PR #7463, opened 2016-10-13
Ensure input services can be safely opened and closed

e-dard:
###### Required for all non-trivial PRs
- [x] Rebased/mergable
- [x] Tests pass

This PR ensures that input services (`udp`, `graphite`, `opentsdb` and `collectd`) can be idempotently opened and closed.

It also fixes a goroutine leak where a listener was not being closed properly.

PR #7477, opened 2016-10-18
Input plugins should retry database creation

e-dard:
- [x] Rebased/mergable
- [x] Tests pass

#7463 needs to be merged first.

This PR changes how input plugins open. Instead of trying to create a database on open they will instead check if they have successfully created a database when processing data.

@jwilder


services/graphite/service.go
```diff
@@ -397,6 +439,9 @@ func (s *Service) processBatches(batcher *tsdb.PointBatcher) {
 	for {
 		select {
 		case batch := <-batcher.Out():
+			// Will attempt to create database if not yet created.
+			s.createInternalStorage()
```
jwilder: Shouldn't this check for an error being returned?

services/opentsdb/service.go
```diff
@@ -408,20 +455,21 @@ func (s *Service) serveHTTP() {
...
 // processBatches continually drains the given batcher and writes the batches to the database.
 func (s *Service) processBatches(batcher *tsdb.PointBatcher) {
-	defer s.wg.Done()
 	for {
 		select {
+		case <-s.done:
+			return
 		case batch := <-batcher.Out():
+			// Will attempt to create database if not yet created.
+			s.createInternalStorage()
```
jwilder: Check for an error?

e-dard: Since there is nothing we can do but log the error, it seems cleaner to have `createInternalStorage` check for an error and log it itself.

jwilder: But then it continues on to try the write which will fail as well and tie up resources.  Seems like we should abort the write here since we already know it will fail.

e-dard: Ah sorry I see what you mean. OK I'll refactor that.

PR #7480, opened 2016-10-18
Memoize output of FileStore.Stats

e-dard:
- [x] Rebased/mergable
- [x] Tests pass

This should significantly reduce the time spent planning compactions when there are large numbers of TSM files within a shard, by memoizing the output of `FileStore.Stats`. When new TSM files are added or removed the output of `FileStore.Stats` will need to be recalculated.

Before:

```
⇒  gt -run NONE -bench BenchmarkFileStore_Stats
BenchmarkFileStore_Stats-4   	    5000	    215589 ns/op	   82094 B/op	       2 allocs/op
PASS
ok  	github.com/influxdata/influxdb/tsdb/engine/tsm1	9.185s
```

After:

```
⇒  gt -run NONE -bench BenchmarkFileStore_Stats
BenchmarkFileStore_Stats-4   	20000000	        99.2 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/influxdata/influxdb/tsdb/engine/tsm1	15.388s
```

@jwilder @benbjohnson 

/cc @beckettsean

### DIFF
--- a/tsdb/engine/tsm1/file_store.go
+++ b/tsdb/engine/tsm1/file_store.go
@@ -124,6 +124,9 @@ const (
 type FileStore struct {
 	mu           sync.RWMutex
 	lastModified time.Time
+	// Most recently known file stats. If nil then stats will need to be
+	// recalculated
+	lastFileStats []FileStat
 
 	currentGeneration int
 	dir               string
@@ -258,6 +261,7 @@ func (f *FileStore) Add(files ...TSMFile) {
 	for _, file := range files {
 		atomic.AddInt64(&f.stats.DiskBytes, int64(file.Size()))
 	}
+	f.lastFileStats = f.lastFileStats[:0] // Will need to be recalculated on next call to Stats.
 	f.files = append(f.files, files...)
 	sort.Sort(tsmReaders(f.files))
 	atomic.StoreInt64(&f.stats.FileCount, int64(len(f.files)))
@@ -285,6 +289,7 @@ func (f *FileStore) Remove(paths ...string) {
 			atomic.AddInt64(&f.stats.DiskBytes, -int64(file.Size()))
 		}
 	}
+	f.lastFileStats = f.lastFileStats[:0] // Will need to be recalculated on next call to Stats.
 	f.files = active
 	sort.Sort(tsmReaders(f.files))
 	atomic.StoreInt64(&f.stats.FileCount, int64(len(f.files)))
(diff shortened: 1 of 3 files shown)


tsdb/engine/tsm1/file_store.go
```diff
@@ -258,6 +261,7 @@ func (f *FileStore) Add(files ...TSMFile) {
 	for _, file := range files {
 		atomic.AddInt64(&f.stats.DiskBytes, int64(file.Size()))
 	}
+	f.lastFileStats = nil // Will need to be recalculated on next call to Stats.
```
jwilder: Lower down in the readers, I just resliced this to `f.lastFileStats = f.lastFileStats[:0]` to avoid the extra allocations.  Maybe do that here and check for `len() == 0`?

tsdb/engine/tsm1/file_store.go
```diff
@@ -449,6 +454,7 @@ func (f *FileStore) Close() error {
 		file.Close()
 	}
 
+	f.lastFileStats = nil
```
jwilder: This also needs to be done in `Replace` which is the primary way that files are added to the `FileStore` via compactions.

tsdb/engine/tsm1/file_store.go
```diff
@@ -598,6 +612,7 @@ func (f *FileStore) Replace(oldFiles, newFiles []string) error {
 	// Tell the purger about our in-use files we need to remove
 	f.purger.add(inuse)
 
+	f.lastFileStats = nil // Will need to be recalculated on next call to Stats.
```
jwilder: `[:0]`

tsdb/engine/tsm1/file_store.go
```diff
@@ -485,12 +491,20 @@ func (f *FileStore) KeyCursor(key string, t int64, ascending bool) *KeyCursor {
 
 func (f *FileStore) Stats() []FileStat {
 	f.mu.RLock()
-	defer f.mu.RUnlock()
+	if len(f.lastFileStats) > 0 {
+		f.mu.RUnlock()
+		return f.lastFileStats
```
jwilder: I think you need to assign this under the read lock and return the assign var or else your accessing the slice outside the lock.

tsdb/engine/tsm1/file_store.go
e-dard: This one is in `Close` so I think it's OK to set it to `nil`.

jwilder:
The rename in your test fails on windows/appveyor because the file is open. Probably need to do something like: https://github.com/influxdata/influxdb/blob/master/tsdb/engine/tsm1/compact_test.go#L2175

PR #7495, opened 2016-10-20
Rewrite some regex conditions to use index-compatible conditions

### LINKED ISSUE #7486: Optimize regexp on single tag values
Queries matching a single tag value using regexp :

```
SELECT myaggregation(myfield) FROM "mymeasurement" WHERE mytag =~ /^mytagvalue$/ AND time > now() - 15m
```

can be optimized to work on a direct tag match instead:

```
SELECT myaggregation(myfield) FROM "mymeasurement" WHERE mytag = 'mytagvalue' AND time > now() - 15m
```

E.g. it is possible for Grafana to generate such queries when selecting a single value from a template (e.g. tag values as templates) because regexps are available in the query editor.

InfluxDB could detect ` =~ /^foo$/` and rewrite that to ` = 'foo'` and turn the linear scan into an index lookup.

e-dard:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

Fixes #7486.

This PR adds support for replacing regexes with non-regex conditions
when possible. Currently the following regexes are supported:

 - `host =~ /^server1$/` will be converted into `host = 'server1'`
 - `host !~ /^server1$/` will be converted into `host != 'server1'`

*Note*: if the regex expression contains repetition, groupings or character classes, then rewriting may not take place. For example, the condition: `name =~ /^foo.$/` will not
be rewritten as it cannot be rewritten as an exact match.

Support for some of these cases may arrive in the future.

Regexes that can be converted into simpler expression will be able to
take advantage of the tsdb index, making them significantly faster.

#### Benchmarks

The following benchmarks run a query along the lines of:
 
```
SELECT count(value) FROM cpu WHERE host = 'server1'
``` 

for the non-regex ones, and: 

```
SELECT count(value) FROM cpu WHERE host =~ /^server1$/
```

 for the regex ones.

`master` 6fd74a6a

```
BenchmarkServer_Query_Count_Where_500-4          	    1000	   1392067 ns/op	  909527 B/op	     572 allocs/op
BenchmarkServer_Query_Count_Where_1K-4           	    1000	   1349177 ns/op	  909332 B/op	     571 allocs/op
BenchmarkServer_Query_Count_Where_10K-4          	    1000	   1564794 ns/op	  908493 B/op	     569 allocs/op
BenchmarkServer_Query_Count_Where_100K-4         	    1000	   2161792 ns/op	  906965 B/op	     566 allocs/op
BenchmarkServer_Query_Count_Where_Regex_500-4    	     100	  21228299 ns/op	 1717204 B/op	     642 allocs/op
BenchmarkServer_Query_Count_Where_Regex_1K-4     	     100	  22877996 ns/op	 1717317 B/op	     642 allocs/op
BenchmarkServer_Query_Count_Where_Regex_10K-4    	      50	  20535616 ns/op	 1717303 B/op	     642 allocs/op
BenchmarkServer_Query_Count_Where_Regex_100K-4   	     100	  18272231 ns/op	 1717068 B/op	     641 allocs/op
PASS
ok  	github.com/influxdata/influxdb/cmd/influxd/run	21.132s
```

This PR produces the following results:

```
BenchmarkServer_Query_Count_Where_500-4          	    1000	   1272334 ns/op	  909657 B/op	     572 allocs/op
BenchmarkServer_Query_Count_Where_1K-4           	    1000	   1316998 ns/op	  909420 B/op	     571 allocs/op
BenchmarkServer_Query_Count_Where_10K-4          	    1000	   1770097 ns/op	  908471 B/op	     569 allocs/op
BenchmarkServer_Query_Count_Where_100K-4         	    1000	   2368152 ns/op	  906952 B/op	     566 allocs/op
BenchmarkServer_Query_Count_Where_Regex_500-4    	    1000	   2053715 ns/op	  913136 B/op	     635 allocs/op
BenchmarkServer_Query_Count_Where_Regex_1K-4     	    1000	   2823591 ns/op	  913161 B/op	     635 allocs/op
BenchmarkServer_Query_Count_Where_Regex_10K-4    	     500	   2273258 ns/op	  913110 B/op	     635 allocs/op
BenchmarkServer_Query_Count_Where_Regex_100K-4   	    1000	   2327778 ns/op	  913072 B/op	     635 allocs/op
PASS
ok  	github.com/influxdata/influxdb/cmd/influxd/run	24.178s
```

So you can see that the regexes have been rewritten and are now effectively running the same benchmarks.

##### Ratings repo

A real world query on the [ratings](github.com/jwilder/ratings) dataset:

`1.0.2`
```
⇒  time influx -database ML1M -execute "SELECT mean(rating) FROM movielens WHERE userid =~ /^48$/"
name: movielens
---------------
time	mean
0	3.068561872909699

influx -database ML1M -execute   0.00s user 0.01s system 0% cpu 1.945 total
```

`PR`
```
⇒  time influx -database ML1M -execute "SELECT mean(rating) FROM movielens WHERE userid =~ /^48$/"
name: movielens
---------------
time	mean
0	3.068561872909699

influx -database ML1M -execute   0.00s user 0.01s system 0% cpu 1.399 total
```

@jwilder @jsternberg

### DIFF
--- a/influxql/ast.go
+++ b/influxql/ast.go
@@ -1241,8 +1241,8 @@ func (s *SelectStatement) RewriteFields(ic IteratorCreator) (*SelectStatement, e
 //
 // Conditions that can currently be simplified are:
 //
-//     - host ~= /^foo$/ becomes host = 'foo'
-//     - host ~! /^foo$/ becomes host != 'foo'
+//     - host =~ /^foo$/ becomes host = 'foo'
+//     - host !~ /^foo$/ becomes host != 'foo'
 //
 // Note: if the regex contains groups, character classes, repetition or
 // similar, it's likely it won't be rewritten. In order to support rewriting
--- a/influxql/ast.go
+++ b/influxql/ast.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
+	"regexp/syntax"
 	"sort"
 	"strconv"
 	"strings"
@@ -1233,6 +1234,93 @@ func (s *SelectStatement) RewriteFields(ic IteratorCreator) (*SelectStatement, e
 	}
 
 	return other, nil
+}
+
+// RewriteRegexExprs rewrites regex conditions to make better use of the
+// database index.
+//
+// Conditions that can currently be simplified are:
+//
+//     - host ~= /^foo$/ becomes host = 'foo'
+//     - host ~! /^foo$/ becomes host != 'foo'
+//
+// Note: if the regex contains groups, character classes, repetition or
+// similar, it's likely it won't be rewritten. In order to support rewriting
+// regexes with these characters would be a lot more work.
+func (s *SelectStatement) RewriteRegexConditions() {
+	s.Condition = RewriteExpr(s.Condition, func(e Expr) Expr {
+		be, ok := e.(*BinaryExpr)
+		if !ok || (be.Op != EQREGEX && be.Op != NEQREGEX) {
+			// This expression is not a binary condition or doesn't have a
+			// regex based operator.
+			return e
+		}
+
+		// Handle regex-based condition.
+		rhs := be.RHS.(*RegexLiteral) // This must be a regex.
+
+		val, ok := matchExactRegex(rhs.Val.String())
+		if !ok {
+			// Regex didn't match.
+			return e
+		}
+
+		// Remove leading and trailing ^ and $.
+		be.RHS = &StringLiteral{Val: val}
+
+		// Update the condition operator.
+		if be.Op == EQREGEX {
+			be.Op = EQ
+		} else {
+			be.Op = NEQ
+		}
+		return be
+	})
+}
+
+// matchExactRegex matches regexes that have the following form: /^foo$/. It
+// considers /^$/ to be a matching regex.
+func matchExactRegex(v string) (string, bool) {
+	re, err := syntax.Parse(v, syntax.Perl)
+	if err != nil {
+		// Nothing we can do or log.
+		return "", false
+	}
+
+	if re.Op != syntax.OpConcat {
+		return "", false
+	}
+
+	if len(re.Sub) < 2 || len(re.Sub) > 3 {
+		// Regex has too few or too many subexpressions.
+		return "", false
+	}
+
+	start := re.Sub[0]
+	if !(start.Op == syntax.OpBeginLine || start.Op == syntax.OpBeginText) {
+		// Regex does not begin with ^
+		return "", false
+	}
+
+	end := re.Sub[len(re.Sub)-1]
+	if !(end.Op == syntax.OpEndLine || end.Op == syntax.OpEndText) {
+		// Regex does not end with $
+		return "", false
+	}
+
+	if len(re.Sub) == 3 {
+		middle := re.Sub[1]
+		if middle.Op != syntax.OpLiteral {
+			// Regex does not contain a literal op.
+			return "", false
+		}
+
+		// We can rewrite this regex.
+		return string(middle.Rune), true
+	}
+
+	// The regex /^$/
+	return "", true
 }
 
 // RewriteDistinct rewrites the expression to be a call for map/reduce to work correctly
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ The following configuration changes in the `[data]` section may need to changed 
 - [#7496](https://github.com/influxdata/influxdb/pull/7496): Filter out series within shards that do not have data for that series.
 - [#7480](https://github.com/influxdata/influxdb/pull/7480): Improve compaction planning performance by caching tsm file stats.
 - [#7320](https://github.com/influxdata/influxdb/issues/7320): Update defaults in config for latest best practices
+- [#7495](https://github.com/influxdata/influxdb/pull/7495): Rewrite regexes of the form host = /^server-a$/ to host = 'server-a', to take advantage of the tsdb index.
 
 ### Bugfixes
 
--- a/cmd/influxd/run/server_bench_test.go
+++ b/cmd/influxd/run/server_bench_test.go
@@ -7,25 +7,17 @@ import (
 	"testing"
 )
 
-func BenchmarkServer_Query_Count_1(b *testing.B) {
-	benchmarkServerQueryCount(b, 1)
-}
+var strResult string
 
-func BenchmarkServer_Query_Count_1K(b *testing.B) {
-	benchmarkServerQueryCount(b, 1000)
-}
-
-func BenchmarkServer_Query_Count_100K(b *testing.B) {
-	benchmarkServerQueryCount(b, 100000)
-}
-
-func BenchmarkServer_Query_Count_1M(b *testing.B) {
-	benchmarkServerQueryCount(b, 1000000)
-}
+func BenchmarkServer_Query_Count_1(b *testing.B)    { benchmarkServerQueryCount(b, 1) }
+func BenchmarkServer_Query_Count_1K(b *testing.B)   { benchmarkServerQueryCount(b, 1000) }
+func BenchmarkServer_Query_Count_100K(b *testing.B) { benchmarkServerQueryCount(b, 100000) }
+func BenchmarkServer_Query_Count_1M(b *testing.B)   { benchmarkServerQueryCount(b, 1000000) }
 
 func benchmarkServerQueryCount(b *testing.B, pointN int) {
-	s := OpenDefaultServer(NewConfig())
-	defer s.Close()
+	if _, err := benchServer.Query(`DROP MEASUREMENT cpu`); err != nil {
+		b.Fatal(err)
+	}
 
 	// Write data into server.
 	var buf bytes.Buffer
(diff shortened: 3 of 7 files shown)


pauldix:
hawt

influxql/ast.go
```diff
@@ -1235,6 +1235,71 @@ func (s *SelectStatement) RewriteFields(ic IteratorCreator) (*SelectStatement, e
...
+	s.Condition = RewriteExpr(s.Condition, func(e Expr) Expr {
+		be, ok := e.(*BinaryExpr)
+		if !ok || (be.Op != EQREGEX && be.Op != NEQREGEX) {
+			// This expression is not a binary condition or doesn't have a
+			// regex based operator.
+			return e
+		}
+
+		// Handle regex-based condition.
+		var matched bool
```
stevenh: There is only one case for matched so why take the hit on the defer, surely a straight inline the switch.

influxql/ast.go
```diff
@@ -1235,6 +1235,71 @@ func (s *SelectStatement) RewriteFields(ic IteratorCreator) (*SelectStatement, e
...
+		defer func() {
+			// If we rewrote a regex then we'll need to update the condition's
+			// operator.
+			if !matched {
+				return
+			}
+			switch be.Op {
```
stevenh: This can be optimised to:
```golang
if be.Op == EQREGEX {
   be.op = EQ
} else {
   be.op = NEQ
}

**Comment by mark-rushakoff** [comment]
Any reason for not using [regexp/syntax](https://golang.org/pkg/regexp/syntax/) to inspect the regular expression, rather than inspecting the string? That would make it more straightforward to handle some of the other one-of-many-exact matches, e.g. `/^(foo|bar)$/`.

File: influxql/ast.go [hunk 3]
```diff
@@ -1235,6 +1235,71 @@ func (s *SelectStatement) RewriteFields(ic IteratorCreator) (*SelectStatement, e
...
+				be.Op = EQ
+			case NEQREGEX:
+				be.Op = NEQ
+			}
+		}()
+
+		rhs := be.RHS.(*RegexLiteral) // This must be a regex.
+		val := rhs.Val.String()
+
+		if strings.ContainsRune(val, '\\') {
```
jwilder: What about other regex special chars?  Can you add some tests for things like `/^foo[12]$/`?

stevenh: Indeed the following, which deals with most basic regexps, may be useful: https://play.golang.org/p/EwEb9rXfX1

All you need to do is rewrite if you get a none empty slice back

influxql/ast.go
e-dard: There is no measurable hit on a `defer` in this context. Query parsing is not a bottleneck when executing a query.

stevenh: That as may be but its actually unnecessary so should be avoided as it will add overhead all be it small and every little bit helps, but mainly as complicates the flow of the code,

influxql/ast.go
e-dard: Thanks @stevenh I'll take a look!

influxql/ast.go
e-dard: The reason it's there is in preparation for when I add further regex rewrites along side this exact match rewrite. Once they're added the `defer` block will be cleaner since there will be multiple matched return points. Happy to remove it until then though, if it's complicating things 👍

stevenh: That makes more sense then, thanks for explaining 👍

influxql/ast.go
stevenh: Be aware its intentionally limited to only do the basics, specifically:
- It only deals with anchored i.e /^....$/
- It only deals with captures at the base level i.e. /^(...)$/
- It only deals with concats containing only literals and char classes
- It only deals with alternatives containing literals and concats

So it doesn't deal with all cases as that would open it up to potential issues but here's some that it does deal with
```go
/^1test$/
/^(tester|wibble|blip)$/
/^(1test[1-2][4-5]|wibble[g-hk])$/
/^(2test1|3test2|3test3|3test5|3test7|3test9|3testx)$/ 
````

Comment (e-dard) [inline influxql/ast.go]: @stevenh I'm currently working on an alternative approach which will just involve rejecting all regexes with meta characters in them (with the exception of beginning and ending with `^` and `$` of course). It will be quite restrictive (probably more restrictive than the playground snippet you linked to?), but easier to review and test.

We can improve on it for `1.2`.

**Comment by e-dard** [comment]
@jwilder @jsternberg updated.

@mark-rushakoff I'll admit I'm pretty clueless about`regexp/syntax`. In the end I just went for a simple regex to identify strings like `^foo$` and then a simple rejection of any regex with a meta character. Can `regexp/syntax` do this for us in as simple a way?

**Comment by daviesalex** [comment]
This looks like a great idea to achieve a free lunch!

This may be out of the scope of this PR, but a really huge win here (that I briefly chatted to Paul about a few weeks ago) would be to add a more optimal "begins with" comparison, and filter those out. 99.9% of our regex's are host =~ /^server1/  style (may be expressed at host =~ /^server.*$/  or host =~ /^server.*/)

There are cases where the rex drops down to cases covered by this PR, but its not common (for us!)

**Comment by e-dard** [comment]
@daviesalex we have your use-case on our minds. I think the main problem preventing us solving the prefix issue with our current in-memory index, is that we can only do direct lookups for series keys. To do prefix lookups we would need to store series in-order.

The new `tsi` index approach over in https://github.com/influxdata/influxdb/tree/tsi could allow us to search through series keys in this way. Assuming that we add query support for prefix matches, we could then potentially rewrite `/^server1/`, `/server1./`, `/server1.+$` etc to use the index-compatible prefix search.

**Comment by jsternberg** [comment]
@e-dard `regexp/syntax actually seems to be pretty simple. I can work on it with you later today if you want. It would certainly be better and would allow us to support more types of optimizations such as character classes more easily.

**Comment by jsternberg** [comment]
Example of some code that I haven't finished yet:

```
func literalMatches(re *regexp.Regexp) ([]string, error) {
    s, err := syntax.Parse(re.String(), 0)
    if err != nil {
        return nil, err
    }

    // We are looking for EXACT matches which means we need
    // a concatenation of a ^, some text, then $.
    if s.Op != syntax.OpConcat && len(s.Sub) < 3 {
        return nil, nil
    }

    start := s.Sub[0]
    if !(start.Op == syntax.OpBeginLine || start.Op == syntax.OpBeginText) {
        return nil, nil
    }
    end := s.Sub[len(s.Sub)-1]
    if !(end.Op == syntax.OpEndLine || end.Op == syntax.OpEndText) {
        return nil, nil
    }

    // Look at the inside to try and put together a string.
    subs := s.Sub[1 : len(s.Sub)-1]
    if len(subs) == 0 {
        return []string{""}, nil
    }

    // now process the inside of the regular expression to try and find exact matches
    return nil, nil
}
```

**Comment by e-dard** [comment]
@jsternberg though it's important to remember that in the scope of this PR we can't support character classes, since they don't map to literal strings. In the future (beyond the scope of this PR) it would be cool if we could take `host =~ /^server-[ab]$/`  and convert that into `host = 'server-a' OR host = 'server-b'` 😃 .

**Comment by mark-rushakoff** [comment]
+1 for the code @jsternberg showed for this, and strong -1 from me on attempting to parse stringified regular expressions, with regular expressions. 

I think adding a check for `len(subs) == 1` and `subs[0].Op == syntax.OpLiteral` would more or less implement the issue as written. I think we'd want to run [(syntax.Regexp).Simplify](https://golang.org/pkg/regexp/syntax/#Regexp.Simplify) at the start of that code, too.

@stevenh linked to [this playground link](https://play.golang.org/p/EwEb9rXfX1) in a now-hidden comment, which seems like a good example of a fuller solution, perhaps for 1.2.

**Comment by stevenh** [comment]
I believe the original use case came from us identifying that grafana generates its queries that reference vars with regexp so you can't (from the gui) get:
```go
mytag = '$var'
```
instead you get:
```go
mytag = /^$var$/
```
However if $var is multiple then this results in:
```go
mytag =/^(value1|value2)$/
```
This seems like a simple case however syntax reduces this on parse to:
```go
mytag = /^value[1-2]$/
```

This is why I prototyped out a bit more in https://play.golang.org/p/EwEb9rXfX1 so we can actually deal with what grafana generates commonly.

influxql/ast.go
```diff
@@ -1235,6 +1236,96 @@ func (s *SelectStatement) RewriteFields(ic IteratorCreator) (*SelectStatement, e
...
+	if err != nil {
+		// Nothing we can do or log.
+		return "", false
+	}
+
+	if re.Op != syntax.OpConcat {
+		return "", false
+	}
+
+	if len(re.Sub) < 2 || len(re.Sub) > 3 {
```
stevenh: So you mean len(re.Sub) != 3 then yes?

influxql/ast.go
```diff
@@ -1235,6 +1236,96 @@ func (s *SelectStatement) RewriteFields(ic IteratorCreator) (*SelectStatement, e
...
+		return "", false
+	}
+
+	start := re.Sub[0]
+	if !(start.Op == syntax.OpBeginLine || start.Op == syntax.OpBeginText) {
+		// Regex does not begin with ^
+		return "", false
+	}
+
+	end := re.Sub[len(re.Sub)-1]
```
stevenh: As this must be 3 in len you could just use literal 2 here.

influxql/ast.go
```diff
@@ -1235,6 +1236,96 @@ func (s *SelectStatement) RewriteFields(ic IteratorCreator) (*SelectStatement, e
...
+	if !(end.Op == syntax.OpEndLine || end.Op == syntax.OpEndText) {
```
stevenh: I think technically the combination of this and the start check is wrong as it could be unbalanced.

influxql/ast.go
```diff
@@ -1235,6 +1236,96 @@ func (s *SelectStatement) RewriteFields(ic IteratorCreator) (*SelectStatement, e
...
+		// Regex does not end with $
+		return "", false
+	}
+
+	var res []byte
+	if len(re.Sub) == 3 {
```
stevenh: This must be the case given the check above
```go
if len(re.Sub) < 2 || len(re.Sub) > 3 {
```

influxql/ast.go
```diff
@@ -1235,6 +1236,96 @@ func (s *SelectStatement) RewriteFields(ic IteratorCreator) (*SelectStatement, e
...
+		middle := re.Sub[1]
+		if middle.Op != syntax.OpLiteral {
+			// Regex does not contain a literal op.
+			return "", false
+		}
+
+		// We can rewrite this regex.
+		res = make([]byte, 0, len(middle.Rune))
```
stevenh: yes you can

influxql/ast.go
```diff
@@ -1235,6 +1236,96 @@ func (s *SelectStatement) RewriteFields(ic IteratorCreator) (*SelectStatement, e
...
+	})
+}
+
+// matchExactRegex matches regexes that have the following form: /^foo$/. It
+// considers /^$/ to be a matching regex.
+func (s *SelectStatement) matchExactRegex(v string) (string, bool) {
+	re, err := syntax.Parse(v, syntax.Perl)
+	if err != nil {
+		// Nothing we can do or log.
+		return "", false
```
stevenh: For repeat returns of all Zero values like this, declaring the variables in the return declaration so you can just do "return" could be considered cleaner.

influxql/ast.go
e-dard: No I mean if there are fewer than 2 or more than 3 subexpressions. We will want to catch one regex with two subexpressions, namely `/^$/`, because that would be more efficiently represented as ` = ''` than by using a regex.

influxql/ast.go
e-dard: See comment above.

influxql/ast.go
e-dard: Not sure I follow? Could you elaborate?

mark-rushakoff: I believe @stevenh is pointing out the code currently would accept OpBeginLine with OpEndText, or OpBeginText with OpEndLine, instead of only accepting Line/Line and Text/Text.

jsternberg: That's likely fine. I'm not 100% sure, but `OpBeginText` is likely `\A` and that's fully compatible with `OpEndLine` which is `$`. I don't believe they have to match. If we want to normalize it, I think we can use the flag `syntax.OneLine` so it's always `OpBeginText` and `OpEndText` instead of checking for both.

influxql/ast.go
stevenh: Ahh right, I see that's good :)

PR #7496, opened 2016-10-20
Filter out series within shards that do not have data for that series

jsternberg:
Previously, we would return a full tag set for every shard and the tag
set would include all series that existed in the database index
including series that didn't physically exist within that shard. This
led to the tag sets returned being incredibly huge when we had high
cardinality but sparse data. Since the data was sparse, it was
unexpected that it would cause such a large strain on the system by most
people.

Now we filter out the series ids that are not assigned to the current
shard when computing a tag set for that shard. This lowers the memory
usage for high cardinality sparse data drastically and allows queries on
those to complete successfully.

This does not resolve issues for high cardinality data in every shard
that is also spread out over a long series of time. That situation isn't
nearly as common as the above situation though.


jwilder:
I think this change may not be needed w/ TSI, but is need for the current index.  @e-dard @benbjohnson can you take a look?

benbjohnson:
Correct, it's not necessarily for TSI since it'll only know about its series internally.

PR #7515, opened 2016-10-25
Return parse error from delete/drop when db or rp is specified

### LINKED ISSUE #7053: Delete statement ignores a retention policy
### Bug report

Delete statement ignores a specified retention policy and deletes data across all retention policies.

__System info:__ InfluxDB 1.0 beta3, InfluxDB nightly (1.1.0~n201607220800) Ubuntu 14.04

__Steps to reproduce:__

```
# create database
curl http://localhost:8086/query --data-urlencode "q=CREATE DATABASE mydb"

# create non-default retention policy myrp
curl http://localhost:8086/query --data-urlencode "q=CREATE RETENTION POLICY myrp ON mydb DURATION 365d REPLICATION 1"

# write data to default rp
curl -X POST "http://localhost:8086/write?db=mydb" --data-binary "cpu,region=useast,host=server_1,service=redis value=61"

# write data to myrp
curl -X POST "http://localhost:8086/write?db=mydb&rp=myrp" --data-binary "cpu,region=useast,host=server_1,service=redis value=62"

# delete data from myrp
curl "http://localhost:8086/query?db=mydb" --data-urlencode 'q=DELETE FROM myrp.cpu'
# alternatively tried
# curl "http://localhost:8086/query?db=mydb&rp=myrp" --data-urlencode 'q=DELETE FROM myrp.cpu'

# Query default rp
# Bug: expected it to return data, but no data comes back.
curl -G "http://localhost:8086/query?db=mydb" --data-urlencode 'q=SELECT * from cpu'
```

__Expected behavior:__ Data deleted only from `myrp` retention policy

__Actual behavior:__ Data deleted in all retention policies

jwilder:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

The delete and drop statements apply to the measurement within a db.
The parser allowed a db or rp to be specified and these values were
silently ignored.  This could cause data loss as someone would think
they are only deleting the series within a rp, but they are actually
deleting all their data.

Instead, we return a parse error if a db or rp is specified in the
delete or drop statements.  Ideally, we'd be able to respect the
db and rp, but that requires significant work in the query engine
and tsdb store to make that work.

Fixes #7053

cc @jsternberg


influxql/parser.go
```diff
@@ -1030,6 +1030,25 @@ func (p *Parser) parseDeleteStatement() (Statement, error) {
 		if stmt.Sources, err = p.parseSources(); err != nil {
 			return nil, err
 		}
+
+		var err error
+		WalkFunc(stmt.Sources, func(n Node) {
+			if t, ok := n.(*Measurement); ok {
+				// Don't allow database or retention policy in from clause for delete
```
jsternberg: I think the database is supported. It's retention policy that isn't and causes the issue.

CHANGELOG.md
```diff
@@ -59,6 +59,7 @@ The following configuration changes in the `[data]` section may need to changed
 - [#7334](https://github.com/influxdata/influxdb/issues/7334): Panic with unread show series iterators during drop database
 - [#7482](https://github.com/influxdata/influxdb/issues/7482): Fix issue where point would be written to wrong shard.
 - [#7431](https://github.com/influxdata/influxdb/issues/7431): Remove /data/process_continuous_queries endpoint.
+- [#7053](https://github.com/influxdata/influxdb/issues/7053): Delete statement ignores a retention policy
```
jsternberg: I think this should say something like "delete statement returns error when retention policy is specified" since that describes the fix rather than the problem.

jwilder: From the docs, it's just measurement in a database.  The code doesn't seem to work when a DB is specified.

jsternberg: Just tried it out and it seems that's true. We'll fix that in a future version.

jwilder: Updated changelog.

PR #7519, opened 2016-10-25
Change default time boundaries for raw queries

### LINKED ISSUE #4461: [feature request] change default time boundaries
To address https://github.com/influxdb/influxdb/issues/2703, https://github.com/influxdb/influxdb/issues/3369, https://github.com/influxdb/influxdb/issues/3789, and to clear up confusion about https://influxdb.com/docs/v0.9/troubleshooting/frequently_encountered_issues.html#querying-after-now, it would make sense to change the default lower bound from `epoch 0` to the smallest timestamp and the default upper bound from `now()` to the largest possible timestamp.

I believe the smallest is -9023372036854775808, although https://github.com/influxdb/influxdb/issues/3367 means that's not valid right now. The largest timestamp is 9023372036854775807. These were derived from testing, not from parsing code.

jsternberg:
Changes the default time boundaries for raw queries so raw queries will
range until the end of time. Aggregate queries continue to have their
default end time be `now()`.

Fixes #4461.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ The following configuration changes in the `[data]` section may need to changed 
 - [#7320](https://github.com/influxdata/influxdb/issues/7320): Update defaults in config for latest best practices
 - [#7495](https://github.com/influxdata/influxdb/pull/7495): Rewrite regexes of the form host = /^server-a$/ to host = 'server-a', to take advantage of the tsdb index.
 - [#6704](https://github.com/influxdata/influxdb/issues/6704): Optimize first/last when no group by interval is present.
+- [#4461](https://github.com/influxdata/influxdb/issues/4461): Change default time boundaries for raw queries.
 
 ### Bugfixes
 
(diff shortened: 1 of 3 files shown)


jsternberg:
@rkuchan this change will need to be documented.

beckettsean:
@jsternberg is my understanding correct?

```
SELECT * FROM foo			# will select until max(int64)-1
SELECT MEAN(*) FROM foo		# will select until now()
SELECT MAX(*) FROM foo		# will select until now()
```

jsternberg:
I need to clarify the changelog since that's wrong. This is it:

```
SELECT * FROM foo                                   # will select until max(int64)-1
SELECT MEAN(*) FROM foo                             # will select until max(int64)-1
SELECT MAX(*) FROM foo                              # will select until max(int64)-1
SELECT MAX(*) FROM foo WHERE ... GROUP BY time(...) # will select until now()

beckettsean:
Thanks for the clarification, @jsternberg. 

The original comment has:`Aggregate queries continue to have their default end time be now()`, but if `SELECT MEAN(*) FROM foo` doesn't use `now()`, should that sentence in the comment be more like `Queries with a group by time continue to have the default end time of now()`?

PR #7527, opened 2016-10-27
updating status on httpd handler for partial write

sebito91:
- [ ] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

Changing the behaviour of a partial-write to the backend. When writing to the httpd endpoint, the handler returns a 400 in all cases where this is an error of any kind. This unfortunately is a misnomer and leads to false-negatives in the following cases:
1. when a write is partially completed (i.e. 3 of 4 measurements commit fine, measurement 4 is malformed) we actually increment the PointsWrittenOk by the len of the 3 successful writes only to return complete failure to the client
2. when actually sending broken data alongside good data, you cannot tell which packet is actually bad unless you parse resp.Body at all times (which you most likely won't be doing if writing via telegraf)

If we're trying to troubleshoot situations where resp.StatusCode != 204, figuring out the difference between various 400 replies is difficult and time consuming. Figuring out the difference between 206 and a true 400 though can dramatically speed up debug.

### DIFF
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -632,9 +632,9 @@ func (h *Handler) serveWrite(w http.ResponseWriter, r *http.Request, user *meta.
 	} else if parseError != nil {
 		// We wrote some of the points
 		atomic.AddInt64(&h.stats.PointsWrittenOK, int64(len(points)))
-		// The other points failed to parse which means the client sent invalid line protocol.  We return a 400
-		// response code as well as the lines that failed to parse.
-		h.httpError(w, fmt.Sprintf("partial write:\n%v", parseError), http.StatusBadRequest)
+		// The other points failed to parse which means the client sent invalid line protocol.  We return a 202
+		// response code as well as the lines that failed to parse (partial write).
+		h.httpError(w, fmt.Sprintf("partial write:\n%v", parseError), http.StatusAccepted)
 		return
 	}
 


services/httpd/handler.go
```diff
@@ -632,9 +632,9 @@ func (h *Handler) serveWrite(w http.ResponseWriter, r *http.Request, user *meta.
 	} else if parseError != nil {
 		// We wrote some of the points
 		atomic.AddInt64(&h.stats.PointsWrittenOK, int64(len(points)))
-		// The other points failed to parse which means the client sent invalid line protocol.  We return a 400
-		// response code as well as the lines that failed to parse.
-		h.httpError(w, fmt.Sprintf("partial write:\n%v", parseError), http.StatusBadRequest)
+		// The other points failed to parse which means the client sent invalid line protocol.  We return a 206
+		// response code as well as the lines that failed to parse (partial write).
+		h.httpError(w, fmt.Sprintf("partial write:\n%v", parseError), http.StatusPartialContent)
```
jwilder: While a `206` makes sense in that it is used for partial content, I believe we also have to return a `Content-Range` header according to the HTTP spec.  Maybe a `202` would be more appropriate which does have this requirement and is still different from `400`.

@joelegasse what do you think?

sebito91: A 202 would be fine by me, something that at least lets us know we completed a part of the payload. Committing in two mins...

sebito91:
This missed the 1.1 release?

sebito91:
Closing based on feedback from @jwilder

PR #7532, opened 2016-10-27
Don't normalize drop/delete series statements

jwilder:
- [x] Rebased/mergable
- [x] Tests pass

7093 causes a parse error to be returned from delete and drop
statements.  Normalizing them cause an invalid statement to be generated
which cannot be reparse if converted to a string and back.

@jsternberg

### DIFF
--- a/internal/meta_client.go
+++ b/internal/meta_client.go
@@ -19,7 +19,7 @@ type MetaClientMock struct {
 	CreateUserFn                        func(name, password string, admin bool) (*meta.UserInfo, error)
 
 	DatabaseFn  func(name string) *meta.DatabaseInfo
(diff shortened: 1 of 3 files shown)


internal/meta_client.go
```diff
@@ -19,7 +19,7 @@ type MetaClientMock struct {
 	CreateUserFn                        func(name, password string, admin bool) (*meta.UserInfo, error)
 
 	DatabaseFn  func(name string) *meta.DatabaseInfo
-	DatabasesFn func() ([]meta.DatabaseInfo, error)
+	DatabasesFn func() []meta.DatabaseInfo
```
corylanou: This looks unrelated to this PR?  Either update PR comments or move to different PR?

PR #7538, opened 2016-10-28
Add config option to messages when limits exceeded

jwilder:
- [x] Rebased/mergable
- [x] Tests pass

When a limit is exceeded, we return errors and sometimes log (if appropriate)
that a limit was exceeded.  The messages don't always provide an indication
as to where or how they are configured.

Instead, return the config option (easily searchable for) as well as the limit
currently set and the value that exceeded it when possible.


jwilder:
I'd like to try the the shorter messages for now.  We can always change it later if we still get questions about these error messages.

PR #7545, opened 2016-10-30
Add note about stability of master branch

corylanou:
- [ ] Rebased/mergable

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # InfluxDB [![Circle CI](https://circleci.com/gh/influxdata/influxdb/tree/master.svg?style=svg)](https://circleci.com/gh/influxdata/influxdb/tree/master) [![Go Report Card](https://goreportcard.com/badge/github.com/influxdata/influxdb)](https://goreportcard.com/report/github.com/influxdata/influxdb) [![Docker pulls](https://img.shields.io/docker/pulls/library/influxdb.svg)](https://hub.docker.com/_/influxdb/)
 
+**Note**: The master branch may be in an unstable or even broken state during development. Please use one of the [pre-built packages](https://influxdata.com/downloads/#influxdb) instead of the master branch in order to get stable binaries.
+
 ## An Open-Source Time Series Database
 
 InfluxDB is an open source **time series database** with


README.md
```diff
@@ -1,5 +1,7 @@
 # InfluxDB [![Circle CI](https://circleci.com/gh/influxdata/influxdb/tree/master.svg?style=svg)](https://circleci.com/gh/influxdata/influxdb/tree/master) [![Go Report Card](https://goreportcard.com/badge/github.com/influxdata/influxdb)](https://goreportcard.com/report/github.com/influxdata/influxdb) [![Docker pulls](https://img.shields.io/docker/pulls/library/influxdb.svg)](https://hub.docker.com/_/influxdb/)
 
+**Note**: The master branch may be in an unstable or even broken state during development. Please use one of the [pre-built packages](https://influxdata.com/downloads/#influxdb) instead of the master branch in order to get stable binaries.
```
pauldix: Is this really true? Sounds like a bit of fear-mongering to  me. We never merge in stuff that is knowingly broken. Master should always be green, thus to the best of our knowledge, it works. It's just not the battle-hardened super tested stable release.

e-dard: @corylanou I would probably reword to `The master branch may be in an unstable state during development. ...`

corylanou: I don't think anything is not true in the statement.  We certainly have had things in a broken state in the past and not known it.  This is not about fear mongering, it's about letting people know the risks they are assuming by using the master branch.  Of course we try to make sure it isn't broken, but in no way is the master branch tested to the extent that our releases are.  Master being `green` is in no way an indication that it doesn't have the potential for serious bugs that we aren't aware of.

I find it's easier to set the bar low and then exceed it.  However, setting the bar high and missing it leads to bad reddit threads.

pauldix: Fair enough. Just didn't want to give the impression that we were knowingly merging in broken stuff. As in, "we never merge anything that isn't green, but master could be broken or have bugs that we don't know about. Most users should be running the latest stable build that can be found here..."

corylanou:
Closing this as I don't feel like we came to a consensus.  Will try again in the future.

PR #7549, opened 2016-10-31
Fix output duration units for SHOW QUERIES

### LINKED ISSUE #7548: show queries output has wrong units
### Bug report

__System info:__ 

InfluxDB 1.0.x

__Steps to reproduce:__

`SHOW QUERIES` elapsed times use incorrect units

__Actual behavior:__

elapsed times are given in nanoseconds or other units, but marked as microseconds.

```
qid query database duration 
329183 SHOW QUERIES 45042u 
329181 SELECT * FROM system WHERE time > now() - 1w telegraf 817262117u

qid query database duration 
329185 SHOW QUERIES 63741u 
329181 SELECT * FROM system WHERE time > now() - 1w telegraf 1s

qid query database duration 
329181 SELECT * FROM system WHERE time > now() - 1w telegraf 2s 
329187 SHOW QUERIES 102072u
```

jsternberg:
The previous version was showing the microseconds unit when it was
outputting nanoseconds. Now we correctly identify which sub-second unit
to use (milliseconds, microseconds, or nanoseconds) and use the correct
unit while dividing the duration unit correctly to produce the correct
output.

Also updated to use the default duration string instead of our own
custom formatters. It turns out that the string method for
`time.Duration` does the correct thing as long as we truncate the value
first.

Fixes #7548.

### DIFF
--- a/influxql/task_manager.go
+++ b/influxql/task_manager.go
@@ -95,15 +95,16 @@ func (t *TaskManager) executeShowQueriesStatement(q *ShowQueriesStatement) (mode
 	for id, qi := range t.queries {
 		d := now.Sub(qi.startTime)
 
-		var ds string
-		if d == 0 {
-			ds = "0s"
-		} else if d < time.Second {
-			ds = fmt.Sprintf("%du", d)
-		} else {
-			ds = (d - (d % time.Second)).String()
+		switch {
+		case d >= time.Second:
+			d = d - (d % time.Second)
+		case d >= time.Millisecond:
+			d = d - (d % time.Millisecond)
+		case d >= time.Microsecond:
+			d = d - (d % time.Microsecond)
 		}
-		values = append(values, []interface{}{id, qi.query, qi.database, ds})
+
+		values = append(values, []interface{}{id, qi.query, qi.database, d.String()})
 	}
 
 	return []*models.Row{{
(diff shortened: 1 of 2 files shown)


influxql/task_manager.go
```diff
@@ -95,15 +95,15 @@ func (t *TaskManager) executeShowQueriesStatement(q *ShowQueriesStatement) (mode
...
 		d := now.Sub(qi.startTime)
 
-		var ds string
-		if d == 0 {
-			ds = "0s"
-		} else if d < time.Second {
-			ds = fmt.Sprintf("%du", d)
-		} else {
-			ds = (d - (d % time.Second)).String()
+		if d >= time.Second {
```
e-dard: It would be [idiomatic](https://golang.org/doc/effective_go.html#switch) to write this as:

```go
switch {
  case d >= time.Second:
      d = d - (d % time.Second)
  case d >= time.Millisecond:
      d = d - (d % time.Millisecond)
  case d >= time.Microsecond:
      d = d - (d % time.Microsecond)
}
```

but it's not a blocker for me 👍

jsternberg: Updated to use `switch` instead. I'll merge when it hits green unless something else is found. Thanks.

Issue #7222, opened 2016-08-26
Low WAL performance for many small write requests


mhamrle:
### Feature Request

**Proposal:** Reduce number of fsync calls

**Current behavior:** WAL writes write requests into a file and calls fsync after every written request.

**Desired behavior:** I would like to reduce number of fsync calls during writing into WAL file.

**Use case:** :
I am using Influxdb in in small deployment with 15 services. From each service I need to send metrics into one central influxdb server. Even it is not many metrics the services generates many http write requests and these requests significantly load influxdb server. After a while I figured out that the load is caused by queued write request into WAL. WAL writes one request into a file and then calls fsync. Tracing linux storage subsystem showed up that writing data into file is negligible compared to fsync. Fsync takes about 20ms on SSD disc, data writing about 2ms.

So it clear that I cannot expect more that 50 http requests per second because influxdb spends all time only by fsyncing  20ms \* 50 = 1000ms. Big improvement can be achieved by reducing number of fsyncs. Instead of calling fsync after every write the fsync will be called after given time interval - for example 100ms. So WAL will process incoming write requests and write them sequentially into file (without fsync), all written requests would wait for one collective fsync. After given interval fsync is called and all waiting finished requests would return with success.

It is trade off between low processing time of http requests and total throughput.

bitwalker:
I just wanted to add in my own observations using InfluxCloud - the average write request hovers between 50ms and 150ms, often taking as long as 500ms-1s. We have data coming from a large number and variety of sources, in most cases a single point at a time, in others the opportunity to batch writes is possible (and something I'm implementing at the application layer), but it does seem like we're more or less capped at ~10 requests per second before we plateau. I can get ~70 reqs/sec testing against a local Influx instance, but when you factor in talking across the internet, and against a cluster instead of a standalone instance, the ~7x performance penalty seems high, but hard to know how much of that is spent where (latency just reaching InfluxCloud, time spent fsyncing the writes, replication/consensus, serialization, etc.)

I was really hoping Influx could handle a much higher level of concurrency than this with small writes, but it looks like that is not a thing which will change soon. Anyway, hopefully some extra data will help.

jwilder:
Fixed via #8143

PR #7565, opened 2016-11-02
Improve WAL performance

mhamrle:
This commit improve WAL performance for use case when influxdb handles
many small write requests. In old code after every write the WAL file is
synced. Sync is expensive operation that takes significant time.
In new code WAL file is synced less frequently. Once per 100ms on high
load or immediately if write load is low (there are no other waiting requests).

There is significant performance improvement tested on stress test
(BatchSize=50, WriteConcurrency=150) 2000 requests, average request size
is 2760B. Old code handles 2000 points per second, new code handles
98300 requests per second. So new code has 50x better write performance.

Closing #7222

### DIFF
--- a/pkg/limiter/fixed.go
+++ b/pkg/limiter/fixed.go
@@ -16,3 +16,7 @@ func (t Fixed) Take() {
 func (t Fixed) Release() {
 	<-t
 }
+
+func (t Fixed) InUse() int {
+	return len(t)
+}
--- a/tsdb/engine/tsm1/wal.go
+++ b/tsdb/engine/tsm1/wal.go
@@ -307,6 +307,16 @@ func (l *WAL) LastWriteTime() time.Time {
 }
 
 func (l *WAL) writeToLog(entry WALEntry) (int, error) {
+	id, ch, err := l.writeToLogDo(entry)
+	if err != nil {
+		return -1, err
+	}
+	// wait for writer is synced
+	<-ch
+	return id, nil
+}
+
+func (l *WAL) writeToLogDo(entry WALEntry) (int, chan struct{}, error) {
 	// limit how many concurrent encodings can be in flight.  Since we can only
 	// write one at a time to disk, a slow disk can cause the allocations below
 	// to increase quickly.  If we're backed up, wait until others have completed.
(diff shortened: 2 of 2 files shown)


daviesalex:
I have not looked at the code, but this gets a +1 from us. Risking a few hundred ms of data in exchange for massively higher performance is a tradeoff we are certainly prepared to make.

jwilder:
Are you using SSD or HDD?  I don't see the same improvement using SSDs based on the benchmarks I have in this [gist](https://gist.github.com/jwilder/e1260ac5fec90e97ffa67710eb4a356f).

```
go test -v -run non -bench BenchmarkWAL_Write_1 -benchmem ./tsdb/engine/tsm1/ > /tmp/new.txt
benchcmp /tmp/old.txt /tmp/new.txt
benchmark                  old ns/op     new ns/op     delta
BenchmarkWAL_Write_1-8     64741         62046         -4.16%

benchmark                  old allocs     new allocs     delta
BenchmarkWAL_Write_1-8     2              3              +50.00%

benchmark                  old bytes     new bytes     delta
BenchmarkWAL_Write_1-8     13            109           +738.46%
```

Also, I get a deadlock if I increase concurrency (running `BenchmarkWAL_Write_10`) benchmark.  All goriness are blocked at:

```
goroutine 73 [chan receive]:
github.com/influxdata/influxdb/tsdb/engine/tsm1.(*WAL).writeToLog(0x4206b8000, 0x532d40, 0x421009860, 0x1, 0x0, 0x0)
	/Users/jason/go/src/github.com/influxdata/influxdb/tsdb/engine/tsm1/wal.go:315 +0x7f
github.com/influxdata/influxdb/tsdb/engine/tsm1.(*WAL).WritePoints(0x4206b8000, 0x420674510, 0x1, 0x0, 0x0)
	/Users/jason/go/src/github.com/influxdata/influxdb/tsdb/engine/tsm1/wal.go:233 +0x72
github.com/influxdata/influxdb/tsdb/engine/tsm1_test.benchmarkWALWrite.func1(0x420670510, 0x4206ba140, 0x4206b8000, 0x420674510)
	/Users/jason/go/src/github.com/influxdata/influxdb/tsdb/engine/tsm1/wal_test.go:746 +0x89
created by github.com/influxdata/influxdb/tsdb/engine/tsm1_test.benchmarkWALWrite
	/Users/jason/go/src/github.com/influxdata/influxdb/tsdb/engine/tsm1/wal_test.go:750 +0x659
```
 
It looks likes the `syncWaiter` channel is never closed if another write does not come in so all pending writes will wait indefinitely.  

I think this could be fixed using a `time.Ticker` that fires every 100ms and sync if necessary.

PR #7576, opened 2016-11-03
Client failover to meta commands when stdin is empty in non-interactive mode

### LINKED ISSUE #7575: influx client non-interactive mode doesn't accept meta commands anymore
### Bug report

__System info:__

influxdb 1.1.0-rc1 in a Docker container based on Alpine linux 3.4

__Steps to reproduce:__

1. Build a Docker image in which influxd is launched in background and the influx CLI is used to execute a query (or import a file)
2. Design a docker-compose file containing an influxdb service and other services depending on the influxdb service
3. Start the influxdb service (docker-compose run influxdb)
4. Stop it (docker-compose down)
5. Start an other service depending on influxdb (docker compose run sut)

__Expected behavior:__

The CLI commands are executed on the influxdb server, and queries are seen in the influxdb container logs.

__Actual behavior:__

The expected behavior is observed on step 3.
The expected behavior is not observed on step 5, influxdb container logs complain about an empty `q` parameter.

__Additional info:__

Until influxdb 1.0.2 this was working as expected, the regression is observed since commit [2f5f9957828c09e73bbc725729f7da9cd66633be](https://github.com/influxdata/influxdb/commit).

The Docker image and the docker-compose.test.yml files are implemented in the [appcelerator/influxdb](https://github.com/appcelerator/docker-influxdb/tree/influxdb-1.1) repo, branch influxdb-1.1. A patch is included to workaround the issue.

ndegory:
resolves #7575

### DIFF
--- a/cmd/influx/cli/cli.go
+++ b/cmd/influx/cli/cli.go
@@ -90,7 +90,10 @@ func (c *CommandLine) Run() error {
 		if err != nil {
 			return err
 		}
-		return c.ExecuteQuery(string(cmd))
+		if string(cmd) != "" {
+			return c.ExecuteQuery(string(cmd))
+		}
+		// no stdin, switch back to meta commands
 	}
 
 	if !c.IgnoreSignals {


jsternberg:
Already resolved by #7586.

Issue #7445, opened 2016-10-11
Time weighted average/mean


deandob:
Proposal:
Provide an aggregate function that provides a time weighted mean. Similar to the interpolation function in OpenTSB [http://opentsdb.net/docs/build/html/user_guide/query/aggregators.html](url)

I am new to influxdb and possibly I have not understood how this can be done, but from spending an hour researching this it looks like it is not natively possible in Influxdb and without the possibility of doing user defined functions this type of calculation has to be done in the host application, negating some of the benefits and performance of influxdb.

Use Case:
For IoT there are a lot of sensor scenarios where data isn't sent on a regular time basis, mostly where the sensors are reacting to changes (which saves battery and network bandwidth compared to sending the same unchanging data value in regular intervals). For example, temperature measurements won't change while a freezer door is closed (which is most of the time) sending no data, but will send many data points when the door is opened for a short while (getting warmer) then closed (getting colder again). A conventional mean will not represent the average temperature in the freezer as most of the sensor data ingested is from the short time the door was open so needs to be weighted by the time interval between sensor readings.

Apologies if there is already a feature request for this but I couldn't find anything similar in the GH issues log. I like the philosophy behind influxdb and the active development - so hoping this is a feature that could be added. Also surprised that this isn't a more requested feature as it should be a common use case. I don't want to use Kapacitor as my servers are Windows not Linux.

Thanks for listening.

jsternberg:
I think you might be able to do something similar to this using `FILL()`. You, unfortunately, can't use fill unless you use an aggregate, but you might be able to use `moving_average` along with `mean` and an aggregate interval to get something similar to what you want. We still haven't added a weighted moving average yet though which seems like it would be better for your purposes.

deandob:
Hi,

Fill doesn't really help as it interpolates to produce a linear time series (values at regular time intervals based on interpolated values) from a delta time series which would give incorrect readings if the changes are step changes. See the examples below.

Sensors sends the following in response to an environment changes at 10:23 and 4 am:
10:20am value = 3
10:23am value = 10
10:24am value = 8
If we were to get a time weighted average between 10:20am and 10:25am the calculation is:
(3 x 3min + 10 x 1min + 8 * 1min)  /  5min =  5.4

If we were to use FILL then a mean we would get:
10:20am value = 3
10:21am value = 5.33 (interpolated)
10:22am value = 7.66 (interpolated)
10:23am value = 10
10:24am value = 8
[...]

PR #6508, opened 2016-04-29
Implement INTEGRAL() to show the sum over time

slamplug:
###### Required for all non-trivial PRs
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated
- [x] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

###### Required only if applicable
_You can erase any checkboxes below this note if they are not applicable to your Pull Request._

Issue #5930, opened 2016-03-07
[[feature collection]] Functions


beckettsean:
This issue contains a list of related feature requests that are not on the near-term roadmap. The feature requests in this issue are all new functions that have been requested. If you want to request a function not already listed please make a comment on this issue, and we will add it to the checklist.

- [ ] moving averages function: https://github.com/influxdata/influxdb/issues/77
- [ ] lag variables https://github.com/influxdata/influxdb/issues/142
- [ ] 
- [ ] 
- [ ] 
- [ ] 
- [ ] 
- [ ]

tbowmo:
Is there any timeline for these functions? (I'm very interested in the aggregate integral functions, to calculate kWh from watts). Seems that a feature request was opened a year ago.

beckettsean:
There is no timeline for each specific function/feature. All work on new functions was on hold while the query engine was refactored, and that refactoring was merged into InfluxDB 0.11. We plan to introduce a few functions with each release from now on.

timgriffiths:
Would also be great to get #3633 Added to your list this feature would be very useful on the client side when graphing data.

beckettsean:
Thanks, @timgriffiths, added!

dengliu:
hi, would like to see some build-in time series anomaly detection functions in influx QL.
E.g. https://blog.twitter.com/2015/introducing-practical-and-robust-anomaly-detection-in-a-time-series
http://robjhyndman.com/hyndsight/yahoo-data/

beckettsean:
@dengliu can you open a new issue so I can reference it here?

dengliu:
@beckettsean I have updated the issue id in my previous comment.

beckettsean:
@dengliu as referenced in the other issue, anomaly detection functions should go into Kapacitor. The results can be fed back into InfluxDB for visualization, but that doesn't require any new functionality in InfluxDB.

nelg:
Awesome work.  I'm +1 for #813  -  I've got data in Influxdb that I'd like to run it on.

pavelpep:
+1 for exponents and logarithms #659

brandoncazander:
The lack of a mode function is really hurting us in switching from 0.8.9. In particular, we need to be able to select the most common occurrence of a specific string over a group by, which we have to do in post-processing now.

beckettsean:
[...]

Issue #1400, opened 2015-01-28
Aggregate Integral Function


gesquive:
A function we use often in graphite is [integral](http://graphite.readthedocs.org/en/latest/functions.html#graphite.render.functions.integral). It's pretty self explanitory. We really want to completely replace graphite with influxdb and this would really help our case.

balistof:
Provide also an aggreate Integral Average function. So that you not take the Mean (sum of values divided by count) but instead use the time as the weight. Basically like Integral but dividing the Integral value by the group by time() range.

C4ptainCrunch:
Any update on this ?

martinpallmann:
I also like to know if there is happening something regarding this ticket. We're using influx right know and the missing integral function is a big problem for us.

beckettsean:
@martinpallmann it is not yet scheduled for a particular release, so it is 6+ weeks away. Watch this issue for updates to the Milestone. That's out backlog.

michaelsproul:
I've implemented an integration aggregation function in #3344 and would appreciate feedback and help getting it merged :smile:

hotsnow77:
+1

enavarro222:
+1
This will be very useful to get energy consumption (Kwh) from power logging (W)

markusr:
@enavarro222 
We use:  ``` SELECT sum(power)/count(power) FROM "data" WHERE $timeFilter GROUP BY time(1h) ```
to calculate the energy consumption in [kWh]. power is [W]

carlosandroni:
@enavarro222 
Your solution works only if you have a constant sample time.

pakozm:
I'm also interested in this feature... Is it planed to be introduced in InfluxDB?

oluies:
+1

sischnei:
+1: Would allow to count the amount of events processed!!

beckettsean:
As mentioned in my [post to the mailing list](https://groups.google.com/d/msg/influxdb/CkKzZoXDgiU/5E-vDNVGCwAJ) we are experimenting with simplifying our open GitHub Issues. This feature request has been rolled into an [aggregate issue](https://github.com/influxdata/influxdb/issues/5930) for all function requests, so that we can close this issue until we are ready to work on it. 

You may continue to make comments here. Closing the issue does not mean we are rejecting this idea.

lukecyca:
+1

lbuchy:
+1

JarrettR:
+1

[...]

PR #7591, opened 2016-11-06
Add "integral" function to InfluxQL

Tomcat-Engineering:
- [x] Rebased/mergable
- [x] Tests pass

- [x] Provide example syntax

I'll open a pull request against the documentation repository, but essentially this is an aggregator function which returns a single value for the "area under the curve".  

You can optionally specify a time interval argument which determines the units of the returned values.  This defaults to one second, so

    SELECT integral(value) from power_in_watts

will return the energy in units of watt-seconds, whereas

    SELECT integral(value, 1h) from power_in_watts

will return the energy in watt-hours.

I think the only potentially contentious/confusing thing is the behaviour at the "edges" of a time period.  As for other InfluxQL functions, we only analyse the datapoints within the specified period (or group-by interval).   This means that the integral does not include the time between the start of the interval and the first datapoint, nor the time between the last datapoint and the end of the interval.  

This will lead to slight inconsistencies, such as a set of daily integral values not quite adding up to the corresponding weekly integral.  The only way to solve that would be to look outside the time period being analysed to find the neighbouring points, so that we can interpolate the curve to the edges of the period, but I don't see a mechanism for doing that.  It could be potentially very expensive, as you might have to search forwards and back through all time in case there were points to use.

We already have some slight inconsistencies of a similar nature - for example the `derivative` function doesn't produce a value for the first point in each interval (for the same reason) so a set of derivative series calculated for each day doesn't quite match the result of analysing the whole week at once.

I propose that we just explain the above limitation in the documentation for now.

### DIFF
--- a/influxql/call_iterator.go
+++ b/influxql/call_iterator.go
@@ -1295,16 +1295,17 @@ func newSampleIterator(input Iterator, opt IteratorOptions, size int) (Iterator,
 
 // newIntegralIterator returns an iterator for operating on a integral() call.
 func newIntegralIterator(input Iterator, opt IteratorOptions, interval Interval) (Iterator, error) {
+	groupByTime := !opt.Interval.IsZero()
 	switch input := input.(type) {
 	case FloatIterator:
 		createFn := func() (FloatPointAggregator, FloatPointEmitter) {
-			fn := NewFloatIntegralReducer(interval)
+			fn := NewFloatIntegralReducer(interval, groupByTime)
 			return fn, fn
 		}
 		return &floatReduceFloatIterator{input: newBufFloatIterator(input), opt: opt, create: createFn}, nil
 	case IntegerIterator:
 		createFn := func() (IntegerPointAggregator, FloatPointEmitter) {
-			fn := NewIntegerIntegralReducer(interval)
+			fn := NewIntegerIntegralReducer(interval, groupByTime)
 			return fn, fn
 		}
 		return &integerReduceFloatIterator{input: newBufIntegerIterator(input), opt: opt, create: createFn}, nil
--- a/influxql/call_iterator.go
+++ b/influxql/call_iterator.go
@@ -1292,3 +1292,23 @@ func newSampleIterator(input Iterator, opt IteratorOptions, size int) (Iterator,
 		return nil, fmt.Errorf("unsupported elapsed iterator type: %T", input)
 	}
 }
+
+// newIntegralIterator returns an iterator for operating on a integral() call.
+func newIntegralIterator(input Iterator, opt IteratorOptions, interval Interval) (Iterator, error) {
+	switch input := input.(type) {
+	case FloatIterator:
+		createFn := func() (FloatPointAggregator, FloatPointEmitter) {
+			fn := NewFloatIntegralReducer(interval)
+			return fn, fn
+		}
+		return &floatReduceFloatIterator{input: newBufFloatIterator(input), opt: opt, create: createFn}, nil
+	case IntegerIterator:
+		createFn := func() (IntegerPointAggregator, FloatPointEmitter) {
+			fn := NewIntegerIntegralReducer(interval)
+			return fn, fn
+		}
+		return &integerReduceFloatIterator{input: newBufIntegerIterator(input), opt: opt, create: createFn}, nil
+	default:
+		return nil, fmt.Errorf("unsupported integral iterator type: %T", input)
+	}
+}
--- a/influxql/functions.go
+++ b/influxql/functions.go
@@ -760,16 +760,17 @@ func (r *FloatHoltWintersReducer) constrain(x []float64) {
 
 // FloatIntegralReducer calculates the time-integral of the aggregated points.
 type FloatIntegralReducer struct {
+	groupByTime bool
 	count     uint32
 	interval  Interval
 	sum       float64
 	prev      FloatPoint
-	startTime int64
 }
 
 // NewFloatIntegralReducer creates a new FloatIntegralReducer.
-func NewFloatIntegralReducer(interval Interval) *FloatIntegralReducer {
+func NewFloatIntegralReducer(interval Interval, groupByTime bool) *FloatIntegralReducer {
 	return &FloatIntegralReducer{
+		groupByTime: groupByTime,
 		interval: interval,
 		prev:     FloatPoint{Nil: true},
 	}
@@ -780,7 +781,6 @@ func (r *FloatIntegralReducer) AggregateFloat(p *FloatPoint) {
 
 	// If this is the first point, just save it
 	if r.prev.Nil {
-		r.startTime = p.Time
 		r.prev = *p
 		r.count++
 		return
@@ -806,26 +806,38 @@ func (r *FloatIntegralReducer) AggregateFloat(p *FloatPoint) {
 }
 
 // Emit emits the time-integral of the aggregated points as a single point.
+// InfluxQL convention dictates that outside a group-by-time clause we return
+// a timestamp of zero.  Within a group-by-time, we can set the time to ZeroTime
+// and a higher level will change it to the start of the time group.
 func (r *FloatIntegralReducer) Emit() []FloatPoint {
-	return []FloatPoint{{
-		Time:       r.startTime,
-		Value:      r.sum,
-		Aggregated: r.count,
-	}}
+	if r.groupByTime {
+		return []FloatPoint{{
+			Time:       ZeroTime,
+			Value:      r.sum,
+			Aggregated: r.count,
+		}}
+	} else {
+		return []FloatPoint{{
+			Time:       0,
+			Value:      r.sum,
+			Aggregated: r.count,
+		}}
+	}
 }
 
 // IntegerIntegralReducer calculates the time-integral of the aggregated points.
 type IntegerIntegralReducer struct {
+	groupByTime bool
 	count     uint32
 	interval  Interval
 	sum       float64
 	prev      IntegerPoint
-	startTime int64
 }
 
 // NewIntegerIntegralReducer creates a new IntegerIntegralReducer.
-func NewIntegerIntegralReducer(interval Interval) *IntegerIntegralReducer {
+func NewIntegerIntegralReducer(interval Interval, groupByTime bool) *IntegerIntegralReducer {
 	return &IntegerIntegralReducer{
+		groupByTime: groupByTime,
 		interval: interval,
 		prev:     IntegerPoint{Nil: true},
 	}
@@ -833,10 +845,10 @@ func NewIntegerIntegralReducer(interval Interval) *IntegerIntegralReducer {
 
 // AggregateInteger aggregates a point into the reducer.
 func (r *IntegerIntegralReducer) AggregateInteger(p *IntegerPoint) {
+
 	// If this is the first point, just save it
 	if r.prev.Nil {
 		r.prev = *p
-		r.startTime = p.Time
 		r.count++
 		return
 	}
@@ -861,10 +873,21 @@ func (r *IntegerIntegralReducer) AggregateInteger(p *IntegerPoint) {
 }
 
 // Emit emits the time-integral of the aggregated points as a single FLOAT point
+// InfluxQL convention dictates that outside a group-by-time clause we return
+// a timestamp of zero.  Within a group-by-time, we can set the time to ZeroTime
+// and a higher level will change it to the start of the time group.
 func (r *IntegerIntegralReducer) Emit() []FloatPoint {
-	return []FloatPoint{{
-		Time:       r.startTime,
-		Value:      r.sum,
-		Aggregated: r.count,
-	}}
+	if r.groupByTime {
+		return []FloatPoint{{
+			Time:       ZeroTime,
+			Value:      r.sum,
+			Aggregated: r.count,
+		}}
+	} else {
+		return []FloatPoint{{
+			Time:       0,
+			Value:      r.sum,
+			Aggregated: r.count,
+		}}
+	}	
 }
--- a/influxql/functions.go
+++ b/influxql/functions.go
@@ -757,3 +757,114 @@ func (r *FloatHoltWintersReducer) constrain(x []float64) {
 		x[3] = 0
 	}
 }
+
+// FloatIntegralReducer calculates the time-integral of the aggregated points.
+type FloatIntegralReducer struct {
+	count     uint32
+	interval  Interval
+	sum       float64
+	prev      FloatPoint
+	startTime int64
+}
+
+// NewFloatIntegralReducer creates a new FloatIntegralReducer.
+func NewFloatIntegralReducer(interval Interval) *FloatIntegralReducer {
+	return &FloatIntegralReducer{
+		interval: interval,
+		prev:     FloatPoint{Nil: true},
+	}
+}
+
+// AggregateFloat aggregates a point into the reducer.
+func (r *FloatIntegralReducer) AggregateFloat(p *FloatPoint) {
+
+	// If this is the first point, just save it
+	if r.prev.Nil {
+		r.startTime = p.Time
+		r.prev = *p
+		r.count++
+		return
+	}
+
+	// If this point has the same timestamp as the previous one,
+	// use it but don't add anything to the integral.  Effectively
+	// this means that we allow the curve we are integrating
+	// to have step changes in it, but who knows whether the points
+	// actually arrive in any particular order...?
+	if r.prev.Time == p.Time {
+		r.prev = *p
+		r.count++
+		return
+	}
+
+	// Normal operation: update the sum using the trapezium rule
+	elapsed := float64(p.Time-r.prev.Time) / float64(r.interval.Duration)
+	r.sum += 0.5 * (p.Value + r.prev.Value) * elapsed
+	r.prev = *p
+	r.count++
+	return
+}
+
+// Emit emits the time-integral of the aggregated points as a single point.
+func (r *FloatIntegralReducer) Emit() []FloatPoint {
+	return []FloatPoint{{
+		Time:       r.startTime,
+		Value:      r.sum,
+		Aggregated: r.count,
+	}}
+}
+
+// IntegerIntegralReducer calculates the time-integral of the aggregated points.
+type IntegerIntegralReducer struct {
+	count     uint32
+	interval  Interval
+	sum       float64
+	prev      IntegerPoint
+	startTime int64
+}
+
+// NewIntegerIntegralReducer creates a new IntegerIntegralReducer.
+func NewIntegerIntegralReducer(interval Interval) *IntegerIntegralReducer {
+	return &IntegerIntegralReducer{
+		interval: interval,
+		prev:     IntegerPoint{Nil: true},
+	}
+}
+
+// AggregateInteger aggregates a point into the reducer.
+func (r *IntegerIntegralReducer) AggregateInteger(p *IntegerPoint) {
+	// If this is the first point, just save it
+	if r.prev.Nil {
+		r.prev = *p
+		r.startTime = p.Time
+		r.count++
+		return
+	}
+
+	// If this point has the same timestamp as the previous one,
+	// use it but don't add anything to the integral.  Effectively
+	// this means that we allow the curve we are integrating
+	// to have step changes in it, but who knows whether the points
+	// actually arrive in any particular order...?
+	if r.prev.Time == p.Time {
+		r.prev = *p
+		r.count++
+		return
+	}
+
+	// Normal operation: update the sum using the trapezium rule
+	elapsed := float64(p.Time-r.prev.Time) / float64(r.interval.Duration)
+	r.sum += 0.5 * float64(p.Value+r.prev.Value) * elapsed
+	r.prev = *p
+	r.count++
+	return
+}
+
+// Emit emits the time-integral of the aggregated points as a single FLOAT point
+func (r *IntegerIntegralReducer) Emit() []FloatPoint {
+	return []FloatPoint{{
+		Time:       r.startTime,
+		Value:      r.sum,
+		Aggregated: r.count,
+	}}
+}
--- a/influxql/functions.go
+++ b/influxql/functions.go
@@ -761,18 +761,18 @@ func (r *FloatHoltWintersReducer) constrain(x []float64) {
 // FloatIntegralReducer calculates the time-integral of the aggregated points.
 type FloatIntegralReducer struct {
 	groupByTime bool
-	count     uint32
-	interval  Interval
-	sum       float64
-	prev      FloatPoint
+	count       uint32
+	interval    Interval
+	sum         float64
+	prev        FloatPoint
 }
 
 // NewFloatIntegralReducer creates a new FloatIntegralReducer.
 func NewFloatIntegralReducer(interval Interval, groupByTime bool) *FloatIntegralReducer {
 	return &FloatIntegralReducer{
 		groupByTime: groupByTime,
-		interval: interval,
-		prev:     FloatPoint{Nil: true},
+		interval:    interval,
+		prev:        FloatPoint{Nil: true},
 	}
 }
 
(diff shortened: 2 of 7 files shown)


Tomcat-Engineering:
#1400 
#5930 

BTW, I propose a separate `cumulative_integral` function that outputs one value for each datapoint rather than just the aggregate (last) value.  I decided that it would be cleaner if this was kept separate from the main `integral` function.  Happy to implement `cumulative_integral` if people would find it useful.

Tomcat-Engineering:
Documentation pull request is at https://github.com/influxdata/docs.influxdata.com/pull/855

jwilder:
@jsternberg @nathanielc

nathanielc:
The "area under the curve" via the trapezoidal rule looks great. I reviewed the implementation of the aggregator and it makes sense to me. The function signature also makes sense to me.

As for the edge cases around the group by time boundaries, @jsternberg do you know if the derivative takes into account the last point of the previous group by interval? I thought I remembered you working on that at some point.

As for the rest of the code of boiler plate QL functions it looked good to me but I'll defer to @jsternberg  on any of that code.

Tomcat-Engineering:
Looks like `derivative` doesn't do anything fancy like looking for the previous point outside the time interval being analysed, it just doesn't emit anything for the first datapoint.

jsternberg:
@nathanielc derivative will try to take into account the previous point, but only when there is a group by interval.

The only concern I have with this is that derivative will eliminate the first point so if there are 10 points, derivative will return 9. I figure that integral, which should be the opposite of derivative, should restore that point. While we wouldn't be able to restore the `+ c` constant for the integral (which isn't a problem to me personally), I don't see how we would restore the first point that was eliminated by derivative.

If this isn't an issue, then I'm fine with this. @nathanielc comments?

nathanielc:
> derivative will try to take into account the previous point, but only when there is a group by interval.

This was my concern, as derivative will correctly take into account group by time boundaries, where the current implementation will skip the trapezoid that spans the group by time boundary. 

Specifically for this test case https://github.com/influxdata/influxdb/pull/7591/files#diff-4efc5fdc06e36abc4addef9f831c2345R2677

I would expect the integral over the entire time range to be the same as the sum of the integrals over each group by time interval. The data for the test case is (in line protocol):

```
m value=20  10
m value=10  15
m value=0   20
m value=-10 30
```

The integral over the entire range is `15*5+5*5+-5*10 = 50`. 
The integral over the 20s group by time intervals (as it is currently implemented) is `15*5 = 75` for the first interval and `-5*10 = -50` for the second interval. The sum of those two interval is `75+-50 = 25`. The `5*5` term is missing and so the result is off of `25` from the integral of the entire interval.

I would expect the behavior to be that the first group by interval value is  `15*5 = 75` and the second is `5*5 + -5*10 = -25`, then the sum of both intervals is `75+-25 = 50` which is the same value we get if we had performed the integral over the entire interval. 

In short just like derivative will carry over the last point from the previous interval if we are grouping by time, I think that integral should do the same. And obviously for the first interval there is nothing to carry over. @Tomcat-Engineering @jsternberg @jwilder Does my expectation make sense? Or would you expect each group by time interval to be independent?

As for the `restore` the first point question, I don't think it is relevant. An integral is a sum of points, so the integral of `n` points always returns a single point.

Tomcat-Engineering:
The `cumulative_integral` function which I proposed in the comments above would be the one which is the inverse of the `derivative` function.  The `integral` function in this PR is an aggregator, so it returns a single point rather than a time series.  

PS - the `+ c` constant doesn't apply in this case because we are dealing with definite integrals.

Tomcat-Engineering:
@nathanielc your expectation makes sense and I'll have a go at implementing it... can anyone point me at the relevant code from `derivative`?  I'm struggling to find the magic bit that handles group-by-time.

bbinet:
Same here, integral function would be very useful.
What is the status of this PR? Is it planned to be merged soon?

Tomcat-Engineering:
The basic problem is that it is not possible to efficiently find the neighbouring points immediately before and after the queried time interval.  These are necessary to give consistent integral results, as discussed above, but you may have to search the whole of time to find them which would be slow.

It may be possible to do clever things to pass points between adjacent group-by-time intervals, but I don't think we would want to do that as the results would be even more inconsistent: a particular time-group could then have **four** different integral values depending on whether it is at the beginning, middle or end of the query interval, or is the only group in the interval.

Having thought about it for a while, I can't come up with anything better than my original algorithm, despite its limitations.  Anyone else got any bright ideas?

jsternberg:
I think you're right. If we find a solution to that in the future, I assume it won't be a problem for us to add back on that one extra point since it will be at a different time than the previous values.

Can you rebase this and resolve the conflicts? Unfortunately, we're in the weird period where 1.2 has been branched off into its own branch, but we're not merging 1.3 bound PRs to master yet. I also don't have time to look at this in full until probably next week. I'm sorry for the inconvenience. If you can also update the changelog to target 1.3, I'll try to get this merged early next week.

Sorry for taking so long to respond to this.

influxql/call_iterator.go
```diff
@@ -1292,3 +1292,24 @@ func newSampleIterator(input Iterator, opt IteratorOptions, size int) (Iterator,
...
+// newIntegralIterator returns an iterator for operating on a integral() call.
+func newIntegralIterator(input Iterator, opt IteratorOptions, interval Interval) (Iterator, error) {
+	groupByTime := !opt.Interval.IsZero()
+	switch input := input.(type) {
+	case FloatIterator:
+		createFn := func() (FloatPointAggregator, FloatPointEmitter) {
+			fn := NewFloatIntegralReducer(interval, groupByTime)
+			return fn, fn
+		}
+		return &floatReduceFloatIterator{input: newBufFloatIterator(input), opt: opt, create: createFn}, nil
```
jsternberg: Does `integral()` have the same behavior as `derivative()`? I think this should be using the stream iterator instead. That way, it can get called on every incoming point rather than only after a group by interval is read.

influxql/select_test.go
```diff
@@ -2645,6 +2645,104 @@ func TestSelect_Elapsed_Boolean(t *testing.T) {
 	}
 }
 
+func CheckPoints(t *testing.T, ic *IteratorCreator, query string, expected [][]influxql.Point) {
```
jsternberg: I don't think this is necessary. I think you can use the same pattern the other tests use. I see that this is used in other parts of the code:

```
    if err != nil {
        t.Fatal(err)
    } else if a, err := Iterators(itrs).ReadAll(); err != nil {
        t.Fatalf("unexpected point: %s", err)
    } else if !deep.Equal(a, [][]influxql.Point{
        {&influxql.FloatPoint{Name: "cpu", Tags: ParseTags("host=A"), Time: 0 * Second, Value: 19, Aggregated: 2}},
        {&influxql.FloatPoint{Name: "cpu", Tags: ParseTags("host=B"), Time: 0 * Second, Value: 10, Aggregated: 1}},
        {&influxql.FloatPoint{Name: "cpu", Tags: ParseTags("host=A"), Time: 10 * Second, Value: 2, Aggregated: 2}},
        {&influxql.FloatPoint{Name: "cpu", Tags: ParseTags("host=A"), Time: 30 * Second, Value: 100, Aggregated: 1}},
    }) {
        t.Fatalf("unexpected points: %s", spew.Sdump(a))
    }
```

influxql/call_iterator.go
Tomcat-Engineering: `integral` returns a single total value aggregating all the incoming points, whereas `derivative` returns a value for each incoming value, hence the different iterators.  

See above discussion of the proposed `cumulative_integral` function for an integral which returns a point for each incoming point.

influxql/select_test.go
Tomcat-Engineering: Yup, I just thought that rather than duplicating the same code loads of times it would be best practice to write it once and wrap it in a function - but I'm happy to remove the function and get copy-pasting if that is the preferred style!

jsternberg:
I'm afraid I might be a bit confused about the direction of this PR and I think some new features (like subqueries) may also factor into that confusion.

Can you explain to me in words which of the following queries are valid and what they are supposed to look like?

```
SELECT integral(value) FROM cpu
SELECT integral(mean(value)) FROM cpu
SELECT integral(value) FROM cpu GROUP BY time(5m)
SELECT integral(mean(value)) FROM cpu GROUP BY time(5m)
```

The queries are abbreviated so just assume they have the proper where clause. If I'm understanding correctly, all of those queries would return one value that would just be the sum of the area under the curve. Is that correct?

I'm also thinking subqueries may influence what we do here. Before 1.2, you would have needed a special function to find the sum of an integral. We had one person who really needed a sum of derivatives functionality for a similar purpose. I'm thinking that we might want to make this function more similar to `derivative()` though so it acts as an exact opposite. That way, the current implementation would be done with:

```
SELECT sum(integral) FROM (SELECT integral(value) FROM cpu)
```

Basically, make the proposed `cumulative_integral()` into the default `integral()` and suggest that people use subqueries for the sum of that curve.

Thoughts?

Tomcat-Engineering:
`SELECT integral(value) FROM cpu`
Join each point in cpu to its neighbours with straight lines.  Return the total area under this shape.

`SELECT integral(mean(value)) FROM cpu`
Doesn't work, both `integral` and `mean` are aggregators so this is rejected by the parser.

`SELECT integral(value) FROM cpu GROUP BY time(5m)`
Within each time-bin, join each point in cpu to its neighbours with straight lines.  Return the total area under this shape for each bin.

`SELECT integral(mean(value)) FROM cpu GROUP BY time(5m)`
In theory this could compute the mean value for each 5-minute bin and then integrate the resulting curve - but the parser doesn't allow similar nested queries like `sum(mean(value))` so for consistency I haven't updated the parser to allow this and the query will be rejected.

Tomcat-Engineering:
Perhaps the confusion is mostly about naming - if you expect `integral` to be the inverse of `derivative` then we should use my proposed `cumulative_integral` as the `integral` function as you suggest.  I think the "integral as aggregator" was the feature that people meant when they originally requested this, but I'm happy to change the name if it is less confusing.

However we define it, `integral` will never be an exact inverse of `derivative` - apart from anything else, `derivative` returns one fewer points than it receives so we can never get back to the original data.

I think the current implementation is still useful enough to include, perhaps with the name `total_integral` or something.  Based on the comments in the original feature request, this function is what people are going to use most of the time and it would be annoying to have to use subqueries all the time for this common operation.

This is exactly analogous to the `sum` function: you could implement this by taking the last value from the `cumulative_sum` results in a subquery, but the aggregator is used more often so gets its own function.  

Likewise, to get the total integral out of the `cumulative_integral` result you take the `last` value.  You can see where I got the idea for the naming from!

jsternberg:
Yea, you'll never be able to reverse the derivative completely and you'll still lose a point, but there will always be limitations and integrating in calculus already has the `+ c` thing because you can't retrieve the constant after it has been removed when finding the derivative.

I think changing this to be the proposed `cumulative_integral()` would give more flexibility though because the current implementation of `integral()` could be implemented with using the subquery that I mentioned in my last comment.

@nathanielc thoughts? Maybe I'm off-base and I don't understand the use cases very well. I'm also not sure which time the integral should output. I would say the time of the second point rather than the first point (like derivative).

Tomcat-Engineering:
`integral` has been listed in [the documentation](https://docs.influxdata.com/influxdb/v1.2/query_language/functions/) as an aggregation for ages, so someone at least was expecting it to return a single point!  But happy to rename it to `total_integral` if you think it is more clear.

dreamon-dd:
For the use cases, if I understand the current implementation correctly, this function will basically resolve the need of time-weighted average, described in #7445, if divided by the time difference between last and first point in each interval. The result is still using interpolation between points, not steps for the value change as described in #7445, but will be already much better than in case of using simple mean() which doesn't consider time. Therefore it will be essential for downsampling time-unregular sensor data.

jsternberg:
I don't think integral will solve any problem of time weighted averages because I don't think integral performs any time weighted average. I think we need to implement the `weighted_moving_average()` function for that.

I'll have a longer post responding to the specifics of integral in a bit. I need some time to organize my thoughts into understandable words.

dreamon-dd:
The value of the integral on the interval divided by the duration of this interval is the average value on this interval, isn't it?

jsternberg:
@Tomcat-Engineering 

I talked with @pauldix a bit today about the `integral()` function and we came back to looking at the function in graphite so we could be certain we implemented what was in graphite. Here's a link to their documentation: http://graphite.readthedocs.io/en/latest/functions.html#graphite.render.functions.integral

I'm a bit confused by the description of how integral works here, so I launched a graphite instance and grafana to try some simple graphs to see what these functions did without having to read the documentation. The descriptions are pretty accurate, but they're also confusing because the definition of `integral()` in the Grafana documentation is not actually what an integral is. It just seems to sum every value that comes in. Now, it doesn't normalize it for time and I don't think you can do that with basic time math in the query so we can talk about a function for doing that, but I'm not sure we should use the name `integral()`. While it might be confusing for those coming from graphite, I think we should be focusing on implementing `integral()` and not focusing on implementing what graphite incorrectly thinks an integral is. I looked around for some confirmation about my thoughts here and found this [blog post](https://grafana.com/blog/2016/03/03/25-graphite-grafana-and-statsd-gotchas/#deriving.integration) from grafana confirming that graphite is just wrong here.

I would also like to know if `cumulative_sum()` matches the current function. From what I see, `cumulative_sum()` should implement the same functionality as graphite's `integral()`, but without the time normalization. Maybe we can add an argument to `cumulative_sum()` for it to normalize by time or think of a function name to do that.

Going back to the difference between aggregate and transformation, I don't think an aggregate is correct for integral. I think we put that down in the documentation before anybody took a serious look at implementing it or deciding what that meant. An aggregate is something that will either be used on the entire span of the series and produce one value (usually without skipping values like derivative sometimes does) or it will perform that same operation on disparate intervals. Aggregates don't consider other intervals when calculating the current interval. I don't think that applies to integral and grafana categorizes both derivative and integral for graphite as transformations. So I think we should be thinking of integral as a transformation rather than an aggregate.

I think an initial question to answer is this: should integral in InfluxDB match integral in Graphite or should we implement an actual integral?

I'm also going to write up a follow-up of how I think integral should be implemented, but I wanted to write this to get some initial thoughts so we could be on the same page.

pauldix:
Thanks for the writeup @jsternberg. Seeing this I think we should implement an actual integral function. Sounds like we already have the equivalent to Graphite's function with cumulative_sum.

jsternberg:
A few more comments on this and how we should interpret the query. I think we need to define what integral should encompass and determine the syntax. While it doesn't fit into an aggregate, it may also not be the same as derivative and I would like to explore that a bit.

The integral is the area under the curve, but we also need to figure out which part of the area under the curve we're referring to. If we have an example where there are 100 points equally spaced apart, is it desirable to find the area between each of these points? That would result in 99 different areas. There's also the question of whether or not we're finding the cumulative sum of these areas. If you were to use an integral to reverse a derivative, the function that would graph that would be the cumulative sum of these areas. So if the original function is `x = t` and the derivative is `v = 1`, then the point at `t = 1` would be the integral from 0 to 1 while the points at `t = 2` would be the integral from 0 to 2.

So the first question. Should the integral be the cumulative sum of the areas or should it just be the area of the current interval? My preference is that it is the area of the current interval (so from `t = 0` to `t = 1` and from `t = 1` to `t = 2` and etc). If you want to get the cumulative sum of the areas, you can just use a subquery to find the cumulative sum from the output of integral.

The next question. If you ask for an integral of a point and specify nothing else, should this produce a single value (the area under the entire curve) or should it produce the area under the curve between each point? The former might be what is expected by users, but the latter is more flexible because you can use the latter with a subquery and `sum()` to find the former, but you can't do the same with the latter.

That also ties into bucketing. If we want to find the area under the curve for every 10 minute interval, we could use a subquery with a `GROUP BY time(...)` to implement that if the area of each section under the curve is returned. Unfortunately, if we have a bucket that is partially in one bucket and partially in another, we lose that information if we attempt to do it through a subquery rather than implementing it directly as part of the function.

As confusing as all of that is, I propose two different syntaxes for consideration.

First syntax:

This has integral work as a pseudo-aggregate. It has some properties of streams and some properties of aggregates. It will likely need its own custom implementation since it doesn't neatly fit into either of them.

```
/* Find the area under the entire curve.
   Internally, this will use the trapezoid rule to calculate the area
   between each point and aggregate all of them together. */
SELECT integral(value, 1s) FROM cpu

/* Find the area for each grouping.
   If the area between two points is partially in one bucket and partially in
   another, linear interpolation is used to divide the two.
*/
SELECT integral(value, 1s) FROM cpu GROUP BY time(1m)

/* Find the integral curve of each grouping by accumulating from previous buckets. */
SELECT cumulative_sum(integral) FROM (SELECT integral(value, 1s) FROM cpu GROUP BY time(1m))

/* It is not possible to find the area between each point with this syntax as a separate bucket. */
```

Second syntax:

```
/* Find the area between each point. If there are 100 points, this will emit 99 values. */
SELECT integral(value, 1s) FROM cpu

/* Find the area for each grouping. This will not be able to use linear interpolation when some area is divided. */
SELECT sum(integral) FROM (SELECT integral(value, 1s) FROM cpu) GROUP BY time(1m)

/* Find the area under the entire curve. */
SELECT cumulative_sum(integral) FROM (SELECT integral(value, 1s) FROM cpu)

/* Find the integral curve of each grouping by accumulating from previous buckets. */
SELECT cumulative_sum(sum) FROM (SELECT sum(integral) FROM (SELECT integral(value, 1s) FROM cpu) GROUP BY time(1m))
```

I'm leaning towards the first. While I think it'll be slightly more complicated to implement because it doesn't have a precedent, I think it will be more clear to users. The second lends itself to using an aggregate within the function better than the first, but I don't like that syntax anyway and I don't consider the syntax necessary now that we have subqueries. The first also has an advantage in that it seems like it matches what people would expect more.

The negative is the first does not make it easy to find the cumulative sum of the integral since it isn't easy to find the dividing line between each of the lines. But you can simulate it easily by just selecting a grouping time like 1s or 10s. Since I'm proposing that it will interpolate the points between the buckets, then you could have points 10 second apart from each other and still use a 1 second grouping since it will interpolate the areas of the buckets.

Thoughts?

Tomcat-Engineering:
Thanks for the notes @jsternberg.

As far as I can tell, with one minor exception your "first syntax" is exactly what I have implemented.

The minor exception is that my algorithm doesn't interpolate across time-bin boundaries in a group-by-time query.  This is because it is really hard (or impossible?) to do this efficiently and I think would require re-writing large parts of the query engine.  This was discussed quite a bit in previous comments, and I thought that we had come to the conclusion that we should release it as-is for now, with suitable notes in the documentation, because lots of users are waiting for this feature.  

PS - your conclusions on Graphite's integral are correct.  You based the current InfluxQ `cumulative_sum` function on [someone trying to implement `integral` by copying Graphite!](https://github.com/influxdata/influxdb/pull/6508)

Tomcat-Engineering:
PS - I think the idea of using any form of `integral` to reverse (invert) a `derivative` is a red herring.  I've attached a sketch showing why - apologies for the slightly wonky drawing!

img_0834

The first graph shows a simple time series with four points.  Each square on the paper is one unit.  The second graph is the result of the existing `derivative` function.  The third graph shows that the output of any attempt to integrate the second graph with linear interpolation is just going to be a load of zeroes.  I'm obviously not suggesting that you will always get zeroes, just that the output is rarely likely to be very meaningful or useful.

To get the correct results, we would need to define `integral` to use a bizarre backwards-sample-and-hold (zero order) interpolation as shown in this version of the second graph, where we interpolate horizontally *backwards* from each point:

img_0835

If you apply the cumulative version of your emit-lots-of points integrator to this interpolation you get back to the first curve (though the constant of integration is only correct because the first point happened to be at t=0).  

However, we will definitely not be popular if we define our general-purpose `integral` function to use this crazy type of interpolation!

jsternberg:
Ok, I'll review it with an assumption of the first syntax then. I would like the interpolation (and I don't think it would be too difficult), but let's get this merged. Now that I understand the syntax better, it'll be easier for me to review this.

influxql/functions.go
```diff
@@ -757,3 +757,137 @@ func (r *FloatHoltWintersReducer) constrain(x []float64) {
...
+		r.count++
+		return
+	}
+
+	// If this point has the same timestamp as the previous one,
+	// use it but don't add anything to the integral.  Effectively
+	// this means that we allow the curve we are integrating
+	// to have step changes in it, but who knows whether the points
+	// actually arrive in any particular order...?
+	if r.prev.Time == p.Time {
```
jsternberg: This is a potential error location. Points aren't guaranteed to be ordered. There are two ways to resolve this. Either have integral request the points be ordered from the storage engine (see `median()` in `influxql/select.go`) or by having this function operate on the entire slice instead (see `median()` again).

influxql/functions.go
```diff
@@ -757,3 +757,137 @@ func (r *FloatHoltWintersReducer) constrain(x []float64) {
...
+	r.count++
+	return
+}
+
+// Emit emits the time-integral of the aggregated points as a single point.
+// InfluxQL convention dictates that outside a group-by-time clause we return
+// a timestamp of zero.  Within a group-by-time, we can set the time to ZeroTime
+// and a higher level will change it to the start of the time group.
+func (r *FloatIntegralReducer) Emit() []FloatPoint {
+	if r.groupByTime {
```
jsternberg: This shouldn't be needed. If you use `ZeroTime`, one of the other iterators takes care of setting the start time correctly. Using `0` would result in the wrong time here.

influxql/iterator.go
```diff
@@ -882,6 +882,16 @@ func (opt IteratorOptions) ElapsedInterval() Interval {
 	return Interval{Duration: time.Nanosecond}
 }
 
+// IntegralInterval returns the time interval for the integral function.
+func (opt IteratorOptions) IntegralInterval() Interval {
```
jsternberg: This looks slightly different from `DerivativeInterval()`. Is there a reason this is different?

influxql/select.go
```diff
@@ -279,7 +279,7 @@ func buildExprIterator(expr Expr, ic IteratorCreator, opt IteratorOptions, selec
 			opt.Interval = Interval{}
 
 			return newHoltWintersIterator(input, opt, int(h.Val), int(m.Val), includeFitData, interval)
-		case "derivative", "non_negative_derivative", "difference", "moving_average", "elapsed":
+		case "derivative", "non_negative_derivative", "difference", "moving_average", "elapsed", "integral":
```
jsternberg: `integral()` seems to be different than the rest of these. The rest of these are here because they aren't aggregates so they don't support `GROUP BY` or `fill()`. If this is going to be an aggregate, it needs to be moved to the area below this.

influxql/select_test.go
jsternberg: Yea, if you can remove this that would be great. I understand the desire and I would consider a PR that changes everything for this pattern, but I don't want to mix test styles at the moment. When I've tried that in the past, it became unruly and I had to undo all of it.

jsternberg:
@Tomcat-Engineering I'm going to use your PR as a base for the functionality I described in the first syntax as we discussed before. I'll also be rebasing this onto a more recent commit and making it compatible with subqueries along with trying to implement the interpolation aspect I talked about.

Just giving you a heads up. Is that ok?

Tomcat-Engineering:
@jsternberg - sounds great, thanks a lot and sorry I didn't get around to doing it myself!

jsternberg:
I've done the fixup and adapted the PR. I'm going to close this one and open up a new PR.

I still want to add the interpolation feature I mentioned above so integral gives more accurate results. I'm hesitant to merge it without that since adding it later will likely count as a breaking change. The difficulty is that, because of that, integral acts differently than both streams and aggregates and it's showing some holes in the method employed to implement both of those. I'll be continuing to work on that, but I wanted to get up what I currently had for the sake of openness.

jsternberg:
Closing this in favor of #8194.

Issue #7439, opened 2016-10-10
What is func (c *Call) Fields() for?


Tomcat-Engineering:
I was adding a new InfluxQL function when I came across [this function](https://github.com/influxdata/influxdb/blob/master/influxql/ast.go#L3217).

I have a few questions about it:

1. What is it used for?  I can't see any obvious in the code which uses it.  If I delete it, everything still compiles.  Is it used for some top-secret part of the commercial clustering code?  Or is it old code which can be removed?

2. The comment says that it only applies to "specific calls".  How do I know which calls it should apply to?  For example, it currently works for `mean` and `mode` but not `median`... is there some subtle reason for this or is it just an oversight?  How do I know whether it should apply to my new function?

Thank you!

jsternberg:
Probably just an old function that's not used anymore. There are a bunch of them that I try to purge whenever I find them. The query engine especially went through a major refactor in 0.11 and a lot of these kinds of functions that it used were not removed since they still compiled perfectly fine.

PR #7592, opened 2016-11-06
Remove old code which is no longer used.

Tomcat-Engineering:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CLA signed

As per #7439 

When I was adding new functions to InfluxQL I came across this and spent a while trying to figure out whether I needed to update it for my functions - it seems to do things for some InfluxQL functions but not others.  

Turns out it is redundant code leftover from a refactoring and is not used anyway.  I propose that we remove it to prevent other people wasting time trying to figure out what it is for.


jwilder:
Thanks @Tomcat-Engineering!

PR #7610, opened 2016-11-09
Prune shards in meta data

### LINKED ISSUE #7601: Prune data in meta store for deleted shards
## Problem Statement

Currently, when a shard is deleted, it is simply marked with a timestamp in the `DeletedAt` field.

For long running instances, this can cause a build up of shard infos in the data store that create unnecessary garbage when being cloned as well as unnecessary cycles during searches of the shards.

## Proposal

- Each data node that successfully drops the shard will remove itself from the owner group of the shard.
-  A periodic service that runs every 30 minutes will `prune` all shards from the meta store that have no owners, or the `DeletedAt` is more than 2 weeks old.

corylanou:
fixes #7601 

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/services/meta/client.go
+++ b/services/meta/client.go
@@ -28,6 +28,10 @@ const (
 	SaltBytes = 32
 
 	metaFile = "meta.db"
+
+	// ShardGroupDeletedExpiration is the amount of time before a shard group info will be removed from cached
+	// data after it has been marked deleted (2 weeks)
+	ShardGroupDeletedExpiration = -2 * 7 * 24 * time.Hour
 )
 
 var (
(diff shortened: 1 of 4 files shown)


services/meta/client.go
```diff
@@ -654,6 +658,33 @@ func (c *Client) DropShard(id uint64) error {
...
+					// we are safe to delete the shard group as it's been marked deleted for the required expiration
+					s := append(rp.ShardGroups[:i], rp.ShardGroups[i+1:]...)
+					data.Databases[i].RetentionPolicies[j].ShardGroups = s
+					changed = true
+				}
+			}
+		}
+	}
+	if changed {
+		if err := c.commit(data); err != nil {
```
jwilder: Can be `return c.commit(data)`

PR #7611, opened 2016-11-09
WIP: Change influx_stress to influx-stress

desa:
This PR removes `influx_stress` and replaces it with [`influx-stress`](https://github.com/influxdata/influx-stress).

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

##### Todo
- [x] Add tests


jwilder:
We should keep `influx-stress` in a separate repo and deprecate `influx_stress`.

desa:
@jwilder when should we remove `influx_stress` from the main repo. Should I open up a PR that does that now or should we just add something to the 1.2 release notes saying that we're going to deprecate it?

jwilder:
I think adding to the release notes that we're deprecating it would be fine.  We don't need to remove it immediately.

PR #7620, opened 2016-11-11
Update Elapsed function to return zero when only one point was aggreg…

### LINKED ISSUE #7619: Update Elapsed function to return 0 for aggregations of less than two points
Feature Request

Currently the elapsed function returns nil if it aggregated less than 2 points. It makes sense to me to have it return a point with value of zero, when it aggregates less than 2 points. Since zero time has elapsed form one point to itself, or from no points.

Use case:

This use case came up here in this mailing list discussion https://groups.google.com/forum/?pli=1#!topic/influxdb/1CdjyXrhbjY

The basic idea is that elapsed time is being calculated between pairs of points. But if a point is missing no value is returned instead of reporting that no time elapsed.

nathanielc:
Fixes #7619 

This makes it so the elapsed function returns zero when only one point was aggregated. If no points were aggregated no points are returned.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - [#7066](https://github.com/influxdata/influxdb/issues/7066): Add support for secure transmission via collectd.
 - [#7554](https://github.com/influxdata/influxdb/pull/7554): update latest dependencies with Godeps.
+- [#7619](https://github.com/influxdata/influxdb/issues/7619): Update Elapsed function to return zero, when only one point was aggregated.
 
 ### Bugfixes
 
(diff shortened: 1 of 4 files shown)


nathanielc:
@jsternberg @jwilder So this isn't as easy as I had hoped. The tests are failing now since the elapsed node emits a `0` point at the beginning of each group by time interval. This is because `elapsed` behaves like derivative and `emit` is called after each point, so we always get a `0` point at the beginning. Is there any way to distinguish between a "normal" emit call and the "last" emit call?

jsternberg:
I'm going to close this. I think it's too late to change this behavior.

PR #7631, opened 2016-11-16
Allocation/pointer removal

jwilder:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

This PR has a few changes to remove pointers and allocations.

* There are several long running goroutines that would `time.Sleep` every second.  This ends up allocating as seen by `GODEBUG=allocfreetrace=1`.  These were replaced with `time.Timer` which avoids the allocation and is more idiomatic.
* `findGenerations` is called a lot and ends up allocation quite a bit when many shards exists.  The result is usually the same because the generations do not change until a compaction completes.  This caches the result to avoid all the re-allocations.
* Converts all the `*Value` types to be non-pointers.   They don't need to be pointers since they are immutable once created.

### DIFF
--- a/tsdb/engine/tsm1/compact.go
+++ b/tsdb/engine/tsm1/compact.go
@@ -73,6 +73,13 @@ type DefaultPlanner struct {
 
 	// lastPlanCheck is the last time Plan was called
 	lastPlanCheck time.Time
+
+	mu sync.RWMutex
+	// lastFindGenerations is the last time findGenerations was run
+	lastFindGenerations time.Time
+
+	// lastGenerations is the last set of generations found by findGenerations
+	lastGenerations tsmGenerations
 }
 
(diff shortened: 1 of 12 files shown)


tsdb/engine/tsm1/compact.go
```diff
@@ -73,6 +73,13 @@ type DefaultPlanner struct {
 
 	// lastPlanCheck is the last time Plan was called
 	lastPlanCheck time.Time
+
+	mu sync.RWMutex
+	// lastFindGenerations is the last time findGenerations was run
+	lastFindGenerations time.Time
+
```
e-dard: nit: since `lastGenerations` is protected by `mu`, the blank line could be removed so it's in the same block as `lastFindGenerations`.

PR #7633, opened 2016-11-16
Improve TSM1 cache performance

e-dard:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

This PR optimises the TSM1 cache, improving write performance under load by up to 70% from `1.1`.

It works by using a simple (crude, really) Hash Ring, which allows any series key to be consistently mapped to a partition in the ring. Each partition contains a map of series keys to TSM1 entries, and manages safe access to that map. Entries are distributed uniformly between a small (currently fixed to 128) partitions.

Since each partition manages access to the entries that map to it, contention over the entire cache is minimised, and restricted to cases when keys that map to the same partition as each other need to operate on the cache.

Further, currently the cache is essentially re-allocated and re-build after a snapshot is taken, which is expensive. This PR adds some logic that keeps track of how entries within the cache are currently allocated as _hints_, and uses these hints to pre-allocate the cache after a snapshot, so that subsequent additions of entries do not result in new allocations within the cache data-structure.

##### Performance

I've been mainly testing performance on a `c4.8xlarge` AWS instance, firing a combined `1.8 M writes/sec` at the instance using `influx-stress`. Specifically:

```
# host 1
$ influx-stress insert -n 30000000 --pps 900000 --host http://influxdb:8086

# host 2
$ influx-stress insert -n 30000000 --pps 900000 --host http://influxdb:8086
```
In total 600M points are written.

 - Write Throughput InfluxDB 1.1 RC1: *1,082,621 writes/sec*
 - Write Throughput er-cache: *1,682,816 writes/sec*

This is a typical run, where the points have a single value. The case where there are multiple points have a similar difference.

### DIFF
--- a/tsdb/engine/tsm1/ring.go
+++ b/tsdb/engine/tsm1/ring.go
@@ -137,9 +137,10 @@ func (r *ring) keys(sorted bool) []string {
 }
 
 // apply applies the provided function to every entry in the ring under a read
-// lock. The provided function will be called with each key and the
-// corresponding entry. The first error encountered will be returned, if any.
-// apply is safe for use by multiple goroutines.
+// lock using a separate goroutine for each partition. The provided function
+// will be called with each key and the corresponding entry. The first error
+// encountered will be returned, if any. apply is safe for use by multiple
+// goroutines.
 func (r *ring) apply(f func(string, *entry) error) error {
 
 	var (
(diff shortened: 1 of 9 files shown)


tsdb/engine/tsm1/ring.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+// newring returns a new ring initialised with n partitions. n must always be a
+// power of 2, and for performance reasons should be larger than the number of
+// cores on the host. The supported set of values for n is:
+//
+//     {1, 2, 4, 8, 16, 32, 64, 128, 256}.
+//
+func newring(n int) (*ring, error) {
+	if n <= 0 || n > 256 {
```
jwilder: Move `256` to a `const` since it's repeated a few times in this func?

jwilder: Also, we should probably test w/ higher partitions under high load to see if that has any benefit.

tsdb/engine/tsm1/ring.go
```diff
@@ -0,0 +1,289 @@
...
+
+	for _, p := range r.partitions {
+		wg.Add(1)
+
+		go func(p *partition) {
+			defer wg.Done()
+
+			p.mu.RLock()
+			for k, e := range p.store {
+				if err := f(k, e); err != nil {
```
jwilder: When I got a lot of timeouts under high load, this was one of the spots that showed up where many goroutines where busy.  The issue was a that each `entry` was being sorted which was expensive and there were millions of them which I think caused CPU starvation for the waiting write HTTP goroutines.

sebito91: This is huge, thank you very much for adding this!

PR #7637, opened 2016-11-17
Return the time from a percentile call on an integer

### LINKED ISSUE #7634: Inconsistency whether percentile iterators should return time
### Bug report

__System info:__ InfluxDB 1.1

__Additional info:__ 

Unsure whether this is actually a bug or not, however I noticed the inconsistency when I was trying to see how percentile() handles null or missing data.  I thought it better to ask whether an issue than ignore it.  Forgive for being unable to develop a process to confirm if an actual bug or not.

1. [float percentile iterator returning time](https://github.com/influxdata/influxdb/blob/master/influxql/call_iterator.go#L1108)
```
/ NewFloatPercentileReduceSliceFunc returns the percentile value within a window.
func NewFloatPercentileReduceSliceFunc(percentile float64) FloatReduceSliceFunc {
	return func(a []FloatPoint) []FloatPoint {
		length := len(a)
		i := int(math.Floor(float64(length)*percentile/100.0+0.5)) - 1

		if i < 0 || i >= length {
			return nil
		}

		sort.Sort(floatPointsByValue(a))
		return []FloatPoint{{Time: a[i].Time, Value: a[i].Value, Aux: a[i].Aux}}
	}
}
```
2. [integer percentile iterator returning ZeroTime](https://github.com/influxdata/influxdb/blob/master/influxql/call_iterator.go#L1123)
```
// NewIntegerPercentileReduceSliceFunc returns the percentile value within a window.
func NewIntegerPercentileReduceSliceFunc(percentile float64) IntegerReduceSliceFunc {
	return func(a []IntegerPoint) []IntegerPoint {
		length := len(a)
		i := int(math.Floor(float64(length)*percentile/100.0+0.5)) - 1

		if i < 0 || i >= length {
			return nil
		}

		sort.Sort(integerPointsByValue(a))
		return []IntegerPoint{{Time: ZeroTime, Value: a[i].Value, Aux: a[i].Aux}}
	}
}
```

jsternberg:
`percentile()` is supposed to be a selector and return the time of the
point, but that only got changed when the input was a float. Updating
the integer processor to also return the time of the point rather than
the beginning of the interval.

Fixes #7634.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 ### Bugfixes
 
 - [#7625](https://github.com/influxdata/influxdb/issues/7625): Fix incorrect tag value in error message.
+- [#7634](https://github.com/influxdata/influxdb/issues/7634): Return the time from a percentile call on an integer.
 
 ## v1.1.0 [2016-11-14]
 ## v1.1.0 [unreleased]
--- a/cmd/influxd/run/server_test.go
+++ b/cmd/influxd/run/server_test.go
@@ -4163,6 +4163,84 @@ func TestServer_Query_AggregateSelectors(t *testing.T) {
 	}
 }
 
+func TestServer_Query_Selectors(t *testing.T) {
+	t.Parallel()
+	s := OpenServer(NewConfig())
+	defer s.Close()
+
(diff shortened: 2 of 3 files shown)


jwilder:
Can you add a test for this?

jsternberg:
Added a test for this.

PR #7644, opened 2016-11-18
Quote the empty string as an ident

jsternberg:
Without this quoting, the function `max("")` turns into `max()` and will
not be reparsed correctly.


jwilder:
```
--- FAIL: TestQuoteIdent (0.00s)
	parser_test.go:2820: 0. []: mismatch:  != ""
```

Issue #4098, opened 2015-09-15
Enable `golint` on the code base


corylanou:
# Description

We want to enable [golint](https://github.com/golang/lint) on our codebase.

First, get golint:

```
go get -u github.com/golang/lint
```

Then run it on any package you want to lint:

``` sh
# will lint entire project
golint ./...

# will lint single package in project
golint ./meta
```

Also, you can use this script to get an output of how many lint errors are in each package, and which packages are currently passing (thanks @Wind0r):

``` sh
#!/bin/bash

#All Packages
PACKAGES=$(find . -name '*.go' -print0 | xargs -0 -n1 dirname | sort --unique)

echo 'packages needing linting'
for pkg in ${PACKAGES[@]}; do
    echo $pkg $(golint $pkg | wc -l) | grep -v '\s0$'
done

echo ''
echo 'packages passing lint'
for pkg in ${PACKAGES[@]}; do
    echo $pkg $(golint $pkg | wc -l) | grep '\s0$'
done
```

We want to do this for several reasons:
- We want to improve code quality
- We need objective filters on quality to help us discriminate bad pull requests
# How to

Following is the list of all `github.com/influxdb/influxdb` subpackages. For each we need to:
- Submit a PR cleaning all existing golint warnings. One thing to keep in mind is that golint will require that every exported symbol has a comment: favor making symbols private where possible before documenting.
- Enable golint on this package as part of the Circle CI jobs to make sure we don't diverge (something that individual contributors can do, but @otoolep who owns the testing infrastructure will help with that).

We also need to make it easy for contributors to easily give golint a local run before submitting the PR. We should probably have that script only cover a hardcoded list of packages that we can augment over time.
# Packages:
- [x] [root] #4340 thanks @skanjo
- [x] client #4439 thanks @skanjo
- [x] cluster #4266 thanks @lenko-d 
- [ ] cluster/internal
- [X] cmd/influx #4703 thanks @pablolmiranda
- [x] cmd/influx_stress #5822 thanks @Gouthamve
- [x] cmd/influxd #5822 thanks @Gouthamve
- [x] cmd/influxd/backup #5822 thanks @Gouthamve
- [x] cmd/influxd/help #5822 thanks @Gouthamve
- [x] cmd/influxd/restore #5822 thanks @Gouthamve
- [x] cmd/influxd/run #5822 thanks @Gouthamve
- [ ] cmd/inspect
- [ ] importer/v8
- [ ] influxql
- [ ] meta
[...]

PR #7645, opened 2016-11-19
removed parentheses to make tcp golintable

shaybix:
Issue: Enable golint on the code base #4098

All required was removing parantheses trailing the method name at the start of the comment.


e-dard:
Thanks @shaybix!

PR #7651, opened 2016-11-22
Expose Shard.LastModified

jwilder:
- [x] Rebased/mergable
- [x] Tests pass

This returns the LastModified time of the shard.  The LastModified
time is the wall time when a change to the shards state occurred.
It uses the WAL or FileStore to determine the max mod time.

### DIFF
--- a/tsdb/engine/tsm1/engine_test.go
+++ b/tsdb/engine/tsm1/engine_test.go
@@ -609,6 +609,63 @@ func TestEngine_DeleteSeries(t *testing.T) {
 
 }
 
+func TestEngine_LastModified(t *testing.T) {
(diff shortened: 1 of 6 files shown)


tsdb/engine/tsm1/engine_test.go
```diff
@@ -609,6 +609,65 @@ func TestEngine_DeleteSeries(t *testing.T) {
 
 }
 
+func TestEngine_LastModified(t *testing.T) {
+	// Generate temporary file.
+	f, _ := ioutil.TempFile("", "tsm")
```
joelegasse: Are you using the generated filename as a directory? Why not just use `iotuil.TempDir`?

PR #7652, opened 2016-11-23
Make monitor golintable

shaybix:
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

contributes to issue #4098 

Added a comment to the Reporter interface
Fixed a capitalisation typo which presented a linting error
Added comment to type Statistics.

### DIFF
--- a/monitor/reporter.go
+++ b/monitor/reporter.go
@@ -2,6 +2,7 @@ package monitor
 
 import "github.com/influxdata/influxdb/models"
 
+// Reporter ...
 type Reporter interface {
 	Statistics(tags map[string]string) []models.Statistic
 }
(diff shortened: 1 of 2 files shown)


monitor/reporter.go
```diff
@@ -2,6 +2,7 @@ package monitor
 
 import "github.com/influxdata/influxdb/models"
 
+// Reporter ...
```
e-dard: This needs a proper comment. Perhaps something like: `Reporter gathers internal statistics.`

monitor/service.go
```diff
@@ -451,6 +451,7 @@ func (s *Statistic) ValueNames() []string {
 	return a
 }
 
+// Statistics ....
```
e-dard: This needs a proper comment. Perhaps something like `Statistics is a collection of sortable statistics.`

shaybix:
I can't make sense of why the tests would have failed in CircleCI, I ran them locally and were just fine? Only changes i made were they ones requests in the documentation?

PR #7658, opened 2016-11-28
Use slash separator for filenames in tar archives

### LINKED ISSUE #7656: Import of backup generated on Windows does not work on Linux
### Bug report

I have an installation of inluxdb running on Windows 10. I made a backup of the metadata and my database and tried to restore it on my linux machine. 

The retstore finished without errors, but as you can see in the following snippet, the path seperator is wrong and as a result, only a file named "stationdata\autogen\2\000000002-000000001.tsm" is generated.

`sudo influxd restore -database stationdata -datadir /var/lib/influxdb/data stationdata.bak
Restoring from backup stationdata.bak/stationdata.*
unpacking /var/lib/influxdb/data/stationdata\autogen\2\000000001-000000001.tsm
unpacking /var/lib/influxdb/data/stationdata\autogen\2\000000001-000000001.tsm
unpacking /var/lib/influxdb/data/stationdata\autogen\2\000000002-000000001.tsm
`

Manually renaming the file and adding the folders works.

__System info:__ 
Windows 10 64bit
Arch Linux 64bit
Influx server and client: 1.1.0

__Steps to reproduce:__

1. Make Backup on Windows
2. Restore Backup on Linux

__Expected behavior:__ Database should be imported up correctly.

__Actual behavior:__ Database is not imported, because the path seperator is wrong, i.e. "\\" instead of "/"

allenpetersen:
- [x] CHANGELOG.md updated
- [x] Rebased/mergable
- [x] Tests pass

NO-OP on platforms with Unix path separator.
On Windows paths get converted to slashes before adding to archive and back to backslashes during restore.

Fixes #7656

### DIFF
--- a/cmd/influxd/restore/restore.go
+++ b/cmd/influxd/restore/restore.go
@@ -309,7 +309,8 @@ func (cmd *Command) unpackTar(tarFile string) error {
 
 // unpackFile will copy the current file from the tar archive to the data dir
 func (cmd *Command) unpackFile(tr *tar.Reader, fileName string) error {
-	fn := filepath.Join(cmd.datadir, fileName)
+	nativeFileName := filepath.FromSlash(fileName)
+	fn := filepath.Join(cmd.datadir, nativeFileName)
 	fmt.Printf("unpacking %s\n", fn)
 
 	if err := os.MkdirAll(filepath.Dir(fn), 0777); err != nil {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 - [#7621](https://github.com/influxdata/influxdb/issues/7621): Expand string and boolean fields when using a wildcard with `sample()`.
 - [#7616](https://github.com/influxdata/influxdb/pull/7616): Fix chuid argument order in init script @ccasey
+- [#7656](https://github.com/influxdata/influxdb/issues/7656): Fix cross-platform backup/restore
 
 ## v1.1.1 [unreleased]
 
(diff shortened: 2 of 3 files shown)


allenpetersen:
I was looking for a simple issue to start with to get familiar with the codebase.

This was easy to reproduce and manually test the fix. I still have a few questions but I thought I would submit the pull request to get feedback.

It looks like the backup happens via Engine.Restore() but the CLI directly calls influxd/restore.go unpackFile() rather than RestorShard/Restore. So I put the filepath.FromSlash() code in both places.

cmd/influxd/restore/restore.go
```diff
@@ -309,7 +309,8 @@ func (cmd *Command) unpackTar(tarFile string) error {
 
 // unpackFile will copy the current file from the tar archive to the data dir
 func (cmd *Command) unpackFile(tr *tar.Reader, fileName string) error {
-	fn := filepath.Join(cmd.datadir, fileName)
+	nativeFileName := filepath.FromSlash(fileName)
```
jsternberg: If I'm reading the documentation for `filepath.FromSlash` correctly, this will convert `/` characters and leave `\` alone on Windows so this is backwards compatible, right?

allenpetersen: Yes, this is backwards compatible since FromSlash called with a filename with backslashes will not find any slashes to convert. To be sure I wrote a little test program to make sure I understood the docs correctly.

On Windows
```
Input:     c:\foo\bar
ToSlash:   c:/foo/bar
FromSlash: c:\foo\bar

Input:     /foo/bar
ToSlash:   /foo/bar
FromSlash: \foo\bar
```

On Linux
```
Input:     c:\foo\bar
ToSlash:   c:\foo\bar
FromSlash: c:\foo\bar

Input:     /foo/bar
ToSlash:   /foo/bar
FromSlash: /foo/bar
```

joelegasse:
LGTM

PR #7662, opened 2016-11-29
Use X-Forwarded-For IP address in HTTP logger if present

### LINKED ISSUE #7356: Support XFF logging for HTTP requests
It is common practice for Load Balancers and Proxies to substitute the source IP address of a packet with their own and (optionally) provide the originating IP address under the X-Forwarded-For header.

The influxd HTTP listener should be able to parse and log the address in the [X-Forwarded-For](https://en.wikipedia.org/wiki/X-Forwarded-For) header, if it exists.

The logs could include both the source and the originating address, e.g:

`[httpd] 172.30.22.4 - XFF 77.49.88.96 - [24/Sep/2016:16:17:33 +0300] "POST /write?consistency=any&db=telegraf&precision=ns&rp= HTTP/1.1" 204 0 "-" "InfluxDBClient" 411c6f60-8259-11e6-8029-000000000000 2680`

jsternberg:
Fixes #7356.

### DIFF
--- a/services/httpd/response_logger.go
+++ b/services/httpd/response_logger.go
@@ -5,6 +5,7 @@ import (
 	"net"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/influxdata/influxdb/influxql"
@@ -84,9 +85,13 @@ func buildLogLine(l *responseLogger, r *http.Request, start time.Time) string {
 	username := parseUsername(r)
 
 	host, _, err := net.SplitHostPort(r.RemoteAddr)
-
 	if err != nil {
 		host = r.RemoteAddr
+	}
+
+	if xff := r.Header["X-Forwarded-For"]; xff != nil {
+		addrs := append(xff, host)
+		host = strings.Join(addrs, ",")
 	}
 
 	uri := r.URL.RequestURI()
--- a/services/httpd/handler_test.go
+++ b/services/httpd/handler_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"mime/multipart"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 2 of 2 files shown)


phemmer:
The change as is is insecure. If someone has direct access to the server (ability to bypass proxy), they could forge the x-forwarded-for header and then perform whatever they wanted and the server would log the wrong address.
The way this is handled on most other apps that use x-forwarded-for is to either have a whitelist of trusted sources which are allowed to provided x-forwarded-for, or to just append x-forwarded-for to the actual address so both are logged.

jsternberg:
@phemmer is there a standard log format that I can use for the latter? I don't think a whitelist is likely practical unless it just included `127.0.0.1` and that would also preclude something like an ELB from being used. I'm not very familiar with the CLF format used by web servers.

To be fair, I don't think the current method is that secure to begin with. I don't remember completely, but I think the remote address can also be forged. It's not as easy, but I thought I remembered that as possible. But given that method of attack is much harder, it might just be enough to write both IP addresses and call it a day.

phemmer:
Since the x-forwarded-for header is already a chain of addresses, I've usually seen it logged as if the socket IP were just added to the chain.

E.G.:

    X-Forwarded-For: 1.1.1.1, 2.2.2.2, 3.3.3.3

Your proxy's address is `4.4.4.4`:
Log:

    1.1.1.1,2.2.2.2,3.3.3.3,4.4.4.4 - johndoe [02/Jan/2006:15:04:05 -0700] ...

phemmer:
> I think the remote address can also be forged

Unless you can hijack the box's gateway (or some other box along the IP route), it's not possible to forge the socket's remote address.

jsternberg:
Updated so it would include all IP addresses in `X-Forwarded-For` and the remote IP address comma separated in the log.

phemmer:
The header might have comma+space delimited values (`X-Forwarded-For: 1.1.1.1, 2.2.2.2`), and the code would just copy that straight into the log. This will probably be rather uncommon though. So the current PR seems acceptable.

services/httpd/response_logger.go
```diff
@@ -84,11 +85,19 @@ func buildLogLine(l *responseLogger, r *http.Request, start time.Time) string {
 	username := parseUsername(r)
 
 	host, _, err := net.SplitHostPort(r.RemoteAddr)
-
 	if err != nil {
 		host = r.RemoteAddr
 	}
 
+	if xff := r.Header["X-Forwarded-For"]; xff != nil {
+		addrs := make([]string, 0, len(xff)+1)
```
e-dard: Instead of:

```go
addrs := make([]string, 0, len(xff)+1)
for _, addr := range xff {
	addrs = append(addrs, addr)
}
addrs = append(addrs, host)
```

just do: 

```go
addrs := append(xff, host)
```

jsternberg:
I just merged and realized I forgot the changelog. I'll have another PR up for the changelog for this.

PR #7671, opened 2016-11-30
Switch logging to use structured logging everywhere

### LINKED ISSUE #7036: Refactor logging to use a structured logger and log levels
### Feature Request

Refactor our logging to use a structured logger and log levels. Structured logging allows us to use a consistent method of logging messages and a way of filtering messages for support. It also allows us to format messages for either the traditional log file that we normally use and like along with JSON messages for consumption by automated log tools like Splunk.

**Proposal:** Refactor logging to a structured logger and log levels

**Current behavior:** We use the Go standard `log` package

**Desired behavior:** Using the structured logger

**Use case:** It makes log files easier to follow and filter out messages that may only be a concern to those doing development.

jsternberg:
The logging library has been switched to use uber-go/zap. While the
logging has been changed to use structured logging, this commit does not
change any of the logging statements to take advantage of the new
structured log or new log levels. Those changes will come in future
commits.

Fixes #7036.


tsdb/shard.go
```diff
@@ -150,24 +150,22 @@ func NewShard(id uint64, index *DatabaseIndex, path string, walPath string, opti
...
 	}
 	return s
 }
 
-// SetLogOutput sets the writer to which log output will be written. It is safe
-// for concurrent use.
-func (s *Shard) SetLogOutput(w io.Writer) {
-	s.logger.SetOutput(w)
+func (s *Shard) WithLogger(log zap.Logger) {
+	s.baseLogger = log.With(
```
joelegasse: This seems like only half of the change to switch to a full structured log. Without changing the current message being printed later on, it results in long messages with duplicated information:
```
[I] 2016-12-14T16:25:12Z /home/jlegasse/.influxdb/data/_internal/monitor/9 database index loaded in 8.860841ms shard=9 path=/home/jlegasse/.influxdb/data/_internal/monitor/9 service=shard
```

I think for this change, the extra context should be left out. We can revisit later and switch from formatted strings to zap context values all at once, then.

jsternberg: Removed this, but kept the scaffolding code so it's easy to readd.

PR #7672, opened 2016-11-30
Enforce minimum shard duration when creating retention policies

### LINKED ISSUE #7563: RP should not allow `INF` or `0` as a shard duration
### Bug report

InfluxDB 1.0.x

CREATE RP and ALTER RP queries will accept `INF` as a valid duration, which should not be accepted, as shards must have an explicit end time. What ends up happening is the shards have a `1s` duration, effectively meaning all data is dropped as soon as it is written.

__Desired behavior:__

Throw an error and reject the RP query.

corylanou:
Fixes #7563 

Here are examples of what the new results are when running `CREATE RETENTION POLICY` commands:

```sql
> create database foo
> CREATE RETENTION POLICY policy3 ON foo DURATION INF REPLICATION 1 SHARD DURATION 0m
> CREATE RETENTION POLICY policy4 ON foo DURATION INF REPLICATION 1 SHARD DURATION 1s
> show retention policies on foo
name    duration        shardGroupDuration      replicaN        default
----    --------        ------------------      --------        -------
autogen 0s              168h0m0s                1               true
policy3 0s              168h0m0s                1               false
policy4 0s              1h0m0s                  1               false

> CREATE RETENTION POLICY policy4 ON foo DURATION INF REPLICATION 1 SHARD DURATION INF
ERR: error parsing query: invalid duration INF for shard duration at line 1, char 82
Warning: It is possible this error is due to not setting a database.
Please set a database with the command "use <database>".
```

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

- [x] Provide example syntax

### DIFF
--- a/influxql/parser_test.go
+++ b/influxql/parser_test.go
@@ -2059,6 +2059,26 @@ func TestParser_ParseStatement(t *testing.T) {
 				ShardGroupDuration: 30 * time.Minute,
 			},
 		},
+		{
+			s: `CREATE RETENTION POLICY policy1 ON testdb DURATION 1h REPLICATION 2 SHARD DURATION 0s`,
+			stmt: &influxql.CreateRetentionPolicyStatement{
+				Name:               "policy1",
+				Database:           "testdb",
+				Duration:           time.Hour,
+				Replication:        2,
+				ShardGroupDuration: 0,
+			},
+		},
+		{
+			s: `CREATE RETENTION POLICY policy1 ON testdb DURATION 1h REPLICATION 2 SHARD DURATION 1s`,
+			stmt: &influxql.CreateRetentionPolicyStatement{
+				Name:               "policy1",
+				Database:           "testdb",
+				Duration:           time.Hour,
+				Replication:        2,
+				ShardGroupDuration: time.Second,
+			},
+		},
 
 		// ALTER RETENTION POLICY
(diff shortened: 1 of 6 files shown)


services/meta/client_test.go
```diff
@@ -239,8 +239,8 @@ func TestMetaClient_CreateRetentionPolicy(t *testing.T) {
 	rp0 := meta.RetentionPolicyInfo{
 		Name:               "rp0",
 		ReplicaN:           1,
-		Duration:           time.Hour,
-		ShardGroupDuration: time.Hour,
+		Duration:           2 * time.Hour,
```
corylanou: This change was required as tests further down rely on changing the RP duration and it was dropping below the new min enforced floor.  This was the easiest way to fix that test other than rewriting the entire test.

influxql/parser_test.go
```diff
@@ -2059,6 +2059,36 @@ func TestParser_ParseStatement(t *testing.T) {
 				ShardGroupDuration: 30 * time.Minute,
 			},
 		},
+		{
```
corylanou: Added these tests just for coverage.  This is just validating the parser allows for these values, and parses the durations correctly.  The integration tests then make sure the minimums are being enforced.

PR #7673, opened 2016-12-01
Reject invalid subscription urls

### LINKED ISSUE #7615: Validate Subscription properties early
# Bug report

InfluxDB: 1.0.2

1.  Create a subscription with a URL scheme of `bad://`
2.  Try and create a new valid subscription for a database with a name alphabetically after the first subscription

## Expected behavior: 
It should throw an error at step one

## Actual behavior:

No error is thrown and in step 2 the subscription will be created but will fail to start because of the first subscription

## Suggested solution

There are two issues: 

1. Validate malformed subscriptions at query time and return an error to the user.
2. Do not fail early when updating subscriptions in the subscriber because there is no way to return an error to the user at that level. Do a best effort update.

allenpetersen:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

The url must have a scheme of udp or tcp in addition to a port number.
CREATE SUBSCRIPTION will fail if there are invalid destinations.

Additionally Service.createSubscription will skip over any invalid
destinations on previously created subscriptions.

Fixes #7615

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,13 +20,14 @@ The stress tool `influx_stress` will be removed in a subsequent release. We reco
 
 - [#7621](https://github.com/influxdata/influxdb/issues/7621): Expand string and boolean fields when using a wildcard with `sample()`.
 - [#7616](https://github.com/influxdata/influxdb/pull/7616): Fix chuid argument order in init script @ccasey
-- [#7656](https://github.com/influxdata/influxdb/issues/7656): Fix cross-platform backup/restore
+- [#7656](https://github.com/influxdata/influxdb/issues/7656): Fix cross-platform backup/restore @allenpetersen
 - [#7650](https://github.com/influxdata/influxdb/issues/7650): Ensures that all user privileges associated with a database are removed when the database is dropped.
 - [#7659](https://github.com/influxdata/influxdb/issues/7659): Fix CLI import bug when using self-signed SSL certificates.
 - [#7698](https://github.com/influxdata/influxdb/pull/7698): CLI was caching db/rp for insert into statements.
 - [#6527](https://github.com/influxdata/influxdb/issues/6527): 0.12.2 Influx CLI client PRECISION returns "Unknown precision....
 
 - [#7396](https://github.com/influxdata/influxdb/issues/7396): CLI should use spaces for alignment, not tabs.
+- [#7615](https://github.com/influxdata/influxdb/issues/7615): Reject invalid subscription urls @allenpetersen
 
 ## v1.1.1 [unreleased]
 
--- a/services/meta/data.go
+++ b/services/meta/data.go
@@ -3,6 +3,8 @@ package meta
 import (
 	"errors"
 	"fmt"
+	"net"
+	"net/url"
 	"sort"
 	"strings"
 	"sync"
@@ -470,8 +472,33 @@ func (data *Data) DropContinuousQuery(database, name string) error {
 	return ErrContinuousQueryNotFound
 }
 
+// validateURL returns an error if the URL does not have a port or uses a scheme other than UDP or TCP.
+func validateURL(input string) error {
+	u, err := url.Parse(input)
+	if err != nil {
+		return ErrInvalidSubscriptionURL(input)
+	}
+
+	if u.Scheme != "udp" && u.Scheme != "http" && u.Scheme != "https" {
+		return ErrInvalidSubscriptionURL(input)
+	}
+
+	_, port, err := net.SplitHostPort(u.Host)
+	if err != nil || port == "" {
+		return ErrInvalidSubscriptionURL(input)
+	}
+
+	return nil
+}
+
 // CreateSubscription adds a named subscription to a database and retention policy.
 func (data *Data) CreateSubscription(database, rp, name, mode string, destinations []string) error {
+	for _, d := range destinations {
+		if err := validateURL(d); err != nil {
+			return err
+		}
+	}
+
 	rpi, err := data.RetentionPolicy(database, rp)
 	if err != nil {
 		return err
--- a/services/subscriber/service.go
+++ b/services/subscriber/service.go
@@ -192,11 +192,11 @@ func (s *Service) createSubscription(se subEntry, mode string, destinations []st
 	for _, dest := range destinations {
 		u, err := url.Parse(dest)
 		if err != nil {
-			return nil, fmt.Errorf("Failed to parse subscription url: %s", dest)
+			return nil, fmt.Errorf("failed to parse destination: %s", dest)
 		}
 		w, err := s.NewPointsWriter(*u)
 		if err != nil {
-			return nil, fmt.Errorf("Failed to create PointsWriter for subscription url: %s", dest)
+			return nil, fmt.Errorf("failed to create writer for destination: %s", dest)
 		}
 		writers = append(writers, w)
 		stats = append(stats, writerStats{dest: dest})
@@ -288,7 +288,7 @@ func (s *Service) updateSubs(wg *sync.WaitGroup) {
 				sub, err := s.createSubscription(se, si.Mode, si.Destinations)
 				if err != nil {
 					atomic.AddInt64(&s.stats.CreateFailures, 1)
-					s.Logger.Println(err)
+					s.Logger.Printf("Subscription creation failed for '%s' with error: %s", si.Name, err)
 					continue
 				}
 				cw := chanWriter{
--- a/services/subscriber/service.go
+++ b/services/subscriber/service.go
@@ -19,8 +19,9 @@ import (
 
 // Statistics for the Subscriber service.
 const (
-	statPointsWritten = "pointsWritten"
-	statWriteFailures = "writeFailures"
+	statCreateFailures = "createFailures"
+	statPointsWritten  = "pointsWritten"
+	statWriteFailures  = "writeFailures"
 )
 
 // PointsWriter is an interface for writing points to a subscription destination.
@@ -123,8 +124,9 @@ func (s *Service) SetLogOutput(w io.Writer) {
 
 // Statistics maintains the statistics for the subscriber service.
 type Statistics struct {
-	WriteFailures int64
-	PointsWritten int64
+	CreateFailures int64
+	PointsWritten  int64
+	WriteFailures  int64
 }
 
 // Statistics returns statistics for periodic monitoring.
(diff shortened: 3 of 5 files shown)


allenpetersen:
Is a useful default protocol and port number to use for subscriptions? The current docs and tests fully qualify but use udp and port 9090. Also the docs describe the destinations as "host" so I could see a user just putting in a hostname and expecting it to work.

allenpetersen:
The InfluxDB glossary says "subscriptions can use TCP or UDP for transmitting the writes." Service.newPointsWriter() only accepts udp/http/https. I wrote validateURL based on the glossary docs. Should the docs and validate function change to match newPointsWriter()?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - [#7621](https://github.com/influxdata/influxdb/issues/7621): Expand string and boolean fields when using a wildcard with `sample()`.
 - [#7616](https://github.com/influxdata/influxdb/pull/7616): Fix chuid argument order in init script @ccasey
 - [#7656](https://github.com/influxdata/influxdb/issues/7656): Fix cross-platform backup/restore
+- [#7615](https://github.com/influxdata/influxdb/issues/7615): Reject invalid subscription urls.
```
e-dard: Please feel free to give yourself credit (`add @allenpetersen`)

services/meta/data.go
```diff
@@ -465,8 +466,33 @@ func (data *Data) DropContinuousQuery(database, name string) error {
 	return ErrContinuousQueryNotFound
 }
 
+// validateURL returns an error if the URL does not have a port or uses a scheme other than UDP or TCP.
+func validateURL(input string) error {
+	u, err := url.Parse(input)
+	if err != nil {
+		fmt.Println(err)
```
e-dard: Remove this.

services/meta/data.go
```diff
@@ -465,8 +466,33 @@ func (data *Data) DropContinuousQuery(database, name string) error {
...
+		return ErrInvalidSubscriptionURL(input)
+	}
+
+	if u.Scheme != "udp" && u.Scheme != "tcp" {
+		return ErrInvalidSubscriptionURL(input)
+	}
+
+	hostParts := strings.Split(u.Host, ":")
```
e-dard: I'm not sure what the intention is here. Are you checking that a port has been set on the URL?

services/meta/data.go
```diff
@@ -465,8 +466,33 @@ func (data *Data) DropContinuousQuery(database, name string) error {
...
+		return ErrInvalidSubscriptionURL(input)
+	}
+	return nil
+}
+
 // CreateSubscription adds a named subscription to a database and retention policy.
 func (data *Data) CreateSubscription(database, rp, name, mode string, destinations []string) error {
+	for _, d := range destinations {
+		if err := validateURL(d); err != nil {
+			return ErrInvalidSubscriptionURL(d)
```
e-dard: `validateURL` only returns an error created with `ErrInvalidSubscriptionURL` so I think you can just `return err` here.

services/subscriber/service.go
```diff
@@ -183,20 +183,28 @@ func (s *Service) createSubscription(se subEntry, mode string, destinations []st
 	default:
 		return nil, fmt.Errorf("unknown balance mode %q", mode)
 	}
-	writers := make([]PointsWriter, len(destinations))
-	stats := make([]writerStats, len(writers))
-	for i, dest := range destinations {
+	writers := []PointsWriter{}
```
e-dard: Why have you changed these? Usually slices have been pre-allocated for a reason.

services/subscriber/service.go
```diff
@@ -183,20 +183,28 @@ func (s *Service) createSubscription(se subEntry, mode string, destinations []st
...
+	stats := []writerStats{}
+	// add only valid destinations
+	for _, dest := range destinations {
 		u, err := url.Parse(dest)
 		if err != nil {
-			return nil, err
+			s.Logger.Println("failed to parse subscription url:", dest)
+			continue
```
e-dard: What's the rationale for ignoring (aside from logging the error) any invalid subscriptions? It's unlikely operators are going to notice that their subscriptions have not been added unless they're checking the logs.

services/subscriber/service.go
```diff
@@ -183,20 +183,28 @@ func (s *Service) createSubscription(se subEntry, mode string, destinations []st
...
-	stats := make([]writerStats, len(writers))
-	for i, dest := range destinations {
+	writers := []PointsWriter{}
+	stats := []writerStats{}
+	// add only valid destinations
+	for _, dest := range destinations {
 		u, err := url.Parse(dest)
 		if err != nil {
-			return nil, err
+			s.Logger.Println("failed to parse subscription url:", dest)
```
e-dard: Log messages should be correct sentences, and `Printf` is more common in the project, i.e.,

```go
s.Logger.Printf("Failed to parse subscription url: %s", dest)
```

Same issue applies to all other log messages in the PR.

services/subscriber/service.go
```diff
@@ -285,7 +293,8 @@ func (s *Service) updateSubs(wg *sync.WaitGroup) error {
 				}
 				sub, err := s.createSubscription(se, si.Mode, si.Destinations)
 				if err != nil {
-					return err
+					s.Logger.Println("createSubscription failed for:", si.Name)
```
e-dard: `s.Logger.Printf("Subscription creation failed for: %s", si.Name)`

e-dard: Again, shouldn't this be an error? Why would it be sensible to continue after encountering this?

e-dard:
@allenpetersen we don't support raw `tcp` so the implication in the docs is that subscriptions can use UDP or HTTP/HTTPS (TCP).

services/subscriber/service.go
allenpetersen: I first wrote it similar to your suggestion but noticed it didn't match the rest of the file. Being new to the project I just followed the pattern. There are a 11 Logger calls in the file, all of them use Println and most are sentence fragments. I've fixed my log messages.

services/meta/data.go
allenpetersen: oops, forgot to remove this debug logging. fixed.

services/subscriber/service.go
allenpetersen: Returning the error will only silently stop processing the rest of the subscriptions. The new code will log the error and continue setting up the rest of the subscriptions.

services/meta/data.go
allenpetersen: The issue addresses having an invalid scheme/protocol but using udp without knowing the port number is also not correct.

Currently it just requires a non-empty string to be set for the port. This is related to my questions about default port numbers and tcp vs http. Once I better understand the requirements this will most likely change, at a minimum it could verify the port is a positive integer.

e-dard: [net.SplitHostPort](https://golang.org/pkg/net/#SplitHostPort) might be a better fit, here.

services/subscriber/service.go
e-dard: You can pre-allocate and still use `append`. You just need to allocate the capacity and a length of zero. For example: `writers := make([]PointsWriter, 0, len(destinations))`

allenpetersen: I think using a slice literal is cleaner and easier to read. It is my default unless there is a clear performance reason to do otherwise.

Checking now I notice that using make with a zero length is a common pattern in the InfluxDB codebase. I'll change these lines.

e-dard:
@nathanielc can you take a look at this please?

allenpetersen:
I removed tcp and added http(s). I suppose that if they are using http(s) there are known default ports but Kapacitor's default port is 9092 so using 80/443 is most likely not what the user wants to do.

Should create subscription require a fully qualified `protocol://host:port` destination or apply some defaults for parts that are not supplied?

nathanielc:
>Should create subscription require a fully qualified protocol://host:port destination or apply some defaults for parts that are not supplied?

For now I would require a fully qualified `protocol://host:port`. If we find a useful set of defaults they can be added later.

services/subscriber/service.go
```diff
@@ -285,7 +293,8 @@ func (s *Service) updateSubs(wg *sync.WaitGroup) error {
 				}
 				sub, err := s.createSubscription(se, si.Mode, si.Destinations)
 				if err != nil {
-					return err
+					s.Logger.Printf("Subscription creation failed for: %s", si.Name)
+					continue
```
nathanielc: Now that no error is returned we should change the return of the updateSubs method to return nothing, since there is no longer any case where an error is returned.

f we run into this error it means that we have a configured subscription but it is not running. Since there is no way to query the status of the subscriptions we should add a statistic for this error so that it is easier to detect. That way a user can setup an alert on the statistic to know if/when subscriptions are failing.

allenpetersen: I removed the unnecessary error return value and a new `createFailures` statistic.

nathanielc: I expected the `CreateFailures` stat to be incremented after a call to `createSubscription` so that one call can increment the stat by at most one. Can you change it so that the stat is incremented only after the call to `createSubscriptions`. 

If we also want to track the failures per destination then a new stats should be added to the `writerStats` type.

allenpetersen: I simplified things a little since the new validation should prevent us from getting into this code path. Subscriptions are back to being 'all or nothing' but one bad subscription doesn't prevent others from being created.

If a subscription has a single bad destination it fails with an error and the `CreateFailures` stat will get incremented. Now there is no need to track the per destination failures.

services/subscriber/service.go
```diff
@@ -183,20 +186,22 @@ func (s *Service) createSubscription(se subEntry, mode string, destinations []st
...
-	stats := make([]writerStats, len(writers))
-	for i, dest := range destinations {
+	writers := make([]PointsWriter, 0, len(destinations))
+	stats := make([]writerStats, 0, len(destinations))
+	// add only valid destinations
+	for _, dest := range destinations {
 		u, err := url.Parse(dest)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("Failed to parse subscription url: %s", dest)
```
e-dard: Error messages should not be capitalised. `s/Failed/failed` Same thing a few lines down

allenpetersen:
@e-dard @nathanielc I think I've addressed all of the review comments. There was a conflict with the changelog so I rebased the commits.

e-dard:
@nathanielc can you approve if this now LGTY.

nathanielc:
@allenpetersen @e-dard Looks like the error messages are still capitalized, maybe the rebase messed up the intended change. Other than that LGTM.

allenpetersen:
@nathanielc Yes, I did miss a commit on my last rebase. I've rebased the branch again and it should be good to go now.

nathanielc:
@e-dard LGTM

e-dard:
Nice work @allenpetersen; thanks!

PR #7679, opened 2016-12-01
Fix string fields w/ trailing slashes

jwilder:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

A string field w/ a trailing slash before the quote would parse incorrectly
because the quote would be seen as escaped.  We have to treat \\ as an
escape sequence within strings in order to handle this.

@e-dard

### DIFF
--- a/models/points.go
+++ b/models/points.go
@@ -1038,8 +1038,8 @@ func scanFieldValue(buf []byte, i int) (int, []byte) {
 	start := i
 	quoted := false
 	for i < len(buf) {
-		// Only escape char for a field value is a double-quote
-		if buf[i] == '\\' && i+1 < len(buf) && buf[i+1] == '"' {
+		// Only escape char for a field value is a double-quote and backslash
+		if buf[i] == '\\' && i+1 < len(buf) && (buf[i+1] == '"' || buf[i+1] == '\\') {
 			i += 2
 			continue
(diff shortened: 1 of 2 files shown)


models/points.go
```diff
@@ -1038,8 +1038,8 @@ func scanFieldValue(buf []byte, i int) (int, []byte) {
 	start := i
 	quoted := false
 	for i < len(buf) {
-		// Only escape char for a field value is a double-quote
-		if buf[i] == '\\' && i+1 < len(buf) && buf[i+1] == '"' {
+		// Only escape char for a field value is a double-quote and backslash
+		if buf[i] == '\\' && i+1 < len(buf) && (buf[i+1] == '"' || buf[i+1] == '\\') {
```
e-dard: From recollection of when I did the [er-escape-backslash](https://github.com/influxdata/influxdb/compare/er-escape-backslash) branch, I don't think this quite captures the desirable behaviour. 

If the rule is "the only escapable characters are a double quote and a backslash" then what happens if we hit a backslash that isn't followed by a double quote or a backslash. For example:

```
cpu value="foo\s"
```

This current approach I think will skip over the single slash silently, but I feel like maybe it should be an error? Possibly we have discussed this before..

Issue #7687, opened 2016-12-05
Support for bound parameters in V2 Client


harryrose:
### Feature Request

__Proposal:__ Support for bound parameters was added in #6634.  However, the golang client library was not updated to support the new behaviour.  I propose that we extend the `Query` struct to contain a map of parameter names to values, which are then posted to the InfluxDB client when the query is executed.

__Use case:__ Developers may build queries which depend on user-provided strings.  For example, filtering results using a name provided by a user.

PR #6634, opened 2016-05-16
Support bound parameters in the parser

jsternberg:
The parser can be passed a map of keys to literal values to be replaced
into the query. Parameters are preceded by a dollar sign (`$`). If a
parameter key is missing, an error is thrown by the parser.

Fixes #2926.

PR #7688, opened 2016-12-05
Add bound parameters map to Query object

harryrose:
Addresses #7687.  Support for bound parameters was added in #6634. However, the golang client package was not updated to support the new behaviour. This PR extends the Query object to take a list of parameters, which are POSTed to the influxdb server when the query is executed.

- [x] CHANGELOG.md updated (No, as it is a trivial change which doesn't affect the server)
- [x] Rebased/mergable
- [x] Tests pass (See below)

## Tests
I've added a test that the parameters are encoded, (which passes).

The following tests fail when I run ` go test -v ./...`.  However, this test is also failing with a clean check-out of the master repository, so I believe it is unrelated to this PR.

FAIL	github.com/influxdata/influxdb/tsdb/engine/tsm1	21.017s

```
[tsm1] 2016/12/05 16:14:47 Snapshot for path /tmp/tsm497135727 written in 5.765042ms
--- FAIL: TestEngine_Backup (0.02s)
	engine_test.go:185: file name doesn't match:
			got: 000000002-000000001.tsm
			exp: /tmp/tsm497135727/000000001-000000001.tsm
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The stress tool `influx_stress` will be removed in a subsequent release. We reco
 - [#7036](https://github.com/influxdata/influxdb/issues/7036): Switch logging to use structured logging everywhere.
 - [#3188](https://github.com/influxdata/influxdb/issues/3188): [CLI feature request] USE retention policy for queries.
 - [#7709](https://github.com/influxdata/influxdb/pull/7709): Add clear command to cli.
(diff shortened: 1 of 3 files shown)


harryrose:
This is probably redundant now, as influxdata/influxdb-client#3 has an implementation.

jsternberg:
Can you rebase this and update the changelog? That other repository isn't yet ready for primetime and it would be nice to offer bound parameters with the old API for those who want to use that.

harryrose:
Hi.  sorry it has taken me a while to get to this.  I've rebased, and updated the changelog as requested.

PR #7698, opened 2016-12-07
CLI was caching db/rp for insert into statements

corylanou:
## Before:

```sql
> create database foo
> use foo
Using database foo
> insert cpu value=1
> insert into badrp mem value=100
Using retention policy badrp
ERR: {"error":"retention policy not found: badrp"}

> insert cpu value=2
ERR: {"error":"retention policy not found: badrp"}
```

## After:
```
> create database foo
> use foo
Using database foo
> insert cpu value=1
> insert into badrp mem value=100
ERR: {"error":"retention policy not found: badrp"}

> insert cpu value=2
> insert into autogen mem value=100
> select * from cpu
name: cpu
time                value
----                -----
1481123342516456239 1
1481123351091949866 2

> select * from mem
name: mem
time                value
----                -----
1481123368541597661 100
```

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): https://github.com/influxdata/docs.influxdata.com/pull/914


joelegasse:
Have you filed a docs issue to remove the note at the end of the `insert` section on the CLI page?

https://docs.influxdata.com/influxdb/v1.1/tools/shell/#write-data-to-influxdb-with-insert

PR #7705, opened 2016-12-08
Optimize Values.Deduplicate

jwilder:
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

`Deduplicate` is called from various places in the engine and can cause
a lot of garbage to get created.  It first creates a map and then
adds each value to the map in order (1st alloc).  It then creates a
new slice (2nd alloc) and appends everything from the map to the slice.
Finally, it sorted the new slice (3rd alloc).

This switches the algorithm to use stable sorting and reusing the existing
slice to avoid allocations.

```
$ benchcmp /tmp/old.txt /tmp/new.txt
benchmark                         old ns/op     new ns/op     delta
BenchmarkValues_Deduplicate-8     447726        57155         -87.23%

benchmark                         old allocs     new allocs     delta
BenchmarkValues_Deduplicate-8     67             1              -98.51%

benchmark                         old bytes     new bytes     delta
BenchmarkValues_Deduplicate-8     140583        32            -99.98%
```

### DIFF
--- a/tsdb/engine/tsm1/encoding.gen.go.tmpl
+++ b/tsdb/engine/tsm1/encoding.gen.go.tmpl
@@ -53,18 +53,20 @@ func (a {{.Name}}Values) assertOrdered() {
 // Deduplicate returns a new slice with any values that have the same timestamp removed.
 // The Value that appears last in the slice is the one that is kept.
 func (a {{.Name}}Values) Deduplicate() {{.Name}}Values {
-	m := make(map[int64]{{.Name}}Value)
-	for _, val := range a {
-		m[val.UnixNano()] = val
+	if len(a) == 0 {
+		return a
 	}
+	sort.Stable(a)
+	var i int
+	for j := 1; j < len(a); j++ {
+		v := a[j]
+		if v.UnixNano() != a[i].UnixNano() {
+			i++
+		}
+		a[i] = v
 
-	other := make({{.Name}}Values, 0, len(m))
-	for _, val := range m {
-		other = append(other, val)
 	}
-
-	sort.Sort(other)
-	return other
+	return a[:i+1]
 }
 
 //  Exclude returns the subset of values not in [min, max]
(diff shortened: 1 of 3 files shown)


toddboom:
Nice one.

tsdb/engine/tsm1/encoding.gen.go.tmpl
```diff
@@ -53,18 +53,23 @@ func (a {{.Name}}Values) assertOrdered() {
...
 func (a {{.Name}}Values) Deduplicate() {{.Name}}Values {
-	m := make(map[int64]{{.Name}}Value)
-	for _, val := range a {
-		m[val.UnixNano()] = val
+	if len(a) == 0 {
+		return a
 	}
+	sort.Stable(a)
+	var i int
+	for j, v := range a {
```
e-dard: How about for j := 1; j < len(a); j++{ and then you can drop the if j == 0 check?

tsdb/engine/tsm1/encoding.gen.go.tmpl
```diff
@@ -53,18 +53,23 @@ func (a {{.Name}}Values) assertOrdered() {
 // Deduplicate returns a new slice with any values that have the same timestamp removed.
 // The Value that appears last in the slice is the one that is kept.
 func (a {{.Name}}Values) Deduplicate() {{.Name}}Values {
-	m := make(map[int64]{{.Name}}Value)
-	for _, val := range a {
-		m[val.UnixNano()] = val
+	if len(a) == 0 {
+		return a
 	}
+	sort.Stable(a)
```
e-dard: This could be more expensive than an unstable sort. Do we need it to be stable given we're only going to be removing duplicate elements anyway?

jwilder: Yes.  It's important that it's a stable sort because duplicates (t1 == t2), may have different values and we need the values added later to overwrite existing values.  If it's not a stable sort, the ordering could get swapped causing the latest write to be lost.

e-dard: Oh wow OK!

PR #7717, opened 2016-12-13
Use local RNG in SampleReducer

mark-rushakoff:
- [x] Rebased/mergable
- [ ] Tests pass

The reducers already had a local RNG but mistakenly did not use it when
sampling points.

Because the local RNG is not protected by a mutex, there is a slight
speedup as a result of this change:

```
benchmark                          old ns/op     new ns/op     delta
BenchmarkSampleIterator_1k-4       418           418           +0.00%
BenchmarkSampleIterator_100k-4     434           422           -2.76%
BenchmarkSampleIterator_1M-4       449           439           -2.23%

benchmark                          old allocs     new allocs     delta
BenchmarkSampleIterator_1k-4       3              3              +0.00%
BenchmarkSampleIterator_100k-4     3              3              +0.00%
BenchmarkSampleIterator_1M-4       3              3              +0.00%

benchmark                          old bytes     new bytes     delta
BenchmarkSampleIterator_1k-4       304           304           +0.00%
BenchmarkSampleIterator_100k-4     304           304           +0.00%
BenchmarkSampleIterator_1M-4       304           304           +0.00%
```

The speedup would presumably increase when multiple sample iterators are
used concurrently.

### DIFF
--- a/influxql/functions.gen.go
+++ b/influxql/functions.gen.go
@@ -409,7 +409,7 @@ func (r *FloatSampleReducer) AggregateFloat(p *FloatPoint) {
 	// Generate a random integer between 1 and the count and
 	// if that number is less than the length of the slice
 	// replace the point at that index rnd with p.
-	rnd := rand.Intn(r.count)
+	rnd := r.rng.Intn(r.count)
 	if rnd < len(r.points) {
 		r.points[rnd] = *p
 	}
@@ -823,7 +823,7 @@ func (r *IntegerSampleReducer) AggregateInteger(p *IntegerPoint) {
 	// Generate a random integer between 1 and the count and
 	// if that number is less than the length of the slice
 	// replace the point at that index rnd with p.
-	rnd := rand.Intn(r.count)
+	rnd := r.rng.Intn(r.count)
 	if rnd < len(r.points) {
 		r.points[rnd] = *p
 	}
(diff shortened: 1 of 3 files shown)


desa:
@mark-rushakoff any reason why the appveyor tests are taking so long?

mark-rushakoff:
Looks like something wrong on AppVeyor's side. I'll force push this branch in hopes of re-triggering a build.

desa:
LGTM. Pending Appveyor actually working.

mark-rushakoff:
@desa the AllSamplesSeen test is failing repeatedly on Appveyor, even though I bumped up the iterations from 6 to now 24. It seems like increasing it more is masking an underlying issue. Thoughts?

desa:
Very weird. Is there a way to ssh into the container like on Circle? If we change it back to using the rng with a mutex does the problem still exist?

mark-rushakoff:
I don't think you can debug the containers on Appveyor. 

I think it was passing before because:
* Go tests run in their own executable per package
* The test was using the global RNG, so it was working off the same seed every time, which happened to pass the test at 6 iterations

I could believe an occasional flaky run at 6 or 24 iterations due to "bad luck" with the RNG, but it's very concerning that it's failed multiple times in a row at 24 iterations with what appears to be a properly seeded, local RNG.

I can try a separate branch with the global RNG again, perhaps seeding it, but I'd expect that to behave the same as it does currently.

desa:
Yeah I completely agree. The thing that I find odd is that it isn't an issue with the Circle build. Maybe it's a windows thing?

PR #7724, opened 2016-12-14
Use logfmt as the default output for influxdb logging

jsternberg:


cmd/influxd/main.go
```diff
@@ -61,7 +62,7 @@ type Main struct {
 func NewMain() *Main {
 	return &Main{
 		Logger: zap.New(
-			zap.NewTextEncoder(),
+			zaplogfmt.NewEncoder(),
```
e-dard: What about all the other places we're currently using `zap.NewTextEncoder()`? Do we need to replace those too?

jsternberg: Nope. But I'll modify the one in `cmd/influxd/run/server.go` anyway. The rest are in test files.

jwilder:
The current `zap.TextEncoder` is pretty readable.  I don't think we need to change the format again.

jsternberg:
Closing this then.

PR #7727, opened 2016-12-15
Converting all int types to int64 on bound parameters

ricardomaraschini:
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

As the IntegerLiteral uses int64 internally we are unable to set
a bound parameter to any other size of integers except 64.
This patch converts any size of integer to 64 when creating an
IntegerLiteral.

### DIFF
--- a/influxql/parser.go
+++ b/influxql/parser.go
@@ -2539,6 +2539,14 @@ func (p *Parser) parseUnaryExpr() (Expr, error) {
 		switch v := v.(type) {
 		case float64:
 			return &NumberLiteral{Val: v}, nil
+		case int:
+			return &IntegerLiteral{Val: int64(v)}, nil
+		case int8:
+			return &IntegerLiteral{Val: int64(v)}, nil
+		case int16:
+			return &IntegerLiteral{Val: int64(v)}, nil
+		case int32:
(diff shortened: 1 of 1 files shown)


jsternberg:
Can you give me the context where this is used? I think you would only run into this if you used the package directly in code.

ricardomaraschini:
Well, I understand your thought as that is exactly what I am doing :-D I read a few things about a way to make bound parameters but i could not find how to make it through the client.Query struct. Maybe I am wrong but I am doing the query parser process happens before I set the client.Query.Command value. Well, I guess even if there is a way to implement bound parameters through client.Query I would have the same problem or extra concern.

jsternberg:
Ah, this code is only run on the server which is why I was questioning it. If you need to do this on the client, I don't think we have a way yet to set bound parameters there. Someone submitted a PR for that in #7688 though. I can take another look at that.

ricardomaraschini:
According to #7688 on version 3 we are going to have this feature, that would be great! You may ignore this one. Just as a matter of curiosity I have moved those casts to my own code in order to be able to keep using the official repository. Thanks for your great work.

PR #7733, opened 2016-12-16
TSI cardinality estimation integration

e-dard:
- [ ] Rebased/mergable
- [ ] Tests pass

This PR adds cardinality estimation to TSI.


tsdb/store.go
```diff
@@ -820,6 +816,11 @@ func (s *Store) WriteToShard(shardID uint64, points []models.Point) error {
 	return sh.WritePoints(points)
 }
 
+func (s *Store) CanWrite(n int) bool {
+	// if s.
```
benbjohnson: then?

PR #7735, opened 2016-12-16
Reduce lock contention when snapshotting

jwilder:
- [x] Rebased/mergable
- [x] Tests pass

This reduces some lock contention in the write path when snapshots are running.  This doesn't prevent write timeouts from ever occurring, but does seem to reduce their occurrences.

### DIFF
--- a/tsdb/engine/tsm1/cache.go
+++ b/tsdb/engine/tsm1/cache.go
@@ -83,7 +83,7 @@ func (e *entry) add(values []Value) error {
 		return nil // Nothing to do.
 	}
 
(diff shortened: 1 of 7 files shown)


tsdb/engine/tsm1/cache.go
```diff
@@ -101,22 +88,11 @@ func (e *entry) add(values []Value) error {
 		if e.vtype != valueType(v) {
 			return tsdb.ErrFieldTypeConflict
 		}
-
```
e-dard: Since we're no longer checking if the values are sorted we can alter the comment about them being "out of order" on line 86.

PR #7747, opened 2016-12-19
Miscellaneous lint cleanup

mark-rushakoff:
- [x] Rebased/mergable
- [x] Tests pass

### DIFF
--- a/internal/meta_client.go
+++ b/internal/meta_client.go
@@ -87,7 +87,7 @@ func (c *MetaClientMock) Databases() []meta.DatabaseInfo {
 }
 
 func (c *MetaClientMock) DeleteShardGroup(database string, policy string, id uint64) error {
-	return c.DeleteShardGroup(database, policy, id)
+	return c.DeleteShardGroupFn(database, policy, id)
 }
 
 func (c *MetaClientMock) DropContinuousQuery(database, name string) error {
--- a/tsdb/engine/tsm1/compact.go
+++ b/tsdb/engine/tsm1/compact.go
@@ -288,7 +288,7 @@ func (c *DefaultPlanner) Plan(lastWrite time.Time) []CompactionGroup {
 	generations := c.findGenerations()
 
 	// first check if we should be doing a full compaction because nothing has been written in a long time
-	if c.CompactFullWriteColdDuration > 0 && time.Now().Sub(lastWrite) > c.CompactFullWriteColdDuration && len(generations) > 1 {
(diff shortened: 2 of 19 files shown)


coordinator/statement_executor_test.go
```diff
@@ -196,10 +196,6 @@ func NewQueryExecutor() *QueryExecutor {
 	}
 	e.QueryExecutor.StatementExecutor = e.StatementExecutor
 
-	var out io.Writer = &e.LogOutput
-	if testing.Verbose() {
-		out = io.MultiWriter(out, os.Stderr)
-	}
 	e.QueryExecutor.WithLogger(zap.New(
 		zap.NewTextEncoder(),
 		zap.Output(os.Stderr),
```
jwilder: What this supposed to be `out` from L199?

tsdb/engine/tsm1/compact.go
```diff
@@ -1208,7 +1208,7 @@ func (k *tsmKeyIterator) combine(dedup bool) blocks {
 }
 
 func (k *tsmKeyIterator) chunk(dst blocks) blocks {
-	for len(k.mergedValues) > k.size {
+	if len(k.mergedValues) > k.size {
```
jwilder: This should be `for` since we are slicing values off on L1225 within the loop.

coordinator/statement_executor_test.go
mark-rushakoff: Probably. I'll amend the commit.

tsdb/engine/tsm1/compact.go
jwilder: Hmm.  I'll have to look at when that changed.  That may be a bug.

internal/meta_client.go
```diff
@@ -87,7 +87,7 @@ func (c *MetaClientMock) Databases() []meta.DatabaseInfo {
 }
 
 func (c *MetaClientMock) DeleteShardGroup(database string, policy string, id uint64) error {
-	return c.DeleteShardGroup(database, policy, id)
+	return c.DeleteShardGroupFn(database, policy, id)
```
corylanou: Clearly this was never being called in a test :man_facepalming:

e-dard:
We should try and get this merged soon given it touches a lot of the code-base.

mark-rushakoff:
Waiting on @jwilder to make a decision on the (non-)loop in `(*tsmKeyIterator).chunk`. I can remove that part of the PR if we're going to defer that decision for later.

tsdb/engine/tsm1/compact.go
jwilder: This should be an `if` as you have it.  When the return was added, I forgot to switch it from a loop.

PR #7755, opened 2016-12-21
Mitigate timing attack to identify valid username

mark-rushakoff:
Refer to https://cwe.mitre.org/data/definitions/208.html for details on
this common weakness.

Prior to this change, a locally running influxd instance would respond
to a query with an invalid username in about 14ms, but would take about
93ms to respond when the username was valid but the password was
incorrect. Now, both valid and invalid usernames take about 93ms.

Likewise, the newly added benchmark for an invalid username now takes
about the same time as the benchmark for a valid username with an
invalid password. Prior to this change, the invalid username benchmark
ran in about 0.004% of the invalid password time.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/services/meta/client.go
+++ b/services/meta/client.go
@@ -539,7 +539,23 @@ func (c *Client) Authenticate(username, password string) (*UserInfo, error) {
 	userInfo := c.cacheData.User(username)
 	c.mu.RUnlock()
 	if userInfo == nil {
-		return nil, ErrUserNotFound
+		// Instead of instantly returning when the username is not found, compare against a valid bcrypt hash
+		// so that we don't leak whether the username was valid via how quickly the server responds.
+		// See https://cwe.mitre.org/data/definitions/208.html for more details.
+
+		// Create fake bcrypt hash with Version=2, Cost=<current cost setting>, Salt+Hash=<doesn't really matter>.
+		fakeHash := fmt.Sprintf("$2$%02d$00000000000000000000000000000000000000000000000000000", bcryptCost)
+
+		// Compare the password against the fake bcrypt hash so we eat up the right time according to bcryptCost.
+		err := bcrypt.CompareHashAndPassword([]byte(fakeHash), []byte(password))
+
+		// If we get an error other than ErrMismatchedHashAndPassword, there's a bug in how we generated fakeHash, so return that error.
+		if err != nil && err != bcrypt.ErrMismatchedHashAndPassword {
+			return nil, err
+		}
+
+		// Now that we've spent the right amount of time calculating the hash, return an authentication error.
+		return nil, ErrAuthenticate
 	}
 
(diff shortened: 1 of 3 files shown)


services/meta/client.go
```diff
@@ -539,7 +539,23 @@ func (c *Client) Authenticate(username, password string) (*UserInfo, error) {
 	userInfo := c.cacheData.User(username)
 	c.mu.RUnlock()
 	if userInfo == nil {
-		return nil, ErrUserNotFound
+		// Instead of instantly returning when the username is not found, compare against a valid bcrypt hash
+		// so that we don't leak whether the username was valid via how quickly the server responds.
+		// See https://cwe.mitre.org/data/definitions/208.html for more details.
+
+		// Create fake bcrypt hash with Version=2, Cost=<current cost setting>, Salt+Hash=<doesn't really matter>.
+		fakeHash := fmt.Sprintf("$2$%02d$00000000000000000000000000000000000000000000000000000", bcryptCost)
```
joelegasse: This should probably be done once for the package, rather than every time an unknown user is attempted.

If we decide to keep this change, then this time-wasting hash should be generated with `bcrypt.GenerateFromPassword`, and not `fmt.Sprintf`.

services/meta/client.go
```diff
@@ -539,7 +539,23 @@ func (c *Client) Authenticate(username, password string) (*UserInfo, error) {
...
+
+		// Compare the password against the fake bcrypt hash so we eat up the right time according to bcryptCost.
+		err := bcrypt.CompareHashAndPassword([]byte(fakeHash), []byte(password))
```
joelegasse: If we're trying to match the behavior below of attempting a password on a known user, then this error can just be ignored, and skip straight to returning `ErrAuthenticate` as it does on line 576.

mark-rushakoff: In benchmarking this implementation, I observed that ignoring the error seemed to instruct the compiler to discard the entire call to `bcrypt.CompareHashAndPassword`.

services/meta/client.go
mark-rushakoff: The difficulty of doing this once for the package is that `bcryptCost` is reassigned during in the package tests, so the benchmarks become skewed if this path uses an expensive bcrypt call and all the other paths use bcrypt.MinCost.

It was unclear to me whether `bcrypt.GenerateFromPassword` had the same cost as `bcrypt.CompareHashAndPassword`. I probably should have benchmarked it.

mark-rushakoff:
> I argue that if you have someone doing an on-line password guessing attempt, we're just helping them take down the server by promising to churn the CPU for a bit with every invalid request.

Sold me on closing this PR. Thanks for the well-thought-out response.

PR #7756, opened 2016-12-22
Bump jwt-go to 2.7.0

aviau:
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

This patch was written by @guillemj for Debian. (thanks to him!) I have modified it to also fix tests.

We had to write the patch because we don't have an older version of jwt-go in Debian.


guillemj:
Hey!

Actually the problem in Debian is that we have a too old version, so that's why I didn't submit this upstream. I don't think this should be merged TBH. But I think upgrading jwt-go seemed problematic as it was API incompatible so it would have started a transition in Debian which would be undesirable at this point of the release process.

corylanou:
This is the revision history from the current version to the new one which should be reviewed as part of accepting this PR:

[github.com/dgrijalva/jwt-go](https://github.com/dgrijalva/jwt-go/compare/268038b363c7a8d7306b8e35bf77a1fde4b0c402...24c63f56522a87ec5339cc3567883f1039378fdb)

aviau:
Oh that is my mistake. I'll close this.

@guillemj We can upload 3.X in Debian with a different package name. But that is unrelated to this issue!

PR #7761, opened 2016-12-24
Require fields when marshalling Point

mark-rushakoff:
I haven't been able to reproduce creating a point without any fields,
but we've seen points in the wild that have been marshalled with no
fields - that is, the length header for fields is uint32(0) and a
well-formed encoded time follows.

Attempting to unmarshal points via NewPointFromBytes returns
ErrPointMustHaveAField, so it seems better to fail earlier with the same
error, rather than allowing those points to be serialized in the first
place.

- [x] Rebased/mergable
- [x] Tests pass

### DIFF
--- a/models/points.go
+++ b/models/points.go
@@ -1406,6 +1406,10 @@ func (p *point) StringSize() int {
 }
 
 func (p *point) MarshalBinary() ([]byte, error) {
+	if len(p.fields) == 0 {
+		return nil, ErrPointMustHaveAField
+	}
+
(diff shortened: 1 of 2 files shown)


mark-rushakoff:
The existing test file was in the separate `models_test` package, so I added a new test file in the `models` package, since I couldn't find an obvious way to create a point without fields using the exported API.

corylanou:
I fine with internal tests when needed like this.

PR #7762, opened 2016-12-25
Support timezone offsets for queries

### LINKED ISSUE #6541: Support timezone for a user-friendly way of setting the timezone offset
From a conversation in #2074.

We should support a more user-friendly way to set timezone offsets. In 0.13, you will be able to set an offset with an extra parameter in the `GROUP BY time(...)` section like this:

```
SELECT mean(value) FROM cpu WHERE time > now() - 7d GROUP BY time(1d, -5h)
```

It should also be possible to set this timezone offset through doing something like this:

```
SELECT mean(value) FROM cpu WHERE time > now() - 7d GROUP BY time(1d) TZ("America/New York")
```

This should automatically set the group by offset and could potentially be used for something else like handling daylight savings time switches.

jsternberg:
The timezone for a query can now be added to the end with something like
`TZ("America/Los_Angeles")` and it will localize the results of the
query to be in that timezone. The offset will automatically be set to
the offset for that timezone and offsets will automatically adjust for
daylight savings time so grouping by a day will result in a 25 hour day
once a year and a 23 hour day another day of the year.

The automatic adjustment of intervals for timezone offsets changing will
only happen if the group by period is greater than the timezone offset
would be. That means grouping by an hour or less will not be affected by
daylight savings time, but a 2 hour or 1 day interval will be.

The default timezone is UTC and existing queries are unaffected by this
change.

When times are returned as strings (when `epoch=1` is not used), the
results will be returned using the requested timezone format in RFC3339
format.

Fixes #6541.


jsternberg:
This still requires tests. I've only verified a single query with this and haven't tried the edge cases or tested the efficiency. I also think the changes to the fill iterator are going to need to be different so it's faster, but that's only a hunch at the moment.

For those following this issue and seeing this PR, it will not end up in 1.2 and is potentially targeted at 1.3 at the moment. I just wanted to get it up now because I finally thought of a way to do this and wanted to write the code down before I forgot.

jsternberg:
This now has tests and fixed some bugs when traveling between time offsets. This is likely fine to be reviewed and tested more thoroughly now.

jgeiger:
:+1:  This would be useful for pulling CSV data that no longer needs to be modified by the pulling application.

dgnorton:
@jsternberg I think the syntax should be single quotes.
```
TZ('America/Los_Angeles')
```
Thoughts?

jsternberg:
I can do that. I wasn't sure which one to use anyway since we muddle the water between identifiers and strings in a few too many places.

openbufo:
Updated to 1.2.3 using deb package on ubuntu. While trying queries with TZ('America/Los_Angeles') at the end, I m getting the following error:
ERR: error parsing query: found TZ, expected ; at line 1 
Exact query statement: SELECT mean(value) FROM cpu WHERE time >= '2017-05-04T00:00:00Z' GROUP BY time(1d) TZ("America/Los_Angeles").
Where am I going wrong? Isnt this feature available on 1.2.3 deb pkg yet? Or do i need to build from source?

jsternberg:
Time zones are slated for 1.3. They should be in master right now.

openbufo:
Okay. So, I build influxdb from the git master branch on a docker image. TZ function is now recognizable but cant find the timezone. Using tz('America/Los_Angeles') throws an error "unable to find time zone America/Los_Angeles". Error goes with other timezones as well.

jsternberg:
Please direct usage questions to our community website at https://community.influxdata.com.

jsternberg:
Updated the PR comment to use single quotes instead of double quotes. That got changed during review but never got changed in the PR comment.

PR #7765, opened 2016-12-27
Add modulo operator to the query language

### LINKED ISSUE #7553: % operator
Proposal: Would like the % operator

Current behavior: doesn't exist

Desired behavior: % works

Use case:
I have a data set that has a state value of 0: off, 1: heat, 2: cool.
I'd like to convert this value to be cool < off < heat to make the singlestat grafana plugin able to display it nicely. the % operator could be used to reformat the data nicely in this case.

jsternberg:
Fixes #7553.


influxql/ast_test.go
```diff
@@ -1086,6 +1086,8 @@ func TestReduce(t *testing.T) {
 		{in: `foo(bar(2 + 3), 4)`, out: `foo(bar(5), 4)`},
 		{in: `4 / 0`, out: `0.000`},
 		{in: `1 / 2`, out: `0.500`},
+		{in: `2 % 3`, out: `2`},
+		{in: `5 % 2`, out: `1`},
```
corylanou: Can you test the zero edge cases and negative edge cases?

jsternberg: Done. Looks like a had a bug anyway when doing `float % integer` which is now fixed.

jwilder:
@jsternberg @rkuchan Are there docs that need to be updated?

rkuchan:
@jwilder - thanks! I created a docs issue. Does this operator work in both the `SELECT` clause and the `WHERE` clause?

jsternberg:
It should work in the `WHERE` clause, but it isn't. It looks like I missed some stuff. I'll have a follow-up PR to fix that.

jsternberg:
Fixed those issues in #7990.

PR #7781, opened 2017-01-02
Godoc cleanup

mark-rushakoff:
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

I went through basically the entire project, except for templated or go-generated files, and cleaned up Godoc formatting and tried to add Godoc for all exported symbols. I made minimal changes to actual code, but I'm going to comment on the PR indicating what was changed and why.


cmd/influxd/run/command.go
```diff
@@ -204,7 +205,7 @@ func (cmd *Command) ParseConfig(path string) (*Config, error) {
 	return config, nil
 }
 
-var usage = `Runs the InfluxDB server.
+const usage = `Runs the InfluxDB server.
```
mark-rushakoff: Changed var to const - usage should not change at runtime.

models/points.go
```diff
@@ -26,104 +27,158 @@ var (
 		'=': []byte(`\=`),
 	}
 
-	ErrPointMustHaveAField  = errors.New("point without fields is unsupported")
-	ErrInvalidNumber        = errors.New("invalid number")
-	ErrInvalidPoint         = errors.New("point is invalid")
-	ErrMaxKeyLengthExceeded = errors.New("max key length exceeded")
```
mark-rushakoff: Removed this error - we still report max key length exceeded, but the error is constructed at runtime. `ErrMaxKeyLengthExceeded` is still an exported error in the tsm1 package.

tsdb/engine/tsm1/cache.go
```diff
@@ -21,10 +21,12 @@ import (
 const ringShards = 128
 
 var (
-	ErrCacheInvalidCheckpoint = fmt.Errorf("invalid checkpoint")
```
mark-rushakoff: This error has been unused since b0fb8a0a27d5380f7255275939c3881801caeb3e (Dec 2015).

tsdb/engine/tsm1/reader.go
```diff
@@ -145,11 +148,12 @@ func (b *BlockIterator) Next() bool {
...
 }
 
-func (b *BlockIterator) Read() (string, int64, int64, uint32, []byte, error) {
+// Read reads information about the next block to be iterated.
+func (b *BlockIterator) Read() (key string, minTime int64, maxTime int64, checksum uint32, buf []byte, err error) {
 	if b.err != nil {
 		return "", 0, 0, 0, nil, b.err
 	}
-	checksum, buf, err := b.r.readBytes(&b.entries[0], nil)
+	checksum, buf, err = b.r.readBytes(&b.entries[0], nil)
```
mark-rushakoff: I named the return values in this method, which changed this assignment all to existing variables.

PR #7784, opened 2017-01-03
Fix broken return statements swallowing errors

mark-rushakoff:
There was no comment on either case specifying that the `return nil`
was deliberate instead of `return err`, so I'm assuming these were
typos. I added tests to conserve the error-returning behavior.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The stress tool `influx_stress` will be removed in a subsequent release. We reco
 
 ### Bugfixes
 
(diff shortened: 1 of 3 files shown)


mark-rushakoff:
@corylanou your name came up on git blame for the changed lines. Does it seem accurate that these `return nil`s were likely typos, not deliberate?

corylanou:
Doh!  Those had to be cut/paste errors. Ugg.  Nice catch.

PR #7786, opened 2017-01-03
Use one atomic operation in (*Cache).decreaseSize

mark-rushakoff:
The previous implementation was susceptible to a race condition (of
correctness) since c.decreaseSize is called without a lock in
(*Cache).WriteMulti.

There were already tests which asserted the correctness of the result of
decreaseSize, so no tests were added or modified.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/tsdb/engine/tsm1/cache.go
+++ b/tsdb/engine/tsm1/cache.go
@@ -423,8 +423,8 @@ func (c *Cache) increaseSize(delta uint64) {
 
 // decreaseSize decreases size by delta.
 func (c *Cache) decreaseSize(delta uint64) {
-	size := atomic.LoadUint64(&c.size)
-	atomic.StoreUint64(&c.size, size-delta)
+	// Per sync/atomic docs, bit-flip delta minus one to perform subtraction within AddUint64.
+	atomic.AddUint64(&c.size, ^(delta - 1))
 }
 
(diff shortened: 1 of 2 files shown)


tsdb/engine/tsm1/cache.go
```diff
@@ -423,8 +423,9 @@ func (c *Cache) increaseSize(delta uint64) {
 
 // decreaseSize decreases size by delta.
 func (c *Cache) decreaseSize(delta uint64) {
-	size := atomic.LoadUint64(&c.size)
-	atomic.StoreUint64(&c.size, size-delta)
+	// Since there isn't an explicit atomic operation for subtraction,
+	// use addition and overflow to effectively subtract delta from c.size.
+	atomic.AddUint64(&c.size, math.MaxUint64-delta+1)
```
jwilder: nit: I think this is still correct, but the docs suggest a slightly shorter version:

```
atomic.AddUint64(&c.size, ^(delta-1))
```

https://golang.org/pkg/sync/atomic/#AddUint64

mark-rushakoff: I overlooked that in the docs. I just pushed an amended commit to use that approach.

PR #7789, opened 2017-01-03
Require database name on monitor config

mark-rushakoff:
- [x] Rebased/mergable
- [x] Tests pass

### DIFF
--- a/monitor/config.go
+++ b/monitor/config.go
@@ -40,5 +40,8 @@ func (c Config) Validate() error {
 	if c.StoreInterval <= 0 {
 		return errors.New("monitor store interval must be positive")
 	}
+	if c.StoreDatabase == "" {
+		return errors.New("monitor store database name must not be empty")
+	}
 	return nil
 }
--- a/monitor/config_test.go
+++ b/monitor/config_test.go
@@ -28,3 +28,25 @@ store-interval="10m"
 		t.Fatalf("unexpected store-interval:  %s", c.StoreInterval)
 	}
 }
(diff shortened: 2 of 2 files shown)


jwilder:
Or is this preventing the case where someone explicitly configures it to `""`?

mark-rushakoff:
It covers the explicit set to empty. Given a monitor section of the config:

```toml
[monitor]
  store-enabled = true
  store-database = "" # Explicitly empty for some reason
  store-interval = "10s"
```

you will now get the error:

```
[I] 2017-01-03T23:21:00Z Using configuration at: /tmp/ic

run: monitor store database name must not be empty. To generate a valid configuration file run `influxd config > influxdb.generated.conf`
exit status 1
```

whereas before, influxd would start correctly, and you would see:

```
[I] 2017-01-03T23:07:48Z Storing statistics in database '' retention policy 'monitor', at interval 10s service=monitor
...
[I] 2017-01-03T23:08:00Z failed to create database '', failed to create storage: database name required service=monitor
[I] 2017-01-03T23:08:00Z failed to store statistics: database not found:  service=monitor
[I] 2017-01-03T23:08:10Z failed to create database '', failed to create storage: database name required service=monitor
[I] 2017-01-03T23:08:10Z failed to store statistics: database not found:  service=monitor
```

If the `store-database` line is commented out or otherwise missing, influxdb still falls back to the default of `_internal`.

jwilder:
Ah.  Makes sense now.

PR #7803, opened 2017-01-07
Added option to follow redirects and use last URL in redirect path.

lrsmith:
This is code to address https://github.com/influxdata/telegraf/issues/2191
For redirects and puts/posts they are not followed. This PR creates an option 'InsecureFollowRedirect' which is triggered will have an initial GET done and the redirect will be followed and the URL for the last redirect will be placed in the connection.

- [ ] Rebased/mergable
- [X] Tests pass
- [ ] CHANGELOG.md updated
- [X ] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

- [ ] Provide example syntax

### DIFF
--- a/client/v2/client.go
+++ b/client/v2/client.go
@@ -77,6 +77,8 @@ type Client interface {
 	// the UDP client.
 	Query(q Query) (*Response, error)
 
+	CheckHTTPRedirect() (bool, error)
+
 	// Close releases any resources a Client may be using.
 	Close() error
 }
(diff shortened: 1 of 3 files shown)


e-dard:
Thanks for the PR @lrsmith.

I'm currently 👎  on this though, because I don't think there is a common enough need for this feature. Other members of the team may feel differently though so I'll leave this PR open and see if there is some more feedback.

jsternberg:
I'm a big 👎 on this too. There's a very good reason why POST and PUT's do not follow redirects. Looking at the original issue, it looks like you should either have an Nginx proxy in front of InfluxDB that can handle both HTTP and HTTPS or just fix your configuration to use HTTPS. You can use the Nginx proxy to accept HTTP connections while you work on fixing your configuration files to communicate over HTTPS and then switch everything to HTTPS.

I'm going to close this since there are two members of the team who don't like this for two different reasons.

PR #7809, opened 2017-01-09
Fix race in (*tsm1.Cache).values

mark-rushakoff:
Without this read lock, this race would happen during a concurrent
snapshot compaction and query.

The race was observed in [this build](https://circleci.com/gh/influxdata/influxdb/14150?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) and was easy to reproduce locally with `GOMAXPROCS=1 go test -race -run=TestServer_BackupAndRestore ./cmd/influxd/run`.

- [x] Rebased/mergable
- [x] Tests pass

### DIFF
--- a/tsdb/engine/tsm1/cache.go
+++ b/tsdb/engine/tsm1/cache.go
@@ -554,14 +554,18 @@ func (c *Cache) SetMaxSize(size uint64) {
 	c.mu.Unlock()
 }
 
-// values returns the values for the key. It doesn't lock and assumes the data is
-// already sorted. Should only be used in compact.go in the CacheKeyIterator
+// values returns the values for the key. It assumes the data is already sorted.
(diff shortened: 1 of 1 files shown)


tsdb/engine/tsm1/cache.go
```diff
@@ -561,7 +561,10 @@ func (c *Cache) values(key string) Values {
 	if e == nil {
 		return nil
 	}
-	return e.values
+	e.mu.RLock()
```
e-dard: The method comment should probably be updated to reflect the fact a lock is now taken.

mark-rushakoff: Comment updated.

e-dard:
Good stuff @mark-rushakoff \o/

PR #7813, opened 2017-01-10
Fix #7812 - Panic when pruning shard groups

### LINKED ISSUE #7812: Panic: slice out of bounds when pruning shard groups
### Bug report

__System info:__ [Influxdb 1.2.0 [unreleased] built from current master branch, Ubuntu 14.04 64Bits, Go 1.7.4.]

__Steps to reproduce:__

I have an instance for testing and it has a lot of databases and shards. This database is tested rarely, making it full of expired shards. With the latest master, I ran into this issue.

So, if you have at least one of the following conditions, this could also happen to you:

1. Have more than one database or
2. Deleting more than one shard group at once

__Expected behavior:__ No panic

__Actual behavior:__ Slice out of bounds panic and process shutdown because of it

__Additional info:__
 
The problem is in the following function https://github.com/influxdata/influxdb/blob/master/services/meta/client.go#L671, which was introduced by PR #7610 .
It has two main problems:
* It is using the index variable i from the Databases loop, which will cause an out of bounds error if we have more databases than shard groups;
* It is changing the slice it is iterating for, which may cause problems when removing as the index changes when we change the slice (eg: we can remove the wrong shard group or cause an out of bounds panic if the shard group is beyond the current size of the slice).

I will send a PR trying to fix this.

vladlopes:
- [x] Rebased/mergable
- [X] Tests pass
- [X] CHANGELOG.md updated

This PR fixes #7812 using this filtering trick from the [Go](https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating).

I changed the test to trigger the behavior described in the issue asserting that this change fix it.

The only problem with this solution is that we will be changing the retention policy shard groups slice every time this function is called, different from the previous one which only changed when deleting (we had performance, but no correctness).

### DIFF
--- a/services/meta/client.go
+++ b/services/meta/client.go
@@ -676,15 +676,15 @@ func (c *Client) PruneShardGroups() error {
 	data := c.cacheData.Clone()
 	for i, d := range data.Databases {
 		for j, rp := range d.RetentionPolicies {
+			var remainingShardGroups []ShardGroupInfo
 			for _, sgi := range rp.ShardGroups {
 				if sgi.DeletedAt.IsZero() || !expiration.After(sgi.DeletedAt) {
+					remainingShardGroups = append(remainingShardGroups, sgi)
 					continue
 				}
-				// we are safe to delete the shard group as it's been marked deleted for the required expiration
-				s := append(rp.ShardGroups[:i], rp.ShardGroups[i+1:]...)
-				data.Databases[i].RetentionPolicies[j].ShardGroups = s
 				changed = true
 			}
+			data.Databases[i].RetentionPolicies[j].ShardGroups = remainingShardGroups
 		}
 	}
 	if changed {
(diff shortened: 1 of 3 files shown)


services/meta/client.go
```diff
@@ -676,15 +676,15 @@ func (c *Client) PruneShardGroups() error {
 	data := c.cacheData.Clone()
 	for i, d := range data.Databases {
 		for j, rp := range d.RetentionPolicies {
+			remainingshards := data.Databases[i].RetentionPolicies[j].ShardGroups[:0]
```
corylanou: This would be more readable as:

```go
var remainingShards []ShardGroupInfo
```

vladlopes: I should definitely change the name of the variable. I think ```remainingShardGroups``` would be even better. But if I declare a new variable, we will loose the "no allocation" we gain using this trick. In the end, we are not allocating a new array, just using the old one. After the appends, it will have the size that we need, but, if fact, what we did was change the contents of the original array.
But it is up to you to decide... Maybe we can consider a premature optimization...

corylanou: This code is rarely called, so optimizations are not really a concern here.  I understand that we can share the memory, but I would favor readability here over a premature optimization.  Thanks!

vladlopes: Changed the name to ```remainingShardGroups``` and removed the allocation trick

PR #7820, opened 2017-01-11
fix marshaling of zero time

### LINKED ISSUE #7689: Unix epoch time in ShardGroupInfo changes to zero-time
### Bug report

__System info:__ InfluxDB (v1.1), Linux

__Steps to reproduce:__

1. Create a retention policy with infinite duration and `1d` shard duration
2. Insert data at time `0`
3. Restart instance

the `1d` shard duration isn't important, but it ensures that the start time of the shard including the unix epoch will be the unix epoch.

__Expected behavior:__ 

The shard info should be preserved through the restart

__Actual behavior:__

The timestamp of `1970-01-01T00:00:00Z` turns to `0001-01-01T00:00:00Z` for the start time of the created shard

__Additional info:__

If any of the timestamps of a `ShardGroupInfo` struct happen to be the unix epoch, they will be interpreted as the zero time because of the way the protobuf is defined and the decisions made around encoding timestamps.

https://github.com/influxdata/influxdb/blob/master/services/meta/data.go#L1482

Both `UnmarshalTime` and `MarshalTime` use `0` to represent no time being set, but then return/decode any other time as the unix nanoseconds. This means that `MarshalTime` will return `0` for both unix epoch and the zero `time.Time` value. `UnmarshalTime` has to pick one, and chooses the zero time.

The binary form of the metadata is only used to load/store, and will only be noticed after a restart:
```
> create retention policy thing on foo duration inf replication 1 shard duration 1d default
> show retention policies on foo
name    duration        shardGroupDuration      replicaN        default
----    --------        ------------------      --------        -------
thing   0s              24h0m0s                 1               true

> use foo
Using database foo
> insert x v=1 0
> show shards
name: foo
id      database        retention_policy        shard_group     start_time              end_time                expiry_time             owners
--      --------        ----------------        -----------     ----------              --------                -----------             ------
3       foo             thing                   3               1970-01-01T00:00:00Z    1970-01-02T00:00:00Z    1970-01-02T00:00:00Z
```
Note the correct timestamp, but after a restart:
```
> show shards
name: foo
id      database        retention_policy        shard_group     start_time              end_time                expiry_time             owners
--      --------        ----------------        -----------     ----------              --------                -----------             ------
3       foo             thing                   3               0001-01-01T00:00:00Z    1970-01-02T00:00:00Z    1970-01-02T00:00:00Z
```
It changes to the zero time.

This could be fixed for start/end times by not checking for zero, since they are required fields. `DeletedAt` is set to `time.Now()` when shards are marked for deletion, so as long as people don't have their clocks set that far in the past, that shouldn't cause a problem, and the current logic of "zero means unset" should be fine there.

`TruncatedAt` is somewhere in the middle, since the time being relative to start/end is important. This means that if a shard group spans the unix epoch, `0` is a perfectly valid value for `TruncatedAt`. That being said, in practice, it is only used currently for real-time changes to shard groups, and should also not be a problem.

corylanou:
There is a very rare edge case that this will introduce.  If TruncatedAt or DeletedAt happens at epoch, when it's marshaled, it will get reset to zero time again.  The bigger fix would be to make all zero time defaults actually be epoch, but that would introduce yet another edge case likely.

This is what I came up with so far.  Let's discuss.

//cc @jwilder @joelegasse 

fixes #7689 

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/services/meta/data.go
+++ b/services/meta/data.go
@@ -1248,8 +1248,16 @@ func (sgi *ShardGroupInfo) marshal() *internal.ShardGroupInfo {
 // unmarshal deserializes from a protobuf representation.
 func (sgi *ShardGroupInfo) unmarshal(pb *internal.ShardGroupInfo) {
 	sgi.ID = pb.GetID()
-	sgi.StartTime = UnmarshalTime(pb.GetStartTime())
-	sgi.EndTime = UnmarshalTime(pb.GetEndTime())
+	if i := pb.GetStartTime(); i == 0 {
+		sgi.StartTime = time.Unix(0, 0).UTC()
+	} else {
+		sgi.StartTime = UnmarshalTime(i)
+	}
+	if i := pb.GetEndTime(); i == 0 {
+		sgi.EndTime = time.Unix(0, 0).UTC()
+	} else {
+		sgi.EndTime = UnmarshalTime(i)
+	}
 	sgi.DeletedAt = UnmarshalTime(pb.GetDeletedAt())
 
 	if pb != nil && pb.TruncatedAt != nil {
--- a/services/meta/data_internal_test.go
+++ b/services/meta/data_internal_test.go
@@ -54,3 +54,36 @@ func TestShardGroupSort(t *testing.T) {
 		t.Fatal("unstable sort for ShardGroupInfos")
 	}
 }
+
+func Test_Data_RetentionPolicy_MarshalBinary(t *testing.T) {
+	zeroTime := time.Time{}
+	epoch := time.Unix(0, 0).UTC()
+
+	startTime := zeroTime
+	sgi := &ShardGroupInfo{
+		StartTime: startTime,
+	}
+	isgi := sgi.marshal()
+	sgi.unmarshal(isgi)
+	if got, exp := sgi.StartTime.UTC(), epoch.UTC(); got != exp {
+		t.Errorf("unexpected start time.  got: %s, exp: %s", got, exp)
+	}
+
+	startTime = time.Unix(0, 0)
+	endTime := startTime.Add(time.Hour * 24)
+	sgi = &ShardGroupInfo{
+		StartTime: startTime,
+		EndTime:   endTime,
+	}
+	isgi = sgi.marshal()
+	sgi.unmarshal(isgi)
+	if got, exp := sgi.StartTime.UTC(), startTime.UTC(); got != exp {
+		t.Errorf("unexpected start time.  got: %s, exp: %s", got, exp)
+	}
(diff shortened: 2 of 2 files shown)


services/meta/data.go
```diff
@@ -1531,16 +1531,19 @@ func (leases *Leases) Acquire(name string, nodeID uint64) (*Lease, error) {
 
 // MarshalTime converts t to nanoseconds since epoch. A zero time returns 0.
 func MarshalTime(t time.Time) int64 {
-	if t.IsZero() {
+	if t.IsZero() || t.UTC() == time.Unix(0, 0).UTC() {
```
joelegasse: This changes nothing since `t.UnixNano()` will be `0`, anyway

services/meta/data.go
```diff
@@ -1248,12 +1248,12 @@ func (sgi *ShardGroupInfo) marshal() *internal.ShardGroupInfo {
 // unmarshal deserializes from a protobuf representation.
 func (sgi *ShardGroupInfo) unmarshal(pb *internal.ShardGroupInfo) {
 	sgi.ID = pb.GetID()
-	sgi.StartTime = UnmarshalTime(pb.GetStartTime())
-	sgi.EndTime = UnmarshalTime(pb.GetEndTime())
-	sgi.DeletedAt = UnmarshalTime(pb.GetDeletedAt())
+	sgi.StartTime = UnmarshalTime(pb.GetStartTime(), true)
```
joelegasse: Rather than modifying `UnmarshalTime`, we could just add some code below that runs: `if sgi.(Start|End)Time.IsZero() { sgi.(Start|End)Time = epochTime }`

services/meta/data.go
corylanou: That's what I thought, but that's not true.  `UnixNano` on a zero time is actually undefined.  

```go
package main

import "time"

func main() {
	var t time.Time
	println(t.String())
	i := t.UnixNano()
	println(i)
	t = time.Unix(0, i)
	println(t.String())
}
```

results: 

```
0001-01-01 00:00:00 +0000 UTC
-6795364578871345152
1754-08-30 22:43:41.128654848 +0000 UTC
```

Play Example: https://play.golang.org/p/P0s7lbZtnK

And the source:

https://golang.org/src/time/time.go?s=24148:24178#L827

corylanou: oh, I see, you mean `UnixNano()` will be zero.  That is correct.  I'll fix.

corylanou:
@joelegasse addressed your comments.

PR #7829, opened 2017-01-12
Add some checks before removing directories

### LINKED ISSUE #7822: Drop database will delete /influxdb/data directory
### Bug report

__System info:__ InfluxDB v1.1.0 (downloaded, not built), Ubuntu 15.04 64-bit Intel

__Steps to reproduce:__

1. Startup InfluxDB and create a database.
    `curl -XPOST "http://influxdb:8086/query" --data-urlencode "q=CREATE DATABASE \"Test1\""`

2. Add content to create measurement.
    `curl -XPOST 'http://influxdb:8086/write?db=Test1' --data-binary 'cpu_load_short,host=server01,region=us-west value=0.64 1434055562000000000'`
    You should see the /var/lib/influxdb/data directory with Test1 and _internal sub-directories.

3. Execute 'DROP DATABASE' without a name specifically like the following:
    `curl -XPOST "http://influxdb:8086/query" --data-urlencode "q=DROP DATABASE \"\""`
    No error is returned or logged but the entire /var/lib/influxdb/data is gone (including _internal).

__Expected behavior:__ Return an error if no database name is provided.

__Actual behavior:__ Deleted the entire /influxdb/data directory.

The DROP DATABASE with no name was a bug in my code where a null name got through but the consequences are pretty bad since all the data files are gone. The system worked ok for a couple days until influxdb was restarted of course. Errors about missing files started appearing only in the logs a few seconds after the drop.

joelegasse:
Fixes #7822

This change first ensures that databases and retention policies exist
before attempting to remove them from the Store. It also adds some
checks in the `DeleteDatabase` and `DeleteRetentionPolicy` to ensure
that maliciously named entries won't remove anything outside of the
configured data directory.

This also adds some very basic name validation with the following
plain-english description: names must be non-zero sequence of printable
characters that do not contain slashes ('/' or '\') and are not equal to
either "." or "..".

The intent is that, since we currently just use database and retention
policy names directly as path elements, these rules will hopefully leave
us with names that should be at least close to valid directory names.

Ideally, we would restrict names even further or not use them as path
elements directly, but this should be a step towards the former without
restricting names "too much"

- [x] Rebased/mergable
- [ ] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/coordinator/statement_executor.go
+++ b/coordinator/statement_executor.go
@@ -244,6 +244,12 @@ func (e *StatementExecutor) executeCreateContinuousQueryStatement(q *influxql.Cr
 }
 
 func (e *StatementExecutor) executeCreateDatabaseStatement(stmt *influxql.CreateDatabaseStatement) error {
+	if !meta.ValidName(stmt.Name) {
+		// TODO This should probably be in `(*meta.Data).CreateDatabase`
+		// but can't go there until 1.1 is used everywhere
+		return meta.ErrInvalidName
+	}
+
 	if !stmt.RetentionPolicyCreate {
 		_, err := e.MetaClient.CreateDatabase(stmt.Name)
 		return err
@@ -260,6 +266,12 @@ func (e *StatementExecutor) executeCreateDatabaseStatement(stmt *influxql.Create
 }
 
 func (e *StatementExecutor) executeCreateRetentionPolicyStatement(stmt *influxql.CreateRetentionPolicyStatement) error {
+	if !meta.ValidName(stmt.Name) {
+		// TODO This should probably be in `(*meta.Data).CreateRetentionPolicy`
+		// but can't go there until 1.1 is used everywhere
+		return meta.ErrInvalidName
+	}
+
 	spec := meta.RetentionPolicySpec{
 		Name:               stmt.Name,
 		Duration:           &stmt.Duration,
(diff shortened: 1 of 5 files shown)


coordinator/statement_executor.go
```diff
@@ -305,6 +305,10 @@ func (e *StatementExecutor) executeDropContinuousQueryStatement(q *influxql.Drop
 // It does not return an error if the database was not found on any of
 // the nodes, or in the Meta store.
 func (e *StatementExecutor) executeDropDatabaseStatement(stmt *influxql.DropDatabaseStatement) error {
+	if e.MetaClient.Database(stmt.Name) == nil {
```
jwilder: This needs to be a no-op instead of returning an error for API compatibility.

coordinator/statement_executor.go
```diff
@@ -348,6 +352,16 @@ func (e *StatementExecutor) executeDropShardStatement(stmt *influxql.DropShardSt
 }
 
 func (e *StatementExecutor) executeDropRetentionPolicyStatement(stmt *influxql.DropRetentionPolicyStatement) error {
+
+	dbi := e.MetaClient.Database(stmt.Database)
+	if dbi == nil {
+		return influxql.ErrDatabaseNotFound(stmt.Database)
```
jwilder: Same for both of these.

coordinator/statement_executor.go
joelegasse: Fixed. The conditional won't change, though.

PR #7830, opened 2017-01-12
Cache snapshotting performance improvements

jwilder:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

This PR has has a few changes to improve the cache snapshotting throughput.  Under heavy write load, snapshots can start to backup.  This can cause 500 to be returned with a `cache-max-memory-size` error as well as more frequent write timeouts due to the cache being lock for a longer period of time.  

The images below are examples of 1.1 cache and heap memory size under very heavy write load (~2M/s).  They were run a c4.8xlarge (36 core/ 64 GB RAM) with both 100k series and 1M series.  

The saw tooth pattern below is the snapshots getting progressively larger and taking longer to complete.  At the peaks, write errors for `cache-max-memory-size` being exceeded could be returned depending on the configuration of the limit.  When they get larger, timeouts also start to occur when compactions kick in and use up a CPU.

screen shot 2017-01-12 at 8 38 28 am

Heap also continues to grow in this case:
screen shot 2017-01-12 at 8 42 37 am

With this PR, the same write load now looks like:

screen shot 2017-01-12 at 8 45 11 am

screen shot 2017-01-12 at 8 45 41 am

A longer 10B point run across 1M series looks similar whereas 1.1 would start hitting 500s fairly quickly.

screen shot 2017-01-12 at 8 47 53 am

In both cases, the cache size stayed well under the default 1GB max memory size.  

The actual changes in this PR are the following:

* `Cache.Snapshot` was simplified to just swap the hot cache with the snapshot cache instead of iterating of very key and copy values.  This reduces lock contention with the cache is write locked.
* `CacheKeyIterator` encodes keys in parallel.  Previously, the writing of a snapshot was single thread and would iterate over each key in order and encode all the values.  This now encodes all the keys concurrently while the writer goroutine writes them out to disk in order as soon as the encoded blocks are ready.  The makes snapshotting leverage all available cores now.
* Fix cache size getting incremented to early.  If the cache limit was exceeded, the cache size counter would continue to grow with every failed write.  When the snapshot completed, it would sometimes go negative.  
* Fix an issue where compaction scheduling would sometime stop.  I'm not exactly sure how this happened locally, but the cache compactions for a shard just got stuck on a `Timer.C` channel that never fired.  This code was reusing a `Timer` incorrectly according to the docs so it has been switch to a `Ticker` instead.

With these changes, the chances of getting a `cache-max-memory-size` limit error decrease significantly.  Timeouts can still occur depending so increasing `write-timeout` may still be needed in some cases.  The timeout should occur much less frequently though.

### DIFF
--- a/tsdb/engine/tsm1/compact.go
+++ b/tsdb/engine/tsm1/compact.go
@@ -1320,14 +1320,12 @@ func NewCacheKeyIterator(cache *Cache, size int) KeyIterator {
 }
 
 func (c *cacheKeyIterator) encode() {
-	concurrency := runtime.NumCPU()
+	concurrency := runtime.GOMAXPROCS(0)
 	n := len(c.ready)
 
 	// Divide the keyset across each CPU
 	chunkSize := 128
 	idx := uint64(0)
-	var wg sync.WaitGroup
-	wg.Add(concurrency)
 	for i := 0; i < concurrency; i++ {
 		// Run one goroutine per CPU and encode a section of the key space concurrently
 		go func() {
@@ -1342,10 +1340,8 @@ func (c *cacheKeyIterator) encode() {
 				}
 				c.encodeRange(start, end)
 			}
-			wg.Done()
 		}()
 	}
-	wg.Wait()
 }
 
 func (c *cacheKeyIterator) encodeRange(start, stop int) {
@@ -1373,7 +1369,7 @@ func (c *cacheKeyIterator) encodeRange(start, stop int) {
 				err:     err,
 			})
 		}
-		// Notify this key is full encoded
+		// Notify this key is fully encoded
 		c.ready[i] <- struct{}{}
 	}
 }
@@ -1387,7 +1383,7 @@ func (c *cacheKeyIterator) Next() bool {
 	}
 	c.i++
 
-	if c.i >= len(c.order) {
+	if c.i >= len(c.ready) {
 		return false
 	}
 
--- a/tsdb/engine/tsm1/compact.go
+++ b/tsdb/engine/tsm1/compact.go
@@ -17,8 +17,10 @@ import (
 	"math"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/influxdata/influxdb/tsdb"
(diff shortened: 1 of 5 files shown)


tsdb/engine/tsm1/compact.go
```diff
@@ -1282,53 +1284,120 @@ func (k *tsmKeyIterator) Close() error {
...
+		order:  keys,
+		ready:  chans,
+		blocks: make([][]cacheBlock, len(keys)),
+	}
+	go cki.encode()
+	return cki
+}
+
+func (c *cacheKeyIterator) encode() {
+	concurrency := runtime.NumCPU()
```
mark-rushakoff: Shouldn't this be `runtime.GOMAXPROCS(0)` instead of `runtime.NumCPU`, to respect the `GOMAXPROCS` environment setting?

As a side note, there's a few other places in the tsm1 package where we're using `NumCPU`. I think those ought to all be `GOMAXPROCS`.

tsdb/engine/tsm1/compact.go
```diff
@@ -1282,53 +1284,120 @@ func (k *tsmKeyIterator) Close() error {
...
+				end := start + chunkSize
+				if end > n {
+					end = n
+				}
+				c.encodeRange(start, end)
+			}
+			wg.Done()
+		}()
+	}
+	wg.Wait()
```
mark-rushakoff: Because this method is only called in its own goroutine, the WaitGroup seems unnecessary.

tsdb/engine/tsm1/compact.go
jwilder: I used `NumCPU` because it sounds like `GOMAXPROCS` might go away at some point: https://golang.org/pkg/runtime/#GOMAXPROCS.  I also assumed that `NumCPU` is the same as `GOMAXPROCS(0)`.  That might be incorrect though.

tsdb/engine/tsm1/compact.go
jwilder: Good point.  Will update.

e-dard:
@jwilder I've added 73ed864

tsdb/engine/tsm1/compact.go
jwilder: Looks like `NumCPU` is different than `GOMAXPROCS(0)`.  Will update to use `GOMAXPROCS(0)`.

mark-rushakoff: If and when it does go away, [gofix](https://blog.golang.org/introducing-gofix) will be updated to minimize the effort to update our code.

> The GOMAXPROCS variable limits the number of operating system threads that can execute user-level Go code simultaneously.

There may be situations where someone would want to set it lower, possibly in conjunction with [taskset](http://linuxcommand.org/man_pages/taskset1.html) to deliberately limit how many CPUs influxdb can use. In other situations, we as developers may want to increase GOMAXPROCS to shake out contention issues.

Issue #7810, opened 2017-01-10
memory leak crashing influxdb


rybit:
I am seeing influxdb crash every day or two, because it consumes all the memory on the box. I took it from a 4GB to a 16GB and it just pushed how long it can go. Below is the chart. I find it odd that the Alloc and HeapAlloc are identical, but I looked [here ](https://github.com/influxdata/influxdb/blob/390a16925d8bce2955ef7a27bc423762566cd931/monitor/service.go#L301) and it seems right. Still the growth and eventual OOM is a problem. 

![image](https://cloud.githubusercontent.com/assets/406376/21790787/97acb126-d693-11e6-8c0d-74c9c8805904.png)

```
name: database
tags: database=metrics
numMeasurements	numSeries
---------------	---------
37		373362
```

influxdb version: InfluxDB v1.1.1 (git: master e47cf1f2e83a02443d7115c54f838be8ee959644)
linux version: ubuntu 14.04

```
[general]
bind-address = ":8088"

[meta]
dir = "/mnt/db/meta"

[data]
dir = "/mnt/db/data"
wal-dir = "/mnt/influx"

[admin]
enabled = true
port = ":8083"
https-enabled = true
https-certificate = "/etc/cfssl/certs/combined.pem"

[hinted-handoff]
dir = "/mnt/db/hh"

[http]
bind-address = ":8086"
https-enabled = true
https-certificate = "/etc/cfssl/certs/combined.pem"
auth-enabled = true
```

mark-rushakoff:
Many (most?) users are not experiencing this problem. If there's a memory leak, we want to fix it, but we need more detail to track down the source of the issue.

Please collect the following diagnostic information that was requested in the issue template. It will be most helpful if you gather this multiple times, each an hour or two apart, as you observe the memory growing.

```
influx -execute "show shards" > "shards-$(date +%s).txt"
influx -execute "show stats" > "stats-$(date +%s).txt"
influx -execute "show diagnostics" > "diagnostics-$(date +%s).txt"
curl -o "goroutine-$(date +%s).txt" "http://localhost:8086/debug/pprof/goroutine?debug=1" 
curl -o "heap-$(date +%s).txt" "http://localhost:8086/debug/pprof/heap?debug=1" 
```

rybit:
Yeah, I thought the behaviour was very strange. As requested here is [the gist with the results](https://gist.github.com/rybit/37606a70430eee2b2238c5aea7d6b2fc).

[...]

Issue #7778, opened 2016-12-30
fatal error "out of memory" sometimes (mostly around 3:00am if this helps)


Apollon77:
I use most current Influxdb package on Ubuntu Xenial on ARMHF device (Cubietruck).

Now for the third time after working some days the influxd crashes at night (around 3am german winter time) with a fatal (see attachement) ... System has 2 GB memory and DB is around 1,3GB.

Influxdb automatically restarts itself.

Hope the stack trace helps

[Uploading influxd-fatal.txt…]()

desa:
@Apollon77 What version of InfluxDB are you using? Also what kind of load is your system under?

desa:
The link doesn't currently point to the stack trace.

PR #7832, opened 2017-01-13
Fix memory leak of retained HTTP write payloads

### LINKED ISSUE #7827: Performance deterioration after upgrading to 1.1
### Bug report

InfluxDB 1.1, Centos7, hosted on EC2, c4.xlarge.
Write is done using HTTP.
Our current model includes about 30 databases with ~120 ops points written and 60 ops queries.
This is after limiting the number of connections to 300 because we had memory issues (We are reporting from about 450 servers once a minute).
Also we are using kapacitor that adds to the querying and writes and have Grafana querying the db.
Until a couple of days ago we were using 1.0 and decided to upgrade to 1.1 because of performance improvements.
Unfortunately, our performance seriously deteriorated:
image

I don't have the performance dumps, I'm afraid, because I had to downgrade as this was a production issue, but I do have the data in the _internal db so if you want me to query something let me know.

mark-rushakoff:
This leak seems to have been introduced in 8aa224b22dc007e,
present in 1.1.0 and 1.1.1.

When points were parsed from HTTP payloads, their tags and fields
referred to subslices of the request body; if any tag set introduced a
new series, then those tags then were stored in the in-memory series
index objects, preventing the HTTP body from being garbage collected. If
there were no new series in the payload, then the request body would be
garbage collected as usual.

Now, we clone the tags before we store them in the index. This is an
imperfect fix because the Point still holds references to the original
tags, and the Point's field iterator also refers to the payload buffer.
However, the current write code path does not retain references to the
Point or its fields; and this change will likely be obsoleted when TSI
is introduced.

This change likely fixes #7827, #7810, #7778, and perhaps others.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/models/points.go
+++ b/models/points.go
@@ -233,6 +233,9 @@ func ParsePointsString(buf string) ([]Point, error) {
 }
 
 // ParseKey returns the measurement name and tags from a point.
+//
+// NOTE: to minimize heap allocations, the returned Tags will refer to subslices of buf.
+// This can have the unintended effect preventing buf from being garbage collected.
 func ParseKey(buf []byte) (string, Tags, error) {
 	// Ignore the error because scanMeasurement returns "missing fields" which we ignore
 	// when just parsing a key
@@ -249,6 +252,9 @@ func ParseKey(buf []byte) (string, Tags, error) {
 
 // ParsePointsWithPrecision is similar to ParsePoints, but allows the
 // caller to provide a precision for time.
+//
+// NOTE: to minimize heap allocations, the returned Points will refer to subslices of buf.
+// This can have the unintended effect preventing buf from being garbage collected.
 func ParsePointsWithPrecision(buf []byte, defaultTime time.Time, precision string) ([]Point, error) {
 	points := make([]Point, 0, bytes.Count(buf, []byte{'\n'})+1)
 	var (
@@ -1617,6 +1623,22 @@ type Tag struct {
 	Value []byte
 }
 
+// Clone returns a shallow copy of Tag.
+//
+// Tags associated with a Point created by ParsePointsWithPrecision will hold references to the byte slice that was parsed.
+// Use Clone to create a Tag with new byte slices that do not refer to the argument to ParsePointsWithPrecision.
+func (t Tag) Clone() Tag {
+	other := Tag{
+		Key:   make([]byte, len(t.Key)),
+		Value: make([]byte, len(t.Value)),
+	}
+
+	copy(other.Key, t.Key)
+	copy(other.Value, t.Value)
+
+	return other
+}
+
 // Tags represents a sorted list of tags.
 type Tags []Tag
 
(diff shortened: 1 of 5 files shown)


mark-rushakoff:
If anyone is running into this issue now and needs a workaround until this is merged or released, the best option is to keep the HTTP payload as small as possible for writes that introduce a new series. Writes that only affect existing series should not contribute to the leak. Once new series have been written, I believe the only way to release the memory of the associated HTTP payloads is to restart the server.

joelegasse:
If we're concerned about total allocations, this will create a lot of smaller allocations. We could instead just perform one large allocations for cloning `Tags`, rather than cloning each tag Key/Value separately. It would be a two-pass `Clone()`, and there wouldn't be any `Clone()` defined for a single `Tag`, but I think given the usage and goal, that would be OK

mark-rushakoff:
@joelegasse I see what you mean. We have a couple options for a more efficient `Tags.Clone`:

* One backing slice containing all deduplicated keys and values
  * Advantage: one allocation
  * Disadvantage: that one allocation stays in memory until all referring series are deleted
* One backing slice for deduplicated keys, one backing slice for deduplicated values
  * Unlikely to be any better than the previous option
* One backing slice for deduplicated keys, one backing slice per deduplicated value:
  * Advantage: any individual value slice can be freed if all series containing that value are dropped
  * Disadvantage: potentially many small allocations (but potentially not that many, and likely still less than the implementation in cdbdd15)

Ultimately it probably would be best if `tsdb.NewSeries` was instead a method on `tsdb.DatabaseIndex` so that backing slices for tags' keys and values could all be shared. I'll look into this approach tomorrow, and if it seems too complex to get into 1.2, we can fall back to probably the third option above.

tsdb/shard.go
```diff
@@ -580,6 +580,10 @@ func (s *Shard) validateSeriesAndFields(points []models.Point) ([]models.Point,
 				continue
 			}
 
+			// If the tags were created by models.ParsePointsWithPrecision,
+			// they refer to subslices of the buffer containing line protocol.
+			// To ensure we don't refer to that buffer, preventing that buffer from being garbage collected, clone the tags.
+			tags = tags.Clone()
```
e-dard: This is going to cause unnecessary  allocations a lot of the time because `CreateSeriesIndexIfNotExists` only holds onto the subsliced tags if a new series needs to be created.

I think it would be better to push `tags` through to `NewSeries` as-is, and then clone the tags inside `CreateSeriesIndexIfNotExists` only when `AddSeries` is called. Maybe you could just do something like:

```go
series.Tags = series.Tags.Clone()
m.AddSeries(series)
```

models/points.go
```diff
@@ -1633,6 +1655,19 @@ func NewTags(m map[string]string) Tags {
 	return a
 }
 
+// Clone returns a shallow copy of Tags.
+//
+// Tags associated with a Point created by ParsePointsWithPrecision will hold references to the byte slice that was parsed.
+// Use Clone to create Tags with new byte slices that do not refer to the argument to ParsePointsWithPrecision.
+func (a Tags) Clone() Tags {
```
e-dard: Just a suggestion: `a` is not at risk of holding onto anything, if we remove all the references it contains. Therefore I don't think we need to allocate a new `Tags` value here. We can just do `a[i] = a[i].Clone()` and it will probably make this method significantly faster. 

Whether or not it's still appropriate to call this `Clone` with that implementation I'm not sure about. 😄

PR #7835, opened 2017-01-13
Bind RPC to localhost by default, add to sample config

mark-rushakoff:
Bind the RPC port to localhost by default so it is not potentially exposed to the public internet.

Add toplevel `bind-address` to the sample config because it was missing, and fix a couple typos in the config.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 The admin UI is removed and unusable in this release. The `[admin]` configuration section will be ignored.
 
+### Configuration Changes
+
+* The top-level config `bind-address` now defaults to `localhost:8088`.
+  The previous default was just `:8088`, causing the backup and restore port to be bound on all available interfaces (i.e. including interfaces on the public internet).
+
 ### Features
 
 - [#8143](https://github.com/influxdata/influxdb/pull/8143): Add WAL sync delay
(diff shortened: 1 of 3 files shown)


jwilder:
Actually, just realized this is the `8088` port and not the `8086` one.  This might be ok.

gunnaraasen:
We should also reevaluate whether we need to use a separate port `8088` at some point, since it only handles backup and restore (I think).

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ The stress tool `influx_stress` will be removed in a subsequent release. We reco
 - [#7585](https://github.com/influxdata/influxdb/pull/7585): Return Error instead of panic when decoding point values.
 - [#7812](https://github.com/influxdata/influxdb/issues/7812): Fix slice out of bounds panic when pruning shard groups. Thanks @vladlopes
 - [#7822](https://github.com/influxdata/influxdb/issues/7822): Drop database will delete /influxdb/data directory
+- [#7835](https://github.com/influxdata/influxdb/pull/7835): Bind backup and restore port to localhost by default
```
jwilder: Needs to be moved to 1.3 section now.

Also, I think we need to add a note under a configuration changes (like prior releases) that nodes the change in default.  Some people _may_ be relying on the current default for remote backups.

mark-rushakoff:
Finally updated the changelog and rebased, @jwilder.

Issue #7815, opened 2017-01-10
Mismatched type error in batch causes points to get registered but not written


sparrc:
### Bug report

When writing point batches that contain a mismatched type, some points get registered in the index but not written.

This appears to only happen for some points that appear in the batch _before_ the mismatched type point.

__System info:__ [Include InfluxDB version, operating system name, and other relevant details]

current master (commit 8c2cfd1)

__Steps to reproduce:__

1. Create database: `> create database telegraf`
2. Write `test` measurement:
```
$ curl -i "http://localhost:8086/write?db=telegraf" --data-binary "test value=1"
HTTP/1.1 204 No Content
Content-Type: application/json
Request-Id: e88b50c9-d74f-11e6-8009-000000000000
X-Influxdb-Version: v1.1.0
Date: Tue, 10 Jan 2017 16:14:48 GMT
```
3. Show measurements:
```
> show measurements
name: measurements
name
----
test
```
4. Write `test2` and `test3` measurements in a batch with a mismatched type `test` point at the end:
```
$ curl -i "http://localhost:8086/write?db=telegraf" --data-binary "test2 value=1                                                                
test3 value=1
test value=1i
"
HTTP/1.1 400 Bad Request
Content-Type: application/json
Request-Id: fc6afa4c-d74f-11e6-800b-000000000000
X-Influxdb-Version: v1.1.0
Date: Tue, 10 Jan 2017 16:15:21 GMT
Content-Length: 125

{"error":"field type conflict: input field \"value\" on measurement \"test\" is type integer, already exists as type float"}
```
5. Observe that `show measurements` now shows `test3`:
```
> show measurements
name: measurements
name
----
test
test3
```
6. But there is no data in `test3`, and it disappears on restart:
```
> select * from test3
>
```

__Additional info:__ [Include gist of relevant config, logs, etc.]

```
> show shards
name: _internal
id database  retention_policy shard_group start_time           end_time             expiry_time          owners
-- --------  ---------------- ----------- ----------           --------             -----------          ------
1  _internal monitor          1           2017-01-10T00:00:00Z 2017-01-11T00:00:00Z 2017-01-18T00:00:00Z


name: telegraf
id database retention_policy shard_group start_time           end_time             expiry_time          owners
-- -------- ---------------- ----------- ----------           --------             -----------          ------
3  telegraf autogen          3           2017-01-09T00:00:00Z 2017-01-16T00:00:00Z 2017-01-16T00:00:00Z

> show diagnostics
[...]

PR #7836, opened 2017-01-13
allow partial writes on field conflicts

### LINKED ISSUE #7814: InfluxDB should do a partial write on mismatched type errors
### Bug report

When a point is written with a mismatched type, it is handled differently than a malformed point. 

With a malformed point, a partial write is done of all well-formed points that are provided in the batch, and a "partial write" error is returned.

With a mismatched point, no points are written and a "mismatched type" error is returned.

both cases return http status code 400.

I think that we should treat mismatched types the same as malformed points, performing a partial write of the well-formed and valid points in the batch.

InfluxDB master, commit 8c2cfd1 (master some ways after 1.1.0)

__Steps to reproduce:__

1. write a metric called `test` as a float:
```
curl -i "http://localhost:8086/write?db=telegraf" --data-binary "test value=1"
```
2. write a batch of two metrics, first with `test` as an integer and then a new metric called `test2`:
```
curl -i "http://localhost:8086/write?db=telegraf" --data-binary "test value=1i
test2 value=1
"
```
3. Observe that the `test2` metric does not get created, even though it is valid.
4. Now write a batch of two metrics, first with `test` completely malformed and then a new metric called `test2`:
```
curl -i "http://localhost:8086/write?db=telegraf" --data-binary "test value
test2 value=1
"
```
5. Now observe that `test2` does get created, though a 400 status code and "partial write" is still returned.

Not sure, but this might be a dupe of #4856

corylanou:
fixes #7814 

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [ ] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): https://github.com/influxdata/docs.influxdata.com/issues/972

### DIFF
--- a/cmd/influxd/run/server_test.go
+++ b/cmd/influxd/run/server_test.go
@@ -387,6 +387,58 @@ func TestServer_UserCommands(t *testing.T) {
 	}
 }
 
+// Ensure the server will write all points possible with exception to the field type conflict.
+// This should return a partial write and a status of 400
+func TestServer_Write_FieldTypeConflict(t *testing.T) {
+	t.Parallel()
+	s := OpenServer(NewConfig())
+	defer s.Close()
+
+	if err := s.CreateDatabaseAndRetentionPolicy("db0", newRetentionPolicySpec("rp0", 1, 0), true); err != nil {
+		t.Fatal(err)
+	}
+
+	if res, err := s.Write("db0", "rp0", fmt.Sprintf("cpu value=1i %d", mustParseTime(time.RFC3339Nano, "2015-01-01T00:00:01Z").UnixNano()), nil); err != nil {
+		t.Fatal(err)
+	} else if exp := ``; exp != res {
+		t.Fatalf("unexpected results\nexp: %s\ngot: %s\n", exp, res)
+	}
+
+	// Verify the data was written.
+	if res, err := s.Query(`SELECT * FROM db0.rp0.cpu`); err != nil {
+		t.Fatal(err)
+	} else if exp := `{"results":[{"statement_id":0,"series":[{"name":"cpu","columns":["time","value"],"values":[["2015-01-01T00:00:01Z",1]]}]}]}`; exp != res {
+		t.Fatalf("unexpected results\nexp: %s\ngot: %s\n", exp, res)
+	}
+
+	writes := []string{
+		fmt.Sprintf("cpu value=2i %d", mustParseTime(time.RFC3339Nano, "2015-01-01T00:00:02Z").UnixNano()),
+		fmt.Sprintf("cpu value=3  %d", mustParseTime(time.RFC3339Nano, "2015-01-01T00:00:03Z").UnixNano()),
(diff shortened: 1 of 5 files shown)


coordinator/points_writer.go
```diff
@@ -353,6 +353,12 @@ func (w *PointsWriter) writeToShard(shard *meta.ShardInfo, database, retentionPo
 		return nil
 	}
 
+	// If this is a partial write error, that is also ok.
```
corylanou: @jwilder can you verify this is the correct thing to do here?  Otherwise it was re-trying which has no effect for a partial write.

tsdb/shard.go
```diff
@@ -543,7 +546,9 @@ func (s *Shard) validateSeriesAndFields(points []models.Point) ([]models.Point,
 
 	// get the shard mutex for locally defined fields
 	n = 0
+	var skipForFieldTypeConflict bool
 	for i, p := range points {
+		skipForFieldTypeConflict = false
```
corylanou: I hate this name.  @benbjohnson needs to pull out his uber naming skilz and help me!

coordinator/points_writer.go
jwilder: Shouldn't this increment write failed stat?

tsdb/shard.go
jwilder: what about just `skip`?

cmd/influxd/run/server_test.go
```diff
@@ -387,6 +387,52 @@ func TestServer_UserCommands(t *testing.T) {
 	}
 }
 
+// Ensure the server will write all points possible with exception to the field type conflict.
+// This should return a partial write and a status of 204
```
sparrc: partial write should return a status of `400`

sparrc:
this would also fix #7815

cmd/influxd/run/server_test.go
phemmer: Well, if we're going to debate the return code, shouldn't it be '409 conflict'?

phemmer:
Do we want to allow the possibility of somehow indicating which point caused the error? Maybe a request parameter that turns on a response body indicating the status of each point. Something like `[200,200,200,409,200]`. Or just indicating failed points, `{"3": 409}`

coordinator/points_writer.go
corylanou: yup, whoops. fixed.

services/httpd/handler.go
```diff
@@ -667,6 +667,7 @@ func (h *Handler) serveWrite(w http.ResponseWriter, r *http.Request, user *meta.
 		h.httpError(w, fmt.Sprintf("partial write: %v", werr), http.StatusBadRequest)
 		return
 	} else if err != nil {
+		log.Println("$$$$$$$$$$$$$$$$$$$$$$$$------ ", err.Error())
```
jwilder: Remove?

PR #7837, opened 2017-01-13
General tidy up and subtle bug fixes

e-dard:
- [x] Rebased/mergable
- [x] Tests pass

This PR uses `unused` and `staticcheck` to fix some subtle (mainly testing) bugs, as well as remove a bunch of dead code.


client/v2/client_test.go
```diff
@@ -257,16 +258,19 @@ func TestClient_Concurrent_Use(t *testing.T) {
 	wg.Add(3)
 	n := 1000
 
+	errC := make(chan error, 3)
```
joelegasse: Rather than counting the number of times this gets used, I would suggest using an unbuffered channel and launch a separate goroutine that waits on the WaitGroup and closes the channel. The last few lines would then be:

```go
go func() {
    wg.Wait()
    close(errC)
}

for e := range errC {
    if e != nil {
        t.Error(e)
    }
}
```

services/httpd/listen_test.go
```diff
@@ -82,15 +82,22 @@ func BenchmarkLimitListener(b *testing.B) {
...
 	for i := 0; i < b.N; i++ {
 		go func() {
 			c, err := l.Accept()
 			if err != nil {
-				b.Fatal(err)
+				errC <- err
+				return
 			}
 			c.Close()
 			wg.Done()
```
joelegasse: This feels like a bug... the WaitGroup will never finish if there is an error.

services/httpd/listen_test.go
```diff
@@ -82,15 +82,22 @@ func BenchmarkLimitListener(b *testing.B) {
 	wg.Add(b.N)
 
 	l := httpd.LimitListener(&fakeListener{}, b.N)
+	errC := make(chan error, 1)
```
joelegasse: Same comment as above: unbuffered channel, goroutine that waits on the WaitGroup and closes the channel, then use `range` and `t.Error()` instead of `Fatal`

tcp/mux_test.go
```diff
@@ -41,6 +42,8 @@ func TestMux(t *testing.T) {
 		if !testing.Verbose() {
 			mux.Logger = log.New(ioutil.Discard, "", 0)
 		}
+
+		errC := make(chan error, n)
```
joelegasse: Same comment. Unbuffered, goroutine, `range`, `t.Error`

tsdb/engine/tsm1/cache_race_test.go
```diff
@@ -85,23 +85,33 @@ func TestCacheRace(t *testing.T) {
 			c.Values(s)
 		}(s)
 	}
+
+	errC := make(chan error, 1)
```
joelegasse: Same.

tsdb/engine/tsm1/cache_race_test.go
```diff
@@ -138,18 +148,20 @@ func TestCacheRace2Compacters(t *testing.T) {
 	fileCounter := 0
 	mapFiles := map[int]bool{}
 	mu := sync.Mutex{}
+	errC := make(chan error, 1000)
```
joelegasse: Last comment, do the above things anywhere you find this stuff...

client/v2/client_test.go
e-dard: Looks like you commented just as I realised I had some redundant loops in the test handling code. The pattern instead of `for` should just be `if err := <-errC; err != nil` since you only need to find one error

e-dard: The problem with your approach is that `wg.Wait` will never get done because the second goroutine could be stuck trying to send on an unbuffered channel.

services/httpd/listen_test.go
e-dard: Yeah that does look like a bug.

client/v2/client_test.go
joelegasse: the `range` is executing concurrently with the `Wait`, and `t.Error` will allow the channel to drain

e-dard:
@joelegasse changes made. I don't know what app veyor is up to though.

PR #7853, opened 2017-01-17
Report subset of config values in SHOW DIAGNOSTICS

### LINKED ISSUE #7821: Expose some configuration values via SHOW DIAGNOSTICS
To reduce back-and-forth cycles between developers and users, we should include a subset of runtime configuration values in the output of `SHOW DIAGNOSTICS`. We already ask for diagnostics output in our issue template.

#### Rationale

* `SHOW DIAGNOSTICS` output will reflect any configuration set as environment variables (which is higher precedence than the config file)
* Diagnostics output is available without needing shell access to the machine running InfluxDB and without knowing where the config file resides on disk
* Diagnostics output will not contain large and numerous blocks of comments like most config files do, so it will be easier for developers to read

#### Example configuration settings likely to be included

A configuration setting is likely to be included if it will help us diagnose issues users are experiencing.

* `data.cache-snapshot-memory-size`
* `http.https-enabled`
* `udp.batch-timeout`

#### Example configuration settings likely to be excluded

A configuration setting will not be included if it is not helpful in troubleshooting or if exposing that value would make the system less secure.

* `data.query-log-enabled`: whether certain logs are enabled is usually not interesting to developers
* `http.shared-secret`: security-related, should not be exposed via any external tooling

mark-rushakoff:
Fixes #7821.

- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG updated

### DIFF
--- a/cmd/influxd/run/config.go
+++ b/cmd/influxd/run/config.go
@@ -17,6 +17,7 @@ import (
 	"github.com/BurntSushi/toml"
 	"github.com/influxdata/influxdb/coordinator"
 	"github.com/influxdata/influxdb/monitor"
+	"github.com/influxdata/influxdb/monitor/diagnostics"
 	"github.com/influxdata/influxdb/services/admin"
 	"github.com/influxdata/influxdb/services/collectd"
 	"github.com/influxdata/influxdb/services/continuous_querier"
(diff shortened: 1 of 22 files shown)


cmd/influxd/run/config.go
```diff
@@ -306,3 +307,60 @@ func (c *Config) applyEnvOverrides(prefix string, spec reflect.Value, structKey
...
+	return diagnostics.RowFromMap(map[string]interface{}{
+		"reporting-disabled": c.ReportingDisabled,
+		"bind-address":       c.BindAddress,
+	}), nil
+}
+
+func (c *Config) diagnosticsClients() map[string]diagnostics.Client {
+	// Config settings that are always present.
+	m := map[string]diagnostics.Client{
+		"config": c,
```
e-dard: If we're including `c` do we need to explicitly include all the other things? I'm a bit unclear on this.

services/collectd/config.go
```diff
@@ -128,3 +129,37 @@ func (c *Config) Validate() error {
...
+func (c Configs) Diagnostics() (*diagnostics.Diagnostics, error) {
+	d := &diagnostics.Diagnostics{
+		Columns: []string{"enabled", "bind-address", "database", "retention-policy", "batch-size", "batch-pending", "batch-timeout"},
+	}
+
+	for _, cc := range c {
+		if !cc.Enabled {
+			r := make([]interface{}, 1, len(d.Columns))
+			r[0] = false
+			d.AddRow(r)
```
e-dard: You could shorten this somewhat to just:

```go
d.AddRow([]interface{}{false})
```

services/graphite/config.go
```diff
@@ -253,3 +254,37 @@ func (c *Config) validateTag(keyValue string) error {
...
+func (c Configs) Diagnostics() (*diagnostics.Diagnostics, error) {
+	d := &diagnostics.Diagnostics{
+		Columns: []string{"enabled", "bind-address", "protocol", "database", "retention-policy", "batch-size", "batch-pending", "batch-timeout"},
+	}
+
+	for _, cc := range c {
+		if !cc.Enabled {
+			r := make([]interface{}, 1, len(d.Columns))
+			r[0] = false
+			d.AddRow(r)
```
e-dard: Same as above

cmd/influxd/run/config.go
mark-rushakoff: `c` implements `diagnostics.Client` directly with its own two, top-level config values `reporting-disabled` and `bind-address`, distinct from the other config sections.  The presence of the top-level / unscoped config settings at all is what makes this a little awkward.

Each key-value pair in the map is registered as a diagnostics client. This results in output like:

```
> show diagnostics
name: build
Branch  Build Time Commit  Version
------  ---------- ------  -------
unknown            unknown unknown


name: config
bind-address reporting-disabled
------------ ------------------
:8088        false


name: config-coordinator
log-queries-after max-concurrent-queries max-select-buckets max-select-point max-select-series query-timeout write-timeout
----------------- ---------------------- ------------------ ---------------- ----------------- ------------- -------------
0s                0                      0                  0                0                 0s            10s


name: config-cqs
enabled run-interval
------- ------------
true    1s

...
```

services/collectd/config.go
mark-rushakoff: I don't recall why I thought the capacity of the row had to match the length of the header row. Changing this and the other occurrences.

mark-rushakoff:
Rebased and removed commit message that noted the config information was included in /debug/vars output. I think that may have been true when the PR was originally opened but that is no longer the case after rebasing. I think it's fine if this is only available via SHOW DIAGNOSTICS.

mark-rushakoff:
Not merging this yet: before this change, `TestShard_WritePoints_FieldConflictConcurrentQuery` completes in about 3s on my machine; after, it takes about 99s! It doesn't look like this change should have affected this test but I'm going to continue digging into it.

mark-rushakoff:
Turns out the timing of that test was already sporadic on master as of fe5a5f38d7fafaa08260c30a5edaab571af5e909, prior to this commit, so I'm merging this PR. I'll still try and bisect to find where the test started getting inconsistent timing, but that's out of scope from the PR now.

```
$ time ./tsdb.test -test.run=ConcurrentQuery -test.cpuprofile /tmp/cq.cpuprof
PASS

real	0m0.486s
user	0m0.795s
sys	0m0.103s

$ time ./tsdb.test -test.run=ConcurrentQuery -test.cpuprofile /tmp/cq.cpuprof
PASS

real	1m20.424s
user	0m1.183s
sys	0m0.455s
```

PR #7854, opened 2017-01-17
Update subqueries so groupings are propagated to inner queries

### LINKED ISSUE #7846: (Possibly) inconsistent tag treatment with subqueries
### Bug report

Ubuntu 16.04, 1.3.0~n201701150800

__Steps to reproduce:__

1. Write some data:
Where `color` is a tag and `value` is a field.
```
> SELECT * FROM pillows
name: pillows
time                 color  value
----                 -----  -----
2017-01-16T19:27:10Z blue   2
2017-01-16T19:27:10Z yellow 12
2017-01-16T19:27:15Z blue   1
2017-01-16T19:27:15Z yellow 13
2017-01-16T19:27:20Z blue   3
2017-01-16T19:27:20Z yellow 15
2017-01-16T19:27:25Z blue   4
2017-01-16T19:27:25Z yellow 16
```
2. Run one query with a subquery and `GROUP BY *` in the main query:
```
> SELECT MEAN("last") FROM (SELECT LAST("value"),"color" FROM "pillows" WHERE time >= '2017-01-16T19:27:10Z' AND time <= '2017-01-16T19:27:25Z' GROUP BY time(10s)) GROUP BY *
name: pillows
time                 mean
----                 ----
1970-01-01T00:00:00Z 14.5
```
3.  Run another query with the same subquery and `GROUP BY` a specific tag key in the main query:
```
> SELECT MEAN("last") FROM (SELECT LAST("value"),"color" FROM "pillows" WHERE time >= '2017-01-16T19:27:10Z' AND time <= '2017-01-16T19:27:25Z' GROUP BY time(10s)) GROUP BY "color"
name: pillows
tags: color=blue
time                 mean
----                 ----
1970-01-01T00:00:00Z 2.5

name: pillows
tags: color=yellow
time                 mean
----                 ----
1970-01-01T00:00:00Z 14.5
```

__Expected behavior:__
Based on [this discussion](https://github.com/influxdata/influxdb/pull/7646#issuecomment-265547406) I would have expected InfluxDB to treat `color` as a field in both of the main queries. So I would have expected to get the result in the second step when I `GROUP BY *` and when I `GROUP BY color`.

__Actual behavior:__ 
I get different results depending on if I `GROUP BY *` or `GROUP BY color` in the main query. When I `GROUP BY *` InfluxDB can't find any tags to group by because all tags have been converted to fields in the subquery (I think). But, when I `GROUP BY color`, InfluxDB applies that clause to both the subquery and the main query so I end up with `color` being treated as a tag.

I don't love that tags get turned into fields but besides that, I don't find the behavior I described above to be consistent. I'm also just getting acquainted with subqueries so there could be an easy explanation for this 😬

jsternberg:
Previously, only time expressions got propagated inwards. The reason for
this was simple. If the outer query was going to filter to a specific
time range, then it would be unnecessary for the inner query to output
points within that time frame. It started as an optimization, but became
a feature because there was no reason to have the user repeat the same
time clause for the inner query as the outer query. So we allowed an
aggregate query with an interval to pass validation in the subquery if
the outer query had a time range. But `GROUP BY` clauses were not
propagated because that same logic didn't apply to them. It's not an
optimization there. So while grouping by a tag in the outer query
without grouping by it in the inner query was useless, there wasn't any
particular reason to care.

Then a bug was found where wildcards would propagate the dimensions
correctly, but the outer query containing a group by with the inner
query omitting it wouldn't correctly filter out the outer group by. We
could fix that filtering, but on further review, I had been seeing
people make that same mistake a lot. People seem to just believe that
the grouping should be propagated inwards. Instead of trying to fight
what the user wanted and explicitly erase groupings that weren't
propagated manually, we might as well just propagate them for the user
to make their lives easier. There is no useful situation where you would
want to group into buckets that can't physically exist so we might as
well do _something_ useful.

This will also now propagate time intervals to inner queries since the
same applies there. But, while the interval propagates, the following
query will not pass validation since it is still not possible to use a
grouping interval with a raw query (even if the inner query is an
aggregate):

    SELECT * FROM (SELECT mean(value) FROM cpu) WHERE time > now() - 5m GROUP BY time(1m)

This also means wildcards will behave a bit differently. They will
retrieve dimensions from the sources in the inner query rather than just
using the dimensions in the group by.

Fixes #7846.

### DIFF
--- a/cmd/influxd/run/server_test.go
+++ b/cmd/influxd/run/server_test.go
@@ -4653,6 +4653,11 @@ func TestServer_Query_Subqueries(t *testing.T) {
 			command: `SELECT min(value) FROM (SELECT max(usage_user), usage_user - usage_system AS value FROM cpu GROUP BY host) WHERE time >= '2000-01-01T00:00:00Z' AND time < '2000-01-01T00:00:30Z'`,
 			exp:     `{"results":[{"statement_id":0,"series":[{"name":"cpu","columns":["time","min"],"values":[["2000-01-01T00:00:10Z",-44]]}]}]}`,
 		},
+		&Query{
+			params:  url.Values{"db": []string{"db0"}},
+			command: `SELECT min(value) FROM (SELECT top(usage_user, 2), usage_user - usage_system AS value FROM cpu) WHERE time >= '2000-01-01T00:00:00Z' AND time < '2000-01-01T00:00:30Z' GROUP BY host`,
+			exp:     `{"results":[{"statement_id":0,"series":[{"name":"cpu","tags":{"host":"server01"},"columns":["time","min"],"values":[["2000-01-01T00:00:10Z",-10]]},{"name":"cpu","tags":{"host":"server02"},"columns":["time","min"],"values":[["2000-01-01T00:00:10Z",-44]]}]}]}`,
+		},
 	}...)
 
 	for i, query := range test.queries {
--- a/influxql/ast.go
+++ b/influxql/ast.go
@@ -1922,7 +1922,7 @@ func (s *SelectStatement) validateAggregates(tr targetRequirement) error {
 					return fmt.Errorf("invalid group interval: %v", err)
 				}
 
-				if c, ok := expr.Args[0].(*Call); ok && groupByInterval == 0 {
+				if c, ok := expr.Args[0].(*Call); ok && groupByInterval == 0 && tr != targetSubquery {
 					return fmt.Errorf("%s aggregate requires a GROUP BY interval", expr.Name)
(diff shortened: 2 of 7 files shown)


jsternberg:
@jwilder @desa @rkuchan

corylanou:
Should we have some integration tests that illustrate the bug and the fixes and prove that they are working properly with actual data?  I feel that any change to the query engine needs to have integration tests proving out the changes.

Also, we need to check benchmarking as well prior to this PR being accepted.  Same line of thinking, any time we make a change that affects the query engine, we should benchmark as well.

PR #7862, opened 2017-01-22
Adds handler for returning a profile archive

e-dard:
- [x] Rebased/mergable
- [ ] Tests pass
- [x] CHANGELOG.md updated

Either @gunnaraasen or @rossmcdonald sowed the seed of this idea, so thanks to whichever one of you it was. 👍 

Currently, when debugging issues with InfluxDB, we often ask for the
following profiles and queries:

```
curl -o block.txt "http://localhost:8086/debug/pprof/block?debug=1"
curl -o goroutine.txt "http://localhost:8086/debug/pprof/goroutine?debug=1"
curl -o heap.txt "http://localhost:8086/debug/pprof/heap?debug=1"
curl -o cpu.txt "http://localhost:8086/debug/pprof/profile

influx -exectute 'SHOW SHARDS' > shards.txt
influx -exectute 'SHOW STATS' > stats.txt
influx -exectute 'SHOW DIAGNOSTICS' > diagnostics.txt
```

Collecting these can be bothersome for users, or even difficult if they're unfamiliar with `cURL` (or it's not on their system).

This commit adds a new endpoint: `/debug/pprof/all` which will return a single compressed archive of all of the above profiles and queries. The CPU profile is optional, and not returned by default. To include a CPU profile, the URL to request should be: `/debug/pprof/all?cpu=true`. It's also possible to vary the length of the CPU profile by adding a `seconds=x` parameter, where `x` defaults to `30`, if absent.

The new command for gathering profiles from users should now be:

```
curl -o profiles.tar.gz "http://localhost:8086/debug/pprof/all"
```

Or, if we need to see a CPU profile:

```
```

It's important to remember that a CPU profile is a blocking operation, and by default it will take 30 seconds for the response (archive) to be returned to the user.

Finally, if the user is unfamiliar with `cURL`, they will now be able to visit `http://localhost:8086/debug/pprof/all` in a web browser, and the archive will be downloaded to their machine.

The archive contents look like:

```
⇒  tree ~/Downloads/profiles
/Users/edd/Downloads/profiles
├── block.txt
├── cpu.txt
├── diagnostics.txt
├── goroutine.txt
├── heap.txt
├── shards.txt
└── stats.txt

0 directories, 7 files
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,11 +22,12 @@ The admin UI is removed and unusable in this release. The `[admin]` configuratio
 - [#8302](https://github.com/influxdata/influxdb/pull/8302): Write throughput/concurrency improvements
 - [#8273](https://github.com/influxdata/influxdb/issues/8273): Remove the admin UI.
 - [#8327](https://github.com/influxdata/influxdb/pull/8327): Update to go1.8.1
-- [#8348](https://github.com/influxd	ata/influxdb/pull/8348): Add max concurrent compaction limits
+- [#8348](https://github.com/influxdata/influxdb/pull/8348): Add max concurrent compaction limits
 - [#8366](https://github.com/influxdata/influxdb/pull/8366): Add TSI support tooling.
 - [#8350](https://github.com/influxdata/influxdb/pull/8350): Track HTTP client requests for /write and /query with /debug/requests.
+<<<<<<< 8f8ff0ec612e6e77ff618f572b302e069de8e5c3
 - [#8384](https://github.com/influxdata/influxdb/pull/8384): Write and compaction stability
-- [#7862](https://github.com/influxdata/influxdb/pull/7861): Add new profile endpoint for gathering all debug profiles single in archive.
+- [#7862](https://github.com/influxdata/influxdb/pull/7861): Add new profile endpoint for gathering all debug profiles and querues in single archive.
 
 ### Bugfixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,10 +22,11 @@ The admin UI is removed and unusable in this release. The `[admin]` configuratio
 - [#8302](https://github.com/influxdata/influxdb/pull/8302): Write throughput/concurrency improvements
 - [#8273](https://github.com/influxdata/influxdb/issues/8273): Remove the admin UI.
 - [#8327](https://github.com/influxdata/influxdb/pull/8327): Update to go1.8.1
-- [#8348](https://github.com/influxdata/influxdb/pull/8348): Add max concurrent compaction limits
+- [#8348](https://github.com/influxd	ata/influxdb/pull/8348): Add max concurrent compaction limits
 - [#8366](https://github.com/influxdata/influxdb/pull/8366): Add TSI support tooling.
 - [#8350](https://github.com/influxdata/influxdb/pull/8350): Track HTTP client requests for /write and /query with /debug/requests.
 - [#8384](https://github.com/influxdata/influxdb/pull/8384): Write and compaction stability
+- [#7862](https://github.com/influxdata/influxdb/pull/7861): Add new profile endpoint for gathering all debug profiles single in archive.
 
 ### Bugfixes
 
--- a/services/httpd/pprof.go
+++ b/services/httpd/pprof.go
@@ -1,0 +1,152 @@
+package httpd
+
+import (
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
+	"io"
+	"net/http"
+	httppprof "net/http/pprof"
+	"runtime/pprof"
+	"strconv"
+	"time"
+)
+
+// handleProfiles determines which profile to return to the requester.
+func handleProfiles(w http.ResponseWriter, r *http.Request) {
+	switch r.URL.Path {
+	case "/debug/pprof/cmdline":
+		httppprof.Cmdline(w, r)
+	case "/debug/pprof/profile":
+		httppprof.Profile(w, r)
+	case "/debug/pprof/symbol":
+		httppprof.Symbol(w, r)
+	case "/debug/pprof/all":
+		archiveProfiles(w, r)
+	default:
+		httppprof.Index(w, r)
+	}
+}
+
+// prof describes a profile name and a debug value, or in the case of a CPU
+// profile, the number of seconds to collect the profile for.
+type prof struct {
+	Name  string
+	Debug int64
+}
+
+// archiveProfiles collects the following profiles:
+//	- goroutine profile
+//	- heap profile
+//	- blocking profile
+//	- (optionally) CPU profile
+//
+// All profiles are added to a tar archive and then compressed, before being
+// returned to the requester as an archive file. Where profiles support debug
+// parameters, the profile is collected with debug=1. To optionally include a
+// CPU profile, the requester should provide a `cpu` query parameter, and can
+// also provide a `seconds` parameter to specify a non-default profile
+// collection time. The default CPU profile collection time is 30 seconds.
+//
+// Example request including CPU profile:
+//
+//	http://localhost:8086/debug/pprof/all?cpu=true&seconds=45
+//
+// The value after the `cpu` query parameter is not actually important, as long
+// as there is something there.
+//
+func archiveProfiles(w http.ResponseWriter, r *http.Request) {
+	var all = []*prof{
+		{Name: "goroutine", Debug: 1},
+		{Name: "block", Debug: 1},
+		{Name: "heap", Debug: 1},
+	}
+
+	// Capture a CPU profile?
+	if r.FormValue("cpu") != "" {
+		profile := &prof{Name: "cpu"}
+
+		// For a CPU profile we'll use the Debug field to indicate the number of
+		// seconds to capture the profile for.
+		profile.Debug, _ = strconv.ParseInt(r.FormValue("seconds"), 10, 64)
+		if profile.Debug <= 0 {
+			profile.Debug = 30
+		}
+		all = append([]*prof{profile}, all...) // CPU profile first.
+	}
+
+	var (
+		resp bytes.Buffer // Temporary buffer for entire archive.
+		buf  bytes.Buffer // Temporary buffer for each profile.
+	)
+
+	gz := gzip.NewWriter(&resp)
+	tw := tar.NewWriter(gz)
+	for _, profile := range all {
+		if profile.Name == "cpu" {
+			if err := pprof.StartCPUProfile(&buf); err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+
+			sleep(w, time.Duration(profile.Debug)*time.Second)
+			pprof.StopCPUProfile()
+		} else {
+			prof := pprof.Lookup(profile.Name)
+			if prof == nil {
+				http.Error(w, "unable to find profile "+profile.Name, http.StatusInternalServerError)
+				return
+			}
+
+			if err := prof.WriteTo(&buf, int(profile.Debug)); err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+		}
+
+		// Write the profile file's header.
+		err := tw.WriteHeader(&tar.Header{
+			Name: profile.Name + ".txt",
+			Mode: 0600,
+			Size: int64(buf.Len()),
+		})
+		if err != nil {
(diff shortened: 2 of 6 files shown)


services/httpd/pprof.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// Close the gzip writer.
+	if err := gz.Close(); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+	}
+
+	// Return the gzipped archive.
+	w.Header().Set("Content-Disposition", "attachment; filename=profiles.tar.gz")
+	w.Header().Set("Content-Type", "application/gzip")
+	fmt.Fprint(w, resp.String())
```
mark-rushakoff: Nit: This looks like it would allocate a new string for the entire response. Would this be better as `io.Copy(w, resp)`?

services/httpd/pprof.go
```diff
@@ -0,0 +1,151 @@
...
+// Example request including CPU profile:
+//
+//	http://localhost:8086/debug/pprof/all?cpu=true&seconds=45
+//
+// The value after the `cpu` query parameter is not actually important, as long
+// as there is something there.
+//
+func archiveProfiles(w http.ResponseWriter, r *http.Request) {
+	var all = []*prof{
+		{Name: "goroutine", Debug: 1},
```
mark-rushakoff: The `goroutine` profile accepts debug=2, but I think we're always more interested in debug=1 output, since it combines goroutines with identical stack traces. Maybe add a comment noting that we deliberately aren't supporting debug=2?

e-dard: `debug=2` is not a valid format for use with the `pprof` tool, so it's not as useful in my opinion. I'll add something to make it explicit though.

e-dard:
@jwilder I did consider adding other information but it didn't feel like the right place for it, since it's under the `/debug/pprof` endpoint. Plus executing queries could be a lot more expensive than gathering otherwise helpful profiles.

Maybe we should add a further endpoint, which will execute the queries and also call out to the `/debug/pprof` endpoint, then wrap everything up into an archive? We can direct users to run either one dependent on their circumstances?

jwilder:
@e-dard Those three queries are returned directly from the meta cache.

e-dard:
@jwilder cool. I'll add them.

services/httpd/pprof.go
mark-rushakoff: https://golang.org/pkg/runtime/pprof/

> The predefined profiles may assign meaning to other debug values; for example, when printing the "goroutine" profile, debug=2 means to print the goroutine stacks in the same form that a Go program uses when dying due to an unrecovered panic.

```
$ curl -s http://localhost:8086/debug/pprof/goroutine?debug=0 | head
goroutine profile: total 55
5 @ 0x300ba 0x3f4c9 0x3e2bc 0x1de964 0x5ffb1
5 @ 0x300ba 0x3f4c9 0x3e2bc 0x477d16 0x4b1241 0x5ffb1
5 @ 0x300ba 0x3f4c9 0x3e2bc 0x477f72 0x4b1073 0x5ffb1
5 @ 0x300ba 0x3f4c9 0x3e2bc 0x477f72 0x4b1113 0x5ffb1
5 @ 0x300ba 0x3f4c9 0x3e2bc 0x477f72 0x4b11b3 0x5ffb1
5 @ 0x300ba 0x3f4c9 0x3e2bc 0x4780f5 0x4b0fd1 0x5ffb1
5 @ 0x300ba 0x3f4c9 0x3e2bc 0x4a3da7 0x5ffb1
1 @ 0x1205b 0x43387 0x844d2 0x5ffb1
1 @ 0x23710f 0x236f10 0x233c31 0x425876 0x425bf5 0x19fd56 0x360fdd 0x35d987 0x5ffb1

$  curl -s http://localhost:8086/debug/pprof/goroutine?debug=1 | head
goroutine profile: total 55
5 @ 0x300ba 0x3f4c9 0x3e2bc 0x1de964 0x5ffb1
#	0x1de963	github.com/influxdata/influxdb/tsdb.(*Shard).monitor+0x3b3	/Users/mr/go/src/github.com/influxdata/influxdb/tsdb/shard.go:918

5 @ 0x300ba 0x3f4c9 0x3e2bc 0x477d16 0x4b1241 0x5ffb1
#	0x477d15	github.com/influxdata/influxdb/tsdb/engine/tsm1.(*Engine).compactCache+0x475			/Users/mr/go/src/github.com/influxdata/influxdb/tsdb/engine/tsm1/engine.go:963
#	0x4b1240	github.com/influxdata/influxdb/tsdb/engine/tsm1.(*Engine).enableSnapshotCompactions.func1+0x60	/Users/mr/go/src/github.com/influxdata/influxdb/tsdb/engine/tsm1/engine.go:266

5 @ 0x300ba 0x3f4c9 0x3e2bc 0x477f72 0x4b1073 0x5ffb1
#	0x477f71	github.com/influxdata/influxdb/tsdb/engine/tsm1.(*Engine).compactTSMLevel+0x161			/Users/mr/go/src/github.com/influxdata/influxdb/tsdb/engine/tsm1/engine.go:1003

$ curl -s http://localhost:8086/debug/pprof/goroutine?debug=2 | head
goroutine 102 [running]:
runtime/pprof.writeGoroutineStacks(0xa46a20, 0xc420f7f040, 0xa68940, 0x30)
	/usr/local/Cellar/go/1.7.3/libexec/src/runtime/pprof/pprof.go:585 +0x79
runtime/pprof.writeGoroutine(0xa46a20, 0xc420f7f040, 0x2, 0x0, 0xc420132000)
	/usr/local/Cellar/go/1.7.3/libexec/src/runtime/pprof/pprof.go:574 +0x44
runtime/pprof.(*Profile).WriteTo(0xa6a360, 0xa46a20, 0xc420f7f040, 0x2, 0xc420f7f040, 0xc4207af9c0)
	/usr/local/Cellar/go/1.7.3/libexec/src/runtime/pprof/pprof.go:298 +0x341
net/http/pprof.handler.ServeHTTP(0xc420b163d1, 0x9, 0xa4f0a0, 0xc420f7f040, 0xc420f81950)
	/usr/local/Cellar/go/1.7.3/libexec/src/net/http/pprof/pprof.go:209 +0x1a6
net/http/pprof.Index(0xa4f0a0, 0xc420f7f040, 0xc420f81950)
```

e-dard: @mark-rushakoff I think you misunderstood what I meant. `debug=2` is not compatible with `go tool pprof`, though it is of course valid with the `runtime/pprof` package handler, and will output something useful (to a human). It's just you can't feed that profile into the `pprof` tool.

```
edd@Alameda:~|
⇒  go tool pprof $GOPATH/bin/ossinfluxd-v1.2.1 http://localhost:8086/debug/pprof/goroutine\?debug\=1
Fetching profile from http://localhost:8086/debug/pprof/goroutine?debug=1
Saved profile in /Users/edd/pprof/pprof.ossinfluxd-v1.2.1.localhost:8086.goroutine.001.pb.gz
Entering interactive mode (type "help" for commands)
(pprof) quit
edd@Alameda:~|
⇒  go tool pprof $GOPATH/bin/ossinfluxd-v1.2.1 http://localhost:8086/debug/pprof/goroutine\?debug\=2
Fetching profile from http://localhost:8086/debug/pprof/goroutine?debug=2
parsing profile: unrecognized profile format
edd@Alameda:~|
⇒  go tool pprof $GOPATH/bin/ossinfluxd-v1.2.1 http://localhost:8086/debug/pprof/goroutine\?debug\=0
Fetching profile from http://localhost:8086/debug/pprof/goroutine?debug=0
Saved profile in /Users/edd/pprof/pprof.ossinfluxd-v1.2.1.localhost:8086.goroutine.002.pb.gz
Entering interactive mode (type "help" for commands)
(pprof) quit
edd@Alameda:~|
⇒
```

e-dard:
@jwilder I was going to fix this up the other day but then I realised we need to be able to handle meta queries where the server has authentication enabled. We can chat about that offline.

e-dard:
@jwilder @mark-rushakoff see https://github.com/influxdata/influxdb/pull/7862/commits/d692e4730282c45c08b895da0923143884827623. I've added support for `SHOW SHARDS`, `SHOW STATS` and `SHOW DIAGNOSTICS`.

CHANGELOG.md
```diff
@@ -22,10 +22,10 @@ The admin UI is removed and unusable in this release. The `[admin]` configuratio
 - [#8302](https://github.com/influxdata/influxdb/pull/8302): Write throughput/concurrency improvements
 - [#8273](https://github.com/influxdata/influxdb/issues/8273): Remove the admin UI.
 - [#8327](https://github.com/influxdata/influxdb/pull/8327): Update to go1.8.1
-- [#8348](https://github.com/influxd	ata/influxdb/pull/8348): Add max concurrent compaction limits
+- [#8348](https://github.com/influxdata/influxdb/pull/8348): Add max concurrent compaction limits
 - [#8366](https://github.com/influxdata/influxdb/pull/8366): Add TSI support tooling.
 - [#8350](https://github.com/influxdata/influxdb/pull/8350): Track HTTP client requests for /write and /query with /debug/requests.
-- [#7862](https://github.com/influxdata/influxdb/pull/7861): Add new profile endpoint for gathering all debug profiles single in archive.
+- [#7862](https://github.com/influxdata/influxdb/pull/7861): Add new profile endpoint for gathering all debug profiles and querues in single archive.
```
mark-rushakoff: s/querues/queries

services/httpd/pprof.go
```diff
@@ -123,6 +136,76 @@ func archiveProfiles(w http.ResponseWriter, r *http.Request) {
...
+		rows, err := query.fn()
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+		}
+
+		for i, row := range rows {
+			var out []byte
+			// Write the columns
+			for _, col := range row.Columns {
+				out = append(out, []byte(col+"\t")...)
```
mark-rushakoff: Should be `out = append(out, col...); out = append(out, '\t')`?

services/httpd/pprof.go
```diff
@@ -123,6 +136,76 @@ func archiveProfiles(w http.ResponseWriter, r *http.Request) {
...
+			}
+			out = append(out, []byte("\n")...)
```
mark-rushakoff: `out = append(out, '\n')`?

services/httpd/pprof.go
```diff
@@ -123,6 +136,76 @@ func archiveProfiles(w http.ResponseWriter, r *http.Request) {
...
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+			}
+
+			// Write all the values
+			for _, val := range row.Values {
+				out = out[:0]
+				for _, v := range val {
+					out = append(out, []byte(fmt.Sprintf("%v\t", v))...)
+				}
+				out = append(out, []byte("\n")...)
```
mark-rushakoff: `out = append(out, '\n')`?

services/httpd/pprof.go
```diff
@@ -139,6 +222,107 @@ func archiveProfiles(w http.ResponseWriter, r *http.Request) {
...
+			values = append(values, stat.Values[k])
+		}
+		row.Values = [][]interface{}{values}
+		rows = append(rows, row)
+	}
+	return rows, nil
+}
+
+// joinUint64 returns a comma-delimited string of uint64 numbers.
+func joinUint64(a []uint64) string {
```
mark-rushakoff: `strconv.FormatUint` will create a new string and then write it to the buffer. I would expect this method to be more efficient as

```go
func joinUint64(a []uint64) string {
  var buf []byte // Could take a guess at initial size here.
  for i, x := range a {
    if i != 0 {
      buf = append(buf, ',')
    }
    buf = strconv.AppendUint(buf, x, 10)
  }
  return string(buf)
}
```

services/httpd/pprof.go
e-dard: `col` is a string not a `[]byte`. It looks better as far as I can tell when we have a tab between each column in the output (including between each column header).

services/httpd/pprof.go
e-dard: I always forget this...

services/httpd/pprof.go
e-dard: I pilfered this from https://github.com/influxdata/influxdb/blob/master/coordinator/statement_executor.go#L1182; I didn't look at the complexity at all since this is not a hot path. I don't mind changing it for that implementation.

PR #7863, opened 2017-01-22
Add system information to /debug/vars

### LINKED ISSUE #7776: [feature request] Add database uptime to /debug/vars endpoint
### Feature Request

__Proposal:__

Currently the database uptime is included in the output to `SHOW DIAGNOSTICS`, but is not included in the `/debug/vars` endpoint output.

__Current behavior:__

Uptime is not included in the `/debug/vars` output.

__Desired behavior:__

Uptime will be included in the `/debug/vars` output. Here are the vars from SHOW DIAGNOSTICS:

```
name: system
PID  currentTime                    started                        uptime
---  -----------                    -------                        ------
8850 2017-01-20T00:30:30.417981478Z 2017-01-20T00:30:21.130836519Z 9.287145055s
```

__Use case:__ 

If uptime was exposed, Kapacitor could be used to monitor when the InfluxDB process crashes or restarts.

e-dard:
- [x] Rebased/mergable
- [x] CHANGELOG.md updated

It should fix #7776. The implementation is not very elegant, as `diagnostics.Diagnostics` is not in a format that lends itself well to being partially converted into JSON. Possibly a better solution would be to add a `json.Marshaler` implementation to each `diagnostics.Diagnostics`.

Hitting up `/debug/vars` will now should something like:

```json
{
  "system": {
    "currentTime": "2017-01-22T04:02:39.752460895Z",
    "started": "2017-01-22T04:02:38.744828465Z",
    "uptime": 1007632562
  },
  "cmdline": [
    "influxd"
  ],
....
```

The uptime of the system is shown in seconds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,17 @@
+## v1.3.0 [unreleased]
+
+### Features
+
+- [#7776](https://github.com/influxdata/influxdb/issues/7776): Add system information to /debug/vars.
+
+
 ## v1.2.1 [unreleased]
 
 ### Bugfixes
 
 - [#7877](https://github.com/influxdata/influxdb/issues/7877): Fix mapping of types when the measurement uses a regex
 - [#7888](https://github.com/influxdata/influxdb/pull/7888): Expand query dimensions from the subquery.
+
 
 ## v1.2.0 [2017-01-24]
 
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -24,6 +24,7 @@ import (
 	"github.com/influxdata/influxdb/influxql"
 	"github.com/influxdata/influxdb/models"
 	"github.com/influxdata/influxdb/monitor"
+	"github.com/influxdata/influxdb/monitor/diagnostics"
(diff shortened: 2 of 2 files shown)


rossmcdonald:
@e-dard It looks like uptime is in ns. Do you think that could cause issues for very long-running instances? I'm wondering if we should convert it to seconds instead.

Also, if we keep it as nanoseconds, we should probably change the key to `uptime_ns` to make it clear what the unit is.

e-dard:
@rossmcdonald it will cause a problem if someone has their instance up for longer than 292 years. You're right about clarity though. I'll change it to seconds.

e-dard:
@rossmcdonald updated to seconds.

rossmcdonald:
@e-dard Looks great, thanks for your help.

rossmcdonald:
@e-dard Do you need anyone else to review this before merging?

e-dard:
@rossmcdonald I'll get this rebased and merged.

e-dard:
@corylanou yep I agree. I have manually tested this for now. @rossmcdonald if you could test this on a `master` build that would be great too.

PR #7881, opened 2017-01-25
Fix mapping of types when the measurement uses a regex

### LINKED ISSUE #7877: Incorrect results when select uses measurement regex without field regex
### Bug report

__System info:__ [1.1.2, MacOs, 10.12.3] 

__Steps to reproduce:__

Two measurements, M1 and M2 with same fields: A and B. These are fields, not tags. 

1. select "A","B" from /M1|M2/ =>empty
2. select A,/B/ from /M1|M2/ => correct result for B
3. select /A/,/B/ from /M1|M2/ =>correct result A,B

__Expected behavior:__ [the value of the fields "A" and "B" are correctly selected in all queries]

__Actual behavior:__ [it appears that regex in field expressions TRIGGERS parsing regex in measurement expression]

__Additional info:__ [the query 1 works fine with influx 1.1.0-1 on Windows and MacOs] This new behavior in influxdb 1.2.0-1 will break Grafana.

jsternberg:
With the new shard mapper implementation, regexes were just ignored so
it attempted to look up the field type inside of a measurement with no
name (which cannot possibly exist) so it would think the field didn't
exist and map it as the unknown type.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

Fixes #7877.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## v1.2.1 [unreleased]
+
+### Bugfixes
+
+- [#7877](https://github.com/influxdata/influxdb/issues/7877): Fix mapping of types when the measurement uses a regex
+
 ## v1.2.0 [2017-01-24]
 
 ### Release Notes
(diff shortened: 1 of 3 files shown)


pauldix:
LGTM

PR #7907, opened 2017-01-30
Fix ORDER BY time DESC with ordering series keys

### LINKED ISSUE #7905: ORDER BY time DESC probably returns the wrong results sometimes
### Bug report

__System info:__ InfluxDB 1.2.0

__Steps to reproduce:__

```
> CREATE DATABASE db0
> USE db0
> INSERT cpu,host=server01 value=2 1000000000
> INSERT cpu,host=server01 value=3 2000000000
> INSERT cpu,host=server02 value=4 1000000000
> INSERT cpu,host=server02 value=5 2000000000
> SELECT value, host FROM cpu ORDER BY time DESC
name: cpu
time                value host
----                ----- ----
2000000000          3     server01
2000000000          5     server02
1000000000          2     server01
1000000000          4     server02
```

The order should be `server02` before `server01`. There is also an error in the Emitter which causes this one:

```
> CREATE DATABASE db0
> USE db0
> INSERT mem,host=server01 free=1 1000000000
> INSERT mem,host=server01 free=2 2000000000
> INSERT mem,host=server02 used=3 1000000000
> INSERT mem,host=server02 used=4 2000000000
> SELECT first("used") AS "used", first("free") AS "free" FROM "mem" WHERE time >= '1970-01-01T00:00:01Z' AND time <= '1970-01-01T00:00:02Z' GROUP BY host, time(1s) FILL(none) ORDER BY time DESC
name: mem
tags: host=server01
time       used free
----       ---- ----
2000000000      2
1000000000      1

name: mem
tags: host=server02
time       used free
----       ---- ----
2000000000 4
1000000000 3
```

`server02` is supposed to be before `server01`.

__Additional info:__

This section of code seems wrong and an earlier version of the code seems to do something different:

```
        // Update range values if lower and emitter is in time ascending order.
        if e.ascending {
            if (itrName < name) || (itrName == name && itrTags.ID() < tags.ID()) || (itrName == name && itrTags.ID() == tags.ID() && itrTime < t) {
                t, name, tags = itrTime, itrName, itrTags
            }
            continue
        }

        // Update range values if higher and emitter is in time descending order.
        if (itrName < name) || (itrName == name && itrTags.ID() < tags.ID()) || (itrName == name && itrTags.ID() == tags.ID() && itrTime < t) {
            t, name, tags = itrTime, itrName, itrTags
        }
```

It seems like the second statement there have the conditions in the wrong direction.

jsternberg:
The order of series keys is in ascending alphabetical order, not
descending alphabetical order, when it is ordered by descending time.
This fixes the ordering so points are returned in descending order. The
emitter also had the conditions for choosing which iterator to use in
the wrong direction (which only affects aggregates with `FILL(none)`).

Fixes #7905.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - [#7895](https://github.com/influxdata/influxdb/issues/7895): Fix incorrect math when aggregates that emit different times are used.
 - [#7946](https://github.com/influxdata/influxdb/issues/7946): Fix authentication when subqueries are present.
 - [#7885](https://github.com/influxdata/influxdb/issues/7885): Fix LIMIT and OFFSET when they are used in a subquery.
+- [#7905](https://github.com/influxdata/influxdb/issues/7905): Fix ORDER BY time DESC with ordering series keys.
 
 ## v1.2.0 [2017-01-24]
 
--- a/influxql/result.go
+++ b/influxql/result.go
@@ -35,6 +35,14 @@ func (t *TagSet) Swap(i, j int) {
 	t.Filters[i], t.Filters[j] = t.Filters[j], t.Filters[i]
 }
 
+// Reverse reverses the order of series keys and filters in the TagSet.
+func (t *TagSet) Reverse() {
+	for i, j := 0, len(t.Filters)-1; i < j; i, j = i+1, j-1 {
+		t.Filters[i], t.Filters[j] = t.Filters[j], t.Filters[i]
+		t.SeriesKeys[i], t.SeriesKeys[j] = t.SeriesKeys[j], t.SeriesKeys[i]
+	}
+}
+
 // Message represents a user-facing message to be included with the result.
 type Message struct {
 	Level string `json:"level"`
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 - [#7776](https://github.com/influxdata/influxdb/issues/7776): Add system information to /debug/vars.
 
+### Bugfixes
+
+- [#7905](https://github.com/influxdata/influxdb/issues/7905): Fix ORDER BY time DESC with ordering series keys.
```
jwilder: Can you move this to 1.2.1 sections?

influxql/result.go
```diff
@@ -35,6 +35,22 @@ func (t *TagSet) Swap(i, j int) {
 	t.Filters[i], t.Filters[j] = t.Filters[j], t.Filters[i]
 }
 
+func (t *TagSet) Reverse() *TagSet {
```
jwilder: Does this need to return a new `TagSet`?  Seems like it could just reverse the existing one to avoid the extra allocations.

CHANGELOG.md
jsternberg: Yea, let me close this PR and open a new one targeting 1.2 then.

PR #7942, opened 2017-02-03
Reduce write timeouts

jwilder:
This PR has a few changes to reduce write timeouts which can cause `500s` via the HTTP API.

* The number of cache partitions is increased from 256 to 4096 which reduces the lock contention in `partition.write`.
* If enough compactions goroutine are running concurrently, they may tie up a CPU core for long periods of time and not allow the go runtime to schedule other goroutines.  `runtime.Gosched()` was added in a few spots during compactions where this could occur to allow write goroutines to continue.
* `FileStore` locks were modified to reduce contention while the `Engine` is locked.


sebito91:
Using this patch in Production at the moment, significant performance improvement (reducing 500s to virtually nil during high read + write volumes).

PR #7954, opened 2017-02-06
Fix LIMIT and OFFSET when they are used in a subquery

### LINKED ISSUE #7885: LAST function in subquery
First of all, I love the great new subquery feature - it helps me in many cases where a specific function etc. is not (yet) available.

But I see an issue when using the ```LAST``` function in combination with subqueries:

Using the ```TOP``` function I get a correct count of the values in the database:
```
> SELECT count(*) FROM (SELECT top("value", 1) FROM "consul_value" GROUP BY *)
name: consul_value
time                 count_top
----                 ---------
1970-01-01T00:00:00Z 439
```

But using the ```LAST``` function, I get a way too long number:
```
> SELECT count(*) FROM (SELECT last("value") FROM "consul_value" GROUP BY *)
name: consul_value
time                 count_last
----                 ----------
1970-01-01T00:00:00Z 2
```

It seems that somehow only two values make it from the subquery to the outer query when using the ```LAST``` function:
```
SELECT last FROM (SELECT last("value") FROM "consul_value" GROUP BY *)
name: consul_value
time                           last
----                           ----
2017-01-25T17:14:26.65126633Z  0
2017-01-25T17:14:26.651284406Z 0
```

```
SELECT top FROM (SELECT top("value", 1) FROM "consul_value" GROUP BY *)
name: consul_value
time                           top
----                           ---
2017-01-25T13:30:44.401091Z    0
2017-01-25T13:30:44.401105Z    1
2017-01-25T13:30:44.401116Z    0
2017-01-25T13:30:44.401128Z    0
2017-01-25T13:30:44.40314Z     1
2017-01-25T13:30:44.403162Z    0
2017-01-25T13:30:44.403181Z    0
2017-01-25T13:30:44.403195Z    1
2017-01-25T13:30:44.403211Z    0
2017-01-25T13:30:44.403225Z    0
2017-01-25T13:30:44.403244Z    1
2017-01-25T13:30:44.403257Z    0
2017-01-25T13:30:44.403297Z    0
2017-01-25T13:30:46.631056Z    0
(...)
```

jsternberg:
This fixes LIMIT and OFFSET when they are used in a subquery where the
grouping of the inner query is different than the grouping of the outer
query. When organizing tag sets, the grouping of the outer query is
used so the final result is in the correct order. But, unfortunately,
the optimization incorrectly limited the number of points based on the
grouping in the outer query rather than the grouping in the inner query.

The ideal solution would be to use the outer grouping to further
organize it by the grouping for the inner subquery, but that's more
difficult to do at the moment. As an easier fix, the query engine now
limits the output of each series. This may result in these types of
queries being slower in some situations like this one:

    SELECT mean(value) FROM (SELECT value FROM cpu GROUP BY host LIMIT 1)

This will be slower in a situation where the `cpu` measurement has a
high cardinality and many different tags.

This also fixes `last()` and `first()` when they are used in a subquery
because those functions use `LIMIT 1` as an internal optimization.

Fixes #7885.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -1349,14 +1349,36 @@ func (e *Engine) createVarRefIterator(measurement string, opt influxql.IteratorO
 				continue
 			}
 
+			// If we have a LIMIT or OFFSET and the grouping of the outer query
+			// is different than the current grouping, we need to perform the
+			// limit on each of the individual series keys instead to improve
+			// performance.
+			if (opt.Limit > 0 || opt.Offset > 0) && len(opt.Dimensions) != len(opt.GroupBy) {
+				for i, input := range inputs {
+					inputs[i] = newLimitIterator(input, opt)
+				}
+			}
+
 			itr, err := influxql.Iterators(inputs).Merge(opt)
 			if err != nil {
 				influxql.Iterators(inputs).Close()
 				return err
 			}
 
+			// Apply a limit on the merged iterator.
 			if opt.Limit > 0 || opt.Offset > 0 {
-				itr = newLimitIterator(itr, opt)
+				if len(opt.Dimensions) == len(opt.GroupBy) {
+					// When the final dimensions and the current grouping are
+					// the same, we will only produce one series so we can use
+					// the faster limit iterator.
+					itr = newLimitIterator(itr, opt)
+				} else {
+					// When the dimensions are different than the current
+					// grouping, we need to account for the possibility there
+					// will be multiple series. The limit iterator in the
+					// influxql package handles that scenario.
+					itr = influxql.NewLimitIterator(itr, opt)
+				}
 			}
 			itrs = append(itrs, itr)
 		}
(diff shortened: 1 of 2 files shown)


tsdb/engine/tsm1/engine.go
```diff
@@ -1349,14 +1349,36 @@ func (e *Engine) createVarRefIterator(measurement string, opt influxql.IteratorO
...
 			}
 
+			// Apply a limit on the merged iterator.
 			if opt.Limit > 0 || opt.Offset > 0 {
-				itr = newLimitIterator(itr, opt)
+				if len(opt.Dimensions) == len(opt.GroupBy) {
+					// When the final dimensions and the current grouping are
+					// the same, we will only produce one series so we can use
+					// the faster limit iterator.
+					itr = newLimitIterator(itr, opt)
```
jwilder: nit: We might want to rename this iterator.  `newLimitIterator` and `influxql.NewLimitIterator` is a little confusing.

PR #7974, opened 2017-02-09
Allow non-admin users to execute SHOW DATABASES

### LINKED ISSUE #4785: SHOW DATABASES is Unauthorized for non-admin users
expected behavior is for influx to show databases for which the user has grants.

mark-rushakoff:
This commit introduces a new interface type, influxql.Authorizer, that
is passed as part of a statement's execution context and determines
whether the context is permitted to access a given database. In the
future, the Authorizer interface may be expanded to other resources
besides databases. In this commit, the Authorizer interface is
specifically used to determine which databases are returned when
executing SHOW DATABASES.

When HTTP authentication is enabled, the existing meta.UserInfo struct
implements Authorizer, meaning admin users can SHOW every database, and
non-admin users can SHOW only databases for which they have read and/or
write permission.

When HTTP authentication is disabled, all databases are visible through
SHOW DATABASES.

This addresses a long-standing issue where Chronograf or Grafana would
be unable to list databases if the logged-in user did not have admin
privileges.

Fixes #4785.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): https://github.com/influxdata/docs.influxdata.com/issues/1016

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,11 @@
 
 ### Features
 
-- [#7776](https://github.com/influxdata/influxdb/issues/7776): Add system information to /debug/vars.
-- [#7948](https://github.com/influxdata/influxdb/pull/7948): Reduce memory allocations by reusing gzip.Writers across requests
-- [#7553](https://github.com/influxdata/influxdb/issues/7553): Add modulo operator to the query language.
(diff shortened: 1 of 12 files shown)


mark-rushakoff:
Per Joe's comment and an offline conversation with him, I'll be pushing up a cleaner implementation most likely today.

Issue #7945, opened 2017-02-03
Lock contention in Index.CreateSeriesListIfNotExists


jwilder:
Using this `influx_stress` config which writes 100M series (no queries to avoid #7944) I eventually can't write and everything starts to return 500s.

```
diff --git a/stress/template.go b/stress/template.go
index 71572ff..20cc65b 100644
--- a/stress/template.go
+++ b/stress/template.go
@@ -12,8 +12,8 @@ var s = `
   [write.point_generator]
     [write.point_generator.basic]
       enabled = true
-      point_count = 100
-      series_count = 100000
+      point_count = 1
+      series_count = 100000000
       tick = "10s"
       jitter = true
       measurement = "cpu"
@@ -41,20 +41,6 @@ var s = `
       concurrency = 10
       ssl = false
       format = "line_http" # line_udp, graphite_tcp, graphite_udp
-
-[read]
-  [read.query_generator]
-    [read.query_generator.basic]
-      template = "SELECT count(value) FROM cpu where host='server-%v'"
-      query_count = 250
-
-  [read.query_client]
-    [read.query_client.basic]
-      enabled = true
-      addresses = ["localhost:8086"]
-      database = "stress"
-      query_interval = "100ms"
-      concurrency = 1
 `
```

The goroutine profiles show all the write goroutines blocking on `Index.Lock`.  

```
goroutine profile: total 135
84 @ 0x102e2da 0x102e3be 0x103e8fa 0x103e5e4 0x106dbcd 0x106eb7d 0x155589d 0x15a16f0 0x1426782 0x1425a3f 0x1433efc 0x1473ee0 0x1481fba 0x105a6f1
#	0x103e5e3	sync.runtime_SemacquireMutex+0x33								/usr/local/go/src/runtime/sema.go:62
#	0x106dbcc	sync.(*Mutex).Lock+0x9c										/usr/local/go/src/sync/mutex.go:87
#	0x106eb7c	sync.(*RWMutex).Lock+0x2c									/usr/local/go/src/sync/rwmutex.go:86
#	0x155589c	github.com/influxdata/influxdb/tsdb/index/tsi1.(*Index).CreateSeriesListIfNotExists+0x7c	/Users/jason/go/src/github.com/influxdata/influxdb/tsdb/index/tsi1/index.go:473
[...]

PR #7976, opened 2017-02-09
Reduce allocations

e-dard:
- [ ] Rebased/mergable
- [ ] Tests pass

This PR contributes toward some of #7945 by reducing allocations.

 - `FilterNamesTags` was accounting for around 4% of allocations (by size). It no longer allocates.
 - `execSeriesEntry` was allocating a bunch with `string(myBuf)` calls.

Not included in this PR (because I already pushed it to `tsi`, is 21e821e5f).

@benbjohnson I added a mock `FileSet` implementation, which may come in handy.

### DIFF
--- a/tsdb/index/tsi1/log_file.go
+++ b/tsdb/index/tsi1/log_file.go
@@ -544,13 +544,14 @@ func (f *LogFile) execSeriesEntry(e *LogEntry) {
 	// Generate key & series, if not exists.
 	// TODO(edd) use a pool of buffers?
 	key := AppendSeriesKey(nil, e.Name, e.Tags)
-	serie := mm.series[string(key)]
+	keyStr := string(key)
+	serie := mm.series[keyStr]
 	if serie == nil {
 		serie = &logSerie{name: e.Name, tags: e.Tags, deleted: deleted}
-		mm.series[string(key)] = serie
+		mm.series[keyStr] = serie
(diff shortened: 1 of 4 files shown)


tsdb/index/tsi1/log_file.go
```diff
@@ -544,21 +544,22 @@ func (f *LogFile) execSeriesEntry(e *LogEntry) {
 	// Generate key & series, if not exists.
 	// TODO(edd) use a pool of buffers?
 	key := AppendSeriesKey(nil, e.Name, e.Tags)
-	serie := mm.series[string(key)]
+	keyStr := string(key)
+	serie := mm.series[keyStr]
```
mark-rushakoff: Did this actually reduce allocations?

https://github.com/golang/go/wiki/CompilerOptimizations#map-lookup-by-byte

> For a map m of type `map[string]T` and `[]byte b`, `m[string(b)]` doesn't allocate. (the temporary string copy of the byte slice isn't made)

e-dard: @mark-rushakoff that compiler optimisation is moot here because we have to allocate `keyStr` anyway (since further down we save it in multiple maps).

In this case, since we already need to allocate `keyStr`, `serie := mm.series[keyStr]` and `serie := mm.series[string(key)]` are equivalent.

PR #7982, opened 2017-02-09
Don't load meta data when using TSI

### LINKED ISSUE #7978: LoadMetadataIndex called from tsi index
`LoadMetadataIndex` is only needed with the `inmem` index. It shouldn't be called when using `tsi1`.

This is causing significant allocations on startup (over `14GB` for `38M` series system).

e-dard:
Fixes #7978.

On my test instance with 38M series, `tsi` was taking around 10m to start up. With this change it starts up in less than 2 seconds.

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -18,6 +18,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/influxdata/influxdb/tsdb/index/inmem"
+
 	"github.com/influxdata/influxdb/influxql"
 	"github.com/influxdata/influxdb/models"
 	"github.com/influxdata/influxdb/pkg/estimator"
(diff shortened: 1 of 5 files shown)


tsdb/engine/tsm1/engine.go
```diff
@@ -475,6 +477,10 @@ func (e *Engine) WithLogger(log zap.Logger) {
 
 // LoadMetadataIndex loads the shard metadata into memory.
 func (e *Engine) LoadMetadataIndex(shardID uint64, index tsdb.Index) error {
+	if index.Type() != inmem.IndexName {
```
jwilder: Instead of doing this check, what do you think about calling a func on the `Index` itself.  The `inmem` would do this and `tsi` would be a `hoop`.  I think adding `index.Type() == foo` could lead to complicated logic in the code over time.

benbjohnson: That would be better but the `LoadMetadataIndex()` is fairly complicated between walking the file store and cache. If a refactor isn't too bad then I agree that we should push it down into the index.

jwilder: Makes sense

e-dard: Yep. I looked at making it seamless @jwilder with noop calls but it looked like a lot of trouble.

PR #7983, opened 2017-02-09
Increase parallel iterator buffers to improve group by query speed

jwilder:
This increases the result buffers size used by the parallel iterators to allow more concurrency in certain group by queries.

In my test dataset spanning 2yr and ~1000 queries such as:

```
SELECT mean(value) FROM stress.autogen.cpu WHERE time >= '2015-01-29T00:00:00Z' AND time < now() GROUP BY time(24h)
``` 

Latency improved from from **54s** to **4.5s** on an `m4.16xlarge` and memory usage (RSS) stayed the same at ~3.8GB.  This instance type has 64 cores which allowed them all to be utilized instead of just 1 pegged at 100%.

### DIFF
--- a/influxql/iterator.gen.go
+++ b/influxql/iterator.gen.go
@@ -453,7 +453,7 @@ type floatParallelIterator struct {
 func newFloatParallelIterator(input FloatIterator) *floatParallelIterator {
 	itr := &floatParallelIterator{
 		input:   input,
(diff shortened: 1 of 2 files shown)


jwilder:
@jsternberg Removed the protobuf change.  That was an unintended.

PR #7989, opened 2017-02-10
Add chunked processing back into v2 client

### LINKED ISSUE #7977: Add chunked request processing back into the Go client v2
### Feature Request
__Proposal:__ Add chunked request processing back into v2

__Current behavior:__ Requests only return the first 10000 rows of a query

__Desired behavior:__ Return all rows for a query that includes more than 10000 rows

__Use case:__ We're attempting to use InfluxDB as an event store, but we have series that have more than 10000 entries. We allow people to download the entire history and right now I need to run multiple queries with LIMIT and OFFSET. 

Doing a single query with chunked=true and chunk_size=10000 returns 350000 rows in about 5 seconds, while the LIMIT/OFFSET takes significantly (12x) longer.

jgeiger:
- Moving the to v2 client removed this functionality. This copies
  code back into the client. The tests were also added back into
  the test suite.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

Fixes #7977

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - [#7776](https://github.com/influxdata/influxdb/issues/7776): Add system information to /debug/vars.
 - [#7948](https://github.com/influxdata/influxdb/pull/7948): Reduce memory allocations by reusing gzip.Writers across requests
 - [#7553](https://github.com/influxdata/influxdb/issues/7553): Add modulo operator to the query language.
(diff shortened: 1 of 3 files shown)


jsternberg:
Can you rebase this and then I'll merge it?

jgeiger:
Done. Thanks.

jsternberg:
The build error looks unrelated to I'm going to merge this.

PR #7990, opened 2017-02-10
Fixed String() output for MOD operator and added MOD op precedence

jsternberg:


corylanou:
Are there any tests to verify these changes work?

jsternberg:
To be fair, there appear to be no tests for anything related to operator precedence. I'm going to go fix that.

jsternberg:
Added some tests and updated the expression parsing test to attempt to reparse the output from `String()`. We had similar code for statement parsing, but we didn't use it when testing expressions. Having it on both of them will increase the amount of reparse coverage we have.

PR #8011, opened 2017-02-15
Fix tag dereferencing panic.

benbjohnson:
## Overview

Clones series tags under lock during var ref iterator creation.

## TODO
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


benbjohnson:
@jwilder Added `CHANGELOG` entry in 8e79ca5.

PR #8015, opened 2017-02-16
Expand code coverage for undercovered packages

jsternberg:
This PR is to track the work to expand code coverage to more sections of the code in an effort to improve some of the tests. I think we have too many things that are covered in the `cmd/influxd/run` integration tests and we would benefit from building a more comprehensive test suite that more thoroughly tests the individual components.

The first step is to add tests to some components that have almost no test coverage.

Output of the following is generated with the following command:

```
$ go list ./... | grep -v cmd | xargs '-I{}' go test -cover '{}' | ruby -n -e 'puts $_.split("\t").tap {|l| if l.size == 4; l.delete_at(2); end }.join("\t")'
```

Diff between master and this branch:

```
-ok     github.com/influxdata/influxdb/monitor  coverage: 3.6% of statements
+ok     github.com/influxdata/influxdb/monitor  coverage: 92.5% of statements
-ok     github.com/influxdata/influxdb/services/precreator      coverage: 34.3% of statements
+ok     github.com/influxdata/influxdb/services/precreator      coverage: 91.2% of statements
-ok     github.com/influxdata/influxdb/services/retention       coverage: 9.4% of statements
+ok     github.com/influxdata/influxdb/services/retention       coverage: 92.5% of statements
-ok     github.com/influxdata/influxdb/services/snapshotter     coverage: 0.0% of statements
+ok     github.com/influxdata/influxdb/services/snapshotter     coverage: 82.1% of statements
-ok     github.com/influxdata/influxdb/tcp      coverage: 52.6% of statements
+ok     github.com/influxdata/influxdb/tcp      coverage: 59.1% of statements
```

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/monitor/system_test.go
+++ b/monitor/system_test.go
@@ -1,0 +1,55 @@
+package monitor_test
+
+import (
+	"os"
+	"reflect"
+	"testing"
+	"time"
+
+	"github.com/influxdata/influxdb/monitor"
+)
+
+func TestDiagnostics_System(t *testing.T) {
+	s := monitor.New(nil, monitor.Config{})
+	if err := s.Open(); err != nil {
+		t.Fatalf("unexpected error: %s", err)
+	}
+	defer s.Close()
+
+	d, err := s.Diagnostics()
+	if err != nil {
+		t.Errorf("unexpected error: %s", err)
+		return
+	}
+
+	diags, ok := d["system"]
+	if !ok {
+		t.Fatal("no diagnostics found for 'system'")
+	}
+
+	if got, exp := diags.Columns, []string{"PID", "currentTime", "started", "uptime"}; !reflect.DeepEqual(got, exp) {
+		t.Errorf("unexpected columns: got=%v exp=%v", got, exp)
+	}
+
+	// So this next part is nearly impossible to match, so just check if they look correct.
+	if exp, got := 1, len(diags.Rows); exp != got {
+		t.Fatalf("expected exactly %d row, got %d", exp, got)
+	}
+
+	if got, exp := diags.Rows[0][0].(int), os.Getpid(); got != exp {
+		t.Errorf("unexpected pid: got=%v exp=%v", got, exp)
+	}
+
+	currentTime := diags.Rows[0][1].(time.Time)
+	startTime := diags.Rows[0][2].(time.Time)
+	if !startTime.Before(currentTime) {
+		t.Errorf("start time is not before the current time: %s (start), %s (current)", startTime, currentTime)
+	}
+
+	uptime, err := time.ParseDuration(diags.Rows[0][3].(string))
+	if err != nil {
+		t.Errorf("unable to parse uptime duration: %s: %s", diags.Rows[0][3], err)
+	} else if got, exp := uptime, currentTime.Sub(startTime); got != exp {
+		t.Errorf("uptime does not match the difference between start time and current time: got=%v exp=%v", got, exp)
+	}
+}
--- a/services/precreator/service_test.go
+++ b/services/precreator/service_test.go
@@ -1,55 +1,55 @@
-package precreator
+package precreator_test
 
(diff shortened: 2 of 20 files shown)


jsternberg:
Note: I'm going to be trying to go for `> 80%` test coverage on as many packages as I can. That's the end goal.

jsternberg:
I think the added tests here are noteworthy themselves and I think waiting to modify every package will prevent the changes in here from doing good now.

monitor/system_test.go
```diff
@@ -0,0 +1,57 @@
...
+	d, err := s.Diagnostics()
+	if err != nil {
+		t.Errorf("unexpected error: %s", err)
+		return
+	}
+
+	diags, ok := d["system"]
+	if !ok {
+		t.Error("no diagnostics found for 'system'")
+		return
```
e-dard: If you don't want to continue with the test then why not use `t.Fatal`?

monitor/system_test.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	if got, exp := diags.Columns, []string{"PID", "currentTime", "started", "uptime"}; !reflect.DeepEqual(got, exp) {
+		t.Errorf("unexpected columns: got=%v exp=%v", got, exp)
+	}
+
+	// So this next part is nearly impossible to match, so just check if they look correct.
+	if len(diags.Rows) != 1 {
+		t.Errorf("expected exactly 1 row, got %d", len(diags.Rows))
```
e-dard: `t.FatalF`

e-dard: also your `got` and `exp` are the wrong way around here.

services/precreator/service_test.go
```diff
@@ -1,55 +1,61 @@
...
-	wg.Wait() // Ensure metaClient test function is called.
-	return
+	s := precreator.NewService(config)
+	s.WithLogger(zap.New(zap.NewTextEncoder()))
+	return s
 }
 
-// PointsWriter represents a mock impl of PointsWriter.
-type metaClient struct {
+type MetaClient struct {
```
e-dard: Rather than defining one here, you could just use the `internal.MetaClientMock`

services/snapshotter/service_test.go
```diff
@@ -1 +1,449 @@
...
+
+	// The snapshotter needs to be used with a tcp.Mux listener.
+	mux := tcp.NewMux()
+	go mux.Serve(l)
+
+	s.Listener = mux.Listen(snapshotter.MuxHeader)
+	return s, l, nil
+}
+
+type TSDBStore struct {
```
e-dard: You can DRY this up by using `internal.TSDBStoreMock{}`.

tcp/mux.go
```diff
@@ -203,8 +207,20 @@ func (ln *listener) Accept() (c net.Conn, err error) {
 	return conn, nil
 }
 
-// Close is a no-op. The mux's listener should be closed instead.
-func (ln *listener) Close() error { return nil }
+// Close removes this listener from the parent mux and closes the channel.
+func (ln *listener) Close() error {
+	ln.mux.mu.Lock()
```
e-dard: Personally I don't like objects reaching into their Fields' muxtexes and controlling them. I'd rather see a `Close() error` method defined on `*Mux` that takes listener and iterates over it's map of listeners, looking for the right one to close.

services/snapshotter/service_test.go
jsternberg: Yea. I'll do that. I wrote this before that existed.

tcp/mux.go
jsternberg: To be fair, this listener is not exposed and other places in the listener reach into the mux's mutex. I'll try to rearrange it though.

PR #8027, opened 2017-02-17
Remove Tags.shouldCopy, replace with forceCopy on series creation.

benbjohnson:
## Overview

Previously, tags had a `shouldCopy` flag to indicate if those tags referenced an underlying buffer and should be copied to allow GC. Unfortunately, this prevented tags from being copied that were created and referenced the mmap which caused segfaults.

This change removes the `shouldCopy` flag and replaces it with a `forceCopy` argument in `CreateSeriesIfNotExists()`. This allows the write path to indicate that tags must be cloned on insert.

## TODO
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/tsdb/meta.go
+++ b/tsdb/meta.go
@@ -145,7 +145,7 @@ func (d *DatabaseIndex) SeriesShardN(shardID uint64) int {
 }
 
 // CreateSeriesIndexIfNotExists adds the series for the given measurement to the index and sets its ID or returns the existing series object
-func (d *DatabaseIndex) CreateSeriesIndexIfNotExists(measurementName string, series *Series) *Series {
+func (d *DatabaseIndex) CreateSeriesIndexIfNotExists(measurementName string, series *Series, forceCopy bool) *Series {
 	d.mu.RLock()
 	// if there is a measurement for this id, it's already been added
 	ss := d.series[series.Key]
@@ -173,7 +173,9 @@ func (d *DatabaseIndex) CreateSeriesIndexIfNotExists(measurementName string, ser
 	series.measurement = m
 
 	// Clone the tags to dereference any short-term buffers
-	series.Tags = series.Tags.Clone()
+	if forceCopy {
+		series.CopyTags()
+	}
 	d.series[series.Key] = series
 
 	m.AddSeries(series)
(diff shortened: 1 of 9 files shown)


tsdb/meta.go
```diff
@@ -796,9 +798,12 @@ func (m *Measurement) TagSets(shardID uint64, dimensions []string, condition inf
 		tags := make(map[string]string, len(dimensions))
 
 		// Build the TagSet for this series.
+		s.mu.RLock()
 		for _, dim := range dimensions {
-			tags[dim] = s.Tags.GetString(dim)
+			tags[dim] = s.tags.GetString(dim)
```
jwilder: Shouldn't this use `Tags()` now?  Could do it just before the for loop.  That way the `Series` is the only code responsible for locking it's `tags`.

tsdb/meta.go
```diff
@@ -1981,14 +2000,16 @@ func (m *Measurement) tagValuesByKeyAndSeriesID(tagKeys []string, ids SeriesIDs)
 
 		// Iterate the tag keys we're interested in and collect values
 		// from this series, if they exist.
+		s.mu.RLock()
 		for _, tagKey := range tagKeys {
-			if tagVal := s.Tags.GetString(tagKey); tagVal != "" {
+			if tagVal := s.tags.GetString(tagKey); tagVal != "" {
```
jwilder: Same here?  Why not:
```
seriesTags  := s.Tags()
```
just before the for loop?

jwilder: Actually, could this use `s.GetTagString(tagKey)`?

tsdb/meta.go
jwilder: Or `s.GetTagString(dim)`?

benbjohnson: I was trying to avoid extra locking but I can change it to `s.GetTagString()`. I'd rather not generate a bunch of garbage by calling `Tags()`.

tsdb/meta.go
benbjohnson: I'll switch it.

tsdb/meta.go
benbjohnson: Fixed in 2aae363.

tsdb/meta.go
benbjohnson: Fixed in 2aae363.

PR #8062, opened 2017-02-27
Implement both single and multiline comments in influxql

### LINKED ISSUE #8025: Feature request: InfluxQL comments
__Proposal:__ Comments in InfluxQL.

__Current behavior:__ Comments are not supported.

__Desired behavior:__ Allow line-oriented comments in InfluxQL. Example:

```
# This query computes a bar from foo.
select bar from foo
```

__Use case:__ When performing data analysis, I end up using scripting pipelines that look like this:
```
cat my_query.infuxql | influx > output.txt
```

Inside `my_query.influxql` I have no way to document the queries (and now, subqueries). Comments would increase maintainability.

jsternberg:
A single line comment will read until the end of a line and is started
with `--` (just like SQL). A multiline comment is with `/* */`. You
cannot nest multiline comments.

Fixes #8025.

- [x] Rebased/mergable
- [ ] Tests pass
- [x] CHANGELOG.md updated
- [x] Provide example syntax
- [x] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): issue filed or pull request submitted influxdata/docs.influxdata.com#1058

### DIFF
--- a/influxql/parser.go
+++ b/influxql/parser.go
@@ -2686,13 +2686,15 @@ func (p *Parser) parseResample() (time.Duration, time.Duration, error) {
 // scan returns the next token from the underlying scanner.
 func (p *Parser) scan() (tok Token, pos Pos, lit string) { return p.s.Scan() }
 
-// scanIgnoreWhitespace scans the next non-whitespace token.
+// scanIgnoreWhitespace scans the next non-whitespace and non-comment token.
 func (p *Parser) scanIgnoreWhitespace() (tok Token, pos Pos, lit string) {
-	tok, pos, lit = p.scan()
-	if tok == WS {
+	for {
 		tok, pos, lit = p.scan()
+		if tok == WS || tok == COMMENT {
+			continue
+		}
+		return
 	}
-	return
 }
 
 // consumeWhitespace scans the next token if it's whitespace.
--- a/influxql/parser_test.go
+++ b/influxql/parser_test.go
@@ -43,6 +43,29 @@ func TestParser_ParseQuery_Empty(t *testing.T) {
 	}
 }
 
+// Ensure the parser will skip comments.
+func TestParser_ParseQuery_SkipComments(t *testing.T) {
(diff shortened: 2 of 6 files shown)


influxql/parser_test.go
```diff
@@ -43,6 +43,26 @@ func TestParser_ParseQuery_Empty(t *testing.T) {
 	}
 }
 
+// Ensure the parser will skip comments.
+func TestParser_ParseQuery_SkipComments(t *testing.T) {
```
mark-rushakoff: There are three more specific test cases I'd like to see covered:

1. `-- some comment containing /*` is treated as a line comment, not a block comment (I think that behavior is already true, just looking for test coverage)
2. Block comment inside a statement, e.g. `SELECT user /*, system, idle */ FROM cpu` is equivalent to `SELECT user FROM cpu`
3. A well-formed regex looks like it contains a block comment but remains a regex, e.g. `SELECT value FROM /foo\/*bar/ WHERE x = 1`, and maybe `/foo\/*bar*/` as well

influxql/parser.go
```diff
@@ -2688,11 +2688,13 @@ func (p *Parser) scan() (tok Token, pos Pos, lit string) { return p.s.Scan() }
 
 // scanIgnoreWhitespace scans the next non-whitespace token.
 func (p *Parser) scanIgnoreWhitespace() (tok Token, pos Pos, lit string) {
```
mark-rushakoff: Probably this method name and its comment should reflect that it skips whitespace and comments now.

jsternberg: I updated the comment, but I'm having difficulty thinking of a name that isn't too long for this function. I personally consider comments to be whitespace, but I would be willing to change the name as long as it's not `scanIgnoreWhitespaceAndComments()`.

influxql/parser_test.go
jsternberg: Updated so these are tested now.

PR #8069, opened 2017-02-27
Treat non-reserved measurement names with underscores as normal measurements

### LINKED ISSUE #8044: [1.2.0] Cannot read from measurement if its name is starting with an underscore.
### Bug report

Hey,

in  InfluxDB version 1.2.0 it is not possible to read from measurements starting with an underscore. If the name of a measurement starts with "_" the response is always empty. In a previous version (e.g. 1.0.1), the measurements could be queried as expected.

__System info:__ Windows 7 / InfluxDB 1.2.0

__Steps to reproduce:__

1. Create a measurement starting with an underscore. E.g.: _data
2. Reading from the measurement. E.g.: SELECT * FROM _data

__Expected behavior:__ The data contained in the measurement

__Actual behavior:__ An empty response

jsternberg:
A measurement name that begins with an underscore and does not conflict
with one of the reserved measurement names will now be passed untouched
to the underlying shards rather than being intercepted as an empty
measurement.

A user still shouldn't rely on measurements that begin with underscores
to always be accessible, but this will prevent the most common use case
from causing unexpected behavior since we will very rarely, if ever, add
additional system sources.

Fixed #8044.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - [#7968](https://github.com/influxdata/influxdb/issues/7968): Properly select a tag within a subquery.
 - [#8028](https://github.com/influxdata/influxdb/issues/8028): Fix panic in collectd when configured to read types DB from directory.
 - [#7880](https://github.com/influxdata/influxdb/issues/7880): Dividing aggregate functions with different outputs doesn't panic.
+- [#8044](https://github.com/influxdata/influxdb/issues/8044): Treat non-reserved measurement names with underscores as normal measurements.
 
 ## v1.2.0 [2017-01-24]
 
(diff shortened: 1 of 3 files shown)


mark-rushakoff:
> we will very rarely, if ever, add additional system sources.

If we haven't specified any other reserved system sources, then we shouldn't follow the `_$NAME` convention going forward, because we run the risk of breaking an existing user measurement of the same name.

We could either pick a prefix of an arbitrary, unlikely-to-be-used Unicode character (see http://stackoverflow.com/a/9454080 for some time-related characters) or we could "reserve" `_influxdb_$NAME` for the future.

gerrickw:
Will a 1.2.1 rc build be made soon with this change?  I could pick it up and confirm it works on my side, although the unit test looks like it tests the functionality pretty well.

PR #8080, opened 2017-03-01
Point.UnmarshalBinary() bounds check

benbjohnson:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/models/points.go
+++ b/models/points.go
@@ -6,6 +6,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"io"
 	"math"
 	"sort"
 	"strconv"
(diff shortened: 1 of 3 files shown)


models/points.go
```diff
@@ -1492,21 +1493,36 @@ func (p *point) MarshalBinary() ([]byte, error) {
...
-	i += fieldLen
+	// Read fields.
+	if len(b) < n {
+		return io.ErrShortBuffer
+	}
+	p.fields, b = b[:n], b[n:]
 
-	p.time = time.Now()
-	p.time.UnmarshalBinary(b[i:])
+	// Read key.
```
joelegasse: this should be the timestamp

benbjohnson: Fixed in dffd123.

PR #8116, opened 2017-03-09
Level compaction allocation improvements

jwilder:
This PR improves the allocations created during level compactions.  A user had a particular shard that would cause the process to stall when a compaction was at a particular series key.  This key had thousands of blocks that need to be merged and many overlapped in large ranges of time.  This caused an excessive amount of allocations which seems to have made the Go runtime pause for long periods of time.

The are two main changes here:
1. The excessive allocations was a bug where we did not clamp the window of values we needed to merge to the minimum time necessary.   This happened where there were multiple overlapping blocks for large ranges of time.  This is now fixed.  The overlapping blocks likely occurred by repeatedly overwriting data or backfilling data.
2. The compaction code used the `Value` interface which caused many small allocations converting from the interface to the specific value type (`*FloatValue`) and back again.  This causes every value being decoded/encoded to trigger an allocation.  It also causes a lot of pointer jumping around the heap when looking at the timestamp of the value.  This has been changed to use type specific values for level compactions via code generation similar to what we do with query side.  This results in a significant decrease in memory allocations and GC pressure.  The generated code does create some duplication because the Cache, WAL and snapshot compactions still rely on the general `Value` code.

There is still more work to do w/ snapshot compactions, WAL and cache to remove `Value` everywhere, but this allowed this one problem shard to finish compactions without issue.  The initial run stalled around 1.5-2hrs and had to be killed.  After this fix, the compaction completed in `23m`.  Other testing seems to improve level compactions times around 30-40% in general.

screen shot 2017-03-09 at 11 49 33 am

This graph show the `runtime.Mallocs` decreased from 20-60M down to ~1.5M.

screen shot 2017-03-09 at 11 49 46 am

Objects in the heap is also lower overall.  That last spike at the end is due to `_internal` monitoring with values still in the cache.


tsdb/engine/tsm1/encoding.go
```diff
@@ -366,6 +366,53 @@ func encodeFloatBlock(buf []byte, values []Value) ([]byte, error) {
...
+		b = packBlock(buf, BlockFloat64, tb, vb)
+
+		return nil
+	}()
+
+	putTimeEncoder(tsenc)
+	putFloatEncoder(venc)
+
+	return b, err
+}
```
benbjohnson: These `encode*ValuesBlock()` functions looks similar. Should they go into a `tmpl`?

jwilder: Probably.  The Encoder interfaces are all slightly different so I can't generate them with the current code.  I'll see if it's easy to unify the interfaces.

jwilder: Fixed in ca9c67a

PR #8119, opened 2017-03-10
add chunked/chunk size as setting/options in cli

corylanou:
While working on another PR, I noticed that `chunking` was hard coded in the CLI.  This adds options to turn chunking on/off as well setting the chunk size.

This is very useful when doing any type of scripted querying or testing.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

- [x] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): [PR 1055 Filed](https://github.com/influxdata/docs.influxdata.com/pull/1055)

### DIFF
--- a/cmd/influx/cli/cli.go
+++ b/cmd/influx/cli/cli.go
@@ -47,6 +47,7 @@ type CommandLine struct {
 	ShowVersion     bool
 	Import          bool
 	Chunked         bool
+	ChunkSize       int
 	Quit            chan struct{}
 	IgnoreSignals   bool // Ignore signals normally caught by this process (used primarily for testing)
 	ForceTTY        bool // Force the CLI to act as if it were connected to a TTY
@@ -64,6 +65,7 @@ func New(version string) *CommandLine {
 		ClientVersion: version,
 		Quit:          make(chan struct{}, 1),
 		osSignals:     make(chan os.Signal, 1),
+		Chunked:       true,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


cmd/influx/cli/cli.go
```diff
@@ -31,6 +31,10 @@ import (
 // ErrBlankCommand is returned when a parsed command is empty.
 var ErrBlankCommand = errors.New("empty input")
 
+// DefaultChunkSize is the default number of results to return
+// per response if chunked responses are turned on
+const DefaultChunkSize = 10000
```
jsternberg: Default chunk size should just be set to the zero value. We don't want to duplicate the value from the server in the client. If the chunk size is set to zero, don't include it in the client request and that will cause the server to use the default.

cmd/influx/cli/cli.go
```diff
@@ -490,6 +506,30 @@ func (c *CommandLine) retentionPolicyExists(db, rp string) bool {
...
 }
 
+// SetChunkSize sets the chunk size
+// 0 sets it back to the default
+func (c *CommandLine) SetChunkSize(cmd string) {
+	// normalize cmd
+	cmd = strings.ToLower(cmd)
+
+	// Remove the "chunk size" keyword if it exists
+	cmd = strings.TrimSpace(strings.Replace(cmd, "chunk size", "", -1))
```
jsternberg: Can this check for `chunk size` at the beginning and trim it from the beginning? It seems to me like replace is unnecessary.

corylanou: I'm not sure what you are suggesting.  Can you elaborate?

jsternberg: I'm saying just use `strings.HasPrefix` instead and then trimming the prefix instead of replacing anywhere within the line. The current implementation allows `chunk size 4 chunk size`.

cmd/influx/cli/cli.go
corylanou: ah, that makes sense.  I see that the handler already sets to the default if we pass zero.

corylanou:
@jsternberg issues addressed. Please review again when you have a change. Thanks.

corylanou:
@jsternberg ping

PR #8122, opened 2017-03-10
suppress headers in output for influx when they are the same

corylanou:
If you encountered a chunked result while processing the results of a query, you would get the headers inserted for no reason. This PR suppresses the header when they are the same so as to not inflate the output.  This is very useful when doing queries and grabbing a line count for adhoc testing, not to mention it created issues with csv layouts.

There is no good way to add testing for this without a larger refactor, so I've done extensive scenario adhoc testing.  I'll list those below, but if you are reviewing this PR, the most important thing to look for is any scenarios I might of missed.

### Before Change
```
> select * from m2
name,time,area,cpu,mem,region,val
m2,1489160933298193430,,,,west,8
m2,1489160955106239552,north,99.8,123254,,
m2,1489160980553494642,south,98.8,123525,,
m2,1489160999872615861,east,97.8,123523,,
m2,1489161005520282539,east,97.8,123522,,
name,time,area,cpu,mem,region,val
m2,1489161007072346588,east,97.8,123521,,
m2,1489161024602161179,east,97.8,123520,,
```

### After Change
```
> select * from m2
name,time,area,cpu,mem,region,val
m2,1489160933298193430,,,,west,8
m2,1489160955106239552,north,99.8,123254,,
m2,1489160980553494642,south,98.8,123525,,
m2,1489160999872615861,east,97.8,123523,,
m2,1489161005520282539,east,97.8,123522,,
m2,1489161007072346588,east,97.8,123521,,
m2,1489161024602161179,east,97.8,123520,,
```

### Before Change
```
> select * from m2
name: m2
time                area  cpu  mem    region val
----                ----  ---  ---    ------ ---
1489160933298193430                   west   8
1489160955106239552 north 99.8 123254
1489160980553494642 south 98.8 123525
1489160999872615861 east  97.8 123523
1489161005520282539 east  97.8 123522

name: m2
time                area cpu  mem    region val
----                ---- ---  ---    ------ ---
1489161007072346588 east 97.8 123521
1489161024602161179 east 97.8 123520

```

### After Change
```
> select * from m2
name: m2
time                area  cpu  mem    region val
----                ----  ---  ---    ------ ---
1489160933298193430                   west   8
1489160955106239552 north 99.8 123254
1489160980553494642 south 98.8 123525
1489160999872615861 east  97.8 123523
1489161005520282539 east  97.8 123522
1489161007072346588 east  97.8 123521
1489161024602161179 east  97.8 123520
```

### Before Change
```
> select * from m2
name: m2
time                area  cpu  mem    region val
----                ----  ---  ---    ------ ---
1489160933298193430                   west   8
1489160955106239552 north 99.8 123254
1489160980553494642 south 98.8 123525
1489160999872615861 east  97.8 123523
1489161005520282539 east  97.8 123522

name: m2
time                area cpu  mem    region val
----                ---- ---  ---    ------ ---
1489161007072346588 east 97.8 123521
1489161024602161179 east 97.8 123520

> select * from /m*/
name: m1
time                area cpu field mem region val
----                ---- --- ----- --- ------ ---
1489160894168214149                    east   1
1489160896124618043                    east   2
1489160897460385000                    east   3
1489160898924532600                    east   4
1489160900644363008                    east   5

name: m1
time                area cpu field mem region val
----                ---- --- ----- --- ------ ---
1489160907827927527                    west   6
1489160910315118036                    west   7

name: m2
time                area  cpu  field mem    region val
----                ----  ---  ----- ---    ------ ---
1489160933298193430                         west   8
1489160955106239552 north 99.8       123254
1489160980553494642 south 98.8       123525
1489160999872615861 east  97.8       123523
1489161005520282539 east  97.8       123522

name: m2
time                area cpu  field mem    region val
----                ---- ---  ----- ---    ------ ---
1489161007072346588 east 97.8       123521
1489161024602161179 east 97.8       123520

name: m3
time                area cpu field mem region val
----                ---- --- ----- --- ------ ---
1489161208341811442          0
```

### After Change
```
> select * from /m*/
name: m1
time                area cpu field mem region val
----                ---- --- ----- --- ------ ---
1489160894168214149                    east   1
1489160896124618043                    east   2
1489160897460385000                    east   3
1489160898924532600                    east   4
1489160900644363008                    east   5
1489160907827927527                    west   6
1489160910315118036                    west   7

name: m2
time                area  cpu  field mem    region val
----                ----  ---  ----- ---    ------ ---
1489160933298193430                         west   8
1489160955106239552 north 99.8       123254
1489160980553494642 south 98.8       123525
1489160999872615861 east  97.8       123523
1489161005520282539 east  97.8       123522
1489161007072346588 east  97.8       123521
1489161024602161179 east  97.8       123520

name: m3
time                area cpu field mem region val
----                ---- --- ----- --- ------ ---
1489161208341811442          0
```

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/cmd/influx/cli/cli.go
+++ b/cmd/influx/cli/cli.go
@@ -31,6 +31,10 @@ import (
 // ErrBlankCommand is returned when a parsed command is empty.
 var ErrBlankCommand = errors.New("empty input")
 
+// DefaultChunkSize is the default number of results to return
+// per response if chunked responses are turned on
+const DefaultChunkSize = 10000
+
 // CommandLine holds CLI configuration and state.
 type CommandLine struct {
 	Line            *liner.State
--- a/cmd/influx/cli/cli.go
+++ b/cmd/influx/cli/cli.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"reflect"
 	"sort"
 	"strconv"
 	"strings"
@@ -30,10 +31,6 @@ import (
 
 // ErrBlankCommand is returned when a parsed command is empty.
 var ErrBlankCommand = errors.New("empty input")
-
-// DefaultChunkSize is the default number of results to return
-// per response if chunked responses are turned on
-const DefaultChunkSize = 10000
 
 // CommandLine holds CLI configuration and state.
 type CommandLine struct {
@@ -777,16 +774,41 @@ func (c *CommandLine) writeJSON(response *client.Response, w io.Writer) {
 	fmt.Fprintln(w, string(data))
 }
 
+func tagsEqual(prev, current map[string]string) bool {
+	return reflect.DeepEqual(prev, current)
+}
+
+func columnsEqual(prev, current []string) bool {
+	return reflect.DeepEqual(prev, current)
+}
+
+func headersEqual(prev, current models.Row) bool {
+	if prev.Name != current.Name {
+		return false
+	}
+	return tagsEqual(prev.Tags, current.Tags) && columnsEqual(prev.Columns, current.Columns)
+}
+
 func (c *CommandLine) writeCSV(response *client.Response, w io.Writer) {
 	csvw := csv.NewWriter(w)
+	var previousHeaders models.Row
 	for _, result := range response.Results {
+		suppressHeaders := len(result.Series) > 0 && headersEqual(previousHeaders, result.Series[0])
+		if !suppressHeaders && len(result.Series) > 0 {
+			previousHeaders = models.Row{
+				Name:    result.Series[0].Name,
+				Tags:    result.Series[0].Tags,
+				Columns: result.Series[0].Columns,
+			}
+		}
+
 		// Create a tabbed writer for each result as they won't always line up
-		rows := c.formatResults(result, "\t")
+		rows := c.formatResults(result, "\t", suppressHeaders)
 		for _, r := range rows {
 			csvw.Write(strings.Split(r, "\t"))
 		}
-		csvw.Flush()
 	}
+	csvw.Flush()
 }
 
 func (c *CommandLine) writeColumns(response *client.Response, w io.Writer) {
@@ -794,21 +816,40 @@ func (c *CommandLine) writeColumns(response *client.Response, w io.Writer) {
 	writer := new(tabwriter.Writer)
 	writer.Init(w, 0, 8, 1, ' ', 0)
 
-	for _, result := range response.Results {
+	var previousHeaders models.Row
+	for i, result := range response.Results {
 		// Print out all messages first
 		for _, m := range result.Messages {
 			fmt.Fprintf(w, "%s: %s.\n", m.Level, m.Text)
 		}
-		csv := c.formatResults(result, "\t")
-		for _, r := range csv {
+		// Check to see if the headers are the same as the previous row.  If so, suppress them in the output
+		suppressHeaders := len(result.Series) > 0 && headersEqual(previousHeaders, result.Series[0])
+		if !suppressHeaders && len(result.Series) > 0 {
+			previousHeaders = models.Row{
+				Name:    result.Series[0].Name,
+				Tags:    result.Series[0].Tags,
+				Columns: result.Series[0].Columns,
+			}
+		}
+
+		// If we are suppressing headers, don't output the extra line return. If we
+		// aren't suppressing headers, then we put out line returns between results
+		// (not before the first result, and not after the last result).
+		if !suppressHeaders && i > 0 {
+			fmt.Fprintln(writer, "")
+		}
+
+		rows := c.formatResults(result, "\t", suppressHeaders)
+		for _, r := range rows {
 			fmt.Fprintln(writer, r)
 		}
-		writer.Flush()
+
 	}
+	writer.Flush()
 }
 
 // formatResults will behave differently if you are formatting for columns or csv
-func (c *CommandLine) formatResults(result client.Result, separator string) []string {
+func (c *CommandLine) formatResults(result client.Result, separator string, suppressHeaders bool) []string {
 	rows := []string{}
 	// Create a tabbed writer for each result as they won't always line up
 	for i, row := range result.Series {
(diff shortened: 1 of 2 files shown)


cmd/influx/cli/cli.go
```diff
@@ -777,38 +778,103 @@ func (c *CommandLine) writeJSON(response *client.Response, w io.Writer) {
 	fmt.Fprintln(w, string(data))
 }
 
+func tagsEqual(prev, current map[string]string) bool {
+	if prev == nil && current == nil {
```
e-dard: You can remove this check; `nil` maps have zero length.

cmd/influx/cli/cli.go
```diff
@@ -777,38 +778,103 @@ func (c *CommandLine) writeJSON(response *client.Response, w io.Writer) {
...
+		return true
+	}
+	if len(prev) != len(current) {
```
e-dard: You can probably just remove this too. I checked the source of `DeepEqual` and one of the first things it does it check the [length of two maps](https://golang.org/src/reflect/deepequal.go?s=5823:5860#L110).

cmd/influx/cli/cli.go
```diff
@@ -777,38 +778,103 @@ func (c *CommandLine) writeJSON(response *client.Response, w io.Writer) {
...
+		return false
+	}
+	return reflect.DeepEqual(prev, current)
+}
+
+func columnsEqual(prev, current []string) bool {
+	if prev == nil && current == nil {
+		return true
+	}
+	if len(prev) != len(current) {
```
e-dard: Remove redundant checks, as with `tagsEqual`.

cmd/influx/cli/cli.go
```diff
@@ -777,38 +778,103 @@ func (c *CommandLine) writeJSON(response *client.Response, w io.Writer) {
...
+		return false
+	}
+	return reflect.DeepEqual(prev, current)
+}
+
+func headersEqual(prev, current models.Row) bool {
+	if prev.Name != current.Name {
+		return false
+	}
+	if prev.Tags == nil && current.Tags == nil {
```
e-dard: You can remove this check.

cmd/influx/cli/cli.go
```diff
@@ -777,38 +778,103 @@ func (c *CommandLine) writeJSON(response *client.Response, w io.Writer) {
...
+	}
+	if !tagsEqual(prev.Tags, current.Tags) {
+		return false
+	}
+	if !columnsEqual(prev.Columns, current.Columns) {
+		return false
+	}
+
+	return true
```
e-dard: How about you remove `807–812` and instead do:

```go
return tagsEqual(prev.Tags, current.Tags) && columnsEqual(prev.Columns, current.Columns)
```

cmd/influx/cli/cli.go
```diff
@@ -777,38 +778,103 @@ func (c *CommandLine) writeJSON(response *client.Response, w io.Writer) {
...
+
 		// Create a tabbed writer for each result as they won't always line up
-		rows := c.formatResults(result, "\t")
+		rows := c.formatResults(result, "\t", suppressHeaders)
 		for _, r := range rows {
 			csvw.Write(strings.Split(r, "\t"))
 		}
-		csvw.Flush()
 	}
+	csvw.Flush()
```
e-dard: What's the reason for flushing less often?

cmd/influx/cli/cli.go
```diff
@@ -777,38 +778,103 @@ func (c *CommandLine) writeJSON(response *client.Response, w io.Writer) {
...
+				Name:    result.Series[0].Name,
+				Tags:    result.Series[0].Tags,
+				Columns: result.Series[0].Columns,
+			}
+		}
+
+		// if we are not at the end of the results, and we aren't suppressing headers, put in a new line
+		// to break up the results
+		// We don't print one if we are the first result, or the last result
+		if !suppressHeaders && i > 0 && i < len(response.Results) {
```
e-dard: Shouldn't this be `... && i < len(response.Results)-1 {`? The last result will have an index `< len(response.Results)`.

cmd/influx/cli/cli.go
```diff
@@ -777,38 +778,103 @@ func (c *CommandLine) writeJSON(response *client.Response, w io.Writer) {
...
+		}
+
+		rows := c.formatResults(result, "\t", suppressHeaders)
+		for _, r := range rows {
 			fmt.Fprintln(writer, r)
 		}
-		writer.Flush()
+
 	}
+	writer.Flush()
```
e-dard: Again, flush moving down. I'm sure this is fine but I'm not sure on the reasoning behind it.

cmd/influx/cli/cli.go
corylanou: You only need to call flush once.  It's always streaming.  And if you call flush to early, it actually messes up the alignment on the output (didn't read the source, but assuming it resets the column width tracking).  There is no benefit to calling flush more than once regardless.

cmd/influx/cli/cli.go
corylanou: This is putting line returns between results, and it outputs it before the call to the headers.  So the actual fix is to take out the check for `i < len(response.Results)` as it is always true.  Good catch.

cmd/influx/cli/cli.go
corylanou: Same as above.

corylanou:
@e-dard comments addressed. Take another gander when you get a chance.  Thanks.

corylanou:
also, this PR is based off of https://github.com/influxdata/influxdb/pull/8119 (chunked/size settings) to be able to test it out, so I'll rebase/merge once that one is merged.

jsternberg:
This might be more simple to implement by just starting to switch the CLI to using the alternative client. The alternative client already concatenates partial responses so you wouldn't have to check the headers.

You can also just check to see if the statement id has changed or not now that the statement id gets returned with the response. I think it would be easier than checking the headers for equality.

corylanou:
Moving to a new client is bigger change, and one I agree should be done, but this fixes a real problem now.

PR #8132, opened 2017-03-14
Ensure privileges can't be set on non-existent DB

### LINKED ISSUE #8124: Can grant permissions for non existing databases
__System info:__

InfluxDB shell version: 1.2.0
InfluxDB  version: 1.2.0
Debian 8.7

__Steps to reproduce:__

1. Create user
2. `GRANT WRITE ON doesnotexist TO user`
3. `SHOW GRANTS FOR user`

```
database     privilege
--------     ---------
telegrad     NO PRIVILEGES
telegraf     WRITE
doesnotexist WRITE

```
__Expected behavior:__ 

Error that the DB does not exist.

__Actual behavior:__

Privs are granted without issue.

e-dard:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

Fixes #8124.

This PR prevents privileges from being added to non-existent databases, by checking for the existence of the database in `SetPrivilege`.

### DIFF
--- a/services/meta/data.go
+++ b/services/meta/data.go
@@ -611,6 +611,10 @@ func (data *Data) SetPrivilege(name, database string, p influxql.Privilege) erro
 		return ErrUserNotFound
 	}
 
+	if data.Database(database) == nil {
+		return influxdb.ErrDatabaseNotFound(database)
+	}
+
 	if ui.Privileges == nil {
 		ui.Privileges = make(map[string]influxql.Privilege)
 	}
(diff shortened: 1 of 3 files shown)


services/meta/data.go
```diff
@@ -613,6 +613,10 @@ func (data *Data) SetPrivilege(name, database string, p influxql.Privilege) erro
 		return ErrUserNotFound
 	}
 
+	if data.Database(database) == nil {
```
corylanou: I'm pretty sure this gets called in plutonium and can now err out where previously it couldn't.  This could potentially corrupt the meta store if it replays in the raft log.

PR #7368, opened 2016-09-27
Introduce syntax for marking a partial response with chunking

jsternberg:
The `partial` tag has been added to the JSON response of a series and
the result so that a client knows when more of the series or result will
be sent in a future JSON chunk.

This helps interactive clients who don't want to wait for all of the
data to know if it is done processing the current series or the current
result. Previously, the client had to guess if the next chunk would
refer to the same result or a new result and it had to match the name
and tags of the two series to know if they were the same series. Now,
the client just needs to check the `partial` field included with the
response to know if it should expect more.

PR #8133, opened 2017-03-14
Disable max-row-limit by default

### LINKED ISSUE #8050: influxdb & grafana, absence of data on the graphs
__System info:__ 
OS: CentOS Linux release 7.3.1611 (Core)
influxdb: influxdb-1.2.0-1.x86_64 (https acccess)
grafana: grafana-4.1.2-1486989747.x86_64

__Steps to reproduce:__
I faced the problem, that grafana shows always "InfluxDB Error: undefined" after influxdb works for awhile.
Here is how it is presented, you can see exclamation mark on the upper left side with error (InfluxDB Error: undefined).

Dashboard loads
refresh1

Waited ~2 seconds and clicked on refresh

refresh2

So after clicking refresh some graphs show data and some not.  
The InfluxQL queries are registered in logs without errors.
But in grafana logs some queries have response code 200 and some have 502, when queries are proxied via grafana.

__Additional info:__
If I only open one graph, then response is always successful.
Moreover the problem disappears after influxdb restarted and then appears after 5-10 minutes again.

my config is [here](https://gist.githubusercontent.com/oshamin/aef4dbf095ddce13c8b55b0006ee5dae/raw/09742f67160f49605074f6d09adba2d12ce843a2/gistfile1.txt)

What can be a cause of such inconsistent behavior?

jwilder:
`max-row-limit` was set at 10000 since 1.0, but due to a bug it was
effectively 0 (disabled).  1.2 fixed this bug via #7368, but this
caused a breaking change w/ Grafana and any users upgrading from <1.2
who had not disabled the config manually.

This switches the default to `0` to match the prior behavior.

Fixes #8050 

Also related: https://github.com/grafana/grafana/issues/7380

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+## v1.2.2 [2017-03-14]
+
+### Release Notes
+
+### Configuration Changes
+
+#### `[http]` Section
+
+* `max-row-limit` now defaults to `0`.  The previous default was `10000`, but due to a bug, the value in use since `1.0` was `0`.
+
+### Bugfixes
+
+- [#8050](https://github.com/influxdata/influxdb/issues/8050): influxdb & grafana, absence of data on the graphs
+
 ## v1.2.1 [2017-03-08]
 
 ### Release Notes
--- a/cmd/influxd/run/server_test.go
+++ b/cmd/influxd/run/server_test.go
@@ -5960,7 +5960,7 @@ func TestServer_Query_Chunk(t *testing.T) {
 	}
 
 	writes := make([]string, 10001) // 10,000 is the default chunking size, even when no chunking requested.
-	expectedValues := make([]string, 10000)
+	expectedValues := make([]string, len(writes))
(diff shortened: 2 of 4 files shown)


joelegasse:
LGTM, but one of the tests is now failing around chunking queries

jwilder:
@joelegasse Tests fixed.

joelegasse:
The test looks like it was meant to test chunking. Instead of disabling chunking, should the config be updated to set a limit?

jwilder:
@joelegasse This is how the test was working before #7368.  https://github.com/influxdata/influxdb/pull/7368/commits/b4db76cee23dcb7d57f8e99001c23d1fdec150a7#diff-09d8dc75f001e1e2a64439ee67dd14cbR5469

torkelo:
yes! so glad this is changed back. so many issues around this and haven't had time to implement chunked response handling in Grafana.

PR #8143, opened 2017-03-15
Add WAL sync delay

jwilder:
This is a follow up to address an issue introduced #7942 where write performance was lower due to an 100ms delay added to batch up fsyncs calls to the WAL.  This reworks it so that there is no delay by default, but multiple, concurrent fsyncs are coalesced into one.  It also adds a `wal-fsync-delay` config option that can be increased to add a delay.  The previous version also added another long running goroutine per shard.  That goroutine is removed now.

This change helps with high write load use cases, users not running on SSDs and when there are lots of small writes.

This also avoids some of the work done to collect shard stat data.  Previously, we'd `stat` the shard data and wal dirs on every collection interval (10s).  For cold shards, this these stats don't change.  When there are many shards (1000's), these thousands of syscalls can be taxing and on the system when run so frequently.

These stats are only collected now when the shard data on disk has actually changed due to a write or a compaction.

I tested this on an `m4.xlarge` w/ `sc1` volume (HDD 250 IOPS) using various workloads.

For small writes using `influx-stress insert --pps 1200 -n 100000000 -b 10`, the disk is `100%` utilized continuously and write latency is around 5.7s.

### 1.2

```
Device:         rrqm/s   wrqm/s     r/s     w/s    rkB/s    wkB/s avgrq-sz avgqu-sz   await r_await w_await  svctm  %util
xvda              0.00     0.00    0.00    0.00     0.00     0.00     0.00     0.00    0.00    0.00    0.00   0.00   0.00
xvdf              0.00    23.00    0.00   63.00     0.00   348.00    11.05     1.02   16.32    0.00   16.32  15.87 100.00
```

### New w/  `100ms` delay

Write latency improves to `1.5s` and disk util stays around 2-3%.

```
Device:         rrqm/s   wrqm/s     r/s     w/s    rkB/s    wkB/s avgrq-sz avgqu-sz   await r_await w_await  svctm  %util
xvda              0.00     1.00    0.00    8.00     0.00    36.00     9.00     0.00    0.00    0.00    0.00   0.00   0.00
xvdf              0.00    20.00    0.00   32.00     0.00   212.00    13.25     0.03    0.88    0.00    0.88   0.75   2.40
```

Testing with SSDs shows some improvement as well, but doesn't require the delay.

### DIFF
--- a/tsdb/engine/tsm1/wal.go
+++ b/tsdb/engine/tsm1/wal.go
@@ -90,6 +90,12 @@ type WAL struct {
 	closing chan struct{}
 	// goroutines waiting for the next fsync
 	syncWaiters chan chan error
+	syncCount   uint64
+
+	// syncDelay sets the duration to wait before fsyncing writes.  A value of 0 (default)
+	// will cause every write to be fsync'd.  This must be set before the WAL
+	// is opened if a non-default value is required.
+	syncDelay time.Duration
 
 	// WALOutput is the writer used by the logger.
 	logger       zap.Logger // Logger to be used for important messages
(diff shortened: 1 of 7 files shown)


jwilder:
@mark-rushakoff Fixed in c72f69c

tsdb/engine/tsm1/wal.go
```diff
@@ -129,6 +131,13 @@ func (l *WAL) enableTraceLogging(enabled bool) {
 	}
 }
 
+// setSyncDelay sets the duration to wait before fsync writes.  A value of 0
+// will cause every write to be fsync'd.  This must be called before the WAL
+// is opened.
+func (l *WAL) setSyncDelay(d time.Duration) {
```
jwilder: I was trying to avoid adding another parameter to `NewWAL` and passing in the whole `Config` was kind of messy too.

joelegasse: I'd say just set the field directly, then. Creating an unexported "setter" method just feels wrong since it's not doing anything other than a field assignment.

tsdb/engine/tsm1/wal.go
```diff
@@ -221,30 +230,41 @@ func (l *WAL) Open() error {
...
-	t := time.NewTicker(100 * time.Millisecond)
-	defer t.Stop()
-	for {
+// sync will schedule an fsync to the current wal segment and notify any
+// waiting gorutines.  If an fsync is already scheduled, subsequent calls will
+// not schedule a new fsync and will be handle by the existing scheduled fsync.
+func (l *WAL) sync() {
+	// If we're not the first to increment the sync counter, then another
+	// goroutine is and will fsync the wal for us.
+	if n := atomic.AddInt64(&l.syncCount, 1); n > 1 {
```
jwilder: Good idea.  I'll give that try.

PR #8155, opened 2017-03-17
Reduce cost of admin user check

e-dard:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

This PR simplifies the process of checking for the existence of an admin user when handling a query.

### DIFF
--- a/services/meta/data.go
+++ b/services/meta/data.go
@@ -42,6 +42,10 @@ type Data struct {
 	Databases []DatabaseInfo
 	Users     []UserInfo
 
+	// adminUserExists provides a constant time mechanism for determining
+	// if there is at least one admin user.
+	adminUserExists bool
+
 	MaxShardGroupID uint64
 	MaxShardID      uint64
 }
@@ -568,6 +572,11 @@ func (data *Data) CreateUser(name, hash string, admin bool) error {
 		Admin: admin,
 	})
 
+	// We know there is now at least one admin user.
+	if admin {
+		data.adminUserExists = true
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


e-dard:
@joelegasse @dgnorton @mark-rushakoff please review 9d9b3254

services/meta/data.go
```diff
@@ -576,9 +585,12 @@ func (data *Data) DropUser(name string) error {
 	for i := range data.Users {
 		if data.Users[i].Name == name {
 			data.Users = append(data.Users[:i], data.Users[i+1:]...)
+			// Maybe we dropped the only admin user?
+			data.adminUserExists = data.hasAdminUser()
```
joelegasse: You could add a check here to see if the user being removed is an admin, first.

services/meta/data.go
```diff
@@ -576,9 +585,12 @@ func (data *Data) DropUser(name string) error {
 	for i := range data.Users {
 		if data.Users[i].Name == name {
 			data.Users = append(data.Users[:i], data.Users[i+1:]...)
+			// Maybe we dropped the only admin user?
```
dgnorton: It should prevent this.

services/meta/data.go
```diff
@@ -731,6 +751,17 @@ func (data *Data) UnmarshalBinary(buf []byte) error {
 	return nil
 }
 
+// hasUserExists exhaustively checks for the presence of at least one admin
```
mark-rushakoff: `s/hasUserExists/hasAdminUser`

services/meta/data.go
e-dard: @joelegasse but there could still be an admin lurking in the background.

services/meta/data.go
e-dard: Isn't that out of the scope of this PR though?

dgnorton: true

services/meta/data.go
joelegasse: I meant if the user isn't an admin, there's not a point in refreshing the cached value

Issue #7811, opened 2017-01-10
the kill query doesn't work in 1.0.2 version


pingliu:
I use kill query for test.
when I run:
```
select value from "cpu.clock"
```
then run:
```
kill query 187
```
the query is killed ,but the memory of servers still rise up .
![image](https://cloud.githubusercontent.com/assets/6415493/21804948/a1bf6904-d76c-11e6-803f-0122aa66016a.png)

desa:
@pingliu can you verify that this is still an issue with `1.1.1`.

pingliu:
@desa  I test it with 1.1.1. 
it is still issue with 1.1.1.

![image](https://cloud.githubusercontent.com/assets/6415493/22005920/341182e4-dca2-11e6-8aff-8d482ca699ca.png)
![image](https://cloud.githubusercontent.com/assets/6415493/22005932/428312b6-dca2-11e6-8055-38c570140128.png)

jwilder:
Fixed via #8158

PR #8158, opened 2017-03-17
Fix killing queries

### LINKED ISSUE #7457: KILL QUERY should work during all phases of a query
### Feature Request

Opening a feature request kicks off a discussion.
Requests may be closed if we're not actively planning to work on them.

**Proposal:** 

Make it so queries can be interrupted during the planning phase, not just the execution phase.

**Current behavior:**

`KILL QUERY` can't interrupt queries until the query is at the execution phase, actually sampling data from disk. With large datasets, a query like `SELECT * FROM /.*/` will spend a long time and a very large amount of RAM just instantiating the query engine iterators. During that phase, `KILL QUERY` won't interrupt the process and the RAM use will continue to rise.

**Desired behavior:** 

`KILL QUERY` should terminate the query as quickly as feasible no matter what phase the query is in.

**Use case:**

Currently there are queries that can't be killed before they OOM the system even if the query is detected and a `KILL` is issued.

jwilder:
While troubleshooting some high cardinality issues, I discovered that `max-select-series` did not actually kill a query until after it was planned.  In this particular case, RAM blew up during planning. (#7457)

This change fixes `max-select-series` so that it's killed as soon as we know the query will exceed the limit.  It has the side effect that queries can now be killed during the planning phase even with `kill query`.

There was also a bug in the `InterruptIterators` where even when a query was killed, it would still continue to execute until completion (#7811).  The client would get an error saying it was interrupted, but it actually wasn't completely.

- [x] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

Fixes #7457 #7811


influxql/iterator.gen.go
```diff
@@ -767,7 +767,7 @@ func (itr *floatInterruptIterator) Next() (*FloatPoint, error) {
 	if itr.count&0xFF == 0xFF {
 		select {
 		case <-itr.closing:
-			return nil, nil
+			return nil, itr.Close()
```
jsternberg: I think the reason this returned nil was mostly because another part of the code is supposed to call `Close()` and I didn't want to double close. Is the query engine not calling close correctly? It's supposed to call `Next()` until it gets a `nil` and then call `Close()`.

tsdb/engine/tsm1/engine.go
```diff
@@ -1352,7 +1360,7 @@ func (e *Engine) createVarRefIterator(measurement string, opt influxql.IteratorO
 	}
 
 	// Determine tagsets for this measurement based on dimensions and filters.
-	tagSets, err := mm.TagSets(e.id, opt.Dimensions, opt.Condition)
+	tagSets, err := mm.TagSets(e.id, opt.Dimensions, opt.Condition, opt)
```
jsternberg: Should this just take a `<-chan struct{}` instead? If we want to change it to take the iterator options, we should just remove the `opt.Dimensions` and `opt.Condition` and pass that as part of the IteratorOptions.

influxql/iterator.gen.go
jwilder: This is the code path where a query is interrupted either via `kill query` or a hitting a limit.  The cause of #7811 is that `itr.Close()` is not called in this case.  I can change it to call `itr.Close()` and return nil though.

tsdb/engine/tsm1/engine.go
jwilder: Removed these in 8177df2

influxql/iterator.gen.go
jsternberg: Ok, I don't think there are any iterators that can't accept a double close. If there are, we can add a check to the interrupt iterator so it will only perform a single close.

PR #8160, opened 2017-03-17
Fixed typo in etc/config.sample.toml

eplanet:
This is a very trivial PR to correct a type in the config sample.


mark-rushakoff:
Nice catch, thanks.

Issue #8042, opened 2017-02-22
More operators please.


taffmak:
### Feature Request
__Proposal:__ 
Implement the full complement of arithmetic operators including logical bitwise operators for and, or, not (invert) and xor.

__Current behavior:__ 
Not currently supported.

__Desired behavior:__ 
Ability to perform bitwise logical operations on integer field values.

__Use case:__
Enable the use of field values as bit fields that can be easily queried with bitwise logical operators.
We are instrumenting our applications and streaming measurements through a processing application in real-time.  The application outputs measurements to influx in addition to identifying issues in real-time.  E.g. stalling queues, counter resets - it can also detect issues over several measurements that are difficult to query for, such as a queue where latency is increasing over some period of time.

We are raising events in our system when these things occur as they are important to us in real-time.

It would also useful to pass through to influx the live detected status as bits within a field so we can query back these occurrences after the fact.

e.g. We can query the number of times per month that a queue stalls, is reset or it's length exceeds a threshold (where the thresholds may be different for each series and are not known to influx).

Tomcat-Engineering:
Like you, we use InfluxDB to store lots of status words so this could be useful for us too.

There is an issue with implementing bitwise-not as an operator: what word size do we assume?

E.g. what should "NOT 0" return?  If the word size is 8 bits, we should return 0b11111111 (i.e. 0xFF, decimal 255) but if the word size is bigger obviously we would need to return something different.

If implementing a generic operator we would have to assume that the word size is the biggest that InfluxDB supports, 64 bits, but unfortunately because integers in InfluxDB are defined as the signed `int64` type, "NOT 0" would return binary of 64 ones which is interpreted and displayed as minus one!  Neither 64 binary ones nor a negative number is going to be very helpful to most users.

I guess you could always AND the result with a bitmask representing the word width to get it back to a reasonable size and get rid of the negatives (it is only the very highest bit out of the 64 which makes things be interpreted as negative), but it would be a bit annoying to have to do something like `(NOT x) AND 0xFF` every time.

One option is to define bitwise-not as a function rather than an operator, e.g. `bitwise_not(value, 8)` where the second argument is the word width in bits.  Would that be acceptable and better than having to do `(NOT x) AND 0xFF` every time?

Happily this only applies to NOT - the other bitwise operators that you mention (AND, OR, XOR) can all just operate on standard integers and you will get sensibly-sized non-negative results because all the unused higher bits of the result will be zero.

[...]

PR #8163, opened 2017-03-19
Influxql bitwise operators

Tomcat-Engineering:
Relates to #8042 

This implements bitwise AND, OR and XOR operators.   They can be used in both the main part of a SELECT statement, and in a WHERE clause.  They accept any non-negative integer (field or constant) on either side.

There is no bitwise-NOT operator, for reasons explained in [the documentation](https://github.com/Tomcat-Engineering/docs.influxdata.com/blob/add-bitwise-operators/content/influxdb/v1.2/query_language/math_operators.md#bitwise-not).  There is a simple workaround which avoids the need for a new influxql function, also explained in the documentation.

Negative numbers are not supported, because representing a bitfield as a negative number is a very advanced use-case, involving detailed knowledge of twos-complement arithmetic and the underlying `int64` implementation.  In practice all this means is that we support bitfields up to 63 bits wide instead of 64.  If you try to use one of these operators with a negative number then you just get 0 as a result - this is the same behaviour as InfluxQL uses for other "invalid operands" like divide-by-zero errors.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): https://github.com/influxdata/docs.influxdata.com/pull/1067


jsternberg:
What's the rationale for restricting this to positive numbers? It might be worth limiting this to int64's only rather than floats, but I'm not sure why any of the relevant operators are affected by if they are positive or negative.

jwilder:
Should these support `boolean` fields as well?

Tomcat-Engineering:
@jsternberg The rationale for excluding negative integers was explained in the description of the pull request and on the ticket (#8042),  but it seems to be confusing people so I have removed the restriction.

The new operators don't work with floats.

@jwilder I am adding support for booleans too.  But you won't be able to mix integers and booleans.

influxql/select.go
```diff
@@ -1429,6 +1503,16 @@ func buildTransformIterator(lhs Iterator, rhs Iterator, op Token, opt IteratorOp
...
 		}
 		return newIntegerBooleanExprIterator(left, right, opt, fn), nil
+	case func(bool, bool) bool:
+		left, ok := lhs.(BooleanIterator)
+		if !ok {
+			return nil, fmt.Errorf("type mismatch on LHS, unable to use %T as a BooleanIterator", lhs)
+		}
+		right, ok := rhs.(BooleanIterator)
+		if !ok {
+			return nil, fmt.Errorf("type mismatch on LHS, unable to use %T as a BooleanIterator", rhs)
```
jsternberg: This looks like a copy/paste error. Shouldn't this be `RHS`?

influxql/select_test.go
```diff
@@ -2284,6 +2284,33 @@ func TestSelect_BinaryExpr_Integer(t *testing.T) {
 				{&influxql.FloatPoint{Name: "cpu", Time: 9 * Second, Value: 1}},
 			},
 		},
+		{
```
jsternberg: Can we also get tests added for the boolean points?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### Features
 
+- [#8042](https://github.com/influxdata/influxdb/issues/8042): Add bitwise AND, OR and XOR operators to the query language
```
jsternberg: Minor nitpick. Can you move this to the bottom after you rebase? This is really annoying me for some reason, but it really shouldn't annoy me.

jsternberg:
After the unit tests pass, can you squash this to a single commit and then we'll merge it? Thanks. It looks good to me at the moment.

jsternberg:
Can you squash the changelog tweak into the first commit?

Tomcat-Engineering:
Done - sorry, thought you had a magic "squash and merge" button these days!

jsternberg:
Apparently, I do have a button that does that. I was not aware of that. Thanks though! I guess I'll try that out next time.

PR #8168, opened 2017-03-20
Fix a regression when math was used with selectors

### LINKED ISSUE #8167: [1.2 regression] Math on multiple selectors returns distinct points
v1.2.1 behavior:

```
> insert metrics value=1
> insert metrics value=2
> SELECT min(value),max(value) from metrics
name: metrics
time min max
---- --- ---
0    1   2

> SELECT min(value)*1,max(value)*1 from metrics
name: metrics
time                min max
----                --- ---
1490013334977108401 1   
1490013337400734540     2
```

1.1.4 behavior:

```
> insert metrics value=1
> insert metrics value=2
> SELECT min(value),max(value) from metrics
name: metrics
time	min	max
----	---	---
0	1	2

> SELECT min(value)*1,max(value)*1 from metrics
name: metrics
time	min	max
----	---	---
0	1	2
```

May be related to https://github.com/influxdata/influxdb/issues/8072

@jsternberg

jsternberg:
If there were multiple selectors and math, the query engine would
mistakenly think it was the only selector in the query and would not
match their timestamps.

Fixed the query engine to pass whether the selector should be treated as
a selector so queries like `max(value) * 1, min(value) * 1` will match
the timestamps of the result.

Fixes #8167.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/influxql/select.go
+++ b/influxql/select.go
@@ -971,13 +971,13 @@ func (b *exprIteratorBuilder) buildBinaryExprIterator(expr *BinaryExpr) (Iterato
 			return nil, fmt.Errorf("unable to construct an iterator from two literals: LHS: %T, RHS: %T", lhs, rhs)
 		}
 
-		lhs, err := buildExprIterator(expr.LHS, b.ic, b.sources, b.opt, IsSelector(expr.LHS))
+		lhs, err := buildExprIterator(expr.LHS, b.ic, b.sources, b.opt, b.selector)
 		if err != nil {
 			return nil, err
 		}
 		return buildRHSTransformIterator(lhs, rhs, expr.Op, b.opt)
 	} else if lhs, ok := expr.LHS.(Literal); ok {
-		rhs, err := buildExprIterator(expr.RHS, b.ic, b.sources, b.opt, IsSelector(expr.RHS))
+		rhs, err := buildExprIterator(expr.RHS, b.ic, b.sources, b.opt, b.selector)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


jsternberg:
@rbetts proposed for backporting to 1.2.

influxql/select.go
```diff
@@ -971,13 +971,13 @@ func (b *exprIteratorBuilder) buildBinaryExprIterator(expr *BinaryExpr) (Iterato
...
 
-		lhs, err := buildExprIterator(expr.LHS, b.ic, b.sources, b.opt, IsSelector(expr.LHS))
+		lhs, err := buildExprIterator(expr.LHS, b.ic, b.sources, b.opt, b.selector)
 		if err != nil {
 			return nil, err
 		}
 		return buildRHSTransformIterator(lhs, rhs, expr.Op, b.opt)
 	} else if lhs, ok := expr.LHS.(Literal); ok {
-		rhs, err := buildExprIterator(expr.RHS, b.ic, b.sources, b.opt, IsSelector(expr.RHS))
+		rhs, err := buildExprIterator(expr.RHS, b.ic, b.sources, b.opt, b.selector)
```
mark-rushakoff: The current implementation of IsSelector:

```
func IsSelector(expr Expr) bool {
	if call, ok := expr.(*Call); ok {
		switch call.Name {
		case "first", "last", "min", "max", "percentile", "sample", "top", "bottom":
			return true
		}
	}
	return false
}
```

I'm concerned about the other uses of `IsSelector` being broken in the same way it was here – missing a Call in a binary expression, etc. Is there a separate issue to track/audit the other uses?

jsternberg: `IsSelector` wasn't broken here. It was just improperly used. `IsSelector` correctly identified the expression was a selector, but it was supposed to recognize an earlier conclusion by the query engine not to treat the selectors as selectors.

The reason is because `max()`, when it's treated as a selector, uses the time of the point it selects. When it's treated as an aggregate, it normalizes the time to be the beginning of the current interval. When we deal with any expression where there are multiple selectors (including binary expressions or separately selected), it's not supposed to treat the selectors as selectors.

mark-rushakoff: I understand now - `IsSelector` is working as intended. Fine by me to merge.

PR #8185, opened 2017-03-22
Fix query parser when using addition and subtraction without spaces

### LINKED ISSUE #8174: query parser is confused by subtraction without whitespace
### Bug report

__System info:__ 1.2.2

__Steps to reproduce:__

1. start influx CLI shell
2. enter

e.g.

    > select * from my_measurement where time > now()-1m and db='mydb';

__Expected behavior:__ 

No parse error, the same result as with:

    > select * from my_measurement where time > now()- 1m and db='mydb';

or

    > select * from my_measurement where time > now() - 1m and db='mydb';

__Actual behavior:__

this error is printed:

    ERR: error parsing query: found -1m, expected ; at line 1, char 46

__Additional info:__ 

This violates the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). I don't know any other shell/expression language that has this limitation - dealing with negative signs when parsing infix expressions is a standard exercise.

jsternberg:
Additionally, support unary addition and subtraction for variables,
calls, and parenthesis expressions. Doing `-value` will be the
equivalent of doing `-1 * value` now.

Fixes #8174.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/influxql/parser_test.go
+++ b/influxql/parser_test.go
@@ -2827,6 +2827,15 @@ func TestParser_ParseExpr(t *testing.T) {
 		// Primitives
 		{s: `100.0`, expr: &influxql.NumberLiteral{Val: 100}},
 		{s: `100`, expr: &influxql.IntegerLiteral{Val: 100}},
+		{s: `-100.0`, expr: &influxql.NumberLiteral{Val: -100}},
+		{s: `-100`, expr: &influxql.IntegerLiteral{Val: -100}},
+		{s: `100.`, expr: &influxql.NumberLiteral{Val: 100}},
+		{s: `-100.`, expr: &influxql.NumberLiteral{Val: -100}},
+		{s: `.23`, expr: &influxql.NumberLiteral{Val: 0.23}},
+		{s: `-.23`, expr: &influxql.NumberLiteral{Val: -0.23}},
+		{s: `1s`, expr: &influxql.DurationLiteral{Val: time.Second}},
+		{s: `-1s`, expr: &influxql.DurationLiteral{Val: -time.Second}},
+		{s: `-+1`, err: `found +, expected identifier, number, duration, ( at line 1, char 2`},
 		{s: `'foo bar'`, expr: &influxql.StringLiteral{Val: "foo bar"}},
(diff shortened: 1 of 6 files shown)


jsternberg:
Updated the InfluxQL specification to include the "+" and "-" as part of the syntax.

influxql/parser_test.go
```diff
@@ -2827,6 +2827,15 @@ func TestParser_ParseExpr(t *testing.T) {
...
 		{s: `100`, expr: &influxql.IntegerLiteral{Val: 100}},
+		{s: `-100.0`, expr: &influxql.NumberLiteral{Val: -100}},
+		{s: `-100`, expr: &influxql.IntegerLiteral{Val: -100}},
+		{s: `100.`, expr: &influxql.NumberLiteral{Val: 100}},
+		{s: `-100.`, expr: &influxql.NumberLiteral{Val: -100}},
+		{s: `.23`, expr: &influxql.NumberLiteral{Val: 0.23}},
+		{s: `-.23`, expr: &influxql.NumberLiteral{Val: -0.23}},
+		{s: `1s`, expr: &influxql.DurationLiteral{Val: time.Second}},
+		{s: `-1s`, expr: &influxql.DurationLiteral{Val: -time.Second}},
+		{s: `++-+- -1`, expr: &influxql.IntegerLiteral{Val: -1}},
```
mark-rushakoff: Is this a case we want to support? It seems very unlikely that someone would intentionally write an expression like this. 

It also seems odd that `++` would work but presumably `--` would begin a line comment. I think I'd prefer to see a single unary operator allowed before a number, and if someone wants to do multiple of them, they have to parenthesize the arguments, e.g. `-(+(-(mean(x))))`.

jsternberg: I'll look at modifying that and also verifying parenthesis in front of a parenthetical expression works. I don't actually think that does work at the moment.

jsternberg:
Done. Also added a few more goodies like supporting a unary negative expression. See the description for details.

PR #8191, opened 2017-03-23
redact passwords before saving history in cli

### LINKED ISSUE #8190: History file should redact passwords before saving to history
Statements like `create user username with password 'password'` are stored in clear text in .influx_history

This can be considered as a security vulnerability since passwords are saved in a clear text file. Passwords should be concealed in the history file.

corylanou:
fixes #8190 

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 ### Bugfixes
 
+- [#8190](https://github.com/influxdata/influxdb/issues/8190): History file should redact passwords before saving to history.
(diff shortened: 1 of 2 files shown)


corylanou:
@dgnorton auth doesn't log anything to the history, it's strictly an interactive command that sets in memory state.

corylanou:
@dgnorton can I get another look at this.  Auth does't record any data to the history log.

PR #8179, opened 2017-03-21
Speedup shutdown for high cardinality databases

e-dard:
###### Required for all non-trivial PRs
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

Fixes #8171.

The current system [times out](https://github.com/influxdata/influxdb/blob/master/cmd/influxd/main.go#L108) after 30 seconds if the system has not gracefully shutdown before then.

For databases with a large number of series sometimes this hard limit is often triggered because it's very expensive to remove all of the series from the index. Of course, there is not actually a need to carefully remove all the series from the index, because we're shutting down the entire system.

Much like #6623 we can just close the shards and move on.

For `v1.2.1` running on a 36 core box with 400K series:

```
vice=retention
[I] 2017-03-21T13:08:55Z Listening for signals
[I] 2017-03-21T13:08:55Z Storing statistics in database '_internal' retention policy 'monitor', at interval 10s service=monitor
[I] 2017-03-21T13:08:55Z Sending usage statistics to usage.influxdata.com
^C[I] 2017-03-21T13:09:02Z Signal received, initializing clean shutdown...
[I] 2017-03-21T13:09:02Z Waiting for clean shutdown...
[I] 2017-03-21T13:09:02Z shutting down monitor system service=monitor
[I] 2017-03-21T13:09:02Z snapshot listener closed service=snapshot
[I] 2017-03-21T13:09:02Z terminating storage of statistics service=monitor
[I] 2017-03-21T13:09:02Z Precreation service terminating service=shard-precreation
[I] 2017-03-21T13:09:02Z continuous query service terminating service=continuous_querier
[I] 2017-03-21T13:09:02Z retention policy enforcement terminating service=retention
[I] 2017-03-21T13:09:32Z time limit reached, initializing hard shutdown
ubuntu@ip-172-30-2-55:~$
```

Will now shutdown in under a second:

```
[I] 2017-03-21T13:13:18Z Listening on HTTP:[::]:8086 service=httpd
[I] 2017-03-21T13:13:18Z Starting retention policy enforcement service with check interval of 30m0s service=retention
[I] 2017-03-21T13:13:18Z Listening for signals
[I] 2017-03-21T13:13:18Z Sending usage statistics to usage.influxdata.com
^C[I] 2017-03-21T13:13:23Z Signal received, initializing clean shutdown...
[I] 2017-03-21T13:13:23Z Waiting for clean shutdown...
[...]

PR #8204, opened 2017-03-24
Speed up drops/deletes and startup

jwilder:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

This PR speeds up all drop and deletes statements as well as startup times.  The drops and delete statements were slow due to cleaning up the in-memory index as well as some inefficient code paths in the TSM tombstoning path.

* Deleting series keys using `FileStore.Walk` which iterated over every key in every TSM file sequentially.  It spent a lot of time and memory deduplicating keys and converting them to the underlying series keys.  This was switch to leverage the already sorted TSM file indexes to merge and iterate over them in parallel.  
* A side effect of optimizing `FileStore.Walk` is that it is also used at startup to reload the index.  This now reduces the number of keys that are parsed across all files.  I noticed ~50% speedup with my dataset, but did not experiment with other datasets.  It's possible the _optimize_ compactions may be able to be removed in the future if this change keeps start times consistent.
* Writing tombstone files did not buffer writes.  When deleting many series keys, a lot of IO time due to small writes would slow the writes down.  This now uses a buffered writer.
* Deleting a retention policy or `delete series` used a slower code path that needed to remove series ids from the in-memory measurement index.  There is a sorted slice of series IDs that was constantly maintained and cause memory to be continuously re-copied.  Keeping this sorted during large bulk deletes was too slow.  Instead, we create the sorted slice on-demand the first time it's requested by a query.  Deletes now just delete from a map which is much faster.
* Deleting keys in various parts of the TSM engine used some intermediate maps.  This created garbage and high memory usage.  These were switched to use approaches that leveraged the already sorted series keys to avoid the intermediate allocations.

Some benchmarks from testing w/ a dataset of 5M series in some unoptimized TSM files (e.g the keys were repeated across many big TSM files).  I test the same dataset w/ 1.2.2 and this branch:

### Drop retention policy
Before **10m35s**, after **7s**

### Drop measurement
Before **25m6s**, after **29s**

Delete series and drop database (also optimized further in #8179) had similar speedups.

### DIFF
--- a/tsdb/engine/tsm1/file_store.go
+++ b/tsdb/engine/tsm1/file_store.go
@@ -1,6 +1,7 @@
 package tsm1
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"math"
@@ -299,16 +300,34 @@ func (f *FileStore) Remove(paths ...string) {
 // exists in multiple files, it will be invoked for each file.
 func (f *FileStore) WalkKeys(fn func(key []byte, typ byte) error) error {
 	f.mu.RLock()
-	defer f.mu.RUnlock()
+	if len(f.files) == 0 {
+		f.mu.RUnlock()
+		return nil
+	}
 
+	readers := make([]chan seriesKey, 0, len(f.files))
 	for _, f := range f.files {
-		for i := 0; i < f.KeyCount(); i++ {
-			key, typ := f.KeyAt(i)
-			if err := fn(key, typ); err != nil {
-				return err
+		ch := make(chan seriesKey, 1)
+		readers = append(readers, ch)
+
+		go func(c chan seriesKey, r TSMFile) {
+			n := r.KeyCount()
+			for i := 0; i < n; i++ {
+				key, typ := r.KeyAt(i)
+				c <- seriesKey{key, typ}
 			}
+			close(ch)
+		}(ch, f)
+	}
+	f.mu.RUnlock()
+
+	merged := merge(readers...)
+	for v := range merged {
+		if err := fn(v.key, v.typ); err != nil {
+			return err
 		}
 	}
+
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


tsdb/engine/tsm1/file_store.go
```diff
@@ -318,12 +337,19 @@ func (f *FileStore) Keys() map[string]byte {
...
+		return nil
+	}); err != nil {
+		return nil
 	}
+	// for _, f := range f.files {
+	// 	for i := 0; i < f.KeyCount(); i++ {
+	// 		key, typ := f.KeyAt(i)
+	// 		uniqueKeys[string(key)] = typ
+	// 	}
+	// }
```
benbjohnson: Can this be deleted?

tsdb/engine/tsm1/file_store.go
```diff
@@ -318,12 +337,19 @@ func (f *FileStore) Keys() map[string]byte {
 	defer f.mu.RUnlock()
 
 	uniqueKeys := map[string]byte{}
-	for _, f := range f.files {
-		for i := 0; i < f.KeyCount(); i++ {
-			key, typ := f.KeyAt(i)
-			uniqueKeys[string(key)] = typ
-		}
+	if err := f.WalkKeys(func(key []byte, typ byte) error {
+		//println(string(key))
```
benbjohnson: Can you delete this line?

jwilder:
@benbjohnson commented lines removed.  The merge is essentially [this](https://en.wikipedia.org/wiki/Merge_algorithm#Merging_two_lists) adapted to use channels.  The nil checks on the channels make it a little more verbose, but protect it from getting deadlocked.  The first version I had used a heap, but was slower.

jwilder:
@mark-rushakoff Removed.

PR #8226, opened 2017-03-29
Fix variable shadowing and simplify some logic.

matiasinsaurralde:
- [x] Rebased/mergable
- [x] Tests pass
- [x] [CLA](https://influxdata.com/community/cla/) signed


jsternberg:
I believe the variable shadowing is usually done on purpose to keep those variables in scope while not having to find new variable names. Can I see the linter errors you're talking about?

PR #8240, opened 2017-04-03
condense compactionGroups code

oiooj:
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/tsdb/index/tsi1/index.go
+++ b/tsdb/index/tsi1/index.go
@@ -802,21 +802,11 @@ func (i *Index) compactionGroups(fileSet FileSet) [][]*IndexFile {
 
 		// Skip over log files. They compact themselves.
 		if !ok {
-			if isCompactableGroup(group, i.CompactionFactor) {
-				group, groups = nil, append(groups, group)
-			} else {
-				group = nil
-			}
 			continue
 		}
 
 		// If file is currently compacting then stop current group.
 		if indexFile.Compacting() {
-			if isCompactableGroup(group, i.CompactionFactor) {
-				group, groups = nil, append(groups, group)
-			} else {
-				group = nil
-			}
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


tsdb/index/tsi1/index.go
```diff
@@ -802,21 +802,11 @@ func (i *Index) compactionGroups(fileSet FileSet) [][]*IndexFile {
 
 		// Skip over log files. They compact themselves.
 		if !ok {
-			if isCompactableGroup(group, i.CompactionFactor) {
-				group, groups = nil, append(groups, group)
-			} else {
-				group = nil
-			}
 			continue
 		}
```
benbjohnson: This section is needed because there can be `tsi` files before it reaches a log file. Those files need to be grouped in a compaction group.

tsdb/index/tsi1/index.go
```diff
@@ -802,21 +802,11 @@ func (i *Index) compactionGroups(fileSet FileSet) [][]*IndexFile {
...
 
 		// If file is currently compacting then stop current group.
 		if indexFile.Compacting() {
-			if isCompactableGroup(group, i.CompactionFactor) {
-				group, groups = nil, append(groups, group)
-			} else {
-				group = nil
-			}
 			continue
 		}
```
benbjohnson: This section is needed because we're looking for contiguous runs of non-compacting files. If we hit a file that is compacting then we need to make a compaction group with the previous files, if applicable.

tsdb/index/tsi1/index.go
```diff
@@ -829,11 +819,11 @@ func (i *Index) compactionGroups(fileSet FileSet) [][]*IndexFile {
...
 		group = append(group, indexFile)
-	}
 
-	// Append final group, if compactable.
-	if isCompactableGroup(group, i.CompactionFactor) {
-		groups = append(groups, group)
+		// Append group, if compactable.
+		if isCompactableGroup(group, i.CompactionFactor) {
+			group, groups = nil, append(groups, group)
+		}
```
benbjohnson: This section was outside the `for` loop to grab any remaining files in a compaction group.

By moving it inside the loop it is forming compaction groups as soon as a valid one exists instead of trying to find the longest contiguous group of files.

benbjohnson:
Closing this PR as it changes the compaction functionality.

PR #8243, opened 2017-04-04
fix close index file twice

oiooj:
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

Close index file twice will panic.

### DIFF
--- a/tsdb/index/tsi1/index.go
+++ b/tsdb/index/tsi1/index.go
@@ -876,7 +876,6 @@ func (i *Index) compactGroup(files []*IndexFile) {
 		log.Printf("%s: error creating compaction files: %s", IndexName, err)
 		return
 	}
-	defer f.Close()
 
 	srcIDs := joinIntSlice(IndexFiles(files).IDs(), ",")
 	log.Printf("%s: performing full compaction: src=%s, path=%s", IndexName, srcIDs, path)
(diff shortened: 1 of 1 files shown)


tsdb/index/tsi1/index.go
```diff
@@ -876,7 +876,6 @@ func (i *Index) compactGroup(files []*IndexFile) {
 		log.Printf("%s: error creating compaction files: %s", IndexName, err)
 		return
 	}
-	defer f.Close()
```
benbjohnson: Is this causing issues? `f.Close()` should be idempotent.

tsdb/index/tsi1/index.go
```diff
@@ -885,6 +884,7 @@ func (i *Index) compactGroup(files []*IndexFile) {
 	n, err := IndexFiles(files).WriteTo(f)
 	if err != nil {
 		log.Printf("%s: error compacting index files: src=%s, path=%s, err=%s", IndexName, srcIDs, path, err)
+		f.Close()
 		return
 	}
```
benbjohnson: Changing to an explicit close here has two issues: 1) Any panic before this will cause `f` to not get closed, and 2) If someone else adds a another return in this function there's a decent chance that they could forget to add `f.Close()` which will cause issues. The `defer` is a catch all for any `return`.

tsdb/index/tsi1/index.go
benbjohnson: @oiooj Can you show the panic that's occurring? Closing twice shouldn't be a problem.

oiooj: @benbjohnson  sry, my bad, Closing twice is not a problem.

PR #8248, opened 2017-04-04
fix index lock

oiooj:
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated


oiooj:
fix #7945 maybe.

benbjohnson:
@oiooj Removing this recheck for compactions will disable compactions in many cases. The check is done inside a goroutine so there's not a deadlock issue.

PR #8251, opened 2017-04-04
Fix delete time fields creating unparseable points

jwilder:
If a field was named time was written and was subsequently dropped,
it could leave a trailing comma in the series key causing it to fail
to be parseable in other parts of the code.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/models/points.go
+++ b/models/points.go
@@ -1919,7 +1919,9 @@ func (p *point) Delete() {
 	switch {
 	case p.it.end == p.it.start:
 	case p.it.end >= len(p.fields):
-		p.fields = p.fields[:p.it.start]
+		// Remove the trailing comma if there are more than one fields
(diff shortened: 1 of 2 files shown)


models/points.go
```diff
@@ -1919,7 +1919,13 @@ func (p *point) Delete() {
 	switch {
 	case p.it.end == p.it.start:
 	case p.it.end >= len(p.fields):
-		p.fields = p.fields[:p.it.start]
+		// If we're not deleting first field, trim the comma
```
dgnorton: Should this be if we're not deleting the last field?

jwilder: Switched to use `bytes.TrimSuffix`

PR #8258, opened 2017-04-05
Restrict fill(none) and fill(linear) to be usable only with aggregate queries

### LINKED ISSUE #8067: [Query Fuzz] Fill(none) in query hangs
This issue was found by fuzzing the database with a query generator tool. It may be possible to simplify the query to help identify the root cause.

##### Dataset
```
$ curl https://s3.amazonaws.com/noaa.water-database/NOAA_data.txt -o NOAA_data.txt
$ influx -import -path=NOAA_data.txt
```

##### Query

```
USE NOAA_water_database

SELECT  "pH" / "degrees", "degrees" FROM  "average_temperature" GROUP BY "pq" FILL(none)
```

##### Result

The query hangs. Heap remains very low, no CPU activity.

See attached `SIGQUIT` on the server.

If you remove the `FILL(none)` the query returns fine. If  you change the query to:
```
SELECT  "pH" / "degrees", "degrees" FROM  "average_temperature" GROUP BY "pq" FILL(null)
```

it also returns fine.

[1488220500085551134-freeze.txt](https://github.com/influxdata/influxdb/files/804771/1488220500085551134-freeze.txt)

jsternberg:
Fixes #8067.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/influxql/ast.go
+++ b/influxql/ast.go
@@ -1740,6 +1740,10 @@ func (s *SelectStatement) validate(tr targetRequirement) error {
 		return err
 	}
 
+	if err := s.validateFill(); err != nil {
+		return err
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


influxql/ast.go
```diff
@@ -2112,6 +2116,20 @@ func (s *SelectStatement) validateAggregates(tr targetRequirement) error {
 	return nil
 }
 
+// validateFill ensures that the fill option matches the query type.
+func (s *SelectStatement) validateFill() error {
+	info := newSelectInfo(s)
+	if len(info.calls) == 0 {
+		switch s.Fill {
```
dgnorton: Does it make sense to allow `fill(null)` and `fill(previous)`?

jsternberg: Yea, `fill(null)` is necessary anyway since that's the default, but it's also possible since `fill(null)` will always produce a point and doesn't require looking at future points. Same with `fill(previous)`. Both of them will give valid values. `fill(none)` is problematic because it doesn't produce any point (which causes the iterators to become unaligned) and `fill(linear)` requires reading a future point which auxiliary iterators can't do.

PR #8261, opened 2017-04-05
Fix LIMIT and OFFSET for certain aggregate queries

### LINKED ISSUE #8066: [Query Fuzz] LIMIT / OFFSET in query hangs
This issue was found by fuzzing the database with a query generator tool. It may be possible to simplify the query to help identify the root cause.

##### Dataset
```
$ curl https://s3.amazonaws.com/noaa.water-database/NOAA_data.txt -o NOAA_data.txt
$ influx -import -path=NOAA_data.txt
```

##### Query

```
USE NOAA_water_database

SELECT  "degrees",   SAMPLE("degrees", 2) FROM  "average_temperature" LIMIT 42747 OFFSET 323971
```

##### Result

The query hangs. Heap remains very low, no CPU activity.

See attached `SIGQUIT` on the server.

I have managed to simplify a query that triggers this issue to: 
```
SELECT  "degrees",   SAMPLE("degrees", 2) FROM  "average_temperature" LIMIT 1 OFFSET 1
```

Note that all the following queries return results, or at least return an empty result immediately:

```
SELECT  "degrees",   SAMPLE("degrees", 2) FROM  "average_temperature"
SELECT  "degrees",   SAMPLE("degrees", 2) FROM  "average_temperature" LIMIT 1
```

However, `SELECT  "degrees",   SAMPLE("degrees", 2) FROM  "average_temperature" OFFSET 1` also hangs the server. I note in this case the docs say that `OFFSET` should not be used without `LIMIT`, but as you can see above, using it with `LIMIT` still hangs the server.

[1488220454176671673-freeze.txt](https://github.com/influxdata/influxdb/files/804760/1488220454176671673-freeze.txt)

jsternberg:
When LIMIT and OFFSET were used with any functions that were not handled
directly by the query engine (anything other than count, max, min, mean,
first, or last), the input to the function would be limited instead of
receiving the full stream of values it was supposed to receive.

This also fixes a bug that caused the server to hang when LIMIT and
OFFSET were used with a selector. When using a selector, the limit and
offset should be handled before the points go to the auxiliary iterator
to be split into different iterators. Limiting happened afterwards which
caused the auxiliary iterator to hang forever.

Fixes #8066.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/tests/server_test.go
+++ b/tests/server_test.go
@@ -7934,6 +7934,64 @@ func TestServer_Query_Sample_Wildcard(t *testing.T) {
 	}
 }
 
+func TestServer_Query_Sample_LimitOffset(t *testing.T) {
+	t.Parallel()
+	s := OpenServer(NewConfig())
+	defer s.Close()
+
+	if err := s.CreateDatabaseAndRetentionPolicy("db0", newRetentionPolicySpec("rp0", 1, 0), true); err != nil {
+		t.Fatal(err)
+	}
+
+	writes := []string{
+		fmt.Sprintf(`cpu float=1,int=1i %d`, mustParseTime(time.RFC3339Nano, "2000-01-01T00:00:00Z").UnixNano()),
+		fmt.Sprintf(`cpu float=2,int=2i %d`, mustParseTime(time.RFC3339Nano, "2000-01-01T00:01:00Z").UnixNano()),
+		fmt.Sprintf(`cpu float=3,int=3i %d`, mustParseTime(time.RFC3339Nano, "2000-01-01T00:02:00Z").UnixNano()),
+	}
+
+	test := NewTest("db0", "rp0")
+	test.writes = Writes{
+		&Write{data: strings.Join(writes, "\n")},
+	}
+
+	test.addQueries([]*Query{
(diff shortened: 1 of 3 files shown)


tests/server_test.go
```diff
@@ -7930,6 +7930,68 @@ func TestServer_Query_Sample_Wildcard(t *testing.T) {
...
+		},
+		/*
+			&Query{
+				name:    "sample() with limit 1 offset 1",
+				params:  url.Values{"db": []string{"db0"}},
+				command: `SELECT sample(float, 3), int FROM cpu LIMIT 1 OFFSET 1`,
+				exp:     `{"results":[{"statement_id":0,"series":[{"name":"cpu","columns":["time","sample","int"],"values":[["2000-01-01T00:01:00Z",2,2]]}]}]}`,
+			},
+		*/
+	}...)
```
benbjohnson: Are these tests supposed to be commented out?

jsternberg: No, that's a good catch. I must have forgotten about putting them back.

PR #8272, opened 2017-04-07
Fix dropping fields creating data corruption

jwilder:
The Point is intended to be immutable after being parsed since it
is shared by several goroutines.  When dropping a field (e.g. time),
corrupted data can result if one goroutine is deleting the field
while another is marshaling the underlying byte slices.

To avoid this, the shard will just skip invalid fields and series
instead of trying to mutate them by deleting them.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated


jwilder:
I fixed this in 1.2 directly because the code in master has diverged quite a bit and requires a different fix now.

tsdb/shard.go
```diff
@@ -546,26 +546,27 @@ func (s *Shard) validateSeriesAndFields(points []models.Point) ([]models.Point,
...
-		// verify the tags and fields
+
+		// Drop any series w/ a "time" tag, these are illegal
 		tags := p.Tags()
+		skip = tags.Get(timeBytes) != nil
 		if v := tags.Get(timeBytes); v != nil {
-			s.logger.Info(fmt.Sprintf("dropping tag 'time' from '%s'\n", p.PrecisionString("")))
-			tags.Delete(timeBytes)
-			p.SetTags(tags)
+			dropped++
```
e-dard: Given `tags.Get` is `O(n)` you could argue that having a `var skip bool` and then setting `skip` to true inside this if block to reduce the number of calls to `tags.Get` would be worthwhile in some cases.

jwilder: Ah.  I forgot to remove L552.

PR #8279, opened 2017-04-10
Use go junit formatter to output tests in junit format on CircleCI

jsternberg:
- [x] Rebased/mergable
- [x] Tests pass


jsternberg:
Of note, this doesn't affect appveyor. There's a way to get this working with appveyor too, but it's incredibly confusing and not clearly documented IMO.

PR #8282, opened 2017-04-11
Fix influxql select tests

### LINKED ISSUE #8281: Fix the influxql select tests
### Bug report

The influxql package has bad tests for `influxql.Select`. The big problem is that many of them have a preconstructed input that is defined by the person who wrote the test because these tests don't use the underlying query engine. This sometimes results in the programmer who wrote them writing the order of the points incorrectly when other aspects of the query engine change such as when we did the subquery refactor.

As an example, look at the following test:

```go
func TestSelect_Min(t *testing.T) {
    var ic IteratorCreator
    ic.CreateIteratorFn = func(m *influxql.Measurement, opt influxql.IteratorOptions) (influxql.Iterator, error) {
        if m.Name != "cpu" {
            t.Fatalf("unexpected source: %s", m.Name)
        }
        if !reflect.DeepEqual(opt.Expr, MustParseExpr(`min(value)`)) {
            t.Fatalf("unexpected expr: %s", spew.Sdump(opt.Expr))
        }

        return influxql.NewCallIterator(&FloatIterator{Points: []influxql.FloatPoint{
            {Name: "cpu", Tags: ParseTags("region=west,host=A"), Time: 0 * Second, Value: 20},
            {Name: "cpu", Tags: ParseTags("region=west,host=B"), Time: 5 * Second, Value: 10},
            {Name: "cpu", Tags: ParseTags("region=east,host=A"), Time: 9 * Second, Value: 19},
            {Name: "cpu", Tags: ParseTags("region=east,host=A"), Time: 10 * Second, Value: 2},
            {Name: "cpu", Tags: ParseTags("region=west,host=A"), Time: 11 * Second, Value: 3},
            {Name: "cpu", Tags: ParseTags("region=west,host=A"), Time: 31 * Second, Value: 100},
        }}, opt)
    }

    // Execute selection.
    itrs, err := influxql.Select(MustParseSelectStatement(`SELECT min(value) FROM cpu WHERE time >= '1970-01-01T00:00:00Z' AND time < '1970-01-02T00:00:00Z' GROUP BY time(10s), host fill(none)`), &ic, nil)
    if err != nil {
        t.Fatal(err)
    } else if a, err := Iterators(itrs).ReadAll(); err != nil {
        t.Fatalf("unexpected point: %s", err)
    } else if !deep.Equal(a, [][]influxql.Point{
        {&influxql.FloatPoint{Name: "cpu", Tags: ParseTags("host=A"), Time: 0 * Second, Value: 19, Aggregated: 2}},
        {&influxql.FloatPoint{Name: "cpu", Tags: ParseTags("host=B"), Time: 0 * Second, Value: 10, Aggregated: 1}},
        {&influxql.FloatPoint{Name: "cpu", Tags: ParseTags("host=A"), Time: 10 * Second, Value: 2, Aggregated: 2}},
        {&influxql.FloatPoint{Name: "cpu", Tags: ParseTags("host=A"), Time: 30 * Second, Value: 100, Aggregated: 1}},
    }) {
        t.Fatalf("unexpected points: %s", spew.Sdump(a))
    }
}
```

The first issue here is that the order of the points going into the call iterator is incorrect and doesn't match what the query engine is expecting. The TSM1 engine wouldn't return the points in that order. Since there is a `GROUP BY host`, the point that is second should be last because `host=B` comes after `host=A` for that query. `host=B` would only come before the point at 10 or 30 seconds if it were being fed into a subquery.

A better way to do this that would test the query engine more thoroughly is to make more liberal use of `Merge()` in the query tests. Separate each series into their own iterator and then merge them instead of trying to guess what the order should be.

jsternberg:
The inputs are now sent to the tested iterator in the correct order so
we can more accurately test each individual select statement.

Fixes #8281.

- [x] Rebased/mergable
- [x] Tests pass

### DIFF
--- a/influxql/select_test.go
+++ b/influxql/select_test.go
@@ -25,14 +25,24 @@ func TestSelect_Min(t *testing.T) {
 			t.Fatalf("unexpected expr: %s", spew.Sdump(opt.Expr))
 		}
 
-		return influxql.NewCallIterator(&FloatIterator{Points: []influxql.FloatPoint{
-			{Name: "cpu", Tags: ParseTags("region=west,host=A"), Time: 0 * Second, Value: 20},
-			{Name: "cpu", Tags: ParseTags("region=west,host=B"), Time: 5 * Second, Value: 10},
-			{Name: "cpu", Tags: ParseTags("region=east,host=A"), Time: 9 * Second, Value: 19},
-			{Name: "cpu", Tags: ParseTags("region=east,host=A"), Time: 10 * Second, Value: 2},
-			{Name: "cpu", Tags: ParseTags("region=west,host=A"), Time: 11 * Second, Value: 3},
-			{Name: "cpu", Tags: ParseTags("region=west,host=A"), Time: 31 * Second, Value: 100},
-		}}, opt)
+		input, err := influxql.Iterators{
+			&FloatIterator{Points: []influxql.FloatPoint{
+				{Name: "cpu", Tags: ParseTags("region=west,host=A"), Time: 0 * Second, Value: 20},
+				{Name: "cpu", Tags: ParseTags("region=west,host=A"), Time: 11 * Second, Value: 3},
+				{Name: "cpu", Tags: ParseTags("region=west,host=A"), Time: 31 * Second, Value: 100},
+			}},
+			&FloatIterator{Points: []influxql.FloatPoint{
+				{Name: "cpu", Tags: ParseTags("region=east,host=A"), Time: 9 * Second, Value: 19},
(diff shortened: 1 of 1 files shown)


jsternberg:
Looks like there are more of these kinds of tests in `influxql/call_iterator_test.go`. I'll be working on modifying those in the same method.

jsternberg:
The call iterator stuff is more complicated and likely will need to be changed as I try to refactor some of the ordering code. The ordering code, while it works, likely doesn't give us as good of performance as it could.

Issue #3691, opened 2015-08-17
[feature request] Should support ceil and floor functions


pauldix:
Some users want to round results from math and aggregates. These functions should act like transformation functions (like derivative). For example say you have a series where you're reporting uptime in seconds for devices and you want to show how many days it has been up.

```sql
-- ceil rounds up, floor rounds down
-- the first argument is the thing you're taking the ceiling of
-- this would give you as many data points as you have uptime reports, many of which
-- would be the same because of the large divisor
select ceil(value / 86400) from uptime where time > now() - 30d

-- if you're reporting uptime frequently, you may want to limit the number of points here
-- this would give you only 30 data points back
select ceil(last(value) / 86400) from uptime where time > now() - 30d group by time(1d) 

-- ceil and floor should support an optional second argument that specifies the precision to round to
select ceil(value / 86400, .2f) from uptime where time > now() - 30d
```

Just my initial thoughts on this. Feedback welcome.

laf:
Thanks @pauldix 

Exactly what we're looking for :)

musl:
@pauldix Great idea. This will make some history graphs easier to read - especially for those targeted at an audience that's not expected to be well versed in statistics. I think it'll also help with clarity and style in other cases.

toddboom:
Also adding a request for `round()` in response to #5435.

beckettsean:
As mentioned in my [post to the mailing list](https://groups.google.com/d/msg/influxdb/CkKzZoXDgiU/5E-vDNVGCwAJ) we are experimenting with simplifying our open GitHub Issues. This feature request has been rolled into an [aggregate issue](https://github.com/influxdata/influxdb/issues/5930) for all function requests, so that we can close this issue until we are ready to work on it. 

You may continue to make comments here. Closing the issue does not mean we are rejecting this idea.

MikiDi:
Hi,
I would like to have a go at implementing the `CEIL()` and `FLOOR()` functions.
However, I'm new to the project ... could anyone give some pointers on what to look at first, which special cases to consider, etc? :-)

Michaël

timhallinflux:
closed via #9632

Issue #659, opened 2014-06-18
Support more math (power, logarithm, trigonometric functions)


jordens:
Basic math operations in queries, like a power operator, square roots, some logarithm, exponentials and trigonometric functions would be great.
I would especially like to see power and logarithms. They would come in handy especially with #438 . I would like to be able to do:

`select ain0**2 from /.*/ limit 1`

`select log10(pressure) from /.*/ limit 1`

eliperelman:
Adding my +1 behind this.

toddboom:
@eliperelman any functions in particular that you're needing?

eliperelman:
I'm specifically needing square root so I can calculate upper bounds:

`select mean(value) + (1.96 * stddev(value) / sqrt(count(value)) as "upperBound" from measure ...`

eliperelman:
Although having an upper bound function might be more succinct. :)

`select upperbound(value, 95) from myseries`

beckettsean:
As mentioned in my [post to the mailing list](https://groups.google.com/d/msg/influxdb/CkKzZoXDgiU/5E-vDNVGCwAJ) we are experimenting with simplifying our open GitHub Issues. This feature request has been rolled into an [aggregate issue](https://github.com/influxdata/influxdb/issues/5930) for all function requests, so that we can close this issue until we are ready to work on it. 

You may continue to make comments here. Closing the issue does not mean we are rejecting this idea.

PR #8283, opened 2017-04-11
Implement value-transform functions such as sin() and abs().

Tomcat-Engineering:
This adds 17 "value transforms" to InfluxQL, where these are defined as functions which operate on the value of one datapoint at a time, leaving the timestamp unaffected.

The full list of functions added is: `abs`, `sin`, `cos`, `tan`, `asin`, `acos`, `atan`, `log`, `log10`, `exp`, `sqrt`, `rad2deg`, `deg2rad`, `pow`, `ceiling`, `floor`, `round`.

All of these work for float and integer datatypes.  You get integer points back where possible (e.g. `round(my_float)` or `abs(my_int)`), otherwise you get float points.

Internally, they are called via the Auxiliary Value route rather than as standard Calls (which are basically for aggregators, selectors and other things which don't necessarily return a new value for each input point).  This means that they play nicely with multi-field queries, selectors, nesting and subqueries; the following queries on the standard dataset all produce the expected results:

```
SELECT water_level, 3 * abs(water_level - 100.4) AS fred FROM h2o_feet LIMIT 3

SELECT max(water_level), sqrt(abs(water_level)) FROM h2o_feet

SELECT -4 * abs(fred) FROM (SELECT mean(water_level) AS fred FROM h2o_feet)

SELECT mean(*) FROM (SELECT abs(water_level) FROM h2o_feet)
```

More unit tests are required, but implementing the tests seems to be harder than implementing the code, particularly for more complex queries like the ones above.  Any pointers or examples welcome.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x] Provide example syntax
- [x] [InfluxData Documentation] https://github.com/influxdata/docs.influxdata.com/pull/1097

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - [#7393](https://github.com/influxdata/influxdb/issues/7393): Add "non_negative_difference" function to InfluxQL.
 - [#8042](https://github.com/influxdata/influxdb/issues/8042): Add bitwise AND, OR and XOR operators to the query language.
 - [#8302](https://github.com/influxdata/influxdb/pull/8302): Write throughput/concurrency improvements
+- [#659](https://github.com/influxdata/influxdb/issues/659): Add value-transform functions to the query language (abs, sin, cos, tan, asin, acos, atan, log, log10, exp, sqrt, rad2deg, deg2rad, pow, ceiling, floor, round)
 
 ### Bugfixes
 
--- a/influxql/ast.go
+++ b/influxql/ast.go
@@ -1761,6 +1761,10 @@ func (s *SelectStatement) validate(tr targetRequirement) error {
 		return err
 	}
 
+	if err := s.validateValueTransforms(); err != nil {
+		return err
+	}
+
 	if err := s.validateAggregates(tr); err != nil {
 		return err
 	}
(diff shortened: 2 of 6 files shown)


Tomcat-Engineering:
Addresses parts of #5930 #659 #3691.

fdhex:
Anyone looked at this? @rbetts ?

rbetts:
We'll add this to the work queue for review. However, in the short term it needs to be rebased and it needs test cases for argument and type validation to the functions that are added.

jsternberg:
Closing this in favor of #9424. @Tomcat-Engineering I'm sorry it took so long for this feature and for not properly looking at this PR, but we refactored the engine to make this kind of functionality much easier to validate and implement. #9424 implements only the basic trigonometry functions, but we would be very interested in adding more value transform functions if you have the time. Please refer to that PR to see the new method. Thanks.

jsternberg:
#9499 is the PR. #9424 is the issue.

Tomcat-Engineering:
@jsternberg no worries, glad the feature made it in eventually, and good that you came up with a neater way of integrating with the engine.  I might have a look at adding some of the other value transforms if I have time... I assume that the mechanism for implementing value transforms is now considered stable enough to add more?

jsternberg:
Yes. I would consider it stable enough.

PR #8290, opened 2017-04-12
TSI Compaction Refactor

benbjohnson:
## Overview

This changes the `TagBlock` & `MeasurementBlock` to store series ID lists as delta-encoded uvarints. This reduces the storage size significantly. Previously my tests were showing a 461b per series (for a 130M series dataset). Now tests are showing 225b per series (for a 430M series dataset).

The total size can still be reduced further but this is a good start.

- [x] Rebased/mergable
- [x] Tests pass

### DIFF
--- a/tsdb/index/tsi1/file_set.go
+++ b/tsdb/index/tsi1/file_set.go
@@ -23,16 +23,18 @@ type FileSet struct {
 }
 
 // NewFileSet returns a new instance of FileSet.
-func NewFileSet(levels []CompactionLevel, files []File) *FileSet {
+func NewFileSet(levels []CompactionLevel, files []File) (*FileSet, error) {
 	fs := &FileSet{levels: levels, files: files}
-	fs.buildFilters()
-	return fs
+	if err := fs.buildFilters(); err != nil {
+		return nil, err
+	}
+	return fs, nil
 }
 
 // Close closes all the files in the file set.
 func (p FileSet) Close() error {
 	var err error
-	for _, f := range p {
+	for _, f := range p.files {
 		if e := f.Close(); e != nil && err == nil {
 			err = e
 		}
@@ -55,7 +57,7 @@ func (fs *FileSet) Release() {
 }
 
 // Prepend returns a new file set with f added at the beginning.
-func (fs *FileSet) Prepend(f File) *FileSet {
+func (fs *FileSet) Prepend(f File) (*FileSet, error) {
 	return NewFileSet(fs.levels, append([]File{f}, fs.files...))
 }
 
@@ -87,7 +89,11 @@ func (fs *FileSet) MustReplace(oldFiles []File, newFile File) *FileSet {
 	other[i] = newFile
 	copy(other[i+1:], fs.files[i+len(oldFiles):])
 
-	return NewFileSet(fs.levels, other)
+	fs, err := NewFileSet(fs.levels, other)
+	if err != nil {
+		panic("cannot build file set: " + err.Error())
+	}
+	return fs
 }
 
 // MaxID returns the highest file identifier.
@@ -99,6 +105,11 @@ func (fs *FileSet) MaxID() int {
 		}
 	}
 	return max
+}
+
+// Files returns all files in the set.
+func (fs *FileSet) Files() []File {
+	return fs.files
 }
 
 // LogFiles returns all log files from the file set.
@@ -849,10 +860,10 @@ func (fs *FileSet) seriesByBinaryExprVarRefIterator(name, key []byte, value *inf
 }
 
 // buildFilters builds a series existence filter for each compaction level.
-func (fs *FileSet) buildFilters() {
+func (fs *FileSet) buildFilters() error {
 	if len(fs.levels) == 0 {
 		fs.filters = nil
-		return
+		return nil
 	}
 
 	// Generate filters for each level.
@@ -875,9 +886,11 @@ func (fs *FileSet) buildFilters() {
 
 		// Merge filter.
 		if err := fs.filters[level].Merge(f.Filter()); err != nil {
-			panic(err)
+			return err
 		}
 	}
+
+	return nil
 }
 
 // File represents a log or index file.
--- a/tsdb/index/tsi1/file_set.go
+++ b/tsdb/index/tsi1/file_set.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/influxdata/influxdb/influxql"
 	"github.com/influxdata/influxdb/models"
+	"github.com/influxdata/influxdb/pkg/bloom"
 	"github.com/influxdata/influxdb/pkg/bytesutil"
 	"github.com/influxdata/influxdb/pkg/estimator"
 	"github.com/influxdata/influxdb/pkg/estimator/hll"
@@ -15,7 +16,18 @@ import (
 )
 
 // FileSet represents a collection of files.
-type FileSet []File
+type FileSet struct {
+	levels  []CompactionLevel
+	files   []File
+	filters []*bloom.Filter // per-level filters
+}
+
+// NewFileSet returns a new instance of FileSet.
+func NewFileSet(levels []CompactionLevel, files []File) *FileSet {
+	fs := &FileSet{levels: levels, files: files}
+	fs.buildFilters()
+	return fs
+}
 
 // Close closes all the files in the file set.
 func (p FileSet) Close() error {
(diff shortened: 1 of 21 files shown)


tsdb/index/tsi1/measurement_block.go
```diff
@@ -321,16 +326,17 @@ func (e *MeasurementBlockElem) TagBlockOffset() int64 { return e.tagBlock.offset
...
-func (e *MeasurementBlockElem) SeriesID(i int) uint64 {
-	return binary.BigEndian.Uint64(e.series.data[i*SeriesIDSize:])
-}
-
 // SeriesIDs returns a list of decoded series ids.
 func (e *MeasurementBlockElem) SeriesIDs() []uint64 {
-	a := make([]uint64, e.series.n)
-	for i := 0; i < int(e.series.n); i++ {
-		a[i] = e.SeriesID(i)
+	a := make([]uint64, 0, e.series.n)
```
jwilder: Could this end up creating a huge slice allocation?  E.g. if you had 1B series for a measurement, would this try to allocation 8GB?

tsdb/index/tsi1/measurement_block.go
```diff
@@ -321,16 +326,17 @@ func (e *MeasurementBlockElem) TagBlockOffset() int64 { return e.tagBlock.offset
...
+	var prev uint64
+	for data := e.series.data; len(data) > 0; {
+		delta, n := binary.Uvarint(data)
```
jwilder: What is the average delta size?

tsdb/index/tsi1/index.go
```diff
@@ -27,7 +27,7 @@ const IndexName = "tsi1"
 // Default compaction thresholds.
 const (
 	DefaultMaxLogFileSize   = 5 * 1024 * 1024
-	DefaultCompactionFactor = 1.8
+	DefaultCompactionFactor = 10 // 1.8
```
jwilder: Can you add a comment to this constant?  The value and commented out `// 1.8` are very cryptic.

tsdb/index/tsi1/measurement_block.go
e-dard: Does this method make sense for high cardinality situations? Feels like we should be exposing an iterator rather than allocating the entire list of ids.

tsdb/index/tsi1/tag_block.go
```diff
@@ -316,16 +316,17 @@ func (e *TagBlockValueElem) Value() []byte { return e.value }
...
-func (e *TagBlockValueElem) SeriesID(i int) uint64 {
-	return binary.BigEndian.Uint64(e.series.data[i*SeriesIDSize:])
-}
-
 // SeriesIDs returns a list decoded series ids.
 func (e *TagBlockValueElem) SeriesIDs() []uint64 {
-	a := make([]uint64, e.series.n)
-	for i := 0; i < int(e.series.n); i++ {
-		a[i] = e.SeriesID(i)
+	a := make([]uint64, 0, e.series.n)
```
e-dard: Same question regarding memory size when we have a lot of series sharing the same tag value. Could we end up with huge allocations here?

tsdb/index/tsi1/measurement_block.go
benbjohnson: This function is just used for testing. I can remove it so it doesn't get used by the main code accidentally.

tsdb/index/tsi1/measurement_block.go
benbjohnson: For the measurement block, the average delta should be the average length of the series key.

tsdb/index/tsi1/tag_block.go
benbjohnson: This is also just used for testing. I'll move it into the `_test` package.

tsdb/index/tsi1/measurement_block.go
benbjohnson: I added a `NOTE` about this function in 26effda. It was a pain to move it only to the `_test` repo.

benbjohnson:
@jwilder I ended up building on the original delta encoding when I refactored the compaction so this pull request has grown quite a bit. This new code adds memory-mapped bloom filters to the index files and distinct compaction levels. Each compaction level has the same bloom filter bit size (`m`) and function count (`k`) so they get merged together which reduces the lookups. I also reduced the number of files merged at once to speed up compaction time.

The new code seems to double the series creation speed on the machines I've tested it on. There are still some additional write optimizations that can be done but I'm going to move to read optimizations for now.

tsdb/index/tsi1/file_set.go
```diff
@@ -765,12 +837,46 @@ func (fs FileSet) seriesByBinaryExprVarRefIterator(name, key []byte, value *infl
...
+
+		// Initialize a filter if it doesn't exist.
+		if fs.filters[level] == nil {
+			lvl := fs.levels[level]
+			fs.filters[level] = bloom.NewFilter(lvl.M, lvl.K)
+		}
+
+		// Merge filter.
+		if err := fs.filters[level].Merge(f.Filter()); err != nil {
+			panic(err)
```
jwilder: Should return err here.

tsdb/index/tsi1/index.go
```diff
@@ -756,102 +769,68 @@ func (i *Index) compact() {
...
-
-		// If file is currently compacting then stop current group.
-		if indexFile.Compacting() {
-			if isCompactableGroup(group, i.CompactionFactor) {
-				group, groups = nil, append(groups, group)
-			} else {
-				group = nil
-			}
+		if size < i.levels[level+1].MinSize {
+			// log.Printf("tsi1: SKIP, too small: level=%d, size=%d, target=%d", level, size, i.levels[level+1].MinSize)
```
jwilder: Remove?

tsdb/index/tsi1/measurement_block.go
jwilder: I was asking more about the range of values.  If it's < 128, varint encoding would use 2 bytes when 1 of a few bits are sufficient.  My guess is that series keys won't differ by a large range very frequently (since I believe they are sorted) so I think using a different encoding might save a lot of space here.

I think you mentioned offline that you were going to switch series ids back to `uint32`?  Is that still planned?

tsdb/index/tsi1/file_set.go
benbjohnson: Fixed in 2524df3.

tsdb/index/tsi1/index.go
benbjohnson: Fixed in 2524df3.

benbjohnson:
@jwilder I rebased, fixed up some of the changes, and converted the series IDs back from `uint64` to `uint32`.

tsdb/index/tsi1/index.go
```diff
@@ -762,102 +783,67 @@ func (i *Index) compact() {
...
-			f.Retain()
-			f.setCompacting(true)
+	// Iterate over each level we are going to compact.
+	// We skip the first level (0) because it is log files and they are compacted separately.
+	// We skip the last level because the files have no higher level to compact into.
+	minLevel, maxLevel := 1, len(i.levels)-2
+	for level := minLevel; level <= maxLevel; level++ {
+		// Skip level if it is currently compacting.
+		if i.levelCompacting[level] {
+			// log.Printf("tsi1: SKIP, already compacting: level=%d", level)
```
jwilder: Remove?

tsdb/index/tsi1/index.go
```diff
@@ -1036,10 +1022,9 @@ func (i *Index) compactLogFile(logFile *LogFile) {
 		log.Printf("%s: error updating manifest: %s", IndexName, err)
 		return
 	}
-	log.Printf("%s: finished compacting log file: file=%s, t=%v, sz=%d", IndexName, logFile.Path(), time.Since(start), n)
+	log.Printf("%s: log file compacted: file=%s, t=%0.03fs", IndexName, filepath.Base(logFile.Path()), time.Since(start).Seconds())
```
jwilder: I realized TSI types need a logger instance so they use the same format as the rest of the code.  Something to fix in another PR I think.

tsdb/index/tsi1/index.go
benbjohnson: Removed in 48456d8.

tsdb/index/tsi1/index.go
benbjohnson: I'll fix the logging in a separate PR.

PR #8302, opened 2017-04-19
Write throughput/concurrency improvements

jwilder:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

This PR has a few performance improvements and bug fixes to the write path which improve overall write throughput when there are many concurrent writers.

The graphs below show the write latency and throughput changes from 1.2.3 to this PR (based off of 1.3) using batches of 5k and 10k and varying concurrent writers from 50-200 in each run.  Each of the humps is a write load of 500M values over 100k series with varying concurrency and batch sizes.  The top row is 5k batches.  Second row is 10k batches.  Left column is 1.2 and right column is this PR.  These were run a m4.16xlarge w/ 1000 IOPS SSDs using go1.8.1 and `inmem` and `tsi1` index.

For 5k batches, write throughput improved from 1.1M w/s to 2.3M w/s and latency decreased by ~50% and stayed below 500ms after the change.

For 10k batches, write throughput improved from 1.8M w/s to 2.3M w/s and latency decreased by ~30%.

The majority of the changes focus on reducing lock contention and allocations in the hot path.

The throughput was remarkably similar between `inmem` and `tsi`.  `tsi` on master hit write timeouts under higher concurrency immediately when the run started.  With this PR those, timesout were fixed as well.

### inmem

perf

### tsi

tsi

### DIFF
--- a/tsdb/engine/tsm1/pools.go
+++ b/tsdb/engine/tsm1/pools.go
@@ -1,15 +1,26 @@
 package tsm1
 
-import "github.com/influxdata/influxdb/pkg/pool"
+import "sync"
 
-var bufPool = pool.NewBytes(10)
+var bufPool sync.Pool
 
 // getBuf returns a buffer with length size from the buffer pool.
-func getBuf(size int) []byte {
-	return bufPool.Get(size)
+func getBuf(size int) *[]byte {
+	x := bufPool.Get()
+	if x == nil {
+		b := make([]byte, size)
+		return &b
+	}
+	buf := x.(*[]byte)
+	if cap(*buf) < size {
+		b := make([]byte, size)
+		return &b
+	}
+	*buf = (*buf)[:size]
+	return buf
 }
 
 // putBuf returns a buffer to the pool.
-func putBuf(buf []byte) {
+func putBuf(buf *[]byte) {
 	bufPool.Put(buf)
 }
--- a/tsdb/engine/tsm1/wal.go
+++ b/tsdb/engine/tsm1/wal.go
@@ -382,16 +382,16 @@ func (l *WAL) writeToLog(entry WALEntry) (int, error) {
 	//defer l.limiter.Release()
 
 	// encode and compress the entry while we're not locked
-	bytes := getBuf(walEncodeBufSize)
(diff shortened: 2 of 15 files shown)


tsdb/engine/tsm1/pools.go
```diff
@@ -1,12 +1,20 @@
...
+var bufPool sync.Pool
 
 // getBuf returns a buffer with length size from the buffer pool.
 func getBuf(size int) []byte {
-	return bufPool.Get(size)
+	x := bufPool.Get()
+	if x == nil {
+		return make([]byte, size)
+	}
+	buf := x.([]byte)
```
nussjustin: Correct me if I'm wrong, but putting a []byte (or any slice) into a sync.Pool will allocate as the conversion from slice to interface needs to allocate space for the slice on the heap. This can be avoided by pooling *[]byte instead. For example:

```go
// getBuf returns a buffer with length size from the buffer pool.
func getBuf(size int) *[]byte {
	x := bufPool.Get()
	if x == nil {
		b := make([]byte, size)
		return &b
	}
	buf := x.(*[]byte)
	if cap(*buf) < size {
		b := make([]byte, size)
		return &b
	}
	*buf = (*buf)[:size]
	return buf
}

// putBuf returns a buffer to the pool.
func putBuf(buf *[]byte) {
	bufPool.Put(buf)
}
```

A quick local benchmark shows that this avoids an allocation.

tsdb/engine/tsm1/wal.go
```diff
@@ -121,7 +121,7 @@ func NewWAL(path string) *WAL {
 		// these options should be overriden by any options in the config
 		SegmentSize: DefaultSegmentSize,
 		closing:     make(chan struct{}),
-		syncWaiters: make(chan chan error, 256),
+		syncWaiters: make(chan chan error, 1024),
```
alexdaviesjump: <nit> Commit message says "This increases the buffer to 4096"

jwilder: Fixed.

jwilder:
Added `tsi` tests and fixes.

PR #8305, opened 2017-04-19
Optimize WAL segment reading

jwilder:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

Reading WAL segments generates a lot of garbage and was doing some things inefficiently.  This change improves reading segments a bit.  

These changes help a little, but most of the garbage (and time) is still dominated by `tsm1.Value` interface and conversion when re-creating the type specific values to be added back to the `tsm1.Cache`.  Fixing that is a bit more complicated.

```
benchmark                       old ns/op     new ns/op     delta
BenchmarkWALSegmentReader-8     40012977      23665143      -40.86%

benchmark                       old allocs     new allocs     delta
BenchmarkWALSegmentReader-8     1000500        500601         -49.96%

benchmark                       old bytes     new bytes     delta
BenchmarkWALSegmentReader-8     24236300      16236804      -33.01%
```

### DIFF
--- a/tsdb/engine/tsm1/wal.go
+++ b/tsdb/engine/tsm1/wal.go
@@ -743,31 +743,28 @@ func (w *WriteWALEntry) UnmarshalBinary(b []byte) error {
 
 		switch typ {
 		case float64EntryType:
+			if i+16*nvals > len(b) {
+				return ErrWALCorrupt
+			}
+
 			values := make([]Value, 0, nvals)
 			for j := 0; j < nvals; j++ {
-				if i+16 > len(b) {
-					return ErrWALCorrupt
-				}
-
 				un := int64(binary.BigEndian.Uint64(b[i : i+8]))
 				i += 8
-
 				v := math.Float64frombits((binary.BigEndian.Uint64(b[i : i+8])))
 				i += 8
-
 				values = append(values, NewFloatValue(un, v))
 			}
 			w.Values[k] = values
 		case integerEntryType:
+			if i+16*nvals > len(b) {
+				return ErrWALCorrupt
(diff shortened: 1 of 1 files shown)


tsdb/engine/tsm1/wal.go
```diff
@@ -709,87 +710,69 @@ func (w *WriteWALEntry) UnmarshalBinary(b []byte) error {
...
+			for j := 0; j < nvals; j++ {
+				if i+12 > len(b) {
 					return ErrWALCorrupt
 				}
 
+				un := int64(binary.BigEndian.Uint64(b[i : i+8]))
+				i += 8
+
 				length := int(binary.BigEndian.Uint32(b[i : i+4]))
 				if i+length > int(uint32(len(b))) {
```
joelegasse: Can we remove the double-cast while we're in here? `leb(b)` is already an `int` :-)

tsdb/engine/tsm1/wal.go
```diff
@@ -709,87 +710,69 @@ func (w *WriteWALEntry) UnmarshalBinary(b []byte) error {
...
-
-			un := int64(binary.BigEndian.Uint64(b[i : i+8]))
-			i += 8
-
-			switch typ {
-			case float64EntryType:
-				if i+8 > len(b) {
+			values := make([]Value, 0, nvals)
+			for j := 0; j < nvals; j++ {
+				if i+16 > len(b) {
```
joelegasse: After seeing the rest of the code, it doesn't seem like a partial read is done, as the `values` slice is never stored.

PR #8336, opened 2017-04-28
prefix partial write errors with `partial write:`

stuartcarnie:
NOTE: parser errors (via http API) are also transformed into
PartialWriteError

- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/tsdb/shard.go
+++ b/tsdb/shard.go
@@ -554,7 +554,7 @@ func (s *Shard) validateSeriesAndFields(points []models.Point) ([]models.Point, 
 		if v := tags.Get(timeBytes); v != nil {
 			dropped++
 			if reason == "" {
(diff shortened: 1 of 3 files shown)


tsdb/shard.go
```diff
@@ -553,6 +553,9 @@ func (s *Shard) validateSeriesAndFields(points []models.Point) ([]models.Point,
 		tags := p.Tags()
 		if v := tags.Get(timeBytes); v != nil {
 			dropped++
+			if reason == "" {
+				reason = fmt.Sprintf("invalid field name: input tag \"%s\" on measurement \"%s\" is invalid", "time", p.Name())
```
jwilder: Should probably be `invalid tag key` since this is a tag.

stuartcarnie: Updated both messages; when you accept, I'll update the Godeps in the plutonium PR as this should end up being a fast-forward

PR #8337, opened 2017-04-28
Add TSM import capabilities

jwilder:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

This adds some of the lower level bits to import a shard while the `tsdb.Store` is open.  There are also a few other clean ups and exposing of functions that are needed by external tools.


tsdb/engine/tsm1/engine.go
```diff
@@ -612,73 +612,146 @@ func (e *Engine) writeFileToBackup(name string, shardRelativePath, fullPath stri
...
 
+	// Load any new series keys to the index
+	readers := make([]chan seriesKey, 0, len(newFiles))
+	for _, f := range newFiles {
+		ch := make(chan seriesKey, 1)
+		readers = append(readers, ch)
+
+		if strings.HasSuffix(f, ".tmp") {
+			f = strings.TrimSuffix(f, ".tmp")
+		}
```
benbjohnson: Why would some files have a `.tmp` extension?

jwilder: When importing as new files they are written into the shard as `.tmp` and the `FileStore.Replace` renames them when they are installed.

tsdb/engine/tsm1/engine.go
```diff
@@ -612,73 +612,146 @@ func (e *Engine) writeFileToBackup(name string, shardRelativePath, fullPath stri
...
+		defer r.Close()
+
+		go func(c chan seriesKey, r *TSMReader) {
+			n := r.KeyCount()
+			for i := 0; i < n; i++ {
+				key, typ := r.KeyAt(i)
+				c <- seriesKey{key, typ}
+			}
+			close(ch)
+		}(ch, r)
```
benbjohnson: The `close(ch)` references the `ch` outside the closure while the arg passed in is called `c`. I don't think it should cause an issue since `ch` is defined inside the `for` loop but I think it could lead to subtle bugs if the `ch` definition is moved outside the loop at some point.

tsdb/engine/tsm1/engine.go
benbjohnson: It doesn't seem like you need the `HasSuffix()` at all. You can just call `TrimSuffix()` since it'll be a no-op if the suffix doesn't exist.

tsdb/engine/tsm1/engine.go
jwilder: Fixed.

tsdb/engine/tsm1/engine.go
jwilder: Good point. Updated.

PR #8344, opened 2017-05-01
Set the CSV output to an empty string for null values

### LINKED ISSUE #8343: CSV output uses previous point value when it should be a null
### Bug report
X-Influxdb-Version →1.2.2

__Steps to reproduce:__

1. `/query?q=DROP DATABASE csv_test`
2. `/query?q=CREATE DATABASE csv_test`
3. `/write?db=csv_test`
```
example a=1,b=2 1465839830100400000 
example a=1 1465839830100500000 
```

__Expected behavior:__ 
I would expect the second point to have a null value for 'b' as it does in the json response:
```
"values":[["2016-06-13T17:43:50.1004Z",1,2],["2016-06-13T17:43:50.1005Z",1,null]]
```
__Actual behavior:__ 
JSON format gives the expected results, but CSV does not

`/query?db=csv_test&q=SELECT * FROM example`
Gives:
```
name,tags,time,a,b,c
example,,1465839830100400000,1,2
example,,1465839830100500000,1,2
```

jsternberg:
Fixes #8343.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


ryantxu:
thanks!

Issue #8276, opened 2017-04-09
Limit number of concurrent compactions


hpbieker:
Proposal:
I would like InfluxDB to have a configuration setting to limit the number of concurrent compactions.

Current behavior:
InfluxDB starts as many compactions as it can concurrently. It is not limited to the number of MAXGOPROCS as far as I understand.

Desired behavior:
The user can limit the number of compactions running at the same time.

Use case:
On my system I had 170 full compactions running concurrently after a restart of InfluxDB as all of them are basically started at the same time. This caused the machine to run out of memory and starts swapping. Thuse it because useless. It might also cause the machine to run out of disk space.

I have 16 cores and 64 GB RAM. 1500 shards distributed on ~10 databases. 1.7 TB data.

By limiting the number of concurrent compactions, these situations can be prevented.

jwilder:
Can you attach some profile when this is occurring? Also, what version are you running?

```
curl -o block.txt "http://localhost:8086/debug/pprof/block?debug=1" 
curl -o goroutine.txt "http://localhost:8086/debug/pprof/goroutine?debug=1" 
curl -o heap.txt "http://localhost:8086/debug/pprof/heap?debug=1" 
curl -o vars.txt "http://localhost:8086/debug/vars" 
iostat -xd 1 30 > iostat.txt
influx -execute "show shards" > shards.txt
influx -execute "show stats" > stats.txt
influx -execute "show diagnostics" > diagnostics.txt
```

hpbieker:
Hi,
Fortuantly we were able to "fix" this by ensuring that only a ~10 shards that needed compacting had write permission, and then restarting and updating the permission once it was done processing the shards. I am therefore not able to provide these stats.

hpbieker:
Here is a copy of (a part of) the log which shows that very many compactions are started at the same time:
https://gist.github.com/hpbieker/c2107428fd5dc787b359c0e023397efd

This can easyly be reproduced by:
* Disable full compaction by setting "compact-full-write-cold-duration" to a large value
* Have say 100 large shards (~2 GB each)
* Write one or more point to each of the shards
* Wait for the non full compactions to complete their jobs
* Then change the "compact-full-write-cold-duration" to "1m" and restart Influx

Influx will then start to do a full compaction of all 100 shards at once. Each compaction may take a few minutes so they do not have the time to complete before others are started, and the system will start trashing (100 % utilization of swap media) or you will get a OOM.

hpbieker:
[...]

Issue #8123, opened 2017-03-12
tsm compaction does not remove .tmp on error (e.g. disk full)


hpbieker:
System info: InfluxDB 1.2.1, Linux

**Actual behavior:**
When tsm compaction failes because of disk full, it reports:
Mar 12 07:58:20 PPS-UbuntuLTS influxd[11583]: [I] 2017-03-12T06:58:20Z error compacting TSM files: write /media/db_disk/data/XXXX/autogen/84/000000294-000000003.tsm.tmp: no space left on device engine=tsm1

But the file is not removed (three hours later):
$ sudo ls -l /media/db_disk/data/XXXX/autogen/84/000000294-000000003.tsm.tmp
-rw-r--r-- 1 influxdb influxdb 125829120 Mar 12 10:59 /media/db_disk/data/history/autogen/84/000000294-000000003.tsm.tmp

**Desired behaviour:**
The .tmp file from the compaction should be removed. (It is removed when Influx is restarted.) This is especially a problem because InfluxDB tries to compact many directories at the same time, which eats up the disk space.

Looking at the comment line 734-737 (https://github.com/influxdata/influxdb/blob/549b7adf35512e8b1fa9f7649ac2883ec134b4c2/tsdb/engine/tsm1/compact.go#L734) in tsdb/engine/tsm1/compact.go it looks like it is supposed to remove the file, but the actual implementation does not.
		// We hit an error but didn't finish the compaction.  Remove the temp file and abort.
		if err != nil {
			return nil, err
		}

hpbieker:
I guess the files in the "files" list should be deleted in addition to the file in "fileName".

jwilder:
Fixed via #8348

Issue #7640, opened 2016-11-18
Influxdb crashes after 2 hours


rubycut:
### Bug report

__System info:__ 
Influxdb 1.1.0
Os: Debian

__Steps to reproduce:__

1. Start influxdb
2. Create 500 different databases with 500 users
3. Auth should be on

__Expected behavior:__ 

Should run normally

__Actual behavior:__ 

Crashes after two hours

rubycut:
Crash exception is over 70Mb:

https://www.dropbox.com/s/hn2sde9liw6dbo7/error.log?dl=0

rubycut:
Another link, maybe this will work better:

https://www.dropbox.com/s/1nqioi5pf18gz0q/error2.log?dl=0

rubycut:
At the time of crash, there was 114k goroutines in the system:
```
grep "goroutine " error2.log|wc -l
114551
```

Roughly half of them were stuck at the same place:
```
grep engine.go:979 error2.log|wc -l
56118
```

```
goroutine 19985 [sleep]:
time.Sleep(0x3b9aca00)
	/usr/local/go/src/runtime/time.go:59 +0xe1
github.com/influxdata/influxdb/tsdb/engine/tsm1.(*Engine).compactTSMLevel(0xc429826100, 0xb74601, 0x1, 0xc428dd2300)
	/root/go/src/github.com/influxdata/influxdb/tsdb/engine/tsm1/engine.go:979 +0x90
```

jwilder:
@rubycut Can you attach the full panic?  The two links provided cut off the actual cause of the panic.

Can you also run the following and attach to the issue:
```
curl -o block.txt "http://localhost:8086/debug/pprof/block?debug=1" 
curl -o goroutine.txt "http://localhost:8086/debug/pprof/goroutine?debug=1" 
curl -o heap.txt "http://localhost:8086/debug/pprof/heap?debug=1" 
curl -o vars.txt "http://localhost:8086/debug/vars" 
iostat -xd 1 30 > iostat.txt
uname -a > uname.txt
influx -execute "show shards" > shards.txt
influx -execute "show stats" > stats.txt
influx -execute "show diagnostics" > diagnostics.txt
```

rubycut:
```
uname -a
Linux dedxxxx 3.10.95xx #30 SMP Fri Jan 29 11:54:46 CST 2016 x86_64 GNU/Linux
```

rubycut:
[...]

PR #8348, opened 2017-05-02
Compaction limits

### LINKED ISSUE #7425: move some compaction logs to trace level
`aborted level 1 group (0). compaction in progress` sounds scary, but isn't. 

See https://groups.google.com/d/msgid/influxdb/88679401-4e5b-4cee-9f20-80266333994f%40googlegroups.com for context.

@jwilder

jwilder:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

This change adds a new limit to control how many concurrent full/level compactions can run at one time.  It also stops background compaction goroutines for cold shards.

* `max-concurrent-compactions` can be set to limit the number of full and level compactions that run concurrently.  This limit does not apply to snapshot compactions as they must run to avoid filling the cache.  A value of `0`, default, sets the limit to `runtime.GOMAXPROCS(0)`.  Any value greater than 0 will limit compactions to that number.  Compactions that are scheduled to run when the limit is met, will block until one completes.  This limit can be used to throttle CPU usage due to many concurrent compactions.
* The startup process has changed to disable compactions until all shards have opened.  Previously, a shard would start compactions as soon as it was opened even if other shards were still opening.  This could slow down startup due to many compactions kicking in and consuming CPU/IO while other shards are loading.
* Shards have about 5 goroutines used for level/full compactions.  When there are thousands of shards, the number of goroutines can be high.  While this hasn't been an issue per-se, it is inefficient considering most of these goroutines are not doing anything once a shard goes cold and is fully compacted.  These goroutines are now stopped once the shard goes cold and they are restarted after new writes/deletes to the shard occur.
* Compaction planning runs independently for each level.  Sometimes the planning can assign the same file to different plans.  This was handled by the `Compactor` which would keep track of which files are currently being compacted.  If a duplicate file was seen, the second compaction would be aborted leading to messages as seen in #7425.  The planner was updated to keep track of what files have been assigned to plans to prevent plans being returned with overlapping files.  #7425 should not occur now, but the safeguards in the `Compactor` are still in place.
* There was bug where `tmp` files would not be cleaned up when an error occurred during a compaction.  These are now removed.
* `monitor` goroutine per shard has been removed and handled by a single goroutine for all shards on the store.
* `Cache.Size` didn't include the snapshot size which cause the size to be misreported.

Fixes #7425 #8276 #8123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The admin UI is removed and unusable in this release. The `[admin]` configuratio
 - [#8302](https://github.com/influxdata/influxdb/pull/8302): Write throughput/concurrency improvements
 - [#8273](https://github.com/influxdata/influxdb/issues/8273): Remove the admin UI.
 - [#8327](https://github.com/influxdata/influxdb/pull/8327): Update to go1.8.1
+- [#8348](https://github.com/influxdata/influxdb/pull/8348): Add max concurrent compaction limits
 
 ### Bugfixes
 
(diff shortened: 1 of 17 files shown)


jwilder:
@benbjohnson Found a bug while testing and pushed up a fix.  I also reworked the `monitor` goroutine so there is no longer 1 per shard.

benbjohnson:
@jwilder 👍

mahaveer1707:
https://github.com/influxdata/influxdb/issues/7640#issuecomment-532125601

PR #8350, opened 2017-05-02
Track HTTP client requests for /write and /query with /debug/requests

jsternberg:
After using `/debug/requests`, the client will wait for 30 seconds
(configurable by specifying `seconds=` in the query parameters) and the
HTTP handler will track every incoming query and write to the system.
After that time period has passed, it will output a JSON blob that looks
very similar to `/debug/vars` that shows every IP address and user
account (if authentication is used) that connected to the host during
that time.

In the future, we can add more metrics to track. This is an initial
start to aid with debugging machines that connect too often by looking
at a sample of time (like `/debug/pprof`).

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): issue filed or pull request submitted \<[influxdata/docs.influxdata.com#1122](https://github.com/influxdata/docs.influxdata.com/issues/1122)\>

### DIFF
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -38,6 +38,10 @@ const (
 	//
 	// This has no relation to the number of bytes that are returned.
 	DefaultChunkSize = 10000
+
+	DefaultDebugRequestsInterval = 10 * time.Second
+
+	MaxDebugRequestsInterval = 6 * time.Hour
 )
 
 // AuthenticationMethod defines the type of authentication used.
@@ -99,16 +103,19 @@ type Handler struct {
 	Logger    zap.Logger
 	CLFLogger *log.Logger
 	stats     *Statistics
+
+	requestTracker *RequestTracker
 }
 
 // NewHandler returns a new instance of handler with routes.
 func NewHandler(c Config) *Handler {
 	h := &Handler{
-		mux:       pat.New(),
-		Config:    &c,
-		Logger:    zap.New(zap.NullEncoder()),
-		CLFLogger: log.New(os.Stderr, "[httpd] ", 0),
-		stats:     &Statistics{},
+		mux:            pat.New(),
+		Config:         &c,
+		Logger:         zap.New(zap.NullEncoder()),
+		CLFLogger:      log.New(os.Stderr, "[httpd] ", 0),
+		stats:          &Statistics{},
+		requestTracker: NewRequestTracker(),
 	}
 
 	h.AddRoutes([]Route{
@@ -257,6 +264,8 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	} else if strings.HasPrefix(r.URL.Path, "/debug/vars") {
 		h.serveExpvar(w, r)
+	} else if strings.HasPrefix(r.URL.Path, "/debug/requests") {
+		h.serveDebugRequests(w, r)
 	} else {
 		h.mux.ServeHTTP(w, r)
 	}
@@ -282,6 +291,7 @@ func (h *Handler) serveQuery(w http.ResponseWriter, r *http.Request, user *meta.
 	defer func(start time.Time) {
 		atomic.AddInt64(&h.stats.QueryRequestDuration, time.Since(start).Nanoseconds())
 	}(time.Now())
+	h.requestTracker.Add(r, user)
 
 	// Retrieve the underlying ResponseWriter or initialize our own.
 	rw, ok := w.(ResponseWriter)
@@ -584,6 +594,7 @@ func (h *Handler) serveWrite(w http.ResponseWriter, r *http.Request, user *meta.
 		atomic.AddInt64(&h.stats.ActiveWriteRequests, -1)
 		atomic.AddInt64(&h.stats.WriteRequestDuration, time.Since(start).Nanoseconds())
 	}(time.Now())
+	h.requestTracker.Add(r, user)
 
 	database := r.URL.Query().Get("db")
 	if database == "" {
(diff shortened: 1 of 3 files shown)


jsternberg:
This is an initial try at doing this. Likely, this needs a configuration option to turn it off (or turn it on) and likely needs to keep more statistics than it currently keeps.

services/httpd/requests.go
```diff
@@ -0,0 +1,113 @@
...
+		return
+	}
+
+	info.IPAddr = host
+	if user != nil {
+		info.Username = user.Name
+	}
+
+	// There is an entry in the request tracker. Increment that.
+	if n, ok := rt.requests[info]; ok {
```
jwilder: This needs an `RLock` to prevent a concurrent map panic.

services/httpd/requests.go
```diff
@@ -0,0 +1,113 @@
...
+			return
+		}
+
+		rt.mu.Lock()
+		if len(rt.requests) == 0 {
+			continue
+		}
+
+		requests := rt.requests
+		rt.requests = make(map[RequestInfo]*int64, len(requests))
```
jwilder: Wiping the map each minute might be confusing as a user.  In one second, you would see a bunch of writers and in the next they would all be gone, then they might start to show up again.  Recording it over time might show very spiky usage too.

services/httpd/requests.go
jsternberg: Ah, I was under the impression that maps could be read from concurrently. That may mean that we can't return the in-use map to be written to `/debug/vars`.

services/httpd/requests.go
jsternberg: It might be better to just take the aggregated map and store it and return that aggregated map from `/debug/vars`.

Maybe it would be better to just increment the request info always? But then the person would need to read the values over time and use derivative like they do with the other internal monitoring. It seems to me like any attempt to do this without storing some historical data in the database may not work. We would also have to eventually cull any clients that just disappear.

jwilder: You might be able to use a ring buffer of these maps where each slot is some fixed sized time window (e.g. 10s).  As you roll over to new slots, you would wipe them clean of old entries.

services/httpd/requests.go
jwilder: Concurrent readers are fine provided there are no writers.  There is a writer here: https://github.com/influxdata/influxdb/pull/8350/files#diff-badb764b9e9789c5a3bf719f6c815aadR71

There is also a data race because the map is replaced here: https://github.com/influxdata/influxdb/pull/8350/files#diff-badb764b9e9789c5a3bf719f6c815aadR110

services/httpd/requests.go
```diff
@@ -0,0 +1,128 @@
...
+		Requests: make(map[RequestInfo]*RequestStats),
+		tracker:  rt,
+	}
+
+	rt.mu.Lock()
+	defer rt.mu.Unlock()
+
+	for {
+		n := rt.rand.Int63()
+		if _, ok := rt.profiles[n]; ok {
```
jwilder: Why not just use a slice and append the new profile to it?  Id would be `len(rt.profiles)`.

jwilder:
Other dimensions that might be useful to record as well are `db` and `rp`.

services/httpd/handler.go
```diff
@@ -204,6 +206,15 @@ func (h *Handler) Statistics(tags map[string]string) []models.Statistic {
 	}}
 }
 
+// TrackRequests begins tracking requests made to the Handler.
+// This should be called before this Handler is used to serve requests.
+func (h *Handler) TrackRequests() {
+	if h.requestTracker != nil {
+		return
+	}
+	h.requestTracker = NewRequestTracker()
```
jsternberg: This can be removed. This was only used previously because the request tracker had a timer. Now it doesn't have a timer anymore so I can just move it to the handler creation.

services/httpd/handler.go
```diff
@@ -834,6 +855,52 @@ func (h *Handler) serveExpvar(w http.ResponseWriter, r *http.Request) {
...
+	case <-timer.C:
+		profile.Stop()
+	case <-closing:
+		// Connection was closed early.
+		profile.Stop()
+		timer.Stop()
+		return
+	}
+
+	w.Header().Add("Connection", "close")
```
jsternberg: Anything that uses `CloseNotify()` needs to have this since the HTTP connection can't be reused. I can remove the `CloseNotify()` part, but then a lost connection will result in requests continuing to be tracked after the connection is lost.

jwilder: Ok.  Makes sense.

services/httpd/requests.go
jsternberg: Wouldn't that make it difficult to remove? A linked list might actually be better for this kind of thing.

jsternberg:
When it comes to `/write` and `/query`, do we want the format to be:

```
{
    "127.0.0.1": {
        "writes": 12,
        "queries": 5,
    },
}
```

Or

```
{
    "/query": {
        "127.0.0.1": 5
    },
    "/write": {
        "127.0.0.1": 12,
    },
}
```

If you're only looking for one of the two, the second might be more useful since it separates the information to only other data that's related to it. But the first might make more sense organizationally.

services/httpd/requests.go
jwilder: Is there 1 `profiles` map for each request to `/track-requests`?  I can't imagine that there would be a lot of concurrent tracking requests in practice.

jsternberg: The idea is that each request to profile gets its own map so it can have its own tracking. Then, internally, there's a map for the requests themselves.

So there's one profiles map (or list) and there's on entry in that map (or list) per request. A slice would technically work, but then removing the entry when the time is finished is `O(n)` for removal and would shift the index of all of the other profiles making it harder for them to remove.

jwilder:
I was thinking the first:
```
{
    "127.0.0.1": {
        "writes": 12,
        "queries": 5,
    },
}
```

Mainly because I was looking for a specific IP first and then figuring out what it was doing.  The latter would requirement to look at writes and then find my ip. The queries and find my IP again.

services/httpd/requests.go
jsternberg: I've pushed the change that I made to use `list.List` instead which should be better for this situation.

services/httpd/handler.go
```diff
@@ -834,6 +851,58 @@ func (h *Handler) serveExpvar(w http.ResponseWriter, r *http.Request) {
...
+	requests := make(map[string]*RequestStats, len(profile.Requests))
+	for req, st := range profile.Requests {
```
jsternberg: Since the profile has been stopped, it shouldn't be modified any longer. The lock should only be needed while the profile is happening.

jsternberg:
I'm also thinking this should output the JSON in a similar manner to `/debug/vars`, but I'll do that after we work out the other details.

jsternberg:
I squashed the PR, modified the format to look more like `/debug/vars` for easier readability, and updated the changelog.

PR #8357, opened 2017-05-03
Fix panic in AST rewriter when (*SelectStatement).Condition == nil

rw-influxdata:
The AST rewriter tries to assert `nil` to be an `influxql.Expr` and thus causes a panic. This patch checks that the field is not `nil` before running the rewriter.

### DIFF
--- a/influxql/ast.go
+++ b/influxql/ast.go
@@ -4300,7 +4300,15 @@ func Rewrite(r Rewriter, node Node) Node {
 		n.Fields = Rewrite(r, n.Fields).(Fields)
 		n.Dimensions = Rewrite(r, n.Dimensions).(Dimensions)
 		n.Sources = Rewrite(r, n.Sources).(Sources)
-		n.Condition = Rewrite(r, n.Condition).(Expr)
+
+		// Rewrite may return nil. Nil does not satisfy the Expr
+		// interface. We only assert the rewritten result to be an
+		// Expr if it is not nil:
+		if cond := Rewrite(r, n.Condition); cond != nil {
+			n.Condition = cond.(Expr)
+		} else {
+			n.Condition = nil
+		}
 
 	case *SubQuery:
 		n.Statement = Rewrite(r, n.Statement).(*SelectStatement)


rw-influxdata:
The appveyor failure does not seem to have to do with my PR:

```
==2084==ERROR: ThreadSanitizer failed to allocate 0x80000 (524288) bytes of DenseSlabAllocator (error code: 1455)
FATAL: ThreadSanitizer CHECK failed: gotsan.cc:6498 "((0 && "unable to mmap")) != (0)" (0x0, 0x0)
FAIL	github.com/influxdata/influxdb/tsdb/engine/tsm1	122.440s
```

influxql/ast.go
```diff
@@ -4300,7 +4300,13 @@ func Rewrite(r Rewriter, node Node) Node {
 		n.Fields = Rewrite(r, n.Fields).(Fields)
 		n.Dimensions = Rewrite(r, n.Dimensions).(Dimensions)
 		n.Sources = Rewrite(r, n.Sources).(Sources)
-		n.Condition = Rewrite(r, n.Condition).(Expr)
+
+		// n.Condition may be nil. In that case, Rewrite will return
+		// nil. This causes a panic in the type assertion to Expr,
+		// because nil does not satisfy the Expr interface:
+		if n.Condition != nil {
+			n.Condition = Rewrite(r, n.Condition).(Expr)
```
jsternberg: This looks like it's trying to assert the return value is an expression. It seems to me this should instead by something like this:

```go
if cond := Rewrite(r, n.Condition); cond != nil {
    n.Condition = cond.(Expr)
} else {
    n.Condition = nil
}
```

This way, it's possible for the rewrite to return a nil condition. It appears that this appears to cause an error when nil is passed in because it doesn't touch the passed in nil and then tries to assert the nil is an `Expr`.

rw-influxdata: I agree. I updated the code.

rw-influxdata:
Squashed.

PR #8370, opened 2017-05-08
Fixes races/memory usage

jwilder:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

This is a follow up to #8348 that fixes data race and reduces memory allocations when monitoring shards.


hpbieker:
@jwilder Thank you. I see that it is already merged, we will test it when we do our next upgrade.

PR #8381, opened 2017-05-11
fix NewShardOwner to not return a current owner

dgnorton:
Fixed meta.NewShardOwner() so it wouldn't return an ID for a data node
that already owns the shard.

- [x] Rebased/mergable
- [ ] Tests pass

### DIFF
--- a/services/meta/data.go
+++ b/services/meta/data.go
@@ -50,10 +50,10 @@ type Data struct {
 	MaxShardID      uint64
 }
 
-// NewShardOwner sets the owner of the provided shard to the data node
-// that currently owns the fewest number of shards. If multiple nodes
-// own the same (fewest) number of shards, then one of those nodes
-// becomes the new shard owner.
+// NewShardOwner returns the ID of a data node that the shard should be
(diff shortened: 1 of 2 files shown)


services/meta/data.go
```diff
@@ -50,10 +50,10 @@ type Data struct {
...
 
-// NewShardOwner sets the owner of the provided shard to the data node
-// that currently owns the fewest number of shards. If multiple nodes
-// own the same (fewest) number of shards, then one of those nodes
-// becomes the new shard owner.
+// NewShardOwner returns the ID of a data node that the shard should be
+// reassigned to based on which node currently owns the fewest shards.
+// If multiple nodes own the same (fewest) number of shards, one of those
+// nodes is chosen as the new shard owner.
 func NewShardOwner(s ShardInfo, ownerFreqs map[int]int) (uint64, error) {
```
jwilder: Is this called via a Raft command?

dgnorton:
Closing because this function is no longer needed and will be removed by a followup PR.

PR #8384, opened 2017-05-11
Write and compaction stability

jwilder:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

These are some changes to improve stability for writes and compactions.  #8302 improved write throughput by ~2x.  As a result, snapshot compactions are written much more frequently which cause TSM files to accumulate much faster.  The current compaction planning attempts to kicks off as many compactions as possible and allow the go runtime to manage them.

When there are many shards, this can cause problems with many full or expensive compactions running at once spiking CPU and using a lot of disk space.  #8348 added a config option to limit the global number of running level and full compactions across shards, but an individual level could still kick off more compactions than the limit.  The PR limits concurrent compactions within a level and reduces some allocations to improve stability and jitter.

I ran a 2B write load w/ 50 concurrent writers against 1.2.3, master and this PR.  Total series is 100k w/ 1 value.  There was no query load.  This is on a `c4.8xlarge`.  I also ran a larger 20B run against master and this PR to see when many level/full compactions within a shard would be kicked off an what the impact was.

The main changes in this PR:
* Convert `Point.Name` to return `[]byte` instead of string.  We were converting from `[]byte` to `string` to `[]byte` in many places causes lots of garbage.
* Switch the TSM WAL buffer pool back to a managed pool from `sync.Pool`.  The `sync.Pool` wasn't able to hold onto the allocations so we were allocating many `[]byte` over and over.  The pool used is not limited to drop buffers that are too big.
* Limit concurrent compactions within a shard.  Instead of kicking off as many goroutines as there are compactions groups, cap it at 4 (determined empirically to work well).
* Rework key parsing to not allocate and intermediate tag map that we ignore. 

### Writes

This is HTTP write throughput and latency.  1.2.3 sustains about 1.1M w/s and latency varies around 210ms-320ms.  `master` and this PR sustain around 2.3M w/s.  `master` latency fluctuates around 100-110ms and with this change it reduces to 90-100ms.  In the `master` run, you can see the throughput jump around quite a bit and is slowly going down.  On larger test runs (20B), this variance can be more dramatic.  This variance appears to be due to CPU starvation caused by lots of allocations and compactions competing for CPU.

#### 2B Point Run
writes

#### 20B Point Run

This larger test run shows what happens to writes when too many level or full compactions kick in at once within a shard.

`master`

screen shot 2017-05-11 at 3 02 45 pm

In the middle of the graph, the write throughput fluctuations are correlated with about 21 concurrent compactions running (14 full and 7 level 2).

screen shot 2017-05-11 at 3 07 48 pm

This is the same write load against this PR.

screen shot 2017-05-11 at 3 05 34 pm

Compactions are limited in this run and larger variations are gone.

screen shot 2017-05-11 at 3 15 17 pm

### Compactions

Watching the TSM file count shows that `master` compactions are not keeping up with the increased write load and are slowly backing up.  This is a 2B point run.

filecount

The current compaction planning is also spinning up a lot of compactions that are competing with each other and slowing progress.

running

With this change, file count stays more steady with the increased write load and the number of concurrent compactions stays more consistent.  

While I was not measuring it, I noticed CPU util and load was significantly reduced from `master` to this change as well.


models/points.go
```diff
@@ -248,6 +251,20 @@ func ParseKey(buf []byte) (string, Tags, error) {
...
+
+func ParseName(buf []byte) ([]byte, error) {
+	// Ignore the error because scanMeasurement returns "missing fields" which we ignore
+	// when just parsing a key
+	state, i, err := scanMeasurement(buf, 0)
+	if state == tagKeyState {
+		return buf[:i-1], nil
+	}
+	return buf[:i], err
+}
```
benbjohnson: The comment says "ignore the error" but it gets passed back. I just want to make sure that's intentional.

jwilder: Good catch, it should be returning nil here and err should err should be `_`.

tsdb/engine/tsm1/wal.go
```diff
@@ -383,20 +387,17 @@ func (l *WAL) writeToLog(entry WALEntry) (int, error) {
 	// limit how many concurrent encodings can be in flight.  Since we can only
```
jwilder: Yes.  I didn't see these comments earlier before I merged.  I'll submit a new PR to remove these lines.

PR #8389, opened 2017-05-15
Import Override Support

pmyjavec:
Allow users to override the DML parameters found in dump/backup files.

This is especially useful when importing large databases as it doesn't
require the file to be extracted to change these values.

* Allow users to specify the destination database name.
* Allow users to override the retention policy value.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

- [ ] Provide example syntax


cmd/influx/main.go
```diff
@@ -56,6 +56,8 @@ func main() {
 	fs.IntVar(&c.ImporterConfig.PPS, "pps", defaultPPS, "How many points per second the import will allow.  By default it is zero and will not throttle importing.")
 	fs.StringVar(&c.ImporterConfig.Path, "path", "", "path to the file to import")
 	fs.BoolVar(&c.ImporterConfig.Compressed, "compressed", false, "set to true if the import file is compressed")
+	fs.StringVar(&c.ImporterConfig.RetentionPolicy, "retention-policy", "", "Overrides the retention policy when importing a databse from file")
```
e-dard: `importing a databse`

cmd/influx/main.go
```diff
@@ -56,6 +56,8 @@ func main() {
...
+	fs.StringVar(&c.ImporterConfig.DestinationDatabase, "destination-database", "", "Overrides the database name when importing a databse from file")
```
e-dard: `importing a databse`

cmd/influx/main.go
```diff
@@ -96,6 +98,10 @@ func main() {
        Path to file to import
   -compressed
        Set to true if the import file is compressed
+  -retention-policy
+		Overrides the retention policy when importing a databse from file. Must be used with -import.
```
e-dard: `importing a databse`

cmd/influx/main.go
```diff
@@ -96,6 +98,10 @@ func main() {
...
+  -destination-database
```
e-dard: `-destination-databse`

importer/v8/importer.go
```diff
@@ -153,18 +157,37 @@ func (i *Importer) processDDL(scanner *bufio.Scanner) {
...
 		}
-		i.queryExecutor(line)
+
+		i.createDatabaseQuery = line
 	}
 }
 
-func (i *Importer) processDML(scanner *bufio.Scanner) {
+// processDML actually processes each of the points once it has created the target database. The database
+// databae can either be specified by the user, or it will be read from the import file.
```
e-dard: `databae` can go.

importer/v8/importer.go
```diff
@@ -153,18 +157,37 @@ func (i *Importer) processDDL(scanner *bufio.Scanner) {
...
+//The saame goes for the retention policy it too can be overriden by specifying `rp` or else it too is set
```
e-dard: `saame`

man/influx.txt
```diff
@@ -78,4 +78,10 @@ OPTIONS
 -version::
   Outputs the version of the influx client.
 
+-destination-database::
+  Overrides the database name found in the import file and sets it to the specified value, must be used with -import.
+
+-retention-policy::
+  Overrides the retention policy when importing a databse from file. must be used with -import.
```
e-dard: `databse`

cmd/influx/main.go
e-dard: I think that this, and the `destination-database` flag should say something like "writes the import data to the provided retention policy when using -import". `Overrides` seemed a little confusion to me as to whether it was modifying the import file or changing the location the imported data was written to.

pmyjavec:
Ok, 

@e-dard all issues have been fixed.

importer/v8/importer.go
```diff
@@ -153,18 +157,36 @@ func (i *Importer) processDDL(scanner *bufio.Scanner) {
...
 			continue
 		}
-		i.queryExecutor(line)
+
+		i.createDatabaseQuery = line
 	}
 }
 
-func (i *Importer) processDML(scanner *bufio.Scanner) {
+// processDML actually processes each of the points once it has created the target database. The databae can
```
e-dard: `databae`

e-dard:
@pmyjavec this PR will also need one more review from a platform team member.

pmyjavec:
@e-dard,

Ok no problem, thanks for that. Sorry about the typos, I'll blame it on my new ergonomic keyboard :)

cmd/influx/main.go
```diff
@@ -56,6 +56,8 @@ func main() {
 	fs.IntVar(&c.ImporterConfig.PPS, "pps", defaultPPS, "How many points per second the import will allow.  By default it is zero and will not throttle importing.")
 	fs.StringVar(&c.ImporterConfig.Path, "path", "", "path to the file to import")
 	fs.BoolVar(&c.ImporterConfig.Compressed, "compressed", false, "set to true if the import file is compressed")
+	fs.StringVar(&c.ImporterConfig.RetentionPolicy, "retention-policy", "", "Writes the import data to the specified retention policy when using -import.")
```
corylanou: for consistency, should it be `destination-retention-policy` to match `destination-database`?

cmd/influx/main.go
```diff
@@ -96,6 +98,10 @@ func main() {
        Path to file to import
   -compressed
        Set to true if the import file is compressed
+  -retention-policy
```
corylanou: we we do update the parameter per my comment above, this reference would need to be updated as well

man/influx.txt
```diff
@@ -78,4 +78,10 @@ OPTIONS
 -version::
   Outputs the version of the influx client.
 
+-retention-policy::
```
corylanou: may need to update this reference if we update the name.

cmd/influx/main.go
pmyjavec: Hello @corylanou,

I did think about using that parameter name, I'm pretty big consistency too but I think it's really a little too long and considering it's not used for anything else, nor I can't think of any case where it ever will, I felt that this name is probably fine.

Thoughts?

pmyjavec: @corylanou, rather than updating the README, how about we just change the wording to something like:

> Overrides all retention policy statements found in the import file and sets them to the specified parameter when using -import

Something similar for the `-destination-database` parameter.

pmyjavec: @e-dard, @corylanou any thoughts?

stromnet:
Any ETA on this? Just updated to 1.3 in the hope to see this landed but unfortunately not.. :)

importer/v8/importer.go
```diff
@@ -153,18 +157,36 @@ func (i *Importer) processDDL(scanner *bufio.Scanner) {
 		if strings.TrimSpace(line) == "" {
 			continue
 		}
-		i.queryExecutor(line)
+
+		i.createDatabaseQuery = line
```
rbetts: This constantly sets `createDatabaseQuery` to the last line of the DDL section. Is that correct? Does `CREATE DATABASE` have to be the final line of the DDL preamble?

Also, whatever `line` is -- it was executed before being stored here; meaning at minimum, we will CREATE the database in the DDL section and then create the database for the override; that seems incorrect. If an override is specified, I wouldn't expect the database in the file to be created.

importer/v8/importer.go
```diff
@@ -153,18 +157,36 @@ func (i *Importer) processDDL(scanner *bufio.Scanner) {
...
+// either be specified by the user, or it will be read from the import file.  The same goes for the retention
+// policy it too can be overriden by specifying `rp` or else it too is set to whatever is spcfied in the DML.
+func (i *Importer) processDML(scanner *bufio.Scanner, dboverride string, rpoverride string) {
+
+	// If a user specified a dboverride, override the command specified in the DDL
+	if dboverride != "" {
+		i.createDatabaseQuery = fmt.Sprintf("CREATE DATABASE %s", dboverride)
+		i.database = dboverride
+	}
+	i.queryExecutor(i.createDatabaseQuery)
```
rbetts: If `dboverride == ""`, this command was already executed in `processDDL()`.  Also, I don't expect `CREATE DATABASE` to be called in a function named `processDML()`.

pmyjavec: I'm unclear on how that happens because the processDDL just finds the create database statement and sets it for processing later. It used to be the case that processDDL creates the database but in the PR that is no longer the case. From the previous snippet:

` i.createDatabaseQuery = line `

pmyjavec: Notice that it just sets `i.createDatabaseQuery`, it does not execute the queryExecutor.

pmyjavec:
Hey @rbetts,

I've addressed your concerns by removing any database creation statements when the overrides are specified. I think this makes more sense anyway.

Thanks

PR #8391, opened 2017-05-15
make client/v2/BatchPoints precision match models/Points

zuercher:
Currently client/v2/BatchPoints enforces precisions that are valid golang time.Duration units (ns, us, ms, s, m, or h). Furthermore, it will accept a precision string like "0s" (since "10s" is a valid time.Duration). However, models/Points.PrecisionString, which is ultimately used by the client to generate the line protocol data for writing points, requires one of ns, u, ms, s, m, or h. Otherwise it silently ignores the precision and generates a nanosecond precision timestamp. The two sets of units are the same except for microseconds (us vs u).

I swapped out the calls to `time.ParseDuration` in favor of explicitly checking for the valid values. It now accepts "u" (seemingly the definitive unit name for microseconds) as well as "us" (since some code might use the client with this value, even though it doesn't work correctly). I threw in "µ" and "µs" since these are valid elsewhere in the API. I moved tests of NewBatchPoints and BatchPoints.SetPrecision to a separate, exhaustive, test.

- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated
- [x] Sign [CLA](https://influxdata.com/community/cla/)

### DIFF
--- a/client/v2/client.go
+++ b/client/v2/client.go
@@ -211,10 +211,10 @@ type BatchPoints interface {
 
 // NewBatchPoints returns a BatchPoints interface based on the given config.
 func NewBatchPoints(conf BatchPointsConfig) (BatchPoints, error) {
+	var err error
 	if conf.Precision == "" {
 		conf.Precision = "ns"
-	}
-	if _, err := time.ParseDuration("1" + conf.Precision); err != nil {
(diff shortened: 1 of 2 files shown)


zuercher:
Since you're clearly not interested in even discussing this patch, let's just close it.

PR #8394, opened 2017-05-16
Optimize top() and bottom() using an incremental aggregator

jsternberg:
The previous version of `top()` and `bottom()` would gather all of the
points to use in a slice, filter them (if necessary), then use a
slightly modified heap sort to retrieve the top or bottom values.

This performed horrendously from the standpoint of memory. Since it
consumed so much memory and spent so much time in allocations (along
with sorting a potentially very large slice), this affected speed too.

These calls have now been modified so they keep the top or bottom points
in a min or max heap. For `top()`, a new point will read the minimum
value from the heap. If the new point is greater than the minimum point,
it will replace the minimum point and fix the heap with the new value.
If the new point is smaller, it discards that point. For `bottom()`, the
process is the opposite.

It will then sort the final result to ensure the correct ordering of the
selected points.

When `top()` or `bottom()` contain a tag to select, they have now been
modified so this query:

    SELECT top(value, host, 2) FROM cpu

Essentially becomes this query:

    SELECT top(value, 2), host FROM (
        SELECT max(value) FROM cpu GROUP BY host
    )

This should drastically increase the performance of all `top()` and
`bottom()` queries.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ The admin UI is removed and unusable in this release. The `[admin]` configuratio
 - [#8384](https://github.com/influxdata/influxdb/pull/8384): Write and compaction stability
 - [#7862](https://github.com/influxdata/influxdb/pull/7861): Add new profile endpoint for gathering all debug profiles and querues in single archive.
 - [#8390](https://github.com/influxdata/influxdb/issues/8390): Add nanosecond duration literal support.
+- [#8394](https://github.com/influxdata/influxdb/pull/8394): Optimize top() and bottom() using an incremental aggregator.
 
 ### Bugfixes
 
--- a/influxql/call_iterator.go
+++ b/influxql/call_iterator.go
@@ -1,8 +1,6 @@
 package influxql
 
 import (
-	"bytes"
-	"container/heap"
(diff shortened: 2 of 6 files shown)


jsternberg:
Benchmark comparison:

```
benchmark                    old ns/op     new ns/op     delta
BenchmarkSelect_Top_1K-8     381236352     64565949      -83.06%

benchmark                    old allocs     new allocs     delta
BenchmarkSelect_Top_1K-8     110            66             -40.00%

benchmark                    old bytes     new bytes     delta
BenchmarkSelect_Top_1K-8     529304706     12168         -100.00%
```

There should also be an improvement when using `SELECT top(value, host, <n>) ...`, but the code to benchmark this is less straightforward since the old implementation worked completely differently than this new implementation. I can try if needed, but it wouldn't be a 1-1 match between the code.

discoduck2x:
@jsternberg very interesting , is this part of the nightly build now?

jsternberg:
This has not been merged yet. It still needs to be reviewed by someone before it gets merged.

PR #8397, opened 2017-05-17
Tidy up Contributing Guidelines & Makefile issues

pmyjavec:
Tidy up the `CONTRIBUTION.md` as some of the information in there was a little
out of date.

Also fix minor issues with Makefile and add ability to build the project
using `make`, this makes things simpler for new comers to contribute.

* Tidy up contribution documentation
* Add some new targets to the `Makefile` for testing and building.
* Rule separation as found in https://www.gnu.org/software/make/manual/make.html#Rule-Syntax
* Added default `make` target to make the dev workflow easier.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,178 +1,127 @@
-Contributing to InfluxDB
-========================
+# Contributing to InfluxDB
 
-Bug reports
----------------
-Before you file an issue, please search existing issues in case it has already been filed, or perhaps even fixed. If you file an issue, please include the following.
+Welcome prospective Influx contributor!
+
+The following guide will help you get started contributing to InfluxDB, to make this process as smooth as possible, we've prepared this short guide for you. It's split it into three sections for your convenience:
+
+* [Getting started](#getting-started) contains instructions for setting up your development environment and links to any pre-required reading material.
+* [Guidelines](#guidelines) outlines steps for submitting bugs, feature requests and pull-request.
+* [Advanced Topics](#advanced-topics) for experienced contributors and developers, information about building packages etc.
+* [Assistance](#assistance) have further questions or need help? Checkout this section.
+
+## Getting Started
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -1,178 +1,127 @@
...
 
 ```bash
-    mkdir $HOME/gocodez
-    export GOPATH=$HOME/gocodez
-    go get github.com/influxdata/influxdb
+go get github.com/influxdata/influxdb
 ```
 
-You can add the line `export GOPATH=$HOME/gocodez` to your bash/zsh file to be set for every shell instead of having to manually run it everytime.
+#### Working with a Fork
```
e-dard: I don't think this is the best way contributors can work with a fork of the project. For one thing it can get tricky or annoying with imports, if their editors/IDEs try to use their fork's location in the import paths. A better way to work with forks in Go is provided by [Francesc Campoy](http://blog.campoy.cat/2014/03/github-and-go-forking-pull-requests-and.html). We should probably suggest people do that.

pmyjavec:
Hey @rbetts and @e-dard, there isn't a move away from using official go tooling. Make is used in this instance as a convenience wrapper to run bunch of official go tooling and runs them  in the correct stages to produce a build, automates a lot of the steps required to get started and produce binaries.

It's entirely optional whether people use it or not; However it seems to be popular in open source Go projects I've contributed too so I didn't think it would hurt fixing ours up a little. 

Thoughts?

rbetts:
@pmyjavec Can you change the documented instructions in `CONTRIBUTING.md` to use the `go` commands directly? I like fixing the Makefile (it actually confused me the very first time, too); but I prefer that we document the standard go tooling in our suggestions.

I think that gets us a fixed `makefile` as well as an improved `contributing.md` file.

pmyjavec:
@rbetts no problems :)

jsternberg:
I'm going to close this for now since it seems there isn't any momentum on this PR. If you want to rebase this PR and update it, I'd be willing to take another look at it.

PR #8398, opened 2017-05-17
Maintain the tags of points selected by top() or bottom() when writing the results

### LINKED ISSUE #7129: CQ or Backfill using Top() with tag does not create tag in result
**System info:** 1.0.0-beta3, official docker image

**Steps to reproduce:**
1. Create a measurement with some fields and tags
2. Create a continuous query or a backfill query with INTO clause while doing a select that looks like this:
select top(summary_field, tag1, 100) as total, tag2, tag3 INTO tops FROM previous_measurement where time < now() GROUP BY tag4, time(1d) fill(none);

**Expected behavior:**
Output should mimic this:
total, tag2, tag3 should be become fields, due to the way they're selected.
tag1 and tag4 should be tags.

**Actual behavior:**
Tag1 is created as a field, not a tag.

This behavior is irrelevant if you're just selecting, however when used as a select INTO, because it is not created as a tag, you don't get the data you should.

Observe:

```
> select... INTO...:
name: result
------------
time    written
0       5181

> select count(total) from tops;
name: tops
---------------------------
time    count
0       100

> show tag keys from tops
name: tops
---------------------------
tagKey
tag4
```

Because it did not create tag1, the series was not unique enough to accommodate all the data, and the points were considered duplicates, and overwrote one another. This is why there were 5181 points written, but only 100 are available in the destination measurement of the INTO. 

I don't believe this issue is specific to my config. It appears as though this section of code hasn't been touched much since Feb 2016, and this situation is not covered by any tests in the code or documentation, or google group.

Let me know if I can provide additional information.

jsternberg:
When a `SELECT ... INTO ...` is used with `top()` or `bottom()` used with
tags, the points will be written with the tags still intact instead of
converted to fields.

Fixes #7129.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): issue filed or pull request submitted \<[influxdata/docs.influxdata.com#1135](https://github.com/influxdata/docs.influxdata.com/issues/1135)\>

### DIFF
--- a/influxql/call_iterator.go
+++ b/influxql/call_iterator.go
@@ -781,39 +781,47 @@ func IntegerSpreadReduceSlice(a []IntegerPoint) []IntegerPoint {
 	return []IntegerPoint{{Time: ZeroTime, Value: max - min}}
 }
 
-func newTopIterator(input Iterator, opt IteratorOptions, n int) (Iterator, error) {
+func newTopIterator(input Iterator, opt IteratorOptions, n int, keepTags bool) (Iterator, error) {
 	switch input := input.(type) {
 	case FloatIterator:
 		createFn := func() (FloatPointAggregator, FloatPointEmitter) {
 			fn := NewFloatTopReducer(n)
 			return fn, fn
 		}
-		return newFloatReduceFloatIterator(input, opt, createFn), nil
+		itr := newFloatReduceFloatIterator(input, opt, createFn)
+		itr.keepTags = keepTags
+		return itr, nil
 	case IntegerIterator:
 		createFn := func() (IntegerPointAggregator, IntegerPointEmitter) {
 			fn := NewIntegerTopReducer(n)
 			return fn, fn
 		}
-		return newIntegerReduceIntegerIterator(input, opt, createFn), nil
+		itr := newIntegerReduceIntegerIterator(input, opt, createFn)
+		itr.keepTags = keepTags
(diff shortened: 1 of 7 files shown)


jsternberg:
I wanted to get this up to allow discussion about it and methods (along with letting those who are interested build the PR themselves), but this explicitly needs tests before it is merged. The existing tests ensure that it doesn't change current behavior for queries, but the new behavior is strange enough that the lack of tests is a blocker to merging.

influxql/call_iterator.go
```diff
@@ -781,39 +781,47 @@ func IntegerSpreadReduceSlice(a []IntegerPoint) []IntegerPoint {
 	return []IntegerPoint{{Time: ZeroTime, Value: max - min}}
 }
 
-func newTopIterator(input Iterator, opt IteratorOptions, n int) (Iterator, error) {
+func newTopIterator(input Iterator, opt IteratorOptions, n int, writeMode bool) (Iterator, error) {
```
jwilder: Should `writeMode` be named `keepTags`?  Seems like that is what it indicates as opposed to having multiple values.

jsternberg: That sounds fine to me. The name of the variable was rather arbitrary to begin with. I couldn't think of something good.

PR #8407, opened 2017-05-19
Return partial write error when points outside of retention policy ar…

### LINKED ISSUE #8392: Points beyond retention policy scope are dropped silently
Points with timestamps outside of the retention policy's duration are accepted but not persisted.

An error response code should be returned in this case, to inform the client that the points are in fact discarded.

Repro steps:

* Create a database with a finite retention policy:

```
> create database example 
> create retention policy week on example duration 168h replication 1
> show retention policies on example
name    duration shardGroupDuration replicaN default
----    -------- ------------------ -------- -------
autogen 0s       168h0m0s           1        true
week    168h0m0s 24h0m0s            1        false
```

* Write a point beyond the retention policy's duration:

```
# curl -v -X POST "http://localhost:8086/write?db=example&rp=week" --data-binary "data value=1 1484565335000000000"
< HTTP/1.1 204 No Content
< Content-Type: application/json
< Request-Id: 7496a993-3a29-11e7-8042-000000000000
< X-Influxdb-Version: 1.2.4
```

* Point is silently dropped despite receiving a 204 success HTTP response:
```
> show shards
name: example
id database retention_policy shard_group start_time end_time expiry_time owners
-- -------- ---------------- ----------- ---------- -------- ----------- ------
> 

```

jwilder:
…e dropped

Writing points outside of a retention policy range were silently dropped. They
are dropped to prevent creating a shard that will be immediately deleted.  These
dropped points were silent and did not return an error respone to the caller.

Fixes #8392

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/coordinator/points_writer_test.go
+++ b/coordinator/points_writer_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/influxdata/influxdb/coordinator"
 	"github.com/influxdata/influxdb/models"
 	"github.com/influxdata/influxdb/services/meta"
+	"github.com/influxdata/influxdb/tsdb"
 )
 
 // TODO(benbjohnson): Rewrite tests to use cluster_test.MetaClient.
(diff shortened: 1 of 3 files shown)


coordinator/points_writer_test.go
```diff
@@ -234,8 +235,12 @@ func TestPointsWriter_MapShards_Invalid(t *testing.T) {
...
 	}
 
-	if exp := 0; len(shardMappings.Points) != exp {
-		t.Errorf("MapShards() len mismatch. got %v, exp %v", len(shardMappings.Points), exp)
+	if got, exp := len(shardMappings.Points), 0; got != exp {
+		t.Errorf("MapShards() len mismatch. got %v, exp %v", got, exp)
+	}
+
+	if got, exp := len(shardMappings.Dropped), 1; got != exp {
+		t.Fatalf("MapShard() dropped mismathc: got %v, exp %v", got, exp)
```
jsternberg: Looks like a spelling error here.

jwilder: Fixed

smifun:
This is breaking change. In out project we were relying on this behaviour and since we are not controlling influx client we cannot handle it properly. 
I would expect that there should be config property available to control whether error is return or not.

PR #8447, opened 2017-06-01
Fix SHOW TAG VALUES deduplication.

### LINKED ISSUE #8443: TSI branch has duplicate tag values
### Bug report

__System info:__ 
* 1.3.0_tsi_6608f97 (from master)
* RHEL7.3 -- 3.10.0-514.6.1.el7.x86_64
* Dell PowerEdge R720xd
* Other details as needed

__Steps to reproduce:__

1. show tag values from cpu with key in (host)
2. compare this between v1.3.0 and v1.2.3.

__Expected behavior:__ 
* Would see unique values

__Actual behavior:__ 
* Seeing duplicate values

```
[sborza@deskies]:~/src/stuffs/python/metrics (master *)$ python query.py
--- tsi host 1 ---
159674 <type 'list'> [u'host', u'hostname']
4330 <type 'set'>
--- tsm host 1 ---
7998 <type 'list'> [u'host', u'hostname']
7998 <type 'set'>
--- tsi host 2 ---
9603 <type 'list'> [u'host', u'hostname']
873 <type 'set'>
[sborza@deskies]:~/src/stuffs/python/metrics (master *)$ cat query.py
#!/usr/bin/env python

import urllib
import requests

def runit():
    """ simple query handler to prove duplicate host keys """
    query = "show tag values from cpu with key in (host)"

    for a in xrange(1, 4):
        host = "influx0{}".format(a)
        print "--- {} ---".format(host)

        r = requests.get("http://{}:8086/query?{}".format(host, urllib.urlencode({"db": "tg_udp", "q": query})))

        data = r.json()
        results = data['results'][0]['series'][0]['values']
        print len(results), type(results), results[10]

        unis = set([v for _, v in results])
        print len(unis), type(unis)

if __name__ == "__main__":
    runit()
```

benbjohnson:
## Overview

The merging of values & `SHOW TAG VALUES` broke during the transition to TSI.

Fixes #8443.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@ The admin UI is removed and unusable in this release. The `[admin]` configuratio
 - [#8387](https://github.com/influxdata/influxdb/issues/8387): Fix TSM tmp file leaked on disk
 - [#8417](https://github.com/influxdata/influxdb/issues/8417): Fix large field keys preventing snapshot compactions
 - [#7957](https://github.com/influxdata/influxdb/issues/7957): URL query parameter credentials take priority over Authentication header.
+- [#8443](https://github.com/influxdata/influxdb/issues/8443): TSI branch has duplicate tag values.
 
 
 ## v1.2.4 [2017-05-08]
--- a/tsdb/store.go
+++ b/tsdb/store.go
@@ -948,6 +948,12 @@ type TagValues struct {
 	Values      []KeyValue
 }
 
+type TagValuesSlice []TagValues
+
+func (a TagValuesSlice) Len() int           { return len(a) }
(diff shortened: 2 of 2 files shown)


benbjohnson:
@jwilder Yes, the names & tags are sorted. I started working on a faster implementation but it gets quite a bit more complicated than the current code since there's multiple levels (name/tag/value) and I don't think that's a bottleneck in `SHOW TAG VALUES`. We can re-evaluate in a separate PR if users are seeing performance problems but I'm going to merge this as-is to fix the bug.

Issue #8461, opened 2017-06-07
`influxd backup` writes to stderr by default


xginn8:
### Bug report

__System info:__ InfluxDB v1.1.0

__Steps to reproduce:__

1. Run `influxd backup` for any database

__Expected behavior:__ Normal process output should be printed as stdout, not stderr

PR #8462, opened 2017-06-07
backup output should go to stdout not stderr

xginn8:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
 
fixes https://github.com/influxdata/influxdb/issues/8461


e-dard:
@xginn8 thanks for this. However I think this isn't the right sort of change. Looking through the backup code there are cases where both logging to `stdout` and `stderr` are appropriate. I think it might be preferable to either have two loggers (one for `stderr` and one for `stdout`) _or_ to use the logging approach used in the InfluxDB server, which is to use a `zap.Logger`.

e-dard:
@xginn8 Thanks! I'll add a `CHANGELOG` entry and get this merged.

Issue #8188, opened 2017-03-23
Write CQ stats to _internal


dgnorton:
Make the CQ service write stats to _internal. This will provide users visibility into CQ activity and aid in troubleshooting.

dgnorton:
- Number of CQs executed
- Number of CQ errors
- CQ durations
- Maybe node(s) that CQ was distributed to?

stuartcarnie:
@dgnorton, what are your thoughts regarding extending the `monitor` service to provide a new API for consumers to _push_ points in addition to the current pull model. The justification for this approach is the monitor pulls aggregated statistics at a set interval (10s by default) however a service like CQs may execute queries very infrequently, resulting in many identical points until the next CQ is run. We are also forced to aggregate individual metrics and accumulate a large number of statistics in the CQ service for each CQ. By writing individual points each time a CQ is run, we can write the value immediately in addition to having accurate, point-in-time data to correlate slow queries with other system events, such as compactions, high CPU load, etc and to be able to analyze trends for individual CQs.

> Maybe node(s) that CQ was distributed to?

That information is internalized by the query and statement executors when sources are mapped to shards. We could add a new API to the executor to list the shards that will be queried for a given

dgnorton:
@stuartcarnie This is probably a dumb question but instead of a separate API, could it use the `/write` endpoint?

Re adding an API to the executor, there have been discussions about adding an `EXPLAIN` feature or ways of retrieving info from the query engine about how a query will be executed.

stuartcarnie:
@dgnorton it may be possible to use `/write` but we would incur additional overhead going through auth, point parsing, etc. We would also add a number of dependencies inside the CQ service as to which DB is used for statistics, which port to write to, credentials to use (if auth is enabled). A direct API on `Monitor` would be relatively easy to implement and abstracts the details of how / where to write this data. Conceivably, we could extend the monitor to send to a separate influxd instance, etc. 

Sounds like we could expand the stats when the `EXPLAIN` feature is added.

stuartcarnie:
Fixed by #8471

jgysel:
Although #8188 is mentioned to be part of v1.3.0 (see https://github.com/influxdata/influxdb/blob/master/CHANGELOG.md) , it looks like this is going to be released on v1.4.0. It wasn't merged to 1.3.0. The changelog on master seems to be outdated.

PR #8471, opened 2017-06-10
CQ statistics written to monitor database, addresses #8188

stuartcarnie:
* off by default, enabled by `query-stats-enabled`
* writes metrics to `cq_query` measurement of configured monitor database (`_internal` by default)
* see CHANGELOG for schema of individual points

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
...
+
+#### Continuous Query Statistics
+
+When enabled, each time a continuous query is completed, a number of details regarding the execution are written to the `cq_query` measurement of the internal monitor database (`_internal` by default). The tags and fields of interest are
+
+| tag / field | description                                           |
+|:----------- |:----------------------------------------------------- |
+| `db`        | name of database                                      |
+| `cq`        | name of continuous query                              |
+| `duration`  | query execution time in nanoseconds                   |
```
jwilder: Other stats use `durationNs` in the field name.  I'd suggest that for consistency.

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
...
+| `startTime` | lower bound of time range                             |
+| `endTime`   | upper bound of time range                             |
+| `written`   | number of points written to the target measurement    |
```
jwilder: I'd suggest `pointsWrittenOK` for consistency with other stats.

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 ## v1.3.0 [unreleased]
 
+### Release Notes
+
+#### Continuous Query Statistics
+
+When enabled, each time a continuous query is completed, a number of details regarding the execution are written to the `cq_query` measurement of the internal monitor database (`_internal` by default). The tags and fields of interest are
```
jwilder: Why disabled by default?

monitor/service.go
```diff
@@ -117,6 +117,9 @@ func (m *Monitor) Open() error {
 
 	// If enabled, record stats in a InfluxDB system.
 	if m.storeEnabled {
+		hostname, _ := os.Hostname()
+		m.SetGlobalTag("hostname", hostname)
```
jwilder: This is already handled externally when wiring up the service.

stuartcarnie: `hostname` was previously being set in the `storeStatistics` function, which is called on a separate goroutine. I moved it here to remove a race condition that it may not be set before `WritePoints` is called if the goroutine is not running.

CHANGELOG.md
stuartcarnie: Done

CHANGELOG.md
stuartcarnie: Done

CHANGELOG.md
stuartcarnie: Figured it was new behavior and wanted to give admins the option to enable, but if you think it is fine to default to enabled, I will switch it.

monitor/service.go
```diff
@@ -386,9 +415,6 @@ func (m *Monitor) storeStatistics() {
 	m.Logger.Info(fmt.Sprintf("Storing statistics in database '%s' retention policy '%s', at interval %s",
 		m.storeDatabase, m.storeRetentionPolicy, m.storeInterval))
 
-	hostname, _ := os.Hostname()
-	m.SetGlobalTag("hostname", hostname)
-
```
stuartcarnie: @jwilder this is where the `hostname` tag was moved from

CHANGELOG.md
jwilder: We typically just enable all stats for simplicity, but maybe it would be better to make them selectively enabled (or disabled).

jwilder: nit: Stats are only written to `_internal` is `[monitor] store-enabled = true`.  If that is not enabled, they are just available via `show stats` and the debug endpoints.

Issue #8470, opened 2017-06-09
index file fd leak in tsi branch


marchtea:
### Bug report

I've using latest master branch https://github.com/influxdata/influxdb/commit/4957b3d8be5fff66b1150f3fe894da09092e923a
after merging changes in https://github.com/influxdata/influxdb/pull/8468 

OS: `centos 6`
go version: `1.7.4`

It appears that index file fd is leaking.

```
cat maps | grep delete
result:
7f83f0f87000-7f83f74e3000 r--s 00000000 08:04 24182797                   /data/influxdb/data/app_6/duration_7/2740/index/L4-00000155.tsi (deleted)
7f8406810000-7f8407974000 r--s 00000000 08:04 24182837                   /data/influxdb/data/app_6/duration_7/2740/index/L2-00000207.tsi (deleted)
7f8408714000-7f840b991000 r--s 00000000 08:04 24182821                   /data/influxdb/data/app_6/duration_7/2740/index/L3-00000196.tsi (deleted)
7f8419a04000-7f841a50d000 r--s 00000000 08:04 24182840                   /data/influxdb/data/app_6/duration_7/2740/index/L1-00000200.tsi (deleted)
7f8420be7000-7f8426bf3000 r--s 00000000 08:04 24182819                   /data/influxdb/data/app_6/duration_7/2740/index/L4-00000187.tsi (deleted)
7f842759a000-7f8428f52000 r--s 00000000 08:04 24182831                   /data/influxdb/data/app_6/duration_7/2740/index/L2-00000201.tsi (deleted)
7f8428f52000-7f842ab64000 r--s 00000000 08:04 24182825                   /data/influxdb/data/app_6/duration_7/2740/index/L2-00000195.tsi (deleted)
7f842ab64000-7f842dec4000 r--s 00000000 08:04 24182817                   /data/influxdb/data/app_6/duration_7/2740/index/L3-00000186.tsi (deleted)
7f842e9c3000-7f842fd98000 r--s 00000000 08:04 24182848                   /data/influxdb/data/app_6/duration_7/2740/index/L1-00000210.tsi (deleted)
7f842fd98000-7f8433807000 r--s 00000000 08:04 30294029                   /data/influxdb/data/app_6/duration_7/2740/index/L3-00000164.tsi (deleted)
7f843faec000-7f8441ab8000 r--s 00000000 08:04 30294021                   /data/influxdb/data/app_6/duration_7/2740/index/L2-00000143.tsi (deleted)
[...]

PR #8475, opened 2017-06-12
fix index file fd leak

marchtea:
Fixes https://github.com/influxdata/influxdb/issues/8470
During index file compacting, it will be mmap twice.
First is calling `mapIndexFileSeriesBlock`, mmap the file and write index data into it, and then munmap it. After these, it will be opened as index file util being compacted into higher level file, munmap and removed from system.
The leaking is caused by the first mmap. The function `mapIndexFileSeriesBlockFile` calls `mmap`, get the data and truncates it from start which changes the address `data` points to. So, the `munmap` can't release the fd, thus, caused the fd leak.   

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


benbjohnson:
Good catch. 👍

PR #7154, opened 2016-08-15
Add msgpack handler for smarter JSON encoding

jsternberg:
The msgpack handler correctly retains information about whether the
number is a float or an integer, unlike JSON. While the format is not
human-readable, it makes a good interchange format for applications that
don't necessarily care about a human readable output.

PR #8477, opened 2017-06-12
Refactor how results are returned to the client

### LINKED ISSUE #8450: Proposal: New response structure
This issue is directly tied to #8330 and really can't be separated from it. If accepted, both of these would be implemented at the same time.

### Feature Request

I think we should reconsider the response structure that we want to be used for the client. I do not think the current one is user-friendly towards either users or API writers. To be clear, when I say response structure, I mean the way the data is structured as it comes back. I do not mean the format like JSON, MessagePack, Protocol Buffers, etc. I think the format itself is broken and based on concepts that haven't existed in the query engine since before 1.0 (or at least not how I think of them while working on the query engine). Changing the format will not fix underlying issues that get run into when programming clients.

I've seen issues that indicate people have difficulty with a large amount of data and I have experienced the problem myself while trying to work on the experimental client.

Large amounts of data means I want to be able to process the stream iteratively, but the current format makes this difficult because we return "results" that contain "series" (which we call "rows" in the code and "series" everywhere that is user facing) and inside of those series are the "values" (which should be called rows).

When we need to break down the data so it gets returned in chunks, we use the concept of a partial result or a partial series. The partial result says that the next result is actually part of the current result and a partial series means the next series is part of the current series. But, when you get into nesting and actually reading from the cursor, you end up with this absurdly complicated code to correctly find which result the next series is in. As an example, here is the code that I created for the experimental client that fully handles partials in a way so the user does not have to worry about partials: https://github.com/influxdata/influxdb-client/blob/2d000abc81dd6588d94ee5578810e9fc64bec5ab/json_cursor.go

I consider this code to be unreadable and I believe it would hamper adoption since you cannot realistically use chunking. I say this as the person who added that feature as an answer to large amounts of data.

But, the idea behind chunking was to make it a default so data would be streamed rather than forcing a potentially large amount of memory to be allocated on the server. It was supposed to _solve_ problems. While it has not solved those problems, that doesn't mean the original problems went away. We still need a way to stream points back to the user instead of returning it as a JSON blob.

With that, I think we should discuss ways that we can potentially change the format. Internal to the query engine, we already return the data in a way that would work. But then we aggregate it and return it in our monster format that's difficult to impossible to read correctly. We will obviously have to maintain the same format with JSON and CSV, but I think we should consider a structure that works better with streams for MessagePack before we merge that and then start using that for any future clients.

jsternberg:
This refactors the internal result returning system to match more
closely how we iterative return points between result sets, series, and
rows within the series. It uses the updated terminology rather than
older terminology that no longer matches how we refer to things in the
documentation or within the query engine.

The refactor moves the aggregation and chunking behavior from
`influxql.Emitter` to the HTTP service so that behavior is isolated to
one location rather than sprinkled around in multiple locations. This
refactor also allows us to use different output structures. One of these
new output structures is implemented when using `application/x-msgpack`
as the output format.

This new format is more suited to client designs than the old JSON
format for a few reasons. First, it maintains floats and integers as
separate types. So, unlike JSON, you do not have to guess if a returned
point is a float or an integer when it is a whole number. That is
returned as part of the format.

Additionally, it removes the current difficult to use chunking and
creates a new iterative format that shouldn't be as difficult to
process. For this new format, there is no non-chunking format.

- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated
- [ ] Provide example syntax

Fixes #8450.


jsternberg:
This PR is _mostly_ completed, but still a long way to final merge. There are a lot of changes in here and it isn't the cleanest PR. A decent amount of file renaming has to happen.

This PR is described by [this gist](https://gist.github.com/jsternberg/02bb27ce1087ff26e7340981795ebc35) and before we merge, this gist should be modified to act as internal documentation for the format (and then used to make developer facing documentation that isn't as coarse).

I also want to start discussion between our various client library implementors to get some feedback on if this is a decent idea or if there are other concerns that they have experienced when implementing their own libraries. The motivation behind this have mostly been my own struggles so I'm interested to hear from @dmke and @aviau.

/cc @jackzampolin for feedback on how this might help developers who use InfluxDB or to voice any concerns or improvements that could be made before we would merge this.

This PR will likely be used instead of #7154 since it is more complete and solves more problems.

Thanks everyone.

jackzampolin:
@timhallinflux

dmke:
I will look into this later today. Thanks for the heads up!

dmke:
Sorry for the delay, my laptop fan blew and I need to wait for replacement parts. Bad timing... Anyhow, here are my initial thoughts:

At a first glance, the proposed response format change sounds reasonable. I'm not sure yet what the performance implications on the client side will be. Yes, we'd potentially need (far) less memory, but:

- currently we can leverage speed from Ruby's JSON library being implemented in C (or Java for JRuby)
- the streaming format would require a state machine to keep track of the response chunks, which also implies some sort of format validation
- instantiating the parser (via `JSON.parse`) also comes with a cost; parsing the response data (header+rows, i.e. not the number counts) looks like a high number of parser instances. Worst case would be if the `Row` data comes one-by-one, e.g. `SeriesHeader, 1, Row, 1, Row, 1, ..., Row, 1, Row, 0, 0`

The last point might be just a problem with the JSON stdlib. I don't know how, for example MsgPack or other streaming-enabled libraries would behave in these situations.

I will setup a benchmark when I've replaced my fan and report back (might be next week at the earliest though).

jsternberg:
Thanks for the feedback. MessagePack should also get some of the benefits of being a C library since I believe the MessagePack implementation in Ruby is built using C. I'm not sure about JRuby though.

I created a small sample library in Ruby along the lines of what I was thinking of for a proof of concept that this would be parseable in other languages. Do you want me to push it to a repository on GitHub so you can see the rough outline of the code? It might also help with profiling.

dmke:
Indeed, [msgpack](https://rubygems.org/gems/msgpack/versions) seems to be JRuby compatible and  supports [streaming deserialization](https://github.com/msgpack/msgpack-ruby#streaming-deserialization).

> Do you want me to push it to a repository on GitHub

That would be nice to have, maybe as (orphaned) branch in influxdata/influxdb-ruby?

jsternberg:
@dmke I've just put the demo here: https://github.com/jsternberg/influxdb-ruby

jsternberg:
The implementation for the cursor is in here: https://github.com/jsternberg/influxdb-ruby/blob/master/lib/influxdb/cursor.rb

jsternberg:
I'm going to be making one slight modification to the spec format above (and I'll edit the gist). The first is that the spec example didn't match the actual text in one area. I've updated that. The second is I would like some way in the format for a reader to know if they are reading the final chunk of a section at the time they are reading the number. I believe we have this information on the server (we can't modify that because the old structure needs it).

The reason I want to make this change is I want to make it possible for a cursor to be able to give a length hint similar to here: https://github.com/influxdata/influxdb-client/blob/2d000abc81dd6588d94ee5578810e9fc64bec5ab/cursor.go#L52

That way, if you want to make an array out of the iterative results, you get some help with knowing what memory you have to allocate ahead of time.

To see the differences, please look at the revisions to the gist. If anybody has any better ideas (or think this is confusing/shouldn't be done/any other concerns), please bring them up. Thanks.

dmke:
(FYI: two different suppliers were not able to send me the correct replacment fan. I'm hoping to get the correct one from China now, but the shipment will take another 2-4 weeks or so... :sob:)

jsternberg:
I thought of another idea I want to add to this as a part of the new format. It might be nice to have the ability to return information for a query after the query has ended instead of having the entire thing be in the header. The current format doesn't really differentiate between these two because it doesn't need to.

Right now, the footer would be empty, but it might be useful for meta information that is only known after the query is finished in the future. One idea I was thinking of was building some profiling into the query engine. Right now, we have very little insight into how long individual portions of the query takes without using pprof and, even then, it can be hard to diagnose noise in the cpuprofile and get it correct. Building in timing for the different stages of the query that gets returned at the end of the response would do a lot of good for understanding these issues.

Of course, that infrastructure hasn't been built yet, but having a way to return that data is important to get into the format now so it can be added without breaking backwards compatibility later.

services/httpd/emitter.go
```diff
@@ -0,0 +1,290 @@
...
+
+type ResponseFormatter interface {
+	WriteResponse(w io.Writer, resp Response) (int, error)
+	WriteError(w io.Writer, err error)
+	ContentType() string
+}
+
+func NewEncoder(r *http.Request, config *Config) Encoder {
+	epoch := strings.TrimSpace(r.FormValue("epoch"))
+	switch r.Header.Get("Accept") {
```
phemmer: The parsing of the `Accept` header needs to be more complex. It's possible for the client to accept multiple types. This will be a very likely occurrence once msgpack is introduced as clients will prefer msgpack, but also accept json.
For example:

```
Accept: application/x-msgpack, application/json;q=0.7
```

It's also technically valid to do:
```
Accept: application/x-msgpack
Accept: application/json
```
Though this is rather uncommon.

services/httpd/msgp_emitter.go
```diff
@@ -0,0 +1,160 @@
...
+	values := make([][]interface{}, 0, e.ChunkSize)
+
+	enc := msgp.NewWriter(w)
+	enc.WriteMapHeader(1)
+	enc.WriteString("results")
+	enc.WriteInt(header.Results)
+
+	enc.Flush()
+	if w, ok := w.(http.Flusher); ok {
+		w.Flush()
```
jsternberg: We'll have to figure out a good buffering size for this. For this to be a streaming protocol, it needs to not buffer the entire thing. I do think that we can probably remove some of these flushes though for speed. The first flush that sends the headers is likely necessary though and the flush after sending the result header is another one that seems useful.

services/httpd/msgp_emitter.go
```diff
@@ -0,0 +1,160 @@
...
+			if w, ok := w.(http.Flusher); ok {
+				w.Flush()
+			}
+			continue
+		}
+
+		for series := range result.SeriesCh() {
+			enc.WriteArrayHeader(2)
+			enc.WriteInt(1)
+			enc.WriteBool(true)
```
jsternberg: The reason for encoding it as an array is so that it's a single thing that can be decoded.

The reason why it doesn't use `msgp:tuple` is because I was unaware that existed. Thanks.

services/httpd/msgp_emitter.go
```diff
@@ -0,0 +1,160 @@
...
+
+			if series.Err != nil {
+				err := series.Err.Error()
+				header := SeriesError{Error: err}
+				header.EncodeMsg(enc)
```
jsternberg: The reason why I do this is because of a limitation with `tinylib/msgp`. The library doesn't take kindly to not encoding certain arguments and will instead encode a blank string with them instead of just omitting the entry. So I use two structures that are supposed to be one structure to trick msgp into generating the correct code.

The intention is that the `error` key is part of the `SeriesHeader`.

services/httpd/msgp_emitter.go
```diff
@@ -0,0 +1,160 @@
...
+					enc.WriteBool(false)
+					for _, v := range values {
+						val := Row{Value: v}
+						val.EncodeMsg(enc)
+					}
+					values = values[:0]
+
+					err := RowError{Error: row.Err.Error()}
+					err.EncodeMsg(enc)
+					continue
```
jsternberg: The intention is to drain the result set just in case there's a bug. That should be made explicit.

services/httpd/msgp_emitter.go
```diff
@@ -0,0 +1,160 @@
...
+	var convertToEpoch func(row *influxql.Row)
+	if e.Epoch != "" {
+		convertToEpoch = epochConverter(e.Epoch)
+	}
+	values := make([][]interface{}, 0, e.ChunkSize)
+
+	enc := msgp.NewWriter(w)
+	enc.WriteMapHeader(1)
+	enc.WriteString("results")
+	enc.WriteInt(header.Results)
```
phemmer: There are 2 ways the client knows when to stop reading a sequence of objects. Either the client knows how many objects to read, or it receives something which tells it to stop.

But my point is that having an indicator when to stop reading works for everything. Having a count that tells the client how many to read only works for some things (when you know what that count is). Thus for consistency (and less confusion), the former works better.

services/httpd/msgp_emitter.go
phemmer: Why do we need to flush after a certain type of data is sent, instead of flushing when buffer is full?
Or alternatively, what is the purpose of flushing headers when there is no data? The client can read the headers yes, but if it has no data, the headers don't serve much purpose.

services/httpd/msgp_emitter.go
```diff
@@ -0,0 +1,160 @@
...
+					continue
+				}
+
+				if convertToEpoch != nil {
+					convertToEpoch(&row)
+				}
+
+				if len(values) == cap(values) {
+					enc.WriteArrayHeader(2)
+					enc.WriteInt(len(values))
```
jsternberg: If you can give me an idea what you mean by true streaming, I'm all for that. The old response structure had a problem that it wasn't true streaming and the results were embedded in a weird way that made reading from it as a client impossible.

The reason why this encodes each row is so that it can read one row at a time. If you encode as an array, you have to decode the entire array rather than just the next row. Then you have to buffer the full results from the chunk. While this method requires maintaining state for how much you are going to read, that's just an integer and a boolean rather than an array. And you would have to still maintain state for the boolean (do I need to read more when I finish reading the array?) and for the array itself. So both maintain roughly the same state.

Now reading an array of rows as a chunk may perform better. If it does, we should consider the pros and cons. I haven't performed any testing of that.

services/httpd/emitter.go
jsternberg: If you make a PR for this, I would definitely review it and accept it.

services/httpd/msgp_emitter.go
phemmer: > The reason why I do this is because of a limitation with tinylib/msgp.

Should we let limitations of the library dictate the protocol?

> The intention is that the error key is part of the SeriesHeader.

It would need to be encoded as a map entry, or as a item in an array. Here it's just a standalone object which can show up unexpectedly on the wire.

services/httpd/msgp_emitter.go
jsternberg: It's possible to know the response header. I don't know about other HTTP libraries, but the Go one will block until it receives the headers. Once it receives the headers, you can read the body whenever you want.

So it's mostly to make Ctrl+C functionality possible. If you don't flush the headers immediately, the client won't be able to get to the section of code where handling an interrupt is possible.

A (non-compiling) example:

```
resp, err := http.DefaultClient.Do(req)
if err != nil {
    panic(err)
}
// Does not get here until the headers are sent back.
signals := make(os.Signal, 0, 4)
signal.Notify(signals, os.Interrupt)
go func() {
    <-signals
    resp.Body.Close()
}()

for {
    // read points from the stream
}
```

Something like that, but the code above doesn't work and has issues.

services/httpd/msgp_emitter.go
jsternberg: The limitation of the library doesn't dictate the protocol. The protocol has the `error` key in the series header. In this PR, I separate the structs so that the proper output is written to the wire. The protocol states that the series header is written and the series header contains an optional error key. If the error key is set, there was an error.

The gist describes the full protocol. I should update this PR to add the protocol as part of this PR.

phemmer: Oh, ok, I see what you're doing now. Using 2 different `struct`s with the intention that the client decodes them into the same one. Sorry, didn't realize that until now.

services/httpd/msgp_emitter.go
jsternberg: There's two advantages to having a count in the protocol. The first is that the client knows what to expect. So if you are trying to create a slice of all of the points, I know before reading any points how many are going to be in the current chunk and I also know if I'm expecting more. I can allocate an array that's appropriate.

The second advantage is there is less guessing in the protocol. If you have some kind of signal at the end that tells the stream if it has finished reading, then you have to determine what that signal is and try to find it. It adds an additional condition to the code and requires that you read at least one byte that you then have to buffer just in case that byte is not the EOF signal. If you use an actual EOF signal, that isn't very kind to reusing a socket connection. This is made more difficult because we would need multiple EOF signals. One for the end of a series, another for the end of a result, and another for an end to the response itself.

Since we know what we are going to send before we send it, the first seems less complicated from the client side and has fewer conditions.

jsternberg:
I have added a formal writeup of the cursor protocol. Please take a look at it and I will then update the code to match the changes I have made.

services/httpd/cursor.md
```diff
@@ -0,0 +1,162 @@
...
+then read the `RowBatchHeader`.
+
+    // RowBatchHeader contains information about the current batch.
+    type RowBatchHeader struct {
+        // Length contains the number of rows that are in this batch.
+        Length int
+
+        // Continue is set to true if there is another batch that
+        // should be read.
+        Continue bool
```
phemmer: One awkward thing about doing it this way is that it means the server has to delay the batch until it knows whether another batch will follow. While this is likely to be an infinitesimally small delay, it just feels weird.

jsternberg: It actually already needs to do this to be backwards compatible with the old format. It is a bit weird, but it's already there and makes it easier in the long run. If we change it in the future so the code doesn't do that, we can always mark one batch as `continue: true` and then have zero length for the next batch when we learn there aren't any more entries.

services/httpd/cursor.md
```diff
@@ -0,0 +1,162 @@
...
+
+        // Error is present if an error was encountered while
+        // emitting the rows for the query.
+        Error string
+    }
+
+The `RowChunkHeader` contains metadata for the rows that are about to be
```
jsternberg: Yea, I changed the name several times. I hate all of the names. Feel free to give feedback on them. I'm about ready to name them using `/dev/urandom` at this point.

services/httpd/cursor.md
```diff
@@ -0,0 +1,162 @@
...
+    {"length":2}
+    ["2010-01-01T00:00:00Z",2]
+    ["2010-01-01T00:00:00Z",3]
+    false
+    {"id":1}
+    true
+    {"name":"databases","columns":["name"]}
+    {"length":2}
+    {"values":["db0"]}
+    {"values":["db1"]}
```
jsternberg: Yes. I messed up here.

services/httpd/cursor.md
```diff
@@ -0,0 +1,163 @@
...
+a specific order.
+
+## Definitions
+
+Unless otherwise stated, every struct is meant to be encoded as the
+equivalent of a hash map.
+
+## MIME Type
+
+The MIME type for the InfluxDB Cursor Protocol is
```
jsternberg: Another possibility for this which might be better is `application/vnd.influxdb-cursor+msgpack; version=1.0`. So the version is a parameter instead of part of the accept header.

phemmer:
Just curious, is there any idea when this change might make it in?
I've been running a custom build based on the original code (as it was on Aug 7), with a custom client, and the performance is much better than json encoding, at least for the client. InfluxDB still bottlenecks on other things, but I have an nginx cache in front of InfluxDB to address that.

If we think this is close to the final iteration, I can do another build on top of a more recent master, update the client to match, and provide more feedback.

jsternberg:
I'm going to rebase this and modify the code to match the README. I don't think they currently match anymore. That will allow you to run whatever tests. If this performs better, then that may throw some momentum towards getting this merged.

@jwilder can you take a look at the cursor.md file contained in this PR so we can start the process of determining if this proposal is suitable?

jsternberg:
I have updated this PR so now the code aligns with the specification in `services/httpd/cursor.md`.

jsternberg:
We're going to be revisiting this idea in the future. We're working on a successor to influxql and are not going to be making this change to influxql at the current moment.

Sorry for the confusion.

phemmer:
For posterity, msgpack was added over in #8897

PR #8480, opened 2017-06-12
Change the default stats interval to 1 second instead of 10 seconds

jsternberg:
Update the configuration file's comment to mention the limitations of `max-select-point`.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated


jsternberg:
@rbetts this is separated into two commits so we can cherry pick the first commit to 1.3.

PR #8481, opened 2017-06-12
Decouple Subscriber from PointsWriter

jwilder:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

This decouples the `PointsWriter` from `Subscriber` a bit by making the `PointsWriter` maintain a slice of points channels as opposed to a 1:1 reference to subscriber services channel.  When the server is wired up, `Subscriber` service is added as a subscriber.  This also allows for adding other listeners type services that want to be notified when points are written.

### DIFF
--- a/coordinator/points_writer.go
+++ b/coordinator/points_writer.go
@@ -60,9 +60,6 @@ type PointsWriter struct {
 		WriteToShard(shardID uint64, points []models.Point) error
 	}
 
-	Subscriber interface {
-		Points() chan<- *WritePointsRequest
-	}
 	subPoints []chan<- *WritePointsRequest
 
 	stats *WriteStatistics
@@ -127,9 +124,6 @@ func (w *PointsWriter) Open() error {
 	w.mu.Lock()
 	defer w.mu.Unlock()
 	w.closing = make(chan struct{})
-	if w.Subscriber != nil {
-		w.AddWriteSubscriber(w.Subscriber.Points())
-	}
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


coordinator/points_writer.go
```diff
@@ -312,14 +310,18 @@ func (w *PointsWriter) WritePoints(database, retentionPolicy string, consistency
...
+		select {
+		case ch <- pts:
+			ok = true
+		default:
+		}
 	}
 	w.mu.RUnlock()
+
 	if ok {
 		atomic.AddInt64(&w.stats.SubWriteOK, 1)
```
nathanielc: This stat now counts if at least one succeed. Do we want to change to count if all succeeded instead of at least one succeeds?

jwilder: Good point.  I'll update it.

jwilder: Fixed in 71579fc

nathanielc:
@jwilder I was testing this (indirectly) and I have discovered that this change has caused all subscription points to be dropped. Reverting to 77afe50f7ebe8fc6cc2c635fb9971dfe4278578b fixes the issue. I took a look over the code and its not clear why but the `default` select case is always being taken ,https://github.com/influxdata/influxdb/pull/8481/files#diff-5cc502cf8a324b161bd7409ca0ed3655R320

jwilder:
I think I know what's going on.  I'll push up a fix.

PR #8491, opened 2017-06-14
Add support for TSI shard streaming and shard size

e-dard:
This commit firstly ensures that a shard's size on disk is accurately
reported when using the `tsi1` index, by including the on-disk size of the
`tsi1` index in the calculation.

Secondly, this commit add support for shard streaming/copying when using
the `tsi1` index. Prior to this, a `tsi1` index would not be correctly
restored when streaming shards.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - [#8495](https://github.com/influxdata/influxdb/pull/8495): Improve CLI connection warnings
 - [#9084](https://github.com/influxdata/influxdb/pull/9084): Handle high cardinality deletes in TSM engine
 - [#9162](https://github.com/influxdata/influxdb/pull/9162): Improve inmem index startup performance for high cardinality.
+- [#8491](https://github.com/influxdata/influxdb/pull/8491): Add further tsi support for streaming/copying shards.
 
 ### Bugfixes
 
@@ -93,8 +94,7 @@ refuse to open, and will most likely see the following error message:
 - [#8897](https://github.com/influxdata/influxdb/pull/8897): Add message pack format for query responses.
 - [#8886](https://github.com/influxdata/influxdb/pull/8886): Improved compaction scheduling
 - [#8690](https://github.com/influxdata/influxdb/issues/8690): Implicitly decide on a lower limit for fill queries when none is present.
-- [#8947](https://github.com/influxdata/influxdb/pull/8947): Add `EXPLAIN ANALYZE` command, which produces a detailed execution plan of a `SELECT` statement.
-- [#8963](https://github.com/influxdata/influxdb/pull/8963): Streaming inmem2tsi conversion.
+- [#8947](https://github.com/influxdata/influxdb/pull/8947): Add `EXPLAIN ANALYZE` command, which produces a detailed execution plan of a `SELECT` statement.- [#8963](https://github.com/influxdata/influxdb/pull/8963): Streaming inmem2tsi conversion.
 - [#8995](https://github.com/influxdata/influxdb/pull/8995): Sort & validate TSI key value insertion. 
 - [#8968](https://github.com/influxdata/influxdb/issues/8968): Make client errors more helpful on downstream errs. Thanks @darkliquid!
 - [#8984](https://github.com/influxdata/influxdb/pull/8984): EXACT and estimated CARDINALITY queries.
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -123,8 +123,9 @@ const (
 
 // Engine represents a storage engine with compressed blocks.
 type Engine struct {
-	mu     sync.RWMutex
-	_index tsdb.Index
+	mu sync.RWMutex
+
+	index tsdb.Index
 
 	// The following group of fields is used to track the state of level compactions within the
 	// Engine. The WaitGroup is used to monitor the compaction goroutines, the 'done' channel is
@@ -197,7 +198,7 @@ func NewEngine(id uint64, idx tsdb.Index, database, path string, walPath string,
 		id:           id,
 		database:     database,
 		path:         path,
-		_index:       idx,
+		index:        idx,
 		logger:       logger,
 		traceLogger:  logger,
 		traceLogging: opt.Config.TraceLoggingEnabled,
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -33,18 +37,22 @@ The admin UI is removed and unusable in this release. The `[admin]` configuratio
...
-  
+<<<<<<< HEAD
+  the configured maximum size, a `413 Request Entity Too Large` HTTP response is returned.
+
 #### `[continuous_queries]` Section
 
 * `query-stats-enabled` was added with a default of `false`. When set to `true`, continuous query execution statistics are written to the default monitor store.
+=======
+  the configured maximum size, a `413 Request Entity Too Large` HTTP response is returned.
+>>>>>>> Add support for TSI shard streaming and shard size
```
benbjohnson: Merge conflict?

tsdb/engine/tsm1/engine.go
```diff
@@ -666,6 +662,8 @@ func (e *Engine) overlay(r io.Reader, basePath string, asNew bool) error {
 			return nil, err
 		}
 
+		// The filestore will only handle tsm files. Other file types will be
+		// ignored.
```
benbjohnson: Why must you leave orphaned words in comment lines!?!? :rage:

tsdb/engine/tsm1/file_store_test.go
```diff
@@ -2253,9 +2253,9 @@ func TestFileStore_Stats(t *testing.T) {
 		"mem": []tsm1.Value{tsm1.NewValue(0, 1.0)},
 	})
 
-	replacement := files[2] + "-foo" + ".tmp" // Assumes new files have a .tmp extension
+	replacement := fmt.Sprintf("%s.%s.%s", files[2], tsm1.TmpTSMFileExtension, tsm1.TSMFileExtension) // Assumes new files have a .tmp extension
```
benbjohnson: Should Tmp extension be at the end?

tsdb/engine/tsm1/engine.go
e-dard: Your rage brings me so much joy ❤️ .

tsdb/engine/tsm1/file_store_test.go
e-dard: @benbjohnson I changed the implementation to check for either `.tsm.tmp` or `.tsm`. The previous test was creating a filename ending in `-foo.tmp`, which didn't match the new filters, so I updated the test to come up with a new filename that matched one of the filters.

tsdb/engine/tsm1/engine.go
```diff
@@ -1093,9 +1101,15 @@ func (e *Engine) CreateSnapshot() (string, error) {
 	}
 
 	e.mu.RLock()
-	defer e.mu.RUnlock()
+	path, err := e.FileStore.CreateSnapshot()
+	if err != nil {
+		e.mu.RUnlock()
+		return "", err
+	}
+	e.mu.RUnlock()
```
jwilder: I think the engine should still be locked while the index is snapshotted.  If it's unlocked, it's possible the index and tsm files could be inconsistent if deletes and writes were running.

e-dard: Looks like this was a pre-existing bug in that case. I'll add the lock protection.

CHANGELOG.md
```diff
@@ -107,7 +108,11 @@ The following new configuration options are available.
 * `max-body-size` was added with a default of 25,000,000, but can be disabled by setting it to 0.
   Specifies the maximum size (in bytes) of a client request body. When a client sends data that exceeds
   the configured maximum size, a `413 Request Entity Too Large` HTTP response is returned.
+<<<<<<< HEAD
```
jwilder: Bad merge...

tsdb/engine/tsm1/engine.go
```diff
@@ -707,15 +705,27 @@ func (e *Engine) overlay(r io.Reader, basePath string, asNew bool) error {
 		return err
 	}
 
+	if idx, ok := e.index.(*tsi1.Index); ok {
```
jwilder: I think this may create a race now since it's accessing the index w/o a RLock.

jwilder: There may be other accesses to `e.index` that need to be protected now too.

e-dard:
@jwilder is this good to go now?

@benbjohnson can you check 4d14ff058181bda032e9fbed702f4255381be099. I didn't include these fields originally and as I was rebasing it occurred to me that we probably need to set them.

tsdb/engine/tsm1/file_store.go
```diff
@@ -363,8 +368,9 @@ func (f *FileStore) Open() error {
 	if err != nil {
 		return err
 	}
+	ext := fmt.Sprintf(".%s", TmpTSMFileExtension)
 	for _, fi := range tmpfiles {
-		if fi.IsDir() && strings.HasSuffix(fi.Name(), ".tmp") {
+		if fi.IsDir() && strings.HasSuffix(fi.Name(), ext) {
```
jwilder: This is missing the "." now.

tsdb/engine/tsm1/file_store.go
```diff
@@ -565,16 +571,20 @@ func (f *FileStore) replace(oldFiles, newFiles []string, updatedFn func(r []TSMF
 	f.mu.RUnlock()
 
 	updated := make([]TSMFile, 0, len(newFiles))
+	tsmTmpExt := fmt.Sprintf("%s.%s", TSMFileExtension, TmpTSMFileExtension)
 
 	// Rename all the new files to make them live on restart
 	for _, file := range newFiles {
 		var newName = file
-		if strings.HasSuffix(file, ".tmp") {
+		if strings.HasSuffix(file, tsmTmpExt) {
```
jwilder: "." is missing as well.

tsdb/index/tsi1/index.go
```diff
@@ -826,6 +879,13 @@ func (i *Index) TagSets(name []byte, opt query.IteratorOptions) ([]*query.TagSet
 	return sortedTagsSets, nil
 }
 
+// DiskSizeBytes returns the size of the index on disk.
+func (i *Index) DiskSizeBytes() int64 {
+	fs := i.RetainFileSet()
+	defer fs.Release()
+	return fs.Size()
```
jwilder: `Shard.DiskSize` is called fairly frequently.  It created a lot of lock contention in tsm and I suspect it will cause a problem here as well.  I ended up switching the sizes in tsm to be atomic counters to fix it.

tsdb/engine/tsm1/engine.go
jwilder: `e.index` is still accessed outside of a `RLock`.

tsdb/index/tsi1/index.go
e-dard: @jwilder this calls down to [here](https://github.com/influxdata/influxdb/blob/master/tsdb/index/tsi1/index_file.go#L122) on each file in the fileset. It's should be cheap.

e-dard:
@jwilder see c61f533a4d256cffda0736a1fa00a000066c034f

benbjohnson:
@e-dard 👍 on 4d14ff0.

tsdb/engine/tsm1/file_store.go
```diff
@@ -635,16 +641,20 @@ func (f *FileStore) replace(oldFiles, newFiles []string, updatedFn func(r []TSMF
...
 			// The new TSM files have a tmp extension.  First rename them.
 			newName = file[:len(file)-4]
 			if err := os.Rename(file, newName); err != nil {
 				return err
 			}
+		} else if !strings.HasSuffix(file, TSMFileExtension) {
```
jwilder: I think this should be `TmpTSMFileExtension`.  New files passed in are always tmp one.

e-dard: That check happens in the `if` above. It checks for `tsmTmpExt` which is `.tsm.tmp`. This `else if` skips any non-tsm or non-tmp-tsm files (such as index files).

In the `if` above we rename the `tmp` tsm file and drop the `.tmp`. Then below this `else if` we process the `tsm` file.

Without this `else if` to skip non-tsm files I believe we could get caught up on `tsi` index files.

PR #8521, opened 2017-06-22
switch tsi debugging info to logs/errors

oiooj:
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated


benbjohnson:
@oiooj The `assert()` calls exist as invariants in the code and they don't return an error because there's nothing to be done to handle them. They should always be true and if they're not then something is fundamentally wrong.

Returning or logging an error and stopping compaction causes a larger issue where files would simply get backed up and clog the system. I think that panicking immediately in these situations is the best course of action.

PR #8523, opened 2017-06-23
8426 add a config option so collectd can generate multiple fields, not multiple measurements

aanthony1243:
- [X] Rebased/mergable
- [X] Tests pass
- [X] CHANGELOG.md updated

- [X] Config changes: update sample config (`etc/config.sample.toml`), server `NewDemoConfig` method, and `Diagnostics` methods reporting config settings, if necessary
- [X] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): issue filed or pull request submitted <Doc updated implicitly via new Readme.md on this PR>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,20 @@
 ## v1.4.0 [unreleased]
 
+### Configuration Changes
+
+#### `[collectd]` Section
+
+* `parse-multivalue-plugin` was added with a default of `split`.  When set to `split`, multivalue plugin data will be split into separate measurements.  When set to `join`, multivalue plugin will be stored as a single multi-value measurement.  
+
+### Features
+
+- [#8426] (https://github.com/influxdata/influxdb/issues/8426): Add `parse-multivalue-plugin` to allow users to choose how multivalue plugins should be handled by the collectd service.  
+
 ### Bugfixes
 
 - [#8480](https://github.com/influxdata/influxdb/pull/8480): Change the default stats interval to 1 second instead of 10 seconds.
 - [#8466](https://github.com/influxdata/influxdb/issues/8466): illumos build broken on syscall.Mmap
-- [#8426](https://github.com/influxdata/influxdb/issues/8426): some collectd information that was split into separate measurements can now be grouped
+
 ## v1.3.0 [unreleased]
 
 ### Release Notes
@@ -46,6 +56,10 @@ The following new configuration options are available.
 #### `[continuous_queries]` Section
 
 * `query-stats-enabled` was added with a default of `false`. When set to `true`, continuous query execution statistics are written to the default monitor store.
+
+#### `[collectd]` Section
+
+* `parse-multivalue-plugin` was added with a default of `split`.  When set to `split`, multivalue plugin data will be split into separate measurements.  When set to `join`, multivalue plugin will be stored as a single multi-value measurement.  
 
 ### Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,11 @@
 
 #### `[collectd]` Section
 
-* `parse-multivalue-plugin` was added with a default of `split`.  When set to `split`, multivalue plugin data will be split into separate measurements.  When set to `join`, multivalue plugin will be stored as a single multi-value measurement.  
+* `parse-multivalue-plugin` was added with a default of `split`.  When set to `split`, multivalue plugin data (e.g. df free:5000,used:1000) will be split into separate measurements (e.g., (df_free, value=5000) (df_used, value=1000)).  When set to `join`, multivalue plugin will be stored as a single multi-value measurement (e.g., (df, free=5000,used=1000)).  
 
 ### Features
 
-- [#8426] (https://github.com/influxdata/influxdb/issues/8426): Add `parse-multivalue-plugin` to allow users to choose how multivalue plugins should be handled by the collectd service.  
+- [#8426](https://github.com/influxdata/influxdb/issues/8426): Add `parse-multivalue-plugin` to allow users to choose how multivalue plugins should be handled by the collectd service.  
 
 ### Bugfixes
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 ## v1.4.0 [unreleased]
 
+### Configuration Changes
+
+#### `[collectd]` Section
+
+* `parse-multivalue-plugin` was added with a default of `split`.  When set to `split`, multivalue plugin data will be split into separate measurements.  When set to `join`, multivalue plugin will be stored as a single multi-value measurement.
```
jwilder: This config option needs to be added to the sample config in `etc/config.sample.toml` too.

services/collectd/service.go
```diff
@@ -385,6 +390,53 @@ func (s *Service) writePoints() {
...
+// The difference is that the original provided measurements like (PLUGIN_DSNAME, ["value",xxx])
+// while this one will provide measurements like (PLUGIN, {["DSNAME",xxx]})
+// this effectively joins collectd data that should go together, such as:
+//  (df, {["used",1000],["free",2500]})
+func (s *Service) UnmarshalValueListPacked(vl *api.ValueList) []models.Point {
+	timestamp := vl.Time.UTC()
+
+	var name = vl.Identifier.Plugin
+	tags := make(map[string]string)
+	fields := make(map[string]interface{})
```
mark-rushakoff: Although Go allows you to `make` maps (and slices) without specifying a size, but we've run into many bottlenecks that were (at least partially) addressed by specifying an initial size hint. 

Here, I'd go with 
```
tags := make(map[string]string, 4)
fields := make(map[string]interface{}, len(vl.Values))
```
We might not use all 4 `tags` slots, which is fine, but giving `fields` the right capacity from the start will avoid creating some garbage as the map is grown from a smaller capacity.

services/collectd/service.go
```diff
@@ -385,6 +390,53 @@ func (s *Service) writePoints() {
...
+	if vl.Identifier.Type != "" {
+		tags["type"] = vl.Identifier.Type
+	}
+	if vl.Identifier.TypeInstance != "" {
+		tags["type_instance"] = vl.Identifier.TypeInstance
+	}
+
+	for i := range vl.Values {
+		fieldName := vl.DSName(i)
+		switch value := vl.Values[i].(type) {
```
mark-rushakoff: It's generally preferable to use the two-argument range function rather than indexing again, i.e.

```
for i, v := range vl.Values {
  fieldName := vl.DSName(i)
  switch val := v.(type) {
```

services/collectd/service.go
```diff
@@ -385,6 +390,53 @@ func (s *Service) writePoints() {
...
+		case api.Counter:
+			fields[fieldName] = float64(value)
+		}
+	}
+	// Drop invalid points
+	p, err := models.NewPoint(name, models.NewTags(tags), fields, timestamp)
+	if err != nil {
+		s.Logger.Info(fmt.Sprintf("Dropping point %v: %v", name, err))
+		atomic.AddInt64(&s.stats.InvalidDroppedPoints, 1)
+	}
```
mark-rushakoff: If `err` is not nil, we may not be sure of the state of `p`. This block continues to the return statement. Should it `return nil` instead?

services/collectd/service.go
```diff
@@ -385,6 +390,53 @@ func (s *Service) writePoints() {
...
+
+	// return as an array for compatability with rest of API
```
mark-rushakoff: Probably not a necessary comment, and it's returning a slice, not an array. https://blog.golang.org/go-slices-usage-and-internals

services/collectd/service_test.go
```diff
@@ -260,13 +260,73 @@ func TestService_BatchSize(t *testing.T) {
 	}
 }
 
+// Test that the collectd service correctly batches points by BatchSize.
```
mark-rushakoff: Looks like a copy-pasted comment.

services/collectd/service_test.go
```diff
@@ -260,13 +260,73 @@ func TestService_BatchSize(t *testing.T) {
...
+	}
+
+	// Send the test data to the service.
+	if n, err := conn.Write(testData); err != nil {
+		t.Fatal(err)
+	} else if n != len(testData) {
+		t.Fatalf("only sent %d of %d bytes", n, len(testData))
+	}
+
+	points := []models.Point{}
```
mark-rushakoff: Since you're only appending to this, it's okay to write `var points []models.Point`. Go will do the right thing when you say `points = append(points, foo)`.

services/collectd/service_test.go
```diff
@@ -260,13 +260,73 @@ func TestService_BatchSize(t *testing.T) {
...
+
+Loop:
+	for {
+		select {
+		case p := <-pointCh:
+			points = append(points, p)
+			if len(points) == totalPoints {
+				break Loop
+			}
+		case <-time.After(time.Second):
```
mark-rushakoff: Use of `time.After` in a  loop like this is an antipattern. 

Every time the select block is entered, a new timer is created. The timer continues to run in the background for the full second, and there's overhead in the runtime for managing any running timer.

It's better to use a single instance of [`time.Timer`](https://golang.org/pkg/time/#Timer) and call `Reset` on each iteration of the loop.

CHANGELOG.md
aanthony1243: forgot to push...it's there now.

mark-rushakoff: It's hard to follow what a multivalue plugin means exactly. Is this something that collectd users are expected to know? If not, it would be helpful to have an example of the differences between split and join.

services/collectd/service.go
aanthony1243: This snippet was pulled from another test, should I fix the other tests too?

services/collectd/service_test.go
aanthony1243: done.

services/collectd/service.go
aanthony1243: done.

services/collectd/service.go
```diff
@@ -385,6 +390,53 @@ func (s *Service) writePoints() {
 	}
 }
 
+// UnmarshalValueListPacked is an alternative to the original UnmarshalValueList
```
aanthony1243: done.

PR #8540, opened 2017-06-28
Fix TSI issue with spaces in tag values (1.3)

benbjohnson:
- [x] Rebased/mergable
- [x] Tests pass

### DIFF
--- a/tsdb/index/tsi1/index.go
+++ b/tsdb/index/tsi1/index.go
@@ -707,7 +707,7 @@ func (i *Index) TagSets(name []byte, opt influxql.IteratorOptions) ([]*influxql.
 				}
 			}
 			// Associate the series and filter with the Tagset.
-			tagSet.AddFilter(string(SeriesElemKey(e)), e.Expr())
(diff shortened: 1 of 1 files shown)


tsdb/index/tsi1/index.go
```diff
@@ -707,7 +707,7 @@ func (i *Index) TagSets(name []byte, opt influxql.IteratorOptions) ([]*influxql.
 				}
 			}
 			// Associate the series and filter with the Tagset.
-			tagSet.AddFilter(string(SeriesElemKey(e)), e.Expr())
+			tagSet.AddFilter(string(models.MakeKey(e.Name(), e.Tags())), e.Expr())
```
e-dard: Is this only for read-only operations, because it seems like the behaviour of `SeriesElemKey` and `MakeKey` are quite different? I ask only in terms of backwards compatibility if this result is persisted anywhere.

benbjohnson: Yes, it's just for read operations. It needs to match the format that the query engine uses internally.

PR #8550, opened 2017-06-29
Allow panic recovery mechanism to be disabled

e-dard:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

This PR adds a new environment variable `INFLUXDB_PANIC_CRASH`, which
when set to a truthy value, e.g., `true`, `TRUE`, `1`, will prevent bot the HTTP server
and the `QueryExecutor` from recovering from a panic.

When `INFLUXDB_PANIC_CRASH` is switched on, the panic will be emitted to the logs and then a further emission of the current stacktrace of all goroutines will follow.

Further, this commit adds `_internal` stats that will monitor the
occurrence of panics all the time (regardless of if `INFLUXDB_PANIC_CRASH`
has been set to `true` or not).

The recovered panic frequency can be inspected with the following
queries:

`SELECT "recoveredPanics" FROM "_internal"."monitor"."httpd";`
`SELECT "recoveredPanics" FROM "_internal"."monitor"."queryExecutor";`

### DIFF
--- a/influxql/query_executor.go
+++ b/influxql/query_executor.go
@@ -3,7 +3,9 @@ package influxql
 import (
 	"errors"
 	"fmt"
+	"os"
 	"runtime/debug"
+	"strconv"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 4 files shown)


services/httpd/handler.go
```diff
@@ -1186,16 +1189,33 @@ func (h *Handler) responseWriter(inner http.Handler) http.Handler {
...
+		willCrash = false
+	}
+
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		start := time.Now()
 		l := &responseLogger{w: w}
 
 		defer func() {
 			if err := recover(); err != nil {
+
```
jsternberg: Nitpick: Why the added newline?

e-dard:
@jsternberg updated, including moving env var checks to `init` (thanks @joelegasse) and also panicking in `QueryExecutor` to let any possible defers run. Maintained `os.Exit` in http Handler since the Go server will recover another `panic` if I used that instead.

influxql/query_executor.go
```diff
@@ -392,13 +402,33 @@ LOOP:
...
 			StatementID: -1,
 			Err:         fmt.Errorf("%s [panic:%s]", query.String(), err),
 		}
+
+		if willCrash {
+			e.Logger.Error(fmt.Sprintf("\n\n=====\nAll goroutines now follow:"))
+			buf := make([]byte, 1<<16)
+			stackSize := runtime.Stack(buf, true)
+			e.Logger.Error(fmt.Sprintf("%s\n", string(buf[0:stackSize])))
+			panic(err) // Ensure any defers run
```
jsternberg: This will produce a double stack trace, wouldn't it? Is it possible to rethrow the panic and maintain the original stack instead?

e-dard: @jsternberg I thought about this but I wasn't aware of a way to do that. I think once `recover()` has been executed you're done with the original panic and you can't "rethrow" it can you? Given that I was left with `panic(err)` or `panic(nil)` for example.

PR #8553, opened 2017-06-30
Updated Makefile to support installing

kulinacs:
Update the Makefile to support the standard options found in telegraf


jsternberg:
We do not use the Makefile in this project so I'm going to close this. I am removing the file in #8917.

PR #8557, opened 2017-07-02
Introduce a new dynamic language mechanism

jsternberg:
The language is now defined in a way similar to many HTTP routers with
the left prefix being placed into a parse tree and then eventually
invoking a function to parse the arguments.

This allows dynamically adding additional components to the parse tree
for either query language extensions or enterprise.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/influxql/parse_tree.go
+++ b/influxql/parse_tree.go
@@ -1,0 +1,217 @@
+package influxql
+
+import "fmt"
+
+var Language = &ParseTree{}
+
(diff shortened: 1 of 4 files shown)


benbjohnson:
The code itself looks fine. I'm a little hesitant about making the initial token parsing different than token parsing after the `ParseTree` but it's probably not a big deal.

PR #8568, opened 2017-07-06
Compress tombstone files

### LINKED ISSUE #8558: Dropping measurement used several GB disk space
### Bug report

Version: e3918e0
OS: Linux

__Steps to reproduce:__

1. `drop measurement foo`

Reduce disk usage

Increased disk usage

Before running `drop measurement`:
```
# df -h .
Filesystem                Size  Used Avail Use% Mounted on
/dev/mapper/sys-influxdb   30G   16G   13G  57% /mnt/influxdb
```
After running `drop measurement`:
```
# df -h .
Filesystem                Size  Used Avail Use% Mounted on
/dev/mapper/sys-influxdb   30G   23G  5.8G  80% /mnt/influxdb
```

In this specific case I was trying to remove unneeded data to address performance issues, not disk usage issues. But if I were low on disk space, and trying to remove data to get some back, this would be a big problem. I would not expect dropping a measurement to result in using more disk space. Especially an additional 43% of the original size. If I were low on disk space, dropping a measurement in attempt to get some space back could end up causing me to run out of space entirely.

jwilder:
This adds a v3 tombstone format that is a gzip compressed version of the v2
format.  It reduces the size of tombstone files substantially without
having to support a more feature rich file format for tombstones.  Prior versions can still be read, but new tombstone will use the v3 format.

Testing with 1M series and dropping all of them in my local env reduced disk usage for tombstone files from 397MB to 11MB.  Each file is about 98% smaller.

### Before
```
$ du -c -h *.tombstone
 99M	000000004-000000003.tombstone
 99M	000000008-000000003.tombstone
 99M	000000011-000000002.tombstone
 99M	000000012-000000001.tombstone
397M	total
```

### After
```
$ du -c -h *.tombstone
2.8M	000000004-000000003.tombstone
2.8M	000000008-000000003.tombstone
2.8M	000000011-000000002.tombstone
2.8M	000000012-000000001.tombstone
 11M	total
```

Fixes #8558 

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/tsdb/engine/tsm1/tombstone.go
+++ b/tsdb/engine/tsm1/tombstone.go
@@ -2,6 +2,7 @@ package tsm1
 
 import (
 	"bufio"
+	"compress/gzip"
 	"encoding/binary"
 	"io"
 	"io/ioutil"
@@ -13,8 +14,9 @@ import (
 )
 
 const (
-	v2header     = 0x1502
-	v2headerSize = 4
+	headerSize = 4
+	v2header   = 0x1502
+	v3header   = 0x1503
 )
 
 // Tombstoner records tombstones when entries are deleted.
@@ -164,7 +166,10 @@ func (t *Tombstoner) Walk(fn func(t Tombstone) error) error {
 		return err
 	}
 
-	if binary.BigEndian.Uint32(b[:]) == v2header {
+	header := binary.BigEndian.Uint32(b[:])
+	if header == v3header {
+		return t.readTombstoneV3(f, fn)
+	} else if header == v2header {
 		return t.readTombstoneV2(f, fn)
 	}
 	return t.readTombstoneV1(f, fn)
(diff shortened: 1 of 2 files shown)


tsdb/engine/tsm1/tombstone.go
```diff
@@ -323,6 +338,70 @@ func (t *Tombstoner) readTombstoneV2(f *os.File, fn func(t Tombstone) error) err
...
+	b := make([]byte, 4096)
+	for {
+		if _, err = io.ReadFull(gr, b[:4]); err == io.EOF || err == io.ErrUnexpectedEOF {
+			return nil
+		} else if err != nil {
+			return err
+		}
+		n += 4
+
+		keyLen := int(binary.BigEndian.Uint32(b[:4]))
```
e-dard: Paranoid me... But.. On a 32-bit system isn't it going to be possible for the `keyLen` to be a value larger than `math.MaxInt32`, and therefore when we convert to an `int` (which will be an `int32` on a 32-bit system), we may overflow.

Probably unlikely, but it could be a bit safer to do: 

```go
keyLen := binary.BigEndian.Uint32(b[:4])
if keyLen > uint32(len(b)) {
...
}
```

Just a thought.

tsdb/engine/tsm1/tombstone.go
```diff
@@ -323,6 +338,70 @@ func (t *Tombstoner) readTombstoneV2(f *os.File, fn func(t Tombstone) error) err
...
 
+// readTombstoneV3 reads the third version of tombstone files that are capable
+// of storing keys and the range of time for the key that points were deleted. This
+// format is a binary and compressed with gzip.
+func (t *Tombstoner) readTombstoneV3(f *os.File, fn func(t Tombstone) error) error {
+	// Skip header, already checked earlier
+	if _, err := f.Seek(headerSize, io.SeekStart); err != nil {
+		return err
+	}
+	n := int64(headerSize)
```
e-dard: Aside from keeping track of the bytes read, you don't seem to be doing anything with `n`? I can't find a read anywhere.

tsdb/engine/tsm1/tombstone.go
```diff
@@ -181,30 +186,40 @@ func (t *Tombstoner) writeTombstone(tombstones []Tombstone) error {
 
 	bw := bufio.NewWriterSize(tmp, 1024*1024)
 
-	binary.BigEndian.PutUint32(b[:4], v2header)
+	binary.BigEndian.PutUint32(b[:4], v3header)
 	if _, err := bw.Write(b[:4]); err != nil {
 		return err
 	}
 
+	if err := bw.Flush(); err != nil {
```
jwilder: Yes.  It might not be needed though. I was getting some weird decompresssion results when testing and added this, but it was actually that I was not using `io.ReadFull`

tsdb/engine/tsm1/tombstone.go
jwilder: That's copy/paste leftover from the v2 version.  I'll remove it since it use `io.EOF` now to check for termination.

tsdb/engine/tsm1/tombstone.go
jwilder: The max key length is 65k so we shouldn't ever get near a value that can overflow.

jwilder:
@e-dard All fixed.

PR #8572, opened 2017-07-06
Log server error response messages

e-dard:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

This PR provides more insight into server errors by both setting
the error on a response header, and, in the case of server errors `(5xx)`,
logging those error messages to the HTTP log, if:

```toml
[http] 
  log-enabled = true
```

So, a client error would be available via the response's `X-InfluxDB-Error` header, but wouldn't appear in the server logs. A `500` error would be in both the header and also appear in the server logs. In that case it would look something like:

```
[I] 2017-07-06T16:46:12Z Storing statistics in database '_internal' retention policy 'monitor', at interval 10s service=monitor
[I] 2017-07-06T16:46:12Z Sending usage statistics to usage.influxdata.com
[httpd] ::1 - - [06/Jul/2017:17:46:14 +0100] "POST /write?consistency=all&db=&precision=ns&rp= HTTP/1.1" 500 33 "-" "InfluxDBShell/unknown" 9fc4ecf8-626a-11e7-8001-000000000000 241
[E] 2017-07-06T16:46:14Z [500] - "some annoying error from deep within" service=httpd
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - [#8669](https://github.com/influxdata/influxdb/pull/8669): TSI Index Migration Tool
 - [#7195](https://github.com/influxdata/influxdb/issues/7195): Support SHOW CARDINALITY queries.
 - [#8711](https://github.com/influxdata/influxdb/pull/8711): Batch up writes for monitor service
+- [#8572](https://github.com/influxdata/influxdb/pull/8572): All errors from queries or writes are available via X-InfluxDB-Error header, and 5xx error messages will be written to server logs.
 
 ### Bugfixes
 
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"math"
 	"net/http"
 	"os"
 	"runtime/debug"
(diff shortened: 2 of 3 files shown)


jwilder:
Can we do this for `/write` as well?  At least logging the error that caused the 500?

e-dard:
@jwilder this should work for anything that calls `h.httpError`, which is currently both writes and queries.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 - [#8426](https://github.com/influxdata/influxdb/issues/8426): Add `parse-multivalue-plugin` to allow users to choose how multivalue plugins should be handled by the collectd service.
 - [#8548](https://github.com/influxdata/influxdb/issues/8548): Allow panic recovery to be disabled when investigating server issues.
+- [#8572](https://github.com/influxdata/influxdb/pull/8572): All errors from queries or writes are available via X-INFLUXDB-ERROR header, and 5xx error messages will be written to server logs.
```
jsternberg: Nitpick: Can you have this use the normalized header value like `Influxdb-Error` in the changelog or at least `InfluxDB-Error`? It's purely a nitpick, but I don't like the changelog yelling at me 😛

services/httpd/handler.go
```diff
@@ -1177,6 +1181,14 @@ func (h *Handler) logging(inner http.Handler, name string) http.Handler {
 		l := &responseLogger{w: w}
 		inner.ServeHTTP(l, r)
 		h.CLFLogger.Println(buildLogLine(l, r, start))
+
+		// Always log server errors.
```
jsternberg: Can this comment be changed to "Log server errors"? There's no mechanism to log non-server errors so saying we always log server errors gives no contrast to any other type of logging we do for errors. Also, it seems to me like this should say, "Log server error if present in the HTTP header."

services/httpd/handler.go
```diff
@@ -954,14 +955,17 @@ func parseSystemDiagnostics(d *diagnostics.Diagnostics) (map[string]interface{},
 }
 
 // httpError writes an error to the client in a standard format.
-func (h *Handler) httpError(w http.ResponseWriter, error string, code int) {
```
jsternberg: I think `error` is a better parameter name. At the least, `errmsg` so that the parameter name describes that it's expecting the text of an error.

CHANGELOG.md
e-dard: Yeah I should have changed this...

services/httpd/handler.go
e-dard: I'll settle on `errmsg`. I don't like `error` as a variable name as it's also a predeclared identifier and my editor paints it all bright purple. 😄

PR #8576, opened 2017-07-07
Fix index inconsistency after deletes

jwilder:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

This fixes an issue where the inmem index would be inconsistent with the data stored on disk.  It also fixes a bug where deletes do not write tombstones for data that should be deleted.

The inconsistency can occur under write load to a measurement and concurrently dropping the measurement.  There is a race where the incoming writes create index entries, but the writes have not hit the cache/wal yet.  The delete completes and removes them series from the index and then the write completes and stores the data on disk.  The end result is that the data is on disk, but the index is inconsistent and is not aware that it exists within the shard.  Restarting the server causes the index to be reloaded and the data appears again.

There wasn't a good way to fix this without significantly reworking how the index and engine work together or introducing a lot of lock contention mainly to handle this cases of deletes.  I went with an approach that refreshes the index from the newly compacted files as well as what is remaining in the cache when the files are installed for use.  This resolves the issue, but there can be a brief period of time between when the the delete removes the series from the index and before the index is refreshed where series can appear missing.

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -179,6 +179,8 @@ func NewEngine(id uint64, idx tsdb.Index, database, path string, walPath string,
 		compactionLimiter: opt.CompactionLimiter,
 	}
 
+	fs.OnReplace = e.onFileStoreReplace
+
 	// Attach fieldset to index.
 	e.index.SetFieldSet(e.fieldset)
 
@@ -521,7 +523,7 @@ func (e *Engine) LoadMetadataIndex(shardID uint64, index tsdb.Index) error {
 			return err
 		}
 
-		if err := e.addToIndexFromKey(key, fieldType, index); err != nil {
+		if err := e.addToIndexFromKey(key, fieldType); err != nil {
 			return err
 		}
 		return nil
@@ -536,7 +538,7 @@ func (e *Engine) LoadMetadataIndex(shardID uint64, index tsdb.Index) error {
 			e.logger.Info(fmt.Sprintf("error getting the data type of values for key %s: %s", key, err.Error()))
 		}
 
-		if err := e.addToIndexFromKey([]byte(key), fieldType, index); err != nil {
+		if err := e.addToIndexFromKey([]byte(key), fieldType); err != nil {
 			return err
 		}
 		return nil
@@ -735,7 +737,7 @@ func (e *Engine) overlay(r io.Reader, basePath string, asNew bool) error {
 			return err
 		}
 
-		if err := e.addToIndexFromKey(v.key, fieldType, e.index); err != nil {
+		if err := e.addToIndexFromKey(v.key, fieldType); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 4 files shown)


tsdb/engine/tsm1/engine.go
```diff
@@ -1227,9 +1229,59 @@ func (e *Engine) compactTSMFull(quit <-chan struct{}) {
 				// Release the files in the compaction plan
 				e.CompactionPlan.Release(s.compactionGroups)
 			}
+		}
+	}
+}
+
+// onFileStoreReplace is callback handler invoked when the FileStore
+// has replace one set of TSM files with a new set.
```
e-dard: nit `s/replace/replaced`

tsdb/engine/tsm1/engine.go
```diff
@@ -1227,9 +1229,59 @@ func (e *Engine) compactTSMFull(quit <-chan struct{}) {
...
+func (e *Engine) onFileStoreReplace(newFiles []TSMFile) {
+	// Load any new series keys to the index
+	readers := make([]chan seriesKey, 0, len(newFiles))
+	for _, r := range newFiles {
+		ch := make(chan seriesKey, 1)
+		readers = append(readers, ch)
+
+		go func(c chan seriesKey, r TSMFile) {
+			n := r.KeyCount()
```
e-dard: nit: you could just put `i < r.KeyCount()`

tsdb/engine/tsm1/engine.go
```diff
@@ -1227,9 +1229,59 @@ func (e *Engine) compactTSMFull(quit <-chan struct{}) {
...
+		if err := e.addToIndexFromKey(v.key, fieldType); err != nil {
+			continue
 		}
 	}
+
+	// load metadata from the Cache
+	e.Cache.ApplyEntryFn(func(key string, entry *entry) error {
+		fieldType, err := entry.values.InfluxQLType()
+		if err != nil {
+			// ignore
```
e-dard: Could we at least log any errors here, or are we expecting them?

tsdb/engine/tsm1/engine.go
```diff
@@ -1227,9 +1229,59 @@ func (e *Engine) compactTSMFull(quit <-chan struct{}) {
...
+		}
+
+		if err := e.addToIndexFromKey([]byte(key), fieldType); err != nil {
+			// ignore
+			return nil
+		}
+		return nil
+	})
+
+	return
```
e-dard: This `return` is redundant.

tsdb/engine/tsm1/reader.go
```diff
@@ -895,7 +895,7 @@ func (d *indirectIndex) OverlapsTimeRange(min, max int64) bool {
 
 // OverlapsKeyRange returns true if the min and max keys of the file overlap the arguments min and max.
 func (d *indirectIndex) OverlapsKeyRange(min, max string) bool {
-	return d.minKey <= max && d.maxKey >= max
+	return d.minKey <= max && d.maxKey >= min
```
e-dard: This looks like it could have been the source of other bugs unrelated to the dropping/missing data issue. Do we have any other open issues that could be hitting this, and how do you think they would manifest themselves?

tsdb/engine/tsm1/engine.go
jwilder: `KeyCount` ends up taking `RLocks` on the reader and I didn't want to lock the readers with every loop iteration.

tsdb/engine/tsm1/engine.go
jwilder: There shouldn't be any errors here.  I can log them though.

tsdb/engine/tsm1/reader.go
jwilder: This is only called via `DeleteRange` and would appear that a delete didn't actually delete everything.

jwilder:
@e-dard All fixed.

PR #8595, opened 2017-07-14
#6563 Support Ctrl+C to cancel a running query in the Influx CLI

### LINKED ISSUE #6563: Support Ctrl+C to cancel a running query in the Influx CLI
The Influx CLI does not allow you to use Ctrl+C to cancel a running query. We should catch the signal and close the connection on the client side.

This may require a change to the Go client API.

emluque:
Fixes #6563

Solved the issue by using context on the http request on the client.
Some common functionality between the queries with context and the ones without it was extracted to avoid code duplication.
Wrote tests for everything.

- [ X] Rebased/mergable
- [ X] Tests pass
- [ ] CHANGELOG.md updated
- [ X] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

### DIFF
--- a/client/influxdb.go
+++ b/client/influxdb.go
@@ -180,6 +180,12 @@ func (c *Client) SetPrecision(precision string) {
 
 // Query sends a command to the server and returns the Response
 func (c *Client) Query(q Query) (*Response, error) {
+	return c.QueryContext(context.Background(), q)
+}
+
+// QueryContext sends a command to the server and returns the Response
+// It uses a context that can be cancelled by the command line client
+func (c *Client) QueryContext(ctx context.Context, q Query) (*Response, error) {
 	u := c.url
 
 	u.Path = "query"
@@ -205,6 +211,8 @@ func (c *Client) Query(q Query) (*Response, error) {
 	if c.username != "" {
 		req.SetBasicAuth(c.username, c.password)
 	}
+
+	req = req.WithContext(ctx)
 
 	resp, err := c.httpClient.Do(req)
 	if err != nil {
--- a/cmd/influx/cli/cli.go
+++ b/cmd/influx/cli/cli.go
@@ -3,6 +3,7 @@ package cli // import "github.com/influxdata/influxdb/cmd/influx/cli"
 
 import (
 	"bytes"
+	"context"
 	"encoding/csv"
 	"encoding/json"
 	"errors"
@@ -729,8 +730,31 @@ func (c *CommandLine) ExecuteQuery(query string) error {
 		}
 		query = pq.String()
 	}
-	response, err := c.Client.Query(c.query(query))
+
+	ctx := context.Background()
+	if !c.IgnoreSignals {
+		done := make(chan struct{})
+		defer close(done)
+
+		var cancel func()
+		ctx, cancel = context.WithCancel(ctx)
+		go func() {
+			select {
+			case <-done:
+			case <-c.osSignals:
+				cancel()
+			}
+		}()
+	}
+
+	response, err := c.Client.QueryContext(ctx, c.query(query))
 	if err != nil {
+		if err.Error() == "" {
+			err = ctx.Err()
+			if err == context.Canceled {
+				err = errors.New("aborted by user")
+			}
+		}
 		fmt.Printf("ERR: %s\n", err)
 		return err
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - [#8611](https://github.com/influxdata/influxdb/issues/8611): Respect X-Request-Id/Request-Id headers.
 - [#8572](https://github.com/influxdata/influxdb/issues/8668): InfluxDB now uses MIT licensed version of BurntSushi/toml.
 - [#8752](https://github.com/influxdata/influxdb/pull/8752): Use system cursors for measurement, series, and tag key meta queries.
+- [#6563](https://github.com/influxdata/influxdb/issues/6563): Support Ctrl+C to cancel a running query in the Influx CLI. Thanks @emluque!
 
 ### Bugfixes
 
--- a/client/influxdb_test.go
+++ b/client/influxdb_test.go
@@ -1,6 +1,7 @@
 package client_test
 
 import (
+	"context"
 	"crypto/tls"
 	"encoding/json"
 	"errors"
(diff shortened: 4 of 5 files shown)


emluque:
Hi @jsternberg , I've sent this more than a month ago, it solves an issue that was labelled "Help Wanted". I would love to get some feedback, If there is anything that needs to be changed or if the approach is incorrect I would happily correct this, Thanks in advance.

client/influxdb.go
```diff
@@ -180,6 +180,23 @@ func (c *Client) SetPrecision(precision string) {
 
 // Query sends a command to the server and returns the Response
 func (c *Client) Query(q Query) (*Response, error) {
+
```
jsternberg: Can we remove these extraneous spaces? And instead of adding a new function `createHTTPRequest` and duplicating code, can you move the implementation of this function to `QueryContext` and have this method be:

```
func (c *Client) Query(q Query) (*Response, error) {
    return c.QueryContext(q, context.Background())
}
```

It seems like it would require less code and fewer changes.

client/influxdb.go
```diff
@@ -249,6 +265,26 @@ func (c *Client) Query(q Query) (*Response, error) {
 		return &response, fmt.Errorf("received status code %d from server", resp.StatusCode)
 	}
 	return &response, nil
+
+}
+
+// QueryWithContext sends a command to the server and returns the Response
+// It uses a context that can be cancelled by the command line client
+func (c *Client) QueryWithContext(ctx context.Context, q Query) (*Response, error) {
```
jsternberg: Can you rename this to `QueryContext`?

client/influxdb_internal_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	bod := `{"results":[{"statement_id":0,"series":[{"name":"average_temperature","columns":["time","degrees","location"],"values":[[1439856000000000000,82,"coyote_creek"]]}]}]}`
+
+	status := 200
+
+	NewRespBodyFromString := func(body string) io.ReadCloser {
+		return &dummyReadCloser{strings.NewReader(body)}
+	}
+
+	resp :=
```
jsternberg: This looks like bad formatting. Did you run this through `go fmt`?

cmd/influx/cli/cli.go
```diff
@@ -729,9 +758,30 @@ func (c *CommandLine) ExecuteQuery(query string) error {
 		}
 		query = pq.String()
 	}
-	response, err := c.Client.Query(c.query(query))
+	return query, nil
+}
+
+// ExecuteQueryWithCancel runs any query statement. It is cancellable by the user.
+func (c *CommandLine) ExecuteQueryWithCancel(query string) error {
```
jsternberg: I think you can do this without necessarily having a separate control flow for `ExecuteQuery` and `ExecuteQueryCancel`. The `ExecuteQuery` function has access to the signals being received by the program already so you don't need two methods. All you really need is something like this.

```
ctx := context.Background()
if !c.IgnoreSignals {
    done := make(chan struct{})
    defer close(done)

    var cancel func()
    ctx, cancel = context.WithCancel(ctx)
    go func() {
        // wait for either a signal to be sent or the done channel to have been closed
        select {
        case <-done:
        case <-c.osSignals:
            cancel()
        }
    }()
}
// actually execute the query here with the context
```

cmd/influx/cli/cli.go
```diff
@@ -746,6 +796,30 @@ func (c *CommandLine) ExecuteQuery(query string) error {
...
+// or ctx.Done is closed while the query of f is running, it cancels the query
+// waits for f to exit, and returns an errot. Otherwise it returns f's error.
+func clientDoQuery(ctx context.Context, cancel context.CancelFunc, c *CommandLine, query string, f func(*client.Response, error) error) error {
+	// Run the HTTP request in a goroutine and pass the response to f.
+	ch := make(chan error, 1)
+	go func() { ch <- f(c.Client.QueryWithContext(ctx, c.query(query))) }()
+	select {
+	case <-c.osSignals:
+		cancel()
+		fmt.Println("Aborting query..")
```
jsternberg: I don't think we need to print that the query was aborted. The person knows what they did.

client/influxdb.go
emluque: Done.

client/influxdb.go
emluque: Done.

client/influxdb_internal_test.go
emluque: This tests are not necessary after previous changes.

cmd/influx/cli/cli.go
emluque: No, I can't do that, because of the way that contexts work. I need on ExecuteQueryCancel to call clientDoQuery to be able to a) catch cancels and b) get the response on the closure. This is the standard way to use contexts as described here: https://blog.golang.org/context 
Also, the main loop has access to the signals, ExecuteQuery does not, that's why I has to find a way to make them accessible and that's why I'm using context.

cmd/influx/cli/cli.go
emluque: Done. I took the message out, but on my opinion when testing this on the cli I found returning a "Cancelled" message to be clearer. That's my opinion as a User.

emluque:
No problem @jsternberg , I made all the changes except for one, because I don't think it can be done differently. See my comments about it.

jsternberg:
I added a commit to expand on what I said earlier. Can you take a look and see if you have any objections? I kept my changes to a separate commit so they were separately reviewable and they didn't interfere with your work.

If you're fine with the changes I made, please squash my changes into your commit and rebase this on top of master, give yourself credit in the changelog, and we can potentially get this merged by Thursday/Friday.

Thanks.

emluque:
I'm sorry I didn't completely understand what you meant on the comment before. Yes, this is way simpler and clearer. However the only problem I see is that we are showing an error message after context is cancelled. Specifically it says "ERR: context canceled" after Ctrl-C, which I think it's pretty cryptic for the user. Maybe we can take it out or change it to "Query Aborted". What do you think?

jsternberg:
Yea, a different error message is fine. Feel free to change it, squash the PR into one commit, and then we can get this merged.

emluque:
Done.

cmd/influx/cli/cli.go
```diff
@@ -729,9 +730,31 @@ func (c *CommandLine) ExecuteQuery(query string) error {
...
+				cancel()
+			}
+		}()
+	}
+
+	response, err := c.Client.QueryContext(ctx, c.query(query))
 	if err != nil {
-		fmt.Printf("ERR: %s\n", err)
+		if err.Error() == "" {
+			fmt.Println("Aborting Query..")
```
jsternberg: This shouldn't print aborting query. You should do:

```
if err := ctx.Err(); err == context.Canceled {
    return errors.New("aborted by user")
} else {
    return err
}
```

Then the normal error printing loop will print the correct message.

jsternberg:
You also may have to rebase this on top of master because of the changelog update I made.

emluque:
Done and thank you for the message!

emluque:
Sorry it's not done yet. I forgot to change the message. I changed it and now it seems one test is failing. I'll look at this tomorrow, I can't continue today.

jsternberg:
I updated the commit slightly to have a better flow when printing "aborted by user". I'm approving the PR and then I'm going to see if I can get all of the CI checks to run correctly. They should, but our CI system is a bit flaky at the moment.

PR #8603, opened 2017-07-17
Fix the cq start and end times to use unix timestamps

jsternberg:
The Go timestamp leads Truncate to start a week on Monday, but the query
engine uses unix timestamps which has the week start on a Thursday.

Updating the service so it uses a custom truncate method that uses the
unix timestamp instead of `time.Time`.

Backport of #8600.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## v1.2.5 [unreleased]
+
+### Bugfixes
+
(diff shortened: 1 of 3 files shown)


jsternberg:
Yes. It's a backport of #8600.

PR #8606, opened 2017-07-19
Accomodate for time zones with continuous query scheduling

jsternberg:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/services/continuous_querier/service.go
+++ b/services/continuous_querier/service.go
@@ -287,6 +287,12 @@ func (s *Service) ExecuteContinuousQuery(dbi *meta.DatabaseInfo, cqi *meta.Conti
 		return false, err
 	}
 
+	// Set the time zone on the now time if the CQ has one. Otherwise, force UTC.
+	now = now.UTC()
+	if cq.q.Location != nil {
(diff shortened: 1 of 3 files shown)


jsternberg:
I'm going to hold off on merging this one just yet. There are some conditions that I'm having difficulty with that are located in #8607.

desa:
@jsternberg Makes sense.

services/continuous_querier/service.go
```diff
@@ -287,6 +287,13 @@ func (s *Service) ExecuteContinuousQuery(dbi *meta.DatabaseInfo, cqi *meta.Conti
 		return false, err
 	}
 
+	// Set the time zone on the now time if the CQ has one. Otherwise, force UTC.
+	if cq.q.Location != nil {
```
desa: This is just a stylistic thing, but it can be changed to 

```go
now = now.UTC()
if cq.q.Location != nil {
    now = now.In(cq.q.Location)
}

jsternberg:
@desa can you review this again? I think I fixed the time zone boundary sections and incorporated your feedback.

PR #8615, opened 2017-07-21
uint64 support for storage layer

stuartcarnie:
### TL;DR

This PR is strictly to enhance the storage engine in order to store unsigned 64-bit integers. As such, there are no user-facing features including the ability to write and query `uint64` values.

- [x] Rebased/mergeable
- [x] Tests pass

---

### Purpose

The goal of this PR is to add `uint64` support to the storage engine. Consequently, minor changes where introduced to the `influxql` and `models` packages, addding constants or approapriate APIs to existing interfaces in order to facilitate completion of the storage changes. As such, complete support for `uint64` values within the `influxql` package is **incomplete**.
 

### Naming

Following the established convention of using `Float` and `Integer` to name APIs related to `float64` and `int64` values, `Unsigned` was chosen to to identify `uint64` specific APIs.

### tsdb / storage engine

Complimentary unit tests were added for Unsigned values, sourced from their Integer counterparts.

#### TSM

Unsigned values are represented by a new block type, `BlockUnsigned = 4`. These blocks utilize the same encoding as `int64` values (RLE, simple8b or uncompressed) and therefore reuse the existing `IntegerEncoder` / `IntegerDecoder` types.

#### wal.go

To encode `uint64` values, the wal was extended with a new `unsignedEntryType = 5` and appropriate codecs added, following the established `integerEntryType`.

#### reader.go

The `readUnsignedBlock` was added to the `blockAccessor` interface and an appropriate implementation added to the `mmapAccessor`.

#### code generation

`Unsigned` was added to the appropriate code generation template data files. Templates were modified, where appropriate, to facilitate generation of this new type. New cursor types were added to cast between float64, int64 and uint64 streams.


tsdb/engine/tsm1/compact.gen.go
```diff
@@ -439,6 +439,220 @@ func (k *tsmKeyIterator) chunkInteger(dst blocks) blocks {
...
+	}
+
+	dedup := len(k.mergedUnsignedValues) != 0
+	if len(k.blocks) > 0 && !dedup {
+		// If we have more than one block or any partially tombstoned blocks, we many need to dedup
+		dedup = len(k.blocks[0].tombstones) > 0 || k.blocks[0].partiallyRead()
+
+		// Quickly scan each block to see if any overlap with the prior block, if they overlap then
+		// we need to dedup as there may be duplicate points now
+		for i := 1; !dedup && i < len(k.blocks); i++ {
```
e-dard: I think this loop could be simplified to:

```go
for i := 1; !dedup && i < len(k.blocks); i++ {
    dedup = k.blocks[i].partiallyRead() ||
	    k.blocks[i].minTime <= k.blocks[i-1].maxTime || 
            len(k.blocks[i].tombstones) > 0
}
```

What do you think?

stuartcarnie: Good call – the resulting ASM is simpler too. I'll update the code-gen templates.

e-dard:
The changes look good.

Issue #8611, opened 2017-07-19
Feature Request: Allow the client to specify the X-Request-Id header for correlating requests across services


timhallinflux:
### Feature Request

__Proposal:__ 
InfluxDB logs requests with a correlation Id and returns the correlation Id to the client in the Request-Id HTTP header for correlating requests and responses.

For example, an HTTP GET on http://localhost:8086/ping results in a log message like the following:

[httpd] 172.17.0.1 - - [19/Jul/2017:18:38:08 +0000] "GET /ping HTTP/1.1" 204 0 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/59.0.3071.115 Safari/537.36" 68beb75a-6cb1-11e7-8001-000000000000 1811

The correlation Id 68beb75a-6cb1-11e7-8001-000000000000 is returned in the Request-Id header.

To track requests across disparate services, it would be ideal if InfluxDB honored the X-Request-Id header so that the caller could pass a correlation Id as part of the request. If the header exists, use this correlation Id for logging and the Request-Id header in the response. If it does not exist, generate the correlation Id for the request as it is done now.

__Current behavior:__ 
X-Request-Id cannot be specified by the caller.

__Desired behavior:__ [What you would like to happen]
Allow X-Request-Id to be sent by the caller and return it in the Request-Id header in the response.

__Use case:__ 
Required for correlating service-to-service calls when building applications which leverage InfluxDB

PR #8619, opened 2017-07-24
Support pre-existing request ID headers

e-dard:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

Implements #8611.

Other applications or services sometimes expose a header containing a
unique ID, which can then be included in logging or response information
to allow an operator to link inter-service requests. The most common
header name used by services in the wild appears to be `X-Request-Id`,
but `Request-Id` is also used.

This commit adds support for specifying either `X-Request-Id` or
`Request-Id` headers, which will then be used by InfluxDB when logging
request information, and also in the `X-Request-Id` and `Request-Id`
response headers.

In the HTTP response, we populate both `X-Request-Id` and `Request-Id` to maintain backwards
compatibility with previous version, and to support the more common
`X-Request-Id` header name.

If both `X-Request-Id` and `Request-Id` are specified, then
`X-Request-Id` is used.

If neither header is specified, then in line with previous behaviour, we
generate a v1 UUID.

### DIFF
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -1170,9 +1170,30 @@ func cors(inner http.Handler) http.Handler {
 
 func requestID(inner http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		uid := uuid.TimeUUID()
-		r.Header.Set("Request-Id", uid.String())
-		w.Header().Set("Request-Id", r.Header.Get("Request-Id"))
+		// X-Request-Id takes priority.
+		rid := r.Header.Get("X-Request-Id")
+
+		// If X-Request-Id is empty, then check Request-Id
+		if rid == "" {
+			rid = r.Header.Get("Request-Id")
+		}
+
+		// If Request-Id is empty then generate a v1 UUID.
+		if rid == "" {
+			rid = uuid.TimeUUID().String()
+		}
+
+		// We read Request-Id in other handler code so we'll use that naming
+		// convention from this point in the request cycle.
+		r.Header.Set("Request-Id", rid)
+
+		// Set the request ID on the response headers.
+		// X-Request-Id is the most common name for a request ID header.
+		w.Header().Set("X-Request-Id", rid)
+
+		// We will also set Request-Id for backwards compatibility with previous
+		// versions of InfluxDB.
+		w.Header().Set("Request-Id", rid)
 
 		inner.ServeHTTP(w, r)
 	})
--- a/services/httpd/handler_test.go
+++ b/services/httpd/handler_test.go
@@ -698,6 +698,72 @@ func TestHandler_XForwardedFor(t *testing.T) {
 	}
 }
 
+func TestHandler_XRequestId(t *testing.T) {
+	var buf bytes.Buffer
+	h := NewHandler(false)
+	h.CLFLogger = log.New(&buf, "", 0)
+
+	cases := []map[string]string{
+		{"X-Request-Id": "abc123", "Request-Id": ""},          // X-Request-Id is used.
+		{"X-REQUEST-ID": "cde", "Request-Id": ""},             // X-REQUEST-ID is used.
+		{"X-Request-Id": "", "Request-Id": "foobarzoo"},       // Request-Id is used.
+		{"X-Request-Id": "abc123", "Request-Id": "foobarzoo"}, // X-Request-Id takes precedence.
+		{"X-Request-Id": "", "Request-Id": ""},                // v1 UUID generated.
+	}
+
+	for _, c := range cases {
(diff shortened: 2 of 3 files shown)


services/httpd/handler.go
```diff
@@ -1165,9 +1165,30 @@ func cors(inner http.Handler) http.Handler {
 
 func requestID(inner http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		uid := uuid.TimeUUID()
-		r.Header.Set("Request-Id", uid.String())
-		w.Header().Set("Request-Id", r.Header.Get("Request-Id"))
+		// X-Request-ID takes priority.
+		var rid = r.Header.Get("X-Request-ID")
```
mark-rushakoff: I learned through previous benchmarking that we can avoid some garbage creation by following canonicalization rules in our calls to Header.Get. This should be `-Id`, not `-ID`.

services/httpd/handler_test.go
```diff
@@ -698,6 +698,60 @@ func TestHandler_XForwardedFor(t *testing.T) {
...
+	}{
+		{XRequestID: "abc123", RequestId: ""},          // X-Request-ID is used.
+		{XRequestID: "", RequestId: "foobarzoo"},       // Request-Id is used.
+		{XRequestID: "abc123", RequestId: "foobarzoo"}, // X-Request-ID takes precedence.
+		{XRequestID: "", RequestId: ""},                // v1 UUID generated.
+	}
+
+	for _, c := range cases {
+		t.Run(fmt.Sprintf("X-Request-ID: %q Request-Id: %q", c.XRequestID, c.RequestId), func(t *testing.T) {
+			buf.Reset() // Reset the buffer.
```
mark-rushakoff: download
I'm used to seeing higher quality comments from you.

services/httpd/handler_test.go
```diff
@@ -698,6 +698,60 @@ func TestHandler_XForwardedFor(t *testing.T) {
...
+			req := MustNewRequest("GET", "/ping", nil)
+			req.RemoteAddr = "127.0.0.1"
+
+			// Set the relevant request ID headers
+			req.Header.Set("X-Request-ID", c.XRequestID)
```
mark-rushakoff: Should also be `-Id` for consistency.

services/httpd/handler_test.go
```diff
@@ -698,6 +698,60 @@ func TestHandler_XForwardedFor(t *testing.T) {
...
+			req.Header.Set("Request-Id", c.RequestId)
+
+			w := httptest.NewRecorder()
+			h.ServeHTTP(w, req)
+
+			parts := strings.Split(buf.String(), " ")
```
mark-rushakoff: `parts[12]` throughout the rest of this function is a bit mysterious. Is there a better way to extract the returned ID? Will the returned ID always be at index 12, even if we add or remove headers later?

services/httpd/handler_test.go
```diff
@@ -698,6 +698,60 @@ func TestHandler_XForwardedFor(t *testing.T) {
...
+					t.Fatalf("got ID of %q, expected %q", got, exp)
+				}
+			} else {
+				if got, exp := parts[12], c.RequestId; got != exp {
+					t.Fatalf("got ID of %q, expected %q", got, exp)
+				}
+			}
+
+			// Check response headers
+			if got, exp := w.Header().Get("Request-ID"), parts[12]; got != exp {
```
mark-rushakoff: `-Id` in this block as well.

services/httpd/handler_test.go
e-dard: I must have been drunk.

services/httpd/handler_test.go
e-dard: There probably is but it would be convoluted I expect, as we would have to parse a log entry. I could add a comment explaining that this will need to be adjusted if we change what we output in the HTTP logs

e-dard:
@mark-rushakoff updated.

services/httpd/handler.go
```diff
@@ -1170,9 +1170,30 @@ func cors(inner http.Handler) http.Handler {
 
 func requestID(inner http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		uid := uuid.TimeUUID()
-		r.Header.Set("Request-Id", uid.String())
-		w.Header().Set("Request-Id", r.Header.Get("Request-Id"))
+		// X-Request-Id takes priority.
+		var rid = r.Header.Get("X-Request-Id")
```
jsternberg: Should this be:

```
rid := r.Header.Get("X-Request-Id")
```

?

e-dard: Good catch. I'll fix that when I rebase.

PR #8631, opened 2017-07-26
Test that bound params generate single statement

rbetts:
Scratches an itch - I was curious about sql injection and a few days ago and noticed @mark-rushakoff answered a community question pointing to bound parameters. I wrote a test case to prove to myself that bound parameters don't parse to two statements.  

Thoughts? Reasonable test case?

### DIFF
--- a/influxql/parser_test.go
+++ b/influxql/parser_test.go
@@ -81,18 +81,27 @@ func TestParser_ParseQuery_NoSemicolon(t *testing.T) {
 	}
 }
 
-func TestParser_ParseBoundParameter(t *testing.T) {
-	parser := influxql.NewParser(strings.NewReader(`SELECT * FROM m WHERE t = $v`))
-	params := map[string]interface{}{
-		"v": "1; select * from m where t = v",
+func TestParser_SQLInjection(t *testing.T) {
+	injections := []string{
+		"1; select * from m where t = v",
+		`1 "; drop measurement m;"`,
+		`1 '; drop measurement m;'`,
 	}
-	parser.SetParams(params)
-	q, err := parser.ParseQuery()
-	if err != nil {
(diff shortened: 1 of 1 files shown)


mark-rushakoff:
And I think if you rebase on master you won't get the `More CircleCI nodes allocated than tests environments to run!` error from Circle.

jsternberg:
I think there's another test of this, but it isn't so explicit and I don't think it tests SQL injection specifically. Bound parameters are performed at the AST level so they barely interact with the parser. It's not substituting strings into the query and then parsing, but parsing and then doing substitutions.

I think this change though is good to make that explicit so nobody tries to implement it in the other way and makes it vulnerable to SQL injection. After @mark-rushakoff's comments are addressed, I think this can be merged.

influxql/parser_test.go
```diff
@@ -81,6 +81,21 @@ func TestParser_ParseQuery_NoSemicolon(t *testing.T) {
 	}
 }
 
+func TestParser_ParseBoundParameter(t *testing.T) {
```
jsternberg: If you can change this test name to `SQLInjection`, I would like that more. We already test bound parameters in another section and this is a test for sql injection.

rbetts:
@mark-rushakoff is this what you were thinking?

```
Parameterized result: SELECT * FROM m WHERE t = '1; select * from m where t = v'
Parameterized result: SELECT * FROM m WHERE t = '1 "; drop measurement m;"'
Parameterized result: SELECT * FROM m WHERE t = '1 \'; drop measurement m;\''
```

mark-rushakoff:
Yes, those cases cover some very common injection attempts.

mark-rushakoff:
Still needs rebased on master to get Circle tests passing, however.

PR #8641, opened 2017-07-27
Update udp port in README to be in valid range

jheusser:
`80891` is not a valid port and outside of range, changed to `8189`.

### DIFF
--- a/services/udp/README.md
+++ b/services/udp/README.md
@@ -116,7 +116,7 @@ Multiple UDP listeners
 [[udp]]
   # High-traffic UDP
   enabled = true
-  bind-address = ":80891" # the bind address
+  bind-address = ":8189" # the bind address
   database = "mymetrics" # Name of the database that will be written to
   batch-size = 5000 # will flush if this many points get buffered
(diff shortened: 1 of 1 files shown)


services/udp/README.md
```diff
@@ -116,7 +116,7 @@ Multiple UDP listeners
 [[udp]]
   # High-traffic UDP
   enabled = true
-  bind-address = ":80891" # the bind address
+  bind-address = ":8189" # the bind address
```
jsternberg: It looks like this is supposed to be `8089` according to the code and the `1` was a typo. Can you change it to that number?

jheusser: This is on multiple udp sections, so `8089` is taken in the section above the `High-traffic UDP` config.

jsternberg: Ah, my mistake. Diffs only show so much.

jsternberg:
I forgot to ask before pushing merge, but just to confirm, have you signed the CLA and could you if you haven't?

https://www.influxdata.com/legal/cla/

Thanks.

jheusser:
@jsternberg done.

PR #8642, opened 2017-07-27
Parse time literals using the time zone in the select statement

### LINKED ISSUE #8639: Timezone bug
Greetings,

and thank you all for developing such an amazing product as Influxdb!

I want to report a bug that I ve noticed with the new 1.3.1 version of InfluxDB regarding the new and useful Timezone feature you added. The bug is that when I query my database to get me some sensor readings that happened within a time range, for example at the date: 26-07-2017 and I add the new tz('Europe/Athens') function at the end, which is +03 hours from UTC, it does not display the readings that happened at 25-07-2017 between 21:00 - 24:00 and also displays some that normally do not belong there, which are the ones from 21:00 - 24:00 of 26-07-2017 that belong to 27-07-2017 in my Timezone.

The reason for this I believe is that the query first fetches the UTC values of that specific date and then applies the Timezone function tz(), where the correct thing to do is to first apply the tz() function and then see which readings belong to the requested time range.

Below I show you an example to be more clear on this:

```
> select * from events where dev_id='yyyyyyyyyyyy' and time > '2017-07-25' and time < '2017-07-26' tz('Europe/Athens')
name: events
time                      alert country dev_id       host         location
----                      ----- ------- ------       ----         ------------
2017-07-25T05:50:50+03:00 true  FRA     yyyyyyyyyyyy xxxxxxxxxxxx Test_Home1
2017-07-25T08:41:40+03:00 true  FRA     yyyyyyyyyyyy xxxxxxxxxxxx Test_Home1
2017-07-25T11:32:50+03:00 true  FRA     yyyyyyyyyyyy xxxxxxxxxxxx Test_Home1
2017-07-25T14:24:00+03:00 true  FRA     yyyyyyyyyyyy xxxxxxxxxxxx Test_Home1
2017-07-25T17:15:10+03:00 true  FRA     yyyyyyyyyyyy xxxxxxxxxxxx Test_Home1
2017-07-25T20:05:50+03:00 true  FRA     yyyyyyyyyyyy xxxxxxxxxxxx Test_Home1
2017-07-25T22:57:20+03:00 true  FRA     yyyyyyyyyyyy xxxxxxxxxxxx Test_Home1
2017-07-26T01:48:30+03:00 true  FRA     yyyyyyyyyyyy xxxxxxxxxxxx Test_Home1


> select * from events where dev_id='yyyyyyyyyyyy' and time > '2017-07-26' and time < '2017-07-27' tz('Europe/Athens')
name: events
time                      alert country dev_id       host         location
----                      ----- ------- ------       ----         ------------
2017-07-26T06:21:20+03:00 true  FRA     yyyyyyyyyyyy xxxxxxxxxxxx Test_Home1
2017-07-26T09:42:30+03:00 true  FRA     yyyyyyyyyyyy xxxxxxxxxxxx Test_Home1
2017-07-26T13:04:30+03:00 true  FRA     yyyyyyyyyyyy xxxxxxxxxxxx Test_Home1
2017-07-26T16:26:00+03:00 true  FRA     yyyyyyyyyyyy xxxxxxxxxxxx Test_Home1
2017-07-26T19:48:00+03:00 true  FRA     yyyyyyyyyyyy xxxxxxxxxxxx Test_Home1
2017-07-26T23:09:40+03:00 true  FRA     yyyyyyyyyyyy xxxxxxxxxxxx Test_Home1
```

As you can see the last reading from the 1st query is wrongly placed there and is 
missing from the 2nd query that I did for the next day.

Hope this helps

jsternberg:
Fixes #8639.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): issue filed or pull request submitted \<[influxdata/docs.influxdata.com#1226](https://github.com/influxdata/docs.influxdata.com/issues/1226)\>

### DIFF
--- a/influxql/ast.go
+++ b/influxql/ast.go
@@ -1482,7 +1482,7 @@ func (s *SelectStatement) RewriteTimeCondition(now time.Time) error {
 	if err != nil {
 		return err
 	} else if interval > 0 && s.Condition != nil {
-		_, tmax, err := TimeRange(s.Condition)
+		_, tmax, err := TimeRange(s.Condition, s.Location)
 		if err != nil {
 			return err
 		}
@@ -3778,19 +3778,23 @@ func (l *StringLiteral) IsTimeLiteral() bool {
 }
 
 // ToTimeLiteral returns a time literal if this string can be converted to a time literal.
-func (l *StringLiteral) ToTimeLiteral() (*TimeLiteral, error) {
+func (l *StringLiteral) ToTimeLiteral(loc *time.Location) (*TimeLiteral, error) {
+	if loc == nil {
+		loc = time.UTC
+	}
+
 	if isDateTimeString(l.Val) {
-		t, err := time.Parse(DateTimeFormat, l.Val)
+		t, err := time.ParseInLocation(DateTimeFormat, l.Val, loc)
 		if err != nil {
 			// try to parse it as an RFCNano time
-			t, err = time.Parse(time.RFC3339Nano, l.Val)
+			t, err = time.ParseInLocation(time.RFC3339Nano, l.Val, loc)
 			if err != nil {
 				return nil, ErrInvalidTime
 			}
 		}
 		return &TimeLiteral{Val: t}, nil
 	} else if isDateString(l.Val) {
-		t, err := time.Parse(DateFormat, l.Val)
+		t, err := time.ParseInLocation(DateFormat, l.Val, loc)
 		if err != nil {
 			return nil, ErrInvalidTime
 		}
@@ -4040,7 +4044,7 @@ func OnlyTimeExpr(expr Expr) bool {
 
 // TimeRange returns the minimum and maximum times specified by an expression.
 // It returns zero times if there is no bound.
-func TimeRange(expr Expr) (min, max time.Time, err error) {
+func TimeRange(expr Expr, loc *time.Location) (min, max time.Time, err error) {
 	WalkFunc(expr, func(n Node) {
 		if err != nil {
 			return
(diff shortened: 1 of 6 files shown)


influxql/ast.go
```diff
@@ -4868,7 +4877,7 @@ func reduceBinaryExprBooleanLHS(op Token, lhs *BooleanLiteral, rhs Expr) Expr {
 	return &BinaryExpr{Op: op, LHS: lhs, RHS: rhs}
 }
 
-func reduceBinaryExprDurationLHS(op Token, lhs *DurationLiteral, rhs Expr) Expr {
+func reduceBinaryExprDurationLHS(op Token, lhs *DurationLiteral, rhs Expr, loc *time.Location) Expr {
```
desa: Kind of sucks that this bleeds into the some of the binary expression reducing. Guess it has to though.

influxql/ast.go
```diff
@@ -5300,9 +5309,16 @@ type Valuer interface {
 	Value(key string) (interface{}, bool)
 }
 
+// ZoneValuer is the interface that specifies the current time zone.
+type ZoneValuer interface {
```
desa: Do we really need this interface? Seems like the only place this gets used is

```
+	loc := time.UTC
 +	if v, ok := valuer.(ZoneValuer); ok {
 +		loc = v.Zone()
 +	}
```
Is there any anticipation that there will be anything else that implements `ZoneValuer`?

I guess I've never quite understood why people use create an interface (it seems pretty common) for something where a concrete type would suffice.

jsternberg: It's because this is an extension of the `Valuer` interface and not all `Valuer`s can have a Zone method.

influxql/ast.go
jsternberg: Yea, I didn't like this. Might be improved in the future with some better parsing of the expression.

influxql/ast.go
desa: But you could just as easily casted to a `NowValuer`

PR #8652, opened 2017-07-30
Reduce allocations using nil cursors and literal value cursors

stuartcarnie:
* Replaces individual literal cursors with a single interface value. This has the benefit that the value is immediately wrapped in an `iface`, removing `convT2I` calls and subsequent allocations every time the value is read.
* Statically allocates the nil value cursors, saving further allocations as they are stateless and safe for concurrent use.

```
benchmark                           old ns/op     new ns/op     delta
BenchmarkIntegerIterator_Next-8     82.8          22.7          -72.58%

benchmark                           old allocs     new allocs     delta
BenchmarkIntegerIterator_Next-8     3              0              -100.00%

benchmark                           old bytes     new bytes     delta
BenchmarkIntegerIterator_Next-8     32            0             -100.00%
```

- [x] Rebased/mergable
- [x] Tests pass

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -1852,19 +1852,19 @@ func (e *Engine) createVarRefSeriesIterator(ref *influxql.VarRef, name string, s
 				// If a field was requested, use a nil cursor of the requested type.
 				switch ref.Type {
 				case influxql.Float, influxql.AnyField:
-					aux[i] = &floatNilLiteralCursor{}
+					aux[i] = nilFloatLiteralValueCursor
 					continue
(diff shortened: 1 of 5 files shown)


jwilder:
@jsternberg You are more familiar with this code than I am. LGTM though.

stuartcarnie:
@jsternberg / @jwilder what should we put in the CHANGELOG for this one?

jsternberg:
I'm not really sure. Likely goes in the features section and references this PR. Just a one sentence blurb about optimizing the memory usage for some cursors works. Enough for a person who is perusing the changelog to click the link for more information if they desire.

PR #8669, opened 2017-08-07
TSI Index Migration Tool

benbjohnson:
## Overview

This adds an `inmem2tsi` command to `influx_inspect` that allows for the conversion of an in-memory index to a `TSI` index. The index is written to a temporary location until complete and then moved to its permanent location. If any error occurs then it will fallback to the in-memory index.

This is not intended as a hot shard conversion. New shards can be set to use TSI as their new index and old shards can be converted as needed. A server restart is required after a shard is converted.

## Example

```sh
$ influx_inspect inmem2tsi -path ~/.influxdb/data/stress/autogen/1 ~/.influxdb/wal/stress/autogen/1
```

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

- [x] Provide example syntax
- [x] Update man page when modifying a command

### DIFF
--- a/cmd/influx_inspect/inmem2tsi/inmem2tsi.go
+++ b/cmd/influx_inspect/inmem2tsi/inmem2tsi.go
@@ -1,0 +1,164 @@
+// Package inmem2tsi reads an in-memory index and exports it as a TSI index.
+package inmem2tsi
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/influxdata/influxdb/models"
+	"github.com/influxdata/influxdb/tsdb"
+	"github.com/influxdata/influxdb/tsdb/index/inmem"
+	"github.com/influxdata/influxdb/tsdb/index/tsi1"
+	"github.com/uber-go/zap"
+)
+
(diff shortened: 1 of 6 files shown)


cmd/influx_inspect/inmem2tsi/inmem2tsi.go
```diff
@@ -0,0 +1,160 @@
...
+	cmd.Logger.Info("opening tsi index in temporary location", zap.String("path", tmpPath))
+	if err := tsiIndex.Open(); err != nil {
+		return err
+	}
+	defer tsiIndex.Close()
+
+	cmd.Logger.Info("iterating over measurements")
+
+	// Iterate over each series & insert into new index.
+	for _, name := range inmemIndex.MeasurementNames() {
```
jwilder: I think there is a `ForEachMeasurementName` func that exist that does the same as this new method?  Any reason to not re-use that one?

dazoot:
Can the conversion be ran while influxdb is running ?

benbjohnson:
No, this tool is for offline conversion. The idea is to switch new shards to TSI and then do old shard conversion offline if necessary.

PR #8685, opened 2017-08-10
Run relevant unit tests on both indexes

e-dard:
- [x] Rebased/mergable
- [ ] Tests pass

This PR ensure that where appropriate, unit tests are run on both indexes using sub-tests. Benchmarks have also been updated in a similar way.

In terms of the `tests` package tests, I found that the test output was too noisy when using Go's verbose `-v` mode, which itself is useful for seeing the sub-test result output. Therefore I've added a new `-vv` flag for that package, if the server logs are desirable output. /cc @mark-rushakoff @aanthony1243.

Finally, regarding the `tests` package, all queries for each test will now run as sub-tests.

```
go test -v github.com/influxdata/influxdb/tests -vv
```

### DIFF
--- a/tests/server_test.go
+++ b/tests/server_test.go
@@ -40,14 +40,18 @@ func TestMain(m *testing.M) {
 		benchServer = OpenDefaultServer(c)
 
 		// Run test suite.
-		fmt.Printf("============= Running all tests for %q index =============\n", indexType)
+		if testing.Verbose() {
+			fmt.Printf("============= Running all tests for %q index =============\n", indexType)
+		}
 		if thisr := m.Run(); r == 0 {
 			r = thisr // We'll always remember the first time r is non-zero
 		}
 
 		// Cleanup
 		benchServer.Close()
-		fmt.Println()
+		if testing.Verbose() {
+			fmt.Println()
+		}
 	}
 	os.Exit(r)
 }
@@ -8413,7 +8417,7 @@ func TestServer_Query_LargeTimestamp(t *testing.T) {
 	// Open a new server with the same configuration file.
 	// This is to ensure the meta data was marshaled correctly.
 	s2 := OpenServer((s.(*LocalServer)).Config)
-	defer s2.Close()
+	defer s2.(*LocalServer).Server.Close()
 
 	for _, query := range test.queries {
 		t.Run(query.name, func(t *testing.T) {
--- a/tests/server_test.go
+++ b/tests/server_test.go
@@ -66,15 +66,16 @@ func TestServer_DatabaseCommands(t *testing.T) {
 	test := tests.load(t, "database_commands")
 
 	for _, query := range test.queries {
-		if query.skip {
-			t.Logf("SKIP:: %s", query.name)
-			continue
-		}
-		if err := query.Execute(s); err != nil {
-			t.Error(query.Error(err))
-		} else if !query.success() {
-			t.Error(query.failureMessage())
-		}
+		t.Run(query.name, func(t *testing.T) {
+			if query.skip {
+				t.Skipf("SKIP:: %s", query.name)
+			}
+			if err := query.Execute(s); err != nil {
+				t.Error(query.Error(err))
+			} else if !query.success() {
+				t.Error(query.failureMessage())
+			}
+		})
 	}
 }
 
@@ -90,20 +91,21 @@ func TestServer_Query_DropAndRecreateDatabase(t *testing.T) {
 	}
 
 	for i, query := range test.queries {
-		if i == 0 {
-			if err := test.init(s); err != nil {
-				t.Fatalf("test init failed: %s", err)
+		t.Run(query.name, func(t *testing.T) {
+			if i == 0 {
+				if err := test.init(s); err != nil {
+					t.Fatalf("test init failed: %s", err)
+				}
 			}
-		}
-		if query.skip {
-			t.Logf("SKIP:: %s", query.name)
-			continue
-		}
-		if err := query.Execute(s); err != nil {
-			t.Error(query.Error(err))
-		} else if !query.success() {
-			t.Error(query.failureMessage())
-		}
+			if query.skip {
+				t.Skipf("SKIP:: %s", query.name)
+			}
+			if err := query.Execute(s); err != nil {
+				t.Error(query.Error(err))
+			} else if !query.success() {
+				t.Error(query.failureMessage())
+			}
+		})
 	}
 }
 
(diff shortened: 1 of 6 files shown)


jwilder:
```
=== RUN   TestStore_Cardinality_Compactions/inmem
[I] 2017-08-10T15:21:33Z Using data dir: /tmp/influxdb-tsdb-960455470 service=store
panic: test executed panic(nil) or runtime.Goexit

goroutine 87907 [running]:
testing.tRunner.func1(0xc42056f110)
	/usr/local/go/src/testing/testing.go:622 +0x55f
runtime.Goexit()
	/usr/local/go/src/runtime/panic.go:366 +0x15c
testing.(*common).SkipNow(0xc42005f860)
	/usr/local/go/src/testing/testing.go:555 +0x59
testing.(*common).Skip(0xc42005f860, 0xc420479e60, 0x1, 0x1)
	/usr/local/go/src/testing/testing.go:536 +0x7d
github.com/influxdata/influxdb/tsdb_test.testStoreCardinalityCompactions(0xc42005f860, 0xc4201300d0)
	/root/go/src/github.com/influxdata/influxdb/tsdb/store_test.go:766 +0x142
github.com/influxdata/influxdb/tsdb_test.TestStore_Cardinality_Compactions.func1(0xc85fad, 0x5)
	/root/go/src/github.com/influxdata/influxdb/tsdb/store_test.go:823 +0x155
github.com/influxdata/influxdb/tsdb_test.TestStore_Cardinality_Compactions.func2(0xc42056f110)
	/root/go/src/github.com/influxdata/influxdb/tsdb/store_test.go:827 +0x72
testing.tRunner(0xc42056f110, 0xc4205b1c00)
	/usr/local/go/src/testing/testing.go:657 +0x108
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:697 +0x544
FAIL	github.com/influxdata/influxdb/tsdb	78.467s
```

tests/server_test.go
```diff
@@ -90,20 +103,21 @@ func TestServer_Query_DropAndRecreateDatabase(t *testing.T) {
...
-			continue
-		}
-		if err := query.Execute(s); err != nil {
-			t.Error(query.Error(err))
-		} else if !query.success() {
-			t.Error(query.failureMessage())
-		}
+			if query.skip {
+				t.Skipf("SKIP:: %s", query.name)
+			}
```
benbjohnson: This doesn't have to be done in this PR but we can take out all the `.skip` fields now that we're using `t.Run()`.

e-dard:
@jwilder that must be a flaky test

tsdb/engine/tsm1/engine_test.go
```diff
@@ -578,123 +578,100 @@ func TestEngine_CreateIterator_Condition(t *testing.T) {
...
-	// Create a few points.
-	p1 := MustParsePointString("cpu,host=A value=1.1 1000000000")
-	p2 := MustParsePointString("cpu,host=B value=1.2 2000000000")
-	p3 := MustParsePointString("cpu,host=A sum=1.3 3000000000")
+	for _, index := range tsdb.RegisteredIndexes() {
+		t.Run(index, func(t *testing.T) {
+			// Create a few points.
+			p1 := MustParsePointString("cpu,host=A value=1.1 1000000000")
+			p2 := MustParsePointString("cpu,host=B value=1.2 2000000000")
+			p3 := MustParsePointString("cpu,host=A sum=1.3 3000000000")
```
benbjohnson: Can we use a helper function (e.g. `RunForEachIndex(func(*testing.T))`) that loops the registered indexes?

tsdb/engine/tsm1/engine_test.go
```diff
@@ -767,120 +744,77 @@ func BenchmarkEngine_CreateIterator_Limit_1M(b *testing.B) {
...
+				pp = append(pp, p)
+			}
 
-	b.ResetTimer()
-	b.ReportAllocs()
-	for i := 0; i < b.N; i++ {
-		err := e.WritePoints(pp)
-		if err != nil {
-			b.Fatal(err)
+			b.Run(fmt.Sprintf("%s_%d", index, sz), func(b *testing.B) {
```
benbjohnson: Can this be two nested `b.Run()` calls instead of concatenating `index` & `sz`?

tsdb/engine/tsm1/engine_test.go
```diff
@@ -767,120 +744,77 @@ func BenchmarkEngine_CreateIterator_Limit_1M(b *testing.B) {
...
-		for i := 0; i < cpus; i++ {
-			wg.Add(1)
-			go func(i int) {
-				defer wg.Done()
-				from, to := i*batchSize, (i+1)*batchSize
-				err := e.WritePoints(pp[from:to])
-				if err != nil {
-					errC <- err
-					return
+			b.Run(fmt.Sprintf("%s_%d", index, sz), func(b *testing.B) {
```
benbjohnson: Nest `b.Run()` here as well.

tsdb/shard_test.go
```diff
@@ -600,799 +600,874 @@ func TestShard_Close_RemoveIndex(t *testing.T) {
 
 // Ensure a shard can create iterators for its underlying data.
 func TestShard_CreateIterator_Ascending(t *testing.T) {
-	sh := NewShard()
+	var (
+		sh  *Shard
+		itr influxql.Iterator
+	)
```
benbjohnson: I personally loathe the block `var` declaration. :)

Why use four lines for what can be declared in two?

tsdb/shard_test.go
```diff
@@ -600,799 +600,874 @@ func TestShard_Close_RemoveIndex(t *testing.T) {
...
+			Time:  time.Unix(0, 0).UnixNano(),
+			Value: 100,
+			Aux:   []interface{}{(*float64)(nil)},
+		}) {
+			t.Fatalf("unexpected point(2): %s", spew.Sdump(p))
+		}
+	}
+
+	for _, index := range tsdb.RegisteredIndexes() {
+		t.Run(index, func(t *testing.T) { test(index) })
```
benbjohnson: Why is this calling an variable function instead of being nested inside `Run`?

tests/server_test.go
e-dard: You mean the `skip` fields on the `query` type? How would we skip tests we don't want to run? I'll admit I have no idea why we're skipping any of them in the first place :)

tsdb/engine/tsm1/engine_test.go
e-dard: I didn't know that was an option. Neat.

tsdb/engine/tsm1/engine_test.go
e-dard: Here or _everywhere_? Are you concerned about indentation?

tsdb/shard_test.go
e-dard: I was concerned about the indentation of nesting everything inside of `t.Run`

tests/server_test.go
benbjohnson: I believe the `skip` field was there to allow someone to isolate a test instead of commenting out the other tests. With subtests we can isolate with `-run`.

tsdb/engine/tsm1/engine_test.go
benbjohnson: I thought it was being used in a bunch of places but it seems to be less than I thought. Indentation was one concern. Another concern was DRY and consistency.

tests/server_test.go
jwilder: I think we still need skip.  We've marked things as skip in the past because the tests ended up being flaky and need to be be disabled.  If we just comment them out, the output in the log is lost so we can't easily just search the output to see if any tests happened to be skipped.

benbjohnson: We can use `t.Skip()` on the subtest.

tsdb/engine/tsm1/engine_test.go
e-dard: Hmmm:

```
BenchmarkEngine_WritePoints/10/inmem-4   	   20000	     77541 ns/op	    2701 B/op	      48 allocs/op
BenchmarkEngine_WritePoints/10#01/tsi1-4 	   20000	     75302 ns/op	    2702 B/op	      48 allocs/op
```

Not exactly as nice.

benbjohnson: Does it work if you invert `index` & `sz`? Does it change it to:

```
BenchmarkEngine_WritePoints/inmem/10-4   	   20000	     77541 ns/op	    2701 B/op	      48 allocs/op
BenchmarkEngine_WritePoints/tsi1/10/-4 	   20000	     75302 ns/op	    2702 B/op	      48 allocs/op
```

PR #8691, opened 2017-08-11
Compressed Query Execution

jwilder:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

This adds a new `count` iterator that executes queries directly on compressed TSM data and avoids decompression when possible as well as uses compression specific counts to speed up query execution.  This seems to improve all `count` queries by 98% in my tests.  I have only implemented this for counts, but should be able to extend this to `min`, `max`, `first`, `last`, `sum` and `mean` in the future.  The optimization is only enabled when the query filter criteria includes only time and tag filters.  If a field value needs to filters (e.g. `value > 100`), the existing count is used.  This can be further optimized later by pushing down predicates.  In the case where there are tombstones or overlapping blocks, a slower path is used that decodes and merges the values.  This could also be optimized further to avoid decoding values.

The results below are using two different data sets.  The first is a database w/ 100k series and 1B values spanning 50 years.  The second is a database w/ 3.2B values across 5 years with 50k series.  I ran these a few times to allow the page cache to warm up.

These are the queries I used:
```
select count(value) from cpu
select count(value) from cpu WHERE time > '1969-12-31T23:59:59.999995001Z' GROUP BY time(365d)
select count(value) from cpu WHERE time > '1969-12-31T23:59:59.999995001Z' GROUP BY time(30d)
select count(value) from cpu where time > '2015-01-01T00:00:00Z' and host =~ /^server\d$/ group by *, time(365d)
select count(value) from cpu where time > '2015-01-01T00:00:00Z' and host =~ /^server\d{1,2}$/ group by *, time(365d)
select count(value) from cpu where time > '2015-01-01T00:00:00Z' and host =~ /^server\d{1,3}$/ group by *, time(365d)
select count(value) from cpu where time > '2015-01-01T00:00:00Z' and host =~ /^server\d{1,4}$/ group by *, time(365d)
select count(value) from cpu where time > '2015-01-01T00:00:00Z' and host =~ /^server\d{1,5}$/ group by *, time(365d)
```

```
$ query-comparer -time_delta 1 /tmp/old.log /tmp/new.log
2017/08/11 14:05:48 Query 0 tooks 26.493525106s in first instance, but 428.484116ms in second. One input is 98.4% faster
2017/08/11 14:05:48 Query 1 tooks 31.237530382s in first instance, but 512.416927ms in second. One input is 98.4% faster
2017/08/11 14:05:48 Query 2 tooks 31.661536852s in first instance, but 488.075753ms in second. One input is 98.5% faster
2017/08/11 14:05:48 Query 3 tooks 78.321071ms in first instance, but 13.666434ms in second. One input is 82.6% faster
2017/08/11 14:05:48 Query 4 tooks 660.307493ms in first instance, but 22.787475ms in second. One input is 96.5% faster
2017/08/11 14:05:48 Query 5 tooks 6.541713447s in first instance, but 104.728203ms in second. One input is 98.4% faster
2017/08/11 14:05:48 Query 6 tooks 1m7.887726037s in first instance, but 876.052676ms in second. One input is 98.7% faster
2017/08/11 14:05:49 Query 7 tooks 7m31.876794893s in first instance, but 4.683279507s in second. One input is 99.0% faster
2017/08/11 14:05:49 Total pairs of queries compared: 8. Passed: 8
2017/08/11 14:05:49 Total duration first input: 10m16.437455281s, total duration second input: 7.129491091s. Difference 98.8%
2017/08/11 14:05:49 Total differences: 0
```

The performance improvements are due to a few things:

* Avoiding value decompression - Counts do not need to decode the values.  In the unoptimized version, all values are fully decoded and counted individually (ignoring the values).  In this version, values are not decoded at all and only the timestamp portion of a block is inspected.
* Reduced allocations - since the counts are pushed down to the lower level storage, we do not need to allocate a `*Value` for every item to be counted.  We only return a single count for a time range.
* Compression specific counts -  Timestamps are encoded 3 different ways: run length encoded (RLE), simple8b (bit packed), and raw.  Counting RLE blocks only needs to returns the count attribute in the block header.  Simple8b only needs to inspect the selector bits to determine the count for an `uint64` and then sum them.  Raw values can return the count attribute in the header.
* Compression specific partial counts -  When a block overlaps a time window, a faster partial count is used that operates directly on the compressed data.


query/call_iterator.go
```diff
@@ -39,6 +39,11 @@ func NewCallIterator(input Iterator, opt IteratorOptions) (Iterator, error) {
 	name := opt.Expr.(*influxql.Call).Name
 	switch name {
 	case "count":
+		// Count is optimized in the lower level storage to return an aggregate sum,
+		// per window, if there are no field conditions.
+		if opt.IndexOnlyConditions() {
```
jsternberg: This shouldn't be here. If you have are merging multiple iterators into a call iterator, it will already use the sum iterator instead. So you just need to ensure that each series inside of the storage engine uses the optimized count cursor and the other logic should take care of the rest.

Look for the comment "Wrap each series in a call iterator". Since the decision to create the count cursor needs to be done inside of build cursor, it might be best to move the logic in this area into the creation of the iterators too.

Sorry if that's a bit confusing. Please tell me if you need help with that.

influxql/ast.go
```diff
@@ -472,6 +472,7 @@ func IsSystemName(name string) bool {
 	case "_fieldKeys",
 		"_measurements",
 		"_series",
+		"_seriesKey",
```
jwilder: Yes.  Because its called from here: https://github.com/influxdata/influxdb/pull/8691/files#diff-0cae4acf5479b75328bd79505447da9eR980 and to fix `TestServer_Query_ShowSeriesCardinality`.

query/call_iterator.go
jwilder: The issue I ran into (and why I did it this way) was that a `count` seems to expects N values to be returned from the series and it counts each one.  The fast iterator return 1 point with the value as the count.  Using the existing logic, it would see that as 1 value.  I needed it to sum each point though.

I'll need to pair up with you so I can better understand what you mean.

tsdb/engine/tsm1/file_store.go
```diff
@@ -1051,6 +1056,42 @@ func (c *KeyCursor) Next() {
 	}
 }
 
+// NextWindow resets the cursor to read values between min, max.  Values already
+// read not be returned again.  Returns true if there are more values that can
```
jwilder: Yep.  Will fix.

epa095:
What happened to this @jwilder ? It seems really useful, and for graphing large series a similar optimization for first/last seems equally cool. Is this PR superseded by something, or is it just dead in the water?

jwilder:
@epa095 I didn't get a chance to update this after the initial PR review.  It's not dead, just low priority at the moment.  I may break some of this up into smaller pieces so that it can be used by IFQL work.

CAFxX:
@jwilder just a note to say we are running into something that could be solved by this. We have certain series that contain individual (non-aggregated) events; most of the time the number of events per time interval is very low but sometimes a burst of activity occurs and the result is that `count()` and `sum()` over such series, if the query time range includes one such period, spikes all cores for minutes (data is already in the kernel page cache, so IO is basically not existant).

PR #8706, opened 2017-08-15
fix: cursor leak when cur == nil and aux or conds is not empty

stuartcarnie:
* symptom would be locked `.tsm` files resulting in an accumulation of `.tsm.tmp` files after a compaction

Scenarios

* aggregate queries that included fields in the `where` condition, where the aggregate field does not exist for all shards. example `select count(*) where some_field = 0` in which case `conds` cursors will leak
* selectors that include additional fields and the field in the selector does not exist across all shards, like `select first(foo), v0 from m0 limit 10`, in which case `aux` cursors will leak

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/tsdb/engine/tsm1/iterator.go
+++ b/tsdb/engine/tsm1/iterator.go
@@ -140,3 +140,11 @@ func (c *stringSliceCursor) nextString() (int64, string) {
 	c.values = c.values[1:]
 	return 0, value
 }
+
+type cursorsAt []cursorAt
+
+func (c cursorsAt) close() {
+	for _, cur := range c {
+		cur.close()
+	}
+}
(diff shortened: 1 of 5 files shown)


tsdb/engine/tsm1/iterator.go
```diff
@@ -121,3 +121,9 @@ var (
 	nilStringLiteralValueCursor   cursorAt = &literalValueCursor{value: (*string)(nil)}
 	nilBooleanLiteralValueCursor  cursorAt = &literalValueCursor{value: (*bool)(nil)}
 )
+
+func closeAllCursorAt(cur []cursorAt) {
```
jwilder: Could you do something like:

```
type CursorsAt []cursorAt
func (a CursorsAt) Close() {
    for _, c := range a {
        c.close()
    }
}
```

And then `CursorsAt(itr.aux).Close()` or maybe change the type of `aux` and `conds.curs` to `CursorsAt` so you can just call `itr.aux.Close()`?

stuartcarnie: I should have gone with that first, as it did cross my mind. I'll update it

stuartcarnie: I'll keep it unexported, given `cursorAt` is also

PR #8708, opened 2017-08-16
meta: apply authorization to SHOW MEASUREMENTS

joelegasse:
This change will allow limiting the results from `SHOW MEASUREMENTS` according to enterprise authorization rules.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/tsdb/index/inmem/meta.go
+++ b/tsdb/index/inmem/meta.go
@@ -395,6 +395,20 @@ func (m *Measurement) TagSets(shardID uint64, opt influxql.IteratorOptions) ([]*
 	return sortedTagsSets, nil
 }
 
+func (m *Measurement) AuthorizeRead(auth influxql.Authorizer) bool {
+	if auth == nil {
+		return true
+	}
+
+	for _, s := range m.SeriesByIDMap() {
+		if auth.AuthorizeSeriesRead(m.database, m.name, s.tags) {
+			return true
+		}
+	}
+
+	return false
+}
+
 // intersectSeriesFilters performs an intersection for two sets of ids and filter expressions.
 func intersectSeriesFilters(lids, rids SeriesIDs, lfilters, rfilters FilterExprs) (SeriesIDs, FilterExprs) {
 	// We only want to allocate a slice and map of the smaller size.
--- a/coordinator/statement_executor.go
+++ b/coordinator/statement_executor.go
@@ -684,7 +684,7 @@ func (e *StatementExecutor) executeShowMeasurementsStatement(q *influxql.ShowMea
 		return ErrDatabaseNameRequired
 	}
 
-	names, err := e.TSDBStore.MeasurementNames(q.Database, q.Condition)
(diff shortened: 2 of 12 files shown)


e-dard:
General question: what's the reason behind not implementing FGA on `SHOW MEASURMENTS` for the `tsi1` index?

joelegasse:
Just not implemented yet, I requested the reviews to get some early
feedback.

e-dard:
@joelegasse sorry, I edited my message as you replied 😄

joelegasse:
The authorizer is the user. Ultimately we need to perform lower level operations (filtering series) per-user. The original code base assumes full access at the database level, so execution context (originating user) was dropped after the initial check was done in the query handler. The Authorizer interface gives us an abstraction of access rules that can be expanded later if needed.

tsdb/index/inmem/meta.go
```diff
@@ -395,6 +395,20 @@ func (m *Measurement) TagSets(shardID uint64, opt influxql.IteratorOptions) ([]*
 	return sortedTagsSets, nil
 }
 
+func (m *Measurement) AuthorizeRead(auth influxql.Authorizer) bool {
```
e-dard: I'm concerned about how this is going to perform in some possibly unusual/extreme situations. Apologies if I'm not understanding how the `auth` methods work internally though...

If I understand this properly, this is relying on using unordered iteration of the series on the measurement to find at least one series that we're authorised to read from. So it's the case that a user can read a measurement if they can read at least one series on that measurement, right?

I'm assuming the worse case here would be if we have some tags that we can read, but none of the series on the measurement have those tags? In that case we have to iterate over all the series to determine that we can't read any, and therefore can't read this measurement?

When the number of series for a measurement is large, I'm wondering if we would be better off taking in turn the (likely) small set of tag key/value grants we're interested in and binary searching over the ordered list of series. I realise that's probably too significant a change to the API though...

I guess the type of scenarios I'm thinking of would be something like a user being allowed to read values for a certain container ID (or some other high cardinality data), and the measurement in question containing many container IDs, but not the one the user can read from.

joelegasse: Currently the only primitive exposed from the `Authorizer` is read/write access to a specific series. Ideally, yes, there would be a way to invert the rules to more efficiently filter through available series. @jwilder and I discussed possibly trying to extract an `influxql.Expr` from the rules, but they might have to be on a per-measurement basis (potentially large tree checking `_name` for different values). Also, since the representation of a measurement and the relevant series is per-index, we would have to try to export primitives useful to both `inmem` and `tsi`, and `MeasurementNames` already takes an `Expr` for filtering.

I'm all for finding a better way, which is why I asked for some early feedback. This works with what we have now, and is a starting point for improvements. :-)

joelegasse:
I'm going to close this in favor of a simpler version based on #8752

PR #8712, opened 2017-08-16
Remove TimeRange function and replace with a more accurate ConditionExpr function

jsternberg:
The ConditionExpr function is more accurate because it parses the condition
and ensures that time conditions are actually used correctly. That means
that attempting to combine conditions with OR will not result in the query
silently pretending it's an AND and nested conditions work correctly so
there is only one way to read the query.

It also extracts the non-time conditions into a separate condition so we
can stop attempting to parse around the time conditions in lower layers of
the storage engine. This change does not remove those hacks, but a
following commit should be able to sanitize the condition and remove them.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated


benbjohnson:
lgtm but needs a rebase with master to update the `inmem2tsi` command's reference to `query`.

PR #8716, opened 2017-08-17
Ensure inputs are closed on error. Add runtime GC finalizer as additional guard to close iterators

stuartcarnie:
`newMergeGuardIterator` ensures `inputs` are closed if `Merge` returns error and wraps the returned merge iterator in `<type>FinalizerIterator`

* <type>FinalizerIterator sets a runtime finalizer and calls `Close`
  when garbage collected. This will ensure any associated cursors
  are closed and ref counts on related TSM files released
* `query.Iterators#Merge` call could return an error and the inputs
  would not be closed, causing a cursor leak, hence the `newMergeGuardIterator`

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
 - [#8677](https://github.com/influxdata/influxdb/issues/8677): Fix backups when snapshot is empty.
 - [#8706](https://github.com/influxdata/influxdb/pull/8706): Cursor leak, resulting in an accumulation of `.tsm.tmp` files after compactions.
 - [#8712](https://github.com/influxdata/influxdb/pull/8712): Force time expressions to use AND and improve condition parsing.
+- [#8716](https://github.com/influxdata/influxdb/pull/8716): Ensure inputs are closed on error. Add runtime GC finalizer as additional guard to close iterators
 
 ## v1.3.4 [unreleased]
 
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -1563,7 +1563,7 @@ func (e *Engine) CreateIterator(measurement string, opt query.IteratorOptions) (
 				if err != nil {
 					return nil, err
 				}
-				return query.Iterators(itrs).Merge(opt)
+				return newMergeGuardIterator(itrs, opt)
 			}
 		}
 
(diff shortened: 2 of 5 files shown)


stuartcarnie:
updated commits with debug code removed

benbjohnson:
@stuartcarnie I haven't used `SetFinalizer()` before but there are quite a few concerns here: https://groups.google.com/forum/#!topic/golang-dev/DMiUkpS1uyQ

What's the goal?

stuartcarnie:
@benbjohnson to address any consumers of `CreateIterator` failing to call `Close` and consequently leaking cursors. The ultimate issue is the ref count of the associated TSMFile objects never return to 0 and subsequent compactions generating `.tsm.tmp` files until after a restart. The issue is we use a manual, reference counted mechanism to detemine usage of `TSMFile`s associated with seeks in the cursors. That means we rely on consumers to call `Close` to ensure the ref count is decremented. This is a guard until we can spend more time throughly reviewing the query iterators or refactoring the code. Naturally, this is based on the assumption that the iterator returned from `Engine#CreateIterator` is garbage collected and hot referenced indefinitely due to a hung goroutine, etc.

---

I have read that post before. It was started by Dmitry because the Go GC was not a precise garbage collector, thus you could not rely on GC finalizers being called. That has since been rectified with 1.3. Rob Pike [responded](https://groups.google.com/d/msg/golang-dev/DMiUkpS1uyQ/0sHpTacG00YJ) that it wasn't as gloomy as Dmitry made it out to be and later Rob clarified [they were not going to be removed](https://groups.google.com/d/msg/golang-dev/DMiUkpS1uyQ/r6T9N_8ezw4J). They are used for a [similar][0] [purpose][1] in the `net` package to ensure the `Close` method of garbage collected file descriptors is called in order to free resources.

[0]: https://github.com/golang/go/blob/master/src/net/fd_unix.go#L52
[1]: https://github.com/golang/go/blob/0b81c023a722304304539cdbb0c6229bca481b09/src/os/file_unix.go#L118

stuartcarnie:
Agreed @e-dard, the long term goal would be to make sure `Close` is called.

The other struct names for the iterators in the `iterator.gen.go` file start with the data type (`float`, `integer`, etc), so I followed that convention

**Q:** Do you think we should change that? 

`query.Iterators(input).Merge(opt)` returns `query.Iterator` and the `<type>FinalizerIterator` wraps any `query.Iterator`, including merge iterators or the various call iterators returned by the `Merge` function.

I'll rename the `newMergeGuardIterator` to `newMergeFinalizerIterator`. It is intended as a guard to make sure the `inputs` are closed if the call to `query.Iterators(inputs).Merge(opt)` returned an error. If it succeeds, it wraps the resulting `query.Iterator` in the `Finalizer` iterator.

e-dard:
@stuartcarnie I think the changes are good. 🍏

PR #8725, opened 2017-08-21
Refactor how results are returned to the client

jsternberg:
This refactors the internal result returning system to match more
closely how we iterative return points between result sets, series, and
rows within the series. It uses the updated terminology rather than
older terminology that no longer matches how we refer to things in the
documentation or within the query engine.

The refactor moves the aggregation and chunking behavior from
`influxql.Emitter` to the HTTP service so that behavior is isolated to
one location rather than sprinkled around in multiple locations.

- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated


jsternberg:
This includes the refactor from #8477, but rebased to resolve merge conflicts and it removes the new response format. So this just refactors the internal structure of how results are constructed instead of adding additional functionality on top of that.

PR #8735, opened 2017-08-23
Ensure that sorted heaps are merged correctly

### LINKED ISSUE #8695: System iterators returning data out of order
Currently system iterators are not being merged correctly across multiple shards. Specifically they're being merged in an arbitrary order. 

While it appears as if this issue is only affecting the `tsi1` index, the iterators are incorrectly being merged in both the `inmem` and `tsi1` indexes. However, since the `inmem` index is actually duplicated across shards, the result of merging system iterators is correct (even though the logic to merge them is faulty!!!)

This caused me much frustration and amusement at the same time when debugging this issue.

Using the `tsi1` index, here are some examples of incorrect behaviour:

##### SHOW SERIES
```
>
> create database db
> insert cpu,host=serverB value=1
> insert cpu,host=serverA value=1 1689894000000000000
>
>
> show series
key
---
cpu,host=serverB
cpu,host=serverA
>
>
> show shards
name: _internal
id database  retention_policy shard_group start_time           end_time             expiry_time          owners
-- --------  ---------------- ----------- ----------           --------             -----------          ------
1  _internal monitor          1           2017-08-11T00:00:00Z 2017-08-12T00:00:00Z 2017-08-19T00:00:00Z
10 _internal monitor          10          2017-08-14T00:00:00Z 2017-08-15T00:00:00Z 2017-08-22T00:00:00Z
```

##### SHOW TAG KEYS
```
> create database db
> insert cpu,region=west value=1
> insert cpu,host=serverA value=1 1689894000000000000
>
>
> show tag keys
name: cpu
tagKey
------
region
host
>
>
> show shards
name: _internal
id database  retention_policy shard_group start_time           end_time             expiry_time          owners
-- --------  ---------------- ----------- ----------           --------             -----------          ------
1  _internal monitor          1           2017-08-11T00:00:00Z 2017-08-12T00:00:00Z 2017-08-19T00:00:00Z
10 _internal monitor          10          2017-08-14T00:00:00Z 2017-08-15T00:00:00Z 2017-08-22T00:00:00Z
```

##### SHOW FIELD KEYS
```
> create database db
> insert cpu value=1
> insert cpu count=1 1689894000000000000
> show field keys
name: cpu
fieldKey fieldType
-------- ---------
value    float
count    float
>
>
> show shards
name: _internal
id database  retention_policy shard_group start_time           end_time             expiry_time          owners
-- --------  ---------------- ----------- ----------           --------             -----------          ------
1  _internal monitor          1           2017-08-11T00:00:00Z 2017-08-12T00:00:00Z 2017-08-19T00:00:00Z
10 _internal monitor          10          2017-08-14T00:00:00Z 2017-08-15T00:00:00Z 2017-08-22T00:00:00Z
```

`SHOW TAG VALUES` and `SHOW MEASUREMENTS` are unaffected because they use the index directly and are not converted into `SELECT` queries.

The issue boils down to the merging operations in `influxql`, which currently considers the following facets of a point when merging it with others: (1) its measurement name; (2) its tag keys/values; (3) time.

Since regular points usually have all three of those, they're always sorted correctly across shards. The results of system iterators however do not have names, tags or time values and so end up being merged arbitrarily between streams of iterators. Since the `inmem` index streams are always identical though, the result is always correct.

e-dard:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

When merging streams of system iterators we don't use tags or time.
Instead we add series keys (in the case of, for example, `SHOW SERIES`)
to the `Aux` field of the iterators' elements. This is because we only
emit merged and sorted sets of series key to the client.

We currently use `SortedMergeHeap`s to merge together multiple
iterators, and the comparitor function did not consider `Aux` fields
when determining which heap to pop the next item off during a merge. As
such, `SHOW SERIES` and `SHOW TAG KEYS` (any meta query that gets
converted into a special type of `SELECT`) were returning results in
arbitrary order.

This issue was never noticed on the `inmem` index because the streams
are always duplicates of each other, and of course it doesn't matter if
you arbitrarily merge together two idential, sorted streams...

The issue first manifested itself on the `tsi1` index, but this fix will
apply to both indexes.

Fixes #8695.

### DIFF
--- a/query/iterator.gen.go
+++ b/query/iterator.gen.go
@@ -395,6 +395,12 @@ func (itr *floatSortedMergeIterator) pop() (*FloatPoint, error) {
 }
 
 // floatSortedMergeHeap represents a heap of floatSortedMergeHeapItems.
+// Items are sorted with the following priority:
+//     - By their measurement name;
+//     - By their tag keys/values;
+//     - By time; or
+//     - By their Aux field values.
+//
 type floatSortedMergeHeap struct {
 	opt   IteratorOptions
 	items []*floatSortedMergeHeapItem
@@ -411,7 +417,26 @@ func (h *floatSortedMergeHeap) Less(i, j int) bool {
 		} else if xTags, yTags := x.Tags.Subset(h.opt.Dimensions), y.Tags.Subset(h.opt.Dimensions); !xTags.Equals(&yTags) {
 			return xTags.ID() < yTags.ID()
 		}
-		return x.Time < y.Time
+
+		if x.Time != y.Time {
+			return x.Time < y.Time
+		}
+
+		if len(x.Aux) > 0 && len(x.Aux) == len(y.Aux) {
+			for i := 0; i < len(x.Aux); i++ {
+				v1, ok1 := x.Aux[i].(string)
+				v2, ok2 := y.Aux[i].(string)
+				if !ok1 || !ok2 {
+					// Unsupported types used in Aux fields. Maybe they
+					// need to be added here?
+					return false
+				} else if v1 == v2 {
+					continue
+				}
+				return v1 < v2
+			}
+		}
+		return false // Times and/or Aux fields are equal.
 	}
 
 	if x.Name != y.Name {
@@ -419,7 +444,26 @@ func (h *floatSortedMergeHeap) Less(i, j int) bool {
 	} else if xTags, yTags := x.Tags.Subset(h.opt.Dimensions), y.Tags.Subset(h.opt.Dimensions); !xTags.Equals(&yTags) {
 		return xTags.ID() > yTags.ID()
 	}
-	return x.Time > y.Time
+
+	if x.Time != y.Time {
+		return x.Time > y.Time
+	}
+
+	if len(x.Aux) > 0 && len(x.Aux) == len(y.Aux) {
+		for i := 0; i < len(x.Aux); i++ {
+			v1, ok1 := x.Aux[i].(string)
+			v2, ok2 := y.Aux[i].(string)
+			if !ok1 || !ok2 {
+				// Unsupported types used in Aux fields. Maybe they
+				// need to be added here?
+				return false
+			} else if v1 == v2 {
+				continue
+			}
+			return v1 > v2
+		}
+	}
+	return false // Times and/or Aux fields are equal.
 }
 
 func (h *floatSortedMergeHeap) Push(x interface{}) {
@@ -3742,6 +3786,12 @@ func (itr *integerSortedMergeIterator) pop() (*IntegerPoint, error) {
 }
 
 // integerSortedMergeHeap represents a heap of integerSortedMergeHeapItems.
+// Items are sorted with the following priority:
+//     - By their measurement name;
+//     - By their tag keys/values;
+//     - By time; or
+//     - By their Aux field values.
+//
 type integerSortedMergeHeap struct {
 	opt   IteratorOptions
 	items []*integerSortedMergeHeapItem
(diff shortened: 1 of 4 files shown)


query/iterator.gen.go.tmpl
```diff
@@ -409,14 +415,44 @@ func (h *{{$k.name}}SortedMergeHeap) Less(i, j int) bool {
...
+						// Unsupported types used in Aux fields. Maybe they
+						// need to be added here?
+						break
+					}
+					return v1 < v2
+				}
+			}
+		}
 		return x.Time < y.Time
 	}
```
benbjohnson: I think this function would flow better if you handle the time comparison first & then compare aux values, and then handle aux lengths last:

```go
if x.Time != y.Time {
	return x.Time < y.Time
}

for i := 0; i < len(x.Aux) && i < len(y.Aux); i++ {
	if cmp := Compare(x.Aux[i], y.Aux[i]); cmp != 0 {
		return cmp == -1
	}
}

return len(x.Aux) < len(y.Aux)
```

Then move the comparison function out:

```go
func Compare(a, b interface{}) int {
	switch a := a.(type) {
	case string:
		if b, ok := b.(string); ok {
			return strings.Compare(a, b)
		}
	}
	return 0
}
```

e-dard: @benbjohnson ha, I noticed the indentation almost as soon as I submitted the PR 😄

e-dard: `strings.Compare` is always slower than using the builtin operator so I didn't move the comparison into a separate function because it becomes clunky when you want to do either `a < b` or `a > b`. 

A little bit of duplication seemed appropriate in this case.

benbjohnson: Sure, that's a fair point about perf. 👌

query/iterator.gen.go
```diff
@@ -411,14 +417,48 @@ func (h *floatSortedMergeHeap) Less(i, j int) bool {
...
+				if !ok1 || !ok2 {
+					// Unsupported types used in Aux fields. Maybe they
+					// need to be added here?
+					return false
+				} else if v1 == v2 {
+					continue
+				}
+				return v1 < v2
+			}
+		}
```
benbjohnson: This only compares `Aux` fields if they are of equal length. They *should* always be equal length but it's incorrect if they aren't.

query/iterator.gen.go
```diff
@@ -411,14 +417,48 @@ func (h *floatSortedMergeHeap) Less(i, j int) bool {
...
+				v2, ok2 := y.Aux[i].(string)
+				if !ok1 || !ok2 {
+					// Unsupported types used in Aux fields. Maybe they
+					// need to be added here?
+					break
+				}
+				return v1 > v2
+			}
+		}
+	}
```
benbjohnson: This block needs the same changes as above: 1) Consolidate time comparison, 2) Don't return on first aux.

e-dard: Doh. I think I've done enough coding today....

query/iterator.gen.go
e-dard: If they're of different lengths then we wouldn't want to sort based on that would we? Given we can't return an error my thinking was "if the aux fields are not comparable then ignore that aspect of comparison", which leaves us with two equal times, so we `return false`.

stuartcarnie: Given the `<type>Point#Aux` field is populated, assuming the length of the original `IteratorOptions#Aux` field, [here][0], therefore I would argue this is currently an invariant. Additionally, the only place the `aux` field of the iterator is built is via [buildVarRefSeriesIterator][1], again derived from the `IteratorOptions#Aux` field

[0]: https://github.com/influxdata/influxdb/blob/d189621d072a6fc3224cfb46c52855ac085d5317/tsdb/engine/tsm1/iterator.gen.go.tmpl#L220-L222
[1]: https://github.com/influxdata/influxdb/blob/d189621d072a6fc3224cfb46c52855ac085d5317/tsdb/engine/tsm1/engine.go#L1853

stuartcarnie: If `len(aux) < len(opt.Aux)` that first link would panic, too.

PR #8743, opened 2017-08-25
Prepare for Go 1.9 math/bits package.

e-dard:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

**Note** Should not be merged until we decide to move to Go 1.9.

This PR replaces the `github.com/dgryksi/go-bits` dependency with the standard library `math/bits` package.

### DIFF
--- a/pkg/bits/bits.go
+++ b/pkg/bits/bits.go
@@ -1,0 +1,19 @@
+// +build !go1.9
+
+package bits
+
+import (
+	"github.com/dgryski/go-bits"
+)
+
+// LeadingZeros64 counts leading zeroes.
+var LeadingZeros64 = bits.Clz
+
+// TrailingZeros64 counts trailing zeroes.
+var TrailingZeros64 = bits.Ctz
+
+// LeadingZeros32 uses a 64-bit clz implementation and
+// reduces the result by 32, to get a clz for a word.
+func LeadingZeros32(x uint32) uint64 {
+	return bits.Clz(uint64(x)) - 32
+}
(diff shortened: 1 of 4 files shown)


e-dard:
@jwilder sorry I did run the benchmarks before pushing this, but didn't attach them to the issue. Here are the ones for the `Float{Encoder, Decoder}`:

```
⇒  benchstat /tmp/old.txt /tmp/new.txt
name            old time/op    new time/op    delta
FloatEncoder-4    7.15µs ±10%    6.95µs ± 6%    ~           (p=0.119 n=11+10)
FloatDecoder-4    1.89µs ± 7%    1.81µs ± 3%  -4.23%        (p=0.003 n=11+11)

name            old alloc/op   new alloc/op   delta
FloatEncoder-4      624B ± 0%      624B ± 0%    ~     (all samples are equal)
FloatDecoder-4    0.00B ±NaN%    0.00B ±NaN%    ~     (all samples are equal)

name            old allocs/op  new allocs/op  delta
FloatEncoder-4      4.00 ± 0%      4.00 ± 0%    ~     (all samples are equal)
FloatDecoder-4     0.00 ±NaN%     0.00 ±NaN%    ~     (all samples are equal)
```

Not a huge amount in it; I think the change is fine.

I'll add build tags.

e-dard:
@jwilder I think this is in good shape now. Had to do some fiddly abstractions to ensure that the committed code worked both on Go 1.9 and previous versions. When we commit to Go 1.9 I'll remove the `pkg/bits` package and directly call into `math/bits` from the `hll` and `tsm1`.

PR #8751, opened 2017-08-28
Update parser and AST for explain statement

jsternberg:
- [x] Rebased/mergable
- [ ] Tests pass


jsternberg:
I've now added this same code in at least 10 different branches so I wanted to get this part in to reduce the diff of any implementation. I can wait to add this if desired, but adding this means I don't have to worry about any AST related stuff while working on the implementation.

PR #8752, opened 2017-08-28
Use system cursors for measurement, series, and tag key meta queries.

benbjohnson:
## Overview

This pull request changes several meta queries to use the new system cursors instead of the higher level system iterators used before. This allows the queries to filter based on time (with only shard-level time granularity).

```
SHOW SERIES
SHOW MEASUREMENTS
SHOW TAG KEYS
```

This does not include `SHOW TAG VALUES` or `SHOW FIELD KEYS`.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


influxql/statement_rewriter.go
```diff
@@ -457,3 +454,18 @@ func rewriteSourcesCondition(sources Sources, cond Expr) Expr {
 	}
 	return scond
 }
+
+func rewriteSources2(sources Sources, database string) Sources {
+	if len(sources) == 0 {
+		sources = Sources{&Measurement{Regex: &RegexLiteral{Val: regexp.MustCompile(`.+`)}}}
```
joelegasse: rather than recompiling the regex every time, the compilation can be done in an `init()` function, then [`Regexp.Copy()`](https://golang.org/pkg/regexp/#Regexp.Copy) can be used here.

influxql/ast.go
```diff
@@ -1066,6 +1066,12 @@ type SelectStatement struct {
 	// Removes the "time" column from the output.
 	OmitTime bool
 
+	// Removes measurement name from resulting query. Useful for meta queries.
+	StripName bool
```
jsternberg: I'm not a big fan of adding these two options, but I don't see an alternative at the moment.

PR #8754, opened 2017-08-28
Force subqueries to match the parent queries ordering

### LINKED ISSUE #8699: subquery doesn't support order by
InfluxDB version: 1.3.1  

When I run the subquery with `order by`, it doesn't work:

```
> select * from cpu order by time desc
name: cpu
time                host    region  value
----                ----    ------  -----
1502764231978282504 serverA us_west 0.6
1502763228833832379 serverA us_west 0.61
1502763224144379689 serverA us_west 0.64
> 
> select * from (select * from cpu) order by time desc
name: cpu
time                host    region  value
----                ----    ------  -----
1502763224144379689 serverA us_west 0.64
1502763228833832379 serverA us_west 0.61
1502764231978282504 serverA us_west 0.6
```

jsternberg:
Previously, subqueries would honor their own ordering. We never really
supported that and I have no idea if it would work since most parts in
the query engine assume that points are being delivered in only one
ordering.

Subqueries have now been modified so if a person tries to do different
ordering, they get an error when running the query. If they specify an
ordering in the top most query, that ordering gets propagated to all
subqueries.

Fixes #8699.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/query/compile.go
+++ b/query/compile.go
@@ -38,6 +38,9 @@ type compiledStatement struct {
 	// a query that shouldn't have an interval to fail.
 	InheritedInterval bool
 
+	// Ascending is true if the time ordering is ascending.
+	Ascending bool
+
 	// FunctionCalls holds a reference to the call expression of every function
 	// call that has been encountered.
 	FunctionCalls []*influxql.Call
@@ -114,6 +117,7 @@ func Compile(stmt *influxql.SelectStatement, opt CompileOptions) (Statement, err
 
 // preprocess retrieves and records the global attributes of the current statement.
 func (c *compiledStatement) preprocess(stmt *influxql.SelectStatement) error {
+	c.Ascending = stmt.TimeAscending()
 	c.Limit = stmt.Limit
 	c.HasTarget = stmt.Target != nil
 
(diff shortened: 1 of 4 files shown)


query/compile.go
```diff
@@ -764,6 +768,19 @@ func (c *compiledStatement) subquery(stmt *influxql.SelectStatement) error {
...
+	if subquery.Ascending != c.Ascending {
+		// If the ordering is different, but the subquery has no sort fields
+		// (so the user never specified they wanted it to be ascending), inherit
+		// the ordering from the parent query.
+		if len(stmt.SortFields) == 0 {
+			subquery.Ascending = c.Ascending
+		} else {
+			return errors.New("subqueries must be ordered in the same direction as the query itself")
+		}
+	}
```
benbjohnson: Do the sort fields need to be the same if `Ascending` is the same?

jsternberg: Well, the sort fields have to be the same if `Ascending` is the same. Although I think we could largely change this logic to `if len(stmt.SortFields) != 0 && subquery.Ascending != c.Ascending`. I'll change it to that so there's less nesting.

PR #8770, opened 2017-08-30
Reduce how long it takes to walk the varrefs in an expression

jsternberg:
This is used quite a bit to determine which fields are needed in a
condition. When the condition gets large, the memory usage begins to
slow it down considerably and it doesn't take care of duplicates.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


jsternberg:
Benchmark comparison:

```
benchmark                old ns/op     new ns/op     delta
BenchmarkExprNames-8     131087        8389          -93.60%

benchmark                old allocs     new allocs     delta
BenchmarkExprNames-8     308            10             -96.75%

benchmark                old bytes     new bytes     delta
BenchmarkExprNames-8     135424        240           -99.82%
```

PR #8771, opened 2017-08-31
This restores the old time range behavior

jsternberg:
All time ranges are combined with AND regardless of context and
regardless of whether it makes any logical sense.

That was the previous behavior and, unfortunately, a lot of people rely
on it.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

This change modifies the changelog for #8712.


query/compile_test.go
```diff
@@ -145,7 +145,7 @@ func TestCompile_Failures(t *testing.T) {
 		{s: `SELECT bottom(value, 2.5) FROM cpu`, err: `expected integer as last argument in bottom(), found 2.500`},
 		{s: `SELECT bottom(value, -1) FROM cpu`, err: `limit (-1) in bottom function must be at least 1`},
 		{s: `SELECT bottom(value, 3) FROM cpu LIMIT 2`, err: `limit (3) in bottom function can not be larger than the LIMIT (2) in the select statement`},
-		{s: `SELECT value FROM cpu WHERE time >= now() - 10m OR time < now() - 5m`, err: `cannot use OR with time conditions`},
+		//{s: `SELECT value FROM cpu WHERE time >= now() - 10m OR time < now() - 5m`, err: `cannot use OR with time conditions`},
```
benbjohnson: Can you either remove this line or add a comment about why it's commented out?

PR #8776, opened 2017-08-31
Initial implementation of explain plan

jsternberg:
It prints the statistics of each iterator that will access the storage
engine. For each access of the storage engine, it will print the number
of shards that will potentially be accessed, the number of files that
may be accessed, the number of series that will be created, the number
of blocks, and the size of those blocks.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): issue filed or pull request submitted \<[influxdata/docs.influxdata.com#1256](https://github.com/influxdata/docs.influxdata.com/issues/1256)>


jsternberg:
I pushed a small change so that attempting to use analyze would return an error.

I want to keep the AST parsing for analyze there because we want to add it, but it's not implemented and I don't want to give people the impression it is implemented.

PR #8784, opened 2017-09-03
Add support for Prometheus remote read and write API.

pauldix:
Adds a new package prometheus for converting from remote reads and writes to Influx queries and points. Adds two new endpoints to the httpd handler to support prometheus remote read at /api/v1/prom/read and remote write at /api/v1/prom/write.

I've tested just against a local Prometheus server. Will have to prove it out over a large installation so we'd probably want to release it as a beta feature.

I wanted to get this up for review for the approach. Could add unit tests to the `prometheus` package, but the handler test should hit the main points. Was thinking that I may just hold this and base the query stuff off what @stuartcarnie is doing for the v2 read API. It'll have a similar structure as Prometheus to support the v2 data model.


phemmer:
Just curious, why add this here instead of as a telegraf input?
Does this mean that we might see current telegraf listener inputs ported to InfluxDB in the future?

pauldix:
@phemmer It's just another input method kind of like we support OpenTSDB, Carbon, and CollectD. It's writing data directly to the DB, where Telegraf is more of a collector. Although admittedly Telegraf has gained more than that over time. Which listener inputs were you thinking of in Telegraf?

prometheus/converters.go
```diff
@@ -0,0 +1,170 @@
...
+			Op:  influxql.LTE,
+			LHS: &influxql.VarRef{Val: "time"},
+			RHS: &influxql.TimeLiteral{Val: time.Unix(0, q.EndTimestampMs*int64(time.Millisecond))},
+		},
+	}, nil
+}
+
+// TagsToLabelPairs converts a map of Influx tags into a slice of Prometheus label pairs
+func TagsToLabelPairs(tags map[string]string) []*remote.LabelPair {
+	var pairs []*remote.LabelPair
```
e-dard: `pairs := make([]*remote.LabelPair, 0, len(tags))`

services/httpd/handler.go
```diff
@@ -760,6 +777,277 @@ func convertToEpoch(r *query.Result, epoch string) {
...
+	// Convert the Prometheus remote write request to Influx Points
+	var req remote.WriteRequest
+	if err := proto.Unmarshal(reqBuf, &req); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	points, err := prometheus.WriteRequestToPoints(&req)
+	if err != nil {
+		h.Logger.Info(fmt.Sprintf("promerr: %s", err.Error()))
```
e-dard: So if `WriteTracing` is on we log `err` twice? Seems a bit redundant.

services/httpd/handler.go
```diff
@@ -760,6 +777,277 @@ func convertToEpoch(r *query.Result, epoch string) {
...
+			h.httpError(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+	}
+
+	// Determine required consistency level.
+	level := r.URL.Query().Get("consistency")
+	consistency := models.ConsistencyLevelOne
+	if level != "" {
+		var err error
```
e-dard: You don't need to declare `err` here. There is already one in scope.

services/httpd/handler.go
```diff
@@ -760,6 +777,277 @@ func convertToEpoch(r *query.Result, epoch string) {
...
+	}
+	atomic.AddInt64(&h.stats.WriteRequestBytesReceived, int64(buf.Len()))
+
+	if h.Config.WriteTracing {
+		h.Logger.Info(fmt.Sprintf("Prom write body received by handler: %s", buf.Bytes()))
+	}
+
+	reqBuf, err := snappy.Decode(nil, buf.Bytes())
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
e-dard: Can you use `h.httpError` instead? That way we get more error message context in headers and logs.

services/httpd/handler.go
```diff
@@ -760,6 +777,277 @@ func convertToEpoch(r *query.Result, epoch string) {
...
+		return
+	}
+
+	// Convert the Prometheus remote write request to Influx Points
+	var req remote.WriteRequest
+	if err := proto.Unmarshal(reqBuf, &req); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
e-dard: `h.httpError`

services/httpd/handler.go
```diff
@@ -760,6 +777,277 @@ func convertToEpoch(r *query.Result, epoch string) {
...
+	}
+
+	atomic.AddInt64(&h.stats.PointsWrittenOK, int64(len(points)))
+	h.writeHeader(w, http.StatusNoContent)
+}
+
+// servePromRead will convert a Prometheus remote read request into an InfluxQL query and
+// return data in Prometheus remote read protobuf format.
+func (h *Handler) servePromRead(w http.ResponseWriter, r *http.Request, user meta.User) {
+	compressed, err := ioutil.ReadAll(r.Body)
```
e-dard: Every `http.Error` in here should be a `h.httpError`.

services/httpd/handler.go
```diff
@@ -760,6 +777,277 @@ func convertToEpoch(r *query.Result, epoch string) {
...
+		}
+
+		// read the series data and convert into Prometheus samples
+		for _, s := range r.Series {
+			ts := &remote.TimeSeries{
+				Labels: prometheus.TagsToLabelPairs(s.Tags),
+			}
+
+			for _, v := range s.Values {
+				timestamp := v[0].(time.Time).UnixNano() / int64(time.Millisecond) / int64(time.Nanosecond)
```
e-dard: `s.Values` is a type from the prometheus package right?  If so, I would be inclined to avoid the unsafe type assertion on elements in `Values`. Only in case a third party bug causes `v[0]` to not be a `time.Time` (which would cause a panic). 

I'd use `ts, ok := v[0].(time.Time)` I think and check `ok`.

One other thing—`time.Nanosecond == 1`. You don't need to divide by it.

services/httpd/handler.go
```diff
@@ -760,6 +777,277 @@ func convertToEpoch(r *query.Result, epoch string) {
...
+				ts.Samples = append(ts.Samples, &remote.Sample{
+					TimestampMs: timestamp,
+					Value:       v[1].(float64),
```
e-dard: Similar concern here with `v[1].(float64)`

services/httpd/handler.go
```diff
@@ -760,6 +777,277 @@ func convertToEpoch(r *query.Result, epoch string) {
...
+				})
+			}
+
+			resp.Results[0].Timeseries = append(resp.Results[0].Timeseries, ts)
+		}
+	}
+
+	data, err := proto.Marshal(resp)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
e-dard: `h.httpError`

services/httpd/handler.go
```diff
@@ -760,6 +777,277 @@ func convertToEpoch(r *query.Result, epoch string) {
...
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/x-protobuf")
+	w.Header().Set("Content-Encoding", "snappy")
+
+	compressed = snappy.Encode(nil, data)
+	if _, err := w.Write(compressed); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
e-dard: `h.httpError`

services/httpd/handler_test.go
```diff
@@ -524,6 +529,137 @@ func TestHandler_Query_CloseNotify(t *testing.T) {
...
+	b := bytes.NewReader(compressed)
+	h := NewHandler(false)
+	h.MetaClient.DatabaseFn = func(name string) *meta.DatabaseInfo {
+		return &meta.DatabaseInfo{}
+	}
+	called := false
+	h.PointsWriter.WritePointsFn = func(db, rp string, _ models.ConsistencyLevel, _ meta.User, points []models.Point) error {
+		called = true
+		point := points[0]
+		if point.UnixNano() != 1*int64(time.Millisecond) {
```
e-dard: `1*` is redundant.

prometheus/converters.go
```diff
@@ -0,0 +1,170 @@
+package prometheus
```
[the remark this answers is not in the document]
pauldix: Maybe? I didn't put it under there because there wasn't an associated service running. I think of things under `services` as things that have network listeners.

prometheus/converters.go
```diff
@@ -0,0 +1,170 @@
...
+	"time"
+
+	"github.com/influxdata/influxdb/influxql"
+	"github.com/influxdata/influxdb/models"
+	"github.com/prometheus/prometheus/storage/remote"
+)
+
+const (
+	// MeasurementName is where all prometheus time series go to
+	MeasurementName = "_"
```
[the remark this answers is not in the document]
pauldix: Probably :)

prometheus/converters.go
```diff
@@ -0,0 +1,170 @@
...
+		IsRawQuery: true,
+		Fields: []*influxql.Field{
+			{Expr: &influxql.VarRef{Val: FieldName}},
+		},
+		Sources: []influxql.Source{&influxql.Measurement{
+			Name:            MeasurementName,
+			Database:        db,
+			RetentionPolicy: rp,
+		}},
+		Dimensions: []*influxql.Dimension{{Expr: &influxql.Wildcard{}}},
```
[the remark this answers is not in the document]
pauldix: Also used when doing a `GROUP BY *`, which this needs to do.

services/httpd/handler.go
pauldix: Good call on the check. On the divide by `time.Nanosecond`, I did that even though it's a 1 because apparently those underlying things aren't guaranteed. Based on the pedantic conversation in the answers of this post: https://stackoverflow.com/questions/24122821/go-golang-time-now-unixnano-convert-to-milliseconds

prometheus/converters.go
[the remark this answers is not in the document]
e-dard: That's a good point.

services/httpd/handler_test.go
pauldix: ah, was because that was the timestamp that was passed in, but will remove

services/httpd/handler.go
pauldix: Addressed in f0ffe201615dc9d75c9a6fd064faa076eb99eb16

services/httpd/handler.go
pauldix: Addressed in f0ffe201615dc9d75c9a6fd064faa076eb99eb16

services/httpd/handler.go
pauldix: Addressed in f0ffe201615dc9d75c9a6fd064faa076eb99eb16

services/httpd/handler.go
pauldix: Addressed in f0ffe201615dc9d75c9a6fd064faa076eb99eb16

services/httpd/handler.go
pauldix: Addressed in f0ffe201615dc9d75c9a6fd064faa076eb99eb16

services/httpd/handler.go
pauldix: Addressed in f0ffe201615dc9d75c9a6fd064faa076eb99eb16

prometheus/converters.go
```diff
@@ -0,0 +1,170 @@
...
+
+const (
+	// MeasurementName is where all prometheus time series go to
+	MeasurementName = "_"
+
+	// FieldName is the field all prometheus values get written to
+	FieldName = "f64"
+)
+
+var ErrNaNDropped = errors.New("")
```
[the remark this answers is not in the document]
pauldix: Addressed in a2d7024f4aeea92643c84bf6d47e5179224d7286

e-dard:
@pauldix I updated `LICENSE_OF_DEPENDENCIES.md` in https://github.com/influxdata/influxdb/pull/8784/commits/e7a85a1a21d7c7302de93b8561d7fd631236282b

services/httpd/handler.go
jsternberg: Is it appropriate to use `h.httpError` here? That will output the error in a format for InfluxDB. While at the moment we use JSON, it could be any format like CSV. If we are attempting to send an error back that matches what Prometheus is expecting, it probably isn't correct to use `h.httpError`.

services/httpd/handler.go
```diff
@@ -760,6 +777,276 @@ func convertToEpoch(r *query.Result, epoch string) {
...
+				close(closing)
+			}
+		}()
+		opts.AbortCh = done
+	} else {
+		defer close(closing)
+	}
+
+	// Execute query.
+	results := h.QueryExecutor.ExecuteQuery(q, opts, closing)
```
jsternberg: This entire section here should probably be refactored because it is about 95% in common with `serveQuery`, but that doesn't have to be done now. I think #8725 makes it a lot easier.

So no action on this, but I just wanted to get it on the record.

prometheus/converters.go
```diff
@@ -0,0 +1,175 @@
...
+
+// ReadRequestToInfluxQLQuery converts a Prometheus remote read request to an equivalent InfluxQL
+// query that will return the requested data when executed
+func ReadRequestToInfluxQLQuery(req *remote.ReadRequest, db, rp string) (*influxql.Query, error) {
+	if len(req.Queries) != 1 {
+		return nil, errors.New("Prometheus read endpoint currently only supports one query at a time")
+	}
+	promQuery := req.Queries[0]
+
+	q := &influxql.Query{}
```
jsternberg: It would probably be easier to just create this at the bottom when you need it with a literal. The literal will likely allocate less memory to boot since it doesn't use `append`.

stuartcarnie:
@pauldix, really just a nit, but given our only dependency is the `.proto` definition file, we could pull only the [`remote.proto`](https://github.com/prometheus/prometheus/blob/master/storage/remote/remote.proto) file in. This would allow us to generate our own `.pb.go` using gogo, which is more efficient and would avoid pulling entire prometheus repository.

services/httpd/handler.go
pauldix: Yeah, the same is true of the `servePromWrite` method, but I wanted to get this in without modifying any of the existing code paths to make it a very low risk merge.

services/httpd/handler.go
pauldix: The Prometheus remote client doesn't do anything with the response body on error so it's probably not an issue: https://github.com/prometheus/prometheus/blob/master/storage/remote/client.go#L123-L134

pauldix:
@stuartcarnie @e-dard @jsternberg ok, removed the prometheus dependency and just copied the `remote.proto` file. Should be ready for final review.

prometheus/remote.proto
```diff
@@ -0,0 +1,70 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package prometheus; // change package from remote to prometheus
```
jsternberg: Can we put this file in an internal package named remote? I do not want us doing any manual modifications to this file if at all possible. Something like this should work:

    prometheus/
      internal/
        remote/
          remote.proto

And then modify the `go generate` command to output the `remote.pb.go` file to this internal directory. Since it's in the remote folder, the package name will still be `remote` and it will be accessible to the `prometheus` package, but it won't be accessible to anyone else. That will make updating that file easier when we need to because there are no modifications.

prometheus/converters.go
```diff
@@ -0,0 +1,176 @@
...
+	"errors"
+	"fmt"
+	"math"
+	"time"
+
+	"github.com/influxdata/influxdb/influxql"
+	"github.com/influxdata/influxdb/models"
+)
+
+//go:generate protoc -I$GOPATH/src -I. --gogofaster_out=. remote.proto
```
jsternberg: In influxql, we just use `--gogo_out`. Are those different generators and is this adding a new dependency? We likely should avoid using multiple protoc generators if we can. Also, is the `-I` needed? I don't see any included dependencies in the proto file.

prometheus/remote.proto
pauldix: Why the `internal` before? Why not just put it in `remote`?

jsternberg: It means it's not an interface we expose. While not necessary, I think it would be a good habit for us to start getting into.

prometheus/converters.go
stuartcarnie: `gogofaster` is a superset of functionality of `gogo` with additional optimizations in the generated code and less pointer fields. This generally makes the structs easier and more efficient to work with in Go.

prometheus/remote.proto
pauldix: That feels odd to me. Go already has the public vs. private in a package. If we go down this route we'll end up having `internal` in every package in the project, no?

prometheus/converters.go
stuartcarnie: The protobuf wire format is unchanged

jsternberg: Should we be using it instead of `gogo` everywhere and avoid having multiple generators? I'm not opposed to switching generators, I would just like it if we used one. It also doesn't have to be this PR that switches everything to a new generator.

stuartcarnie: Agree and we should move to the new generators. In many cases, we will need to do a significant amount of refactoring but the effort will be worth while for performance and ease of use.

jsternberg: Just to make sure the other part of the comment isn't lost, is the `-I` needed?

pauldix: ok, so use the vanilla generator for now and we can evaluate gogofaster later?

stuartcarnie: Go with `gogo_faster`. We are using it for the new storage APIs and should plan on migrating others over. It does not introduce additional dependencies. The `-I` includes mean that if future versions of the .proto file import other proto files, you won't have to update the `generate` line, so it is safe to leave it in.

pauldix: ok, should be all moved around and good now

PR #8791, opened 2017-09-05
Include the number of scanned cached values in the iterator cost

jsternberg:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -17,16 +17,15 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/influxdata/influxdb/query"
-	"github.com/influxdata/influxdb/tsdb/index/inmem"
-
 	"github.com/influxdata/influxdb/influxql"
 	"github.com/influxdata/influxdb/models"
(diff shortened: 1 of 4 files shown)


tsdb/engine/tsm1/engine.go
```diff
@@ -2204,6 +2202,33 @@ func (e *Engine) IteratorCost(measurement string, opt query.IteratorOptions) (qu
 	return cost, nil
 }
 
+func (e *Engine) seriesCost(seriesKey, field string, tmin, tmax int64) query.IteratorCost {
+	key := SeriesFieldKeyBytes(seriesKey, field)
+	c := e.FileStore.Cost(key, tmin, tmax)
+
+	// Retrieve the range of values within the cache.
+	cacheValues := e.Cache.Values(key)
```
jwilder: I think you can simplify this func to:

`c.CachedValues = len(e.Cache.Values(key).Include(tmin, tmax-1))`

That will filter return the count of values in the cache that are between [tmin, max).

jsternberg: The maximum is inclusive so I'll just put `tmax`, but thanks! Done.

PR #8827, opened 2017-09-14
Implement exponential moving average (exp_moving_average)

stevenroose:
In the same fashion as `moving_average`, this commit implements an
average function `exp_moving_average` that implements an exponentially
weighted moving average (EWMA).

The implementation is based on the implementation of Vivid Cortex Inc.
in [their repository](https://github.com/VividCortex/ewma) which is MIT
licensed.

I'm new to the codebase and have some questions:

1. I looked at the VividCortex implementation to implement this. It's only [11 lines of code](https://github.com/VividCortex/ewma/blob/master/ewma.go#L97-L107) so I don't think we need a copyright notice. If we do, where should I put it?
2. I'm new to the codebase, so I just looked at the implementation of `moving_average`. Please review it thoroughly.
3. I will try to write tests as good as the ones for `moving_average`. I posted the PR before that in the hope to get some review on the code.
4. I can't build/run tests for now. But I'm trying to resolve that. (`query/query_executor.go:15:2: code in directory /home/steven/gocode/src/github.com/uber-go/zap expects import "go.uber.org/zap"`)

Checklist:
- [x] rebased/mergable
- [x] existing tests pass
- [x] new tests provided: I looked for all tests mentioning normal moving averages and made ones for exponential moving averages in the same style. The only ones I miss are the ones from `tests/server_test.go`. They seem pretty advanced and leveraging exact encoded query response json.
- [ ] CHANGELOG.md updated: I think you guys better put this change in the right place in the CHANGELOG.
- [x] sign [CLA](https://influxdata.com/community/cla/) (if not already signed)
- [ ] provide example syntax (where? there is no mention of `moving_average` in this repo outside of code)


e-dard:
@stevenroose Hi. Thanks for this. 

It's the end of my day right now so I don't have time to review the PR, but I can help you with a couple of your other questions:

 Q1. I don't think we need copyright notice, but it would be a good idea to add a comment linking to the source of the approach you've taken for updating the EWMA values. Where we've used other bits of code before we usually link back to the repo and call out the author in a comment.
 Q4. You need to `checkout` revision `fbae0281ffd546fa6d1959fec6075ac5da7fb577` of `github.com/uber-go/zap`. Technically, all your dependencies should be at the revisions specified in out [Godeps](https://github.com/influxdata/influxdb/blob/master/Godeps) file.

You can use [gdm](https://github.com/sparrc/gdm) if you want to automatically change your dependencies to those, but consider that it will modify the dependencies in your `GOPATH`.

e-dard:
I can confirm that @stevenroose has signed the CLA.

query/functions.go
```diff
@@ -566,6 +566,76 @@ func (r *UnsignedMovingAverageReducer) Emit() []FloatPoint {
...
+// exponentialMovingAverageWarmupSamples is the number of samples used to warm up
+// the exponential moving average calculation.
+//
+// For best results, the moving average should not be initialized to the
+// samples it sees immediately.  The book "Production and Operations
+// Analysis" by Steven Nahmias suggests initializing the moving average to
+// the mean of the first 10 samples.  Until the exponential moving average
+// reducer has seen this many samples, it is not "ready" to be queried for the
+// value of the moving average.
+const exponentialMovingAverageWarmupSamples = 10
```
phemmer: This is a very subjective value. For example a recommendation I see all the time, and what I personally use, is to use the period count as the warmup size. Meaning if your taking the EMA over a 20-period span, your warmup is 20.

query/functions.go
```diff
@@ -566,6 +566,76 @@ func (r *UnsignedMovingAverageReducer) Emit() []FloatPoint {
...
+
+// aggregates aggregates a value to the reducer.
+func (r *ExponentialMovingAverageReducer) aggregate(value float64, time int64) {
+	r.count++
+	r.time = time
+	switch {
+	case r.count < exponentialMovingAverageWarmupSamples:
+		r.value += value
+	case r.count == exponentialMovingAverageWarmupSamples:
+		r.value = r.value / float64(exponentialMovingAverageWarmupSamples)
```
phemmer: This is another highly contentious point. How you warm the moving average can have a significant impact on the values at the start. Some people like using a simple moving average (what you've done), and others like jumping straight into the exponential moving average (which is my preference).

It's also possible to start emitting values without waiting for warmup to complete. The values might not be very stable, but depending on your usage, unstable data might be better than no data. Effectively you just start the algorithm with your weight as if the warmup count were 1, and then as the algorithm gets more data, you scale the weight until you reach the real warmup count.

query/functions.go
stevenroose: True. I know it's bad to use magic numbers and perhaps it'd better be configurable by the user, but this value is usually very insignificant when looking at longer time spans. I think it just prevents having a EMA graph that starts with some crazy outliers and then smooths.

stevenroose: Like I said, I looked for similar implementations and took the one that at least mentions some sources on their magic numbers.

query/functions.go
stevenroose: Again, of what importance is the accuracy of the first few values of the graph? In a dataset with highly varying data, some outliers could make the EMA graph potentially very ugly and stretched across the axis (making the rest of the graph scale weird). I think this warmup period only serves to avoid that.

query/functions.go
phemmer: I would argue the opposite, that the value is very significant. If you're taking an EMA with a period in the range of thousands or more, 10 periods is insignificant. It might as well be 0. The first chunk of your results aren't going to be smooth at all.

stevenroose: @phemmer  EMA doesn't really have a "period". All samples incorporate all past ones; there is just a decay factor. So let's say a sample that goes like `[1000, 0, 1000, 0, 1000, 0, ...]`, will have a very ugly first 5-10 samples, whereas with a small warmup period, it startss with 500 and will go up and down slightly when it keeps iterating.

Or f.e. a sample that is quite regular but has an outlier like every 100 samples. If you're unlucky and this outlier is in your first few samples (`[3, 6, 100, 2, 2, 1, 4, ...]`). Without the warmup period, that will look ugly. With the warmup period, the 100 will still put the EMA too high in the beginning, but at least it won't jump from ~5 to ~100 and start falling by then.

I'm no expert in the maths behind this, don't get me wrong. I just understood that the value is not that important because EMA's are mainly used for longer series; but that it's just a convenience to prevent ugly data.

If you have a concrete proposal (like making it 0) that would make this change more acceptable to the maintainers, I'd be willing to change the implementation.

rbetts:
Milestone 1.4 is being closed to new features at this point in time.

stevenroose:
@phemmer Do you have a concrete idea on how to proceed regarding to the warmup samples constant?

phemmer:
If you're asking about how the difference of opinion on what the value should be, I was thinking something along the lines of letting the user control it with a function interface such as: `exp_moving_average(<field>, <sample_count>[, <warmup_sample_count>, [<warmup_type>]])`.
While I have strong opinions about what the warmup_sample_count and warmup_type should be, I'd be less concerned about the default values if they were customizable.

jsternberg:
Closing this in favor of #9260.

stevenroose:
Perfectly happy about that! :)

PR #8835, opened 2017-09-15
Add uint support into the write protocol

jsternberg:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated
- [ ] Provide example syntax

### DIFF
--- a/models/points.go
+++ b/models/points.go
@@ -224,6 +224,9 @@ const (
 	// the number of characters for the smallest possible int64 (-9223372036854775808)
 	minInt64Digits = 20
 
+	// the number of characters for the largest possible uint64 (18446744073709551615)
+	maxUint64Digits = 20
+
 	// the number of characters required for the largest float64 before a range check
 	// would occur during parsing
 	maxFloat64Digits = 25
@@ -827,7 +830,7 @@ func isNumeric(b byte) bool {
 // error if a invalid number is scanned.
 func scanNumber(buf []byte, i int) (int, error) {
 	start := i
-	var isInt bool
+	var isInt, isUnsigned bool
 
 	// Is negative number?
 	if i < len(buf) && buf[i] == '-' {
@@ -853,8 +856,12 @@ func scanNumber(buf []byte, i int) (int, error) {
 			break
 		}
 
-		if buf[i] == 'i' && i > start && !isInt {
+		if buf[i] == 'i' && i > start && !(isInt || isUnsigned) {
 			isInt = true
+			i++
+			continue
+		} else if buf[i] == 'u' && i > start && !(isInt || isUnsigned) {
+			isUnsigned = true
 			i++
 			continue
 		}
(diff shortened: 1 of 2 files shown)


models/points.go
```diff
@@ -926,6 +933,22 @@ func scanNumber(buf []byte, i int) (int, error) {
 				return i, fmt.Errorf("unable to parse integer %s: %s", buf[start:i-1], err)
 			}
 		}
+	} else if isUnsigned {
+		// Make sure the last char is a 'u' for unsigned
+		if buf[i-1] != 'u' {
+			return i, ErrInvalidNumber
+		}
+		// Make sure the first char is not a '-' for unsigned
```
rbetts: This check is redundant with the initial check at line 833?

models/points_test.go
```diff
@@ -2173,7 +2175,7 @@ m v=42i
 m v="string"
 m v=true
 m v="string\"with\"escapes"
-m v=42i,f=42,g=42.314
```
rbetts: I might be reading the test incompletely - but need to be sure we have tests for a negative unsigned int (error path), for max value, and for something greater than the max value.

models/points.go
jsternberg: I don't believe so. The check at line 833 prevents a person from writing `m=-`. This checks to make sure a negative wasn't used while looking at an unsigned integer.

models/points_test.go
jsternberg: I'll add those.

jsternberg:
I wasn't planning on having this merged until the query engine stuff was done, but I didn't think about the other applications. Should I build it as a configuration option or some compilation switch? I can also just block all incoming writes of unsigned with no compilation switch and no configuration option.

jsternberg:
It also may be difficult to have a switch that enables this. The section of code that does the parsing doesn't really have access to any configuration and catching it later than that would probably add some undesired overhead.

phemmer:
also is there any planned way to be able to change the type of an existing field?
I'm just imagining the scenario in telegraf where the various input plugins are storing their types as `uint64`, and the telegraf marshaller is sending them as signed integers. But once this PR is merged, and the telegraf marshaller is changed so that `uint64` is actually sent as an unsigned integer, lots of writes will start failing because of conflicting field type.

rbetts:
@phemmer The ability to retroactively rewrite field types is not in scope. We are thinking through issues like the one you've raised but our immediate goal is to get to a functional uint64 field type.

rbetts:
@jsternberg I would avoid a configuration parameter - as this won't be long lived toggle once everything is done.  A command line flag or environment variable might be preferable. I would like to get these bits landed in reasonable chunks on master.

jsternberg:
@rbetts done.

```
$ go install ./cmd/influxd
$ influx
> create database db0
> use db0
> insert cpu value=1u
ERR: {"error":"unable to parse 'cpu value=1u': invalid number"}

> exit
$ go install -tags uint ./cmd/influxd
$ influx
> use db0
> insert cpu value=1u
> exit
```

jsternberg:
I need to fix the tests since they fail since uint support is turned off. I'm also adding some additional parser tests for uint64 on top of the existing ones.

Issue #8789, opened 2017-09-05
cli can't handle database names with space, http api can


knan-nrk:
### Bug report

__System info:__ 

Connected to http://localhost:8086 version 1.3.4
InfluxDB shell version: 1.3.4
Ubuntu 16.04

__Steps to reproduce:__

1. create database "with space"
2. use "with space"

__Expected behavior:__ 

database "with space" created, ability to use it in cli queries

__Actual behavior:__ 

database "with space" created
> use "with space"
Could not parse database name from "use \"with space\"".

__Additional info:__ 

HTTP api has no problems accepting data for the database, seems to be a pure cli problem. Can create and drop the database "with space" via the cli, just not use/inspect it.

joelegasse:
Fixed by #8845

PR #8845, opened 2017-09-18
Fix CLI to allow quoted database names

andrewhare:
Change the CLI to support quoted database names in `use` statements.
This also allows for all database names to be specified, including names
that contain spaces.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


cmd/influx/cli/cli.go
```diff
@@ -396,7 +396,7 @@ func (c *CommandLine) clear(cmd string) {
 }
 
 func (c *CommandLine) use(cmd string) {
-	args := strings.Split(strings.TrimSuffix(strings.TrimSpace(cmd), ";"), " ")
+	args := strings.SplitAfterN(strings.TrimSuffix(strings.TrimSpace(cmd), ";"), " ", 2)
```
joelegasse: This should probably be a `SplitN`, not a `SplitAfterN`. It doesn't really change anything in this case (N=2, and only care about the second item), but the separator isn't important after the split, which would be the only reason to use `SplitAfter`, since it keeps the returned strings.

e-dard:
This should fix #8789.

e-dard:
@andrewhare can you rebase or resolve the `CHANGELOG` via the Github editor?

andrewhare:
@e-dard Rebased

e-dard:
@andrewhare merge away.

PR #8847, opened 2017-09-19
Fix various engine races

### LINKED ISSUE #8842: Races caused by unprotected engine access
There are a number of unprotected accesses to the `Engine`, via the `Store`, which if triggered can lead to the server panicking.

For example:

 - Interaction between `SHOW TAG VALUES` and `DROP DATABASE`.
 - Interaction between opening a shard and `DROP DATABASE`.
 - Interaction between `SHOW FIELD KEYS` and `DROP DATABASE`.
 - Interaction between the monitoring service and `DROP DATABASE`.

e-dard:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

Fixes #8842.

This PR fixes a number of races on a shard's Engine. There are a few places where we have unprotected access to an Engine. A shard's Engine is set to `nil` when we close the shard, so anything that reads from the Engine variable without being under an `RLock` is looking at a `nil` pointer in certain circumstances.

The main situations where this would show itself is when deleting a database whilst other operations such as certain meta queries or series deletes are in progress.

### DIFF
--- a/tsdb/shard.go
+++ b/tsdb/shard.go
@@ -735,15 +735,38 @@ func (s *Shard) validateSeriesAndFields(points []models.Point) ([]models.Point, 
 // MeasurementNamesByExpr returns names of measurements matching the condition.
 // If cond is nil then all measurement names are returned.
 func (s *Shard) MeasurementNamesByExpr(cond influxql.Expr) ([][]byte, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	if err := s.ready(); err != nil {
+		return nil, err
+	}
+
 	return s.engine.MeasurementNamesByExpr(cond)
 }
 
 // MeasurementFields returns fields for a measurement.
+// TODO(edd): This method is currently only being called from tests; do we
+// really need it?
 func (s *Shard) MeasurementFields(name []byte) *MeasurementFields {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	if err := s.ready(); err != nil {
+		return nil
+	}
+
 	return s.engine.MeasurementFields(name)
 }
 
+// MeasurementExists returns true if the shard contains name.
+// TODO(edd): This method is currently only being called from tests; do we
+// really need it?
 func (s *Shard) MeasurementExists(name []byte) (bool, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	if err := s.ready(); err != nil {
+		return false, nil
+	}
+
 	return s.engine.MeasurementExists(name)
 }
 
--- a/tsdb/shard.go
+++ b/tsdb/shard.go
@@ -118,9 +118,9 @@ type Shard struct {
 
 	options EngineOptions
 
-	mu     sync.RWMutex
-	engine Engine
-	index  Index
+	mu      sync.RWMutex
+	_engine Engine
+	index   Index
 
 	closing chan struct{}
 	enabled bool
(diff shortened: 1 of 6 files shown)


e-dard:
@rbetts 🔎  👍

e-dard:
@benbjohnson @jwilder  please see further commits de8a7cb and 6633abc.

Couple of points about the further changes:

  - `SeriesSketches` and `MeasurementsSketches` used to return `nil` when they encountered a `nil` engine. This would be the case if the engine had been closed. While this was racy (and is now fixed) I wasn't sure about returning an error here so I have kept it as a `nil` error return when the engine is closed.
 - A couple of the tests in `shard_test` seem to have had the invariant that when the shard has been closed, the methods `ForEachMeasurementTagKey` and `TagKeyCardinality` should not return errors, but should return zero values. This seems strange to me given most of the other methods on the shard will return an error when the shard (engine) has been closed. I have switched the assertions on those tests.

Let me know if you think any of this will cause a problem.

tsdb/shard.go
```diff
@@ -1081,13 +1153,19 @@ func (a Shards) Less(i, j int) bool { return a[i].id < a[j].id }
...
+// provided regex, for all the shards.
 func (a Shards) MeasurementsByRegex(re *regexp.Regexp) []string {
 	m := make(map[string]struct{})
 	for _, sh := range a {
-		names := sh.MeasurementsByRegex(re)
-		for _, name := range names {
-			m[name] = struct{}{}
+		sh.mu.RLock()
+		if err := sh.ready(); err == nil {
+			names := sh.measurementsByRegex(re)
```
jwilder: Can you have this call `MeasurementsByRegex` instead of accessing the `sh.mu` directly outside of `Shard`?

tsdb/shard.go
```diff
@@ -997,6 +1048,17 @@ func (s *Shard) ExpandSources(sources influxql.Sources) (influxql.Sources, error
 	return expanded, nil
 }
 
+// Backup backs up the shard by creating a tar archive of all TSM files that
+// have been modified since the provided time. See Engine.Backup for more details.
+func (s *Shard) Backup(w io.Writer, basePath string, since time.Time) error {
+	s.mu.RLock()
```
jwilder: This is going to block cache snapshotting now, when it does not need to.  Can you hold onto a reference for the engine so that that you don't need to hold a read lock while the backup is running?

tsdb/shard.go
```diff
@@ -761,17 +821,18 @@ func (s *Shard) WriteTo(w io.Writer) (int64, error) {
 
 // CreateIterator returns an iterator for the data in the shard.
 func (s *Shard) CreateIterator(measurement string, opt query.IteratorOptions) (query.Iterator, error) {
+	s.mu.RLock()
```
jwilder: Can you grab a reference to the engine and use that instead of holding a read lock for the whole func?  The locks were lower to avoid lock contention when snapshotting the cache.

tsdb/shard.go
```diff
@@ -1306,15 +1334,12 @@ func (a Shards) ExpandSources(sources influxql.Sources) (influxql.Sources, error
...
 	for _, sh := range a {
 		sh.mu.RLock()
-		if err := sh.ready(); err != nil {
-			return nil, err
-		}
 		expanded, err := sh.expandSources(sources)
-		sh.mu.RUnlock()
-
 		if err != nil {
+			sh.mu.RUnlock()
```
jwilder: You should be able to this once before the `err != nil` check.

PR #8850, opened 2017-09-19
WIP: TSI Series File

benbjohnson:
@e-dard @jwilder This is still a rough cut of the series file. It's working but the series IDs are currently `uint32` offsets so it has an upper bound of a few 10s of millions of series. I'll fix that but I wanted to get this PR up so you can take a look in case you're interested. Also, I'm still investigating the performance. It's not where I want it to be.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

- [ ] Provide example syntax


tsdb/index/tsi1/series_file.go
```diff
@@ -0,0 +1,446 @@
...
+		offset += uint32(len(key))
+	}
+
+	f.hashMap = m
+
+	return nil
+}
+
+// CreateSeriesIfNotExists creates series if it doesn't exist. Returns the offset of the series.
+func (f *SeriesFile) CreateSeriesIfNotExists(name []byte, tags models.Tags, buf []byte) (offset uint32, err error) {
```
jwilder: Is the offset returned as the series ID stable?  Will it change each time the hash is re-indexed?  If so, then I won't be able to use it in TSM files as I need a stable ID.

jwilder: Also, I need the field type and field name to be part of the key as well.

tsdb/shard.go
```diff
@@ -524,16 +524,6 @@ func (s *Shard) SeriesN() int64 {
 	return s.engine.SeriesN()
 }
 
-// SeriesSketches returns the series sketches for the shard.
-func (s *Shard) SeriesSketches() (estimator.Sketch, estimator.Sketch, error) {
```
jwilder: Why is this removed?

tsdb/index/tsi1/series_file.go
```diff
@@ -0,0 +1,446 @@
...
+
+	// Memory map file data.
+	data, err := mmap.Map(f.path, f.MaxSize)
+	if err != nil {
+		return err
+	}
+	f.data = data
+
+	// Index all series.
+	if err := f.reindex(); err != nil {
```
jwilder: This looks like it's rebuilding the hash index each time the file is open?  This needs to be an on-disk hash index that we can just open and start using or startup times will be slow.

benbjohnson: I still need to add persistence for the hash map. I went with a simpler approach for now so I could test the ingest performance.

tsdb/index/tsi1/series_file.go
benbjohnson: The offset is from the series file which is append-only so it will be stable. I'll add field name/type as well.

tsdb/shard.go
benbjohnson: Do we still need series sketches if we aren't merging multiple files? It seems like we'd be able to have an accurate count at all times with a single series file.

tsdb/index/tsi1/series_file.go
jwilder: Makes sense.

tsdb/index/tsi1/series_file.go
jwilder: Cool

jwilder:
This was incorporated in another PR.

PR #8866, opened 2017-09-22
Implicitly decide on the lower limit for fill queries when none is present

### LINKED ISSUE #8690: Query with time < now() produces negative timestamps and null data
InfluxDB 1.3.2 on Debian 9.

This query worked as I expect it to (only return existing data and fill between first and last points) in 1.2.x, but now it starts with completely odd negative timestamp (like -9223372036) which breaks my stuff entirely.

Was this an expected change, as I don't see anything particular in the changelog.

The full query is: `SELECT time, MAX(value) FROM concurrent_players_daily WHERE appid = '440' AND time < now() - 1d GROUP BY time(1d) fill(linear)`

This appears to directly related to `GROUP BY` clause as removing `fill(linear)` does not change anything.

jsternberg:
This allows the query:

    SELECT mean(value) FROM cpu GROUP BY time(1d)

To function in some way that makes sense. The upper limit is implicitly
the `now()` starting time and the lower limit will be whichever interval
the lowest point falls into.

When no lower bound is specified and `max-select-buckets` is specified,
the query will only consider points that would satisfy
`max-select-buckets`. So if you have one point written in 1970, have
another point within the last minute, and then do the above query with
`max-select-buckets` being equal to 10, the older point from 1970 will
not be considered.

Fixes #8690.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): issue filed or pull request submitted \<[influxdata/docs.influxdata.com#1289](https://github.com/influxdata/docs.influxdata.com/issues/1289)>


jsternberg:
The behavior here may conflict with `max-select-buckets`.

jsternberg:
@benbjohnson I think it would block this query for having too many buckets. The validation occurs before the iterator is created and would calculate that based off of the minimum time.

I'm not sure how to deal with `max-select-buckets` for this scenario. Any ideas?

jsternberg:
@benbjohnson I think I came up with a solution for the `max-select-buckets` problem. Can you take a look and see if it makes sense? I edited the description of the PR with an explanation of what happens with `max-select-buckets`.

PR #8878, opened 2017-09-26
Do not report an error when dropping a CQ on a non-existent DB/RP

andrewhare:
This makes the behavior similar to other places in the DB where we
don't return an error when we try to drop an object from a non-
existent database.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


jsternberg:
@andrewhare can you rebase this, remove the changelog update (we're now doing that automatically), and change the line David pointed out? Thanks.

PR #8884, opened 2017-09-27
Implement uint64 logic into eval

jsternberg:
Any operation between an int64 and uint64 results in the type becoming a
uint64. This is because the most common will usually be an unsigned cursor
being modified by an int literal. Even in cases where we were to add an
unsigned literal to an integer iterator, that would just result in it being
recast back and overflow.

Includes a minor fix for dividing an integer by zero (should have resulted in
an integer, but resulted in a float) and I forgot to include UnsignedLiteral in
the logic for EvalType.

- [x] Rebased/mergable
- [x] Tests pass


influxql/ast.go
```diff
@@ -3988,7 +3988,7 @@ func evalBinaryExpr(expr *BinaryExpr, m map[string]interface{}) interface{} {
 				if !ok {
 					return nil
 				} else if rhs == 0 {
-					return float64(0)
+					return int64(0)
```
jsternberg: So this is a lot more confusing than I thought it would be. We appear to have an inconsistency. If you do division of two integers in a condition, you get integer division. I verified that the division below is between two integers. But, if you do division between two integers in a field (binary math), you get float division and will get a float (which EvalType doesn't return correctly and so would screw up subqueries).

I think we need to straighten out this behavior and backport that specific section to the 1.3 series.

CC @rbetts

PR #8885, opened 2017-09-27
Fix race on Cache entry

e-dard:
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

Fixes a race on the `Cache` `entry`.

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -1331,7 +1331,7 @@ func (e *Engine) onFileStoreReplace(newFiles []TSMFile) {
 
 	// load metadata from the Cache
 	e.Cache.ApplyEntryFn(func(key []byte, entry *entry) error {
-		fieldType, err := entry.values.InfluxQLType()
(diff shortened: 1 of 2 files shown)


tsdb/engine/tsm1/engine.go
```diff
@@ -1331,7 +1331,9 @@ func (e *Engine) onFileStoreReplace(newFiles []TSMFile) {
 
 	// load metadata from the Cache
 	e.Cache.ApplyEntryFn(func(key []byte, entry *entry) error {
+		entry.mu.RLock()
```
jwilder: Instead of touching the entries lock from the engine, can you change this to have a `entry.InfluxQLType()` instead that keeps the locking private to the `entry`?

e-dard: Sure.

e-dard:
@jwilder updated.

e-dard:
Backport https://github.com/influxdata/influxdb/pull/8887

PR #8886, opened 2017-09-27
Improved compaction scheduling

jwilder:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

This fixes some regressions and new issues due to recent changes from #8872 and #8856.

* #8872 fixed some long process stalls due to major page faults, but also causes a significant reduction in throughput for lower cardinalities.  This reverts of `O_SYNC` flag for the wal for now.
* #8856 increased the throughput for snapshot compactions which keeps memory usage lower at higher cardinalities, but increases the number of level 1 tsm files more quickly.  This exposes a issue with the current compaction planning process where the planning for a level would get block on a slower compaction within the level.  For example, a level 3 compaction might plan for 4 groups to be compacted concurrently.  If 3 completed quickly, the level 3 planner would wait until the last compaction completed.  Since lower levels are getting written more quickly, a large backlog could accumulate.  This changes the planning to better handle these situations and make use of all the cores allocated for compactions (limited by `max-concurrent-compactions`).

There are also two bug fixes where the temp TSM index file could be left around when running out of disk space as well as fixing the cache mem bytes stat to include the size of the series key.


tsdb/engine/tsm1/writer.go
```diff
@@ -463,6 +467,18 @@ func (d *directIndex) Close() error {
 	return os.Remove(d.fd.Name())
 }
 
+// Remove removes the index from any tempory storage
+func (d *directIndex) Remove() error {
+	if d.fd == nil {
+		return nil
+	}
+
+	if err := d.fd.Close(); err != nil {
```
e-dard: So if there is a problem closing the file we're not going to clean up, but we're not going to return an error to the caller either? This scenario seems unusual enough that we don't want to ignore it or not log it.

tsdb/engine/tsm1/writer.go
```diff
@@ -652,6 +668,19 @@ func (t *tsmWriter) Close() error {
...
 }
 
+// Remove removes any temporary storage used by the writer.
+func (t *tsmWriter) Remove() error {
+	if err := t.index.Remove(); err != nil {
+		return err
+	}
+
+	if f, ok := t.wrapped.(*os.File); ok {
+		f.Close()
```
e-dard: In this case we're going to ignore a `Close` error (fine) but then still try and remove the file. This seems at odds with how the `directIndex` is handled.

tsdb/engine/tsm1/writer.go
jwilder: That should be returning the `err` here.  Will fix.

tsdb/engine/tsm1/writer.go
jwilder: I'll fix this up too.

PR #8891, opened 2017-09-27
Allow human-readable byte sizes in config

andrewhare:
Update support in the `toml` package for parsing human-readble byte sizes.
Supported size suffixes are "k" or "K" for kibibytes, "m" or "M" for
mebibytes, and "g" or "G" for gibibytes. If a size suffix isn't specified
then bytes are assumed.

In the config, `cache-max-memory-size` and `cache-snapshot-memory-size` are
now typed as `toml.Size` and support the new syntax.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

- [x] Provide example syntax

### DIFF
--- a/etc/config.sample.toml
+++ b/etc/config.sample.toml
@@ -71,10 +71,12 @@
 
   # CacheMaxMemorySize is the maximum size a shard's cache can
   # reach before it starts rejecting writes.
+  # Values without a size suffix (e.g. 30k, 20m, 10g) are in bytes.
   # cache-max-memory-size = 1048576000
 
   # CacheSnapshotMemorySize is the size at which the engine will
   # snapshot the cache and write it to a TSM file, freeing up memory
+  # Values without a size suffix (e.g. 30k, 20m, 10g) are in bytes.
   # cache-snapshot-memory-size = 26214400
 
   # CacheSnapshotWriteColdDuration is the length of time at
--- a/etc/config.sample.toml
+++ b/etc/config.sample.toml
@@ -71,13 +71,15 @@
 
   # CacheMaxMemorySize is the maximum size a shard's cache can
   # reach before it starts rejecting writes.
-  # Values without a size suffix (e.g. 30k, 20m, 10g) are in bytes.
-  # cache-max-memory-size = 1048576000
+  # Valid size suffixes are k, m, or g (case insensitive, 1024 = 1k).
+  # Vaues without a size suffix are in bytes.
+  # cache-max-memory-size = "1g"
 
   # CacheSnapshotMemorySize is the size at which the engine will
   # snapshot the cache and write it to a TSM file, freeing up memory
-  # Values without a size suffix (e.g. 30k, 20m, 10g) are in bytes.
-  # cache-snapshot-memory-size = 26214400
+  # Valid size suffixes are k, m, or g (case insensitive, 1024 = 1k).
+  # Values without a size suffix are in bytes.
+  # cache-snapshot-memory-size = "25m"
 
   # CacheSnapshotWriteColdDuration is the length of time at
   # which the engine will snapshot the cache and write it to
--- a/toml/toml.go
+++ b/toml/toml.go
@@ -3,13 +3,11 @@ package toml // import "github.com/influxdata/influxdb/toml"
 
 import (
 	"fmt"
+	"math"
 	"strconv"
 	"time"
 	"unicode"
 )
-
-// maxInt is the largest integer representable by a word (architecture dependent).
-const maxInt = int64(^uint(0) >> 1)
 
 // Duration is a TOML wrapper type for time.Duration.
 type Duration time.Duration
@@ -55,7 +53,7 @@ func (s *Size) UnmarshalText(text []byte) error {
 
 	// The multiplier defaults to 1 in case the size has
 	// no suffix (and is then just raw bytes)
-	mult := int64(1)
+	mult := uint64(1)
 
 	// Preserve the original text for error messages
 	sizeText := text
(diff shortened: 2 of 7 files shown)


toml/toml.go
```diff
@@ -42,31 +43,54 @@ func (d Duration) MarshalText() (text []byte, err error) {
...
-	// Parse unit of measure ("m", "g", etc).
-	switch suffix := text[len(text)-1]; suffix {
-	case 'm':
-		size *= 1 << 20 // MB
-	case 'g':
-		size *= 1 << 30 // GB
-	default:
-		return fmt.Errorf("unknown size suffix: %c", suffix)
+	if maxInt/mult < size {
+		return fmt.Errorf("size would overflow the max size (%d) of an int: %s", maxInt, string(text))
```
jsternberg: This should say `a uint` because now it's an unsigned integer. I don't think we need to specify it is a uint64. Can you also use `math.MaxUint64` for the size check? It would make it so `maxInt` isn't needed anymore.

toml/toml.go
```diff
@@ -42,31 +43,54 @@ func (d Duration) MarshalText() (text []byte, err error) {
...
 	}
 
+	size *= mult
+
 	// Check for overflow.
 	if size > maxInt {
-		return fmt.Errorf("size %d cannot be represented by an int", size)
+		return fmt.Errorf("size %d is too large", size)
```
jsternberg: I don't think this check is possible. I don't think the previous code worked and I don't see any tests that verified this if statement ever worked.

etc/config.sample.toml
```diff
@@ -71,10 +71,12 @@
 
   # CacheMaxMemorySize is the maximum size a shard's cache can
   # reach before it starts rejecting writes.
+  # Values without a size suffix (e.g. 30k, 20m, 10g) are in bytes.
```
jsternberg: Can you update this to say: `Valid size suffixes are k, m, or g (case insensitive). Values without a size suffix are in bytes.`

jsternberg:
@andrewhare @mark-rushakoff I implemented the feedback to this so we can get it in for 1.4. Can you two check to make sure that there isn't anything I missed?

etc/config.sample.toml
```diff
@@ -71,10 +71,14 @@
 
   # CacheMaxMemorySize is the maximum size a shard's cache can
   # reach before it starts rejecting writes.
+  # Valid size suffixes are k, m, or g (case insensitive).
+  # Values without a size suffix (e.g. 30k, 20m, 10g) are in bytes.
```
mark-rushakoff: "without a size suffix" - should it be 30, 20, 10 in that case? Or maybe just say that all values represent bytes.

To reduce any other confusion, it would help if this comment clarified k means 1000 and not 1024.

toml/toml.go
```diff
@@ -42,32 +41,50 @@ func (d Duration) MarshalText() (text []byte, err error) {
...
+	if !unicode.IsDigit(rune(suffix)) {
+		switch suffix {
+		case 'k', 'K':
+			mult = 1 << 10 // KiB
+		case 'm', 'M':
+			mult = 1 << 20 // MiB
+		case 'g', 'G':
+			mult = 1 << 30 // GiB
+		default:
+			return fmt.Errorf("unknown size suffix: %c", suffix)
```
mark-rushakoff: It would be helpful if this error mentioned that "valid suffixes are k, m, g".

tsdb/config.go
```diff
@@ -71,8 +71,8 @@ type Config struct {
 	QueryLogEnabled bool `toml:"query-log-enabled"`
 
 	// Compaction options for tsm1 (descriptions above with defaults)
-	CacheMaxMemorySize             uint64        `toml:"cache-max-memory-size"`
-	CacheSnapshotMemorySize        uint64        `toml:"cache-snapshot-memory-size"`
+	CacheMaxMemorySize             toml.Size     `toml:"cache-max-memory-size"`
```
mark-rushakoff: I'd like to see an addition to `tsdb/config_test.go` that sets one of these new toml.Size values to something with a suffix.

etc/config.sample.toml
jsternberg: I'm not sure that's the case. Since the sizes are byte shifts, then `1 << 10` is 1024. I'll see if I can clarify this comment though since what it's trying to say is that 30k, 20m, and 10g have size suffixes.

mark-rushakoff: You're right, it's 1024 and not 1000. Should still be clarified for users.

toml/toml.go
jsternberg: Done.

tsdb/config.go
jsternberg: Done.

etc/config.sample.toml
```diff
@@ -71,11 +71,15 @@
 
   # CacheMaxMemorySize is the maximum size a shard's cache can
   # reach before it starts rejecting writes.
-  # cache-max-memory-size = 1048576000
+  # Valid size suffixes are k, m, or g (case insensitive).
+  # Vaues without a size suffix are in bytes.
+  # cache-max-memory-size = "1g"
```
jsternberg: Just a note for those who might check this, the sample configuration file has the wrong value. The code has 1 GB as the default and 1 GB in bytes is 1073741824 bytes, not 1048576000.

mark-rushakoff: I see, the old sample config value was 1000m but the `const DefaultCacheMaxMemorySize` is actually 1g (i.e. 1024m). Fixing the sample config to 1g makes sense here.

etc/config.sample.toml
jsternberg: I updated the configuration file to include a note that 1024 = 1k. That should be good enough for someone to extrapolate that these are 1024 and not 1000.

jsternberg:
The circle tests continue to fail for some reason, but the appveyor ones succeeded and the ones that failed, the race tests, seem to work on my computer and have little to nothing to do with this PR. I'm going to merge this since I don't think the unit tests are actual failures.

PR #8896, opened 2017-09-28
Fix a minor memory leak when batching points for some services.

### LINKED ISSUE #8895: TSDB batcher doesn't free unused timers causing a potential memory leak.
### Bug report

__System info:__ Influxd 1.3.5

__Steps to reproduce:__

1. configure influxdb to use small batch sizes for collectd/graphite/opentsdb/udp
2. send large quantities of points through the service

__Additional info:__ Everytime a batch is emitted in the batcher the batch is cleared but the old timer is replaced and the previous time will end up creating a time.Time and attempt to send it on the channel that nothing is looking at.

lyondhill:
fixes #8895

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/tsdb/batcher.go
+++ b/tsdb/batcher.go
@@ -55,14 +55,23 @@ func (b *PointBatcher) Start() {
 		return
 	}
 
-	var timer *time.Timer
+	// initialize the timer variable
+	timer := &time.NewTimer(b.duration)
+	timer.Stop()
 	var batch []models.Point
-	var timerCh <-chan time.Time
 
 	emit := func() {
+
+		timer.Stop()
+
+		// Nothing batched?
+		if len(batch) == 0 {
+			return
+		}
 		b.out <- batch
 		atomic.AddUint64(&b.stats.BatchTotal, 1)
 		batch = nil
+
 	}
 
 	b.wg = &sync.WaitGroup{}
@@ -75,16 +84,15 @@ func (b *PointBatcher) Start() {
 			case <-b.stop:
 				if len(batch) > 0 {
 					emit()
-					timerCh = nil
 				}
 				return
 			case p := <-b.in:
 				atomic.AddUint64(&b.stats.PointTotal, 1)
 				if batch == nil {
 					batch = make([]models.Point, 0, b.size)
+
 					if b.duration > 0 {
-						timer = time.NewTimer(b.duration)
-						timerCh = timer.C
+						timer.Reset(b.duration)
 					}
 				}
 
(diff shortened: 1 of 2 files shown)


tsdb/batcher.go
```diff
@@ -55,14 +55,23 @@ func (b *PointBatcher) Start() {
...
-	var timer *time.Timer
+	// initialize the timer variable
+	timer := &time.NewTimer(b.duration)
+	timer.Stop()
 	var batch []models.Point
-	var timerCh <-chan time.Time
 
 	emit := func() {
+
+		timer.Stop()
```
joelegasse: We are reusing a timer with `Reset`, so we need to handle the case where we've reached this point other than the `<-timer.C` case, but the timer fired concurrently with this function call. That means that stop wasn't effective, and the `<-timer.C` case will fire immediately once we loop around again. This won't matter much since `batch` will be empty, and it will just return after draining the channel. But we should probably avoid the extra call if we can:

```go
timer.Stop()
select {
    case <-timer.C:
    default:
}
```

tsdb/batcher.go
```diff
@@ -55,14 +55,23 @@ func (b *PointBatcher) Start() {
 		return
 	}
 
-	var timer *time.Timer
+	// initialize the timer variable
+	timer := &time.NewTimer(b.duration)
```
joelegasse: This should just be `timer := time.NewTimer(time.Hour)`, otherwise `timer`'s type is going to be `**time.Timer`. Also, picking an arbitrarily large value is fine, just in case there isn't a duration set, and we're stopping it on the next line, anyway.

joelegasse: This doesn't even compile... :-)

tsdb/batcher.go
```diff
@@ -75,33 +88,30 @@ func (b *PointBatcher) Start() {
 			case <-b.stop:
 				if len(batch) > 0 {
```
joelegasse: While we're here... this check isn't needed, as it's in `emit()`, and calling `emit()` will force the timer to stop.

tsdb/batcher.go
```diff
@@ -75,33 +88,30 @@ func (b *PointBatcher) Start() {
...
 
 				batch = append(batch, p)
 				if len(batch) >= b.size { // 0 means send immediately.
 					atomic.AddUint64(&b.stats.SizeTotal, 1)
 					emit()
-					timerCh = nil
 				}
 
 			case <-b.flush:
 				if len(batch) > 0 {
```
joelegasse: Same here, this case can just call `emit()` without the check

PR #8840, opened 2017-09-17
Convert duration to string without dot and zeros

liketic:
Closes #8813 . 
Format `Duration` to `[w][d][h][m][s][ms][u][ns]`. Another resolution is `parser.FormatDuration`.

PR #8901, opened 2017-10-02
Revert "Convert duration to string without dot and zeros"

jsternberg:
Reverts influxdata/influxdb#8840

The new output makes it so the duration is not parseable with `time.ParseDuration` from the Go standard library. While this isn't necessarily a breaking change, it's in a gray enough area that the change is likely not worth it for a minor release.


rbetts:
LGTM

jsternberg:
The previous build on CircleCI passed and the failure is just a timeout. Since this is a low risk revert, I'm going to just merge it.

PR #8903, opened 2017-10-02
Use `command` instead of `which`

mark-rushakoff:
Some systems don't have `which` installed.

Also fix some whitespace issues.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


mark-rushakoff:
Added Needs Backporting label in case we have a 1.3.7 release.

PR #8906, opened 2017-10-03
Series Map

benbjohnson:
This includes an RHH implementation of an on-disk series map. Currently this rebuilds the entire map after going past a threshold for a given number of series.

This currently blocks inserts (which is obviously bad) and I started doing a background compaction but realized that it might make more sense to create incremental maps instead.

Also, there was a bug in the partition loading that was causing partitions not to load. I'll add a comment inline.

### DIFF
--- a/tsdb/index/tsi1/index.go
+++ b/tsdb/index/tsi1/index.go
@@ -147,27 +147,28 @@ func (i *Index) Open() error {
 	}
 
 	// Open all the Partitions in parallel.
+	partitionN := len(i.partitions)
 	n := i.availableThreads()
 
 	// Store results.
-	errC := make(chan error, n)
+	errC := make(chan error, partitionN)
 
 	// Run fn on each partition using a fixed number of goroutines.
 	var pidx uint32 // Index of maximum Partition being worked on.
 	for k := 0; k < n; k++ {
-		go func() {
+		go func(k int) {
 			for {
 				idx := int(atomic.AddUint32(&pidx, 1) - 1) // Get next partition to work on.
-				if idx >= len(i.partitions) {
+				if idx >= partitionN {
 					return // No more work.
 				}
 				errC <- i.partitions[idx].Open()
 			}
-		}()
+		}(k)
 	}
 
 	// Check for error
-	for i := 0; i < cap(errC); i++ {
+	for i := 0; i < partitionN; i++ {
 		if err := <-errC; err != nil {
 			return err
 		}
@@ -175,7 +176,7 @@ func (i *Index) Open() error {
 
 	// Mark opened.
 	i.opened = true
-	i.logger.Info(fmt.Sprintf("index opened with %d partitions", len(i.partitions)))
+	i.logger.Info(fmt.Sprintf("index opened with %d partitions", partitionN))
 	return nil
 }
 
--- a/tsdb/index/tsi1/series_file.go
+++ b/tsdb/index/tsi1/series_file.go
@@ -35,6 +35,10 @@ const DefaultMaxSeriesFileSize = 32 * (1 << 30) // 32GB
 // MaxSeriesFileHashSize is the maximum number of series in a single hash.
 const MaxSeriesFileHashSize = (1048576 * LoadFactor) / 100
 
+// SeriesMapThreshold is the number of series to hold in the in-memory series map
+// before compacting and rebuilding the on-disk map.
+const SeriesMapThreshold = 100000
+
 // SeriesFile represents the section of the index that holds series data.
 type SeriesFile struct {
 	mu   sync.RWMutex
@@ -42,9 +46,10 @@ type SeriesFile struct {
 	data []byte
 	file *os.File
 	w    *bufio.Writer
+	size int64
 
-	n       uint64
-	hashMap *rhh.HashMap
+	seriesMap           *seriesMap
+	compactingSeriesMap *seriesMap
 
 	// MaxSize is the maximum size of the file.
 	MaxSize int64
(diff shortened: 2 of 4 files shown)


tsdb/index/tsi1/index.go
```diff
@@ -147,35 +147,36 @@ func (i *Index) Open() error {
...
+		}(k)
 	}
 
 	// Check for error
-	for i := 0; i < cap(errC); i++ {
+	for i := 0; i < partitionN; i++ {
 		if err := <-errC; err != nil {
 			return err
 		}
 	}
```
benbjohnson: This error check would continue on after receiving `threadN` responses instead of `partitionN` responses. The channel has a capacity of `threadN` so once this continues on then the goroutines above get blocked from adding more error responses.

This issue may have only been noticeable on computers with a smaller number of cores since as `threadN` gets closer to `partitionN` then it's less likely to occur. I have 4 threads on my machine and 16 partitions so it happened on every startup.

tsdb/index/tsi1/series_file.go
```diff
@@ -196,22 +187,35 @@ func (f *SeriesFile) CreateSeriesListIfNotExists(names [][]byte, tagsSlice []mod
...
+			f.compactingSeriesMap.inmem.Put(key, keyRange.offset)
+		}
+	}
+
+	// Begin compaction if in-memory map is past threshold.
+	if f.seriesMap.inmem.Len() >= SeriesMapThreshold {
+		if err := f.compactSeriesMap(); err != nil {
+			return nil, err
+		}
 	}
```
benbjohnson: This is the blocking compaction. I'm trying to figure out the best approach here. Rebuilding the whole map requires more writes but it makes lookups faster later on. Performing incremental maps reduces write load but requires more work during lookups.

tsdb/index/tsi1/series_file.go
```diff
@@ -257,17 +260,61 @@ func (f *SeriesFile) HasSeries(name []byte, tags models.Tags, buf []byte) bool {
...
 func (f *SeriesFile) SeriesIDIterator() SeriesIDIterator {
 	return &seriesFileIterator{
 		offset: 1,
-		data:   f.data[1:f.n],
+		data:   f.data[1:f.size],
 	}
 }
 
+func (f *SeriesFile) compactSeriesMap() error {
+	// TEMP: Compaction should occur in parallel.
```
e-dard: Can we change this to `TODO` so it's more searchable?

tsdb/index/tsi1/series_file.go
```diff
@@ -450,3 +497,173 @@ func (a seriesKeys) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
...
+	if len(m.data) == 0 {
+		return 0
+	}
+
+	hash := rhh.HashKey(key)
+	for d, pos := int64(0), hash&m.mask; ; d, pos = d+1, (pos+1)&m.mask {
+		elem := m.data[SeriesMapHeaderSize+(pos*SeriesMapElemSize):]
+		elem = elem[:SeriesMapElemSize]
+
+		h := int64(binary.LittleEndian.Uint64(elem[:8]))
```
e-dard: We don't use `LittleEndian` representations anywhere else in the codebase; is there a specific reason for choosing it in this case?

tsdb/index/tsi1/index.go
e-dard: @benbjohnson I fixed this issue in the other methods. Got about a 5% bump in ingestion. I didn't update `Open` so your PR should be able to merge on top.

tsdb/index/tsi1/index.go
```diff
@@ -147,35 +147,36 @@ func (i *Index) Open() error {
...
 
 	// Open all the Partitions in parallel.
+	partitionN := len(i.partitions)
```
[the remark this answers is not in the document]
benbjohnson: Fixed

tsdb/index/tsi1/index.go
```diff
@@ -147,35 +147,36 @@ func (i *Index) Open() error {
...
 
 	// Store results.
-	errC := make(chan error, n)
+	errC := make(chan error, partitionN)
 
 	// Run fn on each partition using a fixed number of goroutines.
 	var pidx uint32 // Index of maximum Partition being worked on.
 	for k := 0; k < n; k++ {
-		go func() {
+		go func(k int) {
```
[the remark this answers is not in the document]
benbjohnson: Fixed.

tsdb/index/tsi1/index.go
```diff
@@ -147,35 +147,36 @@ func (i *Index) Open() error {
...
+	for i := 0; i < partitionN; i++ {
 		if err := <-errC; err != nil {
 			return err
 		}
 	}
 
 	// Mark opened.
 	i.opened = true
-	i.logger.Info(fmt.Sprintf("index opened with %d partitions", len(i.partitions)))
+	i.logger.Info(fmt.Sprintf("index opened with %d partitions", partitionN))
```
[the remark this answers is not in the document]
benbjohnson: Fixed.

tsdb/index/tsi1/series_file.go
```diff
@@ -90,10 +95,12 @@ func (f *SeriesFile) Open() error {
 	}
 	f.data = data
 
-	// Index all series.
-	if err := f.reindex(); err != nil {
+	// Load series map.
+	m := newSeriesMap(f.path+SeriesMapFileSuffix, f)
```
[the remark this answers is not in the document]
benbjohnson: I wanted to make the file `seriesmap` so I did a simple concatenation.

tsdb/index/tsi1/series_file.go
benbjohnson: Fixed.

tsdb/index/tsi1/series_file.go
benbjohnson: Fixed.

PR #8909, opened 2017-10-03
Use correct variable style in influxdb.service

mark-rushakoff:
From the systemd.service docs:

Use "${FOO}" as part of a word, or as a word of its own, on the command
line, in which case it will be replaced by the value of the environment
variable including all whitespace it contains, resulting in a single
argument. Use "$FOO" as a separate word on the command line, in which
case it will be replaced by the value of the environment variable split
at whitespace, resulting in zero or more arguments.

- [x] Rebased/mergable
- [ ] Tests pass
- [x] CHANGELOG.md updated


mark-rushakoff:
Added a Needs Backporting label in case there is a 1.3.7 release.

mark-rushakoff:
Backported in #8910, removing that label.

PR #8931, opened 2017-10-05
Update to next version of Go with security fixes

mark-rushakoff:
- [x] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -69,7 +69,7 @@ second to sign our CLA, which can be found
 
 Installing Go
 -------------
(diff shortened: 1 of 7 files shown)


mark-rushakoff:
We've been running unit tests successfully with Go 1.9 but I'm not aware of any formal integration tests with Go 1.9. I think at least some of the developers use Go 1.9 on their workstations.

Go 1.9.X and influxdb 1.4.0 is a conversation we need to revisit with the team.

PR #8941, opened 2017-10-09
OSS export for use in Enterprise Import

aanthony1243:
This is ready for a full review.  Still need to update changelog and produce a short doc for export.  Thanks for all the feedback so far.  

example usage: 
for existing backup function use -since or no flag: 
influxd  backup -since 2017-09-01T00:00:00Z -database collectd_db /Users/adam/tmp/

for export, use start/end: 
influxd  backup -start 2017-09-01T00:00:00Z -end 2017-10-01T11:11:11Z -database collectd_db /Users/adam/tmp/

- [X] Rebased/mergable
- [X] Tests pass
- [ ] CHANGELOG.md updated

- [x] Provide example syntax

###### these need to be done when the Import command ships, but isn't worth much now.  


tsdb/engine/tsm1/engine.go
```diff
@@ -653,6 +653,162 @@ func (e *Engine) Backup(w io.Writer, basePath string, since time.Time) error {
...
+		if !strings.HasSuffix(file, ".tsm") {
+			continue
+		}
+
+		var tombstonePath string
+		f, err := os.Open(filepath.Join(path, file))
+		if err != nil {
+			return err
+		}
+		r, err := NewTSMReader(f)
```
jwilder: Opening a TSMReader multiple times is kind of expensive.  It's done here (and thrown away), then again here: https://github.com/influxdata/influxdb/pull/8941/files#diff-82e4e65bd72f00208a49c4797d466ecdR768

It would better to only have to open each file once.

tsdb/engine/tsm1/engine.go
```diff
@@ -653,6 +653,162 @@ func (e *Engine) Backup(w io.Writer, basePath string, since time.Time) error {
...
+func (e *Engine) filterFileToBackup(name, shardRelativePath, fullPath string, start, end int64, tw *tar.Writer) error {
+	f, err := os.Stat(fullPath)
+	if err != nil {
+		return err
+	}
+
+	h := &tar.Header{
+		Name:    filepath.ToSlash(filepath.Join(shardRelativePath, name)),
+		ModTime: f.ModTime(),
+		Size:    f.Size(),
```
jwilder: This size won't be correct since we're filtering out data from the original file.

tsdb/engine/tsm1/engine.go
```diff
@@ -653,6 +653,162 @@ func (e *Engine) Backup(w io.Writer, basePath string, since time.Time) error {
...
+		return err
+	}
+	defer func() {
+		closeErr := w.Close()
+		if err == nil {
+			err = closeErr
+		}
+
+		// Check for errors where we should not remove the file
+		// REVIEWER:  do I need to check anything here? lifted this code from compact.go
```
jwilder: nit: You can review/comment on your own PR for things like this vs committing them as comments in the code.

tsdb/engine/tsm1/engine.go
```diff
@@ -653,6 +653,162 @@ func (e *Engine) Backup(w io.Writer, basePath string, since time.Time) error {
...
+
+		if err != nil {
+			w.Remove()
```
jwilder: This isn't necessary as you are writing a TSM file to a tar stream.  This would attempt to delete it from the tar stream which you can't really do.

tsdb/engine/tsm1/engine.go
```diff
@@ -653,6 +653,162 @@ func (e *Engine) Backup(w io.Writer, basePath string, since time.Time) error {
...
+
+	for bi.Next() {
+		// not concerned with typ of checksum since we are just blindly writing back, with no decoding
+		key, minTime, maxTime, _, _, buf, err := bi.Read()
+		if err != nil {
+			return err
+		}
+		if minTime >= start && minTime <= end ||
+			maxTime >= start && maxTime <= end {
+			w.WriteBlock(key, minTime, maxTime, buf)
```
jwilder: Missing err check.

tsdb/engine/tsm1/engine.go
```diff
@@ -653,6 +653,162 @@ func (e *Engine) Backup(w io.Writer, basePath string, since time.Time) error {
...
+		return err
+	}
+	fr, err := os.Open(fullPath)
+	if err != nil {
+		return err
+	}
+
+	defer fr.Close()
+
+	// TODO (Adam):  instead of io.copy as in writeFileToBackup, we will need to:
```
jwilder: nit: remove these comments?  Looks like in-progress notes for you?

cmd/influxd/export/export.go
```diff
@@ -0,0 +1,367 @@
...
+	if err := json.NewDecoder(conn).Decode(&r); err != nil {
+		return nil, err
+	}
+
+	return &r, nil
+}
+
+// printUsage prints the usage message to STDERR.
+func (cmd *Command) printUsage() {
+	fmt.Fprintf(cmd.Stdout, `TODO: write usage for export`)
```
jwilder: Needs usage filled in: `TODO: write usage for export`

cmd/influxd/export/export.go
```diff
@@ -0,0 +1,367 @@
...
+	// Rename temporary file to final path.
+	if err := os.Rename(tmppath, path); err != nil {
+		return fmt.Errorf("rename: %s", err)
+	}
+
+	return nil
+}
+
+// download downloads a snapshot of either the metastore or a shard from a host to a given path.
+func (cmd *Command) download(req *snapshotter.Request, path string) error {
```
jwilder: A lot of this `download` related code looks copied form the backup command.  Maybe extract that into something re-usable by both?

cmd/influxd/export/export.go
```diff
@@ -0,0 +1,367 @@
...
+}
+
+// exportDatabase will request the database information from the server and then backup the metastore and
+// every shard in every retention policy in the database. Each shard will be written to a separate tar.
+func (cmd *Command) exportDatabase(tsStart, tsEnd time.Time) error {
+	cmd.StdoutLogger.Printf("backing up db=%s start %s end %s", cmd.database, tsStart, tsEnd)
+
+	req := &snapshotter.Request{
+		Type:     snapshotter.RequestDatabaseInfo,
+		Database: cmd.database,
```
joelegasse: `parseFlags` doesn't require `-database` to be set, but this call does require a single database. Is the command meant to work for multiple databases, or on a per-database basis? Either make the flag required, or update the code to allow exporting all databases when none are specified, whichever one is the intended behavior.

cmd/influxd/export/export.go
```diff
@@ -0,0 +1,367 @@
...
+	}
+	cmd.StdoutLogger.Println("made it to before the export response paths")
```
joelegasse: This sounds like a debug print statement. Either remove it or update it to be a user-meaningful message

cmd/influxd/export/export.go
```diff
@@ -0,0 +1,367 @@
...
+	response, err := cmd.requestInfo(req)
+	if err != nil {
+		return err
+	}
+	cmd.StdoutLogger.Println("made it to the export response paths")
```
joelegasse: update/remove debug statement

cmd/influxd/export/export.go
```diff
@@ -0,0 +1,367 @@
...
+	return cmd.exportResponsePaths(response, tsStart, tsEnd)
+}
+
+// exportRetentionPolicy will request the retention policy information from the server and then backup
+// the metastore and every shard in the retention policy. Each shard will be written to a separate tar.
+func (cmd *Command) exportRetentionPolicy(retentionPolicy string, tsStart, tsEnd time.Time) error {
```
joelegasse: This method is never called from anywhere...

cmd/influxd/export/export.go
```diff
@@ -0,0 +1,367 @@
...
+	}
+
+	cmd.StdoutLogger.Printf("backing up metastore to %s", metastoreArchivePath)
+
+	req := &snapshotter.Request{
+		Type: snapshotter.RequestMetastoreBackup,
+	}
+
+	return cmd.downloadAndVerify(req, metastoreArchivePath, func(file string) error {
+		binData, err := ioutil.ReadFile(file)
```
joelegasse: Reading in an entire file just to check 8 bytes? Seems silly. I would suggest either verifying the entire file (can decode the blob) or just reading in the magic bytes.

cmd/influxd/export/export.go
```diff
@@ -0,0 +1,367 @@
...
+
+// exportResponsePaths will backup the metastore and all shard paths in the response struct
+func (cmd *Command) exportResponsePaths(response *snapshotter.Response, tsStart, tsEnd time.Time) error {
+	if err := cmd.exportMetastore(); err != nil {
+		return err
+	}
+
+	// loop through the returned paths and back up each shard
+	for _, path := range response.Paths {
+		rp, id, err := retentionAndShardFromPath(path)
```
joelegasse: I think this approach is very wasteful. Your'e requesting the database info, which is really just a bunch of relative shard paths, and then parsing those paths to get the rp name and shard ID. All of this while having a full binary metadata snapshot that can be decoded and contains that information for all databases.

I think the overall approach of this command should take that into account: grab the latest metadata snapshot, and filter shards based on that. You'll even have local access to the timestamps for each shard, and won't need to return "empty" shards from the server.

cmd/influxd/export/export.go
joelegasse: After looking further, I think this blob should definitely just be decoded and used as the source of metadata for which shards to pull down, and not rely on shard paths that don't contain timestamp info.

tsdb/engine/tsm1/engine.go
```diff
@@ -653,6 +653,162 @@ func (e *Engine) Backup(w io.Writer, basePath string, since time.Time) error {
...
+		min, max := r.TimeRange()
+		if min > end.UnixNano() || max < start.UnixNano() {
+			r.Close()
+			continue
+		}
+
+		r.Close()
+
+		if tombstonePath != "" {
+			whole = append(filtered, tombstonePath)
```
joelegasse: `whole = append(filtered, ...)`? Is this intentional, or a typo?

tsdb/engine/tsm1/engine.go
```diff
@@ -653,6 +653,162 @@ func (e *Engine) Backup(w io.Writer, basePath string, since time.Time) error {
...
+		}
+
+		if min > start.UnixNano() && max < end.UnixNano() {
+			whole = append(filtered, file)
```
joelegasse: `a = append(b,...)` again

cmd/influxd/export/export.go
aanthony1243: I'm going to prioritize other work and make a ticket for this.  #8964

cmd/influxd/export/export.go
aanthony1243: optimized this to get just the first 8 bytes.  See above RE: metadata crawling.  probably I will get to this, since we'll be doing work on the OSS import/restore side next.

tsdb/engine/tsm1/engine.go
```diff
@@ -653,6 +654,135 @@ func (e *Engine) Backup(w io.Writer, basePath string, since time.Time) error {
...
+		// so the helper functions will do the closing.
+		// to write the next file to the gzip stream, we just reset the gzip writer.
+		// later, this zipped stream must be read with zr.Multistream(false) so that it will stop reading on each new file.
+		zw.Reset(w)
+	}
+
+	return nil
+}
+
+func (e *Engine) filterFileToGzip(r *TSMReader, name, shardRelativePath, fullPath string, start, end int64, zw *gzip.Writer) error {
```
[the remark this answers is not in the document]
joelegasse: After some digging, it looks like `TSMWriter.Close()` will close the underlying writer, as this is abnormal behavior from other Go wrapped `Writer` implementations, I would add a comment above that deferred close that it closes the underlying writer (`zw` in this case).

tsdb/engine/tsm1/engine_test.go
```diff
@@ -250,6 +251,274 @@ func TestEngine_Backup(t *testing.T) {
...
+			continue
+		}
+		if !equalBuffers(fullExp, v) {
+			t.Errorf("2Difference in data between backup and Export for file %s", th.Name)
+		}
+
+	}
+}
+
+func equalBuffers(bufA, bufB *bytes.Buffer) bool {
```
joelegasse: Is there a reason you didn't use https://golang.org/pkg/bytes/#Equal here on the `Bytes()` calls? If `bufB.Len() < bufA.Len()`, this will panic. Also, it is only checking that `bufB` has a prefix of `bufA`, not that they are the same length and have the same contents.

tsdb/engine/tsm1/engine_test.go
```diff
@@ -250,6 +251,274 @@ func TestEngine_Backup(t *testing.T) {
...
+	}
+
+	for k, v := range ex3Data {
+		fullExp, ok := fileData[k]
+		if !ok {
+			t.Errorf("Extracting subset resulted in file not found in full export: %s", err.Error())
+			continue
+		}
+		if !equalBuffers(fullExp, v) {
+			t.Errorf("2Difference in data between backup and Export for file %s", th.Name)
```
joelegasse: typo with `2Difference` everywhere, or am I missing something?

tsdb/engine/tsm1/engine_test.go
```diff
@@ -250,6 +251,274 @@ func TestEngine_Backup(t *testing.T) {
...
+
+	th, err := tr.Next()
+	for err == nil {
+		expData, ok := fileData[th.Name]
+		if !ok {
+			t.Errorf("Extra file in backup: %q", th.Name)
+			continue
+		}
+
+		buf := new(bytes.Buffer)
```
joelegasse: tar readers know the size of the files. You could replace this with a fixed-size buffer and use `ReadFull` rather than having to use a `bytes.Buffer` everywhere.

cmd/influxd/backup/backup.go
```diff
@@ -62,106 +68,189 @@ func (cmd *Command) Run(args ...string) error {
...
-		since, err = time.Parse(time.RFC3339, sinceArg)
+		cmd.since, err = time.Parse(time.RFC3339, sinceArg)
 		if err != nil {
-			return
+			return err
 		}
 	}
+	if startArg != "" {
+		if cmd.isBackup {
+			return errors.New("backup command uses one of -since or -start/-end")
```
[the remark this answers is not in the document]
aanthony1243: if the user is not paying attention, they may set '-since T1  -end T2'  and it will set isBackup to true.

tsdb/engine/tsm1/engine.go
```diff
@@ -700,111 +698,87 @@ func (e *Engine) Export(w io.Writer, basePath string, start time.Time, end time.
...
 		min, max := r.TimeRange()
-		if min > end.UnixNano() || max < start.UnixNano() {
-			r.Close()
-			continue
-		}
 
-		r.Close()
+		// We overlap time ranges, we need to filter the file
+		if min >= start.UnixNano() && min <= end.UnixNano() ||
+			max >= start.UnixNano() && max <= end.UnixNano() {
```
jwilder: nit: I think you can simplify this to:

```
min <= end.UnixNano() && max >= start.UnixNano()
```

tsdb/engine/tsm1/engine.go
```diff
@@ -700,111 +698,87 @@ func (e *Engine) Export(w io.Writer, basePath string, start time.Time, end time.
...
+			if err := e.filterFileToGzip(r, file, basePath, filepath.Join(path, file), start.UnixNano(), end.UnixNano(), zw); err != nil {
+				return err
```
jwilder: This looks like it will leak the reader being open when there is an error.

tsdb/engine/tsm1/engine.go
```diff
@@ -700,111 +698,87 @@ func (e *Engine) Export(w io.Writer, basePath string, start time.Time, end time.
...
 
-		if min > start.UnixNano() && max < end.UnixNano() {
-			whole = append(filtered, file)
-		} else {
-			// at this point, we know we're not totally outside or inside start and end so we'll need to scan the blocks
-			filtered = append(filtered, file)
-		}
-	}
+		// above is the only case where we need to keep the reader open.
+		r.Close()
```
jwilder: This looks like it should be within the loop above or every reader will leak except the last one.

jwilder: Also, error is ignored here.

cmd/influxd/backup/backup.go
```diff
@@ -62,106 +68,189 @@ func (cmd *Command) Run(args ...string) error {
...
+		}
+		cmd.end, err = time.Parse(time.RFC3339, endArg)
+		if err != nil {
+			return err
+		}
+	}
+
+	// some validations
+	// 1.  -database is required
```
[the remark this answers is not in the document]
aanthony1243: removed '-database' requirement and fixed the comments.

tsdb/engine/tsm1/engine.go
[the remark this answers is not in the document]
aanthony1243: yeah, that's actually an annoying side effect that makes the code less clear.  But I figured changing TSMWriter would have a....huge...ripple effect.

tsdb/engine/tsm1/engine.go
```diff
@@ -700,111 +698,87 @@ func (e *Engine) Export(w io.Writer, basePath string, start time.Time, end time.
...
+		// if this TSM file had a tombstone we'll write out the whole thing too.
+		if tombstonePath != "" {
+			if err := e.writeFileToGzip(tombstonePath, basePath, filepath.Join(path, tombstonePath), zw); err != nil {
+				return err
+			}
 		}
+		// when writing a file to Gzip, zw.Close() is how you signify you are done writing the file.
+		// so the helper functions will do the closing.
+		// to write the next file to the gzip stream, we just reset the gzip writer.
+		// later, this zipped stream must be read with zr.Multistream(false) so that it will stop reading on each new file.
```
jwilder: I don't think we use multistream on import.  This may break some existing things in enterprise.

tsdb/engine/tsm1/engine.go
```diff
@@ -700,111 +698,87 @@ func (e *Engine) Export(w io.Writer, basePath string, start time.Time, end time.
...
 			maxTime >= start && maxTime <= end {
-			w.WriteBlock(key, minTime, maxTime, buf)
+			err := w.WriteBlock(key, minTime, maxTime, buf)
+			if err != nil {
+				return err
+			}
 		}
 	}
-	w.WriteIndex()
+
```
jwilder: There is a recent change in #8981 which adds an `Err` func for `BlockIterator`.  You'll need to ensure that is `nil` before writing the index and returning success.

```
if err := bi.Err(); err != nil {
   return err
}
```

cmd/influxd/backup/backup.go
```diff
@@ -62,106 +68,189 @@ func (cmd *Command) Run(args ...string) error {
...
-	} else if retentionPolicy != "" {
-		err = cmd.backupRetentionPolicy(retentionPolicy, since)
+		err = cmd.extractRetentionPolicy()
 	} else if cmd.database != "" {
-		err = cmd.backupDatabase(since)
+		// always backup the metastore
+		if err := cmd.extractMetastore(cmd.database); err != nil {
+			return err
+		}
+		err = cmd.extractDatabase()
```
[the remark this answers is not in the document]
aanthony1243: fixed.

cmd/influxd/backup/backup.go
```diff
@@ -62,106 +68,189 @@ func (cmd *Command) Run(args ...string) error {
...
 	} else {
-		err = cmd.backupMetastore()
+		// always backup the metastore
```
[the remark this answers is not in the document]
aanthony1243: un-required it, forgot to push that.

cmd/influxd/backup/backup.go
```diff
@@ -62,106 +68,189 @@ func (cmd *Command) Run(args ...string) error {
...
+		if err := cmd.extractMetastore(""); err != nil {
+			return err
+		}
+		cmd.StdoutLogger.Printf("No database, retention policy or shard ID given. Full meta store backed up.")
```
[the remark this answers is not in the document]
aanthony1243: fixed

cmd/influxd/backup/backup.go
```diff
@@ -62,106 +68,189 @@ func (cmd *Command) Run(args ...string) error {
...
 	}
 
 	if err != nil {
 		cmd.StderrLogger.Printf("backup failed: %v", err)
 		return err
 	}
-
-	cmd.StdoutLogger.Println("backup complete")
+	cmd.StdoutLogger.Printf("backup complete")
```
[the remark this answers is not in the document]
aanthony1243: fixed

cmd/influxd/backup/backup.go
```diff
@@ -62,106 +68,189 @@ func (cmd *Command) Run(args ...string) error {
...
+	}
+
+	// some validations
+	// 1.  -database is required
+	if cmd.database == "" {
+		return errors.New("-database <dbname> is a required argument")
+	}
+
+	// 2.  start should be < end
+	if cmd.start.After(cmd.end) {
```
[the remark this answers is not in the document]
aanthony1243: fixed

cmd/influxd/backup/backup.go
```diff
@@ -62,106 +68,189 @@ func (cmd *Command) Run(args ...string) error {
...
 	cmd.path = fs.Arg(0)
 
 	err = os.MkdirAll(cmd.path, 0700)
 
-	return
+	return err
+}
+
+func (cmd *Command) extractShard(rp, sid string) error {
+	if cmd.isBackup {
```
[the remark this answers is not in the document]
aanthony1243: good suggestion.  the only difference is the flag in the request.  fixed.

cmd/influxd/backup/backup.go
```diff
@@ -173,52 +262,51 @@ func (cmd *Command) backupDatabase(since time.Time) error {
...
 		}
 	}
 
 	return nil
 }
 
-// backupMetastore will backup the metastore on the host to the passed in path. Database and retention policy backups
-// will force a backup of the metastore as well as requesting a specific shard backup from the command line
-func (cmd *Command) backupMetastore() error {
+// extractMetastore will backup the whole metastore on the host to the passed in path.
```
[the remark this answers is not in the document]
aanthony1243: fixed

cmd/influxd/backup/backup.go
```diff
@@ -227,16 +315,27 @@ func (cmd *Command) backupMetastore() error {
...
 
 	req := &snapshotter.Request{
-		Type: snapshotter.RequestMetastoreBackup,
+		Type:     snapshotter.RequestMetastoreBackup,
+		Database: useDB,
 	}
 
 	return cmd.downloadAndVerify(req, metastoreArchivePath, func(file string) error {
-		binData, err := ioutil.ReadFile(file)
+		f, err := os.Open(file)
```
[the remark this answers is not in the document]
aanthony1243: fixed.

cmd/influxd/backup/backup.go
```diff
@@ -227,16 +315,27 @@ func (cmd *Command) backupMetastore() error {
...
 		if err != nil {
 			return err
 		}
 
-		magic := binary.BigEndian.Uint64(binData[:8])
+		magicByte := make([]byte, 8)
+		n, err := f.Read(magicByte)
```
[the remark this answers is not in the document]
aanthony1243: fixed

services/snapshotter/service.go
```diff
@@ -133,9 +133,20 @@ func (s *Service) handleConn(conn net.Conn) error {
...
-func (s *Service) writeMetaStore(conn net.Conn) error {
+func (s *Service) writeMetaStore(conn net.Conn, database string) error {
 	// Retrieve and serialize the current meta data.
-	metaBlob, err := s.MetaClient.MarshalBinary()
+	// if the database name is non-empty, then we drop all the DB's from the metadata
+	// that aren't being exported.
+	data := s.MetaClient.Data()
+	if database != "" {
+		for _, db := range data.Databases[:] {
+			if db.Name != database {
```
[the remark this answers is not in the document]
aanthony1243: thanks, that's so much better!

services/snapshotter/service.go
```diff
@@ -133,9 +133,20 @@ func (s *Service) handleConn(conn net.Conn) error {
...
 
-func (s *Service) writeMetaStore(conn net.Conn) error {
+func (s *Service) writeMetaStore(conn net.Conn, database string) error {
 	// Retrieve and serialize the current meta data.
-	metaBlob, err := s.MetaClient.MarshalBinary()
+	// if the database name is non-empty, then we drop all the DB's from the metadata
+	// that aren't being exported.
+	data := s.MetaClient.Data()
+	if database != "" {
+		for _, db := range data.Databases[:] {
```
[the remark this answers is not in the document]
aanthony1243: ok, removed for other reasons but what I was trying to avoid (and I see my error now) was modifying the array that I was iterating.  So I needed a copy.   I got a bit pythonic there, need to unlearn some things :)

aanthony1243:
Rolled into PR#9040

PR #8947, opened 2017-10-10
EXPLAIN ANALYZE implementation

stuartcarnie:
EXPLAIN ANALYZE SELECT ... will measure a number of counters whilst
executing the query. All data is processed, providing an accurate
account of blocks that were decoded, cursors created for each
expression in a SELECT statement.

The total planning and execution times are measured.

- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated


stuartcarnie:
@jsternberg I've broken the PR into multiple commits so it is easier for you to see the changes

stuartcarnie:
[Example output](https://gist.github.com/stuartcarnie/280ae035d364b2727c2861b5716f385a)

coordinator/shard_mapper_test.go
```diff
@@ -74,7 +75,7 @@ func TestLocalShardMapper(t *testing.T) {
 		t.Fatalf("unexpected number of shard mappings: %d", len(m.ShardMap))
 	}
 
-	if _, err := ic.CreateIterator(measurement, query.IteratorOptions{}); err != nil {
+	if _, err := ic.CreateIterator(context.TODO(), measurement, query.IteratorOptions{}); err != nil {
```
jsternberg: Can you modify these various `context.TODO()` calls to what they're supposed to be?

pkg/tracing/context.go
```diff
@@ -0,0 +1,32 @@
+package tracing
```
jsternberg: Can you add a `doc.go` file to this package to explain the behavior? I think I've figured it out, but it's a bit confusing and I don't see very many doc comments.

If I'm correctly, you create a `Trace` at the beginning of the thing you want to trace. You then use the `Span` to create child spans. All spans are managed by the `Trace` and each `Span` only keeps track of its own level. It does not keep track of anything else.

Godeps
```diff
@@ -17,8 +17,10 @@ github.com/peterh/liner 88609521dc4b6c858fd4c98b628147da928ce4ac
 github.com/philhofer/fwd 1612a298117663d7bc9a760ae20d383413859798
 github.com/retailnext/hllpp 38a7bb71b483e855d35010808143beaf05b67f9d
 github.com/spaolacci/murmur3 0d12bf811670bf6a1a63828dfbd003eded177fce
+github.com/stretchr/testify 890a5c3458b43e6104ff5da8dfa139d013d77544
```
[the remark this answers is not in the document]
stuartcarnie: I agree with not using other _testing_ packages. This package is included only to leverage the assertion package `testify/assert`. It removes a lot of noise from unit tests when asserting for expected behavior / values. If the assertion fails, it will generate a standard t.Error, t.Fail, etc. For more complex types, like collections, diffs will also be produced. 

`testify/assert` contains APIs for simplifying a number of scenarios and uses the standard Go `*testing.T` type to report errors or failures. Some example

* [assert.PanicsWithValue](https://github.com/influxdata/influxdb/pull/8947/files#diff-82e168f130315192395e34c360d6910eR18) – asserts the code panics and returns an expected value. Much simpler than writing `defer`s, etc

Others I like include:

* `assert.NotNil`
* `assert.NoError`
* `assert.Empty`

This little [introduction](https://github.com/stretchr/testify#assert-package) is helpful. You can also review [the docs](https://godoc.org/github.com/stretchr/testify/assert) for the available assertions.

coordinator/shard_mapper_test.go
stuartcarnie: I'll change them to `Background`

tsdb/engine/tsm1/engine.go
```diff
@@ -18,11 +18,15 @@ import (
 	"sync/atomic"
 	"time"
 
+	"strconv"
```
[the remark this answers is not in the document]
stuartcarnie: done

coordinator/shard_mapper_test.go
stuartcarnie: Done

Godeps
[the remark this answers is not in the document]
jsternberg: I'm aware of this testing package. But, we don't use it anywhere else and I don't like the idea of including an additional dependency for only one section of code. Introducing this likely means that it will start being used in other locations and the unit tests won't have the same pattern anymore. They don't currently, but adding additional unit test patterns to a code base that already has a bunch seems like too much to me.

pkg/tracing/context.go
stuartcarnie: Done

Godeps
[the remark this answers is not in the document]
stuartcarnie: Removed in favor of `pkg/testing/assert`

LICENSE_OF_DEPENDENCIES.md
```diff
@@ -25,3 +25,5 @@
 - github.com/uber-go/zap [MIT LICENSE](https://github.com/uber-go/zap/blob/master/LICENSE.txt)
 - golang.org/x/crypto [BSD LICENSE](https://github.com/golang/crypto/blob/master/LICENSE)
 - jquery 2.1.4 [MIT LICENSE](https://github.com/jquery/jquery/blob/master/LICENSE.txt)
+- github.com/xlab/treeprint [MIT LICENSE](https://github.com/xlab/treeprint/blob/master/LICENSE)
+- github.com/stretchr/testify [MIT LICENSE](https://github.com/stretchr/testify/blob/master/LICENSE)
```
jsternberg: This license needs to be removed if we aren't using this dependency.

PR #8949, opened 2017-10-10
Engine race fixes

jwilder:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

This fixes two races in the engine.  Once occurs on the engine `WaitGroup` when `Add` is called while another goroutine is still `Wait`ing.  The other occurs when deletes occur while rebuilding the in-memory measurement index asynchronously.

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -97,15 +97,11 @@ type Engine struct {
 
 	// The following group of fields is used to track the state of level compactions within the
 	// Engine. The WaitGroup is used to monitor the compaction goroutines, the 'done' channel is
-	// used to signal those goroutines to shutdown. Every request to disable level compactions will
-	// call 'Wait' on 'wg', with the first goroutine to arrive (levelWorkers == 0 while holding the
-	// lock) will close the done channel and re-assign 'nil' to the variable. Re-enabling will
-	// decrease 'levelWorkers', and when it decreases to zero, level compactions will be started
-	// back up again.
+	// used to signal those goroutines to shutdown.
 
-	wg           sync.WaitGroup // waitgroup for active level compaction goroutines
-	done         chan struct{}  // channel to signal level compactions to stop
-	levelWorkers int            // Number of "workers" that expect compactions to be in a disabled state
+	wg      sync.WaitGroup // waitgroup for active level compaction goroutines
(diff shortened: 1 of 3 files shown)


tsdb/engine/tsm1/engine.go
```diff
@@ -211,72 +207,63 @@ func (e *Engine) SetEnabled(enabled bool) {
...
+func (e *Engine) disableLevelCompactions() {
+	// Already disabled?
+	if atomic.LoadInt32(&e.running) == 0 {
+		return
 	}
 
-	if old == 0 && e.done != nil {
-		// Prevent new compactions from starting
-		e.Compactor.DisableCompactions()
+	e.mu.Lock()
```
e-dard: In `enableLevelCompactions` `mu`'s unlocking is deferred. Is there a reason why we can't defer this one?

tsdb/index/inmem/meta.go
```diff
@@ -286,7 +286,10 @@ func (m *Measurement) Rebuild() *Measurement {
 	// expunged.  Note: we're using SeriesIDs which returns the series in sorted order so that
 	// re-adding does not incur a sort for each series added.
 	for _, id := range m.SeriesIDs() {
-		nm.AddSeries(m.SeriesByID(id))
+		s := m.SeriesByID(id)
```
e-dard: nit, you can put this on one line `if s := m.SeriesByID(id); s != nil {`

e-dard: Also, if we don't want to add `nil` Series to the new measurement can't we just add the no-op for the `nil` case to the `AddSeries` implementation?

tsdb/engine/tsm1/engine.go
jwilder: We need to unlock before call `Wait` so that any goroutines waiting to acquire the ending lock can acquire and exit.

tsdb/index/inmem/meta.go
jwilder: Yeah, I can have `AddSeries` handle adding `nil`.

tsdb/engine/tsm1/engine.go
e-dard: Ah yeah I missed the `Wait` call 👍

PR #8955, opened 2017-10-12
remove unnecessary err value

stuartcarnie:
`readKey` never sets `err`, therefore call sites checking for errors are unnecessary

- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated


stuartcarnie:
Sufficient space in the buffer was validated when [`UnmarshalBinary`](https://github.com/influxdata/influxdb/blob/a0848eac8c9bb1e7faf66043b00e1ee293b8b6f8/tsdb/engine/tsm1/reader.go#L968-L1006) was called

PR #8959, opened 2017-10-12
Use multistage builds in a Dockerfile to compile and create an image from source

jsternberg:
The multistage build lets us build the image inside of one container and
then copy over the results to a new image so the build dependencies
aren't needed in the final container.

The resulting container should be identical to the official built ones
for every release, but coming from source instead.

- [x] Rebased/mergable
- [x] Tests pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM golang:1.9.2 as builder
+RUN go get -u github.com/sparrc/gdm
+COPY Godeps /go/src/github.com/influxdata/influxdb/Godeps
+WORKDIR /go/src/github.com/influxdata/influxdb
+RUN gdm restore
+COPY . /go/src/github.com/influxdata/influxdb
+RUN go install ./cmd/...
+
+FROM debian:stretch
+COPY --from=builder /go/bin/* /usr/bin/
+COPY --from=builder /go/src/github.com/influxdata/influxdb/etc/config.sample.toml /etc/influxdb/influxdb.conf
+
+EXPOSE 8086
+VOLUME /var/lib/influxdb
+
+COPY docker/entrypoint.sh /entrypoint.sh
+COPY docker/init-influxdb.sh /init-influxdb.sh
+ENTRYPOINT ["/entrypoint.sh"]
+CMD ["influxd"]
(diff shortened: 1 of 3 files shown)


docker-init-influxdb.sh
```diff
@@ -0,0 +1,120 @@
...
+
+AUTH_ENABLED="$INFLUXDB_HTTP_AUTH_ENABLED"
+
+if [ -z "$AUTH_ENABLED" ]; then
+	AUTH_ENABLED="$(grep -iE '^\s*auth-enabled\s*=\s*true' /etc/influxdb/influxdb.conf | grep -io 'true' | cat)"
+else
+	AUTH_ENABLED="$(echo ""$INFLUXDB_HTTP_AUTH_ENABLED"" | grep -io 'true' | cat)"
+fi
+
+INIT_USERS=$([ ! -z "$AUTH_ENABLED" ] && [ ! -z "$INFLUXDB_ADMIN_USER" ] && echo 1 || echo)
```
mark-rushakoff: Is there a reason to use `! -z` instead of `-n`?

And are the environment variables in this file documented anywhere?

jsternberg: Yea, they're documented in the docker docs themselves. I copied the file directly from there.

I'm not sure at the moment how we should keep the two connected with each other. I just thought this feature was cool and a decent way to allow someone to more easily build a Docker image from master.

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+COPY Godeps /go/src/github.com/influxdata/influxdb/Godeps
+WORKDIR /go/src/github.com/influxdata/influxdb
+RUN gdm restore
+COPY . /go/src/github.com/influxdata/influxdb
+RUN go install ./cmd/...
+
+FROM debian:stretch
+COPY --from=builder /go/bin/* /usr/bin/
+RUN mkdir -p /etc/influxdb && \
+    influxd config > /etc/influxdb/influxdb.conf
```
jwilder: This shouldn't be `influxd config`, should copy in the the `/etc/config.sample.toml` as the config.

jsternberg: Done.

jsternberg:
@jwilder I updated this to use go 1.9.2 and moved the docker specific files to a folder instead.

PR #8962, opened 2017-10-13
Modify each service to use a different logging pattern

jsternberg:
Each package that logs has its own interface for the logger to
implement. The interface is focused more on the actions that occur
rather than log messages and the log messages and control of the logger
is controlled in the `diagnostics` package.

This changes each one of the packages except for the tsdb packages to
this new pattern. The tsdb packages are more complicated to convert so
that will be a separate commit that can be reviewed on its own.

This also removes the logger from the execution context in the query
package altogether. It wasn't being used.

- [ ] Rebased/mergable
- [ ] Tests pass


cmd/influxd/run/server.go
```diff
@@ -63,7 +64,8 @@ type Server struct {
 	BindAddress string
 	Listener    net.Listener
 
-	Logger zap.Logger
+	Logger            zap.Logger
```
desa: Should we upgrade to a more recent version of zap? I think that this is an older version.

desa: This might be a change for later, but could also be useful to do now.

cmd/influxd/run/server.go
```diff
@@ -247,6 +252,7 @@ func (s *Server) appendRetentionPolicyService(c retention.Config) {
 	srv := retention.NewService(c)
 	srv.MetaClient = s.MetaClient
 	srv.TSDBStore = s.TSDBStore
+	srv.With(s.DiagnosticService.InRetentionContext())
```
desa: Any thoughts on removing the `In` prefix from these and just having 

```
srv.With(s.DiagnosticService.RetentionContext())
```

I think it reads pretty well.

coordinator/points_writer.go
```diff
@@ -40,12 +38,16 @@ var (
 	ErrWriteFailed = errors.New("write failed")
 )
 
+type Diagnostic interface {
+	WriteFailed(shardID uint64, err error)
```
desa: I really like having specific errors like this.

coordinator/points_writer.go
```diff
@@ -148,8 +149,8 @@ func (w *PointsWriter) AddWriteSubscriber(c chan<- *WritePointsRequest) {
 }
 
 // WithLogger sets the Logger on w.
-func (w *PointsWriter) WithLogger(log zap.Logger) {
-	w.Logger = log.With(zap.String("service", "write"))
+func (w *PointsWriter) With(d Diagnostic) {
```
desa: I like this pattern as well

coordinator/points_writer.go
```diff
@@ -380,15 +381,18 @@ func (w *PointsWriter) writeToShard(shard *meta.ShardInfo, database, retentionPo
 	if err == tsdb.ErrShardNotFound {
 		err = w.TSDBStore.CreateShard(database, retentionPolicy, shard.ID, true)
 		if err != nil {
-			w.Logger.Info(fmt.Sprintf("write failed for shard %d: %v", shard.ID, err))
-
+			if w.Diagnostic != nil {
```
desa: Not a huge fan of this nil check on the diagnostic, but I guess we have to :/

diagnostic/handlers.go
```diff
@@ -0,0 +1,582 @@
...
+func (h *CollectdHandler) UnableToReadDirectory(path string, err error) {
+	h.l.Info(fmt.Sprintf("Unable to read directory %s: %s\n", path, err))
+}
+
+func (h *CollectdHandler) LoadingPath(path string) {
+	h.l.Info(fmt.Sprintf("Loading %s\n", path))
+}
+
+func (h *CollectdHandler) TypesParseError(path string, err error) {
+	h.l.Info(fmt.Sprintf("Unable to parse collectd types file: %s\n", path))
```
desa: Any reason for not using `zap.Fields` in each of these methods?

Something along the line of 

```go
h.l.Info("Unable to parse collectd types file", zap.String("path", path))
```

services/collectd/service.go
```diff
@@ -53,13 +52,28 @@ func TypesDBFile(path string) (typesdb *api.TypesDB, err error) {
...
+	Listening(addr net.Addr)
+	Closed()
+	UnableToReadDirectory(path string, err error)
+	LoadingPath(path string)
+	TypesParseError(path string, err error)
+
+	ReadFromUDPError(err error)
+	ParseError(err error)
+	DroppingPoint(name string, err error)
+	InternalStorageCreateError(err error)
```
desa: I really like how having specific errors like this. I know I said it previously, just felt like saying it again.

services/collectd/service.go
```diff
@@ -194,7 +214,9 @@ func (s *Service) Open() error {
 	}
 	s.conn = conn
 
-	s.Logger.Info(fmt.Sprint("Listening on UDP: ", conn.LocalAddr().String()))
+	if s.Diagnostic != nil {
```
desa: I'm really not a fan of these nil checks, Isn't it possible that they old line could have panicked in the same way that theses are trying to guard against?

cmd/influxd/run/server.go
jsternberg: I think we should be upgrading to a newer version, but I thought this change would get unwieldy if I did it right here. The change is already decently hard to track and I've had a bad track record when I try to mix moving refactors with other stuff.

cmd/influxd/run/server.go
jsternberg: I like this.

coordinator/points_writer.go
jsternberg: The primary reason for having these nil checks everywhere is to avoid making a nil version of the interface for every package and attempting to use it as a default. The old code didn't need it because I set the logger to the null logger. I'm removing that instead and using this pattern.

diagnostic/handlers.go
jsternberg: No reason at the moment. Right now, I'm just doing pretty straight translation. We only ever half did structured logging in InfluxDB. I moved the code to a structured logger, but used `Sprintf` so basically nothing else changed. The intention was to then have follow up commits that moved to structured logging, but then we couldn't figure out how we wanted to proceed on the logging front.

cmd/influxd/run/server.go
desa: This meaning having the `In` prefix? I don't feel strongly one way or another 👍

jsternberg: Oh no, I'm sorry. Removing the `In` prefix.

coordinator/points_writer.go
desa: I see. Maybe have a Diagnostic thats local to this package that does something like this

```go
type LocalDiag struct {
Diag Diagnostic
}

func (d *LocalDiag) DiagMethod(...) {
if d == nil {
return
}

d.Diag.DiagMethod(...)
}
```

So that we can avoid having these nil checks everywhere. I don't feel super strongly about it, but having 

```
if d.diag != nil {
}
```
everywhere definitely is an eye sore

diagnostic/handlers.go
desa: I see. That makes sense.

jsternberg:
I've updated with the various changes we've discussed both on this issue and the ones over chat. This is ready for review again.

It still needs some additional work, but this seems like a good jumping off points.

jsternberg:
I've updated this to update the version of zap so we no longer are using a non-existant hash of zap.

The console output is used as the default and it has been updated upstream to look like this:

```

 8888888           .d888 888                   8888888b.  888888b.
   888            d88P"  888                   888  "Y88b 888  "88b
   888            888    888                   888    888 888  .88P
   888   88888b.  888888 888 888  888 888  888 888    888 8888888K.
   888   888 "88b 888    888 888  888  Y8bd8P' 888    888 888  "Y88b
   888   888  888 888    888 888  888   X88K   888    888 888    888
   888   888  888 888    888 Y88b 888 .d8""8b. 888  .d88P 888   d88P
 8888888 888  888 888    888  "Y88888 888  888 8888888P"  8888888P"

2017-10-17T16:04:24Z    info    InfluxDB starting, version unknown, branch unknown, commit unknown
2017-10-17T16:04:24Z    info    Go version go1.9.1, GOMAXPROCS set to 8
2017-10-17T16:04:24Z    info    no configuration provided, using default settings
2017-10-17T16:04:24Z    info    Using data dir: /Users/jsternberg/.influxdb/data        {"service": "store"}
2017-10-17T16:04:24Z    info    reading file /Users/jsternberg/.influxdb/wal/_internal/monitor/1/_00001.wal, size 923   {"engine": "tsm1", "service": "cacheloader"}
2017-10-17T16:04:24Z    info    reading file /Users/jsternberg/.influxdb/wal/_internal/monitor/1/_00004.wal, size 0     {"engine": "tsm1", "service": "cacheloader"}
2017-10-17T16:04:24Z    info    /Users/jsternberg/.influxdb/data/_internal/monitor/1 opened in 2.703438ms       {"service": "store"}
2017-10-17T16:04:24Z    info    opened service  {"service": "subscriber"}
```

I'm also working on logfmt integration so we can have the logger be configurable.

CC @rbetts.

cmd/influxd/run/server.go
```diff
@@ -303,6 +312,7 @@ func (s *Server) appendGraphiteService(c graphite.Config) error {
 	srv.PointsWriter = s.PointsWriter
 	srv.MetaClient = s.MetaClient
 	srv.Monitor = s.Monitor
+	//srv.With(s.DiagnosticService.GraphiteContext())
```
desa: Any reason why this is commented out?

diagnostic/README.md
```diff
@@ -0,0 +1,69 @@
+# Diagnostic Package
```
desa: :1st_place_medal: I really like this doc

query/task_manager.go
```diff
@@ -174,15 +173,14 @@ func (t *TaskManager) AttachQuery(q *influxql.Query, database string, interrupt
 	t.queries[qid] = query
 
 	go t.waitForQuery(qid, query.closing, interrupt, query.monitorCh)
-	if t.LogQueriesAfter != 0 {
+	if t.LogQueriesAfter != 0 && t.Diagnostic != nil {
```
desa: I know it's mostly obvious whats going on here, but it might be useful to note why we're checking if the diagnostic is nil here. I could see that as possibly being confusing if someone didn't have the context we do.

cmd/influxd/run/server.go
jsternberg: Likely a mistake. I'll fix that.

jsternberg:
@desa about the nil checks, maybe I should try wrapping the interface in a struct? The main problem is that interfaces have no default implementation they can use. But structs can be null so you can have something like this:

```
type Context struct {
    Handler Handler
}

type Handler interface {
    Info(msg string)
}

func (c *Context) Info(msg string) {
    if c != nil && c.Handler != nil {
        c.Handler.Info(msg)
    }
}
```

Then you can use `Context` as a pointer in every struct and it will operate correctly.

jsternberg:
So I'll try to address the more technical questions. I can't really address the idea behind why we're converting to this method other than "Kapacitor is the model and Kapacitor decided on this." The idealogical argument is that we need to have some method where all of the products log using the same kind of pattern and with the same logger so there is some uniformity to the logs themselves. The Kapacitor team got the privilege/responsibility to do it and InfluxDB is just following. Some of the stuff in this PR isn't in Kapacitor, but follows with the same pattern as we figure out what works for every product. The end goal is to then decide on a specific logging implementation that will be shared between the products. Most likely, everything is going to be output in logfmt as a default.

1. Yes. This is unfortunately the case. At the same time, it abstracts away the logger itself and makes it easier for us to potentially add i18n support for logs in the future because what is logged and the act of logging are separated from each other. While it is an initial pain, the architecture lends itself better to long term maintenance. As an example, updating the zap logger version was much easier since I only had to look inside of the global `diagnostic` package rather than go hunting everywhere.

2. This seems like a good idea. I can do this.

3. Anonymous interfaces result in a lot of duplication and also don't play well with nested types. You need to reference the interface in the global `diagnostic` package along with the interface in the package that is logging. So you would have something like this:

```
// query/query_executor.go
type QueryExecutor interface {
    Diagnostic interface {
        PrintStatement(query string)
        QueryPanic(query string, err interface{}, stack []byte)
    }
}

// diagnostic/query_handler.go
func (s *Service) QueryContext() interface {
    PrintStatement(query string)
    QueryPanic(query string, err interface{}, stack []byte)
} {
    return ...
}
```

Then there's also the problem of a default interface. We're potentially moving to using a struct with a single interface inside of it so the struct can have a default zero value of logging nothing.

jwilder:
> The idealogical argument is that we need to have some method where all of the products log using the same kind of pattern and with the same logger so there is some uniformity to the logs themselves. 

The format and style in the way we log is orthogonal to much of what is changed in this PR though.  The changes here are more complicated than they need to be to only achieve a consistent log format between products.  

For example, instead of having `zap.Logger` (a hard external dependency) in each module, replace that with a generic logger interface (e.g. `influxdb.Logger`) that is used by each module.  We then have one implementation of that interface that is based on `zap` and injected just as we have now.

This addresses several things:

1) It's a smaller change and easier to digest/review.
2) The dependency on `zap.Logger` everywhere is reduced to one place which makes is easier to change.
3) It avoids the subtle addition of extension points everywhere which is very dangerous if the module author did not originally intend an extension point to exist there.  The interface is that of a logger which is clearer.

jsternberg:
I'm going to close this and not delete the branch. I don't think this is something we're going to do so I'll remove the clutter.

PR #8968, opened 2017-10-16
Make client errors more helpful on downstream errs

### LINKED ISSUE #8966: Client obscures errors due from HTTP proxies/loadbalancer
### Bug report

__System info:__ influxdb 1.2.x, Linux

__Steps to reproduce:__

1. Setup nginx proxy to influxdb's HTTP endpoint
2. Don't actually run influxdb (so the requests to the proxy fails with 503 or similar)
3. Make a request via the client to the proxy

__Expected behavior:__ Error message indicating an issue with communicating with the influxdb server

__Actual behavior:__ Error regarding json decoding"invalid character '<'"

__Additional info:__
The main issue here is one of clarity - the current error handling in the HTTP client makes a number of assumptions that aren't always true. namely:

 * the response is JSON
 * the response is from an influxdb server

If you are running influxdb behind a proxy, loadbalancer or some kind of application firewall, any one of those things may be introducing errors and responding as the influxdb server instead of passing the request on to it. In these cases, the client should return an error that indicates that the error did not originate with the influxdb server itself but rather is an downstream problem.

darkliquid:
When a downstream server such as a proxy or loadbalancer between
influxdb and the client produces an error, the client currently does
not make this very obvious.

This change introduces checks on both the content type and the
influx version header to identify whether a request was served by
influxdb itself and returns a more appropriate error in the cases
where it can be determined a downstream issue is at play.

Fixes #8966

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/client/v2/client.go
+++ b/client/v2/client.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"mime"
 	"net/http"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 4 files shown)


e-dard:
@darkliquid has signed the CLA.

client/v2/client.go
```diff
@@ -534,7 +548,7 @@ func (c *client) Query(q Query) (*Response, error) {
 				break
 			}
 		}
-	} else {
+	} else if strings.Contains(resp.Header.Get("Content-Type"), "application/json") {
```
e-dard: I'm not sure about needing to separate out unmarshaling a body that comes from a response with an `application/json` content type, and dealing with a body that comes from a response with a different content type.

We could just attempt to unmarshal the body into `dec`, and if there is an error then return that, along with the body.

e-dard:
@darkliquid needs a `CHANGELOG` entry too.

client/v2/client.go
darkliquid: The reasoning behind separating it out is two-fold:

1. downstream error pages (maybe your proxy does auth and throws a 403 non-json response) gets handled without an explicit 500+ status code check
2. saves wasting time dropping into the json decode machinery and allows us to fail early _and_ keep the body reader content - dropping into the decoder will read from the body reader into a private buffer inside the decoder we can't access, so we can't return any error messages it may contain.

e-dard: Sounds reasonable 👍

darkliquid:
Hows this looking now? Any further changes required?

e-dard:
@darkliquid looks good, thanks! I just need to poke someone else on the team to take a look and then I'll merge it.

client/v2/client.go
jsternberg: This check should be for the response as a whole. A chunked response should have this check too.

You also don't want to use `strings.Contains` here. You will want to use this: https://godoc.org/mime#ParseMediaType

So something like:

```
var contentType string
if header := resp.Header.Get("Content-Type"); header != "" {
    contentType, _, _ = mime.ParseMediaType(header)
}

if contentType != "application/json" {
    // the code you added
}

// the normal code that was already there
```

darkliquid: Ahh yes, good point. I'll fix that shortly and resubmit.

darkliquid:
@jsternberg how is this looking now? I _think_ I've addressed everything.

e-dard:
Thanks @darkliquid!

PR #8971, opened 2017-10-17
Create a command to truncated shard groups

andrewhare:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


CHANGELOG.md
```diff
@@ -38,6 +38,7 @@
 - [#8897](https://github.com/influxdata/influxdb/pull/8897): Add message pack format for query responses.
 - [#8886](https://github.com/influxdata/influxdb/pull/8886): Improved compaction scheduling
 - [#8690](https://github.com/influxdata/influxdb/issues/8690): Implicitly decide on a lower limit for fill queries when none is present.
+- [#7355](https://github.com/influxdata/influxdb/issues/7355): Create a command to truncated shard groups
```
e-dard: `to truncate shard group`

services/meta/data.go
```diff
@@ -746,6 +746,32 @@ func (data *Data) UnmarshalBinary(buf []byte) error {
...
+	for i := range data.Databases {
+		dbi := &data.Databases[i]
+
+		for j := range dbi.RetentionPolicies {
+			rpi := &dbi.RetentionPolicies[j]
+
+			for k := range rpi.ShardGroups {
+				sgi := &rpi.ShardGroups[k]
+
+				if !t.Before(sgi.EndTime) || sgi.Deleted() || (sgi.Truncated() && sgi.TruncatedAt.Before(t)) {
```
e-dard: It took me a little while to wrap my head around this logic; maybe a little commentary could have helped.

services/meta/data_test.go
```diff
@@ -215,6 +215,104 @@ func TestData_SetPrivilege(t *testing.T) {
 	}
 }
 
+func TestData_TruncateShardGroups(t *testing.T) {
+	data := &meta.Data{}
+
+	must := func(err error) {
```
e-dard: I think I'd rather see this as `assertNil` or something. `must` usually implies that if there is an error you `panic`.

e-dard: Another approach we often take in the database (though it seems overkill for one test) is to create a `Data` type in the test, and then write methods like `MustCreateDatabase`, `MustCreateShardGroup` etc.

PR #8975, opened 2017-10-17
Copy returned bytes from TSI meta functions.

benbjohnson:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/tsdb/index/tsi1/tsi1.go
+++ b/tsdb/index/tsi1/tsi1.go
@@ -792,25 +792,6 @@ func (a byteSlices) Len() int           { return len(a) }
 func (a byteSlices) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
 func (a byteSlices) Less(i, j int) bool { return bytes.Compare(a[i], a[j]) == -1 }
 
-// copyBytes returns a copy of b.
-func copyBytes(b []byte) []byte {
-	if b == nil {
-		return nil
-	}
-	buf := make([]byte, len(b))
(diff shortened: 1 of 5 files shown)


tsdb/index/tsi1/tsi1.go
```diff
@@ -802,6 +802,15 @@ func copyBytes(b []byte) []byte {
 	return buf
 }
 
+// copyBytes returns a copy of a slice of byte slices.
+func copyByteSlices(a [][]byte) [][]byte {
```
e-dard: I don't have any qualms about this being here, but it could also live happily alongside `Union(a, b [][]byte) [][]byte` and `Intersect(a, b [][]byte) [][]byte` in the `bytesutil` package. Maybe as `Clone(a [][]byte) [][]byte`?

benbjohnson: Good point. I moved these up to `bytesutil` in d17d0f1.

benbjohnson:
@e-dard I added comments about the byte copying in 8ad2048. The slices just need to be copied if they will be used after the fileset is released. Within the `FileSet.Retain()` & `FileSet.Release()` the files are guaranteed to stay memory-mapped.

e-dard:
@benbjohnson ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ ⭐️ DoUbLe ApProVE ⭐️ ⭐️ ⭐️ ⭐️ ⭐️

PR #8980, opened 2017-10-18
Prevent deadlock during collectd, graphite, opentsdb, and udp shutdown

### LINKED ISSUE #8780: Clean shutdown impossible with collectd plugin
influxdb 1.3.2 on linux

When I start influxd and send a `SIGTERM` to it the following output is shown
```
[I] 2017-09-02T13:02:46Z Listening on UDP: [::]:25826 service=collectd
[I] 2017-09-02T13:02:46Z Listening for signals
[I] 2017-09-02T13:03:38Z Signal received, initializing clean shutdown...
[I] 2017-09-02T13:03:38Z Waiting for clean shutdown...
[I] 2017-09-02T13:03:38Z shutting down monitor system service=monitor
[I] 2017-09-02T13:03:38Z snapshot listener closed service=snapshot
[I] 2017-09-02T13:03:38Z continuous query service terminating service=continuous_querier
[I] 2017-09-02T13:03:38Z collectd ReadFromUDP error: read udp [::]:25826: use of closed network connection service=collectd

[I] 2017-09-02T13:04:08Z time limit reached, initializing hard shutdown
```
I don't think this should happen and influxd should be able to cleanly shutdown with collectd enabled.
Looking through the old/existing issues, this might be similar to #4016

jsternberg:
All of these services start up goroutines and then wait for the
goroutines to finish. Each of them has a `tsdb.PointBatcher` that may
return a point during the shutdown sequence. During the shutdown
sequence, a lock was held. This lock may get accessed when attempting to
write the point that came back from the `tsdb.PointBatcher`. This caused
the read lock attempt to wait forever for the write lock to be unlocked
during `Close()`.

This modifies these methods so that the write lock is released while
waiting for goroutines to finish in these three services.

Fixes #8780.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/services/collectd/service.go
+++ b/services/collectd/service.go
@@ -95,7 +95,7 @@ func (s *Service) Open() error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	if !s.closed() {
+	if s.done != nil {
 		return nil // Already open.
 	}
 	s.done = make(chan struct{})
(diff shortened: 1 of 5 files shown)


services/collectd/service.go
```diff
@@ -211,24 +211,34 @@ func (s *Service) Open() error {
 
 // Close stops the service.
 func (s *Service) Close() error {
-	s.mu.Lock()
-	defer s.mu.Unlock()
+	if wait := func() bool {
```
e-dard: Just a thought: this anonymous function only returns `false` in one place, and with that `false` value we then return from `Close()`.  Seems like it would be fewer lines of code and the block one would be indented less if we just add the appropriate `s.mu.Unlock()` calls before the early return, and remove the anonymous function.

jsternberg: The reason why I chose to do it this way was to take advantage of defer. I though there was enough going on there and enough function calls that protecting against a panic was a decent idea. Do you think I should get rid of the defer though?

e-dard: I don't mind to be honest. The way it is now involves four extra lines, while adding in two `s.mu.Unlock()` calls is only two lines. After I made that comment I noticed other cases of the same pattern where it made a lot of sense (because there were errors involved) so I think either approach would be reasonable. 👍

Issue #8937, opened 2017-10-09
Data spontaneously dropped from shard, retention policy possibly fails


daviskirk:
### Bug report

__System info:__ influxdb:1.3.3, running in docker persisted via host mount

__Steps to reproduce:__

1. Write data to default database (starting using influx 0.10)
2. Upgrade database and continue to write 
3. Change shard duration from 1 week to 3 months
4. Change retention policy to 3650d
5. Continue writing

__Expected behavior:__

Writes continue and data is persisted for 10 years in line with the retention policy (and the expiration date of the shards).  

__Actual behavior:__ 

* 50% of the series from a shard (`378 default      autogen          2017-01-16T00:00:00Z 2017-04-10T00:00:00Z 2027-04-08T00:00:00Z`) are no longer queryable (there is a 3 month hole in all the data) 
* Writes continue to succeed (for the other 50% of time series)
* The shard itself was not empty (300MB vs. 550MB before)
* In the client code there were no "drop shard" commands used, `delete from` is used with specific time series (but never for exactly the shard period) 

__Additional info:__ 

The relevant part of the output of `show shard groups`:

```
220 default      autogen          2016-12-19T00:00:00Z 2016-12-26T00:00:00Z 2026-12-24T00:00:00Z
197 default      autogen          2016-12-26T00:00:00Z 2017-01-02T00:00:00Z 2026-12-31T00:00:00Z
196 default      autogen          2017-01-02T00:00:00Z 2017-01-09T00:00:00Z 2027-01-07T00:00:00Z
195 default      autogen          2017-01-09T00:00:00Z 2017-01-16T00:00:00Z 2027-01-14T00:00:00Z
293 default      autogen          2017-01-16T00:00:00Z 2017-01-23T00:00:00Z 2027-01-21T00:00:00Z
378 default      autogen          2017-01-16T00:00:00Z 2017-04-10T00:00:00Z 2027-04-08T00:00:00Z
432 default      autogen          2017-04-10T00:00:00Z 2017-07-03T00:00:00Z 2027-07-01T00:00:00Z
433 default      autogen          2017-07-03T00:00:00Z 2017-09-25T00:00:00Z 2027-09-23T00:00:00Z
563 default      autogen          2017-09-25T00:00:00Z 2017-12-18T00:00:00Z 2027-12-16T00:00:00Z
564 default      autogen          2017-12-18T00:00:00Z 2018-03-12T00:00:00Z 2028-03-09T00:00:00Z
```

[...]

PR #8981, opened 2017-10-19
Fix data being dropped during compactions

### LINKED ISSUE #8822: Lost all datas after DELETE command
### Bug report
Influxdb version = 1.3.5
On linux go1.8.3
Intel(R) Xeon(R) CPU E5-2620 with 64Go of RAM

Sometimes, when we try to delete somes points on a specific measurement , we lost 2 or 3 days of datas in our database. ( all measurements are impacted ! )

Exemple of DELETE query  : 

DELETE FROM events WHERE application = 'cosmos' AND time < 1505223439180000001 AND time > 1505223439179999999

Same issue with :

DELETE FROM events WHERE application = 'cosmos' AND time = 1505223439180000000

You can find a extract log when the issue appear with many messages about full compaction aborted... 
If you need a more details logs , tell me.
We have many application which write data in parallel.

[delete.zip](https://github.com/influxdata/influxdb/files/1298810/delete.zip)

jwilder:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

This fixes #8822 and #8937.  

If multiple tombstones exists for a series that ended up causing the
full data to be deleted, the blocks were not removed from the offsets
in the TSM index.  This causes the TSMReader to report that a key exist
but does not have any data which should never happen.

During a compaction, every key should have at least one value.  Since
this invariant was broken, the compaction aborted early and ends up
dropping all series keys that are lexigraphically greater than where
the breakage occured.  This would cause data to be dropped during the
compaction.

To repro this in a live server you can run the following script:

First start the server with:
```
INFLUXDB_DATA_CACHE_SNAPSHOT_WRITE_COLD_DURATION=1s influxd
```

Then run:

```bash
#!/bin/bash

influx -execute "drop database test; create database test"

curl -v "http://localhost:8086/write?db=test" --data-binary 'cpu value=1 0
cpu value=1 1
mem value=1 1
zzz value=1 2
aaa value=1 3'

echo "Waiting 2 seconds for snapshot"
sleep 2

echo "All data is there"
influx -database test -execute "select * from /.*/"

echo "Delete cpu series by exact times"
influx -database test -execute "delete from cpu where time >= 0 and time <= 0; delete from cpu where time >= 1 and time <= 1;"

echo "Waiting for compaction"
sleep 2
echo "All series lexicographically after aaa are gone"

influx -database test -execute "select * from /.*/"
```

This fixes a potential bug where the BlockIterator would skip blocks
if the underlying TSMReader had deletes on it concurrently.  This
could possibly occur due to changes in 91eb9de that now use the
existing TSMReaders from the FileStore instead of creating new ones
during compaction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,7 @@
 - [#8952](https://github.com/influxdata/influxdb/issues/8952): Fix WAL panic: runtime error: makeslice: cap out of range
 - [#8975](https://github.com/influxdata/influxdb/pull/8975): Copy returned bytes from TSI meta functions.
 - [#7797](https://github.com/influxdata/influxdb/issues/7706): Fix data deleted outside of time range
+- [#8822](https://github.com/influxdata/influxdb/issues/8822): Fix data dropped incorrectly during compaction
 
 ## v1.3.4 [unreleased]
 
--- a/tsdb/engine/tsm1/compact.go
+++ b/tsdb/engine/tsm1/compact.go
@@ -1045,6 +1045,11 @@ func (c *Compactor) write(path string, iter KeyIterator) (err error) {
 		}
 	}
 
+	// Were there any errors encountered during iteration?
+	if err := iter.Err(); err != nil {
+		return err
+	}
+
 	// We're all done.  Close out the file.
 	if err := w.WriteIndex(); err != nil {
 		return err
@@ -1089,6 +1094,9 @@ type KeyIterator interface {
 
 	// Close closes the iterator.
 	Close() error
+
+	// Err returns any errors encountered during iteration.
+	Err() error
 }
 
 // tsmKeyIterator implements the KeyIterator for set of TSMReaders.  Iteration produces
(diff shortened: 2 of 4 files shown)


stuartcarnie:
Are these scenarios ok:

Insert data

```
cpu,tagkey=v0 value=0 1500
cpu,tagkey=v0 value=0 1800
cpu,tagkey=v0 value=0 2000
```

### Scenario 1

Delete all the individual points

```
delete from cpu where tagkey=v0 and time=1500
delete from cpu where tagkey=v0 and time=1800
delete from cpu where tagkey=v0 and time=2000
```

### Scenario 2

Delete outer points

```
delete from cpu where tagkey=v0 and time=1500
delete from cpu where tagkey=v0 and time=2000
```

jwilder:
@stuartcarnie Scenario 1 is covered in `TestTSMReader_MMAP_TombstoneMultipleRangesFull` Scenario 2 is covered in `TestTSMReader_MMAP_TombstoneMultipleRangesNoOverlap`

PR #8990, opened 2017-10-20
Code cleanup for removing the pid file

jsternberg:


jsternberg:
@e-dard I pushed the previous commit to this without realizing I had a fixup for some formatting issues in the previous one. Can you take a quick look at this?

PR #8995, opened 2017-10-20
Sort & validate TSI key value insertion.

benbjohnson:
## Overview 

The `TagBlockEncoder` value sorting was removed at some point and the underlying input was done via a range over a map which has unpredictable sorting. This fixes the sorting and adds additional validation to ensure inputs are always ordered when compacting index files.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/tsdb/index/tsi1/tag_block.go
+++ b/tsdb/index/tsi1/tag_block.go
@@ -479,7 +479,8 @@ type TagBlockEncoder struct {
 	trailer TagBlockTrailer
 
 	// Track tag keys.
-	keys []tagKeyEncodeEntry
+	keys      []tagKeyEncodeEntry
+	prevValue []byte
 }
 
 // NewTagBlockEncoder returns a new TagBlockEncoder.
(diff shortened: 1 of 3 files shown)


tsdb/index/tsi1/tag_block.go
```diff
@@ -539,6 +543,13 @@ func (enc *TagBlockEncoder) EncodeValue(value []byte, deleted bool, seriesIDs []
 		return fmt.Errorf("zero length tag value not allowed")
 	}
 
+	// Validate that keys are in-order.
```
e-dard: How often will we call `EncodeValue`? I only ask as this `bytes.Compare` call could be an expensive way to test the encoder is working properly. Or is returning an error a "normal path" for this encoder?

benbjohnson: It's called once for every tag value in the index. I think the cost should be relatively small compared to the merging and writing. We can remove it if it becomes a performance bottleneck but I think it'd be good to keep it in the short-term at least to ensure we don't store unsorted data.

Issue #8986, opened 2017-10-19
Buffer overflow occours when a large compressed file is imported to influx


lets00:
### Description

I needed export my data using influx_inspect program. I used this command to generate a output file on line protocol format.
```
influx_inspect export -database monasca -datadir monasca_influx/data/ -retention autogen -waldir monasca_influx/wal -out monasca.dmp -start 2017-10-08T00:00:00Z -end 2017-10-08T23:59:59Z -compress
```

This file has 950MB compressed, when I try to import it using influx command (```influx -import -compressed -path=monasca.dmp```) the error occours:

```
ERROR: reading standard input: bufio.Scanner: token too long
```

### Bug report

__System info:__ 
Ubuntu 16.04
Influx 1.3.5 and Influx 1.1.0
Database full
24GB RAM 

__Steps to reproduce:__

1. Export database to line protocol file in compress mode (the output file must have more than 600MB)
2. Try to import this database using influx import command

__Expected behavior:__ 

The line protocol series is added on influxdb.

__Actual behavior:__ 

Nothing is added on influxdb.

__Additional info:__ 

I look at the code (https://github.com/influxdata/influxdb/blob/master/importer/v8/importer.go#L95-L112), When the file is uncompressed, the final file size is probably larger than the available RAM in the system. Chunk the decompression process on smaller parts of the file maybe solves the problem (e.g. read 1024 bytes of file, decompress it, add on the db, and so on).

e-dard:
@lets00 this is probably our fault for the  implementation we used to scan import files. I think you have one or more particularly long files in your import file.

If you're on `Ubuntu` can you try `$ wc -L monasca.dmp` and see how long the longest line is? 

In order for the importer to accept this we'll need to fix the importer, or you'll need to remove the long lines from the input (which isn't particularly practical).

lets00:
@e-dard Thank you for answer me. I ran the command, the output is described below:
```
$ wc -L monasca.dmp
1678 monasca.dmp
```

The max line lenght in this file is 1678. The problem in remove the long lines is that, uncompressed, this file is extremely huge (1 GB becomes some like 200GB). And try to edit manually all long line is impractical.

e-dard:
@lets00 is that the line length of the compressed file? I probably should have been more specific that it would be interesting to know the max size of a line in the uncompressed (line protocol) format. If the file is too big to handle then don't worry about that.

[...]

PR #8999, opened 2017-10-23
Replace NewScanner to NewReader to fix bug #8986

lets00:
Replace NewScanner to NewReader to fix bug #8986 

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


CHANGELOG.md
```diff
@@ -82,6 +82,7 @@
 - [#8822](https://github.com/influxdata/influxdb/issues/8822): Fix data dropped incorrectly during compaction
 - [#8780](https://github.com/influxdata/influxdb/issues/8780): Prevent deadlock during collectd, graphite, opentsdb, and udp shutdown.
 - [#8983](https://github.com/influxdata/influxdb/issues/8983): Remove the pidfile after the server has exited.
+- [#8986](https://github.com/influxdata/influxdb/issues/8986): Replace NewScanner to NewReader in importer.go file.
```
e-dard: How about `Add long-line support to client importer` ?

importer/v8/importer.go
```diff
@@ -126,9 +126,9 @@ func (i *Importer) Import() error {
 	i.processDML(scanner)
 
 	// Check if we had any errors scanning the file
-	if err := scanner.Err(); err != nil {
-		return fmt.Errorf("reading standard input: %s", err)
-	}
+	//if err := scanner.Err(); err != nil {
```
e-dard: Please remove this commented out code.

importer/v8/importer.go
```diff
@@ -144,42 +144,49 @@ func (i *Importer) Import() error {
 	return nil
 }
 
-func (i *Importer) processDDL(scanner *bufio.Scanner) {
-	for scanner.Scan() {
-		line := scanner.Text()
+func (i *Importer) processDDL(scanner *bufio.Reader) {
+	for {
```
e-dard: There are a couple of issues here. (1) you could use `ReadString(byte('\n'))` instead of `ReadLine()`. That way you don't need to litter `string(line)` everywhere. Also, I think `ReadLine()` can return short reads, in that it might not return an entire line. 

The second thing is you're not checking for errors that are not nil, but also not `io.EOF`. Those errors would now need to be returned to the caller too.

How about we go with something like this?

```go
func (i *Importer) processDDL(scanner *bufio.Reader) error {
    for {
        line, err := scanner.ReadString(byte('\n'))
        if err != nil && err != io.EOF {
            return err
        }

	// If we find the DML token, we are done with DDL
        if strings.HasPrefix(line, "# DML") {
            return nil
        }

        if strings.HasPrefix(line, "#") {
            continue
        }

        // Skip blank lines
        if strings.TrimSpace(line) == ""  {
            continue
        }
        
        i.queryExecutor(string(line))

        // Check if this was the final token.
        if err == io.EOF { 
            return nil
        }
    }
    return nil
}
```

importer/v8/importer.go
```diff
@@ -144,42 +144,49 @@ func (i *Importer) Import() error {
...
 
-func (i *Importer) processDML(scanner *bufio.Scanner) {
+func (i *Importer) processDML(scanner *bufio.Reader) {
 	start := time.Now()
-	for scanner.Scan() {
-		line := scanner.Text()
-		if strings.HasPrefix(line, "# CONTEXT-DATABASE:") {
-			i.database = strings.TrimSpace(strings.Split(line, ":")[1])
+	for {
+		line, _, err := scanner.ReadLine()
```
e-dard: You should take a similar approach to the one I suggested previously in the review.

importer/v8/importer.go
e-dard: Don't forget you'll need to add error handling to the call to `processDDL()` inside of `Import()`.

lets00:
I'm working on the proposed changes. Thank you for the feedback.

e-dard:
I can confirm @lets00 has signed the CLA.

importer/v8/importer.go
```diff
@@ -112,7 +112,10 @@ func (i *Importer) Import() error {
 	scanner := bufio.NewReader(r)
 
 	// Process the DDL
-	i.processDDL(scanner)
+	ddl_err := i.processDDL(scanner)
```
e-dard: `if err := i.processDDL(scanner); err != nil {`

importer/v8/importer.go
```diff
@@ -112,7 +112,10 @@ func (i *Importer) Import() error {
...
+	if ddl_err != nil {
+		return fmt.Errorf("reading standard input: %s", ddl_err)
```
e-dard: `err` not `ddl_err`

importer/v8/importer.go
```diff
@@ -123,12 +126,10 @@ func (i *Importer) Import() error {
 	i.lastWrite = time.Now()
 
 	// Process the DML
-	i.processDML(scanner)
-
-	// Check if we had any errors scanning the file
-	//if err := scanner.Err(); err != nil {
-	//	return fmt.Errorf("reading standard input: %s", err)
-	//}
+	dml_err := i.processDML(scanner)
```
e-dard: `if err := i.processDML(scanner); err != nil {`

importer/v8/importer.go
```diff
@@ -123,12 +126,10 @@ func (i *Importer) Import() error {
...
+	if dml_err != nil {
+		return fmt.Errorf("reading standard input: %s", dml_err)
```
e-dard: `err` not `ddl_err`

importer/v8/importer.go
```diff
@@ -144,52 +145,61 @@ func (i *Importer) Import() error {
...
 		}
-		i.batchAccumulator(string(line), start)
+		i.batchAccumulator(line, start)
+		//End of File
+		if err == io.EOF {
+			return nil
+		}
 	}
 	// Call batchWrite one last time to flush anything out in the batch
 	i.batchWrite()
```
lets00: Maybe put this on if err ==io.EOF statement

lets00:
I saw a infinite loop scenario when i.batchAccumulator() run after end of file. This not happen before(witch NewScanner), because the "for" statement not execute it. For fix this, I put, io.EOF verification before i.batchAccumulator(). I will open a new pull request to fix this. Sorry about that.

e-dard:
@lets00 no need for a new pull request, just push a commit up to this one 😄

lets00:
@e-dard I fix the problem putting ```if io.EOF...``` statement below the ```if err != nil ...``` statement in both the ```processDDL``` and ```processDML``` methods. I test in my environment, importing a ```monasca.dmp``` database and this solution works for me.

importer/v8/importer.go
```diff
@@ -160,17 +166,27 @@ func (i *Importer) processDDL(scanner *bufio.Scanner) {
...
+		}
+		//End of file
+		if err == io.EOF {
+			// Call batchWrite one last time to flush anything out in the batch
+			i.batchWrite()
+			return nil
+		}
 		if strings.HasPrefix(line, "# CONTEXT-DATABASE:") {
-			i.database = strings.TrimSpace(strings.Split(line, ":")[1])
+			i.database = strings.TrimSpace(strings.Split(string(line), ":")[1])
```
e-dard: You don't need `string(line)` since `line` is already a `string`.

importer/v8/importer.go
```diff
@@ -160,17 +166,27 @@ func (i *Importer) processDDL(scanner *bufio.Scanner) {
...
 		}
 		if strings.HasPrefix(line, "# CONTEXT-RETENTION-POLICY:") {
-			i.retentionPolicy = strings.TrimSpace(strings.Split(line, ":")[1])
+			i.retentionPolicy = strings.TrimSpace(strings.Split(string(line), ":")[1])
```
e-dard: You don't need `string(line)` since `line` is already a `string`.

importer/v8/importer.go
```diff
@@ -160,17 +166,27 @@ func (i *Importer) processDDL(scanner *bufio.Scanner) {
...
-func (i *Importer) processDML(scanner *bufio.Scanner) {
+func (i *Importer) processDML(scanner *bufio.Reader) error {
 	start := time.Now()
-	for scanner.Scan() {
-		line := scanner.Text()
+	for {
+		line, err := scanner.ReadString(byte('\n'))
+		if err != nil && err != io.EOF {
+			return err
+		}
```
e-dard: This can be an `} else if err == io.EOF {`.

importer/v8/importer.go
```diff
@@ -144,12 +143,19 @@ func (i *Importer) Import() error {
...
 
-func (i *Importer) processDDL(scanner *bufio.Scanner) {
-	for scanner.Scan() {
-		line := scanner.Text()
+func (i *Importer) processDDL(scanner *bufio.Reader) error {
+	for {
+		line, err := scanner.ReadString(byte('\n'))
+		if err != nil && err != io.EOF {
+			return err
+		}
```
e-dard: This can be an `} else if err == io.EOF {`.

PR #2776, opened 2015-06-04
Enforce retention policies

otoolep:
This change adds a new service to the system for Retention Policy enforcement.

Enforcement is composed of two parts: 

- a goroutine, that only runs on the leader, which periodically walks all `ShardGroups` and deletes any `ShardGroups` that have expired. Note that with this change ShardGroups are not actually deleted. Instead they have a tombstone flag set. This is so it is explicit that a `ShardGroup`, and its associated shards, can be safely deleted.
- a second goroutine that runs on every node, which checks each shard in the TSBD store against the list of shards which should actually exist. If the shard is part of a tombstoned `ShardGroup` it is deleted.

PR #9017, opened 2017-10-26
Ensure retention service removes shards locally 

### LINKED ISSUE #8819: Key space is not cleanup up by retention policy
affected version: influx 1.3.1

We are trying to use influxdb in a memory limited environment. Our data scheme requires to assign a medium number (1000 to 10000) of tag values.

We already found out that this is not optimal regarding the memory usage of influxdb.

Our current attempt is to limit the number of series with max-series-per-database = 10000.
So far, it works as expected. Whenever we hit the limit, new tag values are dropped during import, which is what we try to achieve.

But the key-space of the tag values seems to be not released by the retention policies. Even when the series should have been dropped we cannot insert new entries.
As a result the database looks empty (doing a select * query) but we cannot insert new elements.

This issue can be solved by restarting influxd. After restart we can insert new tag values again.

Any idea how we can overcome this?

e-dard:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

Fixes #8819.

As well as the main fix, this PR adds a generic `tsdb.Store` mock to the `internal` package, which can be used in other parts of the codebase. It also adds some initial testing to the Retention Service, as well as a repro test for #8819.

Previously, the process of dropping expired shards according to the retention policy duration, was managed by two independent goroutines in the retention policy service. This behaviour was introduced in #2776, at a time when there were both data and meta nodes in the OSS codebase.
The idea was that only the leader meta node would run the meta data deletions in the first goroutine, and all other nodes would run the local deletions in the second goroutine.

InfluxDB no longer operates in that way, and so we ended up with two independent goroutines that were carrying out an action that was in reality dependent on each other.

If the second goroutine (local deleter) runs before the first (meta deleter), then it may not have yet seen the meta data changes indicating which shards should be deleted. Therefore, it won't
delete any of those shards locally. Shortly after this, the first goroutine will run and remove the meta data for the shard groups.

This results in a situation where it looks like the shards have gone, e.g., according to `SHOW SHARDS` or `SHOW SERIES`, but in fact the shards remain on disk (and importantly, their series remains within the index) until the next time the second goroutine runs. By default that's another 30 minutes.

In the case where the shards to be removed would have removed the last occurrences of some series, then it's possible that if the database was already at its maximum series limit (or tag limit for that matter), no further new series can be inserted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,7 @@
 - [#8983](https://github.com/influxdata/influxdb/issues/8983): Remove the pidfile after the server has exited.
 - [#9005](https://github.com/influxdata/influxdb/pull/9005): Return `query.ErrQueryInterrupted` for successful read on `InterruptCh`.
 - [#8989](https://github.com/influxdata/influxdb/issues/8989): Fix race inside Measurement index.
+- [#8819](https://github.com/influxdata/influxdb/issues/8819): Ensure retention service always removes local shards.
 
 ## v1.3.4 [unreleased]
 
--- a/internal/meta_client.go
+++ b/internal/meta_client.go
@@ -32,6 +32,8 @@ type MetaClientMock struct {
 
 	OpenFn func() error
 
+	PruneShardGroupsFn func() error
+
 	RetentionPolicyFn func(database, name string) (rpi *meta.RetentionPolicyInfo, err error)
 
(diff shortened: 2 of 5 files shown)


e-dard:
Oh yeah I forgot to add in the description: prior to this PR it was impossible to disabled the retention service via the config 😂

PR #9021, opened 2017-10-26
Update to go 1.9.2

jwilder:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

Updates to go1.9.2

### DIFF
--- a/Dockerfile_build_ubuntu64_go19
+++ b/Dockerfile_build_ubuntu64_go19
@@ -23,8 +23,8 @@ RUN gem install fpm
 ENV GOPATH /root/go
 ENV GO_VERSION 1.9.2
 ENV GO_ARCH amd64
-RUN wget https://storage.googleapis.com/golang/go${GO_VERSION}.linux-${GO_ARCH}.tar.gz; \
-   tar -C /usr/local/ -xf /go${GO_VERSION}.linux-${GO_ARCH}.tar.gz ; \
+RUN wget https://storage.googleapis.com/golang/go${GO_VERSION}.linux-${GO_ARCH}.tar.gz && \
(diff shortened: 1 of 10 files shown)


e-dard:
@jwilder my only suggestion would be we switch out the 1.9 test runner for a 1.10 one. The 1.10 branch is open. I assume Circle doesn't fail the build if the testing-future-go-version runner fails does it?

mark-rushakoff:
There's a couple other mentions of Go versions that need to be updated. I'm locating those now and I'll drop them in here.

jwilder:
@e-dard It does fail the build.

mark-rushakoff:
@jwilder there's some go 1.8.4 occurrences in CONTRIBUTING.md that need updated, then I think this PR is good to go.

jwilder:
@mark-rushakoff Fixed

Dockerfile_build_ubuntu64_go19
```diff
@@ -21,7 +21,7 @@ RUN gem install fpm
 
 # Install go
 ENV GOPATH /root/go
-ENV GO_VERSION 1.9.1
+ENV GO_VERSION 1.9.2
```
mark-rushakoff: The _go19 file isn't really necessary - it should be _go110 but I'm not sure whether that's available yet. I'm fine leaving the file around so we don't have to undo and redo the work to have circle test the next Go version.

jwilder: Yeah.  I left it as is since I couldn't find a go1.10.

e-dard:
@jwilder dang. I guess we should probably hold the 1.10 runner off until the branch is a bit more stable.

PR #9028, opened 2017-10-28
Fix linting on cmd files

cbushko:
In reference to [Enable `golint` on the code base](https://github.com/influxdata/influxdb/issues/4098)

FYI:  meta.pb.go is a generated file and the newest version of golint will ignore it if the header is
 > // Code generated by protoc-gen-go. DO NOT EDIT.
 
The newer versions of protoc generates the proper single line header.

- [x] Rebased/mergable
- [x] Tests pass


cmd/store/query/query.go
```diff
@@ -107,32 +107,34 @@ func (cmd *Command) Run(args ...string) error {
 
 	// set defaults
 	if start != "" {
-		if t, err := parseTime(start); err != nil {
+		t, err := parseTime(start)
```
e-dard: We can reduce these two blocks down by three lines. 

On line 108 define `var err error`, then on line 110:

```go
if cmd.startTime, err = parseTime(start); err != nil {
    return err
}
```

and also line 120:

```go
if cmd.endTime, err = parseTime(end); err != nil {
    return err
}
```

e-dard:
@cbushko ah hang on. I just had a thought. What revision of `protoc` did you use to generate the proto file? You must ensure you only use the version of `protoc` that's specified in InfluxDB's `Godeps` file, which is currently: `github.com/gogo/protobuf 1c2b16bc280d6635de6c52fc1471ab962dc36ec9`

cbushko:
I did not regenerate the files. I only changed the header at the top so that linting would pass. I was worried about causing a regression in the code.

PR #9033, opened 2017-10-28
Parallelize first and last in the tsm1 engine

ssiddhantsharma:
Also prevent the first and last optimization from being used if the
outer dimensions do not match the inner dimensions. This is because the
optimization will not return the correct points if these groupings
differ from each other.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

- [ ] Provide example syntax


jsternberg:
Please don't submit my own commits as a pull request under your own name.

PR #9056, opened 2017-11-03
Add logging configuration to the main configuration file

jsternberg:
Make logfmt the default log output format with json and a more user
friendly console into possible options.

This also adds the ability to suppress the logo output in the configuration
file. If the configuration file fails to be parsed in any way, these options
also fail and the defaults are used.

This change makes it so InfluxDB has a more consistent user output and follows
along with Kapacitor for the default output. That should make it easier to
consolidate Kapacitor and InfluxDB to use the same logging framework and
patterns.

The console logging output can only be used by specifying `auto` and writing to a TTY. It is not an available log format for logging to a file.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated
- [x] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): issue filed or pull request submitted \<[influxdata/docs.influxdata.com](https://github.com/influxdata/docs.influxdata.com/issues/1418)\>


Artanicus:
Would this change also suppress the extra timestamp printed currently? (Hoping that it does.) Or would that become configurable. (Didn't see any clear options for choosing parameters.)

jsternberg:
@Artanicus can you clarify what you mean? Thanks.

logger/logger.go
```diff
@@ -1,24 +1,81 @@
...
 func New(w io.Writer) *zap.Logger {
+	config := NewConfig()
+	l, _ := config.New(w)
+	return l
+}
+
+func (c *Config) New(defaultOutput io.Writer) (*zap.Logger, error) {
+	w := defaultOutput
+	format := c.Format
+	if format == "console" {
```
stuartcarnie: this will work, but you could also switch this to a white list

```go
if format != "json" && format != "logfmt" && format != "auto" {
    return nil, fmt.Errorf("unknown logging format: %s", format)
}
```

jsternberg: The error message for most log formats is handled by `newEncoder`. The reason why this is handled specially is because I didn't want to modify `newEncoder` to accept `auto`. Since `newEncoder` can't make the decision since it doesn't have access to the `io.Writer`, I wanted to make it explicit that `console` was specified. Since that means console is a valid input, I had to make it so a user-specified console would be intercepted and would pretend to be an invalid format.

PR #9073, opened 2017-11-07
Improve SHOW TAG KEYS performance.

benbjohnson:
- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/coordinator/statement_executor.go
+++ b/coordinator/statement_executor.go
@@ -188,6 +188,8 @@ func (e *StatementExecutor) ExecuteStatement(stmt influxql.Statement, ctx query.
 		rows, err = e.executeShowStatsStatement(stmt)
 	case *influxql.ShowSubscriptionsStatement:
 		rows, err = e.executeShowSubscriptionsStatement(stmt)
+	case *influxql.ShowTagKeysStatement:
+		return e.executeShowTagKeys(stmt, &ctx)
 	case *influxql.ShowTagValuesStatement:
 		return e.executeShowTagValues(stmt, &ctx)
 	case *influxql.ShowUsersStatement:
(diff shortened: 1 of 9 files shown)


coordinator/statement_executor.go
```diff
@@ -932,6 +934,98 @@ func (e *StatementExecutor) executeShowSubscriptionsStatement(stmt *influxql.Sho
 	return rows, nil
 }
 
+func (e *StatementExecutor) executeShowTagKeys(q *influxql.ShowTagKeysStatement, ctx *query.ExecutionContext) error {
+	var dis []meta.DatabaseInfo
+	if q.Database != "" {
```
e-dard: I don't think that `q.Database` can be empty. The `StatementRewriter` should have set it I think. Did you encounter a case where it was empty?

coordinator/statement_executor.go
```diff
@@ -932,6 +934,98 @@ func (e *StatementExecutor) executeShowSubscriptionsStatement(stmt *influxql.Sho
...
+		di := e.MetaClient.Database(q.Database)
+		if di == nil {
+			return fmt.Errorf("database not found: %s", q.Database)
+		} else if di.DefaultRetentionPolicy == "" {
+			return fmt.Errorf("database %s does not have default retention policy", q.Database)
+		}
+		dis = []meta.DatabaseInfo{*di}
+	} else {
+		dis = e.MetaClient.Databases()
```
e-dard: I don't think `SHOW TAG VALUES` should work across databases.

tsdb/store.go
```diff
@@ -1000,6 +1000,159 @@ func (s *Store) MeasurementSeriesCounts(database string) (measuments int, series
...
+func (a tagKeysSlice) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a tagKeysSlice) Less(i, j int) bool { return bytes.Compare(a[i].name, a[j].name) == -1 }
+
+// TagKeys returns the tag keys in the given database, matching the condition.
+func (s *Store) TagKeys(auth query.Authorizer, shardIDs []uint64, cond influxql.Expr) ([]TagKeys, error) {
+	cond = influxql.Reduce(influxql.RewriteExpr(cond, func(e influxql.Expr) influxql.Expr {
+		switch e := e.(type) {
+		case *influxql.BinaryExpr:
+			// Remove time clause from condition
+			if ref, ok := e.LHS.(*influxql.VarRef); ok && strings.ToLower(ref.Val) == "time" {
```
e-dard: The `time` clause should already be removed by `influxql.ConditionExpr` in the `StatementExecutor` I think.

coordinator/statement_executor.go
benbjohnson: I think you're right. Not sure where I was thinking it could be empty. Fixed in 156f25a.

coordinator/statement_executor.go
benbjohnson: Fixed in 156f25a.

tsdb/store.go
benbjohnson: Removed in 156f25a.

PR #9092, opened 2017-11-10
Initial jenkinsfile

jsternberg:
- [ ] Rebased/mergable
- [ ] Tests pass

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,79 @@
+readTrusted 'Dockerfile_jenkins_ubuntu32'
+
+pipeline {
+  agent none
+
+  stages {
+    stage('64bit') {
+      agent {
+        docker {
+          image 'golang:1.9.2'
+        }
+      }
+
+      steps {
+        sh """
+        mkdir -p /go/src/github.com/influxdata
+        cp -a $WORKSPACE /go/src/github.com/influxdata/influxdb
+
+        cd /go/src/github.com/influxdata/influxdb
+        go get github.com/sparrc/gdm
+        gdm restore
+        """
+
+        sh """
+        cd /go/src/github.com/influxdata/influxdb
+        go test -parallel=1 ./...
+        """
+      }
+    }
+
+    stage('32bit') {
+      agent {
+        dockerfile {
+          filename 'Dockerfile_jenkins_ubuntu32'
+        }
+      }
+
+      steps {
+        sh """
+        mkdir -p /go/src/github.com/influxdata
+        cp -a $WORKSPACE /go/src/github.com/influxdata/influxdb
+
+        cd /go/src/github.com/influxdata/influxdb
+        go get github.com/sparrc/gdm
+        gdm restore
+        """
+
+        sh """
+        cd /go/src/github.com/influxdata/influxdb
+        go test -parallel=1 ./...
+        """
+      }
+    }
+
+    stage('tsi index') {
+      agent {
+        docker {
+          image 'golang:1.9.2'
+        }
+      }
+
+      steps {
+        sh """
+        mkdir -p /go/src/github.com/influxdata
+        cp -a $WORKSPACE /go/src/github.com/influxdata/influxdb
+
+        cd /go/src/github.com/influxdata/influxdb
+        go get github.com/sparrc/gdm
+        gdm restore
+        """
+
+        sh """
+        cd /go/src/github.com/influxdata/influxdb
+        INFLUXDB_DATA_INDEX_VERSION=tsi1 go test -parallel=1 ./...
+        """
+      }
+    }
+  }
+}
--- a/services/retention/service.go
+++ b/services/retention/service.go
@@ -79,6 +79,12 @@ func (s *Service) run() {
 			return
 
 		case <-ticker.C:
+			select {
+			case <-s.done:
+				return
+			default:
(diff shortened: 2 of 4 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,79 @@
...
+        sh """
+        mkdir -p /go/src/github.com/influxdata
+        cp -a $WORKSPACE /go/src/github.com/influxdata/influxdb
+
+        cd /go/src/github.com/influxdata/influxdb
+        go get github.com/sparrc/gdm
+        gdm restore
+        """
+
+        sh """
```
mark-rushakoff: Why are these separate `sh` blocks instead of one contiguous `sh` block?

Jenkinsfile
```diff
@@ -0,0 +1,79 @@
...
+        cd /go/src/github.com/influxdata/influxdb
+        go test -parallel=1 ./...
```
mark-rushakoff: What do you think about archiving the output of `go test`, and running that output through go-junit-report so Jenkins can parse the individual test states?

jsternberg: I'm concerned that there's too much output and part of the reason why we never figure out the problems with Circle is because we were always drowned in so much output. I've got an idea for filtering the verbose output. Maybe we can have it so whatever filter will do, it can do junit and also output the full verbose text to a file we can archive?

Jenkinsfile
jsternberg: I think there's going to be a feature in the future to name the blocks. I wanted it so these would be shown as separate steps in the UI.

Jenkinsfile
mark-rushakoff: As-is, all of the `go test` output from each stage will be shown in Jenkins' console output page, which is approximately as bad as the current situation with Circle. 

Changing this to `go test -parallel=1 ./... 2>&1 | tee /path/to/full-test-log.txt` would allow us to then archive that file, so we could view the full test output from a single run, and we can convert it to junit so we get the easily navigated individual test breakdown. That's a pretty small change that I would prefer to see now, to hold over until the more sophisticated solution you're describing.

e-dard: Why are we running with `parallel=1`? That's only going to run one package's tests at a time.

Jenkinsfile
```diff
@@ -0,0 +1,79 @@
...
+        """
+      }
+    }
+
+    stage('tsi index') {
```
e-dard: We don't need a `tsi` specific runner. Where it makes sense to run tests on the `tsi` index they're already told to do so inside the test code.

services/retention/service.go
```diff
@@ -79,6 +79,12 @@ func (s *Service) run() {
 			return
 
 		case <-ticker.C:
+			select {
```
e-dard: Why did you add this here?

Jenkinsfile
jsternberg: We are currently doing that for CI on Circle. It's supposed to make sure the output is properly aligned, but I just recently learned that it doesn't do that so this will likely change.

Jenkinsfile
jsternberg: Same with this one. It's currently like that in the CircleCI tests. I can remove it though.

services/retention/service.go
jsternberg: One of the tests set the check interval really low to 1 millisecond. This meant that, even if the service was done, the timer kept getting set off and Go kept choosing this branch so it never exited. This goes and ensures that the done channel gets priority even if there is a pending check interval.

Jenkinsfile
e-dard: Yep it can go now, I think.

Jenkinsfile
mark-rushakoff: This is how many concurrent `t.Parallel` tests are run.

```
	-parallel n
	    Allow parallel execution of test functions that call t.Parallel.
	    The value of this flag is the maximum number of tests to run
	    simultaneously; by default, it is set to the value of GOMAXPROCS.
	    Note that -parallel only applies within a single test binary.
	    The 'go test' command may run tests for different packages
	    in parallel as well, according to the setting of the -p flag
	    (see 'go help build').
```

If we needed one package to run at a time, the `-p` help says:

```
	-p n
		the number of programs, such as build commands or
		test binaries, that can be run in parallel.
		The default is the number of CPUs available.
```

e-dard: @mark-rushakoff I'm always getting `-parallel` and `-p` mixed up 😄

services/retention/service.go
e-dard: @jsternberg The timer shouldn't have kept getting set off. When the service is closed then the `<-s.done` case will be available. When two competing cases are available they're chosen uniformly randomly. I doubt it would have taken more than a few `ms` for the `<-s.done` case to be chosen.

I don't think adding this logic to the service for the sake of a test is the right way to go. What problems were you having with the test (I wrote it)—maybe we can refactor it?

jsternberg: Ok. I'll try to figure out why it wasn't finishing and revert this back to the original code.

Even then, I'm not sure we should bother with choosing randomly. The intention appears to be that if the service is done, the service is done. Go doesn't seem to have a very good pattern for "prefer this channel over others".

PR #9095, opened 2017-11-13
shell shows error and usage when given extra parameters

### LINKED ISSUE #8593: Shell ignores extra parameters
### Feature Request

__Proposal:__ trigger an error when launching the shell with extra parameters (arguments, not options).

__Current behavior:__ when you launch the shell, arguments are simply ignored. This can be misleading.

Each database has its own CLI, and you could forget whether you need to pass the hostname to connect to as an option or as an argument. If you pass the hostname as an argument and have a local InfluxDB database running, it will connect to your local DB, but you might think it successfully connected to the remote host.

```
$ influx host_I_want_to_connect_to
Connected to http://localhost:8086 version v1.3.0
InfluxDB shell version: v1.3.0
>
```

__Desired behavior:__ an error would be ideal. It should print the usage and indicate that the extra argument is not valid.

__Use case:__ this just saves time in case people make a mistake in their command line.

stop-start:
Fixes #8593 

When try to launch the shell with arguments the shell won't launch, instead will show an error (with the unknown args) as well as the usage.

- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/cmd/influx/main.go
+++ b/cmd/influx/main.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/influxdata/influxdb/client"
 	"github.com/influxdata/influxdb/cmd/influx/cli"
+	"strings"
 )
 
 // These variables are populated via the Go linker.
@@ -107,6 +108,13 @@ Examples:
 `)
 	}
 	fs.Parse(os.Args[1:])
+
+	argsNotParsed := os.Args[len(os.Args)-fs.NArg():]
+	if len(argsNotParsed) > 0 {
+		fmt.Fprintf(os.Stderr, "unknown arguments: %s\n", strings.Join(argsNotParsed, " "))
+		fs.Usage()
+		os.Exit(1)
+	}
 
 	if c.ShowVersion {
 		c.Version()
--- a/cmd/influx/main.go
+++ b/cmd/influx/main.go
@@ -109,7 +109,7 @@ Examples:
 	}
 	fs.Parse(os.Args[1:])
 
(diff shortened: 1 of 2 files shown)


cmd/influx/main.go
```diff
@@ -108,6 +109,13 @@ Examples:
 	}
 	fs.Parse(os.Args[1:])
 
+	argsNotParsed := os.Args[len(os.Args)-fs.NArg():]
+	if len(argsNotParsed) > 0 {
```
mark-rushakoff: This block should use `flag.Args()` instead.

https://golang.org/pkg/flag/#Arg
> Args returns the non-flag command-line arguments.

mark-rushakoff:
@stop-start would you mind adding a line like

```
- [#9065](https://github.com/influxdata/influxdb/pull/9065): Refuse extra arguments to influx CLI
```

to the 1.5.0 / Bugfixes section of CHANGELOG.md?

stop-start:
@mark-rushakoff  Done

mark-rushakoff:
Thanks @stop-start!

stop-start:
Thanks @mark-rushakoff, it's my second pull request but first looked at and approved!

Luit:
The PR number in CHANGELOG.md is the wrong one, though. 9065 instead of 9095.

mark-rushakoff:
Good catch @Luit, that was a typo on my part. Care to open a small PR to fix it?

mark-rushakoff:
I opened a PR to fix it in #9182.

PR #9101, opened 2017-11-13
pkg/escape: Preallocate output in Unescape

mjs:
Preallocating the capacity of the output is faster and uses less memory than letting append() do its own (over)allocation. 

Benchmarking shows that this is ~42% faster than the original. The benchmark used is here: https://gist.github.com/mjs/bc3219a79d4e4a1d7848c5aaa12b3c40. The results on my machine are:

```
BenchmarkOrig-4   	20000000	        85.1 ns/op
BenchmarkNew-4    	30000000	        49.4 ns/op
```

- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated (trivial change - not required?)

### DIFF
--- a/pkg/escape/bytes.go
+++ b/pkg/escape/bytes.go
@@ -78,7 +78,11 @@ func Unescape(in []byte) []byte {
 
 	i := 0
 	inLen := len(in)
-	var out []byte
+
+	// The output size will be no more than inLen. Preallocating the
+	// capacity of the output is faster and uses less memory than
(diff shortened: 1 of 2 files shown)


e-dard:
@mjs has signed the CLA.

@mjs thanks for this. Would you mind adding some benchmarks to `pkg/escape/bytes_test.go`?

mjs:
I've added benchmarks for all bytes escape and unescape functions.

PR #9104, opened 2017-11-14
fix descending queries

stuartcarnie:
* did not handle cached values correctly
* sort shards by time in either ascending or descending
  order depending on the RPC request ordering to ensure they
  are traversed in the correct order.

- [x] Rebased/mergable
- [x] Tests pass


stuartcarnie:
@nathanielc I am adding a few unit tests for additional verification

PR #9114, opened 2017-11-14
Add capability to force a full compaction

jwilder:
This adds the capability to the engine to force a full compaction
to be scheduled.  When called, it snapshots any data in the cache,
aborts running compactions and prevents level plans from returning
level plans.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/tsdb/engine/tsm1/compact.go
+++ b/tsdb/engine/tsm1/compact.go
@@ -77,6 +77,10 @@ type CompactionPlanner interface {
 	PlanOptimize() []CompactionGroup
 	Release(group []CompactionGroup)
 	FullyCompacted() bool
+
+	// ForceFull causes the planner to return a full compaction plan the next
+	// time Plan() is called if there are files that could be compacted.
+	ForceFull()
 }
 
 // DefaultPlanner implements CompactionPlanner using a strategy to roll up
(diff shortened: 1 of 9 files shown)


tsdb/engine/tsm1/compact.go
```diff
@@ -276,6 +304,16 @@ func (c *DefaultPlanner) PlanLevel(level int) []CompactionGroup {
 // to optimize the index across TSM files.  Each returned compaction group can be
 // compacted concurrently.
 func (c *DefaultPlanner) PlanOptimize() []CompactionGroup {
+	c.mu.RLock()
+	forceFull := c.forceFull
+	c.mu.RUnlock()
+
+	// If a full plan has been requested, don't plan any levels which will prevent
+	// the full plan from acquiring them.
```
e-dard: Can you clarify this language a bit? To me this suggests that the check here is to prevent another compaction coming in from another goroutine and getting in the way of a previously requested full compaction. Is that the case? 

If that's the case, without this `if forceFull` check would it be possible for competing compactions to cause deadlocks or hangs?

e-dard: > The forceFull here is to ensure that only a full plan will be returned in the next scheduler loop which causes the scheduler to pick a full plan and try to execute it.

I guess what I'm wondering is that if this racy check is a problem or not. Because you're checking `forceFull` outside of a lock, it's possible that when you do the check `forceFull != c.ForceFull`. Might it be safer to do the check under the `RLock`?

jwilder: I see what you're saying.  Yeah, I can switch that up.

PR #9127, opened 2017-11-16
Implement FGA on SHOW COMMANDS

e-dard:
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

This PR does the following:

   - Fix bug in `tsi1` where database names were not propagating properly. This would have been a problem for FGA.
   - Implement FGA on the `SeriesPointIterator`. This will be needed by `SHOW SERIES`, and I suspect `ifql`.
   - Add test coverage for FGA when using `SHOW TAG KEYS` and `SHOW TAG VALUES`. It looks like FGA was working on them even after the recent `SHOW` command refactor.
  - Implement FGA on `SHOW MEASUREMENTS` now that `SHOW MEASUREMENTS` is using its own command, rather than a `SelectStatement`.
  - Fix a bug in `tsi1` where negation operations on tags would not return the correct values.
  - Add significant test coverage for FGA on `SHOW MEASUREMENTS`.

### DIFF
--- a/tsdb/index/inmem/inmem.go
+++ b/tsdb/index/inmem/inmem.go
@@ -392,24 +392,26 @@ func (i *Index) TagsForSeries(key string) (models.Tags, error) {
 
 // MeasurementNamesByExpr takes an expression containing only tags and returns a
 // list of matching meaurement names.
-func (i *Index) MeasurementNamesByExpr(expr influxql.Expr) ([][]byte, error) {
+func (i *Index) MeasurementNamesByExpr(auth query.Authorizer, expr influxql.Expr) ([][]byte, error) {
 	i.mu.RLock()
 	defer i.mu.RUnlock()
 
 	// Return all measurement names if no expression is provided.
 	if expr == nil {
 		a := make([][]byte, 0, len(i.measurements))
-		for name := range i.measurements {
-			a = append(a, []byte(name))
+		for _, m := range i.measurements {
+			if m.Authorized(auth) {
+				a = append(a, m.name)
+			}
 		}
 		bytesutil.Sort(a)
 		return a, nil
 	}
 
-	return i.measurementNamesByExpr(expr)
+	return i.measurementNamesByExpr(auth, expr)
 }
 
-func (i *Index) measurementNamesByExpr(expr influxql.Expr) ([][]byte, error) {
+func (i *Index) measurementNamesByExpr(auth query.Authorizer, expr influxql.Expr) ([][]byte, error) {
 	if expr == nil {
 		return nil, nil
 	}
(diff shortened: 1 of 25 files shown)


tsdb/index/inmem/inmem.go
```diff
@@ -490,20 +493,29 @@ func (i *Index) measurementNamesByNameFilter(op influxql.Token, val string, rege
...
 // measurementNamesByTagFilters returns the sorted measurements matching the filters on tag values.
-func (i *Index) measurementNamesByTagFilters(filter *TagFilter) [][]byte {
+func (i *Index) measurementNamesByTagFilters(auth query.Authorizer, filter *TagFilter) [][]byte {
 	// Build a list of measurements matching the filters.
 	var names [][]byte
 	var tagMatch bool
+	var authorized bool
+
+	// valEqual determins if the provided []byte is equal to the tag value
+	// to be filtered on.
```
benbjohnson: `s/determins/determines/`

tsdb/index/inmem/inmem.go
```diff
@@ -490,20 +493,29 @@ func (i *Index) measurementNamesByNameFilter(op influxql.Token, val string, rege
...
+	valEqual := func(b []byte) bool {
+		if filter.Op == influxql.EQ || filter.Op == influxql.NEQ {
+			return bytes.Equal([]byte(filter.Value), b)
+		}
+		return filter.Regex.Match(b)
+	}
```
benbjohnson: Does this need to be an inline function? Can you move it out and just pass `filter` in?

tsdb/index/inmem/inmem.go
```diff
@@ -513,37 +525,53 @@ func (i *Index) measurementNamesByTagFilters(filter *TagFilter) [][]byte {
 		}
 
 		tagMatch = false
+		authorized = true
+		if auth != nil {
+			authorized = false // Authorization must be explicitly granted.
+		}
```
benbjohnson:
```go
authorized := auth != nil // Authorization must be explicitly granted.
```

tsdb/store_test.go
```diff
@@ -959,6 +960,193 @@ func TestStore_TagValues(t *testing.T) {
...
+		}
+
+		// values should not contain any tag values associated with a series containing
+		// a secret tag.
+		expValues := 2
+		var gotValues int
+		for _, tv := range values {
+			if got, exp := tv.Measurement, "cpu"; got != exp {
+				return fmt.Errorf("got measurment %q, expected %q", got, exp)
+			}
```
benbjohnson: `s/measurment/measurement/`

tsdb/index/tsi1/file_set.go
```diff
@@ -639,62 +643,115 @@ func (fs *FileSet) measurementNamesByNameFilter(op influxql.Token, val string, r
...
+		if auth != nil {
+			authorized = false // Authorization must be explicitly granted.
+		}
+
+		vitr := fs.TagValueIterator(me.Name(), []byte(key))
+		if vitr != nil {
+			for ve := vitr.Next(); ve != nil; ve = vitr.Next() {
+				// If a match was found for the EQ or EQREGEX operator then
+				// the measurement should be included (if it's authorized).
+				if valEqual(ve.Value()) {
```
joelegasse: `if !valEqual() { continue }` and save the indent?

tsdb/index/tsi1/index_test.go
```diff
@@ -394,3 +338,11 @@ func (idx *Index) CreateSeriesSliceIfNotExists(a []Series) error {
 	}
 	return nil
 }
+
+func BytesToStrings(a [][]byte) []string {
```
joelegasse: I see this in a couple places, would it make sense to add to the `bytesutils` package?

e-dard: Sounds good. I did consider this, but was concerned others' might object to putting code only used in tests into non-test packages.

e-dard:
@benbjohnson @joelegasse please see a5af19fc0 where I have addressed feedback. I also added an extra optimisation on `inmem` where we only continue checking for tag matches if we're [using a regex](https://github.com/influxdata/influxdb/pull/9127/commits/a5af19fc06e6fe319f30162c14abbb88b24b3edf#diff-b27ccb4c1a2b59f94983d00fd2381c5cR551)

benbjohnson:
👌 Don't forget the `CHANGELOG`.

PR #9128, opened 2017-11-16
Fix field dimensions when one shard has no series for a particular measurement

jsternberg:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated


jsternberg:
I think I'm wrong. I don't think this is actually required to make it work correctly and the new code uses this better style. As long as it's working in 1.3, it doesn't have to be pretty.

I'll double check, but I'm closing this for now since I think it's unnecessary.

PR #9143, opened 2017-11-21
SHOW TAG KEYS with high cardinality and many shards 

### LINKED ISSUE #9125: Show tag keys performance on 1.4.1 (tsi1)
### Bug report

__System info:__
influxDB 1.4.1 with tsi1
Amazon Linux AMI release 2015.09 (centos based) 
Machine type: r4.4xlarge

__Description:__

We have a high cardinality use case, as can be seen by:

```
SELECT numSeries FROM "_internal".."database" WHERE time > now() - 10s GROUP BY "database" ORDER BY desc LIMIT 1

name: database
tags: database=something
time                numSeries
----                ---------
1510849360000000000 31199677
```

This is mostly due to a specific measurement.

On 1.3.7 when running "show tag keys from measurement" it executes in a few seconds. 

We tried to upgrade to 1.4.1, to do so we did the following:
 - stop influx
 - delete the previous index files - given that this version has incompatible indexes as stated on #8857 
 - use influx_inspect inmem2tsi to create the new index files
 - start influx

__Expected behavior:__ 

Have similar performance between 1.3.7 and 1.4.1

__Actual behavior:__ 

 "show tag keys from measurement"  takes around 10 minutes to execute.

Tried other lower cardinality measurements on the same database and the performance looks similar to 1.3.7

__Additional info:__ 

We are trying to build a way to reproduce the issue that we can share, possibly a script that generates similar cardinality but we haven't had time to finish it yet.

e-dard:
- [ ] Rebased/mergable
- [ ] Tests pass

Fixes #9125.

This PR fixes a performance issue where `SHOW TAG KEYS` on a system using the `tsi1` index and many shards / high cardinality would become very expensive. This was due to the way with checked for filtered tag values. Filtered tag values would exist is you did a query like `SHOW TAG KEYS WHERE "region" = 'west'`. In the case where no `WHERE` clause is present we were actually retrieving all tag values, rather than retrieving none at all. The latter would obviously be preferable from a performance perspective.

This PR adds a fix for this by adding a new method on the engine that lets you determine if there exists an authorised series for a given measurement and tag key, without having to instantiate the expensive `TagValueSeriesIterator` iterator, and without having to materialise millions of tag values. We short-circuit where appropriate to speed up performance.

### Benchmark

For a database with 3M series across 500 shards using `tsi1`:

```
// v1.4.2
$ edd@work:~|⇒  time influx -database stress -execute 'show tag keys'               
name: m0
tagKey
------
tag0
tag1
tag2
tag3
tag4
tag5
tag6
influx -database stress -execute 'show tag keys'  0.01s user 0.00s system 0% cpu 15.548 total
```

**15.5** seconds.

```
// This PR
edd@work:~|⇒  time influx -database stress -execute 'show tag keys'
name: m0
tagKey
------
tag0
tag1
tag2
tag3
tag4
tag5
tag6
influx -database stress -execute 'show tag keys'  0.00s user 0.01s system 24% cpu 0.021 total
```

**0.021** seconds.

### DIFF
--- a/tsdb/index/inmem/inmem.go
+++ b/tsdb/index/inmem/inmem.go
@@ -274,6 +274,48 @@ func (i *Index) MeasurementTagKeysByExpr(name []byte, expr influxql.Expr) (map[s
 	return mm.TagKeysByExpr(expr)
 }
 
+// TagKeyHasAuthorizedSeries determines if there exists an authorized series for
+// the provided measurement name and tag key.
+func (i *Index) TagKeyHasAuthorizedSeries(auth query.Authorizer, name []byte, key string) bool {
+	i.mu.RLock()
+	mm := i.measurements[string(name)]
+	i.mu.RUnlock()
+
+	if mm == nil {
+		return false
+	}
+
+	// TODO(edd): This looks like it's inefficient. Since a series can have multiple
+	// tag key/value pairs on it, it's possible that the same unauthorised series
+	// will be checked multiple times. It would be more efficient if it were
+	// possible to get the set of unique series IDs for a given measurement name
+	// and tag key.
+	var authorized bool
+	mm.SeriesByTagKeyValue(key).Range(func(_ string, seriesIDs SeriesIDs) bool {
+		if auth == nil || auth == query.OpenAuthorizer {
+			authorized = true
+			return false
+		}
+
+		for _, id := range seriesIDs {
+			s := mm.SeriesByID(id)
+			if s == nil {
+				continue
+			}
+
+			if auth.AuthorizeSeriesRead(i.database, mm.name, s.Tags()) {
+				authorized = true
+				return false
+			}
+		}
+
+		// This tag key/value combination doesn't have any authorised series, so
+		// keep checking other tag values.
+		return true
+	})
+	return authorized
+}
+
 // MeasurementTagKeyValuesByExpr returns a set of tag values filtered by an expression.
 //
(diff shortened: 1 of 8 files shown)


tsdb/index/inmem/inmem.go
```diff
@@ -274,6 +274,48 @@ func (i *Index) MeasurementTagKeysByExpr(name []byte, expr influxql.Expr) (map[s
...
+	}
+
+	// TODO(edd): This looks like it's inefficient. Since a series can have multiple
+	// tag key/value pairs on it, it's possible that the same unauthorised series
+	// will be checked multiple times. It would be more efficient if it were
+	// possible to get the set of unique series IDs for a given measurement name
+	// and tag key.
+	var authorized bool
+	mm.SeriesByTagKeyValue(key).Range(func(_ string, seriesIDs SeriesIDs) bool {
+		if auth == nil || auth == query.OpenAuthorizer {
```
joelegasse: This check can be moved up to the top of this method. Also, since `OpenAuthorizer` is now just a public package level variable instead of a type, it should not be checked for equality. However unlikely, it could be reassigned a value that no longer just returns `true` for all its methods. This is my fault, anyway... I should have just allowed for a `nil` Authorizer everywhere initially to mean "bypass authorization checks", rather than trying to mock a valid authorizer.

e-dard: @joelegasse https://github.com/influxdata/influxdb/issues/9159 to track this.

PR #9147, opened 2017-11-22
Speedup startup time for TSI

jwilder:
This fixes the startup time issue with TSI indexes.  Startup time is slow because the field types for each measurement need to be reloaded at startup.  These types are not part of the TSI index and the only way to find them is to scan all the TSM file indexes.  Where cardinality is high or there are a lot of TSM files on disk, this can be slow.

This PR allows the field types to be persisted to a shard local `fields.idx` file that is updated when new fields are added/removed from a measurement.  If the file does not exists, we fallback to the old method and scan all the TSM index.  If the file exists, we just load this smaller file and skip scanning the index.

This file only affects TSI startup times as we still need to re-populate the index with `inmem` which requires scanning all the TSM indexes.

Local benchmarking of 32m series (32M values total) improved from about `1m6s` to `2s`.  Other cardinalities started up in about `2s` as well.  This is just with a single shard.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated


tsdb/engine/tsm1/engine.go
```diff
@@ -583,6 +593,12 @@ func (e *Engine) LoadMetadataIndex(shardID uint64, index tsdb.Index) error {
 	// Save reference to index for iterator creation.
 	e.index = index
 
+	// If we have the cached fields index on disk and and we're using TSI, we
+	// can skip scanning all the TSM files.
+	if e.index.Type() != inmem.IndexName && !e.fieldset.IsEmpty() {
+		return nil
+	}
```
benbjohnson: `s/and and/and/`

tsdb/shard.go
```diff
@@ -63,6 +66,13 @@ var (
 	// ErrShardDisabled is returned when a the shard is not available for
 	// queries or writes.
 	ErrShardDisabled = errors.New("shard is disabled")
+
+	// ErrUnknownFieldsFormat is returned when the fields index file is not identifiable by
+	// the files magic number.
```
e-dard: `s/files/file's`

tsdb/internal/meta.proto
```diff
@@ -7,21 +9,25 @@ package meta;
...
 message Series {
-  required string Key = 1;
+  string Key = 1;
   repeated Tag Tags = 2;
 }
 
 message Tag {
-  required string Key = 1;
-  required string Value = 2;
+  string Key = 1;
```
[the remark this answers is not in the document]
jwilder: These protobufs were dead code.  I removed them since I was re-purposing the code.

jwilder:
@e-dard Yeah, I held off of moving stuff around while @benbjohnson is still making changes to the index.  I agree, moving them to `index` makes sense.

PR #9151, opened 2017-11-22
initial opentrace implementation for ifql RPC interface

stuartcarnie:
NOTE: does not include a default tracer until configuration across
projects is standardized

- [x] Rebased/mergable
- [x] Tests pass


services/storage/rpc_service.go
```diff
@@ -31,35 +41,66 @@ func (r *rpcService) Hints(context.Context, *types.Empty) (*HintsResponse, error
...
+	ctx = tsm1.NewContextWithMetricsGroup(ctx)
+
+	var agg Aggregate_AggregateType
+	if req.Aggregate != nil {
+		agg = req.Aggregate.Type
+	}
+	pred := PredicateToExprString(req.Predicate)
+	groupKeys := strings.Join(req.Grouping, ",")
+	span.
+		SetTag("predicate", pred).
```
nathanielc: This could be quite long, do we want it always attached to the span?

services/storage/rpc_service.go
```diff
@@ -107,5 +148,14 @@ func (r *rpcService) Read(req *ReadRequest, stream Storage_ReadServer) error {
 
 	w.flushFrames()
 
+	span.SetTag("values", w.vc)
```
nathanielc: Is this a counter? Maybe a name the `num_values` would be more clear?

tsdb/engine/tsm1/engine_cursor.go
```diff
@@ -23,6 +24,10 @@ func (e *Engine) CreateCursor(ctx context.Context, r *tsdb.CursorRequest) (tsdb.
 		return nil, nil
 	}
 
+	if grp := metrics.GroupFromContext(ctx); grp != nil {
+		grp.GetCounter(numberOfRefCursorsCounter).Add(1)
```
nathanielc: Whats this?

services/storage/rpc_service.go
stuartcarnie: `num_values` is good

tsdb/engine/tsm1/engine_cursor.go
stuartcarnie: the number of TSM cursors that have been created; it is ultimately serialized as a tag

services/storage/rpc_service.go
stuartcarnie: are you referring to the `groupKeys` or the `pred` (predicate)?

nathanielc: The predicate

stuartcarnie: Actually, how about I use this pattern and only enable if we set an environment variable:

https://github.com/opentracing/opentracing-go#conditionally-capture-a-field-using-lognoop

stuartcarnie: Perhaps we can stick with a `max-tag-length`, that can later be exposed as a configuration option

stuartcarnie: It appears that is already taken care of by the client; `jaeger` for example limits the tag length to 256

nathanielc: ok, yeah a limit sounds good, its just informational so if we truncate it, its no big deal. Then we have the data so we can correlate it later.

services/storage/rpc_service.go
```diff
@@ -107,5 +148,14 @@ func (r *rpcService) Read(req *ReadRequest, stream Storage_ReadServer) error {
 
 	w.flushFrames()
 
+	span.SetTag("num_values", w.vc)
```
stuartcarnie: @nathanielc renamed to `num_values`

services/storage/rpc_service.go
stuartcarnie: added `truncateString` and limited to 256 also

PR #9162, opened 2017-11-27
improve startup performance

stuartcarnie:
* replaces coordinating goroutines for single k-way heap merge iterator
* removes contention sending keys across buffered channels

startup time from 46s -> 28s for iterating 1MM keys across 14 shards

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/tsdb/engine/tsm1/file_store_key_iterator.go
+++ b/tsdb/engine/tsm1/file_store_key_iterator.go
@@ -1,0 +1,111 @@
+package tsm1
+
+import (
+	"bytes"
+	"container/heap"
+)
+
+type keyIterator struct {
+	f    TSMFile
+	p, n int
+	key  []byte
+	typ  byte
+}
+
+func newKeyIterator(f TSMFile, seek []byte) *keyIterator {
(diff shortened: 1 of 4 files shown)


tsdb/engine/tsm1/file_store_key_iterator.go
```diff
@@ -0,0 +1,111 @@
+package tsm1
+
+import (
+	"bytes"
+	"container/heap"
+)
+
+type keyIterator struct {
+	f    TSMFile
+	p, n int
```
jwilder: Can you add a comment what `p` and `n` are?

jwilder: Assuming `p` is the current index and `n` is the total number of keys?

stuartcarnie: will do 👍🏻

stuartcarnie: done

PR #9164, opened 2017-11-27
inmem startup improvments

stuartcarnie:
* only call ParseTags when necessary
* remove dependency on inmem.Series in tsdb test package
* Measurement and Series are no longer exported. Their use is restricted
  to the inmem package
* improve Measurement and Series types by exporting immutable
  fields and removing unnecessary APIs and locks

Reduced startup time from 28s to 17s. Overall improvement including
#9162 reduces startup from 46s to 17s for 1MM series across 14 shards.

- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/models/points.go
+++ b/models/points.go
@@ -281,8 +281,8 @@ func ParseKeyBytes(buf []byte) ([]byte, Tags) {
 	return buf[:i], tags
 }
 
-func ParseTags(buf []byte) (Tags, error) {
-	return parseTags(buf), nil
+func ParseTags(buf []byte) Tags {
+	return parseTags(buf)
 }
 
 func ParseName(buf []byte) ([]byte, error) {
@@ -1528,9 +1528,12 @@ func parseTags(buf []byte) Tags {
 		return nil
 	}
 
-	tags := make(Tags, 0, bytes.Count(buf, []byte(",")))
+	tags := make(Tags, bytes.Count(buf, []byte(",")))
+	p := 0
 	walkTags(buf, func(key, value []byte) bool {
-		tags = append(tags, NewTag(key, value))
+		tags[p].Key = key
+		tags[p].Value = value
+		p++
 		return true
 	})
 	return tags
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -1122,7 +1122,9 @@ func (e *Engine) addToIndexFromKey(key []byte, fieldType influxql.DataType) erro
 
 	// Build in-memory index, if necessary.
 	if e.index.Type() == inmem.IndexName {
-		e.index.InitializeSeries(seriesKey, name)
+		if err := e.index.InitializeSeries(seriesKey, name, models.ParseTags(seriesKey)); err != nil {
+			return err
+		}
 	}
 
 	return nil
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -690,9 +690,9 @@ func (e *Engine) LoadMetadataIndex(shardID uint64, index tsdb.Index) error {
 	}
 
 	if err := e.FileStore.WalkKeys(nil, func(key []byte, typ byte) error {
-		fieldType, err := tsmFieldTypeToInfluxQLDataType(typ)
-		if err != nil {
-			return err
+		fieldType := BlockTypeToInfluxQLDataType(typ)
+		if fieldType == influxql.Unknown {
+			return fmt.Errorf("unknown block type: %v", typ)
 		}
 
 		if err := e.addToIndexFromKey(key, fieldType); err != nil {
@@ -1050,9 +1050,9 @@ func (e *Engine) overlay(r io.Reader, basePath string, asNew bool) error {
 	// lock contention on the index.
 	merged := merge(readers...)
 	for v := range merged {
-		fieldType, err := tsmFieldTypeToInfluxQLDataType(v.typ)
-		if err != nil {
-			return err
+		fieldType := BlockTypeToInfluxQLDataType(v.typ)
+		if fieldType == influxql.Unknown {
+			return fmt.Errorf("unknown block type: %v", v.typ)
 		}
 
 		if err := e.addToIndexFromKey(v.key, fieldType); err != nil {
@@ -1122,10 +1122,7 @@ func (e *Engine) addToIndexFromKey(key []byte, fieldType influxql.DataType) erro
 
 	// Build in-memory index, if necessary.
 	if e.index.Type() == inmem.IndexName {
-		tags, _ := models.ParseTags(seriesKey)
-		if err := e.index.InitializeSeries(seriesKey, name, tags); err != nil {
-			return err
-		}
+		e.index.InitializeSeries(seriesKey, name)
 	}
 
 	return nil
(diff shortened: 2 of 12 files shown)


models/points.go
```diff
@@ -1528,9 +1528,12 @@ func parseTags(buf []byte) Tags {
 		return nil
 	}
 
-	tags := make(Tags, 0, bytes.Count(buf, []byte(",")))
+	tags := make(Tags, bytes.Count(buf, []byte(",")))
+	p := 0
 	walkTags(buf, func(key, value []byte) bool {
-		tags = append(tags, NewTag(key, value))
+		tags[p].Key = key
```
stuartcarnie: @jwilder this change removes a `runtime.typedmemmove`

models/points.go
```diff
@@ -281,8 +281,8 @@ func ParseKeyBytes(buf []byte) ([]byte, Tags) {
 	return buf[:i], tags
 }
 
-func ParseTags(buf []byte) (Tags, error) {
-	return parseTags(buf), nil
+func ParseTags(buf []byte) Tags {
```
stuartcarnie: the `error` return value was always `nil`, so this removes a stack push / pop for every `ParseTags` call

tsdb/index/inmem/meta.go
```diff
@@ -19,43 +19,43 @@ import (
...
 // goroutine safe while un-exported functions assume the caller will use the
 // appropriate locks.
-type Measurement struct {
-	database string
-	Name     string `json:"name,omitempty"`
-	name     []byte // cached version as []byte
+type measurement struct {
+	Database  string
+	Name      string `json:"name,omitempty"`
+	NameBytes []byte // cached version as []byte
```
jwilder: Do we need both a string and a byte version of name?  Wonder if we can just have `Name []byte`.

stuartcarnie: 👍 I'll take a look and adjust if `Name` is no longer needed

stuartcarnie: `Name` is still used in a number of places, however I pushed an additional commit that replaces `[]byte(Measurement.Name)` with `Measurement.NameBytes` to save allocations

jwilder: Cool.  Can remove the other places some other time.

tsdb/engine/tsm1/engine.go
```diff
@@ -2544,21 +2541,22 @@ func SeriesFieldKeyBytes(seriesKey, field string) []byte {
...
+		BlockFloat64:  influxql.Float,
+		BlockInteger:  influxql.Integer,
+		BlockBoolean:  influxql.Boolean,
+		BlockString:   influxql.String,
+		BlockUnsigned: influxql.Unsigned,
+	}
+)
+
+func BlockTypeToInfluxQLDataType(typ byte) influxql.DataType {
+	if int(typ) < len(blockToFieldType) {
```
e-dard: As we're on an über optimisation drive here, you could also have:

```go
fieldTypesN byte = len(blockToFieldType)
...
...
...
func BlockTypeToInfluxQLDataType(typ byte) influxql.DataType {
    if typ < fieldTypesN {
```

tsdb/engine/tsm1/engine_test.go
```diff
@@ -816,6 +816,25 @@ func TestEngine_CreateCursor_Descending(t *testing.T) {
...
+		r[i] = b % m
+	}
+	return r
+}
+
+func BenchmarkBlockTypeToInfluxQLDataType(b *testing.B) {
+	t := makeBlockTypeSlice(100)
+	for i := 0; i < b.N; i++ {
+		for j := 0; j < len(t); j++ {
+			tsm1.BlockTypeToInfluxQLDataType(t[j])
```
e-dard: Are you sure this can't get optimised away by the compiler? That used to be a problem so I always got into the habit of defining a `var result influxql.Unknown` somewhere outside the benchmark.

tsdb/index/inmem/meta.go
```diff
@@ -1164,33 +1164,43 @@ func (a Measurements) Union(other Measurements) Measurements {
...
-	measurement *Measurement
-	shardIDs    map[uint64]struct{} // shards that have this series defined
-	deleted     bool
+	Measurement *measurement
+	Key         string
+	Tags        models.Tags
+
+	mu       sync.RWMutex
+	shardIDs map[uint64]struct{} // shards that have this series defined
+	deleted  bool
```
e-dard: Can we move `deleted` down to the bottom? There will be 7 bytes of padding after `deleted`, and if it gets left here then we'll end up with a larger than necessary struct size if we add anything that's less than 8 bytes underneath.

tsdb/index/inmem/meta.go
```diff
@@ -1367,35 +1339,42 @@ func (t *TagKeyValue) Range(f func(tagValue string, a SeriesIDs) bool) {
...
 // Store stores ids under the value key.
-func (t *TagKeyValue) Store(value string, ids SeriesIDs) {
+func (t *tagKeyValue) Store(value string, ids seriesIDs) {
 	t.mu.Lock()
-	defer t.mu.Unlock()
 	t.valueIDs[value] = ids
+	t.mu.Unlock()
+}
+
+// StoreByte stores ids under the value key.
```
e-dard: It's not clear to me why we need this new method.

tsdb/engine/tsm1/engine_test.go
stuartcarnie: Good catch – if the function was not inlined, it would have made no difference. We can deduce the inner loop overhead is thus about 45ns :)

Original benchmark was about 212ns, less loop overhead is 167ns. I added an assignment and it jumped from 45ns to 70ns. So about 30ns vs 167ns or about a 5.5x improvement.

tsdb/engine/tsm1/engine.go
stuartcarnie: I tried that, but alas the bounds check elimination (BCE) optimization can't be used. We can only declare `FieldTypesN` as a `var` and the compiler cannot prove this value doesn't change outside the `BlockTypeToInfluxQLDataType` function.

tsdb/index/inmem/meta.go
stuartcarnie: To complement the `LoadByte` as there is a drive to move to `[]byte` keys vs `string` keys. This change also removes a call to `runtime.slicebytetostring` at the call site in `measurement#AddSeries`.

stuartcarnie:
Will hold off until @e-dard merges the `er-tsi-index-part`

tsdb/engine/tsm1/engine.go
e-dard: Interesting. 👍

tsdb/engine/tsm1/engine_test.go
e-dard: Still a good improvement.

tsdb/index/inmem/meta.go
e-dard: I get the push to `[]byte`; it's something that's been slowly chipped away at over time but seems never ending 😄 

> The cast in StoreByte is for the map key, so we benefit from the compiler optimization that it skips the alloc.

That was kind of the underlying point of my confusion I guess.. We can't benefit from that particular compiler optimisation here because we need to alloc in order to guarantee that the map key is stable don't we? I'm pretty sure that particular optimisation would only work on the read-side wouldn't it?

With `Store(value string, ids seriesIDs)` I would assume there would be a `runtime.slicebytetostring` on the `string(value)` at the call site but then there would be no further allocations would there? Or maybe there would be another allocation when we did `t.valueIDs[value] = ids` in `Store`?

stuartcarnie: Indeed you are correct. Whilst map's won't allocate if a space exists to store a new key in a bucket, the assignment does indeed cause an allocation every time. It seems the compiler is not smart enough to delay the allocation _if_ the key does not exist.

However, that leads to another issue with this logic anyhow. The `*measurement#AddSeries` method calls `tagKeyValue#LoadByte` to fetch the seriesIDs slice and later overwrites with the `StoreByte` method. This is a race condition, as if two concurrent processes mutated the slice, one would overwrite the other on the `StoreByte` call. We know this won't happen, as it is only mutated via `*measurement#AddSeries` that takes an exclusive lock on the `measurement`. We should just remove all this complicated locking logic :)

e-dard: If you're referring to the lock on `TagKeyValue`, it was added specifically to fix an existing race (I forget where).

stuartcarnie:
This PR was merged as it still had a number of useful improvements, including cleaning up some `inmem` APIs

PR #9167, opened 2017-11-27
Make template parsing more flexible

seebs-markley:
A system we work with likes to send values with names like
	h1.Interface_xe0.in
	h1.Interface_xe0.out
	h1.Interface_bge0.in
	h1.Interface_bge0.out

We can parse these as "host.measurement.field", but then we have
measurements with names like "Interface_xe0" or "Interface_bge0".
In fact, in our current data set, that's over 1000 distinct
measurements.

This patch introduces a "subfield" functionality for templates.
In the new functionality, the template is:
	host.(_,measurement_interface).field
This results in all the measurements being named "Interface",
and having tags of
	{host: h1, interface: xe0}
	{host: h1, interface: bge0}
along with the expected set of fields.

The parsing is slightly complicated because of the existing
expectation that specifying a separator doesn't change how things
are split, only how they are rejoined when joining multiple
matching tags.

This patch also makes the '*' functionality generic, allowing
it to be used with arbitrary names, not just 'measurement' and
'field'. It also moves the check for more than one greedy name
into the initial template parse, since it's detectable at that
point, and the corresponding test is moved. The test for
multiple specifications of 'field' is dropped, because that
actually works perfectly well, and is just like a more selective
form of 'field*'.

Signed-off-by: seebs <seebs@markleygroup.com>

- [*] Rebased/mergable
- [*] Tests pass
- [*] CHANGELOG.md updated

(The CLA will probably take a bit to percolate to people who have authority, but I don't anticipate them objecting.)

- [*] Provide example syntax
- [*] Update man page when modifying a command

(Does "updated README.md" count, or do I need a separate issue for that?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - [#8495](https://github.com/influxdata/influxdb/pull/8495): Improve CLI connection warnings
 - [#9084](https://github.com/influxdata/influxdb/pull/9084): Handle high cardinality deletes in TSM engine
 - [#9162](https://github.com/influxdata/influxdb/pull/9162): Improve inmem index startup performance for high cardinality.
+- [#9165]: Allow fancier parsing for Graphite-protocol templates.
 
 ### Bugfixes
 
--- a/services/graphite/README.md
+++ b/services/graphite/README.md
@@ -23,7 +23,9 @@ To extract tags from metrics, one or more templates must be configured to parse 
 
 Templates allow matching parts of a metric name to be used as tag keys in the stored metric.  They have a similar format to Graphite metric names.  The values in between the separators are used as the tag keys.  The location of the tag key that matches the same position as the Graphite metric section is used as the value.  If there is no value, the Graphite portion is skipped.
 
-The special value _measurement_ is used to define the measurement name.  It can have a trailing `*` to indicate that the remainder of the metric should be used.  If a _measurement_ is not specified, the full metric name is used.
+The special value _measurement_ is used to define the measurement name.  If a _measurement_ is not specified, the full metric name is used.
+
+Any metric value can have a trailing ```*``` to indicate that it should use the rest of the metric name. You probably don't want to use this for anything but 'measurement' or 'field'. Only one metric value may have a trailing ```*``` in any given template.
 
 ### Basic Matching
 
@@ -33,8 +35,7 @@ The special value _measurement_ is used to define the measurement name.  It can 
 
 ### Multiple Measurement & Tags Matching
 
-The _measurement_ can be specified multiple times in a template to provide more control over the measurement name. Tags can also be
-matched multiple times. Multiple values will be joined together using the _Separator_ config variable.  By default, this value is `.`.
+A component can be specified multiple times in a template to provide more control over its contents. Multiple values will be joined together using the _Separator_ config variable.  By default, this value is `.`.
 
 `servers.localhost.localdomain.cpu.cpu0.user`
 * Template: `.host.host.measurement.cpu.measurement`
@@ -57,9 +58,7 @@ Additional tags can be added to a metric if they don't exist on the received met
 
 ### Fields
 
-A field key can be specified by using the keyword _field_. By default if no _field_ keyword is specified then the metric will be written to a field named _value_.
-
-The field key can also be derived from the second "half" of the input metric-name by specifying ```field*``` (eg ```measurement.measurement.field*```). This cannot be used in conjunction with "measurement*"!
+A field key can be specified by using the keyword _field_. By default if no _field_ keyword is specified then the metric will be written to a field named _value_. As with other keys, you can specify ```field*``` to mark the remainder of the input metric name as the field.
 
 It's possible to amend measurement metrics with additional fields, e.g:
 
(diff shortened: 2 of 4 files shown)


seebs-markley:
(signature should be done now.)

jsternberg:
I like this idea, but I'm also hesitant to try and add more complex parsing to the current format.

As an idea, how would you feel about being able to use regular expressions to parse the incoming graphite metrics? For most cases, it would be unnecessary and we would continue supporting and using the current parser as a default, but for more advanced cases it may be better.

seebs-markley:
Regular expressions seem like a reasonable thing. I was thinking about this more and am not sure my solution does a great job either. Really, in the use case I'm looking at, what I really want to do is recognize that "Interface_xe0" should be a measurement named "interface", with a tag "interface=xe0". (Which could, possibly, be handled by allowing "measurement" to be specified in a filtered rule the same way tags can be specified.) And in some cases, it might make sense to want to be able to match-and-discard specific literal strings.

Hmm.

Okay, so imagine a spelling like:

```/([^.]*).([^.]*).Interface_([^.*]).([^.]*)/ site=$1 host=$2 measurement=interface interface=$3 field=$4```

so if the "filter" starts with a slash, and ends with a slash and possible trailing suffix like "i", then we expect a list of tag-looking things, one of which should be "measurement".

I don't like how often I end up writing `([^.*])` in that, but it does work and seems like it would be handy. And that's specific to a given use case; not every setup would be like that. And it won't break any existing stuff unless people have existing filters with leading and trailing slashes, I guess? That would probably be a much nicer design than mine, which... well, it works but it was really feeling awkward, and I probably should have thought about that and taken a step back.

Thing I like about this: It allows you to combine the filter match and specification of fields. It can handle the multiple-field case in a couple of ways ("measurement=$1$2" for instance). Probably want to clean up the syntax on the assignment side to avoid ambiguity problems.

That said! I do think the underlying thing of converting way from field\*/measurement\* and just allowing any one field to get a *, and allowing arbitrary tags to be specified more than once rather than specially excluding field, is probably a good improvement to the existing parser, and won't break anything.

**Comment by seebs-markley** [comment]
Regular expressions seem like a reasonable thing. I was thinking about this more and am not sure my solution does a great job either. Really, in the use case I'm looking at, what I really want to do is recognize that "Interface_xe0" should be a measurement named "interface", with a tag "interface=xe0". (Which could, possibly, be handled by allowing "measurement" to be specified in a filtered rule the same way tags can be specified.) And in some cases, it might make sense to want to be able to match-and-discard specific literal strings.

Hmm.

Okay, so imagine a spelling like:

```/([^.]*).([^.]*).Interface_([^.*]).([^.]*)/ site=$1 host=$2 measurement=interface interface=$3 field=$4```

so if the "filter" starts with a slash, and ends with a slash and possible trailing suffix like "i", then we expect a list of tag-looking things, one of which should be "measurement".

I don't like how often I end up writing `([^.*])` in that, but it does work and seems like it would be handy. And that's specific to a given use case; not every setup would be like that. And it won't break any existing stuff unless people have existing filters with leading and trailing slashes, I guess? That would probably be a much nicer design than mine, which... well, it works but it was really feeling awkward, and I probably should have thought about that and taken a step back.

Thing I like about this: It allows you to combine the filter match and specification of fields. It can handle the multiple-field case in a couple of ways ("measurement=$1$2" for instance). Probably want to clean up the syntax on the assignment side to avoid ambiguity problems.

That said! I do think the underlying thing of converting way from field\*/measurement\* and just allowing any one field to get a *, and allowing arbitrary tags to be specified more than once rather than specially excluding field, is probably a good improvement to the existing parser, and won't break anything.

**Comment by jsternberg** [comment]
I'll work on a PR for the regular expression idea. Do you think you would be able to try it out to see if it's useful for you? I do agree that the number of `([^.]+)` parts is obnoxious, but it wouldn't be the default and would only be applied to uses like this that are more complex.

I'll take a look at the improvements you made to the parser and see if we can use some of that too. I noticed that the code is a bit awkward in some areas.

**Comment by seebs-markley** [comment]
oh, yes, the code is definitely awkward. I'm still sorta new to go, and I was still trying to rough out the design while I worked on it. I think a "tag" type with a "multi" flag is sensible, but the nested tag-or-tagset design is awful and should probably be cleansed with fire.

I'd probably be able to have a look at the regular-expression case. I'd also been thinking about a fancy thing with a Kapacitor plugin and basically having all the data written to a temporary database, with something that reads it in, cleans it up, and dumps it out... but that seems like a much heavier solution than "more flexible template parsing".

The other use case that comes immediately to mind when I think of template parsing, BTW:
```
siteA.hostA.Temperature_Inlet_Temperature_SensorInlet0.temp
siteA.hostA.Temperature_Inlet_Temperature_Sensordie_CPU.temp
siteA.hostA.Temperature_Inlet_Temperature_Sensordie_FIA.temp
```

With the current parser, we end up with "measurements" named things like Temperature_Inlet_Temperature_SensorInlet0, and fields named things like "temp". And it seems to me that in this case, we really want "measurement" temp, and "field" "Inlet_Temperature_SensorInlet0". (Or possibly even just "Inlet0".)

But the extra "Temperature_" in front is the sort of thing the regex matching could probably improve.

(I suppose from a usability standpoint, allowing an empty () to be treated as implicitly `([^.]*)` might help, but extra complexity for relatively little value.)

jsternberg:
#9194 is an alternate implementation for this problem.

PR #9192, opened 2017-12-01
ensure tsmWriter#Write returns ErrMaxBlocksExceeded

stuartcarnie:
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated


stuartcarnie:
@jwilder I updated the test so that the first file is 1 block short and the second file writes two blocks.

PR #9194, opened 2017-12-01
Add support for advanced graphite template parsing

jsternberg:
The old method of specifying templates is still supported, but an advanced
usage using an array of tables can also be used to use more advanced template
specifications.

Inside of the graphite section, you can do this:

    [[graphite.templates]]
      format = "simple"
      template = "measurement.field"

This section supports `format`, which specifies the format (default is simple);
`template`, which specifies the template text; and `filter`, which is a filter
that determines whether this template should be used.

The available formats are `simple` and `regexp`. The `simple` format is the
default format that is currently in place. The `regexp` format allows you to
specify a regular expression and set the values based on the named captures. So
the following regular expression:
`(?P<host>[^.]+)\\.(?P<measurement>[^.]+)_(?P<interface>[^.]+)\\.(?P<field>[^.]+)`
will produce tags for `host` and `interface`. The special named captures
`measurement` and `field` set the measurement and field. The `measurement`
named capture is required.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated


jsternberg:
@seebs-markley

seebs-markley:
A thing that occurred to me: It might make sense to check both subregexpnames *and* fixed/default tags for "measurement".

Consider the case where a strangely-configured thing produces graphite-format names which don't actually *contain* the string that would be the best measurement name. There's existing support in other templates for Additional Default Tags. If you had a reasonably-specific regex match, and then just specified a default tag of "measurement=foo", that would also make sense.

jsternberg:
If I understand correctly, that feature would apply to the original simple format too, correct?

seebs-markley:
That's actually not a bad idea at all. It wouldn't make much *sense* in the absence of a filter, but I could easily see cases where, with a given filter, that would be a good choice.

krystiancharubin:
If we want to specify multiple advanced templates then we would have multiple 
[[graphite.template]]?
Any updates on getting this merged?

Issue #3019, opened 2015-06-17
File name of the snapshot created on backup


leelasharma:
For incremental backups(or even full snapshot) it would be nice if the output also had the file name of the numbered snapshot just created.

beckettsean:
@leelasharma can you provide an example of the output you want? We don't quite follow what you're asking for.

leelasharma:
I meant something along the lines of  either a) accept a -snapshot-filename or -backup-filename as an argument or 2) a message like backup complete, available at <filename>

beckettsean:
So you would like to be able to either provide a target filename for the snapshot, or receive a message to STDOUT containing the snapshot filename once written?

leelasharma:
@beckettsean  yes.

zstyblik:
> or receive a message to STDOUT containing the snapshot filename once written

+1

PR #9207, opened 2017-12-07
Update restore functionality to run in online mode, consume Enterprise backup files. 

aanthony1243:
Added a flag -online  that will run a database restore on a live DB instance.  A side effect of this change is that the existing DB's in the destination system will be unaffected, while the imported database must not exist in the system.  

Also added a -enterprise flag that will enable restore to consume enterprise data files.  This mode always runs as an online update.  

Existing restore functionality is preserved.  

- [X] Rebased/mergable
- [X] Tests pass
- [X] CHANGELOG.md updated

- [X] Update man page when modifying a command
- [X] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): issue filed or pull request submitted \<link to issue or pull request\>


pkg/tar/file_unix.go
```diff
@@ -0,0 +1,20 @@
+// +build !windows
```
aanthony1243: copied from plutonium.

pkg/tar/file_windows.go
```diff
@@ -0,0 +1,19 @@
+package tar
```
aanthony1243: copied from plutonium

pkg/tar/stream.go
```diff
@@ -0,0 +1,151 @@
...
+	"archive/tar"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+)
+
+func StreamFunc(w io.Writer, dir, relativePath string, writeFunc func(f os.FileInfo, shardRelativePath, fullPath string, tw *tar.Writer) error) error {
```
aanthony1243: copied and adapted from plutonium to be more flexible for the OSS use-case where we filter files before they go to the archive.  the original API is below, and is functionally equivalent to the original.

cmd/influxd/restore/restore.go
```diff
@@ -212,11 +295,134 @@ func (cmd *Command) unpackMeta() error {
...
+	}
+	defer conn.Close()
+
+	if _, err := conn.Write([]byte{byte(req.Type)}); err != nil {
+		return b.Bytes(), err
+	}
+
+	if req.Type != snapshotter.RequestShardUpdate { // Write the request
+		if err := json.NewEncoder(conn).Encode(req); err != nil {
+			return b.Bytes(), fmt.Errorf("encode snapshot request: %s", err)
```
joelegasse: I already mentioned it should be changed, anyway, but if `[]byte` were a good return type here, this should still be `return nil, err`. `b` hasn't seen any writes up to this point.

cmd/influxd/restore/restore.go
```diff
@@ -212,11 +295,134 @@ func (cmd *Command) unpackMeta() error {
...
+		}
+	}
+
+	if n, err := io.Copy(conn, upStream); (err != nil && err != io.EOF) || n != req.UploadSize {
+		return b.Bytes(), fmt.Errorf("error uploading file: err=%v, n=%d, uploadSize: %d", err, n, req.UploadSize)
+	}
+
+	cmd.StdoutLogger.Printf("wrote %d bytes", req.UploadSize)
+
+	b.Reset()
```
joelegasse: `b` was declared at the beginning of this function, but received no writes, and is being `Reset` here? Why not just use `ioutil.ReadAll`? I don't see the benefit in using `Buffer` here.

cmd/influxd/restore/restore.go
```diff
@@ -212,11 +295,134 @@ func (cmd *Command) unpackMeta() error {
...
+	if n, err := b.ReadFrom(conn); err != nil || n == 0 {
+		return b.Bytes(), fmt.Errorf("updating metadata on influxd service failed: err=%v, n=%d", err, n)
+	}
+
+	return b.Bytes(), nil
```
joelegasse: Decode the bytes, don't just return arbitrary data.

cmd/influxd/restore/restore.go
```diff
@@ -227,41 +433,153 @@ func (cmd *Command) unpackShard(shardID string) error {
...
-	pat := filepath.Join(cmd.backupFilesPath, fmt.Sprintf(backup_util.BackupFilePattern, cmd.database, cmd.retention, id))
+	pat := filepath.Join(cmd.backupFilesPath, fmt.Sprintf(backup_util.BackupFilePattern, cmd.destinationDatabase, cmd.restoreRetention, id))
 	return cmd.unpackFiles(pat + ".*")
 }
 
-// unpackDatabase will look for all backup files in the path matching this database
+// unpackFiles will look for backup files matching the pattern and restore them to the data dir
+func (cmd *Command) uploadShardsLive() error {
+	// find the destinationDatabase backup files
+	if cmd.enterprise {
```
joelegasse: abstract this away

cmd/influxd/restore/restore.go
```diff
@@ -10,29 +10,53 @@ import (
...
+	path string
+
+	backupFilesPath     string
+	metadir             string
+	datadir             string
+	destinationDatabase string
+	sourceDatabase      string
+	backupRetention     string
+	restoreRetention    string
+	shard               string
```
joelegasse: Probably update this to be a `uint` or `int` while we're in here updating things.

cmd/influxd/restore/restore.go
```diff
@@ -227,41 +433,153 @@ func (cmd *Command) unpackShard(shardID string) error {
...
+// unpackFiles will look for backup files matching the pattern and restore them to the data dir
+func (cmd *Command) uploadShardsLive() error {
+	// find the destinationDatabase backup files
+	if cmd.enterprise {
+		for _, file := range cmd.manifest.Files {
+			if cmd.sourceDatabase == "" || cmd.sourceDatabase == file.Database {
+				if cmd.backupRetention == "" || cmd.backupRetention == file.Policy {
+					if cmd.shard == "" || cmd.shard == strconv.FormatUint(file.ShardID, 10) {
+						cmd.StdoutLogger.Printf("Restoring shard %d live from backup %s\n", file.ShardID, file.FileName)
+						f, err := os.Open(filepath.Join(cmd.backupFilesPath, file.FileName))
```
joelegasse: This is never closed.

cmd/influxd/restore/restore.go
```diff
@@ -227,41 +433,153 @@ func (cmd *Command) unpackShard(shardID string) error {
...
+			parts := strings.Split(fn, ".")
+
+			if len(parts) != 4 {
+				cmd.StderrLogger.Printf("Skipping mis-named backup file: %s", fn)
+			}
+			shardID, err := strconv.ParseUint(parts[2], 10, 64)
+			if err != nil {
+				return err
+			}
+			f, err := os.Open(fn)
```
joelegasse: File is never closed.

cmd/influxd/restore/restore.go
```diff
@@ -227,41 +433,153 @@ func (cmd *Command) unpackShard(shardID string) error {
...
+	defer conn.Close()
+	if err != nil {
+		return err
+	}
+
+	if _, err := conn.Write([]byte{byte(snapshotter.RequestShardUpdate)}); err != nil {
+		return err
+	}
+
+	// 0.  write the shard ID to pw
```
joelegasse: What is `pw`? I also feel like this code should be abstracted away, maybe in the `snapshotter` package, so that clients don't have to know the exact byte order and incantations to perform operations.

cmd/influxd/restore/restore.go
```diff
@@ -227,41 +433,153 @@ func (cmd *Command) unpackShard(shardID string) error {
...
+
+	for {
+		hdr, err := tr.Next()
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			return err
+		}
+
+		names := strings.Split(hdr.Name, "/")
```
joelegasse: This feels like it should be using something other than `strings.Split`, considering you don't just use `strings.Join( ... , "/")` later.

cmd/influxd/restore/restore.go
```diff
@@ -227,41 +433,153 @@ func (cmd *Command) unpackShard(shardID string) error {
...
-	restorePath := filepath.Join(cmd.datadir, cmd.database)
+	restorePath := filepath.Join(cmd.datadir, cmd.destinationDatabase)
 	if _, err := os.Stat(restorePath); err != nil && !os.IsNotExist(err) {
-		return fmt.Errorf("database already present: %s", restorePath)
+		return fmt.Errorf("destinationDatabase already present: %s", restorePath)
 	}
 
-	// find the database backup files
-	pat := filepath.Join(cmd.backupFilesPath, cmd.database)
+	// find the destinationDatabase backup files
```
joelegasse: "search-and-replace-all" strikes again?

cmd/influxd/restore/restore.go
```diff
@@ -227,41 +433,153 @@ func (cmd *Command) unpackShard(shardID string) error {
...
+
+// unpackDatabase will look for all backup files in the path matching this destinationDatabase
 // and restore them to the data dir
 func (cmd *Command) unpackDatabase() error {
 	// make sure the shard isn't already there so we don't clobber anything
-	restorePath := filepath.Join(cmd.datadir, cmd.database)
+	restorePath := filepath.Join(cmd.datadir, cmd.destinationDatabase)
 	if _, err := os.Stat(restorePath); err != nil && !os.IsNotExist(err) {
-		return fmt.Errorf("database already present: %s", restorePath)
+		return fmt.Errorf("destinationDatabase already present: %s", restorePath)
```
joelegasse: "search-and-replace-all" gone wrong?

cmd/influxd/restore/restore.go
```diff
@@ -289,27 +607,20 @@ func (cmd *Command) unpackTar(tarFile string) error {
...
+	pathParts := strings.Split(filepath.Base(tarFile), ".")
 
-	for {
-		hdr, err := tr.Next()
-		if err == io.EOF {
-			return nil
-		} else if err != nil {
-			return err
-		}
+	shardPath := filepath.Join(cmd.datadir, pathParts[0], pathParts[1], strings.Trim(pathParts[2], "0"))
```
joelegasse: Length check missing. `pathParts[1]` could panic?

cmd/influxd/restore/restore.go
```diff
@@ -319,37 +630,68 @@ func (cmd *Command) unpackFile(tr *tar.Reader, fileName string) error {
...
-	defer ff.Close()
 
 	if _, err := io.Copy(ff, tr); err != nil {
+		if err := ff.Close(); err != nil {
+			return err
+		}
 		return err
 	}
 
+	if err := ff.Close(); err != nil {
```
joelegasse: `return ff.Close()`

cmd/influxd/restore/restore.go
```diff
@@ -319,37 +630,68 @@ func (cmd *Command) unpackFile(tr *tar.Reader, fileName string) error {
...
             TSM files.
     -shard <id>
-            Optional. If given, database and retention are required. Will restore the shard's
+            Optional. If given, -database and -retention are required. Will restore the shard's
             TSM files.
+	-online
+	        Optional. If given, the restore will be done using the new process, detailed below.  All above legacy
+	        arguments are ignored.
+
+The -enterprise restore mode consumes files in an improved format that includes a file manifest.  PATH should indicate
```
joelegasse: If we're going to call this `enterprise`, it should mimic the enterprise restore, accepting either a single manifest or a directory containing multiple manifests (as created by incremental backups)

man/influxd-restore.txt
```diff
@@ -3,32 +3,71 @@ influxd-restore(1)
...
 ----
-influxd-restore - Restores the metastore, databases, retention policies, or specific shards
+influxd-restore - Restores the metastore, databases, retention policies, or specific shards.  Legacy mode requires
+the instance to be stopped before running, and will wipe all databases from the system (e.g., for disaster recovery).
+The improved online mode requires the instance to be running, and the database name used must not already exist.
 
 SYNOPSIS
 --------
-'influxd restore' [options] PATH
+'influxd restore' [-legacy] [flags] PATH
```
joelegasse: Is it `-legacy` or `-enterprise`? Which is the default?

man/influxd-restore.txt
```diff
@@ -3,32 +3,71 @@ influxd-restore(1)
...
 
 -datadir <path>::
-  If set, the restore process will recover the specified database, retention policy, or shard to the given directory. Optional.
+            Optional. If set the restore process will recover the specified
+            destinationDatabase, retention policy or shard to the given directory.
 
 -database <name>::
-  Will restore the database TSM files. Required if no metadir is given. Optional.
+            Optional. Required if no metadir given. Will restore the database's
+            TSM files.
```
joelegasse: Just say "a single database's data". TSM files are an implementation detail of the current storage engine.

man/influxd-restore.txt
```diff
@@ -3,32 +3,71 @@ influxd-restore(1)
...
 
 -retention <name>::
-  Will restore the retention policy's TSM files. If given, database is required. Optional.
+            Optional. If given, -database is required. Will restore the retention policy's
+            TSM files.
```
joelegasse: Don't reference TSM files

services/meta/data.go
```diff
@@ -783,6 +784,91 @@ func (data *Data) hasAdminUser() bool {
 	return false
 }
 
+// ImportData imports selected data into the current metadata.
+// opts provides options to possibly rename the Database or Retention Policy, or change the Replication Factor.
+func (data *Data) ImportData(other Data, backupDBName, restoreDBName, backupRPName, restoreRPName string) (map[uint64]uint64, error) {
+
+	if backupDBName == "" {
```
joelegasse: Invert this for the early single-DB case, and then un-indent the rest of the function

services/meta/data.go
```diff
@@ -783,6 +784,91 @@ func (data *Data) hasAdminUser() bool {
...
+		for _, dbi := range other.Databases {
+			if dbi.Name == "_internal" {
+				continue
+			}
+			tmpMap, err := data.ImportOneDB(other, dbi.Name, "", "", "")
+			if err != nil {
+				return shardIDMap, err
+			}
+
+			for k, v := range tmpMap {
```
joelegasse: Maybe just pass the shard map into the function, rather than copying all the entries around? Also, perhaps `ImportOneDB` should be unexported?

services/meta/data.go
```diff
@@ -783,6 +784,91 @@ func (data *Data) hasAdminUser() bool {
...
+	if dbImport == nil {
+		return shardIDMap, fmt.Errorf("imported metadata does not have datbase named %s", backupDBName)
+	}
+
+	if restoreDBName == "" {
+		restoreDBName = backupDBName
+	}
+
+	// change the names if we want/need to
+	dbImport.Name = restoreDBName
```
joelegasse: Don't do this, `dbImport` is a pointer. You're modifying data that should be read-only for purposes of this operation.

services/meta/data.go
```diff
@@ -783,6 +784,91 @@ func (data *Data) hasAdminUser() bool {
...
+
+	if backupRPName != "" {
+		rpImport, err := other.RetentionPolicy(restoreDBName, backupRPName)
```
joelegasse: Use `dbImport.RetentionPolicy`

services/meta/data.go
```diff
@@ -783,6 +784,91 @@ func (data *Data) hasAdminUser() bool {
...
+				restoreRPName = backupRPName
+			}
+			rpImport.Name = restoreRPName
+			dbImport.RetentionPolicies = []RetentionPolicyInfo{*rpImport}
+		} else {
+			return shardIDMap, fmt.Errorf("retention Policy not found in meta backup: %s", backupRPName)
+		}
+
+	}
+	if data.Database(restoreDBName) != nil {
```
joelegasse: Check this first, before scanning the old data. No need to go through all of the above if there's a simple and quick way to check for failure.

services/meta/data.go
joelegasse: You should create a new database info object instead.

services/meta/data.go
```diff
@@ -783,6 +784,91 @@ func (data *Data) hasAdminUser() bool {
...
+		return shardIDMap, errors.New("database already exists")
+	}
+
+	for _, rpImport := range dbImport.RetentionPolicies {
```
joelegasse: I would strongly recommend updating this to create new info objects throughout this code, rather than mutating what's already here. The metadata structs are full of slices, so you may be inadvertently modifying shared data.

services/snapshotter/service.go
```diff
@@ -274,12 +359,46 @@ func (s *Service) writeRetentionPolicyInfo(conn net.Conn, database, retentionPol
...
 // readRequest unmarshals a request object from the conn.
-func (s *Service) readRequest(conn net.Conn) (Request, error) {
+func (s *Service) readRequest(conn net.Conn) (Request, []byte, error) {
 	var r Request
-	if err := json.NewDecoder(conn).Decode(&r); err != nil {
-		return r, err
+	d := json.NewDecoder(conn)
+
+	if err := d.Decode(&r); err != nil {
+		return r, []byte{}, err
```
joelegasse: `nil` should be used for an empty slice

services/snapshotter/service.go
```diff
@@ -274,12 +359,46 @@ func (s *Service) writeRetentionPolicyInfo(conn net.Conn, database, retentionPol
...
 	}
-	return r, nil
+
+	bits := make([]byte, r.UploadSize+1)
+
+	if r.UploadSize > 0 {
+
+		remainder := d.Buffered()
+
+		//// no time to investigate.  It seems the JSON encoder puts one extra byte on the stream,
```
joelegasse: What is going on here?

joelegasse: OK, I just read through... and it looks like at one point the protocol is `[byte].[byte].[JSON-object].[binary-data]`... That seems far too fragile to be useful or maintainable.

services/snapshotter/service.go
```diff
@@ -274,12 +359,46 @@ func (s *Service) writeRetentionPolicyInfo(conn net.Conn, database, retentionPol
...
+		// it is a bit random but sometimes the Json decoder will consume all the bytes and sometimes
+		// it will leave a few behind.
+		if err != io.EOF && n < int(r.UploadSize+1) {
+			n, err = conn.Read(bits[n:])
+		}
+
+		if err != nil && err != io.EOF {
+			return r, bits, err
+		}
+		return r, bits[1:], nil
```
joelegasse: Why is one byte being stripped off?

services/snapshotter/service.go
```diff
@@ -274,12 +359,46 @@ func (s *Service) writeRetentionPolicyInfo(conn net.Conn, database, retentionPol
...
+
+		n, err := remainder.Read(bits)
+		if err != nil && err != io.EOF {
+			return r, bits, err
+		}
+
+		// it is a bit random but sometimes the Json decoder will consume all the bytes and sometimes
+		// it will leave a few behind.
+		if err != io.EOF && n < int(r.UploadSize+1) {
+			n, err = conn.Read(bits[n:])
```
joelegasse: In general, prefer `ioutil.ReadAll` or `io.ReadFull` to ensure draining of readers or reading sufficient data.

services/meta/data.go
aanthony1243: 👍

services/meta/data.go
aanthony1243: 👍

services/meta/data.go
aanthony1243: 👍

services/meta/data.go
aanthony1243: 👍

services/meta/data.go
aanthony1243: solved using dbInfo.clone()  which does a deep copy.

services/snapshotter/service.go
aanthony1243: 👍

services/snapshotter/service.go
aanthony1243: the JSON encoder seems to put an extra byte on the stream on the client side.  I'll add a comment.

cmd/influxd/restore/restore.go
```diff
@@ -212,11 +295,134 @@ func (cmd *Command) unpackMeta() error {
 	return nil
 }
 
+// updateMetaLive takes a metadata backup and sends it to the influx server
+// for a live merger of metadata.
+func (cmd *Command) updateMetaLive() error {
+
+	var metaBytes []byte
+	if cmd.enterprise {
```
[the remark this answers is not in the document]
aanthony1243: I'm not sure I agree with this point.  I did tidy up the logic a bit so it's not "scattered around" so much.

cmd/influxd/restore/restore.go
```diff
@@ -212,11 +295,134 @@ func (cmd *Command) unpackMeta() error {
...
+		fileBytes, err := ioutil.ReadFile(fileName)
+		if err != nil {
+			return err
+		}
+		if cmd.manifest.Platform == "OSS" {
+			metaBytes = fileBytes
+		} else if cmd.manifest.Platform == "ENT" || cmd.manifest.Platform == "" {
+			// allow "" to be compatible with pre-1.5 Enterprise backups
+			var pb EnterpriseData
+			if err := proto.Unmarshal(fileBytes, &pb); err != nil {
```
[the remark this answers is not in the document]
aanthony1243: will get to this Monday.

pkg/tar/stream.go
jwilder: Does this need to be exported?  Also, needs docs.

pkg/tar/stream.go
```diff
@@ -0,0 +1,151 @@
...
+	})
+}
+
+// Stream writes a tar archive to w, containing the files and sub-directories in the directory dir
+// stored relative to relativePath.
+func Stream(w io.Writer, dir, relativePath string) error {
+	return StreamFunc(w, dir, relativePath, TarFile)
+}
+
+func SinceFilterTarFile(since time.Time) func(f os.FileInfo, shardRelativePath, fullPath string, tw *tar.Writer) error {
```
jwilder: Needs docs.

pkg/tar/stream.go
```diff
@@ -0,0 +1,151 @@
...
+	return func(f os.FileInfo, shardRelativePath, fullPath string, tw *tar.Writer) error {
+		if f.ModTime().After(since) {
+			return TarFile(f, shardRelativePath, fullPath, tw)
+		}
+		return nil
+	}
+}
+
+func TarFile(f os.FileInfo, shardRelativePath, fullPath string, tw *tar.Writer) error {
```
jwilder: Does this need to be exported?

jwilder: Also, has a stutter `tar.TarFile`.

pkg/tar/stream.go
jwilder: `StreamFunc` is an odd name for a function to call.  Maybe have `Stream` take an optional func arg that defaults to the `TarFile` if nil?

pkg/tar/stream.go
aanthony1243: 👍

pkg/tar/stream.go
aanthony1243: will doc new Stream function, but yes this should be exported.  The design is that you can write a custom filter for Stream.  For example, see timeStampFilterTarFile    in engine.go, which is too specific to be in pkg/tar.  A custom function may, and in many cases should, use TarFile internally.  Will rename it to remove stutter.

pkg/tar/stream.go
aanthony1243: 👍   needs to be exported, esp since it now replaces Stream(...)

services/snapshotter/client.go
```diff
@@ -22,6 +26,138 @@ func NewClient(host string) *Client {
...
+
+	if n, err := io.Copy(conn, upStream); (err != nil && err != io.EOF) || n != req.UploadSize {
+		return nil, fmt.Errorf("error uploading file: err=%v, n=%d, uploadSize: %d", err, n, req.UploadSize)
+	}
+
+	if n, err := b.ReadFrom(conn); err != nil || n == 0 {
+		return nil, fmt.Errorf("updating metadata on influxd service failed: err=%v, n=%d", err, n)
+	}
+
+	resp := b.Bytes()
```
joelegasse: In this scenario, prefer `ioutil.ReadAll` rather than creating a `bytes.Buffer`, using `ReadFrom`, and then only calling `Bytes()`.

services/snapshotter/client.go
```diff
@@ -22,6 +26,138 @@ func NewClient(host string) *Client {
...
+		}
+		shardIDMap[k] = v
+	}
+
+	return shardIDMap, nil
+}
+
+func (c *Client) decodeUintPair(bits []byte) (uint64, uint64, error) {
+	if len(bits) != 16 {
+		return 0, 0, fmt.Errorf("array must have exactly 16 bytes")
```
joelegasse: `errors.New`, there is no formatting occurring here. Also, the error message says "array", when it is talking about a slice.

services/snapshotter/client.go
```diff
@@ -22,6 +26,138 @@ func NewClient(host string) *Client {
...
+		return nil, fmt.Errorf("updating metadata on influxd service failed: err=%v, n=%d", err, n)
+	}
+
+	resp := b.Bytes()
+
+	if err != nil {
+		return nil, err
+	}
+
+	header, npairs, err := c.decodeUintPair(resp[:16])
```
joelegasse: No length check here, could panic.

services/snapshotter/client.go
```diff
@@ -22,6 +26,138 @@ func NewClient(host string) *Client {
...
+		if err != nil {
+			return nil, err
+		}
+		shardIDMap[k] = v
+	}
+
+	return shardIDMap, nil
+}
+
+func (c *Client) decodeUintPair(bits []byte) (uint64, uint64, error) {
```
joelegasse: This does not need to be a method on `Client`, it doesn't use any fields of the client, it is just a helper function.

services/snapshotter/client.go
```diff
@@ -22,6 +26,138 @@ func NewClient(host string) *Client {
...
+	if header != BackupMagicHeader {
+		return nil, fmt.Errorf("Response did not contain the proper header tag.")
+	}
+
+	if (len(pairs)/8)%2 != 0 {
+		return nil, fmt.Errorf("expected an even number of integer pairs in update meta repsonse")
+	}
+
+	shardIDMap := make(map[uint64]uint64)
+	for i := 0; i < int(npairs); i++ {
```
joelegasse: `for i := 0; i < len(pairs); i+= 16 {`

services/snapshotter/client.go
```diff
@@ -22,6 +26,138 @@ func NewClient(host string) *Client {
...
+		offset := i * 16
+		k, v, err := c.decodeUintPair(pairs[offset : offset+16])
```
joelegasse: Inline the two calls to `binary.BigEndian.Uint64`, you've already checked the length (once you update the logic above)

services/snapshotter/client.go
```diff
@@ -22,6 +26,138 @@ func NewClient(host string) *Client {
...
+		return nil, fmt.Errorf("DB metadata not changed. database may already exist")
+	}
+
+	pairs := resp[16:]
+
+	if header != BackupMagicHeader {
+		return nil, fmt.Errorf("Response did not contain the proper header tag.")
+	}
+
+	if (len(pairs)/8)%2 != 0 {
```
joelegasse: The logic here is all over the place. You already have the number of expected pairs (`npairs`), the expected length is then `npairs * 16`. If the length isn't exactly that, there was a problem.

services/snapshotter/client.go
```diff
@@ -22,6 +26,138 @@ func NewClient(host string) *Client {
...
+		return nil, fmt.Errorf("expected an even number of integer pairs in update meta repsonse")
+	}
+
+	shardIDMap := make(map[uint64]uint64)
```
joelegasse: Provide a size hint here: `shardIDMap := make(map[uint64]uint64, npairs)`

services/snapshotter/client.go
```diff
@@ -22,6 +26,138 @@ func NewClient(host string) *Client {
...
+	defer conn.Close()
+	if err != nil {
+		return err
+	}
+
+	if _, err := conn.Write([]byte{byte(RequestShardUpdate)}); err != nil {
+		return err
+	}
+
+	var shardBytes [8]byte
```
joelegasse: Minor nit: make this `[9]byte` and put `RequestShardUpdate` in the first byte, then encode the ID into `shardBytes[1:9]`. You're currently making two small writes of known length and content back-to-back, it would be better to combine them.

pkg/tar/stream.go
```diff
@@ -0,0 +1,151 @@
...
+		if err != nil {
+			return err
+		}
+
+		// Skip adding an entry for the root dir
+		if dir == path && f.IsDir() {
+			return nil
+		}
+
+		// Figure out the the full relative path including any sub-dirs
```
joelegasse: Use `filepath.Rel`

pkg/tar/stream.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil
+	}
+}
+
+func TarFile(f os.FileInfo, shardRelativePath, fullPath string, tw *tar.Writer) error {
+	h, err := tar.FileInfoHeader(f, f.Name())
+	if err != nil {
+		return err
+	}
+	h.Name = filepath.Join(shardRelativePath, f.Name())
```
joelegasse: This should probably use `filepath.ToSlash` to keep all paths in tar forward slash-based, and then use `filepath.FromSlash` on extraction. Otherwise backups from one system might not be able to be restored onto a system with a different separator.

cmd/influxd/restore/restore.go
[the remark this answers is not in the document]
joelegasse: It's still adds large blocks of code that a split by a single `else`. This needs to either be abstracted away as a common interface, or just split the codepaths (call different methods).

cmd/influxd/restore/restore.go
joelegasse: 👎 It is not closed in the case of an error

cmd/influxd/restore/restore.go
joelegasse: :-1: handle the error case as well

cmd/influxd/restore/restore.go
```diff
@@ -227,41 +367,108 @@ func (cmd *Command) unpackShard(shardID string) error {
...
+
+						if err := cmd.client.UploadShard(file.ShardID, cmd.shardIDMap[file.ShardID], cmd.destinationDatabase, cmd.restoreRetention, tr); err != nil {
+							return err
+						}
+						f.Close()
+					}
+				}
+			}
+		}
+	} else {
```
joelegasse: Split this off somehow. Two giant blocks of code split by a single `else` makes it difficult to maintain.

cmd/influxd/restore/restore.go
```diff
@@ -212,11 +295,134 @@ func (cmd *Command) unpackMeta() error {
...
+
+	var metaBytes []byte
+	if cmd.enterprise {
+		fileName := filepath.Join(cmd.backupFilesPath, cmd.manifest.Meta.FileName)
+
+		fileBytes, err := ioutil.ReadFile(fileName)
+		if err != nil {
+			return err
+		}
+		if cmd.manifest.Platform == "OSS" {
```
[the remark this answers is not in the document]
joelegasse: If that's the case, this needs to be called something else, then. "enterprise" is not a state of mind, it is a different version of the software. The open-source database does not produce "enterprise" backups, it's just an updated backup format on-disk. Calling it "enterprise" but allowing it come from the open-source database is very confusing.

cmd/influxd/restore/restore.go
```diff
@@ -289,67 +496,65 @@ func (cmd *Command) unpackTar(tarFile string) error {
...
-	nativeFileName := filepath.FromSlash(fileName)
-	fn := filepath.Join(cmd.datadir, nativeFileName)
-	fmt.Printf("unpacking %s\n", fn)
-
-	if err := os.MkdirAll(filepath.Dir(fn), 0777); err != nil {
-		return fmt.Errorf("error making restore dir: %s", err.Error())
+	// should get us ["db","rp", "00001", "00"]
+	pathParts := strings.Split(filepath.Base(tarFile), ".")
+	if len(pathParts) != 4 {
+		return fmt.Errorf("backup tarfile name incorrect length")
```
joelegasse: It's not the length that is incorrect, it's the format.

services/snapshotter/client.go
aanthony1243: 👍

services/snapshotter/client.go
aanthony1243: 👍

services/snapshotter/client.go
aanthony1243: 👍

cmd/influxd/restore/restore.go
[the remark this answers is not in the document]
aanthony1243: fixed

cmd/influxd/restore/restore.go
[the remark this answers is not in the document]
aanthony1243: fixed, but the tag is still in the manifest until I can change + confirm on plutonium.

aanthony1243:
@joelegasse   should have covered all your change requests.  I rolled in a small feature #3019 to backup that collects a file listing.  Initially I was using it to drive some tests, though your change requests designed that out.    Also added is support for incremental manifest file collections.

rbetts:
This PR has been open for a long time and has been read and iterated against by reviewers and by Adam. I understand that there are some lingering differences but at this point I've decided to allow this code to master. 

The most recent addition to the PR is largely my fault - I asked Adam to simultaneously hold this work back from master until TSI landed and to go ahead and implement the remaining functionality to support restore of incremental backups to OSS.

Adam is going to put together docs for the supported changes. I'll arrange a review of those with PM.

Thank you everyone for working together on this - I know it has been a challenging experience.

PR #9211, opened 2017-12-08
pkg/tar moved here from plutonium and extended a bit to help with OSS restore

aanthony1243:
reviewing separate from OSS restore since it stands alone.  

- [X] Rebased/mergable
- [X] Tests pass


aanthony1243:
aa_8880_liverestore includes some bugfixes, we can merge it there.

PR #9232, opened 2017-12-15
[BUGFIX] client url path

### LINKED ISSUE #9154: Client - Allow path in address
### Bug report

__Steps to reproduce:__
https://github.com/influxdata/influxdb/blob/2ed0d2d1c918ea4624dfe45dad0e6c5fd288e56f/client/v2/client.go#L26
With a `Addr` like `https://example.org/influxdb/` like behind a proxy server or so

concat path with e.g. `ping`.

override path with e.g. `ping`.

https://github.com/influxdata/influxdb/blob/2ed0d2d1c918ea4624dfe45dad0e6c5fd288e56f/client/v2/client.go#L123

https://github.com/influxdata/influxdb/blob/2ed0d2d1c918ea4624dfe45dad0e6c5fd288e56f/client/v2/client.go#L370

https://github.com/influxdata/influxdb/blob/2ed0d2d1c918ea4624dfe45dad0e6c5fd288e56f/client/v2/client.go#L476

https://github.com/influxdata/influxdb/blob/2ed0d2d1c918ea4624dfe45dad0e6c5fd288e56f/client/influxdb.go#L191

https://github.com/influxdata/influxdb/blob/2ed0d2d1c918ea4624dfe45dad0e6c5fd288e56f/client/influxdb.go#L267

https://github.com/influxdata/influxdb/blob/2ed0d2d1c918ea4624dfe45dad0e6c5fd288e56f/client/influxdb.go#L345

https://github.com/influxdata/influxdb/blob/2ed0d2d1c918ea4624dfe45dad0e6c5fd288e56f/client/influxdb.go#L391

genofire:
fix #9154 

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/client/influxdb.go
+++ b/client/influxdb.go
@@ -14,6 +14,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"path"
 	"strconv"
 	"strings"
 	"time"
@@ -196,8 +197,8 @@ func (c *Client) Query(q Query) (*Response, error) {
 // It uses a context that can be cancelled by the command line client
 func (c *Client) QueryContext(ctx context.Context, q Query) (*Response, error) {
 	u := c.url
+	u.Path = path.Join(u.Path, "query")
 
-	u.Path = "query"
 	values := u.Query()
 	values.Set("q", q.Command)
 	values.Set("db", q.Database)
@@ -276,7 +277,7 @@ func (c *Client) QueryContext(ctx context.Context, q Query) (*Response, error) {
 // If an error occurs, Response may contain additional information if populated.
 func (c *Client) Write(bp BatchPoints) (*Response, error) {
 	u := c.url
-	u.Path = "write"
+	u.Path = path.Join(u.Path, "write")
 
 	var b bytes.Buffer
 	for _, p := range bp.Points {
@@ -354,7 +355,7 @@ func (c *Client) Write(bp BatchPoints) (*Response, error) {
 // If an error occurs, Response may contain additional information if populated.
 func (c *Client) WriteLineProtocol(data, database, retentionPolicy, precision, writeConsistency string) (*Response, error) {
 	u := c.url
-	u.Path = "write"
+	u.Path = path.Join(u.Path, "write")
 
 	r := strings.NewReader(data)
 
@@ -399,8 +400,9 @@ func (c *Client) WriteLineProtocol(data, database, retentionPolicy, precision, w
 // Ping returns how long the request took, the version of the server it connected to, and an error if one occurred.
 func (c *Client) Ping() (time.Duration, string, error) {
 	now := time.Now()
+
 	u := c.url
-	u.Path = "ping"
+	u.Path = path.Join(u.Path, "ping")
 
 	req, err := http.NewRequest("GET", u.String(), nil)
 	if err != nil {
--- a/client/v2/client.go
+++ b/client/v2/client.go
@@ -12,6 +12,7 @@ import (
 	"mime"
 	"net/http"
 	"net/url"
+	"path"
 	"strconv"
 	"strings"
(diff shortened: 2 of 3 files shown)


genofire:
@e-dard @benbjohnson @hpbieker @jsternberg could somebody make any reaction - i am feeling like forgotten

hpbieker:
Hi @genofire! Thank you for mentioning me alongside InfluxDB's great heros, but I am just a poor/happy user of InfluxDB :-)

client/influxdb.go
```diff
@@ -186,9 +187,9 @@ func (c *Client) Query(q Query) (*Response, error) {
 // QueryContext sends a command to the server and returns the Response
 // It uses a context that can be cancelled by the command line client
 func (c *Client) QueryContext(ctx context.Context, q Query) (*Response, error) {
-	u := c.url
+	u, _ := url.Parse(c.url.String())
```
e-dard: Please don't ignore any errors from `url.Parse`. Return them to the caller

e-dard:
Hi @genofire, sorry your PR got overlooked. I've checked the CLA, but I can't find you listed anywhere. Are you sure you signed [this form here?](https://www.influxdata.com/community/cla/)

genofire:
Better now? - i  did not know why your ci failed.

client/influxdb.go
```diff
@@ -195,9 +196,13 @@ func (c *Client) Query(q Query) (*Response, error) {
 // QueryContext sends a command to the server and returns the Response
 // It uses a context that can be cancelled by the command line client
 func (c *Client) QueryContext(ctx context.Context, q Query) (*Response, error) {
-	u := c.url
+	// clone url struct
```
jsternberg: I don't like this cloning the struct thing. The original line, `u := c.url`, already does a soft clone of the struct without the additional requirement to render a string and then reparse that string back into the struct. The only thing that doesn't get cloned properly is the `.User` variable on the struct, but that isn't being modified so a shallow clone should be fine.

It seems to me like it would be better to just modify line 200 in the original diff to:

```
u.Path += "/query"
```

And that's all that would be required to implement this feature. If the path is empty, the path should still be preceded by a slash anyway and it outputs the correct output. See this example: https://play.golang.org/p/veRZfeI9Grj

genofire: soft clone did not work, it is a pointer:
https://play.golang.org/p/X9vztSPdo4P

genofire: ???

jsternberg: Sorry, I didn't see you had replied.

If you look at the struct definition, you will see this:

```go
type Client struct {
    url        url.URL
    unixSocket string
    username   string
    password   string
    httpClient *http.Client
    userAgent  string
    precision  string
}
```

If you notice, `url.URL` is not a pointer. So when you do this:

```
u := c.url
```

That creates a copy of the `url.URL` struct. We can then look at [url.URL](https://godoc.org/net/url#URL) in the documentation and we see that almost all of the attributes in there are values. There's one pointer, but we don't modify it so it doesn't matter. That means it is safe for us to modify without causing any kind of race condition.

So my original statement stands. It would be easier to modify the original code to just add the additional path component rather than to do a complex rendering of the url to a string and then reparsing it while checking for an error that would never happen.

genofire: oh sry, u are right

genofire: https://play.golang.org/p/8FHZDXpLnK_c

jsternberg:
@e-dard can you take a look?

jsternberg:
@genofire also,  have you signed the CLA? We need you to sign that before we can merge this.

genofire:
screenshot_20180207-200457

PR #9234, opened 2017-12-15
Add a gometalinter run to the output

jsternberg:
- [ ] Rebased/mergable
- [ ] Tests pass


jsternberg:
I'm going to continue working on this when I have time in the future (so I'm keeping it up), but this doesn't work at the moment since the diff doesn't work properly. The output of the tool isn't consistent and, if line numbers change, the diff will be incorrect too.

PR #9265, opened 2017-12-29
Sequential series file id & series file segmentation

benbjohnson:
## Overview

Previously the `SeriesFile` used the file offset as the identifier. This meant that there was a single jump to get from an offset to a key. This pull request changes it so there is now a sequential identifier used which points to an offset which then points to the key so it is two jumps. 

This change allows us compact the underlying series data and remove tombstones since the series are not attached to a specific offset. This change also allows us to separate the file into multiple segments so that we avoid having to mmap a large fixed space (which causes issues on Windows).

Segments are allocated in increasing size starting at 4MB and doubling until reaching a maximum of 256MB. Segments will be able to be compacted in the future but that is not included in this PR. Only the index is rebuilt during compaction right now.

The `SeriesFile` is broken into two subcomponents now: `SeriesSegment` and `SeriesIndex`. The segments are simply logs of series inserts & tombstones. When a segment is full then a new segment is created and appended to. The index is a two tier (disk & in-memory) index that combines the key-to-id mapping and id-to-offset mapping as well as the set of tombstones since the last rebuild.

- [x] Rebased/mergable
- [x] Tests pass

### DIFF
--- a/tsdb/engine/tsm1/engine_test.go
+++ b/tsdb/engine/tsm1/engine_test.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/influxdata/influxdb/logger"
 	"github.com/influxdata/influxdb/models"
 	"github.com/influxdata/influxdb/pkg/deep"
 	"github.com/influxdata/influxdb/query"
@@ -1690,6 +1691,7 @@ func NewEngine(index string) (*Engine, error) {
 	}
 
 	sfile := tsdb.NewSeriesFile(seriesPath)
+	sfile.Logger = logger.New(os.Stdout)
 	if err = sfile.Open(); err != nil {
 		return nil, err
 	}
--- a/tsdb/engine/tsm1/engine_test.go
+++ b/tsdb/engine/tsm1/engine_test.go
@@ -1685,7 +1685,7 @@ func NewEngine(index string) (*Engine, error) {
 	}
 
 	// Setup series file.
-	seriesPath, err := ioutil.TempDir(dbPath, "series")
+	seriesPath, err := ioutil.TempDir(dbPath, tsdb.SeriesFileDirectory)
 	if err != nil {
 		return nil, err
 	}
--- a/tsdb/engine/tsm1/engine_test.go
+++ b/tsdb/engine/tsm1/engine_test.go
@@ -1684,19 +1684,12 @@ func NewEngine(index string) (*Engine, error) {
 	}
 
 	// Setup series file.
-	f, err := ioutil.TempFile(dbPath, "series")
+	seriesPath, err := ioutil.TempDir(dbPath, "series")
 	if err != nil {
 		return nil, err
 	}
-	f.Close()
-	sfile := tsdb.NewSeriesFile(f.Name())
 
-	// If we're running on a 32-bit system then reduce the SeriesFile size, so we
-	// can address is in memory.
-	if runtime.GOARCH == "386" {
-		sfile.MaxSize = 1 << 27 // 128MB
-	}
-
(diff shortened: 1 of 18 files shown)


tsdb/engine/tsm1/engine_test.go
```diff
@@ -1729,19 +1722,11 @@ type SeriesFile struct {
 
 // NewSeriesFile returns a new instance of SeriesFile with a temporary file path.
 func NewSeriesFile() *SeriesFile {
-	file, err := ioutil.TempFile("", "tsdb-series-file-")
+	dir, err := ioutil.TempDir("", "tsdb-series-file-")
```
e-dard: Do you need to change line `1743` (`SeriesFile.Close)` to use `RemoveAll` now that the series file is a collection of files?

tsdb/series_segment.go
```diff
@@ -0,0 +1,396 @@
...
+		if flag == 0 {
+			break
+		}
+		s.size += uint32(sz)
+	}
+
+	// Open file handler for writing & seek to end of data.
+	if s.file, err = os.OpenFile(s.path, os.O_WRONLY|os.O_CREATE, 0666); err != nil {
+		return err
+	} else if _, err := s.file.Seek(int64(s.size), os.SEEK_SET); err != nil {
```
e-dard: `os.SEEK_SET` is [deprecated](https://golang.org/pkg/os/#pkg-constants), you should use `io.SeekCurrent` instead.

tsdb/series_file.go
```diff
@@ -900,3 +746,5 @@ func hashReader(r io.Reader) ([]byte, error) {
 	}
 	return h.Sum(nil), nil
 }
+
+func hexdump(data []byte) { os.Stderr.Write([]byte(hex.Dump(data))) }
```
e-dard: This isn't called by anything; maybe define it in a test file if it's still useful?

tsdb/series_file.go
```diff
@@ -690,196 +530,202 @@ func (a seriesKeys) Less(i, j int) bool {
...
-			hash, key, val = h, k, v
+			hash, id, offset = elemHash, elemID, elemOffset
 
 			// Update current distance.
 			dist = d
 		}
+
+		if i > capacity {
+			panic("rhh map full")
+		}
```
e-dard: The `hashReader` function further down in this file can be deleted.

tsdb/engine/tsm1/engine_test.go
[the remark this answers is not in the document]
benbjohnson: Fixed in 3900c94.

tsdb/series_file.go
benbjohnson: Fixed in 3900c94.

tsdb/series_file.go
benbjohnson: Removed in 3900c94.

tsdb/series_segment.go
benbjohnson: Fixed in 3900c94.

tsdb/series_segment.go
```diff
@@ -0,0 +1,396 @@
...
+}
+
+// IsValidSeriesSegmentFilename returns true if filename is a 4-character lowercase hexidecimal number.
+func IsValidSeriesSegmentFilename(filename string) bool {
+	return seriesSegmentFilenameRegex.MatchString(filename)
+}
+
+// ParseSeriesSegmentFilename returns the id represented by the hexidecimal filename.
+func ParseSeriesSegmentFilename(filename string) uint16 {
+	i, _ := strconv.ParseUint(filename, 16, 32)
```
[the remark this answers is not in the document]
benbjohnson: Fixed in 3900c94.

Issue #9255, opened 2017-12-21
Blocks are overlapping after full compaction


hpbieker:
### Bug report

__System info:__ 
InfluxDB 1.3.7, Ubuntu 16.4, 128 GB RAM, 16 cores.

__Steps to reproduce:__

Make sure a "full compaction" of the shard has ran.
```
November 3rd 2017, 17:04:13.000	compacted full 4 files into 3 files in 15m46.744294449s engine=tsm1
November 3rd 2017, 17:04:13.000	compacted full group (0) into /media/db_volume/data/XXX/autogen/5269/000001594-000000004.tsm.tmp (#0) engine=tsm1
November 3rd 2017, 17:04:13.000	compacted full group (0) into /media/db_volume/data/XXX/autogen/5269/000001594-000000005.tsm.tmp (#1) engine=tsm1
November 3rd 2017, 17:04:13.000	compacted full group (0) into /media/db_volume/data/XXX/autogen/5269/000001594-000000006.tsm.tmp (#2) engine=tsm1
November 3rd 2017, 16:48:27.000	compacted full group (0) into /media/db_volume/data/XXX/autogen/5269/000001594-000000002.tsm.tmp (#0) engine=tsm1
November 3rd 2017, 16:48:27.000	compacting full group (0) /media/db_volume/data/XXX/autogen/5269/000000886-000000005.tsm (#0) engine=tsm1
November 3rd 2017, 16:48:27.000	compacting full group (0) /media/db_volume/data/XXX/autogen/5269/000001594-000000002.tsm (#2) engine=tsm1
November 3rd 2017, 16:48:27.000	compacted full group (0) into /media/db_volume/data/XXX/autogen/5269/000001594-000000003.tsm.tmp (#1) engine=tsm1
November 3rd 2017, 16:48:27.000	compacting full group (0) /media/db_volume/data/XXX/autogen/5269/000000886-000000006.tsm (#1) engine=tsm1
November 3rd 2017, 16:48:27.000	compacting full group (0) /media/db_volume/data/XXX/autogen/5269/000001594-000000003.tsm (#3) engine=tsm1
November 3rd 2017, 16:41:35.000	compacting full group (0) /media/db_volume/data/XXX/autogen/5269/000001591-000000003.tsm (#4) engine=tsm1
November 3rd 2017, 16:41:35.000	compacting full group (0) /media/db_volume/data/XXX/autogen/5269/000001594-000000001.tsm (#6) engine=tsm1
[...]

PR #9267, opened 2018-01-02
Sort blocks by time when compacting

hpbieker:
### Overview
This PR solves https://github.com/influxdata/influxdb/issues/9255 where blocks where left unsorted after a compaction. So given a key and two blocks, you where not garanteed that minTime[0] <= maxTime[0] < minTime[1] <= maxTime[1].

This also caused some slow queries and high memory consumption at runtime because of merging https://github.com/influxdata/influxdb/issues/9241.

This is my first PR in InfluxDB and I am not that familiar with the source code, so please review the change before merging.

- [X] Rebased/mergable
- [X] Tests pass
- [X] CHANGELOG.md updated
  #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 - [#9226](https://github.com/influxdata/influxdb/issues/9226): Allow lone boolean literals in a condition expression.
 - [#9235](https://github.com/influxdata/influxdb/pull/9235): Improve performance when writes exceed `max-values-per-tag` or `max-series`.
 - [#9216](https://github.com/influxdata/influxdb/issues/9216): Prevent a panic when a query simultaneously finishes and is killed at the same time.
+- [#9255](https://github.com/influxdata/influxdb/issues/9255): Fix missing sorting of blocks by time when compacting.
 
 ## v1.4.3 [unreleased]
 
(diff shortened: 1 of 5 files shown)


hpbieker:
I see that there is a build failure, but it looks like other PR are also currently failing.

hpbieker:
I have tested your fix, and it works very well. I updated the branch with your suggestions and merged master into the branch.

jwilder:
Thanks @hpbieker!

hpbieker:
Thank you @jwilder . Unfortuantly, it looks like I was not able to sync the changes in the  compact.gen.go.tmpl and the  compact.gen.go file. As you can see, the sort function is only in the  compact.gen.go.tmpl file, and the modified if check is only in the  compact.gen.go file.

However, the test passes without the sort for some reason!?

jwilder:
@hpbieker Ah.  I didn't catch that.  I'll fix that up and submit a PR.  To rebuild the `.gen.go` files, you would run `go generate ./tsdb/engine/tsm1`.  You would need to install https://github.com/benbjohnson/tmpl as well which processes the `.tmpl` files.

jwilder:
@hpbieker Looks like you might have already submitted a fix #9278.

PR #9220, opened 2017-12-11
Prevent a panic when a query simultaneously finishes and is killed at the same time

jsternberg:
There is a strange race condition where a query can be killed and finish
at approximately the same time. If this happens, the query gets
retrieved by the killing task, the query gets closed by the normal
processing thread, and then the killing task attempts to kill the query
afterwards. Since the close doesn't mark the query as already killed
(since it's not killed, just merely unused), the killing thread attempts
to close the channel again.

Mark the query as killed whenever it is closed to prevent a double close
from happening. This should never cause the status to be erroneously
reported since the query status is removed from the query table within
the same lock scope.

Fixes #9216.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

PR #9272, opened 2018-01-02
Prevent a panic when a query simultaneously finishes and is killed at the same time

jsternberg:
There is a strange race condition where a query can be killed and finish
at approximately the same time. If this happens, the query gets
retrieved by the killing task, the query gets closed by the normal
processing thread, and then the killing task attempts to kill the query
afterwards. Since the close doesn't mark the query as already killed
(since it's not killed, just merely unused), the killing thread attempts
to close the channel again.

Mark the query as killed whenever it is closed to prevent a double close
from happening. This should never cause the status to be erroneously
reported since the query status is removed from the query table within
the same lock scope.

Backport of #9220.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated


leafonsword:
Is 1.4.3 merge this feature?

PR #9278, opened 2018-01-04
Fixed issue where compacting did not sort when blocks are unsorted and overlapping

hpbieker:
This PR fixes some issues with https://github.com/influxdata/influxdb/pull/9267.

In the case where you have both overlapping and and unsorted blocks, the previous code did not work because sort was missing. Furthermore, the template file compact.gen.go.tmpl and compact.gen.go was out of sync.

@jwilder, please review this as this is a follow-up on https://github.com/influxdata/influxdb/pull/9267.

- [X] Rebased/mergable
- [X] Tests pass
- [X] CHANGELOG.md updated

### DIFF
--- a/tsdb/engine/tsm1/compact.gen.go
+++ b/tsdb/engine/tsm1/compact.gen.go
@@ -424,7 +424,7 @@ func (k *tsmKeyIterator) mergeUnsigned() {
 		// we need to dedup as there may be duplicate points now
 		for i := 1; !dedup && i < len(k.blocks); i++ {
 			dedup = k.blocks[i].partiallyRead() ||
-				k.blocks[i].minTime <= k.blocks[i-1].maxTime ||
+				k.blocks[i].overlapsTimeRange(k.blocks[i-1].minTime, k.blocks[i-1].maxTime) ||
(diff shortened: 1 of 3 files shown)


hpbieker:
@jwilder, thank you. The file is updated now.

hpbieker:
BTW -- will the sorting somehow cause overwritten values to be left unmodified?

Lets say you have the following TSM files:
TSM1 with one blocK: T = [2], V = [1]
TSM2 with one blocK: T = [1, 2], V = [2, 3]

After the merge, will you will have T = [1, 2], but will V = [2, 3] or [1, 3]?

jwilder:
@hpbieker Values with the same timestamp in later TSM files should overwrite prior ones after merging.

hpbieker:
Thanks again @jwilder, I just verified that now. I see that Less(), which is used by sort.Stable(), returns true only if:
* The key is less, or
* The **whole** block i is before block j

https://github.com/influxdata/influxdb/blob/176a7e1e98e2d11d2abe7cbe7db565ea7e30b817/tsdb/engine/tsm1/compact.go#L1270-L1276

I was worried that it only checked the minTime.

PR #9285, opened 2018-01-05
WIP: Close series file on database deletion.

benbjohnson:
- [x] Rebased/mergable
- [x] Tests pass

### DIFF
--- a/tsdb/store.go
+++ b/tsdb/store.go
@@ -566,6 +566,7 @@ func (s *Store) DeleteDatabase(name string) error {
 		// no files locally, so nothing to do
 		return nil
 	}
+	sfile := s.sfiles[name]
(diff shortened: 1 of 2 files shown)


tsdb/store.go
```diff
@@ -566,6 +566,7 @@ func (s *Store) DeleteDatabase(name string) error {
 		// no files locally, so nothing to do
 		return nil
 	}
+	sfile := s.sfiles[name]
```
e-dard: This is racy with another goroutine creating a new database. You should use `seriesFile(database string)`.

jwilder: L563 takes an `RLock` which is what `seriesFile()` does I believe.  This looks ok to me.

e-dard: Missed that `RLock`.

PR #9288, opened 2018-01-05
Fix series file compaction stall.

benbjohnson:
## Overview

The series file compaction previously did not snapshot the max
offset before compacting and would keep compacting until it reached
the end of segment file. This caused more entries than expected into
the RHH map and this map gets exponentially slower as it gets close
to full.

- [x] Rebased/mergable
- [x] Tests pass


benbjohnson:
The series file needs some ref counting added to fix CI but I'm going to go ahead and merge because it's better than what's in `master` right now. Will fix ref counting in a separate PR.

PR #9295, opened 2018-01-09
Partition series file

benbjohnson:
## Overview

This pull request splits the `SeriesFile` into multiple partitions each with independent locking. There are also some memory optimizations around bulk generation of series keys & removing allocations during TSI index compaction.

- [x] Rebased/mergable
- [ ] Tests pass

### DIFF
--- a/pkg/binaryutil/binaryutil.go
+++ b/pkg/binaryutil/binaryutil.go
@@ -1,0 +1,22 @@
+package binaryutil
+
+// VarintSize returns the number of bytes to varint encode x.
+// This code is copied from encoding/binary.PutVarint() with the buffer removed.
+func VarintSize(x int64) int {
+	ux := uint64(x) << 1
+	if x < 0 {
+		ux = ^ux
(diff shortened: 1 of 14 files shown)


pkg/binaryutil/binaryutil.go
```diff
@@ -0,0 +1,22 @@
+package binaryutil
```
e-dard: Maybe this could just be `binary`?

tsdb/series_partition.go
```diff
@@ -0,0 +1,665 @@
...
+	}
+
+	// Mark tombstone in memory.
+	p.index.Delete(id)
+
+	return nil
+}
+
+// IsDeleted returns true if the ID has been deleted before.
+func (p *SeriesPartition) IsDeleted(id uint64) bool {
```
e-dard: Do series that have never been added to this partition return `true`? Does this cover series that have been deleted and also series that have never existed?

PR #9297, opened 2018-01-10
client v2 support http proxy

coooold:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

- [ ] Provide example syntax

### DIFF
--- a/client/v2/client.go
+++ b/client/v2/client.go
@@ -44,6 +44,9 @@ type HTTPConfig struct {
 	// TLSConfig allows the user to set their own TLS config for the HTTP
 	// Client. If set, this option overrides InsecureSkipVerify.
 	TLSConfig *tls.Config
+
+	// HttpProxy can be set to create http.Transport. "" for none-proxy
+	HttpProxy string
 }
 
 // BatchPointsConfig is the config data needed to create an instance of the BatchPoints struct.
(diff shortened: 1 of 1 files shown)


client/v2/client.go
```diff
@@ -44,6 +44,9 @@ type HTTPConfig struct {
 	// TLSConfig allows the user to set their own TLS config for the HTTP
 	// Client. If set, this option overrides InsecureSkipVerify.
 	TLSConfig *tls.Config
+
+	// HttpProxy can be set to create http.Transport. "" for none-proxy
```
jsternberg: Can you modify this to be the same function signature as `Proxy`? I think it's `func (*http.Request) (*url.URL, error)`. That way, a client can use the `http.ProxyURL(*url.URL)` function if they want to and we can use `http.ProxyFromEnvironment` in the CLI program.

Can you also modify the name to be `Proxy` and change the documentation comment to be:

```
Proxy specifies a function to return a proxy for a given
Request. If the function returns a non-nil error, the
request is aborted with the provided error.
```

It's basically the same text from the `Proxy` field documentation on the Transport struct.

jsternberg:
I'm closing this in favor of #9429.

Issue #9299, opened 2018-01-10
TSI: revert to 1.3 DELETE semantics.


e-dard:
During the development of TSI we decided we're going to stick with the current delete/drop semantics.

The InfluxDB documentation states that `DELETE` never removes data from the index, however the database has been behaving as follows, since version 1.3.

 - `DELETE` doesn't remove series data from the index, if there exist other undeleted data for the same series key.
 - `DELETE` will remove the series from the index if the last data for that series are removed with the `DELETE` command.
 - `DROP` always removes data from the index.

This poses some challenges for the different indexes. 

The `inmem` index is shared over all shards. Therefore a shard cannot remove a series from the index without knowing first if any other shards have data associated with that series. The current way this is handled is by tracking which shards own which series.

The `tsi1` index is not shared between shards. Each shard has a separate index. Therefore it's a simple matter of removing that series from the shard's index when all remnants of data associated with it are removed.

#### Series File

With recent changes, a database-wide series file has been introduced, to help performance across all indexes. Each index now will not store a reference to series keys, but will instead store `series ids`, which themselves reference series keys in a global (database-wide) series file.

This creates some additional challenges for both the `inmem` index and the `tsi1` index.

 1) Since all shards share the series file, marking a series as deleted in the series file will affect all shards. Therefore, `tsi1` indexes can no longer be considered self contained with respect to `DELETE/DROP` semantics.

 2) The `inmem` index should no longer be determining whether a series is deleted or not. It should be checking against the series file.

 3) The series file will help support higher cardinality databases. Tracking which shards own which series by mapping shard values to each series value will not scale well for many millions of series.

#### Proposal

The proposal is as follows:

 1) Introduce a compressed bitset, in this case probably using [roaring bitmaps](http://roaringbitmap.org/) into each index, to track series creation and deletion. Each bitmap will store the series ID associated with the series in the series file. It's important that in the case of the `inmem` index, each shard has its own bitmap, and they're not shared.

2) Revert the previous delete/drop changes introduced in https://github.com/influxdata/influxdb/pull/9150.

3) For the `inmem` index, during series deletion, determine if any other shards' bitsets contain the series that is proposed to be removed from the shard's index. If no other shards contain the series then it can be removed from both the global `inmem` index, and also from the series file.

[...]

PR #9301, opened 2018-01-10
Track series IDs at a per-shard level

e-dard:
- [ ] Rebased/mergable
- [ ] Tests pass

This PR is the first step in #9299.

#### Changes

  - Adds a compressed bitset to the `ShardIndex`, which allows us to maintain a per-shard bitset for the `inmem` index. The bitset is injected into the `inmem` series creation method calls, so each shard's bitset is updated correctly.
 - Alters the existing behaviour of the `tsi1` bitset, which was introduced in https://github.com/influxdata/influxdb/pull/9275. Previously, this bitset was only ever added to, with the idea being that if a series was re-added to the index it would have a different series id in the series file. However, this assumption is incorrect if you consider removing a series from only one shard and then re-adding it. In that case the same series id would be in the series file. So, we fix this problem by ensuring that series ids are removed from this bitset when they're deleted from the `tsi1` index.
 - Leverage this `tsi1` bitset to use as the tsi bitset. Since it's a per-partition bitset, a method is added to the `tsi1.Index` type that creates a union of these bitsets when needed.


tsdb/index/tsi1/index.go
```diff
@@ -806,7 +818,16 @@ func (i *Index) SetFieldName(measurement []byte, name string) {}
...
+// UnassignShard removes the provided series key from the index. The naming of
+// this method stems from a legacy index logic that used to track which shards
+// owned which series.
+func (i *Index) UnassignShard(k string, id uint64, ts int64) error {
+
+	// TODO(edd): temporarily munging series id and shard id into same value,
+	// to test prototype without having to change Index API.
+	// seriesID := id >> 32
+	// i.seriesIDSet.Remove(seriesID)
+
```
benbjohnson: Should this commented block be removed?

tsdb/engine/tsm1/engine.go
```diff
@@ -1443,7 +1444,18 @@ func (e *Engine) deleteSeriesRange(seriesKeys [][]byte, min, max int64, removeIn
 			}
 
 			// Remove the series from the index.
-			if err := e.index.UnassignShard(string(k), e.id, ts); err != nil {
+			// TODO(edd): if a DROP MEASUREMENT, or even a DROP SERIES had been
```
jwilder: This comment isn't correct.  You can't remove from series file/index until after we've verified that no concurrent writes were in place (e.g. no values added to cache while this delete is running).

tsdb/index/inmem/inmem.go
```diff
@@ -211,6 +210,9 @@ func (i *Index) CreateSeriesIfNotExists(shardID uint64, key, name []byte, tags m
 	// Add the series to the series sketch.
 	i.seriesSketch.Add(key)
 
+	// This series needs to be added to the bitset tracking undeleted series IDs.
+	seriesSet.Add(seriesID)
```
jwilder: I think this might make the need for the `seriesSketch` on L211 unnecessary.  Might want to look at that another time.

PR #9316, opened 2018-01-15
Cleanup Shard assignment/unassignment 

e-dard:
- [x] Rebased/mergable
- [ ] Tests pass

This PR relies on #9315. I will rebase this, once that is merged.

This PR cleans up the `inmem` index by removing all methods and data structures associated with shard assignment/unassignment.

Now that each shard-local index is maintaining a bitset of series ids, tracking the series present in the local shard's tsm engine, there is no need to track shards in the `inmem` index.

It also removes the `Shard.CloseFast` method, as all shard closes should be "fast".


e-dard:
This PR now relies on #9321 being merged first.

PR #9319, opened 2018-01-15
Remove series when shard rolls over

e-dard:
- [ ] Rebased/mergable
- [ ] Tests pass

This PR builds on #9315  and #9316; they should be merged first.

The only commit directly relevant to this PR is https://github.com/influxdata/influxdb/commit/2bb0b79b0da6c8e4411b49b95711eb124d20462c.

Series should only be removed from the series file when they're no
longer present in any shard. This commit ensures that during a shard
rollover, the series local to the shard are checked against all other
series in the database.

Series that are no longer present in any other shards' bitsets, are then
marked as deleted in the series file.

### DIFF
--- a/tsdb/store.go
+++ b/tsdb/store.go
@@ -534,28 +534,67 @@ func (s *Store) DeleteShard(shardID uint64) error {
 		return nil
 	}
 
-	// Remove the shard from the database indexes before closing the shard.
-	// Closing the shard will do this as well, but it will unload it while
-	// the shard is locked which can block stats collection and other calls.
-	sh.UnloadIndex()
-
-	if err := sh.Close(); err != nil {
-		return err
-	}
-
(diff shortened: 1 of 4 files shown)


tsdb/store.go
```diff
@@ -534,28 +534,67 @@ func (s *Store) DeleteShard(shardID uint64) error {
...
+	s.walkShards(shards, func(sh *Shard) error {
+		index, err := sh.Index()
+		if err != nil {
+			return err
+		}
+
+		if i, ok := index.(interface {
+			SeriesIDSet() *SeriesIDSet
+		}); ok {
+			ss.Diff(i.SeriesIDSet())
```
jwilder: Does this return a new bitset or mutate the existing one?

e-dard: It mutates the existing one, making it slightly different to the `AndNot` method in the series set.

PR #9335, opened 2018-01-17
Prevent race condition caused by WaitGroup re-use

joelegasse:
This has been an elusive issue, but this should fix the data race seen occasionally during testing.

- [x] Rebased/mergable
- [ ] Tests pass
- [x] CHANGELOG.md updated


tsdb/engine/tsm1/engine.go
```diff
@@ -136,12 +136,12 @@ type Engine struct {
 	// decrease 'levelWorkers', and when it decreases to zero, level compactions will be started
 	// back up again.
 
-	wg           sync.WaitGroup // waitgroup for active level compaction goroutines
-	done         chan struct{}  // channel to signal level compactions to stop
-	levelWorkers int            // Number of "workers" that expect compactions to be in a disabled state
+	wg           *sync.WaitGroup // waitgroup for active level compaction goroutines
```
mark-rushakoff: I had to read the code pretty closely to understand why `wg` is kept as a field even though it's passed through some methods as arguments. The comment here or in `enableLevelCompactions` could probably explain it better.

PR #9341, opened 2018-01-18
TSI/SeriesFile Compaction Disabling

jwilder:
This adds the capability to disable TSI and SeriesFile compactions.  Disabling them prevents new compactions from starting, but does not end any in-progress.  This fixes deleting w/ TSI where a TSI/SeriesFile compaction would run while deleting causing deleted data to re-appear.

This also disables the compactions (like TSM) when deleting series.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/tsdb/index/tsi1/partition.go
+++ b/tsdb/index/tsi1/partition.go
@@ -76,7 +76,7 @@ type Partition struct {
 	MaxLogFileSize int64
 
 	// Frequency of compaction checks.
-	compactionsDisabled       bool
+	compactionsDisabled       int
 	compactionMonitorInterval time.Duration
 
 	logger *zap.Logger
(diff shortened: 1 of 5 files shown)


tsdb/index/tsi1/partition.go
```diff
@@ -761,11 +761,32 @@ func (i *Partition) Compact() {
 	i.compact()
 }
 
+func (i *Partition) DisableCompactions() {
```
e-dard: Just a thought—by using a counting semaphore instead of a binary one, it means one has to `EnableCompactions` as many times as `DisableCompactions`.

I realise that we may have an invariant somewhere that says we won't call `DisableCompactions` multiple times without calling `EnableCompactions`, but could we maybe just use a binary semaphore, and have it so that a subsequent call to `DisableCompactions` when they're already disabled, is a no-op?

tsdb/series_partition.go
```diff
@@ -362,6 +363,26 @@ func (p *SeriesPartition) SeriesCount() uint64 {
 	return n
 }
 
+func (p *SeriesPartition) DisableCompactions() {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	p.compactionsDisabled++
```
e-dard: Same question here about semaphore choice.

tsdb/index/tsi1/partition.go
jwilder: The problem with the binary one is that if we disable here, we don't want another disable/enable pair that completes faster than the first, to start compactions again. That would cause compactions to run during the the first delete.

e-dard: @jwilder Ah I see. OK, that makes sense.

PR #9349, opened 2018-01-21
Cleanup of codebase using static analysis tools

e-dard:
- [ ] Rebased/mergable
- [ ] Tests pass

This PR cleans up the code-base using static analysis tools available in the [go-tools](https://github.com/dominikh/go-tools) repo. Specifically, the following are used:

  - `unused`: identifies unused identifiers.
  - `staticcheck`: identifies bugs and inefficient code.
  - `simplify`: identifies code that could be rewritten in a simpler or more idiomatic way.

Using the output of these tools, I went through each report line-item and updated the code accordingly. 

I found a number of bugs, typically situations where we're deferring before checking an error or failing to check error conditions.

The codebase should now pass all the above tools. I plan to add a bot to run these tools on future PRs.


e-dard:
@jsternberg could you take a look at 21f0c64? I had to rewrite it a bit to be more idiomatic. Previously the code contained a loop that was unconditionally terminating on the first iteration. I realise it was probably written like that for brevity, but I think it would be preferable to have an entire codebase that passes the static analysis tools.

e-dard:
@jwilder could you look at https://github.com/influxdata/influxdb/pull/9349/commits/821b784fa00446dfa150904b8903b68ab0333640? I think that we were just using `filepath.HasPrefix` as a fancy way of doing a string prefix check. The static tooling shout about the use of `filepath.HasPrefix` because it's a deprecated method in the Go standard library.

cmd/influxd/restore/restore.go
```diff
@@ -416,7 +419,7 @@ func (cmd *Command) uploadShardsPortable() error {
 						targetDB = file.Database
 					}
 
-					if err := cmd.client.UploadShard(file.ShardID, cmd.shardIDMap[file.ShardID], cmd.destinationDatabase, cmd.restoreRetention, tr); err != nil {
+					if err := cmd.client.UploadShard(file.ShardID, cmd.shardIDMap[file.ShardID], targetDB, cmd.restoreRetention, tr); err != nil {
```
e-dard: @aanthony1243 I think this was buggy before, right? We were never actually using `targetDB`.

pkg/bytesutil/bytesutil.go
```diff
@@ -155,7 +152,7 @@ func CloneSlice(a [][]byte) [][]byte {
 // Pack converts a sparse array to a dense one.  It removes sections of a containing
 // runs of val of length width.  The returned value is a subslice of a.
 func Pack(a []byte, width int, val byte) []byte {
-	var i, j, iStart, jStart, end int
+	var i, j, jStart, end int
```
e-dard: @jwilder  `iStart` was completely unused. Was it supposed to be used? i.e., was/is this code buggy?

cmd/influxd/restore/restore.go
aanthony1243: i don't recall any bugs at this level.  any bugs occurring were a step further back in the process.  I will have to take a closer look at this change, but if the tests pass it is probably ok.

aanthony1243: @e-dard  took a look before sutting down my laptop.  you are right, this is a correct fix.  For better or worse, the client also checks the passed value for "" before it's submitted, so it shouldn't have been a bug.  I may take a closer look later to be sure that the checks are identical but I believe that they are.

tsdb/series_set.go
```diff
@@ -100,7 +100,7 @@ func (s *SeriesIDSet) Equals(other *SeriesIDSet) bool {
 	s.RLock()
 	defer s.RUnlock()
 	other.RLock()
-	other.RUnlock()
+	defer other.RUnlock()
```
stuartcarnie: good catch!

PR #9350, opened 2018-01-22
escape non-printing characters for columnar output on cli

phemmer:
This causes the influx CLI utility to escape non-printable characters when using column formatting.
My use case is that I have fields containing short binary data (binary encoded UUIDs), and I frequently run commands like `select * from foo`. I could explicitly list the columns I want to omit the binary ones, but this is cumbersome. Also seeing the escaped binary is sometimes useful for correlating data across rows.

- [X] Rebased/mergable
- [X] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/cmd/influx/cli/cli.go
+++ b/cmd/influx/cli/cli.go
@@ -975,7 +975,7 @@ func (c *CommandLine) formatResults(result client.Result, separator string, supp
 			}
 
 			for _, vv := range v {
-				values = append(values, interfaceToString(vv))
+				values = append(values, interfaceToString(vv, c.Format == "column"))
 			}
 			rows = append(rows, strings.Join(values, separator))
 		}
@@ -983,7 +983,7 @@ func (c *CommandLine) formatResults(result client.Result, separator string, supp
 	return rows
 }
 
-func interfaceToString(v interface{}) string {
+func interfaceToString(v interface{}, escapeString bool) string {
 	switch t := v.(type) {
 	case nil:
 		return ""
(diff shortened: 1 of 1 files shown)


cmd/influx/cli/cli.go
```diff
@@ -994,7 +994,11 @@ func interfaceToString(v interface{}) string {
 	case float32, float64:
 		return fmt.Sprintf("%v", t)
 	default:
-		return fmt.Sprintf("%v", t)
+		if !escapeString {
+			return fmt.Sprintf("%v", t)
+		}
+		v := fmt.Sprintf("%q", t)
```
phemmer: Aside from binary data getting escaped, this will have an effect on 2 non-binary characters, `\` and `"`, which will get translated to `\\` and `\"` respectively.
The whole reason for doing this escaping on columnar output only was to only affect usage by humans, so this seems like a minor break. But open to other options (such as removing the escaping on `"`).

phemmer:
Aside from binary data getting escaped, this will have an effect on 2 non-binary characters, `\` and `"`, which will get translated to `\\` and `\"` respectively.
The whole reason for doing this escaping on columnar output only was to only affect usage by humans, so this seems like a minor break. But I'm open to other options (such as some sort of flag to toggle the behavior, or removing the escaping on `"`).

jsternberg:
This seems like an appropriate method if we wanted to do that: https://godoc.org/unicode#IsPrint

phemmer:
I personally like the compact output that `%q` provides. Especially if it's only a single character, such as `\n` or `\t` which is present. However that's just a personal preference. base64 / hex would work fine. The one technical detail that springs to mind is that doing `IsPrint` on every character might be expensive. But I dunno that it's any more expensive than whatever `%q` is doing under the hood.

PR #9352, opened 2018-01-22
Fix imports of multiple databases in one export file

### LINKED ISSUE #9336: influx -import may import to wrong database
### Bug report

__System info:__ CentOS 7. InfluxDB 1.4.2

__Steps to reproduce:__

1. Create two databases. Insert a point into each

```
CREATE DATABASE db0
USE db0
INSERT my_measurement,dest=db0 value=0
CREATE DATABASE db1
USE db1
INSERT my_measurement,dest=db1 value=1
```

2. Export with influx_inspect

```
influx_inspect export -out [some path] (datadir, waldir, etc)
```

This gives

```
# INFLUXDB EXPORT: 1677-09-21T00:12:43Z - 2262-04-11T23:47:16Z
# DDL
CREATE DATABASE db0 WITH NAME autogen
CREATE DATABASE db1 WITH NAME autogen
# DML
# CONTEXT-DATABASE:db0
# CONTEXT-RETENTION-POLICY:autogen
# writing wal data
my_measurement,dest=db0 value=0 1516242288102872319
# CONTEXT-DATABASE:db1
# CONTEXT-RETENTION-POLICY:autogen
# writing wal data
my_measurement,dest=db1 value=1 1516242314672449241
```

3. Drop databases

```
DROP DATABASE db0
DROP DATABASE db1
```

4. Import with `influx -import -path ...`

```
2018/01/18 02:33:47 Processed 2 commands
2018/01/18 02:33:47 Processed 2 inserts
2018/01/18 02:33:47 Failed 0 inserts
```

__Expected behavior:__ Points have been fully restored from the export files

__Actual behavior:__ Every line protocol has been inserted into the second database db1

```
USE db0
SELECT * FROM my_measurement
(nothing)

USE db1
SELECT * FROM my_measurement

name: my_measurement
time                dest value
----                ---- -----
1516242288102872319 db0  0
1516242314672449241 db1  1
```
__Additional info:__

I believe this is an issue in the way batching was [implemented in](https://github.com/influxdata/influxdb/blob/master/importer/v8/importer.go) `importer/v8/importer.go`. The lines from the export file like

```
# CONTEXT-DATABASE:some_db_name
# CONTEXT-RETENTION-POLICY:some_retention_policy
```

Are parsed and stored in `i.database` and `i.retentionPolicy` in the `Importer` struct, but then the line protocol commands are batched into a hardcoded 5000 commands. When `batchWrite` runs, it uses the last parsed values of `i.database` and `i.retentionPolicy` which are not necessarily correct.

I originally thought I might have run into same issues brought up in #7516 but I don't think that is the case. First explanation offered in that issue was importing points whose timestamps were already older than the retention policy. That's not the case here. Second example given later in the thread was a confusion over what retention policy is used when running SELECT. Hopefully I have not made a mistake there, but it is always possible!

As far as a fix, is it worth trying to come up with a patch for `influx -import` or is there a plan to deprecate that command? I just started working with Influx a few weeks back and I'm still confused over the vision for the command line tooling.  [This community thread](https://community.influxdata.com/t/influxdb-cli-tools-limitations-is-there-a-roadmap-for-tools/504) outlined a lot of the same confusion I was experiencing. Will the export format from influx_inspect be supported going forward?

Thanks for the time!

wwilfinger:
Fixes #9336

I'm not happy with some of the names here. I also want to add a test that hits the 5000 batchSize to verify everything works there. Will probably look at `cmd/influx_inspect/export/export_test.go` to see how to generate larger test dbs.

batchWrite was using the last database and retention policy read from
the input file. Because batchWrite was called only every batchSize lines
or at EOF, databases with fewer than batchWrite points could be imported
into the incorrect database.

This change forces a flush with batchWrite whenever processDML reads a
change in database or retention policy.

- [x] Rebased/mergable
- [ ] Tests pass
- [x] CHANGELOG.md updated

### DIFF
--- a/importer/v8/importer.go
+++ b/importer/v8/importer.go
@@ -42,6 +42,7 @@ type Importer struct {
 	failedInserts         int
 	totalCommands         int
 	throttlePointsWritten int
+	startTime             time.Time
 	lastWrite             time.Time
 	throttle              *time.Ticker
 
@@ -167,7 +168,7 @@ func (i *Importer) processDDL(scanner *bufio.Reader) error {
 }
 
 func (i *Importer) processDML(scanner *bufio.Reader) error {
-	start := time.Now()
+	i.startTime = time.Now()
 	for {
 		line, err := scanner.ReadString(byte('\n'))
 		if err != nil && err != io.EOF {
@@ -178,9 +179,11 @@ func (i *Importer) processDML(scanner *bufio.Reader) error {
 			return nil
 		}
 		if strings.HasPrefix(line, "# CONTEXT-DATABASE:") {
+			i.batchWrite()
 			i.database = strings.TrimSpace(strings.Split(line, ":")[1])
 		}
 		if strings.HasPrefix(line, "# CONTEXT-RETENTION-POLICY:") {
+			i.batchWrite()
 			i.retentionPolicy = strings.TrimSpace(strings.Split(line, ":")[1])
 		}
 		if strings.HasPrefix(line, "#") {
@@ -190,7 +193,7 @@ func (i *Importer) processDML(scanner *bufio.Reader) error {
 		if strings.TrimSpace(line) == "" {
 			continue
 		}
-		i.batchAccumulator(line, start)
+		i.batchAccumulator(line)
 	}
 }
 
@@ -210,18 +213,10 @@ func (i *Importer) queryExecutor(command string) {
 	i.execute(command)
 }
 
-func (i *Importer) batchAccumulator(line string, start time.Time) {
+func (i *Importer) batchAccumulator(line string) {
 	i.batch = append(i.batch, line)
 	if len(i.batch) == batchSize {
 		i.batchWrite()
-		i.batch = i.batch[:0]
-		// Give some status feedback every 100000 lines processed
-		processed := i.totalInserts + i.failedInserts
-		if processed%100000 == 0 {
-			since := time.Since(start)
-			pps := float64(processed) / since.Seconds()
-			i.stdoutLogger.Printf("Processed %d lines.  Time elapsed: %s.  Points per second (PPS): %d", processed, since.String(), int64(pps))
-		}
 	}
 }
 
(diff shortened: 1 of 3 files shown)


importer/v8/importer.go
```diff
@@ -178,10 +181,18 @@ func (i *Importer) processDML(scanner *bufio.Reader) error {
 			return nil
 		}
 		if strings.HasPrefix(line, "# CONTEXT-DATABASE:") {
-			i.database = strings.TrimSpace(strings.Split(line, ":")[1])
+			i.dmlDatabase = strings.TrimSpace(strings.Split(line, ":")[1])
```
jsternberg: Is there a reason why you don't just call `i.batchWrite()` here before changing the database? It seems like there would be less of a code change if the batch were just flushed whenever the context database or retention policy is changed. It looks like `batchWrite()` is pretty safe to call multiple times. The only caveat is that, for some reason, this section of code:

```go
i.batch = i.batch[:0]
```

This is in `batchAccumulator` rather than inside of `batchWrite`. If you move that piece of code to the end of `batchWrite` and remove it from `batchAccumulator`, you can just call `batchWrite` whenever you want to flush the writes.

wwilfinger: You're right, thanks. That was much easier. I rebased and pushed.

Not sure how to write a test with more than 5000 points without pasting in a gigantic string of json. Open to suggestions.

jsternberg:
I resolved the changelog conflict and pushed an update. As soon as it passes tests, I'm going to merge this.

Thanks for the help! We should have this change included in the 1.5 release.

jsternberg:
Hi @wwilfinger, can you sign the link to the CLA that's [here](https://influxdata.com/community/cla/)? Thanks.

jsternberg:
@wwilfinger ping. Can you sign the CLA otherwise we're going to have to revert this change and redo it.

Thanks.

wwilfinger:
@jsternberg My employer's legal team is taking their sweet time to get back to me if I can sign the CLA. Go ahead and revert. I apologize for the waste of time.

PR #9369, opened 2018-01-29
Add recent server logs to `/pprof/all` endpoint

e-dard:
- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

This PR adds recent server logs to the `/debug/pprof/all` profiling endpoint. This has two advantages:

  1) It makes it easier to investigate user issues because the logs can often contain useful information.
  2) The logs are taken as the last step in the profile gathering stage, which means they correlate well to the time period the profiles were captured over. Sometimes we're sent profiling data and logs that are from different time periods, making it hard to understand what's going on.

A circular buffer is used to maintain a fixed size of logging data at all times. By default, `2MB` of logs will be made available in the profile archive. Calls to the profiling endpoint copy this buffer and write it out to an `influxdb.log` file.

In order to ensure we capture all logs within this buffer, `os.Stderr` and `os.Stdout` are swapped throughout the program for `logger.Stderr` and `logger.Stdout`, which are actually multi-writers, teeing all data to the circular buffer and the respective original locations.

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -2,12 +2,36 @@ package logger
 
 import (
 	"io"
+	"os"
 	"time"
 
+	"github.com/influxdata/influxdb/pkg/bytesutil"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
 
+// LoggingBuffer maintains a globally available buffer of the most recent logs.
+// It is injected into the server within the influxd main package, and is
+// currently accessed by the custom profiling archive generated using the
+// /debug/pprof/all URL.
+var LoggingBuffer *bytesutil.CircularBuffer
+
+// Stderr should be used whenever any package is logging to os.Stderr. This
+// package ensures that any writes to this writer are duplicated to the logging
+// buffer defined above.
+var Stderr io.Writer
+
+// Stdout is similar to Stderr.
+var Stdout io.Writer
+
+func init() {
+	// Initialize LoggingBuffer with 2MB size.
+	LoggingBuffer = bytesutil.NewCircularBuffer(1 << 21)
+	Stderr = io.MultiWriter(LoggingBuffer, os.Stderr)
+	Stdout = io.MultiWriter(LoggingBuffer, os.Stdout)
+}
+
+// New initializes a new Logger.
 func New(w io.Writer) *zap.Logger {
 	config := zap.NewProductionEncoderConfig()
 	config.EncodeTime = func(ts time.Time, encoder zapcore.PrimitiveArrayEncoder) {
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -1,8 +1,10 @@
 package logger
 
 import (
+	"fmt"
 	"io"
 	"os"
+	"strconv"
 	"time"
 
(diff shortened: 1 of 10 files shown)


e-dard:
N.B, I might be adding a configuration setting to turn this off, depending on the outcome of some internal discussions.

logger/logger.go
```diff
@@ -1,13 +1,49 @@
...
+var Stderr io.Writer
+
+// Stdout is similar to Stderr.
+var Stdout io.Writer
+
+func init() {
+	size := 1 << 21
+	env := os.Getenv("INFLUXDB_PRIVATE_LOG_BUFFER_SIZE")
+	if env != "" {
+		if n, err := strconv.ParseInt(env, 10, 64); err != nil {
```
jwilder: Maybe handle `0` size to disable?

pkg/bytesutil/bytesutil.go
```diff
@@ -4,8 +4,82 @@ import (
...
+
+	if len(p)+b.end <= b.n { // p fits into existing capacity.
+		copy(b.buf[b.end:], p)
+		b.end += len(p)
+		return len(p), nil
+	}
+
+	// Move buffer down such that `end` is `len(p)` further down the buffer.
+	remain := b.n - b.end
+	copy(b.buf, b.buf[len(p)-remain:])
```
jwilder: I'm concerned this is going to be expensive since once the buffer is full, we'll always be shifting ~2MB of memory around under a global write lock when something is logged.  Did you look at using a circular buffer with start and end indexes into `p.buf`.  See [Circular Buffer](https://en.wikipedia.org/wiki/Circular_buffer)

e-dard: @jwilder I benchmarked the worst case—writing to a full buffer—and it came in at `0.41 ms` on my laptop, which is worse than I expected. I'll re-implement it as a full ring (I was prototyping this for internal discussion to be honest anyway).

jsternberg:
I have concerns with this idea. Consul has a similar thing where it has an endpoint that emits logs through an HTTP endpoint. Since consul is configured to communicate with local unix sockets by default, it's not a huge deal, but it causes some issues that I think are worthy of discussion.

First, HTTP doesn't have the same concepts of user permissions. So any permissions that you set on the HTTP endpoint do not use the unix permissions. Log files can potentially contain sensitive information. While we scrape out passwords, I don't know if we have done a security audit on everything that gets emitted in the log.

That's my main concern. I don't like the idea of exposing the log over an HTTP endpoint since it compromises potential security sensitive sections and compromises the typical unix permissions. I have the same concerns with Consul and, while I loved that feature while I was using Consul, this was my primary reason for not making logs available through an HTTP endpoint for a long time. At least unix sockets still use unix permissions to some extent even if the application never sees them. Making it available by a TCP socket completely compromises that.

To be fair, this complaint is true of everything under `/debug/pprof`, but the rest of the information in `/debug/pprof` is less sensitive to being scraped by a bad actor and less capable of creating a security hole.

e-dard:
@jsternberg this is the first thing I considered when implementing this. We are having an internal discussion about it.

PR #9376, opened 2018-01-29
Move ApplyEnvOverrides to toml package

mark-rushakoff:
This code has been duplicated to other projects and its implementations
have grown out of sync. Now the code can live in one place instead of
coupled with a particular struct.

- [x] Rebased/mergable
- [x] Tests pass

### DIFF
--- a/cmd/influxd/run/config.go
+++ b/cmd/influxd/run/config.go
@@ -1,16 +1,13 @@
 package run
 
 import (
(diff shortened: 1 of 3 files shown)


mark-rushakoff:
This is out of date as of https://github.com/influxdata/influxdb/pull/9553. I'll update it later.

PR #5221, opened 2015-12-24
Compaction concurrency

jwilder:
This PR changes how compactions run so that multiple compactions can run concurrently.  It also fixes several bugs that were discovered during this refactor.

Previously, the compaction planning was based on sizes of files and a single planning step.  The sizes were somewhat arbitrary and worked well for lower numbers of series <100k.  In the larger number of series, compactions could pile up and get bogged down compacting larger files causing lots of smaller files to pile up.  The single planner also could get backlogged causing files to pile.  Also, this size based strategy sometimes created files that would get isolated between larger files and never compacted further depending on the shape of the data and how it was compacted.

This PR switches the compactions to dedicated level based compactions.  It works as follows:
* Each level has a dedicated goroutine that compacts a given level.  It runs a planning step for that level and then kicks off multiple goroutines to compact that level concurrently.
* Cache snapshot compactions run continuously writing out level 1 files
* Level 1 files are compacted in multiple groups of 2 concurrently using a fast compaction (explained later).  These compactions create level 2 files.
* Level 2 files are compacted in multiple groups of 3 concurrently using a fast compaction method.  These produce level 3 files.
* Level 3 files are compacted in multiple groups of 4 concurrently using an optimal compaction method.  These produce level 4 files.
* Level 4 files are compacted in multiple groups of 4 concurrently using an optimal compaction method.  These are compacted back into new level 4 files until they hit the max size at which point they are skipped.

The fast compaction method basically combines multiple TSM files, but does not decode and re-chunk blocks for the same key unless there is an overlap in the points.  The optimal compaction method will decode and re-chunk to produce the fewest number of blocks containing 1000 points each.  My first attempt was to always run the optimal compaction for all levels, but it can get backed up causing many TSM files to accumulate.  Level 1 and 2 use the fast compaction to try to combine all the files produced (rather quickly) from the snapshot compactions so that they don't grow too fast an impact query performance.  When they get to a larger size in level 3 & 4, the optimal compaction increases the compression ratio by combining more files at once.  This takes more time though so we only want to do it when we know we are going to produce full blocks.

This strategy was worked out over multiple tests of varying sizes of series (up to 2.5M) and seems to work well in those tests.  Unfortunately, the code could be simpler and I plan to clean it up in subsequent PR as I am short on time.

Other changes in this PR are:

[...]

PR #9380, opened 2018-01-30
Series creation improvements

jwilder:
This fixes a few performance issues I was seeing with series creation.
* There was lock contention on the series file when `LogFile.execSeriesEntry` was run.  We already had the parsed series data further up the stack and `LogFile.execSeriesEntry` would need to look it up again and re-parse it.  I tacked on the parsed series data to the `LogEntry` and `execSeriesEntry` will use it if exists, otherwise fall back to looking it up in the `SeriesFile`.
* Inserting to RHH was creating lots of little allocations.  I switched it to use some temp buffers that can be re-used.
* The `mmap.Map` call `madvise(MADV_RANDOM)` which was trigger a lot of read IOPS on the series files.  I removed that and seems to reduce read IOPS.
* Loading a big TSM file was also trigger lots of read IOPS due to page faults when scanning the index.  I added a hint `madvise(MADV_WILLNEED)` that helps a bit.  I think a better solution would be to store some extra info in the TSM index to avoid scanning at load time.

- [ ] Rebased/mergable
- [ ] Tests pass
- [ ] CHANGELOG.md updated

### DIFF
--- a/pkg/mmap/mmap_unix.go
+++ b/pkg/mmap/mmap_unix.go
@@ -10,7 +10,6 @@ package mmap
 import (
 	"os"
 	"syscall"
-	"unsafe"
 )
 
 // Map memory-maps a file.
@@ -35,11 +34,6 @@ func Map(path string, sz int64) ([]byte, error) {
 
 	data, err := syscall.Mmap(int(f.Fd()), 0, int(sz), syscall.PROT_READ, syscall.MAP_SHARED)
 	if err != nil {
-		return nil, err
-	}
-
-	if _, _, err := syscall.Syscall(syscall.SYS_MADVISE, uintptr(unsafe.Pointer(&data[0])), uintptr(len(data)), uintptr(syscall.MADV_RANDOM)); err != 0 {
-		Unmap(data)
 		return nil, err
 	}
 
--- a/pkg/rhh/rhh.go
+++ b/pkg/rhh/rhh.go
@@ -19,6 +19,8 @@ type HashMap struct {
 	threshold  int64
 	mask       int64
 	loadFactor int
+
+	tmpKey []byte
 }
 
 func NewHashMap(opt Options) *HashMap {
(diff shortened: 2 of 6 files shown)


pkg/rhh/rhh.go
```diff
@@ -268,6 +270,16 @@ func pow2(v int64) int64 {
 	panic("unreachable")
 }
 
+func assign(x, v []byte) []byte {
+	if cap(x) < len(v) {
+		x = make([]byte, len(v))
+	} else {
```
e-dard: Could make this function a little shorter like this?

```go
func assign(x, v []byte) []byte {
	if cap(x) < len(v) {
		x = make([]byte, len(v))
	}
	copy(x, v)
	return x[:len(v)]
}
```

tsdb/engine/tsm1/reader.go
```diff
@@ -1374,6 +1375,10 @@ func (m *mmapAccessor) init() (*indirectIndex, error) {
 		return nil, fmt.Errorf("mmapAccessor: invalid indexStart")
 	}
 
+	// Hint to the kernal that we will be reading the file.  It would be better to hint
+	// that we will be reading the index section, but that doesn't seem to work ATM.
+	_ = madvise(m.b, syscall.MADV_WILLNEED)
```
e-dard: Could we add this call to the Series File index? We will always need to read from that file too right?

pkg/mmap/mmap_unix.go
```diff
@@ -38,11 +37,6 @@ func Map(path string, sz int64) ([]byte, error) {
 		return nil, err
 	}
 
-	if _, _, err := syscall.Syscall(syscall.SYS_MADVISE, uintptr(unsafe.Pointer(&data[0])), uintptr(len(data)), uintptr(syscall.MADV_RANDOM)); err != 0 {
```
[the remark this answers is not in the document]
jwilder: I've had to remove any place where we're using `MADV_RANDOM` in the past due to adverse performance problems.  See: https://github.com/influxdata/influxdb/pull/8872 and https://github.com/influxdata/influxdb/pull/5221.  Other projects (https://github.com/boltdb/bolt/issues/691) have seen similar issues.

I only add `madavise` calls in specific spots after testing them quite extensively and can see that they improve what I'm measuring.  Despite what the docs say they do, it's kind of a black box as to what kernel actually does.

tsdb/engine/tsm1/reader.go
jwilder: I was not measuring series file IOPS so I would not add this there without seeing the actual change in performance.   This does have _some_ benefit with larger TSM files that are not in the page cache though.

pkg/mmap/mmap_unix.go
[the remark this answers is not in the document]
e-dard: @jwilder I guess what I'm saying is: we should do some explicit experimentation in the future on all the sub systems and see if any advice other than `NORMAL` is appropriate anywhere. If we're doing really sequential stuff then `MADV_SEQUENTIAL` might be even better for read IOPS than `MADV_NORMAL` for example.

PR #9388, opened 2018-02-02
update to logs to print start/end values when relevant

aanthony1243:
update variable value check, and log messages to be more informative of what time bounds are set in backup.  

- [X] Rebased/mergable
- [X] Tests pass

### DIFF
--- a/cmd/influxd/backup/backup.go
+++ b/cmd/influxd/backup/backup.go
@@ -231,10 +231,10 @@ func (cmd *Command) backupShard(db, rp, sid string) error {
 
 	if cmd.isBackup {
 		cmd.StdoutLogger.Printf("backing up db=%v rp=%v shard=%v to %s since %s",
-			db, rp, sid, shardArchivePath, cmd.since)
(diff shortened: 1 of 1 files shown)


cmd/influxd/backup/backup.go
```diff
@@ -229,9 +229,13 @@ func (cmd *Command) backupShard(db, rp, sid string) error {
 		return err
 	}
 
-	cmd.StdoutLogger.Printf("backing up db=%v rp=%v shard=%v to %s since %s",
-		db, rp, sid, shardArchivePath, cmd.since)
-
+	if cmd.isBackup {
+		cmd.StdoutLogger.Printf("backing up db=%v rp=%v shard=%v to %s since %s",
+			db, rp, sid, shardArchivePath, cmd.since)
```
e-dard: Can we format the time appropriately? The default stringer for a `time.Time` will include the monotonic clock offset, if it's set, e.g., something like `m=+0.000416596`. I don't think this is relevant to the user.

PR #9396, opened 2018-02-05
Fix stddev() call to report itself as always returning a float

jsternberg:
Backport of #9390.

- [x] Rebased/mergable
- [ ] Tests pass
- [x] CHANGELOG.md updated


jsternberg:
This isn't needed. We're going to catch up the 1.5 branch to master on the next release candidate instead.

PR #9399, opened 2018-02-05
Allow customizing the unix socket group and permissions created by the server

### LINKED ISSUE #9266: Unix domain socket doesn't have the right file permission
## Bug report
**System info:** Ubuntu 16.04

Steps to reproduce:

1. install influxdb by 
````sh
curl -sL https://repos.influxdata.com/influxdb.key | sudo apt-key add -
source /etc/lsb-release
echo "deb https://repos.influxdata.com/${DISTRIB_ID,,} ${DISTRIB_CODENAME} stable" | sudo tee /etc/apt/sources.list.d/influxdb.list
````
2. edit `/etc/influxdb/influxdb.conf` enable http service over unix domain socket

3. restart influxdb

influxdb default run with user `influxdb` and create unix domain socket `influxdb.sock` without write permission, finally it will cause nginx could connect this socket!

so i have 2 chooses:
1. custom influxdb to run with root or my nginx user group.
2. or execute `chmod a+x influxdb.sock` evenytime influxdb restart.

### PR DESCRIPTION
**Pull request by jsternberg**
Fixes #9266.

- [x] Rebased/mergable
- [x] Tests pass

### DIFF
--- a/toml/toml.go
+++ b/toml/toml.go
@@ -3,9 +3,11 @@ package toml // import "github.com/influxdata/influxdb/toml"
 
 import (
 	"encoding"
+	"errors"
 	"fmt"
 	"math"
 	"os"
+	"os/user"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: toml/toml.go [hunk 1]
```diff
@@ -89,3 +91,51 @@ func (s *Size) UnmarshalText(text []byte) error {
...
+	if err != nil {
+		return err
+	}
+	*m = FileMode(mode)
+	return nil
+}
+
+func (m FileMode) MarshalText() (text []byte, err error) {
+	if m != 0 {
+		return []byte(fmt.Sprintf("0%03o", m)), nil
```
mark-rushakoff: Could be `%04o`?

jsternberg: Is there a reason for `%04o`?

jsternberg: One thing I just noticed is that the output from `MarshalText` appears to be different than what `UnmarshalText` parses so I'll update that.

mark-rushakoff: `%03o` sets the output width to three and left pads with `0`s. The leading `0` could be be part of the padding, i.e. `%04o` should be equivalent to `0%03o` which would be the same as `00%02o` etc., as long as the value `m` didn't exceed the given width.

jsternberg: I think it would be safer to explicitly put the zero. It's possible for a file to have the mode `1777` because of things like the [sticky bit](https://en.wikipedia.org/wiki/Sticky_bit).

On the other hand, the value here is always going to be an octal so we could go and declare that `777` will always be interpreted as an octal, even if there is no leading zero. Then we would use `%04` instead (and get rid of the trailing `o` since it's just confusing).

jsternberg: The trailing zero is to tell `Sprintf` to use octal and I just remembered that so just ignore that part. We would use `%04o`.

jsternberg:
@mark-rushakoff I think I addressed your comments on this.

PR #9401, opened 2018-02-06
Fix windows history file location

oiooj:
- [x] Rebased/mergable
- [x] Tests pass
- [ ] CHANGELOG.md updated

An issue reported by a windows user in `InfluxDB China User Group`.

The `windows` OS  have no `$HOME` env, so he got an empty `c.historyFilePath`, but there is an error every time. 

Now use `USERPROFILE ` env on windows.

### DIFF
--- a/cmd/influx/cli/cli.go
+++ b/cmd/influx/cli/cli.go
@@ -15,6 +15,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"reflect"
+	"runtime"
 	"sort"
 	"strconv"
 	"strings"
@@ -198,9 +199,20 @@ func (c *CommandLine) Run() error {
 	c.Version()
 
 	// Only load/write history if HOME environment variable is set.
+	var historyDir string
+	if runtime.GOOS == "windows" {
+		if userDir := os.Getenv("USERPROFILE"); userDir != "" {
+			historyDir = userDir
+		}
+	}
+
 	if homeDir := os.Getenv("HOME"); homeDir != "" {
-		// Attempt to load the history file.
-		c.historyFilePath = filepath.Join(homeDir, ".influx_history")
+		historyDir = homeDir
+	}
+
+	// Attempt to load the history file.
+	if historyDir != "" {
+		c.historyFilePath = filepath.Join(historyDir, ".influx_history")
 		if historyFile, err := os.Open(c.historyFilePath); err == nil {
 			c.Line.ReadHistory(historyFile)
 			historyFile.Close()
@@ -1054,6 +1066,9 @@ func (c *CommandLine) history() {
 }
 
 func (c *CommandLine) saveHistory() {
+	if c.historyFilePath == "" {
+		return
+	}
 	if historyFile, err := os.Create(c.historyFilePath); err != nil {
 		fmt.Printf("There was an error writing history file: %s\n", err)
 	} else {


cmd/influx/cli/cli.go
```diff
@@ -198,13 +198,20 @@ func (c *CommandLine) Run() error {
...
-		c.historyFilePath = filepath.Join(homeDir, ".influx_history")
-		if historyFile, err := os.Open(c.historyFilePath); err == nil {
-			c.Line.ReadHistory(historyFile)
-			historyFile.Close()
-		}
+		historyDir = homeDir
+	} else if userDir := os.Getenv("USERPROFILE"); userDir != "" {
+		historyDir = userDir
+	} else {
+		historyDir = os.TempDir()
```
jsternberg: I think we should avoid this as it becomes a potential security hole. If a password gets put into a command such as from create user, I don't know if that's redacted in the history file and it probably isn't. The file in the temp folder would make it open to everyone on the system.

I would rather we just don't open the history file at all in that case and don't save history. Maybe warn the user that's what we're doing.

cmd/influx/cli/cli.go
```diff
@@ -198,13 +198,20 @@ func (c *CommandLine) Run() error {
...
+	var historyDir string
 	if homeDir := os.Getenv("HOME"); homeDir != "" {
-		// Attempt to load the history file.
-		c.historyFilePath = filepath.Join(homeDir, ".influx_history")
-		if historyFile, err := os.Open(c.historyFilePath); err == nil {
-			c.Line.ReadHistory(historyFile)
-			historyFile.Close()
-		}
+		historyDir = homeDir
+	} else if userDir := os.Getenv("USERPROFILE"); userDir != "" {
```
jsternberg: Do you know of an example of another piece of software that reads `USERPROFILE` on Linux because of Windows? I'm not necessarily against this, but I would like to see some precedent for doing this.

I think it might be better to only check `USERPROFILE` if the OS is Windows.

oiooj:
@jsternberg  You're right.

cmd/influx/cli/cli.go
```diff
@@ -198,9 +199,20 @@ func (c *CommandLine) Run() error {
...
+
+	if runtime.GOOS == "windows" {
+		if userDir := os.Getenv("USERPROFILE"); userDir != "" {
+			historyDir = userDir
+		}
+	}
+
+	// Attempt to load the history file.
+	if historyDir != "" {
+		c.historyFilePath = filepath.Join(historyDir, ".influx_history")
```
jsternberg: This is almost what we need. I think you need to modify `saveHistory()` so it won't panic when trying to save the history to an empty file.

cmd/influx/cli/cli.go
```diff
@@ -198,9 +199,20 @@ func (c *CommandLine) Run() error {
...
 
 	// Only load/write history if HOME environment variable is set.
+	var historyDir string
 	if homeDir := os.Getenv("HOME"); homeDir != "" {
-		// Attempt to load the history file.
-		c.historyFilePath = filepath.Join(homeDir, ".influx_history")
+		historyDir = homeDir
+	}
+
+	if runtime.GOOS == "windows" {
```
jsternberg: Can you make it so that `USERPROFILE` doesn't get read if `HOME` is set? That way, if both `HOME` and `USERPROFILE` are set, then it uses `HOME`.

oiooj:
@jsternberg Done

PR #9406, opened 2018-02-06
Use dep for dependency management instead of gdm

### LINKED ISSUE #9404: Use dep for dependency management instead of gdm
### Feature Request

__Proposal:__ Use dep for dependency management instead of gdm.

__Current behavior:__ We use gdm for dependency management. It is simple, allows us to pin dependenies, and keeps our workspaces in order. But, it is unsupported and not typically used within the general Go community. It also causes a lot of problems when it comes to managing our local workspaces. Some developers use a global go path and some create their own go path for each project to avoid the problems.

__Desired behavior:__ `dep` is the new standard for Go dependency management. The file format is now considered stable and it is likely safe for us to start transferring to use it. The general opinion of most on the team is either neutral or positive towards `dep`. While there are other options for dependency management, `dep` is the only one blessed by the general Go team and is one that integrates best with the default Go compiler chain. Tools like glide usually require glide to build. I personally like `govendor`, but `dep` is the one that will likely be integrated with the Go toolchain at some point and `dep` is supposed to include versioning capabilities and constraints.

__Use case:__ An overall improvement to our build process. It will make it easier to allow us to ship source only packages with the vendored dependencies included in the source package and it will make it easier for newcomers to build without them having to learn how to use gdm to get the proper dependencies.

jsternberg:
Fixes #9404.

- [x] Rebased/mergable
- [x] Tests pass


jsternberg:
At the moment, this just takes everything from `Godeps` and uses a revision constraint in the `Gopkg.toml` file. We can then migrate to actually using versions to make it easier to update project dependencies in the future.

mark-rushakoff:
Due to how late we are in the 1.5 release cycle, I'd prefer to hold off on this until we branch to 1.6 development.

jsternberg:
That was my intention. I'll add a 1.6 milestone to this to clarify that.

rbetts:
I approve moving to `dep` after 1.5 is branched.

e-dard:
We will need quite a bit of coordination around this, as once it's merged a lot of internal stuff will break. We should get all our ducks in a row before merging.

jwilder:
@e-dard @jsternberg What do you think about keeping both `gdm` and `dep`bits in place (for now) when first merged, update all the other places that use `gdm` to switch to `dep`, then remove all the `gdm`bits?  I think we could merge and move forward with this in smaller pieces, without breaking everything else that uses `gdm` at the same time.

e-dard:
@jwilder I would be in favour of that if it's feasible.

jsternberg:
I think that would result in a dependency hell. It's not too difficult for us to just grep for `gdm` and replace them and doing it all at once avoids a system where `dep` gives one set of dependencies and `gdm` gives another set and the build only uses on of the two.

e-dard:
@jsternberg I think you're misunderstanding the issue. We're basically suggesting to not remove the `Godeps` file in this PR. That way other internal systems and tools (static checkers, nightly harnesses, performance benchmark robots) can still go and build InfluxDB even after this PR is merged using `gdm`.

Then we can independently update those tools/robots/whatevers to use `dep`. Then we can remove the `Godeps` file.

jsternberg:
My concern is that two different lock files will get it if sync. I can potentially add a script though to ensure that the lock file and Godeps are consistent. Then we could keep both in.

rbetts:
@jsternberg I don't want to spend time scripting support for syncing the two systems. Support for both is only transitional.

I like the idea of support both `gdm` and `dep` for a week or two as we transition.

@e-dard @mark-rushakoff @aanthony1243 Let's make a list of stuff not in this PR that will need to learn `go dep`?

jsternberg:
I don't think I posted it here, but I modified the PR to not delete the `Godeps` file for a smoother transition.

Gopkg.lock
```diff
@@ -0,0 +1,301 @@
...
+[[projects]]
+  name = "github.com/matttproud/golang_protobuf_extensions"
+  packages = ["pbutil"]
+  revision = "c12348ce28de40eed0136aa2b644d0ee0650e56c"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/mschoch/smat"
+  packages = ["."]
+  revision = "90eadee771aeab36e8bf796039b8c261bebebe4f"
```
mark-rushakoff: Both github.com/mschoch/smat and github.com/willf/bitset are present here but not in Godeps. I'm not sure if that means they're a transitive dependency that gdm didn't pick up, maybe?

All the other revisions appear to match gdm at a glance.

jsternberg: This one in particular seems to be a direct dependency of roaring bitmap so that is definitely strange that GDM didn't pick it up.

PR #9407, opened 2018-02-06
Install autoconf in the circle build for fpm

jsternberg:
Upstream must have changed and now autoconf isn't installed as an
implicit dependency of one of the packages anymore. Install is
explicitly so the fpm portion of the circle build works correctly.

- [x] Rebased/mergable
- [x] Tests pass


jsternberg:
Looks like we also needed to install `libtool` as a separate package.

PR #9415, opened 2018-02-08
restore `MetaClient`, which is needed by store

stuartcarnie:
Was incorrectly flagged as unused by some automated tools


hercules-influx:
```/tmp/042603362/src/github.com/influxdata/influxdb/services/storage/service.go:21:3: func Database is unused (U1000)
/tmp/042603362/src/github.com/influxdata/influxdb/services/storage/service.go:22:3: func ShardGroupsByTimeRange is unused (U1000)
```

stuartcarnie:
@e-dard looks like `megacheck` is flagging a false positive here, as I had to re-add the `MetaClient`

PR #9423, opened 2018-02-09
Fix imports of multiple databases in a single import file from `influx -import`

### LINKED ISSUE #9336: influx -import may import to wrong database

jsternberg:
If multiple databases were specified, then the earlier writes would be
written to the later database and/or retention policy because points were
only written out after the batch size was reached. This forces the batcher
to flush whenever the database context switches.

Fixes #9336.


hercules-influx:
``````

hercules-influx:
``````

jsternberg:
@rbetts this is a revert of a change where I didn't remember to check for the CLA before merging. The contributor couldn't figure out the policy for how to sign the CLA so I reverted his change and redid it. I think the main part of the code that was unique was the test case and the other parts in common are trivial (or were recommended by me to add during code review).

PR #9439, opened 2018-02-14
Remove IsAdmin() method from meta.User interface

mark-rushakoff:
The method was only called in tests.


hercules-influx:
``````

PR #9445, opened 2018-02-14
Suppress the InfluxDB banner if the log output is not a TTY

stuartcarnie:
This PR removes the InfluxDB logo when output is redirected to a non-TTY device in order to strip extraneous data from log files.

### DIFF
--- a/cmd/influxd/run/command.go
+++ b/cmd/influxd/run/command.go
@@ -94,7 +94,7 @@ func (cmd *Command) Run(args ...string) error {
 	}
 
 	// Print sweet InfluxDB logo.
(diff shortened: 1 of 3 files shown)


cmd/parse.go
```diff
@@ -27,3 +32,13 @@ func ParseCommandName(args []string) (string, []string) {
 	}
 	return "", args
 }
+
+// IsTerminal checks if w is a file and whether it is an interactive terminal session.
+func IsTerminal(w io.Writer) bool {
```
jsternberg: I think it would be better to just expose this method from the logger package rather than to copy it here.

Issue #9438, opened 2018-02-14
Improve config options for HTTP logging 


gunnaraasen:
### Feature Request

__Proposal:__ Improve HTTP logging management. This is a two part proposal. 
- [ ] Add a log level config setting for HTTP logs (e.g. “info” level logs all requests and “error” level logs only non-2xx responses)
- [x] Add a way to split out HTTP logs into a separate access log. Fixed with #9449

__Current behavior:__ HTTP logs are always interspersed with regular database internal logs. Theres no way to separate them aside from filtering for lines with the `[httpd]` string. There’s also no way to selectively limit logs for successful HTTP response logs.

__Use case:__ There are often times where having only HTTP errors logged separately is significantly more useful for debugging InfluxDB issues than having to filter out HTTP logs from InfluxDB internal service logs. Full HTTP logs for all responses can quickly become verbose with many logs for 200 and 204 responses that are usually filtered out in most cases. Overall, the usefulness of HTTP logs in InfluxDB is limited without another tool to filter them. Separating the logs into a separate output is also a common idiom used in servers like Apache.

PR #9449, opened 2018-02-15
Added option to write HTTP request logs to separate file

stuartcarnie:
Added a new option for the `[http]` service configuration to redirect request logging to a separate file, to address part of #9438

> Add a way to split out HTTP logs into a separate access log.

The new key is as follows

```
[http]
  # When HTTP request logging is enabled, this option specifies the path where 
  # log entries should be written. If unspecified, the default is to write to stderr, which
  # intermingles HTTP logs with internal InfluxDB logging.
  #
  # If influxd is unable to access the specified path, it will log an error and fall back to writing
  # the request log to stderr.
  # access-log-path = ""
```

**NOTES:** 

* when request logging is redirected to a file, the `[httpd] ` prefix is stripped, so access log parsing tools like [lnav](https://lnav.org) can render them without additional modification.
* to rotate the log file, use the `copytruncate` method of `logrotate` or similar to leave the original file in place.

### DIFF
--- a/etc/config.sample.toml
+++ b/etc/config.sample.toml
@@ -217,6 +217,10 @@
   # Determines whether HTTP request logging is enabled.
   # log-enabled = true
 
+  # Specifies whether HTTP request logging is written to the specified path when enabled.
+  # If influxd is unable to access the specified path, it will log an error and fall back to stderr.
+  # access-log-path = ""
+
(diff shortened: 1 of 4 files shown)


stuartcarnie:
I improved the config help

PR #9452, opened 2018-02-15
Reduce time spent under Lock

e-dard:
This PR moves the `inmem` index lock around so that it's not held for as long as it was previously.

It improve `inmem` series creation performance by ~30% with my testing.


e-dard:
This didn't work so well after we changed the way we use `CreateSeriesIfNotExists` in `inmem`.

PR #9460, opened 2018-02-16
Fix a performance regression in subqueries with implicit fills

### LINKED ISSUE #9271: 1.4.0 regression: fill() in a subquery causes long delay and incorrect result
### Bug report

__System info:__ influxdb-1.4.2-1.x86_64 on CentOS7

__Steps to reproduce:__

On my database:

`SELECT sum("BTU_Hr") FROM (SELECT mean("value") as "BTU_Hr" FROM "modbus_value" WHERE ("type_instance" = 'firing-rate') AND time >= now() - 1h GROUP BY time(5m) fill(36))
`

From scratch:

```
> create database testdb
> use database testdb
> insert test value=2
> SELECT mean("value") FROM "test" WHERE time >= now() - 10m GROUP BY time(5m) fill(36)
name: test
time                mean
----                ----
1516650300000000000 36
1516650600000000000 36
1516650900000000000 2
>
```

(note first timestamp, use that in next query to work around other parsing bug(?))

```
> SELECT SUM(mean) FROM (SELECT mean("value") FROM "test" WHERE time >= 1516650300000000000 GROUP BY time(5m) fill(36))
name: test
time sum
---- ---
0    3138415850
>
```
Note the incorrect value and the very long wait.

__Expected behavior:__ A value is quickly returned
__Actual behavior:__ Influxdb returns after long delay with incorrect value

__Additional info:__ 

This works if I use fill(null) instead:

```
> SELECT SUM(mean) FROM (SELECT mean("value") FROM "test" WHERE time >= 1516650300000000000 GROUP BY time(5m) fill(null))
name: test
time sum
---- ---
0    2
```
This seems to be a regression between 1.3.8 and 1.4.0

jsternberg:
During the compilation refactor, it was intended to retrieve the
information from the query into a new data format that would then be
used to build the query. The problem is that the actual execution is
still done on the AST so the implicit change of `fill(null)` to
`fill(none)` (which is done to prevent the outer query from churning
through and ignoring a bunch of null points) was never done.

Fixes #9271.

### DIFF
--- a/query/compile.go
+++ b/query/compile.go
@@ -782,6 +782,14 @@ func (c *compiledStatement) subquery(stmt *influxql.SelectStatement) error {
 	// null values with a redundant fill iterator.
 	if !subquery.Interval.IsZero() && subquery.FillOption == influxql.NullFill {
 		subquery.FillOption = influxql.NoFill
+		// TODO(jsternberg): The query engine should use the compiled data we
+		// retrieve during compilation and not require setting the fill option
+		// on the statement, but it currently doesn't so set the original
+		// statement to fill(none). This makes the previous statement useless,
+		// but we want to eventually use that information instead of this for
+		// the actual execution.
+		// See https://github.com/influxdata/influxdb/issues/9271.
+		stmt.Fill = influxql.NoFill
 	}
 
 	// Inherit the grouping interval if the subquery has none.


jsternberg:
This does, in fact, not fix the issue it says it will fix. I think I've figured out the actual issue so I'll update this when I'm done with the fix.

PR #9461, opened 2018-02-17
Do not drop on the floor small buffers

CAFxX:
Currently if a buffer from the buffer is too small to satisfy its request then we simply drop it and allocate a new one.

This change puts it back in the pool and then allocates a new one.


jwilder:
Thanks @CAFxX!

PR #9462, opened 2018-02-20
Add bootstrap service, and stop allowing unauthenticated access when auth is enabled and no admin user exists

aeijdenberg:
### Background

Currently a new `influxd` with `auth-enabled` set to `true` has the surprising behavior of allowing unauthenticated access, presumably in order to allow the first user who finds the new server to be able to bootstrap it.

Unauthenticated installers combined with modern techniques such as use of Let's Encrypt can have surprising side-effects.

See for example [this DefCon presentation](https://media.defcon.org/DEF%20CON%2025/DEF%20CON%2025%20presentations/DEFCON-25-Hanno-Boeck-Abusing-Certificate-Transparency-Logs.pdf) last year as it relates to dealing with unauthenticated WordPress installations, which follow a similar anti-pattern.

Relevant certificate search for registered certs with a name beginning with `influx.*`: <https://crt.sh/?q=influx.%25>

There are likely a lot of publicly exposed servers out there - defaults are important (separate to this PR, it would be good to consider enabling auth by default).

### Proposed change

This PR does two things:

1. Fixes the `TODO` to stop allowing unauthenticated access when `AdminUserExists()` returns `false`.

2. Provides `[bootstrap]` stanza to the TOML configuration that allows an operator to specify a username / password for an admin user, along with a list of seed databases to create:

```toml
[bootstrap]
  admin_username = "admin"
  admin_password = "s3cr4t"
  seed_databases = ["foo", "bar", "baz"]
```

No default values are set, so not specifying these will not create a bootstrap user. These must be explicitly set to apply.

Note also that being able to specify bootstrap data via configuration is a pattern that works well with many orchestration layers such as [Bosh](https://bosh.io/) which we intend to use for integration with Prometheus in our CloudFoundry monitoring stack.

When this PR lands, we'll update our fork of the [influxdb-boshrelease](https://github.com/govau/influxdb-boshrelease) to add.

Thank you for your consideration.


aeijdenberg:
@corylanou as your GitHub handle was on the TODO that this addresses.

aeijdenberg:
Gentle ping?

mark-rushakoff:
This looks like a useful and very sane pattern to adopt.

@aeijdenberg we'll need you to sign [our CLA](https://www.influxdata.com/legal/cla/) before we can review or accept the PR.

I think we can target this for 1.6, if okay with @timhallinflux and @rbetts.

aeijdenberg:
Thanks @mark-rushakoff - CLA is now signed.

jsternberg:
Hello, can you split the bootstrap portions from the security portions? I'm interested in merging the "Stop allowing unauthenticated access when auth is enabled and no admin user exists" portion, but less interested in merging the bootstrap portion. The kind of bootstrapping that I see in the code seems like it is something that is better served by puppet/chef/ansible than the database itself and I would like to separate them so this can get into 1.6.

Although I do think it would be interesting to allow configuring the schema in the configuration file. I would like to separate it though so the security fix (which I would count as a kind of bug) is separated from the bootstrap (which would act more as a feature).

aeijdenberg:
@jsternberg , yes it should be easy enough to split it out, and I've started to so - however I'm running into problems with the tests as a result. The bootstrap functionality also made it easier.

In any case, I've added what I've had time to do today to a new PR (https://github.com/influxdata/influxdb/pull/9797), and I've renamed this one slightly to disambiguate.

jsternberg:
Thanks for splitting it out, but on review, it looks like there's still some discussion about what the final goal of this is and how we can get there without breaking everyone's current bootstrap process for authentication.

It seems to me there are 3 options.

1. Allow a user to only run `CREATE USER` when auth is enabled and there is not admin users. The `CREATE USER` command needs to create an admin user which can then be used for subsequent commands. Add an optional bootstrap section so that this initial command can require authentication if the user supplies it. This is the version that I would not consider a breaking change, even though it will technically break some workflows. The workflows it breaks are completely invalid and I think they're worthy of invalidating.
2. The above, but require the bootstrap section or generate unique credentials on startup and output them to the log. Jenkins will generate unique credentials for its bootstrap user and output them to the log so there's precedent for this. If the bootstrap section is filled out, we would not generate credentials. This is a breaking change.
3. Force the user to set bootstrap credentials. This is a breaking change.

The first one is insecure by default and doesn't really solve the issue, but two and three would break a lot of authentication bootstrap methods at the moment and I don't think we can merge that kind of thing this close to when we release 1.6.

CC @timhallinflux

PR #9465, opened 2018-02-20
Mark a shard as in process of being deleted

benbjohnson:
Without this, deleting a shard could trigger things so that a write would attempt to create the shard again before it was actually deleted.

### DIFF
--- a/tsdb/store.go
+++ b/tsdb/store.go
@@ -54,6 +54,10 @@ type Store struct {
 	// shared per-database indexes, only if using "inmem".
 	indexes map[string]interface{}
 
+	// Maintains a set of shards that are in the process of deletion.
+	// This prevents new shards from being created while old ones are being deleted.
+	pendingShardDeletes map[uint64]struct{}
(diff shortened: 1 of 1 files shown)


tsdb/store.go
```diff
@@ -537,11 +548,29 @@ func (s *Store) DeleteShard(shardID uint64) error {
...
 	delete(s.shards, shardID)
+	s.pendingShardDeletes[shardID] = struct{}{}
 	s.mu.Unlock()
 
+	// Ensure the pending deletion flag is cleared on exit.
+	defer func() {
+		s.mu.Lock()
+		defer s.mu.Unlock()
+		delete(s.pendingShardDeletes, shardID)
+	}()
```
benbjohnson: I added this `defer` so  that it would clean up the `pendingShardDeletes` flag once it's finished.

benbjohnson:
@mark-rushakoff I added an issue to track the test coverage (#9467)

PR #9466, opened 2018-02-20
Fix the logging sample config

jsternberg:
The original PR that modified this was modified without changing the
sample config so this makes the sample config more accurate.


jsternberg:
I'm just going to merge this. It's a change to the sample log file.

PR #9477, opened 2018-02-23
rename + namespace keys, fix doc comment

stuartcarnie:
All keys and operation names use underscore. Do not use periods for any key or identifier names.

Namespacing keys like `level` to `tsm1_level`

* avoid future collisions and 
* confusion with a global `level` which might be interpreted as the log level

## relevant tsm1 keys

| old key | new key |
| ------- | ------- |
| `level`    | `tsm1_level`    |
| `strategy` | `tsm1_strategy` |
| `optimize` | `tsm1_optimize` |

## relevant tsi keys

| old key | new key |
| ------- | ------- |
| `index.tsi.compact_to_level` | `tsi1_compact_to_level` |
| `index.tsi.compact_log_file` | `tsi1_compact_log_file` |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,3 +74,5 @@ man/*.1.gz
 
 # test outputs
 /test-results.xml
+
+/prof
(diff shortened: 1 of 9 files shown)


stevebang:
Docs updated with these changes.

PR #9488, opened 2018-02-27
Improve startup time of inmem index

### LINKED ISSUE #9486: `inmem` index startup is about 3X slower
Compared to the 1.3 or 1.4 releases, the current startup time for the `inmem` index is about 3x slower.

After writing 4M series with `inch -t 400000`, shutdown the database and then restart it:

| release  | startup time |
| ------------- | ------------- |
|1.3.9 |   27s |
|1.4.2  |   27s |
|1.5.0rc3 |  91s |

e-dard:
This PR improves the startup time when using the `inmem` index by
ensuring that the series are created in the index and series file in
batches of 10000, rather than individually.

Fixes #9486.

Using the same dataset as #9486 with 4M series:

| release  | startup time |
| ------------- | ------------- |
|1.3.9 |   27s |
|1.4.2  |   27s |
|1.5.0rc3 |  91s | 
| PR | 20s |

So that's an improvement of `~25%` on `1.3.9`. However @jwilder @stuartcarnie I'm not set on this PR yet as we may not be able to re-use the `keys` buffer as I'm currently doing. I'm going to dig around a bit and see if it needs to be changed to allocate a new `[][]byte` for each batch rather than re-using the same one.

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -725,34 +725,68 @@ func (e *Engine) LoadMetadataIndex(shardID uint64, index tsdb.Index) error {
 		return nil
 	}
 
+	keys := make([][]byte, 0, 10000)
+	fieldTypes := make([]influxql.DataType, 0, 10000)
+
 	if err := e.FileStore.WalkKeys(nil, func(key []byte, typ byte) error {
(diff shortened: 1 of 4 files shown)


e-dard:
Not re-using the buffers doesn't make any real difference to the startup time, it just means more garbage.

PR #9491, opened 2018-02-27
Evaluate a true boolean literal when calculating tag sets

### LINKED ISSUE #9290: 1.4 doesn't allow binary operations on literals
### Bug report

__System info:__ Any 1.4.x or the last 1.5 nightly build

__Steps to reproduce:__

Execute query:
`select count * (1 / 1) from measurement`
or
`select * from measurement where 'a'='a'`

__Expected behavior:__ The same as in the previous versions, i.e. 1.3.x and older. The expression which can be reduced to a constant during AST processing is accepted and correctly evaluated. `'a'='a'` would be equivalent to `true`, `count * (1 / 1)` would be the same as `count * 1`

__Actual behavior:__ The expression is refused by the parser. With `ERR: cannot perform a binary expression on two literals`, or respectively `ERR: invalid expression: 'a' = 'a'`

__Additional info:__ 
Might be related to #9226 

Rule `binary_op` in InfluxQL reference .

__Use case:__ Some use-cases in Grafana using templating variables. 

The first query is derived from autoscaling implementation:
```
SELECT sum("count")/($__interval_ms/3600000) GROUP BY time($__interval)
```
the second is used to add some modes to Grafana dashboard by switching some queries on/off based on templating variables.

jsternberg:
Fixes #9290.

### DIFF
--- a/Godeps
+++ b/Godeps
@@ -14,7 +14,7 @@ github.com/gogo/protobuf 1c2b16bc280d6635de6c52fc1471ab962dc36ec9
 github.com/golang/protobuf 1e59b77b52bf8e4b449a57e6f79f21226d571845
 github.com/golang/snappy d9eb7a3d35ec988b8585d4a0068e462c27d28380
 github.com/google/go-cmp 18107e6c56edb2d51f965f7d68e59404f0daee54
-github.com/influxdata/influxql 7c0f432656229c2084ee825680ef39a2228ccdb3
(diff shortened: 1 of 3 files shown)


jsternberg:
This should fix #9290 in 1.5. It does not backport cleanly to 1.4 since this specific area of the code was modified for TSI. The fix for 1.4 is pretty similar though and is partially shared with this change.

This PR does not require, but is helped out by influxdata/influxql#11. While that PR makes this one unnecessary, I think that this functionality is still useful to reduce confusion when coding.

PR #9499, opened 2018-02-28
Implement basic trigonometry functions

### LINKED ISSUE #9424: Implement trigonometry functions
### Feature Request

__Proposal:__ Implement the trigonometry functions: `sin()`, `cos()`, `tan()`.

__Current behavior:__ They are not implemented.

__Desired behavior:__ They would be implemented.

__Use case:__ Using the trigonometry functions.

Earlier versions of the query engine had a problem with implementing trigonometry functions because these would be a new class of functions labeled transform functions. A transform function gets parsed as a call, but it is effectively another binary operator so query validation works differently. Since earlier versions of the query parser relied on assuming that a call was a selector or an aggregator, this was difficult or impossible.

In 1.4, I changed the query compilation so this assumption is no longer made... as frequently. I think there are a few places where this assumption is still made, but they should be reasonably easy to remove and then we can implement the query engine to handle these calls as part of the normal processing of a binary operator.

jsternberg:
This adds support for math functions into the query language. Math
functions are special because they are transformations and do not access
the filesystem in the same way aggregate functions do. A transformation
takes one point and always outputs one point making it more similar to
binary expressions so these math functions follow the same rules as
binary expressions.

This also supports using math literals (so you can do `sin(1)`) and the
math functions can be used anywhere such as in a field or an expression.
Both of the following are supported:

    SELECT sin(value) FROM cpu
    SELECT value FROM cpu WHERE sin(value) > 0.5

Arguments are in radians. Degrees is not supported.

Fixes #9424.

- [x] Provide example syntax
- [x] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): issue filed or pull request submitted \<[influxdata/docs.influxdata.com#1457](https://github.com/influxdata/docs.influxdata.com/issues/1457)>

### DIFF
--- a/query/compile.go
+++ b/query/compile.go
@@ -250,8 +250,7 @@ func (c *compiledField) compileExpr(expr influxql.Expr) error {
 		return nil
 	case *influxql.Call:
 		if isMathFunction(expr) {
-			// TODO(jsternberg): Implement validation for any math functions.
-			return nil
+			return c.compileTrigFunction(expr)
 		}
 
 		// Register the function call in the list of function calls.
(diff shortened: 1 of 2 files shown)


hercules-influx:
``````

jsternberg:
I'm likely going to refactor this to make use of #9511 instead of doing custom iterators. I think it will use less code and offer greater flexibility so we can shortly follow this with `sqrt(x)` and `pow(x, y)` along with related trigonometry functions like `atan2(x, y)`.

The idea is that all of these functions will be implemented in this project (the `query` package) and they would not be present in the `influxql` project. But, for that, we need to expose an interface there. influxdata/influxql#13 is for that.

jsternberg:
This has been refactored to make use of the math refactor in #9511. It's much simpler now and it will be easy to implement the other value transform iterators without advanced iterator nesting logic.

PR #9509, opened 2018-03-05
Add a logging level to the access log

jsternberg:
If set to error, it will only print 5xx responses. If set to info or
lower, it will print everything.

### DIFF
--- a/etc/config.sample.toml
+++ b/etc/config.sample.toml
@@ -225,6 +225,11 @@
   # the request log to stderr.
   # access-log-path = ""
 
+  # Determines which requests should be logged. If this is set to "error", only 5xx responses will be
+  # logged to the access log or stderr if the access-log-path is empty. If set to "info", it will print
+  # everything.
+  # access-log-level = "info"
+
   # Determines whether detailed write logging is enabled.
   # write-tracing = false
 
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -37,6 +37,7 @@ import (
 	"github.com/influxdata/influxql"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
 const (
@@ -108,11 +109,12 @@ type Handler struct {
 		WritePoints(database, retentionPolicy string, consistencyLevel models.ConsistencyLevel, user meta.User, points []models.Point) error
 	}
 
-	Config    *Config
-	Logger    *zap.Logger
-	CLFLogger *log.Logger
-	accessLog *os.File
-	stats     *Statistics
+	Config         *Config
+	Logger         *zap.Logger
+	CLFLogger      *log.Logger
+	accessLog      *os.File
+	accessLogLevel zapcore.Level
+	stats          *Statistics
 
 	requestTracker *RequestTracker
 }
@@ -1530,7 +1532,14 @@ func (h *Handler) logging(inner http.Handler, name string) http.Handler {
 		start := time.Now()
 		l := &responseLogger{w: w}
 		inner.ServeHTTP(l, r)
-		h.CLFLogger.Println(buildLogLine(l, r, start))
+
+		level := zapcore.InfoLevel
+		if l.Status()/100 == 5 {
+			level = zapcore.ErrorLevel
+		}
+		if h.Config.AccessLogLevel.Enabled(level) {
+			h.CLFLogger.Println(buildLogLine(l, r, start))
+		}
 
 		// Log server errors.
 		if l.Status()/100 == 5 {
--- a/services/httpd/config.go
+++ b/services/httpd/config.go
@@ -1,6 +1,9 @@
 package httpd
 
-import "github.com/influxdata/influxdb/monitor/diagnostics"
+import (
+	"github.com/influxdata/influxdb/monitor/diagnostics"
+	"go.uber.org/zap/zapcore"
+)
 
 const (
 	// DefaultBindAddress is the default address to bind to.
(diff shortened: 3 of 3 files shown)


hercules-influx:
```/tmp/709933322/src/github.com/influxdata/influxdb/services/httpd/handler.go:116:2: field accessLogLevel is unused (U1000)
```

gunnaraasen:
I think it would be more useful to have the `error` log level print any response that is not `2xx`. The `4xx` range of errors can help with debugging and shouldn't show up in normal operations.

etc/config.sample.toml
```diff
@@ -225,6 +225,11 @@
   # the request log to stderr.
   # access-log-path = ""
 
+  # Determines which requests should be logged. If this is set to "error", only 5xx responses will be
+  # logged to the access log or stderr if the access-log-path is empty. If set to "info", it will print
+  # everything.
+  # access-log-level = "info"
```
gunnaraasen: If we do change the error level to print all non-`2xx` codes, we should change the default setting in the example config to `error`.

e-dard:
@jsternberg I don't know what's going on with the static check, but the error is:

```
⇒  megacheck github.com/influxdata/influxdb/...
services/httpd/handler.go:116:2: field accessLogLevel is unused (U1000)
```

rbetts:
@jsternberg PM would like this for 1.7.0; I added the appropriate milestone.

nathanielc:
This should be straightforward to finish up for the 1.7.0 release. @jsternberg and I have some questions about how the feature should work:

1. Should the log level control which http status codes are logged? If so how? I.e. Does error map to 5xx only, does it include 4xx? Does info include all error codes?
2. Alternatively should the user simply specify a mapping of HTTP status codes to log levels, i.e. the use can specifiy all 5xx requests are errors and they will be logged with the error log level. And 4xx could be mapped to errors or infos depending on the users needs.

In other words should this feature introduce a mapping from log levels to status codes, Or status codes to log levels?

timhallinflux:
I think we should just go by the following:
https://www.restapitutorial.com/httpstatuscodes.html

400s and 500s should be logged as errors.
100s-300s should be logged as info.

e-dard:
On the face of it, I don't agree that `4xx` codes should be logged as errors. According to our own [logging style guide](https://github.com/influxdata/influxdb/blob/master/logger/style_guide.md), we should only log things at an `error` level when the _system cannot continue correctly if no action is taken_, and an engineer could remedy the situation operationally.

Most (all??) `4xx` errors do not fall into that category, since the fault is on behalf of the user. 

Note, I'm not making a judgement on which log messages get logged for one setting, and which get logged for another. I'm just saying that we should be setting the log level of each message appropriately, regardless of whether we have set in the config that we want to ignore that log entry or not.

It seems to me that what this feature actually wants to be is "I want to see all log messages associated with non-2xx status codes please", or alternatively "I want to see all log messages for all status codes".

I don't think the log level of those messages is relevant to the config setting.

rbetts:
Let's make the `access-log-level` config setting take a list of HTTP status codes (by major number would be okay) instead of trying to map HTTP status codes to debug/info/error levels of logging. If we are letting people filter the access-log content by HTTP status code, let's do that directly instead of trying to map access logs to the syslog-ish log level.

We already allow the user to log access logs to a separate, dedicated access log file. It doesn't really make sense to try to map that to our syslog logging level.

jsternberg:
I've updated this PR to use the status filter.

etc/config.sample.toml
```diff
@@ -255,6 +255,12 @@
   # the request log to stderr.
   # access-log-path = ""
 
+  # Filters which requests should be logged. Each filter is of the pattern NNN, NNX, or NXX where N is
+  # a number and X is a wildcard for any number. To filter all 5xx responses, use the string 5xx.
```
mark-rushakoff: `a number (1-5, inclusive)` perhaps?

services/httpd/handler.go
```diff
@@ -236,6 +238,15 @@ func (h *Handler) Open() {
 		h.Logger.Info("opened HTTP access log", zap.String("path", path))
 	}
 
+	for _, s := range h.Config.AccessLogStatusFilters {
+		sf, err := newStatusFilter(s)
+		if err != nil {
+			h.Logger.Warn("Dropping invalid access log status filter", zap.String("filter", s), zap.Error(err))
```
mark-rushakoff: It would be better to properly fail on an invalid access log filter. There are often a lot of logs flying by at startup and it would be easy to miss this warning, and then incorrectly believe your filters are configured as intended.

It doesn't look like httpd/config has a Validate method, but other configs do.

services/httpd/handler.go
```diff
@@ -1841,3 +1856,64 @@ func (t *Throttler) Handler(h http.Handler) http.Handler {
...
+// Match will check if the status code matches this filter.
+func (sf *statusFilter) Match(statusCode int) bool {
+	return statusCode/sf.divisor == sf.base
+}
+
+// reStatusFilter ensures that the format is digits optionally followed by X values.
+var reStatusFilter = regexp.MustCompile(`^(\d+)([xX]*)$`)
+
+// newStatusFilter will create a new status filter from the string.
+func newStatusFilter(s string) (*statusFilter, error) {
```
mark-rushakoff: There should be a couple tests for status filter, like testing that `0xx` and `6xx` and `4x0` are rejected filters, `40x` matches `404` but not `411`.

etc/config.sample.toml
jsternberg: I really like the string pattern because it feels natural to me.

services/httpd/handler.go
jsternberg: Removed. It's now a native type for TOML so decoding will just fail.

services/httpd/handler.go
jsternberg: Done. `6xx` won't be rejected though since it matches with the pattern. Do you want me to change the regex to `[1-5]` instead of `\d`?

jsternberg: I did this anyway so `6xx` will now be rejected.

mark-rushakoff: `[1-5]`  would probably be better, if you don't mind. I've skimmed through the spec, and although I don't see where 0xx or 6xx are disallowed, everything in the registered status codes appears to be [100-599].

PR #9525, opened 2018-03-08
Turn the ExecutionContext into a context.Context

jsternberg:
Along with modifying ExecutionContext to be a context and have the
TaskManager return the context itself, this also creates a Monitor
interface and exposes the Monitor through the Context. This way, we can
access the monitor from within the query.Select method and keep all of
the limits inside of the query package instead of leaking them into the
statement executor.

An eventual goal is to remove the InterruptCh from the IteratorOptions
and use the Context instead, but for now, we'll just assign the done
channel from the Context to the IteratorOptions so at least they refer
to the same channel.

### DIFF
--- a/coordinator/statement_executor.go
+++ b/coordinator/statement_executor.go
@@ -56,10 +56,10 @@ type StatementExecutor struct {
 }
 
 // ExecuteStatement executes the given statement with the given execution context.
-func (e *StatementExecutor) ExecuteStatement(stmt influxql.Statement, ctx query.ExecutionContext) error {
+func (e *StatementExecutor) ExecuteStatement(stmt influxql.Statement, ctx *query.ExecutionContext) error {
 	// Select statements are handled separately so that they can be streamed.
 	if stmt, ok := stmt.(*influxql.SelectStatement); ok {
-		return e.executeSelectStatement(context.Background(), stmt, &ctx)
+		return e.executeSelectStatement(stmt, ctx)
 	}
 
 	var rows models.Rows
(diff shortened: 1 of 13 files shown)


hercules-influx:
``````

**Comment by hercules-influx** [comment]
```/tmp/884848090/src/github.com/influxdata/influxdb/query/monitor_test.go:49:11: this value of err is never used (SA4006)
/tmp/884848090/src/github.com/influxdata/influxdb/tsdb/index/tsi1/log_file.go:139:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
/tmp/884848090/src/github.com/influxdata/influxdb/query/execution_context.go:8:6: type key is unused (U1000)
```

stuartcarnie:
@jsternberg did you see the `megacheck` message?

coordinator/statement_executor.go
```diff
@@ -56,10 +56,10 @@ type StatementExecutor struct {
 }
 
 // ExecuteStatement executes the given statement with the given execution context.
-func (e *StatementExecutor) ExecuteStatement(stmt influxql.Statement, ctx query.ExecutionContext) error {
+func (e *StatementExecutor) ExecuteStatement(stmt influxql.Statement, ctx *query.ExecutionContext) error {
 	// Select statements are handled separately so that they can be streamed.
 	if stmt, ok := stmt.(*influxql.SelectStatement); ok {
-		return e.executeSelectStatement(context.Background(), stmt, &ctx)
+		return e.executeSelectStatement(context.Background(), stmt, ctx)
```
stuartcarnie: This is confusing as we now have two independent `context.Context` instances. What is the reasoning behind embedding it within `ExecutionContext` and leaving it as an argument?

The longer term goal is that the `context.Context` would be passed in via the `QueryExecutor#Execute` method, thus coming from the HTTP request or whoever initiated the query.

jsternberg: That appears to be a mistake. The intention is to remove the first argument and use the third (now second) as the first argument.

hercules-influx:
```/tmp/100126914/src/github.com/influxdata/influxdb/query/monitor_test.go:49:11: this value of err is never used (SA4006)
/tmp/100126914/src/github.com/influxdata/influxdb/tsdb/index/tsi1/log_file.go:139:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
```

PR #9531, opened 2018-03-08
Modify the Select call to return a Cursor

jsternberg:
The Cursor returned will be capable of scanning rows into a structure.
It replaces part of the function for why the Emitter existed. The
Emitter would both join the resulting rows and then transform the values
into a models.Row so it could be returned to the results.

In the future, we will be able to use the Cursor directly to write out
values which should be more memory efficient.


hercules-influx:
```/tmp/600041284/src/github.com/influxdata/influxdb/query/monitor_test.go:49:7: this value of err is never used (SA4006)
/tmp/600041284/src/github.com/influxdata/influxdb/tsdb/index/tsi1/log_file.go:139:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
```

hercules-influx:
```/tmp/661102358/src/github.com/influxdata/influxdb/query/monitor_test.go:49:7: this value of err is never used (SA4006)
/tmp/661102358/src/github.com/influxdata/influxdb/tsdb/index/tsi1/log_file.go:139:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
```

hercules-influx:
```/tmp/320425144/src/github.com/influxdata/influxdb/query/monitor_test.go:49:7: this value of err is never used (SA4006)
/tmp/320425144/src/github.com/influxdata/influxdb/tsdb/index/tsi1/log_file.go:139:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
```

PR #9536, opened 2018-03-09
Fix the deadlocks in the monitor tests

jsternberg:
This was previous fixed in b740567 for one test, but the other tests in
the package were still vulnerable. The tests are all updated to use a
cancel context instead since then it doesn't need to be wrapped in a
sync.Once.

### DIFF
--- a/monitor/service_test.go
+++ b/monitor/service_test.go
@@ -2,12 +2,12 @@ package monitor_test
 
 import (
 	"bytes"
+	"context"
 	"expvar"
 	"fmt"
 	"os"
 	"reflect"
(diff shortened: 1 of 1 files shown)


hercules-influx:
```/tmp/956464810/src/github.com/influxdata/influxdb/query/monitor_test.go:49:11: this value of err is never used (SA4006)
/tmp/956464810/src/github.com/influxdata/influxdb/tsdb/index/tsi1/log_file.go:139:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
```

PR #9537, opened 2018-03-09
Simplify tsi1/log_file.go according to megacheck

jsternberg:


hercules-influx:
```/tmp/101126508/src/github.com/influxdata/influxdb/query/monitor_test.go:49:11: this value of err is never used (SA4006)
```

PR #9556, opened 2018-03-12
Series File Tombstone Tracking

### LINKED ISSUE #9514: Tombstone aggregate tracking
Add ability to track tombstone aggregates on series files segments to determine if we need to compact them.

benbjohnson:
## Overview

This pull request adds series & tombstone statistic tracking to the series file compactor. These are calculated on a per-segment basis to later determine if they are past a threshold for compacting.

Fixes #9514.


hercules-influx:
```/tmp/210506072/src/github.com/influxdata/influxdb/query/monitor_test.go:49:11: this value of err is never used (SA4006)
/tmp/210506072/src/github.com/influxdata/influxdb/tsdb/index/tsi1/log_file.go:139:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
```

tsdb/series_partition.go
```diff
@@ -614,6 +648,9 @@ func (c *SeriesPartitionCompactor) compactIndexTo(index *SeriesIndex, seriesN ui
 		return err
 	}
 
+	// Attach stats to compactor.
+	c.Stats = stats
+
 	return nil
 }
```
benbjohnson: Stats are built using the current index snapshot to calculate tombstones. Orphaned tombstones are calculated based on a bitset to determine if a tombstone's series has already been compacted away in an segment.

hercules-influx:
``````

**Comment by hercules-influx** [comment]
```/tmp/795134988/src/github.com/influxdata/influxdb/query/monitor_test.go:49:11: this value of err is never used (SA4006)
```

e-dard:
@benbjohnson `hercules-influx` with an empty comment is a bug in the static checker. Sometimes it means that the branch doesn't build.

e-dard:
@benbjohnson I don't know why it's not emitting the error. I will have to look at it. The issue anyway is:

```
⇒  megacheck github.com/influxdata/influxdb/...
query/monitor_test.go:49:11: this value of err is never used (SA4006)
```

tsdb/series_partition.go
```diff
@@ -389,6 +393,13 @@ func (p *SeriesPartition) compactionsEnabled() bool {
 	return p.compactionsDisabled == 0
 }
 
+// SeriesPartitionStats returns stats generated by compaction.
+func (p *SeriesPartition) Stats() *SeriesPartitionStats {
+	p.mu.Lock()
```
e-dard: `RLock` / `RUnlock`?

jwilder: +1

tsdb/series_partition.go
```diff
@@ -56,6 +58,7 @@ func NewSeriesPartition(id int, path string) *SeriesPartition {
 		CompactThreshold: DefaultSeriesPartitionCompactThreshold,
 		Logger:           zap.NewNop(),
 		seq:              uint64(id) + 1,
+		stats:            NewSeriesPartitionStats(),
```
jwilder: What happens on restart/re-open?  Are any stats accumulated prior to the restart lost?

e-dard:
@benbjohnson what is the status of this PR? Is there an advantage to getting it into the next release? /cc @rbetts

benbjohnson:
@e-dard I need to update based on the feedback but it's been low priority compared to EARs. There's no benefit to adding this to the next release.

e-dard:
don't close.

danxmoran:
@benbjohnson working on clearing our old PR queue, is this something that would make sense to re-target at 2.x?

PR #9568, opened 2018-03-13
Update influxql to include the function type evaluators in the query package

jsternberg:


query/select.go
```diff
@@ -13,6 +13,10 @@ import (
 	"github.com/influxdata/influxql"
 )
 
+var DefaultTypeMapper = influxql.MultiTypeMapper(
```
stuartcarnie: this will allocate a `TypeMapper` interface{} wrapper each time:

```go
valuer := influxql.TypeValuerEval{
 		TypeMapper: DefaultTypeMapper,
 	}
```

However, you can avoid that allocation if you change the declaration to:

```go
var DefaultTypeMapper TypeMapper = influxql.MultiTypeMapper(
```

jsternberg: Done.

PR #9569, opened 2018-03-13
 bugfix for shard collisions resulting from attempted _internal import

aanthony1243:

### DIFF
--- a/tests/backup_restore_test.go
+++ b/tests/backup_restore_test.go
@@ -83,7 +83,6 @@ func TestServer_BackupAndRestore(t *testing.T) {
 
 		// it takes some time for _internal to be created, we want to make sure it's treated
 		// correctly.
-		// wait for the snapshot to write
 		time.Sleep(time.Second * 10)
 
 		// now backup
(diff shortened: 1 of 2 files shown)


tests/backup_restore_test.go
```diff
@@ -81,6 +81,10 @@ func TestServer_BackupAndRestore(t *testing.T) {
 			t.Fatalf("query results wrong:\n\texp: %s\n\tgot: %s", expected, res)
 		}
 
+		// it takes some time for _internal to be created, we want to make sure it's treated
+		// correctly.
+		time.Sleep(time.Second * 10)
```
e-dard: You might want to bump this a little to be on the safe side. What with goroutine scheduling and all, it might not be exactly 10 seconds.

tests/backup_restore_test.go
```diff
@@ -81,6 +81,10 @@ func TestServer_BackupAndRestore(t *testing.T) {
...
+
```
e-dard: In order to guarantee that the `_internal` database has been created, and that your code correctly skips its shards, you could add an assertion here that the `_internal` database exists.

PR #9582, opened 2018-03-15
Improve performance of SHOW SERIES query

stuartcarnie:
Use single `SeriesPointIterator` and combine each shard index into a single `IndexSet`. Assuming series keys are relatively consistent between shards, this shows a 2x improvement in performance for 2 shards. Should show linear improvement in performance, so 10 shards, 10x.

### DIFF
--- a/tsdb/shard.go
+++ b/tsdb/shard.go
@@ -408,6 +408,15 @@ func (s *Shard) Index() (Index, error) {
 	return s.index, nil
 }
 
+func (s *Shard) seriesFile() (*SeriesFile, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	if err := s.ready(); err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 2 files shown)


tsdb/shard.go
```diff
@@ -1237,6 +1251,29 @@ func (a Shards) CreateIterator(ctx context.Context, measurement *influxql.Measur
 	return query.Iterators(itrs).Merge(opt)
 }
 
+func (a Shards) createSeriesIterator(ctx context.Context, opt query.IteratorOptions) (_ query.Iterator, err error) {
+	var (
+		idxs  []Index
+		sfile *SeriesFile
+	)
```
benbjohnson: Can you preallocate `idxs` to `make([]Index, len(a))`?

stuartcarnie: Done 👍

PR #9586, opened 2018-03-15
Allow SHOW SERIES kill

benbjohnson:
Previously, large `SHOW SERIES` queries would block until merging was complete before being interrupted. Interruption checks have been added to the iterator and the merge loop. I also tried adding interruption checks to the sort phase but it's difficult to break out of `sort.Sort()` upon interruption.

### DIFF
--- a/tsdb/index.go
+++ b/tsdb/index.go
@@ -752,12 +752,20 @@ func (itr *seriesPointIterator) readSeriesKeys(name []byte) error {
 
 	// Slurp all series keys.
 	itr.keys = itr.keys[:0]
-	for {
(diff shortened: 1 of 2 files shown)


stuartcarnie:
@benbjohnson would this also need to be wrapped?

https://github.com/influxdata/influxdb/blob/7b1c0e75d395b21047c9c7fefb4c20960a3d17a2/tsdb/shard.go#L1273

benbjohnson:
@stuartcarnie I don't think that section needs to be wrapped. I believe it gets wrapped by `SELECT`.

PR #9595, opened 2018-03-17
Update dependencies and remove unnecessary constraints from Gopkg.toml

jsternberg:
This also adds a small script to write out the dependencies to gdm.
These can be written with:

    $ ./write-gdm-deps.sh > Godeps

This is a temporary script during the migration.

The dependency overrides have also been removed in favor of them being
locked by Gopkg.lock.


mark-rushakoff:
I don't understand why some entries are no longer specified in Gopkg.toml but are still present in Gopkg.lock. For example, github.com/dgryski/go-bitstream and github.com/influxdata/yamux are direct dependencies that are only in Gopkg.lock now. Does the lock file just use the HEAD revision on those packages?

jsternberg:
It confused me too, but that's all part of the process. The `Gopkg.toml` file just specifies _constraints_. It doesn't list dependencies. So if you have a dependency that you don't specifically want to constrain, usually because you don't care, you don't need to put it into the `Gopkg.toml` file.

The default picker for what to use is described in the dep documentation, although it's a bit hidden in the FAQ: https://golang.github.io/dep/docs/FAQ.html#how-does-dep-decide-what-version-of-a-dependency-to-use

So it doesn't necessarily use the HEAD revision for a dependency. For the two you listed, I think it does use HEAD and then lock the revision because neither of those two uses versions.

Gopkg.lock
```diff
@@ -134,7 +141,8 @@
 [[projects]]
   name = "github.com/mattn/go-isatty"
   packages = ["."]
-  revision = "6ca4dbf54d38eea1a992b3c722a76a5d1c4cb25c"
+  revision = "0360b2af4f38e8d38c7fce2a9f4e702702d73a39"
```
mark-rushakoff: This revision went backwards.

Gopkg.lock
```diff
@@ -154,12 +162,14 @@
     "ext",
     "log"
   ]
-  revision = "1361b9cd60be79c4c3a7fa9841b3c132e40066a7"
+  revision = "1949ddbfd147afd4d964a9f00b24eb291e0e7c38"
```
mark-rushakoff: This revision went backwards.

Gopkg.lock
```diff
@@ -154,12 +162,14 @@
...
+  version = "v1.0.2"
 
 [[projects]]
   name = "github.com/paulbellamy/ratecounter"
   packages = ["."]
-  revision = "5a11f585a31379765c190c033b6ad39956584447"
+  revision = "524851a93235ac051e3540563ed7909357fe24ab"
```
mark-rushakoff: This revision went backwards but we only lost a test coverage commit so it shouldn't matter.

Gopkg.lock
```diff
@@ -204,12 +214,14 @@
 [[projects]]
   name = "github.com/retailnext/hllpp"
   packages = ["."]
-  revision = "38a7bb71b483e855d35010808143beaf05b67f9d"
+  revision = "9fdfea05b3e55bebe7beb22d16c7db15d46cd518"
```
mark-rushakoff: This revision went backwards.

Gopkg.lock
jsternberg: I'm going to switch the others to follow the master branch and keep this one the same. I would rather move us to using versions if possible.

PR #9600, opened 2018-03-19
Updated all dependencies

jsternberg:
Locked the prometheus client to a specific revision because we use a
revision more recent than the latest released version.


Gopkg.lock
```diff
@@ -147,7 +159,8 @@
 [[projects]]
   name = "github.com/matttproud/golang_protobuf_extensions"
   packages = ["pbutil"]
-  revision = "c12348ce28de40eed0136aa2b644d0ee0650e56c"
+  revision = "3247c84500bff8d9fb6d579d800f20b3e091582c"
```
mark-rushakoff: This one went backwards but we only lost an upstream test change.

PR #9609, opened 2018-03-20
Add capability change compaction planner

jwilder:
This adds the capability to use a different compaction planner programatically.

### DIFF
--- a/tsdb/engine/tsm1/compact.go
+++ b/tsdb/engine/tsm1/compact.go
@@ -83,6 +83,8 @@ type CompactionPlanner interface {
 	// ForceFull causes the planner to return a full compaction plan the next
 	// time Plan() is called if there are files that could be compacted.
 	ForceFull()
+
+	SetFileStore(fs *FileStore)
 }
 
 // DefaultPlanner implements CompactionPlanner using a strategy to roll up
@@ -176,6 +178,10 @@ func (t *tsmGeneration) hasTombstones() bool {
 		}
 	}
 	return false
+}
+
+func (c *DefaultPlanner) SetFileStore(fs *FileStore) {
+	c.FileStore = fs
 }
 
 // FullyCompacted returns true if the shard is fully compacted.
--- a/tsdb/engine.go
+++ b/tsdb/engine.go
@@ -153,6 +153,7 @@ type EngineOptions struct {
 	ShardID       uint64
 	InmemIndex    interface{} // shared in-memory index
 
+	CompactionPlannerCreator    CompactionPlannerCreator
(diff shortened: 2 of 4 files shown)


tsdb/engine/tsm1/compact.go
```diff
@@ -1279,21 +1300,26 @@ func (a blocks) Swap(i, j int) { a[i], a[j] = a[j], a[i] }
...
 // size indicates the maximum number of values to encode in a single block.
-func NewTSMKeyIterator(size int, fast bool, interrupt chan struct{}, readers ...*TSMReader) (KeyIterator, error) {
+func NewTSMKeyIterator(size int, fast bool, filter tsdb.CompactionFilter, interrupt chan struct{}, readers ...*TSMReader) (KeyIterator, error) {
 	var iter []*BlockIterator
 	for _, r := range readers {
 		iter = append(iter, r.BlockIterator())
 	}
 
+	if filter == nil {
+		filter = &noFilter{}
```
stuartcarnie: Can you make this a static var, so it doesn't incur any allocations

```go
var defaultFilter tsdb.CompactionFilter = &noFilter{}
```

tsdb/engine/tsm1/compact.go
```diff
@@ -661,6 +661,19 @@ func (c *DefaultPlanner) Release(groups []CompactionGroup) {
...
 
+// noFilter is a CompactionFilter that does not filter anything
+type noFilter struct {
+}
+
+func (f *noFilter) FilterBlock(key []byte, minTime, maxTime int64) bool {
+	return false
+}
+
+func (f *noFilter) FilterTimeRange(key []byte) (int64, int64) {
```
stuartcarnie: Is it worth moving `tsm1.TimeRange` to the `tsdb` package and returning that value here. By doing so, we could add a concept like `IsZero() bool` or `IsEmpty() bool` that can be used elsewhere, such as:

```go

// Remove any values filtered globally
filterRange := k.compactionFilter.FilterTimeRange(k.blocks[i].key)
if !filterRange.IsZero() {
    v = IntegerValues(v).Exclude(filterRange.Min, filterRange.Max)
    // or better still:
    v = IntegerValues(v).Exclude(filterRange)
}
```

jwilder: Maybe. I'm going to hold off of that change for now.

stuartcarnie: I am happy to take than on later 👍

jwilder:
I removed the filtering part of this PR.  That can lead to problems where the index is inconsistent w/ TSM data.  I've left the hooks to change the compaction planner though.

PR #9614, opened 2018-03-21
1.5: Revert "Use MADV_WILLNEED when loading TSM files"

benbjohnson:
This reverts commit ee270e1dd2135d91dbaf0b3adb25d9c2801b6639.


shilicqupt:
may i ask the reason for remove the madviseWillNeed(m.b) reason, indirectIndex warm-up is no better

PR #9622, opened 2018-03-22
Add multi-tenant support to storage RPC API

stuartcarnie:
This change adds additional fields to the `Read` RPC API to execute queries over series keys formatted as `<orgid>\x00\x00<bucketid>,_measurement=<measurement>,…`


stuartcarnie:
@nathanielc once this merges, you will be able to update `IFQL`. I've also extended [ingen](https://github.com/influxdata/ingen) to write either single or multi-tenant series data.

services/storage/predicate_influxql.go
```diff
@@ -7,10 +7,14 @@ import (
 	"github.com/influxdata/influxql"
 )
 
+var (
+	measurementRemap = map[string]string{"_measurement": "_name"}
```
e-dard: Nit: bro, why `var ( )` when you `var`?  😛

cmd/store/query/query.go
```diff
@@ -31,6 +31,7 @@ type Command struct {
 	addr            string
 	cpuProfile      string
 	memProfile      string
+	tenant          string
```
e-dard: This should be `org` now.

PR #9629, opened 2018-03-23
Do not overwrite _measurement tag for multi-tenant reads

stuartcarnie:


services/storage/series_cursor.go
```diff
@@ -181,12 +183,16 @@ RETRY:
...
-		c.filterset = mapValuer{"_name": string(sr.Name)}
+		c.filterset = make(mapValuer)
 		for _, tag := range c.tags {
 			c.filterset[string(tag.Key)] = string(tag.Value)
 		}
 
-		c.tags.Set(measurementKey, sr.Name)
+		if !c.multiTenant {
+			c.filterset["_name"] = string(sr.Name)
+			c.tags.Set(measurementKey, sr.Name)
```
nathanielc: If it is multiTenant is the mesurementKey already part of the tags?

stuartcarnie: exactly – which means less allocations and more efficient 🎉

stuartcarnie: …and now we just need to get `_field` in there

nathanielc: Makes sense

PR #9641, opened 2018-03-27
Add some unit tests to subqueries

jsternberg:
This is not complete, but it is a starting point for more thorough tests
of subqueries.

This also reorders the use of `cmp.Diff` so the `want` is first and
`got` is second. This way, the `want` shows up as a minus sign in the
diff rather than, confusingly, as a plus sign.


jsternberg:
I also fixed the subquery regression from the math refactor so I've pushed that as a separate commit with this one.

PR #9647, opened 2018-03-28
Delete deleted shards in retention service.

benbjohnson:
## Overview

Previously the `retension.Service` would only delete expired shards but ignore already deleted shards in the local filesystem. If a shard directory was not deleted for some reason then this would cause the shard to never get checked and removed later.


kevinhogan111:
thanks @benbjohnson...

@rbetts @gshif  will this be part of a 1.5.2 release?  

We'd like to get this fix over to PayPal as soon as possible

PR #9649, opened 2018-03-28
Allow math functions to be used in the condition

jsternberg:


jsternberg:
I don't think I have, but I just ran a quick test on tsi1 and it works there.

PR #9670, opened 2018-04-02
Add adjustable TSI log file size.

benbjohnson:
## Overview

This commit adds the `max-index-log-file-size` configuration flag so that users can restrict the maximum size of log files before compaction. The default limit was also lowered from `5MB` to `1MB`. The original size was set before we partitioned the index so the change reflects this.

### DIFF
--- a/tsdb/index/tsi1/index.go
+++ b/tsdb/index/tsi1/index.go
@@ -39,8 +39,8 @@ func init() {
 		DefaultPartitionN = uint64(i)
 	}
 
-	tsdb.RegisterIndex(IndexName, func(_ uint64, db, path string, _ *tsdb.SeriesIDSet, sfile *tsdb.SeriesFile, _ tsdb.EngineOptions) tsdb.Index {
-		idx := NewIndex(sfile, db, WithPath(path))
+	tsdb.RegisterIndex(IndexName, func(_ uint64, db, path string, _ *tsdb.SeriesIDSet, sfile *tsdb.SeriesFile, opt tsdb.EngineOptions) tsdb.Index {
+		idx := NewIndex(sfile, db, WithPath(path), WithMaximumLogFileSize(int64(opt.Config.MaxIndexLogFileSize)))
 		return idx
 	})
 }
@@ -111,7 +111,7 @@ type Index struct {
 // NewIndex returns a new instance of Index.
 func NewIndex(sfile *tsdb.SeriesFile, database string, options ...IndexOption) *Index {
 	idx := &Index{
-		maxLogFileSize: DefaultMaxLogFileSize,
+		maxLogFileSize: tsdb.DefaultMaxIndexLogFileSize,
 		logger:         zap.NewNop(),
 		version:        Version,
 		sfile:          sfile,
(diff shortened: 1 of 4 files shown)


tsdb/index/tsi1/index.go
```diff
@@ -111,7 +111,7 @@ type Index struct {
 // NewIndex returns a new instance of Index.
 func NewIndex(sfile *tsdb.SeriesFile, database string, options ...IndexOption) *Index {
 	idx := &Index{
-		maxLogFileSize: DefaultMaxLogFileSize,
+		maxLogFileSize: tsdb.DefaultMaxIndexLogFileSize,
```
e-dard: Shouldn’t this be `tsdb.MaxIndexLogFileSize`?

tsdb/index/tsi1/partition.go
```diff
@@ -97,7 +94,7 @@ func NewPartition(sfile *tsdb.SeriesFile, path string) *Partition {
 		seriesIDSet: tsdb.NewSeriesIDSet(),
 
 		// Default compaction thresholds.
-		MaxLogFileSize: DefaultMaxLogFileSize,
+		MaxLogFileSize: tsdb.DefaultMaxIndexLogFileSize,
```
e-dard: Shouldn’t this be `tsdb.MaxIndexLogFileSize`?

tsdb/index/tsi1/index.go
```diff
@@ -39,8 +39,8 @@ func init() {
 		DefaultPartitionN = uint64(i)
 	}
 
-	tsdb.RegisterIndex(IndexName, func(_ uint64, db, path string, _ *tsdb.SeriesIDSet, sfile *tsdb.SeriesFile, _ tsdb.EngineOptions) tsdb.Index {
-		idx := NewIndex(sfile, db, WithPath(path))
+	tsdb.RegisterIndex(IndexName, func(_ uint64, db, path string, _ *tsdb.SeriesIDSet, sfile *tsdb.SeriesFile, opt tsdb.EngineOptions) tsdb.Index {
+		idx := NewIndex(sfile, db, WithPath(path), WithMaximumLogFileSize(int64(opt.Config.MaxIndexLogFileSize)))
```
e-dard: Was `WithMaximumLogFileSize` added in another PR? I can’t see it in this change set (but I am on my phone).

tsdb/index/tsi1/partition.go
benbjohnson: `tsdb.MaxIndexLogFileSize` would indicate that there is a fixed setting. It's using the default value when initialized and then the `Index` overrides the value with its value here: https://github.com/influxdata/influxdb/blob/dc71a8d82bf94ce4ed62818760d6a28b1abf8996/tsdb/index/tsi1/index.go#L180

tsdb/index/tsi1/index.go
benbjohnson: Yes, it was there previously but only applied from the `Index` down to the `Partition`.

tsdb/index/tsi1/partition.go
e-dard: Ah I see. Because I could see the `With` overrider in the chageset I missed that.

e-dard: Ah I see. Because I could see the `With` overrider in the chageset I missed that.

PR #9682, opened 2018-04-04
Avoid a panic when using show diagnostics with text/csv

### LINKED ISSUE #9504: Panic when running SHOW DIAGNOSTICS with text/csv
### Bug report

__System info:__ InfluxDB 1.5 and likely older

__Steps to reproduce:__

```
$ curl -G 'http://localhost:8086/query' -H 'Accept: text/csv' --data-urlencode 'q=SHOW DIAGNOSTICS'
```

__Expected behavior:__ It likely shouldn't panic.

__Actual behavior:__ It does panic.

__Additional info:__ This is likely because `SHOW DIAGNOSTICS` is weird. It is the only command that changes which columns exist in the row within the same statement, which is pretty weird. Most queries use the same columns for the entire statement, but this command doesn't.

jsternberg:
If the columns change between series, it will now act as if it was a new
statement id and reprint the headers. This only happens with show
diagnostics at the moment and we shouldn't add this functionality
anywhere else anyway.

Fixes #9504.

### DIFF
--- a/services/httpd/response_writer.go
+++ b/services/httpd/response_writer.go
@@ -147,7 +147,29 @@ func (w *csvFormatter) WriteResponse(resp Response) (n int, err error) {
 			}
 		}
 
-		for _, row := range result.Series {
+		for i, row := range result.Series {
+			if i > 0 && !stringsEqual(result.Series[i-1].Columns, row.Columns) {
+				// The columns have changed. Print a newline and reprint the header.
+				csv.Flush()
+				if err := csv.Error(); err != nil {
+					return n, err
+				}
+
+				out, err := io.WriteString(w, "\n")
+				if err != nil {
+					return n, err
+				}
+				n += out
+
+				w.columns = make([]string, 2+len(row.Columns))
+				w.columns[0] = "name"
+				w.columns[1] = "tags"
+				copy(w.columns[2:], row.Columns)
+				if err := csv.Write(w.columns); err != nil {
+					return n, err
+				}
+			}
+
 			w.columns[0] = row.Name
 			if len(row.Tags) > 0 {
 				w.columns[1] = string(models.NewTags(row.Tags).HashKey()[1:])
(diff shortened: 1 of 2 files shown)


services/httpd/response_writer.go
```diff
@@ -147,7 +148,29 @@ func (w *csvFormatter) WriteResponse(resp Response) (n int, err error) {
 			}
 		}
 
-		for _, row := range result.Series {
+		for i, row := range result.Series {
+			if i > 0 && !reflect.DeepEqual(result.Series[i-1].Columns, row.Columns) {
```
mark-rushakoff: Since these are both string slices, can you make a helper method to compare them instead of using reflect?

services/httpd/response_writer.go
```diff
@@ -147,7 +148,29 @@ func (w *csvFormatter) WriteResponse(resp Response) (n int, err error) {
...
+					return n, err
+				}
+				n += out
+
+				w.columns = make([]string, 2+len(row.Columns))
+				w.columns[0] = "name"
+				w.columns[1] = "tags"
+				copy(w.columns[2:], row.Columns)
+				if err := csv.Write(w.columns); err != nil {
+					return n, err
```
mark-rushakoff: The `n` tracking is missing most of the csv calls - it's unfortunate that the csv.Writer doesn't expose the bytes written.

Assuming we do want to track `n` (for the QueryRequestBytesTransmitted stat), what do you think about using io.MultiWriter with the underlying `w` and some fake writer like

```go
type countWriter struct { n int64 }
func (w *countWriter) Write(p []byte) (int, err) {
  w.n += len(p)
  return len(p), nil
}
```

so that we can properly track `n`?

jsternberg: That would be good. Can we open up another PR for that though? The behavior is already broken and I'm not breaking it anymore than it previously was.

services/httpd/response_writer.go
jsternberg: Sure.

services/httpd/response_writer.go
mark-rushakoff: Sure thing - opened https://github.com/influxdata/influxdb/issues/9690 to track it.

PR #9686, opened 2018-04-05
Adding additional aggregate benchmarks for tsm1

jsternberg:
This will help us address performance problems in the underlying tsm1
implementations of the aggregate iterators.

### DIFF
--- a/tsdb/engine/tsm1/engine_test.go
+++ b/tsdb/engine/tsm1/engine_test.go
@@ -1675,75 +1675,6 @@ func TestEngine_DisableEnableCompactions_Concurrent(t *testing.T) {
 	}
 }
 
-func BenchmarkEngine_CreateIterator_Count_1K(b *testing.B) {
-	benchmarkEngineCreateIteratorCount(b, 1000)
-}
-func BenchmarkEngine_CreateIterator_Count_100K(b *testing.B) {
-	benchmarkEngineCreateIteratorCount(b, 100000)
-}
-func BenchmarkEngine_CreateIterator_Count_1M(b *testing.B) {
-	benchmarkEngineCreateIteratorCount(b, 1000000)
-}
-
-func benchmarkEngineCreateIteratorCount(b *testing.B, pointN int) {
-	benchmarkIterator(b, query.IteratorOptions{
-		Expr:      influxql.MustParseExpr("count(value)"),
-		Ascending: true,
-		StartTime: influxql.MinTime,
-		EndTime:   influxql.MaxTime,
-	}, pointN)
-}
-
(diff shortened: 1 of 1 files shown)


hercules-influx:
```/tmp/587897810/src/github.com/influxdata/influxdb/tsdb/index.go:765:4: should use a simple channel send/receive instead of select with a single case (S1000)
/tmp/587897810/src/github.com/influxdata/influxdb/tsdb/engine/tsm1/engine_test.go:1753:6: func benchmarkIterator is unused (U1000)
```

tsdb/engine/tsm1/engine_test.go
```diff
@@ -1819,28 +1750,168 @@ func BenchmarkEngine_WritePoints_Parallel(b *testing.B) {
...
+				Duration: time.Hour,
+			}
+			return opt
+		},
+	},
 }
 
-func benchmarkIterator(b *testing.B, opt query.IteratorOptions, pointN int) {
+func BenchmarkEngine_CreateIterator_Default_1K(b *testing.B) {
+	benchmarkIteratorDefault(b, 1000)
```
e-dard: Why not something like:

```go
func BenchmarkEngine_CreateIterator_Default(b *testing.B) {
    sizes := []int{1000, 100000, 1000000}
    for _, size := range sizes {
		e := MustInitDefaultBenchmarkEngine(size)
		b.Run(fmt.Sprint(size), func(b *testing.B) {
			benchmarkEngine(b, e)
		})
	}
}
```

That way you don't need separate `BenchmarkEngine_CreateIterator_Default_X` functions, and you get nicer output in the results. You will get everything namespaced by `/size/`

tsdb/engine/tsm1/engine_test.go
```diff
@@ -1819,28 +1750,168 @@ func BenchmarkEngine_WritePoints_Parallel(b *testing.B) {
...
+			name := tt.name + "_" + variant.name
+			opt := variant.modify(tt.opt)
+			b.Run(name, func(b *testing.B) {
+				b.ReportAllocs()
+				for i := 0; i < b.N; i++ {
+					itr, err := e.CreateIterator(context.Background(), "cpu", opt)
+					if err != nil {
+						b.Fatal(err)
+					}
+					query.DrainIterator(itr)
```
e-dard: Does `DrainIterator` have a significant cost or any allocations? You could surround it in a `stop` / `reset` or whatever if it does.

jsternberg: It reads the iterator so all of the logic we're testing happens because that's initiating the calls to `Next()`. So that needs to be part of the profile.

Issue #9421, opened 2018-02-09
Add CA certificate, CRL and HTTPS client certificate verification support for InfluxDB


mjf:
Hello,

it would be nice to add *CA* certificate, *CRL* and *HTTPS* **client certificate verification** support for *InfluxDB*. As far as I know *HAProxy* or another proxy is needed in front of *InfluxDB* to achieve client certificate verification for now. Desired configuration could look like the following example one:

    [http]
        # ... snip ...
        https-enabled = true

        # ... snip ...
        https-certificate = "/path/to/some.crt"
        https-key = "/path/to/some.key"

        # NEW OPTIONS
        https-ca-certificate = "/path/to/some.crt"
        # https-crl = "/path/to/some.crl"
        https-client-certificate-verify = true
        https-client-certificate-verify-required = true

-----
Snippet from the *HAProxy* v1.8 documentation on how this works in it:

``verify [ none | optional | required ]``

*[...] If set to 'none', client certificate is not requested. In other cases, a client certificate is requested.
If the client does not provide a certificate after the request and if 'verify' is set to 'required', then the
handshake is aborted, while it would have succeeded if set to 'optional'. The certificate provided by the
client is always verified using CAs from 'ca-file' and optional CRLs from 'crl-file'. [...]*

-----
(also see [HAProxy Configuration Manual](https://cbonte.github.io/haproxy-dconv/1.8/configuration.html#5.1-verify))

FendiJatmiko:
anyone currently working on this ..? i would like to working on this issue, please. .

mahekchheda:
@mgaffney is this issue still needed? I would like to start contributing to this project so don't mind picking this up.

mgaffney:
@mahekchheda I'm not sure what the status of this is.

OmgImAlexis:
Can this please be reopened?

PR #9698, opened 2018-04-10
Add CA certificate and HTTPS client certificate verification support for InfluxDB

gentstr:
#9421

SSL/TLS client certificate authentication allows using a certificate and key to authenticate instead of using a username password. The primary use case is to remove clear-text passwords from configuration files (e.g. Telegraf).

This implementation validates client certificates against the CA certificate provided to InfluxDB through a new configuration option. If the client certificate is valid, the Common Name (CN) is extracted and compared against valid usernames. This is a model implemented with common cert authentication in postgresql, mysql, rabbitmq, etc.

Note: There is no support for CRLs. It is not implemented in the crypto/tls package and the creator of that package has publicly stated his reasons why https://news.ycombinator.com/item?id=7615273. Client certificates are no longer valid if the username in the Common Name is removed.

* New configuration option https-ca-certificate
* Optional client certificate, if provided, is validated against CA certificate
* Username is extracted from Common Name (CN)

- [x] Config changes: update sample config (`etc/config.sample.toml`), server `NewConfig` method, and `Diagnostics` methods reporting config settings, if necessary
- [x] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): pull request submitted influxdata/docs.influxdata.com#1529


services/httpd/config.go
```diff
@@ -28,6 +28,7 @@ type Config struct {
 	WriteTracing          bool          `toml:"write-tracing"`
 	PprofEnabled          bool          `toml:"pprof-enabled"`
 	HTTPSEnabled          bool          `toml:"https-enabled"`
+	HTTPSCACertificate string `toml:"https-ca-certificate"`
```
jsternberg: You need to use `go fmt` to format the files. I also don't think this will work for configuration because it needs to be made abundantly clear this is an alternative for authentication. It may be interpreted as a normal client certificate (which should also be supported) to validate a client, but not as authentication and it may get interpreted that way by a casual reading of the options.

So there will need to be one configuration for validating a client certificate to begin with and one that enables reading the CN of the client certificate to perform authentication.

I'm not sure what that looks like right now so just format the file for now.

services/httpd/handler.go
```diff
@@ -1356,6 +1359,14 @@ func parseCredentials(r *http.Request) (*credentials, error) {
 		}
 	}
 
+	// Check for TLS client certificate
```
jsternberg: The comment style for this seems to be to use a period at the end.

jsternberg:
Before you put too much work in this, I'm going to kick this up to be reviewed by our product team to make sure that this is a feature we want to merge.

jsternberg:
Thanks for the heads up. I'll try to get this on the radar for things we need to look at. Until then, I've made a comment on the docs issue saying we need to revert this.

services/httpd/config.go
mark-rushakoff: I agree - there should be separate config settings for:

- the certificate to use for verification
- whether client verification is disabled, on but optional, or on and required
- maybe whether to use CN or some other attribute for the username? I didn't find any quick answers when I searched for client certificate usernames.

kaspergrubbe:
@jsternberg Did you hear back from the product team?

kaspergrubbe:
@gentstr Thanks for your contribution, this is a welcomed change so we don't need to run Influxdb behind Haproxy. Have you thought about how the `influx` command will work with this change, should it be possible to specify which client cert/key-pair to use for the connection?

forsberg:
Minor detail: The http request log does not show any username when authenticated via client certificate, would be nice if it listed the CN of the certificate in use.

e-dard:
@gentstr if you were still interested in pushing for this to get into a future 1.x release could you fix up the existing review comments? I realise the PR is pretty stale now so if I don't hear back I can just close it out. Thanks!

leonboot:
What a pity this PR didn't make the grade. It's exactly what I'm looking for. In fact, I ended up here because I was searching for user certificate authentication for InfluxDB.

Are the main arguments against merging this PR the lack of tests, as [mentioned](pullrequestreview-126521251) by @mark-rushakoff?

Does it make sense to have a setting for certificate verification in this case? And is there any other field suitable for providing the username besides the CN?

If @gentstr isn't able to provide the appropriate tests for this feature, couldn't the PR be adopted by InfluxData?

mark-rushakoff:
Tests are one part of it, but adding a new way to log in or authenticate is a weighty enough feature on its own. Combined with all the subtleties around TLS configuration and verification, we need very careful consideration on how this interacts with the rest of InfluxDB holistically.

To be honest, I don't recall any feature of comparable size that was introduced by a community PR -- normally the design originates internally from InfluxData.

Maybe it would be best to start a discussion on https://community.influxdata.com, for more general visibility from other community members and from InfluxData's product team, to see how to move this feature forward.

Forbzy:
I'm using influxdb 2.0.4 and I couldn't see an easy way for the client to use TLS/SSL.  @mark-rushakoff I see that this issue was raised in 2018. Are we any closer to having this?

PR #9705, opened 2018-04-11
Check for errors from binary.Uvarint when reading TSI logs

### LINKED ISSUE #9425: TSI: log entry checksum should be validated before decoding entry
Garbage input to (*LogEntry).UnmarshalBinary can cause panics in some instances (e.g. if binary.Uvarint returns a negative length [here](https://github.com/influxdata/influxdb/blob/90c921fc0f7f0186955641fcf4b4c9e256b3696d/tsdb/index/tsi1/log_file.go#L1038-L1039)).

Assuming these panics come from corrupted input files, the CRC validation should happen at the start of UnmarshalBinary rather than validated at the end.

jacobmarble:
Closes #9425

This change doesn't do anything with the CRC check because the CRC sum isn't available until after the panic likely occurs. This change does check the result of `binary.Uvarint` which is the likely source of panic.

### DIFF
--- a/tsdb/index/tsi1/log_file.go
+++ b/tsdb/index/tsi1/log_file.go
@@ -1040,7 +1040,7 @@ func (e *LogEntry) UnmarshalBinary(data []byte) error {
 	// Parse series id.
 	if len(data) < 1 {
 		return io.ErrShortBuffer
-	} else if seriesID, n = binary.Uvarint(data); n == 0 {
+	} else if seriesID, n = binary.Uvarint(data); n == 0 || n > len(data) {
 		return io.ErrShortBuffer
 	} else if n < 0 {
 		return fmt.Errorf("parsing binary-encoded uint64 value failed; binary.Uvarint() returned %d", n)
@@ -1050,7 +1050,7 @@ func (e *LogEntry) UnmarshalBinary(data []byte) error {
 	// Parse name length.
 	if len(data) < 1 {
 		return io.ErrShortBuffer
-	} else if sz, n = binary.Uvarint(data); n == 0 {
+	} else if sz, n = binary.Uvarint(data); n == 0 || n > len(data) {
 		return io.ErrShortBuffer
 	} else if n < 0 {
 		return fmt.Errorf("parsing binary-encoded uint64 value failed; binary.Uvarint() returned %d", n)
@@ -1065,7 +1065,7 @@ func (e *LogEntry) UnmarshalBinary(data []byte) error {
 	// Parse key length.
 	if len(data) < 1 {
 		return io.ErrShortBuffer
-	} else if sz, n = binary.Uvarint(data); n == 0 {
+	} else if sz, n = binary.Uvarint(data); n == 0 || n > len(data) {
 		return io.ErrShortBuffer
 	} else if n < 0 {
 		return fmt.Errorf("parsing binary-encoded uint64 value failed; binary.Uvarint() returned %d", n)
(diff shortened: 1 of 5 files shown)


tsdb/index/tsi1/log_file.go
```diff
@@ -1039,15 +1040,20 @@ func (e *LogEntry) UnmarshalBinary(data []byte) error {
 	// Parse series id.
 	if len(data) < 1 {
 		return io.ErrShortBuffer
+	} else if seriesID, n = binary.Uvarint(data); n == 0 {
+		return io.ErrShortBuffer
+	} else if n < 0 {
+		return fmt.Errorf("parsing binary-encoded uint64 value failed; binary.Uvarint() returned %d", n)
 	}
-	seriesID, n := binary.Uvarint(data)
 	e.SeriesID, data = uint64(seriesID), data[n:]
```
mark-rushakoff: Still need to check `n < len(data)`, or we could get an index out of range panic on this line.

jacobmarble: Done

jacobmarble:
@mark-rushakoff PTAL

e-dard:
It seems to me that we could DRY this up quite a bit. How about, inside of `UnmarshalBinary` we have: 

```go
uvarint := func(data []byte) (v int, n int, err error) {
    if len(data) < 1 {
        return 0, 0, io.ErrShortBuffer
    } else if v, n = binary.Uvarint(data); n == 0 || n > len(data) {
        return 0, 0, io.ErrShortBuffer
    } else if n < 0 {
        return 0, 0, fmt.Errorf("parsing binary-encoded uint64 value failed; binary.Uvarint() returned %d", n)
     }
    return v, n, err
}


// Parse series id.
v, n, err := uvarint(data)
if err != nil {
    return err
}

...

// Parse key length.
if v, n, err = uvarint(data); err != nil {
    return err
}

...

// Parse value length.
if v, n, err = uvarint(data); err != nil {
    return err
}

...
```

jacobmarble:
SGTM

tsdb/index/tsi1/log_file.go
```diff
@@ -1024,8 +1024,22 @@ type LogEntry struct {
...
+	uvarint := func(data []byte) (value uint64, n int, err error) {
+		if len(data) < 1 {
+			err = io.ErrShortBuffer
+		} else if value, n = binary.Uvarint(data); n == 0 || n > len(data) {
+			err = io.ErrShortBuffer
+		} else if n < 0 {
+			err = fmt.Errorf("parsing binary-encoded uint64 value failed; binary.Uvarint() returned %d", n)
+		}
+		return
+	}
```
mark-rushakoff: `uvarint` can be promoted to a package-level function since it doesn't (and shouldn't) hold a closure around anything scoped to `(*LogEntry).UnmarshalBinary`.

jacobmarble: I was going to say that too, but this package is relatively large, so privatizing the function inside `UnmarshalBinary` is at least more readable.

OTOH, `uvarint` is useful anywhere that `binary.Uvarint` is called. There are 10 more uses of `binary.Uvarint` just within the `tsi1` package, and none of them check `n`.

jacobmarble: PTAL. I left 4 other uses within `tsi1`, they are nested at least two-deep in non-error-returning functions.

Issue #8677, opened 2017-08-09
[BUG] Backup stopped working after upgrade to 1.3.2


skladd:
After yesterday's upgrade to version 1.3.2 (from 1.3.1 via Debian repo), backup fails with this error message:

`influxd backup -database telegraf /tmp/influxbackup`

```
2017/08/09 10:05:21 backing up db=telegraf since 0001-01-01 00:00:00 +0000 UTC
2017/08/09 10:05:21 backing up metastore to /tmp/influxbackup/meta.01
2017/08/09 10:05:21 backing up db=telegraf rp=autogen shard=169 to /tmp/influxbackup/telegraf.autogen.00169.00 since 0001-01-01 00:00:00 +0000 UTC
2017/08/09 10:05:21 Download shard 169 failed copy backup to file: err=<nil>, n=0.  Retrying (0)...
2017/08/09 10:05:22 Download shard 169 failed copy backup to file: err=<nil>, n=0.  Retrying (1)...
2017/08/09 10:05:23 Download shard 169 failed copy backup to file: err=<nil>, n=0.  Retrying (2)...
2017/08/09 10:05:24 Download shard 169 failed copy backup to file: err=<nil>, n=0.  Retrying (3)...
2017/08/09 10:05:25 Download shard 169 failed copy backup to file: err=<nil>, n=0.  Retrying (4)...
2017/08/09 10:05:26 Download shard 169 failed copy backup to file: err=<nil>, n=0.  Retrying (5)...
2017/08/09 10:05:27 Download shard 169 failed copy backup to file: err=<nil>, n=0.  Retrying (6)...
2017/08/09 10:05:28 Download shard 169 failed copy backup to file: err=<nil>, n=0.  Retrying (7)...
2017/08/09 10:05:29 Download shard 169 failed copy backup to file: err=<nil>, n=0.  Retrying (8)...
2017/08/09 10:05:30 Download shard 169 failed copy backup to file: err=<nil>, n=0.  Retrying (9)...
2017/08/09 10:05:31 backup failed: copy backup to file: err=<nil>, n=0
backup: copy backup to file: err=<nil>, n=0
```
Version: 1.3.2
OS: Debian jessie on x86_64

skladd:
On another server, where influxdb was installed much later (thus a more recent version), backup still works after the upgrade.

skladd:
[...]

PR #9706, opened 2018-04-12
Changed file download retry rate to have exponential wait times

aanthony1243:
helps reduce failures mentioned in #8677

### DIFF
--- a/cmd/influxd/backup/backup.go
+++ b/cmd/influxd/backup/backup.go
@@ -19,6 +19,7 @@ import (
 	"github.com/influxdata/influxdb/cmd/influxd/backup_util"
 	"github.com/influxdata/influxdb/services/snapshotter"
 	"github.com/influxdata/influxdb/tcp"
+	"math"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


cmd/influxd/backup/backup.go
```diff
@@ -541,8 +542,13 @@ func (cmd *Command) download(req *snapshotter.Request, path string) error {
 		}(); err == nil {
 			break
 		} else if err != nil {
-			cmd.StderrLogger.Printf("Download shard %v failed %s.  Retrying (%d)...\n", req.ShardID, err, i)
-			time.Sleep(time.Second)
+			min := 2 * time.Second
```
e-dard: Can you move this out of the loop. Reinitialising it ten times is bothering me 😛

cmd/influxd/backup/backup.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/influxdata/influxdb/cmd/influxd/backup_util"
 	"github.com/influxdata/influxdb/services/snapshotter"
 	"github.com/influxdata/influxdb/tcp"
+	"math"
```
e-dard: This import doesn't live here; it should be between `log` and `os`.

PR #9710, opened 2018-04-12
Update client README - close client resources

jangaraj:
The best practice is to close used resources at the end. I've had a problem when I used this code in AWS Golang Lambda function without `c.Close()`: `socket: too many open files` - I've reached connection limit because InfluxDB connections have been still opened.

- [X] Provide example syntax

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -72,7 +72,8 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-
+	defer c.Close()
+	
 	// Create a new point batch
 	bp, err := client.NewBatchPoints(client.BatchPointsConfig{
 		Database:  MyDB,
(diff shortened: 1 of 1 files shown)


client/README.md
```diff
@@ -100,6 +100,9 @@ func main() {
 	if err := c.Write(bp); err != nil {
 		log.Fatal(err)
 	}
+	
+	// Close client resources
+	c.Close()
```
jsternberg: I think this code would be better. On line 75, add this:

```golang
defer c.Close()
```

This ensures that close is always called.

At the end, instead, do this:

```golang
if err := c.Close(); err != nil {
    log.Fatal(err)
}
```

Looking at the code, this can't fail, but it's a good practice and the Close method should be capable of being called multiple times. In case we ever needed to modify the code to report some kind of failure (unlikely), it's good to be able to handle it.

jsternberg:
@jangaraj can you squash the commits for this?

jangaraj:
@jsternberg done

Issue #9554, opened 2018-03-12
Flaky Test


e-dard:
During CI checks for #9551 `TestService_ModeANY` kept failing in Circle. I was unable to make it fail locally. The output looks like:

```
--- FAIL: TestService_ModeANY (0.02s)
	service_test.go:254: expected points request```

I have skipped the test in #9551

**Comment by jacobmarble** [comment]
I'll start by writing a script that just runs the test lots of times, record a baseline failure rate.

**Comment by jacobmarble** [comment]
This command builds the test, then runs it 1000000 times, completes in 66 seconds:
```
go test -run TestService_ModeANY -count 1000000
```

It passes on my workstation.

jacobmarble:
Playing with CircleCI, see if I can reduce latency to complete. Currently ~18m, but only ~5m on my laptop after the first run, because the docker image doesn't have to be rebuilt. Multiple improvements are available.

jacobmarble:
Test has been re-enabled, will watch for failure over the next few days.

PR #9711, opened 2018-04-12
update CircleCI config to 2.0 syntax

jacobmarble:
Expect flakiness described in #9554.

Updating CircleCI config to 2.0 is required by aug-2018, and it gives us access to some fancy CI features.


jacobmarble:
The first couple of executions with this new config show a small performance improvement (~2m) just from this change. Might be that 1.0 was running in a docker container, while I have the 2.0 config running on a VM, which has 2 cores. Docker Layer Caching would improve performance more dramatically, perhaps by ~8m.

jacobmarble:
@e-dard I should note that this PR removes a lot of logic, for example there is no more `circle-test.sh`, the CircleCI config just runs `test.sh` and doesn't persist STDOUT. I think this is appropriate, please confirm.

jacobmarble:
Latest changes improve (cached) test runtime to ~8m.

services/subscriber/service_test.go
```diff
@@ -185,7 +185,6 @@ func TestService_ModeALL(t *testing.T) {
 }
 
 func TestService_ModeANY(t *testing.T) {
-	t.Skip("TODO: flaky test.")
```
e-dard: Is this implication here that you think moving to this new infrastructure will fix this flaky test?

.gitignore
```diff
@@ -66,6 +66,7 @@ config.toml
 
 # test data files
 integration/migration_data/
+test-logs/
```
mark-rushakoff: Is this necessary?

CONTRIBUTING.md
```diff
@@ -277,4 +277,4 @@ func BenchmarkSomething(b *testing.B) {
 
 Continuous Integration testing
 -----
-InfluxDB uses CircleCI for continuous integration testing. To see how the code is built and tested, check out [this file](https://github.com/influxdata/influxdb/blob/master/circle-test.sh). It closely follows the build and test process outlined above. You can see the exact version of Go InfluxDB uses for testing by consulting that file.
+InfluxDB uses CircleCI for continuous integration testing. To see how the code is built and tested, check out [this file](https://github.com/influxdata/influxdb/blob/master/test.sh). It closely follows the build and test process outlined above. You can see the exact version of Go InfluxDB uses for testing by consulting that file.
```
jsternberg: Should we include instructions for how to use the shell script here? I think this section needs to be updated with the recommended testing procedures that are useful to a contributor. I'm not sure I would recommend using any of our CI scripts for testing in all honesty.

mark-rushakoff: The last sentence of this line is false and should be removed.

.gitignore
jacobmarble: `./test.sh 0` writes to that directory

circle.yml
```diff
@@ -1,23 +0,0 @@
-machine:
-    services:
-        - docker
-    environment:
-      GODIST: "go1.9.2.linux-amd64.tar.gz"
```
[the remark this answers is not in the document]
jacobmarble: This has no effect. `./test.sh n` creates a container, installs go (versioned, see `test.sh`) then executes tests.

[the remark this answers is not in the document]
jacobmarble: In other words, this installs go on the host container/VM, then executes the tests in guest containers.

jacobmarble:
@mark-rushakoff PTAL

jacobmarble:
FYI a failing version of this PR, just added `t.Fail()` to one test:
https://github.com/influxdata/influxdb/pull/9717

PR #9260, opened 2017-12-24
Add additional technical analysis algorithms

phemmer:
This adds several technical analysis algorithms:

* [`exponential_moving_average`](http://stockcharts.com/school/doku.php?id=chart_school:technical_indicators:moving_averages#exponential_moving_average_calculation)
* [`double_exponential_moving_average`](https://en.wikipedia.org/wiki/Double_exponential_moving_average)
* [`triple_exponential_moving_average`](https://en.wikipedia.org/wiki/Triple_exponential_moving_average)
* [`relative_strength_index`](http://stockcharts.com/school/doku.php?id=chart_school:technical_indicators:relative_strength_index_rsi)
* [`triple_exponential_average`](http://stockcharts.com/school/doku.php?id=chart_school:technical_indicators:trix)
* [`kaufmans_efficiency_ratio`](http://stockcharts.com/school/doku.php?id=chart_school:technical_indicators:kaufman_s_adaptive_moving_average) (commonly referred to as just "Efficiency Ratio")
* [`kaufmans_adaptive_moving_average`](http://stockcharts.com/school/doku.php?id=chart_school:technical_indicators:kaufman_s_adaptive_moving_average)
* [`chande_momentum_oscillator`](https://www.fidelity.com/learning-center/trading-investing/technical-analysis/technical-indicator-guide/cmo)

The `exponential_moving_average` algorithm also has a PR open at #8827, but this implementation has a few differences:
1. The holdback period is customizable. Defaults to the algorithm's warmup period.
2. Starts the algorithm using a exponential moving average with scaling alpha instead of a simple average. This results in a faster time in which the algorithm can start producing accurate values. Starting the algorithm with a simple average is still offered as a parameter.

The algorithms themselves are pulled in from an external library of mine. I have a personal closed-source application which uses them. I pulled the algorithms out of the app and dumped them into their own repo with MIT license. However I can instead copy them into the InfluxDB source if desired.

Will write documentation & tests if this clears initial review.

### Sample data
```
> select v from test where time >= 0 and time <= 8
name: test
time v
---- -
0    1
1    2
2    3
3    4
4    5
5    4
6    3
7    2
8    1
```

### `exponential_moving_average`:
```
> select last(v),exponential_moving_average(last(v),3) from test where time >= 0 and time <= 8 group by time(1ns)
name: test
time last exponential_moving_average
---- ---- --------------------------
0    1    
1    2    
2    3    2.333333333333333
3    4    3.1666666666666665
4    5    4.083333333333333
[...]

PR #9733, opened 2018-04-17
Add additional technical analysis algorithms

jsternberg:
Adapted from #9260.


query/call_iterator.go
```diff
@@ -1247,6 +1247,32 @@ func newMovingAverageIterator(input Iterator, n int, opt IteratorOptions) (Itera
 	}
 }
 
+// newExponentialMovingAverageIterator returns an iterator for operating on an exponential_moving_average() call.
+func newExponentialMovingAverageIterator(input Iterator, n int, nHold int, warmupType string, opt IteratorOptions) (Iterator, error) {
```
jsternberg: `n, nHold int` you don't need to repeat the argument twice.

The argument for `warmupType` should be the `gota.WarmupType`. The string to warmup type can be somewhere else.

query/compile.go
```diff
@@ -550,6 +552,61 @@ func (c *compiledField) compileMovingAverage(args []influxql.Expr) error {
...
+		case *influxql.IntegerLiteral:
+			if arg2.Val < 0 && arg2.Val != -1 {
+				return fmt.Errorf("exponential_moving_average hold period must be greater than or equal to 0")
+			}
+		default:
+			return fmt.Errorf("exponential_moving_average hold period must be an integer")
+		}
+	}
+
+	if len(args) >= 4 {
```
jsternberg: This isn't possible. The portion above expects the number of arguments to be 3 or less.

query/compile.go
```diff
@@ -550,6 +552,61 @@ func (c *compiledField) compileMovingAverage(args []influxql.Expr) error {
...
+		switch arg3 := args[3].(type) {
+		case *influxql.StringLiteral:
+			switch arg3.String() {
+			case "'exponential_moving_average'":
```
jsternberg: The string here should not have single quotes. That will get removed on its own. I think using a name of `exponential_moving_average` isn't the best. Maybe this should just be `exponential`?

query/functions.go
```diff
@@ -638,6 +639,59 @@ func (r *UnsignedMovingAverageReducer) Emit() []FloatPoint {
...
+	r.aggregate(float64(p.Value), p.Time)
+}
+func (r *ExponentialMovingAverageReducer) aggregate(v float64, t int64) {
+	r.v = r.ema.Add(v)
+	r.t = t
+	r.count++
+}
+func (r *ExponentialMovingAverageReducer) Emit() []FloatPoint {
+	if r.count <= r.holdPeriod {
+		return []FloatPoint(nil)
```
jsternberg: I think you can just return `nil`.

query/select.go
```diff
@@ -288,6 +288,28 @@ func (b *exprIteratorBuilder) buildCallIterator(ctx context.Context, expr *influ
...
+			}
+
+			nHold := -1
+			if len(expr.Args) >= 3 {
+				nHold = int(expr.Args[2].(*influxql.IntegerLiteral).Val)
+			}
+
+			var warmupType string
+			if len(expr.Args) >= 4 {
+				warmupType = expr.Args[3].(*influxql.StringLiteral).String()
```
jsternberg: There's no need to do this. Just do `expr.Args[3].(*influxql.StringLiteral).Val` and you won't have to strip the quotes.

query/compile.go
```diff
@@ -569,6 +569,9 @@ func (c *compiledField) compileExponentialMovingAverage(name string, args []infl
 	if len(args) >= 3 {
 		switch arg2 := args[2].(type) {
 		case *influxql.IntegerLiteral:
+			if name == "triple_exponential_average" && arg2.Val < 1 && arg2.Val != -1 {
+				return fmt.Errorf("%s hold period must be greater than or equal to 1", name)
```
jsternberg: Make some mention here about -1 being the option to set it to the default value. Maybe zero should set it to the default value instead? I'm not really sure I care which one signals the default value as long as one of them does and the error message mentions it.

query/select.go
```diff
@@ -342,6 +342,28 @@ func (b *exprIteratorBuilder) buildCallIterator(ctx context.Context, expr *influ
...
```
jsternberg: Same things I mentioned before apply here too.

jsternberg:
@phemmer above are my comments. If you have time, can you take a look at those and then you can push back to your own branch if you are OK with the changes I made?

query/compile.go
phemmer: Following from your other comments, I'm assuming you mean if `arg3.Val` is used instead. Valid point. I probably just didn't notice `Val` when I was doing the discovery & implementation.

As for the name, yes it's long, but it makes it clear what it is. "exponential" is ambiguous. Most other technical analysis applications just refer to it as "ema", but influxdb doesn't abbreviate anything, and the function name is also "exponential_moving_average", so I kept the long name for consistency.
Perhaps support both? `ema` and `exponential_moving_average`?

jsternberg:
@natezimmer see #9260. This was just a PR for aiding in collaboration on this. #9260 is the real one. These functions should be available in the upcoming 1.6 release.

PR #9743, opened 2018-04-18
Add 'influx_inspect deletetsm'

benbjohnson:
## Overview

Provides a tool for bulk deleting a measurement from raw TSM files.

### DIFF
--- a/cmd/influx_inspect/deletetsm/deletetsm.go
+++ b/cmd/influx_inspect/deletetsm/deletetsm.go
@@ -1,0 +1,153 @@
+// Package deletetsm bulk deletes a measurement from a raw tsm file.
+package deletetsm
+
+import (
+	"flag"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+	"time"
+
+	"github.com/influxdata/influxdb/models"
+	"github.com/influxdata/influxdb/tsdb/engine/tsm1"
+)
+
+// Command represents the program execution for "influxd deletetsm".
+type Command struct {
+	// Standard input/output, overridden for testing.
+	Stderr io.Writer
+	Stdout io.Writer
+
+	measurement string // measurement to delete
+	verbose     bool   // verbose logging
+}
+
+// NewCommand returns a new instance of Command.
+func NewCommand() *Command {
+	return &Command{
+		Stderr: os.Stderr,
+		Stdout: os.Stdout,
+	}
+}
+
+// Run executes the command.
+func (cmd *Command) Run(args ...string) (err error) {
+	fs := flag.NewFlagSet("deletetsm", flag.ExitOnError)
+	fs.StringVar(&cmd.measurement, "measurement", "", "")
+	fs.BoolVar(&cmd.verbose, "v", false, "")
+	fs.SetOutput(cmd.Stdout)
+	fs.Usage = cmd.printUsage
+	if err := fs.Parse(args); err != nil {
+		return err
+	} else if fs.NArg() == 0 {
+		fmt.Printf("path required\n\n")
+		fs.Usage()
+		return nil
+	}
+
+	if !cmd.verbose {
+		log.SetOutput(ioutil.Discard)
+	}
+
+	// Validate measurement flag.
+	if cmd.measurement == "" {
+		return fmt.Errorf("measurement name required")
+	}
+
+	// Process each TSM file.
+	for _, path := range fs.Args() {
+		log.Printf("processing: %s", path)
+		if err := cmd.process(path); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (cmd *Command) process(path string) error {
+	// Open TSM reader.
+	input, err := os.Open(path)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/influx_inspect/deletetsm/deletetsm.go
```diff
@@ -0,0 +1,161 @@
...
+	defer input.Close()
+
+	r, err := tsm1.NewTSMReader(input)
+	if err != nil {
+		return fmt.Errorf("unable to read %s: %s", path, err)
+	}
+	defer r.Close()
+
+	// Remove previous temporary files.
+	outputPath := path + ".rewriting"
```
jwilder: I'd use `.tmp` so that the server will remove any failed runs.  `.rewriting` will linger forever if the process crashes and is not retried.

cmd/influx_inspect/deletetsm/deletetsm.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	// Create TSMWriter to temporary location.
+	output, err := os.Create(outputPath)
+	if err != nil {
+		return err
+	}
+	defer output.Close()
+
+	w, err := tsm1.NewTSMWriterWithDiskBuffer(output)
```
jwilder: Since this is an external tool, you can use `tsm1.NewTSMWriter` which won't be as taxing on the disks and should run faster.

cmd/influx_inspect/deletetsm/deletetsm.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return err
+	}
+	defer w.Close()
+
+	// Iterate over the input blocks.
+	itr, err := tsm1.NewTSMKeyIterator(tsdb.DefaultMaxPointsPerBlock, true, nil, r)
```
jwilder: You can use `r.BlockIterator` here.  That's the fastest way to iterate over all blocks.

cmd/influx_inspect/deletetsm/deletetsm.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// Write index & close.
+	if err := w.WriteIndex(); err != nil {
+		return err
+	} else if err := w.Close(); err != nil {
+		return err
+	}
+
+	// Replace original file with new file.
+	if err := os.Rename(outputPath, path); err != nil {
```
e-dard: `return os.Rename(outputPath, path)`

cmd/influx_inspect/deletetsm/deletetsm.go
```diff
@@ -0,0 +1,161 @@
...
+		log.SetOutput(ioutil.Discard)
+	}
+
+	// Validate measurement flag.
+	if cmd.measurement == "" {
+		return fmt.Errorf("measurement name required")
+	}
+
+	// Process each TSM file.
+	for _, path := range fs.Args() {
```
e-dard: I don't think this is a blocker to merging the tool, but I think it's annoying to use if you have to provide all TSM files. It seems too easy for users to miss one out for example, unless they're familiar with tools like `xargs` and so on.

I think by default we look for tsm files (so we can use other unix tools to generate that set), but we should also provide a way for a user to provide a shard, retention policy or even a database as the argument.

cmd/influx_inspect/deletetsm/deletetsm.go
benbjohnson: Fixed in 8c9d179.

cmd/influx_inspect/deletetsm/deletetsm.go
benbjohnson: Fixed in 8c9d179.

cmd/influx_inspect/deletetsm/deletetsm.go
benbjohnson: Fixed in 8c9d179.

cmd/influx_inspect/deletetsm/deletetsm.go
benbjohnson: Fixed in 8c9d179.

cmd/influx_inspect/deletetsm/deletetsm.go
benbjohnson: Users want to work on a subset of time ranges so it seems like there's some fine grained filters we'd need to add to work for all use cases. I think it makes sense to keep the tooling low level to allow people to filter TSM files how they want external to this command. A separate TSM finder command would make sense to me instead of baking it into each subcommand.

e-dard: SGTM. Either way, I think we could make the experience a bit better. I think a tsm filter that you can pipe into this would be nice.

stuartcarnie: Sounds like we're thinking similarly, as I mentioned a similar approach to @rbetts regarding how I might do deletes with the `influx-tools` using a pipe filter:

```
$ influx-tools export | TSM filter | influx-tools import
```

PR #9748, opened 2018-04-20
Prevent series type conflict 

jwilder:
This fixes the `Engine` to prevent type conflicts within a series.  This is handled by using a radix tree to map series keys to types.  The radix tree provides compression for the mostly duplicate prefixes of the series keys and functions as a map like structure.

There is no persistence of this structure and it does not require it.  The first time a series is written, the mapping is checked and if the series does not exists, it is looked up from the `Cache` and/or `FileStore`.   This means there is no extra disk IO at write time or startup.

This does not currently handle deletes (will follow up with separate PR) and stats are not recorded for type conflicts.

### DIFF
--- a/pkg/radix/tree.go
+++ b/pkg/radix/tree.go
@@ -1,0 +1,575 @@
+package radix
+
+// This is a fork of https://github.com/armon/go-radix that removes the
+// ability to update nodes as well as uses fixed int value type.
+
+import (
+	"sort"
+	"strings"
+	"sync"
+)
+
+// WalkFn is used when walking the tree. Takes a
+// key and value, returning if iteration should
+// be terminated.
+type WalkFn func(s string, v int) bool
+
+// leafNode is used to represent a value
+type leafNode struct {
+	key string
+	val int
+}
+
+// edge is used to represent an edge node
+type edge struct {
+	label byte
+	node  *node
+}
+
+type node struct {
+	// leaf is used to store possible leaf
+	leaf *leafNode
+
+	// prefix is the common prefix we ignore
+	prefix string
+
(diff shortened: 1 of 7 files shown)


e-dard:
@jwilder I'm going to try and review this today or tomorrow. It would be good to see some of the relevant benchmark results / comparisons to `master`.

e-dard:
@jwilder I tested this in two scenarios:

 1) 30M series across 3 tag keys (`inch -b 10000 -c 20 -t 10,10,100000,3 -m 1 -p 1 -v`)
 2) 10M series across 7 tag keys (`inch -b 10000 -c 20 -t 10,10,10,10,10,10,10 -m 1 -p 1 -v`)

In both cases I compared this branch to master and ran inch from a clean empty server run as follows:

```
INFLUXDB_DATA_INDEX_VERSION=tsi1 
INFLUXDB_HTTP_LOG_ENABLED=false INFLUXDB_DATA_CACHE_SNAPSHOT_MEMORY_SIZE=50000000 influxd
```

### Results


| Scenario        | master           | jw-series-type  | slowdown |
| ------------- |:-------------:| -----:| -----:|
| 1)      | 155s (193,355 v/s) | 712s (42,137 v/s) | 4.59 X
| 2)      | 44s (224,213 v/s)      | 231s (43,141 v/s)   | 5.25 X

I have attached a 2 minute CPU profile.

My suggestion is that we will merge this at some point, but with the series type checker switch **off** (or turned on via a private env var). Then the performance can be improved and smaller incremental PRs can be made until we are happy with the performance.

jwilder:
@e-dard Pushed up a feature commit to disable this by default.  Can be enabled w/ an env var.

PR #9753, opened 2018-04-20
Add retention policy query parameter to queries

jsternberg:
Included a note in the client README about the alternative library
location with semantic versioning.


hercules-influx:
``````

PR #9757, opened 2018-04-23
Add suppress-write-log option to disable the write log when the log is enabled

jsternberg:


jsternberg:
@mark-rushakoff done.

PR #9761, opened 2018-04-23
Fix tsdb/index/inmem benchmark tests

jacobmarble:
Fixes broken benchmarks.


hercules-influx:
```/tmp/256827022/src/github.com/influxdata/influxdb/tsdb/index/inmem/inmem_test.go:127:29: func (*seriesFileWrapper).reopen is unused (U1000)
```

Issue #9745, opened 2018-04-19
Dropped point count can over-count dropped points.


e-dard:
If you have multiple field conflict errors on a single point, e.g., you have two fields on a point and they are _both_ the wrong type, then the `dropped` count (the number of points in a batch we log as being dropped, will be `2`, when it should be `1`.

See [here](https://github.com/influxdata/influxdb/blob/master/tsdb/shard.go#L634).

e-dard:
From a cursory glance it also looks like we [don't reset the field iterator](https://github.com/influxdata/influxdb/blob/master/tsdb/shard.go#L586-L589) if we have a field called `time` on the point.

PR #9762, opened 2018-04-24
Add FieldValidator to allow custom validations on measurements

zeebo:
No appreciable changes in benchmark results. It seems like this
function is less than 4% of cpu time in the write workloads in the
benchmarks at least.

### DIFF
--- a/tsdb/field_validator.go
+++ b/tsdb/field_validator.go
@@ -1,0 +1,70 @@
+package tsdb
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/influxdata/influxdb/models"
+	"github.com/influxdata/influxql"
+)
+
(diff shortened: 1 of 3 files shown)


zeebo:
I believe this also addresses #9745

tsdb/field_validator.go
```diff
@@ -0,0 +1,70 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+// dataTypeFromModelsFieldType returns the influxql.DataType that corresponds to the
+// passed in field type. If there is no good match, it returns Unknown and false.
+func dataTypeFromModelsFieldType(fieldType models.FieldType) (influxql.DataType, bool) {
```
stuartcarnie: you could probably eliminate the `bool` return and let the caller use `influxql.Unknown` as the signal

PR #9770, opened 2018-04-24
Add verify-seriesfile to influx_inspect

### LINKED ISSUE #9767: Create a `verify series-file` tool
Currently, we have a verification tool for [TSM data](https://github.com/influxdata/influxdb/tree/master/cmd/influx_inspect/verify).

It's accessed via `influx_inspect verify /path/to/root/storage`.

This tool iterates over all TSM files in all shards in all rps and databases, and verifies that they are intact and healthy.

I would like to see a tool created (`influx_inspect verify-seriesfile`) performs the same task but on a Series File.

It should carry out the following:

 - Default to `~/.influxdb/data` as the root tsdb data dir.
 - Allow the user to change the data dir argument.
 - Allow the user to provide a specific database to verify.
 - Iterate over all series file partitions and verify their correctness.
 - Verify the correctness of the series file index.

zeebo:
Fixes #9767


zeebo:
I'm gonna be adding tests, but I figured it'd be useful to show the implementation to get a quicker feedback cycle. Specifically,

- is it a good idea to duplicate this much logic about the layout and semantics of the things? It provides much better diagnostics about what is wrong this way.
- is the logging an appropriate way to provide feedback? too verbose? not enough?
- anything I missed about integrating the new command (docs, etc.)?

cmd/influx_inspect/verify_seriesfile/verify_seriesfile.go
```diff
@@ -0,0 +1,298 @@
...
+}
+
+// Run executes the command.
+func (cmd *Command) Run(args ...string) error {
+	fs := flag.NewFlagSet("verify-seriesfile", flag.ExitOnError)
+	fs.SetOutput(cmd.Stdout)
+
+	dataDir := fs.String("dir", os.Getenv("HOME")+"/.influxdb", "data directory.")
+	dbPath := fs.String("db", "", "database path. overrides data directory.")
+	filePath := fs.String("file", "", "series file path. overrides db and data directory.")
```
e-dard: Maybe we could match up with `dumptsi` here? That tool has a `fs.StringVar(&cmd.seriesFilePath, "series-file", "", "Path to series file")`

The description could be `Path to series file. This overrides -db and -dir` or similar?

cmd/influx_inspect/verify_seriesfile/verify_seriesfile.go
```diff
@@ -0,0 +1,298 @@
...
+	if err := fs.Parse(args); err != nil {
+		return err
+	}
+
+	return cmd.run(*dataDir, *dbPath, *filePath)
+}
+
+// run performs all of the validations, preferring a filePath, then a dbPath,
+// and then a dataDir.
+func (cmd *Command) run(dataDir, dbPath, filePath string) error {
```
e-dard: Could we change this up a bit so there is a programatic entrypoint to the tool? I would like eventually for all tools to be able to be run from inside an InfluxDB (or other) process. I don't think that's currently possible if the exported method handles flag stuff. Feels like this method should be exported (and I guess renamed not to clash with the one that handles flags?)

cmd/influx_inspect/verify_seriesfile/verify_seriesfile.go
```diff
@@ -0,0 +1,298 @@
...
+		_, err := cmd.runFile(filepath.Join(dbPath, "_series"))
+		return err
+	}
+
+	dbs, err := ioutil.ReadDir(filepath.Join(dataDir, "data"))
+	if err != nil {
+		return err
+	}
+
+	for _, db := range dbs {
```
e-dard: What do you think about parallelising this? It would be nice to add a `-p` flag that sets the maximum number of OS-threads we would dedicate to the task, and then spin up that many goroutines to work on a series file each.

e-dard: I'm open to names other than `-p` and it would default to `runtime.GOMAXPROCS(0)`.

cmd/influx_inspect/verify_seriesfile/verify_seriesfile.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	return nil
+}
+
+// runFile performs validations on a series file. The error is only returned if
+// there was some fatal problem with operating, not if there was a problem with
+// the series file.
+func (cmd *Command) runFile(filePath string) (valid bool, err error) {
+	logger := logger.New(cmd.Stderr).With(zap.String("path", filePath))
```
e-dard: nit: this could be set higher up, and then this line could be: `logger := cmd.logger.With(zap.String("path", filePath))`

cmd/influx_inspect/verify_seriesfile/verify_seriesfile.go
```diff
@@ -0,0 +1,298 @@
...
+	logger.Info("starting validation")
```
e-dard: `Starting validation`.

cmd/influx_inspect/verify_seriesfile/verify_seriesfile.go
```diff
@@ -0,0 +1,288 @@
+// Package verify_seriesfile verifies integrity of series files.
+package verify_seriesfile
```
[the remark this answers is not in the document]
zeebo: seems fine to me. since they're command oriented i'd prefer the latter.

cmd/influx_inspect/verify_seriesfile/verify_seriesfile.go
zeebo: I'll do some refactoring to make this more consumable externally.

cmd/influx_inspect/verify_seriesfile/verify_seriesfile.go
zeebo: I think the parallel wins would probably come from doing it at the partition level. Maybe both. I'll make it go fast. :dash:

cmd/influx_inspect/verify_seriesfile/verify_seriesfile.go
```diff
@@ -0,0 +1,298 @@
...
+
+			sLogger := pLogger.With(zap.String("segment", segmentInfo.Name()))
+			sLogger.Info("validating segment")
+
+			segment := tsdb.NewSeriesSegment(segmentID, segmentPath)
+			if err := segment.Open(); err != nil {
+				sLogger.Error("opening segment", zap.Error(err))
+				return false, nil
+			}
+			defer segment.Close()
```
benbjohnson: This `Close()` won't be deferred until the very end of `runFile()`. That probably isn't a huge issue unless there are a bunch of segments and limited file descriptors. Maybe refactor this block to a `runSegment()` function so `defer` runs after ever segment?

cmd/influx_inspect/verify_seriesfile/verify_seriesfile.go
```diff
@@ -0,0 +1,298 @@
...
+
+		// validate the index if it exists
+		indexPath := filepath.Join(partitionPath, "index")
+		_, err = os.Stat(indexPath)
+		if os.IsNotExist(err) {
+			continue
+		}
+		if err != nil {
+			return false, err
+		}
```
benbjohnson: This is mostly personal preference but I generally group my error checks in a single `if`/`else if` block:

```go
if _, err := os.Stat(indexPath); os.IsNotExist(err) {
	continue
} else if err != nil {
	return false, err
}
```

cmd/influx_inspect/verify_seriesfile/verify_seriesfile.go
```diff
@@ -0,0 +1,298 @@
...
+	data := segment.Data()
+
+	advance := func(n int64) error {
+		if int64(len(data)) < n {
+			return fmt.Errorf("unable to advance %d bytes: %d remaining", n, len(data))
+		}
+		offset += n
+		data = data[n:]
+		return nil
+	}
```
benbjohnson: I'm personally not a fan of inline functions. I find it harder to read the code. I would either move the function out to a helper or inline the code to where it's used.

cmd/influx_inspect/verify_seriesfile/verify_seriesfile_test.go
```diff
@@ -0,0 +1,169 @@
...
+	}}
+}
+
+func (t *TestCommand) StdoutString() string {
+	return t.Stdout.(*bytes.Buffer).String()
+}
+
+func (t *TestCommand) StderrString() string {
+	return t.Stderr.(*bytes.Buffer).String()
+}
```
benbjohnson: We don't generally use assert helpers in the code. Although now that I do a quick search in the code we do have a `pkg/testing/assert` package.

@e-dard Did we change our stance on testing style?

benbjohnson:
> _is it a good idea to duplicate this much logic about the layout and semantics of the things? It provides much better diagnostics about what is wrong this way._

Which part are you thinking is duplication?

> _is the logging an appropriate way to provide feedback? too verbose? not enough?_

Yes, I think logging seems appropriate. Maybe add a `-v` flag to toggle verbosity.

> _anything I missed about integrating the new command (docs, etc.)?_

No, i think it looks good.

zeebo:
> Which part are you thinking is duplication?

For example, the parts that understand how to iterate over series files and the directory layout of partitions and series files.

cmd/influx_inspect/verify_seriesfile/verify_seriesfile_test.go
zeebo: I personally don't like big assertion libraries either, because they end to include a bunch of reflect magic that makes it hard to know if what it claims to do is actually what it's doing. I don't mind killing it, but I thought this might be a middle ground between a huge testing assertion package, and verbose failure calls.

cmd/influx_inspect/verify_seriesfile/verify_seriesfile.go
zeebo: I actually intend for them not to be closed until the end since we use them for checking the index later. Defer in a for loop is definitely a red flag, so I'll leave a comment to avoid false positives.

benbjohnson:
I think the iteration piece is OK since it's using `tsdb` functions to read the entries. I agree that the directory layout could be handled better. @e-dard and I have discussed adding commands to `influx_inspect` to filter paths so they can be piped together.

```
$ influx_inspect find-series-files -db DB | xargs influx_inspect verify-seriesfile
```

Then we'd only need to take `PATH [PATH...]` as the argument. However, that's out of the scope for this PR.

cmd/influx_inspect/verify_seriesfile/verify_seriesfile.go
zeebo: Oh nope, I'm wrong. That's an artifact from when it was just checking a single partition in a function. Thanks.

hercules-influx:
```/tmp/614425028/src/github.com/influxdata/influxdb/cmd/influx_inspect/verify/seriesfile/verify_test.go:115:3: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
/tmp/614425028/src/github.com/influxdata/influxdb/cmd/influx_inspect/verify/seriesfile/verify_test.go:115:3: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
/tmp/614425028/src/github.com/influxdata/influxdb/tsdb/index/inmem/inmem_test.go:127:29: func (*seriesFileWrapper).reopen is unused (U1000)
```

cmd/influx_inspect/verify/seriesfile/verify_test.go
```diff
@@ -0,0 +1,163 @@
...
+			return err
+		}
+
+		// wait for compaction to make sure we detect issues with the index
+		partitions := seriesFile.Partitions()
+	wait:
+		for _, partition := range partitions {
+			if partition.Compacting() {
+				time.Sleep(100 * time.Millisecond)
+				goto wait
```
e-dard: I really am not a fan of `goto`, but this seems legit.

cmd/influx_inspect/verify/seriesfile/verify.go
```diff
@@ -0,0 +1,399 @@
...
+	v.Logger.Debug("Verifying segment")
+
+	// Open up the segment and grab it's data.
+	segmentID, err := tsdb.ParseSeriesSegmentFilename(segmentName)
+	if err != nil {
+		return false, err
+	}
+	segment := tsdb.NewSeriesSegment(segmentID, segmentPath)
+	if err := segment.Open(); err != nil {
+		v.Logger.Error("Error opening segment", zap.Error(err))
```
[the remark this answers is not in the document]
zeebo: I think you're actually right. I think I messed this up during the verification.

cmd/influx_inspect/verify/seriesfile/verify.go
```diff
@@ -0,0 +1,399 @@
...
+				return false, nil
+			}
+
+			firstID = false
+			prevID = id
+
+			if ids != nil {
+				ids[id] = IDData{
+					Offset: tsdb.JoinSeriesOffset(segment.ID(), uint32(buf.offset)),
+					Key:    append([]byte(nil), key...),
```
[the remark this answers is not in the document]
zeebo: That would be `[]byte{nil}`, not `[]byte(nil)`. This is a way to make a copy of the key.

cmd/influx_inspect/verify/seriesfile/verify.go
```diff
@@ -0,0 +1,399 @@
...
+			return false, err
+		} else if !result.valid {
+			return false, nil
+		}
+	}
+
+	return true, nil
+}
+
+// VerifyPartition performs verifications on a parition of a series file. The error is only returned
```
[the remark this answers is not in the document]
zeebo: My eternal issue is spelling.

cmd/influx_inspect/verify/seriesfile/command.go
```diff
@@ -0,0 +1,119 @@
...
+	fs.SetOutput(cmd.Stdout)
+	fs.Usage = cmd.printUsage
+
+	if err := fs.Parse(args); err != nil {
+		return err
+	}
+
+	config := logger.NewConfig()
+	if cmd.verbose {
+		config.Level = zapcore.DebugLevel
```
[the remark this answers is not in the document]
zeebo: The reason I went with DEBUG is because INFO was on by default. I needed some level below INFO that was off, that could be turned on with verbose. Thoughts?

cmd/influx_inspect/verify/seriesfile/command.go
```diff
@@ -0,0 +1,119 @@
...
+	fs.StringVar(&cmd.dir, "dir", filepath.Join(os.Getenv("HOME"), ".influxdb", "data"),
+		"Data directory.")
+	fs.StringVar(&cmd.db, "db", "",
+		"Only use this database inside of the data directory.")
+	fs.StringVar(&cmd.seriesFile, "series-file", "",
+		"Path to a series file. This overrides -db and -dir.")
+	fs.BoolVar(&cmd.verbose, "v", false,
+		"Verbose output.")
+	fs.IntVar(&cmd.parallel, "p", runtime.GOMAXPROCS(0),
+		"How many parallel workers to run.")
```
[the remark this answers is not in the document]
zeebo: I could change the word to "concurrent"? The users would be in control of GOMAXPROCS to control upper bounds on parallelism.

cmd/influx_inspect/verify/seriesfile/verify.go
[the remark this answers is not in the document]
e-dard: I think the fact that I misread the parentheses for braces is probably a sign that this pattern for copying is a bit confusing 😄

cmd/influx_inspect/verify/seriesfile/command.go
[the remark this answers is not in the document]
e-dard: Ah I see. A couple of ways this gets done:

  - two loggers. One logs to stderr one to stdout. `-v` turns on the stdout one.
  - Only use the logger for the verbose stuff (at `INFO` level). So that means `-v` initialises a logger and its absence means we use the `NOP` one. Error messages get emitted via `fmt.FprintX`.

Or we just stick with what you have done, which on reflection seems completely fine. @benbjohnson what do you think?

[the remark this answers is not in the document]
benbjohnson: @e-dard What about having `-v` set level to `INFO` and no `-v` sets the level to `WARN`? Or just leave it. That works too.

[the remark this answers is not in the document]
e-dard: All sounds good to me

e-dard:
LGTM :+1:

PR #9778, opened 2018-04-25
Include the query task status in the QueryInfo struct

jsternberg:
Previously, the task manager was modified to keep the query status so it
could track which queries were running and which ones were killed.
In those previous versions, we removed a task from the process table
as soon as it was killed and did not remove it after it had finished
executing. This meant there could be zombie goroutines running in the
background that were impossible to see.

When the task manager was updated to track the task status, we forgot to
expose the status in the public interface so consumers could see the
task status.

### DIFF
--- a/query/task_manager.go
+++ b/query/task_manager.go
@@ -1,7 +1,9 @@
 package query
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"sync"
 	"time"
@@ -21,7 +23,7 @@ type TaskStatus int
 
 const (
 	// RunningTask is set when the task is running.
-	RunningTask TaskStatus = iota
+	RunningTask TaskStatus = iota + 1
 
 	// KilledTask is set when the task is killed, but resources are still
 	// being used.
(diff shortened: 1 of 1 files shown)


hercules-influx:
```/tmp/348578972/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/command.go:77:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
/tmp/348578972/src/github.com/influxdata/influxdb/tsdb/index/inmem/inmem_test.go:127:29: func (*seriesFileWrapper).reopen is unused (U1000)
```

hercules-influx:
```/tmp/231737902/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/command.go:77:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
/tmp/231737902/src/github.com/influxdata/influxdb/tsdb/index/inmem/inmem_test.go:127:29: func (*seriesFileWrapper).reopen is unused (U1000)
```

e-dard:
@jsternberg could we get some context on this PR?

PR #9783, opened 2018-04-27
tsm1: Replace goroutine `merge` with k-way merge

stuartcarnie:
Previously replaced `WalkKeys` implementation for a considerable
improvement to startup time. The previous commit for reference https://github.com/influxdata/influxdb/commit/e1ec33104886452a46b26e18ca4bd1764f7e8e31


hercules-influx:
```/tmp/132448102/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/command.go:77:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
/tmp/132448102/src/github.com/influxdata/influxdb/tsdb/index/inmem/inmem_test.go:127:29: func (*seriesFileWrapper).reopen is unused (U1000)
```

PR #9784, opened 2018-04-27
Changes /ping route to return status code 200 instead of 204.

### LINKED ISSUE #9772: HTTP 204 returned from /ping endpoint is unsuitable for Google Kubernetes Cloud healthcheck
## Use case

I want to deploy an InfluxDB instance to the Google Kubernetes Cloud. In order to expose the application to the outer world, I need to create an Ingress.

## Problem

The Ingress requires a health check. I can configure the endpoint of the health check but it needs to return a HTTP 200. The `/ping` endpoint currently returns an HTTP 204 which is unsuitable.

## Sugggestion

Consider adding an endpoint that returns a HTTP 200 or make the status code returned from `/ping` configureable.

ryuhhnn:
Closes #9772

### DIFF
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -784,8 +784,14 @@ func (h *Handler) serveOptions(w http.ResponseWriter, r *http.Request) {
 
 // servePing returns a simple response to let the client know the server is running.
 func (h *Handler) servePing(w http.ResponseWriter, r *http.Request) {
+	statusCode := r.URL.Query().Get("status")
 	atomic.AddInt64(&h.stats.PingRequests, 1)
-	h.writeHeader(w, http.StatusOK)
+
+	if statusCode == "200" {
+		h.writeHeader(w, http.StatusOK)
+	} else {
+		h.writeHeader(w, http.StatusNoContent)
+	}
 }
 
 // serveStatus has been deprecated.


jsternberg:
Since the `/ping` route returns a content type of `application/json`, this should probably return either `{}` in the body or something like `{"status":"ok"}`.

ryuhhnn:
@jsternberg this is a good point. What about changing the endpoint to return a content type of `text/plain`? This seems more appropriate to me than returning `application/json`.

mark-rushakoff:
@timhallinflux can you weigh in on this behavior change?

I reckon there are existing user scripts that expect a 204 return code. If we need to preserve existing behavior, we can have the status default to 204 and add support for something like `/ping?status=200` to get a 200 return code.

ryuhhnn:
@mark-rushakoff @timhallinflux thank you guys for your feedback! I've updated the PR per your suggestions.

jsternberg:
Can you change the query parameter to `verbose=true` and make it so it returns a JSON body like this?

```
{"version": "1.5"}
```

This information is already included in a header, but I do want some JSON text output and this seems like something that could be added.

jsternberg:
The original repository doesn't exist anymore so I'm going to modify this and make a new PR.

ryuhhnn:
@jsternberg will you allow me to create a new PR since I was the original author of this one? I deleted the original repository as part of some housekeeping on my repos and wasn't seeing any movement on this PR.

jsternberg:
That's fine. I was going to ensure you stayed as the author on the actual commit (I just end up showing up as the committer). Here's a modified diff of your original that I was going to apply and commit.

```
commit c93068794f9e72d2aa3daa2cf25c4baeb49c0df4 (HEAD -> pr-9784)
Author: Ryan Coleman <me@ryuhhnn.com>
Date:   Wed May 2 16:34:57 2018 -0500

    Changes the functionality of /ping to return a status code of 200 when /ping?verbose=1 is used

diff --git a/services/httpd/handler.go b/services/httpd/handler.go
index 7bf44498a..6e7253f2a 100644
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -784,8 +784,16 @@ func (h *Handler) serveOptions(w http.ResponseWriter, r *http.Request) {

 // servePing returns a simple response to let the client know the server is running.
 func (h *Handler) servePing(w http.ResponseWriter, r *http.Request) {
+       verbose := r.URL.Query().Get("verbose")
        atomic.AddInt64(&h.stats.PingRequests, 1)
-       h.writeHeader(w, http.StatusOK)
+
+       if verbose != "" && verbose != "0" && verbose != "false" {
+               h.writeHeader(w, http.StatusOK)
+               b, _ := json.Marshal(map[string]string{"version": h.Version})
+               w.Write(b)
+       } else {
+               h.writeHeader(w, http.StatusNoContent)
+       }
 }

 // serveStatus has been deprecated.
```

If you want to apply it locally and submit a PR for github stats, I can merge that one instead. I don't really have a preference on that.

PR #9785, opened 2018-04-27
Rename & log corrupt tsm files on load

benbjohnson:
While opening, this change updates TSM files to have a `.tsm.bad` extension if the files are corrupt.


hercules-influx:
```/tmp/232802298/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/command.go:77:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
/tmp/232802298/src/github.com/influxdata/influxdb/tsdb/index/inmem/inmem_test.go:127:29: func (*seriesFileWrapper).reopen is unused (U1000)
```

PR #9792, opened 2018-04-30
TSM: TSMReader.Close blocks until reads complete

### LINKED ISSUE #9786: TSM files not closed when shard deleted
InfluxDB current master 28-apr-2018 running on current MacOS.

The service removes, but does not close, TSM files when a shard is dropped while the TSM files are in use by a query.

- start influxdb
- create a 1h retention policy
- write hundreds of megabytes to the new RP
- `select sum(foo) from m` takes a long time to complete
- run that ^^ query in an infinite loop, 10 goroutines
- `drop shard X`
- shard directory and contained files disappear from filesystem as viewed via `ls`
- `lsof -p PID` shows shard .tsm files still open
- stop query load
- `lsof` still shows .tsm files open

jacobmarble:
Fixes #9786

`TSMReader.Close` returns an error when queries are referencing the TSM file, but `FileStore.Close` ignores the error.

In this change, `FileStore.Close` checks the error, and `TSMReader.Close` blocks, rather than returning the error. This makes it clear to the caller of `FileStore.Close` (`drop shard X` for example) that the file is safely closed, rather than requiring the caller to retry up to thousands of times when the query load is constantly high.

After this change, TSM files are always closed when `drop shard X` returns, and writing to a shard that is currently being deleted returns a partial error.

### DIFF
--- a/coordinator/points_writer.go
+++ b/coordinator/points_writer.go
@@ -2,6 +2,7 @@ package coordinator
 
 import (
 	"errors"
+	"fmt"
 	"sort"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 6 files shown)


hercules-influx:
```/tmp/589773538/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/command.go:77:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
/tmp/589773538/src/github.com/influxdata/influxdb/tsdb/index/inmem/inmem_test.go:127:29: func (*seriesFileWrapper).reopen is unused (U1000)
```

jacobmarble:
Tests are timing out. That's probably important in this change, so I'm digging in now.

jacobmarble:
This should fix tests. Just needed to close iterators and cursors.

jacobmarble:
@e-dard Thanks!

PR #9793, opened 2018-04-30
Add key sanitization to deletetsm

benbjohnson:
## Overview

This commit adds the `-sanitize` flag to `influx_inspect deletetsm` which will delete all keys that contain invalid, non-printable, or replacement character unicode.

## Usage:

```sh
$ influx_inspect deletetsm -sanitize PATH
```

/cc @timhallinflux

### DIFF
--- a/cmd/influx_inspect/deletetsm/deletetsm.go
+++ b/cmd/influx_inspect/deletetsm/deletetsm.go
@@ -21,6 +21,7 @@ type Command struct {
 	Stdout io.Writer
 
 	measurement string // measurement to delete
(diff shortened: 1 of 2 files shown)


hercules-influx:
```/tmp/569486538/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/command.go:77:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
/tmp/569486538/src/github.com/influxdata/influxdb/tsdb/index/inmem/inmem_test.go:127:29: func (*seriesFileWrapper).reopen is unused (U1000)
```

Issue #9766, opened 2018-04-24
Provide an API to determine the byte size of the inmem index


e-dard:
It's not currently possible to easily determine how much memory the `inmem` index takes up. By providing an API to access this information, it would be possible to provide more accurate sizing guidelines based on empirical tests done with different series key lengths, series cardinality etc.

PR #9794, opened 2018-05-01
tsdb/index: Add Bytes() methods

jacobmarble:
Provides internal API to make progress on #9766 

Memory counting is a bit complicated because `go vet` does not allow copying instances of `sync.Mutex` and similar data structures, and `unsafe.Sizeof` violates that check. To work around this, there is a lot of field-by-field counting in this change, but I tried to consistently comment exceptions and count fields in the order they are presented in each struct.

### DIFF
--- a/pkg/estimator/hll/hll.go
+++ b/pkg/estimator/hll/hll.go
@@ -102,7 +102,7 @@ func NewPlus(p uint8) (*Plus, error) {
 	return hll, nil
 }
 
-// Bytes returns an estimate of the memory footprint of this instance of Plus, in bytes.
+// Bytes estimates the memory footprint of this Plus, in bytes.
 func (h *Plus) Bytes() int {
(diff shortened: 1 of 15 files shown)


hercules-influx:
``````

hercules-influx:
``````

jacobmarble:
This whole PR originally used `unsafe.Sizeof` more liberally, but [this bug](https://github.com/golang/go/issues/21800) in go vet prevents me from calling `unsafe.Sizeof(mu)` because that implies copying the mutex. This bug exists in golang 1.9, which is what we're currently building with. I also tried `reflect.TypeOf(foo).Size()`, same result.

1. See above.
2. I don't have a preference, and I'm not aware of any advantage.
3. See above.

benbjohnson:
Make sense. lgtm

PR #9801, opened 2018-05-02
Add option for unicode validation.

benbjohnson:
## Overview

This pull request adds a `validate-keys` configuration field that requires shard writes to check for valid, printable UTF-8 characters. Points with invalid keys will be dropped and a partial write error is returned.

### DIFF
--- a/etc/config.sample.toml
+++ b/etc/config.sample.toml
@@ -67,6 +67,10 @@
   # log any sensitive data contained within a query.
   # query-log-enabled = true
 
+  # Validates incoming writes to ensure keys only have valid unicode characters.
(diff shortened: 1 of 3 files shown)


etc/config.sample.toml
```diff
@@ -67,6 +67,10 @@
   # log any sensitive data contained within a query.
   # query-log-enabled = true
 
+  # Validates incoming writes to ensure keys only have valid unicode characters.
+  # This setting will incur a small overhead because every key must be checked.
```
e-dard: nit: I would remove `small`; I think the overhead could be significant depending on the key length for example.

PR #9812, opened 2018-05-07
tsdb/WAL: Better respect for WAL disabled

jacobmarble:
While toying with the TSM Engine, I notice that WAL files are created when WAL is disabled. Turns out that the WAL object is instantiated, but only checked for sometimes, when configured to be disabled.

This change removes instantiation and all use of `Engine.WAL` when the option is disabled.


hercules-influx:
```/tmp/720110960/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

PR #9822, opened 2018-05-09
Implement SHOW STATS FOR 'indexes'

### LINKED ISSUE #9766: Provide an API to determine the byte size of the inmem index

jacobmarble:
Closes #9766 

This change adds `SHOW STATS FOR 'indexes'` which returns an estimate of memory use of all indexes. Index memory use is *not* reported with `SHOW STATS` because it is a potentially expensive operation, and the implementation of `SHOW STATS` is also triggered every 10s to populate the `_internal` database.

Tested by comparing the output of this new function to the heap pprof. Results are comparable, not perfect, but pprof doesn't hold absolute truth either.

### DIFF
--- a/cmd/influxd/run/server.go
+++ b/cmd/influxd/run/server.go
@@ -183,7 +183,7 @@ func NewServer(c *Config, buildInfo *BuildInfo) (*Server, error) {
 	s.QueryExecutor.StatementExecutor = &coordinator.StatementExecutor{
 		MetaClient:  s.MetaClient,
 		TaskManager: s.QueryExecutor.TaskManager,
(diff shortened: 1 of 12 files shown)


hercules-influx:
``````

**Comment by jacobmarble** [comment]
Working on the failing test.

**Comment by hercules-influx** [comment]
``````

**Comment by hercules-influx** [comment]
```/tmp/195668190/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

PR #9824, opened 2018-05-10
radix: optimize for our use case

zeebo:
- reduce allocations by making leaf a value type with a bool
- make longestPrefix inlineable and have no bounds checks
- delete any code for functions we don't plan to use
- operate on []byte and only copy when necessary
- inline calls to sort.Search to avoid allocations and indirections
- insert directly in the correct location for addEdge
- reduce allocations during copying with a buffer helper

results:

```
name              old time/op    new time/op     delta
Tree_Insert-8       1.10ms ± 4%     0.73ms ± 4%  -33.54%  (p=0.000 n=10+10)
Tree_InsertNew-8    3.18ms ± 2%     1.91ms ± 6%  -39.90%  (p=0.000 n=10+10)

name              old speed      new speed       delta
Tree_Insert-8     9.12MB/s ± 4%  13.72MB/s ± 4%  +50.46%  (p=0.000 n=10+10)
Tree_InsertNew-8  3.15MB/s ± 2%   5.24MB/s ± 6%  +66.42%  (p=0.000 n=10+10)

name              old alloc/op   new alloc/op    delta
Tree_InsertNew-8    1.62MB ± 0%     1.60MB ± 0%   -1.28%  (p=0.000 n=10+9)

name              old allocs/op  new allocs/op   delta
Tree_InsertNew-8     35.0k ± 0%      15.0k ± 0%  -57.04%  (p=0.000 n=10+10)
```

MB/sec in this case is 1 byte per key inserted, so it's really millions
of keys inserted per second.

### DIFF
--- a/pkg/radix/tree.go
+++ b/pkg/radix/tree.go
@@ -4,20 +4,16 @@ package radix
 // ability to update nodes as well as uses fixed int value type.
 
 import (
+	"bytes"
 	"sort"
-	"strings"
 	"sync"
 )
 
-// WalkFn is used when walking the tree. Takes a
-// key and value, returning if iteration should
-// be terminated.
-type WalkFn func(s string, v int) bool
-
 // leafNode is used to represent a value
 type leafNode struct {
-	key string
-	val int
+	valid bool // true if key/val are valid
+	key   []byte
+	val   int
 }
 
 // edge is used to represent an edge node
@@ -28,10 +24,10 @@ type edge struct {
 
 type node struct {
 	// leaf is used to store possible leaf
-	leaf *leafNode
+	leaf leafNode
 
 	// prefix is the common prefix we ignore
-	prefix string
+	prefix []byte
 
 	// Edges should be stored in-order for iteration.
 	// We avoid a fully materialized slice to save memory,
@@ -40,12 +36,26 @@ type node struct {
 }
 
 func (n *node) isLeaf() bool {
-	return n.leaf != nil
+	return n.leaf.valid
 }
 
 func (n *node) addEdge(e edge) {
-	n.edges = append(n.edges, e)
-	n.edges.Sort()
+	// find the insertion point with bisection
+	num := len(n.edges)
+	i, j := 0, num
+	for i < j {
+		h := int(uint(i+j) >> 1)
+		if n.edges[h].label < e.label {
+			i = h + 1
+		} else {
+			j = h
+		}
+	}
+
+	// make room, copy the suffix, and insert.
+	n.edges = append(n.edges, edge{})
+	copy(n.edges[i+1:], n.edges[i:])
+	n.edges[i] = e
 }
 
 func (n *node) replaceEdge(e edge) {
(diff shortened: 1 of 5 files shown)


hercules-influx:
```/tmp/037151180/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

pkg/radix/tree.go
```diff
@@ -61,79 +70,55 @@ func (n *node) replaceEdge(e edge) {
...
-}
-
-func (e edges) Sort() {
-	sort.Sort(e)
-}
-
 // Tree implements a radix tree. This can be treated as a
 // Dictionary abstract data type. The main advantage over
 // a standard hash map is prefix-based lookups and
-// ordered iteration.  The tree is safe for concurrent access.
```
jwilder: I'd rather we keep this goroutine safe.  The existing code that uses this relies on this type being safe for concurrent access.  Otherwise, we'll have to perform these operations under a separate mutex or one that already exists in `Engine`.  The spots in the `Engine` where this would be called only need an `RLock` and this would need a `Lock`.  If it's a separate mutex, we open up the possibility of deadlocks.

pkg/radix/tree.go
```diff
@@ -190,18 +187,23 @@ func (t *Tree) Insert(s string, v int) (int, bool) {
 
 		// No edge, create one
 		if n == nil {
+			newNode := &node{
+				leaf: leafNode{
+					key:   byteCopy(s),
```
jwilder: I think you could save an allocation here with something like:
```
buf := make([]byte, len(s)+len(search)
copy(buf[:len(s)], s)
keyCopy := buf[:len(s)]
copy(len(s):], search)
searchCopy := buf[len(s):]
```

pkg/radix/tree.go
```diff
@@ -145,29 +130,39 @@ func (t *Tree) Len() int {
...
+		return int(i)
 	}
-	return i
+	i++
+	goto loop
+}
+
+// byteCopy makes a copy of the byte slice.
+func byteCopy(x []byte) []byte {
+	out := make([]byte, len(x))
```
jwilder: I wonder if it would help to allocate a slightly larger slice and hand out sub-slices while there is enough space for it.  When it's too small, create a new slice to slice up further.

pkg/radix/tree.go
```diff
@@ -396,174 +338,25 @@ func (t *Tree) Get(s string) (int, bool) {
...
-		}
-	}
-	if last != nil {
-		return last.key, last.val, true
-	}
-	return "", 0, false
-}
-
-// Minimum is used to return the minimum value in the tree
-func (t *Tree) Minimum() (string, int, bool) {
```
jwilder: `Minimum` and maybe `Maximum` are needed in a branch I need to push up.  Can you add them back?

pkg/radix/tree.go
zeebo: Oops, yep. I wasn't careful enough at validating the engine side locks.

pkg/radix/tree.go
zeebo: nice!

pkg/radix/tree.go
zeebo: I see, so automatically do the optimization above. I'll give that a shot.

pkg/radix/tree.go
zeebo: No problem. Sorry for going overboard!

hercules-influx:
```/tmp/702309904/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

hercules-influx:
```/tmp/599297474/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

PR #9826, opened 2018-05-10
TSM Filename Injection

benbjohnson:
Provides the ability to inject functions that can customize the naming of TSM filenames. Provides existing implementations as `DefaultParseFileName()` & `DefaultFormatFileName()`.


hercules-influx:
```/tmp/689905600/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

tsdb/engine/tsm1/compact.go
```diff
@@ -681,6 +682,9 @@ type Compactor struct {
 	// RateLimit is the limit for disk writes for all concurrent compactions.
 	RateLimit limiter.Rate
 
+	// Returns the filename used for files generated by CompactFast().
+	CompactFastFormatFileName FormatFileNameFunc
```
jwilder: No need to separate out `Fast` type compactions.  All the names have to be the same.

tsdb/engine/tsm1/engine.go
```diff
@@ -175,6 +175,12 @@ type Engine struct {
 	// writes will only exist in the cache and can be lost if a snapshot has not occurred.
 	WALEnabled bool
 
+	// Invoked when creating a backup file "as new".
+	FormatBackupFileName FormatFileNameFunc
```
jwilder: No need to have separate funcs for these cases.  They should all be the same.

tsdb/engine/tsm1/file_store.go
```diff
@@ -978,8 +978,20 @@ func (f *FileStore) CreateSnapshot() (string, error) {
 	return tmpPath, nil
 }
 
-// ParseTSMFileName parses the generation and sequence from a TSM file name.
-func ParseTSMFileName(name string) (int, int, error) {
+// FormatFileNameFunc is executed when generating a new TSM filename.
+// Source filenames are provided via src.
+type FormatFileNameFunc func(generation, sequence int, src []string) string
```
jwilder: I think this is adding too much flexibility.  The generation/sequence have specific semantics that apply for compactions and query processing.  I think it would safer to allow the generation to part of the file to be injected.

The contract is that it has to create an increasing integer and cant return an a value less than anything that already exists as a TSM file.

hercules-influx:
```/tmp/968114862/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

benbjohnson:
@jwilder I simplified the naming a lot in 4a9c6f2. Can you take a look again?

hercules-influx:
```/tmp/862337132/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

benbjohnson:
I pushed up a callback on `EngineOptions` to allow the caller to inject options to the `Engine` at initialization: a82efef.

tsdb/engine/tsm1/engine.go
```diff
@@ -1779,6 +1784,9 @@ func (e *Engine) WriteSnapshot() error {
 			return
 		}
 
+		// Determine filename generation function under lock.
+		formatFileName = e.GenerateFormatFileNameFunc()
```
zeebo: How do you plan to have whatever this implementation is know what kafka index is the upper bound for the cache snapshot and wal segments?

PR #9827, opened 2018-05-10
Correct godoc for SeriesFile.CreateSeriesListIfNotExists

jacobmarble:
The existing godoc doesn't match the implementation (SeriesPartition in particular) or the way CreateSeriesListIfNotExists is used in the two places where the returned `ids` value isn't thrown away. In fact, those two usages look like they might break if current godoc were true.

### DIFF
--- a/tsdb/series_partition.go
+++ b/tsdb/series_partition.go
@@ -174,7 +174,7 @@ func (p *SeriesPartition) Path() string { return p.path }
 func (p *SeriesPartition) IndexPath() string { return filepath.Join(p.path, "index") }
 
 // CreateSeriesListIfNotExists creates a list of series in bulk if they don't exist.
-// The returned ids list returns values for new series and zero for existing series.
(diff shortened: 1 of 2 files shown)


tsdb/series_partition.go
```diff
@@ -174,7 +174,7 @@ func (p *SeriesPartition) Path() string { return p.path }
 func (p *SeriesPartition) IndexPath() string { return filepath.Join(p.path, "index") }
 
 // CreateSeriesListIfNotExists creates a list of series in bulk if they don't exist.
-// The returned ids list returns values for new series and zero for existing series.
+// The ids parameters is modified to contain series IDs for all keys belonging to this partition.
```
benbjohnson: `s/parameters/parameter/

hercules-influx:
```/tmp/024000582/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

PR #9838, opened 2018-05-14
Include the query task status in the QueryInfo struct

jsternberg:
Previously, the task manager was modified to keep the query status so it
could track which queries were running and which ones were killed.
In those previous versions, we removed a task from the process table
as soon as it was killed and did not remove it after it had finished
executing. This meant there could be zombie goroutines running in the
background that were impossible to see.

When the task manager was updated to track the task status, we forgot to
expose the status in the public interface so consumers could see the
task status.

Backport of #9778.

### DIFF
--- a/query/task_manager.go
+++ b/query/task_manager.go
@@ -1,6 +1,8 @@
 package query
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"sync"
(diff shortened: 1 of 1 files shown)


jsternberg:
Updated the first comment with the backport PR number.

PR #9848, opened 2018-05-15
Fix broken unit tests after function signature change

jacobmarble:


hercules-influx:
``````

PR #9852, opened 2018-05-16
file store: send notifications about new/deleted tsm files.

zeebo:
just adds some interface for hooks about when these files come and go.
we do them before the action is taken so that if the hook has an
error, it doesn't have any consistency problems.

### DIFF
--- a/tsdb/engine.go
+++ b/tsdb/engine.go
@@ -185,6 +185,8 @@ type EngineOptions struct {
 	FieldValidator FieldValidator
 
 	OnNewEngine func(Engine)
+
+	FileStoreObserver FileStoreObserver
 }
 
 // NewEngineOptions returns the default options.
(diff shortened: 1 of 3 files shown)


hercules-influx:
```/tmp/346853302/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

tsdb/engine.go
```diff
@@ -201,3 +203,12 @@ func NewEngineOptions() EngineOptions {
 var NewInmemIndex func(name string, sfile *SeriesFile) (interface{}, error)
 
 type CompactionPlannerCreator func(cfg Config) interface{}
+
+// FileStoreObserver is passed notifications before the file store adds or deletes files.
```
jacobmarble: Please explain briefly why the notification is sent before the file change.

PR #9853, opened 2018-05-16
Fix the log2 function to return values

### LINKED ISSUE #9850: log2() function doesn't work, returns empty values
### Bug report

__System info:__ MacOS 10.13.4, InfluxDB (built from master on 15 May 2018)

__Steps to reproduce:__

1. Run a query using the `log2()` function

__Expected behavior:__ Should return log2 values for associated field values.

__Actual behavior:__ Returns empty values

---

_**Example using the NOAA sample data:**_
```
SELECT LOG2("water_level") FROM "h2o_feet" WHERE time >= '2015-08-18T00:00:00Z' AND time <= '2015-08-18T00:30:00Z' AND "location" = 'santa_monica'

name: h2o_feet
time                 log2
----                 ----
2015-08-18T00:00:00Z
2015-08-18T00:06:00Z
2015-08-18T00:12:00Z
2015-08-18T00:18:00Z
2015-08-18T00:24:00Z
2015-08-18T00:30:00Z
```

jsternberg:
The passed in argument wasn't correct so it always returned null instead
of the appropriate value.

Also includes unit tests for all of the math functions and restricts the
`asin()` and `acos()` functions to floats only since those functions
don't give any meaningful results when using integers or unsigned.

Fixes #9850.

### DIFF
--- a/query/math.go
+++ b/query/math.go
@@ -23,14 +23,60 @@ func (MathTypeMapper) MapType(measurement *influxql.Measurement, field string) i
 
 func (MathTypeMapper) CallType(name string, args []influxql.DataType) (influxql.DataType, error) {
 	switch name {
-	case "abs", "sin", "cos", "tan", "asin", "acos", "atan", "atan2", "exp", "log", "ln", "log2", "log10", "sqrt", "pow":
-		return influxql.Float, nil
-	case "floor", "ceil", "round":
-		switch args[0] {
+	case "sin", "cos", "tan", "atan", "exp", "log", "ln", "log2", "log10", "sqrt":
+		var arg0 influxql.DataType
+		if len(args) > 0 {
+			arg0 = args[0]
+		}
+		switch arg0 {
+		case influxql.Float, influxql.Integer, influxql.Unsigned, influxql.Unknown:
+			return influxql.Float, nil
+		default:
+			return influxql.Unknown, fmt.Errorf("invalid argument type for the first argument in %s(): %s", name, arg0)
+		}
+	case "asin", "acos":
+		var arg0 influxql.DataType
+		if len(args) > 0 {
+			arg0 = args[0]
+		}
+		switch arg0 {
+		case influxql.Float, influxql.Unknown:
+			return influxql.Float, nil
+		default:
+			return influxql.Unknown, fmt.Errorf("invalid argument type for the first argument in %s(): %s", name, arg0)
+		}
+	case "atan2", "pow":
+		var arg0, arg1 influxql.DataType
+		if len(args) > 0 {
+			arg0 = args[0]
+		}
+		if len(args) > 1 {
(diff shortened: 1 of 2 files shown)


mark-rushakoff:
Is there a test that should be added or updated here?

mark-rushakoff:
Are you saying you will add tests to this PR or a separate PR?

If it's this PR, I'll wait to approve. If it's a different PR that will happen later, I can approve this PR if you link back with an issue tracking the missing tests.

hercules-influx:
```/tmp/547485954/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

jsternberg:
@mark-rushakoff I've added the unit tests. I found an issue with `asin()` and `acos()` in the current implementation so I've updated those too and updated the description.

query/math.go
```diff
@@ -23,14 +23,59 @@ func (MathTypeMapper) MapType(measurement *influxql.Measurement, field string) i
...
+	case "sin", "cos", "tan", "atan", "exp", "log", "ln", "log2", "log10", "sqrt":
+		var arg0 influxql.DataType
+		if len(args) > 0 {
+			arg0 = args[0]
+		}
+		switch arg0 {
+		case influxql.Float, influxql.Integer, influxql.Unsigned, influxql.Unknown:
+			return influxql.Float, nil
+		default:
+			return influxql.Unknown, fmt.Errorf("invalid argument type for the first argument in %s(): %s", name, args[0])
```
mark-rushakoff: Looks like this should be `..., arg0)`, since earlier code implies len(args) could be zero. Occurs a couple times in this change.

query/math.go
```diff
@@ -23,14 +23,59 @@ func (MathTypeMapper) MapType(measurement *influxql.Measurement, field string) i
...
+		case influxql.Float, influxql.Integer, influxql.Unsigned, influxql.Unknown:
+		default:
+			return influxql.Unknown, fmt.Errorf("invalid argument type for the first argument in %s(): %s", name, args[0])
+		}
+
+		switch arg1 {
+		case influxql.Float, influxql.Integer, influxql.Unsigned, influxql.Unknown:
+			return influxql.Float, nil
+		default:
+			return influxql.Unknown, fmt.Errorf("invalid argument type for the second argument in %s(): %s", name, args[0])
```
mark-rushakoff: This one should be `arg1` rather than `args[0]`?

query/math.go
jsternberg: Fixed all of these.

hercules-influx:
```/tmp/216100182/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

PR #9860, opened 2018-05-16
rename "triple_exponential_average" -> "triple_exponential_derivative"

phemmer:
This renames the function `triple_exponential_average` to `triple_exponential_derivative` to more accurately reflect what it does. Ref https://github.com/influxdata/influxdb/pull/9260#issuecomment-389333924

Per the referenced discussion, instead of naming it "trix", as this is indeed what it is generally known as, I still opted for the long name due to consistency. What influx calls "average" is generally known as "sma" (simple moving average), and renaming this isn't an option, so I think consistency is more important.

@jsternberg 

### DIFF
--- a/query/call_iterator.go
+++ b/query/call_iterator.go
@@ -1311,29 +1311,29 @@ func newRelativeStrengthIndexIterator(input Iterator, n int, nHold int, warmupTy
 	}
 }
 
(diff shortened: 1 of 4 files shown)


jsternberg:
@sanderson can you mark this down to be changed in the docs? Can you also note that this algorithm is also called TRIX?

sanderson:
Will do. Thanks for the heads-up.

PR #9868, opened 2018-05-17
Close TSMReaders from FileStore.Close after releasing FileStore mutex

jacobmarble:
Backport of #9866 again fixing #9786


hercules-influx:
```/tmp/972882458/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

PR #9858, opened 2018-05-16
Return the correct auxiliary values for top/bottom

jsternberg:
When `top()` or `bottom()` were used and selected auxiliary values, they
would return the wrong values that would be equal to the last point
selected. This is because the aggregators saved the memory address of
the auxiliary fields instead of copying them over. Since the same
auxiliary fields memory location is used for every value returned by the
storage engine, this resulted in the values being incorrect because they
were overwritten with incorrect values.

This fixes that so the auxiliary fields are copied out when they are
saved rather than only the memory location.

Fixes #9197.

PR #9869, opened 2018-05-17
Return the correct auxiliary values for top/bottom

jsternberg:
When `top()` or `bottom()` were used and selected auxiliary values, they
would return the wrong values that would be equal to the last point
selected. This is because the aggregators saved the memory address of
the auxiliary fields instead of copying them over. Since the same
auxiliary fields memory location is used for every value returned by the
storage engine, this resulted in the values being incorrect because they
were overwritten with incorrect values.

This fixes that so the auxiliary fields are copied out when they are
saved rather than only the memory location.

Backport of #9858.

### DIFF
--- a/query/functions.go
+++ b/query/functions.go
@@ -1337,11 +1337,14 @@ func (r *FloatTopReducer) AggregateFloat(p *FloatPoint) {
 		if !r.h.cmp(&r.h.points[0], p) {
 			return
 		}
-		r.h.points[0] = *p
+		p.CopyTo(&r.h.points[0])
 		heap.Fix(r.h, 0)
 		return
 	}
-	heap.Push(r.h, *p)
+
+	var clone FloatPoint
+	p.CopyTo(&clone)
+	heap.Push(r.h, clone)
 }
 
 func (r *FloatTopReducer) Emit() []FloatPoint {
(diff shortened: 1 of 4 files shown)


hercules-influx:
```/tmp/934505966/src/github.com/influxdata/influxdb/tsdb/index/tsi1/log_file.go:139:2: 'if err != nil { return err }; return nil' can be simplified to 'return err' (S1013)
/tmp/934505966/src/github.com/influxdata/influxdb/query/iterator_test.go:1595:2: field point is unused (U1000)
/tmp/934505966/src/github.com/influxdata/influxdb/query/iterator_test.go:1625:2: field point is unused (U1000)
/tmp/934505966/src/github.com/influxdata/influxdb/query/iterator_test.go:1655:2: field point is unused (U1000)
/tmp/934505966/src/github.com/influxdata/influxdb/query/iterator_test.go:1685:2: field point is unused (U1000)
/tmp/934505966/src/github.com/influxdata/influxdb/query/iterator_test.go:1715:2: field point is unused (U1000)
```

stuartcarnie:
@jsternberg megacheck errors?

jsternberg:
I think I fixed those.

PR #9870, opened 2018-05-17
Fix the new top/bottom unit tests

jsternberg:
The new tests accidentally labeled the `p3` variable as the data type
rather than a string.


hercules-influx:
```/tmp/538746192/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

PR #9836, opened 2018-05-14
Performance optimization suggestions

huhue:
In my test, this optimization can improve performance 20%.  This optimization can make full use of my machine's CPU.
The big reason is that my test machine performance is better, 64-core cpu, ssd disk.
This optimization is what I found out through the flame graph, but sorry for the flame graph is lost.

I use influxdb own stress tools. stress config like this:
```
[provision]
  [provision.basic]
    enabled = true
    address = "xxxx:xxxx"
    database = "xxxx"
    reset_database = false

[write]
  [write.point_generator]
    [write.point_generator.basic]
      point_count = 1000
      series_count = 1000000
      tick = "10s"
      jitter = true
      precision = "s"
      measurement = "cpu"
      start_date = "2009-Jan-02"
      [[write.point_generator.basic.tag]]
        key = "host"
        value = "server"
      [[write.point_generator.basic.tag]]
        key = "location"
        value = "us-west"
      [[write.point_generator.basic.field]]
        key = "value"
        value = "float64" # supported types: float64, int, bool
      [[write.point_generator.basic.field]]
        key = "test_filed_int"
        value = "int" # supported types: float64, int, bool
      [[write.point_generator.basic.field]]
        key = "test_filed_bool"
        value = "bool" # supported types: float64, int, bool

  [write.influx_client]
    [write.influx_client.basic]
      enabled = true
      addresses = ["xxxx:xxxx"] 
      database = "stress"
      precision = "s"
      batch_size = 10000
      batch_interval = "0s"
      concurrency = 100
      ssl = false
      format = "line_http" 
      retention-policy = "xxxx"
```

PR #9885, opened 2018-05-21
Revert "Performance optimization suggestions"

stuartcarnie:
Reverts influxdata/influxdb#9836; need signed CLA


hercules-influx:
```/tmp/921052144/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

PR #9892, opened 2018-05-23
tsdb: observe tombstone files as well

zeebo:
I worry that I have tangled myself into a mess with the `WithObserver` pattern in order to avoid a mandatory, optional, parameter everywhere. It seems fragile to future changes in that one must consider if upon any opening of a TSMFile, if it may generate a tombstone. Open to suggestions.

### DIFF
--- a/tsdb/engine/tsm1/tombstone.go
+++ b/tsdb/engine/tsm1/tombstone.go
@@ -13,6 +13,8 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+
+	"github.com/influxdata/influxdb/tsdb"
 )
 
 const (
@@ -50,6 +52,9 @@ type Tombstoner struct {
 	pendingFile       *os.File
 	tmp               [8]byte
 	lastAppliedOffset int64
+
+	id  uint64
+	obs tsdb.FileStoreObserver
 }
 
 // Tombstone represents an individual deletion.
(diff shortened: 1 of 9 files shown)


hercules-influx:
```/tmp/916097944/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

tsdb/engine/tsm1/tombstone.go
```diff
@@ -50,6 +52,9 @@ type Tombstoner struct {
 	pendingFile       *os.File
 	tmp               [8]byte
 	lastAppliedOffset int64
+
+	id  uint64
```
jacobmarble: I notice that you added the shard id in FileStore as well. I suggest either add shard ID to the Tombstoner and FileStore constructor so that it's always available, or add it to the observer implementation, with a new instance per shard. I favor the latter.

hercules-influx:
```/tmp/603434334/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

hercules-influx:
```/tmp/825562176/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

PR #9896, opened 2018-05-23
fix(tsm1): Avoid searching index if key outside index bounds

stuartcarnie:
This improvement avoids performing a binary search on the index by
first checking the key against the lower and upper bounds. Particularly
useful for multiple, fully-compacted TSM files.


hercules-influx:
```/tmp/825581644/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

PR #9907, opened 2018-05-25
Fix the influx_inspect man page

jsternberg:


hercules-influx:
```/tmp/462425946/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

PR #9910, opened 2018-05-25
tsdb/tsm1: Clean up TSM filename format/parse

jacobmarble:
Making adjustments to the recently-implemented TSM filename format and parse code. Some of the parsers/formatters were hard-coded to the default implementation. This change cleans up the approach to just one: safe default and `With` methods to call from an `OnNewEngine` function.


hercules-influx:
```/tmp/195394350/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

hercules-influx:
```/tmp/614553744/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

PR #9921, opened 2018-05-29

### LINKED ISSUE #9904: influx_inspect buildtsi adds backslashes to spaces in measurement names
### Bug report

__System info:__ 1.5.2 and current master, MacOS 10.13 (and Linux?)

__Steps to reproduce:__

- Start InfluxDB 1.5.2 with inmem index (default for 1.5)
- `curl -i -XPOST http://localhost:8086/query --data-urlencode "q=CREATE DATABASE mydb"`
- `curl -i -XPOST 'http://localhost:8086/write?db=mydb' --data-binary 'cpu\ load\ short,host=server01,region=us-west value=0.64 1434055562000000000'`
- `show measurements` shows measurement name with spaces and no backslashes (expected)
- Stop InfluxDB
- `influx_inspect buildtsi --datadir=path/to/data --waldir=path/to/wal`
- Change InfluxDB config to `index-version = "tsi1"`
- Start InfluxDB

__Expected behavior:__

`show measurements` shows measurement name with spaces and no backslashes

__Actual behavior:__

```
> show measurements
name: measurements
name
----
cpu\ load\ short
```

__Additional info:__

Bug is likely in `buildtsi`.

Maybe a clue, WAL dump contains backslashes:
```
$ influx_inspect dumptsmwal before.wal
[write] sz=77
cpu\ load\ short,host=server01,region=us-west#!~#value 0.64 1434055562000000000
```

__Alternate reproduce steps:__

These steps reinforce the theory that the bug is in `buildtsi`.

- new database with TSI enabled
- insert with escaped spaces
- `show measurements` works as intended
- stop influxd
- delete TSI index, then rebuild the index
- start influxd
- show measurements shows escaped spaces

### Other side effects

When `buildtsi` is invoked against a TSI index with measurement names containing spaces, and then points are written to that same space-containing measurement, two measurements result. For example, `buildtsi` creates "a\ b" from measurement "a b", then subsequent writes to "a b" are written to a new measurement "a b", so "a\ b" and "a b" exist.

jacobmarble:
When `influx_inspect buildtsi` is used to create a new `tsi1` index, spaces in measurement names are escaped, so measurement "a b" is changed to "a\ b".

This change modifies `models.ParseKeyBytes()` and `models.ParseName()` to unescape measurement names. `models.ParseKeyBytes()` returns unescaped tag keys, so this seems like the natural place to unescape measurement names.

Also followed `scanMeasurement()` to see what other code could be problematic, and this should be everything (the result of one other use of `scanMeasurement()` is later escaped).

Removed `tsdb.MeasurementFromSeriesKey()`. These methods are exported, so checked for side effects in other InfluxData repositories.

After this change, existing TSI indexes that include backslashed measurements created by this bug will need to be rebuilt with `influx_inspect buildtsi`. This consolidates measurements "a b" and "a\ b" into just "a b".

Fixes #9904

### DIFF
--- a/models/points.go
+++ b/models/points.go
@@ -273,8 +273,8 @@ func ParsePointsString(buf string) ([]Point, error) {
 // NOTE: to minimize heap allocations, the returned Tags will refer to subslices of buf.
 // This can have the unintended effect preventing buf from being garbage collected.
 func ParseKey(buf []byte) (string, Tags) {
-	meas, tags := ParseKeyBytes(buf)
-	return string(meas), tags
+	name, tags := ParseKeyBytes(buf)
+	return string(name), tags
 }
 
 func ParseKeyBytes(buf []byte) ([]byte, Tags) {
@@ -282,27 +282,34 @@ func ParseKeyBytes(buf []byte) ([]byte, Tags) {
 	// when just parsing a key
 	state, i, _ := scanMeasurement(buf, 0)
 
+	var name []byte
 	var tags Tags
 	if state == tagKeyState {
 		tags = parseTags(buf)
 		// scanMeasurement returns the location of the comma if there are tags, strip that off
-		return buf[:i-1], tags
+		name = buf[:i-1]
+	} else {
+		name = buf[:i]
 	}
-	return buf[:i], tags
+	return unescapeMeasurement(name), tags
 }
 
 func ParseTags(buf []byte) Tags {
 	return parseTags(buf)
 }
 
-func ParseName(buf []byte) ([]byte, error) {
+func ParseName(buf []byte) []byte {
 	// Ignore the error because scanMeasurement returns "missing fields" which we ignore
 	// when just parsing a key
 	state, i, _ := scanMeasurement(buf, 0)
+	var name []byte
 	if state == tagKeyState {
-		return buf[:i-1], nil
+		name = buf[:i-1]
+	} else {
+		name = buf[:i]
 	}
-	return buf[:i], nil
+
+	return unescapeMeasurement(name)
 }
 
 // ParsePointsWithPrecision is similar to ParsePoints, but allows the
--- a/models/points_test.go
+++ b/models/points_test.go
@@ -2388,6 +2388,50 @@ func TestEscapeStringField(t *testing.T) {
 	}
 }
 
+func TestParseKeyBytes(t *testing.T) {
+	testCases := []struct {
+		input        string
+		expectedName string
+		expectedTags map[string]string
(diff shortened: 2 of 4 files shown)


jacobmarble:
Identified any side effects of this change.

`influx-tools import` (master)
- `seriesWriter.AddSeries` calls `models.ParseKeyBytes`
- works properly before and after change

`tsm1.Engine.deleteSeriesRange` calls `models.ParseKeyBytes`
- called by `delete from "a b"`
- called by `drop series from "a b"
- TSI - `Index.DropSeries`
  - called by `Engine.deleteSeriesRange`, but never hit when name contains space because of the current bug, but the input arg is exactly the same pointer as the call from 
`models.ParseKeyBytes`
- broken before change
  - calls `ParseKeyBytes` with escaped measurement name
  - Returned name used to call `SeriesFile.SeriesID`, still escaped
  - this calls `AppendSeriesKey` with escaped name
  - `SeriesFile.SeriesID` can't find the series, so returns zero
  - `show measurements` still lists the relevant measurement
  - FIXED

`influx_inspect buildtsi` - processShard AND processTSMFile call `models.ParseKey`
- `processShard` calls `ParseKey` to populate index from WAL/cache, uses returned escaped name to call `TSI Index.CreateSeriesIfNotExists`
- `processTSMFile` calls `ParseKey` to populate index from TSM, uses returned escaped name exactly as `processShard`
- resulting measurement name is escaped
- FIXED

`influx_inspect deletetsm` (master) - `Command.process` calls `ParseKey`
- `influx_inspect deletetsm -measurement "a b" ~/.influxdb/data/db/autogen/2/000000001-000000001.tsm`
- measurement not actually removed
- FIXED

`influx_inspect report` - `Command.Run` calls `models.ParseKey`
- `influx_inspect report -detailed ~/.influxdb/data/db/autogen/1/`
- escaped measurement names are printed
- FIXED

`tsm1.Engine.createVarRefSeriesIterator` calls `models.ParseKey`
- measurement name thrown away
- nothing to do

`tsdb.MeasurementFromSeriesKey` calls `models.ParseName`
- unescapes the result of `ParseName`
- only called from `Engine.addToIndexFromKey`, so safe to refactor
- REFACTORED

`show series` shows escaped measurement names
- working as intended?
- NOT FIXED

jacobmarble:
Spent some more time confirming each of the "FIXED" cases in my last comment, discovered a [bug in `deletetsm`](https://github.com/influxdata/influxdb/issues/9929), but otherwise everything checks out.

PR #9928, opened 2018-05-31
TSM: Enforce OpenLimiter must not be nil

jacobmarble:
Enterprise was failing to start because `EngineOptions.OpenLimiter` was `nil`, and reading from a `nil` channel blocks forever.


jacobmarble:
Fixes bug introduced in #9918. Follow up to this in #9933

PR #9930, opened 2018-06-01

jacobmarble:
Backport of #9921 fixing #9904


hercules-influx:
``````

PR #9932, opened 2018-06-01

jacobmarble:
Backport of #9921 fixing #9904


hercules-influx:
```/tmp/148615346/src/github.com/influxdata/influxdb/cmd/influx-tools/importer/importer.go:73:3: this value of rpi is never used (SA4006)
```

PR #9933, opened 2018-06-01
tests: Fix OpenLimiter is nil

jacobmarble:


jacobmarble:
Fixes test broken by #9928

PR #9884, opened 2018-05-21
Add -series-file flag to dumptsi command help.

benbjohnson:
###### Required only if applicable
- [x] Update man page when modifying a command

PR #9946, opened 2018-06-07
Functions percentage_between() and count_between() implemented

dkuk:

- [ ] [InfluxQL Spec](https://github.com/influxdata/influxdb/blob/master/influxql/README.md) updated (unavailable!)
- [ ] Provide example syntax 
- [x] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): issue filed or pull request submitted [functions docs](https://github.com/influxdata/docs.influxdata.com/pull/1667)


dkuk:
Looks like auto-checks does not works. Same test problem with pull-request https://github.com/influxdata/influxdb/pull/9884 - missing test files?

jsternberg:
cc @rbetts 

@dkuk thank you for your contribution. Is there anything that `count_between()` does that isn't done better by using a `WHERE` clause? It seems to me like it is duplicating that behavior since you could just do:

```
SELECT count(value) FROM ... WHERE value >= 0 AND value <= 5
```

This has more flexibility too.

On the other hand, `percentage_between()` is not something influxql can currently do so we are primarily looking at this PR because of that function. We'll get back to you on this PR soon. Thanks again.

dkuk:
It is not for simple cases when you need to get only one value with some conditions. So, yours example does not provide to calculate ratios for complex service level such as ratio of operations duration between 0 and 200ms to all requests (which is most common case implemented with percentage_between()), or correlation between very fast an abnormal slow operations:
```sql
SELECT count("duration",10,3600)/count_between("duration", 0, 0.2) FROM "operations"
```
I agreed with that is not so common case, as percentage, but another simple example is a trigger for percentage of slow queries with condition on absoltue number of such queries:

```sql
SELECT count("request_time",10,600) as "num_slow", percentage_between("request_time",10,600) FROM "requests"
```
Trigger alert only when number of problems exceeds N and percentage is greater than X.

jsternberg:
Ah, so the issue is mostly that you want to be able to run a different condition on two different functions used together.

We'll try to take a look at this. The feature set for 1.6 is mostly frozen but we'll take a look at this when we start on the 1.7 cycle. This will also be built-in behavior that will be easy to do in the new flux language that we are developing.

dkuk:
Is it possible to bring this functions into 1.6? 
Is the different conditions per defferent fuctions confirmed for 1.7 release and the roadmap for this release defined?

rbetts:
@nathanielc @jsternberg This needs an estimate of effort for the flux transpiler. If that is sane, we want to land this for 1.7.0.

nathanielc:
This is possible to implement in Flux and the transpiler would see the count_between function as a `filter|>count` pair. Same for `percentage_between`.

PR #9959, opened 2018-06-11
Improve the number of regex patterns that are optimized to static OR conditions

### LINKED ISSUE #9574: Improve RewriteRegexConditions to handle regex OR
The current [`RewriteRegexConditions`][rewrite] API handles the following case

* `tag =~ /^foo$/`

and rewrites it to

* `tag = foo`

allowing the use of a faster index search. However, dashboards generate more complex regular expressions that could also be rewritten for improved performance. Specifically

* `tag =~ /^foo$|^bar$|^fee$/`
* `tag =~ /^(foo|bar|fee)$/`

could be rewritten to

* `tag = foo OR tag = bar OR tag = fee`

Again allowing faster index lookups.

When updating this API, it should be moved to a top-level function so it can be reused for all conditions rather than just those attached to a select statement.

[rewrite]: https://github.com/influxdata/influxql/blob/master/ast.go#L1371-L1411

jsternberg:
See influxdata/influxql#21 for details.

Fixes #9574.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -31,10 +31,6 @@
   revision = "d9eb7a3d35ec988b8585d4a0068e462c27d28380"
 
 [[constraint]]
-  name = "github.com/influxdata/influxql"
-  revision = "145e0677ff6418fa00ee7e5dd434305631ab44ea"
-
-[[constraint]]
   name = "github.com/mattn/go-isatty"
   branch = "master"
 
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -113,7 +113,7 @@
     ".",
     "internal"
   ]
-  revision = "145e0677ff6418fa00ee7e5dd434305631ab44ea"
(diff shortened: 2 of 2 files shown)


Gopkg.toml
```diff
@@ -32,7 +32,7 @@
 
 [[constraint]]
   name = "github.com/influxdata/influxql"
-  revision = "145e0677ff6418fa00ee7e5dd434305631ab44ea"
+  revision = "8d2756c43bb02f5d3e505c51cf83a1ac9234c419"
```
e-dard: We should change this to `branch = master`.

jsternberg: This is always set to an exact revision so the same revision is inherited by the closed source version.

jsternberg: I've removed this constraint. I'll create something in the closed source to watch for these kinds of things.

PR #10007, opened 2018-06-27
allow tag keys to contain underscores

### LINKED ISSUE #9522: WHERE matches always with SHOW TAG VALUES if tag key begins with underscore
I came across with this behavior when trying to use the template facilities of Grafana while using data stored in InfluxDB by the Prometheus interface. The interface stores the names of the metrics in the tag key `__name__`, so it would occasionally make sense to retrieve, for example, all "instance" values that have a certain series defined by the method described below.

I figure names like these might be interfering with some internal mechanisms, but I couldn't find any way to quote the tag key in such a way that it would work either. If this is by design, then the naming established in the Prometheus interface might require reworking, and that is a completely another story.

__System info:__
InfluxDB 1.4.2 on Centos 7 x86_64 from official InfluxDB repository.

__Steps to reproduce:__

1. Create measurements with a tag with underscores
```
> select * from testseries
name: testseries
time       __name__ meas uniq
----       -------- ---- ----
1439856000 bar      1    ccc
1439856000 bar      1    ddd
1439856000 foo      1    aaa
1439856000 foo      1    bbb
```
2.  Try to show tag values
```
> SHOW TAG VALUES FROM testseries WITH KEY = "uniq" WHERE "__name__" = 'bar'
```

__Expected behavior:__
Values of tag key "uniq" where tag `__name__` equals "bar" are returned.
```
name: testseries
key  value
---  -----
uniq ccc
uniq ddd
```
__Actual behavior:__
All possible values for the tag key "uniq" are returned. The result is always the same if the tag key after WHERE begins with an underscore, eg. `WHERE "_asdfa" = ...`
```
name: testseries
key  value
---  -----
uniq aaa
uniq bbb
uniq ccc
uniq ddd
```

dgnorton:
This PR changes the `tsdb/store` to call `influxql.IsSystemName(tagKey)` instead of filtering out all tag keys that contain a leading `_`. Fixes #9522

### DIFF
--- a/tsdb/store.go
+++ b/tsdb/store.go
@@ -1500,7 +1500,7 @@ func (s *Store) TagValues(auth query.Authorizer, shardIDs []uint64, cond influxq
 			switch e.Op {
 			case influxql.EQ, influxql.NEQ, influxql.EQREGEX, influxql.NEQREGEX:
 				tag, ok := e.LHS.(*influxql.VarRef)
-				if !ok || strings.HasPrefix(tag.Val, "_") {
+				if !ok || influxql.IsSystemName(tag.Val) {
 					return nil
 				}
 			}


e-dard:
@dgnorton can you try rebasing on master and re-pushing?

hercules-influx:
```/tmp/141104795/src/github.com/influxdata/influxdb/cmd/store/query/query.go:390:5: this value of line is never used (SA4006)
```

PR #10013, opened 2018-06-28
Filter iterators as high up a possible

e-dard:
Previously we were filtering iterators at a very low level, which meant we were filtering series that would have been merged out by higher iterators anyway.

Filtering is expensive because it involves locking on the series file partitions, which has a fixed cost of ~35ns per seriesID checked. For millions of series that can expensive at planning time.

This PR moves the filtering to as high a level as possible.

### DIFF
--- a/tsdb/index.go
+++ b/tsdb/index.go
@@ -1629,7 +1629,7 @@ func (is IndexSet) measurementSeriesIDIterator(name []byte) (SeriesIDIterator, e
 			a = append(a, itr)
 		}
 	}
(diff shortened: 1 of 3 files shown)


tsdb/index.go
```diff
@@ -1768,6 +1790,9 @@ func (is IndexSet) MeasurementSeriesKeysByExpr(name []byte, expr influxql.Expr)
 	}
 	defer itr.Close()
 
+	// measurementSeriesByExprIterator filters deleted series; no need to do so
+	// here.
```
benbjohnson: `"here"` looks lonely.

Issue #10027, opened 2018-07-03
tsi inverted index may lose index data when server crash


huhue:
__System info:__ [InfluxDB 1.5, Darwin Kernel Version 16.7.0]

__Steps to reproduce:__

1. create new series 
2. series data persist in tsl file
3. tsl log file use bufio.Writer module, call bufio.Writer.Write() and bufio.Writer will buffer the series index data, then if server crash, will lose up to 4KB bytes data. 

__Additional info:__ [Include gist of relevant config, logs, etc.]

I think need do bufio.Writer.Flush() in every write case. 
Strictly speaking, I think may need do f.Sync() like wal writer.

e-dard:
Thanks for the issue and related PR @huhue!

This is has already been fixed in #9982, and will be part of the 1.6 release.

PR #9982, opened 2018-06-19
Flush/sync TSI1 WAL

benbjohnson:
Flushes the `bufio.Writer` and fsync's to disk after writes.

Fixes #9863
Fixes #9957
Fixes #9987

PR #10029, opened 2018-07-03
suggestion for prevent lose inverted index data when server crash

huhue:

#10027


e-dard:
Thanks for the PR @huhue. This is has already been fixed in #9982, and will be part of the 1.6 release.

Issue #10023, opened 2018-07-02
Offline tooling to better understand TSI cardinality


e-dard:
__Proposal:__

To write a `reportsi` tool that will walk one or more TSI indexes in a database and produce statistics about the cardinality of series for measurements or tag keys, on a per-shard basis.

__Current behavior:__ 

There is no tooling to do this. We have TSM-level tooling that can do something similar but (1) it either uses estimations or it allocates a lot to count exactly; (2) it's slow because it scan TSM files; (3) it can't count across shards

__Desired behavior:__

 - see top n measurements by cardinality for each shard in the database, and overall.
 - see top n tag keys by cardinality for each shard in the database, and overall.
 - work on TSI indexes to leverage performance.
 - Use bitmaps for accurate yet heap efficient counting.

__Use case:__ 

Increasingly we see uses or customers who have an unknown increase in cardinality, and it's difficult to track down what is responsible for this cardinality. We have `SHOW SERIES CARDINALITY` and friends, but the queries don't work on a per-shard basis. If you run `SHOW SERIES CARDINALITY FROM mymeasurement` then it's incredibly slow because it uses the query engine.

With this tool, it would be possible to see easily what measurements or tag keys are responsible for cardinality, and which shards introduced the issue.

PR #10032, opened 2018-07-03
Add new TSI index cardinality report to influx_inspect 

e-dard:
Resolves https://github.com/influxdata/influxdb/issues/10023

This PR adds a new tool `reporttsi` to `influx_inspect`. You run it as follows:

```
$ influx_inspect reporttsi -db-path ~/.influxdb/data/stress -top 10
```

The report does the following:

  - calculates the total exact series cardinality in the database;
  - segments that cardinality by measurement, and emits those cardinality values;
  - optionally limits the result to the "top n";
  - Emits total exact cardinality for each shard in the database;
  - Segments for each shard the exact cardinality for each measurement in the shard;
  - optionally limits the results in each shard to the "top n".

This tool is primarily of use when there has been a change in cardinality, and it's not clear which measurement is responsible for this change, and further, roughly _when_ that change happened.

Emitting an accurate cardinality breakdown for each measurement, and for each shard will help answer those questions.

#### Performance

  - The tool uses simple slice/maps to store low cardinality measurements, which saves on the cost of initialising bitmaps.
  - For high cardinality measurements the tool uses roaring bitmaps, which means we don't need to store all series ids on the heap while running the tool.
 - Conversion from low-cardinality to high-cardinality representations is done automatically while the tool runs.

I tested the tool on a dataset of 100M series across 25 shards. The tool completed in ~11 seconds, which means it worked in that case at approximately `9M series / second`.

#### Todo

  - Add support for segmenting by tag key instead of measurement.
  - Allow tool to run on a sub-set of shards for the database rather than all of them.

### DIFF
--- a/cmd/influx_inspect/reporttsi/report.go
+++ b/cmd/influx_inspect/reporttsi/report.go
@@ -159,9 +159,9 @@ func (cmd *Command) run() error {
 
 	// Calculate cardinalities of shards.
 	fn := cmd.cardinalityByMeasurement
-	if cmd.byTagKey {
-		// TODO(edd)
-	}
+	// if cmd.byTagKey {
+	// TODO(edd)
+	// }
 
 	// Blocks until all work done.
 	cmd.calculateCardinalities(fn)
@@ -290,7 +290,7 @@ OUTER:
 		var e tsdb.SeriesIDElem
 		for e, err = sitr.Next(); err == nil && e.SeriesID != 0; e, err = sitr.Next() {
 			if e.SeriesID > math.MaxUint32 {
-				panic(fmt.Sprintf("series ID is too large: %d (max %d). Corrupted series file?", e.SeriesID, math.MaxUint32))
+				panic(fmt.Sprintf("series ID is too large: %d (max %d). Corrupted series file?", e.SeriesID, uint32(math.MaxUint32)))
 			}
 			c.add(e.SeriesID)
 		}
--- a/cmd/influx_inspect/reporttsi/report.go
+++ b/cmd/influx_inspect/reporttsi/report.go
@@ -21,6 +21,12 @@ import (
 	"github.com/influxdata/influxdb/tsdb/index/tsi1"
 )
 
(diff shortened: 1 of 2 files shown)


hercules-influx:
```/tmp/033724926/src/github.com/influxdata/influxdb/cmd/influx_inspect/reporttsi/report.go:162:2: empty branch (SA9003)
```

cmd/influx_inspect/reporttsi/report.go
```diff
@@ -0,0 +1,486 @@
...
+	}
+}
+
+// Run executes the command.
+func (cmd *Command) Run(args ...string) error {
+	fs := flag.NewFlagSet("reporttsi", flag.ExitOnError)
+	fs.StringVar(&cmd.dbPath, "db-path", "", "Path to database. Required.")
+	fs.StringVar(&cmd.seriesFilePath, "series-file", "", "Optional path to series file. Defaults /path/to/db-path/_series")
+	fs.BoolVar(&cmd.byMeasurement, "measurements", true, "Segment cardinality by measurements")
+	fs.BoolVar(&cmd.byTagKey, "tag-key", false, "Segment cardinality by tag keys (overrides `measurements`")
```
stuartcarnie: Perhaps worth hiding this switch altogether until it is implemented?

cmd/influx_inspect/reporttsi/report.go
```diff
@@ -0,0 +1,486 @@
...
+	"github.com/influxdata/influxdb/tsdb/index/tsi1"
+)
+
+const (
+	// Number of series IDs to stored in slice before we convert to a roaring
+	// bitmap. Roaring bitmaps have a non-trivial initial cost to construct.
+	useBitmapN = 25
+)
+
+// Command represents the program execution for "influxd dumptsi".
```
zeebo: nit: s/dumptsi/reporttsi/

e-dard:
Feedback addressed in 08c3800.

PR #10044, opened 2018-07-05
Implement bitset iterator

benbjohnson:
This pull request implements iterators based on `SeriesIDSet`. This allows faster merging of large series sets. This improves planning time by approximately 20%, however, a large amount of time is still spent building the bitsets. This change, along with changes from @e-dard, and some additional caching should help remove the bitset building time.

### DIFF
--- a/tsdb/index.go
+++ b/tsdb/index.go
@@ -164,6 +164,53 @@ type SeriesIDIterator interface {
 	Close() error
 }
 
+// SeriesIDSetIterator represents an iterator that can produce a SeriesIDSet.
+type SeriesIDSetIterator interface {
+	SeriesIDIterator
+	SeriesIDSet() *SeriesIDSet
+}
+
+type seriesIDSetIterator struct {
+	ss  *SeriesIDSet
+	itr SeriesIDSetIterable
+}
+
+func NewSeriesIDSetIterator(ss *SeriesIDSet) SeriesIDSetIterator {
+	if ss == nil || ss.bitmap == nil {
+		return nil
+	}
+	return &seriesIDSetIterator{ss: ss, itr: ss.Iterator()}
+}
+
+func (itr *seriesIDSetIterator) Next() (SeriesIDElem, error) {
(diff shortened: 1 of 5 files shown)


hercules-influx:
```/tmp/121093881/src/google.golang.org/genproto/googleapis/rpc/status/status.pb.go:111:28: InternalMessageInfo not declared by package proto
couldn't load packages due to errors: google.golang.org/genproto/googleapis/rpc/status
```

tsdb/index.go
```diff
@@ -9,6 +9,7 @@ import (
 	"sort"
 	"sync"
 
+	"github.com/RoaringBitmap/roaring"
```
e-dard: Can we not dip into the implementation of `SeriesIDSet`? If we need something from the `roaring.Bitmap` we could maybe just add it to `SeriesIDSet`? It will make future optimisations to the `SeriesIDSet` a bit easier.

tsdb/series_set.go
```diff
@@ -126,6 +135,15 @@ func (s *SeriesIDSet) AndNot(other *SeriesIDSet) *SeriesIDSet {
 	return &SeriesIDSet{bitmap: roaring.AndNot(s.bitmap, other.bitmap)}
 }
 
+// Or returns a new SeriesIDSet containing elements that were present in s or other.
+func (s *SeriesIDSet) Or(other *SeriesIDSet) *SeriesIDSet {
```
e-dard: It seems like `a.Or(b)` is equivalent to:

```go
ss := NewSeriesIDSet()
ss.Merge(a, b)
```

Since we already have a `Merge` method available. Maybe this is faster than `Merge` as we don't need to allocate a `[]*roaring.Bitmap`?

benbjohnson: Fixed in 3b77f18.

tsdb/index.go
benbjohnson: Fixed in 979d790.

e-dard:
🎉

PR #10046, opened 2018-07-05
httpd: fix flaky test in timeout handler

zeebo:
there were two problems with this code:

1. the send on pending did not imply that the handler was running
2. there was a race starting the handler with timing out

1 is fixed by sending to a begin channel inside the handler. it is
then guaranteed that the timeout handler code has been entered.

2 is fixed by attempting to acquire the semaphore channel once before
checking the timeout channel. in this way, if there is capacity, which
in this test there is known to be, it is guaranteed to be taken. if
we check with the timer at the same time and the timer has already
fired, there is a pseudorandom chance the timer will be taken even
if there is capacity.

### DIFF
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -1703,13 +1703,19 @@ func (t *Throttler) Handler(h http.Handler) http.Handler {
 			}
 		}
 
-		// Wait for a spot in the list of concurrent requests.
+		// First check if we can immediately send in to current because there is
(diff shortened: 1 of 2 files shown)


hercules-influx:
```/tmp/123059649/src/google.golang.org/genproto/googleapis/rpc/status/status.pb.go:111:28: InternalMessageInfo not declared by package proto
couldn't load packages due to errors: google.golang.org/genproto/googleapis/rpc/status
```

PR #10048, opened 2018-07-05
Improve performance of buildtsi tool / series creation

### LINKED ISSUE #10019: influx_inspect too slow
I ran `influx_inspect buildtsi` on a single shard/tsm file, which was about 1.5GB, and it still hadn't processed it after 2 hours. Something going on..

Further, when using `buildtsi` in internal clusters I am seeing it sometimes only working at around 6GB a minute, which is only about 100MB/s.

e-dard:
Fixes #10019 

This PR significantly improves the performance of `influx_inspect buildtsi`. It makes building a TSI index around 4-10 times faster than 1.5.3, and 17–46 times faster than current `master`. 😃 

#### New tool features

 - Shards will be processed concurrently (adjustable with `-c`);
 - Series batch size configurable. This is an "expert" option, and users should typically stick with the default value, which is 10,000; and
 - Much. More. Fasterer.

#### Test data

I wrote 100,000,000 series with inch across 25 shards.

  - 4 measurements
  - 5 tag keys

```
inch -b 10000 -c 5 -t 10,10,10,100,1000 -f 1 -m 4 -p 1 -start-time '2018-07-02T00:00:00Z' -time 100h  -v
```

#### TLDR Results

These results show the time taken to build the TSI index with a pre-existing series file. Most of the performance boost comes from changes to how the index is built so I focussed on that results-wise:

| Description | master time | PR time| time Δ  | master series/s | PR series/s | series/s Δ |
| ---: | :---: | :---: | :---: | :---: | :---: | :---: |
| single shard (4M series) |  8m 15s | 28s | ~94% |  9K  | ~143K  | ~94% |
| 25 shards (100M series) |  ~200 min | 4m 20s | ~98% |  ~8.3K  | ~﻿384K  | ~98% |

N.B these results were against `master`. 1.5.3 might not be _as bad_ as the master results because it doesn't contain #9982, which hurts `buildtsi` performance. 

#### Breakdown

There are some other useful commits in the set, but these are the ones that made the most difference:

 - f6facb5 and 2c40e80: we have to flush buffers and syncing files on every write to a `LogFile` for durability reasons. However, we don't need to do this for offline tooling since in the worst-case failure, we can just re-run the tool. Let's turn that off! 
 - 10a45f3: rather than reading series one at a time out of a TSM file, we can read a batch and then send them all at once to be written to the index. 10,000 is a good, empirically determined, default batch size.
 - f78ff85: A `LogEntry` is 11-12 bytes per series. If you write a batch of 10,000 new series that's going to involve three fsyncs. By making the buffer configurable for the `buildtsi` tool we can reduce the frequency of these syscalls.
 - 4792ba4: We maintain inverted indexes for `measurement_name -> all_series_with_name` amd `tag_key_value -> all_series_with_tag_key_value`. Rather than storing the sets of series ids for these indexes in maps, we use bitmaps. This reduces their size and improves performance when we need to pull them out in order. This will significantly help with TSI query planning too. Note: bitsets have a non-trivial cost for low cardinality sets (`~1000 ns` and `~15` allocs),  so we don't always use them.
 - 0bf0180: re-use tag containers... This one I need some scrutiny on. I'm almost certain that it is OK to do this for the offline tooling, but please take a close look @benbjohnson @stuartcarnie.... It made a massive improvement: `~37s -> ~28s`.
 - a85e951 concurrent shard index building. This will help for machines with many cores. How it helps with machines with fewer than 8 OS-threads is unclear because the series file and index already use multiple cores during insertion anyway.

#### Todo 

 I want to abstract a bunch of the "when to use bitsets and when not to" into the `tsdb.SeriesIDSet`. I will do that in a future PR.

### DIFF
--- a/cmd/influx_inspect/buildtsi/buildtsi.go
+++ b/cmd/influx_inspect/buildtsi/buildtsi.go
@@ -279,14 +279,14 @@ func (cmd *Command) processShard(sfile *tsdb.SeriesFile, dbName, rpName string, 
 
 	for _, key := range cache.Keys() {
 		seriesKey, _ := tsm1.SeriesAndFieldFromCompositeKey(key)
-		name, tags := models.ParseKey(seriesKey)
+		name, tags := models.ParseKeyBytes(seriesKey)
 
 		if cmd.Verbose {
 			cmd.Logger.Info("Series", zap.String("name", string(name)), zap.String("tags", tags.String()))
 		}
 
 		keysBatch = append(keysBatch, seriesKey)
-		namesBatch = append(namesBatch, []byte(name))
+		namesBatch = append(namesBatch, name)
 		tagsBatch = append(tagsBatch, tags)
 
 		// Flush batch?
@@ -347,14 +347,14 @@ func (cmd *Command) processTSMFile(index *tsi1.Index, path string) error {
 	for i := 0; i < r.KeyCount(); i++ {
 		key, _ := r.KeyAt(i)
 		seriesKey, _ := tsm1.SeriesAndFieldFromCompositeKey(key)
-		name, tags := models.ParseKey(seriesKey)
+		name, tags := models.ParseKeyBytes(seriesKey)
 
 		if cmd.Verbose {
 			cmd.Logger.Info("Series", zap.String("name", string(name)), zap.String("tags", tags.String()))
 		}
 
 		keysBatch = append(keysBatch, seriesKey)
-		namesBatch = append(namesBatch, []byte(name))
+		namesBatch = append(namesBatch, name)
 		tagsBatch = append(tagsBatch, tags)
 
 		// Flush batch?
--- a/cmd/influx_inspect/buildtsi/buildtsi.go
+++ b/cmd/influx_inspect/buildtsi/buildtsi.go
@@ -232,8 +232,13 @@ func (cmd *Command) processShard(sfile *tsdb.SeriesFile, dbName, rpName string, 
 	}
 
 	// Open TSI index in temporary path.
-	tsiIndex := tsi1.NewIndex(sfile, dbName, tsi1.WithPath(tmpPath), tsi1.WithMaximumLogFileSize(cmd.maxLogFileSize))
+	tsiIndex := tsi1.NewIndex(sfile, dbName,
+		tsi1.WithPath(tmpPath),
+		tsi1.WithMaximumLogFileSize(cmd.maxLogFileSize),
+		tsi1.DisableFsync())
+
 	tsiIndex.WithLogger(cmd.Logger)
+
 	cmd.Logger.Info("Opening tsi index in temporary location", zap.String("path", tmpPath))
 	if err := tsiIndex.Open(); err != nil {
 		return err
--- a/cmd/influx_inspect/buildtsi/buildtsi.go
+++ b/cmd/influx_inspect/buildtsi/buildtsi.go
@@ -6,9 +6,12 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"os"
 	"os/user"
 	"path/filepath"
+	"runtime"
+	"runtime/pprof"
 	"strconv"
 	"strings"
 
@@ -61,7 +64,46 @@ func (cmd *Command) Run(args ...string) error {
 	}
 	cmd.Logger = logger.New(cmd.Stderr)
 
+	finish := startProfiles()
+	defer finish()
 	return cmd.run(*dataDir, *walDir)
+}
+
+func startProfiles() func() {
+	runtime.MemProfileRate = 100 // Sample 1% of allocations.
+
+	paths := []string{"/tmp/buildtsi.mem.pprof", "/tmp/buildtsi.cpu.pprof"}
+	var files []*os.File
+	for _, pth := range paths {
+		f, err := os.Create(pth)
+		if err != nil {
+			log.Fatalf("memprofile: %v", err)
+		}
+		log.Printf("writing profile to: %s\n", pth)
+		files = append(files, f)
+
+	}
+
+	closeFn := func() {
+		// Write the memory profile
+		if err := pprof.Lookup("heap").WriteTo(files[0], 0); err != nil {
+			panic(err)
+		}
+
+		// Stop the CPU profile.
+		pprof.StopCPUProfile()
+
+		for _, fd := range files {
+			if err := fd.Close(); err != nil {
+				panic(err)
+			}
+		}
+	}
+
+	if err := pprof.StartCPUProfile(files[1]); err != nil {
+		panic(err)
+	}
+	return closeFn
 }
 
 func (cmd *Command) run(dataDir, walDir string) error {
--- a/cmd/influx_inspect/buildtsi/buildtsi.go
+++ b/cmd/influx_inspect/buildtsi/buildtsi.go
@@ -23,7 +23,7 @@ import (
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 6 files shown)


e-dard:
Nit: I forgot to set a limit of 128K on the log buffer size in f78ff85. I will add a commit to the set to do that.

Incidentally, 128K seems to work pretty well for me, but I'm not sure if it's maybe at bit high?

cmd/influx_inspect/buildtsi/buildtsi.go
```diff
@@ -61,9 +71,50 @@ func (cmd *Command) Run(args ...string) error {
 	}
 	cmd.Logger = logger.New(cmd.Stderr)
 
+	// Uncomment for profiling
+	// finish := startProfiles()
+	// defer finish()
```
benbjohnson: Can this be a CLI flag or removed?

cmd/influx_inspect/buildtsi/buildtsi.go
```diff
@@ -61,9 +71,50 @@ func (cmd *Command) Run(args ...string) error {
...
+// 			}
+// 		}
+// 	}
+
+// 	if err := pprof.StartCPUProfile(files[1]); err != nil {
+// 		panic(err)
+// 	}
+// 	return closeFn
+// }
+
```
benbjohnson: Delete or uncomment.

cmd/influx_inspect/buildtsi/buildtsi.go
```diff
@@ -133,13 +184,18 @@ func (cmd *Command) processDatabase(dbName, dataDir, walDir string) error {
...
 
 	fis, err := ioutil.ReadDir(dataDir)
 	if err != nil {
 		return err
 	}
 
+	var shards []struct {
+		ID   uint64
+		Path string
+	}
```
benbjohnson: Can you create a type for this struct so it doesn't have to be repeated below in the `append()`.

cmd/influx_inspect/buildtsi/buildtsi.go
```diff
@@ -152,25 +208,50 @@ func (cmd *Command) processRetentionPolicy(sfile *tsdb.SeriesFile, dbName, rpNam
...
+				i := int(atomic.AddUint32(&maxi, 1) - 1) // Get next partition to work on.
+				if i >= len(shards) {
+					return // No more work.
+				}
+
+				id, name := shards[i].ID, shards[i].Path
+				errC <- cmd.processShard(sfile, dbName, rpName, id, filepath.Join(dataDir, name), filepath.Join(walDir, name))
+			}
+		}()
+	}
```
benbjohnson: Is this something we can generalize? It seems like a common approach we do all over the place.

models/points.go
```diff
@@ -1562,20 +1565,33 @@ func walkFields(buf []byte, fn func(key, value []byte) bool) {
...
+
 	// Series keys can contain escaped commas, therefore the number of commas
 	// in a series key only gives an estimation of the upper bound on the number
 	// of tags.
-	tags := make(Tags, 0, bytes.Count(buf, []byte(",")))
+	var i int
 	walkTags(buf, func(key, value []byte) bool {
-		tags = append(tags, Tag{Key: key, Value: value})
+		dst[i].Key, dst[i].Value = key, value
+		// tags = append(tags, Tag{Key: key, Value: value})
```
benbjohnson: Remove line?

tsdb/index/tsi1/log_file.go
```diff
@@ -265,7 +288,10 @@ func (f *LogFile) MeasurementHasSeries(ss *tsdb.SeriesIDSet, name []byte) bool {
 	if !ok {
 		return false
 	}
-	for id := range mm.series {
+
+	// TODO(edd): if mm is using a bitmap then this could be changed to do a fast
+	// intersection.
```
benbjohnson: `s/\nintersection/intersection/`

tsdb/index/tsi1/log_file.go
```diff
@@ -739,11 +741,25 @@ func (f *LogFile) SeriesIDIterator() tsdb.SeriesIDIterator {
 	defer f.mu.RUnlock()
 
 	ss := tsdb.NewSeriesIDSet()
+	allBitMaps := make([]*tsdb.SeriesIDSet, 0, len(f.mms))
```
benbjohnson: Can we `s/bitmap/seriesSet/` for consistent naming?

tsdb/index/tsi1/log_file.go
```diff
@@ -801,8 +817,8 @@ func (f *LogFile) CompactTo(w io.Writer, m, k uint64, cancel <-chan struct{}) (n
 	default:
 	}
 
-	// Wrap in bufferred writer.
-	bw := bufio.NewWriter(w)
+	// Wrap in bufferred writer with a buffer equivalent to the LogFile size.
+	bw := bufio.NewWriterSize(w, 1<<17) // 128K
```
benbjohnson: Can we move this number to a constant?

tsdb/index/tsi1/log_file.go
```diff
@@ -1177,30 +1193,96 @@ type logMeasurement struct {
...
+	m.series[x] = struct{}{}
+
+	// If the map is getting too big it can be converted into a roaring bitmap.
+	if len(m.series) > 25 {
+		m.bitmap = tsdb.NewSeriesIDSet()
+		for id := range m.series {
+			m.bitmap.AddNoLock(id)
+		}
+		m.series = nil
+	}
```
benbjohnson: I'd be curious to see what the impact is in practice for `<25` series sets. It seems like it's going to make the code a lot more complicated by handling both cases.

cmd/influx_inspect/buildtsi/buildtsi.go
e-dard: I think so yeah, but should probably do it in a future PR and apply it to the entire codebase.

tsdb/index/tsi1/log_file.go
e-dard: I have some benchmarks somewhere. I plan on abstracting all this away into the `SeriesIDSet` implementation, so we can make changes once and they'll apply everywhere.

e-dard:
@benbjohnson please see 22bd3b5

models/points.go
```diff
@@ -1562,20 +1565,33 @@ func walkFields(buf []byte, fn func(key, value []byte) bool) {
...
-func parseTags(buf []byte) Tags {
+// parseTags parses buf into the provided destination tags, returning destination
+// Tags, which may have a different length and capacity.
+func parseTags(buf []byte, dst Tags) Tags {
 	if len(buf) == 0 {
 		return nil
 	}
 
+	n := bytes.Count(buf, []byte(","))
+	if len(dst) < n {
```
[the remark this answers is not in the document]
e-dard: This is preferable yeah.

models/points.go
```diff
@@ -1562,20 +1565,33 @@ func walkFields(buf []byte, fn func(key, value []byte) bool) {
...
+		dst = make(Tags, n)
+	}
+
+	if dst == nil { // Ensure existing behaviour.
```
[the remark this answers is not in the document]
e-dard: It's to cover the case you provide a `nil` slice and the point has no tags. I wanted to keep the existing behaviour.

e-dard:
Final performance improvements on a box with 24 OS threads:

| Description | v1.5.3 time | PR time| time Δ 
| ---: | :---: | :---: | :---: | 
| single shard (4M series) |  2m 2s | 12s | ~90% | 
| 25 shards (100M series) |  ~51m 40s | 1m 15s | ~97.5% |

PR #10050, opened 2018-07-05
Resolve deadlock deleting from many measurements concurrently

### LINKED ISSUE #9636: [bug] drop/delete terribly slow
### Bug report

Influx version: **v1.5.1**
OS version: **Ubuntu Linux 17.04**

__Steps to reproduce:__
1. Have a big database with > 20000 series
2. Try to execute `drop series from /{{regexp_pattern}}/`.
3. or try to execute `drop measurement /{{regexp_pattern}}/`.
4. or try to execute `delete from /{{regexp_pattern}}/`.

It will take > 30 minutes.

__Expected behavior:__ 
1. Well deleting data should take miliseconds?
2. Or there should be some special command modifier that will do it fast (I think this is becaus esome influx DB cleaning stuff). And there should be another comment to execute this "long" part manually, so for example I can delete ton of series and run cleanup process manually (to reclaim space) from a cron job or something like this.

jacobmarble:
TSI LogFile compactions occasionally race with insert and delete
operations because the index partition FileSet is retained needlessly by
the method that calls Partition.CheckLogFile.

In this change:
- TSI LogFile compaction respects enable/disable compactions
- Partition FileSet.Release before log compaction is triggered

An alternative to the second step is to handle log file compaction in a
new goroutine. Log file compaction errors would be logged and not
returned to the caller.

After this change, `DELETE FROM /regex/` does not deadlock; performance:
- 30s to delete 100 measurements
- 5m30s to delete 1000 measurements

Fixes #9636

### DIFF
--- a/tsdb/index/tsi1/partition.go
+++ b/tsdb/index/tsi1/partition.go
@@ -103,35 +103,35 @@ func NewPartition(sfile *tsdb.SeriesFile, path string) *Partition {
 }
 
 // bytes estimates the memory footprint of this Partition, in bytes.
-func (i *Partition) bytes() int {
+func (p *Partition) bytes() int {
 	var b int
 	b += 24 // mu RWMutex is 24 bytes
-	b += int(unsafe.Sizeof(i.opened))
+	b += int(unsafe.Sizeof(p.opened))
 	// Do not count SeriesFile because it belongs to the code that constructed this Partition.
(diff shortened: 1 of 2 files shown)


jacobmarble:
Currently, tests fail because of a mutex deadlock, so this is likely to change tomorrow.

PR #5565, opened 2016-02-06
Configurable precision on UDP services

tpitale:
* pass configured precision string to point parsing
* add Precision configuration to UDP config
* default configured precision to match what it appears to be now (from ParsePoints)

PR #10051, opened 2018-07-06
Update example config with UDP precision option

gunnaraasen:

This updates the config and diagnostic info for the UDP service to include the `precision` option, which was added in PR https://github.com/influxdata/influxdb/pull/5565.

### DIFF
--- a/etc/config.sample.toml
+++ b/etc/config.sample.toml
@@ -487,6 +487,9 @@
   # bind-address = ":8089"
   # database = "udp"
   # retention-policy = ""
+ 
+  # InfluxDB precision for timestamps on received points ("" or "n", "u", "ms", "s", "m", "h")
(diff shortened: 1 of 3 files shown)


rbetts:
Before we merge this documentation change, it looks like the original PR didn't include a testcase. Also, we need to review this for Enterprise in case this is a configuration stanza that needs to be updated in both places.

mark-rushakoff:
Yep, looks like services/udp/service_test.go still hasn't been updated with a precision test case since that original PR.  This PR so far looks good, but +1 to adding a test case.

gunnaraasen:
Enterprise already has precision documented in the sample config file. I think not including it here was an oversight.

I've updated the text to reflect the Enterprise config snippet and also added precision to the UDP config test.

Issue #10056, opened 2018-07-09
TSI Log File Lock Contention


e-dard:
The TSI index appears to be deadlocking. I think this is a regression introduced by #10050.

Attached are some profiles. Often these can be double `RLocks`.

[block.txt](https://github.com/influxdata/influxdb/files/2176284/block.txt)
[goroutine-debug2.txt](https://github.com/influxdata/influxdb/files/2176285/goroutine-debug2.txt)
[goroutine.txt](https://github.com/influxdata/influxdb/files/2176286/goroutine.txt)

e-dard:
I don't think this is right: https://github.com/influxdata/influxdb/pull/10050/commits/681f22b07853097fc6fd54aa3337f609e83b580e#diff-5a703050674f5cd277f93f7f47b8f6e2R1066

We're now going to be calling `p.compactLogFile(logFile)` under a `Lock`, which will be very damaging to performance, and could possibly be the source of the deadlock.

benbjohnson:
Yeah, good catch. That'll back it up quite a bit. I can take a look at that.

jacobmarble:
This is easiest to reproduce when:
- the database is empty (the index doesn't already have series)
- writes are concurrent (race)

The following locks are blocking 30m after the writes started to time out:
- `tsm1.partition.write()` calls `p.mu.Lock()` line 246
- `tsi1.Partition.RetainFileSet()` calls `p.mu.RLock()` line 432
- `tsi1.Partition.CheckLogFile()` calls `p.mu.RLock()` line 1040
- `tsdb.SeriesPartition.CreateSeriesListIfNotExists()` calls `p.mu.RLock()` line 180
- `tsdb.SeriesPartition.CreateSeriesListIfNotExists()` calls `p.mu.Lock()` line 210

This log file shows that the timeouts start exactly when TSI log compaction starts, at 13:57:11. TSI log file compaction is what I was fixing in #10050, so that makes sense.

[log.txt](https://github.com/influxdata/influxdb/files/2178141/log.txt)

PR #10057, opened 2018-07-09
Revert "Resolve deadlock"

jacobmarble:
This reverts commit 681f22b07853097fc6fd54aa3337f609e83b580e.

Temporary fix for #10056


e-dard:
Let's give the current approach to #10056 another day or so then we can apply this reversion if things are still not fruitful?

jacobmarble:
Cool. This is here for anyone to merge if patience runs out while I'm out-of-timezone.

PR #10062, opened 2018-07-10
[WIP] Add per-measurement series id sets.

benbjohnson:
This pull request adds per-measurement series id sets to TSI1.

_Note: This is a work-in-progress. Do not merge._

### DIFF
--- a/tsdb/index/tsi1/index.go
+++ b/tsdb/index/tsi1/index.go
@@ -617,17 +617,17 @@ func (i *Index) InitializeSeries(keys, names [][]byte, tags []models.Tags) error
 // DropSeries drops the provided series from the index.  If cascade is true
 // and this is the last series to the measurement, the measurment will also be dropped.
 func (i *Index) DropSeries(seriesID uint64, key []byte, cascade bool) error {
+	// Extract measurement name.
+	name, _ := models.ParseKeyBytes(key)
+
 	// Remove from partition.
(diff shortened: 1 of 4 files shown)


benbjohnson:
@e-dard I added per-measurement series id sets but after reviewing your profile again I realized that the time is mainly spent on key/value series sets and not measurement sets. InfluxQL only fetches series from measurement block records for `SELECT FROM m` and `SELECT FROM m WHERE tag != 'foo'`. So the problem with this approach is that we would need series id sets for both measurement blocks and tag value elements. That would require a lot of write overhead.

Another option that's probably better long term would be to store the roaring bitmaps in the index files and materialize them with their containers pointing at the raw mmap. We'd need to fork `RoaringBitmap/roaring` to add mmap support but all the containers use little endian values:

- https://github.com/RoaringBitmap/roaring/blob/master/arraycontainer.go#L10
- https://github.com/RoaringBitmap/roaring/blob/master/bitmapcontainer.go#L12
- https://github.com/RoaringBitmap/roaring/blob/master/runcontainer.go#L49-L65

The bitmap would replace the existing series id lists in the index. We could do it as an incremental change to TSI1 so it would support both formats.

What do you think?

e-dard:
@benbjohnson let's discuss.

Current flame graph is looking like this.

benbjohnson:
I'm closing this until we see a need in profiles to move series id sets to persistent roaring bitmaps in TSI1.

PR #10063, opened 2018-07-10
Make store include context in logs

zeebo:
If some error or message is in the context of some shard or database
be sure to include it in the message.

### DIFF
--- a/tsdb/store.go
+++ b/tsdb/store.go
@@ -106,15 +106,16 @@ func (s *Store) Statistics(tags map[string]string) []models.Statistic {
 	databases := s.Databases()
 	statistics := make([]models.Statistic, 0, len(databases))
 	for _, database := range databases {
+		log := s.Logger.With(logger.Database(database))
 		sc, err := s.SeriesCardinality(database)
 		if err != nil {
-			s.Logger.Info("Cannot retrieve series cardinality", zap.Error(err))
+			log.Info("Cannot retrieve series cardinality", zap.Error(err))
 			continue
 		}
 
 		mc, err := s.MeasurementsCardinality(database)
 		if err != nil {
-			s.Logger.Info("Cannot retrieve measurement cardinality", zap.Error(err))
+			log.Info("Cannot retrieve measurement cardinality", zap.Error(err))
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


tsdb/store.go
```diff
@@ -108,13 +108,15 @@ func (s *Store) Statistics(tags map[string]string) []models.Statistic {
 	for _, database := range databases {
 		sc, err := s.SeriesCardinality(database)
 		if err != nil {
-			s.Logger.Info("Cannot retrieve series cardinality", zap.Error(err))
+			s.Logger.Info("Cannot retrieve series cardinality", zap.Error(err),
```
e-dard: How about `log := s.Logger.With(logger.Database(database))`? I know it's only two places but I'm a fan of injecting the shared structures into the logger; it feels a bit cleaner.

tsdb/store.go
```diff
@@ -1755,7 +1757,9 @@ func (s *Store) monitorShards() {
 			for _, sh := range s.shards {
 				if sh.IsIdle() {
 					if err := sh.Free(); err != nil {
-						s.Logger.Warn("Error while freeing cold shard resources", zap.Error(err))
+						s.Logger.Warn("Error while freeing cold shard resources",
+							zap.Error(err),
+							zap.Uint64("shard", sh.ID()))
```
e-dard: I think we have something in the `logger` package for a shard.

hercules-influx:
```/tmp/506351119/src/github.com/influxdata/influxdb/tsdb/shard.go:625:11: this value of keys is never used (SA4006)
/tmp/506351119/src/github.com/influxdata/influxdb/tsdb/shard.go:625:17: this value of names is never used (SA4006)
/tmp/506351119/src/github.com/influxdata/influxdb/tsdb/shard.go:625:24: this value of tagsSlice is never used (SA4006)
/tmp/506351119/src/github.com/influxdata/influxdb/tsdb/shard.go:626:3: this value of types is never used (SA4006)
```

hercules-influx:
```/tmp/845391129/src/github.com/influxdata/influxdb/cmd/store/query/query.go:390:5: this value of line is never used (SA4006)
/tmp/845391129/src/github.com/influxdata/influxdb/cmd/store/query/query.go:404:5: this value of line is never used (SA4006)
```

PR #10073, opened 2018-07-12
Update to Go 1.10

e-dard:
This PR updates the Go runtime used to 1.10.

In 289a05e I have updated the "next" version we run in tests to Go 1.11, however until that release is available I have left the actual version at 1.10.3.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -69,7 +69,7 @@ second to sign our CLA, which can be found
 
 Installing Go
 -------------
-InfluxDB requires Go 1.9.2.
+InfluxDB requires Go 1.10.3.
 
 At InfluxDB we find gvm, a Go version manager, useful for installing Go. For instructions
 on how to install it see [the gvm page on github](https://github.com/moovweb/gvm).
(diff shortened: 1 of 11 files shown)


CONTRIBUTING.md
```diff
@@ -69,16 +69,16 @@ second to sign our CLA, which can be found
...
 
 At InfluxDB we find gvm, a Go version manager, useful for installing Go. For instructions
 on how to install it see [the gvm page on github](https://github.com/moovweb/gvm).
 
 After installing gvm you can install and set the default go version by
 running the following:
 
-    gvm install go1.9.2
-    gvm use go1.9.2 --default
+    gvm install go1.10.3
```
mark-rushakoff: I'm pretty sure nobody on our team uses gvm anymore. But we can deal with that doc change later.

Dockerfile_build_ubuntu64_go1.11
```diff
@@ -24,7 +24,10 @@ RUN gem install fpm
 
 # Install go
 ENV GOPATH /root/go
-ENV GO_VERSION 1.9.2
+
+# TODO(edd) this needs to be updated to 1.11 when the branch is available.
+ENV GO_VERSION 1.10.3
```
mark-rushakoff: Possibly we could use 1.11beta1 for now on this: https://golang.org/dl/#unstable

But I can live with it set to 1.10 for now.

e-dard: Yeah I don't fancy the grief. Let's wait for the 1.11 release, update to that, and then we can do testing on it until we ship Influx 1.7/1.8  with Go 1.11 if there have been no issues over the summer/autumn.

PR #10080, opened 2018-07-13
Improve performance of buildtsi and TSI planning 

e-dard:
Backport of #10048.


hercules-influx:
```/tmp/215543923/src/github.com/influxdata/influxdb/cmd/store/query/query.go:390:5: this value of line is never used (SA4006)
```

PR #10005, opened 2018-06-27
Missing hardwareAddr in uuid v1 generation

amallia:
###### Required for all non-trivial PRs
- [X] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

###### Required only if applicable
_You can erase any checkboxes below this note if they are not applicable to your Pull Request._
- [ ] [InfluxQL Spec](https://github.com/influxdata/influxdb/blob/master/influxql/README.md) updated
- [ ] Provide example syntax
- [ ] Update man page when modifying a command
- [ ] Config changes: update sample config (`etc/config.sample.toml`), server `NewDemoConfig` method, and `Diagnostics` methods reporting config settings, if necessary
- [ ] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): issue filed or pull request submitted \<link to issue or pull request\>

PR #10084, opened 2018-07-13
Update changelog for #10005

jacobmarble:


hercules-influx:
```/tmp/759036291/src/github.com/influxdata/influxdb/cmd/store/query/query.go:390:5: this value of line is never used (SA4006)
/tmp/759036291/src/github.com/influxdata/influxdb/cmd/store/query/query.go:404:5: this value of line is never used (SA4006)
```

PR #10090, opened 2018-07-17
Improve performance of storage read service

e-dard:
This PR significantly improves the performance of the `storage` `read` service when querying on a single measurement.

This primarily impacts on the Prometheus read  endpoint, and the `flux` query language.

In the `read` service, we fetch series from the index, and then have to determine the fields that needed to be combined with the series to enable us to fetch TSM data. Previously we did this by  using the current `influxql` query engine to grab all the field keys for each measurement.

One of the changes in this PR is to expose the mapping we already maintain within a shard between measurements and fields, providing a more efficient means to get the field keys for a measurement.

The next change we made in this PR was to determine when an expression only contains a single measurement, and then to provide an iterator that only iterates over that single measurement, rather than iterating over all measurements. This has a significant performance improvement when there are many measurements in the database. 

Currently we limit supported expressions to things like:

```
_measurement = 'm0' AND tag1 = 'foo'
```

but not

```
_measurement = 'm0' AND (tag1 = 'foo' OR tag0 = 'bar')
_measurement =~ /m0/ 
```

### Benchmarks 

I wrote `1M`  series across `1700` measurements. These were laid over 25 shards.

I then executed the following expression via the `store` command:

```
store query -database stress -expr "_measurement = 'm0' AND tag0 = 'value4' AND tag1 = 'value2' AND tag2 = 'value7' AND tag3 = 'value2'"
```

which should return 100 points.

#### Results

|| master  | PR |
|------| ------------- | ------------- |
|| 3065 ms  | 3.9 ms   |
|| 3049 ms | 3.3 ms   |
|| 3517 ms  | 3.8 ms  |
|| 3070 ms |  3.4 ms |
|| 3062 ms  | 4.1 ms |
|μ |3152 ms| 3.7ms|

The latency has been reduced by 99.9% (850X faster).


hercules-influx:
```/tmp/212052029/src/github.com/influxdata/influxdb/cmd/store/query/query.go:390:5: this value of line is never used (SA4006)
/tmp/212052029/src/github.com/influxdata/influxdb/cmd/store/query/query.go:404:5: this value of line is never used (SA4006)
```

hercules-influx:
```/tmp/196532599/src/github.com/influxdata/influxdb/cmd/store/query/query.go:390:5: this value of line is never used (SA4006)
/tmp/196532599/src/github.com/influxdata/influxdb/cmd/store/query/query.go:404:5: this value of line is never used (SA4006)
```

services/storage/predicate_influxql.go
```diff
@@ -242,6 +242,56 @@ func (v *hasRefs) Visit(node influxql.Node) influxql.Visitor {
...
+// HasSingleMeasurementNoOR determines if an index optimisation is available.
+//
+// Typically the read service will use the query engine to retrieve all field
+// keys for all measurements that match the expression, which can be very
+// inefficient if it can be proved that only one measurement matches the expression.
+//
+// This condition is determined when the following is true:
+//
+//		* there is only one occurrence of the tag key `_measurement`.
+//		* there are no OR operators in the expression tree.
```
dgnorton: If there's only one occurrence of `_measurement`, does it matter if there is an OR operator?

e-dard: @dgnorton yes because you can have `_measurement = 'foo' OR _measurement = 'bar'`

dgnorton: That doesn't count as one measurement?

e-dard: No. In this model, a measurement is just a tag called `_measurement`, therefore we can't tell easily how many measurements in the index we need to select from.

stuartcarnie: @dgnorton flux would allow a query like:

```flux
filter(fn: (d) => (d._measurement == "cpu" and d._field == "idle") OR (d._measurement == "mem" and d._field == "free")
```

we would like to add future improvements, such that the optimization would still apply in the following scenario, which still applies to a single measurement:

```flux
filter(fn: (d) => (d._measurement == "cpu" and (d._field == "system" OR d._field == "user"))
```

services/storage/series_cursor.go
```diff
@@ -12,9 +12,11 @@ import (
 	"github.com/opentracing/opentracing-go"
 )
 
+const measurementKey = "_measurement"
+
 var (
-	measurementKey = []byte("_measurement")
-	fieldKey       = []byte("_field")
+	measurementKeyBytes = []byte(measurementKey)
+	fieldKey            = []byte("_field")
```
stuartcarnie: nit: `fieldKeyBytes` for consistency

zeebo:
My main worry is that since the optimization is so useful/large, there can be some pretty drastic performance pitfalls for some very innocuous query changes.

PR #10091, opened 2018-07-17
Allow tag keys to contain underscores

dgnorton:
Backport of PR #10007


e-dard:
@dgnorton can you (1) put a reference to the original PR in the description; (2) update the PR title to something that would be OK to see in the changelog; and (3) add the `kind/bug` label. Thanks!

PR #10092, opened 2018-07-17
Update Windows CI Go to 1.10

e-dard:
A companion to #10073.


hercules-influx:
```/tmp/293576191/src/github.com/influxdata/influxdb/cmd/store/query/query.go:390:5: this value of line is never used (SA4006)
/tmp/293576191/src/github.com/influxdata/influxdb/cmd/store/query/query.go:404:5: this value of line is never used (SA4006)
```

PR #10094, opened 2018-07-17
Fix a panic when matching on a specific type of regular expression

### LINKED ISSUE #10069: Regression in query regex handling (panic)
__System info:__ InfluxDB 1.6.0, Debian stretch

__Steps to reproduce:__
With our existing data, the following query used to work without issue:
```
SELECT non_negative_derivative(mean("tx"), 1s) * 8
FROM "interface_if_octets"
WHERE ("host" =~ /^gw101[ab]-\d\./ AND "plugin_instance" =~ /^eth2\.1\d{3}$/) AND time >= now() - 5m
GROUP BY time(1m), "host"
```

__Expected behavior:__ Results returned as per previously successful operation.

__Actual behavior:__ No data shown in Grafana, InfluxDB CLI reports a panic
Since upgrading to InfluxDB 1.6.0, it fails, and a panic can be see in the `influx` CLI tool:
```
> SELECT non_negative_derivative(mean("tx"), 1s) * 8 FROM "interface_if_octets" WHERE ("host" =~ /^gw101[ab]-\d\./ AND "plugin_instance" =~ /^eth2\.1\d{3}$/) AND time >= now() - 5m GROUP BY time(1m), "host"
ERR: SELECT non_negative_derivative(mean(tx), 1s) * 8 FROM perfdb."default".interface_if_octets WHERE (host =~ /^gw101[ab]-\d\./ AND plugin_instance =~ /^eth2\.1\d{3}$/) AND time >= now() - 5m GROUP BY time(1m), host [panic:runtime error: index out of range]
```

__Additional info:__
After some experimentation looking for workarounds, it seems to be failing on the `/^eth2\.1\d{3}$/` regex, specifically the `$` end-of-line match. Removing the `$` allows the query to work.

Also, changing the regex to `/^eth2\.1\d{1,3}$/` works, however `/^eth2\.1\d{3,3}$/` panics.

The tag in the database contains values such as `eth2.1xxx`, where `x` is a digit 0-9.

This has the hallmarks of an off-by-one error when comparing to the tag value, since the `/^eth2\.1\d{3}$/` regex would eat the entire string length of the values we have for that tag.

Somewhat mysteriously, the regex `/^e[a-z0-9]+\.1\d{3}$/` works successfully.

jsternberg:
Fixes #10069.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -119,7 +119,7 @@
     ".",
     "internal"
   ]
-  revision = "5e999e6a81820d4450f2a1f35c5597b569258f01"
+  revision = "c661ab7db8ad858626cc7a2114e786f4e7463564"
 
 [[projects]]
   branch = "master"


hercules-influx:
```/tmp/157938637/src/github.com/influxdata/influxdb/cmd/store/query/query.go:390:5: this value of line is never used (SA4006)
```

e-dard:
@jsternberg don't we want this to appear in the `1.6.1` CHANGELOG rather than the `1.7` `CHANGELOG`? Should the `kind/bug` label be on the backport PR?

jsternberg:
It should be in both sections so this is fine. The `kind/bug` label should be on the backport PR and I forgot so I added that before merging.

PR #10120, opened 2018-07-24
Fix bug on tsdb/arrayvalues

linxGnu:

Reproduce:
- code base would fail with the following test case (`tsdb/arrayvalues.gen_test.go`, function `TestIntegerArray_Exclude`):
```
{"excl suffix", 16, 18, []int64{10, 12, 14}},
```

Cause:
- due to wrong handle to exclude values. When max value equals to max elem, cause the array unchanged.

Impact:
- may cause wrong compaction and wrong queries


stuartcarnie:
@linxGnu thanks for the fix 👍

PR #10137, opened 2018-07-30
fix restore shard bug

z1q1q7:

###### Reproduce
* operating system: CentOS release 6.7 (Final)
* version: 1.6.0 (download from official site)
* test case: Any shard that ends with "0" will restore to wrong shard.


e-dard:
@aanthony1243 this makes sense to me, can you take a look?

z1q1q7:
> @z1q1q7 before we can merge this, could you please sign the [CLA](https://www.influxdata.com/community/cla/).

So sorry for my late reply. I have already sigh the CLA.
image

PR #10147, opened 2018-08-02
series file type checking

zeebo:
# Overview

This implements doing type checking per series. This PR introduces the `tsdb.SeriesID` and `tsdb.SeriesIDTyped` values that abstract the internal representation of a series id. This allowed the compiler to track down all of the places that needed to be modified, flagging the spots that needed consideration. The `ID` field is unfortunately exported on the types due to an issue with some reflect like tests (the diffing business). If this is considered to be too overbearing of a change, I wouldn't mind changing back to either typed uint64s or naked uint64s now that the hard work is done.

The next large portion of the PR is to have the code use a new abstraction: the `tsdb.SeriesCollection`. This type exists so that we can avoid passing around 4 parallel slices both in and out of many methods. It also helps with the logic of invalidating bad points, keeping track of partial write errors, and shuffling around the slices cheaply.

The portion of the complexity in the collection comes from how the series file handles doing lookups/writes of new series in parallel across its partitions. Since that's also the spot where we know what type the series id is supposed to be, invalidating the series at that moment becomes challenging. Using the fact that most writes won't have a problem, we make it cheap and allocation free in the case that there are no conflicts.

Lastly, one change I would like to draw attention to is in tsi1. When creating series in tsi1, it would first split the work into 8 subsets and pass those to its partitions. Then, each tsi1 partition would individually pass *their* subset of the data to the series file. Then, the series file would split that work against *its* 8 partitions, causing a total of 64 splits, and much ado. Instead, tsi1 now passes the entire collection to the series file *before* splitting the work for its individual partitions. Not only do I think this is better, it makes the job of the series file efficiently dropping incorrectly typed series easier since it can operate on all of the data at once, and no merging of the individual pieces is required.

## Benchmarks

Some benchmark results for inmem[1], names elided for space:

```
old time/op    new time/op    delta
  2.70ms ± 1%    2.64ms ± 1%  -2.31%          (p=0.000 n=8+7)
   314µs ± 2%     315µs ± 1%    ~             (p=0.442 n=8+8)
   314µs ± 4%     315µs ± 1%    ~             (p=0.536 n=8+7)
  2.02ms ± 1%    2.00ms ± 1%  -0.88%          (p=0.001 n=8+7)

old alloc/op   new alloc/op   delta
   477kB ± 0%     477kB ± 0%  -0.04%          (p=0.000 n=8+8)
   15.0B ± 0%     15.0B ± 0%    ~     (all samples are equal)
   15.0B ± 0%     15.0B ± 0%    ~     (all samples are equal)
   501kB ± 0%     501kB ± 0%  -0.04%          (p=0.000 n=8+8)

old allocs/op  new allocs/op  delta
    37.0 ± 0%      38.0 ± 0%  +2.70%          (p=0.000 n=8+8)
   0.00 ±NaN%     0.00 ±NaN%    ~     (all samples are equal)
   0.00 ±NaN%     0.00 ±NaN%    ~     (all samples are equal)
    23.0 ± 0%      23.0 ± 0%    ~     (all samples are equal)
```

tsi1 has a more comprehensive benchmark suite, but I had trouble running it, so I used inch to create 30 million series by running `inch -b 10000 -c 20 -t 10,10,100000,3 -m 1 -p 1 -v`:

| strategy    | vals/sec          | seconds        | ratio |
|-------------|-------------------|----------------|-------|
| no checking | 80829.1 val/sec.  | 371.2 seconds  | 1x    |
| checking    | 81896.5 val/sec.  | 366.3 seconds  | 0.98x |

which I believe can be attributed to noise.

---

[1] The inmem benchmarks were also benchmarking the initial add in some situations, causing the allocation amounts to vary based on how many times it ran, since only the first time allocated. The above benchmarks include backported changes to fix this.

### DIFF
--- a/cmd/influx_inspect/buildtsi/buildtsi.go
+++ b/cmd/influx_inspect/buildtsi/buildtsi.go
@@ -268,41 +268,42 @@ func IndexShard(sfile *tsdb.SeriesFile, dataDir, walDir string, maxLogFileSize i
 		}
 
 		log.Info("Iterating over cache")
-		keysBatch := make([][]byte, 0, batchSize)
-		namesBatch := make([][]byte, 0, batchSize)
-		tagsBatch := make([]models.Tags, 0, batchSize)
+		collection := &tsdb.SeriesCollection{
+			Keys:  make([][]byte, 0, batchSize),
+			Names: make([][]byte, 0, batchSize),
+			Tags:  make([]models.Tags, 0, batchSize),
+			Types: make([]models.FieldType, 0, batchSize),
+		}
 
 		for _, key := range cache.Keys() {
 			seriesKey, _ := tsm1.SeriesAndFieldFromCompositeKey(key)
 			name, tags := models.ParseKeyBytes(seriesKey)
+			typ, _ := cache.Type(key)
 
 			if verboseLogging {
 				log.Info("Series", zap.String("name", string(name)), zap.String("tags", tags.String()))
 			}
 
-			keysBatch = append(keysBatch, seriesKey)
-			namesBatch = append(namesBatch, name)
-			tagsBatch = append(tagsBatch, tags)
+			collection.Keys = append(collection.Keys, seriesKey)
+			collection.Names = append(collection.Names, name)
+			collection.Tags = append(collection.Tags, tags)
+			collection.Types = append(collection.Types, typ)
 
 			// Flush batch?
-			if len(keysBatch) == batchSize {
-				if err := tsiIndex.CreateSeriesListIfNotExists(keysBatch, namesBatch, tagsBatch); err != nil {
+			if collection.Length() == batchSize {
+				if err := tsiIndex.CreateSeriesListIfNotExists(collection); err != nil {
 					return fmt.Errorf("problem creating series: (%s)", err)
 				}
-				keysBatch = keysBatch[:0]
-				namesBatch = namesBatch[:0]
-				tagsBatch = tagsBatch[:0]
+				collection.Truncate(0)
 			}
 		}
 
 		// Flush any remaining series in the batches
-		if len(keysBatch) > 0 {
-			if err := tsiIndex.CreateSeriesListIfNotExists(keysBatch, namesBatch, tagsBatch); err != nil {
+		if collection.Length() > 0 {
+			if err := tsiIndex.CreateSeriesListIfNotExists(collection); err != nil {
 				return fmt.Errorf("problem creating series: (%s)", err)
 			}
-			keysBatch = nil
-			namesBatch = nil
-			tagsBatch = nil
+			collection = nil
 		}
 	}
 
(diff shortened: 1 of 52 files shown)


hercules-influx:
```/tmp/265371993/src/github.com/influxdata/influxdb/tsdb/index/tsi1/tsi1.go:516:6: type uint64Slice is unused (U1000)
/tmp/265371993/src/github.com/influxdata/influxdb/tsdb/series_file.go:471:6: type uint64Slice is unused (U1000)
/tmp/265371993/src/github.com/influxdata/influxdb/tsdb/series_id.go:68:2: const seriesIDStructSize is unused (U1000)
/tmp/265371993/src/github.com/influxdata/influxdb/tsdb/series_id.go:69:2: const seriesIDTypedStructSize is unused (U1000)
```

zeebo:
Circle CI seems to be exposing this failure:

```
panic: interface conversion: tsm1.Value is tsm1.IntegerValue, not tsm1.FloatValue

goroutine 42196 [running]:
github.com/influxdata/influxdb/tsdb/engine/tsm1.(*floatAscendingCursor).peekCache(0x1b0bae70, 0x1, 0x2, 0x2, 0x24)
	/root/go/src/github.com/influxdata/influxdb/tsdb/engine/tsm1/iterator.gen.go:410 +0xb1
github.com/influxdata/influxdb/tsdb/engine/tsm1.(*floatAscendingCursor).nextFloat(0x1b0bae70, 0x8d0c93c, 0x0, 0x1, 0x80556d8)
	/root/go/src/github.com/influxdata/influxdb/tsdb/engine/tsm1/iterator.gen.go:441 +0x25
github.com/influxdata/influxdb/tsdb/engine/tsm1.(*floatIterator).Next(0x1acd4240, 0x87de2e0, 0x1, 0xf54b62b8)
	/root/go/src/github.com/influxdata/influxdb/tsdb/engine/tsm1/iterator.gen.go:236 +0x556
github.com/influxdata/influxdb/tsdb_test.TestShard_WritePoints_FieldConflictConcurrentQuery.func2(0x1acb4140, 0x1ad64e40)
	/root/go/src/github.com/influxdata/influxdb/tsdb/shard_test.go:627 +0xa95
created by github.com/influxdata/influxdb/tsdb_test.TestShard_WritePoints_FieldConflictConcurrentQuery
	/root/go/src/github.com/influxdata/influxdb/tsdb/shard_test.go:580 +0x3d2
FAIL	github.com/influxdata/influxdb/tsdb	8.708s
```

I have a feeling this change is causing it. Tickling it out with cpulimit.

zeebo:
Alright, I cannot for the life of me get this to fail locally, so I'm going to start the arduous process of pushing commits with a bunch of temporary logging to extract more information out of CircleCI.

zeebo:
The issue discovered here by CircleCI is now #10169.

tsdb/series_id.go
```diff
@@ -0,0 +1,74 @@
...
+	seriesIDValueMask = 0xFFFFFFFF                // series ids numerically are 32 bits
+	seriesIDTypeShift = 32                        // we put the type right after the value info
+	seriesIDTypeMask  = 0xFF << seriesIDTypeShift // a mask for the type byte
+)
+
+// SeriesIDHasType returns if the raw id contains type information.
+func SeriesIDHasType(id uint64) bool { return id&seriesIDTypeFlag > 0 }
+
+// SeriesID is the type of a series id.
+type SeriesID struct{ ID uint64 }
```
benbjohnson: What's the benefit of using a struct instead of:

```go
type SeriesID uint64
```

tsdb/series_index.go
```diff
@@ -144,50 +144,52 @@ func (idx *SeriesIndex) OnDiskCount() uint64 { return idx.count }
...
 	idx.execEntry(SeriesEntryInsertFlag, id, offset, key)
 }
 
 // Delete marks the series id as deleted.
-func (idx *SeriesIndex) Delete(id uint64) {
-	idx.execEntry(SeriesEntryTombstoneFlag, id, 0, nil)
+func (idx *SeriesIndex) Delete(id SeriesID) {
+	// TODO(jeff): WithType(0) kinda sucks here, but we know it will be masked off.
+	idx.execEntry(SeriesEntryTombstoneFlag, id.WithType(0), 0, nil)
 }
```
benbjohnson: Should this `TODO` be renamed to `NOTE`? Or is there a change to be made?

tsdb/series_index_test.go
```diff
@@ -6,9 +6,14 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/influxdata/influxdb/models"
 	"github.com/influxdata/influxdb/tsdb"
 )
 
+func typedSeriesIDHelper(id uint64) tsdb.SeriesIDTyped {
+	return tsdb.NewSeriesID(id).WithType(models.Empty)
+}
```
benbjohnson: Can this be called `toSeriesIDTyped()`? "helper" doesn't explain what it does very well IMO.

cmd/influx_inspect/buildtsi/buildtsi.go
```diff
@@ -416,3 +424,20 @@ func isRoot() bool {
...
+	case tsm1.BlockBoolean:
+		return models.Boolean
+	case tsm1.BlockString:
+		return models.String
+	case tsm1.BlockUnsigned:
+		return models.Unsigned
+	default:
+		return models.Empty
+	}
+}
```
benbjohnson: Can this be moved to `tsm1` since it's duplicated in the `importer` package as well?

tests/server_test.go
```diff
@@ -6289,17 +6289,17 @@ func TestServer_Query_Where_Fields(t *testing.T) {
 		},
 	}...)
 
-	for i, query := range test.queries {
+	var once sync.Once
+	for _, query := range test.queries {
 		t.Run(query.name, func(t *testing.T) {
-			if i == 0 {
+			once.Do(func() {
```
benbjohnson: ty. These `i == 0` checks drive me nuts. 💯

tsdb/series_collection.go
```diff
@@ -0,0 +1,334 @@
...
+
+// Copy will copy the element at src into dst in all slices that can: x[dst] = x[src].
+func (s *SeriesCollection) Copy(dst, src int) {
+	if dst == src {
+		return
+	}
+	udst, usrc := uint(dst), uint(src)
+	if len := uint(len(s.Points)); udst < len && usrc < len {
+		s.Points[udst] = s.Points[usrc]
+	}
```
benbjohnson: I'd personally rename `len` to `n` so you don't shadow the `len()` function. It makes it a little easier to read, however, it's not really a big deal.

tsdb/series_collection.go
```diff
@@ -0,0 +1,334 @@
...
+}
+
+// PartialWriteError returns a PartialWriteError if any entries have been marked as invalid. It
+// returns an error to avoid `return collection.PartialWriteError()` always being non-nil.
+func (s *SeriesCollection) PartialWriteError() error {
+	if s.Dropped == 0 {
+		return nil
+	}
+	droppedKeys := bytesutil.SortDedup(s.DroppedKeys)
+	return PartialWriteError{
```
stuartcarnie: I'm glad to see this is consistently a value vs mixed pointer / value. I verified Enterprise does not expect a pointer anywhere either 👍

tsdb/series_id.go
zeebo: It doesn't automatically convert from an untyped constant. The only reason the ID field is actually exported is because a ton of tests were using some reflect business that didn't handle the unexported field. I wanted to be totally opaque so that I caught all usages of it. Many places were directly comparing to 0, or using <.

tests/server_test.go
zeebo: It caused me significant pain when some tests were failing but only if I ran that specific test (since it was still index 3 or whatever, and 0 never ran!)

tsdb/series_id.go
```diff
@@ -0,0 +1,76 @@
...
```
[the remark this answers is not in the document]
zeebo: See my response to ben, above (?? where is github gonna put it?)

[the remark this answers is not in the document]
stuartcarnie: WFM 👍

cmd/influx_inspect/buildtsi/buildtsi.go
zeebo: As it turns out, they aren't really duplicated. One operates on the encoded byte in the tsm block, and the other operates on some value out of a protobuf. These values don't match.

tsdb/series_id.go
e-dard: @zeebo seems like a sensible choice. I think it might be worth just documenting that on the type so it doesn't get changed back by someone with good intentions.

PR #10152, opened 2018-08-03
Add option to hint MADV_WILLNEED to kernel

### LINKED ISSUE #10117: Consider making TSM MADV_WILLNEED a configuration option
__Proposal:__ 

In 1.5.1 we introduced a feature that used the `mmap` [hint](http://man7.org/linux/man-pages/man2/madvise.2.html) `MADV_WILLNEED` when loading TSM files: https://github.com/influxdata/influxdb/pull/9380. The general idea was that because we need to scan TSM indexes on startup, using `MADV_WILLNEED` would let the kernel know it could read ahead pages, or at least prep them in the disk cache. Of course, we can't really control what the kernel does, just hint.

Due to https://github.com/influxdata/influxdb/issues/9534#issuecomment-375086909 we ended up having to remove the `MADV_WILLNEED` hint in `1.5.2`. See https://github.com/influxdata/influxdb/pull/9614

__Current behavior:__

`MADV_WILLNEED` is not used as a hint to the kernel that we plan to scan the TSM index. 

__Desired behavior:__ 

We have found that in some cases, users have moved from `1.5.1` (where we used `MADV_WILLNEED`) to `>= 1.5.2` (where we don't), and have experienced significant IO Wait on `>= 1.5.2`. It's possible this is down to slow disks, and that in their case, on `1.5.1` their kernel is paging in TSM files in a way that is helping their IO performance.

We could consider making the `MADV_WILLNEED` hint a configuration option.

e-dard:
Fixes #10117.

This PR adds a new configuration option - `tsm-use-madv-willneed` / `INFLUXDB_DATA_TSM_USE_MADV_WILLNEED`, which, when true, will hint to the kernel that we plan on sequentially reading TSM files.

This option defaults to false because we have had some issues in the past with certain kernels/operating systems. However, it may be useful to some users with slow disks.

### DIFF
--- a/tsdb/engine/tsm1/reader.go
+++ b/tsdb/engine/tsm1/reader.go
@@ -30,7 +30,8 @@ type TSMReader struct {
 	refs   int64
 	refsWG sync.WaitGroup
 
-	mu sync.RWMutex
+	madviseWillNeed bool // Hint to the kernel with MADV_WILLNEED.
+	mu              sync.RWMutex
 
 	// accessor provides access and decoding of blocks for the reader.
 	accessor blockAccessor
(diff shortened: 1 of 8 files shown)


tsdb/engine/tsm1/reader.go
```diff
@@ -1325,6 +1339,15 @@ func (m *mmapAccessor) init() (*indirectIndex, error) {
 		return nil, fmt.Errorf("mmapAccessor: byte slice too small for indirectIndex")
 	}
 
+	// Hint to the kernel that we will be reading the file.  It would be better to hint
+	// that we will be reading the index section, but that's not been
+	// implemented as yet.
+	if m.mmapWillNeed {
+		if err := madviseWillNeed(m.b); err != nil {
```
zeebo: munmap here? i guess this comment applies to a bunch of spots in this function lol. in addition to file handle resources. rename has the same problem. seems big enough for another ticket/bug maybe?

PR #10164, opened 2018-08-07
influx_inspect: Make buildtsi code resusable

jacobmarble:
Exported two methods in `buildtsi.go` so that they can be used from other code.

Follow-up to #10019

### DIFF
--- a/cmd/influx_inspect/buildtsi/buildtsi.go
+++ b/cmd/influx_inspect/buildtsi/buildtsi.go
@@ -183,7 +183,8 @@ func (cmd *Command) processRetentionPolicy(sfile *tsdb.SeriesFile, dbName, rpNam
 				}
 
 				id, name := shards[i].ID, shards[i].Path
-				errC <- cmd.processShard(sfile, dbName, rpName, id, filepath.Join(dataDir, name), filepath.Join(walDir, name))
+				log := cmd.Logger.With(logger.Database(dbName), logger.RetentionPolicy(rpName), logger.Shard(id))
+				errC <- IndexShard(sfile, filepath.Join(dataDir, name), filepath.Join(walDir, name), cmd.maxLogFileSize, cmd.batchSize, log, cmd.Verbose)
 			}
 		}()
 	}
@@ -197,8 +198,7 @@ func (cmd *Command) processRetentionPolicy(sfile *tsdb.SeriesFile, dbName, rpNam
 	return nil
 }
 
-func (cmd *Command) processShard(sfile *tsdb.SeriesFile, dbName, rpName string, shardID uint64, dataDir, walDir string) error {
-	log := cmd.Logger.With(logger.Database(dbName), logger.RetentionPolicy(rpName), logger.Shard(shardID))
+func IndexShard(sfile *tsdb.SeriesFile, dataDir, walDir string, maxLogFileSize int64, batchSize int, log *zap.Logger, verboseLogging bool) error {
 	log.Info("Rebuilding shard")
 
 	// Check if shard already has a TSI index.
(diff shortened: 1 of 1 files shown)


cmd/influx_inspect/buildtsi/util.go
```diff
@@ -0,0 +1,218 @@
+package buildtsi
+
+import (
```
e-dard: I feel like this code could just live in the same `buildtsi.go` file.

cmd/influx_inspect/buildtsi/util.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/influxdata/influxdb/tsdb/index/tsi1"
+	"go.uber.org/zap"
+)
+
+func IndexShard(sfile *tsdb.SeriesFile, dataDir, walDir string, maxLogFileSize int64, batchSize int, log *zap.Logger, verboseLogging bool) error {
+	log.Info("Rebuilding shard")
+
+	// Check if shard already has a TSI index.
+	indexPath := filepath.Join(dataDir, "index")
+	log.Info("Checking index path", zap.String("path", indexPath))
```
e-dard: Maybe we could do a `log = log.With(zap.String("path", indexPath)` here?

cmd/influx_inspect/buildtsi/util.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	log.Info("Opening shard")
+
+	// Find shard files.
+	tsmPaths, err := collectTSMFiles(dataDir)
+	if err != nil {
+		return err
+	}
+	walPaths, err := collectWALFiles(walDir)
```
e-dard: Have you tested what happens when we run this in an environment without a WAL? I think `collectWALFiles` would return an error, which means rebuilding will fail. If that's the case we can either treat `walDir == ""` as a special case in `collectWALFiles` or we can check here, and only call `collectWALFiles` in the case `walDir != ""`.

cmd/influx_inspect/buildtsi/util.go
```diff
@@ -0,0 +1,218 @@
...
+	tsiIndex := tsi1.NewIndex(sfile, "",
+		tsi1.WithPath(tmpPath),
+		tsi1.WithMaximumLogFileSize(maxLogFileSize),
+		tsi1.DisableFsync(),
+		// Each new series entry in a log file is ~12 bytes so this should
+		// roughly equate to one flush to the file for every batch.
+		tsi1.WithLogFileBufferSize(12*batchSize),
+	)
+
+	tsiIndex.WithLogger(log)
```
e-dard: Ah maybe this would cause problems if we added the `path` key above. I'm not sure how zap behaves if you add the same key multiple times...

cmd/influx_inspect/buildtsi/util.go
e-dard: ^ I think adding `path` to the logger might break things further down, so maybe keep it as it is 😄

cmd/influx_inspect/buildtsi/util.go
```diff
@@ -0,0 +1,218 @@
...
+	log.Info("Iterating over tsm files")
+	for _, path := range tsmPaths {
+		log.Info("Processing tsm file", zap.String("path", path))
+		if err := IndexTSMFile(tsiIndex, path, batchSize, log, verboseLogging); err != nil {
+			return err
+		}
+	}
+
+	// Write out wal files.
+	log.Info("Building cache from wal files")
```
e-dard: We don't need to do this all the way to line 120 if we don't have a WAL.

jacobmarble: Done

cmd/influx_inspect/buildtsi/util.go
jacobmarble: Done. I disagree, but my real problem is that this lives inside a tool, not inside a library package. I originally tried to move this to the `tsi1` package, but that creates a recursive dependency.

cmd/influx_inspect/buildtsi/buildtsi.go
```diff
@@ -386,7 +390,13 @@ func (cmd *Command) collectTSMFiles(path string) ([]string, error) {
 	return paths, nil
 }
 
-func (cmd *Command) collectWALFiles(path string) ([]string, error) {
+func collectWALFiles(path string) ([]string, error) {
+	if path == "" {
+		return []string{}, os.ErrNotExist
```
e-dard: It's rarely correct to return an allocated slice with an error. This should just be `return nil, os.ErrNotExist`. Same below.

PR #10168, opened 2018-08-07
Add getter and setter for config of a BatchPoints on client

FdeFabricio:
I am working on project that fragments one BatchPoints into multiples, so the configuration needs to be duplicated. This two functions will be useful to avoid verbose code.


e-dard:
@FdeFabricio thanks for this PR. I don't think think we're going to take it though. The more general use-case isn't clear.

PR #10172, opened 2018-08-08
Fix iterator panic in read service 

### LINKED ISSUE #10171: Read service panics under certain conditions due to nil SeriesIDIterator
Certain requests with a predicate containing a regular expression may cause a panic due to a `nil` value in a `tsdb.seriesIDMergeIterator`. These `nil` values  have been identified to occur in 4 possible places:

https://github.com/influxdata/influxdb/blob/979d790154acc2e1d3815c8a83be5f66324886e6/tsdb/index.go#L2183

https://github.com/influxdata/influxdb/blob/979d790154acc2e1d3815c8a83be5f66324886e6/tsdb/index.go#L2226

https://github.com/influxdata/influxdb/blob/979d790154acc2e1d3815c8a83be5f66324886e6/tsdb/index.go#L2257

and

https://github.com/influxdata/influxdb/blob/979d790154acc2e1d3815c8a83be5f66324886e6/tsdb/index.go#L2287

stuartcarnie:
Do not add `nil` values when calling `MergeSeriesIDIterators`, which will result in a panic. A `nil` value may occur when a predicate contains regular expressions.

fixes #10171

### DIFF
--- a/tsdb/index.go
+++ b/tsdb/index.go
@@ -2179,8 +2179,9 @@ func (is IndexSet) matchTagValueEqualEmptySeriesIDIterator(name, key []byte, val
 				itr, err := is.tagValueSeriesIDIterator(name, key, e)
 				if err != nil {
 					return err
+				} else if itr != nil {
+					itrs = append(itrs, itr)
 				}
-				itrs = append(itrs, itr)
 			}
 		}
 		return nil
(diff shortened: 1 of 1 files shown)


stuartcarnie:
@e-dard this is a different iterator bug and has currently only manifested itself via the regex predicates and storage read RPC calls.  Worth noting it could affect Prometheus. I will update issue #10171 with the note that it is regex conditions and also add that to the PR.

PR #10173, opened 2018-08-08
test(stress): Fix flaky test

stuartcarnie:
Fixes a flaky test point values with a leading zero (`0` or `0i`) are generated

```
=== RUN   Test_typeArr
--- FAIL: Test_typeArr (0.00s)
    stress_test.go:86: Expected line protocol int got 0i
```


stuartcarnie:
@zeebo you're right, changing it

PR #10184, opened 2018-08-09
backport SHOW MEASUREMENTS panic fix

jacobmarble:
backports #10175


hercules-influx:
```/tmp/152982057/src/github.com/influxdata/influxdb/cmd/store/query/query.go:347:5: this value of line is never used (SA4006)
/tmp/152982057/src/github.com/influxdata/influxdb/cmd/store/query/query.go:361:5: this value of line is never used (SA4006)
```

PR #10186, opened 2018-08-09
[Experimental] TSI Series ID set caching

e-dard:
**Experimental Work**

This PR introduces a caching layer to the TSI index, which aims to improve the performance of retrieving sets of Series IDs associated with a `{measurment_name, tag_key, tag_value}` tuple. For example, you often ask the index "get me all the series ids for any series under the `cpu` measurement with tag pair `region = west`.

Since the TSI index is partitioned, and since within each partiition we can have many `LogFile`s (`.tsl`) and `IndexFile`s (`.tsi`), getting the set of matching series ids can be expensive. 

It involves merging at least eight bitmaps, and often more.

This work is an experiment to improve the performance of the mechanism by adding a cache at the `tsi1.Index` layer. Therefore rather than gathering all the individual sets and merging them we cache them.

Since the index is mutable as new series arrive into `LogFile`s, we also have to check if any cached bitsets need to be updated.

We implement the cache as a simple LRU (see `tsdb/index/tsi1/cache.go`), where bitsets are moved to the front of a linked list (the eviction queue) as they're retrieved from the cache. Although the bitsets are roaring bitmaps, and thus well compressed, we limit the cache size to minimise impact on the heap.

This PR has some serious caveats!!

  1) It *does not support deletes*. Not implemented yet. If you delete a series that is held in a cached bitset then much sadness will ensue. 
  2) It only supports series id sets for `{name, key, value}` tuples. That's what you would use when you do a query invoving a tag filter. It doesn't help you if you do something like `SELECT COUNT(v0) FROM cpu`. To implement...
 3) It has only been tested on a very particular workload. It needs more rigourous testing. That's why it's experimental.

Finally, there is also a [commit in here](https://github.com/influxdata/influxdb/commit/4275a0c0bcf5ba1a580a125d0620fc68f06615d2) that reduces the garbage being generated by `TagSets` when the TSI index is been used. That commit will help all the lovely people using TSI  🎉 ☮️ .

#### Performance Characteristics 

The caching layer is going to be most helpful when there are concurrent write and query load, and those queries benefit from the cache.

Here are a set of benchmarks for the index level iterator that the cache is helping with:

```
⇒  benchstat /tmp/old-tv.txt /tmp/new-tv.txt
name                                                                              old time/op    new time/op    delta
Index_IndexFile_TagValueSeriesIDIterator/78888_series_TagValueSeriesIDIterator-8     169µs ± 7%       1µs ± 0%  -99.62%  (p=0.000 n=10+9)

name                                                                              old alloc/op   new alloc/op   delta
Index_IndexFile_TagValueSeriesIDIterator/78888_series_TagValueSeriesIDIterator-8     168kB ± 0%       1kB ± 0%  -99.56%  (p=0.000 n=9+10)

name                                                                              old allocs/op  new allocs/op  delta
Index_IndexFile_TagValueSeriesIDIterator/78888_series_TagValueSeriesIDIterator-8       524 ± 0%        13 ± 0%  -97.52%  (p=0.000 n=10+10)
```

And an example of what happens when there is a cache miss

```
BenchmarkIndex_IndexFile_TagValueSeriesIDIterator/78888_series_TagValueSeriesIDIterator/cache-8   	 2000000	       699 ns/op	     744 B/op	      13 allocs/op
BenchmarkIndex_IndexFile_TagValueSeriesIDIterator/78888_series_TagValueSeriesIDIterator/no_cache-8         	   10000	    134489 ns/op	  124952 B/op	     350 allocs/op
```

Here is a benchmark that shows how long it takes to write 1M series to the TSI index whilst also executing 100,000 queries against the index concurrently:

```
BenchmarkIndex_ConcurrentWriteQuery/partition_8/queries_100000/cache-8            	1	1714771527 ns/op	2341518456 B/op		23450621 allocs/op
BenchmarkIndex_ConcurrentWriteQuery/partition_8/queries_100000/no_cache-8         	1	22242979152 ns/op	28178915600 B/op	80156305 allocs/op
```

##### Real world example

I have been testing against a real-world internally. It has around 1.5 million series and is under constant write load. Using the `qprof` tool I have been measuring the performance characteristics of different releases. Here are some results where a `LAST` query that has three predicates on it is executed with 10 concurrent workers as many times as possible over a 90s period. (1.6-exp is the tip of the branch we're merging into).

| | v1.5.3 (ms) | v1.6.1 (ms)  | 1.6-exp (ms)|  PR (ms)|
|:-------------:|:-------------:|:-------------:|:-------------:|:-------------:|
|μ| 1150   |  238 | 65| 5.98 | 
|Med| 1081   |  203 | 28| 1.92 |
|Max|  3637 | 1071 |1033 | 633 |

The throughput was as follows:

| | v1.5.3 (q/s) | v1.6.1 (q/s)  | 1.6-exp (q/s)| PR (q/s)|
|:-------------:|:-------------:|:-------------:|:-------------:|:-------------:|
|| 8   |  41 | 153| 1676 |


hercules-influx:
```/tmp/654492739/src/github.com/influxdata/influxdb/tsdb/index/tsi1/log_file_test.go:75:7: this value of err is never used (SA4006)
/tmp/654492739/src/github.com/influxdata/influxdb/tsdb/index/tsi1/index.go:1075:6: type ssElement is unused (U1000)
```

hercules-influx:
```/tmp/007620461/src/github.com/influxdata/influxdb/tsdb/index/tsi1/log_file_test.go:75:7: this value of err is never used (SA4006)
/tmp/007620461/src/github.com/influxdata/influxdb/tsdb/index/tsi1/index.go:1075:6: type ssElement is unused (U1000)
```

tsdb/index/tsi1/cache.go
```diff
@@ -0,0 +1,175 @@
...
+				goto EVICT
+			}
+
+			// Add the set to the map
+			tkmap[string(value)] = listElement
+			goto EVICT
+		}
+
+		// No series set map for the tag key - first tag value for the tag key.
+		mmap[string(key)] = map[string]*list.Element{string(value): listElement}
```
zeebo: I don't know how often these lines happen vs the `tkmap[string(value)] = listElement` above, but i think doing it like this causes an extra allocation on `string(value)`. Maybe the compiler is smart enough, though?

tsdb/index/tsi1/index.go
```diff
@@ -590,7 +607,54 @@ func (i *Index) CreateSeriesListIfNotExists(keys [][]byte, names [][]byte, tagsS
...
+							// Taking the cache lock will block all queries, but is one lock. Taking each series set
+							// lock might be many lock/unlocks but will only block a query that needs that particular set.
+							//
+							// Need to think on it, but I think taking a lock on each series id set is the way to go.
+							//
+							// One other option here is to take a lock on the series id set when we first encounter it
+							// and then keep it locked until we're done with all the ids.
+							//
+							// Note: this will only add `id` to the set if it exists.
+							i.tagValueCache.addToSet(name, pair.Key, pair.Value, id) // Takes a lock on the series id set
```
zeebo: one other option to consider here is to just drop the cache entries for the `name`. that will be less good if measurements have a really big scope, though.

e-dard:
I need to update the default cache size before this is mergae.

tsdb/index/tsi1/index.go
```diff
@@ -845,7 +946,19 @@ func (i *Index) TagValueSeriesIDIterator(name, key, value []byte) (tsdb.SeriesID
...
+	itr := tsdb.MergeSeriesIDIterators(a...)
+	if !EnableBitsetCache {
+		return itr, nil
+	}
+
+	// Check if the iterator contains only series id sets. Cache them...
+	if ssitr, ok := itr.(tsdb.SeriesIDSetIterator); ok {
+		ss := ssitr.SeriesIDSet()
+		ss.SetCOW(true) // This is important to speed the clone up.
+		i.tagValueCache.Put(name, key, value, ss)
```
zeebo: seems like concurrent calls to `TagValueSeriesIDIterator` can cause that debugging condition in `Put` to happen. Maybe either do the `RLock` + check + `Lock` + check + insert song and dance, or maybe it's ok to just drop it on the floor in the `Put` call since it should be the same value anyway?

tsdb/index/tsi1/index.go
```diff
@@ -956,3 +1069,12 @@ func IsIndexDir(path string) (bool, error) {
...
 }
+
+// ssElement is is stored within the LRU. It makes it possible to do constant
+// time access via a hashmap and also storage inside a linked list.
+type ssElement struct {
+	name        []byte
+	key         []byte
+	value       []byte
+	SeriesIDSet *tsdb.SeriesIDSet
+}
```
zeebo: vestigial?

tsdb/index/tsi1/log_file.go
```diff
@@ -550,21 +551,25 @@ func (f *LogFile) AddSeriesList(seriesSet *tsdb.SeriesIDSet, names [][]byte, tag
 	seriesSet.Lock()
 	defer seriesSet.Unlock()
 
-	for i := range entries {
+	for i := range entries { // NB - this doesn't evaluate all series ids returned from series file.
 		entry := &entries[i]
 		if seriesSet.ContainsNoLock(entry.SeriesID) {
 			// We don't need to allocate anything for this series.
+			seriesIDs[i] = 0
```
zeebo: the index in entries isn't necessarily the same index in seriesIDs.

tsdb/series_file.go
```diff
@@ -355,18 +355,38 @@ func ReadSeriesKeyTag(data []byte) (key, value, remainder []byte) {
...
+	var name []byte
 	_, data = ReadSeriesKeyLen(data)
 	name, data = ReadSeriesKeyMeasurement(data)
 
 	tagN, data := ReadSeriesKeyTagN(data)
-	tags = make(models.Tags, tagN)
+
+	if got, want := len(dst), tagN; got < want {
+		dst = append(dst, make(models.Tags, want-got)...)
+	}
```
zeebo: maybe you want to reslice `dst = dst[:cap]` before doing this checking? additionally, it looks like you need to do a `dst = dst[:tagN]` after ensuring it has enough capacity.

zeebo:
This is really cool. It looks like you took the path of attempting to keep the cache updated as changes happen. Do you think it's worthwhile to instead assume a more steady state and carefully evict some entries if we detect they could have changed? Then we don't have to worry about keeping the data in sync with the source of truth.

tsdb/index/tsi1/cache.go
e-dard: I'll note that for the final PR but I don't think it's a biggie for this experiment.

tsdb/index/tsi1/index.go
e-dard: Adding to the set under lock is tens of nanoseconds, depending on the value being added. Rebuilding the cached series id set is hundreds of microseconds. I think if there was constant series creation for a general tag (like "region = west") then dropping the cached set each time would be more expensive.

tsdb/index/tsi1/index.go
e-dard: We have to `Lock` the cache anyway because we have to put the item at the beginning of the eviction queue. So even if we drop the item on the floor we still would have had to have `Lock`ed the cache.

e-dard:
@zeebo I have updated. My concern with evicting on change is that for our internal projects where measurement would be an entire database of data, any new series would evict the entire cache, which would defeat its purpose in the first place.

hercules-influx:
```/tmp/119024881/src/github.com/influxdata/influxdb/tsdb/index/tsi1/log_file_test.go:75:7: this value of err is never used (SA4006)
```

hercules-influx:
```/tmp/273743947/src/github.com/influxdata/influxdb/tsdb/index/tsi1/log_file_test.go:75:7: this value of err is never used (SA4006)
```

PR #10197, opened 2018-08-15
pkg/snowflake: reduce contention in Next calls

zeebo:
use atomics rather than mutexes to synchronize state between calls.

```
name            old time/op  new time/op  delta
Next-8           244ns ± 0%    30ns ± 2%  -87.70%  (p=0.000 n=8+7)
NextParallel-8   215µs ±60%     0µs ± 1%  -99.96%  (p=0.000 n=8+8)
```

The results for NextParallel are around ~80ns/op, but that doesn't
show up in the benchstat output.

### DIFF
--- a/pkg/snowflake/gen_test.go
+++ b/pkg/snowflake/gen_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"math/rand"
 	"sort"
+	"sync/atomic"
 	"testing"
 
 	"github.com/influxdata/influxdb/pkg/testing/assert"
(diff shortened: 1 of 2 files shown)


hercules-influx:
```/tmp/610360901/src/github.com/influxdata/influxdb/pkg/snowflake/gen.go:17:2: const sequenceMask is unused (U1000)
```

pkg/snowflake/gen_test.go
```diff
@@ -60,6 +66,24 @@ func BenchmarkEncode(b *testing.B) {
...
+
+func BenchmarkNextParallel(b *testing.B) {
+	g := New(1)
+
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			g.Next()
+		}
+	})
+}
```
docmerlin: Looks like the call to g.Next is being elided out, you may want to make sure the compiler knows not to do that.

zeebo: It actually isn't in this case (confirmed with -gcflags=-S), but I'll update to make sure it doesn't happen in the future as best as I'm able.

hercules-influx:
```/tmp/462434495/src/github.com/influxdata/influxdb/pkg/snowflake/gen.go:17:2: const sequenceMask is unused (U1000)
```

PR #10200, opened 2018-08-16
httpd/meta: use open auth when unrestricted

tmgordeeva:
This changes adds additional auth shortcutting, primarily for enterprise
usecases with simple FGA setups. OSS users won't see any changes.

### DIFF
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -475,8 +475,12 @@ func (h *Handler) serveQuery(w http.ResponseWriter, r *http.Request, user meta.U
 	}
 
 	if h.Config.AuthEnabled {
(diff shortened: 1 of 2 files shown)


services/httpd/handler.go
```diff
@@ -475,8 +475,12 @@ func (h *Handler) serveQuery(w http.ResponseWriter, r *http.Request, user meta.U
 	}
 
 	if h.Config.AuthEnabled {
-		// The current user determines the authorized actions.
-		opts.Authorizer = user
+		if (user != nil) && user.AuthorizeUnrestricted() {
```
e-dard: nit: these parentheses are redundant.

PR #10201, opened 2018-08-16
pkg/snowflake: update platform branch.

zeebo:
Applies #10197 and #10202 to platform branch.


zeebo:
Actually, this isn't the right change, and I made some improvements. Wait a bit.

PR #10212, opened 2018-08-21
Ensure orhpaned series cleaned up with shard drop

e-dard:
This PR fixes an issue where series could be left in the `inmem` index, when they should have been removed.

It could be triggered as follows:

  - insert some series into a shard that only exist in that shard;
  - wait for the retention policy enforcement to remove the shard;
  - run a `SHOW` query such as `SHOW TAG KEYS`;
  - The series that were removed with the shard will be missing (as expected);
  - re-insert the same series;
  - run a `SHOW` query such as `SHOW TAG KEYS`;
  - The series will **not** be present (they should bed).

This manifests itself because the series were removed from the series file, but not from the `inmem` index. If the server was restarted the issue would resolve itself.

Commit f52de2d ensures that the series are removed from the `inmem` index if they're the last occurence of those series in the index.

TODO: whilst doing this work, it's become apparent that dropping shards with the `inmem` index is particularly racy, when one considers concurrent writes of the same series. I will add a ticket to track that.

### DIFF
--- a/tsdb/store.go
+++ b/tsdb/store.go
@@ -352,7 +352,7 @@ func (s *Store) loadShards() error {
 
 					// Existing shards should continue to use inmem index.
 					if _, err := os.Stat(filepath.Join(path, "index")); os.IsNotExist(err) {
-						opt.IndexVersion = "inmem"
+						opt.IndexVersion = InmemIndexName
 					}
 
 					// Open engine.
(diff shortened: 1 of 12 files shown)


tsdb/store.go
```diff
@@ -701,10 +698,42 @@ func (s *Store) DeleteShard(shardID uint64) error {
...
+					}
+
+					name, tagsBuf = ParseSeriesKeyInto(skey, tagsBuf)
+					keyBuf = models.AppendMakeKey(keyBuf, name, tagsBuf)
+					if err = index.DropSeriesGlobal(keyBuf); err != nil {
+						return
+					}
+				})
+
+				if err != nil {
```
jacobmarble: Not sure this conditional does anything.

Issue #10146, opened 2018-08-01
Add counter when exceeding http.max-row-limit


benbjohnson:
__Proposal:__ Add an internal metric to track how many requests hit the `http.max-row-limit` value.

__Current behavior:__ No metric is reported.

__Desired behavior:__ Add `1` to a counter whenever a request exceeds the row limit.

__Use case:__ Operators wish to limit the number of rows that users can query but have no visibility into the settings and how to tune them.

liukun4515:
Can I implement this feature in the `httpd.hander`?
@benbjohnson

liukun4515:
This feature is reveiewed by PM?

PR #10225, opened 2018-08-23
http handler: add metric for hit the max row limit

liukun4515:
#10146


liukun4515:
PTAL @benbjohnson

e-dard:
@liukun4515 thanks for the PR. I don't think we're going to accept this though.

PR #10227, opened 2018-08-23
Use snappy compression for digests

dgnorton:
This PR backports https://github.com/influxdata/influxdb/pull/10215 to 1.6.

### DIFF
--- a/tsdb/engine/tsm1/digest.go
+++ b/tsdb/engine/tsm1/digest.go
@@ -142,6 +142,7 @@ func DigestFresh(dir string, files []string, shardLastMod time.Time) (bool, stri
 	if err != nil {
 		return false, fmt.Sprintf("Can't read digest: err=%s", err)
 	}
+	defer dr.Close()
 
 	mfest, err := dr.ReadManifest()
 	if err != nil {
--- a/tsdb/engine/tsm1/digest_reader.go
+++ b/tsdb/engine/tsm1/digest_reader.go
@@ -1,84 +1,73 @@
 package tsm1
 
 import (
-	"bufio"
-	"compress/gzip"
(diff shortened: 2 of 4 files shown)


tsdb/engine/tsm1/digest.go
```diff
@@ -142,6 +142,7 @@ func DigestFresh(dir string, files []string, shardLastMod time.Time) (bool, stri
 	if err != nil {
 		return false, fmt.Sprintf("Can't read digest: err=%s", err)
 	}
+	defer dr.Close()
```
jacobmarble: `defer` is good to ensure cleanup on early return, but can you also check the error returned by `dr`, perhaps just after the last usage? `Close` should be idempotent.

tsdb/engine/tsm1/digest_reader.go
```diff
@@ -1,84 +1,73 @@
...
 
 	return string(b), ts, nil
 }
 
-func (w *DigestReader) Close() error {
-	if err := w.gr.Close(); err != nil {
-		return err
-	}
-	return w.r.Close()
+func (r *DigestReader) Close() error {
```
jacobmarble: No need to close `r.sr`?

dgnorton: Surprised me, too, but `snappy.Reader` doesn't have a `Close`. https://github.com/golang/snappy/blob/master/decode.go

tsdb/engine/tsm1/digest.go
dgnorton: Good idea. Since this is a backport of something on `master`, I'll make a note and change it on a future commit.

PR #10228, opened 2018-08-23
update influxql version

jlapacik:
This PR updates the version of influxql to this commit https://github.com/influxdata/influxql/commit/a7267bff5327e316e54c54342b0bc9598753e3d5 . This fixes a bug where predicate expressions were getting parsed incorrectly.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -113,7 +113,7 @@
     ".",
     "internal"
   ]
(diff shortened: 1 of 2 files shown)


nathanielc:
It looks like you have used a newer version of `dep` than is currently in use for this repo. I am not sure if that is a problem.

PR #10231, opened 2018-08-24
fix(storage): Fix potential panic if nil is returned

stuartcarnie:


hercules-influx:
```/tmp/371808939/src/github.com/influxdata/influxdb/services/storage/group_resultset_test.go:106:19: invalid operation: rs (variable of type GroupResultSet) has no field or method rows
couldn't load packages due to errors: github.com/influxdata/influxdb/services/storage
```

PR #10234, opened 2018-08-26
Improve TSI query planning with cached bitsets

e-dard:
This PR introduces a caching layer to the TSI index, which aims to improve the performance of retrieving sets of Series IDs associated with a `{measurment_name, tag_key, tag_value}` tuple. For example, you often ask the index "get me all the series ids for any series under the `cpu` measurement with tag pair `region = west`.

Since the TSI index is partitioned, and since within each partiition we can have many `LogFile`s (`.tsl`) and `IndexFile`s (`.tsi`), getting the set of matching series ids can be expensive because it involves unioning at least eight bitmaps, and often more.

This work improves the performance of this process by adding a cache at the `tsi1.Index` layer. Therefore, rather than gathering all the individual sets and merging them, we cache them.

Since the index is modified as new series arrive into `LogFile`s, we also have to check if any cached bitsets need to be updated.

The cache is implemented as a simple LRU (see `tsdb/index/tsi1/cache.go`), where bitsets are moved to the front of a linked list (the eviction queue) as they're retrieved from the cache. Although the bitsets are roaring bitmaps, and thus well compressed, we limit the cache size to minimise impact on the heap.

#### Performance Characteristics 

The caching layer is going to be most helpful when there are concurrent write and query load, and those queries benefit from the cache.

Here are a set of benchmarks for the index level iterator that the cache is helping with:

```
⇒  benchstat /tmp/old-tv.txt /tmp/new-tv.txt
name                                                                              old time/op    new time/op    delta
Index_IndexFile_TagValueSeriesIDIterator/78888_series_TagValueSeriesIDIterator-8     169µs ± 7%       1µs ± 0%  -99.62%  (p=0.000 n=10+9)

name                                                                              old alloc/op   new alloc/op   delta
Index_IndexFile_TagValueSeriesIDIterator/78888_series_TagValueSeriesIDIterator-8     168kB ± 0%       1kB ± 0%  -99.56%  (p=0.000 n=9+10)

name                                                                              old allocs/op  new allocs/op  delta
Index_IndexFile_TagValueSeriesIDIterator/78888_series_TagValueSeriesIDIterator-8       524 ± 0%        13 ± 0%  -97.52%  (p=0.000 n=10+10)
```

And an example of what happens when there is a cache miss

```
BenchmarkIndex_IndexFile_TagValueSeriesIDIterator/78888_series_TagValueSeriesIDIterator/cache-8   	 2000000	       699 ns/op	     744 B/op	      13 allocs/op
BenchmarkIndex_IndexFile_TagValueSeriesIDIterator/78888_series_TagValueSeriesIDIterator/no_cache-8         	   10000	    134489 ns/op	  124952 B/op	     350 allocs/op
```

Here is a benchmark that shows how long it takes to write 1M series to the TSI index whilst also executing 100,000 queries against the index concurrently:

```
BenchmarkIndex_ConcurrentWriteQuery/inmem/queries_100000/cache-8   	       1	6008440191 ns/op	2499732776 B/op	23964208 allocs/op
BenchmarkIndex_ConcurrentWriteQuery/inmem/queries_100000/no_cache-8         	       1	5714187427 ns/op	2499339096 B/op	23962364 allocs/op
BenchmarkIndex_ConcurrentWriteQuery/tsi1/queries_100000/cache-8             	       1	2113678244 ns/op	2216227120 B/op	23049356 allocs/op
BenchmarkIndex_ConcurrentWriteQuery/tsi1/queries_100000/no_cache-8          	       1	22474623589 ns/op	28310531776 B/op	79742846 allocs/op
```

##### Real world example

I have been testing against a real-world internally. It has around 1.5 million series and is under constant write load. Using the `qprof` tool I have been measuring the performance characteristics of different releases. Here are some results where a `LAST` query that has three predicates on it is executed with 10 concurrent workers as many times as possible over a 90s period. (1.6-exp is the tip of the branch we're merging into).

| | v1.5.3 (ms) | v1.6.1 (ms)  |   PR (ms)|
|:-------------:|:-------------:|:-------------:|:-------------:|
|μ| 1150   |  238 |5.98 | 
|Med| 1081   |  203 | 1.92 |
|Max|  3637 | 1071 | 633 |

The throughput was as follows:

| | v1.5.3 (q/s) | v1.6.1 (q/s)  | PR (q/s)|
|:-------------:|:-------------:|:-------------:|:-------------:|
|| 8   |  41 |  1676 |


tsdb/index/tsi1/cache.go
```diff
@@ -0,0 +1,198 @@
...
+		value:       value,
+		SeriesIDSet: ss,
+	})
+
+	// Add the listElement to the set of items.
+	if mmap, ok := c.cache[string(name)]; ok {
+		if tkmap, ok := mmap[string(key)]; ok {
+			if _, ok := tkmap[string(value)]; ok {
+				// FIXME(edd): only here whilst testing.
+				fmt.Println("existence of cache item breaks invariant")
```
zeebo: leftover?

tsdb/index/tsi1/index_test.go
```diff
@@ -411,3 +579,271 @@ func (idx *Index) CreateSeriesSliceIfNotExists(a []Series) error {
...
+				b.Fatal("got nil iterator")
+			}
+		}
+	}
+
+	// This benchmark will merge eight bitsets each containing ~10,000 series IDs.
+	b.Run("78888 series TagValueSeriesIDIterator", func(b *testing.B) {
+		b.ReportAllocs()
+		b.Run("cache", func(b *testing.B) {
+			runBenchMark(b)
```
zeebo: this should probably set `tsi1.EnableBitsetCache = true` to avoid ordering issues with other possible benchmarks? and/or default changes.

PR #10249, opened 2018-08-31
Prevent DELETE FROM to rebuild series files for shards where nothing is deleted

### LINKED ISSUE #10218: DELETE query is very slow -- still slow in InfluxDb 1.6.0
__System info:__
InfluxDb 1.6.0
Ubuntu 16.04

__Steps to reproduce:__
I reported this before in #9015 for InfluxDb 1.3.x, but the issue remains: deletes are very slow.

I get a lot of the following in the log:
```
ts=2018-08-22T11:50:40.932742Z lvl=warn msg="Detected slow query: DELETE FROM \"XXX\" WHERE (time >= '2018-08-11 00:00:00' AND time < '2018-08-12 00:00:00') (qid: 1092268, database: xxx, threshold: 10s)" log_id=09NfXRsW000 service=query
```

__Expected behavior:__ 
Deletes should be fast.

__Actual behavior:__
Deletes takes maybe 5-10 seconds in some cases.

__Profile:__
I did a profile while running a delete from in a loop, each taking around 4 seconds.
[block.txt](https://github.com/influxdata/influxdb/files/2332378/block.txt)
[cpu.txt](https://github.com/influxdata/influxdb/files/2332379/cpu.txt)
[diagnostics.txt](https://github.com/influxdata/influxdb/files/2332380/diagnostics.txt)
[goroutine.txt](https://github.com/influxdata/influxdb/files/2332381/goroutine.txt)
[mutex.txt](https://github.com/influxdata/influxdb/files/2332382/mutex.txt)
[shards.txt](https://github.com/influxdata/influxdb/files/2332383/shards.txt)
[stats.txt](https://github.com/influxdata/influxdb/files/2332384/stats.txt)

hpbieker:

Queries such as 'DELETE FROM "measurement" WHERE time > someTime' will typically only affect a few shards. However, this causes all series indexes to be rebuilt as long as the series existed in the shard, but it did not care about the min/max time.

It looks like the existing if check was wrong because it just checked the length of the series array, but did not care about the series name that was set to empty in the loop above.

This PR should fix #10218, or at least partially.

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -1616,7 +1616,14 @@ func (e *Engine) deleteSeriesRange(seriesKeys [][]byte, min, max int64) error {
 
 	// Have we deleted all values for the series? If so, we need to remove
 	// the series from the index.
-	if len(seriesKeys) > 0 {
+	hasDeleted := false;
+	for _, k := range seriesKeys {
+			if len(k) > 0 {
+					hasDeleted = true
+					break
+			}
+	}
+	if hasDeleted {
 		buf := make([]byte, 1024) // For use when accessing series file.
 		ids := tsdb.NewSeriesIDSet()
 		measurements := make(map[string]struct{}, 1)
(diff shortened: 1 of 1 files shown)


hpbieker:
@benbjohnson Would it be possible get this reviewed?

hpbieker:
Any comments?

hpbieker:
@e-dard @benbjohnson @jwilder ? Any chance to get this reviewed? I think it will reduce the delete times significantly...

e-dard:
@hpbieker sorry, this seems to have slipped through the net.... @benbjohnson can you take a look at this too?

benbjohnson:
@hpbieker Sorry about that. The change lgtm. Thanks! 👍

PR #10258, opened 2018-09-05
Fix #10257 add `QueryAsChunk`

### LINKED ISSUE #10257: Add chunked query support to v2 client
__Proposal:__ 
Add chunked query support to v2 client

__Current behavior:__ 
The `Query` structure has chunked option, but all the fetched result points in memory. Therefore it is ineffective

__Desired behavior:__ 
Add the query method that returns `ChunkedReponse` to the client interface.

__Use case:__ 
When fetching large data at once.

miya-masa:


timhallinflux:
Fixes #10258

aaronbee:
Just a heads up, adding a method to an interface is a backwards incompatible change. This broke our build because our mock client used in tests no longer implemented the influxdb.Client interface.

rbetts:
@aaronbee Sorry for the regression.

PR #10278, opened 2018-09-12
Remove TSI1 HLL sketches from heap.

benbjohnson:
## Overview

This commit removes the HLL sketches on each `tsi1.LogFile` and `tsi1.IndexFile` and instead caches the data at the `tsi1.Index` level. This reduces the heap size significantly for servers with many TSI-enabled shards.

Tests using `inch` showed a reduction from `~900MB` to `~200MB` for a `638` shard server with this change. The reduction is larger if there are more compacted index files in each shard.

### DIFF
--- a/tsdb/index/tsi1/log_file.go
+++ b/tsdb/index/tsi1/log_file.go
@@ -62,9 +62,6 @@ type LogFile struct {
 	size    int64            // tracks current file size
 	modTime time.Time        // tracks last time write occurred
 
-	mSketch, mTSketch estimator.Sketch // Measurement sketches
-	sSketch, sTSketch estimator.Sketch // Series sketche
-
 	// In-memory series existence/tombstone sets.
 	seriesIDSet, tombstoneSeriesIDSet *tsdb.SeriesIDSet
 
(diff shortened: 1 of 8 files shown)


tsdb/index/tsi1/log_file.go
```diff
@@ -1035,32 +1008,45 @@ type logFileMeasurementCompactInfo struct {
...
-// MergeMeasurementsSketches merges the measurement sketches belonging to this
-// LogFile into the provided sketches.
-//
-// MergeMeasurementsSketches is safe for concurrent use by multiple goroutines.
-func (f *LogFile) MergeMeasurementsSketches(sketch, tsketch estimator.Sketch) error {
+// MeasurementsSketches returns sketches for existing and tombstoned measurement names.
+func (f *LogFile) MeasurementsSketches() (sketch, tSketch estimator.Sketch, err error) {
 	f.mu.RLock()
 	defer f.mu.RUnlock()
+	return f.measurementsSketches()
```
docmerlin: since you have named return variables there is no need to use a `defer` here.
It doesn't really matter but
```go
f.mu.RLock()
sketch, tSketch, err = f.measurementsSketches()
f.mu.RUnlock()
return sketch, tSketch, err
``` 
Would be a tiny bit more performant.  I doubt it would make much of a difference here, though.

PR #10283, opened 2018-09-14
new: support for gzip in client v2

primalmotion:
This patch adds support for gzip compression in the http client.
It can be enabled by using setting `Query{Compressed: true}`


primalmotion:
I don't understand why the test failed. I can't find any failure in appveyor

primalmotion:
I'm tired. This is done automatically by the http.Client

PR #10300, opened 2018-09-24
Improve Compaction Performance

e-dard:
This PR adds new batch-oriented `tsm1.Encoders` that aim to improve the encoding performance of the various TSM blocks (Float, Integer, Unsigned, Time, String and Boolean) with a more efficient implementation of the various encoding schemes.

No encoding/compression behaviour had been altered, and these encoders are compatible with the existing `tsm1` format.

By improving the performance of the encoders, the aim is that compactions themseleves will see improved performance.

This PR concerns itself with the following changes:

  - ac90e5f: a batch oriented float encoder (`FloatArrayEncodeAll`) using Gorilla compression, but avoiding the function call and poor cache effects of using the existing `bitstream` package. Instead, we take all the float values to encode, and encode them all at once. Benchmarks show an improvement of around **5x - 16x** in encoding performance for a 1,000 value block, when compared to the existing encoder implementation. (See `a2724e2` for the updated benchmarks)
 - 5a5ffe6: batch oriented integer/unsigned integer encoders (`IntegerArrayEncodeAll` and `UnsignedArrayEncodeAll`) using either uncompressed, RLE or simple8b compression. Benchmarks show an improvement of around **1.5x - 1.7x** in encoding performance for a 1,000 value block.
  - 5decd99: a batch oriented timestamp encoder (`TimeArrayEncodeAll`) using either uncompressed, RLE or simple8b compression. Benchmarks show an improvement of around **1.5x - 2.5x** in encoding performance for a 1,000 value block. These improvements carry to all blocks, since they all encode timestamps.
 - dd13d79 : a batch oriented string encoder (`StringArrayEncodeAll`). Benchmarks show an improvement of around **13% – 66%** in encoding performance for a 1,000 value block. The original encoder was already quite performant.

 - 0a5eb7a: add block encoders for all types that encode the values and timestamps into the tsm block. 
 - 902ba41 add a new key iterator `tsmBatchKeyIterator` that uses the batch decode and encode functions to compact TSM blocks.

#### Not implemented
There are several things that have not been implemented in this PR, and could be considered future work:

 - Pooling of the buffers used for the encoding. I put a lot of effort into reusing the provided buffer. However, it's a non trivial amount of work to create the memory pools necessary to reuse those buffers. We need to know for example when the buffer has been safely written to disk so we can mutate the buffer. This could be future work.
 - Cache snapshots. The snapshots do not currently use the new encoders. They could however.
 - Encoding in some offline tooling (though the offline compactor will use these changes).

#### TODO 
 
I need to add some compaction benchmarks here....


tsdb/engine/tsm1/compact.gen.go.tmpl
```diff
@@ -203,4 +207,213 @@ func (k *tsmKeyIterator) chunk{{.Name}}(dst blocks) blocks {
...
+				}
+
+				k.merged{{.Name}}Values.Merge(&v)
+			}
+		}
+
+		// Since we combined multiple blocks, we could have more values than we should put into
+		// a single block.  We need to chunk them up into groups and re-encode them.
+		return k.chunk{{.Name}}(nil)
+	} else {
```
benbjohnson: This `else` is not needed since you're returning in the previous block.

tsdb/engine/tsm1/compact.go
```diff
@@ -1556,6 +1556,295 @@ func (k *tsmKeyIterator) Err() error {
...
+	if len(k.blocks) > 0 {
+		k.merge()
+		if len(k.merged) > 0 || k.hasMergedValues() {
+			return true
+		}
+	}
+
+	// Read the next block from each TSM iterator
+	for i, v := range k.buf {
+		if len(v) == 0 {
```
benbjohnson: You can unindent the code in the block by changing this to:

```go
if len(v) != 0 {
	continue
}
```

PR #10307, opened 2018-09-26
Backup continue on errors

aanthony1243:
there are some edge cases where a shard backup fails on a live backup if it is deleted by a retention policy 
after the shard is identified, but before the download starts.  This fix allows the backup process to continue, even with failures so that empty/deleted shards are ignored.  

Adds a new flag : 
`# influxd backup -portable -skip-errors -db telegraf /tmp/backup`

if skip-errors is set, then errors will be skipped as described above.  The default for skip-errors is false, maintaining backward compatability.  

docs PR:  https://github.com/influxdata/docs.influxdata.com/pull/1844

### DIFF
--- a/cmd/influxd/backup/backup.go
+++ b/cmd/influxd/backup/backup.go
@@ -382,6 +382,7 @@ func (cmd *Command) backupResponsePaths(response *snapshotter.Response) error {
 		err = cmd.backupShard(db, rp, id)
 
 		if err != nil && !cmd.continueOnError {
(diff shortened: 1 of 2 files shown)


cmd/influxd/backup/backup.go
```diff
@@ -379,7 +381,7 @@ func (cmd *Command) backupResponsePaths(response *snapshotter.Response) error {
 
 		err = cmd.backupShard(db, rp, id)
 
-		if err != nil {
+		if err != nil && !cmd.continueOnError {
```
e-dard: This error should be loudly logged when it's being ignored.

PR #10326, opened 2018-10-02
Fix append of possible nil iterator.

benbjohnson:
This commit updates an iterator list to ignore `nil` iterators.
Adding a `nil` caused the `SeriesIterators.Close()` to panic.


jsternberg:
Related to #10367.

PR #10339, opened 2018-10-05
Fix series file tombstoning.

benbjohnson:
This commit fixes an issue with the series file compaction process where tombstones are lost after compaction and series existence checks are not correct. This commit also fixes some smaller flushing issues within the series file that mainly related to testing.

### DIFF
--- a/tsdb/series_file_test.go
+++ b/tsdb/series_file_test.go
@@ -124,6 +124,42 @@ func TestSeriesFileCompactor(t *testing.T) {
 	}
 }
 
+// Ensure series file deletions persist across compactions.
+func TestSeriesFile_DeleteSeriesID(t *testing.T) {
+	sfile := MustOpenSeriesFile()
+	defer sfile.Close()
+
+	ids0, err := sfile.CreateSeriesListIfNotExists([][]byte{[]byte("m1")}, []models.Tags{nil})
+	if err != nil {
+		t.Fatal(err)
+	} else if _, err := sfile.CreateSeriesListIfNotExists([][]byte{[]byte("m2")}, []models.Tags{nil}); err != nil {
+		t.Fatal(err)
+	} else if err := sfile.ForceCompact(); err != nil {
+		t.Fatal(err)
+	}
+
+	// Delete and ensure deletion.
+	if err := sfile.DeleteSeriesID(ids0[0]); err != nil {
+		t.Fatal(err)
+	} else if _, err := sfile.CreateSeriesListIfNotExists([][]byte{[]byte("m1")}, []models.Tags{nil}); err != nil {
+		t.Fatal(err)
+	} else if !sfile.IsDeleted(ids0[0]) {
+		t.Fatal("expected deletion before compaction")
+	}
+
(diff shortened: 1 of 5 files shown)


tsdb/series_file_test.go
```diff
@@ -124,6 +124,42 @@ func TestSeriesFileCompactor(t *testing.T) {
...
+	} else if !sfile.IsDeleted(ids0[0]) {
+		t.Fatal("expected deletion after compaction")
+	}
+
+	if err := sfile.Reopen(); err != nil {
+		t.Fatal(err)
+	} else if !sfile.IsDeleted(ids0[0]) {
+		t.Fatal("expected deletion after reopen")
+	}
+}
```
benbjohnson: This is the original failing test case that now works with the new existence check.

tsdb/series_index.go
```diff
@@ -155,8 +155,10 @@ func (idx *SeriesIndex) Delete(id uint64) {
 
 // IsDeleted returns true if series id has been deleted.
 func (idx *SeriesIndex) IsDeleted(id uint64) bool {
-	_, ok := idx.tombstones[id]
-	return ok
+	if _, ok := idx.tombstones[id]; ok {
+		return true
+	}
+	return idx.FindOffsetByID(id) == 0
 }
```
benbjohnson: Tombstones only exist for IDs since the last compaction so we now check if the ID exists in the map as well.

tsdb/series_index.go
```diff
@@ -261,6 +263,11 @@ func (idx *SeriesIndex) Clone() *SeriesIndex {
 		tombstones[id] = struct{}{}
 	}
 
+	idOffsetMap := make(map[uint64]int64)
+	for k, v := range idx.idOffsetMap {
+		idOffsetMap[k] = v
+	}
```
benbjohnson: Since we now check for existence in the map & tombstones, we need to copy over the id/offset map during `Clone()` (which is used to snapshot the current index state).

tsdb/series_partition.go
```diff
@@ -309,6 +309,13 @@ func (p *SeriesPartition) DeleteSeriesID(id uint64) error {
 		return err
 	}
 
+	// Flush active segment write.
+	if segment := p.activeSegment(); segment != nil {
+		if err := segment.Flush(); err != nil {
+			return err
+		}
+	}
```
benbjohnson: This was an unrelated bug but it caused series deletions to not be visible in compactions until another series was added afterward which caused the segment to be flushed. Now all deletions cause a segment flush.

tsdb/series_partition.go
```diff
@@ -572,14 +580,14 @@ func (c *SeriesPartitionCompactor) compactIndexTo(index *SeriesIndex, seriesN ui
 				return fmt.Errorf("unexpected series partition log entry flag: %d", flag)
 			}
 
+			// Save max series identifier processed.
+			hdr.MaxSeriesID, hdr.MaxOffset = id, offset
+
```
benbjohnson: This is another unrelated bug but it caused deleted series to not move the max id/offset during compaction. This is now done before the deletion check.

PR #10361, opened 2018-10-10
Fixing the stream iterator to not ignore the error

jsternberg:
The stream iterator would ignore an error that happened when reading
points. This may have caused it to potentially return an error that got
ignored and then to try invoking `Next()` on an iterator in an invalid
state and that iterator would then actually return a point which it
wasn't supposed to.

Also added some defensive coding to that same call to prevent a nil map
from being assigned to in the event of an invalid iterator returning
junk data.


jsternberg:
Both backports have been created and added to the relevant milestone. I have removed the needs backport tag for both of them from this issue.

PR #10368, opened 2018-10-11
Update platform / flux packages and implement from source

stuartcarnie:


stuartcarnie:
These tests will pass when https://github.com/influxdata/platform/pull/1064 is merged

PR #10375, opened 2018-10-16
Modify context key type

KimMachineGun:
I think that define type with `struct{}` and using it as a context key is more go-ish.

And, actually [golang/go](https://github.com/golang/go) project has been using it.


stuartcarnie:
@KimMachineGun could you sign the [CLA](https://www.influxdata.com/legal/cla/)

KimMachineGun:
sure 😊

e-dard:
@KimMachineGun Thanks for you contribution. As well as signing the CLA you will need to fix the conflict in `context.go`.

PR #10387, opened 2018-10-18
Add shards' index types to /debug/vars

e-dard:
This commit adds an `indexType` key to the shard sections of the
`/debug/vars` endpoint, as well as the `_internal` shard statistics.

The tag will be reported as `"indexType": "inmem"` or `"indexType":
"tsi1"`.

Note, this will increase the cardinality of the `_internal` database.

### DIFF
--- a/tsdb/shard.go
+++ b/tsdb/shard.go
@@ -264,6 +264,13 @@ func (s *Shard) Statistics(tags map[string]string) []models.Statistic {
 	seriesN := engine.SeriesN()
 
 	tags = s.defaultTags.Merge(tags)
+
+	// Set the index type on the tags.  N.B this needs to be checked since it's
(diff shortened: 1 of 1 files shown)


tsdb/shard.go
```diff
@@ -264,6 +264,13 @@ func (s *Shard) Statistics(tags map[string]string) []models.Statistic {
 	seriesN := engine.SeriesN()
 
 	tags = s.defaultTags.Merge(tags)
+
+	// Set the index type on the tags.  N.B this needs to be checked since it's
+	// only set when the shard is opened.
```
dgnorton: What does `N.B` mean?

e-dard: @dgnorton "note well". https://en.wikipedia.org/wiki/Nota_bene

PR #10391, opened 2018-10-19
Reduce allocations for high measurement writes

tmgordeeva:
Removes cloning measurement fields for a lighter way to handle contention between writes adding fields.

### DIFF
--- a/tsdb/shard.go
+++ b/tsdb/shard.go
@@ -604,8 +604,6 @@ func (s *Shard) validateSeriesAndFields(points []models.Point) ([]models.Point, 
 		}
 	}
 
-	// Create a MeasurementFields cache.
-	mfCache := make(map[string]*MeasurementFields, 16)
 	j = 0
 	for i, p := range points {
 		// Skip any points with only invalid fields.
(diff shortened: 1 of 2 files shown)


tmgordeeva:
Ran `inch -t 10 -m 50000 -f 1`, before this change it's 141.9 seconds and with the change it's 108.9 seconds (additionally the benchmarks included here show similar improvements).

zeebo:
I'm worried about the case where we create a lot of fields on the same measurement. Specifically, the new logic in [CreateFieldIfNotExists](https://github.com/influxdata/influxdb/pull/10391/files#diff-5794b17db5ab55866ed4ee776df9a0eaR1500) will be O(N^2) when adding N new fields to the same measurement. This pattern is what happens in [createFieldsAndMeasurements](https://github.com/influxdata/influxdb/pull/10391/files#diff-5794b17db5ab55866ed4ee776df9a0eaR695) after we figure out what fields we have to create.

I see at least two ways to resolve this:

1. It doesn't matter and I'm worried about nothing
2. We can batch all of the new fields we're going to add at once, rather than add them one at a time.

I don't know what a typical number of fields per measurement is, but how do the benchmarks compare with something I'd guess is too big, like `-f 1000`?

zeebo:
I wasn't talking about deserializing. When you add a field, it copies all the existing fields into a new map, then adds the new field. That causes O(N^2) copies if you add the fields one at a time.

I was unaware of the 255 field cap. That, plus the benchmark results, remove all of my fears.

PR #10396, opened 2018-10-22
Make digest reader skip manifest if needed

dgnorton:
This change makes the digest reader read and discard the manifest if
needed. Not all readers of a digest are interested in the manifest.

This change also makes it a requirement for the writer to write a
manifest because it is a non-optional part of a digest file.


tsdb/engine/tsm1/digest_reader.go
```diff
@@ -28,10 +29,22 @@ func (r *DigestReader) ReadManifest() (*DigestManifest, error) {
 	lr := io.LimitReader(r.sr, int64(n))
 
 	m := &DigestManifest{}
-	return m, json.NewDecoder(lr).Decode(m)
+	if err := json.NewDecoder(lr).Decode(m); err != nil {
```
stuartcarnie: given this is an exported method, you may want to add a guard at the entry point that returns an error if the manifest has already been read and this API is called again.

```go
if r.manifestRead {
    return nil, <some error>
}
```

dgnorton: Good point. What do you thing about caching the manifest and returning it on future reads? Would that make sense or should that be up to the caller?

stuartcarnie: Sounds fair to me, if you think that is a valid use case

dgnorton: After thinking about it, returning an error probably makes more sense with a method named `Read...` and it's easy enough for the caller to store it if needed.

PR #10414, opened 2018-10-25
reuse ValuerEval objects

seebs:
Scanner objects and iterators often need a ValuerEval. This
object is created, often with a function call, and has at
least one interface in it, so it allocates storage. Then it's
dropped again right away. The only part of it that might be
subject to change is usually a map. While the map's contents
change over time, the actual map doesn't change for the
lifetime of the object.

So, in both iterators and scanners, stash the ValuerEval
and continue reusing it. On a query returning a fair number
of data points, this produces a small (<5% in practice)
improvement in observed performance, visible as a significant
reduction in time spent in runtime (mallocgc, newobject,
etcetera).

The performance improvement isn't big, but it's reasonably
easy to evaluate it and establish that it's a safe change
to make.

Signed-off-by: seebs <seebs@seebs.net>

[none apply]

Notes:
* I didn't create an issue for this, but it's a side-effect of getting curious about query-response performance and wanting to do science.
* This is mostly a thing I'm messing with in order to learn my way around the code because I had thoughts about possibly trying to reduce the number of `interface{}` being slung around.
* The performance difference is pretty small, but it is more noticeable if you use msgpack. Still pretty small.

### DIFF
--- a/tsdb/engine/tsm1/iterator.gen.go
+++ b/tsdb/engine/tsm1/iterator.gen.go
@@ -196,6 +196,7 @@ type floatIterator struct {
 	statsLock sync.Mutex
 	stats     query.IteratorStats
 	statsBuf  query.IteratorStats
+	valuer    influxql.ValuerEval
 }
 
 func newFloatIterator(name string, tags query.Tags, opt query.IteratorOptions, cur floatCursor, aux []cursorAt, conds []cursorAt, condNames []string) *floatIterator {
@@ -222,6 +223,13 @@ func newFloatIterator(name string, tags query.Tags, opt query.IteratorOptions, c
 	}
 	itr.conds.names = condNames
 	itr.conds.curs = conds
+
+	itr.valuer = influxql.ValuerEval{
+		Valuer: influxql.MultiValuer(
+			query.MathValuer{},
+			influxql.MapValuer(itr.m),
+		),
+	}
 
 	return itr
 }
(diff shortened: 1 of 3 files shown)


jsternberg:
At first glance it seems good. We're doing a release very soon though so I'm going to circle back to this later and double check if it's ok to merge.

seebs:
Seems reasonable. In terms of actual memory allocation load, it's basically gonna be invisible compared to the interfaces anyway. :P

e-dard:
@seebs we will look at this for the next release. Can you remove your changes to the `CHANGELOG`? We handle editing that. @jsternberg can you make a review if you get time?

seebs:
Updated with the changelog thing dropped. This doesn't apply to master, I think because of code reorganization, but I haven't looked very closely.

jsternberg:
I'll take care of rebasing it  if needed. Thanks for the help.

I'll be looking at this tomorrow so it will likely be merged then.

tsdb/engine/tsm1/iterator.gen.go
```diff
@@ -223,6 +224,17 @@ func newFloatIterator(name string, tags query.Tags, opt query.IteratorOptions, c
 	itr.conds.names = condNames
 	itr.conds.curs = conds
 
+	// note: if opt.Condition is nil, itr.m can be nil.
```
jsternberg: You can remove this comment. Passing `nil` to the `MapValuer` should be fine regardless. Assigning to it would cause a panic, but that's not relevant to the valuer.

seebs: Good point, that hadn't occurred to me.

PR #10428, opened 2018-10-26
update flux dependency after planner integration

jlapacik:

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -397,7 +397,7 @@
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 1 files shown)


jlapacik:
@nathanielc @stuartcarnie there seems to be some non-deterministic behavior coming from the `tests` package where a test will fail if a service that it uses isn't up. This has only occurred for me as part of CI. I'm not sure if this is normal behavior. I'm hesitant to merge this update if this is not known or expected behavior.

jsternberg:
If the test that's failing has nothing to do with flux, then this is likely good to merge.

PR #10436, opened 2018-10-30
Handle msgpack output in client

seebs:
This adds support for requesting JSON or msgpack encoding
in the client, and decoding msgpack format data directly.

The code was written pretty much straight through top-to-bottom
by keeping a copy of response_writer.go up and comparing
bits and pieces, and it's not been carefully tested on edge
cases or everything.

The reason I think it's worth sending in anyway is
that executing the same query both ways gets runtime
of:

	JSON: 15.855865892s
	msgpack: 6.881726139s

this is on the client side, but previous experiments with
curl suggest that using msgpack roughly *halves* wall-clock
time to respond to a request, not to mention significantly
reducing allocation load. It's even lower bandwidth
usage!

So the advantage here isn't just that the client's faster;
it's that each client using msgpack *also* reduces load on
the server. And given that the server can respond to the same
query in half as much time using msgpack, it seems to me
that quite possibly, disk access and database reads are *not*
dominating runtime in these trivial cases.

Code is probably full of typos and inadequately tested. All
I've verified is that a quick hacked-together thing that tries
to do a deep compare of two Result objects says they're
the same.

... I haven't done this, and probably should, but I figure first I should have this up so people can have a look at it.

Other issues: The attempt to figure out what might be an error message on the end of msgpack data we couldn't parse is probably bad. I am not sure I can do better in a sane way.

### DIFF
--- a/client/v2/client.go
+++ b/client/v2/client.go
@@ -16,8 +16,10 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode"
 
 	"github.com/influxdata/influxdb/models"
+	"github.com/tinylib/msgp/msgp"
 )
 
 // HTTPConfig is the config data needed to create an HTTP Client.
@@ -48,6 +50,10 @@ type HTTPConfig struct {
 
 	// Proxy configures the Proxy function on the HTTP client.
 	Proxy func(req *http.Request) (*url.URL, error)
+
+	// Encoding specifies the encoding to use.
+	// Values are "json", "msgpack", with "" treated as "json".
+	Encoding string
 }
 
 // BatchPointsConfig is the config data needed to create an instance of the BatchPoints struct.
@@ -80,7 +86,7 @@ type Client interface {
 
 	// QueryAsChunk makes an InfluxDB Query on the database. This will fail if using
 	// the UDP client.
-	QueryAsChunk(q Query) (*ChunkedResponse, error)
+	QueryAsChunk(q Query) (ChunkedResponse, error)
 
 	// Close releases any resources a Client may be using.
 	Close() error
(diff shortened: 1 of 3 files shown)


seebs:
d'oh. i somehow wrote this against the tree without the recent QueryAsChunk change. I'll rebase.

seebs:
So that was an adventure: My implementation worked only for non-chunked usage, because `msgp.Decode()` creates a new reader *with an internal buffer*, then discards the Reader, and its buffer, after it's done reading one object. Fixed that. Also now checked performance chunked. Not as impressive, but still an improvement:

JSON: 8.772141564s
msgpack: 5.512790251s

dswarbrick:
I'm always happy to see people realize how inefficient and CPU intensive JSON is for bulk data transfer. But since InfluxDB is primarily a Go-based project, wouldn't it make more sense to use Protobuf?

seebs:
It might, but the server *already supports* msgpack and csv, so adding msgpack to the client was a trivial patch I could do in something like two hours.

dswarbrick:
Ah, I was not aware that msgpack was already implemented in the backend. As an aside, I've just realized (to my disappointment and surprise) that the Go client/v2 does not implement gzip encoding for the line protocol (despite the server supporting it). This would also be about a 10:1 bandwidth reduction. Telegraf seems to implement its own InfluxDB client, which _does_ support gzip.

seebs:
Huh, that would be interesting to test. I don't know what size uploads are most common, so I'm not sure how significant gzip would be; it's obviously way more for large data sets than small ones.

dswarbrick:
From my own personal experience, after writing an InfluxDB writer plugin for collectd, we observed compression ratios of between 8:1 and 10:1 when using 256 Kb batch size. The data was typical host metrics such as CPU load, memory util, disk / network IO etc. Obviously the compression efficiency will vary depending on how dynamic the values are, cardinality of labels etc.

e-dard:
@seebs are you interested in getting this back on track?

seebs:
I think so? I have limited free time these days, but it seemed like a pretty easy patch.

PR #10459, opened 2018-11-06
Update flux dep constraint to 0.7.0

jsternberg:


jsternberg:
We are going to reset the 1.7 branch to the current master so this isn't needed.

PR #10462, opened 2018-11-07
Improve Unix socket handling

guillemj:

- [x] Update man page when modifying a command
- [x] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): issue filed or pull request submitted <https://github.com/influxdata/docs.influxdata.com/pull/1967>

This PR includes several changes to improve the Unix socket support usage. I'm not sure how you'd like to handle the change in default though, but given that it required manual specification before it might not be such a big issue?

### DIFF
--- a/client/influxdb.go
+++ b/client/influxdb.go
@@ -29,6 +29,9 @@ const (
 	// DefaultPort is the default port used to connect to an InfluxDB instance
 	DefaultPort = 8086
 
+	// DefaultUnixSocket is the default unix socket used to connect to an InfluxDB instance
+	DefaultUnixSocket = "/run/influxdb/influxdb.sock"
+
 	// DefaultTimeout is the default connection timeout used to connect to an InfluxDB instance
 	DefaultTimeout = 0
 )
--- a/cmd/influx/cli/cli.go
+++ b/cmd/influx/cli/cli.go
@@ -94,6 +94,19 @@ func (c *CommandLine) Run() error {
 		return errors.New("unable to prompt for a password with no TTY")
 	}
 
+	if c.ClientConfig.UnixSocket == "" && c.Host == "" && c.Port == 0 {
+		fi, err := os.Stat(client.DefaultUnixSocket)
+		if (err == nil && fi.Mode() & os.ModeSocket != 0) {
+			c.ClientConfig.UnixSocket = client.DefaultUnixSocket
+		}
+	}
+	if c.Host == "" {
+		c.Host = client.DefaultHost
+	}
+	if c.Port == 0 {
+		c.Port = client.DefaultPort
+	}
+
 	// Read environment variables for username/password.
 	if c.ClientConfig.Username == "" {
 		c.ClientConfig.Username = os.Getenv("INFLUX_USERNAME")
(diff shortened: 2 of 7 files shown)


guillemj:
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you for your contributions.

Err, ping then? :/

guillemj:
Ping again…

jsternberg:
Can you describe what is being changed with this PR? I don't see anything except for changing defaults and the path location for stuff and I'm very hesitant to change any defaults. Is there something else this does? I see that it adds the `-socket` option to the man page which is really good, but I don't see anything else in the PR.

guillemj:
Did you check the commit messages? I'm happy to expand or clarify those, or describe further in here anything that's not clear from these. :)

jsternberg:
Ah I had not. I only usually read the PR message, not the individual commits.

I don't think this is a change we can accept. I understand the desire is to have it so if the socket is available that it will default instead of being forced to use `influx -socket`, but this changes a default behavior in a way that I think might be a bit too risky since the existence of the file forces the client to use the unix socket instead of the tcp socket when the previous behavior was to always use the tcp socket. And I'm not going to be able to change the default unix socket location between a minor version.

I apologize for taking so long to respond to the PR just to give that response. We are actively trying to be better at responding to PRs. If you change the PR just to document the `-socket` option in the man page though, that's something that's easy to approve and accept.

guillemj:
Ok, so I think the patch to try the unix socket should be quite safe, as the default for the influxd daemon is to not enable the unix socket, so that it will not be present and will not be picked up by default by the influx CLI. In addition the defaul tcp address in the CLI is localhost, so I don't think it would be an issue even if the unix socket is present, as both would (supposedly) be talking  to a local instance. Well I guess you could have a influx-relay or a port redirection in place but that seems rather odd?

I agree the change to the default location for the unix socket in the daemon is a potential breaking change. And that's why I mentioned it in the initial MR description. I still think it makes sense as otherwise you have a hard time running the daemon as non-root. So I'd like to see this merged in eventually, I don't mind when though.

If by splitting this into multiple MR makes merging this incrementally more feasible, then I'm happy to do that, but I'd really like to see these changes in at some point. We currently use a forked influxdb so that we can apply this changes.

jsternberg:
If I may ask, what is the particular use case for this? My assumption has been this is just to make it easier for you to run `influxd` using a unix socket and connect to it without using any command line flags. Possibly for making local development easier. Can you tell me more about your use case for why this is needed? I may be able to propose a solution that would be able to help in your situation.

I do want to be honest with you though. I am unlikely to merge this PR mostly because the benefit is a small bit of convenience for someone using the unix socket feature, which isn't a very prominent one to my knowledge, and a change in behavior for those who don't use the feature. Even if it is unlikely, choosing to connect to the unix socket if it is present before localhost changes the default behavior and can lead to an unintentional attack on a client that could potentially be writing data to the wrong place. I know it doesn't seem like much, but it does add a potential attack angle to the CLI tool.

jsternberg:
I'll pay attention to any replies, but I'm going to close this because we're not going to accept this contribution. Thank you for submitting a PR and I'm happy to help give advice on how to make your workflow easier if I can.

PR #10474, opened 2018-11-12
Fix go:generate line to refer to correct proto file

### LINKED ISSUE #10473: storage.proto missing from the source
Hello!

It looks like ``services/storage/storage.proto`` is missing from the source.

I am getting the following when running go generate:
```
src/github.com/influxdata/influxdb/services/storage/gen.go
storage.proto: No such file or directory
```

stuartcarnie:
Closes #10473


stuartcarnie:
Fixed in 1.7 via 5d675c5

PR #10477, opened 2018-11-13
time weighted average implementation

gpomykala:
This adds support for "time weighted average" as requested in #7445.

### Example usage
**Sample data from [https://s3.amazonaws.com/noaa.water-database/NOAA_data.txt](url) **

> select water_level from h2o_feet where "location" = 'santa_monica' limit 20
name: h2o_feet
time			water_level
----			-----------
1439856000000000000	2.064
1439856360000000000	2.116
1439856720000000000	2.028
1439857080000000000	2.126
1439857440000000000	2.041
1439857800000000000	2.051
1439858160000000000	2.067
1439858520000000000	2.057
1439858880000000000	1.991
1439859240000000000	2.054
1439859600000000000	2.018
1439859960000000000	2.096
1439860320000000000	2.1
1439860680000000000	2.106
1439861040000000000	2.146
1439861400000000000	2.1
1439861760000000000	2.136
1439862120000000000	2.182
1439862480000000000	2.306
1439862840000000000	2.448

**time_weighted_average:**

> select time_weighted_average("water_level") from h2o_feet where "location" = 'santa_monica' and time <= '2015-08-18T01:54:00Z'  group by time(10m)
name: h2o_feet
time			time_weighted_average
----			---------------------
1439856000000000000	2.0848
1439856600000000000	2.0652
1439857200000000000	2.075
1439857800000000000	2.0574
1439858400000000000	2.0458
1439859000000000000	2.0288
1439859600000000000	2.0492
1439860200000000000	2.1004
1439860800000000000	2.13
1439861400000000000	2.1144
1439862000000000000	2.1976
1439862600000000000	2.3912

- [X] Rebased/mergable
- [X] Tests pass
- [x] CHANGELOG.md updated

- [X] Provide example syntax
- [ ] [InfluxData Documentation](https://github.com/influxdata/docs.influxdata.com): issue filed or pull request submitted


nyhu:
Hello @gpomykala,

Thanks for this feature that will be most useful for iot developers.

I will not be able to review code correctly since I miss the whole influxdb implementation context, but I can review tests:

Could you add cases where there are missing points on a window?

`{
			name: "TimeWeightedAverage_Integer_Grouping_OneSamplePerGroup_WithMissingPoint",
			q:    `SELECT time_weighted_average(value) FROM cpu WHERE time >= '1970-01-01T00:00:00Z' AND time < '1970-01-01T00:00:14Z' GROUP BY time(2s)'`,
			typ:  influxql.Integer,
			itrs: []query.Iterator{
				&IntegerIterator{Points: []query.IntegerPoint{
					{Name: "cpu", Time: 0 * Second, Value: 1},
					{Name: "cpu", Time: 1 * Second, Value: 2},
					{Name: "cpu", Time: 5 * Second, Value: 3},
				}},
			},
			rows: []query.Row{
				{Time: 0 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(1.5)}},
				{Time: 2 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(2)}},
				{Time: 4 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(2.5)}},
			},
		},`

I'll add more request if I'm welcome.
Best regards

gpomykala:
> Hello @gpomykala,
> 
> Thanks for this feature that will be most useful for iot developers.
> 
> I will not be able to review code correctly since I miss the whole influxdb implementation context, but I can review tests:
> 
> Could you add cases where there are missing points on a window?
> 
> ```
> {
> 	name: "TimeWeightedAverage_Integer_Grouping_OneSamplePerGroup_WithMissingPoint",
> 	q:    `SELECT time_weighted_average(value) FROM cpu WHERE time >= '1970-01-01T00:00:00Z' AND time < '1970-01-01T00:00:14Z' GROUP BY time(2s)'`,
> 	typ:  influxql.Integer,
> 		itrs: []query.Iterator{
> 			&IntegerIterator{Points: []query.IntegerPoint{
> 				{Name: "cpu", Time: 0 * Second, Value: 1},
> 				{Name: "cpu", Time: 1 * Second, Value: 2},
> 				{Name: "cpu", Time: 5 * Second, Value: 3},
> 		}},
> 	},
> 	rows: []query.Row{
> 		{Time: 0 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(1.5)}},
> 		{Time: 2 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(2)}},
> 		{Time: 4 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(2.5)}},
> 	},
> },
> ```
> I'll add more request if I'm welcome.
> Best regards

Hello @nyhu ,
Thanks for your feedback. Your test case revealed a bug in my implementation - it does return rows with nil values.
{[]query.Row}[?->5]:
    			-: <non-existent>
    			+: query.Row{Time: 10000000000, Series: query.Series{Name: "cpu", id: 0x01}, Values: []interface {}{interface {}(nil)}}

I would appreciate if you could suggest more scenarios

nyhu:
Hello @gpomykala 
Sorry for the delay.

Before adding many test cases I would gladly have a feedback from a influx team member for the implementation context. I'll explain progressively.

The test case I suggest here raise the main issue I see for weighted_average : the maintained state AKA PrevElement(type)Iterator. I suppose it's a very common process for event based data but not yet for influxdb aggregation function.

This need for a state should appear in moving_average transformation implementation if a example can help you validate your current implementation. (my understanding of transformation/aggregation comes mainly from:
https://docs.influxdata.com/influxdb/v1.7/query_language/functions)

I saw you generate most of it and it should please the actual reviewer. I personally think it's the right and only way.

Weighted_average, by the test case I suggested enter in conflict with fill() management. fill values "previous" and "linear" loses their meaning and "none" / "null" are usefull only when there is no data at the beginning of dataset. Filled value on interval without timestamp are filled with last known raw value and not last aggregated value. In fine we end up with a new kind of fill: "previous_raw". The problem with previous raw is that the state is extended to fill management -> code contagion. That's probably why this very useful feature is not yet implemented. (It was my first tough on this matter five month ago).

Before extensive tests are done, I suggest we talk about fill to be sure it will be coherent for everyone.
From my usecase perspective, if weighted average is used for event base data, there is no such thing as "no data" on a GROUP BY time(), there is "unchanged then not sent" data. Then "previous_raw" doesn't really exist and "previous"/"linear" just have nothing to fill. It's the easiest/non-breaking way, I hope the right way.

This behavior is currently unreachable with influx and a real pain for iot world. We currently have two solutions : precalculate aggregation from another service and load our database with terabytes of raw datas. Or do a rough estimation by doing statistical average on a subquery (mean(value) group by time(1s)). This is wrong but close.

As for remaining missing tests, they are designed by:
- no data on AND before the requested period should end up with nil
- data before BUT not on the requested period should end up with the last value repeated again and again.
Otherwise Tests I reviewed seems to cover pretty well!

We could converse on https://github.com/influxdata/influxdb/issues/7445 if it's more appropriate.

See you and thanks for all the fish.

nyhu:
As a response for your edition, now that you understand my point of view:
```
{
      name: "TimeWeightedAverage_Integer_Grouping_OneSamplePerGroup_WithMissingPoint",
      q:    `SELECT time_weighted_average(value) FROM cpu WHERE time >= '1970-01-01T00:00:00Z' AND time < '1970-01-01T00:00:14Z' GROUP BY time(2s)'`,
      typ:  influxql.Integer,
        itrs: []query.Iterator{
          &IntegerIterator{Points: []query.IntegerPoint{
            {Name: "cpu", Time: 0 * Second, Value: 1},
            {Name: "cpu", Time: 1 * Second, Value: 2},
            {Name: "cpu", Time: 5 * Second, Value: 3},
        }},
      },
      rows: []query.Row{
        {Time: 0 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(1.5)}},
        {Time: 2 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(2)}},
        {Time: 4 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(2.5)}},
        {Time: 6 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(3)}},
        {Time: 8 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(3)}},
        {Time: 10 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(3)}},
        {Time: 12 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(3)}},
      },
    },
```

gpomykala:
Hi,
I've taken your comments onboard.
Regarding your proposed case:
```
{
      name: "TimeWeightedAverage_Integer_Grouping_OneSamplePerGroup_WithMissingPoint",
      q:    `SELECT time_weighted_average(value) FROM cpu WHERE time >= '1970-01-01T00:00:00Z' AND time < '1970-01-01T00:00:14Z' GROUP BY time(2s)'`,
      typ:  influxql.Integer,
        itrs: []query.Iterator{
          &IntegerIterator{Points: []query.IntegerPoint{
            {Name: "cpu", Time: 0 * Second, Value: 1},
            {Name: "cpu", Time: 1 * Second, Value: 2},
            {Name: "cpu", Time: 5 * Second, Value: 3},
        }},
      },
      rows: []query.Row{
        {Time: 0 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(1.5)}},
        {Time: 2 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(2)}},
        {Time: 4 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(2.5)}},
        {Time: 6 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(3)}},
        {Time: 8 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(3)}},
        {Time: 10 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(3)}},
        {Time: 12 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(3)}},
      },
    },
`

Your assertion is met when FILL(linear) is applied. With implicit FILL(null) you'll still get as specified in [:](https://docs.influxdata.com/influxdb/v1.7/query_language/data_exploration/#group-by-time-intervals-and-fill). If you disagree, please compare the fill behaviour with other aggregation ie. MEAN
```
       {Time: 2 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{nil}},
`

I've added test cases with all FILL options.

Regarding your suggested test cases:
* no data on AND before the requested period should end up with nil DONE
* data before BUT not on the requested period should end up with the last value repeated again and again. INVALID since WHERE constraint precedes any aggregation therefore therefore aggregate function gets no data to get the previous values from.

Best regards,
Grzegorz

nyhu:
Hello @gpomykala,

I add hopes and doubts at the same time when you told me fill() will solve this.
Sadly I stick on my position, fill('previous') repeat the previous already aggregated value, in fact, the result is wrong.

Let me give a true story example based on a sensor data I saw recently. Simplified to fit with our dear test case.
```
{
      name: "TimeWeightedAverage_Integer_Grouping_OneSamplePerGroup_WithMissingPoint",
      q:    `SELECT time_weighted_average(value) FROM cpu WHERE time >= '1970-01-01T00:00:00Z' AND time < '1970-01-01T00:00:14Z' GROUP BY time(2s)'`,
      typ:  influxql.Integer,
        itrs: []query.Iterator{
          &IntegerIterator{Points: []query.IntegerPoint{
            {Name: "cpu", Time: 0 * Second, Value: 30000}, // machine reach cap value
            {Name: "cpu", Time: 0.4 * Second, Value: 29988},
            {Name: "cpu", Time: 0.6 * Second, Value: 29998},
            {Name: "cpu", Time: 0.8 * Second, Value: 30000},  // machine stops on cap value
            {Name: "cpu", Time: 0.9 * Second, Value: 0}, // in real scenario this is more than a few seconds
            {Name: "cpu", Time: 8 * Second, Value: 12345}, // machine restarts 
            {Name: "cpu", Time: 10 * Second, Value: 30000}, // machine stays on cap
        }},
      },
      rows: []query.Row{
        {Time: 0 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(
           30000 * 0.4 + 29988 * 0.2 + 29998 * 0.2 + 30000 * 0.1 + 0 * 0.1)}},
        {Time: 2 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(0)}},
        {Time: 4 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(0)}},
        {Time: 6 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(0)}},
        {Time: 8 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(12345)}},
        {Time: 10 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(30000)}},
        {Time: 12 * Second, Series: query.Series{Name: "cpu"}, Values: []interface{}{float64(30000)}},
      },
    },
```
A fill('previous') break any data analisis. Obviously the machine is at stop between 2 and 8 seconds. To have a zero here is important. 
A fill(0) assume that the machine is stopped if there is no data. But the machine can be a long time on capping value -> At 12 second a fill(0) would virtually stop the machine.

Maybe I overstep with my user story, but it still seems legitimate for iot scenario.  
Handling each machine separately with fill is sometimes impossible like the upper case. And even when possible it would require a lot of experience with machine behavior and human leverage to make it work properly. I would not feel perfectly safe with your current implementation but would still use it -> most of the time I would gain in precision.

Again, if it was simple, the influx team would probably have done it 2 years ago. With my request, it's like a specific aggregation that does not behave like others on an empty group by time() so I understand your point of view.

For the second part of the discussion:
>  INVALID since WHERE constraint precedes any aggregation therefore therefore aggregate function gets no data to get the previous values from.

I can live with that fact. If you do not have the data, I won't ask to change the whole influx implementation. Your proposition is such a gain in precision, having a first point with reduced precision is almost irrelevant. It's good to know and should be specified in documentation as a known issue.

If you need me to dive into the code more and more, just tell me, I'd hope some influx commando would rush in. I can give you a working branch with my proposal with time and if I'm not mistaken on the concept.

Bests regards.

nyhu:
PS: The tests with fill you added are testing fill behavior, seems to me they are not necessary in this implementation.

query/select_test.go
```diff
@@ -17,6 +17,808 @@ import (
...
+			typ:  influxql.Float,
+			itrs: []query.Iterator{
+				&FloatIterator{Points: []query.FloatPoint{	
+					{Name: "cpu", Time: 0 * Second, Value: 2},				
+				}},
+			},
+			rows: []query.Row(nil),
+		},		
+		{
+			name: "TimeWeightedAverage_Float_NoGrouping_ValuesOnlyAfterRequestedWindow",
```
nyhu: Sample doesn't fit with test name. Should be : 
`{Name: "cpu", Time: 14 * Second, Value: 2},`

nyhu:
I verified every result and match expected behavior (expect our conversation of course). Job well done.

e-dard:
@nathanielc what do you think, and does this impact flux parity?

nathanielc:
@e-dard I took a quick look at the code looks good. We will need to reimplement for Flux but that shouldn't matter.

query/call_iterator.go
```diff
@@ -53,6 +53,8 @@ func NewCallIterator(input Iterator, opt IteratorOptions) (Iterator, error) {
 		return newLastIterator(input, opt)
 	case "mean":
 		return newMeanIterator(input, opt)
+	case "time_weighted_average":
```
jsternberg: This section is meant for functions that can be pushed down to the storage engine so anything that is in this section will effectively be called twice. All of the implementations here are idempotent except for `count` which has a special implementation so that count is performant.

For a function to be in this section, this must be true:

```
X(X(y)) = X(y)
```

Count is special cased so that it does this:

```
sum(count(y)) = count(y)
```

Here's the place where that happens: https://github.com/influxdata/influxdb/blob/e9bada090f10ea04f07ab56a839690ab15328269/query/iterator.go#L139

It is much safer to add this to the appropriate place in `query/select.go`.

query/call_iterator.go
```diff
@@ -510,6 +512,71 @@ func NewDistinctIterator(input Iterator, opt IteratorOptions) (Iterator, error)
...
+	}
+}
+
+func newUnsignedReduceFloatWithPrevElementIterator(input UnsignedIterator, opt IteratorOptions, createFn func(it UnsignedPrevIterator) (UnsignedPointAggregator, FloatPointEmitter)) *unsignedReduceFloatIterator {
+	prevIt := newPrevElementUnsignedIterator(input)
+	it := newBufUnsignedIterator(prevIt)
+	return &unsignedReduceFloatIterator{
+		input: it,
+		create: func() (UnsignedPointAggregator, FloatPointEmitter) {
+			return createFn(prevIt)
```
jsternberg: I think this is wrong, but I'm not aware enough of how time weighted averages work. The one thing I do know is that this seems to be attempting to retrieve a value through state that was intentionally kept separate for this specific iterator.

The `mean()` operator and others relating to it are intentionally separated into their own buckets. They are not allowed to communicate with each other. That's not necessarily correct for this code to use and you'll run into potential bugs if we try to get around that structure.

But, I don't know enough about time weighted average to say which method should be used. So here are the three methods that exist that may be applicable.

1. The method is completely bucketed. The buckets have nothing to do with each other. These are your standard aggregation/selector functions (`mean()`, `max()`, `min()`, `first()`, `last()`, etc).
2. The method has one result per bucket, but the points from a previous bucket may influence the value of the current bucket. The only existing function that does this is `integral()`.
3. The method is not bucketed at all. It consumes a point and produces a point (sometimes). It may produce fewer points than the original stream, but it is mostly 1 point in, 1 point out. The `derivative()`, `difference()`, and `moving_average()` functions are the best example of this. These methods may _operate_ on bucketed data, but they don't care about whether values are in buckets by themselves.

I'm thinking `integral()` is the best example of what you probably need, but I don't know if the result should be bucketed or not.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ v1.7.0 [unreleased]
 -	[#9964](https://github.com/influxdata/influxdb/pull/9964): Enable the storage service by default.
 -	[#9996](https://github.com/influxdata/influxdb/pull/9996): Ensure read service regexes get optimised.
 -	[#10408](https://github.com/influxdata/influxdb/pull/10408): Add Flux support to the influx CLI command.
+-	[#10477](https://github.com/influxdata/influxdb/pull/10477): Support for time_weighted_average aggregation
```
jsternberg: We'll handle the changelog so you can remove this.

query/call_iterator.go
gpomykala: Thanks for a tip, i will be happy to move it into buildCallIterator in query/select.go. Although there are 2 separate places where switch expr.Name takes place and I am not sure why is that the case and which section time_weighted_average belongs to

query/call_iterator.go
gpomykala: "The method has one result per bucket, but the points from a previous bucket may influence the value of the current bucket." This is valid for time weighted averages. I will try to change the implementation

gpomykala: time_weighted_average reimplemented with streaming iterator - similarly to integral

gpomykala:
The requested changes had been submitted to the PR - changes pending review

pon., 22.07.2019, 23:32 użytkownik stale[bot] <notifications@github.com>
napisał:

Talon876:
Are there any plans of continuing forward with this feature?

inselbuch:
Time-weighted averages are a critical and core feature for a time-series database.

Sdelausnay:
When will the feature be available?

fgheysels:
It would be sad if implementation of this feature would stop, given the time and effort that has already been invested.   Is there a reason why the PR hasn't been merged / reviewed any further ?  I'm not a Go developer, but otherwise I'd like to contribute.

PR #10502, opened 2018-11-20
No rebuild series index on delete when the series still exists in cache

### LINKED ISSUE #10503: Delete rebuilds series index when series to be deleted are only found in cache
__System info:__ 
InfluxDB 1.7.0

__Steps to reproduce:__

1. Start influxd with tsi1
2. Run `insert someMeasurement value=1` in the influx client to create a new series that only lives in the cache.
3. Run `delete from "someMeasurement" where time=0`  in the influx client (which deletes nothing because I do not have any data with time=0)

Influx now decides to rebuild the series index files because the series "someMeasurement" does not exist in any tsm file so it is considered "deleted".

If I do a restart of InfluxDB between 2. and 3. it will not rebuild the series on the delete.

This issue is related to #9015, because it causes deletes to be very slow when the series only exist in cache. I get the profile below when running a query like `delete from "someMeasurement" where time=0` in a loop.

image

__Expected behavior:__
The series index should only be rebuilt when we actually delete data.

__Actual behavior:__ 
The series index is rebuilt if it only exist in cache.

__Additional info:__

hpbieker:
This PR fixes #10503.


hpbieker:
@e-dard please review?

hpbieker:
@rbetts please review?

e-dard:
@benbjohnson can you take a look?

benbjohnson:
@hpbieker This looks good. Thanks for the PR and sorry for the delay!

PR #10506, opened 2018-11-20
Update Platform dependency

stuartcarnie:
This PR includes some minor changes to simplify Flux support in Enterprise

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -445,8 +445,8 @@
   revision = "32c6aa80de5eb09d190ad284a8214a531c6bce57"
 
 [[projects]]
-  branch = "master"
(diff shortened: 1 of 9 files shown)


stuartcarnie:
1. The http service is reused in its entirety in Enterprise and there are no extension points to inject middleware or extend handlers. InfluxQL similarly parses the `node_id` in the OSS handler. We could do some refactoring to expose a way to inject middleware, but not sure that is good use of our time right now.

2. There are differences in how series keys and fields are handled between OSS and Platform, so there is a small amount of specialized code in that package.

e-dard:
@stuartcarnie makes sense. No need for any changes in my opinion.

Issue #9694, opened 2018-04-09
DROP MEASUREMENT "xyz" not working


mgf909:
### Bug report

Cannot delete measurments.
I accidentally created a bunch of MS SQL server measurements into my Telegraf db, and want to remove them. When i try to remove them with DROP MEASUREMENT "<name of measurement>" it doesnt actually seem to drop it...

My mistake seems to have created over 1000 measurements, so i suspect the bug/issue is due to the number of measurements.
If i create a new test database i can drop measurements as expected.

__System info:__
Influx 1.5.1 running on Red Hat 4.8.5-11

__Steps to reproduce:__

Create a lot of measurements - i did this by trying to use the MS SQL Server telegraf input.

1. USE telegraf
2.  DROP MEASUREMENT "xa7_sessions"  
It doesnt matter what measurement i specify, i cannot remove it. There is no error
3. SHOW MEASUREMENTS
Its still there!

__Expected behavior:__ 
Id expect the measurement to no longer be returned after running the DROP MEASUREMENT command.

__Actual behavior:__ 
SHOW MEASUREMENT continues to return the measurement..and will not die ;-(

__Additional info:__ [Include gist of relevant config, logs, etc.]

mgf909:
[telegraf_measurements.zip](https://github.com/influxdata/influxdb/files/1888591/telegraf_measurements.zip)

foxmask:
+1 I cant drop anything at all.
Even SERIES are still there :(

max3163:
Same issue too, no problem when we switch back to 1.3 version.

foxmask:
@max3163 thanks for the tips, it seems that 1.4.1 works fine too

binary0111:
I moved from 1.4.3 to 1.5.2, everything seems to work fine for me.

kichristensen:
Same issue here in version 1.5.2

yellowpattern:
In my experience, dropping a series is not always "instant". If a "drop" is not given the correct parameters, it can silently do nothing without an error.

This command has no observability and what it does (or doesn't) do is hidden from users. This needs to be improved and "drop series" made more transparent in what it does (or at least the option of having it behave that way.)

A more "verbose" drop command is required, telling you how many series were found to drop, including 0 if none are to be dropped.

Plus include something in the log file - maybe write out to the log file the name of each series dropped and the total number dropped.

fhriley:
Same here. Can't delete old measurements I don't want:

```
[...]

PR #10509, opened 2018-11-20
tsdb: don't allow deletes to a database in mixed index mode

zeebo:
TSI1 and inmem indexes have different properties during deletes.
Specifically, inmem shares a global index across all shards, where
every tsi1 index is contained to a specific shard. When deleting
a series, it may cause the last reference to the series across all
shards to be dropped, necessitating a removal from the series file.
Since the inmem index shares the index across all shards, removing
the series when it's removed from the series file is sufficient.

However, in the case of a mixed index database, if the last shard
is a TSI1 shard, the other inmem indexes are not available when we
discover that it was the last reference to the series. This ends
up leaving the series in the inmem index without a series id in
the series file, causing all sorts of misbehavior.

Rather than continue curling ourselves into a ball to try to fix
this unsupported mode, give a helpful error message to the user
that they must run their database in a non-mixed index mode to
allow deletes.

Updates #9694.

### DIFF
--- a/tsdb/store.go
+++ b/tsdb/store.go
@@ -33,6 +33,9 @@ var (
 	ErrStoreClosed = fmt.Errorf("store is closed")
 	// ErrShardDeletion is returned when trying to create a shard that is being deleted
 	ErrShardDeletion = errors.New("shard is being deleted")
+	// ErrMultipleIndexTypes is returned when trying to do deletes on a database with
+	// multiple index types.
+	ErrMultipleIndexTypes = errors.New("cannot delete data. DB contains shards using both inmem and tsi1 indexes. Please convert all shards to use the same index type to delete data.")
 )
 
 // Statistics gathered by the store.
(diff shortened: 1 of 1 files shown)


tsdb/store.go
```diff
@@ -33,6 +33,9 @@ var (
 	ErrStoreClosed = fmt.Errorf("store is closed")
 	// ErrShardDeletion is returned when trying to create a shard that is being deleted
 	ErrShardDeletion = errors.New("shard is being deleted")
+	// ErrMultipleIndexTypes is returned when trying to do deletes on a database with
+	// multiple index types.
+	ErrMultipleIndexTypes = errors.New("cannot delete with multiple index types")
```
e-dard: Can we expand on this a bit? I think this is going to surprise some users, so we should be quick to help resolve the situation for them. Maybe something like:

```
cannot delete data. DB contains shards using both inmem and tsi1 indexes. Please convert all shards to use the same index type to delete data. 
```

Or something like that...

gshif:
@e-dard, Usually, CHANGELOG is updated automatically if the PR is labeled correctly (`kind/bug`, `kind/feature`), but I can always put an extra info if needed.

PR #10513, opened 2018-11-21
tsm1: disallow concurrent writes and deletes

zeebo:
There is a race described in the comments about how a delete
concurrent with a write can cause the index to miss the series.
The severity of that issue isn't described in the comment, though.

If a series is missing from the index, then it is invisible to future
deletes, meaning that you can have entries in the fields.idx for
series that don't exist, and data in the tsm file that is unreachable.
That is, until another write comes in for that series for that shard.
In that case, all of that ghost data and field constraints come back
causing great confusion for users, engineers, and the database.

There may be cheaper ways to fix the race, but this definitely
fixes the race, and we can iterate on making it faster while keeping
it correct. Additionally, leaving these races in makes it harder
to diagnose and reproduce other possible issues with deletes.

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -1497,6 +1497,12 @@ func (e *Engine) deleteSeriesRange(seriesKeys [][]byte, min, max int64) error {
 		max = math.MaxInt64
 	}
 
+	// This sucks, but writes while we're deleting are SUPER racy and can lead to amazing
+	// mismatches between what's on disk, in the cache, in the wal, in the index, in the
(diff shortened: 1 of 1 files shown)


e-dard:
@zeebo this is a fix for https://github.com/influxdata/influxdb/issues/9694#issuecomment-435488714 isn't it?

zeebo:
I've been able to reproduce the effects described in that issue concurrently writing and deleting some points, so it may indeed be a fix.

zeebo:
Closed in favor of #10516 which will do this with less contention.

PR #10516, opened 2018-11-22
tsdb: conflict based concurrency resolution

zeebo:
There are some problematic races that occur when deletes happen
against writes to the same points at the same time. This change
introduces guards and an epoch based system to coordinate these
modifications.

A guard matches a point based on the time, measurement name, and
some conditions loaded from an influxql expression. The intent
is to be as precise as possible without allowing any false
neagatives: if a point would be deleted, the guard must match it.
We are allowed to match more points than necessary, at the cost
of slowing down writes.

The epoch based system keeps track of outstanding writes and
deletes and their associated guards. When a delete operation
is going to start, it waits until all current writes are
done, and installs its guard, blocking all future writes that
contain points that may conflict with the delete. This allows
writes to disjoint points to proceed uncontended, and the
implementation is optimized for assuming there are few
outstanding deletes. For example, in the case that there are no
deletes, a write just has to take a mutex, bump a counter, and
compare a value against zero. The epoch trackers are per shard,
so that different shards never have to contend with one another.

Updates #9694.

### DIFF
--- a/tsdb/epoch_tracker.go
+++ b/tsdb/epoch_tracker.go
@@ -1,0 +1,147 @@
+package tsdb
+
+import (
+	"sync"
+)
+
+// TODO(jeff): using a mutex is easiest, but there may be a way to do
+// this with atomics only, and in a way such that writes are minimally
+// blocked.
+
+// epochTracker keeps track of epochs for write and delete operations
+// allowing a delete to block until all previous writes have completed.
+type epochTracker struct {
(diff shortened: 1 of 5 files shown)


zeebo:
Note this has a bit of a weird merge destination because it conflicts with changes in #10509, but I'm putting this up so that it's easiest to review. We can rebase and change the merge destination once #10509 is merged.

PR #10519, opened 2018-11-23
fix /debug/pprof url path

cfz:
the pprof [page index](https://golang.org/src/net/http/pprof/pprof.go) use relative path in href. the tailing '/' is necessary when navigating form the index page. otherwise, all links will be directed to /debug/*

### DIFF
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -330,7 +330,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	w.Header().Add("X-Influxdb-Version", h.Version)
 	w.Header().Add("X-Influxdb-Build", h.BuildType)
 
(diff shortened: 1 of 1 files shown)


e-dard:
@cfz thanks for the contribution. Unfortunately it's not clear to me what this PR attempts to do. Can you explain further what problem this solves?

cfz:
hi, sorry for the slow response. 
occasionally, i'm using web browser to open the debug urls, like "localhost:8086/debug/pprof", i will get a page (generated by `httppprof.Index()` ) containing links directing to the sub path like "block" or "goroutine" by a relative path
without the trailing backslash, the index page will be served at "/debug/pprof" and the "block" hyperlink will be pointing to "/debug/block", which  is not expected. the correct url for show the blocking time of mutex should be "/debug/pprof/block"
so my solution is to make the index page shown at "/debug/pprof/" (with the trailing backslash), and every relative path to the debug endpoint will get right

PR #10536, opened 2018-12-04
Limit force-full and cold compaction size.

benbjohnson:
This pull request limits the number of files that can be compacted in a single group when forcing a full compaction or when a shard becomes cold. This is to prevent too many files being compacted at the same time.

### DIFF
--- a/tsdb/engine/tsm1/compact.go
+++ b/tsdb/engine/tsm1/compact.go
@@ -39,6 +39,10 @@ const (
 	TSMFileExtension = "tsm"
 )
 
+// ColdCompactionSplitSize is the maximum number of files to compact together
+// when performing a compaction on a cold shard.
+const ColdCompactionSplitSize = 8
+
 var (
 	errMaxFileExceeded     = fmt.Errorf("max file exceeded")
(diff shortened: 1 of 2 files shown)


tsdb/engine/tsm1/compact.go
```diff
@@ -453,11 +453,11 @@ func (c *DefaultPlanner) Plan(lastWrite time.Time) []CompactionGroup {
 			return nil
 		}
 
-		group := []CompactionGroup{tsmFiles}
-		if !c.acquire(group) {
+		groups := splitCompactionGroups(tsmFiles, 4)
```
e-dard: Can we make `4` a constant somewhere up top? I wonder if `4` is also too low.

e-dard:
@benbjohnson  iIf we're doing as forced full compaction and we don't compact all the files in the shard then I think we would need to keep the `c.forceFull = true` wouldn't we? 

We don't want a forced full compaction to result in not all the TSM files being compacted down.

benbjohnson:
@e-dard Ready for re-review.

benbjohnson:
@e-dard I verified that the cold compaction eventually processes all the files (in groups of 8) and compacts them down to a single file.

Issue #10540, opened 2018-12-06
Influxd deadlocks while starting with corrupt timeseries file


gpomykala:
System info: InfluxDB 1.7rc0 + linux64 / arm32

Steps to reproduce:

1. Induce an error to any file in timeseries storage
2. Start Influxd

Expected behavior: InfluxDB should report an error and exit or recover

Actual behavior: Influxd process hangs forever

Additional info:

Deadlock in tsdb/series_file.go:68 - Open() method acquires a lock upon entry, error handling within tries to call Close() which also tries to acquire a lock before Open() released it.

The issue was introduced with #9308

Proposed fix in #10541

PR #10541, opened 2018-12-06
Fix open/close race in SeriesFile

gpomykala:
…() method

### DIFF
--- a/tsdb/series_file.go
+++ b/tsdb/series_file.go
@@ -65,6 +65,10 @@ func (f *SeriesFile) Open() error {
 		p := NewSeriesPartition(i, f.SeriesPartitionPath(i))
 		p.Logger = f.Logger.With(zap.Int("partition", p.ID()))
 		if err := p.Open(); err != nil {
+			f.Logger.Error("Unable to open time series data",
+				zap.String("path", f.path),
+				zap.Int("partition", p.ID()),
+				zap.Error(err))
 			f.doClose(false)
 			return err
 		}
--- a/tsdb/series_file.go
+++ b/tsdb/series_file.go
@@ -69,7 +69,7 @@ func (f *SeriesFile) Open() error {
 				zap.String("path", f.path),
 				zap.Int("partition", p.ID()),
 				zap.Error(err))
-			f.close(false)
+			f.close()
 			return err
 		}
 		f.partitions = append(f.partitions, p)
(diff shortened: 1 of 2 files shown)


tsdb/series_file.go
```diff
@@ -65,7 +65,11 @@ func (f *SeriesFile) Open() error {
 		p := NewSeriesPartition(i, f.SeriesPartitionPath(i))
 		p.Logger = f.Logger.With(zap.Int("partition", p.ID()))
 		if err := p.Open(); err != nil {
-			f.Close()
+			f.Logger.Error("Unable to open time series data",
```
e-dard: Please change this line to `"Unable to open series file"`

tsdb/series_file.go
```diff
@@ -74,12 +78,12 @@ func (f *SeriesFile) Open() error {
 	return nil
 }
 
-// Close unmaps the data file.
-func (f *SeriesFile) Close() (err error) {
```
e-dard: Please change this method to `close() (err error) {` and remove the refs lock call.

tsdb/series_file.go
```diff
@@ -89,6 +93,11 @@ func (f *SeriesFile) Close() (err error) {
 	return err
 }
 
+// Close unmaps the data file.
+func (f *SeriesFile) Close() (err error) {
+	return f.doClose(true)
```
e-dard: Instead of doing this, you can just do this:

```go
func (f *SeriesFile) Close() (err error) {
    f.refs.Lock()
    defer f.refs.Unlock()
    return f.close()
}
```

tsdb/series_file.go
gpomykala: changed as suggested

tsdb/series_file.go
gpomykala: changed as suggested

tsdb/series_file.go
gpomykala: changed as suggested

e-dard:
@gpomykala that's not going to build. `close(false)` should just be `close`.

gpomykala:
> @gpomykala that's not going to build. `close(false)` should just be `close`.

Sorry for that. forgot to add that change. fixed

jacobmarble:
Fixes #10540

PR #10546, opened 2018-12-07
clientv2 - QueryCtx()

hekmon:
A simple modification allowing to pass a context to the clientv2 when performing queries. This is most helpful when a client launches thru a reverse proxy a long read and closes the connection before the batch is finished.

Dummy example:
```golang
func readHandler(w http.ResponseWriter, r *http.Request) {
	// snappy decode
	// [...]
	// protobuff unmarshall
	// [...]

	// spawn our influx client
	influx, err := influxdb.New("http://<host>:8086", database, user, password)
	if err != nil {
		fmt.Println(err)
		if r.Context().Err() == nil {
			http.Error(w, fmt.Sprintf("can't spawn InfluxDB controller: %v", err),
				http.StatusInternalServerError)
		}
		return
	}
	defer func() {
		if err = influx.Close(); err != nil {
			fmt.Println(err)
		}
	}()

	// Build & perform request
	// [...]
	resp, err := influx.QueryCtx(r.Context(), req)
	if err != nil {
		fmt.Println("read query to influxdb failed:", err)
		if r.Context().Err() == nil {
			http.Error(w, fmt.Sprintf("read query to influxdb failed: %v", err),
				http.StatusInternalServerError)
		}
		return
	}
	// [...]

	// protobuff marshal
	// [...]
	// snappy encode
	// [...]
}
```

~I did not signed it, but this is a trivial PR :)~

- [x] Provide example syntax

### DIFF
--- a/client/v2/udp.go
+++ b/client/v2/udp.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"net"
@@ -104,6 +105,10 @@ func (uc *udpclient) Write(bp BatchPoints) error {
 }
 
 func (uc *udpclient) Query(q Query) (*Response, error) {
+	return nil, fmt.Errorf("Querying via UDP is not supported")
+}
+
+func (uc *udpclient) QueryCtx(ctx context.Context, q Query) (*Response, error) {
 	return nil, fmt.Errorf("Querying via UDP is not supported")
 }
 
--- a/client/v2/udp.go
+++ b/client/v2/udp.go
@@ -105,15 +105,15 @@ func (uc *udpclient) Write(bp BatchPoints) error {
 }
 
 func (uc *udpclient) Query(q Query) (*Response, error) {
-	return nil, fmt.Errorf("Querying via UDP is not supported")
+	return nil, fmt.Errorf("querying via UDP is not supported")
 }
 
 func (uc *udpclient) QueryCtx(ctx context.Context, q Query) (*Response, error) {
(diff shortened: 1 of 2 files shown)


timhallinflux:
I appreciate you think this is trivial.  But, we would prefer that the CLA is signed in order to accept your contribution.  We appreciate your contribution very much!!

hekmon:
I understand... And I just did :)

client/v2/udp.go
```diff
@@ -107,6 +108,10 @@ func (uc *udpclient) Query(q Query) (*Response, error) {
 	return nil, fmt.Errorf("Querying via UDP is not supported")
 }
 
+func (uc *udpclient) QueryCtx(ctx context.Context, q Query) (*Response, error) {
+	return nil, fmt.Errorf("Querying via UDP is not supported")
```
e-dard: Error messages in Go start with a lower case letter. `"querying via UDP ...`

hekmon: You are absolutly right. But I thought it would be best to keep the error message like the existing ones:
https://github.com/influxdata/influxdb/blob/71ac0134162d440702ced4824e0bd365ff540b2f/client/v2/udp.go#L108
https://github.com/influxdata/influxdb/blob/71ac0134162d440702ced4824e0bd365ff540b2f/client/v2/udp.go#L116

hekmon:
Should be good now (fixed the old ones as well).

e-dard:
Thanks for you contribution @hekmon!

Issue #10489, opened 2018-11-16
fill(previous) does not properly work with group by


Kasokaso:
__System info:__ [ influx version 1.6.2, windows]

__Steps to reproduce:__

> select * from channel_site_count where  time > '2018-11-12T10:50:00Z'and time <'2018-11-12T11:08:00Z'
name: channel_site_count

```
time                     count site
----                     ----- ----
2018-11-12T10:55:52.899Z 4     410-1%%730
2018-11-12T10:55:52.899Z 10    SIP%%730
2018-11-12T10:55:52.899Z 4     410-2%%730
2018-11-12T10:55:52.899Z 30    E1%%730
2018-11-12T10:55:58.047Z 4     410-2%%730
2018-11-12T10:55:58.047Z 30    E1%%730
2018-11-12T10:55:58.047Z 4     410-1%%730
2018-11-12T10:55:58.047Z 10    SIP%%730
2018-11-12T10:56:02.983Z 4     410-1%%730
2018-11-12T10:56:02.983Z 10    SIP%%730
2018-11-12T10:56:02.983Z 4     410-2%%730
2018-11-12T10:56:02.983Z 30    E1%%730
2018-11-12T10:56:08.161Z 4     410-2%%730
2018-11-12T10:56:08.161Z 30    E1%%730
2018-11-12T10:56:08.161Z 4     410-1%%730
2018-11-12T10:56:08.161Z 10    SIP%%730
2018-11-12T10:56:13.198Z 4     410-1%%730
2018-11-12T10:56:13.198Z 10    SIP%%730
2018-11-12T10:56:13.198Z 4     410-2%%730
2018-11-12T10:56:13.198Z 30    E1%%730
2018-11-12T10:56:18.12Z  4     410-2%%730
2018-11-12T10:56:18.12Z  27    E1%%730
2018-11-12T10:56:18.12Z  4     410-1%%730
2018-11-12T10:56:18.12Z  10    SIP%%730
2018-11-12T10:56:23.172Z 4     410-1%%730
2018-11-12T10:56:23.172Z 9     SIP%%730
2018-11-12T10:56:23.172Z 4     410-2%%730
2018-11-12T10:56:23.172Z 26    E1%%730
2018-11-12T10:56:33.292Z 4     410-2%%730
2018-11-12T10:56:33.292Z 30    E1%%730
2018-11-12T10:56:33.292Z 4     410-1%%730
2018-11-12T10:56:33.292Z 10    SIP%%730
[...]

PR #10555, opened 2018-12-11
fix fill(previous) mentioned in #10489

uoryon:

fix fill(previous) metioned in #10489 

we need reset `cursor.m` for every new series or group.


uoryon:
@stuartcarnie please review?

PR #10576, opened 2018-12-17
Update Go runtime to 1.10.6

e-dard:
This PR updates the runtime to `go 1.10.3`. At the time of this PR `go 1.10.4` was not available in the Docker repos.

It also adds `go 1.11.6` as the release to run as an extra test in CI.


e-dard:
This fixes #10576

PR #10580, opened 2018-12-18
Add user authentication and authorization support for Flux HTTP requests

stuartcarnie:
* Add `AuthorizeDatabase` API to `QueryAuthorizer` to verify a user has appropriate access to the specified database
* Update `serveFluxQuery` handler, adding a `meta.User` parameter, which triggers the Flux API requires a user when auth is enabled
* update Flux `createFromSource` and `createBucketsSource` dependencies to require `Authorizer` when auth is enabled in configuration
* update `createFromSource` verify read permissions for each bucket specified in a Flux query
* update `BucketsDecoder`, which implements the `buckets()` Flux function, to limit the displayed buckets to those the user has read or write permissions to. This replicates the behavior of the `SHOW DATABASES` InfluxQL command
* add unit tests to verify authentication is required for Flux HTTP requests when auth is enabled in configuration

### DIFF
--- a/cmd/influxd/run/server.go
+++ b/cmd/influxd/run/server.go
@@ -281,7 +281,8 @@ func (s *Server) appendHTTPDService(c httpd.Config) {
 	}
 	srv := httpd.NewService(c)
 	srv.Handler.MetaClient = s.MetaClient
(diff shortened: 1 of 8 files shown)


stuartcarnie:
@nathanielc agree – none of the improvements felt unnatural.

PR #10599, opened 2018-12-21
Add -check-utf8 to verify sub-command

stuartcarnie:
`-check-utf8` will check all series keys and verify they are valid UTF-8

NOTE: to complete the check as fast as possible, `-check-utf8` will only validate series keys for UTF-8 validity and not perform any other TSM validation.

### DIFF
--- a/cmd/influx_inspect/verify/tsm/verify.go
+++ b/cmd/influx_inspect/verify/tsm/verify.go
@@ -10,8 +10,10 @@ import (
 	"path/filepath"
 	"text/tabwriter"
 	"time"
+	"unicode/utf8"
 
 	"github.com/influxdata/influxdb/tsdb/engine/tsm1"
+	"github.com/pkg/errors"
 )
 
 // Command represents the program execution for "influx_inspect verify".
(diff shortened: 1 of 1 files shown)


cmd/influx_inspect/verify/tsm/verify.go
```diff
@@ -34,87 +36,189 @@ func (cmd *Command) Run(args ...string) error {
...
+	tw := tabwriter.NewWriter(cmd.Stdout, 16, 8, 0, '\t', 0)
+
+	var runner verifier
+	if checkUTF8 {
+		runner = &verifyUTF8{}
+	} else {
+		runner = &verifyChecksums{}
+	}
+	err := runner.Run(tw, dataPath)
+	tw.Flush()
```
benbjohnson: Not sure if it matters but the error from the `Flush()` is being discarded.

cmd/influx_inspect/verify/tsm/verify.go
```diff
@@ -34,87 +36,189 @@ func (cmd *Command) Run(args ...string) error {
...
+            Defaults to "%[1]s/.influxdb".
+    -check-utf8 
+            Verify series keys are valid UTF-8.
+            This check skips verification of block checksums.
+ `, os.Getenv("HOME"))
 
-	brokenBlocks := 0
-	totalBlocks := 0
+	fmt.Fprintf(cmd.Stdout, usage)
+}
```
benbjohnson: Can you simply call `fmt.Printf()` instead of `fmt.Sprintf()` -> `fmt.Fprintf()`?

PR #10664, opened 2019-01-09
Marked functions that always return floats as always returning floats

jsternberg:
These functions, when used on non-float types in subqueries, would
return incorrect results as the type system thought they would return a
different type than the actual implementation would return.


nathanielc:
What does this fix? Is it easy to test?

jsternberg:
It ensures that the type mapper returns the correct type which is used by subqueries.

It doesn't look like this is currently tested exhaustively. There are a few subquery tests, but I think it would be a lot to test all of the functions.

Issue #10837, opened 2019-01-10
cli: authorization permissions invalid ID


fntlnz:
### Observations

When trying to create a new authorization with `--delete-user` permissions I get an error:

```
influx  auth create -u yolo7  --delete-user
Error: invalid ID
Usage:
  influx auth create [flags]

Flags:
      --create-user                grants the permission to create users
      --delete-user                grants the permission to delete users
  -h, --help                       help for create
      --read-bucket stringArray    bucket id
  -u, --user string                user name (required)
      --write-bucket stringArray   bucket id

Global Flags:
      --host string    HTTP address of Influx (default "http://localhost:9999")
      --local          Run commands locally against the filesystem
  -t, --token string   API token to be used throughout client calls

invalid ID
```

### Analysis

I noticed two possible root causes for this problem:

Client side (cli) while creating a new `Permission` the function `NewPermission` doesn't set any ID on the newly created permission here:
https://github.com/influxdata/platform/blob/3429e8d0c6753e7f1dab6b080397580a5ba94205/authz.go#L166-L173

When creating the authorization, only permissions are passed to it, while the authorization needs a valid ID.

https://github.com/influxdata/platform/blob/3429e8d0c6753e7f1dab6b080397580a5ba94205/cmd/influx/authorization.go#L101-L103

Issue #2157, opened 2015-04-03
Can we have IN filter?


flowbehappy:
Hi! Great job for InfluxDB. 

And I wonder If it can support IN filter in where clause, so far it seems no supported:

```
SELECT * FROM mydb WHERE service IN ('aa', 'bb');

{
    "error": "error parsing query: found IN, expected SELECT, DELETE, SHOW, CREATE, DROP, GRANT, REVOKE, ALTER at line 1, char 321"
}
```

version: bdd469db26c483a0acdcadbb937829a3913d9be5

otoolep:
We don't have support for `IN` right now but you can use `OR`, for the query you provided:

`SELECT * from mydb WHERE service = 'aa' OR servce = 'bb'`

otoolep:
I too would like this, and it should be easy with some query re-writing of the AST. Let me see if I can get to it.

beckettsean:
A potentially easier workaround than multiple `OR` statements would be to use a regex. I think it should be faster, too. E.g.

`SELECT * from mydb WHERE service =~ /aa|bb/`

srikara:
+1

ch33hau:
Hi Guys,

Is this feature request still valid? If it is valid, can I work on this task?

Thanks

jsternberg:
Yes, we are still interested in this feature.

ch33hau:
Thanks, I will start working on this during weekend.

kalidasya:
We have another usecase, the support for `IN`  would make it possible to integrate Grafana's multiselect feature with influx
@ch33hau do you have any update?

ch33hau:
Hi @kalidasya sorry I was stop it half way, will working on it sooner. 

However, I found that `IN` already exists in influxdb for other purpose. Due to the way we code, might not able to use `IN` for the purpose of this issue. 
Is it possible to rename it to other syntax, eg, `WITHIN`?

kalidasya:
@ch33hau good to hear, `WITHIN` to me more time related but I also cannot really come up with anything better

stampy88:
@ch33hau any updates on this?

5k3105:
This would be great guys! It's been a few years but ???

taabodim:
Please implement this faster, this feature is very very mich needed

yshaul:
Is anybody working on this? I was thinking of taking a crack at this tomorrow

jsternberg:
[...]

PR #10947, opened 2019-01-10
Fix create authorization command 

desa:
Link https://github.com/influxdata/platform/pull/2185
Closes https://github.com/influxdata/influxdb/issues/10837

Briefly describe your proposed changes:
Previously, authorizations required the users id to be set explicitly on the authorization. As of #2157 the user id is retrieved off of the authorization or session used in the http request.

We now allow users to specify an explicit user when creating authorizations. If not id is provided, the user from the authorizer will be used.

Additionally, we now need to pass the organization id when creating an authorization.

### DIFF
--- a/cmd/influx/authorization.go
+++ b/cmd/influx/authorization.go
@@ -23,7 +23,8 @@ var authorizationCmd = &cobra.Command{
 
 // AuthorizationCreateFlags are command line args used when creating a authorization
 type AuthorizationCreateFlags struct {
-	org string
+	user string
+	org  string
 
 	createUserPermission bool
 	deleteUserPermission bool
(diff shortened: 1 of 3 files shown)


mark-rushakoff:
This has some merge conflicts now, and needs to be rebased to get CI sorted out. But when I built merged locally to master and built, and took a blind guess on the merge conflicts, it appeared to fix  the issue of being unable to create authorizations for other users.

cmd/influx/authorization.go
```diff
@@ -229,6 +228,23 @@ func authorizationCreateF(cmd *cobra.Command, args []string) error {
 		OrgID:       o.ID,
 	}
 
+	if authorizationCreateFlags.user != "" {
+		// if the user flag is supplied, the set the user ID explicitly on the request
```
leodido: Perhaps a typo here @desa?

`then set`?

Issue #10778, opened 2019-01-08
Add org to dashboards


alexpaxton:
Need to display the owner of a dashboard in the index view, just need Name & ID

PR #10948, opened 2019-01-10
Add organization to dashboards

desa:
Closes https://github.com/influxdata/influxdb/issues/10778

This PR adds `OrganizationID` to dashboards supports filtering by organization id via the API.

### DIFF
--- a/bolt/dashboard.go
+++ b/bolt/dashboard.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"sync"
 	"time"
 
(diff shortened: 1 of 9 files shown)


kelwang:
I really like the idea of having orgID index. But I am a little confused about the use of user resource mapping. Shall we build the additional org id based index in user resource mapping instead?

desa:
@kelwang I'm a bit confused how the user resource mapping is related here. My understanding was that the user resource mapping maps users to the resources that they own/created. This seems independent from which resources belong to which organizations.

Issue #10960, opened 2019-01-10
Filter missing resources from authorizations response


desa:
If a resource is removed, its authorization will still contain a like to that resource.

When we fetch the list of resources names in the lookup service, we should filter out resources that no longer exist.

PR #10971, opened 2019-01-10
fix(influxdb): add org id to permissions

desa:
Closes https://github.com/influxdata/influxdb/issues/10960 

This PR fixes various issues that were introduced by https://github.com/influxdata/platform/pull/2367.

Specifically, the following were done

* `OrgID` was added to permission
* Restructured the string version of a permission was `read:buckets:10` -> `read:orgs/1/buckets/10`
* Allowed does a hierarchical check of permissions `read:orgs/1/buckets` grants `read:org/1/buckets/10` 
* Ensures that authorization permissions `OrgID` match the authorizations `OrgID`.
* Addes `{write,read}-{users,buckets}` flags to CLI.
* Adds `OrgID` to the appropriate tests.


authz.go
```diff
@@ -40,15 +40,21 @@ func PermissionAllowed(p Permission, ps []Permission) bool {
...
-			permID = *perm.ID
-			if !permID.Valid() {
-				return false
+		if perm.ID == nil {
+			if perm.Action == p.Action && perm.Resource == p.Resource && perm.OrgID == p.OrgID {
+				return true
+			}
+		} else {
+			permID := ID(0)
+			if perm.ID != nil {
```
goller: Probably simplify as you have already done a nil check

bolt/onboarding.go
```diff
@@ -118,13 +118,12 @@ func (c *Client) Generate(ctx context.Context, req *platform.OnboardingRequest)
 		return nil, err
 	}
 
-	perms := platform.OperPermissions()
-	perms = append(perms, platform.OrgAdminPermissions(o.ID)...)
-	writeBucketPerm, err := platform.NewPermissionAtID(bucket.ID, platform.WriteAction, platform.BucketsResource)
+	perms := platform.OperPermissions(o.ID)
+	writeBucketPerm, err := platform.NewPermissionAtID(bucket.ID, platform.WriteAction, platform.BucketsResource, bucket.OrganizationID)
```
goller: Shouldn't the onboarding token be for all buckets rather than just a single bucket?

bolt/onboarding.go
```diff
@@ -118,13 +118,12 @@ func (c *Client) Generate(ctx context.Context, req *platform.OnboardingRequest)
...
 	if err != nil {
 		return nil, err
 	}
-	readBucketPerm, err := platform.NewPermissionAtID(bucket.ID, platform.ReadAction, platform.BucketsResource)
+	readBucketPerm, err := platform.NewPermissionAtID(bucket.ID, platform.ReadAction, platform.BucketsResource, bucket.OrganizationID)
```
goller: perhaps all buckets?

inmem/onboarding.go
```diff
@@ -93,13 +93,12 @@ func (s *Service) Generate(ctx context.Context, req *platform.OnboardingRequest)
 		return nil, err
 	}
 
-	perms := platform.OperPermissions()
-	perms = append(perms, platform.OrgAdminPermissions(o.ID)...)
-	writeBucketPerm, err := platform.NewPermissionAtID(bucket.ID, platform.WriteAction, platform.BucketsResource)
+	perms := platform.OperPermissions(o.ID)
+	writeBucketPerm, err := platform.NewPermissionAtID(bucket.ID, platform.WriteAction, platform.BucketsResource, bucket.OrganizationID)
```
goller: (same comment... all buckets?)

authz.go
```diff
@@ -40,15 +40,14 @@ func PermissionAllowed(p Permission, ps []Permission) bool {
...
-		if perm.ID != nil {
-			permID = *perm.ID
-			if !permID.Valid() {
-				return false
+		if perm.ID == nil {
+			if perm.Action == p.Action && perm.Resource == p.Resource && perm.OrgID == p.OrgID {
+				return true
+			}
+		} else {
+			if (*perm.ID).Valid() && *perm.ID == pID && perm.Action == p.Action && perm.Resource == p.Resource && perm.OrgID == p.OrgID {
```
jademcgough: `(*perm.ID).Valid() && *perm.ID == pID`

Won't this always be false? pID is ID(0), which by definition is not valid:
```
// Valid checks whether the receiving ID is a valid one or not.
func (i ID) Valid() bool {
	return i != 0
}
```

jademcgough: `(*perm.ID).Valid()`

You shouldn't need to check this and `*perm.ID == pID`; we already checked that `pID` is valid on line 37, so anything equal to that must also be valid.

desa: Good point 👍

bolt/onboarding.go
desa: yup. It is. Dunno why I left this here.

bolt/onboarding.go
desa: yup

desa:
@goller @jademcgough made suggested changes

auth.go
```diff
@@ -23,6 +24,20 @@ type Authorization struct {
 	Permissions []Permission `json:"permissions"`
 }
 
+// Valid ensures that the authorization is valid.
+func (a *Authorization) Valid() error {
+	for _, p := range a.Permissions {
+		if p.OrgID != a.OrgID {
```
goller: Zero value also works here.  Is there already a check for p.OrgID being valid?

authz.go
```diff
@@ -40,15 +40,14 @@ func PermissionAllowed(p Permission, ps []Permission) bool {
 	}
 
 	for _, perm := range ps {
-		permID := ID(0)
-		if perm.ID != nil {
-			permID = *perm.ID
-			if !permID.Valid() {
-				return false
+		if perm.ID == nil {
```
goller: What do you think of this to remove else:

```go
if perm.Action != p.Action || perm.Resource != p.Resource || perm.OrgID != p.OrgID {
    continue
}
```

auth.go
desa: There are other checks, but I'll add one here to ensure that the provided orgID is valid.

authz.go
desa: my intuition says keep the explicit if else in this case, but I'm open to switching it if you'd prefer to have it the other way.

PR #10973, opened 2019-01-10
fix: Add 'Basic Usage' fodder to README

jacobmarble:
Improve README to help folks get started using 2.0 quickly.


zeebo:
Given the visibility of a top level README, I don't know who/if someone else should look at this, so I'm not comfortable just fully approving.

That said, it looks fine to me from a technical perspective.

PR #10974, opened 2019-01-10
ci: only run Jenkinsfile in 2.0 development

mark-rushakoff:
We have separate Jenkins instances for 1.x development, that must not
attempt to run the 2.0 Jenkinsfile.


mark-rushakoff:
This approach didn't work.

PR #11001, opened 2019-01-11
Only update the changelog on 1.x branches

mark-rushakoff:
Master is switching to 2.x development today, and we are not going to
continue using the changelog strategy we used in 1.x.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,8 +9,7 @@ pipeline {
 
       when {
         anyOf {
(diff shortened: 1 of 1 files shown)


mark-rushakoff:
I spent a bit of time yesterday trying to figure out how to filter Jenkins builds to PRs only targeting specific branches, but I wasn't able to find a way to do it. ¯\\\_(ツ)_/¯

PR #11015, opened 2019-01-11
feat(task): Add validations for all task actions

### LINKED ISSUE #10767: Task permission validation is not being tested in task services.
Task validation is tested separate from the task service. It would be useful to test the validation failure in the context of an api call.

lyondhill:
Closes #10767


task/validator.go
```diff
@@ -35,6 +35,40 @@ func NewValidator(ts platform.TaskService, bs platform.BucketService) platform.T
...
+	perm, err := platform.NewPermission(platform.ReadAction, platform.TasksResource, task.Organization)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := validatePermission(ctx, *perm); err != nil {
+		return nil, err
+	}
+
+	return task, err
```
mark-rushakoff: `return task, nil` please.

task/validagtor_test.go
```diff
@@ -0,0 +1,482 @@
+package task_test
```
mark-rushakoff: Please fix the filename, to `task/validator_test.go`

PR #11029, opened 2019-01-11
Update onboarding to be able to make telegraf configs

ischolten:
Currently the api returns an error for get requests to the /telegrafs endpoint. Formerlly the ui would make that request to see if the config was already created to determine whether to update an existing config or create a new one. This check and update were removed until that fix is made to make onboarding work.

Additionally, we now store the organization id returned from setting up an admin since a telegraf post request now requires it.

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/onboarding/apis/index.ts
```diff
@@ -46,9 +51,10 @@ export interface SetupParams {
 
 export const setSetupParams = async (
   setupParams: SetupParams
-): Promise<void> => {
+): Promise<OnboardingResponse> => {
```
alexpaxton: Noice

PR #11032, opened 2019-01-12
refactor(flux): make packages mirror Flux namespaces

nathanielc:


query/options/task.go
```diff
@@ -1,23 +0,0 @@
...
-import (
-	"github.com/influxdata/flux"
-	"github.com/influxdata/flux/values"
-)
-
-func init() {
-	flux.RegisterBuiltInOption("task", taskObject())
-}
-
-func taskObject() values.Object {
```
jlapacik: Where did this go? Could you create an issue to add this back in? The easiest way is to probably just move this to the flux universe package. It will be a larger job to define pure flux packages here.

nathanielc: We chatted, and we don't need it at all. If we do need it again in the future we can add it.

PR #11034, opened 2019-01-12
Add save query as task functionality to Data Explorer

### LINKED ISSUE #10896: Add the ability to save the query from data explorer as a Task
Depends on #1868

as a data explorer
when i open the save as modal from the data explorer
then i would like the ability to save this query as a task
then i would like a way to set the name of the task
then i would like a way to set the org the task belongs to
then i would like the task to be saved as inactive
then i would like a notification telling me that the task was saved
then i would like a way to quickly jump to the task after creating it to continue configuring  and editing it

AlirieGray:
Closes #10896

This PR adds the `TaskForm` component to the Save As overlay in the Data Explorer to create a new task with the active query in the explorer.

Also adds form validation, so that a task can only be saved if it has a name, a schedule (interval/cron), and a script. 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/dataExplorer/components/SaveAsButton.tsx
+++ b/ui/src/dataExplorer/components/SaveAsButton.tsx
@@ -45,11 +45,11 @@ class SaveAsButton extends PureComponent<Props, State> {
           icon={IconFont.Export}
           text="Save As"
           onClick={this.handleShowOverlay}
-          color={ComponentColor.Success}
+          color={ComponentColor.Primary}
           titleText="Save your query as a Dashboard Cell or a Task"
         />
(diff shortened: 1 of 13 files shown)


ui/src/dataExplorer/components/SaveAsButton.tsx
```diff
@@ -76,7 +76,9 @@ class SaveAsButton extends PureComponent<Props, State> {
               {saveAsOption === SaveAsOption.Dashboard && (
```
mavarius: We could use a getter here since we never need these both at the same time and there will be more options soon.

PR #11042, opened 2019-01-14
Improve delete bucket performance

e-dard:
This PR significantly improves the performance of deleting a bucket in the storage engine.

During testing of the drop bucket feature I measured the following times:

| Series Cardinality  | Time to delete |
| ------------- | ------------- |
| 10,000  | 45s |
| 10,000,000  | ~∞  |

Where `~∞ > 8 hours` (the amount of time I left the delete running.

Clearly the times were unacceptable for any non-trivial workloads.

This PR improves the performance by doing the following (`s/measurement/bucket`):

  - Deleting an entire measurement from the index where possible rather than many series.
  - Rather than doing a key-->id lookup in the series file to get a series id, tracking the series ids belonging to all series for a measurement from the index, and then deleting those ids in the series file.
  - Reducing fsyncs in the `LogFile` by implementing a mass delete method.

After implementing these changes I see the following performance when deleting a bucket.

| Series Cardinality  | Time to delete |
| ------------- | ------------- |
| 10,000  | 8.7s |
| 10,000,000  | 18s  |

There are further improvements that can be made, but the above results show there is now sub-linear growth in the time to delete a bucket with respect to the cardinality.


tsdb/tsm1/engine_delete_measurement.go
```diff
@@ -143,6 +147,44 @@ func (e *Engine) DeletePrefix(prefix []byte, min, max int64) error {
...
+			}
+
+			var elem tsdb.SeriesIDElem
+			for elem, err = itr.Next(); err != nil; elem, err = itr.Next() {
+				if elem.SeriesID.IsZero() {
+					break
+				}
+
+				set.AddNoLock(elem.SeriesID)
+			}
```
zeebo: missing an error check unless Close is guaranteed to return the same one?

jacobmarble: Good eye!

e-dard:
@zeebo I knew I was going to forget that, the moment when I said to myself "remember to add that error check in, when you have tested this implementation"...

Fixed in 9ff65f60168848b45ee8a2176692e53e67987407

PR #11058, opened 2019-01-14
chore(ci): Fix nightly build bug

BondAnthony:
The project is currently using goreleaser to manage building nightly binaries, docker images and uploads to s3.
The tool has a strict dependency on semantic versioned tags. InfluxDB history does not have a valid semantic tag causing the tool to fail. This commit creates a valid tag for the current time to allow nightly builds to continue.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,7 +98,6 @@ jobs:
           name: Restore Yarn package cache
           keys:
             - chronograf-npm-packages-{{ checksum "ui/package-lock.json" }}
-
       - run: make build
 
   deploy:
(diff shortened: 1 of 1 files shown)


jacobmarble:
I would like to add that the tag that causes problems is currently `v1.6.0rc0`. To be strictly semver compliant, we would need to add a hyphen as `v1.6.0-rc0`. We wouldn't have to change all past tags, just future tags, but that might break some release workflows. `goreleaser` doesn't actually use the semver here, so it's easiest to just ignore the imperfect semver format with this change.

PR #11062, opened 2019-01-14
feat(http): add renew session

kelwang:
Closes https://github.com/influxdata/influxdb/issues/10773

renew the session with 5 min. when a new authorized request came in.
Manually tested with UI. 

  - [x] Rebased/mergeable
  - [x] Tests pass


bolt/session.go
```diff
@@ -22,6 +22,27 @@ func (c *Client) initializeSessions(ctx context.Context, tx *bolt.Tx) error {
 	return nil
 }
 
+// RenewSession extends the expire time by expiredAt.
+func (c *Client) RenewSession(ctx context.Context, session *platform.Session, expiredAt time.Time) error {
+	op := getOp(platform.OpRenewSession)
+	if session == nil {
```
jademcgough: when would we expect session to be nil?

bolt/session.go
```diff
@@ -22,6 +22,27 @@ func (c *Client) initializeSessions(ctx context.Context, tx *bolt.Tx) error {
 	return nil
 }
 
+// RenewSession extends the expire time by expiredAt.
+func (c *Client) RenewSession(ctx context.Context, session *platform.Session, expiredAt time.Time) error {
```
jademcgough: nitpick, but maybe `newExpiration` instead of `ExpiredAt`?

bolt/session.go
kelwang: Hi, Jade, thanks for the comment. I totally agree there should not be any issue for now. I was wondering if in the future, we might have some kind of map storing the session, this might be helpful to trigger the error.

PR #11066, opened 2019-01-14
Update Line Protocol button styles, submit form on file upload.

### LINKED ISSUE #11024: loading line protocol doesn't move to verify step
when loading a file in line protocol, the load succeeded but i was not moved to the next step.  attached is an example file to test with.

[all.out.zip](https://github.com/influxdata/influxdb/files/2751335/all.out.zip)

### LINKED ISSUE #11027: cancel on line protocol button needs styling
need to style the cancel button on file upload in onboarding

mavarius:
Closes #11027 
Closes #11024

Update Line Protocol button styles. Submit form on file upload to move to verify step.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/DragAndDrop.tsx
+++ b/ui/src/shared/components/DragAndDrop.tsx
@@ -1,5 +1,6 @@
 import React, {PureComponent} from 'react'
 import classnames from 'classnames'
+import {Button, ComponentColor, ComponentSize, ButtonType} from 'src/clockface'
 
 import './DragAndDrop.scss'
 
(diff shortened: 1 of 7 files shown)


ui/src/shared/components/DragAndDrop.tsx
```diff
@@ -146,11 +146,14 @@ class DragAndDrop extends PureComponent<Props, State> {
...
-        <button className="btn btn-sm btn-success" onClick={this.handleSubmit}>
+        <button
+          className="button button-sm button-success"
+          onClick={this.handleSubmit}
+        >
           {submitText}
         </button>
         <button
-          className="btn btn-sm btn-default"
+          className="button button-sm button-default"
```
alexpaxton: Want to import the clockface button and use it here? That way we don't have to worry about this component falling out of sync

PR #11093, opened 2019-01-15
Make task validation errors more clear

### LINKED ISSUE #10987: creating an empty task fails with a 500 error
on the UI, we should handle this gracefully.

in the api, it should return an error message for invalid task.

lyondhill:
fixes #10987

When task validation returns an error it should be of the bas model errors type so that the api can determine what status codes to return. This allows us to return the proper 400 status code when its an invalid request.

### DIFF
--- a/task/validator.go
+++ b/task/validator.go
@@ -254,11 +254,17 @@ func validateBucket(ctx context.Context, script string, preAuth query.PreAuthori
 
 	spec, err := flux.Compile(ctx, script, time.Now())
 	if err != nil {
-		return err
+		return platform.NewError(
+			platform.WithErrorErr(err),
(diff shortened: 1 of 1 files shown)


task/validator.go
```diff
@@ -254,11 +254,17 @@ func validateBucket(ctx context.Context, script string, preAuth query.PreAuthori
 
 	spec, err := flux.Compile(ctx, script, time.Now())
 	if err != nil {
-		return err
+		return platform.NewError(
+			platform.WithErrorErr(err),
+			platform.WithErrorMsg("Failed to parse flux script."),
```
mark-rushakoff: Maybe this should be `compile` rather than `parse`?

lyondhill: I like compile here!

PR #11112, opened 2019-01-15
build(ui): update Makefile to understand binary asset sources

nathanielc:
The Makefile for the binary assets didn't understand what its sources
were. As a result the binary assets were not rebuilt when the sources
changed.


BondAnthony:
Closing this PR since the suggested change was included in https://github.com/influxdata/influxdb/pull/11117

PR #11113, opened 2019-01-15
Only render SortBy dropdown if FieldOptions exist

### LINKED ISSUE #11077: A table option fails render
screen shot 2019-01-14 at 5 38 45 pm

AlirieGray:
Closes #11077 

Previously, the `SortBy` dropdown was crashing when field options could not be fetched from the server.

This PR ensures that the `SortBy` dropdown is only rendered if the field options array has at least one item.

  - [x] Rebased/mergeable
  - [x] Tests pass


AlirieGray:
alternatively, we could render some sort of an empty/error state in place of the dropdown if there are no field options.

Issue #10814, opened 2019-01-07
Start authorizing all http requests


desa:
Currently, all HTTP requests are authenticated, but not authorized.

This was done initially so that development would go more smoothly, however before the alpha, we must authorize requests appropriately.

Authorization should take place in the `authorizer` package, by creating a wrapped service that implements the associated service.

## Only requires Read/Write access to the resource (include org id for the time as it will be used when someone has an org level permission)
- [x] Buckets (https://github.com/influxdata/influxdb/pull/11135)
- [x] Dashboards (https://github.com/influxdata/influxdb/pull/11207)
- [x] Macros (https://github.com/influxdata/influxdb/pull/11203)  
- [x] Scraper (https://github.com/influxdata/influxdb/issues/11210)
- [x] Source (https://github.com/influxdata/influxdb/pull/11208)  
- [x] Telegrafs (https://github.com/influxdata/influxdb/issues/11159)

## Special (create requires write to the global organizations permission)
- [x] Organization (https://github.com/influxdata/influxdb/pull/11116)
- [x] Users (https://github.com/influxdata/influxdb/pull/11221)

## Each permission on an authorization must be allowed by the authorizer on context
- [x] Authorizations (https://github.com/influxdata/influxdb/pull/11261)

## Unsure of what will be involved for authorizing
- [x] User resource mapping (https://github.com/influxdata/influxdb/pull/11301)

# Will implement after alpha
- [x] Secret Service (https://github.com/influxdata/influxdb/issues/11345)
- [x] Label Service (https://github.com/influxdata/influxdb/issues/11362)

russorat:
@desa @goller looks like we can close this?

PR #11116, opened 2019-01-15
Authorize organization requests.

desa:
Closes part of #10814 

This PR introduces an `authorizer` package that is used to authorize service methods. It pulls the authorizer from context and uses it to authorize an action.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/authorizer/org.go
+++ b/authorizer/org.go
@@ -1,0 +1,143 @@
+package authorizer
+
+import (
+	"context"
+
+	"github.com/influxdata/influxdb"
+)
+
+var _ influxdb.OrganizationService = (*OrgService)(nil)
+
+// OrgService wraps a influxdb.OrganizationService and authorizes actions
+// against it appropriately.
+type OrgService struct {
+	s influxdb.OrganizationService
+}
+
+// NewOrgService constructs an instance of an authorizing org serivce.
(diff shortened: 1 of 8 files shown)


authorizer/org.go
```diff
@@ -0,0 +1,143 @@
...
+var _ influxdb.OrganizationService = (*OrgService)(nil)
+
+// OrgService wraps a influxdb.OrganizationService and authorizes actions
+// against it appropriately.
+type OrgService struct {
+	s influxdb.OrganizationService
+}
+
+// NewOrgService constructs an instance of an authorizing org serivce.
+func NewOrgService(s influxdb.OrganizationService) *OrgService {
```
jademcgough: Love this wrapper pattern, it's super intuitive

errors.go
```diff
@@ -17,6 +17,7 @@ const (
 	EEmptyValue       = "empty value"
 	EUnavailable      = "unavailable"
 	EForbidden        = "forbidden"
+	EUnauthorized     = "unauthorized"
```
kelwang: can you also add the status code mapping in http/errors.go as well?

bednar:
Hi @desa,

after this PR I cannot create an Organization with the session authentication scheme.

Can you look at the #11344?

PR #11119, opened 2019-01-15
Add the ability to download config from collectors

### LINKED ISSUE #10789: Add the ability to download config for a collector
as an org manager
then i would like to be able to download config for a collector

Palakp41:
Closes #10789 

Add the function to call telegraf config api to download config toml
Add notify to collectors if error occurs during download

kapture 2019-01-15 at 14 33 54


ui/src/organizations/components/CollectorRow.tsx
```diff
@@ -30,6 +31,7 @@ export default class BucketRow extends PureComponent<Props> {
               size={ComponentSize.Small}
               color={ComponentColor.Secondary}
               text={'Download Config'}
+              onClick={this.handleDownloadConfig}
```
alexpaxton: Can you combine this column and the last? So `Download` is in the same column as `Delete`?
You can wrap the buttons in a clockface `ComponentSpacer` to ensure there is a little gap between the two. See the action buttons in the dashboards list for an example of how to do this

Palakp41: Fixed it! Thanks :)

PR #11120, opened 2019-01-15
Ensure view retains cellID, and use cellID or id in updateVeiw

### LINKED ISSUE #11031: VEO doesn't save single stat graph config
i tried to build a cell and make it a single stat and each time i saved it, my dashboard still showed up with a line graph.

ebb-tide:
Closes #11031

### DIFF
--- a/ui/src/dashboards/actions/v2/views.ts
+++ b/ui/src/dashboards/actions/v2/views.ts
@@ -45,14 +45,16 @@ export const readView = (dashboardID: string, cellID: string) => async (
 export const updateView = (dashboardID: string, view: View) => async (
   dispatch: Dispatch<Action>
 ): Promise<View> => {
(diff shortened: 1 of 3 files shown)


ui/src/dashboards/actions/v2/views.ts
```diff
@@ -45,14 +45,15 @@ export const readView = (dashboardID: string, cellID: string) => async (
 export const updateView = (dashboardID: string, view: View) => async (
   dispatch: Dispatch<Action>
 ): Promise<View> => {
-  dispatch(setView(view.cellID, null, RemoteDataState.Loading))
+  const viewID = view.cellID || view.id // these are currently equivalent, views wil soon not have ids
```
bthesorceror: If views are not going to have ids why are we adding view.id here?

PR #11122, opened 2019-01-15
Feat/edit labels on tasks

### LINKED ISSUE #10893: add a way to edit labels on a task
as a user
when i see the list of labels on a task
then i would like the ability to edit those labels

AlirieGray:
Closes #10893 

This PR adds the ability to add and remove labels from a task.

Note that the `addTaskLabelsAsync` thunk will be changing when the the API is updated. In this PR, this thunk checks for which labels already exist on the task and only adds those which do not already exist. The API is going to be updated to handle not adding duplicates on the backend.

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/shared/components/EditLabelsOverlay.tsx
```diff
@@ -0,0 +1,172 @@
...
+        return ComponentStatus.Loading
+      }
+
+      return ComponentStatus.Default
+    }
+
+    return ComponentStatus.Disabled
+  }
+
+  private get changes(): {
```
Palakp41: Would it be better if there was a return type included in this getter?

PR #11124, opened 2019-01-15
Add the ability to delete collector from admin ui

### LINKED ISSUE #10788: Add the ability to remove collector from the org
as an org manager
then i would like to be able to remove a collector from the org

Palakp41:
Closes #10788 

Update swagger and cur_swagger to include delete for telegrafs/{telegrafID}
Add function to call delete api
Add onConfirm on collector to call delete function when clicked

kapture 2019-01-15 at 15 50 37


ui/src/organizations/components/CollectorRow.tsx
```diff
@@ -50,4 +52,7 @@ export default class BucketRow extends PureComponent<Props> {
   private handleDownloadConfig = (): void => {
     this.props.onDownloadConfig(this.props.collector.id)
   }
+  private handleDeleteConfig = (): void => {
+    this.props.onDelete(this.props.collector.id)
```
ebb-tide: so collectors are telegrafs? I'm really confused about these collectors

ui/src/organizations/components/CollectorRow.tsx
```diff
@@ -17,6 +17,7 @@ interface Props {
   collector: Telegraf
   bucket: string
   onDownloadConfig: (telegrafID: string) => void
+  onDelete: (telegrafID: string) => void
 }
 
 export default class BucketRow extends PureComponent<Props> {
```
ebb-tide: I feel like this class name got left behind from the copy and paste? :)

Palakp41: Oops

PR #11142, opened 2019-01-16
Use bucket api for listing buckets in time machine

### LINKED ISSUE #11068: Use buckets endpoint to populate buckets dropdown
instead of a GET request to `/v2/query` with flux script `buckets()`
we should get list of buckets by:
making a GET request to `/v2/buckets`

bthesorceror:
Closes #11068

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/actions/v2/queryBuilder.ts
+++ b/ui/src/shared/actions/v2/queryBuilder.ts
@@ -4,6 +4,8 @@ import {
   CancellationError,
 } from 'src/shared/apis/v2/queryBuilder'
 
+import {bucketsAPI} from 'src/utils/api'
+
 // Utils
 import {
   getActiveQuerySource,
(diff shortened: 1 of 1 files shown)


ui/src/shared/actions/v2/queryBuilder.ts
```diff
@@ -202,12 +204,11 @@ export const loadBuckets = () => async (
   dispatch: Dispatch<Action>,
   getState: GetState
 ) => {
-  const queryURL = getActiveQuerySource(getState()).links.query
-
   dispatch(setBuilderBucketsStatus(RemoteDataState.Loading))
 
   try {
-    const buckets = await fetcher.findBuckets(queryURL)
+    const {data} = await bucketsAPI.bucketsGet('')
```
ischolten: do you need to pass the empty string in?

bthesorceror: yeah, because we don't want any specific org, rather we want any buckets you have access to

PR #11145, opened 2019-01-16
Update final onboarding step to offer next step options to the user

### LINKED ISSUE #11047: Create dataLoaders start data collection step
as a user,
in the onboarding wizard,
after i do the admin setup,
i should see a page where i can select quick start, advanced, or skip data loading

image

mavarius:
Closes #11047 

Add options to the final step of onboarding to allow the user to decide when/how to add data.

image

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/onboarding/components/CompletionQuickStartButton.tsx
+++ b/ui/src/onboarding/components/CompletionQuickStartButton.tsx
@@ -1,0 +1,43 @@
+// Libraries
+import React, {PureComponent} from 'react'
+import {withRouter, WithRouterProps} from 'react-router'
+import _ from 'lodash'
+
+// Components
+import {Button, ComponentColor, ComponentSize} from 'src/clockface'
+import {ErrorHandling} from 'src/shared/decorators/errors'
+
+// Types
+import {Dashboard} from 'src/api'
+
+interface OwnProps {
+  dashboards: Dashboard[]
+}
+
+type Props = OwnProps & WithRouterProps
+
+@ErrorHandling
+class CompletionQuickStartButton extends PureComponent<Props> {
+  public render() {
+    return (
+      <Button
+        text="Quick Start"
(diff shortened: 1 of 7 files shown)


ui/src/onboarding/components/CompletionQuickStartButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+        color={ComponentColor.Success}
+        size={ComponentSize.Large}
+        onClick={this.handleAdvanced}
+      />
+    )
+  }
+
+  private handleAdvanced = (): void => {
+    const {router, dashboards} = this.props
+    const id = _.get(dashboards, 0, 'id')
```
ebb-tide: I think this get will give you the first dashboard itself or the string 'id' I think you want `_.get(dashboards, '[0].id', null)` so that the not(id) else clause is reachable

ui/src/onboarding/components/CompletionStep.tsx
```diff
@@ -19,27 +32,63 @@ class CompletionStep extends PureComponent<OnboardingStepProps> {
...
 
   public render() {
-    const {onExit, onDecrementCurrentStepIndex} = this.props
+    const {onExit} = this.props
 
     return (
       <div className="wizard--bookend-step">
+        <h3 className="wizard-step--title">You are ready to go!</h3>
+        <h5 className="wizard-step--sub-title">
+          Your InfluxDB 2.0 has 1 organization, 1 user, and 1 bucket.
```
ebb-tide: is this always going to be hardcoded? I think there's actually no way for them to have more than one of each of these right? If it needs to be dynamic later on, can we make an issue for it?

mavarius: They can only create 1 during onboarding so it's fine if it's hardcoded.

PR #11156, opened 2019-01-16
Add the ability to delete a task from organization

### LINKED ISSUE #11147: Add the ability to delete a task from org
as an org manager
when i view the list of tasks
then i would like to be able to delete a task from the org

Palakp41:
Closes #11147 

Remove the hardcoded delete value and add the button instead
Add functionality to delete a task from task row

kapture 2019-01-16 at 12 19 29

### DIFF
--- a/ui/src/organizations/components/TaskRow.tsx
+++ b/ui/src/organizations/components/TaskRow.tsx
@@ -1,0 +1,48 @@
+// Libraries
+import React, {PureComponent} from 'react'
+import {Link} from 'react-router'
+
+// Components
+import {
+  ComponentSize,
+  IndexList,
+  ConfirmationButton,
+  Alignment,
+} from 'src/clockface'
+
+// Api
+import {Task} from 'src/api'
+
+interface Props {
+  task: Task
+  onDelete: (taskID: string) => void
+}
+
+export default class TaskRow extends PureComponent<Props> {
+  public render() {
+    const {task} = this.props
+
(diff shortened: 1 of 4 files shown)


ui/src/organizations/components/TaskRow.tsx
```diff
@@ -0,0 +1,48 @@
+// Libraries
```
AlirieGray: there is already a `TaskRow` component in src/tasks/components, is there any way we could use that component here instead of making a new one?

ui/src/organizations/components/Tasks.tsx
```diff
@@ -10,10 +10,12 @@ import FilterList from 'src/shared/components/Filter'
 
 // Types
 import {Task} from 'src/api'
+import {deleteTask} from 'src/tasks/api/v2/index'
```
AlirieGray: same thing here, there's a `Tasks` component in src/tasks/components, perhaps we could make that component more generic so it can be used in both places?

ui/src/organizations/components/TaskRow.tsx
Palakp41: Going to create a new issue for this

ui/src/organizations/components/Tasks.tsx
Palakp41: Creating a new issue for this. Thanks!

Issue #11159, opened 2019-01-16
Authorize Telegrafs Service operations.


kelwang:
Part of https://github.com/influxdata/influxdb/issues/10814

Should follow pattern similar to https://github.com/influxdata/influxdb/pull/11135

PR #11179, opened 2019-01-16
feat(authorizer): add telegraf authorizer

kelwang:
Closes https://github.com/influxdata/influxdb/issues/11159

TODO:
- [x] Add tests
- [x] Hook up authorized http telegraf service

Closes Part of #10814 

This PR introduces a bucket service for authorizing requests.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/authorizer/telegraf.go
+++ b/authorizer/telegraf.go
@@ -1,0 +1,152 @@
+package authorizer
+
+import (
+	"context"
+
+	"github.com/influxdata/influxdb"
+)
+
+var _ influxdb.TelegrafConfigStore = (*TelegrafConfigService)(nil)
+
+// TelegrafConfigService wraps a influxdb.TelegrafConfigStore and authorizes actions
+// against it appropriately.
+type TelegrafConfigService struct {
+	s influxdb.TelegrafConfigStore
+	influxdb.UserResourceMappingService
+}
+
+// NewTelegrafConfigService constructs an instance of an authorizing telegraf serivce.
+func NewTelegrafConfigService(s influxdb.TelegrafConfigStore, urm influxdb.UserResourceMappingService) *TelegrafConfigService {
+	return &TelegrafConfigService{
+		s:                          s,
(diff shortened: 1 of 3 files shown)


http/api_handler.go
```diff
@@ -74,6 +74,7 @@ func NewAPIHandler(b *APIBackend) *APIHandler {
 	h := &APIHandler{}
 	b.BucketService = authorizer.NewBucketService(b.BucketService)
 	b.OrganizationService = authorizer.NewOrgService(b.OrganizationService)
+	b.TelegrafService = authorizer.NewTelegrafConfigStore(b.TelegrafService, b.UserResourceMappingService)
```
desa: I'd recommend against wrapping it at the top level like this. I did this originally and it caused a bunch of unintended problems. I'd only wrap the specific handle that needs it like

https://github.com/influxdata/influxdb/blob/f889675ef7934df81bd0743c102f680ddd720789/http/api_handler.go#L80

See https://github.com/influxdata/influxdb/pull/11176 for more details

authorizer/telegraf.go
```diff
@@ -0,0 +1,152 @@
...
+	if err := IsAllowed(ctx, *p); err != nil {
+		return err
+	}
+
+	return s.s.CreateTelegrafConfig(ctx, tc, userID)
+}
+
+// UpdateTelegrafConfig checks to see if the authorizer on context has write access to the telegraf config provided.
+func (s *TelegrafConfigService) UpdateTelegrafConfig(ctx context.Context, id influxdb.ID, upd *influxdb.TelegrafConfig, userID influxdb.ID) (*influxdb.TelegrafConfig, error) {
+	tc, err := s.FindTelegrafConfigByID(ctx, id)
```
desa: One quick note, this call should be to `s.s`. It likely won't cause an issue for people using the UI, but if people are using tokens for this it'll be problematic.

PR #11211, opened 2019-01-17
chore: update Flux to v0.16.0

nathanielc:


nathanielc:
We have a data race https://github.com/influxdata/flux/issues/879 we will fix and update to a point release.

PR #11223, opened 2019-01-17
Show organization names in task api actions

### LINKED ISSUE #10876: api: /tasks should return a organization's name that it belongs to in order to match the rest of the organization resources.

lyondhill:
Closes #10876 

Actions on resources under an organization show the org id as well as the name. This should bring the tasks resource inline with the rest of the resources in the same category. It also grants the advantage of allowing the api to accept org name instead of org id when creating a task.

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -5151,7 +5151,7 @@ components:
         orgID:
           description: The ID of the organization that owns this Task.
           type: string
-        organization:
+        org:
           description: The organization that owns this Task.
           type: string
         name:
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -5151,6 +5151,9 @@ components:
         orgID:
           description: The ID of the organization that owns this Task.
           type: string
+        organization:
+          description: The organization that owns this Task.
+          type: string
         name:
           description: A description of the task.
           type: string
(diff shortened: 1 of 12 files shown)


http/swagger.yml
```diff
@@ -5151,6 +5151,9 @@ components:
         orgID:
           description: The ID of the organization that owns this Task.
           type: string
+        organization:
```
mark-rushakoff: There are a couple cases of `organization` but it looks like `org` is more frequent.

task.go
```diff
@@ -22,7 +22,8 @@ const (
 // Task is a task. 🎊
 type Task struct {
 	ID              ID     `json:"id,omitempty"`
-	Organization    ID     `json:"orgID"`
+	OrganizationID  ID     `json:"orgID"`
+	Organization    string `json:"organization"`
```
mark-rushakoff: As above, I'm leaning towards `org` rather than `organization` for the JSON key.

lyondhill:
I used bucket as the template for the change. I can make it org instead of organization. Should I change it to OrgID as well?

mark-rushakoff:
I don't have a preference on the field name, just the JSON tag. `orgID` is already a JSON tag. I'm just asking for `` `json:"organization"` `` -> `` `json:"org"` ``.

PR #11233, opened 2019-01-17
feat(http): add vnd.flux content-type support

### LINKED ISSUE #10745: Add application/vnd.flux media type support for issuing plain-text Flux queries
Executing Flux queries via the HTTP endpoint using tools such as `curl` require the content be JSON encoded, which is tedious. Add support for a new media type, `application/vnd.flux` that accepts Flux queries as plain text.

* [ ] Update [Swagger documentation][swagger]
* [ ] Update [handler][handler]

[swagger]: https://github.com/influxdata/platform/blob/61d60a3c6113707067c3eb94dac2752e0db71823/http/swagger.yml#L1221
[handler]: https://github.com/influxdata/platform/blob/61d60a3c6113707067c3eb94dac2752e0db71823/http/query.go#L136

nathanielc:
Fixes #10745

This change means that 1.x and 2.x should be in sync, and the easy way to curl a query to the server is using the `application/vnd.flux

```
curl -XPOST -sS -H 'Authorization: Token foo:pass' -H accept:application/csv -H 'content-type:application/vnd.flux' http://localhost:9999/api/v2/query -d 'from(bucket:"test") |> range(start:-1000h) |> group(columns:["_measurement"], mode:"by") |> sum()'
```

### DIFF
--- a/http/query.go
+++ b/http/query.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"mime"
 	"net/http"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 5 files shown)


http/query.go
```diff
@@ -330,23 +332,27 @@ func QueryRequestFromProxyRequest(req *query.ProxyRequest) (*QueryRequest, error
...
+	}
+	switch mt {
+	case "application/vnd.flux":
+		body, err := ioutil.ReadAll(r.Body)
+		if err != nil {
+			return nil, err
 		}
-	} else {
+		req.Query = string(body)
+	default:
```
goller: minor nit... and only for "documentation" purposes, but, do you mind adding
a 

```case "application/json":
      fallthrough
```

to send it to default?

PR #11235, opened 2019-01-17
feat(ui/dataLoaders): Add new scraper from scrapers tab

### LINKED ISSUE #10785: Add the ability to open data loader wizard to scraping configurations
as an org manager
when i view the list of scrapers
then i would like to be able to add new scrapers to the bucket

ischolten:
Closes #10785

Clicking the Create scraper button on the scrapers tab in admin ui will open the the dataloaders wizard to the correct step and type. 
Additionally, fix a bug with creating scrapers.

  - [x] Rebased/mergeable
  - [x] Tests pass


ischolten:
can do

PR #11251, opened 2019-01-17
Create the scraper for the bucket selected in dropdown

### LINKED ISSUE #10784: Add the ability to select bucket from dropdown on scraper overlay
as an org manager
then i would like to be able to select a bucket when adding a new scraper to the org

Palakp41:
Closes #10784 

Updated redux state to the correct bucketID on dropdown select change

### DIFF
--- a/ui/src/onboarding/components/configureStep/Scraping.tsx
+++ b/ui/src/onboarding/components/configureStep/Scraping.tsx
@@ -14,6 +14,7 @@ import {
   setScraperTargetURL,
   saveScraperTarget,
 } from 'src/onboarding/actions/dataLoaders'
+import {setBucketInfo} from 'src/onboarding/actions/steps'
 
 // Types
 import {Bucket} from 'src/api'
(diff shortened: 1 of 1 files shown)


ui/src/onboarding/components/configureStep/Scraping.tsx
```diff
@@ -18,6 +18,7 @@ import {
 // Types
 import {Bucket} from 'src/api'
 import {AppState} from 'src/types/v2/index'
+import {setBucketInfo} from 'src/onboarding/actions/steps'
```
ischolten: nit: should this go under `// Actions`

ui/src/onboarding/components/configureStep/Scraping.tsx
```diff
@@ -101,7 +103,14 @@ export class Scraping extends PureComponent<Props> {
   }
 
   private handleSelectBucket = (bucket: string) => {
-    this.props.onSetScraperTargetBucket(bucket)
+    const {buckets, onSetScraperTargetBucket, onSetBucketInfo} = this.props
+
+    const bucketID = buckets.find(b => b.name === bucket).id
+    const org = buckets.find(b => b.name === bucket).organization
+    const orgID = buckets.find(b => b.name === bucket).organizationID
```
ischolten: so rather than finding the same bucket 3 times over again, can you in one step find the bucket and then extract the additional information that you need off of it?

PR #11279, opened 2019-01-18
test(influxdb): testing table generation and transformations

affo:
Tests for cursor and table generation.

This tests uses the `reads.Reader` by mocking `Store` and cursors to return test-specific data.

At the moment I am not checking table keys, but only data.  
I cannot get any data, because, when the transformation starts processing, it tries to add columns that already exist, because they are part of the group key.  
If I remove the group key from the input data, then I get a panic because a point must have at least 2 tags.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/query/querytest/reader.go
+++ b/query/querytest/reader.go
@@ -1,0 +1,83 @@
+package querytest
+
+import (
+	"context"
+	"math"
+	"sort"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/execute/executetest"
+	"github.com/influxdata/influxdb/query/stdlib/influxdata/influxdb"
+	"github.com/influxdata/influxdb/storage/reads"
+)
+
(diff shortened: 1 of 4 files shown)


affo:
@stuartcarnie I think this is way better

affo:
@stuartcarnie I pushed another version of the code.

I was trying to implement `GroupRead` for the `mockStore` and I discovered it could be a bit painful and that, in the end, I could end up mocking too much. In the end I decided to leave the tests as they were in their last update.

The original purpose of this test was to create a test case for when a groupBy gets pushed down to storage and causes the `MultiShardArrayCursor` error (https://github.com/influxdata/flux/issues/321), this tests cannot generate that error at the moment, because `GroupRead` is not implemented.

However, @aanthony1243 is working on end-to-end tests for storage and I think that that is the proper way to cause the aforementioned error in a test.

We can decide either to merge this code if it is a relevant contribution, or not if it can be directly covered by end-to-end tests.

affo:
As said above, this PR overlaps with end-to-end tests functionality. I close the PR, for now.

Issue #10529, opened 2018-11-29
segmentation violation in roaring.difference used by tsdb.DifferenceSeriesIDIterators


nimdanitro:
__System info:__ InfluxDB v1.7.1, Linux amd64

__Steps to reproduce:__ sporadic SIGSEGV after upgrade from 1.6.4 to 1.7.1
__Expected behavior:__  no panic and SIGSEGV

__Actual behavior:__  ```
unexpected fault address 0x7f4821482641
fatal error: fault
[signal SIGSEGV: segmentation violation code=0x1 addr=0x7f4821482641 pc=0x97ebb1]

goroutine 3941050647 [running]:
runtime.throw(0x11ed3e4, 0x5)
        /opt/OSAGgobuild/latest/src/runtime/panic.go:608 +0x72 fp=0xcbb48a8220 sp=0xcbb48a81f0 pc=0x42e072
runtime.sigpanic()
        /opt/OSAGgobuild/latest/src/runtime/signal_unix.go:397 +0x275 fp=0xcbb48a8270 sp=0xcbb48a8220 pc=0x443945
github.com/influxdata/influxdb/vendor/github.com/influxdata/roaring.difference(0xc03685ed40, 0x1d, 0x1d, 0x7f4821482641, 0x1, 0xf9c2d, 0xc03685f400, 0x0, 0x1d, 0x25)
        /usr/src/packages/src/pkg-osaginfluxdb_1orwWK9oi2cm9yPgpTMlhjYs3yl4Gvp8/src/github.com/influxdata/influxdb/vendor/github.com/influxdata/roaring/setutil.go:30 +0x71 fp=0xcbb48a8280 sp=0xcbb48a8270 pc=0x97ebb1
github.com/influxdata/influxdb/vendor/github.com/influxdata/roaring.(*arrayContainer).andNotArray(0xec0a5b0fa0, 0xca0c6cfca0, 0x13697e0, 0xec0a5b1aa0)
        /usr/src/packages/src/pkg-osaginfluxdb_1orwWK9oi2cm9yPgpTMlhjYs3yl4Gvp8/src/github.com/influxdata/influxdb/vendor/github.com/influxdata/roaring/arraycontainer.go:654 +0xe9 fp=0xcbb48a82f0 sp=0xcbb48a8280 pc=0x950e59
[...]

PR #11307, opened 2019-01-18
fix: Retain series file properly

jacobmarble:
Should help #10529

### DIFF
--- a/tsdb/index.go
+++ b/tsdb/index.go
@@ -51,13 +51,16 @@ func (itr *seriesIteratorAdapter) Next() (SeriesElem, error) {
 		}
 
 		// Skip if this key has been tombstoned.
+		release := itr.sfile.Retain()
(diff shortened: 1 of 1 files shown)


jacobmarble:
Adding a series file retain to an iterator deadlocks `TestIndex_SeriesIDSet` (package tsm1). `DeleteMeasurement/deleteMeasurement` retains the series file (read lock) when it creates an iterator representing series to delete. Later, `DeleteSeriesRange` consumes that iterator, first `.wait()`'ing the series file (write lock), then deleting from it with `SeriesFile.DeleteSeriesID()`.

jacobmarble:
Another example, `seriesIteratorAdaptor` does not retain/release it's series file, despite holding a reference.

https://github.com/influxdata/influxdb/blob/178579abde87155b983e7c33d8daae2173aabc50/tsdb/index.go#L30-L35

PR #11309, opened 2019-01-18
Add urm scraper

kelwang:
Closes https://github.com/influxdata/influxdb/issues/11237

add user resource mapping for scraper http endpoint.
user resource mapping become part of scraper creating transaction. 
when create the scraper, creator should become the owner.
when delete the scraper, related user resource mapping will be removed

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/authorizer/scraper.go
+++ b/authorizer/scraper.go
@@ -11,13 +11,15 @@ var _ influxdb.ScraperTargetStoreService = (*ScraperTargetStoreService)(nil)
 // ScraperTargetStoreService wraps a influxdb.ScraperTargetStoreService and authorizes actions
 // against it appropriately.
 type ScraperTargetStoreService struct {
+	influxdb.UserResourceMappingService
 	s influxdb.ScraperTargetStoreService
 }
 
 // NewScraperTargetStoreService constructs an instance of an authorizing scraper target store serivce.
-func NewScraperTargetStoreService(s influxdb.ScraperTargetStoreService) *ScraperTargetStoreService {
+func NewScraperTargetStoreService(s influxdb.ScraperTargetStoreService, urm influxdb.UserResourceMappingService) *ScraperTargetStoreService {
 	return &ScraperTargetStoreService{
-		s: s,
+		UserResourceMappingService: urm,
+		s:                          s,
 	}
 }
 
@@ -94,7 +96,7 @@ func (s *ScraperTargetStoreService) ListTargets(ctx context.Context) ([]influxdb
 }
 
 // AddTarget checks to see if the authorizer on context has write access to the global scraper target resource.
-func (s *ScraperTargetStoreService) AddTarget(ctx context.Context, st *influxdb.ScraperTarget) error {
+func (s *ScraperTargetStoreService) AddTarget(ctx context.Context, st *influxdb.ScraperTarget, userID influxdb.ID) error {
 	p, err := influxdb.NewPermission(influxdb.WriteAction, influxdb.ScraperResourceType, st.OrgID)
 	if err != nil {
 		return err
@@ -104,11 +106,11 @@ func (s *ScraperTargetStoreService) AddTarget(ctx context.Context, st *influxdb.
 		return err
 	}
 
-	return s.s.AddTarget(ctx, st)
+	return s.s.AddTarget(ctx, st, userID)
 }
 
 // UpdateTarget checks to see if the authorizer on context has write access to the scraper target provided.
-func (s *ScraperTargetStoreService) UpdateTarget(ctx context.Context, upd *influxdb.ScraperTarget) (*influxdb.ScraperTarget, error) {
+func (s *ScraperTargetStoreService) UpdateTarget(ctx context.Context, upd *influxdb.ScraperTarget, userID influxdb.ID) (*influxdb.ScraperTarget, error) {
 	st, err := s.s.GetTargetByID(ctx, upd.ID)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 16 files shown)


authorizer/scraper.go
```diff
@@ -11,13 +11,15 @@ var _ influxdb.ScraperTargetStoreService = (*ScraperTargetStoreService)(nil)
 // ScraperTargetStoreService wraps a influxdb.ScraperTargetStoreService and authorizes actions
 // against it appropriately.
 type ScraperTargetStoreService struct {
+	influxdb.UserResourceMappingService
```
desa: I'm a bit uneasy about this being embedded like this. What's the reasoning for doing this?

http/scraper_service.go
```diff
@@ -8,47 +8,75 @@ import (
...
-	h.HandlerFunc("GET", targetPath+"/:id", h.handleGetScraperTarget)
-	h.HandlerFunc("PATCH", targetPath+"/:id", h.handlePatchScraperTarget)
-	h.HandlerFunc("DELETE", targetPath+"/:id", h.handleDeleteScraperTarget)
+	h.HandlerFunc("POST", targetsPath, h.handlePostScraperTarget)
+	h.HandlerFunc("GET", targetsPath, h.handleGetScraperTargets)
+	h.HandlerFunc("GET", targetsPath+"/:id", h.handleGetScraperTarget)
+	h.HandlerFunc("PATCH", targetsPath+"/:id", h.handlePatchScraperTarget)
+	h.HandlerFunc("DELETE", targetsPath+"/:id", h.handleDeleteScraperTarget)
+
+	h.HandlerFunc("POST", targetsIDMembersPath, newPostMemberHandler(h.UserResourceMappingService, h.UserService, influxdb.ScraperResourceType, influxdb.Member))
```
desa: Does this work? `h.UserResourceMappingService` will be `nil` here.

http/scraper_service.go
```diff
@@ -8,47 +8,75 @@ import (
...
+	h.HandlerFunc("GET", targetsIDMembersPath, newGetMembersHandler(h.UserResourceMappingService, h.UserService, influxdb.ScraperResourceType, influxdb.Member))
+	h.HandlerFunc("DELETE", targetsIDMembersIDPath, newDeleteMemberHandler(h.UserResourceMappingService, influxdb.Member))
+
+	h.HandlerFunc("POST", targetsIDOwnersPath, newPostMemberHandler(h.UserResourceMappingService, h.UserService, influxdb.ScraperResourceType, influxdb.Owner))
+	h.HandlerFunc("GET", targetsIDOwnersPath, newGetMembersHandler(h.UserResourceMappingService, h.UserService, influxdb.ScraperResourceType, influxdb.Owner))
+	h.HandlerFunc("DELETE", targetsIDOwnersIDPath, newDeleteMemberHandler(h.UserResourceMappingService, influxdb.Owner))
+
+	h.HandlerFunc("GET", targetsIDLabelsPath, newGetLabelsHandler(h.LabelService))
```
desa: Same thing here `h.LabelService` will be `nil`

http/swagger.yml
```diff
@@ -709,7 +1029,8 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
-  '/macros/{macroID}':
+
```
desa: What is this for?

mock/middle_ware.go
```diff
@@ -0,0 +1,27 @@
+package mock
```
desa: filename should be one word `middleware.go`

desa: Actually, what do you think about making this a subpackage of `http`? So this file would be in `http/mock/middleware.go`?

mock/middle_ware.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"net/http"
+
+	"github.com/influxdata/influxdb"
+	platcontext "github.com/influxdata/influxdb/context"
+)
+
+// NewMiddlewareHandler create a mocked middleware handler.
+func NewMiddlewareHandler(handler http.Handler, auth influxdb.Authorizer) http.Handler {
```
desa: Seems like this is a very generic name, since it specifically is am authentication middleware mock.

scraper.go
```diff
@@ -28,11 +28,12 @@ type ScraperTarget struct {
 
 // ScraperTargetStoreService defines the crud service for ScraperTarget.
 type ScraperTargetStoreService interface {
+	UserResourceMappingService
```
desa: I don't understand why this is embedded. What's the reasoning for doing this?

scraper.go
```diff
@@ -28,11 +28,12 @@ type ScraperTarget struct {
...
 	ListTargets(ctx context.Context) ([]ScraperTarget, error)
-	AddTarget(ctx context.Context, t *ScraperTarget) error
+	AddTarget(ctx context.Context, t *ScraperTarget, userID ID) error
```
desa: Why does this function take a userID?

mock/middle_ware.go
kelwang: sure

http/scraper_service.go
kelwang: good catch!

scraper.go
kelwang: this is for creating the user resource mapping.

authorizer/scraper.go
kelwang: I would like to have the urm as part of the same transaction when creating the scraper. the user should become the owner of the scraper once created. also when delete the scraper, all the related urm should be removed as well

kelwang: I did the same thing of telegraf as well.

desa: Why embed it though?

scraper.go
desa: I see. I think for now this is fine, but we should probably revisit this pattern to see if this is something we want to repeat.

desa: Specifically, I'm concerned about us mixing abstractions and tightly coupling pieces together.

kelwang: 👍 yeah, totally agree. I think it's better maybe we can think about adapting a search engine thing for the indexing.

PR #11317, opened 2019-01-18
Update url input to now allow empty strings

### LINKED ISSUE #11271: Dont allow empty string to be saved for scraper
as a user using the dataloader wizard,
if i select scraper,
i must enter something inthe scraper url

Palakp41:
Closes #11271 

Update scraping to disable continue button if url is empty
Update scraper target to display error if url input is empty

  - [ ] Rebased/mergeable
  - [ ] Tests pass


ui/src/onboarding/components/configureStep/ScraperTarget.tsx
```diff
@@ -65,6 +76,13 @@ export class ScraperTarget extends PureComponent<Props> {
     ))
   }
 
+  private get urlEmpty(): boolean {
+    if (this.props.url === '') {
```
AlirieGray: I think you might just be able to do something like `return !this.props.url`

PR #11364, opened 2019-01-22
Move the WAL from the tsm1.Engine to the storage.Engine

### LINKED ISSUE #11094: Move the WAL up from tsm1 to storage engine
# Tear down that WAL

This document is a proposal with cost/benefits for moving the WAL out of the `tsm1.Engine` and
in to the `storage.Engine`. The high level goals are:

- Keeping consistency with all the on disk structures
- A closer architecture to cloud 2.0
- Reducing moving parts for more involved TSM work

## Consistency

Right now, the flow of data is:

```
Writes:
┌──────────────────────────────────────────────────────────────┐
│Storage Engine                                                │
│                        ┌────────────────────────────────────┐│
│                        │TSM1 Engine                         ││
│┌───────────────────┐   │┌───┐   ┌─────┐   ┌────────────────┐││
││Index + Series File├──>││WAL├──>│Cache├──>│Compaction (TSM)│││
│└───────────────────┘   │└───┘   └─────┘   └────────────────┘││
│                        └────────────────────────────────────┘│
└──────────────────────────────────────────────────────────────┘

Deletes:
┌───────────────────────────────────────────────┐
│TSM1 Engine                                    │
│┌───┐   ┌───┐   ┌─────┐   ┌───────────────────┐│
││TSM├──>│WAL├──>│Cache├──>│Index + Series File││
│└───┘   └───┘   └─────┘   └───────────────────┘│
└───────────────────────────────────────────────┘
```

Writes have a problem if there is a crash or hard exit between the index + series file and the WAL.
Deletes have a problem if there are deletes during the TSM delete (partial tombstones written), or
after the WAL, or during the index + series file (partial series deletion). The partial deletes are
a problem because delete operations do not know the full set of keys that could be deleted since
they operate on buckets. This means that we may not know to delete a series from the index or
series file because we won't observe the series being deleted.

The proposal is to change the flow to look like this:

```
Writes:
┌──────────────────────────────────────────────────────────────┐
│Storage Engine                                                │
│                                ┌────────────────────────────┐│
│                                │TSM1 Engine                 ││
│┌───┐   ┌───────────────────┐   │┌─────┐   ┌────────────────┐││
││WAL├──>│Index + Series File├──>││Cache├──>│Compaction (TSM)│││
│└───┘   └───────────────────┘   │└─────┘   └────────────────┘││
│                                └────────────────────────────┘│
└──────────────────────────────────────────────────────────────┘

Deletes:
┌─────────────────────────────────────────────────┐
│Storage Engine                                   │
│        ┌───────────────────────────────────────┐│
│        │TSM1 Engine                            ││
│┌───┐   │┌─────┐   ┌───────────────────┐   ┌───┐││
││WAL├──>││Cache├──>│Index + Series File├──>│TSM│││
│└───┘   │└─────┘   └───────────────────┘   └───┘││
│        └───────────────────────────────────────┘│
└─────────────────────────────────────────────────┘

```

This solves the above issues because the operations are idempotent. Moving the TSM deletes last
in the pipeline for deletes makes it so that partial failures are still ok.

## Closer architecture

This is closer to cloud 2.0 which means that testing, performance and bug fixes for that will
be more applicable to our open source code, and visa versa. It's synergistic!

## Reducing moving parts

By moving the WAL up out of the tsm engine, we are allowed to assume inside of it that any
partial operations will be replayed automatically. This greatly simplifies the concerns at
those layers, making it so that we just have to worry about keeping the state in a working
state. Additionally, it can help performance because we can batch up operations in memory or
delay fsyncs until more work has been done confident that any uncomitted work will be replayed
by the WAL.

# Challenges

The challenges for this proposal include

- The WAL is tied into the tsm1 package and engine fairly deeply
- Knowing when it's safe to truncate the WAL requires upward signaling
- It only replays into a cache

## Deep ties

TODO: Investigate all the spots the WAL is used and under what locks

## Knowing when to truncate

The WAL right now truncates writes when compactions happen. This is problematic because we also
need to keep track of when deletes have been fully applied for the truncate to be possible.
Additionally, knowing when compactions happen and exactly what values have been compacted will
require signaling from the `tsm1.Engine` upward to the `storage.Engine`, with possibly some
risk with locks and stuff.

## Replay

The cache kinda directly loads from the internal WAL data structures, which causes some problems
for moving it. We need the WAL to replay not only to the cache, but to the index as well. It
will become more like an op log and less like some special purpose data structure to load values
into a cache. That may be challenging to keep performance up.

zeebo:
Closes #11094

This moves the WAL from being in the tsm1.Engine to being in the storage.Engine. This required a number of changes including:

- Some weird upward signaling and lock management
- Changing retention to use DeleteBucketRange
- Removing some delete methods from the engine
- Removing some delete type from the WAL
- Adding a new delete type to the WAL
- Replay the WAL rather than directly load into the Cache
- Move the tsm1 Value type into its own package


zeebo:
### Added TODOs

- https://github.com/influxdata/influxdb/pull/11364/files#diff-18b0fccd77bafb3a93fcacdfdd86ec34R228
	This existed before this PR. This TODO just calls out the poor behavior.

- https://github.com/influxdata/influxdb/pull/11364/files#diff-b44a6551f36995f10c237a575eec2bf7R558
	This TODO is just moving the lock usage one into the function body where it happens and adding a comment about keeping track of the keys. I should probably remove the key comment because it's actually just a slice header per key being removed, since the cache already keeps all of the keys in memory.

- https://github.com/influxdata/influxdb/pull/11364/files#diff-b633e12e59f6e30d9f2f96ccc8a301e2R108
	I'd like some help in knowing which metrics to provide here. How does adding timing + status of every delete for every bucket/org combo sound? Worried about cardinality? Just a histogram of timing + counts of errors/ok?

- https://github.com/influxdata/influxdb/pull/11364/files#diff-18b0fccd77bafb3a93fcacdfdd86ec34R420
	I'm waiting for ben's change for improving performance of explode so that I can make this work well with that.

### Action items

- Update comment in the cache.
- Add metrics when which ones are agreed upon.

storage/retention.go
```diff
@@ -98,149 +91,41 @@ func (s *retentionEnforcer) run() {
...
-			mu.Unlock()
-			atomic.AddUint64(&seriesSkipped, 1)
-			return 0, 0, false
-		}
-		if retentionPeriod == 0 {
-			return 0, 0, false
+		max := now.Add(-b.RetentionPeriod).UnixNano()
+		err := s.Engine.DeleteBucketRange(b.OrganizationID, b.ID, math.MinInt64, max)
+		if err != nil {
+			// TODO(jeff): metrics?
```
e-dard: I don't see where `err` can be returned here? Maybe change the API so it doesn't return an `error`? 

Then you could move the failed metrics error down from here https://github.com/influxdata/influxdb/pull/11364/files#diff-b633e12e59f6e30d9f2f96ccc8a301e2R84 into this branch?

We should track failed deletions segmented by org, bucket, which is what https://github.com/influxdata/influxdb/pull/11364/files#diff-b633e12e59f6e30d9f2f96ccc8a301e2R89 tracks.

zeebo:
### Performance

    inch -bucket b -org o -p 10 -m 10 -t 100,100,10 -v -c 8

#### Master
```
10000000 points written (100%).
Total throughput: 689477.0 pt/sec | 689477.0 val/sec.
Current throughput: 350000 val/sec.
Errors: 0
μ: 114.874573ms
90%: 150.491507ms
95%: 173.744145ms
99%: 233.363966ms
```

#### This change

```
10000000 points written (100%).
Total throughput: 693866.6 pt/sec | 693866.6 val/sec.
Current throughput: 330000 val/sec.
Errors: 0
μ: 114.452816ms
90%: 156.055782ms
95%: 176.183093ms
99%: 220.710342ms
```

### Conclusion

There is no performance change. We're doing the same amount of work, just in a different order.

Issue #11254, opened 2019-01-17
api: JSON key conflict in dashboard response


mark-rushakoff:
The http.dashboardResponse type has a `json:"cells"` tag and it embeds an influxdb.Dashboard.

https://github.com/influxdata/influxdb/blob/5baa4ba397f6bc7b0dec0dafc9f7b82893c83cfe/http/dashboard_service.go#L99-L104

But influxdb.Dashboard also has a `json:"cells"` tag:

https://github.com/influxdata/influxdb/blob/60009cb1b78eb82bddc6b71e1022229a8fd8eb81/dashboard.go#L70-L77

I don't know which `cells` wins when a dashboardResponse is serialized, or if it's even defined behavior.

Found via go vet on the latest Go 1.12:

```
# github.com/influxdata/influxdb/http
http/dashboard_service.go:101:2: struct field Cells repeats json tag "cells" also at ../dashboard.go:75
```

mark-rushakoff:
FWIW we could use `go test -vet=off` in CI to delay fixing this and unblock upgrading to go 1.12. But we really should fix this.

PR #11390, opened 2019-01-22
ci: use go1.12 on CircleCI

mark-rushakoff:


mark-rushakoff:
Currently blocked on #11254, which is a new vet check in 1.12.

PR #11418, opened 2019-01-22
feat: Onboarding without interaction

### LINKED ISSUE #10684: cli: support non-interactive onboarding
During development, I frequently want to start fresh without any data, and I always use the same values during onboarding. It would be more convenient if I could specify all the arguments as command line flags or pass in a config file, rather than having to interactively type in each argument every time I run `influx setup`.

jacobmarble:
Closes #10684 

Makes this possible:
```
$ influx setup --username user --password s3^R#tpA$s --org my-org --bucket my-bucket --retention 168 --token my-token --force
Welcome to InfluxDB 2.0!
UserID                  Username        Organization    Bucket
0349dcdc02ab4000        user            my-org          my-bucket
Your token has been stored in /Users/jacobmarble/.influxdbv2/credentials.
```


bolt/authorization.go
```diff
@@ -226,17 +226,20 @@ func (c *Client) CreateAuthorization(ctx context.Context, a *platform.Authorizat
...
+			a.Token = token
 		}
-		a.Token = token
 
 		a.ID = c.IDGenerator.ID()
 
+		var err error
 		pe := c.putAuthorization(ctx, tx, a)
 		if pe != nil {
 			pe.Op = op
```
mark-rushakoff: I'd probably do `return pe` in this block and `return nil` outside it so you don't need to declare `var err error` -- slightly easier to read without the extra declaration IMO.

cmd/influx/setup.go
```diff
@@ -3,15 +3,14 @@ package main
...
-	"strconv"
-	"strings"
-
 	platform "github.com/influxdata/influxdb"
 	"github.com/influxdata/influxdb/cmd/influx/internal"
 	"github.com/influxdata/influxdb/http"
 	"github.com/spf13/cobra"
-	input "github.com/tcnksm/go-input"
+	"github.com/tcnksm/go-input"
+	"os"
```
mark-rushakoff: Fix import order.

cmd/influx/setup.go
```diff
@@ -76,21 +97,47 @@ func getOnboardingRequest() (req *platform.OnboardingRequest, err error) {
...
+	if setupFlags.bucket != "" {
+		req.Bucket = setupFlags.bucket
+	} else {
+		req.Bucket = getInput(ui, "Please type your primary bucket name", "")
+	}
+	if setupFlags.retention >= 0 {
+		req.RetentionPeriod = uint(setupFlags.retention)
+	} else {
+		for {
+			rpStr := getInput(ui, "Please type your retention period in hours (exp 168 for 1 week).\r\nOr press ENTER for infinite.", strconv.Itoa(platform.InfiniteRetention))
```
mark-rushakoff: I know you didn't introduce this line, but `s/exp/ex/`? Or maybe something else that is more clear, or maybe just drop the parenthetical altogether.

PR #11438, opened 2019-01-22
Update error message to api response message on tasks

### LINKED ISSUE #11165: Tasks UI should show the errors being reported by the api
when building a task, the UI isn't showing the error message returned from the API. it is showing a generic one with the http status code

Palakp41:
Closes #11165

Update notifications in Tasks UI to api response error messages

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/tasks/actions/v2/index.ts
+++ b/ui/src/tasks/actions/v2/index.ts
@@ -231,7 +231,8 @@ export const updateTaskStatus = (task: Task) => async dispatch => {
     dispatch(populateTasks())
   } catch (e) {
     console.error(e)
-    dispatch(notify(taskDeleteFailed()))
+    const message = getErrorMessage(e)
+    dispatch(notify(taskDeleteFailed(message)))
   }
 }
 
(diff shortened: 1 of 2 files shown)


ui/src/tasks/actions/v2/index.ts
```diff
@@ -357,7 +362,8 @@ export const saveNewScript = (
     dispatch(populateTasks())
   } catch (e) {
     console.error(e)
-    dispatch(notify(taskNotCreated(e.message)))
+    const message = _.get(e, 'response.data.error.message', '')
```
ischolten: could you use getErrorMessage here?

ui/src/tasks/actions/v2/index.ts
```diff
@@ -390,7 +396,7 @@ export const importScript = (script: string, fileName: string) => async (
     dispatch(notify(taskImportSuccess(fileName)))
   } catch (error) {
     console.error(error)
-    const message = getDeep<string>(error, 'message', '')
+    const message = _.get(error, 'response.data.error.message', '')
```
ischolten: same here

ui/src/tasks/actions/v2/index.ts
Palakp41: Actually, don't need it here because the api comes back with response.data.error.message.

ui/src/tasks/actions/v2/index.ts
Palakp41: Same as above. It calls the same post api for tasks

PR #11454, opened 2019-01-23
Add destination to task flux script

### LINKED ISSUE #11168: Save As task should allow the user to specify a To with a bucket and org
the save as for a task should allow the user to select a bucket and an org for the task. Doing so should add a `|> to(bucket: "bucket_name", org: "org_name")` to the end of the task query.

ebb-tide:
Closes #11168

### DIFF
--- a/ui/src/tasks/components/TasksOptionsBucketDropdown.tsx
+++ b/ui/src/tasks/components/TasksOptionsBucketDropdown.tsx
@@ -1,0 +1,91 @@
+// Libraries
+import React, {PureComponent} from 'react'
+import _ from 'lodash'
+
+// Components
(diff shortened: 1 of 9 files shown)


ui/src/tasks/components/TasksOptionsBucketDropdown.tsx
```diff
@@ -0,0 +1,95 @@
...
+      return buckets.map(bucket => {
+        return (
+          <Dropdown.Item id={bucket.name} key={bucket.name} value={bucket.name}>
+            {bucket.name}
+          </Dropdown.Item>
+        )
+      })
+    } else {
+      return [
+        React.createElement(
```
OfTheDelmer: Why no JSX for `Dropdown.Item`?

PR #11458, opened 2019-01-23
When all tags have selector, don't show add button

### LINKED ISSUE #11406: Disable the + button on the builder when there are no more tags to display.
I shouldn't be able to do this

image

bthesorceror:
Closes #11406

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/TimeMachineQueryBuilder.tsx
+++ b/ui/src/shared/components/TimeMachineQueryBuilder.tsx
@@ -14,16 +14,21 @@ import FunctionSelector from 'src/shared/components/FunctionSelector'
 import {loadBuckets, addTagSelector} from 'src/shared/actions/v2/queryBuilder'
 
 // Utils
-import {getActiveQuery} from 'src/shared/selectors/timeMachines'
+import {
+  getActiveQuery,
+  getActiveTimeMachine,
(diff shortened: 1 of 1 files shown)


ui/src/shared/components/TimeMachineQueryBuilder.tsx
```diff
@@ -56,25 +61,40 @@ class TimeMachineQueryBuilder extends PureComponent<Props, State> {
...
+        customClass="query-builder--add-tag-selector"
+      />
+    )
+  }
 }
 
 const mstp = (state: AppState): StateProps => {
   const tagFiltersLength = getActiveQuery(state).builderConfig.tags.length
+  const moreTags = !getActiveTimeMachine(state).queryBuilder.tags.find(
+    t => !t.keys.length && t.keysStatus === RemoteDataState.Done
```
chnn: I think this will hide the add button when a user searches for tag keys in any `TagSelector` and the search returns no results.

I think that's fine. Maybe a small improvement would be to only perform this check on the last `queryBuilder.tags` object (which corresponds to the right-most `TagSelector`).

bthesorceror: fair enough

PR #11480, opened 2019-01-23
feat(cmd/influx): pass context by default

fntlnz:
Signed-off-by: Lorenzo Fontana <lo@linux.com>

Always pass main context in cmd instead of creating it multiple times


fntlnz:
This is rebased and g2g @desa @goller !

PR #11493, opened 2019-01-23
Fix typos

asashour:
In 1.7 branch.

PR #11494, opened 2019-01-23
Fix typos

asashour:
In 1.8 branch


asashour:
@jacobmarble: this is similar to https://github.com/influxdata/influxdb/pull/11493, but against `1.8`

Issue #11427, opened 2019-01-22
Integrate /owners routing with tasks


lyondhill:
Proper integration will be a somewhat involved change:

- [x] On task create, assign owning user and org via UserResourceMappingService
- [x] On task delete, clean up UserResourceMappingService
- [x] FindTasks needs to look up task IDs through the UserResourceMappingService
- [x] Task store needs to stop maintaining its own user and owner index
- [ ] Remove task.Owner from the task struct.

lyondhill:
The next steps of this process needs to be put on hold.

In order to update the task find function to use the UserResourceMap to pull a list of tasks by a user id. The current implementation of URM is using a different backend storage than tasks. We need to unify tasks with the other storage systems before we can reasonably make a clean change here

mark-rushakoff:
As discussed in Slack:

To unblock FindTasks integration with users/members, the HTTP task handler should make one call to the URM to discover the valid task IDs, and then make many individual calls to `task/backend.(store.FindTaskByID)`. It certainly won't be performant, but it will be correct. Moving forward on this work will leave us better informed on what we need from the eventual refactor simplifying the task backend.

PR #11504, opened 2019-01-23
Remove owner from task json responses

### LINKED ISSUE #11164: api: tasks returns owner in response but should not.
Ownership of resources is to be handled at the /api/v2/{RESOURCE}/{id}/owners endpoint.

For now, tasks needs to remove the field `owner` from the task API.

This is a response that contains owner, but, should not: 

screen shot 2019-01-16 at 12 36 20 pm

/cc @Palakp41 @ebb-tide

lyondhill:
Closes #11164

### DIFF
--- a/task/servicetest/servicetest.go
+++ b/task/servicetest/servicetest.go
@@ -55,7 +55,7 @@ func TestTaskService(t *testing.T, fn BackendComponentFactory) {
 		// but more importantly, it should exercise concurrency to catch data races.
 
 		t.Run("Task CRUD", func(t *testing.T) {
-			// t.Parallel()
+			t.Parallel()
 			testTaskCRUD(t, sys)
 		})
 
@@ -108,15 +108,7 @@ type System struct {
 }
 
 func testTaskCRUD(t *testing.T, sys *System) {
-	orgID := idGen.ID()
-	userID := idGen.ID()
-	if sys.CredsFunc != nil {
-		var err error
-		orgID, userID, _, err = sys.CredsFunc()
-		if err != nil {
-			t.Fatal(err)
-		}
-	}
+	orgID, userID, _ := creds(t, sys)
 
 	// Create a task.
 	task := &platform.Task{OrganizationID: orgID, Owner: platform.User{ID: userID}, Flux: fmt.Sprintf(scriptFmt, 0)}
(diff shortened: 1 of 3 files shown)


task/servicetest/servicetest.go
```diff
@@ -55,7 +55,7 @@ func TestTaskService(t *testing.T, fn BackendComponentFactory) {
 		// but more importantly, it should exercise concurrency to catch data races.
 
 		t.Run("Task CRUD", func(t *testing.T) {
-			t.Parallel()
+			// t.Parallel()
```
mark-rushakoff: Uncomment?

task/servicetest/servicetest.go
```diff
@@ -110,6 +110,13 @@ type System struct {
...
+	if sys.CredsFunc != nil {
+		var err error
+		orgID, userID, _, err = sys.CredsFunc()
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
 
 	// Create a task.
 	task := &platform.Task{OrganizationID: orgID, Owner: platform.User{ID: userID}, Flux: fmt.Sprintf(scriptFmt, 0)}
```
mark-rushakoff: I don't necessarily see the path forward here, with omitting the owner from the response but still accepting it in the create arguments.

I understand that https://github.com/influxdata/influxdb/issues/11427 is to track the big picture, and if work needs to be pushed there, that's fine.

Should this test be updated to check the owner/member API to ensure that the owner matches what was provided? Or would that assertion be part of https://github.com/influxdata/influxdb/issues/11427?

lyondhill: You're right. I think that all TaskCreates should only be able to accept the owner through the context and not in the task struct. It is strange to me that you can set the owner of a task as someone other than yourself.

task/servicetest/servicetest.go
lyondhill: Your right. We should keep this parallel. The problem is we are using the same "system" for all the tests when we should generate a new system for each test.

mark-rushakoff: I think the reason for using the same system, and t.Parallel, was to exercise that the task systems work correctly under concurrent operation. I would prefer to rearrange the test setup over duplicating systems.

lyondhill: We wrote a test just for the purpose of testing concurrency. `testTaskConcurrency` It is entirely unnecessary to use the crud test to check the concurrency as well.

mark-rushakoff: `testTaskConcurrency` should be renamed, as it doesn't exercise all the  methods covered in the other set of tests.

task/servicetest/servicetest.go
lyondhill: After our conversation I thing this is going to be removed in a later pr and is left here until we can make other changes in the backend.

PR #11507, opened 2019-01-23
fix(http): consistently use influxdb errors in task handlers

### LINKED ISSUE #11426: task: use influxdb.Error in all task endpoints
There is currently a mixture of kit errors and influxdb errors in the task endpoints.

Inconsistency will lead to errors that just show up as an EOF, because the JSON parser attempts to read an empty body; or as an empty string because a missing HTTP header is inspected.

https://github.com/influxdata/influxdb/pull/11327 is one such PR that changed several returned errors to explicitly user influxdb.Error.  We need to apply the same process to the other endpoints in http/task_service.go, and make sure `CheckError(err, true)` is used consistently in cmd/influx/task.go.

mark-rushakoff:
And ensure that the task HTTP client consistently expects influxdb
errors, so that they're properly extracted and returned to the caller.

Closes #11426.

### DIFF
--- a/http/task_service.go
+++ b/http/task_service.go
@@ -205,18 +205,29 @@ func (h *TaskHandler) handleGetTasks(w http.ResponseWriter, r *http.Request) {
 
 	req, err := decodeGetTasksRequest(ctx, r)
 	if err != nil {
+		err = &platform.Error{
+			Err:  err,
+			Code: platform.EInvalid,
+		}
 		EncodeError(ctx, err, w)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


mark-rushakoff:
Fair enough, I can make another pass through it and clean those up.

There might be a couple spots where that can't be added - where we have a contract around returning a specific error type, like on a Retry Already Queued error.

mark-rushakoff:
I'm going to merge this now and create a separate issue for better messaging.

PR #11514, opened 2019-01-23
Add types for JSON representation of a Flux AST

chnn:
These types were lifted from the [Flux source](https://github.com/influxdata/flux/blob/master/ast/ast.go).

Eventually it would be nice to generate them directly from the source, rather than writing them by hand. This might prove a bit challenging, since Go's type system is at odds with TypeScript's (there are no union types in Go).

### DIFF
--- a/ui/src/shared/apis/v2/ast.ts
+++ b/ui/src/shared/apis/v2/ast.ts
@@ -1,0 +1,8 @@
+import {queryAPI} from 'src/utils/api'
+import {Package} from 'src/types/ast'
+
+export const getAST = async (query: string): Promise<Package> => {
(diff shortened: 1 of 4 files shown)


nathanielc:
@chnn Could you give me some context on what these types are being used for? Why are they needed instead of using the simple JSON objects that can already serialize the Flux AST?

chnn:
@nathanielc We do use the JSON objects, these types only define the shape/schema of those objects.
They help us program against the AST JSON in the front end more carefully (with compile-time assistance, rather than runtime trial-and-error).

chnn:
Ok great, thanks. I'll keep an eye on the release notes.

PR #11515, opened 2019-01-23
refactor(query/stlib): move from implementation from Flux

affo:
Moves `from` from Flux.

This is the counterpart of https://github.com/influxdata/flux/pull/920.  
Should be fixed once 920 gets merged and reaches flux-staging.
 
  - [ ] Rebased/mergeable
  - [ ] Tests pass


affo:
Changes to `go-mod` and `go.sum` will be removed once 920 gets merged.

The current release this PR is pointing to is:

```
go get github.com/influxdata/flux@refactor/move-from
```

query/stdlib/influxdata/influxdb/from.go
```diff
@@ -3,23 +3,576 @@ package influxdb
...
+	"github.com/influxdata/flux/semantic"
 	"github.com/influxdata/flux/stdlib/influxdata/influxdb"
+	"github.com/influxdata/flux/stdlib/universe"
+	"github.com/influxdata/flux/values"
 	platform "github.com/influxdata/influxdb"
 	"github.com/influxdata/influxdb/query"
 	"github.com/pkg/errors"
 )
 
+const FromKind = "influxDBFrom"
```
wolffcm: Does the `influxdb.from` need a different kind than the one in vanilla Flux?  This seems okay, I'm just curious.

query/stdlib/influxdata/influxdb/from.go
```diff
@@ -3,23 +3,576 @@ package influxdb
...
+	fromSignature := semantic.FunctionPolySignature{
+		Parameters: map[string]semantic.PolyType{
+			"bucket":   semantic.String,
+			"bucketID": semantic.String,
+		},
+		Required: nil,
+		Return:   flux.TableObjectType,
+	}
+
+	flux.ReplacePackageValue("influxdata/influxdb", influxdb.FromKind, flux.FunctionValue(influxdb.FromKind, createFromOpSpec, fromSignature))
```
wolffcm: So for this replacement, you are passing the `influxdb.FromKind`, which is `"from"` and not `"influxDBFrom"`... should it be the latter?

query/stdlib/influxdata/influxdb/from.go
```diff
@@ -3,23 +3,576 @@ package influxdb
...
+
+	newFromSpec := fromSpec.Copy().(*FromProcedureSpec)
+	newFromSpec.LimitSet = true
+	newFromSpec.PointsLimit = -1
+
+	if err := fromNode.ReplaceSpec(newFromSpec); err != nil {
+		return nil, false, err
+	}
+
+	return keysNode, true, nil
```
wolffcm: I'm assuming that you only had to make minimal changes to move this code here so I'm not looking at it thoroughly... let me know if I should look at it closer...

query/stdlib/influxdata/influxdb/from_test.go
```diff
@@ -145,3 +149,703 @@ func TestFromOpSpec_BucketsAccessed(t *testing.T) {
...
+	}
+
+	ps := plantest.CreatePlanSpec(&testSpec)
+	pp := plan.NewPhysicalPlanner(plan.OnlyPhysicalRules())
+	_, err := pp.Plan(ps)
+
+	if err == nil {
+		t.Error("Expected from with no range to fail physical planning")
+	}
+}
```
wolffcm: I'm also assuming that this code was copied in a straightforward manner from the `flux` repo.

query/stdlib/influxdata/influxdb/from.go
affo: Yep, it has to, otherwise, when registering the opSpec it would result in an error. Kinds must be unique, while package values can be replaced.

query/stdlib/influxdata/influxdb/from.go
affo: That is required to replace the actual call of `from`, so that it is overridden by InfluxDB's from. Otherwise, a caller should write:

```
import "influxdb"

influxdb.influxDBFrom(...) |> ...
``` 
What I think is incorrect is here:

```
flux.FunctionValue(influxdb.FromKind, createFromOpSpec, fromSignature)
```

I think it should be:

```
flux.FunctionValue(FromKind, createFromOpSpec, fromSignature)
```

I tried both ways and tests passes, maybe I am missing something.

query/stdlib/influxdata/influxdb/from.go
affo: Exactly, I only changed imports.

query/stdlib/influxdata/influxdb/from_test.go
affo: Same as above :)

query/stdlib/influxdata/influxdb/from.go
wolffcm: Would be good to get to the bottom of that.  I would like to understand it better too.

affo: Yeah, got to the bottom with @jlapacik and it turns out that the kind passed to `FunctionValue` is redundant and not used. The interpreter needs only to know that a certain id in a package is mapped to some function value that invokes a precise `createOpSpec`, no more. However, I'll change to 

```
flux.FunctionValue(FromKind, createFromOpSpec, fromSignature)
```

For consistency with other implementations.

affo:
@wolffcm 
fixed some pieces according to the changes in `querytest/compiler` in Flux.

wolffcm:
Hey @affo Thanks for fixing the compatibility issues... I had touched base with @docmerlin yesterday about the issue with tasks, and the issue with `fieldsAsCols` will be fixed with by @aanthony1243 in this PR:
https://github.com/influxdata/influxdb/pull/11666
So you may want to touch base with them before merging.

affo:
Ok, thank you. I privately asked them and added as reviewers, once they approve, I'll merge.

task/backend/query_logreader.go
```diff
@@ -109,12 +109,14 @@ func (qlr *QueryLogReader) ListRuns(ctx context.Context, runFilter platform.RunF
...
 	listScript := fmt.Sprintf(`
+import "influxdata/influxdb/v1"
+
 from(bucketID: "000000000000000a")
   |> range(start: -24h)
 	|> filter(fn: (r) => r._measurement == "records" and r.taskID == %q)
 	|> drop(columns: ["_start", "_stop"])
 	|> group(columns: ["_measurement", "taskID", "scheduledFor", "status", "runID"])
-	|> influxFieldsAsCols()
+	|> v1.fieldsAsCols()
```
aanthony1243: this is fine.  Maybe we should get one of the task team members to sign off on this?  @mark-rushakoff ?   

I had a related PR where I solved this by replacing with an equivalent pivot()  operation.  whatever the task team wants to use, that's good for me.

affo:
Perfect. I added @mark-rushakoff as reviewer.

query/influxql/end_to_end_test.go
```diff
@@ -15,12 +15,13 @@ import (
 	"github.com/influxdata/flux/execute/executetest"
 	ifql "github.com/influxdata/flux/influxql"
 	"github.com/influxdata/flux/memory"
-	"github.com/influxdata/flux/querytest"
+	fluxQueryTest "github.com/influxdata/flux/querytest"
```
mark-rushakoff: > Good package names are short and clear. They are lower case, with no under_scores or mixedCaps.

https://blog.golang.org/package-names

Please rename to `fluxquerytest`.

query/querytest/compiler.go
```diff
@@ -0,0 +1,20 @@
+package querytest
+
+import (
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/querytest"
+	"github.com/influxdata/flux/stdlib/influxdata/influxdb/v1"
+	"github.com/influxdata/influxdb/query/stdlib/influxdata/influxdb"
+)
+
+// FromInfluxJSONCompiler returns a compiler that replaces all From operations with FromJSON
```
mark-rushakoff: Godoc should be a complete sentence, with punctuation at the end. https://blog.golang.org/godoc-documenting-go-code

task/backend/query_logreader.go
mark-rushakoff: Fine by me. Tests are still passing.

query/stdlib/influxdata/influxdb/from.go
```diff
@@ -3,23 +3,576 @@ package influxdb
...
+		} else {
+			bucket = s.BucketID
+		}
+		return fmt.Errorf(`%s: results from "%s" must be bounded`, id, bucket)
+	}
+
+	return nil
+}
+
+// MergeFromRangeRule pushes a `range` into a `from`
```
mark-rushakoff: Some of the Godoc here and immediately below should have punctuation too.

PR #11520, opened 2019-01-23
chore(docker): Include ca-certificates

BondAnthony:
Include ca-certificates and tzdata in influxdb image.

### DIFF
--- a/docker/influxd/Dockerfile
+++ b/docker/influxd/Dockerfile
@@ -1,13 +1,19 @@
 FROM debian:stable-slim
-COPY influxd /usr/bin/influxd
-COPY influx /usr/bin/influx
+COPY influx influxd /usr/bin/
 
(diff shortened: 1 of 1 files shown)


BondAnthony:
I rebased and played around with moving this to alpine. Did we choose debian for a specific reason? Running this under an alpine image would give back ~74Mb to the users and speed up image build times.

Alpine Dockerfile:
```
FROM alpine:3.8
COPY influx influxd /usr/bin/

EXPOSE 9999

COPY docker/influxd/entrypoint.sh /entrypoint.sh

RUN apk --no-cache add \
		libc6-compat \
		ca-certificates \
		tzdata

ENTRYPOINT ["/entrypoint.sh"]
CMD ["influxd"]
```
The `entrypoint.sh` would need to use `/bin/sh` instead of `/bin/bash`.

PR #11525, opened 2019-01-24
Allow TSI bitset cache to be configured

e-dard:
This PR allows the TSI series id set cache (bitset cache) to be configured via the config. The new config option is as follows:

```toml
[data]
  ...
  tsi-cache-size = 100
  ...
```

The value defaults to `100`. Setting the value to `0` will disable the cache completely. Disabling the cache can lead to query performance degradation. 

From a docs perspective, I would describe this setting as follows:

>The size of the internal cache used in the TSI index to store previously calculated series results. Cached results will be returned quickly from the cache rather than needing to be recalculated when a subsequent query with a matching tag key/value predicate is executed. Setting this value to 0 will disable the cache, which may lead to query performance issues. This value should only be increased if it is known that the set of regularly used tag key/value predicates across all measurements for a database is larger than 100. An increase in cache size may lead to an increase in heap usage.

/cc @stevebang

### DIFF
--- a/tsdb/config.go
+++ b/tsdb/config.go
@@ -66,6 +66,9 @@ const (
 	// DefaultMaxIndexLogFileSize is the default threshold, in bytes, when an index
 	// write-ahead log file will compact into an index file.
 	DefaultMaxIndexLogFileSize = 1 * 1024 * 1024 // 1MB
+
+	// DefaultSeriesIDSetCacheSize is the default number of series ID sets to cache in the TSI index.
+	DefaultSeriesIDSetCacheSize = 100
 )
 
 // Config holds the configuration for the tsbd package.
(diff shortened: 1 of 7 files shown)


tsdb/config.go
```diff
@@ -120,6 +123,11 @@ type Config struct {
 	// be compacted less frequently, store more series in-memory, and provide higher write throughput.
 	MaxIndexLogFileSize toml.Size `toml:"max-index-log-file-size"`
 
+	//  is the number items that can be cached within the TSI index. TSI caching can help
+	// with query performance when the same tag key/value predicates are commonly used on queries.
+	// Setting TSICacheSize disables the cache.
+	TSICacheSize int `toml:"tsi-cache-size"`
+
```
benbjohnson: Why is the field called `TSICacheSize`, the default called `DefaultSeriesIDSetCacheSize`, and the underlying field in `tsi1.Index` called `tagValueCacheSize`? Can we standardize them all to "series id set cache size" or "tag value cache size"? The `TSICacheSize` seems limiting since there could be other caches added.

e-dard: SGTM

tsdb/index/tsi1/index_test.go
```diff
@@ -736,9 +736,14 @@ func BenchmarkIndex_ConcurrentWriteQuery(b *testing.B) {
 	names := make([][]byte, 0, 1e6)
 	tags := make([]models.Tags, 0, 1e6)
 
+	gopath := os.Getenv("GOPATH")
+	if gopath == "" {
+		gopath = build.Default.GOPATH
+	}
```
zeebo: Why not a symlink in the tsi1/testdata folder to the line protocol file?

e-dard:
@benbjohnson @zeebo addressed

PR #11534, opened 2019-01-24
Remove copy-on-write when caching bitmaps

e-dard:
In the case of caching TSI bitmaps belonging to immutable .tsi files,
the underlying bitset data can be mmapped. It is possible, though rare,
for this data to be unmapped (e.g., via a TSI compaction) but for the
cached bitmap to be subsequently read. This leads to a segfault.

This only happens when copy-on-write is set to true on the roaring
bitmap, because in that case only the internal pointers are cloned.

This change will reduce the TSI cache performance by around 10%, which I
have deemed to account for only a few microseconds typically.

### DIFF
--- a/tsdb/series_set.go
+++ b/tsdb/series_set.go
@@ -86,16 +86,6 @@ func (s *SeriesIDSet) Contains(id SeriesID) bool {
 	return x
 }
 
-// SetCOW sets the copy-on-write status of the underlying bitmap. When SetCOW(true)
-// is called, modifications to the bitmap will result in copies of the relevant
-// data structures being made, preventing consumers of clones of the bitmap from
-// seeing those modifications.
-func (s *SeriesIDSet) SetCOW(b bool) {
-	s.Lock()
-	defer s.Unlock()
-	s.bitmap.SetCopyOnWrite(b)
-}
-
 // ContainsNoLock returns true if the id exists in the set. ContainsNoLock is
 // not safe for use from multiple goroutines. The caller must manage synchronization.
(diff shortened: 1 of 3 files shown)


e-dard:
Closing in favour of https://github.com/influxdata/influxdb/pull/11602

mr-wing:
> In the case of caching TSI bitmaps belonging to immutable .tsi files,
> the underlying bitset data can be mmapped. It is possible, though rare,
> for this data to be unmapped (e.g., via a TSI compaction) but for the
> cached bitmap to be subsequently read. This leads to a segfault.
> 
> This only happens when copy-on-write is set to true on the roaring
> bitmap, because in that case only the internal pointers are cloned.
> 
> This change will reduce the TSI cache performance by around 10%, which I
> have deemed to account for only a few microseconds typically.


Hello, I want to ask you a question. According to your statement, it should be not clone before setCOW(true) and there is a clone after setCOW(true), which causes the cache content to point to the data of mmap. But the code I see is just the opposite, not only before setCOW(true) there is clone and there is no clone after setCOW(true).
`	itr := tsdb.MergeSeriesIDIterators(a...)                                                         // clone in or function
	if !EnableBitsetCache {
		return itr, nil
	}

	// Check if the iterator contains only series id sets. Cache them...
	if ssitr, ok := itr.(tsdb.SeriesIDSetIterator); ok {
		ss := ssitr.SeriesIDSet()
		ss.SetCOW(true) // This is important to speed the clone up.	
		i.tagValueCache.Put(name, key, value, ss)                                            // there is no clone before into cache
	}
`

PR #11603, opened 2019-01-25
feat: Add influx ping subcommand

jacobmarble:
For environments that do not have `nc`, `lsof`, `nmap`, `netstat`, but just a way to know if influxdb is up, this subcommand should work nicely.

I'll use it in Kubernetes with the nightly influxdb image, something like this:
```
...
      containers:
      - name: influxdb
        image: quay.io/influxdb/influx:nightly
        lifecycle:
          postStart:
            exec:
              command:
              - sh
              - -c
              - until influx ping; do echo waiting for influxd; sleep 1; done; influx setup -u $USERNAME -p $PASSWORD -o $ORGANIZATION -b $BUCKET -r 720 -t $INFLUX_TOKEN -f
...
```

### DIFF
--- a/cmd/influx/ping.go
+++ b/cmd/influx/ping.go
@@ -3,11 +3,11 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/influxdata/influxdb/kit/check"
-	"github.com/spf13/cobra"
-	"io/ioutil"
 	"net/http"
 	"time"
+
+	"github.com/influxdata/influxdb/kit/check"
+	"github.com/spf13/cobra"
 )
 
 var pingCmd = &cobra.Command{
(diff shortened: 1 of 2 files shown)


cmd/influx/ping.go
```diff
@@ -0,0 +1,54 @@
...
+
+	c := http.Client{
+		Timeout: 5 * time.Second,
+	}
+	url := flags.host + "/health"
+	resp, err := c.Get(url)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode/100 != 2 {
```
mark-rushakoff: Should `defer resp.Body.Close()` here.

cmd/influx/ping.go
```diff
@@ -0,0 +1,54 @@
...
+		return fmt.Errorf("got %d from '%s'", resp.StatusCode, url)
+	}
+
+	b, err := ioutil.ReadAll(resp.Body)
```
mark-rushakoff: Prefer `json.NewDecoder(resp.Body).Decode(&healthResponse)` so that you don't have to read the entire response body up-front (i.e. don't trust the network).

cmd/influx/ping.go
```diff
@@ -0,0 +1,49 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/influxdata/influxdb/kit/check"
```
mark-rushakoff: Just noticed import grouping isn't split.

Issue #11491, opened 2019-01-23
Task: adding member (owner) to a Task does not work


bednar:
When I trying add member to the Task then the response is unsuccessful with an error:

```bash
-> POST 
-> http://127.0.0.1:9999/api/v2/tasks/034a72f16c10e000/members
-> {"id":"034a72f1741c6000"}

<- RESPONSE
<- {"code":"internal error","message":"unsupported resource type tasks"}
```

The problem is that the [FindResourceOrganizationID](https://github.com/influxdata/influxdb/blob/abe6ed3b8ba87bbc06bb72e722158602f487a40e/bolt/organization.go#L514) does not have implementation for the [TasksResourceType](https://github.com/influxdata/influxdb/blob/abe6ed3b8ba87bbc06bb72e722158602f487a40e/authz.go#L109).

desa:
Note that #11701 does not solve this issue.

jademcgough:
@bednar I'll deal with this now that #11664 has been merged, sorry to leave you hanging.

bednar:
@jademcgough Thanks for that

jademcgough:
Believe this was closed by #11741

jademcgough:
Reopening until ^ has been merged

PR #11664, opened 2019-02-04
Add user resource management to the task API

lyondhill:
step 1 of #11427

### DIFF
--- a/http/task_service.go
+++ b/http/task_service.go
@@ -361,7 +361,7 @@ func (h *TaskHandler) handlePostTask(w http.ResponseWriter, r *http.Request) {
 		// clean up the task if we fail to map the user and resource
 		// TODO(lh): Multi step creates could benefit from a service wide transactional request
 		if derr := h.TaskService.DeleteTask(ctx, req.Task.ID); derr != nil {
-			err = fmt.Errorf("%s: failed to clean up task", err.Error())
+			err = fmt.Errorf("%s: failed to clean up task: %s", err.Error(), derr.Error())
 		}
 
 		err = &platform.Error{
@@ -587,7 +587,9 @@ func (h *TaskHandler) handleDeleteTask(w http.ResponseWriter, r *http.Request) {
 		h.logger.Warn("failed to pull user resource mapping", zap.Error(err))
 	} else {
 		for _, m := range urms {
-			h.UserResourceMappingService.DeleteUserResourceMapping(ctx, m.ResourceID, m.UserID)
+			if err := h.UserResourceMappingService.DeleteUserResourceMapping(ctx, m.ResourceID, m.UserID); err != nil {
+				h.logger.Warn(fmt.Sprintf("failed to remove user resource mapping for task %s", m.ResourceID.String()), zap.Error(err))
+			}
 		}
 	}
 


http/task_service.go
```diff
@@ -554,6 +567,17 @@ func (h *TaskHandler) handleDeleteTask(w http.ResponseWriter, r *http.Request) {
 		EncodeError(ctx, err, w)
 		return
 	}
+	// clean up resource maps for deleted task
+	urms, _, err := h.UserResourceMappingService.FindUserResourceMappings(ctx, platform.UserResourceMappingFilter{
+		ResourceID:   req.TaskID,
+		ResourceType: platform.TasksResourceType,
+	})
+
+	if err == nil {
```
mark-rushakoff: We should do something when err is non-nil. Maybe just `log.Warn` it?

http/task_service_test.go
```diff
@@ -760,3 +760,64 @@ func TestTaskHandler_NotFoundStatus(t *testing.T) {
...
+		},
+	}
+	h.handlePostTask(w, r)
+	r = httptest.NewRequest("DELETE", "http://any.url/api/v2/tasks/"+taskID.String(), nil)
+
+	h.ServeHTTP(w, r)
+
+	if created.UserID != deletedUser {
+		t.Fatalf("deleted user (%s) doesn't match created user (%s)", deletedUser, created.UserID)
+
```
mark-rushakoff: Remove blank line.

http/task_service_test.go
```diff
@@ -760,3 +760,64 @@ func TestTaskHandler_NotFoundStatus(t *testing.T) {
...
+		OrganizationID: 1,
+	}
+
+	b, err := json.Marshal(task)
+	if err != nil {
+		t.Fatalf("failed to unmarshal task: %v", err)
+	}
+
+	r := httptest.NewRequest("POST", "http://any.url/v1", bytes.NewReader(b))
+	ctx := pcontext.SetAuthorizer(context.TODO(), &platform.Authorization{UserID: 2})
```
mark-rushakoff: `context.Background()` instead of `context.TODO` here.

http/task_service.go
```diff
@@ -350,6 +350,19 @@ func (h *TaskHandler) handlePostTask(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	urm := &platform.UserResourceMapping{
+		UserID:       auth.GetUserID(),
+		UserType:     platform.Owner,
+		ResourceType: platform.TasksResourceType,
+		ResourceID:   req.Task.ID,
+	}
+	if err := h.UserResourceMappingService.CreateUserResourceMapping(ctx, urm); err != nil {
+		err = &platform.Error{
```
mark-rushakoff: Need to do something with this error. Offhand I don't know what pattern is used in the other parts of the codebase if this call fails.

lyondhill: Yah I will look around. I dont know if we delete the newly created task here or just return an error response?

http/task_service.go
```diff
@@ -350,6 +350,29 @@ func (h *TaskHandler) handlePostTask(w http.ResponseWriter, r *http.Request) {
...
+		UserID:       auth.GetUserID(),
+		UserType:     platform.Owner,
+		ResourceType: platform.TasksResourceType,
+		ResourceID:   req.Task.ID,
+	}
+	if err := h.UserResourceMappingService.CreateUserResourceMapping(ctx, urm); err != nil {
+		// clean up the task if we fail to map the user and resource
+		// TODO(lh): Multi step creates could benefit from a service wide transactional request
+		if derr := h.TaskService.DeleteTask(ctx, req.Task.ID); derr != nil {
+			err = fmt.Errorf("%s: failed to clean up task", err.Error())
```
mark-rushakoff: Let's make sure `derr` is reflected here too. 

I think there's a way you can chain platform.Error together, if you make a platform.Error out of `err` and then associate `derr` somehow.

http/task_service.go
```diff
@@ -554,6 +577,19 @@ func (h *TaskHandler) handleDeleteTask(w http.ResponseWriter, r *http.Request) {
...
+	urms, _, err := h.UserResourceMappingService.FindUserResourceMappings(ctx, platform.UserResourceMappingFilter{
+		ResourceID:   req.TaskID,
+		ResourceType: platform.TasksResourceType,
+	})
+
+	if err != nil {
+		h.logger.Warn("failed to pull user resource mapping", zap.Error(err))
+	} else {
+		for _, m := range urms {
+			h.UserResourceMappingService.DeleteUserResourceMapping(ctx, m.ResourceID, m.UserID)
```
mark-rushakoff: DeleteUserResourceMapping returns an error too, and should be logged if non-nil.

bednar:
Hi @lyondhill 

When I trying create the Task via HTTP then the response is unsuccessful with an error:

```bash
-> POST 
-> http://127.0.0.1:9999/api/v2/tasks
-> {
->	"flux": "option task = {\n    name: \"it task1549451552382-IT\",\n    every: 1h\n}\n\nfrom(bucket:\"my-bucket\") |> range(start: 0) |> last()",
->	"links": {},
->	"name": "it task1549451552382-IT",
->	"orgID": "035c9e6a90ff9000",
->	"status": "active"
-> }

<- RESPONSE
<- {"code":"internal error","message":"failed to add user permissions","error":{"code":"","message":"unsupported resource type tasks"}}
```

The problem is same as in #11491.

Can you look at it?

mark-rushakoff:
@bednar the task creation error you mentioned is most likely fixed by https://github.com/influxdata/influxdb/pull/11701, which is still WIP.

bednar:
@mark-rushakoff I can confirm that it works

PR #11675, opened 2019-02-05
refactor(ui/dataLoaders): Update dataLoaders bucket dropdowns to be the same component

### LINKED ISSUE #11585: Make bucket dropdown component
Both the scraper configuration and the telegraf bundle selection steps use a dropdown with buckets. both of these use buckets that basically work the same way. Let make it one component and reuse so when they need to change, we can change just one.

ischolten:
Closes #11585

Previously, both scrapers and telegraf configs had a dropdown for the buckets in which to write data. Now, there is a bucket component so that they use the same component

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/dataLoaders/components/BucketsDropdown.test.tsx
+++ b/ui/src/dataLoaders/components/BucketsDropdown.test.tsx
@@ -1,0 +1,50 @@
+// Libraries
+import React from 'react'
+import {shallow} from 'enzyme'
+
+// Components
+import BucketsDropdown from 'src/dataLoaders/components/BucketsDropdown'
+import {Dropdown, ComponentStatus} from 'src/clockface'
+import {bucket} from 'mocks/dummyData'
+
+// Mocks
+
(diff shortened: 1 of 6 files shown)


ui/src/dataLoaders/components/BucketsDropdown.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+describe('DataLoading.Components.BucketsDropdown', () => {
+  describe('when no buckets', () => {
+    it('renders disabled dropdown', () => {
+      const {wrapper} = setup()
+
+      const dropdown = wrapper.find(Dropdown)
+
+      expect(wrapper.exists()).toBe(true)
+      expect(dropdown.exists()).toBe(true)
+      expect(dropdown.prop('status')).toBe(ComponentStatus.Disabled)
```
chnn: this would be a great test to write with `react-testing-library`!

PR #11693, opened 2019-02-05
feat(ui/dataLoaders): Use env var for telegraf config auth token

### LINKED ISSUE #11609: Use environmental variable for auth token when saving the telegraf config
We should hardcode using the environmental variable `$INFLUX_TOKEN` in the telegraf config as the auth token for the influxdb_v2 output plugin rather than passing the actual auth token.

blocked on #11608

ischolten:
Closes #11609

Rather than using the actual auth token for the telegraf config, the environmental variable `$INFLUX_TOKEN` will now get saved into the config.

  - [x] Rebased/mergeable
  - [x] Tests pass
  - [x] Update Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## v2.0.0-alpha.2 [unreleased]
 
 ## Features
+1. [11693](https://github.com/influxdata/influxdb/pull/11693): Save the $INFLUX_TOKEN environmental variable in telegraf configs
 
 ## Bug Fixes
 1. [11678](https://github.com/influxdata/influxdb/pull/11678): Update the System Telegraf Plugin bundle to include the swap plugin
--- a/ui/src/dataLoaders/actions/dataLoaders.ts
+++ b/ui/src/dataLoaders/actions/dataLoaders.ts
@@ -283,7 +283,7 @@ export const removePluginBundleWithPlugins = (
   dispatch(removeBundlePlugins(bundle))
 }
 
-export const createOrUpdateTelegrafConfigAsync = (authToken: string) => async (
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## v2.0.0-alpha.2 [unreleased]
 
 ## Features
+1. [11693](https://github.com/influxdata/influxdb/pull/11693): Save the $INFLUX_TOKEN environmental variable in telegraf configs
```
OfTheDelmer: Nice

ui/src/dataLoaders/actions/dataLoaders.ts
```diff
@@ -299,7 +299,7 @@ export const createOrUpdateTelegrafConfigAsync = (authToken: string) => async (
     type: TelegrafPluginOutputInfluxDBV2.TypeEnum.Output,
     config: {
       urls: ['http://127.0.0.1:9999'],
-      token: authToken,
+      token: '$INFLUX_TOKEN',
```
OfTheDelmer: So we won't store the exact token here anymore?  Does that mean the auth token will need to be fetched separately for the `TelegrafInstructionsOverlay` in the other PR I've got open where it was using `plugin.config.token`?

OfTheDelmer: https://github.com/influxdata/influxdb/pull/11677

ischolten: Oh yes i do believe that will affect the Instructions Overlay in that way. Since we are not going to store that anymore, fetching it will be the only to get that value.

PR #11700, opened 2019-02-05
Feat/tasks page adminui

Palakp41:
Making TasksPage component more generic so it is shared with Organization page as well. Created a tasks container component which gets tasks and sends them as props in tasks page

  - [x] Rebased/mergeable
  - [x] Tests pass

screen shot 2019-02-05 at 3 25 58 pm


ui/src/organizations/components/OrgTasksPage.tsx
```diff
@@ -0,0 +1,239 @@
...
+          </FilterList>
+        </Page>
+        {this.renderImportOverlay}
+      </>
+    )
+  }
+
+  private get filteredTasks() {
+    const {tasks, showInactive} = this.props
+    const mappedTasks = tasks.filter(t => {
```
bthesorceror:
```suggestion
    if (showInactive) {
      return tasks
    }
    const mappedTasks = tasks.filter(t => {
```

ui/src/organizations/components/OrgTasksPage.tsx
```diff
@@ -0,0 +1,239 @@
...
+      let activeFilter = true
+      if (!showInactive) {
```
bthesorceror:
```suggestion
      return t.status === TaskAPI.StatusEnum.Active
```

PR #11729, opened 2019-02-06
refactor(query): separate "from" into logical and physical procedures

wolffcm:
Fixes influxdata/flux#270.

  - [x] Rebased/mergeable
  - [x] Tests pass


query/stdlib/influxdata/influxdb/from_test.go
```diff
@@ -335,23 +335,33 @@ func TestFromRangeRule(t *testing.T) {
 
 	for _, tc := range tests {
 		tc := tc
-		t.Run(tc.Name, func(t *testing.T) {
-			t.Parallel()
-			plantest.RuleTestHelper(t, &tc)
-		})
+		if tc.Name == "cannot push range into from" {
```
jlapacik: Why just in this case?

wolffcm: Oops, that was temporary code to debug an issue.  I'll remove it.

PR #11747, opened 2019-02-07
enable Litmus Smoke tests on daily PRs

gshif:
- created a litmus_daily job
- added persist_to_workspace to build job to have access to influxd binary
- added litmus_daily to build workflow

Closes #

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -181,10 +181,10 @@ jobs:
           keys:
             - chronograf-npm-packages-{{ checksum "ui/package-lock.json" }}
       - run: make build
-      persist_to_workspace:
-        root: .
-        paths:
-          - bin/linux/influxd
+      - persist_to_workspace:
+          root: .
+          paths:
+            - bin/linux/influxd
 
   deploy-nightly:
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,9 +6,9 @@ jobs:
       - attach_workspace:
           at: ~/project
       - run: docker login -u=$QUAY_USER -p=$QUAY_PASS quay.io
-      - run: docker run --entrypoint "./run_litmus_tests_oss.sh" -e ONE_TEST=src/cloud/rest_api/smoke/test_smoke.py -e BINARYPATH=/Litmus/result/bin/linux/influxd -e BOLTPATH=/Litmus/result/influxd_test/influxd.bolt -e ENGINEPATH=/Litmus/result/influxd_test --net host -v /var/run/docker.sock:/var/run/docker.sock -v ~/project:/Litmus/result quay.io/influxdb/litmus:latest
+      - run: docker run --entrypoint "./run_litmus_tests_oss.sh" -e ONE_TEST=src/cloud/rest_api/smoke/test_smoke.py -e PROTOSPATH=/Liltmus/result/influxd_test -e BINARYPATH=/Litmus/result/bin/linux/influxd -e BOLTPATH=/Litmus/result/influxd_test/influxd.bolt -e ENGINEPATH=/Litmus/result/influxd_test --net host -v /var/run/docker.sock:/var/run/docker.sock -v ~/project:/Litmus/result quay.io/influxdb/litmus:latest
       - run:
-          name: Litmus Nightly Tests Success
+          name: Litmus Smokte Tests Success
           when: on_success
           command: |
             curl -X POST https://slack.com/api/chat.postMessage \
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -1,10 +1,52 @@
 version: 2
 jobs:
+  litmus_daily:
+    machine: true
+    steps:
+      - attach_workspace:
+          at: ~/project
+      - run: docker login -u=$QUAY_USER -p=$QUAY_PASS quay.io
```
goller: Would you turn these 2 lines into a shell script in etc?

.circleci/config.yml
```diff
@@ -1,10 +1,52 @@
...
+      - run: docker run --entrypoint "./run_litmus_tests_oss.sh" -e ONE_TEST=src/cloud/rest_api/smoke/test_smoke.py -e PROTOSPATH=/Litmus/result/influxd_test -e BINARYPATH=/Litmus/result/bin/linux/influxd -e BOLTPATH=/Litmus/result/influxd_test/influxd.bolt -e ENGINEPATH=/Litmus/result/influxd_test --net host -v /var/run/docker.sock:/var/run/docker.sock -v ~/project:/Litmus/result quay.io/influxdb/litmus:latest
+      - run:
+          name: Litmus Smokte Tests Success
+          when: on_success
+          command: |
+            curl -X POST https://slack.com/api/chat.postMessage \
```
goller: Would you turn these slack posts into a script in etc?

.circleci/config.yml
```diff
@@ -1,10 +1,52 @@
...
+            )
+      - store_artifacts:
+          path: ~/project
+
   litmus_nightly:
     machine: true
     steps:
       - run: docker login -u=$QUAY_USER -p=$QUAY_PASS quay.io
-      - run: docker run --entrypoint "./run_litmus_tests_oss.sh" -e TEST_LIST=tests_lists/gateway_api_tests.list --net host -v /var/run/docker.sock:/var/run/docker.sock -v ~/project:/Litmus/result quay.io/influxdb/litmus:latest
+      - run: docker run --entrypoint "./run_litmus_tests_oss.sh" -e TEST_LIST=tests_lists/gateway_api_tests.list -e DOCKERIMAGE=true --net host -v /var/run/docker.sock:/var/run/docker.sock -v ~/project:/Litmus/result quay.io/influxdb/litmus:latest
```
goller: Shell script here too

.circleci/config.yml
```diff
@@ -139,6 +181,10 @@ jobs:
           keys:
             - chronograf-npm-packages-{{ checksum "ui/package-lock.json" }}
       - run: make build
+      - persist_to_workspace:
+          root: .
+          paths:
+            - bin/linux/influxd
```
goller: Perhaps also persist bin/linux/influx

.circleci/config.yml
BondAnthony: Would this orb work?

https://circleci.com/orbs/registry/orb/circleci/slack

goller: @gshif Yeah, I like @BondAnthony 's orb idea for sure!

gshif: Since slack notifications for litmus nightly job are already there, I would not want to make the changes right now. I can create the ticket to look into that when I have more time. My priority is to add more tests. Would that work?

goller: Nah... I think you should do them now.  The reason is that it'll set the pattern for other builds sending to slack.  This will lead us in a really positive direction.

gshif: Why don't we use the slack notifications that are already there? This PR is about adding Litmus daily test runs and I do not want to hold it because of notifications.

.circleci/config.yml
gshif: Added `- bin/linux/influx` to be shared between jobs in a workspace

.circleci/config.yml
gshif: Created two shell scripts to be used for slack posts.

gshif: Re: orbs suggested by @BondAnthony. It is not just import and use. It requires changes to CircleCI configuration, that I would like to leave to SREs.

BondAnthony:
LGTM, sorry I misspoke about the orbs. I didn’t realize those were only 2.1 features. I’ll add orbs to the list of changes that should be done once 2.1 goes GA. @goller upgrading to CircleCI 2.1 brings additional risk since it’s classified as experimental, however once it’s released we will have a number of features to leaverage.

BondAnthony:
We would need to enable build processing (preview) feature after converting all the configuration to 2.1. I have this on the list,  I think we need to run a number of tests to validate builds still work as expected but I’m not sure how stable their “preview” is.

PR #11750, opened 2019-02-07
Update tagKeyValue mutex to write lock.

benbjohnson:

This commit changes the read lock to a write lock when calling the
`ids()` function because `ids()` can mutate the underlying series
ids slice.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/tsdb/index/inmem/meta.go
+++ b/tsdb/index/inmem/meta.go
@@ -1128,8 +1128,14 @@ func (t *tagKeyValue) Load(value string) seriesIDs {
 
 	t.mu.RLock()
 	entry := t.entries[value]
-	ids := entry.ids()
+	ids, changed := entry.ids()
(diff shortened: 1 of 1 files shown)


e-dard:
Any performance implications here? What code paths use `Range` and `Load` that are no longer going to be running concurrently?

benbjohnson:
It looks like `Range()` is used directly when checking for authorized series. `RangeAll()` is for `foo != ''` and regex queries as well anywhere that all tag keys & tag values need to be iterated over. `Load()` is used when fetching IDs for individual tag values.

It seems like the main performance implication will be adding planning time for regex & NEQ queries .

benbjohnson:
@e-dard This implements the `changed` return you suggested. Can you take a look please?

PR #11756, opened 2019-02-07
Fix the inline edit for org tasks page

Palakp41:
Fix the inline edit for org tasks page

Update TaskRow to take in optional parameter for updateTask name.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/EditableName.tsx
+++ b/ui/src/shared/components/EditableName.tsx
@@ -15,7 +15,7 @@ import 'src/shared/components/EditableName.scss'
 interface Props {
   onUpdate: (name: string) => void
   name: string
-  onEditName: () => void
+  onEditName: (e?: any) => void
   hrefValue?: string
   placeholder?: string
 }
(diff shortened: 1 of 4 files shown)


ui/src/tasks/components/TaskRow.tsx
```diff
@@ -98,6 +94,33 @@ export class TaskRow extends PureComponent<Props & WithRouterProps> {
     )
   }
 
+  private get resourceNames(): JSX.Element {
+    const {onUpdate, task} = this.props
+    if (onUpdate) {
```
OfTheDelmer: nice

ui/src/shared/components/EditableName.tsx
```diff
@@ -15,7 +15,7 @@ import 'src/shared/components/EditableName.scss'
 interface Props {
   onUpdate: (name: string) => void
   name: string
-  onEditName: () => void
+  onEditName: (e?: any) => void
```
OfTheDelmer: might be better to use a ChangeEvent here

Palakp41: I agree, I will fix it in next pr.

PR #11760, opened 2019-02-07
fix(query/influxql): normalize selectors when an interval is present

### LINKED ISSUE #10747: influxql: selectors do not normalize their time to the start time
We are going to need some way to join times together for different aggregate streams. When a selector is used as an aggregate, it still retains the original time in the `_time` column within the transpiler. If we try to join it with an aggregate, it won't succeed.

But, if we try to join with `_start`, we need to make sure to perform that join before using `window(every: inf)` when we put them all into the same table since that process will erase the start time of the interval. And we also need to do this in circumstances where there is no `window()` function because it happens with aggregate queries too.

jsternberg:
The selector will be normalized to the start time interval when `GROUP
BY time(X)` is present in the query.

Fixes #10747.

### DIFF
--- a/query/influxql/README.md
+++ b/query/influxql/README.md
@@ -149,7 +149,7 @@ If the aggregate is combined with conditions, the column name of `_value` is rep
 
 #### <a name="normalize-time"></a> Normalize the time column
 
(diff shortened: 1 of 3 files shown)


jsternberg:
Those are skip test declarations.

PR #11770, opened 2019-02-08
Add support for describing data-generation schema via a TOML file

stuartcarnie:
The `influx_tools generate` command has a new option, `-schema`, for
specifying a TOML file which describes a desired schema.

Use the `-example` switch to print out documentation for the schema, similar to the following:

```toml
title = "CLI schema"

# limit the maximum number of series generated across all measurements
#
# series-limit: integer, optional (default: unlimited)
# multiple measurements are merged together
[[measurements]]
# name of measurement

name = "cpu"

# sample: float; where 0 < sample ≤ 1.0 (default: 0.5)
#   sample a subset of the tag set
#
# sample 25% of the tags
#
# sample = 0.25

# Keys for defining a tag
#
# name: string, required
#   Name of field
#
# source: array<string> or object
# 
#   A literal array of string values defines the tag values.
#
#   An object defines more complex generators. The type key determines the
#   type of generator.
#
# source types:
#
# type: "sequence" 
#   generate a sequence of tag values
#
#       format: string
#           a format string for the values (default: "value%s")
#       start: int (default: 0)
#           beginning value 
#       count: int, required
#           ending value
#
# type: "file"
#   generate a sequence of tag values from a file source.
#   The data in the file is sorted, deduplicated and verified is valid UTF-8
#
#       path: string
#           absolute path or relative path to current toml file
tags = [
    # example sequence tag source. The range of values are automatically prefixed with 0s
    # to ensure correct sort behavior.
    { name = "host",   source = { type = "sequence", format = "host-%s", start = 0, count = 5 } },

    # tags can also be sourced from a file. The path is relative to the schema.toml.
    # Each value must be on a new line. The file is also sorted, validated for UTF-8 and deduplicated.
    # { name = "region", source = { type = "file", path = "files/regions.txt" } },

    # Example string array source, which is also deduplicated and sorted
    { name = "region", source = ["us-west-01","us-west-02","us-east"] },
]

# Keys for defining a field
#
# name: string, required
#   Name of field
#
# count: int, required
#   Number of values to generate. When multiple fields have the same 
#   count, they will share timestamps.
#
# time-precision: string (default: ms)
#   The precision for generated timestamps. 
#   One of ns, us, ms, s, m, h
#
# source: int, float, boolean, string, array or object
# 
#   A literal int, float, boolean or string will produce 
#   a constant value of the same data type.
#
#   A literal array of homogeneous values will generate a repeating 
#   sequence.
#
#   An object defines more complex generators. The type key determines the
#   type of generator.
#
# source types:
#
# type: "rand<float>" 
#   generate random float values
#       seed: seed to random number generator (default: 0)
#       min:  minimum value (default: 0.0)
#       max:  maximum value (default: 1.0)
#
# type: "zipf<integer>" 
#   generate random integer values using a Zipf distribution
#   The generator generates values k ∈ [0, imax] such that P(k) 
#   is proportional to (v + k) ** (-s). Requirements: s > 1 and v ≥ 1.
#   See https://golang.org/pkg/math/rand/#NewZipf for more information.
#
#       seed: seed to random number generator (default: 0)
#       s:    float > 1 (required)
#       v:    float ≥ 1 (required)
#       imax: integer (required)
#
fields = [
    # Example constant float
    { name = "system", count = 5000, source = 2.5 },
    
    # Example random floats
    { name = "user",   count = 5000, source = { type = "rand<float>", seed = 10, min = 0.0, max = 1.0 } },
]

# Multiple measurements may be defined.
[[measurements]]
name = "mem"
tags = [
    { name = "host",   source = { type = "sequence", format = "host-%s", start = 0, count = 5 } },
    { name = "region", source = ["us-west-01","us-west-02","us-east"] },
]
fields = [
    # An example of a sequence of integer values
    { name = "free", count = 17, source = [10,15,20,25,30,35,30], time-precision = "ms" },
    { name = "low_mem", count = 17, source = [false,true,true], time-precision = "ms" },
]
```


pkg/data/gen/specs_test.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	spec, err := NewSpecFromSchema(&out)
+	if err != nil {
+		t.Error(err)
+	}
+	t.Log(spec)
+}
+
+func TestSpecFromSchemaError(t *testing.T) {
```
zeebo: Seems like this test is the same thing as above? It doesn't appear to be checking for an error.

stuartcarnie: Ahh you are right – I'll take a look at that. Neither test should be logging either

PR #11792, opened 2019-02-10
Revert "Limit force-full and cold compaction size."

benbjohnson:
This reverts commit 381f206f4acbd34698b1349582117a551a918d62. (https://github.com/influxdata/influxdb/pull/10593)


johnnyjian:
We encounter the issue of full compacting too many tsm files (1000+) which consuming too many heap memory. I wonder why this commit is reverted.

PR #11795, opened 2019-02-11
Add histogram visualization type

### LINKED ISSUE #11730: Display tooltip on bar hover
When hovering over a bar in the histogram, a tooltip should display what group the bar belongs to, and how many observations the bar corresponds to.

### LINKED ISSUE #11732: Add color scheme configuration to histogram
I should be able to choose which color scheme is used in a histogram, a lá

screen shot 2019-02-06 at 2 48 18 pm

chnn:
Closes #11730 
Closes #11732 

Creates a histogram visualization type.

Data in the histogram can be grouped and regrouped on the fly:

reground

The binning statistic can also be recomputed dynamically:

bins

Histograms can either be stacked or overlaid:

stacked

Histograms have a tooltip on hover:

tooltip

## Implementation

This PR sketches out a new visualization framework (everything under `src/minard`), which implements the histogram visualization. I plan to post an RFC later today or tomorrow that motivates and describes this approach in more depth. For now, I'm opening this PR so we can get some feedback on the histogram from the next alpha release.

The gist of the approach is that the framework expects a single data table in a column-oriented format. This is passed to a `Plot` component, which sets up XY scales, axes, margins, and so forth. Then different columns of the table are mapped to different _aesthetics_ of the plot (e.g. "map the `_value` column to the _x_ aesthetic of the plot"). As children, the `Plot` accepts one or more layers. A full example:

```
<Plot table={table} width={600} height={400}>
  {env => <Histogram x="_value" fill="host" env={env} />}
</Plot>
```
 
Right now the `Histogram` is the only implemented layer. But in the future there could be many other types of layers... a scatter plot and linear regression layer perhaps, or a heatmap, etc.

```
<Plot table={table} width={600} height={400} x="_time" y="_value" fill="cpu">
  {env => (
    <Point env={env} />
    <Smooth env={env} />
  )}
</Plot>
```

If a layer doesn't have aesthetic mappings specified, like above, it will inherit the aesthetic mappings from the parent `Plot`. Layers can aggregate or transform data however they please (for example, to compute a histogram).

There's a few other details to the design described [here](http://vita.had.co.nz/papers/layered-grammar.pdf). This is the same approach used with great success in [ggplot2](https://ggplot2.tidyverse.org/reference/index.html).

Note that the implementation is still very rough. Various `TODO`s are strewn about the code, which I will address in the next few days. Some notable missing pieces:

- The visualization options for the histogram are hard coded
- The histogram view type isn't defined serverside yet, so histograms cannot be saved to a cell yet
- The framework hasn't been extracted to a package yet

### DIFF
--- a/ui/src/minard/components/Axes.tsx
+++ b/ui/src/minard/components/Axes.tsx
@@ -1,0 +1,97 @@
+import React, {useRef, useLayoutEffect, SFC} from 'react'
+
+import {PlotEnv, TICK_PADDING_RIGHT, TICK_PADDING_TOP} from 'src/minard'
+import {clearCanvas} from 'src/minard/utils/clearCanvas'
+
+interface Props {
+  env: PlotEnv
+  axesStroke?: string
+  tickFont?: string
+  tickFill?: string
+}
+
+export const drawAxes = (
+  canvas: HTMLCanvasElement,
+  env: PlotEnv,
+  axesStroke: string,
+  tickFont: string,
+  tickFill: string
+) => {
+  const {
+    width,
+    height,
+    margins,
+    xTicks,
+    yTicks,
+    defaults: {
+      scales: {x: xScale, y: yScale},
+    },
+  } = env
+
+  clearCanvas(canvas, width, height)
+
+  const context = canvas.getContext('2d')
+  const xAxisY = height - margins.bottom
+
+  // Draw x axis line
+  context.strokeStyle = axesStroke
+  context.beginPath()
+  context.moveTo(margins.left, xAxisY)
+  context.lineTo(width - margins.right, xAxisY)
+  context.stroke()
+
+  // Draw y axis line
+  context.beginPath()
+  context.moveTo(margins.left, xAxisY)
+  context.lineTo(margins.left, margins.top)
+  context.stroke()
+
+  context.font = tickFont
+  context.fillStyle = tickFill
+  context.textAlign = 'center'
+  context.textBaseline = 'top'
+
+  // Draw and label each tick on the x axis
+  for (const xTick of xTicks) {
+    const x = xScale(xTick) + margins.left
+
+    context.beginPath()
+    context.moveTo(x, xAxisY)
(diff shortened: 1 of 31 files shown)


ui/src/minard/components/Axes.tsx
```diff
@@ -0,0 +1,93 @@
...
+import {clearCanvas} from 'src/minard/utils/clearCanvas'
+
+interface Props {
+  env: PlotEnv
+  axesStroke?: string
+  tickFont?: string
+  tickFill?: string
+}
+
+export const drawAxes = (
```
AlirieGray: would it be possible to break this function up, or add some comments to explain what each block does?

ui/src/minard/utils/assert.ts
```diff
@@ -0,0 +1,5 @@
+export const assert = (message: string, condition: boolean) => {
```
AlirieGray: nice

ui/src/minard/utils/plotEnvReducer.ts
```diff
@@ -0,0 +1,247 @@
...
+    } else if (defaultLayer.table) {
+      fillCol = defaultLayer.table.columns[fillColName]
+    }
+
+    assert(`couldnt find column ${fillColName} for fill`, !!fillCol)
+
+    return fillCol
+  }
+
+  // Compute fill scales for layers that require their own scale
```
AlirieGray: nice

ui/src/shared/utils/toMinardTable.test.ts
```diff
@@ -0,0 +1,100 @@
...
+          table: [0, 0, 1, 1],
+        },
+      },
+    }
+
+    expect(actual).toEqual(expected)
+  })
+
+  test('with a schema conflict', () => {
+    const CSV = `#group,false,false,true,true,false,false,true,true,true,true
```
AlirieGray: we could put these CSV strings in a constants/dummy data file

PR #11804, opened 2019-02-11
Create Tasks Run Page for dispalying runs for specific TaskID

### LINKED ISSUE #11501: Add the ability to view the runs of a specific task
as a task user
with admin
when i view the list of tasks 
then i would like to view the runs of a specific task

Palakp41:
Closes #11501 

Updated route to display new page when clicked on view runs on tasks page. Using dummy data to display the runs, will create a new card to update the client and use the api response once the api is fixed.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/tasks/components/TaskRunsPage.tsx
+++ b/ui/src/tasks/components/TaskRunsPage.tsx
@@ -1,0 +1,73 @@
+// Libraries
+import React, {PureComponent} from 'react'
+
+// Types
+import {Run} from '@influxdata/influx'
+import {IndexList} from 'src/clockface'
+import {taskRuns} from '../dummyData'
+import {Page} from 'src/pageLayout'
+
+interface Props {
+  taskID: string
+  runs: Run[]
+}
+
+const dummyData = taskRuns
+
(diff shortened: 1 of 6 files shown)


ui/src/tasks/components/TaskRunsPage.tsx
```diff
@@ -0,0 +1,65 @@
...
+        </Page>
+      </>
+    )
+  }
+
+  public get listRuns(): JSX.Element[] {
+    const taskRuns = dummyData.map(t => (
+      <IndexList.Row key={`task-id--${t.id}`}>
+        <IndexList.Cell>{t.requestedAt.toDateString()}</IndexList.Cell>
+        <IndexList.Cell>{t.startedAt.toDateString()}</IndexList.Cell>
```
bthesorceror: We might want to be a bit more granular than day here, i.e. if I have multiple runs of a task in the same day.

Palakp41: Fixed it! :)

PR #11827, opened 2019-02-12
Update task http handler to use user resource mapping.

lyondhill:
part 3 of #11427

### DIFF
--- a/http/task_service.go
+++ b/http/task_service.go
@@ -262,14 +262,39 @@ func (h *TaskHandler) handleGetTasks(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	tasks, _, err := h.TaskService.FindTasks(ctx, req.filter)
-	if err != nil {
-		err = &platform.Error{
-			Err: err,
-			Msg: "failed to find tasks",
+	var tasks []*platform.Task
+
+	if req.filter.User != nil {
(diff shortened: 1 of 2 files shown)


http/task_service.go
```diff
@@ -262,14 +262,39 @@ func (h *TaskHandler) handleGetTasks(w http.ResponseWriter, r *http.Request) {
...
+		ownedTasks, _, err := h.UserResourceMappingService.FindUserResourceMappings(
+			ctx,
+			platform.UserResourceMappingFilter{
+				UserID:       *req.filter.User,
+				UserType:     platform.Owner,
+				ResourceType: platform.TasksResourceType,
+			},
+		)
+		if err != nil {
+			EncodeError(ctx, err, w)
```
mark-rushakoff: For consistency, we should probably wrap `err` with a `platform.Error`.

http/task_service.go
```diff
@@ -262,14 +262,39 @@ func (h *TaskHandler) handleGetTasks(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+		for _, ownedTask := range ownedTasks {
+			task, err := h.TaskService.FindTaskByID(ctx, ownedTask.ResourceID)
+			if err != nil {
+				EncodeError(ctx, err, w)
```
mark-rushakoff: Probably want to wrap this error too.

PR #11831, opened 2019-02-12
Convert TagValueSeriesIDCache to use string fields

benbjohnson:
Backport of https://github.com/influxdata/influxdb/pull/11829


stuartcarnie:
This should also close #11639

PR #11832, opened 2019-02-12
Ensure that cached series id sets are Go heap backed

zeebo:
This way we don't have problems cloning it for reads if the backing memory gets unmapped.


e-dard:
@zeebo this is some of your finest work.

PR #11855, opened 2019-02-13
feat(kv): add lookup service

### LINKED ISSUE #11717: kv: port lookup service from bolt

goller:
Closes #11717 

Implements lookup service using the key-value store.  Ported from BoltDB impl.

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/kv/lookup_service.go
+++ b/kv/lookup_service.go
@@ -45,15 +45,12 @@ func (s *Service) Name(ctx context.Context, resource influxdb.ResourceType, id i
 			return "", err
 		}
 		return r.Name, nil
-		// TODO (goller): once telegraf is merged into kv/service
-		/*
-			case influxdb.TelegrafsResourceType: // 6
-				r, err := s.FindTelegrafConfigByID(ctx, id)
-				if err != nil {
-					return "", err
-				}
-				return r.Name, nil
-		*/
+	case influxdb.TelegrafsResourceType: // 6
+		r, err := s.FindTelegrafConfigByID(ctx, id)
+		if err != nil {
+			return "", err
+		}
+		return r.Name, nil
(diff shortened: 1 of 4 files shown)


kv/lookup_service.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"context"
+
+	"github.com/influxdata/influxdb"
+)
+
+var _ influxdb.LookupService = (*Service)(nil)
+
+// Name returns the name for the resource and ID.
+func (s *Service) Name(ctx context.Context, resource influxdb.ResourceType, id influxdb.ID) (string, error) {
```
desa: comepletely unrelated, but what if we did something like a generic `Find` with something like this

```go
func (s *Service) Find(ctx context.Context, resource influxdb.ResourceType, id influxdb.ID, v interface{})  error {}
```
we could map `resource -> kv.bucket` and lookup the look up the ID of the resource in that bucket.

kv/lookup_service.go
```diff
@@ -0,0 +1,66 @@
...
+		return "", err
+	}
+
+	if ok := id.Valid(); !ok {
+		return "", influxdb.ErrInvalidID
+	}
+
+	switch resource {
+	case influxdb.TasksResourceType: // 5 // TODO(goller): unify task storage here so we can lookup names
+	case influxdb.AuthorizationsResourceType: // 0 TODO(goller): authorizations should also have optional names
```
desa: Maybe we should change `Description` to be name?

kv/lookup_service.go
goller: Yeah I think a generic find would be nice for sure.  This refactor has shown me we could have a few functions like that Find that could be used by our implementations.

Issue #11669, opened 2019-02-04
task: allow changing AuthorizationID


mark-rushakoff:
Follows https://github.com/influxdata/influxdb/issues/11668.

Allow `authorizationID` as a patch parameter on a task, which will override the authorization ID set on its StoreTaskMeta. Only Owners of a task can change the authorizationID, so this may need to wait until that part of https://github.com/influxdata/influxdb/issues/11427 is ready.

Update the swagger document to reflect this API change.

mark-rushakoff:
Closed by #11876.

PR #11876, opened 2019-02-13
Plumb authorization through tasks

### LINKED ISSUE #11668: task: store authorizationID with task metadata
1. Add a new field, AuthorizationID, to StoreTaskMeta.
2. When a new task is created, take the AuthorizationID associated with the create request, and store it with the StoreTaskMeta.
3. Outgoing requests originated from a task -- particularly, such as requests to the query service -- should associate that AuthorizationID with that request's context.
4. If the authorization ID is invalid when the task executes, continue to run and fail; do not deactivate the task.

mark-rushakoff:
Closes #11668, #11669.


http/swagger.yml
```diff
@@ -6929,4 +6933,35 @@ components:
...
+          type: string
+        cron:
+          description: Override the 'cron' option in the flux script.
+          type: string
+        offset:
+          description: Override the 'offset' option in the flux script.
+          type: string
+        token:
+          description: Override the existing token associated with the task.
+          type: string
```
docmerlin: add carriage return to end of file.

docmerlin:
Looks good, only saw one stylistic issue, nothing substantive.

http/swagger.yml
mark-rushakoff: I'm not going to wait for another CI cycle to get the EOL there, when it wasn't there before. Do we need to standardize on [EditorConfig](https://editorconfig.org/) or something else to ensure the style going forward?

PR #11879, opened 2019-02-13
Combine permisssions in the ui on token overlay

### LINKED ISSUE #11780: permissions should be combined in the UI when appropriate
when viewing the token permissions in the UI, the read/write sections for a particular resource should be combined.

Palakp41:
Closes #11780 

Updated token overlay to create an object with unique keys and values array for multiple actions

  - [x] Rebased/mergeable
  - [x] Tests pass

screen shot 2019-02-13 at 2 08 49 pm

### DIFF
--- a/ui/src/me/components/account/ViewTokenOverlay.tsx
+++ b/ui/src/me/components/account/ViewTokenOverlay.tsx
@@ -30,6 +30,15 @@ export default class ViewTokenOverlay extends PureComponent<Props> {
     const {description, permissions} = this.props.auth
     const {onNotify} = this.props
 
+    const permissionsByType = {}
+    for (const key of permissions) {
+      if (permissionsByType[key.resource.type]) {
+        permissionsByType[key.resource.type].push(key.action)
+      } else {
+        permissionsByType[key.resource.type] = [key.action]
+      }
+    }
+
     return (
       <OverlayContainer>
         <OverlayHeading title={description} onDismiss={this.handleDismiss} />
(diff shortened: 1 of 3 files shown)


ui/src/me/components/account/ViewTokenOverlay.tsx
```diff
@@ -15,21 +15,26 @@ import {Authorization, Permission} from '@influxdata/influx'
...
 }
 
-const actions = [Read, Write]
-
 export default class ViewTokenOverlay extends PureComponent<Props> {
   public render() {
     const {description, permissions} = this.props.auth
     const {onNotify} = this.props
 
+    const obj = {}
```
chnn: I often see objects like this called something like `permissionsByType`, letting you know that it's an object whose keys are types and values are permissions.

A more functional style of building up that object could go something like:
```
permissions.reduce((acc, {action, resource: {type}) => ({
  ...acc,
  [type]: [...acc[type], action]
}), {})
```
Whether or not that's more readable is not so certain :)

ui/src/me/components/account/ViewTokenOverlay.tsx
```diff
@@ -39,20 +44,20 @@ export default class ViewTokenOverlay extends PureComponent<Props> {
             mode={PermissionsWidgetMode.Read}
             heightPixels={500}
           >
-            {permissions.map((p, i) => {
+            {Object.keys(obj).map((p, i) => {
```
chnn: `Object.entries` could be helpful here:

```suggestion
            {Object.entries(obj).map(([type, permissions]) => {
```

then you don't need to index back into `obj` later

ui/src/me/components/account/ViewTokenOverlay.tsx
```diff
@@ -39,20 +44,20 @@ export default class ViewTokenOverlay extends PureComponent<Props> {
...
                   mode={PermissionsWidgetMode.Read}
                 >
-                  {actions.map((a, i) => (
+                  {obj[p].map((a, i) => (
                     <PermissionsWidget.Item
                       key={i}
                       id={this.itemID(p, a)}
                       label={a}
-                      selected={this.selected(p, a)}
+                      selected={PermissionsWidgetSelection.Selected}
```
chnn: Should each `Item` always be selected?

Palakp41: Thanks for finding that, I fixed it now

ui/src/me/components/account/ViewTokenOverlay.tsx
Palakp41: I don't want to loop through an array of type then array of permissions. So that's why I am looping on the keys only.

ui/src/me/components/account/ViewTokenOverlay.tsx
Palakp41: Thanks!

Issue #11063, opened 2019-01-14
task: investigate consolidating bolt and in-mem store behind kv interface


mark-rushakoff:

lyondhill:
After investigating the current shortcomings of our system and a few meetings a refactor is in order. It should relieve a significant amount of technical debt and put us in a good place for code readability, stability, and transactionality. 

As a result of this investigation I have created a number of new issues.
#12724 
#12725 
#12726
#12727 
#12728 
#12729 
#12730

PR #11882, opened 2019-02-13
Clean out user from the task store's

lyondhill:
completes step 4 of #11427

### DIFF
--- a/task/platform_adapter.go
+++ b/task/platform_adapter.go
@@ -219,11 +219,10 @@ func (p pAdapter) DeleteTask(ctx context.Context, id platform.ID) error {
 
 	if err != nil {
 		return err
-	} else {
-		for _, m := range urms {
-			if err := p.urm.DeleteUserResourceMapping(ctx, m.ResourceID, m.UserID); err != nil {
-				return err
-			}
+	}
+	for _, m := range urms {
+		if err := p.urm.DeleteUserResourceMapping(ctx, m.ResourceID, m.UserID); err != nil {
+			return err
 		}
 	}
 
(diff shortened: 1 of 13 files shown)


task/platform_adapter.go
```diff
@@ -12,6 +12,10 @@ import (
 	"github.com/influxdata/influxdb/task/options"
 )
 
+// TODO(lh): A temporary error response for all finds that use user id's
+// This should be removed when we transition to the new kv store.
+var ErrTempDisabled = errors.New("temporarily disabled")
```
mark-rushakoff: We started but didn't resolve an offline conversation about this.

We shouldn't be using this error. The platform adapter should instead accept a UserResourceMappingService, and handle the User filter case by looking up that user's tasks and then calling `FindTaskByID` on each individual task.

We will refactor that to a more efficient approach as part of the proposed refactor in https://github.com/influxdata/influxdb/issues/11063.

lyondhill: I will make that change. It should be a clean change and allow it to be a little more permanent then the current solution.

task/platform_adapter.go
```diff
@@ -161,8 +206,28 @@ func (p pAdapter) UpdateTask(ctx context.Context, id platform.ID, upd platform.T
...
+
+	// clean up resource maps for deleted task
+	urms, _, err := p.urm.FindUserResourceMappings(ctx, platform.UserResourceMappingFilter{
+		ResourceID:   id,
+		ResourceType: platform.TasksResourceType,
+	})
+
+	if err != nil {
+		return err
+	} else {
```
mark-rushakoff: Nit: could remove the `else` here so that the for loop isn't indented.

task/servicetest/servicetest.go
```diff
@@ -174,17 +174,17 @@ func testTaskCRUD(t *testing.T, sys *System) {
...
 	}
-	f = findTask(fs, task.ID)
+	f = findTask(fs, tsk.ID)
 	found["FindTasks with Organization filter"] = f
 
 	fs, _, err = sys.ts.FindTasks(sys.Ctx, platform.TaskFilter{User: &cr.UserID})
 	if err != nil {
 		t.Fatal(err)
+	} else {
+		f = findTask(fs, tsk.ID)
```
mark-rushakoff: This could/should be moved out of the `else` too. `else` looks unusual after a `t.Fatal`.

PR #11886, opened 2019-02-13
Add orgID property to Macro (variable) type in swagger

AlirieGray:
This PR adds an orgID to the Macro type in the swagger documentation

  - [x] Rebased/mergeable
  - [x] Tests pass


desa:
👍 

Would it be possible to update https://github.com/influxdata/influxdb/blob/8f0e3f5e78cbe93333ab1be5786eb7643122b1e2/macro.go#L48

to match the swagger . This might require changing a few go tests as well.

PR #11894, opened 2019-02-14
fix(query/stdlib/influxdb): make FromOpSpec a BucketAwareSpec

affo:
Closes https://github.com/influxdata/flux/issues/114

there was a cyclic dependency between Flux and InfluxDB because of `from` being in Flux.
Now that `from` is in InfluxDB, pre-authorization can work for `from`.

I fixed every test that contained "https://github.com/influxdata/flux/issues/114" in the `Skip` reason.  
If there is something else I should be aware of, please tell me.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/query/stdlib/influxdata/influxdb/to.go
+++ b/query/stdlib/influxdata/influxdb/to.go
@@ -172,7 +172,13 @@ func (ToOpSpec) Kind() flux.OperationKind {
 
 // BucketsAccessed returns the buckets accessed by the spec.
 func (o *ToOpSpec) BucketsAccessed() (readBuckets, writeBuckets []platform.BucketFilter) {
-	bf := platform.BucketFilter{Name: &o.Bucket, Organization: &o.Org}
+	bf := platform.BucketFilter{}
+	if o.Bucket != "" {
+		bf.Name = &o.Bucket
+	}
+	if o.Org != "" {
+		bf.Organization = &o.Org
+	}
 	if o.OrgID != "" {
 		id, err := platform.IDFromString(o.OrgID)
 		if err == nil {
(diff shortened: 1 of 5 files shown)


gshif:
@affo , the litmus tests is failing to create a tasks with the error `Failed to authorize. no read permission for bucket: "<bucket name>"`. I am working on getting more details.

affo:
@gshif is pre-authorization making some other test fail?  
This PR is precisely about that

gshif:
@affo , The task creation is now validating the existence of a from/to buckets and then verifying user has read/write permissions for these buckets. I have updated the smoke test to use `super user`'s token to create tasks (has all the permissions in the world). Before, the token I was using missed a `read` permission. I have a PR, once it is green, I will merge into master and re-run the litmus_smoke test.

affo:
@gshif oh yeah, now I remember that you were talking exactly about it yesterday at standup!  
Got it 👍

gshif:
@affo , smoke test passed

affo:
Thank you @gshif 👍

query/stdlib/influxdata/influxdb/to.go
```diff
@@ -172,7 +172,13 @@ func (ToOpSpec) Kind() flux.OperationKind {
 
 // BucketsAccessed returns the buckets accessed by the spec.
 func (o *ToOpSpec) BucketsAccessed() (readBuckets, writeBuckets []platform.BucketFilter) {
-	bf := platform.BucketFilter{Name: &o.Bucket, Organization: &o.Org}
+	bf := platform.BucketFilter{}
+	if o.Bucket != "" {
```
aanthony1243: no need to change this, but it appears equivalent to the original line.  You avoid 2 assignments in the case where the values are known to be nil

affo: Yeah, you are right

PR #11897, opened 2019-02-14
Histogram polish

chnn:
Adds an empty state to the histogram "Column" option:

screen shot 2019-02-14 at 10 42 49 am

Updates the tooltip positioning so that it can overflow the `Plot` component. It will detect collisions with the edge of the screen instead:

positioning

### DIFF
--- a/ui/src/minard/utils/useMousePos.ts
+++ b/ui/src/minard/utils/useMousePos.ts
@@ -1,40 +1,38 @@
-import {useState, useEffect, MutableRefObject} from 'react'
+import {useState, useEffect} from 'react'
 
-export const useMousePos = (
-  ref: MutableRefObject<Element>
-): [number, number] => {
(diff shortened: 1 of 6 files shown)


ui/src/minard/utils/useMousePos.ts
```diff
@@ -1,18 +1,16 @@
-import {useState, useEffect, MutableRefObject} from 'react'
+import {useState, useEffect} from 'react'
 
-export const useMousePos = (
-  ref: MutableRefObject<Element>
-): [number, number] => {
+export const useMousePos = (el: Element): [number, number] => {
```
bthesorceror: Small nit. I would prefer this to return an object, simply because an array of two numbers doesn't convey much information to the user of the function.

chnn: seems good, I've updated it

PR #11919, opened 2019-02-14
Generic import overlay

### LINKED ISSUE #11864: Create generic import overlay
as a user of the UI, when I click on an `import` functionality, 
I see a generic import overlay
which is configured to be able to take JSON verification and instantiation functions as arguments, so that it may be customized for different template types

screen shot 2019-02-13 at 11 43 23 am

ebb-tide:
Closes #11864

### DIFF
--- a/ui/src/shared/components/ImportOverlay.tsx
+++ b/ui/src/shared/components/ImportOverlay.tsx
@@ -8,15 +8,16 @@ import {
   OverlayBody,
   OverlayContainer,
   OverlayHeading,
+  OverlayFooter,
   Radio,
 } from 'src/clockface'
+import {Button, ComponentColor} from '@influxdata/clockface'
 
 // Constants
-import {
-  dashboardImportFailed,
-  dataWriteFailed,
-  dataWritten,
-} from 'src/shared/copy/notifications'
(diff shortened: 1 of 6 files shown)


ui/src/shared/components/ImportOverlay.tsx
```diff
@@ -0,0 +1,183 @@
...
+    if (selectedImportOption === ImportOption.Upload) {
+      return (
+        <DragAndDrop
+          submitText="Upload"
+          handleSubmit={this.handleSetImportContent}
+          submitOnDrop={true}
+          onCancel={this.clearImportContent}
+        />
+      )
+    } else if (selectedImportOption === ImportOption.Paste) {
```
121watts: a plain old `if` will do

ui/src/shared/components/ImportOverlay.tsx
```diff
@@ -0,0 +1,183 @@
...
+      return (
+        <TextArea
+          value={importContent}
+          onChange={this.handleSetImportContent}
+        />
+      )
+    }
+  }
+
+  private get submitButton(): JSX.Element {
```
121watts: This could be a component :)

Issue #11893, opened 2019-02-14
Add check in the influx setup command for a password that has at least 8 characters


goller:

Zyqsempai:
I can take this one if no one minds;)

goller:
Hey @Zyqsempai, go for it!  Here is a starting place for it:

https://github.com/influxdata/influxdb/blob/master/cmd/influx/setup.go#L192

PR #11934, opened 2019-02-15
Added password length check

Zyqsempai:
Closes https://github.com/influxdata/influxdb/issues/11893

Added check that password length is more then `8` on SetUp step.

  - [ ] Rebased/mergeable
  - [ ] Tests pass


Zyqsempai:
cc @knz

PR #11938, opened 2019-02-15
fix(ui): add ordering to org resources

### LINKED ISSUE #11891: All list views should be alphabetical by name
noticed this on the buckets page. let's update the component to default to sort by name

image

OfTheDelmer:
Closes #11891

Adds ordering to labels and org resources

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/organizations/containers/OrganizationView.tsx
+++ b/ui/src/organizations/containers/OrganizationView.tsx
@@ -96,7 +96,7 @@ class OrganizationView extends PureComponent<Props> {
                 url="tasks_tab"
                 title="Tasks"
               >
-                <GetOrgResources<Task[]> organization={org} fetcher={getTasks}>
+                <GetOrgResources<Task> organization={org} fetcher={getTasks}>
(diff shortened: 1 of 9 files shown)


ui/src/organizations/containers/OrganizationView.tsx
```diff
@@ -196,6 +204,7 @@ class OrganizationView extends PureComponent<Props> {
                 <GetOrgResources<Variable[]>
                   organization={org}
                   fetcher={getVariables}
+                  orderBy={{keys: ['name']}}
```
ebb-tide: should orderBy be an optional prop with default= `{keys:['name]}`

OfTheDelmer: @ebb-tide I was wondering that after most places it was used benefited from a simple name ordering. It should just ignore ordering if the collection doesn't have that key. Hmm would it be worth typing the orderedBy.keys property names of the collection item?

PR #11939, opened 2019-02-15
feat(http): define external preprocessing options via http api

### LINKED ISSUE #11702: Add Options to HTTP Query API
Flux has a language feature called `options`. Options are the mechanism by which configuration can be injected into a Flux script. Options can be set during runtime and referenced anywhere in the main package scope. 

We should update the `/query` API endpoint to take a new `options` property that contains key value pairs for setting options on a query. Then we need to update the implementation to set those options at runtime. 

A couple of questions I have:

* Should options values be ast.Nodes or raw values and we define how to unmarshal them?
* Should the flux.Compile method support generic options as values.Value or should the http server inject the options into the AST of the Flux script?

I am leaning towards using values.Value for option, since this means they can be injected at runtime without having to do a special parse + inject + format workflow.

The use case that drives this feature:

The UI needs to provide context under which the Flux query executes, things like the dashboard time range, time interval and any other user defined template variables.

By passing options on the HTTP API the frontend does not need to do any manipulation of the Flux text itself, rather it can submit two bits of info: the query text, and the options context.

It is possible that tasks will also benefit from this feature.

jlapacik:
Closes #11702 

Define external query preprocessing options via an `extern` property in the request body of an http request to the Query endpoint. An `extern` property is of type object and is unmarshalled to a Flux `File` ast node on the backend.

  - [x] Rebased/mergeable
  - [x] Tests pass


http/query.go
```diff
@@ -259,17 +220,24 @@ func (r QueryRequest) proxyRequest(now func() time.Time) (*query.ProxyRequest, e
...
-		var err error
-		r.Spec, err = toSpec(r.AST, now)
+		pkg, err := flux.Parse(r.Query)
 		if err != nil {
 			return nil, err
 		}
-		compiler = lang.SpecCompiler{
-			Spec: r.Spec,
+		if r.Extern != nil {
+			pkg.Files = append([]*ast.File{r.Extern}, pkg.Files...)
```
nathanielc: Since order doesn't matter should we just append this instead of prepending?

nathanielc: Do we want to add the extern concept to the ASTCompiler itself, so that the appending of the file is normal and standard?

nathanielc: I have been playing around with this and I think it makes sense to leave it how it is. This way its more flexible. In the testing code I am writing I am likely going to append multiple ast.Files so exposing a single Extern ast.File is not that helpful. Exposing a list seems to specific, so I say lets leave it.

jlapacik: The interpreter still executes statements sequentially. Dependency reordering has not been implemented yet, which is why I am prepending this file for right now.

http/query.go
```diff
@@ -259,17 +220,24 @@ func (r QueryRequest) proxyRequest(now func() time.Time) (*query.ProxyRequest, e
...
+		}
+		compiler = lang.ASTCompiler{
+			AST: pkg,
+			Now: now,
+		}
+	} else if r.AST != nil {
+		if r.Extern != nil {
```
goller: I'd rather see Extern as part of the lang.ASTCompiler here.  That way the API does not need to know about code changes to the AST as much.

```go
compiler = lang.ASTCompiler{
			AST: r.AST,
			Extern: r.Extern,
			Now: now,
		}
```

The compiler would handle the nil checking and adding it to the appropriate structure

http/query.go
```diff
@@ -259,17 +220,24 @@ func (r QueryRequest) proxyRequest(now func() time.Time) (*query.ProxyRequest, e
...
-	} else if r.AST != nil {
-		var err error
-		r.Spec, err = toSpec(r.AST, now)
+		pkg, err := flux.Parse(r.Query)
 		if err != nil {
 			return nil, err
 		}
-		compiler = lang.SpecCompiler{
-			Spec: r.Spec,
+		if r.Extern != nil {
```
goller: Same as my ASTCompiler note... I'd rather see extern as part of the SpecCompiler to decrease datastructure knowledge in the HTTP API.

http/query_test.go
```diff
@@ -86,6 +108,19 @@ func TestQueryRequest_Validate(t *testing.T) {
 			},
 			wantErr: true,
 		},
+		{
+			name: "no extern with spec",
```
goller: from slack convo: perhaps "query cannot have both extern and spec"

http/query_test.go
```diff
@@ -482,8 +532,104 @@ func Test_decodeProxyQueryRequest(t *testing.T) {
...
+				Dialect: &csv.Dialect{
+					ResultEncoderConfig: csv.ResultEncoderConfig{
+						NoHeader:  false,
+						Delimiter: ',',
+					},
+				},
+			},
+		},
+		{
+			name: "valid post query request",
```
goller: perhaps describe it as "valid query including extern definition"

http/swagger.yml
```diff
@@ -4564,6 +4566,28 @@ components:
...
           $ref: "#/components/schemas/Dialect"
+    FileASTNode:
+      description: flux File AST node
+      type: object
+      properties:
+        type:
+          description: type of AST node
+          type: string
+          enum:
+            - File
```
goller: Is this enumeration validated on query (that it can only be `File`)?  I didn't see it... perhaps it is done in flux unmarshaling code?

http/swagger.yml
```diff
@@ -4564,6 +4566,28 @@ components:
...
+        package:
+          description: package declaration
+          type: object
```
goller: This object looks like it has these properties:

BaseNode
```json
{
    "location": {
        "file": "string",
        "start": {
            "line": 1,
            "column": 1
        },
    },
    "errors": [
        {"msg": "string"}
    ]
}
```
and

File

```json
        {
            "name": "string",
            "package": [
                {"name": "string"}
            ],
            "imports": [
                {
                    "as": {"name": "string"},
                    "path": {"value": "string"}
                }
            ],
            "body": [
                // oneOf BadStatement, VariableAssignment, MemberAssignment, ExpressionStatement, etc
            ]
        }
```

Would you define the name field and fill in the package, imports and body objects?  Best would be to create more definitions that can be referred to here.

http/swagger.yml
```diff
@@ -4564,6 +4566,28 @@ components:
           type: string
         dialect:
           $ref: "#/components/schemas/Dialect"
+    FileASTNode:
+      description: flux File AST node
```
goller: I'd use the comment from the Flux code here:
https://github.com/influxdata/flux/blob/master/ast/ast.go#L186

Would you add descriptions for package, imports, name, and body using the godoc comments?

http/swagger.yml
jlapacik: Yes this is handled in the unmarshaling code.

http/swagger.yml
jlapacik: BaseNode is only used by the Flux parser to record source code location information. It should never be specified as part of this API which is why I didn't include it. But I can add the other fields.

jlapacik:
@goller I added the Flux AST to the swagger file. Could you please give this another review?

http/query.go
```diff
@@ -259,18 +223,27 @@ func (r QueryRequest) proxyRequest(now func() time.Time) (*query.ProxyRequest, e
...
+			c.PrependFile(r.Extern)
 		}
+		compiler = c
+	} else if r.AST != nil {
+		c := lang.ASTCompiler{
+			AST: r.AST,
+			Now: now,
+		}
+		if r.Extern != nil {
+			c.PrependFile(r.Extern)
```
goller: do all compilers now have PrependFile?  If so, perhaps this could be done after all flavors of the compiler have been constructed.  Decreasing the number of if statements is the goal.

jlapacik: @goller `PrependFile` is unique to this compiler (ASTCompiler). This seemed like the best place for this check since these extra statements are optional in terms of the HTTP API.

http/swagger.yml
```diff
@@ -4711,6 +4713,396 @@ components:
           type: string
         dialect:
           $ref: "#/components/schemas/Dialect"
+    File:
```
nathanielc: I think we want to namespace all of these definitions to be `fluxast` or something. According to the OpenAPI spec the `components` can be extended to contain our own namespace.

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#componentsObject 

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#specificationExtensions

So maybe we should put these into a components field `x-fluxast` instead of `schemas`. But before we do that we should check that the code generators will still generate the structures if we do. 

There is a doc on how do use the code generators, I am digging it up. Once I find it I'll link it here.

bthesorceror:
Looks like trying to generate our javascript client from this updated swagger is causing an error

```
-attribute components.schemas.CallExpression.properites is unexpected
```

http/swagger.yml
jlapacik: @nathanielc after doing some experimenting I found that the package used to validate the swagger specification (https://github.com/getkin/kin-openapi) does not accept such extensions.

nathanielc: Sounds good, lets leave it as it is for now then.

PR #11944, opened 2019-02-15
Enforce single org for series key reads

stuartcarnie:

These changes modify the parameters to a series cursor request and render the intent explicitly; series keys returned by the cursor must never cross organization boundaries.

* `SeriesCursorRequest` takes a single `Name` parameter, specifying the 16-byte tsdb encoded org and bucket ID
* `seriesCursor` ensures the first 8 bytes of the measurement name returned from the series index matches the current organization ID

**Blocked by #12010**

### DIFF
--- a/storage/series_cursor.go
+++ b/storage/series_cursor.go
@@ -4,8 +4,8 @@ import (
 	"bytes"
 	"errors"
 	"sort"
-	"sync"
 
+	"github.com/influxdata/influxdb"
 	"github.com/influxdata/influxdb/models"
 	"github.com/influxdata/influxdb/tsdb"
 	"github.com/influxdata/influxdb/tsdb/tsi1"
@@ -24,20 +24,19 @@ type SeriesCursor interface {
 
 type SeriesCursorRequest struct {
 	// Name contains the tsdb encoded org and bucket ID
-	Name [16]byte
+	Name [influxdb.IDLength]byte
 }
 
 // seriesCursor is an implementation of SeriesCursor over an tsi1.Index.
 type seriesCursor struct {
-	once  sync.Once
 	index *tsi1.Index
 	sfile *tsdb.SeriesFile
-	mitr  tsdb.MeasurementIterator
-	orgID []byte
+	name  [influxdb.IDLength]byte
 	keys  [][]byte
 	ofs   int
 	row   SeriesCursorRow
 	cond  influxql.Expr
+	init  bool
 }
 
 type SeriesCursorRow struct {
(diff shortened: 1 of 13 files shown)


storage/series_cursor.go
```diff
@@ -98,6 +103,9 @@ func (cur *seriesCursor) Next() (*SeriesCursorRow, error) {
 		}
 
 		cur.row.Name, cur.row.Tags = tsdb.ParseSeriesKey(cur.keys[cur.ofs])
+		if !bytes.HasPrefix(cur.row.Name, cur.orgID) {
```
e-dard: I think we should either change the `orgID` type to something stronger, or if we leave `orgID` as a `[]byte` then maybe `len(cur.row.Name) < 8 || !bytes.Equal(cur.row.Name[:8], cur.orgID)`.

storage/series_cursor_test.go
```diff
@@ -0,0 +1,38 @@
...
+		return tsdb.AppendSeriesKey(nil, name[:], nil)
+	}
+
+	nm := tsdb.EncodeName(0x0f0f, 0xb0b0)
+	cur := &seriesCursor{
+		keys: [][]byte{
+			makeKey(0x0f0f, 0xb0b0),
+			makeKey(0xffff, 0xb0b0),
+		},
+		orgID: nm[:8],
```
e-dard: Can we change `[:8]` to `[:influxdb.OrgIDLength]` or something? I just want to make it simpler to change the id scheme if we need to.

storage/series_cursor.go
```diff
@@ -11,20 +12,28 @@ import (
...
+	Name [16]byte
 }
 
 // seriesCursor is an implementation of SeriesCursor over an tsi1.Index.
 type seriesCursor struct {
 	once  sync.Once
 	index *tsi1.Index
+	sfile *tsdb.SeriesFile
 	mitr  tsdb.MeasurementIterator
+	orgID []byte
```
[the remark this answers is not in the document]
stuartcarnie: We could assert `len(orgID) == influxdb.IDBytesLen` or similar in `newSeriesCursor` to strengthen this precondition

storage/series_cursor.go
stuartcarnie: `bytes.HasPrefix` is implemented as 

```go
len(name) >= len(orgID) && Equal(name[0:len(prefix)], orgID)
```

So by adding an assertion in `newSeriesCursor`, such as `len(orgID) == influxdb.IDBytesLen`, we could ensure this invariant.

storage/series_cursor_test.go
stuartcarnie: Yeah, sounds good; I'll use that constant in the `newSeriesCursor` to validate too

stuartcarnie:
@e-dard, cleaned up the code a bit more:

* replaced `orgID` with `name [influxdb.IDLength]byte` and use that to test the prefix with `name[:influxdb.OrgIDLength]`
* removed all the measurement iterator stuff, as it was redundant

That end-to-end test you shared does exercise this new code and passes, however, there is no easy way to test the negative case, besides explicitly adding ways to break the invariants in the series and index logic.

PR #11958, opened 2019-02-16
refactor(task): remove explicit org from run/log lookup

### LINKED ISSUE #11807: task: infer orgID from taskID when listing runs
There is currently a mismatch between the swagger specification and the actual implementation for `GET /tasks/{taskID}/runs`. 

The swagger says nothing about `org` or `orgID`, which is correct.

The current code for the endpoint expects `org` or `orgID` to be passed explicitly, which is incorrect. It should look up the task to figure out the right organization ID to use.

mark-rushakoff:
Task ID is now a required value on run and log filters. It was
effectively required by all implementations before anyway, so now those
types reflect that requirement.

Organization ID was removed from those same fields. The TaskService
looks up the organization ID via the task in cases where we need it at a
lower layer.

Closes #11807.

### DIFF
--- a/cmd/influx/task.go
+++ b/cmd/influx/task.go
@@ -378,7 +378,6 @@ func taskDeleteF(cmd *cobra.Command, args []string) error {
 type TaskLogFindFlags struct {
 	taskID string
 	runID  string
-	orgID  string
 }
 
 var taskLogFindFlags TaskLogFindFlags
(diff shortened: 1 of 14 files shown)


bednar:
Hi @mark-rushakoff

Can you please look at #11960? 
The current implementation return 500 instead 404.

Thx

PR #11959, opened 2019-02-16
chore: add .editorconfig and enable in CI

### LINKED ISSUE #11895: Proposal: add .editorconfig to enforce style on non-source files
After [some discussion on a separate PR](https://github.com/influxdata/influxdb/pull/11876#pullrequestreview-203833845) regarding the EOL at the end of the swagger.yml file, it would probably be a better use of time to automate style enforcement for YAML and other non-source files.

As far as I'm aware, [EditorConfig](https://editorconfig.org/) is as close to a standard tool as we'll get there. Many editors support it already, and many other editors support EditorConfig through a plugin.

For that particular case of the trailing newline, it looks like the config would be as simple as:

```ini
[*.yaml]
insert_final_newline: true
```

There are [standalone implementations](https://github.com/editorconfig/editorconfig/wiki/FAQ#are-there-any-tools-to-check-orand-reformat-existing-files-against-editorconfig-rules) that we could use as a CI step to enforce the rules outside of an editor.

mark-rushakoff:
Closes #11895.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+# EditorConfig helps us maintain consistent formatting on non-source files.
+# Visit https://editorconfig.org/ for details on how to configure your editor to respect these settings.
+
+# This is the terminal .editorconfig in this repository.
+root = true
+
+[*.yml]
+insert_final_newline = true
+trim_trailing_whitespace = true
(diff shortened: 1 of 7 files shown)


mark-rushakoff:
I intentionally let this fail CI once so anyone can see what the failure looks like: https://circleci.com/gh/influxdata/influxdb/26574

mark-rushakoff:
Requesting review from those who left +1s on #11895.

PR #11973, opened 2019-02-19
fix(ui): add label from resource labels editor

### LINKED ISSUE #11624: Add ability to create a new label while adding labels to a resource.
Add ability to create a new label while adding labels to a resource. Labels should be able to be created on the fly instead of only on the config page.

OfTheDelmer:
Closes #11624

Add resource label creation form when label is not present. Tune label filtering to not aggressively match labels not in the list to more easily enable user to create a new label. Add a random color button instead of a color dropdown which does not work well in the label list dropdown.

  - [ ] Rebased/mergeable
  - [ ] Tests pass
kapture 2019-02-21 at 7 29 57

### DIFF
--- a/ui/src/configuration/components/RandomLabelColor.tsx
+++ b/ui/src/configuration/components/RandomLabelColor.tsx
@@ -1,0 +1,46 @@
+import React, {Component} from 'react'
+import _ from 'lodash'
+
+import {Button} from '@influxdata/clockface'
+
+import {PRESET_LABEL_COLORS} from 'src/configuration/constants/LabelColors'
+
+// Styles
+import 'src/configuration/components/RandomLabelColor.scss'
+
+interface Props {
+  colorHex: string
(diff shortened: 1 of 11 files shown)


ui/src/configuration/components/RandomLabelColor.tsx
```diff
@@ -0,0 +1,45 @@
...
+import 'src/configuration/components/RandomLabelColor.scss'
+
+interface Props {
+  colorHex: string
+  onClick: (newRandomHex: string) => void
+}
+
+export default class RandomLabelColorButton extends Component<Props> {
+  public componentDidMount() {
+    this.handleClick()
```
chnn: Using the mount of the component to eventually provide data from higher up in the render tree feels a bit funky. Can the color be generated for the first time outside of this component and then passed in?

mavarius:
Closes #11978

PR #11990, opened 2019-02-19
feat(ui): Add custom time range option in time range dropdown

### LINKED ISSUE #11692: Bring over the calendar selection from 1.7.x (no Rome)
as a query builder
with admin
when i select the time rage for my query
then i would like to be able to select an arbitrary start and end time using a calendar

we will need to ensure the range function of the query uses start and stop time if selected

ischolten:
Closes #11692

Add a '
`Custom Time Range` option to the timerange dropdown that opens a calendar selection to choose the time range
Change the clickOutside component to listen to `mousedown` rather than `click`
Add react-datepicker as dependency
add stop to queries built by the query builder that default to `now()` if a custom time is not selected

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/dateRangePicker/DateRangePicker.tsx
+++ b/ui/src/shared/components/dateRangePicker/DateRangePicker.tsx
@@ -1,0 +1,136 @@
+// Libraries
+import React, {PureComponent, createRef, CSSProperties} from 'react'
+
+// Components
+import DatePicker from 'src/shared/components/dateRangePicker/DatePicker'
+import {ClickOutside} from 'src/shared/components/ClickOutside'
+
+// Styles
+import 'src/shared/components/dateRangePicker/DateRangePicker.scss'
+
+// Types
+import {TimeRange} from 'src/types'
+import {Button, ComponentColor, ComponentSize} from '@influxdata/clockface'
+
+interface Props {
+  timeRange: TimeRange
+  onSetTimeRange: (timeRange: TimeRange) => void
+  position?: {top: number; right: number}
+  onClose: () => void
+}
+
+interface State {
+  lower: string
+  upper: string
+  bottomPosition?: number
+  topPosition?: number
+}
+
+const PICKER_HEIGHT = 410
+const HORIZONTAL_PADDING = 2
+const VERTICAL_PADDING = 15
+
+class DateRangePicker extends PureComponent<Props, State> {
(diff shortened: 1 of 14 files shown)


ui/src/shared/components/dateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,132 @@
...
+          />
+        </div>
+      </ClickOutside>
+    )
+  }
+
+  private get stylePosition(): CSSProperties {
+    const {position} = this.props
+    const {bottomPosition, topPosition} = this.state
+    const pickerHeight = 410
```
chnn: _nit_ Want to extract this to a named screaming snake case constant at the top of the file?

ui/src/shared/components/dateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+    if (topPosition) {
+      return {
+        top: '14px',
+        right: `${right + 2}px`,
+      }
+    }
+
+    const bottomPx =
+      (bottomPosition || window.innerHeight - top - 15) - pickerHeight / 2
```
chnn: What's the `15` for?

ui/src/shared/components/dateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,132 @@
...
+    return {
+      bottom: `${bottomPx}px`,
+      right: `${right + 2}px`,
```
chnn: What's the `2` for?

ui/src/shared/components/dateRangePicker/DateRangePicker.tsx
ischolten: can do

ui/src/shared/components/dateRangePicker/DateRangePicker.tsx
ischolten: some additional padding

ui/src/shared/components/dateRangePicker/DateRangePicker.tsx
ischolten: some additional padding; i can make into a constant

PR #11994, opened 2019-02-19
fix(ui): add guard for hasKey check

### LINKED ISSUE #11993: Org page tabs spins indefinitely if no resources exist
If no variables exist page spins indefinitely. I would expect a page load completed status or no resources found page. 

kapture 2019-02-19 at 14 16 31

OfTheDelmer:
Fixes checking key in primitive or null types

Closes #11993 

Empty resources and primitive data types could cause the `in` operator to error and not load resources.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/organizations/components/GetOrgResources.tsx
+++ b/ui/src/organizations/components/GetOrgResources.tsx
@@ -97,10 +97,10 @@ export default class GetOrgResources<T> extends PureComponent<
     resources: T[],
     key: string | number | symbol
   ): key is keyof T {
-    const resource = resources[0]
+    const resource = _.get(resources, '0', null)
     // gaurd against null and primitive types
     const isObject = !!resource && typeof resource === 'object'
 
-    return isObject && key in resource
+    return isObject && _.hasIn(resource, key)
   }
 }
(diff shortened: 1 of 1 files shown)


ui/src/organizations/components/GetOrgResources.tsx
```diff
@@ -97,6 +97,10 @@ export default class GetOrgResources<T> extends PureComponent<
     resources: T[],
     key: string | number | symbol
   ): key is keyof T {
-    return key in resources[0]
+    const resource = resources[0]
```
ebb-tide: maybe also `resource = get(resources, '0', null)` ? just in case resources comes back undefined or not an array?

ui/src/organizations/components/GetOrgResources.tsx
```diff
@@ -97,6 +97,10 @@ export default class GetOrgResources<T> extends PureComponent<
...
+    // gaurd against null and primitive types
+    const isObject = !!resource && typeof resource === 'object'
+
+    return isObject && key in resource
```
ebb-tide: an alternative could be to use lodash [_.keys](https://lodash.com/docs/4.17.11#keys) function to get the array of keys and check that `keys.includes(key)`

OfTheDelmer: Yeah I was going to add that but didn't like it's behavior with strings

```
_.keys("abc") // =>  ["0", "1", "3"]
```

PR #12011, opened 2019-02-20
Add ability to update a variable

### LINKED ISSUE #11844: Create a way to update a template variable

AlirieGray:
Closes #11844

This PR adds the ability to edit a variable's name and query from the Variables tab of the orgs page. It also includes notifications for when an update succeeds or fails.

  - [x] Rebased/mergeable
  - [x] Tests pass
  - [x] CHANGELOG.md updated


ui/src/organizations/components/UpdateVariableOverlay.tsx
```diff
@@ -0,0 +1,146 @@
...
+  nameErrorMessage: string
+  nameInputStatus: ComponentStatus
+}
+
+export default class UpdateVariableOverlay extends PureComponent<Props, State> {
+  constructor(props) {
+    super(props)
+
+    const {variable} = this.props
+    const script = _.get(variable, 'arguments.values.query')
```
ischolten: does there need to be a default in this get? like `""`?

ui/src/organizations/components/UpdateVariableOverlay.tsx
```diff
@@ -0,0 +1,146 @@
...
+            <OverlayFooter>
+              <Button
+                text="Cancel"
+                color={ComponentColor.Danger}
+                onClick={onCloseOverlay}
+              />
+              <Button
+                text="Submit"
+                type={ButtonType.Submit}
+                onClick={this.handleSubmit}
```
ischolten: Forms have an `onSubmit` prop. Since this Button is already type submit, you could remove `this.handleSubmit` to the onSubmit prop on the form and then get default browser behavior here

ui/src/organizations/components/UpdateVariableOverlay.tsx
```diff
@@ -0,0 +1,146 @@
...
+    )
+  }
+
+  private get isVariableValid(): boolean {
+    const {variable, script} = this.state
+
+    return !!variable.name && !!script
+  }
+
+  private handleSubmit = (e): void => {
```
ischolten: could you type annotate e?

ui/src/organizations/components/VariableList.tsx
```diff
@@ -35,13 +53,45 @@ class VariableList extends PureComponent<Props> {
...
+  private get variable(): Variable {
+    return this.props.variables.find(v => v.id === this.state.variableID)
+  }
+
+  private get isVariableOverlayVisible(): boolean {
+    return this.state.variableOverlayState === OverlayState.Open
+  }
+
+  private handleCloseOverlay = () => {
+    this.setState({variableOverlayState: OverlayState.Closed})
```
ischolten: should this also reset variableID to `null`?

ui/src/organizations/components/Variables.tsx
```diff
@@ -165,4 +168,22 @@ export default class Variables extends PureComponent<Props, State> {
...
+    variable: Partial<Variable>
+  ): Promise<void> => {
+    const {notify, onChange} = this.props
+
+    try {
+      const updatedVariable = await client.variables.update(
+        variable.id,
+        variable
+      )
+      notify(updateVariableSuccess(updatedVariable.name))
```
ischolten: nice!

ui/src/organizations/components/VariableRow.tsx
```diff
@@ -11,10 +11,13 @@ import {
 
 // Types
 import {Variable} from '@influxdata/influx'
+import EditableName from 'src/shared/components/EditableName'
 
 interface Props {
   variable: Variable
   onDeleteVariable: (variable: Variable) => void
+  onUpdateVariable: (variable: Partial<Variable>) => void
+  onEditVariable: (variable: Variable) => void
```
ischolten: perhaps there is a better name distinction between these two props? both of those names seem like they would do the same thing.

ui/src/organizations/components/VariableList.tsx
AlirieGray: good catch!

Issue #11781, opened 2019-02-08
CLI: task find runs only accepts org-id but should accept org name


russorat:

tkyocum:
Currently blocked by https://github.com/influxdata/influxdb/pull/12022.

PR #12022, opened 2019-02-20
feat(tasks): look up tasks by orgname

### LINKED ISSUE #11822: get tasks rest api : `org` query param is being ignored.
Whether or not you are using a valid org id in `org` query param, the get tasks request will all ways return all of the tasks.

- Create tasks for different orgs using  `influx` command:
`./influx task create 'option task = {name:"new org",every:1h} from(bucket:"b_1") |> range(start: -1h)' --org-id "0364bd08a56fa000" -t "OuNJYaSH5Vqdy3Jd8aZ7ID9C3VKynUket8EmzLR3ILNcJpvbjRZFkt1DDnFNoAs1BYEGUuG-zIPUvxYSxVidzA=="`

- Get all of the tasks using `influx` command:
`./influx task find -t "OuNJYaSH5Vqdy3Jd8aZ7ID9C3VKynUket8EmzLR3ILNcJpvbjRZFkt1DDnFNoAs1BYEGUuG-zIPUvxYSxVidzA=="`

The above command will return all of the tasks for all of the orgs:
```
ID			Name		OrganizationID		Organization	Status	Every	Cron
0364b59ea340e000	n_1		0364b4fc0d6fa000	test_1		active	1h0m0s	
0364b5e11600e000	n_1		0364b4fc0d6fa000	test_1		active	1h0m0s	
0364b5f13400e000	n_2		0364b4fc0d6fa000	test_1		active	1h0m0s	
0364b5f7ea00e000	n_3		0364b4fc0d6fa000	test_1		active	1h0m0s	
0364bd37ca40e000	new org		0364bd08a56fa000	test_2		active	1h0m0s	
0364bd40b980e000	new org 1	0364bd08a56fa000	test_2		active	1h0m0s	
0364bd4679c0e000	new org 2	0364bd08a56fa000	test_2		active	1h0m0s	
```

- Use rest api to get tasks for one of the organizations:
`curl -GET "http://localhost:9999/api/v2/tasks?org='0364bd08a56fa000'" -H "Authorization:Token OuNJYaSH5Vqdy3Jd8aZ7ID9C3VKynUket8EmzLR3ILNcJpvbjRZFkt1DDnFNoAs1BYEGUuG-zIPUvxYSxVidzA=="`

It will return tasks for all of the orgs.

docmerlin:
closes #11822 and #11781

This allows lookups by org name and makes the api more consistent.

### DIFF
--- a/http/task_service.go
+++ b/http/task_service.go
@@ -17,6 +17,7 @@ import (
 	platform "github.com/influxdata/influxdb"
 	"github.com/influxdata/influxdb/authorizer"
 	pcontext "github.com/influxdata/influxdb/context"
+	"github.com/influxdata/influxdb/kv"
 	"github.com/influxdata/influxdb/query"
 	"github.com/influxdata/influxdb/task/backend"
 	"github.com/influxdata/influxdb/task/options"
@@ -259,7 +260,7 @@ func newRunsResponse(rs []*platform.Run, taskID platform.ID) runsResponse {
 func (h *TaskHandler) handleGetTasks(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 
-	req, err := decodeGetTasksRequest(ctx, r)
+	req, err := decodeGetTasksRequest(ctx, r, h.OrganizationService)
 	if err != nil {
 		err = &platform.Error{
 			Err:  err,
@@ -323,7 +324,7 @@ type getTasksRequest struct {
 	filter platform.TaskFilter
 }
 
-func decodeGetTasksRequest(ctx context.Context, r *http.Request) (*getTasksRequest, error) {
+func decodeGetTasksRequest(ctx context.Context, r *http.Request, orgs platform.OrganizationService) (*getTasksRequest, error) {
 	qp := r.URL.Query()
 	req := &getTasksRequest{}
 
@@ -335,12 +336,26 @@ func decodeGetTasksRequest(ctx context.Context, r *http.Request) (*getTasksReque
 		req.filter.After = id
 	}
 
-	if orgID := qp.Get("organization"); orgID != "" {
-		id, err := platform.IDFromString(orgID)
+	if orgName := qp.Get("org"); orgName != "" {
+		o, err := orgs.FindOrganization(ctx, platform.OrganizationFilter{Name: &orgName})
+		if err != nil {
+			if pErr, ok := err.(*platform.Error); ok && pErr != nil {
+				if kv.IsNotFound(err) || pErr.Code == platform.EUnauthorized {
+					return nil, &platform.Error{
+						Err: errors.New("org not found or unauthorized"),
+						Msg: "org " + orgName + " not found or unauthorized",
+					}
+				}
+			}
+			return nil, err
+		}
+		req.filter.Organization = &o.ID
+	} else if oid := qp.Get("orgID"); oid != "" {
+		orgID, err := platform.IDFromString(oid)
 		if err != nil {
 			return nil, err
 		}
-		req.filter.Organization = id
+		req.filter.Organization = orgID
 	}
 
 	if userID := qp.Get("user"); userID != "" {
@@ -1409,7 +1424,7 @@ func (t TaskService) FindTasks(ctx context.Context, filter platform.TaskFilter) 
 		val.Add("after", filter.After.String())
 	}
 	if filter.Organization != nil {
-		val.Add("organization", filter.Organization.String())
+		val.Add("orgID", filter.Organization.String())
 	}
 	if filter.User != nil {
 		val.Add("user", filter.User.String())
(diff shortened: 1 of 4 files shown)


http/task_service.go
```diff
@@ -334,12 +334,18 @@ func decodeGetTasksRequest(ctx context.Context, r *http.Request) (*getTasksReque
 		req.filter.After = id
 	}
 
-	if orgID := qp.Get("organization"); orgID != "" {
-		id, err := platform.IDFromString(orgID)
+	if orgName := qp.Get("org"); orgName != "" {
+		o, err := orgs.FindOrganization(ctx, platform.OrganizationFilter{Name: &orgName})
```
goller: I think there needs to be some way to obfuscate the error returned here, so, it is not possible to fish for organization names.

Perhaps the API should return an empty list of tasks of the org name is unknown?

http/task_service.go
```diff
@@ -1184,7 +1190,7 @@ func (t TaskService) FindTasks(ctx context.Context, filter platform.TaskFilter)
 		val.Add("after", filter.After.String())
 	}
 	if filter.Organization != nil {
-		val.Add("organization", filter.Organization.String())
```
goller: Perhaps add name to TaskFilter allowing one to search for tasks with name?  (could be follow-on work)

task_test.go
```diff
@@ -64,4 +64,21 @@ from(bucket:"x")
 			t.Fail()
 		}
 	})
+	t.Run("test add new option", func(t *testing.T) {
```
goller: Is this test related to the new code?  Perhaps add a test using org name if it doesnt exist?

http/task_service.go
mark-rushakoff: I think in other places (possibly only in some wip), I've used the phrase "unauthorized or unknown" to keep it vague.

task_test.go
mark-rushakoff: Same question, this test doesn't seem related to this PR.

http/task_service.go
```diff
@@ -322,7 +322,7 @@ type getTasksRequest struct {
 	filter platform.TaskFilter
 }
 
-func decodeGetTasksRequest(ctx context.Context, r *http.Request) (*getTasksRequest, error) {
+func decodeGetTasksRequest(ctx context.Context, r *http.Request, orgs platform.OrganizationService) (*getTasksRequest, error) {
```
mark-rushakoff: @lyondhill you moved some org service calls into the platform adapter in #12004. Does it make sense to move this into the platform adapter too, or should it stay here?

lyondhill: TaskService `FindTasks` currently doesn't support org name. It can be moved to the platform adapter if we wanted to update the FindTasks method in the interface. 
Out of this context The FindTasks probably needs to be updated to use `FindOptions` sometime. When that happens would be a good time to move this out of the http handler.

task_test.go
docmerlin: Whoops, is related to another PR, the files got mixed.

http/task_service.go
docmerlin: Yah, I think something like that would be better one in follow on.

task_test.go
docmerlin: Its still a useful test, and it didn't get included in the merged pr it was supposed to, so I'll leave it in?

http/task_service_test.go
```diff
@@ -38,6 +39,20 @@ func NewMockTaskBackend(t *testing.T) *TaskBackend {
...
 				org := &platform.Organization{}
 				if filter.Name != nil {
+					if *filter.Name == "non-existant-org" {
+						return nil, &platform.Error{
+							Err:  errors.New("org not found or unauthorized"),
+							Msg:  "org " + *filter.Name + " not found or unauthorized",
+							Code: platform.ENotFound,
+						}
+					}
+					if *filter.Name == "unauthorized" {
```
mark-rushakoff: It doesn't look like the `*filter.Name == "unauthorized"` path is exercised. It should be exercised or removed.

task_test.go
```diff
@@ -64,4 +64,21 @@ from(bucket:"x")
...
+			t.Fatal(err)
+		}
+
+		op, err := options.FromScript(*tu.Flux)
+		if err != nil {
+			t.Error(err)
+		}
+		if op.Offset != 30*time.Second {
+			t.Logf("expected every to be 10s but was %s", op.Every)
+			t.Fail()
```
mark-rushakoff: `t.Logf; t.Fail` can be just `t.Fatalf`. And it looks like the message should read 30s.

http/task_service.go
docmerlin: Here's the issue for it:
https://github.com/influxdata/influxdb/issues/12170

PR #12025, opened 2019-02-20
Clockface/spinners

mavarius:
Use Spinners in Clockface Library

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/me/components/account/__snapshots__/Tokens.test.tsx.snap
```diff
@@ -56,14 +56,16 @@ exports[`Account rendering renders! 1`] = `
   <ResourceFetcher
     fetcher={[Function]}
   >
-    <SpinnerContainer
+    <t
```
chnn: Maybe we should turn off uglification in the clockface builds so that we get more descriptive snapshots.

mavarius: Ya, this was something we also discussed with @bthesorceror, we may just.

PR #12068, opened 2019-02-21
Allow tasks to skip catchup

### LINKED ISSUE #11365: task: skip task catch-up on startup, by default
If the (OSS) system was shut down, then presumably there weren't any writes, so tasks can't do anything to catch up.

I believe the solution here is to update LatestCompleted to `time.Now()` for every task, before starting the scheduler.

lyondhill:
Closes #11365

### DIFF
--- a/task/backend/bolt/bolt.go
+++ b/task/backend/bolt/bolt.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"time"
 
 	bolt "github.com/coreos/bbolt"
@@ -46,6 +47,8 @@ type Store struct {
 	db     *bolt.DB
 	bucket []byte
 	idGen  platform.IDGenerator
+
+	minLatestCompleted int64
 }
 
 const basePath = "/tasks/v1/"
@@ -59,8 +62,14 @@ var (
 	runIDs       = []byte(basePath + "run_ids")
 )
 
+// Option is a optional configuration for the store.
+type Option func(*Store)
+
+// NoCatchUp allows you to skip any task that was supposed to run during down time.
+func NoCatchUp(st *Store) { st.minLatestCompleted = time.Now().Unix() }
+
 // New gives us a new Store based on "github.com/coreos/bbolt"
-func New(db *bolt.DB, rootBucket string) (*Store, error) {
+func New(db *bolt.DB, rootBucket string, opts ...Option) (*Store, error) {
 	if db.IsReadOnly() {
 		return nil, ErrDBReadOnly
 	}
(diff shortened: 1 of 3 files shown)


task/backend/bolt/bolt.go
```diff
@@ -87,7 +95,11 @@ func New(db *bolt.DB, rootBucket string) (*Store, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &Store{db: db, bucket: bucket, idGen: snowflake.NewDefaultIDGenerator()}, nil
+	st := &Store{db: db, bucket: bucket, idGen: snowflake.NewDefaultIDGenerator()}
```
mark-rushakoff: Seems like `latestReplace` should be initialized to `math.MinInt64`. Otherwise it's plausible that a task that was created to run over historical data will "fast-forward" to zero time.

task/backend/bolt/bolt.go
```diff
@@ -46,6 +46,8 @@ type Store struct {
 	db     *bolt.DB
 	bucket []byte
 	idGen  platform.IDGenerator
+
+	latestReplace int64
```
mark-rushakoff: What do you think about calling this `minLatestCompleted`?

ui/package-lock.json
```diff
@@ -6074,8 +6074,7 @@
           "version": "2.1.1",
           "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
           "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
-          "dev": true,
-          "optional": true
+          "dev": true
```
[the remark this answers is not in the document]
lyondhill: I did a `make` and it changed that. not sure why.

task/backend/bolt/bolt.go
```diff
@@ -434,6 +447,10 @@ func (s *Store) FindTaskMetaByID(ctx context.Context, id platform.ID) (*backend.
 		return nil, err
 	}
 
+	if stm.LatestCompleted < s.minLatestCompleted {
```
mark-rushakoff: Looking at it now, my concern with this approach is that we're returning something different from what's in the task store.

I think it will be okay for now, so I'm approving. But we'll need to keep in mind that this comes at a risk of difficulty debugging later if what we're exposing to the higher layers doesn't match exactly what's stored on disk.

If we run into debugging that, and if it's that painful, we can refactor to the other discussed approach of modifying every task meta record at startup.

lyondhill: I agree, It could be difficult to debug. We will have to keep an eye on it.

PR #12071, opened 2019-02-21
Some preperation for delete safety

zeebo:
The tsm1 package had a lot of exported fields that were unused making it hard to reason about changes to locks. The exported fields were only used in the tests and only because the tests were of the `tsm1_test` style. Just make it internal and do some cleanups.

The tsi1 package had some methods that were only used to be api compatible with inmem, which is no longer a thing. Baleeted.

The cache ring had code to support multiple sizes of rings that was totally dead and broken. Additionally, it had races that could cause data loss during compaction enable/disable. Removes all that support, and fixes up benchmarks/tests.

### DIFF
--- a/pkg/lifecycle/lifecycle.go
+++ b/pkg/lifecycle/lifecycle.go
@@ -1,38 +1,38 @@
-package storage
+package lifecycle
 
 import (
 	"io"
 )
 
-// opener is something that can be opened and closed.
-type opener interface {
+// Resource is something that can be opened and closed.
+type Resource interface {
 	Open() error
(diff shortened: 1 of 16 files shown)


e-dard:
@zeebo worth fixing this up for next week?

e-dard:
@zeebo shall we get this fixed up?

zeebo:
I think this has rotted quite a bit, and many of the changes have been made already in other PRs. I'm going to close it out just so that it's not in anyone's backlog, evaluate if the branch is worth saving, and push up anything from it.

PR #12072, opened 2019-02-21
Update admin to show error if password is not 8 characters

### LINKED ISSUE #12028: ensure setup succeeds before continuing to the next step of onboarding

Palakp41:
Closes #12028 

Update admin page to not proceed to next step if there is an error with passwords that are not 8 characters long

  - [x] Rebased/mergeable
  - [x] Tests pass

kapture 2019-02-21 at 11 43 50

### DIFF
--- a/ui/src/onboarding/actions/index.ts
+++ b/ui/src/onboarding/actions/index.ts
@@ -1,3 +1,6 @@
+// Libraries
+import _ from 'lodash'
+
 // Constants
 import {StepStatus} from 'src/clockface/constants/wizard'
 import {SetupSuccess, SetupError} from 'src/shared/copy/notifications'
@@ -5,6 +8,7 @@ import {SetupSuccess, SetupError} from 'src/shared/copy/notifications'
 // Actions
 import {notify} from 'src/shared/actions/notifications'
 
+// APIs
 import {client} from 'src/utils/api'
 
 // Types
@@ -62,7 +66,9 @@ export const setBucketID = (bucketID: string): SetBucketID => ({
   payload: {bucketID},
 })
 
-export const setupAdmin = (params: ISetupParams) => async dispatch => {
+export const setupAdmin = (params: ISetupParams) => async (
+  dispatch
+): Promise<boolean> => {
   try {
     dispatch(setSetupParams(params))
     const response = await client.setup.create(params)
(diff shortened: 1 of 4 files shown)


ui/src/onboarding/actions/index.ts
```diff
@@ -1,10 +1,12 @@
...
+import _ from 'lodash'
+
 // Constants
 import {StepStatus} from 'src/clockface/constants/wizard'
 import {SetupSuccess, SetupError} from 'src/shared/copy/notifications'
 
 // Actions
 import {notify} from 'src/shared/actions/notifications'
-
 import {client} from 'src/utils/api'
```
ebb-tide: nit- maybe this should go up under `//Libraries` ?

ui/src/onboarding/actions/index.ts
```diff
@@ -62,7 +64,7 @@ export const setBucketID = (bucketID: string): SetBucketID => ({
   payload: {bucketID},
 })
 
-export const setupAdmin = (params: ISetupParams) => async dispatch => {
+export const setupAdmin = (params: ISetupParams) => async (dispatch): Promise<boolean | null> => {
```
ebb-tide: Each type in typescript also contains null as a subtype, so you can just say boolean here and null should be accepted

ui/src/shared/copy/notifications.ts
```diff
@@ -99,10 +99,10 @@ export const SetupSuccess: Notification = {
   message: 'Initial user details have been successfully set',
 }
 
-export const SetupError: Notification = {
+export const SetupError= (message: string): Notification => ({
   ...defaultErrorNotification,
```
ebb-tide: yeah as we discussed, I think it would be better to notify some of our copy "could not do x" followed by message from back end.

ui/src/onboarding/actions/index.ts
Palakp41: I added it under // APIs :)

ui/src/shared/copy/notifications.ts
Palakp41: Got it!

ui/src/onboarding/actions/index.ts
Palakp41: Ok, I fixed it to be boolean type

PR #12080, opened 2019-02-21
Fix query bug resulting from missing org ID

chnn:
The `/api/v2/query` endpoint requires an organization or organization ID query parameter.

Previously, there existed a bug in the API where if the organization parameters were left off, the API would use the first organization created in the backend, rather than returning a 400 error.

We built the entire UI on top of this bug, but it has now been fixed. So in every location where we use the `/api/v2/query` endpoint, we need to supply an organization.

This commit updates all such locations to use the first organization present in our Redux store as the organization parameter, thus roughly reproducing the behavior of the load bearing bug.

This is just a quick fix. Long term, we will want to think about what organization queries should run under and build an appropriate UI around that design.

### DIFF
--- a/ui/src/dataLoaders/components/verifyStep/DataListening.tsx
+++ b/ui/src/dataLoaders/components/verifyStep/DataListening.tsx
@@ -1,9 +1,11 @@
 // Libraries
 import React, {PureComponent} from 'react'
+import {connect} from 'react-redux'
 import _ from 'lodash'
 
 // Apis
 import {executeQuery} from 'src/shared/apis/v2/query'
+import {getActiveOrg} from 'src/organizations/selectors'
 
 // Components
 import {ErrorHandling} from 'src/shared/decorators/errors'
@@ -18,11 +20,18 @@ import ConnectionInformation, {
 } from 'src/dataLoaders/components/verifyStep/ConnectionInformation'
 
 // Types
+import {AppState, Organization} from 'src/types/v2'
 import {InfluxLanguage} from 'src/types/v2/dashboards'
 
-export interface Props {
+interface OwnProps {
   bucket: string
 }
+
+interface StateProps {
+  activeOrg: Organization
+}
+
+type Props = OwnProps & StateProps
 
 interface State {
   loading: LoadingState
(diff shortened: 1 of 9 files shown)


chnn:
I'm fixing the test failures now...

ui/src/dataLoaders/components/verifyStep/DataListening.tsx
```diff
@@ -123,6 +132,7 @@ class DataListening extends PureComponent<Props, State> {
     try {
       const response = await executeQuery(
         '/api/v2/query',
+        activeOrg.id,
```
ischolten: if no active org is found, should we have a  different error here?

ui/src/timeMachine/actions/queryBuilder.ts
```diff
@@ -203,11 +204,13 @@ export const loadBuckets = () => async (
   dispatch: Dispatch<Action>,
   getState: GetState
 ) => {
+  const queryURL = getActiveQuerySource(getState()).links.query
+  const orgID = getActiveOrg(getState()).id
```
ischolten: should we handle the org not being found here?

ui/src/timeMachine/actions/queryBuilder.ts
```diff
@@ -247,6 +250,7 @@ export const loadTagSelector = (index: number) => async (
 
   const tagPredicates = tags.slice(0, index)
   const queryURL = getActiveQuerySource(getState()).links.query
+  const orgID = getActiveOrg(getState()).id
```
ischolten: same here

ui/src/timeMachine/actions/queryBuilder.ts
```diff
@@ -299,6 +304,7 @@ const loadTagSelectorValues = (index: number) => async (
   const {buckets, tags} = getActiveQuery(getState()).builderConfig
   const tagPredicates = tags.slice(0, index)
   const queryURL = getActiveQuerySource(getState()).links.query
+  const orgID = getActiveOrg(getState()).id
```
ischolten: and here

PR #12101, opened 2019-02-22
fix(swagger): quote keys named "y"

mark-rushakoff:
The YAML parser used by the go-openapi libraries treats an unquoted y as
a boolean key, which will lead to a difficult-to-understand parser
error:

> types don't match expect map key string or int got: bool

See also https://yaml.org/type/bool.html.


goller:
Nice...lgtm

PR #12104, opened 2019-02-22
Check error when writing token file during setup

juliusv:
Signed-off-by: Julius Volz <julius.volz@gmail.com>

This tripped me up because the CLI does not actually create the directory it wants to write the token into, and at the same time the CLI always uses the user's home dir, while the server allows configuring the storage path (which is what I did, which is why the `~/.influxdbv2` didn't exist for the CLI to find). So that should probably also be fixed (create the dir if it doesn't exist) and configurable.


nathanielc:
@juliusv Thanks!

PR #12121, opened 2019-02-22
Add import dropdown to empty task and dashboard states

ebb-tide:


ebb-tide:
accomplished by : https://github.com/influxdata/influxdb/pull/12661

PR #12128, opened 2019-02-22
Clean up histogram implementation

chnn:
This PR contains some refactoring of the histogram implementation

- Improve the typing of various vis objects
- Use a different `Table` shape for plot layers that colocates data and data types
- Renames the "aesthetics" field of a plot layer to "mappings", since "aesthetics" is difficult to say, difficult to spell, and is less clear than "mappings"

### DIFF
--- a/ui/src/minard/components/Histogram.tsx
+++ b/ui/src/minard/components/Histogram.tsx
@@ -47,7 +47,7 @@ export const Histogram: SFC<Props> = ({
   const layer = useLayer(
     env,
     () => {
-      const [table, aesthetics] = bin(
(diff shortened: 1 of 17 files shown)


ui/src/minard/components/Histogram.tsx
```diff
@@ -47,7 +47,7 @@ export const Histogram: SFC<Props> = ({
   const layer = useLayer(
     env,
     () => {
-      const [table, aesthetics] = bin(
+      const [table, mappings] = bin(
```
alexpaxton: Hell yeah

Issue #12117, opened 2019-02-22
Improve validation of field types in WAL


e-dard:
Since we moved the WAL out of the engine (https://github.com/influxdata/influxdb/pull/11364) the writing client is no longer protected by the error handling in the Cache:

https://github.com/influxdata/influxdb/blob/b9d5a7bec2529d7696afb016646f6eb0b4139c2c/tsdb/tsm1/cache.go#L74-L86

As such, the user will now see the following error if they have two series with different field types within a single batch:

```
edd@Alameda:~|⇒  influx write --bucket-id 0371a3adc1755001  @/tmp/writes.txt
Error: Failed to write data: http/handleWrite: incorrect value found in int64 slice: value.FloatValue.
Usage:
  influx write line protocol or @/path/to/points.txt [flags]
```

We should improve the error message by possibly using the same error message as the `Cache`. We can't remove the type checking from the cache because we need it for instances where a WAL is not in use.

zhulongcheng:
@e-dard 
Is it possible to add type checking to `PointsToValues`?
https://github.com/influxdata/influxdb/blob/b9d5a7bec2529d7696afb016646f6eb0b4139c2c/tsdb/tsm1/value.go#L53-L108

zeebo:
In the case of conflicting points in the same batch, I think it's going to be tricky to maintain the same semantics. It used to first validate all of the points against the index, then would write into the cache, and if any of those points were problematic, it would skip writing to the WAL (source: https://github.com/influxdata/influxdb/blob/1.8/tsdb/engine/tsm1/engine.go#L1348-L1356). That doesn't seem possible to maintain with the WAL first.

Given the above challenges, I propose that, as @zhulongcheng suggested, we report errors when converting into the values map, drop the conflicting points where first point sets the type for the rest of the points in the batch, and report a partial write error.

Since that's fairly small to implement, I'm going to start work on it and we can change directions if the proposal isn't sufficient.

PR #12135, opened 2019-02-24
storage: unify field type check error

zhulongcheng:
Updates #12117

  - [ ] Rebased/mergeable
  - [ ] Tests pass


zeebo:
#12349 addresses this by ensuring that the values map we pass to the WAL is never invalid. If the values map ever is invalid, that's an internal error, so user presentation is less important than precise information.

Thanks for all your awesome PRs!

PR #12136, opened 2019-02-25
spike(http): add swagger validator middleware

mark-rushakoff:
This adds an HTTP middleware that performs validation on both requests
and responses to the API.

If running without the assets build tag, it looks in a few reasonable
places on disk to detect swagger.yml. Otherwise, it expects the
swagger.yml Asset to exist.

The validation from kin-openapi/openapi3filter has a lot of whitespace,
so in this commit, instead of logging it which would escape the newlines
and tabs, we just printf the message, which looks like:

```
SWAGGER RESPONSE VALIDATION ERROR: Error at "/authorizations/0/permissions/16/resource/type":JSON value is not one of the allowed values
Schema:
  {
    "enum": [
      "authorizations",
      "buckets",
      "dashboards",
      "orgs",
      "sources",
      "tasks",
      "telegrafs",
      "users"
    ],
    "type": "string"
  }

Value:
  "variables"
```

An enhancement here would be to do something smarter with the message --
perhaps embed it into an HTTP header or trailer.

This commit does not include tests for the swaggervalidator middleware.

It also has to record the request and response in a separate
bytes.Buffer, which is created from scratch on every request. There is
likely room for improvement on how we deal with those allocations.


mark-rushakoff:
I just rebased this on master. It didn't require any extra changes.

The couple pieces I'm specifically unsure about:

- Whether just printing to stdout is acceptable here. Logging is generally preferred, but logfmt does not play well with messages that contain a lot of whitespace like these validation errors. Maybe shoving the error in an HTTP header would be nice since that can be readily inspected from client code. We can set a normal header for request validation, but I think response validation would have to be an HTTP trailer.
- Where the middleware should be instantiated. Right now it's happening inside the PlatformHandler which doesn't quite feel right. Maybe that's okay for now.
- Under what circumstances the middleware should be short-circuited. It's probably desirable to be on by default in development (i.e. without the `assets` build tag) and off by default in production (i.e. _with_ the `assets` build tag), but configurable either way.

So far this has been very helpful in clearly calling out swagger discrepancies. See issues I've recently filed with the `area/api-docs` label: https://github.com/influxdata/influxdb/issues?q=is%3Aissue+is%3Aopen+label%3Aarea%2Fapi-docs

PR #12144, opened 2019-02-25
Let Plot components size themselves

chnn:
Let `Plot` components measure the dimensions of their containing element automatically. I think this is a slightly nicer public API.


ui/src/shared/components/Histogram.tsx
```diff
@@ -99,36 +98,24 @@ const Histogram: SFC<Props> = ({
...
-                position={position}
-                tooltip={HistogramTooltip}
-                colors={colorHexes}
-              />
-            )}
-          </MinardPlot>
-        )
-      }}
-    </AutoSizer>
+    <MinardPlot
```
chnn: This is an example of how the public API get's simplified.

PR #12153, opened 2019-02-25
Update tasks runs to use the api instead of dummy data

### LINKED ISSUE #12120: Update TasksRuns to use runs api instead of dummy data

Palakp41:
Closes #12120 

Connected tasks runs page with redux to use the state to get runs and display runs on the page

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/tasks/actions/v2/index.ts
+++ b/ui/src/tasks/actions/v2/index.ts
@@ -50,6 +50,7 @@ export type Action =
   | SetAllTaskOptions
   | AddTaskLabels
   | RemoveTaskLabels
+  | SetRuns
 
 type GetStateFunc = () => AppState
 interface Task extends TaskAPI {
@@ -148,6 +149,13 @@ export interface RemoveTaskLabels {
   }
 }
 
+export interface SetRuns {
+  type: 'SET_RUNS'
+  payload: {
+    runs: Run[]
+  }
+}
+
 export const setTaskOption = (taskOption: {
   key: TaskOptionKeys
   value: string
@@ -198,6 +206,11 @@ export const setShowInactive = (): SetShowInactive => ({
 export const setDropdownOrgID = (dropdownOrgID: string): SetDropdownOrgID => ({
   type: 'SET_DROPDOWN_ORG_ID',
   payload: {dropdownOrgID},
+})
+
+export const setRuns = (runs: Run[]): SetRuns => ({
+  type: 'SET_RUNS',
+  payload: {runs},
 })
 
 const addTaskLabels = (taskID: string, labels: Label[]): AddTaskLabels => ({
@@ -468,9 +481,14 @@ export const getErrorMessage = (e: any) => {
   return message
 }
 
-export const getRuns = async (taskID: string): Promise<Run[]> => {
-  const runs = await client.tasks.getRunsByTaskID(taskID)
-  return runs
+export const getRuns = (taskID: string) => async (dispatch): Promise<void> => {
+  try {
+    const runs = await client.tasks.getRunsByTaskID(taskID)
+
+    dispatch(setRuns(runs))
+  } catch (error) {
+    console.error(error)
+  }
 }
 
 export const runTask = (taskID: string) => async dispatch => {
--- a/ui/src/tasks/actions/v2/index.ts
+++ b/ui/src/tasks/actions/v2/index.ts
@@ -20,6 +20,7 @@ import {
   taskCloneSuccess,
   taskCloneFailed,
   taskRunSuccess,
+  taskGetFailed,
 } from 'src/shared/copy/v2/notifications'
 
 // Types
@@ -34,6 +35,7 @@ import {
   TaskOptions,
   TaskSchedule,
 } from 'src/utils/taskOptionsToFluxScript'
+import {RemoteDataState} from '@influxdata/clockface'
 
 export type Action =
   | SetNewScript
@@ -153,6 +155,7 @@ export interface SetRuns {
   type: 'SET_RUNS'
   payload: {
     runs: Run[]
+    runStatus: RemoteDataState
   }
 }
 
@@ -208,9 +211,9 @@ export const setDropdownOrgID = (dropdownOrgID: string): SetDropdownOrgID => ({
   payload: {dropdownOrgID},
 })
 
-export const setRuns = (runs: Run[]): SetRuns => ({
+export const setRuns = (runs: Run[], runStatus: RemoteDataState): SetRuns => ({
   type: 'SET_RUNS',
-  payload: {runs},
+  payload: {runs, runStatus},
 })
 
 const addTaskLabels = (taskID: string, labels: Label[]): AddTaskLabels => ({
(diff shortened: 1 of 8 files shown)


ui/src/tasks/actions/v2/index.ts
```diff
@@ -468,9 +481,14 @@ export const getErrorMessage = (e: any) => {
...
-export const getRuns = async (taskID: string): Promise<Run[]> => {
-  const runs = await client.tasks.getRunsByTaskID(taskID)
-  return runs
+export const getRuns = (taskID: string) => async (dispatch): Promise<void> => {
+  try {
+    const runs = await client.tasks.getRunsByTaskID(taskID)
+
+    dispatch(setRuns(runs))
+  } catch (error) {
+    console.error(error)
```
chnn: Can this display a notification instead?

ui/src/tasks/components/TaskRunsPage.tsx
```diff
@@ -1,25 +1,47 @@
...
 
-// DummyData
-import {taskRuns} from 'src/tasks/dummyData'
+// Actions
+import {getRuns} from 'src/tasks/actions/v2'
+import {Run} from '@influxdata/influx'
+import {RemoteDataState} from 'src/types'
 
-interface Props {
+interface PassedInProps {
```
chnn: We've been calling this `OwnProps` elsewhere. Could you use that name instead for consistency?

ui/src/tasks/components/TaskRunsPage.tsx
```diff
@@ -1,25 +1,47 @@
...
   params: {id: string}
+}
+
+interface ConnectedDispatchProps {
```
chnn:
```suggestion
interface DispatchProps {
```

ui/src/tasks/components/TaskRunsPage.tsx
```diff
@@ -1,25 +1,47 @@
...
+  getRuns: typeof getRuns
+}
+
+interface ConnectedStateProps {
```
chnn:
```suggestion
interface StateProps {
```

ui/src/tasks/components/TaskRunsPage.tsx
```diff
@@ -1,25 +1,47 @@
...
-const dummyData = taskRuns
+interface State {
+  loading: RemoteDataState
+}
+
+type Props = PassedInProps & ConnectedDispatchProps & ConnectedStateProps
 
-class TaskRunsPage extends PureComponent<Props & WithRouterProps> {
+class TaskRunsPage extends PureComponent<Props, State> {
+  constructor(props: Props) {
```
chnn: _nit_ I think a slightly more declarative way of doing this is to use a instance property instead of the `constructor`:

```
class TaskRunsPage extends PureComponent<Props, State> {
  public state: State = {loading: RemoteDataState.NotStarted}

  // ...
```

ui/src/tasks/components/TaskRunsPage.tsx
```diff
@@ -1,25 +1,47 @@
...
+    super(props)
+
+    this.state = {loading: RemoteDataState.NotStarted}
+  }
   public render() {
-    const {params} = this.props
+    const {params, runs} = this.props
+
+    if (this.state.loading !== RemoteDataState.Done) {
+      return <div>Runs not loaded</div>
```
chnn: Could we use a clockface spinner here?

ui/src/tasks/components/TaskRunsPage.tsx
```diff
@@ -1,25 +1,47 @@
...
+  getRuns: typeof getRuns
+}
+
+interface ConnectedStateProps {
   runs: Run[]
 }
 
-const dummyData = taskRuns
+interface State {
+  loading: RemoteDataState
```
chnn: It seems like storing the runs in redux but storing the loading state for the runs outside of redux could potentially cause us problems in the future. In particular,

- It splits state across the application, which makes thinking about state more difficult (and thinking about state is the hardest part of building a web app)
- The state can get out of sync if the `TaskRunsPage` is unmounted and remounted for some reason. A react component should be able to be unmounted/remounted/rerendered whenever and still accurately reflect the state of the application

I know we do this pattern elsewhere a lot, but it's something I would like to change.

What do you think about adding a `runsStatus: RemoteDataState` property to the task reducer and storing the state there instead?

Then the `setRuns` action could look something like this instead:

```
export interface SetRuns {
  type: 'SET_RUNS'
  payload: {
    runs: Run[]
    runsStatus: RemoteDataState
  }
}
```

ui/src/tasks/components/TaskRunsPage.tsx
```diff
@@ -1,25 +1,63 @@
...
+  constructor(props: Props) {
+    super(props)
+
+    this.state = {loading: RemoteDataState.NotStarted}
+  }
   public render() {
-    const {params} = this.props
+    const {params, runs} = this.props
+
+    if (this.state.loading !== RemoteDataState.Done) {
```
chnn: Could we display a spinner while this is loading? The lack of feedback seems disorienting.

ui/src/tasks/actions/v2/index.ts
Palakp41: Yup! I added a notification:)

ui/src/tasks/components/TaskRunsPage.tsx
Palakp41: Yes, I added it!

ui/src/tasks/components/TaskRunsPage.tsx
Palakp41: Yes, that makes more sense! Thanks for reviewing :)

ui/src/tasks/components/TaskRunsPage.tsx
Palakp41: Added the spinner

ui/src/tasks/components/TaskRunsPage.tsx
```diff
@@ -31,13 +45,29 @@ class TaskRunsPage extends PureComponent<Props & WithRouterProps> {
...
+              <TaskRunsList taskID={params.id} runs={runs} />
             </div>
           </Page.Contents>
         </Page>
-      </>
+      </SpinnerContainer>
     )
   }
+
+  public async componentDidMount() {
```
chnn: Since this isn't setting state anymore, this can be
```
  public componentDidMount() {
    this.props.getRuns(this.props.params.id)
  }
```
I think that makes it a little more clear that it is issuing a side effect that takes place outside of the component.

ui/src/tasks/actions/v2/index.ts
```diff
@@ -468,9 +484,15 @@ export const getErrorMessage = (e: any) => {
   return message
 }
 
-export const getRuns = async (taskID: string): Promise<Run[]> => {
-  const runs = await client.tasks.getRunsByTaskID(taskID)
-  return runs
+export const getRuns = (taskID: string) => async (dispatch): Promise<void> => {
+  try {
+    const runs = await client.tasks.getRunsByTaskID(taskID)
```
chnn:
```suggestion
    dispatch(setRuns(runs, RemoteDataState.Loading))
    
    const runs = await client.tasks.getRunsByTaskID(taskID)
```

ui/src/tasks/components/TaskRunsPage.tsx
Palakp41: Okay, I removed it

ui/src/tasks/actions/v2/index.ts
Palakp41: I added loading state before get runs api call

PR #12177, opened 2019-02-26
docs(http): add members/owners links

zhulongcheng:
Closes #

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/scraper_service_test.go
+++ b/http/scraper_service_test.go
@@ -132,7 +132,9 @@ func TestService_handleGetScraperTargets(t *testing.T) {
 						  "links": {
 						    "bucket": "/api/v2/buckets/0000000000000212",
 						    "organization": "/api/v2/orgs/0000000000000211",
-						    "self": "/api/v2/scrapers/0000000000000111"
+						    "self": "/api/v2/scrapers/0000000000000111",
+						    "members": "/api/v2/scrapers/0000000000000111/members",
+						    "owners": "/api/v2/scrapers/0000000000000111/owners"
 						  }
 						},
 						{
@@ -147,7 +149,9 @@ func TestService_handleGetScraperTargets(t *testing.T) {
 						  "links": {
 						    "bucket": "/api/v2/buckets/0000000000000212",
 						    "organization": "/api/v2/orgs/0000000000000211",
-						    "self": "/api/v2/scrapers/0000000000000222"
+						    "self": "/api/v2/scrapers/0000000000000222",
+						    "members": "/api/v2/scrapers/0000000000000222/members",
+						    "owners": "/api/v2/scrapers/0000000000000222/owners"
 						  }
                         }
 					  ]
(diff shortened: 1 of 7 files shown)


http/scraper_service_test.go
```diff
@@ -675,7 +683,9 @@ func TestService_handlePatchScraperTarget(t *testing.T) {
 		              "links":{
 		                "bucket": "/api/v2/buckets/0000000000000212",
 		                "organization": "/api/v2/orgs/0000000000000211",
-		                "self":"/api/v2/scrapers/%[1]s"
+		                "self":"/api/v2/scrapers/%[1]s",
```
goller: These IDs look odd; would you check why they aren't `targetOneID` ?

zhulongcheng:
```
body: fmt.Sprintf(`..., "self": "/api/v2/scrapers/%[1]s",...`, targetOneIDString) 
```

zhulongcheng: The `%[1]s` is to maintain the same code style as the previous code.
I think It is clearer to remove `fmt.Sprintf` and replace   `%[1]s` with `targetOneID`.

goller: Ok. would you remove the [1] from the string?  That looks like a bug to me.

zhulongcheng: Actually, that is not a bug. Just different code styles.
```
fmt.Sprintf(
	`
{
  "id": "%[1]s",
  "name": "name",
  "type": "prometheus",
  "url": "www.example.url",
  "organization": "org1",
  "orgID": "0000000000000211",
  "bucket": "bucket1",
  "bucketID": "0000000000000212",
  "links": {
    "bucket": "/api/v2/buckets/0000000000000212",
    "organization": "/api/v2/orgs/0000000000000211",
    "self": "/api/v2/scrapers/%[1]s",
    "members": "/api/v2/scrapers/%[1]s/members",
    "owners": "/api/v2/scrapers/%[1]s/owners"
  }
}`,
	targetOneIDString,
)
```
All `%[1]s` will be replace with `targetOneIDString`
(https://golang.org/pkg/fmt  #Explicit argument indexes)

If remove `[1]` from the string, that will like:
```
fmt.Sprintf(
	`
{
  "id": "%s",
  "name": "name",
  "type": "prometheus",
  "url": "www.example.url",
  "organization": "org1",
  "orgID": "0000000000000211",
  "bucket": "bucket1",
  "bucketID": "0000000000000212",
  "links": {
    "bucket": "/api/v2/buckets/0000000000000212",
    "organization": "/api/v2/orgs/0000000000000211",
    "self": "/api/v2/scrapers/%s",
    "members": "/api/v2/scrapers/%s/members",
    "owners": "/api/v2/scrapers/%s/owners"
  }
}`,
	targetOneIDString, targetOneIDString, targetOneIDString, targetOneIDString
)
```
Which style do you prefer?  @goller

zhulongcheng: `[1]` removed.

leodido: Not @goller is not a bug. It was intended to use only one parameter in the `fmt.Sprintf` tail rather than repeting the same variable 3 times (I was the author of this choice months ago).

leodido: Personally I prefer the original one.

Just because I do not like function calls (and also definitions) with so many parameters ...

leodido: BTW, since I was the only to use this style and all the other APIs impl. do not use it I'm okay with conforming to the common style the team is using.

goller:
Awesome again :)

Thanks!

PR #12206, opened 2019-02-26
storage: fix series file retention in some iterators

zeebo:
These iterators should hold on to the resources they need to do their job. The callers don't/can't for them.

Also, the SeriesIterator and SeriesIteratorAdapter are dead.


zeebo:
Fixed by #12207.

PR #12207, opened 2019-02-26
storage: fix problems with keeping resources alive

zeebo:
This commit adds the `pkg/lifecycle.Resource` type to help manage opening, closing, and leasing out references to some resource. A resource cannot be closed until all acquired references have been released. If the `debug_ref` tag is enabled, all resource acquisitions keep track of the stack trace that created them and have a finalizer associated with them to print on stderr if they are leaked. It also registers a handler on SIGUSR2 to dump all of the currently live
resources.

Having resources tracked in a uniform way with a data type allows us to do more sophisticated tracking with the `debug_ref` tag, as well. For example, we could show references and panic the process if a resource cannot be closed within a certain time frame, or attempt to figure out the DAG of resource ownership dynamically. If we go all in, we can require that callers present a reference in order to use some APIs to statically ensure that you're holding a reference.

This commit also fixes some issues around resources, correctness during error scenarios, reporting of errors, idempotency of open and close, tracking of memory for some data structures, resource leaks in tests, and out of order dependency closes in tests.

### DIFF
--- a/pkg/lifecycle/resource.go
+++ b/pkg/lifecycle/resource.go
@@ -1,0 +1,94 @@
+package lifecycle
+
+import (
+	"sync"
+	"sync/atomic"
+	"unsafe"
+)
+
+// Resource keeps track of references and has some compile time debug hooks
+// to help diagnose leaks. It keeps track of if it is open or not and allows
+// blocking until all references are released.
+type Resource struct {
+	sem  sync.RWMutex
+	open bool
+}
+
+// Open waits for any outstanding references, of which there should be none
+// and marks the reference counter as open.
+func (res *Resource) Open() {
+	res.sem.Lock()
+	res.open = true
+	res.sem.Unlock()
+}
+
+// Close waits for any outstanding references and marks the reference counter
+// as closed, so that Acquire returns an error.
+func (res *Resource) Close() {
+	res.sem.Lock()
+	res.open = false
+	res.sem.Unlock()
+}
+
+// Opened returns true if the resource is currently open.
+func (res *Resource) Opened() bool {
+	res.sem.RLock()
+	open := res.open
+	res.sem.RUnlock()
+	return open
+}
+
+// Acquire returns a Reference used to keep alive some resource.
+func (res *Resource) Acquire() (*Reference, error) {
+	res.sem.RLock()
+	if !res.open {
+		res.sem.RUnlock()
+		return nil, resourceClosed()
+	}
+	// RLock intentionally left open.
+
+	return live.track(&Reference{
+		res: res,
+		id:  0, // required because staticcheck
+	}), nil
+}
+
+// Reference is an open reference for some resource.
+type Reference struct {
+	res *Resource
+	id  uint64
+}
+
+// Release causes the Reference to be freed. It is safe to call multiple times.
+func (r *Reference) Release() {
+	// Inline a sync.Once using the res pointer as the flag for if we have
+	// called unlock or not. This reduces the size of a ref.
+	addr := (*unsafe.Pointer)(unsafe.Pointer(&r.res))
+	old := atomic.LoadPointer(addr)
+	if old != nil && atomic.CompareAndSwapPointer(addr, old, nil) {
+		live.untrack(r)
+		(*Resource)(old).sem.RUnlock()
+	}
+}
+
+// Close makes a Reference an io.Closer. It is safe to call multiple times.
+func (r *Reference) Close() error {
+	r.Release()
+	return nil
+}
+
+// References is a helper to aggregate a group of references.
+type References []*Reference
+
+// Release releases all of the references. It is safe to call multiple times.
+func (rs References) Release() {
+	for _, r := range rs {
+		r.Release()
+	}
+}
+
+// Close makes references an io.Closer. It is safe to call multiple times.
+func (rs References) Close() error {
+	rs.Release()
+	return nil
+}
--- a/pkg/lifecycle/resource_debug_enabled.go
+++ b/pkg/lifecycle/resource_debug_enabled.go
@@ -1,0 +1,118 @@
+// +build debug_ref
+
+package lifecycle
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"os/signal"
+	"path/filepath"
+	"runtime"
+	"sync"
+	"syscall"
+)
+
+// When in debug mode, we associate each reference an id and with that id the
+// stack trace that created it. We can't directly refer to the reference here
+// because we also associate a finalizer to print to stderr if a reference is
+// leaked, including where it came from if possible.
+
+func init() {
+	// This goroutine will dump all live references and where they were created
+	// when SIGUSR2 is sent to the process.
+	go func() {
+		ch := make(chan os.Signal, 1)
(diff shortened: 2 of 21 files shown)


pkg/lifecycle/resource_debug_enabled.go
```diff
@@ -0,0 +1,115 @@
...
+	live: make(map[uint64][]uintptr),
+}
+
+// finishId informs the liveReferences that the id is no longer in use.
+func (l *liveReferences) untrack(r *Reference) {
+	l.mu.Lock()
+	if _, ok := l.live[r.id]; ok {
+		delete(l.live, r.id)
+		runtime.SetFinalizer(r, nil)
+	}
```
jacobmarble: Is it worth checking `!ok`?

pkg/lifecycle/resource_debug_enabled.go
```diff
@@ -0,0 +1,115 @@
...
+
+// withFinalizer associates a finalizer with the Reference that will cause it
+// to print a leak message if it is not closed before it is garbage collected.
+func (l *liveReferences) track(r *Reference) *Reference {
+	var buf [32]uintptr
+	pcs := append([]uintptr(nil), buf[:runtime.Callers(3, buf[:])]...)
+
+	l.mu.Lock()
+	r.id, l.id = l.id, l.id+1
+	l.live[r.id] = pcs
```
jacobmarble: Again, worth checking if the map entry already exists?

pkg/lifecycle/resource.go
```diff
@@ -0,0 +1,94 @@
...
+	sem  sync.RWMutex
+	open bool
+}
+
+// Open waits for any outstanding references, of which there should be none
+// and marks the reference counter as open.
+func (res *Resource) Open() {
+	res.sem.Lock()
+	res.open = true
+	res.sem.Unlock()
```
jacobmarble: It's great that these `Unlock()` calls are not deferred. I've been bitten by `defer`s building up over time in a busy process.

storage/series_cursor.go
```diff
@@ -63,16 +66,30 @@ func newSeriesCursor(req SeriesCursorRequest, index *tsi1.Index, sfile *tsdb.Ser
...
+		index:    index,
+		indexref: indexref,
+		sfile:    sfile,
+		sfileref: sfileref,
+		name:     req.Name,
+		cond:     cond,
 	}, nil
 }
 
 // Close closes the iterator.
```
jacobmarble: "safe to call multiple times" wouldn't hurt

tsdb/index.go
```diff
@@ -207,86 +136,31 @@ func (a SeriesIDIterators) Close() (err error) {
...
-			}
-		}
-		return &itr.point, nil
-	}
-}
-
 // filterUndeletedSeriesIDIterator returns all series which are not deleted.
 type filterUndeletedSeriesIDIterator struct {
 	sfile *SeriesFile
+	ref   *lifecycle.Reference
```
jacobmarble: For consistency, please rename `ref` to `sfileref` as in `seriesCursor`

tsdb/series_file.go
```diff
@@ -45,7 +46,9 @@ type SeriesFile struct {
 	defaultMetricLabels prometheus.Labels
 	metricsEnabled      bool
 
-	refs sync.RWMutex // RWMutex to track references to the SeriesFile that are in use.
+	// res keeps track of outstanding references to the series file.
+	resmu sync.Mutex
```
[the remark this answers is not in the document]
jacobmarble: Looks like the comment should say something about exclusivity, not reference tracking. It changed from a `RWMutex` to just a `Mutex` and is useful to protect multiple `Open()`s, for example.

tsdb/index.go
```diff
@@ -3,83 +3,12 @@ package tsdb
...
 	"fmt"
-	"sync"
 
-	"github.com/influxdata/influxdb/models"
+	"github.com/influxdata/influxdb/pkg/lifecycle"
 	"github.com/influxdata/influxdb/query"
 	"github.com/influxdata/influxql"
 )
 
-// SeriesElem represents a generic series element.
```
jacobmarble: I assume this was unused code. Please confirm.

tsdb/tsi1/file_set.go
```diff
@@ -2,38 +2,56 @@ package tsi1
...
-	levels       []CompactionLevel
 	sfile        *tsdb.SeriesFile
+	sfileref     *lifecycle.Reference
 	files        []File
+	filesref     lifecycle.References
 	manifestSize int64 // Size of the manifest file in bytes.
 }
 
 // NewFileSet returns a new instance of FileSet.
-func NewFileSet(levels []CompactionLevel, sfile *tsdb.SeriesFile, files []File) (*FileSet, error) {
```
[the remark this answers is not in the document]
benbjohnson: I think the idea was to allow us to change compaction level settings in future releases but retain compaction settings for existing indexes.

pkg/lifecycle/resource_debug_enabled.go
```diff
@@ -0,0 +1,115 @@
...
+		return
+	}
+
+	fmt.Fprintln(os.Stderr, "=====================================================")
+	fmt.Fprintln(os.Stderr, "=== Leaked a reference! Created from")
+	summarizeStack(os.Stderr, pcs)
+	fmt.Fprintln(os.Stderr, "=====================================================")
+}
+
+func init() {
```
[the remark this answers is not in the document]
zeebo: No problem.

tsdb/series_file.go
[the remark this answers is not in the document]
zeebo: Jacob is correct, and this code is wrong/misleading. It will be changed to be a mutex to just protect against concurrent Open and Close calls.

tsdb/tsi1/file_set.go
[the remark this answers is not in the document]
zeebo: Yeah there does seem to be support for that, but I was unable to find a reason for it to exist in the FileSet, and so I removed it.

pkg/lifecycle/resource_debug_enabled.go
zeebo: I think I might get rid of the checking in general since the parent ensures that it's only called once and on tracked references, and the protected calls are idempotent/no-ops on untracked references.

pkg/lifecycle/resource_debug_enabled.go
zeebo: That should be impossible due to the mutex and the increment right above it.

tsdb/index.go
zeebo: Yep, I could find no usages of it in any repo.

zeebo:
Retain and release bugs do manifest as a deadlock. In the debug mode, we have more leeway. I'd prefer to run in debug mode (maybe even a harsher one that panics on leaks) in as many situations as possible if performance allows.

jacobmarble:
> Retain and release bugs do manifest as a deadlock. In the debug mode, we have more leeway. I'd prefer to run in debug mode (maybe even a harsher one that panics on leaks) in as many situations as possible if performance allows.

+1 to panic on leak.

Issue #12162, opened 2019-02-25
Run query to show available values for a Variable


AlirieGray:

chnn:
Blocked pending answers to the following questions:

- When a user is using the data explorer, what variables do they see?
- When a user is viewing a dashboard, what variables do they see?
- When a user edits a cell on a dashboard, what variables do they see?
- If a user saves a cell to a dashboard that uses a variable that they were not using before does it appear in the variable toolbar for the dashboard? Or do they add the variable explicitly?
- If a user changes the value for a variable in a dropdown in the CEO, does it change the value selected on the dashboard?
- If a user can see every variable when opening the CEO, does that mean we load the values for every variable? What's the max number of variables that we expect a user would have access to, and would this be too slow?
- Once executed, should the query for a variable ever reexecute? Does this matter for variables with `now` etc in them? Should there be a way for a user to explicitly refresh a variable?
- Does a dashboard save variable value selections? If so, is it to the server, or to browser local storage?
- Can variables be used in tasks?

I've set up a meeting for tomorrow to resolve this.

PR #12209, opened 2019-02-27
Centralize variables state in Redux store

chnn:
Working towards #12162

Moves `Variable` state from local component state to the Redux store.

### DIFF
--- a/ui/src/variables/reducers/index.ts
+++ b/ui/src/variables/reducers/index.ts
@@ -1,0 +1,70 @@
+// Libraries
+import {produce} from 'immer'
+
+// Types
+import {RemoteDataState} from 'src/types'
+import {Action} from 'src/variables/actions'
+import {Variable} from '@influxdata/influx'
+
+const initialState = (): VariablesState => ({
+  status: RemoteDataState.NotStarted,
+  variables: {},
+})
+
+export interface VariablesState {
+  status: RemoteDataState // Loading status of the entire variables collection
+  variables: {
+    [variableID: string]: {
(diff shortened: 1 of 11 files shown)


ui/src/variables/reducers/index.ts
```diff
@@ -0,0 +1,70 @@
...
+          draftState.variables[id].status = status
+        }
+
+        return
+      }
+
+      case 'REMOVE_VARIABLE': {
+        const {id} = action.payload
+
+        delete draftState.variables[id]
```
AlirieGray: could we map over the variables and filter out the one with this ID, rather than using the delete?

ui/src/variables/reducers/index.ts
```diff
@@ -0,0 +1,70 @@
...
+        const {id, status, variable} = action.payload
+        const variableExists = !!draftState.variables[id]
+
+        if (variable || !variableExists) {
+          draftState.variables[id] = {variable, status}
+        } else {
+          draftState.variables[id].status = status
+        }
+
+        return
```
AlirieGray: Seems like this is directly modifying state, whereas we usually make a copy of the state and return it?

chnn: It's all wrapped in an [immer `produce` function](https://github.com/mweststrate/immer), so the state never does get modified.

ui/src/variables/reducers/index.ts
chnn: This also because the reducer is written in the immer style (see my other comment).

PR #12212, opened 2019-02-27
chore(ci):Fork PR tests

BondAnthony:
Closes #

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,7 +108,7 @@ test-js: node_modules
 	make -C ui test
 
 test-go:
-	$(GO_TEST) ./...
+	GO_TEST=env GOTRACEBACK=all GO111MODULE=on gotestsum --junitfile junit-test.xml
 
 test-integration: GO_TAGS=integration
 test-integration:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -108,7 +108,7 @@ test-js: node_modules
 	make -C ui test
 
 test-go:
-	$(GO_TEST) ./...
+	GO_TEST=env GOTRACEBACK=all GO111MODULE=on gotestsum --junitfile junit-test.xml
```
mark-rushakoff: I don't think this is quite the right change, for a couple reasons:

- `gotestsum` is not part of the go toolchain. We can make it part of this project, but we need team buy-in first.
- If there was a way to conditionally only run `gotestsum` when we know we're running in CircleCI, but to otherwise run `go test` like we were before, that would be acceptable today without team buy-in, IMO.
- There should be a comment with a link to https://circleci.com/blog/level-up-go-test-with-gotestsum/ so that a reader can immediately learn what `gotestsum` is and why it's useful on CircleCI.
- I'm pretty sure `GO_TEST=env` is literally setting the environment variable `GO_TEST` to the string `env`. AFAIK nothing inspects that environment variable. The old way actually invoked the definition of `GO_TEST` declared on line 23.

mark-rushakoff:
It would also be helpful if you pushed a temporary commit that included a deliberate test failure, and then include a link to that test failure so we can see what Circle does when a Go test fails. This could be as simple as:

```go
func TestFail(t *testing.T) {
  t.Fatal("forced failure")
}
```

PR #12221, opened 2019-02-27
test(e2e/de): shows an error syntactically incorrect query is submitted

### LINKED ISSUE #12192: test(de): submitting a synatctically incorrect query should give a valid error message to the user

121watts:
Closes #12192

### DIFF
--- a/ui/src/shared/components/EmptyGraphMessage.tsx
+++ b/ui/src/shared/components/EmptyGraphMessage.tsx
@@ -2,11 +2,25 @@ import React, {SFC} from 'react'
 
 interface Props {
   message: string
+  type?: string
 }
 
-const EmptyGraphMessage: SFC<Props> = ({message}) => {
+export enum GraphMessageType {
+  Empty = 'empty',
+  Error = 'error',
+  NoData = 'no-data',
+  RemoteDataState = 'remote-data-state',
+}
+
+const EmptyGraphMessage: SFC<Props> = ({
+  message,
+  type = GraphMessageType.Empty,
+}) => {
   return (
-    <div className="cell--view-empty">
+    <div
+      className="cell--view-empty"
+      data-testid={`empty-graph-message ${type}`}
+    >
       <h4>{message}</h4>
     </div>
   )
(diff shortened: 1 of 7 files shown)


ui/src/shared/components/EmptyGraphMessage.tsx
```diff
@@ -2,11 +2,25 @@ import React, {SFC} from 'react'
...
+export enum GraphMessageType {
+  Empty = 'empty',
+  Error = 'error',
+  NoData = 'no-data',
+  RemoteDataState = 'remote-data-state',
+}
+
+const EmptyGraphMessage: SFC<Props> = ({
+  message,
+  type = GraphMessageType.Empty,
```
chnn: seems like this in only used for testing, but is requiring a bunch of changes to the `EmptyQueryView`. could we assert on the actual message itself in tests instead and leave the app code as is?

121watts: I opted for this approach because a message we display to the user is more likely to change.

chnn: what if someone incorrectly passes the `type` option to the `EmptyGraphMessage` component? Then the tests could be a false positive or false negative, since they are testing an implementation detail of the app that only exists for tests

121watts: That's a risk for sure.  I think the greater risk is that the error message contents change.  However, I think a happy medium is searching for `error` within `EmptyGraphMessage` with a static test id.

chnn: I think this is getting back to our discussion about using `data-testid` vs. other data in the DOM for assertions. This feels to me like a concrete case where we _do_ want the test to fail if the message changes.

Like if I were to introduce the following bug, we would want the tests to fail:
```
    if (hasNoResults) {
      return (
        <EmptyGraphMessage message="Your query was syntactically invalid" />
      )
    }
```
With the data-testid stuff we are no longer asserting against the actual behavior of the app, but we are instead asserting against a special implementation that only exists in tests.

It's nbd either way so sorry to be so pedantic. I would be curious to hear what others think about this though.

Issue #12236, opened 2019-02-27
Labels Cleanup


alexpaxton:
Once #12232, #12233, #12234, and #12234  are completed there is some cleanup to do.

- [ ] Delete all clockface auxillary components (everything that is not the base `<Label />` component)

Issue #12234, opened 2019-02-27
Consolidate multiple label types for dashboards


alexpaxton:
- [ ] The single type should be `Label` as defined in `src/types/v2/labels` which has `properties` on it
- [ ] Labels presentational components should not use the Clockface `LabelType` and instead use `Label`

PR #12241, opened 2019-02-28
Components for handling inline label editing

### LINKED ISSUE #12232: Create components for editing labels inline

alexpaxton:
Closes #12232 

Adds a set of components for adding/removing/creating labels inline (instead of using an overlay) which are intended to replace the labels management overlay UX in the long term

Goal is to only rely on the base `Label` component from clockface so that all non-essential clockface label components can be removed (see #12236)

  - [x] Rebased/mergeable
  - [x] Tests pass


alexpaxton:
Waiting for the labels typing work in #12234

OfTheDelmer:
awaiting client type updates from #12243

ui/src/shared/components/inlineLabels/InlineLabelPopover.tsx
```diff
@@ -0,0 +1,120 @@
...
+      Math.max(highlightedIndex + direction, 0),
+      filteredLabels.length - 1
+    )
+
+    const adjacentID = filteredLabels[adjacentIndex].id
+
+    onUpdateSelectedItem(adjacentID)
+  }
+
+  private handleReFocusInput = (e: ChangeEvent<HTMLInputElement>): void => {
```
ischolten: nit: in `handleRefocusInput`, should the `f` in focus be lowercase

ui/src/shared/components/inlineLabels/InlineLabels.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  private handleDeleteLabel = (labelID: string): void => {
+    const {onRemoveLabel, selectedLabels} = this.props
+    const label = selectedLabels.find(label => label.id === labelID)
+
+    onRemoveLabel(label)
+  }
+
+  private handleCreateLabel = (name: string): void => {
+    console.log(name)
```
ischolten: can we remove the `console.log`?
should other logic be here?

ui/src/shared/components/inlineLabels/InlineLabelsEditor.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+// Utils
+import {validateLabelUniqueness} from 'src/configuration/utils/labels'
+
+// Styles
+import 'src/shared/components/inlineLabels/InlineLabelsEditor.scss'
+
+import {ErrorHandling} from 'src/shared/decorators/errors'
+
+enum OverlayState {
```
ischolten: there might (i think) be an overlay state enum already defined elsewhere that could be used

ui/src/shared/components/inlineLabels/InlineLabelsEditor.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+    onAddLabel(label)
+  }
+
+  private handleUpdateSelectedItem = (selectedItemID: string): void => {
+    this.setState({selectedItemID})
+  }
+
+  private handleShowPopover = () => {
+    const {availableLabels} = this
```
ischolten: nit: is this destructuring necessary since `this.availableLabels` is being called again in line 162

ui/src/shared/components/inlineLabels/InlineLabelsList.tsx
```diff
@@ -0,0 +1,108 @@
...
+  private get createNewLabelButton(): JSX.Element {
+    const {searchTerm, filteredLabels, onStartCreatingLabel} = this.props
+
+    if (!searchTerm) {
+      return null
+    }
+
+    const searchTermHasExactMatch = filteredLabels.reduce(
+      (acc: boolean, current: Label) => {
+        if (acc === true || current.name === searchTerm) {
```
ischolten: nit: this could be simplified to just `return acc === true || current.name === searchTerm`

ui/src/shared/components/inlineLabels/InlineLabelsEditor.tsx
alexpaxton: I think you are right

ui/src/configuration/components/Labels.tsx
```diff
@@ -168,7 +168,9 @@ class Labels extends PureComponent<Props, State> {
   }
 
   private handleNameValidation = (name: string): string | null => {
-    return validateLabelName(this.state.labelTypes, name)
+    const names = this.state.labelTypes.map(label => label.name)
```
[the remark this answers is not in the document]
alexpaxton: That's actually how it was before, I just changed it so it doesn't need to know about the label type. Otherwise I'd fall down a rabbit hole updating types in multiple places. I think once the label types are all standardized in the UI it would be trivial to make this accept an array of labels again

ui/src/configuration/utils/labels.ts
```diff
@@ -9,22 +8,17 @@ import {
 export const randomPresetColor = () =>
   _.sample(PRESET_LABEL_COLORS.slice(1)).colorHex
 
-export const validateLabelName = (
-  labels: LabelType[],
-  name: string,
-  labelID: string = null
-) => {
+export const validateLabelUniqueness = (labelNames: string[], name: string) => {
```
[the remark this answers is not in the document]
alexpaxton: Good catch

ui/src/configuration/components/Labels.tsx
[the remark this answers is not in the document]
ischolten: sounds good

ui/src/shared/components/inlineLabels/InlineLabels.tsx
alexpaxton: Handled!

PR #12244, opened 2019-02-28
Eradicate OrganizationView and begin eradicating GetOrgResources

ebb-tide:
This PR: 
Moves the scrapers, tasks and variables fetching from the `OrganizationView` container in to their own separate containers. 
Moves the handling of the export overlay in orgs->tasks to the router. 
Begins the work for moving org resource fetching to redux. I've completed the work for the tasks endpoint. More to come!


ui/src/organizations/containers/OrgTasksIndex.tsx
```diff
@@ -0,0 +1,140 @@
...
+}
+
+@ErrorHandling
+class OrgTasksIndex extends Component<Props, State> {
+  public state = {
+    loadingState: RemoteDataState.NotStarted,
+  }
+
+  public componentDidMount = async () => {
+    this.setState({loadingState: RemoteDataState.Loading})
```
chnn: Could this loading state exist in Redux instead? Since the data is loaded by a Redux action, I think it makes sense that the loading state would be stored in Redux as well.

It also enables the pattern of connecting the loading state to the component and doing something like:

```
if (loadingState === RemoteDataState.NotStarted) {
  getTasks()
}
```

That way we only load tasks if they haven't already been loaded. This is useful since the tasks are used on multiple pages, and we don't need to reload them when switching pages.

chnn: Also kinda related but not just specific to tasks... when we fetch data through a fetch-and-cache Redux action, we should probably have some way of limiting the concurrency of that action. I.e. if we call `getTasks` twice in a row, the requests will both be setting Redux state. I would expect instead that the first call gets canceled, and the state only gets set once. This will become more important if we store loading states in Redux as well.

ui/src/organizations/containers/OrgVariablesIndex.tsx
```diff
@@ -0,0 +1,106 @@
...
+          <div className="col-xs-12">
+            <Tabs>
+              <OrganizationNavigation tab={'variables'} orgID={org.id} />
+              <Tabs.TabContents>
+                <TabbedPageSection
+                  id="org-view-tab--variables"
+                  url="variables"
+                  title="Variables"
+                >
+                  <GetOrgResources<Variable>
```
chnn: I thought I removed this, how has it risen from the dead??

ui/src/organizations/containers/OrgTasksIndex.tsx
```diff
@@ -0,0 +1,140 @@
...
+import {RemoteDataState} from 'src/types'
+
+interface RouterProps {
+  params: {
+    orgID: string
+  }
+}
+
+interface DispatchProps {
+  notify: NotificationsActions.PublishNotificationActionCreator
```
chnn: Looks like this is unused.

ui/cypress/e2e/variables.test.ts
```diff
@@ -6,7 +6,7 @@ describe('Variables', () => {
 
     cy.signin().then(({body}) => {
       cy.wrap(body.org).as('org')
-      cy.visit(`organizations/${body.org.id}/variables_tab`)
+      cy.visit(`organizations/${body.org.id}/variables`)
```
121watts: looks like youll need to update the other tests using org resource routes to remove the `_tab`

ui/src/organizations/containers/OrganizationsIndex.tsx
```diff
@@ -13,7 +13,7 @@ import {Button, IconFont, ComponentColor} from '@influxdata/clockface'
 import {OverlayTechnology} from 'src/clockface'
 
 // Actions
-import {createOrg, deleteOrg} from 'src/organizations/actions'
+import {createOrg, deleteOrg} from 'src/organizations/actions/orgs'
```
121watts: nit.  why create another file called orgs under `organizations/actions`?

ebb-tide: the reducer was in a file called orgs so thought they should match?

ui/src/organizations/containers/OrgTasksIndex.tsx
ebb-tide: I will consider and implement this in the next PR.

PR #12263, opened 2019-02-28
fix(http/swagger/dashboard): update the swagger post dashboards

OfTheDelmer:
Updates the swagger definition for the post dashboard request.  Using `Dashboard` for both Request and Response causes typing paining in the UI where `labels`, `cells`, etc are optional. 

Our first steps for fixing `labels` being generated as optional on `Bucket`, `Dashboard`, `Telegraf`, and `Task` is to create PRs for separating out their post request schema from the response schema. 

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -1683,7 +1683,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/Dashboard"
+                $ref: "#/components/schemas/CreateDashboardRequest"
       responses:
         '201':
           description: Added dashboard
@@ -6615,6 +6615,20 @@ components:
             $ref: "#/components/schemas/Cells"
         labels:
             $ref: "#/components/schemas/Labels"
+    CreateDashboardRequest:
+      properties:
+        orgID:
+          type: string
+          description: id of the organization that owns the dashboard
+        name:
+          type: string
+          description: user-facing name of the dashboard
+        description:
+          type: string
+          description: user-facing description of the dashboard
+      required:
+        - orgID
+        - name
     Dashboards:
       type: object
       properties:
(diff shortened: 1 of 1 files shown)


desa:
@OfTheDelmer My only hesitancy about this change is that we've been toying with the idea of allowing cells and labels to be posted with a dashboard on create.

> ... causes typing paining in the UI where `labels`, `cells`, etc are optional.

Could you expand a bit more on why that is? Looking at the backend code they should always come back non-nil.

OfTheDelmer:
@desa so there are two issues created by `Dashboard` properties not being marked as required 

- this makes the generated `Dashboard` have all its properties be optional. The causes more types to be created to rewrap in another type with more guarantees on the properties.
- the other issue is that it's unclear what actually is required to be posted to `/dashboards`.

OfTheDelmer:
> My only hesitancy about this change is that we've been toying with the idea of allowing cells and labels to be posted with a dashboard on create

@desa could we resolve that by updating the `CreateDashboardRequest` schema to include `labels`

http/swagger.yml
```diff
@@ -6615,6 +6615,20 @@ components:
             $ref: "#/components/schemas/Cells"
         labels:
             $ref: "#/components/schemas/Labels"
+    CreateDashboardRequest:
```
desa: One possible change would be to have Something like a `DashboardPartial` or `DashboardBody` that both Create and Dashboard could reference so that we don't have to update fields in multiple locations whenever we add a field to dashboards.

desa:
Had a call, just to summarize the results. We are adding an explicit create structure because the expectations of get vs create are different. As we need more fields on create, we can add them.

PR #12284, opened 2019-03-01
Add retry and bail-out to reader

stuartcarnie:
This work is necessary to support gRPC streams.

It is possible a `StreamReader` (gRPC) may return an empty response. This change adds retry and bail-out support. When a bail-out occurs, `reads.ErrStreamNoData` is returned.

### DIFF
--- a/storage/reads/stream_reader.go
+++ b/storage/reads/stream_reader.go
@@ -20,7 +20,16 @@ var (
 	// ErrPartitionKeyOrder means the partition keys for a
 	// GroupResultSetStreamReader were incorrectly ordered.
 	ErrPartitionKeyOrder = errors.New("invalid partition key order")
+
+	// ErrStreamNoData means the StreamReader repeatedly returned no data
+	// when calling Recv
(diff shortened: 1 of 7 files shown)


storage/reads/stream_reader.go
```diff
@@ -340,6 +349,9 @@ func (r *frameReader) init() {
 }
 
 func (r *frameReader) peekFrame() *datatypes.ReadResponse_Frame {
+	retries := peekFrameRetries
+
+RETRY:
```
tmgordeeva: If we have a previous `r.buf` and retry, won't `r.p == 0` so we just return the previous buffer?

stuartcarnie: You are right – I'll clear r.buf on retry. Great catch!

stuartcarnie: I created a test to catch that, however, I was reminded `Recv` always returns a non-nil value, so r.buf was always being overwritten. Regardless, I am explicitly clearing for safety.

PR #12289, opened 2019-03-01
fix(tasks): fix task update validation

### LINKED ISSUE #12100: Not able to override cron option with every in a flux script
I am not sure if that is supported, but I am seeing errors trying to override `cron` option with `every` and vise versa

Running `8016599605faf3b85ae2f3ef98213f8586080dc8@master` of `influxdb` 

- Create a task with `cron` or `every` in flux script:
Task is being created with following params:
```
Organization ID : '0370c48cf85aa000'
Flux Script: 'option task = {name: "task_uceifs", every: 1h} from(bucket:"read_from_b_query_params") |> range(start: -1h)'
Status : 'active'
Token : 'EepQdPw4MGQMmVIwwuy0DUALJBWKRfxuVIWmb9SG1Qv3lM5R2XcuqI9_eZoc0D3TYw9OJblGaR8hHlWukFqLtg=='
```

- Update Task
Request Body:
```
'{"cron": "1 2 3 4 5"}'
```

Error is returned:
```
error_message_headers = 'None'
e_msg = 'failed to update task'
e_code = 'internal error'
e_op = 'None'
e_err = 'cannot use both cron and every in task options'
error_message = '{u'message': u'failed to update task', u'code': u'internal error', u'error': u'cannot use both cron and every in task options'}'
```

docmerlin:
Closes #12100

### DIFF
--- a/task/backend/bolt/bolt.go
+++ b/task/backend/bolt/bolt.go
@@ -173,6 +173,7 @@ func (s *Store) CreateTask(ctx context.Context, req backend.CreateTaskRequest) (
 
 func (s *Store) UpdateTask(ctx context.Context, req backend.UpdateTaskRequest) (backend.UpdateTaskResult, error) {
 	var res backend.UpdateTaskResult
+	fmt.Println("here1")
 	op, err := backend.StoreValidator.UpdateArgs(req)
 	if err != nil {
 		return res, err
--- a/task/backend/store.go
+++ b/task/backend/store.go
@@ -11,9 +11,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/influxdata/flux/ast"
-	"github.com/influxdata/flux/ast/edit"
-	"github.com/influxdata/flux/parser"
 	platform "github.com/influxdata/influxdb"
 	"github.com/influxdata/influxdb/task/options"
 )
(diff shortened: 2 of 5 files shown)


task/backend/bolt/bolt.go
```diff
@@ -173,6 +173,7 @@ func (s *Store) CreateTask(ctx context.Context, req backend.CreateTaskRequest) (
 
 func (s *Store) UpdateTask(ctx context.Context, req backend.UpdateTaskRequest) (backend.UpdateTaskResult, error) {
 	var res backend.UpdateTaskResult
+	fmt.Println("here1")
```
lyondhill: I use these too!!  I sometimes get up to "here17". You might not want to leave it in though.

task/backend/store.go
```diff
@@ -189,54 +186,23 @@ type UpdateTaskRequest struct {
...
-	if ast.Check(parsedPKG) > 0 {
-		return ast.GetError(parsedPKG)
+	if t.Options.IsZero() {
+		return nil
 	}
-	parsed := parsedPKG.Files[0] //TODO: remove this line when flux 0.14 is upgraded into platform
-	// so we don't allocate if we are just changing the status
-	if t.Every != 0 && t.Cron != "" {
-		return errors.New("cannot specify both every and cron")
+	tu := platform.TaskUpdate{
```
lyondhill: If I'm understanding this correctly the backend store no long does its own update flux functionality but now relies on the influxdb.TaskUpdate to do it. If thats true then great. My quiestion is why have this function at all? I think it could be removed.

task/backend/bolt/bolt.go
docmerlin: DOH.

task/backend/store.go
docmerlin: It is still in, because the format of the update objects is different between the two packages, the internal update format needs to be beefier than the external facing one, and the zero values are a little different (needed some string pointers) :-(

PR #12307, opened 2019-03-01
test(explorer): add end-to-end test for empty state

### LINKED ISSUE #12178: test(de): visualization should display empty state when no data is present

AlirieGray:
Closes #12178

This PR adds an end-to-end test that checks for the empty state in the Data Explorer visualization when the query entered returns no results.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/cypress/e2e/explorer.test.ts
+++ b/ui/cypress/e2e/explorer.test.ts
@@ -85,6 +85,28 @@ describe('DataExplorer', () => {
       cy.get('.input-field').type('covariance')
       cy.getByTestID('toolbar-function').should('have.length', 1)
     })
+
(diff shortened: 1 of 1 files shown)


ui/cypress/e2e/explorer.test.ts
```diff
@@ -85,6 +85,28 @@ describe('DataExplorer', () => {
       cy.get('.input-field').type('covariance')
       cy.getByTestID('toolbar-function').should('have.length', 1)
     })
+
+    it.only('shows the empty state when the query returns no results', () => {
```
121watts: still have an only right there :)

PR #12338, opened 2019-03-04
fix(variables): close create overlay when new variable created

### LINKED ISSUE #12332: Sort variables list by name

### LINKED ISSUE #12335: Creating a variable fails to close the overlay

AlirieGray:
Closes #12332
Closes #12335

This PR fixes a couple issues with the manage variables page: the create overlay now closes when the submit button is pressed, and the variables in the index are now sorted by name. Also, a notification is displayed to the user when a variable is successfully created.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/variables/actions/index.ts
+++ b/ui/src/variables/actions/index.ts
@@ -9,6 +9,7 @@ import {
   createVariableFailed,
   updateVariableFailed,
   deleteVariableFailed,
+  createVariableSuccess,
 } from 'src/shared/copy/notifications'
 
 // Types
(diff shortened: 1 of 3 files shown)


ui/src/variables/actions/index.ts
```diff
@@ -104,7 +105,9 @@ export const createVariable = (variable: Variable) => async (
   } catch (e) {
     console.log(e)
     dispatch(notify(createVariableFailed()))
+    return
   }
+  dispatch(notify(createVariableSuccess(variable.name)))
```
chnn: _nit_ It might be easier to read the control here if this was right after line 104 (in the "happy path" for the request)

PR #12344, opened 2019-03-04
feat(http): add an OnFinished handler to the query service

jsternberg:
This is so we can optionally run some code when a query has been
finished and the result has been written out.


jsternberg:
We found a different way to do this.

PR #12347, opened 2019-03-05
refactor(ui/labels): store label state in redux store

121watts:


ui/src/labels/reducers/index.ts
```diff
@@ -0,0 +1,73 @@
...
+
+        return
+      }
+
+      case 'ADD_LABEL': {
+        const {label} = action.payload
+
+        draftState.list.push(label)
+
+        return
```
ebb-tide: shouldn't you be returning state here? how is this working ?

PR #12349, opened 2019-03-05
storage: detect conflicting types in a single batch of points

### LINKED ISSUE #12117: Improve validation of field types in WAL

zeebo:
When the WAL was moved up, the validation that happened at the cache was skipped. This moves the field type validation for a batch of points up ahead of the WAL again.

Closes #12117.

### DIFF
--- a/storage/engine.go
+++ b/storage/engine.go
@@ -405,7 +405,7 @@ func (e *Engine) WritePoints(ctx context.Context, points []models.Point) error {
 	}
 
 	// Convert the points to values for adding to the WAL/Cache.
-	values, err := tsm1.PointsToValues(collection.Points)
+	values, err := tsm1.CollectionToValues(collection)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


storage/engine.go
```diff
@@ -405,7 +405,7 @@ func (e *Engine) WritePoints(ctx context.Context, points []models.Point) error {
 	}
 
 	// Convert the points to values for adding to the WAL/Cache.
```
e-dard: `s/points/collection`

PR #12360, opened 2019-03-05
refactor(query): make queryd present ProxyQueryService

wolffcm:
Closes #

  - [ ] Rebased/mergeable
  - [ ] Tests pass


http/influxdb/source_proxy_query_service.go
```diff
@@ -91,28 +91,39 @@ func (s *SourceProxyQueryService) fluxQuery(ctx context.Context, w io.Writer, re
...
+		return flux.Statistics{}, err
 	}
 	defer resp.Body.Close()
 	if err := platformhttp.CheckError(resp); err != nil {
-		return 0, err
+		return flux.Statistics{}, err
 	}
-	return io.Copy(w, resp.Body)
+
+	_, err = io.Copy(w, resp.Body)
```
jsternberg:
```
if _, err := io.Copy(w, resp.Body); err != nil {
    return flux.Statistics{}, err
}
```

This is a code style I like because it reduces the scope of the variable name and prevents accidentally leaking it to somewhere else.

http/influxdb/source_proxy_query_service.go
```diff
@@ -132,22 +143,32 @@ func (s *SourceProxyQueryService) influxQuery(ctx context.Context, w io.Writer,
...
 	}
 
 	csvDialect, ok := req.Dialect.(csv.Dialect)
 	if !ok {
-		return 0, fmt.Errorf("unsupported dialect %T", req.Dialect)
+		return flux.Statistics{}, fmt.Errorf("unsupported dialect %T", req.Dialect)
 	}
 
-	return csv.NewMultiResultEncoder(csvDialect.ResultEncoderConfig).Encode(w, influxql.NewResponseIterator(res))
+	_, err = csv.NewMultiResultEncoder(csvDialect.ResultEncoderConfig).Encode(w, influxql.NewResponseIterator(res))
```
jsternberg: Same here.

http/proxy_query_service.go
```diff
@@ -57,25 +73,39 @@ func (h *ProxyQueryHandler) handlePostQuery(w http.ResponseWriter, r *http.Reque
...
 	if !ok {
 		EncodeError(ctx, fmt.Errorf("unsupported dialect over HTTP %T", req.Dialect), w)
 		return
 	}
 	hd.SetHeaders(w)
 
-	n, err := h.ProxyQueryService.Query(ctx, w, &req)
+	cw := iocounter.Writer{Writer: w}
+	stats, err := h.ProxyQueryService.Query(ctx, &cw, &req)
+	n := cw.Count()
```
jsternberg: I see this is done to preserve the old code, but it seems unnecessary. You can get rid of line 87 and do `cw.Count() == 0` on line 89. This reduces the scope where this variable could be used which is good for code readability.

http/proxy_query_service.go
```diff
@@ -90,31 +120,75 @@ type ProxyQueryService struct {
...
+		return flux.Statistics{}, err
 	}
 	defer resp.Body.Close()
 	if err := CheckError(resp); err != nil {
-		return 0, err
+		return flux.Statistics{}, err
 	}
-	return io.Copy(w, resp.Body)
+
+	_, err = io.Copy(w, resp.Body)
```
jsternberg: Here too.

http/query_handler.go
```diff
@@ -336,14 +352,25 @@ func (s *FluxService) Query(ctx context.Context, w io.Writer, r *query.ProxyRequ
...
+		return flux.Statistics{}, err
 	}
 	defer resp.Body.Close()
 
 	if err := CheckError(resp); err != nil {
-		return 0, err
+		return flux.Statistics{}, err
+	}
+
+	_, err = io.Copy(w, resp.Body)
```
jsternberg: Here too.

http/proxy_query_service.go
```diff
@@ -9,37 +9,53 @@ import (
...
+	"github.com/influxdata/flux/iocounter"
+	icontext "github.com/influxdata/influxdb/context"
 	"github.com/influxdata/influxdb/query"
 	"github.com/julienschmidt/httprouter"
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
 )
 
+// This file contains ProxyQueryHandler and ProxyQueryService
+// which implement services and handlers for both queryd and transpilerde.
```
jsternberg: I don't think we should reference names for the enterprise product in open source. It's not a problem with information security, but more that the core product should be able to stand on its own as an individual piece of code in my opinion.

http/query_service.go
```diff
@@ -1,242 +0,0 @@
...
-	Logger *zap.Logger
-
-	csvDialect csv.Dialect
-
-	QueryService     query.QueryService
-	CompilerMappings flux.CompilerMappings
-}
-
-// NewQueryHandler returns a new instance of QueryHandler.
-func NewQueryHandler() *QueryHandler {
```
jsternberg: This makes this into a pretty big breaking change in my opinion and we should caution against doing this. It's something we're going to need to discuss if we accept this change.

I think the easiest thing to do is to keep this code and have `QueryHandler` change to using proxy query service. That's a breaking change for Go code, but the HTTP API remains the same and it has a far smaller surface area.

http/proxy_query_service.go
wolffcm: This is a good point.  Actually this code is only used in the enterprise product, not in OSS at all.  A follow-up would be to move this into idpe.  I can change the comment in the meantime.

http/query_service.go
wolffcm: I think a goal that @nathanielc had with this refactoring was to make it so that the code that gateway uses to talk to `queryd` and `transpilerde` is the same to reduce technical debt.  

We should discuss with Nathaniel when we're all around.

jsternberg: I think that's fine and a good goal. My only issue is that it's combining too many things in this PR which makes it more complex to review and adds additional points where it may break.

We can discuss this with Nathaniel though to see if it's worth going down that road.

http/query_handler.go
```diff
@@ -98,20 +99,34 @@ func (h *FluxHandler) handleQuery(w http.ResponseWriter, r *http.Request) {
...
+	// Write statistics trailer
+	data, err := json.Marshal(stats)
+	if err != nil {
+		h.Logger.Info("Failed to encode statistics",
+			zap.String("handler", "flux"),
+			zap.Error(err))
+		return
 	}
+
+	w.Header().Set(QueryStatsTrailer, string(data))
```
nathanielc: @goller You are correct, this should not be changed. The public API should not expose the trailer and should not be changing.

http/query_service.go
nathanielc: This change was quite big and effected lots of different bits,as such I saw it as valuable to delete this dead code as that was the intent of this change. By deleting the code we are confident that we put everything back together correctly without relying on this old code.

http/proxy_query_service.go
```diff
@@ -9,37 +9,50 @@ import (
...
+	icontext "github.com/influxdata/influxdb/context"
 	"github.com/influxdata/influxdb/query"
 	"github.com/julienschmidt/httprouter"
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/zap"
 )
 
 const (
-	proxyQueryPath = "/api/v2/queryproxysvc"
+	proxyQueryPath    = "/api/v2/queryproxysvc"
```
nathanielc: should we rename this to be `/queryproxysvc` since its not really part of the public API?

http/source_proxy_service.go
```diff
@@ -90,12 +96,17 @@ func (s *SourceProxyQueryService) queryInfluxQL(ctx context.Context, w io.Writer
...
-		return 0, err
+		return flux.Statistics{}, err
 	}
-	return io.Copy(w, resp.Body)
+
+	if _, err = io.Copy(w, resp.Body); err != nil {
+		return flux.Statistics{}, err
+	}
+
+	return flux.Statistics{}, nil
```
nathanielc: Why does these not return a complete flux.Statistics struct?

query/bridges.go
```diff
@@ -23,35 +22,76 @@ func (b QueryServiceBridge) Query(ctx context.Context, req *Request) (flux.Resul
...
 
-	encoder := req.Dialect.Encoder()
-	n, err := encoder.Encode(w, results)
+	go func() {
+		var stats flux.Statistics
+		var err error
+		defer func() {
+			_ = w.CloseWithError(err)
+			statsChan <- stats
+		}()
```
nathanielc: Why is this a defer instead of following the call to Query?

http/source_proxy_service.go
wolffcm: This file seems to be talking to the public API, based on the path in the URL.  So there won't be any statistics in the response.

query/bridges.go
wolffcm: If the call to `Query` panics somewhere, we may not close the pipe and write the stats, is that right?  I wanted to make sure we do that even if a panic occurs.

http/proxy_query_service.go
wolffcm: Is there a convention in place for public vs. internal URLs?

nathanielc: Not to my knowledge.

http/source_proxy_service.go
nathanielc: I am confused why this implements a query.ProxyQueryService then if it never needs the stats. Perhaps we should just create an issue to investigate as that is not really part of this change.

query/bridges.go
nathanielc: If a panic occurs it will crash the whole queryd process. Since a new go routine was created if that go routine panics then the process will exit. In that case there is no real need to clean up the pipe. I think we can ignore panics here as it means a complete process shutdown.

PR #12366, opened 2019-03-05
fix: flux imports

### LINKED ISSUE #12265: Cannot save a script with import statements as a task from the DE
1. Write a script in the Data Explorer that uses import statements, e.g.
   ```

   v1.tagValues(bucket: "a", tag: "_field")
   ```
2. Attempt to save this script as a task by clicking the `Save As...` button

Creating the task will fail, since it attempts to create a task with the following script:

```
option task = {
  // ...
}

v1.tagValues(bucket: "a", tag: "_field")
```

Or if variables are used in the script, it will attempt to create the following:

```
option task = {
  // ...
}

option v = {
  // variable definitions here
}

v1.tagValues(bucket: "a", tag: "_field")
```

These are invalid Flux scripts, since `import` statements must always come first.

The solution to this may involve fetching the AST for the script and using it to seperate the imports and non-imports sections of the script, so that the `option`s may be inserted in between. Another solution would be to add support for the `extern` parameter to the tasks API, and always encode the options via the `extern` parameter.

121watts:
Closes #12265

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -183,6 +183,7 @@
     "redux-auth-wrapper": "^1.0.0",
     "redux-thunk": "^1.0.3",
     "rome": "^2.1.22",
+    "ui": "^0.2.4",
     "uuid": "^3.2.1"
   }
 }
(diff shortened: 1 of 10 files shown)


ui/package.json
```diff
@@ -183,6 +183,7 @@
     "redux-auth-wrapper": "^1.0.0",
     "redux-thunk": "^1.0.3",
     "rome": "^2.1.22",
+    "ui": "^0.2.4",
```
mavarius: Is this supposed to be in this PR?

PR #2443, opened 2015-04-28
Update collectd plugin to write values under the field "value"

josh-padnick:
This is a fix for #2412.  I'm a golang newb, so I'll describe my testing procedure to make sure it's sufficient.

1. Forked repo
2. Used `go get x` to install each dependency
3. Ran `go test` and received `PASS`
4. Made fix
5. Ran `go test` and received `PASS`

Simple change, but this was holding us up.  Thanks.

PR #12382, opened 2019-03-06
Add templates

desa:
Closes #2443

This PR implements templates via a generic documents resource type.

TODO

- [x] add swagger definition
- [x] test http endpoints
- [x] comment code and resolve TODOs

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -147,6 +147,113 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/OnboardingResponse"
+  /documents/templates:
+    get:
+      tags:
+        - Templates
+      parameters:
+          - $ref: '#/components/parameters/TraceSpan'
+          - in: query
+            name: org
+            description: specifies the name of the organization of the template
+            required: true
+            schema:
+              type: string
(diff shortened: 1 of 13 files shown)


http/swagger.yml
```diff
@@ -147,6 +147,113 @@ paths:
...
+          application/json:
+            schema:
+              $ref: "#/components/schemas/DocumentCreate"
+      responses:
+        '201':
+          description: Template created
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Telegraf"
```
goller: this would be a document schema

```$ref: "#/components/schemas/Document"```

File: http/swagger.yml [hunk 2]
```diff
@@ -6753,6 +6860,64 @@ components:
...
+      properties:
+        meta:
+          $ref: "#/components/schemas/DocumentMeta"
+        content:
+          type: object
+        org:
+          type: string
+        orgID:
+          type: string
+        labels:
```
goller: would be good to add a comment on labels that they are strings of labels that will be attached to this object.

Also, mention that only one of org/orgID is required.

PR #12384, opened 2019-03-06
Inline Label Editing from Dashboards Index

### LINKED ISSUE #12233: Implement inline label editing components on dashboards index
Requires https://github.com/influxdata/influxdb/issues/12232 be completed first

alexpaxton:
Closes #12233 

- Using `ILabel` type from `@influxdata/influx` wherever possible, had to touch a lot of files because of this
- Updated labels config page slightly
- Created `createLabelAJAX` which returns the newly created label for use in inline label creation (must be created the immediately added to a dashboard)
- Removed dashboards unit test in favor of stronger e2e dashboards index tests
- Split cypress tool `createLabel` into two functions:
  - `createAndAddLabel`
  - `createLabel`

Checklist:

  - [x] Rebased/mergeable
  - [ ] Tests pass


proto/bin_gen.go
```diff
@@ -84,7 +84,7 @@ func systemJson() (*asset, error) {
 		return nil, err
 	}
 
-	info := bindataFileInfo{name: "system.json", size: 30129, mode: os.FileMode(420), modTime: time.Unix(1548273842, 0)}
+	info := bindataFileInfo{name: "system.json", size: 30129, mode: os.FileMode(420), modTime: time.Unix(1548960115, 0)}
```
mavarius: What is this? Was it meant to be part of this PR?

ui/src/clockface/components/label/Label.tsx
```diff
@@ -59,34 +59,37 @@ class Label extends Component<Props, State> {
...
     }
   }
 
   private handleDelete = (): void => {
     const {id, onDelete} = this.props
 
-    if (onDelete) {
-      onDelete(id)
-    }
+    onDelete(id)
```
mavarius: `onDelete` is optional, perhaps this should remain in an `if` statement.

proto/bin_gen.go
alexpaxton: I’ll have to revert these changes

ui/src/clockface/components/label/Label.tsx
alexpaxton: I took it out because the button that calls the delete function is also conditionally rendered with the same check, seemed redundant. I can add this guard back in to be extra safe

PR #12391, opened 2019-03-06
Reorder series key tagset

e-dard:
This PR makes a **breaking** change to the 2.x line.

Currently we use the special tag keys `_m` and `_f` to index the measurement name and field key. 
This results in keys ordered like this:

```
<orgbucketid>,ABBA=cool,_f=temp,_m=cpu,server=b
<orgbucketid>,ABBA=cool,_f=zz,_m=mem,server=a
<orgbucketid>,_f=temp,_m=cpu,server=a
<orgbucketid>,_f=yy,_m=mem,server=a
<orgbucketid>,_f=yy,_m=mem,server=b
```

With this change we will now use the null byte for measurement name, and byte `255` for the field key.

Keys will now be ordered as follows:

```
<orgbucketid>,\x00=cpu,ABBA=cool,server=b,\xff=temp
<orgbucketid>,\x00=cpu,server=a,\xff=temp
<orgbucketid>,\x00=mem,ABBA=cool,server=a,\xff=zz
<orgbucketid>,\x00=mem,server=a,\xff=yy
<orgbucketid>,\x00=mem,server=b,\xff=yy
```

which is preferable in terms of access patterns, and means we need two bytes less per series key.

Finally, this change will ensure that tag sets are ordered the same way in series keys on both the 1.x and 2.x lines. We can guarantee this because we will only ever be adding a tag pair to the beginning and end of a tag set. This should lead to more elegant and performant implementations of import/export of TSM data between the 1.x and 2.x lines.

Further, this PR adds another change:

  - **all** user provided tag keys, values and field key will be validated as only containing valid utf-8 characters.


models/points.go
```diff
@@ -2462,9 +2475,10 @@ func appendField(b []byte, k string, v interface{}) []byte {
 	return b
 }
 
-// ValidKeyToken returns true if the token used for measurement, tag key, or tag
-// value is a valid unicode string and only contains printable, non-replacement characters.
-func ValidKeyToken(s string) bool {
+// ValidToken returns true if the provided token is a valid unicode string, and
+// only contains printable, non-replacement characters.
+func ValidToken(a []byte) bool {
+	s := string(a)
```
zeebo: There's a `uf8.Valid` that takes a `[]byte`. Given this is in a hot path, that's probably useful.

storage/config.go
```diff
@@ -9,9 +9,10 @@ import (
 	"github.com/influxdata/influxdb/tsdb/tsm1"
 )
 
+// Default configuration values.
 const (
-	DefaultRetentionInterval = 1 * time.Hour
-	DefaultValidateKeys      = false
+	DefaultRetentionInterval   = time.Hour
+	DefaultTraceLoggingEnabled = false
```
zeebo: IIRC this got removed by @jacobmarble's PR for tracing?

storage/engine.go
```diff
@@ -355,40 +355,62 @@ func (e *Engine) CreateCursorIterator(ctx context.Context) (tsdb.CursorIterator,
...
+
+		// Last tag key is not field tag.
+		if !bytes.Equal(fkey, models.FieldKeyTagKeyBytes) {
+			dropPoint(iter.Key(), fmt.Sprintf("missing required field key tag as last tag, got: %q", tags[0].Key))
+			continue
+		}
+
+		// The value representing the underlying field key is invalid if it's "time".
+		if bytes.Equal(fval, timeBytes) {
+			dropPoint(iter.Key(), fmt.Sprintf("invalid field key: input field %q is invalid", timeBytes))
```
zeebo: I've always wondered and this is a good time to ask: why does this need to be invalid?

tsdb/explode.go
```diff
@@ -45,19 +34,22 @@ func ExplodePoints(org, bucket platform.ID, points []models.Point) ([]models.Poi
...
+
+		tags[0] = models.NewTag(models.MeasurementTagKeyBytes, pt.Name())
 		pt.ForEachTag(func(k, v []byte) bool {
 			tags = append(tags, models.NewTag(k, v))
 			return true
 		})
 
 		t := pt.Time()
 		itr := pt.FieldIterator()
+		tags = append(tags, models.Tag{}) // Make room for
```
zeebo: nit: the comment is trunc

models/points.go
e-dard: dang, not sure how I missed that.

storage/engine.go
e-dard: I can't remember the 1.x tsm details. We can loosen this restriction later.

PR #12404, opened 2019-03-06
fix(storage/reads): track memory allocations when reading from storage

jsternberg:
Related to influxdata/flux#1021.


jsternberg:
We should definitely consider adding unit tests for `storage.go`. But, that currently doesn't exist so I'll try to add it in the future.

PR #12406, opened 2019-03-06
feat(ui/oauth/redirect): redirects to /api/v2/signin

OfTheDelmer:
Creates a redirect for `/api/v2/signin` in the ui when `CLOUD=true` otherwise does the typical auth flow for OSS. 

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/Signin.tsx
+++ b/ui/src/Signin.tsx
@@ -67,6 +67,7 @@ export class Signin extends PureComponent<Props, State> {
 
   private checkForLogin = async () => {
     try {
+      this.setState({loading: RemoteDataState.Loading})
       await client.users.me()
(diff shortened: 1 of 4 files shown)


ui/src/Signin.tsx
```diff
@@ -68,12 +68,20 @@ export class Signin extends PureComponent<Props, State> {
   private checkForLogin = async () => {
     try {
       await client.users.me()
+      this.setState({loading: RemoteDataState.Done})
```
ischolten: do we need a loading state before the await?

OfTheDelmer: @ischolten ahh it looks like it was transitioning from `NotStarted` to `Done` before I can add a loading state.

PR #12433, opened 2019-03-07
Update users dropdown to filter out the users that already exist in org

### LINKED ISSUE #12426: Adding a member to an org doesn't update the member list with new members

Palakp41:
Closes #12426 #12411 

Changed users to be a key value object instead of an array. Added filter on users object for the members that already exist in an org. Updated the component to re-fetch members when save is clicked on the overlay. 

  - [x] Rebased/mergeable
  - [x] Tests pass

kapture 2019-03-07 at 11 28 02

### DIFF
--- a/ui/src/organizations/components/AddMembersForm.tsx
+++ b/ui/src/organizations/components/AddMembersForm.tsx
@@ -5,11 +5,13 @@ import React, {PureComponent} from 'react'
 import {Button, ButtonType} from '@influxdata/clockface'
 import {Form, Grid} from 'src/clockface'
 import SelectUsers from 'src/organizations/components/SelectUsers'
-import {User} from '@influxdata/influx'
+
+// Types
+import {UsersMap} from 'src/organizations/components/Members'
 
 interface Props {
   onCloseModal: () => void
-  users: User[]
+  users: UsersMap
   onSelect: (selectedIDs: string[]) => void
   selectedUserIDs: string[]
   onSave: () => void
--- a/ui/src/organizations/components/Members.tsx
+++ b/ui/src/organizations/components/Members.tsx
@@ -14,6 +14,7 @@ import {
 import MemberList from 'src/organizations/components/MemberList'
 import FilterList from 'src/shared/components/Filter'
 import AddMembersOverlay from 'src/organizations/components/AddMembersOverlay'
+import {Button, ComponentColor} from '@influxdata/clockface'
 
 // Actions
 import * as NotificationsActions from 'src/types/actions/notifications'
@@ -21,11 +22,10 @@ import * as NotificationsActions from 'src/types/actions/notifications'
 // Types
 import {
   ResourceOwner,
-  User,
   AddResourceMemberRequestBody,
+  User,
 } from '@influxdata/influx'
 import {OverlayState} from 'src/types'
-import {Button, ComponentColor} from '@influxdata/clockface'
 
 // APIs
 import {client} from 'src/utils/api'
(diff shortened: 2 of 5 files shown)


ui/src/organizations/components/Members.tsx
```diff
@@ -34,17 +30,29 @@ import {
   memberAddFailed,
 } from 'src/shared/copy/v2/notifications'
 
+export interface UserObject {
+  [key: string]: {
```
OfTheDelmer:
```suggestion
  [userID: string]: User
```

Is the value here just a User interface or maybe `Pick<User, 'id' | 'name' | 'links'>`

ui/src/organizations/components/AddMembersForm.tsx
```diff
@@ -5,11 +5,13 @@ import React, {PureComponent} from 'react'
 import {Button, ButtonType} from '@influxdata/clockface'
 import {Form, Grid} from 'src/clockface'
 import SelectUsers from 'src/organizations/components/SelectUsers'
-import {User} from '@influxdata/influx'
+
+// Types
+import {UserObject} from 'src/organizations/components/Members'
```
OfTheDelmer: We might be able to use a `Map` here? It could be typed as `Map<User['id'], User>`

ui/src/organizations/components/Members.tsx
```diff
@@ -109,8 +117,28 @@ export default class Members extends PureComponent<Props, State> {
...
 
   private async getUsers() {
+    const {members} = this.props
+
     const data = await client.users.getAllUsers()
-    this.setState({users: data.users})
+
+    const users = {}
+
+    for (const key of data.users) {
```
OfTheDelmer: We could make this functional using `data.users.forEach`

ui/src/organizations/components/Members.tsx
```diff
@@ -109,8 +117,28 @@ export default class Members extends PureComponent<Props, State> {
...
+      } else {
+        users[key.id] = key
+      }
+    }
+
+    members.forEach(m => {
+      if (users[m.id]) {
+        delete users[m.id]
+      }
+      return
```
OfTheDelmer:
```suggestion
   
```

ui/src/organizations/components/Members.tsx
```diff
@@ -109,8 +117,28 @@ export default class Members extends PureComponent<Props, State> {
...
   private async getUsers() {
+    const {members} = this.props
+
     const data = await client.users.getAllUsers()
-    this.setState({users: data.users})
+
+    const users = {}
+
+    for (const key of data.users) {
+      if (users[key.id]) {
```
OfTheDelmer:
```suggestion
      users[key.id] = key
```

ui/src/organizations/components/Members.tsx
```diff
@@ -34,17 +30,29 @@ import {
   memberAddFailed,
 } from 'src/shared/copy/v2/notifications'
 
+export interface UserObject {
```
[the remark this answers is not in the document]
Palakp41: Yes, I changed it to UsersMap

ui/src/organizations/components/Members.tsx
Palakp41: Yes it's the user interface. I gave it the same type

ui/src/organizations/components/AddMembersForm.tsx
Palakp41: I prefer the use of object with key and value here for the users

ui/src/organizations/components/Members.tsx
Palakp41: Yes, I updated it

Issue #12439, opened 2019-03-07
cli: influx setup should not overwrite existing credentials file


mark-rushakoff:
I just ran `influx setup` against a different instance than usual, and it looks like it silently overwrote what used to be in `~/.influxdbv2/credentials`.

https://github.com/influxdata/influxdb/issues/12161 is about supporting different profiles, which is a better long-term solution. Until then, the simple solution I might propose would be to immediately fail `influx setup` if the credentials file already exists, so that the user can decide whether to move the file out of the way, provide a different directory, or delete the existing data.

PR #12443, opened 2019-03-08
fix(influx): return the error when token file already exists

kelwang:
Closes https://github.com/influxdata/influxdb/issues/12439

manually tested

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/cmd/influx/setup.go
+++ b/cmd/influx/setup.go
@@ -62,6 +62,18 @@ func setupF(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("instance at %q has already been setup", flags.host)
 	}
 
+	dPath, dir, err := defaultTokenPath()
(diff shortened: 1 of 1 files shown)


cmd/influx/setup.go
```diff
@@ -62,6 +62,18 @@ func setupF(cmd *cobra.Command, args []string) error {
...
 
+	dPath, dir, err := defaultTokenPath()
+	if err != nil {
+		return err
+	}
+
+	if _, err := os.Stat(dPath); err == nil {
+		return &platform.Error{
+			Code: platform.EConflict,
+			Msg:  fmt.Sprintf("Your token has already been existed in %s", dPath),
```
goller: perhaps "token already exists at %s"

PR #12444, opened 2019-03-08
Use alternate runtime APIs to reduce allocs / cpu

stuartcarnie:
```
benchmark                               old ns/op     new ns/op     delta
BenchmarkLocal_StartSpanFromContext-8   1382          718           -48.05%

benchmark                               old allocs    new allocs    delta
BenchmarkLocal_StartSpanFromContext-8   6             4             -33.33%

benchmark                               old bytes     new bytes     delta
BenchmarkLocal_StartSpanFromContext-8   376           224           -40.43%
```


jacobmarble:
Added a small commit to the PR, to help understand performance composition. Thanks for this Stuart!

stuartcarnie:
Good call - allows us to understand the relative perf of those runtime functions

PR #12459, opened 2019-03-08
fix(kv): deleted organizations leave orphaned buckets

desa:
This PR removed an organizations bucket when the organization is deleted.

### DIFF
--- a/kv/org.go
+++ b/kv/org.go
@@ -407,12 +407,28 @@ func (s *Service) updateOrganization(ctx context.Context, tx Tx, id influxdb.ID,
 	return o, nil
 }
 
+func (s *Service) deleteOrganizationsBuckets(ctx context.Context, tx Tx, id influxdb.ID) error {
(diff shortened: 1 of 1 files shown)


kv/org.go
```diff
@@ -407,12 +407,28 @@ func (s *Service) updateOrganization(ctx context.Context, tx Tx, id influxdb.ID,
...
+	filter := influxdb.BucketFilter{
+		OrganizationID: &id,
+	}
+	bs, err := s.findBuckets(ctx, tx, filter)
+	if err != nil {
+		return err
+	}
+	for _, b := range bs {
+		if err := s.deleteBucket(ctx, tx, b.ID); err != nil {
+			return err
```
mark-rushakoff: If something went wrong deleting one bucket you probably want to keep going. Maybe just log the error here instead of returning it.

desa: good idea

Issue #12462, opened 2019-03-08
Finding a bucket by name returns the first match when the name exists in multiple orgs


mark-rushakoff:
We're hitting this specifically when creating a task. Here's what happens:

1. Orgs O1 and O2 both contain a bucket named "mybucket".
1. The user in org O2 creates a task, with content like `from(bucket: "mybucket") |> ...`.
1. The task validator checks if the current authorization is allowed to access the buckets in the task... https://github.com/influxdata/influxdb/blob/afcfb1085919d7ada1240b30a81b0ac16d070843/task/validator.go#L92-L110
1. The query preauthorizer looks up the bucket https://github.com/influxdata/influxdb/blob/afcfb1085919d7ada1240b30a81b0ac16d070843/query/preauthorizer.go#L31-L74 by using the `from` function's BucketsAccessed https://github.com/influxdata/influxdb/blob/afcfb1085919d7ada1240b30a81b0ac16d070843/query/stdlib/influxdata/influxdb/from.go#L83-L102
1. At this point, BucketService.FindBucket is called with just the bucket name "mybucket" and returns the bucket belonging to O1, which the user is not allowed to access.

The solution to this problem seems subtle.

- Should BucketsAccessed require an organization ID?
- Should BucketService.FindBucket require an organization ID?
- Should BucketService.FindBucket use the authorizer on its context to filter the results down to what is allowed to be accessed?

Until this is fixed, the workarounds are: use unique bucket names, or use `from(bucketID: "012345")` instead of the bucket name.

PR #12465, opened 2019-03-08
fix(query): fix same bucket name with different orgs

kelwang:
Closes https://github.com/influxdata/influxdb/issues/12462

Add an optional orgID to the preAuthorizer func

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/task/validator.go
+++ b/task/validator.go
@@ -102,7 +102,7 @@ func (ts *taskServiceValidator) CreateTask(ctx context.Context, t platform.TaskC
 		return nil, err
 	}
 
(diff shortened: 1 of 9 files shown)


task/validator.go
```diff
@@ -294,7 +294,7 @@ func validatePermission(ctx context.Context, perm platform.Permission) error {
 	return nil
 }
 
-func validateBucket(ctx context.Context, script string, preAuth query.PreAuthorizer) error {
+func validateBucket(ctx context.Context, script string, preAuth query.PreAuthorizer, OrganizationID platform.ID) error {
```
mark-rushakoff: Argument names should start with a lowercase letter.

PR #12471, opened 2019-03-08
feat(ui/export/json): format export using jsonapi

### LINKED ISSUE #12324: On export, fetch associated resources and convert dashboard to template

OfTheDelmer:
Closes #12324 

Transforms the dashboard, cells and views using the json api relationships, attributes, included, and id,  and type.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/shared/utils/resourceToTemplate.ts
+++ b/ui/src/shared/utils/resourceToTemplate.ts
@@ -1,6 +1,6 @@
 import _ from 'lodash'
 import {getDeep} from 'src/utils/wrappers'
-import {Task, Label, Dashboard} from 'src/types/v2'
+import {Task, Label, Dashboard, Cell, View} from 'src/types/v2'
 import {ITemplate, TemplateType} from '@influxdata/influx'
 
 const CURRENT_TEMPLATE_VERSION = '1'
@@ -15,9 +15,6 @@ const blankTaskTemplate = () => {
   const baseTemplate = blankTemplate()
   return {
     ...baseTemplate,
-    meta: {
-      ...baseTemplate.meta,
-    },
     content: {
       ...baseTemplate.content,
       data: {...baseTemplate.content.data, type: TemplateType.Task},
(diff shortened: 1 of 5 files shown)


ui/src/shared/utils/resourceToTemplate.ts
```diff
@@ -96,8 +96,87 @@ export const taskToTemplate = (
...
+      }
+    }
+  }
+}
+
+const cellToRelationship = (cell: Cell) => ({
+  type: TemplateType.Cell, id: cell.id
+})
+
+// Todo: update template to match task format
```
ebb-tide: is this still a todo?

ui/src/shared/utils/resourceToTemplate.ts
```diff
@@ -96,8 +96,87 @@ export const taskToTemplate = (
...
+export const dashboardToTemplate = (
+  dashboard: Dashboard,
+  views: View[],
+  baseTemplate = blankTaskTemplate(),
```
ebb-tide: should probably use blankDashboard Template?

ui/src/shared/utils/resourceToTemplate.ts
```diff
@@ -96,8 +96,87 @@ export const taskToTemplate = (
...
+      ...baseTemplate.meta,
+      name: templateName,
+      version: CURRENT_TEMPLATE_VERSION,
+      description: `template created from dashboard: ${dashboardName}`,
+    },
+    content: {
+      ...baseTemplate.content,
+      data: {
+        ...baseTemplate.content.data,
+        type: TemplateType.Dashboard,
```
ebb-tide: so I had put type in the baseTemplate for a dashboard, but I'm not sure that it has much utility anymore. I would love to test this but we don't have tests in client.

ui/src/shared/utils/resourceToTemplate.ts
```diff
@@ -96,8 +96,87 @@ export const taskToTemplate = (
...
+}
+
+const cellToRelationship = (cell: Cell) => ({
+  type: TemplateType.Cell, id: cell.id
+})
+
+// Todo: update template to match task format
+export const dashboardToTemplate = (
+  dashboard: Dashboard,
+  views: View[],
```
ebb-tide: I think it would be appropriate for the client to worry about getting views instead of expecting to be passed.  Now that I think of it, this should be a function that takes just a dashboard id and does everything else itself. Let's get this in as is, but I'll open an issue to refactor.

ui/src/shared/utils/resourceToTemplate.ts
ischolten: nope. ill remove it

ui/src/shared/utils/resourceToTemplate.ts
ischolten: good catch!

Issue #12306, opened 2019-03-01
Import Overlay should use semantic html


ebb-tide:
Import overlay is just a text area and not a form, we could benefit from using semantic html

PR #12496, opened 2019-03-11
feat(ui/dash/import): create dashboards from imported templates

### LINKED ISSUE #11533: On import create dashboard from template, also create any included resources
as a user
with admin
when i see a list of resources (dashboard/task/variable for now)
then i would like the ability to export resource as JSON, 
and import it back from that JSON, to be instantiated as that resource. 

If imported JSON is not valid for that resource, I would like to see an error. 

what happens to other resources bundled with that resource? like variables included in dashboard?

ischolten:
Closes #11533

Create a dashboard from the imported dashboard.
Make import button in import overlay visible and disabled rather than not visible

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/components/ImportDashboardOverlay.tsx
+++ b/ui/src/dashboards/components/ImportDashboardOverlay.tsx
@@ -1,70 +1,80 @@
+// Libraries
 import React, {PureComponent} from 'react'
 import _ from 'lodash'
+import {connect} from 'react-redux'
 
-import Container from 'src/clockface/components/overlays/OverlayContainer'
-import Heading from 'src/clockface/components/overlays/OverlayHeading'
-import Body from 'src/clockface/components/overlays/OverlayBody'
-import DragAndDrop from 'src/shared/components/DragAndDrop'
+// Constants
 import {dashboardImportFailed} from 'src/shared/copy/notifications'
 
-import {Dashboard} from 'src/types/v2'
-import {Notification} from 'src/types/notifications'
+// Actions
+import {notify as notifyAction} from 'src/shared/actions/notifications'
+import {getDashboardsAsync} from 'src/dashboards/actions/v2'
 
-interface Props {
+// Types
+import ImportOverlay from 'src/shared/components/ImportOverlay'
+import {createDashboardFromTemplate as createDashboardFromTemplateAction} from 'src/dashboards/actions/v2'
+
+interface OwnProps {
   onDismissOverlay: () => void
-  onImportDashboard: (dashboard: Dashboard) => void
-  notify: (message: Notification) => void
+  orgID: string
+  isVisible: boolean
 }
+interface DispatchProps {
+  notify: typeof notifyAction
+  createDashboardFromTemplate: typeof createDashboardFromTemplateAction
+  populateDashboards: typeof getDashboardsAsync
+}
+
+type Props = OwnProps & DispatchProps
+
 class ImportDashboardOverlay extends PureComponent<Props> {
   constructor(props: Props) {
     super(props)
   }
 
   public render() {
-    const {onDismissOverlay} = this.props
+    const {isVisible, onDismissOverlay} = this.props
 
     return (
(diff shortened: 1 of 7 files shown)


ui/src/dashboards/components/ImportDashboardOverlay.tsx
```diff
@@ -1,70 +1,84 @@
...
-      const {dashboard} = JSON.parse(uploadContent)
+      const template = JSON.parse(uploadContent)
 
-      if (!_.isEmpty(dashboard)) {
-        onImportDashboard(dashboard)
+      if (_.isEmpty(template)) {
         onDismissOverlay()
-      } else {
-        notify(dashboardImportFailed('No dashboard found in file'))
+        return
```
bthesorceror: when would the template be empty and not have failed? Shouldn't we show a user an error message?

ui/src/dashboards/components/dashboard_index/DashboardsIndex.tsx
```diff
@@ -229,13 +208,13 @@ class DashboardIndex extends PureComponent<Props, State> {
...
+    const {orgs} = this.props
 
     return (
-      <ImportOverlay
-        isVisible={isImportingDashboard}
-        resourceName="Dashboard"
+      <ImportDashboardOverlay
         onDismissOverlay={this.handleToggleImportOverlay}
-        onSubmit={this.handleImportDashboard}
+        orgID={get(orgs, '0.id', '')}
```
bthesorceror: Why is this defaulting to the first org? Shouldn't we let the user select org?

ui/src/organizations/components/Dashboards.tsx
```diff
@@ -205,46 +202,20 @@ class Dashboards extends PureComponent<Props, State> {
...
-      <OverlayTechnology visible={isImportingDashboard}>
-        <ImportDashboardOverlay
-          onDismissOverlay={this.handleToggleOverlay}
-          onImportDashboard={this.handleImportDashboard}
-          notify={notify}
-        />
-      </OverlayTechnology>
+      <ImportDashboardOverlay
+        onDismissOverlay={this.handleToggleOverlay}
+        orgID={_.get(orgs, '0.id', '')}
```
bthesorceror: same comment as above

ui/src/shared/components/ImportOverlay.tsx
```diff
@@ -110,18 +112,21 @@ export default class ImportOverlay extends PureComponent<Props, State> {
...
-      )
-    }
+    const status = isEnabled
+      ? ComponentStatus.Default
+      : ComponentStatus.Disabled
+
+    return (
+      <Button
+        text={`Import JSON as ${resourceName}`}
+        onClick={this.submit}
```
bthesorceror: Can we make a card about this not being a form?

ui/src/dashboards/components/dashboard_index/DashboardsIndex.tsx
ischolten: this is defaulting to the first org since that's what creating a dashboard does. In a separate pr, i will add an org dropdown to the import overlay so that a user may choose.

ui/src/organizations/components/Dashboards.tsx
ischolten: same as above

ui/src/shared/components/ImportOverlay.tsx
ischolten: there already is one: #12306

ui/src/dashboards/components/ImportDashboardOverlay.tsx
ischolten: it wouldnt. i will take out that check.

ui/src/dashboards/components/dashboard_index/DashboardsIndex.tsx
ischolten: #12496

PR #12525, opened 2019-03-11
Populate and substitute variables on dashboard load

### LINKED ISSUE #12470: Detect, populate, and substitute variables when loading a dashboard
To load a dashboard:

1. Load all the variables the current user has access to, the dashboard, and all of the views for the dashboard
2. Scan the queries in each view to find which variables are in use on the dashboard
3. [Populate the values for the variables in use](https://github.com/influxdata/influxdb/blob/ae5bdeb997c2d8b21321f0d377a18d5ba3ca3017/ui/src/variables/actions/index.ts#L168)
4. Render the dashboard and all of its views
5. Execute the queries for each view, substituting the appropriate variable values where necessary

chnn:
Closes #12470

This enables a user to create a cell that uses an external variable (e.g. `from(bucket: v.my_bucket)`). Then when that dashboard is first loaded, an appropriate value for the `my_bucket` variable will be loaded, selected, and substituted before executing the query for that cell.

It doesn't yet refetch the values for variables when a cell is created or edited, and doesn't allow for selecting different values. These will follow in a future PR.

### DIFF
--- a/ui/src/shared/components/cells/Cell.tsx
+++ b/ui/src/shared/components/cells/Cell.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {Component, ComponentClass} from 'react'
+import React, {Component} from 'react'
 import {connect} from 'react-redux'
 import {get} from 'lodash'
 
@@ -8,13 +8,12 @@ import CellHeader from 'src/shared/components/cells/CellHeader'
 import CellContext from 'src/shared/components/cells/CellContext'
 import ViewComponent from 'src/shared/components/cells/View'
 import {ErrorHandling} from 'src/shared/decorators/errors'
-import Conditional from 'src/shared/components/Conditional'
 
-// Actions
-import {getView} from 'src/dashboards/actions/v2/views'
+// Utils
+import {getView} from 'src/dashboards/selectors'
 
 // Types
-import {RemoteDataState, TimeRange} from 'src/types'
+import {TimeRange} from 'src/types'
 import {AppState, ViewType, View, Cell} from 'src/types/v2'
 
 // Styles
@@ -22,14 +21,9 @@ import './Cell.scss'
 
 interface StateProps {
   view: View
-  viewStatus: RemoteDataState
 }
 
-interface DispatchProps {
-  onGetView: typeof getView
-}
-
-interface PassedProps {
+interface OwnProps {
   cell: Cell
   timeRange: TimeRange
   autoRefresh: number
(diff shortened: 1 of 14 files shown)


ui/src/variables/selectors/index.tsx
```diff
@@ -19,19 +25,19 @@ const getVariablesForOrgMemoized = memoizeOne(
...
   state: AppState,
   contextID: string
 ): ValueSelections => {
-  const contextValues: VariableValues = get(
-    state,
-    `variables.values.${contextID}.values`,
-    {}
-  )
+  const contextValues: VariableValues =
+    get(state, `variables.values.${contextID}.values`) || {}
```
bthesorceror: does this sometimes have the key but it is set to null?

chnn: Yep

ui/src/variables/selectors/index.tsx
```diff
@@ -48,3 +54,29 @@ export const getValueSelections = (
...
+  (
+    valuesState: ValuesState,
+    variablesState: VariablesState
+  ): VariableAssignment[] => {
+    if (!valuesState) {
+      return []
+    }
+
+    return Object.entries(valuesState.values).map(([variableID, values]) => {
+      const variableName = variablesState[variableID].variable.name
```
bthesorceror: what do you think about using a `get` here?

ui/src/shared/components/cells/Cell.tsx
```diff
@@ -133,21 +112,17 @@ class CellComponent extends Component<Props> {
   }
 }
 
-const mstp = (state: AppState, ownProps: PassedProps): StateProps => {
-  const entry = state.views[ownProps.cell.id]
+const mstp = (state: AppState, ownProps: OwnProps): StateProps => {
+  const entry = state.views.views[ownProps.cell.id]
```
bthesorceror: naming is weird here `views.views`, but more relevant is maybe we use a `get` here?

chnn: I extracted it to a selector function, and used `get` within the selector.

The naming is weird... it comes up any time we have a slice of Redux state that has loading state:

```
interface ViewsState {
  status: RemoteDataState
  views: {
    [viewID: string]: {
      status: RemoteDataState
      view: View
    }
  }
}
```

This also happens with some variation in the variables, telegrafs, buckets, and tokens state. Consistency would be nice eventually ¯\_(ツ)_/¯

ui/src/variables/selectors/index.tsx
chnn: 👍

PR #12555, opened 2019-03-12
Fix auth permissions

### LINKED ISSUE #12552: Permissions should not match independently from org
An example is worth a 1000 words.

Given this existing permission A:

```
{
    Action: platform.WriteAction,
    Resource: platform.Resource{
        Type: platform.BucketsResourceType,
        ID:   influxdbtesting.IDPtr(1),
    },
}
```

Should the following permission B match it?

```
{
    Action: platform.WriteAction,
    Resource: platform.Resource{
        Type: platform.BucketsResourceType,
        ID:   influxdbtesting.IDPtr(1),
        OrgID: influxdbtesting.IDPtr(10),
    },
}
```

Currently they match but imho they should not since we should always consider the organization a resource belongs to.

We do not need to implement an use case like "allow someone to write to any bucket with ID 1 **disregarding** the org to which it belongs to".

Furthermore as per #12553 permission are not sufficient. They need to be evaluated always in the context of the authorization (and the related org) they belongs to.

### LINKED ISSUE #12553: Permissions should be validated in the context of authorizations
Authorizations contains an organization ID and a list of permissions, to simplify.

Also every permission contains an organization ID.

We should ensure `(a.OrgID, p_i.OrgID, ..., p_n.OrgID)` are always the same ...

This means that to validate permissions we should always consider also the authorization (ie., context) they belong to.

leodido:
Closes #12553 
Closes #12552

Permissions are not sufficient.
They needs to be checked in the context of the authorization they belong to.
This PR also fixes the matching logic of permissions.
It also adds some test for particular cases about authorizations and permissions.

cc @influxdata/api 

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/authorizer/bucket_test.go
+++ b/authorizer/bucket_test.go
@@ -61,8 +61,9 @@ func TestBucketService_FindBucketByID(t *testing.T) {
 				permission: influxdb.Permission{
 					Action: "read",
 					Resource: influxdb.Resource{
-						Type: influxdb.BucketsResourceType,
-						ID:   influxdbtesting.IDPtr(1),
+						Type:  influxdb.BucketsResourceType,
+						OrgID: influxdbtesting.IDPtr(10),
+						ID:    influxdbtesting.IDPtr(1),
 					},
 				},
 				id: 1,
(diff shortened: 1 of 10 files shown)


kelwang:
Hey @leodido

Is this mergeable?

leodido:
Imho it could be. But I'd like to have a review from @goller since we talked about this stuff together some time ago.

PR #12568, opened 2019-03-12
Expose debug pprof port 6060

benbjohnson:
This pull request exposes port `6060` in the `Dockerfile` to allow the port to be accessed when the `http.debug-pprof-enabled` flag is set.


e-dard:
shall we merge this?

PR #12585, opened 2019-03-13
Label names

jademcgough:
Allow label names to be updated.


goller:
Great! I think labels are such an awesome feature of this API.

http/swagger.yml
```diff
@@ -7709,6 +7709,8 @@ components:
     LabelUpdate:
       type: object
       properties:
+        name:
```
OfTheDelmer: Awesome!

PR #12586, opened 2019-03-13
fix: Use simple function name for tracing operation name

jacobmarble:

Every span has an operation name, which shows up in multiple places in the Jaeger UI. Verbose function names like
```
github.com/influxdata/influxdb/kit/tracing.StartSpanFromContext
```
are not readable in this UI. Shorter function names like
```
StartSpanFromContext
```
improve UX. Context is not lost because (1) the filename and line number are logged with the span and (2) spans are understood in the context of other spans, marked with service name and operation name.

Extra cost is ~22 ns per call to StartSpanFromContext. No additional memory is allocated.

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/kit/tracing/tracing.go
+++ b/kit/tracing/tracing.go
@@ -3,11 +3,12 @@ package tracing
 import (
 	"context"
 	"errors"
-	"github.com/opentracing/opentracing-go"
-	"github.com/opentracing/opentracing-go/log"
 	"net/http"
 	"runtime"
 	"strings"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/log"
 )
 
 // LogError adds a span log for an error.
--- a/kit/tracing/tracing.go
+++ b/kit/tracing/tracing.go
@@ -3,11 +3,11 @@ package tracing
 import (
 	"context"
 	"errors"
-	"net/http"
-	"runtime"
-
 	"github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/log"
+	"net/http"
+	"runtime"
+	"strings"
 )
 
 // LogError adds a span log for an error.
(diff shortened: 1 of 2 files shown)


kit/tracing/tracing.go
```diff
@@ -83,6 +83,9 @@ func StartSpanFromContext(ctx context.Context) (opentracing.Span, context.Contex
 	}
 	fn := runtime.FuncForPC(pcs[0])
 	name := fn.Name()
+	if lastPeriod := strings.LastIndexByte(name, '.'); lastPeriod > 0 {
+		name = name[lastPeriod+1:]
+	}
```
benbjohnson: Should we include the package name too (e.g. `LastIndexByte()` on `/` instead)? I'm not sure how much room is in the UI.

kit/tracing/tracing.go
```diff
@@ -3,11 +3,11 @@ package tracing
...
 	"errors"
-	"net/http"
-	"runtime"
-
 	"github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/log"
+	"net/http"
+	"runtime"
+	"strings"
 )
```
stuartcarnie: nit: `goimports`

kit/tracing/tracing.go
jacobmarble: @benbjohnson great question. Here are the specific pain points. @stuartcarnie and I talked about tradeoffs yesterday.

- The first screenshot demonstrates a user experience with no flexibility for string length unless we want to modify the Jaeger UI. Some function names, alone, will overflow the selector.
- The second screenshot demonstrates how spans are contextualized in a trace. Click a span and you get the filename and line number in a log. Because of these extra sources of context, I'm just taking the function name as "operation" for now.

Screen Shot 2019-03-13 at 9 25 12 AM

Screen Shot 2019-03-13 at 9 24 11 AM

jacobmarble: I changed my mind. I think this screenshot demonstrates that there is too much ambiguity in just the function name. I'll follow up with another commit, going to slice from the last `/` rather than the last `.`. It would be cool to clean up the string a bit, but that would cost another allocation to build the string, rather than slicing into an existing string.

Screen Shot 2019-03-13 at 9 41 47 AM

PR #12593, opened 2019-03-13
Ensure when a manual run is called we run it as soon as possible

### LINKED ISSUE #12454: Tasks that are manually run should start almost immediately
I understand that when clicking a manual run of a task, there may be a slight delay, but there is a large delay even when there are very few tasks running. To reproduce, create a task to run every 1m, view the run logs, then manually run the task a few times, and notice that the runs don't kick off for 10 to 15 seconds after manually running them.

kapture 2019-03-08 at 11 09 56

lyondhill:
Closes #12454

### DIFF
--- a/task/backend/coordinator/coordinator.go
+++ b/task/backend/coordinator/coordinator.go
@@ -151,3 +151,15 @@ func (c *Coordinator) DeleteOrg(ctx context.Context, orgID platform.ID) error {
 func (c *Coordinator) CancelRun(ctx context.Context, taskID, runID platform.ID) error {
 	return c.sch.CancelRun(ctx, taskID, runID)
 }
+
+func (c *Coordinator) ManuallyRunTimeRange(ctx context.Context, taskID platform.ID, start, end, requestedAt int64) (*backend.StoreTaskMetaManualRun, error) {
+	r, err := c.Store.ManuallyRunTimeRange(ctx, taskID, start, end, requestedAt)
+	if err != nil {
+		return r, err
+	}
+	t, m, err := c.Store.FindTaskByIDWithMeta(ctx, taskID)
+	if err != nil {
+		return nil, err
+	}
+	return r, c.sch.UpdateTask(t, m)
+}
--- a/task/backend/coordinator/coordinator_test.go
+++ b/task/backend/coordinator/coordinator_test.go
@@ -207,7 +207,6 @@ func TestCoordinator_ManuallyRunTimeRange(t *testing.T) {
 
 	ch := sched.TaskUpdateChan()
 	manualRunTime := time.Now().Unix()
-	// Deleting the task through the coordinator should succeed.
 	if _, err := coord.ManuallyRunTimeRange(context.Background(), id, manualRunTime, manualRunTime, manualRunTime); err != nil {
 		t.Fatal(err)
 	}
--- a/task/backend/coordinator/coordinator_test.go
+++ b/task/backend/coordinator/coordinator_test.go
@@ -192,3 +192,40 @@ func TestCoordinator_ClaimExistingTasks(t *testing.T) {
 		}
 	}
 }
+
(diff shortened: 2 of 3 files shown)


task/backend/coordinator/coordinator.go
```diff
@@ -151,3 +151,15 @@ func (c *Coordinator) DeleteOrg(ctx context.Context, orgID platform.ID) error {
...
+func (c *Coordinator) ManuallyRunTimeRange(ctx context.Context, taskID platform.ID, start, end, requestedAt int64) (*backend.StoreTaskMetaManualRun, error) {
+	r, err := c.Store.ManuallyRunTimeRange(ctx, taskID, start, end, requestedAt)
+	if err != nil {
+		return r, err
+	}
+	t, m, err := c.Store.FindTaskByIDWithMeta(ctx, taskID)
+	if err != nil {
+		return nil, err
+	}
+	return r, c.sch.UpdateTask(t, m)
```
mark-rushakoff: I see that calling UpdateTask cancels in-flight work:

https://github.com/influxdata/influxdb/blob/ab2fd3ba2fee091c70c5988da784a7497c1c4942/task/backend/scheduler.go#L335-L339

IIRC a canceled run won't be restarted. What do you think about adding a new method on the `TickScheduler`, like `CheckManualRuns(taskID platform.ID)`? That would need to call `taskScheduler.Start` and make sure that the `taskScheduler.hasQueue` is set to true so it knows to call `DesiredState.CreateNextRun`.

task/backend/coordinator/coordinator_test.go
```diff
@@ -192,3 +192,40 @@ func TestCoordinator_ClaimExistingTasks(t *testing.T) {
...
+
+	// Create an isolated task directly through the store so the coordinator doesn't know about it.
+	id, err := coord.CreateTask(context.Background(), backend.CreateTaskRequest{Org: 1, AuthorizationID: 3, Script: script})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ch := sched.TaskUpdateChan()
+	manualRunTime := time.Now().Unix()
+	// Deleting the task through the coordinator should succeed.
```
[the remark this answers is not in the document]
lyondhill: OO yep. Copy Paste error!

task/backend/coordinator/coordinator.go
lyondhill: Seems reasonable. Do you think it should be added to the Scheduler interface or just the implementation of the TickScheduler? Im leaning toward the tick scheduler, only because this manual run problem is a consequence of the tick scheduler's caching of the run data. Thoughts?

lyondhill: or maybe the update task shouldnt cancel work in progress. It seems like any work in progress should proceed and any new work should get the update?

mark-rushakoff: I vaguely recall that we consciously decided that Update should cancel in progress work, because we can't make assumptions about what changed during an update. 

Without playing with the code, it's hard for me to say whether just updating TickScheduler will work out cleanly as opposed to updating the Scheduler interface too. I don't know if there's enterprise code which refers to the Scheduler interface, and which would need to know about this new method.

But if it works out fine just updating TickScheduler, I'm okay with that approach.

lyondhill: After Looking back at the reason why we replaced the task scheduler (https://github.com/influxdata/platform/pull/1058). Im now sure we wrote it to replace the task scheduler as a quick solution not because of any specific intent. I think It would be quite simple to update the `UpdateTask` to not replace but instead update.

mark-rushakoff: Okay, fine by me.

docmerlin: I prefer this behavior to canceling.

PR #12610, opened 2019-03-13
Fine tune inline label keyboard interactions

### LINKED ISSUE #12609: Fix Keyboard interactions when using inline labels
- There should always be a "selected item" in the list so long as there is a search term
- `ArrowUp` and `ArrowDown` should change the selected item
- `Enter` should take action on the visibly selected item
- If previously selected item is no longer in the list then select the first item available

alexpaxton:
Closes #12609 

- Text field in `Add Label` popover is always focused (to receive keyboard commands)
- There is always a selected item in the list (so that hitting `Enter` always does something)
- Hitting `ArrowUp`/`ArrowDown` changes the selected item in the list
- Abiity to select the suggestion via keyboard shortcuts
- Selected item is much brighter
- Wrote unit tests for every mouse & keyboard interaction associated with inline labels

inline-labels-keyboard-interactions
inline-labels-keyboard-empty

### Checklist

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/shared/components/inlineLabels/InlineLabelsEditor.tsx
+++ b/ui/src/shared/components/inlineLabels/InlineLabelsEditor.tsx
@@ -21,7 +21,7 @@ import {OverlayState} from 'src/types/overlay'
 
 // Constants
 export const ADD_NEW_LABEL_ITEM_ID = 'add-new-label'
-const ADD_NEW_LABEL_LABEL: ILabel = {
+export const ADD_NEW_LABEL_LABEL: ILabel = {
   id: ADD_NEW_LABEL_ITEM_ID,
   name: '',
   properties: {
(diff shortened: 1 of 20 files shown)


ui/src/shared/components/inlineLabels/InlineLabelsEditor.tsx
```diff
@@ -160,35 +182,63 @@ class InlineLabelsEditor extends Component<Props, State> {
...
-    const {availableLabels} = this
-    let selectedItemID = this.state.selectedItemID
     const searchTerm = e.target.value
+    const filteredLabels = this.filterLabels(searchTerm)
 
-    const selectedItemIDAvailable = availableLabels.find(
-      al => al.name === selectedItemID
-    )
+    if (filteredLabels.length) {
+      const selectedItemID = filteredLabels[0].id
```
bthesorceror: It always feels weird when I see code that just selects the first on from the list, there is no clear definition as to why the first element gets special status over the whole list.

PR #12611, opened 2019-03-13
Add variables to configuration page and navigation for config tabs

### LINKED ISSUE #12255: Add Variables tab to config page
We are moving any kind of configuration workflows into the config page since that's where users tend to look first

- [ ] Create `Variables` tab within `Configuration` page
- [ ] Create `Variables` tab within `Configuration` navbar tab
- [ ] List variables across all orgs
- [ ] Add `Owner` column to table
- [ ] Variables page in `Configuration` should otherwise look the same as the org level variables page

Palakp41:
Closes #12255 

  - [x] Rebased/mergeable
  - [x] Tests pass

Kapture 2019-03-13 at 14 02 09

### DIFF
--- a/ui/src/configuration/components/Variables.tsx
+++ b/ui/src/configuration/components/Variables.tsx
@@ -1,0 +1,199 @@
+// Libraries
+import React, {PureComponent, ChangeEvent} from 'react'
+import _ from 'lodash'
+import {connect} from 'react-redux'
+
+// Utils
+import {
+  getVariables,
+  createVariable,
+  updateVariable,
+  deleteVariable,
+} from 'src/variables/actions'
+
+// Components
+import TabbedPageHeader from 'src/shared/components/tabbed_page/TabbedPageHeader'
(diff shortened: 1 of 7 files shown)


ui/src/configuration/components/Variables.tsx
```diff
@@ -0,0 +1,200 @@
...
+          {v => (
+            <VariableList
+              variables={v}
+              emptyState={this.emptyState}
+              onDeleteVariable={this.handleDeleteVariable}
+              onUpdateVariable={this.handleUpdateVariable}
+            />
+          )}
+        </FilterList>
+        <Overlay visible={createOverlayState === OverlayState.Open}>
```
bthesorceror: Lets move this overlay component into the CreateVariableOverlay and have that take a visible prop; this will make it easier to move this into the router.

Palakp41: Got it, I added it as the prop

PR #12614, opened 2019-03-13
ci: fail build if binary files committed

mark-rushakoff:
Binary files can be whitelisted with the special line 'Adds-Binary:
<path/to/binary/file>' included in the commit message body. Deliberately
checked-in binary files should be very rare.

With minimal effort, this could probably be repurposed as a local
postcommit hook.


mark-rushakoff:
Example failure due to introduced binary files: https://github.com/influxdata/influxdb/pull/12617

PR #12616, opened 2019-03-13
DO NOT MERGE: testing CI binary check

mark-rushakoff:
Testing #12614.


mark-rushakoff:
I think Circle isn't cooperating with this one since I originally opened it based on master, not the CI branch I'm trying to test, by mistake. Changing the base isn't doing what I expect.

PR #12618, opened 2019-03-14
VEO uses react router

bthesorceror:
VEO use react router for rendering and moves view retrieving into the VEO.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/components/VEO.tsx
+++ b/ui/src/dashboards/components/VEO.tsx
@@ -1,6 +1,10 @@
 // Libraries
 import React, {PureComponent} from 'react'
+import {withRouter, WithRouterProps} from 'react-router'
 import {connect} from 'react-redux'
+import _ from 'lodash'
+
+import {Overlay} from 'src/clockface'
 
 // Components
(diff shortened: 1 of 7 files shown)


ui/src/dashboards/components/VEO.tsx
```diff
@@ -1,87 +1,153 @@
...
+  const {cellID, dashboardID} = params
+  const {
+    views: {views},
+    dashboards,
+  } = state
   const {view, draftQueries} = getActiveTimeMachine(state)
+  const dashboard = dashboards.find(d => d.id === dashboardID)
+  let existingView = null
+  if (cellID) {
+    existingView = (views[cellID] || {view: null}).view
```
ebb-tide: can this be views[cellID].view || null instead? hard to read

bthesorceror: no because if `views[cellID]` returns null then `views[cellID].view` will throw an error

ebb-tide: hrmm _.get?

PR #12619, opened 2019-03-14
Migrate TimeSeries state into Redux

### LINKED ISSUE #12613: Migrate TimeSeries state into Redux in DE/CEO
The root of the `TimeMachine` tree [contains a `TimeSeries` component](https://github.com/influxdata/influxdb/blob/8ef815943d6416ef898d6359fa53384936449463/ui/src/timeMachine/components/TimeMachine.tsx#L82-L87). This component handles fetching time series data used in the `TimeMachine`, and stores the loading state for the fetch as well.

This state is widely used throughout the `TimeMachine`, but most other state in the `TimeMachine` is stored in Redux. This causes a few problems.

- Fetching time series data is now a 2-step process: first fetch the variables (happens outside of the `TimeSeries` component / happens in Redux), then execute the queries with the variables substituted (happens inside the `TimeSeries` component). It's awkward to display an appropriate loading indicator since the location of that state is split up.
- Excessive prop drilling 
- Some [awkward two-way data binding](https://github.com/influxdata/influxdb/blob/8ef815943d6416ef898d6359fa53384936449463/ui/src/shared/components/Histogram.tsx#L87-L91) that has lead to [defensive programming](https://github.com/influxdata/influxdb/blob/8ef815943d6416ef898d6359fa53384936449463/ui/src/shared/components/Histogram.tsx#L44-L50)

The aim of this refactor would be to move all state from the `TimeSeries` component into Redux and all logic into a thunk. This would only apply to the `TimeSeries` component as it is used in the `TimeMachine`; other usages of the `TimeSeries` component would remain as is.

chnn:
Closes #12613


ui/src/timeMachine/actions/queries.ts
```diff
@@ -0,0 +1,129 @@
...
+      executeQueryWithVars(queryURL, orgID, text, variableAssignments)
+    )
+
+    const results = await Promise.all(pendingResults.map(r => r.promise))
+
+    const duration = Date.now() - startTime
+
+    const files = results.map(r => r.csv)
+
+    dispatch(setQueryResults(RemoteDataState.Done, files, duration))
```
OfTheDelmer: Nice

ui/src/timeMachine/selectors/index.ts
```diff
@@ -1,3 +1,12 @@
+// Libraries
+import memoizeOne from 'memoize-one'
+import {flatten} from 'lodash'
```
OfTheDelmer:
```suggestion
import {flatMap} from 'lodash'
```

chnn: hell ya 😎

ui/src/timeMachine/actions/queries.ts
```diff
@@ -0,0 +1,142 @@
...
+  } catch (e) {
+    if (e instanceof CancellationError) {
+      return
+    }
+
+    console.log(e)
```
mavarius: Should this error or warn instead of log?

chnn: ah yep, good catch

PR #12621, opened 2019-03-14
Fix dashboards in org

### LINKED ISSUE #12574: Dashboards in org view should only the be the dashboards for that org
Currently dashboards in the org view include all dashboards rather than just those in that org.

ebb-tide:
Closes #12574 

Org view dashboards were loading dashboards in a number of places in the component hierarchy and there was a lot of general confusion in the components. 
I did a lot of clean up, as well as centralizing where dashboard get was being performed.

### DIFF
--- a/ui/src/dashboards/components/dashboard_index/DashboardsIndex.tsx
+++ b/ui/src/dashboards/components/dashboard_index/DashboardsIndex.tsx
@@ -20,16 +20,12 @@ import {
   addDashboardLabelsAsync,
   removeDashboardLabelsAsync,
 } from 'src/dashboards/actions/v2'
-import {setDefaultDashboard} from 'src/shared/actions/links'
 import {retainRangesDashTimeV1 as retainRangesDashTimeV1Action} from 'src/dashboards/actions/v2/ranges'
 import {notify as notifyAction} from 'src/shared/actions/notifications'
 
 // Constants
 import {DEFAULT_DASHBOARD_NAME} from 'src/dashboards/constants/index'
-import {
-  dashboardSetDefaultFailed,
-  dashboardCreateFailed,
-} from 'src/shared/copy/notifications'
+import {dashboardCreateFailed} from 'src/shared/copy/notifications'
 
 // Types
 import {Notification} from 'src/types/notifications'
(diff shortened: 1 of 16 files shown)


ui/src/dashboards/components/dashboard_index/DashboardsIndex.tsx
```diff
@@ -132,20 +124,6 @@ class DashboardIndex extends PureComponent<Props, State> {
     )
   }
 
-  private handleSetDefaultDashboard = async (
```
chnn: 🔥

ui/src/organizations/actions/orgView.ts
```diff
@@ -35,6 +35,23 @@ export const getTasks = (orgID: string) => async dispatch => {
   dispatch(populateTasks(tasks))
 }
 
+export interface PopulateDashboards {
+  type: ActionTypes.PopulateDashboards
```
chnn: Should we use the top level dashboards reducer instead to do this? It seems unfortunate that there would be two copies of the same dashboard in Redux (one from the dashboards reducer, and one from the org view reducer).

ui/src/organizations/containers/OrgDashboardsIndex.tsx
```diff
@@ -33,73 +34,110 @@ interface RouterProps {
...
+  public state = {
+    loadingState: RemoteDataState.NotStarted,
+  }
+
+  public componentDidMount = async () => {
+    this.setState({loadingState: RemoteDataState.Loading})
+
+    const {getDashboards, org} = this.props
+
+    await getDashboards(org.id)
```
chnn: What do you think about putting all this fetching logic in a separate data-fetching component?

ui/src/organizations/actions/orgView.ts
ebb-tide: I might agree but wanted to continue with existing pattern, would be easy to switch

ui/src/organizations/containers/OrgDashboardsIndex.tsx
ebb-tide: I will look in to this idea, but won't do it in this PR

PR #12623, opened 2019-03-14
Death to v2 folders dashboards

ebb-tide:
Some v2 folders are benign, and just pose extra nesting, others come with legacy non v2 versions, of very very old code.


ui/src/dashboards/apis/index.ts
```diff
@@ -1,102 +1,161 @@
...
-    console.error(error)
-    throw error
+export const createDashboard = async (
+  props: CreateDashboardRequest
+): Promise<Dashboard> => {
+  const dashboard = await client.dashboards.create(props)
+
+  return {
+    ...dashboard,
+    labels: dashboard.labels.map(addLabelDefaults),
```
ischolten: not for this pr, but perhaps we should consider doing this or having a function for this in the client library so we dont have to implement it everywhere.

ui/src/dashboards/apis/index.ts
```diff
@@ -1,102 +1,161 @@
...
+  await client.dashboards.deleteCell(dashboardID, cell.id)
+}
+
+export const addDashboardLabels = async (
+  dashboardID: string,
+  labels: ILabel[]
+): Promise<ILabel[]> => {
+  const addedLabels = await Promise.all(
+    labels.map(async label => {
+      return client.dashboards.addLabel(dashboardID, label.id)
```
ischolten: i believe there is already a `client.dashboards.addLabels` that takes in a list  of label ids and does this. how do you feel about using that function rather than reimplementing it here ?

ui/src/dashboards/apis/index.ts
```diff
@@ -1,102 +1,161 @@
...
-      method: 'PUT',
-      url: templateVariable.links.self,
-      data: templateVariable,
+export const removeDashboardLabels = async (
+  dashboardID: string,
+  labels: ILabel[]
+): Promise<void> => {
+  await Promise.all(
+    labels.map(label => {
+      return client.dashboards.removeLabel(dashboardID, label.id)
```
ischolten: i believe the same for this one, `removeLabels`, (if its not implemented i think thats an oversight and could be added)

121watts:
yaas!

ui/src/dashboards/apis/index.ts
ebb-tide: will do in next PR

ui/src/dashboards/apis/index.ts
ebb-tide: will do in next PR

ui/src/dashboards/apis/index.ts
ebb-tide: will do in next PR

PR #12628, opened 2019-03-14
feat(query/mock): create a mock object for the query interface

jsternberg:


jsternberg:
This is an implementation of `flux.Query` which is the asynchronous one. I think you're thinking of `flux.ResultIterator` which is the synchronous version of this interface.

Issue #10758, opened 2018-05-30
Break out each Read physical operation into its own RPC call


nathanielc:
Currently we have a Read call with various parameters. Not all combinations of parametes are valid, and certain combinations require that the read service behave differently. 

By explicitly exposing those individual physical operations we prevent invalid combinations, expose to the planner the cost of each operation, protect against breaking existing behavior and separate concerns between the client and the server.

jlapacik:
Each physical read operation now has its own RPC call, so closing this issue.

PR #12651, opened 2019-03-15
[WIP] feat(query/stdlib/influxdb): rule to push window to storage

### LINKED ISSUE #12487: Explore pushing down fixed windows to storage
The storage source is capable of making many requests to storage for fixed windows of time. This means we can push down window transformations in the cases that the windows are fixed length.

This may improve performance because we can emit watermark events more frequently and therefore flush data through transformations in small chunks.

affo:
Closes #12487 

Adds a rule to merge `from` with `window`.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/query/stdlib/influxdata/influxdb/from.go
+++ b/query/stdlib/influxdata/influxdb/from.go
@@ -43,6 +43,7 @@ func init() {
 		FromDistinctRule{},
 		MergeFromGroupRule{},
 		FromKeysRule{},
+		MergeFromWindowRule{},
 	)
 	execute.RegisterSource(PhysicalFromKind, createFromSource)
 }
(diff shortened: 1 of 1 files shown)


affo:
This is WIP.

The merge of operations produces a correct behavior, because the code is already backed to accomodate this use case: 
 - here: https://github.com/influxdata/influxdb/blob/3abb66ebb3052527149c27a2f730a8f73cc167a4/query/stdlib/influxdata/influxdb/from.go#L732 the `from` source is created with the `window` injected by the custom rule;
 - here: https://github.com/influxdata/influxdb/blob/9549cc4d97b50aba8b886b9f3dd56daa9d4b188c/query/stdlib/influxdata/influxdb/storage.go#L117 the source repeatedly calls `next` to get tables and propagate them to downstream transformations;
 - here: https://github.com/influxdata/influxdb/blob/9549cc4d97b50aba8b886b9f3dd56daa9d4b188c/query/stdlib/influxdata/influxdb/storage.go#L149 `next` uses the `window` in the source to issue the correct RPC call to storage.

The bad behavior right now is that, if there are holes on the time axis, RPC calls for empty windows are performed. So if in e2e tests we have:

```
option now = () => (2018-05-22T19:59:00Z)

inData = "
#datatype,string,long,dateTime:RFC3339,double,string,string,string
#group,false,false,false,false,true,true,true
#default,_result,,,,,,
,result,table,_time,_value,_field,_measurement,host
,,0,2018-05-22T19:53:26Z,63.053321838378906,used_percent,mem,host.local
,,0,2018-05-22T19:54:36Z,62.71536350250244,used_percent,mem,host.local
,,0,2018-05-22T19:55:46Z,62.38760948181152,used_percent,mem,host.local
,,0,2018-05-22T19:56:56Z,62.74595260620117,used_percent,mem,host.local
,,0,2018-05-22T19:57:06Z,62.78183460235596,used_percent,mem,host.local
,,0,2018-05-22T19:58:16Z,62.46745586395264,used_percent,mem,host.local
"

// a test that windows every minute
...
```

we get around a million RPC calls (1 minute for 2 years) to get empty windows. We should prevent this bad behavior by knowing in advance the holes and skip RPC calls for empty windows. IDK if that is possible.

affo:
At the moment, actually, tests fail for wrong results. @jlapacik and I were more worried about the slow outcome than correctness of e2e tests overall. We only focused on `window_start_bound.flux` test that produced only a column order mismatch. We should also further investigate why other results are different then expected.

wolffcm:
One thing that concerns me about this change is that there are now several operations that can be pushed down into storage:
- range
- filter
- group
- window
- (in the future aggregators or selectors can go here too)

Storage must evaluate these 4 things in some static order, but we don't push down operations in any particular order in the planner.  This means we are changing the order of operations during execution, but in a sneaky way that could cause wrong answers that are hard to understand.

Maybe in most cases these 4 operations can be evaluated in any order and get the same answer, but it won't be universally true.  This is why I'd like to logical planning to come up with a canonical ordering of operations, and model our gRPC calls around that as described here:
https://github.com/influxdata/influxdb/issues/10758

I understand the need to improve performance in the short term, and maybe this change is okay.  I haven't thought it all through.

Really curious about the wrong answers that the e2e tests are finding!  Glad that they're working now.  :)

affo:
This PR depends on having a proper stack for applying pushed-down transformations (as @wolffcm mentioned:  https://github.com/influxdata/influxdb/issues/10758).
 
And on cost-based optimizations: the planner should know the cost of pushing down a `window` transformation if that brings to performing too many RPC calls. However, our rules are always positive, and not cost based.

We should wait for modeling cost-based rules in the planner and the pushed-down operation stack in InfluxDB before working again on this feature.

PR #12655, opened 2019-03-15
Dashboards/variables toolbar

### LINKED ISSUE #12375: In a dashboard, show a list of dropdowns with the variables used in the cells in that dashboard
Screen Shot 2019-03-08 at 3 02 45 PM

Each dropdown in the list allows the user to set the value for that variable.

AlirieGray:
Closes #12375

This PR adds a new toolbar to the dashboard with dropdowns for all of the variables used in that dashboard. The dropdowns allow the user to select the value that will be used for a given variable everywhere that variable is used on the dashboard they are viewing.

The toolbar can be shown or hidden with a button in the Dashboard Header.

  - [x] Rebased/mergeable
  - [x] Tests pass
  - [x] CHANGELOG.md updated

### DIFF
--- a/ui/src/clockface/components/empty_state/EmptyState.tsx
+++ b/ui/src/clockface/components/empty_state/EmptyState.tsx
@@ -1,5 +1,6 @@
 // Libraries
 import React, {Component} from 'react'
+import classnames from 'classnames'
 
 // Components
 import EmptyStateText from 'src/clockface/components/empty_state/EmptyStateText'
@@ -16,6 +17,7 @@ import {ErrorHandling} from 'src/shared/decorators/errors'
 
 interface PassedProps {
   children: JSX.Element | JSX.Element[]
+  customClass?: string
 }
 
 interface DefaultProps {
@@ -36,15 +38,22 @@ class EmptyState extends Component<Props> {
   public static SubText = EmptyStateSubText
 
   public render() {
-    const {children, size, testID} = this.props
-
-    const className = `empty-state empty-state--${size}`
+    const {children, testID} = this.props
 
     return (
-      <div className={className} data-testid={testID}>
+      <div className={this.className} data-testid={testID}>
         {children}
       </div>
     )
+  }
+
+  private get className(): string {
+    const {customClass, size} = this.props
+
+    return classnames('empty-state', {
+      [`empty-state--${size}`]: size,
+      [customClass]: customClass,
+    })
   }
 }
 
--- a/ui/src/dashboards/components/variablesControlBar/VariableDropdown.tsx
+++ b/ui/src/dashboards/components/variablesControlBar/VariableDropdown.tsx
@@ -1,0 +1,93 @@
+// Libraries
+import React, {PureComponent} from 'react'
+import {connect} from 'react-redux'
+import _ from 'lodash'
+
+// Components
+import {Dropdown, DropdownMenuColors} from 'src/clockface'
+
+// Utils
+import {
+  getValuesForVariable,
+  getSelectedValueForVariable,
+} from 'src/variables/selectors'
+
+// Styles
+import 'src/dashboards/components/variablesControlBar/VariableDropdown.scss'
+
+// Types
+import {AppState} from 'src/types/v2'
+
+interface StateProps {
+  values: string[]
+  selectedValue: string
+}
+
+interface OwnProps {
+  name: string
+  variableID: string
+  dashboardID: string
+  onSelect: (variableID: string, value: string) => void
+}
+
+type Props = StateProps & OwnProps
+
+class VariableDropdown extends PureComponent<Props> {
+  render() {
+    const {name, selectedValue} = this.props
+
+    return (
+      <div className="variable-dropdown">
+        {/* TODO: Add variable description to title attribute when it is ready */}
+        <div className="variable-dropdown--label">
+          <span>{name}</span>
+        </div>
+        <Dropdown
+          selectedID={selectedValue}
+          onChange={this.handleSelect}
+          widthPixels={140}
+          customClass="variable-dropdown--dropdown"
+          menuColor={DropdownMenuColors.Amethyst}
+        >
+          {this.dropdownItems}
+        </Dropdown>
+      </div>
+    )
+  }
+
+  private get dropdownItems(): JSX.Element[] {
+    const {values} = this.props
+
+    return values.map(v => {
+      return (
+        <Dropdown.Item key={v} id={v} value={v}>
+          {v}
+        </Dropdown.Item>
+      )
+    })
+  }
+
+  private handleSelect = (value: string) => {
+    const {variableID, onSelect} = this.props
+    onSelect(variableID, value)
+  }
+}
+
+const mstp = (state: AppState, props: OwnProps): StateProps => {
+  const {dashboardID, variableID} = props
+
+  const values = getValuesForVariable(state, variableID, dashboardID)
+
+  const selectedValue = getSelectedValueForVariable(
+    state,
+    variableID,
+    dashboardID
+  )
+
+  return {values, selectedValue}
+}
+
+export default connect<StateProps, {}, OwnProps>(
+  mstp,
+  null
+)(VariableDropdown)
--- a/ui/src/dashboards/components/variablesControlBar/VariablesControlBar.tsx
+++ b/ui/src/dashboards/components/variablesControlBar/VariablesControlBar.tsx
@@ -1,0 +1,92 @@
+// Libraries
+import React, {PureComponent} from 'react'
+import {connect} from 'react-redux'
+import _ from 'lodash'
+
+// Components
+import VariableDropdown from 'src/dashboards/components/variablesControlBar/VariableDropdown'
+import {EmptyState, ComponentSize} from 'src/clockface'
+
+// Utils
+import {getVariablesForDashboard} from 'src/variables/selectors'
+
+// Styles
+import 'src/dashboards/components/variablesControlBar/VariablesControlBar.scss'
+
+// Actions
+import {selectValue} from 'src/variables/actions'
+
+// Types
+import {AppState} from 'src/types/v2'
+import {Variable} from '@influxdata/influx'
+
+// Decorators
+import {ErrorHandling} from 'src/shared/decorators/errors'
+
+interface OwnProps {
+  dashboardID: string
+}
+
+interface StateProps {
+  variables: Variable[]
+}
+
+interface DispatchProps {
+  selectValue: typeof selectValue
+}
+
+type Props = StateProps & DispatchProps & OwnProps
+
+@ErrorHandling
+class VariablesControlBar extends PureComponent<Props> {
+  render() {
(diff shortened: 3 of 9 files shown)


ui/src/clockface/components/empty_state/EmptyState.tsx
```diff
@@ -16,6 +16,7 @@ import {ErrorHandling} from 'src/shared/decorators/errors'
 
 interface PassedProps {
   children: JSX.Element | JSX.Element[]
+  className?: string
```
chnn: I think most of the other clockface components are using `customClass` for a prop like this

ui/src/dashboards/components/variablesControlBar/VariablesControlBar.tsx
```diff
@@ -0,0 +1,129 @@
...
+interface DispatchProps {
+  selectValue: typeof selectValue
+}
+
+type Props = StateProps & DispatchProps & OwnProps
+
+@ErrorHandling
+class VariablesControlBar extends PureComponent<Props> {
+  render() {
+    // return <div className="variables-control-bar" />
```
chnn: delete me!

ui/src/dashboards/components/variablesControlBar/VariablesControlBar.tsx
```diff
@@ -0,0 +1,129 @@
...
+          >
+            <EmptyState.Text text="To see variable controls here, use a variable in a cell query" />
+          </EmptyState>
+        </div>
+      )
+    }
+
+    return (
+      <div className="variables-control-bar">
+        {Object.keys(this.variablesForDashboard).map(variableID => {
```
chnn: How about just passing only the `id` of the variable to the `VariableDropdown`, and letting the `VariableDropdown` read its own state from Redux?

ui/src/dashboards/components/variablesControlBar/VariablesControlBar.tsx
```diff
@@ -0,0 +1,129 @@
...
+    )
+  }
+
+  private get variablesForDashboard() {
+    const {variables} = this.props.variables
+
+    const variableIDs = Object.keys(this.valuesForDashboard)
+    let variablesForDash = {}
+
+    Object.keys(variables).forEach(variableID => {
```
chnn: I think this stuff would belong better in a "selector" utility function. A selector is a function that reads Redux state and transforms/filters it into a nice format for a component to use. Selectors get used in the `mstp` for a component. That way the component stays focused only on presentation, and the selection logic can be reused across components.

[Here](https://github.com/influxdata/influxdb/blob/ccc8165b168d7db823091ed0d02f61a4a27a70ca/ui/src/timeMachine/selectors/index.ts#L13-L18) is an example of a simple selector and [where it gets used](https://github.com/influxdata/influxdb/blob/ccc8165b168d7db823091ed0d02f61a4a27a70ca/ui/src/timeMachine/components/Queries.tsx#L128-L130).

A lot of people also [memoize selectors](https://github.com/reduxjs/reselect) for performance reasons, but it's probably a detail not worth bothering with at the moment.

ui/src/dashboards/components/variablesControlBar/VariablesControlBar.tsx
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    return (
+      <div className="variables-control-bar">
+        {Object.keys(this.variablesForDashboard).map(variableID => {
+          const initialSelected = _.get(
+            this.valuesForDashboard,
+            `${variableID}.selectedValue`
+          )
+          const name = _.get(
```
chnn: I commented below about selectors... I think this is also a good example of what could be a selector that gets used within the `mstp` for the `VariableDropdown`.

ui/src/variables/actions/index.ts
```diff
@@ -69,6 +74,20 @@ const removeVariable = (id: string): RemoveVariable => ({
   payload: {id},
 })
 
+interface SelectValue {
```
chnn: heh I just wrote something very similar. Let's merge yours first and then I'll consolidate after merging my PR.

ui/src/variables/reducers/index.ts
```diff
@@ -78,6 +78,13 @@ export const variablesReducer = (
         return
       }
 
+      case 'SELECT_VALUE': {
+        const {contextID, id, value} = action.payload
+        draftState.values[contextID].values[id].selectedValue = value
```
chnn: We might want to check that this path exists before trying to set it (e.g. that `values[contextID]` exists and `values[id]` exists and so on).

ui/src/dashboards/components/variablesControlBar/VariableDropdown.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+// Styles
+import 'src/dashboards/components/variablesControlBar/VariableDropdown.scss'
+
+interface Props {
+  name: string
+  variableID: string
+  values: string[] //todo
+  initialSelected: string // todo
+  onSelect: (variableID: string, value: string) => void // todo
```
chnn: `todo`?

AlirieGray: oh yeah, I just wanted to check with you that the values should be strings

ui/src/variables/reducers/index.ts
AlirieGray: would something like `if (!!draftState.values[contextID].values[id].selectedValue)` work?

ui/src/dashboards/components/variablesControlBar/VariableDropdown.tsx
chnn: oh, yep!

ui/src/variables/reducers/index.ts
chnn: I think that will still throw an error if the `values[contextID]` part doesn't exist, or if the `values[id]` part doesn't exist (would see something like `Error: can't get selectedValue of null`)

How about:

```
        const {contextID, id, selectedValue} = action.payload

        const valuesExist = !!get(
          draftState,
          `values.${contextID}.values.${id}`
        )

        if (!valuesExist) {
          return
        }

        draftState.values[contextID].values[id].selectedValue = selectedValue
```
since what

mavarius: This is a pretty long chain, is there any point where one of these 6 items could be undefined?

AlirieGray: good point

ui/src/dashboards/components/variablesControlBar/VariablesControlBar.tsx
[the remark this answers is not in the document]
AlirieGray: agreed

ui/src/clockface/components/empty_state/EmptyState.tsx
```diff
@@ -36,12 +37,14 @@ class EmptyState extends Component<Props> {
   public static SubText = EmptyStateSubText
 
   public render() {
-    const {children, size, testID} = this.props
+    const {children, size, testID, className} = this.props
 
-    const className = `empty-state empty-state--${size}`
+    const classname = className
```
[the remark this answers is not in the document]
AlirieGray: is there a good example of this pattern with a `customClass` property?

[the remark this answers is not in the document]
mavarius: `IndexListRow.tsx` is a good example.

ui/src/dashboards/components/variablesControlBar/VariableDropdown.tsx
```diff
@@ -0,0 +1,70 @@
...
+import React, {PureComponent} from 'react'
+import {Dropdown, DropdownMenuColors} from 'src/clockface'
+
+// Styles
+import 'src/dashboards/components/variablesControlBar/VariableDropdown.scss'
+
+interface Props {
+  name: string
+  variableID: string
+  values: string[] //todo
```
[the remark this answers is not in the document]
AlirieGray: good catch

ui/src/variables/selectors/index.tsx
```diff
@@ -32,6 +32,48 @@ export const getVariablesForOrg = (
...
+): Variable[] => {
+  const {
+    variables: {variables, values},
+  } = state
+
+  let variablesForDash = []
+
+  const variablesIDs = Object.keys(get(values, `${dashboardID}.values`))
+
+  variablesIDs.forEach(variableID => {
```
chnn: I think this might supposed to be:

```
  variablesIDs.forEach(variableID => {
    const variable = get(variables, `${variableID}.variable`)

    if (variable) {
      variablesForDash.push(variable)
    }
  })
```

ui/src/dashboards/components/variablesControlBar/VariableDropdown.tsx
```diff
@@ -0,0 +1,97 @@
...
+// Types
+import {AppState} from 'src/types/v2'
+
+interface StateProps {
+  values: string[]
+  selectedValue: string
+}
+
+interface OwnProps {
+  name: string
```
chnn: This and `onSelect` could come from Redux too (but it's also fine like this)

ui/src/dashboards/components/variablesControlBar/VariableDropdown.tsx
```diff
@@ -0,0 +1,97 @@
...
+  variableID: string
+  dashboardID: string
+  onSelect: (variableID: string, value: string) => void
+}
+
+type Props = StateProps & OwnProps
+
+class VariableDropdown extends PureComponent<Props> {
+  constructor(props) {
```
chnn: delete me!

ui/src/variables/selectors/index.tsx
```diff
@@ -32,6 +32,48 @@ export const getVariablesForOrg = (
...
+): string[] => {
+  const {variables} = state
+  return get(variables, `values.${contextID}.values.${variableID}.values`)
+}
+
+export const getSelectedValueForVariable = (
+  state: AppState,
+  variableID: string,
+  contextID: string
+): string => {
```
chnn: nice

ui/src/dashboards/components/variablesControlBar/VariableDropdown.tsx
AlirieGray: I thought of that but I thought a utility was a lot for something we already have peeled off in the parent

chnn: there's already a selector that could be useful: 

https://github.com/influxdata/influxdb/blob/243f1ea5112058b74f838d0eae8da8d03f4795b7/ui/src/variables/selectors/index.tsx#L115-L117

then `mstp` could look something like
```
const mstp = (state: AppState, props: OwnProps): StateProps => {
  const {variableID} = props

  const {name} = getVariable(state, variableID)

  return {name}
}
```

just as an example for how selectors can be reused.... but this is totally down the weeds and nitpicky, just wanted to share. it's totally cool as is

PR #12657, opened 2019-03-15
Add a task control service to facilitate a new structure for tasks

lyondhill:


task/backend/task.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/influxdata/influxdb"
+)
+
+// TaskControlService is a task service with control functions.
+type TaskControlService interface {
+
```
mark-rushakoff: Remove blank line.

task/backend/task.go
```diff
@@ -0,0 +1,31 @@
...
+	// NextDueRun returns the Unix timestamp of when the next call to CreateNextRun will be ready.
+	// The returned timestamp reflects the task's delay, so it does not necessarily exactly match the schedule time.
+	NextDueRun() (int64, error)
+
+	// UpdateRunState sets the run state and the respective time.
+	UpdateRunState(ctx context.Context, taskID, runID influxdb.ID, when time.Time, state RunStatus) error
+
+	// AddRunLog adds a log line to the run.
+	AddRunLog(ctx context.Context, taskID, runID influxdb.ID, when time.Time, log string) error
```
mark-rushakoff: For UpdateRunState and AddRunLog, there probably should be some documentation of these logs not being written to the storage layer until FinishRun is called and succeeds... which raises the question of what to do with those logs if the call to FinishRun fails.

task/backend/task.go
```diff
@@ -0,0 +1,31 @@
+package backend
+
+import (
+	"context"
+	"time"
+
+	"github.com/influxdata/influxdb"
+)
+
+// TaskControlService is a task service with control functions.
```
mark-rushakoff: This doesn't really inform the reader much about the type. Maybe something like:

> TaskControlService is a low-level controller interface, intended to be passed to
> task executors and schedulers, which allows creation, completion, and status updates of runs.

lyondhill: Thanks that's very clear

task/backend/task.go
lyondhill: That's a good question. Maybe finish run would always return a run for logging. I also thing a failed run finish and should be retried by the executer

task/backend/task.go
```diff
@@ -0,0 +1,31 @@
...
+	"time"
+
+	"github.com/influxdata/influxdb"
+)
+
+// TaskControlService is a low-level controller interface, intended to be passed to
+// task executors and schedulers, which allows creation, completion, and status updates of runs.
+type TaskControlService interface {
+	// CreateNextRun attempts to create a new run.
+	// The new run's now is assigned the earliest possible time according to task's cron,
```
mark-rushakoff: Reading through this again, I see that we are inconsistently using the word `now`.

`influxdb.Run` has a `ScheduledFor` field. `task/backend.QueuedRun` calls it `Now`.

I think in general `ScheduledFor` is more descriptive than `Now`.

But the `now` argument to `CreateNextRun` really is "now", the current time that the scheduler thinks it is. That `now` can stay, or if you have a better name for it, I'm open to that too.

task/backend/task.go
```diff
@@ -0,0 +1,31 @@
...
+	// that is later than any in-progress run and LatestCompleted run.
+	// If the run's now would be later than the passed-in now, CreateNextRun returns a RunNotYetDueError.
+	CreateNextRun(ctx context.Context, taskID influxdb.ID, now int64) (RunCreation, error)
+
+	// FinishRun removes runID from the list of running tasks and if its `now` is later then last completed update it.
+	FinishRun(ctx context.Context, taskID, runID influxdb.ID) (*influxdb.Run, error)
+
+	// NextDueRun returns the Unix timestamp of when the next call to CreateNextRun will be ready.
+	// The returned timestamp reflects the task's delay, so it does not necessarily exactly match the schedule time.
```
mark-rushakoff: `s/delay/offset/`?

task/backend/task.go
```diff
@@ -0,0 +1,31 @@
...
+	NextDueRun() (int64, error)
+
+	// UpdateRunState sets the run state and the respective time.
```
mark-rushakoff: `s/and/at/`?

task/backend/task.go
```diff
@@ -0,0 +1,31 @@
...
+// TaskControlService is a low-level controller interface, intended to be passed to
+// task executors and schedulers, which allows creation, completion, and status updates of runs.
+type TaskControlService interface {
+	// CreateNextRun attempts to create a new run.
+	// The new run's now is assigned the earliest possible time according to task's cron,
+	// that is later than any in-progress run and LatestCompleted run.
+	// If the run's now would be later than the passed-in now, CreateNextRun returns a RunNotYetDueError.
+	CreateNextRun(ctx context.Context, taskID influxdb.ID, now int64) (RunCreation, error)
+
+	// FinishRun removes runID from the list of running tasks and if its `now` is later then last completed update it.
```
mark-rushakoff: `s/now/scheduledFor/`?

Issue #12674, opened 2019-03-18
InfluxDB 2.0: The "/sources/{sourceID}/buckets" response type is defined as "Buckets" but returns a list of "Bucket" [area/api]


bednar:
The `/sources/{sourceID}/buckets` response type is defined as `Buckets`:

```yaml
/sources/{sourceID}/buckets:
    get:
      tags:
        - Sources
        - Buckets
      summary: Get a sources buckets (will return dbrps in the form of buckets if it is a v1 source)
      parameters:
          - $ref: '#/components/parameters/TraceSpan'
          - in: path
            name: sourceID
            schema:
              type: string
            required: true
            description: ID of the source
          - in: query
            name: org
            description: specifies the organization of the resource
            required: true
            schema:
              type: string
      responses:
        '200':
          description: a source
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Buckets"
        '404':
          description: source not found
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Error"
        default:
          description: unexpected error
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Error"
```

but HTTP response returns an array of `Bucket`:

```bash
-> GET 
-> http://127.0.0.1:9999/api/v2/sources/039042392f62d000/buckets

<- 200 
<- OK
<- [{
	"name": "telegraf",
	"rp": "autogen",
	"retentionPeriod": 0
}, {
	"name": "_internal",
	"rp": "monitor",
	"retentionPeriod": 604800000000000
}]
```

kelwang:
cc @bthesorceror

PR #12677, opened 2019-03-18
fix(http): change source buckets from slice to obj

kelwang:
Closes https://github.com/influxdata/influxdb/issues/12674

fixed the response, manually tested as well 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/api_handler.go
+++ b/http/api_handler.go
@@ -119,8 +119,7 @@ func NewAPIHandler(b *APIBackend) *APIHandler {
 
 	sourceBackend := NewSourceBackend(b)
 	sourceBackend.SourceService = authorizer.NewSourceService(b.SourceService)
-	sourceBackend.NewBucketService = b.NewBucketService
-	sourceBackend.NewQueryService = b.NewQueryService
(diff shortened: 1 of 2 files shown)


http/api_handler.go
```diff
@@ -119,8 +119,8 @@ func NewAPIHandler(b *APIBackend) *APIHandler {
 
 	sourceBackend := NewSourceBackend(b)
 	sourceBackend.SourceService = authorizer.NewSourceService(b.SourceService)
-	sourceBackend.NewBucketService = b.NewBucketService
-	sourceBackend.NewQueryService = b.NewQueryService
+	sourceBackend.LabelService = authorizer.NewLabelService(b.LabelService)
```
jademcgough: I don't think it's necessary to wrap labelService in authorizer here. If the user has access to a bucket, we can assume that they can see the bucket's labels. You can move this line to have it be set in `NewSourceBackend` and remove the authorizer.

kelwang: thanks @jademcgough, fixed

PR #12678, opened 2019-03-18
 Enable selecting variable values in time machine

### LINKED ISSUE #12122: Display a variable's value on hover in the script editor menu
As a user 
When I hover over a variable in the side menu in the script editor,
Then I would like to see a popup appear to the left of that variable showing me the value held by that variable
If there are multiple values that can be selected for that variable,
Then I would like to see a dropdown where I can select which value I want to use wherever that variable appears in the script I am currently editing

chnn:
Closes #12122 

pr


ui/src/timeMachine/components/fluxFunctionsToolbar/ToolbarFunction.tsx
```diff
@@ -37,7 +39,11 @@ class ToolbarFunction extends PureComponent<Props, State> {
         onMouseLeave={this.handleStopHover}
         data-testid={testID}
       >
-        {this.tooltip}
+        {isActive && (
```
chnn: The variables list in the script editor needed the same style of tooltips as the functions list, so I extracted them to a component called `BoxTooltip`.


Screen Shot 2019-03-18 at 8 58 13 AM
Screen Shot 2019-03-18 at 8 58 34 AM

ui/src/clockface/components/dropdowns/Dropdown.tsx
```diff
@@ -174,55 +173,52 @@ class Dropdown extends Component<Props, State> {
       children,
       testID,
     } = this.props
+
     const {expanded} = this.state
 
-    if (expanded) {
```
chnn: Hey @alexpaxton @mavarius, could I get your opinion on this change? I removed the validation / dev assertion that a `Dropdown` must have at least 1 child.

I needed to display a dropdown that loads its items asynchronously. So it may no have any items (either hasn't been loaded, or has been loaded but no results were returned).

Screen Shot 2019-03-18 at 8 58 59 AM

Screen Shot 2019-03-18 at 8 59 08 AM

ui/src/timeMachine/components/variableToolbar/VariableItem.tsx
```diff
@@ -1,25 +1,42 @@
...
+  let triggerRect: DOMRect = null
+
+  if (trigger.current) {
+    triggerRect = trigger.current.getBoundingClientRect() as DOMRect
   }
+
+  return (
+    <div
+      className="variables-toolbar--item"
+      onMouseEnter={() => setTooltipVisible(true)}
```
121watts: nit. these anonymous functions will be recreated every time the component renders.

chnn: I was thinking that's OK in this case, since they aren't being passed to a child component that uses them in deciding whether to render or not.

https://reactjs.org/docs/hooks-faq.html#are-hooks-slow-because-of-creating-functions-in-render

ui/src/shared/components/GetVariables.tsx
```diff
@@ -0,0 +1,57 @@
...
+type Props = StateProps & DispatchProps
+
+const GetVariables: FunctionComponent<Props> = ({
+  variablesStatus,
+  onGetVariables,
+  children,
+}) => {
+  useEffect(() => {
+    if (variablesStatus === RemoteDataState.NotStarted) {
+      onGetVariables()
```
121watts: reluctant to use `useEffect` for data fetching until the suspense api has been released. especially since a component that fetches arbitrary resources and calls the thunks that sets them on redux state already exists.  See `GetResources`

ui/src/timeMachine/components/variableToolbar/VariableItem.tsx
121watts: oh, nice!

ui/src/shared/components/GetVariables.tsx
chnn: It does seem likely that any data fetching in hooks we start to do now will be at odds with suspense patterns. I think we're in the same spot with how we fetch data in class components, but I can see value in sticking to just one style, even if it's also incompatible.  So for this PR, I'll replace this hook usage with `GetResources`.

The compatibility with redux with respect to suspense is also pretty unclear—I was reading [this issue](https://github.com/reduxjs/react-redux/issues/1177) the other day, which lays out some of the challenges that the react-redux lib is facing. It's an interesting read.

markerikson: Yeah, tbh, the aspect of "Redux and compat with Suspense" is _very_ unclear atm.  

Dan Abramov had some [thoughts on hypothetical gradiations of "compatibility"](https://github.com/reduxjs/react-redux/issues/1164#issuecomment-457196712).  Not sure where we're going to end up on that list, and I don't think we can even really figure that out until Suspense itself is truly ready for release.

I'd suggest writing code in whatever way works best for you now, and deal with Suspense once it's actually out.  Current patterns won't suddenly stop working - it's just that there may be a new alternative that _could_ simplify things.

chnn: Gotcha. I think we're also trying to grapple with the multitude of different, yet functionally equivalent patterns for fetching data with Redux as it stands today. Trying to orient ourselves towards future compatibility with Suspense does seem like a bit much to take on at once.

I appreciate you chiming in, and am looking forward to where Redux ends up!

Issue #12671, opened 2019-03-18
InfluxDB 2.0: The Telegraf Plugin configuration is required even if it's not necessary [area/api] 


bednar:
When I trying create a new `Telegraf` configuration with a `Kernel Input Plugin` then the configuration of plugin is required even if it's not necessary:

```bash
-> POST 
-> http://127.0.0.1:9999/api/v2/telegrafs
-> {
	"name": "TelegrafConfig1552908917061-IT",
	"description": "test-config",
	"agent": {
		"collectionInterval": 1000
	},
	"plugins": [{
		"comment": "Output to Influx 2.0",
		"name": "influxdb_v2",
		"type": "output",
		"config": {
			"urls": ["http://127.0.0.1:9999"],
			"token": "$INFLUX_TOKEN",
			"organization": "my-org",
			"bucket": "my-bucket"
		}
	}, {
		"name": "kernel",
		"type": "input"
	}],
	"organizationID": "03899e997851e000"
}

<- 400 
<- Bad Request
<- {
	"code": "invalid",
	"message": "An internal error has occurred.",
	"op": "unmarshal telegraf config raw plugin",
	"error": "unexpected end of JSON input"
}
```

The [`Kernel Input Plugin`](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/kernel#configuration) doesn't have a configuration.

PR #12682, opened 2019-03-18
feat(telegraf): allow config to be empty

kelwang:
Closes https://github.com/influxdata/influxdb/issues/12671

added one more test case


telegraf.go
```diff
@@ -336,6 +336,9 @@ func decodePluginRaw(tcd *telegrafConfigDecode, tc *TelegrafConfig) (err error)
 		}
 		if ok {
 			config = tpFn()
+			if len(string(pr.Config)) == 0 {
```
goller: comment this... why is this here for example?

PR #12705, opened 2019-03-18
chore: npm audit updates

121watts:
vulnerability count now = 0


chnn:
@121watts  Hey can you check if this breaks HMR for you? We were holding off on upgrading to `1.12.0` due to this, and also since the audit warnings are a false positive.

121watts:
@chnn def breaks hmr.  closing pr.

PR #12708, opened 2019-03-18
test(stdlib): skip failing window test due to new trigger optimization

jlapacik:
Closes #

Skip failing end-to-end test due to https://github.com/influxdata/flux/issues/1067

  - [ ] Rebased/mergeable
  - [ ] Tests pass


nathanielc:
@jlapacik Why does table order effect the test case? I thought diff didn't care about table order.

jlapacik:
@nathanielc as I mentioned in standup conversation, the test case is terminated by a `group()`, so the table ordering affects the row ordering of the single output table. That's why the test is failing.

PR #12734, opened 2019-03-19
Scope labels to organizations

jademcgough:
- adds an organization ID to labels
- labels must specify their org on creation; a user can create that label if they have read permissions for that org (is that weird?)
- user resource mappings are created for a newly created label, to allow auth enforcement over non-token session

This is red right now because it breaks some chronograf tests (chronograf needs to be updated include an `orgID` in the label body when posting a new label).

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -1638,7 +1638,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/Label"
+                $ref: "#/components/schemas/LabelCreateRequest"
       responses:
         '201':
           description: Added label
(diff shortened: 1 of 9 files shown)


http/swagger.yml
```diff
@@ -1638,7 +1638,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/Label"
+                $ref: "#/components/schemas/LabelCreateRequest"
```
OfTheDelmer: oh! good catch

jademcgough:
I don't think the permissions logic is correct for scoping the label authorizer. I'm making a few changes, should have the commit up shortly.

PR #12770, opened 2019-03-20
fix(authorization): ignore the labels

kelwang:
This fix will ignore the old labels without append them.


goller:
An issue need to be created to remove this code later.  Additionally, existing labels need to be removed that have this problem.

PR #12780, opened 2019-03-20
Improve bulk series file writes.

benbjohnson:
Reapplying https://github.com/influxdata/platform/pull/1599 to `influxdb` repo.

### DIFF
--- a/tsdb/series_index.go
+++ b/tsdb/series_index.go
@@ -155,6 +155,14 @@ func (idx *SeriesIndex) Recover(segments []*SeriesSegment) error {
 	return nil
 }
 
+// GrowBy preallocates the in-memory hashmap to a larger size.
+func (idx *SeriesIndex) GrowBy(delta int) {
+	if delta < 0 {
+		return
+	}
+	idx.keyIDMap.Grow(((idx.keyIDMap.Len() + int64(delta)) * 100) / int64(idx.keyIDMap.LoadFactor()))
+}
+
 // Count returns the number of series in the index.
 func (idx *SeriesIndex) Count() uint64 {
 	return idx.OnDiskCount() + idx.InMemCount()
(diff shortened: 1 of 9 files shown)


tsdb/series_partition.go
```diff
@@ -226,6 +224,13 @@ func (p *SeriesPartition) CreateSeriesListIfNotExists(collection *SeriesCollecti
 	newKeyRanges := make([]keyRange, 0, writeRequired)
 	newIDs := make(map[string]SeriesIDTyped, writeRequired)
 
+	// Pre-grow index for large writes.
+	if writeRequired >= 10000 {
```
e-dard: Can you make `10000` the default, but a configurable value. In other setups we may want to tweak this.

tsdb/series_index.go
```diff
@@ -155,6 +155,14 @@ func (idx *SeriesIndex) Recover(segments []*SeriesSegment) error {
 	return nil
 }
 
+// GrowBy preallocates the in-memory hashmap to a larger size.
+func (idx *SeriesIndex) GrowBy(delta int) {
+	if delta < 0 {
+		return
+	}
+	idx.keyIDMap.Grow(((idx.keyIDMap.Len() + int64(delta)) * 100) / int64(idx.keyIDMap.LoadFactor()))
```
e-dard: can `LoadFactor` ever be `0`?

e-dard: Under what conditions would a `delta` value result in less than a doubling of the map? Seems to me at a minimum we always want to at least double the map right? So maybe we should set a min value for `delta`?

benbjohnson: `LoadFactor` of `0` would have serious issues way before it got to the point of needing to grow. We only adjust the load factor via a constant right now anyway.

re: `delta`, the `rhh.Map` requires that the capacity be a power of two so it will automatically do the doubling.

PR #12791, opened 2019-03-20
Use selected time range for metaqueries in DE

### LINKED ISSUE #11535: Historical metadata never shows in query builder
Organizations > org > Buckets > add data > line protocol > Enter Manually

Clicking through to this option and then pasting in several rows of line protocol then clicking "continue" produces a message that says "Data Written Successfully!" but when I check for the measurements associated with that bucket, there are none, and it appears that there is no data in the bucket.

### LINKED ISSUE #12779: Update meta queries in DE to use selected time range

chnn:
Closes #12779 
Closes #11535

reloading

Screen Shot 2019-03-20 at 1 38 54 PM

### DIFF
--- a/ui/src/timeMachine/components/TagSelector.scss
+++ b/ui/src/timeMachine/components/TagSelector.scss
@@ -52,12 +52,20 @@ button.tag-selector--remove,
   pointer-events: none;
   flex: 1 1 0;
   display: flex;
+  flex-direction: column;
   justify-content: center;
   align-items: center;
+  padding: $ix-marg-c $ix-marg-c $ix-marg-d $ix-marg-c;
   color: $g8-storm;
   font-weight: 500;
   font-size: 16px;
-  text-transform: uppercase;
   text-align: center;
-  padding: 12px 12px 36px 12px;
+  text-transform: uppercase;
+
+  small {
+    display: block;
+    font-size: 14px;
+    text-transform: none;
+    margin-top: $ix-marg-b;
+  }
 }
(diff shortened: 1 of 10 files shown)


chnn:
Connect https://github.com/influxdata/flux/issues/1071

ui/src/timeMachine/components/TagSelector.scss
```diff
@@ -52,12 +52,20 @@ button.tag-selector--remove,
   pointer-events: none;
   flex: 1 1 0;
   display: flex;
+  flex-direction: column;
   justify-content: center;
   align-items: center;
+  padding: 12px 12px 36px 12px;
```
alexpaxton: Can you use margin variables from `_variables.scss`?

Could probably do something like:
```
padding: $ix-marg-c;
padding-bottom: $ix-marg-d;
```

ui/src/timeMachine/components/TagSelector.scss
```diff
@@ -52,12 +52,20 @@ button.tag-selector--remove,
...
-  text-transform: uppercase;
   text-align: center;
-  padding: 12px 12px 36px 12px;
+  text-transform: uppercase;
+
+  small {
+    display: block;
+    font-size: 14px;
+    text-transform: none;
+    margin-top: 10px;
```
alexpaxton: Same comment about variables

ui/src/timeMachine/components/TagSelector.scss
chnn: Yep can do

PR #12793, opened 2019-03-20
fix(tasks): clear existing task schedule when changing types

### LINKED ISSUE #12690: Impossible to change a task to cron after using every
when editing a task, if you try to switch it to cron, you get an error about evey and cron being specified. If someone switches between cron and every, we should remove the non selected one so it can be saved properly.

image

AlirieGray:
Closes #12690 

This PR fixes an issue where switching between the "cron" and "interval" schedule types would cause an error when creating or editing a task. Now the reducer for setting a task option will clear the previous task schedule type when a new one is added.

The v2 folders for task actions and reducers are also removed in this PR.

  - [x] Rebased/mergeable
  - [ ] Tests pass


ui/src/tasks/reducers/tasks.test.ts
```diff
@@ -0,0 +1,50 @@
+import tasksReducer, {
+  defaultState,
+  defaultTaskOptions,
+} from 'src/tasks/reducers'
+import {setTaskOption} from 'src/tasks/actions'
+import {TaskSchedule} from 'src/utils/taskOptionsToFluxScript'
+
+describe('tasksReducer', () => {
+  it('clears the cron property from the task options when interval is selected', () => {
```
121watts: these tests could have a another describe for the action these reducers are responding to.  i.e."

```
describe('tasksReducer', () => {
  describe('setTaskOption', () => {
    ... setTaskOption tests here
   })
})
```

Issue #12772, opened 2019-03-20
InfluxDB 2.0: The "me/password" doesn't have specified a response content [area/api]


bednar:
The `me/password` doesn't have specified a response content:

```yaml
/me/password:
    put:
      tags:
        - Users
      summary: Update password
      security:
        - basicAuth: []
      parameters:
        - $ref: '#/components/parameters/TraceSpan'
      requestBody:
        description: new password
        required: true
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/PasswordResetBody"
      responses:
        '204':
          description: password succesfully updated
        default:
          description: unsuccessful authentication
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Error"
```

but the http response contains an `User`:

```bash
-> PUT 
-> http://127.0.0.1:9999/api/v2/me/password
-> {"password":"my-password"}

<- 200 
<- OK
<- {
	"links": {
		"logs": "/api/v2/users/0391ed592f162000/logs",
		"self": "/api/v2/users/0391ed592f162000"
	},
	"id": "0391ed592f162000",
	"name": "my-user"
}
```

PR #12801, opened 2019-03-20
fix(http): swagger change user pass reset resp

kelwang:
Closes https://github.com/influxdata/influxdb/issues/12772

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -4356,7 +4356,7 @@ paths:
               $ref: "#/components/schemas/PasswordResetBody"
       responses:
         '204':
-          description: password succesfully updated
+          description: password successfully updated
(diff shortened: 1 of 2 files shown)


http/swagger.yml
```diff
@@ -4334,8 +4334,12 @@ paths:
             schema:
               $ref: "#/components/schemas/PasswordResetBody"
       responses:
-        '204':
+        '200':
           description: password succesfully updated
```
jademcgough: succesfully -> successfully

kelwang:
> I'm not sure why the user itself should be returned on updating the user's password. IMO the swagger is correct, we should change the response to be an empty 204.

Good catch, thanks

PR #12806, opened 2019-03-20
fix(ui): fetch queries when opening veo

### LINKED ISSUE #12797: after opening the veo from clicking the edit cell, the graph should be here
Currently, it just says create a query, even though there is a query selected in the builder.

open_veo

ischolten:
Closes #12797

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/components/VEO.tsx
+++ b/ui/src/dashboards/components/VEO.tsx
@@ -32,6 +32,7 @@ import {VEO_TIME_MACHINE_ID} from 'src/timeMachine/constants'
 
 // Nofication Messages
 import {cellAddFailed} from 'src/shared/copy/notifications'
+import {executeQueries} from 'src/timeMachine/actions/queries'
 
 interface StateProps {
   draftView: QueryView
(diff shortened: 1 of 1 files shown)


ui/src/dashboards/components/VEO.tsx
```diff
@@ -58,6 +60,15 @@ interface OwnProps extends WithRouterProps {
 type Props = OwnProps & StateProps & DispatchProps
 
 class VEO extends PureComponent<Props, {}> {
+  public componentDidMount() {
+    const {existingView} = this.props
+    const view = existingView || createView<XYView>(ViewType.XY)
+
+    this.props.setActiveTimeMachine(VEO_TIME_MACHINE_ID, {view})
```
AlirieGray: is it possible to destructure these on line 64?

ischolten: because the name of the action creator is the same as the name that the prop function was named, destructuring would create a shadowed name.

PR #12807, opened 2019-03-21
Cache cleanups

zeebo:
This PR does some cleanups on the cache. Specifically

- `init` and `Free` are unnecessary and can lead to data loss in some race conditions.
- the cache `*ring` type only works with 16 partitions, so update the code and remove comments claiming otherwise.
- fix up the benchmarks that attempt to use rings with different partitions.
- refactor some types into their own files to keep single file length down.
- create an export a type to allow consumers to detect when a write fails due to the cache being full.


tsdb/tsm1/ring.go
```diff
@@ -49,24 +48,12 @@ type ring struct {
...
-	}
-
-	r := ring{
-		partitions: make([]*partition, n), // maximum number of partitions.
-	}
-
-	// The trick here is to map N partitions to all points on the continuum,
-	// such that the first eight bits of a given hash will map directly to one
-	// of the N partitions.
+func newRing() *ring {
```
e-dard: Can ditch the acceptable values for `n` comment in the braces.

e-dard:
🍏

PR #12813, opened 2019-03-21
Drop all unsupported Prometheus values written to the remote write endpoint

### LINKED ISSUE #10490: prometheus api and +/-Inf values
__System info:__ [Include InfluxDB version, operating system name, and other relevant details

InfluxDB v1.7.0 (git: 1.7 dac4c6f571662c63dc0d73346787b8c7f113222a)

__Steps to reproduce:__

Send +/-Inf values with prometheus' remote_write.

Ignore +/-Inf values

The whole batch fails.

Here is an example setup with docker-compose:

https://gist.github.com/datamuc/d6085cc4c4190619796f0ea5a8661dce

```
influxdb_1       | [httpd] 192.168.208.3 - admin [16/Nov/2018:16:34:55 +0000] "POST /api/v1/prom/write?db=prom HTTP/1.1" 400 59 "-" "Go-http-client/1.1" 8ce96825-e9bd-11e8-8001-0242c0a8d002 88584
prometheus_1     | level=warn ts=2018-11-16T16:34:55.788646973Z caller=queue_manager.go:531 component=remote queue="0:http://influxdb:8086/api/v1/prom/write?db=prom" msg="Error sending samples to remote storage" count=375 err="server returned HTTP status 400 Bad Request: {\"error\":\"+/-Inf is an unsupported value for field value\"}"
```

All metrics sent in this POST request are lost because of a single +Inf value.

e-dard:
Fixes #10490.

This PR extends the Prometheus remote write endpoint to drop
unsupported Prometheus values, rather than reject the entire batch.

InfluxDB does not support `NaN`, `-Inf` or `+Inf`, but Prometheus does. The
remote write endpoint will now drop these and write valid values in the
provided batch.

If the user enabled write trace logging (`[http] write-tracing = true`)
then summaries of any dropped values within a batch will be logged.

If a batch of values contains any values that are subsequently dropped,
the returned status code will be `204`.

/cc @stevebang we should edit https://docs.influxdata.com/influxdb/v1.7/supported_protocols/prometheus/#how-prometheus-metrics-are-parsed-in-influxdb so that it explains what happens if `NaN`,  `Inf` or `-Inf` are written to InfluxDB via the Prometheus remote write endpoint.

### DIFF
--- a/prometheus/converters.go
+++ b/prometheus/converters.go
@@ -30,7 +30,18 @@ const (
 	measurementTagKey = "_measurement"
 )
 
-var ErrNaNDropped = errors.New("dropped NaN from Prometheus since they are not supported")
+// A DroppedValuesError is returned when the prometheus write request contains
+// unsupported float64 values.
+type DroppedValuesError struct {
+	nan  uint64
+	ninf uint64
+	inf  uint64
+}
+
+// Error returns a descriptive error of the values dropped.
+func (e DroppedValuesError) Error() string {
+	return fmt.Sprintf("dropped unsupported Prometheus values: [NaN = %d, +Inf = %d, -Inf = %d]", e.nan, e.inf, e.ninf)
+}
 
 // WriteRequestToPoints converts a Prometheus remote write request of time series and their
 // samples into Points that can be written into Influx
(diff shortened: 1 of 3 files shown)


stevebang:
For 1.7.6 release, appended the following to the Prometheus (https://docs.influxdata.com/influxdb/v1.7/supported_protocols/prometheus/#how-prometheus-metrics-are-parsed-in-influxdb):

* [v1.7.6 and later] Prometheus remote write endpoint drops unsupported Prometheus values (`NaN`,`-Inf`, and `+Inf`) rather than reject the entire batch.
  * If [write trace logging is enabled (`[http] write-tracing = true`)](/influxdb/v1.7/administration/config/#write-tracing-false), then summaries of dropped values are logged.
  * If a batch of values contains values that are subsequently dropped, HTTP status code `204` is returned.

PR #12821, opened 2019-03-21
feat(dashboard): add ability to change order of variable dropdowns

### LINKED ISSUE #12376: Store the user's selection for a Variable's value locally
As a user
When I select a value for a value in a dashboard 
Then I would like that variable to have the value I selected when I come back to that dashboard

(use local storage)

### LINKED ISSUE #12512: Change the order of variables on a dashboard
As a user
With editor permissions or higher
When I am viewing the list of variables at the top of the dashboard page
Then I would like to be able to change the order of variables as they appear in that list

This order should be persisted, so that I see the variables in the same order when I close and return to the dashboard page

AlirieGray:
Closes #12512
Closes #12376

This PR allows variables to be moved left and right within the `VariablesControlBar` in a dashboard. The order of the variables on a given dashboard is stored in the `values` property of the Variables state in local storage.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/components/variablesControlBar/VariablesControlBar.tsx
+++ b/ui/src/dashboards/components/variablesControlBar/VariablesControlBar.tsx
@@ -1,10 +1,11 @@
 // Libraries
 import React, {PureComponent} from 'react'
 import {connect} from 'react-redux'
-import _ from 'lodash'
+import {isEmpty} from 'lodash'
+import {DragDropContext} from 'react-dnd'
+import HTML5Backend from 'react-dnd-html5-backend'
 
 // Components
-import VariableDropdown from 'src/dashboards/components/variablesControlBar/VariableDropdown'
 import {EmptyState, ComponentSize} from 'src/clockface'
 import {TechnoSpinner} from '@influxdata/clockface'
 
@@ -14,6 +15,12 @@ import {
   getDashboardValuesStatus,
 } from 'src/variables/selectors'
 
+// Styles
+import 'src/dashboards/components/variablesControlBar/VariablesControlBar.scss'
+
+// Actions
+import {moveVariable} from 'src/variables/actions'
+
 // Types
 import {AppState} from 'src/types/v2'
 import {Variable} from '@influxdata/influx'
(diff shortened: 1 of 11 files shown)


ui/src/dashboards/components/variablesControlBar/VariablesControlBar.tsx
```diff
@@ -1,10 +1,12 @@
...
 import React, {PureComponent} from 'react'
 import {connect} from 'react-redux'
-import _ from 'lodash'
+import {isEmpty} from 'lodash'
+import {DragDropContext} from 'react-dnd'
+import HTML5Backend from 'react-dnd-html5-backend'
 
 // Components
-import VariableDropdown from 'src/dashboards/components/variablesControlBar/VariableDropdown'
+// import VariableDropdown from 'src/dashboards/components/variablesControlBar/VariableDropdown'
```
chnn:
```suggestion
```

ui/src/variables/reducers/index.ts
```diff
@@ -81,13 +82,29 @@ export const variablesReducer = (
 
       case 'SET_VARIABLE_VALUES': {
         const {contextID, status, values} = action.payload
+        const originalOrder = draftState.values[contextID].order
```
chnn: This line was preventing me from loading a dashboard that had never been loaded before. In this case there are no values in localStorage, so `draftState.values[contextID]` doesn't exist, so accessing its `order` property is a run time error.

I think we can also simplify the logic that maintains the existing order:

```
        const prevOrder = get(draftState, `values.${contextID}.order`) || []

        if (values) {
          // Maintain the existing sort order, placing new variables at the
          // beginning
          const order = Object.keys(values).sort(
            (a, b) => prevOrder.indexOf(a) - prevOrder.indexOf(b)
          )

          draftState.values[contextID] = {
            status,
            values,
            order,
          }
        } else if (draftState.values[contextID]) {
```

AlirieGray: clever! I like it

PR #12826, opened 2019-03-21
Improve display of query errors

### LINKED ISSUE #12686: Add the ability to copy error message from a cell
as a dashboard user
with editor
when i see a flux error in a dashboard cell
then i would like the ability to copy that error

I can't copy the error text in this cell:
image

chnn:
Closes #12686

error

Screen Shot 2019-03-21 at 1 14 21 PM

Also tweaks the query execution code to check for [error messages that appear as a Flux table](https://github.com/influxdata/flux/blob/master/docs/SPEC.md#errors). Previously we ignored these errors and attempted to plot the data anyway.

### DIFF
--- a/ui/src/shared/components/EmptyGraphError.tsx
+++ b/ui/src/shared/components/EmptyGraphError.tsx
@@ -1,0 +1,63 @@
+// Libraries
+import React, {useState, FunctionComponent} from 'react'
+import CopyToClipboard from 'react-copy-to-clipboard'
+
+// Components
+import {
+  Button,
+  ComponentSize,
+  ComponentColor,
+  IconFont,
+} from '@influxdata/clockface'
+import FancyScrollbar from 'src/shared/components/fancy_scrollbar/FancyScrollbar'
+
+interface Props {
+  message: string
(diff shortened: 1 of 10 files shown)


ui/src/shared/components/EmptyGraphError.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const onClick = () => {
+    setDidCopy(true)
+    setTimeout(() => setDidCopy(false), 2000)
+  }
+
+  return (
+    <div className="cell--view-empty" data-testid={testID}>
+      <div className="empty-graph-error">
+        <CopyToClipboard text={message}>
+          <Button
```
mavarius: Speak with @ischolten about standardizing copy to clipboard implementation across the application.

chnn: Will do. I first tried using the `CodeSnippet` component used in the Telegraf configuration screen. The always-visible big purple "Copy to Clipboard" button felt out of place on the dashboard, which is why I created yet another way of doing the same thing :|

PR #12834, opened 2019-03-21
fix(http): ignore an empty trailer inside of the proxy query service client

jsternberg:


http/proxy_query_service.go
```diff
@@ -182,8 +182,10 @@ func (s *ProxyQueryService) Query(ctx context.Context, w io.Writer, req *query.P
 
 	data := []byte(resp.Trailer.Get(QueryStatsTrailer))
 	var stats flux.Statistics
-	if err := json.Unmarshal(data, &stats); err != nil {
-		return stats, tracing.LogError(span, err)
+	if len(data) > 0 {
+		if err := json.Unmarshal(data, &stats); err != nil {
```
wolffcm: Can you put a note here and link to the issue: https://github.com/influxdata/idpe/issues/2841 
It's really a bug in Envoy that the trailers are missing.  Queryd always sends them, so it's confusing.

PR #12847, opened 2019-03-22
Snapshot cache based on age 

e-dard:
This PR adds a configuration option to the cache that allows it to be snapshotted based on its age, regardless of the cache size.

This option is useful when running the system without a WAL.


e-dard:
@stuartcarnie @zeebo dang I forgot one thing extra I plan to do.... I'm going to provide metrics around what caused the cache to be snapshotted (hence the different statuses)...

I'll add that commit on Monday.

tsdb/tsm1/config.go
```diff
@@ -79,10 +75,12 @@ type CompactionConfig struct {
 	MaxConcurrent int `toml:"max-concurrent"`
 }
 
+// Default Cache configuration values.
 const (
-	DefaultCacheMaxMemorySize             = 1024 * 1024 * 1024 // 1GB
-	DefaultCacheSnapshotMemorySize        = 25 * 1024 * 1024   // 25MB
-	DefaultCacheSnapshotWriteColdDuration = time.Duration(10 * time.Minute)
+	DefaultCacheMaxMemorySize             = toml.Size(1024 << 20)           // 1GB
```
benbjohnson: Minor nit but why not `1 << 30`?

tsdb/tsm1/engine.go
```diff
@@ -866,39 +874,68 @@ func (e *Engine) compactCache() {
...
+		return CacheStatusAgeExceeded
+	}
+
+	// Cache has not been written to for a long time.
+	if t.Sub(e.Cache.LastWriteTime()) > e.CacheFlushWriteColdDuration {
+		return CacheStatusColdNoWrites
 	}
 
-	return t.Sub(e.Cache.LastWriteTime()) > e.CacheFlushWriteColdDuration
+	return 0
```
benbjohnson: Should this be `CacheStatusOkay`?

tsdb/tsm1/config.go
e-dard: Just seemed clearer to use the same base (MB) for all the constants).

PR #12866, opened 2019-03-25
Fix/task health

wolffcm:
Make the QueryService interfaces implement check.Checker.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/query_handler.go
+++ b/http/query_handler.go
@@ -17,14 +17,15 @@ import (
 	"github.com/influxdata/flux/csv"
 	"github.com/influxdata/flux/iocounter"
 	"github.com/influxdata/flux/parser"
-	"github.com/julienschmidt/httprouter"
-	"github.com/prometheus/client_golang/prometheus"
-	"go.uber.org/zap"
-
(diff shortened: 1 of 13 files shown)


http/query_handler.go
```diff
@@ -495,3 +504,45 @@ func SimpleQuery(addr, flux, org, token string) ([]byte, error) {
...
+	u, err := newURL(url, "/health")
+	if err != nil {
+		return check.Response{
+			Name:    "query health",
+			Status:  check.StatusFail,
+			Message: errors.Wrap(err, "could not form URL").Error(),
+		}
+	}
+
+	insecureSkipVerify := true
```
mark-rushakoff: Hardcoding insecureSkipVerify to true seems a little suspect. Would it be better off as an explicit parameter to `QueryHealthCheck`? Off the top of my head I'm not sure how that gets plumbed through all the layers of query services.

wolffcm: You are right.  All of the callers have `insecureSkipVerify` as a member, so I added it as a parameter to `QueryHealthCheck`.

Issue #12864, opened 2019-03-25
InfluxDB 2.0: The "GET" method of "/documents/templates/{templateID}" should return "Not Found" for a template that doesn't exists [area/api]


bednar:
The `/documents/templates/{templateID}` should return `404 HTTP response` for a template that doesn't exists:

```bash
-> GET 
-> http://127.0.0.1:9999/api/v2/documents/templates/020f755c3c082000

<- 500 
<- Internal Server Error
<- {
	"code": "internal error",
	"error": "key not found"
}
```

but currently the HTTP response is `500: Internal Server Error`.

introduced by #12382, cc @desa, @goller

PR #12868, opened 2019-03-25
fix(http): fix template not found error

kelwang:
Closes https://github.com/influxdata/influxdb/issues/12864
Closes https://github.com/influxdata/influxdb/issues/12871

add an error check for not found errors for documents

### DIFF
--- a/kv/document.go
+++ b/kv/document.go
@@ -581,6 +581,13 @@ func (s *DocumentStore) FindDocuments(ctx context.Context, opts ...influxdb.Docu
 		return nil
 	})
 
+	if IsNotFound(err) {
+		return nil, &influxdb.Error{
+			Code: influxdb.ENotFound,
+			Msg:  influxdb.ErrDocumentNotFound,
+		}
+	}
+
 	if err != nil {
 		return nil, err
 	}
--- a/document.go
+++ b/document.go
@@ -4,6 +4,9 @@ import (
 	"context"
 )
 
+// ErrDocumentNotFound is the error msg for a missing document.
+const ErrDocumentNotFound = "document not found"
+
 // DocumentService is used to create/find instances of document stores.
 type DocumentService interface {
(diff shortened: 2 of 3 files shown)


kv/document.go
```diff
@@ -81,6 +81,12 @@ func (s *Service) FindDocumentStore(ctx context.Context, ns string) (influxdb.Do
 	})
 
 	if err != nil {
+		if IsNotFound(err) {
```
desa: I don't think this solves #12864. This function is just used to lookup a namespace not the actual document.

kelwang: Thanks a lot

kv/document.go
```diff
@@ -582,6 +582,12 @@ func (s *DocumentStore) FindDocuments(ctx context.Context, opts ...influxdb.Docu
...
```
desa: What do you think about un-nesting the conditional so it's something like

```go
if IsNotFoundErr(err) {
   ...
}

if err != nil {
...
}
```

kelwang:
> Could you add a test for this?

i added another ticket for the testing, https://github.com/influxdata/influxdb/issues/12871

desa:
there are tests, they just aren't unit tests, it should be possible to extend them.

https://github.com/influxdata/influxdb/blob/6c6247efb3a45558be2a211c81dff8c3bae9fe75/testing/document.go#L15

kelwang:
> there are tests, they just aren't unit tests, it should be possible to extend them.
> 
> [influxdb/testing/document.go](https://github.com/influxdata/influxdb/blob/6c6247efb3a45558be2a211c81dff8c3bae9fe75/testing/document.go#L15)
> 
> Line 15 in [6c6247e](/influxdata/influxdb/commit/6c6247efb3a45558be2a211c81dff8c3bae9fe75)
> 
>  func NewDocumentIntegrationTest(store kv.Store) func(t *testing.T) {

added

Issue #12860, opened 2019-03-25
InfluxDB 2.0: The Document with a label that doesn't exists should throw clear error message [area/api]


bednar:
When I create a `Document` with a `label` that doesn't exists:

```bash
-> POST 
-> http://127.0.0.1:9999/api/v2/documents/templates
-> {
	"meta": {
		"name": "document-name1553508154030-IT",
		"version": "1"
	},
	"content": "templates content",
	"orgID": "0399033576e2b000",
	"labels": ["not_exists_label_1553508154030-IT"]
}

<- 500 
<- Internal Server Error
<- {
	"code": "internal error",
	"message": "found multiple labels matching the name provided"
}
```

then the HTTP response should be `404` with a clear message like `label not found`.

introduced by #12382, cc @desa, @goller

PR #12870, opened 2019-03-25
fix(http): fix doc label not found error

kelwang:
Closes https://github.com/influxdata/influxdb/issues/12860


kelwang:
> LGTM, only thing I think that would be good is to add a test for this behavior.

Done 👍

Issue #12814, opened 2019-03-21
InfluxDB 2.0: The "/sources/{sourceID}/buckets" returns a default bucket for "Source" that doesn't exists  [area/api]


bednar:
The `Source` with id = **ffffffffffffffff** doesn't exists but the response contains an one bucket: 

```bash
-> GET 
-> http://127.0.0.1:9999/api/v2/sources/ffffffffffffffff/buckets

<- 200 
<- OK
<- {
	"links": {
		"self": "/api/v2/buckets?descending=false\u0026limit=20\u0026offset=0"
	},
	"buckets": [{
		"id": "039416d357144001",
		"organizationID": "039416d357144000",
		"organization": "my-org",
		"name": "my-bucket",
		"retentionRules": [],
		"links": {
			"labels": "/api/v2/buckets/039416d357144001/labels",
			"logs": "/api/v2/buckets/039416d357144001/logs",
			"members": "/api/v2/buckets/039416d357144001/members",
			"org": "/api/v2/orgs/039416d357144000",
			"owners": "/api/v2/buckets/039416d357144001/owners",
			"self": "/api/v2/buckets/039416d357144001",
			"write": "/api/v2/write?org=039416d357144000\u0026bucket=039416d357144001"
		},
		"labels": []
	}]
}
```

The result should be empty or throw `HTTP 404 Not found` exception.

Probably caused by #12677

cc @kelwang

PR #12875, opened 2019-03-25
fix(http): fix invalid source buckets

kelwang:
Closes https://github.com/influxdata/influxdb/issues/12814

### DIFF
--- a/http/source_service.go
+++ b/http/source_service.go
@@ -226,7 +226,13 @@ func (h *SourceHandler) handleGetSourcesBuckets(w http.ResponseWriter, r *http.R
 		return
 	}
 
-	bs, _, err := h.BucketService.FindBuckets(ctx, req.filter)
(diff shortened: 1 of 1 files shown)


http/source_service.go
```diff
@@ -226,7 +226,13 @@ func (h *SourceHandler) handleGetSourcesBuckets(w http.ResponseWriter, r *http.R
 		return
 	}
 
-	bs, _, err := h.BucketService.FindBuckets(ctx, req.filter)
+	_, err = h.SourceService.FindSourceByID(ctx, req.getSourceRequest.SourceID)
```
desa: I've been out of the loop for a while. Is this route expected to work? I thought we were rethinking how sources would work.

kelwang: yeah, i am not aware of any new plans.

PR #12922, opened 2019-03-26
Variables/style polish

### LINKED ISSUE #12844: Polish Variables Control Bar styles
- [x] Add padding to top of VariablesControlBar in presentation mode
- [x] Remove space between variable dropdown and label

AlirieGray:
Closes #12844

This PR fixes two small style bugs in the `VariablesControlBar` component. 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/dashboards/components/variablesControlBar/VariablesControlBar.scss
+++ b/ui/src/dashboards/components/variablesControlBar/VariablesControlBar.scss
@@ -21,14 +21,17 @@ $variables-control-bar--gutter: $ix-marg-a;
   }
 
   .variable-dropdown {
-    margin: 0 $variables-control-bar--gutter / 2;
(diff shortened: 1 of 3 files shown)


ui/src/dashboards/components/variablesControlBar/VariablesControlBar.scss
```diff
@@ -40,3 +39,7 @@ $variables-control-bar--gutter: $ix-marg-a;
 .variables-control-bar > .techno-spinner {
   margin-left: 10px;
 }
+
+.variables-control-bar.presentation-mode {
+  padding-top: 8px !important;
```
alexpaxton: Usually I try to avoid using `!important` as little as possible. I feel like there's a way to achieve the same look without having to use it

PR #12933, opened 2019-03-27
Fix stale render flash upon opening VEO

### LINKED ISSUE #12789: Cell editor showing previously configured cell viz
Cell editor shows visualization from the previously clicked cell.

Kapture 2019-03-20 at 16 24 49

chnn:
Closes #12789

A VEO contains a time machine. When the VEO opens, we reset the time machine state according to the current view being edited (for example, the queries in the time machine should start as the queries for the view being edited).

We previously added additional state to the time machine that depends on the view (https://github.com/influxdata/influxdb/pull/12619) , but neglected to reset this state whenever a new time machine was activated.

In particular, the `queryResults` field of the time machine state was not reset, which led to stale data being momentarily rendered in the time machine visualization when a time machine was activated (#12789).

This commit ensures the `queryResults` field is reset whenever a new time machine is activated. It also ensures other data that only correspond to the lifetime of a particular time machine are reset as well.

### DIFF
--- a/ui/src/dashboards/components/VEO.tsx
+++ b/ui/src/dashboards/components/VEO.tsx
@@ -48,13 +48,14 @@ class VEO extends PureComponent<Props, State> {
   }
 
   public componentDidUpdate() {
-    if (
-      !this.state.hasActivatedTimeMachine &&
-      this.props.viewsStatus === RemoteDataState.Done
-    ) {
-      this.props.onSetActiveTimeMachine(VEO_TIME_MACHINE_ID, {
-        view: this.props.view,
-      })
+    const {view, onSetActiveTimeMachine} = this.props
+    const {hasActivatedTimeMachine} = this.state
+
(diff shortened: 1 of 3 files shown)


ui/src/dashboards/components/VEO.tsx
```diff
@@ -48,13 +48,14 @@ class VEO extends PureComponent<Props, State> {
...
-        view: this.props.view,
-      })
+    const {view, onSetActiveTimeMachine} = this.props
+    const {hasActivatedTimeMachine} = this.state
+
+    const timeMachineShouldActivate =
+      !hasActivatedTimeMachine && this.loading === RemoteDataState.Done
+
+    if (timeMachineShouldActivate) {
+      onSetActiveTimeMachine(VEO_TIME_MACHINE_ID, {view})
```
OfTheDelmer: tidy

PR #12949, opened 2019-03-27
update task schedler to use the new task control service

### LINKED ISSUE #12725: Refactor task scheduler to use the TaskControlService
backend.Scheduler relies on the backend.Store which will be deprecated soon. We need to update the scheduler to rely on the new TaskControlService.

lyondhill:
Closes #12725


task/mock/task_control_service.go
```diff
@@ -0,0 +1,326 @@
...
+	defer d.mu.Unlock()
+
+	run := d.runs[taskID.String()][runID.String()]
+	run.Status = state.String()
+	switch state {
+	case backend.RunStarted:
+		run.StartedAt = when.Format(time.RFC3339Nano)
+	case backend.RunSuccess, backend.RunFail, backend.RunCanceled:
+		run.FinishedAt = when.Format(time.RFC3339Nano)
+	}
```
mark-rushakoff: For safety, maybe add a `default` case that panics?

task/mock/scheduler.go
```diff
@@ -22,28 +19,19 @@ type Scheduler struct {
 
 	lastTick int64
 
-	claims map[string]*Task
-	meta   map[string]backend.StoreTaskMeta
+	claims map[string]*platform.Task
```
mark-rushakoff: Since we're in this code anyway, can you change this to `map[influxdb.ID]*platform.Task`?

task/backend/task.go
```diff
@@ -29,3 +33,192 @@ type TaskControlService interface {
...
+}
+
+func (tcs *taskControlAdaptor) ManualRuns(ctx context.Context, taskID influxdb.ID) ([]*influxdb.Run, error) {
+	t, m, err := tcs.s.FindTaskByIDWithMeta(ctx, taskID)
+	if err != nil {
+		return nil, err
+	}
+
+	var rtn = make([]*influxdb.Run, len(m.CurrentlyRunning))
+	for i, cr := range m.CurrentlyRunning {
```
mark-rushakoff: This should be `m.ManualRuns`, right? This should have been caught by a test -- please add an appropriate one.

task/backend/task.go
```diff
@@ -29,3 +33,192 @@ type TaskControlService interface {
...
+		rtn[i] = &influxdb.Run{
+			ID:           influxdb.ID(cr.RunID),
+			TaskID:       t.ID,
+			ScheduledFor: time.Unix(cr.Now, 0).Format(time.RFC3339),
```
mark-rushakoff: This should be `time.Unix(...).UTC().Format(...)` to strip the server's time zone.

We probably are missing a test to ensure that the parsed time from a run has `Zone()` with offset zero.

task/backend/task.go
```diff
@@ -29,3 +33,192 @@ type TaskControlService interface {
...
+		}
+		if cr.RequestedAt != 0 {
+			rtn[i].RequestedAt = time.Unix(cr.RequestedAt, 0).Format(time.RFC3339)
```
mark-rushakoff: `UTC()` here too.

task/backend/task.go
```diff
@@ -29,3 +33,192 @@ type TaskControlService interface {
...
+		}
+	}
+	return rtn, nil
+}
+
+func (tcs *taskControlAdaptor) NextDueRun(ctx context.Context, taskID influxdb.ID) (int64, error) {
+	_, m, err := tcs.s.FindTaskByIDWithMeta(ctx, taskID)
```
mark-rushakoff: Even though this code won't be around too long, I guess it could be `FindTaskMetaByID`.

task/backend/task.go
```diff
@@ -29,3 +33,192 @@ type TaskControlService interface {
...
+	// in the old system the log store may not have the run until after the first
+	// state update, so we will need to pull the currently running.
+	if schedFor.IsZero() {
+		for _, cr := range m.CurrentlyRunning {
+			if influxdb.ID(cr.RunID) == runID {
+				schedFor = time.Unix(cr.Now, 0)
+				reqAt = time.Unix(cr.RequestedAt, 0)
+			}
+		}
+
```
mark-rushakoff: Remove blank line.

task/backend/task.go
```diff
@@ -29,3 +33,192 @@ type TaskControlService interface {
...
```
mark-rushakoff: Remove blank line.

task/backend/task.go
```diff
@@ -29,3 +33,192 @@ type TaskControlService interface {
...
+		pt.Offset = opts.Offset.String()
+	}
+	if m != nil {
+		pt.Status = string(m.Status)
+		pt.LatestCompleted = time.Unix(m.LatestCompleted, 0).Format(time.RFC3339)
+		if m.CreatedAt != 0 {
+			pt.CreatedAt = time.Unix(m.CreatedAt, 0).Format(time.RFC3339)
+		}
+		if m.UpdatedAt != 0 {
+			pt.UpdatedAt = time.Unix(m.UpdatedAt, 0).Format(time.RFC3339)
```
mark-rushakoff: `.UTC()` on the three times here.

task/backend/scheduler.go
```diff
@@ -286,7 +269,7 @@ func (s *TickScheduler) Stop() {
 	s.executor.Wait()
 }
 
-func (s *TickScheduler) ClaimTask(task *StoreTask, meta *StoreTaskMeta) (err error) {
+func (s *TickScheduler) ClaimTask(authCtx context.Context, task *platform.Task) (err error) {
```
mark-rushakoff: `authCtx` is a good annotation here that we expect the task to have an authorizer.

Should ClaimTask fail early if there is no authorizer on it?

task/backend/scheduler.go
```diff
@@ -337,20 +325,33 @@ func (s *TickScheduler) UpdateTask(task *StoreTask, meta *StoreTaskMeta) error {
...
 	ts.nextDueMu.Lock()
 	ts.hasQueue = hasQueue
 	ts.nextDue = next
+	ts.authCtx = authCtx
 	ts.nextDueMu.Unlock()
-
 	// check the concurrency
 	// todo(lh): In the near future we may not be using the scheduler to manage concurrency.
-	maxC := int(meta.MaxConcurrency)
+	opt, err := options.FromScript(task.Flux)
```
mark-rushakoff: Since `options.FromScript` could fail, could that be moved earlier in the method so we can avoid some locking if it happens to produce an error?

task/backend/scheduler.go
```diff
@@ -675,16 +688,21 @@ func (r *runner) fail(qr QueuedRun, runLogger *zap.Logger, stage string, reason
 func (r *runner) executeAndWait(ctx context.Context, qr QueuedRun, runLogger *zap.Logger) {
 	defer r.wg.Done()
 
+	defer func() {
```
mark-rushakoff: Undo this change to defer the call to FinishRun. The specific error messages at the specific points are necessary debugging information. Without spending time to trace through a lot of code, I can't say what happens when FinishRun happens _after_, rather than before, calls to r.fail or r.startFromWorking.

task/backend/scheduler.go
```diff
@@ -722,45 +739,28 @@ func (r *runner) executeAndWait(ctx context.Context, qr QueuedRun, runLogger *za
...
-		RunScheduledFor: qr.Now,
-		RequestedAt:     qr.RequestedAt,
-	}
 	stats := rr.Statistics()
 
 	b, err := json.Marshal(stats)
 	if err == nil {
-		r.logWriter.AddRunLog(r.ctx, rlb, time.Now(), string(b))
+		// authctx can be updated mid process
+		r.ts.nextDueMu.RLock()
```
mark-rushakoff: Is this lock right?

The type definition doesn't place authCtx near nextDueMu, and I didn't see any other accesses using a lock.

task/backend/scheduler_test.go
```diff
@@ -456,13 +440,13 @@ func TestScheduler_Queue(t *testing.T) {
 	// Tick the scheduler so the next natural run will happen once 180 finishes.
 	o.Tick(3062)
 
-	// Cancel 180. Next run should be 3060, the next natural schedule.
+	// Cancel 180. Next run should be 240, manual runs get priority.
 	e.RunningFor(task.ID)[0].Cancel()
-	pollForRun(3060)
+	pollForRun(240)
 
 	// Cancel the 3060 run; 240 should pick up.
```
mark-rushakoff: Good catch on manual runs getting priority.  This and the next comment are now stale.

task/backend/scheduler_test.go
```diff
@@ -471,59 +455,125 @@ func TestScheduler_Queue(t *testing.T) {
...
 
 	t.Logf("Didn't find message %q in logs:", exp)
 	for _, log := range logs {
-		t.Logf("\t%s", log.Message)
+		t.Logf("\t%s", log)
 	}
 	t.FailNow()
 }
 
+// LogListener allows us to act as a middleware and see if specific logs have been written
```
mark-rushakoff: Stale comment.

task/backend/scheduler_test.go
```diff
@@ -471,59 +455,125 @@ func TestScheduler_Queue(t *testing.T) {
...
+type runListener struct {
+	mu sync.Mutex
+
+	backend.TaskControlService
+
+	rs map[string][]*platform.Run
```
mark-rushakoff: Prefer platform.ID for the map key.

task/backend/scheduler.go
lyondhill: There is a race condition where a `UpdateTask` can be called during execution. Im using nextDueMu to ensure the write to the authCtx doesn't cause a failure here. We could rename the nextDueMu to be a little more inclusive.

task/backend/scheduler.go
lyondhill: In the new system we wont be able to update the status of a task or add to its logs after it is "finished" something that we were guilty of doing quite a bit in the scheduler. I think we can still accomplish more clear error messaging but keep the finish as a defer.

task/backend/scheduler.go
lyondhill: Im not actually sure. Auth is required during the transition period because the TaskControlAdaptor makes calls to our storage system. I think validation in the future will probably happen before things get to the scheduler. I think for now we can catch it below the scheduler but It may be arranged differently in the stack in the future.

task/backend/task.go
lyondhill: This should have been picked up by the TestScheduler_Queue test. I will dig in and see whats going on here.

PR #12951, opened 2019-03-27
fix(tasks): make durations visibly show up nicely

docmerlin:


task/backend/meta.go
```diff
@@ -43,20 +45,30 @@ func (stm *StoreTaskMeta) AlignLatestCompleted() {
...
 			// For example, every 1m truncates to exactly on the minute.
 			// But the input request is schedule after, not "on or after".
 			// Add one interval.
-			t += int64(every / time.Second)
+			tafter, err := every.Add(t)
+			if err != nil {
+				return
+			}
+			t = tafter
+			//t.A += //int64(every / time.Second)
```
mark-rushakoff: Remove comment.

task/options/options.go
```diff
@@ -40,22 +45,129 @@ type Options struct {
...
+	}
+	a.Node = *q
+	return nil
+}
+
+func (a Duration) MarshalText() ([]byte, error) {
+	return []byte(a.String()), nil
+}
+
+func (a Duration) MartialJSON() ([]byte, error) {
```
mark-rushakoff: s/Martial/Marshal/

task/options/options.go
```diff
@@ -40,22 +45,129 @@ type Options struct {
...
+func (a *Duration) IsZero() bool {
+	for i := range a.Node.Values {
+		if a.Node.Values[i].Magnitude != 0 {
+			return false
+		}
+	}
+	return true
+}
+
+func (a *Duration) UnmartialJSON(b []byte) error {
```
mark-rushakoff: s/Martial/Marshal/

task/options/options.go
```diff
@@ -40,22 +45,129 @@ type Options struct {
...
+	return []byte(a.String()), nil
+}
+
+func (a Duration) MartialJSON() ([]byte, error) {
+	q := a.String()
+	res := make([]byte, len(q)+2)
+	res[0] = '"'
+	res[len(res)-1] = '"'
+	copy(res[1:len(res)-1], q)
+	return res, nil
```
mark-rushakoff: This would probably be simpler as `return json.Marshal(a.String())`...

Since the method name is misspelled, json.Marshal doesn't look at this method, but it does see `MarshalText`:

> If no MarshalJSON method is present but the value implements encoding.TextMarshaler instead, Marshal calls its MarshalText method and encodes the result as a JSON string.

So maybe you can just remove the MarshalJSON and UnmarshalJSON methods altogether?

task/options/options.go
```diff
@@ -40,22 +45,129 @@ type Options struct {
...
+		}
+	}
+	return true
+}
+
+func (a *Duration) UnmartialJSON(b []byte) error {
+	if len(b) < 2 {
+		return errors.New("duration json must be at least 2 bytes long including the quotes")
+	}
+	x, err := parseSignedDuration(string(b[1 : len(b)-1]))
```
mark-rushakoff: I would probably rely on json.Unmarshal here too, like 

```go
var s string
if err := json.Unmarshal(b, &s); err != nil {
  return err
}

x, err := parseSignedDuration(s)
// ...
```

If we even need it, given the presence of MarshalText.

task/options/options.go
```diff
@@ -80,6 +192,42 @@ const (
...
+		}
+	}
+	return false
+}
+
+func grabTaskOptionAST(p *ast.Package, keys ...string) map[string]ast.Expression {
+	res := make(map[string]ast.Expression, 2) // we preallocate two keys for the map, as that is how many we will use at maximum (offset and every)
+	for i := range p.Files {
+		for j := range p.Files[i].Body {
+			if p.Files[i].Body[j].Type() == "OptionStatement" {
```
mark-rushakoff: This would probably be a little more readable with inverted conditions and continues, for reduced indentation, like

```
for ... {
  if p.Files[i].Body[j].Type() != "OptionStatement" {
    continue
  }
  opt := (p.Files[i].Body[j]).(*ast.OptionStatement)
  if opt.Assignment.Type() != "VariableAssignment" {
    continue
  }
  asmt := opt.Assignment.(*ast.VariableAssignment)
  // ...
}

File: task/options/options_test.go [hunk 7]
```diff
@@ -74,6 +91,7 @@ func TestFromScript(t *testing.T) {
 			continue
 		}
 		if !cmp.Equal(o, c.exp) {
+			fmt.Println(c.exp)
```
mark-rushakoff: Remove print.

task/options/options_test.go
```diff
@@ -191,3 +209,85 @@ func TestEffectiveCronString(t *testing.T) {
...
+}
+
+func TestDurationMath(t *testing.T) {
+	dur := options.MustParseDuration("10s")
+	d, err := dur.DurationFrom(time.Now())
+	if err != nil {
+		t.Fatal(err)
+	}
+	if d != 10*time.Second {
+		t.Fatal("expected duration to be 10s but it was ")
```
mark-rushakoff: Missing `%s", d`?

task/servicetest/servicetest.go
```diff
@@ -937,7 +936,7 @@ func testTaskConcurrency(t *testing.T, sys *System) {
 			// Create a run for the last task we found.
 			// The script should run every minute, so use max now.
 			tid := tasks[len(tasks)-1].ID
-			if _, err := sys.TaskControlService.CreateNextRun(sys.Ctx, tid, math.MaxInt64); err != nil {
+			if _, err := sys.TaskControlService.CreateNextRun(sys.Ctx, tid, math.MaxInt64>>6); err != nil {
```
mark-rushakoff: Is `>>6` needed here?

task/options/options_test.go
```diff
@@ -45,20 +45,37 @@ func scriptGenerator(opt options.Options, body string) string {
...
+func TestNegDurations(t *testing.T) {
+	dur := options.MustParseDuration("-1m")
+	d, err := dur.DurationFrom(time.Now())
+	if err != nil {
+		t.Fatal(err)
+	}
+	if d != -time.Minute {
+		t.Fatalf("expected duration to be -1m but was %s", d)
+	}
+
```
mark-rushakoff: Remove blank line.

task/options/options.go
```diff
@@ -40,22 +45,129 @@ type Options struct {
...
+func MustParseDuration(s string) (dur *Duration) {
+	dur = &Duration{}
+	if err := dur.Parse(s); err != nil {
+		panic(err)
+	}
+	return dur
+}
+
+// parseSignedDuration will convert a string into a possibly negative DurationLiteral.
+// TODO(docmerlin): switch to the one in flux
```
jlapacik: Why not just use the one you added to Flux?
https://github.com/influxdata/flux/blob/master/parser/strconv.go#L22-L37

task/options/options.go
```diff
@@ -80,6 +192,42 @@ const (
...
+}
+
+func grabTaskOptionAST(p *ast.Package, keys ...string) map[string]ast.Expression {
+	res := make(map[string]ast.Expression, 2) // we preallocate two keys for the map, as that is how many we will use at maximum (offset and every)
+	for i := range p.Files {
+		for j := range p.Files[i].Body {
+			if p.Files[i].Body[j].Type() == "OptionStatement" {
+				opt := (p.Files[i].Body[j]).(*ast.OptionStatement)
+				if opt.Assignment.Type() == "VariableAssignment" {
+					asmt := opt.Assignment.(*ast.VariableAssignment)
```
jlapacik: Perhaps the type assertion should happen in the if statement?
```
if asmt, ok := opt.Assignment.(*ast.VariableAssignment); ok {
...
}
```

task/servicetest/servicetest.go
docmerlin: Yes, because math.MaxInt64 is too large and causes errors when converting to a time and then adding durations etc.

task/options/options.go
docmerlin: I think, I am going to switch to using the continue pattern that mark suggested.

task/backend/meta.proto
```diff
@@ -25,7 +25,7 @@ message StoreTaskMeta {
   string effective_cron = 5;
 
   // Task's configured delay, in seconds.
-  int32 offset = 6;
+  string offset = 6;
```
mark-rushakoff: I guess this will need to be called out as a breaking change. I'm not sure what will fail when reading back an existing record that was int32 here. I think it's fine to break in this way rather than worrying about backwards compatibility for now.

task/options/options.go
jlapacik: I agree, that's way more readable. Really nice. I was more referring to performing the check on the actual concrete type vs the string returned by the `Type()` method. I don't expect those methods or the types to be modified. It just seemed like a safer alternative for reducing the chances of unexpected behavior in the future because of an AST change in Flux.

PR #12957, opened 2019-03-27
Update tasks runs list to display 20 latest runs

### LINKED ISSUE #12874: Run logs page should limit to the last 20 runs
as a tasks user
with admin
when i view a list of task runs
then i would like to limit that view to the last 20 records

Palakp41:
Closes #12874 

Sorting was not working since sort key was null in state so updated sort key to "scheduleFor"
Updated runs to filter for 20 runs to display latest 20 runs

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/tasks/components/TaskRunsList.tsx
+++ b/ui/src/tasks/components/TaskRunsList.tsx
@@ -27,7 +27,7 @@ export default class TaskRunsList extends PureComponent<Props, State> {
   constructor(props) {
     super(props)
     this.state = {
-      sortKey: null,
+      sortKey: 'scheduledFor',
       sortDirection: Sort.Descending,
     }
   }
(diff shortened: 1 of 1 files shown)


ui/src/tasks/components/TaskRunsList.tsx
```diff
@@ -104,9 +104,18 @@ export default class TaskRunsList extends PureComponent<Props, State> {
 
   public listRuns = (runs: Run[]): JSX.Element => {
     const {taskID} = this.props
+
+    let recentRuns = []
```
mavarius: You could just use `const recentRuns = runs.slice(19)` here.

Palakp41: I tried that but it messes up the sorting. Doesn't show the latest test runs first.

PR #12958, opened 2019-03-27
feat(ui): Create import overlay to import variables

### LINKED ISSUE #12884: Add ability to import variables
As a user I should be able to import a variable from a template

ischolten:
Closes #12884

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/variables/components/VariableImportOverlay.tsx
+++ b/ui/src/variables/components/VariableImportOverlay.tsx
@@ -1,0 +1,83 @@
+import React, {PureComponent} from 'react'
+import {withRouter, WithRouterProps} from 'react-router'
+import {connect} from 'react-redux'
+
+// Components
+import ImportOverlay from 'src/shared/components/ImportOverlay'
+
+// Actions
+import {
+  createVariableFromTemplate as createVariableFromTemplateAction,
+  getVariables as getVariablesAction,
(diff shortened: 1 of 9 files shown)


ui/src/variables/components/VariableImportOverlay.tsx
```diff
@@ -0,0 +1,83 @@
...
+  private handleImportVariable = async (
+    uploadContent: string,
+    orgID: string
+  ): Promise<void> => {
+    const {createVariableFromTemplate, getVariables} = this.props
+
+    const template = JSON.parse(uploadContent)
+    await createVariableFromTemplate(template, orgID)
+
+    await getVariables()
```
Palakp41: I think it would be better to dispatch the `SetVariable` action from `createVariableFromTemplate` thunk, rather than having an api call to get all variables. What do you think?

ischolten: the problem with this is that other variables can be included in this template. If i have a variable that uses another variable in my query, then when i import this variable, both should be created. Therefore just setting one variable is not enough. but if we just fetch these variables again, we will get all variables created.

Palakp41: Okay, makes sense! Thanks :)

PR #12995, opened 2019-03-28
Add variable to system dashboard template and create at new org

ebb-tide:
Add variable `bucket` to system dashboard template. 
Create system dashboard template at new org creation. 
Remove builder configs from queries (they were wrong)


ui/src/templates/constants/defaultTemplates.ts
```diff
@@ -0,0 +1,989 @@
+export const systemTemplate = () => ({
```
bthesorceror: will these templates eventually live on the backend as a document?

ebb-tide: no we are responsible for them :(

PR #13002, opened 2019-03-28
Exclude links that don't work in cloud

bthesorceror:


ui/src/shared/components/cloud/CloudExclude.tsx
```diff
@@ -0,0 +1,13 @@
+import {PureComponent} from 'react'
+
+export default class extends PureComponent {
```
ischolten: Theres already a component called cloud feature flag that does basically this. should that component be used rather than creating anew one?

PR #13049, opened 2019-04-01
Upgrade flux to the latest version and remove the platform dependency

jsternberg:
This integrates the influxdb 1.x series to the latest version of Flux
and updates the code to use it. It also removes the dependency on
platform and copies the necessary code from storage into the 1.x series
so the dependency is unneeded.

The flux functions specific to 1.x have been moved to the same structure
that flux changed to with having a `stdlib` directory instead of a
`functions` directory. It also adds a `databases()` function that
returns the databases from the meta client.


Gopkg.toml
```diff
@@ -72,12 +72,8 @@
 
 [[constraint]]
   name = "github.com/influxdata/flux"
-  version = "0.12.0"
-
-[[constraint]]
-  name = "github.com/influxdata/platform"
-  revision = "0f79e4ea3248354c789cba274542e0a8e55971db"
+  version = "= 0.23.0"
```
stuartcarnie: 🎉

PR #13060, opened 2019-04-01
Redirect to quartz logout in cloud

bthesorceror:
  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/Logout.tsx
+++ b/ui/src/Logout.tsx
@@ -6,7 +6,7 @@ import {withRouter, WithRouterProps} from 'react-router'
 import {client} from 'src/utils/api'
 
 // Constants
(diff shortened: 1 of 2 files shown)


ui/src/Logout.tsx
```diff
@@ -25,7 +25,7 @@ export class Logout extends PureComponent<Props> {
 
   private handleSignOut = async () => {
     if (CLOUD) {
-      window.location.pathname = CLOUD_SIGNOUT_PATHNAME
+      window.location.href = CLOUD_SIGNOUT_URL
```
OfTheDelmer: Should we also update the following: https://github.com/influxdata/influxdb/blob/916f74569a479fe59a930b031b290411df12844c/ui/src/Signin.tsx#L81

Issue #13076, opened 2019-04-02
Allow getAll templates by orgID


ebb-tide:
Currently a GET to `/documents/templates` accepts an `org` but there is no option to get all by orgID
https://github.com/influxdata/influxdb/blob/83000361c3a9bbb5f164bcf719fc55c4c3a9f95a/http/swagger.yml#L157

PR #13085, opened 2019-04-02
feat(influxdb): get docs by org id

kelwang:
Closes https://github.com/influxdata/influxdb/issues/13076

Added the option of getting docs by org id

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/document_service.go
+++ b/http/document_service.go
@@ -192,10 +192,24 @@ func (h *DocumentHandler) handleGetDocuments(w http.ResponseWriter, r *http.Requ
 		return
 	}
 
-	opt := influxdb.AuthorizedWhere(a)
+	var opt func(influxdb.DocumentIndex, influxdb.DocumentDecorator) ([]influxdb.ID, error)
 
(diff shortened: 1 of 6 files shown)


http/document_service.go
```diff
@@ -194,7 +194,9 @@ func (h *DocumentHandler) handleGetDocuments(w http.ResponseWriter, r *http.Requ
 
 	opt := influxdb.AuthorizedWhere(a)
```
jademcgough: imo would be nice to have a check here to see if both `Org` and `OrgID` and return an error, to avoid ambiguous behavior in that situation.

kelwang:
> See comment ^
> 
> Would also be great to have some testing for this as well, but not a blocker to merging this if it needs to be merged asap. There's no unit testing for the http service currently, but there's integration testing this can be added to.

tests added

http/document_service.go
kelwang: 👍

PR #13088, opened 2019-04-02
Fix security vulnerability when shared secret is blank

### LINKED ISSUE #12927: Password bypass vulnerability
We can authorize by any user. For that we can use jwt token with empty shared_secret.
It's happen because https://github.com/influxdata/influxdb/blob/e2af85d6503ecaaa4dbdbdd9ddad740741dae582/services/httpd/handler.go#L1585-L1602 here we not check that h.Config.SharedSecret is not empty string.
In `Authentication and authorization in InfluxDB` document we not mention anything about `shared-secret` variable.

Also sems like we also not check exp state of token. And once generated token valid forewer.

dgnorton:
Closes #12927

### DIFF
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -240,6 +240,10 @@ func (h *Handler) Open() {
 	}
 	h.accessLogFilters = StatusFilters(h.Config.AccessLogStatusFilters)
 
+	if h.Config.AuthEnabled && h.Config.SharedSecret == "" {
+		h.Logger.Info("Auth is enabled but shared-secret is blank. BearerAuthentication is disabled.")
+	}
+
 	if h.Config.FluxEnabled {
 		h.registered = true
 		prom.MustRegister(h.Controller.PrometheusCollectors()...)
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -1595,6 +1595,11 @@ func authenticate(inner func(http.ResponseWriter, *http.Request, meta.User), h *
 					return
 				}
 			case BearerAuthentication:
+				if h.Config.SharedSecret == "" {
+					atomic.AddInt64(&h.stats.AuthenticationFailures, 1)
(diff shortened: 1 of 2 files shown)


services/httpd/handler.go
```diff
@@ -1595,6 +1595,11 @@ func authenticate(inner func(http.ResponseWriter, *http.Request, meta.User), h *
 					return
 				}
 			case BearerAuthentication:
+				if h.Config.SharedSecret == "" {
+					atomic.AddInt64(&h.stats.AuthenticationFailures, 1)
+					h.httpError(w, "bearer auth disabled", http.StatusUnauthorized)
```
e-dard: Shouldn't this message indicate why the auth failed, i.e., that the provided shared secret was empty?

dgnorton: I debated what message to return to the client and decided less info is better for security. Thought about logging it but that would allow a DoS by filling the log. Maybe I should log it using `sync.Once`?

e-dard: Maybe I'm confused. Isn't this the case where the user has setup bearer auth but they have forgotten to fill out a config value?

mark-rushakoff: There are other existing ways that a malicious unauthenticated user can spam the server logs, especially if HTTP logging is enabled.

I'm fine with using sync.Once here, but I'm also fine looking for a pattern to spread widely on those other cases.

dgnorton: There isn't a separate config setting for enabling bearer auth. If the shared secret is set, it is enabled. If it's not, bearer auth is disabled.

dgnorton:
@mark-rushakoff instead of sync.Once, I added a log entry when the handler opens: https://github.com/influxdata/influxdb/pull/13088/commits/0ab2a30497b58ceb64aca2a8cfb58c712188106f. Look okay?

PR #13089, opened 2019-04-02
fix(ui/variables): filter vars with prefix names

### LINKED ISSUE #12939: Exported variable template should contain all variables that are in the dependency tree
Variable queries can contain other variables. So an exported variable template must include all other variables in the dependency tree. 

https://github.com/influxdata/influxdb/blob/df8a0d0ddd70a2ec3b3ec46ab1303a113998d4d0/ui/src/variables/utils/hydrateVars.ts#L57-L60 
^^ searches all variables to find those that are in variable query.

OfTheDelmer:
Closes #12939 
Closes https://github.com/influxdata/influxdb/issues/13090

Match exported and dashboard var names up to a word boundary so var names like `v.be` don't display as dependent vars of `v.before` 

Also updates the variable exports util to use this word boundary filter to find all dependent vars in the variable graph resolving #13090 

  - [ ] Rebased/mergeable
  - [ ] Tests pass


ui/src/variables/utils/hydrateVars.ts
```diff
@@ -212,16 +227,24 @@ const constVariableValues = (
...
+
+  // reduce children
   for (const child of node.children) {
-    acc.push(child)
-    collectDescendants(child)
+    if (visited.has(child)) {
+      continue
+    }
+    visited.add(child)
+    descendants = descendants.concat(child, collectDescendants(child, visited))
```
chnn: I feel like keeping track of two accumulators (`children`, `visited`) is pretty confusing... could this work like the `collectAncestors` method instead?

```
const collectDescendents = (
  node: VariableNode,
  acc = new Set()
): VariableNode[] => {
  for (const child of node.children) {
    if (!acc.has(child)) {
      acc.add(child)
      collectAncestors(child, acc)
    }
  }

  return [...acc]
}
```

ui/src/variables/constants/index.ts
```diff
@@ -2,3 +2,70 @@ export const OPTION_NAME = 'v'
 export const TIME_RANGE_START = 'timeRangeStart'
 export const TIME_RANGE_STOP = 'timeRangeStop'
 export const WINDOW_PERIOD = 'windowPeriod'
+
+enum FluxBoundaries {
```
chnn: Does this need to be a enum?

ui/src/variables/utils/hydrateVars.test.ts
```diff
@@ -41,16 +41,64 @@ class FakeFetcher implements ValueFetcher {
...
+      const vars = [a, b, c, d, e, f, g, h]
 
       const actual = exportVariables([a], vars)
+      const expected = new Set([a, b, c, e, f, g])
 
-      expect(actual).toEqual([a, b, c])
+      expect(new Set(actual)).toEqual(expected)
+    })
+
+    const examples = [
```
chnn: beautiful

ui/src/variables/utils/hydrateVars.ts
```diff
@@ -35,25 +35,26 @@ interface HydrateVarsOptions {
...
-  allVariables: Variable[]
+  varGraph: VariableNode[]
 ): Variable[] => {
-  const query: string = variable.arguments.values.query
-  const dependencies = allVariables.filter(maybeChild =>
-    query.includes(`${OPTION_NAME}.${maybeChild.name}`)
-  )
-
-  return dependencies
+  const node = varGraph.find(n => n.variable.id === variable.id)
```
chnn: Do you think you could move the export specific stuff to a separate utility module? `findDependentVariables` and `exportVariables`

Just because the `hydrateVars` module is so gnarly as is, I want to keep it as focused as possible

ui/src/variables/constants/index.ts
OfTheDelmer: ahh true it could just be be vars

ui/src/variables/utils/hydrateVars.ts
OfTheDelmer: was thinking that before that last commit. Will do.

ui/src/variables/utils/hydrateVars.ts
OfTheDelmer: works for me.

ui/src/variables/utils/exportVariables.test.ts
```diff
@@ -0,0 +1,65 @@
+import {exportVariables} from 'src/variables/utils/exportVariables'
+// Mocks
+import {createVariable} from 'src/variables/mocks'
+
+describe('exportVariables', () => {
+  test('should find dependent variables', () => {
```
121watts: `it`

ui/src/variables/utils/exportVariables.test.ts
```diff
@@ -0,0 +1,65 @@
...
+  examples.forEach(example => {
+    test(`should filter vars with shared prefix: ${example.name}`, () => {
+      const target = createVariable('target', 'match me!')
+      const partial = createVariable('tar', 'broke!')
+      const vars = [example, target, partial]
+
+      const actual = exportVariables([example], vars)
+
+      expect(actual).toEqual([example, target])
+    })
```
121watts: nice!

PR #13093, opened 2019-04-02
Improve Navbar

alexpaxton:
Making some changes to the navbar in preparation of moving it to Clockface

- Move highlighting logic to container
- Remove `lodash` from presentational components
- Add `testID` props to all presentational components
- Add `className` props to all presentational components
- Make `NavMenuItem` and `NavMenuSubItem` configurable via a `type` prop
  - Can be either `RouterLink` or `HTTPLink`

### Checklist

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/pageLayout/components/NavMenu.tsx
+++ b/ui/src/pageLayout/components/NavMenu.tsx
@@ -9,12 +9,21 @@ import NavMenuSubItem from 'src/pageLayout/components/NavMenuSubItem'
 // Types
 import {ErrorHandling} from 'src/shared/decorators/errors'
 
-interface Props {
+interface PassedProps {
   children: JSX.Element[]
 }
+interface DefaultProps {
+  testID?: string
+}
+
+type Props = PassedProps & Partial<DefaultProps>
 
 @ErrorHandling
 class NavMenu extends PureComponent<Props> {
+  public static defaultProps = {
+    testID: 'nav-menu',
+  }
+
   public static Item = NavMenuItem
   public static SubItem = NavMenuSubItem
 
@@ -23,10 +32,10 @@ class NavMenu extends PureComponent<Props> {
   }
 
   public render() {
-    const {children} = this.props
+    const {children, testID} = this.props
 
     return (
-      <nav className="nav" data-testid="nav">
+      <nav className="nav" data-testid={testID}>
         {children}
       </nav>
     )
--- a/ui/src/pageLayout/components/NavMenu.tsx
+++ b/ui/src/pageLayout/components/NavMenu.tsx
@@ -15,7 +15,7 @@ interface PassedProps {
 }
 
 interface DefaultProps {
-  testID?: string
+  testID: string
 }
 
 type Props = PassedProps & Partial<DefaultProps>
--- a/ui/src/pageLayout/components/NavMenu.tsx
+++ b/ui/src/pageLayout/components/NavMenu.tsx
@@ -11,7 +11,9 @@ import {ErrorHandling} from 'src/shared/decorators/errors'
 
 interface PassedProps {
   children: JSX.Element[]
+  className?: string
 }
+
 interface DefaultProps {
   testID?: string
 }
@@ -34,8 +36,12 @@ class NavMenu extends PureComponent<Props> {
   public render() {
     const {children, testID} = this.props
 
+    const className = this.props.className
+      ? `nav ${this.props.className}`
+      : 'nav'
+
     return (
-      <nav className="nav" data-testid={testID}>
+      <nav className={className} data-testid={testID}>
         {children}
       </nav>
     )
--- a/ui/src/pageLayout/components/NavMenuItem.tsx
+++ b/ui/src/pageLayout/components/NavMenuItem.tsx
@@ -16,8 +16,8 @@ interface PassedProps {
 }
 
 interface DefaultProps {
-  type?: NavMenuType
-  testID?: string
+  type: NavMenuType
+  testID: string
 }
 
 type Props = PassedProps & Partial<DefaultProps>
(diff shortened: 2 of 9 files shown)


ui/src/pageLayout/components/NavMenu.tsx
```diff
@@ -9,12 +9,23 @@ import NavMenuSubItem from 'src/pageLayout/components/NavMenuSubItem'
...
+interface PassedProps {
   children: JSX.Element[]
+  className?: string
 }
 
+interface DefaultProps {
+  testID?: string
+}
+
+type Props = PassedProps & Partial<DefaultProps>
```
ischolten: @121watts did some work with the default props pattern here influxdata/influxdb#12956. in order to follow that pattern, `testID` should be just a required prop in passed props and typescript will interpret the type of the defaultProps.

ui/src/pageLayout/components/NavMenuItem.tsx
```diff
@@ -1,48 +1,81 @@
...
-        <span className={`icon sidebar--icon ${icon}`} />
-      </Link>
-      <div className="nav--item-menu">
-        <Link className="nav--item-header" to={link}>
-          {title}
-        </Link>
-        {children}
+interface DefaultProps {
+  type?: NavMenuType
+  testID?: string
```
ischolten: if these have default props, they probably shouldnt be optional. See other default props comment.

ui/src/pageLayout/components/NavMenuItem.tsx
```diff
@@ -1,48 +1,81 @@
...
+              {title}
+            </Link>
+            {children}
+          </div>
+        </div>
+      )
+    }
+
+    return (
+      <div className={classnames('nav--item', {active})} data-testid={testID}>
```
ischolten: should the classname prop be able to be used in this case too?

ischolten: should the test ids for the cases be the same?

ui/src/pageLayout/components/NavMenuSubItem.tsx
```diff
@@ -1,30 +1,59 @@
...
   title: string
-  link: string
-  location: string
-  highlightPaths: string[]
+  path: string
+  active: boolean
+  className?: string
+}
+
+interface DefaultProps {
```
ischolten: same default props comments as above

ui/src/pageLayout/containers/Nav.test.tsx
```diff
@@ -2,29 +2,38 @@
...
 
+// Constants
+const DASHBOARDS_NAV_ITEM = 'Dashboards'
+const TASKS_NAV_ITEM = 'Tasks'
+
 function Nav(props) {
   const {pathname} = props
+
+  const activateNavItem = (keywords: string[]): boolean => {
+    const parentPath = _.get(pathname.split('/'), '3', '')
```
ischolten: what is this `3`? could you name this variable so its obvious what is happening here?

ui/src/pageLayout/containers/Nav.test.tsx
```diff
@@ -2,29 +2,38 @@
...
+    return keywords.some(path => path === parentPath)
+  }
+
   return (
-    <>
+    <NavMenu>
       <NavMenu.Item
-        title="Dashboards"
-        link="/orgs/036ef4599dddb000/dashboards"
+        title={DASHBOARDS_NAV_ITEM}
```
ischolten: nice

alexpaxton:
@ischolten I realize I could make all the nav components into `PureComponent`s what do you think?

ui/src/pageLayout/containers/Nav.tsx
```diff
@@ -38,63 +39,56 @@ class SideNav extends PureComponent<Props> {
...
+            active={this.activateNavItem(['tokens_tab'])}
           />
         </NavMenu.Item>
         <CloudNav />
       </NavMenu>
     )
   }
+
+  private activateNavItem = (keywords: string[]): boolean => {
+    const parentPath = _.get(location.pathname.split('/'), '3', '')
```
ischolten: same, could you assign this 3 to a descriptive variable.

ui/src/pageLayout/containers/Nav.test.tsx
```diff
@@ -2,29 +2,38 @@
...
 import NavMenu from 'src/pageLayout/components/NavMenu'
 
+// Constants
+const DASHBOARDS_NAV_ITEM = 'Dashboards'
+const TASKS_NAV_ITEM = 'Tasks'
+
 function Nav(props) {
   const {pathname} = props
+
+  const activateNavItem = (keywords: string[]): boolean => {
```
ischolten: it seems weird that this function is being defined in the test as well as the component. should it be moved to a util to be reused. Additionally, should this test include the exact implementation of determining what should be highlighted?

ischolten:
@alexpaxton I think PureComponent

ui/src/pageLayout/components/NavMenuItem.tsx
alexpaxton: ah good catch

alexpaxton: another good catch!

ui/src/pageLayout/containers/Nav.test.tsx
alexpaxton: That's a good call

ui/src/pageLayout/containers/Nav.test.tsx
alexpaxton: I actually have no idea what the 3 does it was there before, I just moved the code to a different place. I'll ask @121watts about it since I think he wrote it

alexpaxton: so it skips the first two items in the URL which are currently always org and org ID

PR #13101, opened 2019-04-02
feat(ui/template/labels): show template labels

### LINKED ISSUE #12907: display labels on templates
Currently, labels on templates are not displayed. They should be displayed.

OfTheDelmer:
Closes #12907 

Display labels for templates, but hold off on adding and removing templates labels in the UI. This is accomplished by adding a `editMode` prop to the `InlineLabels` component. 

  - [ ] Rebased/mergeable
  - [ ] Tests pass

Screen Shot 2019-04-02 at 5 24 35 PM
Clicking also filters
Screen Shot 2019-04-02 at 5 26 24 PM

### DIFF
--- a/ui/src/organizations/components/OrgTemplatesPage.tsx
+++ b/ui/src/organizations/components/OrgTemplatesPage.tsx
@@ -8,6 +8,7 @@ import TemplatesHeader from 'src/templates/components/TemplatesHeader'
 import OrgTemplatesList from 'src/organizations/components/OrgTemplatesList'
 import {ErrorHandling} from 'src/shared/decorators/errors'
 import SearchWidget from 'src/shared/components/search_widget/SearchWidget'
+import GetLabels from 'src/configuration/components/GetLabels'
 
 // Types
 import {TemplateSummary} from '@influxdata/influx'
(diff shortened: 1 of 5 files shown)


ui/src/organizations/components/OrgTemplatesPage.tsx
```diff
@@ -45,20 +46,22 @@ class OrgTemplatesPage extends PureComponent<Props, State> {
...
-          {ts => (
-            <OrgTemplatesList
-              searchTerm={searchTerm}
-              templates={ts}
-              onFilterChange={this.setSearchTerm}
-              onImport={onImport}
-            />
-          )}
-        </FilterList>
+        <GetLabels>
```
chnn: hooray

ui/src/shared/components/inlineLabels/InlineLabels.tsx
```diff
@@ -32,19 +42,22 @@ export default class InlineLabels extends Component<Props> {
...
     return (
       <div className="inline-labels--container">
-        <InlineLabelsEditor
-          labels={labels}
-          selectedLabels={selectedLabels}
-          onAddLabel={onAddLabel}
-          onCreateLabel={onCreateLabel}
-        />
+        {this.isEditable && (
+          <InlineLabelsEditor
```
chnn: If this is a long term thing, it could be nice to factor it out into a few different components, rather than one `InlineLabels` component that conditionally renders different stuff.

Maybe something like

- `EditableInlineLabels` which is composed of an`InlineLabels` and `InlineLabelsEditor`
- `InlineLabels`

Then the `TemplateCard` could render an `InlineLabels` directly.

OfTheDelmer: Went with the enum because I think of three modes we might need to be aware of:

- user can `create` labels and `read`, `add` and `remove` on a resource
- user can `read`, `add` and `remove` labels on a  resource
- user can `read` labels on a resource

I'll create an issue for `InlineLabels` in TechDebt because its component name should clarify it's handling these concerns.

OfTheDelmer: Note: I still have to verify that we'll want the user experience where only an org owner can add/remove labels. I think the intent was to have labels create/delete be restricted but generally allow users to add/remove labels. So I'm not sure atm.

PR #13107, opened 2019-04-02
refactor(query): update testing InfluxQL testing compiler to use new interface

wolffcm:
Instead of using `ReplaceSpecCompiler`, create a new compiler that employs a rule to replace `from` with `fromInfluxJSON` during logical planning.

With this set of changes, most of the Flux tests pass in `influxdb`, with the exception of some formatting issues.


wolffcm:
@aanthony1243 The pure Flux way is nice and if it could be done that way I would prefer it.  For these tests, we are producing a Flux AST via the transpiler.  Maybe at some point in the future we could use `option` to tell Flux so apply a given planner rule.

PR #13122, opened 2019-04-03
ReadRange planner operation and ReadFilter storage call

jlapacik:
Closes #

This PR adds a new physical operation to the planner called ReadRange. It is represented by the `ReadRangePhysSpec`. This new physical operation is constructed from the pattern `from |> range`. In essence it is the result of pushing a 'range' into a 'from'.

This new planner operation maps to a new ReadFilter storage call which performs a simple read with an optional predicate.

  - [ ] Rebased/mergeable
  - [ ] Tests pass


query/stdlib/influxdata/influxdb/operators.go
```diff
@@ -0,0 +1,55 @@
...
+	"fmt"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/plan"
+	"github.com/influxdata/flux/values"
+)
+
+const ReadRangePhysKind = "ReadRangePhysKind"
+
+type ReadRangePhysSpec struct {
```
wolffcm: This is a divergence from how we've named the other `ProcedureSpec`s.  I think I like it though.  If there is a 1:1 correspondence with physical specs and RPCs, maybe it should match the RPC name as close as possible?

query/stdlib/influxdata/influxdb/source.go
```diff
@@ -0,0 +1,158 @@
...
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/memory"
+	"github.com/influxdata/flux/plan"
+	platform "github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/query"
+	"github.com/influxdata/influxdb/tsdb/cursors"
+)
+
```
wolffcm: So everything in this file will be the common parts of the various InfluxDB read operations, is that right?

query/stdlib/influxdata/influxdb/operators.go
jlapacik: Perhaps change `ReadRangePhysSpec` to just `ReadRange`? I wouldn't mind that.

query/stdlib/influxdata/influxdb/source.go
jlapacik: Correct. We will need a `ReadGroupSource` and a source for meta queries which will be placed here.

jlapacik:
@wolffcm yes there will be a 1:1 mapping between physical read operations and RPC calls with one caveat. The planner makes a distinction between ReadRange and ReadFilter, but storage does not. That is, both ReadRange and ReadFilter will map to the same RPC call as there is no difference to how they are answered at storage.

This simplifies the filter push down rule slightly, since a filter will only be able to be pushed down into a ReadRange operation. In the case that a "non-pushable" filter remains after applying the rule, no attempt will be made to push it down to storage.

storage/reads/datatypes/storage_common.proto
```diff
@@ -16,6 +16,9 @@ service Storage {
   // Read performs a read operation using the given ReadRequest
   rpc Read (ReadRequest) returns (stream ReadResponse);
 
+  // ReadFilter performs a filter operation at storage
+  rpc ReadFilter (ReadFilterRequest) returns (stream ReadResponse);
```
stuartcarnie: Presumably you are dropping `Range` (as in `ReadRangeFilter`) given a time range is assumed for all requests and therefore redundant

jlapacik: Yes there's no reason to have a separate ReadRange RPC as everything is a filter from the perspective of storage. The planner will ensure a range is always included in a call to storage or else planning will fail.

PR #13133, opened 2019-04-03
Fix security vulnerability when shared secret is blank

dgnorton:
Backport #13088 from 1.8 to 1.6

### DIFF
--- a/services/httpd/handler_test.go
+++ b/services/httpd/handler_test.go
@@ -227,6 +227,24 @@ func TestHandler_Query_Auth(t *testing.T) {
 		t.Fatalf("unexpected body: %s", body)
 	}
 
+	// Test that auth fails if shared secret is blank.
+	origSecret := h.Config.SharedSecret
+	h.Config.SharedSecret = ""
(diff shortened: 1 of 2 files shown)


services/httpd/handler_test.go
```diff
@@ -227,6 +227,24 @@ func TestHandler_Query_Auth(t *testing.T) {
...
+	signedToken, err = token.SignedString([]byte(h.Config.SharedSecret))
+	if err != nil {
+		t.Fatal(err)
+	}
+	req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", signedToken))
+	w = httptest.NewRecorder()
+	h.ServeHTTP(w, req)
+	if w.Code != http.StatusUnauthorized {
+		t.Fatalf("unexpected status: %d: %s", w.Code, w.Body.String())
+	} else if body := strings.TrimSpace(w.Body.String()); body != `{"error":"bearer auth disabled"}` {
```
jacobmarble: Is this string found in a constant somewhere, or can it be changed to that effect? Simply comparing a string this way usually causes pain months later. Not a big deal.

PR #13138, opened 2019-04-03
fix(reads): HasFieldValueKey searches for "$", not "_value"

jlapacik:
Closes #

The function `HasKeyFieldValue` searched an influxql expression for references to `_value`. But the function `NodeToExpr` takes an influxql expression (protobuf message) and replaces all references to `_value` with `$`.

This change makes it so `HasKeyFieldValue` searches for references to `$`. This fixes a bug where a value predicate would mistakenly be assigned as predicate consisting only of tag values.
https://github.com/influxdata/influxdb/blob/a5478652ba0eefa2c1bef8dc091305f29086d405/storage/readservice/cursor.go#L67

This would cause a series cursor error for any queries that filtered on `_value` like so:

    from() |> range() |> filter(fn: (r) => r._value < 3000)

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/storage/reads/predicate.go
+++ b/storage/reads/predicate.go
@@ -18,6 +18,7 @@ const (
 	fieldKey       = "_field"
 	measurementKey = "_measurement"
 	valueKey       = "_value"
+	fieldRef       = "$"
 )
 
 // NodeVisitor can be called by Walk to traverse the Node hierarchy.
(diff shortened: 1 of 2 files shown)


zeebo:
The changes seem fine, but I don't have a deeper understanding of what's going on here. Does this fix some open bug I can look at?

jlapacik:
@zeebo here is an issue explaining the bug https://github.com/influxdata/influxdb/issues/13159

PR #13139, opened 2019-04-03
feat(ui): store orgID in local storage

### LINKED ISSUE #13062: add orgID to local storage
As a user, when I log back in to the app, I expect to be directed to the last org I was logged in to. 
GetOrganizations component should update local storage as part of routing. 
This may make orgs.org in redux state redundant

ischolten:
Closes #13062

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/shared/containers/RouteToOrg.tsx
+++ b/ui/src/shared/containers/RouteToOrg.tsx
@@ -1,0 +1,46 @@
+// Libraries
+import {PureComponent} from 'react'
+import {connect} from 'react-redux'
+import {WithRouterProps} from 'react-router'
+
+// Types
+import {AppState} from 'src/types'
+import {Organization} from '@influxdata/influx'
+
+// Decorators
+
+interface StateProps {
+  orgs: Organization[]
+  org: {id?: string}
+}
+
+type Props = StateProps & WithRouterProps
+
+class RouteToOrg extends PureComponent<Props> {
+  public componentDidMount() {
+    const {orgs, router, org} = this.props
+
+    // org from local storage
+    if (org && org.id) {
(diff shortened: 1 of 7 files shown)


ui/src/shared/containers/RouteToOrg.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+// Types
+import {AppState} from 'src/types'
+import {Organization} from '@influxdata/influx'
+
+// Decorators
+
+interface StateProps {
+  orgs: Organization[]
+  org: {id?: string}
```
ebb-tide: org has id and name, I'm surprised this isn't causing linter error

ui/src/types/localStorage.ts
```diff
@@ -1,11 +1,13 @@
...
 import {UserSettingsState} from 'src/userSettings/reducers'
+import {OrgsState} from 'src/organizations/reducers/orgs'
 
 export interface LocalStorage {
   VERSION: string
   app: AppState
   ranges: any[]
   variables: VariablesState
   userSettings: UserSettingsState
+  orgs: OrgsState
```
ebb-tide: I think this should be `org:{id:string, name:string}`

ui/src/shared/containers/RouteToOrg.tsx
ischolten: org has an optional id and name but this interface only requires it to have an optional id. Since its just an interface, it doesnt need to match exactly, just meet the expectations of the interface.

ui/src/types/localStorage.ts
ischolten: It uses org state to match the state of what redux expects

PR #13170, opened 2019-04-04
Add task name to Tasks Runs page

### LINKED ISSUE #13046: Task name should be on the runs page
as a user
with admin
when i view the list of task runs
then i would like to know which task i am viewing the runs for.

let's add the task name to the runs page of a task.

Palakp41:
Closes #13046 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/tasks/actions/index.ts
+++ b/ui/src/tasks/actions/index.ts
@@ -458,7 +458,10 @@ export const getRuns = (taskID: string) => async (dispatch): Promise<void> => {
   try {
     dispatch(setRuns([], RemoteDataState.Loading))
 
-    const runs = await client.tasks.getRunsByTaskID(taskID)
+    const [runs] = await Promise.all([
+      client.tasks.getRunsByTaskID(taskID),
+      dispatch(selectTaskByID(taskID)),
+    ])
 
     const runsWithDuration = runs.map(run => {
       const finished = new Date(run.finishedAt)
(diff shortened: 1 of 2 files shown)


ui/src/tasks/components/TaskRunsPage.tsx
```diff
@@ -49,7 +51,9 @@ class TaskRunsPage extends PureComponent<Props> {
         <Page titleTag="Runs">
           <Page.Header fullWidth={false}>
             <Page.Header.Left>
-              <PageTitleWithOrg title="Runs" />
+              <PageTitleWithOrg
+                title={`${_.get(currentTask, 'name', '')} - Runs`}
```
chnn: Should the `- ` be there if the name lookup fails? Might look weird having ` - Runs` as a title.

ui/src/tasks/actions/index.ts
```diff
@@ -458,7 +458,7 @@ export const importTask = (script: string) => async (
 export const getRuns = (taskID: string) => async (dispatch): Promise<void> => {
   try {
     dispatch(setRuns([], RemoteDataState.Loading))
-
+    await dispatch(selectTaskByID(taskID))
```
chnn: Could these be done in parallel?
```
const [runs] = await Promise.all([
  client.tasks.getRunsByTaskID(taskID)
  dispatch(selectTaskByID(taskID))
])
```

ui/src/tasks/components/TaskRunsPage.tsx
Palakp41: I updated it with a getter for title

ui/src/tasks/actions/index.ts
Palakp41: Yes, they don't depend on each other, I updated it

Issue #13063, opened 2019-04-01
Reconcile orgView and config pages


ebb-tide:
As a user when I click on the settings, 
I see a tabbed page view of resources in my organization. 
Please follow orgView patterns instead of configurations patterns. 
This story includes moving some redux state from `orgView.resource` to top level `resource`

- [x] Variables
- [x] Buckets
- [ ] Telegrafs (https://github.com/influxdata/influxdb/issues/13201)
- [ ] Scrapers (https://github.com/influxdata/influxdb/issues/13198)
- [ ] Templates (https://github.com/influxdata/influxdb/issues/13196)
- [x] Tokens
- [ ] Members (https://github.com/influxdata/influxdb/issues/13197)

ebb-tide:
settings should contain resources as described in https://github.com/influxdata/influxdb/issues/12941

PR #13186, opened 2019-04-05
Make variables org centric

ebb-tide:
Closes Variables part of #13063


ui/src/shared/components/ExportOverlay.tsx
```diff
@@ -151,21 +142,14 @@ class ExportOverlay extends PureComponent<Props> {
...
-    const template = addOrgIDToTemplate(resource, orgID)
-
-    try {
-      await client.templates.create(template)
-      notify(resourceSavedAsTemplate(resourceName))
-    } catch (error) {
-      notify(saveResourceAsTemplateFailed(resourceName, error))
-    }
+    const {resource, onDismissOverlay, resourceName} = this.props
+    createTemplateFromResource(resource, resourceName)
```
ischolten: woot

ui/src/templates/actions/index.ts
```diff
@@ -115,6 +114,22 @@ export const createTemplate = (template: DocumentCreate) => async dispatch => {
   }
 }
 
+export const createTemplateFromResource = (
```
ischolten: Noice!

ui/src/variables/actions/index.ts
```diff
@@ -141,11 +145,16 @@ export const selectValue = (
...
+  getState: GetState
+) => {
   try {
     dispatch(setVariables(RemoteDataState.Loading))
-
-    const variables = await client.variables.getAll()
+    const {
+      orgs: {org},
+    } = getState()
+    const variables = (await client.variables.getAll(org.id)) as IVariable[]
```
ischolten: why do we need type coercion here? does get all not return it as the type we need?

ui/src/variables/components/VariableImportOverlay.tsx
```diff
@@ -40,23 +32,19 @@ class VariableImportOverlay extends PureComponent<Props> {
...
-    orgID: string
+    uploadContent: string
   ): Promise<void> => {
-    const {createVariableFromTemplate, getVariables} = this.props
+    const {createVariableFromTemplate} = this.props
 
     const template = JSON.parse(uploadContent)
-    await createVariableFromTemplate(template, orgID)
-
-    await getVariables()
```
ischolten: is this no longer required to be able to see all imported variables?

ebb-tide: with variable templates including other variables, it's probably better to getVariables.

ui/src/variables/actions/index.ts
ebb-tide: this needs to be fixed in client

PR #13208, opened 2019-04-05
feat(http): add document labe handler

kelwang:
Closes https://github.com/influxdata/influxdb/issues/12937

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/document_service.go
+++ b/http/document_service.go
@@ -18,6 +18,7 @@ type DocumentBackend struct {
 	Logger *zap.Logger
 
 	DocumentService influxdb.DocumentService
+	LabelService    influxdb.LabelService
 }
 
 // NewDocumentBackend returns a new instance of DocumentBackend.
@@ -25,6 +26,7 @@ func NewDocumentBackend(b *APIBackend) *DocumentBackend {
 	return &DocumentBackend{
 		Logger:          b.Logger.With(zap.String("handler", "document")),
 		DocumentService: b.DocumentService,
+		LabelService:    b.LabelService,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


http/document_service.go
```diff
@@ -58,6 +64,10 @@ func NewDocumentHandler(b *DocumentBackend) *DocumentHandler {
 	h.HandlerFunc("PUT", documentPath, h.handlePutDocument)
 	h.HandlerFunc("DELETE", documentPath, h.handleDeleteDocument)
 
+	h.HandlerFunc("GET", documentLabelsPath, h.handleGetDocumentLabel)
```
jademcgough: Other services are using a function factory here (e.g. `newGetLabelsHandler`). Is it possible to use it here as well?

http/swagger.yml
```diff
@@ -226,6 +226,98 @@ paths:
...
+          description: unexpected error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
+  '/documents/templates/{templateID}/labels/{labelID}':
+    delete:
+      tags:
+        - Templates
+      summary: delete a label from a telegraf config
```
jademcgough: telegraf config -> document

http/document_service.go
```diff
@@ -260,46 +270,145 @@ func decodeGetDocumentsRequest(ctx context.Context, r *http.Request) (*getDocume
...
+		EncodeError(ctx, err, w)
+		return
+	}
+
+	label, err := h.LabelService.FindLabelByID(ctx, req.Mapping.LabelID)
 	if err != nil {
 		EncodeError(ctx, err, w)
 		return
 	}
+	d.Labels = append(d.Labels, label)
```
jademcgough: I think Labels on the document struct are intended to be read-only and stored in the labels service, not as part of the document.

http/document_service.go
```diff
@@ -260,46 +270,145 @@ func decodeGetDocumentsRequest(ctx context.Context, r *http.Request) (*getDocume
...
+	if err := s.UpdateDocument(ctx, d, influxdb.Authorized(a)); err != nil {
```
jademcgough: This can be removed. There's no need to update the document, labels are not stored on it.

http/document_service.go
```diff
@@ -260,46 +270,145 @@ func decodeGetDocumentsRequest(ctx context.Context, r *http.Request) (*getDocume
...
 	}
+	// remove label from document
+	for k, l := range d.Labels {
+		if req.LabelID == l.ID {
+			d.Labels = append(d.Labels[:k], d.Labels[k+1:]...)
+			break
+		}
+	}
+
+	if err := s.UpdateDocument(ctx, d, influxdb.Authorized(a)); err != nil {
```
jademcgough: See above comment. The document does not need to be updated; labels aren't stored within it.

PR #13219, opened 2019-04-08
refactor(task): make tasks use new Query interface

### LINKED ISSUE #12883: Update Tasks for private Spec
The flux.Spec is being made private see https://github.com/influxdata/flux/issues/1085. We need to ensure that the uses from within Tasks use the AST version of the API endpoints instead of the Spec.

wolffcm:
Fixes #12883.

@mark-rushakoff I included you as a reviewer for this PR since it makes changes to the tasks backend.

These changes targeting the feature branch `refactor/private-spec` and update the tasks backend and pipeline tests to work with the new `Query` interface:
https://github.com/influxdata/flux/blob/refactor/private-spec/query.go#L7-L29
This is part of a broader issue to get dynamic queries working:
https://github.com/influxdata/flux/issues/1090

These changes pass all the go unit tests in `influxdb`.  `idpe` is not yet updated so the litmus tests are still failing there.

  - [ ] Rebased/mergeable
  - [ ] Tests pass


task/backend/executor/executor.go
```diff
@@ -303,38 +303,43 @@ func (p *asyncRunPromise) followQuery(wg *sync.WaitGroup) {
...
+SelectLoop:
+	for {
+		select {
+		case <-p.ready:
+			// The promise was finished somewhere else, so we don't need to call p.finish.
+			// But we do need to cancel the flux. This could be a no-op.
+			p.q.Cancel()
 			return
+		case r, ok := <-p.q.Results():
+			if ok {
```
mark-rushakoff: Minor style nit: I would probably go with

```go
if !ok {
  break SelectLoop
}

rwg.Add(1)
// ...
```

just to reduce indentation in the `ok` path.

PR #13244, opened 2019-04-08
Fix broken profile link

### LINKED ISSUE #13239: Remove link to profile on the homepage
Clicking the profile link (under `Account`) in `/me` should open an overlay with the form that was previously in `Settings`

alexpaxton:
Closes #13239 

Clicking `Profile` link under "Account" header in me page opens an overlay with the form in it
Route for updating a user's name is no longer present so the form is disabled until that is back

Screen Shot 2019-04-08 at 2 47 01 PM

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/index.tsx
```diff
@@ -121,6 +122,9 @@ class Root extends PureComponent {
                         <Route path="new" component={CreateOrgOverlay} />
                         <Route path=":orgID" component={SetOrg}>
                           <IndexRoute component={MePage} />
+                          <Route path="" component={MePage}>
+                            <Route path="profile" component={ProfileOverlay} />
```
alexpaxton: Threw this in here so the profile overlay has something in the background. It's a bit of a workaround because you can't give `IndexRoute` a child

chnn: Could the `IndexRoute` be removed altogether?
```
 <Route path=":orgID" component={SetOrg}>
   <Route path="" component={MePage}>
     <Route path="profile" component={ProfileOverlay} />
```
Not sure, I'm not very familiar with `IndexRoute`s.

ui/src/me/components/ProfileOverlay.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+interface State {
+  me: AppState['me']
+}
+
+export class Settings extends PureComponent<Props, State> {
+  constructor(props) {
+    super(props)
+    this.state = {
+      me: this.props.me,
```
chnn: This is temporary placeholder, right? Until the form is no longer disabled?

alexpaxton: Yeah basically leaving this here so when the update user route is back this will be an easy piece of work

ui/src/index.tsx
alexpaxton: I have no idea

alexpaxton: @bthesorceror thoughts?

bthesorceror: your overlay route should be a child of the IndexRoute

ui/src/me/components/ProfileOverlay.tsx
bthesorceror: If its really just for a later scenario, remove it. We can pull it back out of history later.

ui/src/me/components/ProfileOverlay.tsx
```diff
@@ -0,0 +1,84 @@
...
+// Components
+import {Form, Input, Button, ComponentColor} from '@influxdata/clockface'
+import {Overlay} from 'src/clockface'
+
+// Types
+import {AppState} from 'src/types'
+import {ComponentSize, ComponentStatus} from '@influxdata/clockface'
+
+interface StateProps {
+  me: AppState['me']
```
bthesorceror: This is a good example of where we should use a small interface that represents what we need for the component rather than pulling a type from AppState.

alexpaxton:
Closed in favor of https://github.com/influxdata/influxdb/pull/13551

PR #13245, opened 2019-04-08
refactor(variables): update CreateVariableOverlay to use routes

### LINKED ISSUE #13243: Create Variable Overlay should use React Router
Selecting "New Variable" from the `AddResourceDropdown` in the "Manage Variables" page should navigate to the variables/new route. Currently it uses open/close state managed by the `VariablesTab` component.

AlirieGray:
Closes #13243

This PR changes the `CreateVariableOverlay` to use React Router rather than open/close state in the parent component.

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/index.tsx
```diff
@@ -208,6 +209,10 @@ class Root extends PureComponent {
                               path=":id/export"
                               component={VariableExportOverlay}
                             />
+                            <Route
+                              path="new"
```
alexpaxton: I haven’t checked but I am wondering if “new” is consistent with the other Create overlays

alexpaxton:
I think the failing e2e test is in master

ui/src/index.tsx
AlirieGray: seems to be!

Issue #13215, opened 2019-04-08
implement delete with predicate endpoint


jademcgough:
- [ ] wire `DeleteBucketRangePredicate` up to the API
- [ ] cli variant

PR #13254, opened 2019-04-09
feat(http): stub out delete with predicate endpoint

jademcgough:
Ref #13215


jademcgough:
Closing this as stale. I think @desa has done work since this that makes this redundant.

Issue #13251, opened 2019-04-09
Implement storage schema RPC APIs


stuartcarnie:
Dependent on stubs from #13241

## High level tasks

- [x] Extend the `Storage` service definition

The `Storage` service defined in the [protobuf definition](https://github.com/influxdata/influxdb/blob/e9600b1f0be0cced84f376e0a633998e78970bd0/storage/reads/datatypes/storage_common.proto#L15) will need to expose the new `TagKeys` and `TagValues` APIs. These APIs should return a `stream` similar to the [`Read`](https://github.com/influxdata/influxdb/blob/e9600b1f0be0cced84f376e0a633998e78970bd0/storage/reads/datatypes/storage_common.proto#L17) API in order to handle very large responses in an efficient manner. The chunked response would likely be slices of byte slices (`[][]byte`) representing the tag or key values.

----

- [x] Extend the `reads.Store` interface with new `TagKeys` and `TagValues` APIs

The [`reads.Store`](https://github.com/influxdata/influxdb/blob/d0603457b7a2ef876e62364d8667cc7a97e7ca05/storage/reads/store.go#L77-L81) interface is intended to abstract the various physical operations offered by the storage layer. The `TagKeys` and `TagValues` APIs should be added to this interface. The implementation of this interface exists in OSS and Cloud 2.0.

Note that this API returns interfaces also defined in the `reads` package. These abstractions  typically follow an iterator pattern with a `Next` API. The iterator approach allows us to implement in-process or inter-process responses efficiently, as the internal details can batch network requests as appropriate.

See the `reads.ResultSet` details below which walks through the implementation for in-process and inter-process behavior. The process should be similar for the `TagKeys` and `TagValues` APIs, albeit much simpler, given these APIs are ultimately streaming a sequence of string or byte slices. 

The result returned by `TagKeys` or `TagValues` via  `storage.Engine` will be an string iterator similar to:

```go
type StringIterator interface {
    Next() bool    // Returns false it iteration has concluded
    Value() string // Returns the current value.
}
```

A similar definition could exist in the `reads` package that can be reused whenever an RPC needs to return a sequence of strings.

----

- [x] Extend [`readservice.store`](https://github.com/influxdata/influxdb/blob/d0603457b7a2ef876e62364d8667cc7a97e7ca05/storage/readservice/store.go#L17) to implement refactored `reads.Store` interface

This is the OSS implementation of the `reads.Store` interface. A separate issue will be created to track the Cloud 2.0 work.

[...]

PR #13263, opened 2019-04-09
Add TagKeys and TagValues methods to Storage service

desa:
Link #13251

TODO:

- [x] Run `go generate ./storage/...`

This PR extends the `Storage` service interface by adding two methods `TagKeys` and `TagValues`. Both methods share request and response structures as they are particularly similar. ~The `TagsResponse` message shares a structure similar to that of the `ReadResponse`, but without all of the other frame types.~

  - [ ] Rebased/mergeable
  - [ ] Tests pass


storage/reads/datatypes/storage_common.proto
```diff
@@ -208,6 +214,22 @@ message TimestampRange {
...
+  Predicate predicate = 3;
+}
+
+// Response message for Storage.TagKeys and Storage.TagValues.
+message TagsResponse {
+  message Frame {
+    repeated bytes values = 1;
+  }
+
+  repeated Frame frames = 1 [(gogoproto.nullable) = false];
```
stuartcarnie: nit: I don't believe we'll need this inner `Frame` message; we can probably just include the values property in the `TagsResponse`.

desa: 👍

jlapacik:
Should a request for tag values include an explicit field for the tag key(s) being queried?

stuartcarnie:
> Should a request for tag values include an explicit field for the tag key(s) being queried?

@jlapacik per the original RFC, Edd proposed the following API:

```go
// TagValues returns an iterator of tag values for the bucket matching predicate within the time range (start, end].
func (e *Engine) TagValues(orgID, bucketID influxdb.ID, start, end int64, predicate influxql.Expr) StringIterator {
    // This method would be invoked when the consumer wants to get the following schema information for an arbitrary
    // time range in a single bucket:
    // 
    // * All measurement names;
    // * All field keys;
    // * All tag values filtered by an arbitrary predicate, e.g., tag values for a measurement and tag key.
    //
}
```

It does not mention a tag key for which set of tag values to produce. @desa has undoubtedly based his definition from this. I suspect we will need to specify a tag key. It technically could be expressed in the predicate with a sentinel like `TAGKEY == "foo_key"`, which is similar to the old metadata queries

desa:
Closing PR in favor of https://github.com/influxdata/influxdb/pull/13263

Issue #13278, opened 2019-04-09
[pvt-spec] Re-implement BucketsAccessed to respect private flux.Spec


wolffcm:
The `BucketsAccessed` method looks like this:
```go
func BucketsAccessed(q *flux.Spec, orgID *platform.ID) (readBuckets, writeBuckets []platform.BucketFilter, err error)
```
However there is no way to get at a `flux.Spec` from a Flux script in the `influxdb` repo anymore.  The `BucketsAccessed` is needed by the `PreAuthorization` interface.

`BucketsAccessed` should be re-implemented to accept an `ast.Package` and probably passed to a helper method in the `flux` repo that can visit each node in the corresponding Spec.

PR #13282, opened 2019-04-09
refactor(query): allow code to compile despite BucketsAccessed

### LINKED ISSUE #13275: [pvt-spec] Preauthorizer should accept an AST not a Spec

wolffcm:
BucketsAccessed doesn't work currently with a private flux.Spec.
See this issue: https://github.com/influxdata/influxdb/issues/13278

This set of changes just allows code to compile until #13278 is fixed.
Note that preauthorization is not working in the meantime.

Fixes #13275.

  - [ ] Rebased/mergeable
  - [ ] Tests pass


query/querytest/compile.go
```diff
@@ -34,43 +30,9 @@ var opts = append(
...
-		if diagnostic := verifyBuckets(*tc.WantReadBuckets, gotReadBuckets); diagnostic != "" {
-			t.Errorf("Could not verify read buckets: %v", diagnostic)
-		}
-	}
-
-	if tc.WantWriteBuckets != nil {
-		if diagnostic := verifyBuckets(*tc.WantWriteBuckets, gotWriteBuckets); diagnostic != "" {
-			t.Errorf("Could not verify write buckets: %v", diagnostic)
-		}
-	}
```
affo: isn't it better to comment out these lines to not lose track of them?

query/preauthorizer.go
```diff
@@ -2,8 +2,7 @@ package query
 
 import (
 	"context"
-
-	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/ast"
```
affo: nit, ordered imports

query/querytest/compile.go
wolffcm: Yes, I think you're right.

Issue #13274, opened 2019-04-09
Bucket.organization and scrapper organization rename to org


kelwang:
part of 
https://github.com/influxdata/influxdb/issues/13148
and
https://github.com/influxdata/influxdb/issues/13147

bednar:
Hi @kelwang,

can you also change the definition in swagger for the `Bucket`?

```yaml
    Bucket:
      properties:
        links:
          type: object
          readOnly: true
          example:
            labels: "/api/v2/buckets/1/labels"
            logs: "/api/v2/buckets/1/logs"
            members: "/api/v2/buckets/1/members"
            org: "/api/v2/orgs/2"
            owners: "/api/v2/buckets/1/owners"
            self: "/api/v2/buckets/1"
            write: "/api/v2/write?org=2&bucket=1"
          properties:
            labels:
              description: URL to retrieve labels for this bucket
              $ref: "#/components/schemas/Link"
            logs:
              description: URL to retrieve operation logs for this bucket
              $ref: "#/components/schemas/Link"
            members:
              description: URL to retrieve members that can read this bucket
              $ref: "#/components/schemas/Link"
            org:
              description: URL to retrieve parent organization for this bucket
              $ref: "#/components/schemas/Link"
            owners:
              description: URL to retrieve owners that can read and write to this bucket.
              $ref: "#/components/schemas/Link"
            self:
              description: URL for this bucket
              $ref: "#/components/schemas/Link"
            write:
              description: URL to write line protocol for this bucket
              $ref: "#/components/schemas/Link"
        id:
          readOnly: true
          type: string
        name:
          type: string
        organizationID:
          type: string
        organization:
          type: string
        rp:
          type: string
        retentionRules:
          type: array
          description: rules to expire or retain data.  No rules means data never expires.
          items:
            type: object
            properties:
              type:
                type: string
                default: expire
                enum:
                  - expire
              everySeconds:
                type: integer
                description: duration in seconds for how long data will be kept in the database.
                example: 86400
                minimum: 1
            required: [type, everySeconds]
        labels:
          $ref: "#/components/schemas/Labels"
      required: [name, retentionRules]
```

should be:

```yaml
    Bucket:
      properties:
        links:
          type: object
          readOnly: true
          example:
            labels: "/api/v2/buckets/1/labels"
            logs: "/api/v2/buckets/1/logs"
            members: "/api/v2/buckets/1/members"
            org: "/api/v2/orgs/2"
[...]

PR #13286, opened 2019-04-10
Bucket scrapper org rename

kelwang:
Closes https://github.com/influxdata/influxdb/issues/13274

Rename to Organization to Org
Added Migration Service
Added Migration Test Cases

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/bucket.go
+++ b/bucket.go
@@ -22,7 +22,7 @@ const InfiniteRetention = 0
 type Bucket struct {
 	ID                  ID            `json:"id,omitempty"`
 	OrganizationID      ID            `json:"orgID,omitempty"`
-	Organization        string        `json:"organization,omitempty"`
+	Org                 string        `json:"org,omitempty"`
 	Name                string        `json:"name"`
 	RetentionPolicyName string        `json:"rp,omitempty"` // This to support v1 sources
 	RetentionPeriod     time.Duration `json:"retentionPeriod"`
@@ -73,7 +73,7 @@ type BucketFilter struct {
 	ID             *ID
 	Name           *string
 	OrganizationID *ID
-	Organization   *string
+	Org            *string
 }
 
 // QueryParams Converts BucketFilter fields to url query params.
@@ -91,8 +91,8 @@ func (f BucketFilter) QueryParams() map[string][]string {
 		qp["orgID"] = []string{f.OrganizationID.String()}
 	}
 
-	if f.Organization != nil {
-		qp["org"] = []string{*f.Organization}
+	if f.Org != nil {
+		qp["org"] = []string{*f.Org}
 	}
 
 	return qp
@@ -112,8 +112,8 @@ func (f BucketFilter) String() string {
 	if f.OrganizationID != nil {
 		parts = append(parts, "Org ID: "+f.OrganizationID.String())
 	}
-	if f.Organization != nil {
-		parts = append(parts, "Org Name: "+*f.Organization)
+	if f.Org != nil {
+		parts = append(parts, "Org Name: "+*f.Org)
 	}
 	return "[" + strings.Join(parts, ", ") + "]"
 }
--- a/bucket.go
+++ b/bucket.go
@@ -21,8 +21,7 @@ const InfiniteRetention = 0
 // Bucket is a bucket. 🎉
 type Bucket struct {
 	ID                  ID            `json:"id,omitempty"`
-	OrganizationID      ID            `json:"orgID,omitempty"`
-	Org                 string        `json:"org,omitempty"`
+	OrgID               ID            `json:"orgID,omitempty"`
 	Name                string        `json:"name"`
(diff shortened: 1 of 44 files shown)


bucket.go
```diff
@@ -22,7 +22,7 @@ const InfiniteRetention = 0
 type Bucket struct {
 	ID                  ID            `json:"id,omitempty"`
 	OrganizationID      ID            `json:"orgID,omitempty"`
-	Organization        string        `json:"organization,omitempty"`
+	Org                 string        `json:"org,omitempty"`
```
desa: Do we even need this field on the Bucket struct? We had initially added it because we wanted to present things nicely going out the door for the API, but we've since developed patterns for doing presentation in the http layer. Might be worthwhile to simply remove this field all together.

cmd/influx/bucket.go
```diff
@@ -181,15 +181,15 @@ func bucketFindF(cmd *cobra.Command, args []string) error {
...
 		"OrganizationID",
 	)
 	for _, b := range buckets {
 		w.Write(map[string]interface{}{
 			"ID":             b.ID.String(),
 			"Name":           b.Name,
 			"Retention":      b.RetentionPeriod,
-			"Organization":   b.Organization,
+			"Org":            b.Org,
 			"OrganizationID": b.OrganizationID.String(),
```
desa: Should this be `OrgID`?

data_migration.go
```diff
@@ -0,0 +1,53 @@
+package influxdb
```
desa: I'm concern that this pattern isn't going to be something that works long term. I imagine that this will likely be something we repeat with some regularity.

data_migration.go
```diff
@@ -0,0 +1,53 @@
...
+
+// BucketMigrationService will migrate old bucket to the most recent bucket schema.
+type BucketMigrationService interface {
+	IsBucketMigrated(ctx context.Context) bool
+	ConvertBucketToNew(ctx context.Context) error
+}
+
+// OldBucket should includes all old fields of previous bucket schemas,
+// as well as the new fields incase of any partial conversion.
+type OldBucket struct {
```
desa: What happens when we have to change the schema of buckets again?

data_migration.go
```diff
@@ -0,0 +1,53 @@
...
+	// CheckAndMigrate will determine if data already migrated.
+	// Usually we will put an index to make sure this the newest version.
+	// Then do a scan to the storage and convert every thing related.
+	CheckAndMigrate(ctx context.Context) error
+	BucketMigrationService
+}
+
+// BucketIsMigratedIndex will be the index check to determine
+// if the newest bucket schema is applied.
+var BucketIsMigratedIndex = []byte("bucketIsMigrated_org")
```
desa: Why is this defined here?

kv/data_migration.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+		return nil
+	}); err != nil {
+		return false
+	}
+	return true
+}
+
+// ConvertBucketToNew to do a scan to the storage and convert every thing related.
+func (s *Service) ConvertBucketToNew(ctx context.Context) error {
```
desa: I'm concerned about this pattern for cloud. I imagine we're going to need a more sophisticated migration strategy as we will likely have to have instances that are using both the new schema and the old schema concurrently. Additionally, this is likely contain too many operations to do in a single etcd transaction.

cmd/influxd/launcher/launcher.go
```diff
@@ -410,6 +410,12 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 		return err
 	}
 
+	// use unlimited context for data migration for now.
+	if err := m.kvService.CheckAndMigrate(context.Background()); err != nil {
```
desa: What happens if this operation fails dues to programmer error and leaves the user database in a malformed state?

data_migration.go
kelwang: @desa, you can simple replace the BucketIsMigratedIndex with a new value, for example: bucketIsMigrated_orgID. any deleted value, or new added value has be show up in the oldBucket.

data_migration.go
kelwang: Because the reason above. This is should be changed any time, there is a schema change. So it should not global for all storage service.

cmd/influx/bucket.go
kelwang: Yeah, I believe there is another ticket for that one.

bucket.go
kelwang: Yeah, I agree. We should remove it, in that case.

kelwang:
@desa, I will pull out the migrator from this PR, since it's not really required for now.
And a new issue for the migrator: https://github.com/influxdata/influxdb/issues/13404,

PR #13291, opened 2019-04-10
fix: encode IDs as JSON map keys properly

### LINKED ISSUE #13277: influxdb.ID not properly JSON-encoded as map keys
Consider this updated test (applied against master at 6b300ae30d817d99aa901e8a4ace4bdf132667df):

```diff
diff --git a/id_test.go b/id_test.go
index 13bf7ce17..9d3f7ebfd 100644
--- a/id_test.go
+++ b/id_test.go
@@ -7,6 +7,7 @@ import (
        "reflect"
        "testing"

+       "github.com/influxdata/influxdb"
        platform "github.com/influxdata/influxdb"
        platformtesting "github.com/influxdata/influxdb/testing"
 )
@@ -181,6 +182,28 @@ func TestMarshalling(t *testing.T) {
        if !bytes.Equal(bytes1, bytes2) {
                t.Errorf("error marshalling/unmarshalling ID")
        }
+
+       // When used as a map key, IDs must use their string encoding.
+       // If you only implement json.Marshaller, they will be encoded with Go's default integer encoding.
+       b, err := json.Marshal(map[influxdb.ID]int{0x1234: 5678})
+       if err != nil {
+               t.Fatal(err)
+       }
+       const exp = `{"0000000000001234":5678}`
+       if string(b) != exp {
+               t.Fatalf("expected map to json.Marshal as %s; got %s", exp, string(b))
+       }
+
+       var idMap map[influxdb.ID]int
+       if err := json.Unmarshal(b, &idMap); err != nil {
+               t.Fatal(err)
+       }
+       if len(idMap) != 1 {
+               t.Fatalf("expected length 1, got %d", len(idMap))
+       }
+       if idMap[0x1234] != 5678 {
+               t.Fatalf("unmarshalled incorrectly; exp 0x1234:5678, got %v", idMap)
+       }
 }
```

This test is desirable so that encoding IDs as map keys will be human-readable, in the format we're trained to read them everywhere else.

id.go as-is will fail:

```
--- FAIL: TestMarshalling (0.00s)
    id_test.go:194: expected map to json.Marshal as {"0000000000001234":5678}; got {"4660":5678}
```

Given the notes on map keys in encoding/json:

> The map's key type must either be a string, an integer type, or implement encoding.TextMarshaler. The map keys are sorted and used as JSON object keys by applying the following rules, subject to the UTF-8 coercion described for string values above:
> 
> - string keys are used directly
> - encoding.TextMarshalers are marshaled
> - integer keys are converted to strings

It seems like the naive solution would be to replace ID's MarshalJSON with an equivalent MarshalText:

```go
// MarshalText implements text marshaller for IDs.
//
// Note that by implementing MarshalText, json.Marshaler is also satisfied.
// This also has the desired effect of using the string encoding for map keys,
// instead of the default integer encoding.
func (i ID) MarshalText() ([]byte, error) {
	enc, err := i.Encode()
	if err != nil {
		return nil, err
	}
	return enc, nil
}
```

But then other tests start failing in a way I don't immediately understand: `invalid use of ,string struct tag, trying to unmarshal "020f755c3c082001" into influxdb.ID`

I'm skeptical of the current implementation of UnmarshalJSON:

```go
// UnmarshalJSON implements JSON unmarshaller for IDs.
func (i *ID) UnmarshalJSON(b []byte) error {
	if b[0] == '"' {
		b = b[1:]
	}

	if b[len(b)-1] == '"' {
		b = b[:len(b)-1]
	}

	return i.Decode(b)
}
```

Conditionally trimming the quotes can't be correct. I think this may be masking some other incorrect behavior related to the above `invalid use of ,string struct tag` test failure.

I suspect fixing the map key encoding behavior will be a breaking change one way or another, given the likelihood of IDs being JSON-encoded somewhere in bolt. We should fix this (or decide not to fix it) sooner rather than later.

mark-rushakoff:
The encoding/json docs explain that you need to provide a MarshalText
method to encode integer types as map keys, otherwise they will be
formatted as a string version of the decimal number.

Providing MarshalText and UnmarshalText also uses those methods as a
fallback for MarshalJSON and UnmarshalJSON, so we no longer need
explicit versions of those latter methods.

Apparently Sources were using IDs as map keys and were providing the
,string attribute on the JSON tag on the struct. This was not correct so
that attribute has been removed. Existing sources will no longer be
readable as a result of this change.

Fixes #13277.

### DIFF
--- a/id.go
+++ b/id.go
@@ -3,7 +3,6 @@ package influxdb
 import (
 	"encoding/binary"
 	"encoding/hex"
-	"encoding/json"
 	"errors"
 	"reflect"
 	"strconv"
@@ -119,24 +118,17 @@ func (i ID) GoString() string {
 	return `"` + i.String() + `"`
 }
 
-// UnmarshalJSON implements JSON unmarshaller for IDs.
-func (i *ID) UnmarshalJSON(b []byte) error {
-	if b[0] == '"' {
-		b = b[1:]
-	}
-
-	if b[len(b)-1] == '"' {
-		b = b[:len(b)-1]
-	}
-
-	return i.Decode(b)
+// MarshalText encodes i as text.
+// Providing this method is a fallback for json.Marshal,
+// with the added benefit that IDs encoded as map keys will be the expected string encoding,
+// rather than the effective fmt.Sprintf("%d", i) that json.Marshal uses by default for integer types.
+func (i ID) MarshalText() ([]byte, error) {
+	return i.Encode()
 }
 
-// MarshalJSON implements JSON marshaller for IDs.
-func (i ID) MarshalJSON() ([]byte, error) {
-	enc, err := i.Encode()
-	if err != nil {
-		return nil, err
-	}
-	return json.Marshal(string(enc))
+// UnmarshalText decodes i from a byte slice.
+// Providing this method is also a fallback for json.Unmarshal,
+// also relevant when IDs are used as map keys.
+func (i *ID) UnmarshalText(b []byte) error {
+	return i.Decode(b)
 }
--- a/id_test.go
+++ b/id_test.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/influxdata/influxdb"
 	platform "github.com/influxdata/influxdb"
 	platformtesting "github.com/influxdata/influxdb/testing"
 )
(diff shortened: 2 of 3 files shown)


mark-rushakoff:
I have been told that Sources can be removed (https://github.com/influxdata/influxdb/issues/13325), so removing the `,string` part from the JSON tag on the Source struct should be noncontroversial.

jademcgough:
I'm nervous about merging a breaking change like this shortly before our upcoming milestone, but it might be better to get this in sooner rather than later, to avoid having to do a migration.

PR #13293, opened 2019-04-10
Update variables to use view overlay instead of export overlay

### LINKED ISSUE #12914: Overlay to view template should not use export overlay
As a user when I click on a template's name, 
I see an overlay, that looks a lot like the export resource overlay (what shows up now) but should be its own separate component distinct from exportOverlay. 
The overlay should not have export written at the top left. 
There should be no save as template button at the bottom. 
Download as json and Copy to clipboard buttons should remain

Palakp41:
Closes #12914

Created a generic component for ViewOverlay
Created a TemplateViewOverlay that gets the template and passes into ViewOverlay as a resource.
Updated the route for viewing templates to `/view`

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/components/ViewOverlay.tsx
+++ b/ui/src/shared/components/ViewOverlay.tsx
@@ -1,0 +1,106 @@
+// Libraries
+import React, {PureComponent} from 'react'
+
+// Components
+import {Button, SpinnerContainer, TechnoSpinner} from '@influxdata/clockface'
+import {Overlay, ComponentSize} from 'src/clockface'
+import {Controlled as ReactCodeMirror} from 'react-codemirror2'
+import CopyButton from 'src/shared/components/CopyButton'
+
+// Types
+import {DocumentCreate} from '@influxdata/influx'
+import {ComponentColor} from '@influxdata/clockface'
+import {RemoteDataState} from 'src/types'
+
+interface Props {
+  onDismissOverlay: () => void
+  resource: DocumentCreate
+  overlayHeading: string
+  status: RemoteDataState
+  isVisible: boolean
+}
+
+export default class ViewOverlay extends PureComponent<Props> {
+  public static defaultProps = {
+    isVisible: true,
+  }
+
+  public render() {
+    const {isVisible, overlayHeading, onDismissOverlay, status} = this.props
+
+    return (
+      <Overlay visible={isVisible}>
+        <Overlay.Container maxWidth={800}>
+          <Overlay.Heading
+            title={overlayHeading}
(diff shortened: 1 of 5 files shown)


ui/src/shared/components/ViewOverlay.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+interface Props {
+  onDismissOverlay: () => void
+  resource: DocumentCreate
+  overlayHeading: string
+  status: RemoteDataState
+  isVisible: boolean
+}
+
+export default class ViewOverlay extends PureComponent<Props> {
```
ischolten: do we need to add the errors decorator here?

ui/src/templates/components/TemplateCard.tsx
```diff
@@ -128,7 +128,7 @@ class TemplateCard extends PureComponent<Props & WithRouterProps> {
 
   private handleExport = () => {
     const {router, template, org} = this.props
-    router.push(`orgs/${org.id}/templates/${template.id}/export`)
+    router.push(`/orgs/${org.id}/templates/${template.id}/view`)
```
ischolten: should this be a part of the handleExport function? this is not dealing with exporting anymore? perhaps this function just requires a name change? it seems like this would be more of a handleNameClick concern.

ui/src/templates/components/TemplateViewOverlay.tsx
```diff
@@ -0,0 +1,87 @@
...
+    const {router, clearExportTemplate} = this.props
+
+    router.goBack()
+    clearExportTemplate()
+  }
+}
+
+const mstp = (state: AppState): StateProps => ({
+  exportTemplate: state.templates.exportTemplate.item,
+  status: state.templates.exportTemplate.status,
```
ischolten: could exportTemplate ever be undefined? should a lodash get be used here to ensure you dont get an error?

ui/src/shared/components/ViewOverlay.tsx
Palakp41: I updated it!

ui/src/templates/components/TemplateCard.tsx
Palakp41: Yup, forgot to update the name for handler

ui/src/templates/components/TemplateViewOverlay.tsx
Palakp41: It's being set as null in the state, so won't be ever undefined.

PR #13296, opened 2019-04-10
chore(tracing): Cleanup

jacobmarble:

Various tracing cleanups:
- Add parent spans where child spans were triggering root spans (eg Scheduler, Flux from/to)
- Remove a Tracer pointer (we use `opentracing.GlobalTracer`)
- Add some `tracing.LogError` love
- Inject and extract HTTP headers with library rather than copy-pasted code
- Log file/line from `tracing.LogError`
- Add `tracing.StartSpanFromContextWithOperationName`
- Generally replace opentracing usage with `kit/tracing` for consistency and better spans

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/query/stdlib/influxdata/influxdb/from.go
+++ b/query/stdlib/influxdata/influxdb/from.go
@@ -1,7 +1,9 @@
 package influxdb
 
 import (
+	"context"
 	"fmt"
+	"github.com/influxdata/influxdb/kit/tracing"
 
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/ast"
@@ -649,6 +651,9 @@ func (FromKeysRule) Rewrite(keysNode plan.Node) (plan.Node, bool, error) {
 }
 
 func createFromSource(prSpec plan.ProcedureSpec, dsid execute.DatasetID, a execute.Administration) (execute.Source, error) {
+	span, ctx := tracing.StartSpanFromContext(context.TODO())
+	defer span.Finish()
+
 	spec := prSpec.(*PhysicalFromProcedureSpec)
 	var w execute.Window
 	bounds := a.StreamContext().Bounds()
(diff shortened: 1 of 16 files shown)


query/stdlib/influxdata/influxdb/from.go
```diff
@@ -649,6 +651,9 @@ func (FromKeysRule) Rewrite(keysNode plan.Node) (plan.Node, bool, error) {
 }
 
 func createFromSource(prSpec plan.ProcedureSpec, dsid execute.DatasetID, a execute.Administration) (execute.Source, error) {
+	span, ctx := tracing.StartSpanFromContext(context.TODO())
```
docmerlin: Are you sure about this being a TODO? it seems that context.Background() would work better here.

query/stdlib/influxdata/influxdb/source.go
```diff
@@ -114,6 +115,9 @@ func (s *readFilterSource) run(ctx context.Context) error {
 }
 
 func createReadFilterSource(s plan.ProcedureSpec, id execute.DatasetID, a execute.Administration) (execute.Source, error) {
+	span, ctx := tracing.StartSpanFromContext(context.TODO())
```
docmerlin: Again, TODO -> Background

query/stdlib/influxdata/influxdb/to.go
```diff
@@ -439,6 +440,9 @@ func (s Stats) Update(o Stats) {
 }
 
 func writeTable(t *ToTransformation, tbl flux.Table) error {
+	span, ctx := tracing.StartSpanFromContext(context.TODO())
```
docmerlin: I'm not sure about this one.  Are you intending to eventually modify writeTable to accept a context?

query/stdlib/influxdata/influxdb/from.go
jacobmarble: Just a note for the query team. I figure they're neck-deep in refactors, so this is enough of a billboard.

query/stdlib/influxdata/influxdb/to.go
jacobmarble: That would be the idea, yes. Again, query team's decision. When you touch I/O, it's a good place for tracing and a good place for context.

PR #13303, opened 2019-04-10
Add prometheus counters for tracking write/query usage

desa:
Closes # (not sure if there's a direct issue for this)

This PR adds usage tracking for the write and query endpoints. In particular we track

1. The write request size (after gzip) and number of requests, per org
2. The query response size and number of query requests, per org

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/http/write_handler.go
+++ b/http/write_handler.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/influxdata/influxdb/http/metric"
 	"github.com/julienschmidt/httprouter"
 	"go.uber.org/zap"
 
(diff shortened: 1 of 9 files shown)


desa:
After talking with @mark-rushakoff I'm going to put some work into making this a bit more generic.

desa:
@mark-rushakoff @goller Made the interface a bit more generic, but the whole thing ended up being a whole lot messier than I would have preferred. In particular what was difficult was getting the size of the body of the request--specifically because the time when we handle the body of a request is deep inside the handler itself and [http.Handler](https://golang.org/pkg/net/http/#Handler) expects that the request body will not be modified.

Would be good to get a bit of feedback before putting in the effort of plumbing this work through idpe.

http/write_handler.go
```diff
@@ -61,13 +66,15 @@ const (
 
 // NewWriteHandler creates a new handler at /api/v2/write to receive line protocol.
 func NewWriteHandler(b *WriteBackend) *WriteHandler {
+
```
mark-rushakoff: Remove blank line.

PR #13315, opened 2019-04-11
feat(http): filter label response by organization

### LINKED ISSUE #13125: Labels GET should accept orgID as a query parameter
Similarly to how other resources have the query parameter orgID to return only the resources for the specified org, labels GET request should also take that orgID.

http://localhost:9999/api/v2/labels?orgID=00000012344234

jademcgough:
Closes #13125


label.go
```diff
@@ -113,8 +113,8 @@ type LabelUpdate struct {
 
 // LabelFilter represents a set of filters that restrict the returned results.
 type LabelFilter struct {
-	Name  string
-	OrgID *ID
+	Name           string
+	OrganizationID *ID
```
kelwang: I think we want to keep it as OrgID, been working on another ticket to replace all the OrganizationID to OrgID

http/label_service.go
```diff
@@ -334,23 +359,23 @@ func newGetLabelsHandler(b *LabelBackend) http.HandlerFunc {
...
 
-type getLabelsRequest struct {
-	filter platform.LabelMappingFilter
+type getLabelsMappingRequest struct {
+	filter influxdb.LabelMappingFilter
 }
 
-func decodeGetLabelsRequest(ctx context.Context, r *http.Request, rt platform.ResourceType) (*getLabelsRequest, error) {
-	req := &getLabelsRequest{}
+func decodeGetLabelMappingsRequest(ctx context.Context, r *http.Request, rt influxdb.ResourceType) (*getLabelsMappingRequest, error) {
```
kelwang: Should this be decode GetLabelsMappingRequest ?

kv/label.go
```diff
@@ -85,7 +85,8 @@ func (s *Service) findLabelByID(ctx context.Context, tx Tx, id influxdb.ID) (*in
 
 func filterLabelsFn(filter influxdb.LabelFilter) func(l *influxdb.Label) bool {
 	return func(label *influxdb.Label) bool {
-		return (filter.Name == "" || (filter.Name == label.Name))
+		return (filter.Name == "" || (filter.Name == label.Name) &&
+			filter.OrganizationID == nil || (*filter.OrganizationID == label.OrganizationID))
```
kelwang: If filter.Name != label.Name && filter.OrganizationID == nil, this might fail. To be safe add a check with filter.ID!=nil &&(*filter.OrganizationID == label.OrganizationID)

label.go
jademcgough: +1 couldn't remember what we'd decided for that

http/label_service.go
jademcgough: Other way around; I'm changing the struct name to match.

kv/label.go
jademcgough: Yep. Tests caught this too, was causing a segfault.

bednar:
Hi @jademcgough 

Why is added the `orgID` parameter to `POST` method?

```yaml
 /labels:
    post:
      tags:
        - Labels
      summary: Create a label
      parameters:
          - $ref: '#/components/parameters/TraceSpan'
          - in: query
            name: orgID
            description: return only labels that belong to the specified organization
            schema:
              type: string
      requestBody:
          description: label to create
          required: true
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/LabelCreateRequest"
      responses:
        '201':
          description: Added label
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/LabelResponse"
        default:
          description: unexpected error
          content:
            application/json:
              schema:
```

jademcgough:
@bednar looks like an error on my part, thanks

PR #13332, opened 2019-04-11
Update Telegraf Config overlay to be on a route

Palakp41:
Co-authored-by: Alirie Gray<alirie.gray@gmail.com>

Closes https://github.com/influxdata/applications-team-issues/issues/298

Add new route for viewing telegraf configs

  - [x] Rebased/mergeable
  - [ ] Tests pass


ui/src/telegrafs/components/CollectorRow.tsx
```diff
@@ -159,11 +167,22 @@ class CollectorRow extends PureComponent<Props> {
...
-    )
+    const {
+      collector,
+      router,
+      org,
+      onSetTelegrafConfigID,
+      onSetTelegrafConfigName,
+    } = this.props
+    onSetTelegrafConfigID(collector.id)
+    onSetTelegrafConfigName(collector.name)
```
ischolten: How do you feel about moving these to the telegraf config overlay? So it doesnt matter where its opened from or if you were on a previous page to have this information

ui/src/telegrafs/components/TelegrafConfigOverlay.tsx
```diff
@@ -69,18 +56,22 @@ export class TelegrafConfigOverlay extends PureComponent<Props> {
...
-      )
-    } catch (error) {
-      this.props.notify(getTelegrafConfigFailed())
-    }
+    const {
+      onDownloadTelegrafConfig,
+      telegrafConfigName,
+      telegrafConfigID,
+    } = this.props
+    onDownloadTelegrafConfig(telegrafConfigID, telegrafConfigName)
```
ischolten: could the config id from the url be used as the config id rather than getting one from redux?

ui/src/telegrafs/actions/index.ts
```diff
@@ -156,3 +159,15 @@ export const removeTelelgrafLabelsAsync = (
     dispatch(removeTelelgrafLabelFailed())
   }
 }
+
+export const downloadTelegrafConfig = (
+  telegrafConfigID: string,
+  telegrafConfigName: string
+) => async (dispatch): Promise<void> => {
+  try {
+    const config = await client.telegrafConfigs.getTOML(telegrafConfigID)
```
ischolten: since this config needs to be displayed by the overlay, could the same config be passed in so we dont have to make this request twice?

ui/src/telegrafs/components/TelegrafConfigOverlay.tsx
```diff
@@ -1,102 +1,109 @@
...
-      const config = await client.telegrafConfigs.getTOML(
-        this.props.telegrafConfigID
-      )
-      downloadTextFile(
-        config,
-        `${this.props.telegrafConfigName || 'config'}.toml`
+  private get overlay(): JSX.Element {
+    const {telegraf, status} = this.props
+
+    if (status === RemoteDataState.Done) {
```
ischolten: because of this check, we will never see the spinner below. How do you feel about removing this check so the spinner can do its thing?

ui/src/telegrafs/components/TelegrafConfigOverlay.tsx
```diff
@@ -1,102 +1,109 @@
...
+      return (
+        <Overlay visible={true}>
+          <Overlay.Container maxWidth={1200}>
+            <SpinnerContainer
```
ischolten: how do you feel about moving the spinner deeper within this tree? does the Overlay Heading rely on this status? If the TelegrafConfig component is the only thing relying on this spinner, then the rest of these contents like header and body could be rendered and within them, the spinner would spin.

ui/src/telegrafs/components/TelegrafConfigOverlay.tsx
Palakp41: The check for status is needed here because if the telegraf list is not done loading, it throws an error on telegraf name that is being displayed in the overlay heading.

ui/src/telegrafs/components/TelegrafConfigOverlay.tsx
Palakp41: I forgot I put a spinner in the overlay body, so I will remove this one.

PR #13345, opened 2019-04-12
adding new default template

russorat:
Closes #

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/templates/constants/defaultTemplates.ts
+++ b/ui/src/templates/constants/defaultTemplates.ts
@@ -2012,7 +2012,7 @@ export const gettingStartedWithFluxTemplate = () => ({
        "query": "buckets()\n  |> distinct(column: \"name\")"
       }
      },
-     "name": "buckets",
+     "name": "bucket",
      "selected": null
     },
     "id": "039d86a307713000",
--- a/ui/src/templates/constants/defaultTemplates.ts
+++ b/ui/src/templates/constants/defaultTemplates.ts
@@ -1330,7 +1330,7 @@ export const gettingStartedWithFluxTemplate = () => ({
     "attributes": {
      "name": "Name this Cell",
      "properties": {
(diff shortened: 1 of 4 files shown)


ui/src/templates/constants/defaultTemplates.ts
```diff
@@ -987,3 +987,1043 @@ export const systemTemplate = () => ({
...
+        "suffix": ""
+       }
+      },
+      "colors": [],
+      "geom": "line",
+      "legend": {},
+      "note": "# Uh oh, something has gone wrong!\n\nIf you're seeing this note, it means the queries running in this cell aren't returning any data. That might mean that your Telegraf instance hasn't sent any data to InfluxDB during the time range set in the Dashboard, or it might mean there is no data at all. Please install Telegraf or check your Telegraf configuration before continuing.\n\nYou can find detailed information about [setting up Telegraf in the InfluxDB 2.0 Documentation](https://v2.docs.influxdata.com/v2.0/collect-data/use-telegraf/).\n\nYou should configure the following plugins:\n\n- cpu\n- disk\n- network\n\nIf you need additional help, the best place to ask questions is on the [community site](https://community.influxdata.com/).",
+      "queries": [
+       {
+        "builderConfig": {
```
ischolten: this builder config should be set to the default "empty state" of a builderConfig if the edit mode is `advanced` so that the builder config isnt given hard coded values that the user doesnt have.

```      
                builderConfig: {	           
                  buckets: [],	                
                  tags: [{key: '_measurement', values: []}],	                
                  functions: [],	                  
                },	                
```

ischolten: also this should be the case for all the cells in edit mode

russorat: We should open an issue for this and fix the template export if that's the case. the goal is the exact same thing that is exported as a template from the app can be used here.

ischolten: will do

Issue #11969, opened 2019-02-19
Add metrics for level of TSM files


e-dard:
Currently we have a metric to track the number of TSM files in a storage engine:

https://github.com/influxdata/influxdb/blob/cbef811c63bf76ae0ce844fc1d5e95e39d53a871/tsdb/tsm1/metrics.go#L134-L155

These get updated when a `tsm1.FileStore` is opened and closed. They also get updated whenever a compaction completes.

https://github.com/influxdata/influxdb/blob/cbef811c63bf76ae0ce844fc1d5e95e39d53a871/tsdb/tsm1/file_store.go#L939-L950

I would like to add a new label to these metrics called `level`. This metric would have the following possible values: `level = {1, 2, 3, 4+}`. The label value would be set by iterating over the file set when they're replaced via a compaction and inspecting the filenames to determine the level. Simply put:

```
000000001.tsm <- level = "1"
000000002.tsm <- level = "2"
000000003.tsm <- level = "3"
000000004.tsm <- level = "4+"
...
...
...
000000092.tsm <- level = "4+"
```

tmgordeeva:
PR at https://github.com/influxdata/influxdb/pull/13353

PR #13353, opened 2019-04-12
feat(storage): add tsm level metrics

tmgordeeva:
Adds prometheus metrics recording compaction levels for TSM files.

Closes https://github.com/influxdata/influxdb/issues/11969

### DIFF
--- a/tsdb/tsm1/file_store.go
+++ b/tsdb/tsm1/file_store.go
@@ -303,42 +303,67 @@ type fileTracker struct {
 	metrics   *fileMetrics
 	labels    prometheus.Labels
 	diskBytes uint64
-	fileCount uint64
+	levels    uint64
 }
 
 func newFileTracker(metrics *fileMetrics, defaultLabels prometheus.Labels) *fileTracker {
 	return &fileTracker{metrics: metrics, labels: defaultLabels}
 }
 
+// Labels returns a copy of the default labels used by the tracker's metrics.
+// The returned map is safe for modification.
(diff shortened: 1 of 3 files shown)


tsdb/tsm1/file_store.go
```diff
@@ -631,13 +657,23 @@ func (f *FileStore) Open(ctx context.Context) error {
...
+		name := filepath.Base(res.r.Stats().Path)
+		_, seq, err := f.parseFileName(name)
+		if err != nil {
+			return err
+		}
+		if _, ok := counts[seq]; !ok {
+			counts[seq] = 1
+		} else {
+			counts[seq]++
+		}
```
zeebo: Neat trick: only the `++` branch is needed, since maps return the zero value for missing keys. https://play.golang.org/p/NKqO6UFZHn5

tsdb/tsm1/file_store.go
```diff
@@ -948,18 +984,27 @@ func (f *FileStore) replace(oldFiles, newFiles []string, updatedFn func(r []TSMF
...
+			size += uint64(ts.Size)
+		}
+		_, seq, err := f.parseFileName(file.Path())
+		if err != nil {
+			return err
+		}
+		if _, ok := sizes[seq]; !ok {
+			sizes[seq] = size
+		} else {
+			sizes[seq] += size
```
zeebo: Similar to the `++` stuff, you can just do `+=`.

PR #13382, opened 2019-04-12
fix(tsdb): Remove TSI stats file cache

benbjohnson:
Removes the `STATS` file generated during TSI compaction as it had
potential for becoming inconsistent with the index data. Instead,
stats are recalculated on start up and on each compaction on a
per-partition basis.

Computing stats for 10M series across 10K measurements takes
approximately 0.171s.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/tsdb/tsi1/file_set.go
+++ b/tsdb/tsi1/file_set.go
@@ -403,6 +403,39 @@ func (fs *FileSet) TagValueSeriesIDIterator(name, key, value []byte) (tsdb.Serie
 	return tsdb.NewSeriesIDSetIterator(ss), nil
 }
 
+// Stats computes aggregate measurement cardinality stats from the raw index data.
+func (fs *FileSet) Stats() MeasurementCardinalityStats {
+	stats := make(MeasurementCardinalityStats)
+	mitr := fs.MeasurementIterator()
+	if mitr == nil {
+		return stats
+	}
+
+	for {
+		// Iterate over each measurement and set cardinality.
+		mm := mitr.Next()
+		if mm == nil {
+			return stats
+		}
+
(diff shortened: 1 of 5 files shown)


tsdb/tsi1/file_set.go
```diff
@@ -403,6 +403,39 @@ func (fs *FileSet) TagValueSeriesIDIterator(name, key, value []byte) (tsdb.Serie
...
+			return stats
+		}
+
+		// Obtain all series for measurement.
+		sitr := fs.MeasurementSeriesIDIterator(mm.Name())
+		if sitr == nil {
+			continue
+		}
+
+		// All iterators should be series id set iterators except legacy 1.x data.
```
e-dard: Should it be possible to have legacy 1.x data?

benbjohnson: We could but it's going to take a long time to compute. This is specifically pre-roaring 1.x. Later 1.x data works ok.

benbjohnson:
@e-dard What level of end-to-end test are you thinking? We don't currently expose the stats beyond `storage.Engine.MeasurementCardinalityStats()` in OSS.

tsdb/tsi1/file_set.go
e-dard: No, I mean, there shouldn't be any legacy TSI data in a running 2.x instance.

benbjohnson: Ah, yes, that's correct. We could panic here instead but I figured just skipping it would be better.

benbjohnson:
@e-dard Yes, I can add a test on the in-house PR.

PR #13389, opened 2019-04-13
fix(query): return statistics even in encoder error case

mark-rushakoff:
The ProxyQueryServiceAsyncBridge was not returning statistics when there
was an encoder error. Because the encoder was just writing to an
io.Writer, it was possible that a remote disconnect could happen and
statistics could not be reported.

### DIFF
--- a/query/bridges_test.go
+++ b/query/bridges_test.go
@@ -1,0 +1,67 @@
+package query_test
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/csv"
+	"github.com/influxdata/flux/execute/executetest"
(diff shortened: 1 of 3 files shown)


query/bridges_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/influxdata/flux/execute/executetest"
+	"github.com/influxdata/influxdb/query"
+	"github.com/influxdata/influxdb/query/mock"
+)
+
+type failWriter struct {
+	Err error
+}
+
+// Write returns max(p, w.Take), w.Err,
```
wolffcm: _nit_ this comment must refer to a member `Take` that you later removed.

mark-rushakoff: Yes, good catch -- I'll fix this comment and merge.

PR #13397, opened 2019-04-15
fix(package): make rpm verify pass after installation

affo:
This change sets the ownership of directories on package generation with fpm instead that in post-install to make `rpm -V` pass.

For more details, see: influxdata/EAR#716

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/scripts/post-install.sh
+++ b/scripts/post-install.sh
@@ -24,14 +24,6 @@ function install_chkconfig {
     chkconfig --add influxdb
 }
 
-id influxdb &>/dev/null
-if [[ $? -ne 0 ]]; then
-    useradd --system -U -M influxdb -s /bin/false -d $DATA_DIR
-fi
-
-chown -R -L influxdb:influxdb $DATA_DIR
-chown -R -L influxdb:influxdb $LOG_DIR
-
 # Add defaults file, if it doesn't exist
 if [[ ! -f /etc/default/influxdb ]]; then
     touch /etc/default/influxdb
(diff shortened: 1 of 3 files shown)


affo:
I think this would need backport.

scripts/post-install.sh
```diff
@@ -53,6 +50,12 @@ if [[ -f /etc/redhat-release ]]; then
         install_chkconfig
     fi
 elif [[ -f /etc/debian_version ]]; then
+    # Ownership for RH-based platforms is set in build.py via the `rmp-attr` option.
+    # We perform ownership change only for Debian-based systems.
```
gunnaraasen: Do these options also need to be added to the Amazon Linux block at line 69?

gunnaraasen: Never mind, Amazon Linux and Amazon Linux are both CentOS-based with yum. Might be worth manually testing though.

e-dard:
@affo I don't know enough about this to review it. @mark-rushakoff might have a better insight.

scripts/post-install.sh
```diff
@@ -53,6 +50,12 @@ if [[ -f /etc/redhat-release ]]; then
...
+    # Moving these lines out of this if statement would make `rmp -V` fail after installation.
+    chown -R -L influxdb:influxdb $DATA_DIR
```
mark-rushakoff: The complemetary PRs https://github.com/influxdata/telegraf/pull/5716/ and https://github.com/influxdata/chronograf/pull/5146/ conditionally make the directories and `chmod 755` then. Does this need the same logic for the data and log dirs?

affo: I see your point. What I did here is to move the lines responsible for ownership change in order to make `rpm -V` pass and to leave the script's behavior the same as before when not using RHEL.  
It seems that the script didn't need those lines before.

Anyhow, I think I can add `chmod` lines without affecting the "correctness" of the script.

affo:
@mark-rushakoff I added the `chmod` lines and moved user creation to pre-install phase.

I did because of an issue with `rpm-attr` and non-existing user at installation time (see https://github.com/influxdata/EAR/issues/716#issuecomment-487758645).

PR #13406, opened 2019-04-15
Add route to new member overlay

Palakp41:
Closes https://github.com/influxdata/applications-team-issues/issues/292

The users in new members overlay are now stored in redux state.
Updated the route to /new and include orgID

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/members/components/AddMembersOverlay.tsx
+++ b/ui/src/members/components/AddMembersOverlay.tsx
@@ -1,20 +1,31 @@
 // Libraries
 import React, {PureComponent, ChangeEvent} from 'react'
+import {connect} from 'react-redux'
+import {withRouter, WithRouterProps} from 'react-router'
 
 // Components
 import {Overlay} from 'src/clockface'
 import AddMembersForm from 'src/members/components/AddMembersForm'
 import FilterList from 'src/shared/components/Filter'
+import {SpinnerContainer, TechnoSpinner} from '@influxdata/clockface'
+
+// Actions
+import {getUsers, addNewMember} from 'src/members/actions'
 
 // Types
-import {UsersMap} from 'src/members/components/Members'
+import {UsersMap} from 'src/members/reducers'
 import {User} from '@influxdata/influx'
-import {Member} from 'src/types'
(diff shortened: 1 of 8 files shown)


ui/src/members/components/AddMembersOverlay.tsx
```diff
@@ -35,36 +47,46 @@ export default class AddMembersOverlay extends PureComponent<Props, State> {
...
+                )}
+              </FilterList>
+            </SpinnerContainer>
+          </Overlay.Body>
+        </Overlay.Container>
+      </Overlay>
     )
   }
+  public componentDidMount() {
+    this.props.getUsers()
```
ebb-tide: Hrmm.. Could we use the GetResources component here instead since it handles the spinner container and everything. That way we have one component that handles fetching data?

ui/src/members/components/Members.tsx
```diff
@@ -101,33 +81,19 @@ class Members extends PureComponent<Props, State> {
...
-    this.setState({overlayState: OverlayState.Open})
-  }
-
-  private handleCloseModal = (): void => {
-    this.setState({overlayState: OverlayState.Closed})
-  }
-
-  private async getUsers() {
-    const {members} = this.props
-    const apiUsers = await client.users.getAll()
```
ebb-tide: yay! these client calls should not be happening here.

ui/src/members/containers/MembersIndex.tsx
```diff
@@ -54,6 +54,7 @@ class MembersIndex extends Component<Props> {
                 >
                   <GetResources resource={ResourceTypes.Members}>
                     <Members />
+                    {this.props.children}
```
ebb-tide: Here overlay is rendered as a child of GetResources

ui/src/members/components/AddMembersOverlay.tsx
ebb-tide: Hrmmm actually, I think you are already wrapping this component in a GetResources ? in `ui/src/members/containers/MembersIndex.tsx` below

PR #13409, opened 2019-04-15
feat(storage): Implement storage schema RPC de/serializer, merge, APIs

### LINKED ISSUE #13251: Implement storage schema RPC APIs

### LINKED ISSUE #13271: Implement storage schema RPC serializer / deserializer and merge
Dependent on stubs from #13241

This issue tracks the required implementations of types responsible for serializing, deserializing and merging `StringIterator` instances over gRPC.

See #13251 for additional detail on the definition of `StringIterator` and the `reads.ResultSet` section, which describes the behavior of the following types.

## High level tasks 

- [x] Implement a `StringIterator` gRPC writer / serializer

This type is analagous to the `ResponseWriter` detailed in the `reads.ResultSet` section of #13251, and is reponsible for serializing the `StringIterator` to a gRPC stream

------

- [x] Implement a `StringIterator` gRPC reader / deserializer

This type is analagous to the `ResultSetStreamReader` detailed in the `reads.ResultSet` section of #13251, and is reponsible for deserializing the `StringIterator` from a gRPC stream. This type must implement a `StringIterator` interface or similar.

------

- [x] Implement a `StringIterator` merger

This type is analagous to the `mergedResultSet` detailed in the `reads.ResultSet` section of #13251, and is reponsible for merging and deduplicating multiple `StringIterator` instances into a single `StringIterator`.

jacobmarble:
Closes #13271
Closes #13251
Helps https://github.com/influxdata/idpe/issues/3114

Adds schema support to storage APIs


storage/reads/datatypes/storage_common.proto
```diff
@@ -24,6 +24,12 @@ service Storage {
 
   rpc Hints (google.protobuf.Empty) returns (HintsResponse);
 
+  // TagKeys performs a read operation for tag keys
+  rpc TagKeys (TagsRequest) returns (stream TagsResponse);
```
stuartcarnie: need distinct `TagKeysRequest` and `TagValuesRequest` arguments to these APIs, as they have different parameters

storage/reads/datatypes/storage_common.proto
```diff
@@ -24,6 +24,12 @@ service Storage {
...
+
+  // TagValues performs a read operation for tag values
+  rpc TagValues (TagsRequest) returns (stream TagsResponse);
```
stuartcarnie: Add `TagValuesRequest` that has the same parameters as `TagKeysRequest` with the addition of `string tag_key` to specify which tag key the values should be produced for

storage/reads/datatypes/storage_common.proto
```diff
@@ -208,6 +214,18 @@ message TimestampRange {
...
 
+// TagsRequest is the request message for the Storage.TagKeys and Storage.TagValues
+message TagsRequest {
+  google.protobuf.Any tags_source = 1 [(gogoproto.customname) = "TagsSource"];
+  TimestampRange range = 2 [(gogoproto.nullable) = false];
+  Predicate predicate = 3;
+}
+
+// Response message for Storage.TagKeys and Storage.TagValues.
+message TagsResponse {
```
stuartcarnie: This should be renamed to something more generic, like `StringValuesResponse` given it represents the generic `StringIterator` and its associated merge operation.

storage/reads/merge.go
```diff
@@ -114,10 +115,47 @@ func (h *resultSetHeap) Push(x interface{}) {
...
+	}
+}
+
+func (mr *MergedStringIterator) Next() bool {
+	for len(mr.iterators) > 0 {
+		iterator := mr.iterators[0]
+
+		for iterator.Next() {
+			mr.nextValue = iterator.Value()
+			if _, found := mr.uniqueValues[mr.nextValue]; !found {
```
stuartcarnie: A `StringIterator` returned by the engine's `TagValues` or `TagKeys` APIs will produce sorted results, so you can rely on that to implement a k-way merge sort. 

Given there is still a lot to do, we can add that as a TODO in this code and come back to it later this week with a separate PR.

jacobmarble: I tried to find that written down somewhere, because I thought you mentioned it verbally. Anyway, #13440

PR #13414, opened 2019-04-15
feat(ui/variables): add name validation to overlay

### LINKED ISSUE #13297: Variable name should be unique
Variables can be created with the same name using the create overlay or when importing.

- [x] Verify variable name is unique in Variable Overlay
> -  Import variable name should error if name already exists

This could be enforced in api side. However, we don't currently do this for labels, which also requires name uniqueness in the ui.

OfTheDelmer:
Closes #13297 

Add a variables uniqueness validation to the variable creation form.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/dataExplorer/components/SaveAsVariable.tsx
+++ b/ui/src/dataExplorer/components/SaveAsVariable.tsx
@@ -8,10 +8,12 @@ import VariableForm from 'src/variables/components/VariableForm'
 
 // Utils
 import {createVariable} from 'src/variables/actions'
+import {extractVariablesList} from 'src/variables/selectors'
 
 // Types
 import {AppState} from 'src/types'
 import {getActiveQuery} from 'src/timeMachine/selectors'
+import {IVariable as Variable} from '@influxdata/influx'
 
 interface OwnProps {
   onHideOverlay: () => void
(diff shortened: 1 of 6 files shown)


ui/src/dataExplorer/components/SaveAsVariable.tsx
```diff
@@ -23,28 +25,38 @@ interface DispatchProps {
...
         initialScript={initialScript}
+        variables={variables}
       />
     )
   }
 }
 
 const mstp = (state: AppState): StateProps => {
   const activeQuery = getActiveQuery(state)
+  const variables = extractVariablesList(state)
```
OfTheDelmer: Note: variables state is loaded by the `DataExplorer`, which is a sibling. If we want to be more explicit we could lift that `GetResources` to the `DataExplorerPage`

ui/src/variables/components/VariableForm.tsx
```diff
@@ -32,31 +30,36 @@ export default class VariableForm extends PureComponent<Props, State> {
...
-                    value={name}
-                    onChange={this.handleChangeInput}
-                    status={nameInputStatus}
-                  />
-                </Form.Element>
+                <Form.ValidationElement
+                  label="Name"
+                  value={name}
+                  required={true}
+                  validationFunc={this.handleNameValidation}
```
AlirieGray: nice

PR #13418, opened 2019-04-15
Create an analytical storage middleware layer

### LINKED ISSUE #12728: Convert the current PointLogWriter to be a TaskService middleware
The current implementation of the Log writer is built as a stand alone log/run Writer. To simplify the way we do these function calls we can make a middleware layer in top of the kv task service that writes to the analytical store when the run in finalized. This will allow writer to only write one time and pull the data in a simple and precise way.

The middleware here will also act as a log/run reader for data that isn't in the kv service.

lyondhill:
Closes #12728

### DIFF
--- a/task.go
+++ b/task.go
@@ -71,6 +71,11 @@ func (r *Run) ScheduledForTime() (time.Time, error) {
 	return time.Parse(time.RFC3339, r.ScheduledFor)
 }
 
+// StartedAtTime gives the time.Time that the run is scheduled for.
+func (r *Run) StartedAtTime() (time.Time, error) {
+	return time.Parse(time.RFC3339Nano, r.StartedAt)
+}
+
 // RequestedAtTime gives the time.Time that the run was requested.
 func (r *Run) RequestedAtTime() (time.Time, error) {
 	return time.Parse(time.RFC3339, r.RequestedAt)
--- a/task.go
+++ b/task.go
@@ -71,7 +71,7 @@ func (r *Run) ScheduledForTime() (time.Time, error) {
 	return time.Parse(time.RFC3339, r.ScheduledFor)
 }
 
-// StartedAtTime gives the time.Time that the run is scheduled for.
+// StartedAtTime gives the time.Time that the run was started.
 func (r *Run) StartedAtTime() (time.Time, error) {
 	return time.Parse(time.RFC3339Nano, r.StartedAt)
 }
--- a/task/backend/analytical_storage_test.go
+++ b/task/backend/analytical_storage_test.go
@@ -1,0 +1,132 @@
+package backend_test
+
+import (
+	"context"
+	// "fmt"
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/influxdata/flux/control"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/influxdb"
+	icontext "github.com/influxdata/influxdb/context"
+	"github.com/influxdata/influxdb/inmem"
+	"github.com/influxdata/influxdb/kv"
+	"github.com/influxdata/influxdb/query"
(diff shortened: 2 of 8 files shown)


task.go
```diff
@@ -71,6 +71,11 @@ func (r *Run) ScheduledForTime() (time.Time, error) {
 	return time.Parse(time.RFC3339, r.ScheduledFor)
 }
 
+// StartedAtTime gives the time.Time that the run is scheduled for.
```
mark-rushakoff: ... the time that the run started?

ui/package-lock.json
```diff
@@ -6176,21 +6176,20 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": false,
+          "resolved": "",
```
mark-rushakoff: Can you undo your changes to this file?

task/servicetest/servicetest.go
```diff
@@ -1188,11 +1197,101 @@ func testRetryAcrossStorage(t *testing.T, sys *System) {
...
+		t.Fatal(err)
+	}
+
+	logs, _, err := sys.TaskService.FindLogs(sys.Ctx, influxdb.LogFilter{Task: task.ID})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(logs) != 7 {
+		for _, log := range logs {
+			fmt.Printf("log: %+v\n", log)
```
mark-rushakoff: This should be `t.Logf` so it displays correctly with the `t.Fatalf` below.

task/servicetest/servicetest.go
```diff
@@ -1188,11 +1197,101 @@ func testRetryAcrossStorage(t *testing.T, sys *System) {
...
+	if err := sys.TaskControlService.UpdateRunState(sys.Ctx, task.ID, rc1.Created.RunID, startedAt, backend.RunStarted); err != nil {
+		t.Fatal(err)
+	}
+
+	// Mark the second run finished.
+	if err := sys.TaskControlService.UpdateRunState(sys.Ctx, task.ID, rc1.Created.RunID, startedAt.Add(time.Second), backend.RunSuccess); err != nil {
+		t.Fatal(err)
+	}
+
+	sys.TaskControlService.AddRunLog(sys.Ctx, task.ID, rc0.Created.RunID, time.Now(), "0-0")
```
mark-rushakoff: Can you add more comments explaining the intent in this new test?

I think something is going on between rc0 and rc1 but it's not really obvious to me.

task/backend/analytical_storage_test.go
```diff
@@ -0,0 +1,132 @@
+package backend_test
+
+import (
+	"context"
+	// "fmt"
```
mark-rushakoff: Remove comment.

task/backend/analytical_storage_test.go
```diff
@@ -0,0 +1,132 @@
...
+	if err := os.RemoveAll(lrw.rootDir); err != nil {
+		t.Error(err)
+	}
+}
+
+func newAnalyticalBackend(t *testing.T, orgSvc influxdb.OrganizationService, bucketSvc influxdb.BucketService) *analyticalBackend {
+	// Mostly copied out of cmd/influxd/main.go.
+	logger := zaptest.NewLogger(t)
+
+	rootDir, err := ioutil.TempDir("", "task-logreaderwriter-")
```
mark-rushakoff: Does rootDir get removed somewhere?

ui/package-lock.json
lyondhill: yah. my bad :(

task/backend/analytical_storage_test.go
lyondhill: in the close method I remove it. But it is currently possible for it to not be removed if we fail to setup the engine. I will fix that.

PR #13437, opened 2019-04-16
feat(http): Default http transport uses the actual default values

fntlnz:
Signed-off-by: Lorenzo Fontana <lo@linux.com>
Co-authored-by: Leonardo Di Donato <leodidonato@gmail.com>

  - [x] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/http/transport.go
+++ b/http/transport.go
@@ -1,0 +1,48 @@
+// Shared transports for all clients to prevent leaking connections
+package http
+
+import (
+	"crypto/tls"
+	"net"
+	"net/http"
+	"time"
+)
+
+// defaultTransport is the default implementation of Transport and is
+// used by traceClient. It establishes network connections as needed
+// and caches them for reuse by subsequent calls. It uses HTTP proxies
+// as directed by the $HTTP_PROXY and $NO_PROXY (or $http_proxy and
+// $no_proxy) environment variables.
+//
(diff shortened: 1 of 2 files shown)


http/transport.go
```diff
@@ -0,0 +1,48 @@
...
+		DualStack: true,
+	}).DialContext,
+	MaxIdleConns:          100,
+	IdleConnTimeout:       90 * time.Second,
+	TLSHandshakeTimeout:   10 * time.Second,
+	ExpectContinueTimeout: 1 * time.Second,
+}
+
+// skipVerifyTransport is the default implementation of Transport and is
+// used by traceClient (newClient with insecure set to true). It establishes network connections as needed
```
goller: Would you mention why? something like:

```skipVerifyTransport accepts self-signed certificates.

File: http/transport.go [hunk 2]
```diff
@@ -0,0 +1,48 @@
...
+package http
+
+import (
+	"crypto/tls"
+	"net"
+	"net/http"
+	"time"
+)
+
+// defaultTransport is the default implementation of Transport and is
```
goller: Would you mention that this is the same as http.DefaultTransport (and if there are values that are different than in the standard library, please note 'em)

http/transport.go
fntlnz: @goller your comment is empty.

PR #13441, opened 2019-04-16
Fix the sort order for aggregates so that they are sorted by tag and then time

jsternberg:
The reduce iterators would read in the points for a window, which
matched the grouping of the outermost query, and then it would sort them
by the time before emitting the points.

When there were multiple series, this would sometimes cause a conflict
because it would change the sorting of the inner query output when
selectors were used within a subquery. Then, these emitted points would
be output in the wrong order and they wouldn't join correctly when
multiple cursors were used.

This fixes it so the sorting happens per series grouping rather than on
all of the points together so they retain their tag order which is the
correct sorting method.

### DIFF
--- a/query/iterator.gen.go
+++ b/query/iterator.gen.go
@@ -1115,6 +1115,8 @@ func (itr *floatReduceFloatIterator) reduce() ([]FloatPoint, error) {
 	}
 
 	// Reverse sort points by name & tag if our output is supposed to be ordered.
+	// This is because the final array needs to have the first
+	// point at the end since they are popped off of the end.
 	keys := make([]string, 0, len(m))
 	for k := range m {
 		keys = append(keys, k)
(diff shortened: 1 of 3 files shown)


jlapacik:
I'm not sure I completely follow. So reduce iterators (which correspond to selectors as well as aggregates?) operate on a list of points? And multiple series could be represented in this list of points? The iterator assumes the points for each series are clustered together, but it used to sort its entire output by time, regardless of how many series were present in the points?

jsternberg:
That is correct. A sorted input is supposed to be by series and then by time, but it was sorting by time first so it mixed up the ordering.

jsternberg:
I think this may need to be reverted. Basically, while it fixes aggregates, it causes raw queries to return the wrong sorting order because they end up with the points out of time order since it sorts by the tag in the inner query. When it is then read by the outer query which doesn't have the same grouping, this causes it to return points in the wrong order.

PR #13459, opened 2019-04-17
Fill resets the previous value when a new series or window is encountered

aanthony1243:
- [X] Rebased/mergeable
  - [X] Tests pass


query/cursor.go
```diff
@@ -233,6 +233,12 @@ func newScannerCursor(s IteratorScanner, fields []*influxql.Field, opt IteratorO
 
 func (s *scannerCursor) scan(m map[string]interface{}) (int64, string, Tags) {
 	ts, name, tags := s.scanner.Peek()
+	// if a new series, clear the map of previous values
+	if name != s.series.Name || tags.ID() != s.series.Tags.ID() {
+		for k := range m {
```
jsternberg: Can you encapsulate this logic with a `clear(m)` method on the `scannerCursorBase`?

PR #13466, opened 2019-04-17
feat(ui/variables/map): add map type variables

### LINKED ISSUE #12832: Allow user to create map type variables
When creating a variable in the "manage variables" page, the user should be shown a dropdown allowing them to select which type of variable they would like to create.

Currently, only query type variables can be created from this page.

The user should be able to create a map type variable.

Alongside this issue, validation should also be added for map type variables: https://github.com/influxdata/influxdb/issues/12008

Screen Shot 2019-03-21 at 2 00 48 PM

OfTheDelmer:
Closes #12832

Add map type variables to create and update overlay as part of 12832. Ports over the tests from chrongoraf 1.7

Kapture 2019-04-17 at 17 15 57

  - [ ] Rebased/mergeable
  - [ ] Tests pass


ui/src/variables/components/MapVariableBuilder.tsx
```diff
@@ -0,0 +1,145 @@
+import React, {PureComponent} from 'react'
+import _ from 'lodash'
+import {connect} from 'react-redux'
+
+// Component
+// import TemplatePreviewList from 'src/tempVars/components/TemplatePreviewList'
```
mavarius: remove this if not used

ui/src/variables/components/MapVariableBuilder.tsx
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    return Object.entries(values).map(([key, value]) => ({
+      key,
+      value,
+    }))
+  }
+
+  private get pluralizer(): string {
+    return Object.keys(this.props.values).length === 1 ? '' : 's'
```
mavarius: This should really be a utility, there are many places in the application that this is repeated.

ui/src/variables/components/MapVariableBuilder.tsx
OfTheDelmer: indeed

PR #13491, opened 2019-04-17
fix(storage): Assume sorted StringIterators, and retain sorted order

### LINKED ISSUE #13440: MergedStringIterator should assume sorted input, output
A `StringIterator` returned by the engine's `TagValues` or `TagKeys` APIs will produce sorted results, so you can rely on that to implement a k-way merge sort. 

_Originally posted by @stuartcarnie in https://github.com/influxdata/influxdb/pull/13409_

jacobmarble:
Assume that `StringIterator` always returns values in order, then retain that order on merge. Cost is `log(n)` per call to `MergedStringIterator.Next()` where `n` is the quantity of non-empty `StringIterator`s.

Closes #13440

### DIFF
--- a/storage/reads/merge.go
+++ b/storage/reads/merge.go
@@ -125,7 +125,7 @@ func (h *resultSetHeap) Pop() interface{} {
 // It sorts and deduplicates adjacent values, so the output is sorted iff all inputs are sorted.
 // If all inputs are not sorted, then output order and deduplication are undefined and unpleasant.
 type MergedStringIterator struct {
-	heap      *stringIteratorHeap
+	heap      stringIteratorHeap
 	nextValue string
 }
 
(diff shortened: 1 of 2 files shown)


storage/reads/merge.go
```diff
@@ -122,42 +122,81 @@ func (h *resultSetHeap) Pop() interface{} {
...
 }
 
+// MergedStringIterator merges multiple storage.StringIterators into one.
+// It sorts and deduplicates adjacent values, so the output is sorted iff all inputs are sorted.
+// If all inputs are not sorted, then output order and deduplication are undefined and unpleasant.
 type MergedStringIterator struct {
-	iterators    []storage.StringIterator
-	uniqueValues map[string]struct{}
-	nextValue    string
+	heap      *stringIteratorHeap
```
stuartcarnie: You can embed this to avoid an additional allocation

```suggestion
	heap      stringIteratorHeap
```

storage/reads/merge.go
```diff
@@ -122,42 +122,81 @@ func (h *resultSetHeap) Pop() interface{} {
...
 }
 
-func (mr *MergedStringIterator) Next() bool {
-	// TODO assume that each iterator is sorted, and iterate in sorted order
-	// https://github.com/influxdata/influxdb/issues/13440
-	for len(mr.iterators) > 0 {
-		iterator := mr.iterators[0]
+func (msi *MergedStringIterator) Next() bool {
+	for msi.heap.Len() > 0 {
+		iterator := heap.Pop(msi.heap).(storage.StringIterator)
```
stuartcarnie: Not a required change, however an alternative approach, using `heap.Fix`. 

A property of the min-heap is that element zero is always the minimum, so you can avoid two heap orderings per iteration (`heap.Pop`, `heap.Push`):

```go
for msi.heap.Len() > 0 {
	iterator := msi.heap.iterators[0]

	haveNext := false
	if proposedNextValue := iterator.Value(); proposedNextValue != msi.nextValue { // Skip dupes.
		msi.nextValue = proposedNextValue
		haveNext = true
	}

	if iterator.Next() {
		heap.Fix(msi.heap, 0)
	} else {
		heap.Pop(msi.heap)
	}

	if haveNext {
		return true
	}
}

return false
```

Issue #12851, opened 2019-03-22
Meta queries should have their own RPC call


jlapacik:
Meta queries in Flux come in 2 flavors:

* Queries that fetch tag keys
* Queries that fetch tag values

Q: Are there more? SHOW SERIES?

Storage has efficient APIs for fetching both of these and which don't require touching TSM files like normal queries do. Unfortunately right now, unbounded flux queries (queries without a range) are not allowed, so all meta queries specify a default range of 30 days, and this guarantees that we touch TSM files which negatively impacts the performance of these meta queries.

These queries should instead be answered using two specialized RPC calls that use the storage APIs for fetching tag keys and values. These RPC calls would correspond to two physical index read operations at the query planning level:

* ReadTagKeys
* ReadTagValues

These types of queries correspond to the following flux patterns respectively:

```
from |> filter |> keys |> keep |> unique     // tag keys
from |> filter |> keep |> unique             // tag values
```

Note that a call to `from` as well as a terminating `keep |> unique` is required. Any unbounded query not satisfying this pattern would fail planning.

I'm not entirely if we need 2 RPC calls or just one, as the data access pattern should be mostly the same between the two types of queries (when a filter is specified).

It would also be nice to incorporate this into the stack based planning method defined in #10758. A `unique` call at the top of the stack would map to the meta query RPC call. I don't know if this would work though, as `from |> range |> filter |> keep |> unique` is a meta query that must touch TSM files whereas `from |> filter |> keep |> unique` does not. Also a plain `from |> filter` is not valid.

wolffcm:
I like this approach a lot.  I think it will work with the stack-based approach from #10758.

If there is no `range`, then the logical planner would still produce `from |> filter ...` and the desired patterns metaqueries in the physical planner could match what's described here.  So basically the physical planner would have a set of rules:
- `from |> range |> ...` --> produce a normal read gRPC call
- `from |> filter |> ...` --> produce a meta-query gRPC call if possible

If someone tried to roll their own metaquery that included a call to `range` it might be slow.  But most users (and the UI) will use the `influxdata/influxdb/v1` wrappers, which we can optimize to use the patterns above.

jlapacik:
All queries must be time bounded. That is a product requirement. Therefore meta queries will need to include a range. So the pattern for meta queries then becomes:

```
from |> range |> keys |> keep |> unique  // tag keys
from |> range |> keep |> unique          // tag values
```

jsternberg:
[...]

PR #13500, opened 2019-04-18
feat(storage): add tag keys reader to the storage reader interface

jsternberg:
Related to #12851.

### DIFF
--- a/storage/reads/reader.go
+++ b/storage/reads/reader.go
@@ -13,6 +13,7 @@ import (
 	"github.com/influxdata/flux/values"
 	"github.com/influxdata/influxdb/models"
 	"github.com/influxdata/influxdb/query/stdlib/influxdata/influxdb"
+	"github.com/influxdata/influxdb/storage"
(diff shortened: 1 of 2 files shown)


storage/reads/reader.go
```diff
@@ -694,3 +715,70 @@ func groupKeyForGroup(kv [][]byte, readSpec *influxdb.ReadSpec, bnds execute.Bou
...
+		}
+	}
+
+	// Construct the table and add to the reference count
+	// so we can free the table later.
+	tbl, err := builder.Table()
+	if err != nil {
+		return err
+	}
+	tbl.RefCount(1)
```
jlapacik: Where exactly are table reference counts used?

jsternberg: It's used by the table implementation for releasing the memory. We're not very good about it and we should probably update this API to be `Retain` and `Release` and for the default reference count to be 1, but it seemed like the default right now was 0.

jsternberg:
I'm not sure what happened, but this code or code similar to it has already been merged. I'm going to close this PR.

PR #13502, opened 2019-04-18
fix(query): add panic stack to logs for query logging recovery

nathanielc:


query/logging.go
```diff
@@ -26,9 +30,12 @@ func (s *LoggingProxyQueryService) Query(ctx context.Context, w io.Writer, req *
 
 	var n int64
 	defer func() {
-		r := recover()
-		if r != nil {
+		if r := recover(); r != nil {
 			err = fmt.Errorf("panic: %v", r)
+			if entry := s.Logger.Check(zapcore.InfoLevel, "QueryLogging panic"); entry != nil {
+				entry.Stack = string(debug.Stack())
```
mark-rushakoff: Should this instead be `entry.Write(zap.Stack("<some key name>"))`?

Looks like zap does a bit of optimization in the allocations for the buffer for the stack.

nathanielc: I expected only a single Write call to be valid, but I am not sure. The docs don't really state either way. 

Also I am not sure what the difference between `zap.Stack` and `Entry.Stack` are. This code came from https://github.com/influxdata/flux/blob/master/execute/dispatcher.go#L74 as part of https://github.com/influxdata/flux/commit/6e20ca2ed2045476f8150f0a94e67b956c3dba6d 

@jsternberg Can you provide some clarity here?

mark-rushakoff: I was looking at the example for zap.Logger.Check: https://godoc.org/go.uber.org/zap#Logger.Check

kit/prom/registry_test.go has a test to check logging output if you're so inclined to add a test to check for the stack being included.

jsternberg: I think stack is treated in a special way if it is set on the entry. Let me try and find that.

jsternberg: Here's where the support was added into zap-logfmt: jsternberg/zap-logfmt@677997e.

Here's where I modeled it from: https://github.com/uber-go/zap/blob/master/zapcore/console_encoder.go#L114

There's not much of a difference for the logfmt encoder because it tries to use logfmt for everything so the stacktrace output ends up being the same as if it were added to a key. But, the console encoder will print out the log message, a newline, and then the stacktrace below in a readable way that is easier to read so I tried to keep with making that possible.

Otherwise, the console encoder would include it in the json context that it prints instead of as a newline.

I think otherwise the `Write` call would be correct. Stacktraces are just special.

jsternberg: Unrelated, but `Write` I think can only be called once. It can be called with multiple arguments though so you could pass `zap.Stack` into it along with the error. I think setting the stack on the entry though is more correct IMO.

nathanielc: Sounds good and its working so I vote to leave it as is.

PR #13503, opened 2019-04-18
feat(ui/variables/csv): add csv variable

### LINKED ISSUE #12831: Allow user to create constant type variable
When creating a variable in the "manage variables" page, the user should be shown a dropdown allowing them to select which type of variable they would like to create.

Currently, only query type variables can be created from this page.

The user should be able to create a constant type variable.

Validation should also be added for constant type variables: https://github.com/influxdata/influxdb/issues/12008

Screen Shot 2019-03-21 at 2 00 48 PM

OfTheDelmer:
Closes #12831 

Does some clean up for map var notifications and adds the constant variable type to the create/update overlays.

Kapture 2019-04-18 at 11 47 29

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/variables/components/CSVVariableBuilder.tsx
+++ b/ui/src/variables/components/CSVVariableBuilder.tsx
@@ -82,7 +82,7 @@ export default class CSVTemplateBuilder extends PureComponent<Props, State> {
     const {onChange} = this.props
     const {csv} = this.state
 
-    const update = this.getValuesFromString(csv)
+    const update = this.getUniqueValuesFromCSV(csv)
 
     onChange(update)
   }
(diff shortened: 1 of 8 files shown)


ui/src/variables/components/CSVVariableBuilder.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  private handleChange = (csv: string): void => {
+    this.setState({csv})
+  }
+
+  private getValuesFromString(csv: string) {
+    const parsedTVS = Papa.parse(csv)
+    const templateValuesData: string[][] = _.get(parsedTVS, 'data', [[]])
+
+    const valueSet = new Set()
```
ischolten: is this to ensure that values are unique? if so should the name of this function indicate that? something like `getUniqueValuesFromString` or just `getUniqueValues`?

PR #13505, opened 2019-04-18
fix(query): fix misc tests with new master Flux

nathanielc:


query/influxql/spectests/show_tag_values_multiple_measurements.go
```diff
@@ -8,7 +8,7 @@ func init() {
 
 from(bucketID: "")
 	|> range(start: -1h)
-	|> filter(fn: (r) => r._measurement == "cpu" or r._measurement == "mem" or r._measurement == "gpu")
+	|> filter(fn: (r) => r._measurement == "cpu" or (r._measurement == "mem" or r._measurement == "gpu"))
```
jsternberg: Is this a bug with the formatter?

nathanielc: I thought it might be, but then I added a test case for it and it passed without parens, so I am not sure what was going on there. Maybe the way the transpiler builds the AST vs the way the parser does?

jsternberg: That must be it. We should fix that, but that's unrelated to this and doesn't matter in the grand scheme of things.

PR #13509, opened 2019-04-18
feat(http): pretty stack trace for the http panic recovery handler

jsternberg:
When printing out a stacktrace with a logged error message, it is better
to set the `Stack` property on the entry than to include it as a string
within the context fields. It is also better than using `zap.Stack()`
too.

This is because certain encoders will perform a pretty print of the
stacktrace. The default zap-logfmt will read this property and include
it as a field so this code is identical when using that encoder, but the
console encoder, which is commonly used in local development because it
is automatically used with a TTY available, will print out a newline and
pretty print the stacktrace if it is included on the entry itself.

### DIFF
--- a/http/router.go
+++ b/http/router.go
@@ -11,6 +11,7 @@ import (
 	influxlogger "github.com/influxdata/influxdb/logger"
 	"github.com/julienschmidt/httprouter"
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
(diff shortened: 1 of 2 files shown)


jsternberg:
@goller this is related to a conversation I had with @mark-rushakoff earlier. I found this completely separately from that conversation while I was working on my own tasks. I encountered a panic and I made this change to make it easier for me to read the stacktrace in local development.

As mentioned in that comment, this change should not effect how the message is displayed in production because production will be using the logfmt encoder which writes it as a string field.

PR #13512, opened 2019-04-18
feat(query): add push down rule for filters to read range procedure

jsternberg:
The read range procedure that uses the new rpc endpoints will now accept
filters using the same rules as the previous physical from procedure
spec.

Related to #12851.

### DIFF
--- a/query/stdlib/influxdata/influxdb/rules.go
+++ b/query/stdlib/influxdata/influxdb/rules.go
@@ -2,12 +2,14 @@ package influxdb
 
 import (
 	"github.com/influxdata/flux/plan"
+	"github.com/influxdata/flux/semantic"
 	"github.com/influxdata/flux/stdlib/universe"
 )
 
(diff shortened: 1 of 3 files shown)


query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -35,3 +37,78 @@ func (rule PushDownRangeRule) Rewrite(node plan.Node) (plan.Node, bool, error) {
 		Bounds:   rangeSpec.Bounds,
 	}), true, nil
 }
+
+// MergeFromFilterRule is a rule that pushes filters into from procedures to be evaluated in the storage layer.
```
jlapacik: This comment can be removed now.

jsternberg: It seems I forgot to rename it. Usually the refactors take care of that, but I think I manually changed the name.

PR #13513, opened 2019-04-18
feat(query): implement the read tag keys rpc call in the query engine

jsternberg:
If a pattern is seen that matches reading the tag keys, it will be
replaced with a direct RPC call to read the tag keys which should be
better optimized than reading from the storage engine tsm1 files.

Related to #12851.

### DIFF
--- a/query/stdlib/influxdata/influxdb/rules.go
+++ b/query/stdlib/influxdata/influxdb/rules.go
@@ -10,6 +10,7 @@ import (
 // 	plan.RegisterPhysicalRules(
 // 		PushDownRangeRule{},
 // 		PushDownFilterRule{},
+// 		PushDownReadTagKeysRule{},
 // 	)
 // }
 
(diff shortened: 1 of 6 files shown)


query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -112,3 +113,68 @@ func (PushDownFilterRule) Rewrite(pn plan.Node) (plan.Node, bool, error) {
...
+		return nil, false, nil
+	}
+
+	// We have passed all of the necessary prerequisites
+	// so construct the procedure spec.
+	newNode := plan.CreatePhysicalNode(fromNode.ID(), &ReadTagKeysPhysSpec{
+		ReadRangePhysSpec: *fromSpec.Copy().(*ReadRangePhysSpec),
+		ValueColumnName:   keysSpec.Column,
+	})
+	plan.ReplaceNode(fromNode, newNode)
```
jlapacik: This is a leaf node in the query plan, so not necessary to replace it. You can just return `newNode`.

query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -112,3 +113,66 @@ func (PushDownFilterRule) Rewrite(pn plan.Node) (plan.Node, bool, error) {
 
 	return pn, true, nil
 }
+
+// ReadTagKeysRule matches 'ReadRange |> keys() |> keep() |> distinct()'.
+// The 'from()' must have already been merged with 'range' and, optionally,
+// may have been merged with 'filter'.
+// If any other properties have been set on the from procedure,
+// this rule will not rewrite anything.
+type ReadTagKeysRule struct{}
```
jlapacik: Would you mind renaming the rule to `PushDownTagKeysRule` just for consistency?

query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -112,3 +113,66 @@ func (PushDownFilterRule) Rewrite(pn plan.Node) (plan.Node, bool, error) {
...
+	}
+
+	// The distinct spec should keep only the value column.
+	if distinctSpec.Column != keysSpec.Column {
+		return nil, false, nil
+	}
+
+	// We have passed all of the necessary prerequisites
+	// so construct the procedure spec.
+	return plan.CreatePhysicalNode(fromNode.ID(), &ReadTagKeysPhysSpec{
```
jlapacik: Sorry I didn't catch this the first time. Instead of copying the node ID from `from`, could you call it `ReadTagKeys`? This way it will be clear which planner operation it is when printing the query plan.

query/stdlib/influxdata/influxdb/rules_test.go
```diff
@@ -118,3 +121,216 @@ func TestPushDownRangeRule(t *testing.T) {
...
+			After: &plantest.PlanSpec{
+				Nodes: []plan.Node{
+					plan.CreatePhysicalNode("ReadRange", readTagKeysSpec(false)),
+					plan.CreatePhysicalNode("count", &universe.CountProcedureSpec{}),
+				},
+				Edges: [][2]int{{0, 1}},
+			},
+		},
+		{
+			Name: "with multiple successors",
```
jlapacik: I think this is a `ReadRange` test. Perhaps you meant to write a similar test for `ReadTagKeys`? I'm fine if you just want to remove it.

query/stdlib/influxdata/influxdb/rules.go
jsternberg: Fixed.

query/stdlib/influxdata/influxdb/rules.go
jsternberg: Done.

query/stdlib/influxdata/influxdb/rules_test.go
jsternberg: I intentionally copied this. I thought it would be a useful test for here too.

jsternberg: Updated the comments.

PR #13527, opened 2019-04-19
Implementation of storage schema TagKeys API

### LINKED ISSUE #13377: Implementation of storage schema TagKeys API
Provide an implementation of the `TagKeys` API

```go
// TagKeys returns an iterator where the values are tag keys for the bucket
// matching the predicate within the time range (start, end].
func (e *Engine) TagKeys(orgID, bucketID influxdb.ID, start, end int64, predicate influxql.Expr) StringIterator
```

This API will be responsible for resolving schema requests such as the following:

* retrieve all tag key names for a bucket
* retrieve all tag key names for a provided measurement `<name>` in a bucket, where
    * predicate is `_measurement == "<name>"`

If the `predicate` is empty or strictly `_measurement == "<name>" [OR _measurement == "<name>", ...]`, the request can be answered entirely by accessing TSM data and scanning the prefix of the series keys.

If the predicate specifies any other tag keys, TSI (index) will need to be consulted. Details of the algorithm are covered in the RFC.

Requires #13309

### LINKED ISSUE #13541: `StringIterator` returned by `TagValues` and `TagKeys` must return accurate `CursorStats`
These APIs must gather TSM reads and return a non-zero-value `CursorStats`

stuartcarnie:
Provide an implementation of the `TagKeys` API

Closes #13377

----

Last commit implements statistics for billing. Closes #13541

### DIFF
--- a/models/points.go
+++ b/models/points.go
@@ -334,6 +334,10 @@ func ParseTags(buf []byte) Tags {
 	return parseTags(buf, nil)
 }
 
+func ParseTagsWithTags(buf []byte, tags Tags) Tags {
+	return parseTags(buf, tags)
+}
+
 func ParseName(buf []byte) []byte {
 	// Ignore the error because scanMeasurement returns "missing fields" which we ignore
 	// when just parsing a key
--- a/models/points_test.go
+++ b/models/points_test.go
@@ -2648,10 +2648,33 @@ func BenchmarkEscapeString_QuotesAndBackslashes(b *testing.B) {
 	}
 }
 
-func BenchmarkParseTags(b *testing.B) {
-	tags := []byte("cpu,tag0=value0,tag1=value1,tag2=value2,tag3=value3,tag4=value4,tag5=value5")
+func BenchmarkParseKeyBytes(b *testing.B) {
+	buf := []byte("cpu,tag0=value0,tag1=value1,tag2=value2,tag3=value3,tag4=value4,tag5=value5")
 	for i := 0; i < b.N; i++ {
-		models.ParseTags(tags)
+		models.ParseKeyBytes(buf)
+	}
+}
+
+func BenchmarkParseKeyBytesWithTags(b *testing.B) {
+	var tags models.Tags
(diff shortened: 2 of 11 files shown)


models/points.go
```diff
@@ -331,6 +331,10 @@ func ParseTags(buf []byte) Tags {
 	return parseTags(buf, nil)
 }
 
+func ParseTagsWithTags(buf []byte, tags Tags) Tags {
```
jacobmarble: Could you just rename `parseTags` to `ParseTags` ?

stuartcarnie: Alas, there is already an exported API called `ParseTags`. I'm following the existing convention where `WithTags` is added to the end when the function takes a reusable tags buffer as an argument.

For example, this API does not take an existing `tags` buffer as an argument

```go
func ParseKeyBytes(buf []byte) ([]byte, Tags)
```

this one does

```go
func ParseKeyBytesWithTags(buf []byte, tags Tags) ([]byte, Tags)
```

stuartcarnie:
@jacobmarble may I request a review of the last commit, which teaches `Tags.Get` how to go faster

models/points.go
```diff
@@ -2088,16 +2092,43 @@ func CompareTags(a, b Tags) int {
 
 // Get returns the value for a key.
 func (a Tags) Get(key []byte) []byte {
-	// OPTIMIZE: Use sort.Search if tagset is large.
-
-	for _, t := range a {
-		if bytes.Equal(t.Key, key) {
-			return t.Value
+	if len(a) < 25 {
```
e-dard: Whilst I'm always happy to see `TODO`s being sorted out, adding a binary search here concerns me a little bit unless it's accompanied by an audit of the codebases.

What I'm mainly worried about is that Tags are being created that are not in order. I assume it wouldn't be too hard to verify the API is being used correctly of course.

stuartcarnie: Moved this commit to #13582 to complete audit of `models.Tags` usage and fix any issues related #13581

Issue #13489, opened 2019-04-17
Document: Add a lastUpdatedAt to each Document


russorat:
as a user
with admin
when i view the list of all documents
then i would like to know when each one was updated

Issue #13488, opened 2019-04-17
Document: Add a createdAt to each Document


russorat:
as a user
with admin
when i view the list of documents
then i would like to know when each one was created

PR #13552, opened 2019-04-22
feat(http): add audit log

kelwang:
Closes https://github.com/influxdata/influxdb/issues/13489
Closes https://github.com/influxdata/influxdb/issues/13488

add audit log for document response, 

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/http/document_service.go
+++ b/http/document_service.go
@@ -140,6 +140,8 @@ func (h *DocumentHandler) handlePostDocument(w http.ResponseWriter, r *http.Requ
 		return
 	}
 
+	h.Logger.Info("document created")
+
 	if err := encodeResponse(ctx, w, http.StatusCreated, newDocumentResponse(req.Namespace, req.Document)); err != nil {
 		logEncodingError(h.Logger, r, err)
 		return
@@ -459,6 +461,8 @@ func (h *DocumentHandler) handleDeleteDocument(w http.ResponseWriter, r *http.Re
 		return
 	}
 
+	h.Logger.Info("document deleted")
+
 	w.WriteHeader(http.StatusNoContent)
 }
 
@@ -525,6 +529,8 @@ func (h *DocumentHandler) handlePutDocument(w http.ResponseWriter, r *http.Reque
 		EncodeError(ctx, err, w)
 		return
 	}
+
+	h.Logger.Info("document updated")
 
 	ds, err := s.FindDocuments(ctx, influxdb.WhereID(req.Document.ID), influxdb.IncludeContent)
 	if err != nil {
--- a/kv/kvlog.go
+++ b/kv/kvlog.go
@@ -108,7 +108,10 @@ func (s *Service) initializeKVLog(ctx context.Context, tx Tx) error {
 	return nil
 }
 
-var errKeyValueLogBoundsNotFound = fmt.Errorf("oplog not found")
+var errKeyValueLogBoundsNotFound = &platform.Error{
+	Code: platform.ENotFound,
+	Msg:  "oplog not found",
+}
 
 func (s *Service) getKeyValueLogBounds(ctx context.Context, tx Tx, key []byte) (*keyValueLogBounds, error) {
 	k := encodeKeyValueIndexKey(key)
@@ -381,7 +384,7 @@ func (s *Service) LastLogEntry(ctx context.Context, k []byte) ([]byte, time.Time
 func (s *Service) firstLogEntry(ctx context.Context, tx Tx, k []byte) ([]byte, time.Time, error) {
 	bounds, err := s.getKeyValueLogBounds(ctx, tx, k)
 	if err != nil {
-		return nil, bounds.StartTime(), err
+		return nil, time.Time{}, err
 	}
 
 	return s.getLogEntry(ctx, tx, k, bounds.StartTime())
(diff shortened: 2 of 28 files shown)


http/kvlog.go
```diff
@@ -0,0 +1,49 @@
...
+
+func logKey(r influxdb.ResourceType, id influxdb.ID) []byte {
+	return []byte(fmt.Sprintf("%s_%s_auditlog", r, id))
+}
+
+func addLog(ctx context.Context, svc influxdb.KeyValueLog, r influxdb.ResourceType, id influxdb.ID, desc string, now time.Time) error {
+	return svc.AddLogEntry(ctx, logKey(r, id), []byte(desc), now)
+}
+
+func time2Str(t time.Time) string {
```
jademcgough: Nitpicky, but time.time2Str is a bit stuttery. Maybe just `toStr`?

http/document_service.go
```diff
@@ -72,30 +80,34 @@ func NewDocumentHandler(b *DocumentBackend) *DocumentHandler {
 }
 
 type documentResponse struct {
-	Links map[string]string `json:"links"`
+	Links         map[string]string `json:"links"`
+	CreatedAt     string            `json:"createAt"`
```
jademcgough: These should be `createdAt` and `updatedAt`, to be consistent with dashboards and tasks. These should also be embedded within dashboard in the response.

kv/kvlog.go
```diff
@@ -381,7 +381,7 @@ func (s *Service) LastLogEntry(ctx context.Context, k []byte) ([]byte, time.Time
 func (s *Service) firstLogEntry(ctx context.Context, tx Tx, k []byte) ([]byte, time.Time, error) {
 	bounds, err := s.getKeyValueLogBounds(ctx, tx, k)
 	if err != nil {
-		return nil, bounds.StartTime(), err
```
jademcgough: changing to zero value 👍

http/document_service.go
jademcgough: The swagger also needs to be updated for this.

jademcgough:
Just a few comments, but generally looks good.

My assumption from the issues was that this was just about adding `UpdatedAt` and `CreatedAt` fields, and setting those on create or update; we're doing that in tasks and dashboards currently. But I like adding an audit log as part of this, and using that as backing for createdAt / updatedAt information. It seems like the `kvlog` was created with this in mind.

http/document_service.go
```diff
@@ -140,7 +152,16 @@ func (h *DocumentHandler) handlePostDocument(w http.ResponseWriter, r *http.Requ
 		return
 	}
 
-	if err := encodeResponse(ctx, w, http.StatusCreated, newDocumentResponse(req.Namespace, req.Document)); err != nil {
+	now := h.timeGenerator.Now()
+
+	if err := addLog(ctx, h.KeyValueLog, influxdb.DocumentsResourceType, req.Document.ID,
```
jademcgough: I wonder if it might be better to have logging be on a lower layer than the http service (specifically, in `kv`). We don't seem to have documents in our cli right now, but if/when we do, logs won't be created from the cli.

http/document_service.go
```diff
@@ -464,6 +492,14 @@ func (h *DocumentHandler) handleDeleteDocument(w http.ResponseWriter, r *http.Re
 		return
 	}
 
+	now := h.timeGenerator.Now()
+	if err := addLog(ctx, h.KeyValueLog, influxdb.DocumentsResourceType, req.ID,
```
jademcgough: see above ^

http/document_service.go
```diff
@@ -72,30 +80,34 @@ func NewDocumentHandler(b *DocumentBackend) *DocumentHandler {
...
 type documentResponse struct {
-	Links map[string]string `json:"links"`
+	Links         map[string]string `json:"links"`
+	CreatedAt     string            `json:"createAt"`
+	LastUpdatedAt string            `json:"lastUpdatedAt"`
 	*influxdb.Document
 }
 
-func newDocumentResponse(ns string, d *influxdb.Document) *documentResponse {
+func newDocumentResponse(ns string, d *influxdb.Document, createdAt, lastUpdatedAt time.Time) *documentResponse {
```
jademcgough: Just an thought, but a pattern I've seen @desa use to avoid having to decorate this information in the response is doing the decoration on the kv layer, using `DocumentDecorator`. This has the added benefit of decoupling KeyValueLog from the Document service.

http/document_service.go
```diff
@@ -72,30 +80,34 @@ func NewDocumentHandler(b *DocumentBackend) *DocumentHandler {
 }
 
 type documentResponse struct {
-	Links map[string]string `json:"links"`
+	Links     map[string]string `json:"links"`
+	CreatedAt string            `json:"createAt"`
```
jademcgough: `json:"createAt"` should be `json:"createdAt"`

jademcgough:
My one big worry about this is using etcd for maintaining an audit log. BoltDB we can do whatever we want with, but etcd has a hard cap of 8GB of data.

This is really elegant, but I'd feel more comfortable storing `createdAt` and `updatedAt` as fields on document, and maintaining an audit log with a separate service (maybe using the existing logging service?).

@desa @goller Do either of you have thoughts/opinions?

http/document_service.go
desa: I agree, this shouldn't be exclusive the the HTTP service

desa:
> My one big worry about this is using etcd for maintaining an audit log. BoltDB we can do whatever we want with, but etcd has a hard cap of 8GB of data.

Yeah this will be an issue for sure. Ideally we could use a system bucket for these values since a log is essentially a time series.

> This is really elegant, but I'd feel more comfortable storing createdAt and updatedAt as fields on document, and maintaining an audit log with a separate service (maybe using the existing logging service?).

Totally agree.

http/document_service.go
```diff
@@ -140,6 +140,8 @@ func (h *DocumentHandler) handlePostDocument(w http.ResponseWriter, r *http.Requ
 		return
 	}
 
+	h.Logger.Info("document created")
```
jademcgough: 👍logging for handlePutDocument (updating) would be great too

PR #13558, opened 2019-04-22
chore(storage): Merge StringIterator cursor stats

jacobmarble:
Merge stats with StringIterator.

Helps https://github.com/influxdata/idpe/issues/3198

### DIFF
--- a/storage/reads/merge.go
+++ b/storage/reads/merge.go
@@ -127,6 +127,7 @@ func (h *resultSetHeap) Pop() interface{} {
 type MergedStringIterator struct {
 	heap      stringIteratorHeap
 	nextValue string
+	stats     cursors.CursorStats
 }
 
 // API compatibility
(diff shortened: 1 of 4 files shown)


storage/reads/merge.go
```diff
@@ -127,23 +127,27 @@ func (h *resultSetHeap) Pop() interface{} {
...
 func NewMergedStringIterator(iterators []cursors.StringIterator) *MergedStringIterator {
 	nonEmptyIterators := make([]cursors.StringIterator, 0, len(iterators))
+	var stats cursors.CursorStats
 
 	for _, iterator := range iterators {
 		// All iterators must be Next()'d so that their Value() methods return a meaningful value, and sort properly.
 		if iterator.Next() {
 			nonEmptyIterators = append(nonEmptyIterators, iterator)
 		}
+		stats.Add(iterator.Stats())
```
stuartcarnie: Stats have to be accumulated at the end, when the `StringIterator` concludes. This is important for gRPC `StringIterator` types, as the data doesn't arrive until the trailing header is sent. 

See:

https://github.com/influxdata/influxdb/blob/662cf578c99fa25383b88067cae73de18472eaf2/storage/reads/merge.go#L50-L58

PR #13561, opened 2019-04-22
Rate Limiting notifications on read/write endpoints

### LINKED ISSUE #13422: The UI should display easy to understand notifications on rate/limiting errors for queries
As a user makes a query or series of queries that exceed usage rates- the backend will reject the request with an HTTP 429 [#3109](https://github.com/influxdata/idpe/issues/3109)

The UI should display a notification that is easy for the user to understand that a rate limit has been reached.

In particular, let's test a dashboard in expectation that it may be easy for users to exceed max: 2 concurrency limits on queries. 

Check: data explorer and dashboard View Editor Overlays

ebb-tide:
Closes #13422

### DIFF
--- a/ui/src/dataLoaders/actions/dataLoaders.ts
+++ b/ui/src/dataLoaders/actions/dataLoaders.ts
@@ -44,7 +44,9 @@ import {notify} from 'src/shared/actions/notifications'
 import {
   TelegrafConfigCreationError,
   TelegrafConfigCreationSuccess,
+  writeLimitReached,
 } from 'src/shared/copy/notifications'
+import {RATE_LIMIT_ERROR_STATUS} from 'src/shared/constants/errors'
 
 type GetState = () => AppState
 
@@ -547,6 +549,13 @@ export const writeLineProtocolAction = (
     dispatch(setLPStatus(RemoteDataState.Done))
   } catch (error) {
     dispatch(setLPStatus(RemoteDataState.Error))
+
+    const errorMessage = _.get(error, 'response.data.message', '')
+    console.error(errorMessage || error)
+
+    if (error.response.status === RATE_LIMIT_ERROR_STATUS) {
+      dispatch(notify(writeLimitReached(errorMessage)))
+    }
   }
 }
 
--- a/ui/src/shared/copy/notifications.ts
+++ b/ui/src/shared/copy/notifications.ts
@@ -519,7 +519,7 @@ export const dashboardExportFailed = (
 
 export const dashboardCreateFailed = () => ({
   ...defaultErrorNotification,
-  message: 'Failed to created dashboard.',
+  message: 'Failed to create dashboard.',
 })
 
 export const dashboardSetDefaultFailed = (name: string) => ({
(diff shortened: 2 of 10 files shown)


ui/src/shared/copy/notifications.ts
```diff
@@ -948,3 +948,42 @@ export const deleteBucketFailed = (): Notification => ({
   ...defaultErrorNotification,
   message: 'Failed to delete bucket',
 })
+
+// Limits
+export const writeLimitReached = (errorMessage: string): Notification => ({
+  ...defaultErrorNotification,
+  message: `You have reached a write limit on this resource: ${errorMessage}.`,
+  duration: INFINITE,
+})
```
lukevmorris: I wonder if `writeLimitReached` and `readLimitReached` should become convenience mehtods for `readWriteLimitReached`?

ui/src/dataLoaders/actions/dataLoaders.ts
```diff
@@ -547,6 +549,13 @@ export const writeLineProtocolAction = (
     dispatch(setLPStatus(RemoteDataState.Done))
   } catch (error) {
     dispatch(setLPStatus(RemoteDataState.Error))
+
+    const errorMessage = _.get(error, 'response.data.message', '')
+    console.error(errorMessage || error)
+
+    if (error.response.status == RATE_LIMIT_ERROR_STATUS) {
```
lukevmorris: I can has threequals?

ui/src/shared/copy/notifications.ts
ebb-tide: I was using the writeLimitReached for line protocol writes, so it felt reasonable here. I'm expecting feedback on error messages.

ui/src/dataLoaders/actions/dataLoaders.ts
ebb-tide: yes you most certainly can

ui/src/limits/actions/index.ts
```diff
@@ -0,0 +1,36 @@
...
+) => {
+  try {
+    const {
+      orgs: {org},
+    } = getState()
+
+    const limits = await getReadWriteLimitsAJAX(org.id)
+
+    const readLimit = limits.read.status === LimitStatus.EXCEEDED
+    const writeLimit = limits.write.status === LimitStatus.EXCEEDED
```
ischolten: nit: `readLimit` and `writeLimit` sound like they would be the actual limit like a number. seems like these should be something along the lines of `readLimitReached`

ui/src/shared/copy/notifications.ts
```diff
@@ -519,7 +519,7 @@ export const dashboardExportFailed = (
 
 export const dashboardCreateFailed = () => ({
   ...defaultErrorNotification,
-  message: 'Failed to created dashboard.',
+  message: 'Failed to create dashboard.',
```
ischolten: nice catch!

ui/src/shared/copy/notifications.ts
ischolten: +1

lukevmorris: I really like the message you've written on L974. Seems like a good wording until we introduce response headers that will return more specific info.

ui/src/limits/actions/index.ts
ebb-tide: 👍

PR #13570, opened 2019-04-22
Generic Builder Card Components

alexpaxton:
Screen Shot 2019-04-22 at 4 50 57 PM

Introduce generic presentational components for the query builder and use them wherever possible
Replace use of `FancyScrollbars` with `DapperScrollbars`
Start taking steps towards new designs for query builder
Refactor and reorganize associated styles

  - [x] Rebased/mergeable
  - [x] Tests pass


ui/src/timeMachine/components/TagSelector.tsx
```diff
@@ -67,7 +73,17 @@ class TagSelector extends PureComponent<Props> {
...
 
   public render() {
-    return <div className="tag-selector">{this.body}</div>
+    const {index} = this.props
+
+    return (
+      <BuilderCard>
+        <BuilderCard.Header
+          title="Filter"
+          onDelete={index !== 0 && this.handleRemoveTagSelector}
```
OfTheDelmer: I feel like `index !== 0` should be a guard inside the handler

alexpaxton: It's definitely funky. If a function is passed in at all the delete button appears on the card, so I threw the guard in here so it doesn't end up with a non-functional delete button. I think long term this will change as cards will be more flexible

PR #13588, opened 2019-04-23
Add rate limiting to dashboard asset

### LINKED ISSUE #13421: The UI should display easy to understand notifications on rate/limiting errors for assets
When a user makes a request to create a resource (bucket/dashboard/task) that exceeds their usage limits the backend will reject the request with a 429 ? (Issue [#3100](https://github.com/influxdata/idpe/issues/3100))

The UI should display a notification that is easy for the user to understand that a rate limit has been reached

- [ ] buckets
- [x] dashboards
- [ ] tasks

ebb-tide:
Closes #13421  for dashboards

### DIFF
--- a/ui/src/cloud/actions/limits.ts
+++ b/ui/src/cloud/actions/limits.ts
@@ -1,5 +1,5 @@
 // API
-import {getReadWriteLimitsAJAX} from 'src/cloud/apis/limits'
+import {getReadWriteLimitsAJAX, getLimitsAJAX} from 'src/cloud/apis/limits'
 
 // Types
 import {AppState} from 'src/types'
@@ -8,11 +8,83 @@ import {AppState} from 'src/types'
 import {notify} from 'src/shared/actions/notifications'
 
 // Constants
-import {readWriteLimitReached} from 'src/shared/copy/notifications'
+import {
+  readWriteLimitReached,
+  resourceLimitReached,
+} from 'src/shared/copy/notifications'
 
-enum LimitStatus {
+// Types
+import {RemoteDataState} from '@influxdata/clockface'
+
+export enum LimitStatus {
   OK = 'ok',
   EXCEEDED = 'exceeded',
+}
+
+interface Limits {
+  rate: {
+    readKBs: number
+    concurrentReadRequests: number
+    writeKBs: number
+    concurrentWriteRequests: number
+  }
+  bucket: {
+    maxBuckets: number
+  }
+  task: {
+    maxTasks: number
+  }
+  dashboard: {
+    maxDashboards: number
+  }
+}
+
+export enum ActionTypes {
+  SetLimits = 'SET_LIMITS',
+  SetLimitsStatus = 'SET_LIMITS_STATUS',
+  SetDashboardLimitStatus = 'SET_DASHBOARD_LIMIT_STATUS',
+}
+
+export type Actions = SetLimits | SetLimitsStatus | SetDashboardLimitStatus
+
+export interface SetLimits {
+  type: ActionTypes.SetLimits
+  payload: {limits: Limits}
+}
+
+export const setLimits = (limits: Limits): SetLimits => {
+  return {
+    type: ActionTypes.SetLimits,
+    payload: {limits},
+  }
+}
+
+export interface SetDashboardLimitStatus {
+  type: ActionTypes.SetDashboardLimitStatus
+  payload: {limitStatus: LimitStatus}
+}
+
+export const setDashboardLimitStatus = (
+  limitStatus: LimitStatus
+): SetDashboardLimitStatus => {
+  return {
+    type: ActionTypes.SetDashboardLimitStatus,
+    payload: {limitStatus},
+  }
+}
+
+export interface SetLimitsStatus {
+  type: ActionTypes.SetLimitsStatus
+  payload: {
+    status: RemoteDataState
+  }
+}
+
+export const setLimitsStatus = (status: RemoteDataState): SetLimitsStatus => {
+  return {
+    type: ActionTypes.SetLimitsStatus,
+    payload: {status},
+  }
 }
 
 export const getReadWriteLimits = () => async (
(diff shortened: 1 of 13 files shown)


ui/src/cloud/apis/limits.ts
```diff
@@ -13,3 +13,17 @@ export const getReadWriteLimitsAJAX = async (orgID: string) => {
     throw error
   }
 }
+
+export const getLimitsAJAX = async (orgID: string) => {
```
bthesorceror: this is a total nit but I don't think we need to specify AJAX here :)

ui/src/dashboards/actions/index.ts
```diff
@@ -212,6 +213,65 @@ export const removeDashboardLabels = (
...
+      dashboards,
+    } = getState()
+
+    const allDashboardNames = dashboards.list.map(d => d.name)
+
+    const clonedName = incrementCloneName(allDashboardNames, dashboard.name)
+
+    const data = await client.dashboards.clone(dashboard.id, clonedName, org.id)
+
+    dispatch(push(`/orgs/${org.id}/dashboards/${data.id}`))
```
bthesorceror: the router change should be last thing we do in a sequence, if the action after this is needed for the page then doing this here hides the implicit coupling and makes it harder to debug in the future.

ebb-tide: 👍

ui/src/cloud/actions/limits.ts
```diff
@@ -34,3 +106,48 @@ export const getReadWriteLimits = () => async (
...
+) => {
+  dispatch(setLimitsStatus(RemoteDataState.Loading))
+  try {
+    const {
+      orgs: {org},
+    } = getState()
+
+    const limits = (await getLimitsAJAX(org.id)) as Limits
+    dispatch(setLimits(limits))
+    dispatch(setLimitsStatus(RemoteDataState.Done))
```
chnn: Might be nice to combine this and the above `setLimits` action into a single action, to avoid double renders due to the synchronous nature of Redux.

ui/src/cloud/components/GetAssetLimits.tsx
```diff
@@ -0,0 +1,57 @@
...
+// Actions
+import {getAssetLimits as getAssetLimitsAction} from 'src/cloud/actions/limits'
+
+// Components
+import {
+  TechnoSpinner,
+  SpinnerContainer,
+  RemoteDataState,
+} from '@influxdata/clockface'
+import {AppState} from 'src/types'
```
chnn:
```suggestion

// Types
import {AppState} from 'src/types'
```

ui/src/cloud/components/GetAssetLimits.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+interface DispatchProps {
+  getAssetLimits: typeof getAssetLimitsAction
+}
+
+type Props = StateProps & DispatchProps
+
+class LimitChecker extends PureComponent<Props> {
+  public componentDidMount() {
+    if (process.env.CLOUD === 'true') {
```
chnn: I think we should use the `CLOUD` variable in `src/shared/constants/index.ts` instead (here and on line 35).

ui/src/cloud/components/GetAssetLimits.tsx
```diff
@@ -0,0 +1,57 @@
...
+  status: RemoteDataState
+}
+
+interface DispatchProps {
+  getAssetLimits: typeof getAssetLimitsAction
+}
+
+type Props = StateProps & DispatchProps
+
+class LimitChecker extends PureComponent<Props> {
```
chnn: Seems to be a different name than the file name.

```suggestion
class GetAssetLimits extends PureComponent<Props> {
```

ui/src/dashboards/components/dashboard_index/DashboardsIndex.tsx
```diff
@@ -87,35 +82,38 @@ class DashboardIndex extends PureComponent<Props, State> {
...
             <div className="col-md-12">
               <GetResources resource={ResourceTypes.Dashboards}>
-                <DashboardsIndexContents
-                  filterComponent={() => (
-                    <SearchWidget
-                      placeholderText="Filter dashboards..."
-                      onSearch={this.handleFilterDashboards}
+                <GetResources resource={ResourceTypes.Labels}>
+                  <GetAssetLimits>
+                    <DashboardsIndexContents
```
chnn: I'm worried that this new component + redux pattern we have for fetching resources will discourage us from fetching resources in parallel, rather than in serial. In this example it looks like the dashboards, labels, and asset limits could all be fetched at the same time.

I don't have a sense of what the user impact of this is (does it make loading the dashboards index much slower?) Also not something to fix in this PR, just thinkin about it.

ui/src/dashboards/components/dashboard_index/DashboardsIndex.tsx
```diff
@@ -125,54 +123,6 @@ class DashboardIndex extends PureComponent<Props, State> {
     )
   }
 
-  private handleCreateDashboard = async (): Promise<void> => {
```
chnn: Woot! Happy to see this extracted into a thunk.

ui/src/cloud/actions/limits.ts
```diff
@@ -34,3 +106,48 @@ export const getReadWriteLimits = () => async (
...
+  } catch (e) {
+    dispatch(setLimitsStatus(RemoteDataState.Error))
+  }
+}
+
+export const checkDashboardLimits = () => async (
```
chnn:
```suggestion
export const checkDashboardLimits = () => (
```

ui/src/dashboards/components/dashboard_index/DashboardsIndexContents.tsx
```diff
@@ -1,34 +1,55 @@
...
+  dashboards: Dashboard[]
+  limitStatus: RemoteDataState
+}
+
+type Props = DispatchProps & StateProps & OwnProps
+
 @ErrorHandling
-export default class DashboardsIndexContents extends Component<Props> {
+class DashboardsIndexContents extends Component<Props> {
+  public async componentDidMount() {
```
chnn:
```suggestion
  public componentDidMount() {
```

ui/src/cloud/components/GetAssetLimits.tsx
ebb-tide: thanks!!

ui/src/dashboards/components/dashboard_index/DashboardsIndex.tsx
ebb-tide: I actually moved these all the fetchers to one component one after another to highlight that it was silly that we were doing them serially. I'll make an issue to consider ways to parallelize/ GetResources could probably take an array and wait until all promises are resolved.

ui/src/dashboards/components/dashboard_index/DashboardsIndex.tsx
ebb-tide: :)

ui/src/dashboards/components/dashboard_index/DashboardsIndex.tsx
chnn: Awesome. It might also be worth just writing specialized fetcher components whenever we need them, rather than trying to make `GetResources` a generic super-flexible component.

PR #13589, opened 2019-04-23
fix(http): don't 400 when auth GET decoration fails

jademcgough:


jademcgough:
@mark-rushakoff oops, thanks for the catch; I assumed it was append style already

PR #13590, opened 2019-04-23
Update add users dropdown with an empty state

### LINKED ISSUE #13472: Add Members dropdown has no empty state
Screen Shot 2019-04-17 at 2 28 15 PM

Should have some sort of empty state in the dropdown menu or show an empty state instead of the dropdown

Palakp41:
Closes #13472 

Updated the dropdown to be disabled if there are no users or users array is empty.

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/members/components/SelectUsers.tsx
+++ b/ui/src/members/components/SelectUsers.tsx
@@ -6,6 +6,7 @@ import {MultiSelectDropdown, Dropdown} from 'src/clockface'
 
 // Types
 import {User} from '@influxdata/influx'
+import {ComponentStatus} from '@influxdata/clockface'
 
 interface Props {
   users: User[]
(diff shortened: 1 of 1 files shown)


ui/src/members/components/SelectUsers.tsx
```diff
@@ -15,22 +16,39 @@ interface Props {
...
   public render() {
-    const {users} = this.props
+    return <>{this.dropdownUsers}</>
+  }
+  private get dropdownUsers() {
+    const {users, selectedUserIDs, onSelect} = this.props
 
-    return (
-      <>
+    if (!users || !users.length) {
```
alexpaxton: I think you could make getters for the `emptyText` and `status` props that use this logic, rather than having to ensure that changes to `MultiSelectDropdown` don't have to be made in two places

ui/src/members/components/SelectUsers.tsx
```diff
@@ -15,22 +16,39 @@ interface Props {
 
 export default class SelectUsers extends PureComponent<Props> {
   public render() {
-    const {users} = this.props
+    return <>{this.dropdownUsers}</>
```
alexpaxton: I don't think you need a fragment here and can just combine `render` and `dropdownUsers()`

ui/src/members/components/SelectUsers.tsx
```diff
@@ -15,22 +16,37 @@ interface Props {
...
+    }
+    return 'Select user'
+  }
+
+  private get dropdownStatus(): ComponentStatus {
+    const {users} = this.props
+    if (!users || !users.length) {
+      return ComponentStatus.Disabled
+    }
+    return
```
alexpaxton: I would return `ComponentStatus.Default` here but not 100% necessary

PR #13621, opened 2019-04-24
refactor(query): enable new planner rules and remove old rules

nathanielc:


query/stdlib/influxdata/influxdb/from.go
```diff
@@ -38,14 +37,6 @@ func init() {
 	flux.ReplacePackageValue("influxdata/influxdb", influxdb.FromKind, flux.FunctionValue(FromKind, createFromOpSpec, fromSignature))
 	flux.RegisterOpSpec(FromKind, newFromOp)
 	plan.RegisterProcedureSpec(FromKind, newFromProcedure, FromKind)
-	plan.RegisterPhysicalRules(
-		FromConversionRule{},
-		MergeFromRangeRule{},
-		MergeFromFilterRule{},
-		FromDistinctRule{},
-		MergeFromGroupRule{},
```
jsternberg: Are we just not going to push down group rules? I'm fine with that, but I just want to confirm that's intended.

nathanielc: I thought it was incompatible with the new rules? Is that not the case?

jsternberg: It is. I just wanted to confirm that, at least temporarily, this was ok. We should create an issue to redo this rule with the new rules also.

jlapacik: @nathanielc that is correct, none of the old rules are compatible.
@jsternberg Group will need to be a separate PR.

query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -2,20 +2,21 @@ package influxdb
...
-// 		PushDownFilterRule{},
-// 		PushDownReadTagKeysRule{},
-// 		PushDownReadTagValuesRule{},
-// 	)
-// }
+func init() {
+	plan.RegisterPhysicalRules(
+		PushDownRangeRule{},
+		PushDownFilterRule{},
+		//PushDownReadTagKeysRule{},
```
jlapacik: I thought we can un-comment this one too?

nathanielc: Its merged to master but not deployed yet. So I wanted to hold off on that one so we can deploy this change without blocking on a storage deploy.

query/stdlib/influxdata/influxdb/from.go
nathanielc: Sounds good, Issue created https://github.com/influxdata/influxdb/issues/13623

Can either of you add more details to that issue if they are needed? Also does that rule need a new RPC call to be exposed from storage?

jlapacik: Sure I'll add it. And yes that will be the final RPC call to be exposed from storage.

PR #13622, opened 2019-04-24
Use streaming interface from client library

bthesorceror:


ui/src/shared/apis/query.ts
```diff
@@ -163,7 +83,7 @@ export const executeQuery = (
   used in the query.
 */
 export const executeQueryWithVars = (
-  url: string,
+  _url: string,
```
ebb-tide: why keep _url if we are not using it?

PR #13638, opened 2019-04-25
Clean up notifications

ebb-tide:
Move notifications action types in to actions file
Use immer library in notifications reducer
Remove v2 notifications, and delete unused notification copy


ui/src/organizations/actions/orgs.ts
```diff
@@ -19,12 +19,12 @@ import {
...
   orgRenameSuccess,
   orgRenameFailed,
-} from 'src/shared/copy/v2/notifications'
+} from 'src/shared/copy/notifications'
 
 // Types
 import {Bucket} from '@influxdata/influx'
 import {Organization, RemoteDataState} from 'src/types'
-import {PublishNotificationAction} from 'src/types/actions/notifications'
+import {PublishNotificationAction} from 'src/shared/actions/notifications'
```
chnn: Could/should this go in `src/types`? Watts centralized a bunch of types there a while ago and I found I really liked the change (don't have to remember where the various types live).

ui/src/shared/actions/notifications.ts
```diff
@@ -1,20 +1,43 @@
...
-export const notify: NotificationsActions.PublishNotificationActionCreator = (
-  notification: NotificationsModels.Notification
-): NotificationsActions.PublishNotificationAction => ({
-  type: 'PUBLISH_NOTIFICATION',
+export type Action =
+  | PublishNotificationAction
+  | DismissNotificationAction
+  | DismissAllNotificationsAction
+
+export enum ActionTypes {
```
chnn: Is there any reason to make a separate `ActionTypes` enum rather than embedding a string literal type in the `Action` types directly? My take is that it provides no benefit and is superfluous boilerplate.... but I haven't been able to get consensus on this yet. nbd either way.

https://influxdata.slack.com/archives/C9RL1S494/p1542669173011600

ebb-tide: oh hmm. that does make sense.

PR #13645, opened 2019-04-25
Fix permissions on view token overlay

### LINKED ISSUE #13633: Bucket with read write permissions shows only read permissions
on the tokens page,
click generate > read / write token
select a bucket in the read section and select the same bucket in the write section and create the token

Click on the token description to see the overview of the token. It will show read but not write. It should show that the bucket is both read and write.

Palakp41:
Closes #13633 

The reduce function to create the object with keys and values to display the permissions was not adding actions correctly. Updated the reduce function and created the test for it in any case if changed in future, it will fail if not meeting the expected values for permissions.

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/authorizations/components/ViewTokenOverlay.test.tsx
+++ b/ui/src/authorizations/components/ViewTokenOverlay.test.tsx
@@ -6,7 +6,8 @@ import {shallow} from 'enzyme'
 import ViewTokenOverlay from 'src/authorizations/components/ViewTokenOverlay'
 
 // Fixtures
-import {authorization as auth} from 'src/authorizations/apis/__mocks__/data'
+import {auth} from 'mocks/dummyData'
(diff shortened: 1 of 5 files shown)


ui/src/authorizations/components/ViewTokenOverlay.test.tsx
```diff
@@ -6,26 +6,50 @@ import {shallow} from 'enzyme'
...
     it('renders!', () => {
-      const {wrapper} = setup()
+      const wrapper = setup()
 
       expect(wrapper.exists()).toBe(true)
       expect(wrapper).toMatchSnapshot()
     })
   })
+
+  describe('if there is all access tokens', () => {
```
alexpaxton: This is great! I like more better tests

PR #13655, opened 2019-04-26
fix(http): use influxdb.Error in ProxyQueryHandler

mark-rushakoff:
In our internal use of this handler, some of these errors weren't
wrapped in the expected influxdb.Error type.

### DIFF
--- a/http/proxy_query_service.go
+++ b/http/proxy_query_service.go
@@ -4,13 +4,13 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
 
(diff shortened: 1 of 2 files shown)


http/proxy_query_service.go
```diff
@@ -4,13 +4,13 @@ import (
...
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
 
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/iocounter"
+	influxdb "github.com/influxdata/influxdb"
```
wolffcm: I guess that you don't need to alias the package here.

PR #13657, opened 2019-04-26
fix(ui) Update flux function docs and examples

sanderson:
Closes #

  - [x] Rebased/mergeable
  - [x] Tests pass

### DIFF
--- a/ui/src/shared/constants/fluxFunctions.ts
+++ b/ui/src/shared/constants/fluxFunctions.ts
@@ -16,10 +16,10 @@ export const FROM: FluxToolbarFunction = {
   ],
   desc:
     'Used to retrieve data from an InfluxDB data source. It returns a stream of tables from the specified bucket. Each unique series is contained within its own table. Each record in the table represents a single point in the series.',
-  example: 'from(bucket: "telegraf/autogen")',
+  example: 'from(bucket: "default")',
   category: 'Inputs',
   link:
-    'https://v2.docs.influxdata.com/v2.0/reference/flux/functions/inputs/from',
+    'https://v2.docs.influxdata.com/v2.0/reference/flux/functions/built-in/inputs/from/',
 }
 
 export const RANGE: FluxToolbarFunction = {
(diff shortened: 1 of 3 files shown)


ui/src/shared/constants/fluxFunctions.ts
```diff
@@ -119,36 +123,28 @@ export const FLUX_FUNCTIONS: FluxToolbarFunction[] = [
...
+    package: '',
     desc: 'Applies an aggregate function to fixed windows of time.',
     example: 'aggregateWindow(every: 1m, fn: mean)',
     category: 'Aggregates',
     link:
-      'https://v2.docs.influxdata.com/v2.0/reference/flux/functions/transformations/aggregates/aggregatewindow',
+      'https://v2.docs.influxdata.com/v2.0/reference/flux/functions/built-in/transformations/aggregates/aggregatewindow/',
   },
   {
-    name: 'assertEquals',
```
russorat: This seems like a strange update. did they really remove the assertEquals function? it's still in the spec: https://github.com/influxdata/flux/blob/master/docs/SPEC.md#assertequals

ui/src/shared/constants/fluxFunctions.ts
```diff
@@ -306,24 +354,25 @@ export const FLUX_FUNCTIONS: FluxToolbarFunction[] = [
...
         desc: 'The column name for the time values. Defaults to `"_time"`.',
         type: 'String',
       },
     ],
+    package: '',
     desc:
       'Computes the rate of change per unit of time between subsequent non-null records. The output table schema will be the same as the input table.',
     example:
-      'derivative(unit: 1s, nonNegative: true, columns: ["_value"], timeColumn: "_time")',
+      'derivative(unit: 1s, nonNegative: true, column: "_value", timeColumn: "_time")',
```
russorat: the spec still says this is "columns": https://github.com/influxdata/flux/blob/master/docs/SPEC.md#derivative

ui/src/shared/constants/fluxFunctions.ts
sanderson: They didn't remove it. It got moved into the `testing` package.

ui/src/shared/constants/fluxFunctions.ts
```diff
@@ -513,9 +581,9 @@ export const FLUX_FUNCTIONS: FluxToolbarFunction[] = [
         type: 'Integer',
       },
       {
-        name: 'columns',
-        desc: 'List of columns by which to sort. Default is `["_value"]`.',
-        type: 'Array of Strings',
+        name: 'column',
+        desc: 'Column by which to sort. Default is `"_value"`.',
```
russorat: spec still has columns: https://github.com/influxdata/flux/blob/master/docs/SPEC.md#highestlowest

ui/src/shared/constants/fluxFunctions.ts
sanderson: You're right. Got over aggressive with my aggregate updates. Will fix this.

ui/src/shared/constants/fluxFunctions.ts
sanderson: Looks like the SPEC needs to be updated here. This was updated with all the aggregate functions so they only accept a single column now:
https://github.com/influxdata/flux/blob/928e70566bea61accd45c43ec04b1692f740964e/stdlib/universe/universe.flux#L171

Issue #13658, opened 2019-04-26
Variable dropdown does not update for keys with same value 


OfTheDelmer:
When a user selects a key name from the variable dropdown that has the same value as another key the dropdown cannot update the selection.

PR #13659, opened 2019-04-26
fix(ui/var/dropdown): display map key names in dropdown

### LINKED ISSUE #13651: Map type variables do not display properly in a dashboard
when i create a map type variable:
image (16)

And then try to use it in a dashboard, it shows the value, but it should show the key.
image (15)

OfTheDelmer:
Closes #13651 

Map variables need to display their key name instead of their key values. This adds a selector to return an array of value entries from a selector. Using entries allows the variable dropdown to handle and render items without switching on types. However, displaying keys and values revealed an issue where variable two keys could have the same value. This can be resolved in a separate issue, https://github.com/influxdata/influxdb/issues/13658.

Screen Shot 2019-04-26 at 11 08 19 AM

Screen Shot 2019-04-26 at 11 08 55 AM

Screen Shot 2019-04-26 at 11 10 41 AM

  - [ ] Rebased/mergeable
  - [ ] Tests pass

### DIFF
--- a/ui/src/dashboards/components/variablesControlBar/VariableDropdown.tsx
+++ b/ui/src/dashboards/components/variablesControlBar/VariableDropdown.tsx
@@ -10,14 +10,14 @@ import {Dropdown, DropdownMenuColors} from 'src/clockface'
 import {selectVariableValue} from 'src/dashboards/actions/index'
 
 // Utils
-import {getValuesForVariable} from 'src/variables/selectors'
+import {getVariableValuesForDropdown} from 'src/dashboards/selectors'
 
 // Types
 import {AppState} from 'src/types'
 
 interface StateProps {
-  values: string[]
-  selectedValue: string
+  values: [string, string][]
+  selectedKey: string
 }
 
 interface DispatchProps {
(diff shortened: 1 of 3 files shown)


ui/src/dashboards/components/variablesControlBar/VariableDropdown.tsx
```diff
@@ -10,14 +10,14 @@ import {Dropdown, DropdownMenuColors} from 'src/clockface'
...
-import {getValuesForVariable} from 'src/variables/selectors'
+import {getVariableValuesForDropdown} from 'src/dashboards/selectors'
 
 // Types
 import {AppState} from 'src/types'
 
 interface StateProps {
-  values: string[]
-  selectedValue: string
+  values: [string, string][]
```
chnn: `Array<{name: string, value: string}>` might be easier to understand. Have you been writing elixir? This feels elixirish 😸

ui/src/dashboards/components/variablesControlBar/VariableDropdown.tsx
```diff
@@ -37,47 +37,54 @@ type Props = StateProps & DispatchProps & OwnProps
...
           titleText="No Values"
           customClass="variable-dropdown--dropdown"
           menuColor={DropdownMenuColors.Amethyst}
         >
-          {dropdownValues.map(v => (
-            <Dropdown.Item key={v} id={v} value={v}>
-              {v}
+          {dropdownValues.map(([key]) => (
+            /*
+              Use key as value since they are unique otherwise
```
chnn: If I understand this, it is handling the case where there are multiple items in a map variable with the same key, right? Like:
```
foo,123
foo,456
```
In this case, shouldn't `foo` appear twice in the dropdown? Not sure.

ui/src/variables/selectors/index.tsx
```diff
@@ -67,6 +67,33 @@ export const getValuesForVariable = (
...
+
+interface ArgumentMap {
+  map: {[key: string]: string}
+  list: string[]
+}
+
+export const getArgumentValuesForVariable = <T extends keyof ArgumentMap>(
+  state: AppState,
+  variableID: string
+): ArgumentMap[T] => {
```
chnn: This feels a bit too clever... took a while to understand how it could possibly be correct. What about returning `VariableArguments` instead and then using type assertions in the calling code? As far as I can tell, the generic return type is a roundabout way of doing a type assertion.

ui/src/dashboards/components/variablesControlBar/VariableDropdown.tsx
chnn: Oh nevermind, I think I misunderstood. Disregard.

ui/src/variables/selectors/index.tsx
OfTheDelmer: sgtm. I was a bit concerned about the cleverness

PR #13663, opened 2019-04-26
fix(ui): task update

### LINKED ISSUE #12305: bug: chronograf on task update sends both every and cron.
bug: chronograph on task update sends both every and cron.  It should only send one or the other.

121watts:
Closes #12305


121watts:
Also made a change to client https://github.com/influxdata/influxdb-client-js/pull/118

PR #13691, opened 2019-04-29
fix(storage/reads): reserve data for the tags column when building a table

jsternberg:


storage/reads/table.go
```diff
@@ -127,6 +127,7 @@ func (t *table) appendTags() {
 		if v != nil {
 			b := arrow.NewStringBuilder(t.alloc)
 			b.Reserve(t.l)
+			b.ReserveData(t.l*len(v))
```
jlapacik: Why a call to both `Reserve` and `ReserveData`?

jsternberg: The call to `Reserve` is for the number of entries and the second is for the binary data that will be appended so they're different numbers.

PR #13698, opened 2019-04-29
fix(storage): reduce tsm level metrics cardinality

tmgordeeva:
This should have cut off TSM file levels at 4+.

### DIFF
--- a/tsdb/tsm1/file_store.go
+++ b/tsdb/tsm1/file_store.go
@@ -309,7 +309,6 @@ type fileTracker struct {
 	metrics   *fileMetrics
 	labels    prometheus.Labels
 	diskBytes uint64
-	levels    uint64
 }
 
 func newFileTracker(metrics *fileMetrics, defaultLabels prometheus.Labels) *fileTracker {
@@ -334,7 +333,7 @@ func (t *fileTracker) SetBytes(bytes map[int]uint64) {
 	total := uint64(0)
 	labels := t.Labels()
 	for k, v := range bytes {
-		labels["level"] = fmt.Sprintf("%d", k)
+		labels["level"] = formatLevel(uint64(k))
 		t.metrics.DiskSize.With(labels).Set(float64(v))
 	}
 	atomic.StoreUint64(&t.diskBytes, total)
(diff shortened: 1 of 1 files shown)


tsdb/tsm1/file_store.go
```diff
@@ -339,22 +343,29 @@ func (t *fileTracker) AddBytes(bytes uint64, level int) {
...
+			level = 5
+		}
+		labels["level"] = fmt.Sprintf("%d", level)
+		if level > maxLevel {
+			maxLevel = level
 		}
 		t.metrics.Files.With(labels).Set(float64(v))
 	}
-	atomic.StoreUint64(&t.levels, level)
+	atomic.StoreUint64(&t.levels, maxLevel)
```
e-dard: What is `t.levels` used for? It's never returned from anywhere? Is this a regression that's been introduced?

tsdb/tsm1/file_store.go
```diff
@@ -339,22 +343,29 @@ func (t *fileTracker) AddBytes(bytes uint64, level int) {
...
+	maxLevel := uint64(0)
 	for k, v := range files {
-		labels["level"] = fmt.Sprintf("%d", k)
-		if uint64(k) > level {
-			level = uint64(k)
+		level := uint64(k)
+		if level > 5 {
+			level = 5
+		}
+		labels["level"] = fmt.Sprintf("%d", level)
```
e-dard: If the level is 5 or over then the label should be `5+` really.

dgnorton: nit: `6+` since it's `if level > 5`

tsdb/tsm1/file_store.go
tmgordeeva: Used to keep track of metrics to clear later, but we could equally just set all the levels we care about to 0.

e-dard: Yeah I don't understand what it's for. We should clear this up.

benbjohnson:
It looks like this does levels `1`, `2`, `3`, `4`, & `5+`. Should it just go up to `4+` instead?

tmgordeeva:
Now goes to 4+ instead of 5+.

PR #13700, opened 2019-04-29
Newly activated tasks should not attempt to 'catch up'

lyondhill:


task/backend/coordinator/coordinator_test.go
```diff
@@ -212,6 +212,52 @@ func TestCoordinator(t *testing.T) {
 	}
 }
 
+func TestCoordinator_ClaimTask(t *testing.T) {
+	ts := inmemTaskService()
```
mark-rushakoff: Since there are time-based delays in this test it should be `t.Parallel()`.

task/backend/coordinator/coordinator.go
```diff
@@ -126,7 +131,14 @@ func (c *Coordinator) UpdateTask(ctx context.Context, id platform.ID, upd platfo
 	}
 
 	// If enabling the task, claim it after modifying the script.
-	if task.Status == string(backend.TaskActive) {
+	if task.Status != oldTask.Status && task.Status == string(backend.TaskActive) {
+		// don't catch up on all the missed task runs while disabled
+		newLatestCompleted := time.Now().UTC().Format(time.RFC3339)
```
mark-rushakoff: The other uses of time.Now in `influxdb/task` and its subpackages are related to timestamps for run logs, setting created/updated at, and managing the scheduler's ticker.

Instead of calling time.Now() here, add a new method to the Scheduler interface that exposes the scheduler's current time, and use that. Then you can avoid the hardcoded time.Sleep in the new test.

task/backend/coordinator/coordinator_test.go
```diff
@@ -212,6 +212,52 @@ func TestCoordinator(t *testing.T) {
 	}
 }
 
+func TestCoordinator_ClaimTask(t *testing.T) {
```
mark-rushakoff: This test name should be more descriptive.

PR #13705, opened 2019-04-29
feat(stdlib): enable the tag keys optimization for meta queries

jsternberg:


jlapacik:
I found two issues:

1. Pushing down tagKeys to storage does not return `_start` or `_stop`, but the Flux implementation does.
2. Pushing down tagKeys to storage returns `\x00` for `_measurement`.

This was not caught during testing, so we need the proper tests to accompany this update. A flux end to end test for `v1.tagKeys` should do the trick.

`_start` and `_stop` aren't really tag keys, so I'm not sure if they should be returned from the call. However this would require updating the Flux implementation of `tagKeys` which will need to be treated as a breaking and change and first communicated with the Chronograf team. It might be simpler to just add those keys to the pushed down result. But eventually I think they should be removed.

jsternberg:
Both of the above have been fixed in #13707.

PR #13707, opened 2019-04-29
fix(storage/reads): translate measurement and field tag key names

jsternberg:
Translate the measurement and field tag key names to their non-storage
names and add the `_start` and `_stop` tag keys to the output since
they aren't real tags, but ones that are added by range.

https://github.com/influxdata/influxdb/pull/13705#issuecomment-487752507


jlapacik:
@jsternberg Can you create an issue for adding the appropriate Flux end-to-end test, to exercise this code path?

jsternberg:
#13710

PR #13709, opened 2019-04-29
fix(task): lookup fully hydrated task after creation

### LINKED ISSUE #12021: TaskCreate does not return time of task creation
Swagger doc mentions that task create response object should have `createdAt`, `updatedAt` and `lastCompleted` entries.
I am not sure if `updatedAt` and `lastCompleted` are really should be there, since task has just been created, but `createdAt` make sense.

- Create task (my examples are using python):
```
requests.post('http://localhost:9999/api/v2/tasks', data='{"orgID":"036f43714301f000","name":"task_1","flux":%s}' % json.dumps(flux_script), headers={'Authorization':'Token %s' % token})
```
where `flux_script` is 
```
'option task = {name:"test_task_name", every:60s} from(bucket:"b_1") |> range(start: -1h)'
```

- request response: 
```
{
  "authorizationID": "036f42af1a81f002", 
  "every": "1m0s", 
  "flux": "option task = {name:\"test_task_name\", every:60s} from(bucket:\"b_1\") |> range(start: -1h)", 
  "id": "036f46407fcbf000", 
  "labels": [], 
  "links": {
    "labels": "/api/v2/tasks/036f46407fcbf000/labels", 
    "logs": "/api/v2/tasks/036f46407fcbf000/logs", 
    "members": "/api/v2/tasks/036f46407fcbf000/members", 
    "owners": "/api/v2/tasks/036f46407fcbf000/owners", 
    "runs": "/api/v2/tasks/036f46407fcbf000/runs", 
    "self": "/api/v2/tasks/036f46407fcbf000"
  }, 
  "name": "test_task_name", 
  "org": "o_1", 
  "orgID": "036f43714301f000", 
  "status": "active"
}
```

goller:
Closes #12021

When creating a task, not all fields were available in the response.  This loads the task from the backing store to return all fields.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/kv/task.go
+++ b/kv/task.go
@@ -558,6 +558,7 @@ func (s *Service) createTask(ctx context.Context, tx Tx, tc influxdb.TaskCreate)
 		tc.Status = string(backend.TaskActive)
 	}
 
+	createdAt := time.Now().UTC().Format(time.RFC3339)
 	task := &influxdb.Task{
 		ID:              s.IDGenerator.ID(),
 		OrganizationID:  org.ID,
@@ -568,7 +569,8 @@ func (s *Service) createTask(ctx context.Context, tx Tx, tc influxdb.TaskCreate)
 		Flux:            tc.Flux,
 		Every:           opt.Every.String(),
 		Cron:            opt.Cron,
-		CreatedAt:       time.Now().UTC().Format(time.RFC3339),
+		CreatedAt:       createdAt,
+		LatestCompleted: createdAt,
 	}
 	if opt.Offset != nil {
 		task.Offset = opt.Offset.String()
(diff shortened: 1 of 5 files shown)


kv/task.go
```diff
@@ -568,7 +569,8 @@ func (s *Service) createTask(ctx context.Context, tx Tx, tc influxdb.TaskCreate)
 		Flux:            tc.Flux,
 		Every:           opt.Every.String(),
 		Cron:            opt.Cron,
-		CreatedAt:       time.Now().UTC().Format(time.RFC3339),
+		CreatedAt:       createdAt,
+		LatestCompleted: createdAt,
```
docmerlin: I don't understand this line.  Why are we setting the LatestCompleted to createdAt?

lyondhill: When a task is first created we don't have a latest completed time. If a task goes to the scheduler without latest completed it will start at 1970. A while ago we decided to make latest completed never be before createdAt. KV was not following the standard set by the other stores and by updating the test it exposed this difference.

PR #13717, opened 2019-04-30
fix: notes covered by adjacent cells

### LINKED ISSUE #13669: Notes popover isn't on top
when you hover over a cell with a note, the note should be on top.

image

I also noticed that in chrome 73.0.3683.103, you have to actually click on the speech bubble to get the note to show up.

121watts:
Closes #13669

### Before
Kapture 2019-04-30 at 11 20 18

### After
Kapture 2019-04-30 at 11 19 12


ui/src/shared/components/cells/CellHeaderNoteTooltip.scss
```diff
@@ -1,8 +1,11 @@
 @import "src/style/modules";
 
+#notes-portal {
+  z-index: $z--notes-portal;
```
ischolten: nice!

PR #13720, opened 2019-04-30
Feat/read group

### LINKED ISSUE #13623: Replace MergeFromGroupRule with new RPC based rules
We have switched over to using the new planner rules that use a single RPC call per physical operation. We need to finish replacing the existing rules, specifically the MergeFromGroupRule was removed and its replacement needs to be created.

jlapacik:
Closes #13623 .

Re-introduce group pushdown to storage.

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/query/stdlib/influxdata/influxdb/operators.go
+++ b/query/stdlib/influxdata/influxdb/operators.go
@@ -14,9 +14,35 @@ import (
 
 const (
 	ReadRangePhysKind     = "ReadRangePhysKind"
+	ReadGroupPhysKind     = "ReadGroupPhysKind"
 	ReadTagKeysPhysKind   = "ReadTagKeysPhysKind"
 	ReadTagValuesPhysKind = "ReadTagValuesPhysKind"
 )
+
+type ReadGroupPhysSpec struct {
+	plan.DefaultCost
+	ReadRangePhysSpec
+
+	GroupMode flux.GroupMode
+	GroupKeys []string
+
+	AggregateMethod string
+}
+
+func (s *ReadGroupPhysSpec) Kind() plan.ProcedureKind {
+	return ReadGroupPhysKind
+}
+
+func (s *ReadGroupPhysSpec) Copy() plan.ProcedureSpec {
+	ns := new(ReadGroupPhysSpec)
+	ns.ReadRangePhysSpec = *s.ReadRangePhysSpec.Copy().(*ReadRangePhysSpec)
+
+	ns.GroupMode = s.GroupMode
+	ns.GroupKeys = s.GroupKeys
+
+	ns.AggregateMethod = s.AggregateMethod
+	return ns
+}
 
 type ReadRangePhysSpec struct {
 	plan.DefaultCost
--- a/query/stdlib/influxdata/influxdb/rules.go
+++ b/query/stdlib/influxdata/influxdb/rules.go
@@ -13,6 +13,7 @@ func init() {
 	plan.RegisterPhysicalRules(
 		PushDownRangeRule{},
 		PushDownFilterRule{},
+		PushDownGroupRule{},
 		PushDownReadTagKeysRule{},
 		PushDownReadTagValuesRule{},
 	)
--- a/query/stdlib/influxdata/influxdb/rules.go
+++ b/query/stdlib/influxdata/influxdb/rules.go
@@ -18,6 +18,44 @@ func init() {
 	)
 }
 
+// PushDownGroupRule pushes down a group operation to storage
+type PushDownGroupRule struct{}
+
+func (rule PushDownGroupRule) Name() string {
+	return "PushDownGroupRule"
+}
+
+func (rule PushDownGroupRule) Pattern() plan.Pattern {
+	return plan.Pat(universe.GroupKind, plan.Pat(ReadRangePhysKind))
+}
+
+func (rule PushDownGroupRule) Rewrite(node plan.Node) (plan.Node, bool, error) {
+	src := node.Predecessors()[0].ProcedureSpec().(*ReadRangePhysSpec)
+	grp := node.ProcedureSpec().(*universe.GroupProcedureSpec)
+
+	switch grp.GroupMode {
+	case
+		flux.GroupModeBy:
+	default:
+		return node, false, nil
+	}
+
+	for _, col := range grp.GroupKeys {
+		// Storage can only group by tag keys.
+		// Note the columns _start and _stop are ok since all tables
+		// coming from storage will have the same _start and _values.
+		if col == execute.DefaultTimeColLabel || col == execute.DefaultValueColLabel {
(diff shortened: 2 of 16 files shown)


mock/storage_reads.go
```diff
@@ -78,7 +78,7 @@ type GroupResultSet struct {
 	ErrFunc   func() error
 }
 
-func NewGroupResultSet() *GroupResultSet {
+func NewGroupedResultSet() *GroupResultSet {
```
wolffcm: I don't have a strong opinion either way, but it would be good for the name of this function to match the type it returns, either `Group` or `Grouped`.

query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -13,11 +13,48 @@ func init() {
...
+func (rule PushDownGroupRule) Pattern() plan.Pattern {
+	return plan.Pat(universe.GroupKind, plan.Pat(ReadRangePhysKind))
+}
+
+func (rule PushDownGroupRule) Rewrite(node plan.Node) (plan.Node, bool, error) {
+	src := node.Predecessors()[0].ProcedureSpec().(*ReadRangePhysSpec)
+	grp := node.ProcedureSpec().(*universe.GroupProcedureSpec)
+
+	// GroupExcept not a supported
+	if grp.GroupMode == flux.GroupModeExcept {
```
wolffcm: Instead checking for the unsupported group mode, maybe look for the group modes that _are_ supported, and return otherwise?  That way if another group mode is added, we won't accidentally push it down.

query/stdlib/influxdata/influxdb/rules_test.go
```diff
@@ -122,6 +122,216 @@ func TestPushDownRangeRule(t *testing.T) {
...
+			Name: "un-group",
+			// ReadRange -> group() => ReadGroup
+			Rules: []plan.Rule{
+				influxdb.PushDownGroupRule{},
+			},
+			Before: &plantest.PlanSpec{
+				Nodes: []plan.Node{
+					plan.CreateLogicalNode("ReadRange", &readRange),
+					plan.CreateLogicalNode("group", &universe.GroupProcedureSpec{
+						GroupMode: flux.GroupModeNone,
```
wolffcm: There has been confusion in the past about what `GroupModeNone` means.

The Flux definition seems to describe it as a no-op:
https://github.com/influxdata/flux/blob/master/group_mode.go#L7-L8

This would be consistent with just letting storage produce the natural grouping, leaving the output of the read grouped by tag keys.  

Is it possible to produce a `GroupModeNode` in a `GroupProcedureSpec` with Flux syntax?  I _think_ the answer is "no"---I believe that a plain `group()` in Flux will produce `GroupModeBy`, with an empty column list.

I think that we needed `GroupModeNone` in the past because there was just one read RPC, and we needed a way to express, "don't do anything with respect to grouping."  My guess is that it's obsolete now.

query/stdlib/influxdata/influxdb/rules_test.go
```diff
@@ -122,6 +122,216 @@ func TestPushDownRangeRule(t *testing.T) {
...
+						GroupKeys: []string{"_measurement", "tag0", "tag1"},
+					}),
+					plan.CreatePhysicalNode("count", &universe.CountProcedureSpec{}),
+				},
+				Edges: [][2]int{
+					{0, 1},
+					{0, 2},
+				},
+			},
+			After: &plantest.PlanSpec{
```
wolffcm: You can use the `NoChange: true` field to indicate that a rule should not be applied, I find it makes tests a bit easier to read.

query/stdlib/influxdata/influxdb/operators.go
```diff
@@ -14,10 +14,36 @@ import (
...
 )
 
+type ReadGroupPhysSpec struct {
+	plan.DefaultCost
+	ReadRangePhysSpec
+
+	GroupMode flux.GroupMode
+	GroupKeys []string
+
+	AggregateMethod string
```
wolffcm: I guess this field is not used yet, but will be in the future?

query/stdlib/influxdata/influxdb/rules_test.go
wolffcm: Just thinking aloud, if someday `ReadGroup` can also have aggregates pushed into it, you might want `GroupModeNode` to indicate you want to aggregate over the natural grouping.

query/stdlib/influxdata/influxdb/source.go
```diff
@@ -156,6 +157,84 @@ func createReadFilterSource(s plan.ProcedureSpec, id execute.DatasetID, a execut
...
+		s.alloc,
+	)
+	if err != nil {
+		return err
+	}
+	return s.processTables(ctx, tables, stop)
+}
+
+func createReadGroupSource(s plan.ProcedureSpec, id execute.DatasetID, a execute.Administration) (execute.Source, error) {
+	span, ctx := tracing.StartSpanFromContext(context.TODO())
```
wolffcm: You could use `a.Context()` instead of `context.TODO()` here, although keeping contexts as fields in structs is discouraged, so it seems like `execute.Administration` shouldn't do that.  Ideally the context would be an argument to this function.

storage/reads/group_resultset.go
```diff
@@ -42,7 +43,7 @@ func GroupOptionNilSortLo() GroupOption {
 	}
 }
 
-func NewGroupResultSet(ctx context.Context, req *datatypes.ReadRequest, newCursorFn func() (SeriesCursor, error), opts ...GroupOption) GroupResultSet {
+func NewGroupedResultSet(ctx context.Context, req *datatypes.ReadGroupRequest, newCursorFn func() (SeriesCursor, error), opts ...GroupOption) GroupResultSet {
```
wolffcm: Same comment as above regarding `Group` vs `Grouped`.

query/stdlib/influxdata/influxdb/operators.go
jlapacik: Correct.

query/stdlib/influxdata/influxdb/rules_test.go
jlapacik: Looking into it now. I thought GroupModeNone meant un-grouping, i.e all results in a single table. But I will confirm that.

jlapacik: @wolffcm you are correct. GroupModeNone is obsolete. I will fix.

jlapacik:
@wolffcm that's correct. There's also still some refactoring left to do. In particular the ReadGroup RPC message still has a `Hints` field which is only used in tests. However the tests are useful and I did not want to remove them. And refactoring them would require a non-trivial amount of work. So instead I created an issue to tackle once this is merged #13730.

PR #13740, opened 2019-05-01
fix(query/influxql): update transpiler for moved/renamed code

jsternberg:
The `databases()` function was moved into the `influxdata/influxdb/v1`
package and so it wouldn't work anymore. The `percentile()` call was
changed to `quantile()` and the argument was changed to `q`. This also
updates `median()` to use `median()` since we now produce AST's and not
the spec so we can use the `median()` definition instead.

### DIFF
--- a/query/influxql/transpiler.go
+++ b/query/influxql/transpiler.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"path/filepath"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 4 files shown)


query/influxql/transpiler.go
```diff
@@ -705,3 +716,26 @@ func (t *transpilerState) assignment(expr ast.Expression) *ast.Identifier {
...
 	}
 }
+
+// requireImport will ensure the import is included in the file and return
+// the variable used to access the package.
+func (t *transpilerState) requireImport(pkgpath string) *ast.Identifier {
+	// Check to see if the import already exists.
+	for _, decl := range t.file.Imports {
+		if decl.Path.Value == pkgpath {
+			return decl.As
```
jlapacik: This is guaranteed not to be empty, even if not explicitly specified right?

Issue #12891, opened 2019-03-26
Flaky: Flux end to end tests


e-dard:
I got this red run on a PR: https://circleci.com/gh/influxdata/influxdb/36696

```
--- FAIL: TestFluxEndToEnd (23.09s)
    --- PASS: TestFluxEndToEnd/aggregate_empty_window.flux (0.42s)
    --- FAIL: TestFluxEndToEnd/aggregate_fill_window.flux (0.61s)
    --- FAIL: TestFluxEndToEnd/aggregate_window.flux (0.53s)
    --- FAIL: TestFluxEndToEnd/limit_offset.flux (0.58s)
    --- FAIL: TestFluxEndToEnd/lowestCurrent.flux (0.53s)
```

I've attached the build file.

[build_36696_step_108_container_0.txt](https://github.com/influxdata/influxdb/files/3007865/build_36696_step_108_container_0.txt)

aanthony1243:
@e-dard I can dig deeper into this this week, but it looks like if there's something flaky (e.g. they sometimes pass), that it's likely the write-path in the test isn't reliable.  will try to figure if the reliability is expected (e.g. it works but is slower than the test expects) or if it seems to be something more critical in how the test framework is set up.  

but if you look at the diffs printed from the tests, the test query returns with zero rows and no errors, which usually is the case when no data comes across from storage.

e-dard:
@aanthony1243 I would be surprised if this was to do with the write path since it's synchronous—clients are blocked until the write is made readable.

aanthony1243:
@e-dard this involves the `to` function which is sort-of-on and sort-of-off the write path.  it reuses some of the same code, but I'm not sure that it blocks in the way that you describe.  could you point me to where that block occurs in the standard write path?  it would be idea if `to` would follow the same convention as closely as possible.

e-dard:
If you call `storage.Engine.WritePoints` then it will block until it's readable.

jlapacik:
When a Flux end-to-end test succeeds, it should return 3 query results with the following names:

* want
* got
* diff

In the case that it is run against influxdb, there should also be a result labeled `to1`. This is because tables written to a bucket as a result of a call to `to` will be implicitly yielded to the client.

So far when I get test failures, I only see results labeled `want`, `diff`, and `to1`. This seems to suggest that the write to the bucket is performed but that the query itself is never actually executed.

[...]

PR #13752, opened 2019-05-02
fix(readservice): normalize special tag keys after reducing request p…

affo:
…redicate

When InfluxDB receives a read filter request, it parses the predicate and transforms special tag keys (`_measurement` and `_field`) to its internal encoding:

https://github.com/influxdata/influxdb/blob/2909927ab0eaa2943905f5beab5bbb8d71ca18b8/storage/reads/predicate.go#L260

The encoding is later reverted to reflect the emitted format.

However, if the predicate contains a filtering expression on the value, that expression is reduced to be evaluated later.

The encoding must be reverted after this operation is performed, otherwise the reduction will produce wrong values.

- [ ] Rebased/mergeable
- [ ] Tests pass


affo:
@jlapacik that test would require something similar to an end-to-end test that does not pass through Flux (issues with flaky tests https://github.com/influxdata/influxdb/issues/12891).

I filed an issue to write custom end-to-end tests to test `influxdb.Reader` #13757 .

PR #13756, opened 2019-05-02
Ensure Tag(Keys|Values) APIs never return (nil, nil)

stuartcarnie:
Formalized this post condition in the documentation and added additional
unit tests.

Added a nil guard and unit test to WriteStringIterator.

### DIFF
--- a/storage/reads/string_iterator_writer.go
+++ b/storage/reads/string_iterator_writer.go
@@ -38,6 +38,10 @@ func (w *StringIteratorWriter) WrittenN() int {
 }
 
 func (w *StringIteratorWriter) WriteStringIterator(si cursors.StringIterator) error {
+	if si == nil {
+		return nil
+	}
+
 	for si.Next() {
 		v := si.Value()
 		if v == "" {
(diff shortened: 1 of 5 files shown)


tsdb/tsm1/engine_schema_test.go
```diff
@@ -238,6 +238,23 @@ memB,host=EB,os=macOS value=1.3 201`)
 			exp:      []string{"0B", "AB", "BB", "CB", "DB", "EB"},
 			expStats: cursors.CursorStats{ScannedValues: 3, ScannedBytes: 24},
 		},
+
+		// ***********************
+		// * other scenarios
+		// ***********************
+		{
+			// ensure StringIterator is never nil
+			name: "predicate/no candidate series",
```
jlapacik: I'm assuming this test case would panic previously?

storage/reads/string_iterator_writer.go
```diff
@@ -38,6 +38,10 @@ func (w *StringIteratorWriter) WrittenN() int {
 }
 
 func (w *StringIteratorWriter) WriteStringIterator(si cursors.StringIterator) error {
+	if si == nil {
```
jlapacik: This should never be true, right? Just a precaution?

tsdb/tsm1/engine_schema_test.go
stuartcarnie: correct – no more panic

storage/reads/string_iterator_writer.go
stuartcarnie: right - only precautionary

PR #13765, opened 2019-05-03
Refactor/store

jlapacik:
Closes #

This PR removes the `Read` method from the gRPC API.
It also removes the old `FromProcedureSpec` as we now have multiple physical read calls.

- [ ] Rebased/mergeable
- [ ] Tests pass


query/stdlib/influxdata/influxdb/storage.go
```diff
@@ -65,138 +63,27 @@ func (l StaticLookup) Watch() <-chan struct{} {
...
 type GroupMode int
 
 const (
-	// GroupModeDefault specifies the default grouping mode, which is GroupModeAll.
-	GroupModeDefault GroupMode = 0
 	// GroupModeNone merges all series into a single group.
-	GroupModeNone GroupMode = 1 << iota
-	// GroupModeAll produces a separate table for each series.
-	GroupModeAll
+	GroupModeNone GroupMode = iota
```
wolffcm: Is this correct?  I would think that `GroupModeNone` wouldn't do any grouping, i.e., leave the output grouped by the tag keys.

jlapacik: I was definitely confused by this, but Flux as it stands right now has no concept of GroupModeNode. It can do group by, or group except, that's it. Storage however defines GroupNone to to be that the result is "all one group", and it can make optimizations for that specific case. To remove the ambiguity, I removed GroupModeAll, because there's no way to specify that type of grouping using a `group()` operation. It is just the default from storage.

stuartcarnie: We should plan to drop the `GroupModeNone` constant completely and have a new physical operation that represents this behavior. After all, it is implemented separately in storage too and therefore should be a separate RPC.

storage/reads/stream_reader.go
```diff
@@ -44,23 +45,28 @@ func (*emptyStatistics) Stats() cursors.CursorStats {
...
-	c       datatypes.Storage_ReadClient
+	client  streamClient
 	trailer metadata.MD
 }
 
 // NewStorageReadClient returns a new StorageReadClient which implements
 // StreamReader and reads the gRPC trailer to return CursorStats.
-func NewStorageReadClient(c datatypes.Storage_ReadClient) *StorageReadClient {
-	return &StorageReadClient{c: c}
+func NewStorageReadClient(client streamClient) *StorageReadClient {
```
stuartcarnie: This is an exported API, so I would expect `streamClient` to be exported as `StreamClient` for correctness.

PR #13767, opened 2019-05-03
Update task run to not display NaN seconds

### LINKED ISSUE #13766: Task run displaying Nan seconds in duration
Task run displays nan seconds when task is ran while being inactive. Update task to display error if not scheduled to run.

Palakp41:
Closes #13766 

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/tasks/actions/index.ts
+++ b/ui/src/tasks/actions/index.ts
@@ -501,6 +501,8 @@ export const runTask = (taskID: string) => async dispatch => {
     await client.tasks.startRunByTaskID(taskID)
     dispatch(notify(taskRunSuccess()))
   } catch (error) {
+    const message = getErrorMessage(error)
+    dispatch(notify(copy.taskRunFailed(message)))
     console.error(error)
   }
 }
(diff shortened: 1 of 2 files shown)


ui/src/tasks/actions/index.ts
```diff
@@ -558,12 +560,15 @@ export const createTaskFromTemplate = (template: ITaskTemplate) => async (
 export const runDuration = (finishedAt: Date, startedAt: Date): string => {
   let timeTag = 'seconds'
 
-  let diff = (finishedAt.getTime() - startedAt.getTime()) / 1000
+  if (!isNaN(finishedAt.getTime()) || !isNaN(startedAt.getTime())) {
```
ebb-tide: so I think this statement will still be true if one is nan but one isn't.  and should be (!isNan(...) && !isNan(...)) and instead of or.

ebb-tide: But actually I suggest doing
```
if (isNaN(finishedAt.getTime()) || isNaN(startedAt.getTime())){
return '' 
}
// rest of logic
```

Palakp41: Will update!

Issue #13690, opened 2019-04-29
Cardinality metrics disagree with each other


e-dard:
`storage_series_file_series_total` disagrees with `storage_store_series_active`.

Ideally they should report the same cardinality value as the Series File and TSI index should be in sync.

I'm seeing dramatic differences in the values. My hunch is that the series file metric is still counting series that have been deleted, and the TSI index one is not.

benbjohnson:
Update: The tombstones are not being included in `SeriesIndex.Count()`. I'm trying to fix that but I'm working through issues in the unit test where I'm getting unexpected results.

Another issue is that `Store.monitorShardCardinality()` updates `storage_store_series_active` every `MonitorCardinalityInterval` (`10s` by default). However, the `storage_series_file_series_total` only seems to be updated on `Open()` so that number will lag behind. @e-dard Do we want this updated on an interval in a goroutine or can I update the value whenever the count changes?

e-dard:
@benbjohnson what are the tradeoffs of each approach? Given the TSI one (`storage_store_series_active`) is only updated and observed periodically, I wonder if we can do the same with the series file metric.

benbjohnson:
I'm not sure there's strong tradeoffs either way. It's probably better to keep it consistent with the other metric and throw it in a goroutine.

benbjohnson:
Fixed by #13770. Additional goroutine unnecessary as series are incrementally tracked here: 

- Add: https://github.com/influxdata/influxdb/blob/a181e60d70c02a3eed360a9414d90f07051f3665/tsdb/series_partition.go#L305
- Delete: https://github.com/influxdata/influxdb/blob/a181e60d70c02a3eed360a9414d90f07051f3665/tsdb/series_partition.go#L380

PR #13770, opened 2019-05-03
fix(tsdb): Fix series file count

benbjohnson:
Previously the series file did not include tombstones in the total count. This commit now includes tombstones in the count as well as fixes an issue where replayed tombstone records could exist but their underlying ID did not exist. This caused the count to become negative and with the count being `uint64` it caused the count to rollover to `math.Uint64Max`.

Related: https://github.com/influxdata/influxdb/issues/13690

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/tsdb/series_index.go
+++ b/tsdb/series_index.go
@@ -165,7 +165,11 @@ func (idx *SeriesIndex) GrowBy(delta int) {
 
 // Count returns the number of series in the index.
 func (idx *SeriesIndex) Count() uint64 {
-	return idx.OnDiskCount() + idx.InMemCount()
+	n := int64(idx.OnDiskCount()+idx.InMemCount()) - int64(len(idx.tombstones))
+	if n < 0 {
+		n = 0
+	}
+	return uint64(n)
 }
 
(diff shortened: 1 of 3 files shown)


tsdb/series_index.go
```diff
@@ -165,7 +165,11 @@ func (idx *SeriesIndex) GrowBy(delta int) {
 
 // Count returns the number of series in the index.
 func (idx *SeriesIndex) Count() uint64 {
-	return idx.OnDiskCount() + idx.InMemCount()
+	n := int64(idx.OnDiskCount()+idx.InMemCount()) - int64(len(idx.tombstones))
+	if n < 0 {
```
e-dard: This bothers me. How is this possible?

benbjohnson: It's not. We return a `uint64` though so if it does happen to cross into the negative threshold then we'll get stats at `math.UintMax`.

benbjohnson:
This is what causes https://github.com/influxdata/influxdb/issues/13690. We also need a goroutine to continually update it so I didn't mark it as `Fix #...`.

PR #13780, opened 2019-05-03
Fix/range error msg

jlapacik:
Closes #

The last two commits give a better error message when a range is either not given or not pushed down.

- [ ] Rebased/mergeable
- [ ] Tests pass


query/stdlib/influxdata/influxdb/operators.go
```diff
@@ -80,14 +80,31 @@ func (s *ReadRangePhysSpec) Copy() plan.ProcedureSpec {
...
-		return fmt.Errorf(`%s: results from "%s" must be bounded`, id, bucket)
+		beg, err := s.Bounds.Start.MarshalText()
+		if err != nil {
+			return err
+		}
+		end, err := s.Bounds.Stop.MarshalText()
+		if err != nil {
+			return err
+		}
+		return fmt.Errorf("empty time range queried from bucket %s; start: %s, stop: %s",
```
wolffcm: Why is this an error?  Couldn't we fulfill this request, even if it's likely a mistake by the user?

wolffcm: If storage cannot handle the empty request, that's a different story.

jlapacik:
@wolffcm Yes my first commit https://github.com/influxdata/influxdb/pull/13780/commits/cb0a4600a0eb5fcb3a359d36b0b197845fd58ad7 conflicts with changes already in master. So I figured it's a good time to merge master into flux-staging just so that it doesn't cause someone else trouble when they go to update flux-staging.

What do you think?

wolffcm:
Seems fine to merge master into flux-staging as long as you do it for both idpe and influxdb, but I would do it in separate PRs, using merge commits as usual.

query/stdlib/influxdata/influxdb/operators.go
jlapacik: Storage can handle an empty time range, but a query specifying an empty range is most likely an unintended user error. In that case maybe it's best to highlight it vs executing the query and putting some load on the storage engine just to return no data?

wolffcm: I can imagine uses for querying an empty range.  Maybe the user just wants to see if their query _would_ execute okay.  In general I don't like to second-guess the user's intention.

One could imagine creating a no-op source that generates no data in the future, as an optimization for things like `filter(fn: (r) => false)` as well as the empty range.

I know erroring on an empty range is the current behavior, so leaving it as-is would be okay too.

PR #13783, opened 2019-05-03
fix(ui): ensure map type variables can get selected key

### LINKED ISSUE #13749: Map variable type doesn't work
to repro, try to create a map variable type and use it in a dashboard. you cannot select any value.

Can we create an e2e test for this?

Kapture 2019-05-01 at 23 08 11

ischolten:
Closes #13749

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/dashboards/components/variablesControlBar/VariableDropdown.test.tsx
+++ b/ui/src/dashboards/components/variablesControlBar/VariableDropdown.test.tsx
@@ -1,0 +1,82 @@
+// Libraries
+import React from 'react'
+import {fireEvent} from 'react-testing-library'
+
+// Components
+import VariableDropdown from 'src/dashboards/components/variablesControlBar/VariableDropdown'
+
+// Utils
+import {renderWithRedux} from 'src/mockState'
+import {AppState} from 'src/types'
+
+const values = {
+  def: 'defbuck',
+  def2: 'defbuck',
+  foo: 'foobuck',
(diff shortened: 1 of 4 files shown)


ui/src/dashboards/selectors/index.ts
```diff
@@ -58,7 +58,7 @@ export const getVariableValuesForDropdown = (
       const selection = list.find(({value}) => value === selectedValue)
 
       return {
-        selectedKey: selection[0],
+        selectedKey: selection.name,
```
OfTheDelmer: wooops!

ui/src/dashboards/components/variablesControlBar/VariableDropdown.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+            },
+          },
+          order: ['03cbdc8a53a63000'],
+        },
+      },
+    },
+  }
+}
+
+describe('Dashboards.Components.VariablesControlBar.VariableDropdown', () => {
```
alexpaxton: Yay more tests

ui/src/dashboards/selectors/index.ts
alexpaxton: Does this need a guard?

ischolten: i can add a _.get

PR #13794, opened 2019-05-06
test(flux/e2e): auto-skip e2e test if no table gets read from bucket

affo:
This auto-skips flaky tests because the bucket doesn't contain input data.

- [ ] Rebased/mergeable
- [ ] Tests pass


affo:
This is still buggy, because sometimes `yield` doesn't return the `input-data` result. I think this is related to #1234. Once that got fixed, this should work.

nathanielc:
Unfortunately this didn't work, we still have flaky tests.

PR #13800, opened 2019-05-06
Generate more idiomatic Flux in query builder

### LINKED ISSUE #13775: Autogenerated Flux code does not use "aggregateWindow"
Use the query builder to create a Flux query and add at least one function.

For example click:

* telegraf - bucket
* disk - measurement
* used_percent - field
* mean - function

This code is generated

```
from(bucket: "telegraf")
  |> filter(fn: (r) => r._measurement == "cpu" or r._measurement == "disk")
  |> filter(fn: (r) => r._field == "used_percent")
  |> window(period: v.windowPeriod)
  |> mean()
  |> group(columns: ["_value", "_time", "_start", "_stop"], mode: "except")
  |> yield(name: "mean")
```

The above code can be simplified to use `aggregateWindow` like:

```
from(bucket: "telegraf")
  |> filter(fn: (r) => r._measurement == "cpu" or r._measurement == "disk")
  |> filter(fn: (r) => r._field == "used_percent")
  |> aggregateWindow(every: v.windowPeriod, fn: mean)
  |> yield(name: "mean")
```

As users will likely learn Flux best by example I think it is important to introduce good concepts in the Flux code we generate. Using `aggregateWindow` is something we want users to learn to do. This way they do not have to mange the groups so directly.

Previously only aggregate functions like `mean` could be used with `aggregateWindow` while selector functions like `max` could not. This is no longer the case and so I think we should push users to learn `aggregateWindow`.

chnn:
Closes #13775

Updates the Flux code generated by the query builder to use the `aggregateWindow` helper when possible. Queries previously generated like so:

```
from(bucket: "telegraf")
  |> filter(fn: (r) => r._measurement == "cpu" or r._measurement == "disk")
  |> filter(fn: (r) => r._field == "used_percent")
  |> window(period: v.windowPeriod)
  |> mean()
  |> group(columns: ["_value", "_time", "_start", "_stop"], mode: "except")
  |> yield(name: "mean")
```

Will now be generated as:

```
from(bucket: "telegraf")
  |> filter(fn: (r) => r._measurement == "cpu" or r._measurement == "disk")
  |> filter(fn: (r) => r._field == "used_percent")
  |> aggregateWindow(every: v.windowPeriod, fn: mean)
  |> yield(name: "mean")
```

### DIFF
--- a/ui/src/timeMachine/constants/queryBuilder.ts
+++ b/ui/src/timeMachine/constants/queryBuilder.ts
@@ -3,28 +3,75 @@ import {WINDOW_PERIOD, OPTION_NAME} from 'src/variables/constants'
 export interface QueryFn {
   name: string
   flux: string
-  aggregate: boolean
 }
 
 export const FUNCTIONS: QueryFn[] = [
-  {name: 'mean', flux: `|> mean()`, aggregate: true},
-  {name: 'median', flux: '|> toFloat()\n  |> median()', aggregate: true},
-  {name: 'max', flux: '|> max()', aggregate: true},
-  {name: 'min', flux: '|> min()', aggregate: true},
-  {name: 'sum', flux: '|> sum()', aggregate: true},
+  {
+    name: 'mean',
+    flux: `|> aggregateWindow(every: ${OPTION_NAME}.${WINDOW_PERIOD}, fn: mean)`,
+  },
+  {
+    name: 'median',
+    // TODO: https://github.com/influxdata/influxdb/issues/13806
+    flux: `|> window(period: ${OPTION_NAME}.${WINDOW_PERIOD})
+  |> toFloat()
+  |> median()
+  |> group(columns: ["_value", "_time", "_start", "_stop"], mode: "except")`,
+  },
+  {
+    name: 'max',
+    flux: `|> aggregateWindow(every: ${OPTION_NAME}.${WINDOW_PERIOD}, fn: max)`,
+  },
+  {
+    name: 'min',
+    flux: `|> aggregateWindow(every: ${OPTION_NAME}.${WINDOW_PERIOD}, fn: min)`,
+  },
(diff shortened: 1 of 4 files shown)


ui/src/timeMachine/constants/queryBuilder.ts
```diff
@@ -3,28 +3,77 @@ import {WINDOW_PERIOD, OPTION_NAME} from 'src/variables/constants'
...
-  {name: 'sum', flux: '|> sum()', aggregate: true},
+  {
+    name: 'mean',
+    flux: `|> aggregateWindow(every: ${OPTION_NAME}.${WINDOW_PERIOD}, fn: mean)`,
+  },
+  {
+    name: 'median',
+    // TODO: Update to use `aggregateWindow` once
+    // https://github.com/influxdata/flux/issues/1244 is resolved. The
+    // `toFloat` call should still be present in the updated version.
```
ebb-tide: maybe an issue instead of a todo here?

chnn: sure how about both, in case someone isn't aware of the issue when they are looking at the code

https://github.com/influxdata/influxdb/issues/13806

chnn:
> any idea how we could keep track of the number type issue nathaniel highlighted? It's certainly going to come up again.

Yeah not sure... maybe it makes sense to wait until the issue is reported by a user, so we have a more acute description of the problem. Or we could create an issue—my worry is that the issue would be too broad of a problem statement and unactionable.

nathanielc:
> any idea how we could keep track of the number type issue nathaniel highlighted? It's certainly going to come up again.

Eventually I expect this to be something that a Flux language server could expose. Not sure exactly the details of how that will all work, but this kind of thing is a type error which in many cases should be discoverable statically.

PR #13805, opened 2019-05-06
feat(task): Track queue duration

lyondhill:
Closes ##13792

Track the time it takes for a task to go from it's dueAt time until it actually starts

- [ ] Rebased/mergeable
- [x] Tests pass


task/backend/scheduler_metrics.go
```diff
@@ -59,6 +61,13 @@ func newSchedulerMetrics() *schedulerMetrics {
 			Name:      "claims_active",
 			Help:      "Total number of claims currently held.",
 		}),
+		queueDelta: prometheus.NewSummary(prometheus.SummaryOpts{
+			Namespace:  namespace,
+			Subsystem:  subsystem,
+			Name:       "run_queue_delta",
+			Help:       "The time it task from when a task is due until it is started.",
```
mark-rushakoff: `The duration in seconds between a run being due to start and actually starting.`

task/backend/scheduler_metrics.go
```diff
@@ -71,12 +80,14 @@ func (sm *schedulerMetrics) PrometheusCollectors() []prometheus.Collector {
 		sm.runsActive,
 		sm.claimsComplete,
 		sm.claimsActive,
+		sm.queueDelta,
 	}
 }
 
 // StartRun adjusts the metrics to indicate a run is in progress for the given task ID.
-func (sm *schedulerMetrics) StartRun(tid string) {
+func (sm *schedulerMetrics) StartRun(tid string, queueDelta float64) {
```
mark-rushakoff: Add godoc and maybe rename the new parameter to reflect that it is in seconds.

We do have tests for the scheduler metrics. I don't very well understand how prometheus histograms work, but it wouldn't hurt to add a test for it.

lyondhill: Ill update the godoc, and also look into the test. good eye :)

task/backend/scheduler_metrics.go
```diff
@@ -71,12 +80,15 @@ func (sm *schedulerMetrics) PrometheusCollectors() []prometheus.Collector {
...
 		sm.claimsComplete,
 		sm.claimsActive,
+		sm.queueDelta,
 	}
 }
 
 // StartRun adjusts the metrics to indicate a run is in progress for the given task ID.
-func (sm *schedulerMetrics) StartRun(tid string) {
+// We are also storing the delta time between when a run is due to start and actually starting.
+func (sm *schedulerMetrics) StartRun(tid string, queueDelta float64) {
```
goller: Rather than leak prometheus' types into this StartRun function would you take a time.Duration instead?  Then you could do the conversion as close to the prom library as possible.

task/backend/scheduler.go
```diff
@@ -51,6 +51,9 @@ type QueuedRun struct {
 	// The Unix timestamp (seconds since January 1, 1970 UTC) that will be set when a run a manually requested
 	RequestedAt int64
 
+	// The Unix timestamp representing when this run was due to run.
+	DueAt int64
```
goller: Could this be a time.Time rather than int64?  The majority of this patch is conversions of time.Time to this number

lyondhill: The rest of the time stamps in the QueuedRun are int64 unix time stamps. I assumed it would be better to have the model be comparable. This could be a time.Time. I don't really have a preference.

PR #13816, opened 2019-05-07
test(flux/e2e): export skip list

affo:
See https://github.com/influxdata/idpe/issues/3300

- [ ] Rebased/mergeable
- [ ] Tests pass


query/stdlib/testing/testing.go
```diff
@@ -1 +1,107 @@
 package testing
+
+var FluxEndToEndSkipList = map[string]string{
```
affo: Here I merged the 2 skip lists and tried to keep the most meaningful reasons.

jsternberg: Merged from where?

affo: From the one that I removed here: https://github.com/influxdata/idpe/pull/3434

jsternberg: I think it might be better if we just expose the skip list from influxdb and then in the enterprise product, we can just check both the influxdb skip list and the idpe specific-skip list.

affo: I'm doing it now

PR #13828, opened 2019-05-07
Extract "retry after" from query limit error response headers and reveal to user in notifications

ebb-tide:
Query limit notification should include cool down timer
Documentation on error headers :https://github.com/influxdata/influxdb/blob/master/http/swagger.yml#L1556-L1563


ebb-tide:
https://github.com/influxdata/influxdb/blob/master/http/swagger.yml#L1556-L1563

PR #13844, opened 2019-05-08
fix(http): do not discard non-json encoded errors when using `CheckError`

jsternberg:
If the error returned was normal text instead of JSON, the JSON parsing
would fail and the JSON error would be returned instead of the actual
error.

This has been modified so that when it fails to parse the text as JSON,
it will construct an error with the actual text and then wrap it using
the JSON error.

### DIFF
--- a/http/errors.go
+++ b/http/errors.go
@@ -1,12 +1,16 @@
 package http
 
 import (
+	"bytes"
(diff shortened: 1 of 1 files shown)


lukevmorris:
Thanks Jade

PR #13850, opened 2019-05-08
feat(tasks): add description field to tasks

### LINKED ISSUE #12386: no description field on tasks
it's a UI req. that each resource has a description field, the tasks swagger spec does not contain one. The parameter should be a string field which is optional.

AlirieGray:
Closes #12386

This PR adds a `Description` property to Tasks in the Swagger definition as well as in the API.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/task_service_test.go
+++ b/http/task_service_test.go
@@ -87,6 +87,7 @@ func TestTaskHandler_handleGetTasks(t *testing.T) {
 							{
 								ID:              1,
 								Name:            "task1",
+								Description:     "A little Task",
 								OrganizationID:  1,
 								Organization:    "test",
 								AuthorizationID: 0x100,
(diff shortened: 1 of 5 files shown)


http/task_service_test.go
```diff
@@ -137,7 +138,8 @@ func TestTaskHandler_handleGetTasks(t *testing.T) {
       },
       "id": "0000000000000001",
       "name": "task1",
-			"labels": [
+	  "description: "A little Task",
```
jademcgough: `"description` -> `"description"` will fix the test failure

task.go
```diff
@@ -133,6 +134,7 @@ type TaskService interface {
 // TaskCreate is the set of values to create a task.
 type TaskCreate struct {
 	Flux           string `json:"flux"`
+	Description    string `json:description,omitempty`
```
jademcgough: `json:"description,omitempty"`

Issue #13852, opened 2019-05-08
Avoid unnecessary Flux CSV parsing


chnn:
Currently, every time we fetch Flux CSV data we parse it into a `FluxTable[]` representation, but we don't always use it. In most situations we use the @influxdata/vis `Table` parsed representation instead. We should delay parsing CSV data into a `FluxTable[]` until absolutely necessary.

Parsing Flux CSV data is a significant bottleneck in the rendering of the UI and negatively impacts the UX of dashboards. Using the @influxdata/vis `Table` format exclusively would enable an effective implementation of https://github.com/influxdata/influxdb/issues/13849 as well.

PR #13853, opened 2019-05-08
fix(ui): improve single stat computation

### LINKED ISSUE #13824: Single stat complaining about non-numeric values
Details to reproduce coming shortly...

Screen Shot 2019-05-07 at 9 06 20 AM

chnn:
The method we used to compute a single stat / gauge value previously did account for missing data. If the latest value in a response was part of a numeric column but was null/NaN/not defined, the single stat computation would fail and a user would see an error message "Could not display single stat because your values are non-numeric".

This commit updates the single stat computation to find the latest *defined* numeric values.

If no latest valid numeric values are found, we will either:

- Display an error message if using the compuation within a single stat visualization
- Display nothing if using the computation within a within a line + single stat visualization (i.e. display the line vis only)

If multiple latest values are found, we make an arbitrary selection (same as previous behavior). The goal is to eventually expose UI elements to the user so they can make this selection themselves.

This commit also updates the single stat computation to use the @influxdata/vis `Table` format as an intermediate/parsed representation of a Flux CSV response. This unlocks the possibility for performance gains in our CSV parsing. See #13852.

Closes #13824


chnn:
Looks like this breaks a few of the single stats on the system dashboard template. Looking into it...

PR #13869, opened 2019-05-09
Extract and update templates

bthesorceror:


ui/src/templates/constants/defaultTemplates.ts
```diff
@@ -1,2965 +1,16 @@
...
-            {
-              id: '03afa77a9b582000',
-              type: 'cell',
-            },
-            {
-              id: '03afa77aa3582000',
-              type: 'cell',
-            },
-            {
-              id: '03afa77aa6182000
```
chnn: I think these were originally functions returning objects rather than just objects to guard against someone mutating them and causing subtle bugs.

We could preserve the existing behavior by importing the JSON files as text and returning `JSON.parse(theText)` from these functions... or we could not worry about it and assume no one will mutate them down the line. If we do the latter then I think we should update whatever uses these functions to just use the `SYSTEM_TEMPLATE` or whatever constant object directly.

PR #13875, opened 2019-05-09
Update members to use ResourceList instead of IndexList

### LINKED ISSUE #13543: Members list should be using ResourceList instead of IndexList
Members list uses index list right now. We should switch members list to use resource list to keep the code consistent.

Palakp41:
Closes #13543 

Updated members list to use resource list components
Renamed ResourceName to ResourceEditableName for the editable name resources

- [ ] Rebased/mergeable
- [ ] Tests pass


ui/src/clockface/components/resource_list/ResourceEditableName.tsx
```diff
@@ -0,0 +1,172 @@
...
+        >
+          <span className="icon pencil" />
+        </div>
+        {this.input}
+      </div>
+    )
+  }
+
+  private get input(): JSX.Element {
+    const {placeholder, inputTestID} = this.props
```
bthesorceror: Could we extract this to its own component? Usually a getter for a JSX element is usually a good indicator.

ui/src/members/components/MemberCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+          testID="task-card"
+          contextMenu={() => this.contextMenu}
+          name={() => <ResourceList.Name name={member.name} />}
+          metaData={() => [<>Role: {member.role}</>]}
+        />
+      </>
+    )
+  }
+
+  private get contextMenu(): JSX.Element {
```
bthesorceror: Same comment as above.

PR #13895, opened 2019-05-13
Bonitoo cypress tokens

karel-rehor:
Closes #

- [ ] Rebased/mergeable
- [ ] Tests pass


ui/cypress/e2e/tokens.test.ts
```diff
@@ -0,0 +1,388 @@
...
+                    status: resp.body.status === 'active',
+                    id: resp.body.id,
+                  })
+                })
+              })
+              .then(() => {
+                //sanity check that data is seeded
+                cy.get<Organization>('@org').then(({id}) => {
+                  cy.request('api/v2/authorizations?orgID=' + id).then(resp => {
+                    expect(resp.body.authorizations).to.have.length(4)
```
121watts: I don't think this sanity check is necessary.  This would also eliminate the need to chain all 3 of these create calls in nested `then` blocks.

karel-rehor: The check has been removed and the seeding block replaced with a loop 
```javascript
    testTokens.forEach(token => {
          cy.createToken(
            token.id,
            token.description,
            token.status,
            token.permissions
          ).then(resp => {
            expect(resp.body).to.exist
            authData.push({
              description: resp.body.description,
              status: resp.body.status === 'active',
              id: resp.body.id,
            })
          })
        })
```
Also the `cy.visit` call into the UI has been placed below this loop.  So it runs stably now headless and in Chrome.

PR #13914, opened 2019-05-14
Extract templates to external module

bthesorceror:
Moves template json files to external dependency.


ui/package.json
```diff
@@ -141,6 +141,8 @@
   "dependencies": {
     "@influxdata/clockface": "0.0.13",
     "@influxdata/influx": "0.3.4",
+    "@influxdata/influxdb-template": "git+ssh://git@github.com:influxdata/influxdb-templates.git",
```
chnn: should this line exist? Looks like it's being imported from the pluralized version instead (line 145)

ui/package.json
```diff
@@ -128,7 +128,7 @@
     "junit-viewer": "^4.11.1",
     "mocha": "^5.2.0",
     "mocha-junit-reporter": "^1.18.0",
-    "parcel": "^1.11.0",
+    "parcel": "^1.12.3",
```
chnn: we'll have to fix HMR, but that seems OK. I spiked on a solution a few weeks back.

ui/package.json
bthesorceror: yeah missed that, thanks

PR #13915, opened 2019-05-14
Update Labels list to use ResourceList instead of IndexList

### LINKED ISSUE #13548: Labels list should be using ResourceList instead of IndexList
Labels list uses index list right now. We should switch labels list to use resource list to keep the code consistent.

Palakp41:
Closes #13548 

- [ ] Rebased/mergeable
- [ ] Tests pass


ui/src/labels/components/LabelList.tsx
```diff
@@ -56,22 +56,25 @@ export default class LabelList extends PureComponent<Props, State> {
     const {sortKey, sortDirection, onClickColumn} = this.props
     return (
       <>
-        <IndexList>
-          <IndexList.Header>
-            <IndexList.HeaderCell
+        <ResourceList>
+          <ResourceList.Header>
+            <ResourceList.Sorter
+              name="Name"
```
ischolten: could this.headerKeys[0] be used here? rather than the string "Name"?

Palakp41: Yes, it can. I updated it!

PR #13962, opened 2019-05-17
chore(swagger): update swagger to include operationId

docmerlin:
This adds operationId to the swagger template.

For the client generation in golang, operationId defines the function names.


goller:
I totally want to see what you did to to all this 😺

Looks good but let’s talk to Brandon to see if this changes the java generation.

PR #13964, opened 2019-05-17
Add honeybadger overrides for cloud

bthesorceror:

### DIFF
--- a/ui/src/cloud/utils/overrides.ts
+++ b/ui/src/cloud/utils/overrides.ts
@@ -1,0 +1,31 @@
+import _ from 'lodash'
+import HoneyBadger from 'honeybadger-js'
+import {GIT_SHA} from 'src/shared/constants'
+
+if (process.env.CLOUD === 'true') {
+  HoneyBadger.configure({
+    apiKey: process.env.HONEYBADGER_KEY,
+    revision: GIT_SHA,
+    environment: process.env.HONEYBADGER_ENV,
+  })
+
+  const override = () => {
+    if (!window || !window.console) {
+      return
+    }
+
+    const logError = console.error
+
+    console.error = function(...args) {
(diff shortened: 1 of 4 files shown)


ui/src/cloud/utils/overrides.ts
```diff
@@ -0,0 +1,33 @@
+import _ from 'lodash'
+import HoneyBadger from 'honeybadger-js'
+import {GIT_SHA} from 'src/shared/constants'
+
+if (process.env.CLOUD === 'true') {
```
OfTheDelmer: Configure accepts a disabled true/false key where this value could be used.

ui/src/cloud/utils/overrides.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  HoneyBadger.configure({
+    apiKey: process.env.HONEYBADGER_KEY,
+    revision: GIT_SHA,
+    environment: process.env.HONEYBADGER_ENV,
```
OfTheDelmer: This accepts a context where we could add more helpful information like user_id. Or in the honeybadger client we could set context using the honeybadger.resetContext or .addContext

ui/src/cloud/utils/overrides.ts
```diff
@@ -0,0 +1,33 @@
...
+  })
+
+  const override = () => {
+    if (!window || !window.console) {
+      return
+    }
+
+    const logError = console.error
+
+    console.error = function (...args) {
```
OfTheDelmer: Nice strategy

ui/src/cloud/utils/overrides.ts
bthesorceror: I would prefer we didn't even hit configure when its not running in cloud.

PR #14010, opened 2019-05-29
feat: add "none" selection to query builder function selector aggreation

### LINKED ISSUE #13806: Use `aggregateWindow` when generating a Flux query with a `median` call
Depends on https://github.com/influxdata/flux/issues/1244

The query builder currently builds `median` queries like so:

```
from(bucket: "telegraf")
  |> filter(fn: (r) => r._measurement == "mem")
  |> filter(fn: (r) => r._field == "active")
  |> window(period: v.windowPeriod)
  |> toFloat()
  |> median()
  |> group(columns: ["_value", "_time", "_start", "_stop"], mode: "except")
  |> yield(name: "median")
```

When possible, we should update the builder to generate a query like this:

```
from(bucket: "telegraf")
  |> filter(fn: (r) => r._measurement == "mem")
  |> filter(fn: (r) => r._field == "active")
  |> toFloat()
  |> aggregateWindow(every: v.windowPeriod, fn: median)
  |> yield(name: "median")
```

See https://github.com/influxdata/influxdb/pull/13800, https://github.com/influxdata/influxdb/issues/13775, [discussion](https://github.com/influxdata/influxdb/pull/13800/files/d6f371a7d6c452f8f7ce43bc05f77fe76c23ec1d#r281347551).

### LINKED ISSUE #14000: Story breakdown for Alerts and Notification

121watts:
Closes #14000
Closes #13806

### DIFF
--- a/ui/src/timeMachine/constants/queryBuilder.ts
+++ b/ui/src/timeMachine/constants/queryBuilder.ts
@@ -7,7 +7,7 @@ export interface QueryFn {
 }
 
 export const genFlux = (func: string, period?: string) => {
-  if (period === 'none') {
+  if (period === AGG_WINDOW_NONE) {
     return `|> ${func}()`
   }
 
--- a/ui/src/timeMachine/constants/queryBuilder.ts
+++ b/ui/src/timeMachine/constants/queryBuilder.ts
@@ -6,77 +6,103 @@ export interface QueryFn {
   flux: (period?: string) => string
 }
 
+export const genFlux = (func: string, period?: string) => {
+  if (period === 'none') {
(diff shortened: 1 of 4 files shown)


ui/src/timeMachine/constants/queryBuilder.ts
```diff
@@ -6,77 +6,103 @@ export interface QueryFn {
   flux: (period?: string) => string
 }
 
+export const genFlux = (func: string, period?: string) => {
+  if (period === 'none') {
```
ischolten: should this be checking against the AGG_WINDOW_NONE rather than the string `none`?

PR #14016, opened 2019-05-29
fix(storage/reads): fix the storage tables to work correctly with multiple transformations

### LINKED ISSUE #12891: Flaky: Flux end to end tests

jsternberg:
The storage table reader will now work correctly when there are multiple
outputs. The table interface now implements the new table and column
reader interfaces and works properly with `execute.CopyTable`. The
source uses `execute.CopyTable` to buffer the table in memory when there
are multiple output transformations.

Fixes #12891.

### DIFF
--- a/query/stdlib/influxdata/influxdb/source.go
+++ b/query/stdlib/influxdata/influxdb/source.go
@@ -55,17 +55,7 @@ func (s *Source) Metadata() flux.Metadata {
 
 func (s *Source) processTables(ctx context.Context, tables TableIterator, watermark execute.Time) error {
 	err := tables.Do(func(tbl flux.Table) error {
-		for _, t := range s.ts {
-			if err := t.Process(s.id, tbl); err != nil {
-				return err
-			}
-			//TODO(nathanielc): Also add mechanism to send UpdateProcessingTime calls, when no data is arriving.
-			// This is probably not needed for this source, but other sources should do so.
-			if err := t.UpdateProcessingTime(s.id, execute.Now()); err != nil {
-				return err
-			}
-		}
-		return nil
+		return s.processTable(ctx, tbl)
 	})
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


query/stdlib/influxdata/influxdb/source.go
```diff
@@ -84,6 +74,30 @@ func (s *Source) processTables(ctx context.Context, tables TableIterator, waterm
 	return nil
 }
 
+func (s *Source) processTable(ctx context.Context, tbl flux.Table) error {
+	if len(s.ts) == 0 {
+		tbl.Done()
+		return nil
+	} else if len(s.ts) == 1 {
+		return s.ts[0].Process(s.id, tbl)
```
jlapacik: I see. So if there are multiple downstream transformations, for each table we send copies downstream and then release our specific instance. But if there's only one downstream transformation, there's no need to copy the table, we just send it downstream and it is the job of that transformation to release it. Correct?

jsternberg: Yes. That is correct.

storage/reads/table.gen.go.tmpl
```diff
@@ -73,36 +72,37 @@ func (t *{{.name}}Table) Do(f func(flux.ColReader) error) error {
...
-		t.err = f(t)
-		for !t.isCancelled() && t.err == nil && t.advance() {
-			t.err = f(t)
+	for !t.isCancelled() && t.err == nil  {
+		cr := t.advance()
+		if cr == nil {
+			break
 		}
+		t.err = f(cr)
+		cr.Release()
```
jlapacik: Is it possible for Release to be called before another transformation Retains it?

jsternberg: I think the contract for `Do` says that as soon as the function ends it goes out of scope so it shouldn't be possible to have anything like the `consecutiveTransport` as that would break that contract (unless the transport retained it itself).

So I don't think this should cause a race condition unless the passed in function is breaking the contract.

PR #14029, opened 2019-05-30
fix(tasks): adds check for inactive task in ForceRun and Retry

### LINKED ISSUE #13771: Running a task returns incorrect error message
When a task is inactive and run task button is clicked, the error comes back with message `task not claimed`. Error should return `task not active`.

AlirieGray:
Closes #13771

This PR adds a check for if a task is inactive to the `ForceRun` and `Retry` functions, and returns an error if the task is inactive.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/authorizer/task.go
+++ b/authorizer/task.go
@@ -2,14 +2,15 @@ package authorizer
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	"github.com/influxdata/flux"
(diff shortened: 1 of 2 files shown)


AlirieGray:
This change does not take into account the possibility that a use may want to manually run a task that is inactive (without setting it to active). I will make an issue/RFC to address possibly adding that functionality in another PR

PR #14044, opened 2019-06-03
fix(promql): synch generated code

### LINKED ISSUE #13955: promql: running go generate will modify the file and fail CI
Using `go generate` should produce the file that is committed. This appears to not be true for the `query/promql` package. Fix it so the modifications made will remain permanent if regenerated.

affo:
Closes #13955 

The change was not detected by simply running `make`.

I had to `make clean` and then `make`. Is this the expected behavior?

- [ ] Rebased/mergeable
- [ ] Tests pass


jsternberg:
The `make clean` followed by `make` is expected behavior.

query/promql/promql.go
```diff
@@ -3,8 +3,6 @@
 // Package promql implements a promql parser to build flux query specifications from promql.
 package promql
 
-//lint:file-ignore SA6001 Ignore all unused code, it's generated
```
jsternberg: I think you need to find some way to add this lint line.

affo: Thank you

query/promql/promql.go
```diff
@@ -23,42 +21,44 @@ import (
 
 // DO NOT EDIT: This file is auto generated by the pigeon PEG parser generator.
 
+//lint:file-ignore SA6001 Ignore all unused code, it's generated
```
jsternberg: Wow it did not occur to me that you could just do this but it's so obvious it hurts me to see that I didn't realize it.

PR #14048, opened 2019-06-03
fix(query/stdlib): remove skips for tests that were flaky from table.Do

jsternberg:


query/stdlib/testing/testing.go
```diff
@@ -102,8 +80,6 @@ var FluxEndToEndSkipList = map[string]string{
 	"rowfn_with_import":           "imported libraries are not visible in user-defined functions (https://github.com/influxdata/flux/issues/1000)",
 	"string_trim":                 "imported libraries are not visible in user-defined functions (https://github.com/influxdata/flux/issues/1000)",
 
-	"window_group_mean_ungroup": "window trigger optimization modifies sort order of its output tables (https://github.com/influxdata/flux/issues/1067)",
```
jlapacik: Why was this test fixed with the table.Do changes? From what I remember this test was skipped due to an unrelated issue.

jsternberg: It passed with `-count=10` which was my only criteria. I'll take another look at this one.

jsternberg: I don't know why it's not failing, but I'm going to add it back in and we'll review this again later.

PR #14052, opened 2019-06-03
Add x and y column pickers to graph types

### LINKED ISSUE #14024: Allow user to select y-axis Column to plot in the vis library
Currently the UI library looks for the `_value` column and fails if no `_value` column exists.

ebb-tide:
Closes #14024


ui/src/timeMachine/components/view_options/HeatmapOptions.tsx
```diff
@@ -100,34 +89,19 @@ const HeatmapOptions: FunctionComponent<Props> = props => {
...
-          titleText="None"
-        >
-          {props.numericColumns.map(columnName => (
-            <Dropdown.Item id={columnName} key={columnName} value={columnName}>
-              {columnName}
-            </Dropdown.Item>
-          ))}
-        </Dropdown>
-      </Form.Element>
+      <ColumnSelector
```
chnn: much better :)

ui/src/shared/components/ScatterContainer.tsx
```diff
@@ -75,15 +86,17 @@ const ScatterContainer: FunctionComponent<Props> = ({
         onSetFillColumns(filteredGroupKeys)
       }
     }
+    if (!xColumn) {
```
chnn: Is this to handle existing `XYViewProperties` that don't have an `xColumn` setting yet? Wowza I really wish we had the ability to migrate data...

I think if that is the case, there's no need to be setting the `xColumn` on the view properties by dispatching an effect, since those view properties will never be persisted. I think we can keep the x column resolution a purely render-path concern by declaring the `xColumn` could outside the effect like so:

```
const xColumn = storedXColumn || chooseXColumn(table)
```

It might help to extract that logic to a helper function and document it with a comment, so that we know we are dealing with an issue evolving data (like https://github.com/influxdata/influxdb/blob/e4772f06b238065e1bc25bc642042da73e926c4b/ui/src/shared/utils/vis.ts#L22-L32).

ui/src/shared/components/XYContainer.tsx
```diff
@@ -49,15 +62,24 @@ const XYContainer: FunctionComponent<Props> = ({
...
 }) => {
   const {table, fluxGroupKeyUnion} = useMemo(
     () => fromFlux(files.join('\n\n')),
     [files]
   )
 
-  // Eventually these will be configurable in the line graph options UI
-  const xColumn = chooseXColumn(table)
-  const yColumn = chooseYColumn(table)
+  useEffect(() => {
```
chnn: (same comment about moving this resolution outside of an effect)

ui/src/shared/components/ScatterContainer.tsx
```diff
@@ -50,9 +57,13 @@ const ScatterContainer: FunctionComponent<Props> = ({
...
+    xColumn,
+    yColumn,
   },
   onSetFillColumns,
   onSetSymbolColumns,
+  onSetXColumn,
+  onSetYColumn,
 }) => {
   useEffect(() => {
     if (fluxGroupKeyUnion && (!storedSymbol || !storedFill)) {
```
chnn: What is an instance where there is no `storedSymbol` / `storedFill` for a scatter view? I think this should never happen... or if it is happening, then the fix should be elsewhere so that we can ensure it never happens rather than programming defensively here. Can we ensure that the `storedSymbol` and `storedFill` settings are valid when the view is created (in the VEO)?

I think this `useEffect` introduces two-way data flow (component accepting data and emitting effects to change that data) into a component that should only observe one way data flow (component only accepts data). It feels like we are fighting React a lot here.

ui/src/shared/components/ScatterContainer.tsx
```diff
@@ -75,15 +86,17 @@ const ScatterContainer: FunctionComponent<Props> = ({
...
+    if (!yColumn) {
+      onSetYColumn(chooseYColumn(table))
+    }
   })
 
   const fillColumns = storedFill || []
   const symbolColumns = storedSymbol || []
 
-  // TODO: allow xcolumn and ycolumn to be user selectable
-  const xColumn = chooseXColumn(table)
```
chnn: I think we need to add in some guards to make sure the `xColumn` selection is valid. Something like this:

https://github.com/influxdata/influxdb/blob/e4772f06b238065e1bc25bc642042da73e926c4b/ui/src/shared/components/HeatmapContainer.tsx#L62-L67

ui/src/shared/components/XYContainer.tsx
```diff
@@ -49,15 +62,24 @@ const XYContainer: FunctionComponent<Props> = ({
...
+    if (!xColumn) {
+      xColumn = chooseXColumn(table)
+      onSetXColumn(xColumn)
+    }
+    if (!yColumn) {
+      yColumn = chooseYColumn(table)
+      onSetYColumn(yColumn)
+    }
+  })
```
chnn: (same comment about ensuring the `xColumn` and `yColumn` selections actually exist in the `table`)

dashboard.go
```diff
@@ -591,6 +591,8 @@ type LinePlusSingleStatProperties struct {
 	DecimalPlaces     DecimalPlaces    `json:"decimalPlaces"`
 	Note              string           `json:"note"`
 	ShowNoteWhenEmpty bool             `json:"showNoteWhenEmpty"`
+	XColumn           string           `json:"xColumn"`
```
chnn: Seems like we might need to feature flag the UI parts of the this feature, so that we can ensure these API changes go out first.

ui/src/timeMachine/selectors/index.ts
```diff
@@ -107,59 +106,81 @@ const getYColumnSelectionMemoized = memoizeOne(selectXYColumn)
...
 
-    return []
+  return getYColumnSelectionMemoized(
+    validYColumns,
+    preference,
+    defaultSelection
+  )
+}
+
+const getColumnSelection = (
```
chnn:
```suggestion
const getGroupColumnsSelection = (
```
maybe?

ui/src/shared/components/XYContainer.tsx
```diff
@@ -73,7 +66,14 @@ const XYContainer: FunctionComponent<Props> = ({
     storedYDomain,
     columnKeys.includes(yColumn) ? table.getColumn(yColumn, 'number') : []
   )
-  if (!xColumn || !yColumn) {
+
+  const isValidView =
```
chnn: I tested this out with some line graphs that I had created before this change. It looks like this will invalidate all of them, since they don't have stored `xColumn` and `yColumn` settings:

Screen Shot 2019-06-04 at 2 27 33 PM

To fix this I think that we could fall back to using `chooseXColumn(table)` if a `xColumn` doesn't exist on the view properties, and similarly for the `yColumn`.

```
const xColumn = storedXColumn || chooseXColumn(table)
const yColumn = storedYColumn || chooseYColumn(table)
```

chnn: We might also have to do this for the scatterplot too.

ebb-tide: thanks!

PR #14059, opened 2019-06-04
Enable formatting line graph y ticks with binary prefix

### LINKED ISSUE #14051: Add tick formatting presents to line graph options
Previously we had a "Y-Value's Format" option for line graphs that was removed in the transition to the vis library. We should add it back, using the same fields in the stored `XYView` to persist the setting.

Screen Shot 2019-04-29 at 3 55 19 PM

chnn:
Closes #14051 

bytes

When formatting ticks with a binary unit prefix, the generated tick intervals aren't super nice. I've created https://github.com/influxdata/vis/issues/79 and https://github.com/influxdata/vis/issues/80 to track this issue.

### DIFF
--- a/ui/src/timeMachine/components/view_options/YAxisBase.tsx
+++ b/ui/src/timeMachine/components/view_options/YAxisBase.tsx
@@ -23,34 +23,34 @@ class YAxisBase extends PureComponent<Props> {
 
     return (
       <Grid.Column widthXS={Columns.Twelve}>
-        <Form.Element label="Y-Value's Format">
+        <Form.Element label="Y-Value Unit Prefix">
           <Radio shape={ButtonShape.StretchToFit}>
             <Radio.Button
               id="y-values-format-tab--raw"
               value=""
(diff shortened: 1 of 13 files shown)


ui/src/timeMachine/components/view_options/YAxisBase.tsx
```diff
@@ -23,34 +23,34 @@ class YAxisBase extends PureComponent<Props> {
...
             >
-              Raw
+              None
             </Radio.Button>
             <Radio.Button
               id="y-values-format-tab--kmb"
               value={BASE_10}
               active={base === BASE_10}
-              titleText="Thousand / Million / Billion"
+              titleText="Format values using a SI unit prefix"
```
alexpaxton: What do you think of spelling out the full name of `SI` in the `titleText`?

chnn: Good point! I updated it.

Issue #12963, opened 2019-03-28
scraper: add scraper interval option


zhulongcheng:
https://github.com/influxdata/influxdb/blob/a64c4fd138ef11336de84b4bd42bc0f49b7d8d1d/cmd/influxd/launcher/launcher.go#L587-L591

Currently,  scraper scheduler interval hard code to `10*time.Second`, and all scrapers share the same interval.

Should each scraper has an interval option？

cc @russorat

russorat:
@zhulongcheng yes they should. thanks for reporting!

zhulongcheng:
don't close

NF1198:
This would be a very useful feature. +1

russorat:
@NF1198 have you tired configuring a Flux Task to do the scraping using the https://docs.influxdata.com/influxdb/v2.0/reference/flux/stdlib/experimental/prometheus/scrape/ function?

scrobbleme:
Really needed +1

> @NF1198 have you tired configuring a Flux Task to do the scraping using the https://docs.influxdata.com/influxdb/v2.0/reference/flux/stdlib/experimental/prometheus/scrape/ function?

Tried this, but wasn't able to get it working...

Nevertheless, scrapers are just a very very easy to setup and use... only an interval option is missing

d0rkster:
Using a scraper would be useful if the interval could be set. My data takes longer than 10 seconds to GET. And if your data doesn't change or update this often it's extra data not needed and workloads used that could be moved elsewhere.

alehmann76:
Would be a very useful feature!
+1

niyantaz:
Are there any updates? Looking for methods to set scraping-interval per scrapper

ebc-conscia:
Would be super useful to be able to set intervals, this would allow us to create Prometheus API endpoints in front of everything, even old PDUs, terminals etc. Just the 10-second interval can be a bit much for these dated types of equipment and at the same time it also saves processing time, reducing hardware requirements for the Influx server.

peterpeerdeman:
+1 for the interval setting on scrapers. I've tried the feature and removed these because 20sec interval was too much for this specific case. Without the interval I won't be using this feature at all. I've migrated to a task now and that works, for anyone (@scrobbleme ) interested in the workaround for this issue:

1. create notebook

2. build a query like:
```
import "experimental/prometheus"

prometheus.scrape(url: "http://192.168.x.x:9092/metrics")

|> to(bucket: "bucketname")
[...]

PR #14070, opened 2019-06-06
feat(http):  add scraper interval and timeout

zhulongcheng:
Updates #12963

add scraper interval and timeout options, so scrapers can be scheduled base on customized interval and timeout, not global options.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/scraper.go
+++ b/scraper.go
@@ -2,6 +2,7 @@ package influxdb
 
 import (
 	"context"
+	"github.com/influxdata/influxdb/task/options"
 )
 
 // ErrScraperTargetNotFound is the error msg for a missing scraper target.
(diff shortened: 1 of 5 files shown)


scraper.go
```diff
@@ -16,14 +17,21 @@ const (
 	OpUpdateTarget  = "UpdateTarget"
 )
 
+var (
+	DefaultScraperInterval = *options.MustParseDuration("10s")
+	DefaultScraperTimeout  = *options.MustParseDuration("30s")
```
zhulongcheng: the default value base on previous usage:
https://github.com/influxdata/influxdb/blob/a64c4fd138ef11336de84b4bd42bc0f49b7d8d1d/cmd/influxd/launcher/launcher.go#L587-L591

kelwang:
hey @zhulongcheng, thanks for this implementation. Quick question, looks like the interval won't really working, unless we implement this in the scheduler right?

zhulongcheng:
> hey @zhulongcheng, thanks for this implementation. Quick question, looks like the interval won't really working, unless we implement this in the scheduler right?

Yes. 
This PR added interval and timeout options, and the next step is to apply the options to the scheduler.

zhulongcheng:
Closes #12963 -> Updates #12963

rpatel3001:
looks like this is stuck in limbo, still a useful feature that I would appreciate.

Issue #13404, opened 2019-04-15
Make an OSS bolt data migrator for schema changes


kelwang:
Derived from https://github.com/influxdata/influxdb/pull/13286

1. Support convert all data in bolt storage in case any of schema change.
2. Give user an option to backup old data. In case user would like downgrade.

sub ticket of https://github.com/influxdata/idpe/issues/3649

PR #14073, opened 2019-06-06
Data migrator Prototype

kelwang:
Closes #https://github.com/influxdata/influxdb/issues/13404

The prototype of data migrator. will be able upgrade and downgrade.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/migrator.go
+++ b/migrator.go
@@ -1,0 +1,100 @@
+package influxdb
+
+import (
+	"context"
+)
+
+// MigrationManager is the manager of migrators.
+type MigrationManager interface {
+	Register(ms ...Migrator)
+	DryRun(ctx context.Context, errCh chan<- error, done chan<- struct{})
+	Migrate(ctx context.Context, up bool) (err error)
+}
+
+type migrationManager struct {
+	// Limit states how many ops per transaction
+	Limit int
+	MigratorService
+	Migrators []Migrator
+}
+
+// NewMigrationManager create a new NewMigrationManager.
+func NewMigrationManager(svc MigratorService, limit int) MigrationManager {
+	return &migrationManager{
+		Limit:           limit,
+		MigratorService: svc,
+		Migrators:       make([]Migrator, 0),
+	}
+}
+
+// Register add migrator to the manager.
+func (mg *migrationManager) Register(ms ...Migrator) {
+	mg.Migrators = append(mg.Migrators, ms...)
+}
+
+// DryRun will perform the dry run, and stream out error.
+func (mg *migrationManager) DryRun(ctx context.Context, errCh chan<- error, done chan<- struct{}) {
+	for _, m := range mg.Migrators {
+		if !m.NeedMigration() {
+			continue
+		}
+		mg.MigratorService.DryRun(ctx, m, true, errCh)
+	}
+	// close it
+	close(done)
+}
+
+// Migrate will perform the migration, store the value.
+func (mg *migrationManager) Migrate(ctx context.Context, up bool) (err error) {
+	for _, m := range mg.Migrators {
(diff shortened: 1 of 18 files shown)


migrator.go
```diff
@@ -0,0 +1,84 @@
...
+	NeedMigration() bool
+	// Up convert the old version to the new version.
+	Up(src []byte) (dst []byte, err error)
+	// Down convert the new version to the old version.
+	Down(src []byte) (dst []byte, err error)
+}
+
+// MigragorService is the service to fetch and store bytes.
+type MigragorService interface {
+	ScanBytes(ctx context.Context, bucket []byte) (data map[string][]byte, err error)
```
lyondhill: Any attempt to read all key's and values from an entire bucket will cause the system to run out of memory.

migrator.go
```diff
@@ -0,0 +1,84 @@
...
+	errCh <- nil
+}
+
+// Migrator is the interface of schema migration.
+type Migrator interface {
+	MigragorService
+	// Bucket is the bucket of the data stored.
+	Bucket() []byte
+	// NeedMigration will return whether this migrator is needed at runtime.
+	NeedMigration() bool
```
lyondhill: Migrations need to be able to execute in order. If multiple migrations are in a single update to the binary we can't run them in random order.

migrator.go
```diff
@@ -0,0 +1,84 @@
...
+// MigrationManager is the manager of migrators.
+type MigrationManager struct {
+	Migrators []Migrator
+}
+
+// Migrate will perform the migration.
+func (mg MigrationManager) Migrate(ctx context.Context, Up bool, errCh chan<- error) {
+	for _, m := range mg.Migrators {
+		if !m.NeedMigration() {
+			return
```
lyondhill: Return here would mean the first time a migrator on the list of migrators didn't need to migrate the rest of the list would never be executed.

migrator.go
```diff
@@ -0,0 +1,84 @@
...
+
+import "context"
+
+// MigrationManager is the manager of migrators.
+type MigrationManager struct {
+	Migrators []Migrator
+}
+
+// Migrate will perform the migration.
+func (mg MigrationManager) Migrate(ctx context.Context, Up bool, errCh chan<- error) {
```
lyondhill: If we are going to use a channel its best to have the `Migrate` function return a channel. `Migrate(ctx context.Context, up bool) (chan <- error) `. But if this isn't being executed in a separate go routine there is no reason to use channels.

lyondhill: Also Up -> up. No reason to capitalize a argument

migrator.go
```diff
@@ -0,0 +1,84 @@
...
+			}
+		}
+	}
+	// close it
+	errCh <- nil
+}
+
+// Migrator is the interface of schema migration.
+type Migrator interface {
+	MigragorService
```
lyondhill: So if I understand correctly all migrations have to implement the migration service? This introduces a lot of boilerplate on each request and gives no advantage as far as i can see.

migrator.go
```diff
@@ -0,0 +1,84 @@
+package influxdb
+
+import "context"
+
+// MigrationManager is the manager of migrators.
+type MigrationManager struct {
```
lyondhill: I dont think we will want more then one MigrationManger per execution of the binary. I would make this work similar to how http deals with adding new handlers (https://golang.org/src/net/http/server.go?s=74743:74787#L2424). I would even go as far to say that it would be best to make the migration manager private and expose a `Register(migrator Migrator)` and a `Migrate` function

migrator.go
kelwang: Good catch

PR #14078, opened 2019-06-06
perf: dashboard cell and view rendering

### LINKED ISSUE #14049: Dashboard perf - only load the cells in view in a dashboard
similar to 1.x, we should only load the data in cells that are in view in the browser.

121watts:
Closes #14049 

Used the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to determine if a cell is coming into the viewport.  If a cell is not in view do not render the contents or fetch any data.

Kapture 2019-06-06 at 14 10 35

### DIFF
--- a/ui/src/shared/components/RefreshingView.tsx
+++ b/ui/src/shared/components/RefreshingView.tsx
@@ -24,7 +24,6 @@ import {SpinnerContainer, TechnoSpinner} from '@influxdata/clockface'
 
 interface OwnProps {
   timeRange: TimeRange
-  inView: boolean
   manualRefresh: number
   properties: QueryViewProperties
   dashboardID: string
@@ -62,7 +61,7 @@ class RefreshingView extends PureComponent<Props, State> {
   }
 
   public render() {
-    const {inView, properties, manualRefresh, variablesStatus} = this.props
+    const {properties, manualRefresh, variablesStatus} = this.props
     const {submitToken} = this.state
 
     return (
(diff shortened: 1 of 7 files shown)


ui/src/shared/components/RefreshingView.tsx
```diff
@@ -62,7 +61,7 @@ class RefreshingView extends PureComponent<Props, State> {
   }
```
ebb-tide: line 44 references `inView` can be deleted

ui/src/shared/components/cells/Cell.tsx
```diff
@@ -66,8 +93,12 @@ class CellComponent extends Component<Props> {
...
           />
         )}
-        <div className="cell--view" data-testid="cell--view-empty">
-          {this.view}
+        <div
+          className="cell--view"
+          data-testid="cell--view-empty"
+          ref={this.cellRef}
+        >
+          {inView && this.view}
```
ebb-tide: instead of returning null here when inView===false, should we be rendering Spinner Container? this.view returns spinnercontainer if viewsstatus is done, maybe viewsstatus could be loading if inView is false.

ui/src/shared/components/cells/Cells.tsx
```diff
@@ -38,17 +32,15 @@ interface Props {
 }
 
 interface State {
-  rowHeight: number
+  mounted: boolean
```
ebb-tide: mounted !== animated?

PR #14087, opened 2019-06-08
Separate input element from date-picker

### LINKED ISSUE #13829: pressing enter while manually selecting a time range removes the calendar
Kapture 2019-05-07 at 11 55 26

ebb-tide:
Closes #13829 

Add separate input element to handle text based inputs to the date picker. 
If text input is invalid do not update the timeRange, but continue displaying in text input. 
When opening date picker for the first time, select time range that was already selected in Time Range Dropdown. So that if Past 5m were selected -> lower= now-5m and upper = now.

### DIFF
--- a/ui/src/shared/components/dateRangePicker/DatePicker.tsx
+++ b/ui/src/shared/components/dateRangePicker/DatePicker.tsx
@@ -4,14 +4,14 @@ import ReactDatePicker from 'react-datepicker'
 import moment from 'moment'
 
 // Components
-import {Form, Input, Grid} from '@influxdata/clockface'
+import {Input, Grid, ComponentStatus} from '@influxdata/clockface'
 
 // Styles
 import 'react-datepicker/dist/react-datepicker.css'
 
 // Types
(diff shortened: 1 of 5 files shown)


ui/src/shared/components/dateRangePicker/DatePicker.tsx
```diff
@@ -1,35 +1,56 @@
...
 // Components
-import {Form, Input, Grid} from '@influxdata/clockface'
+import {Input, Grid, ComponentStatus} from '@influxdata/clockface'
 
 // Styles
 import 'react-datepicker/dist/react-datepicker.css'
 
 // Types
 import {Columns, ComponentSize} from '@influxdata/clockface'
+import {Form} from 'src/clockface'
```
alexpaxton: Since we used the workaround of passing in the unicode space characters when there is no error, do we still need to use the local `Form`? My hunch is we could import the clockface version and this should still work

ebb-tide: Ah, good point, thanks!

PR #14095, opened 2019-06-10
chore(tasks): update run limits in swagger and FindRuns

### LINKED ISSUE #3119: [0.9.0] Unable to create a database with the name of a query paramater
Trying to create a database called `servers` returns an error:

`{"error":"error parsing query: found SERVERS, expected identifier at line 1, char 15"}`

You can focre it by creating it with double quotes `CREATE DATABASE "servers"` (single quotes throw the same error) but then you get into funky broken territory where it can't write/query it to it unless it's double quoted in the query.

I've run a few tests and get the same behaviour when trying to name a database a name that would be part of a query string (drop, show, database, where, etc...).

AlirieGray:
This PR updates the swagger document to reflect the correct default and maximum number of runs returned by `FindRuns`. It also updates the `FindRuns` function to set the number returned to the default if the limit is 0 or not set, and set it to the maximum if it was higher than the maximum. Previously, it was set to the maximum whether it was zero or greater than the max.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -4349,8 +4349,8 @@ paths:
           schema:
             type: integer
             minimum: 1
-            maximum: 100
-            default: 20
+            maximum: 500
+            default: 100
           description: the number of runs to return
         - in: query
           name: afterTime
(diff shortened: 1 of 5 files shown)


AlirieGray:
closes idpe #3119

PR #14106, opened 2019-06-11
fix(query): recover from panics during compiler and start steps

wolffcm:
Fixes influxdata/idpe#3409.


wolffcm:
Yeah, I think we will recover from a panic on any goroutine spawned from a query request now.

PR #14124, opened 2019-06-12
fix(task): add logging into the task analytical store

lyondhill:


task/backend/analytical_storage.go
```diff
@@ -339,7 +345,7 @@ func (re *runReader) readRuns(cr flux.ColReader) error {
 				logBytes := cr.Strings(j).Value(i)
 				err := json.Unmarshal(logBytes, &r.Log)
 				if err != nil {
-					return err
+					re.logger.Info("failed to parse log data", zap.Error(err))
```
docmerlin: Not sure why you don't return an error here.

lyondhill: so on occasion (especially in cloud beta) some data has been stored that is either missing any log data in the column or the data is incomplete. So if we return an error here people wont see their runs if because of incomplete data on our side.

lyondhill: I made a few improvements to address your concerns :)

PR #14131, opened 2019-06-13
chore(tasks): consolidate task errors into task_errors.go

### LINKED ISSUE #14119: Unify task errors
Currently errors related to tasks can be found in:
- [x] task/backend/task.go
- [x] task/backend/analytical_storage.go
- [x] task/backend/executor/executor.go
- [x] task/backend/coordinator/coordinator.go
- [x] kv/task.go
...

Many of these errors are duplicate errors, like `task not found` and `run not found`

I suggest we move all task related errors to the root directory in a `task_errors.go` file and allow for all task subsystems to use the same error objects.

AlirieGray:
Closes #14119

This PR moves task errors from various locations in the project into a single file in the root directory called `task_errors.go`.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/task_errors.go
+++ b/task_errors.go
@@ -1,0 +1,122 @@
+package influxdb
+
+import (
+	"fmt"
+	"time"
+)
+
+var (
+	// ErrRunCanceled is returned from the RunResult when a Run is Canceled.  It is used mostly internally.
+	ErrRunCanceled = Error{
+		Code: EInternal,
+		Msg:  "run canceled",
+	}
+
+	// ErrTaskNotClaimed is returned when attempting to operate against a task that must be claimed but is not.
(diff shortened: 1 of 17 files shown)


task_errors.go
```diff
@@ -0,0 +1,124 @@
...
+	return &Error{
+		Code: EInvalid,
+		Msg:  fmt.Sprintf("invalid options; Err: %v", err),
+		Op:   "kv/taskOptions",
+		Err:  err,
+	}
+}
+
+// RunNotYetDueError is returned from CreateNextRun if a run is not yet due.
+type RunNotYetDueError struct {
```
lyondhill: I would like to see this in the same format as the other errors. A func instead of a struct.

PR #14135, opened 2019-06-13
refactor(http): make http.Query use FluxCompiler when query is passed as text

wolffcm:
This makes it so that we use `FluxCompiler` to pass the request to the controller when a request comes over http using Flux text instead of AST.

This has two purposes:
- It keeps us from parsing Flux in the http layer, which seems like the wrong place to do it
- It makes it easier to test behavior when there are errors in the `compile` step, since parsing now actually happens there.  

Related PR in Flux:
https://github.com/influxdata/flux/pull/1371


http/task_service.go
```diff
@@ -379,7 +379,7 @@ func (h *TaskHandler) createBootstrapTaskAuthorizationIfNotExists(ctx context.Co
 		return nil, nil
 	}
 
-	prog, err := lang.Compile(t.Flux, time.Now())
+	prog, err := lang.Compile(nil, t.Flux, time.Now())
```
jsternberg: This seems weird and I can't find the corresponding change on flux master.

wolffcm: It's in `lang/compiler.go` line 87 (at the top in the files tab in the Flux PR)

jsternberg: I see it now. I commented there.

PR #14141, opened 2019-06-14
fix(task): bug when urm's are left in the database for task lookup

lyondhill:


kv/task.go
```diff
@@ -368,11 +372,17 @@ func (s *Service) findTaskByOrg(ctx context.Context, tx Tx, filter influxdb.Task
 
 			t, err := s.findTaskByID(ctx, tx, *id)
 			if err != nil {
-				return nil, 0, err
+				if err != backend.ErrTaskNotFound {
```
docmerlin: Shouldn’t that be ==

lyondhill: I think != is what we want here. If we have an error and the error is not a "task not found" error we want to return. otherwise we dont. 
maybe it would be more clear if we did this 
```
if err != nil && err != backend.ErrTaskNotFound {
   return nil, 0, err
}

PR #14147, opened 2019-06-17
fix(influxdb): telegraf compatible mode

kelwang:
Close https://github.com/influxdata/idpe/issues/3758

This change will allow existing old telegraf configs still work with UI before any migration. 
This change won't change the data stored in bolt, unless user made an update.
Manually tested as well.

- [x] Rebased/mergeable
- [x] Tests pass


kelwang:
> Sweet, this'll work. Do we still want to do a migration and later remove `OrganizationID`? Would be good to at least leave a comment.

Yes, once this is deployed, we can do the migration and cleanup anytime

PR #14149, opened 2019-06-17
feat(me): return oauth id with user object

lukevmorris:
`/me` will now include an `oauthID` attribute. If influxdb has a configured OAuth provider, then we should expect `/me` to include the corresponding provider id for that user. If influxdb does not have OAuth configured, and is reliant on the local username/password bolt store, then `oauthID` will be omitted.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -6089,7 +6089,6 @@ components:
           readOnly: true
           type: string
         oauthID:
-          readOnly: true
           type: string
         name:
           type: string
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -6088,6 +6088,9 @@ components:
         id:
           readOnly: true
           type: string
+        oauthID:
+          readOnly: true
+          type: string
         name:
           type: string
         status:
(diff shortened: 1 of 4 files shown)


http/swagger.yml
```diff
@@ -6088,6 +6088,9 @@ components:
         id:
           readOnly: true
           type: string
+        oauthID:
+          readOnly: true
+          type: string
```
kelwang: Looks like the oauthID is added to the session, which is part of authorization, not really part of a user.
I suggest maybe we could make a Me schema, which includes user response and authorization response.

Palakp41: Updated the User interface to include the oauthID instead of the session

Palakp41:
This is the first step. Whenever someone logs in with oauth, we need to store their oauthID in the user repo.

http/swagger.yml
```diff
@@ -6088,6 +6088,9 @@ components:
         id:
           readOnly: true
           type: string
+        oauthID:
+          readOnly: true
```
kelwang: looks good, guess we can remove readOnly for oauthID, since it could be updated

PR #14154, opened 2019-06-18
Add type metadata to User Templates

### LINKED ISSUE #14098: Templates Index should display descriptions and type
The template data structure does have a description field, but the templates index list does not display them. 
It should probably be possible to edit a description as well. 
The template meta also has type information, which should be visible to user, but can not be edited.

mavarius:
Closes #14098

Add type metadata to User Templates

image

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/templates/components/TemplateCard.tsx
+++ b/ui/src/templates/components/TemplateCard.tsx
@@ -114,8 +114,8 @@ class TemplateCard extends PureComponent<Props & WithRouterProps> {
 
   private get description(): JSX.Element {
     const {template} = this.props
-    const description = _.get(template, 'meta.description')
-    const name = _.get(template, 'meta.name')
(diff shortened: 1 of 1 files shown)


ui/src/templates/components/TemplateCard.tsx
```diff
@@ -126,6 +126,16 @@ class TemplateCard extends PureComponent<Props & WithRouterProps> {
     )
   }
 
+  private get templateType(): JSX.Element {
+    const {template} = this.props
+
+    return (
+      <div className="resource-list--meta-item">
+        {_.get(template, 'meta.type')}
```
121watts: Need a default value?

mavarius: Yep, fixed.

PR #14157, opened 2019-06-18
Update client version to resolve query bottleneck

### LINKED ISSUE #14084: Flux response limiting is bottleneck in dashboards
We have a utility that executes a Flux query and returns the results. It has been through several iterations:

1. It was first introduced as a simple wrapper around Axios + the query endpoint
2. We realized that we needed the ability to limit the size of Flux responses, and so we [rewrote the utility](https://github.com/influxdata/chronograf/pull/3596) using `fetch` and the [Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API)
3. We realized that the Streams API was behind a feature flag in Firefox, and thus the utility wasn't working in Firefox. We [rewrote it](https://github.com/influxdata/chronograf/pull/4643) using `XMLHttpRequest`
4. We [extracted this code to `influxdb-client-js`](https://github.com/influxdata/influxdb-client-js/pull/102), wrapping the `XMLHttpRequest` limiting logic in a Node.js Stream

It turns out that the stream wrapper is an extremely costly abstraction, at least it its current incarnation.

I created a dashboard with 9 cells, each containing a single series with about 42k points.

dashboard used to test

I refreshed all cells on this dashboard and recorded a performance profile. The `influxd` instance was on my laptop and no additional network latency was introduced.

half the time is spent receiving data

52% of the total time was spent piping data through the stream wrapper. It took ~4x longer to pipe data through the stream wrapper than it did to parse it.

I ran another test, replacing the stream wrapper limiting logic with our previous `fetch` / Streams API approach. Profiling the same interaction showed that the time spent reading the Flux responses was entirely negligable (0.3% of total time):

fetch api is much faster

Since the Streams API was de-feature flagged in Firefox 65, this is now a viable option for removing this dashboards bottleneck. Note that it has slightly different behavior: the node stream approach limits responses to a maximum number of rows, while the `fetch` approach limits responses to a maximum number of bytes.

Some other possible approaches:

- Remove the stream wrapper in the client in favor of just exposing a promise
- Replace the stream with something more efficient
- Tweak the stream wrapper so that it doesn't output each line individual, but outputs chunks of lines (maybe once per Flux result?)

The need to fix this bottleneck begs the question: should the Node stream wrapper be in the client at all, given its performance issues?

chnn:
Closes #14084

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+## v2.0.0-alpha.13 [unreleased]
+
+### Features
+
+### Bug Fixes
+
+### UI Improvements
+
+1. [14157](https://github.com/influxdata/influxdb/pull/14157): Remove rendering bottleneck when streaming Flux responses
+
 ## v2.0.0-alpha.13 [2019-06-13]
 
 ### Features
--- a/ui/src/shared/apis/query.ts
+++ b/ui/src/shared/apis/query.ts
@@ -1,68 +1,36 @@
-import Deferred from 'src/utils/Deferred'
 import {getWindowVars} from 'src/variables/utils/getWindowVars'
 import {buildVarsOption} from 'src/variables/utils/buildVarsOption'
 import {client} from 'src/utils/api'
 
-import {File} from '@influxdata/influx'
+import {
+  File,
+  CancellationError as ClientCancellationError,
+} from '@influxdata/influx'
 
 // Types
 import {WrappedCancelablePromise, CancellationError} from 'src/types/promises'
 import {VariableAssignment} from 'src/types/ast'
 
-const MAX_ROWS = 50000
-
-export interface ExecuteFluxQueryResult {
-  csv: string
-  didTruncate: boolean
-  rowCount: number
-}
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## v2.0.0-alpha.13 [unreleased]
+
+### Features
+
+### Bug Fixes
+
+### UI Improvements
+
+1. [14157](https://github.com/influxdata/influxdb/pull/14157): Remove rendering bottleneck when streaming Flux responses
```
OfTheDelmer: nice!

ui/src/shared/apis/query.ts
```diff
@@ -1,68 +1,36 @@
...
-    if (rowCount < MAX_ROWS) {
-      return
-    }
-
-    didTruncate = true
-    conn.cancel()
+): WrappedCancelablePromise<string> => {
+  const {promise, cancel} = client.queries.execute(orgID, query, {
+    extern,
+    limitChars: MAX_RESPONSE_CHARS,
```
OfTheDelmer: I like the clean up here.

PR #14163, opened 2019-06-19
feat(task): Build interfaces for a generic scheduler.

lyondhill:


task/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,78 @@
...
+	// if Now is zero we can just do the first 2 steps (This is how we would trigger manual runs).
+	// Errors returned from the execute request imply that this attempt has failed and
+	// should be put back in scheduler and re executed at a alter time. We will add scheduler specific errors
+	// so the time can be configurable.
+	Execute(ctx context.Context, id influxdb.ID, meta interface{}, now time.Time) (Execution, error)
+
+	CancelRun(ctx context.Context, taskID, runID influxdb.ID) error
+}
+
+type Execution interface {
```
docmerlin: Not sure that execution needs to be an interface instead of a concrete.

task/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/influxdata/influxdb"
+)
+
+type Storage interface {
+	// UpdateLatestCompleted will allow us to set a latest completed time for an id.
+	UpdateLatestCompleted(ctx context.Context, id influxdb.ID, lastRun time.Time) error
+
+	// retrieve the latest completed for this schedulable id.
+	// If GetLatestCompleted is empty scheduler will use "now" as its latest completed
```
docmerlin: This would prevent you from running historical jobs?

task/backend/scheduler/scheduler.go
lyondhill: the reason I thought of the execution as a interface is I wanted the executor to be able to define how each of those pieces respond. I believe by using a static type anything that implements the "executor" interface would have less power.

task/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,78 @@
+package scheduler
+
+import (
+	"context"
+	"time"
+
+	"github.com/influxdata/influxdb"
+)
+
+type Storage interface {
```
goller: Would you comment // Storage ...

task/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,78 @@
...
+	// UpdateLatestCompleted will allow us to set a latest completed time for an id.
+	UpdateLatestCompleted(ctx context.Context, id influxdb.ID, lastRun time.Time) error
+
+	// retrieve the latest completed for this schedulable id.
+	// If GetLatestCompleted is empty scheduler will use "now" as its latest completed
+	GetLatestCompleted(ctx context.Context, id influxdb.ID) (time.Time, error)
```
goller: Comment should be in the form // GetLatestCompleted ...

task/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+type Executor interface {
+	// Execute is used to execute run's for any schedulable object.
```
goller: Add comment // Executor ...

task/backend/scheduler/scheduler.go
lyondhill: So historical jobs can be run strictly inside the executor in this iteration. We could expand the storage options to allow a list of time's we need to attempt to run again? what are you thinking?

task/backend/scheduler/scheduler.go
goller: What do you think of changing `GetLatestCompleted` to `CompletionTime` or `LatestTime` or `LatestCompletionTime` ?  I think it is good to have units and that generally no need for a `Get` prefix.

task/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/influxdata/influxdb"
+)
+
+type Storage interface {
+	// UpdateLatestCompleted will allow us to set a latest completed time for an id.
+	UpdateLatestCompleted(ctx context.Context, id influxdb.ID, lastRun time.Time) error
```
goller: What do you think of the name `SetCompletionTime`?  That way it would include units.  Also, does that fact that it is the "latest" matter to the users of storage?

task/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,78 @@
...
+	Execute(ctx context.Context, id influxdb.ID, meta interface{}, now time.Time) (Execution, error)
+
+	CancelRun(ctx context.Context, taskID, runID influxdb.ID) error
+}
+
+type Execution interface {
+	// ID is a unique ID usable to look up or cancel a run
+	ID() influxdb.ID
+
+	// Status can be used to view what the current status of this execution is
```
goller: Could the Status() return type be a type other than string? Is the Status() string a user-facing message?  If so, would you mention that it is user-facing in the comments?

task/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+type Executor interface {
+	// Execute is used to execute run's for any schedulable object.
+	// the executor can go through manual runs, clean currently running, and then create a new run based on `now`.
+	// if Now is zero we can just do the first 2 steps (This is how we would trigger manual runs).
+	// Errors returned from the execute request imply that this attempt has failed and
+	// should be put back in scheduler and re executed at a alter time. We will add scheduler specific errors
+	// so the time can be configurable.
+	Execute(ctx context.Context, id influxdb.ID, meta interface{}, now time.Time) (Execution, error)
```
goller: Is there any way to give `meta` more structure?  It always feels too free-form to me that an interface defines a function that has an interface argument.

task/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,78 @@
...
+	// This is a hanging call until Done() is closed.
+	Error() error
+}
+
+type Schedulable interface {
+	// ID a unique ID we can to lookup a scheduler
+	ID() influxdb.ID
+
+	// Schedule is the schedule you want execution
+	Schedule() string // cron or every dx
```
goller: Is this string only user-facing?  If not, I would think it should be a struct of some sort; if so, would you comment that it is user-facing?

task/backend/scheduler/scheduler.go
docmerlin: If GetLatestCompleted is empty, why not return a zero time?
Then the scheduler or whatever can decide what to do based on that, but it knows that it has a zero time.

docmerlin: I guess another option would be to return a special error a "NeverCompletedBefore" error that the scheduler (or whoever is asking then knows how to respond.

lyondhill: I agree. I think the scheduler could decide how it handles a "not completed" response.

task/backend/scheduler/scheduler.go
lyondhill: I like `LatestCompletionTime`

task/backend/scheduler/scheduler.go
lyondhill: the "storage" may not need to know its the "latest" completed. so your thinking `CompletionTime` and `SetCompletionTime`?

task/backend/scheduler/scheduler.go
lyondhill: We have a system now that is a RunStatus int that has a `String` and some json marshaling function. I suspect that would be better here. I suspect the status would be used for logging/metrics. Inside the influxdb system this status would be used in the run list http action.

task/backend/scheduler/scheduler.go
lyondhill: What do you think a struct here would look like?

task/backend/scheduler/scheduler.go
lyondhill: I had thought about that, I dont love that it can be anything. I wanted it to be agnostic to our usage. I worried that by defining what it would look like may restrict its usefulness as a open source package.

task/backend/scheduler/scheduler.go
lyondhill: Whatever status ends up being I would like it to be some public type. I think it would make anything implementing the executor interface much cleaner if it didn't have to import the scheduler.

task/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,103 @@
...
+
+// Promise is the currently executing element from the a call to Execute.
+type Promise interface {
+	// ID is a unique ID usable to look up or cancel a run
+	ID() ID
+
+	// Cancel a promise, identical to calling executor.Cancel()
+	Cancel(ctx context.Context)
+
+	// Done() returns a read only channel that when closed indicates the execution is complete
```
goller: Probably just Done rather than Done()

task/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,103 @@
...
+// Schedulable is a object that can be scheduled by the scheduler.
+type Schedulable interface {
+	// ID a unique ID we can to lookup a scheduler
+	ID() ID
+
+	// Schedule is the schedule you want execution.
+	Schedule() Schedule
+}
+
+type Schedule struct {
```
goller: Comment

task/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,103 @@
...
+	Schedule string
+	Offset   time.Duration
+}
+
+func (s Schedule) Valid() error {
```
goller: Comment

task/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,103 @@
...
+	_, err := cron.Parse(s.Schedule)
+	return err
+}
+
+// Next returns the next time a a schedule needs to run
+func (s Schedule) Next(checkpoint time.Time) (time.Time, error) {
```
goller: would you mention in the comment that the checkpoint time is the next time to schedule a run after the checkpoint?

task/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/influxdata/influxdb"
+	cron "gopkg.in/robfig/cron.v2"
+)
+
+// ID
+type ID influxdb.ID
+
+// Checkpointer allows us to restart a service from the last time we executed
+type Checkpointer interface {
+
```
goller: this looks like extra whitespace to me

task/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,103 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/influxdata/influxdb"
+	cron "gopkg.in/robfig/cron.v2"
+)
+
+// ID
```
goller: Might as well add your ideas as a TODO comment about how you don't want to import ID

task/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,103 @@
...
+type ID influxdb.ID
+
+// Checkpointer allows us to restart a service from the last time we executed
```
goller: Period at the end of the comment

PR #14167, opened 2019-06-20
deps: upgrade flux to v0.33.2

jsternberg:


jsternberg:
I found another possibility. The influxdb tables no longer report `Empty()` correctly because they don't know until `Do` has been called. The previous fix assumed that the table buffer was the only one that reported it incorrectly.

I added a new change to flux in a branch that would handle more situations and would more aggressively protect against the possibility of reporting this incorrectly. I'm updated this to use a specific revision to see if the tests succeed and then that can get reviewed and be part of `v0.33.2`.

jsternberg:
The issue will be fixed by #14193.

PR #14171, opened 2019-06-21
fix(ui/cloudNav): Update cloud subnav feature flag to use env var

ischolten:
Subnav should be visible in acceptance but feature flagged in production
env var is `CLOUD_BILLING_VISIBLE`

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/pageLayout/components/CloudNav.tsx
+++ b/ui/src/pageLayout/components/CloudNav.tsx
@@ -25,7 +25,7 @@ export default class CloudNav extends PureComponent {
             </a>
           )}
         >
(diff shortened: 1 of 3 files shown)


ui/src/pageLayout/components/CloudNav.tsx
```diff
@@ -25,7 +25,7 @@ export default class CloudNav extends PureComponent {
             </a>
           )}
         >
-          <FeatureFlag name="cloudSubNav">
+          <FeatureFlag name="cloudBilling">
```
OfTheDelmer: nice

Issue #14161, opened 2019-06-19
Create simple controller for 1.x Flux


nathanielc:
The controller was removed from Flux. This means that we need to create a simple implementation of the controller for 1.x

See this branch `chore/update-flux-0.33.0` for a starting point.

nathanielc:
See https://github.com/influxdata/influxdb/commit/0f997d53c0b839fbb7655555b3070875135f5f87 for what needs to change in part.

jsternberg:
I think the decision was just to compile the program and call start on it so I'm going to close this issue since that is done now.

PR #14177, opened 2019-06-21
Fix and implement the controller and http service for flux v0.33.0

### LINKED ISSUE #14159: Update http code for Flux changes in 1.x
There were some changes to the http bits for Flux. On this branch `chore/update-flux-0.33.0` you will see that the code no longer compiles.

jsternberg:
Fixes #14159 and #14161.


services/httpd/handler.go
```diff
@@ -1796,7 +1796,7 @@ func (h *Handler) recovery(inner http.Handler, name string) http.Handler {
 
 // Store describes the behaviour of the storage packages Store type.
 type Store interface {
-	Read(ctx context.Context, req *datatypes.ReadRequest) (reads.ResultSet, error)
+	ReadFilter(ctx context.Context, req *datatypes.ReadFilterRequest) (reads.ResultSet, error)
```
affo: Is this a piece of another PR?

internal/flux_controller.go
```diff
@@ -14,9 +15,12 @@ type FluxControllerMock struct {
...
 		QueryFn: func(ctx context.Context, compiler flux.Compiler) (query flux.Query, e error) {
-			// See https://github.com/influxdata/influxdb/issues/14161
-			//return NewFluxQueryMock(), nil
-			return nil, nil
+			p, err := compiler.Compile(ctx)
+			if err != nil {
+				return nil, err
+			}
+			alloc := &memory.Allocator{}
+			return p.Start(ctx, alloc)
```
affo: Can't you adapt `FluxQueryMock` to implement `flux.Query` and return that instead?

services/httpd/handler_test.go
```diff
@@ -1078,7 +1080,9 @@ func TestHandler_Flux_QueryJSON(t *testing.T) {
 			t.Fatalf("unexpected query -got/+exp\n%s", cmp.Diff(c.Query, exp))
 		}
 		called = true
-		return internal.NewFluxQueryMock(), nil
+
+		p := &mock.Program{}
+		return p.Start(ctx, nil)
```
affo: You could  use `FluxQueryMock` if that implements `flux.Query`. Otherwise, I think you can just remove that.

services/httpd/handler.go
jsternberg: Yea. This was a change to the storage interface.

internal/flux_controller.go
jsternberg: I thought it was easier to just have the default use the compiler. The purpose of this is to make it possible to override the query function for the controller I believe.

services/httpd/handler_test.go
jsternberg: We added a mock package at some point to flux probably after the last time we updated this so I thought it was easier to just start using that since it would actually be maintained while the offshoot struct in the 1.x line of InfluxDB would likely get out of sync again.

cmd/store/query/query.go
```diff
@@ -1,3 +1,5 @@
+//+build ignore
```
affo: Are you ignoring these just to make it build?

Do we have issues to fix commands?

jsternberg: We've apparently fixed it on InfluxDB 2.0 and it's an internal command that uses a now removed interface according to @stuartcarnie so he said it was fine if I just ignore the files for now. Since it's not part of the release, I figured it would be fine.

PR #14181, opened 2019-06-24
fix(storage): fix counts for level 4+ files

tmgordeeva:
The counts weren't adding all the level 4+ files, so the last one to be counted
would override the rest.


tsdb/tsm1/file_store.go
```diff
@@ -332,8 +332,13 @@ func (t *fileTracker) Bytes() uint64 { return atomic.LoadUint64(&t.diskBytes) }
 func (t *fileTracker) SetBytes(bytes map[int]uint64) {
 	total := uint64(0)
```
jacobmarble: What is `total` for? Looks like it never changes from zero?

Issue #14020, opened 2019-05-29
User selects token for task


bthesorceror:
In Data Explorer Save As:

Given I am a user creating a task and a list of my existing tokens
When I create a task I can select a token from a dropdown
Then the created task will be run with that token

bthesorceror:
Selecting a token from the dropdown should add the token's id to the request to create a token. A task should not be able to be made without a token.

lukevmorris:
- Display input and output buckets
- Display only the relevant tokens for this task

PR #14187, opened 2019-06-24
chore(ui): upgrade dependencies

chnn:
Upgrades giraffe and influxdb-client-js dependencies.

Makes a few source tweaks to accommodate changes in the new client version.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -140,10 +140,10 @@
   },
   "dependencies": {
     "@influxdata/clockface": "0.0.13",
-    "@influxdata/influx": "github:influxdata/influxdb-client-js#0911a12853c53bbc3e20092d78e6b1cacf377878",
+    "@influxdata/giraffe": "0.12.2",
+    "@influxdata/influx": "0.4.0",
     "@influxdata/influxdb-templates": "0.2.0",
     "@influxdata/react-custom-scrollbars": "4.3.8",
-    "@influxdata/giraffe": "0.12.1",
     "axios": "^0.19.0",
     "babel-polyfill": "^6.26.0",
     "bignumber.js": "^4.0.2",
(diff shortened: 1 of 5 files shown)


ui/package.json
```diff
@@ -140,10 +140,10 @@
   },
   "dependencies": {
     "@influxdata/clockface": "0.0.13",
-    "@influxdata/influx": "github:influxdata/influxdb-client-js#0911a12853c53bbc3e20092d78e6b1cacf377878",
+    "@influxdata/giraffe": "0.12.2",
```
OfTheDelmer: 🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒🦒

ui/src/tasks/actions/index.ts
```diff
@@ -422,7 +422,7 @@ export const saveNewScript = (script: string, preamble: string) => async (
       orgs: {org},
     } = getState()
 
-    await client.tasks.createByOrgID(org.id, fluxScript)
+    await client.tasks.createByOrgID(org.id, fluxScript, null)
```
OfTheDelmer: What is the null for here? Could we name this in a const and provide it for clarity?

OfTheDelmer: **is there a diff between providing `null` vs excluding it as `undefined`.

chnn: Palak is working on https://github.com/influxdata/influxdb/issues/14020, which required updating the client so that a token is always passed. But the influxdb side of her work hasn't landed yet and I wanted to update the dependency in the meantime, hence this funkiness. 

Passing `null` just means to preserve the old behavior. Not obvious at all, but it will be shortly removed once Palak merges her work.

chnn: 😬

Issue #14190, opened 2019-06-24
Buggy gauge / single stat threshold settings


chnn:
The thresholds list used in the gauge and single settings has some issues:

- Displays a state inconsistent with the settings it emits:

   ![buggy thresholdslist](https://user-images.githubusercontent.com/638955/60054727-33cc6100-9690-11e9-87e9-3e2d03abcb2d.gif)

- Doesn't allow setting negative numbers by typing them, but it does allow setting them by entering a positive number first then pressing the down arrow key

- Cannot delete existing number in input and then type a new one since it inserts 0 when the last character is deleted

Here's how I think it should work instead:

- Let a user enter whatever they want into an input

- On input blur or when enter is pressed, the component attempts to validate and emit the updated thresholds

   - If the entered thresholds are valid then they are emitted

   - If the entered thresholds are invalid, an error is displayed and the thresholds are not emitted

   - The error does not clear until a user performs either a blur or presses in an input, in which case the validation is rerun

- The component can be in an invalid state either because an individual input is invalid (e.g. user types "baloney" into an input but we wanted a number) or because some validation between inputs is invalid (e.g. the min is greater than the max)

- If the component receives new thresholds at any time, it should reset its internal state to display those thresholds and their validation state

Here's an example of a similar UI control that exhibits the desired behavior:

![not buggy](https://user-images.githubusercontent.com/638955/60055378-49db2100-9692-11e9-9306-fc1d8be65829.gif)

Issue #14175, opened 2019-06-21
Gauges don't handle negative number correctly


davidgs:
__Steps to reproduce:__
List the minimal actions needed to reproduce the behavior.

1. Create a gauge
2. set the range from -120 to -10
3. Enjoy the results. 

__Expected behavior:__
Describe what you expected to happen.
Gauge should add negative numbers to the the tick-marks

__Actual behavior:__
Describe What actually happened.
![Screen Shot 2019-06-21 at 1 57 37 PM](https://user-images.githubusercontent.com/2071898/59942163-f3b97400-942c-11e9-984a-6fcbb9dcda75.png)

__Environment info:__

* System info: Run `uname -srm` and copy the output here
Linux 3.10.0-862.14.4.el7.x86_64 x86_64
* InfluxDB version: Run `influxd version` and copy the output here
[davidgs@whm ~]$ ./influxd version
Error: unknown command "version" for "influxd"'

2.0 Alpha doesn't understand the version, but this is Alpha 13
* Other relevant environment details: Container runtime, disk info, etc

__Config:__
Copy any non-default config values here or attach the full config as a gist or file.

__Logs:__
Include snippet of errors in log.

__Performance:__
Generate profiles with the following commands for bugs related to performance, locking, out of memory (OOM), etc.

```sh
# Commands should be run when the bug is actively.
# Note: This command will run for at least 30 seconds.
curl -o profiles.tar.gz "http://localhost:8086/debug/pprof/all?cpu=true"
curl -o vars.txt "http://localhost:8086/debug/vars"
iostat -xd 1 30 > iostat.txt
# Attach the `profiles.tar.gz`, `vars.txt`, and `iostat.txt` output files.
```

PR #14194, opened 2019-06-24
Gauge polish

chnn:
I investigated https://github.com/influxdata/influxdb/issues/14175 + https://github.com/influxdata/EAR/issues/823  and discovered the root issue is that the threshold settings component emits nonsensical settings. I opened up https://github.com/influxdata/influxdb/issues/14190 to track resolution.

In the meantime I made some minor gauge tweaks. It now displays better on HiDPI / Retina screens.

Before (you will need a HiDPI screen to see the blurriness): 

before

After:

after

I also tweaked how the gauge behaves when the needle position is beyond the domain of the gauge (for example if the current value is 6,000,000 and the max value for the gauge is 100). Before the needle would display at an essentially random position ¯\_(ツ)_/¯ (some sort of modular arithmetic).

Now it will be clamped to just beyond the domain:

Screen Shot 2019-06-24 at 3 30 58 PM

### DIFF
--- a/ui/src/shared/components/Gauge.tsx
+++ b/ui/src/shared/components/Gauge.tsx
@@ -282,15 +282,14 @@ class Gauge extends Component<Props> {
     const {prefix, suffix, decimalPlaces} = this.props
     const {degree, lineCount, labelColor, labelFontSize} = GAUGE_SPECS
 
-    const incrementValue = (maxValue - minValue) / lineCount
+    const tickValues = [
+      ..._.range(minValue, maxValue, Math.abs(maxValue - minValue) / lineCount),
+      maxValue,
+    ]
 
-    const labels = []
-    for (let g = minValue; g < maxValue; g += incrementValue) {
-      const valueString = formatStatValue(g, {decimalPlaces, prefix, suffix})
-      labels.push(valueString)
-    }
-
-    labels.push(formatStatValue(maxValue, {decimalPlaces, prefix, suffix}))
+    const labels = tickValues.map(tick =>
+      formatStatValue(tick, {decimalPlaces, prefix, suffix})
+    )
 
     const startDegree = degree * 135
     const arcLength = Math.PI * 1.5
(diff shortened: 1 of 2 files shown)


ui/src/shared/components/Gauge.tsx
```diff
@@ -123,6 +118,23 @@ class Gauge extends Component<Props> {
     this.drawNeedle(ctx, radius, minValue, maxValue)
   }
 
+  private resetCanvas = () => {
+    const canvas = this.canvasRef.current
+    const ctx = canvas.getContext('2d')
+    const {width, height} = this.props
+    const dpRatio = window.devicePixelRatio || 1
+
+    // Set up canvas to draw on HiDPI / Retina screens correctly
```
ischolten: wooot!

ui/src/shared/components/Gauge.tsx
```diff
@@ -282,15 +282,14 @@ class Gauge extends Component<Props> {
     const {prefix, suffix, decimalPlaces} = this.props
     const {degree, lineCount, labelColor, labelFontSize} = GAUGE_SPECS
 
-    const incrementValue = (maxValue - minValue) / lineCount
+    const tickValues = [
+      ..._.range(minValue, maxValue, Math.abs(maxValue - minValue) / lineCount),
```
ischolten: 🚀

ui/src/shared/components/Gauge.tsx
```diff
@@ -282,15 +282,14 @@ class Gauge extends Component<Props> {
...
+      maxValue,
+    ]
 
-    const labels = []
-    for (let g = minValue; g < maxValue; g += incrementValue) {
```
ischolten: 🔥 🔥 🔥

ui/src/shared/components/Gauge.tsx
```diff
@@ -349,10 +348,18 @@ class Gauge extends Component<Props> {
...
   private drawNeedle = (ctx, radius, minValue, maxValue) => {
     const {gaugePosition} = this.props
-    const {degree, needleColor0, needleColor1} = GAUGE_SPECS
+    const {degree, needleColor0, needleColor1, overflowDelta} = GAUGE_SPECS
     const arcDistance = Math.PI * 1.5
 
-    const needleRotation = (gaugePosition - minValue) / (maxValue - minValue)
+    let needleRotation: number
+
+    if (gaugePosition <= minValue) {
```
ischolten: This is awesome! how do you feel about a function like `getNeedleRotation` and sticking that logic there?

PR #14199, opened 2019-06-25
Implement the TagKeys and TagValues storage rpc calls for 1.x

jsternberg:


flux/stdlib/influxdata/influxdb/rules.patch
```diff
@@ -0,0 +1,48 @@
+diff --git a/flux/stdlib/influxdata/influxdb/rules.go b/flux/stdlib/influxdata/influxdb/rules.go
```
aanthony1243: need to git rm this?

services/storage/store.go
```diff
@@ -187,11 +190,219 @@ func (s *Store) ReadGroup(ctx context.Context, req *datatypes.ReadGroupRequest)
...
+		if found := reads.HasFieldValueKey(expr); found {
+			return nil, errors.New("field values unsupported")
+		}
+		expr = influxql.Reduce(influxql.CloneExpr(expr), nil)
+		if reads.IsTrueBooleanLiteral(expr) {
+			expr = nil
+		}
+	}
+
+	// TODO(jsternberg): Use a real authorizer.
```
aanthony1243: create an issue for this and link?

flux/stdlib/influxdata/influxdb/rules.patch
jsternberg: That was intentionally kept. I was thinking about maybe doing it for everything that I modified so I don't have to do this again? I wasn't sure if it was a good idea or not.

PR #14201, opened 2019-06-25
fix(storage): replace panic with error in influx_inspect verify

### LINKED ISSUE #9763: influx_inspect verify panics when data dir is missing
`influx_inspect verify` doesn't have a very nice UX with respect to the data dir. If it's missing, e.g., because you're on a system that uses a non-default data directory (not in `.influxdb`) then it panics.

It shouldn't panic, and the usage instructions should be better.

adamperlin:
Closes #9763 

Replaces panic with error handling in `influx_inspect verify` if the storage directory cannot be read from. 

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influx_inspect/verify/tsm/verify.go
+++ b/cmd/influx_inspect/verify/tsm/verify.go
@@ -67,7 +67,8 @@ func (cmd *Command) printUsage() {
 Usage: influx_inspect verify [flags]
 
     -dir <path>
-            Root storage path
+            The root storage path.
+            Must be changed if you are using a non-default storage directory.
             Defaults to "%[1]s/.influxdb".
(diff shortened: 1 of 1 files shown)


cmd/influx_inspect/verify/tsm/verify.go
```diff
@@ -66,9 +66,10 @@ func (cmd *Command) printUsage() {
 
 Usage: influx_inspect verify [flags]
 
-    -dir <path>
-            Root storage path
-            Defaults to "%[1]s/.influxdb".
+	-dir <path>
```
jacobmarble: Replace tabs with spaces, as the surrounding text is

PR #14206, opened 2019-06-26
deps: upgrade to flux v0.34.1

jsternberg:


jsternberg:
The test failure seems to be unrelated. The dependency we are upgrading to is identical to the previous one, but the version is different because `v0.34.0` had to be pulled because it had a committed binary in it.

PR #14208, opened 2019-06-26
Update flux version to v0.33.2

jsternberg:
The flux in influxdb has been upgraded to use v0.33.2. A lot of
interfaces for the storage engine were changed during this so code had
to change to accomodate the new interfaces and remove the old ones.

Included in this commit is a patch file for the changes that were made.
A patch was generated for the following packages:

* `flux/stdlib/influxdata/influxdb`
* `storage/reads`
* `tsdb/cursors`

These are the three packages that are in common with version 2 of the
database and the first of these packages contains the specific
implementations that are used for version 1.

It is very possible that the next time we upgrade this, the patch will
not apply cleanly just like it wouldn't have applied cleanly to this
update. The patch is mostly meant to document exactly what changed
during the copy over to help ensure we don't forget things when adapting
the interfaces.

Add a patch file to hopefully make this easier in the future


jsternberg:
Yes. I will attempt to use the patch to apply this to 1.7 after this is merged.

jsternberg:
I think the implementations of `TagKeys` and `TagValues` are completely new.

Issue #14193, opened 2019-06-24
storage/reads: table should return empty as true when the table is empty


jsternberg:
During a past refactor for `storage/reads`, the `Empty()` method was set to always return false for simplicity. We should see if it is easy enough to have `Empty()` return a meaningful value to improve query efficiency and reduce confusion about the return value.

wolffcm:
It seems really unintuitive to have a method called `Empty()` that will return false when the table is empty.  I understand why we may not know if the table is empty until we actually try to iterate over it, but the potential for a bug resulting from this behavior seems high.

Other possible ways to address this confusion:
- Change the name of the method to `EmptyHint()` or something
- Make `Empty()` part of an optional interface that not all implementors of `flux.Table` have, if the underlying data model doesn't support it.

nathanielc:
From my comment here https://github.com/influxdata/flux/pull/1414#discussion_r296946105

`Empty` is not a hint and all implementations need to implement it correctly. This need stems from the need to be able to encode and decode empty tables in CSV. I don't think this need is unique to the CSV format and I would expect other formats to need to understand how to encode empty tables explicitly. This is because while tables are empty (they have no rows) the table still has a group key and its values that need to be encoded.

I believe this issues is a non-issue though because the storage RPC calls never return empty tables. Meaning at the storage layer if there was no block to decode with values then no table is produced and returned over the RPC interface. As such the current `return false` implementation of the table implementation in `storage/reads` is correct.

jsternberg:
@nathanielc I don't think that's true, but we'll have to talk to @stuartcarnie because the previous implementation of `storage/reads` did return an actual value for it and the test seems to pass on CI if I stop relying on `Empty()`.

I'm not sure what you mean since I think it would be possible to encode and decode empty tables in CSV without `Empty()` (you just delay writing the header until you're in the call to `Do`). It also makes the `Do` implementations more complicated for each of the tables. But, I've added a bug label to this so it can be fixed.

We can talk about it more tomorrow. If we're deciding that the problem is in influxdb, then influxdata/flux#1414 can likely be closed instead of merged.

PR #14234, opened 2019-06-28
fix(storage/reads): storage table implementation passes table tests

jsternberg:
#14193


jsternberg:
Everything in this PR was necessary, but a lot of the stuff was partially reverting the code.

The big new things were `used` to track if a table has been used or not and the implementation of empty tracks how things used to work.

PR #14246, opened 2019-07-02
Clears tsm files when replace fails

### LINKED ISSUE #14058: Temporary TSM files not cleaned up when compaction rename fails
As a result of analyzing #11339, it was determined there is an issue with the final stage of compaction during the exchange of old with new TSM files. Specifically, if an error occurs during the `FileStore.replace` call, which is called indirectly via `ReplaceWithCallback` in the `compactGroup` API:

https://github.com/influxdata/influxdb/blob/aa3dfc066260674699ef8b01a4499bdcb425537a/tsdb/engine/tsm1/engine.go#L2209-L2214

Note that the error is simply logged.

Alternatively, when a level-0 (cache snapshot) is written and `Replace` fails, the temporary files are cleaned up:

https://github.com/influxdata/influxdb/blob/aa3dfc066260674699ef8b01a4499bdcb425537a/tsdb/engine/tsm1/engine.go#L1925-L1933

Per #11339, `FileStore.replace` failed because the server was low on memory and was unable to `mmap` the new TSM files.

maxunt:
Closes #14058 

When compaction fails, instead of simply printing error message, clear tmp.tsm files so we can try again

- [ ] Rebased/mergeable
- [ ] Tests pass


tsdb/tsm1/engine.go
```diff
@@ -1170,6 +1170,14 @@ func (s *compactionStrategy) compactGroup(ctx context.Context) {
 		log.Info("Error replacing new TSM files", zap.Error(err))
 		s.tracker.Attempted(s.level, false, "", 0)
 		time.Sleep(time.Second)
+
+		// Remove the new snapshot files. We will try again.
+		for _, file := range files {
+			if err := os.Remove(file); err != nil {
+				log.Info("Unable to remove file", zap.String("path", file), zap.Error(err))
```
e-dard: this should be an `Error` level not `Info`.

maxunt:
alchemy please merge

PR #14257, opened 2019-07-03
feat(tasks): add last and checkpoint functions

AlirieGray:
This PR implements the `Checkpoint` and `Last` functions using a Task Service

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/task/backend/task_service_checkpointer.go
+++ b/task/backend/task_service_checkpointer.go
@@ -1,0 +1,47 @@
+package backend
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/influxdata/influxdb"
+)
+
+// Checkpointer allows us to restart a service from the last time we executed.
+type TaskServiceCheckpointer struct {
+	ts influxdb.TaskService
+}
+
+func NewTaskServiceCheckpointer(ts influxdb.TaskService) *TaskServiceCheckpointer {
+	return &TaskServiceCheckpointer{
+		ts: ts,
+	}
+}
+
+// Checkpoint updates a task's LatestCompleted value with the given time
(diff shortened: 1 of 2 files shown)


task/backend/scheduler/scheduler.go
```diff
@@ -13,15 +13,6 @@ import (
 // TODO(lh): maybe make this its own thing sometime in the future.
 type ID influxdb.ID
 
-// Checkpointer allows us to restart a service from the last time we executed.
```
lyondhill: The interface is required for the scheduler. You are writing a implementation of this interface. We will have several implementations of the interface.

task/backend/task_service_checkpointer.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/influxdata/influxdb"
+)
+
+// Checkpointer allows us to restart a service from the last time we executed.
+type Checkpointer struct {
```
lyondhill: This is a "task service checkpointer" meaning it implements the checkpointer interface but uses a task service to do it. A more specific name would be better here.

task/servicetest/servicetest.go
```diff
@@ -368,6 +368,23 @@ func testTaskCRUD(t *testing.T, sys *System) {
 		t.Fatalf("expected task status to be active, got %q", f.Status)
 	}
 
+	// Update task: checkpoint & last
```
lyondhill: the test suite in servicetest is specifically designed to test anything that implements the `TaskService` interface. You need to write a suite of tests that tests the `Checkpointer`. I think your test suite should in the backend somewhere.

PR #14262, opened 2019-07-04
Storage reads should be in domain [start, end)

### LINKED ISSUE #14249: storage read RPC assumes inclusive stop boundary, should be exclusive
the RPC defines an inclusive boundary:  https://github.com/influxdata/influxdb/blob/faab75968b7cf66e824e95506f439c1301c10365/storage/reads/datatypes/storage_common.proto#L176

while the spec defines an exclusive boundary: 
https://github.com/influxdata/flux/blob/master/docs/SPEC.md#range

e-dard:
Closes https://github.com/influxdata/flux/issues/1427
Closes #14249

This PR changes how storage read cursors work. Previously storage returned points in the domain `[start, end]`.

However, Flux's `range` function uses the range `[start, stop)`. 

Storage now uses the domain `[start, end)`.

### DIFF
--- a/storage/reads/datatypes/storage_common.proto
+++ b/storage/reads/datatypes/storage_common.proto
@@ -173,7 +173,7 @@ message TimestampRange {
   // Start defines the inclusive lower bound.
   int64 start = 1;
 
-  // End defines the inclusive upper bound.
+  // End defines the exclusive upper bound.
   int64 end = 2;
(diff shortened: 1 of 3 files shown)


e-dard:
This is affecting users in 1.x also, e.g., see https://github.com/influxdata/flux/issues/2170

PR #14263, opened 2019-07-04
fix(storage): add key field to keyRange

zhulongcheng:
add `key` field to `keyRange`.

so no need to read SeriesKey from the segment file.


benbjohnson:
Yes, this seems right to me. 👍

PR #14280, opened 2019-07-08
feat(fs): API for replacing os calls

maxunt:
This PR replaces calls to os.Create and os.Rename with our own API for consistent error messaging and to prevent writing over files that already exist

- [x] Rebased/mergeable
- [x] Tests pass


chronograf/etc/build.py
```diff
@@ -697,15 +698,15 @@ def package(build_output, pkg_name, version, nightly=False, iteration=1, static=
                                 #     logging.debug("Changing package output version from {} to {} for RPM.".format(version, package_version))
                                 # Strip nightly version from package name
                                 new_outfile = outfile.replace("{}-{}".format(package_version, package_iteration), "nightly")
-                                os.rename(outfile, new_outfile)
+                                fs.RenameFile(outfile, new_outfile)
```
e-dard: What's going on with this Python code?

e-dard:
@maxunt needs at least one more careful review.

pkg/fs/fs_test.go
```diff
@@ -0,0 +1,160 @@
...
+		panic(fmt.Sprintf("failed to create temp file: %v", err))
+	}
+
+	name := f.Name()
+	f.WriteString(name)
+	if err := f.Close(); err != nil {
+		panic(err)
+	}
+	return name
+}
```
benbjohnson: Minor nit but it's not clear from the name that it's writing its own name to the filename. I think it would be clearer from the caller to pass in a string for the `data` to write. Also, probably worth adding an error check on `os.File.WriteString()`:

```go
// MustCreateTempFile creates a temporary file returning the path to the file.
func MustCreateTempFile(data string) string {
	f, err := ioutil.TempFile("", "fs-test")
	if err != nil {
		panic(fmt.Sprintf("failed to create temp file: %v", err))
	} else if _, err := f.WriteString(data); err != nil {
		panic(err)
	} else if err := f.Close(); err != nil {
		panic(err)
	}
	return name
}
```

Another approach I've been doing more lately is making the function a helper and naming it something like `CreateTempFileOrFail()`. That way it only fails the test instead of the whole program execution (although temp files are unlikely to fail).

```go
// CreateTempFileOrFail creates a temporary file returning the path to the file.
func MustCreateTempFile(t testing.TB, data string) string {
	tb.Helper()
	
	f, err := ioutil.TempFile("", "fs-test")
	if err != nil {
		tb.Fatalf("failed to create temp file: %v", err)
	} else if _, err := f.WriteString(data); err != nil {
		t.Fatal(err)
	} else if err := f.Close(); err != nil {
		t.Fatal(err)
	}
	return name
}
```

pkg/fs/fs_windows.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// RenameFileWithReplacement will replace any existing file at newpath with the contents
+// of oldpath.
+//
+// If no file already exists at newpath, newpath will be created using the contents
+// of oldpath. If this function returns successfully, the contents of newpath will
+// be identical to oldpath, and oldpath will be removed.
+func RenameFileWithReplacement(oldpath, newpath string) error {
+	if _, err := os.Stat(newpath); err == nil {
```
tmgordeeva: Would it be safer to rename the existing `newpath` to a temporary file before renaming oldpath, and then doing the removal last?

maxunt: I'm not exactly sure what you are asking, but using `RenameFileWithReplacement` is safe if you expect and are planning for any old contents of `newpath` to be truncated and lost. The `RenameFile` func exists as a tool to ensure you do not lose any data accidentally.

PR #14286, opened 2019-07-09
Fix issue where fields re-appear after DROP MEASUREMENT

e-dard:
Backport https://github.com/influxdata/influxdb/pull/14266

### DIFF
--- a/tests/server_delete_test.go
+++ b/tests/server_delete_test.go
@@ -281,6 +281,43 @@ func TestServer_Insert_Delete_1515777603585914810(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


tests/server_delete_test.go
```diff
@@ -281,14 +281,65 @@ func TestServer_Insert_Delete_1515777603585914810(t *testing.T) {
...
-	gotSeries, err := seriesFromShowSeries(result)
+	gotSeries, err := valuesFromShowQuery(result)
+	if err != nil {
+		panic(err)
+	}
+	return gotSeries
+}
+
+func mustGetFieldKeys(s Server) []string {
+	// Compare series left in index.
```
dgnorton: nit: don't think this comment is in the right spot

PR #14315, opened 2019-07-11
Update Go version to 1.12

gunnaraasen:
This PR updates the Go runtime used to build InfluxDB to Go 1.12.

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -9,7 +9,7 @@
 #      1: race enabled 64bit tests
 #      2: normal 32bit tests
 #      3: tsi build
-#      4: go 1.11
+#      4: go 1.12
 #      count: print the number of test environments
 #      *: to run all tests in parallel containers
 #
@@ -124,8 +124,8 @@ case $ENVIRONMENT_INDEX in
         rc=$?
         ;;
     4)
-        # go1.11
-        run_test_docker Dockerfile_build_ubuntu64_go1.11 test_64bit --test --junit-report
+        # go1.12
+        run_test_docker Dockerfile_build_ubuntu64_go1.12 test_64bit --test --junit-report
         rc=$?
         ;;
     "count")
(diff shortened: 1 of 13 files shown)


e-dard:
Wondering what broke in CI?

gunnaraasen:
This is failing because the build cache is [required in Go 1.12](https://golang.org/doc/go1.12#gocache) and go does not have permissions to create a `~/.cache` directory in the home directory. Seems like other people have set `GOCACHE=/tmp`. I'm not very familiar with our Jenkins pipeline. Where is the right place to set `GOCACHE`?

gunnaraasen:
@gshif resolved the CI error. @e-dard, can you take another look?

test.sh
```diff
@@ -124,8 +124,8 @@ case $ENVIRONMENT_INDEX in
         rc=$?
         ;;
     4)
-        # go1.11
-        run_test_docker Dockerfile_build_ubuntu64_go1.11 test_64bit --test --junit-report
+        # go1.12
+        run_test_docker Dockerfile_build_ubuntu64_go1.12 test_64bit --test --junit-report
```
mark-rushakoff: I'm not sure if this is needed now, or if we should be using 1.13beta for this. The previous stages should be testing with 1.12 already so this seems like duplicate work.

gunnaraasen: This line should be updated to 1.13beta along with the [release scripts](https://github.com/influxdata/influxdb/blob/1.8/releng/_go_versions.sh#L5). Go 1.13beta won't be out until August 1st though.

mark-rushakoff: 1.13beta1 has already been released: https://groups.google.com/forum/?pli=1#!topic/golang-dev/Qn7Gc9-y-nw

1.13beta images appear to be available on Docker hub.

gunnaraasen:
@mark-rushakoff I just pushed up an Appveyor config update.

PR #14316, opened 2019-07-11
feat(inspect): add verify-seriesfile tool to 2.x

### LINKED ISSUE #14279: Convert influx_inspect verify series file to 2.x
We need to take the `influx_inspect verify-seriesfile` tool from 1.x and convert it to a 2.x tool.

The original PR is here: https://github.com/influxdata/influxdb/pull/9770

The guts of the logic currently lives in the 1.8 branch here: https://github.com/influxdata/influxdb/tree/1.8/cmd/influx_inspect/verify/seriesfile

In the 2.x world, each tool should also have its own API. The API should live close to the package being verified/reported on. In this case I would expect to see a file `tsdb/series_verify.go`.

See these other examples of 1.x influx inspect tools that have been converted to the 2.x format:

https://github.com/influxdata/influxdb/pull/11964
https://github.com/influxdata/influxdb/pull/14221
https://github.com/influxdata/influxdb/pull/14233

Notice that they all provide an exported API, so that they can be called programmatically. Then we provide a binary that wraps the API so that we can call the tool with `influxd verify seriesfile` for example.

maxunt:
Closes #14279

Adds the influx_inspect verify-seriesfile tool to 2.x, called now with `influxd inspect verify-seriesfile`.

By default it will look for all series files in `$HOME/.influxdbv2/engine/_series`. If the `--series-file` flag is set, only that file is verified. The `--c` flag specifies the maximum number of concurrent workers used during verification. The `--db` flag that was available in 1.x has been removed.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influxd/inspect/inspect.go
+++ b/cmd/influxd/inspect/inspect.go
@@ -18,7 +18,11 @@ func NewCommand() *cobra.Command {
 		NewReportTSMCommand(),
 		NewVerifyTSMCommand(),
 		NewVerifyWALCommand(),
+<<<<<<< HEAD
 		NewReportTSICommand(),
+=======
+		NewVerifySeriesFileCommand(),
+>>>>>>> 26899d169... adds verify-seriesfile to 2.x
 	}
 
 	for _, command := range subCommands {
--- a/cmd/influxd/inspect/inspect.go
+++ b/cmd/influxd/inspect/inspect.go
@@ -18,11 +18,8 @@ func NewCommand() *cobra.Command {
 		NewReportTSMCommand(),
 		NewVerifyTSMCommand(),
 		NewVerifyWALCommand(),
-<<<<<<< HEAD
 		NewReportTSICommand(),
-=======
 		NewVerifySeriesFileCommand(),
->>>>>>> 26899d169... adds verify-seriesfile to 2.x
 	}
 
 	for _, command := range subCommands {
--- a/cmd/influxd/inspect/verify_series.go
+++ b/cmd/influxd/inspect/verify_series.go
@@ -1,9 +1,7 @@
 package inspect
 
 import (
-	"io/ioutil"
 	"os"
-	"path/filepath"
 	"runtime"
 
 	"github.com/influxdata/influxdb/logger"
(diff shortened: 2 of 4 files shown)


cmd/influxd/inspect/verify_series.go
```diff
@@ -0,0 +1,91 @@
...
+
+// NewVerifySeriesFileCommand returns a new instance of verifySeriesCommand
+// for execution of "influx_inspect verify-seriesfile".
+func NewVerifySeriesFileCommand() *cobra.Command {
+	verifySeriesCommand := &cobra.Command{
+		Use:   "verify-seriesfile",
+		Short: "Verifies the integrity of Series files",
+		Long: `Verifies the integrity of Series files.
+		Usage: influx_inspect verify-seriesfile [flags]
+			-dir <path>
```
e-dard: Since we only support a single series file per data directly, we don't need a `--dir` option

cmd/influxd/inspect/verify_series.go
```diff
@@ -0,0 +1,91 @@
...
+					Root data path.
+					Defaults to "` + os.Getenv("HOME") + `/.influxdbv2/engine/_series".
+			-series-file <path>
```
e-dard: series-file can default to `os.Getenv("HOME") + /.influxdbv2/engine/_series`

cmd/influxd/inspect/verify_series.go
```diff
@@ -0,0 +1,91 @@
...
+		_, err := v.VerifySeriesFile(VerifySeriesFlags.seriesFile)
+		return err
+	}
+
+	dbs, err := ioutil.ReadDir(VerifySeriesFlags.dir)
+	if err != nil {
+		return err
+	}
+
+	for _, db := range dbs {
```
e-dard: There are no dbs to iterate over

maxunt:
alchemy please merge

cmd/influxd/inspect/inspect.go
```diff
@@ -19,6 +19,7 @@ func NewCommand() *cobra.Command {
 		NewVerifyTSMCommand(),
 		NewVerifyWALCommand(),
 		NewVerifySeriesFileCommand(),
+    NewReportTSICommand(),
```
e-dard: This looks like it needs to be go formatted.

PR #14322, opened 2019-07-12
Update to Flux v0.36.2

nathanielc:


nathanielc:
@wolffcm Those tests are to emulate string manipulation the same way pandas does it. The idea being when we fix them we can say that we have rough feature parity with how pandas does string manipulation.

wolffcm:
Ah, ok.  I wasn't sure if it was a copy-paste mishap or something.

PR #14338, opened 2019-07-15
feat(http): handle flux errors that are returned from the query service

jsternberg:
Flux errors that are returned from the query service will be adapted to
an influxdb.Error and then will be written to HTTP correctly.

### DIFF
--- a/http/query_handler.go
+++ b/http/query_handler.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/codes"
 	"github.com/influxdata/flux/complete"
 	"github.com/influxdata/flux/csv"
 	"github.com/influxdata/flux/iocounter"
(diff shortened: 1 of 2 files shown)


http/query_handler.go
```diff
@@ -179,6 +175,51 @@ func (h *FluxHandler) handleQuery(w http.ResponseWriter, r *http.Request) {
...
+		codes.ResourceExhausted,
+		codes.FailedPrecondition,
+		codes.Aborted,
+		codes.OutOfRange,
+		codes.Unimplemented:
+		code = influxdb.EInvalid
+	case codes.PermissionDenied:
+		code = influxdb.EForbidden
+	case codes.Unauthenticated:
+		code = influxdb.EUnauthorized
```
wolffcm: Seems weird to match unauthenticated to unauthorized since they are different but easily confused.  I'm not sure what it should map to instead though.

jsternberg: It's because these are very closely related to the HTTP status codes where the 401 unauthorized basically means "you have not authenticated" while 403 forbidden is "you authenticated, but you don't have permission".

Issue #14340, opened 2019-07-15
Update influxdb flux sources to return flux errors


jsternberg:
The influxdb flux sources need to return flux errors so they can be interpreted correctly. The example that came up is with buckets that aren't found which should return a `NotFound` code.

jsternberg:
To clarify, this only applies to the `query/stdlib` code. It does not apply to `storage/reads`.

PR #14346, opened 2019-07-15
feat(query/stdlib): update influxdb sources to use flux error codes

jsternberg:
Update the influxdb sources to be annotated with flux error codes. This
will enable us to identify different response codes for these
operations.

#14340


jsternberg:
I'm not sure if the best way is to have this return flux errors which is what I did here or have it return influxdb errors. We may want to consider influxdb errors at some point, but I'm trying it this way first.

PR #14353, opened 2019-07-16
chore(tasks): remove old auth code and allow only token auth

### LINKED ISSUE #12409: task: require explicit authorization when creating task
In the HTTP handler for tasks today, we conditionally create a new authorization for the task if you are using a Session to make the request: https://github.com/influxdata/influxdb/blob/fe6ce24c8051a78fa952407e3ea8972e7fcd7855/http/task_service.go#L495-L526

If you are making an HTTP request to create a task, and in that request you're using a Token associated with an Authorization, and your create request doesn't explicitly set the task's token field, then we use the token (really the authorization ID) associated with the request.

https://github.com/influxdata/influxdb/blob/fe6ce24c8051a78fa952407e3ea8972e7fcd7855/task/platform_adapter.go#L143-L157
https://github.com/influxdata/influxdb/blob/fe6ce24c8051a78fa952407e3ea8972e7fcd7855/task/platform_adapter.go#L338-L354

Both of these behaviors are too implicit and will ultimately surprise users.

Instead, we should make the token a required field when creating a task, and cut out the conditional authorization creation and the current token grabbing.

Before we can make the token required, the UI needs to support setting a token when creating a task.

AlirieGray:
Closes #12409
Closes #14353

This PR makes it so that a token must be provided in order to create a task. 

- [x] Rebased/mergeable
- [ ] Tests pass


AlirieGray:
not ready to merge
blocked on data team adding ability to create tasks with a token

PR #14357, opened 2019-07-16
Subquery ordering with aggregates in descending mode was wrong

jsternberg:
The sort order of points when performing aggregates never took into
account if they were ascending or descending so when multiple series
were aggregated, it would ensure they were sorted in the correct order.
But it wouldn't reverse this order when descending was used.

Additionally, it seems that the iterator template and the iterator file
itself became out of sync. It seems the template was not reverted
correctly from a previously incorrect change and only the float type was
changed to the correct version and the tests used the float version.

Related to #14307.

### DIFF
--- a/query/iterator.gen.go
+++ b/query/iterator.gen.go
@@ -1114,28 +1114,28 @@ func (itr *floatReduceFloatIterator) reduce() ([]FloatPoint, error) {
 		rp.Aggregator.AggregateFloat(curr)
 	}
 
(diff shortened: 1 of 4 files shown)


jsternberg:
The tests were failing and it looks like some of the changes that went into 1.7 to fix subqueries had not been included in the 1.8 version so there's an extra test and removing one line of code that had been removed already in 1.7 included in this PR that wasn't included in the original 1.7 PR.

PR #14373, opened 2019-07-18
fix(tsdb): add missing err in SeriesPartition.Open

zhulongcheng:
Closes #

If `SeriesIndex.Recover` fails, the most recent series cannot be queried because the WAL segment has not yet been loaded into the cache.

Therefore, if the `SeriesIndex.Recover` fails, the `SeriesPartition` should not open successfully.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/tsdb/series_partition.go
+++ b/tsdb/series_partition.go
@@ -92,7 +92,7 @@ func (p *SeriesPartition) Open() error {
 
 		if err := p.index.Open(); err != nil {
 			return err
-		} else if p.index.Recover(p.segments); err != nil {
+		} else if err = p.index.Recover(p.segments); err != nil {
 			return err
 		}
 		return nil


zhulongcheng:
If `SeriesIndex.Recover` fails, the most recent series cannot be queried because the WAL segment has not yet been loaded into the cache.

Therefore, if the `SeriesIndex.Recover` fails, the `SeriesPartition` should not open successfully.

/cc @e-dard @benbjohnson

zhulongcheng:
Now, `SeriesIndex.Recover` always returns nil (even if there is an error in the method signature).
So there is nothing to worry about.
Just missing the err statement (`err := `).

tsdb/series_partition.go
```diff
@@ -92,7 +92,7 @@ func (p *SeriesPartition) Open() error {
 
 		if err := p.index.Open(); err != nil {
 			return err
-		} else if p.index.Recover(p.segments); err != nil {
+		} else if err := p.index.Recover(p.segments); err != nil {
```
e-dard: nit: this can just be `err = p.index.Recover(p.segments); err != nil {`

PR #14374, opened 2019-07-18
Use v4 UUID instead of v1 for UI IDs

ctavan:
I'm a co-author of the [uuid npm module](https://github.com/kelektiv/node-uuid) which is being used in some places within the code base of influxdb.

I'm currently trying to understand real-world use cases of [time-based UUIDs ("v1 UUIDs")](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_1_(date-time_and_MAC_address)).

I found three occurrence where v1 UUIDs were being used instead of v4 UUIDs and I was wondering if this case really requires the semantically much more complex v1 UUIDs or whether we could live equally well with purely random v4 UUIDs?

At least the test suite still passes after my changes, so apparently this doesn't seem to introduce any known regressions.

I'd be really curious to understand the motivation for choosing v1 over v4 UUIDs in the first place, so any feedback on this would be highly appreciated!

@chnn, as far as I can tell from the git history you introduced this in https://github.com/influxdata/influxdb/commit/201c4f4304f75d06e72da104dafe702888414905, I'd be really curious about your feedback!

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/src/workers/JobManager.ts
+++ b/ui/src/workers/JobManager.ts
@@ -68,7 +68,7 @@ class JobManager {
 
   // @ts-ignore
   private postJobRequest = async (type: string, payload: any) => {
-    const id = uuid.v1()
+    const id = uuid.v4()
     const deferred = new Deferred()
 
     this.jobs[id] = deferred
--- a/ui/src/workers/utils.ts
+++ b/ui/src/workers/utils.ts
@@ -4,7 +4,7 @@ import uuid from 'uuid'
 import {RequestMessage, ResponseMessage} from 'src/workers/types'
 
 export const postSuccessResponse = async (requestID: string, payload: any) => {
-  const id = uuid.v1()
+  const id = uuid.v4()
 
   await DB.put(id, payload)
 
@@ -18,7 +18,7 @@ export const postSuccessResponse = async (requestID: string, payload: any) => {
 }
 
 export const postErrorResponse = (requestID: string, err: Error) => {
-  const id = uuid.v1()
+  const id = uuid.v4()
 
   const response: ResponseMessage = {
     id,


chnn:
Hey @ctavan, I think there's no real reason v1 UUIDs were chosen here :) I ported the code over from the [Chronograf repo](https://github.com/influxdata/chronograf/blob/master/ui/src/worker/JobManager.ts), where there is a bit more history. I believe it is actually unused here, and should be deleted instead.

But in general, the places we use UUIDs (v1 or v4) probably don't need to be UUIDs at all. We use them only to differentiate client-side resources used by a single user and of a single type. If persisted, those IDs are entirely opaque to our server as well,  so our needs for a truly unique ID are quite relaxed. The uuid package was convenient though—thank you for that!

ctavan:
@chnn thanks for the quick reply!

> Hey @ctavan, I think there's no real reason v1 UUIDs were chosen here :) I ported the code over from the [Chronograf repo](https://github.com/influxdata/chronograf/blob/master/ui/src/worker/JobManager.ts), where there is a bit more history. I believe it is actually unused here, and should be deleted instead.

Does that mean that it is still being used in the [Chronograf repo](https://github.com/influxdata/chronograf/blob/master/ui/src/worker/JobManager.ts) so it would make sense for me to submit a PR over there instead?

Will you take care of removing it here or shall I create an issue for that?

> But in general, the places we use UUIDs (v1 or v4) probably don't need to be UUIDs at all. We use them only to differentiate client-side resources used by a single user and of a single type. If persisted, those IDs are entirely opaque to our server as well, so our needs for a truly unique ID are quite relaxed. The uuid package was convenient though—thank you for that!

At a first glance it sounds like there's nothing wrong with using UUIDs for the purpose that you describe here, I think they are a very good fit (even if you don't absolutely need the strong universal uniqueness properties)!

I just believe that `v4` UUIDs would be more suitable if you really only need some unique identifiers because they are much simpler in nature and come with less unexpected side effects.

chnn:
@ctavan I'll take care of removing the unused code from this repo. I just saw and merged your PR to chronograf. Thanks!

PR #14400, opened 2019-07-22
fix(storage): run snapshot before retention deletes

### LINKED ISSUE #14399: Snapshot before deleting for retention
Deletes block writes while they're deleting from the cache. Snapshotting first to clear the cache can reduce this contention.

tmgordeeva:
Deleting from the cache takes a lock which blocks writes. Snapshot to clear the
cache before deleting to reduce the lock contention.

Closes #14399

### DIFF
--- a/storage/engine.go
+++ b/storage/engine.go
@@ -531,6 +531,11 @@ func (e *Engine) DeleteBucket(orgID, bucketID platform.ID) error {
 
 // DeleteBucketRange deletes an entire bucket from the storage engine.
 func (e *Engine) DeleteBucketRange(orgID, bucketID platform.ID, min, max int64) error {
+	// Snapshot to clear the cache to reduce write contention.
+	if err := e.engine.WriteSnapshot(context.Background()); err != nil {
+		return err
+	}
+
 	e.mu.RLock()
 	defer e.mu.RUnlock()
 	if e.closing == nil {


storage/engine.go
```diff
@@ -531,6 +531,11 @@ func (e *Engine) DeleteBucket(orgID, bucketID platform.ID) error {
 
 // DeleteBucketRange deletes an entire bucket from the storage engine.
 func (e *Engine) DeleteBucketRange(orgID, bucketID platform.ID, min, max int64) error {
+	// Snapshot to clear the cache to reduce write contention.
+	if err := e.engine.WriteSnapshot(context.Background()); err != nil {
+		return err
+	}
+
```
dgnorton: Do we need to worry about multiple deletes causing too many snapshots? Not sure how expensive it is to write snapshots.

tmgordeeva: This is in the "delete a whole range from a bucket" code path called by retention and bucket deletion, so we shouldn't have to worry about that. We definitely don't want to call snapshot on every delete as it is expensive, but the expensive part of the snapshot at least won't block incoming writes.

storage/engine.go
```diff
@@ -531,6 +531,11 @@ func (e *Engine) DeleteBucket(orgID, bucketID platform.ID) error {
 
 // DeleteBucketRange deletes an entire bucket from the storage engine.
 func (e *Engine) DeleteBucketRange(orgID, bucketID platform.ID, min, max int64) error {
+	// Snapshot to clear the cache to reduce write contention.
```
e-dard: This isn't the right place to snapshot the cache. We may have thousands of buckets, and we don't want to check if we can snapshot the cache thousands of times. Instead, the call to snapshot the cache should probably come from within the retention enforcer.

storage/engine.go
```diff
@@ -531,6 +531,11 @@ func (e *Engine) DeleteBucket(orgID, bucketID platform.ID) error {
...
+	if err := e.engine.WriteSnapshot(context.Background()); err != nil {
```
e-dard: This is going to abort the deletion if we can't snapshot the cache. Sometimes we can't snapshot the cache for good reasons, but that doesn't mean we should abort a retention check / delete bucket range.

One concrete example of this is if the cache is already snapshotting, in which case `tsm1.ErrSnapshotInProgress` is going to be returned.

PR #14410, opened 2019-07-22
build(webpack): parcel to webpack

### LINKED ISSUE #14380: Switch from Parcel to Webpack
Parcel does not have WASM support.  Webpack does.  This is a blocker for any work that requires the Rust flux parser.

121watts:
Closes #14380 

### Pros
- We got our HMR back!  There's even more we can do (like preserving state in between hot-reloads!) 
- Similar if not faster dev and prod build times
- The wasm support we need

### Cons
- Configuration is 💩 

### Notes
For local development we will use `localhost:8080` (the route to the dev server).  All other folks that build for production can still use the lovely 9999.


ui/webpack.dev.ts
```diff
@@ -0,0 +1,24 @@
+export {}
```
chnn: what's this for?

ui/webpack.common.ts
```diff
@@ -0,0 +1,94 @@
...
+const GIT_SHA = require('child_process')
+  .execSync('git rev-parse HEAD')
+  .toString()
+
+const devMode = process.env.NODE_ENV !== 'production'
+
+module.exports = {
+  context: __dirname,
+  output: {
+    filename: '[name].bundle.js',
```
chnn: We should probably add a hash in production.

ui/src/shared/components/EditableName.tsx
```diff
@@ -33,7 +33,7 @@ interface State {
 @ErrorHandling
 class EditableName extends Component<Props, State> {
   public static defaultProps = {
-    hrefValue: '#',
+    hrefValue: 'javascript:void(0);',
```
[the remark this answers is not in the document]
121watts: The `#` was causing a route change to `/#` which, for some reason, causes the app to reload and re-render at the root of the project.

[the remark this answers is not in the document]
ebb-tide: shall we change it everywhere then?

ui/webpack.common.ts
121watts: Good catch added

PR #14420, opened 2019-07-23
fix(storage): move retention snapshot out of per bucket calls

### LINKED ISSUE #14399: Snapshot before deleting for retention

tmgordeeva:
Fixes for https://github.com/influxdata/influxdb/pull/14400

Closes #14399

### DIFF
--- a/storage/retention.go
+++ b/storage/retention.go
@@ -25,7 +25,7 @@ type Deleter interface {
 
 // A Snapshotter implementation can take snapshots of the entire engine.
 type Snapshotter interface {
-	WriteSnapshot(ctx context.Context) error
+	WriteSnapshot(ctx context.Context, status tsm1.CacheStatus) error
 }
 
 // A BucketFinder is responsible for providing access to buckets via a filter.
@@ -121,12 +121,10 @@ func (s *retentionEnforcer) expireData(ctx context.Context, buckets []*influxdb.
 	defer logEnd()
 
 	// Snapshot to clear the cache to reduce write contention.
-	span, _ := tracing.StartSpanFromContextWithOperationName(ctx, "retention_snapshot")
-	if err := s.Snapshotter.WriteSnapshot(context.Background()); err != nil {
-		if err != tsm1.ErrSnapshotInProgress {
-			logger.Warn("unable to snapshot cache before retention", zap.Error(err))
-			tracing.LogError(span, err)
-		}
+	span, _ := tracing.StartSpanFromContext(ctx)
+	if err := s.Snapshotter.WriteSnapshot(ctx, tsm1.CacheStatusRetention); err != nil && err != tsm1.ErrSnapshotInProgress {
+		logger.Warn("Unable to snapshot cache before retention", zap.Error(err))
+		tracing.LogError(span, err)
 	}
 	span.Finish()
 
--- a/storage/retention.go
+++ b/storage/retention.go
@@ -121,12 +121,9 @@ func (s *retentionEnforcer) expireData(ctx context.Context, buckets []*influxdb.
 	defer logEnd()
 
 	// Snapshot to clear the cache to reduce write contention.
-	span, _ := tracing.StartSpanFromContext(ctx)
 	if err := s.Snapshotter.WriteSnapshot(ctx, tsm1.CacheStatusRetention); err != nil && err != tsm1.ErrSnapshotInProgress {
 		logger.Warn("Unable to snapshot cache before retention", zap.Error(err))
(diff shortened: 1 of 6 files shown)


storage/retention.go
```diff
@@ -111,6 +120,16 @@ func (s *retentionEnforcer) expireData(ctx context.Context, buckets []*influxdb.
 	logger, logEnd := logger.NewOperation(ctx, s.logger, "Data deletion", "data_deletion")
 	defer logEnd()
 
+	// Snapshot to clear the cache to reduce write contention.
+	span, _ := tracing.StartSpanFromContextWithOperationName(ctx, "retention_snapshot")
```
jacobmarble: The correct way to trace `WriteSnapshot()` is to simply pass the context object `ctx` in to the `WriteSnapshot()` method. As I see things, this is not the right place to create a new context (`context.Background()`).

Looks like `Engine.WriteSnapshot()` [does create a span](https://github.com/influxdata/idpe/blob/8d40e7b3f788e2ef79d7fa7e2a996b10500a8467/vendor/github.com/influxdata/influxdb/tsdb/tsm1/engine.go#L780), so this should be simple.

storage/retention.go
```diff
@@ -111,6 +120,16 @@ func (s *retentionEnforcer) expireData(ctx context.Context, buckets []*influxdb.
...
+	if err := s.Snapshotter.WriteSnapshot(context.Background()); err != nil {
+		if err != tsm1.ErrSnapshotInProgress {
```
e-dard: `if err := s.Snapshotter.WriteSnapshot(context.Background()); err != nil && err != tsm1.ErrSnapshotInProgress {`

storage/retention.go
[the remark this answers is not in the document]
tmgordeeva: I'm uncertain if the name should be the same or different - if it's the same, will the rest of the trace have some indication this is a snapshot from retention and not the usual check?

However I also found that the call doing `Engine.compactCache <-t.C` updates `compactionTracker` as well which we'll also want here.

jacobmarble: FWIW, `s.Engine.DeleteBucketRange` is wrapped in a span because it doesn't take a context object (which needs to be fixed; all IO operations should be traced).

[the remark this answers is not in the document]
jacobmarble: No name necessary, see https://github.com/influxdata/influxdb/pull/14420/files#r306421742

storage/retention.go
```diff
@@ -121,12 +121,10 @@ func (s *retentionEnforcer) expireData(ctx context.Context, buckets []*influxdb.
 	defer logEnd()
 
 	// Snapshot to clear the cache to reduce write contention.
-	span, _ := tracing.StartSpanFromContextWithOperationName(ctx, "retention_snapshot")
-	if err := s.Snapshotter.WriteSnapshot(context.Background()); err != nil {
-		if err != tsm1.ErrSnapshotInProgress {
-			logger.Warn("unable to snapshot cache before retention", zap.Error(err))
-			tracing.LogError(span, err)
-		}
+	span, _ := tracing.StartSpanFromContext(ctx)
```
jacobmarble: Delete this line. `WriteSnapshot()` (now `writeSnapshot()`) creates it's own span, inheriting from the parent in `ctx`

storage/retention.go
```diff
@@ -121,12 +121,10 @@ func (s *retentionEnforcer) expireData(ctx context.Context, buckets []*influxdb.
...
+	if err := s.Snapshotter.WriteSnapshot(ctx, tsm1.CacheStatusRetention); err != nil && err != tsm1.ErrSnapshotInProgress {
+		logger.Warn("Unable to snapshot cache before retention", zap.Error(err))
+		tracing.LogError(span, err)
 	}
 	span.Finish()
```
jacobmarble: Delete this line.

tsdb/tsm1/engine.go
```diff
@@ -877,12 +892,10 @@ func (e *Engine) compactCache() {
 			span, ctx := tracing.StartSpanFromContextWithOperationName(context.Background(), "Engine.compactCache <-t.C")
```
jacobmarble: FYI we do create a wrapping span here to provide context in the UI. This function is run as a long-lived goroutine, and it wouldn't help us to create spans for an hours-long trace, so this is the right place to create a new trace, and give it an explicit, easy to read name. We sample every one of these traces.

It wouldn't hurt to change the operation name to `compact cache`.

PR #14421, opened 2019-07-23
Fix the http handler to not mislabel series as partial

jsternberg:
If a series was split by the encoder because of chunking and it was
reconstructed by the http handler, it would not reset the partial
indicator for the series to indicate if the series was still partial or
not. That meant that a result that returned more than the 10,000 values
in a single series with chunking disabled would say that the series was
partial, but it was not.

This fixes it so the handler now correctly sets the partial attribute of
the series to indicate if the series is still partial or not. This was
done when merging results, but was not done with series.

#14402


jsternberg:
Yea. This is because I make the PR on 1.7 first and then cherry pick to 1.8.

PR #14434, opened 2019-07-24
fix(ui/dashboards): prevent default on anchor tag

OfTheDelmer:
__Describe your proposed changes here.__
Fixes an issue where clicking on a dashboard name from the dashboards index caused an incorrect redirect due to default behavior on the anchor tag. 

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/src/tasks/components/TaskCard.tsx
+++ b/ui/src/tasks/components/TaskCard.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {PureComponent} from 'react'
+import React, {PureComponent, MouseEvent} from 'react'
 import {connect} from 'react-redux'
 import {withRouter, WithRouterProps} from 'react-router'
 
@@ -116,7 +116,9 @@ export class TaskCard extends PureComponent<Props & WithRouterProps> {
     )
   }
 
-  private handleNameClick = () => {
+  private handleNameClick = (e: MouseEvent) => {
+    e.preventDefault()
+
     this.props.onSelect(this.props.task)
   }
 
--- a/ui/src/tasks/components/TaskCard.tsx
+++ b/ui/src/tasks/components/TaskCard.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {PureComponent, MouseEvent} from 'react'
+import React, {PureComponent} from 'react'
 import {connect} from 'react-redux'
 import {withRouter, WithRouterProps} from 'react-router'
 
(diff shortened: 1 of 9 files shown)


OfTheDelmer:
@chnn yeah I noticed the resource card from `@influxdata/clockface` differs in that it doesn't use an anchor tag.  It looks like none of the components using `ResourceList.EditableName` specify an `hrefValue` so it could probably be removed. However, only `3` of `8` components currently have the `e.preventDefault()`. I'll go ahead and add it there. Maybe after we update to `0.0.17` we can swap these with the `@influxdata/clockface` implementation.

chnn:
@OfTheDelmer awesome. I do think we're using the `hrefValue` in a few places:

https://github.com/influxdata/influxdb/blob/efa9851685f1a52e5d24d5a0ffc40e692a240a89/ui/src/buckets/components/BucketCard.tsx#L48

https://github.com/influxdata/influxdb/blob/efa9851685f1a52e5d24d5a0ffc40e692a240a89/ui/src/variables/components/VariableCard.tsx#L63

I tried real quick adding `e.preventDefault` to the `EditableName` component itself, but that didn't seem to work. We could also try replacing the `a` with a `div` or `Link` in that component so that we don't have to `e.preventDefault()` everywhere it's used.

Also found a duplicate `EditableName` component that can probably be deleted if the one place it is used is updated:

https://github.com/influxdata/influxdb/blob/master/ui/src/shared/components/EditableName.tsx

OfTheDelmer:
@chnn didn't realize the `ResourceList.Name` also had this `hrefValue` implementation, but since none of the `ResourceList.EditableName` are currently using it I'll just swap it out with an `div` and see how that goes.

OfTheDelmer:
> I tried real quick adding e.preventDefault to the EditableName component itself, but that didn't seem to work. 

@chnn I experienced something similar this morning when trying to test the initial prevent default fix, but now find that running `yarn build` instead of `yarn start` reveals this does seem to fix this issue.

chnn:
Oo weird, I think @121watts was experiencing something similar yesterday.

Does the behavior exist when using the `hrefValue` with a `Link`? My hunch is that is the combination we should be using everywhere.

ui/src/tasks/components/TaskCard.tsx
```diff
@@ -116,9 +116,7 @@ export class TaskCard extends PureComponent<Props & WithRouterProps> {
     )
   }
 
-  private handleNameClick = (e: MouseEvent<HTMLAnchorElement>) => {
-    e.preventDefault()
-
+  private handleNameClick = () => {
```
OfTheDelmer: not able to create tasks so I'm not able to verify this change

OfTheDelmer: checked telegraf list instead and this still needed the prevent default.

PR #14444, opened 2019-07-25
fix(query): make show series exact cardinality count only distinct series

### LINKED ISSUE #14386: Incorrect behavior with `SHOW SERIES EXACT CARDINALITY` / `SHOW SERIES CARDINALITY FROM ...`
On 1.x with `tsi1`, `SHOW SERIES EXACT CARDINALITY`, as well as `SHOW SERIES CARDINALITY FROM <measurement>` returns the sum of the number of unique series for each shard, as opposed to the number of unique series __across__ all shards. This issue is actively being discussed and was filed mainly for documentation.

adamperlin:
Closes #14386 

Wrote this fix last week after discussions in storage re cardinality. 

Adds a `distinct` call to the rewritten queries for `SHOW SERIES EXACT CARDINALITY` and `SHOW SERIES CARDINALITY FROM <measurement>`.

See updated unit tests for the exact rewrites. 

This is the most straightforward fix to the current cardinality bug on 1.x `tsi1`, in which series keys are counted more than once if they are duplicated across shards. The only series cardinality query which is not affected by the bug is `SHOW SERIES CARDINALITY` (run across an entire database) since it is not rewritten and does bitmap level calculations or HLL if on a cluster. 

It is now confirmed that the cardinality bugs were behind https://github.com/influxdata/EAR/issues/992, so I want to continue the conversation and work towards a fix to avoid future customer confusion in this area.  

@stuartcarnie has mentioned that this change would likely cause a decrease in the performance of the affected queries. There could certainly be a more performant solution, and I would appreciate any suggestions!

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/query/statement_rewriter.go
+++ b/query/statement_rewriter.go
@@ -202,7 +202,16 @@ func rewriteShowSeriesCardinalityStatement(stmt *influxql.ShowSeriesCardinalityS
 
 	return &influxql.SelectStatement{
 		Fields: []*influxql.Field{
-			{Expr: &influxql.Call{Name: "count", Args: []influxql.Expr{&influxql.VarRef{Val: "_seriesKey"}}}, Alias: "count"},
+			{
+				Expr: &influxql.Call{
+					Name: "count",
+					Args: []influxql.Expr{&influxql.Call{
+						Name: "distinct",
+						Args: []influxql.Expr{&influxql.VarRef{Val: "_seriesKey"}},
+					}},
+				},
+				Alias: "count",
+			},
 		},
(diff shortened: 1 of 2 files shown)


e-dard:
@adamperlin this needs to be back ported to 1.7 branch if it's reviewed and accepted.

LeiRui:
Hi, I encountered this `show series exact cardinality` bug in version 1.4. The number returned by `show series exact cardinality` is bigger than the number of rows printed by `show series`.
Since version 1.4 supports `show series exact cardinality` clause, does this pr also fix version 1.4?

PR #14454, opened 2019-07-25
feat(tasks): add new executor prometheus metrics

### LINKED ISSUE #14336: add metrics to new executor
The new executor needs to have metrics around the rate of execution, the errors, and the length of time it takes to execute.

AlirieGray:
Closes #14336

This PR refactors the metrics for the new task executor, and adds the following new metrics:
1. Errors Counter
1. Run Duration
1. Percent Workers Busy
1. Percent Promise Queue Usage
1. Manual runs count
1. Resume runs count

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/task/backend/executor/task_executor.go
+++ b/task/backend/executor/task_executor.go
@@ -32,13 +32,8 @@ func MultiLimit(limits ...LimitFunc) LimitFunc {
 // LimitFunc is a function the executor will use to
 type LimitFunc func(*influxdb.Run) error
 
-type Metrics interface {
-	StartRun(influxdb.ID, time.Duration)
-	FinishRun(influxdb.ID, backend.RunStatus)
-}
-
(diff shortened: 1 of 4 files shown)


task/backend/executor/task_executor.go
```diff
@@ -407,6 +405,22 @@ func (w *worker) executeQuery(p *Promise) {
 	w.finish(p, backend.RunSuccess, runErr)
 }
 
+// RunsActive returns the current number of workers, which is equivalent to
+// the number of runs actively running
+func (e *TaskExecutor) RunsActive() float64 {
+	return float64(len(e.workerLimit))
```
lyondhill: For this one I would do the float conversion in the metrics system. Runs active (from the taskExecutor) should be a int or int64.

PR #14466, opened 2019-07-25
fix(task): push task backend logging messages back to debug

sebito91:
Revert extra changes from #14417

### DIFF
--- a/task/backend/executor/executor.go
+++ b/task/backend/executor/executor.go
@@ -142,11 +142,11 @@ func (p *syncRunPromise) finish(res *runResult, err error) {
 		close(p.ready)
 
 		if err != nil {
-			p.logger.Info("Execution failed to get result", zap.Error(err))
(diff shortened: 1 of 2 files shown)


task/backend/executor/executor.go
```diff
@@ -378,7 +378,7 @@ func (p *asyncRunPromise) finish(res *runResult, err error) {
 		} else if res.err != nil {
 			p.logger.Info("Got result with error", zap.Error(res.err))
```
tmgordeeva: Probably someone from tasks can answer, but why is this one `Info` while the other "Got result with error" is `Debug`? Anyway, it's as it is before so no reason to block this change, just was curious.

sebito91: Totally agreed, but to your point I'm just restoring to previous condition. I think this is a known configuration for `tigerteam`

sebito91:
alchemy please merge

PR #14470, opened 2019-07-25
feat(tsi1): add report-tsi tool to 2.x

### LINKED ISSUE #14484: Convert `influx_inspect reporttsi` to 2.x
See #10993.
Issue opened for tracking. 

The 1.x `influx_inspect reporttsi` tool needs to be ported to 2.x as `influxd inspect report-tsi`. The tool will report cardinalities for a given instance segmented by org, bucket, and measurement (`_m` tag). It will also have an API that will allow the tool to be run programmatically.

maxunt:
Resolves #14484.
Add the report-tsi tool to influxdb 2.x for debugging purposes

- [x] Rebased/mergeable
- [x] Tests pass


cmd/influxd/inspect/report_tsi1.go
```diff
@@ -0,0 +1,104 @@
...
+}{}
+
+// NewReportTsiCommand returns a new instance of Command with default setting applied.
+func NewReportTSICommand() *cobra.Command {
+	reportTSICommand := &cobra.Command{
+		Use:   "report-tsi",
+		Short: "Reports the cardinality of tsi files short",
+		Long:  `Reports the cardinality of tsi files long.`,
+		RunE:  RunReportTSI,
+	}
```
benbjohnson: I know it's like this in some other places but I find it easier if you just shorten the variable name to `cmd`.

cmd/influxd/inspect/report_tsi1.go
```diff
@@ -0,0 +1,104 @@
...
+
+	reportTSICommand.Flags().StringVar(&reportTSIFlags.Path, "path", os.Getenv("HOME")+"/.influxdbv2/engine", "Path to data engine. Defaults $HOME/.influxdbv2/engine")
+	reportTSICommand.Flags().StringVar(&reportTSIFlags.SeriesFilePath, "series-file", "", "Optional path to series file. Defaults /path/to/db-path/_series")
+	reportTSICommand.Flags().BoolVarP(&reportTSIFlags.ByMeasurement, "measurements", "m", false, "Segment cardinality by measurements")
+	// fs.BoolVar(&cmd.byTagKey, "tag-key", false, "Segment cardinality by tag keys (overrides `measurements`")
```
benbjohnson: Remove.

cmd/influxd/inspect/report_tsi1.go
```diff
@@ -0,0 +1,104 @@
...
+	report.Logger = log
+	report.ByMeasurement = reportTSIFlags.ByMeasurement
+	report.TopN = reportTSIFlags.TopN
+
+	if reportTSIFlags.Org != "" {
+		if orgID, err := influxdb.IDFromString(reportTSIFlags.Org); err != nil {
+			return err
+		} else {
+			report.OrgID = orgID
+		}
```
benbjohnson: I generally avoid `else` cases like this and either write it as:

```go
orgID, err := influxdb.IDFromString(reportTSIFlags.Org); 
if err != nil {
	return err
}
report.OrgID = orgID
```

or, since `err` is already defined (or I'll declare it as a named return), I'll do the shorter:

```go
if report.OrgID, err = influxdb.IDFromString(reportTSIFlags.Org); ; err != nil {
	return err
}
```

cmd/influxd/inspect/report_tsi1.go
```diff
@@ -0,0 +1,104 @@
...
+		} else {
+			report.BucketID = bucketID
+		}
+	}
+
+	// Run command with printing enabled
+	_, err = report.Run(true)
+	if err != nil {
+		return err
+	}
```
benbjohnson: Since the `err` only lives for this check, I'd inline it in the `if`:

```go
if _, err := report.Run(true); err != nil {
	return err
}
```

tsdb/tsi1/report.go
```diff
@@ -0,0 +1,374 @@
...
+	report.Stdout = os.Stdout
+
+	if report.SeriesDirPath == "" {
+		report.SeriesDirPath = filepath.Join(report.DataPath, "_series")
+	}
+
+	sFile := tsdb.NewSeriesFile(report.SeriesDirPath)
+
+	// TODO: do we actually want the seriesfile logging?
+	// sFile.WithLogger(report.Logger)
```
benbjohnson: Remove.

tsdb/tsi1/report.go
```diff
@@ -0,0 +1,374 @@
...
+// the cardinality of a bucket or org based on its influxdb.ID
+func (report *ReportCommand) Run(print bool) (*Summary, error) {
+	report.start = time.Now()
+	report.Stdout = os.Stdout
+
+	if report.SeriesDirPath == "" {
+		report.SeriesDirPath = filepath.Join(report.DataPath, "_series")
+	}
+
+	sFile := tsdb.NewSeriesFile(report.SeriesDirPath)
```
benbjohnson: We usually name the series file variable `sfile`. Not sure why but it's probably worth doing for consistency. :)

tsdb/tsi1/report.go
```diff
@@ -0,0 +1,374 @@
...
+	defer itr.Close()
+
+OUTER:
+	for {
+		name, err := itr.Next()
+		if err != nil {
+			return err
+		} else if name == nil {
+			break OUTER
+		}
```
benbjohnson: I would remove the `OUTER` label and just do a `return nil` here since there's nothing happening after the `for` loop (except the return).

benbjohnson: Also, I would move everything after this `if` block to a function called `calculateMeasurementCardinalities(name []byte)`. The loop body is pretty long and you'll be able to take advantage of `defer` for the `sitr.Close()` below.

tsdb/tsi1/report.go
```diff
@@ -0,0 +1,374 @@
...
+		} else if report.BucketID != nil && *report.BucketID != bucket {
+			continue
+		}
+
+		sitr, err := idx.MeasurementSeriesIDIterator(name)
+		if err != nil {
+			return err
+		} else if sitr == nil {
+			continue
+		}
```
benbjohnson: This should have a `defer sitr.Close()` to ensure that it's not left open on accident.

tsdb/tsi1/report.go
```diff
@@ -0,0 +1,374 @@
...
+		// initialize total cardinality tracking struct for this bucket
+		if c, ok := report.byOrgBucket[org][bucket]; !ok {
+			bucketCard = &cardinality{name: []byte(bucket.String())}
+			report.byOrgBucket[org][bucket] = bucketCard
+		} else {
+			bucketCard = c
+		}
+
+		var e tsdb.SeriesIDElem
+		for e, err = sitr.Next(); err == nil && e.SeriesID.ID != 0; e, err = sitr.Next() {
```
benbjohnson: This is probably cleaner like this:

```go
for {
	e, err := sitr.Next()
	if err != nil {
		return err
	} else if e.SeriesID.ID == 0 {
		break
	}

	//...
}
```

tsdb/tsi1/report.go
```diff
@@ -0,0 +1,374 @@
...
+			id := e.SeriesID.ID
+
+			if id > math.MaxUint32 {
+				panic(fmt.Sprintf("series ID is too large: %d (max %d). Corrupted series file?", e.SeriesID, uint32(math.MaxUint32)))
+			}
```
benbjohnson: I think an error is more appropriate to return here than panic.

tsdb/tsi1/report.go
```diff
@@ -0,0 +1,374 @@
...
+
+			// add cardinality to bucket
+			bucketCard.add(id)
+
+			// retrieve tags associated with series id so we can get
+			// associated measurement
+			_, tags := report.sfile.Series(e.SeriesID)
+			if len(tags) == 0 {
+				panic(fmt.Sprintf("empty series key"))
+			}
```
benbjohnson: I think an error is more appropriate to return here than panic.

tsdb/tsi1/report.go
```diff
@@ -0,0 +1,374 @@
...
+
+				mCard.add(id)
+			}
+		}
+
+		sitr.Close()
+
+		if err != nil {
+			return err
+		}
```
benbjohnson: The `sitr.Close()` & `err` check aren't needed here if they're handled with a defer as mentioned above and the `sitr.Next()` handles the check.

cmd/influxd/inspect/report_tsi1.go
```diff
@@ -0,0 +1,98 @@
...
+	ByMeasurement bool
+	byTagKey      bool // currently unused
+}{}
+
+// NewReportTsiCommand returns a new instance of Command with default setting applied.
+func NewReportTSICommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "report-tsi",
+		Short: "Reports the cardinality of tsi files short",
+		Long:  `Reports the cardinality of tsi files long.`,
```
e-dard: I'd like to see some real documentation here. For a guide on detail and verbosity take a look at this one: 

https://github.com/influxdata/influxdb/blob/bbc37b4c6814d428d2e423fbac777a9139640be5/cmd/influxd/inspect/report_tsm.go#L29-L57

cmd/influxd/inspect/report_tsi1.go
```diff
@@ -0,0 +1,98 @@
...
+		RunE:  RunReportTSI,
+	}
+
+	cmd.Flags().StringVar(&reportTSIFlags.Path, "path", os.Getenv("HOME")+"/.influxdbv2/engine", "Path to data engine. Defaults $HOME/.influxdbv2/engine")
```
e-dard: I think that this should be the path to the actual index. So it should default to `os.Getenv("HOME")+"/.influxdbv2/engine/index`

cmd/influxd/inspect/report_tsi1.go
```diff
@@ -0,0 +1,98 @@
...
+	cmd.Flags().StringVar(&reportTSIFlags.SeriesFilePath, "series-file", "", "Optional path to series file. Defaults /path/to/db-path/_series")
```
e-dard: This should default to `os.Getenv("HOME")+"/.influxdbv2/engine/_series`

cmd/influxd/inspect/report_tsi1.go
```diff
@@ -0,0 +1,98 @@
...
+	cmd.Flags().BoolVarP(&reportTSIFlags.ByMeasurement, "measurements", "m", false, "Segment cardinality by measurements")
+	cmd.Flags().IntVarP(&reportTSIFlags.TopN, "top", "t", 0, "Limit results to top n")
+	cmd.Flags().StringVarP(&reportTSIFlags.Bucket, "bucket", "b", "", "If bucket is specified, org must be specified")
```
e-dard: nit: `bucket_id` might be preferable. Also be more explicit about what needs to be provided. `a bucket id in as a base-16 string`.

cmd/influxd/inspect/report_tsi1.go
```diff
@@ -0,0 +1,98 @@
...
+	cmd.Flags().StringVarP(&reportTSIFlags.Org, "org", "o", "", "Org to be reported")
```
e-dard: Same thing here: `org_id`

cmd/influxd/inspect/report_tsi1.go
```diff
@@ -0,0 +1,98 @@
...
+
+	return cmd
+}
+
+// RunReportTSI executes the run command for ReportTSI.
+func RunReportTSI(cmd *cobra.Command, args []string) error {
+	// set up log
+	config := logger.NewConfig()
+	config.Level = zapcore.InfoLevel
+	log, err := config.New(os.Stderr)
```
e-dard: Do we log to `Stderr` on the other tools?

cmd/influxd/inspect/report_tsi1.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return err
+	}
+
+	// if path is unset, set to $HOME/.influxdbv2/engine"
+	if reportTSIFlags.Path == "" {
+		reportTSIFlags.Path = path.Join(os.Getenv("HOME"), ".influxdbv2/engine")
```
e-dard: Does this not get set for you when the flags are initialising?

tsdb/tsi1/report.go
```diff
@@ -0,0 +1,375 @@
...
+
+// Run runs the report-tsi tool which can be used to find the cardinality
+// any org or bucket. Run returns a *ReportTSISummary, which contains maps for finding
+// the cardinality of a bucket or org based on its influxdb.ID
+func (report *ReportCommand) Run(print bool) (*Summary, error) {
+	report.start = time.Now()
+
+	report.Stdout = os.Stdout
+
+	if report.SeriesDirPath == "" {
```
e-dard: I think we can just have the `_series` path be explicit rather than inferring it.

tsdb/tsi1/report.go
```diff
@@ -0,0 +1,375 @@
...
+		report.SeriesDirPath = filepath.Join(report.DataPath, "_series")
+	}
+
+	sfile := tsdb.NewSeriesFile(report.SeriesDirPath)
+
+	if err := sfile.Open(context.Background()); err != nil {
+		report.Logger.Error("failed to open series")
```
e-dard: We don't need to log this if the error is being returned.

tsdb/tsi1/report.go
```diff
@@ -0,0 +1,375 @@
...
+
+		// retrieve tags associated with series id so we can get
+		// associated measurement
+		_, tags := report.sfile.Series(e.SeriesID)
+		if len(tags) == 0 {
+			return fmt.Errorf("series ID has empty key: %d", e.SeriesID)
+		}
+
+		// measurement name should be first tag
+		mName := string(tags[0].Value)
```
e-dard: This is right, but let's verify by checking `bytes.Equal(tags[0].Key, models.MeasurementTagKeyBytes)` and returning an error if that's not the case?

maxunt:
alchemy please merge

e-dard:
@maxunt we don't have alchemy on InfluxDB yet :)

Issue #11531, opened 2019-01-24
pkg/estimator/hll: flaky test failure


mark-rushakoff:
Saw this failure once over at https://circleci.com/gh/influxdata/influxdb/22933:

```
--- FAIL: TestPlus_Marshal_Unmarshal_Count (4.91s)
    hll_test.go:594: error was 2e+06 for estimation 0 and true cardinality 2000000
FAIL
FAIL	github.com/influxdata/influxdb/pkg/estimator/hll	4.949s
```

e-dard:
don't close

mark-rushakoff:
Still don't close.

PR #14485, opened 2019-07-26
fix: services/httpd: parse correctly Accept header with extra test cases

### LINKED ISSUE #14296: The HTTP "Accept" header is interpreted in a non-standard way
InfluxDB does not seem to interpret the [Accept HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) according to the HTTP specification. It seems to return a MessagePack encoded response only when the header has the exact value `application/x-msgpack`. Adding a second accepted format to the header makes influxdb switch back to returning JSON.

__Steps to reproduce:__

1. Install influx
2. From a terminal, run 
```sh
curl -G 'http://localhost:8086/query' --data-urlencode 'q=SHOW databases' --header "Accept: application/x-msgpack, text/plain"
```

__Expected behavior:__
A MessagePack response is returned.

__Actual behavior:__
A JSON response is returned.

__Environment info:__

* InfluxDB version: InfluxDB vv1.7.6 (git: master 01c8dd416270f424ab0c40f9291e269ac6921964)

jsign:
Closes #14296

Until the standard lib decides to include a standard function to correctly parse Accept headers, we can rely on [goautoneg](https://github.com/golang/go/issues/19307#issuecomment-511133691) library which is a Github port used by Prometheus.

I also added sub-tests as Table Tests for multiples scenarios around this problem.

Adding a new dependency for this amount of code may hurt a bit, so maybe we could add the code directly too.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/services/httpd/accept.go
+++ b/services/httpd/accept.go
@@ -1,0 +1,80 @@
+// This file is an adaptation of https://github.com/markusthoemmes/goautoneg.
+// The copyright and license header are reproduced below.
+//
+//    Copyright [yyyy] [name of copyright owner]
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+
+package httpd
+
+import (
+	"mime"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+// accept is a structure to represent a clause in an HTTP Accept Header.
+type accept struct {
+	Type, SubType string
+	Q             float64
+	Params        map[string]string
+}
+
+// parseAccept parses the given string as an Accept header as defined in
+// https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.1.
+// Some rules are only loosely applied and might not be as strict as defined in the RFC.
+func parseAccept(headers []string) []accept {
+	var res []accept
+	for _, header := range headers {
+		parts := strings.Split(header, ",")
+		for _, part := range parts {
+			mt, params, err := mime.ParseMediaType(part)
+			if err != nil {
+				continue
+			}
+
+			accept := accept{
+				Q:      1.0, // "[...] The default value is q=1"
+				Params: params,
+			}
+
+			// A media-type is defined as
+			// "*/*" | ( type "/" "*" ) | ( type "/" subtype )
+			types := strings.Split(mt, "/")
+			switch {
+			// This case is not defined in the spec keep it to mimic the original code.
+			case len(types) == 1 && types[0] == "*":
+				accept.Type = "*"
+				accept.SubType = "*"
+			case len(types) == 2:
+				accept.Type = types[0]
+				accept.SubType = types[1]
+			default:
+				continue
+			}
+
+			if qVal, ok := params["q"]; ok {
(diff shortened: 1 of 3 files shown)


jsign:
The failed test case seems to be a [reported flaky test](https://github.com/influxdata/influxdb/issues/11531)

jsign:
@russorat , I rebased it to `1.8` to keep it clean. 
Is `1.8` OK for merging or do you prefer on `master`?

jsign:
@jsternberg, pushed changes in the direction you mentioned.

services/httpd/response_writer.go
```diff
@@ -25,19 +25,27 @@ type ResponseWriter interface {
...
-	case "application/x-msgpack":
-		w.Header().Add("Content-Type", "application/x-msgpack")
-		rw.formatter = &msgpackFormatter{}
-	case "application/json":
-		fallthrough
-	default:
-		w.Header().Add("Content-Type", "application/json")
-		rw.formatter = &jsonFormatter{Pretty: pretty}
+
+	acceptHeaders := parseAccept(r.Header.Get("Accept"))
```
jsternberg: Can you change this to `r.Header` as the argument and change `parseAccept` to take a `[]string` value? I think it's possible for someone to do this:

```
Accept: application/csv
Accept: application/x-msgpack
```

services/httpd/accept.go
```diff
@@ -0,0 +1,68 @@
+/* This file is a copy of github.com/markusthoemmes/goautoneg
```
jsternberg: Can you copy the full license into this file? Something like:

```
// This file is an adaptation of https://github.com/markusthoemmes/goautoneg.
// The copyright and license header are reproduced below.
//
// (copy of the license)
//
// (link to the license)
```

services/httpd/response_writer.go
```diff
@@ -25,19 +25,27 @@ type ResponseWriter interface {
...
+	for _, accept := range acceptHeaders {
+		switch accept.Type + "/" + accept.SubType {
```
jsternberg: It seems to me this code handles wildcards incorrectly. If I am understanding the spec correctly, a `*` would match any one of these. Is this correct? If it is, I have some recommendations for how we can change the code. If it isn't, then we can skip past that.

services/httpd/response_writer_test.go
```diff
@@ -19,63 +19,77 @@ import (
 )
 
 func TestResponseWriter_CSV(t *testing.T) {
-	header := make(http.Header)
-	header.Set("Accept", "text/csv")
-	r := &http.Request{
-		Header: header,
-		URL:    &url.URL{},
+	tableTest := []struct {
```
jsternberg: I'd like some wildcard tests here.

services/httpd/response_writer.go
jsign: Uhm, I agree. Makes sense.

jsign:
@jsternberg, I applied your suggestions, and also switched to performance mode since the code had many allocations that could be avoided. And considering this is a hot-path, deserved some extra attention.

Some points about the changes (apart from your comments):
- make `csvFormatter` stateless
- all known content types and formatters are defined as vars, so to a ton of allocations. This is the reason why I made `csvFormatter` stateless (since it wasn't).
- avoided string concats to save extra allocations (that's why `full` exist apart from splited parts)
- added `*` test case
- in case of `*/*` the switch isn't biased to a definite option. This is strictly correct (since any option complies to request), but maybe another default is wanted.
- as can be seen, adding future content types is much less verbose

jsternberg:
Sorry I haven't had time to review this again. I will review by end of day tomorrow.

dgnorton:
@jsternberg can you review this for the 1.8 release coming up?

jsternberg:
I reverted the changes to the formatters and modified the creation of those formatters to a factory method. I think that the state is actually used by the code so a global instance doesn't work, but I also don't remember well enough to know for certain so I don't want those to change.

I am going to wait for the tests and run some myself and then merge this.

PR #14488, opened 2019-07-28
fix(http): add orgID param to /write endpoint

### LINKED ISSUE #13906: InfluxDB 2.0: The "org" query param should be same for /query and /write  [area/api]
The `/query` and the `/write` are critical parts of API 2.0 and should be friendly for users.

The `/write` endpoint has defined `org` parameter as:

```yaml
- in: query
   name: org
   description: specifies the destination organization for writes
   required: true
   schema:
      type: string
      description: all points within batch are written to this organization.
```

The `org` parameter could be `Organization.name` or `Organization.id` - [see sources](https://github.com/influxdata/influxdb/blob/master/http/write_handler.go#L137).

But the `/query` endpoint has define `org` and `orgID` parameter:
```yaml
- in: query
   name: org
   description: specifies the name of the organization executing the query; if both orgID and org are specified, orgID takes precedence.
   schema:
      type: string
- in: query
   name: orgID
   description: specifies the ID of the organization executing the query; if both orgID and org are specified, orgID takes precedence.
   schema:
      type: string
```

For consistency in API should be better that  the `/query` endpoint will be same as `/write` endpoint.

It will be also useful for clients configuration. For example both of this [Spring actuator configuration](https://github.com/bonitoo-io/influxdb-client-java/tree/master/spring#influxdb2-auto-configuration) will be valid:

```yaml
spring.influx2.url=http://localhost:9999
spring.influx2.org=my-org
spring.influx2.bucket=my-bucket
spring.influx2.token=my-token-123
spring.influx2.username=my-user
```

```yaml
spring.influx2.url=http://localhost:9999
spring.influx2.org=03d2ebc848c1e000
spring.influx2.bucket=my-bucket
spring.influx2.token=my-token
spring.influx2.username=my-user
```

Maybe it is also related to https://github.com/influxdata/influxdb-client-js/pull/115#issuecomment-490310352 from @timhallinflux

zhulongcheng:
Closes #13906

### DIFF
--- a/cmd/influxd/launcher/launcher_helpers.go
+++ b/cmd/influxd/launcher/launcher_helpers.go
@@ -131,7 +131,7 @@ func (tl *TestLauncher) OnBoardOrFail(tb testing.TB, req *platform.OnboardingReq
 // WriteOrFail attempts a write to the organization and bucket identified by to or fails if there is an error.
 func (tl *TestLauncher) WriteOrFail(tb testing.TB, to *platform.OnboardingResults, data string) {
 	tb.Helper()
-	resp, err := nethttp.DefaultClient.Do(tl.NewHTTPRequestOrFail(tb, "POST", fmt.Sprintf("/api/v2/write?org=%s&bucket=%s", to.Org.ID, to.Bucket.ID), to.Auth.Token, data))
+	resp, err := nethttp.DefaultClient.Do(tl.NewHTTPRequestOrFail(tb, "POST", fmt.Sprintf("/api/v2/write?orgID=%s&bucketID=%s", to.Org.ID, to.Bucket.ID), to.Auth.Token, data))
 	if err != nil {
 		tb.Fatal(err)
 	}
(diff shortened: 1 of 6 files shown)


bednar:
I don't think so that this is a good solution. There should be a parameter that accept both variant: `name` and `id`.

For example: If i have a configuration for IOT devices:

```
influx2.url=https://cloud.influxdata/
influx2.org=ReacherGmBH
influx2.token=my-secret
```

the `influx.org` configuration could be _Organization name_ or _Organization id_ and InfluxDB 2.0 should be able to work this way.

kelwang:
cc @jademcgough , @ebb-tide

bednar:
Hi all, 

I prepared PR that introduce interchangeability between ID and Name - see https://github.com/influxdata/influxdb/pull/14540.

I think the #14540 is better solution and it's useful for building tools over the InfluxDB 2.0.

zhulongcheng:
Thanks so much @kelwang @bednar @ebb-tide 

---

There is one question is:
**how to know the org param is an `id` or a `name`?**

Now, if the param can be parsed as an int, then it is an `id`, otherwise, it is a `name`.

This is not entirely correct, because some `name` can be parsed as int.
Like `aaaaaaaaaaaaaaa`, it is a name, but it is treated as an `id`, 
`/write?org=aaaaaaaaaaaaaaa` will return an error (org not found)

InfluxDB is an OSS project, users may name org to any string.
I think this problem should be solved in the system design, rather than expecting that these situations will not occur.

zhulongcheng:
#14540 will fix the issue

PR #14495, opened 2019-07-29
feat(http): gzip compress the query CSV response.

goller:
If client requests with `Accept-Encoding: gzip`, then
compress response and return with `Content-Encoding: gzip`.

This increases the server-side load by about 12%, but saves network bandwidth.

```
benchmark                     old ns/op     new ns/op     delta
Benchmark_Query_no_gzip-4     123609        137885        +11.55%

benchmark                     old allocs     new allocs     delta
Benchmark_Query_no_gzip-4     149            150            +0.67%

benchmark                     old bytes     new bytes     delta
Benchmark_Query_no_gzip-4     14297         15205         +6.35%
```

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/http/query_handler_test.go
+++ b/http/query_handler_test.go
@@ -13,6 +13,7 @@ import (
 	"regexp"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/influxdata/flux"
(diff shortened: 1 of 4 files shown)


http/query_handler_test.go
```diff
@@ -431,3 +433,235 @@ func TestFluxHandler_PostQuery_Errors(t *testing.T) {
...
+		t.Errorf("unexpected difference in identity and compressed bodies:\n%s\n%s", string(identityBody), string(gzippedBody))
+	}
+}
+
+func Benchmark_Query_no_gzip(b *testing.B) {
+	benchmarkQuery(b, true)
+}
+
+func Benchmark_Query_gzip(b *testing.B) {
+	benchmarkQuery(b, false)
```
chnn: I'm wonder if nginx could compress more efficiently.

PR #14496, opened 2019-07-29
fix(component/TemplateBrowserList): sort list of templates alphabetically

### LINKED ISSUE #14465: The list of templates should be sorted alphabetically
this list should be alphabetical:

image

desa:
Closes #14465

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/src/templates/components/createFromTemplateOverlay/TemplateBrowserList.tsx
+++ b/ui/src/templates/components/createFromTemplateOverlay/TemplateBrowserList.tsx
@@ -1,6 +1,6 @@
 // Libraries
 import React, {PureComponent} from 'react'
-import _ from 'lodash'
+import {get, orderBy} from 'lodash'
 
 // Components
 import {DapperScrollbars} from '@influxdata/clockface'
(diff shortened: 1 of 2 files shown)


ui/src/templates/components/createFromTemplateOverlay/TemplateBrowserList.tsx
```diff
@@ -23,16 +23,18 @@ class TemplateBrowser extends PureComponent<Props> {
...
-            key={t.id}
-            template={t}
-            label={t.meta.name}
-            onClick={onSelectTemplate}
-            testID={`template--${t.meta.name}`}
-            selected={_.get(selectedTemplateSummary, 'id', '') === t.id}
-          />
-        ))}
+        {templates
+          .sort((a, b) => (b.meta.name < a.meta.name ? 1 : -1))
```
121watts: probably want a `string.toLocaleLowerCase` when evaluating these string.  also, we have a component which does this for you.  it's called `SortingHat`.  if you don't want to use that, might I suggest using lodash's `sortBy` or `orderBy`.  It just reads nicer.

Issue #14482, opened 2019-07-26
Move API client generation into influxdb


chnn:
See [RFC](https://github.com/influxdata/idpe/pull/4116) for motivation.

- Generate an API client and types from the `http/swagger.yml` spec and [oats](https://github.com/chnn/oats)
- Replace all usages of the `@influxdata/influx` package to use this generated code

### Remaining work

- [ ] Use locally generate API utils
  - [x] `client.auth` (https://github.com/influxdata/influxdb/pull/14611)
  - [x] `client.authorizations` (https://github.com/influxdata/influxdb/pull/14831)
  - [x] `client.buckets`
  - [x] `client.dashboards` (https://github.com/influxdata/influxdb/pull/16438)
  - [x] `client.labels` (https://github.com/influxdata/influxdb/pull/16361)
  - [x] `client.links`
  - [x] `client.organizations`
  - [x] `client.queries`
  - [ ] `client.scrapers`
  - [ ] `client.setup`
  - [x] `client.tasks` (https://github.com/influxdata/influxdb/pull/16353)
  - [ ] `client.telegrafConfigs`
  - [ ] `client.users`
  - [x] `client.variables` (https://github.com/influxdata/influxdb/pull/16400)
  - [x] `client.write`
  - [ ] `client.templates`
  - [x] `client.checks`
  - [x] `client.notificationRules`

PR #14500, opened 2019-07-29
refactor(ui): inline API type generation

chnn:
Previously we used a mixed of handwritten types and generated types from `@influxdata/influx`. This PR replaces some of these types with types generated alongside the rest of the UI code. 

The changed types are those related to:

- dashboards
- views
- view properties
- cells
- checks
- notificationRules
- buckets

I plan to eventually convert all API types in a similar fashion.

See #14482 and [RFC](https://github.com/influxdata/idpe/pull/4116) for more context.


http/swagger.yml
```diff
@@ -8775,24 +8912,26 @@ components:
...
-                  mapping:
-                    greater: "#/components/schemas/GreaterThreshold"
-                    lesser: "#/components/schemas/LesserThreshold"
-                    range: "#/components/schemas/RangeThreshold"
+                $ref: "#/components/schemas/ThresholdCheckThreshold"
+    ThresholdCheckThreshold:
+      oneOf:
+        - $ref: "#/components/schemas/GreaterThreshold"
+        - $ref: "#/components/schemas/LesserThreshold"
+        - $ref: "#/components/schemas/RangeThreshold"
```
kelwang: ThresholdCheckThreshold sounds a little off to me. We would call it ThresholdCheckDetail or ThresholdCheckConfig

For the struct, why don't you make it just one simple object
```go
type ThresholdConfig struct {
	AllValues bool                    `json:"allValues"`
	Level     notification.CheckLevel `json:"level"`
	LowerBound       *float64                `json:"lowerBound,omitempty"`
	UpperBound       *float64                `json:"upperBound,omitempty"`
}
```
So for greater, fill the lowerBound; for lesser fill the upperBound; for range fill both. For within false range, simply reverse the upperBound and lowerBound.

http/swagger.yml
```diff
@@ -8880,6 +9019,18 @@ components:
...
+        - id
+        - notifyEndpointID
+        - orgid
+        - status
+        - name
+        - offset
+        - limitEvery
+        - limit
+        - tagRules
+        - description
```
kelwang: notifyEndpointID , description, offset, Limit, LimitEvery are not required.

http/swagger.yml
```diff
@@ -8978,7 +9126,11 @@ components:
       allOf:
         - $ref: "#/components/schemas/NotificationRuleBase"
         - type: object
+          required: [type, channel, messageTemplate]
```
kelwang: channel is not required, according to @ebb-tide

http/swagger.yml
```diff
@@ -8987,7 +9139,11 @@ components:
       allOf:
         - $ref: "#/components/schemas/NotificationRuleBase"
         - type: object
+          required: [type, subjectTemplate, bodyTemplate, to]
```
kelwang: bodyTemplate is not required

http/swagger.yml
chnn: Thanks for catching these, will update.

http/swagger.yml
chnn: I agree it's a weird name. How about just `Threshold`?

My preference would be to make them separate but similar models, so that it is less ambiguous how each variant should be handled. Doing so helps us on the front end, since that style generates a [discriminated union TypeScript definition](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions) that provides us type checking benefits (it becomes a compile time error to use the a variant in the the wrong way). 

In any event, I think we would still need a `type` field with value either "lesser", "greater", or "range" in order to distinguish how these models are handled on the front end.

ui/src/alerting/components/CheckCard.tsx
```diff
@@ -47,10 +47,8 @@ const CheckCard: FunctionComponent<Props> = ({
   const onClone = () => {}
 
   const onToggle = () => {
-    const status =
-      check.status == CheckBase.StatusEnum.Active
-        ? CheckBase.StatusEnum.Inactive
-        : CheckBase.StatusEnum.Active
+    const status = check.status === 'active' ? 'inactive' : 'active'
```
ebb-tide: I really like that we can compare enum values to their string values.

chnn: Me too! It's a small thing but somehow so nice

ui/src/buckets/actions/index.ts
```diff
@@ -105,12 +109,13 @@ export const createBucket = (bucket: Bucket) => async (
...
-    })
+    const resp = await api.postBucket({data: {...bucket, orgID: org.id}})
 
-    dispatch(addBucket(createdBucket))
+    if (resp.status !== 201) {
+      throw new Error(resp.data.message)
+    }
+
+    dispatch(addBucket(resp.data))
     dispatch(checkBucketLimits())
```
ebb-tide: I think this line (and other checking of asset limits) needs to run both in the case of resp.status=200 and otherwise. can we move them above the error?

chnn: It looks like the only possible responses are `201` and `500`:

```
    post:
      operationId: PostBuckets
      tags:
        - Buckets
      summary: Create a bucket
      parameters:
          - $ref: '#/components/parameters/TraceSpan'
      requestBody:
        description: bucket to create
        required: true
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/Bucket"
      responses:
        '201':
          description: Bucket created
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Bucket"
        default:
          description: unexpected error
          content:
            application/json:
              schema:
```

ebb-tide: yes and we should check bucket limits in both cases.

chnn: ohh I see. So if it returns a `500`, that's because it may have hit the rate limit? And `dispatch(checkBucketLimits())` would be what eventually shows an error to the user?

chnn: I guess I'm a little fuzzy on why we need to check the rate limits for buckets if we tried to add a bucket but failed. I was thinking we'd have the same exact number of buckets in that case, so checking the rate limit would be unnecessary?

ebb-tide: The requirements were to display a static error banner on top of buckets list- rather than just a notification- so yes if you're at the last bucket you can create you succeed in creation and then check your limits. I apparently designed it to only check limits in case of create rather than on get, so I was relying on create to display the warning. I'll add a checkLimits on get to make it work in either case.

chnn: ah, that makes sense. Ok, thank you!

PR #14506, opened 2019-07-30
chore(ui/alerts): change notificationRule to rule

121watts:
I couldn't type or auto-complete `notificationRule` one more time.  With our types and usage of `alerts/notifications/rules.ts` directory structure I don't think we need to prefix all mentions of `rule` with `notification`.


ui/src/alerting/components/AlertColumnHeader.tsx
```diff
@@ -0,0 +1,27 @@
...
+  justifyContent: 'space-between',
+  alignItems: 'center',
+}
+
+interface Props {
+  title: string
+  onCreate: () => void
+}
+
+const AlertsColumnHeader: FC<Props> = ({onCreate, title}) => {
```
ebb-tide: the file name and component name don't match (one's alerts.. the other is alert.. )and that can get confusing, could they match please? :)

ui/src/pageLayout/containers/Nav.tsx
```diff
@@ -10,7 +10,6 @@ import CloudNav from 'src/pageLayout/components/CloudNav'
 import AccountNavSubItem from 'src/pageLayout/components/AccountNavSubItem'
 import CloudExclude from 'src/shared/components/cloud/CloudExclude'
 import CloudOnly from 'src/shared/components/cloud/CloudOnly'
-import {FeatureFlag} from 'src/shared/utils/featureFlag'
```
ebb-tide: I think this is the line that is causing js tests to fail

ui/src/alerting/components/notifications/RulesColumn.tsx
```diff
@@ -0,0 +1,40 @@
...
+type Props = StateProps
+
+const NotificationRulesColumn: FunctionComponent<Props> = ({rules}) => {
+  const handleClick = () => {
+    console.log('creating rules: ')
+  }
+
+  return (
+    <>
+      <AlertsColumnHeader title="Notification Rules" onCreate={handleClick} />
```
ebb-tide: This is helpful :)

ui/src/alerting/components/AlertColumnHeader.tsx
121watts: yaaas!

121watts: good catch

PR #14507, opened 2019-07-30
fix(ui): prevent dashboard cell crash when editing cell

### LINKED ISSUE #14508: Editing a cell causes cell on dashboard to crash

chnn:
Previously, opening the VEO for a cell on a dashboard would cause the view for the cell on the dashboard to reload. In turn, this would cause the dashboard cell to crash.

This commit updates the logic for opening the VEO so that the view is only reloaded when necessary.

Closes #14508

### DIFF
--- a/ui/src/dashboards/components/EditVEO.tsx
+++ b/ui/src/dashboards/components/EditVEO.tsx
@@ -16,12 +16,11 @@ import {saveVEOView} from 'src/dashboards/actions'
 import {setView, getViewForTimeMachine} from 'src/dashboards/actions/views'
 
 // Utils
-import {getActiveTimeMachine} from 'src/timeMachine/selectors'
 import {TimeMachineEnum} from 'src/timeMachine/constants'
+import {getView} from 'src/dashboards/selectors'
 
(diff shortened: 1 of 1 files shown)


chnn:
> A test would be nice for this one 😸 . I could see this regress rather easily.

We were planning on it and thought Cypress would be the best way to cover it. But it's not reproducible without time series data in cypress 😢

121watts:
@chnn can't we just mock the time series?

chnn:
doh! I totally forgot we could do that.

I'll add a test, thanks.

chnn:
After some experimentation it doesn't look too feasible. What I tried:

1. Mocking the `/api/v2/query` endpoint. This endpoint is hit multiple times when we exercise the behavior that causes this crash. So the mock needs to return different responses depending on what each request is, or the order each request comes in. Both of these options seemed fragile, and likely to break for no good reason in the future.

2. Writing data as line protocol before the test. Since this data doesn't get reset between tests, I think this approach would leak state between each test run.

Going to punt for now.

PR #14510, opened 2019-07-30
test(tsm1): TagKeys/TagValues stops scanning when context is cancelled

jlapacik:
Closes #

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/tsdb/tsm1/engine_schema.go
+++ b/tsdb/tsm1/engine_schema.go
@@ -14,11 +14,20 @@ import (
 	"github.com/influxdata/influxql"
 )
 
+// cancelCheckInterval represents the period at which TagKeys and TagValues
+// will check for a canceled context. Specifically after every 64 series
+// scanned, the query context will be checked for cancellation, and if canceled,
+// the calls will immediately return.
+const cancelCheckInterval = 64
+
 // TagValues returns an iterator which enumerates the values for the specific
 // tagKey in the given bucket matching the predicate within the
 // time range (start, end].
 //
 // TagValues will always return a StringIterator if there is no error.
+//
+// If the context is canceled before TagValues has finished processing, a non-nil
+// error will be returned along with a partial result of the already scanned values.
 func (e *Engine) TagValues(ctx context.Context, orgID, bucketID influxdb.ID, tagKey string, start, end int64, predicate influxql.Expr) (cursors.StringIterator, error) {
 	encoded := tsdb.EncodeName(orgID, bucketID)
 
@@ -180,9 +189,8 @@ func (e *Engine) tagValuesPredicate(ctx context.Context, orgBucket, tagKeyBytes 
 	)
 
 	for i := range keys {
-		// to keep cache scans fast,
-		// check context every 64 iteratons.
-		if i%64 == 0 {
+		// to keep cache scans fast, check context every 'cancelCheckInterval' iteratons
+		if i%cancelCheckInterval == 0 {
 			select {
 			case <-ctx.Done():
 				stats = statsFromIters(stats, iters)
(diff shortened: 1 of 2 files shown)


tsdb/tsm1/engine_schema.go
```diff
@@ -141,17 +163,34 @@ func (e *Engine) tagValuesPredicate(ctx context.Context, orgBucket, tagKeyBytes
...
 		tags   models.Tags
 		keybuf []byte
 		sfkey  []byte
-		stats  cursors.CursorStats
 	)
 
 	for i := range keys {
+		// to keep cache scans fast,
+		// check context every 64 iteratons.
+		if i%64 == 0 {
```
stuartcarnie: Can we make this an unexported const and update the references

tsdb/tsm1/engine_schema.go
```diff
@@ -175,6 +214,14 @@ func (e *Engine) tagValuesPredicate(ctx context.Context, orgBucket, tagKeyBytes
 		}
 
 		for _, iter := range iters {
+			// Always check context before performing a tsm file scan
```
stuartcarnie: I'm don't think we need to check it for every iterator seek, as this operation performance a binary search, so it should execute in O(logN) time. I think every 64 iterations of the outer loop should suffice.

tsdb/tsm1/engine_schema.go
```diff
@@ -368,6 +461,14 @@ func (e *Engine) tagKeysPredicate(ctx context.Context, orgBucket []byte, start,
...
```
stuartcarnie: Same here, given seek is binary search, checking every 64 keys of the outer loop should suffice.

tsdb/tsm1/engine_schema.go
```diff
@@ -141,17 +165,33 @@ func (e *Engine) tagValuesPredicate(ctx context.Context, orgBucket, tagKeyBytes
 		return true
 	})
 
+	var stats cursors.CursorStats
+
+	if canceled {
+		stats = statsFromIters(stats, iters)
+		return cursors.NewStringSliceIteratorWithStats(nil, stats), ctx.Err()
```
e-dard: I suppose this is getting a little philosophical now, but is this an error or not? `tagValuesPredicate` didn't hit any particular error conditions, it was just told to finish up early (due to deadline or explicit cancellation). It's clear here that you want to inform the caller of the (partial) work that was done. 

To me, this feels a little bit like the API should not be returning an error, but inside of the result value (the iterator), we should actually have a `Err()` method, which callers should check. That would return an error if `tagValuesPredicate` was told to terminate early.

To me, if a method returns a result and an error, and the error is `non-nil`, then the result is always undefined or useless.

tsdb/tsm1/engine_schema.go
```diff
@@ -14,6 +14,8 @@ import (
 	"github.com/influxdata/influxql"
 )
 
+const contextPeriod = 64
```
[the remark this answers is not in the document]
jlapacik: Sure thing. Yes keys scans.

tsdb/tsm1/engine_schema.go
```diff
@@ -113,7 +129,7 @@ func (e *Engine) tagValuesPredicate(ctx context.Context, orgBucket, tagKeyBytes
 
 	keys, err := e.findCandidateKeys(ctx, orgBucket, predicate)
 	if err != nil {
-		return nil, err
+		return cursors.EmptyStringIterator, err
```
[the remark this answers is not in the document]
jlapacik: Initially I was not returning an error when the context was canceled. Instead I just returned the partial work, i.e the number of bytes/values scanned up until that point. However after looking at the godoc for `context.Done()` as well as other examples, it seemed an error should be returned for a canceled context.

I decided to split the difference, returning a non-nil error in the event the context is canceled while at the same time returning the partial work up until that point. Note there is precedent for such return semantics. As identified in a slack conversations, the stdlib's `io.Reader` is one such example.

Also I have verified that no callers of TagKeys/TagValues rely on a `nil` string iterator in the event of a non-nil error.

tsdb/tsm1/engine_schema.go
jlapacik: Yes I think canceling the context should result in an error. Were we to not treat it as an error, we would run into inconsistencies higher up the stack where a canceled context is treated as an error. Also as mentioned above (from a slack conversation) there is a precedent, set by the stdlib, for returning both a non-nil error as well as non-zero value result.

Having said that I will make it explicit in the godoc for TagKeys/TagValues that this is the case.

jlapacik:
alchemy please merge

PR #14520, opened 2019-07-31
feat(alerts): create notification rule

121watts:
### What has been done
- The UI Component only bits for  the creation of a `NotificationRule`

### What is yet to be done
- Make actual calls to the server to create these.  Im reluctant to add that flow as there is just so much changing right now
- Hook the created notification state up to our Redux store
- Client side validation

### Notes
Given the high velocity of this feature I'm sure the typings and structure of these rules will change.  I would like to merge this in under a feature flag and deal with those changes once they solidify.   This is a good iteration point.

# Behold!
Screen Shot 2019-08-06 at 2 49 31 PM


ui/src/alerting/components/notifications/NewRuleOverlay.reducer.ts
```diff
@@ -0,0 +1,88 @@
+// Libraries
+import {v4} from 'uuid'
+
+// Types
+import {NotificationRuleBox, StatusRuleItem, TagRuleItem} from 'src/types'
+
+export type RuleState = NotificationRuleBox
```
chnn: I'm confused about this name. I know I originally suggested `IDBox`, but now am thinking that doesn't make sense either. The conventional definition of a boxed type is a type that wraps a value with an object to force it to be allocated on the heap. This doesn't wrap an object and it is already allocated on the heap. The only other `Box` type in the UI right now is the `CancelBox`, which does wrap a value but not for the purposes of heap allocation. ¯\_(ツ)_/¯

I do think it would be really nice to have a convention for types that represent our working browser-side copy of some resource that exists on the server. I've seen us use the `Working` prefix for that before, e.g. `WorkingView`. I could also imagine us using `Draft` or `Client`, e.g. `DraftNotificationRule`, `DraftView`, ...

Probably doesn't matter much, but nice names are nice :)

ui/src/alerting/components/notifications/NewRuleOverlay.reducer.ts
```diff
@@ -0,0 +1,88 @@
...
+export type Actions =
+  | {type: 'UPDATE_RULE'; rule: NotificationRuleBox}
+  | {type: 'SET_ACTIVE_SCHEDULE'; schedule: 'cron' | 'every'}
+  | {type: 'UPDATE_STATUS_RULES'; statusRule: StatusRuleItem}
+  | {type: 'ADD_TAG_RULE'; tagRule: TagRuleItem}
+  | {type: 'DELETE_STATUS_RULE'; statusRuleID: string}
+  | {type: 'UPDATE_TAG_RULES'; tagRule: TagRuleItem}
+  | {type: 'DELETE_TAG_RULE'; tagRuleID: string}
+
+export const reducer = (state: RuleState, action: Actions) => {
```
chnn: I've been digging the component-local reducers so much.

ui/src/alerting/components/notifications/NewRuleOverlay.reducer.ts
```diff
@@ -0,0 +1,88 @@
...
+
+      const tagRules = state.tagRules.filter(tr => {
+        return tr.id !== tagRuleID
+      })
+
+      return {...state, tagRules}
+    }
+
+    default:
+      throw new Error('unhandled reducer action in <NewRuleOverlay/>')
```
chnn:
```suggestion
      const neverAction: never = action

      throw new Error(`unhandled action "${(neverAction as any).type}"`)
```

if you're interested in making this a compile time error

ui/src/alerting/components/notifications/NewRuleOverlay.scss
```diff
@@ -0,0 +1,56 @@
+.rule-eo {
+  width: calc(100vw - #{$ix-marg-e});
```
chnn: oh neat. I never knew how to use sass variables with `calc`

ui/src/alerting/components/notifications/NewRuleOverlay.reducer.ts
```diff
@@ -0,0 +1,88 @@
+// Libraries
+import {v4} from 'uuid'
+
+// Types
+import {NotificationRuleBox, StatusRuleItem, TagRuleItem} from 'src/types'
+
+export type RuleState = NotificationRuleBox
+export type Actions =
```
chnn:
```suggestion
export type Action =
```

ui/src/alerting/components/notifications/NewRuleOverlay.tsx
```diff
@@ -1,27 +1,95 @@
...
+import {NotificationRuleBox} from 'src/types'
 
 type Props = WithRouterProps
 
+export const newRuleState: RuleState = {
+  ...newRule,
+  schedule: 'every',
+}
+
+export const NewRuleDispatch = React.createContext<Dispatch<Actions>>(null)
```
chnn: I've been digging this pattern a lot. I was thinking we could codify it into a little utility. Something like:

```
const [MyStateProvider, useMyState, useMyDispatch] = createReducerContext(reducer, initialState)

const MyComponent = ({children}) => {
  return (
    <MyStateProvider>
      <SomeChild>
        <SomeNestedChild />
      </SomeChild>
    </MyStateProvider>
  )
}

const SomeNestedChild = () => {
  const state = useMyState() 
  const dispatch = useMyDispatch()

  // ...
}

```

ui/src/alerting/components/notifications/NewRuleOverlay.tsx
```diff
@@ -1,27 +1,95 @@
...
-    </Overlay>
+    <NewRuleDispatch.Provider value={dispatch}>
+      <Overlay visible={true}>
+        <Overlay.Container maxWidth={800}>
+          <Overlay.Header
+            title="Create a Notification Rule"
+            onDismiss={handleDismiss}
+          />
+          <Overlay.Body>
+            <Grid>
```
chnn: So much `Grid` ahhhh

ui/src/alerting/components/notifications/RulesColumn.tsx
```diff
@@ -35,4 +47,4 @@ const mstp = (state: AppState) => {
 export default connect<StateProps, {}, {}>(
   mstp,
   null
-)(NotificationRulesColumn)
+)(withRouter(NotificationRulesColumn))
```
chnn:
```suggestion
)(withRouter<StateProps>(NotificationRulesColumn))
```

ui/src/alerting/components/notifications/SMTPMessage.tsx
```diff
@@ -0,0 +1,43 @@
...
+      </Form.Element>
+      <Form.Element label="Subject">
+        <Input
+          value={subjectTemplate}
+          name="subjectTemplate"
+          onChange={onChange}
+        />
+      </Form.Element>
+      <Form.Element label="Body">
+        <TextArea
```
chnn: soooo many inputs. I don't even want to think about adding validation...

ui/src/alerting/components/notifications/StatusChangeDropdown.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+// Types
+import {StatusRuleItem} from 'src/types'
+
+// Components
+import {Dropdown} from '@influxdata/clockface'
+import {NewRuleDispatch} from 'src/alerting/components/notifications/NewRuleOverlay'
+
+// Utils
+import {changes, changeStatusRule, activeChange} from './statusChange'
```
chnn: Can you put this in `src/alerting/utils`? I know you're not a fan of utils folders, but I think adding yet another file structure convention that isn't followed consistently is worse.

ui/src/alerting/components/notifications/StatusLevels.tsx
```diff
@@ -0,0 +1,71 @@
...
+        ...status,
+        value,
+      },
+    })
+  }
+
+  return (
+    <ComponentSpacer direction={FlexDirection.Row} margin={ComponentSize.Small}>
+      <TextBlock text="When status" />
+      <ComponentSpacer.FlexChild grow={0} basis={140}>
```
chnn: I wish this `ComponentSpacer` stuff didn't exist. We've adopted all the verbosity of a CSS-in-JS solution with none of the power.

ui/src/alerting/components/notifications/StatusLevels.tsx
```diff
@@ -0,0 +1,71 @@
...
+interface Props {
+  status: StatusRuleItem
+}
+
+const StatusLevels: FC<Props> = ({status}) => {
+  const {currentLevel, previousLevel} = status.value
+  const dispatch = useContext(NewRuleDispatch)
+
+  const onClickLevel = (type, level) => {
+    const value = {
```
chnn: What do you think about moving this logic to the reducer with its own action type?

I think in general when following the action/reducer style, we are too reluctant to add actions for specific/one-off events that a user performs in the UI. The result is that we have very general `SET_FOO` actions that we dispatch in various configurations, and our reducers stay really small/dumb.  Then we don't get one of the big benefits of reducers: that they centralize state changes, so that we are coding an explicit rather than implicit state machine.

Not a big deal here, just something I wanna see us do differently in general.

Related thought-leadering: https://twitter.com/dan_abramov/status/800310164792414208

ui/src/alerting/components/notifications/TagRule.tsx
```diff
@@ -0,0 +1,107 @@
...
+      tagRule: {
+        ...tagRule,
+        value: newValue,
+      },
+    })
+  }
+
+  const onSelectOperator = (operator: Operator) => {
+    dispatch({
+      type: 'UPDATE_TAG_RULES',
```
chnn: (Same comment about dumber actions / smarter reducers.)

ui/src/alerting/components/notifications/endpointBase.ts
```diff
@@ -0,0 +1,23 @@
+import {SlackBase, SMTPBase, PagerDutyBase} from 'src/types'
```
chnn: (Same comment about moving this to a `utils` directory.)

ui/src/alerting/components/notifications/endpointBase.ts
```diff
@@ -0,0 +1,23 @@
...
+
+type EndpointBase = SlackBase | SMTPBase | PagerDutyBase
+
+export const getEndpointBase = (endpoints, id): EndpointBase => {
```
chnn: Add argument types?

ui/src/alerting/components/notifications/statusChange.ts
```diff
@@ -0,0 +1,36 @@
+import {omit} from 'lodash'
+import {StatusRuleItem, LevelRule} from 'src/types'
+
+type Changes = 'changes from' | 'equal'
+export const changes: Changes[] = ['changes from', 'equal']
```
chnn:
```suggestion
export const CHANGES: Change[] = ['changes from', 'equal']
```

ui/src/alerting/components/notifications/statusChange.ts
```diff
@@ -0,0 +1,36 @@
+import {omit} from 'lodash'
+import {StatusRuleItem, LevelRule} from 'src/types'
+
+type Changes = 'changes from' | 'equal'
```
chnn: This is the hill I die on.

```suggestion
type Change = 'changes from' | 'equal'
```

ui/src/alerting/components/notifications/statusChange.ts
```diff
@@ -0,0 +1,36 @@
+import {omit} from 'lodash'
```
chnn: (Same comment about moving this to a utils directory.)

ui/src/pageLayout/containers/Nav.tsx
```diff
@@ -132,31 +131,29 @@ class SideNav extends PureComponent<Props, State> {
           )}
           active={getNavItemActivation(['tasks'], location.pathname)}
         />
-        <FeatureFlag name="alerting">
-          <NavMenu.Item
+        <NavMenu.Item
```
chnn: Wait don't we still need the `<FeatureFlag>`?

ui/src/types/alerting.ts
```diff
@@ -1,9 +1,58 @@
+import {StatusRule, NotificationRuleBase, TagRule} from 'src/client'
+
+export interface AddID<T> {
+  id: string
+  value: T
+}
+
+export type StatusRuleItem = AddID<StatusRule>
```
chnn: Continuing the earlier train of thought, maybe these could be `DraftStatusRule` and `DraftTagRule`, which would both be part of a `DraftNotificationRule`.

ui/src/types/alerting.ts
```diff
@@ -1,9 +1,58 @@
...
+export interface NotificationRuleBaseBox
+  extends ExcludeKeys<NotificationRuleBase> {
+  schedule: 'cron' | 'every'
+  statusRules: StatusRuleItem[]
+  tagRules: TagRuleItem[]
+}
+
+export type NotificationRuleBox = SlackRule | SMTPRule | PagerDutyRule
+
+export type SlackBase = {
```
chnn: Should these be part of the swagger spec instead?

ui/src/alerting/components/notifications/NewRuleOverlay.reducer.ts
121watts: I agree.  This was my least favorite part of this PR.  I like `Draft` as a namespace.

ui/src/alerting/components/notifications/statusChange.ts
121watts: hahahah

ui/src/pageLayout/containers/Nav.tsx
121watts: We do :)

ui/src/types/alerting.ts
121watts: Yup yup

ui/src/types/alerting.ts
121watts: Yuppers!

ui/src/alerting/components/notifications/LevelsDropdown.tsx
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  const button = (active, onClick) => (
+    <Dropdown.Button
+      color={ComponentColor.Default}
+      active={active}
+      onClick={onClick}
+      testID={`levels--dropdown--button ${type}`}
+    >
+      <div className="color-dropdown--item">
```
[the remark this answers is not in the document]
121watts: will do.

ui/src/alerting/components/notifications/NewRuleOverlay.tsx
121watts: Oh, I kinda like that!

ui/src/alerting/components/notifications/SMTPMessage.tsx
121watts: that will be an entirely different story

ui/src/alerting/components/notifications/StatusChangeDropdown.tsx
121watts: what about StatusChange.utils.ts 😄

ui/src/alerting/components/notifications/StatusLevels.tsx
121watts: What if it was just renamed to <FlexBox />?

ui/src/alerting/components/notifications/StatusLevels.tsx
121watts: That's what the ui used to do.  I don't now how we strayed.  But, I really like it.  Makes it super easy to test more complex state changes.

ui/src/alerting/components/notifications/StatusLevels.tsx
chnn: The thing that gets me about it is that it causes us to add a ton more markup / divs everywhere. This looks really noisy when viewing the JSX or the DOM via devtools. Like in the hypothetical scenario:

```tsx
<Nav className="nav">
  <ComponentSpacer direction={FlexDirection.Column}>
    <ComponentSpacer.FlexChild grow={0} shrink={0} basis={50}>
      <NavItem className="nav-item" title="foo" />
    </ComponentSpacer.FlexChild>
  </ComponentSpacer>
</Nav>
```

Could just be

```tsx
<Nav className="nav">
  <NavItem className="nav-item" title="foo" />
</Nav>
```

with some corresponding CSS on `.nav` and `.nav-item`. You still have to define the the styling, but it's separate and once it's done you don't have to think about it or parse through all the `ComponentSpacer` noise to work with the component. 

This is the same thing that bugs me about CSS-in-JS libs like emotion or styled-components. Without a lot of care and strong conventions, you end up adding a lot more noise to your components. It also feels like we are defining a duplicate but less powerful API for something that is already so simple to achieve (putting flexbox styles on an element).

ofc this is all stuff I think should be hashed out systematically via clockface, not in this PR. Just commenting to share what I've been thinking about our current approach

ui/src/types/alerting.ts
121watts: Going to do this in another PR as it looks like swagger was recently updated and there are about a bagillion errors now.

PR #14521, opened 2019-07-31
feat: Add check API

kelwang:
Closes https://github.com/influxdata/influxdb/issues/14325

add check struct, kv, http, authorizer, influxd launcher and testings. based on the swagger file design.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/check.go
+++ b/check.go
@@ -1,0 +1,127 @@
+package influxdb
+
+import (
+	"context"
+	"encoding/json"
+)
+
+// consts for checks config.
+const (
+	CheckDefaultPageSize = 100
+	CheckMaxPageSize     = 500
+)
+
+// Check represents the information required to generate a periodic check task.
+type Check interface {
+	Valid() error
+	Type() string
+	json.Marshaler
+	Updator
+	Getter
+}
+
+// ops for checks error
+var (
+	OpFindCheckByID = "FindCheckByID"
+	OpFindCheck     = "FindCheck"
+	OpFindChecks    = "FindChecks"
+	OpCreateCheck   = "CreateCheck"
+	OpUpdateCheck   = "UpdateCheck"
+	OpDeleteCheck   = "DeleteCheck"
+)
+
+// CheckService represents a service for managing checks.
+type CheckService interface {
+	// UserResourceMappingService must be part of all CheckService service,
+	// for create, delete.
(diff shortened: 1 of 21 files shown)


notification/check/check.go
```diff
@@ -0,0 +1,160 @@
...
+func (b *Base) SetDescription(description string) {
+	b.Description = description
+}
+
+// SetStatus implements influxdb.Updator interface.
+func (b *Base) SetStatus(status influxdb.Status) {
+	b.Status = status
+}
+
+var typToCheck = map[string](func() influxdb.Check){
```
ebb-tide: typToCheck looks like a typo?

notification/check/check.go
```diff
@@ -0,0 +1,160 @@
...
+			Msg:  "Check OrgID is invalid",
+		}
+	}
+	if b.Status != influxdb.Active && b.Status != influxdb.Inactive {
+		return &influxdb.Error{
+			Code: influxdb.EInvalid,
+			Msg:  "invalid status",
+		}
+	}
+	if b.StatusMessageTemplate == "" {
```
ebb-tide: Status message template can be empty.

check.go
```diff
@@ -0,0 +1,127 @@
...
+
+// CheckFilter represents a set of filters that restrict the returned results.
+type CheckFilter struct {
+	ID    *ID
+	Name  *string
+	OrgID *ID
+	Org   *string
+}
+
+// QueryParams Converts NotificationRuleFilter fields to url query params.
```
ebb-tide: Check Filter instead of NotificationRuleFilter here.

kv/check.go
```diff
@@ -0,0 +1,704 @@
...
+			return nil, &influxdb.Error{
+				Err: err,
+			}
+		}
+		return c, nil
+	}
+
+	err = s.kv.View(ctx, func(tx Tx) error {
+		if filter.Org != nil {
+			o, err := s.findOrganizationByName(ctx, tx, *filter.Org)
```
ebb-tide: line 131 suggests that orgID lookup will be efficient but it looks like this function searches for org Name only?

kelwang: Good point, I think it's related the org and orgID filter issue, we will make update for all them in one pr.

Issue #14511, opened 2019-07-25
Issuing generic flux query for too much data causes error


desa:
We're currently hitting https://github.com/influxdata/influxdb-client-js/blob/845de2538ce0cb8dce8fe8b0fa6bf22206badcee/src/wrappers/queries.ts#L83-L85
whenever users issue a query for a very large amount of data. Ideally we should be able to handle this situation gracefully.
![russ error](https://user-images.githubusercontent.com/1477539/61905181-7636c700-aef6-11e9-9271-2bfc9657361c.png)

chnn:
After talking with @russorat, here's the plan:

- In the cloud environment, set the limit for data received to 10MB (after decompression)
- In OSS, set the limit to be much higher (say, 150MB)
- When the limit is hit, display the partial response and a help text letting the user know they can avoid the limit by aggregating their data or selecting a smaller time window

chnn:
Closed by #14529—let's follow up any UX polish we want in another issue.

PR #14529, opened 2019-07-31
feat(ui): display partially received Flux responses

chnn:
A basic fix for #14511. If a Flux response is truncated because it is too large, the partial result will still display as well as a notification.

Screen Shot 2019-07-31 at 4 06 39 PM

The new limits for Flux responses are:

- OSS: 100 MiB
- Cloud 2: 10 MiB

Next we should improve the UX for this event (add suggestions to reduce response size, add button to run query anyway).

I refactored some code along the way (part of #14482).

I also tweaked the first load experience for the dashboard. Before:

before

After:

after


chnn:
> I'm not familiar with TypeScript, so I don't know that I have much of value to say for this PR. Is there a particular part of this that you want me to look at?

I can't speak for kelvin (who requested the view), but I appreciate you taking a look.

> I see there are AST structures in there, are they actually being used?

Yeah, they were in use before. This PR is just shifting around where the type definitions for the various AST nodes are stored. We use them to do things like analyze the duration of a query, or pick out what buckets are read from or written to in a query.

ui/src/shared/apis/query.ts
```diff
@@ -1,19 +1,177 @@
...
+  }
+
+  const body: Query = {
+    query,
     extern,
-    limitChars: MAX_RESPONSE_CHARS,
+    dialect: {annotations: ['group', 'datatype', 'default']},
+  }
+
+  const controller = new AbortController()
```
121watts: Might want a polyfill for this.  Especially for our Safari brethren.

ui/src/shared/utils/insertPreambleInScript.ts
```diff
@@ -15,10 +15,16 @@ export const insertPreambleInScript = async (
     return `${preamble}\n\n${script}`
   }
 
-  // TODO: replace this with `import {parse} from '@influxdata/flux-parser'`
-  // when the flux team adds location.source data to the rust implementation
+  // TODO: Replace this with `import {parse} from '@influxdata/flux-parser'`
+  // when the Flux team adds location.source data to the rust implementation
```
121watts: How dare you.

ui/src/types/queries.ts
```diff
@@ -1,86 +1,4 @@
-import {Source} from 'src/types'
```
121watts: Heck yea!

PR #14531, opened 2019-08-01
refactor(ui): use locally generated client for buckets and orgs

desa:
Link #14482 

This PR uses the locally generated client instead of `@influxdata/influx` for API calls. I figured it would be best to open a bunch of smaller PRs rather than one massive PR for all of the types.

### DIFF
--- a/ui/src/organizations/actions/orgs.ts
+++ b/ui/src/organizations/actions/orgs.ts
@@ -186,9 +186,9 @@ export const createOrgWithBucket = (
     dispatch(addOrg(createdOrg))
     dispatch(push(`/orgs/${createdOrg.id}`))
 
-    bucket.orgID = createdOrg.id
-
-    const bucketResp = await api.postBucket({data: bucket})
+    const bucketResp = await api.postBucket({
+      data: {...bucket, orgID: createdOrg.id},
+    })
 
     if (bucketResp.status !== 201) {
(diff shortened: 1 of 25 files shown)


ui/src/organizations/actions/orgs.ts
```diff
@@ -168,10 +186,13 @@ export const createOrgWithBucket = (
     dispatch(addOrg(createdOrg))
     dispatch(push(`/orgs/${createdOrg.id}`))
 
-    await client.buckets.create({
-      ...bucket,
-      orgID: createdOrg.id,
-    })
+    bucket.orgID = createdOrg.id
```
chnn: would it be possible to remove this line and instead do something like

```
const bucketResp = await api.postBucket({data: {...bucket, orgID: createdOrg.id}})
``` 

right below? As a convention we pretty much never mutate POJOs after they are created, since it tends to cause some subtle bugs.

The type checker might complain, in which case I think a type assertion would probably work:

```
const data = {...bucket, orgID: createdOrg.id} as Bucket

const bucketResp = await api.postBucket({data})
```

desa: Awesome. Thats definitely good feedback. In retrospect yeah, I totally shouldn't have mutated bucket. I've got to get back into the FP style of always just make a new thing.

PR #14533, opened 2019-08-01
fix(ui): improve relevant token detection when saving a query as a task

### LINKED ISSUE #14226: Evaluate DE query using AST
When getting a read bucket name for displaying tokens, instead of string manipulation try to use AST to evaluate the query.

chnn:
Closes #14226

See commits for details

### DIFF
--- a/ui/src/authorizations/utils/permissions.ts
+++ b/ui/src/authorizations/utils/permissions.ts
@@ -160,18 +160,21 @@ export enum BucketTab {
 
 /*
   Given a list of authorizations, return only those that allow performing the
-  supplied `action` to the supplied `bucketName`.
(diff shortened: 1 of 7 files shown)


ui/src/authorizations/utils/permissions.ts
```diff
@@ -171,3 +157,24 @@ export enum BucketTab {
   AllBuckets = 'All Buckets',
   Scoped = 'Scoped',
 }
+
+/*
+  Given a list of authorizations, return only those that allow performing the
+  supplied `action` to the supplied `bucketName`.
```
chnn: This should read as "... to all of the supplied `bucketNames`".

PR #14534, opened 2019-08-01
feat(query): add storage request duration metric

wolffcm:
This adds a prometheus histogram metric to query execution that records the duration of read requests to storage.

influxdata/idpe#4126

### DIFF
--- a/query/stdlib/influxdata/influxdb/metrics.go
+++ b/query/stdlib/influxdata/influxdb/metrics.go
@@ -1,0 +1,88 @@
+package influxdb
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/influxdata/flux/execute"
+	platform "github.com/influxdata/influxdb"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	orgLabel = "org"
+	opLabel  = "op"
+)
+
+type metrics struct {
+	ctxLabelKeys []string
+	requestDur   *prometheus.HistogramVec
+}
+
+// NewMetrics produces a new metrics objects for an influxdb source.
+// Currently it just collects the duration of read requests into a histogram.
+// ctxLabelKeys is a list of labels to add to the produced metrics.
+// The value for a given key will be read off the context.
+// The context value must be a string or an implementation of the Stringer interface.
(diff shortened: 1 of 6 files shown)


query/stdlib/influxdata/influxdb/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+	}, labelKeys)
+	m.ctxLabelKeys = ctxLabelKeys
+	return m
+}
+
+// PrometheusCollectors satisfies the PrometheusCollector interface.
+func (m *metrics) PrometheusCollectors() []prometheus.Collector {
+	if m == nil {
+		return nil
+	}
```
affo: Why is this check for `nil` necessary 🤔 ?

query/stdlib/influxdata/influxdb/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (m *metrics) getLabelValues(ctx context.Context, orgID platform.ID, op string) []string {
+	if m == nil {
+		return nil
+	}
+	labelValues := make([]string, len(m.ctxLabelKeys)+2)
+	for i, k := range m.ctxLabelKeys {
+		value := ctx.Value(k)
```
affo: is it the expected behavior to panic if the context does not contain that key?

affo: Never mind. if value is nil `value.(type)` won't panic.

query/stdlib/influxdata/influxdb/metrics.go
wolffcm: Well... without this check if I just point the current idpe master at this, it will crash in the pipeline tests because `m` is `nil`.  I wanted this set of changes to be okay without any changes to idpe.  Do you think it's okay?  In any case I'll add a comment.

wolffcm:
> One thing I don't know about. Are those implementations of PrometheusCollectors() in the right place to allow auto registration?

Do you mean auto-registration in queryd?  Sadly the answer to that is no.  I thought that maybe queryd would just call `AddControllerConfigDependencies` the same as OSS, but it does not.  It uses a more complicated method instead, which I'll need to update as well:
https://github.com/influxdata/idpe/blob/master/internal/cmd/queryd/queryd.go#L497

Issue #14092, opened 2019-06-10
Translate exists operator for storage pushdowns


jlapacik:
Flux uses (or will use) the `exists` operator for determining if a column value exists on a table. Storage however uses comparisons against the empty string to achieve the same result (#13880).

In order to push existence checks down to storage, `PushDownFilterRule` must perform the following translations:

* `exists v` to `v != ""`
* `not exists v` to `v == ""`

PR #14538, opened 2019-08-01
feat(query/stdlib): push down exists operator to storage

jsternberg:
The `exists` operator now gets pushed down to storage correctly. If
`exists` is used on a tag, then it will be rewritten to `tag != ""`
which is how storage defines if a tag exists. If `not exists` is used,
then it will use `tag == ""` which is how you would query storage for
only if a tag doesn't exist.

The `tag == ""` and `tag != ""` are different. For `tag == ""`, the
predicate is impossible for the storage layer to return true with.
Ideally, we would just rewrite this to return nothing and we wouldn't
bother with even querying storage. Instead, we just do not rewrite this
predicate because it cannot be rewritten to make sense with storage. If
we see `tag != ""`, it is the only one that can be passed through as-is
because `tag != ""` returns the same values as `exists tag`. It will
return true for every non-null value.

#14092

### DIFF
--- a/query/stdlib/influxdata/influxdb/rules.go
+++ b/query/stdlib/influxdata/influxdb/rules.go
@@ -123,6 +123,7 @@ func (PushDownFilterRule) Rewrite(pn plan.Node) (plan.Node, bool, error) {
 		// Nothing could be pushed down, no rewrite can happen
 		return pn, false, nil
 	}
+	pushable, _ = rewritePushableExpr(pushable)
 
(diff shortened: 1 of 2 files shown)


jsternberg:
I've been under the impression that rewrites should always return the same value. I don't really see a way to do things otherwise without making things needlessly complex.

I wonder if there's an alternate way in the UI we could highlight the potential error? Like a linter? But I think we enter dangerous territory if we change the results.

jlapacik:
I agree. Rewrites must preserve semantic equivalence.

PR #14540, opened 2019-08-02
fix(http): The org parameter takes either the ID or Name interchangeably

### LINKED ISSUE #13906: InfluxDB 2.0: The "org" query param should be same for /query and /write  [area/api]

bednar:
Closes #13906

The org parameter could be `Organization.name` or `Organization.id` for '/write' and '/query'.

### DIFF
--- a/http/requests.go
+++ b/http/requests.go
@@ -12,10 +12,31 @@ const (
 	OrgName = "org"
 	// OrgID is the http query parameter to specify an organization by ID.
 	OrgID = "orgID"
+	// Org is the http query parameter that take either the ID or Name interchangeably
+	Org = "org"
 )
 
 // queryOrganization returns the organization for any http request.
 func queryOrganization(ctx context.Context, r *http.Request, svc platform.OrganizationService) (o *platform.Organization, err error) {
+
+	if organization := r.URL.Query().Get(Org); organization != "" {
+		if id, err := platform.IDFromString(organization); err == nil {
+			// Decoded ID successfully. Make sure it's a real org.
+			o, err := svc.FindOrganization(ctx, platform.OrganizationFilter{ID: id})
+			if err == nil {
+				return o, err
+			} else if platform.ErrorCode(err) != platform.ENotFound {
+				return nil, err
+			}
+		}
+
+		o, err := svc.FindOrganization(ctx, platform.OrganizationFilter{Name: &organization})
+		if err != nil {
+			return nil, err
+		}
+		return o, nil
+	}
+
 	filter := platform.OrganizationFilter{}
 	if reqID := r.URL.Query().Get(OrgID); reqID != "" {
 		filter.ID, err = platform.IDFromString(reqID)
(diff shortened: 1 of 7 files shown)


http/requests.go
```diff
@@ -12,21 +12,37 @@ const (
...
+			o, err := svc.FindOrganization(ctx, platform.OrganizationFilter{ID: id})
+			if err == nil {
+				return o, err
+			} else if platform.ErrorCode(err) != platform.ENotFound {
+				return nil, err
+			}
+		}
+
+		o, err := svc.FindOrganization(ctx, platform.OrganizationFilter{Name: &organization})
+		if err != nil {
```
kelwang: I think as long as the IDFromString returns no error, we can assure it is a qualified id, and user want to use orgID. So  we should return "not found error" immediately, without keep doing another search.

It's simpler just doing this
```go
if organization := r.URL.Query().Get(Org); organization != "" {
		if id, err := platform.IDFromString(organization); err == nil {
                         filter.ID = id
                }else{
                         filter.Name = &organization
                }
}
if reqID := r.URL.Query().Get(OrgID); reqID != "" {
		filter.ID, err = platform.IDFromString(reqID)
		if err != nil {
			return nil, err
		}
	}
```

http/requests.go
```diff
@@ -12,21 +12,37 @@ const (
 	OrgName = "org"
```
kelwang: remove this line, if it's not being used everywhere

http/requests.go
bednar: Fixed

http/requests.go
bednar: Fixed

PR #14547, opened 2019-08-02
feat(query): Buffered writer for to function

docmerlin:
This adds a buffered synchronous PointsWriter.  It wraps a regular PointsWriter and allows for fast buffered writing of points.  It is not thread-safe.

### DIFF
--- a/storage/points_writer.go
+++ b/storage/points_writer.go
@@ -2,6 +2,7 @@ package storage
 
 import (
 	"context"
+	"sync"
 
 	"github.com/influxdata/influxdb/models"
 )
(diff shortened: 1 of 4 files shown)


storage/points_writer.go
```diff
@@ -10,3 +11,62 @@ import (
...
+		var n int
+		if b.Buffered() == 0 {
+			// Large write, empty buffer.
+			// Write directly from p to avoid copy.
+			b.err = b.wr.WritePoints(ctx, p)
+		} else {
+			b.n += copy(b.buf[b.n:], p)
+			b.err = b.Flush(ctx)
+		}
+		p = p[n:]
```
stuartcarnie: I don't see `n` being updated anywhere, therefore it will always be 0

storage/points_writer.go
```diff
@@ -10,3 +11,62 @@ import (
...
+}
+
+// WritePoints writes the points to the underlying PointsWriter.
+func (b *BufferedPointsWriter) WritePoints(ctx context.Context, p []models.Point) error {
+	for len(p) > b.Available() && b.err == nil {
+		var n int
+		if b.Buffered() == 0 {
+			// Large write, empty buffer.
+			// Write directly from p to avoid copy.
+			b.err = b.wr.WritePoints(ctx, p)
```
stuartcarnie: I expect you want to add a `break` here or set `n` to `len(p)` so that this does not loop indefinitely or until WritePoints returns an error.

storage/points_writer.go
```diff
@@ -10,3 +11,62 @@ import (
 type PointsWriter interface {
 	WritePoints(context.Context, []models.Point) error
 }
+
+type BufferedPointsWriter struct {
+	sync.Mutex
```
stuartcarnie: This is unused

stuartcarnie:
You'll have to update your PR title to satisfy the *Semantic Pull Request*

Issue #14391, opened 2019-07-19
Convert view type from any to check view when Alerting is selected from a cell 


ebb-tide:
When user adds alerting to a cell, 
Convert view type from the existing view to a check view. 
For this first pass, we will convert view to XY view
This story may require the check view type to contain XY view properties. 
When user removes alerting from cell, convert from check view to XY view. 

![Screen Shot 2019-07-19 at 1.56.28 PM.png](https://images.zenhubusercontent.com/5cfab3d5b261421fbae0e39b/660a5a71-fec1-4e57-86a8-24aef8c89dd2)

ebb-tide:
This issue might require adding view properties to the check view.

Issue #14291, opened 2019-07-09
Add Create Check Page


ebb-tide:
add routes for create Alerts Page. 
Create Alerts Page is accessed from the button on top of the Alert Checks column on the Alerts Page. 
The Create Alerts Page renders TimeMachine with Alert Check builder enabled in the back ground.

![Screen Shot 2019-07-09 at 12.17.23 PM.png](https://images.zenhubusercontent.com/5cfab3d5b261421fbae0e39b/657e8281-c021-4db9-a972-f09234f094b7)

Issue #14537, opened 2019-08-01
Visualize check thresholds with giraffe threshold viewer


ebb-tide:
Checks should be visualized with CheckPlot.

ebb-tide:
I imagine we are going to perform this logic in a bunch of places. What do you think about extracting out a helper like

const getThresholdType = (threshold: Threshold): 'range' | 'greater' | 'lesser' => {
  // ...

ebb-tide:
I think removing this will mean that when a user drags the max below the min, then the threshold type will suddenly change from within: true to within: false, which sounds really jarring.

I imagine a way we can handle this is by determining the within value of a threshold when it is first instantiated, storing that in state, then treating the threshold accordingly. But that sounds complicated. I would really love to get those old Threshold types back.

PR #14550, opened 2019-08-02
feat: Alerting - Add alert builder

### LINKED ISSUE #14292: Add Edit Check Page
Add `checks/{checkID}/edit`route
Clicking on an existing Alert Check from the Alerts Page brings up TimeMachine in an overlay with the Alert check editing panel visible. 
Title should read Edit Alert Check

ebb-tide:
Closes #14292 #14291 #14391

Opens #14554


ui/src/alerting/components/builder/RemoveButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+type Props = DispatchProps
+
+const RemoveButton: FunctionComponent<Props> = ({
+  setActiveTab,
+  setViewType,
+  setCurrentCheck,
+}) => {
+  const handleClick = () => {
+    setCurrentCheck(RemoteDataState.NotStarted, null)
```
chnn: I think one of the main benefits of redux is that it decouples emitting actions from the state transformations. But when we write actions, we often treat them as simple setter functions for some single field in a redux slice. By doing so, I think we are getting all of the indirection of the event emitter abstraction with none of the benefits. 

In this case, a single action was taken in the UI (the alert builder remove button was clicked), but we have to emit 3 different actions. Concretely, this causes an issue where every Redux connected component will render 3 times instead of one, and the intermediate renders will see partially set state. It also means that the state transitions in our UI are spread all over the codebase, rather than being centralized in a reducer (so that we are coding an implicit state machine, rather than an explicit one).

This feels like a really widespread issue in our codebase... but I'd rather not continue it if possible. Could we instead make a single action (`ALERT_BUILDER_REMOVE_BUTTON_CLICKED` or whatever), and put this logic in a reducer?

ui/src/alerting/constants/index.ts
```diff
@@ -60,10 +78,10 @@ export const check2: Check = {
   statusMessageTemplate: 'this is a great message template',
   thresholds: [
     {
-      level: 'WARN',
+      level: 'UNKNOWN',
+      lowerBound: 20,
       allValues: false,
-      type: 'greater',
-    } as GreaterThreshold,
+    } as Threshold,
```
chnn: I don't think this needs an assertion.

ui/src/alerting/reducers/checks.ts
```diff
@@ -49,16 +51,34 @@ export default (
 
       case 'SET_CURRENT_CHECK':
         draftState.current.status = action.payload.status
-        if (action.payload.check) {
-          draftState.current.check = action.payload.check
-        }
+        draftState.current.check = action.payload.check
```
chnn: How useful is immer in this reducer? It seems like immer really shines for deeply nested state objects. But it also risks introducing hard to understand bugs due to it's [limitations](https://github.com/immerjs/immer#pitfalls). I wonder if we should only be using it when its benefits outway its costs.

ui/src/dashboards/components/NewVEO.tsx
```diff
@@ -92,7 +97,7 @@ const mstp = (state: AppState): StateProps => {
 
   let loadingState = RemoteDataState.Loading
 
-  if (activeTimeMachineID === TimeMachineEnum.VEO && viewIsNew) {
+  if (activeTimeMachineID === 'veo' && viewIsNew) {
```
chnn: Loving this change!

ui/src/shared/components/ThresholdMarkers.tsx
```diff
@@ -62,40 +50,41 @@ const ThresholdMarkers: FunctionComponent<Props> = ({
...
-                yScale={yScale}
-                yDomain={yDomain}
-                threshold={threshold}
-                onChangeMinPos={onChangeMinPos}
-                onChangeMaxPos={onChangeMaxPos}
-              />
-            )
-          default:
-            return null
+        if (threshold.lowerBound && threshold.upperBound) {
```
chnn: I imagine we are going to perform this logic in a bunch of places. What do you think about extracting out a helper like

```
const getThresholdType = (threshold: Threshold): 'range' | 'greater' | 'lesser' => {
  // ...
```

ui/src/shared/components/ThresholdMarkers.tsx
```diff
@@ -36,18 +36,6 @@ const ThresholdMarkers: FunctionComponent<Props> = ({
       [field]: yValue,
     }
 
-    if (
```
chnn: I think removing this will mean that when a user drags the max below the min, then the threshold type will suddenly change from `within: true` to `within: false`, which sounds really jarring.

I imagine a way we can handle this is by determining the `within` value of a threshold when it is first instantiated, storing that in state, then treating the threshold accordingly. But that sounds complicated. I would really love to get those old `Threshold` types back.

ui/src/shared/components/ViewSwitcher.tsx
```diff
@@ -69,7 +69,7 @@ const ViewSwitcher: FunctionComponent<Props> = ({
           )}
         </LatestValueTransform>
       )
-
+    case 'check':
```
chnn: There exists a `CheckPlot` component that I think should be used here instead.

ui/src/shared/components/XYPlot.tsx
```diff
@@ -23,14 +23,19 @@ import {DEFAULT_LINE_COLORS} from 'src/shared/constants/graphColorPalettes'
...
+  CheckViewProperties,
+} from 'src/types'
 
 interface Props {
   table: Table
   fluxGroupKeyUnion: string[]
   loading: RemoteDataState
   timeZone: TimeZone
-  viewProperties: XYViewProperties
+  viewProperties: XYViewProperties | CheckViewProperties
```
chnn: (Same as above... it would be great to separate this out to a different component.)

ui/src/timeMachine/components/AlertingButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+const AlertingButton: FunctionComponent<Props> = ({
+  setActiveTab,
+  status,
+  setCurrentCheck,
+  viewType,
+  setViewType,
+}) => {
+  const handleClick = () => {
+    if (viewType !== 'check') {
```
chnn: (Same comment about batching `SET_...` redux action types into a single action.)

ui/src/timeMachine/components/AlertingButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+}
+
+const mstp = (state: AppState): StateProps => {
+  const {
+    activeTab,
+    view: {
+      properties: {type: viewType},
+    },
+  } = getActiveTimeMachine(state)
+  let status = ComponentStatus.Default
```
chnn: (Same comment about moving derivation of data to within the component body.)

ui/src/timeMachine/components/view_options/OptionsSwitcher.tsx
```diff
@@ -29,6 +29,7 @@ class OptionsSwitcher extends PureComponent<Props> {
             <SingleStatOptions />
           </>
         )
+      case 'check':
```
chnn: Similar to the comment in the view switcher. I want to make sure we separate these concepts early so that the `XYPlot` and `LineOptions` don't accumulate a bunch of branching logic later on to deal with checks.

ui/src/timeMachine/components/view_options/ViewTypeDropdown.tsx
```diff
@@ -68,10 +67,14 @@ class ViewTypeDropdown extends PureComponent<Props> {
     ))
   }
 
-  private get selectedView(): string {
+  private get selectedView(): typeof view.properties.type {
```
chnn: There's a `ViewType` type for this.

ui/src/timeMachine/reducers/index.test.ts
```diff
@@ -25,40 +25,33 @@ import {
...
 
-// Constants
-import {TimeMachineEnum} from 'src/timeMachine/constants'
-
 // Types
-import {TimeMachineTab} from 'src/types/timeMachine'
 import {DashboardDraftQuery, QueryViewProperties} from 'src/types/dashboards'
 import {selectAggregateWindow} from '../actions/queryBuilder'
 
 describe('timeMachinesReducer', () => {
```
chnn: We should just delete this test... I've never seen it catch regressions and have had to change it so many times while refactoring.

ui/src/timeMachine/reducers/index.ts
```diff
@@ -67,11 +67,11 @@ export interface TimeMachineState {
 }
 
 export interface TimeMachinesState {
-  activeTimeMachineID: TimeMachineEnum
+  activeTimeMachineID: TimeMachineIDs
   timeMachines: {
-    [TimeMachineEnum.DE]: TimeMachineState
-    [TimeMachineEnum.VEO]: TimeMachineState
-    [TimeMachineEnum.Alerting]: TimeMachineState
+    ['de']: TimeMachineState
```
chnn: This can just be `de: TimeMachineState`

ui/src/timeMachine/reducers/index.ts
```diff
@@ -103,11 +103,11 @@ export const initialStateHelper = (): TimeMachineState => ({
 })
 
 export const initialState = (): TimeMachinesState => ({
-  activeTimeMachineID: TimeMachineEnum.DE,
+  activeTimeMachineID: 'de',
   timeMachines: {
-    [TimeMachineEnum.VEO]: initialStateHelper(),
-    [TimeMachineEnum.DE]: initialStateHelper(),
-    [TimeMachineEnum.Alerting]: initialStateHelper(),
+    ['veo']: initialStateHelper(),
```
chnn: (Same as above: this can just be `veo: initialStateHelper()`)

ui/src/timeMachine/constants/index.ts
```diff
@@ -1,5 +1 @@
-export enum TimeMachineEnum {
-  DE = 'de',
-  VEO = 'veo',
-  Alerting = 'alerting',
-}
+export type TimeMachineIDs = 'de' | 'veo' | 'alerting'
```
chnn:
```suggestion
export type TimeMachineID = 'de' | 'veo' | 'alerting'
```

ui/src/alerting/components/CheckEOHeader.tsx
```diff
@@ -0,0 +1,75 @@
...
+import {DEFAULT_CHECK_NAME, CHECK_NAME_MAX_LENGTH} from 'src/alerting/constants'
+
+interface Props {
+  name: string
+  onSetName: (name: string) => void
+  onCancel: () => void
+  onSave: () => void
+}
+
+const saveButtonClass = 'veo-header--save-cell-button'
```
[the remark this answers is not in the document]
ebb-tide: we will be going through all of these classes on Monday

http/swagger.yml
```diff
@@ -7510,6 +7517,32 @@ components:
...
+          items:
+            $ref: "#/components/schemas/DashboardColor"
+        note:
+          type: string
+        showNoteWhenEmpty:
+          description: if true, will display note when empty
+          type: boolean
+        axes:
+          $ref: '#/components/schemas/Axes'
+        legend:
```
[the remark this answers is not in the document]
ebb-tide: I will address this in: https://github.com/influxdata/influxdb/issues/14537

http/swagger.yml
```diff
@@ -7510,6 +7517,32 @@ components:
...
+          description: Colors define color encoding of data into a visualization
+          type: array
+          items:
+            $ref: "#/components/schemas/DashboardColor"
+        note:
+          type: string
+        showNoteWhenEmpty:
+          description: if true, will display note when empty
+          type: boolean
+        axes:
```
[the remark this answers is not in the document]
ebb-tide: I think this is out of scope for this PR, I have created an issue to address it https://github.com/influxdata/influxdb/issues/14560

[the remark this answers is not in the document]
chnn: It will be much easier to change this for the `CheckView` than for the `XYView` (since the `XYView` requires a migration), so I think they are somewhat separate. I also think it's worth addressing the `CheckView` now, since this is when we are first introducing it's schema. Otherwise we'll build off of it and make the change harder for ourselves in the future. But up to you!

http/swagger.yml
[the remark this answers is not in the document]
chnn: Can we just remove it here?

ui/src/alerting/components/builder/CheckMetaCard.tsx
```diff
@@ -0,0 +1,225 @@
...
+}
+
+interface StateProps {
+  check: Partial<Check>
+}
+
+type Props = DispatchProps & StateProps
+
+const CheckMetaCard: FC<Props> = ({updateCurrentCheck, check}) => {
+  const handleChangeType = (type: typeof check.type) => {
```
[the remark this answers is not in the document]
ebb-tide: love it, thanks!

ui/src/shared/components/ViewSwitcher.tsx
ebb-tide: Agreed! https://app.zenhub.com/workspaces/applicationmonitoring-5d2e06177e9542541ca076ad/issues/influxdata/influxdb/14537

ui/src/shared/components/XYPlot.tsx
ebb-tide: removed!

ui/src/timeMachine/components/view_options/OptionsSwitcher.tsx
ebb-tide: removed!

ui/src/timeMachine/reducers/index.test.ts
ebb-tide: deleted.

ui/src/alerting/actions/checks.ts
```diff
@@ -93,18 +108,35 @@ export const getCurrentCheck = (checkID: string) => async (
...
+    const checkWithOrg = {...check, query: draftQueries[0], orgID} as Check
+
+    const resp = check.id
+      ? await api.patchCheck({checkID: check.id, data: checkWithOrg})
+      : await api.postCheck({data: checkWithOrg})
 
-    if (resp.status !== 201) {
+    if (!(resp.status === 201 || resp.status === 200)) {
       throw new Error(resp.data.message)
     }
```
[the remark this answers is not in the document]
ebb-tide: thank you!

ui/src/shared/components/ThresholdMarkers.tsx
ebb-tide: https://github.com/influxdata/influxdb/issues/14537

ui/src/shared/components/ThresholdMarkers.tsx
ebb-tide: https://github.com/influxdata/influxdb/issues/14537

ui/src/alerting/components/builder/HelpButton.tsx
```diff
@@ -0,0 +1,19 @@
+// Libraries
+import React, {FunctionComponent} from 'react'
+
+// Components
+import {Button} from '@influxdata/clockface'
+
+const HelpButton: FunctionComponent = () => {
+  const handleClick = () => {}
```
chnn: haha no help for you!!

PR #14567, opened 2019-08-05
feat(task): Add task types

lyondhill:
Add a type to task's to allow tasks to be filterable.

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -6556,6 +6556,9 @@ components:
         id:
           readOnly: true
           type: string
+        type:
+          description: The type of task, this can be used for filtering tasks on list actions.
+          type: string
         orgID:
           description: The ID of the organization that owns this Task.
           type: string
@@ -8816,6 +8819,9 @@ components:
     TaskCreateRequest:
       type: object
       properties:
+        type:
+          description: The type of task, this can be used for filtering tasks on list actions.
+          type: string
         orgID:
           description: The ID of the organization that owns this Task.
           type: string
--- a/kv/task.go
+++ b/kv/task.go
@@ -227,6 +227,14 @@ func (s *Service) findTasksByUser(ctx context.Context, tx Tx, filter influxdb.Ta
 			continue
 		}
 
+		if filter.Type == nil {
+			ft := ""
+			filter.Type = &ft
+		}
+		if *filter.Type != influxdb.TaskTypeWildcard && *filter.Type != task.Type {
+			continue
+		}
+
 		ts = append(ts, task)
 
 		if len(ts) >= filter.Limit {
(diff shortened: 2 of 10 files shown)


http/swagger.yml
```diff
@@ -6551,6 +6551,9 @@ components:
         id:
           readOnly: true
           type: string
+        type:
+          description: The type of task, this can be used for filtering
+          type: string
```
chnn: Is there a fixed set of values that this `type` can be? If so, it would be helpful to add an `enum` field here.

lyondhill: Anything except the wildcard "*" no limits currently.

lyondhill:
@gshif We are currently not limiting task types (other than using a "wildcard" as the task type on create). I can make the description in the swagger more descriptive.

gshif:
That would be great

kv/task.go
```diff
@@ -296,6 +304,14 @@ func (s *Service) findTaskByOrg(ctx context.Context, tx Tx, filter influxdb.Task
...
+				ft := ""
+				filter.Type = &ft
+			}
+			if *filter.Type != influxdb.TaskTypeWildcard && *filter.Type != t.Type {
+				t = nil
+			}
+
 			// insert the new task into the list
 			if t != nil {
 				ts = append(ts, t)
```
GeorgeMac: Take it or leave it suggestion. I found the flow a little hard to parse and a bit concerned about all the pointers. I would suggest something like:
```go
if t != nil {
  typ := ""
  if *filter.Type != nil {
    typ = *filter.Type
  }

  // if the filter type matches task type or filter type is a wildcard
  if typ == t.Type  || typ == influxdb.TakeTypeWildcard {
    ts = append(ts, t)
  }
}
```
Actually it appears in the original code that `t` could be nil. At least that nil check guard suggests it could be. So perhaps you do need a nil check on t somewhere first, otherwise the `t.Type` could nil pointer dereference. (either that or t is never really nil which is actually more than likely).

task/servicetest/servicetest.go
```diff
@@ -1437,3 +1443,87 @@ option task = {
...
+	}
+
+	// get all tasks
+	wc := influxdb.TaskTypeWildcard
+	tasks, _, err = sys.TaskService.FindTasks(sys.Ctx, influxdb.TaskFilter{OrganizationID: &cr.OrgID, Type: &wc})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// this is to account for the concurrency test creating tasks in the same organization
```
GeorgeMac: Another take it or leave it:

If you want to get precise here you could update the CredsFunc to take a *testing.T here: https://github.com/influxdata/influxdb/blob/c4fd4d79865873795b02dbdd1bcf26b3fd3e6fcd/http/task_test.go#L70

And then move org creation into the creds func call as well.
Then you can properly scope the organisation to the subtest name. That way you don't share organisations between parallel test runs leading to unpredictable results. Could avoid someone in the future deleting that concurrency test and breaking this one for example.
Also this looks like only happens in the http/task_test.go as that defines its own CredFunc override which causes this problem.

lyondhill: woo fresh eyes are so nice. Thanks! that little change will be very good for this testing suite.

kv/task.go
lyondhill: good find. the call above `influxdb.IDFromString(string(v))` used to be able to return a nil task with no error, I am not sure it does that anymore. I do like the logic in your solution better, its more clean.

PR #14568, opened 2019-08-05
chore(storage): remove defer from SeriesIDSet funcs

jacobmarble:
While working on another issue, I noticed that `SeriesIDSet` uses `defer` a lot, needlessly. [`defer` is known to be slow](https://github.com/golang/go/issues/14939) so this is a cleanup.


sebito91:
I'm a bit on the fence with this one, as it does provide quite the legible breakdown of the functionality. I've had issues before stripping `defer` from code only to find out that I've missed an `Unlock` somewhere that's ultimately caused a panic. 

Are we certain this is something work optimizing?

jacobmarble:
There's some optimized code within the storage engine that isn't terribly readable, but does perform very well. I like knowing that the mutexes are released immediately, not whenever the runtime gets around to it. I haven't benchmarked this change.

With the 30% performance improvement coming in 1.13, and the Golang bug still open, I'll let this one go.

PR #14591, opened 2019-08-07
build(client): generate client on each build

121watts:


kelwang:
Just wonder if we can have a better name for slackRuleBase. Maybe SlackRulePart or SlackRuleDetail? What do you think?🤔

chnn:
We could also define it like this:

    SlackNotificationRule:
      allOf:
        - $ref: "#/components/schemas/NotificationRuleBase"
        - type: object
          required: [type, messageTemplate]
          properties:
            type:
              type: string
              enum: [slack]
            channel:
              type: string
            messageTemplate:
              type: string

To avoid having to give the specialization a name at all.

PR #14592, opened 2019-08-07
fix(ui): sooth honeybadger

chnn:
Closes https://github.com/influxdata/idpe/issues/4222
Closes https://github.com/influxdata/idpe/issues/4045

See commit descriptions. It will be easier to review each commit individually—the first commit contains the important changes, and the second commit is just updating a bagillion imports.


ui/src/shared/components/ErrorBoundary.tsx
```diff
@@ -24,6 +27,29 @@ class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState> {
...
+    //         in ErrorBoundary(MePage) (created by Connect(ErrorBoundary(MePage)))
+    //         in Connect(ErrorBoundary(MePage)) (created by RouterContext)
+    //         in SpinnerContainer (created by SetOrg)
+    //         ...
+    //
+    // We will extract "MePage" as the component name.
+    const componentName = errorInfo.componentStack
+      .trim()
+      .split('\n')
+      .map(s => s.split(' ')[1])[0]
```
121watts: This itself looks like a runtime error waiting to happen ahah.

chnn: Hah definitely. As long as `componentStack` is a string, it should be OK though.

PR #14593, opened 2019-08-07
fix(query/compile.go): time range was exceeding min/max bounds under …

aanthony1243:
…certain conditions

Closes #

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/query/compile.go
+++ b/query/compile.go
@@ -1127,9 +1127,19 @@ func (c *compiledStatement) Prepare(shardMapper ShardMapper, sopt SelectOptions)
 	// Modify the time range if there are extra intervals and an interval.
 	if !c.Interval.IsZero() && c.ExtraIntervals > 0 {
 		if c.Ascending {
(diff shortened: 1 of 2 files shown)


query/compile.go
```diff
@@ -1127,9 +1127,19 @@ func (c *compiledStatement) Prepare(shardMapper ShardMapper, sopt SelectOptions)
 	// Modify the time range if there are extra intervals and an interval.
 	if !c.Interval.IsZero() && c.ExtraIntervals > 0 {
 		if c.Ascending {
-			timeRange.Min = timeRange.Min.Add(time.Duration(-c.ExtraIntervals) * c.Interval.Duration)
+			newTime := timeRange.Min.Add(time.Duration(-c.ExtraIntervals) * c.Interval.Duration)
```
jsternberg: Will this check potentially lead to overflow or underflow? I think you need to check that the addition/subtraction will work before you actually perform it. I guess this works because `time.Time` isn't an `int64`?

PR #14595, opened 2019-08-08
feat(tasks): implement TreeScheduler

docmerlin:
Closes #

This is a refactor for the scheduler, to make it much more easy to test, and make it separable from the kv store.


task/backend/scheduler/treescheduler.go
```diff
@@ -0,0 +1,403 @@
...
+	"math"
+	"sync"
+	"time"
+
+	"github.com/cespare/xxhash"
+	"github.com/google/btree"
+)
+
+const (
+	maxWaitTime          = time.Hour
```
lyondhill: I want to know what these const's are and what they are for.

task/backend/scheduler/treescheduler.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+
+	s.sm = NewSchedulerMetrics(s)
+	s.when = s.time.Now().Add(maxWaitTime)
+	s.timer = s.time.NewTimer(maxWaitTime)
+	if executor == nil {
+		return nil, nil, errors.New("Executor must be a non-nil function")
+	}
+	s.wg.Add(1)
+	go func() {
```
lyondhill: can we break this function out to a separate function just for clarity

task/backend/scheduler/treescheduler.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+	if s.workchans == nil {
+		s.workchans = make([]chan Item, defaultMaxWorkers)
+
+	}
+
+	s.wg.Add(len(s.workchans))
+	for i := 0; i < len(s.workchans); i++ {
+		s.workchans[i] = make(chan Item)
+		go s.work(i)
```
lyondhill: I would rather see this call take a chan than an int. It is hard to follow what the int _means_. If for any reason we wanted to adjust the number of workers passing the index becomes kinda brittle.

task/backend/scheduler/treescheduler.go
```diff
@@ -0,0 +1,403 @@
...
+				s.Lock()
+				s.timer.Stop()
+				// close workchans
+				for i := range s.workchans {
+					close(s.workchans[i])
+				}
+				s.Unlock()
+				return
+			case <-s.timer.C():
+				s.Lock()
```
lyondhill: I think if you break this section of the case out you could have a cleaner lock -> defer unlock style. In the future someone may do an update and forget to unlock.

task/backend/scheduler/treescheduler.go
```diff
@@ -0,0 +1,403 @@
...
+				min := s.scheduled.Min()
+				if min == nil { // grab a new item, because there could be a different item at the top of the queue
+					s.when = s.time.Now().Add(maxWaitTime)
+					s.timer.Reset(maxWaitTime) // we can reset without stop, because its fired.
```
lyondhill: If we can't find an item we will wait an hour?

task/backend/scheduler/treescheduler.go
```diff
@@ -0,0 +1,403 @@
...
+	if !ok {
+		return
+	}
+
+	// delete the old task run time
+	s.scheduled.Delete(Item{id: taskID, ordering: ordering})
+	delete(s.nextTime, taskID)
+}
+
+// Release releases a task, if it doesn't own the task it just returns.
```
lyondhill: I don't see an early return in the function, maybe a comment update?

task/backend/scheduler/treescheduler.go
```diff
@@ -0,0 +1,403 @@
...
+// Release also cancels the running task.
+// Task deletion would be faster if the tree supported deleting ranges.
+func (s *TreeScheduler) Release(taskID ID) {
+	s.sm.release(taskID)
```
lyondhill: maybe this is where an early return might happen?

task/backend/scheduler/treescheduler.go
```diff
@@ -0,0 +1,403 @@
...
+		})
+	}
+	s.nextTime[it.id] = ordering{when: it.next + it.Offset + it.wait}
+
+	// insert the new task run time
+	s.scheduled.ReplaceOrInsert(it)
+	return nil
+}
+
+var maxItem = Item{
```
lyondhill: I don't think this is being used?

task/backend/scheduler/treescheduler.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+}
+
+// Executor is any function that accepts an ID, a time, and a duration.
+func NewScheduler(executor Executor, checkpointer SchedulableService, opts ...treeSchedulerOptFunc) (*TreeScheduler, *SchedulerMetrics, error) {
+	s := &TreeScheduler{
+		executor:     executor,
+		scheduled:    btree.New(degreeBtreeScheduled),
+		nextTime:     map[ID]ordering{},
+		onErr:        func(_ context.Context, _ ID, _ time.Time, _ error) {},
```
lyondhill: Maybe the default behavior is to do some logging (probably in debug)?

task/backend/scheduler/treescheduler.go
```diff
@@ -0,0 +1,403 @@
...
+func (s *TreeScheduler) process() {
+	iter, toReAdd := s.iterator(s.time.Now())
+	s.scheduled.Ascend(iter)
+	for i := range toReAdd.items {
+		s.nextTime[toReAdd.items[i].id] = toReAdd.items[i].ordering
+		s.scheduled.ReplaceOrInsert(toReAdd.items[i])
+	}
+}
+
+func (s *TreeScheduler) iterator(ts time.Time) (btree.ItemIterator, *unsent) {
```
lyondhill: im not sure I like the returning a pointer of unsent items. It feels like the unsent should be done as part of the iteration. Maybe you can elaborate on why this is necessary.

task/backend/scheduler/time.go
```diff
@@ -0,0 +1,237 @@
...
+
+// C returns a <-chan time.Time  and can be used much like time.Timer.C.
+func (t *stdTimer) C() <-chan time.Time {
+	return t.Timer.C
+}
+
+// MockTime is a time that mocks out some methods of time.Time.
+// It doesn't advance the time over time, but only changes it with calls to Set.
+// Use NewMockTime to create Mocktimes, don't instanciate the struct directly unless you want to mess with the sync Cond.
+type MockTime struct {
```
stuartcarnie: Unless there is a good case against it, can you look at using https://github.com/benbjohnson/clock in place of our own. We already have an indirect dependency on his framework.

docmerlin: Sounds like a good idea.  When I built the time wrapper, I didn't know about ben's library.  I'll see if I can switch it out.

task/backend/scheduler/treescheduler.go
```diff
@@ -0,0 +1,403 @@
...
+				s.Lock()
+				min := s.scheduled.Min()
+				if min == nil { // grab a new item, because there could be a different item at the top of the queue
+					s.when = s.time.Now().Add(maxWaitTime)
+					s.timer.Reset(maxWaitTime) // we can reset without stop, because its fired.
+					s.Unlock()
+					continue schedulerLoop
+				}
+				it := min.(Item)
+				if it.when > s.when.UTC().Unix() {
```
lyondhill: If this happens.. it looks like it would be not rescheduled again because the timer is not reset?

task/backend/scheduler/treescheduler.go
docmerlin: If there is no item in the tree at all, it will wait an hour and them check again (honestly, we could set it as long as we want without problem).  As soon as an item is added to the tree, the timer is reset to the appropriate time, so it just waits how long it needs to add.

lyondhill:
Could we just not set the timer at all?

task/backend/scheduler/treescheduler.go
lyondhill: I guess my question is why wouldn't we just not set a new timer? What value does setting it to 1 hour do for us?

task/backend/scheduler/treescheduler.go
```diff
@@ -0,0 +1,411 @@
...
+
+	// delete the old task run time
+	s.scheduled.Delete(Item{id: taskID, ordering: ordering})
+	delete(s.nextTime, taskID)
+}
+
+// Release releases a task.
+// Release also cancels the running task.
+// Task deletion would be faster if the tree supported deleting ranges.
+func (s *TreeScheduler) Release(taskID ID) {
```
stuartcarnie: I don't see any tests for `Release`; could you add them and then we're good to go!

PR #14600, opened 2019-08-08
feat(rule): edit notification rule overlay

121watts:
### Notes
This introduces a new pattern for sharing reducers across multiple components.  The most relevant files are `hooks/index.ts`, `NewRuleOverlay`, `EditRuleOverlay`, the reducer, and any children consuming the context provided `dispatch` function.  Would love to know what you all think.


ui/src/alerting/actions/notifications/rules.ts
```diff
@@ -67,9 +68,8 @@ export const getNotificationRules = () => async (
       throw new Error(resp.data.message)
     }
 
-    dispatch(
-      setAllNotificationRules(RemoteDataState.Done, resp.data.notificationRules)
-    )
+    // TODO(watts): replace with resp.data.notificationRules when implemented
+    dispatch(setAllNotificationRules(RemoteDataState.Done, [mockRule]))
```
chnn: Not relevant here but just on my mind: whenever we have a `DraftSomethingOrOther` resource, we'll need to convert to the draft form whenever receiving from the server, and convert back before persisting changes. I wonder if there's a certain approach we should be taking for this generally applicable problem.

In Ember for example, this is the responsibility of a ["serializer"](https://guides.emberjs.com/release/models/customizing-serializers/).

ui/src/alerting/actions/notifications/rules.ts
```diff
@@ -89,7 +89,11 @@ export const getCurrentRule = (ruleID: string) => async (
       throw new Error(resp.data.message)
     }
 
-    dispatch(setCurrentRule(RemoteDataState.Done, resp.data))
+    // TODO(watts): normalize this data to conform to NotificationRuleDraft
```
chnn: Ah yep I see you are thinking the same thing.

ui/src/alerting/components/notifications/EditRuleOverlay.tsx
```diff
@@ -0,0 +1,114 @@
...
+  Input,
+  Grid,
+  Columns,
+} from '@influxdata/clockface'
+
+// Reducers
+import {reducer, ActionPayload} from './RuleOverlay.reducer'
+
+// Constants
+import {endpoints} from 'src/alerting/constants' // Hooks
```
chnn:
```suggestion
import {endpoints} from 'src/alerting/constants'

// Hooks
```

ui/src/alerting/components/notifications/EditRuleOverlay.tsx
```diff
@@ -0,0 +1,114 @@
...
+  if (!stateRule) {
+    return null
+  }
+
+  const handleDismiss = () => {
+    router.push(`/orgs/${params.orgID}/alerting`)
+  }
+
+  const mode = RuleMode.Edit
+  const [rule, dispatch] = useReducer(reducer(mode), stateRule)
```
chnn: I wonder if it matters that this reducer is recreated every time this component renders.

Only time this bit me in the past is when I tried to create an impure reducer function. When the identity of a reducer changes from render to render, react will sometimes run the reducer twice for a single action, which tripped me up.

ui/src/alerting/components/notifications/EditRuleOverlay.tsx
```diff
@@ -0,0 +1,114 @@
...
+            </Overlay.Body>
+          </Overlay.Container>
+        </Overlay>
+      </EditRuleDispatch.Provider>
+    </RuleModeContext.Provider>
+  )
+}
+
+const mstp = ({rules}: AppState, {params}: Props): StateProps => {
+  const stateRule = rules.list.find(r => r.id === params.ruleID)
```
chnn: A cool thing about hooks I realized lately is that they provide really natural ways to write memoized redux selectors.

Instead of doing [this really awkward annoying thing](https://github.com/reduxjs/reselect#sharing-selectors-with-props-across-multiple-component-instances), you can pass the dependencies of the selector to the component and use `useMemo` instead. Seems nicer to me.

(Not relevant here, I don't think this needs a selector.)

ui/src/alerting/components/notifications/NewRuleOverlay.tsx
```diff
@@ -17,79 +17,83 @@ import {
...
+type Props = WithRouterProps
 
 const NewRuleOverlay: FC<Props> = ({params, router}) => {
   const handleDismiss = () => {
     router.push(`/orgs/${params.orgID}/alerting`)
   }
 
-  const [rule, dispatch] = useReducer(reducer, newRuleState)
+  const mode = RuleMode.New
+  const [rule, dispatch] = useReducer(reducer(mode), newRule)
```
chnn:
```suggestion
  const [rule, dispatch] = useReducer(reducer(mode), NEW_RULE)
```

Issue #4177, opened 2015-09-19
[0.9.4.1] Could not add 2nd data node in a cluster


pnpritchard:
I was able to add and run 3 raft nodes + 1 data node, however when I was trying to add the second data node (which is the 5th node in the cluster), the node could not start up and had the following error in the log :

```
2015/09/17 22:31:48 InfluxDB starting, version 0.9.4.1, branch 0.9.4, commit c4f85f84765e27bfb5e58630d0dea38adeacf543
2015/09/17 22:31:48 Go version go1.5, GOMAXPROCS set to 1
2015/09/17 22:31:48 Using configuration at: /config/config.toml
[retention] 2015/09/17 22:31:48 retention policy enforcement terminating
[monitor] 2015/09/17 22:31:48 shutting down monitor system
run: open server: open meta store: too many peers; influxdb v0.9.0 is limited to 3 nodes in a cluster
```

Original discussion: https://groups.google.com/forum/#!topic/influxdb/LWWrdJRt3kc
cc: @imlk

otoolep:
CC @jwilder @corylanou 

@jwilder - should adding a 2nd data-only node work?

jwilder:
Yes

jsternberg:
Closing this since clustering is no longer supported in the open source version.

PR #14601, opened 2019-08-08
fix(storage/reads): remove duplicate tables from the stream

jsternberg:
If the reader produces more than one table with the same group key, we
discard the later ones because the stream should never give us more than
one table with the same group key.

This is an error and it indicates the server sent us a bad set of data.
This change makes it so that the client is tolerant of that data and
will discard it if it exists.

#4177

### DIFF
--- a/mock/storage_reads.go
+++ b/mock/storage_reads.go
@@ -1,6 +1,9 @@
 package mock
 
 import (
+	"context"
+
+	"github.com/gogo/protobuf/proto"
 	"github.com/influxdata/influxdb/models"
 	"github.com/influxdata/influxdb/storage/reads"
 	"github.com/influxdata/influxdb/storage/reads/datatypes"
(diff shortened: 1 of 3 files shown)


mock/storage_reads.go
```diff
@@ -98,6 +101,38 @@ func (rs *GroupResultSet) Err() error {
 	return rs.ErrFunc()
 }
 
+type FloatArrayCursor struct {
+	CloseFunc func()
+	Errfunc   func() error
```
wolffcm: nit, the `f` in `Errfunc` could be uppercase to be consistent with the other member names.

mock/storage_reads.go
```diff
@@ -185,3 +220,49 @@ func (c *GroupCursor) Err() error {
 func (c *GroupCursor) Stats() cursors.CursorStats {
 	return c.StatsFunc()
 }
+
+type StoreReader struct {
+	ReadFilterFn func(ctx context.Context, req *datatypes.ReadFilterRequest) (reads.ResultSet, error)
```
wolffcm: nit, you use `func` as an abbreviation for "function" earlier in the file, would be good to choose one or the other.  I like `fn` myself.

storage/reads/reader.go
```diff
@@ -648,3 +648,27 @@ func (ti *tagValuesIterator) handleRead(f func(flux.Table) error, rs cursors.Str
...
+		f:    f,
+		seen: execute.NewGroupLookup(),
+	}
+	return filter.Process
+}
+
+func (df *duplicateFilter) Process(tbl flux.Table) error {
+	// Identify duplicate keys within the cursor.
+	if _, ok := df.seen.Lookup(tbl.Key()); ok {
+		// Discard this table.
```
wolffcm: Should we produce a log message here?  If this was happening a lot, then it would be a problem.

jsternberg: I was wondering that, but the logger didn't seem to have easy access here so I opted not to for now.

mock/storage_reads.go
jsternberg: This is duplicated from `IntegerArrayCursor` which was written by somebody else so I was thinking I would keep it the same for consistency.

PR #14603, opened 2019-08-08
feat(ui): add alert history page

### LINKED ISSUE #14393: Create Status History Page
Status history page is akin to a table view of rows in the status bucket, with a selectable query time range and a search input. 
Can be modeled after the tasks runs page for now. 
We may not have components in place to complete this right now, so it might be good to discuss with design as this view may also be used in a 2.x log viewer in the future. 
In this initial implementation, the UI can construct the query and send it directly to the query endpoint. It might be a good idea to consider if we should create an endpoint to generate the query on the backend.
Included in this issue is the task of deciding the query to run against the system bucket to get all checks. (should be able to pass time Range, limit to checkID)

It may be a good idea to implement limits on the rows returned to avoid overwhelming the browser. 
The page title should be: Status History
Table Columns should read: Time, Level, Check, Message, Tags

chnn:
Closes #14393

Adds a page for viewing check statuses. 

The page visualizes events from the status bucket in an infinite table:

1 infinite loading

When first displayed, the table records the current "now", and displays rows ordered from most recent to least recent.

The table can be refreshed to load any new statuses that have been written since you first viewed the page:

2 back to top

The table can also be searched:

4 searching

## Event Viewer

As you'll see in the code, I'm hoping this can provide the foundation for a more general event viewer that can be used in place of the table visualization.

To that end, there's a control for how many rows are loaded at once:

3 can control how many results loaded at once

This lets someone exploring their event data tailor their UI experience to their data. Data sets with less indexing are slower to search, so it may be desirable to use a lower row limit to make the table load faster. How indexed a data set is varies from user to user.

There are also a number of locations in the code where I have attempted to anticipate a broader use case:

- The fields displayed in the table are pluggable with custom components. For example, if we wanted to view syslog events, we could write components for displaying [syslog facility codes](https://en.wikipedia.org/wiki/Syslog#Syslog_message_components) and use the same table code

- The query that fetches results is also generic (must conform to the `LoadRows` interface)

- The table component can be used independently of the controls used on the alert history page. This lets us write custom controls for different event-viewing use cases, or put an event viewer in a dashboard cell with no controls

- An event viewer component owns it's own state entirely, so that multiple can be rendered at once easily (i.e. no redux or context)

## To Do

- [ ] Hook this up to an actual Flux query that retrieves statuses from the system bucket

Also probably want:

- [ ] Scrollbars 😱 
- [ ] Add more advanced ability to navigate time (I think this needs careful design)

Maybe some day in the future:

- [ ] Add ability to resize columns
- [ ] Add ability to sort table by columns (lot of open UX questions of what this actually means)

### DIFF
--- a/ui/src/eventViewer/components/EventViewer.reducer.ts
+++ b/ui/src/eventViewer/components/EventViewer.reducer.ts
@@ -1,0 +1,336 @@
+// Types
+import {Dispatch as ReactDispatch} from 'react'
+import {RemoteDataState} from 'src/types'
+import {Row, LoadRows, SearchExpr} from 'src/eventViewer/types'
+
+export interface State {
+  // All rows currently in the table
+  rows: Row[]
+
+  // The offset to use next time we load data
+  offset: number
+
+  // The number of rows to load next time we load data
+  limit: number
+
+  // The definition of "now" when running queries that are relative to "now"
+  now: number | null
+
+  // A expression used to filter results when we load rows
+  searchExpr: SearchExpr | null
+
+  // Tracks the loading status of the next rows being loading
+  nextRowsStatus: RemoteDataState
+
+  // If the next rows failed to load, this field should contain a
+  // human-friendly error message indicating why the rows failed to load
+  nextRowsErrorMessage: string | null
+
+  // A function that can be used to cancel any rows currently being loaded
+  nextRowsCanceller: null | (() => void)
+
+  // If a user has loaded all the rows that exist
+  hasReachedEnd: boolean
+
+  // When set, the table will render with this index in view
+  nextScrollIndex: number | null
+
+  // How many pixels into the table we have currently scrolled
+  scrollTop: number
+}
+
+export type Action =
+  | {type: 'NEXT_ROWS_LOADED'; rows: Row[]}
+  | {type: 'NEXT_ROWS_FAILED_TO_LOAD'; errorMessage: string}
+  | {type: 'NEXT_ROWS_LOADING'; cancel: () => void; now?: number}
+  | {type: 'SEARCH_ENTERED'; cancel: () => void; expr: SearchExpr; now: number}
+  | {type: 'SEARCH_COMPLETED'; rows: Row[]; limit: number}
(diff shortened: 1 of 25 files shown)


ui/src/eventViewer/utils/fakeLoadRows.ts
```diff
@@ -0,0 +1,60 @@
...
+const ROW_COUNT = 222
+
+const notFizzBuzz = (i: number): CheckStatusLevel => {
+  if (i % 3 === 0 && i % 5 === 0) {
+    return 'CRIT'
+  } else if (i % 5 === 0 || i % 3 === 0) {
+    return 'WARN'
+  } else {
+    return 'OK'
+  }
```
ebb-tide: amazing

ui/src/eventViewer/utils/fakeLoadRows.ts
```diff
@@ -0,0 +1,60 @@
+// This file provides a fake API and should be deleted soon
+
+import {range} from 'lodash'
+import {Row, LoadRows} from 'src/eventViewer/types'
+import {CheckStatusLevel, CancellationError} from 'src/types'
+
+const ROW_COUNT = 222
+
+const notFizzBuzz = (i: number): CheckStatusLevel => {
```
ebb-tide: amazing

ui/src/eventViewer/components/EventViewer.reducer.ts
```diff
@@ -0,0 +1,336 @@
+// Types
+import {Dispatch as ReactDispatch} from 'react'
+import {RemoteDataState} from 'src/types'
+import {Row, LoadRows, SearchExpr} from 'src/eventViewer/types'
+
+export interface State {
+  // All rows currently in the table
```
121watts: I'm loving these comments

ui/src/eventViewer/utils/fakeLoadRows.ts
121watts: bahahahah

PR #14622, opened 2019-08-12
Fix checks bugs

### LINKED ISSUE #14559: Add xy view properties to check view properties type
In the first iteration, check views will be identical to XY views for visualization purposes. 

Add all xy view properties to check view on the back end.

ebb-tide:
Closes #14559

### DIFF
--- a/ui/src/alerting/reducers/checks.ts
+++ b/ui/src/alerting/reducers/checks.ts
@@ -70,13 +70,13 @@ export default (
         if (action.payload.type == 'deadman') {
           draftState.current.check = {
             ...DEFAULT_DEADMAN_CHECK,
-            ...omit(exCheck, 'thresholds'),
+            ...omit(exCheck, ['thresholds', 'type']),
           } as DeadmanCheck
         }
         if (action.payload.type == 'threshold') {
           draftState.current.check = {
             ...DEFAULT_THRESHOLD_CHECK,
-            ...omit(exCheck, ['timeSince', 'reportZero', 'level']),
+            ...omit(exCheck, ['timeSince', 'reportZero', 'level', 'type']),
           } as ThresholdCheck
         }
         return
--- a/ui/src/dashboards/actions/index.ts
+++ b/ui/src/dashboards/actions/index.ts
@@ -22,6 +22,7 @@ import {
   notify,
   PublishNotificationAction,
 } from 'src/shared/actions/notifications'
+import {getChecks} from 'src/alerting/actions/checks'
 import {
   deleteTimeRange,
   updateTimeRangeFromQueryParams,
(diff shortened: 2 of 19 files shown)


ui/src/dashboards/actions/index.ts
```diff
@@ -362,6 +363,8 @@ export const getDashboardAsync = (dashboardID: string) => async (
 
     dispatch(setViews(RemoteDataState.Done, views))
 
+    dispatch(getChecks())
```
chnn: Is it OK for a dashboard to render before these are loaded? Or should this be inserted right after line 356?

ui/src/timeMachine/components/Vis.tsx
```diff
@@ -103,6 +106,7 @@ const TimeMachineVis: SFC<Props> = ({
               files={files}
               loading={loading}
               properties={resolvedViewProperties}
+              check={check}
```
chnn: I wonder if we would be better off connecting the `CheckPlot` to redux, rather than drilling the check from

```
Vis -> ViewSwitcher -> CheckPlot
```

and 
```
Cell -> View -> RefreshingView -> ViewSwitcher -> CheckPlot
```

I'm worried that all the intermediate stages will provide appealing places to introduce check-related logic in the future, even if that logic would be better off in redux or elsewhere.

ui/src/dashboards/selectors/index.ts
```diff
@@ -14,13 +14,16 @@ export const getView = (state: AppState, id: string): View => {
 
 export const getCheckForView = (
   state: AppState,
-  {properties}: View
+  view: View
 ): Partial<Check> => {
-  if (properties.type === 'check') {
-    const checkList = state.checks.list
-    const check = checkList.find(c => c.id == properties.checkID)
+  if (view) {
```
chnn: `_.get` could be helpful for avoiding the nested `if`s here:

```tsx
  const viewType: ViewType = get(view, 'properties.type')
  const checkID = get(view, 'properties.checkID')

  return viewType === 'check'
    ? state.checks.list.find(c => c.id === checkID)
    : null
```

ui/src/alerting/reducers/checks.ts
```diff
@@ -70,13 +70,13 @@ export default (
         if (action.payload.type == 'deadman') {
           draftState.current.check = {
             ...DEFAULT_DEADMAN_CHECK,
-            ...omit(exCheck, 'thresholds'),
+            ...omit(exCheck, ['thresholds', 'type']),
```
chnn: I wonder if a whitelist of still-valid-fields would be safer than using a blacklist here, given that a whitelist can be typechecked more effectively than a blacklist / `omit`.

ui/src/dashboards/selectors/index.ts
ebb-tide: thank you that's much prettier.

ui/src/alerting/reducers/checks.ts
ebb-tide: I feel like if checks change, we will forget to whitelist here, and we will overwrite fields we don't intend to.

PR #14627, opened 2019-08-12
chore(ui): disable eslint via fork-ts-checker-webpack-plugin

chnn:
Disables running ESLint through the fork-ts-checker-webpack-plugin in development, due to its tendency to overuse CPU.

I think this OK since linter errors, unlike typechecking errors, are usually confined to the file you are working within. So they're a good fit to run via an editor, rather than the build tool.


121watts:
Can't we just get you a new computer?

chnn:
Works for me!! I think something with ~128 cores should be sufficient for javascript tooling.

PR #14636, opened 2019-08-12
fix(flux): Fix hanging test by ensuring results are read

### LINKED ISSUE #14574: Flux e2e tests with big input datasets are too slow
Tests with big `inData` run slow. This seems to be due to the `to` transformation.

Even if https://github.com/influxdata/influxdb/pull/14547 got merged. They still run slow.

stuartcarnie:
Closes #14574


nathanielc:
@stuartcarnie Thanks! My guess is this became an issue with holt winters because its the first test case with more than one ColReader, since the csv.from defaults to batching 1000 rows into a single ColReader.

Issue #14581, opened 2019-08-06
threshold need to be typed


kelwang:
convert the current thresholds to 3 types: less, greater, range

PR #14639, opened 2019-08-13
feat(influxdb): Add notification endpoint

kelwang:
Closes https://github.com/influxdata/influxdb/issues/14329

add notification endpoint, package, kv, http, authorizer, launcher, test cases.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/authorizer/notification_endpoint.go
+++ b/authorizer/notification_endpoint.go
@@ -32,52 +32,6 @@ func NewNotificationEndpointService(
 	}
 }
 
-func newNotificationEndpointPermission(a influxdb.Action, orgID, id influxdb.ID) (*influxdb.Permission, error) {
-	return influxdb.NewPermissionAtID(id, a, influxdb.NotificationEndpointResourceType, orgID)
-}
-
-func authorizeReadNotificationEndpoint(ctx context.Context, orgID, id influxdb.ID) error {
-	p, err := newNotificationEndpointPermission(influxdb.ReadAction, orgID, id)
-	if err != nil {
-		return err
-	}
-
-	pOrg, err := newOrgPermission(influxdb.WriteAction, orgID)
-	if err != nil {
-		return err
-	}
-
-	err0 := IsAllowed(ctx, *p)
-	err1 := IsAllowed(ctx, *pOrg)
-
-	if err0 != nil && err1 != nil {
-		return err0
-	}
-
-	return nil
-}
-
-func authorizeWriteNotificationEndpoint(ctx context.Context, orgID, id influxdb.ID) error {
-	p, err := newNotificationEndpointPermission(influxdb.WriteAction, orgID, id)
-	if err != nil {
-		return err
-	}
-
-	pOrg, err := newOrgPermission(influxdb.WriteAction, orgID)
-	if err != nil {
-		return err
-	}
-
-	err0 := IsAllowed(ctx, *p)
-	err1 := IsAllowed(ctx, *pOrg)
-
-	if err0 != nil && err1 != nil {
-		return err0
-	}
-
-	return nil
-}
-
 // FindNotificationEndpointByID checks to see if the authorizer on context has read access to the id provided.
 func (s *NotificationEndpointService) FindNotificationEndpointByID(ctx context.Context, id influxdb.ID) (influxdb.NotificationEndpoint, error) {
 	edp, err := s.s.FindNotificationEndpointByID(ctx, id)
@@ -85,7 +39,7 @@ func (s *NotificationEndpointService) FindNotificationEndpointByID(ctx context.C
 		return nil, err
 	}
 
-	if err := authorizeReadNotificationEndpoint(ctx, edp.GetOrgID(), edp.GetID()); err != nil {
+	if err := authorizeReadOrg(ctx, edp.GetOrgID()); err != nil {
 		return nil, err
 	}
 
@@ -105,7 +59,7 @@ func (s *NotificationEndpointService) FindNotificationEndpoints(ctx context.Cont
 	// https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating
 	endpoints := edps[:0]
 	for _, edp := range edps {
-		err := authorizeReadNotificationEndpoint(ctx, edp.GetOrgID(), edp.GetID())
+		err := authorizeReadOrg(ctx, edp.GetOrgID())
 		if err != nil && influxdb.ErrorCode(err) != influxdb.EUnauthorized {
 			return nil, 0, err
 		}
@@ -149,7 +103,7 @@ func (s *NotificationEndpointService) UpdateNotificationEndpoint(ctx context.Con
 		return nil, err
 	}
 
-	if err := authorizeWriteNotificationEndpoint(ctx, edp.GetOrgID(), id); err != nil {
+	if err := authorizeWriteOrg(ctx, edp.GetOrgID()); err != nil {
 		return nil, err
 	}
 
@@ -163,7 +117,7 @@ func (s *NotificationEndpointService) PatchNotificationEndpoint(ctx context.Cont
 		return nil, err
 	}
 
-	if err := authorizeWriteNotificationEndpoint(ctx, edp.GetOrgID(), id); err != nil {
+	if err := authorizeWriteOrg(ctx, edp.GetOrgID()); err != nil {
 		return nil, err
 	}
 
@@ -177,7 +131,7 @@ func (s *NotificationEndpointService) DeleteNotificationEndpoint(ctx context.Con
 		return err
 	}
 
-	if err := authorizeWriteNotificationEndpoint(ctx, edp.GetOrgID(), id); err != nil {
+	if err := authorizeWriteOrg(ctx, edp.GetOrgID()); err != nil {
 		return err
 	}
 
--- a/authorizer/notification_endpoint.go
+++ b/authorizer/notification_endpoint.go
@@ -1,0 +1,185 @@
+package authorizer
+
+import (
+	"context"
+
+	"github.com/influxdata/influxdb"
+)
+
+var _ influxdb.NotificationEndpointService = (*NotificationEndpointService)(nil)
+
+// NotificationEndpointService wraps a influxdb.NotificationEndpointService and authorizes actions
+// against it appropriately.
+type NotificationEndpointService struct {
+	s influxdb.NotificationEndpointService
+	influxdb.UserResourceMappingService
+	influxdb.OrganizationService
+	influxdb.SecretService
+}
+
+// NewNotificationEndpointService constructs an instance of an authorizing notification endpoint serivce.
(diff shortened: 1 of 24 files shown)


http/swagger.yml
```diff
@@ -8930,54 +8925,19 @@ components:
...
               $ref: "#/components/schemas/CheckStatusLevel"
-    ThresholdBase:
+    Threshold:
       properties:
         level:
          $ref: "#/components/schemas/CheckStatusLevel"
         allValues:
           description: if true, only alert if all values meet threshold
           type: boolean
-    GreaterThreshold:
```
chnn: Is this temporary? I thought that we had settled on the exiting types: https://github.com/influxdata/influxdb/issues/14581

kelwang: yeah, looks like it's just come kind of conflict, let me remove this part.

chnn: gotcha, thanks!

notification_endpoint.go
```diff
@@ -0,0 +1,127 @@
...
+	ErrInvalidNotificationEndpointType = errors.New("unknown notification endpoint type")
+)
+
+// NotificationEndpoint is the configuration describing
+// how to call a 3rd party service. E.g. Slack, Pagerduty
+type NotificationEndpoint interface {
+	Valid() error
+	Type() string
+	json.Marshaler
+	Updator
```
jademcgough: Nit, but should be `Updater`

authorizer/notification_endpoint.go
```diff
@@ -0,0 +1,161 @@
...
+	return &NotificationEndpointService{
+		s:                          s,
+		UserResourceMappingService: urm,
+		OrganizationService:        org,
+		SecretService:              srt,
+	}
+}
+
+func newNotificationEndpointPermission(a influxdb.Action, orgID, id influxdb.ID) (*influxdb.Permission, error) {
+	return influxdb.NewPermissionAtID(id, a, influxdb.NotificationEndpointResourceType, orgID)
```
jademcgough: I think @desa suggested that we piggypack off of Organizations' permissions, to avoid having to introduce new types (looks like we introduced new permission types for checks and notification rules as well).

If we wanted to do that, this would be `influxdb.NewPermissionAtID(id, a, influxdb.OrgsResourceType, orgID)` instead.

I'd strongly prefer to do ^ because it's easier and less of a security risk, but if we do want to introduce new permission types, we'll need to:

1. Write a migration script to add these types to existing superadmin users.
2. Add these new types to the Lookup Service.

jademcgough:
The `patch` method for `/notificationEndpoints/{endpointID}` needs to be changed to a `put` here: https://github.com/influxdata/influxdb/blob/49118a56f359ef39237fd56ed696fe1d1998a524/http/swagger.yml#L5533

http/swagger.yml
```diff
@@ -9239,24 +9237,49 @@ components:
...
+            url:
+              type: string
+            token:
+              type: string
     PagerDutyNotificationEndpoint:
       type: object
       allOf:
         - $ref: "#/components/schemas/NotificationEndpointBase"
         - type: object
+          required: [url, routing-key]
```
jademcgough: routing-key is not a property of this object

http/swagger.yml
```diff
@@ -9239,24 +9237,49 @@ components:
...
+            url:
+              type: string
+            token:
+              type: string
     WebhookNotificationEndpoint:
       type: object
       allOf:
         - $ref: "#/components/schemas/NotificationEndpointBase"
         - type: object
+          required: [url]
```
jademcgough: `method` and `authmethod` should be required

http/swagger.yml
```diff
@@ -9239,24 +9237,49 @@ components:
...
     PagerDutyNotificationEndpoint:
       type: object
       allOf:
         - $ref: "#/components/schemas/NotificationEndpointBase"
         - type: object
+          required: [url, routing-key]
+          properties:
+            url:
+              type: string
+            token:
```
jademcgough: should be `routing-key`

notification/endpoint/webhook.go
```diff
@@ -0,0 +1,154 @@
...
+	if s.Username.Key != "" {
+		arr = append(arr, s.Username)
+	}
+	if s.Password.Key != "" {
+		arr = append(arr, s.Password)
+	}
+	return arr
+}
+
+var goodWebHookAuthMethod = map[string]bool{
```
jademcgough: 👍 for validating enums. This might be easier to read if the return value for these keys was `true`, and this:

`if _, ok := goodWebHookAuthMethod[s.AuthMethod]; !ok {`
was changed to this:
'if goodWebHookAuthMethod[s.AuthMethod] {`.

Same for other places we're doing this.

notification/endpoint/webhook.go
```diff
@@ -0,0 +1,154 @@
...
+		return &influxdb.Error{
+			Code: influxdb.EInvalid,
+			Msg:  "invalid webhook token for bearer auth",
+		}
+	}
+
+	return nil
+}
+
+type webhookAlias WebHook
```
jademcgough: What are these alias types for?

notification_endpoint.go
```diff
@@ -0,0 +1,127 @@
...
+	Getter
+	// SecretFields return available secret fields.
+	SecretFields() []SecretField
+	// BackfillSecretKeys fill back fill the secret field key during the unmarshalling
+	// if value of that secret field is not nil.
+	BackfillSecretKeys()
+	// ParseResponse will parse the response of
+	// the http request.
+	ParseResponse(resp *http.Response) error
```
jademcgough: I don't think ParseResponse is necessary; my understanding is that Flux is responsible for talking to 3rd party services, not us.

notification/endpoint/slack.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Slack is the notification endpoint config of slack.
+type Slack struct {
+	Base
+	// URL is a valid slack webhook URL
+	// TODO(jm): validate this in unmarshaler
+	// example: https://slack.com/api/chat.postMessage
+	URL string `json:"url"`
+	// Token is the bearer token for authorization
+	Token influxdb.SecretField `json:"token"`
```
jademcgough: I would strongly prefer not to store secrets in plaintext unless we have a good reason to do so.

notification/endpoint/webhook.go
```diff
@@ -0,0 +1,154 @@
...
+	webhookPasswordSuffix = "-password"
+)
+
+// WebHook is the notification endpoint config of webhook.
+type WebHook struct {
+	Base
+	// Path is the API path of WebHook
+	URL string `json:"url"`
+	// Token is the bearer token for authorization
+	Token           influxdb.SecretField `json:"token,omitempty"`
```
jademcgough: see ^

notification/endpoint/pagerduty.go
```diff
@@ -0,0 +1,111 @@
...
+const routingKeySuffix = "-routing-key"
+
+// PagerDuty is the notification endpoint config of pagerduty.
+type PagerDuty struct {
+	Base
+	// Path is the PagerDuty API URL, should not need to be changed.
+	URL string `json:"url"`
+	// RoutingKey is a version 4 UUID expressed as a 32-digit hexadecimal number.
+	// This is the Integration Key for an integration on any given service.
+	RoutingKey influxdb.SecretField `json:"routing-key"`
```
jademcgough: I would strongly prefer not to store secret fields here in plaintext.

kv/notification_endpoint.go
```diff
@@ -0,0 +1,544 @@
...
+		Msg:  fmt.Sprintf("Unable to connect to notification endpoint store service. Please try again; Err: %v", err),
+		Op:   "kv/notificationEndpoint",
+	}
+}
+
+// UnavailableNotificationEndpointIndexError is used when the error comes from an internal system.
+func UnavailableNotificationEndpointIndexError(err error) *influxdb.Error {
+	return &influxdb.Error{
+		Code: influxdb.EInternal,
+		Msg:  fmt.Sprintf("unexpected error retrieving notfication endpoint's index bucket; Err %v", err),
```
jademcgough: notfication -> notification

kv/notification_endpoint.go
```diff
@@ -0,0 +1,544 @@
...
+		Op:   "kv/notificationEndpointIndex",
+	}
+}
+
+// InternalNotificationEndpointStoreError is used when the error comes from an
+// internal system.
+func InternalNotificationEndpointStoreError(err error) *influxdb.Error {
+	return &influxdb.Error{
+		Code: influxdb.EInternal,
+		Msg:  fmt.Sprintf("Unknown internal notificationEndpoint data error; Err: %v", err),
```
jademcgough: maybe `notificationEndpoint` -> `notification endpoint` for consistency in our messages?

kv/notification_endpoint.go
```diff
@@ -0,0 +1,544 @@
...
+		_, pe := s.findOrganizationByID(ctx, tx, edp.GetOrgID())
+		if pe != nil {
+			return &influxdb.Error{
+				Op:  influxdb.OpCreateCheck,
+				Err: pe,
+			}
+		}
+	}
+	// notification endpoint name unique
+	if _, err := s.findNotificationEndpointByName(ctx, tx, edp.GetOrgID(), edp.GetName()); err == nil {
```
jademcgough: Is it necessary to enforce uniqueness by name? We don't have a name param in the API.

secret.go
```diff
@@ -25,3 +29,51 @@ type SecretService interface {
...
+	return "secret: " + string(s.Key)
+}
+
+// MarshalJSON implement the json marshaler interface.
+func (s SecretField) MarshalJSON() ([]byte, error) {
+	return json.Marshal(s.String())
+}
+
+// ClearValue will remove the value.
+func (s *SecretField) ClearValue() {
```
jademcgough: This is never used

secret.go
```diff
@@ -25,3 +29,51 @@ type SecretService interface {
 	// DeleteSecret removes a single secret from the secret store.
 	DeleteSecret(ctx context.Context, orgID ID, ks ...string) error
 }
+
+// SecretField contains a key string, and value pointer.
+type SecretField struct {
```
jademcgough: I think there's some value in keeping the concerns of the secret service as simple as possible, instead of adding marshaling/unmarshaling logic (which is more a concern of the http layer).

I'm also not sure if we need the `SecretField` struct. We shouldn't need to store secrets directly in this way to any struct written to bolt/etcd.

http/swagger.yml
kelwang: swagger changes is passed to @121watts , so he won't be blocked.

kv/notification_endpoint.go
kelwang: I think it's fine to make the name, since the dropdown of same name endpoints could confuse user.

PR #14648, opened 2019-08-13
feat(notification/check): generate task from check

### LINKED ISSUE #14331: when a check is created, it generates a task

desa:
Closes #14331

This PR generates a flux script by constructing the corresponding AST by hand and then passing that to `ast.Format`. Additionally, it creates a task when a check is created.

### DIFF
--- a/check.go
+++ b/check.go
@@ -15,6 +15,11 @@ const (
 type Check interface {
 	Valid() error
 	Type() string
+	ClearPrivateData()
+	SetTaskID(ID)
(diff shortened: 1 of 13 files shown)


desa:
I had to do quite a bit of coding around the alerts package not being available yet, but I think I'm in a good state.

kelwang:
mm, the only thing i not sure is integrate auth service into check. Feels it's better just update the authID to token from check struct. and let task worry about where that token is good or not

Issue #14619, opened 2019-08-12
Connect notification rule overlay to redux and backend


ebb-tide:
Saving notification rule should make API call, and add notification rule to backend. 
Connect notification rule card actions (like delete and rename) from <AlertingIndex> to backend and redux.

chnn:
WIP: https://github.com/influxdata/influxdb/pull/14678

Blocked on removal of `authorizationID` from the notification rule API.

PR #14659, opened 2019-08-14
refactor(ui): simplify notification rule state management

chnn:
(Working towards #14619).

1. @121watts and I used the notification rule overlays as an opportunity to workshop a new pattern for passing reducer state/dispatch via a multiton context. Despite the fact that I helped put this pattern in place, I was real confused trying to work in the same code a few days later. I think that's a sign that we can/should simplify the state management, which is what I've attempted here.

   The new approach is almost verbatim what is described [here](https://kentcdodds.com/blog/how-to-use-react-context-effectively), which strikes me as more simple. There's no more mode context or reducer factory, and any remaining provider logic is encapsulated into the `RuleOverlayProvider`. State and dispatch are still consumed via custom hooks, as before.

    There is "one weird trick" in addition to the linked article: the `RuleOverlayProvider` has a sprinkling of `useRef` to ensure that the initial reducer state is reset when a user switches between the new and edit rule overlays. But that's all buried in the `RuleOverlayProvider`, so we don't need to worry about its subtleties. 

2. I renamed some of the constant data we embed in the source, since I was getting confused on whether certain variables were supposed to be dynamic, constant, or temporary fixtures intended to be replaced.

3. I renamed the `AddID<T>` type to `WithClientID<T>`, and renamed it's `id` field to `cid`. I was finding differentiating between the very different `id` fields confusing.


ui/src/alerting/constants/index.ts
```diff
@@ -152,13 +152,13 @@ export const endpoints: NotificationEndpoint[] = [
     orgID: '1',
     userID: '1',
     description: 'interrupt someone by all means known to man',
-    name: 'pagerditty endpoint',
+    name: 'PagerDuty',
```
ebb-tide: REVERT

ui/src/alerting/components/notifications/RuleOverlay.reducer.tsx
```diff
@@ -149,7 +145,7 @@ export const reducer = (mode: RuleMode) => (
     case 'SET_TAG_RULE_OPERATOR': {
       const {tagRuleID, operator} = action
       const tagRules = state.tagRules.map(tagRule => {
-        if (tagRule.id !== tagRuleID) {
+        if (tagRule.cid !== tagRuleID) {
```
121watts: I like the "client id" namespace

ui/src/alerting/constants/index.ts
121watts:
```suggestion
    name: 'pagerditty',
```

ui/src/alerting/components/notifications/RuleOverlay.reducer.tsx
```diff
@@ -167,13 +163,47 @@ export const reducer = (mode: RuleMode) => (
...
+    return reducer(state, action)
+  }, initialState)
+
+  return (
+    <RuleStateContext.Provider value={state}>
+      <RuleDispatchContext.Provider value={dispatch}>
+        {children}
+      </RuleDispatchContext.Provider>
+    </RuleStateContext.Provider>
+  )
```
121watts: Thank you Mr. Dodds!

PR #14662, opened 2019-08-14
Feat/current check to time machine

### LINKED ISSUE #14640: Move current check from ChecksState to TimeMachineState

ebb-tide:
Closes #14640

### DIFF
--- a/ui/src/alerting/actions/checks.ts
+++ b/ui/src/alerting/actions/checks.ts
@@ -20,7 +20,6 @@ import {setCheckStatus, setTimeMachineCheck} from 'src/timeMachine/actions'
 
 // Types
 import {Check, GetState, RemoteDataState} from 'src/types'
-import {check1, check2} from 'src/alerting/constants'
 
 export type Action =
   | ReturnType<typeof setAllChecks>
@@ -61,7 +60,7 @@ export const getChecks = () => async (
     //   throw new Error(resp.data.message)
     // }
 
-    dispatch(setAllChecks(RemoteDataState.Done, [check1, check2]))
+    dispatch(setAllChecks(RemoteDataState.Done, []))
   } catch (e) {
     console.error(e)
     dispatch(setAllChecks(RemoteDataState.Error))
--- a/ui/src/alerting/actions/checks.ts
+++ b/ui/src/alerting/actions/checks.ts
@@ -16,12 +16,11 @@ import {
   Action as NotificationAction,
 } from 'src/shared/actions/notifications'
 import {Action as TimeMachineAction} from 'src/timeMachine/actions'
+import {setCheckStatus, setTimeMachineCheck} from 'src/timeMachine/actions'
 
 // Types
-import {RemoteDataState} from '@influxdata/clockface'
-import {Check, GetState} from 'src/types'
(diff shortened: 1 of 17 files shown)


ui/src/alerting/actions/checks.ts
```diff
@@ -15,19 +15,18 @@ import {
...
   Action as NotificationAction,
 } from 'src/shared/actions/notifications'
+import {Action as TimeMachineAction} from 'src/timeMachine/actions'
 
 // Types
 import {RemoteDataState} from '@influxdata/clockface'
-import {Check, GetState, CheckType} from 'src/types'
+import {Check, GetState} from 'src/types'
+import {check1, check2} from '../constants'
+import {setCheckStatus, setTimeMachineCheck} from 'src/timeMachine/actions'
```
chnn: Looks like the organization of these imports is a bit mixed up

ui/src/alerting/components/EditCheckEO.tsx
```diff
@@ -33,9 +32,9 @@ import {
 
 interface DispatchProps {
   updateCheck: typeof updateCheck
-  setCurrentCheck: typeof setCurrentCheck
-  getCurrentCheck: typeof getCurrentCheck
-  updateCurrentCheck: typeof updateCurrentCheck
+  setTimeMachineCheck: typeof setTimeMachineCheck
```
chnn: We used to have a linting rule that would prevent us from shadowing names in the same scope. Not a big deal but I found it helpful to distinguish between a function of type `T` versus one of type `(dispatch: Dispatch) => T` using an `on` prefix for the connected one (`onSetTimeMachineCheck`).

I don't think the `onSetTimeMachineCheck` style needs to be the one true way, just thought I would share.

ui/src/shared/components/CheckPlot.tsx
```diff
@@ -128,7 +128,7 @@ const CheckPlot: FunctionComponent<Props> = ({
           <ThresholdMarkers
             key="custom"
             thresholds={thresholds}
-            onSetThresholds={updateCheckThresholds}
+            onSetThresholds={updateTimeMachineCheckThresholds}
```
chnn: (For later) Seems like we need some way of making thresholds uneditable when a `CheckPlot` is used in a dashboard cell.

ui/src/alerting/components/EditCheckEO.tsx
ebb-tide: I go back and forth on this- I like to be able to follow a function from its definition to where its called without changing its name, But it is true that I have made the mistake of calling wrong function by forgetting to pass it though dispatch multiple times... so perhaps I should just conform to your suggestion.

ui/src/shared/components/CheckPlot.tsx
ebb-tide: It is currently technically uneditable, but making it explicit would be better

ui/src/alerting/components/EditCheckEO.tsx
ebb-tide: I'll change this here and use on in the future but I'm not going to go through all of this PR

PR #14668, opened 2019-08-15
fix(task): Unify task scheduling across create and query

lyondhill:
Current behavior is that the first execution of a task happens based on the create time
of the task when using a 'every' schedule. If you create a task at 12:02 and want
the task to run every 15m. The first execution would happen at 12:17, and the 2nd would happen
at 12:30.

To fix this behavior I refactored the kv task to give a single source of knowledge.
We now have one function for finding exactly what the last scheduled task was.
We also now have a single method that calculates when the next schedule is due.
By unifying the logic it should always work the same way weather your asking when to run
or when creating a task.


lyondhill:
@docmerlin I think you're right but because that code is just being moved I think I will add a TODO and we can make it more clear in a different PR.

Issue #4321, opened 2015-10-04
[pd-storage-engine] inaccurate values when using mean and group by


parisholley:
when I do a mean() on all values in a measurement, seems fine. soon as I group by time(1m), i get some whacky numbers, but it seems to only happen on some of my data, not all.

i then tried to run a max() on the same query with no interval and it crashed the server with the same error i have seen before:

```
goroutine 2790 [running]:
github.com/influxdb/influxdb/tsdb/engine/tsm1.DecodeBlock(0x7f4bb22f9754, 0x4, 0xc13b8, 0x0, 0x0, 0x0, 0x0, 0x0)
	/root/.gvm/pkgsets/go1.5/global/src/github.com/influxdb/influxdb/tsdb/engine/tsm1/encoding.go:127 +0x37f
github.com/influxdb/influxdb/tsdb/engine/tsm1.(*cursor).decodeBlock(0xc82669ad20, 0xe0040e748)
	/root/.gvm/pkgsets/go1.5/global/src/github.com/influxdb/influxdb/tsdb/engine/tsm1/cursor.go:437 +0x152
github.com/influxdb/influxdb/tsdb/engine/tsm1.(*cursor).seekAscending(0xc82669ad20, 0x140a0ff40ed3b614, 0x140a101e0040e463, 0x0, 0x0)
	/root/.gvm/pkgsets/go1.5/global/src/github.com/influxdb/influxdb/tsdb/engine/tsm1/cursor.go:288 +0x485
github.com/influxdb/influxdb/tsdb/engine/tsm1.(*cursor).SeekTo(0xc82669ad20, 0x140a0ff40ed3b614, 0xffffffffffffffff, 0x0, 0x0)
	/root/.gvm/pkgsets/go1.5/global/src/github.com/influxdb/influxdb/tsdb/engine/tsm1/cursor.go:248 +0x390
github.com/influxdb/influxdb/tsdb/engine/tsm1.(*combinedEngineCursor).SeekTo(0xc8349a3020, 0x140a0ff40ed3b614, 0x648, 0x0, 0x0)
	/root/.gvm/pkgsets/go1.5/global/src/github.com/influxdb/influxdb/tsdb/engine/tsm1/cursor.go:29 +0xc0
github.com/influxdb/influxdb/tsdb.(*TagsCursor).SeekTo(0xc830228cd0, 0x140a0ff40ed3b614, 0x9cfcc0, 0x0, 0x0)
[...]

PR #14674, opened 2019-08-15
fix(storage): Ensure ResponseWriter flushes all buffered data

stuartcarnie:
The `ResponseWriter` would truncate the last series if the byte size of the points frames exceeded the `writeSize` constant, causing a `Flush` to occur and the cumulative `ResponseWriter.sz` to reset to zero. Because `ResponseWriter.sz` was not incremented for each frame, it remained at zero, which resulted in the final `Flush` short circuiting.

This PR implements the `Size` method for the `cursors.Array` types to be used to estimate the size of frame. This is in place of calling the Protocol Buffer `Size` function, which can be very expensive for varint types.

### DIFF
--- a/storage/reads/response_writer_test.go
+++ b/storage/reads/response_writer_test.go
@@ -336,7 +336,7 @@ func TestResponseWriter_WriteResultSet(t *testing.T) {
 	})
 
 	t.Run("issues", func(t *testing.T) {
-		t.Run("#4321 short write", func(t *testing.T) {
+		t.Run("short write", func(t *testing.T) {
 			t.Run("single string series", func(t *testing.T) {
 				exp := sendSummary{seriesCount: 1, stringCount: 1020}
 				var ss sendSummary
--- a/storage/reads/response_writer_test.go
+++ b/storage/reads/response_writer_test.go
@@ -1,12 +1,21 @@
 package reads_test
 
 import (
+	"context"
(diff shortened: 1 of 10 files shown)


storage/reads/response_writer_test.go
```diff
@@ -123,3 +132,403 @@ func TestResponseWriter_WriteGroupResultSet_Stats(t *testing.T) {
...
+
+			cur := newSeriesGeneratorSeriesCursor(makeTypedSeries("m0", "t", "f0", strings.Repeat("0", 1000), 2000, 1))
+			rs := reads.NewFilteredResultSet(context.Background(), &datatypes.ReadFilterRequest{}, cur)
+			_ = w.WriteResultSet(rs)
+			assert.Equal(t, w.Err(), exp)
+		})
+	})
+
+	t.Run("issues", func(t *testing.T) {
+		t.Run("#4321 short write", func(t *testing.T) {
```
jacobmarble: Maybe this one should be called just "short write", since we'll forget about issue #4321 in a few days.

stuartcarnie: Yes, I will remove the issue reference. I had seen it in a couple of other places, but I don’t think it matters

stuartcarnie: Done

PR #14675, opened 2019-08-15
feat(task): impersonate user on task execution

lyondhill:
Passing tokens to tasks is cumbersome and we needed a way to more easily create tasks. With this change we no longer need a token on task create. We take the user that created the task and pass that in as the "owner". As far as the task is concerned the owner is the source of permissions.

This is done by adding an additional field on task create that is `OwnerID`. We will no longer respect the token passed in and it will be deprecated soon.

Things to do still:
Task updates need to allow for owners to be set.

### DIFF
--- a/http/task_service_test.go
+++ b/http/task_service_test.go
@@ -151,7 +151,7 @@ func TestTaskHandler_handleGetTasks(t *testing.T) {
         }
       ],
       "orgID": "0000000000000001",
-	  "ownerID": "0000000000000001",
+      "ownerID": "0000000000000001",
       "org": "test",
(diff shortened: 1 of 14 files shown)


http/task_service_test.go
```diff
@@ -149,9 +151,9 @@ func TestTaskHandler_handleGetTasks(t *testing.T) {
         }
       ],
       "orgID": "0000000000000001",
+	  "ownerID": "0000000000000001",
```
goller: I think these are different space formatting... I betcha you need a go fmt

PR #14678, opened 2019-08-15
feat(ui): integrate notification rule overlays with API

chnn:
Part of #14619

I connected what parts of the notification rule UI I could to the API, but ran into an issue with the required `authorizationID` field. This field needs to be removed, but has not been removed yet. Once it has, I'll finish this up.

save errors


kelwang:
swagger looks good for me

PR #14679, opened 2019-08-15
fix(notification/rule): remove authID requirement

kelwang:
remove the authorization id requirement for creating rule. 
related to https://github.com/influxdata/influxdb/pull/14675

replace with the user id. 
Manually tested

remove the smtp notification rule

- [x] Rebased/mergeable
- [x] Tests pass


kelwang:
> I don't understand- I think we should just remove AuthorizationID and not add an OwnerID. @desa could you check this please?

ownerID is read only, frontend will never pass that in the post

notification/rule/smtp.go
```diff
@@ -1,68 +0,0 @@
-package rule
```
goller: this should be in a different PR... not topical.

http/swagger.yml
```diff
@@ -9057,8 +9057,8 @@ components:
         orgID:
           description: the ID of the organization that owns this notification rule.
           type: string
-        authorizationID:
-          description: The ID of the authorization used to create this notification rule.
+        ownerID:
```
goller: @kelwang i think this should be a readonly field.  It should be sent via the POST/PATCH as owner is implied by the creator.

OwnerID would be gathered from the authorizer context

(context.Authorizer(ctx).GetUserID())

chnn:
At some point I think it would be great for us to start defining separate schemas in the swagger spec for request and response bodies—it would clarify issues like whether or not we are supposed to send the `ownerID` when creating notification rules.

goller:
@chnn agreed

PR #14687, opened 2019-08-16
fix(query): remove http.to and update test cases that used it

nathanielc:


nathanielc:
@wolffcm Yes, the Flux code was never executed, it was just place holder code to ensure that their manipluations of the task source didn't break anything. So changing the uses to `to` instead of `http.to` if semantically equivalent for the test cases.

PR #14693, opened 2019-08-16
feat(notifications): create notification endpoint overlay

### LINKED ISSUE #14396: Add create/edit Endpoint Overlay
As a user
when I click on create endpoint button on Alerting Page, or
when I click on the name of an  endpoint on the Alerting Page,
I see the Endpoint Overlay with appropriate routes.
Create endpoint opens overlay with sensible defaults.
Edit queries for endpoint parameters

Save Endpoint button should be disabled until user selects a destination.  
Destination options are currently slack/smtp/pagerduty/webhook
And input fields below destination dropdown should change depending on selection. 

Send test message is out of scope for this issue

121watts:
Closes #14396 (the create piece)

### DIFF
--- a/ui/src/alerting/components/endpoints/EndpointOverlayProvider.tsx
+++ b/ui/src/alerting/components/endpoints/EndpointOverlayProvider.tsx
@@ -1,0 +1,55 @@
+// Libraries
+import React, {
+  FC,
+  Dispatch,
+  createContext,
+  useReducer,
+  useRef,
+  useContext,
+} from 'react'
+
+// Reducer
+import {EndpointState, Action, reducer} from './EndpointOverlay.reducer'
+
+const EndpointStateContext = createContext<EndpointState>(null)
+const EndpointDispatchContext = createContext<Dispatch<Action>>(null)
+
+export const EndpointOverlayProvider: FC<{initialState: EndpointState}> = ({
+  initialState,
+  children,
+}) => {
+  const prevInitialStateRef = useRef(initialState)
+
+  const [state, dispatch] = useReducer(
+    (state: EndpointState, action: Action) => {
+      if (prevInitialStateRef.current !== initialState) {
+        prevInitialStateRef.current = initialState
+
+        return initialState
+      }
+
+      return reducer(state, action)
+    },
+    initialState
+  )
+
+  return (
+    <EndpointStateContext.Provider value={state}>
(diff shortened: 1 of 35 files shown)


ui/src/alerting/components/endpoints/EndpointOverlayProvider.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const useEndpointState = (): EndpointState => {
+  return useContext(EndpointStateContext)
+}
+
+export const useEndpointDispatch = (): Dispatch<Action> => {
+  return useContext(EndpointDispatchContext)
+}
+
+export const useEndpointReducer = (): [EndpointState, Dispatch<Action>] => {
+  return [useEndpointState(), useEndpointDispatch()]
```
chnn: The case I've seen against bundling state and dispatch in a single hook is that a component using the hook will re-render when either state or dispatch change, even if it only uses dispatch. And dispatch almost never changes.

Seems like a real minor issue though tbh.

ui/src/alerting/components/endpoints/EndpointTypeDropdown.tsx
```diff
@@ -0,0 +1,71 @@
...
+}
+
+interface Props {
+  selectedType: string
+  onSelectType: (type: NotificationEndpointType) => void
+}
+
+const types: EndpointType[] = [
+  {name: 'Slack', type: 'slack', id: 'slack'},
+  {name: 'Pagerduty', type: 'pagerduty', id: 'pagerduty'},
```
chnn:
```suggestion
  {name: 'PagerDuty', type: 'pagerduty', id: 'pagerduty'},
```
where's your respect for The Brand???

ui/src/alerting/constants/index.ts
```diff
@@ -103,6 +103,17 @@ export const NEW_TAG_RULE_DRAFT: TagRuleDraft = {
   },
 }
 
+export const NEW_ENDPOINT_DRAFT: NotificationEndpoint = {
```
chnn: For rules I ended up making `const newRuleDraft = (orgID: string) => ({orgID, ...})`, which was necessary to save the rule successfully. Seems like something similar will be needed here, but we can do it later.

kelwang:
Looks good overall

PR #14694, opened 2019-08-16
feat(tasks): add ability to find tasks by name

### LINKED ISSUE #12170: tasks: should be able to find task by name
tasks: should be able to find task by name
@goller asked that I add this.

- [ ] Check swagger to understand how other resources are searched by name
    - This should determine the type of search (prefix, exact, wildcard, etc)
- [ ] Replicate behavior for finding tasks

AlirieGray:
Closes #12170

Adds ability to find tasks with a name specified in the query parameters. An exact matched is used, to be consistent with the way filtering by name works with other resources such as buckets.

For example, a request to `http://localhost:9999/api/v2/tasks?name=mytask` will return the task(s) with the name `mytask`

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/kv/task.go
+++ b/kv/task.go
@@ -203,7 +203,7 @@ func (s *Service) findTasks(ctx context.Context, tx Tx, filter influxdb.TaskFilt
 	if filter.User != nil {
 		return s.findTasksByUser(ctx, tx, filter)
 	} else if org != nil {
-		return s.findTaskByOrg(ctx, tx, filter)
+		return s.findTasksByOrg(ctx, tx, filter)
 	}
 
 	return s.findAllTasks(ctx, tx, filter)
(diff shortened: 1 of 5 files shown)


kv/task.go
```diff
@@ -475,9 +485,26 @@ func (s *Service) findAllTasks(ctx context.Context, tx Tx, filter influxdb.TaskF
...
 	}
+
+	if filter.Name != nil {
+		ts = filterByName(ts, *filter.Name)
+	}
+
 	return ts, len(ts), err
 }
 
+func filterByName(ts []*influxdb.Task, name string) []*influxdb.Task {
```
lyondhill: Add tasks to name. Just so it's clear is not generic

AlirieGray:
@kelwang No they are not necessarily unique. If there are multiple tasks with the same name, it will just return an array with those tasks

PR #14695, opened 2019-08-16
feat(notifiction_rule): create notification rule task

desa:
This PR generates the base flux for a notification rule and the associated task. The flux that gets generated will likely need a bit of tweaking to match whats here https://github.com/influxdata/idpe/pull/4342#pullrequestreview-276100524, but this should be enough for people to build off of.


http/swagger.yml
```diff
@@ -9153,6 +9163,10 @@ components:
           type: string
         messageTemplate:
           type: string
+    SlackNotificationRule:
+      allOf:
+        - $ref: "#/components/schemas/NotificationRuleBase"
+        - $ref: "#/components/schemas/SlackNotificationRuleBase"
```
chnn: I think @121watts is editing these same schemas at the moment, just a heads up.

kv/notification_rule.go
```diff
@@ -91,6 +99,34 @@ func (s *Service) createNotificationRule(ctx context.Context, tx Tx, nr influxdb
...
+	//	return nil, err
+	//}
+
+	// TODO(desa): pass in non nil notification endpoint.
+	script, err := r.GenerateFlux(nil)
+	if err != nil {
+		return nil, err
+	}
+
+	tc := influxdb.TaskCreate{
```
chnn: I think this task probably needs a version too. If we change how the task is generated in the future, the version will make it easier to migrate the existing data.

notification/rule/http.go
```diff
@@ -0,0 +1,154 @@
...
+
+// HTTP is the notification rule config of http.
+type HTTP struct {
+	Base
+	URL string
+}
+
+// GenerateFlux generates a flux script for the http notification rule.
+func (s *HTTP) GenerateFlux(e influxdb.NotificationEndpoint) (string, error) {
+	// TODO(desa): still needs full implementation
```
chnn: what's missing here? (just curious)

notification/rule/http.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return flux.DefineVariable("http_endpoint", call)
+}
+
+func (s *HTTP) generateFluxASTStatusPipe() ast.Statement {
+	base := flux.Call(flux.Identifier("from"), flux.Object(flux.Property("bucket", flux.String("system_bucket"))))
+
+	calls := []*ast.CallExpression{}
+
+	// TODO(desa): make start negative of every
```
chnn: I think that would be

```
	every := &ast.UnaryExpression{Operator: ast.SubtractionOperator, Argument: (*ast.DurationLiteral)(s.Every)}
	calls = append(calls, flux.Call(flux.Identifier("range"), flux.Object(flux.Property("start", every))))
```

notification/rule/slack.go
```diff
@@ -13,6 +18,129 @@ type Slack struct {
 	MessageTemplate string `json:"messageTemplate"`
 }
 
+// GenerateFlux generates a flux script for the slack notification rule.
```
chnn: Seems like this shares a lot with the other generation code. Does it seem appropriate to remove the repetition?

notification/rule/http.go
desa: The flux we need to generate needs to match the flux here https://github.com/influxdata/idpe/pull/4342#pullrequestreview-276100524

I've had to to some coding around the fact that the alerts package isn't available in flux yet. Hence the jankiness here.

kv/notification_rule.go
desa: yeah for sure. Do tasks have a version?

notification/rule/slack.go
desa: Definitely want to get rid of the repetition. Was in a hurry trying to get the demo here straight.

There's definitely mixed concepts in this PR. It really should have been two PRs. One for generating the notification rule task and the other for generating the flux for a notification rule.

http/swagger.yml
desa: Good to know

kelwang:
Have some suggestions,
The generate ast func, should always include an endpoint object.

So the rule interface should have something like

Func GenerateAst(edp NotificationEndpoint)

In the kv, retrieve the endpoint object by Id.

desa:
@kelwang 

> The generate ast func, should always include an endpoint object.
> ...
> In the kv, retrieve the endpoint object by Id.

Yeah, that's what is in place now, but that means that we can't statically generate a task, which I think adds more complexity than I like. I'll go down that path for now, but long term I'd prefer if that the flux able to be produces statically, without any externalities.

notification/rule/slack.go
desa: @chnn for now I'm going to hold off on reducing any of the repetition. Opened https://github.com/influxdata/influxdb/issues/14702

desa:
Going to aggressively merge this so that things are ready for the demo today. If there's any issues I didn't address let me know and I'll open issues for them.

kv/notification_rule.go
```diff
@@ -353,6 +406,16 @@ func (s *Service) DeleteNotificationRule(ctx context.Context, id influxdb.ID) er
 }
 
 func (s *Service) deleteNotificationRule(ctx context.Context, tx Tx, id influxdb.ID) error {
+	r, err := s.findNotificationRuleByID(ctx, tx, id)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println(r.GetTaskID())
```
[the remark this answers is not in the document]
desa: good catch.

notification/rule/rule.go
```diff
@@ -42,17 +43,18 @@ type Base struct {
 	ID          influxdb.ID     `json:"id,omitempty"`
 	Name        string          `json:"name"`
 	Description string          `json:"description,omitempty"`
-	EndpointID  *influxdb.ID    `json:"endpointID,omitempty"`
+	EndpointID  influxdb.ID     `json:"endpointID,omitempty"`
```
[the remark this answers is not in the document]
desa: I don't think that it will, but I'll confirm.

notification/rule/http.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/notification"
+	"github.com/influxdata/influxdb/notification/flux"
+)
+
+// HTTP is the notification rule config of http.
+type HTTP struct {
+	Base
+	URL string
```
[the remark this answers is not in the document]
desa: For now, we're going to keep it here. The specifics of how rules vs endpoints works is a bit undefined and I'm not sure that I like the behavior of not being able to statically define a task from a rule.

PR #14701, opened 2019-08-19
chore(authorizer): simplify permissions for alerts and notifications

jademcgough:
This simplifies our permissions for a&n. A user with read/write permissions to an org has read/write permissions for a&n.

### DIFF
--- a/authorizer/check.go
+++ b/authorizer/check.go
@@ -26,36 +26,6 @@ func NewCheckService(s influxdb.CheckService, urm influxdb.UserResourceMappingSe
 	}
 }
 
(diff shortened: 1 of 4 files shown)


authorizer/org.go
```diff
@@ -35,6 +36,7 @@ func newOrgPermission(a influxdb.Action, id influxdb.ID) (*influxdb.Permission,
 
 func authorizeReadOrg(ctx context.Context, id influxdb.ID) error {
 	p, err := newOrgPermission(influxdb.ReadAction, id)
+	fmt.Printf("new org permission: %v", p)
```
kelwang: this line can be removed

PR #14712, opened 2019-08-19
feat(task): Remove token's from task structures

lyondhill:
We had previously removed token's from the task api but left the token in place in several locations in the stack.
Now we can cleanly remove the extra tokens.

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -8840,9 +8840,6 @@ components:
         flux:
           description: The Flux script to run for this task.
           type: string
-        token:
-          description: The token to use for authenticating this task when it executes queries.
-          type: string
       required: [flux, token]
     TaskUpdateRequest:
       type: object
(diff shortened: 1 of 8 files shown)


http/swagger.yml
```diff
@@ -8840,9 +8840,6 @@ components:
         flux:
           description: The Flux script to run for this task.
           type: string
-        token:
-          description: The token to use for authenticating this task when it executes queries.
-          type: string
       required: [flux, token]
```
kelwang: does it need to be removed as well?

lyondhill: I asked @jademcgough if we should remove it or deprecate we decided to just remove it.

kelwang: I mean i think u also want to remove the token from the required, right

lyondhill: oo your right Thanks!

PR #14719, opened 2019-08-20
chore(storage): remove metric storage_retention_checks_total

jacobmarble:
Fixes https://github.com/influxdata/idpe/issues/4366

When many buckets are hosted in a single InfluxDB instance, cardinality for the storage_retention_checks_total measurement becomes a material burden. The metric doesn't seem to be valuable, so we'll remove it.


jacobmarble:
alchemy please merge

You're the best, alchemy.

PR #14741, opened 2019-08-22
add metadata for system buckets when org is created

### LINKED ISSUE #14539: Create two system buckets when an org is created
A system bucket is defined as a bucket that cannot be deleted by a user. The user can do anything else they want to the bucket (read data, write data, query, delete data, etc), except delete the bucket. System buckets should be exposed to the user. All other actions to the system buckets should have the same rate limits as any other bucket. they should be differentiated in the UI.

the two system buckets: 
 - task_logs
 - checks_notifications

initial retention policy on both of these should be inf

jademcgough:
Closes #14539

This adds system bucket metadata, visible to users, when an organization is created.

Needs tests before merging, wanted feedback before writing them.

### DIFF
--- a/authorizer/bucket.go
+++ b/authorizer/bucket.go
@@ -2,12 +2,22 @@ package authorizer
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/influxdata/influxdb"
 	"github.com/influxdata/influxdb/kit/tracing"
 )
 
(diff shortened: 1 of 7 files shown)


jademcgough:
Reexamining this, still needs:
- [ ] CLI implementation
- [ ] Cleanup on org deletion

Issue #14326, opened 2019-07-15
as an API user, I can see what flux query a check will generate


jademcgough:

ebb-tide:
define swagger.

jademcgough:
Reopening this. We need a way to preview generated flux for a check without creating it.

kelwang:
@jademcgough , should i just add a post method in the /query endpoint? for preview

PR #14745, opened 2019-08-22
feat(http): checks query endpoint

kelwang:
Closes https://github.com/influxdata/influxdb/issues/14326

Add the endpoint in http, with test, swagger

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/check_service.go
+++ b/http/check_service.go
@@ -59,6 +59,7 @@ type CheckHandler struct {
 const (
 	checksPath            = "/api/v2/checks"
 	checksIDPath          = "/api/v2/checks/:id"
+	checksIDQueryPath     = "/api/v2/checks/:id/query"
 	checksIDMembersPath   = "/api/v2/checks/:id/members"
 	checksIDMembersIDPath = "/api/v2/checks/:id/members/:userID"
 	checksIDOwnersPath    = "/api/v2/checks/:id/owners"
@@ -84,6 +85,7 @@ func NewCheckHandler(b *CheckBackend) *CheckHandler {
 	h.HandlerFunc("POST", checksPath, h.handlePostCheck)
 	h.HandlerFunc("GET", checksPath, h.handleGetChecks)
 	h.HandlerFunc("GET", checksIDPath, h.handleGetCheck)
+	h.HandlerFunc("GET", checksIDQueryPath, h.handleGetCheckQuery)
 	h.HandlerFunc("DELETE", checksIDPath, h.handleDeleteCheck)
 	h.HandlerFunc("PUT", checksIDPath, h.handlePutCheck)
 	h.HandlerFunc("PATCH", checksIDPath, h.handlePatchCheck)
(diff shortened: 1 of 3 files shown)


http/swagger.yml
```diff
@@ -5322,6 +5322,33 @@ paths:
...
+          name: checkID
+          schema:
+            type: string
+          required: true
+          description: ID of check
+      responses:
+        '200':
+          description: the check query requested
+          content:
+            text/plain:
```
chnn:
```suggestion
            application/vnd.flux:
```

http/check_service.go
```diff
@@ -235,6 +237,28 @@ func (h *CheckHandler) handleGetChecks(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (h *CheckHandler) handleGetCheckQuery(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	h.Logger.Debug("check query retrieve request", zap.String("r", fmt.Sprint(r)))
```
goller: This could be pretty heavy weight.  If you want to debug something out of here, would you grab specific request fields?

http/check_service.go
```diff
@@ -235,6 +237,28 @@ func (h *CheckHandler) handleGetChecks(w http.ResponseWriter, r *http.Request) {
...
+		h.HandleHTTPError(ctx, err, w)
+		return
+	}
+	flux, err := chk.GenerateFlux()
+	if err != nil {
+		h.HandleHTTPError(ctx, err, w)
+		return
+	}
+	h.Logger.Debug("check query retrieved", zap.String("check query", flux))
+	fmt.Fprintln(w, flux)
```
goller: The errors are JSON and the output is text.  This feels unpredictable as an API.  Perhaps respond with JSON?

http/check_test.go
```diff
@@ -314,6 +316,143 @@ func mustDuration(d string) *notification.Duration {
...
+				}))
+
+			w := httptest.NewRecorder()
+
+			h.handleGetCheckQuery(w, r)
+
+			res := w.Result()
+			content := res.Header.Get("Content-Type")
+			body, _ := ioutil.ReadAll(res.Body)
+			t.Logf(res.Header.Get("X-Influx-Error"))
```
goller: Do you mind moving this logging of this test into only test failures (by the errors)

http/check_service.go
kelwang: Our "/write" endpoint shows text/plain as response, if any error happens it's still an json. 
Do you think it's worth to do something similar to telegraf response? Based on the "Accept" header we show different responses?

goller: /write should not be a text/plain I wouldn't think.

Errors being returned from /write should be in JSON.

An `Accept` header would be the way to handle this if you really think it is important to have plaintext.  However, if you do plaintext you would need an error representation that was also plaintext.

http/check_service.go
kelwang: I copied from the previous model, shall we take care this issue in another ticket and update them all? I believe we will need the header, request body, the request url, not sure if we need the request ip, probably yes?

goller: I would remove it from this PR, yes.  The further on issue could talk about what kind of debug logs one would want in the API.

http/swagger.yml
```diff
@@ -8861,6 +8888,10 @@ components:
         offset:
           description: Override the 'offset' option in the flux script.
           type: string
+    FluxResponse:
```
goller: Do you mind adding a description here and for the flux property?

PR #14779, opened 2019-08-23
refactor(tasks): separate coordinator and middleware

GeorgeMac:
This is a WIP

This separates the current task coordinators reponsibilities into both a coordinator type and a task service middleware.

The task service middleware handles delegation to a persistent task service implementation and then notifies a middlware.Coordinator implementation that the action has taken place.
It is the job of the new `coordinator.Coordinator` to delegate the appropriate actions to a `backend.Scheduler` implementation.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/task/backend/coordinator.go
+++ b/task/backend/coordinator.go
@@ -1,0 +1,60 @@
+package backend
+
+import (
+	"context"
+	"time"
+
+	"github.com/influxdata/influxdb"
+	"go.uber.org/zap"
+)
+
+var now = func() time.Time {
+	return time.Now().UTC()
+}
+
+// TaskService is a type on which tasks can be listed
+type TaskService interface {
+	FindTasks(context.Context, influxdb.TaskFilter) ([]*influxdb.Task, int, error)
+	UpdateTask(context.Context, influxdb.ID, influxdb.TaskUpdate) (*influxdb.Task, error)
+}
+
+// Coordinator is a type with a single method which
+// is called when a task has been created
+type Coordinator interface {
+	TaskCreated(context.Context, *influxdb.Task) error
+}
+
+// NotifyCoordinatorOfExisting lists all tasks by the provided task service and for
+// each task it calls the provided coordinators task created method
+func NotifyCoordinatorOfExisting(ctx context.Context, ts TaskService, coord Coordinator, logger *zap.Logger) error {
+	// If we missed a Create Action
+	wildcard := influxdb.TaskTypeWildcard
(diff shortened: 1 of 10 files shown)


task/backend/coordinator.go
```diff
@@ -0,0 +1,43 @@
...
+
+// Coordinator is a type with a single method which
+// is called when a task has been created
+type Coordinator interface {
+	TaskCreated(context.Context, *influxdb.Task) error
+}
+
+// NotifyCoordinatorOfCreated lists all tasks by the provided task service and for
+// each task it calls the provided coordinators task created method
+func NotifyCoordinatorOfCreated(ctx context.Context, ts TasksFinder, coord Coordinator) error {
```
lyondhill: I like this change. Instead of the the old way. (nit: i would maybe change "created" to "existing" so we are notifying the coordinator of existing tasks)

task/backend/coordinator.go
```diff
@@ -0,0 +1,43 @@
...
+	// If we missed a Create Action
+	tasks, _, err := ts.FindTasks(ctx, influxdb.TaskFilter{})
+	if err != nil {
+		return err
+	}
+
+	for len(tasks) > 0 {
+		for _, task := range tasks {
+			coord.TaskCreated(ctx, task)
```
lyondhill: One thing that we used to do is if influxdb had been shot down for any amount of time we would not want the oss version to do a "catch up" for the time the task's were down. I am not sure if that code was moved somewhere else but I suspect we will still want to update the existing tasks with a latestCompleted time to time.Now.

task/backend/coordinator.go
```diff
@@ -0,0 +1,43 @@
...
+// is called when a task has been created
+type Coordinator interface {
+	TaskCreated(context.Context, *influxdb.Task) error
+}
+
+// NotifyCoordinatorOfCreated lists all tasks by the provided task service and for
+// each task it calls the provided coordinators task created method
+func NotifyCoordinatorOfCreated(ctx context.Context, ts TasksFinder, coord Coordinator) error {
+	// If we missed a Create Action
+	tasks, _, err := ts.FindTasks(ctx, influxdb.TaskFilter{})
```
lyondhill: A new addition here that was missed before. We will probably need to set the wildcard on the filter so we pick up "checks and notifications" `influxdb.TaskFilter{Type: & influxdb.TaskTypeWildcard}`.

task/backend/coordinator.go
GeorgeMac: oooh good catch

Issue #14609, opened 2019-08-09
Better error handling


desa:
There are a number of places where we do not display the error we get back from the API. This can be confusing to end users. It would be great to have a wholistic approach to how we handle errors in the front-end.

chnn:
Random thoughts:

- I think the most positive change we can make in the UI is to specify error states in mockups to begin with. We almost never do, and I think the result is engineers not thinking about the error path much (or if they do, it's just to display a notification, which is a bit lazy)

- I've noticed some areas where the API could be more consistent. For example, writing a syntactically invalid Flux query returns a `500` error when a `400` or `422` is likely more appropriate. It also returns the useful error message in the `error` field on the JSON response, rather than the `message` field like most other endpoints

- #14482 should help us know about what errors we should be handling while in development

PR #14784, opened 2019-08-23
feat: Alerting context menus

### LINKED ISSUE #14760: Fix Check/Rule/Endpoint Contexts
Context menus on check/rule/endpoints should not have export, should have clone, should have edit that opens the check editor overlay

ebb-tide:
Closes #14760

### DIFF
--- a/ui/src/alerting/actions/notifications/endpoints.ts
+++ b/ui/src/alerting/actions/notifications/endpoints.ts
@@ -77,42 +77,34 @@ export const getEndpoints = () => async (
 export const createEndpoint = (data: NotificationEndpoint) => async (
   dispatch: Dispatch<Action | NotificationAction>
 ) => {
-  try {
-    const resp = await api.postNotificationEndpoint({data})
+  const resp = await api.postNotificationEndpoint({data})
 
-    if (resp.status !== 201) {
-      throw new Error(resp.data.message)
-    }
-
-    dispatch({
-      type: 'SET_ENDPOINT',
-      endpoint: resp.data,
-    })
-  } catch (e) {
-    dispatch(notify(copy.createEndpointFailed(e.message)))
+  if (resp.status !== 201) {
+    throw new Error(resp.data.message)
   }
+
+  dispatch({
+    type: 'SET_ENDPOINT',
+    endpoint: resp.data,
+  })
 }
 
 export const updateEndpoint = (endpoint: NotificationEndpoint) => async (
   dispatch: Dispatch<Action | NotificationAction>
 ) => {
-  try {
-    const resp = await api.putNotificationEndpoint({
-      endpointID: endpoint.id,
-      data: endpoint,
-    })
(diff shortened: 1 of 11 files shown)


ui/src/alerting/actions/notifications/endpoints.ts
```diff
@@ -59,75 +69,92 @@ export const getEndpoints = () => async (
...
     dispatch({type: 'SET_ALL_ENDPOINTS', status: RemoteDataState.Error})
   }
 }
 
 export const createEndpoint = (data: NotificationEndpoint) => async (
-  dispatch: Dispatch<Action>
+  dispatch: Dispatch<Action | NotificationAction>
 ) => {
-  const resp = await api.postNotificationEndpoint({data})
+  try {
```
chnn: This `try / catch` block was left out before so that the endpoint creation form could catch any errors and display an error message inside the form, rather than as a notification. I think that's a nicer UX and would like to keep it—is there any reason this needs to catch errors and display a notification?

It looks like `cloneEndpoint` would still have the same catch behavior if you removed the `try/catch` wrapping, since it also catches and displays a notification.

Related: https://github.com/influxdata/influxdb/issues/14609

ui/src/alerting/actions/notifications/endpoints.ts
```diff
@@ -59,75 +69,92 @@ export const getEndpoints = () => async (
...
 }
 
 export const updateEndpoint = (endpoint: NotificationEndpoint) => async (
   dispatch: Dispatch<Action | NotificationAction>
 ) => {
-  const resp = await api.putNotificationEndpoint({
-    endpointID: endpoint.id,
-    data: endpoint,
-  })
+  try {
```
chnn: (Same comment about `try / catch`)

ui/src/alerting/actions/notifications/endpoints.ts
```diff
@@ -59,75 +69,92 @@ export const getEndpoints = () => async (
...
 
 export const updateEndpointProperties = (
   endpointID: string,
   properties: NotificationEndpointUpdate
 ) => async (dispatch: Dispatch<Action | NotificationAction>) => {
-  const resp = await api.patchNotificationEndpoint({
-    endpointID,
-    data: properties,
-  })
+  try {
```
chnn: (Same comment about `try / catch`)

ui/src/alerting/actions/notifications/endpoints.ts
ebb-tide: Thanks so much- I thought it was an oversight. will remove.

ebb-tide: https://github.com/influxdata/influxdb/issues/14790

ebb-tide: removing from add and edit keeping notifications for other things

Issue #14801, opened 2019-08-26
Create pipeline test for using a secret


jsternberg:
Create a pipeline test for creating and using a secret. There should also be another pipeline test that validates a user who does not have access to a secret cannot use the secret.

This pipeline test should interact with the OSS implementation of the secret service which uses bolt as the backend.

The pipeline tests are located in the `cmd/influxd/launcher` package.

PR #14813, opened 2019-08-26
tests(cmd/influxd/launcher): add a launcher test for the secret service

jsternberg:
The secret service is tested by creating a secret and then attempting to
use it in a flux query. There is one test where accessing the secret
should work and one where it should return that the action is forbidden.

#14801

### DIFF
--- a/cmd/influxd/launcher/query_test.go
+++ b/cmd/influxd/launcher/query_test.go
@@ -3,6 +3,7 @@ package launcher_test
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	nethttp "net/http"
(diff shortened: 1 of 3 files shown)


cmd/influxd/launcher/query_test.go
```diff
@@ -147,3 +151,107 @@ func TestPipeline_QueryMemoryLimits(t *testing.T) {
...
+			{
+				Action: influxdb.ReadAction,
+				Resource: influxdb.Resource{
+					Type:  influxdb.BucketsResourceType,
+					ID:    &l.Bucket.ID,
+					OrgID: &l.Org.ID,
+				},
+			},
+		},
+	}
```
affo: So, this auth is only for reading from a bucket, and not to  communicate with the secret store?

Is it why  it is forbidden?

affo:
@jsternberg I was trying to use Vault in your branch as @goller suggested:

```go
func TestPipeline_Query_LoadSecret_Success(t *testing.T) {
	l := launcher.RunTestLauncherOrFail(t, ctx, "--secret-store", "vault")
	l.SetupOrFail(t)
	defer l.ShutdownOrFail(t, ctx)
...
}
```

And I get a connection refused error:

```
--- FAIL: TestPipeline_Query_LoadSecret_Success (4.50s)
    query_test.go:162: unexpected error: Get https://127.0.0.1:8200/v1/secret/data/0460c34a68403000: dial tcp 127.0.0.1:8200: connect: connection refused
```

Do you think this PR should include that? Or should we open a new issue for it?

I am working on the IDPE counterpart anyways.

Issue #14836, opened 2019-08-28
Check and Notification Rule Task creation should check the status


kelwang:
Currently, a check task or a notification rule task can be created even the status is inactive.

The expected behavior, if any one create an inactive check or update the check to be inactive, the task should be removed.

PR #14857, opened 2019-08-28
fix(kv): fix the status check of check and notification

kelwang:
Closes https://github.com/influxdata/influxdb/issues/14836

Fixed the inactive check or notification rule, shouldn't be able to create a task. Fixed some test cases, but still unable to enable the test.

- [x] Rebased/mergeable
- [x] Tests pass


kv/notification_rule.go
```diff
@@ -203,6 +224,18 @@ func (s *Service) patchNotificationRule(ctx context.Context, tx Tx, id influxdb.
 		nr.SetStatus(*upd.Status)
 	}
 	nr.SetUpdatedAt(s.TimeGenerator.Now())
+
+	t, err := s.updateTask(ctx, tx, nr.GetTaskID(),
```
lyondhill: could you use the new `updateNotificationTask` function here?

desa:
@kelwang why was this merged when it breaks idpe?

PR #14858, opened 2019-08-28
refactor(dashboards): move time series fetcing descision to TimeSeries

121watts:
### What
- Move the `IntersectionObserver` logic out of the `Cell` component and into the `TimeSeries` component
- Removed state from `DashboardPage` and moved the "inView" logic into `TimeSeries`

### Why
- Consolidates all logic of wether a `Cell` should fetch data in one place
- Allows @alexpaxton to replace all local page components with `@influxdata/clockface`
- Enables @alexpaxton to remove `FancyScrollbars`
- Enables extracting `ReactGridLayout` and styles into presentational components that can be shared across multiple projects (like Quartz)

### DIFF
--- a/ui/src/dashboards/components/Dashboard.tsx
+++ b/ui/src/dashboards/components/Dashboard.tsx
@@ -1,4 +1,4 @@
-import React, {PureComponent, MouseEvent} from 'react'
+import React, {PureComponent} from 'react'
 import classnames from 'classnames'
 
 import Cells from 'src/shared/components/cells/Cells'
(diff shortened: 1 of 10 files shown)


ebb-tide:
Why did we do this?

alexpaxton:
@ebb-tide Daniel & myself have laid out some changes to dashboards we want to do, such as moving the presentational elements into Clockface. Some of these changes unblock all that.

The PR started out as an investigation and ended up as usable code

PR #14868, opened 2019-08-29
feat: Add matching rules cards

### LINKED ISSUE #14768: Implement "matching notification rules" in check builder
API is in progress: https://github.com/influxdata/influxdb/pull/14793

Notification rules card should show rule name, rule.every and name of endpoint associated with rule. 

Matched rules should be re-queried: whenever check.tags change and check.thresholds changes. 

image

ebb-tide:
Closes #14768

### DIFF
--- a/ui/src/alerting/components/builder/CheckMatchingRulesCard.tsx
+++ b/ui/src/alerting/components/builder/CheckMatchingRulesCard.tsx
@@ -5,6 +5,8 @@ import {get} from 'lodash'
 
 // Components
 import MatchingRuleCard from 'src/alerting/components/builder/MatchingRuleCard'
+import {SpinnerContainer} from '@influxdata/clockface'
+import {TechnoSpinner} from '@influxdata/clockface'
 
 // Actions
 import {getActiveTimeMachine} from 'src/timeMachine/selectors'
@@ -14,7 +16,7 @@ import * as api from 'src/client'
 
 //Types
 import {NotificationRule, Check, AppState, CheckTagSet} from 'src/types'
-import {EmptyState, ComponentSize} from '@influxdata/clockface'
+import {EmptyState, ComponentSize, RemoteDataState} from '@influxdata/clockface'
 
 interface StateProps {
   check: Partial<Check>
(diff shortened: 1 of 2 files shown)


ui/src/alerting/components/builder/CheckMatchingRulesCard.tsx
```diff
@@ -1,19 +1,86 @@
...
+  orgID: string
+}
+
+const CheckMatchingRulesCard: FunctionComponent<StateProps> = ({
+  check,
+  orgID,
+}) => {
+  const getMatchingRules = async (): Promise<NotificationRule[]> => {
+    const tags: CheckTagSet[] = get(check, 'tags', [])
+    const tagString = tags.map(t => `tag=${t.key}:${t.value}`).join('&')
```
chnn: Check out the [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLsearchParams) API for doing this sort of thing... it's pretty nice and handles URL escaping and all that

ui/src/alerting/components/builder/CheckMatchingRulesCard.tsx
```diff
@@ -1,19 +1,86 @@
...
+interface StateProps {
+  check: Partial<Check>
+  orgID: string
+}
+
+const CheckMatchingRulesCard: FunctionComponent<StateProps> = ({
+  check,
+  orgID,
+}) => {
+  const getMatchingRules = async (): Promise<NotificationRule[]> => {
```
chnn: wanna extract this somewhere else, and pass in the `check` and `orgID` as arguments? I think it would be nice to keep it separate from the component / view logic, since it is complex enough.

ui/src/alerting/components/builder/CheckMatchingRulesCard.tsx
```diff
@@ -1,19 +1,86 @@
...
+    const tags: CheckTagSet[] = get(check, 'tags', [])
+    const tagString = tags.map(t => `tag=${t.key}:${t.value}`).join('&')
+
+    // todo also: get tags from query results
+    const resp = await api.getNotificationRules({
+      query: {orgID, tag: tagString},
+    })
+
+    if (resp.status !== 200) {
+      // throw new Error(resp.data.message)
```
chnn: It would be great to set an error message in local state and display it somewhere inside this component... that's really my hope for how this typed error response stuff shakes out—that we can do more than catch errors and display generic notifications.

ui/src/alerting/components/builder/CheckMatchingRulesCard.tsx
```diff
@@ -1,19 +1,86 @@
...
+      return
+    }
+
+    setMatchingRules(resp.data.notificationRules)
+  }
+
+  const [matchingRules, setMatchingRules] = useState<NotificationRule[]>([])
```
chnn: I think this ought to have a loading state.

ui/src/alerting/components/builder/CheckMatchingRulesCard.tsx
chnn: Actually I'm not sure this will work either way, [oats](https://github.com/chnn/oats) probably needs to be updated to handle serializing an array of values as a query parameters.

Different APIs encode array query parameters differently. Trying to assign the array `['foo', 'bar', 'baz']` to the query paramater `filter` sometimes results in
```
?filter=foo&query=bar&query=baz
```
or sometimes
```
?filter=foo,bar,baz
```
depending on the API. So I'm not sure what to support in oats, without breaking the generality of the generator.

I suppose the practical thing is just to support the former style, which seems to be what Influx uses. Or maybe we just write a custom API utility function—that's probably my preference. We can still use the generated request/response types from oats to ensure it stays in sync with the swagger.

ui/src/alerting/components/builder/CheckMatchingRulesCard.tsx
ebb-tide: I need to call a function that can call `setMatchingRules` because I can't use an async function within useEffect, and it feels a little awkward to pass the `setMatchingRules` to a function extracted to somewhere else.

ui/src/alerting/components/builder/CheckMatchingRulesCard.tsx
ebb-tide: agreed.

ui/src/alerting/components/builder/CheckMatchingRulesCard.tsx
ebb-tide: yup definitely doesn't work as is! let's take a look at this tomorrow?

ebb-tide: solved!

ui/src/alerting/components/builder/CheckMatchingRulesCard.tsx
chnn: oh I see, ok makes sense

ui/src/alerting/components/builder/CheckMatchingRulesCard.tsx
```diff
@@ -1,11 +1,70 @@
...
+  orgID: string
+}
+
+const CheckMatchingRulesCard: FunctionComponent<StateProps> = ({
+  check,
+  orgID,
+}) => {
+  const getMatchingRules = async (): Promise<NotificationRule[]> => {
+    const tags: CheckTagSet[] = get(check, 'tags', [])
+    const tagsList = tags.reduce((acc, t) => {
```
chnn: (real minor) might be easier to read without the `reduce`

```
    const tags: CheckTagSet[] = get(check, 'tags', []).map(t => ['tag', `${t.key}:${t.value}`])
```

ebb-tide: dammit I failed the chnn map test!

chnn: fired

PR #14884, opened 2019-08-30
fix(kv): mock system buckets

jademcgough:
Removes system bucket creation and hardcodes them into the response instead.

This is a short-term, stopgap solution for problems with #14539. Better solution to come in the following days.

### DIFF
--- a/inmem/bucket_service.go
+++ b/inmem/bucket_service.go
@@ -247,7 +247,7 @@ func (s *Service) findSystemBucket(n string) (*platform.Bucket, error) {
 	default:
 		return nil, &platform.Error{
 			Code: platform.ENotFound,
-			Msg:  fmt.Sprintf("bucket %q not found", n),
+			Msg:  fmt.Sprintf("system bucket %q not found", n),
 		}
 	}
 }
--- a/inmem/bucket_service.go
+++ b/inmem/bucket_service.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"time"
 
 	platform "github.com/influxdata/influxdb"
 )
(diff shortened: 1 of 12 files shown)


inmem/bucket_service.go
```diff
@@ -219,6 +228,30 @@ func (s *Service) FindBuckets(ctx context.Context, filter platform.BucketFilter,
...
+		return &platform.Bucket{
+			ID:              platform.MonitoringSystemBucketID,
+			Name:            "_monitoring",
+			RetentionPeriod: time.Hour * 24 * 7,
+			Description:     "System bucket for monitoring logs",
+		}, nil
+	default:
+		return nil, &platform.Error{
+			Code: platform.ENotFound,
+			Msg:  fmt.Sprintf("bucket %q not found", n),
```
kelwang: change it to system bucket %q not found, or add an Op "inmem/findSystemBucket"

kv/bucket.go
```diff
@@ -173,6 +141,30 @@ func (s *Service) FindBucketByName(ctx context.Context, orgID influxdb.ID, n str
...
+		return &influxdb.Bucket{
+			ID:              influxdb.MonitoringSystemBucketID,
+			Name:            "_monitoring",
+			RetentionPeriod: time.Hour * 24 * 7,
+			Description:     "System bucket for monitoring logs",
+		}, nil
+	default:
+		return nil, &influxdb.Error{
+			Code: influxdb.ENotFound,
+			Msg:  fmt.Sprintf("bucket %q not found", n),
```
kelwang: same, change it to system bucket %q not found, or add an Op "kv/findSystemBucket"

inmem/bucket_service.go
```diff
@@ -209,6 +210,14 @@ func (s *Service) findBuckets(ctx context.Context, filter platform.BucketFilter,
 // FindBuckets returns a list of buckets that match filter and the total count of matching buckets.
 // Additional options provide pagination & sorting.
 func (s *Service) FindBuckets(ctx context.Context, filter platform.BucketFilter, opt ...platform.FindOptions) ([]*platform.Bucket, int, error) {
+	if filter.Name != nil {
```
desa: I think we don't need this file. I think this is dead code like the bolt bucket service.

PR #14890, opened 2019-08-30
fix(ui): constrain query builder in check overlay

### LINKED ISSUE #14785: Guide user into selecting valid aggregate window / every duration in check overlay
When a user is building a query with the query builder in the check overlay:

- [x] Remove "none" and "auto" aggregate window duration options from the aggregate card dropdown:

   Screen Shot 2019-08-28 at 1 43 22 PM

- [x] Enforce that an aggregate window is always selected

- [x] Add options for aggregate window durations smaller than `5m` (`1s` being the smallest valid duration)

- [x] The aggregate window duration should always stay in sync with the "every" option of a check; when one changes, so must the other

- [x] Remove the time range dropdown

- [x] For the time range in the visualization, use some fixed multiple of the aggregate window duration. Displaying a time range 100x the aggregate window duration means that the graph will display 100 points; we should pick a reasonable constant based on this fact

- [x] Add validation to the "Schedule every" field in the "Configure a Check" screen, and do not emit changes to the field until they are valid:

   Screen Shot 2019-08-28 at 1 42 22 PM

Bonus polish:

- [x] Rename "Schedule every" to "Schedule Every" in the "Configure a Check" screen

- [x] Update the builder cards to use the single select style (e.g. what the bucket selection card uses):
  
   Screen Shot 2019-08-28 at 1 46 51 PM

chnn:
Closes #14785

Makes a bunch of small tweaks to the query builder when it is used in the check overlays. 

builder


ui/src/alerting/utils/vis.ts
```diff
@@ -0,0 +1,26 @@
...
+// Types
+import {TimeRange} from 'src/types'
+
+const POINTS_PER_CHECK_PLOT = 300
+
+/*
+  Given the duration in a check's `every` field, return a `TimeRange` suitable
+  for visualizing the input data to the check.
+
+  For example, a check might have an `every` value of `1m`. The the check is
```
chnn:
```suggestion
  For example, a check might have an `every` value of `1m`. Then the check is
```

ui/src/alerting/utils/vis.ts
```diff
@@ -0,0 +1,26 @@
...
+  run once every minute, and the input data to the `check` at each run is a
+  single value aggregated from the last `1m`. So to display a visualization of
+  the check data over time, we want to show a series of these aggregated values
```
chnn:
```suggestion
  the check data over time, we want to show a series of these values
```

ui/src/timeMachine/components/FunctionSelector.tsx
```diff
@@ -15,22 +15,44 @@ import {
...
   AGG_WINDOW_AUTO,
+  AGG_WINDOW_NONE,
 } from 'src/timeMachine/constants/queryBuilder'
 
 // Types
 import {AppState, BuilderConfig} from 'src/types'
 
 const FUNCTION_NAMES = FUNCTIONS.map(f => f.name)
 
+const DURATIONS = [
```
ebb-tide: Maybe `DURATIONS` should be equal to the `CHECK_EVERY_OPTIONS` const?

chnn: That seems better, I've updated it.

ui/src/alerting/utils/vis.ts
```diff
@@ -0,0 +1,26 @@
...
+  for visualizing the input data to the check.
+
+  For example, suppose a check has an `every` value of `1m`. Then the check is
+  run once per minute, and the input data to the `check` for each run is a
+  single value aggregated from the last `1m`. To display a visualization of the
+  check data over time, we want to show a series of the values aggregated for
+  each minute. So to display a plot with say, 300 points, we need to query a
+  time range of the last 300 minutes.
+*/
+export const getCheckVisTimeRange = (durationStr: string): TimeRange => {
```
ebb-tide: 👍

ui/src/timeMachine/components/Queries.tsx
```diff
@@ -59,50 +57,30 @@ type Props = StateProps & DispatchProps
...
-              {activeTimeMachineID !== 'alerting' && <CSVExportButton />}
-              <TimeMachineRefreshDropdown />
-              <TimeRangeDropdown
-                timeRange={timeRange}
-                onSetTimeRange={this.handleSetTimeRange}
-                centerPicker={true}
-              />
-              {activeTimeMachineID !== 'alerting' && (
-                <TimeMachineQueriesSwitcher />
+              {!isInCheckOverlay && (
```
ebb-tide: thanks that is better :)

chnn:
> I wonder if product will ask for a freeform input rather than a dropdown for the every

Yeah, I was wondering about this too. I ultimately went with the dropdowns since their selection is always valid, and since they can explain durations in human words (e.g. "Every 1 minute" rather than `1m`). I think we could do the same with a freeform text input with validation and some sort of help tooltip explaining duration literals.

PR #14892, opened 2019-08-30
perf(storage): improve prefix deletion in Cache

e-dard:
This PR adds several commits that significantly improve the performance of executing prefix-based deletes in the `tsm1.Cache`.

All deletes in 2.x are "prefix-based", because series data is segmented between buckets using an `<org, bucket>` id as the measurement name in the TSM engine and the TSI index.

We have recently been seeing issues where contention over the Cache (contention between a delete and some writes) is becoming problematic and affecting write throughput. This often manifests during a retention check (which is executed as a prefix-based delete for each bucket in the system, across varying time periods).

I tested the improvements in this PR with a benchmark that measures CPU time and allocations when deleting from a cache with 100K keys in it. There are two benchmarks: one where some data in the cache needs to be deleted, and one where nothing in the cache matches the key.

**Follow on Work**: there is some forthcoming follow-on work that's not included in this PR:

 - improving performance of deletion against a cache when there are no matching keys. This can be improved significantly with a Bloom filter or similar and I should have something next week.
 - not touching the cache if we know that there can't be any matching timestamps within it. This should basically mean retention checks generally don't need to touch the cache. All of the work in this PR is still relevant for more general deletions though.

Overall the performance improvement is about **4.5x** for deleting from the cache (`~77-83%` reduction in benchmark time).

Here are some benchstats comparing this PR to the baseline benchmark introduced in deb7a59.

```
name                                          old time/op    new time/op    delta
Engine_DeletePrefixRange_Cache/exists-24         268ms ± 5%      59ms ± 3%  -77.89%  (p=0.000 n=10+8)
Engine_DeletePrefixRange_Cache/not_exists-24     265ms ± 5%      45ms ± 1%  -83.00%  (p=0.000 n=10+10)

name                                          old alloc/op   new alloc/op   delta
Engine_DeletePrefixRange_Cache/exists-24        64.1MB ± 0%     3.1MB ± 2%  -95.15%  (p=0.000 n=9+10)
Engine_DeletePrefixRange_Cache/not_exists-24    59.2MB ± 0%     0.1MB ±86%  -99.86%  (p=0.000 n=8+10)

name                                          old allocs/op  new allocs/op  delta
Engine_DeletePrefixRange_Cache/exists-24          711k ± 0%       28k ± 1%  -96.05%  (p=0.000 n=9+10)
Engine_DeletePrefixRange_Cache/not_exists-24      700k ± 0%        1k ±63%  -99.89%  (p=0.000 n=8+10)
```

The performance improvements focus on two main areas: reducing lock contention and reducing small allocations.

### DIFF
--- a/tsdb/tsm1/engine_test.go
+++ b/tsdb/tsm1/engine_test.go
@@ -16,10 +16,12 @@ import (
 	"github.com/influxdata/influxdb"
 	"github.com/influxdata/influxdb/logger"
 	"github.com/influxdata/influxdb/models"
+	"github.com/influxdata/influxdb/toml"
 	"github.com/influxdata/influxdb/tsdb"
 	"github.com/influxdata/influxdb/tsdb/tsi1"
 	"github.com/influxdata/influxdb/tsdb/tsm1"
 	"github.com/influxdata/influxql"
+	"go.uber.org/zap"
 )
 
 // Test that series id set gets updated and returned appropriately.
@@ -132,7 +134,7 @@ func TestEngine_SnapshotsDisabled(t *testing.T) {
 func TestEngine_ShouldCompactCache(t *testing.T) {
 	nowTime := time.Now()
 
-	e, err := NewEngine()
+	e, err := NewEngine(tsm1.NewConfig())
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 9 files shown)


tsdb/tsm1/engine_test.go
```diff
@@ -307,6 +309,56 @@ func BenchmarkEngine_WritePoints_Parallel(b *testing.B) {
...
+			if err = e.WritePoints(pp); err != nil {
+				b.Fatal(err)
+			}
+			b.StartTimer()
+
+			if err := e.DeletePrefixRange(context.Background(), []byte("fooasdasdasdasdasd"), 0, math.MaxInt64, nil); err != nil {
+				b.Fatal(err)
+			}
+		}
+	})
```
benbjohnson: This is a pretty minor nit but I personally find it easier for the subtest name to be camel case (or at least use underscores instead of spaces) since `go test` will replace the spaces with underscores. e.g. `t.Run("NotExists")` or `t.Run("not_exists")`. That way it's easier to search for if you just have the test name.

PR #14901, opened 2019-09-02
perf(storage): expose ability to peek on stream readers

GeorgeMac:
Expose ability to Peek on `reads.ResultSetStreamReader` and `reads.GroupResultSetStreamReader`. The purpose being to allows creators of this type to pre-emptively peek at a frame before creating a cursor.

When creating multiple stream readers like these for multiple partitions and joining them into a single sequential or merged result set, a peek on each partition could be issued concurrently by the caller. This would allow us to quickly discard empty results, potentially reducing latency on reads.

- [x] Rebased/mergeable
- [x] Tests pass


stuartcarnie:
The next step would be to merge this and update the IDPE `Store` type to parallelize the calls to `Peek`. `Store` is used by queryd to make the requests to the storage nodes.

GeorgeMac:
@stuartcarnie my thoughts exactly 💯

PR #14908, opened 2019-09-03
build(webpack): add source maps for @influxdata/giraffe prod builds

goller:
We were getting runtime errors in the prod builds via honeybadger.
They were unreadable because we did not have source maps for giraffe.

We were only able to get this working for prod builds but not dev
despite trying putting it in the webpack common file.  This is good
'nuff for our purpose now, but maybe we'll need further investigation.

Bonus commit... I added a new make target of influxd so I would not
need to type so much.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/webpack.prod.ts
+++ b/ui/webpack.prod.ts
@@ -1,10 +1,15 @@
 export {}
-const merge = require('webpack-merge')
+
+// utils
 const common = require('./webpack.common.ts')
-const OptimizeCSSAssetsPlugin = require('optimize-css-assets-webpack-plugin')
-const TerserJSPlugin = require('terser-webpack-plugin')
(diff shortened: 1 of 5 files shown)


ui/webpack.prod.ts
```diff
@@ -28,6 +33,11 @@ module.exports = merge(common, {
           },
         ],
       },
+      {
+        test: /\.js$/,
+        use: ['source-map-loader'],
+        include: [path.resolve(__dirname, 'node_modules/@influxdata/giraffe')],
```
chnn: Maybe add clockface too

goller: good point... on it

Issue #14876, opened 2019-08-29
Removing organization owner is returning error


ischolten:
__Steps to reproduce:__
Create an organization, the user creating the org should be the default owner.
Add another owner to the organization.
Delete the original owner. 
This should return with an error `user to resource mapping not found`.

__Expected behavior:__
The original owner should be removed successfully. and the api response should be a 204.

__Actual behavior:__
An error response returns.

PR #14918, opened 2019-09-03
fix(kv): ignore not found error when removing org dep resources

kelwang:
Closes https://github.com/influxdata/influxdb/issues/14876

Issue only happens in cloud,
if an org dependent resource doesn't exist, such an error shouldn't be a stopper to prevent the whole remove process. 

Manually tested

- [ ] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/kv/urm.go
+++ b/kv/urm.go
@@ -8,6 +8,7 @@ import (
 	"github.com/influxdata/influxdb"
 	icontext "github.com/influxdata/influxdb/context"
 	"github.com/influxdata/influxdb/kit/tracing"
+	"go.uber.org/zap"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


jademcgough:
> Issue only happens in cloud

Can you expand on this? What changed to make it so that org dep dependencies aren't being created?

kelwang:
> > Issue only happens in cloud
> 
> Can you expand on this? What changed to make it so that org dep dependencies aren't being created?

the cloud onboarding process didn't add the the org dependent bucket resource mapping. So when deleting the urm, the bucket urm couldn't be found.

We will have another pr to figure out the onboarding issue for the bucket urm. But for existed urms in cloud, we shouldn't block them from deleting.

PR #14926, opened 2019-09-04
feat(tasks): report error messages to metrics

AlirieGray:
Closes https://github.com/influxdata/idpe/issues/4262

This PR makes a change to the new executor, which will be put into effect when the new scheduler is merged.

Previously, executor metrics only reported the number of errors thrown by the executor, without any information on what type of errors were occurring. This PR attaches an error type to the error metric, so that developers can observe error trends in the tasks dashboard.

- [x] Rebased/mergeable
- [x] Tests pass


AlirieGray:
TODO: make an issue for categorizing errors based on metrics in the dashboard

AlirieGray:
One alternative to adding another parameter to the `finish` function would be to use the standard InfluxDB error codes. However, most of the errors in the executor are of the standard Go error type, not InfluxDB errors.

PR #14927, opened 2019-09-04
fix(query/stdlib): use auth correctly in the v1.databases() call

jsternberg:
The `v1.databases()` call did not correctly filter buckets based on
auth. Fortunately, it did not cause any improper permissions such as
allowing a person to see buckets that they had no read access to.

The error instead was that if a user did not have read access to one of
the buckets that was returned, the entire command would fail rather than
filter out the bucket that didn't have permissions.

This changes it so that if the user doesn't gets an unauthorized error
when accessing a bucket, it will filter it from the list instead of
failing. It also changes it so the error message is marked as
`ENotFound` instead of as an internal error.

### DIFF
--- a/query/stdlib/influxdata/influxdb/v1/databases.go
+++ b/query/stdlib/influxdata/influxdb/v1/databases.go
@@ -3,6 +3,7 @@ package v1
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/influxdata/flux"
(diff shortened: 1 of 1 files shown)


jsternberg:
Yea. Unfortunately the pipeline test is going to be in idpe because databases isn't actually in OSS when I tried to test it.

PR #14929, opened 2019-09-04
fix(tsi1): Remove TSI cardinality stats cache

benbjohnson:
Removes the original implementation of TSI cardinality stats which would cache all files except for the active log file and then merge the stats together on demand. This caused deleted series in the active log file to be ignored.

This PR changes the computation so all file stats are computed together. It also includes a new time based cache (`stats-ttl`) that defaults to `0` (disabled).

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/tsdb/tsi1/partition.go
+++ b/tsdb/tsi1/partition.go
@@ -58,16 +58,17 @@ type Partition struct {
 	fileSet       *FileSet // current file set
 	seq           int      // file id sequence
 
-	// Computed measurements stats since last compaction.
-	// NOTE: Does not include active log file stats.
-	stats MeasurementCardinalityStats
-
 	// Running statistics
 	tracker *partitionTracker
 
 	// Fast series lookup of series IDs in the series file that have been present
 	// in this partition. This set tracks both insertions and deletions of a series.
 	seriesIDSet *tsdb.SeriesIDSet
+
+	// Stats caching
+	StatsTTL      time.Duration
+	statsCache    MeasurementCardinalityStats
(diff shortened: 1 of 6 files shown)


tsdb/tsi1/partition.go
```diff
@@ -1294,40 +1291,75 @@ func (p *Partition) compactLogFile(ctx context.Context, logFile *LogFile, interr
...
+			return stats, nil
+		}
+
+		// Obtain all series for measurement.
+		sitr := fs.MeasurementSeriesIDIterator(mm.Name())
+		if sitr == nil {
+			continue
+		}
+
+		// All iterators should be series id set iterators except legacy 1.x data.
```
jacobmarble: Will this be a concern when users eventually upgrade from 1.x to 2.x? Want to make sure silently skipping is appropriate.

tsdb/tsi1/partition.go
```diff
@@ -667,6 +662,8 @@ func (p *Partition) DropMeasurement(name []byte) error {
 			}
 			all = append(all, elem.SeriesID)
 
+			n++
```
e-dard: where is `n` getting used?

tsdb/tsi1/partition.go
benbjohnson: This has been in there for a while. IIRC, we decided that users would either need to wait for compactions to update their files automatically or do a manual rebuild if they want to see org/bucket cardinality metrics. Although I'm not sure we even expose those in OSS right now.

PR #14976, opened 2019-09-05
fix(ui): fix max call stack exceeded error when parsing CSV

### LINKED ISSUE #14566: Meta query crashes UI
A call to `tagValues` caused the UI to return the following error message:

ui_crash

Steps to reproduce:

1. Create the following toml file:

    ```toml
    title = "Documented schema"
    [[measurements]]
    name = "mem"
    tags = [
        { name = "AYx2mIOn58", source = { type = "sequence", format = "AYx2mIOn58-%s", start = 0, count = 5000000 } },
        { name = "next", source = { type = "sequence", format = "next-%s", start = 0, count = 2 } },
    ]
    fields = [
        # An example of a sequence of integer values
        { name = "free", count = 10, source = 25, time-interval = "1h" },
        { name = "consumed", count = 1000, source = 25.5, time-precision = "m" },
    ]
    ```
2. Generate the series defined in the above file(influxd should not be running):
```
influxd generate <schema.toml> --org <org> --bucket <bucket> --clean all
```
3. Run `influxd`.
4. Pull up the UI `http://localhost:9999`.
5. Navigate to the data explorer and issue the following flux query:
    ```
    
    v1.tagValues(bucket:"b", tag:"AYx2mIOn58")
    ```
    This query should return `3748952` rows (~70MB of uncompressed data).

Note I am using firefox 68.0.1

chnn:
When parsing a Flux CSV file with hundreds of thousands of lines for the "Raw Data View", we would see a

    Maximum call stack size exceeded

error. This was because every line in the CSV was being passed as an argument in a single `Array.prototype.push` call, and there are [limits][0] to how many arguments can be passed to a function.

This commit avoids passing the arguments all at once.

[0]: https://stackoverflow.com/questions/22747068/is-there-a-max-number-of-arguments-javascript-functions-can-accept

Closes #14566

### DIFF
--- a/ui/src/timeMachine/utils/rawFluxDataTable.ts
+++ b/ui/src/timeMachine/utils/rawFluxDataTable.ts
@@ -19,7 +19,13 @@ export const parseFiles = (responses: string[]): ParseFilesResult => {
       data.push([])
     }
 
-    data.push(...parsedChunks[i])
+    for (let j = 0; j < parsedChunks[i].length; j++) {
+      // Danger zone! Since the contents of each chunk are potentially quite
+      // large, the contents need to be concated using a loop rather than with
+      // `concat`, a splat or similar. Otherwise we see a "Maximum call size
+      // exceeded" error for large CSVs
+      data.push(parsedChunks[i][j])
+    }
 
     // Add an empty line at the end
     data.push([])


ui/src/timeMachine/utils/rawFluxDataTable.ts
```diff
@@ -19,7 +19,9 @@ export const parseFiles = (responses: string[]): ParseFilesResult => {
       data.push([])
     }
 
-    data.push(...parsedChunks[i])
+    for (let j = 0; j < parsedChunks[i].length; j++) {
```
jlapacik: A comment would be useful here just to guard against any regressions in the future.

PR #15008, opened 2019-09-06
fix(task): Update task in scheduler to show updated logs.

lyondhill:
The current behavior is that the update is pushed into the scheduler,
and the scheduler cherry pick's what it needs. This leaves the task itself out
meaning any logging the scheduler did was not going to have the new task information in it.

### DIFF
--- a/task/backend/scheduler.go
+++ b/task/backend/scheduler.go
@@ -359,6 +359,12 @@ func (s *TickScheduler) UpdateTask(authCtx context.Context, task *platform.Task)
 	ts.nextDue = next
 	ts.authCtx = authCtx
 	ts.nextDueMu.Unlock()
+
+	// update the current runners cached task for logs
+	for _, runner := range ts.runners {
+		runner.task = task
+	}
+
 	// check the concurrency
 	// todo(lh): In the near future we may not be using the scheduler to manage concurrency.
 	maxC := len(ts.runners)


task/backend/scheduler.go
```diff
@@ -359,6 +359,12 @@ func (s *TickScheduler) UpdateTask(authCtx context.Context, task *platform.Task)
 	ts.nextDue = next
 	ts.authCtx = authCtx
 	ts.nextDueMu.Unlock()
+
+	// update the current runners cached task for logs
+	for _, runner := range ts.runners {
+		runner.task = task
```
GeorgeMac: Any potential for a race here? I see we are protected when doing this write by the `schedulerMu`, but I notice the runner itself has its own mutexes for state and it has its hands on the taskSchedulers mutexes as well. Also I notice that the runners also spawn goroutines on themselves.
I guess if the race detector is chill this might not be a problem.

lyondhill: Just above we acquire a scheduler wide lock. I will investigate a bit more just incase

lyondhill: I looked into it and haven't found any race that could effect us, negatively. If a task is running with the old script and we are logging the run at the same time we may get one last log with the old flux script in it. But this isn't a failure. To fix that we would need to maintain a database transaction through the scheduler.

GeorgeMac: Nice thanks @lyondhill !

PR #15027, opened 2019-09-06
feat: limit monitoring and alerting

### LINKED ISSUE #14805: Check resourcelimits for checks/rules/endpoints and display appropriate errors to user
Free tier cloud users will have limits on the # of checks/rules/endpoints they can create. We should wrap `<AlertingIndex>` in a` <AssetLimitAlert> `

Consider joining this work with work on https://github.com/influxdata/influxdb/issues/14893

### LINKED ISSUE #14893: Check limits EP before displaying the endpoint types that are available to user
Free tier users do not have access to pagerduty endpoint. 
The list of endpoints that are blocked for a user are available to the UI through the /limits endpoints: as defined here:
https://github.com/influxdata/idpe/pull/4504

UI should query limits, and should not display blocked endpoints as an option in the endpoints dropdown

Consider implementing https://github.com/influxdata/influxdb/issues/14805 in the same PR

ebb-tide:
Closes #14893
Closes #14805


ui/src/alerting/actions/checks.ts
```diff
@@ -168,7 +173,7 @@ export const updateCheck = (check: Partial<Check>) => async (
 }
 
 export const deleteCheck = (checkID: string) => async (
-  dispatch: Dispatch<Action | NotificationAction>
+  dispatch: Dispatch<any>
```
chnn: Do we get anything by typing `dispatch`? Feels like it's more trouble than it's worth.

ui/src/alerting/actions/checks.ts
```diff
@@ -85,6 +88,7 @@ export const getChecks = () => async (
     }
 
     dispatch(setAllChecks(RemoteDataState.Done, resp.data.checks))
+    dispatch(checkChecksLimits())
```
chnn: I wish that the API would return rate limiting information along with resource responses... feels really off that we are doing all this on the client.

ui/src/types/cloud.ts
```diff
@@ -0,0 +1,40 @@
+export interface Limits {
```
chnn: Thanks for moving this.

ui/src/alerting/actions/checks.ts
121watts: We ensure that the actions are of the correct types.

121watts: It's been awesome for me.  It ensures that the actions we pass dispatc conform to the Action generic.

PR #15029, opened 2019-09-06
fix(alertHistory): stop linking to dead resources

### LINKED ISSUE #14930: Clicking a check that has been deleted should be handled gracefully
For now, let's just show a notification in the UI that the Check has been deleted.

image

121watts:
Closes #14930

### What
From the AlertingIndex we stopped linking to resources that don't exist.

Kapture 2019-09-06 at 14 02 46

Co-Author: Michael Desa <mjdesa@gmail.com>


ui/src/alerting/reducers/checks.ts
```diff
@@ -66,3 +66,7 @@ export default (
         return
     }
   })
+
+export const getCheckIDs = (state: ChecksState): {[x: string]: boolean} => {
```
chnn: This would be more similar to most of the other selectors in the app if it lived in a `selectors` directory and accepted an `AppState` rather than `ChecksState` as an argument.

ui/src/alerting/utils/history.ts
```diff
@@ -110,7 +110,7 @@ from(bucket: "${MONITORING_BUCKET}")
   - "notification_rule"
   - "_notification_rule"
 
-  The rewrite rules for the "noficationEndpoint" work similarly.
+  The rewrite rules for the "notificationEndpoint" work similarly.
```
chnn: 😛

ui/src/alerting/components/AlertHistoryIndex.tsx
```diff
@@ -44,39 +45,45 @@ const AlertHistoryIndex: FC<Props> = ({params: {orgID}}) => {
...
-            <div className="alert-history">
-              <EventTable {...props} fields={fields} />
-            </div>
-          </Page.Contents>
-        </Page>
-      )}
-    </EventViewer>
+    <GetResources resource={ResourceTypes.Checks}>
+      <GetResources resource={ResourceTypes.NotificationEndpoints}>
+        <GetResources resource={ResourceTypes.NotificationRules}>
```
chnn: Sooner or later I think we will really want to be able to pass multiple resource types to a single `GetResources` component. It will load faster (parallel rather than serial) and make the spinners look way less janky.

ui/src/alerting/reducers/checks.ts
121watts: Ah, yes they should take all of AppState

ui/src/alerting/components/AlertHistoryIndex.tsx
121watts: Agreed.

PR #15033, opened 2019-09-06
fix(query): Ensure tags are sorted before passing to PointsWriter

### LINKED ISSUE #14963: Flux influxdb.to implementation writes unsorted tag keys to storage via PointsWriter
Writes directly to a PointsWriter require the models.Tags are sorted in lexicographically ascending order. The existing implementation assumes `flux.ColReader` returns the slice of columns in order, which is not guaranteed of this API. Therefore, series keys may be written incorrectly to storage.

stuartcarnie:
This PR resolves an issue with the `to` function that was not validating the tag key, value pairs are sorted in lexicographically ascending order.

Additionally, this PR includes improvements to the `models` package to provide a set of APIs for creating `models.Tags` from a list of key, value pairs, ensuring the result satisfies two invariants:

* `models.Tags` are sorted in lexicographically ascending order
* does not contain any duplicate tag keys 

These APIs allow for a preallocated slice to be passed giving the caller the option to avoid or reduce  allocations.

```
BenchmarkNewTagsKeyValues/sorted/no_dupes/preallocate-8         	20000000	        63.0 ns/op	       0 B/op	       0 allocs/op
BenchmarkNewTagsKeyValues/sorted/no_dupes/allocate-8            	10000000	       124 ns/op	     144 B/op	       1 allocs/op
BenchmarkNewTagsKeyValues/sorted/dupes-8                        	10000000	       181 ns/op	     240 B/op	       1 allocs/op
BenchmarkNewTagsKeyValues/unsorted/no_dupes-8                   	10000000	       204 ns/op	     176 B/op	       2 allocs/op
BenchmarkNewTagsKeyValues/unsorted/dupes-8                      	 5000000	       308 ns/op	     272 B/op	       2 allocs/op
```

Closes #14963


models/points.go
```diff
@@ -55,6 +55,10 @@ var (
 
 	// ErrInvalidPoint is returned when a point cannot be parsed correctly.
 	ErrInvalidPoint = errors.New("point is invalid")
+
+	// ErrInvalidKevValuePairs is returned when the number of key, value pairs
+	// is odd, indicating a missing value.
+	ErrInvalidKevValuePairs = errors.New("key, value pairs is an odd length")
```
benbjohnson: This is a weird nit but can this be `key/value` instead of `key, value`? I feel like it could read weird depending on the context the error is displayed within.

models/points.go
```diff
@@ -2100,6 +2161,34 @@ func (a Tags) Clone() Tags {
...
+}
+
+// sorted returns true if a is sorted and is an optimization
+// to avoid an allocation when calling sort.IsSorted, improving
+// performance as much as 50%.
+func (a Tags) sorted() bool {
+	for i := len(a) - 1; i > 0; i-- {
+		if string(a[i].Key) < string(a[i-1].Key) {
+			return false
+		}
```
benbjohnson: Does `string(a[i].Key)` do an allocation? Would `bytes.Compare()` be faster?

models/points.go
stuartcarnie: I will change the error text 👍 

As it happens, I was back and forth on this too. I chose this format because [`strings.NewReplacer`](https://golang.org/pkg/strings/#NewReplacer) is documented with the words _old, new string pairs_.

models/points.go
stuartcarnie: It won't allocate, however, I will change the code to be clearer per your suggestion.

Incidentally, I reported that `bytes.Equal` was also slower than `string(a) == string(b)` in the Gopher `#performance` slack and someone logged an issue. Go 1.13 was updated to use the cast for `bytes.Equal`:

https://github.com/golang/go/blob/master/src/bytes/bytes.go#L20

models/points.go
stuartcarnie: Done

models/points.go
stuartcarnie: Done

PR #15035, opened 2019-09-06
build(flux): update to Flux v0.45.0

nathanielc:


nathanielc:
@wolffcm  

> This changed a bunch of dependencies, is that expected?

Yes, the google bigtable client was brought in so it updated lots of deps on the Flux side too.

wolffcm:
We've already brought influxdb up to 0.45.1 now.

PR #15049, opened 2019-09-07
feat(kv): random ids without comma, space, backslash for org a…

goller:
At times snowflake id generation would create org and bucket IDs with
characters that had special meaning for the storage engine.

The storage engine concats the org and bucket bytes together into a
single 128 bit value.  That value is used in the old measurement
section.  Measurement was transformed into the tag, _measurement.

However, certain properties of the older measurement data location
are still required for the org/bucket bytes.  We cannot have
commas, spaces, nor backslashes.

This PR puts a specific ID generator in place during the creation of
orgs and buckets.  The IDs are just random numbers but with each
of the restricted chars incremented by one.  While this changes the
entropy distribution somewhat, it does not matter too much for our
purposes.

... because now org and bucket ids are checked for previous existence
transactionally in the key-value stores.  If the ID does already exist
then we try to generate a new key up to 100 times.

### DIFF
--- a/kv/service.go
+++ b/kv/service.go
@@ -24,20 +24,25 @@ type Service struct {
 	Logger *zap.Logger
 	Config ServiceConfig
 
-	IDGenerator    influxdb.IDGenerator
+	IDGenerator influxdb.IDGenerator
+
+	// special ID generator that never returns bytes with backslash,
+	// comma, or space. Used to support very specific encoding of org &
+	// bucket into the old measurement in storage.
+	OrgBucketIDs influxdb.IDGenerator
+
 	TokenGenerator influxdb.TokenGenerator
 	influxdb.TimeGenerator
 	Hash Crypt
-
-	// Organization & bucket specific validation.
-	IsValidOrgBucketID func(influxdb.ID) bool
 }
 
 // NewService returns an instance of a Service.
 func NewService(kv Store, configs ...ServiceConfig) *Service {
 	s := &Service{
-		Logger:         zap.NewNop(),
-		IDGenerator:    snowflake.NewIDGenerator(),
+		Logger:      zap.NewNop(),
+		IDGenerator: snowflake.NewIDGenerator(),
+		// Seed the random number generator with the current time
+		OrgBucketIDs:   rand.NewOrgBucketID(time.Now().UnixNano()),
 		TokenGenerator: rand.NewTokenGenerator(64),
 		Hash:           &Bcrypt{},
 		kv:             kv,
@@ -149,4 +154,12 @@ func (s *Service) Initialize(ctx context.Context) error {
 // Should only be used in tests for mocking.
 func (s *Service) WithStore(store Store) {
 	s.kv = store
+}
+
+// WithSpecialOrgBucketIDs sets the generator for the org
+// and bucket ids.
+//
+// Should only be used in tests for mocking.
+func (s *Service) WithSpecialOrgBucketIDs(gen influxdb.IDGenerator) {
+	s.OrgBucketIDs = gen
 }
--- a/kv/unique.go
+++ b/kv/unique.go
@@ -21,14 +21,14 @@ func UnexpectedIndexError(err error) *influxdb.Error {
 // exists.
 var NotUniqueError = &influxdb.Error{
 	Code: influxdb.EConflict,
-	Msg:  fmt.Sprintf("name already exists"),
+	Msg:  "name already exists",
 }
 
 // NotUniqueIDError is used when attempting to create an org or bucket that already
 // exists.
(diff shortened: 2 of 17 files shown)


kv/service.go
```diff
@@ -24,20 +24,25 @@ type Service struct {
...
 
 // NewService returns an instance of a Service.
 func NewService(kv Store, configs ...ServiceConfig) *Service {
 	s := &Service{
-		Logger:         zap.NewNop(),
-		IDGenerator:    snowflake.NewIDGenerator(),
+		Logger:      zap.NewNop(),
+		IDGenerator: snowflake.NewIDGenerator(),
+		// Seed the random number generator with the current time
+		OrgBucketIDs:   rand.NewOrgBucketID(time.Now().UnixNano()),
```
e-dard: Perhaps you want to provide an option to set the seed? It will make testing easier because the ids generated will be deterministically generated.

kv/unique.go
```diff
@@ -23,6 +24,13 @@ var NotUniqueError = &influxdb.Error{
 	Msg:  fmt.Sprintf("name already exists"),
 }
 
+// NotUniqueIDError is used when attempting to create an org or bucket that already
+// exists.
+var NotUniqueIDError = &influxdb.Error{
+	Code: influxdb.EConflict,
+	Msg:  fmt.Sprintf("ID already exists"),
```
e-dard: nit: don't need `Sprintf` here

rand/id.go
```diff
@@ -0,0 +1,68 @@
...
+// Typically, seed with `time.Now().UnixNano()`
+func NewOrgBucketID(seed int64) *OrgBucketID {
+	return &OrgBucketID{
+		src: rand.New(rand.NewSource(seed)),
+	}
+}
+
+// Seed allows one to override the current seed.
+// Typically, this override is done for tests.
+func (r *OrgBucketID) Seed(seed int64) {
```
e-dard: Ah OK, ignore my previous comment on the main service. This works.

rand/id.go
```diff
@@ -0,0 +1,68 @@
...
+	r.m.Lock()
+	r.src = rand.New(rand.NewSource(seed))
+	r.m.Unlock()
+}
+
+// ID generates an ID that does not have backslashes, commas, or spaces.
+func (r *OrgBucketID) ID() influxdb.ID {
+	r.m.Lock()
```
e-dard: If this ever becomes too slow, we can pre-buffer n random ids that only need read access to distribute.

rand/id.go
```diff
@@ -0,0 +1,68 @@
...
+	n = sanitize(n)
+	return influxdb.ID(n)
+}
+
+func sanitize(n uint64) uint64 {
+	b := make([]byte, 8)
+	binary.BigEndian.PutUint64(b, n)
+	for i := range b {
+		switch b[i] {
+		case 0x5C, 0x2C, 0x20:
```
e-dard: we could maybe add an explicit comment here about these bytes, and maybe even say to go look in the `models` package?

testing/bucket_service.go
```diff
@@ -9,7 +9,7 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
-	platform "github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb"
```
e-dard: 🙏

testing/bucket_service.go
```diff
@@ -157,20 +159,17 @@ func CreateBucket(
 		{
 			name: "basic create bucket",
 			fields: BucketFields{
-				IDGenerator: &mock.IDGenerator{
-					IDFn: func() platform.ID {
-						return MustIDBase16(bucketTwoID)
-					},
-				},
+				IDGenerator:   mock.NewIDGenerator(bucketTwoID, t),
+				OrgBucketIDs:  mock.NewIDGenerator(bucketTwoID, t),
```
e-dard: does it matter that these are the same kind of generators?

testing/organization_service.go
```diff
@@ -3,13 +3,15 @@ package testing
...
 	"bytes"
 	"context"
+	"fmt"
 	"sort"
 	"testing"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
-	platform "github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb"
```
e-dard: 🙏

testing/bucket_service.go
goller: No, this one doesn't matter... The `IDGenerator` is used for all IDs except org/bucket.  However, I didn't remove IDGenerator here because kv would panic later if there were other changes that perhaps included creating other resources while creating buckets.

PR #15068, opened 2019-09-09
fix(alerting): disable save button if query is invalid

### LINKED ISSUE #14894: Disable saving on checks when the query for a check is invalid

121watts:
Closes #14894 
Kapture 2019-09-09 at 12 16 24

### Note
I will update the e2e tests to only run once for each test.

### DIFF
--- a/ui/cypress/support/commands.ts
+++ b/ui/cypress/support/commands.ts
@@ -305,25 +305,14 @@ export const flush = () => {
 }
 
 export const writeData = (
-  lines: string[],
-  chunkSize: number = 100
+  lines: string[]
 ): Cypress.Chainable<Cypress.Response> => {
   return cy.fixture('user').then(({org, bucket}) => {
-    let chunk: string[]
-    let chunkCt: number = 0
-    while (chunkCt < lines.length) {
-      chunk =
-        chunkCt + chunkSize <= lines.length
-          ? lines.slice(chunkCt, chunkCt + chunkSize - 1)
-          : lines.slice(chunkCt, chunkCt + (chunkSize % lines.length))
-      cy.request({
-        method: 'POST',
-        url: '/api/v2/write?org=' + org + '&bucket=' + bucket,
-        body: chunk.join('\n'),
-      })
-      chunkCt += chunkSize
-      chunk = []
-    }
+    cy.request({
+      method: 'POST',
+      url: '/api/v2/write?org=' + org + '&bucket=' + bucket,
+      body: lines.join('\n'),
+    })
   })
 }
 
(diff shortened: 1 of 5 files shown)


ui/cypress/support/commands.ts
```diff
@@ -305,25 +305,14 @@ export const flush = () => {
...
 
 export const writeData = (
-  lines: string[],
-  chunkSize: number = 100
+  lines: string[]
 ): Cypress.Chainable<Cypress.Response> => {
   return cy.fixture('user').then(({org, bucket}) => {
-    let chunk: string[]
-    let chunkCt: number = 0
-    while (chunkCt < lines.length) {
```
chnn: what in tarnation...

ui/src/alerting/components/CheckEOHeader.tsx
```diff
@@ -47,10 +79,17 @@ const CheckEOHeader: FC<Props> = ({name, onSetName, onCancel, onSave}) => {
     }
   }
 
-  const saveButtonStatus =
-    saveStatus === RemoteDataState.Loading
-      ? ComponentStatus.Loading
-      : ComponentStatus.Default
+  const saveButtonStatus = () => {
```
chnn: how come this is a function?

ui/src/timeMachine/utils/queryBuilder.ts
```diff
@@ -46,6 +46,18 @@ export const isDraftQueryAlertable = (
   }
 }
 
+export const isCheckSaveable = (
```
alexpaxton: =D

ui/src/alerting/components/CheckEOHeader.tsx
```diff
@@ -64,7 +103,11 @@ const CheckEOHeader: FC<Props> = ({name, onSetName, onCancel, onSave}) => {
         />
       </Page.Header.Left>
       <Page.Header.Center widthPixels={300}>
-        <CheckAlertingButton />
+        <CheckAlertingButton
```
chnn: What was your motivation for disconnecting it from Redux? We've moved this button around a bunch and it was nice to not worry about reconnecting the right props every time. I bet we'll end up moving it again since I think it makes more sense to be closer to the query/alert builder, which is the part of the UI that changes when you click it.

ui/src/alerting/components/CheckEOHeader.tsx
121watts: I think it reads much nicer as a function.

ui/src/alerting/components/CheckEOHeader.tsx
121watts: Two sibling both required `draftQueries` to make independent display decisions.  Rather than connect two  children to redux, I decided to pull the state up to their shared parent and pass props down one level.  Mayhaps I'll regret this decision.

PR #15072, opened 2019-09-09
chore(alerts): alphabetize endpoints in rule dropdown

### LINKED ISSUE #14430: Sort and filter check/rule/endpoint cards
- [ ] Add filtering to checks/rules/endpoints
- [ ] Add alphabetization to checks/rules/endpoints

### LINKED ISSUE #14999: Sort notification endpoints in notification rule dropdown
sort these alphabetically

image

121watts:
Closes #14999 
Closes #14430

### DIFF
--- a/ui/src/alerting/selectors/index.ts
+++ b/ui/src/alerting/selectors/index.ts
@@ -1,4 +1,5 @@
 import {AppState, Check, ResourceType} from 'src/types'
+import {sortBy, get} from 'lodash'
 
 export const getCheck = (state: AppState, id: string): Check => {
   const checksList = state.checks.list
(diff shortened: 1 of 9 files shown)


ui/src/alerting/selectors/index.ts
```diff
@@ -14,3 +15,17 @@ export const getResourceIDs = (
     {}
   )
 }
+
+interface HasName {
+  name?: string
+}
+
+export const getResourceList = <T extends HasName>(
```
chnn: My preference would be to repeat this code for each resource type so that it can be typechecked.

ui/src/alerting/selectors/index.ts
```diff
@@ -14,3 +15,17 @@ export const getResourceIDs = (
...
+  state: AppState,
+  resource: ResourceType
+): T[] => {
+  const resourceList: T[] = get(state, `${resource}.list`, [])
+  return sortBy(resourceList, l => {
```
chnn: Why `_.sortBy` over `Array.prototype.sort`?

121watts: nice and clean.

ui/src/alerting/selectors/index.ts
121watts: With the above we get yelled at if the generic does not have a `name` field.

Issue #15042, opened 2019-09-06
Add pagerduty client url to notification rule


russorat:
add a field for the user to set the `Pagerduty Client URL` to the notification rule. Default the value to the current $host.

PR #15079, opened 2019-09-09
fix(pagerduty/rule): change url to clientURL for pagerduty endpoint/rule

desa:
Closes https://github.com/influxdata/influxdb/issues/15042

Previously, there was an optional URL provided for the pagerduty
endpoint. However, the pagerduty API url does not change and as a result
it should not have been a parameter. The Pagerduty API does require a
`clientURL` that is presented in the pagerduty UI when an alert is
triggered. Currently that value will default to the alerts history page
for the organization.


http/swagger.yml
```diff
@@ -9760,9 +9760,9 @@ components:
       allOf:
         - $ref: "#/components/schemas/NotificationEndpointBase"
         - type: object
-          required: [url, routingKey]
+          required: [clientURL, routingKey]
           properties:
-            url:
+            clientURL:
```
russorat: no, this is the endpoint. it should not be clientURL

PR #15082, opened 2019-09-09
build(flux): update to Flux v0.45.1

wolffcm:


chronograf/oauth2/google.go
```diff
@@ -90,7 +92,7 @@ func (g *Google) PrincipalID(provider *http.Client) (string, error) {
 
 // Group returns the string of domain a user belongs to in Google
 func (g *Google) Group(provider *http.Client) (string, error) {
-	srv, err := goauth2.New(provider)
+	srv, err := goauth2.NewService(context.TODO(), option.WithHTTPClient(provider))
```
affo: Why are these changes here? 🤔

affo: Did you have to fix something for the updates in `oauth2`?

wolffcm: staticcheck complains about `New` being deprecated in favor of `NewService`, and I had to make this change to make it pass.  (the bigtable changes upgraded the oauth dependency, which is three years old I guess)

PR #15093, opened 2019-09-10
fix(alerts): validate notification rule

### LINKED ISSUE #15092: InfluxDB 2.0: NotificationRule should be validate before its create
The `NotificationRule` should be validate before its create same as is validate in `update`:

https://github.com/influxdata/influxdb/blob/9d67e4bfbfff0c78c41261a138b76391f58ca7d6/kv/notification_rule.go#L248

__Steps to reproduce:__
Create `NotificationRule` without `name`:
```bash
-> POST 
-> http://127.0.0.1:9999/api/v2/notificationRules
-> {"type":"slack","messageTemplate":"template","endpointID":"0472b4bc48843000","orgID":"04591716e3c0f000","every":"1h","tagRules":[],"statusRules":[],"labels":[]}

```
__Expected behavior:__
The `NotificationRule` should be validated before its create.

__Actual behavior:__
The `NotificationRule` isn't validate when is created.

bednar:
Closes #15092

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/kv/notification_rule.go
+++ b/kv/notification_rule.go
@@ -79,6 +79,9 @@ func (s *Service) createNotificationRule(ctx context.Context, tx Tx, nr influxdb
 	nr.SetOwnerID(userID)
 	nr.SetCreatedAt(now)
 	nr.SetUpdatedAt(now)
+	if err := nr.Valid(); err != nil {
+		return err
+	}
 
 	t, err := s.createNotificationTask(ctx, tx, nr)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


kv/notification_rule.go
```diff
@@ -79,6 +79,9 @@ func (s *Service) createNotificationRule(ctx context.Context, tx Tx, nr influxdb
 	nr.SetOwnerID(userID)
 	nr.SetCreatedAt(now)
 	nr.SetUpdatedAt(now)
+	if err := nr.Valid(); err != nil {
+		return err
+	}
```
kelwang: We already did the validation at line `if err := s.putNotificationRule(ctx, tx, nr)`, i don't think this is needed

PR #15111, opened 2019-09-11
docs(http): update NotificationEndpoint swagger definition to reflect reality

bednar:
1. The `NotificationEndpoint.name` should be required - see sources:
https://github.com/influxdata/influxdb/blob/b26ed76d6ae42eaed66e755c9e3c770fbf80ab90/notification/endpoint/endpoint.go#L63
2. The `NotificationEndpointBase` also contains `links` property - see sources:
https://github.com/influxdata/influxdb/blob/7a677e9532835961febc83734837c31d38b07769/http/notification_endpoint.go#L136
3. The `SlackNotificationEndpoint` properties should be optional  - because than the client is assume that has to to fill both property - see sources:
https://github.com/influxdata/influxdb/blob/79a6b353e3dacd9e6588ea695ab0de54de253a9c/notification/endpoint/slack.go#L44 and also as its is in [slack.flux](https://github.com/influxdata/flux/blob/master/stdlib/slack/slack.flux).

---

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -5811,7 +5811,7 @@ paths:
               $ref: "#/components/schemas/NotificationEndpoint"
       responses:
         '201':
-          description: Notification rule created
+          description: Notification endpoint created
           content:
             application/json:
               schema:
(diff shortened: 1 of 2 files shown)


kelwang:
Nice 🥇

Issue #15060, opened 2019-09-09
Cannot save as variable unless you edit the query


russorat:
the create button should be enabled as soon as a valid name for the query is entered.

![Kapture 2019-09-09 at 10 39 46](https://user-images.githubusercontent.com/4805997/64553446-3a14c580-d2ee-11e9-8e41-5050a1d8e0fc.gif)

PR #15115, opened 2019-09-11
fix(variables/components/VariableForm): check to see if form is valid

desa:
Closes https://github.com/influxdata/influxdb/issues/15060

Previously, the create button for a variable wouldn't be enabled unless
the user clicked in the text box, despite a query being present. This
adds an aditional check to the `isFormValid` method that ensures that if
a query is present the create button will be enabled.


desa:
Happy to do things differently if there's a better way

Issue #15104, opened 2019-09-10
Client URL should be optional for PagerDuty


russorat:
Client URL is not required for pagerduty and that field should be optional in the UI.

zoesteinkamp:
This issue is currently blocked by the backend changes in issue #15117

kelwang:
fixed with https://github.com/influxdata/influxdb/pull/15117

zoesteinkamp:
This issue was fixed by the backend change, im not going to add a test for it then close it out

PR #15117, opened 2019-09-11
fix(notification/endpoint): pagerduty clientURL should be optional

kelwang:
Partially https://github.com/influxdata/influxdb/issues/15104

the api and swagger changes for the pagerduty clientURL to be optional

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -9760,7 +9760,7 @@ components:
       allOf:
         - $ref: "#/components/schemas/NotificationEndpointBase"
         - type: object
-          required: [clientURL, routingKey]
+          required: [routingKey]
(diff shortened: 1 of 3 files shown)


desa:
@kelwang can you confirm that this change would be functional in flux? I believe that the issue we have is that the flux function requires it.

kelwang:
Hey @AlirieGray, do you know if it will work?

AlirieGray:
@kelwang is correct, the endpoint only takes in one optional parameter called URL which defaults to the pagerduty API URL (this is different from the client URL). The client URL is passed through the map function and not as a parameter.

PR #15121, opened 2019-09-11
feat(tasks): add WithMaxConcurrency to configure scheduler

AlirieGray:
Part one of idpe issue 4648

Adds a function `WithMaxConcurrency` to set the max concurrency for a scheduler 

- [ ] Rebased/mergeable
- [ ] Tests pass


task/backend/scheduler.go
```diff
@@ -470,19 +475,9 @@ func newTaskScheduler(
...
-	// this will be overwritten if the Flux script for the task has a concurrency set
-	if envConcurrency := os.Getenv("DEFAULT_CONCURRENCY"); envConcurrency != "" {
-		c, err := strconv.Atoi(envConcurrency)
-		if err == nil {
-			maxC = c
-		}
-	}
 
 	if opt.Concurrency != nil {
-		maxC = int(*opt.Concurrency)
```
stuartcarnie: You should no mutate the `TickScheduler.maxConcurrency` field, as a single task that defines a concurrency option will update the default concurrency for all other tasks.

task/backend/scheduler.go
```diff
@@ -470,19 +475,9 @@ func newTaskScheduler(
...
+		s.maxConcurrency = int(*opt.Concurrency)
```
stuartcarnie: You should no mutate the `TickScheduler.maxConcurrency` field, as a single task that defines a concurrency option will update the default concurrency for all other tasks.

AlirieGray: ahh ok, thanks for catching!

Issue #14183, opened 2019-06-24
Refactor Scheduler and Executor


lyondhill:
__Proposal:__
Write a scheduler that is more independent of the task system and can be used as an open source standalone library.

__Current behavior:__
The current scheduler is very tightly coupled to tasks. In addition it is doing scheduling, logging, tracing, and error management and correction.

__Desired behavior:__
We would like a scheduler that only handled scheduling and is portable. 

__Alternatives considered:__
When writing the new scheduler we want to ensure that it is portable so we need to keep influxdb imports and requirements to a minimum. We don't want the scheduler user to have to import any other influx libraries.

The framework for this new scheduler is laid out in https://github.com/influxdata/influxdb/blob/master/task/backend/scheduler/scheduler.go
With the interfaces in there we should be able to write a self contained scheduler.

### Responsibilities

A scheduler is responsible for understanding a cron-like schedule and ensuring these are executed at the desired frequency. Each task has its own schedule. The scheduler SHOULD schedule task runs in order. That is, for each task, the scheduler MUST call `Executor#Execute` with monotonically increasing time.

```go
type SchedulableStore interface {
	// UpdateLastScheduled updates the instance identified by ID with the new time
	UpdateLastScheduled(context.Context, ID, time.Time)
}

// Schedulable is the interface that encapsulates work that
// is to be executed on a specified schedule.
type Schedulable interface {
	// ID is the unique identifier for this Schedulable
	ID() ID

	// Schedule defines the frequency for which this Schedulable should be
	// queued for execution.
	Schedule() Schedule

	// Offset defines a negative or positive duration that should be added
	// to the scheduled time, resulting in the instance running earlier or later
	// than the scheduled time.
	Offset() time.Duration

	// LastScheduled specifies last time this Schedulable was queued
	// for execution.
	LastScheduled() time.Time
}
```

### TODOs

- [ ] Implement decoupled `NewScheduler` (naming) to replace `TickScheduler`
    - Will implement the `scheduler.Scheduler` interface
- [ ] Add unit tests for `NewScheduler`
    - [ ] Mock `Schedulable`
    - [ ] Mock `Executor`
- [ ] Create `type Time interface`
    - this type will be used to provide time to the `NewScheduler`
    - will be mockable for unit testing
- [x] Implement Cron parser to extended functionality

### Responsibilities

The executor is responsible for maintaining the concept of two queues to keep track of scheduled and running work. If the executor is restarted it MUST be capable of resuming outstanding work.

[...]

PR #15123, opened 2019-09-11
feat(task): Change interfaces defining scheduler and executor behavior

stuartcarnie:
This PR includes updates to the interfaces that define the behavior of a scheduler and an executor based on design discussions.

See #14183


stuartcarnie:
@GeorgeMac for reference, we discussed these changes and updated #14183 to an epic.

Issue #14823, opened 2019-08-27
influx CLI doesn't add user to org


gitirabassi:
```shell
$ influx org members add --name=alternative -o=0460a4b4b22c8000                                                                                                           acceptance:(ingress)
Error: Failed to add member: unknown resource type for permission.` 
```

gitirabassi:
@kelwang I'm not sure what i'm doing wrong:

- started the influxdb server
- `influx setup` and create the root token
- `influx user create --name foo`
- `influx org create --name bar`
- `influx org members add -n bar -o <user_id>` and fails with error `Error: Failed to add member: unknown resource type for permission.` 

Other problem, how to I make a user to be an org owner?

kelwang:
@gitirabassi , before you make sure both your influx and influxd are the newest of master. If it is the old cli or old influxd, it won't work.

For the add owners, could you create a new issue?

PR #15140, opened 2019-09-13
fix(cmd/influx): org add/remove members

kelwang:
Closes https://github.com/influxdata/influxdb/issues/14823

fix the base path issue for org members, add/remove.
More cleanup, after we have cli testing frame work merged.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influx/user_resource_mapping.go
+++ b/cmd/influx/user_resource_mapping.go
@@ -1,0 +1,33 @@
+package main
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/influxdata/influxdb"
+)
+
+func membersAddF(urm influxdb.UserResourceMapping, flgs Flags) error {
+	mappingSvc, err := newUserResourceMappingService(flgs)
+	if err != nil {
+		return fmt.Errorf("failed to initialize members service client: %v", err)
+	}
+
+	if err = mappingSvc.CreateUserResourceMapping(context.Background(), &urm); err != nil {
+		return fmt.Errorf("failed to add member: %v", err)
+	}
+	fmt.Printf("user %s has been added as a %s of %s: %s\n", urm.UserID, urm.UserType, urm.ResourceType, urm.ResourceID)
+	return nil
+}
+
+func membersRemoveF(resourceID, userID influxdb.ID, flgs Flags) error {
+	mappingSvc, err := newUserResourceMappingService(flgs)
+	if err != nil {
+		return fmt.Errorf("failed to initialize members service client: %v", err)
+	}
+	if err = mappingSvc.DeleteUserResourceMapping(context.Background(), resourceID, userID); err != nil {
+		return fmt.Errorf("failed to remove member: %v", err)
+	}
+	fmt.Printf("userID %s has been removed from ResourceID %s\n", userID, resourceID)
+	return nil
+}
--- a/http/user_resource_mapping_service.go
+++ b/http/user_resource_mapping_service.go
@@ -10,24 +10,25 @@ import (
 
 	"go.uber.org/zap"
 
-	platform "github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb"
 	"github.com/julienschmidt/httprouter"
 )
 
+// UserResourceMappingService is the struct of urm service
 // TODO(jm): how is basepath going to be populated?
 type UserResourceMappingService struct {
 	Addr               string
 	Token              string
 	InsecureSkipVerify bool
-	BasePath           string
+	basePath           string
 }
 
 type resourceUserResponse struct {
-	Role platform.UserType `json:"role"`
+	Role influxdb.UserType `json:"role"`
 	*userResponse
 }
 
-func newResourceUserResponse(u *platform.User, userType platform.UserType) *resourceUserResponse {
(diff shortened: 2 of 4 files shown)


cmd/influx/user_resource_mapping.go
```diff
@@ -0,0 +1,51 @@
...
+
+	if err = mappingSvc.CreateUserResourceMapping(context.Background(), &urm); err != nil {
+		return fmt.Errorf("failed to add member: %v", err)
+	}
+	writeUserResourceMapping(urm)
+	return nil
+}
+
+func membersRemoveF(resourceID, userID influxdb.ID, flgs Flags) error {
+	mappingSvc, err := newUserResourceMappingService(flags)
```
jademcgough: The arg passed in is `flgs`, not `flags`. Maybe change lines 12, 13 and 25 to be `flags` instead?

jademcgough: the only reason `flags` works here is because it's a global variable :)

http/user_resource_mapping_service.go
```diff
@@ -361,8 +364,10 @@ func (s *UserResourceMappingService) CreateUserResourceMapping(ctx context.Conte
 	return nil
 }
 
-func (s *UserResourceMappingService) DeleteUserResourceMapping(ctx context.Context, resourceID platform.ID, userID platform.ID) error {
-	url, err := NewURL(s.Addr, memberIDPath(s.BasePath, resourceID, userID))
+// DeleteUserResourceMapping will delete user resource mapping based in criteria.
+func (s *UserResourceMappingService) DeleteUserResourceMapping(ctx context.Context, resourceID influxdb.ID, userID influxdb.ID) error {
+	// default to use org resource type, and member resource type since it doesn't matter.
+	url, err := NewURL(s.Addr, resourceIDUserPath(influxdb.OrgsResourceType, resourceID, influxdb.Member, userID))
```
jademcgough: Why are `influxdb.OrgsResourceType` and `influxdb.Member` hardcoded here?

cmd/influx/user_resource_mapping.go
kelwang: good find, that's exactly why we don't want to use any global vars

cmd/influx/user_resource_mapping.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to initialize members service client: %v", err)
+	}
+	if err = mappingSvc.DeleteUserResourceMapping(context.Background(), resourceID, userID); err != nil {
+		return fmt.Errorf("failed to remove member: %v", err)
+	}
+	return nil
```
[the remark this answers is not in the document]
kelwang: I changed to fmt.Errorf("failed to remove member: resourceID: %s, userID: %s ,Err:%v", resourceID, userID, err). Is that okay. Seems resourceID and userID is the primary key.

http/user_resource_mapping_service.go
kelwang: we don't really have a delicate URL for removing members. We only have '/orgs/:id/members' or '/checks/:id/members', or whatever. The really path don't seems matter, as they all redirected to the newDeleteMemberHandler.

jademcgough:
@kelwang for the future, could you make new commits instead of rewriting the previous commit, and then squash/merge it after approval? It makes reviewing PRs a lot easier, since I can see what changed since the last commit :)

kelwang:
> @kelwang for the future, could you make new commits instead of rewriting the previous commit, and then squash/merge it after approval? It makes reviewing PRs a lot easier, since I can see what changed since the last commit :)

sure

cmd/influx/user_resource_mapping.go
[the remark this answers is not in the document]
jademcgough: I meant that we should have a message when removing a member is successful. I think the previous error message was okay.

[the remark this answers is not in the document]
kelwang: @jademcgough , mm, the http service will only output a 204 no content. Shall we update the json output? or we just say fmt.Printf("userID %s has been removed from ResourceID %s", userID, resourceID)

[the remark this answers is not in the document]
jademcgough: Our other delete functions show what was deleted:
https://github.com/influxdata/influxdb/blob/master/cmd/influx/bucket.go#L288-L303

But I think a printf statement like what you suggested should be good.

cmd/influx/user_resource_mapping.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func writeUserResourceMapping(urm influxdb.UserResourceMapping) {
```
jademcgough: I'd prefer just a printf saying that the user has been added as a member or owner. It's confusing/unhelpful to expose system internals that users don't care about.

http/user_resource_mapping_service.go
jademcgough: Looks like a bug introduced by https://github.com/influxdata/influxdb/commit/be37e07bea3270751fd28fd9c53ad4353147b29e. The behavior shouldn't be the same regardless of the path.

PR #15145, opened 2019-09-13
fix(tasks): report error codes to metrics

AlirieGray:
Closes idpe 4708

Errors in the new executor were previously changed to use influxdb.Error objects and report error codes to metrics in [this PR](https://github.com/influxdata/influxdb/pull/14926/).

This PR implements the same change for error reporting in the scheduler which is currently in use, which will give developers a more accurate picture of what system errors are occurring and need to be responded to. Will not need to be updated when the new scheduler goes into effect, as the new executor already handles this behavior. 

TODO: 
- [x] add metrics testing for errors in both old scheduler and new executor
- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/task/backend/scheduler.go
+++ b/task/backend/scheduler.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/influxdata/flux"
+	"github.com/influxdata/influxdb"
 	platform "github.com/influxdata/influxdb"
 	"github.com/influxdata/influxdb/kit/tracing"
 	"github.com/influxdata/influxdb/logger"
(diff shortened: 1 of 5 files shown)


task/backend/scheduler.go
```diff
@@ -779,36 +786,37 @@ func (r *runner) executeAndWait(ctx context.Context, qr QueuedRun, runLogger *za
...
 	if err == nil {
+		err = influxdb.ErrJsonMarshalError(err)
 		// authctx can be updated mid process
 		r.ts.nextDueMu.RLock()
 		authCtx := r.ts.authCtx
 		r.ts.nextDueMu.RUnlock()
 		r.taskControlService.AddRunLog(authCtx, r.task.ID, qr.RunID, time.Now(), string(b))
 	}
-	r.updateRunState(qr, RunSuccess, runLogger)
+	r.updateRunState(qr, RunSuccess, runLogger, err.(*influxdb.Error))
```
stuartcarnie: If an error is returned from `json.Marshal`, this will panic as `err` will not be an instance of `*influxdb.Error`

An alternative approach

```go
var influxErr *influxdb.Error
b, err := json.Marshal(stats)
if err != nil {
    influxErr = influxdb.ErrJsonMarshalError(err)
} else {
	r.ts.nextDueMu.RLock()
	authCtx := r.ts.authCtx
	r.ts.nextDueMu.RUnlock()
	r.taskControlService.AddRunLog(authCtx, r.task.ID, qr.RunID, time.Now(), string(b))
}
r.updateRunState(qr, RunSuccess, runLogger, influxErr)
```

task/backend/scheduler.go
```diff
@@ -819,6 +827,7 @@ func (r *runner) updateRunState(qr QueuedRun, s RunStatus, runLogger *zap.Logger
 		r.taskControlService.AddRunLog(r.ts.authCtx, r.task.ID, qr.RunID, time.Now(), "Completed successfully")
 	case RunFail:
 		r.ts.metrics.FinishRun(r.task.ID.String(), false, time.Since(qr.startedAt))
+		r.ts.metrics.LogError(err)
```
stuartcarnie: Given `RunSuccess` can pass an error if `json.Marshal` fails, you will can move this outside the switch and log an error if `err != nil` so that we can pass an error in for any value of `RunStatus`.

task/backend/scheduler.go
AlirieGray: ohh nice catch, I mistook the `==` for a `!=`

PR #15152, opened 2019-09-16
feat(auth): add jwt support in auth middleware

### LINKED ISSUE #14972: Add support to authorization middleware for jwt based Authorizer
Add support for deserialization / validation of JWT based authorizer within authentication middleware:
https://github.com/influxdata/influxdb/blob/master/http/authentication_middleware.go

## Steps:

1. When attempting to deserialize a token (given we have a JWT secret), first attempt to deserialize it as a JWT token:
https://github.com/influxdata/influxdb/blob/feda6dcd27acd501e43069ac569cf97e30cbda2c/http/authentication_middleware.go#L104-L116
2. If an error is returned because it is **not a well formed token** or no JWT secret exists then attempt to find authorization using the token as we currently do so.
3. If an error is returned because the token is well-formed, however, it is invalid (invalid signature, expired and so on) then return the error.
4. If the token is valid and the permissions can be deserialized then set the authorizer on the context.

## Success Criteria

When configured with a JWT secret, API actions can be triggered using a well formed token. e.g. a token with permissions with write action for a specific bucket within a specific org can be generated using a secret and then used to trigger a write.

GeorgeMac:
Closes #14972 

This PR adds support for JWT authorization to the http auth middleware.

This uses the new `jsonweb` package authorizer types.

When `Token` based authorization is used an attempt is first made to deserialize the payload as a JWT. If the token is well-formed, it is assumed that JWT based authorization is being attempted. The token is then validated and if successful the permission claims are passed down via the usual Authorizer on context strategy.
Given the token is not a well-formed JWT, then it is regarded as a lookup token for an associated Authorization in the AuthorizationService and treated as such. The Authorization is looked up using the token payload as the key.

Currently there is no decision as to how to configure a signing key, so this PR just introduces the mechanisms required for that work in the future. Currently an attempt to use a JWT will result in a no key found error. The token parser's KeyStore must be replace with a key store which returns a valid signing key for a given key id if JWT support is to be enabled.

UPDATE:

~This also now transforms each of our `Authorizer` types into the new `PermissionsAuthorizer` type which is an `Authorizer` which can returns its set of permissions. This is required to create a token from an `Authorizer`.~

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 1. [15151](https://github.com/influxdata/influxdb/pull/15151): Add jsonweb package for future JWT support
 1. [15168](https://github.com/influxdata/influxdb/pull/15168): Added the JMeter Template dashboard
+1. [15152](https://github.com/influxdata/influxdb/pull/15152): Add JWT support to http auth middleware
 
 ### UI Improvements
 
--- a/auth.go
+++ b/auth.go
@@ -8,13 +8,11 @@ import (
 // AuthorizationKind is returned by (*Authorization).Kind().
 const AuthorizationKind = "authorization"
 
-var (
-	// ErrUnableToCreateToken sanitized error message for all errors when a user cannot create a token
-	ErrUnableToCreateToken = &Error{
-		Msg:  "unable to create token",
-		Code: EInvalid,
-	}
-)
+// ErrUnableToCreateToken sanitized error message for all errors when a user cannot create a token
+var ErrUnableToCreateToken = &Error{
+	Msg:  "unable to create token",
+	Code: EInvalid,
+}
 
 // Authorization is an authorization. 🎉
 type Authorization struct {
--- a/http/authentication_middleware.go
+++ b/http/authentication_middleware.go
@@ -8,6 +8,7 @@ import (
 
 	platform "github.com/influxdata/influxdb"
 	platcontext "github.com/influxdata/influxdb/context"
+	"github.com/influxdata/influxdb/jsonweb"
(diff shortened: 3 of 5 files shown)


GeorgeMac:
See `UPDATE` in PR description

docmerlin:
@GeorgeMac This seems to be failing IDPE compatibility. you may want top update IDPE as well.

GeorgeMac:
Cheers @docmerlin 
As per external discussion we have decided to remove this new PermissionsAuthorizer interface now anyway. As its original intention has now been solved without any additional change being required.

GeorgeMac:
I re-requested review and added @jademcgough to this review as I needed to rebase over some recent changes to the auth middleware and could do with a sanity check.

After discussions with the compute team it was decided that permission based auth fits well with the JWT token. The Authorizer interface exposes both permission checking capabilities (via `Allowed()`) and an identity (via `GetUserID()`). Which gets confusing. So I decided to make the JWT Authorizer represent purely permissions and return an invalid ID when asked for a UserID.

The change that just went in just made the authentication middleware dependent on the UserID field of the Authorizer type. Now it has a lookup on the user ID to ensure the associated User is `active`. However, for the JWT token route we are purely interested in the permissions of the authorization and not the associated identity. They are for internal use and will expire shortly after issuing. This is how revokation ultimately will occur.

TL;DR

I updated my change to lean on the validity of the user ID in order to decided whether or not to make a user lookup for activity.
Given an invalid user ID (for example a JWT authorizer) no lookup is required.

PR #15166, opened 2019-09-16
fix(task): Improve Executor#Execute error consistency

### LINKED ISSUE #15161: Improve task query Executor#Execute to return more consistent errors for multiple implementations
`asyncQueryServiceExecutor` and `queryServiceExecutor` both implement the `backend.Executor` interface. The `Execute` method does not return consistent errors between these APIs. This issue is intended to unify the behavior as much as possible.

In summary this means moving most of the logic out of `asyncQueryServiceExecutor#Execute` and into the `asyncRunPromise` type.

stuartcarnie:
Implementations of the backend.Executor produce errors limited to querying the KV store. The remainder of the errors will be processed in the implementation of a `RunPromise`.

Closes #15161

### DIFF
--- a/task/backend/executor/executor.go
+++ b/task/backend/executor/executor.go
@@ -228,26 +228,7 @@ func (e *asyncQueryServiceExecutor) Execute(ctx context.Context, run backend.Que
 		return nil, err
 	}
 
-	pkg, err := flux.Parse(t.Flux)
-	if err != nil {
-		return nil, err
-	}
-
-	req := &query.Request{
-		Authorization:  t.Authorization,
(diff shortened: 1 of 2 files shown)


task/backend/executor/executor.go
```diff
@@ -312,10 +298,31 @@ func (p *asyncRunPromise) Cancel() {
 // followQuery waits for the query to become ready and sets p's results.
```
GeorgeMac: take it or leave it: update doc here

stuartcarnie: good catch; I'll fix that

PR #15174, opened 2019-09-17
refactor(dependencies): use new dependency injection framework

affo:
Connected to https://github.com/influxdata/flux/issues/1795.

This is the counterpart of https://github.com/influxdata/flux/pull/1879.

I moved dependency management from `readservice` to `stdlib/influxdata/influxdb` because of circular dependencies and because that seems its natural place.

Added a Launcher test to check that dynamic queries work.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,7 +36,7 @@ require (
 	github.com/hashicorp/go-msgpack v0.0.0-20150518234257-fa3f63826f7c // indirect
 	github.com/hashicorp/raft v1.0.0 // indirect
 	github.com/hashicorp/vault/api v1.0.2
-	github.com/influxdata/flux v0.47.0
+	github.com/influxdata/flux v0.47.1-0.20190917165150-71861fd0953f
 	github.com/influxdata/influxql v0.0.0-20180925231337-1cbfca8e56b6
 	github.com/influxdata/usage-client v0.0.0-20160829180054-6d3895376368
 	github.com/jessevdk/go-flags v1.4.0
--- a/query/control/controller.go
+++ b/query/control/controller.go
@@ -27,7 +27,6 @@ import (
 
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/codes"
-	"github.com/influxdata/flux/execute"
 	"github.com/influxdata/flux/lang"
 	"github.com/influxdata/flux/memory"
 	"github.com/influxdata/influxdb"
@@ -35,8 +34,9 @@ import (
 	"github.com/influxdata/influxdb/kit/prom"
 	"github.com/influxdata/influxdb/kit/tracing"
 	"github.com/influxdata/influxdb/query"
-	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
+
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
@@ -64,7 +64,7 @@ type Controller struct {
 
 	logger *zap.Logger
 
-	dependencies execute.Dependencies
+	dependencies flux.Dependency
 }
 
 type Config struct {
@@ -85,7 +85,7 @@ type Config struct {
 	// The context value must be a string or an implementation of the Stringer interface.
 	MetricLabelKeys []string
 
-	ExecutorDependencies execute.Dependencies
+	ExecutorDependencies flux.Dependency
 }
 
 func (c *Config) Validate() error {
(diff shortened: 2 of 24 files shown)


go.mod
```diff
@@ -36,7 +36,7 @@ require (
 	github.com/hashicorp/go-msgpack v0.0.0-20150518234257-fa3f63826f7c // indirect
 	github.com/hashicorp/raft v1.0.0 // indirect
 	github.com/hashicorp/vault/api v1.0.2
-	github.com/influxdata/flux v0.47.0
+	github.com/influxdata/flux v0.47.1-0.20190917165150-71861fd0953f
```
affo: This currently points to flux@feat/inject.

Once we will release v0.48 this will change

storage/wal/verifier_test.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 type Test struct {
-	dir string
+	dir          string
```
affo: Changes on this file are caused by go fmt

tsdb/tsm1/metrics_test.go
```diff
@@ -25,7 +25,7 @@ func TestMetrics_Filestore(t *testing.T) {
 	t2.AddBytes(200, 0)
 	t2.SetFileCount(map[int]uint64{0: 4, 4: 3, 5: 1})
 
-	t3.SetBytes(map[int]uint64{0: 300, 1: 500, 4:100, 5: 100})
```
affo: changes to this file are caused by go fmt

query/control/controller.go
```diff
@@ -27,16 +27,16 @@ import (
...
 	"github.com/influxdata/flux/memory"
 	"github.com/influxdata/influxdb"
 	"github.com/influxdata/influxdb/kit/errors"
 	"github.com/influxdata/influxdb/kit/prom"
 	"github.com/influxdata/influxdb/kit/tracing"
 	"github.com/influxdata/influxdb/query"
-	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
+
```
jsternberg: There's an extra newline here.

query/control/controller.go
```diff
@@ -85,7 +85,7 @@ type Config struct {
 	// The context value must be a string or an implementation of the Stringer interface.
 	MetricLabelKeys []string
 
-	ExecutorDependencies execute.Dependencies
+	ExecutorDependencies flux.Dependency
```
jsternberg: Should this be an array of dependencies or `flux.Dependencies`?

query/control/controller_test.go
```diff
@@ -45,7 +46,7 @@ var (
 		ConcurrencyQuota:         1,
 		MemoryBytesQuotaPerQuery: 1024,
 		QueueSize:                1,
-		ExecutorDependencies:     executetest.NewTestExecuteDependencies(),
+		ExecutorDependencies:     influxdb.Dependencies{FluxDeps: executetest.NewTestExecuteDependencies()},
```
jsternberg: Can you make this multiple lines so it looks like:

```go
influxdb.Dependencies{
    FluxDeps: executetest.NewTestExecuteDependencies(),
}
```

task/backend/analytical_storage_test.go
```diff
@@ -15,10 +14,13 @@ import (
 	"github.com/influxdata/influxdb/query"
 	_ "github.com/influxdata/influxdb/query/builtin"
 	"github.com/influxdata/influxdb/query/control"
+	stdlib "github.com/influxdata/influxdb/query/stdlib/influxdata/influxdb"
 	"github.com/influxdata/influxdb/storage"
+	"github.com/influxdata/influxdb/storage/reads"
 	"github.com/influxdata/influxdb/storage/readservice"
 	"github.com/influxdata/influxdb/task/backend"
 	"github.com/influxdata/influxdb/task/servicetest"
+
```
jsternberg: Extra newline.

query/control/controller.go
affo: That's not, that is something that implements `flux.Dependency`, the only thing that OSS needs  to know is that it can call `Inject` on it. Indeed, it passes a `influxdb.Dependencies`

jsternberg: I was thinking that each dependency would eventually be its own dependency and that `ExecutorDependencies` would just be a slice of all of the dependencies you wanted to inject into the executor. At the moment and possibly for longer, we would group some of the more common ones into a single object, but it might be better to prepare this function signature for something else.

affo: Yes, I see your point. I'll do that 👍

PR #15179, opened 2019-09-18
fix(alerts): remove the cron option from checks/rules

### LINKED ISSUE #15184: Remove cron field from check and notification rule in swagger
We can not currently generate the flux that the check/NR tasks need from a cron interval. Let's remove from swagger definition.

bednar:
Closes #15184

Remove the cron option from checks/rules.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -9345,9 +9345,6 @@ components:
         offset:
           description: Duration to delay after the schedule, before executing check.
           type: string
-        cron:
-          description: The check repetition interval in the form `* * * * * *`.
-          type: string
         tags:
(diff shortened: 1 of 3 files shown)


desa:
@bednar yup I agree with @ebb-tide We should just remove the cron option from checks/rules

ebb-tide:
https://github.com/influxdata/influxdb/issues/15184

bednar:
Ok, thanks for a clarification

bednar:
I have been change the PR to remove the cron from checks/rules.

ebb-tide:
thanks !

PR #15182, opened 2019-09-18
fix(tsi1): replace TSI compaction wait group with counter

benbjohnson:
Backport of #14902

Previously the TSI partition would panic if a compaction was started while Wait() was waiting. This commit removes the previous wait group and replaces it with a simple counter. The Wait() function now polls the counter until it reaches zero.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/tsdb/index/tsi1/partition.go
+++ b/tsdb/index/tsi1/partition.go
@@ -58,10 +58,11 @@ type Partition struct {
 	// Close management.
 	once    sync.Once
 	closing chan struct{} // closing is used to inform iterators the partition is closing.
-	wg      sync.WaitGroup
 
 	// Fieldset shared with engine.
 	fieldset *tsdb.MeasurementFieldSet
+
+	currentCompactionN int // counter of in-progress compactions
 
 	// Directory of the Partition's index files.
 	path string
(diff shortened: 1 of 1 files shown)


tsdb/index/tsi1/partition.go
```diff
@@ -322,9 +323,24 @@ func (p *Partition) buildSeriesSet() error {
...
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+	return p.currentCompactionN
+}
+
+// Wait will block until all compactions are finished.
+// Must only be called while they are disabled.
 func (p *Partition) Wait() {
-	p.wg.Wait()
+	ticker := time.NewTicker(10 * time.Millisecond)
```
e-dard: I forgot to add that we should include the changes I made to help with this. Can you cherry pick 030083e1a368725799b2a20b07d2f61d578778b5 into this PR?

benbjohnson: I pulled in the change but slightly tweaked it so it only has one invocation of `CurrentCompactionN()`

https://github.com/influxdata/influxdb/blob/23754e4b259548ef1295e38097fb3aea2bc9009f/tsdb/index/tsi1/partition.go#L333-L344

PR #15195, opened 2019-09-18
fix(query): correctly compute row number when echoing to() output

wolffcm:
This fix is the same as the one here:
https://github.com/influxdata/idpe/pull/4759/files#diff-458ae59fbf7d45cbf871044ade08f7fdR294


wolffcm:
I was totally wrong about this fix... The parameter to `AppendRecord` is the row offset in the `ColReader`, not the builder.  I'm glad our tests prevented me from "fixing" this :/

Issue #14974, opened 2019-09-05
Errors returned from query endpoint different from spec


chnn:
If I run the following query via `/api/v2/query`:

```
from(bucket: "baloney") |> range(start: -1s)
```

Then I get a `404 Not Found` with the following JSON body:

```
{
  "code": "not found",
  "message": "failed to initialize execute state",
  "error": {
    "code": "not found",
    "message": "could not find bucket \"baloney\""
  }
}
```

This schema isn't defined at all in the swagger:

https://github.com/influxdata/influxdb/blob/feda6dcd27acd501e43069ac569cf97e30cbda2c/http/swagger.yml#L3143-L3194

The `X-Influx-Error` or `X-Influx-Reference` headers are not present either.

It looks kinda like the default `Error` schema:

https://github.com/influxdata/influxdb/blob/feda6dcd27acd501e43069ac569cf97e30cbda2c/http/swagger.yml#L7231-L7262

But not quite the same.

What I would like to see: the default `Error` schema with a single descriptive, human-readable `message` field that we can display to users in the UI.

kelwang:
@jsternberg is it safe to update the swagger doc error to use the new error schema?

jsternberg:
Yes, I believe so.

russorat:
@kelwang @jsternberg do all other services return the errors in this new format?

nathanielc:
@russorat I agree, consistency is more important than accurate docs. We should make sure queryd is inline with other services and update the docs if needed.

goller:
We'd like to do this:

```yaml
diff --git a/http/swagger.yml b/http/swagger.yml
index c26a8a4a1..d1dc03ca8 100644
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -7275,6 +7275,7 @@ components:
           type: string
           format: uri
     Error:
       properties:
         code:
           description: code is the machine-readable error code.
@@ -7297,15 +7298,6 @@ components:
           readOnly: true
           description: message is a human-readable message.
           type: string
-        op:
-          readOnly: true
-          description: op describes the logical code operation during error. Useful for debugging.
-          type: string
-        err:
-          readOnly: true
-          description: err is a stack of errors that occurred during processing of the request. Useful for debugging.
-          type: string
-      required: [code, message]
     LineProtocolError:
       properties:
         code:

```

```json
{
  "code": "not found",
  "message": "failed to initialize execute state: could not find bucket \"baloney\"",
  "error": {
[...]

PR #15196, opened 2019-09-18
refactor: http error serialization matches the new error schema

jsternberg:
The http error schema has been changed to simplify the outward facing
API. The `op` and `error` attributes have been dropped because they
confused people. The `error` attribute will likely be readded in some
form in the future, but only as additional context and will not be
required or even suggested for the UI to use.

Errors are now output differently both when they are serialized to JSON
and when they are output as strings. The `op` is no longer used if it is
present. It will only appear as an optional attribute if at all. The
`message` attribute for an error is always output and it will be the
prefix for any nested error. When this is serialized to JSON, the
message is automatically flattened so a nested error such as:

    influxdb.Error{
        Msg: errors.New("something bad happened"),
        Err: io.EOF,
    }

This would be written to the message as:

    something bad happened: EOF

This matches a developers expectations much more easily as most
programmers assume that wrapping an error will act as a prefix for the
inner error.

This is flattened when written out to HTTP in order to make this logic
immaterial to a frontend developer.

The code is still present and plays an important role in categorizing
the error type. On the other hand, the code will not be output as part
of the message as it commonly plays a redundant and confusing role when
humans read it. The human readable message usually gives more context
and a message like with the code acting as a prefix is generally not
desired. But, the code plays a very important role in helping to
identify categories of errors and so it is very important as part of the
return response.

#14974

### DIFF
--- a/authorizer/task_test.go
+++ b/authorizer/task_test.go
@@ -285,7 +285,7 @@ from(bucket:"bad") |> range(start:-5m) |> to(bucket:"bad", org:"thing")`,
 					return fmt.Errorf(errfmt, "platform.Error.Err to be present", perr.Err)
 				}
 
-				if !strings.Contains(perr.Err.Error(), "<not found> bucket \"bad\" not found") {
+				if !strings.Contains(perr.Err.Error(), "bucket \"bad\" not found") {
 					return fmt.Errorf(errfmt, "to container bucket not found", perr.Err)
 				}
 
--- a/bolt/passwords.go
+++ b/bolt/passwords.go
@@ -16,13 +16,15 @@ var (
 	// EIncorrectPassword is returned when any password operation fails in which
 	// we do not want to leak information.
 	EIncorrectPassword = &platform.Error{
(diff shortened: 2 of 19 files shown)


jsternberg:
An example of this:

```
jsternberg:[~/g/s/g/i/influxdb] $ (docs/swagger-query-endpoint *$=)
curl -s -H "Authorization: Token $INFLUXDB_TOKEN" -XPOST http://localhost:9999/api/v2/query'?org=influxdata' -H 'Content-Type: application/json' -d '{"query":"from(bucket:\\"nope\\") |> range(start: -5m)"}'
{"code":"not found","message":"failed to initialize execute state: could not find bucket \"nope\""}                                                           jsternberg:[~/g/s/g/i/influxdb] $ (docs/swagger-query-endpoint *$=) go install ./cmd/influx
jsternberg:[~/g/s/g/i/influxdb] $ (docs/swagger-query-endpoint *$=) influx repl -o influxdata
> from(bucket: "nope") |> range(start: -5m)
Error: failed to initialize execute state: could not find bucket "nope"
```

PR #15197, opened 2019-09-19
fix(tsm1): Fix duplicate points

stuartcarnie:
All seeks must be added to the c.current slice so the min and max read values can be updated on each read pass.


stuartcarnie:
Yes, unfortunately an improvement to reuse the “current” slice also optimized excluding seeks that were outside the range.

jacobmarble:
Some grade A work here. 100+ lines of regression test and 8 lines of deleted code.

PR #15206, opened 2019-09-19
feat(query/stdlib/experimental): add experimental.to() implementation to OSS

wolffcm:
There was already an implementation of `experimental.to()` in enterprise, this adds it to OSS as well.

### DIFF
--- a/query/stdlib/experimental/to.go
+++ b/query/stdlib/experimental/to.go
@@ -1,14 +1,22 @@
 package experimental
 
 import (
+	"context"
+	"errors"
 	"fmt"
 
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/codes"
+	"github.com/influxdata/flux/execute"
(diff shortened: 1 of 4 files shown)


query/stdlib/experimental/to.go
```diff
@@ -157,3 +166,309 @@ func newToProcedure(qs flux.OperationSpec, a plan.Administration) (plan.Procedur
...
+func (t *ToTransformation) Finish(id execute.DatasetID, err error) {
+	if err != nil {
+		t.d.Finish(err)
+		return
+	}
+	err = t.buf.Flush(t.ctx)
+	t.d.Finish(err)
+}
+
+const (
```
wolffcm: From here all the way down to `writeTable`, this code was copied from IDPE to here so it can be reused in both places.

wolffcm:
Corresponding PR in idpe: https://github.com/influxdata/idpe/pull/4765

PR #15212, opened 2019-09-19
feat(influxdb): User disabling

bthesorceror:
#4751

Allow for users status to be updated therefore user can be disabled. There is also a fair bit of refactoring to ensure that notifications and checks reflect the status of their underlying task. This greatly simplified ensuring that all user tasks were disabled as well.

the first commit is only showing up because of interactive rebase, please ignore as that is already in master.

- [ ] Rebased/mergeable
- [x] Tests pass


http/check_service.go
```diff
@@ -47,6 +49,7 @@ type CheckHandler struct {
 	influxdb.HTTPErrorHandler
 	Logger *zap.Logger
 
+	TaskService                influxdb.TaskService
```
kelwang: i think we should do this in KV. instead of the http layer. Just make check.Status inherit from task. in the findCheck service. There is no need to any extra hack for creation and update.

user.go
```diff
@@ -4,11 +4,29 @@ import (
...
+
 // User is a user. 🎉
 type User struct {
-	ID      ID     `json:"id,omitempty"`
-	Name    string `json:"name"`
-	OAuthID string `json:"oauthID,omitempty"`
+	ID      ID         `json:"id,omitempty"`
+	Name    string     `json:"name"`
+	OAuthID string     `json:"oauthID,omitempty"`
+	Status  UserStatus `json:"status,omitempty"`
```
kelwang: shall we use the influxdb.Status instead of creating new status?

bthesorceror: 👍

http/check_service.go
bthesorceror: If we push this down into the KV layer then we will probably want to do a lookup from the store at the end of create, update and patch. Does anyone have objections to that?

bthesorceror: As opposed to modifying the check in place.

notification.go
```diff
@@ -156,6 +142,12 @@ func (f NotificationRuleFilter) QueryParams() map[string][]string {
 	return qp
 }
 
+// NotificationRuleCreate is the struct providing data to create a Notification Rule.
+type NotificationRuleCreate struct {
+	NotificationRule
+	Status *Status `json:"status"`
```
kelwang: Could you change this to struct instead of pointer?

check.go
```diff
@@ -76,6 +80,12 @@ type CheckUpdate struct {
 	Description *string `json:"description,omitempty"`
 }
 
+// CheckCreate represent data to create a new Check
+type CheckCreate struct {
+	Check
+	Status Status
```
kelwang: add `json:"status"`

PR #15213, opened 2019-09-20
Don't write secrets for orgs that don't exist

### LINKED ISSUE #15208: When using the secrets API, the API should not save to an OrgID that does not exist
Kept running into errors while trying to use Flux to retrieve secrets from the secret store.
https://github.com/influxdata/flux/issues/1885

Turns out I sent in the wrong OrgID in:
http://localhost:9999/api/v2/orgs/<orgID>/secrets 

I think I used my cloud orgId? At any rate - setting and getting secrets from the API worked flawlessly, but when I tried to use the secrets with secrets.get(), it failed because my session was of course in my org.

The API should not save secrets into an OrgID if that OrgID doesn't exist.  

(As an aside, I imagine this could also lead to even more weirdness if by chance a new org is created and some secrets had already been stored in there? )

imogenkinsman:
Closes #15208

Check to see if an organization exists before writing secrets for it.


russorat:
@jademcgough is this still relevant?

PR #15222, opened 2019-09-20
feat(httpd): Add option to authenticate debug/pprof and ping endpoints

gunnaraasen:
Adds `pprof-auth-enabled` and `ping-auth-enabled` configuration options which can enable authentication on all `/debug` and `/ping` endpoints. This also covers the prometheus `/metrics` and deprecated `/status` endpoints.

With these additions, all InfluxDB endpoints will require authentication with the following configuration:

```
[http]
  auth-enabled = true
  pprof-auth-enabled = true
  ping-auth-enabled = true
```

Note that `go tool pprof` does not support authentication so it will not work when `pprof-auth-enabled` is enabled.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -20,6 +20,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	httppprof "net/http/pprof"
+
 	"github.com/bmizerany/pat"
 	"github.com/dgrijalva/jwt-go"
 	"github.com/gogo/protobuf/proto"
@@ -154,6 +156,19 @@ func NewHandler(c Config) *Handler {
 		writeLogEnabled = false
 	}
 
+	var authWrapper func(handler func(http.ResponseWriter, *http.Request)) interface{}
+	if h.Config.AuthEnabled && h.Config.PingAuthEnabled {
+		authWrapper = func(handler func(http.ResponseWriter, *http.Request)) interface{} {
+			return func(w http.ResponseWriter, r *http.Request, user meta.User) {
+				handler(w, r)
+			}
+		}
+	} else {
+		authWrapper = func(handler func(http.ResponseWriter, *http.Request)) interface{} {
+			return handler
+		}
+	}
+
 	h.AddRoutes([]Route{
 		Route{
 			"query-options", // Satisfy CORS checks.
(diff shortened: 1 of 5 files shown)


services/httpd/handler.go
```diff
@@ -154,6 +156,19 @@ func NewHandler(c Config) *Handler {
 		writeLogEnabled = false
 	}
 
+	var authWrapper func(handler func(http.ResponseWriter, *http.Request)) interface{}
+	if h.Config.AuthEnabled && h.Config.PingAuthEnabled {
+		authWrapper = func(handler func(http.ResponseWriter, *http.Request)) interface{} {
+			return func(w http.ResponseWriter, r *http.Request, user meta.User) {
```
jacobmarble: I'm looking for something about auth here, but this just looks like a no-op handler wrapper. Can you explain?

services/httpd/handler.go
```diff
@@ -185,26 +200,67 @@ func NewHandler(c Config) *Handler {
...
+			"HEAD", "/status", false, true, authWrapper(h.serveStatus),
 		},
 		Route{
 			"prometheus-metrics",
-			"GET", "/metrics", false, true, promhttp.Handler().ServeHTTP,
+			"GET", "/metrics", false, true, authWrapper(promhttp.Handler().ServeHTTP),
 		},
 	}...)
 
+	// When PprofAuthEnabled is enabled, create debug/pprof endpoints with the
```
jacobmarble: What if `!h.Config.AuthEnabled && h.Config.PprofEnabled`? Where are the endpoints served from in that case?

jacobmarble: Oh, looks like you're adding the pprof endpoints to `h.mux`, then using that later.

services/httpd/handler.go
```diff
@@ -376,7 +432,11 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	w.Header().Add("X-Influxdb-Version", h.Version)
 	w.Header().Add("X-Influxdb-Build", h.BuildType)
 
-	if strings.HasPrefix(r.URL.Path, "/debug/pprof") && h.Config.PprofEnabled {
```
jacobmarble: I prefer to wrap the existing `handleProfiles` method, rather than adding to `h.mux`. Can you reuse the `authenticate()` func for this?

services/httpd/handler.go
gunnaraasen: [`AddRoutes`](https://github.com/influxdata/influxdb/pull/15222/files#diff-719dcaf997cf7e555e32df7f85e7cd60R390) will wrap the handler in `authenticate` if it matches the `func(w http.ResponseWriter, r *http.Request, user meta.User)` signature.

services/httpd/handler.go
gunnaraasen: Endpoints are served from [here](https://github.com/influxdata/influxdb/pull/15222/files#diff-719dcaf997cf7e555e32df7f85e7cd60R439) when any of the three properties are not true. This keeps existing behavior for all cases except when configured. I'll document that in the config description.

services/httpd/handler.go
gunnaraasen: I thought adding the authentication through `mux` would be cleaner than wrapping the handlers with `authenticate` here. Especially since the same wrapping pattern is used for `servePing`. Going through `AddRoutes` wraps a logging handler as well, which I thought might be useful.

services/httpd/handler.go
gunnaraasen: [`AddRoutes`](https://github.com/influxdata/influxdb/pull/15222/files#diff-719dcaf997cf7e555e32df7f85e7cd60R390) will wrap the handler in `authenticate` if it matches the `func(w http.ResponseWriter, r *http.Request, user meta.User)` signature.

services/httpd/handler.go
gunnaraasen: Yup, endpoints are served from the existing handlers [here](https://github.com/influxdata/influxdb/pull/15222/files#diff-719dcaf997cf7e555e32df7f85e7cd60R439) when any of the three properties are not true. This keeps existing behavior for all cases except when configured. I'll add a better explanation in the example config.

gunnaraasen:
I've updated the example config with a note about the `auth-enabled` and `pprof-enabled` dependent settings. Will merge once tests pass.

jacobmarble:
If someone needs a profile from an OSS instance with pprof auth enabled, I
guess they grab it with curl (or whatever) instead of the `go` CLI?

gunnaraasen:
Yup, a profile would need to be pulled with `curl`, for example:

```
curl -o profiles.tar.gz -u username:password "http://localhost:8086/debug/pprof/all?cpu=true"
```

dontsovcmc:
Is it correct: I turn on ping authentification, but InfluxDB responses anonymous request:

```
[http]
  enabled = true
  auth-enabled = true
  pprof-enabled = true 
  pprof-auth-enabled = true
  debug-pprof-enabled = false
  ping-auth-enabled = true
```
```
curl http://localhost:8086/ping -v
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8086 (#0)
> GET /ping HTTP/1.1
> Host: localhost:8086
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 204 No Content
< Content-Type: application/json
< Request-Id: 1c8ddf79-251b-11ea-8007-0242ac180002
< X-Influxdb-Build: OSS
< X-Influxdb-Version: 1.7.9
< X-Request-Id: 1c8ddf79-251b-11ea-8007-0242ac180002
< Date: Mon, 23 Dec 2019 00:28:16 GMT
< 
* Connection #0 to host localhost left intact
```

PR #15226, opened 2019-09-23
Panic when failing to write WAL file

### LINKED ISSUE #9923: client got "no space left on device"  for 1.4.2
influxDB: 1.4.2 
CentOS7

when send info to influxDB, I got messages like below. It gave me tips aoubt no space, but I checked on server, here are lots of spaces over there. Any ideas?

write influx error org.influxdb.InfluxDBException: engine: error writing WAL entry: write /var/lib/influxdb/wal/kafka/autogen/335/_00107.wal: no space left on device 
at org.influxdb.InfluxDBException.buildExceptionFromErrorMessage(InfluxDBException.java:154) 
at org.influxdb.InfluxDBException.buildExceptionForErrorState(InfluxDBException.java:166) 
at org.influxdb.impl.InfluxDBImpl.execute(InfluxDBImpl.java:584) 
at org.influxdb.impl.InfluxDBImpl.write(InfluxDBImpl.java:355) 

### Directions
_GitHub Issues are reserved for actionable bug reports and feature requests._
_General questions should be sent to the [InfluxDB Community Site](https://community.influxdata.com)._

_Before opening an issue, search for similar bug reports or feature requests on GitHub Issues._
_If no similar issue can be found, fill out either the "Bug Report" or the "Feature Request" section below.
_Erase the other section and everything on and above this line._

### Bug report

__Steps to reproduce:__

1. [First Step]
2. [Second Step]
3. [and so on...]

Also, if this is an issue of for performance, locking, etc the following commands are useful to create debug information for the team.

```

iostat -xd 1 30 > iostat.txt
```

**Please note** It will take at least 30 seconds for the first cURL command above to return a response.
This is because it will run a CPU profile as part of its information gathering, which takes 30 seconds to collect.
Ideally you should run these commands when you're experiencing problems, so we can capture the state of the system at that time.

If you're concerned about running a CPU profile (which only has a small, temporary impact on performance), then you can set `?cpu=false` or omit `?cpu=true` altogether.

Please run those if possible and link them from a [gist](http://gist.github.com) or simply attach them as a comment to the issue.

*Please note, the quickest way to fix a bug is to open a Pull Request.*

### Feature Request

Opening a feature request kicks off a discussion.
Requests may be closed if we're not actively planning to work on them.

__Proposal:__ [Description of the feature]

__Current behavior:__ [What currently happens]

__Desired behavior:__ [What you would like to happen]

__Use case:__ [Why is this important (helps with prioritizing requests)]

foobar:
This fixed #9923

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/tsdb/engine/tsm1/wal.go
+++ b/tsdb/engine/tsm1/wal.go
@@ -431,7 +431,11 @@ func (l *WAL) writeToLog(entry WALEntry) (int, error) {
 
 		// write and sync
 		if err := l.currentSegmentWriter.Write(entry.Type(), compressed); err != nil {
-			return -1, fmt.Errorf("error writing WAL entry: %v", err)
+			// If a write failed, most likely there was some IO error and the underlying
+			// buffer writer can hardly recover from it.
+			// panicking and restarting the process can at least truncate the corrupt WAL
+			// entry. Escalating the critical problem also avoids silent failure.
+			panic(err)
 		}
 
 		select {


foobar:
> If I understand from your issue, the problem is that even if the underlying error is ephemeral, the segment can get left in a bad state where it can't be appended to. Is that right?

yes, and this is silent failure

> If so, then a better way to handle this unrecoverable error would be to truncate the segment to the point we can write to it again.

After restart it will truncate the bad WAL entry.
Otherwise it looks complicated to recover: it could be partial write and we may have to rescan the whole segment file.
Basically the panic is not only for recovering, but also to avoid silent failure and escalate the problem so an administrator can be alerted.

PR #15248, opened 2019-09-24
fix(storage): remove level=0 from TSM disk bytes metrics.

Elbehery:
This commit removes the level=0 label from TSM Disk Usage metrics, since TSM level starts always (from 1 up to 4+).
    ex:

 before fix  :

```
            storage_tsm_files_disk_bytes{engine_id="0",level="0",node_id="0"} 0
            storage_tsm_files_disk_bytes{engine_id="0",level="1",node_id="0"} 3.3855125e+07
            storage_tsm_files_disk_bytes{engine_id="0",level="2",node_id="0"} 2.3737882e+07
            storage_tsm_files_disk_bytes{engine_id="0",level="3",node_id="0"} 1.86620984e+08
```

After fix :

```
            storage_tsm_files_disk_bytes{engine_id="0",level="1",node_id="0"} 3.3855125e+07
            storage_tsm_files_disk_bytes{engine_id="0",level="2",node_id="0"} 2.3737882e+07
            storage_tsm_files_disk_bytes{engine_id="0",level="3",node_id="0"} 1.86620984e+08
```

### DIFF
--- a/tsdb/tsm1/file_store.go
+++ b/tsdb/tsm1/file_store.go
@@ -376,6 +376,14 @@ func (t *fileTracker) ClearFileCounts() {
 	}
 }
 
+func (t *fileTracker) ClearDiskSizes() {
+	labels := t.Labels()
+	for i := uint64(0); i <= 4; i++ {
+		labels["level"] = formatLevel(i)
+		t.metrics.DiskSize.With(labels).Set(float64(0))
+	}
+}
+
 func formatLevel(level uint64) string {
 	if level >= 4 {
 		return "4+"
@@ -676,9 +684,9 @@ func (f *FileStore) Open(ctx context.Context) error {
 	}
 
 	var lm int64
-	counts := make(map[int]uint64, 5)
-	sizes := make(map[int]uint64, 5)
-	for i := 0; i <= 5; i++ {
+	counts := make(map[int]uint64, 4)
+	sizes := make(map[int]uint64, 4)
+	for i := 1; i <= 4; i++ {
 		counts[i] = 0
 		sizes[i] = 0
 	}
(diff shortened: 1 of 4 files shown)


tsdb/tsm1/file_store.go
```diff
@@ -377,6 +377,11 @@ func (t *fileTracker) ClearFileCounts() {
 }
 
 func formatLevel(level uint64) string {
+	// sanity check to avoid future bugs.
+	if level == 0 {
+		panic("TSM Files' Level '0' is not supported anymore.")
```
sebito91: We should never `panic` if we can avoid it, especially for a `level0` value here. If anything let's just log and move on but a panic I feel is too strong.

Elbehery:
@sebito91  Please check the updates 👍

Elbehery:
@e-dard Before the statistics being recalculated, we clear the counter in 
https://github.com/influxdata/influxdb/blob/ff2747a72885ac113b81fa440d97274527841676/tsdb/tsm1/file_store.go#L1017 .

I think I should just change the map size and the initialiser, or ?

e-dard:
@Elbehery yes I think you're right. But we also need on of those clear methods for the disk bytes metric, too. Not sure why it's currently missing.

Elbehery:
@e-dard ok I will close this on the initialiser, and we can open another issue for the desk metrics, ok ?

e-dard:
@Elbehery I think it would be fine just to fix it in this PR.

tsdb/tsm1/file_store.go
```diff
@@ -376,6 +376,14 @@ func (t *fileTracker) ClearFileCounts() {
 	}
 }
 
+func (t *fileTracker) ClearDiskSizes() {
+	labels := t.Labels()
+	for i := uint64(0); i <= 4; i++ {
```
e-dard: Doesn't this need to start at `1`?

Elbehery: Yes sure, sorry the I did not see the '0' wrapped into the type :(

Elbehery:
alchemy please merge

PR #15255, opened 2019-09-24
feat(http): delete endpoint

### LINKED ISSUE #13215: implement delete with predicate endpoint

kelwang:
Closes #13215

Add predicate endpoint, service, swagger, permission, test cases.

- [] Rebased/mergeable
- [x] Tests pass


http/requests.go
```diff
@@ -41,3 +45,37 @@ func queryOrganization(ctx context.Context, r *http.Request, svc platform.Organi
...
+// queryBucket returns the bucket for any http request.
+//
+// It checks the bucket= and then bucketID= parameter of the request.
+//
+// This will try to find the bucket using an ID string or
+// the name.  It interprets the &bucket= parameter as either the name
+// or the ID.
+func queryBucket(ctx context.Context, r *http.Request, svc platform.BucketService) (b *platform.Bucket, err error) {
+
+	filter := platform.BucketFilter{}
```
goller: extra newline

http/requests.go
```diff
@@ -41,3 +45,37 @@ func queryOrganization(ctx context.Context, r *http.Request, svc platform.Organi
...
+	if reqID := r.URL.Query().Get(BucketID); reqID != "" {
+		filter.ID, err = platform.IDFromString(reqID)
+		if err != nil {
+			return nil, err
+		}
+	}
+	if filter.ID == nil && filter.Name == nil {
+		return nil, &platform.Error{
+			Code: platform.EInvalid,
+			Msg:  "Please provide either bucketID or bucket",
```
goller: perhaps bucket required (although if in a route then this code can be removed)

predicate/logical.go
```diff
@@ -0,0 +1,116 @@
...
+				Code: influxdb.EInvalid,
+				Msg:  fmt.Sprintf("Err in Child %d, err: %s", k, err.Error()),
+			}
+		}
+	}
+	return nil
+}
+
+// ToDataType convert a LogicalNode to datatypes.Node.
+func (n LogicalNode) ToDataType() (*datatypes.Node, error) {
```
goller: I can't tell about the test coverage here... do these files in this package have extensive tests?

http/delete_handler.go
```diff
@@ -0,0 +1,224 @@
...
+				Msg:  "stop time is not valid RFC3339",
+			}
+		}
+		dr.Stop = stop.UnixNano()
+	}
+	node, err := predicate.UnmarshalJSON(drd.Predicate)
+	if err != nil {
+		return err
+	}
+	dr.Predicate, err = predicate.New(node)
```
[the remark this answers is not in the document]
kelwang: yes

http/api_handler.go
```diff
@@ -229,6 +234,7 @@ var apiLinks = map[string]interface{}{
 	"telegrafs": "/api/v2/telegrafs",
 	"users":     "/api/v2/users",
 	"write":     "/api/v2/write",
+	"delete":    "/api/v2/delete",
```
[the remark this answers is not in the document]
kelwang: I like this idea initially, 
then i realized we already have an endpoint for DELETE /api/v2/buckets/{id}, which will completely delete this bucket.
The difference is the predicate delete will only delete the time series of the bucket.

I wonder if this could trigger some confusion?
 
what do you think? @russorat , @ebb-tide , @desa , @jademcgough

http/delete_test.go
```diff
@@ -0,0 +1,362 @@
...
+				statusCode:  http.StatusBadRequest,
+				contentType: "application/json; charset=utf-8",
+				body: fmt.Sprintf(`{
+					"code": "invalid",
+					"message": "Please provide either bucketID or bucket"
+				  }`),
+			},
+		},
+		{
+			name: "insufficient permissions delete",
```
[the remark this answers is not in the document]
kelwang: Thanks, @goller 
Do I need another test for token auth as well? or Just add the session cookie name will be enough.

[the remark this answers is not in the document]
goller: @kelwang yeah , or, I'd use a mock authorizer as you are testing the unit not sessions

http/delete_test.go
```diff
@@ -0,0 +1,362 @@
...
+						return nil, &influxdb.Error{
+							Code: influxdb.EInvalid,
+							Msg:  "Please provide either orgID or org",
+						}
+					},
+				},
+			},
+			wants: wants{
+				statusCode: http.StatusNoContent,
+				body:       fmt.Sprintf(``),
```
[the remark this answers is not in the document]
kelwang: Yeah,

http/api_handler.go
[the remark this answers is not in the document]
russorat: @goller using that logic, wouldn't `write` also be under a bucket id? i'm not qualified to comment on this but my opinion is that using the name of an HTTP method in the url may cause confusion for people.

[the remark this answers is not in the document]
goller: @russorat I just spoke to @kelwang and @jademcgough about this.  I can restate my primary concerns now... basically, this endpoint has another predicate system.

Perhaps it would be better to have only one predicate system like flux?

Perhaps this delete logic should actually be placed in /api/v2/query depending on the long term vision?

delete.go
```diff
@@ -0,0 +1,14 @@
+package influxdb
+
+import "context"
+
+// Predicate is something that can match on a series key.
+type Predicate interface {
```
e-dard: I'm wondering why this is needed. We have the same interface in `tsm1.Predicate`.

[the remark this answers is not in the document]
kelwang: Hi, @e-dard . I have seen some complex import circle issues. So I ended up copy them to them to the domain level

http/swagger.yml
```diff
@@ -1640,6 +1640,71 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+  /delete:
+    post:
+      summary: delete time-series data from influxdb
+      requestBody:
+          description: predicate delete request
```
sanderson: General nits regarding the format of summaries and descriptions in the swagger.yml – we now auto-generate API docs in the v2 documentation using this swagger file. Because these docs are now hosted and public facing, all summaries and descriptions should be sentence-cased. Descriptions (except response descriptions) should end with a period. 

Other notes:
- "Time series" should not be hyphenated.
- "influxdb" should be formatted as a proper noun: InfluxDB.

delete.go
```diff
@@ -0,0 +1,14 @@
...
+	Matches(key []byte) bool
+	Marshal() ([]byte, error)
+}
+
+// DeleteService will delete a bucket from the range and predict.
```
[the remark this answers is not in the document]
kelwang: Hey @goller, @e-dard 
tsm1 has the interface of Predicate, which imports influxdb. So the  as long as the DeleteService is in the influxdb layer it won't really work. We could move the DeleteService to the storage layer, I feel it's more like an anti-pattern thing. Since similar service like WriteService is in the domain level. What do you guys think?

http/delete_handler.go
```diff
@@ -0,0 +1,207 @@
...
+	err := json.Unmarshal(b, &drd)
+	if err != nil {
+		return &influxdb.Error{
+			Code: influxdb.EInvalid,
+			Msg:  "Invalid delete predicate node request",
+			Err:  err,
+		}
+	}
+	*dr = deleteRequest{}
+	start, err := time.Parse(time.RFC3339Nano, drd.Start)
```
jsteenb2: this might be nice to pull out into a resusable type, something like https://play.golang.org/p/-UH4K1WIJKW . Just a suggestion, there is probably better ways 🤷‍♂, definitely not mandatory

http/delete_handler.go
```diff
@@ -0,0 +1,207 @@
...
+	if err != nil {
+		return &influxdb.Error{
+			Code: influxdb.EInvalid,
+			Op:   "http/Delete",
+			Msg:  "invalid RFC3339Nano for field start",
```
e-dard: Is the user of the API or the UI going to see this message? I think it could be clearer what the problem is. The user might not understand for example that `RFC3339Nano` is a particular time format.

http/delete_handler.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+	}
+	dr.Start = start.UnixNano()
+
+	stop, err := time.Parse(time.RFC3339Nano, drd.Stop)
+	if err != nil {
+		return &influxdb.Error{
+			Code: influxdb.EInvalid,
+			Op:   "http/Delete",
+			Msg:  "invalid RFC3339Nano for field stop",
```
e-dard: Same comment as above.

http/delete_test.go
```diff
@@ -0,0 +1,355 @@
...
+				queryParams: map[string][]string{
+					"org":    []string{"org1"},
+					"bucket": []string{"buck1"},
+				},
+				body: []byte(`{
+					"start":"2009-01-01T23:00:00Z",
+					"stop":"2019-11-10T01:00:00Z",				
+					"nodeType": "logical",
+					"operator":"and",
+					"children":[
```
e-dard: I'm still not convinced by the user experience here. I would much rather see the users provide predicates in a more elegant way. Even something like influxQL predicates would be better.... `foo = "bar" AND a = "b"`.

http/delete_test.go
```diff
@@ -0,0 +1,355 @@
...
+			name: "complex delete",
+			args: args{
+				queryParams: map[string][]string{
+					"org":    []string{"org1"},
+					"bucket": []string{"buck1"},
+				},
+				body: []byte(`{
+					"start":"2009-01-01T23:00:00Z",
+					"stop":"2019-11-10T01:00:00Z",				
+					"nodeType": "logical",
```
e-dard: I would expect to see a `predicate` key in the top-level object. Instead it looks like the contents of the predicate are mixed with the `start` and `stop` keys.

http/delete_handler.go
```diff
@@ -0,0 +1,207 @@
...
+		DeleteService:       b.DeleteService,
+		OrganizationService: b.OrganizationService,
+	}
+
+	h.HandlerFunc("POST", deletePath, h.handleDelete)
+	return h
+}
+
+func (h *DeleteHandler) handleDelete(w http.ResponseWriter, r *http.Request) {
+	span, r := tracing.ExtractFromHTTPRequest(r, "DeleteHandler")
```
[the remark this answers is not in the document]
kelwang: i am not familiar with this one either. Could you create another issue for that?

http/delete_handler.go
kelwang: added an example for RFC3339Nano

http/delete_handler.go
kelwang: yeah, i like the reusable time type idea. I think we have several places like that already. Maybe we should create a new issue to track and convert them all.

http/delete_test.go
```diff
@@ -0,0 +1,384 @@
...
+			name: "unsupported delete",
+			args: args{
+				queryParams: map[string][]string{
+					"org":    []string{"org1"},
+					"bucket": []string{"buck1"},
+				},
+				body: []byte(`{
+					"start":"2009-01-01T23:00:00Z",
+					"stop":"2019-11-10T01:00:00Z",
+					"nodeType": "logical",
```
e-dard: Why do we still have a `nodeType` field?

predicate/logical.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/influxdata/influxdb/storage/reads/datatypes"
+)
+
+// LogicalOperator is a string type of logical operator.
+type LogicalOperator int
+
+// LogicalOperators
+var (
+	LogicalAnd LogicalOperator = 1
+	LogicalOr  LogicalOperator = 2
```
e-dard: SInce we don't support OR why do we have this operator present?

predicate/logical.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+// Value returns the node logical type.
+func (op LogicalOperator) Value() (datatypes.Node_Logical, error) {
+	switch op {
+	case LogicalAnd:
+		return datatypes.LogicalAnd, nil
+	case LogicalOr:
```
e-dard: I think you can remove this section?

predicate/parser.go
```diff
@@ -0,0 +1,250 @@
...
+				}
+				if err != nil {
+					return *n, err
+				}
+				n.Children = append(n.Children[:len(n.Children)-1], LogicalNode{
+					Children: []Node{lastChild, n1},
+					Operator: LogicalAnd,
+				})
+			}
+		case influxql.OR:
```
e-dard: Rather than parsing expressions we don't support could we not just make them parser errors here? So just remove this case and then they can't be parsed?

predicate/parser.go
```diff
@@ -0,0 +1,250 @@
...
+		default:
+			return *n, &influxdb.Error{
+				Code: influxdb.EInvalid,
+				Msg:  fmt.Sprintf("bad logical expression, at position %d", pos.Char),
+			}
+		}
+	}
+}
+
+func (p *parser) parseTagRuleNode() (TagRuleNode, error) {
```
e-dard: There is a lot of stuff here we will parse but not support. It seems to me like it would be better to fail at the parser stage.

predicate/parser_test.go
```diff
@@ -0,0 +1,184 @@
+package predicate
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+
```
e-dard: extra newline here.

predicate/parser_test.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/influxdata/influxdb"
+	influxtesting "github.com/influxdata/influxdb/testing"
+	"github.com/influxdata/influxql"
+)
+
+func TestParseNode(t *testing.T) {
```
e-dard: It's not clear to me why we're adding all this support and test coverage for expressions that we would parse and then reject.

predicate/tag_rule.go
```diff
@@ -0,0 +1,80 @@
...
+func (n TagRuleNode) NodeType() string {
+	return tagRuleNodeType
+}
+
+// NodeTypeLiteral convert a TagRuleNode to a nodeTypeLiteral.
+func NodeTypeLiteral(tr TagRuleNode) *datatypes.Node {
+	switch tr.Operator {
+	case influxdb.RegexEqual:
+		fallthrough
+	case influxdb.NotRegexEqual:
```
e-dard: Do we need to supoprt this?

jademcgough:
Agreed with @edd. We can leave the code in a commit if we want to be able to bring it back at some point, but we probably don't need to parse for tokens we don't support.

http/delete_handler.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	if dr.Bucket, err = queryBucket(ctx, r, bucketSvc); err != nil {
+		return nil, err
+	}
+	return dr, nil
+}
+
+type deleteRequest struct {
+	Org       *influxdb.Organization
```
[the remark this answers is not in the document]
kelwang: Yeah, gonna use the cutomized one: deleteRequestDecode

predicate/parser.go
```diff
@@ -0,0 +1,197 @@
...
+func (p *parser) curr() (tok influxql.Token, pos influxql.Pos, lit string) {
+	buf := &p.buf[(p.i-p.n+len(p.buf))%len(p.buf)]
+	return buf.tok, buf.pos, buf.lit
+}
+
+// scanIgnoreWhitespace scans the next non-whitespace token.
+func (p *parser) scanIgnoreWhitespace() (tok influxql.Token, pos influxql.Pos, lit string) {
+	tok, pos, lit = p.scan()
+	if tok == influxql.WS {
+		tok, pos, lit = p.scan()
```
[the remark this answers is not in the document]
kelwang: yeah, the scanner already taking care of that, so no for loop wrap needed: https://github.com/influxdata/influxql/blob/57f403b00b124eb900835c0c944e9b60d848db5e/scanner.go#L140

predicate/parser.go
```diff
@@ -0,0 +1,197 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxql"
+)
+
+type buffer [3]struct {
```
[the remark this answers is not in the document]
kelwang: added comments

PR #15266, opened 2019-09-24
feat(task): add limit function for task concurrency

lyondhill:
The new task executor handles limit's differently then the old executor
instead of front loading limits by creating a runner for every task that might run
the new executor has a large worker pool and queue. This allow's us to have a unlimited
concurrency per task and helps us avoid a back log of task's execution based on a
arbitrary execution limit. This add's the ability to add an optional task execution limit
so a user can still have the advantages of limiting concurrency.

### DIFF
--- a/task/backend/executor/limits.go
+++ b/task/backend/executor/limits.go
@@ -1,0 +1,56 @@
+package executor
+
+import (
+	"context"
+	"sort"
+
+	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/task/options"
+)
+
+// ConcurrencyLimit creates a concurrency limit func that uses the executor to determine
+// if they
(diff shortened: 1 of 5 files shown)


task/backend/executor/limits.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"context"
+	"sort"
+
+	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/task/options"
+)
+
+// ConcurrencyLimit creates a concurrency limit func that uses the executor to determine
+// if they
```
AlirieGray: is this comment finished?

lyondhill: oops haha. I will fix that.

PR #15268, opened 2019-09-24
refactor(tasks): create coordinator for new scheduler/executor

### LINKED ISSUE #15129: Implement NewCoordinator to interact with NewScheduler
We need a Coordinator in OSS influx based on the current one that works with the new Scheduler.  It needs to accept a Scheduler interface in its creation function.  The interface for scheduler is defined in tasks/backend/scheduler/scheduler.go

The new coordinator must implement the `middleware.Coordinator` interface so that we can substitute the old or new coordinator at runtime.

## Configuration

It is important that the new scheduler is opt-in until the feature is tested and production ready. To that end, the OSS `launcher` must provide a configuration option to determine, at runtime, which scheduler to use. It is anticipated that the following section of code will conditionally choose to configure either the existing or new scheduler:

https://github.com/influxdata/influxdb/blob/080943a9f5620c3de5db5a732f350f995cd1ed45/cmd/influxd/launcher/launcher.go#L546-L584

Also note that the repeated calls to

```go
coordinator := coordinator.New(m.logger, m.scheduler)
```

should use a shared coordinator if possible, so that subsequent services can reuse the correct coordinator:

```go
var checkSvc platform.CheckService
{
	// coordinator := coordinator.New(m.logger, m.scheduler)
	checkSvc = middleware.NewCheckService(m.kvService, m.kvService, coordinator)
}
```

There is no opportunity to reuse configuration or setup code between OSS and Cloud as they are significantly different. It is reasonable to expose a command line flag `--feature-new-task-scheduler` to opt-in to the new scheduler in OSS.

AlirieGray:
Closes #15129

This PR implements a new Coordinator called (temporarily) `TaskCoordinator' to work with the new Scheduler currently in progress.

This addition includes several abstractions, including:
1. `SchedulableTask`, a wrapper around the `Task` struct to make it work with the `Schedulable` type expected by the Scheduler.
1. `SchedulableTaskService`, a wrapper around the `TaskService` to provide the API for updating a task with a new `LatestCompleted` value.

This new coordinator currently lives alongside the old coordinator in `task_coordinator.go`. The old code will be removed once the new system has been fully implemented and tested. 

TODO (when scheduler is completed):
 - [ ] check for env variable in launcher.go to decide whether to use old scheduler system or new

This PR also adds helper functions `LatestCompletedTime` and `OffsetDuration` to the Task struct to return Go Time library objects for the string values `LatestCompleted` and `Offset` on the Task object.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/task.go
+++ b/task.go
@@ -60,7 +60,24 @@ func (t *Task) EffectiveCron() string {
 	return ""
 }
 
-// Run is a record created when a run of a task is scheduled.
+// LatestCompletedTime gives the time.Time that the task was last queued to be run in RFC3339 format.
+func (t *Task) LatestCompletedTime() (time.Time, error) {
+	tm := t.LatestCompleted
+	if tm == "" {
+		tm = t.CreatedAt
+	}
+	return time.Parse(time.RFC3339, tm)
+}
+
+// OffsetDuration gives the time.Duration of the Task's Offset property, which represents a delay before execution
+func (t *Task) OffsetDuration() (time.Duration, error) {
+	if t.Offset == "" {
+		return time.Duration(0), nil
+	}
+	return time.ParseDuration(t.Offset)
+}
+
+// Run is a record createId when a run of a task is scheduled.
 type Run struct {
 	ID           ID     `json:"id,omitempty"`
 	TaskID       ID     `json:"taskID"`
--- a/task/backend/coordinator/task_coordinator.go
+++ b/task/backend/coordinator/task_coordinator.go
@@ -1,0 +1,184 @@
+package coordinator
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/task/backend"
+	"github.com/influxdata/influxdb/task/backend/executor"
+	"github.com/influxdata/influxdb/task/backend/middleware"
+	"github.com/influxdata/influxdb/task/backend/scheduler"
+	"go.uber.org/zap"
+)
+
+var _ middleware.Coordinator = (*Coordinator)(nil)
+var _ Executor = (*executor.TaskExecutor)(nil)
+
+// DefaultLimit is the maximum number of tasks that a given taskd server can own
+const DefaultLimit = 1000
+
+// Executor is an abstraction of the task executor with only the functions needed by the coordinator
+type Executor interface {
+	ManualRun(ctx context.Context, id influxdb.ID, runID influxdb.ID) (executor.Promise, error)
+	Cancel(ctx context.Context, runID influxdb.ID) error
+}
+
+// TaskCoordinator (temporary name) is the intermediary between the scheduling/executing system and the rest of the task system
+type TaskCoordinator struct {
+	logger *zap.Logger
+	sch    scheduler.Scheduler
+	ex     Executor
+
+	limit int
+}
+
+type CoordinatorOption func(*TaskCoordinator)
+
+// SchedulableTask is a wrapper around the Task struct, giving it methods to make it compatible with the Scheduler
+type SchedulableTask struct {
+	*influxdb.Task
+}
+
+func (t SchedulableTask) ID() scheduler.ID {
+	return scheduler.ID(t.Task.ID)
+}
+
+// Schedule takes the time a Task is scheduled for and returns a Schedule object
+func (t SchedulableTask) Schedule() scheduler.Schedule {
+	// TODO (al): use new scheduler's NewSchedule method
+	return scheduler.Schedule{}
+}
+
+// Offset returns a time.Duration for the Task's offset property
+func (t SchedulableTask) Offset() time.Duration {
+	offset, _ := t.OffsetDuration()
+	return offset
+}
+
+// LastScheduled parses the task's LatestCompleted value as a Time object
+func (t SchedulableTask) LastScheduled() time.Time {
+	tm, _ := t.LatestCompletedTime()
+	return tm
+}
+
+func WithLimitOpt(i int) CoordinatorOption {
+	return func(c *TaskCoordinator) {
+		c.limit = i
+	}
+}
+
+// NewSchedulableTask transforms an influxdb task to a schedulable task type
+func NewSchedulableTask(task *influxdb.Task) (SchedulableTask, error) {
+	if offset, err := task.OffsetDuration(); offset != time.Duration(0) && err != nil {
+		return SchedulableTask{}, errors.New("could not create schedulable task: offset duration could not be parsed")
+	}
+
+	if _, err := task.LatestCompletedTime(); err != nil {
+		return SchedulableTask{}, errors.New("could not create schedulable task: latest completed time could not be parsed")
+	}
+	t := SchedulableTask{task}
+
+	return t, nil
+}
+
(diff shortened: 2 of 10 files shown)


task/backend/coordinator/task_coordinator.go
```diff
@@ -0,0 +1,185 @@
...
+// Schedule takes the time a Task is scheduled for and returns a Schedule object
+func (t SchedulableTask) Schedule() scheduler.Schedule {
+	// TODO (al): use new scheduler's NewSchedule method
+	return scheduler.Schedule{}
+}
+
+// Offset returns a time.Duration for the Task's offset property
+func (t SchedulableTask) Offset() time.Duration {
+	offset, _ := t.OffsetDuration()
+	return offset // TODO(al): how to handle parsing errors?
```
AlirieGray: the `Offset` and `LastScheduled` methods as defined in the new scheduler do not include a returned error, so I was unclear on how to handle this error. right now I'm just throwing it out but that doesn't seem optimal

task/backend/coordinator/task_coordinator.go
```diff
@@ -0,0 +1,185 @@
...
+type OptionC func(*TaskCoordinator)
+
+// SchedulableTask is a wrapper around the Task struct, giving it methods to make it compatible with the Scheduler
+type SchedulableTask struct {
+	*influxdb.Task
+}
+
+// SchedulableTaskService provides an API to update the LatestScheduled time of a task
+type SchedulableTaskService struct {
+	ts influxdb.TaskService
```
AlirieGray: currently wrapping the whole task service, but would like to just include the method necessary for the coordinator (which is UpdateTask)

task/backend/coordinator/task_coordinator.go
docmerlin: We could add an error to the signature, to allow for errors being returned.

task/backend/coordinator/task_coordinator.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// UpdateLastScheduled uses the task service to store the latest time a task was scheduled to run
+func (c *SchedulableTaskService) UpdateLastScheduled(ctx context.Context, id scheduler.ID, t time.Time) error {
+	s := t.Format(time.RFC3339Nano)
```
lyondhill: I think the latest completed doesn't need nanoseconds.

task.go
```diff
@@ -60,7 +60,17 @@ func (t *Task) EffectiveCron() string {
 	return ""
 }
 
-// Run is a record created when a run of a task is scheduled.
+// LatestCompletedTime gives the time.Time that the task was last queued to be run at in RFC3339Nano format.
+func (t *Task) LatestCompletedTime() (time.Time, error) {
+	return time.Parse(time.RFC3339Nano, t.LatestCompleted)
```
lyondhill: I believe this is a `time.RFC3339` it is set [here](https://github.com/influxdata/influxdb/blob/a17277f27ae0450d6b4565b149c21833c15e67f8/kv/task.go#L1570) based on the runs scheduledAt which is defined [here](https://github.com/influxdata/influxdb/blob/a17277f27ae0450d6b4565b149c21833c15e67f8/kv/task.go#L1289)

task/backend/coordinator/task_coordinator.go
```diff
@@ -0,0 +1,183 @@
...
+	"go.uber.org/zap"
+)
+
+var _ middleware.Coordinator = (*Coordinator)(nil)
+var _ Executor = (*executor.TaskExecutor)(nil)
+var _ scheduler.SchedulableService = (*SchedulableTaskService)(nil)
+
+// Executor is an abstraction of the task executor with only the functions needed by the coordinator
+type Executor interface {
+	ManualRun(ctx context.Context, id scheduler.ID, runID scheduler.ID) (executor.Promise, error)
```
lyondhill: These should use `influxdb.ID`. Why would a communication between the coordinator and the executor need to communicate using an ID thats non standard for the influxdb system? The scheduler only uses its own id because we want it to be able to stand on its own.

task/backend/coordinator/task_coordinator.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// Schedule takes the time a Task is scheduled for and returns a Schedule object
+func (t SchedulableTask) Schedule() scheduler.Schedule {
+	// TODO (al): use new scheduler's NewSchedule method
+	return scheduler.Schedule{}
+}
+
+// Offset returns a time.Duration for the Task's offset property
+func (t SchedulableTask) Offset() (time.Duration, error) {
```
lyondhill: I think the error here should be caught before we attempt to inform the scheduler. If a task gets to the coordinator and it cant read the offset We shouldn't even attempt to communicate with the scheduler. 
What this is doing is pushing an implementation onto the scheduler. The scheduler shouldn't have to be concerned about if it can get the offset it should always have it.

task/backend/coordinator/task_coordinator.go
```diff
@@ -0,0 +1,183 @@
...
+
+	for _, opt := range opts {
+		opt(c)
+	}
+
+	return c
+}
+
+// TaskCreatedC asks the Scheduler to schedule the newly created task
+func (c *TaskCoordinator) TaskCreatedC(ctx context.Context, task *influxdb.Task) error {
```
lyondhill: Im not sure where `TaskCreatedC` comes from or what the last `C` means. I believe your coordinator is supposed to fulfill the interface the middleware is expecting.
https://github.com/influxdata/influxdb/blob/0cc9caa1d424ad0f8c1151feafba28f34175805a/task/backend/middleware/middleware.go#L14-L21

task/backend/coordinator/task_coordinator.go
lyondhill: I added a comment on this. If an error occur's either in the coordinator side or the scheduler side, the coordinator has to react. Why put the burden on the scheduler if the coordinator still has to do the same thing as if it had caught the error itself?

task/backend/coordinator/task_coordinator.go
AlirieGray: good point, I'll fix that

task/backend/coordinator/task_coordinator.go
AlirieGray: The C is just added because since the two files are in the same package, it doesn't like it having the same name. I can move it to the outer directory though and that would fix it.

cmd/influxd/launcher/launcher.go
```diff
@@ -554,7 +554,7 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 		// define the executor and build analytical storage middleware
 		combinedTaskService := taskbackend.NewAnalyticalStorage(m.logger.With(zap.String("service", "task-analytical-store")), m.kvService, m.kvService, pointsWriter, query.QueryServiceBridge{AsyncQueryService: m.queryController})
 		executor := taskexecutor.NewAsyncQueryServiceExecutor(m.logger.With(zap.String("service", "task-executor")), m.queryController, authSvc, combinedTaskService)
-
+		// TODO: check for env variable to decide whether to use old scheduler system or new
```
stuartcarnie: I will create an issue to unify the OSS and IDPE initialization / setup. We'll want to do this before we integrate the new scheduler, so this TODO will probably be different / redundant.

PR #15286, opened 2019-09-26
feat(task): remove preauth

lyondhill:
Remove the preauth from task execution because it can cause problems. Because task's was the last piece to remove preauth, this pr also cleans up the preauth code.


docmerlin:
Needs fixes to pass e2e and CI tests.

Issue #15160, opened 2019-09-16
Add promql transpiler tests 


fchikwekwe:
Tests for the transpiler need to be fixed and merged. 
https://github.com/influxdata/influxdb/compare/master...juliusv:promql-transpiler-e2e-tests

PR #15293, opened 2019-09-27
test(query): add PromQL end-to-end tests in go submodule

affo:
#15160 

Some notes on this:

 - the submodule depends on InfluxDB, but I replaced with the local one, so that we don't have to update deps for it and the CI can detect if someone breaks PromQL tests by making changes to InfluxDB.
 - the input for tests is a line protocol file of 75MB compressed to a `gz` one of 2MB. Can it be ok to commit that one?
 - Added line in the Makefile to be used in CI


affo:
Ok, I see that our CI prevents us from committing binary files. Now what?
Should I commit the 75MB file?
Should I downsample that? 🤔

juliusv:
Overall, nice! Good to see these being back properly.

In my initial e2e tests, I had checked in a Prometheus TSDB (as this was the most efficient way to represent the input data and it avoided having to re-write the data into two DBs), but I guess checking in the InfluxDB line protocol as input data is more desirable in the InfluxDB context, and to have human-readable source data :) I think compressing it would make sense (seems the CI just expects the commit to have an annotation for it, so it might be fine?).

nathanielc:
> seems the CI just expects the commit to have an annotation for it, so it might be fine?

Correct the CI step is to ensure we don't accidentally commit binary data. If we have an explicit reason to do so then its fine. 

@affo Will compression significantly reduce the file size? I am fine if we commit the binary file we just need to explain why we are doing so.

juliusv:
@nathanielc According to the comment at the top, it's 2MB vs. 75MB, so yes.

mark-rushakoff:
Checking in a gzipped file:
- Git won't be able to apply any compression to the object-store representation
- The file will only use 2MB in your working tree and can therefore be checked out quickly

Checking in the raw line protocol:
- Git will internally compress it, and it will potentially use less than 2MB in the object store
- The file will use 75MB in your working tree, so it will take longer to check out than the binary file
- Could potentially have annoying false positives for `git grep`

Neither option is great. Does the file have to be checked in at all? Could the content be generated from a function in Go, rather than read from a file?

juliusv:
So I originally generated the testdata by having a real Prometheus monitor multiple instances of a demo service (https://github.com/juliusv/prometheus_demo_service) for a short while. While not being real production data, this at least gives the data a certain "real-ness" in terms of spread-out scrape timestamps, and maybe also sample values (although the metrics are mostly synthetically generated). With some care all of this could be procedurally generated, though then it'll be harder to manually inspect the test input data.

affo:
@mark-rushakoff as we discussed, I think that input generation would be valuable, but I think we should leave that for a later PR and do that organically and with a well defined tool so that we  can use it  in multiple cases.

Is that ok for you?

mark-rushakoff:
Does it really need to be 75MB of line protocol still? Could you just sample 10% of the lines in the file and regenerate the tests?

I'm sorry to continue arguing on this, but this is adding 2MB+ on every clone of this already large repository (locally, `du -sh .git` reports around 89MB iafter a `git gc --aggressive`).

affo:
@mark-rushakoff absolutely see your point. I am doing that right now.

affo:
@mark-rushakoff @nathanielc @juliusv  Ok, I downsampled taking the last record per series every 2minutes.

Before downsampling:
```
	// no records: 506590
	// mean table size: 351 (stddev 23)
```

After:
```
	// no records: 24264
	// mean table size: 16.8 (stddev 1.1)
```

.circleci/config.yml
```diff
@@ -122,7 +122,7 @@ jobs:
       - run: make checktidy
       - run: GO111MODULE=on go mod vendor # staticcheck looks in vendor for dependencies.
       - run: GO111MODULE=on go install honnef.co/go/tools/cmd/staticcheck # Install staticcheck from the version we specify in go.mod.
-      - run: staticcheck ./...
+      - run: GO111MODULE=on staticcheck ./...
```
affo: CI failed with:

```
-: go build github.com/influxdata/flux/promql: no Go files in 
go build github.com/influxdata/promql/v2: no Go files in 
go build github.com/prometheus/prometheus/pkg/labels: no Go files in 
go build github.com/prometheus/prometheus/promql: no Go files in 
go build github.com/prometheus/prometheus/storage: no Go files in 
go build github.com/prometheus/prometheus/storage/tsdb: no Go files in 
go build github.com/prometheus/tsdb/wal: no Go files in (compile)
```

On this step.

I SSHed into the machine and saw that it worked like this. I can't exactly explain why though...

affo:
@mark-rushakoff @nathanielc this seems now good for review.

query/promql/promqltests/go.mod
```diff
@@ -0,0 +1,26 @@
...
+	github.com/prometheus/prometheus v2.9.2+incompatible
+	github.com/prometheus/tsdb v0.10.0
+	github.com/spf13/cobra v0.0.5 // indirect
+	github.com/willf/bitset v1.1.10 // indirect
+	go.uber.org/zap v1.10.0 // indirect
+	golang.org/x/time v0.0.0-20190921001708-c4c64cad1fd0 // indirect
+	google.golang.org/grpc v1.24.0 // indirect
+)
+
+replace github.com/influxdata/influxdb => ../../../
```
mark-rushakoff: I thought this replace was a mistake at first, but confirmed that [this is an intended use of a multi-module repo](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive).

query/promql/promqltests/README.md
```diff
@@ -0,0 +1,11 @@
+# PromQL End-to-end Tests 
+
+The code under this directory is organized in the go submodule `github.com/influxdata/promqltests`.
+This was made as a go submodule to avoid bringing unnecessary dependencies into InfluxDB.
+This module is _not_ intended to be exported and used by other modules.
```
mark-rushakoff: I don't remember the details of what went wrong last time we tried to use multiple modules in a closed source repo, so I'm a little wary of this.

Maybe this should have import path `github.com/influxdata/influxdb/query/internal/promqltests`, or something similar with `internal`? That way the go tooling will properly enforce that the module can't be used outside of `github.com/influxdata/influxdb/query/...`.

Possibly another option would be [an empty go.mod](https://github.com/golang/go/wiki/Modules#can-an-additional-gomod-exclude-unnecessary-content-do-modules-have-the-equivalent-of-a-gitignore-file
):

> An empty go.mod in a directory will cause that directory and all of its subdirectories to be excluded from the top-level Go module.

I'm not sure if the `make test-promql-e2e` action would behave as intended that way.

query/promql/promqltests/go.mod
affo: This is cool, I did it intentionally because of the reasons I mention above in the PR description. Now that I know there is a "proper" motivation, that's even better :)

query/promql/promqltests/README.md
affo: I see your point with `internal`. I guess that's a good way of prescribing imports.

I think I want to prevent to things:
 1. InfluxDB shouldn't import this --> solved with `internal`
 2. the module shouldn't be used outside.

I think that in the README I was addressing `2`. I don't know if `internal` would be enough. The submodule is now `github.com/influxdata/promqltests`. I'll see if changing to `github.com/influxdata/internal/promqltests` changes things.

affo:
@mark-rushakoff @nathanielc 

I moved every file in the submodule to an internal folder.

A good news we haven't mentioned before is that tests now require ~2 minutes (due to downsampling).

Waiting for your review :)

mark-rushakoff:
Should it be `query/promql/internal/promqltests`, rather than `query/promql/promqltests/internal`? I assume a go.mod rooted under internal means the module won't be visible externally.

Or maybe I've misunderstood the intent of this layout.

affo:
@mark-rushakoff what do you mean with "not visible". The module, as a submodule, is not visible in general if the super module does not specify it as a dependency.

Enclosing every file into the submodule `internal`, makes it so, even if it is specified as a dependency, nothing can be imported from it.

mark-rushakoff:
Maybe I'm misunderstanding internal modules. I thought:

- `github.com/influxdata/influxdb/query/promql` module: the **module** is visible to the world and importable anywhere, but due to it only having an `internal` package, the subpackages can only be imported by `.../query/promql`. This is kind of like having an exported Go type with no exported fields or methods.
- `github.com/influxdata/influxdb/query/promql/internal`: if the *module* had this name, the *module* would be importable from `github.com/influxdata/influxdb/query/promql` only. This is like an unexported Go type.

I'm advocating for the second approach. Am I misunderstanding the use case here? Are `internal` modules a thing? Is an `internal` module acceptable for this test package, or must it be usable outside influxdb?

query/promql/internal/promqltests/tests.go
```diff
@@ -0,0 +1,533 @@
...
+		},
+		{
+			Query:          `avg_over_time(rate(demo_cpu_usage_seconds_total[1m])[2m:10s])`,
+			SkipComparison: "the implementation cannot guarantee completely identical results for subqueries",
+		},
+	}
+)
+
+func init() {
+	TestCases = getTestCases()
```
mark-rushakoff: nit: this doesn't need to be an `init`, this could just be part of the top-level var declaration and assignment. But fine to leave for this PR.

query/promql/internal/promqltests/tests.go
```diff
@@ -0,0 +1,529 @@
...
+	Skip           string
+	ShouldFail     bool
+}
+
+var (
+	InputPath  = "input.txt"
+	Start      = time.Unix(1550767830, 123000000).UTC()
+	End        = time.Unix(1550767900, 321000000).UTC()
+	Resolution = 10 * time.Second
+	TestCases  = getTestCases()
```
affo: @mark-rushakoff fixed 👍

PR #15295, opened 2019-09-27
fix(http): Ensures users are created with an active status

### LINKED ISSUE #15274: GET http://localhost:9999/api/v2/me returns empty user status
In latest nightly InfluxDB 2.0 build is broken /api/v2/me endpoint.

```json
{
  "links": {
    "logs": "/api/v2/users/04863f56cb5d1000/logs",
    "self": "/api/v2/users/04863f56cb5d1000"
  },
  "id": "04863f56cb5d1000",
  "name": "my-user",
  "status": ""
}
```
Allowed values for status is enum "active" , "inactive".

dearyhud:
Closes #15274

Adds `active` as the default status for the user response.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v2.0.0-alpha.19 [unreleased]
+### Bug Fixes
+1. [15295](https://github.com/influxdata/influxdb/pull/15295): Ensures users are created with an active status 
+
 ## v2.0.0-alpha.18 [2019-09-26]
 
 ### Features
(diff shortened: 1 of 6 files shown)


http/user_service.go
```diff
@@ -323,6 +323,10 @@ type userResponse struct {
 }
 
 func newUserResponse(u *influxdb.User) *userResponse {
+	if u.Status == "" {
```
kelwang: This should be removed. We shall not fake any response.

Issue #15265, opened 2019-09-24
Proposal: Variable Memory Limits for Queries


jsternberg:
Variable Memory Limits
----------------------

### Problem

The Flux engine provides support for tracking the memory usage of each query and provides the ability to restrict the amount of memory that a query may use before it gets killed.

This principle is a cornerstone for how we plan to allow for multi-tenant functionality and long term scalability of the system.

On the other hand, at the moment, we are not reaching that goal. The reason for this is because our memory limits are very simplistic and inelastic. They are unable to adapt to a living system because we only have hard limits and those hard limits dictate a lot about how we set the memory limits.

When calculating how many queries are able to run concurrently, we have to consider a few aspects.

1.	How much memory is each query allowed to use?
2.	How much overhead of Flux runtime memory will be used by the Go runtime and is thus untracked?
3.	How many queries are allowed to execute at the same time?

The first and third are the most important. Because our memory limits are set at the beginning of the runtime and never change, we have to be certain that the amount of memory that the query will use is actually available for use at the time the query executes. This ends up limiting the amount of queries that can be run concurrently because we need to ensure that `N * C` amount of memory is used with `N` being the amount of memory per query and `C` being the concurrency. The value of `N` ends up getting set to a value that we are certain close to 100% of queries will be able to pass through regardless of how much memory may actually get used by the query. This greatly limits the concurrency of the system because we run into one of two situations:

1.	We are unable to feel safe that the memory limits will prevent OOMs because the concurrency multiplied by the memory per query exceeds the memory of the container running the query.
2.	We are unable to set the concurrency to a higher value and we both waste CPU that we are paying for and waste memory that we are paying for because most queries will never use the amount of memory reserved for them.

For this reason, we need a way to allow for memory limits to be changed during the execution of a query. The intention behind this is to be able to reserve a certain amount of memory for each query. This number will be a value that the 99th percentile of queries will require. That means that 99% of queries will only ever use their initial allocation of memory and will never have to request more. The other 1% of queries will still be able to succeed, but may fail if the system is memory starved.

### Proposal

The proposal is to modify the `memory.Allocator` to take an optional `memory.Manager` interface. This interface will be used by the `memory.Allocator` to request more memory so it can raise its own `Limit`. The `Limit` that is set in the `memory.Allocator` will be the currently reserved memory.

```go
package memory

type Manager interface {
    // RequestMemory will request that the given amount of memory
    // be reserved for the caller. The manager will return the number
[...]

PR #15300, opened 2019-09-30
feat(query/control): include variable memory limits for the controller

jsternberg:
The controller now supports setting an initial memory limit and setting
a maximum amount of memory that the controller may use separately from
the memory quota per query and the concurrency quota.

This allows the controller to increase the concurrency quota to a larger
number while setting the maximum amount of memory to a lower amount than
would be required for all queries to use 100% of their allowable memory.

Functionally, this means that a query will have a soft limit for an
initial memory byte quota that a query is guaranteed to have, a shared
pool that it is allowed access to in the case it uses more, and a hard
limit that no query may exceed to prevent runaway queries from taking
over the entire pool.

This change is completely backwards compatible with older configurations
as the new options will default to values that mimic the old behavior
where a query is allocated the full amount of its memory quota and the
maximum amount of memory is based on the concurrency quota and this
maximum memory quota.

In addition to the above, this also fixes a bug in the controller that
allowed it to run more than its concurrency as executing queries. This
happened when the results had finished being sent by the executor, but
the query had not yet been read and/or serialized. The executor would be
freed up and would take the next query even though the previous query
hadn't yet been finalized with `Done()`.

#15265


query/control/controller.go
```diff
@@ -87,23 +100,53 @@ type Config struct {
 	ExecutorDependencies []flux.Dependency
 }
 
+// complete will fill in the defaults, validate the configuration, and
```
fchikwekwe: Thanks for the descriptive commenting.

query/control/controller.go
```diff
@@ -87,23 +100,53 @@ type Config struct {
...
 	if c.ConcurrencyQuota <= 0 {
 		return errors.New("ConcurrencyQuota must be positive")
 	}
 	if c.MemoryBytesQuotaPerQuery <= 0 {
 		return errors.New("MemoryBytesQuotaPerQuery must be positive")
 	}
+	if c.InitialMemoryBytesQuotaPerQuery <= 0 {
+		return errors.New("InitialMemoryBytesQuotaPerQuery must be positive")
+	}
+	if c.MaxMemoryBytes <= 0 {
```
fchikwekwe: Looks like there may be an inconsistency between line 130 and line 139? They are checking the same value.

jsternberg: Fixed. It seems like I accidentally made this same check twice.

query/control/memory.go
```diff
@@ -0,0 +1,132 @@
...
+// RequestMemory will determine if the query can be given more memory
+// when it is requested.
+//
+// Note: This function accesses the memoryManager whose attributes
+// may be modified concurrently. Atomic operations are used to keep
+// it lockless. The data associated with this specific query are only
+// invoked from within a lock so they are safe to modify.
+// Second Note: The errors here are discarded anyway so don't worry
+// too much about the specific message or structure.
+func (q *queryMemoryManager) RequestMemory(want int64) (got int64, err error) {
```
nathanielc: Does this method have a test case? I don't see where RequestMemory is used. Is that planned as a separate PR?

nathanielc: I see its used on the Flux side of things https://github.com/influxdata/flux/pull/1922/files#diff-576d3093e5f774162ca06b1345b6ed07R191

PR #15332, opened 2019-10-07
refactor(ui): global overlays

alexpaxton:
**Goals:**
- Enable more experimentation of UX flows by allowing increased flexibility in accessing overlays
- Hypothesis: Allowing access to overlays from anywhere will make the UI easier to use because the user doesn't need to keep track of the access path for each overlay

**Problem:**
- Overlays must be tied to a specific route
- Accessing an overlay from multiple places means defining a specific route for each one
- Making an overlay accessible from any page is clunky
- Some overlays require props from their parent
- Not all overlays are accessible via the router

**Solution:**
- Place a component within `App.tsx` that listens for query params in the URL and displays overlays based on that
  - example: `8080/orgs/0349f0efc0d46000/data-explorer?overlay=create-bucket`
  - This means any overlay can be opened from any page
- Create a component and utility for opening overlays
- Refactor overlay components so they don't inherit any props from their parent

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -184,6 +184,7 @@
     "papaparse": "^4.6.3",
     "prop-types": "^15.6.1",
     "qs": "^6.5.2",
+    "query-string": "^6.8.3",
     "react": "^16.8.2",
     "react-codemirror2": "^4.2.1",
     "react-copy-to-clipboard": "^5.0.1",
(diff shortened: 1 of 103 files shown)


ebb-tide:
I'm feeling a little nervous about these changes, since other than props from parents, overlays rely on parent components to fetch resources from BE and load them in to redux. It might actually be better for overlays to fetch all the resources they need, but I worry about doing all of them at once. Could we provide the infrastructure for rendering overlays like this, but do the work one at a time?

PR #15343, opened 2019-10-08
fix(logging): blacklist endpoints with sensitive data from logging body

jsteenb2:
blacklists endpoints with sensitive data from having the body logged

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -1,0 +1,153 @@
+package http
+
+import (
+	"bytes"
+	"errors"
+	"io"
+	"net/http"
+	"path"
+	"strings"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+func LoggingMW(logger *zap.Logger) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		fn := func(w http.ResponseWriter, r *http.Request) {
+			srw := &statusResponseWriter{
+				ResponseWriter: w,
+			}
+
+			var buf bytes.Buffer
+			r.Body = &bodyEchoer{
+				rc:    r.Body,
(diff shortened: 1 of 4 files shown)


http/middleware.go
```diff
@@ -0,0 +1,102 @@
...
+		return http.HandlerFunc(fn)
+	}
+}
+
+type isValidMethodFn func(method string) bool
+
+var (
+	ignoreMethod = func(string) bool { return true }
+
+	blacklistEndpoints = map[string]isValidMethodFn{
```
jademcgough: Nice pattern 👍

Can we also add write and query endpoints to the blacklist? They get hit a lot, and we really don't need to log all those requests.

jsteenb2: great point, I'll definitely add that in 👍

jsteenb2: @jademcgough pushed up the changes 👍

jsteenb2: @jademcgough I went through more of the request and found a few more that would have secrets. The org endpoints I totally missed.

jademcgough: @jsteenb2 ah right, I was pretty sure there was something I was forgetting yesterday, it was the org secrets path :)

jsteenb2:
@jademcgough totally agree. I started down that route to do that but it was getting a bit hacky matching on strings. I ended up punting on it b/c the tests were super brittle. I went back and added it back and I'm glad I did! I had missed the `:` prefixes for wildcard matches which would have have all those endpoints body printed. I know have tests for the entire blacklist as well as some base level stuff.  Thanks again for beating the test drum 🤘 !

PR #15358, opened 2019-10-09
fix: Place resource fetch logic in overlay

ebb-tide:
Closes # [4893](https://github.com/influxdata/idpe/issues/4893)

Resource fetching for overlays is a problem that exists in many places in the UI. 
Currently a lot of our overlays do not fetch the resources they need, instead they rely on parent components to fetch resources and place in to redux state. They also do not warn user if the resource they need can not be found in redux state. 
I'm suggesting this PR as a general pattern for data fetching and errors in overlays. 
Would love to hear concerns / feedback

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/src/buckets/components/UpdateBucketOverlay.tsx
+++ b/ui/src/buckets/components/UpdateBucketOverlay.tsx
@@ -1,153 +1,159 @@
 // Libraries
-import React, {PureComponent, ChangeEvent} from 'react'
+import React, {
+  FunctionComponent,
+  useEffect,
+  useState,
+  ChangeEvent,
+  FormEvent,
+} from 'react'
 import {withRouter, WithRouterProps} from 'react-router'
 import {connect} from 'react-redux'
+import {get} from 'lodash'
 
 // Components
-import {Overlay} from '@influxdata/clockface'
+import {
+  Overlay,
+  RemoteDataState,
+  SpinnerContainer,
+  TechnoSpinner,
+} from '@influxdata/clockface'
 import BucketOverlayForm from 'src/buckets/components/BucketOverlayForm'
 
 // Actions
 import {updateBucket} from 'src/buckets/actions'
+import {notify} from 'src/shared/actions/notifications'
+
+// APIs
+import * as api from 'src/client'
 
 // Constants
 import {DEFAULT_SECONDS} from 'src/buckets/components/Retention'
+import {getBucketFailed} from 'src/shared/copy/notifications'
 
-// Types
-import {AppState, Bucket} from 'src/types'
-
-interface State {
-  bucket: Bucket
-  ruleType: 'expire'
-}
-
-interface StateProps {
-  bucket: Bucket
-}
+//Types
+import {Bucket} from 'src/types'
 
 interface DispatchProps {
   onUpdateBucket: typeof updateBucket
+  onNotify: typeof notify
 }
 
-type Props = StateProps & DispatchProps & WithRouterProps
+type Props = DispatchProps & WithRouterProps
 
-class UpdateBucketOverlay extends PureComponent<Props, State> {
-  constructor(props) {
-    super(props)
+const UpdateBucketOverlay: FunctionComponent<Props> = ({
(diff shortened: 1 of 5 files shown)


ui/src/buckets/components/UpdateBucketOverlay.tsx
```diff
@@ -1,153 +1,159 @@
...
+      if (resp.status !== 200) {
+        onNotify(getBucketFailed(bucketID, resp.data.message))
+        handleClose()
+        return
+      }
+      setBucketDraft(resp.data)
 
-  private ruleType = (bucket: Bucket): 'expire' => {
-    const rule = bucket.retentionRules.find(r => r.type === 'expire')
+      const rules = get(resp.data, 'retentionRules', [])
```
drdelambre: it would be cool if this was also in an async function that awaited the bucket data. i know they're super small functions that reuse the same data, they just feel chainable and self contained

ui/src/buckets/components/UpdateBucketOverlay.tsx
```diff
@@ -1,153 +1,159 @@
...
-  private get retentionSeconds(): number {
-    const rule = this.state.bucket.retentionRules.find(r => r.type === 'expire')
+  const [retentionSelection, setRetentionSelection] = useState(DEFAULT_SECONDS)
 
-    if (!rule) {
-      return DEFAULT_SECONDS
-    }
+  useEffect(() => {
+    const fetchBucket = async () => {
+      const resp = await api.getBucket({bucketID})
```
drdelambre: it seems like the friction to this pattern would be that this function isn't cached. for overlays, there is only going to be one per page at a time, but when building out a reusable pattern of `component owns its own state`, certain affordances need to be made around cache invalidation/expiration. not going to use this PR as a springboard for that discussion, just something to start thinking about (multiple components using the same api transform within their useEffect lifecycles)

hoorayimhelping:
i love where your head is at with respect to resource fetching and resource dependency for overlays - it should be clear how an overlay gets the data it needs to render itself, and there should be a strong interface for it that the overlay and the party supplying the data need to follow (even if that party is the overlay itself). 👍

i think it's rad that this makes the overlay a little self contained nugget that knows how to get its own data and then render it. it makes reasoning about an overlay in isolation much simpler. 

the main concern i have is this: overlays are usually shown in the context of a container or a page (i'm obviously unfamiliar with the app at the moment and so i don't know all the patterns and conventions, so if what i'm saying is wrong, please excuse me). we usually display overlays that pertain to the page we're using - they aren't usually shown in isolation, and when they are, they're usually fed simple data (like an error notification).
with this pattern, clicking a bucket launches a request that fetches some data about the bucket. closing the overly and clicking the same bucket again launches the same request for the same data. it's small in this case and probably not a huge deal. and in the context of overlays, there probably isn't a ton of data being displayed. but when used as a general pattern, there's probably a lot of unknowns about how these data fetches might look, which gives me a low level amount of anxiety from time to time. 

re-reading the description, the problem it sounds like you're trying to solve is that containers and overlays aren't communicating well, and that storing the data an overlay uses in redux might not be the best place for it to live? Is that a good description of the problem?

ebb-tide:
Thanks for commenting @hoorayimhelping !
I think the problem I'm solving is that we have begun to, and in the context of what @alexpaxton wants to do with: https://github.com/influxdata/influxdb/pull/15368 will be going further in the direction of overlays being decoupled from the containers that are rendering them. 

In the past, they had been rendered directly by the containers and their visibility was controlled in state, so it was not possible for a user to load an overlay with a resource that didn't exist. Now that they are rendered through the router they can, and none of our overlays handle this situation gracefully. 

You are right that we might be fetching resources too frequently. I don't know how to achieve decoupling and not fetch resources too frequently. @drdelambre was suggesting caching API responses?

hoorayimhelping:
thanks @ebb-tide!

what do we find desirable about having overlays decoupled from their containers? i think i'm missing a common ground everyone is working from :)

PR #15368, opened 2019-10-09
refactor(overlays): proposed new pattern for handling overlays

alexpaxton:
# Global Overlays

### Why

- Current overlay system has some pain points:
  - Some overlays (such as create org) need to be accessed from anywhere but implementing that would require nesting a route under every page in the UI which is overly repetitive and complex
  - Some overlays depend on specific resources having been fetched by their parent

### Goals (or requirements)

- Create a system that empowers design to experiment with userflows with minimal dev work
- Ability to display any overlay from anywhere easily
- All overlays should be rendered at the same point in the DOM tree
- Overlays can receive additional information if necessary
- Overlays can be scoped to certain routes
- Overlays usable independent of context
  - Responsible for fetching required resources
  - No props inherited from parents

### Proposed Solution

- Create a router/link system specifically for overlays
- Use `searchParams` to allow overlays to operate independently of route
  - `OverlayRouter` exists within `App.tsx` putting it close to the top of the DOM which reduces z-index problems and ensures all overlays are rendered consistently
  - `OverlayLink` requires only 1 prop and doesn't render any elements aside from its child, ensuring the DOM tree is preserved and ease of use is simple
- Pass in `resourceID` to `OverlayLink` instead of inferring  context by checking the route for specific `queryParams` 

### Proof of Concept

In this PR I have refactored the Cell Note Editor overlay because:
- The same overlay is used for creation & editing, so it can show off how the `resourceID` prop works
- The overlay isn't really useful outside of dashboards and requires a dashboard ID to work, so it's an example of how we could handle scoping overlays
- Notice the existing search params are preserved when you click `Add Note` (from the dashboard header) or `Edit Note` (from a cell)

Screen Shot 2019-10-09 at 11 55 05 AM

Here's how I propose handling "limited access" of overlays in case someone tries to access one via the URL. In this case the overlay is looking for a `:dashboardID` query param from the URL and does not see it, indicating it is being accessed outside of a dashboard.

### Desired Feedback

- Hence the `Router`/`Link` nomenclature, this feels a bit like re-inventing the router. I wasn't sure how to have 2 independent routers on top of each other, so this was my way of handling that. Open to suggestions here
- Allowing overlays to be accessed from many places means we probably need to re-think how we handle resource fetching app-wide. If we were to move forward with this pattern without addressing resource fetching it would mean wrapping every overlay in a resource fetcher.
- Is there a better way to "scope" overlays while maintaining a "global access" pattern underneath?
- In addition the the last point, can there be a "overlay boundary" wrapper that displays a standardized UI to users who attempt to access an overlay out of scope?
- How would this pattern affect unit and integration tests?
  - I think some of the e2e tests set a URL in order to run tests
- Are there any security issues by having overlays accessible in this way?

- [ ] Rebased/mergeable
- [ ] Tests pass


alexpaxton:
I left a bunch of inline comments to explain in more specifics my approach

alexpaxton:
Connecting this PR to https://github.com/influxdata/influxdb/pull/15358 because relavance

drdelambre:
I feel like you're close to something, but kind of threw the baby out with the bath water. Totally onboard with overlays having a self contained state and injected into a single central point on the dom (probably root), but i dont understand why query params should kill deeplinks. You're missing a central piece to a pattern that is normally called some variation of the 'controller' pattern, and it should have it's own state of [ is it open?, which overlay am i showing?, what do i do when i close? ]. The routes that deeplink are then not used so much to determine what is rendered as to populate the controller's state. for example: /tasks/:id/create might show the exact same dom as /tasks, but the rest of the url is used to populate the controller (or even have a functional component that throws down some state on useEffect(()=>{ read the url }, []) but renders false). and you can use any event / state to populate that controller, from button clicks to server events.

hoorayimhelping:
thanks for starting this conversation and for laying out what you're going for! i think the example helps me understand things a lot better.

i am 100% for using the url to facilitate modifying the state the app to render different things (i.e. pasting &overlay=edit-note). being able to use links that open up overlays opens up all sorts of cool avenues for interactive tutorials and api docs that let you click to put your app into the state you want. i'm on board with helping make this happen, regardless of the overlay discussion.

how many overlays would need to be global vs needing to be scoped to the page/route/app section/feature they're in? is it like 10/90? 50/50? 20/80? usually an overlay is meant to display some kind of contextual data about the page its on, and sometimes system level functions need to display system level information. it seems like global overlays should be kind of called out as special global overlays that need global state and so behave differently.

if global overlays were the exception, we could remove the concept of scoping because all overlays would be scoped to the page/route they're on just by existing and not being a special global overlay. i don't think there's anything wrong with a tight coupling of an overlay to the page it's a part of, as long as the dependencies between overlays and pages are clear (which it sounds like they may not be at the moment)

i'm not sure i necessarily agree or disagree with overlays needing to be at the root of the page specifically, but i totally agree that we should think about the z-index of overlays and how it interacts with the rest of the page and ease our pain there.

>Overlays usable independent of context

this is one i'm confused about. do we not really have a generic overlay that just renders a box on page in a way that makes it look like the box is floating above the content that you can pass content to as props?

(sorry for any ignorance about how we do things here)

alexpaxton:
@hoorayimhelping after talking to @drdelambre a bit about this, my conclusion is that we don't need every overlay to be globally accessible. I'm going to have overlays handled by redux instead of  search params in the URL. The idea is that overlays that need a specific route (or deep link) can still be accessed via the router, but under the hood it will be a component that just fires an action to redux. Hope that makes sense

drdelambre:
I think I can address the last three paragraphs. Currently an overlay is scoped to a route, only to a route, and has a top level dom depth. This PR is exploring how to allow any component [on any route, at any dom depth] access to controlling an arbitrary overlay. A button on route `/user/history/buckets` should be allowed to open the same overlay as a route change event opening `/bucket/:id/edit`, but without polluting the user's history or tightly coupling that button's view requirements with the page it sits on. My suggestion is to couple it to a third party overlay provider so that the overlay/initiator/route entanglement gets decoupled through overlay state change events, but that has the glaring issue of allowing:
```
<OverlayController>
    <UnusedOverlayOne />
    <NeglectedCodeTwo />
    <UserErrorTimeBombThree />
    <BloatedBundleSizeFour />
    <TheOneOverlayThatIsActuallyUsed />
</OverlayController>
```
Wiring up a similar solution appears to be an inevitable step before adding the optimization of dynamically registering overlays from the component that requires it, but we can solve that when the problem shows up.

hoorayimhelping:
Thanks for the explanation, y'all!

One question I have @drdelambre:

> but without polluting the user's history...

Can you explain "polluting the user's history" a little more? What history is being polluted, and with what?

>tightly coupling that button's view requirements with the page it sits on.

Is there anything stopping us from having overlays just accept props? Do they need access to the entire redux store? If overlays had a strong interface via props, there wouldn't be any coupling to the page they're on - the coupling would be to the interface.

drdelambre:
for the first one, i like to think of history pollution as "back button bugs". the classic example looks like:
`user clicks link -> route changes -> overlay is open`
vs
`user deeplinks to route -> overlay is open`
what happens when the user clicks the back button in both cases? does one case leave the user in a weird state? pollution is an extension of trying to navigate all the permutations of route + overlay combinations, pushing imaginary state through the history api, and ending up with a back button that does.... who knows. There are product cases when we want to allow deeplinking and to guard against it dropping the user into a black hole, but decoupling the overlays from routes means that we only have to consider that deliberately defined small subset.

for the second, there's an intent to prevent [as @alexpaxton eloquently puts it] _prop drilling_, which is the tendency to pass props completely through components just to satisfy their childrens requirements. the more boring/bs way of saying this is that there exists a tight intergenerational coupling within a component's hierarchy that is hurting generalized view composition efforts. Props means context, and we're trying to remove the amount of context that needs to be carried around with components to allow them to be more easily swapped in and out, conversely that components should define their own context (and one step further... state and view code collocation!). i believe this echos @ebb-tide's sentiment in #15358, but i feel like on walking a fine line of speaking for others vs regurgitating what i think i heard.

ui/src/dashboards/components/NoteEditorOverlay.tsx
```diff
@@ -91,53 +87,56 @@ class NoteEditorOverlay extends PureComponent<Props, State> {
...
+        </Overlay.Footer>
+      </Overlay.Container>
     )
   }
 
+  private handleDismiss = (): void => {
+    const {onDismiss, router} = this.props
+
+    onDismiss()
+    router.goBack()
```
drdelambre: what does this do?

alexpaxton:
@ebb-tide @drdelambre @hoorayimhelping 

I refactored all the code in this PR. Overall it feels pretty clean and flexible. The tl;dr of this is that redux is the source of truth for overlays. For overlays that require a deep link we use the HOC to handle the router talking to redux. For global overlays we can fire an action to show/hide it. I'll break down some of the elements:

1. `RouteOverlay` HOC
  - This fires the `showOverlay` action with the `overlayID` specified
  - All the query params from the URL are automatically sent to redux
  - See it in use handling the `Cell Note Editor`
2. `OverlayController`
  - This sits at the top of the UI and listens to the `overlay` key in redux
  - Shows only 1 overlay at a time based on the `overlayID` key within `overlay`
3. Overlay State in Redux
  - Overlays can be displayed via the `showOverlay` action
  - Overlays can be hidden via the `dismissOverlay` action
4. Individual Overlays
  - Use the standard `onDismiss` action
  - Within `render` the outermost component should be `<Overlay.Container />` not `<Overlay />`
  - If the overlay is triggered by the router, you will need to fire `router.goBack()` in addition to `onDismiss()` to keep the history clean

ui/src/dashboards/components/NoteEditorOverlay.tsx
alexpaxton: `onDismiss()` just hides the overlay by removing stuff from redux
`router.goBack()` returns the router to the previous state for cleanliness purposes

Wasn't sure if these should be combined or not, since not all overlays are tied to the router

drdelambre:
this is super cool! thanks for entertaining the abstraction. it feels cleaner. i can't decide if the polymorphic nature of pumping the route params into a general store is a good thing (for flexibility) or a bad thing (for complexity) but any potential problems will show up soon enough. lgtm.

ui/src/dashboards/components/NoteEditorOverlay.tsx
```diff
@@ -92,34 +87,56 @@ class NoteEditorOverlay extends PureComponent<Props, State> {
...
+    const {dashboardID} = this.props
+
+    if (!dashboardID) {
+      return (
+        <Overlay.Container maxWidth={360}>
+          <Overlay.Header title="Oh no!" onDismiss={this.handleDismiss} />
+          <Overlay.Body>
+            <h5>
+              This page does not allow creation or editing of notes, better head{' '}
+              to a dashboard to do that.
```
hoorayimhelping: Do you know under what circumstances a user will see error message? Is this a protection against engineers putting overlays in the wrong place after we move forward with the new system?

ui/src/overlays/components/OverlayController.tsx
```diff
@@ -0,0 +1,31 @@
...
+  let activeOverlay = <></>
+
+  switch (overlayID) {
+    case 'add-note':
+    case 'edit-note':
+      activeOverlay = <NoteEditorOverlay />
+      break
+  }
+
+  return <Overlay visible={!!overlayID}>{activeOverlay}</Overlay>
```
hoorayimhelping: what do you think about having a boolean representing `visibility` that is explicitly set, rather than relying on a string cast to a boolean? it's more explicit this way - only when the overlayID is `add-note` or `edit-note` and the `activeOverlay` is an actual component will the overlay be visible. currently, if the overlayID is `delete-note` for example, `visible={!!overlayID}` will evaluate to true, even though the `activeOverlay` is a react fragment, not the note editor overlay.

```ts
  let visible = false
  switch (overlayID) {
    case 'add-note':
    case 'edit-note':
      activeOverlay = <NoteEditorOverlay />
      visible = true
      break
  }

  return <Overlay visible={visible}>{activeOverlay}</Overlay>
```

alexpaxton: ahh that is much cleaner

ui/src/dashboards/components/NoteEditorOverlay.tsx
alexpaxton: Now that you mention it, it is a vestige of this overlay being displayed potentially everywhere. Now that it is only accessible via a single route I don't think it's an issue. I did want to guard against the lack of params since it needs them to render correctly, but maybe there's a more generic & reusable way to handle it

PR #15369, opened 2019-10-09
feat(tasks): deactivate task on unrecoverable error

AlirieGray:
This PR adds a function `IsUnrecoverable` that determines if an error is not recoverable by the system (i.e. requires user intervention to fix) and deactivates the task with a log message if so. Note that this change is only implemented in the new executor system, and so it will not be in use until the new system is switched over.

This PR also generalizes the worker `finish` and metrics `LogError` methods to accept a generic error type, and catches unspecified errors as an "unknown" type.

Future work:
1. Notify the user that their task was deactivated in a more apparent way than a log message (will require notification bucket work)
1. Retry tasks that can be recovered (for example, timeout errors)
1. Record this feature to be documented (ex. in Changelog) when the new executor system is put in place

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/task/backend/check_task_error.go
+++ b/task/backend/check_task_error.go
@@ -1,0 +1,26 @@
+package backend
+
+import (
+	"strings"
+)
+
+// IsUnrecoverable takes in an error and determines if it is permanent (requiring user intervention to fix)
+func IsUnrecoverable(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	errString := err.Error()
+
+	// missing bucket requires user intervention to resolve
+	if strings.Contains(errString, "could not find bucket") {
+		return true
+	}
+
+	// unparseable Flux requires user intervention to resolve
+	if strings.Contains(errString, "could not parse Flux script") {
+		return true
+	}
+
+	return false
+}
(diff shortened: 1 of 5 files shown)


task/backend/check_task_error.go
```diff
@@ -0,0 +1,29 @@
...
+
+type ErrorType int
+
+const (
+	PermanentErr ErrorType = iota
+	InternalErr
+)
+
+// CheckTaskError takes in an error and determines if it is transient, permanent, or internal
+func CheckTaskError(err error) ErrorType {
```
docmerlin: Would be useful to add a comment saying that err must not be nil.

task/backend/check_task_error.go
```diff
@@ -0,0 +1,22 @@
+package backend
+
+import (
+	"strings"
+)
+
+// IsUnrecoverable takes in an error and determines if it is permanent (requiring user intervention to fix)
+func IsUnrecoverable(err error) bool {
+	errString := err.Error()
```
docmerlin: Can you return false early if err==nil?

Issue #14822, opened 2019-08-27
Influx CLI doesn't list members of org


gitirabassi:
`influx org members list --id=<org id>` and `influx org members list --name=<org name>` return empty list

PR #15372, opened 2019-10-09
fix(cmd/influx): fix list org members

kelwang:
Closes https://github.com/influxdata/influxdb/issues/14822

fix the tcp request to list members. use async tcp request for each user.

- [x] Rebased/mergeable
- [ ] Tests pass


cmd/influx/user_resource_mapping.go
```diff
@@ -3,10 +3,74 @@ package main
...
+	}
+	userSVC, err := newUserService(flgs)
+	if err != nil {
+		return fmt.Errorf("failed to initialize users service client: %v", err)
+	}
+	urs := make([]*influxdb.User, len(mps))
+	ursC := make(chan struct {
+		User  *influxdb.User
+		Index int
+	}, len(mps))
```
jsteenb2: what's the motivation for making this a buffered chan?

cmd/influx/user_resource_mapping.go
```diff
@@ -3,10 +3,74 @@ package main
...
+	errC := make(chan error)
+	for k, v := range mps {
+		go func(k int, v *influxdb.UserResourceMapping) {
+			usr, err := userSVC.FindUserByID(ctx, v.UserID)
```
jsteenb2: just out of curiosity, is it possible to provide a `List` operation for the user service? then let that manage the concurrency? I'm not sure if we can obtain a list in one call or not, just curious 🤷‍♂

cmd/influx/user_resource_mapping.go
```diff
@@ -3,10 +3,74 @@ package main
...
+				Index: k,
+			}
+		}(k, v)
+	}
+	for i := 0; i < len(mps); i++ {
+		select {
+		case err := <-errC:
+			return err
+		case item := <-ursC:
+			urs[item.Index] = item.User
```
jsteenb2: can we add `<-ctx.Done` as well? if there is ever a panic in the `usersSVC.FindUserByID`, then this will block forever potentially. Not sure if that is a ticker, maybe give it 10 seconds to complete or something 🤷‍♂

http/user_resource_mapping_service.go
```diff
@@ -274,30 +273,16 @@ func decodeDeleteMemberRequest(ctx context.Context, r *http.Request) (*deleteMem
 
 // FindUserResourceMappings returns the user resource mappings
 func (s *UserResourceMappingService) FindUserResourceMappings(ctx context.Context, filter influxdb.UserResourceMappingFilter, opt ...influxdb.FindOptions) ([]*influxdb.UserResourceMapping, int, error) {
-	url, err := NewURL(s.Addr, s.basePath)
+	url, err := NewURL(s.Addr, resourceIDPath(filter.ResourceType, filter.ResourceID, string(filter.UserType)+"s"))
```
jsteenb2: the `+"s"` piece here is interesting, why is this needed here?

kelwang: i think it's just some old code. We could make another issue to clean them up

cmd/influx/user_resource_mapping.go
kelwang: yeah, that's a good idea

cmd/influx/user_resource_mapping.go
kelwang: Yeah, I initially thought about that as well. But that requires API schema change.

cmd/influx/user_resource_mapping.go
kelwang: added another ticket to wrap up the timeout: https://github.com/influxdata/influxdb/issues/15375

cmd/influx/organization.go
```diff
@@ -356,6 +334,7 @@ type OrganizationMembersAddFlags struct {
 var organizationMembersAddFlags OrganizationMembersAddFlags
 
 func organizationMembersAddF(cmd *cobra.Command, args []string) error {
+	ctx := context.Background()
```
jsteenb2: nit, can we move this closer to where its being used, say like 364? keep it close, line of sight 🤷‍♂

cmd/influx/organization.go
```diff
@@ -428,6 +407,7 @@ type OrganizationMembersRemoveFlags struct {
 var organizationMembersRemoveFlags OrganizationMembersRemoveFlags
 
 func organizationMembersRemoveF(cmd *cobra.Command, args []string) error {
+	ctx := context.Background()
```
jsteenb2: nit, same with moving this `ctx` to where its being used?

cmd/influx/user_resource_mapping.go
```diff
@@ -3,29 +3,96 @@ package main
...
+				Index int
+			}{
+				User:  usr,
+				Index: k,
+			}
+		}(k, v)
+	}
+	for i := 0; i < len(mps); i++ {
+		select {
+		case <-ctx.Done():
```
jsteenb2: this done won't get called until the background context is canceled yah? Perhaps a `context.WithTimeout(ctx, 10*time.second)` or something would suffice here?  You could also use `time.After(10*time.Second)`, or both depending on how you want to signal cancelation from the OS

cmd/influx/user_resource_mapping.go
```diff
@@ -3,29 +3,96 @@ package main
...
+		return fmt.Errorf("failed to initialize users service client: %v", err)
+	}
+	urs := make([]*influxdb.User, len(mps))
+	ursC := make(chan struct {
+		User  *influxdb.User
+		Index int
+	})
+	errC := make(chan error)
+	for k, v := range mps {
+		go func(k int, v *influxdb.UserResourceMapping) {
```
jsteenb2: thsi could create a lot of go routines if the number of `mps` is big, perhaps the semaphore here to govern it?  Something like the following would do the trick to limit the number of outgoing requests

```go
sem := make(chan struct{}, maxActive)
for k, v := range mps {
       sem <- struct{}{}
	go func(k int, v *influxdb.UserResourceMapping) {
             defer func() { <-sem }()
             ... other stuffs
```

cmd/influx/user_resource_mapping.go
kelwang: yeah, i want to handle the timeout in other ticket. so it won't just change here.
https://github.com/influxdata/influxdb/issues/15375

jsteenb2: right on sounds good 👍

cmd/influx/main.go
```diff
@@ -27,6 +27,8 @@ var influxCmd = &cobra.Command{
 	Run:   influxF,
 }
 
+var maxTCPConnections = 3000
```
jsteenb2: this is a high number of concurrent requests, is this number based on benchmarks/metrics somewhere?

kelwang: i will make it 128 for now, see how it goes.

PR #15383, opened 2019-10-10
chore(ui): allow deployment configuration

drdelambre:
hopefully this commit doesnt do anything for everyone else, but it opens up the ability to
change the static folder of the project through the environment variable
STATIC_FOLDER so that we can cache static files and not cache the index
file (allowing more passive deployment strategies), and allows the user
to change the base path of the project with the BASE_PATH variable,
opening up the ability to run an instance behind an nginx proxy,
though that change is pending an update to the `@influxdata/oats` package
before we can turn it on

### DIFF
--- a/ui/src/utils/env.ts
+++ b/ui/src/utils/env.ts
@@ -11,12 +11,18 @@ const STATIC_DIRECTORY = (dir => {
     return dir
   }
 
-  return (
-    dir.slice(
-      dir[0] === '/' ? 1 : 0,
-      dir[dir.length - 1] === '/' ? -1 : undefined
-    ) + '/'
-  )
+  // NOTE: slices for sideeffect free editing (alex)
+  let _dir = dir.slice(0)
+
+  if (_dir[0] === '/') {
+    _dir = _dir.slice(1)
+  }
+
+  if (_dir[_dir.length - 1] === '/') {
+    _dir = _dir.slice(0, -1)
+  }
+
+  return _dir
 })(process.env.STATIC_DIRECTORY || '')
 
 const BASE_PATH = (prefix => {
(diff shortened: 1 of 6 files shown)


ui/src/utils/env.ts
```diff
@@ -0,0 +1,47 @@
+export {}
+
+const GIT_SHA =
+  process.env.INFLUXDB_SHA ||
```
hoorayimhelping: presumably this is so we can override the current git hash with an envvar?

ui/src/utils/env.ts
```diff
@@ -0,0 +1,47 @@
...
+    return prefix
+  }
+
+  return prefix[0] === '/'
+    ? prefix[prefix.length - 1] === '/'
+      ? prefix
+      : prefix + '/'
+    : prefix[prefix.length - 1] === '/'
+    ? '/' + prefix
+    : '/' + prefix + '/'
```
hoorayimhelping: it's probably cause i'm dumber than hell, but this takes a lot of brainpower for me to follow, between the nested ternaries and the length calculations to figuring out where in the string it is. (it also looks like there are identical cases for for the first nesting of `?` and `:` and i can't quite figure it out - again, dumb). do you think you could convert this to a regular `if / else` block or maybe introduce some variables to store intermediate results to make it a little bit easier to read?

ui/webpack.common.ts
```diff
@@ -55,6 +69,7 @@ module.exports = {
       template: './assets/index.html',
       favicon: './assets/images/favicon.ico',
       inject: 'body',
+      base: BASE_PATH.slice(0, -1),
```
hoorayimhelping: if `BASE_PATH` doesn't end with a slash, this is gonna be wonky. what do you think of adding a comment that points to the external file that makes sure basepath is properly terminated so folks reading this know we got it covered?

ui/src/utils/env.ts
```diff
@@ -0,0 +1,47 @@
+export {}
```
hoorayimhelping: what's this export doing?

drdelambre: NO IDEA.. but it was in the other webpack configs. @121watts have any insight here?

ui/webpack.common.ts
drdelambre: it should be all cleaned in the `ui/src/utils/env.ts` file, so you can throw any gibber jabber in there and it'll make it look like a folder to webpack

ui/src/utils/env.ts
drdelambre: or define it externally through a CI pipeline that only has readonly access of the code, but yeah!

ui/src/utils/env.ts
drdelambre: yeah, was conflicted about this as well. i saw a nested ternary or two in the codebase, but this one was loooooong and made it that much harder to read. update forthcoming

ui/webpack.common.ts
hoorayimhelping: >it should be all cleaned in the ui/src/utils/env.ts file, so you can throw any gibber jabber in there and it'll make it look like a folder to webpack

word!

i was poorly trying to say it might be reader friendly to add a comment saying as much (provided it specifically includes the phrase jibber jabber), just to prevent people from having to trace this import to its source to see it's handled there.

PR #15384, opened 2019-10-10
feat(task): Allow tasks to run more isolated from other task systems

lyondhill:
To allow the task internal system to be used for user created tasks as well
as checks, notification and other future additions we needed to take 2 actions:

1 - We need to use type as a first class citizen, meaning that task's have a type
and each system that will be creating tasks will set the task type through the api.
This is a change to the previous assumption that any user could set task types. This change
will allow us to have other service's white label the task service for their own purposes and not
have to worry about colissions between the types.

2 - We needed to allow other systems to add data specific to the problem they are trying to solve.
For this purpose adding a `metadata` field to the internal task system which should allow other systems to
use the task service.

These changes will allow us in the future to allow for the current check's and notifications implementations
to create a task with meta data instead of creating a check object and a task object in the database.
By allowing this new behavior checks, notifications, and user task's can all follow the same pattern:

Field an api request in a system specific http endpoint, use a small translation to the `TaskService` function call,
translate the results to what the api expects for this system, and return results.


http/task_service.go
```diff
@@ -445,6 +444,9 @@ func decodePostTaskRequest(ctx context.Context, r *http.Request) (*postTaskReque
 	}
 	tc.OwnerID = auth.GetUserID()
 
+	// when creating a task we set the type so we can filter later.
+	tc.Type = influxdb.TaskSystemType
```
AlirieGray: 👍

PR #15393, opened 2019-10-11
build(flux): update Flux to v0.50.0

jsternberg:


jsternberg:
Looks like the build is failing and needs to be investigated. I'll add you two back as reviewers once I have figured out how to resolve the issue.

query/control/memory.go
```diff
@@ -0,0 +1,145 @@
...
+		given := q.giveMemory(want, unused)
+
+		// Reserve this memory for our own use.
+		if !q.m.unlimited {
+			if !atomic.CompareAndSwapInt64(&q.m.unusedMemoryBytes, unused, unused-given) {
+				// The unused value has changed so someone may have taken
+				// the memory that we wanted. Retry.
+				continue
+			}
+		}
```
stuartcarnie: My only concern here is when there is contention, this could loop many times. We may wish to consider a retry limit and return an error.

PR #15397, opened 2019-10-11
chore(build): remove Godeps file

jacobmarble:
Remove gdm support, as it was replaced by dep a long time ago. Godeps has been removed in OSS 1.7 and 1.8 branches.


gshif:
@jacobmarble , still need this file as it is being used by plutonium

PR #15406, opened 2019-10-14
refactor(tasks): use Go time objects for timestamps on task Runs

### LINKED ISSUE #14523: Task runs needs to use a less ambiguous time value
We built the task system on the task struct as defined for the api:
```
type Run struct {
	ID           ID     `json:"id,omitempty"`
	TaskID       ID     `json:"taskID"`
	Status       string `json:"status"`
	ScheduledFor string `json:"scheduledFor"`
	StartedAt    string `json:"startedAt,omitempty"`
	FinishedAt   string `json:"finishedAt,omitempty"`
	RequestedAt  string `json:"requestedAt,omitempty"`
	Log          []Log  `json:"log,omitempty"`
}
```
Although this structure works well for the front end I believe using `string` as a timestamp makes it difficult to reason about what is actually stored in the field (unix, utc based time stamps, microseconds, nanoseconds). 
I believe we would benefit by changing all time based values to time.Time and then translating to something api friendly in http/task_service.go instead of using strings for these values on the backend.

AlirieGray:
Closes #14523

This PR converts the the type of the `StartedAt`, `ScheduledFor`, `FinishedAt`, and `RequestedAt` properties of the Task Run struct from strings to Go time.Time objects. This will reduce the complexity of having to parse/format these timestamps throughout the internal task systems. 

These properties will still be stored in the database as strings, for backwards compatibility.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/task/backend/analytical_storage.go
+++ b/task/backend/analytical_storage.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/lang"
@@ -317,10 +318,7 @@ func (as *AnalyticalStorage) RetryRun(ctx context.Context, taskID, runID influxd
 		return run, err
 	}
 
-	sf, err := run.ScheduledForTime()
-	if err != nil {
-		return run, err
-	}
+	sf := run.ScheduledFor
 
 	return as.ForceRun(ctx, taskID, sf.Unix())
 }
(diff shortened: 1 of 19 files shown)


task/backend/analytical_storage.go
```diff
@@ -67,25 +67,20 @@ func (as *AnalyticalStorage) FinishRun(ctx context.Context, taskID, runID influx
...
-		if err != nil {
-			startedAt = time.Now()
+		// cast to UTC here and when we read from the database so that run timestamps are always in UTC format
+		fields[startedAtField] = run.StartedAt.UTC().Format(time.RFC3339Nano)
+		fields[finishedAtField] = run.FinishedAt.UTC().Format(time.RFC3339Nano)
+		fields[scheduledForField] = run.ScheduledFor.UTC().Format(time.RFC3339)
+
+		startedAt := run.StartedAt
+		if !run.IsStarted() {
+			startedAt = time.Now().UTC()
```
AlirieGray: This code previously checked for both an empty StartedAt and RequestedAt. I removed the check for the empty RequestedAt, since it will just be written to the object as the zero time if it does not exist, but I left in the check for StartedAt since that will be used as the timestamp for the call to `NewPoint` below. I also added utility functions to check if a run has been started or finished, but not for requested or scheduled. Open to suggestions on this, we could alternatively use a single IsZeroTime function, remove these functions, etc.

lyondhill: that is probably fine, The run should not get to this point without being started since the we are in the `FinishRun` method.

task/backend/executor/limits.go
```diff
@@ -28,14 +28,10 @@ func ConcurrencyLimit(exec *TaskExecutor) LimitFunc {
...
-			if err != nil {
-				return false
-			}
-			runj, err := runs[j].ScheduledForTime()
-			if err != nil {
-				return false
-			}
+			runi := runs[i].ScheduledFor
+
+			runj := runs[j].ScheduledFor
```
lyondhill: If you wanted to make this shorter you could do `return run[i].ScheduledFor.Before(runs[j].ScheduledFor)` its a bit wordy but it does make it a single line function.

Issue #15243, opened 2019-09-24
members and owners endpoints should check if resource exists


jademcgough:
Most of our endpoints return a 404 if a resource doesn't exist:
```
GET http://localhost:9999/api/v2/orgs/0123456789abcde5:
{
    "code": "not found",
    "message": "organization not found"
}
```

However, member/owner endpoints don't check to ensure that their resource exists, causing them to return an empty array:
```
GET http://localhost:9999/api/v2/orgs/0123456789abcde5/members:
{
    "links": {
        "self": "/api/v2/orgs/0123456789abcde5/members"
    },
    "users": []
}
```

We use a builder pattern for generating handlers for those endpoints:
https://github.com/influxdata/influxdb/blob/bdba601c306101d5448322530667cca733e8e6b9/http/bucket_service.go#L104-L106

We should add code that is executed before that handler logic. One way to do that would be by [chaining http handlers](https://www.alexedwards.net/blog/making-and-using-middleware).

jademcgough:
closed by #15408

PR #15408, opened 2019-10-14
fix: /owners /members endpoints to 404 when resources don't exist

dearyhud:
Closes #
https://github.com/influxdata/influxdb/issues/15243

members and owners endpoints typically return an empty array in the response body, when in that case I think we want to 404.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/influxdata/influxdb"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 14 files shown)


http/middleware.go
```diff
@@ -151,3 +152,39 @@ func (b *bodyEchoer) Read(p []byte) (int, error) {
 func (b *bodyEchoer) Close() error {
 	return b.rc.Close()
 }
+
+func getMembersPrecheckMiddleware(next http.HandlerFunc, backend MemberBackend) http.HandlerFunc {
```
jsteenb2: nit: the signature here you may want to consider using a closure and returning the middleware func. Its a popular pattern in a lot of http middleware in go. Check out the following: https://github.com/influxdata/influxdb/blob/master/http/middleware.go#L15

the func closes over the logger here, and returns a middleware (func(http.handler) http.Handler), which allows you to chain these things very easily. In your case it would look like:

```go
func getMembersPrecheckMiddleware(backend MemberBackend) func(http.Handler) http.Handler {
   return func(next http.Handler) http.Handler {
      fn := func(w http.ResponseWriter, r *http.Request) {
         // req handler stuffs yada yada
      }

     return http.HandlerFunc(fn)
   }
```

in hindsight, I wish I would have made the following type to illuminate that better:

```go
type Middleware func(next http.handler) http.Handler
```

jsteenb2: where are the accompanying tests to verify this behavior?

http/middleware.go
```diff
@@ -151,3 +152,39 @@ func (b *bodyEchoer) Read(p []byte) (int, error) {
...
+		if err != nil {
+			backend.HandleHTTPError(ctx, err, w)
+			return
+		}
+
+		if count == 0 {
+			backend.Logger.Debug("No members/owners retrieved")
+			err := &influxdb.Error{
+				Code: influxdb.ENotFound,
+				Op:   influxdb.OpFindOrganization,
```
jsteenb2: is this `Op` accurate?

121watts:
Looks like the ui client is no longer being successfully built.  Hence the js and e2e failures

dearyhud:
closing this PR in favor of https://github.com/influxdata/influxdb/pull/15504/files

PR #15426, opened 2019-10-15
feat(ui): add 403 signout handler to new api requests

### LINKED ISSUE #15312: 403s from the API should log out user
As a user who is cancelled, anytime the api endpoint comes back with 403 error, the user should be logged out automatically.

We return a 401 for when a user is not authorized to access a resources and 403 when they are disabled. So we should log them out on 403 and not on 401.

hoorayimhelping:
Closes #15312

- Adds a custom response handler for the new api routes created by oats that immediately logs the user out and returns them to `/signin` when a 403 is returned from the API
  - upgrades to oats 0.5.0

out

### Testing Instructions
* If you'd like to see which routes use the new api and go through this handler, you can add a log statement [here](https://github.com/influxdata/influxdb/blob/bs_bugfix_403_logout/ui/src/client/index.ts#L4)
* To force 403s from the API (only works on the api calls that use `client/index`):
  * change [authentication_middleware.go](https://github.com/influxdata/influxdb/blob/bs_bugfix_403_logout/http/authentication_middleware.go#L110) to something [like this](https://gist.github.com/hoorayimhelping/c138a0f393b3face916637b3db29ffe2) (change the id to match yours, which can be found in the url on home page after you log in to the app)
  * re-build and run: `go run ./cmd/influxd --assets-path=ui/build`
  * on the sidebar, click settings -> members (log back in, if necessary)
  * you should be sent back to the `/signin` screen with whatever message you logged in 403 handler.
* To ensure `/me` works (this is a little tricky):
  * start the app normally
  * sign in
  * change [authentication_middleware.go](https://github.com/influxdata/influxdb/blob/bs_bugfix_403_logout/http/authentication_middleware.go#L110) to [check for `/api/v2/me`](https://gist.github.com/hoorayimhelping/09d7d337316411696fd8f970dcd606a3)
   * re-build and run **without reloading and signing back in**: `go run ./cmd/influxd --assets-path=ui/build`
  * wait
  * see the user get logged out
    * if you added a log message in the handler above, you'll see that it doesn't get called. *403s from `/me` already get kicked back to the `/signin` screen.*

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/client/index.ts
+++ b/ui/src/client/index.ts
@@ -1,0 +1,12 @@
+import {setResponseHandler, postSignout} from './generatedRoutes'
+
+setResponseHandler((status, headers, data) => {
+  if (status === 403) {
+    postSignout({})
+    window.location.href = '/signin'
+  }
+
+  return {status, headers, data}
+})
+
+export * from './generatedRoutes'
--- a/.gitignore
+++ b/.gitignore
@@ -43,7 +43,7 @@ ui/src/api/.openapi-generator/VERSION
 ui/src/api/git_push.sh
 
 # UI generated typescript types
-ui/src/client/index.ts
+ui/src/client/generatedRoutes.ts
 
 http/swagger_gen.go
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 ### Features
 1. [15313](https://github.com/influxdata/influxdb/pull/15313): Add shortcut for toggling comments in script editor
 
+### UI Improvements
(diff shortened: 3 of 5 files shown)


drdelambre:
before i start looking at this... GREAT PR DESCRIPTION

ui/src/client/index.ts
```diff
@@ -0,0 +1,12 @@
+import {setResponseHandler, postSignout} from './generatedRoutes'
+
+setResponseHandler((status, headers, data) => {
```
ebb-tide: beautiful!

PR #15429, opened 2019-10-15
fix(ui): adding coverage to ignore file

drdelambre:
that should not have been there 😬


drdelambre:
thanks for the catch and being cool about it!

PR #15430, opened 2019-10-15
fix(tasks): replace deactivation of unrecoverable errors with metric

AlirieGray:
This PR comments out the deactivation logic for unrecoverable errors, to be restored when a user notification system is put in place to alert users when they have a task that is going to be deactivated.

This PR also adds a new metric for the new executor called "Unrecoverable Counter" which will allow engineers to see in one place all of the errors for which we would like to deactivate the offending tasks.

The error message "filesystem service uninitialized" is also added to the list of unrecoverable errors.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/task/backend/check_task_error.go
+++ b/task/backend/check_task_error.go
@@ -22,5 +22,10 @@ func IsUnrecoverable(err error) bool {
 		return true
 	}
 
+	// Flux script uses an API that attempts to read the filesystem
+	if strings.Contains(errString, "filesystem service uninitialized") {
+		return true
+	}
+
 	return false
 }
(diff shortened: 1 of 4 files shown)


task/backend/check_task_error.go
```diff
@@ -22,5 +22,9 @@ func IsUnrecoverable(err error) bool {
 		return true
 	}
 
+	if strings.Contains(errString, "filesystem service uninitialized") {
```
stuartcarnie: Perhaps a comment

```
// Flux script uses an API that attempts to read the filesystem
```

PR #15501, opened 2019-10-18
 chore(ui/GetResources): parallelized GetResources CDM calls

### LINKED ISSUE #13608: It should be possible to parallelize resource fetching
When we need more than one resource to load a page, the current pattern we are using with the GetResources component encourages nested components which execute fetches serially when they could be done in parallel. Let's investigate a solution to this, perhaps GetResources could take an array of resourcetypes as props instead of just one, and wait until all are resolved/ achieved RemoteDataState.Done to render children.

Asalem1:
Closes #13608 

### Problem

GetResources accepted a single ResourceType parameter, forcing developers to wrap child components in multiple GetResources components to retrieve all relevant data

### Solution

Updated the GetResources component to accept an array of ResourceType parameters and refactored the CDM function to run multiple resource fetches in parallel. Also refactored the current uses of GetResources throughout the app to reflect these changes


ui/src/shared/components/GetResources.tsx
```diff
@@ -166,11 +173,34 @@ class GetResources extends PureComponent<Props, StateProps> {
   }
 
   public render() {
-    const {resource, children} = this.props
+    const {resources, children} = this.props
+    let status
+    for (let i = 0; i < resources.length; i++) {
+      const resource = resources[i]
+      // reduce the values of the status to 1 value
```
121watts: Why not use the declarative `Array.prototype.reduce` and put this logic in a getter?

Asalem1: When i refactored this to put the logic in a getter I was getting a type error thrown for status being a string (and refactoring the value to return `this.props[resource].status` for each of the cases seemed a bit repetitive). 

As for using reduce, I'd love to talk about how to integrate it without overriding the status value & breaking out of a reduce

ui/src/shared/components/GetResources.tsx
```diff
@@ -166,13 +175,33 @@ class GetResources extends PureComponent<Props, StateProps> {
...
+        case 'Done': {
+          status = 'Done'
+          break
+        }
+        default: {
+          status = 'Error'
+          break
+        }
+      }
+    }
```
ebb-tide: Yeah I don't think this is gonna work, let's go over this again.

PR #15504, opened 2019-10-18
fix(http): /api/v2/orgs/:id/owners|members to 404

dearyhud:
Closes #

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/http/middleware.go
+++ b/http/middleware.go
@@ -12,7 +12,11 @@ import (
 	"go.uber.org/zap"
 )
 
-func LoggingMW(logger *zap.Logger) func(http.Handler) http.Handler {
+// Middleware constructor.
+type Middleware func(http.Handler) http.Handler
+
+// LoggingMW middleware for logging inflight http requests.
+func LoggingMW(logger *zap.Logger) Middleware {
 	return func(next http.Handler) http.Handler {
 		fn := func(w http.ResponseWriter, r *http.Request) {
 			srw := &statusResponseWriter{
@@ -150,4 +154,16 @@ func (b *bodyEchoer) Read(p []byte) (int, error) {
 
 func (b *bodyEchoer) Close() error {
 	return b.rc.Close()
+}
+
+func applyMW(h http.Handler, m ...Middleware) http.Handler {
+	if len(m) < 1 {
+		return h
+	}
+	wrapped := h
+
+	for i := len(m) - 1; i >= 0; i-- {
+		wrapped = m[i](wrapped)
+	}
+	return wrapped
 }
--- a/http/org_service.go
+++ b/http/org_service.go
@@ -73,6 +73,28 @@ const (
 	organizationsIDLabelsIDPath      = "/api/v2/orgs/:id/labels/:lid"
 )
 
+func checkOrganziationExists(handler *OrgHandler) Middleware {
+	fn := func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			ctx := r.Context()
+			req, err := decodeGetOrgRequest(ctx, r)
+			if err != nil {
+				handler.HandleHTTPError(ctx, err, w)
+				return
+			}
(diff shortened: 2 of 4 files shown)


http/org_service.go
```diff
@@ -73,6 +73,24 @@ const (
 	organizationsIDLabelsIDPath      = "/api/v2/orgs/:id/labels/:lid"
 )
 
+func (h *OrgHandler) handleGetOrgCheck(w http.ResponseWriter, r *http.Request) interface{} {
```
Asalem1: love the helper function!

bthesorceror: Could this just return a organization pointer?

http/org_service.go
```diff
@@ -238,21 +259,17 @@ func decodePostOrgRequest(ctx context.Context, r *http.Request) (*postOrgRequest
...
-		h.HandleHTTPError(ctx, err, w)
+	if organization == nil {
 		return
 	}
-	h.Logger.Debug("org retrieved", zap.String("org", fmt.Sprint(b)))
 
-	if err := encodeResponse(ctx, w, http.StatusOK, newOrgResponse(b)); err != nil {
+	ctx := r.Context()
+
+	org := organization.(*influxdb.Organization)
```
bthesorceror: See comment above, this would remove the need to cast.

http/user_resource_mapping_service.go
```diff
@@ -63,11 +63,16 @@ type MemberBackend struct {
 
 	UserResourceMappingService influxdb.UserResourceMappingService
 	UserService                influxdb.UserService
+	Precheck                   func(w http.ResponseWriter, r *http.Request) interface{}
```
bthesorceror: Could this function type have a defined return type instead of `interface{}`?

http/middleware.go
```diff
@@ -151,3 +151,17 @@ func (b *bodyEchoer) Read(p []byte) (int, error) {
 func (b *bodyEchoer) Close() error {
 	return b.rc.Close()
 }
+
+type middleware func(http.HandlerFunc) http.HandlerFunc
```
jsteenb2: can we export this type here, and refactor the `LoggignMW` to use it as well?

http/org_service.go
```diff
@@ -73,6 +73,41 @@ const (
...
+		return &influxdb.Error{
+			Err: findErr,
+		}
+	}
+
+	return nil
+}
+
+func checkOrganziationExistsWrapper(handler *OrgHandler) func(http.HandlerFunc) http.HandlerFunc {
+	fn := func(next http.HandlerFunc) http.HandlerFunc {
```
jsteenb2: 👏 looking good, very foxused func here, I dig it 👍

http/middleware.go
```diff
@@ -151,3 +151,17 @@ func (b *bodyEchoer) Read(p []byte) (int, error) {
...
+
+func handlerFuncWithMiddlewares(h http.HandlerFunc, m ...middleware) http.HandlerFunc {
```
jsteenb2: can you rename this to something shorter maybe? `applyMW`? this is a mouthful for something you'll be chaining.  Its quite common to see `MW` the type for middleware and how its referenced in a lot of go pkgs

http/org_service.go
```diff
@@ -73,6 +73,41 @@ const (
...
+	if findErr != nil {
+		return &influxdb.Error{
+			Err: findErr,
+		}
+	}
+
+	return nil
+}
+
+func checkOrganziationExistsWrapper(handler *OrgHandler) func(http.HandlerFunc) http.HandlerFunc {
```
jsteenb2: go on and take advantage of the type you just created `func organziationExists(handler *OrgHandler) Middleware {`

one other thing here is that you are passing the org handler as a value instead of using pointer semantics. you want to stick with one or the other to be consistent here. I'd recommend using pointer semantics here as that's what we are sharing across these functional boundaries

PR #15510, opened 2019-10-18
fix(ui/collectorList-telegraf): sort telegrafs based on buckets

### LINKED ISSUE #15247: [UI - Telegrafs] Sort by bucket is not working
__Steps to reproduce:__

1. Create a few telegraf resources cards with distinguishing names and place them variously into two or more buckets which also have distinguishing names
2. Sort the cards by name ascending and descending - works fine
3. Sort the cards by bucket.  Try clicking sort a number of times. 

__Expected behavior:__
That the telegraf cards would be sorted alphanumerically based on the bucket names 

__Actual behavior:__
The logic behind the initial sort is unclear.  For example I have buckets named "qa" and "Duchamp" but the order is {qa, Duchamp, qa, Duchamp, qa}.  Additional clicks trigger no further changes in the sort order.

__Screencast__
SortTelegrafByBuckets01

__Environment info:__

Linux 4.4.0-164-generic x86_64
This is from the latest nightly build - docker image.  From the docker.log
```
ts=2019-09-24T08:43:21.868438Z lvl=info msg="Welcome to InfluxDB" log_id=0I4bsVZ0000 version=nightly commit=c3127c8cd build_date=2019-09-24T07:19:11Z
```
**Browser** Chrome 76

__Config:__
Started with flag enabling testing environment 
```
--e2e-testing=true
```

__Logs:__
Nothing in the console log

Asalem1:
Closes #15247

### Problem

UI/Telegrafs should be sortable based on bucket name

### Solution

Modified the sortKey to accurately reflect the bucket's location in the data, thereby integrating the existing sort functionality.

Kapture 2019-10-18 at 15 14 39

### DIFF
--- a/ui/cypress/e2e/collectors.test.ts
+++ b/ui/cypress/e2e/collectors.test.ts
@@ -1,5 +1,4 @@
 import {Organization} from '../../src/types'
-import {bucket} from '../fixtures/user.json'
 
 describe('Collectors', () => {
   beforeEach(() => {
@@ -47,10 +46,12 @@ describe('Collectors', () => {
           .click()
       })
 
-      cy.getByTestID('resource-card')
-        .should('have.length', 1)
-        .and('contain', newConfig)
-        .and('contain', bucket)
+      cy.fixture('user').then(({bucket}) => {
+        cy.getByTestID('resource-card')
+          .should('have.length', 1)
+          .and('contain', newConfig)
+          .and('contain', bucket)
+      })
     })
 
     it('can update configuration name', () => {
@@ -60,7 +61,9 @@ describe('Collectors', () => {
       const description = 'Config Description'
 
       cy.get('@org').then(({id}: Organization) => {
-        cy.createTelegraf(telegrafConfigName, description, id, bucket)
+        cy.fixture('user').then(({bucket}) => {
+          cy.createTelegraf(telegrafConfigName, description, id, bucket)
+        })
       })
 
       cy.getByTestID('collector-card--name')
@@ -83,7 +86,9 @@ describe('Collectors', () => {
         const telegrafConfigName = 'New Config'
         const description = 'Config Description'
         cy.get('@org').then(({id}: Organization) => {
-          cy.createTelegraf(telegrafConfigName, description, id)
+          cy.fixture('user').then(({bucket}) => {
+            cy.createTelegraf(telegrafConfigName, description, id, bucket)
+          })
         })
       })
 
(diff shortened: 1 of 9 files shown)


drdelambre:
while you're fixing test snapshots.. try to write a test for this one

ui/cypress/e2e/collectors.test.ts
```diff
@@ -1,4 +1,5 @@
 import {Organization} from '../../src/types'
+import {bucket} from '../fixtures/user.json'
```
Asalem1: is this even kosher?

ui/cypress/e2e/collectors.test.ts
```diff
@@ -165,5 +170,99 @@ describe('Collectors', () => {
...
+          cy.getByTestID('resource-list--sorter')
+            .contains('Name')
+            .click()
+          telegrafs.reverse()
+        })
+        .then(() => {
+          cy.getByTestID('collector-card--name').each((val, index) => {
+            expect(val.text()).to.equal(telegrafs[index])
+          })
+        })
```
Asalem1: Couldn't find a working solution with `cypress` other than to chain statements. Would love to hear about any other implementations to guarantee serialization of tests

ui/cypress/e2e/collectors.test.ts
```diff
@@ -165,5 +170,99 @@ describe('Collectors', () => {
...
+      // cy.getByTestID('resource-card').should('have.length', 3)
+      cy.getByTestID('collector-card--name')
+        .not('a')
+        .should('have.length', 3)
+
+      // test to see if telegrafs are initially sorted by name
+      telegrafs.sort()
+
+      cy.getByTestID('collector-card--name')
+        .each((val, index) => {
```
Asalem1: ended up using `each` since it was the `cypress` method that easily made the text of the node available for comparison

ui/cypress/e2e/collectors.test.ts
```diff
@@ -165,5 +170,99 @@ describe('Collectors', () => {
...
+        cy.createTelegraf(firstTelegraf, description, id, firstBucket)
+        cy.createTelegraf(secondTelegraf, description, id, secondBucket)
+        cy.createTelegraf(thirdTelegraf, description, id, thirdBucket)
+      })
+
+      cy.reload()
+
+      // test the order
+      // find all the cf-resource-card--meta-item without children
+      cy.getByTestID('cf-resource-card--meta-item').should('have.length', 6)
```
Asalem1: since we are using built-in testIDs from the `@clockface` dependencies, I couldn't figure out a way to fully customize the selectors. I tried using `not`, `filter`, chaining `get`, `contains`, etc... all without reaching the intended purpose. I would love to work on fixing this since this is shouldn't be merged onto master

ui/cypress/e2e/collectors.test.ts
```diff
@@ -165,5 +170,99 @@ describe('Collectors', () => {
...
+
+      cy.getByTestID('cf-resource-card--meta-item')
+        .each((val, index) => {
+          const text = val.text()
+          const i = Math.floor(index / 2)
```
Asalem1: a side-effect of pulling too many nodes

ui/cypress/e2e/collectors.test.ts
```diff
@@ -165,5 +170,99 @@ describe('Collectors', () => {
...
+          })
+        })
+        .then(() => {
+          cy.getByTestID('resource-list--sorter')
+            .contains('Bucket')
+            .click()
+          bucketz.reverse()
+          cy.getByTestID('cf-resource-card--meta-item').each((val, index) => {
+            const text = val.text()
+            const i = Math.floor(index / 2)
```
Asalem1: same as before

ui/cypress/e2e/collectors.test.ts
121watts: `cy.fixtures`

ui/cypress/e2e/collectors.test.ts
```diff
@@ -121,11 +123,12 @@ describe('Collectors', () => {
...
       const description = 'Config Description'
+      const secondBucket = 'bucket2'
 
       cy.get('@org').then(({id}: Organization) => {
-        cy.createTelegraf(firstTelegraf, description, id)
-        cy.createTelegraf(secondTelegraf, description, id)
-        cy.createTelegraf(thirdTelegraf, description, id)
+        cy.createTelegraf(firstTelegraf, description, id, bucket)
+        cy.createTelegraf(secondTelegraf, description, id, bucket)
+        cy.createTelegraf(thirdTelegraf, description, id, secondBucket)
```
121watts: I would create all of these before this test runs in a `beforeEach`.  You can then nest the sort by name and sort by bucket tests and not have to repeat this creation logic.

ui/cypress/e2e/collectors.test.ts
121watts: chaining statements in cypress is totally fine.

ui/cypress/e2e/collectors.test.ts
121watts: you should be able to pass arbitrary testIDs / overwrite the default testIDs for any clockface component.

ui/cypress/e2e/collectors.test.ts
121watts:
```
describe('sorting', () => {
  beforeEach(() => {
    // creation logic here
  }
  // sort by buckets test here
  // sort by name test here
}
```

ui/cypress/e2e/collectors.test.ts
```diff
@@ -121,11 +128,14 @@ describe('Collectors', () => {
...
 
       cy.get('@org').then(({id}: Organization) => {
-        cy.createTelegraf(firstTelegraf, description, id)
-        cy.createTelegraf(secondTelegraf, description, id)
-        cy.createTelegraf(thirdTelegraf, description, id)
+        cy.fixture('user').then(({bucket}) => {
+          cy.createTelegraf(firstTelegraf, description, id, bucket)
+          cy.createTelegraf(secondTelegraf, description, id, bucket)
+          cy.createTelegraf(thirdTelegraf, description, id, secondBucket)
+        })
```
121watts: Might as well move these into a `beforeEach` as well

ui/cypress/e2e/collectors.test.ts
```diff
@@ -165,5 +177,64 @@ describe('Collectors', () => {
...
+      // sort by buckets test here
+      it('can sort telegraf configs by bucket', () => {
+        cy.getByTestID('bucket-name').should('have.length', 3)
+        cy.getByTestID('bucket-sorter').click()
+
+        bucketz.sort()
+        cy.getByTestID('bucket-name')
+          .each((val, index) => {
+            const text = val.text()
+            expect(text).to.equal(`Bucket: ${bucketz[index]}`)
```
121watts: Might want to do a less strict match here.  `Bucket: ` is a presentational concern and is highly likely to change.  Something like `includes` instead of `equal`

ui/cypress/e2e/collectors.test.ts
```diff
@@ -165,5 +177,64 @@ describe('Collectors', () => {
...
+      // sort by name test here
+      it('can sort telegraf configs by name', () => {
+        cy.getByTestID('collector-card--name').should('have.length', 3)
+
+        // test to see if telegrafs are initially sorted by name
+        telegrafs.sort()
+
+        cy.getByTestID('collector-card--name')
+          .each((val, index) => {
+            expect(val.text()).to.equal(telegrafs[index])
```
121watts: Same thing here.  If we want to add any extra presentation to the string this will break the test but not the functionality

ui/cypress/tsconfig.json
```diff
@@ -6,7 +6,8 @@
     "lib": ["es2017", "dom"],
     "types": ["cypress", "mocha", "node"],
     "jsx": "react",
-    "allowSyntheticDefaultImports": true
+    "allowSyntheticDefaultImports": true,
```
121watts: Yaaaas

PR #15517, opened 2019-10-21
feat: add full compaction semaphore abstraction

e-dard:
This feature adds the concept of a semaphore to the `tsm1.Engine`, which can be used to limit full compactions. By default this feature is disabled; the full compaction behaviour does
not change. 

When this feature is enabled, compactions can be limited across multiple storage engines running in multiple processes through an implementation of an `influxdb.Semaphore`.

Currently the `Semaphore` interface defines non-blocking acquisition methods as that's what's used in the `tsm1.Engine`. We could extend the interface in the future to add blocking methods.

### DIFF
--- a/semaphore.go
+++ b/semaphore.go
@@ -1,0 +1,58 @@
+package influxdb
+
+import (
+	"context"
+	"time"
+)
+
+// DefaultLeaseTTL is used when a specific lease TTL is not requested.
+const DefaultLeaseTTL = time.Minute
+
+// A Semaphore provides an API for requesting ownership of an expirable semaphore.
+//
+// Acquired semaphores have an expiration period. If they're not released or extended
+// during this period then they will expire and ownership of the semaphore will
+// be lost.
+type Semaphore interface {
+	// TODO(edd): add Acquire and AcquireTTL when needed. These should block.
+
+	// TryAcquire attempts to acquire ownership of the semaphore. TryAcquire
+	// must not block. Failure to get ownership of the semaphore should be
+	// signalled to the caller via the return of a nil Lease.
+	TryAcquire(context.Context) (Lease, error)
+
+	// TryAcquireTTL is similar to TryAcquire, but a specific TTL is provided.
+	TryAcquireTTL(ctx context.Context, ttl time.Duration) (Lease, error)
+}
+
+// A Lease represents ownership over a semaphore. It gives the owner the ability
+// to extend ownership over the semaphore or release ownership of the semaphore.
+type Lease interface {
+	// TTL returns the duration of time remaining before the lease expires.
+	TTL(context.Context) (time.Duration, error)
+
+	// Release terminates ownership of the semaphore by revoking the lease.
+	Release(context.Context) error
+
+	// KeepAlive extends the lease back to the original TTL.
+	KeepAlive(context.Context) error
+}
+
+// NopSemaphore is a Semaphore that always hands out leases.
+var NopSemaphore Semaphore = nopSemaphore{}
+
+type nopSemaphore struct{}
+
+func (nopSemaphore) TryAcquire(context.Context) (Lease, error) {
+	return nopLease{}, nil
+}
+
+func (nopSemaphore) TryAcquireTTL(ctx context.Context, ttl time.Duration) (Lease, error) {
+	return nopLease{}, nil
+}
+
+type nopLease struct{}
+
+func (nopLease) TTL(context.Context) (time.Duration, error) { return DefaultLeaseTTL, nil }
+func (nopLease) Release(context.Context) error              { return nil }
+func (nopLease) KeepAlive(context.Context) error            { return nil }
--- a/semaphore.go
+++ b/semaphore.go
@@ -2,27 +2,29 @@ package influxdb
 
 import (
 	"context"
+	"errors"
 	"time"
 )
+
+// ErrNoAcquire is returned when it was not possible to acquire ownership of the
+// semaphore.
+var ErrNoAcquire = errors.New("ownership not acquired")
 
 // DefaultLeaseTTL is used when a specific lease TTL is not requested.
 const DefaultLeaseTTL = time.Minute
 
 // A Semaphore provides an API for requesting ownership of an expirable semaphore.
 //
-// Acquired semaphores have an expiration period. If they're not released or extended
+// Acquired semaphores have an expiration. If they're not released or kept alive
 // during this period then they will expire and ownership of the semaphore will
 // be lost.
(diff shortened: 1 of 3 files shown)


tsdb/tsm1/engine.go
```diff
@@ -1104,15 +1116,66 @@ func (e *Engine) compactFull(ctx context.Context, grp CompactionGroup, wg *sync.
 
 	// Try the lo priority limiter, otherwise steal a little from the high priority if we can.
 	if e.compactionLimiter.TryTake() {
+		// Attempt to get ownership of the semaphore for this engine. If the
+		// default semaphore is in use then ownership will always be granted.
+		ttl := influxdb.DefaultLeaseTTL
+		lastCompaction := time.Duration(atomic.LoadInt64(&e.lastFullCompactionDuration))
+		if lastCompaction > ttl {
+			ttl = lastCompaction // If the last full compaction took > default ttl then set a new TTL
```
goller: What do you think about logging at Info that compaction is using a non-default value?

tsdb/tsm1/engine.go
```diff
@@ -1104,15 +1116,66 @@ func (e *Engine) compactFull(ctx context.Context, grp CompactionGroup, wg *sync.
...
+			for {
+				select {
+				case <-done:
+					if err := lease.Release(ctx); err != nil {
+						e.logger.Warn("Lease on sempahore was not released", zap.Error(err))
+					}
+					return
+				case <-ticker.C:
+					if err := lease.KeepAlive(ctx); err != nil {
+						e.logger.Warn("Unable to extend lease", zap.Error(err))
```
goller: Is it possible to be in a situation where KeepAlive is never able to succeed?  If so, is returning appropriate?

tsdb/tsm1/engine.go
e-dard: That would be the expected case, and might lead to log messages every second whilst the engine was blocked waiting for ownership to the semaphore. We could add a log line when we get the semaphore though and it's not the `no-op` one?

tsdb/tsm1/engine.go
e-dard: The only time that might happen is if `etcd` is broken I think. When this compaction finishes the go routine would be cleaned up anyway.

semaphore.go
```diff
@@ -0,0 +1,58 @@
...
+// Acquired semaphores have an expiration period. If they're not released or extended
+// during this period then they will expire and ownership of the semaphore will
+// be lost.
+type Semaphore interface {
+	// TODO(edd): add Acquire and AcquireTTL when needed. These should block.
+
+	// TryAcquire attempts to acquire ownership of the semaphore. TryAcquire
+	// must not block. Failure to get ownership of the semaphore should be
+	// signalled to the caller via the return of a nil Lease.
+	TryAcquire(context.Context) (Lease, error)
```
benbjohnson: I would personally use an error to signal that the acquisition failed (e.g. `ErrNoAcquire`). I find it's easy for people to forget to check a `nil` `Lease`.

semaphore.go
```diff
@@ -0,0 +1,58 @@
...
+
+	// TryAcquireTTL is similar to TryAcquire, but a specific TTL is provided.
+	TryAcquireTTL(ctx context.Context, ttl time.Duration) (Lease, error)
+}
```
benbjohnson: Did you not like the name, `Lessor`? Acquiring a lease from a semaphore seems odd to me personally.

tsdb/tsm1/engine.go
```diff
@@ -1104,15 +1116,66 @@ func (e *Engine) compactFull(ctx context.Context, grp CompactionGroup, wg *sync.
...
+		done := make(chan struct{}) // Closed when compaction finished.
+		go func(lease influxdb.Lease) {
+			ttl, err := lease.TTL(ctx)
+			if err != nil {
+				e.logger.Warn("unable to get TTL for lease on semaphore", zap.Error(err))
+				ttl = influxdb.DefaultLeaseTTL // This is probably a reasonable fallback.
+			}
+
+			// Renew the lease when ttl is halved
+			ticker := time.NewTicker(ttl / 2)
```
benbjohnson: Can you `defer ticker.Stop()` to make sure we're not leaking goroutines?

tsdb/tsm1/engine.go
```diff
@@ -1104,15 +1116,66 @@ func (e *Engine) compactFull(ctx context.Context, grp CompactionGroup, wg *sync.
...
+					return
+				case <-ticker.C:
+					if err := lease.KeepAlive(ctx); err != nil {
+						e.logger.Warn("Unable to extend lease", zap.Error(err))
+					} else {
+						e.logger.Info("Extended lease on semaphore")
+					}
+				}
+			}
+		}(lease)
```
benbjohnson: This method is already pretty long. Can you move this goroutine to a separate method?

semaphore.go
benbjohnson: Also, if this is meant to support distributed use cases, it seems like all acquisitions should require a TTL (i.e. only implement `TryAcquireTTL()`).

e-dard: I'm open to doing that. We could probably remove `TryAcquire` since in the end it's unlikely to get many callers (mostly you want to dictate the TTL upfront I guess).

semaphore.go
e-dard: If anything `Lease` could be renamed to something else, but not sure what. I kind of worried about this. Usually you wouldn't return anything aside from maybe a `bool` indicating you have ownership, but in this case I wanted something that can be extended and revoked early. It made more sense to return that complex logic in a value. Essentially the semaphore is granting leased ownership over some resource (since leases have a fixed duration).

e-dard:
@benbjohnson PTAL. I added a couple of commits.

tsdb/tsm1/engine.go
```diff
@@ -1104,21 +1116,82 @@ func (e *Engine) compactFull(ctx context.Context, grp CompactionGroup, wg *sync.
...
+				if err := lease.KeepAlive(ctx); err != nil {
+					e.logger.Warn("Unable to extend lease", zap.Error(err))
+				} else {
+					e.logger.Info("Extended lease on semaphore")
+				}
+			}
+		}
+	}(lease)
+	return done
+}
```
benbjohnson: I was thinking that `monitorLease()` wouldn't have a goroutine and that the caller would do `go e.monitorLease()`.

Also, it seems like using `context.WithCancel()` and using `ctx` instead of `done` would be more idiomatic.

PR #15241, opened 2019-09-23
feat(influxdb): Add system buckets on org creation

bthesorceror:
Creates system buckets on organization creation

- [x] Rebased/mergeable
- [x] Tests pass
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)

PR #15529, opened 2019-10-22
fix(kv): FindBucketByName for system buckets fallback to legacy buckets

GeorgeMac:
I am unsure if this is desired behavior or not. However, I am observing lots of `_tasks` bucket not found errors.

A recent change made system buckets an organisation scope entity, rather than using the global buckets (10 for tasks and 11 for monitoring) #15241 
This change just ensures that when using `FindBucketsByName()` it can still locate the legacy locations for these system buckets when an existing organisation hasn't yet migrated to use the organisation scoped system buckets.

- [x] Rebased/mergeable
- [x] Tests pass


stuartcarnie:
This has been closed by #15531

PR #15540, opened 2019-10-22
fix(influxdb): Add OrgID to system buckets

bthesorceror:


kv/bucket.go
```diff
@@ -189,14 +189,17 @@ func (s *Service) findBucketByName(ctx context.Context, tx Tx, orgID influxdb.ID
 				Name:            influxdb.TasksSystemBucketName,
 				RetentionPeriod: influxdb.TasksSystemBucketRetention,
 				Description:     "System bucket for task logs",
+				OrgID:           orgID,
 			}, nil
 		case influxdb.MonitoringSystemBucketName:
+			fmt.Println(n)
```
jademcgough: remove this :)

PR #15548, opened 2019-10-23
feat(ui): choose the midpoint of the data when creating an alert

### LINKED ISSUE #15187: Set threshold values to the midpoint of yDomain at initialization
Currently thresholds are initialized to 20, which is totally arbitrary, midpoint of yDomain makes more sense

drdelambre:
Closes #15187

Giraffe is exposing the range of the vertical axis through redux + memo hooks. This PR routes these events to the alert buttons so that we can default the alert to the middle of the dataset. The output is a meaningful default value that looks nice.


drdelambre:
working on how to write tests around this

ui/src/shared/utils/vis.ts
```diff
@@ -120,6 +120,10 @@ export const parseBounds = (
 }
 
 export const extent = (xs: number[]): [number, number] | null => {
+  if (!xs || !xs.length) {
```
ebb-tide: what does this do?

drdelambre: even though extent is always supposed to be passed a number array, it doesn't when an empty state is loaded from redux. i thought about guarding it at the redux retrieval with a default value, but it seemed that if could happen there, it could probably happen from anywhere and put the guard as low as possible (where it breaks).

PR #15549, opened 2019-10-23
Fix/tasks edit

### LINKED ISSUE #15534: Fetching task gives `Uncaught TypeError: Cannot destructure property `name` of 'undefined' or 'null'.`
This happens in master commit: 804f0b012596ed36be102d7c739dc5ab54acecc3

when i try to view a task that was created a few weeks ago

image

Asalem1:
Closes #15534 

### Problem

The page would crash whenever an existing task was selected to edit. In addition, data on the task form would not populate or persist whenever a task was selected to edit / was toggled between task

### Solution

The app was originally failing because of a race condition where the `currentTask` that was needed to call the `setAllTaskOptions` was only being populate after the componentDidMount function `selectTaskByID` was triggered. Since they were both being triggered in the CDM lifecycle, the `setAllTaskOptions` would never actually populate the data. To solve this, I created a new `setAllTaskOptionsByID` to get the task data based on the `id`. 

In doing so, this fixed our issue of data persistence in the `TaskForm`. To resolve the issue of data persistence when toggling between `schedule tasks` simply required a modification to the following reducer:

https://github.com/influxdata/influxdb/blob/master/ui/src/tasks/reducers/index.ts#L94

It seems like the original intent of this conditional was to prevent data persistence from mucking up the task (i.e. storing interval data and cron data may have set the data for both), but these fears seem unfounded since the data is set correctly without any unintended consequences.

tasks_test

### DIFF
--- a/ui/cypress/e2e/tasks.test.ts
+++ b/ui/cypress/e2e/tasks.test.ts
@@ -19,7 +19,7 @@ describe('Tasks', () => {
     })
 
     cy.fixture('routes').then(({orgs}) => {
-      cy.get<Organization>('@org').then(({id}) => {
+      cy.get('@org').then(({id}: Organization) => {
         cy.visit(`${orgs}/${id}/tasks`)
       })
     })
@@ -60,8 +60,8 @@ from(bucket: "${name}")
       .and('contain', taskName)
   })
 
-  it.only('can delete a task', () => {
-    cy.get<Organization>('@org').then(({id}) => {
+  it('can delete a task', () => {
+    cy.get('@org').then(({id}: Organization) => {
       cy.get<string>('@token').then(token => {
         cy.createTask(token, id)
         cy.createTask(token, id)
@@ -86,7 +86,7 @@ from(bucket: "${name}")
   })
 
   it('can disable a task', () => {
-    cy.get<Organization>('@org').then(({id}) => {
+    cy.get('@org').then(({id}: Organization) => {
       cy.get<string>('@token').then(token => {
         cy.createTask(token, id)
       })
@@ -100,7 +100,7 @@ from(bucket: "${name}")
   })
 
   it('can edit a tasks name', () => {
-    cy.get<Organization>('@org').then(({id}) => {
+    cy.get('@org').then(({id}: Organization) => {
       cy.get<string>('@token').then(token => {
         cy.createTask(token, id)
       })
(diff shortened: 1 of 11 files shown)


ui/cypress/e2e/tasks.test.ts
```diff
@@ -178,18 +178,102 @@ from(bucket: "${name}")
       cy.getByTestID('task-card').should('have.length', 1)
     })
   })
+
+  describe('update & persist data', () => {
+    // address a bug that was reported when editing tasks:
+    // https://github.com/influxdata/influxdb/issues/15534
```
hoorayimhelping: ❤️

ui/src/tasks/reducers/tasks.test.ts
```diff
@@ -5,7 +5,8 @@ import tasksReducer, {
 import {setTaskOption} from 'src/tasks/actions'
 import {TaskSchedule} from 'src/utils/taskOptionsToFluxScript'
 
-describe('tasksReducer', () => {
+// skipping this test since the cron and interval values should not be reset when toggling between schedule tasks
+describe.skip('tasksReducer', () => {
```
hoorayimhelping: if this asserts something that is no longer the case, or it represents outdated assumptions, i'd say just remove the whole assertion. i think keeping it around will be more confusing than helpful. i appreciate the hesitation to delete an entire file someone else wrote though :)

if we're not sure what this is supposed to do, let's get some clarification from russ.

what do you think?

Asalem1: I appreciate the sympathy - definitely didn't wanna go around deleting files and taking names my second week, but it doesn't seem applicable. The only reason why I'd want to keep it would be to have a basis for fleshing out tests on the reducer later on. I'll go ahead and fix the tests so that they're consistent with the new logic of not overwriting the previous values

ui/src/tasks/reducers/tasks.test.ts
```diff
@@ -5,11 +5,13 @@ import tasksReducer, {
 import {setTaskOption} from 'src/tasks/actions'
 import {TaskSchedule} from 'src/utils/taskOptionsToFluxScript'
 
+// skipping this test since the cron and interval values should not be reset when toggling between schedule tasks
```
hoorayimhelping: ✂️ this comment 😉

Asalem1: thanks, good eyes!

PR #15556, opened 2019-10-23
fix(ui/checks): checks render once they've been created

### LINKED ISSUE #15554: Created checks don't display on Monitoring & Alerts page
When creating a new alert where none exist, the alert will not display. A console error appears:

Screen Shot 2019-10-23 at 12 17 57 PM

To reproduce the error, simply create a new alert:

checks_err

Blocking Issue #15518

Asalem1:
Closes #15554 

### Problem

Newly created checks without labels were not rendering on the checks list.

Steps for reproduction can be found on #15554 

### Solution

Ensure that check labels exist before mapping the values. Added testing to ensure that the checks list page should render a check once it has been created

### DIFF
--- a/ui/cypress/e2e/checks.test.ts
+++ b/ui/cypress/e2e/checks.test.ts
@@ -56,4 +56,40 @@ describe('Checks', () => {
       })
     })
   })
+
+  describe('Check should be viewable once created', () => {
+    beforeEach(() => {
+      // creates a check before each iteration
+      // TODO: refactor into a request
+      cy.getByTestID('create-check').click()
+      cy.getByTestID('create-threshold-check').click()
+      cy.getByTestID(`selector-list ${measurement}`).click()
+      cy.getByTestID('save-cell--button').should('be.disabled')
+      cy.getByTestID(`selector-list ${field}`).click()
+      cy.getByTestID('save-cell--button').should('be.disabled')
+      cy.getByTestID('checkeo--header alerting-tab').click()
+      cy.getByTestID('add-threshold-condition-WARN').click()
+      cy.getByTestID('save-cell--button').click()
+      cy.getByTestID('check-card').should('have.length', 1)
+    })
+
+    it('should allow created checks to be selected and routed to the edit page', () => {
+      cy.getByTestID('check-card--name').should('have.length', 1)
+      cy.getByTestID('check-card--name').click()
+      cy.get('@org').then(({id}: Organization) => {
+        cy.fixture('routes').then(({orgs, alerting, checks}) => {
+          cy.url().then(url => {
+            Cypress.minimatch(
+              url,
+              `
+                ${
+                  Cypress.config().baseUrl
+                }${orgs}/${id}${alerting}${checks}/*/edit
+              `
+            )
+          })
+        })
+      })
+    })
+  })
 })
--- a/ui/src/alerting/actions/checks.ts
+++ b/ui/src/alerting/actions/checks.ts
@@ -1,6 +1,7 @@
 // Libraries
 import {Dispatch} from 'react'
 import {push} from 'react-router-redux'
+import {get} from 'lodash'
 
 // Constants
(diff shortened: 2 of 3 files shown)


ui/src/alerting/actions/checks.ts
```diff
@@ -155,7 +155,12 @@ export const saveCheckFromTimeMachine = () => async (
     ...check,
     query: draftQueries[0],
     orgID,
-    labels: check.labels.map(l => l.id),
+    labels:
+      (check &&
+        check.labels &&
+        Array.isArray(check.labels) &&
+        check.labels.map(l => l.id)) ||
+      [],
```
hoorayimhelping: if this is too nitpicky, tell me to stfu and i'll gladly do it! 😃 

from what i've seen in the codebase, i think the more idiomatic way to do this is something like:

```js
import {get} from 'lodash'

//...

const labels = get(check, 'labels', [])

const data = {
  ...check,
  name: clonedName,
  labels: labels.map(l => l.id),
} as PostCheck
```
lodash's [`get`](https://lodash.com/docs/4.17.15#get) will check the first argument (`check`) for the second argument (`'labels'`). it'll do the falsey checks you're doing as it goes down the property list. if it can't find the property you're looking for, it returns the third property - an empty array.

(it's not a huge deal for a single property object, but it's super valuable when there are deeply nested properties, like [here,](https://github.com/influxdata/influxdb/blob/master/ui/src/dashboards/selectors/index.ts#L12)).

using `get` like above, it'll guard against falsey values (`null`, `undefined`, `false`, `''`, etc), and includes the `) || []` that ensures labels stays an array. but it won't guard against truthy values, so if `labels` is anything but an array, like a string or an object, this will fail. but if you search `src/client/generatedRoutes` for `export type Labels`, you'll see it's typed as an array `Label[]`, so i think this check is safe.

the reason i bring it up is i think it's probably a little bit safer and less likely to get tripped up on an edge case, and because we use this pattern (using `get`) in quite a few places i've seen in the code base, it's fairly well understood.

ebb-tide: I ❤️ `get` too

ui/cypress/e2e/checks.test.ts
```diff
@@ -56,4 +56,40 @@ describe('Checks', () => {
...
+      cy.getByTestID('create-check').click()
+      cy.getByTestID('create-threshold-check').click()
+      cy.getByTestID(`selector-list ${measurement}`).click()
+      cy.getByTestID('save-cell--button').should('be.disabled')
+      cy.getByTestID(`selector-list ${field}`).click()
+      cy.getByTestID('save-cell--button').should('be.disabled')
+      cy.getByTestID('checkeo--header alerting-tab').click()
+      cy.getByTestID('add-threshold-condition-WARN').click()
+      cy.getByTestID('save-cell--button').click()
+      cy.getByTestID('check-card').should('have.length', 1)
```
ebb-tide: ❤️ for testing!

ui/src/alerting/actions/checks.ts
```diff
@@ -151,11 +152,12 @@ export const saveCheckFromTimeMachine = () => async (
     alerting: {check},
   } = getActiveTimeMachine(state)
 
+  const labels = get(check, 'labels', []) as Label[]
```
ebb-tide:
```suggestion
get<Check, 'labels', Label[]>(check, 'labels', [])
```

ui/src/alerting/actions/checks.ts
```diff
@@ -253,10 +255,11 @@ export const cloneCheck = (check: Check) => async (
     const allCheckNames = list.map(c => c.name)
 
     const clonedName = incrementCloneName(allCheckNames, check.name)
+    const labels = get(check, 'labels', []) as Label[]
```
ebb-tide:
```suggestion
get<Check, 'labels', Label[]>(check, 'labels', [])
```

PR #15557, opened 2019-10-23
fix(query/stdlib): fix rewritten pushable expressions when a logical expression is present

jsternberg:
When `exists` was used in conjunction with any other pushed down
expression, the `exists` was not rewritten properly because the rewrite
did not descend into logical expressions.

This is now fixed so those expressions will be rewritten correctly. This
affected the following form:

    filter(fn: (r) => r._measurement == "cpu" and exists r.host)

It did not affect the following:

    filter(fn: (r) => r._measurement == "cpu")
    |> filter(fn: (r) => exists r.host)

### DIFF
--- a/query/stdlib/influxdata/influxdb/rules.go
+++ b/query/stdlib/influxdata/influxdb/rules.go
@@ -437,6 +437,15 @@ func rewritePushableExpr(e semantic.Expression) (semantic.Expression, bool) {
 			e.Left, e.Right = left, right
 			return e, true
 		}
+
+	case *semantic.LogicalExpression:
+		left, lok := rewritePushableExpr(e.Left)
+		right, rok := rewritePushableExpr(e.Right)
+		if lok || rok {
+			e = e.Copy().(*semantic.LogicalExpression)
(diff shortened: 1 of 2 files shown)


query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -437,6 +437,15 @@ func rewritePushableExpr(e semantic.Expression) (semantic.Expression, bool) {
...
 		}
+
+	case *semantic.LogicalExpression:
+		left, lok := rewritePushableExpr(e.Left)
+		right, rok := rewritePushableExpr(e.Right)
+		if lok || rok {
+			e = e.Copy().(*semantic.LogicalExpression)
+			e.Left, e.Right = left, right
+			return e, true
+		}
```
jlapacik: `rewritePushableExpr` is used for rewriting `exists` into something storage understands correct? In other words, this has no other side effects right?

jsternberg: Correct.

PR #15558, opened 2019-10-23
feat(backup): backup data and metadata

### LINKED ISSUE #15603: storage 2.0: implement data backup

### LINKED ISSUE #15604: storage 2.0: implement meta data backup

jacobmarble:
Fixes #15603 
Fixes #15604 
Currently blocked on #15961

This command backs up data and metadata:
```
$ influx backup --path backup_dir
```

### DIFF
--- a/backup.go
+++ b/backup.go
@@ -1,0 +1,23 @@
+package influxdb
+
+import (
+	"context"
+	"io"
+)
+
+// BackupService represents the data backup functions of InfluxDB.
+type BackupService interface {
+	// CreateBackup creates a local copy (hard links) of the TSM data for all orgs and buckets.
+	// The return values are used to download each backup file.
+	CreateBackup(context.Context) (backupID int, backupFiles []string, err error)
+	// FetchBackupFile downloads one backup file, data or metadata.
+	FetchBackupFile(ctx context.Context, backupID int, backupFile string, w io.Writer) error
+	// InternalBackupPath is a utility to determine the on-disk location of a backup fileset.
+	InternalBackupPath(backupID int) string
+}
+
+// KVBackupService represents the meta data backup functions of InfluxDB.
+type KVBackupService interface {
+	// Backup creates a live backup copy of the metadata database.
+	Backup(ctx context.Context, w io.Writer) error
+}
--- a/cmd/influx/backup.go
+++ b/cmd/influx/backup.go
@@ -1,0 +1,97 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/bolt"
+	"github.com/influxdata/influxdb/http"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"go.uber.org/multierr"
+)
+
+var backupCmd = &cobra.Command{
+	Use:   "backup",
+	Short: "Backup the data in InfluxDB",
+	Long: fmt.Sprintf(
+		`Backs up data and meta data for the running InfluxDB instance.
+Downloaded files are written to the directory indicated by --path.
+The target directory, and any parent directories, are created automatically.
+Data file have extension .tsm; meta data is written to %s in the same directory.`,
+		bolt.DefaultFilename),
+	RunE: backupF,
+}
+
+var backupFlags struct {
+	Path string
+}
+
+func init() {
+	backupCmd.PersistentFlags().StringVarP(&backupFlags.Path, "path", "p", "", "directory path to write backup files to")
+	err := viper.BindEnv("PATH")
+	if err != nil {
+		panic(err)
+	}
+	if h := viper.GetString("PATH"); h != "" {
+		backupFlags.Path = h
+	}
+}
+
+func newBackupService(flags Flags) (influxdb.BackupService, error) {
+	return &http.BackupService{
+		Addr:  flags.host,
+		Token: flags.token,
+	}, nil
+}
+
+func backupF(cmd *cobra.Command, args []string) error {
+	ctx := context.Background()
+
+	if flags.local {
+		return fmt.Errorf("local flag not supported for backup command")
+	}
+
+	if backupFlags.Path == "" {
+		return fmt.Errorf("must specify path")
+	}
+
+	err := os.MkdirAll(backupFlags.Path, 0770)
+	if err != nil && !os.IsExist(err) {
+		return err
+	}
+
+	backupService, err := newBackupService(flags)
+	if err != nil {
+		return err
+	}
+
+	id, backupFilenames, err := backupService.CreateBackup(ctx)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Backup ID %d contains %d files\n", id, len(backupFilenames))
+
+	for _, backupFilename := range backupFilenames {
+		dest := filepath.Join(backupFlags.Path, backupFilename)
+		w, err := os.OpenFile(dest, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0660)
+		if err != nil {
+			return err
+		}
+		err = backupService.FetchBackupFile(ctx, id, backupFilename, w)
+		if err != nil {
+			return multierr.Append(err, w.Close())
+		}
+		if err = w.Close(); err != nil {
+			return err
+		}
+	}
+
+	fmt.Printf("Backup complete")
+
+	return nil
+}
--- a/http/backup_service.go
+++ b/http/backup_service.go
@@ -1,0 +1,226 @@
+package http
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path"
+	"path/filepath"
+	"strconv"
+	"time"
+
+	"github.com/influxdata/httprouter"
+	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/bolt"
+	"github.com/influxdata/influxdb/kit/tracing"
+	"go.uber.org/multierr"
+	"go.uber.org/zap"
+)
+
+// BackupBackend is all services and associated parameters required to construct the BackupHandler.
+type BackupBackend struct {
+	Logger *zap.Logger
+	influxdb.HTTPErrorHandler
+
+	BackupService   influxdb.BackupService
+	KVBackupService influxdb.KVBackupService
+}
+
+// NewBackupBackend returns a new instance of BackupBackend.
+func NewBackupBackend(b *APIBackend) *BackupBackend {
+	return &BackupBackend{
+		Logger: b.Logger.With(zap.String("handler", "backup")),
+
+		HTTPErrorHandler: b.HTTPErrorHandler,
+		BackupService:    b.BackupService,
+		KVBackupService:  b.KVBackupService,
+	}
+}
+
+type BackupHandler struct {
+	*httprouter.Router
+	influxdb.HTTPErrorHandler
+	Logger *zap.Logger
+
+	BackupService   influxdb.BackupService
+	KVBackupService influxdb.KVBackupService
+}
+
+const (
+	prefixBackup        = "/api/v2/backup"
+	backupIDParamName   = "backup_id"
+	backupFileParamName = "backup_file"
+	backupFilePath      = prefixBackup + "/:" + backupIDParamName + "/file/:" + backupFileParamName
+
+	httpClientTimeout = time.Hour
+)
+
+func composeBackupFilePath(backupID int, backupFile string) string {
+	return path.Join(prefixBackup, fmt.Sprint(backupID), "file", fmt.Sprint(backupFile))
+}
+
+// NewBackupHandler creates a new handler at /api/v2/backup to receive backup requests.
+func NewBackupHandler(b *BackupBackend) *BackupHandler {
+	h := &BackupHandler{
+		HTTPErrorHandler: b.HTTPErrorHandler,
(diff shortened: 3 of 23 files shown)


jacobmarble:
@stephanie-engel data and meta data should backup properly now.

TODO(jacobmarble): auth checks should block any token without permission to read all data and all metadata

http/backup_service.go
```diff
@@ -0,0 +1,223 @@
...
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if err := CheckError(resp); err != nil {
+		return err
+	}
+
+	_, err = io.CopyBuffer(w, resp.Body, make([]byte, 1024*1024))
```
e-dard: Did you test that this was any better than just wrapping `w` in a buffered writer (`bufio.Writer`)? It probably is; I just wondered. I would probably make it a constant.

http/backup_service.go
```diff
@@ -0,0 +1,223 @@
...
+	if err != nil {
```
e-dard: Just `return err`

storage/engine.go
```diff
@@ -662,6 +667,76 @@ func (e *Engine) deleteBucketRangeLocked(ctx context.Context, orgID, bucketID pl
 	return e.engine.DeletePrefixRange(ctx, name, min, max, pred)
 }
 
+func (e *Engine) CreateBackup(ctx context.Context) (int, []string, error) {
+	span, ctx := tracing.StartSpanFromContext(ctx)
+	span.LogKV("path", e.path)
+
+	err := e.engine.WriteSnapshot(ctx, tsm1.CacheStatusBackup)
```
e-dard: nit: you could combine this `err` and `if`

storage/engine.go
```diff
@@ -662,6 +667,76 @@ func (e *Engine) deleteBucketRangeLocked(ctx context.Context, orgID, bucketID pl
 	return e.engine.DeletePrefixRange(ctx, name, min, max, pred)
 }
 
+func (e *Engine) CreateBackup(ctx context.Context) (int, []string, error) {
```
e-dard: Do you need to check if the engine is shutting down?

storage/engine.go
```diff
@@ -662,6 +667,76 @@ func (e *Engine) deleteBucketRangeLocked(ctx context.Context, orgID, bucketID pl
...
+	span, ctx := tracing.StartSpanFromContext(ctx)
+	span.LogKV("path", e.path)
+
+	backupPath := e.engine.FileStore.InternalBackupPath(backupID)
+	if fi, err := os.Stat(backupPath); err != nil {
+		if os.IsNotExist(err) {
+			return errors.Errorf("backup %d not found", backupID)
+		}
+		return errors.WithMessagef(err, "failed to locate backup %d", backupID)
+	} else {
```
e-dard: Since all paths in the `if` lead to a `return` this `else` is redundant. Either make it an `else if` to save a line, or just have another `if` block.

storage/engine.go
```diff
@@ -662,6 +667,76 @@ func (e *Engine) deleteBucketRangeLocked(ctx context.Context, orgID, bucketID pl
...
+	file, err := os.Open(backupFileFullPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return errors.Errorf("backup file %d/%s not found", backupID, backupFile)
+		}
+		return errors.WithMessagef(err, "failed to open backup file %d/%s", backupID, backupFile)
+	}
+	defer file.Close()
+
+	buf := make([]byte, 1024*1024)
```
e-dard: maybe make the buffer a constant?

storage/engine.go
```diff
@@ -662,6 +667,76 @@ func (e *Engine) deleteBucketRangeLocked(ctx context.Context, orgID, bucketID pl
...
+	_, err = io.CopyBuffer(w, file, buf)
+	if err != nil {
+		return errors.WithMessagef(err, "failed to copy backup file %d/%s to writer", backupID, backupFile)
+	}
+
+	err = syscall.Unlink(backupFileFullPath)
```
e-dard: I think this is the first time I have seen `syscall.Unlink` in the Influx codebase. At least outside of the context of a socket. I expect this is because the file is linked when being created in the path, but I don't recall previous backup/restore APIs in influx unlinking, so i'm interested :-)

storage/engine.go
```diff
@@ -662,6 +667,76 @@ func (e *Engine) deleteBucketRangeLocked(ctx context.Context, orgID, bucketID pl
...
+	if err != nil {
```
e-dard: `err` and `if` can be combined

storage/engine.go
```diff
@@ -662,6 +667,76 @@ func (e *Engine) deleteBucketRangeLocked(ctx context.Context, orgID, bucketID pl
...
+		e.logger.Info("failed to unlink backup file after fetch", zap.Error(err), zap.Int("backup_id", backupID), zap.String("backup_file", backupFile))
```
e-dard: `Failed`

storage/engine.go
```diff
@@ -662,6 +667,76 @@ func (e *Engine) deleteBucketRangeLocked(ctx context.Context, orgID, bucketID pl
...
+	}
+
+	return nil
```
e-dard: Did you intend to not return the `Unlink` error?

storage/engine.go
```diff
@@ -662,6 +667,76 @@ func (e *Engine) deleteBucketRangeLocked(ctx context.Context, orgID, bucketID pl
...
+	}
+	filenames := make([]string, len(fileInfos))
+	for i, fi := range fileInfos {
+		filenames[i] = fi.Name()
+	}
+
+	return id, filenames, nil
+}
+
+func (e *Engine) FetchBackupFile(ctx context.Context, backupID int, backupFile string, w io.Writer) error {
```
e-dard: Could you comment the exported methods on the Engine?

tsdb/tsm1/engine.go
```diff
@@ -15,6 +15,10 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/influxdata/influxql"
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/zap"
+
```
e-dard: blank line

tsdb/tsm1/file_store.go
```diff
@@ -1208,7 +1208,7 @@ func (f *FileStore) locations(key []byte, t int64, ascending bool) []*location {
 
 // CreateSnapshot creates hardlinks for all tsm and tombstone files
 // in the path provided.
-func (f *FileStore) CreateSnapshot(ctx context.Context) (string, error) {
+func (f *FileStore) CreateSnapshot(ctx context.Context) (int, string, error) {
```
dgnorton: Would named returns be better, to document what's being returned?

cmd/influx/backup.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/spf13/viper"
+	"go.uber.org/multierr"
+	"os"
+	"path/filepath"
+)
+
+var backupCmd = &cobra.Command{
+	Use:   "backup",
+	Short: "Backup the data in InfluxDB",
+	Long:  `Backs up data and meta data for the InfluxDB instance. OSS only.`,
```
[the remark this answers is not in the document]
e-dard: @dgnorton I don't think that a CLI tool would be the UI used to do backups in Cloud2. Hence `OSS only` seems a bit redundant.

backup.go
```diff
@@ -0,0 +1,16 @@
+package influxdb
+
+import (
+	"context"
+	"io"
+)
+
+type BackupService interface {
+	CreateBackup(context.Context) (int, []string, error)
```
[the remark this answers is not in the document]
jacobmarble: Yes, I tried to go that route.

The two are quite different. `BackupService.CreateBackup` creates hard links, returns a unique ID and list of files. `KVBackupService` writes to a writer, which is hooked up to a file (not hard link) in the directory that `BackupService.CreateBackup` created.

cmd/influx/backup.go
```diff
@@ -0,0 +1,90 @@
...
+	RunE:  backupF,
+}
+
+var backupFlags struct {
+	Path string
+}
+
+func init() {
+	backupCmd.PersistentFlags().StringVarP(&backupFlags.Path, "path", "p", "", "directory path to write backup files to")
+	err := viper.BindEnv("PATH")
```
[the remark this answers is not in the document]
jacobmarble: I agree there are other ways to do this, but I'm following an established pattern:
https://github.com/influxdata/influxdb/blob/f8dee7151658d6fe2ca60f70f28e72e8c09762b0/cmd/influx/delete.go#L25-L28

http/backup_service.go
```diff
@@ -0,0 +1,223 @@
...
+	ID    int      `json:"id,omitempty"`
+	Files []string `json:"files,omitempty"`
+}
+
+func (h *BackupHandler) handleCreate(w http.ResponseWriter, r *http.Request) {
+	span, r := tracing.ExtractFromHTTPRequest(r, "BackupHandler.handleCreate")
+	defer span.Finish()
+
+	ctx := r.Context()
+	defer r.Body.Close()
```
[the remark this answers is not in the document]
jacobmarble: Don't know why this is hard-coded in my brain.

http/backup_service.go
```diff
@@ -0,0 +1,223 @@
...
+
+	return h
+}
+
+type backup struct {
+	ID    int      `json:"id,omitempty"`
+	Files []string `json:"files,omitempty"`
+}
+
+func (h *BackupHandler) handleCreate(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
jacobmarble: How about 1h? The overall backup operation might take longer than that, but each request is either (1) downloading one TSM file (2) downloading one BoltDB file or (3) creating hard links for each TSM file.

cmd/influx/backup.go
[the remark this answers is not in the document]
e-dard: What’s the full name of the env var though? `PATH`?

storage/engine.go
jacobmarble: `os.Remove` removes either a file or a directory. I know this is a file, so I shortcut to `syscall.Unlink`. I just confirmed that we don't use `syscall.Unlink` anywhere, and I don't care to defend this, so `os.Remove` it is.

For the curious:
```golang
package os

func Remove(name string) error {
	// System call interface forces us to know
	// whether name is a file or directory.
	// Try both: it is cheaper on average than
	// doing a Stat plus the right one.
	e := syscall.Unlink(name)
	if e == nil {
		return nil
	}
	e1 := syscall.Rmdir(name)
	if e1 == nil {
		return nil
	}

	// Both failed: figure out which error to return.
	// OS X and Linux differ on whether unlink(dir)
	// returns EISDIR, so can't use that. However,
	// both agree that rmdir(file) returns ENOTDIR,
	// so we can use that to decide which error is real.
	// Rmdir might also return ENOTDIR if given a bad
	// file path, like /etc/passwd/foo, but in that case,
	// both errors will be ENOTDIR, so it's okay to
	// use the error from unlink.
	if e1 != syscall.ENOTDIR {
		e = e1
	}
	return &PathError{"remove", name, e}
}
```

storage/engine.go
jacobmarble: I do not intend to fail the download because of an internal error. The link will be removed eventually.

http/backup_service.go
jacobmarble: That violates a pattern that we have been moving towards (I thought?). The pattern "try to end with `return nil`".

e-dard: That’s a new pattern for me; what’s the rationale?

backup.go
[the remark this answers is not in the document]
jacobmarble: The signatures are very different. `BackupService.CreateBackup` creates hard links in a directory that it figures out. `KVBackupService.Backup` takes a writer, which should point to a file in the directory indicated by `BackupService.CreateBackup`. I tried.

cmd/influx/backup.go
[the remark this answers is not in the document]
jacobmarble: Added a longer description.

If the user does try to backup from their cloud account, they will probably just get a 404.

cmd/influx/backup.go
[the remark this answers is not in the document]
jacobmarble: This follows the pattern here:
https://github.com/influxdata/influxdb/blob/0664dbc47dcc43a8b2ac619fa8d5dfcfa3b96e1d/cmd/influx/query.go#L32-L41

http/backup_service.go
jacobmarble: I did some reading.

Turns out that `io.CopyBuffer` doesn't use the `buf` parameter when the writer implements `io.WriterTo` or the reader implements `io.ReaderFrom`. `os.File` implements both. In this case (API client), the writer is always a file. On the service side, also in this PR, the reader is always a file. Therefore, I've replaced both instances with `io.Copy`, which ultimately does the same thing.

https://github.com/golang/go/issues/32276

storage/engine.go
jacobmarble: See my comment above, which starts with "I did some reading"

jacobmarble:
This PR should not be merged until @russorat and I settle on a better authorization strategy.

russorat:
this cannot be merged until we fix the issues with auth. using a master token gives the error: `Error: read:notificationRules is unauthorized` and using an All Access token gives the error: `Error: read:authorizations is unauthorized`.

jacobmarble:
Squashed and rebased to master.

jacobmarble:
Replaced by #16504

PR #15574, opened 2019-10-24
feat(pkger): add influx packages foundation

jsteenb2:
Inception of influx packages (`pkger`). Starting with buckets here. Supports JSON/YAML formats and has a CLI command to do the business. 

What is packages and what does it solve?

Users would like to be able to create resources from a declarative template. From a maintainer's perspective, we'd like the ability to maintain and extend the functionality as demands grow. Packages provides a way for users to create a declarative config file and ship that to the API and the backend will stand up all the resources described in that package config file. If any issues arise all new state is rolled back to the state before the package were to be applied. 

An example yml file for a package: https://github.com/influxdata/influxdb/blob/40e5ef69a6c39594dba184afa5de9733644433d9/pkger/testdata/label.yml

(can use JSON/YML encoding)

To run this you can run the `influx` cli tool and run the following command:

```
influx pkg --path=PATH_TO_PACKAGE --org-id=$ORGANIZATION_ID
```

and you should see a representation of what will be applied with a confirm step to proceed or exit without applying the package. Confirm and all the resources will be stood up for you. Any errors that occur will be printed in the CLI output.

- [X] Rebased/mergeable
- [x] Tests pass


jsteenb2:
tests failing in e2e 🤷‍♂

e-dard:
@jsteenb2 hi 👋 there is a lot of code here, but less of an explanation of what any of it's for 😄 .

Is there an open issue you can link to? If not, could you provide background information in the PR description? What problem are you solving here and what does that mean for our engineers or our users? Thanks!

jsteenb2:
@e-dard great point. I updated the description with some TLDR around the pkg feature. Holler if there is more info I can provide

jademcgough:
@e-dard this is the basis for composable solutions / apps / whatever we're calling it this week. The goal is for users to be able to install shared packages over the CLI or UI, which will greatly simplify onboarding for new users.

PR #15580, opened 2019-10-25
fix(restore):  restore bugfix and import feedback

z1q1q7:
I find the bug and some can be better when I use influxd restore and influx import tool.
- influxd restore: Any shard that ends with "0" will restore to wrong shard.
- influxd restore: Continuous queries have not restored.
- influxd import: When import multi databases or rps, the feedback have not given.

- [x] Tests pass


z1q1q7:
Sementic Pull Request always Pending.

PR #15589, opened 2019-10-25
fix(ui): allow hmr feature with proxied ingress controllers

drdelambre:
This allows you to pass a root url into the dev server, for when all else fails and you need to hard code some paths. currently, this is used by the hot module replacement mechanism within webpack development server to define the base path of the websocket server from a client's perspective. now we can hide a dev server, running at `http://localhost:9999`, behind the url `https://secure.local:8000`, and HMR should still work. Check the changes in monitor-ci for configuring the ingress controller to pass the correct headers.

### DIFF
--- a/ui/webpack.dev.ts
+++ b/ui/webpack.dev.ts
@@ -21,14 +21,10 @@ module.exports = merge(common, {
       '/api/v2': 'http://localhost:9999',
       '/debug/flush': 'http://localhost:9999',
     },
-    watchOptions: {
-      aggregateTimeout: 300,
-      poll: 1000
-    },
     disableHostCheck: true,
     host: '0.0.0.0',
     port: PORT,
-    public: PUBLIC
+    public: PUBLIC,
   },
   module: {
     rules: [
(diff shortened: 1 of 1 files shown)


ui/webpack.dev.ts
```diff
@@ -5,6 +5,7 @@ const webpack = require('webpack')
 const common = require('./webpack.common.ts')
 const path = require('path')
 const PORT = parseInt(process.env.PORT, 10) || 8080
+const PUBLIC = process.env.PUBLIC || undefined
```
Asalem1: I'd be curious to know how `webpack` handles `undefined` property values. Also, if `process.env.PUBLIC` value isn't passed in, wouldn't it be `undefined` by default?

drdelambre: best case scenario is that the object isn't assigned ANYTHING, but that makes the code hard to read (lots of if(condition){ out.prop = 'whatever' }), and undefined is less truthy than an empty string by 🔬 <- that much.
it handles the undefined value quite well in that it still serves files.

Issue #15239, opened 2019-09-23
Dashboard cell contents disappear after canceled edit, require page refresh to view


glinton:
__Steps to reproduce:__

1. Open dashboard in cloud2
1. Select "Configure" from blue pencil icon in dashboard cell (graph, etc...)
1. Choose the "X" in the top right (next to green checkmark)
1. Observe missing cell contents

__Expected behavior:__
Cell contents, same behavior as selecting the green check in cell configuration mode.

__Actual behavior:__
Contents disappear until a page refresh is executed.

__Environment info:__

* System info: `Linux 5.0.0-27-generic x86_64`
* InfluxDB version: ?? `cloud2.influxdata.com`
* Other relevant environment details: `Chrome Version 77.0.3865.75 (Official Build) (64-bit)`

desa:
@russorat is currently unable to reliably reproduce.

PR #15590, opened 2019-10-25
test(ui): add e2e regression test for issue #15239

hoorayimhelping:
Connects #15239

Adds e2e regression test for above ticket ☝️ 

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/cypress/e2e/dashboardsView.test.ts
+++ b/ui/cypress/e2e/dashboardsView.test.ts
@@ -21,7 +21,7 @@ describe('Dashboard', () => {
     })
   })
 
-  it('can edit a dashboards name', () => {
+  it("can edit a dashboard's name", () => {
     cy.get('@org').then(({id: orgID}: Organization) => {
       cy.createDashboard(orgID).then(({body}) => {
         cy.fixture('routes').then(({orgs}) => {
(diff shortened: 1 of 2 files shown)


ui/cypress/e2e/dashboardsView.test.ts
```diff
@@ -21,7 +21,7 @@ describe('Dashboard', () => {
     })
   })
 
-  it('can edit a dashboards name', () => {
+  it("can edit a dashboard's name", () => {
```
Asalem1: most excellent

desa: double quotes on purpose?

ui/cypress/e2e/dashboardsView.test.ts
```diff
@@ -60,6 +60,39 @@ describe('Dashboard', () => {
...
+
+    // Add an empty celly cell
+    cy.getByTestID('add-cell--button').click()
+    cy.getByTestID('save-cell--button').click()
+    cy.getByTestID('cell--view-empty').should('be.visible')
+
+    let cellContent
+    cy.getByTestID('cell--view-empty')
+      .invoke('text')
+      .then(cellContent => {
```
Asalem1: @hoorayimhelping , why're we calling the parameter that's being passed in the `then` the same thing as the variable that's being instantiated on line 78:

https://github.com/influxdata/influxdb/pull/15590/files#diff-33b992b7a030a90957e222522b90a778R78

ui/cypress/e2e/dashboardsView.test.ts
hoorayimhelping: it used to be

>can edit a dashboards name

so i added the possessive apostrophe and the double quote so we wouldn't have to escape the apostrophe

ui/cypress/e2e/dashboardsView.test.ts
hoorayimhelping: oops, didn't mean to leave `let cellContent` in there. i bet lint will complain now. good catch!

PR #15613, opened 2019-10-28
fix(ui/timerangedropdown): refactored the daterangepicker implementation from modal to popover

### LINKED ISSUE #15420: Rebuild DateRangePicker using Clockface Popover
`DateRangePicker.tsx` located within `TimeRangeDropdown.tsx`

Looks like this one freely floats in the middle of the screen rather than being anchored to any particular element. `Popover` requires you to anchor the popover to an element, in this case it should be the `DropdownButton` within `TimeRangeDropdown`

Screen Shot 2019-10-15 at 9 39 14 AM

Asalem1:
Closes #15420 

### Problem

The current Data Explorer DateRangePicker was set in a modal that would render to a fixed location on the screen (middle)

### Solution

Refactored the way the component was being rendered so that it was relative to the dropdown and within a popover

datetimepopover

### DIFF
--- a/ui/src/shared/components/TimeRangeDropdown.tsx
+++ b/ui/src/shared/components/TimeRangeDropdown.tsx
@@ -4,7 +4,13 @@ import {get} from 'lodash'
 import moment from 'moment'
 
 // Components
-import {Dropdown} from '@influxdata/clockface'
+import {
+  Dropdown,
+  Popover,
+  PopoverPosition,
+  PopoverInteraction,
+  PopoverType,
+} from '@influxdata/clockface'
 import DateRangePicker from 'src/shared/components/dateRangePicker/DateRangePicker'
 
 // Constants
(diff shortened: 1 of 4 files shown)


ui/src/shared/components/TimeRangeDropdown.tsx
```diff
@@ -4,7 +4,13 @@ import {get} from 'lodash'
...
 
 // Components
-import {Dropdown} from '@influxdata/clockface'
+import {
+  Dropdown,
+  Popover,
+  PopoverPosition,
+  PopoverInteraction,
+  PopoverType,
+} from '@influxdata/clockface'
```
zoesteinkamp: Yay importing the popovers!

ui/src/shared/components/dateRangePicker/DateRangePicker.tsx
```diff
@@ -13,7 +13,13 @@ interface Props {
...
   onSetTimeRange: (timeRange: TimeRange) => void
   onClose: () => void
-  position?: {top?: number; right?: number; bottom?: number; left?: number}
+  position?: {
+    top?: number
+    right?: number
+    bottom?: number
+    left?: number
+    position?: string
+  }
```
zoesteinkamp: Improving that line count i see ;)

#new linting for the win

PR #15618, opened 2019-10-29
feat(tasks): add TaskHealthService to monitor active tasks

AlirieGray:
Epic: Improve SLOs

This PR adds a new metrics service called `taskLatency` which will provide a way to monitor the latency of all active tasks. It also adds a new filter option to `TaskFilter` called `active`, which allows us to filter tasks by Status. 

- [x] Rebased/mergeable
- [x] Tests pass


task_health_service.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"context"
+	"errors"
+	"time"
+)
+
+// FindTasksService implements the FindTasks method of a TaskService for the TaskHealthService to use
+type FindTasksService interface {
+	FindTasks(ctx context.Context, filter TaskFilter) ([]*Task, int, error)
+}
```
stuartcarnie: 👍 great approach to define the minimal interface your service requires

task_health_service.go
```diff
@@ -0,0 +1,65 @@
...
+
+// TaskHealthService is responsible for monitoring the latency of active tasks in the kv store
+type TaskHealthService struct {
```
stuartcarnie: I would recommend we limit the responsibility of this service to almost exactly what you state in the comment. We might consider being explicit about the name:

> `TaskLatencyMetricsService` is responsible for determining the run latency of active tasks.

I would expose two methods for this service:

```
UpdateMetrics(ctx context.Context) error
```

Which is responsible for calling find tasks and updating the metrics

and a function like to `storage.Engine#PrometheusCollectors`:

https://github.com/influxdata/influxdb/blob/8f6701d4b1a12c7a5dd2f0adc92ea5ab4967e5f8/storage/engine.go#L220-L228

To return the global metrics.

I would add a second API, `ExtendedPrometheusCollectors` to return the metrics with task_id labels, which will have much higher cardinality. We can add these collectors to a separate HTTP end point, like `/metrics/tasks/usage`, so they can be queried manually but will not be gathered by telegraf by default.

By limiting the concern of this type to gathering metrics of latencies, it will allow us to determine how best to incorporate the type in OSS and Cloud.

task_health_service.go
```diff
@@ -0,0 +1,65 @@
...
+	fts           FindTasksService
+	pollingPeriod time.Duration
+	ctx           context.Context
+	cancel        context.CancelFunc
+}
+
+// NewTaskHealthService takes a TaskHealthService and creates a new TaskHealthService
+func NewTaskHealthService(ctx context.Context, fts FindTasksService, p time.Duration) *TaskHealthService {
```
stuartcarnie: By limiting this to generating metrics, we would no longer require the `ctx` or `p` arguments

task_health_service.go
```diff
@@ -0,0 +1,65 @@
...
+				ticker.Stop()
+				return
+			}
+		}
+	}()
+}
+
+func (p *TaskHealthService) Close() {
+	p.cancel()
+}
```
stuartcarnie: These should be removed.

task_health_service.go
```diff
@@ -0,0 +1,65 @@
...
+
+// PollActiveTasks checks the KV store for Active Tasks on a regular interval scheduled by pollingPeriod
+func (p *TaskHealthService) PollActiveTasks() ([]*Task, error) {
```
stuartcarnie: This would be replaced with the `UpdateMetrics` method, which updates the internal prometheus metrics

kv/task.go
```diff
@@ -267,7 +267,7 @@ func (s *Service) findTasksByUser(ctx context.Context, tx Tx, filter influxdb.Ta
 			continue
 		}
 
-		if taskFilterMatch(filter.Type, task.Type) {
+		if taskTypeFilterMatch(filter.Type, task.Type) && taskStatusFilterMatch(filter.Active, task.Status) {
```
stuartcarnie: We can pair and explore whether there is an opportunity to simplify the `findTask` APIs to use the `newTaskMatchFn` as you have expanded this already.

PR #15642, opened 2019-10-29
Update flux version to v0.50.2

jsternberg:
This upgrades the flux version to v0.50.2.

The secret service, which is used for alerts, is not included. The
`to()` function is also still not included.

### DIFF
--- a/storage/reads/response_writer.gen.go
+++ b/storage/reads/response_writer.gen.go
@@ -57,39 +57,53 @@ func (w *ResponseWriter) streamFloatArrayPoints(cur cursors.FloatArrayCursor) {
 	frame := p.FloatPoints
 	w.res.Frames = append(w.res.Frames, datatypes.ReadResponse_Frame{Data: p})
 
-	var (
-		seriesValueCount = 0
-		b                = 0
-	)
-
+	var seriesValueCount = 0
 	for {
+		// If the number of values produced by cur > 1000,
+		// cur.Next() will produce batches of values that are of
+		// length ≤ 1000.
+		// We attempt to limit the frame Timestamps / Values lengths
+		// the same to avoid allocations. These frames are recycled
+		// after flushing so that on repeated use there should be enough space
+		// to append values from a into frame without additional allocations.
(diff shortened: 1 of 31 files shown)


storage/reads/response_writer.gen.go
```diff
@@ -413,39 +469,53 @@ func (w *ResponseWriter) streamBooleanArrayPoints(cur cursors.BooleanArrayCursor
...
-
+	var seriesValueCount = 0
 	for {
+		// If the number of values produced by cur > 1000,
+		// cur.Next() will produce batches of values that are of
+		// length ≤ 1000.
+		// We attempt to limit the frame Timestamps / Values lengths
+		// the same to avoid allocations. These frames are recycled
+		// after flushing so that on repeated use there should be enough space
+		// to append values from a into frame without additional allocations.
```
fchikwekwe: This makes sense. I'm not sure if I understand all of the memory implications of this change. WhatI perceive is that generally, you're limiting the possible size of each batch to reduce wasted excess allocations. Let me know if I'm missing the point here.

storage/reads/response_writer.gen.go
```diff
@@ -413,39 +469,53 @@ func (w *ResponseWriter) streamBooleanArrayPoints(cur cursors.BooleanArrayCursor
...
 			break
 		}
 
+		seriesValueCount += a.Len()
+		// As specified in the struct definition, w.sz is an estimated
+		// size (in bytes) of the buffered data. It is therefore a
+		// deliberate choice to accumulate using the array Size, which is
+		// cheap to calculate. Calling frame.Size() can be expensive
+		// when using varint encoding for numbers.
+		w.sz += a.Size()
```
fchikwekwe: Do you know why `frame.Size()` is more expensive to calculate?

jsternberg:
Sorry this is just a cherry pick of the 1.7 version. I didn't write a lot of the code as it was just synced over from 2.0.

fchikwekwe:
oh, oops.

PR #15646, opened 2019-10-29
feat(influxdb): Prevent the deletion of system buckets

### LINKED ISSUE #15630: protect system buckets from deletion
If a user attempts to delete a system bucket over the API, the request should be rejected. We need to continue to support internal deletions of system buckets, as cleanup when an organization is deleted.

bthesorceror:
Closes #15630

Keep users from removing system buckets.

### DIFF
--- a/kv/bucket.go
+++ b/kv/bucket.go
@@ -723,6 +723,19 @@ func (s *Service) updateBucket(ctx context.Context, tx Tx, id influxdb.ID, upd i
 func (s *Service) DeleteBucket(ctx context.Context, id influxdb.ID) error {
 	return s.kv.Update(ctx, func(tx Tx) error {
 		var err error
+
+		bucket, err := s.findBucketByID(ctx, tx, id)
+		if err != nil && !IsNotFound(err) {
+			return err
+		}
+
+		if !IsNotFound(err) && bucket.Type == influxdb.BucketTypeSystem {
+			return &influxdb.Error{
+				Code: influxdb.EInvalid,
+				Msg:  "system buckets cannot be deleted",
+			}
+		}
+
 		if pe := s.deleteBucket(ctx, tx, id); pe != nil {
 			err = pe
 		}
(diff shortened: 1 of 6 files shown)


kv/bucket_test.go
```diff
@@ -17,6 +22,75 @@ func TestInmemBucketService(t *testing.T) {
 	influxdbtesting.BucketService(initInmemBucketService, t)
 }
 
+func TestSystemBucketDeletion(t *testing.T) {
```
jademcgough: Is there a reason to have these tests in here, and not in `testing/bucket_service.go`? I'm not super worried about functionality diverging between different kv implementations, but it's a little weird that these only run against bolt.

bthesorceror: the http bucket service doesn't allow for creating buckets with a type of system and it also uses that test group.

bthesorceror: I suppose the question is, should the http and kv levels be running the same tests?

kv/bucket.go
```diff
@@ -683,7 +683,7 @@ func (s *Service) updateBucket(ctx context.Context, tx Tx, id influxdb.ID, upd i
 		b.Description = *upd.Description
 	}
 
-	if upd.Name != nil {
+	if upd.Name != nil && b.Type != influxdb.BucketTypeSystem {
```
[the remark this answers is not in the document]
bthesorceror: 👍

kv/bucket_test.go
jademcgough: `http/bucket_service.go` is different from `testing/bucket_service.go`. The former is only through the http API, the latter is unit testing of public service functions and runs against bolt, inmem and etcd.

kelwang: I think initially we have the same tests in http level to save some code. But recently, it seems http level has different behaviors, so i think it's not that necessary now

PR #15647, opened 2019-10-29
feat(kv): label names to be unique

### LINKED ISSUE #15591: labels should be unique by name
Within an organization, labels should be unique by name; if a label is created or updated to an existing name, the request should fail. The UI enforces this right now, but it is not enforced over the API.

As a bonus, this would allow us to do lookups by name in constant time by indexing on name like we do in `kv/org.go`.

dearyhud:
Closes #15591

- [] Rebased/mergeable
- [] Tests pass
- [] http/swagger.yml updated (if modified Go structs or API)


kv/label.go
```diff
@@ -85,7 +92,7 @@ func (s *Service) findLabelByID(ctx context.Context, tx Tx, id influxdb.ID) (*in
 
 func filterLabelsFn(filter influxdb.LabelFilter) func(l *influxdb.Label) bool {
 	return func(label *influxdb.Label) bool {
-		return (filter.Name == "" || (filter.Name == label.Name)) &&
+		return (filter.Name == "" || (strings.EqualFold(filter.Name, label.Name))) &&
```
jsteenb2: is this change to maintain this with `utf8` chars? what's the motivation behind this change?

kv/label.go
```diff
@@ -248,6 +255,19 @@ func (s *Service) deleteLabelMapping(ctx context.Context, tx Tx, m *influxdb.Lab
...
+		if err := l.Validate(); err != nil {
+			return &influxdb.Error{
+				Code: influxdb.EInvalid,
+				Err:  err,
+			}
+		}
+
+		l.Name = strings.TrimSpace(l.Name)
+
+		if err := s.uniqueLabelName(ctx, tx, l); err != nil {
```
jsteenb2: one thing we may want to do here is move the read for unqiueness outside the update. In cloud this would force a TX imagine, which seems like a bad idea for this. Moving this to a separate read before the update seems like a better idea 🤔. Also, why are we trimming white space on line 265? is that a behavior that we share across the kv implementations?

kv/label.go
```diff
@@ -530,3 +617,41 @@ func (s *Service) deleteLabel(ctx context.Context, tx Tx, id influxdb.ID) error
...
+	if err != nil {
+		return nil, &influxdb.Error{
+			Code: influxdb.EInvalid,
+			Err:  err,
+		}
+	}
+
+	k := make([]byte, influxdb.IDLength+len(l.Name))
+	copy(k, orgID)
+	copy(k[influxdb.IDLength:], []byte(strings.ToLower((l.Name))))
```
jsteenb2: are we making names unique case insensitive now?

dearyhud: Yup. label names should be unique & case agnostic. the UI currently handles all this stuff. feels like it's eagerly loading the labels and doing the label name validation in the browser...

for this change I added, the casing of the label names in the label bucket shouldn't matter & can be a mix of upper and lower case no problem. this [system bucket](https://github.com/influxdata/influxdb/pull/15647/files/6c7dccb1ceb67b80617cd7acc0b09855f9a24f83#diff-f1b793ee8b2acb0554850b94f082205cR17) is going to store the down-cased version of the label name, and we'll be using that to decide if a matching label name string already exists.

kv/label.go
dearyhud: yeah. our CI linter also seemed to agree with the change.

I had changed things originally to deal w/ finding labels case insensitively like so 
`strings.ToLower(filter.Name) == strings.ToLower(label.Name)` and CI called foul on that use.

kv/label.go
dearyhud: re: transactions - I'm just leaving a note here since we chatted about this. This is an even larger issue that may not get addressed in this PR. TX is pretty heavily depended on unfortunately for reading/writing to kv buckets.

PR #15655, opened 2019-10-30
fix(ui): add filter to fieldOptions when dragging a column

### LINKED ISSUE #15569: Table columns are not re-orderable
table-columns-not-reorderable

Steps to reproduce:

1. Go to Data Explorer
2. Create a basic query
3. View that query as a table
4. Go to customize the table settings
5. Attempt to drag columns into a different order

TCL735:
Closes #15569

Change the filter for unused field options into a private method to allow it to be used for drag events.


zoesteinkamp:
Looks good!

PR #15657, opened 2019-10-30
test(query): unskip skipped PromQL transpiler end-to-end .flux tests

wolffcm:


jsternberg:
I'm assuming the fix for these tests is in the current version of flux?

PR #15679, opened 2019-10-31
feat(tests): added tests to validate bug unreproducible

### LINKED ISSUE #15552: Task name in options form does not update
When I click on one task to edit it, the first time I go to edit another task the name that shows up in the options form on the left does not change. Then if I go to the same task again, it does update the name.

Asalem1:
Closes #15552 

### Problem

Task names don't populate the form correctly when toggling between tasks

### Solution

Since this bug could not be reproduced, I added tests to ensure that the correct names are displaying when toggling between tasks using the navbar, the cancel button, and the save button

For more info, please see:

cannot_reproduce

### DIFF
--- a/ui/cypress/e2e/tasks.test.ts
+++ b/ui/cypress/e2e/tasks.test.ts
@@ -257,6 +257,113 @@ from(bucket: "${name}")
       cy.contains('Save').click()
     })
   })
+
+  describe('renders the correct name when toggling between tasks', () => {
+    // addresses an issue that was reported when clicking tasks
+    // this issue could not be reproduced manually | testing:
+    // https://github.com/influxdata/influxdb/issues/15552
+    const firstTask = 'First_Task'
+    const secondTask = 'Second_Task'
+    const taskName = 'First_Task'
+    const interval = '12h'
+    const offset = '30m'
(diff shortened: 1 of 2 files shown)


ui/cypress/e2e/tasks.test.ts
```diff
@@ -257,6 +257,112 @@ from(bucket: "${name}")
...
+      cy.getByInputValue(secondTask)
+      cy.getByTestID('task-save-btn').click()
+      // navigate back to the first task again
+      cy.getByTestID('task-card--name')
+        .contains(firstTask)
+        .click()
+      cy.getByInputValue(firstTask)
+      cy.getByTestID('task-save-btn').click()
+    })
+  })
```
121watts: You could probably just have one "navigation" test which does all of these things in one test.  https://docs.cypress.io/guides/references/best-practices.html#Creating-%E2%80%9Ctiny%E2%80%9D-tests-with-a-single-assertion

Asalem1: true, but the difference here is that we're testing how the state handles different click events. Even though the outcome is to navigate to the tasks page, we're testing to see whether the reducer acts differently depending on whether we're navigated via the navbar or the button clicks

PR #15681, opened 2019-10-31
fix(toolbar): refactored toolbar and variable to overlay

### LINKED ISSUE #15419: Rebuild Flux Function tooltip using Clockface Popover
Located in `ToolbarFunction.tsx`

image

Asalem1:
Closes #15419 

### Problem

ToolbarFunction and VariableItem were not using Popover

### Solution

Refactored ToolbarFunction and VariableItem to use the clockface/Popover

tooltip_popover

### DIFF
--- a/ui/cypress/e2e/explorer.test.ts
+++ b/ui/cypress/e2e/explorer.test.ts
@@ -192,6 +192,24 @@ describe('DataExplorer', () => {
 
       cy.getByTestID('query-builder').should('exist')
     })
+
+    it('should display the popover when hovering', () => {
+      cy.getByTestID('selector-list my_meas').click()
+      cy.getByTestID('selector-list my_field').click()
+
+      cy.getByTestID('switch-to-script-editor').click()
+      cy.getByTestID('flux-editor').should('exist')
+
+      cy.getByTestID('toolbar-popover--contents').should('not.exist')
+
+      cy.getByTestID('flux-function aggregateWindow').trigger('mouseover')
+
+      cy.getByTestID('toolbar-popover--contents').should('exist')
+
+      cy.getByTestID('flux-function aggregateWindow').trigger('mouseout')
+
+      cy.getByTestID('toolbar-popover--contents').should('not.exist')
+    })
   })
 
   describe('raw script editing', () => {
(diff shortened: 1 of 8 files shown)


ui/cypress/e2e/explorer.test.ts
```diff
@@ -192,6 +192,20 @@ describe('DataExplorer', () => {
 
       cy.getByTestID('query-builder').should('exist')
     })
+
+    it('should display the popover when hovering', () => {
```
hoorayimhelping: based on the discussions watts and zoe and i had today, i think we'll want to refactor this to make it behave a little more deterministically so it's less flaky on ci. i think since `my_field` is dependent on `my_meas`,  and switching to the script editor is dependent on that, and checking the popover contents is dependent on that, we'll want to use some `then` handlers:

```suggestion
it('should display the popover when hovering', () => {
      cy.getByTestID('selector-list my_meas').click().then(() => {
        cy.getByTestID('selector-list my_field').click().then(() => {
          cy.getByTestID('switch-to-script-editor').click().then(() => {
            cy.getByTestID('flux-editor').should('exist')
            cy.getByTestID('toolbar-popover--contents').should('not.exist')

            cy.getByTestID('flux-function aggregateWindow').trigger('mouseover').then(() => {
              cy.getByTestID('toolbar-popover--contents').should('exist')
            })
          })
        })
      })
    })
```
this will add a bit of defense to our tests - all of these `then` calls create a new context and ensure that calls wait for the existence of things before preceding

121watts: @hoorayimhelping i think the two `then`s for the async calls for `my_meas` and `my_field` would be great.  The rest of this stuff (to my knowledge) is all synchronous and should work great without any further nesting.  

The default wait time Cypress uses should be more than enough for:
- `switch-to-script-editor`
- `flux-editor`
- `toolbar-popover--contents`
- `flux-function aggregateWindow`
- `toolbar-popover--conents`

ebb-tide: can we change that wait time?

Asalem1: @ebb-tide the cypress default wait times can be modified, and we can also add a `cy.wait(x)` which will wait a certain amount of time, but that shouldn't be used for waiting on specific commands to resolve

121watts: @ebb-tide we have. We changed it to 10s

Issue #15677, opened 2019-10-31
Deleting via a predicate doesn't delete by measurement


e-dard:
I expected to be able to delete some data by specifying the measurement I wanted to filter on. I then expect all of the data for series with that measurement to be deleted. I used the following tag key to specify the measurement: `_measurement`. 

The data in my test was not deleted.

### Repro

```
influx write -o edd -b edd  "disk,abba=v v=$RANDOM `date +%s`000000000"
```

```
influx query -o edd 'from(bucket: "edd") |> range(start: -4h) |> filter(fn: (r) => r._measurement == "disk")'
Result: _result
Table: keys: [_start, _stop, _field, _measurement, abba]
                   _start:time                      _stop:time           _field:string     _measurement:string             abba:string                      _time:time                  _value:float
------------------------------  ------------------------------  ----------------------  ----------------------  ----------------------  ------------------------------  ----------------------------
2019-10-31T06:35:09.719629000Z  2019-10-31T10:35:09.719629000Z                       v                    disk                       v  2019-10-31T10:31:50.000000000Z                         15480
```

```
edd@Alameda:~/go/src/github.com/influxdata/influxdb|master
⇒  influx delete -o edd -b edd --start 2009-01-02T23:00:00Z --stop 2020-01-02T23:00:00Z
edd@Alameda:~/go/src/github.com/influxdata/influxdb|master
⇒  influx delete -o edd -b edd --start 2009-01-02T23:00:00Z --stop 2020-01-02T23:00:00Z -p _measurement="disk"
edd@Alameda:~/go/src/github.com/influxdata/influxdb|master
⇒  influx delete -o edd -b edd --start 2009-01-02T23:00:00Z --stop 2020-01-02T23:00:00Z -p _measurement='disk'
edd@Alameda:~/go/src/github.com/influxdata/influxdb|master
⇒  influx delete -o edd -b edd --start 2009-01-02T23:00:00Z --stop 2020-01-02T23:00:00Z -p _measurement=disk
edd@Alameda:~/go/src/github.com/influxdata/influxdb|master

⇒  influx query -o edd 'from(bucket: "edd") |> range(start: -4h) |> filter(fn: (r) => r._measurement == "disk")'
Result: _result
Table: keys: [_start, _stop, _field, _measurement, abba]
                   _start:time                      _stop:time           _field:string     _measurement:string             abba:string                      _time:time                  _value:float
[...]

PR #15692, opened 2019-10-31
fix(predicate): add special tag key, fixed children number

kelwang:
Closes https://github.com/influxdata/influxdb/issues/15677

add _measurement and _field special tag key.
fixed a logical node issue.
manually tested as well.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/predicate/logical.go
+++ b/predicate/logical.go
@@ -31,7 +31,7 @@ func (op LogicalOperator) Value() (datatypes.Node_Logical, error) {
 // LogicalNode is a node type includes a logical expression with other nodes.
 type LogicalNode struct {
 	Operator LogicalOperator `json:"operator"`
-	Children []Node          `json:"children"`
+	Children [2]Node         `json:"children"`
 }
 
 // ToDataType convert a LogicalNode to datatypes.Node.
--- a/predicate/parser.go
+++ b/predicate/parser.go
@@ -73,9 +73,7 @@ func Parse(sts string) (n Node, err error) {
 }
 
 func (p *parser) parseLogicalNode() (Node, error) {
-	n := &LogicalNode{
-		Children: make([]Node, 0),
-	}
+	n := new(LogicalNode)
 	for {
 		tok, pos, _ := p.scanIgnoreWhitespace()
 		switch tok {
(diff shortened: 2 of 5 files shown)


predicate/logical.go
```diff
@@ -31,7 +31,7 @@ func (op LogicalOperator) Value() (datatypes.Node_Logical, error) {
 // LogicalNode is a node type includes a logical expression with other nodes.
 type LogicalNode struct {
 	Operator LogicalOperator `json:"operator"`
-	Children []Node          `json:"children"`
+	Children [2]Node         `json:"children"`
```
e-dard: What warranted this change?

predicate/parser.go
```diff
@@ -91,9 +89,29 @@ func (p *parser) parseLogicalNode() (Node, error) {
...
 			}
-			n.Children = append(n.Children, tr)
+			if n.Children[0] == nil {
+				n.Children[0] = tr
+			} else {
+				n.Children[1] = tr
+			}
 		case influxql.AND:
 			n.Operator = LogicalAnd
+			if n.Children[1] != nil {
```
jsteenb2: nit: could flip the comparison here and just continue, unindenting the main logic of this block

```go
if n.Children[1] == nil {
  continue
}
other stuffs...
```

predicate/parser.go
```diff
@@ -201,3 +223,13 @@ scanRegularTagValue:
 		}
 	}
 }
+
+// peekRune returns the next rune that would be read by the scanner.
+func (p *parser) peekTok() influxql.Token {
+	tok, _, _ := p.scanIgnoreWhitespace()
+	if tok != influxql.EOF {
```
jsteenb2: curious why there is no unscan when its `EOF`?

predicate/logical.go
jsteenb2: had same question, is this to avoid the panic on zero value slice(`nil`)?

kelwang: this line got me, didn't saw it in the previous commit
https://github.com/influxdata/influxdb/blob/68cfd386c377d1869dd54805caefdd0c08c4860b/tsdb/tsm1/predicate.go#L163

predicate/parser.go
kelwang: the scan will consume the current position until eof.

predicate/parser.go
kelwang: good idea

PR #15697, opened 2019-10-31
feat(kv): Cursor API accepts hints for improving performance

stuartcarnie:
Implementations of the `kv.Bucket#Cursor` API may use the hints to instruct the access or read behavior to the underlying key/value store.

The `findAllTasks` function was also fixed to ensure that paging works as expected when using a name filter. Tests were added to verify this behavior.

Redundant error checks were also removed.

### DIFF
--- a/kv/store.go
+++ b/kv/store.go
@@ -38,14 +38,41 @@ type Tx interface {
 	WithContext(ctx context.Context)
 }
 
+type CursorHints struct {
+	KeyPrefix *string
+	KeyStart  *string
+}
+
+// CursorHint configures CursorHints
+type CursorHint func(*CursorHints)
+
+// WithCursorHintPrefix is a hint to the store
+// that the caller is only interested keys with the
+// specified prefix.
+func WithCursorHintPrefix(prefix string) CursorHint {
+	return func(o *CursorHints) {
+		o.KeyPrefix = &prefix
+	}
+}
+
+// WithCursorHintKeyStart is a hint to the store
+// that the caller is interested in reading keys from
+// start.
+func WithCursorHintKeyStart(start string) CursorHint {
+	return func(o *CursorHints) {
+		o.KeyStart = &start
+	}
+}
+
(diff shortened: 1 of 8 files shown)


kv/store.go
```diff
@@ -38,14 +38,41 @@ type Tx interface {
 	WithContext(ctx context.Context)
 }
 
+type CursorHints struct {
```
GeorgeMac: ❤️ love how this is done like a variadic option

kv/task.go
```diff
@@ -417,6 +413,45 @@ func (s *Service) findTasksByOrg(ctx context.Context, tx Tx, filter influxdb.Tas
...
+	if f.Name != nil {
+		expected := *f.Name
+		prevFn := fn
+		fn = func(t *influxdb.Task) bool {
+			res := prevFn == nil || prevFn(t)
+			return res && (expected == t.Name)
+		}
+	}
+
+	return fn
```
GeorgeMac: This is a take it or leave it nitpick:
I find this function a little hard to grok because of the way it re-composes the closures as you go down. Along with the occassional long boolean expression.
My suggestion would be to do the boring thing and just return a single closure which exits early on any non-filter matching conditions. Something like:
```go
return func(t *influxdb.Task) bool {
    // given a org filter which doesn't match the tasks organization ID
    if org != nil && org.ID != t.OrganizationID {
       return false
    }

    // given a type filter
    if f.Type != nil {
        // legacy case - from when tasks didn't have a type
        // given the filter is not for the empty type and not for the system type
        if t.Type == "" && !(*f.Type == "" || *f.Type == influxdb.TaskSystemType) {
            return false
        }

        if t.Type != *f.Type {
            return false
        }
    }
    // ...
}
```

stuartcarnie: I tend to favor a functional approach, as they often produce efficient predicates and in many cases like this one, we can avoid conditional branches.

I will write some tests for the function, to ensure correctness.

PR #15705, opened 2019-10-31
test(js): remove snapshots

121watts:
- Removes all snapshot testing
- Adds a linting rule that disallows snapshot testing
- Adds a linting rule that breaks the build if `.only` is used (don't worry you can still run and pass the tests locally)


ui/src/dataExplorer/components/SaveAsButton.test.tsx
```diff
@@ -16,7 +16,6 @@ describe('SaveAsButton', () => {
   describe('rendering', () => {
     it('renders', () => {
       expect(wrapper.exists()).toBe(true)
```
drdelambre: you deleted all coverage on this component!

ui/.eslintrc.js
```diff
@@ -67,5 +66,7 @@ module.exports = {
     'react/no-unescaped-entities': 'off',
     'react/prop-types': 'off',
     'react/no-find-dom-node': 'off',
+    'jest/no-large-snapshots': ['error', {maxSize: 0}], // no shapshots please
```
drdelambre: a bit heavy handed and reactionary

drdelambre:
thanks for removing all of those render snapshots! we can update clockface now!

ui/.eslintrc.js
121watts: i thought we all agreed no more snapshots.  linter rules make that automatic.

PR #15709, opened 2019-10-31
test(ui): rejigger the collector tests to be less flaky

hoorayimhelping:
Makes collector e2e test lest flaky

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/cypress/e2e/collectors.test.ts
+++ b/ui/cypress/e2e/collectors.test.ts
@@ -214,19 +214,19 @@ describe('Collectors', () => {
           .then(() => {
             bucketz.sort()
             cy.getByTestID('bucket-name')
-            .should('have.length', 3)
-            .each((val, index) => {
-              const text = val.text()
-              expect(text).to.include(bucketz[index])
-            })
-            .then(() => {
-              cy.getByTestID('bucket-sorter').click()
-              bucketz.reverse()
-              cy.getByTestID('bucket-name').each((val, index) => {
+              .should('have.length', 3)
+              .each((val, index) => {
                 const text = val.text()
                 expect(text).to.include(bucketz[index])
               })
-            })
+              .then(() => {
+                cy.getByTestID('bucket-sorter').click()
+                bucketz.reverse()
+                cy.getByTestID('bucket-name').each((val, index) => {
+                  const text = val.text()
+                  expect(text).to.include(bucketz[index])
+                })
+              })
           })
 
         // sort by name test here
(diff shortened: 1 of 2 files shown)


ui/cypress/e2e/collectors.test.ts
```diff
@@ -138,52 +147,76 @@ describe('Collectors', () => {
...
+                            )
+                          })
+                      })
+                  })
+              })
+          })
 
-        bucketz.sort()
+        // sort by buckets test here
+        cy.reload()
```
121watts: why reload here?

ui/cypress/e2e/collectors.test.ts
```diff
@@ -196,9 +229,9 @@ describe('Collectors', () => {
               expect(text).to.include(bucketz[index])
             })
           })
-      })
-      // sort by name test here
-      it('can sort telegraf configs by name', () => {
+
+        // sort by name test here
+        cy.reload()
         cy.getByTestID('collector-card--name').should('have.length', 3)
```
121watts: and here?

ui/cypress/e2e/collectors.test.ts
```diff
@@ -21,30 +21,41 @@ describe('Collectors', () => {
...
-        cy.getByTestID('streaming').within(() => {
-          cy.get('.cf-button')
-            .contains('Listen for Data')
+      cy.getByTestID('table-row')
+        .should('have.length', 0)
+        .then(() => {
+          cy.contains('Create Configuration')
             .click()
+            .then(() => {
+              cy.getByTestID('overlay--container').then(() => {
```
121watts: `.within` achieves the same sequential behavior as a `then` but also guarantees the rest of our selectors will be within said container.  
```suggestion
              cy.getByTestID('overlay--container').within(() => {
```

ui/cypress/e2e/collectors.test.ts
hoorayimhelping: it's more a conceptual reload than anything. just kind of like a break in the last bit of tests. i'll remove them

ui/cypress/e2e/collectors.test.ts
hoorayimhelping: mistakenly removed this one, didn't go back through the undo stack far enough i think heh

ui/cypress/e2e/collectors.test.ts
hoorayimhelping: so i smooshed three separate specs into a single spec to limit teardown and rebuilding and such, but a consequence of that these is that there is some wonky leftover state from the previous test that affects the current test (like active filtering or sorting for example). these `reload`s are necessary to clear out that state

PR #15710, opened 2019-11-01
feat(ui): tweak telegraf nginx and redis flows

### LINKED ISSUE #15500: [UI - Telegrafs] Update NGINX values popup fails to close when clicking done
__Steps to reproduce:__

1. Open Load Data telegrafs.
2. Click create configuration.  The wizard loads
3. Choose NGINX
4. Click continue
5. In the plugins list click NGINX.  A panel to configure NGinx URLs loads. 
6. Add a URL. 
7. Click done. NOTHING HAPPENS

__Expected behavior:__
That the value in the URLs field will be validated and if valid, the wizard will return to the review step.  If invalid an error message would be supplied. 

Previously it returned to the review step and the plugin was red, green or grey depending on the validity or existence of the plugin URLs value.  

__Actual behavior:__
Clicking the "Done" button did nothing.  Had to dismiss the wizard. 

__Screencast__
TelegrafNginxPluginDetails01

__Environment info:__

Discovered in nightly build docker image
```
ts=2019-10-18T08:29:40.849811Z lvl=info msg="Welcome to InfluxDB" log_id=0IZVctSG000 version=nightly commit=552168d3a build_date=2019-10-18T07:25:22Z
```
browser: __Chrome 77__
__Config:__

hoorayimhelping:
Closes #15500

This wasn't actually a bug, but a confusing interaction. This uri array component expects users to hit enter to add an item to the list (good luck, mobile / tablet users). This adds a button to help out that interaction. It also changes the way the form is handled so that it doesn't fire an empty event, causing an error.

The bug was reported for nginx, but the confusing interaction also appeared in redis, so this fix has been applied there as well. e2e tests were added to both.

in 2 26 45 PM

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/cypress/e2e/collectors.test.ts
+++ b/ui/cypress/e2e/collectors.test.ts
@@ -253,4 +253,121 @@ describe('Collectors', () => {
       })
     })
   })
+
+  describe('configuring plugins', () => {
+    // fix for https://github.com/influxdata/influxdb/issues/15500
+    describe('configuring nginx', () => {
+      beforeEach(() => {
+        // These clicks launch move through configuration modals rather than navigate to new pages
+        cy.contains('Create Configuration')
+          .click()
+          .then(() => {
+            cy.contains('NGINX')
+              .click()
+              .then(() => {
+                cy.contains('Continue')
+                  .click()
+                  .then(() => {
+                    cy.contains('nginx').click()
+                  })
+              })
+          })
+      })
+
+      it('can add and delete urls', () => {
+        cy.getByTestID('input-field').type('http://localhost:9999')
+        cy.contains('Add').click()
+
+        cy.contains('http://localhost:9999').should('exist', () => {
+          cy.getByTestID('input-field').type('http://example.com')
+          cy.contains('Add').click()
+
+          cy.contains('http://example.com')
+            .should('exist')
+            .then($example => {
+              $example.contains('Delete').click()
+              $example.contains('Confirm').click()
+
+              cy.contains('http://example').should('not.exist')
+
+              cy.contains('Done')
+                .click()
+                .then(() => {
+                  cy.get('.icon.checkmark').should('exist')
+                })
+            })
+        })
+      })
+
+      it('handles busted input', () => {
+        // do nothing when clicking done with no urls
+        cy.contains('Done').click()
+        cy.contains('Done').should('exist')
+        cy.contains('Nginx').should('exist')
+
+        cy.getByTestID('input-field').type('youre mom')
+        cy.contains('Add').click()
+
+        cy.contains('youre mom')
+          .should('exist')
+          .then(() => {
+            cy.contains('Done')
+              .click()
+              .then(() => {
+                cy.get('.icon.remove').should('exist')
+              })
+          })
+      })
+    })
+
+    // redis was affected by the change that was written to address https://github.com/influxdata/influxdb/issues/15500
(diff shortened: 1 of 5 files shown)


ui/src/clockface/components/inputs/multipleInput/MultipleInput.tsx
```diff
@@ -3,19 +3,20 @@ import React, {PureComponent, ChangeEvent} from 'react'
...
 
 // Utils
 import {validateURI} from 'src/shared/utils/validateURI'
 
 // Types
 import {
-  Columns,
-  InputType,
   AutoComplete,
+  Columns,
```
drdelambre: WHOOOO alphabetical!!!

ui/cypress/e2e/collectors.test.ts
```diff
@@ -253,4 +253,101 @@ describe('Collectors', () => {
       })
     })
   })
+
+  describe('configuring plugins', () => {
+    // fix for https://github.com/influxdata/influxdb/issues/15500
+    describe('configuring nginx', () => {
+      beforeEach(() => {
+        cy.contains('Create Configuration').click()
```
drdelambre: why this and not just `cy.visit(whateverurl)`? do we want to test that you always get to the page the same way? or is this some weirdness with the tests? also, idk what i'm reading

ui/src/dataLoaders/components/configureStep/streaming/ConfigFieldSwitcher.test.tsx
```diff
@@ -77,7 +77,7 @@ describe('Onboarding.Components.ConfigureStep.Streaming.ConfigFieldSwitcher', ()
       const {wrapper} = setup({fieldName, fieldType, value}, true)
 
       const input = wrapper.find(ArrayFormElement)
-      const formElement = wrapper.find(FormElement)
+      const formElement = wrapper.find(FormElement).first()
```
drdelambre: 👍

ui/cypress/e2e/collectors.test.ts
hoorayimhelping: this all happens in the context of a popups / modals - you click 'Create Configuration', then you click NGINX (or redis), then Cotinue, etc. i'll add a comment explaining it.

drdelambre: might be fine as is. i was just under the impression that all of our modals are currently behind a route and got suspicious about state management within the app.

ui/cypress/e2e/collectors.test.ts
```diff
@@ -253,4 +253,121 @@ describe('Collectors', () => {
...
+                  .then(() => {
+                    cy.contains('nginx').click()
+                  })
+              })
+          })
+      })
+
+      it('can add and delete urls', () => {
+        cy.getByTestID('input-field').type('http://localhost:9999')
+        cy.contains('Add').click()
```
Asalem1: How would you feel about checking to see if the input-field was cleared after the click resolved?

ui/cypress/e2e/collectors.test.ts
Asalem1: @drdelambre depending on the url you are trying to visit since it might be outside of our control. Generally, if we're trying to test out whether or not we are visiting that other page (i.e. within our site) it's just easy to test out that other page separately rather than include that within the scope of this test

PR #15731, opened 2019-11-04
fix(storage): array cursor iterator should return stats of all observed cursors

GeorgeMac:
Closes https://github.com/influxdata/influxdb/issues/15732

1. Adds a failing test which demonstrates the incorrect reporting of stats from the `arrayCursorIterator.Stats()` method.
2. Updates `arrayCursorIterator.Stats()` to accumulate all the cursor field stats values into a single response.

@stuartcarnie showed me that the `builldXArrayCursor` methods actually retain previously read stats and accumlate them on calls to `reset` which is how all read stats are accumulated in the end.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## v2.0.0-alpha.20 [unreleased]
+
+### Bug Fixes
+
+1. [15731](https://github.com/influxdata/influxdb/pull/15731): Ensure array cursor iterator stats accumulate all cursor stats
+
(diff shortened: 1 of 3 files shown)


GeorgeMac:
I just ran this with our signed integer test and it failed.
So I think perhaps the accumulation of all cursors was the desired effect. I will dig further into this assumption.

GeorgeMac:
Cheers @stuartcarnie and @benbjohnson 👍

PR #15746, opened 2019-11-04
Fix(custom-table): provides data consistency when creating a table

### LINKED ISSUE #15570: Custom table settings not persisted after save
custom-table-settings-not-persisted

Steps to reproduce:
1. Add a table view to a dashboard
2. Edit the table view cell
2. Disable the visible columns or rename a column
3. save the cell
4. Edit the cell again

Asalem1:
Closes #15570 

### Problem

Custom Table data that had been set on graphs were being defaulted whenever they were being reconfigured. If custom tables had been set and graph options were being selected from the list of views, the custom data was being overwritten. 

### Solution

The problem was arising in the `TableGraphs` `updateFieldOptions` method. Since no `fieldOptions` were being passed in, the `headers` were being used as a reference to the custom table data and were therefore being used to overwrite the fieldOptions that had previously been set. The solution to this was to create a helperFunction that would set the fieldOptions based on the headers and the previously set fieldOptions. That is, the headers are being used as the primary parameters for the fieldOptions, but if an alias exists for that specific header, we overwrite the default alias. In addition, any previously aliased values should persist regardless of whether the headers have been removed.

### DIFF
--- a/ui/src/timeMachine/reducers/index.ts
+++ b/ui/src/timeMachine/reducers/index.ts
@@ -811,10 +811,10 @@ export const timeMachineReducer = (
         typeof action.payload
       >
       const {fieldOptions} = action.payload
-      const {fieldOptions: options} = workingView.properties
+      const {fieldOptions: prevOptions} = workingView.properties
       const properties = {
         ...workingView.properties,
-        fieldOptions: trueFieldOptions(fieldOptions, options),
+        fieldOptions: trueFieldOptions(fieldOptions, prevOptions),
       }
       const view = {...state.view, properties}
       return {...state, view}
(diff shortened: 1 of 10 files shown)


ui/src/timeMachine/reducers/index.ts
```diff
@@ -981,8 +992,11 @@ const convertView = (
   outType: ViewType
 ): View<QueryViewProperties> => {
   const newView: any = createView(outType)
-
   newView.properties.queries = cloneDeep(view.properties.queries)
+  if (view.properties.type === 'table' && view.properties.fieldOptions) {
```
Asalem1: @desa and I would love some feedback on how to run a specific condition based on which type a property is. This was a workaround

drdelambre: like a switch?

desa: Something like
```ts
interface Square {
    kind: "square";
    size: number;
}
interface Rectangle {
    kind: "rectangle";
    width: number;
    height: number;
}
interface Circle {
    kind: "circle";
    radius: number;
}

type Shape = Square | Rectangle | Circle;

...

    switch (s.kind) {
        case "square": return s.size * s.size;
        case "rectangle": return s.height * s.width;
        case "circle": return Math.PI * s.radius ** 2;
    }
```

where were I don't have to switch on `kind` and can switch on the interface instead like
```ts
    switch (s.kind) {
        case Square: return s.size * s.size;
        case Rectangle: return s.height * s.width;
        case Circle: return Math.PI * s.radius ** 2;
    }
```

desa: Here's an example from go https://tour.golang.org/methods/16

PR #15749, opened 2019-11-04
feat(ui): adding bundle analyzer for prod and dev

drdelambre:
this adds the ability for developers to compare bundle optimizations. when analyzing in develop mode (read: `yarn start`), assets are uncompressed and can be viewed at `http://localhost:9998`. When in production mode (read: `yarn build`), the analysis can be reached by opening `build/bundle-report.html`.

as a point of order, remember that file size isn't as important from a network resource perspective, and that many small files is fine (production will start running http2 soon to remove much of the overhead of asset handshakes), but that large files slow down the time to first render + interaction by increasing time spent parsing the file (read: its an interpreted language). Best case scenario is to send a small file with only enough code to represent the current view, and to start loading everything else while the user clicks around.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -171,6 +171,7 @@
     "redux-thunk": "^1.0.3",
     "reselect": "^4.0.0",
     "rome": "^2.1.22",
-    "uuid": "^3.2.1"
+    "uuid": "^3.2.1",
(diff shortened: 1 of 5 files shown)


drdelambre:
found out all of monitoring team's assets are served with http2 because we are cool. working with e commerce to help get them cool too.

Issue #15734, opened 2019-11-04
Homepage nav icon never in "active" state


alexpaxton:
<img width="595" alt="Screen Shot 2019-11-04 at 12 09 53 PM" src="https://user-images.githubusercontent.com/2433762/68141739-193aac00-fefc-11e9-833c-df8270aa8672.png">

Looks like the first nav item is never in "active" state even when you are on the home page

PR #15757, opened 2019-11-05
fix(ui): update util to allow matching the homepage as the active icon

TCL735:
Closes https://github.com/influxdata/influxdb/issues/15734

### DIFF
--- a/ui/src/pageLayout/utils/index.test.ts
+++ b/ui/src/pageLayout/utils/index.test.ts
@@ -1,0 +1,90 @@
+// Constants
+import {HOMEPAGE_PATHNAME} from 'src/shared/constants'
+
+// Utils
+import {getNavItemActivation} from 'src/pageLayout/utils'
+
+describe('pageLayout.utils.getNavItemActivation', () => {
+  it('matches the homepage when the location pathname has fewer than 3 parts', () => {
+    let pathname = ''
+    expect(
+      getNavItemActivation([HOMEPAGE_PATHNAME, 'account'], pathname)
+    ).toEqual(true)
+
+    pathname = '/orgs'
+    expect(
+      getNavItemActivation([HOMEPAGE_PATHNAME, 'account'], pathname)
+    ).toEqual(true)
+
+    pathname = '/orgs/3491cbaef55b4559'
+    expect(
+      getNavItemActivation([HOMEPAGE_PATHNAME, 'account'], pathname)
+    ).toEqual(true)
+
+    pathname = '/orgs/3491cbaef55b4559/some-path'
+    expect(
+      getNavItemActivation([HOMEPAGE_PATHNAME, 'account'], pathname)
+    ).toEqual(false)
+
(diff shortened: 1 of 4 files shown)


ui/src/pageLayout/constants/navItemActivation.ts
```diff
@@ -0,0 +1 @@
+export const HOMEPAGE_PATHNAME = 'me'
```
drdelambre: can you drop this into this file:
https://github.com/influxdata/influxdb/blob/master/ui/src/shared/constants/index.ts

ui/src/pageLayout/utils/navItemActivation.ts
```diff
@@ -1,8 +1,14 @@
+// Constants
+import {HOMEPAGE_PATHNAME} from 'src/pageLayout/constants/navItemActivation'
+
 export const getNavItemActivation = (
```
drdelambre: should this be the default export as well? or do you envision other things coming out of this file?

ui/src/pageLayout/utils/navItemActivation.ts
```diff
@@ -1,8 +1,14 @@
...
   keywords: string[],
   location: string
 ): boolean => {
   const ignoreOrgAndOrgID = 3
   const parentPath = location.split('/').slice(ignoreOrgAndOrgID)
+  if (!parentPath.length) {
+    parentPath.push(HOMEPAGE_PATHNAME)
+  }
   return keywords.some(path => parentPath.includes(path))
```
drdelambre: i almost feel like includes should also be fixed. cause wouldn't this activate for the url `/new/meme/reviews`?

TCL735: That example has 3 parts and would not activate the homepage icon.

For the other icons, that example would only activate if "reviews" is passed into the 1st argument array.

I have added two additional unit tests to verify the above.

drdelambre: above and beyond man!
👍 🚀 🌠

ui/src/pageLayout/utils/index.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    expect(
+      getNavItemActivation([given], `${base}/${given}${given}${given}`)
+    ).toEqual(false)
+
+    given = 'tokens'
+    expect(getNavItemActivation([given], `${base}////////${given}`)).toEqual(
+      true
+    )
+  })
+})
```
hoorayimhelping: :heart: love these tests!

only nit / style suggestion i could provide is: instead of using variables, just hard code the values in the test - it's a little more repetitive, but i find it makes the tests more explicit and it usually makes the intent of the test and the code a little easier to follow. again, a nit / suggestion, feel free to ignore.

```js
expect(getNavItemActivation(['tokens'], `/orgs/3491cbaef55b4559////////tokens')).toEqual(true)
```

PR #15758, opened 2019-11-05
feat(pkger): add HTTP api endpoints

jsteenb2:
one thing to note here is I did a small discovery on decoupling the HTTP servers. its within the http server files within this PR. No sweat to role it back, just having hard time building confidently on teh current http servers in the `http` pkg.

- [x] Rebased/mergeable
- [x] Tests pass

<details><summary>curl for POST /api/v2/packages</summary>

```sh
curl -X POST -H "Authorization: Token $token" -d '{ "pkgName": "foo_bird", "pkgDescription": "standup desc", "pkgVersion": "v1" }' localhost:9999/api/v2/packages
```

Resp:

```json
{
	"package": {
		"apiVersion": "0.1.0",
		"kind": "package",
		"meta": {
			"description": "standup desc",
			"pkgName": "foo_bird",
			"pkgVersion": "v1"
		},
		"spec": {
			"resources": []
		}
	}
}
```

</details>


jsteenb2:
@kelwang that's a good question, do we want to swagger this during this alpha phase?

http/api_handler.go
```diff
@@ -18,29 +19,31 @@ import (
 // APIHandler is a collection of all the service handlers.
 type APIHandler struct {
 	influxdb.HTTPErrorHandler
-	BucketHandler               *BucketHandler
-	UserHandler                 *UserHandler
-	OrgHandler                  *OrgHandler
+	AssetHandler                *AssetHandler
```
desa: Super awesome that these were alphabetized

http/api_handler.go
```diff
@@ -262,6 +291,14 @@ func (h *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// this for loop is equivalent to the if statements below
```
desa: This is a really cool idea. We've been meaning to have a router here for a while. I think as we have it now, callers of `NewAPIHandler` could prevent any of the other routes from being handled. What do you think about adding each of the handlers on APIHandler to the `ResourceHandler`s and potentially panicking/erring if someone tries to use a reserved route?

pkger/http_server.go
```diff
@@ -0,0 +1,205 @@
...
+
+	"github.com/go-chi/chi"
+	"github.com/go-chi/chi/middleware"
+	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/kit/tracing"
+	"gopkg.in/yaml.v3"
+)
+
+// HTTPServer is a server that manages the packages HTTP transport.
+type HTTPServer struct {
```
desa: Any reason that the http bits here aren't in the http package? In general, we've tried to follow the pattern of packaging on dependencies. See https://medium.com/@benbjohnson/standard-package-layout-7cdbc8391fc1

kelwang:
> @kelwang that's a good question, do we want to swagger this during this alpha phase?

I think we should if we add new http endpoints.

http/api_handler.go
jsteenb2: Yah, that would be my suggestion actually. I love the `chi` router. Super expressive and has zero 3rd party dependencies, and is plenty fast. It would be trivial to mount these. I think having duplicates is panicable, makes sense to me. Not sure where that would happen though 🤔. Adding same http server twice maybe?

pkger/http_server.go
jsteenb2: yah, I've used std pkg laout numerous times and enjoy it. One thing that having it within the domain pkg (`pkger`) is just that, it lives in the domain. The mental model for it is super small b/c you dn't have to worry about concerns from other routers. The http transport layer has a dependency on a service itself. My previous shop, we employed this monolith server approach, which works while thigns are smaller. Howver, as thigns grow, they tend to become harder to test, things end up coupling in very strange ways, and the mental model becomes unmanageable. I eneded up using this approach to stablizie the API around (100s of endpoints). 

The approach I chose here is a DDD approach, one that I picked up at peter bourgon's elegant monolith/ DDD course at GCon last year (forked and annotated repo for the course [here](https://github.com/jsteenb2/sympatico)). I had a lot of success applying this at my previous shop to stabilize an API that had more holes than swiss cheese. Becomes super easy to decorate svc/http/store or whatever layer you want to add orthogonal functionality too, without mudding the abstraction.

The end result is you can take your service and its concerns anywhere without couplin git. If we want to move some service out into its own service running in its own container/pod, its pretty trivial to do with this setup. It's been coined the elegant monolith approach 🤷‍♂, take it with a grain of salt lol

What I see is  having domain implementations, then we can compose those together into an API gateway/Server type, something that has the router with all the routes registered, common middle ware (loggings/auth/tracing/etc), graceful shutdowns, so forth. Makes things trivial to decouple when you want to horizontally scale a specific service, and also keeps the mental model approachable.

jsteenb2: @desa  that being said, and kudos to you if you made it this far 😬 lol, I can always pull this out and move it to the `http` pkg as well. I wanted to throw this out there as a POC, for a slightly modified approach, that lends itself well to breaking things up into decoupled services.

jsteenb2:
@kelwang I'll go on and add it 👍. Going to be painful to write up the pkg itself for the apply/dry run endpoint

http/api_handler.go
jsteenb2: @desa I'd be happy to break the handlers out into their own beasts. I'd like to do it after this body of work though. Hate to grow the scope of this atm. Would it be good to track this with a separate issue if we do it later?

http/api_handler.go
jsteenb2: I had a difficult time finding things until it was alphabetized 😬

jsteenb2:
@kelwang I went ahead and added the swagger docs for the 2 endpoints that were created. I'm going to go soak my wrists now 🤣, lots of typing

pkger/http_server.go
desa: Yeah, I definitely get where you're coming from. In fact, when the project first started out we had things in the go-kit style of implementation. We very intentionally moved away from that to prevent siloing, which was historically a problem here. Each team would ended up building it's own http client, there was slightly different styles of implementation and behaviors, and we had a variety of interdependencies between projects that got messy.

I'm open to changing the pattern to something and directory structure, but would like to do so when it solves an explicit need that we have.

http/api_handler.go
desa: Would be really cool. Does chi support routing to prefixes? For example, if I have

`/api/v2/orgs/:id/owners` can I register `/api/v2/orgs` with chi and have it route to it?

http/swagger.yml
```diff
@@ -4241,6 +4241,73 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+  /packages:
```
desa: Hate to be nit picky about the name packages, but I'm concern that this will create a bunch of confusion when we add flux packages. @jademcgough can you coordinate with @lukevmorris and @nathanielc about what this should be called.

nathanielc: Agreed, we need to decide on a name and its important. Flux calls a `package` a directory of source code. We will need a different name for a collection of packages that are versioned together. Probably call it `module` or something. There is a many to one relationship between Flux packages and modules. 

Also we need to figure out how a Flux module relates to an InfluxDB package. Do we need a third name or reuse the same name?

http/api_handler.go
jsteenb2: sure does 👌, those are table stakes in `chi` world lol

pkger/http_server.go
jsteenb2: only reason I did the POC here was because of the pain incurred working with the current `http` pkg. It feels very complex for an API router. That could be because we don't have a solid service definition yet. Noticed we end up doing lots of logic in the handler itself, which isn't ideal either. Sounds like the initial work didn't have a pattern in place the team should follow. More guidelines with fuzziness that could be implemented differently. That would be a pain if you have a bunch of similar but different stuffs out there.

I'll move this to the http pkg real quick. Should port just as easy

http/api_handler.go
jsteenb2: @desa I put together some of the bits to make that happen here, cleans it up a fair amount. Check out the origins here:

https://github.com/influxdata/influxdb/pull/15758/files#diff-f047f7de48d9a985ade16057d9d48739R126

Then its use here: https://github.com/influxdata/influxdb/pull/15758/files#diff-f047f7de48d9a985ade16057d9d48739R410

http/swagger.yml
jsteenb2: @nathanielc / @desa / @lukevmorris / @jademcgough. Can we put a bow on the naming here soon? This is a blocker for this PR and a lot of the work to follow.

http/api_handler.go
desa: hell yeah!

pkger/http_server.go
jsteenb2: For sure, I totally understand. Just wanted to throw some ideas out there. Is trivial to move things about between pkgs and what not. Fixing up `http` when it gets broken up sounds like a good idea 👌.  The ole 
> while you got the hood open... 😉

http/swagger.yml
jsteenb2: @nathanielc / @desa  / @lukevmorris, I'm goign to merge this as is. Hit me up when you all land on a good name and Ill remedy this. This is blocking all my forward progress.

jsteenb2:
kv failing to build in upstream 😢

PR #15767, opened 2019-11-05
feat(pkger): adds xy chart

dearyhud:
Visual of XY chart added via the pkger CLI
Screen Shot 2019-11-05 at 11 41 21 AM

- [x] Rebased/mergeable
- [x] Tests pass


pkger/parser.go
```diff
@@ -575,6 +575,8 @@ func parseChart(r Resource) (chart, []failure) {
 		YPos:        r.intShort("yPos"),
 		Height:      r.intShort("height"),
 		Width:       r.intShort("width"),
+		Geom:        r.stringShort("geom"),
+		ShadeBelow:  r.boolShort("shadeBelow"),
```
jsteenb2: does this already exist above on line:
https://github.com/influxdata/influxdb/pull/15767/files#diff-bee826069937abe74ce78a93cc101d2fR571

if so then axe this

ui/yarn.lock
```diff
@@ -1,12131 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
jsteenb2: why is this?

pkger/parser_test.go
```diff
@@ -1192,6 +1192,227 @@ spec:
...
+
+		t.Run("handles invalid config", func(t *testing.T) {
+			tests := []struct {
+				name      string
+				jsonStr   string
+				numErrs   int
+				errFields []string
+			}{
+				{
+					name:      "color missing hex value",
```
jsteenb2: lot of missing test cases here it looks like, don't think its a deal breaker here, but lets be cognizant of this moving forward. probably behooves us to put together some helpers to validate these validation errors across multiple view types

dearyhud: totally agree. feels like we should do that sooner than later. the in-lined json/yml is making this file harder to scroll through, and likely will get harder w/ the next set of charts

PR #15769, opened 2019-11-05
fix(bucket-list): wrapped DWP in featureFlag and sorted bucket list

### LINKED ISSUE #15719: Keep delete with predicate should remain under feature flag
currently all of delete with predicate is available without the feature flag. We should keep it behind a feature flag until we're ready to release it.

### LINKED ISSUE #15720: Buckets list should be sorted by bucket name in delete with predicate
Screenshot 2019-11-01 13 25 02

Asalem1:
Closes #15719
Closes #15720 

### Problem

Delete with predicate dropdown bucket list was not alphabetized 

### Solution

Alphabetized dropdown bucket list for delete with predicate, wrote in edge case for default system buckets and added e2e tests

alpha

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@
 ### Bug Fixes
 
 1. [15731](https://github.com/influxdata/influxdb/pull/15731): Ensure array cursor iterator stats accumulate all cursor stats
-1. [15766](https://github.com/influxdata/influxdb/pull/15766): Reset delete with predicate state after submission
 
 ## v2.0.0-alpha.19 [2019-10-30]
 
--- a/ui/src/buckets/selectors/index.test.ts
+++ b/ui/src/buckets/selectors/index.test.ts
@@ -1,10 +1,10 @@
 // Funcs
-import {isSystemBucket, sortBucketsByName, SYSTEM} from './index'
+import {isSystemBucket, getSortedBucketNames, SYSTEM} from './index'
 
 // Types
 import {Bucket} from 'src/types'
 
-describe('Bucket Selector Tests', () => {
+describe('Bucket Selector', () => {
   it('should return true when a default bucket is passed', () => {
     expect(isSystemBucket(SYSTEM)).toEqual(true)
   })
@@ -160,7 +160,7 @@ describe('Bucket Selector Tests', () => {
       },
     ]
 
-    const results = sortBucketsByName(buckets)
+    const results = getSortedBucketNames(buckets)
     const expectedResult = [
       'alpha',
       'buck2',
--- a/ui/src/buckets/selectors/index.test.ts
+++ b/ui/src/buckets/selectors/index.test.ts
@@ -1,19 +1,175 @@
 // Funcs
-import {isDefaultBucket, MONITORING, sortBucketNames, TASKS} from './index'
+import {isSystemBucket, sortBucketNames, SYSTEM} from './index'
+
+// Types
+import {Bucket} from 'src/types'
 
 describe('Bucket Selector Tests', () => {
   it('should return true when a default bucket is passed', () => {
-    expect(isDefaultBucket(MONITORING)).toEqual(true)
-    expect(isDefaultBucket(TASKS)).toEqual(true)
+    expect(isSystemBucket(SYSTEM)).toEqual(true)
   })
   it('should return false when no default bucket is passed', () => {
-    const bucketName = 'bucket_list'
-    expect(isDefaultBucket('tasks')).toEqual(false)
-    expect(isDefaultBucket('monitoring')).toEqual(false)
-    expect(isDefaultBucket('RANDOM')).toEqual(false)
+    expect(isSystemBucket(`_${SYSTEM}`)).toEqual(false)
+    expect(isSystemBucket(`naming_${SYSTEM}`)).toEqual(false)
+    expect(isSystemBucket('SYSTEM')).toEqual(false)
   })
   it('should sort the bucket names alphabetically', () => {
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Bug Fixes
 
 1. [15731](https://github.com/influxdata/influxdb/pull/15731): Ensure array cursor iterator stats accumulate all cursor stats
+1. [15769](https://github.com/influxdata/influxdb/pull/15769): Alphabetized bucket list dropdown in delete with predicate
```
hoorayimhelping: do we want to add a line in the changelog for a feature that's not released?

ui/src/shared/components/DeleteDataForm/BucketsDropdown.tsx
```diff
@@ -22,7 +22,22 @@ const BucketsDropdown: FunctionComponent<Props> = ({
   bucketName,
   onSetBucketName,
 }) => {
-  const bucketNames = buckets.map(bucket => bucket.name)
+  const handleDefaults = name => name === '_monitoring' || name === '_tasks'
```
hoorayimhelping: This is inside a function component which renders on every frame, roughly 60 times a second. This function definition gets called every time `render` gets called, but it's not using any of the information passed in to render. It's basically needlessly being redefined - it can safely be defined outside the function once and just called every frame. 

Was this line run through prettier? This line, at a glance, might be mistaken for a greater than or less than comparison ("is this a function, or is checking if `name` is greater than or equal to `name` equaling `'_monitoring'` or `'_tasks'`?". It might help readability / comprehension to put parens around the argument and braces and a return statement around the body.

ui/src/shared/components/DeleteDataForm/BucketsDropdown.tsx
```diff
@@ -22,7 +22,22 @@ const BucketsDropdown: FunctionComponent<Props> = ({
...
+  const bucketNames = buckets
+    .map(bucket => bucket.name)
+    .sort((a, b) => {
```
hoorayimhelping: what do you think about moving this line out of the component and into the model (i.e. a reducer or maybe selector)? this seems to be an issue that crops up a bit, and it seems like we want buckets to be sorted properly. Doing this in reduxland will ensure future uses of buckets are already sorted out of the box, and all sorting is consistent (even if it's not quite right).

ui/src/shared/components/DeleteDataForm/BucketsDropdown.tsx
Asalem1: Great suggestion about moving it outside of the FC. Prettier removed the `()` :(

ui/src/shared/components/DeleteDataForm/BucketsDropdown.tsx
Asalem1: I think that sounds like a great idea. Lezdoit

CHANGELOG.md
Asalem1: totally right

ui/src/buckets/selectors/index.test.ts
```diff
@@ -0,0 +1,175 @@
+// Funcs
+import {isSystemBucket, sortBucketNames, SYSTEM} from './index'
+
+// Types
+import {Bucket} from 'src/types'
+
+describe('Bucket Selector Tests', () => {
```
hoorayimhelping: you're describing the bucket selector, not the tests :)

ui/src/buckets/selectors/index.test.ts
```diff
@@ -0,0 +1,175 @@
...
+          self: '/api/v2/buckets/adbb0107da2d7d38',
+          write: '/api/v2/write?org=e483753c9bdb47bf&bucket=adbb0107da2d7d38',
+        },
+        labels: [],
+      },
+      {
+        id: 'e2871ad8f92e752a',
+        orgID: 'e483753c9bdb47bf',
+        type: 'user',
+        name: 'disco inferno',
```
hoorayimhelping: so thanks for getting this stuck in my head -____-

ui/src/shared/components/DeleteDataForm/BucketsDropdown.tsx
```diff
@@ -4,10 +4,13 @@ import {connect} from 'react-redux'
...
 // Types
-import {AppState, Bucket} from 'src/types'
+import {AppState} from 'src/types'
+
+// Selectors
+import {sortBucketNames} from 'src/buckets/selectors/index'
 
 interface StateProps {
-  buckets: Bucket[]
+  buckets: string[]
```
hoorayimhelping: i think you should rename this to `bucketNames` since that's now what it represents, right? otherwise it should probably stay a `Bucket[]`

ui/src/buckets/selectors/index.ts
```diff
@@ -0,0 +1,32 @@
+// Types
+import {Bucket} from 'src/types'
+
+export const SYSTEM = 'system'
+
+export const isSystemBucket = (type: string): boolean => type === SYSTEM
+
+export const sortBucketNames = (buckets: Bucket[]) =>
```
hoorayimhelping: this is sorting buckets by name alphabetically. I think it might be better to call it `sortBucketsByName`

ui/src/shared/components/DeleteDataForm/BucketsDropdown.tsx
Asalem1: good point

PR #13039, opened 2019-04-01
fix(inmem): remove the old inmem implementation

zhulongcheng:
Closes #13031

TODO:
- [x] Replace `inmem.Service` with `kv.Service`
- [x] Remove `inmem.Service`

_Briefly describe your proposed changes:_

  - [x] CHANGELOG.md updated with a link to the PR (not the Issue)
  - [x] Rebased/mergeable
  - [x] Tests pass
  - [x] http/swagger.yml updated (if modified Go structs or API)
  - [x] Sign [CLA](https://influxdata.com/community/cla/) (if not already signed)

PR #15784, opened 2019-11-06
feat(createdAt): Added createdat and updatedAt to Authorization

### LINKED ISSUE #15366: Add CreatedAt, LastUpdatedAt to Authorizations
Issue #13066 audited all API resources to make sure they contained the standard; Name, description, labels, CreatedAt, LastUpdatedAt fields. We need to include the authorizations endpoint in this list. Currently `authorizations` is missing CreateA, LastUpdatedAt, name, and labels fields.

Asalem1:
Closes #15366 

### Problem

CreatedAt & UpdatedAt fields were missing from Authorization 

### Solution

Added createdAt & UpdatedAt fields to Authorization. Refactored a few tests to correspond to these changes

### DIFF
--- a/bolt/onboarding_test.go
+++ b/bolt/onboarding_test.go
@@ -33,5 +33,6 @@ func initOnboardingService(f platformtesting.OnboardingFields, t *testing.T) (pl
 }
 
 func TestOnboardingService_Generate(t *testing.T) {
+	t.Skip("This service is not used, we use the kv bolt implementation")
 	platformtesting.Generate(initOnboardingService, t)
 }
--- a/http/auth_service.go
+++ b/http/auth_service.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"net/http"
 	"path"
+	"time"
 
 	"go.uber.org/zap"
 
@@ -85,6 +86,8 @@ type authResponse struct {
 	User        string               `json:"user"`
 	Permissions []permissionResponse `json:"permissions"`
 	Links       map[string]string    `json:"links"`
+	CreatedAt   time.Time            `json:"createdAt"`
+	UpdatedAt   time.Time            `json:"updatedAt"`
 }
 
 func newAuthResponse(a *platform.Authorization, org *platform.Organization, user *platform.User, ps []permissionResponse) *authResponse {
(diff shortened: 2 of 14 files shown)


bolt/onboarding_test.go
```diff
@@ -33,5 +33,6 @@ func initOnboardingService(f platformtesting.OnboardingFields, t *testing.T) (pl
 }
 
 func TestOnboardingService_Generate(t *testing.T) {
+	t.Skip("This service is not used, we use the kv bolt implementation")
```
kelwang: why don't you just delete this file?

http/auth_service.go
```diff
@@ -85,6 +86,8 @@ type authResponse struct {
 	User        string               `json:"user"`
 	Permissions []permissionResponse `json:"permissions"`
 	Links       map[string]string    `json:"links"`
+	CreatedAt   time.Time            `json:"createdAt"`
```
kelwang: you can use platform. CRUDLog  here

inmem/auth_test.go
```diff
@@ -36,5 +37,6 @@ func initAuthorizationService(f platformtesting.AuthorizationFields, t *testing.
 }
 
 func TestAuthorizationService(t *testing.T) {
+	t.Skip("This service is not used, we use the kv inmem implementation")
```
kelwang: delete this file

inmem/onboarding_test.go
```diff
@@ -24,5 +24,6 @@ func initOnboardingService(f platformtesting.OnboardingFields, t *testing.T) (pl
 }
 
 func TestGenerate(t *testing.T) {
+	t.Skip("these are no longer being used, using kv inmem instead")
```
kelwang: delete this file

http/auth_service.go
Asalem1: We ended up using `CreatedAt` and `UpdatedAt` here to be more explicit in the methods that we're using. To me, `CRUD` connotes that the operation has something to do with the Creation, Reading, Updating, and Deletion of data. Since we only care about two of those conditions (`create`, `update`), it felt important to explicit in the intent of the struct

Asalem1: We ended up using CreatedAt and UpdatedAt here to be more explicit in the methods that we're using. To me, CRUD connotes that the operation has something to do with the Creation, Reading, Updating, and Deletion of data. Since we only care about two of those conditions (create, update), it felt important to explicit in the intent of the struct

bolt/onboarding_test.go
desa: I'd prefer to get rid of these all together with something like https://github.com/influxdata/influxdb/pull/13039

inmem/auth_test.go
desa: see https://github.com/influxdata/influxdb/pull/15784#discussion_r343671948

inmem/onboarding_test.go
desa: https://github.com/influxdata/influxdb/pull/15784#discussion_r343671948

PR #15785, opened 2019-11-06
feat(task): Add a latestScheduled field for tasks

lyondhill:
Previously we overwrote the tasks existing latestCompleted to be used for latestCompleted as well as latestScheduled.
For obvious reasons this is confusing and missleading. I believe by seperating the two fields we can have a clear seperation
of concerns.


kv/task.go
```diff
@@ -743,6 +717,13 @@ func (s *Service) updateTask(ctx context.Context, tx Tx, id influxdb.ID, upd inf
 		}
 	}
 
+	if upd.LatestScheduled != nil {
+		// make sure we only update latest scheduled one way
```
AlirieGray: 👍

task.go
```diff
@@ -44,6 +44,7 @@ type Task struct {
 	Cron            string                 `json:"cron,omitempty"`
 	Offset          string                 `json:"offset,omitempty"`
 	LatestCompleted string                 `json:"latestCompleted,omitempty"`
+	LatestScheduled time.Time              `json:"latestScheduled,omitempty"`
```
AlirieGray: nice

http/task_service.go
```diff
@@ -152,10 +152,53 @@ func NewTaskHandler(b *TaskBackend) *TaskHandler {
...
+	Every           string                 `json:"every,omitempty"`
+	Cron            string                 `json:"cron,omitempty"`
+	Offset          string                 `json:"offset,omitempty"`
+	LatestCompleted string                 `json:"latestCompleted,omitempty"`
+	LastRunStatus   string                 `json:"lastRunStatus,omitempty"`
+	LastRunError    string                 `json:"lastRunError,omitempty"`
+	CreatedAt       string                 `json:"createdAt,omitempty"`
+	UpdatedAt       string                 `json:"updatedAt,omitempty"`
+	Metadata        map[string]interface{} `json:"metadata,omitempty"`
+}
```
stuartcarnie: I like that we are separating the representation of the Task over the HTTP API 👍

kv/task.go
```diff
@@ -117,22 +117,6 @@ func (s *Service) findTaskByID(ctx context.Context, tx Tx, id influxdb.ID) (*inf
...
-		if t.LatestCompleted != "" {
-			tlc, err := time.Parse(time.RFC3339, t.LatestCompleted)
-			if err == nil && latestCompleted.After(tlc) {
-				t.LatestCompleted = latestCompleted.Format(time.RFC3339)
-
-			}
-		} else {
-			t.LatestCompleted = latestCompleted.Format(time.RFC3339)
-		}
-	}
```
stuartcarnie: 🖤

task/backend/coordinator/task_coordinator.go
```diff
@@ -59,8 +59,16 @@ func (t SchedulableTask) Offset() time.Duration {
 
 // LastScheduled parses the task's LatestCompleted value as a Time object
 func (t SchedulableTask) LastScheduled() time.Time {
-	tm, _ := t.LatestCompletedTime()
-	return tm
+	if !t.LatestScheduled.IsZero() {
```
docmerlin: need to check for nil since t.LatestSheduled is a pointer, alternatively this would be fine if we just made t,LatestScheduled not a pointer.

task.go
```diff
@@ -232,13 +233,13 @@ type TaskUpdate struct {
 
 	// LatestCompleted us to set latest completed on startup to skip task catchup
 	LatestCompleted *string                `json:"-"`
+	LatestScheduled *time.Time             `json:"-"`
```
docmerlin: no need for this to be a pointer, it simplifies checking if you make it a time.Time

task/backend/coordinator/task_coordinator.go
lyondhill: The LatestScheduled is not a pointer. But on a taskUpdate it is a pointer so we have the option to not update latestScheduled.

task.go
lyondhill: This is the TaskUpdate struct. I believe we want a pointer so we can pass nil latestScheduled and not apply an update.

Issue #15024, opened 2019-09-06
Allow the user to delete all characters from numeric input


russorat:
A user should be able to delete the last character in this component. If they leave the component without a number, then it should be highlighted in red.

![image](https://user-images.githubusercontent.com/4805997/64453948-95507900-d09e-11e9-9a97-3b2f8d69fd03.png)

desa:
@alexpaxton which project board should this issue be on?

alexpaxton:
@desa monitoring!

121watts:
Looks like the culprit is [this](https://github.com/influxdata/clockface/blob/master/src/Components/Inputs/Input.tsx#L142) line in clockface.

alexpaxton:
I did some investigation for about 2 days into this, was very frustrating but here's my conclusions:

1. It is possible to have a fully controlled input of type `number` without this delete problem, there are many in the UI. If it were a clockface issue you'd see this same bug in multiple places

2. The way check editing state is stored is super strict and does not allow an invalid value to be set in state, this forcing the input to have a value value at all times. It so happens that empty string is not a valid value, so it forces it to be `0`

3. I'd propose creating a type called `DraftCheck` that offers more flexibility. This way we can have more options for showing validation to the user. Also the `Draft` pattern is present in other editing overlays in the UI, not sure why `Check` does not follow that

desa:
Just checked and this appears to be fixed, but I'm not certain

TCL735:
There are 2 issues:

- clockface does not handle NaN correctly: it does not update type, status, and value
- Input change handlers avoid passing in NaN

My approach would be to ameliorate both points.

PR #15797, opened 2019-11-07
fix(ui): Update numeric Input components to pass NaN where appropriat…

TCL735:
Closes https://github.com/influxdata/influxdb/issues/15024

Updates numeric Inputs to deliberately pass NaN when user input is completely deleted, as clockface will handle this appropriately, see https://github.com/influxdata/clockface/pull/373

Fixes edge cases with bin sizes in Heat Maps and decimal places in Single Stat.

### DIFF
--- a/ui/cypress/e2e/explorer.test.ts
+++ b/ui/cypress/e2e/explorer.test.ts
@@ -34,6 +34,150 @@ describe('DataExplorer', () => {
     })
   })
 
+  describe('numeric input validation when changing bin sizes in Heat Maps', () => {
+    beforeEach(() => {
+      cy.getByTestID('page-header--right').within(() => {
+        cy.getByTestID('dropdown')
+          .click()
+          .then(() => {
+            cy.get('#heatmap')
+              .click()
+              .then(() => {
+                cy.getByTestID('cog-cell--button').click()
+              })
+          })
+      })
+    })
+    it('should put input field in error status when "10" gets one character deleted and becomes "1"', () => {
+      cy.get('.view-options').within(() => {
+        cy.getByTestID('grid--column').within(() => {
+          cy.getByTestID('input-field')
+            .first()
+            .click()
+            .type('{backspace}')
+            .getByTestID('input-field--error')
+            .should('have.length', 1)
+        })
+      })
+    })
+    it('should put input field in error status when "10" gets all characters deleted and becomes empty', () => {
+      cy.get('.view-options').within(() => {
+        cy.getByTestID('grid--column').within(() => {
+          cy.getByTestID('input-field')
+            .first()
+            .click()
+            .type('{backspace}{backspace}')
+            .getByTestID('input-field--error')
+            .should('have.length', 1)
+        })
+      })
+    })
+    it('should keep input field in error status when "10" becomes "4"', () => {
+      cy.get('.view-options').within(() => {
+        cy.getByTestID('grid--column').within(() => {
+          cy.getByTestID('input-field')
+            .first()
+            .click()
+            .type('{backspace}{backspace}4')
+            .getByTestID('input-field--error')
+            .should('have.length', 1)
+        })
+      })
+    })
+    it('should keep input field in error status when non-numeric characters are typed', () => {
+      cy.get('.view-options').within(() => {
+        cy.getByTestID('grid--column').within(() => {
+          cy.getByTestID('input-field')
+            .first()
+            .click()
+            .type('{backspace}{backspace}abcdefg')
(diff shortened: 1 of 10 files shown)


ui/src/shared/utils/convertUserInput.ts
```diff
@@ -0,0 +1,5 @@
+import React from 'react'
+
+export const convertUserInputToNumOrNaN = (
+  e: React.ChangeEvent<HTMLInputElement>
+) => (e.target.value === '' ? NaN : Number(e.target.value))
```
Asalem1: 🤔 this ternary might cause us issues if `value` is `undefined`. Is there a possibility that the event won't have a `target` or `value` property?

ui/src/shared/utils/convertUserInput.test.ts
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import {convertUserInputToNumOrNaN} from 'src/shared/utils/convertUserInput'
```
Asalem1: Great job on the tests!

TCL735:
> The PR looks great! I think it could be effective to write up a cypress e2e test to see if the app functions as expected as well.

Good idea. I will attempt to write an e2e for this and will update status here.

ui/src/shared/utils/convertUserInput.ts
TCL735: I think React.ChangeEvents always have a target. As long as we have target, value can be undefined. When value is undefined, the ternary will evaluate to false and Number(undefined) will be NaN.

Asalem1: gotcha. Sweet!

ui/src/shared/utils/convertUserInput.test.ts
```diff
@@ -0,0 +1,34 @@
...
+  })
+
+  it('should convert an empty string to NaN', () => {
+    value = ''
+    event.target.value = value
+
+    result = convertUserInputToNumOrNaN(event)
+    expect(Number.isNaN(result)).toEqual(true)
+  })
+})
```
hoorayimhelping: nice tests!

what do you think about adding tests for some falsey values like `null`, `undefined`, and `false` since it seems like from below, the code handles those cases?

ui/src/timeMachine/components/view_options/DecimalPlaces.tsx
```diff
@@ -65,29 +70,23 @@ class DecimalPlacesOption extends PureComponent<Props, State> {
   }
 
   public handleSetValue = (e: ChangeEvent<HTMLInputElement>): void => {
-    const value = Number(e.target.value) || 0
+    const value = convertUserInputToNumOrNaN(e)
     const {digits, onDecimalPlacesChange} = this.props
 
-    if (value === null) {
-      onDecimalPlacesChange({digits, isEnforced: false})
+    if (value === value && value >= MIN_DECIMAL_PLACES) {
```
hoorayimhelping: do you think the logic here is worth testing?

TCL735: Definitely worth testing. I will cover this with e2e tests, which will be pushed once clockface releases 1.0.5

TCL735: So, it turns out that we can adjust the number of decimals, but we will not be able to see the an actual number with decimal places being updated to the correct decimal place in the e2e tests. But we do have tests to validate that the input field is in default or error status.

ui/src/shared/utils/convertUserInput.test.ts
TCL735: Yes, will do.

ui/cypress/e2e/explorer.test.ts
```diff
@@ -34,6 +34,150 @@ describe('DataExplorer', () => {
...
+            .type('{backspace}3')
+            .invoke('val')
+            .should('equal', '3')
+            .getByTestID('input-field--error')
+            .should('have.length', 0)
+        })
+      })
+    })
+  })
+
```
hoorayimhelping: 👏

ui/cypress/e2e/notificationRules.test.ts
```diff
@@ -47,6 +47,234 @@ describe('NotificationRules', () => {
...
+                          })
+                      })
+                  })
+                })
+              })
+          })
+      })
+    })
+  })
+
```
hoorayimhelping: ❤️

PR #15799, opened 2019-11-07
fix(predicate/parser): updated case statement for NotEqual on deleteWithPredicate API

### LINKED ISSUE #15643: Update predicate endpoint to handle `!=`
Related to #13215 

As per the following comment, the UI should allow for users to support the following boolean logic for key-value pairs (`a = "b" AND c = "d" AND w != "s"`). Currently the newly integrated API throws an error whenever a `!=` predicate is submitted.

https://github.com/influxdata/influxdb/issues/15423#issuecomment-542728048

Blocking Issue #15423

Asalem1:
Closes #15643 

### Problem

predicate/parser NotEqual case would `fallthrough` to the default error whenever requests with a `!=` were being made

### Solution

Updated the predicate/parsers `parseTagRuleNode` NEQ case to handle queries where the equality filter is NotEqual

### DIFF
--- a/predicate/parser.go
+++ b/predicate/parser.go
@@ -77,13 +77,7 @@ func (p *parser) parseLogicalNode() (Node, error) {
 	for {
 		tok, pos, _ := p.scanIgnoreWhitespace()
 		switch tok {
-		case influxql.NUMBER:
-			fallthrough
-		case influxql.INTEGER:
-			fallthrough
-		case influxql.NAME:
-			fallthrough
-		case influxql.IDENT:
+		case influxql.NUMBER, influxql.INTEGER, influxql.NAME, influxql.IDENT:
 			p.unscan()
 			tr, err := p.parseTagRuleNode()
 			if err != nil {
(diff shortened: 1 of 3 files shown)


predicate/parser.go
```diff
@@ -77,13 +77,7 @@ func (p *parser) parseLogicalNode() (Node, error) {
...
 		tok, pos, _ := p.scanIgnoreWhitespace()
 		switch tok {
-		case influxql.NUMBER:
-			fallthrough
-		case influxql.INTEGER:
-			fallthrough
-		case influxql.NAME:
-			fallthrough
-		case influxql.IDENT:
+		case influxql.NUMBER, influxql.INTEGER, influxql.NAME, influxql.IDENT:
```
jademcgough: much cleaner

predicate/parser_test.go
```diff
@@ -117,11 +117,8 @@ func TestParseTagRule(t *testing.T) {
 			node: TagRuleNode{Tag: influxdb.Tag{Key: "abc", Value: "-1221"}},
 		},
 		{
-			str: ` abc != "opq"`,
-			err: &influxdb.Error{
-				Code: influxdb.EInvalid,
-				Msg:  `operator: "!=" at position: 5 is not supported yet`,
-			},
+			str:  ` abc != "opq"`,
+			node: TagRuleNode{Tag: influxdb.Tag{Key: "abc", Value: "opq"}, Operator: influxdb.NotEqual},
```
desa: 🙌

predicate/parser.go
Asalem1: Thanks!

PR #15802, opened 2019-11-07
fix(flakey-test): refactored getSortedBucketNames for more consistency and predictability. Finished DWP API functionality

### LINKED ISSUE #15798: jstest is failing in circleCI but passes locally
__Steps to reproduce:__

See log when PR checks fail. Here is a screenshot:

image

__Expected behavior:__
Unit tests should pass

__Actual behavior:__
Unit test fails

Asalem1:
Closes #15798 

### Problem

This PR deals with two separate problems:

1. The issue noted above was the result of inconsistent behavior in the test.
2. The API still couldn't handle `!=` deleteWithPredicate conditions

### Solution

1. Refactored the `getSortedBucketNames` to break the data into two separate arrays that are sorted, concatenated and mapped.
2. Connected the `datatypes.ComparisonNotEqual` to the `influxdb.NotEqual` case in `predicate/tag_rule`

### DIFF
--- a/predicate/tag_rule.go
+++ b/predicate/tag_rule.go
@@ -44,7 +44,7 @@ func NodeComparison(op influxdb.Operator) (datatypes.Node_Comparison, error) {
 	case influxdb.Equal:
 		return datatypes.ComparisonEqual, nil
 	case influxdb.NotEqual:
-		fallthrough
+		return datatypes.ComparisonNotEqual, nil
 	case influxdb.RegexEqual:
 		fallthrough
 	case influxdb.NotRegexEqual:
--- a/ui/src/buckets/selectors/index.ts
+++ b/ui/src/buckets/selectors/index.ts
@@ -5,28 +5,36 @@ export const SYSTEM = 'system'
 
 export const isSystemBucket = (type: string): boolean => type === SYSTEM
 
-export const getSortedBucketNames = (buckets: Bucket[]) =>
-  buckets
-    .sort((a, b) => {
-      const firstBucket = `${a.name}`.toLowerCase()
-      const secondBucket = `${b.name}`.toLowerCase()
-      if (firstBucket === secondBucket) {
-        return 0
-      }
-      if (isSystemBucket(a.type)) {
-        // ensures that the default system types are the last buckets
-        return 1
-      }
-      if (isSystemBucket(b.type)) {
-        // ensures that the default system types are the last buckets
-        return -1
-      }
-      if (firstBucket < secondBucket) {
-        return -1
-      }
-      if (firstBucket > secondBucket) {
-        return 1
-      }
-      return 0
-    })
-    .map(bucket => bucket.name)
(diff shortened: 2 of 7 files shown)


ui/src/buckets/selectors/index.ts
```diff
@@ -5,28 +5,36 @@ export const SYSTEM = 'system'
 
 export const isSystemBucket = (type: string): boolean => type === SYSTEM
 
-export const getSortedBucketNames = (buckets: Bucket[]) =>
-  buckets
-    .sort((a, b) => {
```
jsteenb2: funny thing about JS, each browser implements `sort` just a bit different, maddening at times. Wonder if this was passing in some js runtimes and not others 🤔

Asalem1: it was actually passing locally but not on circleci. I completely thought that the sort method would work deterministically (especially when it came to two values being compared to one another in a particular order) but I guess not.

:sad_trombone:

predicate/tag_rule.go
```diff
@@ -44,7 +44,7 @@ func NodeComparison(op influxdb.Operator) (datatypes.Node_Comparison, error) {
 	case influxdb.Equal:
 		return datatypes.ComparisonEqual, nil
 	case influxdb.NotEqual:
-		fallthrough
+		return datatypes.ComparisonNotEqual, nil
```
kelwang: can you also update the test predicate_test.go?

Asalem1: @kelwang definitely

ui/src/buckets/selectors/index.ts
```diff
@@ -5,28 +5,36 @@ export const SYSTEM = 'system'
...
+  buckets.forEach(bucket => {
+    // separate system buckets from the rest
+    if (isSystemBucket(bucket.type)) {
+      systemBuckets.push(bucket)
+    } else {
+      otherBuckets.push(bucket)
+    }
+  })
+  // alphabetize system buckets
+  systemBuckets.sort(sortFunc)
```
121watts: All of `sortFunc` could be replaced with 
```
_.sortBy((bucket) => bucket.name.toLocaleLowerCase())
```

ui/src/buckets/selectors/index.ts
```diff
@@ -5,28 +5,36 @@ export const SYSTEM = 'system'
...
+  }
+  if (firstBucket > secondBucket) {
+    return 1
+  }
+  return 0
+}
+
+export const getSortedBucketNames = (buckets: Bucket[]) => {
+  const systemBuckets = []
+  const otherBuckets = []
```
hoorayimhelping: minor nit: this should probably be called `userBuckets` but it's not worth blocking on

ui/src/buckets/selectors/index.ts
Asalem1: I very weirdly can't get it to work using the sortBy function

PR #15809, opened 2019-11-08
refactor(ui): update home page animations

alexpaxton:
- Rebuild homepage cards using clockface `Panel` and CSS Grid
- Emphasize "Load Data" panel
- Add buttons into each panel
- Update animations to not look disabled
- Upgrade clockface dependency to `1.0.5` to utilize button mouse interactions

Screen Shot 2019-11-08 at 11 37 43 AM

- [x] Rebased/mergeable
- [x] Tests pass


ui/src/me/components/GettingStarted.tsx
```diff
@@ -1,53 +1,120 @@
...
-          >
-            <GradientBorder />
-            <ExploreGraphic />
-            <h3 className="getting-started--title">Set up alerting</h3>
-          </Link>
-        </div>
-      </div>
-    )
+const GettingStarted: FunctionComponent<WithRouterProps> = ({
+  params: {orgID},
```
drdelambre: it would be super cool if this came from the org store and not the router. feels weird, because we are including the router anyways, but by having redux be the single source of truth and only using the router as an event marker / generator, it makes the code easier to manage overall

ui/src/me/graphics/CollectorGraphic.scss
```diff
@@ -3,34 +3,31 @@
   ------------------------------------------------------------------------------
 */
 
-$collector-graphic-anim-time: 0.5s;
+$collector-graphic-anim-time: 0.1s;
```
drdelambre: SNAPPY

alexpaxton: Need 4 speeeeed

ui/src/me/components/GettingStarted.tsx
alexpaxton: That's a good call, I'll update this guy. We gon' catch all them loose threads

PR #15820, opened 2019-11-08
fix(ui): make a more deterministic build

### LINKED ISSUE #15842: css issue production in production getting started
Screenshot 2019-11-11 12 38 42

Related to https://github.com/influxdata/influxdb/pull/15809

drdelambre:
Closes #15842
Closes influxdata/clockface#362

we saw some weirdness where css inclusion order was changed when building in production mode. this was the result of some drift between our dev and production environments (specifically the use of MiniCssExtract). I went ahead and dragged as much into the common webpack as made sense to hopefully make our build process feel a bit more deterministic.

the second commit is me fixing the error that used to only show up in production, but now should show up everywhere (cause same same, but different)

### DIFF
--- a/ui/src/me/components/GettingStarted.scss
+++ b/ui/src/me/components/GettingStarted.scss
@@ -29,7 +29,6 @@
 }
 
 .getting-started--card {
-  position: absolute;
   width: 100%;
   height: 100%;
   border: $ix-border solid $g4-onyx;
--- a/ui/package.json
+++ b/ui/package.json
@@ -113,7 +113,7 @@
     "sass-loader": "^7.1.0",
     "source-map-loader": "^0.2.4",
     "style-loader": "^0.23.1",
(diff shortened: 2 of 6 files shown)


ui/src/me/components/GettingStarted.scss
```diff
@@ -29,7 +29,6 @@
 }
 
 .getting-started--card {
-  position: absolute;
```
alexpaxton: Does removing this not break anything visually?

drdelambre: i have no idea as these classes aren't scoped, was leaning on you for that 😬

drdelambre: this line WAS getting overwritten on your branch in dev mode, but when production flipped the css rule order, this is what made the giant page thing

PR #15824, opened 2019-11-09
feat(pkger): add support for exporting resource associations

### LINKED ISSUE #15764: export existing resources via influx packages
As a user, I would like to clone desired resources, so I can enable reuse and source control around my newly created package.

initial POST /api/v2/packages should be extended to support the resource kind and the id of existing resource to clone:

request body: 

```
{
    "pkgName": PKG_NAME,
    "pkgDescription": PKG_DESCRIPTION,
    "pkgVersion": PKG_VERSION,
    "resources": [
        {
            "kind": RESOURCE_KIND,
            "id": RESOURCE_ID
        }
        ....
    ]
}
```

jsteenb2:
this PR adds associations for resources that are being exported and adds those to the new package that is created. 

Closes #15764 

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/pkger/models.go
+++ b/pkger/models.go
@@ -57,6 +57,10 @@ func (k Kind) OK() error {
 	return nil
 }
 
+func (k Kind) title() string {
+	return strings.Title(k.String())
+}
+
 func (k Kind) is(comp Kind) bool {
 	normed := Kind(strings.TrimSpace(strings.ToLower(string(k))))
 	return normed == comp
@@ -266,6 +270,11 @@ func (c chartKind) ok() bool {
 	default:
 		return false
 	}
+}
+
+func (c chartKind) title() string {
+	spacedKind := strings.ReplaceAll(string(c), "_", " ")
+	return strings.ReplaceAll(strings.Title(spacedKind), " ", "_")
 }
 
 // SummaryChart provides a summary of a pkg dashboard's chart.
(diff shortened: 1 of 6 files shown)


pkger/models.go
```diff
@@ -57,6 +57,10 @@ func (k Kind) OK() error {
 	return nil
 }
 
+func (k Kind) title() string {
```
kelwang: feels could be helpful if exported

pkger/service.go
```diff
@@ -156,29 +157,52 @@ func (s *Service) CreatePkg(ctx context.Context, setters ...CreatePkgSetFn) (*Pk
...
+	if err := pkg.Validate(ValidWithoutResources()); err != nil {
 		return nil, err
 	}
 
+	sort.Slice(pkg.Spec.Resources, func(i, j int) bool {
+		iName, jName := pkg.Spec.Resources[i].Name(), pkg.Spec.Resources[j].Name()
+		iKind, _ := pkg.Spec.Resources[i].kind()
+		jKind, _ := pkg.Spec.Resources[j].kind()
+
+		kindPriorities := map[Kind]int{
```
kelwang: i will put it outside, so no need to alloc this every time.

pkger/models.go
jsteenb2: I think you're right. Although I think it could be premature to export right now since nothing is using it. I'll make this exported when this becomes useful outside the pkg here.

PR #15848, opened 2019-11-11
feat(pkger): add enhanced HTTP API error handling for the parser errors of pkger

jsteenb2:
This PR aims to provide an enhanced error response when applying an influx package/template/manifest (packemplifest?). The pkger parsing provides rich feedback to users, this is to extend that feedback across the wire to users who are hitting the HTTP endpoints. Example response body:

```
{
    "code": "unprocessable entity",
    "message": "unable to parse package",
    "details": [
        {
            "field": "dashboard",
            "idx": 0,
            "nested": [
                {
                    "field": "charts",
                    "idx": 11,
                    "nested": [
                        {
                            "field": "axes",
                            "reason": "axis not found: \"x\""
                        },
                        {
                            "field": "geom",
                            "reason": "type not found: \"\""
                        },
                        {
                            "field": "height",
                            "reason": "must be greater than 0"
                        },
                        {
                            "field": "query",
                            "idx": 0,
                            "nested": [
                                {
                                    "field": "query",
                                    "reason": "a query must be provided"
                                }
                            ]
                        }
                    ]
                },
                {
                    "field": "charts",
                    "idx": 10,
                    "nested": [
                        {
                            "field": "kind",
                            "reason": "invalid chart kind provided: "
                        }
                    ]
                }
            ]
        },
        {
            "field": "bucket",
            "idx": 3,
            "nested": [
                {
                    "field": "associations",
                    "reason": "label \"label_11\" does not exist in pkg",
                    "idx": 0
                }
            ]
        }
    ]
}
```

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -56,10 +56,21 @@ const (
 //         Err: err,
 //     }.
 type Error struct {
-	Code string
-	Msg  string
-	Op   string
-	Err  error
+	Code    string
+	Msg     string
+	Op      string
+	Err     error
+	Details []ErrDetail
+}
+
+// ErrDetail provides a means for describing an error with
+// fine grained details of the cause. Useful for validation
+// errors.
+type ErrDetail struct {
+	Field  string      `json:"field"`
+	Reason string      `json:"reason,omitempty"`
+	Idx    *int        `json:"idx,omitempty"`
+	Nested []ErrDetail `json:"nested,omitempty"`
 }
 
 // NewError returns an instance of an error.
@@ -90,6 +101,13 @@ func WithErrorCode(code string) func(*Error) {
 func WithErrorMsg(msg string) func(*Error) {
 	return func(e *Error) {
 		e.Msg = msg
+	}
+}
+
+// WithErrDetails sets the validation error body.
+func WithErrDetails(details ...ErrDetail) func(*Error) {
+	return func(e *Error) {
+		e.Details = append(e.Details, details...)
 	}
 }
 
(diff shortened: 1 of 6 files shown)


errors.go
```diff
@@ -56,10 +56,11 @@ const (
...
 type Error struct {
-	Code string
-	Msg  string
-	Op   string
-	Err  error
+	Code              string
+	Msg               string
+	Op                string
+	Err               error
+	ValidationErrBody map[string]interface{}
```
kelwang: I would just use the Err field. implement an ValidationErr 
```go
     type ValidationError struct{...}

    func (e *ValidationError) Error() string{}
```
So the error will become
```go
     &Error{
        Code: ...,
        Msg: ...
        Err: &ValidationError{...}
    }
```

http/errors.go
```diff
@@ -115,12 +115,14 @@ func (h ErrorHandler) HandleHTTPError(ctx context.Context, err error, w http.Res
 	w.Header().Set("Content-Type", "application/json; charset=utf-8")
 	w.WriteHeader(httpCode)
 	var e struct {
-		Code    string `json:"code"`
-		Message string `json:"message"`
+		Code              string                 `json:"code"`
+		Message           string                 `json:"message"`
+		ValidationErrBody map[string]interface{} `json:"validationErrs,omitempty"`
```
kelwang: change this to Err error `json:"err,omitempty"`, check if you need update the swagger file

errors.go
jsteenb2: hmm, the issue with that is I want the error to be a JSON body only for the HTTP API, in the CLI I just want log format, which is what the Error() method is providing now. There is nothing stopping a user from providing an Err in addition to the validation error as is:

```go
&Error{
        Code: ...,
        Msg: ...
        Err: FOO_ERROR,
        ValidationErrs: // detailed view of the error mentioned above
}
```
The biggest concern for me is it becomes a different experience for the same field in 1 place and not others.

http/errors.go
jsteenb2: oh yah, for sure. Wanted to get this API sorted out before I edit the swagger doc

errors.go
kelwang: The err field is initially designed for different kinds of complex error.  One reason to hide that field is to prevent reading difficulties. 

For example, the flux parsing error is now, instead of a complex embedding error.
```
{
  "code": "internal error",
  "message": "compilation failed: loc 2:17-2:23: expected an operator between two expressions"
}
```

Since your error won't be fit in a simple err message, feels like it's wiser to bring that field back, instead of adding new one. Especially, this is only the pkgr error, not really a domain level error.

jsternberg:
I don't have any useful feedback for this. It might be helpful to consult @goller about error messages. The reason why the error message got changed to just the two fields rather than the nested errors was mostly because the UI found it difficult to deal with the nested errors. I like the format as described in the original PR text though. I'm not sure how this impacts things like Flux.

jsteenb2:
closing this in favor of a different API design after speaking with @goller, think it will be the best of both worlds 👍

Issue #15656, opened 2019-10-30
Dashboards index has no empty state


alexpaxton:
<img width="1792" alt="Screen Shot 2019-10-30 at 12 02 20 PM" src="https://user-images.githubusercontent.com/2433762/67876762-cc775f80-fb0e-11e9-8722-f14f69f7c0cb.png">

__Expected behavior:__
There should be an empty state w/ a create button in the center of the screen

__Actual behavior:__
It is blank

TCL735:
Empty state can be restored easily by fixing https://github.com/influxdata/influxdb/blob/master/ui/src/dashboards/components/dashboard_index/Table.tsx#L77 to evaluate to null.

A new related issue is that the empty state is no longer centered because grid is applied (with the assumption that there are dashboards).

See screenshot 
<img width="1680" alt="image" src="https://user-images.githubusercontent.com/10736577/68627724-60441680-0493-11ea-9898-641acc728d90.png">

PR #15853, opened 2019-11-12
fix(ui): Show empty dashboards state

TCL735:
Closes https://github.com/influxdata/influxdb/issues/15656

When dashboards length is 0, props.children evaluates to null rather than false, so that clockface will render the default state.

image


alexpaxton:
@TCL735 can you add a screenshot in the description?

Issue #15841, opened 2019-11-11
Histogram custom bin sizes are hard to manipulate


TCL735:
Data Explorer -> Histogram -> view options (settings wheel)
1. Use backspace to delete the default custom bin size
2. Deleting all of the characters puts the bin size into Auto

3. Attempting to input a single digit bin size is not intuitive, because steps 1 and 2 are what a user normally would do before inputting a single digit. And once the bin size goes into Auto, user must start over again.

See example:

![histogram_bin_size_change](https://user-images.githubusercontent.com/10736577/68608130-062d5c00-0467-11ea-8794-8957f42e2b7f.gif)

__Expected behavior:__
Should be able to input single digit bin sizes after deleting the existing bin size

__Actual behavior:__
Auto selection is too aggressive, making it difficult for user to input single digit bin sizes.

__Environment info:__

latest master

TCL735:
Previously, numeric Input components were made delete-able when issue https://github.com/influxdata/influxdb/issues/15024 was resolved. Bin sizes in Histograms should be one of the use cases for delete-able numeric Input.

PR #15855, opened 2019-11-12
fix(ui): Allow easier manipulation of custom bin sizes in Histograms

TCL735:
Closes https://github.com/influxdata/influxdb/issues/15841

In Histograms, let the user delete all input in Custom bin sizes to input new numbers, and make Auto bin size mode less aggressive.

### DIFF
--- a/ui/cypress/e2e/explorer.test.ts
+++ b/ui/cypress/e2e/explorer.test.ts
@@ -34,6 +34,57 @@ describe('DataExplorer', () => {
     })
   })
 
+  describe('numeric input using custom bin sizes in Histograms', () => {
+    beforeEach(() => {
+      cy.getByTestID('page-header--right').within(() => {
+        cy.getByTestID('dropdown')
+          .click()
+          .then(() => {
+            cy.get('#histogram')
+              .click()
+              .then(() => {
+                cy.getByTestID('cog-cell--button').click()
+              })
+          })
+      })
+    })
+    it('should keep the user on the Custom field and in error status after all input has been deleted even if non-numeric input is typed', () => {
+      cy.get('.view-options').within(() => {
+        cy.getByTestID('auto-input').within(() => {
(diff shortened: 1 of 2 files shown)


ui/cypress/e2e/explorer.test.ts
```diff
@@ -34,6 +34,73 @@ describe('DataExplorer', () => {
     })
   })
 
+  describe('numeric input using custom bin sizes in Histograms', () => {
```
hoorayimhelping: Check out [Cypress - best practices](https://docs.cypress.io/guides/references/best-practices.html#Creating-%E2%80%9Ctiny%E2%80%9D-tests-with-a-single-assertion) around a number of tests with single assertions (I'm not really a fan of the whole "you were told this and you followed it" tone, but whatever):

## Creating “tiny” tests with a single assertion

While technically this runs fine - this is really excessive, and not performant.

Why you did this pattern in unit tests:

    When assertions failed you relied on the test’s title to know what failed
    You were told that adding multiple assertions was bad and accepted this as truth
    There was no performance penalty splitting up multiple tests because they run really fast

Why you shouldn’t do this in Cypress:

    Writing integration tests is not the same as unit tests
    You will always know (and can visually see) which assertion failed in a large test
    Cypress runs a series of async lifecycle events that reset state between tests
    Resetting tests is much slower than adding more assertions

It is common for tests in Cypress to issue 30+ commands. Because nearly every command has a default assertion (and can therefore fail), even by limiting your assertions you’re not saving yourself anything because any single command could implicitly fail.

---

I think it's probably better to try to have this run in a single test, or as few tests as possible, if it makes sense. But like the docs say - "it's fine..." and if you want to leave it this way, feel free.

TCL735: Good point. I have compacted the tests. Please re-review.

PR #15856, opened 2019-11-12
feat(pkger): adds scatter chart support

dearyhud:
Closes #

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/pkger/models.go
+++ b/pkger/models.go
@@ -257,6 +257,7 @@ type chartKind string
 const (
 	chartKindUnknown            chartKind = ""
 	chartKindGauge              chartKind = "gauge"
+	chartKindScatter            chartKind = "scatter"
 	chartKindSingleStat         chartKind = "single_stat"
 	chartKindSingleStatPlusLine chartKind = "single_stat_plus_line"
 	chartKindXY                 chartKind = "xy"
@@ -265,7 +266,7 @@ const (
 func (c chartKind) ok() bool {
 	switch c {
 	case chartKindSingleStat, chartKindSingleStatPlusLine, chartKindXY,
-		chartKindGauge:
+		chartKindGauge, chartKindScatter:
 		return true
 	default:
 		return false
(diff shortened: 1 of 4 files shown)


pkger/parser_test.go
```diff
@@ -1455,6 +1832,206 @@ spec:
 				}
 			})
 		})
+
+		t.Run("pkg with single dashboard xy chart", func(t *testing.T) {
+			t.Run("xy chart", func(t *testing.T) {
```
jsteenb2: I'm a little confused by this big diff here. Thought these tests were all here already? Was thsi just reorded to be aplhabetical? If so, schaawweet 👌

pkger/models.go
```diff
@@ -767,6 +768,22 @@ func (c chart) properties() influxdb.ViewProperties {
...
 			ShowNoteWhenEmpty: c.NoteOnEmpty,
 		}
+	case chartKindScatter:
+		return influxdb.ScatterViewProperties{
+			Type:              influxdb.ViewPropertyTypeScatter,
+			Queries:           c.Queries.influxDashQueries(),
+			ViewColors:        c.Colors.strings(),
+			XColumn:           c.XCol,
+			YColumn:           c.YCol,
+			XAxisLabel:        c.Axes.influxAxes()["x"].Label,
```
jsteenb2: you may want to run the `c.Axes.influxaxes()` call on before the return statement, then reuse its contents here instead of regenerating the influxaxes 6 times

pkger/models.go
```diff
@@ -939,6 +958,16 @@ func (c colors) influxViewColors() []influxdb.ViewColor {
 	return iColors
 }
 
+func (c colors) strings() []string {
+	clrs := []string{}
```
jsteenb2: is it desired to set a `[]` if no colors provided? or does the `nil` value better suffice? When these get json marshaled the one will provide an empty array, whereas the other will be a `null`.

dearyhud: are these api contracts outlined anywhere? i'm curious what the FE folks are expecting, or if either will suffice.

pkger/parser_test.go
dearyhud: yes sir - re ordered things to make it a little easier on the eyes to follow!

pkger/models.go
jsteenb2: think you're cool to leave this as is for now. Should probably check the swagger docs for this 🤔, I'm not sure about teh contract's existence

PR #15857, opened 2019-11-12
feat(pkger): adds heatmap

dearyhud:
Closes #

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/pkger/models.go
+++ b/pkger/models.go
@@ -257,6 +257,7 @@ type chartKind string
 const (
 	chartKindUnknown            chartKind = ""
 	chartKindGauge              chartKind = "gauge"
+	chartKindHeatMap            chartKind = "heatmap"
 	chartKindScatter            chartKind = "scatter"
 	chartKindSingleStat         chartKind = "single_stat"
 	chartKindSingleStatPlusLine chartKind = "single_stat_plus_line"
@@ -265,8 +266,9 @@ const (
 
 func (c chartKind) ok() bool {
 	switch c {
-	case chartKindSingleStat, chartKindSingleStatPlusLine, chartKindXY,
-		chartKindGauge, chartKindScatter:
+	case chartKindGauge, chartKindHeatMap,
+		chartKindScatter, chartKindSingleStat,
+		chartKindSingleStatPlusLine, chartKindXY:
 		return true
 	default:
 		return false
@@ -729,6 +731,7 @@ const (
 	fieldChartXPos          = "xPos"
 	fieldChartYCol          = "yCol"
 	fieldChartYPos          = "yPos"
+	fieldChartBinSize       = "binSize"
 )
 
 type chart struct {
@@ -750,6 +753,8 @@ type chart struct {
 	XCol, YCol    string
 	XPos, YPos    int
 	Height, Width int
+
+	BinSize int
 }
 
 func (c chart) properties() influxdb.ViewProperties {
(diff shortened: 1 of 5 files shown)


dashboard.go
```diff
@@ -681,7 +681,7 @@ type HeatmapViewProperties struct {
 	Type              string           `json:"type"`
 	Queries           []DashboardQuery `json:"queries"`
 	ViewColors        []string         `json:"colors"`
-	BinSize           int32            `json:"binSize"`
```
jsteenb2: can we leave this as is, this may have unintended consequences? What's the motivation here to change it?

pkger/models.go
```diff
@@ -767,6 +771,23 @@ func (c chart) properties() influxdb.ViewProperties {
...
 		}
+	case chartKindHeatMap:
+		return influxdb.HeatmapViewProperties{
+			Type:              influxdb.ViewPropertyTypeHeatMap,
+			Queries:           c.Queries.influxDashQueries(),
+			ViewColors:        c.Colors.strings(),
+			BinSize:           c.BinSize,
+			XColumn:           c.XCol,
+			YColumn:           c.YCol,
+			XAxisLabel:        c.Axes.influxAxes()["x"].Label,
```
jsteenb2: can you pull this `influxAxes()` calls before the return statement?

pkger/parser.go
```diff
@@ -844,6 +845,10 @@ func (r Resource) intShort(key string) int {
 	return i
 }
 
+func (r Resource) int32Short(key string) int32 {
```
jsteenb2: lets leave out all int32 stuffs from teh Resource. Leave it as as int, and convert it at when we assign the value in the inflxudb.ViewProperty type. I don't see it being used anyhow 🤷‍♂

pkger/parser_test.go
```diff
@@ -1012,6 +1012,251 @@ spec:
 			})
 		})
 
+		t.Run("pkg with single dashboard heatmap chart", func(t *testing.T) {
+			testfileRunner(t, "testdata/dashboard_heatmap.json", func(t *testing.T, pkg *Pkg) {
```
jsteenb2: we need to support both `yml` and `json` for the happy path stuff. Make sure to run this as `testdata/dashboard_heatmap` and have goth yml/json avialble to run

pkger/models.go
```diff
@@ -750,6 +753,8 @@ type chart struct {
 	XCol, YCol    string
 	XPos, YPos    int
 	Height, Width int
+
+	BinSize int32
```
jsteenb2: make this type `int` and convert it on https://github.com/influxdata/influxdb/pull/15857/files#diff-26b9adadf7ef8d11b67d2d091e56b593R799

```go
BinSize:           int32(c.BinSize)
```

I want to keep the integration types as vanilla as possible right now.

pkger/parser_test.go
```diff
@@ -1833,7 +2078,7 @@ spec:
 			})
 		})
 
-		t.Run("single xy chart", func(t *testing.T) {
+		t.Run("pkg with single dashboard xy chart", func(t *testing.T) {
```
jsteenb2: please change this back to the what it was before, this creates stutter in the test output. The test this is nested in provides the context of `pkg with single dashboard and single chart`.

dearyhud:
the int32 stuff was resolved [here](https://github.com/influxdata/influxdb/pull/15857/files#diff-bee826069937abe74ce78a93cc101d2fR696) @jsteenb2

pkger/parser_test.go
dearyhud: that's my bad. I must've accepted the wrong diff during my rebase. will fix

PR #15860, opened 2019-11-12
fix(tsi1): address index defect when using negated equality operator

### LINKED ISSUE #15859: != filter does not return matching series in some cases for TSI index
For any series held in `IndexFiles` in the TSI index, any attempt to retrieve them using a negated equality operator will fail. 

For example, suppose you have:

```
cpu region=west
```

and you want `SELECT * FROM cpu WHERE region != ''` then data associated with`cpu region=west` will not be returned if the tag block for that series has been compacted into an `IndexFile`.

However, if the tag block is still in a `LogFile`, or if the query is `region != 'anything_none_empty'` then the data will be returned. In the former case the defect is not triggered within the index code, in the latter case the defective iterator is not used at all.

e-dard:
Fixes #15859.

This PR fixes a defect in the TSI index where a filter using the negated equality (`!=`) operator could result in no matching series being returned for any series stored within the `IndexFile` portions of the index.

The root cause of this was due to missing legacy-handling code in the index for this particular iterator.

This defect was particularly troublesome to track down because there is nothing fundamentally wrong with the TSI index data, only how that data is read in certain circumstances. Around `1.6.1` we added some improvements to the TSI index by persisting sets of series ids as serialised roaring bitmaps. To handle existing TSI indexes we needed to ensure we could deserialise the new bitmaps collections and the legacy collections (variable length encoded blocks of integers IIRC). 

As part of that work we should have ensured that we used the conversion method (`.SeriesIDSet()`) on the `TagKeySeriesIDIterator` but we overlooked it.

The `TagKeySeriesIDIterator` is only really used in one situation—when a predicates is `tag_key != ''`...

### DIFF
--- a/tsdb/tsi1/index_file_test.go
+++ b/tsdb/tsi1/index_file_test.go
@@ -3,7 +3,10 @@ package tsi1_test
 import (
 	"bytes"
 	"context"
+	"fmt"
+	"reflect"
 	"testing"
+	"time"
 
 	"github.com/influxdata/influxdb/models"
 	"github.com/influxdata/influxdb/tsdb"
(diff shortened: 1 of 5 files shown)


e-dard:
Backports:
1.8 https://github.com/influxdata/influxdb/pull/15861
1.7 https://github.com/influxdata/influxdb/pull/15863

tsdb/tsi1/index_file_test.go
```diff
@@ -32,6 +35,55 @@ func TestCreateIndexFile(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		if e.SeriesID.ID == 0 {
+			break
+		}
+		fmt.Println(e.SeriesID.ID)
+
+		name, tags := tsdb.ParseSeriesKey(sfile.SeriesKey(e.SeriesID))
+		got = append(got, string(models.MustNewPoint(string(name), tags, models.Fields{"a": "a"}, time.Time{}).Key()))
```
jacobmarble: What would be the value of `got` before this PR?

e-dard: got would be empty. No series keys would be returned from the index. I just noticed a stray Println above. I'll tidy that up.

PR #15864, opened 2019-11-12
feat(verify-seriesfile): add ability to verify by org/bucket

e-dard:
This PR does the following:

 - Fixes a bug with the tool wherein it would not verify 2.x Series File data (it was not removing the type information from the series ids before verifying them.
 - Adds options to let the operator completely verify a series file even if an error is encountered (`--continue-on-error`).
 - Adds an option to expose a stacktrace for an encountered panic (`--recover=false`).
 - Adds the ability to run series file verification only on series data scoped to either an organisation or a bucket.

### DIFF
--- a/cmd/influxd/inspect/verify_series.go
+++ b/cmd/influxd/inspect/verify_series.go
@@ -1,6 +1,7 @@
 package inspect
 
 import (
+	"encoding/hex"
 	"os"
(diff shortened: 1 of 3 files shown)


danxmoran:
Closing as stale now that we're porting back the 1.x equivalents

PR #15873, opened 2019-11-12
feat: import jsteenb2/testttp directly into pkg

jsteenb2:
removes small external dependency by pulling it in directly

- [x] Rebased/mergeable
- [ ] Tests pass


jsteenb2:
@e-dard good point, I'll clean that up 👍

Issue #15868, opened 2019-11-12
e2e tests performance


TCL735:
__Steps to reproduce:__
Some e2e tests are excessive and should be combined to make multiple assertions.

Example:

<img width="1680" alt="image" src="https://user-images.githubusercontent.com/10736577/68698754-78b93d00-0536-11ea-8b6d-ac542b944bfc.png">

__Expected behavior:__
e2e tests should be compact for performance

__Actual behavior:__
Too many separate tests.

__Environment info:__

latest master

PR #15875, opened 2019-11-12
test: combine tests and remove unnecessary thens

TCL735:
Closes https://github.com/influxdata/influxdb/issues/15868

Make the e2e tests compact for better performance.

### DIFF
--- a/ui/cypress/e2e/explorer.test.ts
+++ b/ui/cypress/e2e/explorer.test.ts
@@ -37,37 +37,21 @@ describe('DataExplorer', () => {
   describe('numeric input using custom bin sizes in Histograms', () => {
     beforeEach(() => {
       cy.getByTestID('page-header--right').within(() => {
-        cy.getByTestID('dropdown')
-          .click()
-          .then(() => {
-            cy.get('#histogram')
-              .click()
-              .then(() => {
-                cy.getByTestID('cog-cell--button').click()
-              })
-          })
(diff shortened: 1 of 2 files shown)


ui/cypress/e2e/explorer.test.ts
```diff
@@ -99,7 +99,7 @@ describe('DataExplorer', () => {
           })
       })
     })
-    it('should put input field in error status when "10" gets one character deleted and becomes "1"', () => {
+    it('should put input field in error status and stay in error status when input is invalid or empty', () => {
```
hoorayimhelping: is it possible to combine these two tests, or make the descriptions of them unique?

at a glance, since they all have the same description, it kind of seems like they're duplicates. a little bit of specificity would be helpful here

TCL735: One test is for Histogram bin sizes, which was compacted as part of https://github.com/influxdata/influxdb/pull/15855 - just that the test description was not updated until now.

The other test is for Heat Map bin sizes. The input there has slightly different behavior but we want to test for the same thing essentially, which is that empty and invalid input will put the component into error status. Here, both the test description and the tests were changed.

TCL735: So I believe the test descriptions are fine. And per watts, the nested ``then``s are unnecessary. I will be removing them.

PR #15885, opened 2019-11-13
fix(storage): skip TSM files with block read errors

### LINKED ISSUE #15804: Corrupt TSM files not always skipped
When we initially open TSM files, if we get an error we skip the file and rename it for future reference. However, future reads don't do the same check so a corrupt block can cause errors. One place we've hit this recently is during compaction, which is convenient because we can set aside the bad file right there. I'm not clear yet if that's sufficient to solve this issue in general.

tmgordeeva:
When we find a bad TSM file during compaction, propagate the error up and move
the bad file aside.

Closes #15804

### DIFF
--- a/tsdb/engine/tsm1/compact.go
+++ b/tsdb/engine/tsm1/compact.go
@@ -68,6 +68,18 @@ func (e errCompactionAborted) Error() string {
 	return "compaction aborted"
 }
 
+type errBlockRead struct {
+	file string
+	err  error
+}
+
+func (e errBlockRead) Error() string {
+	if e.err != nil {
+		return fmt.Sprintf("block read error on %s: %s", e.file, e.err)
+	}
+	return fmt.Sprintf("block read error on %s", e.file)
+}
+
 // CompactionGroup represents a list of files eligible to be compacted together.
 type CompactionGroup []string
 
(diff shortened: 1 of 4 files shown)


tsdb/engine/tsm1/compact.go
```diff
@@ -1752,7 +1772,8 @@ RETRY:
 		}
 
 		if iter.Err() != nil {
-			k.err = iter.Err()
+			k.err = errBlockRead{k.currentTsm, iter.Err()}
+			panic("block read")
```
sebito91: If we panic here do we end up processing the error at all? If we could skip the `panic` that would be awesome and just bubble up the err here, right?

tmgordeeva: oops - this was accidentally left in.

dgnorton:
@tmgordeeva feel like we should add a unit test. How difficult would that be?

tmgordeeva:
@dgnorton I might be able to modify one of the existing tests, I think it by accident already has a file renamed. Let me see if it's easy.

tmgordeeva:
@dgnorton Unfortunately not that simple, the way the compaction planner interacts with the engine means any test like this has to do some tricks to get compactions to actually run at the engine level in the test. Most of our other compaction tests test just the compactor, but this is more integration level because it has to do with the engine, compaction planner, and compactor.

dgnorton:
@tmgordeeva okay. Were you able to test with a corrupt shard and see the compactor move it aside after this change?

tmgordeeva:
Yes, I test corrupting a shard, seeing that compactions failed before the change and that the file was renamed and compactions could resume after.

Thinking about the unit test did make me realize it would be better to do the filestore replace first and then the rename, as the filestore replace has some logic around safely moving around TSM files. I'll push that update now.

PR #15890, opened 2019-11-13
feat(pkger): exports scatter and heatmap charts

dearyhud:
Closes #

- Adds an export for the scatter and heatmap charts
- Also adds 2 additional pkger fields for the heatmap, and scatter charts (`xDomain`, `yDomain`). At first, didn't realized these existed, but you can create these fields when in the chronograf UI. Decided to add these to make sure we're backwards compatible.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/pkger/clone_resource.go
+++ b/pkger/clone_resource.go
@@ -76,6 +76,22 @@ func convertCellView(cv cellView) chart {
 	case influxdb.GaugeViewProperties:
 		setCommon(chartKindGauge, p.ViewColors, p.DecimalPlaces, p.Queries)
 		setNoteFixes(p.Note, p.ShowNoteWhenEmpty, p.Prefix, p.Suffix)
+	case influxdb.HeatmapViewProperties:
+		ch.Kind = chartKindHeatMap
+		ch.Queries = convertQueries(p.Queries)
+		ch.Colors = stringsToColors(p.ViewColors)
+		ch.XCol = p.XColumn
+		ch.YCol = p.YColumn
+		ch.Axes = []axis{
+			{Label: p.XAxisLabel, Prefix: p.XPrefix, Suffix: p.XSuffix, Name: "x", Domain: p.XDomain},
+			{Label: p.YAxisLabel, Prefix: p.YPrefix, Suffix: p.YSuffix, Name: "y", Domain: p.YDomain},
+		}
+		ch.Note = p.Note
+		ch.NoteOnEmpty = p.ShowNoteWhenEmpty
+		ch.BinSize = int(p.BinSize)
+	case influxdb.MarkdownViewProperties:
+		ch.Kind = chartKindMarkdown
+		ch.Note = p.Note
 	case influxdb.LinePlusSingleStatProperties:
 		setCommon(chartKindSingleStatPlusLine, p.ViewColors, p.DecimalPlaces, p.Queries)
 		setNoteFixes(p.Note, p.ShowNoteWhenEmpty, p.Prefix, p.Suffix)
@@ -87,9 +103,18 @@ func convertCellView(cv cellView) chart {
 	case influxdb.SingleStatViewProperties:
 		setCommon(chartKindSingleStat, p.ViewColors, p.DecimalPlaces, p.Queries)
 		setNoteFixes(p.Note, p.ShowNoteWhenEmpty, p.Prefix, p.Suffix)
-	case influxdb.MarkdownViewProperties:
-		ch.Kind = chartKindMarkdown
+	case influxdb.ScatterViewProperties:
+		ch.Kind = chartKindScatter
+		ch.Queries = convertQueries(p.Queries)
+		ch.Colors = stringsToColors(p.ViewColors)
+		ch.XCol = p.XColumn
+		ch.YCol = p.YColumn
+		ch.Axes = []axis{
+			{Label: p.XAxisLabel, Prefix: p.XPrefix, Suffix: p.XSuffix, Name: "x", Domain: p.XDomain},
+			{Label: p.YAxisLabel, Prefix: p.YPrefix, Suffix: p.YSuffix, Name: "y", Domain: p.YDomain},
+		}
 		ch.Note = p.Note
+		ch.NoteOnEmpty = p.ShowNoteWhenEmpty
 	case influxdb.XYViewProperties:
 		setCommon(chartKindXY, p.ViewColors, influxdb.DecimalPlaces{}, p.Queries)
 		setNoteFixes(p.Note, p.ShowNoteWhenEmpty, "", "")
@@ -124,6 +149,10 @@ func convertChartToResource(ch chart) Resource {
 
 	if ch.Legend.Type != "" {
 		r[fieldChartLegend] = ch.Legend
+	}
+
+	if ch.BinSize != 0 {
+		r[fieldChartBinSize] = ch.BinSize
 	}
 
 	ignoreFalseBools := map[string]bool{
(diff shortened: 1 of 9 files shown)


pkger/clone_resource.go
```diff
@@ -278,3 +314,11 @@ func variableToResource(v influxdb.Variable, name string) Resource {
 
 	return r
 }
+
+func v1ColorsToViewColors(clrs []string) colors {
```
jsteenb2: thsi name is a bit misleading. you aren't returning ViewColors here, just colors which is different 🤔

pkger/clone_resource.go
```diff
@@ -87,6 +87,38 @@ func convertCellView(cv cellView) chart {
...
+		ch.Axes = []axis{xAxis, yAxis}
+		ch.Note = p.Note
+		ch.NoteOnEmpty = p.ShowNoteWhenEmpty
+	case influxdb.HeatmapViewProperties:
+		ch.Kind = chartKindHeatMap
+		ch.Queries = convertQueries(p.Queries)
+		ch.Colors = v1ColorsToViewColors(p.ViewColors)
+		ch.XCol = p.XColumn
+		ch.YCol = p.YColumn
+		ch.XDomain = p.XDomain
```
jsteenb2: this feels better off in the axes and not the root level of the chart. What's the motivation here?

pkger/parser.go
```diff
@@ -751,6 +751,20 @@ func parseChart(r Resource) (chart, []failure) {
 		}
 	}
 
+	if _, ok := r[fieldChartXDomain]; ok {
+		for _, str := range r.slcStr(fieldChartXDomain) {
+			val, _ := strconv.ParseFloat(str, 64)
```
jsteenb2: we should be handling this error here

pkger/parser.go
```diff
@@ -751,6 +751,20 @@ func parseChart(r Resource) (chart, []failure) {
...
+			c.XDomain = append(c.XDomain, val)
+		}
+	}
+
+	if _, ok := r[fieldChartYDomain]; ok {
+		for _, str := range r.slcStr(fieldChartYDomain) {
+			val, _ := strconv.ParseFloat(str, 64)
```
jsteenb2: same

dashboard.go
```diff
@@ -836,13 +836,14 @@ func NewBuilderTag(key string, values ...string) struct {
...
-	Base         string   `json:"base"`   // Base represents the radix for formatting axis values
-	Scale        string   `json:"scale"`  // Scale is the axis formatting scale. Supported: "log", "linear"
+	Bounds       []string  `json:"bounds"` // bounds are an arbitrary list of client-defined strings that specify the viewport for a View
+	LegacyBounds [2]int64  `json:"-"`      // legacy bounds are for testing a migration from an earlier version of axis
+	Label        string    `json:"label"`  // label is a description of this Axis
+	Prefix       string    `json:"prefix"` // Prefix represents a label prefix for formatting axis values
+	Suffix       string    `json:"suffix"` // Suffix represents a label suffix for formatting axis values
+	Base         string    `json:"base"`   // Base represents the radix for formatting axis values
+	Scale        string    `json:"scale"`  // Scale is the axis formatting scale. Supported: "log", "linear"
+	Domain       []float64 `json:"domain"` // Scale is the axis formatting scale. Supported: "log", "linear"
```
jsteenb2: We dont want to extend the domain type here. This is purely a pkger concern

pkger/clone_resource.go
```diff
@@ -76,6 +76,24 @@ func convertCellView(cv cellView) chart {
...
+		ch.Kind = chartKindHeatMap
+		ch.Queries = convertQueries(p.Queries)
+		ch.Colors = stringsToColors(p.ViewColors)
+		ch.XCol = p.XColumn
+		ch.YCol = p.YColumn
+
+		xAxis := axis{Label: p.XAxisLabel, Prefix: p.XPrefix, Suffix: p.XSuffix, Name: "x", Domain: p.XDomain}
+		yAxis := axis{Label: p.YAxisLabel, Prefix: p.YPrefix, Suffix: p.YSuffix, Name: "y", Domain: p.YDomain}
+
+		ch.Axes = []axis{xAxis, yAxis}
```
jsteenb2: nit, prefer this

```go
ch.Axes = []axis{
   {Label: p.XAxisLabel, Prefix: p.XPrefix, Suffix: p.XSuffix, Name: "x", Domain: p.XDomain},
   {Label: p.YAxisLabel, Prefix: p.YPrefix, Suffix: p.YSuffix, Name: "y", Domain: p.YDomain},
}
```

pkger/clone_resource.go
```diff
@@ -76,6 +76,24 @@ func convertCellView(cv cellView) chart {
...
+		ch.Note = p.Note
+		ch.NoteOnEmpty = p.ShowNoteWhenEmpty
+
```
jsteenb2: can we axe all the extra spaces here?

pkger/models.go
```diff
@@ -800,11 +805,15 @@ func (c chart) properties() influxdb.ViewProperties {
...
-			Queries:           c.Queries.influxDashQueries(),
-			ViewColors:        c.Colors.strings(),
-			XColumn:           c.XCol,
-			YColumn:           c.YCol,
+			Type:       influxdb.ViewPropertyTypeScatter,
+			Queries:    c.Queries.influxDashQueries(),
+			ViewColors: c.Colors.strings(),
+			XColumn:    c.XCol,
+			YColumn:    c.YCol,
+
```
jsteenb2: go on an axe the spaces extra spaces throughout these changes here

pkger/service_test.go
```diff
@@ -821,24 +874,25 @@ func TestService(t *testing.T) {
 						},
 					},
 					{
-						name:    "guage",
+						name:    "with new name single stat",
```
jsteenb2: this one should be where it was previously, its a different test case than the ones that follow. It is testing the name change and the single stat, but the single stat is kind of a freebie here

dearyhud: I did some more re-ordering/alphabetizing here. the tests for the gauge chart were moved up [here](https://github.com/influxdata/influxdb/blob/cc315a3dc6fceb3d0f7342efffc5c152d641c221/pkger/service_test.go#L787). lmk if anything seems off w/ the tests after I moved them

pkger/models.go
```diff
@@ -1102,16 +1103,26 @@ const (
...
-	Prefix string `json:"prefix,omitempty" yaml:"prefix,omitempty"`
-	Scale  string `json:"scale,omitempty" yaml:"scale,omitempty"`
-	Suffix string `json:"suffix,omitempty" yaml:"suffix,omitempty"`
+	Base   string    `json:"base,omitempty" yaml:"base,omitempty"`
+	Label  string    `json:"label,omitempty" yaml:"label,omitempty"`
+	Name   string    `json:"name,omitempty" yaml:"name,omitempty"`
+	Prefix string    `json:"prefix,omitempty" yaml:"prefix,omitempty"`
+	Scale  string    `json:"scale,omitempty" yaml:"scale,omitempty"`
+	Suffix string    `json:"suffix,omitempty" yaml:"suffix,omitempty"`
+	Domain []float64 `json:"domain,omitempty" yaml:"domain,omitempty"`
```
jsteenb2: 👍

pkger/parser.go
```diff
@@ -746,13 +746,29 @@ func parseChart(r Resource) (chart, []ValidationErr) {
 		c.Axes = presAxes
 	} else {
 		for _, ra := range r.slcResource(fieldChartAxes) {
+			domain := []float64{}
```
jsteenb2: nit: can you make this `var domain []float64`, the behavior when encoding could make this empty slice appear valid instead of it not being there. The `var domain []float64` will be converted to a null to JSON, when you have `domain := []float64{}`, it is converted into an empty slice. Generally you'll want to prefer the `var` approach in pkger, unless the `[]` is desired, which it is in some cases (UI may require it).

dearyhud: good deal

PR #15893, opened 2019-11-13
feat(tasks): add ability to filter by active tasks

AlirieGray:
This PR adds an "Active" filter to the `TaskFilter` struct to enable filtering by Task Status. The "Active" is a pointer to a bool, so it may be true, false, or nil when it is not set explicitly.

This PR also refactors the `findTasksByOrg` function to be more DRY, as well as consolidating multiple filter functions in kv/task.go with duplicated functionality.

- [x] Rebased/mergeable
- [x] Tests pass
- [x] swagger.yml updated

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -4342,6 +4342,14 @@ paths:
             type: string
           description: Filter tasks to a specific organization ID.
         - in: query
+          name: status
+          schema:
+            type: string
+            enum:
+              - active
+              - inactive
+          description: Filter tasks by a status--"inactive" or "active".
+        - in: query
           name: limit
           schema:
             type: integer
--- a/http/task_service.go
+++ b/http/task_service.go
@@ -511,6 +511,12 @@ func decodeGetTasksRequest(ctx context.Context, r *http.Request, orgs influxdb.O
 		req.filter.Limit = influxdb.TaskDefaultPageSize
 	}
 
+	if status := qp.Get("status"); status == "active" {
+		req.filter.Status = &status
+	} else if status := qp.Get("status"); status == "inactive" {
+		req.filter.Status = &status
+	}
+
 	// the task api can only create or lookup system tasks.
 	req.filter.Type = &influxdb.TaskSystemType
 
@@ -1472,6 +1478,10 @@ func (t TaskService) FindTasks(ctx context.Context, filter influxdb.TaskFilter) 
 	}
 	if filter.Limit != 0 {
 		val.Add("limit", strconv.Itoa(filter.Limit))
+	}
+
+	if filter.Status != nil {
+		val.Add("status", *filter.Status)
 	}
 
 	if filter.Type != nil {
--- a/kv/task.go
+++ b/kv/task.go
@@ -310,6 +310,8 @@ func (s *Service) findTasksByUser(ctx context.Context, tx Tx, filter influxdb.Ta
 		return nil, 0, err
 	}
 
+	matchFn := newTaskMatchFn(filter, org)
+
 	for _, m := range maps {
 		task, err := s.findTaskByIDWithAuth(ctx, tx, m.ResourceID)
 		if err != nil && err != influxdb.ErrTaskNotFound {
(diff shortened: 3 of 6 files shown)


kv/task.go
```diff
@@ -425,21 +404,15 @@ func (s *Service) findTasksByOrg(ctx context.Context, tx Tx, filter influxdb.Tas
...
+			if len(ts) >= filter.Limit {
+				break
+			}
 		}
-	}
 
-	if filter.Name != nil {
-		ts = filterByName(ts, *filter.Name)
+		k, v = c.Next()
 	}
```
stuartcarnie: Great! 

This fixes the same bug that I discovered in `findAllTasks`, where the limit is applied before we complete all filtering. It is possible for all tasks to be filtered out by `Name`, preventing the caller from moving to the next page. By moving the check to the `matchFn`, we ensure that the limit is applied after filtering has happened.

http/task_service.go
```diff
@@ -511,6 +511,14 @@ func decodeGetTasksRequest(ctx context.Context, r *http.Request, orgs influxdb.O
...
 	}
 
+	if status := qp.Get("active"); status == "true" {
+		active := false
+		req.filter.Active = &active
+	} else if status := qp.Get("active"); status == "false" {
+		active := false
+		req.filter.Active = &active
+	}
+
```
stuartcarnie: You will need to amend the `swagger.yml` to include this new query argument.

kv/task.go
```diff
@@ -315,13 +315,15 @@ func (s *Service) findTasksByUser(ctx context.Context, tx Tx, filter influxdb.Ta
 			continue
 		}
 
-		if taskFilterMatch(filter.Type, task.Type) {
+		matchFn := newTaskMatchFn(filter, nil)
```
stuartcarnie: This should be moved outside the `for` loop and the second argument can be passed the `org` variable, which is an instance of `*influxdb.Organization`. The `matchFn` will then ensure the org is filtered if set. 

Doing that will also allow you to remove lines 314-316:

```go
if org != nil && task.OrganizationID != org.ID {
	continue
}
```

http/task_service.go
AlirieGray: good catch!

http/swagger.yml
```diff
@@ -4341,6 +4341,11 @@ paths:
           schema:
             type: string
           description: Filter tasks to a specific organization ID.
+        - in: query
+          name: active
+          schema:
+            type: string
+          description: Filter tasks to a specific status--a value of"true" correspondings to active, "false" to inactive.
```
stuartcarnie: nit: space between `of"true"`

http/swagger.yml
```diff
@@ -4341,6 +4341,11 @@ paths:
           schema:
             type: string
           description: Filter tasks to a specific organization ID.
+        - in: query
+          name: active
+          schema:
+            type: string
+          description: Filter tasks to a specific status--a value of "true" corresponds to active, "false" to inactive.
```
jademcgough: A downside to using a boolean here is that it would be harder to extend the API if we were to add a third type of status for tasks in the future.

glinton: Agreed, seems `status` as the name and `active` and `inactive` as current acceptable contents would be a better fit.

stuartcarnie: Originally we weren't going to expose `active` via the HTTP API, but clearly a valid request a client would wish to make.  @AlirieGray and I had a discussion and for parity, will expose `status` as a string enum. 👍

http/swagger.yml
```diff
@@ -4341,6 +4341,11 @@ paths:
           schema:
             type: string
           description: Filter tasks to a specific organization ID.
+        - in: query
+          name: status
+          schema:
+            type: string
+          description: Filter tasks by a status--"inactive" or "active".
```
stuartcarnie: Great! Can you add an `enum` annotation to list the only valid values as `active` and `inactive`, similar to:

https://github.com/influxdata/influxdb/blob/1a28a3256e50b44afb547a14b997c2b0cf7c4527/http/swagger.yml#L6698-L6701

This will improve documentation and code generation.

http/task_service.go
```diff
@@ -511,6 +511,14 @@ func decodeGetTasksRequest(ctx context.Context, r *http.Request, orgs influxdb.O
...
 	}
 
+	if status := qp.Get("active"); status == "true" {
+		active := string(backend.TaskActive)
+		req.filter.Status = &active
+	} else if status := qp.Get("active"); status == "false" {
+		inactive := string(backend.TaskInactive)
+		req.filter.Status = &inactive
+	}
+
```
stuartcarnie: Is this a query parameter? I am not sure if this is redundant or that it should be changed to `status`?

@jademcgough may be able to provide better feedback than I

AlirieGray: I've fixed it to be "status" and use the status query parameter directly!

PR #15900, opened 2019-11-14
fix(tsdb): error instead of panic'ing on corrupted series file

JanBerktold:
We repeatedly see InfluxDB crashing on startup due to a corrupted data directory with a stack trace which looks _rougly_ like this:

```
panic: runtime error: slice bounds out of 

range goroutine 3617 [running]:
github.com/influxdata/influxdb/tsdb.ReadSeriesKey(0x7fde4cc02f45, 0x91b00bb, 0x91b00bb, 0x7fde4c330006, 0xc00037bf40, 0x7fde4c33d709, 0xd9, 0x9a758f7, 0xaaeea615dd120b4e)    
/go/src/github.com/influxdata/influxdb/tsdb/series_file.go:335 +0xa9
github.com/influxdata/influxdb/tsdb.ReadSeriesKeyFromSegments(0xc00238c580, 0x7, 0x8, 0x606e4ff45, 0x7fde4c33d709, 0xd9, 0x9a758f7)    
/go/src/github.com/influxdata/influxdb/tsdb/series_segment.go:292 +0x9c
github.com/influxdata/influxdb/tsdb.(*SeriesIndex).FindIDBySeriesKey(0xc00020d900, 0xc00238c580, 0x7, 0x8, 0xc0f34fd1f5, 0xdc, 0xcf236, 0xce99f)    
/go/src/github.com/influxdata/influxdb/tsdb/series_index.go:204 +0x1c5github.com/influxdata/influxdb/tsdb.(*SeriesPartition).CreateSeriesListIfNotExists(0xc000788580, 0xc0ebf9e000, 0x2710, 0x2710, 0xc0ebfda000, 0x2710, 0x2710, 0xc0ebfee000, 0x2710, 0x2710, ...)   
 /go/src/github.com/influxdata/influxdb/tsdb/series_partition.go:189 +0x109github.com/influxdata/influxdb/tsdb.(*SeriesFile).CreateSeriesListIfNotExists.func1(0x8, 0x151bb18)    
/go/src/github.com/influxdata/influxdb/tsdb/series_file.go:151 +0x7fgithub.com/influxdata/influxdb/vendor/golang.org/x/sync/errgroup.(*Group).Go.func1(0xc02f0da0f0, 0xc0ed2c15c0)    
/go/src/github.com/influxdata/influxdb/vendor/golang.org/x/sync/errgroup/errgroup.go:58 +0x57created by github.com/influxdata/influxdb/vendor/golang.org/x/sync/errgroup.(*Group).Go    
/go/src/github.com/influxdata/influxdb/vendor/golang.org/x/sync/errgroup/errgroup.go:55 +0x66
```

We fix this with some educated guesses and deleting series files until we catch the bad one. Obviously, we should find & resolve the root cause of why this is happening to us, but in the meantime, this PR will at least improve the behaviour to not crash the process due to a panic but rather return an error which happens to state which file/partition is corrupted.

I have 'assumed' at the 'kit/errors' package is what we want to use going forward. Please let me know if that is not correct.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 1. [15713](https://github.com/influxdata/influxdb/pull/15713): Mock missing Flux dependencies when creating tasks
 1. [15731](https://github.com/influxdata/influxdb/pull/15731): Ensure array cursor iterator stats accumulate all cursor stats
 1. [15866](https://github.com/influxdata/influxdb/pull/15866): Do not show Members section in Cloud environments
+1. [15900](https://github.com/influxdata/influxdb/pull/15900): Error instead of crashing on corrupted series file
 
 ### UI Improvements
 1. [15809](https://github.com/influxdata/influxdb/pull/15809): Redesign cards and animations on getting started page
--- a/tsdb/series_file.go
+++ b/tsdb/series_file.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 
 	"github.com/cespare/xxhash"
+	errors2 "github.com/influxdata/influxdb/kit/errors"
 	"github.com/influxdata/influxdb/kit/tracing"
 	"github.com/influxdata/influxdb/logger"
 	"github.com/influxdata/influxdb/models"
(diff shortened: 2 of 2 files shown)


e-dard:
@JanBerktold thanks for the PR! A couple of things:

Are you aware of the tooling we have for verifying the integrity of series files? Are you using 1.7.x or are you using the 2.0 Alpha? For 1.7.x you can use the `influx_inspect verify-seriesfile` command. In 2.0 Alpha we are still working on the tooling, but it would be available under `influxd inspect verify-seriesfile`. These tools can help you track down corrupted files.

The second thing I wanted to chat about was: 

> We repeatedly see InfluxDB crashing on startup due to a corrupted data directory with a stack trace which looks rougly like this

How many times has this happened? Have you checked for faulty hardware? Are you often shutting down influxdb in a none-graceful way? If you're seeing corrupted series files, and it's not through HW or operator error, then that's a defect we need to investigate.

Thirdly, I'm not sure about this approach in the PR. I'm going to need to discuss it. The behaviour should be that the server should not start up if the series file is corrupted. It might be preferable to error out closer to the source of the problem, and provide the user with some explicit feedback about what to do.

Finally, what version are you running? Thanks again!

PR #15920, opened 2019-11-14
fix(ui): arrows in table

### LINKED ISSUE #15888: Arrows in table UI don't do anything
View data in Table View and arrows appear on column headers. When clicked, they don't change until you click out of the table. And none of the data in the table rearranges no matter if the arrow is up/down. 

Why Important: Confusing to users because if they see the arrows they would expect interacting with them to change the data

In screenshot, "Points Against" column is sorted with `desc:false` in Flux query

tableview_arrows

zoesteinkamp:
The fix for getting sorting to work

Closes #15888 

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/src/shared/components/tables/TableGraph.tsx
+++ b/ui/src/shared/components/tables/TableGraph.tsx
@@ -71,7 +71,9 @@ class TableGraph extends PureComponent<Props, State> {
       sortOptions.field = fieldName
       sortOptions.direction = DEFAULT_SORT_DIRECTION
     }
-    this.setState({sortOptions})
+
+    const newSortOptions = {...sortOptions}
+    this.setState({sortOptions: newSortOptions})
   }
 
   private get sortOptions(): SortOptions {
--- a/ui/src/shared/components/tables/TableGraph.tsx
+++ b/ui/src/shared/components/tables/TableGraph.tsx
@@ -62,18 +62,17 @@ class TableGraph extends PureComponent<Props, State> {
   }
 
   public handleSetSort = (fieldName: string) => {
-    const {sortOptions} = this.state
-
-    if (fieldName === sortOptions.field) {
-      sortOptions.direction =
-        sortOptions.direction === ASCENDING ? DESCENDING : ASCENDING
-    } else {
-      sortOptions.field = fieldName
-      sortOptions.direction = DEFAULT_SORT_DIRECTION
-    }
-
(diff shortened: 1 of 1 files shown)


ui/src/shared/components/tables/TableGraph.tsx
```diff
@@ -71,7 +71,9 @@ class TableGraph extends PureComponent<Props, State> {
       sortOptions.field = fieldName
       sortOptions.direction = DEFAULT_SORT_DIRECTION
     }
-    this.setState({sortOptions})
+
+    const newSortOptions = {...sortOptions}
+    this.setState({sortOptions: newSortOptions})
```
TCL735: Why not use the setState updater function here and put everything into the callback, which will update immediately? For reference see: https://reactjs.org/docs/react-component.html#setstate Something like this, which may need to be cleaned up with prettier:

    public handleSetSort = (fieldName: string) => {
      this.setState(({sortOptions}) => {
        const newSortOptions = {...sortOptions}
        if (fieldName === sortOptions.field) {
          newSortOptions.direction =
            sortOptions.direction === ASCENDING ? DESCENDING : ASCENDING
        } else {
          newSortOptions.field = fieldName
          newSortOptions.direction = DEFAULT_SORT_DIRECTION
        }
        return {newSortOptions}
      })
    }

drdelambre: I'm not following. I'm cheating by having offline context, but this change is to fix a condition in which data within sortOptions had changed, but the reference was never updated for `sortOptions` so react didn't know about it. Cool reference though! Maybe i'm still wrapping my head around it.

TCL735: setState takes an object or a function. Using a function makes the component update immediately, instead of possibly waiting (when using an object) due to React batching the state updates.

When a user updates the sort options, they expect to see something update immediately, which seems like a good thing to guarantee with the setState updater function.

drdelambre: 👍 thanks for helping me learn

zoesteinkamp: I tried to do this in the code, and it was complaining about sort options not being assignable, so for now ill just keep what we have :)

TCL735: i edited my snippet. I left off the correct key in the updated state

PR #15934, opened 2019-11-15
fix(pkgr/parser): added failure definition to prevent build failure

Asalem1:
Updated pkgr/parser to define failures to prevent runtime go run build failure

### DIFF
--- a/pkger/parser.go
+++ b/pkger/parser.go
@@ -668,6 +668,7 @@ func parseChart(r Resource) (chart, []ValidationErr) {
 			Msg:   err.Error(),
 		}}
 	}
+	var failures []ValidationErr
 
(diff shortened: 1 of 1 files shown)


pkger/parser.go
```diff
@@ -668,6 +668,7 @@ func parseChart(r Resource) (chart, []ValidationErr) {
 			Msg:   err.Error(),
 		}}
 	}
+	var failures []ValidationErr
```
jsteenb2: nit: can you move this close to the spot it is being used? It looks like line 728 is a good spot for it

Asalem1:
This issue has been resolved in PR:

https://github.com/influxdata/influxdb/pull/15935

PR #15937, opened 2019-11-15
feat(pkger): add pkg new operation to cli to create packages

### LINKED ISSUE #15699: API to generate manifest shell
For this sprint, just generate a shell over the API.

Next sprint, we will use this API to specify which resources should belong in the generated manifest.

jsteenb2:
Closes #15699 

This extends the influx cli with the ability to create packages.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influx/pkg.go
+++ b/cmd/influx/pkg.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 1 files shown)


cmd/influx/pkg.go
```diff
@@ -91,6 +103,83 @@ func pkgApply(orgID, path *string, hasColor, hasTableBorders *bool) func(*cobra.
...
 
+func pkgCreateCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "new",
+		Short: "Create a reusable pkg to create resources in a declarative manner",
+	}
+
+	wd, err := os.Getwd()
+	if err != nil {
+		panic(err)
```
jsteenb2: intentionally panicing here, it should never reach here, and if it does, we have bigger problems 🤷‍♂

cmd/influx/pkg.go
```diff
@@ -91,6 +103,83 @@ func pkgApply(orgID, path *string, hasColor, hasTableBorders *bool) func(*cobra.
 	}
 }
 
+func pkgCreateCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "new",
+		Short: "Create a reusable pkg to create resources in a declarative manner",
+	}
+
+	wd, err := os.Getwd()
```
kelwang: to make it nicer, i will also add an output flag, so user to put it to anywhere they want

jsteenb2: totally agree, I added the following to handle that, the wd is purely for a default val.
https://github.com/influxdata/influxdb/pull/15937/files#diff-0ea570dc74aae6eb4768e34c0d584960R118

kelwang: looks like you already have a flag

Issue #15828, opened 2019-11-11
influx bucket create cannot use organization name


jsternberg:
The `influx bucket create` command doesn't seem to have a `-o` flag for specifying the organization name. Here is what I did on my local box using the latest master at revision a64b976561c798f9aeeb771c4f48fb5d3141ad5f.

```
jsternberg:[~/g/s/g/i/influxdb] $ (master $=) influx bucket create -o influxdata -n NOAA                                              [7/7]
Error: unknown shorthand flag: 'o' in -o
Usage:
  influx bucket create [flags]

Flags:
  -h, --help                 Help for the create command
  -n, --name string          Name of bucket that will be created
      --org-id string        The ID of the organization that owns the bucket
  -r, --retention duration   Duration in nanoseconds data will live in bucket

Global Flags:
      --host string    HTTP address of Influx (default "http://localhost:9999")
      --local          Run commands locally against the filesystem
  -t, --token string   API token to be used throughout client calls

jsternberg:[~/g/s/g/i/influxdb] $ (master $=) influx org find
ID                      Name
97aa81cc0e247dc4        influxdata
jsternberg:[~/g/s/g/i/influxdb] $ (master $=) influx bucket create --org-id 97aa81cc0e247dc4 -n NOAA
ID                      Name    Retention       Org     OrganizationID
b76b31ef2f0341e0        NOAA    0s              <nil>   <nil>
```

Additionally, I found it weird that when I did create the bucket, the `Org` and `OrganizationID` fields were nil. It seems like these aren't getting copied in to the command's display. When I do `influx bucket find` these fields are still nil.

Most other commands support `-o org` as an option and it is a lot more convenient than having to find the organization id explicitly. It would also make writing tutorials easier because then you don't have to teach the person how to retrieve the organization id.

jademcgough:
Acceptance criteria:
- all places where we accept an org-id in the CLI should also accept an organization name
(you'll need to do some research into where this is missing for other CLI functions; we already know it's an issue for bucket, it could also be an issue for others)
- if a user passes both the org-id an the org name, we should have a consistent behavior for how we handle that (either reject with a message, or always only use the org-id or name)

PR #15947, opened 2019-11-16
fix(cli): add organization back to bucket response

jademcgough:
ref #15828 

This fixes a mismatch between tabwriter keys introduced by 9d44ac3635b2800c1dd8e90ac23729e1d298eb0a.

It's probably worth considering adding org names back in the response. Expecting users to work with orgIDs for everything feels like bad UX when it's really an internal implementation detail.

### DIFF
--- a/cmd/influx/bucket.go
+++ b/cmd/influx/bucket.go
@@ -90,14 +90,13 @@ func bucketCreateF(cmd *cobra.Command, args []string) error {
 		"ID",
 		"Name",
 		"Retention",
-		"Org",
 		"OrganizationID",
 	)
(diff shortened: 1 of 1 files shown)


jademcgough:
@kelwang organization IDs do not show in the CLI currently because the keys specified in the tab writer don't match those in the tab writer header. This fixes that.

kelwang:
Ah, okay. Could you change OrganizationID to OrgID, seems most of our code base is using OrgID instead.

jademcgough:
@kelwang we use organizationID in all other places in our CLI, so this is consistent with that. I'd rather merge this and have that be a separate issue if we want it.

PR #15952, opened 2019-11-17
feat(inspect): add influx_inspect verify-tombstone tool

e-dard:
This PR adds a new tool for verifying tombstone files. To use the tool simply point it at an arbitrary path; the default is `~/.influxdb`. 

Use the tool like this:

```
$ influx_inspect verify-tombstone
```

The tool will then find all of the `.tombstone` files recursively under that path and verify each one. Currently it verifies files serially; it would be trivial to parallelise this work in future. The path provided can be arbitrary—it doesn't need to be an InfluxDB data directory for example.

The tool iterates through each tombstone entry in each file. You can set verbosity levels to decide how much information you want:

 - no verbosity set: tool tells you when it has completed verifying a tombstone file;
 - `-v` (verbose): tool tells you when it's verifying a file and will tell you it's progress every 5 million tombstone entries;
 - `-vv (very verbose)`: tool emits **every** series key and time range in the tombstone files. Times are nano seconds since the epoch;
 - `-vvv (very very verbose)`: tool emits **every** series key and time range the tombstone files. Times are `RFC3339 Nano` formatted strings.

### DIFF
--- a/cmd/influx_inspect/main.go
+++ b/cmd/influx_inspect/main.go
@@ -18,6 +18,7 @@ import (
 	"github.com/influxdata/influxdb/cmd/influx_inspect/report"
 	"github.com/influxdata/influxdb/cmd/influx_inspect/reporttsi"
 	"github.com/influxdata/influxdb/cmd/influx_inspect/verify/seriesfile"
+	"github.com/influxdata/influxdb/cmd/influx_inspect/verify/tombstone"
 	"github.com/influxdata/influxdb/cmd/influx_inspect/verify/tsm"
 	_ "github.com/influxdata/influxdb/tsdb/engine"
 )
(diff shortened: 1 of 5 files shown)


cmd/influx_inspect/main.go
```diff
@@ -112,8 +113,13 @@ func (m *Main) Run(args ...string) error {
 		if err := name.Run(args...); err != nil {
 			return fmt.Errorf("verify-seriesfile: %s", err)
 		}
+	case "verify-tombstone":
+		name := tombstone.NewCommand()
+		if err := name.Run(args...); err != nil {
+			return fmt.Errorf("verify-seriesfile: %s", err)
```
rawkode: Should say `verify-tombstone`

cmd/influx_inspect/verify/tombstone/verify.go
```diff
@@ -0,0 +1,145 @@
...
+	vv := fs.Bool("vv", false, "very verbose: emit every tombstone entry key and time range")
+	vvv := fs.Bool("vvv", false, "very very verbose: emit every tombstone entry key and RFC3339Nano time range")
+
+	fs.SetOutput(cmd.Stdout)
+
+	if err := fs.Parse(args); err != nil {
+		return err
+	}
+
+	if *v {
```
rawkode: Could the complexity be simplified by checking `v` first, then `vv`, and finally `vvv`; using the most verbose option?

Would allow removal of complicated if statements

rawkode: That's what I said, but in a terrible fashion :)

e-dard:
@dgnorton thanks for getting to this!

PR #15960, opened 2019-11-18
chore(pkger): add tests to pkg new cli command

jsteenb2:
Add some small integration tests to validation cobra commands are all working as expected.  Inspiration taken from the `kubectl` [here](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/convert/convert_test.go#L108)

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influx/pkg.go
+++ b/cmd/influx/pkg.go
@@ -23,16 +23,18 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-func pkgCmd() *cobra.Command {
-	cmd := pkgApplyCmd()
+type pkgSVCFn func(cliReq httpClientOpts) (pkger.SVC, error)
+
+func pkgCmd(newSVCFn pkgSVCFn) *cobra.Command {
+	cmd := pkgApplyCmd(newSVCFn)
 	cmd.AddCommand(
-		pkgCreateCmd(),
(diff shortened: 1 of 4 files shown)


cmd/influx/pkg.go
```diff
@@ -23,16 +23,18 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-func pkgCmd() *cobra.Command {
-	cmd := pkgApplyCmd()
+type pkgSVCFn func(token, addr string, isLocal, skipVerify bool) (pkger.SVC, error)
```
jsteenb2: DI ftw

kelwang: I think we would like to group `token, addr string, isLocal, skipVerify bool` together as a single dependency, so when we would like to add more parameters, we only need to change one place. Also we don't have to write similar code for every one  else.

```go
     type CliRequest struct{
           Token string
            Addr  string
            IsLocal bool
            SkipVerify bool
     }

     func (c CliRequest) PkgService() *pkger.Service, error{ ... }
```

Also i think it's better if we put CliRequest in a separate  folder under cmd/influx/, other than the main package. The main package is oversized now

jsteenb2:
@kelwang pushed up changes we spoke of on the slack call 👍

PR #15971, opened 2019-11-19
feat(tracing): dont trace spans with full request paths as operation name in ExtractFromHTTPRequest

GeorgeMac:
This updates our automatic span creation from http request method in the tracing kit.

Prior to this change all request paths were being logged in spans as operation names.
This has led to many operation names (and therefore increased cardinality required).

Now we have all operations being reported as `request` along with the following tags:
- `base_path` which is an attempt to capture import resource identifiers e.g. (`/api/v2/buckets`) but not the resource identifier
- `handler_name` which is the handler which is called to invoke this operation

It also logs the full url path with the key `path`. This is for a little more context.

My proposed change has increased the number of allocations required. So I thought I would share here for your consideration:

```
Before:
BenchmarkOpentracing_ExtractFromHTTPRequest-16    	 2000000	      1003 ns/op	     672 B/op	       5 allocs/op

  After:
BenchmarkOpentracing_ExtractFromHTTPRequest-16    	 1000000	      1213 ns/op	     816 B/op	      10 allocs/op
```

Update:

So I have a PR open on `httprouter` https://github.com/julienschmidt/httprouter/pull/286

This now vendors that change and relies on it to get the router match from the request context (this what my PR allows you to do with httprouter).

This means we don't need to do any complicated deriving of some base path. Instead we tag the spans with the router path as was desired.

e.g. a handler mounted on route `/api/v2/buckets/:bucket_id` would produce the tag `route=/api/v2/buckets/:bucket_id` .

```
goos: darwin
goarch: amd64
pkg: github.com/influxdata/influxdb/kit/tracing
BenchmarkOpentracing_ExtractFromHTTPRequest-16    	 2000000	      1008 ns/op	     688 B/op	       7 allocs/op
PASS
ok  	github.com/influxdata/influxdb/kit/tracing	3.032s
```

- [x] Rebased/mergeable
- [x] Tests pass


kit/tracing/tracing.go
```diff
@@ -50,15 +50,44 @@ func InjectToHTTPRequest(span opentracing.Span, req *http.Request) {
...
+}
+
+// basePath derives a suitable base path
+// for example:
+// /foo/bar/1234 -> /foo
+// /api/v2/buckets/1234 -> /api/v2/buckets
+// /api/v1 -> /api
+// / -> /
+// "" -> /
+func basePath(path string) string {
```
jacobmarble: This doesn't seem reliable. For example:
- `/foo/bar/1234` should yield `/foo/bar`
- `/api/v2/query/suggestions` should yield `/api/v2/query/suggestions`

GeorgeMac: Agreed I wasn't happy with what I could get out of this. Hard to say how to resolve this deterministically. Since we don't know which part of the path will be a resource type vs a resource identifier. This brings us around to your suggestion about having the router match rather than just the path. I will dig into our router a bit and see what I can find.
There are other routers that allow you that context when creating middleware.

jacobmarble: I don't know how the current router was chosen, but we would need to rope in more folks to switch, IMO.

> What do you think about adjusting ExtractFromHTTPRequest to take routerPath instead of handlerName?

This option would require adjusting the current pattern across all handlers, but it gives us exactly what is needed here. Ping if you would like to discuss this or other options.

kit/tracing/tracing.go
```diff
@@ -50,15 +50,44 @@ func InjectToHTTPRequest(span opentracing.Span, req *http.Request) {
...
+		return span, req.WithContext(ctx)
+	}
+
+	span := opentracing.StartSpan("request", opentracing.ChildOf(spanContext), ext.RPCServerOption(spanContext))
+	annotateSpan(span, handlerName, req.URL.Path)
+
+	return span, req.WithContext(opentracing.ContextWithSpan(req.Context(), span))
+}
+
+func annotateSpan(span opentracing.Span, handlerName, path string) {
```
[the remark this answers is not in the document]
jacobmarble: Pulling information out of a context will certainly slow performance, though I suspect the context query will only be one or two layers deep in production.

kit/tracing/tracing.go
jacobmarble: Modifying `httprouter` is great. Looks like that repo was last touched on 5-oct, so we should expect something from him.

kit/tracing/tracing.go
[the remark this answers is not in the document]
GeorgeMac: Yeah benchmarks moved back to 7 allocs. Better than 10, but worse than 5 :)

kit/tracing/tracing.go
GeorgeMac: Yeah I would like to get us back to master ASAP.
I used the nice go mod replace so it should be a clean break switching back.

GeorgeMac:
I tagged you in this @mark-rushakoff to review the approach of doing a hard cut-over to an `influxdata` fork of httprouter.

GeorgeMac:
I will come back to this. Do a full replace everywhere means updating flux to use our fork as well.

chronograf/server/influx_test.go
```diff
@@ -10,7 +10,7 @@ import (
 
 	"github.com/influxdata/influxdb/chronograf"
 	"github.com/influxdata/influxdb/chronograf/mocks"
-	"github.com/julienschmidt/httprouter"
+	"github.com/influxdata/httprouter"
```
jacobmarble: I thought the go.mod replace was supposed to make import changes unnecessary?

GeorgeMac: Yeah as @mark-rushakoff says. This is great until the compat tests run and then it breaks :( but seems like bringing the fork under our banner works.

My latest problem is flux depends on the original version 😂 that is my next challenge.

Issue #15977, opened 2019-11-19
Creating user via cli without org renders UI unusable


jsteenb2:
I have the ability to create users via the CLI without an organization, and when I log in as one I just created without an org, the immediate page after the login is an exception and UI does not load. I'd think we'd want to force the CLI user to provide an org id for a user that is created, and we end up mapping that for them in a single call. This would eliminate this use case, but we still have gaps in the platform that wouldn't force it. 

![image](https://user-images.githubusercontent.com/17263167/69180604-699c3700-0ac2-11ea-9277-b5dcf0dec2f7.png)

__Steps to reproduce:__
List the minimal actions needed to reproduce the behavior.

1. create a user and add a password for it
2. load UI and sign in with said user
3. view page

__Expected behavior:__
Expected UI to load.

__Actual behavior:__
UI throws an exception on missing orgID.

PR #15981, opened 2019-11-19
feat: add support for setting password and org for a new user in the cli

jsteenb2:
one thing to note here is that new endpoint was created. there was no
endpoint for setting an initial password that worked. The existing endpoint
was a bit messy and coupled across multiple routes. Having multiple auth
schemes proved incredibly taxing to write against.

update: this came up in discussion in another topic with @goller and from that conversation decided to fixup the password service interface here. To provide the user id instead of name as the first arg to each set/compare func in the PasswordService.

closes: https://github.com/influxdata/influxdb/issues/15977

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influx/main.go
+++ b/cmd/influx/main.go
@@ -17,21 +17,27 @@ import (
 	"github.com/spf13/viper"
 )
 
-var influxCmd = &cobra.Command{
-	Use:   "influx",
-	Short: "Influx Client",
-	Run: func(cmd *cobra.Command, args []string) {
-		if err := checkSetup(flags.host); err != nil {
-			fmt.Printf("Note: %v\n", internal.ErrorFmt(err))
-		}
-		cmd.Usage()
-	},
-}
-
 const maxTCPConnections = 128
 
-func init() {
-	influxCmd.AddCommand(
+func main() {
+	influxCmd := influxCmd()
+	if err := influxCmd.Execute(); err != nil {
+		os.Exit(1)
+	}
+}
+
+func influxCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "influx",
+		Short: "Influx Client",
+		Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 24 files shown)


cmd/influx/main.go
```diff
@@ -17,21 +17,27 @@ import (
 	"github.com/spf13/viper"
 )
 
-var influxCmd = &cobra.Command{
```
jsteenb2: want to make this noted, the init and global state are 100% not needed. Only adds side effects and indirection in this case. Creating constructors allows the ability to provide DI for testing (see [pkg_test.go](https://github.com/influxdata/influxdb/blob/master/cmd/influx/pkg_test.go))

http/user_service.go
```diff
@@ -72,6 +73,10 @@ func NewUserHandler(b *UserBackend) *UserHandler {
 	h.HandlerFunc("GET", usersLogPath, h.handleGetUserLog)
 	h.HandlerFunc("PATCH", usersIDPath, h.handlePatchUser)
 	h.HandlerFunc("DELETE", usersIDPath, h.handleDeleteUser)
+	// the POST doesn't need to be nested under users in this scheme
+	// seems worthwhile to make this a root resource in our HTTP API
```
dearyhud: I can agree w/ this, seems kind of odd to have this nested under the /users/:id path when `id` isn't used

jsteenb2:
tests failing in idpe here, will have follow up PR in idpe to touch that up:

```
# github.com/influxdata/idpe/kv_test [github.com/influxdata/idpe/kv.test]
kv/passwords_test.go:57:44: cannot use f.Users[i].Name (type string) as type influxdb.ID in argument to svc.SetPassword
kv/passwords_test.go:310:54: cannot use tt.args.name (type string) as type influxdb.ID in argument to s.SetPassword
kv/passwords_test.go:485:58: cannot use tt.args.name (type string) as type influxdb.ID in argument to s.ComparePassword
+ record_failure 'go vet failed'
```

PR #15982, opened 2019-11-19
fix(ui): put importing task from template behind a flag

drdelambre:
The current implementation of this feature drops the user off into a black hole. I'm throwing it behind a flag so that product can still give it some love, but our users aren't exposed to it until the feature is more fleshed out.


drdelambre:
After talking to @sjwang90, we should still allow users to use their templates if they have correctly configured them, but instead redirect them to a more meaningful location while we sort out where the product is supposed to go for a user with no task templates.

also got rid of an annoying flash caused by the resource loader thanks to @alexpaxton

ui/src/tasks/components/TemplateBrowserEmpty.tsx
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  private handleButtonClick = (): void => {
+    const {router, org} = this.props
+
+    router.push(`/orgs/${org.id}/tasks/import`)
+  }
+}
+
+const mstp = ({orgs: {org}}: AppState): StateProps => ({
```
Asalem1: Yo dawg, I heard you like `orgs`, so we put `org` in your `orgs`

PR #15986, opened 2019-11-20
fix(tasks): construct permissions for finish run to write to system buckets

GeorgeMac:
This changes the way authentication for the `_tasks` system bucket is crafted within the analytical store.
Prior to this I threaded the author of the tasks authentication onto the FinishRun call. This meant the user needed the permission to write to the system tasks bucket. However, upon reflection the actor here is the system itself attempting to record something. It is archiving a run from short-term to long-term storage. Not the user.
I believe it is therefor safe for the analytical store to craft this permission itself. The downstream storage at the end of the run recorder still has to verify the authenticity of the authorization.

This closes an issue we are experiencing with backwards compatibility with legacy system bucket recording of runs.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/task/backend/analytical_storage.go
+++ b/task/backend/analytical_storage.go
@@ -10,6 +10,7 @@ import (
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/lang"
 	"github.com/influxdata/influxdb"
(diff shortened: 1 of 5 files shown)


GeorgeMac:
@stuartcarnie correct 👍

Issue #15921, opened 2019-11-14
export existing resources over CLI


jademcgough:
- [x] be able to export a resource based on ID
- [x] be able to export resources based on resource type
- [ ] be able to export resources by label name

```bash
# using colon separated args
influx pkg export --buckets=BID_1,BID_2 --labels=LID_1,LID_2 --dashboards=DID_1 --variables=VID_1,VID2

# exporting with TTY input
influx pkg export --resource-type=bucket ARGS|TTY INPUT (piped/bash doc)

# export all
influx pkg export all

# filters are match any within single filter and ANDed with multiple filters, and resource is included in output pkg
# i.e. --filter=labelName=foo --filter=resourceType=bucket 
#   tranlsates to all resoruces that are of type bucket and have an association with label with name foo

# exporting all resources filtered by labelname
influx pkg export all --filter=labelNames=foo,bar,baz 

# exporting all resources filtered by label ID
influx pkg export all --filter=labelID=id_1,id_2

# potentially by multiple filters, i.e. by label name and only buckets
influx pkg export all --filter=labelName=foo,bar,baz --filter=resourceType=bucket
```

if no argument is provided, then return error nothing provided and usage for helping user wade through the cmd

jademcgough:
closed by #16000

PR #16000, opened 2019-11-21
feat(pkger): add ability to export resources to pkg in cli

jsteenb2:
big chunk of the work for: #15921 

couple other odds and ends that got extended here, having abiliy to turn off headers for the `influx bucket find` call (no need to `grep -v` first line), and the ability to take TTY inputs for the `influx pkg` cmd. This enables inputs like `<<EOF` document inputs, which is handy for scripting.

- [x] Rebased/mergeable
- [x] Tests pass


pkger/parser.go
```diff
@@ -473,13 +473,6 @@ func (p *Pkg) graphDashboards() error {
 			}}
 		}
 
-		if _, ok := p.mDashboards[r.Name()]; ok {
```
jsteenb2: heasd up; this should have never been here, dashboards can have same names, there is no restriction on uniqueness for a dashboard

Issue #15984, opened 2019-11-19
Query Builder: Add ability to group by columns


hoorayimhelping:
Related: #15563 

It is currently only possible to generate queries of the form

```
from(bucket: "telegraf")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(<filter>)
  |> aggregateWindow(every: v.windowPeriod, fn: <agg>)
```

Since flux does a default `group by *`, it is not possible to apply a different grouping of data before aggregating.

1. Aggregate all series over time into a single series

![group-none-before-agg](https://user-images.githubusercontent.com/1477539/67502257-98ee8e00-f653-11e9-812c-b6cf723927a8.gif)

More of less the query we'll want to be able to express is

```
from(bucket: "telegraf")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r._measurement == "cpu")
  |> filter(fn: (r) => r._field == "usage_user")
  |> group()
  |> aggregateWindow(every: v.windowPeriod, fn: mean)
  |> yield(name: "mean")
```

hoorayimhelping:
POC I whipped up:
![in](https://user-images.githubusercontent.com/146112/69197203-a75e8700-0ae5-11ea-8d1d-c9c90e7a4b82.gif)

Chatted with Alex Pax about this on Slack. We both think there needs to be more affordance for switching between Filters and Groups so I'm going to make the Filter / Group a dropdown and allow selecting it.

hoorayimhelping:
Chatted with @mavarius about the design and UX of this. For the first initial pass (what I'm calling Group By 0.5) we're going to add a dropdown to select between `filter` and `group`. Adding a new `BuilderCard` will select `filter` by default, initially.
<img width="537" alt="Screen Shot 2019-11-27 at 10 49 35 AM" src="https://user-images.githubusercontent.com/146112/69752808-e9e81b00-1106-11ea-8e6f-3612048329a3.png">

After the initial 0.5 functionality lands, we'll evaluate the best way to change the function with fewer clicks. Some initial thoughts are to have a list of functions to add in place of the single `+` icon, or a dropdown above `Aggregate Functions` with a list of functions to add.
<img width="897" alt="Screen Shot 2019-11-27 at 10 50 18 AM" src="https://user-images.githubusercontent.com/146112/69752952-42b7b380-1107-11ea-8f33-4cfebca71619.png">

[...]

PR #16001, opened 2019-11-21
chore(ui): add filter and group concepts to tags in redux [WIP]

hoorayimhelping:
Connects #15984

**This is a non-user facing change.**

This sets up the redux store ahead of the UI changes to add `group()` in the query builder.

It introduces a new concept: the idea of a builder's aggregate function type. In current production, this would only be `filter`. This change is striving to add `group`, so that in the query builder, a user can toggle between filtering columns and grouping columns. 

The end goal is something resembling this (subject to change)
Screen Shot 2019-11-20 at 5 24 17 PM

**When reviewing this, please think about how this is named and if there's a better name for this concept.**

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/jest.config.js
+++ b/ui/jest.config.js
@@ -23,7 +23,7 @@ module.exports = {
     'ts-jest': {
       tsConfig: 'tsconfig.test.json',
       diagnostics: {
-        ignoreCodes: [6133] // ignore `'foo' is declared but its value is never read.`
+        ignoreCodes: [6133, 6192] // ignore unused variable errors
       },
     },
   },
--- a/ui/src/timeMachine/actions/queryBuilder.ts
+++ b/ui/src/timeMachine/actions/queryBuilder.ts
@@ -10,220 +10,113 @@ import {
 
 // Types
 import {Dispatch} from 'redux-thunk'
-import {GetState} from 'src/types'
-import {RemoteDataState} from 'src/types'
+import {GetState, RemoteDataState} from 'src/types'
+import {BuilderAggregateFunctionType} from 'src/client/generatedRoutes'
 import {BuilderFunctionsType} from '@influxdata/influx'
 
 export type Action =
-  | SetBuilderBucketSelectionAction
-  | SetBuilderBucketsAction
-  | SetBuilderBucketsStatusAction
-  | SetBuilderTagKeysAction
-  | SetBuilderTagKeysStatusAction
-  | SetBuilderTagValuesAction
-  | SetBuilderTagValuesStatusAction
-  | SetBuilderTagKeySelectionAction
-  | SetBuilderTagValuesSelectionAction
-  | AddTagSelectorAction
-  | RemoveTagSelectorAction
-  | SetFunctionsAction
-  | SelectAggregateWindowAction
-  | SetValuesSearchTermAction
-  | SetKeysSearchTermAction
-  | SetBuilderTagsStatusAction
+  | ReturnType<typeof setBuilderAggregateFunctionType>
+  | ReturnType<typeof setBuilderBucket>
+  | ReturnType<typeof setBuilderBuckets>
+  | ReturnType<typeof setBuilderBucketsStatus>
+  | ReturnType<typeof setBuilderTagKeys>
+  | ReturnType<typeof setBuilderTagKeysStatus>
+  | ReturnType<typeof setBuilderTagValues>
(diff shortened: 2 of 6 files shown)


ui/src/timeMachine/actions/queryBuilder.ts
```diff
@@ -32,6 +33,22 @@ export type Action =
   | SetKeysSearchTermAction
   | SetBuilderTagsStatusAction
 
+interface SetBuilderAggregateFunctionTypeAction {
```
drdelambre: we've been slowly moving over to a different pattern for actions. a little less verbose.
https://github.com/influxdata/influxdb/blob/master/ui/src/variables/actions/index.ts#L85

hoorayimhelping: awesome, way less dense!

ui/jest.config.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
     'ts-jest': {
       tsConfig: 'tsconfig.test.json',
       diagnostics: {
-        ignoreCodes: [6133] // ignore `'foo' is declared but its value is never read.`
+        ignoreCodes: [6133, 6192] // ignore unused variable errors
```
drdelambre: what triggered the addition?

ui/src/timeMachine/reducers/index.ts
```diff
@@ -1066,14 +1087,7 @@ const initialQueryBuilderState = (
...
     aggregateWindow: {period: 'auto'},
-    tags: builderConfig.tags.map(_ => ({
-      valuesSearchTerm: '',
-      keysSearchTerm: '',
-      keys: [],
-      keysStatus: RemoteDataState.NotStarted,
-      values: [],
-      valuesStatus: RemoteDataState.NotStarted,
-    })),
+    tags: initialStateHelper().queryBuilder.tags,
```
drdelambre: neat

ui/jest.config.js
hoorayimhelping:
```ts
import {necessaryButUnusedFunctionOne, necessaryButUnusedFunctionTwo} from 'foo'
```

triggers a `6192` which is "all imports must be used"

ui/src/timeMachine/actions/queryBuilder.ts
```diff
@@ -10,220 +10,113 @@ import {
...
+  | ReturnType<typeof setKeysSearchTerm>
+  | ReturnType<typeof setBuilderTagsStatus>
+
+export const setBuilderAggregateFunctionType = (
+  builderAggregateFunctionType: BuilderAggregateFunctionType,
+  index: number
+) => ({
+  type: 'SET_BUILDER_AGGREGATE_FUNCTION_TYPE' as 'SET_BUILDER_AGGREGATE_FUNCTION_TYPE',
+  payload: {builderAggregateFunctionType, index},
+})
```
hoorayimhelping: apologies for having this diff include non-logical structural changes in this file (it's all in service of changing out actions to be less verbose)

this block is the main logical addition to this file

ui/src/timeMachine/reducers/index.test.ts
```diff
@@ -0,0 +1,52 @@
...
+  initialStateHelper,
+  timeMachinesReducer,
+} from 'src/timeMachine/reducers'
+
+/*export const selectBucket = (
+  bucket: string,
+  resetSelections: boolean = false
+)*/
+
+/*export const addTagSelector = () => (
```
drdelambre: these should be removed

ui/src/timeMachine/reducers/index.ts
```diff
@@ -649,6 +652,17 @@ export const timeMachineReducer = (
       })
     }
 
+    case 'SET_BUILDER_AGGREGATE_FUNCTION_TYPE': {
+      return produce(state, draftState => {
+        const {index, builderAggregateFunctionType} = action.payload
+        const draftQuery = draftState.draftQueries[draftState.activeQueryIndex]
+
+        draftQuery.builderConfig.tags[
```
drdelambre: this should check if index is a valid tag before trying to access it

hoorayimhelping: i added the the checks but i'm like 99% certain that the shape of the data in this reducer is consistent with default values

PR #16013, opened 2019-11-21
fix(ui): update the meta tag

### LINKED ISSUE #15780: Mobile display not working on devices - missing <meta> tag
Steps to reproduce:
Either 
-- open the Cloud app in a mobile device and navigate to any page OTHER than usage/billing
-- open in a desktop browser, using browser tools, set to mobile device display

Expected behavior:
Mobile device settings should be triggered (mobile slideout menu, responsive layout)

Actual behavior:
Mobile layout is seen in browser, when in desktop mode, and the browser width is reduced --- but NOT when viewing on a device (or simulating a device). Screen is being displayed full-screen, but scaled to fit.

Suggested Solution:
It appears that monitoring pages don't have the appropriate <meta> tags to trigger display for mobile devices
Sample:  <meta name="viewport" content="width=device-width, initial-scale=1">
These tags are being used for the ecomm/quartz pages, and when I copied over the meta tag to the browser session, it corrected the issue

drdelambre:
Closes #15780

decided to update the meta tag to a newer version that is more inline with quartz after evaluating this ticket

### DIFF
--- a/ui/assets/index.html
+++ b/ui/assets/index.html
@@ -1,11 +1,8 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <meta
-      http-equiv="Content-type"
-      content="text/html; charset=utf-8 width=device-width, initial-scale=1"
-      name="viewport"
-    />
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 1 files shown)


drdelambre:
i'm hoping so too!

PR #16018, opened 2019-11-22
chore(pkger): fixing up the pkger parse error to respect the initial Error call

jsteenb2:
the errors changed during the pkger http server error improvements and this
fixes it to be similar to what it once was, only now its in a flatter fashion
very similar to the pkger http server apply response.

- [x] Rebased/mergeable
- [ ] Tests pass


jsteenb2:
@bthesorceror that's a great point, i forgot to rename that after bringing in the replacement at that same name 🤦‍♂. Will touch it up now 👍

PR #16019, opened 2019-11-22
fix(kv): Add push down predicate hint to filter by keys and values

stuartcarnie:
A KV store may optionally implement a key predicate function to filter on keys or values, which may reduce memory and CPU usage.

Expected `FindByTaskID` resource mapping improvements for a single user:

```
Before:
206966085 ns/op   37672164 B/op   445060 allocs/op

After:
  1514118 ns/op      11184 B/op      131 allocs/op

```


aanthony1243:
didn't see any major problems.  good tests, what you did is straightforward.

PR #16027, opened 2019-11-22
fix(ui): value zero has no background color

### LINKED ISSUE #15803: Background color doesn't set when field value = 0
backgroundtext_0
<!--

zoesteinkamp:
Now fixed, falsy zero is no longer being bad

Closes #15803 

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/src/shared/constants/colorOperations.ts
+++ b/ui/src/shared/constants/colorOperations.ts
@@ -53,7 +53,7 @@ export const generateThresholdsListHexs = ({
   const baseColor = colors.find(color => color.id === BASE_THRESHOLD_ID) || {
     hex: defaultColoring.textColor,
   }
-  
+
   if (lastValue != 0 && !lastValue) {
     return {...defaultColoring, textColor: baseColor.hex}
   }
--- a/ui/src/shared/constants/colorOperations.ts
+++ b/ui/src/shared/constants/colorOperations.ts
@@ -53,8 +53,8 @@ export const generateThresholdsListHexs = ({
   const baseColor = colors.find(color => color.id === BASE_THRESHOLD_ID) || {
     hex: defaultColoring.textColor,
   }
-
-  if (!lastValue) {
+  
+  if (lastValue != 0 && !lastValue) {
     return {...defaultColoring, textColor: baseColor.hex}
   }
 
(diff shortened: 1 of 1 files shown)


ui/src/shared/constants/colorOperations.ts
```diff
@@ -54,7 +54,7 @@ export const generateThresholdsListHexs = ({
     hex: defaultColoring.textColor,
   }
 
-  if (!lastValue) {
+  if (lastValue != 0 && !lastValue) {
```
hoorayimhelping: first, prefer strict equality checking `!==` here:
```js
0 == '0' // true
0 === '0' // false
```
because '0' might pass this 😱 


the second bit is kind of pedantic, but stick with me.

when designing a boolean expression, i like to think of it like a funnel, where it starts wide and broad and gets narrow and specific. i want my widest and broadest expressions first, and my more specific ones last.

i think `!lastValue` is more broad than `lastValue !== 0`: `lastValue !== 0` includes truthy expressions, which is definitely not what we want in this. we want only the falsy expressions that are not 0 (so `''`, and `null` according to [the type hint here](https://github.com/influxdata/influxdb/blob/de4397874b3bdb425d05a4301121fd72f055ee09/ui/src/shared/constants/colorOperations.ts#L38)).

i also like to think about the flow of an `if / then` expression. if we factored

```ts
if (lastValue !== 0 && !lastValue) {
  return {...defaultColoring, textColor: baseColor.hex}
}
```
into separate expressions, it would look like this:

```ts
if (lastValue !== 0) {
  // lastValue can equal any string that isn't '' here
  if (!lastValue) {
    return {...defaultColoring, textColor: baseColor.hex}
  }
}
```

#### now imagine if we flipped the order:
```ts
if (!lastValue && lastValue !== 0) {
  return {...defaultColoring, textColor: baseColor.hex}
}
```
it would factor out to
```ts
if (!lastValue) {
  // last value can only equal null, '', or 0 here
  if (lastValue !== 0) {
    return {...defaultColoring, textColor: baseColor.hex}
  }
}
```
which i think is what we'd rather have here - last value is `falsy,` and it's not the number `0`. it's a subtle thing, but it communicates a lot about intent.

how does all of that strike you?

zoesteinkamp: This seems perfect! even better! Im going to go make the change :)

PR #16029, opened 2019-11-22
feat(pkger): move bucket pkger model to reflect http APIs retention rules

### LINKED ISSUE #15987: Update Bucket Pkger Model to use retention rules
based on feedback on API design, we want to tweak the bucket model to use retention rules instead of retention period to reflect the public API into pkger (learn once, apply everywhere). Minimizing overhead for users new to pkgere or vice versa with the HTTP API.

jsteenb2:
Closes #15987 

after discussion with @goller, we'd like pkger to reflect teh same API as the http bucket api for rentention rules in place of retention period.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/pkger/models.go
+++ b/pkger/models.go
@@ -3,6 +3,7 @@ package pkger
 import (
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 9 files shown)


pkger/models.go
```diff
@@ -1002,14 +1070,14 @@ func (c chart) validBaseProps() []validationErr {
...
-			Field: "width",
+			Field: fieldChartWidth,
 			Msg:   "must be greater than 0",
 		})
 	}
 
 	if c.Height <= 0 {
 		fails = append(fails, validationErr{
-			Field: "height",
+			Field: fieldChartHeight,
```
jsteenb2: all these changes are to couple the field names of the errors with the ones used during parsing, to avoid drift

PR #16031, opened 2019-11-22
feat(ui): test notification endpoints

### LINKED ISSUE #14792: Add endpoint test to endpoint configuration
Allow a test message to be sent via provided configuration information.  This allows the person configuring the endpoint to verify that the information provided is working.

### LINKED ISSUE #15157: Test endpoint UI

121watts:
Closes #14792

Kapture 2019-11-22 at 11 20 40

### DIFF
--- a/http/notification_endpoint.go
+++ b/http/notification_endpoint.go
@@ -6,11 +6,16 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"time"
 
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/csv"
+	"github.com/influxdata/flux/lang"
 	"github.com/influxdata/httprouter"
 	"github.com/influxdata/influxdb"
 	pctx "github.com/influxdata/influxdb/context"
 	"github.com/influxdata/influxdb/notification/endpoint"
+	"github.com/influxdata/influxdb/query"
 	"go.uber.org/zap"
 )
 
@@ -26,6 +31,7 @@ type NotificationEndpointBackend struct {
 	UserService                 influxdb.UserService
 	OrganizationService         influxdb.OrganizationService
 	SecretService               influxdb.SecretService
+	QueryService                query.ProxyQueryService
 }
 
 // NewNotificationEndpointBackend returns a new instance of NotificationEndpointBackend.
@@ -40,6 +46,7 @@ func NewNotificationEndpointBackend(b *APIBackend) *NotificationEndpointBackend 
 		UserService:                 b.UserService,
 		OrganizationService:         b.OrganizationService,
 		SecretService:               b.SecretService,
+		QueryService:                b.FluxService,
 	}
 }
 
@@ -55,11 +62,14 @@ type NotificationEndpointHandler struct {
 	UserService                 influxdb.UserService
 	OrganizationService         influxdb.OrganizationService
 	SecretService               influxdb.SecretService
+	QueryService                query.ProxyQueryService
 }
 
 const (
 	notificationEndpointsPath            = "/api/v2/notificationEndpoints"
 	notificationEndpointsIDPath          = "/api/v2/notificationEndpoints/:id"
+	notificationEndpointsIDTestPath      = "/api/v2/notificationEndpoints/:id/test"
+	notificationEndpointsTestPath        = "/api/v2/notificationEndpointsTest"
 	notificationEndpointsIDMembersPath   = "/api/v2/notificationEndpoints/:id/members"
 	notificationEndpointsIDMembersIDPath = "/api/v2/notificationEndpoints/:id/members/:userID"
 	notificationEndpointsIDOwnersPath    = "/api/v2/notificationEndpoints/:id/owners"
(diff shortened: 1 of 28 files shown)


http/notification_endpoint.go
```diff
@@ -229,7 +242,6 @@ func (h *NotificationEndpointHandler) handleGetNotificationEndpoints(w http.Resp
 		h.HandleHTTPError(ctx, err, w)
 		return
 	}
-	h.Logger.Debug("notificationEndpoints retrieved", zap.String("notificationEndpoints", fmt.Sprint(edps)))
```
kelwang: Um, what's the reason to delete the debug message?

http/notification_endpoint.go
```diff
@@ -403,6 +414,134 @@ func decodePatchNotificationEndpointRequest(ctx context.Context, r *http.Request
...
+		return
+	}
+
+	auth, err := pctx.GetAuthorizer(ctx)
+	if err != nil {
+		h.HandleHTTPError(ctx, err, w)
+		return
+	}
+
+	var token *influxdb.Authorization
```
kelwang: This looks too hacky. @influxdata/compute-team has cleaned this up already for task. We should make sure the query proxy server can handle the context.

http/notification_endpoint.go
```diff
@@ -500,6 +637,7 @@ func (h *NotificationEndpointHandler) handlePutNotificationEndpoint(w http.Respo
 
 	for _, fld := range edp.SecretFields() {
 		if fld.Value != nil {
+			fmt.Printf("hello: %v\n", *fld.Value)
```
kelwang: remove this debug msg

notification_endpoint.go
```diff
@@ -129,5 +131,8 @@ type NotificationEndpointService interface {
 	PatchNotificationEndpoint(ctx context.Context, id ID, upd NotificationEndpointUpdate) (NotificationEndpoint, error)
 
 	// DeleteNotificationEndpoint removes a notification endpoint by ID, returns secret fields, orgID for further deletion.
-	DeleteNotificationEndpoint(ctx context.Context, id ID) (flds []SecretField, orgID ID, err error)
+	DeleteNotificationEndpoint(ctx context.Context, id ID) (flds []*SecretField, orgID ID, err error)
+
+	// TestNotificationEndpoint tests the an endpoints connection to a 3rd party service i.e. Slack, PagerDuty etc.
+	TestNotificationEndpoint(ctx context.Context, ne NotificationEndpoint, userID ID) error
```
kelwang: A method prefix Test is like a reserved keyword to run test uniting in golang, thus might confuse people. If you can give an another name, it would be much preferred.

121watts: Will do

http/notification_endpoint.go
121watts: This leaks secrets in the debug messages.  If anyone has any sensitive info in their headers this would be in our logs.

http/notification_endpoint.go
desa: @kelwang this is what is currently done in the `/api/v2/query` endpoint. I agree it's hacky and that we need a long term solution, but I'm not sure exactly what that looks like.

ui/src/alerting/components/endpoints/EndpointOverlayFooter.tsx
```diff
@@ -12,39 +12,71 @@ import {
...
-interface Props {
+export interface Message {
+  text: string
+  remoteState: RemoteDataState
+}
+interface OwnProps {
   saveButtonText: string
-  onSave: (endpoint: NotificationEndpoint) => void
+  onSave: (endpoint: NotificationEndpoint) => Promise<void>
+  onTest: (endpoint: NotificationEndpoint) => Promise<void>
```
Asalem1: Is there a reason why we've passed `onTest` all the way down here without using it?

ui/src/alerting/actions/notifications/endpoints.ts
```diff
@@ -101,10 +104,47 @@ export const createEndpoint = (endpoint: NotificationEndpoint) => async (
...
+  const labels = endpoint.labels || []
+
+  const data = {
+    ...endpoint,
+    labels: labels.map(l => l.id),
+  } as PostNotificationEndpoint
+
+  const resp = await api.postNotificationEndpointsTest({data})
+
+  if (resp.status !== 204) {
```
Asalem1: It doesn't seem like `testEndpoint` ends being used in some instances, would it be more beneficial to dispatch a notification to the user?

ui/src/alerting/components/endpoints/EndpointOverlayContents.tsx
```diff
@@ -111,13 +115,21 @@ const EndpointOverlayContents: FC<Props> = ({
...
+                  }
+                  icon={
+                    message.remoteState === 'Done'
+                      ? IconFont.Checkmark
+                      : IconFont.AlertTriangle
+                  }
                   style={{width: 'auto', marginTop: '8px'}}
                 >
-                  {errorMessage}
+                  {message.text}
```
Asalem1: What if the `text` is undefined? Should we have a generic message to let the user know that something wrong happened?

PR #16040, opened 2019-11-25
fix(ui): querying a time range without relevant data should show no data for that part of the timerange

### LINKED ISSUE #15908: Graph time range should be based on query time range not returned data
Screenshot 2019-11-14 09 44 59

Note how I've queries for the last 5m, but the domain of the x axis only spans the last 2m. This is because I did not have any data for 3 of the last 5 minutes. This can be confusing when you have graphs where you expect the domains to be similar.

Ideally I could specify whether I'd like the domain to span the entire range queried or the range or the data returned.

Asalem1:
Closes #15908 

### Problem

Graphed time ranges only displayed the timerange based on the results within the selected timerange. That is to say that if a user was looking for data in the past hour, but only had data for the last 30 minutes, the graph would trim the displayed data to show a full graph for the last 30 minutes rather than displaying a half-empty graph for the data that wasn't there.

### Solution

Since the data being returned from the query dictates the boundaries of the x-axis, and since that data that is returned doesn't account for non-existent data (meaning, if no data exists for a given time range, that part of the time range is simply not returned), the solution to the problem was to simply determine the boundaries based on the original input. 


ui/src/timeMachine/selectors/index.ts
```diff
@@ -249,3 +250,75 @@ export const getSaveableView = (state: AppState): QueryView & {id?: string} => {
 
   return saveableView
 }
+
+export const getStartTime = timeRange => {
+  const {lower} = timeRange
+  switch (lower) {
+    case 'now() - 30d':
```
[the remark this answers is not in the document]
ebb-tide: let's talk about this one if you want

ui/src/shared/components/ViewSwitcher.tsx
```diff
@@ -186,4 +206,13 @@ const ViewSwitcher: FunctionComponent<Props> = ({
   }
 }
 
-export default ViewSwitcher
+const mstp = ({timeMachines}: AppState) => {
+  const {activeTimeMachineID, timeMachines: machine} = timeMachines
+  const {timeRange} = machine[activeTimeMachineID]
+  return {
+    startTime: timeRange ? getStartTime(timeRange) : Infinity,
+    endTime: timeRange ? getEndTime(timeRange) : null,
```
ebb-tide: it would be cleaner if the `getStartTime` and `getEndTime` functions handled returning infinity/null in cases where `timeRange` is undefined. I also have a feeling getEndTime should maybe return -Infinity? or startTime should default to `null`, it feels brittle to need to know that their defaults are different.

ui/src/timeMachine/selectors/index.ts
```diff
@@ -249,3 +250,75 @@ export const getSaveableView = (state: AppState): QueryView & {id?: string} => {
...
+      return (
+        moment()
+          .subtract(30, 'days')
+          .unix() * 1000
+      )
+    case 'now() - 7d':
+      return (
+        moment()
```
ebb-tide: I am worried about when this moment() will be instantiated..... doing some tests.

ui/src/shared/components/ViewSwitcher.tsx
```diff
@@ -36,12 +41,19 @@ interface Props {
   statuses: StatusRow[][]
 }
 
-const ViewSwitcher: FunctionComponent<Props> = ({
+interface StateProps {
+  endTime: number
+  startTime: number
+}
+
+const ViewSwitcher: FunctionComponent<Props & StateProps> = ({
```
ebb-tide: I think we should implement start and end time in CheckPlot as well. :)

ui/src/shared/components/HeatmapPlot.tsx
```diff
@@ -85,6 +89,9 @@ const HeatmapPlot: FunctionComponent<Props> = ({
     timeZone,
   })
 
+  xDomain[0] = Math.min(startTime, xDomain[0])
+  xDomain[1] = Math.max(endTime, xDomain[1])
```
hoorayimhelping: i noticed that `xDomain` is coming from `useVisDomainSettings` which seems to be used in a lot of plots. Do all of those plots need the same treatment, or just `ScatterPlot` and `HeatmapPlot`?

i think it's probably clearer and less error prone and easier to reason about to assign these values to a new variable, rather than reassigning them to `xDomain`.  the compiler would complain about this:

```ts
const singleNumberValue = useOtherDomainSettings()
singleNumbervalue = Math.max(someOtherDamnVariable, singleNumberValue)
```

```suggestion
  const remappedXDomain = [Math.min(startTime, xDomain[0]), Math.max(endTime, xDomain[1])]
```

ui/src/shared/components/ViewSwitcher.tsx
```diff
@@ -186,4 +206,13 @@ const ViewSwitcher: FunctionComponent<Props> = ({
   }
 }
 
-export default ViewSwitcher
+const mstp = ({timeMachines}: AppState) => {
+  const {activeTimeMachineID, timeMachines: machine} = timeMachines
+  const {timeRange} = machine[activeTimeMachineID]
+  return {
+    startTime: timeRange ? getStartTime(timeRange) : Infinity,
```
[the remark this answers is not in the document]
Asalem1: Theoretically, any a timeRange with the start time defaulted to 0 would pretty much always be true, whereas setting the default to `Infinity` allows us to pass in a parameter that will default to whatever the start time is based on the data that's returned

ui/src/timeMachine/selectors/index.ts
```diff
@@ -249,3 +250,75 @@ export const getSaveableView = (state: AppState): QueryView & {id?: string} => {
...
 }
+
+export const getStartTime = timeRange => {
+  const {lower} = timeRange
+  switch (lower) {
+    case 'now() - 30d':
+      return (
+        moment()
+          .subtract(30, 'days')
+          .unix() * 1000
```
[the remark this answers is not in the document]
Asalem1: great to know, i'll go ahead and change that over

ui/src/timeMachine/selectors/index.ts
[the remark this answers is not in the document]
ebb-tide: Punted for a separate PR turns out timeRange types are very weird.

ui/src/shared/components/ViewSwitcher.tsx
ebb-tide: I was wrong, nevermind!

ui/src/dashboards/selectors/index.ts
```diff
@@ -16,6 +27,12 @@ export const getViewStatus = (state: AppState, id: string): RemoteDataState => {
   return get(state, `views.views.${id}.status`, RemoteDataState.Loading)
 }
 
+export const getTimeRangeByDashboardID = (
```
hoorayimhelping: probably a good idea to write a test for this too since so many parts of the UI now depend on this

ui/src/shared/utils/useVisDomainSettings.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    expect(beginning).toEqual(data[0])
+  })
+  it("should return the endTime as endTime if it's before the last time in the data array", () => {
+    const range = getValidRange(data, startTime, endTime)
+    expect(range[1]).toEqual(endTime)
+    const newRange = getValidRange(data, endTime, startTime)
+    expect(newRange[1]).toEqual(data[data.length - 1])
+  })
+  it('should return the the start and end times based on the data array if no start / endTime are passed', () => {
+    expect(getValidRange(data)).toEqual([data[0], data[data.length - 1]])
```
ebb-tide: awesome!

ui/src/dashboards/selectors/index.test.ts
```diff
@@ -0,0 +1,40 @@
+// Funcs
+import {getTimeRangeByDashboardID} from 'src/dashboards/selectors/index'
+
+// Types
+import {Range} from 'src/dashboards/reducers/ranges'
+
+// Constants
+import {DEFAULT_TIME_RANGE} from 'src/shared/constants/timeRanges'
+
+describe('Dashboards.Selector', () => {
```
ebb-tide: ❤️!

ui/src/shared/utils/useVisDomainSettings.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    1573767860000,
+    1573767870000,
+    1573767880000,
+    1573767890000,
+    1573767900000,
+    1573767910000,
+    1573767920000,
+    1573767930000,
+    1573767940000,
+  ]
```
hoorayimhelping: i think this belongs in https://github.com/influxdata/influxdb/blob/master/ui/mocks/dummyData.ts

ui/src/dashboards/selectors/index.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    expect(getTimeRangeByDashboardID(ranges, 'Oogum Boogum')).toEqual(
+      DEFAULT_TIME_RANGE
+    )
+  })
+  it('should return the the default range when no ranges are passed in', () => {
+    expect(getTimeRangeByDashboardID([], dashboardID)).toEqual(
+      DEFAULT_TIME_RANGE
+    )
+  })
+})
```
hoorayimhelping: ❤️

ui/src/shared/utils/useVisDomainSettings.ts
```diff
@@ -8,22 +8,38 @@ import {extent} from 'src/shared/utils/vis'
...
+) => {
+  const range = extent((data as number[]) || [])
+  if (range && range.length >= 2) {
+    const start = Math.min(startTime, range[0])
+    const end = Math.max(endTime, range[1])
+    return [start, end]
+  }
+  return range
+}
+
```
hoorayimhelping: good change! note how the difficulty and oddness of testing this clued you in to a design smell - `useVisDomainSettings` might have been doing too much, both front-endy stuff with `useMemo` and calculating the valid range. the act of testing forces one to think, which can be very powerful. noice!

ui/src/shared/utils/useVisDomainSettings.test.ts
```diff
@@ -0,0 +1,134 @@
...
+  it("should return the endTime as endTime if it's before the last time in the data array", () => {
+    const range = getValidRange(data, startTime, endTime)
+    expect(range[1]).toEqual(endTime)
+    const newRange = getValidRange(data, endTime, startTime)
+    expect(newRange[1]).toEqual(data[data.length - 1])
+  })
+  it('should return the the start and end times based on the data array if no start / endTime are passed', () => {
+    expect(getValidRange(data)).toEqual([data[0], data[data.length - 1]])
+  })
+})
```
hoorayimhelping: :heart:

ui/src/timeMachine/selectors/index.ts
```diff
@@ -249,3 +250,63 @@ export const getSaveableView = (state: AppState): QueryView & {id?: string} => {
...
+    return null
+  }
+  const {upper} = timeRange
+  switch (upper) {
+    case null:
+      return null
+    default:
+      return moment(upper).valueOf()
+  }
+}
```
hoorayimhelping: you know what i'm gonna say about this right?

add some tests to `getStartTime` and `getEndTime`. should be very straightforward and will do a few things:

1. clarify your thinking on this by forcing you to think through the cases these functions handle
1. document the behavior of these functions
1. cover us against regressions in this behavior
1. reveal any small bugs that might have slipped through the cracks

i know you know all this, but it's valuable to reiterate it from time to time :)

ui/src/timeMachine/selectors/index.test.ts
```diff
@@ -0,0 +1,119 @@
...
+    expect(getEndTime(timeRange)).toEqual(newYears)
+  })
+  it(`getEndTime should return null when upper is null`, () => {
+    const timeRange = {
+      lower: 'now() - 30d',
+      upper: null,
+    }
+    expect(getEndTime(timeRange)).toEqual(null)
+  })
+})
```
hoorayimhelping: [excited yelping noises]

PR #16041, opened 2019-11-25
feat(storage/reads): add cache to reuse tags when reading from storage

### LINKED ISSUE #15985: Optimize memory usage for bounds and tags in the influxdb storage reader
The storage reader for influxdb that is used by flux will add bounds and tags to each of the tables it reads and constructs. These can consume a lot of memory when reading a lot of data, but the arrays that are created can be shared.

This issue is to optimize the memory usage for these columns to improve read performance.

jsternberg:
This adds an lru cache for the columns that are produced as tags. When
producing the columns that are part of the group key, it will generate
the column and then keep it in an lru cache to reuse for future tables.
The start and stop column are effectively cached for every table because
they are special and will be the same for all of the tables.

For the tags, it retains the most recently used since they may be used
by a future table. That way most of the columns will get shared with
each other.

When the size differs, a slice is used so the underlying data is still
shared, but the size is different.

Fixes #15985.

### DIFF
--- a/storage/reads/reader.go
+++ b/storage/reads/reader.go
@@ -36,6 +36,7 @@ func (r *storeReader) ReadFilter(ctx context.Context, spec influxdb.ReadFilterSp
 		ctx:   ctx,
 		s:     r.s,
 		spec:  spec,
+		cache: newTagsCache(0),
 		alloc: alloc,
 	}, nil
 }
@@ -45,6 +46,7 @@ func (r *storeReader) ReadGroup(ctx context.Context, spec influxdb.ReadGroupSpec
 		ctx:   ctx,
 		s:     r.s,
 		spec:  spec,
+		cache: newTagsCache(0),
 		alloc: alloc,
 	}, nil
 }
@@ -96,6 +98,7 @@ type filterIterator struct {
 	s     Store
 	spec  influxdb.ReadFilterSpec
 	stats cursors.CursorStats
+	cache *tagsCache
 	alloc *memory.Allocator
 }
 
@@ -155,6 +158,7 @@ func (fi *filterIterator) handleRead(f func(flux.Table) error, rs ResultSet) err
 			cur.Close()
 		}
 		rs.Close()
+		fi.cache.Release()
 	}()
 
 READ:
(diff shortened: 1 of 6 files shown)


jsternberg:
Performance metrics from the benchmark.

### Only cache start/stop columns

```
jsternberg:[~/g/s/g/i/influxdb] $ (perf/table-read $<>) benchcmp old.txt new.txt
benchmark                  old ns/op      new ns/op      delta
BenchmarkReadFilter-12     1291650820     1135894099     -12.06%

benchmark                  old allocs     new allocs     delta
BenchmarkReadFilter-12     1350688        1170693        -13.33%

benchmark                  old bytes      new bytes     delta
BenchmarkReadFilter-12     1379473016     983649672     -28.69%
```

### Cache both tags and start/stop column

```
jsternberg:[~/g/s/g/i/influxdb] $ (perf/table-read $<>) benchcmp old.txt new2.txt
benchmark                  old ns/op      new ns/op     delta
BenchmarkReadFilter-12     1291650820     523629173     -59.46%

benchmark                  old allocs     new allocs     delta
BenchmarkReadFilter-12     1350688        833994         -38.25%

benchmark                  old bytes      new bytes     delta
BenchmarkReadFilter-12     1379473016     577756776     -58.12%
```

### Switch the order of t0 and t1 so t0 has a larger cardinality

```
jsternberg:[~/g/s/g/i/influxdb] $ (perf/table-read *$<>) benchcmp old.txt new3.txt
benchmark                  old ns/op      new ns/op     delta
BenchmarkReadFilter-12     1291650820     343158702     -73.43%

benchmark                  old allocs     new allocs     delta
BenchmarkReadFilter-12     1350688        746656         -44.72%

benchmark                  old bytes      new bytes     delta
BenchmarkReadFilter-12     1379473016     420456485     -69.52%
```

This optimization is heavily dependent on the order of the tables coming out of storage, but it does seem like it improves all situations. In the future, we can also improve the performance here by lazily creating the arrow arrays as shown [here](https://github.com/influxdata/flux/pull/2190). That should reduce the need for the lru cache, but not eliminate the need because we would still want to share the column data between these for when the actual column is materialized.

jsternberg:
@nathanielc done.

jsternberg:
Also to clarify, there's no assumptions about the order of how stuff comes out of the cache. I was just noting that the performance can change depending on that order so it's a bit hard to benchmark. I added that note as a comment.

PR #16050, opened 2019-11-26
fix(endpoint): when looking up a endpoint we should allow org only lookup

lyondhill:
In the current system the api always adds "UserID" to the filter. This only
allows the system to look up endpoints that user created. The behavior should be
that we filter based on user input and use authorizer to hide things they shouldn't see.

### DIFF
--- a/authorizer/notification_endpoint.go
+++ b/authorizer/notification_endpoint.go
@@ -45,6 +45,14 @@ func (s *NotificationEndpointService) FindNotificationEndpointByID(ctx context.C
 
 // FindNotificationEndpoints retrieves all notification endpoints that match the provided filter and then filters the list down to only the resources that are authorized.
 func (s *NotificationEndpointService) FindNotificationEndpoints(ctx context.Context, filter influxdb.NotificationEndpointFilter, opt ...influxdb.FindOptions) ([]influxdb.NotificationEndpoint, int, error) {
+	// TODO: This is a temporary fix as to not fetch the entire collection when no filter is provided.
+	if !filter.UserID.Valid() && filter.OrgID == nil {
+		return nil, 0, &influxdb.Error{
+			Code: influxdb.EUnauthorized,
+			Msg:  "cannot process a request without a org or user filter",
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


kelwang:
As the current implementation Authorizer is the last step filter to get all responses. We used to seen an issue, if user make a request /API/v2/endpoints, if the server has for example 2000 endpoints, it will pull like  the first 100 points of all orgs. Then filter by authorizer, so we will very likely see less endpoints, if the user don't have the org parameter. If we want to get rid of this requirement, we have to make the org parameter required or extract the org from the authorizer which is not possible yet, since the session auth has no org id.

desa:
@kelwang 
>  if the server has for example 2000 endpoints, it will pull like the first 100 points of all orgs. 
That sounds like there's a bug in the authorizer then.

lyondhill:
So we would need a temporary solution that ensure's the userID or the org is present? That is until the bug that @desa identified can be resolved?

lyondhill:
the Other option is to error in the auth if you attempt to lookup the entire collection. (ie no filter)

desa:
@kelwang I actually don't understand why this wont fix the issue that @lyondhill is seeing.

lyondhill:
It does fix my issue (tested on oss) but if someone where to hi the api with no filter then the results that are not authorized are filtered from the list. Add in pagination and the user may get a list of 0 if enough org's exist that the user doesn't have access to.

The solution to that problem is either :
1 - reject any "unfiltered" lookups to the colection
2 - Force a user filter if they dont specify any filters.

@desa which of these (or any additional options) would you like to see?

desa:
@lyondhill Thanks for clarifying that. That makes a lot of sense now. It sounds like there's a bug in pagination. I'd like to see pagination implemented correctly. That would solve the issue correct?

lyondhill:
if pagination happened after authorization then yes. It would be a pretty big refactor to the way its being done now. I believe this pagination problem would exist for most the other collections in our system. The problem is that pagination happens on the database layer and for all list's the auth happens as a wrapper around the db using the `/authorizer` package.

PR #16056, opened 2019-11-27
fix(ui): DWP should default bucket name to selected bucket

### LINKED ISSUE #16011: Delete with predicate state persisted across data explorer and buckets page
dwp

Note how the selected bucket in the delete with predicate overlay is not the same as the bucket that I clicked on.

Asalem1:
Closes #16011

### Problem

Selecting to delete with predicate from an active bucket was not updating the selected bucket once an initial bucket was set

### Solution

The solution to this problem was to separate some of the logic from where the initial bucket name was being received, and also making sure to reset the deleteWithPredicate reducer component to its default state when the overlay is closed.

Buckets:
buckets-dwp

Data Explorer:
de-dwp

### DIFF
--- a/ui/src/shared/components/DeleteDataOverlay.tsx
+++ b/ui/src/shared/components/DeleteDataOverlay.tsx
@@ -2,46 +2,53 @@
 import React, {FunctionComponent} from 'react'
 import {connect} from 'react-redux'
 import {withRouter, WithRouterProps} from 'react-router'
-import {Overlay} from '@influxdata/clockface'
-import {get} from 'lodash'
+import {Overlay, SpinnerContainer, TechnoSpinner} from '@influxdata/clockface'
 
 // Components
 import DeleteDataForm from 'src/shared/components/DeleteDataForm/DeleteDataForm'
 
 // Types
-import {Bucket, AppState} from 'src/types'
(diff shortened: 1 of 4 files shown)


ui/src/shared/components/DeleteDataOverlay.tsx
```diff
@@ -2,61 +2,70 @@
...
-  const handleDismiss = () =>
+  const handleDismiss = () => {
+    resetPredicateStateAction()
     router.push(`/orgs/${orgID}/load-data/buckets/${bucketID}`)
-  // separated find logic and name logic since directly routing the a delete-data
-  // endpoint was crashing the app because the bucket is undefined until the component mounts
+  }
   const bucket = buckets.find(bucket => bucket.id === bucketID)
-  const bucketName = bucket && bucket.name ? bucket.name : ''
-  const initialBucketName = selectedBucketName || bucketName
```
drdelambre: how does not having a selected bucket fix the not having a selected bucket issue?

drdelambre: ie: should we instead be making sure that the selected bucket is being correctly maintained throughout the application's lifecycle

PR #16057, opened 2019-11-27
feat(http): add cell properties to dashboard response

### LINKED ISSUE #16052: return nested data in dashboard get request
Getting the required data for a dashboard is a pain. It requires multiple API requests to get the dashboard, its cells, and the nested views for each cell.

In order to make it a non-breaking change, we should pass an optional param to the `GET /dashboards/:id` endpoint to have it return all of that data in a single request.

ref #15898

dearyhud:
Closes #16052

Adds an optional `?include=properties` query string to `GET /dashboards/:dashboardID` that returns each cells properties

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/dashboard_service.go
+++ b/http/dashboard_service.go
@@ -210,6 +210,32 @@ type dashboardCellResponse struct {
 	Links map[string]string `json:"links"`
 }
 
+func (d *dashboardCellResponse) MarshalJSON() ([]byte, error) {
+	r := struct {
+		ID             platform.ID             `json:"id,omitempty"`
+		X              int32                   `json:"x"`
+		Y              int32                   `json:"y"`
+		W              int32                   `json:"w"`
+		H              int32                   `json:"h"`
+		Name           string                  `json:"name,omitempty"`
+		ViewProperties platform.ViewProperties `json:"properties,omitempty"`
+		Links          map[string]string       `json:"links"`
+	}{
+		ID:    d.Cell.ID,
+		X:     d.Cell.X,
+		Y:     d.Cell.Y,
+		W:     d.Cell.W,
+		H:     d.Cell.H,
+		Links: d.Links,
+	}
+
+	if d.Cell.View != nil {
+		r.ViewProperties = d.Cell.View.Properties
+		r.Name = d.Cell.View.Name
+	}
+	return json.Marshal(r)
+}
+
 func (c dashboardCellResponse) toPlatform() *platform.Cell {
 	return &c.Cell
 }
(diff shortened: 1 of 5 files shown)


http/dashboard_service.go
```diff
@@ -472,12 +472,33 @@ func (h *DashboardHandler) handleGetDashboard(w http.ResponseWriter, r *http.Req
 		return
 	}
 
+	include := r.URL.Query().Get("include")
```
jademcgough: could one-line this into `showViewProperties := r.URL.Query().Get("include") == "properties"`

http/dashboard_service.go
```diff
@@ -472,12 +498,28 @@ func (h *DashboardHandler) handleGetDashboard(w http.ResponseWriter, r *http.Req
 		return
 	}
 
+	showViewProperties := r.URL.Query().Get("include") == "properties"
```
jsteenb2: can just inline this in the `509` if block 🤷‍♂

http/dashboard_service.go
```diff
@@ -210,6 +210,32 @@ type dashboardCellResponse struct {
 	Links map[string]string `json:"links"`
 }
 
+func (d *dashboardCellResponse) MarshalJSON() ([]byte, error) {
```
jsteenb2: can you assign the properties directly to the dashboard cell response instead? extending the `dashboardCellResponse` directly on line 209? Think you're better off making this type more xplroable than hiding deets in what is likely unneccssary the MarshalJSON method 🤷‍♂

dearyhud: I hear you - the main reason for me doing this was because I didn't know how to get the `dashboardCellResponse` to surface up the links after adding the view properties to each cell.. Ideally we shouldn't have to marshal the `dashboardCellResponse`, but in my tests I keep getting:
```go
               "cells": [
                 {
                   "h": 4,
                   "id": "da7aba5e5d81e550",
                   "name": "the cell name",
                   "properties": {
                     "axes": null,
                     "colors": null,
                     "geom": "",
                     "legend": {
                     },
                     "note": "",
                     "queries": null,
                     "shadeBelow": false,
                     "showNoteWhenEmpty": false,
                     "type": "xy",
                     "xColumn": "",
                     "yColumn": ""
                   },
                   "w": 3,
                   "x": 1,
                   "y": 2
            +      "links": {
            +        "self": "/api/v2/dashboards/020f755c3c082000/cells/da7aba5e5d81e550",
            +        "view": "/api/v2/dashboards/020f755c3c082000/cells/da7aba5e5d81e550/view"
            +      }
                 }
               ],
```

the code I added in the MarshalJSON adds the links, although it's the same as doing... 
```go
type dashboardCellResponse struct {
	platform.Cell
	Properties platform.ViewProperties `json:"properties,omitempty"`
	Links      map[string]string       `json:"links"`
}
```

jsteenb2: @dearyhud lets chat after standup, I'm not sure I follow the issue here 🤔

dearyhud: sure

http/swagger.yml
```diff
@@ -2217,6 +2217,12 @@ paths:
               type: string
             required: true
             description: The ID of the dashboard to update.
+          - in: query
+            name: include
+            required: false
+            schema:
+              type: string
```
ebb-tide: could we make this an enum so that our client generator generates a type the UI can consume?

http/swagger.yml
```diff
@@ -2217,6 +2217,12 @@ paths:
...
+            description: Includes the cell view properties in the response if set to `properties`
       responses:
           '200':
             description: Get a single dashboard
```
ebb-tide: Could you please also update the schema that may be returned? I can work with you on this if you wish.

dearyhud: yeah sure, I could use your 2-cents on adding this to swagger

http/dashboard_service.go
```diff
@@ -207,21 +207,47 @@ func newDashboardResponse(d *platform.Dashboard, labels []*platform.Label) dashb
 
 type dashboardCellResponse struct {
 	platform.Cell
-	Links map[string]string `json:"links"`
+	Properties platform.ViewProperties `json:"-"`
+	Name       string                  `json:"name,omitempty"`
+	Links      map[string]string       `json:"links"`
+}
+
+func (d *dashboardCellResponse) MarshalJSON() ([]byte, error) {
```
jsteenb2: can you axe this MarshalJSON method now?

PR #16061, opened 2019-11-27
fix(ui): updated telegraf description to be more explicit

### LINKED ISSUE #15990: Highlight "What is Telegraf" Box & information
Objective: Make "What is Telegraf?" information more noticeable for new users
Reason: Many new users of Cloud 2 aren't aware of what Telegraf is

Changes: 
Text box a different color & bigger
Larger Text

image

Asalem1:
Closes #15990

### Problem

Due to its position on the page and the lack of context, it was unclear to users what Telegraf was

### Solution

Added a conditionally rendered textbox with larger text to indicate to the user what Telegraf is. Also updated the background color for the existing bucket and telegraf explainer text to make it stand out more

explainer-update

### DIFF
--- a/ui/src/telegrafs/components/Collectors.tsx
+++ b/ui/src/telegrafs/components/Collectors.tsx
@@ -89,10 +89,14 @@ class Collectors extends PureComponent<Props, State> {
     }
   }
 
+  public static defaultProps = {
+    collectors: [],
+  }
+
   public render() {
     const {collectors} = this.props
     const {searchTerm, sortKey, sortDirection, sortType} = this.state
-
+    const hasTelegrafs = collectors && collectors.length > 0
     return (
       <>
         <NoBucketsWarning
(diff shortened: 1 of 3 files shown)


Asalem1:
Here's a screenshot of the updated styling for the empty state:

Screen Shot 2019-12-02 at 07 50 14

ui/src/telegrafs/components/Collectors.tsx
```diff
@@ -120,8 +123,8 @@ class Collectors extends PureComponent<Props, State> {
           <Grid.Row>
             <Grid.Column
               widthXS={Columns.Twelve}
-              widthSM={Columns.Eight}
-              widthMD={Columns.Ten}
+              widthSM={collectors.length ? Columns.Eight : Columns.Twelve}
```
drdelambre: can you guard here or in the StateProps for collectors being null?

Asalem1: @drdelambre like `collectors && collectors.length`? I'll check out if there are other ways of setting a default prop for stateProps to stabilize it on both ends

PR #16068, opened 2019-12-02
fix(ui): guard against null

drdelambre:
Closes influxdata/honeybadger-errors#2
Closes influxdata/honeybadger-errors#6

it said that there was a null being accessed here, so I let it have a null without failing.

### DIFF
--- a/ui/src/shared/components/cells/View.tsx
+++ b/ui/src/shared/components/cells/View.tsx
@@ -32,19 +32,22 @@ class ViewComponent extends Component<Props> {
     const {view, timeRange, manualRefresh, check} = this.props
     const {dashboardID} = this.props.params
 
-    if (view && view.properties && view.properties.type === 'markdown') {
-      return <Markdown text={view.properties.note} />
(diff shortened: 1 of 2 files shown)


drdelambre:
@ebb-tide I think these might represent the "maybe our resources arent loading properly" problem you mentioned earlier. any thoughts?

ui/src/shared/components/cells/View.tsx
```diff
@@ -32,20 +32,19 @@ class ViewComponent extends Component<Props> {
...
-        return (
-          <RefreshingView
-            timeRange={timeRange}
-            check={check}
-            properties={view.properties}
-            manualRefresh={manualRefresh}
-            dashboardID={dashboardID}
-          />
-        )
+    if (view && view.properties && view.properties.type === 'markdown') {
```
ebb-tide: I would've used lodash get here, just think it's easier to read.

ebb-tide:
@drdelambre I'm really not sure in this case...

Issue #15836, opened 2019-11-11
Add stacked area graph to views


desa:
This may be done as a part of https://github.com/influxdata/giraffe/issues/43, but I wanted to track the work for this separately. This feature would be the ability to use a stacked area graph in a dashboard.

PR #16077, opened 2019-12-02
feat(ui): add positioning option to line graphs

### LINKED ISSUE #15808: Date weirdness in dashboard view
Notice how the date strings overlap in the lower left cell
Screenshot 2019-11-08 10 27 27

When toggling between UTC and Local for their timezone, `undefined` shows up in dashboard view
Screenshot 2019-11-08 at 13 09 59

TCL735:
Closes https://github.com/influxdata/influxdb/issues/15836
Closes #15808 
Add dropdown menu for line positioning in graphs to allow for stacked line graphs.


ui/src/timeMachine/actions/index.ts
```diff
@@ -603,6 +604,18 @@ export const setHistogramPosition = (
   payload: {position},
 })
 
+interface SetLinePositionAction {
```
drdelambre: we're slowly trying to phase this style out in favor of:
https://github.com/influxdata/influxdb/blob/master/ui/src/variables/actions/index.ts#L85

hoorayimhelping: @TCL735 if you convert this over, [here's a diff of a similar change i made recently.](https://github.com/influxdata/influxdb/commit/a90899ebf2e7908d154887bb05b20a22d5281945#diff-b0c7bc23d90cf32723c3bb8675444598R18) maybe it'll help you out with some of the issues i had. in particular, the [`ACTION_TYPE AS ACTION_TYPE`](https://github.com/influxdata/influxdb/commit/a90899ebf2e7908d154887bb05b20a22d5281945#diff-b0c7bc23d90cf32723c3bb8675444598R45) changes tripped me up.

http/swagger.yml
```diff
@@ -8006,6 +8007,9 @@ components:
           type: string
         shadeBelow:
           type: boolean
+        position:
+          type: string
+          enum: [overlaid, stacked]
```
hoorayimhelping: i know that that this pattern has already been established, but out of curiosity, is it possible to create a `LinePositioning` type in one place and use that pre-defined type elsewhere? That way, if we need to change the positioning enum by adding another type of positioning, we only need to make a change in one place instead of tracking down all the possible places we have this enum defined. not sure if that's possible though

TCL735:
There is an issue that needs to resolved: user selection of line position does not persist when cell is saved into a dashboard, and then the cell is viewed in the dashboard. Line position selection is empty and graph defaults to "overlaid" even when "stacked" was selected prior to saving.

TCL735:
> Looks great! Way to go. Does giraffe handle instances where `position` is undefined?

Yes, not giving a position (undefined) is handled through TypeScript. We have an interface that says position is optional. At compile time, TypeScript will convert to JavaScript with the necessary properties (with undefined values for optional properties)

ui/src/timeMachine/actions/index.ts
TCL735: @hoorayimhelping thank you for code example, very helpful! I have refactored this into the preferred style.

http/swagger.yml
TCL735: Golang has enums but I don't think this is possible inside a swagger.yml file just yet. There is a proposal for this. See https://github.com/go-swagger/go-swagger/pull/744

dashboard.go
```diff
@@ -624,6 +624,7 @@ type LinePlusSingleStatProperties struct {
 	XColumn           string           `json:"xColumn"`
 	YColumn           string           `json:"yColumn"`
 	ShadeBelow        bool             `json:"shadeBelow"`
+	Position          string           `json:"position"`
```
jsteenb2: this will be missign from `pkger` will remedy it now, stay tuned

jsteenb2: donezo, commit pushed to address it 👍

Issue #16086, opened 2019-12-03
DWP: Preview Data should be in Raw Data format


Asalem1:
DWP: The preview data should return the raw data form of the incoming data that is trimmed.

Users should have the ability to preview all the raw data

desa:
> Users should have the ability to preview all the raw data

this isn't going to be possible for very large data sets

russorat:
@desa we should be able to build a flux query that returns no more than a fixed number of results. If we cannot, we should look to add that capability into flux. Please work with the flux team to determine the correct query to run.

desa:
we can show a subset of the data, but not all of the data

russorat:
yes a subset is ok.

Asalem1:
Closing this since there doesn't seem to be any progress on this feature

Issue #16085, opened 2019-12-03
DWP: Prepopulate predicates based on data explorer selections


Asalem1:
DWP: When deleting data in the data explorer, the selected tag-key value pairs automatically prepopulate the predicates.

*NOTE* If the user selects multiple values for the given key, the user should be prompted a warning that their selections aren't supported by DWP (since we don't support OR queries). 

*NOTE* This should only happen when the user is in the `query builder` and not in the `script editor`

russorat:
For the warning, basically, i want to make sure that if a user opens the modal, and the filters in the modal do not match the filters they have already selected (for any reason), we should display a message letting them know. This should happen most often when the user selects multiple values in the DE but may also happen if the user selects delete after writing a custom query. 

When the user is in the visual builder, and they select an aggregate, this warning should probably also display. For example, if the user selects "last" as an agg function, they should know that the delete will run against all of their data, not just the "last point in the window".

Does this make sense?

Asalem1:
@russorat we aren't going to support custom query conversion will we?

russorat:
no, you don't have to do custom query support. and after thinking about it, if they have a custom query, click delete, it should be pretty obvious that nothing is configured for them, so we probably don't need a warning there.

Asalem1:
Closing this since there doesn't seem to be any progress on this feature

Issue #16084, opened 2019-12-03
DWP: data explorer deletion should prepopulate based on data explorer


Asalem1:
DWP: When deleting data in the data explorer, the selected timerange should be based on the timerange selected in the data explorer

Currently, the timerange is based on the past hour regardless of what is selected in the data explorer. 

*NOTE*: this should default to a blank time range if selecting `delete data by filter` in the buckets tab

Asalem1:
Closing this since there doesn't seem to be any progress on this feature

Issue #16083, opened 2019-12-03
DWP: remove system buckets from the bucketname list


Asalem1:
DWP: remove system buckets from the delete with predicate bucket list until the API allow for deletion of the system bucket data

Currently, the delete with predicate dropdown allows for users to delete data from default system buckets. Even though this should be available to users, the API is not currently setup to deal with this

desa:
For this issue I'd prefer to see the API for this fixed. System buckets should be treated just like any other bucket.

russorat:
@jademcgough is there an issue for adding support for DWP to system buckets?

russorat:
this issue is stale and we will need to rethink DWP.

PR #16088, opened 2019-12-03
feat(ui): delete with predicate preview

### LINKED ISSUE #15835: preview of data to be deleted with predicate
Before the user confirms a delete we should display the following

* How many points they are going to delete
* Small sample of the data that will be deleted 

This will likely be done using a flux query.

Asalem1:
Closes #15835 

### Problem

Delete with Predicate should return some feedback while the user is selecting specific to delete specific data

### Solution

Implemented a preview based on predicates so that sample data and the number of values to be deleted will be returned to the user to allow them to get a snapshot of what they are deleting

### Going Forward

As mentioned in #15835 this ticket has been expanded upon and no longer meets the end goal it set out to achieve. The ticket has been broken down into several parts that can be worked upon concurrently and succinctly without growing the scope of the feature too much. This work is being pushed up as a means to integrate some of the preview logic into the new larger goal issues. The issues are:

#16080 
#16081 
#16082 
#16083 
#16084 
#16085 
#16086 

### DIFF
--- a/ui/src/shared/actions/predicates.ts
+++ b/ui/src/shared/actions/predicates.ts
@@ -2,9 +2,13 @@
 import {Dispatch} from 'redux-thunk'
 import {extractBoxedCol} from 'src/timeMachine/apis/queryBuilder'
 
-// API
+// Utils
 import {postDelete} from 'src/client'
 import {runQuery} from 'src/shared/apis/query'
+import {getWindowVars} from 'src/variables/utils/getWindowVars'
+import {buildVarsOption} from 'src/variables/utils/buildVarsOption'
+import {getVariableAssignments} from 'src/timeMachine/selectors'
+import {checkQueryResult} from 'src/shared/utils/checkQueryResult'
 
 // Actions
 import {notify} from 'src/shared/actions/notifications'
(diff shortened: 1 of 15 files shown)


ui/src/shared/actions/predicates.ts
```diff
@@ -153,6 +182,47 @@ export const deleteWithPredicate = params => async (
...
+    if (result.didTruncate) {
+      dispatch(notify(resultTooLarge(result.bytesRead)))
+    }
+
+    checkQueryResult(result.csv)
+
+    const files = [result.csv]
+    dispatch(setFiles(files))
+  } catch (e) {
+    if (e.name === 'CancellationError') {
```
drdelambre: should still set the preview status to something here

ui/src/timeMachine/actions/index.ts
```diff
@@ -36,7 +36,7 @@ import {HistogramPosition} from '@influxdata/giraffe'
 import {RemoteDataState} from '@influxdata/clockface'
 import {createView} from 'src/shared/utils/view'
 import {setValues} from 'src/variables/actions'
-import {getTimeRangeByDashboardID} from 'src/dashboards/selectors'
+import {getTimeRangeByDashboardID} from 'src/dashboards/selectors/index'
```
drdelambre: why did you have to be so specific here?

Asalem1: better to be explicit so that there's no doubt about what the function does at a glance

ui/src/shared/actions/predicates.ts
Asalem1: Agreed. We should discuss with design about what feedback beyond a notification we can return to the user

PR #16089, opened 2019-12-03
fix(ui): updated getEndTime to account for now() selections as the upper bound of the query

### LINKED ISSUE #15908: Graph time range should be based on query time range not returned data

Asalem1:
Closes #15908 

### Problem

The previous solution failed to account for query boundaries where `now()` was being set as the upper bound in the `lower` property of the timeRange, and `upper` was being set to `null`

### Solution

Updated `getEndTime` to check to see if `lower` includes `now()` and sets returns a moment reference to `now`

### DIFF
--- a/ui/src/timeMachine/selectors/index.test.ts
+++ b/ui/src/timeMachine/selectors/index.test.ts
@@ -104,16 +104,17 @@ describe('TimeMachine.Selectors.Index', () => {
   })
   it(`getEndTime should return ${newYears} when lower is ${date}`, () => {
     const timeRange = {
-      lower: 'now() - 30d',
+      lower: date,
       upper: date,
     }
     expect(getEndTime(timeRange)).toEqual(newYears)
   })
-  it(`getEndTime should return null when upper is null`, () => {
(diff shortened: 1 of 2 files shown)


ui/src/timeMachine/selectors/index.test.ts
```diff
@@ -104,16 +104,17 @@ describe('TimeMachine.Selectors.Index', () => {
   })
   it(`getEndTime should return ${newYears} when lower is ${date}`, () => {
     const timeRange = {
-      lower: 'now() - 30d',
+      lower: date,
```
Asalem1: This was updated since it was an unrealistic input to expect the lower to be set to a timeRange dropdown selection with the upper property being set to a date.

Issue #16090, opened 2019-12-03
Add check to CI to ensure Flux and promqltests version is the same


wolffcm:
There are two modules defined in the `influxdb`:
- query/promql/internal/promqltests
- The main one

When upgrading Flux, we should also be upgrading Flux in `query/promql/internal/promqltests`.  If this doesn't happen, then:
- We may not be running our PromQL transpiler tests against the current version of Flux
- GoLand tries to be "helpful" and will auto-update the Flux version in promqltests

Users of GoLand will see something like this:
```
--- a/query/promql/internal/promqltests/go.mod
+++ b/query/promql/internal/promqltests/go.mod
@@ -7,7 +7,8 @@ require (
        github.com/google/go-cmp v0.3.1
-       github.com/influxdata/flux v0.54.0
+       github.com/influxdata/changelog v1.1.0 // indirect
+       github.com/influxdata/flux v0.55.1
        github.com/influxdata/influxdb v0.0.0-20190925213338-8af36d5aaedd
```

We should add a check to CI to ensure the two versions are the same to prevent this annoyance and ensure that we test against the latest Flux bits.

fchikwekwe:
Thanks for making the issue so detailed.

PR #16091, opened 2019-12-03
fix(query): update version of Flux in promqltests

wolffcm:
GoLand is auto-updating these files, and it was causing annoyances.


wolffcm:
I filed this issue to ensure we don't forget in the future:
https://github.com/influxdata/influxdb/issues/16090

PR #16095, opened 2019-12-03
fix(ui): updated useEffect to account for times when an organization isn't returned

Asalem1:
Closes https://github.com/influxdata/honeybadger-errors/issues/9

### Problem

If the `getOrgs` api fails to return an organization / errors, `orgs[0]` will be undefined. 

### Solution

Added checks to see if `orgs[0]` exist and added a default route to create a new org if no `orgs` have been returned.

### DIFF
--- a/ui/src/shared/containers/SetOrg.tsx
+++ b/ui/src/shared/containers/SetOrg.tsx
@@ -51,7 +51,11 @@ const SetOrg: FunctionComponent<Props> = ({
     }
 
     // else default to first org
-    router.push(`/orgs/${orgs[0].id}`)
+    if (orgs && orgs.length > 0) {
+      router.push(`/orgs/${orgs[0].id}`)
+    }
+    // if no orgs exist are returned, create one
+    router.push(`/orgs/new`)
   }, [orgID])
 
   return (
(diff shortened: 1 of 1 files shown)


ui/src/shared/containers/SetOrg.tsx
```diff
@@ -51,7 +51,11 @@ const SetOrg: FunctionComponent<Props> = ({
     }
 
     // else default to first org
-    router.push(`/orgs/${orgs[0].id}`)
+    if (orgs && orgs.length > 0) {
+      router.push(`/orgs/${orgs[0].id}`)
+    }
+    // if no orgs exist are returned, create one
+    router.push(`/orgs/new`)
```
TCL735: Should this statement be inside an else? If we have orgs, it will push the first org and then also create a /orgs/new

Is that what is intended?

Asalem1: You're totally right. Good catch!

Asalem1:
This PR is no longer relevant since the issue could not be reproduced and should not be able to exist. We'd need a spec from product in order to determine how best to notify the user of the issue

Issue #16081, opened 2019-12-03
DWP: Tag Keys should be limited based on previous selections


Asalem1:
Delete With Predicate: Tag-Key value pairs should behave like in the query builder, where selecting previous keys should limit the next filterable keys.

Currently, the keys are available based on the bucket name that is selected. We should revise our key availability based on the results of the previous key selected

ebb-tide:
my PRhttps://github.com/influxdata/influxdb/pull/16130 breaks tag key and value setting, I did not fix it because it will be refactored here.

hoorayimhelping:
When implementing this, if it's feasible to create a utility or engine function that the UI can use to remain consistent, I think that would be much preferable to similar / copy pasta logic in multiple places

Asalem1:
Closing this since there doesn't seem to be any progress on this feature

PR #16130, opened 2019-12-04
Chore(ui): Use typed timeRanges throughout the app

ebb-tide:
**Problem:**
Time Ranges were not typed- even though they need to be intrepreted depending on whether they are a duration (like now-5min) or a custom time range(time string to time string)

**Solution:** 
Defined and used types everywhere timeRanges were used. 
This solution breaks the Delete with predicate key value selection logic, but I didn't fix it since it is behind a feature flag, and there is already an issue to fix it: https://github.com/influxdata/influxdb/issues/16081


ui/src/types/queries.ts
```diff
@@ -1,10 +1,43 @@
...
+  upper: null
+  label: string
+  type: 'duration'
+}
+
+export interface CustomTimeRange {
+  lower: string
+  upper: string
+  label: typeof CUSTOM_TIME_RANGE_LABEL | string
+  type: 'custom'
```
hoorayimhelping: This is great, huge improvement from what's there, which is all willy nilly.

I've been marinating on this for a couple hours and I want to get your thoughts on a tweak to the types to see if it's a little more expressive.

So there are a couple of concepts I'm seeing here (I'm naming them after their high level functionality, the names may not map to the variable names in this pr):

* `TimeRange` (as a concept, not as a construct in our application): this is two numbers, a lower and an upper. More formally, this is [an ordered pair](https://en.wikipedia.org/wiki/Ordered_pair) or a [2-tuple](https://www.typescriptlang.org/docs/handbook/basic-types.html#tuple).
* `AppTimeRange`: this has a `TimeRange` (a lower and an upper), a `label`, and a `type`. `Duration` and `Custom` have an `is a` relationship with `AppTimeRange` and `has a` relationship with `TimeRange`
  * A more specific type of `AppTimeRange` is a `SelectableDurationTimeRange`, which is defined by a select dropdown, has no `upper`, and has a specific set of values allowed for `lower`.

Does that jive with what you've seen and are expressing? With all that above said, what do you think of this tweak:

```ts
export type SelectableTimeRangeLower =
  | 'now() - 5m'
  | 'now() - 15m'
  | 'now() - 1h'
  | 'now() - 6h'
  | 'now() - 12h'
  | 'now() - 24h'
  | 'now() - 2d'
  | 'now() - 7d'
  | 'now() - 30d'

export type TimeRange = [string, string | null]

// if it's easier to make TimeRange an interface cause of upper and lower existing all over:
export interface TimeRange {
  lower: string
  upper: string | null
}

export type DashboardRangeType = 'custom' | 'duration' | 'selectable-duration'

// this name may or may not suck, please don't be attached to it if it does
export interface DashboardRange {
  label: string
  range: TimeRange
  type: DashboardRangeType
}

export interface SelectableDashboardRange extends DashboardRange {
  duration: string
  format?: string
  range: [SelectableTimeRangeLower, null]
  seconds: number
}
```
Here's what I'm thinking

1. This makes `TimeRange` a little more reusable and composable, and I think expressive - a range is two date strings - a 2-tuple expresses that very well
   1. I saw a place or two in this PR where a function needed a range-like structure - that kind of inspired the idea that this should be its own concept cause it seems applicable
1. This makes use of [Liskov Substitution](https://en.wikipedia.org/wiki/Liskov_substitution_principle) which is really slick when it works (I think this is one of those cases where it works really well). Basically anywhere you need a dashboard range, you type hint on `DashboardRange`, and any `DashboardRange` or `SelectableDashboardRange` (which extends `DashboardRange`) will work. Any more spefic range we make in the future based off `DashboardRange` will work
1. If this needs to be more specific (for example you want to only allow specific values for `label` on `CustomTimeRange`, we could just do something as simple as:
    ```ts
    export interface CustomTimeRange extends DashboardRange {
      range: [string, string]
      label: typeof CUSTOM_TIME_RANGE_LABEL | string
    }
    ```
So what do you think of that? I'm available to chat on a video call if that's too much words!

ui/src/shared/reducers/predicates.test.ts
```diff
@@ -18,51 +13,57 @@ import {
...
 } from 'src/shared/actions/predicates'
+import {pastHourTimeRange} from 'src/shared/constants/timeRanges'
+import {convertTimeRangeToCustom} from 'src/shared/utils/duration'
+
+const filter: Filter = {key: 'mean', equality: '=', value: '100'}
 
-describe('Predicates reducer test', () => {
-  it('should set the isSerious property', () => {
+describe('Predicates reducer', () => {
+  it('Can set the isSerious property', () => {
```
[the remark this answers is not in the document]
ebb-tide: It wasn't that thoughtful honestly!

ui/src/shared/utils/duration.ts
```diff
@@ -1,5 +1,19 @@
...
+import {TimeRange, CustomTimeRange} from 'src/types'
 import {Duration, DurationUnit} from 'src/types/ast'
+import {TIME_RANGE_FORMAT} from 'src/shared/constants/timeRanges'
+
+export const isDurationParseable = (lower: string): boolean => {
+  if (!lower || !lower.includes('now()')) {
+    return false
+  }
+  // remove spaces and remove "now()-"
+  lower.replace(/\s/g, '').replace(/now\(\)-/, '')
```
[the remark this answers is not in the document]
ebb-tide: love it. doing it.

ui/src/shared/utils/duration.ts
```diff
@@ -1,5 +1,19 @@
...
+
+  const r = /([0-9]+)(y|mo|w|d|h|ms|s|m|us|µs|ns)/g
```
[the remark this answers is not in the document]
ebb-tide: yes definitely. felt bad while duplicating

ui/src/types/queries.ts
ebb-tide: I will remove `label`

ui/src/shared/constants/timeRanges.ts
```diff
@@ -1,87 +1,104 @@
...
+export const pastFifteenMinTimeRange: SelectableDurationTimeRange = {
+  seconds: 900,
+  lower: 'now() - 15m',
+  upper: null,
+  label: 'Past 15m',
+  duration: '15m',
+  type: 'selectable-duration',
+}
+
+export const CUSTOM_TIME_RANGE: {label: string; type: 'custom'} = {
```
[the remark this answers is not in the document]
ebb-tide: this will be addressed by removing the label property from timeRanges

ui/src/shared/reducers/predicates.test.ts
```diff
@@ -18,51 +13,57 @@ import {
...
+  it('Can reset the state after a filter DWP has been successfully submitted', () => {
+    const intermediateState = predicatesReducer(
       initialState,
       setFilter(filter, 0)
-    ).filters
-    const result = predicatesReducer(initialState, resetPredicateState())
-    expect(result).toEqual(state)
+    )
+    const result = predicatesReducer(intermediateState, resetPredicateState())
+    expect(result).toEqual(initialState)
```
Asalem1: should we make a frozen copy of the initialState to ensure that the data we're comparing is referencing a direct copy of the initialState, and not the updated in-memory initialState that may have been updated by `setFilter`?

ebb-tide: yeah-- that's a good idea ! thanks :)

ui/src/dataExplorer/components/DeleteDataOverlay.tsx
```diff
@@ -1,54 +1,62 @@
...
 
 type Props = StateProps & WithRouterProps & DispatchProps
 
 const DeleteDataOverlay: FunctionComponent<Props> = ({
   router,
   params: {orgID},
-  selectedBucketName,
-  selectedTimeRange,
+  bucketNameFromDE,
+  timeRangeFromDE,
```
hoorayimhelping: what is `DE`?

ui/src/shared/utils/duration.ts
```diff
@@ -1,11 +1,29 @@
...
+
+  const removedLower = removeSpacesAndNow(lower)
+
+  return !!removedLower.match(durationRegExp)
+}
 
 export const parseDuration = (input: string): Duration[] => {
-  const r = /([0-9]+)(y|mo|w|d|h|ms|s|m|us|µs|ns)/g
   const result = []
+  const durationRegExp = /([0-9]+)(y|mo|w|d|h|ms|s|m|us|µs|ns)/g
```
hoorayimhelping: ✂️ this, and use the one defined on line 10

ui/src/dashboards/actions/ranges.ts
```diff
@@ -117,49 +100,27 @@ export const updateTimeRangeFromQueryParams = (dashboardID: string) => (
     ignoreQueryPrefix: true,
   })
 
-  const timeRangeFromQueries = {
+  const validatedTimeRangeFromQuery = validateAndTypeRange({
     lower: queryParams.lower,
     upper: queryParams.upper,
```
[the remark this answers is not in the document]
ebb-tide: fixed!

ui/src/dashboards/actions/ranges.ts
```diff
@@ -117,49 +100,27 @@ export const updateTimeRangeFromQueryParams = (dashboardID: string) => (
...
-    upper: validatedTimeRange.upper,
-    zoomedLower: validatedZoomedTimeRange.lower,
-    zoomedUpper: validatedZoomedTimeRange.upper,
-  }
-
-  dispatch(updateQueryParams(updatedQueryParams))
+  dispatch(
+    updateQueryParams({
+      lower: validatedTimeRange.lower,
+      upper: validatedTimeRange.upper,
```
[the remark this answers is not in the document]
ebb-tide: I do think so

ui/src/dataExplorer/components/DeleteDataOverlay.tsx
ebb-tide: data explorer- it's just a long name :(

ui/src/shared/utils/duration.ts
ebb-tide: Turns out ` As with exec() (or in combination with it), test() called multiple times on the same global regular expression instance will advance past the previous match.`

PR #16132, opened 2019-12-05
feat(http): add functionality to retrieve/store telegraf config as toml

glinton:
Centralize the available telegraf plugins and expose via an api endpoint - `/api/v2/telegraf/plugins?type=[input|ouptut|processor|aggregator]`
Update the telegraf config api to be plugin-unaware while still maintaining backwards compatibility with stored plugin-aware data. This is for the upcoming telegraf config editor.

- [ ] Rebased/mergeable
- [ ] Tests pass


http/api_handler.go
```diff
@@ -259,6 +259,7 @@ var apiLinks = map[string]interface{}{
 	"tasks":     "/api/v2/tasks",
 	"checks":    "/api/v2/checks",
 	"telegrafs": "/api/v2/telegrafs",
+	"plugins":   "/api/v2/telegraf/plugins",
```
jsteenb2: part of me thinks this is its own resource `/telegraf-plugins`, its a mouth full, but the nested `/telegraf/plugins` is odd in REST, especially when we have the `telegrafs` as the resource. 🤔

http/api_handler.go
```diff
@@ -364,7 +365,7 @@ func (h *APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if strings.HasPrefix(r.URL.Path, "/api/v2/telegrafs") {
+	if strings.HasPrefix(r.URL.Path, "/api/v2/telegraf") {
```
jsteenb2: thinking this needs to have an addiitinal check here since we'll have both routes available now and the order matters

```go
if strings.HasPrefix(r.URL.Path, "/api/v2/telegrafs") {
... stuffs
}

if strings.HasPrefix(r.URL.Path, "/api/v2/telegraf") {
```

telegraf.go
```diff
@@ -60,313 +61,115 @@ type TelegrafConfigFilter struct {
...
-	}
-	tc.Plugins = append(tc.Plugins, TelegrafPlugin{
-		Config: p,
-	})
-	return nil
+	ID          ID     `json:"id,omitempty"`          // ID of this config object.
+	OrgID       ID     `json:"orgID,omitempty"`       // OrgID is the id of the owning organization.
+	Name        string `json:"name,omitempty"`        // Name of this config object.
+	Description string `json:"description,omitempty"` // Decription of this config object.
+	Config      string `json:"config,omitempty"`      // ConfigTOML contains the raw toml config.
```
jsteenb2: excellent!

telegraf.go
```diff
@@ -60,313 +61,115 @@ type TelegrafConfigFilter struct {
...
 	if err := json.Unmarshal(b, tcd); err != nil {
 		return err
 	}
+
 	orgID := tcd.OrgID
 	if !orgID.Valid() {
 		orgID = tcd.OrganizationID
 	}
+
 	*tc = TelegrafConfig{
```
jsteenb2: can you assign these values straight to the TelegrafConfig instead of deref/assing combo? somethign like

```go
tc.ID = tcd.ID
tc.OrgID = orgID
... yada yada
```

telegraf/plugins/plugin.go
```diff
@@ -0,0 +1,1533 @@
...
+	if err != nil {
+		return nil, err
+	}
+	all.Plugins = append(all.Plugins, t.Plugins...)
+
+	return all, nil
+}
+
+// AvailableInputs returns the base list of available input plugins.
+func AvailableInputs() (*TelegrafPlugins, error) {
```
jsteenb2: what's the motivation for using pointer semantics as the return type here? Feels like you're better off using value semantics for the TelegrafPlugins type 🤷‍♂, just curious

http/api_handler.go
glinton: I did it that way to keep from needing an `/api/v2/` route for everything telegraf related; plugins for now, possibly bundles in the future, etc... By keeping telegraf related things under the `/telegraf` resource, we avoid potential future conflicts with features influxdb might add. the current `/telegrafs` can't serve anything generic yet related to telegraf because the `/telegrafs/:id` bit allows anything.
I'm not sure if it's "proper" as it's been a minute since i've designed api's, but i would have done something like `query` for telegraf's bit:
```go
"telegraf": map[string]string{
	"plugins": "/api/v2/telegraf/plugins",
	"configs": "/api/v2/telegraf/configs", # replace current `/api/v2/telegraf`
}
```
***But***, if it's "the restful way", i definitely agree this should be it's own resource, whether `plugins` or `telegraf-plugins`

http/api_handler.go
glinton: Agreed, i think the plugins logic should use it's own handler too.

telegraf.go
glinton: ah ya.. what about `tc = &TelegrafConfig{` instead?

jsteenb2: I'm fairly certain that will create an issue where the call site won't see the changes, b/c hte pointer value itself changed internally to the method 🤔, at least that's what intelliJ always yells at me 🤣

http/telegraf.go
```diff
@@ -234,6 +189,32 @@ func decodeGetTelegrafRequest(ctx context.Context, r *http.Request) (i platform.
...
+func (h *TelegrafHandler) handleGetTelegrafPlugins(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	telPlugins, err := decodeGetTelegrafPluginRequest(ctx, r)
+	if err != nil {
+		h.HandleHTTPError(ctx, err, w)
+		return
+	}
+
+	h.Logger.Debug("telegraf plugins retrieved", zap.String("plugins", fmt.Sprint(telPlugins)))
```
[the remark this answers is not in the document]
glinton: I don't believe so. I was following the pattern from the [`handleGetTelegrafs` function](https://github.com/influxdata/influxdb/pull/16132/files/6126bbb37ba39f25ce93dc2540c69bd0c5858edd#diff-b0107afcbc22e95e7ebef07287888410R231)

glinton:
It looks like the cypress test is failing because the bucket name isn't showing up any more on a telegraf config. where is that set and, moving forward, should it be set there/associated even? It seems like a user can manually edit their config file and add an `influxdb_v2` output for each of their buckets.

image

http/telegraf.go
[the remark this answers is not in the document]
desa: Lets remove it :)

telegraf.go
```diff
@@ -60,318 +63,213 @@ type TelegrafConfigFilter struct {
...
+
+// telegrafConfigDecode is the helper struct for json decoding. legacy.
+type telegrafConfigDecode struct {
+	ID             *ID                     `json:"id,omitempty"`
+	OrganizationID *ID                     `json:"organizationID,omitempty"`
+	OrgID          *ID                     `json:"orgID,omitempty"`
+	Name           string                  `json:"name,omitempty"`
+	Description    string                  `json:"description,omitempty"`
+	Config         string                  `json:"config,omitempty"`
+	Plugins        *[]telegrafPluginDecode `json:"plugins,omitempty"`
```
desa: Why is this a pointer to a slice?

telegraf.go
```diff
@@ -60,318 +63,213 @@ type TelegrafConfigFilter struct {
...
+		tc.Metadata = m
+	}
 
-// TelegrafPlugin is the general wrapper of the telegraf plugin config
-type TelegrafPlugin struct {
-	Comment string         `json:"comment,omitempty"`
-	Config  plugins.Config `json:"config,omitempty"`
-}
+	if tc.Config == "" {
+		return errors.New("no config provided")
```
desa: Might be good to use an `influxdb.Error` here. Typically any error that isn't an influxdb error ends up giving the user a really awful message in it's API response.

telegraf.go
```diff
@@ -60,318 +63,213 @@ type TelegrafConfigFilter struct {
...
-
-	intervalStr, ok := agent["interval"].(string)
-	if !ok {
-		return errors.New("agent interval is not string")
-	}
-
-	interval, err := time.ParseDuration(intervalStr)
-	if err != nil {
-		return err
+		return errors.New("no config to get buckets")
```
desa: ditto previous comment about errors.

pkger/testdata/dashboard_heatmap.yml
```diff
@@ -38,17 +38,17 @@ spec:
             - hex: "#fbb61a"
             - hex: "#f4df53"
           axes:
-            - name: x
-              label: x_label
-              prefix: x_prefix
-              suffix: x_suffix
+            - name: "x"
```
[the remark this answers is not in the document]
glinton: I can't remember what it was, but there were errors being thrown without them.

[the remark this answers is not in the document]
jsteenb2: I can shed some light on this, when `y` is a value, it is taken as a bool in yaml. Even weirder yaml v2 pkg will convert `"y"` to true as well 😬. This uses v3 which doesn't suffer from taht issue, but as a rule of thumb just been adding `"y"` for all y axis so it doesn't pop up as true

telegraf.go
glinton: Likely so `Plugins` will be `nil` if the json payload is patterned after the new structure. I'm open to suggestions

jsteenb2: the zero value for a slice, is `nil`, and will be omitted by the json marshaler in the case it is a zero value of `nil`, the pointer shouldn't be necessary  🤔

cmd/influxd/launcher/pkger_test.go
```diff
@@ -485,7 +485,22 @@ func newPkg(t *testing.T) *pkger.Pkg {
 	return pkg
 }
 
-const pkgYMLStr = `apiVersion: 0.1.0
+const telConf = `[agent]
+  interval = "10s"
```
jsteenb2: this is freaking awesome, so glad to see this get updated with this!

http/telegraf.go
```diff
@@ -135,55 +135,6 @@ type telegrafLinks struct {
...
-
-	tce := new(telegrafConfigEncode)
-	*tce = telegrafConfigEncode{
-		ID:          r.ID,
-		OrgID:       r.OrgID,
-		Name:        r.Name,
-		Description: r.Description,
-		Agent:       r.Agent,
-		Plugins:     make([]telegrafPluginEncode, len(r.Plugins)),
-		Labels:      r.Labels,
```
jsteenb2: 😢 of joy

http/telegraf.go
```diff
@@ -350,9 +301,8 @@ func decodeTelegrafConfigFilter(ctx context.Context, r *http.Request) (*platform
 }
 
 func decodePostTelegrafRequest(r *http.Request) (*platform.TelegrafConfig, error) {
```
jsteenb2: we don't need this indirection anymore 🤷‍♂

pkger/parser.go
```diff
@@ -12,7 +12,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/BurntSushi/toml"
```
jsteenb2: 👏

pkger/testdata/dashboard_heatmap.yml
```diff
@@ -38,17 +38,17 @@ spec:
...
+              prefix: "x_prefix"
+              suffix: "x_suffix"
               domain:
                 - 0
                 - 10
-            - name: y
-              label: y_label
-              prefix: y_prefix
-              suffix: y_suffix
+            - name: "y"
```
jsteenb2: this should be the only change you need out of all these changes 🤔, we're the tests failing here?

glinton: they were, i'll revert it and test again to get the exact error as i can't remember what the error was

jsteenb2: glad to hear the tests failing though 👍

ui/src/telegrafs/components/Collectors.tsx
```diff
@@ -286,7 +257,7 @@ class Collectors extends PureComponent<Props, State> {
 }
 const mstp = ({telegrafs, orgs: {org}, buckets}: AppState): StateProps => {
   return {
-    collectors: telegrafs.list,
+    hasTelegrafs: !!telegrafs.list.length,
```
hoorayimhelping: i think the more javascript idiomatic way to do this is

```ts
hasTelegrafs: telegrafs.list && telegrafs.list.length > 0
```

i like this personally because it lets the reader understand the intent with fewer mental hops:

```ts
telegrafs.list && telegrafs.list.length > 0
```
the telegrafs has a list - > the list has membership

_vs_

```ts
!!telegrafs.list.length
```
not not telegrafs.list.length -> boolean cast and truthy evaluation -> figure out truthiness of `list.length` (`list.length === 0 // false`, `list.length !== 0 // true`) -> list has membership

ui/src/telegrafs/reducers/index.ts
```diff
@@ -30,7 +29,18 @@ export const telegrafsReducer = (
         draftState.status = status
 
         if (list) {
-          draftState.list = list
+          draftState.list = list.map(l => {
```
hoorayimhelping: should this be `l`? i generally dislike single letter variable names on principle, but even in this case, i think `l` is the wrong abbreviation. it's not a `list` or `list item`, it's a telegraf config, right? you're looping over a list of telegraf configs? should it be `t`?

either way, i think naming the current iteration item, even though it's a small anonymous function, goes a long way in documenting the intent of our code

ui/src/telegrafs/components/CollectorList.tsx
```diff
@@ -83,12 +85,78 @@ export default class CollectorList extends PureComponent<Props> {
...
+      </FilterList>
+    )
+  }
+}
+
+const FilteredList = connect<StateProps, DispatchProps, FilteredOwnProps>(
+  mstp,
+  mdtp
+)(FilteredCollectorList)
+
```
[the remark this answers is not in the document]
glinton: i'm not sure if this would solve it, but it seems that `filteredCollectorList` is more what is wanted. what if we just exported it and stop exporting `collectorList`? again, i'm really not sure if that's what you're talking about or if it's the proper way

ui/src/telegrafs/components/Collectors.tsx
glinton: i wonder if we standardize on one way or the other. it seems like the `!!` is the 'future' https://github.com/influxdata/influxdb/pull/16296/files/0a881156f9839487a1b1431d88e86c2899202ca4#diff-583598f3738558469237fb30beb2e1bdR77

thoughts??

ui/src/telegrafs/components/CollectorList.tsx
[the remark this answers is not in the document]
hoorayimhelping: >what if we just exported it and stop exporting `collectorList`

I guess I should have said: our convention around components seems to be one export per file, not one component per file. so having a helper component that isn't exposed and acts like a private member of the exported component (the exported component is in essence, this file's interface) seems reasonable

pkger/testdata/dashboard_heatmap.yml
glinton: Comment on x makes sense. tests pass without the quotes now, but I'll leave them in. i quoted both for uniformity's sake.

ui/src/telegrafs/components/CollectorList.tsx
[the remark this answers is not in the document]
hoorayimhelping: @glinton do you mean to only export `FilteredCollectorList` or do you want to export both components?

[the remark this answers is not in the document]
glinton: ya, i don't think the other component is used anywhere, so to fit convention i un-exported it

PR #16133, opened 2019-12-05
feat(pkger): add support for exporting telegraf

### LINKED ISSUE #16074: add support for telegraf in pkger
Result:
after discussing with @russorat / @glinton / @drdelambre we're going to leave the pkger side of the house flat. The resulting pkger interface lets the user drop the the entire toml file into the pkger service. The following is an example:

<details>
<summary>example yml</summary>

```yml
apiVersion: 0.1.0
kind: Package
meta:
  pkgName:      pkg_name
  pkgVersion:   1
  description:  pack description
spec:
  resources:
    - kind: Telegraf
      name: first_tele_config
      description: desc
      config: |
        # Configuration for telegraf agent
        [agent]
          ## Default data collection interval for all inputs
          interval = "10s"
          ## Rounds collection interval to 'interval'
          ## ie, if interval="10s" then always collect on :00, :10, :20, etc.
          round_interval = true

          ## Telegraf will send metrics to outputs in batches of at most
          ## metric_batch_size metrics.
          ## This controls the size of writes that Telegraf sends to output plugins.
          metric_batch_size = 1000

          ## For failed writes, telegraf will cache metric_buffer_limit metrics for each
          ## output, and will flush this buffer on a successful write. Oldest metrics
          ## are dropped first when this buffer fills.
          ## This buffer only fills when writes fail to output plugin(s).
          metric_buffer_limit = 10000

          ## Collection jitter is used to jitter the collection by a random amount.
          ## Each plugin will sleep for a random time within jitter before collecting.
          ## This can be used to avoid many plugins querying things like sysfs at the
          ## same time, which can have a measurable effect on the system.
          collection_jitter = "0s"

          ## Default flushing interval for all outputs. Maximum flush_interval will be
          ## flush_interval + flush_jitter
          flush_interval = "10s"
          ## Jitter the flush interval by a random amount. This is primarily to avoid
          ## large write spikes for users running a large number of telegraf instances.
          ## ie, a jitter of 5s and interval 10s means flushes will happen every 10-15s
          flush_jitter = "0s"

          ## By default or when set to "0s", precision will be set to the same
          ## timestamp order as the collection interval, with the maximum being 1s.
          ##   ie, when interval = "10s", precision will be "1s"
          ##       when interval = "250ms", precision will be "1ms"
          ## Precision will NOT be used for service inputs. It is up to each individual
          ## service input to set the timestamp at the appropriate precision.
          ## Valid time units are "ns", "us" (or "µs"), "ms", "s".
          precision = ""

          ## Logging configuration:
          ## Run telegraf with debug log messages.
          debug = false
          ## Run telegraf in quiet mode (error log messages only).
          quiet = false
          ## Specify the log file name. The empty string means to log to stderr.
          logfile = ""

          ## Override default hostname, if empty use os.Hostname()
          hostname = ""
          ## If set to true, do no set the "host" tag in the telegraf agent.
          omit_hostname = false
        [[outputs.influxdb_v2]]
          ## The URLs of the InfluxDB cluster nodes.
          ##
          ## Multiple URLs can be specified for a single cluster, only ONE of the
          ## urls will be written to each interval.
          ## urls exp: http://127.0.0.1:9999
          urls = ["http://localhost:9999"]

          ## Token for authentication.
          token = "$INFLUX_TOKEN"

          ## Organization is the name of the organization you wish to write to; must exist.
          organization = "rg"

          ## Destination bucket to write into.
          bucket = "rucket_3"
        [[inputs.cpu]]
          ## Whether to report per-cpu stats or not
          percpu = true
          ## Whether to report total system cpu stats or not
          totalcpu = true
          ## If true, collect raw CPU time metrics.
          collect_cpu_time = false
          ## If true, compute and report the sum of all non-idle CPU states.
          report_active = false
        [[inputs.disk]]
          ## By default stats will be gathered for all mount points.
          ## Set mount_points will restrict the stats to only the specified mount points.
          # mount_points = ["/"]
          ## Ignore mount points by filesystem type.
          ignore_fs = ["tmpfs", "devtmpfs", "devfs", "overlay", "aufs", "squashfs"]
```

</details>

Todo:
- [x] add parser support
- [x] add dry run/apply support
- [x] add export support

jsteenb2:
Closes #16074 

Adds ability to export telegraf configs to an influx package/template

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influx/pkg_test.go
+++ b/cmd/influx/pkg_test.go
@@ -178,6 +178,7 @@ func Test_Pkg(t *testing.T) {
 			bucketIDs    []influxdb.ID
 			dashIDs      []influxdb.ID
 			labelIDs     []influxdb.ID
+			telegrafIDs  []influxdb.ID
 			varIDs       []influxdb.ID
 			expectedMeta pkger.Metadata
 		}{
@@ -189,7 +190,6 @@ func Test_Pkg(t *testing.T) {
 					flags: []flagArg{
 						{name: "name", val: "new name"},
 						{name: "description", val: "new desc"},
-						{name: "version", val: "new version"},
 						{name: "version", val: "new version"},
 					},
 				},
@@ -209,7 +209,6 @@ func Test_Pkg(t *testing.T) {
 						{name: "name", val: "new name"},
 						{name: "description", val: "new desc"},
 						{name: "version", val: "new version"},
-						{name: "version", val: "new version"},
 					},
 				},
 				dashIDs: []influxdb.ID{1, 2},
@@ -227,7 +226,6 @@ func Test_Pkg(t *testing.T) {
 					flags: []flagArg{
 						{name: "name", val: "new name"},
 						{name: "description", val: "new desc"},
-						{name: "version", val: "new version"},
 						{name: "version", val: "new version"},
 					},
 				},
@@ -247,10 +245,27 @@ func Test_Pkg(t *testing.T) {
 						{name: "name", val: "new name"},
 						{name: "description", val: "new desc"},
 						{name: "version", val: "new version"},
-						{name: "version", val: "new version"},
 					},
 				},
 				varIDs: []influxdb.ID{1, 2},
+				expectedMeta: pkger.Metadata{
+					Name:        "new name",
+					Description: "new desc",
+					Version:     "new version",
+				},
+			},
+			{
+				pkgFileArgs: pkgFileArgs{
+					name:     "telegrafs",
+					encoding: pkger.EncodingYAML,
+					filename: "pkg_0.yml",
+					flags: []flagArg{
+						{name: "name", val: "new name"},
+						{name: "description", val: "new desc"},
+						{name: "version", val: "new version"},
+					},
+				},
+				telegrafIDs: []influxdb.ID{1, 2},
 				expectedMeta: pkger.Metadata{
 					Name:        "new name",
 					Description: "new desc",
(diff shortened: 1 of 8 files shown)


cmd/influx/pkg_test.go
```diff
@@ -256,6 +257,25 @@ func Test_Pkg(t *testing.T) {
...
+			{
+				pkgFileArgs: pkgFileArgs{
+					name:     "telegrafs",
+					encoding: pkger.EncodingYAML,
+					filename: "pkg_0.yml",
+					flags: []flagArg{
+						{name: "name", val: "new name"},
+						{name: "description", val: "new desc"},
+						{name: "version", val: "new version"},
+						{name: "version", val: "new version"},
```
dearyhud: did you mean to specify this twice?

jsteenb2: whoops, that's a mistake 😬

http/telegraf.go
```diff
@@ -455,13 +455,58 @@ var _ platform.TelegrafConfigStore = (*TelegrafService)(nil)
...
+		queryPairs = append(queryPairs, queryPair{
+			k: "userID",
+			v: f.UserID.String(),
+		})
+	}
+
+	var resp struct {
+		Configs []*platform.TelegrafConfig `json:"configurations"`
+	}
+	err := s.client.get(telegrafsPath).
```
glinton: Is there a reason this isn't using the `traceClient` from [`client.go`](https://github.com/influxdata/influxdb/blob/14e7daa6e416d7d2fa0d4c7d83678cb0531ec587/http/client.go#L68) like the other services ([bucket](https://github.com/influxdata/influxdb/blob/5f19c6cace36ceeb68666e080bbd049381d2bef2/http/bucket_service.go#L764))

http/telegraf.go
```diff
@@ -509,6 +553,10 @@ func (c *C) delete(urlPath string) *cReq {
 	return c.newClientReq(http.MethodDelete, urlPath, nil)
 }
 
+func (c *C) get(urlPath string) *cReq {
```
glinton: Why create and use a new client type over using `traceClient`?

http/telegraf.go
jsteenb2: it actually uses that piece, the difference here is the client is more about normalizing the http reqs/resp handling. The tokens/content type/yada yada are all baked in with simple means to override the defaults when necessary (maybe even functional option inputs 🤷‍♂). The client provides some helpers for making the common path super repeatable (like not forgetting status checks on resp and so forth).

http/telegraf.go
jsteenb2: its using traceClient under the hood to make the responses still 👍

PR #16134, opened 2019-12-05
test: add e2e tests for query builder

### LINKED ISSUE #16106: Follow Up on honeybadger-errors #12: Write regression tests
The issue surfaced in [HB 12](https://github.com/influxdata/honeybadger-errors/issues/12) should be tested against in both end-to-end tests and as a regression test in redux to specifically call out this bug.

See: https://github.com/influxdata/influxdb/pull/16105

hoorayimhelping:
Closes #16106

What it says on the tin ☝️. Adds two e2e test cases for creating a query in the query builder, one via the Data Explorer and one via the Dashboard.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/cypress/e2e/queryBuilder.test.ts
+++ b/ui/cypress/e2e/queryBuilder.test.ts
@@ -1,0 +1,140 @@
+import {Organization} from '../../src/types'
+
+const generateRandomSixDigitNumber = () => {
+  const digits = []
+  for (let i = 0; i < 6; i++) {
+    digits[i] = Math.floor(Math.random() * Math.floor(10))
+  }
+  return Number(digits.join(''))
+}
+
+describe('The Query Builder', () => {
+  beforeEach(() => {
+    cy.flush()
+
+    cy.signin().then(({body}) => {
+      cy.wrap(body.org).as('org')
+    })
+
+    cy.writeData([
+      `mem,host=thrillbo-swaggins active=${generateRandomSixDigitNumber()}`,
+      `mem,host=thrillbo-swaggins cached=${generateRandomSixDigitNumber()}`,
+
+      `mem,host=thrillbo-swaggins active=${generateRandomSixDigitNumber()}`,
+      `mem,host=thrillbo-swaggins cached=${generateRandomSixDigitNumber()}`,
+    ])
+  })
+
+  describe('from the data explorer screen', () => {
+    it('creates a query, edits it to add another field, then views its results with pride and satisfaction', () => {
+      cy.get('@org').then((org: Organization) => {
+        cy.visit(`orgs/${org.id}/data-explorer`)
+      })
+
+      cy.contains('mem').click('right') // users sometimes click in random spots
+      cy.contains('active').click('bottomLeft')
+
+      cy.getByTestID('empty-graph--no-queries').should('exist')
+
+      cy.contains('Submit').click()
+
+      cy.get('.giraffe-plot').should('exist')
+
+      cy.contains('Save As').click()
+
+      // open the dashboard selector dropdown
+      cy.contains('Choose at least 1 dashboard').click()
+      cy.contains('Create a New Dashboard').trigger('mousedown')
+      cy.getByTestID('overlay--container').click() // close the blast door i mean dropdown
+
+      cy.get('[placeholder="Add dashboard name"]').type('Basic Ole Dashboard')
+      cy.get('[placeholder="Add optional cell name"]').type('Memory Usage')
+      cy.contains('Save as Dashboard Cell').click()
+
+      // wait for the notification since it's highly animated
+      // we close the notification since it contains the name of the dashboard and interfers with cy.contains
+      cy.wait(250)
+      cy.get('.notification-close').click()
+      cy.wait(250)
+
+      // force a click on the hidden dashboard nav item (cypress can't do the hover)
+      // i assure you i spent a nonzero amount of time trying to do this the way a user would
+      cy.getByTestID('nav-menu_dashboard').click({force: true})
+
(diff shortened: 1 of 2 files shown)


ui/cypress/e2e/queryBuilder.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      cy.get('[placeholder="Add optional cell name"]').type('Memory Usage')
+      cy.contains('Save as Dashboard Cell').click()
+
+      // wait for the notification since it's highly animated
+      // we close the notification since it contains the name of the dashboard and interfers with cy.contains
+      cy.wait(250)
+      cy.get('.notification-close').click()
+      cy.wait(250)
+
+      // force a click on the hidden dashboard nav item (cypress can't do the hover)
```
Asalem1: Cypress can mimic triggering a `mouseover` if that makes a difference:

https://docs.cypress.io/api/commands/trigger.html#Syntax

ui/cypress/e2e/queryBuilder.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      cy.contains('Configure').click()
+
+      cy.get('.giraffe-plot').should('exist')
+      cy.getByTestID('cancel-cell-edit--button').click()
+      cy.contains('Basic Ole Dashboard').should('exist')
+    })
+  })
+
+  describe('from the Dashboard screen', () => {
+    beforeEach(() => {
```
drdelambre: why isn't this logic in the beforeEach above? it appears like all of these tests would need to follow a formula of
- flush
- signin
- goto page
- wait for availability
- [ application specific testing ]

ui/cypress/e2e/queryBuilder.test.ts
hoorayimhelping: it's the left side hover menu, tried getting it to work but it's really finnicky, ended up resorting to `force` https://docs.cypress.io/api/commands/hover.html#Workarounds

ui/cypress/e2e/queryBuilder.test.ts
hoorayimhelping: @drdelambre 

>why isn't this logic in the beforeEach above?

because one set of tests is going to `/dashboards` and one set of tests is going to `/data-explorer`. the other `describe` for `data-explorer` doesn't have a `beforeEach` (and this test probably doesn't need a `beforeEach` since there's only one test) and instead just [`visit`s `orgs/${org.id}/data-explorer`](https://github.com/influxdata/influxdb/pull/16134/files#diff-c432c510f4c140978dbe371ff3e9142dR31)

so the high level flow (after i remove the `beforeEach` here):

```
- describe the data-exporer test
  - flush
  - sign in
  - write a couple of metrics
  - visit `orgs/${org.id}/data-explorer`
  - assertions
- describe the dashboard test
  - flush
  - sign in
  - write a couple of metrics
  - visit `orgs/${org.id}/dashboard`
  - assertions
```

that's reasonable, right?

drdelambre: totes! wasn't a critique, genuine curiosity. it feels like a pattern is forming among the tests we've written and i'm looking for data points to break my mental modal.

ui/cypress/e2e/queryBuilder.test.ts
```diff
@@ -0,0 +1,140 @@
...
+
+      // force a click on the hidden dashboard nav item (cypress can't do the hover)
+      // i assure you i spent a nonzero amount of time trying to do this the way a user would
+      cy.getByTestID('nav-menu_dashboard').click({force: true})
+
+      cy.contains('Basic Ole Dashboard').click()
+      cy.getByTestID('cell-context--toggle').click()
+      cy.contains('Configure').click()
+
+      cy.get('.giraffe-plot').should('exist')
```
drdelambre: is this a sign that giraffe needs better testing affordances (like clockface has?)

ui/cypress/e2e/queryBuilder.test.ts
```diff
@@ -0,0 +1,140 @@
...
+      // we close the notification since it contains the name of the dashboard and interfers with cy.contains
+      cy.wait(250)
+      cy.get('.notification-close').click()
+      cy.wait(250)
+
+      // force a click on the hidden dashboard nav item (cypress can't do the hover)
+      // i assure you i spent a nonzero amount of time trying to do this the way a user would
+      cy.getByTestID('nav-menu_dashboard').click({force: true})
+
+      cy.contains('Basic Ole Dashboard').click()
```
drdelambre: this is a good use of contains. think you can pull it out into a `const` though? to ENSURE it's the same across definition and use?

ui/cypress/e2e/queryBuilder.test.ts
hoorayimhelping: @drdelambre possibly - it might also be an expression of the inherent limitations of testing giraffe, which renders into a canvas which is kind of hard to introspect

ui/cypress/e2e/queryBuilder.test.ts
hoorayimhelping: @drdelambre 
i'd prefer to use a string rather than a variable - even though it creates duplication and the there is the chance of making an error, i think it makes the test much more readable, and i often find myself not wanting to put the context of a variable in my head when reading a test, i'd rather just understand exactly and stupidly what's going on by reading it line by line.

if you're super firm on a variable, i can change it though - it's not a crazy strong preference of mine.

drdelambre: Not super firm in this instance, but I'll get more opinionated as the number of instances grows in a test

PR #16149, opened 2019-12-06
fix(pkger): cleanup leftover bits from telegraf and label mapping misses

jsteenb2:
telegraf was not being mapped correctly for applying the label mapppings
and touched up a bunch of other issues that came up along the way.

- [ ] Rebased/mergeable
- [ ] Tests pass


pkger/service_test.go
```diff
@@ -537,48 +532,104 @@ func TestService(t *testing.T) {
...
+						fakeBktSVC.FindBucketByNameFn = func(_ context.Context, id influxdb.ID, s string) (*influxdb.Bucket, error) {
+							// forces the bucket to be created a new
+							return nil, errors.New("an error")
+						}
+						return []ServiceSetterFn{WithBucketSVC(fakeBktSVC)}
+					},
+				),
+			)
+
+			t.Run("successfully creates dashboards with labels",
```
jsteenb2: putting together my checklist for adding resources, noticed I had holes in coverage for label mappings, this remedies that with the resusable test fn, now all resources have coverage (minus labels/packages b/c htey have no label associations)

PR #16151, opened 2019-12-06
feat(pkger): add support for notification endpoints to parser

jsteenb2:
Add support for notification endpoints to the pkger parser.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/pkger/models.go
+++ b/pkger/models.go
@@ -3,32 +3,40 @@ package pkger
 import (
 	"errors"
 	"fmt"
+	"net/url"
 	"reflect"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/notification/endpoint"
 )
 
 // Package kinds.
 const (
-	KindUnknown   Kind = ""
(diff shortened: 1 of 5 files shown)


notification/endpoint/endpoint.go
```diff
@@ -20,13 +20,11 @@ var typeToEndpoint = map[string](func() influxdb.NotificationEndpoint){
...
-type rawJSON struct {
-	Type string `json:"type"`
-}
-
 // UnmarshalJSON will convert the bytes to notification endpoint.
 func UnmarshalJSON(b []byte) (influxdb.NotificationEndpoint, error) {
-	var raw rawJSON
+	var raw struct {
+		Type string `json:"type"`
+	}
```
kelwang: This strict can be helpful to a lot types, can u move it back?

jsteenb2: it could be, but it isn't atm. Feels awkward to me to pollute the global namespace with something that is never used anywhere else 🤷‍♂. If something wants to use it, they are welcome to abstract it then you know. Main reason I axed it.

PR #16158, opened 2019-12-06
feat(query): add trace response headers

gavincabbage:
Adds `Trace-Id` response header to query endpoint.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/query_handler.go
+++ b/http/query_handler.go
@@ -24,6 +24,7 @@ import (
 	"github.com/influxdata/influxdb/http/metric"
 	"github.com/influxdata/influxdb/kit/check"
 	"github.com/influxdata/influxdb/kit/tracing"
+	"github.com/influxdata/influxdb/logger"
 	influxlogger "github.com/influxdata/influxdb/logger"
 	"github.com/influxdata/influxdb/query"
 	"github.com/pkg/errors"
@@ -32,7 +33,8 @@ import (
 )
 
 const (
-	fluxPath = "/api/v2/query"
+	fluxPath      = "/api/v2/query"
+	traceIDHeader = "Trace-Id"
 )
 
 // FluxBackend is all services and associated parameters required to construct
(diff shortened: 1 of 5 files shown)


http/query_handler.go
```diff
@@ -104,6 +109,11 @@ func (h *FluxHandler) handleQuery(w http.ResponseWriter, r *http.Request) {
 	defer span.Finish()
 
 	ctx := r.Context()
+	if id, sampled, found := logger.TraceInfo(ctx); found {
+		println(id, sampled)
```
kelwang: remove the debug print

logger/fields.go
```diff
@@ -94,6 +94,7 @@ func TraceInfo(ctx context.Context) (traceID string, sampled bool, found bool) {
 		if spanContext, ok := span.Context().(jaeger.SpanContext); ok {
 			traceID = spanContext.TraceID().String()
 			sampled = spanContext.IsSampled()
+			println(traceID, sampled)
```
kelwang: if you really want debug, use zap.Debug

http/query_handler.go
gavincabbage: whoops, thanks! there are two instances of that, ill remove them

gavincabbage:
apologies for leaving those debugs in @kelwang, ive removed them both

http/query_handler.go
```diff
@@ -31,7 +34,9 @@ import (
 )
 
 const (
-	fluxPath = "/api/v2/query"
+	fluxPath           = "/api/v2/query"
+	traceIDHeader      = "Trace-Id"
+	traceSampledHeader = "Trace-Sampled"
```
aanthony1243: I'm trying to decide if the "trace-sampled"  poses a security risk.  it seems to expose some information about the inner workings of our system.  The traceID is useful besides that (query_log, gray_log, etc), so maybe it's OK if they check jaeger and nothing's found?

PR #16161, opened 2019-12-07
feat(pkger): speed up application of pkg with a touch of concurrency

jsteenb2:
governs the concurrency with a simple semaphore. Defaults to 5
concurrent reqs, anything greater, puts a lot of pressure
on the system as a whole (especially OSS/bolt store).

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -1,8 +1,13 @@
 package http
 
 import (
+	"context"
+	"encoding/json"
+	"io"
+	"io/ioutil"
 	"net/http"
 	"net/url"
+	"path"
 
 	"github.com/influxdata/influxdb/kit/tracing"
 )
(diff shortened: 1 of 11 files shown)


http/client.go
```diff
@@ -90,3 +95,126 @@ func (c *traceClient) Do(r *http.Request) (*http.Response, error) {
 	tracing.InjectToHTTPRequest(span, r)
 	return c.Client.Do(r)
 }
+
+// C is a basic http client that can make cReqs with out having to juggle
+// the token and so forth. It provides sane defaults for checking response
+// statuses, sets auth token when provided, and sets the content type to
+// application/json for each request. The token, response checker, and
+// content type can be overidden on the cReq as well.
+type C struct {
```
bthesorceror: Just a note that this struct name could be confusing and without context could be misinterpreted to reference CGO

jsteenb2: great point, totally forgot to comeback and touch that up. Time to do the hardest thing in CS, name something 🤔

PR #16163, opened 2019-12-07
fix(pkger): fix flaky test in pkger service tests

jsteenb2:
tests failing from a data race caused in the tests setup. an incrementing
const needs something to serialize it (atomic in this case) to remove that
data race. This touches that up and provides some reuable thread safe
count mechanism within the mock package. Is pretty brittle to tack these
counters on, and then remember to go back in when things are thrown into
concurrent workflows. This lets us take advantage of describing behavior and
it be safe for concurrent access.

- [x] Rebased/mergeable
- [x] Tests pass


jsteenb2:
tests breaking in unrelated UI test, not a blocker for this work

PR #16165, opened 2019-12-09
feat(httpc): add new httpc pkg to promote better HTTP API clients

jsteenb2:
The new package extracts the new http client from the `http` pkg and provides a more generic and robust solution that can be used in many more places.

- [x] Rebased/mergeable
- [ ] Tests pass


jsteenb2:
@kelwang absolutely, still trying to get last PR to match up in teh upstream, CI is getting wonked up by `fenv` atm

jsteenb2:
@kelwang we should be good on upstream now, updates have been merged

PR #16172, opened 2019-12-09
fix(ui): updated table threshold to set background colors for thresholds correctly

### LINKED ISSUE #15904: Table thresholds are not functional
1. Threshold colors are not applied to table

no-threshold-coloring

2. Threshold settings do not persist

threshold-settings-dont-persist

Asalem1:
Closes #15904 

### Problem

Setting table thresholds would not reflect changes to styling in the table ui

### Solution

Added in check to colorization of the table thresholds to set the background colors correctly. Also updated a table method to correctly determine whether a value is a number or not

### DIFF
--- a/ui/src/shared/components/tables/TableCell.tsx
+++ b/ui/src/shared/components/tables/TableCell.tsx
@@ -137,6 +137,11 @@ class TableCell extends PureComponent<Props> {
     return this.isFirstRow && this.isFirstCol
   }
 
+  private get isTimestamp(): boolean {
+    const {data} = this.props
+    return new Date(data).getTime() > 0
+  }
+
   private get isNumerical(): boolean {
     return !isNaN(Number.parseFloat(this.props.data))
   }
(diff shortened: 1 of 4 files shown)


ui/src/shared/components/tables/TableCell.tsx
```diff
@@ -137,8 +137,12 @@ class TableCell extends PureComponent<Props> {
...
   }
 
-  private get isNumerical(): boolean {
-    return !isNaN(Number.parseFloat(this.props.data))
+  private get isTimestamp(): boolean {
+    return this.props.dataType === 'dateTime:RFC3339'
+  }
+
+  private get isNaN(): boolean {
+    return isNaN(Number(this.props.data))
```
TCL735: Passing in `null` has now changed. Previously, using Number.parseFloat(null) would return `NaN` and ultimately the expression is NaN

Number(null) returns 0, and ultimately the expression is **not** NaN

This looks like an improvement, because we _do_ want to handle null values. Just wanted to make sure it was intentional.

Asalem1: @TCL735 it was intentional in terms of my understanding of how the UI should handle UI values, but it was not explicit in terms of what the criteria should have been. @sjwang90 what're your thoughts on the way we should handle `null` values?

PR #16178, opened 2019-12-10
chore(http): refactor org service to use httpc client

jsteenb2:
Bits of work for the endpiont notification endpoint pkger work. This is broken out to make PRs a bit smaller, easier to reason about.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/http/org_service.go
+++ b/http/org_service.go
@@ -1,7 +1,6 @@
 package http
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -9,10 +8,10 @@ import (
 	"path"
 
 	"github.com/influxdata/httprouter"
-	"go.uber.org/zap"
-
 	"github.com/influxdata/influxdb"
 	"github.com/influxdata/influxdb/kit/tracing"
+	"github.com/influxdata/influxdb/pkg/httpc"
+	"go.uber.org/zap"
 )
 
 // OrgBackend is all services and associated parameters required to construct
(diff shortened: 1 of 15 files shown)


http/org_service.go
```diff
@@ -647,60 +645,42 @@ func (s *OrganizationService) FindOrganizations(ctx context.Context, filter infl
...
-	defer resp.Body.Close()
-
-	if err := CheckError(resp); err != nil {
-		return nil, 0, tracing.LogError(span, err)
-	}
-
 	var os orgsResponse
-	if err := json.NewDecoder(resp.Body).Decode(&os); err != nil {
-		return nil, 0, tracing.LogError(span, err)
+	err := s.Client.
```
jademcgough: this pattern is so clean and readable 😍

http/org_service.go
```diff
@@ -647,60 +645,42 @@ func (s *OrganizationService) FindOrganizations(ctx context.Context, filter infl
...
+		params = append(params, [2]string{"org", *filter.Name})
 	}
 	if filter.ID != nil {
 		span.LogKV("org-id", *filter.ID)
+		params = append(params, [2]string{"orgID", filter.ID.String()})
 	}
 	for _, o := range opt {
 		if o.Offset != 0 {
 			span.LogKV("offset", o.Offset)
+			params = append(params, [2]string{"offset", strconv.Itoa(o.Offset)})
```
jademcgough: options decoding logic is all located in `decodeFindOptions` in `http/paging.go`. I'd prefer not to append these shared params in every http service if possible.

jademcgough: Honestly, the fact that this is looking at options here just for tracing feels like it could be DRY'd, but it doesn't seem like this is uniformly done in other services.

jsteenb2: I'll create a similar method that'll return the params 👍

jsteenb2:
failing test here is failure that keeps popping up from tasks/backend/executor. Knonw issue, issue is created and is tagged with the compute team

PR #16187, opened 2019-12-10
feat(cmd): bucket create to accept org name as flag

### LINKED ISSUE #15828: influx bucket create cannot use organization name

dearyhud:
enables a user to specify an organization name when creating a bucket from the influx cli.

Closes #15828

The influx bucket create command doesn't seem to have a -o flag for specifying the organization name. This PR adds a flag `--org` or `-o` for specifying the org name to avoid the overhead of having to lookup an organization id

- [ ] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influx/inspect.go
+++ b/cmd/influx/inspect.go
@@ -63,6 +63,7 @@ in the following ways:
 	inspectReportTSMCommand.Flags().BoolVarP(&inspectReportTSMFlags.detailed, "detailed", "", false, "emit series cardinality segmented by measurements, tag keys and fields. Warning, may take a while.")
 
 	inspectReportTSMCommand.Flags().StringVarP(&inspectReportTSMFlags.orgID, "org-id", "", "", "process only data belonging to organization ID.")
+	inspectReportTSMCommand.Flags().StringVarP(&inspectReportTSMFlags.orgID, "org", "o", "", "process only data belonging to organization ID.")
 	inspectReportTSMCommand.Flags().StringVarP(&inspectReportTSMFlags.bucketID, "bucket-id", "", "", "process only data belonging to bucket ID. Requires org flag to be set.")
 
 	dir, err := fs.InfluxDir()
--- a/cmd/influx/inspect.go
+++ b/cmd/influx/inspect.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"os"
@@ -18,8 +19,8 @@ type InspectReportTSMFlags struct {
 	exact    bool
 	detailed bool
 
-	orgID, bucketID string
-	dataDir         string
+	orgID, org, bucketID string
+	dataDir              string
 }
 
 var inspectReportTSMFlags InspectReportTSMFlags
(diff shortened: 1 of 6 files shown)


jsteenb2:
can we add the tests in this body of work? a promise to come back and add tests in later, almost never comes to fruition. Be good to start delivering on our testing for cli stuff

cmd/influx/inspect.go
```diff
@@ -63,6 +63,7 @@ in the following ways:
 	inspectReportTSMCommand.Flags().BoolVarP(&inspectReportTSMFlags.detailed, "detailed", "", false, "emit series cardinality segmented by measurements, tag keys and fields. Warning, may take a while.")
 
 	inspectReportTSMCommand.Flags().StringVarP(&inspectReportTSMFlags.orgID, "org-id", "", "", "process only data belonging to organization ID.")
+	inspectReportTSMCommand.Flags().StringVarP(&inspectReportTSMFlags.orgID, "org", "o", "", "process only data belonging to organization ID.")
```
dearyhud: just caught this. will patch this up

cmd/influx/pkg.go
```diff
@@ -108,11 +109,32 @@ func (b *cmdPkgBuilder) cmdPkgApply() *cobra.Command {
...
+			return fmt.Errorf("must specify org-id, or org name not both")
+		}
 		color.NoColor = !b.hasColor
+		var influxOrgID *influxdb.ID
 
-		influxOrgID, err := influxdb.IDFromString(b.orgID)
-		if err != nil {
-			return fmt.Errorf("invalid org ID provided: %s", err.Error())
+		if b.orgID != "" {
+			var err error
```
jsteenb2: this isn't going to work, breaks the pattern for the builder...

jsteenb2: any SVC dep needs to be provided through DI not dumped in here globally

needs fixup

jademcgough: Thanks for catching, we'll fix in the morning.

PR #16190, opened 2019-12-10
feat(metrics): add run latency to executor metrics

AlirieGray:
This PR adds a histogram metric `scheduleInterval` to the Executor metrics that records the time lag between when a task is scheduled to be run and when it actually runs. This can be used to determine the "per minute system concurrency" as described in the queryd SLO RFC.

This PR also adds a property `RunAt` to the Run object, which is used to record the time the Run is scheduled to be executed, including the task's Offset. This is added to the Run object because it is possible that the Run object will be changed by the user in the time between the Run being scheduled and being run, which would make our metrics for that Run inaccurate.

The value `ScheduledAt` is changed to `ScheduledFor` to maintain consistency.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influxd/launcher/launcher.go
+++ b/cmd/influxd/launcher/launcher.go
@@ -647,11 +647,11 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 			sch, sm, err := scheduler.NewScheduler(
 				executor,
 				taskbackend.NewSchedulableTaskService(m.kvService),
-				scheduler.WithOnErrorFn(func(ctx context.Context, taskID scheduler.ID, scheduledAt time.Time, err error) {
+				scheduler.WithOnErrorFn(func(ctx context.Context, taskID scheduler.ID, scheduledFor time.Time, err error) {
 					schLogger.Info(
 						"error in scheduler run",
 						zap.String("taskID", platform.ID(taskID).String()),
-						zap.Time("scheduledAt", scheduledAt),
+						zap.Time("scheduledFor", scheduledFor),
 						zap.Error(err))
 				}),
 			)
--- a/task/backend/executor/executor_metrics.go
+++ b/task/backend/executor/executor_metrics.go
@@ -17,6 +17,7 @@ type ExecutorMetrics struct {
 	manualRunsCounter    *prometheus.CounterVec
 	resumeRunsCounter    *prometheus.CounterVec
 	unrecoverableCounter *prometheus.CounterVec
+	runLatency           *prometheus.HistogramVec
 }
 
 type runCollector struct {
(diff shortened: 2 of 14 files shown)


cmd/influxd/launcher/launcher.go
```diff
@@ -647,11 +647,11 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 			sch, sm, err := scheduler.NewScheduler(
 				executor,
 				taskbackend.NewSchedulableTaskService(m.kvService),
-				scheduler.WithOnErrorFn(func(ctx context.Context, taskID scheduler.ID, scheduledAt time.Time, err error) {
+				scheduler.WithOnErrorFn(func(ctx context.Context, taskID scheduler.ID, scheduledFor time.Time, err error) {
 					schLogger.Info(
 						"error in scheduler run",
 						zap.String("taskID", platform.ID(taskID).String()),
-						zap.Time("scheduledAt", scheduledAt),
+						zap.Time("scheduledFor", scheduledFor),
```
stuartcarnie: Really great to see the consistency with `scheduledFor` and `runAt` 🎉

task/backend/scheduler.go
```diff
@@ -41,6 +41,9 @@ type RunCreation struct {
 	// Unix timestamp for when the next run is due.
 	NextDue int64
 
+	// time.Time for when the next run is due (includes offset)
+	RunAt time.Time
+
```
stuartcarnie: This is not necessary, as you'll be removing the `RunCreation` type with your cleanup PR

task/backend/task.go
```diff
@@ -22,10 +22,10 @@ type TaskControlService interface {
 	NextDueRun(ctx context.Context, taskID influxdb.ID) (int64, error)
 
 	// CreateRun creates a run with a schedule for time.
-	// This differes from CreateNextRun in that it should not to use some scheduling system to determin when the run
+	// This differs from CreateNextRun in that it should not to use some scheduling system to determine when the run
```
stuartcarnie: 👏

task/backend/executor/executor_metrics.go
```diff
@@ -83,6 +84,13 @@ func NewExecutorMetrics(te *TaskExecutor) *ExecutorMetrics {
 			Name:      "resume_runs_counter",
 			Help:      "Total number of runs resumed by task ID",
 		}, []string{"taskID"}),
+
+		scheduleInterval: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "schedule_interval",
+			Help:      "The interval between the time the run was scheduled for and the time the task's next run is due at, by task type",
```
stuartcarnie: The name and wording of this metric is a little confusing. I would consider calling the field `runLatency` and the metric `run_latency_seconds`. [Prometheus guidelines](https://prometheus.io/docs/practices/naming/#metric-names) suggest adding the units (plural) to the end of the metric name.

The `Help` could the be something like:

> Records the latency between the time a task was due to run and the time the task started execution, by task type

PR #16191, opened 2019-12-10
feat(pkger): add notification endpoint dry run/apply functionality

jsteenb2:
adds ability to dry run and apply notification endpoints within a pkg

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/pkger/service_test.go
+++ b/pkger/service_test.go
@@ -230,7 +230,7 @@ func TestService(t *testing.T) {
 								Status:      influxdb.TaskStatusActive,
 							},
 							AuthMethod: "none",
-							Method:     "POST",
+							Method:     "GET",
 							URL:        "https://www.example.com/endpoint/noneauth",
 						},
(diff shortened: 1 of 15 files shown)


pkger/service_test.go
```diff
@@ -75,12 +78,12 @@ func TestService(t *testing.T) {
 			})
 
 			t.Run("single bucket new", func(t *testing.T) {
-				testfileRunner(t, "testdata/bucket", func(t *testing.T, pkg *Pkg) {
+				testfileRunner(t, "testdata/bucket.json", func(t *testing.T, pkg *Pkg) {
```
jsteenb2: this makes it only run against the `.json` version of the file, we support both encodings, but complete coverage is only needed at the parser level, the service is a layer removed from that

jsteenb2:
gotest failing for same data race failure in executor tests (unrelated)

PR #16196, opened 2019-12-11
build: using libflux in OSS

### LINKED ISSUE #15768: Create build using Flux with Rust parser support
The Flux project has a new Rust parser that can be enabled using a Go build tag. We need to create a build of InfluxDB that uses the Flux with the Rust parser. This build should be enabled using a build tag. 

DOD

* Can run a single static binary of InfluxDB that uses the Rust Flux parser
* Go CI tests run against the Rust version of Flux parser
* We only need to support x86 platforms as part of this issue. A follow on issue will tackle the cross-compilation to other platforms/architectures.

yzhang1991:
Closes #15768


.circleci/config.yml
```diff
@@ -1,5 +1,12 @@
 version: 2
 
+aliases:
```
jsternberg: Can you change the circleci version to `2.1` and use [commands](https://circleci.com/docs/2.0/configuration-reference/#commands-requires-version-21) instead?

Also, I believe that rust automatically installed an env script that you can source instead. So you can do `source ~/.cargo/env instead I think.

.circleci/config.yml
```diff
@@ -1,5 +1,12 @@
...
+  - &rustup |
+    curl https://sh.rustup.rs -sSf | sh -s -- -y
+    echo 'export PATH="$HOME/.cargo/bin:$PATH"' >> $BASH_ENV
```
jsternberg: Can you change this to `echo 'source $HOME/.cargo/env' >> $BASH_ENV`? I didn't understand what was going on here before.

PR #16201, opened 2019-12-11
fix(ui): adding times in script editor will default start and endtime for timerange queries

### LINKED ISSUE #16123: graph not handling time range correctly
possibly a side effect from https://github.com/influxdata/influxdb/pull/16040

As a user, i expect the graph to display the range i have specified in the query via the "Range" parameter. Currently, the graphs always use `now()` for the right side of the x axis.

image

Asalem1:
Closes #16123

### Problem

Currently, graph start and end times were being set based on selections from the dropdown regardless of whether the script editor input manual time ranges.

### Solution

Before determining the start and end time, check to see whether a custom time has been input in the script editor

### DIFF
--- a/ui/src/shared/components/RefreshingView.tsx
+++ b/ui/src/shared/components/RefreshingView.tsx
@@ -15,7 +15,7 @@ import {getDashboardValuesStatus} from 'src/variables/selectors'
 import {checkResultsLength} from 'src/shared/utils/vis'
 
 // Selectors
-import {getEndTime, getStartTime} from 'src/timeMachine/selectors/index'
+import {getActiveTimeRange} from 'src/timeMachine/selectors/index'
 import {getTimeRangeByDashboardID} from 'src/dashboards/selectors/index'
 
 // Types
@@ -39,8 +39,7 @@ interface OwnProps {
 }
 
 interface StateProps {
-  endTime: number
-  startTime: number
+  ranges: TimeRange | null
   timeZone: TimeZone
   variableAssignments: VariableAssignment[]
   variablesStatus: RemoteDataState
@@ -73,14 +72,7 @@ class RefreshingView extends PureComponent<Props, State> {
   }
 
   public render() {
-    const {
-      check,
-      endTime,
-      properties,
-      manualRefresh,
-      startTime,
-      timeZone,
-    } = this.props
+    const {check, ranges, properties, manualRefresh, timeZone} = this.props
     const {submitToken} = this.state
 
     return (
@@ -111,12 +103,11 @@ class RefreshingView extends PureComponent<Props, State> {
             >
               <ViewSwitcher
                 check={check}
-                endTime={endTime}
                 files={files}
                 giraffeResult={giraffeResult}
                 loading={loading}
                 properties={properties}
-                startTime={startTime}
+                timeRange={ranges}
                 statuses={statuses}
                 timeZone={timeZone}
               />
@@ -175,8 +166,7 @@ const mstp = (state: AppState, ownProps: OwnProps): StateProps => {
   const timeZone = state.app.persisted.timeZone
 
   return {
-    endTime: getEndTime(timeRange),
-    startTime: getStartTime(timeRange),
+    ranges: getActiveTimeRange(timeRange),
     timeZone,
     variableAssignments,
     variablesStatus: valuesStatus,
--- a/ui/src/shared/components/RefreshingView.tsx
+++ b/ui/src/shared/components/RefreshingView.tsx
@@ -10,9 +10,12 @@ import ViewSwitcher from 'src/shared/components/ViewSwitcher'
 // Utils
 import {GlobalAutoRefresher} from 'src/utils/AutoRefresher'
 import {getTimeRangeVars} from 'src/variables/utils/getTimeRangeVars'
-import {getVariableAssignments} from 'src/variables/selectors'
-import {getDashboardValuesStatus} from 'src/variables/selectors'
+import {
+  getVariableAssignments,
+  getDashboardValuesStatus,
+} from 'src/variables/selectors'
 import {checkResultsLength} from 'src/shared/utils/vis'
+import {getActiveTimeRange} from 'src/timeMachine/selectors/index'
 
 // Types
 import {
@@ -27,7 +30,7 @@ import {
 } from 'src/types'
 
 interface OwnProps {
-  timeRange: TimeRange | null
+  timeRange: TimeRange
   manualRefresh: number
   properties: QueryViewProperties
   dashboardID: string
@@ -35,6 +38,7 @@ interface OwnProps {
 }
 
 interface StateProps {
+  ranges: TimeRange | null
   timeZone: TimeZone
   variableAssignments: VariableAssignment[]
   variablesStatus: RemoteDataState
@@ -67,7 +71,7 @@ class RefreshingView extends PureComponent<Props, State> {
   }
 
   public render() {
-    const {check, timeRange, properties, manualRefresh, timeZone} = this.props
+    const {check, ranges, properties, manualRefresh, timeZone} = this.props
     const {submitToken} = this.state
 
     return (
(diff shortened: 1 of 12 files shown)


ui/src/shared/constants/fluxFunctions.ts
```diff
@@ -1448,7 +1448,7 @@ export const FLUX_FUNCTIONS: FluxToolbarFunction[] = [
         name: 'timeout',
         desc: 'Timeout for the GET request. Default is `30s`.',
         type: 'Duration',
-      }
+      },
```
Asalem1: this was added when I ran `prettier:fix`

ui/src/timeMachine/components/Vis.tsx
```diff
@@ -151,6 +152,7 @@ const mstp = (state: AppState): StateProps => {
   const symbolColumns = getSymbolColumnsSelection(state)
 
   const timeZone = state.app.persisted.timeZone
+  const text = get(viewProperties, 'queries[0].text', '')
```
drdelambre: why isn't this being parsed out into an object before being put into the reducer?

ui/src/shared/components/cells/Cell.tsx
```diff
@@ -140,12 +144,15 @@ class CellComponent extends Component<Props, State> {
 
 const mstp = (state: AppState, ownProps: OwnProps): StateProps => {
   const view = getView(state, ownProps.cell.id)
-
+  const timeRange = getActiveTimeRange(
```
ebb-tide: I think we should do this lower down in RefreshingView.

ui/src/timeMachine/components/Vis.tsx
```diff
@@ -132,7 +130,10 @@ const TimeMachineVis: SFC<Props> = ({
 const mstp = (state: AppState): StateProps => {
   const {
     isViewingRawData,
-    view: {properties: viewProperties},
+    view: {
+      properties: viewProperties,
+      properties: {queries},
```
drdelambre: that's so weird looking!

Asalem1: > this: `query.match(/v.\S*/g).map(m => m.slice(2))` what is this jibberish?

ui/src/shared/components/cells/View.tsx
```diff
@@ -15,7 +15,7 @@ import {ErrorHandling} from 'src/shared/decorators/errors'
 interface OwnProps {
   view: View
   check: Partial<Check>
-  timeRange: TimeRange
+  timeRange: TimeRange | null
```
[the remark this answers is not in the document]
ebb-tide: ah you are... but I think you should not be

ui/src/timeMachine/components/Vis.tsx
drdelambre: oh i'm sorry. didn't realize you weren't fluent in genius.

ui/src/shared/utils/useVisDomainSettings.ts
```diff
@@ -31,15 +38,14 @@ export const getValidRange = (
 export const useVisDomainSettings = (
   storedDomain: number[],
   data: NumericColumnData,
-  startTime: number = Infinity,
-  endTime: number = -Infinity
+  timeRange: TimeRange | null = null
```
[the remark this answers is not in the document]
Asalem1: since we are using the `useVisDomainSettings` function to determine the `y` axis and aren't passing the timerange in those cases, we need to set a default

ui/src/shared/utils/useVisDomainSettings.ts
```diff
@@ -16,11 +19,15 @@ import {extent} from 'src/shared/utils/vis'
 */
 export const getValidRange = (
   data: NumericColumnData = [],
-  startTime: number = Infinity,
-  endTime: number = -Infinity
+  timeRange: TimeRange | null = null
 ) => {
   const range = extent((data as number[]) || [])
+  if (!timeRange) {
```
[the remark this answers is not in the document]
Asalem1: wouldn't we want to return the `range` is timeRange is undefined?

ui/src/timeMachine/selectors/index.ts
```diff
@@ -290,3 +290,18 @@ export const getEndTime = (timeRange: TimeRange): number => {
   }
   return moment().valueOf()
 }
+
+export const getActiveTimeRange = (timeRange: TimeRange, queries: array) => {
+  if (!queries) {
+    return timeRange
```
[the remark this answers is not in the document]
Asalem1: Aw, I see what you're saying. That's a good point!

ui/src/shared/utils/useVisDomainSettings.ts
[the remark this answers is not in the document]
ebb-tide: ok

ui/src/shared/components/RefreshingView.tsx
```diff
@@ -142,6 +132,10 @@ class RefreshingView extends PureComponent<Props, State> {
   private get variableAssignments(): VariableAssignment[] {
     const {timeRange, variableAssignments} = this.props
 
+    if (!timeRange) {
```
ebb-tide: Here you are using whether or not timeRange is defined here as a proxy for whether `v.timeRangeStart` exists in the query- but it will not make sense to the next programmer who looks at this code why timeRange and variableAssigments are related. I'm not sure you need to alter this code at all, but if you feel like you do - you should re-derive whether `v.timeRangeStart` exists in the query.

ui/src/shared/components/RefreshingView.tsx
```diff
@@ -168,15 +162,13 @@ const mstp = (state: AppState, ownProps: OwnProps): StateProps => {
...
   const valuesStatus = getDashboardValuesStatus(state, ownProps.dashboardID)
-
+  const {properties} = ownProps
+  const timeRange = getActiveTimeRange(ownProps.timeRange, properties.queries)
   const timeZone = state.app.persisted.timeZone
 
   return {
-    endTime: getEndTime(timeRange),
-    startTime: getStartTime(timeRange),
+    ranges: timeRange,
```
ebb-tide: let's keep calling this timeRange to signal to the next programmer that this is the same old timeRange that they know from the rest of the app.

ui/src/shared/utils/useVisDomainSettings.ts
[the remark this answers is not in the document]
ebb-tide: that might happen for a variety of reasons. we don't know what to do if timeRange is undefined.

ui/src/timeMachine/selectors/index.ts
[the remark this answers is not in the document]
ebb-tide: Could you test it visually with no query before merging please?

ui/src/shared/components/RefreshingView.tsx
Asalem1: but we have `timeRange` in the `OwnProps` :(

ui/src/timeMachine/selectors/index.ts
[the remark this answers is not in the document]
Asalem1: definitely

ui/src/shared/components/RefreshingView.tsx
Asalem1: I hear ya. I agree. It had originally errored out since the prop was initially set to `null`

ui/src/shared/components/RefreshingView.tsx
ebb-tide: why are we passing the same prop to the same component twice? What is the difference between them? Which one is redundant?

ui/src/shared/utils/useVisDomainSettings.ts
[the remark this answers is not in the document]
Asalem1: that's reasonable

ui/src/shared/components/RefreshingView.tsx
Asalem1: The difference between them is that the `timeRange` refers to the dropdown selection while the `ranges` refer to the specific ranges in the query script editor

ui/src/timeMachine/selectors/index.ts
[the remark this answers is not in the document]
Asalem1: No query = cannot submit without some DOM manipulation (removing the `disabled` flag from the console)

ui/src/shared/components/RefreshingView.tsx
ebb-tide: This is not correct.

PR #16204, opened 2019-12-11
fix: add a timestamp to the telegraf label creation to insure uniqueness

### LINKED ISSUE #15730: [UI - Telegrafs]  Create telegraf transaction fails to create special label
__Steps to reproduce:__
From the Load Data > Telegrafs page
1.  Click Create Configuration
2. Choose a bucket and a plugin (Have been using Docker but issue also appears with others)
3. Click continue
4. Configure the plugin
5. Add a name and description
6. Click Create and Verify 

__Expected behavior:__
If all values are correct expect to see a notification that the Configuration was Successfully saved 

__Actual behavior:__
An error notification appears. "Failed to save configurations"
The console log shows 
```
 POST http://localhost:9999/api/v2/labels 422 (Unprocessable Entity)
```
Which is also shown in the network log 
```
Request URL: http://localhost:9999/api/v2/labels
Request Method: POST
Status Code: 422 Unprocessable Entity
Remote Address: [::1]:9999
Referrer Policy: no-referrer-when-downgrade
```
Payload sent looked like

```
{"orgID":"72548c12d87a9727","name":"@influxdata.token","properties":{"color":"#FFFFFF","description":"token for telegraf config: asdf","tokenID":"04b9730246697000"}}
```
Note that when creating labels directly there is no tokenID in the properties. 

Also can click to another area of the UI and then return to Load Data > Telegrafs and the configuration is present. 

__Screencast__

TelegrafLabel01

__Environment info:__

```
Linux 4.4.0-166-generic x86_64
```
Working with nightly build docker image.  From docker log: 
```
ts=2019-11-04T08:41:47.008799Z lvl=info msg="Welcome to InfluxDB" log_id=0IuP4e00000 version=nightly commit=2d6d38179 build_date=2019-11-04T07:21:10Z
```
browser: __Chrome 77__

__Config:__
Started with e2e-testing enabled
```
--e2e-testing=true
```

hoorayimhelping:
Closes #15730

Labels require unique names. The buggy behavior when we create a telegraf config is to create a label for the config and to name every label `@influxdata.name`. This fix appends a timestamp to `@influxdata.name` (that string is being filtered on, so we want keep `@influxdata.name`)

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/cypress/e2e/collectors.test.ts
+++ b/ui/cypress/e2e/collectors.test.ts
@@ -356,6 +356,30 @@ describe('Collectors', () => {
           })
       })
     })
+
+    // fix for https://github.com/influxdata/influxdb/issues/15730
+    it('creates a configuration with a unique label', () => {
+      cy.contains('Create Configuration').click()
+
+      cy.contains('Docker').click()
+
+      cy.contains('Continue').click()
+
+      cy.contains('docker').click()
+
+      cy.get('[name="endpoint"]').type('http://localhost')
+
+      cy.contains('Done').click()
+      cy.get('input[title="Telegraf Configuration Name"]').type('Label 1')
+      cy.get('input[title="Telegraf Configuration Description"]').type(
+        'Description 1'
+      )
+
+      cy.contains('Create and Verify').click()
+
+      cy.contains('Your configurations have been saved')
+    })
+
     describe('Label creation and searching', () => {
       beforeEach(() => {
         const description = 'Config Description'
--- a/ui/src/dataLoaders/actions/dataLoaders.ts
+++ b/ui/src/dataLoaders/actions/dataLoaders.ts
@@ -454,7 +454,7 @@ const createTelegraf = async (dispatch, getState, plugins) => {
 
     const createdLabel = await client.labels.create({
       orgID: org.id,
-      name: '@influxdata.token',
+      name: `@influxdata.token-${new Date().getTime()}`, // fix for https://github.com/influxdata/influxdb/issues/15730
       properties,
     })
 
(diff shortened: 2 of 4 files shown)


ui/src/dataLoaders/actions/dataLoaders.ts
```diff
@@ -454,10 +454,11 @@ const createTelegraf = async (dispatch, getState, plugins) => {
 
     const createdLabel = await client.labels.create({
       orgID: org.id,
-      name: '@influxdata.token',
+      name: `@influxdata.token-${new Date().getTime()}`,
       properties,
     })
 
+    console.log('label', createdLabel)
```
Asalem1: :(

ui/cypress/e2e/collectors.test.ts
```diff
@@ -356,6 +356,30 @@ describe('Collectors', () => {
           })
       })
     })
+
+    // fix for https://github.com/influxdata/influxdb/issues/15730
+    it('creates a cofiguration with a unique label', () => {
```
121watts:
```suggestion
    it('creates a configuration with a unique label', () => {
```

ui/src/dataLoaders/actions/dataLoaders.ts
```diff
@@ -454,7 +454,7 @@ const createTelegraf = async (dispatch, getState, plugins) => {
 
     const createdLabel = await client.labels.create({
       orgID: org.id,
-      name: '@influxdata.token',
+      name: `@influxdata.token-${new Date().getTime()}`,
```
121watts: perhaps the same comment in the cypress test here?

PR #16206, opened 2019-12-11
fix(cmd): adds org svc dep to pkg

### LINKED ISSUE #15828: influx bucket create cannot use organization name

dearyhud:
Closes #15828
#16187 introduced a regression to the pkg cli. This PR ensures that the org SVC is added to the pkg cli in-line with the builder pattern.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influx/main.go
+++ b/cmd/influx/main.go
@@ -242,3 +242,23 @@ func newLocalKVService() (*kv.Service, error) {
 
 	return kv.NewService(zap.NewNop(), store), nil
 }
+
+func getOrgID(orgSVC influxdb.OrganizationService, id string, name string) (influxdb.ID, error) {
+	if id != "" {
+		influxOrgID, err := influxdb.IDFromString(id)
+		if err != nil {
+			return 0, fmt.Errorf("invalid org ID provided: %s", err.Error())
+		}
+		return *influxOrgID, nil
+	} else if name != "" {
+		org, err := orgSVC.FindOrganization(context.Background(), influxdb.OrganizationFilter{
+			Name: &name,
+		})
+		if err != nil {
+			return 0, fmt.Errorf("%v", err)
+		}
+		return org.ID, nil
+	}
+
+	return 0, fmt.Errorf("")
+}
--- a/cmd/influx/pkg.go
+++ b/cmd/influx/pkg.go
@@ -27,16 +27,16 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-type pkgSVCFn func(cliReq httpClientOpts, opts ...pkger.ServiceSetterFn) (pkger.SVC, error)
+type pkgSVCsFn func(cliReq httpClientOpts, opts ...pkger.ServiceSetterFn) (pkger.SVC, influxdb.OrganizationService, error)
 
-func cmdPkg(svcFn pkgSVCFn, opts ...genericCLIOptfn) *cobra.Command {
+func cmdPkg(svcFn pkgSVCsFn, opts ...genericCLIOptfn) *cobra.Command {
 	return newCmdPkgBuilder(svcFn, opts...).cmdPkg()
 }
 
 type cmdPkgBuilder struct {
 	genericCLIOpts
 
-	svcFn pkgSVCFn
+	svcFn pkgSVCsFn
 
 	applyReqLimit   int
 	file            string
@@ -60,7 +60,7 @@ type cmdPkgBuilder struct {
 	}
 }
 
-func newCmdPkgBuilder(svcFn pkgSVCFn, opts ...genericCLIOptfn) *cmdPkgBuilder {
+func newCmdPkgBuilder(svcFn pkgSVCsFn, opts ...genericCLIOptfn) *cmdPkgBuilder {
 	opt := genericCLIOpts{
 		in: os.Stdin,
 		w:  os.Stdout,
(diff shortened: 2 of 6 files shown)


cmd/influx/pkg.go
```diff
@@ -117,31 +117,31 @@ func (b *cmdPkgBuilder) pkgApplyRunEFn() func(*cobra.Command, []string) error {
 		color.NoColor = !b.hasColor
 		var influxOrgID *influxdb.ID
 
+		svc, orgSVC, err := b.svcFn(flags.httpClientOpts(), pkger.WithApplyReqLimit(b.applyReqLimit))
+		if err != nil {
+			return err
+		}
+
 		if b.orgID != "" {
```
jsteenb2: lets abstract thsi out into a helper function

cmd/influx/main.go
```diff
@@ -242,3 +242,23 @@ func newLocalKVService() (*kv.Service, error) {
...
 }
+
+func getOrgID(orgSVC influxdb.OrganizationService, id string, name string) (influxdb.ID, error) {
+	if id != "" {
+		influxOrgID, err := influxdb.IDFromString(id)
+		if err != nil {
+			return 0, fmt.Errorf("invalid org ID provided: %s", err.Error())
+		}
+		return *influxOrgID, nil
+	} else if name != "" {
```
jsteenb2: nit: you don't need the else here, you can just do

```go
if id != "" {
.... yada yada
}
if name != "" {
.. yada yada
}
return yada yada
```

this is the usual approach in most go libs I've seen/authored 🤔

cmd/influx/pkg.go
```diff
@@ -107,47 +107,38 @@ func (b *cmdPkgBuilder) cmdPkgApply() *cobra.Command {
 	return cmd
 }
 
+func (b *cmdPkgBuilder) validOrgFlags() error {
+	if b.orgID == "" && b.org == "" {
```
jsteenb2: nit: move this to the end of the list of methods (before funcs)

cmd/influx/pkg_test.go
```diff
@@ -126,6 +131,24 @@ func Test_Pkg(t *testing.T) {
...
+					encoding: pkger.EncodingYAML,
+					filename: "pkg_0.yml",
+					flags: []flagArg{
+						{name: "name", val: "new name"},
+						{name: "description", val: "new desc"},
+						{name: "version", val: "new version"},
+						{name: "org", val: "influxdata"},
+					},
+				},
+				expectedMeta: pkger.Metadata{
```
jsteenb2: we need a test for each cmd that updated with org stuff. You'll need to add this to export(export all is separate) func

cmd/influx/main.go
dearyhud: oh okay. the former communicates to me that either one should exist or the other. nbd though, i can update it.

cmd/influx/pkg_test.go
dearyhud: not 100% sure I follow ya

jsteenb2: the export cmd and export all cmd were obth updated. but only 1 test was added. There should be an additional test for the export cmd. You added to the export all. BUt the other is missing

dearyhud: yeah hm. does exporting resources by organization id exist?

jsteenb2: the org id is needed for the exporting of resources, any resource, since the `orgID` is necessary for many of the resource lookups. The export all exports everything for an org, which you have covered in the test right now, so we're good there. The tests are only testing that the correct params are being passed/accessed really. Just tests our public API for the CI

dearyhud: I understand, but where I'm getting at is the implementation of `influx pkg export` doesn't really accept an organization id as input. If that holds true, I'm going to remove the flag for now from the export.

jsteenb2: ahhh yah, your'e right man, that got ripped out a while back 🤦‍♂

PR #16208, opened 2019-12-11
feat: route user to no orgs page

121watts:
Closes https://github.com/influxdata/honeybadger-errors/issues/22

### The Problem
Technically, a user does not need to belong to an organization.  However, the UI is organized around an org context.  Most every page is scoped under `orgs/:orgID`.  So, when a user without an org attempted to use the UI, the app would throw.  

### The Solution
If the user does not belong to an org, direct them to a page instructing them on what they need to do to continue.

Screen Shot 2019-12-11 at 2 40 46 PM


ui/src/index.tsx
```diff
@@ -196,6 +197,7 @@ class Root extends PureComponent {
                 <Route component={GetMe}>
                   <Route component={GetOrganizations}>
                     <Route path="/">
+                      <Route path="no-orgs" component={NoOrgsPage} />
```
drdelambre: why give it a route and not just render the component in routetoorg?

PR #16212, opened 2019-12-12
feat(kv): define forward cursor interface

### LINKED ISSUE #5367: process map shard error on influxdb 0.9.6.1
Hi, I have 3 node influxdb cluster and the cluster seem to be healthy.
show server returns all three nodes.

but I see lot of the error's in the log file(snippet below)
I can insert and read the data normally through curl.

this doesn't seem to be a know issue??
```
2016/01/15 01:37:49 http: TLS handshake error from 127.0.0.1:45489: tls: first record does not look like a TLS handshake
2016/01/15 01:38:19 http: TLS handshake error from 127.0.0.1:45513: tls: first record does not look like a TLS handshake
2016/01/15 01:38:27 http: TLS handshake error from 127.0.0.1:45523: tls: first record does not look like a TLS handshake
[cluster] 2016/01/15 01:42:00 process map shard error: write message type: write tcp 172.16.1.3:42681: connection reset by peer
[cluster] 2016/01/15 01:42:00 process map shard error writing response: write message type: write tcp 172.16.1.3:42681: broken pipe
[cluster] 2016/01/15 01:42:22 process map shard error: write message type: write tcp 172.16.1.2:47428: connection reset by peer
[cluster] 2016/01/15 01:42:22 process map shard error writing response: write message type: write tcp 172.16.1.2:47428: broken pipe
[cluster] 2016/01/15 01:46:53 process map shard error: write message type: write tcp 172.16.1.2:47642: connection reset by peer
[cluster] 2016/01/15 01:46:53 process map shard error writing response: write message type: write tcp 172.16.1.2:47642: broken pipe
[cluster] 2016/01/15 01:46:53 process map shard error: write message type: write tcp 172.16.1.2:47643: connection reset by peer
[cluster] 2016/01/15 01:46:53 process map shard error writing response: write message type: write tcp 172.16.1.2:47643: broken pipe
[cluster] 2016/01/15 01:47:36 process map shard error: write message type: write tcp 172.16.1.2:47706: connection reset by peer
[cluster] 2016/01/15 01:47:36 process map shard error writing response: write message type: write tcp 172.16.1.2:47706: broken pipe
[cluster] 2016/01/15 01:47:37 process map shard error: write message type: write tcp 172.16.1.3:42973: connection reset by peer
[cluster] 2016/01/15 01:47:37 process map shard error writing response: write message type: write tcp 172.16.1.3:42973:
```

GeorgeMac:
Closes #5367

This PR introduces the concept for a `ForwardCursor` as described in the issue above. It's purpose is to create a contract in cursor creators declare upfront the direction and start position in which to create a cursor. This will promote healthier access patterns to the existing cursor. Allowing us to optimize a lot easier in the kv bucket implementations.

1. Define the forward cursor interface. A single direction cursor which is configured with a direction on creation.
```go
c := bucket.ForwardCursor([]byte("runs/"),
        kv.WithCursorDirection(kv.CursorDescending),
        kv.WithCursorHints(kv.WithCursorHintKeyStart("runs/12345")))
```
2. Update all bucket implementations to adhere to new `ForwardCursor()` method.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/bolt/kv.go
+++ b/bolt/kv.go
@@ -13,6 +13,9 @@ import (
 	"go.uber.org/zap"
 )
 
+// check that *KVStore implement kv.Store interface.
+var _ (kv.Store) = (*KVStore)(nil)
+
 // KVStore is a kv.Store backed by boltdb.
 type KVStore struct {
 	path string
@@ -224,10 +227,21 @@ type Cursor struct {
 	key, value []byte
 
 	config kv.CursorConfig
+	closed bool
+}
+
+// Close sets the closed to closed
+func (c *Cursor) Close() error {
+	c.closed = true
+
+	return nil
 }
 
 // Seek seeks for the first key that matches the prefix provided.
 func (c *Cursor) Seek(prefix []byte) ([]byte, []byte) {
+	if c.closed {
+		return nil, nil
+	}
 	k, v := c.cursor.Seek(prefix)
 	if len(k) == 0 && len(v) == 0 {
 		return nil, nil
@@ -237,6 +251,9 @@ func (c *Cursor) Seek(prefix []byte) ([]byte, []byte) {
 
 // First retrieves the first key value pair in the bucket.
 func (c *Cursor) First() ([]byte, []byte) {
+	if c.closed {
+		return nil, nil
+	}
 	k, v := c.cursor.First()
 	if len(k) == 0 && len(v) == 0 {
 		return nil, nil
@@ -246,6 +263,9 @@ func (c *Cursor) First() ([]byte, []byte) {
 
 // Last retrieves the last key value pair in the bucket.
 func (c *Cursor) Last() ([]byte, []byte) {
+	if c.closed {
+		return nil, nil
+	}
 	k, v := c.cursor.Last()
 	if len(k) == 0 && len(v) == 0 {
 		return nil, nil
@@ -255,6 +275,9 @@ func (c *Cursor) Last() ([]byte, []byte) {
 
 // Next retrieves the next key in the bucket.
 func (c *Cursor) Next() (k []byte, v []byte) {
+	if c.closed {
+		return nil, nil
+	}
 	// get and unset previously seeked values if they exist
 	k, v, c.key, c.value = c.key, c.value, nil, nil
 	if len(k) > 0 && len(v) > 0 {
@@ -275,6 +298,9 @@ func (c *Cursor) Next() (k []byte, v []byte) {
 
 // Prev retrieves the previous key in the bucket.
 func (c *Cursor) Prev() (k []byte, v []byte) {
+	if c.closed {
+		return nil, nil
+	}
 	// get and unset previously seeked values if they exist
 	k, v, c.key, c.value = c.key, c.value, nil, nil
 	if len(k) > 0 && len(v) > 0 {
--- a/bolt/kv.go
+++ b/bolt/kv.go
@@ -292,3 +292,8 @@ func (c *Cursor) Prev() (k []byte, v []byte) {
 	}
 	return k, v
 }
+
+// Err always returns nil as nothing can go wrong™ during iteration
+func (c *Cursor) Err() error {
+	return nil
+}
--- a/bolt/kv.go
+++ b/bolt/kv.go
@@ -191,6 +191,17 @@ func (b *Bucket) Delete(key []byte) error {
 	return err
 }
 
+// ForwardCursor retrieves a cursor for iterating through the entries
+// in the key value store in a given direction (ascending / descending).
+func (b *Bucket) ForwardCursor(seek []byte, opts ...kv.CursorOption) (kv.ForwardCursor, error) {
+	cursor := b.bucket.Cursor()
+	cursor.Seek(seek)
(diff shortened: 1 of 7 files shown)


kv/store.go
```diff
@@ -94,19 +94,74 @@ type Bucket interface {
...
 	First() (k []byte, v []byte)
 	// Last moves the cursor to the last key in the bucket.
 	Last() (k []byte, v []byte)
-	// Next moves the cursor to the next key in the bucket.
-	Next() (k []byte, v []byte)
 	// Prev moves the cursor to the prev key in the bucket.
 	Prev() (k []byte, v []byte)
 }
+
+// ForwardCursor is an abstraction for interating/ranging through data in one direction.
```
GeorgeMac: note to self: interating

bolt/kv.go
```diff
@@ -233,19 +254,46 @@ func (c *Cursor) Last() ([]byte, []byte) {
...
+
+	next := c.cursor.Next
+	if c.config.Direction == kv.CursorDescending {
+		next = c.cursor.Prev
+	}
+
+	k, v = next()
 	if len(k) == 0 && len(v) == 0 {
 		return nil, nil
 	}
```
stuartcarnie: I saw this for the original cursor code and wonder why it doesn't just return the key / value, as the caller is required to check `len(key) == 0` to test if it is at the end again.

Not a required change, but perhaps we can clean this up if we remove the old `Cursor` API.

inmem/kv.go
```diff
@@ -232,3 +239,92 @@ func (b *Bucket) getAll(o *kv.CursorHints) ([]kv.Pair, error) {
...
+
+			if fn == nil || fn(j.key, j.value) {
+				batch = append(batch, kv.Pair{Key: j.key, Value: j.value})
+			}
+
+			if len(batch) < cursorBatchSize {
+				return true
+			}
+
+			pairs <- batch
```
stuartcarnie: This goroutine could leak (block forever) if the consumer of this cursor does not call `Next` to read all the data, which happens often when it reaches the last key it is interested in.

One approach that we use for the tsdb Cursor interface is to add a `Close()` API that is responsible for cleaning up any resources.

kv/cursor.go
```diff
@@ -13,9 +13,12 @@ type staticCursor struct {
 }
 
 // Pair is a struct for key value pairs.
+// It also includes an error which should only be non-nil
+// when the Key and Value are nil.
 type Pair struct {
 	Key   []byte
 	Value []byte
+	Err   error
```
stuartcarnie: I missed the use case for an error value on `Pair`. I expected that `ForwardCursor#Err()` would be sufficient to indicate that a read failed and the data should be discarded.

inmem/kv.go
GeorgeMac: good catch!

kv/cursor.go
GeorgeMac: I just thought I would leverage Pair type with an Err field as often I am communicating between a goroutine and a cursor somewhat asynchronously.
I could set the err field on the cursor though directly in the goroutine if this is prefered? Just was concerned about a potential race. But we shouldn't be reading via `Err()` before `Next()` returns nil, at which point the goroutine should have exited. Though I guess this could play into the case where a cursor is exited early?
With the `Close()` semantics would you expect it to be called before `Err()` ?

inmem/kv.go
```diff
@@ -232,3 +239,143 @@ func (b *Bucket) getAll(o *kv.CursorHints) ([]kv.Pair, error) {
...
+		if !ok {
+			break
+		}
+
+		if c.err == nil {
+			for _, pair := range pairs {
+				c.err = pair.err
+			}
+		}
+	}
```
stuartcarnie: See my earlier comment, as you can eliminate this code if we read the `close` channel in the goroutine.

inmem/kv.go
```diff
@@ -232,3 +239,143 @@ func (b *Bucket) getAll(o *kv.CursorHints) ([]kv.Pair, error) {
...
+
+			if fn == nil || fn(j.key, j.value) {
+				batch = append(batch, pair{Pair: kv.Pair{Key: j.key, Value: j.value}})
+			}
+
+			if len(batch) < cursorBatchSize {
+				return true
+			}
+
+			pairs <- batch
```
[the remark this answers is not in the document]
GeorgeMac: @stuartcarnie I actually do this at the top of the function. I could move it down.
I would be worried if I moved it down though that it would mean it tries to fill the batch before acknowledging the stop signal. Since this comes after a length check on the batch as this function is called once per item in the interation.

I guess the best situation would be to do this everywhere. Once on enter incase since the last item we were signalled to stop. Once on send to avoid sending a batch on a close. Once on the final flush.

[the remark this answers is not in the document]
stuartcarnie: @GeorgeMac sorry if I was unclear, I didn't intend for you to remove it from the top. It sounds like you concluded the same thing, which is to probe it in a few places 👍 

This is similar what we see by periodically polling the `context.Done()` channel to determine if the context has been canceled in anyway.

PR #16215, opened 2019-12-12
chore(pkger): use http client/server for integration tests

jsteenb2:
realized as of recent I hadn't set the integration tests to use the server/client. This is a major issue, since I wasn't testing the public API 😳. When changed to use HTTP server, a lot of problems came out of that. This PR remedies the integration tests to use the http server/client and fixes up a bunch of other stuff to make that a reality.

- [x] Rebased/mergeable
- [x] Tests pass
- [x] extend e2e tests to check for dashboard charts (broken atm when going through server, is b/c our dashboard changes only affected the http handler layer, when we remedy the store itself we can axe this)
- [x] fix dashboard to do the looping it has too, unfortunately necessary behind the server side, was never caught b/c e2e tests didn't go through server

### DIFF
--- a/cmd/influx/pkg.go
+++ b/cmd/influx/pkg.go
@@ -27,7 +27,7 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-type pkgSVCsFn func(cliReq httpClientOpts, opts ...pkger.ServiceSetterFn) (pkger.SVC, influxdb.OrganizationService, error)
+type pkgSVCsFn func() (pkger.SVC, influxdb.OrganizationService, error)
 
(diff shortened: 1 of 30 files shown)


cmd/influx/pkg.go
```diff
@@ -138,7 +136,7 @@ func (b *cmdPkgBuilder) pkgApplyRunEFn() func(*cobra.Command, []string) error {
 			return err
 		}
 
-		_, diff, err := svc.DryRun(context.Background(), influxOrgID, pkg)
+		_, diff, err := svc.DryRun(context.Background(), influxOrgID, 0, pkg)
```
jsteenb2: userID set by http client, ignored here, is forced by telegraf/endpoints needing user ID locally in server

PR #16219, opened 2019-12-12
fix(ui): sort by modified now integrated

### LINKED ISSUE #15610: Sort by "Modified" in Dashboards does not sort by modified date
__Steps to reproduce:__
1. Go to Dashboards page
2. Default page is sorted by Name, Click "Modified"
3. Sorting by Modified does not sort by last modified date, continues to sort by name

__Expected behavior:__
When select "Modified", expect dashboards to sort by modified in asc/desc order

__Actual behavior:__
Sorts by Name

dashboards_sortbymodified

Asalem1:
Closes #15610

### Problem

Sorting dashboards by modified dates defaulted to sorting by name

### Solution

Pass in the correct property to sort by so that the orderby function can sort the elements by their appropriate props

### DIFF
--- a/ui/src/shared/utils/useVisDomainSettings.ts
+++ b/ui/src/shared/utils/useVisDomainSettings.ts
@@ -1,6 +1,7 @@
 // Libraries
 import {useMemo} from 'react'
 import {NumericColumnData} from '@influxdata/giraffe'
+import {isNull} from 'lodash'
 
 // Utils
 import {useOneWayState} from 'src/shared/utils/useOneWayState'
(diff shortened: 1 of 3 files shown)


ui/src/shared/utils/useVisDomainSettings.ts
```diff
@@ -22,7 +23,7 @@ export const getValidRange = (
   timeRange: TimeRange | null
 ) => {
   const range = extent((data as number[]) || [])
-  if (!timeRange) {
+  if (isNull(timeRange)) {
```
Asalem1: this never sent up with my last PR so I'm including here as per the PR suggestions

Issue #15814, opened 2019-11-08
Corrupt TSM files not always skipped [backport 1.7]


dgnorton:
Backport fix for https://github.com/influxdata/influxdb/issues/15804 to 1.7

PR #16221, opened 2019-12-13
fix(storage): skip TSM files with block read errors

tmgordeeva:
When we find a bad TSM file during compaction, propagate the error up and move
the bad file aside. The engine will disregard the file so the next compaction
will not hit the same error.

Backport of https://github.com/influxdata/influxdb/pull/15885


tmgordeeva:
@dgnorton does this need a separate ticket too?

dgnorton:
@tmgordeeva here’s the back port issue https://github.com/influxdata/influxdb/issues/15814

PR #16222, opened 2019-12-13
feat(testttp): extend testttp with new JSON methods and refactors test to use it

jsteenb2:
- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/pkg/testttp/http.go
+++ b/pkg/testttp/http.go
@@ -2,6 +2,8 @@ package testttp
 
 import (
 	"bytes"
+	"context"
+	"encoding/json"
 	"io"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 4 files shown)


pkg/testttp/http.go
```diff
@@ -10,76 +12,94 @@ import (
...
-		Req: req,
+		t:   r.t,
+		Req: r.req,
 		Rec: rec,
 	}
 }
 
+// Headers allows the user to set headers on the http request.
+func (r *Req) Headers(k, v string, rest ...string) *Req {
+	headers := append(rest, k, v)
```
kelwang: Why don't you use a map, looks like a little hacky.

jsteenb2: maps makes things unique, don't want to do that here. You can have multiple header values. This design is intentional.

kelwang: The key should unique, right?

jsteenb2: @kelwang sorry for the earlier reply, let that in haste, was rushing out the door. Here is better answer 😬.

When you are using this pkg in a test, the vast majority of hte time (in my experience), has been to add a single header. This design allows for that most common use to be super dead simple:

```go
testttp.
     Post(...).
     Header("Content-Type", "application/json").
      ...  more stuffs
```

if I wnat to add additional headers I still can, but I dont' ahve to sacrifice the ease of using this API to do that. I can simply do the following:

```go
testttp.
     Post(...).
     Header(
         "Content-Type", "application/json",
         "Accept", "text/yaml",
      ).
      ...  more stuffs
```

The signature is a pattern I picked up from Dave Cheney(I'll try and see if I can pull up that blog post he shared about this). It's a means of providing a type safe API with compile time errors when at least 1 arg is required and the rest are optional. Here's an example to illustrate:

```go
func (r *Req) Headers(headerPairs ...string) *Req {...}

// no compile time help here
testtp.Post(...).Headers()... other stuffs 

// k,v estaglish input pattern (meaning they are pairs of k,v)
func (r *Req) Headers(k, v string, rest ...string) *Req {...}

// compile time error here
testtp.Post(...).Headers()... other stuffs 

```

PR #16223, opened 2019-12-13
feat: telegraf editor design work

drdelambre:
@alexpaxton took a pass through the UI for influxdata/idpe#5399 based on feedback from product and with some context of future plans from design. Things got moved around and buttoned up. Also, the color palette isn't so chaotic.


TCL735:
:100: :fire:

PR #16224, opened 2019-12-13
build(e2e): cypress test coverage

121watts:


wbaker85:
We've switched to using UI assets from the `ui` repository instead of using a duplicate copy of the UI source in the `influxdb` repo, see #21090. As such, all of the testing e2e testing setup is handled in the `ui` repo, which has undergone significant development since this PR was opened. So I am closing this PR out!

Issue #16048, opened 2019-11-26
Add environment variables to the influx --help command


russorat:
in the 1.x world, the --help command would show the environment variables that could be set as well so that the command could be shorter. I know about `INFLUX_TOKEN` but are there others for the ORG or BUCKET as well?

It would be helpful to have them here:

```
➜  influxdb git:(master) ✗ ./bin/darwin/influx pkg
Error: required flag(s) "org-id" not set
Usage:
  influx pkg [flags]
  influx pkg [command]

Available Commands:
  export      Export existing resources as a package
  new         Create a reusable pkg to create resources in a declarative manner
  summary     Summarize the provided package
  validate    Validate the provided package

Flags:
  -c, --color               Enable color in output, defaults true (default true)
  -f, --file string         Path to package file
      --force-on-conflict   TTY input, if package will have destructive changes, proceed if set true. (default true)
  -h, --help                Help for the pkg command
  -o, --org-id string       The ID of the organization that owns the bucket
  -q, --quiet               disable output printing
      --table-borders       Enable table borders, defaults true (default true)

Global Flags:
      --host string    HTTP address of Influx (default "http://localhost:9999")
      --local          Run commands locally against the filesystem
      --skip-verify    SkipVerify controls whether a client verifies the server's certificate chain and host name.
  -t, --token string   API token to be used throughout client calls

Use "influx pkg [command] --help" for more information about a command.
```

jademcgough:
A list of all env vars used in the CLI:
INFLUX_ORG_ID
INFLUX_ORG
INFLUX_BUCKET_ID
INFLUX_BUCKET_NAME
INFLUX_TOKEN
INFLUX_HOST
INFLUX_PRECISION

russorat:
this looks like it's been added.

PR #16225, opened 2019-12-13
feat(cmd): apply env vars consistently across cmd

### LINKED ISSUE #16214: consistently apply env vars in CLI
see #16048 for a list of env vars we use.

We use `viper.BindEnv` to access env vars for the CLI tool, but it isn't used in all places where it should be.

Investigate where env vars aren't being used in our CLI and add support in those places.
- the tasks cli commands don't use env vars at all
- the users cli commands also don't use env vars
- possibly other places

dearyhud:
Closes #16214 

Consistently applies env vars to cli commands using flags that could be interchanged with env vars (ie. --org-id INFLUX_ORG_ID, --bucket-name INFLUX_BUCKET_NAME, etc). See https://github.com/influxdata/influxdb/issues/16048 for a more detailed list of env vars.

Also fixes a small bug where we were not setting globally the `INFLUX` prefix before any cobra commands were init.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influx/authorization.go
+++ b/cmd/influx/authorization.go
@@ -8,6 +8,7 @@ import (
 	"github.com/influxdata/influxdb/cmd/influx/internal"
 	"github.com/influxdata/influxdb/http"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 // AuthorizationCreateFlags are command line args used when creating a authorization
@@ -77,6 +78,10 @@ func authCreateCmd() *cobra.Command {
 
 	cmd.Flags().StringVarP(&authCreateFlags.org, "org", "o", "", "The organization name (required)")
 	cmd.MarkFlagRequired("org")
+	viper.BindEnv("ORG")
+	if h := viper.GetString("ORG"); h != "" {
+		authCreateFlags.org = h
+	}
 
 	cmd.Flags().StringVarP(&authCreateFlags.user, "user", "u", "", "The user name")
 
@@ -297,7 +302,16 @@ func authFindCmd() *cobra.Command {
 	cmd.Flags().StringVarP(&authorizationFindFlags.user, "user", "u", "", "The user")
 	cmd.Flags().StringVarP(&authorizationFindFlags.userID, "user-id", "", "", "The user ID")
 	cmd.Flags().StringVarP(&authorizationFindFlags.org, "org", "o", "", "The org")
+	viper.BindEnv("ORG")
+	if h := viper.GetString("ORG"); h != "" {
+		authorizationFindFlags.org = h
+	}
+
 	cmd.Flags().StringVarP(&authorizationFindFlags.orgID, "org-id", "", "", "The org ID")
+	viper.BindEnv("ORG_ID")
+	if h := viper.GetString("ORG_ID"); h != "" {
+		authorizationFindFlags.orgID = h
+	}
 	cmd.Flags().StringVarP(&authorizationFindFlags.id, "id", "i", "", "The authorization ID")
 
 	return cmd
--- a/cmd/influx/bucket.go
+++ b/cmd/influx/bucket.go
@@ -10,6 +10,7 @@ import (
 	"github.com/influxdata/influxdb/cmd/influx/internal"
 	"github.com/influxdata/influxdb/http"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 // Bucket Command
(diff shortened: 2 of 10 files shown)


cmd/influx/main.go
```diff
@@ -83,6 +83,9 @@ func influxCmd() *cobra.Command {
 			cmd.Usage()
 		},
 	}
+
+	viper.SetEnvPrefix("INFLUX")
```
dearyhud: caught this when testing the cmd after these changes - we should've been calling this before adding any of the commands below.

cmd/influx/authorization.go
```diff
@@ -77,6 +78,10 @@ func authCreateCmd() *cobra.Command {
 
 	cmd.Flags().StringVarP(&authCreateFlags.org, "org", "o", "", "The organization name (required)")
 	cmd.MarkFlagRequired("org")
+	viper.BindEnv("ORG")
+	if h := viper.GetString("ORG"); h != "" {
+		authCreateFlags.org = h
```
jsteenb2: this feels off to me, I would suggest flags take precendence over env vars personally

cmd/influx/authorization.go
```diff
@@ -77,6 +78,10 @@ func authCreateCmd() *cobra.Command {
 
 	cmd.Flags().StringVarP(&authCreateFlags.org, "org", "o", "", "The organization name (required)")
 	cmd.MarkFlagRequired("org")
```
jsteenb2: can you make this into a reusable type? somethign like

```go
// rename I don't know what to call it 😬 
type orgStuffs struct {
    orgID string
    organization string
}

func (o *orgStuffs) register(cmd *cobra.Cmd) {

       viper.BindEnv("ORG")
       cmd.Flags().StringVarp(&o.organization,"org", "o", "", "The organization name (required)")
      ... viper stuffs
   
   ... wash rise repeat for org-id in here
}
```

then just make that a dep of the builders or inline it wherever, a step towards reusable patterns for influx cli tool 🤷‍♂

dearyhud: sounds good.

cmd/influx/bucket.go
```diff
@@ -25,9 +26,8 @@ func bucketF(cmd *cobra.Command, args []string) {
 
 // BucketCreateFlags define the Create Command
 type BucketCreateFlags struct {
-	name      string
-	orgID     string
-	org       string
+	name string
+	organization
```
jsteenb2: I'd recommend not embedding the organization type, although, all this will likely get torched when a testable design comes in 🤔

cmd/influx/bucket.go
```diff
@@ -65,9 +64,9 @@ func newBucketService(f Flags) (platform.BucketService, error) {
 }
 
 func bucketCreateF(cmd *cobra.Command, args []string) error {
-	if bucketCreateFlags.orgID == "" && bucketCreateFlags.org == "" {
+	if bucketCreateFlags.organization.id == "" && bucketCreateFlags.organization.name == "" {
 		return fmt.Errorf("must specify org-id, or org name")
-	} else if bucketCreateFlags.orgID != "" && bucketCreateFlags.org != "" {
+	} else if bucketCreateFlags.organization.id != "" && bucketCreateFlags.organization.name != "" {
```
jsteenb2: you can make this a method on the organization type itself and call it here 👍

cmd/influx/main.go
```diff
@@ -230,22 +231,47 @@ func newLocalKVService() (*kv.Service, error) {
...
 }
 
-func getOrgID(orgSVC influxdb.OrganizationService, id string, name string) (influxdb.ID, error) {
-	if id != "" {
-		influxOrgID, err := influxdb.IDFromString(id)
+type organization struct {
+	id, name string
+}
+
+func (org *organization) register(cmd *cobra.Command) {
```
jsteenb2: this looks much improved 👍

cmd/influx/bucket.go
dearyhud: ah - I did https://github.com/influxdata/influxdb/pull/16225/files#diff-0f706fc08b27c10d38fa0be84c6ebf3eR270 just forgot to call it. good catch!

cmd/influx/pkg.go
```diff
@@ -42,9 +42,8 @@ type cmdPkgBuilder struct {
 	hasColor        bool
 	hasTableBorders bool
 	meta            pkger.Metadata
-	orgID           string
-	org             string
-	quiet           bool
+	organization
```
jsteenb2: lets not embed this here, namespace it by a field name, something as simple as the following would be fine.

```go
org organization
```

cmd/influx/bucket.go
dearyhud: totally agree, I wanted this to be easy to revert once we converge on a pattern we're implementing across the board.

cmd/influx/pkg.go
dearyhud: fair enough!

PR #16226, opened 2019-12-13
feat: Add ability to group in Query Builder

### LINKED ISSUE #15984: Query Builder: Add ability to group by columns

hoorayimhelping:
Closes #15984

Adds grouping to query builder

in

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/cypress/e2e/explorer.test.ts
+++ b/ui/cypress/e2e/explorer.test.ts
@@ -349,7 +349,8 @@ describe('DataExplorer', () => {
     })
   })
 
-  describe('raw script editing', () => {
+  // todo: investigate flakiness of this test: https://github.com/influxdata/influxdb/issues/16330
+  describe.skip('raw script editing', () => {
     beforeEach(() => {
       cy.getByTestID('switch-to-script-editor').click()
     })
--- a/ui/src/timeMachine/components/TagSelector.tsx
+++ b/ui/src/timeMachine/components/TagSelector.tsx
@@ -31,17 +31,22 @@ import {
 } from 'src/timeMachine/actions/queryBuilder'
 
 // Utils
-import {toComponentStatus} from 'src/shared/utils/toComponentStatus'
 import DefaultDebouncer from 'src/shared/utils/debouncer'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+import {toComponentStatus} from 'src/shared/utils/toComponentStatus'
 import {
   getActiveQuery,
+  getActiveTagValues,
   getActiveTimeMachine,
   getIsInCheckOverlay,
 } from 'src/timeMachine/selectors'
 
 // Types
-import {AppState, RemoteDataState} from 'src/types'
-import {BuilderAggregateFunctionType} from 'src/client'
+import {
+  AppState,
+  BuilderAggregateFunctionType,
+  RemoteDataState,
+} from 'src/types'
 
 const SEARCH_DEBOUNCE_MS = 500
 
@@ -80,8 +85,6 @@ type Props = StateProps & DispatchProps & OwnProps
 class TagSelector extends PureComponent<Props> {
   private debouncer = new DefaultDebouncer()
 
-  // bucky: this will currently always be 'Filter'
-  // updates to this are imminent
   private renderAggregateFunctionType(
     aggregateFunctionType: BuilderAggregateFunctionType
   ) {
@@ -92,16 +95,29 @@ class TagSelector extends PureComponent<Props> {
   }
 
   public render() {
-    const {aggregateFunctionType, index} = this.props
-
     return (
       <BuilderCard>
-        <BuilderCard.Header
-          title={this.renderAggregateFunctionType(aggregateFunctionType)}
-          onDelete={index !== 0 && this.handleRemoveTagSelector}
-        />
+        {this.header}
         {this.body}
       </BuilderCard>
+    )
+  }
+
+  private get header() {
+    const {aggregateFunctionType, index} = this.props
+
+    return isFlagEnabled('queryBuilderGrouping') ? (
+      <BuilderCard.DropdownHeader
+        options={['filter', 'group']}
+        selectedOption={this.renderAggregateFunctionType(aggregateFunctionType)}
+        onDelete={index !== 0 && this.handleRemoveTagSelector}
+        onSelect={this.handleAggregateFunctionSelect}
+      />
+    ) : (
+      <BuilderCard.Header
+        title={this.renderAggregateFunctionType(aggregateFunctionType)}
+        onDelete={index !== 0 && this.handleRemoveTagSelector}
+      />
     )
   }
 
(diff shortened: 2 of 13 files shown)


ui/src/timeMachine/components/TagSelector.tsx
```diff
@@ -92,19 +91,32 @@ class TagSelector extends PureComponent<Props> {
...
+        {this.header}
         {this.body}
       </BuilderCard>
     )
   }
 
+  private get header() {
+    const {aggregateFunctionType, index} = this.props
+
+    return isFlagEnabled('queryBuilderGrouping') ? (
```
ebb-tide: ❤️

ui/cypress/e2e/explorer.test.ts
```diff
@@ -351,6 +351,10 @@ describe('DataExplorer', () => {
   describe('raw script editing', () => {
     beforeEach(() => {
       cy.getByTestID('switch-to-script-editor').click()
+
+      // explicitly wait, to give the CodeMirror editor a chance to render
+      // may not be necessary when we switch to Monaco
```
121watts: We can explicitly wait on the CodeMirror editor to render by waiting for the `.CodeMirror` class see `'imports the appropriate packages to build a query'` test for an example

ui/src/timeMachine/components/TagSelector.tsx
```diff
@@ -285,13 +305,35 @@ const mstp = (state: AppState, ownProps: OwnProps): StateProps => {
...
+      // if we don't skip the current set of tags, we'll double render them at the bottom of the selector list
+      if (i === ownProps.index) {
+        return
+      }
+      tag.values.forEach(value => {
+        values.push(value)
+      })
+    })
+  }
+
```
121watts: This logic looks like it belongs in its own tested selector.

ui/src/timeMachine/utils/queryBuilder.ts
```diff
@@ -8,7 +8,8 @@ import {
   WINDOW_PERIOD,
 } from 'src/variables/constants'
 import {AGG_WINDOW_AUTO} from 'src/timeMachine/constants/queryBuilder'
-import {BuilderTagsType} from '@influxdata/influx'
+import {BuilderTagsType} from 'src/client/'
```
ebb-tide: I feel like we should be getting this from `@influxdata/influx` if we can?

ui/src/timeMachine/components/TagSelector.tsx
```diff
@@ -92,19 +91,32 @@ class TagSelector extends PureComponent<Props> {
...
+        onSelect={this.handleAggregateFunctionSelect}
+      />
+    ) : (
+      <BuilderCard.Header
+        title={this.renderAggregateFunctionType(aggregateFunctionType)}
+        onDelete={index !== 0 && this.handleRemoveTagSelector}
+      />
+    )
+  }
+
```
121watts: This could probably be its own component.

ui/src/timeMachine/reducers/index.ts
```diff
@@ -829,10 +836,13 @@ export const timeMachineReducer = (
...
-        const selectedValues = draftQuery.builderConfig.tags[index].values
+        let selectedValues = []
+        if (draftQuery) {
+          selectedValues = draftQuery.builderConfig.tags[index].values
 
-        draftQuery.builderConfig.tags.splice(index, 1)
-        draftState.queryBuilder.tags.splice(index, 1)
+          draftQuery.builderConfig.tags.splice(index, 1)
+          draftState.queryBuilder.tags.splice(index, 1)
+        }
```
121watts: A reducer test for this added logic would be 👍

ui/src/timeMachine/utils/queryBuilder.ts
```diff
@@ -110,6 +114,30 @@ function buildQueryHelper(
...
+function buildQueryHelperButWithGrouping(
+  builderConfig: BuilderConfig,
+  fn?: BuilderConfig['functions'][0]
+): string {
+  const [bucket] = builderConfig.buckets
+
+  const tags = Array.from(builderConfig.tags)
+
+  // todo: (bucky) - check to see if we can combine filter calls
+  // https://github.com/influxdata/influxdb/issues/16076
```
ebb-tide: left a note about this

ui/src/timeMachine/utils/queryBuilder.ts
ebb-tide: or imported into and then from `src/types` ?

hoorayimhelping: @ebb-tide i think `@influxdata/influx` is the older definition. `src/client` is the one from swagger and it includes the changes i made earlier regarding `BuilderTagsType` in [16001](https://github.com/influxdata/influxdb/pull/16001/files)

ebb-tide: You're right that `@influxdata/influx` is the older definition but with a lot of other things that come from `src/client` we have been importing into a types file and then importing into and then exporting again from `src/types` it's kinda like saying this is a type we are using in the UI- since there are so many in src/client.

ui/src/timeMachine/components/TagSelector.tsx
hoorayimhelping: it could, but I don't think it adds a ton of value. Once this feature flag goes away, the tag selector will have a `BuilderCard.DropdownHeader` header until we change the interaction, and the `BuilderCard.Header` will be applied to other parts of the query  builder (like the aggregate functions window)

ui/src/timeMachine/utils/queryBuilder.ts
hoorayimhelping: @ebb-tide so the change is that this needs to be imported from `src/types` not `src/client`?

ui/src/timeMachine/reducers/index.ts
hoorayimhelping: @121watts this change i made only adds a check that `draftState` exists. I'll add tests for this, but in a different ticket since this is kind of big and needs to get out.

https://github.com/influxdata/influxdb/issues/16294

hoorayimhelping: @121watts this change i made only adds a check that `draftState` exists. I'll add tests for this, but in a different ticket since this is kind of big and needs to get out.

https://github.com/influxdata/influxdb/issues/16294

ui/cypress/e2e/explorer.test.ts
hoorayimhelping: @121watts i could not get this test to work without the wait. it would always fail on ci (but pass locally). to get this out, do you mind if I keep the `wait` in there and then look at this (and some other tests) next week?

clifforddw:
I just installed the beta and I don't see this functionality anything specif I need ot do ?

PR #16227, opened 2019-12-14
chore: refactor notication endpoints into a service with dep injection

jsteenb2:
this is a blocker for anyone who hits the endpoint services internally(pkger work surfaced this). They had to know that they need to also know of hte secret service then do all that put/delete alongside the operation. This makes that unified inside the store tx. 

- [x] Rebased/mergeable
- [x] Tests pass
- [x] update upstream

### DIFF
--- a/http/notification_endpoint_test.go
+++ b/http/notification_endpoint_test.go
@@ -8,20 +8,19 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
+	"path"
 	"testing"
 
-	platform "github.com/influxdata/influxdb"
-	"github.com/influxdata/influxdb/inmem"
-	"github.com/influxdata/influxdb/kv"
-	platformtesting "github.com/influxdata/influxdb/testing"
-
-	"github.com/google/go-cmp/cmp"
 	"github.com/influxdata/httprouter"
 	"github.com/influxdata/influxdb"
 	pcontext "github.com/influxdata/influxdb/context"
+	"github.com/influxdata/influxdb/inmem"
+	"github.com/influxdata/influxdb/kv"
 	"github.com/influxdata/influxdb/mock"
 	"github.com/influxdata/influxdb/notification/endpoint"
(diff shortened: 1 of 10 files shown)


http/notification_endpoint_test.go
```diff
@@ -503,63 +489,43 @@ func TestService_handlePostNotificationEndpoint(t *testing.T) {
...
-			w := httptest.NewRecorder()
 
-			// this isn't testing anything to do with the routing...
-			h.handlePostNotificationEndpoint(w, r)
+			h := NewNotificationEndpointHandler(zaptest.NewLogger(t), notificationEndpointBackend)
 
-			res := w.Result()
-			content := res.Header.Get("Content-Type")
-			body, _ := ioutil.ReadAll(res.Body)
+			testttp.
```
jsteenb2: this makes me happy :-)

kv/notification_endpoint.go
```diff
@@ -212,6 +215,14 @@ func (s *Service) updateNotificationEndpoint(ctx context.Context, tx Tx, id infl
 	edp.SetCreatedAt(current.GetCRUDLog().CreatedAt)
 	edp.SetUpdatedAt(s.TimeGenerator.Now())
 
+	// TODO: delete old secrets?
+	for _, fld := range edp.SecretFields() {
+		if fld.Value != nil {
+			if err := s.putSecret(ctx, tx, edp.GetOrgID(), fld.Key, *fld.Value); err != nil {
```
kelwang: the secret service is through vault via tcp, not kv. So this won't work at all, and will break the cloud

jademcgough: Riight. I completely forgot that's why this is going through http instead of kv.

I still feel like this could be a smell, but I'd have to think about how we could refactor it.

jsteenb2: regardless of where the secrets are, should still be dep injected into the notification service somewhere. The lack of service definition in the codebase is super painful at times like these. The current abstraction leaks like a sieve 😢

PR #16233, opened 2019-12-16
feat(http): expose list of available telegraf plugins

glinton:
- [ ] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/telegraf.go
+++ b/http/telegraf.go
@@ -65,6 +65,9 @@ const (
 	telegrafsIDOwnersIDPath  = "/api/v2/telegrafs/:id/owners/:userID"
 	telegrafsIDLabelsPath    = "/api/v2/telegrafs/:id/labels"
 	telegrafsIDLabelsIDPath  = "/api/v2/telegrafs/:id/labels/:lid"
+
+	prefixTelegrafPlugins = "/api/v2/telegraf"
+	telegrafPluginsPath   = "/api/v2/telegraf/plugins"
 )
 
 // NewTelegrafHandler returns a new instance of TelegrafHandler.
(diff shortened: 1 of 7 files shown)


http/telegraf.go
```diff
@@ -189,6 +194,30 @@ type telegrafResponses struct {
 	TelegrafConfigs []*telegrafResponse `json:"configurations"`
 }
 
+func decodeGetTelegrafPluginRequest(ctx context.Context, r *http.Request) (*plugins.TelegrafPlugins, error) {
+	t := r.URL.Query().Get("type")
```
jsteenb2: nit: anyway we could change this func signature to just pass in the query vals or even the type? Makes the func signature more useful for inspecting code later 🤷‍♂.

You can also axe the context being passed in since its never being used.

telegraf/plugins/plugins.go
```diff
@@ -0,0 +1,1599 @@
...
+
+// TelegrafPlugins defines a Telegraf version's collection of plugins.
+type TelegrafPlugins struct {
+	Version string   `json:"version,omitempty"` // Version of telegraf plugins are for.
+	OS      string   `json:"os,omitempty"`      // OS the plugins apply to.
+	Plugins []Plugin `json:"plugins,omitempty"` // Plugins this version of telegraf supports.
+}
+
+// ListAvailablePlugins lists available plugins based on type.
+func ListAvailablePlugins(t string) (*TelegrafPlugins, error) {
```
jsteenb2: would a enumerable type be better here than the string?

someting like

```go
type PluginType int

const (
   PluginTypeInput PluginType = iota + 1 // let 0 be the indicator return an error and start valid types at 1
   PluginTypeOutput
   PluginTypeProcessor
   PluginTypeAggregator
   PluginTypeBundle
)

func ListAvailablePlugins(t PluginType) (*TelegrafPlugins, error) {
```

telegraf/plugins/plugins.go
```diff
@@ -0,0 +1,1599 @@
...
+}
+
+// AvailableInputs returns the base list of available input plugins.
+func AvailableInputs() (*TelegrafPlugins, error) {
+	t := &TelegrafPlugins{}
+	err := json.Unmarshal([]byte(availableInputs), t)
+	if err != nil {
+		return nil, err
+	}
+	return t, nil
```
jsteenb2: nit: can we do a sort.Slice on these plugins funcs so they are ordered by name or something? to take away the juggling order in the strings below?

PR #16235, opened 2019-12-16
fix(ui): flux sort no longer being overidden by default FE sort

### LINKED ISSUE #15980: Table View should display results as sorted by underlying flux query
As a user
when configuring a Table View cell in a dashboard
I can opt out of Table View's `Default Sort Field`
so that I can define the `sort()` as part of my flux query

Asalem1:
Closes #15980

### Problem

The previous implementation was setting a default FE sort for columns if they existed without considering whether the results were already sorted

### Solution

Removed the default sort assignment so that values are returned based on the flux query sorting. Added tests to ensure that default values are automatically sorted by `_value`

### DIFF
--- a/ui/src/shared/components/tables/TableCell.tsx
+++ b/ui/src/shared/components/tables/TableCell.tsx
@@ -34,6 +34,22 @@ interface Props extends CellRendererProps {
 class TableCell extends PureComponent<Props> {
   public render() {
     const {data, rowIndex, columnIndex, onHover} = this.props
+    if (rowIndex === 0) {
+      return (
+        <div
+          style={this.style}
+          className={this.class}
+          onClick={this.handleClick}
+          data-column-index={columnIndex}
+          data-row-index={rowIndex}
+          data-testID={`${data}-table-header`}
+          onMouseOver={onHover}
+          title={this.contents}
+        >
+          {this.contents}
+        </div>
+      )
+    }
     return (
       <div
         style={this.style}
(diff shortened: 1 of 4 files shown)


ui/src/shared/components/tables/TableCell.tsx
```diff
@@ -33,14 +33,15 @@ interface Props extends CellRendererProps {
...
-    const {rowIndex, columnIndex, onHover} = this.props
+    const {data, rowIndex, columnIndex, onHover} = this.props
     return (
       <div
         style={this.style}
         className={this.class}
         onClick={this.handleClick}
         data-column-index={columnIndex}
         data-row-index={rowIndex}
+        {...(rowIndex === 0 ? {'data-testID': `${data}-table-header`} : {})} // conditionally adds test-ID for tableCell header
```
drdelambre: dang this is ugly

ui/src/shared/components/tables/TableGraph.tsx
```diff
@@ -81,16 +81,9 @@ class TableGraph extends PureComponent<Props, State> {
...
-      return {...sortOptions, field: '_time'}
-    } else if (headerSet.has('_start')) {
-      return {...sortOptions, field: '_start'}
-    } else if (headerSet.has('_stop')) {
-      return {...sortOptions, field: '_stop'}
-    } else {
-      const headers = table.data[0]
-      return {...sortOptions, field: headers[0]}
     }
+    const headers = table.data[0]
```
drdelambre: what does this do?

ui/src/shared/components/tables/TableCell.tsx
Asalem1: Ya but it’s less that conditionally rendering a component with 1 added prop / adding a testID unnecessarily and sullying the DOM

ui/src/shared/components/tables/TableGraph.tsx
Asalem1: this was the initial default statement set for filtering. While the intention of it is not clear to me, the function of it is to set the default sort to `undefined`

drdelambre: weird

PR #16237, opened 2019-12-17
fix(ui): front end sorting for numeric values now being handled correctly

### LINKED ISSUE #16200: Table sort is not working
Sorting a table in the UI is not working.

See attached screenshots:

image
image

The raw data comes back sorted:

image

Notice the data is an integer type, maybe that is part of the issue?

Asalem1:
Closes #16200

### Problem

Sorting tables via the front end sorted resulted in unordered numbers due to the fact that numbers were being ordered as strings

### Solution

Type checked to-be sorted values to ensure that values are numeric before sorting. Completed testing for table sorting to ensure stability

### DIFF
--- a/ui/cypress/e2e/explorer.test.ts
+++ b/ui/cypress/e2e/explorer.test.ts
@@ -599,7 +599,9 @@ describe('DataExplorer', () => {
       })
     })
 
-    describe('visualize tables', () => {
+    // skipping until the sigin flake gets resolve in:
+    // https://github.com/influxdata/influxdb/issues/16253
+    describe.skip('visualize tables', () => {
       const numLines = 360
 
       beforeEach(() => {
--- a/ui/cypress/e2e/explorer.test.ts
+++ b/ui/cypress/e2e/explorer.test.ts
@@ -623,7 +623,7 @@ describe('DataExplorer', () => {
         })
       })
 
-      it('can view table data', () => {
+      it('can view table data & sort values numerically', () => {
         // build the query to return data from beforeEach
         cy.getByTestID(`selector-list m`).click()
         cy.getByTestID('selector-list v').click()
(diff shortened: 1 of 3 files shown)


ui/cypress/e2e/explorer.test.ts
```diff
@@ -654,8 +654,23 @@ describe('DataExplorer', () => {
           })
         cy.getByTestID('_value-table-header').click()
         cy.get('.table-graph-cell__sort-asc').should('exist')
-        // TODO: complete testing when FE sorting functionality is sorted
-        // https://github.com/influxdata/influxdb/issues/16200
+        cy.getByTestID('_value-table-header').click()
+        cy.get('.table-graph-cell__sort-desc').should('exist')
+        cy.getByTestID('_value-table-header')
+          .should('exist')
```
hoorayimhelping: fyi, `cy.get('foo').should('exist')` is almost always superfluous. the `get` operation implicitly asserts existence.

ui/cypress/e2e/explorer.test.ts
```diff
@@ -623,7 +623,7 @@ describe('DataExplorer', () => {
         })
       })
 
-      it('can view table data', () => {
+      it('can view table data & sort values numerically', () => {
```
hoorayimhelping: this test fails when run with `it.only` which means it's coupled to another part of the test. i think this needs to get fixed.

ui/cypress/e2e/explorer.test.ts
```diff
@@ -654,8 +654,23 @@ describe('DataExplorer', () => {
           })
         cy.getByTestID('_value-table-header').click()
         cy.get('.table-graph-cell__sort-asc').should('exist')
-        // TODO: complete testing when FE sorting functionality is sorted
-        // https://github.com/influxdata/influxdb/issues/16200
+        cy.getByTestID('_value-table-header').click()
+        cy.get('.table-graph-cell__sort-desc').should('exist')
```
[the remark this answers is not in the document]
Asalem1: It's a bit of both. The only indication that a specific cell header isn't toggled by the front end sort is the classname associated with it, hence the check for it.

ui/cypress/e2e/explorer.test.ts
Asalem1: Good point, but it is a way of maintaining UI stability in that we want to know whether the UI is manipulated by the click event

ui/cypress/e2e/explorer.test.ts
Asalem1: I'll take a look and try to reproduce it locally and see whether I can make it less flakey

ui/cypress/e2e/explorer.test.ts
hoorayimhelping: @Asalem1 

>but it is a way of maintaining UI stability in that we want to know whether the UI is manipulated by the click event

i'm curious how `should('exist')` does that where `cy.getByTestID('_value-table-header')` doesn't?

ui/cypress/e2e/explorer.test.ts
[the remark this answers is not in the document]
hoorayimhelping: i don't think it's necessary at all - it really feels like an implementation detail: we could change our underlying css so that we render columns sorted ascending with some kind squiggly underline and columns sorted descending with a dashed outline. making those changes would render this test invalid. 

either way, i think it's enough discussion on a test - i think this is a bit implementation heavy, but if you're comfortable with this, go for it

ui/cypress/e2e/explorer.test.ts
hoorayimhelping: until we sort this out, i think you should `skip` this test and link to this issue: https://github.com/influxdata/influxdb/issues/16253

PR #16241, opened 2019-12-17
refactor(ui): Clockface replacement

alexpaxton:
- Replace local `ResponsiveGridSizer` with `SquareGrid` (from Clockface)
- Delete local `ResponsiveGridSizer` and `GridSizer`
- Replace local `RightClick` with equivalent from Clockface
- Delete local `RightClick`
- Replace `SelectGroup` in save as dialog with tabs
- Add test ids to touched components for more precise element selection

Screen Shot 2019-12-17 at 2 16 55 PM

- [x] Rebased/mergeable
- [x] Tests pass


ui/src/timeMachine/components/QueryTab.tsx
```diff
@@ -169,10 +168,9 @@ class TimeMachineQueryTab extends PureComponent<Props, State> {
     return this.props.queryCount > 1
   }
 
-  private handleRemove = (e: MouseEvent): void => {
+  private handleRemove = (): void => {
```
alexpaxton: The Clockface version of `RightClick` does not pass an event object back to `onClick` so I created an issue to track: https://github.com/influxdata/clockface/issues/420

Also, it looks like `stopPropagation` is already being called from within `RightClickMenuItem` so it may not even need to be called here

PR #16242, opened 2019-12-17
feat(endpoints): drop id prefix req endpoint secret key

### LINKED ISSUE #16240: remove id prefix requirement for notification endpoint secret key value
__Proposal:__
Ideally, we just nuke this requirement and let any secret, as long as it exists, be usable.

__Current behavior:__
Requires key to be prefixed with notification endpoint id for all secrets in notification endpoints.

__Desired behavior:__
Nuke the requirement that the id prefix the secret key. Makes the secrets unusable for other entities within the same org.

Steps to produce:

1. create a secret
2. create endpoint and have the secret key referenced in it
3. receive error

jsteenb2:
Closes #16240 

Relaxes the rigid validation on endpoints to not require the id be prefixed.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/notification/endpoint/slack.go
+++ b/notification/endpoint/slack.go
@@ -45,10 +45,10 @@ func (s Slack) Valid() error {
 	if err := s.Base.valid(); err != nil {
 		return err
 	}
-	if s.URL == "" && s.Token.Key == "" {
+	if s.URL == "" {
 		return &influxdb.Error{
 			Code: influxdb.EInvalid,
-			Msg:  "slack endpoint URL and token are empty",
+			Msg:  "slack endpoint URL must be provided",
 		}
 	}
 	if s.URL != "" {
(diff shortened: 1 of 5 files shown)


notification/endpoint/slack.go
```diff
@@ -45,10 +45,10 @@ func (s Slack) Valid() error {
 	if err := s.Base.valid(); err != nil {
 		return err
 	}
-	if s.URL == "" && s.Token.Key == "" {
+	if s.URL == "" {
```
jsteenb2: token is not required for a url, if URL is "" it shoudl be an eror for slack, token shoudn't matter

jademcgough:
Are we using the id prefix anywhere for lookups?

jsteenb2:
@jademcgough we aren't atm. We save teh entire secrets bit into the store, and never actualy use the notification id to look for the secret, the entire key is provided. It can be `secret: foo` and as long as that foo exists it should work

jsteenb2:
same failign data race test from tasks/backend

PR #16243, opened 2019-12-17
fix(ui): added check to ensure that windowperiod isn't calculated when the query is undefined

Asalem1:
Closes https://app.honeybadger.io/projects/61122/faults/58676136

### Problem

The `FunctionSelector` component was triggering the `getWindowPeriod` function with an empty query when the component was initialized. The query would be blank at this time, so the `getMinDurationFromAST` function would throw an error, resulting a repetitive (not blocking) honeybadger error

### Solution

Added a check before running the `getWindowPeriod` function to see if the `query` is undefined or an empty string to ensure that the error won't be thrown due to an `invalid time range`

### DIFF
--- a/ui/src/shared/utils/errors.ts
+++ b/ui/src/shared/utils/errors.ts
@@ -12,6 +12,8 @@ if (CLOUD) {
 
 interface AdditionalOptions {
   component?: string
+  context?: object
+  name?: string
 }
 
 export const reportError = (
@@ -26,9 +28,9 @@ export const reportError = (
 /*
   Parse React's error boundary info message to provide the name of the
   component an error occured in.
- 
+
   For example, given the following info message:
- 
+
       The above error occurred in the <MePage> component:
           in MePage (created by ErrorBoundary(MePage))
           in ErrorBoundary (created by ErrorBoundary(MePage))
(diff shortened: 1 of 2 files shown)


ui/src/variables/utils/getWindowVars.ts
```diff
@@ -51,6 +51,9 @@ export const getWindowPeriod = (
   query: string,
   variables: VariableAssignment[]
 ): number | null => {
+  if (query === undefined || query.length === 0) {
```
ebb-tide: It makes me nervous that I don't know if and when query can be undefined here. 
I think we should only return null if `(!query.length)`

ui/src/variables/utils/getWindowVars.ts
```diff
@@ -64,7 +67,8 @@ export const getWindowPeriod = (
 
     return Math.round(queryDuration / DESIRED_POINTS_PER_GRAPH)
   } catch (error) {
-    reportError(error)
+    console.error(error)
+    reportError(error, {component: 'getWindowPeriod function'})
```
drdelambre: can we add the query here too?

Asalem1: We should be able to add that

hoorayimhelping:
```suggestion
    reportError(error, {context: { query }, component: 'getWindowPeriod function'})
```

ui/src/shared/utils/errors.ts
```diff
@@ -12,6 +12,8 @@ if (CLOUD) {
 
 interface AdditionalOptions {
   component?: string
+  context?: object
```
hoorayimhelping: any reason this shouldn't be

```ts
context?: {[key: string]: any}
```
?

it's a little more descriptive and makes it easier to limit the values allowed if we find that `any` is too open.

PR #16249, opened 2019-12-17
fix(kv): Prevent infinite loop condition when listing tasks by org.

brettbuddin:
Closes influxdata/idpe#5592

In the event that `findTaskByIDWithAuth` cannot find the task ID contained in the bucket, the outer loop will never terminate.

This ensures that we are calling `Next()` along-side any calls to `continue` while using a `Cursor`.

**Notes:**

I investigated improving the overall ergonomics of these cursor-based loops by hoisting the cursor advancement to the top of the iterations—instead of being at the tail-end of them. However, due to seeking behavior, the pre-conditions of the loop become a bit trickier. Perhaps this is an area of further improvement. Generally, loops that have cross-iteration state interaction tend to cause surprising behavior like this when `continue` gets involved.


stuartcarnie:
@brettbuddin 💯 for looking for ways to improve the loop ergonomics and your pragmatism for the urgency.

We could follow this up later with an improvement like this: https://play.golang.org/p/dygkrQYG2fu

Issue #16099, opened 2019-12-03
Retain user input when JSON parsing fails


hoorayimhelping:
The fix for [the app crashing when parsing invalid JSON](https://github.com/influxdata/honeybadger-errors/issues/3) revealed an issue whereby the JSON that was just parsed is removed from the form, but the form remains and empty.

![template](https://user-images.githubusercontent.com/146112/70089801-c6711480-15cd-11ea-9a40-7b62f52e5a44.gif)

Rather than clearing the form, we want to retain user input when JSON parsing fails.

TCL735:
Additional side effects that should be added:

- the text area should give an indication that the input is wrong. we can achieve this with ComponentStatus
- when the user attempts to correct the invalid input, the text area should no longer indicate that the input is wrong, and the notification should be dismissed

TCL735:
Dismissing the notification automatically when the user changes invalid input might not be necessary or desirable. The notification has details which might be useful for the user to reference while editing. An accidental edit would dismiss this potentially useful information.

As is, the notification will expire in 10 seconds. So, we'll go with that. The user can also manually dismiss the notification.

PR #16255, opened 2019-12-17
fix(ui): retain user input when parsing invalid JSON during import

TCL735:
Closes https://github.com/influxdata/influxdb/issues/16099

Added a third party library that wraps JSON.parse with linting and detailed error messaging.

For now, we are going to put the detailed error messages into the toast. As @hoorayimhelping mentioned, there is a potential for moving the detailed error message into something more permanent and useful for the user.

### DIFF
--- a/ui/cypress/e2e/dashboardsIndex.test.ts
+++ b/ui/cypress/e2e/dashboardsIndex.test.ts
@@ -74,19 +74,39 @@ describe('Dashboards', () => {
       cy.createDashboardTemplate(id)
     })
 
-    cy.getByTestID('empty-dashboards-list')
-      .getByTestID('add-resource-dropdown--button')
-      .click()
-
-    cy.getByTestID('add-resource-dropdown--template').click()
-
+    cy.getByTestID('empty-dashboards-list').within(() => {
+      cy.getByTestID('add-resource-dropdown--button').click()
+      cy.getByTestID('add-resource-dropdown--template').click()
+    })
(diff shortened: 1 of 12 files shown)


ui/cypress/e2e/dashboardsIndex.test.ts
```diff
@@ -74,21 +74,41 @@ describe('Dashboards', () => {
...
+    })
+
+    cy.getByTestID('add-resource-dropdown--import').click()
+    cy.contains('Paste').click()
+    cy.getByTestID('textarea')
+      .click()
+      .type('this is invalid JSON')
+    cy.get('button[title*="Import JSON"]').click()
+    cy.getByTestID('textarea--error').should('have.length', 1)
+    cy.getByTestID('textarea').should($s =>
```
zoesteinkamp: We might want to change this test id to be more descriptive, textarea seems way too vague. Same with page header possibly too. 

Like this is good getByTestID('add-resource-dropdown--button').

TCL735: Likely, this test will change again in the future when a more fully-featured text editor is put in place rather than just a plain <TextArea> per @alexpaxton. When the new editor is put in place, we will be sure to use a more descriptive test ID.

TCL735: Actually, you're right. "textarea" is too generic. I have made the change.

Issue #16138, opened 2019-12-05
More Honeybadger Context


drdelambre:
In the past few days as we've brought honeybadger more into the forefront of our workflow, we've found a few crucial tidbits of information missing from our debug loops. It'd be nice to add these to every error report to speed triage:
- [ ] user id
- [ ] git hash of the code the user is currently on

Additionally, it'd be helpful to have specific context around errors. For example an error in the query builder should show the query being submitted.

PR #16258, opened 2019-12-18
chore: add getUserFlags utility function

hoorayimhelping:
Adds a utility function to `featureFlags`, `getUserFlags`. This was part of work I did on #16138 that I pulled out into its own logical commit, hence no issue number.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/shared/utils/featureFlag.ts
+++ b/ui/src/shared/utils/featureFlag.ts
@@ -1,7 +1,7 @@
 import {FunctionComponent} from 'react'
 import {CLOUD, CLOUD_BILLING_VISIBLE} from 'src/shared/constants'
 
-const OSS_FLAGS = {
+export const OSS_FLAGS = {
   alerting: false,
   deleteWithPredicate: false,
   monacoEditor: false,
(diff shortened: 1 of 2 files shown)


ui/src/shared/utils/featureFlag.ts
```diff
@@ -63,36 +63,21 @@ export const FeatureFlag: FunctionComponent<{
...
-          if (typeof curr[0] === 'boolean') {
-            return prev
-          }
-
-          prev[curr[0]] = curr[1]
-          return prev
-        }, {})
-    )
-  }
+  console.table(getUserFlags())
```
hoorayimhelping: as far as I could tell in my local testing, `getUserFlags` has the same return signature as 

```ts
 Object.keys(CLOUD_FLAGS || OSS_FLAGS)
        .map(k => [k, isFlagEnabled(k)])
        .reduce((prev, curr) => {
          if (typeof curr[0] === 'boolean') {
            return prev
          }

          prev[curr[0]] = curr[1]
          return prev
        }, {})
```

PR #16263, opened 2019-12-18
feat(tasks): remove flux stats from run log and replace with trace id

GeorgeMac:
This removes the flux statistics from the task run log.

In place of the flux statistics we now have the trace ID and whether the trace was sampled.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/task/backend/executor/executor.go
+++ b/task/backend/executor/executor.go
@@ -126,15 +126,18 @@ func (p *syncRunPromise) Cancel() {
 
 func (p *syncRunPromise) finish(res *runResult, err error) {
 	p.finishOnce.Do(func() {
-		defer p.logEnd()
+		defer func() {
+			// Always cancel p's context.
+			// If finish is called before p.qs.Query completes, the query will be interrupted.
+			// If afterwards, then p.cancel is just a resource cleanup.
+			p.cancel()
 
-		// Always cancel p's context.
-		// If finish is called before p.qs.Query completes, the query will be interrupted.
-		// If afterwards, then p.cancel is just a resource cleanup.
-		defer p.cancel()
+			p.logEnd()
+
+			close(p.ready)
+		}()
 
 		p.res, p.err = res, err
-		close(p.ready)
 
 		if err != nil {
 			p.log.Debug("Execution failed to get result", zap.Error(err))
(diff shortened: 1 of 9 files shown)


task/backend/executor/task_executor_test.go
```diff
@@ -19,31 +20,56 @@ import (
...
+		sampler := jaeger.NewConstSampler(true)
+		tracer, closer := jaeger.NewTracer("task_backend_tests", sampler, jaeger.NewInMemoryReporter())
+		defer closer.Close()
+
+		opentracing.SetGlobalTracer(tracer)
+
+		code = m.Run()
+	}()
+
+	os.Exit(code)
```
GeorgeMac: note for reviewer: because of the nature of opentracing using a global variable for the tracer + the fact we use `t.Parallel()` everywhere I was triggering race conditions when setting this tracer in the specific tests I needed it.
I moved it up here to ensure this happens once at the start and then no tracer setting shenanigans could trigger a race later on.

task/backend/executor/executor.go
```diff
@@ -126,15 +126,18 @@ func (p *syncRunPromise) Cancel() {
...
+			p.cancel()
+
+			p.logEnd()
 
-		// Always cancel p's context.
-		// If finish is called before p.qs.Query completes, the query will be interrupted.
-		// If afterwards, then p.cancel is just a resource cleanup.
-		defer p.cancel()
+			close(p.ready)
+		}()
```
GeorgeMac: I changed this because I found a race condition here with the `logEnd()` happening after `close(p.ready)`.
This mean the called to `promise.Wait()` would unblock before the final log was made.

task/backend/executor/task_executor.go
```diff
@@ -400,12 +400,10 @@ func (w *worker) executeQuery(p *promise) {
 
 	it.Release()
 
-	// log the statistics on the run
-	stats := it.Statistics()
-
-	b, err := json.Marshal(stats)
-	if err == nil {
-		w.te.tcs.AddRunLog(p.ctx, p.task.ID, p.run.ID, time.Now().UTC(), string(b))
+	// log the trace id and whether or not it was sampled into the run log
```
gavincabbage: @george there's a helper for this in the logger package that might make things a bit more readable

https://github.com/influxdata/influxdb/blob/master/logger/fields.go#L98

GeorgeMac: @gavincabbage I decided to not use that as I already had the span off the context. What do you think. Split that helper into two parts?

gavincabbage: ah gotcha... yeah you could do that, the reflection makes things harder to grok, but I don't feel strongly either way

PR #16266, opened 2019-12-18
fix: update test to use 'valid' marshalled predicate

rawkode:
- [x] Rebased/mergeable
- [x] Tests pass


storage/wal/dump_test.go
```diff
@@ -118,7 +118,7 @@ func TestWalDumpRun_DeleteRangeEntries(t *testing.T) {
 		BucketID:  influxdb.ID(2),
 		Min:       3,
 		Max:       4,
-		Predicate: []byte("predicate"),
+		Predicate: []byte("\x00"),
```
e-dard: this could also just be `nil` I think

PR #16268, opened 2019-12-18
fix(ui): fixed flakey tests

### LINKED ISSUE #16253: General Sigin/Setup flakiness
### Problem

The `cypress.signin()` function is flakey and sometimes hangs on the `api/v2/setup` causing the tests to fail indiscriminately. 

Screen Shot 2019-12-17 at 10 09 10 AM

Asalem1:
Closes #16253

### Problem

flush & signin would flake if called multiple times in the same test suite

### Solution

consolidated the logic that needed to be retrieved from signing in into the first beforeEach call and removed the multiple usages of `cy.flush` and `cy.signin` since the test suite was logging in successfully and was flaking when we would flush the credentials and sign in again due to idiosyncrasies in the way the tests were running


drdelambre:
so it was already logged in, then broke itself trying to login again?

Asalem1:
> so it was already logged in, then broke itself trying to login again?

Something like that. We were also `flush`ing the credentials so if the operation hadn't stopped running, we would all of a sudden lose context and not have specific properties that may have been available at the outset of the test

PR #16271, opened 2019-12-18
feat: add an env script that will configure the environment for using go with libflux

jsternberg:


jsternberg:
I updated the script so it could differentiate if it was being executed as a script or sourced. This makes it so when executed as a script it can default to outputting the environment. @rockstar can you take a second look at this?

env
```diff
@@ -0,0 +1,25 @@
...
+done
+shift $((OPTIND-1))
+
+CGO_LDFLAGS=$(go run github.com/influxdata/flux/internal/cmd/flux-config --libs $verbose_flag)
+export CGO_LDFLAGS
+
+# If this script is being executed, it will be executed under bash
+# so the bash source variable should be present. If the variable
+# matches the current script, then we are being executed and the
+# default argument should be to print the environment.
```
rockstar: This is a helpful comment.

PR #16278, opened 2019-12-18
chore: add more context to honeybadger errors

### LINKED ISSUE #16138: More Honeybadger Context

hoorayimhelping:
Closes #16138

1. Adds the user id, org id, and current feature flag status to HoneyBadger error context
1. If the environment is not `CLOUD` print a table of the context that would have been sent to honeybadger

[See the context](https://app.honeybadger.io/projects/61122/faults/58718222#notice-context) on a contrived, but actual HoneyBadger error.

Local context in the console:
Screen Shot 2019-12-18 at 10 31 26 AM

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/organizations/actions/orgs.ts
+++ b/ui/src/organizations/actions/orgs.ts
@@ -2,6 +2,7 @@
 import {Dispatch} from 'redux'
 import {ThunkAction} from 'redux-thunk'
 import {push, RouterAction} from 'react-router-redux'
+import HoneyBadger from 'honeybadger-js'
 
 // APIs
 import {getErrorMessage} from 'src/utils/api'
@@ -80,6 +81,9 @@ export interface SetOrg {
 }
 
 export const setOrg = (org: Organization): SetOrg => {
+  HoneyBadger.setContext({
+    orgID: org.id,
+  })
   return {
     type: ActionTypes.SetOrg,
     payload: {org},
--- a/ui/src/shared/actions/me.ts
+++ b/ui/src/shared/actions/me.ts
@@ -1,5 +1,6 @@
 import {MeState} from 'src/shared/reducers/me'
 import {client} from 'src/utils/api'
+import HoneyBadger from 'honeybadger-js'
 
 export enum ActionTypes {
   SetMe = 'SET_ME',
@@ -25,8 +26,12 @@ export const getMe = () => async dispatch => {
   try {
     const user = await client.users.me()
 
+    HoneyBadger.setContext({
+      user_id: user.id,
+    })
+
     dispatch(setMe(user))
-  } catch (e) {
-    console.error(e)
+  } catch (error) {
+    console.error(error)
   }
 }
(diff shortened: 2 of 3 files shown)


ui/src/shared/actions/me.ts
```diff
@@ -25,8 +27,12 @@ export const getMe = () => async dispatch => {
...
 
+    HoneyBadger.setContext({
+      user_id: user.id,
+    })
+
     dispatch(setMe(user))
-  } catch (e) {
-    console.error(e)
+  } catch (error) {
+    console.error(error)
```
Asalem1: It looks like you're importing `reportError` but not using it in the `catch`, did we want to report the error for this?

https://github.com/influxdata/influxdb/pull/16278/files#diff-023642cbf29399507646a13659e14d24R3

ui/src/shared/utils/errors.ts
```diff
@@ -10,18 +12,44 @@ if (CLOUD) {
...
 }
 
-interface AdditionalOptions {
+// See https://docs.honeybadger.io/lib/javascript/guides/reporting-errors.html#additional-options
+interface HoneyBadgerAdditionalOptions {
   component?: string
   context?: {[key: string]: any}
+  cookies?: {[key: string]: any}
   name?: string
+  params?: {[key: string]: any}
```
Asalem1: Great job on linking the docs

ui/src/shared/actions/me.ts
hoorayimhelping: @Asalem1 i had `reportError` being called there, but took it out because that introduces a separate logical change to this PR beyond adding honeybadger context and that's something i'm trying to do less of.

ui/src/organizations/actions/orgs.ts
```diff
@@ -80,6 +81,9 @@ export interface SetOrg {
 }
 
 export const setOrg = (org: Organization): SetOrg => {
+  HoneyBadger.setContext({
```
drdelambre: when a user changes orgs / logs out and logs in as someone else, does this just overwrite in honey badger?

hoorayimhelping: @drdelambre great question, love where your head is at. The docs don't specifically mention this, but from what I've seen, i'm fairly certain that calling `setContext` will overwrite an existing key, but merge new keys.

>Use `Honeybadger.setContext` to add context to error reports. You can call `setContext` as many times as you like, and it will be reset on each page load. New context data is merged with existing data:

https://docs.honeybadger.io/lib/javascript/guides/customizing-error-reports.html#custom-metadata-context

I don't think we want to clear it though, because it will clear out all the keys:

>If you've used `Honeybadger.setContext` to store context data, you can clear it with `Honeybadger.resetContext`

PR #16280, opened 2019-12-18
feat(cmd): flags override env vars

### LINKED ISSUE #16246: cli flags should override environmental variables
We use env vars to supplement our CLI, e.g. allowing users to scope all requests to a specific org without having to specify it in every request.

If a user specifies a flag that conflicts with an env var, the flag should take precedence over the env var.

1) Investigate where this is currently happening in the CLI.
2) Change all instances where env var takes precedence over flags.

dearyhud:
Closes #16246  

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influx/query.go
+++ b/cmd/influx/query.go
@@ -10,7 +10,6 @@ import (
 	platform "github.com/influxdata/influxdb"
 	_ "github.com/influxdata/influxdb/query/stdlib"
 	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
 )
 
 var queryCmd = &cobra.Command{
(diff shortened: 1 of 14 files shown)


cmd/influx/query.go
```diff
@@ -23,31 +22,20 @@ or execute a literal Flux query contained in a file by specifying the file prefi
 }
 
 var queryFlags struct {
-	OrgID string
-	Org   string
+	organization
```
jademcgough: I'm not wild about this semantically, and I'm not sure there's much benefit to encapsulating OrgID/Org if they're going to be accessed directly. Just makes it a little harder to read/understand.

dearyhud: fair point, I tried to tread really lightly when I added this but crossed over boundaries a little. on the flip, this does DRY up a lot of cobra stuff across the board, and organization lookups by name all go through one code path [here](https://github.com/influxdata/influxdb/blob/6fb1b5bfa09d1bf6b2a1b42cdce7e70f251c0a11/cmd/influx/main.go#L254). I'm open to suggestions on where to put this as well

kelwang: I agree it's a little hard to understand at beginning. But overall, it would help us to make all of our flags more consistent. So still worth it in my opinion. Ideally, CLI should inherent all the behavior of the API, plus some interactive feature and file parsing.

russorat:
@jademcgough let's figure out what to do with this PR.

jsteenb2:
this was remedied remedied here: https://github.com/influxdata/influxdb/pull/16491

closing this issue as its completed now

PR #16291, opened 2019-12-19
fix: pass variables into a function

### LINKED ISSUE #16290: Cloud 2.0 GCP: InfluxDB Output plugin download does not match what is in UI
Environment: Cloud 2.0 GCP staging environment
Issue: The information that is displayed on the screen when you hit `InfluxDB Output Plugin` button on `Telegraf Configuration` page is correctly populated with `Organization` and `Bucket`, but when you hit `Download Config` button, it does not seem to download what is shown on screen and therefore the `Organization` and `Bucket` are not the correct entries in the downloaded configuration file. (See attached screenshot)
image

drdelambre:
Closes #16290

just looks like the template function wasn't being passed variables. also took time to normalize some internal state while rummaging around in there.

### DIFF
--- a/ui/src/telegrafs/components/TelegrafOutputOverlay.tsx
+++ b/ui/src/telegrafs/components/TelegrafOutputOverlay.tsx
@@ -71,24 +71,38 @@ class TelegrafOutputOverlay extends PureComponent<Props> {
     return <>{this.overlay}</>
   }
 
-  private get overlay(): JSX.Element {
-    const {server, org, orgID, buckets} = this.props
-    const _buckets = (buckets || [])
+  private get buckets() {
+    const {buckets} = this.props
+    return (buckets || [])
       .filter(item => item.type !== 'system')
       .sort((a, b) => {
         const _a = a.name.toLowerCase()
         const _b = b.name.toLowerCase()
         return _a > _b ? 1 : _a < _b ? -1 : 0
       })
+  }
+
+  private get currentBucket() {
+    const _buckets = this.buckets
     const {selectedBucket} = this.state
-    let bucket_dd = null
-    let bucket = null
 
     if (_buckets.length) {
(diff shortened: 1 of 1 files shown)


ui/src/telegrafs/components/TelegrafOutputOverlay.tsx
```diff
@@ -144,9 +158,15 @@ class TelegrafOutputOverlay extends PureComponent<Props> {
   }
 
   private handleDownloadConfig = () => {
+    const {server, org} = this.props
+    const bucket = this.currentBucket
     const config = transform(
       TELEGRAF_OUTPUT,
-      Object.assign({}, OUTPUT_DEFAULTS, {})
+      Object.assign({}, OUTPUT_DEFAULTS, {
+          server,
```
drdelambre: this is the actual magic

ui/src/telegrafs/components/TelegrafOutputOverlay.tsx
```diff
@@ -144,9 +158,15 @@ class TelegrafOutputOverlay extends PureComponent<Props> {
   }
 
   private handleDownloadConfig = () => {
+    const {server, org} = this.props
+    const bucket = this.currentBucket
     const config = transform(
       TELEGRAF_OUTPUT,
-      Object.assign({}, OUTPUT_DEFAULTS, {})
+      Object.assign({}, OUTPUT_DEFAULTS, {
```
hoorayimhelping: interesting that this isn't being destructured, might be worth considering when thinking about patterns @ebb-tide.

also, it looks like there's an extra space in indentation below, i think you'll need to run `prettier:fix`

drdelambre: i dont understand

PR #16297, opened 2019-12-19
feat(pkger): add parser support for notification rules

jsteenb2:
- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/pkger/models.go
+++ b/pkger/models.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/url"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -29,6 +30,7 @@ const (
 	KindNotificationEndpointPagerDuty Kind = "notification_endpoint_pager_duty"
 	KindNotificationEndpointHTTP      Kind = "notification_endpoint_http"
 	KindNotificationEndpointSlack     Kind = "notification_endpoint_slack"
+	KindNotificationRule              Kind = "notification_rule"
 	KindPackage                       Kind = "package"
 	KindTelegraf                      Kind = "telegraf"
 	KindVariable                      Kind = "variable"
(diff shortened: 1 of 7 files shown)


pkger/models.go
```diff
@@ -107,7 +112,11 @@ func (k Kind) ResourceType() influxdb.ResourceType {
 }
 
 func (k Kind) title() string {
-	return strings.Title(k.String())
+	pieces := strings.Split(string(k), "_")
+	for i := range pieces {
+		pieces[i] = strings.Title(pieces[i])
```
jademcgough: heh, didn't know this was in strings

pkger/models.go
```diff
@@ -922,18 +960,19 @@ func (c *check) valid() []validationErr {
...
-			Msg:   "must be 1 of [active, inactive]",
-		})
-	}
 	if c.statusMessage == "" {
 		vErrs = append(vErrs, validationErr{
 			Field: fieldCheckStatusMessageTemplate,
 			Msg:   `must provide a template; ex. "Check: ${ r._check_name } is: ${ r._level }"`,
 		})
 	}
+	if status := c.Status(); status != influxdb.Active && status != influxdb.Inactive {
```
jademcgough: demorgans law 👍

pkger/models.go
```diff
@@ -1415,6 +1455,163 @@ func (n mapperNotificationEndpoints) Len() int {
...
+func (r *notificationRule) Name() string {
+	return r.name
+}
+
+func (r *notificationRule) ResourceType() influxdb.ResourceType {
+	return KindNotificationRule.ResourceType()
+}
+
+func (r *notificationRule) Status() influxdb.Status {
+	if r.status == "" {
```
jademcgough: When would status be unset for notification rules?

jsteenb2: good question. After speaking with @russorat, we decided that in pkger, you don't have to provide a status, it'll just default to `active` if not provided.

pkger/models.go
jsteenb2: `strings` actually has a ton of useful funcs. Has stuff for `map`, 'trim`ing, and a boat load of other goodies that come up quite often

jsteenb2:
tests failing upstream b/c of go vendor issue. This is not related to this PR work. Should not have any affect on any upstream stuffs, no new vendored code here, nor any changes to existing stuffs being used by idpe

pkger/models.go
jademcgough: got it, that makes sense

PR #16300, opened 2019-12-20
feat: generate types for tests

drdelambre:
tests are failing from the makefile because types aren't being generated from swagger. this runs the generate command more often

### DIFF
--- a/ui/Makefile
+++ b/ui/Makefile
@@ -9,15 +9,18 @@ node_modules:
 	yarn install
 
 e2e: node_modules
+	yarn generate
 	yarn test:junit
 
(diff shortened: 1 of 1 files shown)


ui/Makefile
```diff
@@ -9,15 +9,18 @@ node_modules:
...
 
 e2e: node_modules
+	yarn generate
 	yarn test:junit
 
 build: node_modules $(UISOURCES)
 	yarn build
 
 lint: node_modules $(UISOURCES)
+	yarn generate
```
jsteenb2: probably better off making another make target and making that a dep on line 18 here and elsewhere

PR #16303, opened 2019-12-20
refactor: switch influxdb to use proper module import syntax

jsternberg:


russorat:
@jsternberg what's the status on this PR? do we still want it or should it be closed?

jsternberg:
I would need to run the same script and update it. This is mostly just dependent on when we want to do it.

jsternberg:
Done as part of a different pull request.

PR #16305, opened 2019-12-20
feat(pkger): add application functionality for the notification rules

jsteenb2:
- [x] Rebased/mergeable
- [x] Tests pass

Screen Shot 2019-12-20 at 10 19 48 AM


pkger/service.go
```diff
@@ -1032,6 +1043,18 @@ func (s *Service) Apply(ctx context.Context, orgID, userID influxdb.ID, pkg *Pkg
 		}
 	}
 
+	// this is required after first primary run and before secondary run, b/c this has dependencies
```
jademcgough: Nit, but feels like this comment could be simplified / made more easy to understand.

Something like
"This has to be run after the above primary resources, because it relies on notification endpoints already being applied."

jsteenb2: excellent point! thank you for improving that, struggled to figure out what to write, your version is miles ahead of where mine was lol

PR #16310, opened 2019-12-20
fix(tests): make sure NATS port is open for unit tests

### LINKED ISSUE #10628: failed to start nats streaming server
When using `influx` binaries from nightly builds, starting `influxd` might fail with 
```
2018-12-12T19:16:42.235195Z    error    failed to start nats streaming server    {"log_id": "0CKyR7IW000", "error": "unable to start a NATS Server on localhost:4222"}
Error: unable to start a NATS Server on localhost:4222
Usage:
  influxd [flags]

Flags:
      --bolt-path string           path to boltdb database (default "/Users/gershonshif/.influxdbv2/influxd.bolt")
      --developer-mode             serve assets from the local filesystem in developer mode
      --engine-path string         path to persistent engine files (default "/Users/gershonshif/.influxdbv2/engine")
  -h, --help                       help for influxd
      --http-bind-address string   bind address for the REST HTTP API (default ":9999")
      --log-level string           supported log levels are debug, info, and error (default "info")
      --nats-path string           path to NATS queue for scraping tasks (default "/Users/gershonshif/.influxdbv2/nats")

unable to start a NATS Server on localhost:4222
```

jacobmarble:
Fixes #10628 again

- [x] Rebased/mergeable
- [x] Tests pass

Manually tested #10628 is fixed:
- run influxd
- lsof to verify influxd is listening on port 4222
- git checkout master
- `go test ./cmd/...` fails due to NATS port 4222
- git checkout jgm-fix-nats (this branch)
- `got test ./cmd/...` does not fail

### DIFF
--- a/cmd/influxd/launcher/launcher.go
+++ b/cmd/influxd/launcher/launcher.go
@@ -720,7 +720,6 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 
 	// NATS streaming server
 	natsOpts := nats.NewDefaultServerOptions()
-	nextPort := int64(4222)
 
 	// Welcome to ghetto land. It doesn't seem possible to tell NATS to initialise
 	// a random port. In some integration-style tests, this launcher gets initialised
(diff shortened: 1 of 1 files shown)


cmd/influxd/launcher/launcher.go
```diff
@@ -922,6 +929,18 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 	return nil
 }
 
+// isAddressPortAvailable checks whether the address:port is available to listen,
+// by using net.Listen to verify that the port opens successfully, then closes the listener.
+func isAddressPortAvailable(address string, port int) (bool, error) {
+	if l, err := net.Listen("tcp", fmt.Sprintf("%s:%d", address, port)); err == nil {
+		if err := l.Close(); err != nil {
```
mark-rushakoff: Seems like this would still be racy. Two parallel test packages could contend for the same port if one package opens and closes the port, and the next package immediately opens and closes the same port.

If it's an improvement over the current code, this is probably still mergeable, but it ought to be commented that this is known to be racy.

The better option would be to instruct nats to listen on `:0` and have it report back what port it is listening on, but I assume it doesn't support the latter.

jacobmarble: Right, the NATS package doesn't seem to accommodate the way we are writing unit tests.

I thought the longer-term plan was to remove NATS altogether, but I can't find an open issue for that.

PR #16311, opened 2019-12-20
fix: speeding them tests right up

drdelambre:
tests jumped from ~1min to ~20min on circleci some time last week, prompting parallelization of those tests in order to run in a usable test window. This worked fine until someone made a good case for running the tests outside of the CI pipeline, in which tests took up to 6hrs on one user's machine.

i did some sleuthing

and it turns out that the problem was just that honey badger is being included in all of these tests as of this PR: https://github.com/influxdata/influxdb/pull/16194

so our tests were breaking our tests

I mocked honey badger from the jest tests in the test setup file (applies to all tests), and now tests are back down to ~1min

### DIFF
--- a/ui/jestSetup.ts
+++ b/ui/jestSetup.ts
@@ -1,5 +1,7 @@
 import {cleanup} from 'react-testing-library'
 
+jest.mock('honeybadger-js', () => () => null)
+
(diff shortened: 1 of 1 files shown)


hoorayimhelping:
awesome sluething! love it


EZzRAVv

PR #16319, opened 2019-12-23
Add custom check script editor

### LINKED ISSUE #16075: Add Customize Check Query Button and Script Editor State to Alert Builder
Screen Shot 2019-11-27 at 12 44 01 PM copy

When the button is clicked Alert Builder state should switch to Script Editing mode, with a templated check flux query pre- filled in the Script Editor. 

When the script button is clicked, add `option monitor.write = yield` extern to query 
as described in https://github.com/influxdata/influxdb/issues/16037#issuecomment-561403622

Button Text Should change to: Return to Alert Builder, and should warn users of destructive changes in case user has edited script.

ebb-tide:
Closes #16075 

Adds customize check script editor behind feature flag. 

User can begin creating a threshold or deadman check, and then switch to a check script editor mode to see the flux script that the BE would have created for the parameters in the alert builder. 
User can hit Execute check query and view statuses to view the statuses that the check would have written to monitoring bucket. monitor.write is overwritten by yield function to prevent user from writing to monitoring bucket outside of a check. 
Currently user can not go back to alert builder after customizing check. 

As part of this PR:
Took alertBuilder state out from time machine. 
Changed alertBuilder redux state to store check properties rather than the check itself.


http/check_service.go
```diff
@@ -528,7 +529,6 @@ func (h *CheckHandler) handlePostCheck(w http.ResponseWriter, r *http.Request) {
 		h.HandleHTTPError(ctx, err, w)
 		return
 	}
-	h.log.Debug("Check created", zap.String("check", fmt.Sprint(chk)))
```
kelwang: i think this is still being used. Am I right, @gshif?

desa: @kelwang `fmt.Sprint`ing the raw check is maybe not the thing to do though. This can easily leak sensitive information like the secrets on a check.

notification/check/custom.go
```diff
@@ -0,0 +1,356 @@
...
+	OwnerID      influxdb.ID             `json:"ownerID,omitempty"`
+	OrgID        influxdb.ID             `json:"orgID,omitempty"`
+	Query        influxdb.DashboardQuery `json:"query"`
+	TaskID       influxdb.ID             `json:"taskID,omitempty"`
+	AlertBuilder AlertBuilder            `json:"alertBuilder"`
+	CreatedAt    time.Time               `json:"createdAt"`
+	UpdatedAt    time.Time               `json:"updatedAt"`
+}
+
+// UI
```
ebb-tide: deniz please remove these comments duh

notification/check/custom.go
```diff
@@ -0,0 +1,356 @@
...
+// Valid checks whether check flux is valid, returns error if invalid
+func (c *Custom) Valid() error {
+	if err := c.hasRequiredTaskOptions(); err != nil {
+		return err
+	}
+	if err := c.hasRequiredCheckParameters(); err != nil {
+		return err
+	}
+
+	script, err := c.sanitizeFlux()
```
ebb-tide: add comment about what we do with ids

notification/check/custom_test.go
```diff
@@ -0,0 +1,135 @@
...
+					},
+				},
+			},
+			wants: wants{
+				err:    nil,
+				script: ast.Format(parser.ParseSource(fmt.Sprintf(validQuery, "000000000000000a"))),
+			},
+		},
+		{
+			name: "valid flux script is valid and unchanged",
```
ebb-tide: check these names reflect what test is doing

ui/src/alerting/actions/checks.ts
```diff
@@ -140,50 +148,106 @@ export const saveCheckFromTimeMachine = () => async (
   dispatch: Dispatch<any>,
   getState: GetState
 ) => {
-  const state = getState()
-  const {
-    orgs: {
-      org: {id: orgID},
-    },
-  } = state
+  try {
```
ebb-tide: let's test this

PR #16325, opened 2019-12-23
feat(ui): Added Last run status and error messages to check, tasks, and notification rules

### LINKED ISSUE #15838: add last run status to checks/rules
Similar to tasks, we should return
```
	LatestScheduled time.Time              `json:"latestScheduled,omitempty"`
	LastRunStatus   string                 `json:"lastRunStatus,omitempty"`
	LastRunError    string                 `json:"lastRunError,omitempty"`
```

for checks and notification rules.

We should display something similar to the tasks UI.

Screenshot 2019-11-11 11 38 44

This will involved modifying the checks and notifications UI and API. In the API, we'll need to grab those values from the task that backs the check/rule.

### LINKED ISSUE #15844: Display lastRunStatus/Error in the task UI
The tasks API now returns
```
	LatestScheduled time.Time              `json:"latestScheduled,omitempty"`
	LastRunStatus   string                 `json:"lastRunStatus,omitempty"`
	LastRunError    string                 `json:"lastRunError,omitempty"`
```

We should display those in the UI

Screenshot 2019-11-11 11 38 44

Asalem1:
Closes #15844
Closes #15838 

### Problem

check rules and notification rules would not receive relevant data from the API. The UI wasn't built out to display the data. Using the client library vs the generatedRoutes results in a lot of TS issues

### Solution

Refactored a good chunk of the client library usage related to the ticket to utilize the generatedRoutes to resolve TS issues. Updated the UI to display newly returned data with the help of @alexpaxton 


http/notification_rule.go
```diff
@@ -146,9 +147,13 @@ type notificationRuleLinks struct {
 
 type notificationRuleResponse struct {
 	influxdb.NotificationRule
-	Labels []influxdb.Label      `json:"labels"`
-	Links  notificationRuleLinks `json:"links"`
-	Status string                `json:"status"`
+	Labels          []influxdb.Label      `json:"labels"`
+	Links           notificationRuleLinks `json:"links"`
+	Status          string                `json:"status"`
+	LatestCompleted string                `json:"latestCompleted,omitempty"`
```
kelwang: You can just use time.Time for this field. It will work fine

ui/src/dataLoaders/actions/dataLoaders.ts
```diff
@@ -376,12 +376,19 @@ export const createOrUpdateTelegrafConfigAsync = () => async (
   )
 
   if (telegrafConfigID) {
-    const telegraf = await client.telegrafConfigs.update(telegrafConfigID, {
-      name: telegrafConfigName,
-      description: telegrafConfigDescription,
-      plugins,
+    const resp = await api.putTelegraf({
```
ebb-tide: Are you comparing this function from oats to the one you are replacing from the other client as you go and making sure they are entirely equivalent? It is making me really nervous that we are essentially rewriting the behavior of the app with these changes just to get label types to match. I would strongly prefer if we made these switches from one client to another each in a separate PR instead of one mega PR.

ui/src/tasks/actions/index.ts
```diff
@@ -3,9 +3,9 @@ import {push, goBack} from 'react-router-redux'
 import _ from 'lodash'
 
 // APIs
-import {LogEvent, ITask as Task} from '@influxdata/influx'
 import {client} from 'src/utils/api'
 import {notify} from 'src/shared/actions/notifications'
+import * as api from 'src/client'
```
ebb-tide: I think it's a really bad idea for us to use two different clients to interact with fetching/writing the same resource(tasks). If we are switching to `src/client` for tasks let's make sure each function is swapped with one that is functionally equivalent.

ui/src/tasks/actions/index.ts
```diff
@@ -296,7 +312,7 @@ export const updateTaskStatus = (task: Task) => async dispatch => {
 
 export const updateTaskName = (task: Task) => async dispatch => {
   try {
-    await client.tasks.update(task.id, task)
+    await api.patchTask({taskID: task.id, data: task})
```
ebb-tide: These two functions are not equivalent. client.tasks.update has:
```
    if (!!updates.cron) {
      delete original.every
    }

    if (!!updates.every) {
      delete original.cron
    }
```
which we would have to do in the UI code if we are switching from one client to the other

ui/src/telegrafs/actions/index.ts
```diff
@@ -1,8 +1,10 @@
 // API
 import {client} from 'src/utils/api'
+import * as api from 'src/client'
```
ebb-tide: again makes me really nervous to be using two different clients to be interacting with same resource. Let's either switch or not switch

Asalem1:
> This PR makes a lot more functional changes than is expressed in the PR description. Every place we replace `@influxdata/influx` with `src/client` is a place where we are changing the behavior of the app. I would like for us to make these changes for each resource one by one in separate PRs, rather than as part of making Label types match. The `@influxdata/influx` client adds default properties to labels so that the UI can rely on there existing a color and description property for each label by applying defaults if they don't exist even though they are not guaranteed by the API. oats returns values as they are provided by the API. I am not certain how we should go about replacing this functionality. One option is to add defaults to labels every time we fetch a resource? In this PR we are going in the direction of removing label defaults from resources that used to have them, rather than adding defaults to labels that don't currently have them and I'm not sure this is the direction to go.
> 
> Also, I think it's a bad idea for us to make API calls relating to a particular resource through more than one client. Clients have slightly different expectations/behaviors, and getting from one and posting with another is really difficult to reason about.

I totally hear where you're coming from and would agree with you in that this should've been compartmentalized. The issue kind of transformed itself into what it was based on resolving linter errors, hence the reason for its crazy scope. Once Tasks were transformed to update the type, there was a cascade of linter errors that had to be addressed that ran me down from tasks -> labels -> templates -> variables

http/notification_rule.go
Asalem1: Hey @kelwang - I'm really new to go so my apologies if this is a straightforward SO answer, but I couldn't seem to format the dates using `https://github.com/influxdata/influxdb/pull/16325/files/8a7cd82dc1ad34dcc775b58e6af4526079dc7704#diff-0045aab5f4f34cfeb0f1646f5710df0aR213-R214` when the data was typecast as `time.Time`. Is there something i'm missing?

http/check_test.go
```diff
@@ -703,7 +709,9 @@ func TestService_handlePostCheck(t *testing.T) {
   "every": "5m",
   "level": "WARN",
 	"labels": [],
-	"status": "active"
+	"status": "active",
+	"latestCompleted": "0001-01-01T00:00:00Z",
+	"latestScheduled": "0001-01-01T00:00:00Z"
```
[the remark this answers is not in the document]
Asalem1: I completely agree in that we need to test out the `LastRunStatus` and `LastRunError`. They should be included for tests that result in an error

ui/src/alerting/components/CheckCard.tsx
```diff
@@ -170,7 +171,13 @@ const CheckCard: FunctionComponent<Props> = ({
         />
       }
       metaData={[
+        <>Last completed at {check.latestCompleted}</>,
         <>{relativeTimestampFormatter(check.updatedAt, 'Last updated ')}</>,
+        <LastRunTaskStatus
+          key={2}
```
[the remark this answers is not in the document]
Asalem1: Linter was asking for it saying that the element in the array didn't have a `key` prop. Not sure if it's the way that `clockface` is handling the data, but I could check the source and see if we're mapping the results there without adding a specified `key` that is triggering the linter error

ui/src/alerting/components/notifications/RuleCard.tsx
```diff
@@ -165,7 +161,13 @@ const RuleCard: FC<Props> = ({
         />
       }
       metaData={[
+        <>Last completed at {rule.latestCompleted}</>,
         <>{relativeTimestampFormatter(rule.updatedAt, 'Last updated ')}</>,
+        <LastRunTaskStatus
+          key={2}
```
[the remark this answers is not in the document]
Asalem1: https://github.com/influxdata/influxdb/pull/16325#discussion_r361470351

ui/src/alerting/components/CheckCard.tsx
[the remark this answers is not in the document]
ebb-tide: ah ok- no worries then!

ui/src/dataLoaders/actions/dataLoaders.ts
Asalem1: Hey @ebb-tide , great question since the `putTelegraf` was the only one that's been a question in my mind with respect to this refactor. I did use the `generateRoutes` file in conjunction with the old apis to ensure that the endpoints, data, etc... were all the same:

https://raw.githubusercontent.com/influxdata/influxdb-client-js/master/src/api/api.ts

As I mentioned earlier, this `putTelegraf` one was (and still is) a question mark since we currently don't require the addition of `plugins` for `putTelegraf` as part of the parameters. I'd love to discuss this if you're around today

ui/src/tasks/actions/index.ts
Asalem1: I hear ya. My reasoning behind not refactoring the entire `client` library was because I didn't want this to get too out of scope (which is ironic in a way since it did). I can go ahead and finish refactoring this file if you think that'd be prudent

ui/src/tasks/actions/index.ts
Asalem1: Good eyes. I'll make this change happen

ui/src/tasks/actions/index.ts
ebb-tide: I'm sorry this got out of hand! I think we should go the other direction and add defaults to check and notification rule Labels instead of removing them

ui/src/tasks/actions/index.ts
Asalem1: Great point, let me know if you're around to talk strategy on it

ui/src/telegrafs/actions/index.ts
Asalem1: I hear your concerns, let me know what you think in the previous comment

ui/src/tasks/actions/index.ts
Asalem1: No worries at all. @desa fully warned me, I just underestimated it. Would you mind elaborating a bit on what you mean by going in the other direction?

http/notification_rule.go
kelwang: yeah, that's the default format when you do json encoding of time.Time. see https://golang.org/src/time/time.go#L1264. You can see the example here: https://play.golang.org/p/6sozCf40hMh

Asalem1: Aw perfect, thanks for the heads up!

ui/src/alerting/actions/checks.ts
```diff
@@ -33,9 +33,9 @@ import {
   GetState,
   RemoteDataState,
   CheckViewProperties,
-  Label,
   PostCheck,
 } from 'src/types'
+import {Label} from 'src/client'
```
drdelambre: if you're for serious about this, it'd be easier to just import the Label definition into 'src/types' instead

ui/src/alerting/components/endpoints/EndpointCard.tsx
```diff
@@ -148,7 +144,7 @@ const EndpointCard: FC<Props> = ({
   }
   const labelsComponent = (
     <InlineLabels
-      selectedLabels={endpoint.labels as Label[]}
+      selectedLabels={endpoint.labels}
```
drdelambre: dope

ui/src/dashboards/actions/index.ts
```diff
@@ -16,7 +16,7 @@ import {
   updateView as updateViewAJAX,
 } from 'src/dashboards/apis'
 import {createDashboardFromTemplate as createDashboardFromTemplateAJAX} from 'src/templates/api'
-
+import * as api from 'src/client'
```
drdelambre: be explicit here. it's a contract between modules

ui/src/dataLoaders/actions/dataLoaders.ts
```diff
@@ -3,13 +3,13 @@ import _ from 'lodash'
 
 // Apis
 import {client} from 'src/utils/api'
+import * as api from 'src/client'
```
drdelambre: be explicit

ui/src/dashboards/actions/index.ts
Asalem1: but convention

Issue #16316, opened 2019-12-22
Service Definition for Notification Endpoints


jsteenb2:
This card is to provide a means of defining service boundaries and putting some blueprints in place. It will follow clean architecture principles. 

Things to consider:

- [ ] minimizing mental overhead for services (not needing to know to look some specific place for some spcific logic, we're gonna create some boundaries :-))
- [ ] providing horizontal CI scaling (can we easily run these tests in whole for the service in parallel with other services?)
- [ ] provide an integration to the http router (abstraction -> http.ResourceHandler)

PR #16327, opened 2019-12-26
chore(kv): refactoring store with reusable store type

jsteenb2:
this relates to #16316 

the main gist of this work is to provide some more rigidity around store access. Its a bit all over the place atm. This aims to make the 95% use case completely repeatable, and allow other stores to compose this into even more full fledged solutions

- [x] Rebased/mergeable
- [x] Tests pass
- [x] update upstream tests to pass with slightly changed error messages


kv/notification_endpoint.go
```diff
@@ -116,7 +74,26 @@ func (s *Service) createNotificationEndpoint(ctx context.Context, tx Tx, edp inf
 	edp.SetUpdatedAt(now)
 	edp.BackfillSecretKeys()
 
-	if err := s.putNotificationEndpoint(ctx, tx, edp); err != nil {
+	if err := edp.Valid(); err != nil {
```
jsteenb2: this validation will eventually get moved into service layer in #16316

kv/document_test.go
```diff
@@ -6,16 +6,6 @@ import (
 	influxdbtesting "github.com/influxdata/influxdb/testing"
 )
 
-func TestBoltDocumentStore(t *testing.T) {
```
kelwang: looks like the inmem test for this one is skipped

jsteenb2: same with the notificaiton rules store oddly enough. No idea what all those are about tbh 🤷‍♂.

kelwang: not a big deal, we will build the inmem transaction in another pr

jsteenb2:
upstream tests failing, already have fix upstream ready to merge

PR #16333, opened 2019-12-26
feat(ui): Show users errors from saving check before exiting time machine

### LINKED ISSUE #16257: Checks fails with naming collision when renaming check and no error to user
image

__Steps to reproduce:__

1. create a check in the UI
2. create a 2nd check
3. rename check to same name of 1st check (if no naming collision observed already)
4. should see errors in network tab

__Expected behavior:__
Expected to be notified of naming collision.

__Actual behavior:__
Fails silent, no toast or anything.

NOTE: was also confused I couldn't name or put description when creating it 🤔

ebb-tide:
Closes #16257

### DIFF
--- a/ui/src/alerting/actions/checks.ts
+++ b/ui/src/alerting/actions/checks.ts
@@ -23,6 +23,7 @@ import {
   setActiveTimeMachine,
   updateTimeMachineCheck,
   setCheckStatus,
+  setTimeMachineCheck,
 } from 'src/timeMachine/actions'
 import {executeQueries} from 'src/timeMachine/actions/queries'
 import {checkChecksLimits} from 'src/cloud/actions/limits'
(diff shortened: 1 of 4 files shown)


ui/src/alerting/actions/checks.ts
```diff
@@ -140,50 +141,59 @@ export const saveCheckFromTimeMachine = () => async (
...
-    throw new Error(resp.data.message)
+  if (getCheckResponse.status !== 200) {
+    throw new Error(getCheckResponse.data.message)
   }
 
-  dispatch(setCheck(resp.data))
+  return api.putCheck({
+    checkID: check.id,
+    data: {...check, ownerID: getCheckResponse.data.ownerID},
+  })
```
hoorayimhelping: does this need to dispatch a redux action or is this all good? looks like formerly it was `setCheck` ?

ui/src/alerting/actions/checks.ts
```diff
@@ -140,50 +141,59 @@ export const saveCheckFromTimeMachine = () => async (
...
+      dispatch(setCheck(resp.data))
+      dispatch(checkChecksLimits())
+
+      dispatch(push(`/orgs/${orgID}/alerting`))
+      dispatch(setTimeMachineCheck(RemoteDataState.NotStarted, null))
+    } else {
+      throw new Error(resp.data.message)
+    }
+  } catch (e) {
+    console.error(e)
```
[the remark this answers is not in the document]
ebb-tide: I think no in this case- since these are very reasonable user errors that we are alerting the user of- like they are trying to create a check with a name duplication etc. It is not a broken state in the UI. Thanks for thinking of it and reminding me though I'm sure there are places I should be sending honeybadger errors.

ui/src/alerting/actions/checks.ts
ebb-tide: it is happening on line 171 inside of `saveCheckFromTimeMachine`- the diff is misleading since I've replaced the `updateCheck` function that was being called directly from the component with `updateCheckFromTimeMachine` which is being called from `saveCheckFromTimeMachine` to hide the complexity with the get check we have to do because of the weirdness with putCheck requiring an ownerID.

PR #16335, opened 2019-12-26
chore(redux-thunk): fix typings

121watts:
Prerequisite for normalization.  We had no types for `redux-thunk` and / or they were typed incorrectly.


ui/src/buckets/actions/index.ts
```diff
@@ -180,6 +185,9 @@ export const deleteBucket = (id: string, name: string) => async (
 
     dispatch(removeBucket(id))
     dispatch(checkBucketLimits())
+    dispatch(() => ({
+      type: 'PLERPS',
```
ebb-tide: ? :D

ui/src/labels/actions/index.ts
```diff
@@ -17,7 +17,12 @@ import {
 } from 'src/shared/copy/notifications'
 import {GetState} from 'src/types'
 
-export type Action = SetLabels | AddLabel | EditLabel | RemoveLabel
+export type Action =
+  | SetLabels
+  | AddLabel
+  | EditLabel
+  | RemoveLabel
+  | NotifyAction
```
ebb-tide: Aren't we using this Action to define what the reducer can accept? I'm afraid this reduces the type safety of the reducer. What would you think about removing line 25 and adding another type `ThunkActions = Action | NotifyAction` that we could use for dispatch?

ui/src/timeMachine/actions/index.ts
```diff
@@ -100,6 +100,11 @@ export type Action =
   | ReturnType<typeof updateCheckThreshold>
   | ReturnType<typeof removeCheckThreshold>
 
+type ExternalActions =
```
ebb-tide: ooh love this! Maybe we could even import and export all Actions from a central place?

ui/src/buckets/actions/index.ts
```diff
@@ -1,4 +1,4 @@
-import {Dispatch} from 'redux-thunk'
+import {Dispatch} from 'react'
```
hoorayimhelping: can you give some background in this change _por favor_? like, why is it being imported from React now?

ui/src/timeMachine/actions/index.ts
```diff
@@ -100,6 +100,11 @@ export type Action =
...
+  | ReturnType<typeof loadBuckets>
+  | ReturnType<typeof saveAndExecuteQueries>
+  | ReturnType<typeof setValues>
```
hoorayimhelping: do all the the functions that rely on external actions call all these functions, or is this just the set of all the external functions?

ui/src/labels/actions/index.ts
121watts: What I ultimately want to do is centralize all actions into a single format and location.  So, yes!

ui/src/timeMachine/actions/index.ts
121watts: Yes, I want to do that in another PR.

ui/src/timeMachine/actions/index.ts
121watts: "`ExternalActions`" was just a name I thought of for actions not declared in this file.  Really, I want a central repository and format for all action types.

ui/src/buckets/actions/index.ts
121watts: `redux-thunk` doesn't actually have an exported`Dispatch` type.  So, `Dispatch` was really `any`

ui/src/buckets/actions/index.ts
121watts: oops, ill pull that out.  thanks for spotting that.

PR #16338, opened 2019-12-27
feat(ui): added notification rule & checks last run status & error

### LINKED ISSUE #15838: add last run status to checks/rules

Asalem1:
Closes #15838

### Solution

Added lastrunstatus check and error handling for notification rules and check rules

Screen Shot 2019-12-27 at 09 34 16
Screen Shot 2019-12-27 at 09 34 19

### DIFF
--- a/http/notification_rule_test.go
+++ b/http/notification_rule_test.go
@@ -147,7 +147,9 @@ func Test_newNotificationRuleResponses(t *testing.T) {
       ],
       "type": "slack",
       "updatedAt": "0001-01-01T00:00:00Z",
-      "status": "active"
+      "status": "active",
+			"latestCompleted": "0001-01-01T00:00:00Z",
+			"latestScheduled": "0001-01-01T00:00:00Z"
     },
     {
       "createdAt": "0001-01-01T00:00:00Z",
@@ -170,7 +172,9 @@ func Test_newNotificationRuleResponses(t *testing.T) {
       "runbookLink": "",
       "type": "pagerduty",
       "updatedAt": "0001-01-01T00:00:00Z",
-      "status": "active"
+      "status": "active",
+			"latestCompleted": "0001-01-01T00:00:00Z",
+			"latestScheduled": "0001-01-01T00:00:00Z"
     }
   ]
 }`,
@@ -287,7 +291,9 @@ func Test_newNotificationRuleResponse(t *testing.T) {
    }
  ],
  "type": "slack",
- "updatedAt": "0001-01-01T00:00:00Z"
+ "updatedAt": "0001-01-01T00:00:00Z",
+ "latestCompleted": "0001-01-01T00:00:00Z",
+	"latestScheduled": "0001-01-01T00:00:00Z"
 }`,
 		},
 	}
--- a/ui/src/alerting/actions/checks.ts
+++ b/ui/src/alerting/actions/checks.ts
@@ -182,6 +182,18 @@ export const saveCheckFromTimeMachine = () => async (
   }
 }
 
+export const updateCheck = (check: Partial<Check>) => async (
+  dispatch: Dispatch<Action | NotificationAction>
+) => {
+  const resp = await api.putCheck({checkID: check.id, data: check as Check})
+  if (resp.status === 200) {
+    dispatch(setCheck(resp.data))
+  } else {
+    throw new Error(resp.data.message)
+  }
+  dispatch(setCheck(resp.data))
+}
+
 const updateCheckFromTimeMachine = async (check: Check) => {
   // todo: refactor after https://github.com/influxdata/influxdb/issues/16317
   const getCheckResponse = await api.getCheck({checkID: check.id})
--- a/ui/src/alerting/components/CheckCard.tsx
+++ b/ui/src/alerting/components/CheckCard.tsx
@@ -7,6 +7,7 @@ import {withRouter, WithRouterProps} from 'react-router'
 import {SlideToggle, ComponentSize, ResourceCard} from '@influxdata/clockface'
 import CheckCardContext from 'src/alerting/components/CheckCardContext'
 import InlineLabels from 'src/shared/components/inlineLabels/InlineLabels'
+import LastRunTaskStatus from 'src/shared/components/lastRunTaskStatus/LastRunTaskStatus'
 
 // Constants
 import {DEFAULT_CHECK_NAME} from 'src/alerting/constants'
@@ -170,7 +171,13 @@ const CheckCard: FunctionComponent<Props> = ({
         />
       }
       metaData={[
+        <>Last completed at {check.latestCompleted}</>,
         <>{relativeTimestampFormatter(check.updatedAt, 'Last updated ')}</>,
+        <LastRunTaskStatus
+          key={2}
+          lastRunError={check.lastRunError}
+          lastRunStatus={check.lastRunStatus}
+        />,
       ]}
     />
   )
--- a/ui/src/alerting/components/notifications/RuleCard.tsx
+++ b/ui/src/alerting/components/notifications/RuleCard.tsx
@@ -7,6 +7,7 @@ import {withRouter, WithRouterProps} from 'react-router'
 import {SlideToggle, ComponentSize, ResourceCard} from '@influxdata/clockface'
 import NotificationRuleCardContext from 'src/alerting/components/notifications/RuleCardContext'
 import InlineLabels from 'src/shared/components/inlineLabels/InlineLabels'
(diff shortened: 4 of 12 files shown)


ui/src/alerting/actions/checks.ts
```diff
@@ -182,6 +182,18 @@ export const saveCheckFromTimeMachine = () => async (
   }
 }
 
+export const updateCheck = (check: Partial<Check>) => async (
```
Asalem1: this was previously removed and caused a breaking change. It has been added back here with the intention of refactoring this implementation with a different ticket

ui/src/alerting/components/CheckCard.tsx
```diff
@@ -170,7 +171,13 @@ const CheckCard: FunctionComponent<Props> = ({
         />
       }
       metaData={[
+        <>Last completed at {check.latestCompleted}</>,
         <>{relativeTimestampFormatter(check.updatedAt, 'Last updated ')}</>,
+        <LastRunTaskStatus
+          key={2}
```
Asalem1: added the `key` here to resolve linter errors

ui/src/alerting/components/notifications/RuleCard.tsx
```diff
@@ -165,7 +166,13 @@ const RuleCard: FC<Props> = ({
         />
       }
       metaData={[
+        <>Last completed at {rule.latestCompleted}</>,
         <>{relativeTimestampFormatter(rule.updatedAt, 'Last updated ')}</>,
+        <LastRunTaskStatus
+          key={2}
```
Asalem1: added `key` prop here to resolve linter errors

ui/src/alerting/components/CheckCard.tsx
hoorayimhelping: is this a react warning or a lint error? if react is generating a warning then that's the issue we should address. if lint is mistakenly saying we need a key, we should comment an override here, because it's adding meaning to the code to get around a tool issue. do you understand the difference? a real react warning about a list item not having a key is something we should fix, a mistaken lint error trying to head off that warning is something we should comment about, but not change our code to make go away.

if it's an actual react warning, we should give this a more unique name, if possible. react says to only use the item's list index as a last resort

>When you don’t have stable IDs for rendered items, you may use the item index as a key as a last resort:

>We don’t recommend using indexes for keys if the order of items may change. This can negatively impact performance and may cause issues with component state. 

https://reactjs.org/docs/lists-and-keys.html#keys

same comment and any changes that need to happen apply to `RuleCard`'s `key` change

ui/src/shared/components/lastRunTaskStatus/LastRunTaskStatus.tsx
```diff
@@ -0,0 +1,80 @@
...
+  Icon,
+  IconFont,
+  Popover,
+  Appearance,
+  PopoverInteraction,
+  ComponentColor,
+} from '@influxdata/clockface'
+
+// Styles
+import './LastRunTaskStatus.scss'
```
hoorayimhelping: is this where we want to import this? or should it be imported in in [chronograf.scss](https://github.com/influxdata/influxdb/blob/master/ui/src/style/chronograf.scss) ?

ui/src/shared/components/lastRunTaskStatus/LastRunTaskStatus.tsx
```diff
@@ -0,0 +1,80 @@
...
+interface PassedProps {
+  lastRunError?: string
+  lastRunStatus: string
+}
+
+const LastRunTaskStatus: FC<PassedProps> = ({lastRunError, lastRunStatus}) => {
+  const triggerRef = useRef<HTMLDivElement>(null)
+  const [highlight, setHighlight] = useState<boolean>(false)
+
+  let color = ComponentColor.Success
```
hoorayimhelping: are there implications to setting the default state to `success`? if a task run fails and we don't know about it, and we set the status to success, there might be a nonzero impact on the user? what kind of thought have we put into that?

ui/src/shared/components/lastRunTaskStatus/LastRunTaskStatus.tsx
```diff
@@ -0,0 +1,80 @@
...
+  let icon = IconFont.Checkmark
+  let text = 'Task ran successfully!'
+
+  if (lastRunStatus === 'failed' || lastRunError !== undefined) {
```
hoorayimhelping:
```ts
lastRunError !== undefined
```
is this this check we want here (rather than a more greedy falsy check)?  `lastRunError = false | null | '' | {}` will pass this check and make the app enter an error state. is that desirable?

http/notification_rule_test.go
```diff
@@ -147,7 +147,9 @@ func Test_newNotificationRuleResponses(t *testing.T) {
       ],
       "type": "slack",
       "updatedAt": "0001-01-01T00:00:00Z",
-      "status": "active"
+      "status": "active",
+			"latestCompleted": "0001-01-01T00:00:00Z",
```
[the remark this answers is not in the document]
Asalem1: @hoorayimhelping go won't allow for weird formatting. This may look strange but it's kosher in the actual files

http/notification_rule_test.go
```diff
@@ -287,7 +291,9 @@ func Test_newNotificationRuleResponse(t *testing.T) {
    }
  ],
  "type": "slack",
- "updatedAt": "0001-01-01T00:00:00Z"
+ "updatedAt": "0001-01-01T00:00:00Z",
+ "latestCompleted": "0001-01-01T00:00:00Z",
+	"latestScheduled": "0001-01-01T00:00:00Z"
```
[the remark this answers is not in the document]
Asalem1: same as above

ui/src/shared/components/lastRunTaskStatus/LastRunTaskStatus.tsx
Asalem1: @hoorayimhelping this was input by @alexpaxton. I'm not entirely sure on best practices but if it's the case that it should be input into the chronograf file I can go ahead and do that

http/notification_rule_test.go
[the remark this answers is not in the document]
hoorayimhelping: i know gfmt controls spacing, but those things can get weird sometimes, it's always good to confirm and not just trust the machines

ui/src/alerting/actions/checks.ts
Asalem1: Yes, when I had originally submitted the comment I hadn't had the chance to write up the issue. Here's the breaking change PR:

https://github.com/influxdata/influxdb/pull/16333
but it seems like the changes made to updateCheck weren’t updated here:
https://github.com/influxdata/influxdb/blob/master/ui/src/alerting/components/CheckCard.tsx#L17

and here's the issue that is intended on resolving it :

https://github.com/influxdata/influxdb/issues/16339

ui/src/alerting/components/CheckCard.tsx
Asalem1: It was a lint error - i imagine stemming from the way `clockface` handles metaData being passed dom nodes

http/notification_rule_test.go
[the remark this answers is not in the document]
Asalem1: for sure. It's all good

ui/src/shared/components/lastRunTaskStatus/LastRunTaskStatus.tsx
Asalem1: That's a great point and definitely something to consider. My thinking here was that since the `lastRunError` is an optional parameter (it can be undefined), and since the `lastRunError` should be a string if it is defined, the check for explicitly not undefined (meaning a string should exist) should be sufficient. Anything besides a string should trigger an error for us since our API would be behaving unexpectedly if it did return a more generally false check (like `false` `null`). In this case, an empty string being passed in would be more indicative of improper error handling, but not a reason to prevent the functionality

ui/src/shared/components/lastRunTaskStatus/LastRunTaskStatus.tsx
Asalem1: If a task run fails and we don't know about it, that's a failure on the API

PR #16346, opened 2019-12-27
feat(pkger): fixup flux query normalization in exporting of task

### LINKED ISSUE #16342: pkger tasks model fixing
there are issues with the tasks resource. The coupling of the flux string and the query make it very painful to do correctly. Not a fun one to work with. Need to either put tasks on the shelf for the time being or fix it up now with some ugly regex or something

jsteenb2:
Closes #16342 

Right now we have this duality of making the name/offset/every stuffs both inside the flux query and as siblings on the influxdb.Task type.  For pgker, it needs just the query and not the hand jammed `option task` stuffs, b/c of the brittleness of the current Task API. This makes the initial regex no longer greedy, it was capturing too much. The regex playground below is useful if you want to test drive some more stuffs at it:

regex playground for this fix: https://regex101.com/r/xV9wL4/30

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/pkger/clone_resource.go
+++ b/pkger/clone_resource.go
@@ -502,18 +502,14 @@ func ruleToResource(iRule influxdb.NotificationRule, endpointName, name string) 
 }
 
 // regex used to rip out the hard coded task option stuffs
-var taskFluxRegex = regexp.MustCompile(`option task = {(.|\n)*}`)
+var taskFluxRegex = regexp.MustCompile(`option task = {(.|\n)*?}`)
 
 func taskToResource(t influxdb.Task, name string) Resource {
 	if name == "" {
 		name = t.Name
 	}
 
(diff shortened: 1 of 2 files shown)


pkger/clone_resource.go
```diff
@@ -502,18 +502,14 @@ func ruleToResource(iRule influxdb.NotificationRule, endpointName, name string)
 }
 
 // regex used to rip out the hard coded task option stuffs
-var taskFluxRegex = regexp.MustCompile(`option task = {(.|\n)*}`)
+var taskFluxRegex = regexp.MustCompile(`option task = {(.|\n)*?}`)
```
jsteenb2: needed to make this not greedy

regex plagyround: https://regex101.com/r/xV9wL4/30

PR #16347, opened 2019-12-28
chore(inmem): nuke the duplicative inmem service implementation

jsteenb2:
I'm coming for you next bolt store 📻 

- [x] Rebased/mergeable
- [x] Tests pass


kelwang:
I saw source.go is still there. Are we still using that?

jsteenb2:
@kelwang good catch, and yah, there are a few files left in inmem that are necessary for idpe. I'll double back on the source.go specifically though. Pretty sure its needed for something in the idpe `cache` though

PR #16350, opened 2019-12-29
feat(checks): Add custom check type

### LINKED ISSUE #16075: Add Customize Check Query Button and Script Editor State to Alert Builder

ebb-tide:
Closes #16075

Adds customize check script editor behind feature flag.

User can begin creating a threshold or deadman check, and then switch to a check script editor mode to see the flux script that the BE would have created for the parameters in the alert builder.
User can hit `Execute check query and view statuses` button to view the statuses that the check would have written to monitoring bucket. monitor.write is overwritten by yield function to prevent user from writing to monitoring bucket outside of a check.
Currently user can not return to alert builder after customizing check.

As part of this PR:
Took alertBuilder state out from time machine.
Changed alertBuilder redux state to store all possible check properties in a flat way rather than the check as it is returned from the BE


notification/check/check.go
```diff
@@ -12,19 +12,21 @@ import (
...
+	OwnerID     influxdb.ID             `json:"ownerID,omitempty"`
+	OrgID       influxdb.ID             `json:"orgID,omitempty"`
+	Query       influxdb.DashboardQuery `json:"query"`
 
 	// Care should be taken to prevent TaskID from being exposed publicly.
 	TaskID influxdb.ID `json:"taskID,omitempty"`
-
-	Cron  string                 `json:"cron,omitempty"`
-	Every *notification.Duration `json:"every,omitempty"`
+	// } todo: separate these
```
ebb-tide: Take a look at these comments?

notification/check/custom.go
```diff
@@ -0,0 +1,345 @@
...
+	"github.com/influxdata/flux/parser"
+	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/notification"
+	"github.com/influxdata/influxdb/notification/flux"
+)
+
+var _ influxdb.Check = &Custom{}
+
+// AlertBuilder describes last state of alert builder not called check builder because that stutters
+type AlertBuilder struct {
```
ebb-tide: the UI is not currently using this consider removing it?

ui/src/alerting/utils/customCheck.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    const thresholdExpressions = thresholds.map(t => {
+      const fieldTag = builderConfig.tags.find(t => t.key === '_field')
+      const fieldSelection = get(fieldTag, 'values.[0]')
+
+      // "crit = (r) =>(r.fieldName"
+      const beginning = `${t.level.toLowerCase()} = (r) =>(r.${fieldSelection}`
+
+      if (t.type === 'range') {
+        if (t.within) {
```
hoorayimhelping: another stylistic case where an `else` isn't necessary, and can muddle the intent of the function.

if we treat the `if` as an exceptional case, we can return the normal case (or 'happy path'). seems like `t.within` is a bit rarer, so there should be an `if` statement to check for that, otherwise, just return the default case.

ui/src/alerting/utils/customCheck.ts
```diff
@@ -0,0 +1,153 @@
...
+          return `${beginning} > ${t.min}) and r.${fieldSelection} < ${t.max})`
+        } else {
+          return `${beginning} < ${t.min} and r.${fieldSelection} > ${t.max})`
+        }
+      } else {
```
hoorayimhelping: same thing here about `if/else` - what do you think of the style of removing `else`s?

ui/src/alerting/utils/customCheck.ts
```diff
@@ -0,0 +1,153 @@
...
+    const dataDefinition = [
+      dataFrom,
+      dataRange,
+      ...filterStatements,
+      aggregateFunction,
+    ]
+
+    const thresholdExpressions = thresholds.map(t => {
+      const fieldTag = builderConfig.tags.find(t => t.key === '_field')
+      const fieldSelection = get(fieldTag, 'values.[0]')
```
hoorayimhelping: i'm not familiar with the `values.[0]` syntax, what is that describing? won't that cause an error?

ui/src/alerting/utils/customCheck.ts
```diff
@@ -0,0 +1,153 @@
...
+      if (t.type === 'range') {
+        if (t.within) {
+          return `${beginning} > ${t.min}) and r.${fieldSelection} < ${t.max})`
+        } else {
+          return `${beginning} < ${t.min} and r.${fieldSelection} > ${t.max})`
+        }
+      } else {
+        const equality = t.type === 'greater' ? '>' : '<'
+
+        return `${beginning}${equality} ${t.value})`
```
hoorayimhelping: do you want a space between `${begnning}` and `${equality}`? if i'm reading this correctly it'll be something like `${t.level.toLowerCase()} = (r) =>(r.${fieldSelection}` (`fieldSelection` shouldn't have a space, since it's created from: `const fieldSelection = get(fieldTag, 'values.[0]')`), so it'll make something like

```ts
t.level.toLowerCase()} = (r) =>(r.foo> t.value)
```
?

ui/src/timeMachine/actions/index.ts
```diff
@@ -704,3 +665,26 @@ export const loadNewVEO = (dashboardID: string) => (
   }
   // no need to refresh variable values since there is no query in a new view
 }
+
+export const loadCustomQueryState = () => (
```
hoorayimhelping: is this specifically for checks? it might be good to name this function in a way that makes it clear that it's for checks specifically, if that's the case of course.

ui/src/timeMachine/actions/queries.ts
```diff
@@ -195,6 +199,61 @@ export const saveAndExecuteQueries = () => dispatch => {
...
+    checkQueryResult(result.csv)
+
+    const file = result.csv
+
+    dispatch(setQueryResults(RemoteDataState.Done, [file], duration, null))
+  } catch (e) {
+    if (e.name === 'CancellationError') {
+      return
+    }
+
```
hoorayimhelping: do we want to alert honeybadger here?

ui/src/timeMachine/components/TimeMachineAlerting.tsx
```diff
@@ -23,6 +25,9 @@ const TimeMachineAlerting: FunctionComponent = () => {
             justifyContent={JustifyContent.FlexEnd}
             margin={ComponentSize.Small}
           >
+            <FeatureFlag name="customCheckQuery">
```
hoorayimhelping: more of a personal question: why use the component over the function (`isFlagEnabled`) that the component calls? what abstraction is the component providing over just calling the function directly here?

ui/src/timeMachine/components/TimeMachineFluxEditor.tsx
```diff
@@ -99,11 +105,13 @@ class TimeMachineFluxEditor extends PureComponent<Props, State> {
           return (
             <>
               <div className="toolbar-tab-container">
-                <ToolbarTab
-                  onSetActive={this.hideFluxFunctions}
-                  name="Variables"
-                  active={!this.state.displayFluxFunctions}
-                />
+                {activeTab === 'customCheckQuery' ? null : (
```
hoorayimhelping:
```suggestion
                {activeTab !== 'customCheckQuery' && (
```

ui/tsconfig.json
```diff
@@ -36,8 +36,8 @@
     "jest.config.js",
     "webpack.tsconfig.json",
     "webpack.star.ts",
-    "src/**/*.test.ts",
-    "src/**/*.test.tsx",
+    // "src/**/*.test.ts",
+    // "src/**/*.test.tsx",
```
hoorayimhelping: remove these comments probably?

ebb-tide: thank you!

ui/src/timeMachine/actions/index.ts
ebb-tide: thanks!

ui/src/alerting/components/CheckEOHeader.tsx
```diff
@@ -115,11 +115,13 @@ const CheckEOHeader: FC<Props> = ({
...
-          draftQueries={draftQueries}
-          setActiveTab={setActiveTab}
-        />
+        {activeTab !== 'customCheckQuery' ? (
+          <CheckAlertingButton
+            activeTab={activeTab}
+            draftQueries={draftQueries}
+            setActiveTab={setActiveTab}
+          />
+        ) : null}
```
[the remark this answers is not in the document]
ebb-tide: 👍

ui/src/alerting/actions/checks.ts
```diff
@@ -147,32 +154,62 @@ export const saveCheckFromTimeMachine = () => async (
...
+      type,
+      status,
+      name,
       query: draftQueries[0],
       orgID,
-      labels: labels.map(l => l.id),
-    } as PostCheck
+    } as Check
+
+    if (check.type === 'threshold' || check.type === 'deadman') {
```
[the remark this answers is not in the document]
ebb-tide: 👍

ui/src/alerting/actions/checks.ts
```diff
@@ -101,14 +106,16 @@ export const getChecks = () => async (
 }
 
 export const getCheckForTimeMachine = (checkID: string) => async (
-  dispatch: Dispatch<TimeMachineAction | NotificationAction>,
+  dispatch: Dispatch<
```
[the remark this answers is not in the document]
ebb-tide: Since most actions don't take arguments- I don't really see why we should restrict which actions a thunk can dispatch? I'm really even fine with not typing dispatch at all. Maybe I'm missing something but I just don't see what there is to gain? It's not like it's a strong preference, but that is where I'm at. Open to argument.

ui/src/alerting/components/CheckEOHeader.tsx
```diff
@@ -98,10 +101,7 @@ const CheckEOHeader: FC<Props> = ({
   }
 
   const {singleField, singleAggregateFunc} = isDraftQueryAlertable(draftQueries)
-  const oneOrMoreThresholds =
-    check.type === 'threshold'
-      ? check.thresholds && !!check.thresholds.length
-      : false
+  const oneOrMoreThresholds = checkType === 'threshold' && !!thresholds.length
```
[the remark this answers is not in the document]
ebb-tide: I guess I was being a little more lax about this since thresholds is coming from redux, and not as a parameter from check, but we do overwrite alertBuilder.thresholds from a check coming from the API. so will add extra safety here

ui/src/alerting/components/builder/CheckConditionsCard.tsx
```diff
@@ -13,30 +13,27 @@ import ThresholdConditions from 'src/alerting/components/builder/ThresholdCondit
...
 
-  if (check.type === 'deadman') {
+  if (checkType === 'deadman') {
     cardTitle = 'Deadman'
-    conditionsComponent = <DeadmanConditions check={check} />
+    conditionsComponent = <DeadmanConditions />
   }
 
-  if (check.type === 'threshold') {
+  if (checkType === 'threshold') {
```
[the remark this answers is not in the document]
ebb-tide: There isn't now. since I set `activeTab` to `customCheckQuery` if check.type == "custom", but I think it might be open to breaking during a refactor.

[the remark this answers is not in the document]
ebb-tide: but the alternative is also not that great? It would be to check alertBuilder.checkType on line 38 of `ui/src/timeMachine/components/TimeMachine.tsx` I can't convince myself that this is significantly better. I will throw an error to honeybadger here.

ui/src/alerting/utils/customCheck.ts
```diff
@@ -0,0 +1,153 @@
...
+    type,
+    staleTime,
+    level,
+    timeSince,
+    thresholds,
+  }: AlertBuilderState
+): string {
+  const dataFrom = `data = from(bucket: \"${builderConfig.buckets[0]}\")`
+
+  const filterStatements = builderConfig.tags
```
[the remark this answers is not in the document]
ebb-tide: I think we also have a problem in that this code is duplicating behavior that exists on the backend, and either of these could drift? I think I'm relying on this feature being 'experimental', but also might be introducing patterns that aren't great in the long term.

ui/src/alerting/reducers/alertBuilder.ts
```diff
@@ -0,0 +1,219 @@
...
+        id,
+        type,
+        name,
+        checkStatus: RemoteDataState.Done,
+        query,
+      }
+
+      if (action.payload.check.type === 'custom') {
+        return newState
+      } else if (action.payload.check.type === 'threshold') {
```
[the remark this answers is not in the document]
ebb-tide: 👍

ui/src/alerting/reducers/alertBuilder.ts
```diff
@@ -0,0 +1,219 @@
...
+        const {
+          thresholds,
+          every,
+          offset,
+          tags,
+          statusMessageTemplate,
```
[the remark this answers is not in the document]
ebb-tide: oookayy :P

ui/src/alerting/utils/customCheck.ts
```diff
@@ -0,0 +1,153 @@
...
+  const checkTags = tags
+    ? tags
+        .filter(t => t.key && t.value)
+        .map(t => `${t.key}: \"${t.value}\"`)
+        .join(',')
+    : ''
+
+  const checkStatement = [
+    'check = {',
+    `  _check_id: \"${id || ''}\",`,
```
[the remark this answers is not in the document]
ebb-tide: turns out we don't! thanks!

ui/src/alerting/utils/customCheck.ts
```diff
@@ -0,0 +1,153 @@
...
+      dataRange,
+      ...filterStatements,
+      aggregateFunction,
+    ]
+
+    const thresholdExpressions = thresholds.map(t => {
+      const fieldTag = builderConfig.tags.find(t => t.key === '_field')
+      const fieldSelection = get(fieldTag, 'values.[0]')
+
+      // "crit = (r) =>(r.fieldName"
```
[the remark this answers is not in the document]
ebb-tide: I thought it was helpful! cut. :)

ui/src/alerting/utils/customCheck.ts
ebb-tide: added space. thanks!

ui/src/alerting/utils/customCheck.ts
ebb-tide: they are equivalent, but I can change it since it looks unfamiliar

ui/src/timeMachine/components/TimeMachineAlerting.tsx
ebb-tide: It reads better to me visually, but I do wonder if this makes it evaluate, even though it does not render. testing.

ui/src/alerting/utils/customCheck.ts
ebb-tide: it's so funny I used to be so strongly against `else`s but apparently left to my self I else all the time.

ui/src/timeMachine/actions/queries.ts
ebb-tide: nope! query errors are displayed to user and are helpful to understand that there is something legitimately wrong with users query and what it is. There is nothing for us to fix here.

ui/src/alerting/reducers/alertBuilder.test.ts
```diff
@@ -0,0 +1,264 @@
...
+      const newStatus = RemoteDataState.Error
+      const actual = alertBuilderReducer(
+        initialState(),
+        setAlertBuilderCheckStatus(newStatus)
+      )
+      expect(actual.checkStatus).toEqual(newStatus)
+    })
+  })
+
+  describe('setEvery', () => {
```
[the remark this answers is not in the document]
ebb-tide: I do think we need to address duration inputs. they are strings everywhere but clearly not all strings are valid, but currently the reducer accepts weird input, and relies on the API to reject bad every inputs. What would you have the reducer do for weird input?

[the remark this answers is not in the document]
hoorayimhelping: >and relies on the API to reject bad every inputs
I didn't realize we passed it through wholesale and let the api handle things. i don't think think there's much more to do there then

PR #16352, opened 2019-12-30
feat(storage): restore tool

### LINKED ISSUE #15605: storage 2.0: implement data restore from backup

### LINKED ISSUE #15606: storage 2.0: implement meta data restore from backup

tmgordeeva:
Adds an offline restore tool to replace existing data and metadata with a backup.

Closes #15605 
Closes #15606


cmd/influxd/restore/command.go
```diff
@@ -0,0 +1,239 @@
...
+		return err
+	}
+
+	fmt.Printf("Restored Bolt to %s from %s\n", flags.boltPath, backupBolt)
+	return nil
+}
+
+func restoreEngine() error {
+	dataDir := filepath.Join(flags.enginePath, "/data")
+	if err := os.Mkdir(dataDir, 0777); err != nil {
```
sebito91: Should this be `os.MkdirAll(dataDir, 0777)` in case we need to define parents?

cmd/influxd/restore/command.go
```diff
@@ -0,0 +1,239 @@
...
+
+	if err := removeTmpEngine(); err != nil {
+		return err
+	}
+
+	if err := os.Rename(flags.enginePath, tmpEnginePath()); err != nil {
+		return err
+	}
+
+	return os.Mkdir(flags.enginePath, 0777)
```
sebito91: I think this should be `os.MkdirAll(flags.enginePath, 0777)` to ensure the parents are made too. Since we check to see if `/var/lib/path/to/data/engine` exists in L145, we should make sure that this is fully created. Alternatively if we're just changing the perms we should use `Chmod`?

https://golang.org/pkg/os/#Chmod

cmd/influxd/restore/command.go
```diff
@@ -0,0 +1,239 @@
...
+				return fmt.Errorf("error opening TSM file: %v", err)
+			}
+			defer f.Close()
+
+			tsmPath := filepath.Join(dataDir, filepath.Base(path))
+			w, err := os.OpenFile(tsmPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0660)
+			if err != nil {
+				return err
+			}
+
```
sebito91: add `defer w.Close()` here

cmd/influxd/restore/command.go
jacobmarble: I would rather not worry about chmod'ing an existing dir here. I agree that `MkdirAll` makes sense.

cmd/influxd/restore/command.go
```diff
@@ -0,0 +1,239 @@
...
+	} else if err != nil {
+		return err
+	} else {
+		return os.RemoveAll(path)
+	}
+}
+
+func restoreBolt() error {
+	backupBolt := filepath.Join(flags.backupPath, bolt.DefaultFilename)
+	f, err := os.OpenFile(backupBolt, os.O_RDONLY, 0666)
```
jacobmarble: Use `os.Open` instead.

cmd/influxd/restore/command.go
```diff
@@ -0,0 +1,239 @@
...
+	if err != nil {
+		return fmt.Errorf("no bolt file in backup: %v", err)
+	}
+	defer f.Close()
+
+	w, err := os.OpenFile(flags.boltPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0666)
```
jacobmarble: (do not use `os.Open` here, as the file is opened for writing)

cmd/influxd/restore/command.go
```diff
@@ -0,0 +1,239 @@
...
+func restoreEngine() error {
+	dataDir := filepath.Join(flags.enginePath, "/data")
+	if err := os.Mkdir(dataDir, 0777); err != nil {
+		return err
+	}
+
+	count := 0
+	err := filepath.Walk(flags.backupPath, func(path string, info os.FileInfo, err error) error {
+		if strings.Contains(path, ".tsm") {
+			f, err := os.OpenFile(path, os.O_RDONLY, 0600)
```
jacobmarble: Use `os.Open`

tmgordeeva:
Tested by creating a backup, restoring and verifying:
- Engine starts and sample queries return equivalent results
- TSM files are the same
- Regenerated series files appear the same up to order

cmd/influxd/restore/command.go
```diff
@@ -0,0 +1,240 @@
...
+
+	if err := removeTmpEngine(); err != nil {
+		return err
+	}
+
+	if err := os.Rename(flags.enginePath, tmpEnginePath()); err != nil {
+		return err
+	}
+
+	return os.Mkdir(flags.enginePath, 0777)
```
sebito91: This one still should be changed to `MkdirAll`

jacobmarble:
Replaced by #16504

PR #16353, opened 2019-12-30
chore(ui): refactored tasks to use oats rather than client library

### LINKED ISSUE #16344: refactor(ui): migrate tasks to generated client
Part of #14482

Asalem1:
Closes #16344

### Problem

The client library is deprecated and no longer being supported, need to transition api routes based on the oats generatedRoutes

### Solution

Revised the client api calls that directly correlate to tasks to use the generatedRoutes. Added defaults for labels where applicable in order to refactor this in stages and prevent TS errors


ui/src/tasks/actions/index.ts
```diff
@@ -283,7 +308,7 @@ export const removeTaskLabelsAsync = (
 
 export const updateTaskStatus = (task: Task) => async dispatch => {
   try {
-    await client.tasks.updateStatus(task.id, task.status)
+    await patchTask({taskID: task.id, data: {status: task.status}})
```
Asalem1: Since this only updates the status (and does not pass along any information relating the `cron`, `every` specifics), I did not run a check to see whether or not the cron / every props needed to be updated

ui/src/tasks/actions/index.ts
```diff
@@ -296,7 +321,7 @@ export const updateTaskStatus = (task: Task) => async dispatch => {
 
 export const updateTaskName = (task: Task) => async dispatch => {
   try {
-    await client.tasks.update(task.id, task)
+    await patchTask({taskID: task.id, data: task})
```
Asalem1: Since this only updates the name (and does not pass along any information relating the `cron`, `every` specifics), I did not run a check to see whether or not the cron / every props needed to be updated

ui/src/tasks/actions/index.ts
```diff
@@ -415,7 +462,7 @@ export const updateScript = () => async (dispatch, getState: GetStateFunc) => {
       updatedTask.every = null
     }
 
-    await client.tasks.update(task.id, updatedTask)
+    await patchTask({taskID: task.id, data: updatedTask})
```
Asalem1: The UI is already explicitly setting `cron` and `every` to null values if the other value is set:

https://github.com/influxdata/influxdb/pull/16353/files#diff-3084610c031fd81910381dcc8507c9fcR457-R463

ui/src/tasks/actions/index.ts
```diff
@@ -252,12 +267,18 @@ export const getTasks = () => async (
   }
 }
 
-export const addTaskLabelsAsync = (taskID: string, labels: Label[]) => async (
+export const addTaskLabelAsync = (taskID: string, label: Label) => async (
```
Asalem1: renamed this to accurately reflect its functionality since only one label is ever passed

ui/src/tasks/actions/index.ts
```diff
@@ -266,13 +287,17 @@ export const addTaskLabelsAsync = (taskID: string, labels: Label[]) => async (
   }
 }
 
-export const removeTaskLabelsAsync = (
-  taskID: string,
-  labels: Label[]
-) => async (dispatch): Promise<void> => {
+export const removeTaskLabelAsync = (taskID: string, label: Label) => async (
```
Asalem1: renamed this to accurately reflect its functionality since only one label is ever passed

ui/src/types/logEvent.ts
```diff
@@ -0,0 +1,3 @@
+import {LogEvent as Log} from 'src/client'
+
+export interface LogEvent extends Log {}
```
ebb-tide: I feel like this might be equivalent to just one line:
`export {LogEvent} from 'src/client'` ?

Asalem1: That's a good point. I had originally written it like this in the event that other properties would need to be integrated but I'll go ahead and update it

ui/src/tasks/actions/index.ts
```diff
@@ -460,10 +514,15 @@ export const getRuns = (taskID: string) => async (dispatch): Promise<void> => {
...
-    ])
+    const resp = await getTasksRuns({taskID})
+
+    if (resp.status !== 200) {
+      throw new Error(resp.data.message)
+    }
+
+    let [runs] = await Promise.all([dispatch(selectTaskByID(taskID))])
+
+    runs = runs.concat(resp.data)
```
ebb-tide: selectTaskByID doesn't return anything, so runs is always undefined. and concat on undefined errors. 
try dispatch selectTaskByID above awaiting getTasksRuns. 
In this PR you are touching essential functionality of the UI, and nothing else matters if our data fetching is broken. Please test every place in the app that these functions are being used before moving on to refactoring the next function. The view task runs button on tasks does not work after this change. We don't have enough tests, and maybe that's the problem, but if we don't have tests, we can't move this fast.

ui/src/tasks/components/TaskRunsPage.tsx
```diff
@@ -10,9 +10,8 @@ import TaskRunsList from 'src/tasks/components/TaskRunsList'
 import PageTitleWithOrg from 'src/shared/components/PageTitleWithOrg'
 
 // Types
-import {AppState} from 'src/types'
-import {RemoteDataState} from 'src/types'
-import {Run as APIRun, Task} from '@influxdata/influx'
+import {AppState, RemoteDataState, Task} from 'src/types'
+import {Run as APIRun} from '@influxdata/influx'
```
ebb-tide: Would it be possible to eliminate relying on `@influxdata/influx` for Run types as well?

ui/src/tasks/actions/index.ts
```diff
@@ -19,13 +18,23 @@ import {
...
+  deleteTasksLabel,
+  getTask as apiGetTask,
+  getTasks as apiGetTasks,
+  getTasksRuns,
+  getTasksRunsLogs,
+  patchTask,
+  postTask,
+  postTasksLabel,
+  postTasksRun,
+} from 'src/client'
```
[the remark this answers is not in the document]
Asalem1: I hear what you're saying and had originally set out to use the same manner of importing and using the client routes, however @drdelambre brought up the point of being explicit and I felt it to be a good convention to follow:

https://github.com/influxdata/influxdb/pull/16325#discussion_r361754964

At the end of the day, this is a matter of preference, and while having an identifier like `api` before the function call name is a bit more indicative of the functionality we're performing, I don't think that importing the entire file to use one function from that file should serve that purpose. I am happy to alias the functions to their `apiFunctionName` though since that would help provide greater clarity on what the purpose of the functions are

ui/src/types/tasks.ts
```diff
@@ -1,7 +1,10 @@
-import {Task as TaskAPI, ITask} from '@influxdata/influx'
+import {Task as TaskAPI} from '@influxdata/influx'
+import {Task as ITask} from 'src/client'
+import {Label} from 'src/types'
 
 export const TaskStatus = TaskAPI.StatusEnum
```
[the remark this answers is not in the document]
Asalem1: Thanks for this, that file is a really beneficial resource.

ui/src/tasks/actions/index.ts
```diff
@@ -322,7 +347,19 @@ export const deleteTask = (task: Task) => async dispatch => {
 
 export const cloneTask = (task: Task, _) => async dispatch => {
   try {
-    await client.tasks.clone(task.id)
+    const resp = await apiGetTask({taskID: task.id})
+
+    if (resp.status !== 200 || resp.data.orgID === undefined) {
```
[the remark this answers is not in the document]
Asalem1: That's a great question. No idea anymore. Gonna remove it

ui/src/tasks/actions/index.ts
```diff
@@ -296,7 +321,7 @@ export const updateTaskStatus = (task: Task) => async dispatch => {
 
 export const updateTaskName = (task: Task) => async dispatch => {
```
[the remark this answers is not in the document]
Asalem1: Right, I totally agree! Especially since we are literally doing just that where it's being passed in:

https://github.com/influxdata/influxdb/blob/master/ui/src/tasks/components/TaskCard.tsx#L155-L158

Should I refactor it?

ui/src/tasks/actions/index.ts
Asalem1: I truthfully hadn't even noticed that. Thanks for pointing that out!

ui/src/tasks/components/TaskRunsPage.tsx
Asalem1: Definitely.

ui/src/tasks/actions/index.ts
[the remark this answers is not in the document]
ebb-tide: ok apiFunctionName it is then, yeah- opinions may differ :)

ui/src/templates/utils/index.ts
```diff
@@ -72,3 +73,22 @@ export const getLabelRelationships = (resource: {
...
+  ...l,
+  properties: {
+    ...l.properties,
+    // add default color hex if missing
+    color: (l.properties || {}).color || DEFAULT_LABEL_COLOR,
+    description: (l.properties || {}).description || '',
+  },
+})
+
+export const addDefaults = (task: Task): Task => {
```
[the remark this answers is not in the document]
ebb-tide: I totally agree that all types should be used and accessible from src/types.

ui/src/tasks/components/TaskCard.tsx
```diff
@@ -199,18 +202,18 @@ export class TaskCard extends PureComponent<Props & WithRouterProps> {
...
     }
     return false
   }
 
   private changeToggle = () => {
     const {task, onActivate} = this.props
-    if (task.status === TaskStatus.Active) {
-      task.status = TaskStatus.Inactive
+    if (task.status === 'active') {
+      task.status = 'inactive'
```
ebb-tide: ❤️

ui/src/tasks/components/TaskRunsList.tsx
```diff
@@ -8,7 +8,7 @@ import TaskRunsRow from 'src/tasks/components/TaskRunsRow'
 
 // Types
 import {Sort, ComponentSize} from '@influxdata/clockface'
-import {Run} from 'src/tasks/components/TaskRunsPage'
+import {Run} from 'src/types'
```
ebb-tide: ❤️

Issue #14514, opened 2019-07-31
Support the ability for Flux or queryd to discard output


stuartcarnie:
The primary use case is for tasks, which discards the output of queries. To reduces resource usage and waste, queryd should support the ability to discard the results. Note the following allocations graph, with significant churn as a result of discarding queryd output.

![image.png](https://images.zenhubusercontent.com/5d386a7a09d6d4240f1ac1f6/b81e10a4-8ef8-41ae-857d-0dc06c2707d3)

## New Result Encoder and Decoder

The goal is to encode a minimal response to the client.

* Implement a new result encoder and decoder as a new dialect
* Use the `Accept` header in the request to determine what type of response to send

aanthony1243:
Possible solution:  create a `/dev/null` type encoder that discards rows instead of producing large amounts of CSV.    However, the service must still send a response (perhaps a summary for logging purposes)

affo:
I assigned out team too for possible coordination around the Flux bits.

Would be cool to have a `trap` that does the opposite of `yield` and traps the result.

This is useful I think beyond tasks too. It makes it so one can produce side effects and ignore the results 👍 

Extremely good point

affo:
@aanthony1243 @stuartcarnie also note that it could be damn difficult to post-process the AST and add a `trap`, while adding an option is easy. Would it be better to add an option `discard_output` in package `experimental`?

Options are not well documented anyways, so users are not aware of some of them still.

gianarb:
I didn't know that option exists. This morning I started copy/pasting some code from the stdlib to bootstrap a function, from my understanding and after a conversation with @affo there is a good amount of discussion that needs to be done. 

I like the idea to use `option` because they look general an easy to inject.

docmerlin:
Sending a byte (maybe a carriage return) per table when that table is processed might be a useful way to prevent too-early timeouts.

affo:
Closed via https://github.com/influxdata/influxdb/pull/16356

PR #16356, opened 2019-12-30
feat(query): add dialect to discard output

affo:
This adds a dialect to discard the output in the response of a query.

When somebody specifies `Accept: /dev/null` in his request's headers, the query will return a plain text response containing `null`.

Issue: #14514


affo:
@stuartcarnie I really like your suggestion. I changed to what you proposed.

I set the response to `204 no content` as in the specification you sent.

PR #16367, opened 2019-12-31
chore(kv): fixup the checks service to use the index store

jsteenb2:
closes: #16370 

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/kv/store_base.go
+++ b/kv/store_base.go
@@ -19,34 +19,44 @@ type Entity struct {
 	Body  interface{}
 }
 
+// EncodeEntFn encodes the entity. This is used both for the key and vals in the store base.
 type EncodeEntFn func(ent Entity) ([]byte, string, error)
 
+// EncIDKey encodes an entity into a key that represents the encoded ID provided.
 func EncIDKey(ent Entity) ([]byte, string, error) {
 	id, err := ent.ID.Encode()
 	return id, "ID", err
 }
 
-func EncOrgNameKeyFn(caseSensitive bool) func(ent Entity) ([]byte, string, error) {
+// EncOrgIDNameKey encodes an organization id and name key.
+func EncOrgIDNameKey(caseSensitive bool) func(ent Entity) ([]byte, string, error) {
 	return func(ent Entity) ([]byte, string, error) {
 		key, err := indexByOrgNameKey(ent.OrgID, ent.Name, caseSensitive)
 		return key, "organization ID and name", err
 	}
 }
 
+// EncBodyJSON JSON encodes the entity body and returns the raw bytes and indicates
+// that it uses the entity body.
 func EncBodyJSON(ent Entity) ([]byte, string, error) {
 	v, err := json.Marshal(ent.Body)
 	return v, "entity body", err
 }
 
-type DecodeBucketEntFn func(key, val []byte) (keyRepeat []byte, decodedVal interface{}, err error)
+// DecodeBucketValFn decodes the raw []byte.
+type DecodeBucketValFn func(key, val []byte) (keyRepeat []byte, decodedVal interface{}, err error)
 
-func DecIndexEntFn(key, val []byte) ([]byte, interface{}, error) {
+// DecIndexID decodes the bucket val into an influxdb.ID.
+func DecIndexID(key, val []byte) ([]byte, interface{}, error) {
 	var i influxdb.ID
 	return key, i, i.Decode(val)
 }
 
-type DecodedValToEntFn func(k []byte, v interface{}) (Entity, error)
+// ConvertValToEntFn converts a key and decoded bucket value to an entity.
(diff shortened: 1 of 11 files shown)


kv/store_index.go
```diff
@@ -75,29 +74,97 @@ func (s *IndexStore) DeleteEnt(ctx context.Context, tx Tx, ent Entity) error {
...
-			entVal, err := s.EntStore.FindEnt(ctx, tx, ent)
-			if err != nil {
-				return errors.Wrap(err, "entity lookup")
+	entCaptureFn := opts.CaptureFn
+	kvStream, filterFn := s.indexFilterStream(ctx, tx, opts.FilterEntFn)
+	opts.FilterEntFn = filterFn
+	opts.CaptureFn = func(key []byte, indexVal interface{}) error {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
kelwang: This error needs to be wrapped.

kv/store_base.go
```diff
@@ -26,7 +26,7 @@ func EncIDKey(ent Entity) ([]byte, string, error) {
 	return id, "ID", err
 }
 
-func EncOrgNameKeyFn(caseSensitive bool) func(ent Entity) ([]byte, string, error) {
+func EncOrgNameKey(caseSensitive bool) func(ent Entity) ([]byte, string, error) {
```
kelwang: add comments for exported func

kv/store_base.go
```diff
@@ -40,7 +40,7 @@ func EncBodyJSON(ent Entity) ([]byte, string, error) {
 
 type DecodeBucketEntFn func(key, val []byte) (keyRepeat []byte, decodedVal interface{}, err error)
 
-func DecIndexEntFn(key, val []byte) ([]byte, interface{}, error) {
+func DecIndexEnt(key, val []byte) ([]byte, interface{}, error) {
```
kelwang: comments for exported func

kv/store_base.go
```diff
@@ -401,8 +407,12 @@ func (i *Iter) Next(ctx context.Context) (key []byte, val interface{}) {
 	case len(i.prefix) > 0:
 		i.seek(ctx)
 		nextFn = func() ([]byte, []byte) {
-			kv := <-i.seekChan
-			return kv.k, kv.v
+			select {
+			case <-ctx.Done():
+				return nil, nil
```
kelwang: not sure why does it return nil here

kv/store_base.go
```diff
@@ -495,6 +501,18 @@ func indexByOrgNameKey(orgID influxdb.ID, name string, caseSensitive bool) ([]by
 	}
 	k := make([]byte, influxdb.IDLength+len(name))
 	copy(k, orgIDEncoded)
+
+	if name == "" {
```
kelwang: I would do it before make k

kv/store_index.go
```diff
@@ -75,29 +74,97 @@ func (s *IndexStore) DeleteEnt(ctx context.Context, tx Tx, ent Entity) error {
...
+			if !isValid {
+				close(kvStream)
+			}
+		}()
+		ent, err := s.IndexStore.DecodeToEntFn(key, indexVal)
+		if err != nil {
+			return send(nil, nil, err)
+		}
+
+		entVal, err := s.EntStore.FindEnt(ctx, tx, ent)
```
kelwang: this should always be the id, right?

kv/store_index.go
jsteenb2: the errors get wrapped by the Find cmd in the storage base, extra wrapping here isn't warranted

kv/store_base.go
jsteenb2: I agree, I'm strugglign with naming convention here. The `EncOrgNameKey` key is a key of the org and name.

kv/store_base.go
jsteenb2: good question, the `nil, nil` is a signal to the `Iter.isNext` call, it tells it the loop should exit

kv/store_base.go
jsteenb2: @kelwang, the comment below here explains why this is here. It is to provide a partial key when warranted. Fairly often we lookup entities by `orgID` without a name, this provides that mechanism. When no name is present, it still reponds with an error, but also responds with the key filled with the `orgID`

kv/store_index.go
jsteenb2: I'm a little lost. Which part is suppose to be the `id`? The entity that you see really just the ID. The abstraction works against the `Entity` type though. What this is doing is a conversion between ID to an entity. This one ends up being that the index val is an ID, but that may not always be the case.

desa:
just tested this branch out and it seems like all of my variables are missing

PR #16375, opened 2019-12-31
feat(kv): add migrations to kv store

jsteenb2:
we currently have need to be able to make changes to the kv layer. There are issues at different times with out of sync indexes and other goodies. This is a POC for a stab at correcting that.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/kv/store_migration.go
+++ b/kv/store_migration.go
@@ -1,0 +1,168 @@
+package kv
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	ierrors "github.com/influxdata/influxdb/kit/errors"
+
+	"github.com/influxdata/influxdb/kit/tracing"
+
+	"github.com/influxdata/influxdb"
+)
+
+type Migration struct {
+	Name        string    `json:"name"`
+	Description string    `json:"description"`
+	RunAt       time.Time `json:"runAt"`
+
+	Up func(ctx context.Context, tx Tx, store *IndexStore) error `json:"-"`
+}
+
+type MigrationIndexStore struct {
+	migStore *StoreBase
+
+	*IndexStore
+}
+
+func NewMigrationIndexStore(indexStore *IndexStore) *MigrationIndexStore {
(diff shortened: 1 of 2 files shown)


GeorgeMac:
This is cool! Excited to see what can be done to improve the access patterns around the kv store.

This is a tangential comment as I just discovered all the precursor work you did to get here. I thought you might be interested in https://github.com/influxdata/influxdb/blob/master/kv/store.go#L99
This is something we added (@stuartcarnie came up with the idea) to encourage a better access pattern when using cursors. One which ensured we can perform more optimizations going forward on the particular bolt and etcd cursors.
Not totally relevant here, just food for thought. The idea is that you provide upfront a particular seek destination and a specified direction for a cursor which operates in a single direction. Once iteration stops you check if an error occurred and then close the cursor.

kv/store_migration.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+	if len(errs) > 0 {
+		return &influxdb.Error{Code: influxdb.EConflict, Msg: strings.Join(errs, "\n")}
+	}
+
+	for _, existing := range existingMigs {
+		delete(mMigs, existing.Name)
+	}
+
+	for _, mig := range mMigs {
```
GeorgeMac: I might be just misinterpreting here, but my assumption based on other migrations tools is that migrations are ordered. Is this a characteristic we want from our tool? If so using the map like this will be a problem.

My experience is with tools like https://github.com/golang-migrate/migrate which I believe maintain migration order.

jsteenb2: you're correct. This isn't ready for tracking histories of migrations just yet. We need to settle on a means to identify order still.  I'm fine with adding int IDs, on the migrations itself, but not sure how the team feels about that. Its odd b/c the migrations live inside an evolving codebase.

jsteenb2:
@GeorgeMac the forward cursor stuff would feed right into the new Find method. Enabling it in the new storeBase indexStore would be pretty simple to do. If you want make a card and I can fix that up super easy.

PR #16378, opened 2020-01-01
feat: report telegraf plugin usage metrics

### LINKED ISSUE #16286: feat: Log configured Telegraf plugins
__Proposal:__
For customers who generate their Telegraf configs from the cloud it would be very useful for us to get visibility to which Telegraf plugins they are using, so we can prioritize bug fixes/documentation for the more commonly used plugins.

__Desired behavior:__
Some mechanism (log, report etc.) to see which Telegraf plugins are being included in customers' configs, with a count for each plugin type. We don't need any config-specific data.

__Use case:__
Helps to give us visibility to more important plugins.

glinton:
Addresses #16286 

This adds plugin counts whenever a telegraf config is added/updated and exposes those metrics at `/metrics`

```
# HELP influxdb_telegraf_plugins_count Number of individual telegraf plugins configured
# TYPE influxdb_telegraf_plugins_count gauge
influxdb_telegraf_plugins_count{plugin="inputs.nginx"} 1
influxdb_telegraf_plugins_count{plugin="inputs.redis"} 1
influxdb_telegraf_plugins_count{plugin="outputs.influxdb_v2"} 1
```


kv/telegraf.go
```diff
@@ -195,6 +207,25 @@ func (s *Service) putTelegrafConfig(ctx context.Context, tx Tx, tc *influxdb.Tel
 	if err := bucket.Put(encodedID, v); err != nil {
 		return UnavailableTelegrafServiceError(err)
 	}
+
+	return s.putTelegrafConfigStats(encodedID, tx, tc)
+}
+
+func (s *Service) putTelegrafConfigStats(encodedID []byte, tx Tx, tc *influxdb.TelegrafConfig) error {
```
glinton: Add tests for this and deleteTelegrafConfigStats

jsteenb2: worht giving hte new `StoreBase` or `IndexStore` a run for the money here:
base: https://github.com/influxdata/influxdb/blob/master/kv/store_base.go
index store: https://github.com/influxdata/influxdb/blob/master/kv/store_index.go
example use: https://github.com/influxdata/influxdb/blob/master/kv/check.go

bolt/metrics.go
```diff
@@ -107,7 +116,24 @@ func (c *Client) Collect(ch chan<- prometheus.Metric) {
 		telegrafs = tx.Bucket(telegrafBucket).Stats().KeyN
 		tokens = tx.Bucket(authorizationBucket).Stats().KeyN
 		users = tx.Bucket(userBucket).Stats().KeyN
-		return nil
+		err := tx.Bucket(telegrafPluginsBucket).ForEach(func(k, v []byte) error {
```
jsteenb2: did you consider doing this stat counting incrementally as the telegraf configs come in? this setup won't extedn into cloud as is.... I'm pretty sure 🤔

glinton: yes, but this seemed to be the simplest way of tracking plugins configured in every config on the system. I suppose we could add more logic for when a config is updated vs newly added. on update, we'd diff the old config and apply it to the global configured plugins list..

do you know who's consuming these stats and how often?

jsteenb2: heh, I was goign to ask you that exact question lol. To me its a service concern anyhow, less so a store concern... not a hill for me to die on though

glinton: i'm completely open to other ideas regarding how to keep this aggregated list up to date. this seemed minimally invasive, though i have no hard backed estimates on how performant this would/not be on cloud.

jsteenb2: this wont' even run in cloud, this is bolt specific it looks like, OSS only

glinton: cloud specific stuff would be implemented in idpe though?? or what am i missing. a discussion for next year perhaps

desa: @glinton I think that you'd want to implement the abstraction as a part of the `kv` package. That'd be the best place to get these kinds of metrics.

bolt/metrics.go
```diff
@@ -75,10 +84,43 @@ func (c *Client) Describe(ch chan<- *prometheus.Desc) {
...
+		tick:   make(chan struct{}, 1),
+		timeCh: time.Tick(time.Minute * 59),
+	}
+
+	ticker.tick <- struct{}{}
+
+	go func() {
+		for {
+			select {
+			case <-ticker.timeCh:
```
jsteenb2: why not just use `for t := range ticker.Timech`?  Curious why this is here in bolt and not kv? Are these not intended for Cloud?

glinton: This is the bolt implementation (`bolt/*.go`) as cloud reports metrics differently. The important "shared" logic is in `kv/telegraf.go`. A cloud PR will be forthcoming that utilizes the data stored from that shared logic and reports it as well.

PR #16385, opened 2020-01-02
fix(durations): Add duration input that checks input validity as duration

ebb-tide:
Added duration input component to the check every field.
It's nice in that it gives user examples to choose from without limiting options available to them. 

The duration input does not update the value in the field in redux if it is not valid, but does update local state so that the invalid input is displayed to user in an error state.

### DIFF
--- a/ui/src/alerting/utils/checkValidate.ts
+++ b/ui/src/alerting/utils/checkValidate.ts
@@ -1,0 +1,16 @@
+import {Threshold} from 'src/types'
+import {isNaN} from 'lodash'
+
+export const checkThresholdsValid = (thresholds: Threshold[]) => {
+  thresholds.forEach(t => {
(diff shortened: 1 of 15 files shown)


ui/src/alerting/utils/checkValidate.ts
```diff
@@ -0,0 +1,16 @@
+import {Threshold} from 'src/types'
+import {isNaN} from 'lodash'
```
TCL735: There is a very subtle difference between lodash's isNaN and the global window.isNaN. https://lodash.com/docs/4.17.15#isNaN

lodash isNaN is strictly checking if the value is both a number and equal to NaN, whereas the global window.isNaN will check if the value is undefined.

Do we want to check for undefined here as well? If so, consider using the global window.isNaN

ebb-tide: thanks!

PR #16391, opened 2020-01-02
chore(ui): refactored variables client API to oats generated types and API

### LINKED ISSUE #16362: refactor(ui): migrate variables to generated client
Part of #14482

Asalem1:
Closes #16362

### Problem

Current API and types are dependent upon deprecated client API

### Solution

Refactored the variables to utilize the new oats generated client API


drdelambre:
dope

ui/src/dataLoaders/actions/dataLoaders.ts
```diff
@@ -33,6 +33,7 @@ import {
   BundleName,
   ConfigurationState,
 } from 'src/types/dataLoaders'
+import {ILabel} from '@influxdata/influx'
```
glinton: What about this one?

ui/src/templates/components/TemplatesPage.tsx
```diff
@@ -35,7 +35,7 @@ interface StaticTemplate {
 
 const staticTemplates: StaticTemplate[] = _.map(statics, (template, name) => ({
   name,
-  template,
+  template: template as TemplateSummary,
```
Asalem1: Not sure how this made it onto master. This PR shouldn't be effecting the template type, issue likely stems from an earlier PR on master that didn't set off the linter and should be addressed when Templates are refactored

ui/src/types/templates.ts
```diff
@@ -171,5 +171,5 @@ export interface VariableTemplate extends TemplateBase {
 export type Template = TaskTemplate | DashboardTemplate | VariableTemplate
 
 export interface TemplateSummary extends DocumentListEntry {
-  labels: Label[]
+  labels: ILabel[]
```
Asalem1: Same as before re: Template types

ui/src/dataLoaders/actions/dataLoaders.ts
Asalem1: @glinton was in the process of tossing comments on the PR:

https://github.com/influxdata/influxdb/pull/16391#discussion_r362660986

ui/src/types/labels.ts
```diff
@@ -1,7 +1,7 @@
 import {Label as GenLabel} from 'src/client'
 
 export interface Label extends GenLabel {
-  properties: LabelProperties
+  properties?: LabelProperties
```
ebb-tide: I'm confused about this change. The point of adding label defaults is that we are taking something of `GenLabel` type which has optional and any properties, and adding default properties to ensure they are there and then it becomes a Label type with required properties type.

ui/src/types/variables.ts
```diff
@@ -0,0 +1,7 @@
+export {VariableProperties} from 'src/client'
+import {Variable as GenVariable} from 'src/client'
+import {Label} from 'src/types'
+
+export interface Variable extends GenVariable {
+  labels?: Label[]
```
ebb-tide: I think we are ensuring that labels exists as a param even though it may be empty by adding defaults?

ebb-tide:
Adding labels to tasks doesn't work after this PR.

ui/src/variables/components/VariablesTab.tsx
```diff
@@ -153,7 +152,7 @@ class VariablesTab extends PureComponent<Props, State> {
     router.push(`/orgs/${orgID}/settings/variables/new`)
   }
 
-  private handleUpdateVariable = (variable: Partial<Variable>): void => {
+  private handleUpdateVariable = (variable: Variable): void => {
```
ebb-tide: So, it would not normally work to stronger type something that was more loosely typed- but it worked because this function is not being used anymore! Could you check that that's right and remove all of this if so?

ui/src/templates/api/index.ts
```diff
@@ -229,23 +235,38 @@ const createVariablesFromTemplate = async (
...
+    variables,
     variablesIncluded
   ).map(v => ({...v.attributes, orgID}))
 
-  const createdVariables = await client.variables.createAll(variablesToCreate)
+  const pendingVariables = variablesToCreate.map(vars =>
+    apiPostVariable({data: vars})
+      .then(res => res.data)
+      .catch(e => {
+        throw new Error(e.message)
```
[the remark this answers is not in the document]
Asalem1: That's a really good point. let me try to resolve this discrepancy some other way

ui/src/variables/components/VariablesTab.tsx
Asalem1: Very astute observation!

ui/src/types/labels.ts
Asalem1: That's a good point. I think it had originally been set to a required property to match with some of the expectations of the not-yet-refactored API functions, but removing it and defaulting to `export {Label} from 'src/client` seems to work fine now

ui/src/types/variables.ts
Asalem1: But that doesn't necessarily mean it'll be guaranteed to return from the BE with labels

ebb-tide: from BE -> add defaults function -> labels are guaranteed and Variable describes the type after defaults have been added.

glinton:
> Adding labels to tasks doesn't work after this PR.

@ebb-tide @Asalem1 Is there a new issue (or addition to a pr) to add a test that adds labels to tasks so that doesn't happen again?

Asalem1:
> > Adding labels to tasks doesn't work after this PR.
> 
> @ebb-tide @Asalem1 Is there a new issue (or addition to a pr) to add a test that adds labels to tasks so that doesn't happen again?

@glinton this PR was reverted since there was another review in the pipes when it had been submitted. The issue @ebb-tide brought up about Task Labels not working was addressed in the following PR and stemmed from the fact that the docs didn't align with the correct status code that was being returned:

https://github.com/influxdata/influxdb/pull/16400#pullrequestreview-338157450

PR #16397, opened 2020-01-03
fix(pkger): don't swallow errors when failing to get organization

### LINKED ISSUE #16387: pkgr: auth errors aren't shown to the user
if your token is invalid, there is no error shown to the user.

image

jademcgough:
Closes #16387


cmd/influx/pkg.go
```diff
@@ -129,7 +129,7 @@ func (b *cmdPkgBuilder) pkgApplyRunEFn() func(*cobra.Command, []string) error {
 
 		influxOrgID, err := b.org.getID(orgSVC)
 		if err != nil {
-			return nil
+			return err
```
jsteenb2: 🤦‍♂ how did I miss this in PR review 😳  lol, nice find @jademcgough 👍

PR #16400, opened 2020-01-03
chore(ui): refactored variables client API to oats generated types and API 

### LINKED ISSUE #16362: refactor(ui): migrate variables to generated client

Asalem1:
Closes #16362

Problem
Current API and types are dependent upon deprecated client API

Solution
Refactored the variables to utilize the new oats generated client API


ui/src/tasks/actions/index.ts
```diff
@@ -294,7 +301,10 @@ export const removeTaskLabelAsync = (taskID: string, label: Label) => async (
   dispatch
 ): Promise<void> => {
   try {
-    await apiDeleteTasksLabel({taskID, labelID: label.id})
+    const deleteResp = await apiDeleteTasksLabel({taskID, labelID: label.id})
```
ebb-tide: nit - sometimes the response is called `resp` and sometimes it is called `postResp`- I would prefer uniformity and  `resp`

Asalem1: @ebb-tide since we're already declaring a `resp` further down, I had to assign a different name:

https://github.com/influxdata/influxdb/pull/16400/files/55bd264cd5d7fb5b9e2243b21cf3adfc9d751510#diff-3084610c031fd81910381dcc8507c9fcR308

ui/src/templates/api/index.ts
```diff
@@ -229,23 +235,44 @@ const createVariablesFromTemplate = async (
...
 
-  const createdVariables = await client.variables.createAll(variablesToCreate)
+  const pendingVariables = variablesToCreate.map(vars =>
+    apiPostVariable({data: vars})
+  )
+
+  const resolvedVariables = await Promise.all(pendingVariables)
+  if (
+    resolvedVariables.length > 0 &&
+    resolvedVariables.every(r => r.status !== 201)
```
ebb-tide: I think instead of every this should be if any one of them is not 201

http/swagger.yml
```diff
@@ -4805,7 +4805,7 @@ paths:
             schema:
               $ref: "#/components/schemas/LabelMapping"
       responses:
-        '200':
+        '201':
```
Asalem1: This was updated to reflect what the code is actually returning:

https://github.com/influxdata/influxdb/blob/master/http/task_service.go#L149

Which, when successful, returns:

https://github.com/influxdata/influxdb/blob/master/http/label_service.go#L406

Which returns a `201`

ui/src/tasks/actions/index.ts
ebb-tide: ah ok.

ui/src/types/labels.ts
```diff
@@ -1,8 +1,4 @@
-import {Label as GenLabel} from 'src/client'
-
-export interface Label extends GenLabel {
-  properties: LabelProperties
-}
+export {Label} from 'src/client'
```
ebb-tide: I don't understand this.  should be as it was before. 

The client Label doesn't have properties, we are adding required properties. of type LabelProperties.

ui/src/types/variables.ts
```diff
@@ -0,0 +1,7 @@
+export {VariableProperties} from 'src/client'
+import {Variable as GenVariable} from 'src/client'
+import {Label} from 'src/types'
+
+export interface Variable extends GenVariable {
+  labels?: Label[]
```
ebb-tide: labels need to be required.

ui/src/variables/actions/index.ts
```diff
@@ -179,18 +204,24 @@ export const createVariable = (
...
+    }
+
+    const createdVar = addVariableDefaults(resp.data)
+
+    dispatch(setVariable(createdVar.id, RemoteDataState.Done, createdVar))
     dispatch(notify(createVariableSuccess(variable.name)))
   } catch (e) {
     console.error(e)
-    dispatch(notify(createVariableFailed(e.response.data.message)))
+    dispatch(notify(createVariableFailed(e.message)))
```
ebb-tide: make sure you have tested this please

ui/src/types/labels.ts
Asalem1: The old client library did have LabelProperties that were set differently for the `ILabel`:

https://github.com/influxdata/influxdb-client-js/blob/master/src/types/labels.ts#L13

This had originally been set to ensure that we had type stability. However, now that the labels have been refactored it is no longer necessary

ui/src/types/variables.ts
Asalem1: Why? They aren't required in the swagger file:

https://github.com/influxdata/influxdb/blob/master/http/swagger.yml#L8817

ui/src/variables/actions/index.ts
Asalem1: This actually returned an error when it was set as `e.response.data.message` since the error we're throwing is `e.message`

ui/src/types/variables.ts
Asalem1: The reason for setting labels here (and basically having them do the same thing) is to ensure that the type is declared to the same type. That is, since the default for `Variable` labels are the `Labels` generated in the oats file, they are not the same as the ones that are assigned to the `Labels` in the `src/types` file

Asalem1: But now that we've actually stopped customizing the Label type and are simply exporting it, it seems like this is pretty unnecessary. I'm gonna revert it to the original `export {Variable} from src/client` if the linter returns no errors

kelwang:
API part looks good

ui/src/tasks/actions/index.ts
hoorayimhelping: @Asalem1 @ebb-tide is this an indicator that this operation should be in it's own try/catch block? i think it should since these seem to be discrete actions - `apiGetTask`  doesn't seem to use the response from `apiDeleteTasksLabel`, so they really don't need to be grouped together. the main reason I can think of is to prevent duplication of the `catch` block, but I think duplication is less detrimental than muddled intent.

if they must stay together, i think `resp` should be changed to `getTaskResp` to be in line with `deleteResp`. calling one `deleteResp` and one `resp` gives a bit more weight to `resp` and I'm not sure if that's the intent here.

ui/src/templates/api/index.ts
```diff
@@ -298,39 +325,59 @@ const addTaskLabelsFromTemplate = async (
...
+        await apiPostTasksLabel({taskID: task.id, data: {labelID}})
+    )
+    const resolved = await Promise.all(pending)
+    if (resolved.length > 0 && resolved.every(r => r.status !== 201)) {
+      throw new Error('An error occurred adding task labels from the templates')
+    }
+  } catch (e) {
+    console.error(e)
+  }
 }
```
ebb-tide:
```
const addTaskLabelsFromTemplate = async (
  template: TaskTemplate,
  labelMap: LabelMap,
  task: Task
) => {
  try {
    const relationships = getLabelRelationships(template.content.data)
    const labelIDs = relationships.map(l => labelMap[l.id] || '')
    const pending = labelIDs.map(labelID =>
      apiPostTasksLabel({taskID: task.id, data: {labelID}})
    )
    const resolved = await Promise.all(pending)
    if (resolved.length > 0 && resolved.some(r => r.status !== 201)) {
      throw new Error('An error occurred adding task labels from the templates')
    }
  } catch (e) {
    console.error(e)
  }
}
```

ui/src/templates/api/index.ts
ebb-tide:
```suggestion
    resolvedVariables.some(r => r.status !== 201)
```

ui/src/templates/api/index.ts
```diff
@@ -229,23 +235,44 @@ const createVariablesFromTemplate = async (
...
 
   const addLabelsToVars = variablesIncluded.map(async includedVar => {
     const variable = allVars.find(v => v.name === includedVar.attributes.name)
     const labelRelationships = getLabelRelationships(includedVar)
     const labelIDs = labelRelationships.map(l => labelMap[l.id] || '')
-
-    await client.variables.addLabels(variable.id, labelIDs)
+    const pending = labelIDs.map(async labelID => {
+      await apiPostVariablesLabel({variableID: variable.id, data: {labelID}})
+    })
```
ebb-tide:
```
    const pending = labelIDs.map(labelID => {
      apiPostVariablesLabel({variableID: variable.id, data: {labelID}})
    })
```

ebb-tide: then pending is an array of promises that you do await  Promise.all on.. 👍

ebb-tide: There are a few more examples of this around :)

ui/src/templates/api/index.ts
ebb-tide: there are a few more examples of this.

ui/src/variables/actions/index.ts
```diff
@@ -57,6 +65,13 @@ export type EditorAction =
   | ReturnType<typeof updateMap>
   | ReturnType<typeof updateConstant>
 
+export const addVariableDefaults = (task: IVariable): Variable => {
+  return {
+    ...task,
```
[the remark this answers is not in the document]
Asalem1: 🤦‍♂

ui/src/variables/utils/hydrateVars.test.ts
```diff
@@ -190,7 +190,7 @@ describe('hydrate vars', () => {
           k: 'v',
         },
       },
-    }
+    } as Variable
```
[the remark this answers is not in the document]
Asalem1: Great to know, thanks!

ui/src/templates/api/index.ts
Asalem1: Totally, I think I was aiming for `every === 201`, good catch!

PR #16392, opened 2020-01-02
test: fix flaky explorer tests

hoorayimhelping:
Closes #16330

CodeMirror wraps the code textarea so it can apply eye candy to the text. Our tests were targeting the wrapped textarea and using the `force: true` flag to force cypress to type into it. This change instead targets the CodeMirror div that wraps the text area and doesn't force the typing. This seems to work as cypress waits for the text area to be typeable, which CodeMirror seems to handle.

- [x] [Well-formatted commit messages](https://www.conventionalcommits.org/en/v1.0.0-beta.3/)
- [x] Rebased/mergeable
- [x] Tests pass

PR #16404, opened 2020-01-03
fix(ui): removed buildActiveQuery func to prevent queries from being reset when timeranges are selected.

### LINKED ISSUE #16330: Investigate Flaky Exlporer Tests
Query Builder Grouping (#16226) either introduced or exposed issues in the Explorer tests.

The issues were centered around the CodeMirror editor not loading properly or speedily enough, and the incorrect input being typed into the text area.

https://github.com/influxdata/influxdb/blob/master/ui/cypress/e2e/explorer.test.ts#L450

An errored out run (link may expire):
https://circleci.com/gh/influxdata/influxdb/78933?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

DataExplorer -- editing mode switching -- can switch to and from script editor mode (failed)

### LINKED ISSUE #16369: Query disappears in data explorer when changing time range
disapearing-query

To reproduce

1. switch to data explorer
2. enter the query
 ```
from(bucket: "_monitoring")
  |> filter(fn: (r) => r._measurement == "statuses" and r._field == "_message")
  |> keep(columns: ["_time", "_value", "_check_id", "_check_name", "_level"])
  |> window(every: 1s, timeColumn: "_time", startColumn: "_start", stopColumn: "_stop")
  |> group(columns: ["_start", "_stop"])
  |> filter(fn: (r) => r["_check_id"] == "undefined")
  |> rename(columns: {"_time": "time",
                      "_value": "message",
                      "_check_id": "checkID",
                      "_check_name": "checkName",
                      "_level": "level"})
```
3. change the time range

### PR DESCRIPTION
**Pull request by Asalem1**
Closes #16369

### Problem

Calling `buildActiveQuery` after `SET_BUILDER_TAG_VALUES` was resetting the query whenever the timerange was being set

### Solution

The rationale for calling the `buildActiveQuery` on `SET_BUILDER_TAG_VALUES` was to ensure that:

> it's to build the query when grouping and selecting fields

This seems to already be occurring whenever a query is drafted and is unnecessary. Since this change had originally broken an e2e explorer test, removing this line passes the test.

One key metric to ensure that this is in fact a stable solution would be to:

>  look at switching between filtering and grouping in the query builder

### DIFF
--- a/ui/cypress/e2e/explorer.test.ts
+++ b/ui/cypress/e2e/explorer.test.ts
@@ -349,7 +349,7 @@ describe('DataExplorer', () => {
     })
   })
 
-  describe('raw script editing', () => {
+  describe.skip('raw script editing', () => {
     beforeEach(() => {
       cy.getByTestID('switch-to-script-editor').click()
     })
--- a/ui/cypress/e2e/explorer.test.ts
+++ b/ui/cypress/e2e/explorer.test.ts
@@ -349,7 +349,7 @@ describe('DataExplorer', () => {
     })
   })
 
-  describe.skip('raw script editing', () => {
+  describe('raw script editing', () => {
     beforeEach(() => {
       cy.getByTestID('switch-to-script-editor').click()
     })
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
 1. [16374](https://github.com/influxdata/influxdb/pull/16374): Update influx CLI, only show "see help" message, instead of the whole usage.
 1. [16380](https://github.com/influxdata/influxdb/pull/16380): Fix notification tag matching rules and enable tests to verify
 1. [16376](https://github.com/influxdata/influxdb/pull/16376): Extend the y-axis when stacked graph is selected
+1. [16404](https://github.com/influxdata/influxdb/pull/16404): Fixed query reset bug that was resetting query in script editor whenever dates were changed
 
 ### UI Improvements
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: ui/cypress/e2e/explorer.test.ts [hunk 1]
```diff
@@ -349,8 +349,7 @@ describe('DataExplorer', () => {
     })
   })
 
-  // todo: investigate flakiness of this test: https://github.com/influxdata/influxdb/issues/16330
-  describe.skip('raw script editing', () => {
+  describe('raw script editing', () => {
```
hoorayimhelping: i'd prefer this to stay skipped, i have another pr that is handling this: https://github.com/influxdata/influxdb/pull/16392.

Asalem1: @hoorayimhelping I did run them a few times but I have no problem skipping them here in favor of your PR

hoorayimhelping: @Asalem1 I think they're okay to not be skipped now, I have what I think is the fix in master. I was more curious about the root cause of this issue - I'm having a hard time figuring out why removing `buildActiveQuery` had an effect on these tests and I was wondering if you found anything out in your digging

ui/cypress/e2e/explorer.test.ts
```diff
@@ -349,7 +349,7 @@ describe('DataExplorer', () => {
     })
   })
 
-  describe('raw script editing', () => {
+  describe.skip('raw script editing', () => {
```
hoorayimhelping:
```suggestion
  describe('raw script editing', () => {
```

hoorayimhelping:
@Asalem1 this is good to merge after removing the `skip`

ui/cypress/e2e/explorer.test.ts
Asalem1: @hoorayimhelping aw, I see. Truthfully, I'm not sure I fully comprehend why the test was failing when that line was added, and why it's no longer failing. I imagine that there are unintended consequences to the functionality of the utility function that may have been effecting the query builder / script editor that we are currently unaware of.

PR #16411, opened 2020-01-06
fix(tsdb): incorrect error type in type switch

foobar:
With 0ccea81729, it should be checking `PartialWriteError` instead of `*PartialWriteError`

### DIFF
--- a/tsdb/shard.go
+++ b/tsdb/shard.go
@@ -591,10 +591,7 @@ func (s *Shard) validateSeriesAndFields(points []models.Point) ([]models.Point, 
 	var droppedKeys [][]byte
 	if err := engine.CreateSeriesListIfNotExists(keys, names, tagsSlice); err != nil {
 		switch err := err.(type) {
-		// TODO(jmw): why is this a *PartialWriteError when everything else is not a pointer?
-		// Maybe we can just change it to be consistent if we change it also in all
-		// the places that construct it.
-		case *PartialWriteError:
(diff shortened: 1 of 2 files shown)


foobar:
@e-dard  can you add 1.x label?

tsdb/shard.go
```diff
@@ -591,10 +591,7 @@ func (s *Shard) validateSeriesAndFields(points []models.Point) ([]models.Point,
 	var droppedKeys [][]byte
 	if err := engine.CreateSeriesListIfNotExists(keys, names, tagsSlice); err != nil {
 		switch err := err.(type) {
-		// TODO(jmw): why is this a *PartialWriteError when everything else is not a pointer?
-		// Maybe we can just change it to be consistent if we change it also in all
-		// the places that construct it.
-		case *PartialWriteError:
+		case PartialWriteError:
```
e-dard: Is there a unit test somewhere that can cover this?

foobar: Seems not for the error type. I will add a test case in index/inmem since this error is only populated from inmem index

foobar:
@e-dard can you take a look at the updated version?

dgnorton:
@foobar thanks. Would you mind cherry-picking this to the 1.8 branch also?

dgnorton:
Issue for forward porting this fix to 1.8:  https://github.com/influxdata/influxdb/issues/17020

PR #16417, opened 2020-01-06
build(flux): update Flux to v0.58.0

jsternberg:


query/stdlib/testing/testing.go
```diff
@@ -62,8 +62,8 @@ var FluxEndToEndSkipList = map[string]map[string]string{
 		"keys":               "group key mismatch",
 
 		// failed to read meta data errors: the CSV encoding is incomplete probably due to data schema errors.  needs more detailed investigation to find root cause of error
-		"filter_by_regex":             "failed to read metadata",
-		"filter_by_tags":              "failed to read metadata",
+		// "filter_by_regex":             "failed to read metadata",
```
wolffcm: Any reason not to just delete these lines?

query/promql/internal/promqltests/go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/glycerine/goconvey v0.0.0-20190410193231-58a59202ab31 // indirect
 	github.com/gogo/protobuf v1.3.0 // indirect
 	github.com/google/go-cmp v0.3.1
-	github.com/influxdata/flux v0.57.0
+	github.com/influxdata/flux v0.57.1-0.20191219182109-2fa9b34f588f
```
wolffcm: Shouldn't the version here be `v0.58.0` same as for the main module?

query/stdlib/testing/testing.go
jsternberg: These probably should have been deleted. I will come back to this in the future, but I don't want to make additional modifications beyond updating the version in this PR.

PR #16418, opened 2020-01-06
docs: add developer guide

### LINKED ISSUE #16407: Add Development Guide
There is currently no real guide to setting up Influx locally.

Here is some feedback from a user in the community slack channel:

---

Sure,  I think you need a doc that describes setting up to build Influx and have it run in debug mode (I couldn't see one).

I interpreted most of the environment from the docker file for Flux, and reading the make file / fixing errors.   (Rust for example).

The main confusion starts with the online doc for the Alpha version, which states that it is all running on Port 9999,  which works just fine for that binary. Chronograf is there etc.  The initial setup works fine.   But things fell apart after cloneing the repo and trying a build.

Parts that confused me and req support were:

* Needs Yarn, and the Yarn Install / Yarn Start for the JS  {Shouldn't this be in the Makefile ?  or at least check its up and warn ?} 
* The Port for UI being different than the 9999 default for V2 - which works fine in downloaded binary
* The 500-Internal Server error, with no corresponding error on server log when I hit the 9999 endpoint. (The API is there, but not chronograf)   ZERO info as to problem.
* The 404 Errror on same endpoint after yarn was running.  - Again, no info as to why
* Tracing the handlers in the code is 'challenging'  I found the chronograf handler eventually, but even after putting a breakpoint on every instance of http.StatusInternalServerError  still couldn't see where it was failing to load Chronograf.  

NOTE: The server itself appears to be running fine, my task is running,  telegraf is delivering, it was just the UI that was borked.

[here is the slack convo.](https://influxcommunity.slack.com/archives/CH8RV8PK5/p1577997979103700?thread_ts=1577987613.093500&cid=CH8RV8PK5)

hoorayimhelping:
Closes #16407

* Update docs to reference Go 1.13
* Create developer guide

[View the rendered markdown file here](https://github.com/influxdata/influxdb/blob/13805208c18a76c99a868e24e1d608f046f9f27d/DEVELOPMENT.MD)

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 1. [16348](https://github.com/influxdata/influxdb/pull/16348): Drop legacy bolt service implementation in favor of kv service with bolt dependency
 1. [16014](https://github.com/influxdata/influxdb/pull/16014): While creating check, also display notification rules that would match check based on tag rules
 1. [16389](https://github.com/influxdata/influxdb/pull/16389): Increase default bucket retention period to 30 days
+1. [16418](https://github.com/influxdata/influxdb/pull/16418): Add Developer Documentation
 
 ### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


TCL735:
Excellent guide on getting started.

PR #16427, opened 2020-01-07
fix: show queries built in advanced mode when switching tabs

### LINKED ISSUE #16169: QueryBuilder: valid queries with invalid BuilderConfigs cause wonky tab behavior
when you have an editor, with multiple tabs for queries, deleting a tab that you are not currently viewing deletes all the queries.

Kapture 2019-12-09 at 9 57 26

hoorayimhelping:
Closes #16169

If a query isn't a valid QueryBuilder query, but it still has text (in other words, it was built using the script editor, or by the system), use the text, rather than blanking the text out.

- [x] Rebased/mergeable
- [x] Tests pass


ui/src/telegrafs/reducers/index.ts
```diff
@@ -74,4 +74,4 @@ export const telegrafsReducer = (
         return
       }
     }
-  })
\ No newline at end of file
+  })
```
hoorayimhelping: `prettier:fix`

PR #16433, opened 2020-01-07
POC for a working service definition

### LINKED ISSUE #16316: Service Definition for Notification Endpoints

jsteenb2:
Closes #16316 

The card itself outlines the intentions behind this work.

- [ ] Rebased/mergeable
- [ ] Tests pass


notification_endpoint.go
```diff
@@ -4,13 +4,43 @@ import (
...
 var (
 	// ErrInvalidNotificationEndpointType denotes that the provided NotificationEndpoint is not a valid type
 	ErrInvalidNotificationEndpointType = errors.New("unknown notification endpoint type")
 )
 
+// NotificationEndpointService represents a service for managing notification endpoints.
+type NotificationEndpointService interface {
+	// UserResourceMappingService must be part of all NotificationEndpointStore service,
+	// for create, delete.
+	UserResourceMappingService
```
desa: Why does it need this embedding?

jsteenb2: @desa it doesn't, the original implemntation had both orgSVC and this one embedded to use in interesting ways... mostly for testing it looked like. I just pushed up the change to nuke this dep btw, good timing 👌

kelwang:
Looks much better then the current one, 
I have 2 suggestions:

1. add a count method inside storage, this could be super helpful for limitation services. Currently, we will unmarshal every entry.

2. Seperate the RetrieveOne method, with CUD. for example, most of time we use orgSVC, or bktSVC , we only doing just a retrieving, whether by name or id. If we can split them, our dependency pattern can be much cleaner. So in the future, we might, concurrently fetch org, bucket, and the target svc entry.

Issue #16306, opened 2019-12-20
-0 displayed in single stat for -0.001 with truncated decimals


desa:
![image](https://user-images.githubusercontent.com/1477539/71285012-fda72b80-2318-11ea-9edd-33d24cecf227.png)

We should special case 0 so that we don't have `-0` displayed for `-0.000...`.

desa:
cc @alsargent

TCL735:
To add additional context, JavaScript has two different zeroes: +0 and -0

PR #16439, opened 2020-01-07
fix(ui): Prevent negative zero and allow zero with decimal places

TCL735:
Closes https://github.com/influxdata/influxdb/issues/16306

- Prevent negative zero
- Allow zero to have decimal places

### DIFF
--- a/ui/src/shared/utils/formatStatValue.ts
+++ b/ui/src/shared/utils/formatStatValue.ts
@@ -1,7 +1,14 @@
-import {MAX_DECIMAL_PLACES} from 'src/dashboards/constants'
+// Libraries
 import {isNumber, isString} from 'lodash'
 
+// Types
 import {DecimalPlaces} from 'src/types/dashboards'
+
+// Constants
+import {MAX_DECIMAL_PLACES} from 'src/dashboards/constants'
+
+// Utils
+import {preventNegativeZero} from 'src/shared/utils/preventNegativeZero'
 
 interface FormatStatValueOptions {
   decimalPlaces?: DecimalPlaces
@@ -13,7 +20,7 @@ export const formatStatValue = (
   value: number | string = 0,
   {decimalPlaces, prefix, suffix}: FormatStatValueOptions = {}
 ): string => {
-  let localeFormattedValue = ''
+  let localeFormattedValue
 
   if (isNumber(value)) {
     let digits: number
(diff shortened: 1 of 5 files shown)


ui/src/shared/utils/formatStatValue.ts
```diff
@@ -1,8 +1,15 @@
-import {MAX_DECIMAL_PLACES} from 'src/dashboards/constants'
+// Libraries
 import {isNumber, isString} from 'lodash'
 
+// Types
 import {DecimalPlaces} from 'src/types/dashboards'
 
+// Constants
+import {MAX_DECIMAL_PLACES} from 'src/dashboards/constants'
```
Asalem1: Nice job organizing these

ui/src/shared/utils/preventNegativeZero.ts
```diff
@@ -0,0 +1,8 @@
+export const preventNegativeZero = (
+  value: number | string
+): number | string => {
+  if (Number(value) === -0) {
```
Asalem1: Literally didn't know this was possible

ui/src/shared/utils/formatStatValue.ts
```diff
@@ -26,15 +33,19 @@ export const formatStatValue = (
 
     const roundedValue = value.toFixed(digits)
 
-    localeFormattedValue = Number(roundedValue).toLocaleString(undefined, {
-      maximumFractionDigits: MAX_DECIMAL_PLACES,
-    })
+    localeFormattedValue =
+      Number(roundedValue) === 0
```
hoorayimhelping: what do you think of turning this code into a formatting function and including the negative zero check as part of it?

ui/src/shared/utils/preventNegativeZero.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    expect(preventNegativeZero(nonZeroString)).toEqual(nonZeroString)
+
+    nonZeroString = '-123456789.0001'
+    expect(preventNegativeZero(nonZeroString)).toEqual(nonZeroString)
+  })
+  it('should handle negative zero as a number', () => {
+    expect(preventNegativeZero(-0)).toEqual(0)
+    expect(preventNegativeZero(-0.0)).toEqual(0)
+    /*
+      // Note: prettier prevents us from adding additional tests with these patterns:
```
hoorayimhelping: screw that razzmatazz. our tools are here to help us, not force us to conform to arbitrary patterns.

let's add these tests back and then add this to them

```
// prettier-ignore
```
so that prettier lets us do our thing.

https://prettier.io/docs/en/ignore.html#javascript

ui/src/shared/utils/formatStatValue.ts
TCL735: This piece of code is actually inside a formatting function already. The ternary is to allow positive zero to have decimal places when user-specified.

hoorayimhelping: well then

ui/src/shared/utils/preventNegativeZero.test.ts
```diff
@@ -0,0 +1,43 @@
+import {preventNegativeZero} from 'src/shared/utils/preventNegativeZero'
+
+describe('preventNegativeZero', () => {
+  it('should not alter non-zero numbers', () => {
```
hoorayimhelping: nice test! was going to ask why you can't just use `Math.abs` but this one answers it nicely.

PR #16445, opened 2020-01-08
refactor(ui): normalization of tasks

121watts:
RFC normalization epic: https://github.com/influxdata/idpe/issues/5591


ui/src/schemas/index.ts
```diff
@@ -69,3 +84,10 @@ export const arrayOfTelegrafs = [telegraf]
 
 export const scraper = new schema.Entity(ResourceType.Scrapers)
 export const arrayOfScrapers = [scraper]
+
+export const addLabels = <R extends {labels?: Label[]}>(resource: R): R => {
```
ebb-tide: ooh- do resource types come out of this function with required labels properties? I wonder if the last `R` on this line should be `R & {labels: Label[]}` instead?

ui/src/types/tasks.ts
```diff
@@ -1,6 +1,36 @@
 import {Task as ITask} from 'src/client'
-import {Label} from 'src/types'
+import {Label, NormalizedState, Run, RemoteDataState, LogEvent} from 'src/types'
 
 export interface Task extends ITask {
   labels?: Label[]
```
ebb-tide: I feel like this can be required since we are adding defaults?

ui/src/utils/api.ts
```diff
@@ -6,25 +6,25 @@ import {getAPIBasepath} from 'src/utils/basepath'
...
 
-  if (message === '') {
-    message = get(e, 'message', '')
+  if (!message) {
+    message = get(e, 'message')
   }
 
-  if (message === '') {
+  if (!message) {
     message = 'unknown error'
```
ebb-tide: ah thank you for this!

ui/src/schemas/index.ts
ebb-tide: Separately- should we not be adding labels to buckets?

ebb-tide: There's also a version of this in `src/labels/utils/index.ts` should this replace it?

121watts: I'll be doing a whole labels overhaul during the `Label` normalization epic!

ui/src/types/tasks.ts
121watts: Good catch!  I wanted to do the bare minimum until I normalize the `Labels` resource.  All of this is going to change when labels within a resource are just a list of label ids.

ui/src/utils/api.ts
121watts: be the change

PR #16449, opened 2020-01-08
feat: adding name and description to telegraf [wip]

drdelambre:
this adds the last half of saving to the telegraf editor, without actually saving it. also enables the editor when editing the telegraf config instead of just on creation

stashing while i'm doing other stuff


wbaker85:
We're using the UI assets from the `ui` repo now per #21090, which includes a fully-functional telegraf editor, so I'm closing this PR out!

PR #16451, opened 2020-01-08
chore(http): fixup middleware stack and normalize metrics into same buckets for id fields

jsteenb2:
Closes #

- [x] Rebased/mergeable
- [ ] Tests pass
- [x] fix upstream

### DIFF
--- a/http/handler.go
+++ b/http/handler.go
@@ -5,14 +5,11 @@ import (
 	"encoding/json"
 	"net/http"
 	_ "net/http/pprof" // used for debug pprof at the default path.
-	"strings"
-	"time"
 
+	"github.com/go-chi/chi"
 	"github.com/influxdata/influxdb/kit/prom"
-	"github.com/influxdata/influxdb/kit/tracing"
-	"github.com/opentracing/opentracing-go"
+	kithttp "github.com/influxdata/influxdb/kit/transport/http"
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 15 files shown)


http/ready.go
```diff
@@ -9,26 +9,28 @@ import (
 	"github.com/influxdata/influxdb/toml"
 )
 
-var up = time.Now()
-
 // ReadyHandler is a default readiness handler. The default behaviour is always ready.
-func ReadyHandler(w http.ResponseWriter, r *http.Request) {
-	w.WriteHeader(http.StatusOK)
+func ReadyHandler() http.Handler {
+	up := time.Now()
```
jsteenb2: just wanted to encapsulate a global dependency, no need to bleed it out

http/handler.go
```diff
@@ -31,6 +28,7 @@ const (
 // All other requests are passed down to the sub handler.
 type Handler struct {
 	name string
+	r    chi.Router
 	// MetricsHandler handles metrics requests
 	MetricsHandler http.Handler
 	// ReadyHandler handles readiness checks
```
jsteenb2: the DebugHandler field is never being set anywhere. I'm trying to find anything that uses it but I can't find anything. Can we axe the debug handler since its not being used?

https://github.com/influxdata/influxdb/pull/16451/files

jsteenb2:
upstream branch is ready for when this hits 👍

kelwang:
can you fix the idpe branch?

jsteenb2:
@kelwang I ahve it ready already, relies on thsi being merged first.

http/pkger_http_server.go
```diff
@@ -36,10 +36,11 @@ func NewHandlerPkg(log *zap.Logger, errHandler influxdb.HTTPErrorHandler, svc pk
 	}
 
 	r := chi.NewRouter()
-	r.Use(middleware.RequestID)
-	r.Use(middleware.RealIP)
-	r.Use(traceMW)
```
jademcgough: is pkger tracing being added somewhere else now?

jsteenb2: yah, its being done at the platform handler level. I never saw it before. Its really tucked away in there. It is duplicative with the trace mw that is already in place.

jademcgough: Right on, just noticed that.

jsteenb2: it ws an oversight on my end during the initial pkger work

PR #16469, opened 2020-01-09
feat(http): add configurable limit to points batch size on write endpoint

### LINKED ISSUE #13100: Add limits when decompressing gzipped write requests
## Why

It is possible to construct a gzipped request that can expand to a size that exceeds a reasonable limit, resulting in a denial of service by potentially crashing the server.

The following reads the entire contents of the `io.Reader` into a `[]byte`

https://github.com/influxdata/influxdb/blob/8729cd095975ca55dcd38dc7c5720eba7f622ea8/http/write_handler.go#L208

## How

The amount of data read from the source `io.Reader` should be limited, which can be achieved by using a `LimitReader`.

The default for OSS should remain as unlimited. At some future point, `influxd` will support external configuration (e.g. a .toml file), where it should be available as an option.

The limit should be configurable to the http package such that Cloud 2 services may override the default.

GeorgeMac:
Closes #13100

This enables support for a limit on the points batch size in bytes (after compression if compression is enabled).

To configure this the functional option `WithMaxBatchSizeByte(50 * 1024 * 1024)` should be used when constructing a `http.WriteHandler`.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/write_handler.go
+++ b/http/write_handler.go
@@ -3,6 +3,7 @@ package http
 import (
 	"compress/gzip"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -19,6 +20,13 @@ import (
 	"github.com/influxdata/influxdb/storage"
 	"github.com/influxdata/influxdb/tsdb"
 	"go.uber.org/zap"
+)
+
+var (
+	// ErrMaxBatchSizeExceeded is returned when a points batch exceeds
+	// the defined upper limit in bytes. This pertains to the size of the
+	// batch after inflation from any compression (i.e. ungzipped).
+	ErrMaxBatchSizeExceeded = errors.New("points batch is too large")
 )
 
 // WriteBackend is all services and associated parameters required to construct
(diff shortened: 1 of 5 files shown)


http/write_handler.go
```diff
@@ -369,3 +395,37 @@ func compressWithGzip(data io.Reader) (io.Reader, error) {
...
+
+// Close returns an ErrMaxBatchSizeExceeded when the wrapped reader
+// exceeds the set limit for number of bytes.
+// This is safe to call more than once as long as the provided io.ReadCloser is
+// also safe to close more than once.
+func (l *limitedReader) Close() (err error) {
+	defer func() {
+		if cerr := l.close(); cerr != nil && err == nil {
+			err = cerr
+		}
```
GeorgeMac: note to self: I could swap the close after the first use for a noop one so it ensures l.close() only called once.

http/write_handler.go
```diff
@@ -210,22 +238,28 @@ func (h *WriteHandler) handleWrite(w http.ResponseWriter, r *http.Request) {
...
+
+	// close the reader now that all bytes have been consumed
+	// this will return non-nil in the case of a configured limit
+	// being exceeded
+	if err := in.Close(); err != nil {
+		log.Error("Error reading body", zap.Error(err))
+
+		code := influxdb.EInternal
+		if errors.Is(err, ErrMaxBatchSizeExceeded) {
+			code = influxdb.ETooLarge
```
stuartcarnie: Great to see the use of proper HTTP status codes 🎉

GeorgeMac: You can tell what wikipedia page I spent some time on for this PR 😂

PR #16478, opened 2020-01-09
refactor: normalize variables resource

121watts:
RFC: https://github.com/influxdata/idpe/issues/5591


ui/src/resources/reducers/helpers.ts
```diff
@@ -46,12 +39,9 @@ export const editResource = <R>(
   resource: ResourceType
 ) => {
   const {entities, result} = action.schema
-  if (!draftState.byID) {
-    draftState.byID = {}
-  }
 
-  const bucket = entities[resource][result]
-  draftState.byID[result] = bucket
+  const r = entities[resource][result]
```
hoorayimhelping: why have a transient variable called `r`? `r` doesn't provide any more meaning than `entities[resource][result]` what do you think about changing this to:

```ts
draftState.byID[result] = entities[resource][result]
```
?

ui/src/schemas/index.ts
```diff
@@ -85,9 +95,30 @@ export const arrayOfTelegrafs = [telegraf]
...
+    processStrategy: (v: Variable): Variable => {
+      return {
+        ...v,
+        labels: addLabels(v),
+        status: addStatus(v),
+      }
+    },
   }
+)
+export const arrayOfVariables = [variable]
```
hoorayimhelping: what's this all about? doing a `cmd+f` for it doesn't show it used in this pr

ui/src/types/schemas.ts
```diff
@@ -66,3 +65,11 @@ export interface TaskEntities {
     [uuid: string]: Task
   }
 }
+
+// VariableEntities defines the result of normalizr's normalization
```
hoorayimhelping: 👍

ui/src/types/variables.ts
```diff
@@ -1,7 +1,63 @@
...
+
+export interface ValueSelections {
+  [variableID: string]: string
+}
+
+export interface VariablesState extends NormalizedState<Variable> {
+  values: {
+    // Different variable values can be selected in different
+    // "contexts"---different parts of the app like a particular dashboard, or
+    // the Data Explorer
```
hoorayimhelping: love this comment

ui/src/variables/actions/thunks.ts
```diff
@@ -365,51 +280,57 @@ export const addVariableLabelAsync = (
...
+      schemas.variable
+    )
 
     dispatch(setVariable(variableID, RemoteDataState.Done, variable))
   } catch (error) {
     console.error(error)
-    dispatch(notify(removeVariableLabelFailed()))
+    dispatch(notify(copy.removeVariableLabelFailed()))
   }
 }
```
hoorayimhelping: love the changes in this file, great work!

ui/src/variables/reducers/editor.ts
```diff
@@ -0,0 +1,56 @@
...
+})
+
+export const variableEditorReducer = (
+  state: VariableEditorState = initialEditorState(),
+  action: EditorAction
+): VariableEditorState =>
+  produce(state, draftState => {
+    switch (action.type) {
+      case CLEAR_VARIABLE_EDITOR: {
+        return initialEditorState()
```
hoorayimhelping: does `draftState` need to be set to `initialEditorState` here? this case is kinda different in form from all the other cases.

ui/src/variables/selectors/index.tsx
```diff
@@ -15,29 +15,28 @@ import {
...
+type VariablesState = ResourceState['variables']['byID']
+type ValuesState = ResourceState['variables']['values']['contextID']
 
 const extractVariablesListMemoized = memoizeOne(
   (variablesState: VariablesState): Variable[] => {
-    return Object.values(variablesState)
-      .filter(d => d.status === RemoteDataState.Done)
-      .map(d => d.variable)
+    return Object.values(variablesState).filter(
+      d => d.status === RemoteDataState.Done
```
hoorayimhelping: what is `d`?

121watts: idk.  it was there  when i got there.

ui/src/variables/reducers/editor.ts
121watts: nope! https://immerjs.github.io/immer/docs/return

ui/src/schemas/index.ts
121watts: Should normalizer want to normalize a list of a foo it uses this syntax for it's schema.  I could just inline this, but (imo), `variable` and `[variable]` looked to similar when consumed as arguments to the `normalize` function i.e.:

`normalize(data, fooSchema)` and `normalize(data, [fooSchema])` looked really similar but return different results.  So, to make it easier on the eye I assigned all resource list schemas to a variable.

ui/src/variables/selectors/index.tsx
121watts: idk `d` was there when i got there.  😄

ui/src/variables/reducers/editor.ts
121watts: nope!  https://immerjs.github.io/immer/docs/return

ui/src/resources/reducers/helpers.ts
121watts: its certainly better than `buckets`.  it's really a resource but resource is already taken 🤷‍♂.  Good enough for golang!

ui/src/schemas/index.ts
121watts: I can remove it if it's not used.

ui/src/variables/selectors/index.tsx
hoorayimhelping: this is how i feel about little things like this:

[>don't ya see it? well pick it up!!](https://www.youtube.com/watch?v=wwQ5SmhwGgk&feature=youtu.be&t=156)

basically it's all our job to sweep the floor and we should keep the place tidy for the next person. anything we can do to reduce our collective mental load. to that point, maybe `data` is slightly better than `d`?

ui/src/resources/reducers/helpers.ts
hoorayimhelping: the point i'm making is that it's already clear from the assignment and the name `entities[resource][result]` that `r` represents a result, and if we can eliminate one more line of code without losing any meaning, i'm generally 110% for that.

>Good enough for golang!

we're not writing go though, we're writing typescript / javascript

PR #16491, opened 2020-01-10
fix(influx): add env vars to cli usage and normalize usage and flag/env var priority

### LINKED ISSUE #16048: Add environment variables to the influx --help command

### LINKED ISSUE #16246: cli flags should override environmental variables

jsteenb2:
also rids the pkg of the  cobra tutorial init style and encapsulates things into funcs. One step closer to builder pattern for this cmd pkg.

Closes #16246
Closes #16048 

- [x] Rebased/mergeable
- [x] Tests pass
- [x] Documentation updated or issue created (provide link to issue/pr) https://v2.docs.influxdata.com/v2.0/reference/cli/influx


jsteenb2:
@sanderson this will make a bunch of changes to the CLI help outputs. The env vars will be exposed so folks will know what they are when they see the `-h` usage.

cmd/influx/authorization.go
```diff
@@ -226,15 +218,15 @@ func authorizationCreateF(cmd *cobra.Command, args []string) error {
...
+		OrgID:       orgID,
 	}
 
 	if userName := authCreateFlags.user; userName != "" {
 		userSvc, err := newUserService()
 		if err != nil {
 			return err
 		}
-		user, err := userSvc.FindUser(ctx, platform.UserFilter{
+		user, err := userSvc.FindUser(context.Background(), platform.UserFilter{
```
kelwang: i think we would like to hook up the context sometime in the future, that's why the ctx is defined on top

cmd/influx/authorization.go
```diff
@@ -299,23 +287,13 @@ func authFindCmd() *cobra.Command {
...
-	if h := viper.GetString("ORG_ID"); h != "" {
-		authorizationFindFlags.orgID = h
-	}
 	cmd.Flags().StringVarP(&authorizationFindFlags.id, "id", "i", "", "The authorization ID")
 
 	return cmd
 }
 
-func newAuthorizationService(f Flags) (platform.AuthorizationService, error) {
+func newAuthorizationService() (platform.AuthorizationService, error) {
```
kelwang: I am never a big fan of anything global. If passing the flags bothers you, you could compose a struct and make this function a method.

cmd/influx/main.go
```diff
@@ -73,32 +75,11 @@ func out(w io.Writer) genericCLIOptfn {
...
-	}
-	return cmd
-}
-
-func seeHelp(c *cobra.Command, args []string) {
-	if c = fetchSubCommand(c, args); c == nil {
-		return //return here, since cobra already handles the error
-	}
-	c.Printf("See '%s -h' for help\n", c.CommandPath())
+var flags struct {
```
kelwang: i don't really like such a global thing.

cmd/influx/authorization.go
jsteenb2: I wish that were the case, but the context that was removed above waws just `context.Background()` 🤷‍♂. It just happened to be instantiated well before its use case.

cmd/influx/main.go
jsteenb2: totally agree. I didn't remedy all the issues in this pkg on this go round. However, I agree, the global constructs in this are a major issue and should be fixed. I didn't do it here b/c it was already bigger body of work than I anticipated 😢

cmd/influx/authorization.go
jsteenb2: totally agree. I'd like to see all the commands use the builder pattern of sorts so we can dep inject. We'r enot quite there yet. That'll take some time.

PR #16497, opened 2020-01-10
fix(http): expose max batch size bytes limit option on *APIBackend

GeorgeMac:
I missed a bit. This is a continuation of https://github.com/influxdata/influxdb/commit/a0c18c9ef700fd23c000ef79fe07cf905fef8191

This exposes the MaxBatchSizeBytes property on the *APIBackend. Such this can be configured higher up the call stack.
I also did a little extra housekeeping.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/write_handler.go
+++ b/http/write_handler.go
@@ -146,24 +146,6 @@ func (h *WriteHandler) handleWrite(w http.ResponseWriter, r *http.Request) {
 		})
 	}()
 
-	var in io.ReadCloser = r.Body
-	defer in.Close()
-
-	if r.Header.Get("Content-Encoding") == "gzip" {
-		var err error
-		in, err = gzip.NewReader(r.Body)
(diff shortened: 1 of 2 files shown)


http/write_handler.go
```diff
@@ -306,6 +275,38 @@ func decodeWriteRequest(ctx context.Context, r *http.Request) (*postWriteRequest
...
+	defer func() {
+		// close the reader now that all bytes have been consumed
+		// this will return non-nil in the case of a configured limit
+		// being exceeded
+		if cerr := rc.Close(); err == nil {
+			err = cerr
+		}
+	}()
+
+	if encoding == "gzip" {
```
docmerlin: also check if it is `x-gzip`, as http 1.1 spec recommends you check for it as well, as an alias for `gzip`

Issue #5171, opened 2015-12-19
Inconsistent file permissions


mark-rushakoff:
Sometimes we make directories `0700`, sometimes `0777`:

```
→ git grep -w ' 0[0-9][0-9][0-9]' | grep '.go:' | grep -v _test.go | grep dir
cmd/influxd/restore/restore.go: if err := os.MkdirAll(filepath.Dir(path), 0777); err != nil {
cmd/influxd/run/command.go: err := os.MkdirAll(filepath.Dir(path), 0777)
meta/store.go:  return os.MkdirAll(s.path, 0777)
services/hh/node_processor.go:  if err := os.MkdirAll(n.dir, 0700); err != nil {
services/hh/service.go: if err := os.MkdirAll(s.cfg.Dir, 0700); err != nil {
tsdb/engine/tsm1/engine.go: if err := os.MkdirAll(e.path, 0777); err != nil {
tsdb/engine/tsm1/wal.go:    if err := os.MkdirAll(l.path, 0777); err != nil {
tsdb/engine/wal/wal.go: if err := os.MkdirAll(l.path, 0777); err != nil {
tsdb/store.go:  if err := os.MkdirAll(filepath.Join(s.path, database, retentionPolicy), 0700); err != nil {
tsdb/store.go:  if err := os.MkdirAll(walPath, 0700); err != nil {
tsdb/store.go:  if err := os.MkdirAll(s.path, 0777); err != nil {
```

We usually use `0666` for files, with some exceptions from the CLI history, Raft's peers.json, and the hinted handoff queue.

```
→ git grep -w ' 0[0-9][0-9][0-9]' | grep '.go:' | grep -v _test.go | grep -i file | grep -vi mkdir
cmd/influx/cli/cli.go:      if c.historyFile, err = os.OpenFile(historyFilePath, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0640); err == nil {
cmd/influx_inspect/tsm.go:  f, err := os.OpenFile(filepath.Join(path, FieldsFileExtension), os.O_RDONLY, 0666)
cmd/influx_inspect/tsm.go:  f, err := os.OpenFile(filepath.Join(path, SeriesFileExtension), os.O_RDONLY, 0666)
cmd/influx_inspect/tsm.go:  f, err := os.OpenFile(filepath.Join(path, IDsFileExtension), os.O_RDONLY, 0666)
[...]

PR #16504, opened 2020-01-10
feat(storage): implement backup and restore

### LINKED ISSUE #15603: storage 2.0: implement data backup

### LINKED ISSUE #15604: storage 2.0: implement meta data backup

### LINKED ISSUE #15605: storage 2.0: implement data restore from backup

### LINKED ISSUE #15606: storage 2.0: implement meta data restore from backup

jacobmarble:
Closes #15603
Closes #15604
Closes #15605
Closes #15606

Implements online backup and offline restore.

To test functionality it was necessary to combine #15558 and #16352 into one PR based on recent master/HEAD. Then:

- wrote some data
- read it back
- backed it up with `influx backup`
- terminated influxd
- restored it with `influxd restore` to existing database directory
- started influxd, read it back, terminated influxd
- restored it with `influxd restore` to empty database directory
- started influxd, read it back, terminated influxd

Check before merging:
- [x] "restored it with `influxd restore` to empty database directory" fails
- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influx/backup.go
+++ b/cmd/influx/backup.go
@@ -1,0 +1,111 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/bolt"
+	"github.com/influxdata/influxdb/http"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"go.uber.org/multierr"
+)
+
+func cmdBackup() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "backup",
+		Short: "Backup the data in InfluxDB",
+		Long: fmt.Sprintf(
+			`Backs up data and meta data for the running InfluxDB instance.
+Downloaded files are written to the directory indicated by --path.
+The target directory, and any parent directories, are created automatically.
+Data file have extension .tsm; meta data is written to %s in the same directory.`,
+			bolt.DefaultFilename),
+		RunE: backupF,
+	}
+	opts := flagOpts{
+		{
+			DestP:    &backupFlags.Path,
+			Flag:     "path",
+			Short:    'p',
+			EnvVar:   "PATH",
+			Desc:     "directory path to write backup files to",
+			Required: true,
+		},
+	}
+	opts.mustRegister(cmd)
+
+	return cmd
+}
+
+var backupFlags struct {
+	Path string
+}
+
+func init() {
+	err := viper.BindEnv("PATH")
+	if err != nil {
+		panic(err)
+	}
+	if h := viper.GetString("PATH"); h != "" {
+		backupFlags.Path = h
+	}
+}
+
+func newBackupService() (influxdb.BackupService, error) {
+	return &http.BackupService{
+		Addr:  flags.host,
+		Token: flags.token,
+	}, nil
+}
+
+func backupF(cmd *cobra.Command, args []string) error {
+	ctx := context.Background()
+
+	if flags.local {
+		return fmt.Errorf("local flag not supported for backup command")
+	}
+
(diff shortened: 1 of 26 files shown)


cmd/influx/backup.go
```diff
@@ -0,0 +1,97 @@
...
+Data file have extension .tsm; meta data is written to %s in the same directory.`,
+		bolt.DefaultFilename),
+	RunE: backupF,
+}
+
+var backupFlags struct {
+	Path string
+}
+
+func init() {
```
jsteenb2: can you add some tests here and remove the globals in line with hte direction were heading in the `influx` pkg.  See here for an example of testing in this pkg:
https://github.com/influxdata/influxdb/pull/16491/commits/ee3198fb07ebcd01500c52c19f57b3af257493e0

jacobmarble: Thanks for bringing this up. Looks like you merged just today, so when I rebased for another reason, this fix became necessary. Please take another look.

jacobmarble:
@tmgordeeva this is weird. Restore to existing directory works properly, but restore to empty dir looks like this:
```
$ influx query -o my-org 'from(bucket: "my-bucket") |> range(start: -3h)'
See 'influx query -h' for help
Error: Unauthorized access.
See 'influx query -h' for help
```

I think you can ignore the extra "See 'influx query -h' for help", it's the access that's the issue. I checked the bolt file MD5 sum before and after restore, and it looks identical in both scenarios. I've reproduced this error twice. Please take a look Monday.

tmgordeeva:
@jacobmarble I think the problem is the credentials aren't part of this backup/restore process. I got the same, but restored it to working by copying the credentials back in.

tmgordeeva:
@jacobmarble building idpe fails against this with:
```
etcd/kvservice.go:34:21: cannot use cache (type *KVCache) as type "github.com/influxdata/influxdb/kv".Store in argument to "github.com/influxdata/idpe/kv".NewClient:
	*KVCache does not implement "github.com/influxdata/influxdb/kv".Store (missing Backup method)
etcd/kvservice.go:56:21: cannot use store (type *KVStore) as type "github.com/influxdata/influxdb/kv".Store in argument to "github.com/influxdata/idpe/kv".NewClient:
	*KVStore does not implement "github.com/influxdata/influxdb/kv".Store (missing Backup method)
```

Created https://github.com/influxdata/idpe/pull/5776 to hopefully fix that.

tmgordeeva:
@jacobmarble latest commit should fix the bolt being copied to credentials issue. Problem with my testing was I was still using the token explicitly in queries which I didn't realize overrides whatever's in credentials, retested without it to verify and verified credentials files match before and after.

cmd/influxd/restore/command.go
```diff
@@ -0,0 +1,293 @@
...
+	enginePath string
+	credPath   string
+	backupPath string
+	rebuildTSI bool
+}
+
+func init() {
+	dir, err := fs.InfluxDir()
+	if err != nil {
+		panic(fmt.Errorf("failed to determine influx directory: %v", err))
```
sebito91: Swap to `%s` for string output.

jacobmarble:
> I think this is great, a couple of small nits included but one major question:
> 
> Why do we perform `os.MkdirAll` with `0777` permissions? That seems way too open for my liking, could we make do with `0755` instead? Do we truly need these files to be read/write/execute enable for `other`?

I scanned the code to figure out what we currently do. In the storage engine, 0777 and 0666 is very common, and in https://github.com/influxdata/influxdb/issues/5171 @benbjohnson argues that the user of InfluxDB can `umask` if the defaults are 0777 and 0666. Therefore, I made the change in that direction.

sebito91:
Thanks for that @jacobmarble, appreciate the history on this. I'm fine with the changes as long as you are OK with that `return err` comment.

PR #16511, opened 2020-01-12
feat(pkger): add jsonnet support for package files

jsteenb2:
Clearing up some idea debt. Adds [jsonnet](https://jsonnet.org/) support for influx packages.

- [x] Rebased/mergeable
- [x] Tests pass


pkger/parser.go
```diff
@@ -138,8 +148,14 @@ func parseSource(r io.Reader, opts ...ValidateOptFn) (*Pkg, error) {
 	switch {
 	case strings.Contains(contentType, "json"):
 		return parseJSON(bytes.NewReader(b), opts...)
-	default:
+	case strings.Contains(contentType, "jsonnet"):
```
kelwang: strings.Contains(contentType, "json") will always be true, even if the "content-type" is "jsonnet"

jsteenb2: ohhhh great catch 🥇, will fix that up 👍

jsteenb2: @kelwang push fixed

PR #16521, opened 2020-01-13
fix: paginate dashboards

drdelambre:
prevents large organizations from being crippled after they import their 1.x dashboards using pkgr

the technique is to add dashboard cards one at a time until the user's window is filled, use that number to represent page size (without connecting css rules + js code), and then to render a few pages.

when the user gets to the bottom of the last page, add another one, until there are no more pages to add. it has the downside of getting slow as the user scrolls to the bottom, but we can always nudge user behavior towards using the filter mechanism (keeping it always visible, type ahead, whatevs) if the need presents itself.

this PR also removes some prop drilling that was going on in the area, sorry for the clutter.


ui/src/dashboards/components/DashboardsCardGrid.scss
```diff
@@ -3,12 +3,25 @@
...
+  }
+}
+
 .dashboards-card-grid {
    display: grid;
    grid-template-columns: 1fr;
    grid-template-rows: 150px 150px 150px;
    grid-column-gap: $ix-marg-a;
    grid-row-gap: $ix-marg-a;
+   height: 100%;
```
Asalem1: You’re 100%

ui/src/dashboards/components/dashboard_index/DashboardCards.tsx
```diff
@@ -28,15 +26,110 @@ export default class DashboardCards extends PureComponent<Props> {
     getSortedResources
   )
 
+  state = {
+    hasMeasured: false,
+    pages: 1,
+    windowSize: 0,
+  }
+
+  public componentDidMount() {
```
Asalem1: If the state is being set initially, why are we using componentDidMount to set it again?

drdelambre: it's for when the component gets unmounted (like if you type in a filter that has no matches), and then remounted (remove the filter), to restart the measuring cycle

ui/src/dashboards/components/DashboardsCardGrid.scss
drdelambre: it was the worst drilling 100% height all the way to the frame component, or else it just chooses whatever height it wants to be

ui/webpack.dev.ts
```diff
@@ -11,6 +11,10 @@ module.exports = merge(common, {
   output: {
     filename: '[name].js',
   },
+  watchOptions: {
```
drdelambre: this was to make webpack work in docker on linux

ui/src/dashboards/components/dashboard_index/DashboardCards.tsx
Asalem1: Very interesting. I'll allow it

PR #16522, opened 2020-01-13
feat: resource log

gavincabbage:
This change introduces a resource audit logger to track changes to resources in the system. The operations on tasks, buckets and organizations are augmented to use this logger. 

No implementation is currently provided.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/bucket.go
+++ b/bucket.go
@@ -2,6 +2,7 @@ package influxdb
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"time"
 )
@@ -68,6 +69,7 @@ var (
 	OpFindBucket     = "FindBucket"
 	OpFindBuckets    = "FindBuckets"
 	OpCreateBucket   = "CreateBucket"
+	OpPutBucket      = "PutBucket"
 	OpUpdateBucket   = "UpdateBucket"
 	OpDeleteBucket   = "DeleteBucket"
 )
@@ -152,4 +154,13 @@ func (f BucketFilter) String() string {
 		parts = append(parts, "Org Name: "+*f.Org)
 	}
 	return "[" + strings.Join(parts, ", ") + "]"
+}
+
+func ErrInternalBucketServiceError(op string, err error) *Error {
+	return &Error{
+		Code: EInternal,
+		Msg:  fmt.Sprintf("unexpected error in buckets; Err: %v", err),
+		Op:   op,
+		Err:  err,
+	}
 }
--- a/context/token_test.go
+++ b/context/token_test.go
@@ -8,6 +8,21 @@ import (
 	icontext "github.com/influxdata/influxdb/context"
 )
 
+func TestGetAuthorizer(t *testing.T) {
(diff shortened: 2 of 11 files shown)


context/token_test.go
```diff
@@ -8,6 +8,21 @@ import (
 	icontext "github.com/influxdata/influxdb/context"
 )
 
+func TestGetAuthorizer(t *testing.T) {
```
stuartcarnie: 💯

kv/bucket.go
```diff
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/influxdata/influxdb/resource"
+
```
stuartcarnie: nit: can you remove the blank line and run `goimports` so they group together

bucket.go
```diff
@@ -153,3 +155,12 @@ func (f BucketFilter) String() string {
...
 }
+
+func ErrInternalBucketServiceError(op string, err error) *Error {
+	return &Error{
+		Code: EInternal,
+		Msg:  fmt.Sprintf("unexpected error in buckets; Err: %v", err),
+		Op:   op,
+		Err:  err,
+	}
+}
```
stuartcarnie: This is a great addition 🥇

kv/bucket.go
```diff
@@ -477,14 +479,28 @@ func (s *Service) createBucket(ctx context.Context, tx Tx, b *influxdb.Bucket) (
 		}
 	}
 
-	if err := s.putBucket(ctx, tx, b); err != nil {
+	v, err := json.Marshal(b)
+	if err != nil {
+		return influxdb.ErrInternalBucketServiceError(influxdb.OpCreateBucket, err)
```
stuartcarnie: I ❤️ the distinct errors

resource/resource.go
```diff
@@ -0,0 +1,49 @@
...
+	ResourceType influxdb.ResourceType
+	// OrganizationID of the organization owning the changed resource.
+	OrganizationID influxdb.ID
+	// UserID of the user changing the resource.
+	UserID influxdb.ID
+	// ResourceBody after the change.
+	ResourceBody []byte
+	// Time when the resource was changed.
+	Time time.Time
+}
```
stuartcarnie: Wanted to state the design here, for my own clarification and thinking aloud. tl;dr is this approach is sound 👍 

----

To know the specific changes made to a resource over time, we read its entire history and look at each prior `ResourceBody` to determine the deltas. Knowing that we will store the change history in InfluxDB, this approach is more efficient. Why? Prior values will be in the same column as the `ResourceBody` field and very likely stored in the same TSM block, so will be decompressed together. 💯 

----

For posterity, the alternative would be to include the `PreviousResourceBody` as an additional value as a separate column. Whilst this approach may be better for a row-based data store, it will require separate reads for a columnar-based store like InfluxDB. 

💯

PR #16523, opened 2020-01-13
feat(pkger): make pkger packages CRD compliant

### LINKED ISSUE #16501: Pkger should be compatible with a Kubernetes Object
The manifest format is really close to be a K8s object but not quite there. 
at the top level these should be only four keys: `apiVersion`, `kind`, `metadata` and `spec`
- `apiVersion` right now is `0.1.0` but should be in the line of `influxdata.com/v1alpha1` https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/
- `kind` should be always formatted in a CamelCase format
- should be `metadata` and now it is `meta`

I don't understand why we have a super resource called Package

I imagine something going from this:
https://github.com/influxdata/influxdb/blob/master/pkger/testdata/variables.yml
```yaml
apiVersion: 0.1.0
kind: Package
meta:
  pkgName:      pkg_name
  pkgVersion:   1
  description:  pack description
spec:
  resources:
    - kind: Variable
      name: var_query_1
      description: var_query_1 desc
      type: query
      language: flux
      query: |
        buckets()  |> filter(fn: (r) => r.name !~ /^_/)  |> rename(columns: {name: "_value"})  |> keep(columns: ["_value"])
    - kind: Variable
      name: var_query_2
      description: var_query_2 desc
      type: query
      query: an influxql query of sorts
      language: influxql
    - kind: Variable
      name: var_const_3
      description: var_const_3 desc
      type: constant
      values:
        - first val
    - kind: Variable
      name: var_map_4
      description: var_map_4 desc
      type: map
      values:
        k1: v1
```

to something like

```yaml
apiVersion: influxdata.com/v1alpha1
kind: Variable
metadata:
  name: var_query_1
spec:
  description: var_query_1 desc
  type: query
  language: flux
  query: |
      buckets()  |> filter(fn: (r) => r.name !~ /^_/)  |> rename(columns: {name: "_value"})  |> keep(columns: ["_value"])
---
apiVersion: influxdata.com/v1alpha1
kind: Variable
metadata:
  name: var_query_2
spec:
  description: var_query_2 desc
  type: constant
  query: an influxql query of sorts
  language: influxql
---
apiVersion: influxdata.com/v1alpha1
kind: Variable
metadata:
  name: var_query_3
spec:
  description: var_query_3 desc
  type: constant
  values:
    - first val
---
apiVersion: influxdata.com/v1alpha1
kind: Variable
metadata:
  name: var_query_4
spec:
  description: var_query_4 desc
  type: map
  values:
    k1: v1
```

What do you think @jsteenb2? any thought @BondAnthony ?

Notes for deciding to add CRDs: https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#should-i-add-a-custom-resource-to-my-kubernetes-cluster

jsteenb2:
Closes #16501 

Makes packages compliant with k8s CRDs. More detail in the issue linked above.

Recommend viewing the PR from this link (with only `.go` files showing): https://github.com/influxdata/influxdb/pull/16523/files?file-filters%5B%5D=.go

- [x] Rebased/mergeable
- [x] Tests pass


http/pkger_http_server_test.go
```diff
@@ -261,75 +263,57 @@ func TestPkgerHTTPServer(t *testing.T) {
...
-func bucketPkg(t *testing.T, encoding pkger.Encoding) *pkger.Pkg {
+func bucketPkgKinds(t *testing.T, encoding pkger.Encoding) []pkger.Kind {
 	t.Helper()
 
 	var pkgStr string
 	switch encoding {
 	case pkger.EncodingJsonnet:
 		pkgStr = `
 local Bucket(name, desc) = {
+    apiVersion: 'influxdata.com/v2alpha1',
```
kelwang: can you make a global constant for the api version, also i think we should use beta1, instead of alpha1

pkger/parser.go
```diff
@@ -144,7 +151,7 @@ func parseSource(r io.Reader, opts ...ValidateOptFn) (*Pkg, error) {
 		b = bb
 	}
 
-	contentType := http.DetectContentType(b)
+	contentType := http.DetectContentType(b[:512])
```
kelwang: quick question, i remembered it used to like b[:512], is there any reason to change it back and forth

pkger/parser.go
```diff
@@ -155,7 +162,7 @@ func parseSource(r io.Reader, opts ...ValidateOptFn) (*Pkg, error) {
 		strings.Contains(contentType, "yml"):
 		return parseYAML(bytes.NewReader(b), opts...)
 	default:
-		return parseJsonnet(r, opts...)
+		return parseJsonnet(bytes.NewReader(b), opts...)
```
kelwang: same question above

pkger/parser.go
```diff
@@ -180,19 +187,27 @@ func parse(dec decoder, opts ...ValidateOptFn) (*Pkg, error) {
 	return &pkg, nil
 }
 
+// Kind describes the metadata and raw spec for a package kind.
+type Kind struct {
```
kelwang: I would just call it pkger or PkgerItem. Kind is really confusing. And that kindType should be kind

pkger/parser.go
jsteenb2: I messed up. I dorked it up, but this is back to what it should be. It should only ever check the first 512 bytes. Although, don't think it hurts to give it the entire byte array, just not common to see it that way.

pkger/parser.go
jsteenb2: this was actually a bug. I drain the reader above to get that `[]byte`, then here I was reading from the drained reader 🤦‍♂.  This fixes that so it creates a new reader that can be read from again. If I had a read seeker here, I could just seek back to the beginining, but network streams being waht they are.... that ain't happening anytime soon lol

pkger/parser.go
jsteenb2: point taken. I'm not a fan of the `Item` word per say, nor hte stutter in the name. I'll look at k8s and a few other places to get a feel for their naming convention. I am calling it Kind here b/ that is what k8s calls it, felt better to align than diverge on this. The `KindType`, totally agree, confusing.

pkger/parser.go
jsteenb2: After a little perusing of the k8s API conventions.... I think I'll rename `Kind` to `Object` which is how k8s represents an entity of a kind. And then switch back `KindType` to `Kind`. 

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#types-kinds

PR #16531, opened 2020-01-14
feat(http): Enforce limits in write handler and expose config options

stuartcarnie:
This PR uses the `ParsePointsWithOptions` API for parsing a batch of points submitted to the `/api/v2/write` HTTP endpoint. 

All limits are disabled by default, but may be configured by setting values for the documented fields of the `APIBackend` structure.


GeorgeMac:
Looks good! All the levers!

PR #16532, opened 2020-01-14
Zs optionalsuffix15909

### LINKED ISSUE #15909: Optionally discard suffix in ticks on gauge
Screenshot 2019-11-14 09 47 36

We should allow the user to specify if they'd like to have the suffix appending to each tick in the gauge vs the numeric value that is displayed.

zoesteinkamp:
Closes #15909 

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -8590,19 +8590,7 @@ components:
           type: integer
     GaugeViewProperties:
       type: object
-      required:
-        - type
-        - queries
-        - colors
-        - shape
-        - note
-        - showNoteWhenEmpty
-        - prefix
-        - tickPrefix
-        - suffix
-        - tickSuffix
-        - legend
-        - decimalPlaces
+      required: [type, queries, colors, shape, note, showNoteWhenEmpty, prefix, tickPrefix, suffix, tickSuffix, legend, decimalPlaces]
       properties:
         type:
           type: string
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -8598,7 +8598,9 @@ components:
         - note
         - showNoteWhenEmpty
         - prefix
+        - tickPrefix
         - suffix
+        - tickSuffix
         - legend
         - decimalPlaces
       properties:
@@ -8624,7 +8626,11 @@ components:
           type: boolean
         prefix:
           type: string
+        tickPrefix:
+          type: string
         suffix:
+          type: string
+        tickSuffix:
           type: string
         legend:
           $ref: '#/components/schemas/Legend'
--- a/ui/src/shared/components/Gauge.tsx
+++ b/ui/src/shared/components/Gauge.tsx
@@ -28,7 +28,9 @@ interface Props {
   gaugePosition: number
   colors?: Color[]
   prefix: string
+  tickPrefix: string
   suffix: string
+  tickSuffix: string
   decimalPlaces: DecimalPlaces
 }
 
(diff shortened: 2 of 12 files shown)


ui/src/shared/components/Gauge.tsx
```diff
@@ -340,7 +351,7 @@ class Gauge extends Component<Props> {
     const textContent = formatStatValue(gaugePosition, {
       decimalPlaces,
       prefix,
-      suffix,
+      suffix
```
Asalem1: did prettier do this? I think we have hanging commas set in our linter preferences

ui/src/shared/components/Gauge.tsx
```diff
@@ -279,14 +281,23 @@ class Gauge extends Component<Props> {
...
+    const {tickPrefix, tickSuffix, decimalPlaces} = this.props
+    let {prefix, suffix} = this.props 
     const {degree, lineCount, labelColor, labelFontSize} = GAUGE_SPECS
 
     const tickValues = [
       ..._.range(minValue, maxValue, Math.abs(maxValue - minValue) / lineCount),
       maxValue,
     ]
 
+    if(tickPrefix === "true"){
```
Asalem1: despite being the default values, this seems a bit counterintuitive to have a string value set to `true`, is there any way we can have the defaults set to anything else? What happens if the user automatically sets the prefix or suffic to `"true"`?

ui/src/timeMachine/components/view_options/Affixes.tsx
```diff
@@ -2,21 +2,36 @@
...
+  Form,
+  Input,
+  Grid,
+  Toggle,
+  InputToggleType,
+  InputLabel,
+  FlexBox,
+  AlignItems,
+  ComponentSize,
+  JustifyContent
```
Asalem1: hanging comma

ui/src/shared/components/Gauge.tsx
zoesteinkamp: I did not do this, so possibly we should change our prettier?

ui/src/shared/components/Gauge.tsx
zoesteinkamp: So the reason we are doing this as a string  instead of a boolean is in case in the future we will allow people to input their own custom ticks as a string. So right now we are setting it to string false or string true. In the UI it is a checkbox, so the user can not input anything themselves.

ui/src/timeMachine/components/view_options/Affixes.tsx
zoesteinkamp: this is another prettier issue, as its not failing on the lint, so clearly its cool with this

Asalem1: super weird, 🤷‍♂

ui/src/shared/components/Gauge.tsx
Asalem1: Since we aren't currently focused on user input, I think it would be beneficial to keep things straightforward and have the types be a boolean option. Forward planning is always important, and you did a great job of investigating a solution for it, but since we don't know if this will be a feature request in the future, it might be a little strange for future developers to see `true` and `false` string values

ui/src/shared/components/Gauge.tsx
Asalem1: Nah, I wouldn't worry about it

http/swagger.yml
```diff
@@ -8589,7 +8589,9 @@ components:
         - note
```
jademcgough: This list of required properties has gotten really long. It might help readability to change it to an array format, like:
`required: [type, queries, colors, shape, note, showNoteWhenEmpty, prefix, suffix, tickPrefix, tickSuffix, legend, decimalPlaces]`

zoesteinkamp: Got it! adjusted it!

PR #16535, opened 2020-01-14
refactor: normalization of dashboards

121watts:


ui/src/dashboards/actions/creators.ts
```diff
@@ -0,0 +1,84 @@
...
+import {NormalizedSchema} from 'normalizr'
+
+export const SET_DASHBOARDS = 'SET_DASHBOARDS'
+export const SET_DASHBOARD = 'SET_DASHBOARD'
+export const REMOVE_DASHBOARD = 'REMOVE_DASHBOARD'
+export const DELETE_DASHBOARD_FAILED = 'DELETE_DASHBOARD_FAILED'
+export const EDIT_DASHBOARD = 'EDIT_DASHBOARD'
+export const REMOVE_CELL = 'REMOVE_CELL'
+export const ADD_DASHBOARD_LABEL = 'ADD_DASHBOARD_LABEL'
+export const REMOVE_DASHBOARD_LABEL = 'REMOVE_DASHBOARD_LABEL'
```
hoorayimhelping: what do you think about alphabetizing these? not sure how to do it in the vastly superior vscode, but in sublime, you highlight a block of code and hit `f5`.

ui/src/dashboards/apis/index.ts
```diff
@@ -154,3 +36,47 @@ export const updateView = async (
 
   return viewWithIDs
 }
+
+export const cloneUtilFunc = async (dash: Dashboard, id: string) => {
```
hoorayimhelping: we might want to make a ticket to take a look at this function, it's pretty gnarly

ui/src/dashboards/reducers/dashboards.ts
```diff
@@ -22,83 +42,66 @@ export const dashboardsReducer = (
...
-        const {dashboard} = action.payload
-        draftState.list = _.unionBy([dashboard], state.list, 'id')
+      case SET_DASHBOARD: {
+        const {schema} = action
+        const {entities, result} = schema
+
+        draftState.byID[result] = entities.dashboards[result]
+        const exists = draftState.allIDs.find(id => id === result)
+
+        if (!exists) {
```
hoorayimhelping: is there a chance, however small, that the id it finds could be `0`, and this could be falsy? or will it always be a stringy uuid?

ui/src/dashboards/actions/creators.ts
121watts: hahah.  i think its `f9` in vscode.

ui/src/dashboards/apis/index.ts
121watts: mayhaps indeed

ui/src/dashboards/reducers/dashboards.ts
121watts: nope.  they are all strings of the uuid v4

PR #16536, opened 2020-01-14
feat(ui) More reasonable aggregate window intervals for selected query durations

### LINKED ISSUE #16354: Use clean window period variables
Currently, when we calculate window period we do

https://github.com/influxdata/influxdb/blob/a2071478c804fdbaf8dbefdbdc013f63bd326973/ui/src/variables/utils/getWindowVars.ts#L66-L68

The result is that we end up sending a lot more data to the client than what we would expect.

We should have pre-defined bucket sizes like `10s`, `30s`,` 1m`, `5m`, `10m`, `30m`, `60m`, etc and round to the nearest depending on the duration of the query.

ebb-tide:
Closes #16354 

If user selects `auto` for aggregate window interval, in the Data Explorer or Cell Editor Overlay, the UI calculates an appropriate aggregate window interval for the query duration. The appropriate value is selected so that the UI doesn't request more points than can be displayed by the visulization. Before  this PR we would divide the query duration by the number of pixels that can be displayed- which would result in aggregate windows which were not rounded like 0.83s for a query range of 5m. In this PR the aggregate window intervals are rounded to "more reasonable" values. 
Along the way, the aggregate window interval input was changed to a duration input, so that users can select any valid duration, and so that the duration is properly labeled and indicated to user. 

| duration | window period | change to
|---|---|---|
|  5m | 0.83s | 10s |
| 15m  |  2.5s | 10s |
| 1h  |  10s | 10s |
| 6h  | 60s   | 1m| 
| 12h  | 120s  | 2m |
| 24h  | 240s  | 5m |
| 2d  | 480s  | 10m |
| 7d  | ?  | 30m |
| 30d  | ?  | 1h |

... 

Kapture 2020-01-15 at 18 57 50

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/src/variables/utils/getWindowVars.ts
+++ b/ui/src/variables/utils/getWindowVars.ts
@@ -10,6 +10,7 @@ import {WINDOW_PERIOD} from 'src/variables/constants'
 
 // Types
 import {VariableAssignment, Package} from 'src/types/ast'
+import {SELECTABLE_TIME_RANGES} from 'src/shared/constants/timeRanges'
 
 const DESIRED_POINTS_PER_GRAPH = 360
 const FALLBACK_WINDOW_PERIOD = 15000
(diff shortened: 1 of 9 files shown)


ui/src/variables/utils/getWindowVars.ts
```diff
@@ -63,7 +64,15 @@ export const getWindowPeriod = (
       files: [ast, buildVarsOption(variables)],
     }
 
-    const queryDuration = getMinDurationFromAST(substitutedAST)
+    const queryDuration = getMinDurationFromAST(substitutedAST) // in ms
+
+    const foundDuration = SELECTABLE_TIME_RANGES.find(
+      tr => tr.seconds * 1000 == queryDuration
```
hoorayimhelping: `getMinDurationFromAST` returns a number, i think this can be `===`

ebb-tide: thanks!

PR #16537, opened 2020-01-14
feat(cmd/influx): update password in cli

### LINKED ISSUE #16440: Add the ability to update a user's password from the CLI
as a user
when i use the update capability of the cli
then i would like the ability to update the password of the user

we should add `-p, --password` option to set the password of the user.

```
➜  influxdb git:(master) ✗ ./bin/darwin/influx user update -i 050c7357c0e83000 -h
Update user

Usage:
  influx user update [flags]

Flags:
  -h, --help          Help for the update command
  -i, --id string     The user ID (required)
  -n, --name string   The user name

Global Flags:
```

kelwang:
Closes #16440

Add the feature, so cli can update user's password. manually tested.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influx/setup.go
+++ b/cmd/influx/setup.go
@@ -153,7 +153,7 @@ func interactive() (req *platform.OnboardingRequest, err error) {
 	if setupFlags.password != "" {
 		req.Password = setupFlags.password
 	} else {
-		req.Password = getPassword(ui)
+		req.Password = getPassword(ui, false)
 	}
 	if setupFlags.token != "" {
 		req.Token = setupFlags.token
@@ -240,10 +240,14 @@ var (
 	errPasswordIsTooShort = fmt.Errorf("passwords is too short")
 )
 
-func getPassword(ui *input.UI) (password string) {
+func getPassword(ui *input.UI, showNew bool) (password string) {
+	newStr := ""
+	if showNew {
+		newStr = " new"
+	}
 	var err error
 enterPasswd:
-	query := promptWithColor("Please type your password", colorCyan)
+	query := promptWithColor("Please type your"+newStr+" password", colorCyan)
 	for {
 		password, err = ui.Ask(query, &input.Options{
 			Required:  true,
@@ -269,7 +273,7 @@ enterPasswd:
 		}
 		break
 	}
-	query = promptWithColor("Please type your password again", colorCyan)
+	query = promptWithColor("Please type your"+newStr+" password again", colorCyan)
 	for {
 		_, err = ui.Ask(query, &input.Options{
 			Required:  true,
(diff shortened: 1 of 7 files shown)


cmd/influx/setup.go
```diff
@@ -240,10 +240,14 @@ var (
 	errPasswordIsTooShort = fmt.Errorf("passwords is too short")
 )
 
-func getPassword(ui *input.UI) (password string) {
+func getPassword(ui *input.UI, showNew bool) (password string) {
```
jademcgough: If we passed the prompt string instead of using a boolean flag, it'd keep this additional logic out of the function.

kelwang: we have 2 promo strings, i think it's over complicated to pass them both. And a little weird to pass only one

cmd/influx/user.go
```diff
@@ -60,28 +103,84 @@ func newUserService() (platform.UserService, error) {
 	}, nil
 }
 
-func userUpdateF(cmd *cobra.Command, args []string) error {
-	s, err := newUserService()
+func newUserSVC() (
+	influxdb.UserService,
+	influxdb.OrganizationService,
+	influxdb.PasswordsService,
```
jsteenb2: you may want a dep struct for these dependencies, make it easier to maintain

```go
type CmdUserDeps struct {
    userSVC influxdb.UserService
    orgSvc influxdb.OrgService
   ... rest
}
```

cmd/influx/user_test.go
```diff
@@ -0,0 +1,383 @@
...
+						Name: "new name",
+					},
+				},
+			},
+			{
+				name: "with password",
+				flags: []string{
+					"--name=new name",
+					"--password=pass1",
+					"--org=org name",
```
jsteenb2: can you add some tests for the env var as part of the `--org` and `--org-id` API contract? Here are some examples from the org service that allow for env vars: https://github.com/influxdata/influxdb/blob/master/cmd/influx/organization_test.go#L247

cmd/influx/user_test.go
```diff
@@ -0,0 +1,383 @@
...
+		type called struct {
+			name string
+			id   influxdb.ID
+		}
+
+		tests := []struct {
+			name     string
+			expected called
+			flags    []string
+			envVars  map[string]string
```
jsteenb2: do we need the env vars here? they're never used it looks like

cmd/influx/setup.go
```diff
@@ -200,8 +200,9 @@ var (
 	keyReset    = []byte{keyEscape, '[', '0', 'm'}
 )
 
-func promptWithColor(s string, color []byte) string {
-	return string(color) + s + string(keyReset)
+func promptWithColor(s string, color []byte) []byte {
```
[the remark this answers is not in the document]
kelwang: ah, it's the convert of fmt.Println to the ui.Write, which took the bytes slice instead.

PR #16541, opened 2020-01-14
fix(pkger): fixes issue where pkger can't discern the mime/content type of a package

jsteenb2:
also makes the yaml decoder the default. To foten we end up in application/octet-stream
which is the default for many different mime types. This provides a mechanism
around that so that when the automagical detection fails it can allow the user
to provide it via the CLI.

Closes issues I noticed during testing since the url bits got merged

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/pkger/parser.go
+++ b/pkger/parser.go
@@ -162,7 +162,7 @@ func parseSource(r io.Reader, opts ...ValidateOptFn) (*Pkg, error) {
 		strings.Contains(contentType, "yml"):
 		return parseYAML(bytes.NewReader(b), opts...)
 	default:
-		return parseJsonnet(bytes.NewReader(b), opts...)
+		return parseYAML(bytes.NewReader(b), opts...)
 	}
 }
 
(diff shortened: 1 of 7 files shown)


pkger/parser.go
```diff
@@ -162,7 +162,7 @@ func parseSource(r io.Reader, opts ...ValidateOptFn) (*Pkg, error) {
 		strings.Contains(contentType, "yml"):
 		return parseYAML(bytes.NewReader(b), opts...)
 	default:
-		return parseJsonnet(bytes.NewReader(b), opts...)
+		return parseYAML(bytes.NewReader(b), opts...)
```
jsteenb2: this may end up changing to something else in the future. But right now most folks are using yml, and if the content type is failed to detect or detected as `application/octet-stream` then feels best to fall back to yaml for the time being until we have metrics that tell us we have a different encoding that is more common and we can switch it to be that one if desired.

PR #16542, opened 2020-01-14
fix(storage): Offline Series Compaction

benbjohnson:
This pull request adds the ability to offline compact existing series files into the `influx_inspect buildtsi` tool:

```sh
$ influx_inspect buildtsi -compact-series-file
```

It iterates over all series in each segment and rewrites all series that have not been tombstoned in the index to a new `.tmp` file next to the segment. Once all segments are converted, the temporary files overwrite the original segments.

The included benchmark shows that compacting 1M series takes only about 63ms:

```sh
% go test -bench=BenchmarkSeriesFile_Compaction -run=XXX -v ./tsdb
goos: darwin
goarch: amd64
pkg: github.com/influxdata/influxdb/tsdb
BenchmarkSeriesFile_Compaction-16    	      21	  62932487 ns/op
PASS
ok  	github.com/influxdata/influxdb/tsdb	14.363s
```

- [ ] Rebased/mergeable
- [ ] Tests pass


cmd/influx_inspect/buildtsi/buildtsi.go
```diff
@@ -111,6 +131,126 @@ func (cmd *Command) run(dataDir, walDir string) error {
...
+	if err = sfile.Open(); err != nil {
+		return err
+	}
+
+	compactor := tsdb.NewSeriesPartitionCompactor()
+	for _, partition := range sfile.Partitions() {
+		if err = compactor.Compact(partition); err != nil {
+			return err
+		}
+		fmt.Println("compacted ", partition.Path())
```
e-dard: Should we use the struct's stdout writer for this?

cmd/influx_inspect/buildtsi/buildtsi.go
```diff
@@ -111,6 +131,126 @@ func (cmd *Command) run(dataDir, walDir string) error {
...
+	defer index.Close()
+
+	for _, fi := range fis {
+		segmentID, err := tsdb.ParseSeriesSegmentFilename(fi.Name())
+		if err != nil {
+			continue // skip non-segment file.
+		}
+
+		segmentPath := filepath.Join(path, fi.Name())
+		fmt.Printf("processing segment %q %d\n", path, segmentID)
```
e-dard: Should we use the struct's stdout writer for this?

cmd/influx_inspect/buildtsi/buildtsi.go
```diff
@@ -111,6 +131,126 @@ func (cmd *Command) run(dataDir, walDir string) error {
...
+		}
+		fmt.Println("compacted ", partition.Path())
+	}
+	return nil
+}
+
+func (cmd *Command) compactSeriesFilePartition(path string) error {
+	const tmpExt = ".tmp"
+
+	fmt.Printf("processing partition for %q\n", path)
```
e-dard: Should we use the struct's stdout writer for this?

benbjohnson:
@e-dard I'm testing on an 8-core i9 and it's running in parallel. I was surprised by the speed too though. I added a bunch of `println()` statements as I was benchmarking and it seems to be legit doing real work. 

I'm OOO today and tomorrow but I'll add some additional test coverage. I did hit a bug with the use of `FindOffsetByID()` (which doesn't check tombstones) vs `IsDeleted()` (which does check tombstones).

I'm not concerned about the sizing being off. That was originally added so that we could create a file with a fixed size for Windows.

benbjohnson:
Added test coverage for the series compaction. I'm getting a CI error from _Semantic Pull Request_ but I can't figure out why.

dgnorton:
@benbjohnson the PR title also has to be in the same format

e-dard:
@8none1 I think we should get some real-world testing of this under our belts before 1.8 ships. Perhaps by taking a real workload and verifying the tool does the right thing.

PR #16548, opened 2020-01-15
feat(query): add 'Prefer: return-no-content-with-error' behavior

affo:
kept both `no-content` and `...-with-error`.

This also adds proper testing.

### DIFF
--- a/query/encode.go
+++ b/query/encode.go
@@ -5,14 +5,23 @@ import (
 	"net/http"
 
 	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/csv"
 )
 
-const DialectType = "no-content"
+const (
+	NoContentDialectType     = "no-content"
+	NoContentWErrDialectType = "no-content-with-error"
+)
 
-// AddDialectMappings adds the no-content dialect mapping.
+// AddDialectMappings adds the no-content dialects mappings.
 func AddDialectMappings(mappings flux.DialectMappings) error {
-	return mappings.Add(DialectType, func() flux.Dialect {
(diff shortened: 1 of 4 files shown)


query/encode.go
```diff
@@ -5,14 +5,23 @@ import (
...
 )
 
-const DialectType = "no-content"
+const (
+	NoContentDialectType     = "no-content"
+	NoContentWErrDialectType = "no-content-with-error"
+)
 
-// AddDialectMappings adds the no-content dialect mapping.
+// AddDialectMappings adds the no-content dialects mappings.
```
stuartcarnie:
```suggestion
// AddDialectMappings adds the mappings for the no-content dialects.
```

query/encode.go
```diff
@@ -58,3 +65,70 @@ func (e *NoContentEncoder) Encode(w io.Writer, results flux.ResultIterator) (int
...
+// but it encodes runtime errors from the Flux query in CSV format.
+// To discover if there was any runtime error in the query, one should check the response size.
+// If it is equal to zero, then no error was present.
+// Otherwise one can decode the response body to get the error. For example:
+// ```
+// _, err = csv.NewResultDecoder(csv.ResultDecoderConfig{}).Decode(bytes.NewReader(res))
+// if err != nil {
+//   // we got some runtime error
+// }
+// ```
```
stuartcarnie: This is excellent!

query/encode.go
affo: I'll do this in next PR, to avoid  CI steps 👍

Issue #16468, opened 2020-01-09
 Make Makefiles make again


hoorayimhelping:
Our Makefiles are a bit of a mess. They need to be cleaned up and the main guidelines are:

* They should be the main and official interface into the app's building, testing and scripting processes
* They shouldn't be recursive, one makefile is enough for the whole project
* User friendliness and clear messaging is ideal. Someone unfamiliar with the project should be able to run `make` and go from there

drdelambre:
ask: can we silence `yarn install` on this one? Gunnar gave feedback that the output of that command (with all of its unactionable warnings) gives an unclear picture into the process from a user's perspective (not FE dev).

russorat:
we moved the UI to it's own repo. the current make command works in this repo.

PR #16551, opened 2020-01-15
ci: bail webpack build on tsc error

121watts:
### The Problem

Firstly, webpack did not exit with a non-zero status code if there was a TypeScript compilation error.  Secondly, the Makefile `subdirs` target loop did not consider the exit codes of each sub-make, so it didn't stop. Therefore, only the final status code bubbled up in CI.  See [this for more details](https://stackoverflow.com/questions/16492512/how-to-handle-the-sub-make-in-gnu-make-errors).

### Solution
Copy answers from stack overflow.

Related: #16468


drdelambre:
@zoesteinkamp is working on that flakey test

drdelambre:
whoops

PR #16554, opened 2020-01-15
fix: uuid security

lyondhill:
Update to remove dependency on insecure version of uuid library
https://snyk.io/vuln/SNYK-GOLANG-GITHUBCOMSATORIGOUUID-72488

### DIFF
--- a/chronograf/id/uuid.go
+++ b/chronograf/id/uuid.go
@@ -12,5 +12,9 @@ type UUID struct{}
 
 // Generate creates a UUID v4 string
 func (i *UUID) Generate() (string, error) {
-	return uuid.NewV4().String(), nil
+	uuid, err := uuid.NewV4()
+	if err != nil {
+		return "", err
+	}
+	return uuid.String(), nil
 }


chronograf/id/uuid.go
```diff
@@ -12,5 +12,5 @@ type UUID struct{}
 
 // Generate creates a UUID v4 string
 func (i *UUID) Generate() (string, error) {
-	return uuid.NewV4().String(), nil
+	return uuid.Must(uuid.NewV4()).String(), nil
```
jsternberg: Can you have this return the error instead of panicking?

jsternberg:
It seems like all circle tests are passing for this. If the only thing keeping this from merging is the api compatibility, I'll merge this and we'll start the process of merging the flux staging branch. The api compatibility tests are rerunning right now so I'll merge regardless of the results. If further changes are needed to get the api compatibility tests passing, we'll do that on the flux staging branch as a separate PR.

jsternberg:
The error seems to be this:

```
INFO Data retention check (end) {"test_name": "TestPipeline_Retention", "storage_node": 0, "db_shard_id": 0, "node_id": 0, "engine_id": 0, "service": "storage-engine", "component": "retention_enforcer", "op_name": "data_retention_check", "trace_id": "5b9f3950a747943c", "op_event": "end", "op_elapsed": "51.989018ms"}
query_helpers.go:281: failed to read metadata: missing expected annotations datatype, group, and default
```

It doesn't seem related to this change. Since this is not going to master, I am merging this.

PR #16559, opened 2020-01-16
feat(task): make task query without returning content

affo:
This depends on https://github.com/influxdata/influxdb/pull/16548.

Once  that gets fixed this can land.

The relevant commit is the second one.


query/request.go
```diff
@@ -4,11 +4,19 @@ import (
...
 
 	"github.com/influxdata/flux"
 	platform "github.com/influxdata/influxdb"
 )
 
+const (
+	PreferHeaderKey                = "Prefer"
+	PreferNoContentHeaderValue     = "return-no-content"
+	PreferNoContentWErrHeaderValue = "return-no-content-with-error"
+)
```
affo: Had to move these ones wrt the first commit because of dependency cycles.

http/query.go
```diff
@@ -372,10 +366,10 @@ func decodeQueryRequest(ctx context.Context, r *http.Request, svc influxdb.Organ
...
 
-	switch hv := r.Header.Get(PreferHeaderKey); hv {
-	case PreferNoContentHeaderValue:
+	switch hv := r.Header.Get(query.PreferHeaderKey); hv {
+	case query.PreferNoContentHeaderValue:
 		req.PreferNoContent = true
-	case PreferNoContentWErrHeaderValue:
+	case query.PreferNoContentWErrHeaderValue:
 		req.PreferNoContentWithError = true
 	}
```
stuartcarnie: These cyclic dependencies are unfortunate. Ideally we'd have a the http related constants in the `http` package. If there is no easy way to break that, we can proceed with this PR and identify the problem areas in a future PR.

affo: Yeah, I agree 👍 

The cycle is `query -> http -> authorizer -> task -> query` and I don't see any obvious way to break that for now 😓 

It seemed easy, because `authorizer` depends on `task/backend` only for constants, but when I removed them, the cycle only changed...

affo:
@stuartcarnie just applied your suggested change to the comment  above `AddDialectMappings` 👍

PR #16564, opened 2020-01-16
fix: Fixing master gauge issues

zoesteinkamp:
Closes #

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/src/shared/utils/formatStatValue.ts
+++ b/ui/src/shared/utils/formatStatValue.ts
@@ -45,7 +45,7 @@ export const formatStatValue = (
     return 'Data cannot be displayed'
   }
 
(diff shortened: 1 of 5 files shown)


ui/src/shared/utils/formatStatValue.ts
```diff
@@ -45,7 +45,7 @@ export const formatStatValue = (
     return 'Data cannot be displayed'
   }
 
-  localeFormattedValue = preventNegativeZero(localeFormattedValue)
+  localeFormattedValue = '' + preventNegativeZero(localeFormattedValue)
```
TCL735: Why do we need this as a string? The return value of this function is a different variable which is already a string.

PR #16567, opened 2020-01-16
fix(ui): make localeFormattedValue less confusing

TCL735:
Fixes issues related to recent PRs.

- localeFormattedValue does not need to be coerced, because it is not the return value of the function
- run prettier


TCL735:
It will pass the checks, so master should be fine. (What's gonna happen though is that e2e will flake out after merging to master, just to make me look bad. LOL. circleci loves to stab people in the back.)

PR #16589, opened 2020-01-17
fix(test): Fixing flakey e2e onboarding test in cypress

### LINKED ISSUE #16446: Flakey Onboarding Test
The test 'Can onboard to configure later' in the onboarding cypress test file is not correct anymore. Going to go update so it no longer fails

zoesteinkamp:
Closes #16446 

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/cypress/e2e/onboarding.test.ts
+++ b/ui/cypress/e2e/onboarding.test.ts
@@ -18,7 +18,7 @@ describe('Onboarding', () => {
     cy.visit('/')
   })
 
-  it.skip('Can Onboard to Quick Start', () => {
+  it('Can Onboard to Quick Start', () => {
     cy.server()
 
     //Will want to capture response from this
--- a/ui/cypress/e2e/onboarding.test.ts
+++ b/ui/cypress/e2e/onboarding.test.ts
@@ -111,11 +111,13 @@ describe('Onboarding', () => {
       //advance to Quick Start
       cy.getByTestID('button--quick-start').click()
 
+      cy.wait(20000)
+
       cy.location('pathname').should('equal', '/orgs/' + orgId)
     })
   })
 
-  it.skip('Can onboard to advanced', () => {
+  it('Can onboard to advanced', () => {
     cy.server()
 
     cy.route('POST', 'api/v2/setup').as('orgSetup')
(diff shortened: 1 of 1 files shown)


zoesteinkamp:
> i reject this solution unless you can tell me why a timeout is more robust than looking for dom elements to be available in this situation

We can look for DOM elements, we are currently confirming the route is correct, but we can do this as well. But the underlying problem will not be resolved, as what is currently happening is a time out response, its saying hey i dont see this route you claim we should be on, guess we didnt make it "FAIL". When really it just needs more time to load up the route. Some routes are particularly more heavy and ask for way more info which can take awhile to retrieve. 

BUT, i also dont like this as a solution either. Its a solution presented in the cypress docs and the cypress forms. I think we could do something better. But i think this is an okay band aid.

ui/cypress/e2e/onboarding.test.ts
```diff
@@ -5,6 +5,23 @@ interface TestUser {
...
+    cy.flush()
+
+    cy.fixture('user').then(u => {
+      user = u
+    })
+
+    cy.visit('/')
+  })
+  it('Can redirect to onboarding page', () => {
+    cy.get('[data-testid="init-step--head-main"]', {timeout: 5000})
```
121watts:
```suggestion
    cy.getByTestID("init-step--head-main")
```
You also shouldn't need the timeout as Cypress will already wait 30 seconds to find any element.

ui/cypress/e2e/onboarding.test.ts
```diff
@@ -5,6 +5,23 @@ interface TestUser {
...
+  beforeEach(() => {
+    cy.flush()
+
+    cy.fixture('user').then(u => {
+      user = u
+    })
+
+    cy.visit('/')
+  })
+  it('Can redirect to onboarding page', () => {
```
121watts: Can we add some space for these to breathe?
```suggestion

  it('Can redirect to onboarding page', () => {
```

121watts:
also, to verify this changes flakiness you can wrap any test in a for loop and have it run `n` times.

example:

```
for (let n = 0; n < 50; n++ {
  it('does a thing, () => {
   ....
  })
}
```

ui/cypress/e2e/onboarding.test.ts
```diff
@@ -111,11 +111,13 @@ describe('Onboarding', () => {
       //advance to Quick Start
       cy.getByTestID('button--quick-start').click()
 
+      cy.wait(20000)
```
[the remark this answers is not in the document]
zoesteinkamp: This has been removed, it was for testing and snuck into one of the commits

ui/cypress/e2e/onboarding.test.ts
zoesteinkamp: got it

ui/cypress/e2e/onboarding.test.ts
zoesteinkamp: space added!

PR #16599, opened 2020-01-18
feat(influxdb): Add path to cli

### LINKED ISSUE #10294: Make default InfluxDB client support custom HTTP endpoints
__Proposal:__

Make it possible to connect to an InfluxDB instance running behind a reverse proxy with a custom subpath / endpoint with the default InfluxDB client `influx`.

__Current behavior:__

I have InfluxDB, Grafana and nginx running as docker containers in a test VM at `192.168.122.26`.
InfluxDB is exposed at `192.168.122.26/influxdb/`

I can issue a ping command with `curl` but I can't log in using `influx`:
```shell
$ curl -sL -I l http://192.168.122.26/influxdb/ping
HTTP/1.1 204 No Content
Server: nginx/1.14.0
Date: Thu, 20 Sep 2018 11:21:11 GMT
Content-Type: application/json
Connection: keep-alive
Request-Id: 4714b565-bcc7-11e8-8009-000000000000
X-Influxdb-Build: OSS
X-Influxdb-Version: 1.5.4
X-Request-Id: 4714b565-bcc7-11e8-8009-000000000000

$ influx -host 192.168.122.26/influxdb -port 80
Failed to connect to http://192.168.122.26%2Finfluxdb: parse http://192.168.122.26%2Finfluxdb/ping: invalid URL escape "%2F"
Please check your connection settings and ensure 'influxd' is running.
```

__Desired behavior:__

I'd like to be able to connect to a custom endpoint with `influx`. :)

__Use case:__

I'm just getting started with InfluxDB but I'd like to be able to test the database with `influx` as the server is set up now. I might go back to just exposing the server directly via its default port and no endpoint instead if necessary.

bthesorceror:
Closes #10294

### DIFF
--- a/client/influxdb.go
+++ b/client/influxdb.go
@@ -31,6 +31,7 @@ const (
 
 	// DefaultTimeout is the default connection timeout used to connect to an InfluxDB instance
 	DefaultTimeout = 0
+	DefaultPath    = ""
 )
 
 // Query is used to send a command to the server. Both Command and Database are required.
@@ -64,23 +65,36 @@ type Query struct {
 	NodeID int
 }
 
+func splitPath(v string) (string, string) {
+	parts := strings.Split(v, "/")
+
+	first := parts[0]
+	last := strings.Join(parts[1:], "/")
+
+	return first, last
+}
+
 // ParseConnectionString will parse a string to create a valid connection URL
 func ParseConnectionString(path string, ssl bool) (url.URL, error) {
 	var host string
 	var port int
+	var pth string = ""
 
 	h, p, err := net.SplitHostPort(path)
 	if err != nil {
 		if path == "" {
 			host = DefaultHost
 		} else {
-			host = path
+			host, pth = splitPath(path)
 		}
 		// If they didn't specify a port, always use the default port
 		port = DefaultPort
 	} else {
 		host = h
-		port, err = strconv.Atoi(p)
+		prt, pt := splitPath(p)
+		pth = pt
+
+		port, err = strconv.Atoi(prt)
 		if err != nil {
 			return url.URL{}, fmt.Errorf("invalid port number %q: %s\n", path, err)
 		}
@@ -89,6 +103,7 @@ func ParseConnectionString(path string, ssl bool) (url.URL, error) {
 	u := url.URL{
 		Scheme: "http",
 		Host:   host,
+		Path:   pth,
 	}
 	if ssl {
 		u.Scheme = "https"
--- a/client/influxdb.go
+++ b/client/influxdb.go
@@ -65,13 +65,13 @@ type Query struct {
 	NodeID int
 }
 
-func splitPath(v string) (string, string) {
-	parts := strings.Split(v, "/")
-
-	first := parts[0]
-	last := strings.Join(parts[1:], "/")
-
-	return first, last
+// SplitPath gets the path of a url
+func SplitPath(v string) (string, string) {
+	i := strings.Index(v, "/")
+	if i == -1 {
+		return v, ""
+	}
+	return v[:i] /* first */, v[i+1:] /* rest */
 }
 
 // ParseConnectionString will parse a string to create a valid connection URL
@@ -85,13 +85,13 @@ func ParseConnectionString(path string, ssl bool) (url.URL, error) {
 		if path == "" {
 			host = DefaultHost
 		} else {
-			host, pth = splitPath(path)
+			host, pth = SplitPath(path)
 		}
 		// If they didn't specify a port, always use the default port
 		port = DefaultPort
 	} else {
 		host = h
-		prt, pt := splitPath(p)
+		prt, pt := SplitPath(p)
 		pth = pt
 
 		port, err = strconv.Atoi(prt)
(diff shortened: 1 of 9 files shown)


cmd/influx/cli/cli.go
```diff
@@ -114,6 +116,14 @@ func (c *CommandLine) Run() error {
 		c.ClientConfig.Password = os.Getenv("INFLUX_PASSWORD")
 	}
 
+	addr := fmt.Sprintf("%s:%d/%s", c.Host, c.Port, c.Path)
```
jademcgough: is this going to default the port to 0 if not set, or does port always have a default value?

bthesorceror: it should always have a default at this point as there is a default for the cli params

client/influxdb.go
```diff
@@ -64,23 +65,36 @@ type Query struct {
 	NodeID int
 }
 
+func splitPath(v string) (string, string) {
+	parts := strings.Split(v, "/")
```
jademcgough: Would it be safer to use `url.Parse` for this?

client/influxdb.go
```diff
@@ -64,23 +65,36 @@ type Query struct {
...
 	var port int
+	var pth string = ""
 
 	h, p, err := net.SplitHostPort(path)
 	if err != nil {
 		if path == "" {
 			host = DefaultHost
 		} else {
-			host = path
+			host, pth = splitPath(path)
```
jademcgough: Can probably use `url.Parse` to get this information without manually splitting the string. Might be safer / more readable.

bthesorceror: it won't always be a valid url because it could be `host/path` or `port/path` in which case parse would fail with an error.

cmd/influx/main.go
```diff
@@ -40,6 +40,7 @@ func main() {
 
 	fs := flag.NewFlagSet("InfluxDB shell version "+version, flag.ExitOnError)
 	fs.StringVar(&c.Host, "host", client.DefaultHost, "Influxdb host to connect to.")
+	fs.StringVar(&c.Path, "host", client.DefaultPath, "Influxdb host to connect to.")
```
jademcgough: description is a duplicate of the previous description

query/internal/gota/cmo_test.go
```diff
@@ -15,7 +15,7 @@ func TestCMO(t *testing.T) {
 		}
 	}
 
-	if diff := diffFloats(expList, actList, 1E-7); diff != "" {
+	if diff := diffFloats(expList, actList, 1e-7); diff != "" {
```
jademcgough: were these automatically changed by the formatter?

cmd/influx/main.go
jademcgough: should be "path" instead of "host", as well

query/internal/gota/cmo_test.go
bthesorceror: yes

ayang64:
in general i feel like there might be a better name for what we're doing here.  instead of `path` -- which collides with the `path` package and isn't very clear -- i think i'd use `urlprefix`.

client/influxdb.go
dgnorton: The client allows connect strings like `localhost:8086` that `url.Parse` can't parse but also won't error. Relevant blurb from Go docs:

> Trying to parse a hostname and path without a scheme is invalid but may not necessarily return an error, due to parsing ambiguities.

bthesorceror: @dgnorton I am curious why we allow host and port together when it is not explicitly stated in the host flag usage? This could be a lot cleaner if we had `host` `port` `urlpath` flags and a `connection string` flag for the combination of the 3

bthesorceror:
@ayang64 I am open to a different name but I think urlprefix is a little confusing because its really a postfix

client/influxdb.go
jademcgough: I think we allow them together for the `connect` command (allowing you to connect to another influxdb from within the CLI)

jademcgough:
@bthesorceror What about `-path-prefix`?

client/influxdb.go
bthesorceror: then connection string and the connect command could share logic while allowing the other flags to be independent, there is a bit of tangle logic that could be cleaned up by treating it this way I think. Right now I could supply a port from the `port` flag or maybe I give it in the `host` where as if we had a `connection string` flag we could say, if this is present get port from it and ignore other flags (maybe even give invalid options error if port and connection string are given together.

bthesorceror: maybe instead of `connection string` we say `url` and then we can use `url.Parse` on it

dgnorton: A single command line connection flag that is flexible would be nice:
`http://localhost:8086` (default)
`localhost`
`:9086`
`http://somewhere.com`
`http://somewhere.com:9096`
`http://somewhere.com/influxdb`
`https://somewhere.com:9086/influxdb`
etc.

Need to ensure existing flags continue to work as before. Some users have scripts that depend on existing behavior.

PR #16621, opened 2020-01-22
fix(tsm1): improve series cardinality limit

dgnorton:
Prior to this change, new series would be added to the series file
before checking the series cardinality limit. If the limit was exceeded,
the write was rejected even though the series had already been added to
the series file.

### DIFF
--- a/tsdb/index/inmem/inmem.go
+++ b/tsdb/index/inmem/inmem.go
@@ -189,13 +189,23 @@ func (i *Index) MeasurementIterator() (tsdb.MeasurementIterator, error) {
 func (i *Index) CreateSeriesListIfNotExists(seriesIDSet *tsdb.SeriesIDSet, measurements map[string]int,
 	keys, names [][]byte, tagsSlice []models.Tags, opt *tsdb.EngineOptions, ignoreLimits bool) error {
 
+	// Verify that the series will not exceed limit.
+	if !ignoreLimits {
+		i.mu.RLock()
+		if max := opt.Config.MaxSeriesPerDatabase; max > 0 && len(i.series)+len(keys) > max {
+			i.mu.RUnlock()
+			return errMaxSeriesPerDatabaseExceeded{limit: opt.Config.MaxSeriesPerDatabase}
+		}
+		i.mu.RUnlock()
+	}
+
 	seriesIDs, err := i.sfile.CreateSeriesListIfNotExists(names, tagsSlice)
 	if err != nil {
 		return err
 	}
 
 	i.mu.RLock()
-	// if there is a series for this ids, it's already been added
+	// If there is a series for this ID, it's already been added.
 	seriesList := make([]*series, len(seriesIDs))
 	for j, key := range keys {
 		seriesList[j] = i.series[string(key)]
@@ -256,13 +266,6 @@ func (i *Index) CreateSeriesListIfNotExists(seriesIDSet *tsdb.SeriesIDSet, measu
 	}
 	if newSeriesN == 0 {
 		return nil
-	}
-
-	// Verify that the series will not exceed limit.
-	if !ignoreLimits {
-		if max := opt.Config.MaxSeriesPerDatabase; max > 0 && len(i.series)+len(keys) > max {
-			return errMaxSeriesPerDatabaseExceeded{limit: opt.Config.MaxSeriesPerDatabase}
-		}
 	}
 
 	for j, key := range keys {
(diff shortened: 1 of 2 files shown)


tsdb/index/inmem/inmem.go
```diff
@@ -189,13 +189,23 @@ func (i *Index) MeasurementIterator() (tsdb.MeasurementIterator, error) {
 func (i *Index) CreateSeriesListIfNotExists(seriesIDSet *tsdb.SeriesIDSet, measurements map[string]int,
 	keys, names [][]byte, tagsSlice []models.Tags, opt *tsdb.EngineOptions, ignoreLimits bool) error {
 
+	// Verify that the series will not exceed limit.
+	if !ignoreLimits {
+		i.mu.RLock()
+		if max := opt.Config.MaxSeriesPerDatabase; max > 0 && len(i.series)+len(keys) > max {
```
ayang64: i like simplifying `opt.Config.MaxSeriesPerDatabase` to `max`.  i feel like similarly simplifying `len(i.series)+len(keys)` would make the expression simpler. so instead of:
```
if max := opt.Config.MaxSeriesPerDatabase; max > 0 && len(i.series)+len(keys) > max {
```

maybe:

```
if max, numSeries := opt.Config.MaxSeriesPerDatabase, len(i.series)+len(keys); max > 0 && numSeries > max {
```
i feel like that clarifies the expression.  i'm not sure if `numSeries` is the right name for it but that's the gist.

tsdb/index/inmem/inmem.go
```diff
@@ -189,13 +189,23 @@ func (i *Index) MeasurementIterator() (tsdb.MeasurementIterator, error) {
 func (i *Index) CreateSeriesListIfNotExists(seriesIDSet *tsdb.SeriesIDSet, measurements map[string]int,
 	keys, names [][]byte, tagsSlice []models.Tags, opt *tsdb.EngineOptions, ignoreLimits bool) error {
 
+	// Verify that the series will not exceed limit.
+	if !ignoreLimits {
+		i.mu.RLock()
```
ayang64: seems like there might be a race between here and line 201 where the series might exceed the limit after this check.

it seems like everything between lines 192 and 212 should be in a read lock?

tsdb/index/inmem/inmem.go
```diff
@@ -189,13 +189,23 @@ func (i *Index) MeasurementIterator() (tsdb.MeasurementIterator, error) {
 func (i *Index) CreateSeriesListIfNotExists(seriesIDSet *tsdb.SeriesIDSet, measurements map[string]int,
 	keys, names [][]byte, tagsSlice []models.Tags, opt *tsdb.EngineOptions, ignoreLimits bool) error {
 
+	// Verify that the series will not exceed limit.
+	if !ignoreLimits {
```
ayang64: since we're returning from here and we have to track multiple unlocks, maybe this is better as a separate function or a function literal so that we can use defer to ensure we cleanup like:

```

	if !ignoreLimits {
		i.mu.RLock()
		if max := opt.Config.MaxSeriesPerDatabase; max > 0 && len(i.series)+len(keys) > max {
			i.mu.RUnlock()
			return errMaxSeriesPerDatabaseExceeded{limit: opt.Config.MaxSeriesPerDatabase}
		}
		i.mu.RUnlock()
	}
```

tsdb/index/inmem/inmem.go
dgnorton: Initially, I didn't have the `RLock()` here because the code was moved up from a later part of the function where it had been inside a lock. The race detector caught the race and I added the read lock here. I re-ran the tests multiple times with no race detected. I know the race detector can only report races that actually occur (i.e., no false positives but can have false negatives). I don't think there's a race in the sense that two goroutines might simultaneously access `i.series` but there is a known condition that two or more simultaneous writes might make it past this check and put the database past the limit. The limit isn't an exact limit. We sacrifice exact limit enforcement for performance.

I agree that we could cleanup those bits of code but this is a backport so it's best not to make cosmetic changes here. We should make them on the `1.8` branch first  and then backport.

tsdb/index/inmem/inmem.go
dgnorton: See response above.

tsdb/index/inmem/inmem.go
dgnorton: I initially had this wrapped in a func with a `defer` but it added more lines of code. This is also performance sensitive code executes with every write and `defer` has a cost. We'd also need to make this change on `1.8` branch first and then backport.

ayang64:
LGTM

PR #16627, opened 2020-01-22
chore(tsm1): skip WriteSnapshot during backup if snapshotter is busy

gianarb:
When an InfluxDB database is very busy writing new points the backup
the process can fail because it can not write a new snapshot.

The error is: `operation timed out with error: create snapshot: snapshot in progress`.

This happens because InfluxDB takes almost "continuously" a snapshot
from the cache caused by the high number of points ingested.

This PR skips snapshots if the `snapshotter` does not come available
after three attempts when a backup is requested.

The backup won't contain the data in the cache or WAL.

Signed-off-by: Gianluca Arbezzano <gianarb92@gmail.com>

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -912,9 +912,22 @@ func (e *Engine) Free() error {
 // backup is running. For shards that are still acively getting writes, this
 // could cause the WAL to backup, increasing memory usage and evenutally rejecting writes.
 func (e *Engine) Backup(w io.Writer, basePath string, since time.Time) error {
-	path, err := e.CreateSnapshot()
-	if err != nil {
-		return err
+	var err error
+	var path string
+	for i := 0; i < 3; i++ {
+		path, err = e.CreateSnapshot()
+		if err != nil {
+			switch err {
+			case ErrSnapshotInProgress:
+				backoff := time.Duration(math.Pow(32, float64(i))) * time.Millisecond
+				time.Sleep(backoff)
+			default:
+				return err
+			}
+		}
+	}
+	if err == ErrSnapshotInProgress {
+		e.logger.Warn("Snapshotter busy: Backup proceeding without snapshot contents.")
 	}
 	// Remove the temporary snapshot dir
 	defer os.RemoveAll(path)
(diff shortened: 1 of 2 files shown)


tsdb/engine/tsm1/engine.go
```diff
@@ -1893,8 +1893,22 @@ func (e *Engine) WriteSnapshot() (err error) {
...
 func (e *Engine) CreateSnapshot() (string, error) {
-	if err := e.WriteSnapshot(); err != nil {
-		return "", err
+	var err error
+	for i := 0; i < 3; i++ {
+		err = e.WriteSnapshot()
+		if err != nil {
+			switch err {
+			case ErrSnapshotInProgress:
+				backoff := time.Duration(math.Pow(3.8, float64(i))) * time.Millisecond
```
dgnorton: Not sure what the best backoff would be. 3.8 seems low. That's 14.4ms on the 3rd iteration. Snapshots are generally < 1s but can be longer even in a healthy system sometimes. Maybe 32 instead of 3.8? That would top out at ~1s.

tsdb/engine/tsm1/engine.go
```diff
@@ -1893,8 +1893,22 @@ func (e *Engine) WriteSnapshot() (err error) {
 // CreateSnapshot will create a temp directory that holds
 // temporary hardlinks to the underylyng shard files.
 func (e *Engine) CreateSnapshot() (string, error) {
-	if err := e.WriteSnapshot(); err != nil {
-		return "", err
+	var err error
```
e-dard: Question - is this only used for backups?

gianarb: For what I can tell this is the unique place where we use it

https://github.com/influxdata/influxdb/blob/a56e0226e61bbcfca88b1e0012e27463d10c0fe4/storage/engine.go#L674

So yes

e-dard: I think `CreateSnapshot()` is called from:

https://github.com/influxdata/influxdb/blob/1.8/tsdb/shard.go#L1125
https://github.com/influxdata/influxdb/blob/1.8/tsdb/engine/tsm1/engine.go#L981
https://github.com/influxdata/influxdb/blob/1.8/tsdb/engine/tsm1/engine.go#L915

tsdb/engine/tsm1/engine.go
```diff
@@ -912,9 +912,19 @@ func (e *Engine) Free() error {
...
+		path, err = e.CreateSnapshot()
+		if err != nil {
+			switch err {
+			case ErrSnapshotInProgress:
+				backoff := time.Duration(math.Pow(32, float64(i))) * time.Millisecond
+				time.Sleep(backoff)
+			default:
+				return err
+			}
+		}
```
e-dard: One more tiny nit. Since technically after three attempts we will be potentially ignoring an error we should still handle it. In this case we should log a warning to say that we couldn't snapshot the cache.

PR #16629, opened 2020-01-22
fix: moving an endpoint over to 201

### LINKED ISSUE #16386: PATCH to '/dashboards/{dashboardID}/cells/{cellID}' returns 201 instead of 200 as swagger specifies
swagger line 2451 says a successful PATCH should return a 200, and UI expects a 200, but successful PATCH returns 201 and causes UI to throw an error on the console when trying to resize a cell even though PATCH has succeeded.

drdelambre:
Closes #16386

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -2448,7 +2448,7 @@ paths:
           required: true
           description: The ID of the cell to update.
       responses:
-        '200':
+        '201':
           description: Updated dashboard cell
           content:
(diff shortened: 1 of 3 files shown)


jsteenb2:
fix is in server only here yah?

drdelambre:
flippin auto assigners! i was trying to share something with deniz and the whole world came crashing down upon me

Issue #15530, opened 2019-10-22
[UI - Templates] user templates list is not updating reliably after import template


karel-rehor:
__Steps to reproduce:__
List the minimal actions needed to reproduce the behavior.

This issue is showing up intermittently, so try the following steps maybe a dozen times.  It shows up in one out of every five or six runs of import template tests. 

1. open settings > templates
2. click "user templates"
3. click import template in the list header (top right)
4. in the system dialog browse to a valid template.json file and upload it
5. click the submit button (page reloads to static templates - issue #15514)
6. click "user templates" again (work around for #15514)

__Expected behavior:__
Expect to see reliably (99.9% of the time) the template just uploaded. 

__Actual behavior:__
About one in six times the newly uploaded template fails to appear in the list of user templates. 

Clicking the browser refresh button seems to work to solve this manually, however in the automated tests "refresh()" has not proven a reliable solution. 

__Screencast__
![ImportTemplateStaleList01](https://user-images.githubusercontent.com/25881301/67302642-7a708180-f4f1-11e9-8243-338ae763956f.gif)

__Environment info:__

* System info: Run `uname -srm` and copy the output here
```
Linux 4.4.0-166-generic x86_64
```
* InfluxDB version: Run `influxd version` and copy the output here
Detected in nightly build docker image 
```
ts=2019-10-22T08:09:53.872645Z lvl=info msg="Welcome to InfluxDB" log_id=0Idd5Bp0000 version=nightly commit=1e69c517b build_date=2019-10-22T07:28:00Z
```
* Other relevant environment details: Container runtime, disk info, etc
browser: __Chrome 77__

__Config:__
Copy any non-default config values here or attach the full config as a gist or file.

Started with e2e testing enabled. 

```
--e2e-testing=true
```

TCL735:
TemplatesPage does not refresh when user successfully imports first template after visiting /orgs/[id]/settings/templates

It requires another successful import to see the previous (but not the current) user-imported template. Or a hard refresh will show all user-imported templates.

TCL735:
Another issue: after a successful import of a user defined template, the page goes back to the tab for "Static Templates" rather than staying on the "User Templates" tab.

Issue #15514, opened 2019-10-21
[UI - Templates] Import template ends on static templates while template is imported to user templates


karel-rehor:
__Steps to reproduce:__
List the minimal actions needed to reproduce the behavior.

1. In the templates page switch to User Templates
2. Click Import template
3. Upload a template through system file controls
4. Click Submit (i.e. "Import JSON as Template")

__Expected behavior:__
Expected that the template view would remain on "user templates" list and that the new template card would be immediately visible. 

It makes little sense to switch to static templates from the upload, since it appears all uploaded templates are "User Templates".

__Actual behavior:__
After clicking submit the view switched to the list of static templates.  I did not immediately see the template card in the list.  I had to then click "User Templates" again to view the new template card. 

__Screencast__
![ImportTemplate2StaticList01](https://user-images.githubusercontent.com/25881301/67215209-5fced780-f421-11e9-9f7b-1acee315bd9a.gif)

__Environment info:__

* System info: Run `uname -srm` and copy the output here
```
Linux 4.4.0-165-generic x86_64
```
* InfluxDB version: Run `influxd version` and copy the output here
This is from the nightly build docker image. 
```
ts=2019-10-21T07:30:37.430109Z lvl=info msg="Welcome to InfluxDB" log_id=0IcJRvyW000 version=nightly commit=913e689f4 build_date=2019-10-21T07:23:53Z
```
* Other relevant environment details: Container runtime, disk info, etc
Browser: **Chrome 77**

__Config:__
Copy any non-default config values here or attach the full config as a gist or file.
Started with e2e enabled.
```
--e2e-testing=true
```

karel-rehor:
Still occurs in nightly builds

```
ts=2020-01-10T08:57:47.255905Z lvl=info msg="Welcome to InfluxDB" log_id=0KFfcZyl000 version=nightly commit=63b3a0728 build_date=2020-01-10T08:57:47Z
```

PR #16642, opened 2020-01-23
refactor(templates): normalize Templates

TCL735:
Closes https://github.com/influxdata/influxdb/issues/15530
Closes https://github.com/influxdata/influxdb/issues/15514
as part of RFC normalization epic: influxdata/idpe#5591


ui/src/schemas/index.ts
```diff
@@ -147,6 +148,18 @@ export const telegraf = new schema.Entity(
...
 export const arrayOfTelegrafs = [telegraf]
 
+/* Templates */
+
+// Defines the schema for the "templates" resource
+export const template = new schema.Entity(
+  ResourceType.Templates,
+  {},
+  {
+    processStrategy: (t: TemplateSummary): TemplateSummary => t,
```
121watts: Don't need a `processStrategy` if we aren't modifying the object in some way.

ui/src/templates/actions/creators.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export type Action =
+  | ReturnType<typeof addTemplateSummary>
+  | ReturnType<typeof populateTemplateSummaries>
+  | ReturnType<typeof setExportTemplate>
+  | ReturnType<typeof setTemplatesStatus>
+  | ReturnType<typeof removeTemplateSummary>
+  | ReturnType<typeof setTemplateSummary>
+
+// R
```
121watts: I would describe what this type does here or remove the commend

ui/src/types/templates.ts
```diff
@@ -1,10 +1,28 @@
...
+  Variable,
+  View,
+} from 'src/types'
+
+export interface TemplateSummary extends DocumentListEntry {
+  labels: ILabel[]
+}
+
+export interface TemplatesState extends NormalizedState<TemplateSummary> {
+  status: RemoteDataState
```
121watts: I think `status` is already in `NormalizedState`

ui/src/schemas/index.ts
ebb-tide: I'm surprised to see that we are not modifying the object, especially since the BE guarantees very little about templates?

TCL735: I put that in there as a placeholder. It will be updated.

ui/src/tasks/components/TaskImportFromTemplateOverlay.tsx
```diff
@@ -138,7 +141,13 @@ class TaskImportFromTemplateOverlay extends PureComponent<
...
 
-const mstp = ({templates: {items, status}}: AppState): StateProps => {
+const mstp = (state: AppState): StateProps => {
+  const {
+    resources: {
+      templates: {status},
+    },
+  } = state
+  const items = getAll<TemplateSummary>(state, ResourceType.Templates)
   const filteredTemplates = items.filter(
```
121watts: This could be a selector.

ui/src/templates/actions/thunks.ts
```diff
@@ -141,37 +89,29 @@ export const createTemplate = (template: DocumentCreate) => async (
...
-): SetTemplateSummary => ({
-  type: ActionTypes.SetTemplateSummary,
-  payload: {id, templateSummary},
-})
-
 export const updateTemplate = (id: string, props: TemplateSummary) => async (
-  dispatch
+  dispatch: Dispatch<Action>
 ): Promise<void> => {
   try {
```
121watts:
```suggestion
  setTemplateSummary(id, RemoteDataState.Loading)
  try {
```
Probably want to set the loading state here.

ui/src/types/templates.ts
```diff
@@ -1,10 +1,29 @@
...
+  View,
+} from 'src/types'
+
+export interface TemplateSummary extends DocumentListEntry {
+  labels: ILabel[]
+  status: RemoteDataState
+}
+
+export interface TemplatesState extends NormalizedState<TemplateSummary> {
+  // status: RemoteDataState
```
121watts: Errant comment 😄

PR #16666, opened 2020-01-24
fix: dont crash on malformed label data

drdelambre:
Closes influxdata/honeybadger-errors#131

this broke when this.availableLabels[0] didn't exist because it didn't account for a list being empty AND the popover being open

### DIFF
--- a/ui/src/shared/components/inlineLabels/InlineLabelsEditor.tsx
+++ b/ui/src/shared/components/inlineLabels/InlineLabelsEditor.tsx
@@ -164,7 +164,11 @@ class InlineLabelsEditor extends Component<Props, State> {
     const {availableLabels} = this
     const {isPopoverVisible} = this.state
 
-    if (_.isEmpty(availableLabels) && !isPopoverVisible) {
+    if (_.isEmpty(availableLabels)) {
+      if (isPopoverVisible) {
+        return
+      }
+
       return this.setState({
         isPopoverVisible: true,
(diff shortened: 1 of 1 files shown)


ui/src/shared/components/inlineLabels/InlineLabelsEditor.tsx
```diff
@@ -164,7 +164,11 @@ class InlineLabelsEditor extends Component<Props, State> {
     const {availableLabels} = this
     const {isPopoverVisible} = this.state
 
-    if (_.isEmpty(availableLabels) && !isPopoverVisible) {
+    if (_.isEmpty(availableLabels)) {
```
hoorayimhelping: only `isEmpty` and `differenceBy` are being used from lodash here, what do you think of changing [this import](https://github.com/influxdata/influxdb/blob/4a439d45680dfa873db29ae04cc1a6c67ad39b84/ui/src/shared/components/inlineLabels/InlineLabelsEditor.tsx#L4) to not pull in all of lodash and only grab those two functions?

totally understand if you just want to get this fix out though. i'm not sure where i fall on fixing this bug and changing an import at the same time 🤔

anyway, nice work, i think the nested `if` is easier to read

drdelambre: we totally have a branch that does all of this! we just have to pump zoe up again about merging it

PR #16670, opened 2020-01-24
fix(kv): update tasks kv to use forward cursor

lyondhill:


kv/task.go
```diff
@@ -362,35 +362,31 @@ func (s *Service) findTasksByOrg(ctx context.Context, tx Tx, filter influxdb.Tas
...
-	} else {
-		// if we dont have an after we just move the cursor to the first instance of the orgID
-		key, err := org.ID.Encode()
-		if err != nil {
-			return nil, 0, influxdb.ErrInvalidTaskID
-		}
 
-		k, v = c.Seek(key)
+		// append a extra character because we want to skip "after" record
+		key = []byte(string(key) + "\x00")
```
GeorgeMac: This is making me think we need a handy option like `kv.SkipFirstKey()` or `kv.WithoutFirstKey()`

lyondhill: that would be a nice addition :)

PR #16680, opened 2020-01-28
feat: adding client library option for buckets

### LINKED ISSUE #16661: Add Client Libraries to the Add Data option in buckets
We should add the option to add data from a client library and drop the user on the client libraries tab when that happens.

image

drdelambre:
Closes #16661

This adds in the button for adding in data via the client library, but doesn't carry the bucket id around with it (as the other examples do). I couldn't tell from the ticket if the intent was to illustrate the client library capabilities or to pave the way for auto populating bucket ids into code snippets, so i went with the former.


russorat:
@drdelambre let's get this merged

PR #16688, opened 2020-01-30
fix(tsdb): Fix predicate clone

benbjohnson:
Helps https://github.com/influxdata/idpe/issues/5783

This pull request fixes predicate clone so that the initial clone clones the state but inner nodes all share the same cloned state.


tsdb/tsm1/predicate.go
```diff
@@ -386,9 +387,12 @@ func newPredicateCache(state *predicateState) predicateCache {
 }
 
 // Clone returns a deep copy of p.
-func (p *predicateCache) Clone() *predicateCache {
+func (p *predicateCache) Clone(state *predicateState) *predicateCache {
```
e-dard: I think it's probably worth documenting why `state` needs to be pushed down into this.

PR #16705, opened 2020-02-03
feat(pkger): add metrics svc middleware

jsteenb2:
- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/kit/metric/client.go
+++ b/kit/metric/client.go
@@ -1,0 +1,73 @@
+package metric
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/kit/prom"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// REDClient is a metrics client for collection RED metrics.
+type REDClient struct {
+	// RED metrics
+	reqs *prometheus.CounterVec
+	errs *prometheus.CounterVec
+	durs *prometheus.HistogramVec
+}
+
+// New creates a new REDClient.
+func New(reg *prom.Registry, service string) *REDClient {
+	// MiddlewareMetrics is a metrics service middleware for the notification endpoint service.
+	const namespace = "service"
+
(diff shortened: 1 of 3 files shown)


kit/metric/client.go
```diff
@@ -0,0 +1,77 @@
+package metric
```
kelwang: Is there any special reason to create a new folder under kit? we have a prometheus package under the root. Or we can make a package inside the pkgr folder?

kit/metric/client.go
```diff
@@ -0,0 +1,77 @@
...
+
+// Record returns a record fn that is called on any given return err. If an error is encountered
+// it will register the err metric. The err is never altered.
+func (c *REDClient) Record(method string) func(error) error {
+	start := time.Now()
+	return func(err error) error {
+		c.reqs.With(prometheus.Labels{"method": method})
+
+		if err != nil {
+			code := "unknown"
```
kelwang: if you can ErrorCode() later, for a non wrapped error, it will always return internal

jsteenb2: this coudl be reached if there is a `Error` type that doesn't set the code, which happens fairly frequently in our codebase unfortunately

kit/metric/client.go
jsteenb2: great question, here's the TLDR I gave @goller  in slack about this:

> yah, was going back and forth on that. we definittely could add it to prom. I decided for a new one for 2 reasons.
> 1. metric is dead obvious for someone who looks at the pkg name (“It probably does metrics…“)
> 2. wasn’t sure how tied we are to prom, if we end up using some other metrics engine, its a dependency issue that no one has to worry about except the metric pkg (although dependency change would force updates elsewhere)

kit/metric/client.go
jsteenb2: I take that back, you're right, its not reachable 🤦‍♂, inlined it 👍

PR #16706, opened 2020-02-03
chore(pkger): consolidate pkger http server into pkger

jsteenb2:
this is the last step for pkger to follow the service definition pattern
that is in the works. Some bits from http were moved into kit/transport/http
for reusability. End result is to hopefully axe most if not all of current platform http pkg for resuable types in kit. its a ways off still but we're closing in on some reusable types.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influxd/launcher/launcher.go
+++ b/cmd/influxd/launcher/launcher.go
@@ -30,6 +30,7 @@ import (
 	"github.com/influxdata/influxdb/kit/prom"
 	"github.com/influxdata/influxdb/kit/signals"
 	"github.com/influxdata/influxdb/kit/tracing"
+	kithttp "github.com/influxdata/influxdb/kit/transport/http"
 	"github.com/influxdata/influxdb/kv"
 	influxlogger "github.com/influxdata/influxdb/logger"
 	"github.com/influxdata/influxdb/nats"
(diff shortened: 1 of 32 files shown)


http/middleware.go
```diff
@@ -125,13 +125,14 @@ func ignoreMethod(ignoredMethods ...string) isValidMethodFn {
...
 
+// TODO(@jsteenb2): make this a stronger type that handlers can register routes that should not be logged.
 var blacklistEndpoints = map[string]isValidMethodFn{
 	prefixSignIn:                     ignoreMethod(),
 	prefixSignOut:                    ignoreMethod(),
 	prefixMe:                         ignoreMethod(),
 	mePasswordPath:                   ignoreMethod(),
 	usersPasswordPath:                ignoreMethod(),
-	prefixPackages + "/apply":        ignoreMethod(),
+	"/api/v2/packages/apply":         ignoreMethod(),
```
kelwang: not sure the reason behind this change

cmd/influxd/launcher/launcher.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/influxdata/influxdb/kit/prom"
 	"github.com/influxdata/influxdb/kit/signals"
 	"github.com/influxdata/influxdb/kit/tracing"
+	kithttp "github.com/influxdata/influxdb/kit/transport/http"
```
kelwang: nit picky, should this be kitHTTP?

http/middleware.go
jsteenb2: right now its to decouple the routes in this manner. Ideally we'd have a means of registering these paths instead of hardcoding them. This change is necessary b/c `prefixPackages` is no long er in http.

cmd/influxd/launcher/launcher.go
jsteenb2: with pkgs, its not pretty typical to have it be all lowercase

PR #16709, opened 2020-02-03
fix(flux): Reduce influxd and influx startup time if Flux unused

stuartcarnie:
This PR lazily initializes Flux built-in functions when Flux is used.

It significantly reduces the startup time of the `influxd` and `influx` binaries when Flux is disabled / unused.

Before (4.66s):

```
↳ time bin/18/influx
bin/18/influx  4.66s user 0.19s system 198% cpu 2.441 total
```

After (10ms):

```
↳ time bin/18/influx
bin/18/influx  0.01s user 0.01s system 88% cpu 0.021 total
```

Closes #

### DIFF
--- a/flux/control/controller.go
+++ b/flux/control/controller.go
@@ -7,7 +7,7 @@ import (
 	"github.com/influxdata/flux/lang"
 	"github.com/influxdata/flux/memory"
 	"github.com/influxdata/influxdb/coordinator"
-	_ "github.com/influxdata/influxdb/flux/builtin"
+	"github.com/influxdata/influxdb/flux/builtin"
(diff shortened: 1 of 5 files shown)


flux/control/controller.go
```diff
@@ -17,6 +17,8 @@ type MetaClient = coordinator.MetaClient
 type Authorizer = influxdb.Authorizer
 
 func NewController(mc MetaClient, reader influxdb.Reader, auth Authorizer, authEnabled bool, logger *zap.Logger) *Controller {
+	builtin.Initialize()
```
ayang64: this is awesome!

PR #16711, opened 2020-02-03
chore: remove queryBuilderGrouping flag

hoorayimhelping:
Removes the `queryBuilderGrouping` flag, effectively turning the feature on.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/timeMachine/utils/queryBuilder.ts
+++ b/ui/src/timeMachine/utils/queryBuilder.ts
@@ -14,7 +14,6 @@ import {
   WINDOW_PERIOD,
 } from 'src/variables/constants'
 import {AGG_WINDOW_AUTO} from 'src/timeMachine/constants/queryBuilder'
-import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 export function isConfigValid(builderConfig: BuilderConfig): boolean {
   const {buckets, tags} = builderConfig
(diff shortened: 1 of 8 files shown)


ui/src/timeMachine/utils/queryBuilder.ts
```diff
@@ -95,35 +94,18 @@ export function buildQuery(builderConfig: BuilderConfig): string {
...
-  const {aggregateWindow} = builderConfig
-  const fnCall = fn ? formatFunctionCall(fn, aggregateWindow.period) : ''
-
-  const query = `from(bucket: "${bucket}")
-  |> range(start: ${OPTION_NAME}.${TIME_RANGE_START}, stop: ${OPTION_NAME}.${TIME_RANGE_STOP})${tagFilterCall}${fnCall}`
-
-  return query
-}
-
-function buildQueryHelperButWithGrouping(
```
hoorayimhelping: this diff is wonky. i renamed `buildQueryHelperButWithGrouping` -> `buildQueryFromConfig` and removed `buildQueryHelper` which I thought could have been named a little clearly (hence `buildQueryFromConfig`)

hoorayimhelping:
@drdelambre I think you're right. I'm gonna keep this PR around for when we remove the feature flag, after it's all good. But the point of a flagging system is to set it to true, and if it doesn't work, set it to false, and I think this might be a little bit much.

PR #16712, opened 2020-02-03
feat(query): adds support for influxql as language type for queries

nathanielc:
This change allows for the InfluxQL language type to be used with the
/v2/query API endpoint.

This change also introduces a way to give the transpiler an explicit
bucket name instead of using the DBRPMapping service.
Requests to the endpoint will know the bucket name directly but will
likely not have run the migration step to populate the DBRP mappings.

### DIFF
--- a/http/query.go
+++ b/http/query.go
@@ -23,17 +23,23 @@ import (
 	"github.com/influxdata/influxdb"
 	"github.com/influxdata/influxdb/jsonweb"
 	"github.com/influxdata/influxdb/query"
+	transpiler "github.com/influxdata/influxdb/query/influxql"
 	"github.com/influxdata/influxql"
 )
 
 // QueryRequest is a flux query request.
 type QueryRequest struct {
+	Type  string `json:"type"`
+	Query string `json:"query"`
+
+	// Flux fields
 	Extern  *ast.File    `json:"extern,omitempty"`
 	Spec    *flux.Spec   `json:"spec,omitempty"`
 	AST     *ast.Package `json:"ast,omitempty"`
-	Query   string       `json:"query"`
-	Type    string       `json:"type"`
 	Dialect QueryDialect `json:"dialect"`
+
+	// InfluxQL fields
+	Bucket string `json:"bucket,omitempty"`
 
 	Org *influxdb.Organization `json:"-"`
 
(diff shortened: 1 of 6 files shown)


query/influxql/transpiler.go
```diff
@@ -691,27 +664,75 @@ func (t *transpilerState) from(m *influxql.Measurement) (ast.Expression, error)
...
 	} else {
-		// use mapping bucket id
-		args = []ast.Expression{
-			&ast.ObjectExpression{
-				Properties: []*ast.Property{
-					{
-						Key: &ast.Identifier{
-							Name: "bucketID",
+		if t.dbrpMappingSvc == nil {
+			return nil, errors.New("cannot transpiler, no dbrp mappings provided")
```
goller: prolly: influxdb.Error{Msg: "unable to transpile: db and rp mappings need to be created by some way", Code:  is this internal?}

query/influxql/transpiler.go
```diff
@@ -691,27 +664,75 @@ func (t *transpilerState) from(m *influxql.Measurement) (ast.Expression, error)
...
+		}
+		db, rp := m.Database, m.RetentionPolicy
+		if db == "" {
+			if t.config.DefaultDatabase == "" {
+				return nil, errors.New("database is required")
```
goller: What do you think of also giving guidance in the error message to add db to the query?

http/query.go
```diff
@@ -97,10 +103,14 @@ func (r QueryRequest) Validate() error {
 		}
 	}
 
-	if r.Type != "flux" {
+	if r.Type != "flux" && r.Type != "influxql" {
 		return fmt.Errorf(`unknown query type: %s`, r.Type)
 	}
 
+	if r.Type == "influxql" && r.Bucket == "" {
+		return fmt.Errorf("bucket parameter is required for influxql queries")
```
goller: with this requiring bucket, could the dbrp mapping be removed?

jsternberg: I'm not sure about this API. My understanding was that the API was for connecting the organization and that the bucket would be inferred from the database and retention policy. Is there a way we can avoid adding influxql specific fields to this query request?

nathanielc: The bucket is only required for the `/v2/query` endpoint, but its not required for the legacy `/query` endpoint so we can't remove dbrp mappings. Maybe if we remove that legacy endpoint we could, but that would be a separate change.

query/influxql/transpiler.go
nathanielc: Currently the only way this path is hit is if goreplay fails to send the appropriate parameters. Normally yes informing the user would be good, in this case its an internal code path.

PR #16716, opened 2020-02-04
fix(http): improve error handling and response to api consumer for org service

### LINKED ISSUE #16498: Malformed JSON produces unhelpful error message
__Steps to reproduce:__

`curl --data '{"MYSQL_USER", "davidgs"}' --request PATCH --url https://us-west-2-1.aws.cloud2.influxdata.com/api/v2/orgs/f978dfb87de31ea3/secrets --header 'Authorization: Token token>'`

__Expected behavior:__
Describe what you expected to happen.
Some sort of detailed error message
__Actual behavior:__
Describe What actually happened.
"An internal error has occured"

__Environment info:__

__Config:__

__Logs:__

__Performance:__

```sh
iostat -xd 1 30 > iostat.txt
```

jsteenb2:
Closes #16498

add a reusable api type for decoding and responding to http requests. Improves error handling.

Work is broken up into 2 commits, first one for creating the API decoder/responder, and the 2nd for using it within the org service. The new API type normalizes handling of decoding issues and failed write issues amongst. It is also very extensible, more we can do with it than what meets the eye.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/org_service.go
+++ b/http/org_service.go
@@ -2,7 +2,6 @@ package http
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"net/http"
 	"path"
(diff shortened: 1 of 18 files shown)


http/org_service.go
```diff
@@ -203,399 +201,232 @@ func newOrgResponse(o *influxdb.Organization) *orgResponse {
...
 	}
-}
+	h.log.Debug("Org created", zap.String("org", fmt.Sprint(org)))
 
-type postOrgRequest struct {
-	Org *influxdb.Organization
-}
+	h.API.Respond(w, http.StatusCreated, newOrgResponse(org))
 
-func decodePostOrgRequest(ctx context.Context, r *http.Request) (*postOrgRequest, error) {
```
jsteenb2: this PR axes a bunch of unneccsary indirection, reduce the jumping around and make it easier to follow within the handler itself.

PR #16724, opened 2020-02-04
chore(http): refactor bkt service to use API type

jsteenb2:
- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/bucket_service.go
+++ b/http/bucket_service.go
@@ -2,7 +2,6 @@ package http
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"net/http"
 	"path"
@@ -12,6 +11,7 @@ import (
 	"github.com/influxdata/httprouter"
 	"github.com/influxdata/influxdb"
 	"github.com/influxdata/influxdb/kit/tracing"
+	kithttp "github.com/influxdata/influxdb/kit/transport/http"
 	"github.com/influxdata/influxdb/pkg/httpc"
 	"go.uber.org/zap"
 )
@@ -48,7 +48,7 @@ func NewBucketBackend(log *zap.Logger, b *APIBackend) *BucketBackend {
 // BucketHandler represents an HTTP API handler for buckets.
 type BucketHandler struct {
 	*httprouter.Router
-	influxdb.HTTPErrorHandler
+	api *kithttp.API
 	log *zap.Logger
 
 	BucketService              influxdb.BucketService
@@ -74,9 +74,9 @@ const (
 // NewBucketHandler returns a new instance of BucketHandler.
 func NewBucketHandler(log *zap.Logger, b *BucketBackend) *BucketHandler {
 	h := &BucketHandler{
-		Router:           NewRouter(b.HTTPErrorHandler),
-		HTTPErrorHandler: b.HTTPErrorHandler,
-		log:              log,
+		Router: NewRouter(b.HTTPErrorHandler),
+		api:    kithttp.NewAPI(kithttp.WithLog(log)),
+		log:    log,
 
 		BucketService:              b.BucketService,
 		BucketOperationLogService:  b.BucketOperationLogService,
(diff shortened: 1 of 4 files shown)


http/bucket_service_test.go
```diff
@@ -456,7 +456,7 @@ func TestService_handlePostBucket(t *testing.T) {
 				},
 			},
 			wants: wants{
-				statusCode: http.StatusBadRequest,
+				statusCode: http.StatusUnprocessableEntity,
```
jademcgough: If this is a breaking change, we should surface this to monitoring and document it in our swagger.

jsteenb2: good catch, I'll circle the wagons on that right now

jsteenb2: lol none of the error cases are in the swagger docs 🤦‍♀

jsteenb2: or maybe the status code just doesn't affect it 🤔, clear as mud atm

jsteenb2: one other note, this shouldn't' be considered a breaking change. there was nothing stopping us from returning a 422 earlier. Hopefully it was, hard to track it down, goes through 4 layers of indirection before finding an originating error. Fwiw, the status code it is changed ot is the correct one, a 400 is more for unmarshaling failing to decode/read something. 422 typically outlines validation failure. Milealge may vary 🤷‍♂

jademcgough: We probably overuse 400 responses in the API as kind of a catch-all default error response (and it's, in fact, our `default` for responses in the swagger).

I'd consider this to be breaking in that it's a change in behavior. If javascript in the UI is looking for a 400 response instead of a 400-level response, the UI would break. Pretty minor in this case, and suggests that the UI is brittle if it can't handle all failure situations, but it's a change nonetheless.

Not blocking though - I can surface this information to other people. Would be worthwhile to add the 422 response to the swagger though to make sure it's documented.

http/bucket_service.go
```diff
@@ -353,27 +343,42 @@ type postBucketRequest struct {
...
 
+	// Only support a single retention period for the moment
+	var errs []string
+	if len(b.RetentionRules) > 0 {
+		if _, err := b.RetentionRules[0].RetentionPeriod(); err != nil {
+			errs = append(errs, err.Error())
+		}
 	}
+
+	// names starting with an underscore are reserved for system buckets
```
jademcgough: nice, I like moving this to validation

http/bucket_service_test.go
jsteenb2: good points on all accounts, I pushed up the swagger changes 👍

http/bucket_service.go
```diff
@@ -353,27 +343,42 @@ type postBucketRequest struct {
...
+	if err := validBucketName(b.toInfluxDB()); err != nil {
+		errs = append(errs, err.Error())
+	}
+
+	if len(errs) > 0 {
+		return &influxdb.Error{
+			Code: influxdb.EUnprocessableEntity,
+			Msg:  strings.Join(errs, "; "),
```
jademcgough: How common is this pattern for http error responses? It seems like it would be less computationally intensive (especially if we have validation checks that hit other services or a db) to just do an early return with the first error, and I could see this being difficult to read if there are a lot of errors.

jsteenb2: in go, unfortunately, its all to uncommon to return early. From a user perspective, i want everything that failed validation. We don't have a mechanism for this atm, which is a pain point for me. Takes multiple requests potentially to figure out what all is wrong with a bucket req in this case. Computationally... its a nothing burger 🤷‍♂. If we are relying on multiple services for **validation**, something has gone horribly wrong with our system architecture 😳 . In that case its even better not to return early imo, otherwise you put stress on the entire system for each additional unneeded request 🤷‍♂. 

This `: ` delimiter is definitely pretty lame. I want something better for sure. But the single error message isn't a whole lot better.  i would love to see something like:

```json
{
    code: "unprocessable entity"
    errors: [
       {
          "field": "name",
          "reason": "name must not be prefixed with \"_\""
       },
       {
          "field": "retentionRules",
          "reason": "expiration seconds must be greater than or equal to one second"
       }
    ]
}
```

jsteenb2: I opted not to introduce that as it would be a larger chagne, and would require changes on the FE side. Although, it would make user feedback in the UI programmatic at that point if we did. More than just a toast message would do.  @hoorayimhelping you dig?

jademcgough: I'd much rather have errors returned as an object than as glommed-together messages, but that'd be a big change. We should get buy-in from @desa before making this change.

jsteenb2: sounds good, @desa what say you?

jsteenb2: I'll switch it back to returning early for now. When we get buy in from everyone we can improve this with a collection of failed validations, maybe something like the above 🤷‍♂

desa: Would be really awesome to have

```
{
    "code": "unprocessable entity",
    "errors": [
       {
          "field": "name",
          "reason": "name must not be prefixed with \"_\""
       },
       {
          "field": "retentionRules",
          "reason": "expiration seconds must be greater than or equal to one second"
       }
    ]
}
```

I'm onboard with returning all issues with the request.

jsteenb2: I'll write up a feature req for adding enhanced error return for this situation

jsteenb2: new issue: https://github.com/influxdata/influxdb/issues/16727

jademcgough: I can approve this once the error handling changes are moved out of the PR. We can tackle those changes in ^

PR #16754, opened 2020-02-06
fix(CORS): accept OPTIONS preflight request with Origin header

rawkode:
Browsers send a preflight `OPTIONS` request with a `Origin` header to validate the CORS policy before sending the real request. Currently we block this from working, disallowing any developer to build their own tooling with the Influx API

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/kit/transport/http/middleware.go
+++ b/kit/transport/http/middleware.go
@@ -53,9 +53,13 @@ func Metrics(name string, reqMetric *prometheus.CounterVec, durMetric *prometheu
 
 func SkipOptions(next http.Handler) http.Handler {
 	fn := func(w http.ResponseWriter, r *http.Request) {
-		if r.Method == "OPTIONS" {
+		// Preflight CORS requests from the browser will send an options request,
+		// so we need to make sure we satisfy them
+		if origin := r.Header.Get("Origin"); origin == "" && r.Method == http.MethodOptions {
+			w.WriteHeader(http.StatusMethodNotAllowed)
 			return
 		}
+
 		next.ServeHTTP(w, r)
 	}
 	return http.HandlerFunc(fn)
--- a/kit/transport/http/middleware_test.go
+++ b/kit/transport/http/middleware_test.go
@@ -1,10 +1,12 @@
 package http
 
 import (
+	"net/http"
 	"path"
 	"testing"
 
 	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/pkg/testttp"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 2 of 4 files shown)


kit/transport/http/middleware_test.go
```diff
@@ -43,3 +45,41 @@ func Test_normalizePath(t *testing.T) {
...
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			for _, header := range tt.headers {
+				req.Header.Set(header, "value")
+			}
+
+			rr := httptest.NewRecorder()
+			handler := http.HandlerFunc(SkipOptions(nil))
```
jsteenb2: likely seeing a panic in when the wrapped handler(nil) is called here. One other thing I'd recommend is checking out the `pkg/testttp` pkg here in influxdb. It makes these tests super easy to build... I may be biased b/c I wrote the pkg, but still its handy 😬:

```go
svr := kithttp.SkipOptions(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
	w.Write([]byte(tt.expectedBody))
}))

testttp.
	HTTP(t, tt.method, "/", nil).
	Do(svr).
	ExpectStatus(http.StatusOK).
	ExpectBody(func(body *bytes.Buffer) {
		assert.Equal(t, tt.expectedBody, body.String())
	})
```

I added a expected body value to make sure the interior handler wasn't called. Smalll extension to the testcase struct

kit/transport/http/middleware.go
```diff
@@ -53,7 +53,11 @@ func Metrics(name string, reqMetric *prometheus.CounterVec, durMetric *prometheu
 
 func SkipOptions(next http.Handler) http.Handler {
 	fn := func(w http.ResponseWriter, r *http.Request) {
-		if r.Method == "OPTIONS" {
+		// Preflight CORS requests from the browser will send an options request,
+		// so we need to make sure we satisfy them
+		origin := r.Header.Get(("Origin"))
+
+		if origin == "" && r.Method == "OPTIONS" {
 			return
```
jsteenb2: I know we just return here, but I wonder if we're better off returning a different status code than `200`?

kit/transport/http/middleware.go
```diff
@@ -53,9 +53,13 @@ func Metrics(name string, reqMetric *prometheus.CounterVec, durMetric *prometheu
 
 func SkipOptions(next http.Handler) http.Handler {
 	fn := func(w http.ResponseWriter, r *http.Request) {
-		if r.Method == "OPTIONS" {
+		// Preflight CORS requests from the browser will send an options request,
+		// so we need to make sure we satisfy them
+		if origin := r.Header.Get("Origin"); origin == "" {
+			w.WriteHeader(http.StatusMethodNotAllowed)
```
jsteenb2: love it 👍

pkg/testttp/http.go
```diff
@@ -146,6 +146,15 @@ func (r *Resp) ExpectBody(fn func(body *bytes.Buffer)) *Resp {
 	return r
 }
 
+// ExpectHeaders asserts that multiple headers with values exist in the recorder.
+func (r *Resp) ExpectHeaders(h map[string]string) *Resp {
```
jsteenb2: 👏 nice

rawkode:
@jsteenb2 I don't believe this test failure is my problem, can you confirm? :D

kit/transport/http/middleware.go
jsteenb2: @rawkode we're not checking the Options method here 😳 just realized 🤣 

here's teh if block:

```go
if origin := r.Header.Get("Origin"); origin == "" && r.Method == http.MethodOptions {
```

rawkode: D'oh!

rawkode: Doesn't explain why my tests pass locally though 😯

jsteenb2: it failed for me locally before the fix. Maybe I gave you the wrong test cmd 😬

PR #16760, opened 2020-02-06
fix: remove orgs from token permission list

glinton:
Closes #influxdata/idpe#5940

Error message received when creating all access token:
```json
{
  "code": "forbidden",
  "message": "permission read:orgs/03cf84c67a620000/orgs is not allowed: read:orgs/03cf84c67a620000/orgs is unauthorized"
}
```

Resource `/orgs/{orgID}/orgs` isn't defined, removed the `orgs` entry from the authorizations portion.

Probably works.

### DIFF
--- a/ui/src/authorizations/utils/permissions.ts
+++ b/ui/src/authorizations/utils/permissions.ts
@@ -1,12 +1,12 @@
 import {Bucket, Permission} from 'src/types'
 
-type PermissionTypeValues = Permission['resource']['type']
+type PermissionTypes = Permission['resource']['type']
 
 function assertNever(x: never): never {
   throw new Error('Unexpected object: ' + x)
 }
 
-const allPermissionTypes: PermissionTypeValues[] = [
+const allPermissionTypes: PermissionTypes[] = [
   'authorizations',
   'buckets',
   'checks',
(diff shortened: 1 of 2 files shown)


desa:
I think we do want this permission
```
  {	
    action: 'read',
    resource: {type: 'orgs', id: orgID},
  }
```

glinton:
then something deeper doesn't work, it seems (i assume) that it requests `/orgs/{org_id}/{resource_type}` and `/orgs/id/orgs` doesn't exist. i'll try running/reproducing this locally to verify what gets requested.

desa:
I think that
```
  {	
    action: 'read',
    resource: {type: 'orgs', id: orgID},
  }
```
should create the permission `/orgs/{orgid}`

ui/src/authorizations/utils/permissions.ts
```diff
@@ -59,6 +58,13 @@ const ensureT = (orgID: string) => (t: PermissionTypes) => {
           resource: {type: t, orgID},
         },
       ]
+    case 'orgs':
```
desa: Can you add a  comment here noting specifically that we expect the org permission to be at `id` and not to have `orgID`?

PR #16774, opened 2020-02-07
chore(task): add test around paging using after

GeorgeMac:
Closes https://github.com/influxdata/influxdb/issues/16775

A recent regression surfaced around tasks paging using the `After` field. Causing an infinite loop to occur when consumers try to loop over results.

This was fixed in https://github.com/influxdata/influxdb/pull/16758.

This PR just introduces a test around task paging to ensure this doesn't regress again.

I can confirm I ran this against a commit prior to the fix introduced above and this test failed.

UPDATE:

This now fixes https://github.com/influxdata/influxdb/issues/16775 as this test surfaced the problem.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/kv/task.go
+++ b/kv/task.go
@@ -311,7 +311,21 @@ func (s *Service) findTasksByUser(ctx context.Context, tx Tx, filter influxdb.Ta
 		return nil, 0, err
 	}
 
-	matchFn := newTaskMatchFn(filter, org)
+	var (
+		afterSeen bool
+		after     = func(task *influxdb.Task) bool {
+			if filter.After == nil || afterSeen {
+				return true
+			}
+
+			if task.ID == *filter.After {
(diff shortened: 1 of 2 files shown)


inmem/kv.go
```diff
@@ -284,10 +284,9 @@ func (b *Bucket) ForwardCursor(seek []byte, opts ...kv.CursorOption) (kv.Forward
 		defer close(pairs)
 
 		var (
-			batch     []pair
-			fn        = config.Hints.PredicateFn
-			iterate   = b.ascend
-			skipFirst = config.SkipFirst
+			batch   []pair
```
GeorgeMac: Note for me. I did this because I thought the problem was in `inmem` but I found it wasn't. So perhaps I should put this back 🤔

PR #16786, opened 2020-02-07
feat(cmd/influx): add secret cli

### LINKED ISSUE #16475: Add the ability to manage secrets from the CLI
as a user, i would like to manage my secrets via the cli. something like `./influx secrets -h`

list, create, update, delete should be supported.

kelwang:
Closes #16475

Add the http secret service, cli, and tests

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influx/secret.go
+++ b/cmd/influx/secret.go
@@ -1,0 +1,184 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/http"
+	"github.com/spf13/cobra"
+	input "github.com/tcnksm/go-input"
+)
+
+type secretSVCsFn func() (influxdb.SecretService, influxdb.OrganizationService, func(*input.UI) string, error)
+
+func cmdSecret(opts ...genericCLIOptFn) *cobra.Command {
+	return newCmdSecretBuilder(newSecretSVCs, opts...).cmd()
+}
+
+type cmdSecretBuilder struct {
+	genericCLIOpts
+
+	svcFn secretSVCsFn
+
+	key string
+	org organization
+}
+
+func newCmdSecretBuilder(svcsFn secretSVCsFn, opts ...genericCLIOptFn) *cmdSecretBuilder {
+	opt := genericCLIOpts{
+		in: os.Stdin,
+		w:  os.Stdout,
+	}
+	for _, o := range opts {
+		o(&opt)
+	}
+
+	return &cmdSecretBuilder{
+		genericCLIOpts: opt,
+		svcFn:          svcsFn,
+	}
+}
+
+func (b *cmdSecretBuilder) cmd() *cobra.Command {
+	cmd := b.newCmd("secret", nil)
+	cmd.Short = "Secret management commands"
+	cmd.Run = seeHelp
+	cmd.AddCommand(
+		b.cmdDelete(),
+		b.cmdFind(),
+		b.cmdUpdate(),
+	)
+	return cmd
+}
+
+func (b *cmdSecretBuilder) cmdUpdate() *cobra.Command {
+	cmd := b.newCmd("update", b.cmdUpdateRunEFn)
+	cmd.Short = "Update secret"
+	cmd.Flags().StringVarP(&b.key, "key", "k", "", "The secret key (required)")
+	cmd.MarkFlagRequired("key")
+	b.org.register(cmd, false)
+
+	return cmd
+}
+
+func (b *cmdSecretBuilder) cmdDelete() *cobra.Command {
+	cmd := b.newCmd("delete", b.cmdDeleteRunEFn)
+	cmd.Short = "Delete secret"
+
+	cmd.Flags().StringVarP(&b.key, "key", "k", "", "The secret key (required)")
+	cmd.MarkFlagRequired("key")
+	b.org.register(cmd, false)
+
+	return cmd
+}
+
+func (b *cmdSecretBuilder) cmdUpdateRunEFn(cmd *cobra.Command, args []string) error {
+	scrSVC, orgSVC, getSecretFn, err := b.svcFn()
+	if err != nil {
+		return err
+	}
+	orgID, err := b.org.getID(orgSVC)
+	if err != nil {
+		return err
+	}
+
+	ctx := context.Background()
+
+	ui := &input.UI{
+		Writer: b.genericCLIOpts.w,
+		Reader: b.genericCLIOpts.in,
+	}
+	secret := getSecretFn(ui)
+
+	if err := scrSVC.PatchSecrets(ctx, orgID, map[string]string{b.key: secret}); err != nil {
+		return fmt.Errorf("failed to update secret with key %q: %v", b.key, err)
+	}
+
+	w := b.newTabWriter()
+	w.WriteHeaders("Key", "OrgID", "Updated")
+	w.Write(map[string]interface{}{
+		"Key":     b.key,
+		"OrgID":   orgID,
+		"Updated": true,
+	})
+	w.Flush()
+
+	return nil
+}
+
+func (b *cmdSecretBuilder) cmdDeleteRunEFn(cmd *cobra.Command, args []string) error {
+	scrSVC, orgSVC, _, err := b.svcFn()
+	if err != nil {
+		return err
+	}
+	orgID, err := b.org.getID(orgSVC)
+	if err != nil {
+		return err
+	}
+
+	ctx := context.Background()
+	if err := scrSVC.DeleteSecret(ctx, orgID, b.key); err != nil {
+		return fmt.Errorf("failed to delete secret with key %q: %v", b.key, err)
+	}
+
+	w := b.newTabWriter()
+	w.WriteHeaders("Key", "OrgID", "Deleted")
+	w.Write(map[string]interface{}{
+		"Key":     b.key,
+		"OrgID":   orgID,
+		"Deleted": true,
+	})
+	w.Flush()
+
+	return nil
+}
+
+func (b *cmdSecretBuilder) cmdFind() *cobra.Command {
+	cmd := b.newCmd("find", b.cmdFindRunEFn)
+	cmd.Short = "Find secrets"
+	b.org.register(cmd, false)
+
+	return cmd
+}
+
+func (b *cmdSecretBuilder) cmdFindRunEFn(cmd *cobra.Command, args []string) error {
+
+	scrSVC, orgSVC, _, err := b.svcFn()
+	if err != nil {
+		return err
+	}
+
+	orgID, err := b.org.getID(orgSVC)
+	if err != nil {
+		return err
+	}
+
+	secrets, err := scrSVC.GetSecretKeys(context.Background(), orgID)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve secret keys: %s", err)
+	}
+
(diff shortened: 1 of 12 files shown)


cmd/influx/secret.go
```diff
@@ -0,0 +1,187 @@
...
+
+	ctx := context.Background()
+
+	ui := &input.UI{
+		Writer: b.genericCLIOpts.w,
+		Reader: b.genericCLIOpts.in,
+	}
+	secret := getSecretFn(ui)
+
+	if err := scrSVC.PatchSecrets(ctx, orgID, map[string]string{
```
jsteenb2: nit: can we pull out all multi calls out of the if block?

```go
err := scrSVC.PatchSecrets(ctx, orgID, map[string]string{
...
})
if err != nil {
```

when it grows it turns into a frankenstein like this 😢:
https://github.com/influxdata/influxdb/blob/master/cmd/influxd/launcher/launcher.go#L664

cmd/influx/secret.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		return err
+	}
+
+	ctx := context.Background()
+	if err := scrSVC.DeleteSecret(ctx, orgID, b.key); err != nil {
+		return fmt.Errorf("failed to delete secret with key %q: %v", b.key, err)
+	}
+
+	w := internal.NewTabWriter(b.w)
```
jsteenb2: thought buble just came to me, feels like the `genericCLIOpt` type could be extended with a `tabWriter(headers []string) *tabwriter.Writer` method. That way it always pulls the `b.w` and no one accidentally puts the os.Stdout here. Food for thought 🤔

something like this maybe:

```go
// this exists already
type genericCLIOpts struct {
	in         io.Reader
	w          io.Writer
	runEWrapFn runEWrapFn
}

func (o genericCLIOpts) newTabWriter(headers []string) *internal.TabWriter{
     w := internal.NewTabWriter(b.w)
     w.WriteHeaders(headers)
     return w
}


// secret.go
w := b.newTabWriter("Key", "OrgID", "Deleted")
w.Write(map[string]interface{}{
    "Key":     b.key,
    "OrgID":   orgID,
    "Deleted": true,
})
w.Flush()
```

cmd/influx/secret.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	orgID, err := b.org.getID(orgSVC)
+	if err != nil {
+		return err
+	}
+
+	secrets, err := scrSVC.GetSecretKeys(context.Background(), orgID)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve secret keys: %s", err)
```
jsteenb2: is there more info we can provide the user about how to fix the issue? curious if we can give more details of the error so they have a better idea of what they can do to fix it 🤔

cmd/influx/setup.go
```diff
@@ -236,54 +237,70 @@ You have entered:
...
-)
+func errSecretIsNotMatch(title string) error {
+	return fmt.Errorf(title + "s do not match")
+}
+
+const errTooShort = "too short"
+
+func errSecretIsTooShort(title string) error {
+	return &influxdb.Error{
+		Code: errTooShort,
```
jsteenb2: the code would be `influxdb.UnprocessableEntity` here, `too short` isn't a domain eror code. The message explains how it is out of range.  Let's remove the errTooShort code here

cmd/influx/setup.go
```diff
@@ -236,54 +237,70 @@ You have entered:
...
+				ui.Writer.Write(promptWithColor(strings.ToTitle(title)+" too short - minimum length is 8 characters!\n\r", colorRed))
+				goto enterSecret
+			} else if secret = strings.TrimSpace(secret); secret == "" {
 				continue
 			}
 		}
 		break
 	}
-	query = string(promptWithColor("Please type your"+newStr+" password again", colorCyan))
+	query = string(promptWithColor("Please type your"+newStr+" "+title+" again", colorCyan))
```
jsteenb2: does this need to be `titled` as well?

http/org_service.go
```diff
@@ -426,6 +430,85 @@ func newOrganizationLogResponse(id influxdb.ID, es []*influxdb.OperationLogEntry
...
 }
 
+// SecretService connects to Influx via HTTP using tokens to manage secrets.
+type SecretService struct {
+	Client *httpc.Client
+}
+
+// LoadSecret is not implemented for http
+func (s *SecretService) LoadSecret(ctx context.Context, orgID influxdb.ID, k string) (string, error) {
+	return "", &influxdb.Error{
```
jsteenb2: feels like we should  panic here? not being implemented then it is called that would qualify (in my head) as an exception (panic). Does that make sense?

http/org_service.go
```diff
@@ -426,6 +430,85 @@ func newOrganizationLogResponse(id influxdb.ID, es []*influxdb.OperationLogEntry
...
+		Code: influxdb.EMethodNotAllowed,
+		Msg:  "load secret is not implemented for http",
+	}
+}
+
+// PutSecret is not implemented for http.
+func (s *SecretService) PutSecret(ctx context.Context, orgID influxdb.ID, k string, v string) error {
+	return &influxdb.Error{
+		Code: influxdb.EMethodNotAllowed,
```
jsteenb2: same

http/org_service.go
```diff
@@ -426,6 +430,85 @@ func newOrganizationLogResponse(id influxdb.ID, es []*influxdb.OperationLogEntry
...
+		return nil, err
+	}
+
+	return ss.Secrets, nil
+}
+
+// PutSecrets is not implemented for http.
+func (s *SecretService) PutSecrets(ctx context.Context, orgID influxdb.ID, m map[string]string) error {
+	return &influxdb.Error{
+		Code: influxdb.EMethodNotAllowed,
```
jsteenb2: same

http/org_service.go
```diff
@@ -426,6 +430,85 @@ func newOrganizationLogResponse(id influxdb.ID, es []*influxdb.OperationLogEntry
...
+	}
+}
+
+// PatchSecrets will update the existing secret with new via http.
+func (s *SecretService) PatchSecrets(ctx context.Context, orgID influxdb.ID, m map[string]string) error {
+	span, _ := tracing.StartSpanFromContext(ctx)
+	defer span.Finish()
+
+	if orgID != 0 {
+		span.LogKV("org-id", orgID)
```
jsteenb2: I'm curious, we use the span's in some places, and not i others... feels like a service middleware concern to me.  Just wrap the service with a trace mw and its donezo that way 🤷‍♂

http/org_service_test.go
```diff
@@ -10,12 +10,12 @@ import (
 	"net/http/httptest"
 	"testing"
 
-	platform "github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb"
```
jsteenb2: 👏

cmd/influx/setup.go
kelwang: yeah

http/org_service.go
kelwang: Not a bad suggestion. But I think panic is actually a handling.

In go, I usually prefer handle error, including panics, in one last step. 
It's easier for testing, also easier to track all the panics in our system.

http/org_service.go
kelwang: same here, not sure what's that for

cmd/influx/secret.go
kelwang: Good idea, let me create another issue for that, and get docs team involved.

http/org_service.go
jsteenb2: The point I was driving at, is regardless the language really, an exception is outside the domain the code was implemented for. In those situations, typically we would throw an exception(a panic in go). 

My other reason for a `panic` is that testing code that does is not implemented doesn't make a lot of sense. It should not be used, and if it is then it should panic to indicate to the consumer this is not valid. An error being returned can send the next person who might incorporate the secret service down a rabbit hole. We can avoid that wasted effort with a panic to indicate clearly that the behavior is not implemented.  

Probably shoudl have just wrote all that the first time lol

cmd/influx/secret.go
kelwang: yeah, why not

http/org_service.go
jsteenb2: lets go on and drop them for the time being then 👍

http/org_service.go
kelwang: I kinda agree your point actually. However we have similar pattern in somewhere else, like 
https://github.com/influxdata/influxdb/blob/1f1b1c492e502ca834cae6305e55e523edb81253/task/options/flux_deps.go#L35 The benefit is giving the consumer more flexibility to override and modify.  

I think for a long term wise, we could revisit our dependency strategy, like split the service, so everything make senses. Not just for secret, but for other crud services as well. For now, I think both works, I personally prefer to keep it this way.

cmd/influx/secret.go
kelwang: sure

http/org_service.go
jsteenb2: the points you're making don't really fall in line with what I'm getting at. If you want to leave it as is, go for it. We need to have a definition for these situations as a company, sooner than later. The more  unique each place is, the worse off the code base is generally.

cmd/influx/secret.go
```diff
@@ -0,0 +1,184 @@
...
+	)
+	return cmd
+}
+
+func (b *cmdSecretBuilder) cmdUpdate() *cobra.Command {
+	cmd := b.newCmd("update", b.cmdUpdateRunEFn)
+	cmd.Short = "Update secret"
+	cmd.Flags().StringVarP(&b.key, "key", "k", "", "The secret key (required)")
+	cmd.MarkFlagRequired("key")
+	b.org.register(cmd, false)
```
sanderson: You can pass a key, but how do you update the value of the secret?

cmd/influx/secret.go
```diff
@@ -0,0 +1,184 @@
...
+	})
+	w.Flush()
+
+	return nil
+}
+
+func (b *cmdSecretBuilder) cmdFind() *cobra.Command {
+	cmd := b.newCmd("find", b.cmdFindRunEFn)
+	cmd.Short = "Find secrets"
+	b.org.register(cmd, false)
```
sanderson: This is an overall critique of "find"/"list" commands throughout the `influx` CLI. We are inconsistent. For some it's "list" and others it's "find". The term "find" indicates a searching action and I would expect ways to filter the results. This just lists all the secret keys.

My personal preference would be to make everything consistent and use either `list` or `ls` to align with standard Linux/UNIX command line tools.

@jademcgough

jsteenb2: @sanderson, this x1000! I've had the same feelings towards list|find. I would really love to see it become list if it does not allow for filtering. We do have filtering on some Find commands, either by name or what not. But these are still akin to list operations for me in most nix envs. I would love to see `list` and `ls` as an alias shorthand for calling it 👍

jsteenb2: @sanderson we're going to make these changes 👍. @influxdata/tools-team is on board 🤘

sanderson: @jsteenb2 🎉

cmd/influx/secret.go
jademcgough: @sanderson This is done through a prompt similar to our password flow. I'm adding a card to provide it over the CLI as an arg in the same command.

PR #16789, opened 2020-02-08
feat: add indexes to urm for user lookups

lyondhill:


kv/indexer.go
```diff
@@ -13,71 +13,84 @@ type kvIndexer struct {
...
+				bucket, err := tx.Bucket(batch.bucketName)
 				if err != nil {
 					return err
 				}
-			}
-			return nil
-		})
+				// insert all the keys
+				for k, v := range batch.idxs {
+					err := bucket.Put([]byte(k), v)
```
GeorgeMac: How will we deal with indexing 1 to many here? I imagine that we would really only be interested in creating a composite key here of the thing to index on (foreign) combined with the target indexed item (primary) via some separator. I did this in the generic on here: https://gist.github.com/GeorgeMac/143bd3d1ecab194bada9ff01341e9907#file-indexer-go-L61-L64

This went through my mind when I did the inline one and not when I implemented it for auth lookup for some reason 😂 . Do we want to have the abstraction take a separate `foreign -> primary` mapping and apply a standard here? Or have the caller make a composite key and just send it here to be inserted?
My gut is to have the Indexer take the two separately and combine them itself.

GeorgeMac: Ahh OK I read further down. You're building the composite from the outside cool 👍

kv/urm.go
```diff
@@ -343,6 +471,11 @@ func (s *Service) deleteUserResourceMapping(ctx context.Context, tx Tx, filter i
 	if err := b.Delete(key); err != nil {
 		return UnavailableURMServiceError(err)
 	}
+
+	if err := ib.Delete(ikey); err != nil {
+		return UnavailableURMServiceError(err)
```
GeorgeMac: One thing I just realized is this will fail for authorization net yet in index. Which is a plausible scenario. Perhaps we should just log the error in this situation and continue with the delete?

GeorgeMac: Actually, I think Delete doesn't return an error when key is not found. Just validating that now.

GeorgeMac: Yeah we're good ignore me.

kv/urm_test.go
```diff
@@ -49,3 +63,35 @@ func initUserResourceMappingService(s kv.Store, f influxdbtesting.UserResourceFi
...
+	userCount := b.N / 50
+	if userCount < 1 {
+		userCount = 1
+	}
+	users := make([]influxdb.ID, userCount)
+	for i := 0; i < userCount; i++ {
+		users[i] = idgen.ID()
+	}
+
+	for i := 0; i < b.N; i++ {
```
GeorgeMac: Just checked the docs and this loop should be around the actual function being tested.
namely: `svc.FindUserResourceMappings(...)`.
I tried adapting this benchmark to mine for both a warm and cold cache and it gave bizarre results.
the `b.N` value should be used to loop over and call the function being tested. Rather than populating the state it iterates over.
I changed my benchmark to populated with a large fixed amount of data and then looped over as described above and it reflected the reality I expected.

PR #16805, opened 2020-02-10
fix(cmd): use list instead of find in cli

jademcgough:
long-running ask from @sanderson :)

Change all places in the CLI where we have `find` to use `list`. We inconsistently use the two in our CLI, so this adds more consistency in the UX.

Also adds `find` (to make this non-breaking) and `ls` (to be more unixy) as aliases for the command.

### DIFF
--- a/cmd/influx/bucket_test.go
+++ b/cmd/influx/bucket_test.go
@@ -305,7 +305,7 @@ func TestCmdBucket(t *testing.T) {
 				cmd := builder.cmd(cmdFn)
 
 				if tt.command == "" {
-					tt.command = "find"
+					tt.command = "list"
 				}
 
 				cmd.SetArgs(append([]string{"bucket", tt.command}, tt.flags...))
--- a/cmd/influx/bucket_test.go
+++ b/cmd/influx/bucket_test.go
@@ -177,7 +177,7 @@ func TestCmdBucket(t *testing.T) {
 		}
 	})
 
-	t.Run("find", func(t *testing.T) {
+	t.Run("list", func(t *testing.T) {
 		type called struct {
 			name  string
 			id    influxdb.ID
@@ -189,6 +189,7 @@ func TestCmdBucket(t *testing.T) {
 			name     string
 			expected called
 			flags    []string
+			command  string
 			envVars  map[string]string
 		}{
 			{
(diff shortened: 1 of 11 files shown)


jademcgough:
@rawkode good call

jsteenb2:
we're going to want to update the tests as well to verify it hits the list endpoint, we may also want to run these same tests for the `find` cmd to maintain that backweards compatibility in our testing. Example here:

https://github.com/influxdata/influxdb/blob/master/cmd/influx/organization_test.go#L78

cmd/influx/bucket_test.go
```diff
@@ -289,7 +304,11 @@ func TestCmdBucket(t *testing.T) {
 				cmdFn, calls := cmdFn()
 				cmd := builder.cmd(cmdFn)
 
-				cmd.SetArgs(append([]string{"bucket", "find"}, tt.flags...))
+				if tt.command == "" {
+					tt.command = "find"
```
jsteenb2: should this be `list` as the default?

jademcgough: yes, not sure what I was thinking there

cmd/influx/organization_test.go
```diff
@@ -411,6 +431,8 @@ func TestCmdOrg(t *testing.T) {
 			}
 
 			testMemberFn(t, "list", cmdFn, tests...)
+			testMemberFn(t, "ls", cmdFn, tests[:0]...)
```
jsteenb2: I don't think this is testing anything here :thinking: , did you mean to put `testMemberFn(t, "ls", tests...)`?

`tests[:0] == empty slice`

cmd/influx/secret_test.go
```diff
@@ -85,7 +101,12 @@ func TestCmdSecret(t *testing.T) {
 				)
 				nestedCmdFn, calls := cmdFn()
 				cmd := builder.cmd(nestedCmdFn)
-				cmd.SetArgs(append([]string{"secret", "find"}, tt.flags...))
+
+				if tt.command == "" {
+					tt.command = "find"
```
jsteenb2: default to `list` here?

cmd/influx/user_test.go
```diff
@@ -269,7 +290,12 @@ func TestCmdUser(t *testing.T) {
 				)
 				nestedCmdFn, calls := cmdFn()
 				cmd := builder.cmd(nestedCmdFn)
-				cmd.SetArgs(append([]string{"user", "find"}, tt.flags...))
+
+				if tt.command == "" {
+					tt.command = "find"
```
jsteenb2: same comment about defaults

cmd/influx/organization_test.go
jademcgough: I wanted the first test, since running it against every test seems unnecessary. Fixing.

PR #16831, opened 2020-02-11
chore(ui): moved logic into thunk reducer

### LINKED ISSUE #15956: Save As dashboard reports success but dashboard isn't created
to repro, in data explorer, create a query, save as a dashboard cell, create a new dashboard, name it and the cell, then create. It always returns a success even if the user doesn't have permissions to create a dashboard (read only user).

Asalem1:
Closes #15956

Problem
Read-only users were being notified that created dashboards were successful even though no dashboards were being created

Solution
The dashboard creation process was ingesting any potential errors thrown throughout the process so the default success notification was being triggered. Continued bubbling errors so that the ultimate error was being caught

### DIFF
--- a/ui/src/cells/actions/thunks.ts
+++ b/ui/src/cells/actions/thunks.ts
@@ -135,7 +135,7 @@ export const createCellWithView = (
   }
 }
 
-export const createDashboardWithView = async (
+export const createDashboardWithView = (
   orgID: string,
   dashboardName: string,
   view: View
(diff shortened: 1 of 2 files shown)


ui/src/cells/actions/thunks.ts
```diff
@@ -129,11 +135,37 @@ export const createCellWithView = (
...
+      cells: [],
+    }
+
+    const resp = await postDashboard({data: newDashboard})
+
+    if (resp.status !== 201) {
+      throw new Error(resp.data.message)
+    }
+
+    await createCellWithView(resp.data.id, view)
```
drdelambre: wait this is supposed to be 'dispatch(createCellWithView())`

Asalem1: you're totally right

PR #16836, opened 2020-02-12
refactor(services): start services asynchronously

ayang64:
refactor(service): start services asynchronously
    
These changes allow us to start services (`tsdb`, `http`, `monitor`,  continuous query, etc.) asynchronously so that we can access the http server before the tsdb service is ready.  Before this patch, the http service was loaded after tsdb which makes it impossible to retrieve profiling information that could help diagnose issues during tsdb startup.

This patch replaces the `Open()`/`Close()` metaphor used for starting services with simply `Run()` with concurrency handled by the caller and cancellation done via a context.  It also removes the assumption that `Run()`, `Open()`, and `Close()` can be called more than once on a particular value. The reason for this change is that, in practice, it is very easy to ensure that Run() is only called once and the code to track the state of a service (if it is running or not) adds unnecessary overhead.

This patch also provides a mostly backwards compatible API (`Open()` and `Close()`) so that we can preserve old tests.  

Following the principle of leaving concurrency to the caller, this patch adds a Run() method which accepts a context value and blocks until it is cancelled.  This provides a standard and idiomatic way of cancelling each service. 

The `cancel()` function associated with a context now serves the purpose of the `Close()` method -- once a service starts, it blocks on `ctx.Done()` and performs any necessary cleanup once the go routine is cancelled.

This patch also adds a services package which we use to coordinate and share information between services.

- [x] Rebased/mergeable
- [ ] Tests pass


cmd/influxd/run/server.go
```diff
@@ -436,88 +465,88 @@ func (s *Server) Open() error {
...
 	if s.Listener != nil {
 		s.Listener.Close()
 	}
 
-	// Close services to allow any inflight requests to complete
-	// and prevent new requests from being accepted.
-	for _, service := range s.Services {
-		service.Close()
+	// wait for services to complete.
+	for i, end := 0, len(s.Services); i < end; i++ {
```
e-dard: Let's simplify this a bit, for clarity?

```go
for _ = range s.Services {
    // receive signal when service has ended.
    sem <- struct{}{}
}
```

cmd/integration_config_test.go
```diff
@@ -22,7 +24,10 @@ func TestRetentionAutocreate(t *testing.T) {
 			})
 			defer cmd.Cleanup()
 
-			cmd.MustRun()
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			go cmd.MustRun(ctx)
+			time.Sleep(1 * time.Second) // give cmd.MustRun() time to start
```
e-dard: `time.Second`

cmd/integration_config_test.go
```diff
@@ -57,7 +62,10 @@ func TestCacheMaxMemorySize(t *testing.T) {
 	})
 	defer cmd.Cleanup()
 
-	cmd.MustRun()
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go cmd.MustRun(ctx)
+	time.Sleep(1 * time.Second) // give cmd.MustRun() time to start
```
e-dard: `time.Second`

monitor/service.go
```diff
@@ -210,6 +224,8 @@ func (m *Monitor) SetPointsWriter(pw PointsWriter) error {
 	m.mu.Unlock()
 
 	// Subsequent calls to an already open Monitor are just a no-op.
+
+	// FIXME: should we be running m.Open() here?
```
e-dard: `FIXME` or `TODO`?

services/collectd/service.go
```diff
@@ -17,11 +19,16 @@ import (
...
 	"github.com/influxdata/influxdb/logger"
 	"github.com/influxdata/influxdb/models"
+	"github.com/influxdata/influxdb/services"
 	"github.com/influxdata/influxdb/services/meta"
 	"github.com/influxdata/influxdb/tsdb"
 	"go.uber.org/zap"
 )
 
+func init() {
+	log.SetFlags(log.LstdFlags | log.Lshortfile)
```
e-dard: We don't normally set none-default logging flags as far as I know. Any particular reason for this?

services/collectd/service.go
```diff
@@ -336,32 +358,49 @@ func (s *Service) serve() {
...
-		if err != nil {
-			if strings.Contains(err.Error(), "use of closed network connection") {
-				select {
-				case <-s.done:
-					return
-				default:
-					// The socket wasn't closed by us so consider it an error.
+		case cur := <-reads:
+			if cur.err != nil {
+				if strings.Contains(rp.err.Error(), "use of closed network connection") {
```
e-dard: this method of checking for an error is unfortunate. I don't know what would be better though in this case.

services/collectd/service.go
```diff
@@ -336,32 +358,49 @@ func (s *Service) serve() {
 	//   Ethernet).
 	buffer := make([]byte, 1452)
 
+	type ReadPayload struct {
+		n   int   // bytes read by ReadFromUDP
+		err error // any error returend by ReadFromUDP
+	}
+
+	rp := ReadPayload{}
```
e-dard: nit: `var rp ReadPayload`

services/collectd/service.go
```diff
@@ -336,32 +358,49 @@ func (s *Service) serve() {
...
+
+	reads := make(chan ReadPayload)
+
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				rp.n, _, rp.err = s.conn.ReadFromUDP(buffer)
```
e-dard: I spent five minutes looking this code because initially it looked racy to me. Two goroutines accessing the same value unsynchronised, e.g., `rp.n`. Because `ReadPayload.n` is copied when it's passed to the channel there isn't a race, but it's a bit nuanced. Maybe worth a comment for future folks looking at this code?

services/continuous_querier/service.go
```diff
@@ -274,7 +271,7 @@ func (s *Service) runContinuousQueries(req *RunRequest) {
 // ExecuteContinuousQuery may execute a single CQ. This will return false if there were no errors and the CQ was not run.
 func (s *Service) ExecuteContinuousQuery(dbi *meta.DatabaseInfo, cqi *meta.ContinuousQueryInfo, now time.Time) (bool, error) {
 	// TODO: re-enable stats
-	//s.stats.Inc("continuousQueryExecuted")
+	// s.stats.Inc("continuousQueryExecuted")
```
e-dard: remove this?

services/opentsdb/service.go
```diff
@@ -297,20 +341,48 @@ func (s *Service) Addr() net.Addr {
...
+func (s *Service) serve(ctx context.Context) error {
+	conChan := make(chan net.Conn)
+	errChan := make(chan error)
+
+	// send accepted connections to our conChan so that we can multiplex over
+	// conChand and ctx.Done() in the next for loop.
+	//
+	// this go routine is cancelled once the s.ln listener is closed.
+	//
+	// FIXME: this logic needs to be looked at closely.
```
e-dard: > this logic needs to be looked at closely.

looked at closely, or tested comprehensively and they called good? 😄

cmd/influxd/main.go
```diff
@@ -64,47 +65,67 @@ func NewMain() *Main {
...
+	defer cancel()
+	go func() { errCh <- cmd.Run(ctx, args...) }()
+
+	// wait for message on errCh or signalCh
+	select {
+	case err := <-errCh:
+		// if we've received a value on our errCh at this point, then the
+		// server has stopped and we should exit.
+		return err
+
```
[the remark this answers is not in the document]
ayang64: makes sense.

cmd/influxd/main.go
```diff
@@ -64,47 +65,67 @@ func NewMain() *Main {
...
+	case <-signalCh:
+		// if we received a signal, lets cancel our services.
+		cmd.Logger.Info("Signal received, initializing clean shutdown...")
+		cancel()
+
+		// we store the deadline duration so that we can print it later.
+		deadline := time.Second * 30
```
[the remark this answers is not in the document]
ayang64: i store the duration in `deadline` so that we can print its value later if it is exceeded.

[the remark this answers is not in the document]
ayang64: makes sense -- i'll swap them.

cmd/influxd/run/server.go
ayang64: d'oh.  i'm not sure why i did it this way.

ayang64: changed to:

```
for range s.Services {
  ...
}
```

e-dard: ha indeed! 🤦‍♂

services/collectd/service.go
ayang64: this is stray code.  i did this for testing and forgot about it.  will remove.

services/collectd/service.go
ayang64: agreed.

services/continuous_querier/service.go
ayang64: not even sure where this came from.  will remove it.

services/collectd/service.go
ayang64: ya -- i don't like it.  i'm mostly preserving how it was done previously.

services/collectd/service.go
ayang64: oh -- that's a bug.  `rp` should be `cur` in some parts.  will fix.

services/opentsdb/service.go
ayang64: just looked at. not even read.  just look at the lines and curves that form the letters and words but without assigning any meaning to them.

monitor/service.go
ayang64: ya -- it seems like a bug to me that we're running m.Open() in this place.  it seems like something that needs to be examined with an eye for fixing it.

services/collectd/service.go
stuartcarnie: This is racy on `buffer` as there are multiple concurrent accesses with one being a write.

image

services/graphite/service.go
```diff
@@ -148,24 +166,32 @@ func (s *Service) Open() error {
...
+	case "tcp":
 		s.addr, err = s.openTCPServer()
-	} else if strings.ToLower(s.protocol) == "udp" {
+	case "udp":
 		s.addr, err = s.openUDPServer()
-	} else {
+	default:
 		return fmt.Errorf("unrecognized Graphite input protocol %s", s.protocol)
 	}
+
```
stuartcarnie: 👏

services/opentsdb/service.go
```diff
@@ -297,20 +341,48 @@ func (s *Service) Addr() net.Addr {
...
+				s.Logger.Info("OpenTSDB TCP listener closed")
+				errChan <- err
+				return
+			} else if err != nil {
+				s.Logger.Info("Error accepting OpenTSDB", zap.Error(err))
+				continue
+			}
+			conChan <- c
 		}
+	}()
```
stuartcarnie: My understanding for moving the listener to a separate goroutine and forwarding the new connections over a channel is so that you can read the `context.Context` `Done` channel for cancellation. 

Perhaps in a future update, we could simplify this code by changing `RunWithReady` and build the `Listener` instances with `context.Context`:

```go
var lc ListenConfig
s.ln = lc.Listen(ctx, "tcp", s.BindAddress)
```

as `net.Listen` is:

```go
func Listen(network, address string) (Listener, error) {
	var lc ListenConfig
	return lc.Listen(context.Background(), network, address)
}
```

services/services.go
```diff
@@ -0,0 +1,92 @@
...
+		s:  make(map[string]struct{}),
+		ch: make(map[string]chan struct{}),
+	}
+}
+
+// Implementation of the Registrar interface.  There are no exported fields.
+type Registry struct {
+	mu sync.RWMutex             // rw mutex used to protect our s map
+	s  map[string]struct{}      // map storing service names
+	ch map[string]chan struct{} // map storing service names
```
stuartcarnie:
```suggestion
	ch map[string]chan struct{} // map storing service wait channels
```

services/udp/service.go
```diff
@@ -174,45 +197,57 @@ func (s *Service) writer() {
...
+				s.Logger.Info("Failed to read UDP message", zap.Error(payload.err))
 				continue
 			}
-			atomic.AddInt64(&s.stats.BytesReceived, int64(n))
+			atomic.AddInt64(&s.stats.BytesReceived, int64(payload.nbytes))
 
-			bufCopy := make([]byte, n)
-			copy(bufCopy, buf[:n])
+			bufCopy := make([]byte, payload.nbytes)
+			copy(bufCopy, buf[:payload.nbytes])
```
stuartcarnie: These two goroutines will race on reads / writes to the `buf`, which is read and written to concurrently.

services/collectd/service.go
e-dard: Should have just gone with my gut heh

cmd/influxd/main.go
```diff
@@ -64,47 +65,66 @@ func NewMain() *Main {
...
+		select {
+		case err := <-errCh:
+			// cmd.Run() has completed.
+			return err
+		case <-signalCh:
+			// we got another signal before cmd.Run() has gracefully shut down.
+			return fmt.Errorf("got another signal; forcibly shutting down")
+		case <-t.C:
+			// timeout expired before we cleanly shutdown
+			return fmt.Errorf("server failed to shutdown after %v", deadline)
```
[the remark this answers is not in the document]
ayang64: actually, %v gives me what i want which is a string representation of the duration.

services/udp/service.go
ayang64: fixed!

PR #16842, opened 2020-02-12
feat(lsp): Add lsp wasm and connect to monaco

### LINKED ISSUE #16794: Establish first contact with LSP, enable static completion

ebb-tide:
Closes #16794


ui/src/external/monaco.fluxCompletions.ts
```diff
@@ -1,20 +1,83 @@
...
+): IDisposable => {
+  const completionProvider = monaco.languages.registerCompletionItemProvider(
+    FLUXLANGID,
+    {
+      provideCompletionItems: (model, position, context) => {
+        const wordUntil = model.getWordUntilPosition(position)
+        const defaultRange = new monaco.Range(
+          position.lineNumber,
+          wordUntil.startColumn,
+          position.lineNumber,
```
drdelambre: what if the range is across lines?

ui/src/external/monaco.fluxLang.ts
```diff
@@ -1,7 +1,9 @@
+import {FLUXLANGID} from 'src/types'
```
121watts: nit: `FLUXLANGID` isn't a type its a value.  Should probably live in a `constants` folder.

ui/src/external/monaco.fluxCompletions.ts
```diff
@@ -1,20 +1,83 @@
...
-      ] as any[]
-      return {suggestions: suggestions}
-    },
-  })
+        if (!completionItems) {
+          return
+        }
+        return p2m.asCompletionResult(completionItems, defaultRange)
+      },
+      triggerCharacters: [
```
drdelambre: whaaaaat? totally worth an /* eslint-disable */

ui/src/external/monaco.fluxSyntax.ts
```diff
@@ -25,7 +23,11 @@ export async function addSyntax(monaco: MonacoType) {
 
   // map of monaco "language id's" to TextMate scopeNames
   const grammars = new Map()
-  grammars.set('flux', 'flux')
+  grammars.set(FLUXLANGID, FLUXLANGID)
+
+  monaco.languages.setLanguageConfiguration(FLUXLANGID, {
+    autoClosingPairs: [{open: '"', close: '"'}, {open: '[', close: ']'}],
+  })
```
121watts: what about `'` and `}` and `)`?

ui/src/shared/components/FluxMonacoEditor.tsx
```diff
@@ -1,18 +1,24 @@
 // Libraries
-import React, {FC} from 'react'
+import React, {FC, useEffect, useRef, useState} from 'react'
```
drdelambre: so fancy

ui/src/shared/components/FluxMonacoEditor.tsx
```diff
@@ -28,18 +34,39 @@ interface Props {
...
+  }, [])
+
+  useEffect(() => {
+    return function cleanup() {
+      completionProvider.dispose()
+    }
+  }, [])
+
   const editorWillMount = (monaco: MonacoType) => {
+    monaco.languages.register({id: FLUXLANGID})
```
drdelambre: what happens if you call this a bunch of times, does it reset any of the registrations that happen below?

ui/src/shared/components/FluxMonacoEditor.tsx
```diff
@@ -28,18 +34,39 @@ interface Props {
...
+
+  useEffect(() => {
+    sendMessage(initialize, lspServer.current)
+    sendMessage(didOpen(script), lspServer.current)
+  }, [])
+
+  useEffect(() => {
+    return function cleanup() {
+      completionProvider.dispose()
+    }
```
121watts: This `cleanup` function doesn't need a separate `useEffect`. It can go in between line 50 and 51.

ui/webpack.common.ts
```diff
@@ -29,14 +30,29 @@ module.exports = {
     },
     extensions: ['.tsx', '.ts', '.js', '.wasm'],
   },
+  node: {
```
drdelambre: i know this was required for the vendor build, but is it required for the dist build?

ui/webpack.common.ts
```diff
@@ -29,14 +30,29 @@ module.exports = {
...
+    process: true,
+    module: false,
+    clearImmediate: false,
+    setImmediate: true,
+  },
   module: {
     rules: [
       {
-        test: /\flux_parser_bg.wasm$/,
+        test: /flux_parser_bg.wasm$/,
```
drdelambre: why not join these two?

ui/webpack.common.ts
```diff
@@ -120,6 +136,10 @@ module.exports = {
       API_PREFIX: API_BASE_PATH,
       STATIC_PREFIX: BASE_PATH,
     }),
+    new MonacoWebpackPlugin({
+      languages: ['json', 'javascript', 'go', 'markdown'],
```
drdelambre: did we theme markdown and go? (read: implement the textmate grammer definition and color scheme)

ui/src/shared/components/FluxMonacoEditor.tsx
ebb-tide: I think it might but `editorWillMount` only runs once?

ui/webpack.common.ts
ebb-tide: yup

ui/webpack.common.ts
ebb-tide: I couldn't get the regex to work.

ui/webpack.common.ts
ebb-tide: I don't believe so. I also think we should not have javascript and go now that I think about it.

ui/src/external/monaco.fluxCompletions.ts
ebb-tide: I don't think we provide completion items for multi line selections? it only happens while user is typing. so I don't think it's a problem.

ui/src/shared/components/FluxMonacoEditor.tsx
ebb-tide: good point, thanks!

ui/src/external/monaco.fluxSyntax.ts
ebb-tide: thank you!

PR #16843, opened 2020-02-12
feat(mock): Add SeriesGenerator to ResultSet transformation

stuartcarnie:
This PR adds to helpers for working with `reads.ResultSet` and clients which require a `ResultSet` as input.

`mock.NewResultSetFromSeriesGenerator` is a function for transforming a `SeriesGenerator` from the `gen` package into a `ResultSet`. This is useful for mocking data and testing code which requires a `reads.ResultSet` instance.

Additionally, adds a `reads.ResultSetToLineProtocol` function, which transforms a `ResultSet` into InfluxDB line protocol. This is useful to simplify comparing the data from a `ResultSet` with a string.

### DIFF
--- a/mock/reads_resultset.go
+++ b/mock/reads_resultset.go
@@ -1,0 +1,167 @@
+package mock
+
+import (
+	"github.com/influxdata/influxdb/models"
+	"github.com/influxdata/influxdb/pkg/data/gen"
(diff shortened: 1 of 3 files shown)


gianarb:
lovely!

PR #16849, opened 2020-02-12
Zs gaugecontainerfit15911

### LINKED ISSUE #15911: Gauge doesn't always fit container
guage-scales-poorly

All numeric values on a gauge should always be visible.

zoesteinkamp:
Closes #15911 

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/src/dashboards/components/Dashboard.tsx
+++ b/ui/src/dashboards/components/Dashboard.tsx
@@ -57,7 +57,6 @@ class DashboardComponent extends PureComponent<Props> {
               timeRange={timeRange}
               manualRefresh={manualRefresh}
               onPositionChange={onPositionChange}
-              {...this.props}
             />
           ) : (
             <DashboardEmpty onAddCell={onAddCell} />
--- a/ui/src/dashboards/components/Dashboard.tsx
+++ b/ui/src/dashboards/components/Dashboard.tsx
@@ -57,6 +57,7 @@ class DashboardComponent extends PureComponent<Props> {
               timeRange={timeRange}
               manualRefresh={manualRefresh}
               onPositionChange={onPositionChange}
+              {...this.props}
             />
           ) : (
             <DashboardEmpty onAddCell={onAddCell} />
--- a/ui/src/shared/components/cells/Cells.tsx
+++ b/ui/src/shared/components/cells/Cells.tsx
@@ -33,16 +33,16 @@ interface StateProps {
   state: AppState
 }
 
-interface Props {
+interface OwnProps {
   cells: Cell[]
   timeRange: TimeRange
   manualRefresh: number
   onPositionChange?: (cells: Cell[]) => void
 }
-type AllProps = StateProps & Props
+type Props = StateProps & OwnProps & WithRouterProps
 
 @ErrorHandling
-class Cells extends Component<AllProps & WithRouterProps> {
+class Cells extends Component<Props> {
   public render() {
     const {cells, timeRange, manualRefresh} = this.props
 
@@ -145,10 +145,9 @@ class Cells extends Component<AllProps & WithRouterProps> {
 const mstp = (state: AppState): StateProps => {
   return {state}
 }
-export default compose(
-  withRouter,
-  connect<StateProps, {}, Props>(
+export default withRouter<OwnProps>(
+  connect<StateProps, {}, OwnProps>(
     mstp,
     null
-  )
-)(Cells)
+  )(Cells)
+)
--- a/ui/src/shared/components/cells/Cells.tsx
+++ b/ui/src/shared/components/cells/Cells.tsx
@@ -146,8 +146,5 @@ const mstp = (state: AppState): StateProps => {
   return {state}
 }
 export default withRouter<OwnProps>(
-  connect<StateProps, {}, OwnProps>(
-    mstp,
-    null
-  )(Cells)
+  connect<StateProps, {}, OwnProps>(mstp)(Cells)
 )
--- a/ui/src/shared/components/cells/Cells.tsx
+++ b/ui/src/shared/components/cells/Cells.tsx
@@ -1,5 +1,7 @@
 // Libraries
 import React, {Component} from 'react'
+import {connect} from 'react-redux'
+import {compose} from 'redux'
 import {withRouter, WithRouterProps} from 'react-router'
 import ReactGridLayout, {WidthProvider, Layout} from 'react-grid-layout'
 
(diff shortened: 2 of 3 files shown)


ui/src/dashboards/components/Dashboard.tsx
```diff
@@ -57,6 +57,7 @@ class DashboardComponent extends PureComponent<Props> {
               timeRange={timeRange}
               manualRefresh={manualRefresh}
               onPositionChange={onPositionChange}
+              {...this.props}
```
Asalem1: let's expand this. Using the spread operator leaves it unclear as to what we're actually passing down to the child component

ui/src/shared/components/cells/Cells.tsx
```diff
@@ -10,24 +12,37 @@ import GradientBorder from 'src/shared/components/cells/GradientBorder'
...
+  state: AppState
+}
+
 interface Props {
   cells: Cell[]
   timeRange: TimeRange
   manualRefresh: number
   onPositionChange?: (cells: Cell[]) => void
 }
+type AllProps = StateProps & Props
```
Asalem1: if we're creating a custom type here, we should add `WithRouterProps` as an interface here:

`type AllProps = StateProps & Props & WithRouterProps` 

That way we can write the component as such:

```
class Cells extends Component<AllProps> {
...
```

ui/src/shared/components/cells/Cells.tsx
```diff
@@ -65,16 +80,24 @@ class Cells extends Component<Props & WithRouterProps> {
...
-      .map(c => ({
-        ...c,
-        x: c.x,
-        y: c.y,
-        h: c.h,
-        w: c.w,
-        i: c.id,
-      }))
+      .map(c => {
+        const view = getByID<View>(state, ResourceType.Views, c.id)
```
Asalem1: great job. Exactly what I thought, surprised there wasn't a better way but it makes sense

ui/src/shared/components/cells/Cells.tsx
```diff
@@ -65,16 +80,24 @@ class Cells extends Component<Props & WithRouterProps> {
...
+        let cell = {
```
Asalem1: this should be a `const`

ui/src/shared/components/cells/Cells.tsx
```diff
@@ -65,16 +80,24 @@ class Cells extends Component<Props & WithRouterProps> {
...
+          ...c,
+          x: c.x,
+          y: c.y,
+          h: c.h,
+          w: c.w,
+          i: c.id,
+        }
+        if (view.properties.type === 'gauge') {
+          cell = {...cell, minW: 3, minH: 3}
```
Asalem1: instead of using the spread operator here, we can simply say:

```
cell.minW = 3
cell.minH = 3
```

This will also be true if the variable is a constant

ui/src/shared/components/cells/Cells.tsx
```diff
@@ -119,5 +142,13 @@ class Cells extends Component<Props & WithRouterProps> {
     }
   }
 }
-
-export default withRouter(Cells)
+const mstp = (state: AppState): StateProps => {
+  return {state}
```
Asalem1: Why are we returning the entire state as state here? Which properties are we trying to access?

ui/src/shared/components/cells/Cells.tsx
zoesteinkamp: done

ui/src/shared/components/cells/Cells.tsx
zoesteinkamp: sometimes the best path is the easiest

ui/src/shared/components/cells/Cells.tsx
zoesteinkamp: It cants be, because we edit it below

ui/src/shared/components/cells/Cells.tsx
zoesteinkamp: That did not work... it seemed quite angry. Is it supposed to be cell= {cell.minH etc...}

ui/src/shared/components/cells/Cells.tsx
Asalem1: an obj (and array) can be explicitly reassigned, even as a const.
```
const obj = {a: 1}
console.log(obj.a) // 1
obj.a = 2
console.log(obj.a) // 2
obj.b = 3
console.log(obj) // {a: 2, b: 3}
```

ui/src/shared/components/cells/Cells.tsx
zoesteinkamp: Because we want the entire state? I dont think we can do anything but pass the entire state in

ui/src/shared/components/cells/Cells.tsx
Asalem1: by reassigning the minH and minW, we be overwriting the actual object, so in the end the cell would only have the properties that we've reassigned. What's the error saying? It could be a TypeScript error that's being protective about cell being too loose a type, in which case it would be best for us to create a type for the cell

ui/src/dashboards/components/Dashboard.tsx
zoesteinkamp: Removed

ui/src/shared/components/cells/Cells.tsx
Asalem1: we don't have to pass anything at all. It doesn't seem like we're actually using the redux state or actions to update any specific logic

ui/src/shared/components/cells/Cells.tsx
```diff
@@ -119,5 +141,9 @@ class Cells extends Component<Props & WithRouterProps> {
...
+}
+export default withRouter<OwnProps>(
```
Asalem1: Since we don’t seem to be using redux, we can remove the connect here

ui/src/shared/components/cells/Cells.tsx
```diff
@@ -1,5 +1,6 @@
 // Libraries
 import React, {Component} from 'react'
+import {connect} from 'react-redux'
```
Asalem1: Since we don’t seem to be using redux, we can remove the connect here

zoesteinkamp: got it

ui/src/shared/components/cells/Cells.tsx
zoesteinkamp: got it

PR #16855, opened 2020-02-13
feat(ui): added labels to buckets

### LINKED ISSUE #16740: Add labels to buckets
Currently although the backend supports the buckets resource having labels associated with them, the UI does not display labels on buckets nor does it allow users to add/modify/delete labels associated with buckets. 
It should

Asalem1:
Closes #16740 

### Problem

Buckets did not have labels 

### Solution

Added labels to buckets, normalized some data, add buckets label test

bucket_labels

### DIFF
--- a/ui/cypress/e2e/buckets.test.ts
+++ b/ui/cypress/e2e/buckets.test.ts
@@ -31,7 +31,7 @@ describe('Buckets', () => {
       })
 
       cy.getByTestID(`bucket--card--name ${newBucket}`).should('exist')
-
+      cy.getByTestID(`bucket-card ${newBucket}`).should('exist')
       // Add a label
       cy.getByTestID(`bucket-card ${newBucket}`).within(() => {
         cy.getByTestID('inline-labels--add').click()
--- a/ui/cypress/e2e/buckets.test.ts
+++ b/ui/cypress/e2e/buckets.test.ts
@@ -33,9 +33,9 @@ describe('Buckets', () => {
       cy.getByTestID(`bucket--card--name ${newBucket}`).should('exist')
 
       // Add a label
-      cy.getByTestID(`bucket-card ${newBucket}`).within(() => {
-        cy.getByTestID('inline-labels--add').click()
-      })
+      cy.getByTestID('inline-labels--add')
+        .first()
+        .click()
 
       const labelName = 'l1'
       cy.getByTestID('inline-labels--popover--contents').type(labelName)
@@ -78,8 +78,9 @@ describe('Buckets', () => {
         cy.getByTestID('name-sorter')
           .click()
           .then(() => {
-            cy.get('.cf-resource-name').each((val, index) => {
-              expect(val.text()).to.include(buckets[index])
+            cy.get('[data-testid*="bucket-card"]').each((val, index) => {
+              const testID = val.attr('data-testid')
+              expect(testID).to.include(buckets[index])
             })
           })
 
@@ -87,8 +88,9 @@ describe('Buckets', () => {
           .click()
           .then(() => {
             const asc_buckets = buckets.slice().sort()
-            cy.get('.cf-resource-name').each((val, index) => {
-              expect(val.text()).to.include(asc_buckets[index])
+            cy.get('[data-testid*="bucket-card"]').each((val, index) => {
+              const testID = val.attr('data-testid')
+              expect(testID).to.include(asc_buckets[index])
             })
           })
 
--- a/ui/cypress/e2e/buckets.test.ts
+++ b/ui/cypress/e2e/buckets.test.ts
@@ -74,26 +74,23 @@ describe('Buckets', () => {
 
     describe('Searching and Sorting', () => {
       it('can sort by name and retention', () => {
-        cy.getByTestID('name-sorter').click()
+        const buckets = ['defbuck', '_tasks', '_monitoring']
+        cy.getByTestID('name-sorter')
+          .click()
+          .then(() => {
+            cy.get('.cf-resource-name').each((val, index) => {
+              expect(val.text()).to.include(buckets[index])
+            })
+          })
 
-        cy.get('.cf-resource-card')
-          .first()
-          .contains('defbuck')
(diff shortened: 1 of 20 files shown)


ui/src/buckets/components/UpdateBucketOverlay.tsx
```diff
@@ -61,7 +61,7 @@ const UpdateBucketOverlay: FunctionComponent<Props> = ({
         handleClose()
         return
       }
-      setBucketDraft(resp.data)
+      setBucketDraft(resp.data as Bucket)
```
Asalem1: setting this type here as a way of resolving a TS error that stemmed from a disparity between the expectations of the `useState` func and the data type being returned by the `getBucket` function

ui/cypress/e2e/buckets.test.ts
```diff
@@ -61,27 +75,28 @@ describe('Buckets', () => {
     describe('Searching and Sorting', () => {
       it('can sort by name and retention', () => {
         cy.getByTestID('name-sorter').click()
-        cy.getByTestID('bucket-card')
+
+        cy.get('.cf-resource-card')
```
drdelambre: why did these get less specific? that's a trigger to go add testID params.

Asalem1: Great question. The original was created to test out the sort order. In order to validate those assumptions, we were getting all the bucket-cards (previously by a less specific test-ID that worked). Then we were clicking the sort order to see if the first value changed (since it should've been reflected in the UI). I tried creating a solution where we were getting the resource list body and seeing if the children were in a particular order, but couldn't get that configured in cypress. The solution I came up with was to revert to the original construct (generalized grab all the cards), then see if the first one matched our expectations when changing the sort order

drdelambre: are you talking about not being able to address this:
https://github.com/influxdata/influxdb/pull/16855/files#diff-aef71c9cb50e63e41a914808cdd0aac6R54

with something like this?:
https://github.com/influxdata/influxdb/blob/master/ui/cypress/e2e/collectors.test.ts#L228

Asalem1: Not really. So the label tests are working magically without any issue because of the specificity added to the test-IDs. 

What you linked seems to be a much better way of addressing the tests. I'll go ahead and refactor it so it looks like:

https://github.com/influxdata/influxdb/blob/master/ui/cypress/e2e/collectors.test.ts#L236-L238

ui/src/buckets/components/BucketList.tsx
```diff
@@ -94,9 +95,11 @@ class BucketList extends PureComponent<Props & WithRouterProps, State> {
               testID="retention-sorter"
             />
           </ResourceList.Header>
-          <ResourceList.Body emptyState={emptyState}>
-            {this.listBuckets}
-          </ResourceList.Body>
+          <GetResources resources={[ResourceType.Labels]}>
```
drdelambre: can we put this next to where buckets are fetched?

Asalem1: @drdelambre when I did, the empty state was never fulfilled, so I moved it out one layer (by empty state, I mean there's a message that's typically displayed when there are no buckets output in a search)

drdelambre: here i think:
https://github.com/influxdata/influxdb/blob/a4e3dce135b5a205d9b51abe89928e0c07e20af0/ui/src/buckets/containers/BucketsIndex.tsx#L62

Asalem1: Totally, good eye

ui/cypress/e2e/buckets.test.ts
```diff
@@ -60,28 +74,26 @@ describe('Buckets', () => {
...
-
-        cy.getByTestID('retention-sorter').click()
-        cy.getByTestID('bucket-card')
-          .first()
-          .contains('defbuck')
+        const buckets = ['defbuck', '_tasks', '_monitoring']
+        cy.getByTestID('name-sorter')
+          .click()
+          .then(() => {
+            cy.get('.cf-resource-name').each((val, index) => {
```
drdelambre: still have to change this (and change `bucket + name` back to `bucket` on the testid in the component)

Asalem1: why is that? We changed it for greater specificity to access a specific bucket card.

drdelambre: so use that specificity in the test by generating ids to look up (instead of evaluating the text value). i think we're still straddling two solutions right now and just need to pick one and push it through.

Asalem1: @drdelambre i'm not quite sure I understand your solution, would you mind expanding upon that with an example of what you had in mind?

drdelambre: yeah. so you have the name of the resource in both the test id and in the text content of the dom node you are trying to pinpoint. So you have two sources of truth, and two directions to go to unify them:
1) revert the test id specificity in the component, and test that the value within the component (with `expect(val.text()).to.include(buckets[index])`) is equal to the resource name. that keeps things scoped to buckets (good) but also ties the representation of the key to the value of the key (bad).
2) don't test for the value of the node, but instead look for the test id that meets the key you are looking for( ```cy.getByTestId(`bucket ${cool name}`)``` ). this scopes it to the bucket (good) as well as decouples the representation from the state (good)

seems like you're leaning towards #2. either one is fine, it's just that generic resource selector has got to go.

Asalem1: but wouldn't the second solution defeat the purpose of the test? If we're trying to test the sorting mechanism to present a specific order, targeting a bucket by the name would only validate that the bucket exists. It wouldn't represent the order with which that bucket exists in the context of the page

Asalem1: Oh wait, I think I see what you're saying. Is it possible to compare two nodes?

PR #16871, opened 2020-02-13
fix(pkger): fix check for labels having conflict

jsteenb2:
just stumbled on this issue where labels were seen to be in conflict when they should not have been.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/pkger/models.go
+++ b/pkger/models.go
@@ -317,7 +317,7 @@ func (d DiffLabel) IsNew() bool {
 }
 
 func (d DiffLabel) hasConflict() bool {
-	return d.IsNew() || d.Old != nil && *d.Old != d.New
+	return !d.IsNew() && d.Old != nil && *d.Old != d.New
 }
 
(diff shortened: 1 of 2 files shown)


pkger/models.go
```diff
@@ -317,7 +317,7 @@ func (d DiffLabel) IsNew() bool {
 }
 
 func (d DiffLabel) hasConflict() bool {
-	return d.IsNew() || d.Old != nil && *d.Old != d.New
+	return !d.IsNew() && d.Old != nil && !reflect.DeepEqual(*d.Old, d.New)
```
kelwang: reflect.DeepEqual is a little expensive. Since their type is same, just make a simple func to compare both Color and desc.

jsteenb2: funny enough, i was just changing this :-), good catch 👌

jsteenb2: since the types are struct values, they can be compared directly and go will resolve the equality 👍

PR #16877, opened 2020-02-13
fix(backup): handle backup with no credentials file

### LINKED ISSUE #16723: Backup and Restore OSS: If Credentials file doesn't exist, then the User with the correct permissions receives a 500 error when running the `influx backup` command
**Issue:**
Within my OSS instance, when I attempt to run `influx backup` (with the default user account that has read and write permissions to everything), the following error is returned: `Error: An internal error has occurred`.

**Steps:**
1) Go to the `influxdb` repo and pull latest from `master`
2) If exists, remove your local `.influxdbv2` directory so that a new setup is allowed
3) Start influxd: `bin/$(uname -s | tr [:upper:] [:lower:])/influxd`
4) Create a new org, bucket, and user via the following request:
`curl -X POST \
  http://localhost:9999/api/v2/setup \
  -d '{
    "username": "my-user",
    "password": "password",
    "org": "my-org",
    "bucket": "my-bucket"
}'`

5) Copy the token that is generated from the previous request and, using the influx CLI, run the following command (make sure to replace the directory name and token values with your own values):
`influx backup -p directoryNameForNewBackup -t copiedTokenValue`

**Results:**
- The org, bucket, and user are all successfully created
- Per the `influxd` logs, the Cache snapshot is started and the Snapshot for path written
- The `influx backup` command returns the message: `Error: An internal error has occurred`

**Additional Troubleshooting:**
- I confirmed that the above created user token can be successfully used with the command `influx org find -t copiedTokenValue` and correctly returns the org ID and Name
- I created 2 additional super users at the org, and when I attempted to start a backup with those tokens, I correctly received the following error: `Error: read:authorizations is unauthorized` since user permissions are not fully implemented.

tmgordeeva:
Backups and restores should work whether or not the original installation uses
a credentials file and whether or not the backup contains a credentials file.

Closes #16723

### DIFF
--- a/http/backup_service.go
+++ b/http/backup_service.go
@@ -121,6 +121,7 @@ func (h *BackupHandler) handleCreate(w http.ResponseWriter, r *http.Request) {
 
 	if err != nil {
 		h.HandleHTTPError(ctx, err, w)
+		return
 	}
 
 	if credsExist {
--- a/http/backup_service.go
+++ b/http/backup_service.go
@@ -117,25 +117,15 @@ func (h *BackupHandler) handleCreate(w http.ResponseWriter, r *http.Request) {
 
 	files = append(files, bolt.DefaultFilename)
 
-	credBackupPath := filepath.Join(internalBackupPath, DefaultTokenFile)
+	credsExist, err := h.backupCredentials(internalBackupPath)
 
-	credPath, err := defaultTokenPath()
 	if err != nil {
 		h.HandleHTTPError(ctx, err, w)
-		return
-	}
-	token, err := ioutil.ReadFile(credPath)
-	if err != nil {
-		h.HandleHTTPError(ctx, err, w)
-		return
 	}
 
-	if err := ioutil.WriteFile(credBackupPath, []byte(token), 0600); err != nil {
-		h.HandleHTTPError(ctx, err, w)
-		return
+	if credsExist {
+		files = append(files, DefaultTokenFile)
 	}
-
-	files = append(files, DefaultTokenFile)
 
 	b := backup{
 		ID:    id,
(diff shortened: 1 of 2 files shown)


http/backup_service.go
```diff
@@ -117,26 +117,16 @@ func (h *BackupHandler) handleCreate(w http.ResponseWriter, r *http.Request) {
...
 
-	credPath, err := defaultTokenPath()
-	if err != nil {
-		h.HandleHTTPError(ctx, err, w)
-		return
-	}
-	token, err := ioutil.ReadFile(credPath)
 	if err != nil {
 		h.HandleHTTPError(ctx, err, w)
-		return
```
jacobmarble: Why was this return removed?

http/backup_service.go
```diff
@@ -117,26 +117,16 @@ func (h *BackupHandler) handleCreate(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	if err := ioutil.WriteFile(credBackupPath, []byte(token), 0600); err != nil {
-		h.HandleHTTPError(ctx, err, w)
-		return
```
jacobmarble: Why was this return removed?

http/backup_service.go
tmgordeeva: This was a mistake, fixing it now.

http/backup_service.go
tmgordeeva: This return will be covered by the (currently missing) one just above it.

jacobmarble:
@tmgordeeva I've tested this change by removing the credentials file before and after `influx backup`. That seems to work. However, `influxd restore` still fails:
```
$ influxd restore --backup-path backup
Restored Bolt to /Users/jacobmarble/.influxdbv2/influxd.bolt from backup/influxd.bolt
Error: failed to restore credentials file: no credentials file in backup: open backup/credentials: no such file or directory
Usage:
...
```

PR #16879, opened 2020-02-13
feat: adding in a simple regexp

### LINKED ISSUE #16168: Linked Table cells
__Proposal:__
If a table cell in a dashboard displays a string beginning with `http://` or `https://` it should automatically render as an anchor tag set to open in a new window. We can assume the user wants that cell to be a link because it is a URL

__Current behavior:__
Nothing

__Desired behavior:__
User can click a dashboard cell and have it open the URL in a new window

__Alternatives considered:__
Manually highlighting and copying the URL

__Use case:__
When looking through long running queries there are tables with links to a specific trace. It would reduce friction a little to not have to highlight and copy these links manually

drdelambre:
Closes #16168

this takes the cells within the table view and sees if they match a simple url regexp. if they do, it makes them a link.

Screenshot from 2020-02-13 14-04-13

before it's brought up: https://stackoverflow.com/questions/1500260/detect-urls-in-text-with-javascript/1500501#1500501

this isn't that big of a deal if it fails, as the interaction just reverts to normal, or it's a link that no one wants to go to anyways.

### DIFF
--- a/ui/src/shared/components/tables/TableCell.tsx
+++ b/ui/src/shared/components/tables/TableCell.tsx
@@ -31,6 +31,41 @@ interface Props extends CellRendererProps {
   timeFormatter: (time: string) => string
 }
 
+const URL_REGEXP = /(https?:\/\/[^\s]+)/g
+
+function asLink(str) {
+  if (!URL_REGEXP.test('' + str)) {
+    return str
+  }
+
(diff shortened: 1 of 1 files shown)


drdelambre:
:man_facepalming: i tried to name that one "notes are good"

PR #16880, opened 2020-02-13
feat(kv): add a kv mapper that allows the mapping of one bucket to another

lyondhill:


kv/mapper.go
```diff
@@ -0,0 +1,83 @@
...
+					counter++
+
+				}
+				if counter >= 100 {
+					seek = k
+					break
+				}
+			}
+
+			if err != nil {
```
GeorgeMac:
```suggestion
			if err := cursor.Err(); err != nil {
```

Issue #16809, opened 2020-02-11
Type unification error in algo W branch


wolffcm:
A few tests are failing with an error similar to this (tvar numbers may be different:
```
cannot unify {{}} with {username:t4064 | t4079}: cannot unify {{}} with {username:t4064 | t4079}
```

The affected tests are:
- `TestBoltNotificationRuleStore`
- `TestLauncher_Pkger/apply_a_package_with_env_refs`

The tests will be commented out with this issue URL in the `feat/use-algo-w` branch in influxdb.

PR #16883, opened 2020-02-14
build: remove -tags libflux and FLUX_PARSER_TYPE from build and circleci config

wolffcm:
This also prepends the `env` script (which sets CGO environment variables) that lets `staticcheck` pass.

The `continuous integration` tests will continue to fail until idpe is also green.

The `e2e` test failures seem related to notifications, that might be related to this?
https://github.com/influxdata/influxdb/issues/16809


adriandt:
My understanding is this: We are not ready for the rust parser, but we want to push ahead with ALGO-W, so remove rust parser from the branch as we prepare it for merging to master?

Is that right?

PR #16910, opened 2020-02-18
feat(kv): index utility type

GeorgeMac:
The purpose of this type is to wrap up our learnings for creating and managing indexes into a utility type.

This type owns:

1. Walking a foreign key in the index (read path - disabled by default).
2. Populating the index entirely (intended to be a one time operation before read path enabled).
3. Inserting and removing individual entries from the index (Create / Delete).
4. Verification of an index.

The verification path returns a diff which contains:

1. The set of things correctly indexed.
2. The set of things missing from the index.
3. The set of index entries which point to primary keys no longer present in the source.

It can also provide a list of foreign keys which are "corrupt" (via `diff.Corrupt()`). These are foreign keys which have some of their associated primary keys but not all. Useful in observing the extent of an unhealthy index.

By default, `Walk()` is a no-op. Intended to signify that a full scan is required as the index is empty. When the index is fully populated and all Insert / Delete operations have been correctly configured (i.e. Verify shows the index is healthy) the following option can be provided to `NewIndex` at creation to enable the read path: `WithIndexReadPathEnabled`.

The following is a condensed view of this PR's implementation of the Index's exposed functions.

```go
// IndexMapping is a type which configures and Index to map items
// from a source bucket to an index bucket via a mapping known as
// IndexSourceOn. This function is called on the values in the source
// to derive the foreign key on which to index each item.
type IndexMapping interface {
    SourceBucket() []byte
    IndexBucket() []byte
    IndexSourceOn(body []byte) (foreignKey []byte, err error)
}

// Index is used to define and manage an index for a source bucket.
//
// When using the index you must provide it with an IndexMapping.
// The IndexMapping provides the indexer with the contract it needs to populate
// the entire index and traverse a populated index correctly.
// The IndexMapping provides a way to retrieve the key on which to index with
// when provided with the value from the source.
// It also provides the way to access the source bucket.
type Index struct {
    // ...
}

// Insert creates a single index entry for the provided primary key on the foreign key.
func (i *Index) Insert(tx Tx, foreignKey, primaryKey []byte) error {}

// Delete removes the foreignKey and primaryKey mapping from the underlying index.
func (i *Index) Delete(tx Tx, foreignKey, primaryKey []byte) error {}

// VisitFunc is called for each k, v byte slice pair from the underlying source bucket
// which are found in the index bucket for a provided foreign key.
type VisitFunc func(k, v []byte) error

// Walk walks the source bucket using keys found in the index using the provided foreign key
// given the index has been fully populated.
func (i *Index) Walk(tx Tx, foreignKey []byte, fn VisitFunc) error {}

// Populate does a full population of the index using the provided IndexSourceOnFunc.
// Once completed it marks the index as ready for use.
// It return a nil error on success and the count of inserted items.
func (i *Index) Populate(context.Context, Store) (int,  error) {}

// Verify returns returns difference between a source and its index
// The difference contains items in the source that are not in the index
// and vice-versa.
func (i *Index) Verify(context.Context, Tx) (IndexDiff, error) {}
```

- [x] Rebased/mergeable
- [x] Tests pass


kv/index.go
```diff
@@ -0,0 +1,537 @@
...
+	"errors"
+	"fmt"
+)
+
+const defaultPopulateBatchSize = 100
+
+// Index is used to define and manage an index for a source bucket.
+//
+// When using the index you must provide it with an IndexSource.
+// The IndexSource provides the index with the contract it needs to populate
```
GeorgeMac: note to self: `s/IndexSource/IndexMapping/`.

kv/index.go
```diff
@@ -0,0 +1,537 @@
...
+//
+//      if err := json.Unmarshal(v, auth); err != nil {
+//          return err
+//      }
+//
+//      return auth.UserID.Encode()
+//  }
+//
+//  // configure a write only index
+//  indexByUser := NewIndex(NewSource([]byte(`authorizationsbyuserv1/), byUserID), false)
```
GeorgeMac: note to self: remove last argument `, false` as this is now default behaviour overridden by an option.

kv/index.go
```diff
@@ -0,0 +1,537 @@
...
+//
+//      return nil
+//  })
+//
+//  // populate entire index from source
+//  indexedCount, err := indexByUser.Populate(ctx, store)
+//
+//  // verify the current index against the source and return the differences
+//  // found in each
+//  err := indexByUser.Verify(ctx, tx)
```
GeorgeMac: note to self: this now returns an `IndexDiff`

kv/index.go
```diff
@@ -0,0 +1,537 @@
...
+
+// IndexPopulatorStore is a store which also has a AutoPopulateIndex method
+// The method returns true when the store support auto population of index
+// on initialize.
+type IndexPopulatorStore interface {
+	Store
+	AutoPopulateIndex() bool
+}
+
+// Initialize creates the index bucket on the provided store
```
GeorgeMac: note to self: document how given the store implements `AutoPopulateIndex()` and the method returns true, how ir automatically calls populate with the provided store.

kv/index.go
```diff
@@ -0,0 +1,537 @@
...
+// to derive the foreign key on which to index each item.
+type IndexMapping interface {
+	SourceBucket() []byte
+	IndexBucket() []byte
+	IndexSourceOn(value []byte) (foreignKey []byte, err error)
+}
+
+// IndexSourceOnFunc is a function which can be used to derive the foreign key
+// of a value in a source bucket.
+type IndexSourceOnFunc func([]byte) ([]byte, error)
```
stuartcarnie: Very nice 👌

kv/index.go
```diff
@@ -0,0 +1,537 @@
...
+
+func (i *Index) indexBucket(tx Tx) (Bucket, error) {
+	return tx.Bucket(i.IndexBucket())
+}
+
+func (i *Index) sourceBucket(tx Tx) (Bucket, error) {
+	return tx.Bucket(i.SourceBucket())
+}
+
+func indexKey(foreignKey, primaryKey []byte) (newKey []byte, _ error) {
```
stuartcarnie: nit: Given this always returns `nil` for `error`, it would be prudent to remove it to simplify call sites (remove the redundant `if err != nil` checks)

kv/index.go
```diff
@@ -0,0 +1,537 @@
...
+
+	return
+}
+
+// Insert creates a single index entry for the provided primary key on the foreign key.
+func (i *Index) Insert(tx Tx, foreignKey, primaryKey []byte) error {
+	newKey, err := indexKey(foreignKey, primaryKey)
+	if err != nil {
+		return err
+	}
```
stuartcarnie: nit: As mentioned earlier, by changing the `indexKey` return values, you can remove the redundant `if err != nil` check, given `indexKey` always returns `nil` for the `error` value.

kv/index.go
```diff
@@ -0,0 +1,537 @@
...
+
+	return indexWalk(cursor, sourceBucket, visitFn, func(fk, pk []byte) error {
+		// fail iteration when key not found for item in index
+		return fmt.Errorf("for key %v indexed by %v: %w", pk, fk, ErrKeyNotFound)
+	})
+}
+
+// Populate does a full population of the index using the provided IndexOnFunc.
```
stuartcarnie: Could you clarify what `IndexOnFunc` is? I cannot see where `IndexOnFunc` is provided in this context.

kv/index.go
```diff
@@ -0,0 +1,537 @@
...
+func (i *IndexDiff) Corrupt() (corrupt []string) {
+	for fk := range i.MissingFromIndex {
+		if _, ok := i.PresentInIndex[fk]; ok {
+			corrupt = append(corrupt, fk)
+		}
+	}
+	return
+}
+
+// Verify returns returns difference between a source and its index
```
stuartcarnie:
```suggestion
// Verify returns the difference between a source and its index
```

kv/index.go
GeorgeMac: ooh good catch. I went through a few iterations of this. Once upon a time the currency here with `influxdb.ID` rather than byte slices. But we have a few types which broke that contract (URMs in particular). This is cruft left behind from then.

kv/index.go
GeorgeMac: Ahh another bit of documentation fallout :+1:  . IndexSourceOnFunc used to be called IndexOnFunc. I added source for clarification.

GeorgeMac:
Small Update: I made some fairly sizeable changes to the way populate / verify fulfil their contracts. The tests are the same as we want the same behaviour.

I changed the strategy from a single transaction which curses over one bucket, while using `Get()` on another. Now it reads both buckets into memory and then avoids using the kv.Tx for many individual lookups. 

The original implementation presented a problem when large keyspaces combined with limits imposed by a particular kv.Store implementation regarding the number of operations which can be made in a single transaction.

PR #16922, opened 2020-02-19
feat(http): checks client

gavincabbage:
Adds a checks API client alongside checks API handlers.

This client does not implement `influxdb.CheckService` because that interface returns an `influxdb.Check` interface, which is not suitable for this use case (because, e.g. it provides a `GetTaskID` method, which would require embedding a task service inside the response object).

New unit tests are not added given this client is a thin wrapper on `httpc`. Tests for that package, as well as the handlers themselves, exercise all meaningful logic on these code paths.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/check_service.go
+++ b/http/check_service.go
@@ -6,12 +6,16 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"path"
 	"time"
+
+	"github.com/influxdata/influxdb/kit/tracing"
 
 	"github.com/influxdata/httprouter"
 	"github.com/influxdata/influxdb"
 	pctx "github.com/influxdata/influxdb/context"
 	"github.com/influxdata/influxdb/notification/check"
+	"github.com/influxdata/influxdb/pkg/httpc"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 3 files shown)


http/check_service.go
```diff
@@ -679,3 +683,209 @@ func (h *CheckHandler) handleDeleteCheck(w http.ResponseWriter, r *http.Request)
...
+		DecodeJSON(&cr).
+		Do(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return &cr, nil
+}
+
+func (s *CheckService) FindCheck(ctx context.Context, filter influxdb.CheckFilter) (*Check, error) {
```
AlirieGray: comment for what this does? just returns the first Check found according to a filter?

http/check_service.go
```diff
@@ -679,3 +683,209 @@ func (h *CheckHandler) handleDeleteCheck(w http.ResponseWriter, r *http.Request)
...
 }
+
+func checkIDPath(id influxdb.ID) string {
+	return path.Join(prefixChecks, id.String())
+}
+
+// CheckService is a client to the handlers in this package.
+// It does not implement influxdb.CheckService because it returns a concrete representation of the API response
+// and influxdb.Check as returned by that interface is not appropriate for this use case.
+type CheckService struct {
```
AlirieGray: does this new CheckService get passed to the CheckHandler above?

gavincabbage: No- i clarified the language a bit to make it more clear!

PR #16939, opened 2020-02-19
feat: add side nav for cloud users

121watts:
### What
Adding a cloud only feature flagged "NavBox" to the side nav for the cloud specfic `/users` page. 

### Why
No one could find where it was living in the Quartz nav.

### DIFF
--- a/ui/src/pageLayout/containers/Nav.tsx
+++ b/ui/src/pageLayout/containers/Nav.tsx
@@ -252,19 +252,21 @@ class SideNav extends PureComponent<Props, State> {
           />
         </NavMenu.Item>
         <FeatureFlag name="multiUser">
-          <NavMenu.Item
-            titleLink={className => (
-              <a className={className} href={cloudUsersLink}>
-                Team Members
-              </a>
-            )}
-            iconLink={className => (
-              <a href={cloudUsersLink} className={className}>
-                <Icon glyph={IconFont.UsersTrio} />
-              </a>
-            )}
-            active={getNavItemActivation(['users'], location.pathname)}
-          />
+          <CloudOnly>
+            <NavMenu.Item
+              titleLink={className => (
+                <a className={className} href={cloudUsersLink}>
+                  Team Members
+                </a>
+              )}
+              iconLink={className => (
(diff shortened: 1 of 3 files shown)


ui/src/pageLayout/containers/Nav.tsx
```diff
@@ -89,6 +90,9 @@ class SideNav extends PureComponent<Props, State> {
     const feedbackLink =
       'https://docs.google.com/forms/d/e/1FAIpQLSdGJpnIZGotN1VFJPkgZEhrt4t4f6QY1lMgMSRUnMeN3FjCKA/viewform?usp=sf_link'
 
+    // Cloud
+    const cloudUsersLink = `${CLOUD_URL}/users`
```
drdelambre: making `/users` it's own const would make it more uniform with the convention for quartz urls like `/usage` and `/billing` and `/logout`

ui/src/pageLayout/containers/Nav.tsx
```diff
@@ -247,6 +251,23 @@ class SideNav extends PureComponent<Props, State> {
             key="client-libraries"
           />
         </NavMenu.Item>
+        <FeatureFlag name="multiUser">
```
drdelambre: DAMN! look at all that conditional logic

ui/src/shared/constants/index.ts
```diff
@@ -62,6 +62,7 @@ export const CLOUD_CHECKOUT_PATH = '/checkout'
 export const CLOUD_BILLING_PATH = '/billing'
 export const CLOUD_USAGE_PATH = '/usage'
 export const CLOUD_LOGOUT_PATH = '/logout'
+export const CLOUD_USERS_PATH = '/users'
```
drdelambre: oh.. its here, but not there

ui/src/shared/utils/featureFlag.ts
```diff
@@ -11,6 +11,7 @@ export const OSS_FLAGS = {
 
 export const CLOUD_FLAGS = {
   deleteWithPredicate: false,
+  multiUser: false,
```
OfTheDelmer: nice

PR #16903, opened 2020-02-16
fix(modules): cleanup go.mod and remove Gopkg.toml

FiloSottile:
```
$ rm go.mod
$ go1.14rc1 mod init
go: creating new go.mod: module github.com/influxdata/influxdb
go: copying requirements from Gopkg.lock
go: converting Gopkg.lock: stat github.com/googleapis/gax-go@v2.0.5: github.com/googleapis/gax-go@v2.0.5: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
go: converting Gopkg.lock: stat github.com/google/go-github@dd29b543e14c33e6373773f2c5ea008b29aeac95: github.com/google/go-github@v0.0.0-20181009003523-dd29b543e14c: invalid version: go.mod has post-v0 module path "github.com/google/go-github/v18" at revision dd29b543e14c

Manually added the following lines:
    github.com/googleapis/gax-go/v2 v2.0.5
    github.com/google/go-github/v18 v18.2.1-0.20181009003523-dd29b543e14c

Manually changed the language version:
    go 1.12

$ go1.14rc1 mod tidy
go: gopkg.in/russross/blackfriday.v2@v2.0.1/go.mod: verifying module: gopkg.in/russross/blackfriday.v2@v2.0.1/go.mod: reading https://sum.golang.org/lookup/gopkg.in/russross/blackfriday.v2@v2.0.1: 410 Gone
	server response: not found: gopkg.in/russross/blackfriday.v2@v2.0.1: invalid version: go.mod has non-....v2 module path "github.com/russross/blackfriday/v2" at revision v2.0.1

Manually fixed the blackfriday module path:
    github.com/russross/blackfriday/v2 v2.0.1

$ go1.14rc1 mod tidy
$ git rm Gopkg.lock Gopkg.toml
```

Closes #16901

PR #16944, opened 2020-02-20
feat(modules): Use modules

ayang64:
Use go modules for dependency management.

* Add .mod and .sum files generated by go mod init

* Use a directory outside of $GOPATH for testing.

* Update docker files to use go 1.13.8 instead of 1.12 for most builds.

* Simplify go installation by piping from wget to tar.  This avoids
  having to do any cleanup.

* Build and test in a directory outside of GOPATH since we're using
  modules now.

* Update Jenkinsfile to build using modules by building directly inside
  of $WORKSPACE instead of in $GOPATH.

* Use go mod download instead of dep ensure to collect dependencies

* Remove Gopkg related data files.

* Replace calls to "dep ensure" with "go mod download"

* Replace check for being outside of $GOPATH to a check for being INSIDE
  of $GOPATH which is wrong when we're using modules

* Map $DIR to /root/influxdb in docker run command

* Clone source into /root/influxdb which is outside of our $GOPATH.

Closes #

- [x] Rebased/mergeable
- [x] Tests pass


go.mod
```diff
@@ -0,0 +1,109 @@
+module github.com/influxdata/influxdb
+
+go 1.12
+
+require (
```
dgnorton: Did `go mod` pull dependency versions from the old `Gopkg` files? Did it encounter any errors while doing so? Just to make sure we didn't inadvertently change the version of any dependencies.

ayang64: `go mod` copies the dependencies from the `Gopkg.toml` and `Gopkg.lock` files.

the only issues involve warnings about a couple of imports that don't follow the semantic versioning rules:

```
$ go mod init github.com/influxdata/influxdb
go: creating new go.mod: module github.com/influxdata/influxdb
go: copying requirements from Gopkg.lock
go: converting Gopkg.lock: stat github.com/googleapis/gax-go@v2.0.5: github.com/googleapis/gax-go@v2.0.5: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
go: converting Gopkg.lock: stat github.com/google/go-github@dd29b543e14c33e6373773f2c5ea008b29aeac95: github.com/google/go-github@v0.0.0-20181009003523-dd29b543e14c: invalid version: go.mod has post-v0 module path "github.com/google/go-github/v18" at revision dd29b543e14c
```

dgnorton: I notice in https://github.com/influxdata/influxdb/pull/16903 that @FiloSottile manually added entries for those two. Should this PR do the same?

FiloSottile: I added them manually and then ran `go mod tidy` which did away with them. Not sure if it impacted the graph though.

dgnorton: @FiloSottile thanks. Really appreciate the input on that.

Issue #16876, opened 2020-02-13
influx cli cmd tested


jsteenb2:
At thsi time we have no tests around the root cli cmd. This is no bueno. This issue is to track this to make sure we touch that up and provide that test coverage to catch issues with root lvl persisted flags/env vars like token.

would catch issues like the ones found here: #16875

jsteenb2:
note: refactored a bunch of influx cli to use the entire cmd to test. Meaning we test `influx bucket find` instead of `bucket find`, tieing that missing link. 

What's not donezo?
* adding tests to validate the token/host/globals
* remove global dependencies al-togehter from child cmds

PR #16945, opened 2020-02-20
chore(influx): refactor influx cmd to use factory pattern

jsteenb2:
refactors test to use full call paths in addition. testing the full API for the `influx` cli tool. This is the first of several PRs to update this. The next ones will deal with eliminating the drift in other cmds that haven't been worked into shape yet.

work is in support of: #16876

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influx/main.go
+++ b/cmd/influx/main.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/influxdata/influxdb"
 	"github.com/influxdata/influxdb/bolt"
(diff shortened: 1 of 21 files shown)


cmd/influx/main.go
```diff
@@ -248,19 +273,21 @@ func checkSetup(host string) error {
 	return nil
 }
 
-func wrapCheckSetup(fn func(*cobra.Command, []string) error) func(*cobra.Command, []string) error {
-	return wrapErrorFmt(func(cmd *cobra.Command, args []string) error {
-		err := fn(cmd, args)
-		if err == nil {
-			return nil
-		}
+func wrapCheckSetup(f *globalFlags) func(fn func(*cobra.Command, []string) error) func(*cobra.Command, []string) error {
```
kelwang: This wrap pattern is confusing. using `go to` can be much cleaner 

```go
       type cobraFn func(*cobra.Command, []string) error

       func preChecks(f *globalFlags, cobraFn)cobraFn{
                  var err error
                  if err = checkSetup(f.host, f.skipVerify); err != nil && influxdb.EUnauthorized != influxdb.ErrorCode(err) {
				goto fmtErr
		   }
                   if err= fn(cmd, args);err!=nil{
                        goto fmtErr
                  } 
                  return nil
fmtErr:
                  return internal.ErrorFmt(err)
       } 
```

PR #16948, opened 2020-02-20
feat: initial commit for a flux builder library prototype

jsternberg:


jsternberg:
It seems like we're not going to use this for the transpiler so I'm going to close this and keep it as a prototype in a separate repo for now.

Issue #16861, opened 2020-02-13
Create documents HTTP client


gavincabbage:
Create a client for documents similar to this example for buckets https://github.com/influxdata/influxdb/blob/c8edf66ee4605499548e101a29044722a08ddc27/http/bucket_service.go#L615

PR #16955, opened 2020-02-20
feat(http): add document service

affo:
Issue #16861 

Some (maybe) bugs discovered here and there for authorization (see TODOs).

Don't know if that could be  the expected behavior, though it seems it shouldn't be.

### DIFF
--- a/http/document_service.go
+++ b/http/document_service.go
@@ -5,14 +5,30 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"path"
 
 	"github.com/influxdata/httprouter"
 	"github.com/influxdata/influxdb"
 	pcontext "github.com/influxdata/influxdb/context"
+	"github.com/influxdata/influxdb/kit/tracing"
+	"github.com/influxdata/influxdb/pkg/httpc"
 	"go.uber.org/zap"
 )
 
 const prefixDocuments = "/api/v2/documents"
+
+// DocumentService is an interface HTTP-exposed portion of the document service.
+type DocumentService interface {
+	CreateDocument(ctx context.Context, namespace string, orgID influxdb.ID, d *influxdb.Document) error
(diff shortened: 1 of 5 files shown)


http/document_service.go
```diff
@@ -4,7 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/influxdata/influxdb/kit/tracing"
+	"github.com/influxdata/influxdb/pkg/httpc"
```
affo: uops, I'll change tomorrow 😅

http/document_service.go
```diff
@@ -540,7 +556,7 @@ func (h *DocumentHandler) handlePutDocument(w http.ResponseWriter, r *http.Reque
 		return
 	}
 
-	ds, err := s.FindDocuments(ctx, influxdb.WhereID(req.Document.ID), influxdb.IncludeContent)
+	ds, err := s.FindDocuments(ctx, influxdb.WhereID(req.Document.ID), influxdb.IncludeContent, influxdb.IncludeLabels)
```
affo: changees behavior to include labels. Why including content and not labels?

kv/document.go
```diff
@@ -643,6 +643,12 @@ func (s *DocumentStore) DeleteDocuments(ctx context.Context, opts ...influxdb.Do
 			}
 
 			if err := s.service.deleteDocument(ctx, tx, s.namespace, id); err != nil {
+				if IsNotFound(err) {
```
affo: nicer erroring here

testing/util.go
```diff
@@ -63,3 +65,61 @@ func MustIDBase16Ptr(s string) *platform.ID {
 	id := MustIDBase16(s)
 	return &id
 }
+
+func MustCreateOrgs(ctx context.Context, svc *kv.Service, os ...*platform.Organization) {
```
affo: moved this functions to general testing utils

PR #16979, opened 2020-02-24
feat(ui): added route logic and basic login page format for login epic

Asalem1:
This PR is part of a larger epic to integrate login functionality onto influxdb. This first PR is focused on adding a feature flag for the redirect logic that will navigate users to an ~IDPE~ region-based login page that will allow users to login and sign-up to an account via their ~IDPE cluster~ region.

Edits by Bucky after chatting on Slack for clarification


ui/src/onboarding/containers/LoginPage.tsx
```diff
@@ -0,0 +1,50 @@
...
+  Columns,
+  FlexBox,
+  FlexDirection,
+  Grid,
+  JustifyContent,
+  Page,
+  Panel,
+} from '@influxdata/clockface'
+
+// import LoginPage from 'src/onboarding/containers/LoginPage'
```
hoorayimhelping: ✂️

ui/src/onboarding/containers/LoginPage.tsx
```diff
@@ -0,0 +1,50 @@
...
+              </Grid.Column>
+            </Grid.Row>
+          </FlexBox>
+        </Panel>
+      </Page.Contents>
+    </Page>
+  </AppWrapper>
+)
+
+export default SignUp
```
hoorayimhelping: i mentioned this in slack but i think we should stop exporting default: https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/

even though it's less consistent at the moment, i think it's probably a better pattern to start following

ui/src/onboarding/containers/LoginPage.tsx
```diff
@@ -0,0 +1,50 @@
...
+          >
+            <Grid.Row className="sign-up--full-height">
+              <Grid.Column
+                widthXS={Columns.Twelve}
+                widthMD={Columns.Five}
+                offsetMD={Columns.Four}
+                widthLG={Columns.Four}
+                className="sign-up--full-height"
+              >
+                Login Page will go here
```
hoorayimhelping: is this text being in here a problem? i'm assuming no cause of flagging but it's good to ask

ui/src/onboarding/containers/LoginPage.tsx
```diff
@@ -0,0 +1,50 @@
...
+  AlignItems,
+  AppWrapper,
+  Columns,
+  FlexBox,
+  FlexDirection,
+  Grid,
+  JustifyContent,
+  Page,
+  Panel,
+} from '@influxdata/clockface'
```
hoorayimhelping: 👌 alphabetization

ui/src/onboarding/containers/LoginPage.scss
```diff
@@ -0,0 +1,168 @@
...
+  height: 100%;
+  justify-content: space-between;
+}
+
+.sign-up--panel {
+  height: 100%;
+  max-width: 1600px;
+  min-height: 670px + $ix-marg-d;
+  padding: $ix-marg-d;
+  margin: auto auto;
```
hoorayimhelping: `margin: auto` accomplishes this as well and seems to be more idiomatic

ui/src/onboarding/containers/LoginPage.scss
```diff
@@ -0,0 +1,168 @@
...
+    text-align: center;
+  }
+
+  .sign-up--panel:not(.aws) {
+    height: 80vh;
+    padding: 64px;
+  }
+}
+
+@media screen and (min-width: $grid--breakpoint-md) {
```
hoorayimhelping: i think you can cut this and fold these two class selectors up into the above media declaration, since it does the same thing - `@media screen` and `min-width: $grid--breakpoint-md`

ui/src/onboarding/containers/LoginPage.scss
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  .google-yellow {
+    fill: #f9bb00;
+  }
+
+  .google-red {
+    fill: #e74133;
+  }
+}
```
hoorayimhelping: minor nit, but can you alphabetize the properties in the selectors, like we do with imports?

ui/src/shared/utils/featureFlag.ts
```diff
@@ -7,6 +7,7 @@ export const OSS_FLAGS = {
   telegrafEditor: false,
   customCheckQuery: false,
   matchingNotificationRules: false,
+  IDPELoginPage: false,
```
hoorayimhelping: to keep the language consistent between engineering, product, customer success design, sales, etc, i think we should rename this flag to something akin to `regionBasedLoginPage` since IDPE is a more engineering level construct, while a region is a company wide concept.

ui/src/onboarding/containers/LoginPage.tsx
Asalem1: Not an issue in compiling, but it's also just a temporary placeholder until the next PR

PR #16987, opened 2020-02-24
feat(ui): added auth0 to branch and added form functionality

Asalem1:
This is part two of a larger epic that is integrating auth0 sign-up and log-in capabilities for cloud users on the region they are trying to access. This PR adds a lot of the meat and potatoes for this functionality. Auth0 is added as a dependency to allow users to directly sign-up & log-in through the newly created `login` route & LoginPage. This PR allows users to toggle between the sign-up form and the log-in form, and request to change their password. In addition to the form logic that was built out, this PR also adds a new shared component called `SocialButton` that allows developers to pass button-specific properties (button text, click functionality, and children) to reduce repetition when adding new social sign-up capabilities. Finally, two SVGs were added (github and google) to allow for customized buttons.

**HOW TO TEST**

In order to test this functionality, you will need to run the UI as a cloud-dev user, and enable the feature flag. You will need to replace the current auth0js config variables with valid config variables that can be obtained from Engineering Vault quartz/.env file (auth0 clientID, domain). The redirectURL should be whichever home route your local development is happening on (i.e. localhost:8080/, etc...).


ui/src/onboarding/containers/LoginPageContents.tsx
```diff
@@ -0,0 +1,385 @@
...
+
+// Components
+import {LoginForm} from 'src/onboarding/components/LoginForm'
+import {SignUpForm} from 'src/onboarding/components/SignUpForm'
+import {SocialButton} from 'src/shared/components/SocialButton'
+import {GoogleLogo, GithubLogo} from 'src/clientLibraries/graphics'
+// Types
+import {Auth0Connection} from 'src/types'
+
+const auth0 = new auth0js.WebAuth({
```
Asalem1: These values are simply placeholder values for a later iteration where config variables will be set by data obtained on IDPE

ui/src/onboarding/containers/LoginPageContents.tsx
```diff
@@ -0,0 +1,383 @@
...
+            If you haven't received an email, please ensure that the email you provided is correct.`,
+          notificationVisible: true,
+        })
+        // sets a time to remove the notification
+        setTimeout(() => {
+          this.setState({
+            notificationText: '',
+            notificationVisible: false,
+          })
+        }, 4000)
```
hoorayimhelping: why not use `shared/actions/notifications`, which handles all this for you?

ui/src/onboarding/containers/LoginPageContents.tsx
```diff
@@ -0,0 +1,383 @@
...
+    if (!email) {
+      this.setState({emailError: 'Please enter a valid email address'})
+      return
+    }
+    auth0.changePassword(
+      {
+        email,
+        connection: Auth0Connection.Authentication,
+      },
+      (err, successMessage) => {
```
hoorayimhelping: this callback style is kind of gnarly, and gets kind of deep in some code above. did you look into wrapping this library in a shallow wrapper to make use of promises, which is pretty much how we do async code throughout the codebase?

ui/src/onboarding/containers/LoginPageContents.tsx
```diff
@@ -0,0 +1,383 @@
...
+        )
+      }
+    )
+  }
+
+  private displayErrorMessage = (errors, auth0Err) => {
+    if (auth0Err.code.includes('error in email')) {
+      this.setState({
+        ...errors,
+        emailError: 'Please enter a valid email address',
```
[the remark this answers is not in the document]
Asalem1: Since the auth0Err is provided to us and we aren't performing any actionable email parsing (purposefully), it doesn't seem like we need to elaborate on an email address's validity

ui/src/onboarding/containers/LoginPageContents.tsx
```diff
@@ -0,0 +1,383 @@
...
+      })
+    } else if (auth0Err.code === 'user_exists') {
+      const emailError = `An account with that email address already exists. Try logging in instead.`
+      this.setState({...errors, emailError})
```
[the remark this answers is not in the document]
Asalem1: I don't think we should be, since it could be that the user is trying to sign-up with an email address that already has an associated account. The assumption is that the user is trying to sign-up, not log-in

ui/src/onboarding/containers/LoginPageContents.tsx
Asalem1: copy pasta'd too much 🤦‍♂

ui/src/onboarding/containers/LoginPageContents.tsx
```diff
@@ -0,0 +1,383 @@
...
+
+    auth0.signup(
+      {
+        connection: Auth0Connection.Authentication,
+        email,
+        password,
+        family_name,
+        given_name,
+      },
+      err => {
```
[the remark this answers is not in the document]
Asalem1: Won't renaming the `err` parameter to `error` cause issues with the follow-up declaration here?

https://github.com/influxdata/influxdb/pull/16987/files?file-filters%5B%5D=.json&file-filters%5B%5D=.lock&file-filters%5B%5D=.tsx#diff-c8135c48ec00d98c14033c5ab920ee22R307

ui/src/onboarding/containers/LoginPageContents.tsx
Asalem1: Totally agree, this is based on the docs, but I can see if it's possible to refactor these into async awaits

Asalem1: This is more a result of the design of the API where the second parameter that's accepted is the callback. I tried refactoring this to promisify the result but didn't have the same outcome

PR #16992, opened 2020-02-24
fix: query builder groups on columns, not tag values

### LINKED ISSUE #16783: Group By in the query builder does not show column names
As a user
when i select group by in the visual data explorer
then i would like to see a list of relevant column names that I can select from

The current behavior seems to show the list of measurement values, which doesn't make sense in the context of the group by function. 

Looking at the signature for the group by function, it accepts a list of column names: https://v2.docs.influxdata.com/v2.0/reference/flux/stdlib/built-in/transformations/group/

The UI should show a list of relevant column names based on the query that has already been built.

hoorayimhelping:
Closes #16783 

A picture is worth a thousand words, so here:

**WRONG:**
Screen Shot 2020-02-24 at 2 43 26 PM

**Right:**
Screen Shot 2020-02-24 at 2 42 52 PM

* **Wrong**: The grouping options map to tag values
* **Right:** The grouping options map to columns

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/timeMachine/components/TagSelector.tsx
+++ b/ui/src/timeMachine/components/TagSelector.tsx
@@ -35,7 +35,6 @@ import DefaultDebouncer from 'src/shared/utils/debouncer'
 import {toComponentStatus} from 'src/shared/utils/toComponentStatus'
 import {
   getActiveQuery,
-  getActiveTagValues,
   getActiveTimeMachine,
   getIsInCheckOverlay,
 } from 'src/timeMachine/selectors'
@@ -48,6 +47,9 @@ import {
 } from 'src/types'
 
 const SEARCH_DEBOUNCE_MS = 500
+
+// We don't show these columns in results but they're able to be grouped on
+const ADDITIONAL_GROUP_BY_COLUMNS = ['_start', '_stop', '_time']
 
 interface StateProps {
   aggregateFunctionType: BuilderAggregateFunctionType
@@ -143,6 +145,10 @@ class TagSelector extends PureComponent<Props> {
       )
     }
 
+    const placeholderText =
+      aggregateFunctionType === 'group'
+        ? 'Search group column values'
+        : `Search ${selectedKey} tag values`
     return (
       <>
         <BuilderCard.Menu testID={`tag-selector--container ${index}`}>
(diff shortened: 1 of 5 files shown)


ui/src/timeMachine/selectors/index.ts
```diff
@@ -244,25 +242,6 @@ export const getActiveTimeRange = (
...
-      tag.values.forEach(value => {
-        values.push(value)
-      })
-    })
-    return values
-  }
-
-  return activeQueryBuilderTags[index].values
-}
-
```
hoorayimhelping: this was testing the wrong thing.

ui/src/timeMachine/selectors/index.test.ts
```diff
@@ -62,88 +58,3 @@ describe('TimeMachine.Selectors.Index', () => {
...
-
-  it('returns all previous tag values when the function is group', () => {
-    const actualTags = getActiveTagValues(activeQueryTags, 'group', 2)
-    expect(actualTags).toEqual([
-      ...activeQueryTags[0].values,
-      ...activeQueryTags[1].values,
-      ...activeQueryTags[2].values,
-    ])
-  })
-})
```
hoorayimhelping: this test was testing the wrong thing

ui/src/timeMachine/components/TagSelector.tsx
```diff
@@ -49,6 +48,9 @@ import {
 
 const SEARCH_DEBOUNCE_MS = 500
 
+// We don't show these columns in results but they're able to be grouped on
```
121watts: This is not always true.  There are scenarios where a `_start` and `_stop` are not not columns in a query response.

ui/src/timeMachine/selectors/index.test.ts
121watts: can we test the right thing?

ui/src/timeMachine/components/TagSelector.tsx
hoorayimhelping: Russ mentioned that, but it seems like in the context of query builder, it's safe to assume we can group on those columns? I'll change the comment

ui/src/timeMachine/selectors/index.test.ts
hoorayimhelping: >can we test the right thing?
@121watts only cause you asked https://github.com/influxdata/influxdb/pull/16992/files#diff-c432c510f4c140978dbe371ff3e9142dR93

PR #17028, opened 2020-02-27
fix(ui): updated default positioning for adding aggregate funcs when toggling from query builder

### LINKED ISSUE #16668: Data Explorer: Script Editor should default position to END of script
__Steps to reproduce:__

1. Open Data Explorer 
2. Begin a query using the query builder
3. Toggle to 'Script Editor' view [text editor sets focus on the top line]
4. Click on a function to add to the query
5. Function gets added as the TOP line of the query [after the from() statement]

__Expected behavior:__
Expect that the focus is set so that any additional functions would be added to the END of the script by default (unless I have manually selected a different insertion point in the text editor)

__Actual behavior:__
New items are inserted as first 'argument' (i'm probably using the wrong term)

__Environment info:__

Cloud, Acceptance and Production Version 2.0.0 (8ff62ab)

__Config:__

__Logs:__

__Performance:__

```sh
iostat -xd 1 30 > iostat.txt
```

Asalem1:
Closes #16668

### Problem

Currently toggling from the query builder to the script editor will default aggregate functions that are selected to the initial part of the query.

### Solution

Create a conditional check to update the range that we are adding to to be the end of the existing query. Also added logic to add a new line whenever an aggregate function is selected.

### DIFF
--- a/ui/src/timeMachine/utils/insertFunction.ts
+++ b/ui/src/timeMachine/utils/insertFunction.ts
@@ -20,7 +20,7 @@ export const formatFunctionForInsert = (
     return `\n${fluxFunction}`
   }
 
-  return `  |> ${fluxFunction}`
+  return `\n  |> ${fluxFunction}`
 }
 
 export const generateImport = (
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,11 @@
 ### Features
 
 ### Bug Fixes
+
 1. [17039](https://github.com/influxdata/influxdb/pull/17039): Fixed issue where tasks are exported for notification rules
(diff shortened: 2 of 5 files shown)


ui/src/timeMachine/utils/insertFunction.ts
```diff
@@ -20,7 +20,7 @@ export const formatFunctionForInsert = (
     return `\n${fluxFunction}`
   }
 
-  return `  |> ${fluxFunction}`
+  return `\n  |> ${fluxFunction}`
```
Asalem1: ensures that all selected aggregate functions are inserted onto a new line

PR #17031, opened 2020-02-27
refactor(repl): remove the querier from the repl

jsternberg:
The repl no longer takes in a querier and it will run everything
locally. The spec interface will now not be used and will be removed
from the http endpoint at some point.

### DIFF
--- a/cmd/influx/query.go
+++ b/cmd/influx/query.go
@@ -42,19 +42,20 @@ func fluxQueryF(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to load query: %v", err)
 	}
 
-	orgSvc, err := newOrganizationService()
-	if err != nil {
-		return fmt.Errorf("failed to initialized organization service client: %v", err)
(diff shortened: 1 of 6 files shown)


cmd/influx/query.go
```diff
@@ -42,19 +42,20 @@ func fluxQueryF(cmd *cobra.Command, args []string) error {
...
-	orgSvc, err := newOrganizationService()
-	if err != nil {
-		return fmt.Errorf("failed to initialized organization service client: %v", err)
-	}
-
-	orgID, err := queryFlags.org.getID(orgSvc)
-	if err != nil {
-		return err
-	}
+	// TODO(jsternberg): Restore this when the influxdb client source is merged.
```
wolffcm: What is "influxdb client source"?

jsternberg: That's the `influxdb.from` that's incoming to the flux repository. The influx cli will also rely on it but it will add defaults from the cli.

PR #17036, opened 2020-02-27
feat: add support for pkg-config

### LINKED ISSUE #16980: Integrate pkg-config build method for building libflux for flux dependency
This is an extension of #2553. Once `pkg-config` has been integrated with the flux repository, the influxdb builds are going to need to include it with their builds. This should also involve having it so invoking `./env` will configure `pkg-config` to be used when building without installing `pkg-config` globally by setting the `PKG_CONFIG` environment variable. This can be done by creating a script that will invoke `pkg-config` using `go run` and setting the `PKG_CONFIG` environment variable to that script location. This will allow influxdb developers to install it globally to make it easier to work, but will allow those who don't want to install it globally to still be able to run the build and test commands.

jsternberg:
This adds support for using pkg-config to build libflux inside of the
flux dependency. The build can occur by either installing `pkg-config`
into your path or the `env` script can be used to invoke it from the go
modules.

Fixes #16980.


env
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```
wolffcm: I guess that `BASH_SOURCE[0]` could be a relative path, so you need to use `cd` and `pwd` to get the absolute path?

jsternberg: Yea I think that's it. Maybe I should leave a note to the stackoverflow place I copied it from.

Issue #17038, opened 2020-02-27
Not all of my tasks are exported by pkg


nisley:
I have two manually created tasks in my Cloud account, and they are not exported with this command:

bin/darwin/influx pkg --host "https://us-west-2-1.aws.cloud2.influxdata.com/" --org-id "039346c3777a1000" --token "<myNewAllAccessToken>" export all > ../templateExport/cloudExport.yml

[cloudExport.txt](https://github.com/influxdata/influxdb/files/4264371/cloudExport.txt)

PR #17042, opened 2020-02-28
fix(pkger): correct the export of tasks to include system type tasks

jsteenb2:
wrong assumption in original design. The system tasks are actually
the CRUD tasks. Name is a bit confusing for user generated tasks.

closes: #17038

- [x] Rebased/mergeable
- [x] Tests pass


cmd/influxd/launcher/pkger_test.go
```diff
@@ -408,52 +452,59 @@ spec:
 			assert.Equal(t, expected, secrets[0])
 		})
 
+		t.Run("exporting all resources for an org", func(t *testing.T) {
```
jsteenb2: massive gaping hole in test suite, totally missed this

PR #17044, opened 2020-02-28
feat(ui): added transition effects between form toggling

Asalem1:
This PR is part of a larger PR to add the sign-in functionality to IDPE. This particular PR is intended to add transitioning effects when users toggle between the sign-up and log-in form so that the transition is a bit smoother. Added react-spring as a dependency to keep with existing transition libraries that are used

login_transition

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -182,6 +182,7 @@
     "react-router": "^3.0.2",
     "react-router-redux": "^4.0.8",
     "react-scrollbars-custom": "^4.0.0-alpha.8",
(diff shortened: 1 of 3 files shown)


TCL735:
I'm curious to see what this looks like. Is there a way to include a gif of the toggle happening?

TCL735:
Approved. Missed the click, but this is approved.

Asalem1:
Ya, I'm a fool and saw approved on a different PR that I forgot was up - saw this was ready to merge and merged it in

PR #17045, opened 2020-02-28
feat(ui): added logout functionality to epic

Asalem1:
This is PR part of a larger PR that is intended to add sign-in functionality to IDPE. This particular PR deals specifically with logging out and allowing users to logout using the existing logout button

### DIFF
--- a/ui/src/Logout.tsx
+++ b/ui/src/Logout.tsx
@@ -1,10 +1,11 @@
 // Libraries
-import {PureComponent} from 'react'
+import {FC, useEffect} from 'react'
 import {withRouter, WithRouterProps} from 'react-router'
 import auth0js from 'auth0-js'
 
 // APIs
 import {postSignout} from 'src/client'
+import {getAuth0Config} from 'src/authorizations/apis'
 
 // Constants
 import {CLOUD, CLOUD_URL, CLOUD_LOGOUT_PATH} from 'src/shared/constants'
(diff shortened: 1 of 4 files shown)


ui/src/Logout.tsx
```diff
@@ -34,9 +37,14 @@ export class Logout extends PureComponent<Props> {
...
       }
 
-      this.props.router.push(`/signin`)
+      router.push(`/signin`)
     }
   }
+
+  useEffect(() => {
+    handleSignOut()
+  }, [])
```
TCL735: Nice

ui/src/Logout.tsx
```diff
@@ -1,29 +1,32 @@
...
-
-  public render() {
-    return null
-  }
-
-  private handleSignOut = async () => {
+// Utils
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+
+const Logout: FC<WithRouterProps> = ({router}) => {
```
drdelambre: you need some more types in here. try running this in `influxdb/ui`:
`yarn tsc -noImplicitAny | grep -A 3 -i "$(git diff --name-status master | awk '{ print substr($2,4) }')`

ui/src/authorizations/apis/index.ts
```diff
@@ -17,3 +17,16 @@ export const createAuthorization = async (
     throw error
   }
 }
+
+export const getAuth0Config = async (): Promise<Auth0Config> => {
+  try {
+    const {data} = await AJAX({
```
drdelambre: why are you doubling down on using the old request library for new features?

https://giphy.com/embed/AuIvUrZpzBl04

ui/src/onboarding/containers/LoginPageContents.tsx
```diff
@@ -76,6 +71,16 @@ class LoginPageContents extends PureComponent<DispatchProps> {
     confirmPasswordError: '',
   }
 
+  public async componentDidMount() {
+    const config = await getAuth0Config()
+    this.auth0 = auth0js.WebAuth({
```
drdelambre: dope

ui/src/authorizations/apis/index.ts
Asalem1: That's a great question. I did a global search of influxdb for `/api/v2private` and followed that convention since we aren't actually generating a route for this particular endpoint on IDPE

drdelambre: use fetch then

PR #17047, opened 2020-02-28
feat(stdlib): add influxdb source

jsternberg:
This updates the repl to support the new influxdb source and use it by
default in the repl. It will automatically set some default variables
for the influxdb source to make it easier to use the cli. In particular,
it will set the default organization, token, and the host. The
organization gets set to the one specified in the repl command and the
token gets filled in with the user installed one. The host defaults to
localhost but will change to whichever one was specified on the cli.

In addition, this will replace the http client with one that sets
insecure skip verify if the `--skip-verify` flag is used.


query/stdlib/influxdata/influxdb/from.go
```diff
@@ -20,10 +17,10 @@ type FromOpSpec struct {
 }
 
 func init() {
-	fromSignature := semantic.MustLookupBuiltinType("influxdata/influxdb", "from")
-	runtime.ReplacePackageValue("influxdata/influxdb", influxdb.FromKind, flux.MustValue(flux.FunctionValue(FromKind, createFromOpSpec, fromSignature)))
-	flux.RegisterOpSpec(FromKind, newFromOp)
-	plan.RegisterProcedureSpec(FromKind, newFromProcedure, FromKind)
+	// fromSignature := semantic.MustLookupBuiltinType("influxdata/influxdb", "from")
```
wolffcm: I guess that the signature is the same in `flux` and `influxdb` now, so can you just delete this `init` method?

jsternberg: Yes we can get rid of the replace stuff.

jsternberg:
I need to refactor some of the procedure stuff so I'm working on that at the moment. I'll ask for a rereview in a bit.

query/stdlib/experimental/to_test.go
```diff
@@ -85,53 +83,6 @@ from(bucket:"mydb")
 	}
 }
 
-func TestToOpSpec_BucketsAccessed(t *testing.T) {
```
wolffcm: I guess the `BucketsAccessed` stuff is no longer needed?

query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -20,6 +22,27 @@ func init() {
...
+func (rule FromStorageRule) Name() string {
+	return "influxdata/influxdb.FromStorageRule"
+}
+
+func (rule FromStorageRule) Pattern() plan.Pattern {
+	return plan.Pat(influxdb.FromKind)
+}
+
+func (rule FromStorageRule) Rewrite(node plan.Node) (plan.Node, bool, error) {
+	fromSpec := node.ProcedureSpec().(*influxdb.FromProcedureSpec)
```
wolffcm: Just to be clear, this takes the `from` in the `flux` repo, which is the remote-host `from`, and converts it to an `influxdb` from, which will talk to storage directly.

query/stdlib/influxdata/influxdb/from.go
```diff
@@ -6,107 +6,18 @@ import (
...
-		readBuckets = append(readBuckets, bf)
-	}
-	return readBuckets, writeBuckets
-}
+type (
+	NameOrID   = influxdb.NameOrID
+	FromOpSpec = influxdb.FromOpSpec
+)
 
 type FromProcedureSpec struct {
```
wolffcm: Maybe this should be renamed `FromStorageProcedureSpec` to differentiate it from the the one in the `flux` repo?  I'm finding the differences are getting confusing.

query/stdlib/experimental/to_test.go
jsternberg: I couldn't find it being used anywhere.

query/stdlib/influxdata/influxdb/rules.go
jsternberg: This is correct.

query/stdlib/influxdata/influxdb/from.go
jsternberg: I'll do that rename. I think I'm also going to update the rule to convert to the storage procedure spec if host is unset at all and add a validation to ensure that org isn't set to the storage procedure spec. I think the error messages will get confusing otherwise.

Issue #16901, opened 2020-02-16
v1.7.10 requires a module by the wrong version


FiloSottile:
`github.com/influxdata/influxdb@v1.7.10` requires `github.com/google/go-github@v0.0.0-20181009003523-dd29b543e14c`. Commit google/go-github@dd29b543e14c shows [a `go.mod` with path `github.com/google/go-github/v18`](https://github.com/google/go-github/blob/dd29b543e14c/go.mod) so it can't be used with v0.0.0 and naked `github.com/google/go-github` import path.

Go 1.13 notices and breaks.

```
go: github.com/influxdata/influxdb@v1.7.10 requires
	github.com/google/go-github@v0.0.0-20181009003523-dd29b543e14c: invalid version: go.mod has post-v0 module path "github.com/google/go-github/v18" at revision dd29b543e14c
```

In other words, `github.com/google/go-github@v0.0.0-20181009003523-dd29b543e14c` doesn't exist, that commit is `github.com/google/go-github/v18@v18.2.1-0.20181009003523-dd29b543e14c`.

mark-rushakoff:
Thanks for the report @FiloSottile.

@e-dard introduced go.mod in into the 1.7 branch in cbae66e70765bacc15f22dad7474b18c86a84b53, via https://github.com/influxdata/influxdb/pull/15863 -- there was no mention of go.mod in the commit message or in the PR comments by reviewers @jacobmarble nor @benbjohnson.

I noticed we also still have Gopkg.lock. My suspicion is that we weren't intending to migrate to modules for influxdb 1.7 yet. But now that we've published 1.7.10 with modules, it seems like we ought to fix up go.mod in 1.7.11.

I did try deleting `github.com/google/go-github v0.0.0-20181009003523-dd29b543e14c` line from go.mod and then running `go get github.com/google/go-github/v18@dd29b543e14c`, but that ran into the next error: `go: gopkg.in/russross/blackfriday.v2@v2.0.1: go.mod has non-....v2 module path "github.com/russross/blackfriday/v2" at revision v2.0.1`.

Then, deleting the `gopkg.in/russross/blackfriday.v2 v2.0.1` line from go.mod allows `go get github.com/google/go-github/v18@dd29b543e14c` to run successfully, but it also updates many of the referenced modules' versions.

@dgnorton let's see if we can get modules set up properly for 1.7.11.

FiloSottile:
[...]

PR #17063, opened 2020-03-02
feat(influxdb): Backport modules to 1.7.x

ayang64:
This PR will backport the modules work done in 1.8 to the 1.7 branch.

This involves removing Gopkg.* files, adding go.mod and go.sum files, and updating the supporting CI/CD configuration and scripts.

Closes: #16901
Closes: #17057

- [x] Rebased/mergeable
- [ ] Tests pass


sylr:
I don't believe this should be backported to 1.7 line.

According to semver this kind of change falls into a MINOR change, not a PATCH one.

ayang64:
> I don't believe this should be backported to 1.7 line.
> 
> According to semver this kind of change falls into a MINOR change, not a PATCH one.

so we have a 1.7 branch that the next minor release will likely be based on.  1.7 is the tip of all 1.7.x development so if we release, for example, 1.7.11, we'd like this modules work to land there.

does this make sense or am i missing something?

go.mod
```diff
@@ -3,108 +3,54 @@ module github.com/influxdata/influxdb
 go 1.12
```
dgnorton, line 3: Does this need to be updated?

ayang64: d'oh -- yes, it does.

Dockerfile_build_ubuntu64_go1.12
```diff
@@ -0,0 +1,41 @@
...
+    python \
+    python-boto \
+    asciidoc \
+    xmlto \
+    docbook-xsl
+
+RUN gem install fpm
+
+# setup environment
+ENV GO_VERSION  1.12
```
dgnorton, line 26: Does this need to be updated?

dgnorton: Actually, the name of this file is `Dockerfile_build_ubuntu64_go1.12 ` so I'm not sure if it should be changed here or a new file created??

ayang64:
> @ayang64 Looks like this backport introduces `Dockerfile_build_ubuntu64_go1.12 `. Do we need it in 1.7.x?

I don't think so.  The original author isn't around but as far as I can tell, it was there to test go 1.12 before it was released.

I'll delete the file and resubmit the PR.

ayang64:
Actually -- if it is okay with you, what I'd like to do is leave it in this branch.

The docker file exists in our `master-1.x` branch too.  I'd like to open a ticket to remove it there and backport the fix to 1.7.x.

Does that sound reasonable?

dgnorton:
I'd say just remove it from this branch so we don't have to bother with a backport to remove it.

PR #17072, opened 2020-03-03
fix(ui): fixed regression bug that preventing variable maps from properly assigning values

### LINKED ISSUE #17026: Variable maps aren't working
when you create a variable of type 'map', and you try to use it in a query (add `v.varname`, and then select which key to substitute in the dropdown on the right), it is substituting the key of the map instead of the value

Asalem1:
Closes #17026

### Problem

Variable maps that had previously been created were not mapping variables values out properly. 

### Solution

Undid a change introduced in the following PR which improperly mapped keys of the maps as the value rather than the values:

https://github.com/influxdata/influxdb/commit/78244ce6dec0ed7f5f7be70394d939f3a3d2088a#diff-75e6c7f86f1381fe2013a292310d9d0eL164


desa:
@Asalem1 can you include a test for this?

Asalem1:
@desa definitely. This actually has a few tests that are failing because it messed around with the way that variable maps are intended to be displayed. I'm currently working on trying to update the solution so that the display aspect of this is maintained, but that it is functionally correct.

PR #17093, opened 2020-03-04
fix(http): fix auth in document handler

affo:


authorizer/document.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+//// Private functions to build the options above.
+
+func newDocumentPermission(a influxdb.Action, orgID, id influxdb.ID) (*influxdb.Permission, error) {
+	return influxdb.NewPermissionAtID(id, a, influxdb.DocumentsResourceType, orgID)
+}
+
+func newDocumentOrgPermission(a influxdb.Action, orgID influxdb.ID) (*influxdb.Permission, error) {
+	return influxdb.NewPermission(a, influxdb.DocumentsResourceType, orgID)
```
lyondhill:
```suggestion
	return newOrgPermission(a, orgID)
```

affo:
@lyondhill 

What I did:
 - removed the possibility to generate side effects with the `DocumentIndex`.
 - make it so that that behavior is now in `kv`.
 - separated authorization layer in `authorizer` package.
 - changed the `authorizer.LabelService` so that it accounts for the special case of documents permissions.
 - fixed tests

PR #17096, opened 2020-03-04
chore: Refactor LSP server integration for future work

bthesorceror:

### DIFF
--- a/ui/src/external/monaco.flux.server.ts
+++ b/ui/src/external/monaco.flux.server.ts
@@ -1,44 +1,118 @@
-import {sendMessage, initialize} from 'src/external/monaco.flux.messages'
+import Deferred from '../utils/Deferred'
+import {
+  LSPResponse,
+  parseResponse,
+  initialize,
+  LSPMessage,
+  completion,
+  didOpen,
+  didChange,
+} from 'src/external/monaco.flux.messages'
+import {registerCompletion} from './monaco.flux.completions'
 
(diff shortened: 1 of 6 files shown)


ui/src/external/monaco.flux.server.ts
```diff
@@ -1,44 +1,118 @@
...
-      resolve(_module as LSPServer)
-      return
-    }
+const queue = []
+let server = null,
+  loading = false
+
+export class LSPServer {
+  private server: WASMServer
+  private messageID: number = 0
```
drdelambre: ahhhh.. this is scoped to the server and not the text

ui/src/external/monaco.flux.server.ts
```diff
@@ -1,44 +1,118 @@
...
+  }
+
+  async completionItems(
+    uri: string,
+    position: Position,
+    context: CompletionContext
+  ): Promise<CompletionItem[]> {
+    try {
+      const response = (await this.send(
+        completion(this.currentMessageID, uri, position, context)
```
drdelambre: if we're going all OOPy here.. why not define the messages in their methods on the instance? i dont understand the benefit of keeping the functional interface.

bthesorceror: Readability, also those function construct a type of message but the server has state to pass to them. We don't have to be 100 percent OOP or functional.

Issue #16847, opened 2020-02-12
Refactor "notification" dependency on Flux parser


wolffcm:
The package `influxdb/notification` and its subpackages `check` and `rule` seem to need to parse Flux queries.

This is fine, but there are several very common packages that require `notification` packages:
- `influxdb/kv`
- `influxdb/http`

We should refactor this so these commonly-used packages to not indirectly need the flux parser.

Relevant dependency graph:
![graph](https://user-images.githubusercontent.com/7902964/74387300-04986a80-4dad-11ea-9749-0586e1d9fd64.png)

PR #17104, opened 2020-03-05
refactor(task): tasks will now use the flux language service

### LINKED ISSUE #16844: Refactor tasks dependency on the Flux parser
The root package for `influxdb` makes use of the Flux parser, which will mean that any client code including this package will have a dependency on Rust:

https://github.com/influxdata/influxdb/blob/a5f508de771c6383729baa9f634b5518fedf2ed7/task.go#L297

We should remove this dependency.

The root `influxdb` package also depends on `influxdb/task/options` which also needs the Flux parser.  The root package `influxdb` should not depend on `influxdb/task/options` either.

In addition, the very commonly-used packages `influxdb/http`, `influxdb/kv` and `influxdb/authorizer` require the `influxdb/task/backend` package, which ultimately also requires the Flux parser.  These three commonly used packages should not import `influxdb/task/backend`.  (Perhaps the types declared in `backend` can be moved up into the `influxdb` root package)

jsternberg:
The tasks subsystem will now use the flux language service to parse and
evaluate flux instead of directly interacting with the parser or
runtime. This helps break the dependency on the libflux parser for the
base influxdb package.

This includes the task notification packages which were changed at the
same time.

Fixes #16844 and #16847.

### DIFF
--- a/task.go
+++ b/task.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/influxdata/flux/ast"
 	"github.com/influxdata/flux/ast/edit"
-	"github.com/influxdata/flux/parser"
 	"github.com/influxdata/influxdb/task/options"
 )
 
@@ -284,7 +283,13 @@ func (t TaskUpdate) Validate() error {
 
 // safeParseSource calls the Flux parser.ParseSource function
 // and is guaranteed not to panic.
-func safeParseSource(f string) (pkg *ast.Package, err error) {
+func safeParseSource(parser FluxLanguageService, f string) (pkg *ast.Package, err error) {
+	if parser == nil {
+		return nil, &Error{
+			Code: EInternal,
+			Msg:  "flux parser is not configured; updating a task requires the flux parser to be set",
+		}
+	}
 	defer func() {
 		if r := recover(); r != nil {
 			err = &Error{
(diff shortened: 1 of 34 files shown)


task.go
```diff
@@ -294,25 +293,27 @@ func safeParseSource(f string) (pkg *ast.Package, err error) {
...
 	}
 	toDelete := map[string]struct{}{}
-	parsedPKG, err := safeParseSource(oldFlux)
+	if parser == nil {
+		return &Error{
+			Code: EInternal,
+			Msg:  "flux parser is not configured; updating a task requires the flux parser to be set",
+		}
+	}
+	parsedPKG, err := safeParseSource(parser, oldFlux)
```
fchikwekwe: Is this the only place where we have to check that the parser is properly configured?

task/options/strconv.go
```diff
@@ -0,0 +1,109 @@
...
+)
+
+// TODO(jsternberg): This file copies over code from an internal package
+// because we need them from an internal package and the only way they
+// are exposed is through a package that depends on the core flux parser.
+// We want to avoid a dependency on the core parser so we copy these
+// implementations.
+//
+// In the future, we should consider exposing these functions from flux
+// in a non-internal package outside of the parser package.
```
fchikwekwe: Yeah, I second this. That was my first thought when I saw this file and the other duplicated `ParseDuration` function above. Do we have an issue for this?

jsternberg: No I'll make one.

jsternberg: https://github.com/influxdata/flux/issues/2588

task.go
jsternberg: No, I think we probably should check it in other locations. It just panics in those places if it isn't configured. I will add that before merging.

PR #17109, opened 2020-03-05
refactor(storage/reads): remove the storage dependency on libflux

jsternberg:
This removes the storage dependency on libflux by moving the interfaces
it implements to the `query` package so it can reference the definitions
rather than the package with the implementation and the registration
with the runtime. This breaks the dependency where a storage package
depends on a flux runtime package.


query/storage.go
```diff
@@ -0,0 +1,77 @@
+package query
```
wolffcm: Go package names drive me crazy sometimes.  This is a subpackage of query, that will be referred to as `storage` inside the top-level `storage` package.

jsternberg: It will be referenced as `query` in the storage package. Would you prefer it was referred to as `storage.Reader` instead?

jsternberg: I renamed the type to `StorageReader` so it was clear that it was the query subsystem reading from storage rather than a generic "reader" which could have been misinterpreted as a query reader.

wolffcm: Ah, right.  I was just confused.  I think the change is an improvement.

PR #17110, opened 2020-03-06
fix(ui): disabled group functionality for check query builder

### LINKED ISSUE #16956: Disable group by in query builder in check query builder
Without it the check that is generate is malformed and will cause an error

Asalem1:
Closes #16956

### Problem

Group functionality shouldn't exist on the query builder when setting up checks for alerts

### Solution

Created custom logic to only display `filter` as an option and disable the dropdown to reduce functionality confusion

disable_group

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## v2.0.0-beta.6 [unreleased]
 
 ### Features
+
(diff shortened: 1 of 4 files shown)


desa:
(requested review from monitoring team so I could see if the assigner is working)

PR #17113, opened 2020-03-06
fix(ui): disabled group functionality for check query builder

### LINKED ISSUE #16956: Disable group by in query builder in check query builder

Asalem1:
Closes #16956

### Problem

Group functionality shouldn't exist on the query builder when setting up checks for alerts

### Solution

Created custom logic to only display `filter` as an option and disable the dropdown to reduce functionality confusion

disable_group


ui/src/timeMachine/components/builderCard/BuilderCardDropdownHeader.tsx
```diff
@@ -21,15 +21,30 @@ export default class BuilderCardDropdownHeader extends PureComponent<Props> {
...
+    const menu = () => null
     return (
       <div className="builder-card--header" data-testid={testID}>
-        <SelectDropdown
-          options={options}
-          selectedOption={selectedOption}
-          testID="select-option-dropdown"
-          onSelect={onSelect ? onSelect : emptyFunction}
-        />
+        {options.length === 1 && (
```
hoorayimhelping: Two comments on this block:

Generally: if you're going to do a statement of the form
```
if condition then foo
if !condition then bar
```
it's typically more common, and easier to follow an `if/else`
```jsx
{options.length === 1 ? (<Dropdown />) : (<SelectDropdown />)}
```

Second: This method is inferring what state to put itself into based on, well an inference, rather than checking values or the state of the app.

The length of the options array being 1 _may_ indicate that we should only show the filter, but it may also indicate something else. The issue here isn't hiding group when we're in the checks overlay, it's hiding group when the options list has one item in it - we may only want to show group, but the inference is that if there's only a single item, it must show filter.

What we're interested in is whether this is happening in the checks overlay, right? So how about passing `isInCheckOverlay` as a prop down to `BuilderCardDropdownHeader` - if the prop is present and true, only show `filter`, otherwise show `group` and `filter`.

I know it seems persnickety, but these small things add up in a codebase. I prefer this because it moves the logic into a single place, and it makes it clear, in context, what the intent is - you don't have to look through two files to see what's happening. It also concentrates the logic - rather than spreading the logic to be implicit in the length of the array, which is set based on whether we're in the check overlay, it instead renders a different control based on whether we're in the check overlay. It's like a 'one hop' logic resolution, versus a 'two hop' one with the options array.

ui/src/timeMachine/components/builderCard/BuilderCardDropdownHeader.tsx
```diff
@@ -20,17 +21,38 @@ export default class BuilderCardDropdownHeader extends PureComponent<Props> {
...
-    const {children, options, onSelect, selectedOption, testID} = this.props
+    const {
+      children,
+      isInCheckOverlay,
+      options,
+      onSelect,
+      selectedOption,
+      testID,
+    } = this.props
+    const dropdownOptions = isInCheckOverlay ? ['filter'] : options
```
hoorayimhelping: this is better. the logic is now all in a single component and it's easier to follow.

However, it's mutating internal state needlessly. The length and content of options doesn't need to change - the options should always be `['filter', 'group']` - that represents what options are available to this dropdown. It's a dependency and it's communicating that it needs this dependency to function; we don't want to mutate that dependency. We can reduce this complexity and make the component conceptually simpler by not needing to worry about what the state of `options` is.

```suggestion

```

ui/src/timeMachine/components/builderCard/BuilderCardDropdownHeader.tsx
```diff
@@ -20,17 +21,38 @@ export default class BuilderCardDropdownHeader extends PureComponent<Props> {
...
+            testID="disabled--filter-dropdown"
+          />
+        ) : (
+          <SelectDropdown
+            options={dropdownOptions}
+            selectedOption={selectedOption}
+            testID="select-option-dropdown"
+            onSelect={onSelect ? onSelect : emptyFunction}
+          />
+        )}
```
hoorayimhelping:
```suggestion
        {isInCheckOverlay ? (
          <Dropdown
            button={button}
            menu={menu}
            testID="disabled--filter-dropdown"
          />
        ) : (
          <SelectDropdown
            options={options}
            selectedOption={selectedOption}
            testID="select-option-dropdown"
            onSelect={onSelect ? onSelect : emptyFunction}
          />
        )}
```

following up on the above comment. this is conceptually simpler because it doesn't introduce a new variable. that is the main motivation in asking for this change - i know it's annoying to go back and change a pr like this, but reducing [accidental complexity](https://en.wikipedia.org/wiki/No_Silver_Bullet#Summary) is one of the most valuable things an engineer can spend time on. the complexity of having a new variable (`dropdownOptions`) encode the state of the component, when we already have a variable to do that (`isInCheckOverlay`) is extra complexity that doesn't need to be added.

gueluelue:
Is the group functionality not intended for checks or is it just currently not implemented correctly and thus turned off via the UI?

Asalem1:
Hi @gueluelue thanks for the question. It's actually somewhere in between where the functionality was not intended for checks at the moment, and so it could not be implemented correctly for checks.

PR #17119, opened 2020-03-06
chore(github): change CODEOWNERS for tools-team

jademcgough:
We own the pkger tool, so adding that to CODEOWNERS.

Removing api-team, as that team no longer exists.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -8,6 +8,9 @@
 #
 http/swagger.yml @influxdata/monitoring-team
 
+# dev tools
+/pkger/ @influxdata/tools-team
+
 # Storage code
 /storage/ @influxdata/storage-team-assigner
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -4,9 +4,9 @@
...
 # https://help.github.com/en/articles/about-code-owners
 
-# API and monitoring team will help to review swagger.yml changes.
+# monitoring team will help to review swagger.yml changes.
 #
-http/swagger.yml @influxdata/api @influxdata/monitoring-team
+http/swagger.yml @influxdata/monitoring-team
 
 # Storage code
 /storage/ @influxdata/storage-team-assigner
```
jsteenb2: I don't see the add for pkger anywhere 🤔

jademcgough: derp, lost it just before making the commit

PR #17021, opened 2020-02-26
feat(transpiler): support difference(), group by wildcard, and correc…

yzhang1991:
Supported the `difference()` function, `GROUP BY *`, corrected the `_time` column values

Also, uncommented the test cases for `stddev` and `spread`

PR #17122, opened 2020-03-06
feat(query/influxql): fix cumulative_sum transpilation

### LINKED ISSUE #10732: Transpiler: Implement cumulative_sum

fchikwekwe:
Closes #10732


query/influxql/transpiler_test.go
```diff
@@ -124,7 +124,7 @@ func TestTranspiler_Compile(t *testing.T) {
 		{s: `SELECT derivative(distinct(value)), difference(distinct(value)) FROM cpu WHERE time >= now() - 1m GROUP BY time(5s)`},
 		{s: `SELECT moving_average(distinct(value), 3) FROM cpu WHERE time >= now() - 5m GROUP BY time(1m)`},
 		{s: `SELECT elapsed(distinct(value)) FROM cpu WHERE time >= now() - 5m GROUP BY time(1m)`},
-		{s: `SELECT cumulative_sum(distinct(value)) FROM cpu WHERE time >= now() - 5m GROUP BY time(1m)`},
+		//{s: `SELECT cumulative_sum(distinct(value)) FROM cpu WHERE time >= now() - 5m GROUP BY time(1m)`},
```
rockstar: Why comment this and the others out? What's the benefit of commenting rather than removing with a VCS?

fchikwekwe: What's a VCS?

fchikwekwe: Also, these tests should pass at some point. Most of the failing ones are currently dependent on other unimplemented functions. We can bring them back in when selectors like `top()` `bottom()` etc are implemented.

jlapacik: Why is this one commented out? It looks like it passed before. Does it not pass anymore?

query/influxql/function.go
```diff
@@ -134,6 +134,21 @@ func createFunctionCursor(t *transpilerState, call *influxql.Call, in cursor, no
 		}
 		cur.value = value
 		cur.exclude = map[influxql.Expr]struct{}{call.Args[0]: {}}
+	case "cumulative_sum":
+		value, ok := in.Value(call.Args[0])
```
jlapacik: What does `value` represent?

query/influxql/transpiler_test.go
```diff
@@ -326,12 +326,12 @@ func TestTranspiler_Compile(t *testing.T) {
...
+		//{s: `SELECT cumulative_sum(value) FROM myseries group by time(1h)`, err: `aggregate function required inside the call to cumulative_sum`},
+		//{s: `SELECT cumulative_sum(top(value)) FROM myseries where time < now() and time > now() - 1d group by time(1h)`, err: `invalid number of arguments for top, expected at least 2, got 1`},
+		//{s: `SELECT cumulative_sum(bottom(value)) FROM myseries where time < now() and time > now() - 1d group by time(1h)`, err: `invalid number of arguments for bottom, expected at least 2, got 1`},
+		//{s: `SELECT cumulative_sum(max()) FROM myseries where time < now() and time > now() - 1d group by time(1h)`, err: `invalid number of arguments for max, expected 1, got 0`},
+		//{s: `SELECT cumulative_sum(percentile(value)) FROM myseries where time < now() and time > now() - 1d group by time(1h)`, err: `invalid number of arguments for percentile, expected 2, got 1`},
+		//{s: `SELECT cumulative_sum(mean(value)) FROM myseries where time < now() and time > now() - 1d`, err: `cumulative_sum aggregate requires a GROUP BY interval`},
```
jlapacik: I'm assuming these tests still fail, just with different error messages?

fchikwekwe: Yes, they do still fail. For queries that refer to `top()`, `bottom()`I'm pretty sure it is because they're unimplemented. When I last talked to @yzhang1991 about it, the plan was to revisit this file when more functions are implemented since its kind of hard to fix them at the moment.

jlapacik: Is it possible to just update the error message instead of skipping them? To guard against any potential regressions.

query/influxql/function.go
fchikwekwe: This value becomes the `functionCursor` value. I'm not super sure what its utility is, but it was required for the implementation of functionCursor further down in the file. From lines 398 onward.

query/influxql/transpiler_test.go
fchikwekwe: I could do that, but that would be changing the test case itself. I feel like that does less to guard against regressions and more to just make the test cases fit the current output. Since the output at this stage is wrong, I think its either better to just not merge this change until those other changes have been merged than to change the test cases.

query/influxql/transpiler_test.go
fchikwekwe: As far as I can tell, this test case hasn't been passing since I added these changes. Same as the other ones.

query/influxql/transpiler_test.go
fchikwekwe: FWIW, I was basing commenting out these test cases off this change here: https://github.com/influxdata/influxdb/pull/17021/files#diff-9f6822e7a7cf72e202046334e5da0bef

PR #17133, opened 2020-03-06
refactor(task/backend): move the task/backend constants to the global package

jsternberg:
This moves a few types and constants to the global package so it can be
used without importing the `task/backend` package. These constants are
referenced in non tasks-specific code.

This is needed to break a dependency chain where the task backend will
call into the flux runtime to perform parsing or evaluation of a script
and to prevent the http package from inheriting that dependency.


lyondhill:
The interface is an internal only interface intended for the task system to consume, I am not sure we should move it to influxdb root. That change would imply that we have a public api for these functions and we definitely shouldn't.

jsternberg:
Do you have a recommended place to move this? I just need it to be outside of `task/backend` which has the implementation.

lyondhill:
the "private" ish service is `type TaskControlService interface ` as I do a quick glance around I see the backend imports are really only being used for the status types..
```
type TaskStatus string

const (
	TaskActive   TaskStatus = "active"
	TaskInactive TaskStatus = "inactive"

	DefaultTaskStatus TaskStatus = TaskActive
)

type RunStatus int

const (
	RunStarted RunStatus = iota
	RunSuccess
	RunFail
	RunCanceled
	RunScheduled
)
```
You would know better because you have looked more recently. Maybe we should move the status types but not the `TaskControlService` @jsternberg would that work?

jsternberg:
The `TaskControlService` interface is referenced in the `kv` package, but it's a trivial reference.

```
var _ backend.TaskControlService = (*Service)(nil)
```

If I remove that, I think I might just be able to only move the constants.

PR #17141, opened 2020-03-08
fix: enable variables again

### LINKED ISSUE #17101: [ UI Dashboards > Flux Editor ] Variables of type CSV or Query not fetched correctly
__Steps to reproduce:__

1. Create variables of type Query or CSV 
2. Create a dashboard and add a cell.  
3. Edit the flux script for the cell
4. Open the variable tab of the flux editor
5. Hover over one of the CSV or Query variables - the timemachine variable popover appears
6. Click the value dropdown in the popover 

__Expected behavior:__

Expected the dropdown list to contain the variable values defined by CSV or the Query 

__Actual behavior:__

The dropdown list in the popover contains only numbers e.g. `{0,1,2,3,..}`

__Screencast__

VariableValuesNotFetched01

__Related__ 

A similar issue with map type variables was recently fixed. 

#17026 

__Environment info:__

```
Linux 4.15.0-88-generic x86_64
```
Detected in latest nightly builds
```
ts=2020-03-05T08:57:26.141285Z lvl=info msg="Welcome to InfluxDB" log_id=0LMUTLVG000 version=nightly commit=b1b2444fa build_date=2020-03-05T08:57:26Z
```
Browser: __Chrome 80__

__Config:__

Started with e2e testing enabled. 

```
--e2e-testing=true
```

### LINKED ISSUE #17123: Variable dropdowns are not updating with the value
saw this in latest master

Kapture 2020-03-06 at 10 39 20

drdelambre:
closes #17101
closes #17123

The issue is that when the variables feature was expanded to support maps, it ignored that arrays still existed. this just adds back that concept


ui/src/variables/utils/resolveSelectedValue.ts
```diff
@@ -1,5 +1,3 @@
-import {VariableMapObject} from 'src/types'
```
Asalem1: Let's remove this this type from the `variables` file (and also any other reference to it)

ui/src/variables/utils/resolveSelectedValue.ts
```diff
@@ -17,10 +15,25 @@ export const resolveSelectedKey = (
...
 export const resolveSelectedValue = (
-  values: VariableMapObject,
+  values: string[],
   selectedKey: string,
   defaultSelection?: string
 ): string => {
+  if (
+    values &&
+    typeof values === 'object' &&
+    Object.prototype.toString.call(values) === '[object Array]'
```
Asalem1:
```suggestion
    Array.isArray(values)
```

ui/src/timeMachine/components/variableToolbar/VariableTooltipContents.tsx
```diff
@@ -54,8 +54,11 @@ const VariableTooltipContents: FunctionComponent<Props> = ({
   onSelectVariableValue,
 }) => {
   let dropdownItems = get(values, 'values', []) || []
+  const isObject =
+    typeof dropdownItems === 'object' &&
+    Object.prototype.toString.apply(dropdownItems) !== '[object Array]'
```
Asalem1:
```suggestion
   Array.isArray(dropdownItems) !== true
```

ui/src/variables/utils/resolveSelectedValue.ts
```diff
@@ -17,10 +15,25 @@ export const resolveSelectedKey = (
...
+    if (values.includes(selectedKey)) {
+      return selectedKey
+    }
+
+    if (values.includes(defaultSelection)) {
+      return defaultSelection
+    }
+
+    return values[0]
+  }
```
Asalem1: Is this entire if statement a particular variable condition (like a map)? If so, we could export it into it's own helper function to make things a little clearer

drdelambre: i have no idea why we would add a helper function to a helper function of a helper function if not for a case made around reusability, but we also don't need to reuse this at all outside of this context. the "if array, do this, else do this" seems pretty clear?

ui/src/variables/utils/resolveSelectedValue.ts
drdelambre: it still gets used, just not externally

Asalem1: Do we still want to use it? Does it still serve a purpose? I added that type in to enforce variable mapping types.

drdelambre:
i dont understand how tests allowed the previous PR to get through, so i'm having a hard time figuring out which ones to write

ui/src/variables/utils/resolveSelectedValue.ts
drdelambre: i found it useful. i was even going to extend it to more parts of the app, but typescript decided to... die

Asalem1: 🤷‍♂

ui/src/variables/utils/resolveSelectedValue.ts
Asalem1: Totally. So then maybe we can refactoring the original condition to be:

```
Array.isArray(values)
```

rather than:

```
 if (
    values &&
    typeof values === 'object' &&
    Object.prototype.toString.call(values) === '[object Array]'
  )

Comment (drdelambre) [reply ui/src/variables/utils/resolveSelectedValue.ts]: 👍

**Comment by russorat** [comment]
let's prioritize getting this merged today.

**Comment by drdelambre** [comment]
[_refactoring intensifies_]

**Comment by drdelambre** [comment]
@Asalem1 I made some changes after writing more tests to expand on "arrays and objects should work". I found myself rewriting a lot of functionality for displaying a dropdown of variables, so i ripped it all out and normalized around the dashboard style dropdown. this allows us to make the "is it an array or object" call in one place at the selector, to consolidate the nastiness.

i figured i've made enough changes to warrant another review

File: ui/src/timeMachine/components/variableToolbar/VariableTooltipContents.tsx [hunk 5]
```diff
@@ -67,56 +61,56 @@ const VariableTooltipContents: FunctionComponent<Props> = ({
...
-        )}
+        <Dropdown
+          style={{width: '200px'}}
+          testID="variable--tooltip-dropdown"
+          button={(active, onClick) => (
+            <Dropdown.Button
+              active={active}
+              onClick={onClick}
+              testID="variable-dropdown--button"
+              status={status}
```
Asalem1: got user feedback that it would be helpful to disable the dropdown if there's only 1 option available. Not gonna push for it this time but something to keep in mind moving forward

PR #17145, opened 2020-03-09
feat(kv): support for migrations

GeorgeMac:
This change introduces a new `kv.Migrator` and associated migrations types.
This change sets a new precedent for how we introduce new resource types and indexes into the `kv` package within InfluxDB.

The new `Migrator` type adds support for managing these ongoing types of change as migrations. A migration is effectively a change to kv schema (defined as an `Up(...)` function operation) and a definition of how to revert the operation (defined as a `Down(...)` function operation).

The kinds of "schema" changes expected to be managed by these types (in the near-team) are KV buckets and indexes.

By default, `inmem` and `bolt` kv implementations should continue to operate as they always have. On initialization of the KV service, when backed by one of these KV stores implementations the migrations are all applied automatically.

Outstanding:

- [x] Get https://github.com/influxdata/influxdb/pull/16910 shipped
- [x] Consider either moving the introduction of the new indexes into own PR.
- [x] move tests into `testing` package and invoke for both `inmem` and `bolt`.

Checklist:

- [x] Rebased/mergeable
- [x] Tests pass


kv/migration.go
```diff
@@ -0,0 +1,332 @@
...
+	influxdb "github.com/influxdata/influxdb"
+	"go.uber.org/zap"
+)
+
+var (
+	migrationBucket = []byte("migrationsv1")
+
+	// ErrMigrationSpecNotFound is returned when a migration specification is missing
+	// for an already applied migration.
+	ErrMigrationSpecNotFound = errors.New("migration specification not found")
```
jsteenb2: be good to make this an `inflxudb.Error` type with the necessary code/msg

GeorgeMac: This kind of error shouldn't really ever reach a user

GeorgeMac: It happens on initialize or through manual invocation via a cli.

kv/index.go
```diff
@@ -180,6 +179,51 @@ func indexKeyParts(indexKey []byte) (fk, pk []byte, err error) {
...
+	opts []PopulateOption
+}
+
+// Name returns a readable name for the index migration.
+func (i *IndexMigration) Name() string {
+	return fmt.Sprintf("add index %q", string(i.IndexBucket()))
+}
+
+// Up initializes the index bucket and populates the index.
+func (i *IndexMigration) Up(ctx context.Context, store Store) (err error) {
```
brettbuddin: Would it be better to provide a helper method like `(i *IndexMigration) upErrWrap(err error) error` and delegate to that instead of doing this assignment in `defer`? It's a bit confusing to me.

kv/migration.go
```diff
@@ -0,0 +1,332 @@
...
+// Migrations are applied in order from the lowest indexed migration in a down state.
+//
+// For example, given:
+// 0001 add bucket foo         | (up)
+// 0002 add bucket bar         | (down)
+// 0003 add index "foo on baz" | (down)
+//
+// Up would apply migration 0002 and then 0003.
+func (m *Migrator) Up(ctx context.Context, store Store) (err error) {
+	defer func() {
```
brettbuddin: Similar situation here. I'm not sure if there is a better way, but these assignments in `defer` always confuse me at first read.

PR #17148, opened 2020-03-09
refactor(http): remove the spec and update lang usage

jsternberg:
This removes the spec and updates the lang package usage to make use of
passing in the runtime as a parameter.

It removes all direct dependendencies on the flux runtime from the http
package.


jlapacik:
The spec is currently unused?

jsternberg:
To my knowledge the repl no longer uses the repl endpoint since it was only used by the querier interface within influx.

jsternberg:
It looks like I didn't modifying the mock compiler in the flux repository for the new interface and it is used by influxdb.

https://github.com/influxdata/flux/pull/2601

PR #17154, opened 2020-03-09
fix(ui): fixes issue where submit btn was only loading onClick

### LINKED ISSUE #15668: ctrl + enter to submit a query doesn't update the "submit" button icon
as a user
when i use ctrl + enter to submit a query
then i would like some indication that the query is executing

in the gif below, i am hitting ctrl+enter to submit the query. if i didn't have my network tab open, i wouldn't have any indication that the query was running.

Kapture 2019-10-30 at 13 54 06

Asalem1:
Closes #15668

### Problem

Hitting `ctrl + enter` in the script editor would submit the query without enabling the loading state of the button

### Solution

Removed property that was only enabling the loading state of the button when the button was clicked. It seems like this was set by design so I'd love another opinion on why this was originally introduced and to see if there's a more correct solution.

submit-btn-load


Asalem1:
After speaking with @russorat about the original issue - it seems like it may have been addressed by a different UI modification. Put simply, this PR was meant to address an issue where users didn't have sufficient feedback to their actions whenever using the hot keys to submit queries. However, recent UI changes indicate that the data is reloading, and therefore renders these changes unnecessary

Issue #15302, opened 2019-09-30
influx cli support multiple profiles


gitirabassi:
as of now you can configure the `influx` cli with either `INFLUX_HOST` and `INFLUX_TOKEN` or the token setup in `~/.influxdbv2/credentials` and this is a good starting point.

But soon will rise the need to be handling multiple instances and endpoints or credentials for th e same instance for testing/debug purposes. I'd like to be able to have profiles just like the `aws` cli with the `~/.influxdbv2/credentials` file looking something like

```toml
[default]
address = "http://localhost:9999"
token = "xxxxx"

[staging]
address = "https://example.com:9999"
token = "xxxx"

[frankfurt]
address = "https://eu-central-1-1.aws.influxcloud.net"
token = "xxxx"
```

and the usage can be along the lines of `influx --profile=frankfurt bucket find` or setup the `INFLUX_PROFILE=frankfurt` env variable

jademcgough:
I love this idea, it's like kubenv

russorat:
i agree this would be useful. that was the original reason we used a file.

jademcgough:
Let's spike this out and present it at demo day to see if product likes the UX.

My preference is to borrow heavily from kubectx, as that's an experience that developers are familiar with: https://github.com/ahmetb/kubectx#kubectx1, but open to other ideas on this.

PR #17170, opened 2020-03-10
feat(cmd/influx): add profile management

kelwang:
Closes https://github.com/influxdata/influxdb/issues/15302

add multiple profiles feature as suggested in the ticket

add the ability to setup against another cluster, even if you have a profile file

- [x] Rebased/mergeable
- [x] Tests pass

demo
demo2

### DIFF
--- a/cmd/influx/config.go
+++ b/cmd/influx/config.go
@@ -1,0 +1,265 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/cmd/influx/config"
+	"github.com/spf13/cobra"
+)
+
+func cmdConfig(f *globalFlags, opt genericCLIOpts) *cobra.Command {
+	path, dir, err := defaultConfigPath()
+	if err != nil {
+		panic(err)
+	}
+	builder := cmdConfigBuilder{
+		genericCLIOpts: opt,
+		globalFlags:    f,
+		svc: config.LocalConfigsSVC{
+			Path: path,
+			Dir:  dir,
+		},
+	}
+	builder.globalFlags = f
+	return builder.cmd()
+}
+
+type cmdConfigBuilder struct {
+	genericCLIOpts
+	*globalFlags
+
+	name   string
+	url    string
+	token  string
+	active bool
+	org    string
+
+	svc config.ConfigsService
+}
+
+func (b *cmdConfigBuilder) cmd() *cobra.Command {
+	cmd := b.newCmd("config", nil)
+	cmd.Short = "Config management commands"
+	cmd.Run = seeHelp
+	cmd.AddCommand(
+		b.cmdCreate(),
+		b.cmdDelete(),
+		b.cmdUpdate(),
+		b.cmdList(),
+	)
+	return cmd
+}
+
+func (b *cmdConfigBuilder) cmdCreate() *cobra.Command {
+	cmd := b.newCmd("create", nil)
+	cmd.RunE = b.cmdCreateRunEFn
+	cmd.Short = "Create config"
+	cmd.Flags().StringVarP(&b.name, "name", "n", "", "The config name (required)")
+	cmd.MarkFlagRequired("name")
+	cmd.Flags().StringVarP(&b.token, "token", "t", "", "The config token (required)")
+	cmd.MarkFlagRequired("token")
+	cmd.Flags().StringVarP(&b.url, "url", "u", "", "The config url (required)")
+	cmd.MarkFlagRequired("url")
+
+	cmd.Flags().BoolVarP(&b.active, "active", "a", false, "Set it to be the active config")
+	cmd.Flags().StringVarP(&b.org, "org", "o", "", "The optional organization name")
+	return cmd
+}
+
+func (b *cmdConfigBuilder) cmdCreateRunEFn(*cobra.Command, []string) error {
+	pp, err := b.svc.ParseConfigs()
+	if err != nil {
+		return err
+	}
+	p := config.Config{
+		Host:   b.url,
+		Token:  b.token,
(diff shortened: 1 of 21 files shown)


http/backup_service.go
```diff
@@ -22,8 +22,12 @@ import (
 	"go.uber.org/zap"
 )
 
+// DefaultTokenFile is deprecated, and will be only used for migration.
 const DefaultTokenFile = "credentials"
 
+// DefaultProfilesFile stores cli credentials and hosts.
+const DefaultProfilesFile = "profiles"
```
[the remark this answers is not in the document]
kelwang: because DefaultTokenFile is here

profile.go
```diff
@@ -0,0 +1,135 @@
+package influxdb
```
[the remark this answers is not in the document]
kelwang: yeah, we can move

cmd/influx/profile.go
```diff
@@ -0,0 +1,195 @@
...
+	genericCLIOpts
+	*globalFlags
+
+	name   string
+	url    string
+	token  string
+	active bool
+	org    string
+
+	influxdb.ProfilesService
```
jsteenb2: field name for the interface

```
svc influxdb.ProfileService
```

cmd/influx/config.go
```diff
@@ -0,0 +1,196 @@
...
+	return b.svc.WriteConfigs(pp)
+}
+
+func (b *cmdConfigBuilder) cmdUpdate() *cobra.Command {
+	cmd := b.newCmd("set", b.cmdUpdateRunEFn)
+	cmd.Short = "Update config"
+	cmd.Flags().StringVarP(&b.name, "name", "n", "", "The config name (required)")
+	cmd.MarkFlagRequired("name")
+
+	cmd.Flags().StringVarP(&b.token, "token", "t", "", "The config token (required)")
```
jsteenb2: the global vs local `token` flag feels weird still, we need to figure out how to remedy this 🤔

http/backup_service.go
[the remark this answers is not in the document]
kelwang: replaced the token file with the new config file

cmd/influx/config/mock/config.go
```diff
@@ -0,0 +1,19 @@
+package mock
```
jsteenb2: can we mvoe this file to the `config` pkg and not export it.

cmd/influx/config.go
kelwang: agree, but maybe later.

cmd/influx/setup.go
```diff
@@ -2,13 +2,16 @@ package main
...
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/influxdata/influxdb"
 	platform "github.com/influxdata/influxdb"
```
jsteenb2: lets consolidate here, I'd much prefer to see us drop `platform`, but that's my $0.02

cmd/influx/setup.go
```diff
@@ -47,27 +53,37 @@ func setupF(cmd *cobra.Command, args []string) error {
...
-		return &platform.Error{
-			Code: platform.EConflict,
-			Msg:  fmt.Sprintf("token already exists at %s", dPath),
+		existingConfigs, _ = config.LocalConfigsSVC{
+			Path: dPath,
+			Dir:  dir,
+		}.ParseConfigs()
+		// ignore the error if found nothing
+		if setupFlags.name == "" {
+			return errors.New("flag name is required if you already have existing configs")
```
jsteenb2: can we just fall back to default here?

cmd/influx/setup.go
```diff
@@ -81,12 +97,24 @@ func setupF(cmd *cobra.Command, args []string) error {
...
+			Token: result.Auth.Token,
+			Org:   result.Org.Name,
+			Host:  flags.Host,
+		}
+		config.LocalConfigsSVC{
+			Path: dPath,
+			Dir:  dir,
+		}.WriteConfigs(existingConfigs)
+	} else {
+		err = writeConfigToPath(result.Auth.Token, result.Org.Name, dPath, dir)
```
jsteenb2: why isn't hte LocolConfigSVC called here? can it be instead? feels like we should encapsulate this into the service

cmd/influx/setup.go
```diff
@@ -81,12 +97,24 @@ func setupF(cmd *cobra.Command, args []string) error {
...
+	}
 	if err != nil {
-		return fmt.Errorf("failed to write token to path %q: %v", dPath, err)
+		return fmt.Errorf("failed to write profile to path %q: %v", dPath, err)
 	}
 
-	fmt.Println(string(promptWithColor("Your token has been stored in "+dPath+".", colorCyan)))
+	fmt.Println(string(promptWithColor("Your profile has been stored in "+dPath+".", colorCyan)))
```
jsteenb2: lets add the name as well, as a final gut check. Maybe something like

```go
)"Profile %s has been stored in $s.", name, dPath)
```

cmd/influx/setup.go
kelwang: most likely, it could error out later in the conflict. I would prefer to make this requirement clear here.

cmd/influx/setup.go
kelwang: and you are absolutely right

cmd/influx/setup.go
jsteenb2: this new requirement breaks backwards compatability for anyone who has scripts to automate setup in CI or elsewhere. I don't think this is a viable choice.

jademcgough:
@kelwang could you drop a demo of this in channel before merging? I think it'd be valuable to get feedback on the CLI UX from the team.

PR #17171, opened 2020-03-10
feat(tenant): Build the crud actions for the new tenant service

lyondhill:
To facilitate the isolation of domain specific systems we need to have a
isolated CRUD action system that we can layer business logic on top.

### DIFF
--- a/tenant/storage.go
+++ b/tenant/storage.go
@@ -1,0 +1,115 @@
+package tenant
+
+import (
+	"context"
+	"errors"
+	"sync"
+
+	"github.com/influxdata/influxdb/kv"
+)
+
+var errRollback = errors.New("rollback")
+
+type Store struct {
+	kvStore kv.Store
+}
+
+type TxType int
+
+const (
+	Write TxType = iota
+	Read
+)
+
+type Tx struct {
+	kv.Tx
+	close    chan error // Commit() closes the channel, and Rollback sends a rollback error on the channel
+	response chan error // used to get the response from the transaction handling go routine
+}
+
+func (tx *Tx) Commit() error {
+	close(tx.close)
+	return <-tx.response
+}
+
+func (tx *Tx) Rollback() error {
+	tx.close <- errRollback
+	close(tx.close)
+	err := <-tx.response
+	if err != errRollback {
+		return err
+	}
+	return nil
+}
+
+func NewStore(kvStore kv.Store) (*Store, error) {
+	st := &Store{
+		kvStore: kvStore,
+	}
+	return st, st.setup()
+}
+
+func (s *Store) Begin(txType TxType) (*Tx, error) {
+	tx := Tx{
+		close:    make(chan error, 1),
+		response: make(chan error, 1),
+	}
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		txFunc := s.kvStore.View
+		if txType == Write {
+			txFunc = s.kvStore.Update
+		}
+		tx.response <- txFunc(context.Background(), func(kvTx kv.Tx) error {
+			tx.Tx = kvTx
+			wg.Done()
+			return <-tx.close
+		})
+	}()
+	wg.Wait()
+	return &tx, nil
+}
+
+func (s *Store) setup() error {
+	tx, err := s.Begin(Write)
+	if err != nil {
+		return err
+	}
+
+	if _, err := tx.Bucket(userBucket); err != nil {
+		tx.Rollback()
+		return err
+	}
+
+	if _, err := tx.Bucket(userIndex); err != nil {
+		tx.Rollback()
+		return err
+	}
+
+	if _, err := tx.Bucket(urmBucket); err != nil {
+		tx.Rollback()
+		return err
+	}
+	if _, err := tx.Bucket(organizationBucket); err != nil {
+		tx.Rollback()
+		return err
+	}
+
+	if _, err := tx.Bucket(organizationIndex); err != nil {
+		tx.Rollback()
+		return err
+	}
+
+	if _, err := tx.Bucket(bucketBucket); err != nil {
+		tx.Rollback()
+		return err
+	}
+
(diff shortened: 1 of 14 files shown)


tenant/storage.go
```diff
@@ -0,0 +1,115 @@
...
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		txFunc := s.kvStore.View
+		if txType == Write {
+			txFunc = s.kvStore.Update
+		}
+		tx.response <- txFunc(context.Background(), func(kvTx kv.Tx) error {
+			tx.Tx = kvTx
+			wg.Done()
```
GeorgeMac: I would consider defering this at the top. A panic in `View` or `Update` or elsewhere in this function could lead Done() not getting called.

GeorgeMac: I see that you need `Done()` to be called before blocking on `<-tx.close`. It is also unlikely that `View` or `Update` would panic as well. But I think this does lead to a complex concurrency dance here.

Any reason we can't stick to the shape of the `kv.Tx` (i.e. `View(fn)` or `Update(fn)`) for the KV backed tenant implementation and then have a separate SQL backed version?
I worry we over complicate things by trying to create some in between storage abstraction which we try to shoe-horn both KV and SQL shaped systems behind.

tenant/storage_bucket.go
```diff
@@ -0,0 +1,418 @@
...
+	return s.GetBucket(ctx, tx, id)
+}
+
+type BucketFilter struct {
+	Name           *string
+	OrganizationID *influxdb.ID
+}
+
+func (s *Store) ListBuckets(ctx context.Context, tx *Tx, filter BucketFilter, opt ...influxdb.FindOptions) ([]*influxdb.Bucket, error) {
+	// this isn't a list action its a `GetBucketByName`
```
GeorgeMac: Is this really an error condition or just a bad access pattern by the caller?

tenant/storage_bucket.go
```diff
@@ -0,0 +1,418 @@
...
+
+	cursor, err := b.ForwardCursor(nil)
+	if err != nil {
+		return nil, err
+	}
+	defer cursor.Close()
+
+	count := 0
+	bs := []*influxdb.Bucket{}
+	for k, v := cursor.Next(); k != nil; k, v = cursor.Next() {
```
GeorgeMac: Missing `cursor.Err()` check.

tenant/storage_bucket.go
```diff
@@ -0,0 +1,418 @@
...
+		if o.Offset != 0 && count < o.Offset {
+			count++
+			continue
+		}
+		b, err := unmarshalBucket(v)
+		if err != nil {
+			continue
```
GeorgeMac: Out of interest why don't we care about this error? if we do not, do we want to log or comment as to why?

tenant/storage_bucket.go
```diff
@@ -0,0 +1,418 @@
...
+	cursor, err := b.ForwardCursor(nil)
+	if err != nil {
+		return nil, err
+	}
+	defer cursor.Close()
+
+	count := 0
+	bs := []*influxdb.Bucket{}
+	for k, v := cursor.Next(); k != nil; k, v = cursor.Next() {
+		if o.Offset != 0 && count < o.Offset {
```
GeorgeMac: Do we want to continue supporting the `offset` and `limit` form of pagination here?
I think we spoke about this in the past, around how you used the `after` approach in tasks? Is this a good opportunity to make the transition here for this new service?

tenant/storage_bucket.go
```diff
@@ -0,0 +1,418 @@
...
+		return nil, invalidBucketListRequest
+	}
+
+	// if we dont have any options it would be irresponsible to just give back all orgs in the system
+	if len(opt) == 0 {
+		opt = append(opt, influxdb.FindOptions{
+			Limit: influxdb.DefaultPageSize,
+		})
+	}
+	o := opt[0]
```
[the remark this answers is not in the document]
lyondhill: I like this idea. One thing I will have to be aware of is that the interface cant change yet because we need this new system to act in pairity with the old kv system. So we cant switch to the `After` style lookups yet. But I do like the pattern of passing find options.

tenant/storage_bucket.go
lyondhill: oops., I am pretty sure I forgot that in all the places

tenant/storage_bucket.go
lyondhill: We have to for now unfortunately. We need this new tenant service to drop in as a shadow system  and to confirm identical behavior with the existing kv system. After KV has been removed then Yes I definitely want to switch to the after type lookups

tenant/storage.go
lyondhill: I wanted to use the sql style so when we make the switch the Service layer (the part that uses this) would have very minimal changes. It might have been a bit preemptive.

tenant/storage_bucket.go
lyondhill: I believe it is a pattern we would like to avoid. The reason I added the error is just so when im building the service layer I would get immediate feedback if i accidentally called list when I meant a Get then tests would fail and I could fix it quickly.

tenant/storage_bucket.go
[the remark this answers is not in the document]
GeorgeMac: Ahh ok cool. I see.

tenant/storage_bucket.go
GeorgeMac: Makes sense 👍

tenant/storage_bucket.go
```diff
@@ -0,0 +1,418 @@
...
+		}
+
+		bs = append(bs, b)
+
+		if len(bs) >= o.Limit {
+			break
+		}
+	}
+
+	return bs, nil
```
GeorgeMac:
```suggestion
	return bs, cursor.Err()
```

PR #17187, opened 2020-03-10
build(flux): update Flux to v0.64.0

### LINKED ISSUE #17172: Update Flux dependency for the 1.8 branch
We need to ensure the 1.8 branch is using the most recent version of Flux.

jsternberg:
Fixes #17172.


jsternberg:
cc @dgnorton

flux/stdlib/influxdata/influxdb/rules.go
```diff
@@ -551,3 +566,55 @@ func isPushableFieldOperator(kind ast.OperatorKind) bool {
 
 	return false
 }
+
```
wolffcm: This code must be taken from 2.0 OSS?

jsternberg: Yes I sync this directory from 2.0 OSS when I do an update.

timhallinflux:
@sanderson -- for 1.8

flux/stdlib/influxdata/influxdb/to.go
```diff
@@ -0,0 +1,4 @@
+package influxdb
+
+// TODO(jsternberg): Implement the to method in influxdb 1.x.
```
wolffcm: Is there an issue for this?

jsternberg: I don't think so. It's been this way for awhile, but I think in previous syncs I was just deleting the file. I added this so I could remember that I did this.

jsternberg:
Most of the change was automated by using `rsync` and reapplying the patches. The `storage/reads` package had a big refactor last Friday so, after consulting with Jacob, I synced that one from last Thursday to before the refactor.

PR #17201, opened 2020-03-11
fix(authorizer): fix endpoint handler auth

affo:
This does not actually fix auth.

It only refactors it in a way that is conforming to other services.

Before:

```
Check that a user has permission to read/write an org to indirectly deduce that it can read/write endpoints
```

Now:

```
Check that a user has permission to read/write a endpoints (for the endpoint's org)
```


http/notification_endpoint.go
```diff
@@ -640,7 +638,6 @@ func (s *NotificationEndpointService) CreateNotificationEndpoint(ctx context.Con
 // UpdateNotificationEndpoint updates a single notification endpoint.
 // Returns the new notification endpoint after update.
 func (s *NotificationEndpointService) UpdateNotificationEndpoint(ctx context.Context, id influxdb.ID, ne influxdb.NotificationEndpoint, userID influxdb.ID) (influxdb.NotificationEndpoint, error) {
-	// userID is ignored since userID is grabbed off the http auth set on the client
```
affo: this is not true actually 🤔 

same above

PR #17202, opened 2020-03-11
fix(ui): fixed dashboard cells and check query builder update bug

### LINKED ISSUE #17130: Can't change selected field in existing monitor and alerting check on Cloud 2
If you have an existing threshold check in cloud 2 and try to change the field filter to use a different field, the GUI displays your new choice, but the query actually has selected all the existing fields.

To reproduce, create a new alert check where there are multiple fields to select from, for example "voltage", "current" and "penguins" - pick one of those for your check (say "voltage"), set a threshold and save the check. Open the saved check, deselect "voltage" and select "penguins" instead. Submit the query and examine the raw data returned. It will contain data as if you had selected all the fields even though the GUI shows one selected. You can get around this by deselecting all the filters, picking a different bucket and then select the original bucket and add your filters anew.

Asalem1:
Closes #17130

### Problem

1. Updates made to existing Checks were not registering the changes
2. Creating cells with Group as the aggregate function type were not maintaining the aggregate type

### Solution

Updated the `BuilderConfig tag` struct to account for the `AggregateFunctionType` that had been passed in by the UI but was not being stored on the API. The consequence of this action allows the UI to have access to the `aggregateFunctionType` as a reference for making changes in the query builder. Put simply, the existing query builder is built around the fact that an `aggregateFunctionType` will be passed in as a tag property, but it was not. 

Check Update Fix:
check_update_fix

Dashboard Cell Fix:
dashboard_cell_fix

### DIFF
--- a/pkger/models.go
+++ b/pkger/models.go
@@ -2638,7 +2638,7 @@ func (q queries) influxDashQueries() []influxdb.DashboardQuery {
 			EditMode: "advanced",
 		}
 		// TODO: axe this builder configs when issue https://github.com/influxdata/influxdb/issues/15708 is fixed up
-		newQuery.BuilderConfig.Tags = append(newQuery.BuilderConfig.Tags, influxdb.NewBuilderTag("_measurement", ""))
+		newQuery.BuilderConfig.Tags = append(newQuery.BuilderConfig.Tags, influxdb.NewBuilderTag("_measurement", []string{""}, "filter"))
 		iQueries = append(iQueries, newQuery)
 	}
 	return iQueries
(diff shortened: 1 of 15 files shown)


pkger/models.go
```diff
@@ -2638,7 +2638,7 @@ func (q queries) influxDashQueries() []influxdb.DashboardQuery {
 			EditMode: "advanced",
 		}
 		// TODO: axe this builder configs when issue https://github.com/influxdata/influxdb/issues/15708 is fixed up
```
drdelambre: anyone else curious about this foreboding line?

Asalem1: Yup, definitely shows it to @gavincabbage this morning, especially since the issue was closed a couple days ago. I'd love to see @desa 's opinion on this

PR #17206, opened 2020-03-11
feat(ui): implement new navigation design

alexpaxton:
Phase 2 of the clockface 2 rollout is implementing the new `TreeNav` navigation component as well as slightly re-organizing the items in the navbar. Both the old and new navigation are present and can be toggled with the `treeNav` feature flag

### Changes:

- Implement new `TreeNav` design behind a feature flag `treeNav`
  - Tree nav handles org switching in an overlay instead of custom nav components
- Create `navBarState` key in persisted state to handle expansion and collapse of the `TreeNav`
- Split `Settings` into `Org` and `Settings` to match navigation spec (also behind a feature flag)
  - Created a separate tabbed page for `Org`
- Duplicate `/settings/members` to `/members`
- Duplicate `/settings/profile` to `/about` (Rename "Org Profile" to "About")

### Bugs / Issues encountered

- Having the `TreeNav` set to `expanded` initially causes a bunch of e2e tests to fail. Upon further investigation the failing tests concerned the Alerting page, which features a layout that is very difficult to use on smaller (laptop) screens. By having the `TreeNav` expanded by default the alerting page layout is squished enough to make some user interactions impossible. If we want to have the nav expanded by default we'll have to make the Alerting page layout responsive

Screen Shot 2020-03-18 at 1 54 18 PM

### Preview

Screen Shot 2020-03-24 at 10 19 09 AM
Screen Shot 2020-03-24 at 10 19 15 AM
Screen Shot 2020-03-24 at 10 19 29 AM
Screen Shot 2020-03-24 at 9 30 57 AM
Screen Shot 2020-03-24 at 9 25 11 AM
Screen Shot 2020-03-18 at 3 13 27 PM

- [x] Rebased/mergeable
- [x] Tests pass


ui/src/pageLayout/components/UserWidget.tsx
```diff
@@ -0,0 +1,119 @@
...
+        />
+        <TreeNav.UserItem
+          id="logout"
+          label="Logout"
+          linkElement={className => (
+            <a className={className} href={logoutURL} />
+          )}
+        />
+      </CloudOnly>
+      <CloudExclude>
```
Asalem1: I'd be interested in seeing what would happen if we used the same `Logout` component for CLOUD and OSS since the `Logout` component calls the same `logoutUrl` if it's on "Cloud"

ui/src/pageLayout/containers/Nav.tsx
```diff
@@ -58,314 +54,141 @@ class SideNav extends PureComponent<Props, State> {
...
+const mdtp: DispatchProps = {
+  handleSetNavBarState: setNavBarState,
 }
 
 const mstp = (state: AppState): StateProps => {
   const isHidden = state.app.ephemeral.inPresentationMode
-  const orgs = getAll<Organization>(state, ResourceType.Orgs)
-  const org = getOrg(state)
-  const {me} = state
+  const navBarState = state.app.persisted.navBarState
```
Asalem1: i'd love to see this use `get` just so we don't end up with an error at some point saying `can't read property "persisted" of undefined`

ui/src/types/actions/app.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export interface CollapseNavTreeAction {
+  type: ActionTypes.CollapseNavTree
+}
+
+export type DelayEnablePresentationModeDispatcher = () => DelayEnablePresentationModeThunk
+
+export type DelayEnablePresentationModeThunk = (
+  dispatch: Dispatch<EnablePresentationModeAction>
+) => NodeJS.Timer
```
Asalem1: TIL

ui/src/pageLayout/components/OrgSwitcherItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+  router,
+}) => {
+  const orgSwitcherItemClass = classnames('org-switcher--item', {
+    'org-switcher--item__selected': selected,
+  })
+
+  const orgSwitcherIcon = selected ? IconFont.Checkmark : IconFont.CaretRight
+
+  const handleClick = (): void => {
+    onDismiss()
```
Asalem1: I imagine that we'll pull new data based on the new orgID, but it might be worth checking to see if we'd need to reset the app state for particular parts of the app

ui/src/pageLayout/containers/Nav.tsx
alexpaxton: Are you referring to `get` from lodash or something else?

ui/src/types/actions/app.ts
alexpaxton: I actually think this whole file is supposed to be deleted, might be here as a mishap during rebasing

alexpaxton:
After syncing with the application leads the plan is to roll this out early next week. Before then here's a checklist of things that need to happen:

- [ ] Wrap the new navbar in a feature flag
- [ ] Bring back the old navbar wrapped in an inverse feature flag
- [ ] Update the clockface dependency to the forthcoming 2.0.3 for some small but important improvements to `TreeNav`

ui/src/pageLayout/containers/Nav.tsx
Asalem1: Yes please

ui/src/pageLayout/components/UserWidget.tsx
alexpaxton: I'm gonna wrap this new navbar in a feature flag so we can test it out in acceptance, maybe that is something we can test. Would be nice to reduce separation between cloud/oss as much as possible

ui/src/index.tsx
```diff
@@ -460,6 +460,15 @@ class Root extends PureComponent {
                             path="checks/:checkID"
                             component={CheckHistory}
                           />
+                          <Route path="about" component={OrgProfilePage}>
+                            <Route path="rename" component={RenameOrgOverlay} />
+                          </Route>
+                          <Route path="about" component={OrgProfilePage}>
+                            <Route path="rename" component={RenameOrgOverlay} />
+                          </Route>
```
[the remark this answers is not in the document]
alexpaxton: I needed two different routes for the same page because in `treeNav` mode `about` has a different route than with the existing nav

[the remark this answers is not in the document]
Asalem1: I'm so confused. Aren't these the same routes?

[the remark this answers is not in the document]
alexpaxton: Currently the `about` page is located at `orgs/orgid/settings/about` but in the new tree nav it is `orgs/orgid/about`. Since both can exist I can't remove the old route just yet

ui/src/pageLayout/containers/TreeNav.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+@ErrorHandling
+class SideNav extends PureComponent<Props, State> {
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      isShowingOrganizations: false,
+    }
+  }
```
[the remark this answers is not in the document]
alexpaxton: I ended up removing state entirely from this component since it wasn't being used

ui/src/shared/tabbedPage/TabbedPageTabs.tsx
```diff
@@ -0,0 +1,66 @@
...
+import CloudOnly from 'src/shared/components/cloud/CloudOnly'
+import {FeatureFlag} from 'src/shared/utils/featureFlag'
+
+export interface TabbedPageTab {
+  text: string
+  id: string
+  cloudExclude?: boolean
+  cloudOnly?: boolean
+  featureFlag?: string
+  featureFlagEquals?: boolean
```
hoorayimhelping:
```suggestion
  featureFlagName?: string
  featureFlagValue?: boolean
```

i think `featureFlagName` `featureFlagValue` in context with each other show that we're trying to to get the value of a feature flag by its name.

ui/src/organizations/components/OrgNavigation.tsx
```diff
@@ -0,0 +1,110 @@
...
+        id: 'about',
+        onClick: () => {
+          router.push(`/orgs/${orgID}/about`)
+        },
+      },
+    ]
+
+    return (
+      <Tabs orientation={Orientation.Horizontal} size={ComponentSize.Large}>
+        {tabs.map(t => {
```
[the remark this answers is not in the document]
alexpaxton: That's actually not a bad idea to move the tabs objects into the hierarchy file.

PR #17207, opened 2020-03-11
feat(sampledata): Allow users to query buckets in 'other' org

### LINKED ISSUE #16803: Allow dashboard to run queries against "other" org

ebb-tide:
Closes #16803

- Dashboard Queries search query string for bucket name, and make the query to the org bucket belongs to instead of users current org. 
- Buckets are filtered by org at the UI level instead of the buckets request containing orgID in query parameter
- Fix getResources bug which would cause unmount on refetch of resources. 
- Query builder uses api call to get buckets instead of creating a buckets query. 

Will add bucket type filtering to get sampledata buckets, after extending bucket type.

### DIFF
--- a/ui/src/buckets/actions/thunks.ts
+++ b/ui/src/buckets/actions/thunks.ts
@@ -17,11 +17,13 @@ import {
   Bucket,
   BucketEntities,
   Label,
+  ResourceType,
 } from 'src/types'
 
 // Utils
 import {getErrorMessage} from 'src/utils/api'
 import {getOrg} from 'src/organizations/selectors'
+import {getLabels, getStatus} from 'src/resources/selectors'
 
 // Actions
 import {
@@ -46,7 +48,6 @@ import {
   addBucketLabelFailed,
   removeBucketLabelFailed,
 } from 'src/shared/copy/notifications'
-import {getLabels} from 'src/resources/selectors'
 
 type Action = BucketAction | NotifyAction
 
@@ -55,8 +56,11 @@ export const getBuckets = () => async (
   getState: GetState
 ) => {
   try {
-    dispatch(setBuckets(RemoteDataState.Loading))
-    const org = getOrg(getState())
+    const state = getState()
+    if (getStatus(state, ResourceType.Buckets) === RemoteDataState.NotStarted) {
+      dispatch(setBuckets(RemoteDataState.Loading))
+    }
+    const org = getOrg(state)
 
     const resp = await api.getBuckets({query: {orgID: org.id}})
 
--- a/ui/src/checks/actions/thunks.ts
+++ b/ui/src/checks/actions/thunks.ts
@@ -18,7 +18,7 @@ import {reportError} from 'src/shared/utils/errors'
 import {createView} from 'src/views/helpers'
 import {getOrg} from 'src/organizations/selectors'
 import {toPostCheck, builderToPostCheck} from 'src/checks/utils'
-import {getAll} from 'src/resources/selectors'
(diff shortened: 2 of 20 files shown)


ui/src/buckets/actions/thunks.ts
```diff
@@ -55,17 +56,22 @@ export const getBuckets = () => async (
...
+    const org = getOrg(state)
 
-    const resp = await api.getBuckets({query: {orgID: org.id}})
+    const resp = await api.getBuckets({query: {}})
 
     if (resp.status !== 200) {
       throw new Error(resp.data.message)
     }
 
+    const bucketsInOrg = resp.data.buckets.filter(b => b.orgID === org.id)
```
drdelambre: the idea is to expand this once the sample data org is created to hardcode another org id?

ebb-tide: when I can add sampledata type bucket
the filter will be `b.orgID === org.id || b.type==="sampledata"`

ui/src/checks/actions/thunks.ts
```diff
@@ -66,8 +66,11 @@ export const getChecks = () => async (
   getState: GetState
 ) => {
   try {
-    dispatch(setChecks(RemoteDataState.Loading))
-    const {id: orgID} = getOrg(getState())
+    const state = getState()
+    if (getStatus(state, ResourceType.Checks) === RemoteDataState.NotStarted) {
```
drdelambre: what does this pattern mean?

ebb-tide: this is how I solved the getResources bug we saw last week. Now requests for resources that have already been loaded once do not cause an unmount, because they do not set RemoteDataState to Loaded, unless it's the first time.

ui/src/dashboards/components/DashboardContainer.tsx
```diff
@@ -43,9 +44,11 @@ const DashboardContainer: FC<Props> = ({autoRefresh, dashboard, children}) => {
   return (
     <DashboardRoute>
       <GetResource resources={[{type: ResourceType.Dashboards, id: dashboard}]}>
-        <GetTimeRange />
-        <DashboardPage autoRefresh={autoRefresh} />
-        {children}
+        <GetResources resources={[ResourceType.Buckets]}>
```
drdelambre: why isnt this in the component above?

ebb-tide: Get Resource !== GetResources ?

ui/src/shared/components/TimeSeries.tsx
```diff
@@ -189,7 +200,11 @@ class TimeSeries extends Component<Props & WithRouterProps, State> {
       let statuses = [] as StatusRow[][]
       if (check) {
         const extern = buildVarsOption(variables)
-        this.pendingCheckStatuses = runStatusesQuery(orgID, check.id, extern)
+        this.pendingCheckStatuses = runStatusesQuery(
+          this.props.params.orgID,
```
drdelambre: is this supposed to be `orgID` from line 188?

ui/src/dashboards/components/DashboardContainer.tsx
drdelambre: omg that's gross. but ok

ui/src/shared/components/TimeSeries.tsx
ebb-tide: arghh.. soooooooo....... statuses are always in this orgs _monitoring bucket. so, no... But I realize this is very potentially confusing down the line

PR #17208, opened 2020-03-11
fix(ui): fixed yield bug that was caused by race condition

Asalem1:
Closes https://github.com/influxdata/idpe/issues/6240

### Problem

Race condition was leading to a bug where user was receiving an error due to an undefined variable

### Solution

Replaced the `state` with `getState` to get the most up-to-date version of the state, thereby resolving the race condition error


ui/src/timeMachine/actions/queries.ts
```diff
@@ -131,8 +131,11 @@ export const executeQueries = (dashboardID?: string) => async (
     dispatch(setQueryResults(RemoteDataState.Loading, [], null))
 
     await dispatch(refreshTimeMachineVariableValues(dashboardID))
```
Asalem1: this updates the state. So when we we were using `state` beforehand, the `state` we had referenced was changed here, but those changes weren't reflected on the `state` variable

PR #17225, opened 2020-03-12
fix(http): make URM client use userID in filter on find

affo:
Why it doesn't?

### DIFF
--- a/http/document_test.go
+++ b/http/document_test.go
@@ -93,7 +93,7 @@ var (
 		},
 		Content: "content5",
 	}
-	doc6        = influxdb.Document{
+	doc6 = influxdb.Document{
 		ID: doc6ID,
 		Meta: influxdb.DocumentMeta{
 			Name: "doc6",
--- a/http/user_resource_mapping_client_test.go
+++ b/http/user_resource_mapping_client_test.go
@@ -1,0 +1,234 @@
+package http
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/influxdata/httprouter"
+	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/kit/transport/http"
+	"github.com/influxdata/influxdb/mock"
+	"go.uber.org/zap/zaptest"
+	"net/http/httptest"
+)
+
+type OwnerMemberHandler struct {
+	*httprouter.Router
+}
+
+func urmSetup(t *testing.T, rt influxdb.ResourceType, ut influxdb.UserType) (func() *httptest.Server, func(serverUrl string) influxdb.UserResourceMappingService) {
+	idPath := fmt.Sprintf("/api/v2/%v/:id/%vs", rt, ut)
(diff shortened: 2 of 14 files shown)


http/user_resource_mapping_service.go
```diff
@@ -278,13 +278,15 @@ func (s *UserResourceMappingService) FindUserResourceMappings(ctx context.Contex
...
-	urs := make([]*influxdb.UserResourceMapping, len(results.Users))
-	for k, item := range results.Users {
-		urs[k] = &influxdb.UserResourceMapping{
-			ResourceID:   f.ResourceID,
-			ResourceType: f.ResourceType,
-			UserID:       item.User.ID,
-			UserType:     item.Role,
+	urs := make([]*influxdb.UserResourceMapping, 0, len(results.Users))
+	for _, item := range results.Users {
+		if !f.UserID.Valid() || f.UserID != item.User.ID {
```
jsteenb2: where is the test for this?

affo: For the client?

There's no test for the http client I think

affo: I can create one as part of this PR 👍

jsteenb2: That brings up another question, why isn't the server doing the filtering?

http/document_test.go
```diff
@@ -93,7 +93,7 @@ var (
 		},
 		Content: "content5",
 	}
-	doc6        = influxdb.Document{
+	doc6 = influxdb.Document{
```
affo: go fmt

http/user_resource_mapping_client_test.go
```diff
@@ -0,0 +1,234 @@
...
+		log:                        zaptest.NewLogger(t),
+		ResourceType:               rt,
+		UserType:                   ut,
+		UserResourceMappingService: mock.NewUserResourceMappingService(),
+		UserService:                userSvc,
+	}
+	h := OwnerMemberHandler{Router: NewRouter(backend.HTTPErrorHandler)}
+	h.HandlerFunc("POST", idPath, newPostMemberHandler(backend))
+	h.HandlerFunc("GET", idPath, newGetMembersHandler(backend))
+	h.HandlerFunc("GET", userIDPath, newGetMemberHandler(backend))
```
affo: plan to make this a general util to be reused across handlers 👍

http/user_resource_mapping_service.go
```diff
@@ -203,6 +204,105 @@ func decodeGetMembersRequest(ctx context.Context, r *http.Request) (*getMembersR
...
+			}, w)
+			return
+		}
+		m := mappings[0]
+		u, err := b.UserService.FindUserByID(ctx, m.UserID)
+		if err != nil {
+			b.HandleHTTPError(ctx, err, w)
+			return
+		}
+		b.log.Debug("Member/owner retrieved", zap.String("user", fmt.Sprint(u)))
```
jsteenb2: what is the debug message for?

PR #17230, opened 2020-03-12
fix: ensure we're using go v1.13.8

### LINKED ISSUE #17231: Ensure support scripts are updated to use Go 1.13.x before 1.8 influxdb release
At the moment, some of the release engineering scripts still rely on go 1.12.  As we are moving to go 1.13, those support scripts and configuration files must be updated.

ayang64:
Closes #17231

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/releng/_go_versions.sh
+++ b/releng/_go_versions.sh
@@ -1,5 +1,5 @@
 # These are the current and "next" Go versions used to build influxdb.
 # This file is meant to be sourced from other scripts.
 
-export GO_CURRENT_VERSION=1.12.7
-export GO_NEXT_VERSION=1.12.7
+export GO_CURRENT_VERSION=1.13.8
+export GO_NEXT_VERSION=1.13.8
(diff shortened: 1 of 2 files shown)


releng/_go_versions.sh
```diff
@@ -1,5 +1,5 @@
 # These are the current and "next" Go versions used to build influxdb.
 # This file is meant to be sourced from other scripts.
 
-export GO_CURRENT_VERSION=1.12.7
-export GO_NEXT_VERSION=1.12.7
+export GO_CURRENT_VERSION=1.13.8
+export GO_NEXT_VERSION=1.14
```
dgnorton: Let's set both of these to 1.13.8 for now. I'm fuzzy on the history but I think we only set the one ahead for testing.

ayang64: okay.

PR #17237, opened 2020-03-12
fix(ui/src/alerting): reduce points per check from 300 to 100

desa:
This was done to reduce amount of data returned to the client. This is
problematic when there are a large number of series that get returned
from a query. Long term we will want to limit the topN tables so that
this is not an issue.

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/src/alerting/utils/vis.test.ts
+++ b/ui/src/alerting/utils/vis.test.ts
@@ -2,13 +2,13 @@ import {getCheckVisTimeRange} from 'src/alerting/utils/vis'
 
 const duration = 'duration' as 'duration'
 const TESTS = [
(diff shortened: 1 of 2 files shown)


hoorayimhelping:
>this is problematic when there are a large number of series that get returned
from a query

does this affect smaller queries or normal day-to-day operations? is there a way to ease this on for larger queries? like say, if the query has a certain number of results, use the lower value? is that overkill?

desa:
This is only for the number of points per series that we display in graph when building a check. The main idea behind this feature is to give the user a window of data that they can view. If the user picked a `1m` group by, they would see the last `100m` of data. This is definitely not a long term solution and meant as a bandaid for an existing issue. Long term we'll want to limit the number of series that are returned to only, the top N.

PR #17252, opened 2020-03-13
feat(handler): Add 2.0 compatible health endpoint

### LINKED ISSUE #17248: Add support for 2.0 health API to 1.x
__Proposal:__

Make 1.x support the 2.0 health API.

Together with #16898 it will allow users build robust system that will be prepared to smoothly upgrade to 2.

__1.x Ping API:__

https://docs.influxdata.com/influxdb/v1.7/tools/api/#ping-http-endpoint

__2.0 Health API:__

https://v2.docs.influxdata.com/v2.0/api/#tag/Health

__Proposed Changes:__

Add a `health` endpoint to handle the v2.0 server health check protocol.

InfluxDB v2 implementation: https://github.com/influxdata/influxdb/blob/master/http/health.go

bednar:
Closes #17248

Make 1.x support the 2.0 health API.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -985,7 +985,14 @@ func (h *Handler) servePing(w http.ResponseWriter, r *http.Request) {
 
 // serveHealth maps v2 health endpoint to ping endpoint
 func (h *Handler) serveHealth(w http.ResponseWriter, r *http.Request) {
-	b, _ := json.Marshal(map[string]interface{}{"name": "influxdb", "message": "ready for queries and writes", "status": "pass", "checks": []string{}, "version": h.Version})
+	resp := map[string]interface{}{
+		"name":    "influxdb",
+		"message": "ready for queries and writes",
+		"status":  "pass",
+		"checks":  []string{},
+		"version": h.Version,
+	}
+	b, _ := json.Marshal(resp)
 	h.writeHeader(w, http.StatusOK)
 	w.Header().Set("Content-Type", "application/json; charset=utf-8")
 	if _, err := w.Write(b); err != nil {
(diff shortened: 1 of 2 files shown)


services/httpd/handler.go
```diff
@@ -218,6 +218,10 @@ func NewHandler(c Config) *Handler {
 			"status-head",
 			"HEAD", "/status", false, true, authWrapper(h.serveStatus),
 		},
+		Route{ // Ping
+			"ping",
+			"GET", "/health", false, true, authWrapper(h.serveHealth),
```
dgnorton: Is this under `/api/v2/health` in 2.0? If so, should it be the same here? E.g., we added `/api/v2/write` for 1.8.

timhallinflux: https://v2.docs.influxdata.com/v2.0/api/#tag/Health

I believe the feedback is that we need a /health API in order for the 2.0 clients to successfully perform writes?

bednar: The `/health` endpoint is mapped to root path in 2.0. See: https://github.com/influxdata/influxdb/blob/c8aabaf515dccea989c921ac3d2140beede9fbb9/http/handler.go#L22

We need this because a lot of users uses `/ping` and `/health` endpoint to check if is InfluxDB running... and after success they start writing or querying.

sranka:
@dgnorton I cannot see the code in https://github.com/influxdata/influxdb/tree/master-1.x, there is even no `/health` in a full text search. Shouldn't this be already merged therein?

sranka:
There are more (commits)[https://github.com/influxdata/influxdb/compare/master-1.x...1.8] that have not been probably merged into master-1.x yet. I will add the health endpoint together with #18391, it needs fixes. @bednar

PR #17255, opened 2020-03-13
feat(ui): made slight UI changes to the login page as per design requests

Asalem1:
This PR was made to update the design on the login page after the changes made to clockface:

- Made changes from `Login` to `Log In`
- Uploaded and Updated background image to InfluxCloudBG svg
- Updated some of the components to use new Pendulum components
- Removed old styling that was no longer relevant after Pendulum release
- Fixed a bug that was causing `this.auth0` to be undefined by adding the `new` keyword

Screen Shot 2020-03-13 at 08 06 19


ui/src/onboarding/components/LoginForm.tsx
```diff
@@ -87,12 +91,23 @@ export const LoginForm: FC<Props> = ({
           </Grid.Column>
         </Grid.Row>
       </Grid>
-      <a onClick={handleForgotPasswordClick} className="login--forgot-password">
-        Forgot Password?
-      </a>
+      {
+        // @ts-ignore
+        // TODO: remove this ts-ignore when clockface v2.0.2 is released
+        <Heading
```
Asalem1: I added this `@ts-ignore` here since I was getting a type issue when adding an `onClick` event to the `Heading` Component since the functionality doesn't exist on `clockface v2.0.1`. This PR added that functionality onto the Heading component:

https://github.com/influxdata/clockface/pull/449

but a new version of clockface has yet to be released

PR #17260, opened 2020-03-13
feat(cloud): add some ids to the global context

hoorayimhelping:
Closes https://github.com/influxdata/idpe/issues/6298

* in cloud environments, exposes user id, org id, and username to `window.context` object for use in instrumentation

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/shared/actions/me.ts
+++ b/ui/src/shared/actions/me.ts
@@ -1,6 +1,7 @@
-import {MeState} from 'src/shared/reducers/me'
-import {client} from 'src/utils/api'
 import HoneyBadger from 'honeybadger-js'
+import {MeState} from 'src/shared/reducers/me'
+import {CLOUD} from 'src/shared/constants'
+import {client} from 'src/utils/api'
 
 export enum ActionTypes {
   SetMe = 'SET_ME',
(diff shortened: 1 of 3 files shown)


ui/src/shared/utils/getOrgIDFromUrl.test.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const baseUrl = 'http://influx.application/'
+const orgID = 'f440096db5cdb0b1'
+const pathname = `/orgs/${orgID}`
+const location = {
+  ...window.location,
+  href: `${baseUrl}${pathname}`,
+  pathname,
+}
+Object.defineProperty(window, 'location', {
```
drdelambre: why are you over loading a host variable? there be dragons there, and is not recommended outside of w3c ordained polyfills.

drdelambre: oh.. for tests... hmmmmm

ui/src/shared/actions/me.ts
```diff
@@ -26,6 +28,18 @@ export const getMe = () => async dispatch => {
   try {
     const user = await client.users.me()
 
+    if (CLOUD) {
+      const orgID = getOrgIDFromUrl()
```
drdelambre: this seems fragile and couples a view routing convention (that is currently changing with multi user and sample data work) with business logic that is meant to represent "this is my current organization". why isn't it coming back from the api?

hoorayimhelping: yeah this is pretty brittle

drdelambre: so why not add it to the /me endpoint?

hoorayimhelping:
no longer relevant

PR #17264, opened 2020-03-13
fix(buckets): use no auth urm service internally in auth bucket service

gavincabbage:
Changes the authed bucket service to use the no auth URM service to do its internal system bucket authorization. The authed bucket service returns an error when a single malformed URM is returned, which is not the behavior we want here.

Note: The commit message is wrong here, but I'm not fixing it because this fix is for an ongoing issue we want to stop ASAP.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/authorizer/bucket.go
+++ b/authorizer/bucket.go
@@ -100,7 +100,7 @@ func authorizeReadSystemBucket(ctx context.Context, b *influxdb.Bucket, u influx
 
 	return &influxdb.Error{
 		Code: influxdb.EUnauthorized,
-		Msg:  fmt.Sprintf("unauthorized"),
+		Msg:  "unauthorized",
 	}
 }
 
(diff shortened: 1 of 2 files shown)


authorizer/bucket.go
```diff
@@ -89,6 +90,12 @@ func authorizeReadSystemBucket(ctx context.Context, b *influxdb.Bucket, u influx
 		ResourceType: influxdb.OrgsResourceType,
 	})
 	if err != nil {
+		if errors.Is(err, influxdb.ErrOrgNotFound) {
+			return &influxdb.Error{
+				Code: influxdb.EUnauthorized,
+				Msg:  fmt.Sprintf("unauthorized"),
```
brettbuddin: `Sprint` necessary here?

gavincabbage: lol wow no, copy paste got me

authorizer/bucket.go
```diff
@@ -100,7 +100,7 @@ func authorizeReadSystemBucket(ctx context.Context, b *influxdb.Bucket, u influx
 
 	return &influxdb.Error{
 		Code: influxdb.EUnauthorized,
-		Msg:  fmt.Sprintf("unauthorized"),
+		Msg:  "unauthorized",
```
ayang64: This is great!

PR #17286, opened 2020-03-16
refactor(authorizer): auth micro framework

affo:
This PR tries to lay out foundations for an auth mini framework.

The idea is that every resource will use the same methods exposed in `authorize.go`.

`authorize_find.go` was generated and edited by hand to avoid repeating that logic in every service.

As a PoC, the changes have been introduced in `bucket` and `dashboard`.

The whole point of this is to:
 - clean up the code
- make every resource (almost) authorize in the same way

### DIFF
--- a/authorizer/authorize_find.go
+++ b/authorizer/authorize_find.go
@@ -1,0 +1,259 @@
+package authorizer
+
+import (
+	"context"
+
+	"github.com/influxdata/influxdb"
+)
+
+// AuthorizeFindBuckets takes the given items and returns only the ones that the user is authorized to read.
+func AuthorizeFindBuckets(ctx context.Context, rs []*influxdb.Bucket) ([]*influxdb.Bucket, int, error) {
+	// This filters without allocating
+	// https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating
+	rrs := rs[:0]
+	for _, r := range rs {
+		_, _, err := AuthorizeReadBucket(ctx, r.Type, r.ID, r.OrgID)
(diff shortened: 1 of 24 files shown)


authorizer/authorize_find.go
```diff
@@ -0,0 +1,245 @@
+package authorizer
```
gavincabbage: @affo I forgot I looked at this yesterday and it looks good, but my only question was if we could remove some of the duplication in this file... unfortunately because the loops are over slices of different types, it doesn't look like it would be easy 😞

AlirieGray:
Love this consistency for the authorizer layer! One question, will the final version have something like the other authorize_find here for each method? or will they all be in one place (file)?

authorizer/authorize_find.go
affo: Yes, that would be great!

I did generate this using a template and fixing particular cases by hand.

A good thing would be  to have and `IDer` interface that exposes `ID()` and `OrgID()` and make every resource implement that. At that point generation would only be a metter of iterating over resources that implement `IDer` and generate this file.

I can think about do it in the future 👍

affo:
@AlirieGray Thank you!

> One question, will the final version have something like the other authorize_find here for each method? or will they all be in one place (file)?

The idea is to place it under one file. The best would be to generate it. See the comment above https://github.com/influxdata/influxdb/pull/17286#discussion_r394167810

affo:
@gavincabbage @AlirieGray I think this is now ready to go 👍

PR #17290, opened 2020-03-16
feat(pkger): extend buckets to be able to define a bucket and rename it

jsteenb2:
this work is the first of making resources ALL unique by `metadata.name`. The
`spec.name` is a means to rename an existing resource. This is all to support
pkger idempotency. The metadata.name field will be the unique identifier within
a pkg.

with this new `name` field, the errors are now necessary to improve. The nesting of a field that does not pass validation needs to be nested in the specific object it is from `metadata` or `spec`. This PR adds this additional error handling.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influxd/launcher/pkger_test.go
+++ b/cmd/influxd/launcher/pkger_test.go
@@ -155,7 +155,7 @@ func TestLauncher_Pkger(t *testing.T) {
 
 		bkts := sum.Buckets
 		require.Len(t, bkts, 1)
-		assert.Equal(t, "rucket_1", bkts[0].Name)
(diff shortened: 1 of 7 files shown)


kelwang:
Um, our model is more like orgid+resource_type+name to be unique, what's the meta.name gonna be?

jsteenb2:
@kelwang the `metatada.name` is a unique constraint for the pkg itself. It is used (if no display name) as part of the org id + resource type + name combo you mentioned earlier. This does nto affect the domain for buckets.

Issue #17151, opened 2020-03-09
Add Graphic to OSS Signin page


alexpaxton:
![Screen Shot 2020-03-09 at 10.15.31 AM.jpg](https://images.zenhubusercontent.com/5d09754aba016a26c57226e9/2b8a2b9c-2a0f-4344-9b73-9b5a8ff2495e)

Asalem1:
@mnorriswalker - we should setup a time to talk about this

mnorriswalker:
@Asalem1 - great, we have some time setup to pair soon anyway. The plan is just to replace the SVG background image. (Don't worry about the signup content in the modal - that's just inaccurate filler by the illustrator)

[...]

PR #17291, opened 2020-03-16
feat(ui): update oss login page

alexpaxton:
Completes half the work described in #17151 
The other half is contingent upon this issue: https://github.com/influxdata/clockface/issues/455

- Use `FunnelPage` component in the OSS login page
- Use `InfluxDBLogo` from Clockface in login panel
- Factor out `SplashPage` components and styles

Screen Shot 2020-03-16 at 12 06 40 PM

- [x] Rebased/mergeable
- [x] Tests pass


ui/src/onboarding/components/SigninForm.scss
```diff
@@ -1,8 +1,38 @@
-.cf-panel.signin-panel {
+.signin-page {
+  .cf-funnel-page--content {
+    display: flex;
+    justify-content: center;
+    height: 100%;
+  }
+
+  .cf-dapper-scrollbars--content {
+    display: flex !important;
```
hoorayimhelping: is  this `!important` necessary?

ui/src/onboarding/components/SigninForm.scss
```diff
@@ -1,8 +1,38 @@
...
+
+.signin-page--cubo {
+  background-image: url('../../assets/images/auth-logo.svg');
+  background-size: 100% 100%;
+  background-position: center center;
+  background-repeat: no-repeat;
+  width: 100px;
+  height: 100px;
+  display: inline-block;
+}
```
hoorayimhelping: ninor nit: wanna alphabetize these selectors?

ui/src/onboarding/containers/SigninPage.tsx
```diff
@@ -67,18 +68,18 @@ class SigninPage extends PureComponent<Props, State> {
...
       >
-        <SplashPage>
-          <Panel className="signin-panel">
+        <FunnelPage className="signin-page">
+          <Panel className="signin-page--panel">
             <Panel.Body alignItems={AlignItems.Center}>
-              <SplashPage.Logo />
-              <SplashPage.Header title="InfluxData" />
+              <div className="signin-page--cubo" />
+              <InfluxDBCloudLogo cloud={false} className="signin-page--logo" />
```
hoorayimhelping: what is `cloud={false}` we have a `CLOUD` global variable that denotes whether we're in a cloud context or not.

alexpaxton: that prop controls whether the word "Cloud" renders in the image. https://influxdata.github.io/clockface/?path=/story/graphics-brand-logos--influxdb-cloud

In this instance I left it `false` cuz this page is OSS only

ui/src/onboarding/components/SigninForm.scss
alexpaxton: I can do that, though makes me think we could benefit from more thorough CSS formatting. I think prettier supports that? Maybe a nice little tooling improvement

ui/src/onboarding/components/SigninForm.scss
alexpaxton: Unfortunately yes, in clockface there's an `!important` present to override styles from a library, so the madness trickles down

alexpaxton:
@hoorayimhelping just pushed a commit that sorts the css attributes alphabetically

ui/src/onboarding/components/SigninForm.scss
hoorayimhelping: @alexpaxton deffo. I usually just say "alphabetize absent any other organizing scheme. would be rad to have one in place"

PR #17295, opened 2020-03-16
test(query/stdlib/influxdata/influxdb): update rules_test to use algo-w

wolffcm:
This updates our rules test to create function expressions using the Rust parser and analysis, so that we are testing our rules against the kinds of things we'll actually see in production.

### DIFF
--- a/query/stdlib/influxdata/influxdb/rules_test.go
+++ b/query/stdlib/influxdata/influxdb/rules_test.go
@@ -155,6 +155,26 @@ func TestPushDownFilterRule(t *testing.T) {
 		}
 	}
 
+	makeExprBody := func(expr *semantic.FunctionExpression) *semantic.FunctionExpression {
+		switch e := expr.Block.Body.(type) {
+		case *semantic.Block:
+			if len(e.Body) != 1 {
+				panic("more than one statement in function body")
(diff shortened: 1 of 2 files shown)


query/stdlib/influxdata/influxdb/rules_test.go
```diff
@@ -657,7 +491,6 @@ func TestPushDownFilterRule(t *testing.T) {
 	for _, tc := range tests {
 		tc := tc
 		t.Run(tc.Name, func(t *testing.T) {
-			t.Parallel()
```
jlapacik: Making tests run serially isn't critical to tests passing right?

wolffcm: I think what was happening was that we were re-using function expressions in different cases, and they were getting transformed in-place. Unless we wrap those exprs in closures, the test will have issues.

wolffcm: I guess what I'm saying is that I'm okay with not letting them run in parallel, and leaving the test as-is.

jlapacik: That works for me.

PR #17296, opened 2020-03-16
feat(tenant): Add in service functions and business logic

lyondhill:
Built on top of the crud layer of the system we should now have additional service logic.
The addition of service layer should allow for us to verify similar functionality to the kv system.

### DIFF
--- a/tenant/error.go
+++ b/tenant/error.go
@@ -10,6 +10,20 @@ var (
 		Code: influxdb.EInvalid,
 		Msg:  "name is empty",
 	}
+
+	// NotUniqueIDError is used when attempting to create an org or bucket that already
+	// exists.
+	NotUniqueIDError = &influxdb.Error{
+		Code: influxdb.EConflict,
+		Msg:  "ID already exists",
+	}
+
+	// ErrFailureGeneratingID occurs ony when the random number generator
+	// cannot generate an ID in MaxIDGenerationN times.
+	ErrFailureGeneratingID = &influxdb.Error{
+		Code: influxdb.EInternal,
+		Msg:  "unable to generate valid id",
+	}
 )
 
 // ErrCorruptID the ID stored in the Store is corrupt.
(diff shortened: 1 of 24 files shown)


tenant/service_urm_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	for _, m := range f.UserResourceMappings {
+		if err := svc.CreateUserResourceMapping(context.Background(), m); err != nil {
+			t.Fatalf("failed to populate mappings")
+		}
+	}
+
+	return svc, func() {
+		for _, m := range f.UserResourceMappings {
+			if err := svc.DeleteUserResourceMapping(context.Background(), m.ResourceID, m.UserID); err != nil {
```
AlirieGray: I notice you mostly have tests just for create and delete, is that because find/update is covered by other tests?

tenant/error.go
```diff
@@ -10,6 +10,20 @@ var (
...
+	// NotUniqueIDError is used when attempting to create an org or bucket that already
+	// exists.
+	NotUniqueIDError = &influxdb.Error{
+		Code: influxdb.EConflict,
+		Msg:  "ID already exists",
+	}
+
+	// ErrFailureGeneratingID occurs ony when the random number generator
+	// cannot generate an ID in MaxIDGenerationN times.
+	ErrFailureGeneratingID = &influxdb.Error{
```
AlirieGray: Love all these detailed errors in easy to find places!

tenant/service_urm_test.go
lyondhill: I actually didn't write any tests for these systems. Instead I am relying on the `/testing` packages test suites to make sure that these new service definitions act like the kv system (which is being tested by the same testing suite). I had to make a few updates to the testing framework because I wanted to remove the direct access some of the test required on the service implementer. But in most cases the tests are all in the testing package.

PR #17301, opened 2020-03-16
feat(metrics): add metrics middleware to tenant service

AlirieGray:
This PR adds a metrics middleware wrapping the services that comprise the Tenant Service: User, URM, Bucket, and Org. 

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/tenant/middleware_bucket_metrics.go
+++ b/tenant/middleware_bucket_metrics.go
@@ -1,0 +1,75 @@
+package tenant
+
+import (
+	"context"
+
+	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/kit/metric"
+	"github.com/influxdata/influxdb/kit/prom"
+)
+
+type BucketMetrics struct {
+	// RED metrics
+	rec *metric.REDClient
+
+	bucketService influxdb.BucketService
+}
+
+var _ influxdb.BucketService = (*BucketMetrics)(nil)
+
+// NewBucketMetrics returns a metrics service middleware for the Bucket Service.
+func NewBucketMetrics(reg *prom.Registry, s influxdb.BucketService) *BucketMetrics {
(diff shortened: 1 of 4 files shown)


bucket.go
```diff
@@ -98,6 +98,9 @@ type BucketService interface {
 	FindBucketByName(ctx context.Context, orgID ID, name string) (*Bucket, error)
 }
 
+// BucketMiddleware is a service middleware func.
+type BucketMiddleware func(BucketService) BucketService
```
lyondhill: Usually any middleware definition is concrete. The reason is because most middleware don't share a common signature.

cmd/influxd/launcher/launcher.go
```diff
@@ -516,11 +517,11 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 	m.reg.MustRegister(m.boltClient)
 
 	var (
-		orgSvc                    platform.OrganizationService             = m.kvService
+		orgSvc                    platform.OrganizationService             = tenant.NewOrgMetrics(m.reg)(m.kvService)
```
lyondhill: Let's not plug in any if the tenant stuff yet.

tenant/middleware_bucket_metrics.go
```diff
@@ -0,0 +1,77 @@
...
+	// RED metrics
+	rec *metric.REDClient
+
+	bucketService influxdb.BucketService
+}
+
+var _ influxdb.BucketService = (*bucketMetrics)(nil)
+
+// NewBucketMetrics returns a metrics service middleware for the Bucket Service.
+func NewBucketMetrics(reg *prom.Registry) influxdb.BucketMiddleware {
```
lyondhill: By using a middleware generic function the call here can get confusing 'NewBucketMetrics(reg).(svc)'

PR #17302, opened 2020-03-16
feat(cloud): add userid and org id to gtm

hoorayimhelping:
Closes https://github.com/influxdata/growth/issues/41

(Private repo):
#### add dataLayer indentifiers to idpe

* adds a couple of helper methods to initialize the user in google tag manager
* calls those methods when user data and organization data are loaded

### Testing Instructions
* fire up app
* open your console and look for `window.dataLayer`. it shouldn't exist
* without a lot of config to make cloud work locally, change [this](https://github.com/influxdata/influxdb/pull/17302/files#diff-8c33463caf3e87edc906ec5715cb6eabR68) and [this](https://github.com/influxdata/influxdb/pull/17302/files#diff-023642cbf29399507646a13659e14d24R31) line to say `if (CLOUD || true)`
* open up your console and look for `window.dataLayer` it should be there.
  * if you prefer to just see an object with values, you can do something cutesy like:
    ```ts
      console.log(dataLayer.reduce((identity, item) => {
       return {...identity, ...item.identity}
      }, {}))
    ```

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/shared/utils/analytics.ts
+++ b/ui/src/shared/utils/analytics.ts
@@ -1,0 +1,25 @@
+export const fireOrgIdReady = function fireOrgIdReady(
+  organizationIds: string[]
+) {
+  window.dataLayer = window.dataLayer || []
+  window.dataLayer.push({
+    event: 'cloudAppOrgIdReady',
+    identity: {
+      organizationIds,
+    },
+  })
+}
+
+export const fireUserDataReady = function fireUserDataReady(
+  id: string,
+  email: string
+) {
+  window.dataLayer = window.dataLayer || []
+  window.dataLayer.push({
+    event: 'cloudAppUserDataReady',
+    identity: {
+      email,
+      id,
+    },
+  })
+}
--- a/ui/src/index.tsx
+++ b/ui/src/index.tsx
@@ -150,6 +150,7 @@ const basepath = getBrowserBasepath()
 declare global {
   interface Window {
     basepath: string
+    dataLayer: any[]
   }
 }
 
--- a/ui/src/organizations/actions/thunks.ts
+++ b/ui/src/organizations/actions/thunks.ts
@@ -18,6 +18,8 @@ import {
 } from 'src/organizations/actions/creators'
 
 // Constants
+import {CLOUD} from 'src/shared/constants'
+
 import {
   orgCreateSuccess,
   orgCreateFailed,
@@ -28,6 +30,8 @@ import {
   orgRenameSuccess,
   orgRenameFailed,
 } from 'src/shared/copy/notifications'
+
+import {fireOrgIdReady} from 'src/shared/utils/analytics'
 
 // Schemas
 import {orgSchema, arrayOfOrgs} from 'src/schemas'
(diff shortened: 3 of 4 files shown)


ui/src/shared/utils/analytics.ts
```diff
@@ -0,0 +1,25 @@
...
+) {
+  window.dataLayer = window.dataLayer || []
+  window.dataLayer.push({
+    event: 'cloupAppUserDataReady',
+    identity: {
+      email,
+      id,
+    },
+  })
+}
```
hoorayimhelping: these two methods are somewhat special. the idea is that eventually we'll have a simple interface for firing arbitrary events that abstracts the implementation (currently it's google tag manager). right now, it's just these two methods, but i could see something like:

```ts
export const fireEvent = function fireEvent(eventName: string, contextData: any) {
  // implementation
}
...
componentDidMount() {
  fireEvent('data-explorer-loaded')
}
```

becoming more commonplace

ui/src/shared/utils/analytics.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const fireUserDataReady = function fireUserDataReady(
+  id: string,
+  email: string
+) {
+  window.dataLayer = window.dataLayer || []
+  window.dataLayer.push({
+    event: 'cloupAppUserDataReady',
```
Asalem1: Is this meant to be `cloup` or `cloud`?

hoorayimhelping: cloud heh. good catch

ui/src/shared/utils/analytics.ts
```diff
@@ -0,0 +1,25 @@
...
+  window.dataLayer = window.dataLayer || []
+  window.dataLayer.push({
+    event: 'cloudAppOrgIdReady',
+    identity: {
+      organizationIds,
+    },
+  })
+}
+
+export const fireUserDataReady = function fireUserDataReady(
```
Asalem1: Really interesting setting a const to a function this way. I am curious as to why the parameter is being passed a `user.name` but is expecting an `email`. Is the parameter the source of truth here (in that the `user.name` is an email)?

hoorayimhelping: >Really interesting setting a const to a function this way. 

@Asalem1 it's yavascript learnin' time with uncle bucko

```ts
const foo = function() {
  console.log('anonymous')
}
```
creates an anonymous function and assigns it to the variable `foo`

```ts
const bar = function fineYoungCannibals() {
   console.log('fineYoungCannibals')
}
```
creates a function named `fineYoungCannibals` and assigns it to the variable `bar`. On stack traces, this name will show up, rather than nothing. Check [the first example](https://hackernoon.com/three-reasons-i-avoid-anonymous-js-functions-like-the-plague-7f985c27a006) in this article to see what i mean.

I named the variable `bar` and the function `fineYoungCannibals` to show that the naming of a function and the assigning of a function to a variable are separate things that in the instance above, happen to share the same name.

---

to answer your question re: `user.name` and `email.`

`user.name` seems to represent username, which is what is entered into the form. it doesn't need to be an email (e.g. locally, my login is `bucky` | `password` [plz don't hax0r me]), but for our intents, it -should- be email. (this is confusingly named, not sure about the context behind it).

PR #17309, opened 2020-03-17
test: add tests for tenant service

affo:

### DIFF
--- a/kv/tenant_test.go
+++ b/kv/tenant_test.go
@@ -1,0 +1,99 @@
+package kv_test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/influxdata/influxdb"
+	icontext "github.com/influxdata/influxdb/context"
+	"github.com/influxdata/influxdb/kv"
+	itesting "github.com/influxdata/influxdb/testing"
+	"go.uber.org/zap/zaptest"
+)
+
+func TestBoltTenantService(t *testing.T) {
+	t.Skip("some tests fail using `*kv.Service` as `influxdb.TenantService`, see the NOTEs in `testing/tenant.go`.")
+	itesting.TenantService(t, initBoltTenantService)
+}
+
+func initBoltTenantService(t *testing.T, f itesting.TenantFields) (influxdb.TenantService, func()) {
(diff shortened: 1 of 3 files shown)


testing/tenant.go
```diff
@@ -0,0 +1,100 @@
...
+		{
+			name: "CascadingDelete",
+			fn:   CascadingDelete,
+		},
+		// TODO(affo): add tests.
+		//  - cascading delete urm: for every resource, if you delete it you delete the urm.
+		//  - cascading delete user: if you delete a user you delete every associated resource.
+		//    This should happen only  when that  user is the only member/owner of that resource.
+		//  - cascading delete organization: if you delete an org all its content gets dropped.
+		//    NOTE: do we need a tombstone and an expiry period?
```
affo: some tests ideas

kv/tenant_test.go
```diff
@@ -0,0 +1,79 @@
...
+			t.Fatalf("error populating dashboards: %v", err)
+		}
+	}
+	for _, urm := range f.UserResourceMappings {
+		if err := svc.CreateUserResourceMapping(ctx, urm); err != nil {
+			t.Fatalf("error populating urms: %v", err)
+		}
+	}
+
+	return tenantService{
```
lyondhill: You could probably just return svc here if you wanted.

affo:
@lyondhill I think this is now ready to go 👍

kv/tenant_test.go
```diff
@@ -0,0 +1,86 @@
...
+	itesting "github.com/influxdata/influxdb/testing"
+	"go.uber.org/zap/zaptest"
+)
+
+func TestBoltTenantService(t *testing.T) {
+	t.Skip("some tests fail using `*kv.Service` as `influxdb.TenantService`, see the NOTEs in `testing/tenant.go`.")
+	itesting.TenantService(t, initBoltTenantService)
+}
+
+func TestBoltNewTenantService(t *testing.T) {
```
lyondhill: this function should be in the tenant/tenant_test.go

affo: yes, you are right 😓

PR #17310, opened 2020-03-17
feat(ui): parsing domain on the UI to keep particular logic encompassed

Asalem1:
Implementing a UI-based solution in favor of an API solution (https://github.com/influxdata/idpe/pull/6392) due to the fact that we want to keep logic that is particular to the UI in the UI. That is to say, that since the UI expects the domain to be constructed in a particular way, we want the related logic to be conveniently located near the code where it is being used rather than abstracting it to the API and making it hard to understand both on the API and the UI.

### DIFF
--- a/ui/src/Logout.tsx
+++ b/ui/src/Logout.tsx
@@ -15,13 +15,15 @@ import {ErrorHandling} from 'src/shared/decorators/errors'
 
 // Utils
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+import {parseUrl} from 'src/shared/utils/parseUrl'
 
 const Logout: FC<WithRouterProps> = ({router}) => {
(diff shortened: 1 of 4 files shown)


Asalem1:
@drdelambre not really, since we're using the `Domain` attribute in the API for other purposes, so we don't want to have to update the Domain and have some unintended consequences since we're actually expecting the Domain to be formatted a specific way in IDPE:

https://github.com/influxdata/idpe/blob/master/cmd/gatewayd/main.go#L1000-L1002

Asalem1:
Going to close this in favor of the previous implementation. While a better long-term solution would be to inject a domain without a protocol and prepend that wherever the OIDCConfig.Domain is being used, we would want to simply prepend that domain with an `https`. However, implementing this right now might cause some unintended consequences

PR #17336, opened 2020-03-18
fix: set bucket limit to 100

### LINKED ISSUE #17068: Buckets page cannot show more than 20 user buckets
Once a user creates more than 20 buckets, they can no longer access them on the buckets page. There is no way to search for them either since the search only applies to the 20 buckets returned.

We need to figure out how to page resources in the app since the APIs usually only return 20.

121watts:
Closes #17068 

### The Problem
Users with more than 20 buckets could not see all their buckets

### The (Temporary) Solution
Increase the `limit` query parameter to 100.  

### The Real Solution
Implement an app wide pagination / filtering / infiniscroll solution for all resources.

### DIFF
--- a/ui/src/buckets/actions/thunks.ts
+++ b/ui/src/buckets/actions/thunks.ts
@@ -62,7 +62,7 @@ export const getBuckets = () => async (
     }
     const org = getOrg(state)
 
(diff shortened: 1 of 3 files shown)


121watts:
@hoorayimhelping let's do that

Issue #17233, opened 2020-03-12
Pkger resources are unique by name in pkg


jsteenb2:
We want to make all resource names in a pkg unique by name. We will allow users to change the user facing name/vanity name to be different from the resource name identifier.  Names are unique per resource type. The following is desired for enabling idempotency in support of gitops:

```yml
apiVersion: influxdata.com/v2alpha1
kind: Label
metadata:
  name: label_1 #vanity name defaults to label_1
---
apiVersion: influxdata.com/v2alpha1
kind: Bucket
metadata:
  name: rucket_3 # is the reference name
spec:
  name: new vanity name # will be the vanity name
```

AC
- [ ] resources that provide a `spec.name` override the `metadata.name` field for the resource name
- [ ] when no `spec.name` provided, falls back to use the `metadata.name` field
- [ ] when exporting resources of the same name and type, the `metadata.name` will be appended by a suffix that makes it unique

PR #17337, opened 2020-03-18
chore(pkger): refactors export mechanics to provide unique metadata.n…

jsteenb2:
references #17233 

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influx/pkg_test.go
+++ b/cmd/influx/pkg_test.go
@@ -202,7 +202,7 @@ func TestCmdPkg(t *testing.T) {
 					for _, labelName := range orgIDOpt.LabelNames {
 						pkg.Objects = append(pkg.Objects, pkger.Object{
 							APIVersion: pkger.APIVersion,
-							Type:       pkger.KindLabel,
+							Kind:       pkger.KindLabel,
 							Metadata:   pkger.Resource{"name": labelName},
 						})
 					}
@@ -210,7 +210,7 @@ func TestCmdPkg(t *testing.T) {
 						for _, k := range orgIDOpt.ResourceKinds {
 							pkg.Objects = append(pkg.Objects, pkger.Object{
 								APIVersion: pkger.APIVersion,
-								Type:       k,
+								Kind:       k,
 								Metadata: pkger.Resource{
 									"name": k.String(),
 								},
@@ -222,7 +222,7 @@ func TestCmdPkg(t *testing.T) {
 
 					pkg.Objects = append(pkg.Objects, pkger.Object{
 						APIVersion: pkger.APIVersion,
-						Type:       pkger.KindBucket,
+						Kind:       pkger.KindBucket,
 						Metadata:   pkger.Resource{"name": "bucket1"},
 					})
 					return &pkg, nil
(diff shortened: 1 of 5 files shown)


cmd/influx/pkg_test.go
```diff
@@ -202,15 +202,15 @@ func TestCmdPkg(t *testing.T) {
 					for _, labelName := range orgIDOpt.LabelNames {
 						pkg.Objects = append(pkg.Objects, pkger.Object{
 							APIVersion: pkger.APIVersion,
-							Type:       pkger.KindLabel,
+							Kind:       pkger.KindLabel,
```
kelwang: will these break any existing ones?

jsteenb2: should have no impact on external consumers, the `json` field is still the same, as is the pkg format. The Type part is wrong, just forgot to fix it until now

pkger/clone_resource.go
```diff
@@ -31,6 +37,300 @@ func (r ResourceToClone) OK() error {
...
+		if iKind.is(jKind) {
+			return iName < jName
+		}
+		return kindPriorities[iKind] < kindPriorities[jKind]
+	})
+
+	for _, r := range resourcesToClone {
+		err := ex.resourceCloneToKind(ctx, r, cloneAssFn)
+		if err != nil {
+			return internalErr(fmt.Errorf("failed to clone resource: resource_id=%s resource_kind=%s err=%q", r.ID, r.Kind, err))
```
kelwang: Internal Err is the default code, you can just use fmt.Errorf

jsteenb2: I want to be explicit about using an `influxdb.Error` here. There is associated behavior with that type. That behavior I want the service to guarantee it handles so that consumers of this service have nothing but a error code check to signal how they will handle it.

pkger/clone_resource.go
```diff
@@ -31,6 +37,300 @@ func (r ResourceToClone) OK() error {
...
+
+func (ex *exportCoordinator) Export(ctx context.Context, resourcesToClone []ResourceToClone, labelNames ...string) error {
+	cloneAssFn, err := ex.resourceCloneAssociationsGen(ctx, labelNames...)
+	if err != nil {
+		return err
+	}
+
+	resourcesToClone = uniqResourcesToClone(resourcesToClone)
+	sort.Slice(resourcesToClone, func(i, j int) bool {
+		iName, jName := resourcesToClone[i].Name, resourcesToClone[j].Name
```
jsteenb2: this order here, provides guarantees that dependencies for a resource (associations) are accounted for. I'll add a comment to make that obvious.

pkger/clone_resource.go
```diff
@@ -31,6 +37,300 @@ func (r ResourceToClone) OK() error {
...
+	}
+	if skipResource {
+		return nil
+	}
+
+	mapResource := func(orgID influxdb.ID, k Kind, object Object) {
+		if len(ass) > 0 {
+			object.Spec[fieldAssociations] = ass
+		}
+		key := newExportKey(orgID, k, object.Spec.stringShort(fieldName))
```
kelwang: quick question, where does the `fieldName` come from?

jsteenb2: `fieldName` is a constant that couples together the decoding and encoding fields. You can find it here in the `models.go` file 👍 

https://github.com/influxdata/influxdb/blob/52c7447333c9944e88d71b307770ce9cb1db9b2c/pkger/models.go#L777

PR #17338, opened 2020-03-18
feat(tsdb): Bulk delete series performance improvement

benbjohnson:
## Overview

This pull request significantly improves the performance of deleting with predicate. Local tests show a delete of 1m points going from 20m+ to subsecond. The primary change is batching `Sync()` calls when deleting series & tag values from the index and deleting series from the series file.

### DIFF
--- a/tsdb/tsm1/engine_delete_prefix.go
+++ b/tsdb/tsm1/engine_delete_prefix.go
@@ -12,6 +12,7 @@ import (
 	"github.com/influxdata/influxdb/kit/tracing"
 	"github.com/influxdata/influxdb/models"
 	"github.com/influxdata/influxdb/tsdb"
+	"github.com/influxdata/influxdb/tsdb/tsi1"
 	"github.com/influxdata/influxql"
 )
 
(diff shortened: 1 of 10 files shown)


tsdb/tsm1/engine_delete_prefix.go
```diff
@@ -245,34 +246,53 @@ func (e *Engine) DeletePrefixRange(rootCtx context.Context, name []byte, min, ma
...
-			keyb := []byte(key)
-			keyb, _ = SeriesAndFieldFromCompositeKey(keyb)
+			possiblyDeadKeysSlice = append(possiblyDeadKeysSlice, []byte(key))
+		}
 
-			name, tags := models.ParseKeyBytes(keyb)
-			sid := e.sfile.SeriesID(name, tags, buf)
-			if sid.IsZero() {
-				continue
+		const batchSize = 1000
```
jacobmarble: This is roughly 125 keys per TSI/sfile partition per batch. Any performance difference between 1,000 and 10,000?

benbjohnson: I'm not sure. The performance increase is already several orders of magnitude faster and the trade off would be using more memory. I'm not sure it's worth increasing without an example of slow deletes (after this change).

PR #17340, opened 2020-03-18
fix(documents): prevent adding duplicate labels to documents

### LINKED ISSUE #17092: Label service allows to create same label mapping twice
See documents tests `http/document_service_test.go`:

https://github.com/influxdata/influxdb/blob/9defb8a77f0b4c0f8f6b879cd19a5c7afb6b3485/http/document_service_test.go#L609

AlirieGray:
Closes #17092

This PR adds a check for existing labels on a Document before adding a Label Mapping, to prevent duplicate Label Mappings from being created for a given Document.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/document_service_test.go
+++ b/http/document_service_test.go
@@ -647,16 +647,19 @@ func AddLabels(t *testing.T) {
 	})
 
 	t.Run("add same twice", func(t *testing.T) {
-		// TODO(affo)
-		t.Skip("see https://github.com/influxdata/influxdb/issues/17092")
-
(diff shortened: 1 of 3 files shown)


AlirieGray:
@affo You're right, it could be an issue in other services, I can look into it and add that to this PR.

I considered putting this logic in the label service, but the function `CreateLabelMapping` only gets an ID for the resource, not the resource itself. So I think I'd have to switch on the type, then do a different lookup for each resource.

PR #17350, opened 2020-03-19
fix(ui): treating headers with an error the same as any other value

### LINKED ISSUE #16474: Data with column error should not result in UI error screen
error-column

This happens as we use the simply check for a column called `error` on the csv returned.

We should check for the exact schema returned when there is a runtime error and only display the error screen on that condition. See https://github.com/influxdata/flux/blob/master/docs/SPEC.md#errors

Asalem1:
Closes #16474 

After speaking with @russorat and checking out the previous issue regarding a similar matter (https://app.zenhub.com/workspaces/applicationmonitoring-5e3b05772922e316b3a210a4/issues/influxdata/idpe/5152) it seems clear that the intent of the issue is to treat `error` headers the same as all other columns. As such, this PR removes the logic that dealt with throwing an error whenever a `header` returned an `error`. Below is an example of how this looks:

Screen Shot 2020-03-19 at 08 26 21

### DIFF
--- a/ui/src/shared/utils/checkQueryResult.ts
+++ b/ui/src/shared/utils/checkQueryResult.ts
@@ -1,7 +1,7 @@
 /*
   Given Flux query response as a CSV, check if the CSV contains an error table
   as the first result. If it does, throw the error message contained within
-  that table. 
+  that table.
 
   For example, given the following response:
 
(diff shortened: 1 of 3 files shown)


russorat:
let me validate this in the branch before we merge.

Asalem1:
After speaking with Desa about this PR, there's another change that we want to tackle that hasn't been mentioned in the ticket. This is not ready for merging at this time

ui/src/shared/utils/checkQueryResult.ts
```diff
@@ -24,18 +24,7 @@ export const checkQueryResult = (file: string): void => {
...
-
-  if (index === -1 || !row[index]) {
-    return
-  }
-
-  // Trim off extra quotes at start and end of message
-  const errorMessage = row[index].replace(/^"/, '').replace(/"$/, '')
-
-  throw new Error(errorMessage)
+  return
```
drdelambre, line 27: the error is actually with the assumption made on line 18 (the error could be anywhere), and with the means of asserting an error type in line 29-36. the code should read more like:
```
if (!lines[0].test(/^,error,reference$/)) {
    return
}

const row = lines[1].split(',').map(s => s.trim())

throw new Error(`[${row[2]}] ${row[1]}`)
```

Asalem1:
I'm gonna close this PR out since it's incomplete and I won't be able to work on it until we are code complete on a few other things

PR #17353, opened 2020-03-19
feat(pkger): extend tasks with ability to define and rename it

jsteenb2:
closes: #17233

- [x] Rebased/mergeable
- [ ] Tests pass


pkger/models.go
```diff
@@ -1897,6 +1898,13 @@ func (t *task) Labels() []*label {
 }
 
 func (t *task) Name() string {
+	if displayName := t.displayName.String(); displayName != "" {
```
kelwang: we should make an abstract struct for this.

PR #17360, opened 2020-03-19
chore: cleaning variables [WIP]

drdelambre:
there were too many places where variables could get updated, so changing behavior in one location would break the feature in another location.
consolidating variables to a "getVariables" and "getAllVariables", which derive their value from state and context (optional) gets around that by ensuring that all data is a simple functional transform away from the source of truth.
This resulted in the ability to delete a lot of the code that was transferring state back and forth between formats and locations


ui/cypress/e2e/dashboardsView.test.ts
```diff
@@ -87,9 +87,35 @@ describe('Dashboard', () => {
   })
 
   const getSelectedVariable = (contextID: string, index?: number) => win => {
-    const {resources} = win.store.getState() as AppState
-    const vardawg = resources.variables.values[contextID].values
-    return vardawg[Object.keys(vardawg)[index]].selectedValue
+    const state = win.store.getState() as AppState
```
drdelambre: couldn't load the selector here directly, so this serves as a mock

ui/cypress/e2e/dashboardsView.test.ts
```diff
@@ -151,7 +178,7 @@ describe('Dashboard', () => {
 
           // selected value in cell context is 2nd value
           cy.window()
-            .pipe(getSelectedVariable('veo', 0))
+            .pipe(getSelectedVariable(dashboard.id, 0))
```
drdelambre: the concept of `veo` doesn't really exist any more, instead timemachine is loaded with a contextID (which is just the dashboardID) and loads it's truth from that instead

drdelambre:
NOTE: `asAssignment` is a concept that needs to be killed in this codebase as well, the whole `[jsObj] => astObj => AST => jsObj => var` thing requires too much context to be carried around just to grab the `windowPeriod` occasionally. but this PR was getting too long.

ui/src/variables/selectors/index.tsx
```diff
@@ -247,60 +264,9 @@ export const getTimeMachineValuesStatus = (
...
-): Variable[] => {
-  const hydratedVariableIDs: string[] = Object.keys(
-    get(state, `resources.variables.values.${contextID}.values`, {})
-  )
-
-  const hydratedVariables = Object.values(
-    state.resources.variables.byID
-  ).filter(v => hydratedVariableIDs.includes(v.id))
-
-  return hydratedVariables
```
121watts: Yaaay!

drdelambre:
jumping into jest! stand by

121watts:
looks like your last commit is causing semantic commit to be mad at you.  probably need a space after the `fix:`

Issue #17235, opened 2020-03-12
Allow users to create a stack for managing pkg applications


jsteenb2:
We want to introduce the concept of a stack, which is a container for which to apply packages that maintain state of their application. 

<details><summary>A stack record will look like the following:</summary>

```json
{
  "stack_id": "UUID",
  "created_at": "TIME STAMP",
  "updated_at": "TIME STAMP",
  "config": {
    "urls": ["URL_TO_ORG_PKG", "URL_TO_SHARED_K8s_PKG"]
  },
  "resources": [
    {
      "apiVersion": "v2alpha1",
      "kind": "Bucket",
      "name": "bucket_1",
      "id": "UUID"
    },
    {
      "apiVersion": "v2alpha1",
      "kind": "Label",
      "name": "label_1",
      "id": "UUID"
    },
    {
      "apiVersion": "v2alpha1",
      "kind": "Dashboard",
      "name": "dash_1",
      "id": "UUID"
    },
    {
      "apiVersion": "v2alpha1",
      "kind": "Dashboard",
      "name": "dash_2",
      "id": "UUID"
    },
    {
      "apiVersion": "v2alpha1",
      "kind": "Task",
      "name": "task_1",
      "id": "UUID"
    }
   ...
  ]
}
```

</details> 

The stack record consists of all resources that exist as a result of running the pkg.  This issue is to create the foundation for which pkger will use this stack to manage state of its application of a pkg.  

A new API endpoint is required

`POST /api/v2/packages/stack`

with the following req body (optional):

```json
{
   urls: ["URL_TO_PKG_1", "URL_TO_PKG_2", ...]
}
```

with the following resp body:

```json
{
   "stackID":  "UUID",
   "createdAt": "TIMESTAMP",
   "updatedAt": "TIMESTAMP",
   "config": {
      urls: [...valid urls or empty collection]
   }
}
```

AC
- [x] new endpoint `POST /api/v2/packages/stack` to initialize the stack
  * involves a lot of plumbing
- [x] stack record includes the following when created:

```json
{
"stack_id": "UUID",
  "created_at": "TIME STAMP",
  "updated_at": "TIME STAMP",
  "config": {
    // if provided
    "urls": ["URL_TO_ORG_PKG"] 
  },
  "resources": []
}
```

PR #17365, opened 2020-03-20
feat(pkger): add foundation for pkger stack functionality

jsteenb2:
references #17235 

This is the foundation of the new stack implementation for pkger. The CRUD store, the service stub, and the middleware are all coming together. Wanted to get this a review before that, b/c the size of the review would be pretty big if it was one big PR.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/pkger/service.go
+++ b/pkger/service.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/url"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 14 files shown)


pkger/service.go
```diff
@@ -17,8 +18,26 @@ import (
...
+		APIVersion string
+		ID         influxdb.ID
+		Kind       Kind
+		Name       string
+	}
+)
+
 // SVC is the packages service interface.
 type SVC interface {
+	InitStack(ctx context.Context, orgID, userID influxdb.ID, urls ...url.URL) (Stack, error)
```
kelwang: Could you add some comments for all the exported funcs?

pkger/service.go
```diff
@@ -17,8 +18,26 @@ import (
 // APIVersion marks the current APIVersion for influx packages.
 const APIVersion = "influxdata.com/v2alpha1"
 
+type (
+	Stack struct {
+		ID        influxdb.ID
+		CreatedAt time.Time
```
kelwang: could you use influxdb. CRUDLog here?

PR #17378, opened 2020-03-20
feat: ui lsp integration

bthesorceror:
- Code folding
- Document Symbols
- Rename
- References
- Go to definition


ui/src/external/monaco.flux.completions.ts
```diff
@@ -13,6 +13,83 @@ const m2p = new MonacoToProtocolConverter(),
...
+    provideReferences: async (model, position, context, _token) => {
+      const uri = model.uri.toString()
+      const pos = m2p.asPosition(position.lineNumber, position.column)
+
+      try {
+        const locs = await server.references(uri, pos, context)
+
+        return (locs || []).map(loc => p2m.asLocation(loc))
+      } catch (e) {
+        // console.log(e)
```
drdelambre: why keep the commented code?

ui/src/external/monaco.flux.completions.ts
```diff
@@ -13,6 +13,83 @@ const m2p = new MonacoToProtocolConverter(),
...
+
+  monaco.languages.registerDefinitionProvider(FLUXLANGID, {
+    provideDefinition: async (model, position, _token) => {
+      const uri = model.uri.toString()
+      const pos = m2p.asPosition(position.lineNumber, position.column)
+
+      try {
+        const loc = await server.definition(uri, pos)
+
+        return p2m.asLocation(loc)
```
drdelambre: why doesn't this need the same guard as line 68?

bthesorceror: asLocation takes an optional type so it should handle a null value, whereas 68 we are calling map on a value that could be null.

ui/src/external/monaco.flux.completions.ts
```diff
@@ -13,6 +13,83 @@ const m2p = new MonacoToProtocolConverter(),
   p2m = new ProtocolToMonacoConverter()
 
 export function registerCompletion(monaco: MonacoType, server: LSPServer) {
+  monaco.languages.registerDocumentSymbolProvider(FLUXLANGID, {
```
drdelambre: does it make sense to keep all of these in the "completions" file? not arguing for a different file for each functionality change, just seems confusing that i'd look here to find code folding providers

bthesorceror: yeah I meant to rename that file to monaco.flux.lsp.ts instead but forgot 👍

ui/src/external/monaco.flux.server.ts
```diff
@@ -60,6 +69,46 @@ export class LSPServer {
...
+  async rename(uri, position, newName): Promise<WorkspaceEdit> {
+    const response = (await this.send(
+      rename(this.currentMessageID, uri, position, newName)
+    )) as {result: WorkspaceEdit}
+
+    return response.result
+  }
+
+  async definition(uri, position): Promise<Location> {
+    const response = (await this.send(
```
drdelambre: feels like a lot of copy pasta going on. i forget why we arent just putting the message definitions here?

ui/src/external/monaco.flux.completions.ts
bthesorceror: maybe just a TODO/Note comment would work too, just wanna make sure we remember that this try catch is important and don't remove it

ui/src/external/monaco.flux.server.ts
bthesorceror: I do like having functions to build the messages remove a layer to this. i.e. the server maintains state like the id and the message builders can be pure functions :)

ui/src/external/monaco.flux.completions.ts
bthesorceror: just moved the default return into the catch

PR #17383, opened 2020-03-23
fix(authorizer): fix auth for authorizations

affo:
This PR changes the auth process for authorizations.

__Before__: authorized for a an authorization => authorized to read/write the user in the auth  
__Now__: authorized for an auth => authorized to read/write the auth

Example:

Given auth:

```
&influxdb.Authorization {
  ID: 1,
  UserID: 10,
  OrgID: 100,
}
```

__Before__: example permission

```
{'action': 'write', 'resource': {'id': '10', 'type': 'users'}}
```

__Now__: example permission

```
{'action': 'write', 'resource': {'id': '1', 'orgid': '100' 'type': 'authorizations'}}
```

__UPDATE__:

Now updated to 

```
authorized for an auth => authorized to read/write the auth and the user in the auth
```

This is more conservative and adheres better to the previous implementation 👍 

The permissions now are twofold (auth + user in auth):

```
{'action': 'write', 'resource': {'id': '10', 'type': 'users'}}
{'action': 'write', 'resource': {'id': '1', 'orgid': '100' 'type': 'authorizations'}}
```

This code makes auth adhere with the new auth framework and make auth authorizations more in line with what done for other resources.

### DIFF
--- a/authorizer/auth.go
+++ b/authorizer/auth.go
@@ -28,11 +28,12 @@ func (s *AuthorizationService) FindAuthorizationByID(ctx context.Context, id inf
 	if err != nil {
 		return nil, err
 	}
-
-	if err := authorizeReadAuthorization(ctx, a.UserID); err != nil {
+	if _, _, err := AuthorizeRead(ctx, influxdb.AuthorizationsResourceType, a.ID, a.OrgID); err != nil {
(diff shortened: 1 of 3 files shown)


affo:
FYI @gshif

PR #17385, opened 2020-03-23
feat(tenant): add duplicate read service

affo:
Can now add the nw meta service:

`INFLUXD_NEW_META_STORE` will enable it  (default false)
`INFLUXD_NEW_META_STORE_READ_ONLY` will make it read-only (default true)

### DIFF
--- a/tenant/duplicate_reads.go
+++ b/tenant/duplicate_reads.go
@@ -1,0 +1,333 @@
+package tenant
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/influxdata/influxdb"
+	"github.com/influxdata/influxdb/kv"
+	"go.uber.org/zap"
+)
+
+var _ influxdb.TenantService = (*tenantService)(nil)
+
+// readOnlyStore is a wrapper for kv.Store that ensures that updates are not applied.
+type readOnlyStore struct {
+	kv.Store
+}
+
+func (r readOnlyStore) Update(context.Context, func(kv.Tx) error) error {
+	return nil
+}
+
+type tenantService struct {
+	log *zap.Logger
+
+	oldBucketSvc influxdb.BucketService
+	newBucketSvc influxdb.BucketService
+	oldOrgSvc    influxdb.OrganizationService
+	newOrgSvc    influxdb.OrganizationService
+	oldURMSvc    influxdb.UserResourceMappingService
+	newURMSvc    influxdb.UserResourceMappingService
+	oldUserSvc   influxdb.UserService
+	newUserSvc   influxdb.UserService
+	oldPwdSvc    influxdb.PasswordsService
+	newPwdSvc    influxdb.PasswordsService
+}
+
+// NewReadOnlyStore returns a Store that cannot update the underlying kv.Store.
+func NewReadOnlyStore(store kv.Store) (*Store, error) {
+	return NewStore(readOnlyStore{Store: store})
+}
+
+// NewDuplicateReadTenantService returns a tenant service that duplicates the reads to oldSvc and newSvc.
+// The foreseen use case is to compare two service versions, an old one and a new one.
+// The resulting influxdb.TenantService:
+//  - forwards writes to the old service;
+//  - reads from the old one, if no error is encountered, it reads from the new one;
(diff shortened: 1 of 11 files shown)


tenant/duplicate_reads.go
```diff
@@ -0,0 +1,362 @@
...
+
+// NewDuplicateReadTenantService returns a tenant service that duplicates the reads of a kv.Service to a tenant.Service.
+// It does so by creating a tenant.Service with the given store.
+// For consistent results, the given kv.Service should be built on top of the same store.
+// The foreseen use case is to compare two service versions, an old one and a new one.
+// The resulting influxdb.TenantService:
+//  - forwards writes to the old service;
+//  - reads from the old one, if no error is encountered, it reads from the new one;
+//  - compares the results obtained and logs the difference, if any.
+func NewDuplicateReadTenantService(log *zap.Logger, oldSvc *kv.Service, store kv.Store) (influxdb.TenantService, error) {
```
lyondhill: I would like this function declaration to be a little more abstract. Maybe something like `NewDuplicateReadTenantService(log *zap.Logger, oldSvc influxdb.TenantService, newSvc influxdb.TenantService) (influxdb.TenantService, error)`

I think we may want to use this same pattern in the future dont want to be tied to the *kv.Service or kv.Store.

The purpose of this duplication behavior is not just for read only. Yes right now we are only readying from the new and all writes to the old but this will not be the case very soon. We need to plan ahead for this.

tenant/duplicate_reads.go
```diff
@@ -0,0 +1,362 @@
...
+	"go.uber.org/zap"
+)
+
+var _ influxdb.BucketService = (*dupReadBucketService)(nil)
+var _ influxdb.OrganizationService = (*dupReadOrganizationService)(nil)
+var _ influxdb.UserResourceMappingService = (*dupReadUserResourceMappingService)(nil)
+var _ influxdb.UserService = (*dupReadUserService)(nil)
+
+// readOnlyStore is a wrapper for kv.Store that ensures that updates are not applied.
+type readOnlyStore struct {
```
lyondhill: We will very quickly be rolling in a feature flagging system that allows us to write to the "new" tenant service. I suspect the read only store will cause some headaches when we do this.

tenant/duplicate_reads.go
```diff
@@ -0,0 +1,362 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return tenantService{
+		BucketService:              NewDuplicateReadBucketService(log, oldSvc, svc),
+		OrganizationService:        NewDuplicateReadOrganizationService(log, oldSvc, svc),
+		UserResourceMappingService: NewDuplicateReadUserResourceMappingService(log, oldSvc, svc),
+		UserService:                NewDuplicateReadUserService(log, oldSvc, svc),
+		PasswordsService:           NewDuplicateReadPasswordService(log, oldSvc, svc),
+	}, nil
```
lyondhill: I think the functions could just be on the `tenantService` instead of making separate structs for each sub system. When we rollin feature flagging I wouldn't want all the sub structs to have to share information about who or what systems are going to write. I suspect it would be better to flatten this.

tenant/service_user.go
```diff
@@ -8,6 +8,12 @@ import (
 	"golang.org/x/crypto/bcrypt"
 )
 
+// UserService is a union of UserService and PasswordsService.
+type UserService interface {
+	influxdb.UserService
+	influxdb.PasswordsService
+}
+
```
lyondhill: This is confusing. the system already understands the user service https://github.com/influxdata/influxdb/blob/master/user.go#L43-L64 By redefining it here it will become difficult to understand which implementation we are using.

tenant/duplicate_reads.go
affo: Planning to made this modular, wait for it.

I wouldn't feel safe without it.

I don't want a `Find*` to accidentally issue an `Update`

affo:
@lyondhill I cannot agree more with you. Thanks for the precise comments 👍 

I think I have addressed everything in my last update!

tenant/duplicate_reads.go
lyondhill: I think its ok to leave in and then we can swap it out when we are ready to write with the new.

PR #17391, opened 2020-03-23
fix(ui): fixed threshold check bug that wouldn't allow threshold checks to be created where fields had spaces in them

### LINKED ISSUE #17129: Can't add threshold check in Cloud 2 where field has a space in the name
When creating a threshold check in cloud 2, selecting a field with a space in the name prevents that check from being saved. An error message is displayed similar to:

`Failed to create check: Could not create task from check: invalid options; Err loc 20:3-20:27 expected an operator between two expressions: loc 20:3-20:27:`

The numbers referenced can vary depending on the field name selected.

Asalem1:
Closes #17129 

### Problem

API could not handle spaces in field names

### Solution

Updated the `ast.MemberExpression` Property to be a string rather than `ast.Identifier` so that we forced properties to bracket notation instead of dot notation (which was the cause of the error)

threshold-bug-fix


drdelambre:
does typing in a flux query with `r.beans` instead of `r.["beans"]` still work when initiated from the client side?

Asalem1:
> does typing in a flux query with `r.beans` instead of `r.["beans"]` still work when initiated from the client side?

Yup.

PR #17396, opened 2020-03-24
feat: slightly configurable line writer

### LINKED ISSUE #17369: Create configurable Line Protocol writer
## Background

We want to write to RUM data and performance telemetry to an influxdb 2.0 instance, likely in our tools cluster. We'll want a module to be able to write line protocol data to that cluster.

### Acceptance Criteria

Write a module to write arbitrary line protocol data.

#### The following bits are required:

* Takes a measurement to write to
* Takes an arbitrary length list of tags
* Takes an arbitrary length list of fields

#### The following optional configurations should come with sensible defaults:

* Automatically writes a timestamp based on precision
  * Allows setting of timestamps
* Allows setting of precision
* Allows configurability of environment:
  * url
  * org
  * bucket

### Resources
* https://github.com/influxdata/github-issue-stats/blob/master/api/src/stat/line.js#L53
* https://github.com/influxdata/influxdb/blob/master/ui/cypress/support/commands.ts#L396

hoorayimhelping:
Closes #17369

Adds a module that programmatically writes line data to the server, org and bucket specified.

[here's a gist](https://gist.github.com/hoorayimhelping/9e3fc42f152769c0ed9272fc17ebaa58#file-foo-diff-L93) of it being used to implement RUM

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/utils/lineWriter.ts
+++ b/ui/src/utils/lineWriter.ts
@@ -1,0 +1,174 @@
+export interface Tags {
+  [key: string]: string
+}
+
+export interface Fields {
+  [key: string]: number | string
+}
+
+export enum Precision {
+  ns = 'ns',
+  u = 'u',
+  ms = 'ms',
+  s = 's',
+  m = 'm',
+  h = 'h',
+}
+
+export interface BatchOptions {
+  maxIntervalInSeconds: number
+  maxBatchedLines: number
+}
+
+const defaultBatchOptions: BatchOptions = {
+  maxIntervalInSeconds: 10,
+  maxBatchedLines: 100,
+}
+
+const nowInSeconds = function nowInSeconds() {
+  return Math.floor(Date.now() / 1000)
+}
+
+export class LineWriter {
+  protected url: string
+  protected orgID: number | string
+  protected bucketName: string
+  protected authToken: string
+  protected batchOptions: BatchOptions
+
+  protected timerID: number | null
+  protected batchedLines: number
+  protected lines: string[]
+
+  constructor(
+    url: string,
+    orgID: number | string,
+    bucketName: string,
+    authToken: string,
+    batchOptions: BatchOptions = defaultBatchOptions
+  ) {
+    this.url = url
+    this.orgID = orgID
+    this.bucketName = bucketName
+    this.authToken = authToken
+    this.batchOptions = batchOptions
+
+    this.timerID = null
+    this.batchedLines = 0
+    this.lines = []
+  }
+
+  public createLineFromModel(
+    measurement: string,
+    fields: Fields,
+    tags: Tags = {},
+    timestamp: number = nowInSeconds()
+  ): string {
+    let tagString = ''
+    Object.keys(tags)
+      // Sort keys for a little extra perf
+      // https://v2.docs.influxdata.com/v2.0/write-data/best-practices/optimize-writes/#sort-tags-by-key
+      .sort((a, b) => a.localeCompare(b))
+      .forEach((tagKey, i, tagKeys) => {
+        const tagValue = tags[tagKey]
+        const printableTagKey = tagKey
+          .replace(/\s+/g, '\\ ') // replace any number of spaces with an escaped space
+          .replace(/,/g, '\\,') // replace commas with escaped commas
+          .replace(/=/g, '\\=') // replace equal signs with escaped equal signs
+
+        const printableTagValue = tagValue
+          .replace(/\n/g, '') // remove newlines
+          .replace(/\s+/g, '\\ ') // replace any number of spaces with an escaped space
+          .replace(/,/g, '\\,') // replace commas with escaped commas
+          .replace(/=/g, '\\=') // replace equal signs with escaped equal signs
+
+        tagString = `${tagString}${printableTagKey}=${printableTagValue}`
+
+        // if this isn't the end of the string, append a comma
+        if (i < tagKeys.length - 1) {
+          tagString = `${tagString},`
+        }
+      })
+
+    let fieldString = ''
+    Object.keys(fields).forEach((fieldKey, i, fieldKeys) => {
+      const fieldValue = fields[fieldKey]
+
+      const printableFieldKey = fieldKey
+        .replace(/\s+/g, '\\ ') // replace any number of spaces with an escaped space
+        .replace(/,/g, '\\,') // replace commas with escaped commas
+        .replace(/=/g, '\\=') // replace equal signs with escaped equal signs
+
+      let printableFieldValue = fieldValue
+      if (typeof fieldValue === 'string') {
+        printableFieldValue = fieldValue
+          .replace(/\n/g, '') // remove newlines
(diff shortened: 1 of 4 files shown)


drdelambre:
there's a bit more logic to this. try and inject the regexp filters from here to more closely follow the spec:
https://github.com/influxdata/github-issue-stats/blob/master/api/src/stat/line.js#L80

hoorayimhelping:
@drdelambre 

https://github.com/influxdata/influxdb/pull/17396/commits/800a45d34aa244647d5839a3652f74bd9ccdabcf

ui/src/utils/lineWriter.ts
```diff
@@ -0,0 +1,129 @@
+import AJAX from 'src/utils/ajax'
```
Asalem1: Is there a reason we're using `AJAX` rather than `fetch`?

hoorayimhelping: AJAX uses axios, which uses XMLHTTPRequest which is more widely supported

Asalem1: TIL

ui/src/utils/lineWriter.ts
```diff
@@ -0,0 +1,129 @@
...
+    lineStart = `${lineStart},${tagString}`
+  }
+
+  return `${lineStart} ${fieldString} ${timestamp}`
+}
+
+// Builds a basic line writer that will write a line to the url, org and bucket specified.
+// Returns a function named `writeLine` that utilizes closures to access the configuration
+// arguments passed in to `buildLineWriter`.
+export const buildLineWriter = function buildLineWriter(
```
desa: One thing we might want to consider it building up a buffer of metrics instead of just writing a single line. Maybe something on the order of ~100 metrics before making the actual network request.

Common patterns I've seen for influx writers are to have batch size and interval be configurable options. For example write 100 metrics in a batch or every 10s, whichever is first.

hoorayimhelping: @desa 
right on thanks! i figured that would be a suggestion :)

how does batching writes affect precision, if at all?

for that matter, what kind of precision is appropriate here? is second precision too coarse? could second-level precision lead to a situation where if there are two writes in the same second, the later one overwrites the earlier one?

ui/src/utils/lineWriter.ts
drdelambre: we polyfill fetch though right? since we are actively depreciating anything that uses axios (as it's currently only tied to the old api client), it would be cool to see new code not using it so that we can keep api auth mechanisms centralized in `src/client` (like where tokens are stored and header params and the like). Not a blocker, just bumps the work to the client library tech debt epic right off the bat.

ui/src/utils/lineWriter.ts
desa: Typically each point in a batch shares a common timestamp precision. Second precision here is probably fine. I cant imagine we'd want to track these with sub second precision

desa: > could second-level precision lead to a situation where if there are two writes in the same second, the later one overwrites the earlier one?

Could happen, but I think if we're reporting with that level of frequency we're doing something wrong

ui/src/utils/lineWriter.ts
hoorayimhelping: @drdelambre my thinking is since this isn't a client interacting with our api, but making a request to an external service, it shouldn't really go in `src/client`. i was under the assumption that `src/client` is for communicating with our api via swagger. is that not the case? if it is, do we have a story around how to make requests to external services or resources? i assumed that was the intent of `AJAX`.

either way, first I'd heard we're actively deprecating anything that uses axios. an effective way to let people know this is to put a comment saying something is deprecated and shouldn't be used and link to a place that explains that. do we have any documentation like an rfc around deprecating axios that i can link to? or anything explaining what to use instead?

ui/src/utils/lineWriter.ts
```diff
@@ -0,0 +1,237 @@
...
+  protected authToken: string
+  protected batchOptions: BatchOptions
+
+  protected timerID: number | null
+  protected batchedLines: number
+  protected lines: string[]
+
+  constructor(
+    url: string,
+    orgId: number | string,
```
121watts:
```suggestion
    orgID: number | string,
```
nit here (i prefer `Id`) actually.  sadly, the rest of our app uses `ID`

ui/src/utils/lineWriter.ts
```diff
@@ -0,0 +1,237 @@
...
+    measurement: string,
+    fields: Fields,
+    tags: Tags = {},
+    timestamp: number = nowInSeconds()
+  ): string {
+    let tagString = ''
+    Object.keys(tags)
+      // Sort keys for a little extra perf
+      // https://v2.docs.influxdata.com/v2.0/write-data/best-practices/optimize-writes/#sort-tags-by-key
+      .sort()
```
121watts:
```suggestion
      .sort(((a, b) () => a.localeCompare(b))
```
JavaScripts `sort` array prototype sorts on UTF-16 code units values.  it wont sort lexicographically (alphabetically).

ui/src/utils/lineWriter.ts
121watts: we do indeed polyfill fetch.  would be nice not to rely on axios for this.  this def doesn't need to (or should be) within `src/client`.  i agree that we should just rip axois out altogether and eliminate the confusion.

drdelambre: We had a big talk about it during planning a month ago when bonitoo was releasing the new api client, and made these in an effort to put a bow on an effort that was 90% of the way there:
https://github.com/influxdata/influxdb/issues?q=is%3Aopen+is%3Aissue+label%3Ateam%2Fmonitoring+label%3Akind%2Ftech-debt+generated+client
Everyone has been busy doing other stuff though. We currently treat our queries as an external resource (weird right?) within the app, which you can get a gander of over here:
https://github.com/influxdata/influxdb/blob/master/ui/src/shared/apis/query.ts#L55

ui/src/utils/lineWriter.ts
```diff
@@ -0,0 +1,237 @@
...
+        },
+        true
+      )
+    } catch (error) {
+      console.error(error)
+    }
+  }
+}
+
+// Builds a basic line writer that will write a line to the url, org and bucket specified.
```
drdelambre: whats all this?

ui/src/utils/lineWriter.ts
```diff
@@ -0,0 +1,237 @@
...
+
+  protected writeLine = async (
+    line: string,
+    precision: Precision = Precision.s
+  ) => {
+    try {
+      return await AJAX(
+        {
+          method: 'POST',
+          url: `${this.url}/api/v2/write`,
```
drdelambre: where does this url get set?

drdelambre: found it

Issue #17356, opened 2020-03-19
Remove required '@' prefix for file paths in 'influx' CLI


sanderson:
__Proposal:__
Remove the required `@` prefix when specifying file paths in the `influx` CLI. It breaks any shell's native tab-completion.

__Desired behavior:__
I want to be able to use my shell's native tab-completion to quickly type file paths when using the `influx` CLI.

__Use case:__
Ease of use.

jsteenb2:
toooootally agree, we should support piping everywhere as well withou the arcane `-` requirement

jademcgough:
As part of the story I think it's important to understand why it's designed this way and if there are any implications for changing it, because I don't have context on that right now.

kelwang:
my guess is @ means a file to parse

sranka:
#17398  introduces a new  --file (or -f ) flag that let the user specify the file to write

sanderson:
@sranka does your PR only introduce it to the `write` command? This change also needs to happen with all commands that accept a file as input.:

```
influx write
influx query
influx task create
```

_There may be others_

sranka:
Only the write command

kelwang:
this is fixed by https://github.com/influxdata/influxdb/pull/17398

sanderson:
@kelwang #17398 only address the `write` command. The `query` and `task create` commands still need to be updated with the `--file` flag.

Issue #17259, opened 2020-03-13
clean up persistent flag 


kelwang:
-t --token is a persistent global flag, which is required for `influx config create`, but not required for `influx config update`. 

We need to display the real usage of the token, without bother other sub command

PR #17398, opened 2020-03-24
feat(cmd/influx): influx cli can write CSV files

### LINKED ISSUE #17003: Add the ability to import an annotated Flux CSV response via the influx cli
This should be the signature of the command:
```
#Import annotated Flux CSV response
influx write --bucket fly --org ACME --format csv --file /path/to/flights.csv
```

```
➜  influxdb git:(master) ✗ ./bin/darwin/influx write -h
Write data to InfluxDB via stdin, or add an entire file specified with the -f flag

Usage:
  influx write [flags]

Flags:
  -b, --bucket string      The name of destination bucket; Maps to env var $INFLUX_BUCKET_NAME
      --bucket-id string   The ID of destination bucket; Maps to env var $INFLUX_BUCKET_ID
  -h, --help               Help for the write command
  -o, --org string         The name of the organization that owns the bucket; Maps to env var $INFLUX_ORG
      --org-id string      The ID of the organization that owns the bucket; Maps to env var $INFLUX_ORG_ID
  -p, --precision string   Precision of the timestamps of the lines; Maps to env var $INFLUX_PRECISION (default "ns")
      --format string      The format of the file to load. Supports lp (Line Protocol) or csv (Comma Separated Values). Defaults based on the file extension, or lp if it cannot be determined.
  -f, --file string        The path to the file to import

Global Flags:
      --host string    HTTP address of Influx; Maps to env var $INFLUX_HOST (default "http://localhost:9999")
  -t, --token string   API token to be used throughout client calls; Maps to env var $INFLUX_TOKEN
```

It should be able to load the following csv: 
[2020-02-25-14-22_chronograf_data.csv.zip](https://github.com/influxdata/influxdb/files/4252466/2020-02-25-14-22_chronograf_data.csv.zip)

sranka:
Closes #17003 #17356

This PR 
- supports CSV files in 'influx write'
   - transformation of CSV data to line protocol is driven by CSV annotations
   - annotations described https://v2.docs.influxdata.com/v2.0/reference/syntax/annotated-csv/#annotations are all supported, so that flux query result can be written "as-is"
   - *#datatype* annotation is enhanced:
      * so you can mark non-field data with: 
          * _measurement_, _tag_, _time_
          * _ignore_ value in a column simply ignores the column
          * _dateTime_ is either as a long (int64) number or RFC3339 format, this type is always used to represent time of measurement; or you can specify
             * _dateTime:RFC3339_ for RFC3339 time
             * _dateTime:number_ to expect a long number
          * _time_ value is also supported as an alias for the existing _dateTime_
      * a _field_ data type can be used to let detect field's data type
      * default datatype for a column is =field= unless _field column is present (ignored then)
      * there can be at most 1 _dateTime_ column
   - https://github.com/bonitoo-io/influxdb-csv-import/blob/master/README.md contains demonstrative CSV examples
- adds new options to 'influx write":
```
  -f, --file string        The path to the file to import
      --format string      Input format, either lp (Line Protocol) or csv (Comma Separated Values). Defaults to lp unless '.csv' extension
```
- a new "write dryrun" (i.e. 'write' command with 'dryrun' sub-command) can be used to validate and tune annotations in CSV files; parameters are the as in "write", but line protocol is written to stdout
- keeps the existing functionality (writing line protocol data) without changes
  - the default input format is lp (line protocol) unless a file with ".csv" extension is supplied

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influx/main.go
+++ b/cmd/influx/main.go
@@ -25,7 +25,7 @@ import (
 	"go.uber.org/zap"
 )
 
-const maxTCPConnections = 128
+const maxTCPConnections = 10
 
 func main() {
 	influxCmd := influxCmd()
--- a/cmd/influx/main.go
+++ b/cmd/influx/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -86,6 +87,10 @@ func (o genericCLIOpts) newCmd(use string, runE func(*cobra.Command, []string) e
 	cmd.SetIn(o.in)
 	cmd.SetErr(o.errW)
 	return cmd
+}
+
+func (o genericCLIOpts) writeJSON(v interface{}) error {
+	return writeJSON(o.w, v)
 }
 
 func (o genericCLIOpts) newTabWriter() *internal.TabWriter {
@@ -436,10 +441,39 @@ func (f flagOpts) mustRegister(cmd *cobra.Command) {
 	cli.BindOptions(cmd, f)
 }
 
+func registerPrintOptions(cmd *cobra.Command, headersP, jsonOutP *bool) {
+	var opts flagOpts
+	if headersP != nil {
+		opts = append(opts, cli.Opt{
+			DestP:   headersP,
+			Flag:    "hide-headers",
+			EnvVar:  "HIDE_HEADERS",
+			Desc:    "Hide the table headers; defaults false",
+			Default: false,
+		})
+	}
+	if jsonOutP != nil {
+		opts = append(opts, cli.Opt{
+			DestP:   jsonOutP,
+			Flag:    "json",
+			EnvVar:  "OUTPUT_JSON",
+			Desc:    "Output data as json; defaults false",
+			Default: false,
+		})
+	}
+	opts.mustRegister(cmd)
+}
+
 func setViperOptions() {
 	viper.SetEnvPrefix("INFLUX")
 	viper.AutomaticEnv()
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
+}
+
+func writeJSON(w io.Writer, v interface{}) error {
+	enc := json.NewEncoder(w)
+	enc.SetIndent("", "\t")
+	return enc.Encode(v)
 }
 
 func newBucketService() (influxdb.BucketService, error) {
--- a/cmd/influx/main.go
+++ b/cmd/influx/main.go
@@ -306,10 +306,6 @@ func writeConfigToPath(tok, org, path, dir string) error {
 }
 
 func checkSetup(host string, skipVerify bool) error {
-	if host == "-" {
-		// dry-run executed
-		return nil
-	}
 	s := &http.SetupService{
 		Addr:               host,
 		InsecureSkipVerify: skipVerify,
--- a/cmd/influx/main.go
+++ b/cmd/influx/main.go
@@ -181,12 +181,12 @@ func (b *cmdInfluxBuilder) cmd(childCmdFns ...func(f *globalFlags, opt genericCL
 
 	if flags.Token == "" {
 		// migration credential token
-		if migrateOldCredential() {
-			// this is after the flagOpts register b/c we don't want to show the default value
-			// in the usage display. This will add it as the config, then if a token flag
-			// is provided too, the flag will take precedence.
-			flags.Config = getConfigFromDefaultPath()
-		}
+		migrateOldCredential()
+
+		// this is after the flagOpts register b/c we don't want to show the default value
+		// in the usage display. This will add it as the config, then if a token flag
+		// is provided too, the flag will take precedence.
+		flags.Config = getConfigFromDefaultPath()
 	}
 
 	cmd.PersistentFlags().BoolVar(&flags.local, "local", false, "Run commands locally against the filesystem")
@@ -274,22 +274,21 @@ func getConfigFromDefaultPath() config.Config {
 	return activated
 }
 
-func migrateOldCredential() bool {
+func migrateOldCredential() {
 	dir, err := fs.InfluxDir()
 	if err != nil {
-		return false // no need for migration
+		return // no need for migration
 	}
 	tokB, err := ioutil.ReadFile(filepath.Join(dir, http.DefaultTokenFile))
 	if err != nil {
-		return false // no need for migration
+		return // no need for migration
 	}
 	err = writeConfigToPath(strings.TrimSpace(string(tokB)), "", filepath.Join(dir, http.DefaultConfigsFile), dir)
 	if err != nil {
-		return false
+		return
 	}
 	// ignore the remove err
 	_ = os.Remove(filepath.Join(dir, http.DefaultTokenFile))
-	return true
 }
 
 func writeConfigToPath(tok, org, path, dir string) error {
@@ -308,7 +307,7 @@ func writeConfigToPath(tok, org, path, dir string) error {
 
 func checkSetup(host string, skipVerify bool) error {
 	if host == "-" {
-		// dry run executed
+		// dry-run executed
 		return nil
 	}
 	s := &http.SetupService{
--- a/cmd/influx/main.go
+++ b/cmd/influx/main.go
@@ -68,15 +68,20 @@ type genericCLIOpts struct {
 	runEWrapFn cobraRunEMiddleware
 }
 
-func (o genericCLIOpts) newCmd(use string, runE func(*cobra.Command, []string) error) *cobra.Command {
+func (o genericCLIOpts) newCmd(use string, runE func(*cobra.Command, []string) error, useRunEMiddleware bool) *cobra.Command {
 	cmd := &cobra.Command{
 		Args: cobra.NoArgs,
 		Use:  use,
 		RunE: runE,
 	}
-	if runE != nil && o.runEWrapFn != nil {
+
+	canWrapRunE := runE != nil && o.runEWrapFn != nil
+	if useRunEMiddleware && canWrapRunE {
 		cmd.RunE = o.runEWrapFn(runE)
+	} else if canWrapRunE {
+		cmd.RunE = runE
 	}
+
 	cmd.SetOut(o.w)
 	cmd.SetIn(o.in)
 	cmd.SetErr(o.errW)
@@ -148,7 +153,7 @@ func (b *cmdInfluxBuilder) cmd(childCmdFns ...func(f *globalFlags, opt genericCL
 		setViperOptions()
 	})
 
-	cmd := b.newCmd("influx", nil)
+	cmd := b.newCmd("influx", nil, false)
 	cmd.Short = "Influx Client"
 	cmd.SilenceUsage = true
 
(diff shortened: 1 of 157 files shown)


cmd/influx/main.go
```diff
@@ -176,12 +176,12 @@ func (b *cmdInfluxBuilder) cmd(childCmdFns ...func(f *globalFlags, opt genericCL
...
-
-		// this is after the flagOpts register b/c we don't want to show the default value
-		// in the usage display. This will add it as the config, then if a token flag
-		// is provided too, the flag will take precedence.
-		flags.Config = getConfigFromDefaultPath()
+		if migrateOldCredential() {
+			// this is after the flagOpts register b/c we don't want to show the default value
+			// in the usage display. This will add it as the config, then if a token flag
+			// is provided too, the flag will take precedence.
+			flags.Config = getConfigFromDefaultPath()
```
kelwang: This change will break the multiple configs feature. Usage display override is related to the cobra persistent flags, has nothing to do with migration. https://github.com/influxdata/influxdb/issues/17259

sranka: Understood, I am going to revert this part to a state "before this PR".

cmd/influx/main.go
```diff
@@ -301,6 +301,10 @@ func writeConfigToPath(tok, org, path, dir string) error {
 }
 
 func checkSetup(host string, skipVerify bool) error {
+	if host == "-" {
```
jsteenb2: Why is this necessary here?

cmd/influx/write.go
```diff
@@ -68,94 +75,128 @@ or add an entire file specified with an @ prefix.`
...
+	if writeFlags.DryRun {
+		flags.Host = "-"
 	}
 
-	if writeFlags.Bucket != "" && writeFlags.BucketID != "" {
-		return fmt.Errorf("please specify one of bucket or bucket-id")
-	}
+	var bucketID, orgID platform.ID
+	// validate flags unless writing to stdout
+	if flags.Host != "-" {
```
jsteenb2: I don't understand this, why is this here? Maybe a comment added here at the very least

cmd/influx/write.go
```diff
@@ -68,94 +75,128 @@ or add an entire file specified with an @ prefix.`
...
+	}
 
-	s := write.Batcher{
-		Service: &http.WriteService{
-			Addr:               flags.Host,
-			Token:              flags.Token,
-			Precision:          writeFlags.Precision,
-			InsecureSkipVerify: flags.skipVerify,
-		},
+	if writeFlags.Format == inputFormatCsv {
```
jsteenb2: can we inspect the filename to determine if it is `csv` vs `lp`? if so, we can make this a bit simpler for hte user by not requiring the format, and inspecting the file instead.

cmd/influx/write.go
```diff
@@ -68,94 +75,128 @@ or add an entire file specified with an @ prefix.`
...
-		return err
-	}
+		if !models.ValidPrecision(writeFlags.Precision) {
+			return fmt.Errorf("invalid precision")
+		}
 
-	var filter platform.BucketFilter
-	if writeFlags.BucketID != "" {
-		filter.ID, err = platform.IDFromString(writeFlags.BucketID)
+		bs, err := newBucketService()
```
kelwang: If you flag dry host to be '-', you could get an error here, since bucket service is a tcp request to the host.

cmd/influx/main.go
sranka: There is a new *--dry-run* flag in "influx write" command that helps to tune CSV to line protocol conversion, it prints lines to stdout. This dry-run does not require any check of influx server setup.

jsteenb2: have you considered making dry run a new command instead?

```sh
influx write dryrun
```

kelwang:
with this pr, we can also close #17356

cmd/influx/write.go
```diff
@@ -15,20 +15,27 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	inputFormatCsv          = "csv"
+	inputFormatLineProtocol = "lp"
+)
+
 var writeFlags struct {
 	OrgID     string
```
jsteenb2: can you replace OrgID and Org with the org type that does this:

https://github.com/influxdata/influxdb/blob/master/cmd/influx/main.go#L360

you can see it being used here:

setting the flags: https://github.com/influxdata/influxdb/blob/master/cmd/influx/pkg.go#L100
validating org id: https://github.com/influxdata/influxdb/blob/master/cmd/influx/pkg.go#L108
getting org id from type: https://github.com/influxdata/influxdb/blob/master/cmd/influx/pkg.go#L122

side note: just found redundant code I wrote in that cmd 😬

cmd/influx/write.go
sranka: The ".csv" file extension now determines the CSV format. We might do better by inspecting the first row of the data that comes in, but such a detection might be error-prone (I don't know yet). I would prefer to do this in a separate issue after this PR is merged.

cmd/influx/main.go
sranka: Thank you, I only wanted to bypass the 'setup' middleware for dry-run, and this solution does not appear to me on the first sight. Will do.

cmd/influx/write.go
sranka: will remove this, and introduce a dry run sub-command

cmd/influx/write.go
sranka: :) ok

cmd/influx/write.go
sranka: Is there any reliable way to detect line protocol on input in the existing code base? Or should we simply assume CSV data if input starts with an annotation comment (no real measurements probably start with #)

If the data input is supplied via stdin, the user should better let us know what data type (lp or csv) comes in ... the user knows, but the tool guesses. I could also imagine users supplying wrong data, from which we will then provide false advice about parsing errors.

@jsteenb2 Can you please comment on this?

jsteenb2: totally fine breakign this out into separate issue. Its a quality of life feature. I'm not advocating that we remove the format. I was thinking it coudl be optional if we have a file as an input.  I dont' think we have a reliable way to inspect a buffer and pull out if it is LP or csv 🤔. I could be wrong on that. Maybe our storage reviewer (@sebito91) would know.

jsteenb2: we do something similar to this in the pkg command. We just bet on the file extension being accurate and make decisions off of that if the `encoding` flag isn't set 👍 

https://github.com/influxdata/influxdb/blob/master/cmd/influx/pkg.go#L490

sranka: Yes, the format flag is already optional herein, defaults to lp (backward compatibility) unless a file with .csv extension supplied.

https://github.com/influxdata/influxdb/blob/17003/writeFromCsv/cmd/influx/write.go#L154

cmd/influx/main.go
jsteenb2: did the dryrun command get added? I'm not seeing it 😬

cmd/influx/write.go
```diff
@@ -68,94 +62,133 @@ or add an entire file specified with an @ prefix.`
 		},
 	}
 	opts.mustRegister(cmd)
-
+	cmd.PersistentFlags().StringVar(&writeFlags.Format, "format", "", "Input format, either lp (Line Protocol) or csv (Comma Separated Values). Defaults to lp unless '.csv' extension")
+	cmd.PersistentFlags().StringVarP(&writeFlags.File, "file", "f", "", "The path to the file to import")
+
+	cmdDryRun := opt.newCmd("dryrun", func(command *cobra.Command, args []string) error {
```
jsteenb2: can you break out the `dryrun` command into its own command?

cmd/influx/write.go
```diff
@@ -68,94 +62,133 @@ or add an entire file specified with an @ prefix.`
...
-			InsecureSkipVerify: flags.skipVerify,
-		},
+	if writeFlags.Format == inputFormatCsv {
+		r = write.CsvToProtocolLines(r)
 	}
 
-	ctx = signals.WithStandardSignals(ctx)
-	if err := s.Write(ctx, orgID, bucketID, r); err != nil && err != context.Canceled {
-		return fmt.Errorf("failed to write data: %v", err)
+	if writeFlags.DryRun {
```
jsteenb2: curiosity: feels like the Dry Run functionality can be pulled out from this large func. It shares some things, but those shared things you can pull out as needed and clean up the rest of this here. This command was already doing a fair amount, the dry run stuff baked into here makes it harder to follow. WDYT?

cmd/influx/write.go
sranka: Sure

cmd/influx/write.go
sranka: At the time when I was initially writing this, I thought that it is less error-prone to have a single function that is executed for both write and 'write with dry-run'. Now, after the changes, when the flags settle down and doing a few steps back to see the code better, I have to agree that the code complexity will reduce if there will be separate functions for each command that will call a shared function to obtain a line data reader. Good point, thx :)

sanderson:
@sranka If using the linepart annotation in annotated CSV _without_ the datatype annotations, should the `time` columns should always be an integer (not RFC3339) according to the precision the user specifies (just like line protocol) or will it convert RFC3339 timestamps into an integer for you?

sanderson:
@sranka Also, how do you handle string and boolean field values with the `linepart` annotation? Do you have to include a `datatype` annotation? Or is it smart enough to know?

sranka:
@sanderson *linepart* annotation is now independent on *datatype* annotation, at least conceptually. linepart places column data in the line, *datatype* is just field data type. 

Regarding the data type annotation:
- if no data type is specified for a time column, it is detected from the data, so both number and RFC3339 string is supported, see https://github.com/influxdata/influxdb/blob/17003/writeFromCsv/write/csvTable.go#L349
- if no data type is specified for a field column, the column data are copied to line protocol as-is
- the datatype annotation for measurement and tag columns is ignored, the conversion to line protocol is given for these types

It might be simpler to introduce extra _field_,  _measurement_ or _tag_ (and _ignore_) values in the datatype annotation so that the user could then supply a single annotation line (_#datatype_) to specify fields (with types) and non-field columns (measurement, time, tag and ignored). As I write this comment I tend to change the existing implementation this way and remove the linepart annotation. WDYT?

There is a wide area of enhancements that could be done regarding CSV, once this gets merged: 
- time formats may vary
- header/annotation lines could be added from cli line without changing the CSV file
- constant tags, field values or measurement name could be supplied as well
- CSV with more time columns (such as https://data.humdata.org/dataset/novel-coronavirus-2019-ncov-cases) can be supported OOTB
- field types could be detected and enforced
- better error reporting, a better indication of progress / success / failure
- detection of data format without supplying a format flag
- ...
So this is just a first step to support CSV.

cmd/influx/write.go
```diff
@@ -152,11 +184,27 @@ func fluxWriteF(cmd *cobra.Command, args []string) error {
...
-
 	ctx = signals.WithStandardSignals(ctx)
 	if err := s.Write(ctx, orgID, bucketID, r); err != nil && err != context.Canceled {
 		return fmt.Errorf("failed to write data: %v", err)
 	}
 
 	return nil
 }
+
+func fluxWriteDryrunF(cmd *cobra.Command, args []string) error {
```
jsteenb2: excellent, CLI is looking good from my end. Thanks for running with feedback here 👍

PR #17399, opened 2020-03-24
fix: show more

drdelambre:
Closes influxdata/EAR#1466

no idea why this kinda sometimes worked if state was already loaded into localstorage, and not if it was fetched from the api, had a lot of red herrings based in state management surrounding this one. The core of the issue came down to firefox not rendering _sometimes_ when there was only a single row of results in the infinite scroll. If this doesn't work for the client, the next step is to address a potential race condition where the window is measured before state is loaded, but this fix made the problem disappear locally.


drdelambre:
i have no idea how to

PR #17417, opened 2020-03-25
chore(pkger): refactor influx pkg cmd to not validate in CLI and have server validate

jsteenb2:
the `pkger.ValidSkipParseError` option allows our server to be the one to validate the
the pkg is accurate. If a user has an older version of the UI and our cloud gets updated
with new validation rules,they'll get immediate access to that change without having to
role their CLI build.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influx/pkg.go
+++ b/cmd/influx/pkg.go
@@ -318,8 +318,9 @@ func (b *cmdPkgBuilder) cmdPkgSummary() *cobra.Command {
 		return nil
 	}
 
-	cmd := b.newCmd("summary", runE)
+	cmd := b.newCmd("summary", nil)
 	cmd.Short = "Summarize the provided package"
+	cmd.RunE = runE
 
 	b.registerPkgFileFlags(cmd)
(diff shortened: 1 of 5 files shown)


cmd/influx/pkg.go
```diff
@@ -428,16 +430,21 @@ func (b *cmdPkgBuilder) readPkg() (*pkger.Pkg, bool, error) {
 	}
 	pkgs = append(pkgs, urlPkgs...)
 
+	// the pkger.ValidSkipParseError option allows our server to be the one to validate the
+	// the pkg is accurate. If a user has an older version of the UI and our cloud gets updated
```
jademcgough: By "older version of the UI" do you mean CLI, or do you mean the future UI component?

jsteenb2: whoops, I did indeed 😬, great catch 👍

jsteenb2: fix pushed 👍

PR #17448, opened 2020-03-26
feat(pkger): add stack init cmd to influx cli

jsteenb2:
closes: #17235

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influx/pkg.go
+++ b/cmd/influx/pkg.go
@@ -97,10 +97,9 @@ func (b *cmdPkgBuilder) cmdPkgApply() *cobra.Command {
 
 	b.org.register(cmd, false)
 	b.registerPkgFileFlags(cmd)
+	b.registerPkgPrintOpts(cmd)
 	cmd.Flags().BoolVarP(&b.quiet, "quiet", "q", false, "Disable output printing")
 	cmd.Flags().StringVar(&b.applyOpts.force, "force", "", `TTY input, if package will have destructive changes, proceed if set "true"`)
-	cmd.Flags().BoolVarP(&b.disableColor, "disable-color", "c", false, "Disable color in output")
-	cmd.Flags().BoolVar(&b.disableTableBorders, "disable-table-borders", false, "Disable table borders")
 
 	b.applyOpts.secrets = []string{}
 	cmd.Flags().StringSliceVar(&b.applyOpts.secrets, "secret", nil, "Secrets to provide alongside the package; format should --secret=SECRET_KEY=SECRET_VALUE --secret=SECRET_KEY_2=SECRET_VALUE_2")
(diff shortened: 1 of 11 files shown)


cmd/influx/pkg.go
```diff
@@ -345,6 +345,74 @@ func (b *cmdPkgBuilder) cmdPkgValidate() *cobra.Command {
 	return cmd
 }
 
+func (b *cmdPkgBuilder) cmdStack() *cobra.Command {
+	cmd := b.newCmd("stack", nil, false)
+	cmd.Short = "Stack command is the container for all stack CRUD"
```
jademcgough: Could just be `"stack management commands"` to be consistent with other descriptions.

cmd/influx/pkg.go
```diff
@@ -345,6 +345,74 @@ func (b *cmdPkgBuilder) cmdPkgValidate() *cobra.Command {
...
+	if err != nil {
+		return err
+	}
+
+	orgID, err := b.org.getID(orgSVC)
+	if err != nil {
+		return err
+	}
+
+	const fakeUserID = 0 // is 0 because user is pulled from token...
```
jademcgough: The userID can be inferred from the token if you use the auth service, but I'm not sure I understand what it's being used for here.

Is it a problem for the userID to be 0 in some cases?

cmd/influx/pkg.go
```diff
@@ -345,6 +345,74 @@ func (b *cmdPkgBuilder) cmdPkgValidate() *cobra.Command {
...
+	return cmd
+}
+
+func (b *cmdPkgBuilder) cmdStackInit() *cobra.Command {
+	cmd := b.newCmd("init", b.stackInitRunEFn, true)
+	cmd.Short = "Initialize a stack"
+
+	cmd.Flags().StringVarP(&b.name, "stack-name", "n", "", "Name given to created stack")
+	cmd.Flags().StringVarP(&b.description, "stack-description", "d", "", "Description given to created stack")
+	cmd.Flags().StringArrayVarP(&b.urls, "package-url", "u", nil, "Package urls to pull associate with new stack")
```
jademcgough: Remove "pull"

jsteenb2:
@jademcgough that's a good question. Say you are boostrapping your app for the first time. You've created your org, user, bucket, and now you want a stack to associate with some packages you want in the platform. That stack can be bootstrapped with this call here. The stack can have associated urls, which apply will automatically pull and run for the user associated with the stack_id provided. 

Does that help illuminate the need for this?

cmd/influx/pkg.go
jsteenb2: If the ID is real or not, the user assoicated with the token is what is used when creating resources at this time. That is pretty consistent across the API at this pint.

PR #17459, opened 2020-03-27
fix(ui): include link to non-multi-user members cloud page

alexpaxton:
### Problems:

- The `members` page behind the `multiUser` flag has a different URL than `members` outside `multiUser`
- Links pointing to quartz using the router `<Link />` component fail even if the URL is correct

### Changes:

- `multiUser` feature flag should never render anything when in OSS
- If in cloud and `multiUser` is false, only show the `/about` tab in the organization page
- If in cloud and `multiUser` is true, show the `/members` tab in the organization page
- Pass in `linkElement` as a `<a />` element if the link points to quartz
- Pass in `linkElement` as a `<Link />` element if the link points to influxdb
- Ensure `settings/profile` is supported for the old navbar

### Checklist:

- [x] Rebased/mergeable
- [x] Tests pass


ui/src/pageLayout/constants/navigationHierarchy.ts
```diff
@@ -86,30 +114,87 @@ export const generateNavItems = (orgID: string): NavItem[] => {
...
           label: 'Members',
-          link: `${orgPrefix}/members`,
-          cloudExclude: true,
+          featureFlag: 'multiUser',
+          featureFlagValue: false,
+          link: {
+            type: 'link',
+            location: `${orgPrefix}/members`,
+          },
+        },
```
lukevmorris: This subnavitem only ever renders in OSS.

I think we shouldn't feature flag this subnav item and always link `Members` to `/members` in OSS. That means removing the following subnavitem going to `${CLOUD_URL}/organizations/${orgID}${CLOUD_USERS_PATH}`. Technically OSS has its own MultiUser experience, and we should always render that. When the time comes to ship a new experience in OSS, we can rely on a feature flag that the end user can change as the operator

ui/src/organizations/components/OrgNavigation.tsx
```diff
@@ -42,11 +43,18 @@ class OrgNavigation extends PureComponent<Props> {
         cloudExclude: true,
         link: `/orgs/${orgID}/members`,
       },
```
lukevmorris: This is visible always in OSS and never in Cloud. I think this is correct to go to `/members`.

ui/src/organizations/components/OrgNavigation.tsx
```diff
@@ -42,11 +43,18 @@ class OrgNavigation extends PureComponent<Props> {
...
+      {
+        text: 'Members',
+        id: 'members-quartz',
+        cloudOnly: true,
+        link: `${CLOUD_URL}${CLOUD_USERS_PATH}`,
+        featureFlag: 'multiUser',
+        featureFlagValue: false,
+      },
```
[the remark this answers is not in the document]
alexpaxton: Isn't the current members path `${CLOUD_URL}${CLOUD_USERS_PATH}`?
Before I began working on the new nav there was a multi-user link to cloud in there that used this URL:
`${CLOUD_URL}/organizations/${orgID}${CLOUD_USERS_PATH}`

While I don't have any context in cloud, my read on the situation is that there's a different multi-user page in cloud than a regular users page?

So you are saying we just stick to this url: `${CLOUD_URL}/organizations/${orgID}${CLOUD_USERS_PATH}`?

PR #17467, opened 2020-03-27
feat(tenant): add urm indexing into the tenant service

### LINKED ISSUE #17464: Add indexer to the tenant service
Indexer was added to the urm service in the kv system and we need to roll it in to the tenant urm's to avoid a regression.

lyondhill:
fixes #17464

### DIFF
--- a/tenant/storage.go
+++ b/tenant/storage.go
@@ -2,6 +2,7 @@ package tenant
 
 import (
 	"context"
+	"encoding/json"
 
 	"github.com/influxdata/influxdb"
 	"github.com/influxdata/influxdb/kit/tracing"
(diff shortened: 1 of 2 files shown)


tenant/storage.go
```diff
@@ -13,15 +14,30 @@ const MaxIDGenerationN = 100
...
+				if err := json.Unmarshal(v, &urm); err != nil {
+					return nil, err
+				}
+
+				id, _ := urm.UserID.Encode()
+				return id, nil
+			},
+		)),
 	}
+	kv.WithIndexReadPathEnabled(st.urmByUserIndex)
```
GeorgeMac: You can actually pass this as a variadic option, I was just being weird in my previous implementation because of how I wired in the configuration.
e.g. you can:
```go
urmByUserIndex: kv.NewIndex(kv.NewIndexMapping(
    urmBucket,
    urmByUserIndexBucket,
    func(v []byte) ([]byte, error) {
        var urm influxdb.UserResourceMapping
        if err := json.Unmarshal(v, &urm); err != nil {
            return nil, err
        }

        id, _ := urm.UserID.Encode()
        return id, nil
    },
), kv.WithIndexReadPathEnabled),
```

lyondhill: oo nice. thanks!

PR #17472, opened 2020-03-27
feat(cmd/influx): easy switch config

### LINKED ISSUE #17401: shorten command to switch CLI config
Changing between CLI config is the most common config action, but currently that requires updating the config to active, like this:

`influx config update -n my-config -a`

We should make this way easier to remember, ideally no more than three words and no flags.

My preference is that we shorten it to this:

`influx config my-name`

To make `my-name` the active config.

kelwang:
Closes #17401

Add an easy way to switch config, manually tested

- [x] Rebased/mergeable
- [x] Tests pass


kelwang:
> looks good to me. One nit, not related to this however, I wish we represented the Config with the name inside of it. Instead of having the map[name]config. Its a nit, but it concerns me having to juggle state in two places.... something to think about 🤷‍♂

sure, will definitely think about it

PR #17490, opened 2020-03-30
feat(cmd/influx): add config switch back

### LINKED ISSUE #17402: previous CLI config command
in `kubectx` you can switch to the previous context with `kubectx -`.

We should have a simple command to switch to the previous CLI config, as it's common to switch between two contexts.

My preference is `influx config -`, similar to `cd -` or `kubectx -`.

kelwang:
Closes #17402

use `influx config -`, to switch back to previous configuration. If no previous config, or previous config just deleted, `influx config -` will print err `Error: previous activated config is not found`

create a new config with name `-`, for example `influx config create -n=- ` will be blocked. 

manually tested

- [x] Rebased/mergeable
- [x] Tests pass


cmd/influx/config.go
```diff
@@ -60,12 +60,21 @@ func (b *cmdConfigBuilder) cmdSwitchActiveRunEFn(cmd *cobra.Command, args []stri
...
-			Code: influxdb.ENotFound,
-			Msg:  fmt.Sprintf("name %q is not found", b.name),
+	var p0 config.Config
+	if args[0] == "-" {
+		b.name, p0, err = b.svc.ParsePreviousActiveConfig()
+		if err != nil {
+			return err
+		}
+	} else {
+		b.name = args[0]
```
jsteenb2: you don't need to assign the name to the bulider here, probably better to not do that actually 🤔. No need to add a side effect here imo 🤷‍♂

cmd/influx/config/config.go
```diff
@@ -16,9 +16,10 @@ import (
 type Config struct {
 	Host string `toml:"url" json:"url"`
 	// Token is base64 encoded sequence.
-	Token  string `toml:"token" json:"token"`
-	Org    string `toml:"org" json:"org"`
-	Active bool   `toml:"active" json:"active"`
+	Token          string `toml:"token" json:"token"`
```
jsteenb2: let's add `name` in here as well

cmd/influx/config.go
```diff
@@ -110,10 +119,9 @@ func (b *cmdConfigBuilder) cmdCreateRunEFn(*cobra.Command, []string) error {
 	}
 
 	p := config.Config{
-		Host:   b.url,
-		Token:  b.token,
-		Org:    b.org,
-		Active: b.active,
+		Host:  b.url,
+		Token: b.token,
+		Org:   b.org,
```
jsteenb2: why is active dropped when creating a new one?

kelwang: it will be active, when calling the switch later. It is removed, the the previous active logic can be more consistent, you can check the change of create test case for more details

cmd/influx/config/config.go
kelwang: I actually tried about adding name today. But, I find out still have to check the uniqueness among the configs, and more magics had to done on the toml side. It's easier and cleaner to do a simple map.

cmd/influx/config.go
jsteenb2: the amount of side effects in this PR is scaring me... feels very unobvious what is happening here

cmd/influx/config/config.go
jsteenb2: you can have both a map for uniqueness and the name in the config. I'm advocating you do both at this point.

jsteenb2: the map is an internal implementation detail for the config pkg here to manage imo, no reason the cli has to know about how that uniqueness is done

cmd/influx/config.go
```diff
@@ -60,17 +60,28 @@ func (b *cmdConfigBuilder) cmdSwitchActiveRunEFn(cmd *cobra.Command, args []stri
...
+		}
+		n = p0.Name
+	} else {
+		n = args[0]
+		var ok bool
+		p0, ok = pp[n]
+		if !ok {
+			return &influxdb.Error{
+				Code: influxdb.ENotFound,
+				Msg:  fmt.Sprintf("name %q is not found", b.name),
```
jsteenb2: would it make more sense to say `config ... not found`. Something like:

```go
Msg:  fmt.Sprintf("config %q is not found", b.name),
```

name feels a bit opaque to me 🤔

cmd/influx/config.go
```diff
@@ -123,10 +133,8 @@ func (b *cmdConfigBuilder) cmdCreateRunEFn(*cobra.Command, []string) error {
 	}
 
 	pp[b.name] = p
-	if p.Active {
-		if err := pp.Switch(b.name); err != nil {
-			return err
-		}
+	if err := pp.Switch(b.name); err != nil {
```
jsteenb2: is `Switch` the same as setting the config to active? If so, maybe renaming this to `SetActive` to make that more obvious would help

cmd/influx/config/config.go
```diff
@@ -14,15 +14,18 @@ import (
 
 // Config store the crendentials of influxdb host and token.
 type Config struct {
+	Name string `toml:"-" json:"-"`
```
jsteenb2: why are these not getting encoded?

cmd/influx/config/mock.go
```diff
@@ -2,8 +2,9 @@ package config
 
 // MockConfigService mocks the ConfigService.
 type MockConfigService struct {
```
jsteenb2: why is this exported?

cmd/influx/config/config.go
kelwang: we always export them as map[name]->config, this will prevent user edit any bad data.

cmd/influx/config/mock.go
kelwang: for cli level testing

cmd/influx/config.go
jsteenb2: are you planing on removing the active flag for this command? Otherwise it needs to be added to this config yah?

cmd/influx/config/config.go
jsteenb2: I want to make sure I understand what you are saying. You are saying that b/c the config has a name as the `[name]` object identifier in the `.toml` file, we have to remove it from being added to b/c they could edit one and not the other yah?

cmd/influx/config/config.go
```diff
@@ -14,78 +14,144 @@ import (
...
 type Configs map[string]Config
 
 // ConfigsService is the service to list and write configs.
 type ConfigsService interface {
-	WriteConfigs(pp Configs) error
-	ParseConfigs() (Configs, error)
+	CreateConfig(Config) (Config, error)
+	DeleteConfig(name string) (Config, error)
+	UpdateConfig(Config) (Config, error)
+	SwitchConfig(name string) (Config, error)
```
jsteenb2: can we name this something better? still feels ambiguous to me. something `SetActive` so its more obvious maybe?

cmd/influx/config/config.go
```diff
@@ -14,78 +14,144 @@ import (
...
+func (cfgs *Configs) Switch(name string) error {
+	pc := *cfgs
 	if _, ok := pc[name]; !ok {
 		return &influxdb.Error{
 			Code: influxdb.ENotFound,
 			Msg:  fmt.Sprintf(`config %q is not found`, name),
 		}
 	}
 	for k, v := range pc {
+		v.PreviousActive = v.Active
```
jsteenb2: why are all configs set previously active here?

cmd/influx/config/config.go
```diff
@@ -14,78 +14,144 @@ import (
...
+type store interface {
+	parsePreviousActive() (Config, error)
+	ListConfigs() (Configs, error)
+	writeConfigs(cfgs Configs) error
 }
 
 // Switch to another config.
-func (pp *Configs) Switch(name string) error {
-	pc := *pp
+func (cfgs *Configs) Switch(name string) error {
```
jsteenb2: I just noticed this, `Configs` is a `map`, which means it is a reference type.  The referencing on the line below, doesn't do what I suspect it is intended to be protecting against. It'd be better to remove the `*` on all methods here and to stop dereferencing. You can still mutate the backing hashmap even if it is dereferenced

cmd/influx/config/config.go
```diff
@@ -95,34 +161,161 @@ func (svc LocalConfigsSVC) WriteConfigs(pp Configs) error {
...
+		return Config{}, err
+	}
+	if _, ok := cfgs[cfg.Name]; ok {
+		return Config{}, &influxdb.Error{
+			Code: influxdb.EConflict,
+			Msg:  fmt.Sprintf("config %q already exists", cfg.Name),
+		}
+	}
+	cfgs[cfg.Name] = cfg
+	if err := cfgs.Switch(cfg.Name); err != nil {
```
jsteenb2: why are we implicitly switching the new config to be the active one?

cmd/influx/config/config.go
```diff
@@ -95,34 +161,161 @@ func (svc LocalConfigsSVC) WriteConfigs(pp Configs) error {
...
+	return ioutil.WriteFile(s.Path, b1.Bytes(), 0600)
+}
+
+// parsePreviousActive return the previous active config from the reader
+func (s baseRW) parsePreviousActive() (Config, error) {
+	return s.parseActiveConfig(false)
 }
 
 // ParseActiveConfig returns the active config from the reader.
 func ParseActiveConfig(r io.Reader) (Config, error) {
```
jsteenb2: curious why is this exported? Wont' the service encapsulate this?

cmd/influx/config/mock.go
jsteenb2: then move the mock to the CLI, doesn't belong at the produce side, belongs at the consumer side generally

cmd/influx/config/config.go
kelwang: yeah, that's one reason

cmd/influx/config/config.go
kelwang: what do you think about SwitchActive ?

cmd/influx/config/config.go
kelwang: if a config is active previously, then we can a switching, it's means it's previously active

cmd/influx/config/config.go
kelwang: it's been like that before.

cmd/influx/config/config.go
kelwang: Looks like it's been used by `getConfigFromDefaultPath`, looks like it will ignore some IO errors, so not something i feel comfortable to change this time. No enough testing coverage for this.

cmd/influx/config/config.go
jsteenb2: `SetActive`?

cmd/influx/config/config.go
jsteenb2: ah gotcha, so we don't care specifically which one is active, we just do the check for them all and go about our merry way 👍

cmd/influx/config/config.go
jsteenb2: I understand that, I'm asking why thought? Why are we setting it active implicitly? IF that is the case we should remove the `active` flag from teh cli create command. It efffectively does nothing atm

cmd/influx/config/config.go
jsteenb2: I'm not following the concern here 🤔

cmd/influx/config.go
```diff
@@ -104,40 +78,18 @@ func (b *cmdConfigBuilder) cmdCreate() *cobra.Command {
 }
```
jsteenb2, line 78: need to remove the active flag here if you decide to keep it implictly setting this to active when creating. I think that implicit switch is not something we want here, I'd rather see the active flag from the user take care of that instead of of forcing it every create...

If you decide to keep it implicit, this flag needs to be removed.

jsteenb2: @jademcgough what are your thoughts here?

cmd/influx/config/config.go
kelwang: the concern is that has nothing to do the `switch back` pr, behavior change should be a separated issue, with separated tests

cmd/influx/config.go
kelwang: I create a bug for this one, let's not make any behavior change in this pr. https://github.com/influxdata/influxdb/issues/17534

cmd/influx/config/config.go
kelwang: i already removed the active flag, to reflect the current behavior. For more discussion, let's move to https://github.com/influxdata/influxdb/issues/17534

cmd/influx/config.go
```diff
@@ -104,40 +78,19 @@ func (b *cmdConfigBuilder) cmdCreate() *cobra.Command {
...
-		return err
-	}
-
-	p := config.Config{
+	cfg, err := b.svc.CreateConfig(config.Config{
+		Name:   b.name,
 		Host:   b.url,
 		Token:  b.token,
 		Org:    b.org,
 		Active: b.active,
```
jsteenb2, line 86: 😃 looks good

PR #17498, opened 2020-03-31
test(launcher): e2e test memory limits

affo:
Testing memory limits from the launcher perspective.

Testing if memory gets properly deallocated using metrics.

This PR also adds setting memory limits for OSS via command  line

### DIFF
--- a/cmd/influxd/launcher/launcher_helpers.go
+++ b/cmd/influxd/launcher/launcher_helpers.go
@@ -78,12 +78,15 @@ func RunTestLauncherOrFail(tb testing.TB, ctx context.Context, args ...string) *
 }
 
 // Run executes the program with additional arguments to set paths and ports.
+// Passed arguments will overwrite/add to the default ones.
 func (tl *TestLauncher) Run(ctx context.Context, args ...string) error {
-	args = append(args, "--bolt-path", filepath.Join(tl.Path, bolt.DefaultFilename))
-	args = append(args, "--engine-path", filepath.Join(tl.Path, "engine"))
-	args = append(args, "--http-bind-address", "127.0.0.1:0")
-	args = append(args, "--log-level", "debug")
-	return tl.Launcher.Run(ctx, args...)
+	largs := make([]string, 0, len(args)+8)
+	largs = append(largs, "--bolt-path", filepath.Join(tl.Path, bolt.DefaultFilename))
+	largs = append(largs, "--engine-path", filepath.Join(tl.Path, "engine"))
+	largs = append(largs, "--http-bind-address", "127.0.0.1:0")
+	largs = append(largs, "--log-level", "debug")
+	largs = append(largs, args...)
+	return tl.Launcher.Run(ctx, largs...)
 }
 
 // Shutdown stops the program and cleans up temporary paths.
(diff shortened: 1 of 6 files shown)


cmd/influxd/launcher/query_test.go
```diff
@@ -105,52 +107,220 @@ func TestPipeline_WriteV2_Query(t *testing.T) {
...
-	// this query should error.
-	// spec.Resources = flux.ResourceManagement{
-	// 	MemoryBytesQuota: 100,
-	// }
+// This test:
+//  - initializes a default launcher and sets memory limits;
+//  - writes some data;
+//  - launches several queries that may exceed the memory limit;
+//  - verifies after each query run the used memory.
+func TestPipeline_QueryMemoryLimits_MultipleQueries(t *testing.T) {
```
affo: @GeorgeMac this is the flakey one

http/document_test.go
```diff
@@ -93,7 +93,7 @@ var (
 		},
 		Content: "content5",
 	}
-	doc6        = influxdb.Document{
+	doc6 = influxdb.Document{
```
affo, line 96: go fmt

query/control/memory.go
```diff
@@ -28,6 +28,22 @@ type memoryManager struct {
 	unlimited bool
 }
 
+func (m *memoryManager) getUnusedMemoryBytes() int64 {
+	return atomic.LoadInt64(&m.unusedMemoryBytes)
+}
+
+func (m *memoryManager) getUsedMemoryBytes() int64 {
+	return atomic.LoadInt64(&m.unusedMemoryBytes)
```
GeorgeMac, line 36: This method `getUsedMemoryBytes()` is returning `unusedMemoryBytes`

affo: thanks for catching that.

Certainly forgot to delete that. That's unused actually

affo: Now updated 👍

cmd/influxd/launcher/launcher_helpers.go
```diff
@@ -83,7 +83,16 @@ func (tl *TestLauncher) Run(ctx context.Context, args ...string) error {
 	args = append(args, "--bolt-path", filepath.Join(tl.Path, bolt.DefaultFilename))
 	args = append(args, "--engine-path", filepath.Join(tl.Path, "engine"))
 	args = append(args, "--http-bind-address", "127.0.0.1:0")
-	args = append(args, "--log-level", "debug")
+	logLevel := false
+	for _, arg := range args {
```
jsternberg, line 87: Instead of doing this, I think you can instead do something like:

```
largs := make([]string, 0, len(args)+8)
largs = append(largs, "--bolt-path", filepath.Join(tl.Path, bolt.DefaultFilename))
largs = append(largs, "--engine-path", filepath.Join(tl.Path, "engine"))
largs = append(largs, "--http-bind-address", "127.0.0.1:0")
largs = append(largs, "--log-level", "debug")
largs = append(largs, args...)
return tl.Launcher.Run(ctx, largs...)
```

If there are multiple command line parameters with the same name, the last one generally wins so it is easier to put the default arguments earlier in the command line.

affo: Definitely right, thank you 👍

PR #17501, opened 2020-03-31
fix(kv): make index walk use cursor instead of get

GeorgeMac:
This change swaps a `bucket.Get()` for a `bucket.ForwardCursor()` lookup for a single item in the `kv.Index.Walk()`.

This is in-order to address an issue with lots of individual `bucket.Get` calls in a single transaction when walking the index. This causes an issue with a particular implementation of `kv.Store`. Swapping it for a cursor aids this implementation by making the lookup act outside the transaction in this particular implementation of `kv.Store`.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/kv/index.go
+++ b/kv/index.go
@@ -251,7 +251,7 @@ func (i *Index) Delete(tx Tx, foreignKey, primaryKey []byte) error {
 
 // Walk walks the source bucket using keys found in the index using the provided foreign key
 // given the index has been fully populated.
-func (i *Index) Walk(tx Tx, foreignKey []byte, visitFn VisitFunc) error {
+func (i *Index) Walk(ctx context.Context, tx Tx, foreignKey []byte, visitFn VisitFunc) error {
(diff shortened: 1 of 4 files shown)


kv/index.go
```diff
@@ -549,15 +549,21 @@ func (i *Index) verify(ctx context.Context, store Store, includeMissingSource bo
...
 	}()
 
 	for ik, pk := indexCursor.Next(); ik != nil; ik, pk = indexCursor.Next() {
-		v, err := sourceBucket.Get(pk)
+		// TODO(george): this is a work-around as lots of calls to Get()
+		// on a transaction causes issues with a particular implementation
+		// of kv.Store.
+		// The use of a cursor on this store bypasses the transaction
+		// and gives us the access pattern we desire.
+		// Please do not change back to a bucket.Get().
```
affo: yeah, this also changes the consistency guarantees, but that's ok, we don't want strong consistency

PR #17507, opened 2020-03-31
feat(sampledata): Request buckets with demodata orgID

ebb-tide:
As per 6591, 
FindBuckets filters by orgID, limits the number of responses, and then filters by authorization. 
So if there is no filtering by orgID query parameter the limit may not return enough buckets to find one that user is authorized to read. 

In this PR we query demodata buckets, and peel off the orgID parameter if it exists, from the first returned bucket and supply the demodata orgID as a query parameter to the second getBuckets request to fetch demodata buckets user is authorized to read.

### DIFF
--- a/ui/src/cloud/apis/demodata.ts
+++ b/ui/src/cloud/apis/demodata.ts
@@ -5,7 +5,6 @@ import AJAX from 'src/utils/ajax'
 
 //Utils
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
-import {isDemoData} from 'src/cloud/utils/filterDemoData'
 
 //Types
 import {Bucket} from 'src/types'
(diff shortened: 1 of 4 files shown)


ui/src/cloud/apis/demodata.ts
```diff
@@ -75,17 +74,30 @@ export const deleteDemoDataBucketMembership = async (
   }
 }
 
-export const getDemoDataBucketsFromAll = async (): Promise<Bucket[]> => {
+export const fetchDemoDataBuckets = async (): Promise<Bucket[]> => {
   if (!isFlagEnabled('demodata')) return []
```
hoorayimhelping:
```suggestion
  if (!isFlagEnabled('demodata')) { return Promise.resolve([]) }
```

ui/src/cloud/apis/demodata.ts
```diff
@@ -75,17 +74,30 @@ export const deleteDemoDataBucketMembership = async (
   }
 }
 
-export const getDemoDataBucketsFromAll = async (): Promise<Bucket[]> => {
+export const fetchDemoDataBuckets = async (): Promise<Bucket[]> => {
```
hoorayimhelping: does this return a `Promise<Bucket[]>` or just `Bucket[]`?

ebb-tide: It's returning a Promise because it is an async function.

PR #17510, opened 2020-03-31
ci(promqltests): update promqltests go.mod file to be compatible with go 1.13

jsternberg:
The prometheus project doesn't adhere to the module suffix so any
attempt to use it as a library fails with go 1.13 or greater. The
workaround is to `go get` a specific commit revision that corresponds to
a tag and let go figure out a pseudo-version for it.

In this case, I updated the revision to the one pointed at by `v2.9.2`
since that is what the current `go.mod` file pointed at. I also updated
the go version to 1.13 inside of `go.mod` to be the same as influxdb.

See https://github.com/prometheus/prometheus/issues/6048 for details.

### DIFF
--- a/query/promql/internal/promqltests/go.mod
+++ b/query/promql/internal/promqltests/go.mod
@@ -1,6 +1,6 @@
 module github.com/influxdata/promqltests
 
-go 1.12
+go 1.13
 
 require (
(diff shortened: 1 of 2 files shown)


jsternberg:
cc @stuartcarnie since I know you asked about this. I can't seem to find any issue related to this problem, but this seemed to fix it for me.

PR #17512, opened 2020-03-31
fix(tsdb): Revert "fix: remove some unsafe marshalling to reduce risk of segfault"

benbjohnson:
This reverts commit 30dab0331033af22472df89852247b72e4074e32.

This is to test whether this commit causes a doubling of heap memory.


dgnorton:
@benbjohnson can you update the changelog before merging?

benbjohnson:
No problem. Updated.

PR #17515, opened 2020-03-31
fix: editing a table cell shows the proper values and respects changes

### LINKED ISSUE #17364: Table Cell editor overlay doesn't respect cell options
__Steps to reproduce:__

1. Create a dashboard cell that is a `table` visualization
1. Add custom names to the columns in the `Customize` section
1. Save the cell
1. Observe the correct cell name in dashboard
1. Edit the cell again
1. Observe the incorrect cell names in the overlay

__Expected behavior:__
The custom column names are preserved

__Actual behavior:__
The custom column names are gone and the names revert back to the names in the returned data when editing.

__Environment info:__

* This was in the cloud product
* currently happening in `master`

hoorayimhelping:
Closes #17364

Due to a confusing `falsy` check in a reducer, editing a dashboard cell showed the defaults or the query results for any changed properties. This fixes that.

### Steps to Reproduce
1. Create a dashboard cell that is a `table` visualization
1. Add custom names to the columns in the `Customize` section
1. Save the cell
1. Observe the correct cell name in dashboard
1. Edit the cell again
1. Observe the incorrect cell names in the overlay

This fix changes it so that the final step shows the correct value.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/timeMachine/actions/queries.ts
+++ b/ui/src/timeMachine/actions/queries.ts
@@ -173,7 +173,7 @@ export const executeQueries = (dashboardID?: string) => async (
     // https://github.com/influxdata/idpe/issues/6240
     const variableAssignments = getVariableAssignments(getState())
 
-    const startTime = Date.now()
+    const startTime = window.performance.now()
 
     pendingResults.forEach(({cancel}) => cancel())
 
@@ -186,7 +186,7 @@ export const executeQueries = (dashboardID?: string) => async (
     })
 
     const results = await Promise.all(pendingResults.map(r => r.promise))
-    const duration = Date.now() - startTime
+    const duration = window.performance.now() - startTime
 
     let statuses = [[]] as StatusRow[][]
     if (checkID) {
--- a/ui/src/timeMachine/reducers/index.test.ts
+++ b/ui/src/timeMachine/reducers/index.test.ts
@@ -7,6 +7,8 @@ import {
   timeMachinesReducer,
 } from 'src/timeMachine/reducers'
 
+import {RemoteDataState, TableViewProperties} from 'src/types'
+
 describe('the Time Machine reducer', () => {
   describe('setting the default aggregateFunctionType', () => {
     const store = createStore(timeMachinesReducer, initialState())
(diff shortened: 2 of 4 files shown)


ui/src/timeMachine/actions/queries.ts
```diff
@@ -173,7 +173,7 @@ export const executeQueries = (dashboardID?: string) => async (
     // https://github.com/influxdata/idpe/issues/6240
     const variableAssignments = getVariableAssignments(getState())
 
-    const startTime = Date.now()
+    const startTime = window.performance.now()
```
hoorayimhelping, line 176: https://developer.mozilla.org/en-US/docs/Web/API/Performance/now

ui/src/timeMachine/reducers/index.ts
```diff
@@ -281,7 +281,7 @@ export const timeMachineReducer = (
         draftState.queryResults.status = status
         draftState.queryResults.errorMessage = errorMessage
 
-        if (files) {
+        if (files && files.length) {
```
hoorayimhelping, line 284: [`files`](https://github.com/influxdata/influxdb/blob/master/ui/src/timeMachine/actions/queries.ts#L47) is typed as:

```ts
files?: string[]
```
but iin these flows, it's passed in as `[]`

ui/src/timeMachine/reducers/index.test.ts
```diff
@@ -189,4 +191,37 @@ describe('the Time Machine reducer', () => {
...
+          files: [],
+          fetchDuration: 234,
+        },
+      })
+
+      const endState = store.getState()
+      expect(
+        (endState.timeMachines[endState.activeTimeMachineID].view
+          .properties as TableViewProperties).fieldOptions[0].displayName
+      ).toBe(tableViewProperties.fieldOptions[0].displayName)
```
121watts, line 224:
```suggestion
     const tvp = endState.timeMachines[endState.activeTimeMachineID].view
          .properties as TableViewProperties
          
     const actual = tvp.fieldOptions[0].displayName
     const expected = tableViewProperties.fieldOptions[0].displayName
      
      expect(actual).toBe(expected)
```
or some such cleaner thing to look at

hoorayimhelping:
@121watts do you think it's worth blocking getting a fix out now to write an e2e test? I'd prefer to merge this to get it out and working, then write an e2e test to address it since there's already a test at the reducer level.

121watts:
I think the `e2e` is more valuable than the unit test.  I'll let you make the call of wether you want to merge without it.

hoorayimhelping:
Created a followup issue: #17539

PR #17525, opened 2020-04-01
fix(CORS): added User-Agent to allowed CORS headers

bednar:
`Access-Control-Allow-Headers` header should also contains a `User-Agent` because we want to monitor version of clients - see https://github.com/influxdata/influxdb-client-js/issues/152

cc https://github.com/influxdata/influxdb-client-js/issues/160

Screenshot 2020-04-01 at 09 30 54
Screenshot 2020-04-01 at 09 30 18

- [x] Rebased/mergeable
- [x] Tests pass


rawkode:
Thanks @bednar

PR #17532, opened 2020-04-01
fix(tsdb): Replace panic with error while de/encoding corrupt data

### LINKED ISSUE #17440: Error instead of Panic while Encoding corrupted TSM files
Meanwhile, TSM Encoding/Decoding `panic` when encountering corrupted data. 

For example [here](https://github.com/influxdata/influxdb/blob/90a529e6f0b4a4fb53f4d16c0c5e1e9927050e94/tsdb/tsm1/encoding.go#L182)

There will be always corrupted data, so instead of `panic` we return an `error`.

Elbehery:
fixes #17440

While encoding or decoding corrupt data, the current behaviour is to `panic`.
This commit replaces the `panic` with `error` to be propagated up to the calling `iterator`.

To avoid overwriting other `error`, iterators now wraps a `TSMErrors` which contains ALL the encountered errors. TSMErrors itself implements `Error()`, the returned string contains all the error msgs, separated by `,` delimiter.

### DIFF
--- a/tsdb/engine/tsm1/file_store.go
+++ b/tsdb/engine/tsm1/file_store.go
@@ -924,7 +924,9 @@ func (f *FileStore) BlockCount(path string, idx int) int {
 				}
 			}
 			_, _, _, _, _, block, _ := iter.Read()
-			return BlockCount(block)
+			// on Error, BlockCount(block) returns 0 for cnt
+			cnt, _ := BlockCount(block)
+			return cnt
 		}
 	}
 	return 0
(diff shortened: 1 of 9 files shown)


tsdb/engine/tsm1/file_store.go
```diff
@@ -924,7 +924,9 @@ func (f *FileStore) BlockCount(path string, idx int) int {
 				}
 			}
 			_, _, _, _, _, block, _ := iter.Read()
-			return BlockCount(block)
+			// on Error, BlockCount(block) returns 0 for cnt
+			cnt, _ := BlockCount(block)
```
jacobmarble, line 928: Missing the error handler here

Elbehery: this function returns `0` for other edge cases, the caller should expect the `0` as failure. 
I think we should collect the errors somewhere but not everywhere !!

jacobmarble:
> This looks great. I think `go fmt` will fix up one or two small mistakes.

OK it was a spaces-vs-tabs problem in the .tmpl file. Looks like benbjohnson/tmpl does format after templating, so it wasn't an issue in the .gen.go file. I pushed a one-line commit.

Elbehery:
alchemy please merge

Issue #17542, opened 2020-04-01
Remove FancyScrollbar and react-custom-scrollbars


alexpaxton:
`FancyScrollbar` (a wrapper for the `react-custom-scrollbars` library) should be factored out and replaced with `DapperScrollbars` from Clockface (a wrapper for `react-scrollbars-custom`).

`react-custom-scrollbars` is no longer maintained while `react-scrollbars-custom` is maintained and much easier to work with

PR #17543, opened 2020-04-01
chore(ui): remove most fancy scrollbars

alexpaxton:
Completes most of the work described in #17542 however there are a few instances of `FancyScrollbar` that make use of a prop called `setScrollTop` which there is no equivalent for in `DapperScrollbars`. This issue is for tracking the work in Clockface: https://github.com/influxdata/clockface/issues/476

### Changes

- Replace all but a few instances of `FancyScrollbar` with `DapperScrollbars`
- Update styling of affected components
- Delete a few components that aren't being used

### Preview

Screen Shot 2020-04-01 at 10 02 15 AM
Screen Shot 2020-04-01 at 10 00 20 AM
Screen Shot 2020-04-01 at 9 59 37 AM
Screen Shot 2020-04-01 at 9 57 54 AM
Screen Shot 2020-04-01 at 9 57 47 AM
Screen Shot 2020-04-01 at 10 17 19 AM

- [x] Rebased/mergeable
- [x] Tests pass


alexpaxton:
@Asalem1 The ultimate goal is to remove them all, however I can't do that just yet cuz `DapperScrollbars` doesn't offer the same functionality (yet) as `FancyScrollbar`. This bifurcates the work across 2 PRs but at least most of it gets tackled here

ui/src/shared/components/CodeSnippet.scss
```diff
@@ -4,27 +4,32 @@
...
 .code-snippet--scroll {
   min-height: 26px !important;
-  background-color: $g4-onyx;
+  max-height: 200px;
+  border-radius: $cf-radius-sm;
+  overflow: hidden;
 }
 
 .code-snippet--text {
   font-size: 13px;
```
Asalem1, line 22: is there a font variable we should be using here?

ui/src/shared/components/CodeSnippet.scss
```diff
@@ -4,27 +4,32 @@
...
+  background-color: $g2-kevlar;
+  margin-bottom: $cf-marg-b;
 }
 
 .code-snippet--scroll {
   min-height: 26px !important;
-  background-color: $g4-onyx;
+  max-height: 200px;
+  border-radius: $cf-radius-sm;
+  overflow: hidden;
```
Asalem1, line 18: If we set the overflow to hidden here, aren't we going to confine the DapperScrolls to an artificial div size that won't render all the data?

ui/src/shared/components/views/Markdown.scss
```diff
@@ -3,32 +3,13 @@
...
-  pointer-events: none;
-  content: '';
-  position: absolute;
-  bottom: 0;
-  left: 0;
-  width: 100%;
-  @include gradient-v(rgba($g3-castle, 0), $g3-castle);
-  z-index: 1;
+  padding: $cf-marg-b + $cf-marg-a 26px;
+  font-size: 14px;
```
Asalem1, line 12: The previous font-size was `13px`, and we also have other instances where the app had been using `13pt` font, should we transition the font-sizes from 13->14 / should we set variables to do that?

ui/src/shared/components/CodeSnippet.tsx
```diff
@@ -28,18 +28,17 @@ class CodeSnippet extends PureComponent<Props> {
 
     return (
       <div className="code-snippet" data-testid={testID}>
-        <FancyScrollbar
+        <DapperScrollbars
           autoHide={false}
-          autoHeight={true}
-          maxHeight={400}
+          autoSizeHeight={true}
```
Asalem1, line 33: should we set the maxHeight here?

ui/src/telegrafs/components/TelegrafOutputOverlay.test.tsx
```diff
@@ -2,6 +2,11 @@
 import React from 'react'
 import {render} from 'react-testing-library'
 
+// NOTE: stubbing is required here as the CopyButton component
```
Asalem1, line 5: I'd be curious to know why this was moved out?

alexpaxton: @hoorayimhelping moved it because it previously was failing. There was a strange error that arose when the underlying component used `DapperScrollbars` instead of `FancyScrollbars`

ui/src/shared/components/CodeSnippet.tsx
alexpaxton: I can never decide the best way to handle these things. It's either set it inline using a style object (which I don't like cuz it's separate from the other styles), or create a class for it (which sometimes means making a stylesheet with a single class).

🤷‍♂

ui/src/shared/components/CodeSnippet.scss
alexpaxton: The scroll area will grow to fit the size here so no weird behavior happening. The hidden overflow here ensures the corners are always rounded

ui/src/shared/components/CodeSnippet.scss
alexpaxton: Yes! Good looking out. We have one in the clockface variables

ui/src/shared/components/views/Markdown.scss
alexpaxton: The `14px` deviates but is much nicer to read. Maybe I could set the size using `em` or `rem` so at least it's based on the root font size instead of a fixed number

PR #17547, opened 2020-04-01
refactor(kv): set the default session length when not specified

jsternberg:
At the moment, the default session length only gets set to the default
when no service config is specified. So if a service config is used and
it does not set a value for the session length, then the session length
will be zero.

This modifies the creation of the kv service so that it will set the
default session length if the session length is set to zero (the default
value) instead of only doing it when no service config is specified.


jsternberg:
This change is intended to consolidate a change that was made on the `feat/use-algo-w` branch and another change that was made on master.

PR #17549, opened 2020-04-01
fix: filter variables to only those you are using

### LINKED ISSUE #17513: Variables on dashboards messed up in Safari
I'm able to reproduce on cloud2.
image

drdelambre:
Closes #17513

i let all the variables in. @russorat only wants some of the variables in. the good ones. the kind you can change, and the page works more gooderer.

### DIFF
--- a/ui/src/dashboards/components/variablesControlBar/VariablesControlBar.tsx
+++ b/ui/src/dashboards/components/variablesControlBar/VariablesControlBar.tsx
@@ -166,7 +166,9 @@ const mstp = (state: AppState): StateProps => {
   return {
     variables: filterUnusedVars(
       variables,
-      Object.values(state.resources.views.byID)
+      Object.values(state.resources.views.byID).filter(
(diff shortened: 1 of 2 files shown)


ui/src/dashboards/components/variablesControlBar/VariablesControlBar.tsx
```diff
@@ -163,7 +164,12 @@ const mstp = (state: AppState): StateProps => {
   } = state
 
   return {
-    variables,
+    variables: filterUnusedVars(
+      variables,
+      Object.values(state.resources.views.byID).filter(
+        v => v.dashboardID === dashboardID
```
hoorayimhelping, line 170:
```suggestion
        variable => variable.dashboardID === dashboardID
```
just to improve searchability

drdelambre: 👍

PR #17551, opened 2020-04-01
fix: revert token change to add user ID

gavincabbage:
Reverts https://github.com/influxdata/influxdb/pull/17535


gavincabbage:
I'm going to merge without waiting for CI on this because it reverts something I added earlier today that will might break tasks if allowed to be deployed.

PR #17552, opened 2020-04-01
fix(regression): function inserts now happen on new line

### LINKED ISSUE #17282: Clicking on a function inserts it directly where the cursor is, breaking the query
in previous versions, when clicking a function in the functions list to insert it into the script, it would insert it on the next line without breaking the script. now it inserts it directly where the cursor is, breaking the script. Most likely related to https://github.com/influxdata/influxdb/pull/17028

Kapture 2020-03-15 at 15 09 26

Asalem1:
Closes #17282

### Problem

This bug is a regression that stemmed from our earlier transition from codemirror to monaco. In particular, the issue was that some of the logic that had previously been used to determine the insert line of a function was not integrated into the monaco functionality.

### Solution

Integrated a few logical steps to calculate where a function should be placed when selecting an aggregate function.

func-insert-regression-fix

- [x] Rebased/mergeable

### DIFF
--- a/ui/src/timeMachine/components/TimeMachineFluxEditor.tsx
+++ b/ui/src/timeMachine/components/TimeMachineFluxEditor.tsx
@@ -75,7 +75,7 @@ const TimeMachineFluxEditor: FC<Props> = ({
     const p = editorInstance.getPosition()
     const scriptLines = activeQueryText.split('\n')
 
-    let insertLineNumber = getInsertLineNumber(p.lineNumber, scriptLines)
+    const insertLineNumber = getInsertLineNumber(p.lineNumber, scriptLines)
 
     // sets the range based on the current position
     let range = new window.monaco.Range(
--- a/ui/src/timeMachine/components/TimeMachineFluxEditor.tsx
+++ b/ui/src/timeMachine/components/TimeMachineFluxEditor.tsx
@@ -77,21 +77,23 @@ const TimeMachineFluxEditor: FC<Props> = ({
 
     const insertLineNumber = getInsertLineNumber(p.lineNumber, scriptLines)
 
+    const defaultColumnPosition = 1 // beginning column of the row
+
     // sets the range based on the current position
     let range = new window.monaco.Range(
       insertLineNumber, // the row beneath the cursor
-      1, // beginning column of the row
+      defaultColumnPosition,
       insertLineNumber,
-      1
+      defaultColumnPosition
     )
     // edge case for when user toggles to the script editor
     // this defaults the cursor to the initial position (top-left, 1:1 position)
     const [currentRange] = editorInstance.getVisibleRanges()
     // Determines whether the new insert line is beyond the current range
-    let insertOnLastLine = insertLineNumber > currentRange.endLineNumber
-    if (p.lineNumber === 1 && p.column === 1) {
+    let shouldInsertOnLastLine = insertLineNumber > currentRange.endLineNumber
+    if (p.lineNumber === 1 && p.column === defaultColumnPosition) {
       // adds the function to the end of the query
-      insertOnLastLine = true
+      shouldInsertOnLastLine = true
       range = new window.monaco.Range(
         currentRange.endLineNumber + 1,
         p.column,
@@ -106,7 +108,7 @@ const TimeMachineFluxEditor: FC<Props> = ({
         text: formatFunctionForInsert(
           func.name,
           func.example,
-          insertOnLastLine
+          shouldInsertOnLastLine
         ),
       },
     ]
--- a/ui/src/timeMachine/components/TimeMachineFluxEditor.tsx
+++ b/ui/src/timeMachine/components/TimeMachineFluxEditor.tsx
@@ -56,20 +56,42 @@ const TimeMachineFluxEditor: FC<Props> = ({
     onSetActiveQueryText(editorInstance.getValue())
   }
 
+  const getInsertLineNumber = (
+    currentLineNumber: number,
+    scriptLines: string[]
+  ): number => {
+    const currentLine =
+      scriptLines[currentLineNumber] || scriptLines[scriptLines.length - 1]
+
+    // Insert on the current line if its an empty line
+    if (!currentLine.trim()) {
+      return currentLineNumber
+    }
+
+    return currentLineNumber + 1
+  }
+
   const handleInsertFluxFunction = (func: FluxToolbarFunction): void => {
     const p = editorInstance.getPosition()
+    const scriptLines = activeQueryText.split('\n')
+
+    let insertLineNumber = getInsertLineNumber(p.lineNumber, scriptLines)
+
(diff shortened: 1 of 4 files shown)


ui/src/timeMachine/components/TimeMachineFluxEditor.tsx
```diff
@@ -56,31 +56,58 @@ const TimeMachineFluxEditor: FC<Props> = ({
...
+      insertLineNumber, // the row beneath the cursor
+      1, // beginning column of the row
+      insertLineNumber,
+      1
     )
     // edge case for when user toggles to the script editor
     // this defaults the cursor to the initial position (top-left, 1:1 position)
+    const [currentRange] = editorInstance.getVisibleRanges()
+    // Determines whether the new insert line is beyond the current range
+    let insertOnLastLine = insertLineNumber > currentRange.endLineNumber
```
hoorayimhelping, line 91: strongly recommend this variable be changed to something like `shouldInsertOnLastLine`.

right now, there's a variable called `insertLineNumber` that represents the line to insert a function on, and there's this, called `insertOnLastLine` which is a boolean. The similar names with different values and different semantic meanings is really confusing. Either of these names could also be understood to be verbs, and so someone glancing at this might think they're methods, which is another layer of confusion. These are pretty well named variables other than they're a little bit ambiguous in their intended usage. I don't know a clever way to be precise, so I usually just do it with a long variable name. `lineNumberToInsertAt` vs `insertLineNumber`.

ui/src/timeMachine/utils/insertFunction.test.ts
```diff
@@ -18,10 +18,10 @@ describe('insertFunction', () => {
   test('formatFunctionForInsert', () => {
```
hoorayimhelping, line 18: these tests are okay. they pass. but they're kind of hard to read and they're kind of barebones. the power of rspec-like tests (which jest is) is that we can describe things in human language for people to read. `insertFunction generateImport` is kind of meh compared to something like [this](https://github.com/influxdata/influxdb/blob/master/ui/src/variables/utils/getWindowVars.test.ts) where it's much more natural to read: `getWindowPeriodVariable should return null when passed an empty query string`. 

writing sentences out like this also clarifies our thinking by forcing us to state in words what the code we wrote is doing.

ui/src/timeMachine/components/TimeMachineFluxEditor.tsx
```diff
@@ -56,31 +56,58 @@ const TimeMachineFluxEditor: FC<Props> = ({
...
+
+    const insertLineNumber = getInsertLineNumber(p.lineNumber, scriptLines)
+
     // sets the range based on the current position
     let range = new window.monaco.Range(
-      p.lineNumber,
-      p.column,
-      p.lineNumber,
-      p.column
+      insertLineNumber, // the row beneath the cursor
```
hoorayimhelping, line 82: is this comment true? if there's nothing on the line the cursor is on, it will remain on the current line. i think you should remove the comment

ui/src/timeMachine/components/TimeMachineFluxEditor.tsx
```diff
@@ -56,31 +56,58 @@ const TimeMachineFluxEditor: FC<Props> = ({
...
+      1, // beginning column of the row
+      insertLineNumber,
+      1
```
hoorayimhelping, line 85: can you add a comment here explaining what this magic number is, or store it into a variable that's named what `1` represents?

ui/src/timeMachine/components/TimeMachineFluxEditor.tsx
Asalem1: It is true.

ui/src/timeMachine/components/TimeMachineFluxEditor.tsx
Asalem1: I think it might be a little redundant. The gist of the `Range` monaco object is to define the Range of the editor:

https://microsoft.github.io/monaco-editor/api/classes/monaco.range.html

In this case the `1` represents the start and end position of the column (i.e. the beginning of the column). We can add another comment, but it might be excessive don't you think?

ui/src/timeMachine/utils/insertFunction.ts
```diff
@@ -14,13 +14,18 @@ const functionRequiresNewLine = (funcName: string): boolean => {
 
 export const formatFunctionForInsert = (
   funcName: string,
-  fluxFunction: string
-) => {
+  fluxFunction: string,
+  newLine: boolean
```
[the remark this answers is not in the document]
Asalem1: Ya, it's only used in this one place, but the reason for that added boolean is to handle specific edge cases in cases where the cursor position is at the default position (1:1), or on the last line. I hear what you're saying, but besides creating a specific function that would be called conditionally based on a boolean, I'm not sure what other solution there would be. And since conditionally calling a function seems just as fishy to me, I'd love to hear your thoughts

ui/src/timeMachine/components/TimeMachineFluxEditor.tsx
hoorayimhelping: i think the term "the row beneath" is confusing in this context. it could mean "the line under the line where the cursor is" e.g. cursor is on line 1, place it on line 2 (how i originally interpreted it) or "directly where the cursor is, directly under it" e.g. cursor is on line 1, place the function on line 1 directly under the cursor. this comment added to my confusion, rather than clarifying my understanding.

Asalem1: Gotcha. I'll update the comment to reflect the former string that you said

Asalem1: Since that particular property is only determining the row (and not the column), it's only concerned with the `row` context. In that sense, the column has no bearing on this value so it's only referring to the row

ui/src/timeMachine/utils/insertFunction.ts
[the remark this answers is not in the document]
hoorayimhelping: i'd like to pair up to talk about this, but my calendar just filled up today so i have to type this out.

>I'm not sure what other solution there would be.

the solution is to look at the design of the software as a whole and change the pieces that don't make sense.

right now, there is logic split between the component and this function in utils. `formatFunctionForInsert` can't operate without external logic to tell it whether it should have newlines or not - that logic, in the form of an argument, is handled elsewhere. it makes the function tightly coupled to the code that calls it. what's the point of having a function, which is meant to promote encapsulation and reuse, if you're not encapsulating or enabling reuse? it just becomes a misdirection that you have to look up.

the solution is to consolidate the logic. either keep the logic in the component the way it is now and remove these methods, or make helper methods (which imply reuse) which contain the logic of figuring out where to put the flux function. if it was me, i'd write some helper methods in the component and leave them there until something else needs to use them.

PR #6580, opened 2016-05-08
Fix config file loading to read from a file

jsternberg:
Accidentally used `FromToml()` when trying to read from a file path
rather than `FromTomlFile()` which will read the file and then call
`FromToml()`.

Fixes #6578.

PR #17553, opened 2020-04-01
feat(sampledata): Add demodata bucket type

ebb-tide:
Closes #6580

Create DemoBucket type in the UI and apply `type: demodata` to buckets that are returned from demodata Org. Create a new DemoDataBucketCard to display these buckets in the BucketsTab. 
Also some clean up along the way


ui/src/buckets/components/DemoDataBucketCard.tsx
```diff
@@ -0,0 +1,48 @@
+// Libraries
+import React, {FC} from 'react'
+import {withRouter, WithRouterProps} from 'react-router'
+import _ from 'lodash'
```
drdelambre, line 4: get this out of here!

PR #17555, opened 2020-04-01
fix(ui): change index.html <title> to InfluxDB Cloud

mnorriswalker:
fix: minor text change for name of product from InfluxDB 2.0 to InfluxDB Cloud in index.html

Trying to change so the name of the product aligns between the page and browser tab <title>
image


russorat:
I think this should say InfluxDB Cloud in cloud, but just InfluxDB in OSS.

Issue #17434, opened 2020-03-25
Applying package with a stack is idempotent


jsteenb2:
As a user,
I want to create resources with pkgs and a stack,
so that I don't have to clean up duplicates resources and declare a state of the world

This card is to add idempotency to pkger via a stack. A stack is the state that tracks the side effects of creating a package. Using this, we want to be able to apply packages idempotently and always end up with the state declared in the package.

idempotent means the following:

> the state of the world is always accurate to the package. resources are uniquely identified via its kind and `metadata.name` fields. Any and all secondary resources created (i.e. label mappings) are accurate to the package.

**AC**

- [x] stacks are used when a stack ID is provided to the dry run/apply functionality
   - [x] extends HTTP `POST /api/v2/packages` with `stack_id` field
- [x] package payload is optional if stack_id is provided and stack has URL(s) associated with it
- [x] stacks are updated when side effects are created, if nothing changed between state of the world and package stack is not updated.
- [x] resources are applied idempotently
   - [x] labels
   - [x] buckets
   - [x] checks
   - [x] dashboards
   - [x] notification endpoints
   - [x] notification rules 
   - [x] tasks
   - [x] telegraf
   - [x] variable
- [x] when an association is removed in pkg, is also removed from platform
- [x] support use cases where elements have been deleted by user

PR #17557, opened 2020-04-01
feat(pkger): extend labels diff and summary types to include PkgName and Remove

jsteenb2:
the pkgName refers to the unqiue label resource within the pkg and the
Remove field indicates the label will be removed when applying the pkg

references: #17434 

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influx/pkg.go
+++ b/cmd/influx/pkg.go
@@ -715,29 +715,38 @@ func (b *cmdPkgBuilder) printPkgDiff(diff pkger.Diff) error {
 		return bb
 	}
 
-	tablePrintFn := b.tablePrinterGen()
 	if labels := diff.Labels; len(labels) > 0 {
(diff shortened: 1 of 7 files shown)


cmd/influx/pkg.go
```diff
@@ -1112,6 +1123,7 @@ func (d *diffPrinter) Render() {
 
 	d.setFooter()
 	d.writer.Render()
+	fmt.Fprintln(d.w)
```
kelwang, line 1126: ? does it do anything

jsteenb2: I forgot to add this in last PR. What it does is extend the margin around the table print so that the table has a buffer of 1 newline. Makes it so that when printing a table it takes care of styling itself, including margins 🤷‍♂. Similar to how css has margin but way more hacky lol

before newline buffer:
image

with newline buffer:
image

PR #17563, opened 2020-04-02
fix: Invariant violated: mixed block types for a single series

stuartcarnie:
The root cause is that the `Unsigned` data type has no representation in the `valueType` function in the cache and falls back to the default case of 0.

0 is also a sentinel value in the `entry#add` function that will result in skipping the value type check.

It therefore is possible that unsigned values followed by some other data type is stored in the cache.

It is suspected that the write may be rejected before reaching the cache, and therefore may not occur in practice. Specifically, the series file stores the data types on a per-series basis and would
reject the write.

This commit turns the value types into explicit constants and ensures all existing block types are represented. In addition, it adds a mapping function to convert these to a known Block type,
which will be used by the `MeasurementFields` schema request to determine the type of a series in the cache.

### DIFF
--- a/tsdb/tsm1/encoding.go
+++ b/tsdb/tsm1/encoding.go
@@ -25,6 +25,9 @@ const (
 	// BlockUnsigned designates a block encodes uint64 values.
 	BlockUnsigned = byte(4)
 
+	// blockUndefined represents an undefined block type value.
+	blockUndefined = BlockUnsigned + 1
+
 	// encodedBlockHeaderSize is the size of the header for an encoded block.  There is one
 	// byte encoding the type of the block.
 	encodedBlockHeaderSize = 1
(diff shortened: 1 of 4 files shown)


tsdb/tsm1/encoding.go
```diff
@@ -25,6 +25,9 @@ const (
 	// BlockUnsigned designates a block encodes uint64 values.
 	BlockUnsigned = byte(4)
 
+	// blockUndefined represents an undefined block type value.
+	blockUndefined = BlockUnsigned + 1
```
dgnorton, line 29: Can't think of a scenario but wondering if it would be safer to set `blockUndefined = byte(255)`, a value that won't be used for a block type in the future.

stuartcarnie: I thought about that, but these constants currently must fit in 3 bits so it can be used for this fast look up:

https://github.com/influxdata/influxdb/pull/17563/files#diff-b44a6551f36995f10c237a575eec2bf7R737

It doesn't appear we're going to add more in this version of storage, so this constraint should hold true.

PR #17589, opened 2020-04-03
feat(tsm1): Implementation of MeasurementFields schema API

### LINKED ISSUE #17560: Implement MeasurementFields schema API
## What

Implement the `MeasurementFields` API as prescribed in the **Introduce measurement-based storage schema APIs** design.

stuartcarnie:
The primary purpose of this PR is to provide an implementation of the `MeasurementFields` API. However, it also addresses a data race for the existing `TagKeys` and `TagValues` APIs.

## Data Race

The `TagKeys` and `TagValues` APIs use the `Cache#ApplyEntryFn`. The `ApplyEntryFn` API expects a callback function, which passes a cache `entry` as the only argument. Accessing the `values` field of the `entry` struct is a data race, as it may be mutated by other goroutines performing writes, snapshots, etc. See below for the fix.

## `Cache` improvements

This PR includes a commit which add a single API to the `Cache` data structure to address a data race with the existing `TagKeys` and `TagValues` APIs.

Specifically, `Cache` and `entry` provide `AppendTimestamps`, which returns a slice of the current timestamps for the given key or entry respectively. To reduce allocations, `AppendTimestamps` takes a slice which it will append the values. As noted in the documentation, it is the responsibility of the caller to sort and deduplicate the values, if required. This function is used by `TagKeys`, `TagValues` and `MeasurementFields`.

The `cursors.TimestampArray` type was extended to permit use of the `sort.Sort` API.

Closes #17560

### DIFF
--- a/tsdb/tsm1/cache_entry.go
+++ b/tsdb/tsm1/cache_entry.go
@@ -117,6 +117,19 @@ func (e *entry) size() int {
 	return sz
 }
 
+// AppendTimestamps appends ts with the timestamps from the entry.
+func (e *entry) AppendTimestamps(ts []int64) []int64 {
+	e.mu.RLock()
+	defer e.mu.RUnlock()
+	n := e.values.Len()
+	if n > 0 {
+		for i := range e.values {
+			ts = append(ts, e.values[i].UnixNano())
+		}
+	}
+	return ts
+}
+
 // InfluxQLType returns for the entry the data type of its values.
 func (e *entry) InfluxQLType() (influxql.DataType, error) {
 	e.mu.RLock()
--- a/tsdb/tsm1/engine_schema.go
+++ b/tsdb/tsm1/engine_schema.go
@@ -102,6 +102,8 @@ func (e *Engine) tagValuesNoPredicate(ctx context.Context, orgID, bucketID influ
 		return cursors.NewStringSliceIteratorWithStats(nil, stats), ctx.Err()
 	}
 
+	var ts cursors.TimestampArray
+
 	// With performance in mind, we explicitly do not check the context
 	// while scanning the entries in the cache.
 	tsmKeyprefixStr := string(tsmKeyPrefix)
@@ -122,12 +124,18 @@ func (e *Engine) tagValuesNoPredicate(ctx context.Context, orgID, bucketID influ
 			return nil
 		}
 
-		stats.ScannedValues += entry.values.Len()
-		stats.ScannedBytes += entry.values.Len() * 8 // sizeof timestamp
+		ts.Timestamps = entry.AppendTimestamps(ts.Timestamps[:0])
+		if ts.Len() > 0 {
+			sort.Sort(&ts)
 
-		if entry.values.Contains(start, end) {
-			tsmValues[string(curVal)] = struct{}{}
+			stats.ScannedValues += ts.Len()
+			stats.ScannedBytes += ts.Len() * 8 // sizeof timestamp
+
+			if ts.Contains(start, end) {
+				tsmValues[string(curVal)] = struct{}{}
+			}
 		}
+
 		return nil
 	})
 
@@ -207,6 +215,7 @@ func (e *Engine) tagValuesPredicate(ctx context.Context, orgID, bucketID influxd
 		tags   models.Tags
 		keybuf []byte
 		sfkey  []byte
+		ts     cursors.TimestampArray
 	)
 
 	for i := range keys {
@@ -230,15 +239,23 @@ func (e *Engine) tagValuesPredicate(ctx context.Context, orgID, bucketID influxd
 			continue
 		}
 
-		keybuf = models.AppendMakeKey(keybuf[:0], orgBucketEsc, tags)
+		// orgBucketEsc is already escaped, so no need to use models.AppendMakeKey, which
+		// unescapes and escapes the value again. The degenerate case is if the orgBucketEsc
+		// has escaped values, causing two allocations per key
+		keybuf = append(keybuf[:0], orgBucketEsc...)
+		keybuf = tags.AppendHashKey(keybuf)
 		sfkey = AppendSeriesFieldKeyBytes(sfkey[:0], keybuf, tags.Get(models.FieldKeyTagKeyBytes))
 
-		values := e.Cache.Values(sfkey)
-		stats.ScannedValues += values.Len()
-		stats.ScannedBytes += values.Len() * 8 // sizeof timestamp
+		ts.Timestamps = e.Cache.AppendTimestamps(sfkey, ts.Timestamps[:0])
+		if ts.Len() > 0 {
+			sort.Sort(&ts)
 
-		if values.Contains(start, end) {
-			tsmValues[string(curVal)] = struct{}{}
+			stats.ScannedValues += ts.Len()
+			stats.ScannedBytes += ts.Len() * 8 // sizeof timestamp
+
+			if ts.Contains(start, end) {
+				tsmValues[string(curVal)] = struct{}{}
+			}
 			continue
 		}
 
@@ -378,6 +395,8 @@ func (e *Engine) tagKeysNoPredicate(ctx context.Context, orgID, bucketID influxd
 		return cursors.NewStringSliceIteratorWithStats(nil, stats), ctx.Err()
 	}
 
+	var ts cursors.TimestampArray
+
 	// With performance in mind, we explicitly do not check the context
 	// while scanning the entries in the cache.
 	tsmKeyprefixStr := string(tsmKeyPrefix)
@@ -393,12 +412,18 @@ func (e *Engine) tagKeysNoPredicate(ctx context.Context, orgID, bucketID influxd
 			return nil
 		}
 
-		stats.ScannedValues += entry.values.Len()
-		stats.ScannedBytes += entry.values.Len() * 8 // sizeof timestamp
+		ts.Timestamps = entry.AppendTimestamps(ts.Timestamps[:0])
+		if ts.Len() > 0 {
+			sort.Sort(&ts)
 
-		if entry.values.Contains(start, end) {
-			keyset.UnionKeys(tags)
+			stats.ScannedValues += ts.Len()
+			stats.ScannedBytes += ts.Len() * 8 // sizeof timestamp
+
+			if ts.Contains(start, end) {
+				keyset.UnionKeys(tags)
+			}
 		}
+
 		return nil
 	})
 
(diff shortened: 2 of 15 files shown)


tsdb/tsm1/reader_range_maxtime_iterator.go
```diff
@@ -0,0 +1,154 @@
...
+
+	e, ts := b.getEntriesAndTombstones()
+	if len(e) == 0 {
+		return
+	}
+
+	if len(ts) == 0 {
+		b.lastWrite = e[len(e)-1].MaxTime
+		// no tombstones, fast path will avoid decoding blocks
+		// if queried time interval intersects with one of the entries
```
e-dard, line 94: Can you explain this block to me a little bit? Is the aim to determine if any of the blocks (index entries) for this series overlap with the time range defined by `[b.tr.Min, b.tr.Max]`. Is that right?

I'm assuming that if you don't intersect the entry (the index entry is _not_ subsumed by the desired time range) then a block _may_ have data for the time range, but may not. Is that what you're determining here?

tsdb/tsm1/reader_range_iterator.go
```diff
@@ -105,9 +78,58 @@ func (b *TimeRangeIterator) HasData() bool {
...
+func (b *timeRangeBlockReader) Type() byte {
+	return b.iter.Type()
+}
+
+func (b *timeRangeBlockReader) getEntriesAndTombstones() ([]IndexEntry, []TimeRange) {
+	if b.err != nil {
+		return nil, nil
+	}
+
+	e := excludeEntries(b.iter.Entries(), b.tr)
```
e-dard, line 116: I wouldn't mind understand this a bit more. Doesn't `excludeEntries` only return entries that are _completely_ overlapped by `b.tr`? Wouldn't you be interested in entries that are _partially_ overlapped by `b.tr`?

tsdb/tsm1/reader_range_maxtime_iterator.go
```diff
@@ -0,0 +1,154 @@
...
+		if intersectsEntry(e, b.tr) {
+			b.hasData = true
+			return
+		}
+
+		for i := range e {
```
e-dard, line 100: I think I may have got the wrong end of the stick, but my assumption here is that these entries may or may not contain data within `b.tr`. And you have to decode the block to actually check for points that overlap `b.tr`?

I could think of one example of this being a block that has a time-range of `0–10` and the time-range you care about is `2–4`.

If my understanding of what is going on here is correct then I think you can improve the performance a bit here. Index entries should be stored in TSM files ordered by `min time`. I think the code that handles that is here:

https://github.com/influxdata/influxdb/blob/45250eeda2901cf9843cf63f9d0a9c4de626d23b/tsdb/tsm1/writer.go#L495-L498

Therefore, you should be able to exit early once an index entry's `min` time is greater than `b.tr.Max`.

Or maybe that's handled elsewhere?

tsdb/tsm1/cache_entry.go
```diff
@@ -117,6 +124,35 @@ func (e *entry) size() int {
...
+	contains := e.values.Contains(min, max)
+	e.mu.RUnlock()
+	return contains
+}
+
+// MaxTime returns the maximum timestamp for values in this entry.
+// The entry is deduplicated prior to calling contains.
+// If entry is empty, MaxTime will return InvalidMinNanoTime.
+func (e *entry) MaxTime() int64 {
+	e.deduplicate()
```
e-dard, line 141: It's not clear to me that you need to de-duplicate values in the cache to determine the maximum timestamp.

tsdb/tsm1/cache_entry.go
```diff
@@ -117,6 +124,35 @@ func (e *entry) size() int {
 	return sz
 }
 
+// Contains returns true if values exist for the time interval [min, max]
+// inclusive. The entry is deduplicated prior to calling contains.
+func (e *entry) Contains(min, max int64) bool {
+	e.deduplicate()
```
e-dard, line 130: It's not clear to me that you need to de-duplicate values in the cache overlap the range.

tsdb/tsm1/engine_schema.go
```diff
@@ -230,15 +230,26 @@ func (e *Engine) tagValuesPredicate(ctx context.Context, orgID, bucketID influxd
...
-		stats.ScannedValues += values.Len()
-		stats.ScannedBytes += values.Len() * 8 // sizeof timestamp
+		found := false
+		_ = e.Cache.ApplyEntryFnForKey(sfkey, func(entry *entry) error {
+			stats.ScannedValues += entry.values.Len()
+			stats.ScannedBytes += entry.values.Len() * 8 // sizeof timestamp
 
-		if values.Contains(start, end) {
-			tsmValues[string(curVal)] = struct{}{}
+			if entry.values.Contains(start, end) {
```
e-dard, line 245: I'm not sure I like this. Seems fine, but also seems a little off. The commit is adding new APIs on `entry` to avoid dipping into `values` unsafely, and then here you dip into `values`, albeit safely because you have a read lock via `entry.apply()`.

I think there is a little muddling going on with exported/unexported and locking on `entry`. Probably historical cruft....

Either way, I wonder if either `entry.apply` shouldn't take a lock (and leave the locking to the function argument), or perhaps if there should be an `entry.contains` that doesn't do any locking on `entry` and dips into `entry.values.Contains` for you.

Just some thoughts anyway..

tsdb/tsm1/reader_range_iterator.go
stuartcarnie: Your assumption is correct, and `excludeEntries` does handle partially covered, per these two tests.

https://github.com/influxdata/influxdb/blob/0ae8bebd755cf4aa579b9c38e6ea4792dfdbe53a/tsdb/tsm1/reader_range_iterator_test.go#L337-L352

Which is effectively:

```
+------------------------------------------+
|                 Entries                  |
+----------+---+----------+-----+----------+
| [0, 10]  |   | [12, 15] |     | [19, 21] |
+----------+   +----------+     +----------+
             +----------+                   
             |Time Range|                   
             +----------+                   
             | [11, 13] |                   
             +----------+                   
                   |                        
              excludeEntries                
                     |                      
                     v                      
               +----------+                 
               | [12, 15] |                 
               +----------+                 
                                            
                                            
+------------------------------------------+
|                 Entries                  |
+----------+---+----------+-----+----------+
| [0, 10]  |   | [12, 15] |     | [19, 21] |
+----------+   +----------+     +----------+
           +------------+                   
           | Time Range |                   
           +------------+                   
           |  [10, 13]  |                   
           +------------+                   
                  |                         
                  |                         
        excludeEntries                      
             |                              
             v                              
+-------------------------+                 
|         Entries         |                 
+----------+---+----------+                 
| [0, 10]  |   | [12, 15] |                 
+----------+   +----------+                              
```

tsdb/tsm1/cache_entry.go
stuartcarnie: This because `entry#values` doesn't guarantee sorted timestamps, so `MaxTime` would be incorrect – that is, we would be unable to use `values[len(values)-1].MaxTime()`.

Calling `Cache#Values` also deduplicates before returning the data (twice, when merging with an ongoing snapshot):

https://github.com/influxdata/influxdb/blob/a2a90ce87189c0738bddb5c9f27d71d59e060b8a/tsdb/tsm1/cache.go#L348

Upon further review, we should not put pressure on the cache to deduplicate every time this is called. I think it will be better to simply scan the values array for the `MaxTime`, being an `O(N)` operation.

tsdb/tsm1/cache_entry.go
stuartcarnie: As previous, values are only appended to the slice, so if written out of order, this would be incorrect:

https://github.com/influxdata/influxdb/blob/a2a90ce87189c0738bddb5c9f27d71d59e060b8a/tsdb/tsm1/cache_entry.go#L76-L77

I can do the same as what I propose for `MaxTime`; scan for the min and max values in a single pass and avoid the deduplicate call.

tsdb/tsm1/cache_entry.go
e-dard: Ah I didn't realise that `deduplicate` was the only time we did a lazy sort on the cache. Yeah so we could go `O(N)` (we do try and keep the cache relatively small).

Perhaps add a note that we may need to come back to this w.r.t perf? I feel like running through the entries will be better value than sorting them all the time..

tsdb/tsm1/cache_entry.go
e-dard: Yep, makes sense.

tsdb/tsm1/engine_schema.go
stuartcarnie: In the end I didn't like submitting this change either. I'll remove the API and use the same `MaxTime` and `Contains` APIs

stuartcarnie: (with the proposed changes to do the `O(N)` operations.

tsdb/tsm1/cache_entry.go
stuartcarnie: Overall this sucks compared to having the measurements+field types cache that we have in 1.x.

tsdb/tsm1/reader_range_maxtime_iterator.go
stuartcarnie: I'm going to add additional unit tests to verify this edge case. I agree that you observation is correct – we need to check if we overlap on a boundary of an entry and if not, decode the values and check whether it is contained.

tsdb/tsm1/cache_entry.go
e-dard: That cache didn't help us with time ranges though did it?

stuartcarnie: No, that is a true statement. Adding that dimension will aways make this hairy.

tsdb/tsm1/reader_range_maxtime_iterator.go
stuartcarnie: I have corrected this behavior and updated the unit tests, such that `TimeRangeMaxTimeIterator#MaxTime` will always return a value ≤ `b.tr.Max`.

stuartcarnie:
@e-dard / @jacobmarble I've reworked this PR significant, and simplified the changes to `Cache` and `entry` to a single API to address the data races. It is easy to reason about and will reduce overall allocations by replacing the call to `Cache#Values`, which allocates, sorts and deduplicates the data. We are only interested in the timestamps, hence the `AppendTimestamps` API.

stuartcarnie:
@jacobmarble I've also added comments to describe the struct `timeRangeBlockReader`, which extracts common functionality for the `TimeRangeIterator` and `TimeRangeMaxTimeIterator` structs.

tsdb/tsm1/reader_range_maxtime_iterator.go
stuartcarnie: Agreed; the code now reflects that 👍

PR #17590, opened 2020-04-03
chore(pkger): add stack state management for variables

jsteenb2:
references: #17434 

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/pkger/service.go
+++ b/pkger/service.go
@@ -889,32 +889,27 @@ func (s *Service) dryRunVariables(ctx context.Context, orgID influxdb.ID, pkg *P
 	mExistingLabels := make(map[string]DiffVariable)
 	variables := pkg.variables()
 
-VarLoop:
+	existingVars, _ := s.getAllPlatformVariables(ctx, orgID)
+
+	mIDs := make(map[influxdb.ID]*influxdb.Variable)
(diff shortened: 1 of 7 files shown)


pkger/service.go
```diff
@@ -889,32 +889,32 @@ func (s *Service) dryRunVariables(ctx context.Context, orgID influxdb.ID, pkg *P
 	mExistingLabels := make(map[string]DiffVariable)
 	variables := pkg.variables()
 
-VarLoop:
+	existingVars, _ := s.varSVC.FindVariables(ctx, influxdb.VariableFilter{
+		OrganizationID: &orgID,
+		// TODO: would be ideal to extend find variables to allow for a name matcher
```
jademcgough, line 894: Is it worth doing this now? Will this not work correctly for users with 100+ variables?

jsteenb2: we only make this call once per run. It could be done now, but definitely a nice to have.

jsteenb2: the VarSVC doesn't support a find by name atm 😢

jsteenb2: @jademcgough I pushed up the pagination changes to make sure we get all variables instead of only the first 100 👍

Issue #8304, opened 2017-04-19
Find a way to futher optimize GROUP BY * queries


jsternberg:
At the moment, we optimize queries like this well because we process the various iterators in parallel and perform what amounts to an iterative map/reduce:

```sql
SELECT count(value) FROM cpu
```

Unfortunately, we don't handle this type of query with the same grace. This type of query is bottlenecked at using one core.

```sql
SELECT count(value) FROM cpu GROUP BY *
```

The issue is with how internal processing of the query is done and how the final query is read. For the original query, we can process each series separately and perform an initial aggregate on a single series in parallel. Then we can combine the results of each series together. Since all of these points are processed in parallel with each other and the final result is read from only one location, it's much easier to merge them.

But, when we have `GROUP BY *` or a grouping by something with a high cardinality, the emitter will have to read all of the points from the first series before it can continue to reading the points from the second series. This means that only the first point in the second series is processed in parallel and then it blocks until it's time for the series to be read. Since series are processed in parallel, this makes `GROUP BY *` especially bad since it will inevitably only process series serially.

The emitter emits the series serially because that's what's required out of the emitter. It returns each series in order so it reads all of the series in order. While reading series 1, we cannot continue to process series 2 because the entirety of series 1 has to be emitted before we continue. The only way to continue processing series 2 would be to buffer the results of series 2 until they could eventually be read. With queries that emit a lot of points, this could result in running out of memory. If we limit the number of buffered points to prevent a query from running out of memory, we still don't solve the underlying problem which is that queries will always inevitably lead to using only a single core at some point within the query lifecycle. We just delay the issue.

Another solution might be to try and process a single series in parallel rather than serially like we currently do. This may or may not work. I'll expand on ideas, but we can't have a naive solution to doing this. The reason why we read a series in order is because seeking within the TSM tree generally doesn't improve performance very much. The slowest part of the query is finding the initial cursor and iteration is generally pretty fast after that. **Note: Double check this assumption as it changes the math behind this if it's not true.**

Either way, we will likely need a hybrid approach. Sometimes, processing the series in parallel is faster such as when we're merging series together. Sometimes, processing the series itself in parallel might be faster if we can think of a way to do that efficiently.

Another potential factor to consider is that this query will perform much slower than it likely should:

```sql
SELECT max(min) FROM (SELECT min(value) FROM cpu GROUP BY *)
```

[...]

PR #17596, opened 2020-04-03
 improvement(query): performance improvement for sorted merge iterator

foobar:
Sorted merge iterator has cpu-intensive operations to sort the points
from multiple inputs. Typical queries like `SELECT * FROM m GROUP BY *`
do not behave well due to the comparison of points though in many cases
it doesn't necessarily have to use the slow path.

This patch adds a shortcut.  If each input has a single and unique
series we can just return the points input by input.
The detection of the shortcut introduces slight overhead but the gains
are significant in many slow queries.

See #8304

### DIFF
--- a/query/iterator.gen.go
+++ b/query/iterator.gen.go
@@ -11,6 +11,7 @@ import (
 	"context"
 	"io"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
@@ -374,6 +375,7 @@ func (itr *floatSortedMergeIterator) pop() (*FloatPoint, error) {
 			}
 			itr.heap.items = append(itr.heap.items, item)
 		}
+		itr.heap.detectFast()
 		heap.Init(itr.heap)
 		itr.init = true
 	}
@@ -411,11 +413,57 @@ func (itr *floatSortedMergeIterator) pop() (*FloatPoint, error) {
 type floatSortedMergeHeap struct {
 	opt   IteratorOptions
 	items []*floatSortedMergeHeapItem
+	// if each input comes from a unique single time series, we can make a shortcut.
+	// detection of the shortcut introduces some overhead but it gets significant
+	// performance improvement in cases like SELECT * FROM m GROUP BY *
+	fast bool
+}
+
+func (h *floatSortedMergeHeap) detectFast() {
+	for _, item := range h.items {
+		if item.itr.Stats().SeriesN != 1 {
+			return
+		}
+	}
+
+	hasDup := false
+	s := make([]*floatSortedMergeHeapItem, len(h.items))
+	copy(s, h.items)
+
+	less := func(i, j int) bool {
+		x, y := s[i].point, s[j].point
+		ret := strings.Compare(x.Name, y.Name)
+		if ret == 0 {
+			ret = strings.Compare(x.Tags.ID(), y.Tags.ID())
+		}
+		if ret != 0 {
+			// TT
+			// ret | == -1 | h.opt.Ascending | result
+			//  1  | false |  false          | true
+			// -1  | true  |  false          | false
+			//  1  | false |  true           | false
+			// -1  | true  |  true           | true
+			return ret == -1 == h.opt.Ascending
+		}
+		hasDup = true
+		return false
+	}
+	sort.Slice(s, less)
+	if !hasDup {
+		h.fast = true
+		for i, item := range s {
+			item.fastIdx = i
+		}
+	}
 }
 
 func (h *floatSortedMergeHeap) Len() int      { return len(h.items) }
 func (h *floatSortedMergeHeap) Swap(i, j int) { h.items[i], h.items[j] = h.items[j], h.items[i] }
 func (h *floatSortedMergeHeap) Less(i, j int) bool {
+	if h.fast {
+		return h.items[i].fastIdx < h.items[j].fastIdx
+	}
+
 	x, y := h.items[i].point, h.items[j].point
 
 	if h.opt.Ascending {
@@ -489,6 +537,8 @@ type floatSortedMergeHeapItem struct {
 	point *FloatPoint
 	err   error
 	itr   FloatIterator
+	// index for fast shortcut
+	fastIdx int
 }
 
 // floatIteratorScanner scans the results of a FloatIterator into a map.
@@ -3038,6 +3088,7 @@ func (itr *integerSortedMergeIterator) pop() (*IntegerPoint, error) {
 			}
 			itr.heap.items = append(itr.heap.items, item)
 		}
+		itr.heap.detectFast()
 		heap.Init(itr.heap)
 		itr.init = true
 	}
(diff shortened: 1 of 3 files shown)


foobar:
@jsternberg could you take a look?

query/iterator.gen.go.tmpl
```diff
@@ -410,11 +411,75 @@ func (itr *{{$k.name}}SortedMergeIterator) pop() (*{{$k.Name}}Point, error) {
...
+       }
+
+       hasDup := false
+       s := make([]*{{$k.name}}SortedMergeHeapItem, len(h.items))
+       copy(s, h.items)
+
+       less := func(i, j int) bool {
+               x, y := s[i].point, s[j].point
+               ret := strings.Compare(x.Name, x.Name)
+               if ret == +1 {
```
e-dard, line 434: Please change all these instances of `+1` to `1`.

query/iterator.gen.go.tmpl
```diff
@@ -410,11 +411,75 @@ func (itr *{{$k.name}}SortedMergeIterator) pop() (*{{$k.Name}}Point, error) {
...
 	opt   IteratorOptions
 	items []*{{$k.name}}SortedMergeHeapItem
+	// if each input comes from a unique single time series, we can make a shortcut.
+	// detection of the shortcut introduces some overhead but it gets significant
+	// performance improvement in cases like SELECT * FROM m GROUP BY *
+	fast bool
+}
+
+func (h *{{$k.name}}SortedMergeHeap) detectFast() {
+       for _, item := range h.items {
```
e-dard, line 421: Is it possible that `len(h.items) == 1`? If so, then you could add a shortcut here:

```go
if len(h.items) == 1 {
    return
}
```

query/iterator.gen.go.tmpl
```diff
@@ -410,11 +411,75 @@ func (itr *{{$k.name}}SortedMergeIterator) pop() (*{{$k.Name}}Point, error) {
...
+                       ret = strings.Compare(x.Tags.ID(), y.Tags.ID())
+                       if ret == -1 {
+                               return true
+                       } else if ret == +1 {
+                               return false
+                       }
+                       hasDup = true
+                       return false
+               }
+       }
```
e-dard, line 466:
```suggestion
        less := func(i, j int) bool {
                x, y := s[i].point, s[j].point
                ret := strings.Compare(x.Name, x.Name)
                if ret != 0 {
                        // TT
                        // ret | == -1 | h.opt.Ascending | result
                        //  1  | false |  false          | true
                        // -1  | true  |  false          | false
                        //  1  | false |  true           | false
                        // -1  | true  |  true           | true
                        return ret == -1 == h.opt.Ascending
                }

                ret = strings.Compare(x.Tags.ID(), y.Tags.ID())
                if ret != 0 {
                        // TT
                        // ret | == -1 | h.opt.Ascending | result
                        //  1  | false |  false          | true
                        // -1  | true  |  false          | false
                        //  1  | false |  true           | false
                        // -1  | true  |  true           | true
                        return ret == -1 == h.opt.Ascending
                }
                hasDup = true
                return false
        }

```

e-dard: @foobar what do you think about this change? It will DRY the code up a bit. I'm not sure if I have the condition correct, so you should probably check it works 😄

foobar: neat!  Then how about merging the two returns, like this?:

```go
ess := func(i, j int) bool {
                x, y := s[i].point, s[j].point
                ret := strings.Compare(x.Name, x.Name)
                if ret == 0 {        
                        ret = strings.Compare(x.Tags.ID(), y.Tags.ID())
                }
                if ret != 0 {
                        // TT
                        // ret | == -1 | h.opt.Ascending | result
                        //  1  | false |  false          | true
                        // -1  | true  |  false          | false
                        //  1  | false |  true           | false
                        // -1  | true  |  true           | true
                        return ret == -1 == h.opt.Ascending
                }
                hasDup = true
                return false
        }
```

query/iterator.gen.go.tmpl
foobar: solved now

query/iterator.gen.go.tmpl
foobar: looks it's already handled at line 224:

https://github.com/influxdata/influxdb/blob/a0f2e0c21aada3663f9960f1b773910572f5cc74/query/iterator.go#L220-L226

e-dard:
@foobar hi, I'm happy with  what's here so far. However, I still would like to see some benchmarking and testing of the new code-path. Further  @stuartcarnie  may have his own suggestions. Thanks

foobar:
Thanks for your comments! @e-dard 

> 1. I have a couple of small suggestions. The main one is to remove the use of `+1`, and just use `1` instead, but I also made a simplification to the code and a possible short circuit.

I reworked the code based on your suggestion.

> 2. I don't know if we will put this work into a `1.7` release. Can you change the base branch to `master-1.x`? If accepted, we can manage the backport to the `1.8` branch.

done

> 3. I would like to see some benchmarks showing the changes in performance for this change. I mainly concerned about whether the extra work to detect if the fast condition will be used will be significantly detrimental to other use-cases (queries).

benchmark test added and here is the result on my machine:
```
go test -bench=BenchmarkSorted ./query/...
goos: linux
goarch: amd64
pkg: github.com/influxdata/influxdb/query
BenchmarkSortedMergeIterator_Fast-32                         166           6738013 ns/op
BenchmarkSortedMergeIterator_NotFast-32                       19          61152186 ns/op
BenchmarkSortedMergeIterator_FastCheckOverhead-32        1718110               700 ns/op
```

> 4. can you add some test coverage for `detectFast` in the form of unit tests?

added test cases

foobar:
Thanks @stuartcarnie for reviewing this PR.

> May I suggest the following small change, which may be a little easier for a future maintainer to understand. It is also likely to be a little more efficient and avoids the use of `strings.Compare`, which is not recommended by the [Go documentation](https://pkg.go.dev/strings?tab=doc#Compare) and [the source itself] 

From the source it shouldn't have significant difference in term of performance; for readability, current code is shorter with  @e-dard comments. I'm fine with either. 
@e-dard your thought?

foobar:
@e-dard @stuartcarnie @rickspencer3 any other comments?

e-dard:
@stuartcarnie can you run with this? I got a bunch of other review backed up :-)

stuartcarnie:
I am ok, as-is. Thanks again, @foobar!

@dgnorton if you are happy, you are welcome to merge it

foobar:
hi @dgnorton, had you got a chance to look at this ?

foobar:
@dgnorton could you get it merged?

ayang64:
@foobar, @stuartcarnie 

Please forgive me if I'm off base -- I haven't tested this thoroughly but I think the entire `less` func could be simplified to something like this:

```
less := func(i, j int) bool {
  x, y := s[i].point, s[j].point
  hasDup = x.Name == y.Name
  return ((x.Name < y.Name) || (x.Tags.ID() < y.Tags.ID()) && h.opt.Ascending
}
```

this should be a bit faster.  i haven't tested or benchmarked it though.

query/iterator.gen.go
```diff
@@ -411,11 +413,57 @@ func (itr *floatSortedMergeIterator) pop() (*FloatPoint, error) {
...
+		if item.itr.Stats().SeriesN != 1 {
+			return
+		}
+	}
+
+	hasDup := false
+	s := make([]*floatSortedMergeHeapItem, len(h.items))
+	copy(s, h.items)
+
+	less := func(i, j int) bool {
```
ayang64, line 433: please investigate if something like the following would work as a comparator:

```
less := func(i, j int) bool {
  x, y := s[i].point, s[j].point
  hasDup = hasDup || x.Name == y.Name
  return ((x.Name < y.Name) || (x.Tags.ID() < y.Tags.ID()) && h.opt.Ascending
}
```
i think that should be at least as fast.

stuartcarnie: @ayang64 the `less` function is required to sort ascending or descending order depending on `h.opt.Ascending`. It is unclear to me if the short version above achieves that. Also, it is setting `hasDup` to `true` only if `x.Name == y.Name`, whereas the original function sets `hasDup` to true iif both the `Name` and `Tags` are equal, which is a required property.

I didn't see any obvious performance issues with the existing code, and the overall improvement is significant.

foobar:
hi @timhallinflux, could this catch 1.8.1 milestone?

PR #17597, opened 2020-04-03
fix(tenant): make storage pass tenant tests

affo:
@lyondhill do you think the tests are ok?

I relaxed some assumptions, because I thought about it more and.. That's what I came up with 👍

### DIFF
--- a/tenant/service_bucket.go
+++ b/tenant/service_bucket.go
@@ -108,6 +108,8 @@ func (s *Service) FindBuckets(ctx context.Context, filter influxdb.BucketFilter,
 		return nil, 0, err
 	}
 
+	// NOTE: this is a remnant of the old system.
+	// There are org that do not have system buckets stored, but still need to be displayed.
 	needsSystemBuckets := true
 	for _, b := range buckets {
 		if b.Type == influxdb.BucketTypeSystem {
--- a/tenant/storage_urm.go
+++ b/tenant/storage_urm.go
@@ -13,7 +13,15 @@ var (
 	urmByUserIndexBucket = []byte("userresourcemappingsbyuserindexv1")
 )
 
+// NOTE(affo): On URM creation, we check that the user exists.
+// We do not check that the resource it is pointing to exists.
+// This decision takes into account that different resources could not be in the same store.
+// To perform that kind of check, we must rely on the service layer.
+// However, we do not want having the storage layer depend on the service layer above.
 func (s *Store) CreateURM(ctx context.Context, tx kv.Tx, urm *influxdb.UserResourceMapping) error {
+	if _, err := s.GetUser(ctx, tx, urm.UserID); err != nil {
+		return err
+	}
 	if err := s.uniqueUserResourceMapping(ctx, tx, urm); err != nil {
 		return err
 	}
--- a/tenant/storage_user.go
+++ b/tenant/storage_user.go
@@ -270,13 +270,29 @@ func (s *Store) DeleteUser(ctx context.Context, tx kv.Tx, id influxdb.ID) error 
 		return ErrInternalServiceError(err)
 	}
 
-	// clean up users password
+	// Clean up users password.
 	ub, err := tx.Bucket(userpasswordBucket)
 	if err != nil {
 		return UnavailablePasswordServiceError(err)
 	}
+	if err := ub.Delete(encodedID); err != nil {
+		return err
+	}
 
-	return ub.Delete(encodedID)
+	// Clean up user URMs.
+	urms, err := s.ListURMs(ctx, tx, influxdb.UserResourceMappingFilter{UserID: id})
+	if err != nil {
+		return err
+	}
+	// Do not fail fast on error.
+	// Try to avoid as much as possible the effects of partial deletion.
+	aggErr := NewAggregateError()
+	for _, urm := range urms {
+		if err := s.DeleteURM(ctx, tx, urm.ResourceID, urm.UserID); err != nil {
+			aggErr.Add(err)
+		}
+	}
+	return aggErr.Err()
 }
 
 func (s *Store) GetPassword(ctx context.Context, tx kv.Tx, id influxdb.ID) (string, error) {
--- a/testing/tenant.go
+++ b/testing/tenant.go
@@ -158,6 +158,13 @@ func Create(t *testing.T, init func(*testing.T, TenantFields) (influxdb.TenantSe
 		defer done()
 		ctx := context.Background()
 
+		// Number of buckets prior to user creation.
+		// This is because, for now, system buckets always get returned for compatibility with the old system.
+		_, nbs, err := s.FindBuckets(ctx, influxdb.BucketFilter{})
+		if err != nil {
+			t.Fatal(err)
+		}
+
 		u := &influxdb.User{
 			ID:   1,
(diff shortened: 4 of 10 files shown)


testing/tenant.go
```diff
@@ -599,49 +645,46 @@ func Delete(t *testing.T, init func(*testing.T, TenantFields) (influxdb.TenantSe
...
+			ResourceID:   data.Buckets[0].ID,
+		})
+		if nurms != 0 {
+			t.Errorf("expected bucket1, to be dangling, got: %+v", urms)
 		}
 	})
 
-	// NOTE(affo)(*kv.Service): it deletes urms, but not dangling resources.
-	t.Run("deleting a user deletes every urm and so for dangling resources", func(t *testing.T) {
+	t.Run("deleting a user deletes every related urm and nothing else", func(t *testing.T) {
```
affo, line 687: decided not  to  remove dangling resources, that seems too much.

My thought  is that we do not want to automatically propagate this kind of delete because a resource will always have an owner org.

testing/tenant.go
```diff
@@ -523,8 +568,9 @@ func Delete(t *testing.T, init func(*testing.T, TenantFields) (influxdb.TenantSe
 		}
 	})
 
-	// NOTE(affo)(*kv.Service): it does create dangling resources.
-	t.Run("deleting bucket urm does not create dangling resources", func(t *testing.T) {
+	// NOTE(affo): those resources could not be dangling (URM could be inferred from an user being in the owner org).
+	// We do not want to automatically propagate this kind of delete because an resource will always have an owner org.
+	t.Run("deleting bucket urm does create dangling bucket", func(t *testing.T) {
```
affo, line 573: same here, we don't want to propagate. I think the comment makes it clear

testing/tenant.go
```diff
@@ -295,7 +331,7 @@ func Create(t *testing.T, init func(*testing.T, TenantFields) (influxdb.TenantSe
 		}
 	})
 
-	t.Run("making user part of org creates mappings to buckets", func(t *testing.T) {
+	t.Run("making user part of org creates mapping to org only", func(t *testing.T) {
```
affo, line 334: this is because I know you'll change URM creation soon 👍

testing/tenant.go
```diff
@@ -711,8 +769,8 @@ func Delete(t *testing.T, init func(*testing.T, TenantFields) (influxdb.TenantSe
 
 		// Delete org1.
 		// We expect its buckets to be deleted.
-		// We expect urms to those buckets to be deleted too
-		// user1 should not be deleted because he is member of org2.
+		// We expect urms to those buckets to be deleted too.
+		// No user should be deleted.
```
affo, line 773: a user should have the right to exist even if an org gets deleted

tenant/storage_user.go
```diff
@@ -275,8 +275,25 @@ func (s *Store) DeleteUser(ctx context.Context, tx kv.Tx, id influxdb.ID) error
...
+	// Do not fail fast on error.
+	// Try to avoid as much as possible the effects of partial deletion.
+	var aggErr error
+	for _, urm := range urms {
+		err := s.DeleteURM(ctx, tx, urm.ResourceID, urm.UserID)
+		if aggErr == nil {
+			aggErr = err
+		}
+	}
+	return aggErr
```
affo, line 296: is this the proper way of doing this? 🤔

tenant/service_bucket.go
```diff
@@ -108,35 +108,38 @@ func (s *Service) FindBuckets(ctx context.Context, filter influxdb.BucketFilter,
 		return nil, 0, err
 	}
 
-	needsSystemBuckets := true
-	for _, b := range buckets {
-		if b.Type == influxdb.BucketTypeSystem {
-			needsSystemBuckets = false
-			break
+	// NOTE(affo): why should we artificially append system buckets if they are not there?
```
affo, line 111: yeah, why?

affo:
@lyondhill added comments just to review this easily 👍 

Hope it helps

tenant/service_bucket.go
lyondhill: this is a remnant of the old system. there are old org's that dont have system buckets stored but still need to be displayed.

tenant/storage_urm.go
```diff
@@ -13,7 +14,57 @@ var (
...
+func getBucket(rt influxdb.ResourceType) ([]byte, error) {
+	var bucket []byte
+	switch rt {
+	case influxdb.UsersResourceType:
+		bucket = userBucket
+	case influxdb.OrgsResourceType:
+		bucket = organizationBucket
+	case influxdb.BucketsResourceType:
+		bucket = bucketBucket
+	default:
```
lyondhill, line 26: This kind of switch has to be avoided. Once we add more systems in that use the tenant urm system it would be tempting to add non tenant types to this and that would enforce that all systems exist in the same database. We know that will not be the case in the future.

tenant/storage_urm.go
```diff
@@ -13,7 +14,57 @@ var (
...
+			return errNotFound(influxdb.UsersResourceType)
+		}
+		return err
+	}
+	if _, err := s.get(ctx, tx, resBucket, urm.ResourceID); err != nil {
+		if kv.IsNotFound(err) {
+			return errNotFound(urm.ResourceType)
+		}
+		return err
+	}
```
lyondhill, line 67: I think its a good thing to make sure the user exist for the URM. I dont think once we have other systems (like auth/task) using create URM we will not be able to verify that the resource exists before creating a urm. I think the right action here is to check user (probably using the `s.GetUser` function instead of a generic get) and then create the urm weather the other resource exists or not. a dangling urm isn't a problem I dont think

tenant/storage_user.go
lyondhill: Your `aggErr` is actually just the "first error that occured". If that is what you want that works. If you want to display all the errors You could create a `type aggErr []error`. Then write an `Error()` function that concat's all the error messages from the errors in the list.

tenant/storage_urm.go
affo: That's perfect! I was investing time in making most tests pass given this new logic. I'll rethink about that and update. Thank you

tenant/service_bucket.go
affo: mmmm ok, I had a failing test, because that way asking for buckets always return them.
Maybe I can fix the test if this has to stay here 👍

tenant/storage_urm.go
affo: Yeah,  I think you are right 👍

tenant/storage_urm.go
affo: Thinking more about it, this matches with the problem with the `switch` statement above.

I see one way of doing this: refactor the switch not to return a `bucket`,  but to implement a generic resource find that errors on `not found`. The problem there is that we would end up having the storage layer depend on the service layer. We don't want that. That check should be done at the service layer 👍 

For now, I'll do what you suggest. I'll check user and not the other resource 👍

tenant/storage_urm.go
affo: See below for more

affo:
@lyondhill I think I addressed you comments in the last commit 👍

Issue #17008, opened 2020-02-25
Add the ability to specify the encoding when importing csv via the influx cli


russorat:

sranka:
https://pkg.go.dev/golang.org/x/text@v0.3.2/encoding/ianaindex will simplify the implementation

PR #17599, opened 2020-04-03
feat(cmd/influx/write): enhance the ability to import CSV files

### LINKED ISSUE #17004: Add the ability to import a simple csv file via the influx cli
#17398 introduces CSV data to `influx write` command. As a next step, we should not require (and reduce the need) to change CSV data on input.

_Enhancements in CSV data processing_ 

- `--header` option in `influx write` command let your add annotation or header rows without changing the data on input (supplied via `--file` or stdin)
   - you can supply more `--header` options, new CSV rows will be prepended in the order as they appear on the command line, you can also supply more`--file` options, files follow headers
- the header row (i.e. the row that define column names) can also define column data types when supplied as `name|datatype`; for example `cpu|tag` defines a tag column named _cpu_
  - moreover, it can also specify a default value when supplied as `name|datatype|default`; for example, `count|long|0` defines a field column named _count_ of _long_ data type that will not skip the field if a column value empty, but uses '0' as the column value
- a new `#constant` annotation adds a constant column to the data, so you can set measurement, time, field or tag of every row you import independently on the data you passed in
   - the format of a constant annotation row is `#constant,datatype,name,value`, so you specify a supported datatype, a column name, and a constant value
   - `column` can be omitted for _dateTime_ or _measurement_ columns, so the annotation can be simply `#constant,measurement,cpu`
   - note that you can add constant annotations to existing data using `--header` option of `influx write` cli
- a time column can be of `dateTime:format` datatype to use a custom _format_ to parse column values
   - the format follows the layout described in https://golang.org/pkg/time layout, for example `dateTime:2006-01-02` parses 4-digit-year , '-' , 2-digit month ,'-' , 2 digit day of the month
   - the following formats are predefined:
      - `dateTime:RFC3339` format is 2006-01-02T15:04:05Z07:00
      - `dateTime:RFC3339Nano` format is 2006-01-02T15:04:05.999999999Z07:00
      - `dateTime:number` represent UTCs time since epoch in nanoseconds
   - if the time format includes a time zone, the parsed date time respects the time zone; otherwise the timezone dependends on the presence of `#timezone` annotation; if there is no #timezone annotation, UTC is used
      - `#timezone` annotation specifies time zone using an offset , which is either `+hhmm` or `-hhmm` or `Local` to use the local/computer time zone. Examples:  _#timezone,+0100_  _#timezone -0500_ _#timezone Local_
          
- a _double_ field column can be supplied with `format` as `double:format`
   - the `format`'s first character is used to separate integer and fractional part (usually `.` or `,`), second and next format's characters (such as as `, _`) are removed  from the column value, these removed characters are usually used to visually separate large numbers into groups
   - for example a Spanish locale value `3.494.826.157,123` is of `double:,.` type; the same `double` value is  _3494826157.123_
   - _long_ or _unsignedLong_ data types also support format, everything after and including a fraction character is ignored
   - note that you have to quote column delimiters whenever they appear in a CSV column value, for example:
      - `#constant,"double:,.",myColumn,"1.234,011"`
- a _boolean_ field column can be supplied with `format` as `boolean:truthy:falsy`
   - `truthy` are `falsy` are comma-separated lists of values that are separated by `:` , they can be empty to assume other values as truthy/falsy
   - a  `boolean` without format parses columns values that start with any of _tTyY1_ as `true` values, _fFnN0_ as `false` values and fails on other values
   - a column with an empty value is excluded in the protocol line, setup `#default` annotation for such a column define default value 
   - for example `boolean:sí,yes,ja,oui,ano,да:no,nein,non,ne,нет` 
- a CSV file can start with a line `sep=;` to inform about a character that is used to separate columns in data rows, by default `,` is used as a column separator
- `--skipHeader` flag skips the first line from the input data (the last file or stdin); the optional value of this flag specifies the number of lines to remove; for example `skipHeader=2` removes the first 2 lines
- `--debug` and `--continueOnError` options help with troubleshooting of CSV conversions
   - `--debug` prints (to stderr) metadata about columns that are used to create protocol lines out of CSV data rows
   - `--skipRowOnError` prints CSV data conversion error to stderr, skips the failing row and continues with CSV processing
      - any error stops the processing without this option specified

sranka:
Closes #17004 #17008

There is a bunch of enhancements in the way of how CSV data are processed by `influx write` so that external CSV data can be written to InfluxDB. They are described in detail in #17004 + #17008  

Examples of new capabilities are provided in https://github.com/bonitoo-io/influxdb-csv-import/blob/master/README.md#step-2

New flags in `influx write` :
```
      --debug                Log CSV columns to stderr before reading data rows
      --encoding string      Character encoding of input files or stdin (default "UTF-8")
  -f, --file stringArray     The path to the file to import
      --header stringArray   One or more header lines to prepend to input data
      --skipHeader int[=1]   Skip the first <n> rows from input data
      --skipRowOnError       Log CSV data errors to stderr and continue with CSV processing
```

- [ ] Rebased/mergeable
- [x] Tests pass


cmd/influx/write.go
```diff
@@ -74,41 +90,102 @@ func cmdWrite(f *globalFlags, opt genericCLIOpts) *cobra.Command {
 	return cmd
 }
 
+func (writeFlags *writeFlagsType) dump(args []string) {
+	if writeFlags.Debug {
```
russorat, line 94: don't we have a `log.Debug` function that would be better here?

cmd/influx/write.go
```diff
@@ -74,41 +90,102 @@ func cmdWrite(f *globalFlags, opt genericCLIOpts) *cobra.Command {
...
-func (writeFlags *writeFlagsType) createLineReader(args []string) (r io.Reader, closer io.Closer, err error) {
-	if len(args) > 0 && args[0][0] == '@' {
+func (writeFlags *writeFlagsType) createLineReader(cmd *cobra.Command, args []string) (io.Reader, io.Closer, error) {
+	readers := make([]io.Reader, 0, 2*len(writeFlags.Headers)+2*len(writeFlags.Files)+1)
+	closers := make([]io.Closer, 0, len(writeFlags.Files))
+
+	if len(args) > 0 && len(args[0]) > 1 && args[0][0] == '@' {
 		// backward compatibility: @ in arg denotes a file
-		writeFlags.File = args[0][1:]
+		writeFlags.Files = append(writeFlags.Files, args[0][1:])
```
jsteenb2, line 107: I'm concerned the flags are being used as a bucket for a bunch of side effects here

cmd/influx/write.go
```diff
@@ -74,41 +90,102 @@ func cmdWrite(f *globalFlags, opt genericCLIOpts) *cobra.Command {
...
+			readers = append(readers, decode(f), strings.NewReader("\n"))
+			if len(writeFlags.Format) == 0 && strings.HasSuffix(file, ".csv") {
+				writeFlags.Format = inputFormatCsv
+			}
+		}
+	}
+
+	// add stdin or a single argument
+	switch {
+	case len(args) == 0:
```
jsteenb2, line 149: there is a way to identify if stdin was provided, there is no reason we can't use both file and stdin, nothing stopping us from doing that technically.

check it out here:

https://github.com/influxdata/influxdb/blob/master/cmd/influx/pkg.go#L536

cmd/influx/write.go
sranka: The built-in golang's log package has no Debug fn, OOTB it prints messages to stderr.

cmd/influx/write.go
```diff
@@ -74,41 +90,102 @@ func cmdWrite(f *globalFlags, opt genericCLIOpts) *cobra.Command {
 	return cmd
 }
 
+func (writeFlags *writeFlagsType) dump(args []string) {
```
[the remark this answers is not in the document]
sranka: It helps with the troubleshooting of CSV -> LP conversion. It is easier to find problems in user input when you dump the actual input arguments together with their default values. This is printed only when in --debug mode, together with metadata about columns that are used to extract CSV data to create protocol lines.  

Sure I can remove it, but it will be then harder to resolve user input errors. Opinionated herein, I do insist on having it here.

cmd/influx/write.go
```diff
@@ -74,41 +90,102 @@ func cmdWrite(f *globalFlags, opt genericCLIOpts) *cobra.Command {
...
+		log.Printf("WriteFlags%+v", *writeFlags)
+	}
+}
+
 // createLineReader uses writeFlags and cli arguments to create a reader that produces line protocol
-func (writeFlags *writeFlagsType) createLineReader(args []string) (r io.Reader, closer io.Closer, err error) {
-	if len(args) > 0 && args[0][0] == '@' {
+func (writeFlags *writeFlagsType) createLineReader(cmd *cobra.Command, args []string) (io.Reader, io.Closer, error) {
+	readers := make([]io.Reader, 0, 2*len(writeFlags.Headers)+2*len(writeFlags.Files)+1)
+	closers := make([]io.Closer, 0, len(writeFlags.Files))
```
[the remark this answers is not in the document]
sranka: closers are subset of readers herein, only opened files will be closed, nothing more ... stdin or string readers cannot/must not be closed.

cmd/influx/write.go
sranka: yes, for the sake of backward compatibility @file means a file ... somebody introduced this and I decided to keep this functionality because I don't know where and how it is currently used. When starting from scratch, I would not allow specifying data in an argument at all.

cmd/influx/write.go
sranka: I can change it that way too.

cmd/influx/write_test.go
```diff
@@ -0,0 +1,414 @@
+package main
```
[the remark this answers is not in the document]
sranka: I introduced this test, because I wanted to keep the playground cleaner than before I entered, but it might not look so, ok :) So I will write down more comments herein to know what is being tested.

cmd/influx/write_test.go
```diff
@@ -0,0 +1,414 @@
...
+		{"b x=_measurement", "y=c", "z=d"},
+		{"f2 x=f1", "y=f3", "z=f4"},
+	}, rows)
+
+	// import from csv file as @file + header + skipLines
+	flags = writeFlagsType{}
+	flags.Headers = []string{"x,_measurement,y,z"}
+	flags.Files = []string{}
+	flags.SkipHeader = 1
+	command = cmdWrite(&globalFlags{}, genericCLIOpts{in: strings.NewReader(stdInContents)})
```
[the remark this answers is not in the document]
sranka: will do

cmd/influx/write_test.go
```diff
@@ -0,0 +1,414 @@
...
+		log.SetOutput(os.Stderr)
+		log.SetFlags(oldFlags)
+		log.SetPrefix(oldPrefix)
+	}, &buf
+}
+
+var tempFiles []string
+
+func removeTempFiles() {
+	for len(tempFiles) > 0 {
```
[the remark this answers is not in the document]
sranka: sure

cmd/influx/internal/isCharacterDevice.go
```diff
@@ -0,0 +1,22 @@
+package internal
+
+import (
+	"io"
+	"os"
+)
+
+// IsCharacterDevice returns true if the supplied reader is a character device (a terminal)
+func IsCharacterDevice(reader io.Reader) bool {
```
jsteenb2, line 9: what is the motivation to move this to the internal pkg?

cmd/influx/write.go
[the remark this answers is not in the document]
jsteenb2: I hear you, seems better to do it as ReadClosers with a noop closer on those that don't need a closer, typical pattern in golang.

```go
type noopCloser struct {
    r io.reader
}

func (n *noopCloser) Close() error {
    return nil 
}
```

my big concern here is it makes it easy to lose track of closers/readers when they ares disparate uncoupled return types.

Last thing, please don't resolve conversations, I'll happily resolve them when I feel the conversation has come to fruition 👍

cmd/influx/write.go
jsteenb2: I don't htink my eariler comment was clear. I'd rather see you maintain the files in a sep variable instead of creating a destructive call on the flags. Flags should only be used (imo), for the API contract between CLI and user. Makes things coupled in weird global ways atm.

cmd/influx/internal/isCharacterDevice.go
```diff
@@ -0,0 +1,22 @@
...
+	file, isFile := reader.(*os.File)
+	if !isFile {
+		return false
+	}
+	info, err := file.Stat()
+	if err != nil {
+		return false
+	}
+	if (info.Mode() & os.ModeCharDevice) == os.ModeCharDevice {
```
jsteenb2, line 18: just realized this can be simplified too:

```go
return (info.Mode() & os.ModeCharDevice) == os.ModeCharDevice
```

the if isn't necessary here

cmd/influx/write_test.go
```diff
@@ -0,0 +1,436 @@
...
+	lineData = lineData[:0]
+	flags.Host = server.URL
+	command = cmdWrite(&globalFlags{}, genericCLIOpts{w: ioutil.Discard})
+	// note: my-error-bucket parameter causes the test server to fail
+	command.SetArgs([]string{"--format", "csv", "--org", "my-org", "--bucket", "my-error-bucket"})
+	err = command.Execute()
+	require.Contains(t, fmt.Sprintf("%s", err), "bucket") // failed to retrieve buckets
+
+	// validation: no such bucket found
+	lineData = lineData[:0]
```
jsteenb2, line 396: are these sub tests? its easier to erad, but feels like a lot of coupling between assertions.  Can this be split up into sub tests?

```go
t.Run("validates no bucket found", func(t *testing.T) {
       lineData := lineData[:0]
        command := cmdWrite(&globalFlags{}, genericCLIOpts{w: ioutil.Discard})
	// note: my-empty-org parameter causes the test server to return no buckets
	command.SetArgs([]string{"--format", "csv", "--org", "my-empty-org", "--bucket", "my-bucket"})
	err = :command.Execute()
	require.Contains(t, fmt.Sprintf("%s", err), "bucket") // no such bucket found
})
```

same goes for a lot of these assertions here. They look like sub tests to me but aren't being treated as such. Please correct me if that assumption is wrong 🤔

cmd/influx/write.go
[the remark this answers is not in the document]
sranka: Thank you for the clarification around "conversation resolving", as a newbie herein, I need such advice 👍

I see two concerns herein:
1. Why are readers and closers separate instead of being ReadClosers?
   - files can be provided in non-UTF-8 encoding, a file (ReaderCloser) is then transformed to UTF-8 Reader (not a ReadCloser) 
https://github.com/influxdata/influxdb/blob/b786fc16520aae61482f7e09c149b850aa27ff06/cmd/influx/write.go#L141
   - the last file or stdin can be then changed so that first N lines are skipped when --skipHeader option is present. This transformation also works with Readers, but not ReadClosers https://github.com/influxdata/influxdb/blob/b786fc16520aae61482f7e09c149b850aa27ff06/cmd/influx/write.go#L168   
   - changing the above intermediate  Readers to ReadClosers to have just one ReadCloser array would introduce an extra complexity herein IMO 
2. Why are io.Reader and io.Closer disparate return values?
   - I understand that a composite io.ReaderCloser looks better. It would require to wrap multiple Readers and Closers (or ReadClosers) in a new type that will only delegate calls. I wanted to prefer builtins whenever possible, but there is no multi-reader-closer for reuse.
  - At the same time, it is not a big deal (in my POV) for a non-public function to avoid this wrapping. "A returned Closer must be closed" is an expected contract IMO.

cmd/influx/write.go
sranka: understood, repaired

cmd/influx/write_test.go
sranka: Yes, it looks better when using sub-tests.

sranka:
@rbetts I've seen only a few golang's CSV parser extensions that can do (un)marshaling into structs of a pre-defined layout. We have a different situation herein:
  - the layout is driven by CSV data itself with the help of annotations and/or column headers
  - a CSV with variable row size (such as a result of a flux query) is supported and can be used on input
  - the code should be optimized to transform to line protocol and recognize CSV data annotations that are AFAIK used only in InfluxDB

@russorat Can you please comment on the requested creation of  CSV -> LP conversion library? I mean, there are no technical problems to do so, once you agree and provide a new influxdata git repo.

jsteenb2:
@sranka you should be able to put the parser under the `pkg` directory in the monorepo here in influxdb. If we need to pull it out, we can always do so.

sranka:
@jsteenb2 ok, "csv to line protocol conversion" is now in pkg/csv2lp

jsteenb2:
@sranka, we'd like to see this PR broken up and submitted incrementally. First PR for the csv converter and another for the CLI. We'll have different stakeholders reviewing for the csv PR.

sranka:
@jsteenb2 thanks for letting me know of what is required.

PR #17600, opened 2020-04-03
chore(ui): updated login page styling to match Quartz

Asalem1:
This PR intends to sync up changes that occurred with the Quartz sign-up page. Since the two pages should be identical, any changes made on one should be reflected in the other.


alexpaxton:
Check out this file: https://github.com/influxdata/influxdb/blob/master/ui/src/shared/utils/pageTitles.ts

In either case you may want to use the `CLOUD` environment variable to say either `InfluxDB` or `InfluxDB Cloud`

PR #17605, opened 2020-04-03
fix: sick of seeing this error

drdelambre:
Screen Shot 2020-04-03 at 9 19 09 AM

```
response from /api/v2private/orgs/844910ece80be8bc/limits:
{
    orgID: "844910ece80be8bc"
    rate: null
    bucket: null
    task: null
    dashboard: null
    check: null
    notificationRule: null
    notificationEndpoint: null
    feature: {allowDelete: true}
}
```

### DIFF
--- a/ui/src/cloud/reducers/limits.ts
+++ b/ui/src/cloud/reducers/limits.ts
@@ -71,10 +71,12 @@ export const limitsReducer = (
           draftState.buckets.maxAllowed = maxBuckets
           draftState.buckets.maxRetentionSeconds = maxRetentionDuration / 1e9
         }
+
         if (limits.dashboard) {
           const {maxDashboards} = limits.dashboard
           draftState.dashboards.maxAllowed = maxDashboards
         }
+
         if (limits.task) {
           const {maxTasks} = limits.task
           draftState.tasks.maxAllowed = maxTasks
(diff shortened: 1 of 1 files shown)


ui/src/cloud/reducers/limits.ts
```diff
@@ -66,33 +66,50 @@ export const limitsReducer = (
...
+        }
+        if (limits.task) {
+          const {maxTasks} = limits.task
+          draftState.tasks.maxAllowed = maxTasks
+        }
+
+        if (limits.check) {
+          const {maxChecks} = limits.check
+          draftState.checks.maxAllowed = maxChecks
+        }
```
121watts, line 86:
```suggestion
        }
        
        if (limits.dashboard) {
          const {maxDashboards} = limits.dashboard
          draftState.dashboards.maxAllowed = maxDashboards
        }
        
        if (limits.task) {
          const {maxTasks} = limits.task
          draftState.tasks.maxAllowed = maxTasks
        }

        if (limits.check) {
          const {maxChecks} = limits.check
          draftState.checks.maxAllowed = maxChecks
        }
```

ui/src/cloud/reducers/limits.ts
```diff
@@ -66,33 +66,50 @@ export const limitsReducer = (
...
-        draftState.rules.blocked = blockedNotificationRules
-          .split(',')
-          .map(r => r.trim())
-        draftState.endpoints.blocked = blockedNotificationEndpoints
-          .split(',')
-          .map(r => r.trim())
-        draftState.rate.readKBs.maxAllowed = readKBs
-        draftState.rate.writeKBs.maxAllowed = writeKBs
-        draftState.rate.cardinality.maxAllowed = cardinality
+        if (limits.bucket) {
```
hoorayimhelping, line 69: i couldn't find what the shape of `limits.bucket` or `limits.dashboard` etc are. Only thing that stands out to me is this truthy check might be wonky depending on the shape of `lmiits.bucket`

e.g. if `limits.bucket` is `undefined`, we're good. if `limits.bucket` is `{}`, this check will pass. I just don't have context into this, figured I'd mention it.

drdelambre: it also started begging the question as to why v2private isn't being routed through a swagger defs so that we could normalize this ish and guarantee a shape. i figure that pathway will be ironed out next quarter with the quartz work though and we can look at it then.

PR #17609, opened 2020-04-03
fix: hierarchy of variables is back

### LINKED ISSUE #17606: Variable Template Query Tag Values does not work for each bucket
Tools US 2.x Cluster
Dashboard Log File Studies V1.x - Shards

Variable shard defined as query:

v1.tagValues(bucket: v.bucket, tag: "shard_id")

The above sometimes works and sometimes does not based on the bucket.  For the following bucket, I actually created a cell to show that the value appears valid.

https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards/0569ce471d8dd000?lower=2020-04-01T01%3A00%3A00.000Z&upper=2020-04-01T02%3A00%3A00.000Z

image

I have run through a few dozen buckets (log files from customers) and sometimes the drop down works for the variable, other times it fails.  The l-creditkarma-apr bucket does not enable the drop down but the tag value list should just have the 10 numbers as shown above.

image

Ping me if you cannot repro the problem or if the bucket shown above is gone.

drdelambre:
Closes #17606

variables now hold their own values, and we gotta respect that, you know?

### DIFF
--- a/ui/src/dashboards/actions/thunks.ts
+++ b/ui/src/dashboards/actions/thunks.ts
@@ -286,7 +286,8 @@ export const getDashboard = (dashboardID: string) => async (
       throw new Error(resp.data.message)
     }
 
-    dispatch(hydrateVariables(true))
+    const skipCache = true
+    dispatch(hydrateVariables(skipCache))
 
     const normDash = normalize<Dashboard, DashboardEntities, string>(
       resp.data,
--- a/ui/src/dashboards/actions/thunks.ts
+++ b/ui/src/dashboards/actions/thunks.ts
@@ -286,7 +286,7 @@ export const getDashboard = (dashboardID: string) => async (
       throw new Error(resp.data.message)
     }
 
-    dispatch(hydrateVariables())
+    dispatch(hydrateVariables(true))
 
     const normDash = normalize<Dashboard, DashboardEntities, string>(
       resp.data,
--- a/ui/src/timeMachine/actions/queries.ts
+++ b/ui/src/timeMachine/actions/queries.ts
@@ -12,6 +12,7 @@ import {runStatusesQuery} from 'src/alerting/utils/statusEvents'
 
 // Actions
 import {notify} from 'src/shared/actions/notifications'
+import {hydrateVariables} from 'src/variables/actions/thunks'
 
 // Constants
 import {rateLimitReached, resultTooLarge} from 'src/shared/copy/notifications'
@@ -114,6 +115,8 @@ export const executeQueries = () => async (dispatch, getState: GetState) => {
 
   try {
     dispatch(setQueryResults(RemoteDataState.Loading, [], null))
+
+    await dispatch(hydrateVariables())
 
     //TODO: replace with activeContext selector
     const contextID =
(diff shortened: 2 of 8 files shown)


ui/src/timeMachine/actions/queries.ts
```diff
@@ -115,6 +116,8 @@ export const executeQueries = () => async (dispatch, getState: GetState) => {
   try {
     dispatch(setQueryResults(RemoteDataState.Loading, [], null))
 
+    await dispatch(hydrateVariables())
```
drdelambre, line 119: this refreshes variables that haven't been fetched by another means before sending off the query, just in case. found this while deep linking w/ dependent variables and no application state.

ui/src/variables/selectors/index.tsx
```diff
@@ -175,6 +175,11 @@ export const getVariable = (
   if (!vari.selected) {
     if (vari.arguments.type === 'map') {
       vari.selected = [Object.keys(vari.arguments.values)[0]]
+    } else if (
```
drdelambre, line 178: unrelated change, just an oversight I caught while debugging

ui/src/variables/selectors/index.tsx
```diff
@@ -252,6 +257,9 @@ export const asAssignment = (variable: Variable): VariableAssignment => {
   }
 
   if (variable.arguments.type === 'query') {
+    if (!variable.selected[0]) {
```
drdelambre, line 260: the ensures that the AST doesn't return variables if they don't have a value set, because that causes all sorts of issues in the LSP and on the api

ui/src/variables/utils/buildVarsOption.ts
```diff
@@ -20,7 +20,7 @@ export const buildVarsOption = (variables: VariableAssignment[]): File => ({
         },
         init: {
           type: 'ObjectExpression',
-          properties: variables.map(assignmentToProperty),
+          properties: variables.filter(v => !!v).map(assignmentToProperty),
```
drdelambre, line 23: and this makes sure those variables we removed are filtered from the list before further processing

ui/src/variables/utils/hydrateVars.ts
```diff
@@ -372,7 +372,10 @@ export const hydrateVars = (
     node.status === RemoteDataState.Loading
 
     try {
+      // TODO: remove the concept of node.values, just use node.variable
       node.values = await hydrateVarsHelper(node, options)
+      node.variable.arguments.values.results = node.values.values
```
drdelambre, line 377: here is the actual fix for this PR

ui/src/dashboards/actions/thunks.ts
```diff
@@ -286,7 +286,7 @@ export const getDashboard = (dashboardID: string) => async (
       throw new Error(resp.data.message)
     }
 
-    dispatch(hydrateVariables())
+    dispatch(hydrateVariables(true))
```
ebb-tide, line 289:
```suggestion
    const skipCache = true
    dispatch(hydrateVariables(skipCache))
```

drdelambre: is the request to be more explicit about what it's doing through the var name?

ebb-tide: yup :)

drdelambre: 💯

PR #17610, opened 2020-04-03
fix(storage): Fix query cursor read that caused cursor truncation.

benbjohnson:
## Overview

Filter cursors buffer points in between calls to Next() if the number of read points exceeds 1000. Previously, this buffer was being cleared out before being iterated over which caused queries to return a resultset which had a number of rows divisible by 1000.

This change moves the clearing of the buffer until after the points have been read. This change affects any queries which read more than 1000 points from a single series & have a filter that can be successfully applied to at least one of those points.

### DIFF
--- a/storage/reads/array_cursor.gen.go
+++ b/storage/reads/array_cursor.gen.go
@@ -55,8 +55,6 @@ func (c *floatArrayFilterCursor) Next() *cursors.FloatArray {
 
 	if c.tmp.Len() > 0 {
 		a = c.tmp
-		c.tmp.Timestamps = nil
-		c.tmp.Values = nil
 	} else {
 		a = c.FloatArrayCursor.Next()
 	}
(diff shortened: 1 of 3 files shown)


e-dard:
Looks like this is a similar bug to this one I fixed https://github.com/influxdata/influxdb/pull/16600?

stephanie-engel:
Just finished manually testing this fix:
- first, reproduced the existing issue on `master`
- second, verified the fix is working when running OSS on the current branch: `fix-cursor-buffer-read`

Now, I will be working on a test strategy for incorporating functional / acceptance tests in litmus. I will need to create a utility that will generate thousands of points. Then my overall test plan is to run queries that compare:
- overall count with no-op count
- overall count with slightly no-op count (example: `filter(fn: (r) => r._value >= 1)`
- overall count with op count (example: `filter(fn: (r) => r._value >= 50)`

PR #17615, opened 2020-04-03
refactor: fix imports from go template files

jsternberg:


pkg/data/gen/precision_string.go
```diff
@@ -4,6 +4,18 @@ package gen
 
 import "strconv"
 
+func _() {
```
jacobmarble, line 7: What does this do?

jsternberg: I'm not sure. It seems to be generated by the version of `stringer` that is in our `go.mod` file. This is the comment:

> // An "invalid array index" compiler error signifies that the constant values have changed.
> // Re-run the stringer command to generate them again.

PR #17617, opened 2020-04-05
fix: more variable dependancies

### LINKED ISSUE #17606: Variable Template Query Tag Values does not work for each bucket

drdelambre:
Closes #17606

In this PR, i get dependent variables working again, move the concept of 'current context' deeper into the stack, isolate the "variable node graph" concepts from leaking outside of `hydrateVars`, remove some copy pasta, and update the tests to capture some of the changes made

### DIFF
--- a/ui/src/variables/actions/thunks.ts
+++ b/ui/src/variables/actions/thunks.ts
@@ -290,7 +290,7 @@ export const moveVariable = (originalIndex: number, newIndex: number) => async (
   getState: GetState
 ) => {
   const contextID = currentContext(getState())
-  return dispatch(moveVariableInState(originalIndex, newIndex, contextID))
+  await dispatch(moveVariableInState(originalIndex, newIndex, contextID))
 }
 
 export const convertToTemplate = (variableID: string) => async (
--- a/ui/src/variables/actions/thunks.ts
+++ b/ui/src/variables/actions/thunks.ts
@@ -26,6 +26,7 @@ import {
   getVariable as getVariableFromState,
   getVariables as getVariablesFromState,
   getAllVariables as getAllVariablesFromState,
+  normalizeValues,
 } from 'src/variables/selectors'
 import {variableToTemplate} from 'src/shared/utils/resourceToTemplate'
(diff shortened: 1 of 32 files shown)


ui/src/variables/actions/thunks.ts
```diff
@@ -110,20 +131,24 @@ export const hydrateVariables = (skipCache?: boolean) => async (
...
 
-  vars
-    .filter(v => {
-      return vals[v.id] && v.arguments.type === 'query'
-    })
-    .forEach(v => {
-      v.arguments.values.results = vals[v.id].values
-      v.selected = vals[v.id].selected
-    })
+  const lookup = vals.reduce((prev, curr) => {
```
drdelambre, line 134: this change just reflects that hydratevars returns a list of variables now instead of the variablevalues[] thing

ui/src/variables/utils/hydrateVars.ts
```diff
@@ -124,11 +119,13 @@ const findSubgraph = (
 ): VariableNode[] => {
   const subgraph: Set<VariableNode> = new Set()
 
+  // use an ID array to reduce the chance of reference errors
+  const varIDs = variables.map(v => v.id)
```
drdelambre, line 123: this was the big fix

PR #17625, opened 2020-04-06
chore(ui): refactor logout to utilize new featureFlagging from IDPE

Asalem1:
The purpose of this PR is to update the way we are feature flagging the social sign-on feature. After meeting with Mark, Luke and Pat on Friday afternoon, it was decided that the best way to make sure that a rollout of the social sign-on feature happens smoothly would be to ensure that the feature flag we are using for IDPE and the UI are the same. Since we already have a feature flag in place to set specific config values for IDPE, this PR (in conjunction with https://github.com/influxdata/idpe/pull/6653) aims to update the way we are feature flagging the UI based on the feature flag that is being used in IDPE. 

### What does this mean?

Practically speaking, this means that we will be using a value that is returned from the API to determine whether to toggle the feature rather than using the built-in feature flagging system on the UI. As previously mentioned, we are doing this to ensure that the feature is rolled out simultaneously for the API and the UI.

### DIFF
--- a/ui/src/me/components/LogoutButton.tsx
+++ b/ui/src/me/components/LogoutButton.tsx
@@ -4,30 +4,12 @@ import {Link} from 'react-router'
 
 // Components
 import {Button, ComponentSize} from '@influxdata/clockface'
-import CloudExclude from 'src/shared/components/cloud/CloudExclude'
-import CloudOnly from 'src/shared/components/cloud/CloudOnly'
-import {CLOUD_URL, CLOUD_LOGOUT_PATH} from 'src/shared/constants'
-import {FeatureFlag} from 'src/shared/utils/featureFlag'
 
 const LogoutButton: SFC = () => (
   <>
-    <FeatureFlag name="regionBasedLoginPage">
-      <Link to="/logout">
-        <Button text="Logout" size={ComponentSize.ExtraSmall} />
-      </Link>
-    </FeatureFlag>
-    <FeatureFlag name="regionBasedLoginPage" equals={false}>
-      <CloudExclude>
-        <Link to="/logout">
-          <Button text="Logout" size={ComponentSize.ExtraSmall} />
-        </Link>
-      </CloudExclude>
-      <CloudOnly>
-        <a href={`${CLOUD_URL}${CLOUD_LOGOUT_PATH}`}>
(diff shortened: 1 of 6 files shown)


ui/src/me/components/LogoutButton.tsx
```diff
@@ -4,30 +4,12 @@ import {Link} from 'react-router'
...
 // Components
 import {Button, ComponentSize} from '@influxdata/clockface'
-import CloudExclude from 'src/shared/components/cloud/CloudExclude'
-import CloudOnly from 'src/shared/components/cloud/CloudOnly'
-import {CLOUD_URL, CLOUD_LOGOUT_PATH} from 'src/shared/constants'
-import {FeatureFlag} from 'src/shared/utils/featureFlag'
 
 const LogoutButton: SFC = () => (
   <>
-    <FeatureFlag name="regionBasedLoginPage">
```
Asalem1, line 14: In the process of this revision, I revisited the LogoutButton functionality and ensured that the existing `Logout` component functions identically, removing the need to conditionally render routes based on `CloudExclude` and `CloudOnly`

ui/src/pageLayout/components/UserWidget.tsx
```diff
@@ -49,44 +47,10 @@ const UserWidget: FC<Props> = ({
     return null
   }
 
-  const logoutURL = `${CLOUD_URL}${CLOUD_LOGOUT_PATH}`
-
   const handleSwitchOrganizations = (): void => {
     handleShowOverlay('switch-organizations', {}, handleDismissOverlay)
   }
 
-  const logoutLink = (
```
Asalem1, line 58: See comment above

ui/src/types/auth.ts
```diff
@@ -10,4 +10,5 @@ export type Auth0Config = {
   clientID: string
   domain: string
   redirectURL: string
+  socialSignUpOn: boolean
```
hoorayimhelping, line 13: nit: `toggled` might be more idiotmatic / conventional than `on` eg `socialSignUpToggled`.

Asalem1: Such a great suggestion. Owing to the fact that this is set on IDPE and k8s already as is (and will be a temporary property), I think it would be easiest to keep this naming as is for the time being. But I'll definitely bear in mind that boolean values should have more well-defined names

Issue #17623, opened 2020-04-06
InfluxDB 1.8: Incompatible write precision constant for v2 write URL


vlastahajek:
V2 compatibility URL doesn't support v2 constant for microseconds precision: `us`. It only accepts `u`.
__Steps to reproduce:__
```
curl -v -XPOST "http://localhost:8086/api/v2/write?org=my-org&bucket=test&precision=us" --data-raw "wifi_status,SSID=AB,device=ESP32 rssi=-33 1586189310000001"

* TCP_NODELAY set
* Connected to localhost (::1) port 8086 (#0)
> POST /api/v2/write?org=my-org&bucket=test&precision=us HTTP/1.1
> Host: localhost:8086
> User-Agent: curl/7.55.1
> Accept: */*
> Content-Length: 60
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 60 out of 60 bytes
< HTTP/1.1 204 No Content
< Content-Type: application/json
< Request-Id: c78e2ab5-7823-11ea-8037-d66d6d2aa003
< X-Influxdb-Build: OSS
< X-Influxdb-Version: 1.8.0rc0
< X-Request-Id: c78e2ab5-7823-11ea-8037-d66d6d2aa003
< Date: Mon, 06 Apr 2020 16:29:25 GMT
<
* Connection #0 to host localhost left intact
```
__Expected behavior:__
Write data.

__Actual behavior:__
Due to issue #17622 it returns status as data is successfully written, but it isn't.

glinton:
This seems to have been closed by https://github.com/influxdata/influxdb/pull/17638 comment to reopen if not.

Issue #17622, opened 2020-04-06
InfluxDB 1.x: No error and data are silently ignored in case of invalid time precision


vlastahajek:
__Steps to reproduce:__
Write data with invalid precision constant in the write URL:
```
curl -v -XPOST "http://localhost:8086/write?db=test&precision=x" --data-raw "wifi_status,SSID=AB,device=ESP32 rssi=-31 1586189309000003"

*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8086 (#0)
> POST /write?db=test&precision=x HTTP/1.1
> Host: localhost:8086
> User-Agent: curl/7.55.1
> Accept: */*
> Content-Length: 60
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 60 out of 60 bytes
< HTTP/1.1 204 No Content
< Content-Type: application/json
< Request-Id: 5516b4bc-7822-11ea-8032-d66d6d2aa003
< X-Influxdb-Build: OSS
< X-Influxdb-Version: 1.8.0rc0
< X-Request-Id: 5516b4bc-7822-11ea-8032-d66d6d2aa003
< Date: Mon, 06 Apr 2020 16:19:04 GMT
```
__Expected behavior:__
Either:
1 - write data
2 - return 500 as it InfluxDB 2 does

__Actual behavior:__
The server returns status as if data is written successfully
`< HTTP/1.1 204 No Content`

__Environment info:__
* System info: Windows 10 Professional
* InfluxDB version: 1.8.0rc1

dgnorton:
Fixed in 1.8.0rc3 by https://github.com/influxdata/influxdb/pull/17656

dgnorton:
Fixed in 1.8.0rc3 by #17656

PR #17638, opened 2020-04-07
fix: verify precision parameter in write requests

dgnorton:
Closes # https://github.com/influxdata/influxdb/issues/17623 and https://github.com/influxdata/influxdb/issues/17622

This change updates the v1 and v2 `/write` endpoints to verify values passed in the `precision` parameter. Invalid values will now return a `400 bad request` to the client.

### DIFF
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -812,22 +812,47 @@ func bucket2dbrp(bucket string) (string, string, error) {
 // of an "org" and "bucket" are mapped to v1 "database" and "retention
 // policies".
 func (h *Handler) serveWriteV2(w http.ResponseWriter, r *http.Request, user meta.User) {
+	precision := r.URL.Query().Get("precision")
+	switch precision {
+	case "ns":
+		precision = "n"
+	case "us":
+		precision = "u"
+	case "ms", "s", "":
+		// same as v1 so do nothing
+	default:
+		err := fmt.Sprintf("invalid precision %q (use ns, us, ms or s)", precision)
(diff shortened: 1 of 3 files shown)


services/httpd/handler.go
```diff
@@ -812,22 +812,47 @@ func bucket2dbrp(bucket string) (string, string, error) {
...
 
 // serveWriteV1 handles v1 style writes.
 func (h *Handler) serveWriteV1(w http.ResponseWriter, r *http.Request, user meta.User) {
-	h.serveWrite(r.URL.Query().Get("db"), r.URL.Query().Get("rp"), w, r, user)
+	precision := r.URL.Query().Get("precision")
+	switch precision {
+	case "", "n", "u", "ms", "s", "m", "h":
+		// it's valid
+	default:
+		err := fmt.Sprintf("invalid precision \"%s\" (use n, u, ms, s, m or h)", precision)
```
ayang64, line 843: nit -- i'd use the `%q` verb to quote precision here.

services/httpd/handler.go
```diff
@@ -812,22 +812,47 @@ func bucket2dbrp(bucket string) (string, string, error) {
...
+	precision := r.URL.Query().Get("precision")
+	switch precision {
+	case "ns":
+		precision = "n"
+	case "us":
+		precision = "u"
+	case "ms", "s", "":
+		// same as v1 so do nothing
+	default:
+		err := fmt.Sprintf("invalid precision \"%s\" (use ns, us, ms or s)", precision)
```
ayang64, line 824: nit -- would use %q

PR #17640, opened 2020-04-07
fix(CORS): stop processing in pre-flight CORS requests

sranka:
In CORS (https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS),
- the server stops processing when satisfying pre-flight requests
- Access-Control-Allow-Origin header is included in CORS-allowed responses

This fix is required for any browser client (including influxdb-client-js) to work fine against influxdb OSS. 

cc influxdata/influxdb-client-js#164 

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/kit/transport/http/middleware.go
+++ b/kit/transport/http/middleware.go
@@ -18,13 +18,18 @@ type Middleware func(http.Handler) http.Handler
 func SetCORS(next http.Handler) http.Handler {
 	fn := func(w http.ResponseWriter, r *http.Request) {
 		if origin := r.Header.Get("Origin"); origin != "" {
+			// Access-Control-Allow-Origin must be present in every response
 			w.Header().Set("Access-Control-Allow-Origin", origin)
+		}
+		if r.Method == http.MethodOptions {
+			// allow and stop processing in pre-flight requests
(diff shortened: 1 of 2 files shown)


kit/transport/http/middleware.go
```diff
@@ -18,13 +18,18 @@ type Middleware func(http.Handler) http.Handler
...
 		if origin := r.Header.Get("Origin"); origin != "" {
+			// Access-Control-Allow-Origin must be present in every response
 			w.Header().Set("Access-Control-Allow-Origin", origin)
+		}
+		if r.Method == http.MethodOptions {
+			// allow and stop processing in pre-flight requests
 			w.Header().Set("Access-Control-Allow-Methods", "POST, GET, OPTIONS, PUT, DELETE")
 			w.Header().Set("Access-Control-Allow-Headers", "Accept, Content-Type, Content-Length, Accept-Encoding, Authorization, User-Agent")
+			w.WriteHeader(http.StatusNoContent)
+		} else {
```
rawkode, line 29: Instead of the `else`, could the code return early to avoid branching?

sranka: yes, it also better to signalizes that the processing must be stopped

PR #17650, opened 2020-04-07
refactor(query/stdlib): modify storage filters to use the predicate directly

jsternberg:
The storage filters are modified to use the predicates directly so we do
not have to pass `semantic.FunctionExpression` around. Instead, since
simple expressions are all that are supported anyway, we transform
suitable function expressions into predicates as part of the push down
rule and this simplifies the influxdb reader code.

This also moves the storage predicate conversion code into the standard
library package as it is the only location that uses this code now that
the predicate conversion is done as part of the push down rule.

This refactor was prompted by another refactor of the
`semantic.FunctionExpression` that would cause it to always contain a
`semantic.Block`. Since the push down filter needs the expressions and
to combine them, this refactor allows us not do construct a combined
filter inside of blocks which allows us to have better type safety.


query/stdlib/influxdata/influxdb/storage_predicate.go
```diff
@@ -25,6 +23,39 @@ func toStoragePredicate(f *semantic.FunctionExpression) (*datatypes.Predicate, e
...
+	case ast.AndOperator:
+		value = datatypes.LogicalAnd
+	case ast.OrOperator:
+		value = datatypes.LogicalOr
+	default:
+		return nil, fmt.Errorf("unknown logical operator %v", op)
+	}
+
+	// Nest the predicates backwards. This way we get a tree like this:
+	// a AND (b AND c)
```
rockstar, line 42: Thanks for putting this here.

PR #17651, opened 2020-04-07
fix(ui): fixed dark lines in checks

### LINKED ISSUE #16923: Check view page renders time series with dark color
The time series line is not visible.
Additionally, the events shown do not fall within the time range of the graph that is being displayed

Asalem1:
Closes #16923

### Problem

CheckPlot line and font colors were not being mapped correctly onto the canvas

### Solution

Updated the swagger.yml to accurately reflect the color struct (from an array of strings to an array of DashboardColors). Added mapping logic to isolate the colors in an array so that the colors could be set accurately and added the default options.

- [x] Rebased/mergeable

check-color-normalized


ui/src/timeMachine/components/view_options/CheckOptions.tsx
```diff
@@ -1,49 +0,0 @@
-// Libraries
-import React, {FC} from 'react'
-import {connect} from 'react-redux'
-import {Form, Grid} from '@influxdata/clockface'
-
```
Asalem1, line 5: Since the `CheckPlot` is only ever used in alerting in the VEO, there is no option to select `CheckOptions`

ui/src/timeMachine/components/view_options/OptionsSwitcher.tsx
```diff
@@ -44,8 +43,6 @@ class OptionsSwitcher extends PureComponent<Props> {
         return <HeatmapOptions {...view.properties} />
       case 'scatter':
         return <ScatterOptions {...view.properties} />
-      case 'check':
```
Asalem1, line 47: `CheckOptions` is unused

ui/src/views/helpers/index.ts
```diff
@@ -308,7 +308,7 @@ const NEW_VIEW_CREATORS = {
           },
         },
       ],
-      colors: NINETEEN_EIGHTY_FOUR,
```
Asalem1, line 311: Switching over to a different default allows for a consistent color schema for the CheckViewProperties. This ensures that both the view history and default view are the same type

PR #17659, opened 2020-04-07
feat(tenant): update the client to use the new onboarding package

lyondhill:


cmd/influx/setup.go
```diff
@@ -45,18 +45,92 @@ func cmdSetup(f *globalFlags, opt genericCLIOpts) *cobra.Command {
...
+	s := tenant.OnboardClientService{
+		Client: client,
+	}
+
+	req, err := onboardingRequest()
+	if err != nil {
+		return fmt.Errorf("failed to retrieve data to setup instance: %v", err)
+	}
+
+	fmt.Printf("req: %+v\n", req)
```
AlirieGray, line 91: do we want this print in here?

PR #17665, opened 2020-04-07
fix: escaping tags in flux queries

### LINKED ISSUE #17529: Flux Query Builder is not working for fields with names containing "."
It is not possible to create charts without manual editing of flux command for fields like "service.name". 

__Steps to reproduce:__

Here is flux command generated by query builder:
image

generated filter is:
` |> filter(fn: (r) => r.service.name == “demo-rails-app”)`

should be:
`|> filter(fn: (r) => r["service.name"] == "demo-rails-app")`

__Environment info:__
* InfluxDB version: Version 2.0.0 (a3ff52d)

drdelambre:
Closes #17529

instead of using the dot notation, lets just wrap the fields so that we get some escaping for free

### DIFF
--- a/ui/src/timeMachine/reducers/index.test.ts
+++ b/ui/src/timeMachine/reducers/index.test.ts
@@ -52,7 +52,7 @@ describe('the Time Machine reducer', () => {
 
     it('builds the query if the query builder config is valid', () => {
       const expectedText =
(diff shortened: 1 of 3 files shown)


hoorayimhelping:
thanks for taking care of this one!

PR #17669, opened 2020-04-08
feat(task): Add worker count option to Executor

brettbuddin:
Introduces `WithMaxWorkers` option for `Executor` to control the number of workers allowed.

### DIFF
--- a/task/backend/executor/executor.go
+++ b/task/backend/executor/executor.go
@@ -17,6 +17,11 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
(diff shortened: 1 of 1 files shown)


brettbuddin:
We shouldn't have any issues with maximum number of connections due to a new transport created for every query. https://github.com/influxdata/influxdb/blob/86c6bce2bbdc45d48219871b6fa44c2b40de794c/http/client.go#L136-L147

This isn't ideal. Adding a note to come back to fix it, but it won't affect this change.

PR #17670, opened 2020-04-08
fix(query): Respect the now-time of the compiled query if it's provided

brettbuddin:
We are seeing a mismatch between the expected `now()` time of a Task run and what is actually reported at the time the task is executed. Upon further investigation we found that proxied query requests (via HTTP) aren't using the `Now` time provided by the Task Executor.

This teaches the HTTP layer to use the `Now` time provided if it is presented. Otherwise, it falls back to calculating a time with the epoch.


brettbuddin:
@nathanielc That is correct. I'll get on those changes.

brettbuddin:
@nathanielc Think we're all good here. I'll be opening another PR in IDPE for the integration tests.

PR #17681, opened 2020-04-08
feat: moved parser

drdelambre:
we have to make some weird jumps through spaghetti hoops to get the csv routed from an api response, to timemachine and giraffe for parsing, then back out to display meta data that then goes back into giraffe to choose how to display it.

this can get fixed by just moving the parser out of giraffe

i'm also hoping to get a couple of perf points by using array pointers to build the results table and not cloning of the results for parsing.

existing tests work, but the interface is a little different and readability has been sacrificed for reducing memory footprint where possible

possibly a prereq for #16474

### DIFF
--- a/ui/src/shared/utils/fromFlux.test.ts
+++ b/ui/src/shared/utils/fromFlux.test.ts
@@ -203,4 +203,10 @@ there",5
       '[576] query terminated: reached maximum allowed memory limits'
     )
   })
+
+  test('does nothing with nothing', () => {
+    expect(() => fromFlux('')).not.toThrow()
+    expect(() => fromFlux('\n\n\n\nyolo\ntrash\n\n\n\n')).not.toThrow()
+    expect(fromFlux('\n\n\n\nyolo\ntrash\n\n\n\n').table.length).toEqual(0)
+  })
 })
--- a/ui/src/shared/utils/fromFlux.test.ts
+++ b/ui/src/shared/utils/fromFlux.test.ts
@@ -1,4 +1,4 @@
-import {fromFlux} from './fromFlux'
+import fromFlux from './fromFlux'
 
 describe('fromFlux', () => {
   test('can parse a Flux CSV with mismatched schemas', () => {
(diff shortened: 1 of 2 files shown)


ui/src/shared/utils/fromFlux.ts
```diff
@@ -0,0 +1,262 @@
...
+         values containing newlines.
+
+         [0]: https://github.com/influxdata/influxdb/issues/15017
+     */
+  const output = {},
+    groupKey = {},
+    names = {}
+
+  const chunks = [],
+    regerz = /\n\s*\n#/g
```
Asalem1, line 115: this actually makes sense based on the comment above.

ui/src/shared/utils/fromFlux.test.ts
```diff
@@ -0,0 +1,206 @@
...
+#datatype,string,string,string,string
+#default,,,,
+,error,reference
+,query terminated: reached maximum allowed memory limits,576`
+
+    expect(() => fromFlux(CSV)).toThrowError(
+      '[576] query terminated: reached maximum allowed memory limits'
+    )
+  })
+})
```
Asalem1, line 206: Yo dawg, I heard you like tests. Can we get a test to see what happens if the input is defined || `''`

ui/src/shared/utils/fromFlux.ts
```diff
@@ -0,0 +1,262 @@
...
+const TO_COLUMN_TYPE = {
+  boolean: 'boolean',
+  unsignedLong: 'number',
+  long: 'number',
+  double: 'number',
+  string: 'string',
+  'dateTime:RFC3339': 'time',
+}
+
+function parseValue(value, columnType) {
```
Asalem1, line 12: I know your comments make illuminate the fact that the `columnType` is going to be a `js` type, but maybe we can assign an Enum to this?

ui/src/shared/utils/fromFlux.ts
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    if (
+      parsed[headerLocation][1] === 'error' &&
+      parsed[headerLocation][2] === 'reference'
+    ) {
+      const ref = parsed[headerLocation + 1][2]
+      const msg = parsed[headerLocation + 1][1]
+
+      throw new Error(`[${ref}] ${msg}`)
```
Asalem1, line 158: is this being handled anywhere?

ui/src/shared/utils/fromFlux.ts
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    for (no = 0; no < headerLocation; no++) {
+      annotations[parsed[no][0]] = parsed[no].reduce((p, c, i) => {
+        p[parsed[headerLocation][i]] = c
+        return p
+      }, {})
+    }
+
+    for (no = 1; no < parsed[headerLocation].length; no++) {
```
Asalem1, line 168: why are we resetting `no` here instead of using a new variable? Kinda makes it confusing

ui/src/shared/utils/fromFlux.ts
```diff
@@ -0,0 +1,262 @@
+import Papa from 'papaparse'
```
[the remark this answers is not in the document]
drdelambre: did you notice that i didn't have to update `package.json`? so it's already floating around in there somewhere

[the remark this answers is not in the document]
Asalem1: I know. It's mostly an issue with the developers of the repo naming it and the fact that we have to call it `Papa`. No knock on you, just stating my thoughts

ui/src/shared/utils/fromFlux.ts
drdelambre: i hope so! it'd be in the response parser to connect it to the ui

ui/src/shared/utils/fromFlux.test.ts
```diff
@@ -0,0 +1,212 @@
...
+#default,,,,
+,error,reference
+,query terminated: reached maximum allowed memory limits,576`
+
+    expect(() => fromFlux(CSV)).toThrowError(
+      '[576] query terminated: reached maximum allowed memory limits'
+    )
+  })
+
+  test('does nothing with nothing', () => {
```
drdelambre, line 207: added this test for nothing

ui/src/shared/utils/fromFlux.ts
```diff
@@ -0,0 +1,345 @@
...
+export interface ParsedFlux {
+  table: {
+    columnKeys: string[]
+    columns: ParsedFluxColumnLookup
+    length: number
+  }
+  fluxGroupKeyUnion: string[]
+}
+
+export default function fromFlux(csv: string): ParsedFlux {
```
drdelambre, line 141: here too!

ui/src/shared/utils/fromFlux.ts
```diff
@@ -0,0 +1,345 @@
...
+    start: lastRange,
+    stop: csv.length,
+  })
+
+  let runningTotal = 0,
+    currentChunkIndex,
+    headerLocation,
+    currentLineInHeader,
+    currentColumnIndex,
+    currentLineIndex,
```
drdelambre, line 182: look at these names... they describe EXACTLY what they're for

ui/src/shared/utils/fromFlux.ts
```diff
@@ -0,0 +1,345 @@
...
+    annotations,
+    parsed
+
+  for (
+    currentChunkIndex = 0;
+    currentChunkIndex < chunks.length;
+    currentChunkIndex++
+  ) {
+    annotations = {}
+    parsed = Papa.parse(
```
drdelambre, line 195: i hear library developers love it when you don't rename when declaring a library's default export definition as big papa.
:rip biggie:

PR #17685, opened 2020-04-08
fix(tsm1): Fix temp directory search bug

ayang64:
There's no bug associated with this.  Just found it while perusing the code.

    The original code's intention is to scan a directory for the directory
    with the higest value when converted to an integer.
    
    So directories may be in the form:
    
      0.tmp
      1.tmp
      2.tmp
      30.tmp
      ...
      100.tmp
    
    The loop should scan the directory, strip the basename and extension
    from the file name to leave just a number, then store the higest number
    it finds.
    
    Before this patch, there is a bug that has the code only store the
    higest value if there is an error converting the numeric value into an
    integer.
    
    This patch primarily fixes that logic.
    
    In addition, this patch will save an indent level by inverting logic in
    two places:
    
    Instead if checkig if a file is a directory and has a suffix of ".tmp",
    it is probably better to test if a file is NOT a directory OR does NOT
    have an extension of ".tmp" then continue.
    
    Also, instead of testig if len(ss) == 2, we can test if len(ss) != 2 and
    continue if so.
    
    Both of these save an indent level and keeps our "happy path" to the
    left.
    
    Finally, this patch will use string concatination instead of calling
    fmt.Sprintf() to add periods to "tmp" and "tsm" extension.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+v1.8.1 [unreleased]
+-------------------
+
+### Features
+
+### Bugfixes
+
+-	[#17638](https://github.com/influxdata/influxdb/pull/17638): Verify precision in write requests.
+
 v1.8.0 [unreleased]
 -------------------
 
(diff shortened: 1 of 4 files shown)


ayang64:
the offending old code is:

```
				if i, err := strconv.Atoi(ss[0]); err != nil {	
					if i > f.currentTempDirID {			ss := strings.Split(filepath.Base(fi.Name()), ".")
						f.currentTempDirID = i			if len(ss) != 2 {
					}				continue
				}	
```

notice how we only set `f.currentTempDirID` if `strconv.Atoi()` fails -- which would leave `i` as `0` typically (though there are other unlikely and equally wrong possibilities).

PR #17688, opened 2020-04-08
refactor: remove labels from bucket api

AlirieGray:
Closes https://github.com/influxdata/influxdb/issues/17689

This PR removes the `Labels` property from the `Buckets` object in the API, and updates the thunks to reflect this change.

This change is being implemented because the "get" routes each resource should be separate, rather than having Labels embedded in the Buckets routes.  This is a part of the Gateway API Refactor epic (epic can be seen on the above issue) and the ongoing work to compartmentalize each API so that teams can more easily make changes to the APIs they work with.

Users will still be able to get Labels for a Bucket using the `bucket/[id]/labels` route, and the behavior on the frontend should not change.

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/src/buckets/actions/thunks.ts
+++ b/ui/src/buckets/actions/thunks.ts
@@ -10,7 +10,6 @@ import {bucketSchema, arrayOfBuckets} from 'src/schemas'
 
 // Types
 import {
-  AppState,
   RemoteDataState,
   GetState,
   GenBucket,
@@ -19,12 +18,13 @@ import {
   Label,
   ResourceType,
   OwnBucket,
+  BucketWithLabel,
 } from 'src/types'
 
 // Utils
 import {getErrorMessage} from 'src/utils/api'
 import {getOrg} from 'src/organizations/selectors'
-import {getLabels, getStatus} from 'src/resources/selectors'
+import {getStatus} from 'src/resources/selectors'
 
 // Actions
 import {
@@ -51,6 +51,7 @@ import {
   removeBucketLabelFailed,
 } from 'src/shared/copy/notifications'
 import {LIMIT} from 'src/resources/constants'
+import {isSystemBucket} from 'src/buckets/selectors'
 
 type Action = BucketAction | NotifyAction
 
@@ -75,8 +76,17 @@ export const getBuckets = () => async (
 
     const demoDataBuckets = await fetchDemoDataBuckets()
 
+    const ownBuckets: Array<BucketWithLabel> = resp.data.buckets
+
+    for (let i = 0; i < ownBuckets.length; i++) {
+      if (!isSystemBucket(ownBuckets[i].type)) {
+        const labels = await getLabelsForBucketHelper(ownBuckets[i].id)
+        ownBuckets[i].labels = labels
+      }
+    }
+
     const buckets = normalize<Bucket, BucketEntities, string[]>(
-      [...resp.data.buckets, ...demoDataBuckets],
+      [...ownBuckets, ...demoDataBuckets],
       arrayOfBuckets
     )
 
@@ -85,6 +95,18 @@ export const getBuckets = () => async (
     console.error(error)
     dispatch(setBuckets(RemoteDataState.Error))
     dispatch(notify(getBucketsFailed()))
+  }
+}
+
+const getLabelsForBucketHelper = async (bucketID: string): Promise<api.Labels> => {
+  try {
+    const r = await api.getBucketsLabels({bucketID})
+    if (r.status != 200) {
+      throw new Error(r.data.message)
+    }
+    return r.data.labels
+  } catch (e) {
+    console.error(e)
   }
 }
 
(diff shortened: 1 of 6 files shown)


ui/src/buckets/actions/thunks.ts
```diff
@@ -254,10 +283,8 @@ export const deleteBucketLabel = (bucketID: string, label: Label) => async (
   }
 }
 
-const denormalizeBucket = (state: AppState, bucket: OwnBucket): GenBucket => {
-  const labels = getLabels(state, bucket.labels)
+const denormalizeBucket = (bucket: OwnBucket): GenBucket => {
```
Asalem1, line 286: we should probably just get rid of this

ui/src/buckets/actions/thunks.ts
```diff
@@ -51,6 +51,8 @@ import {
   removeBucketLabelFailed,
 } from 'src/shared/copy/notifications'
 import {LIMIT} from 'src/resources/constants'
+import {getBucketsLabels} from 'src/client'
+import {isSystemBucket} from '../selectors'
```
Asalem1, line 55:
```suggestion
import {isSystemBucket} from 'src/buckets/selectors'
```

ui/src/buckets/actions/thunks.ts
```diff
@@ -88,6 +99,18 @@ export const getBuckets = () => async (
   }
 }
 
+const getLabelsForBucketHelper = async (bucketID: string) => {
```
Asalem1, line 102:
```suggestion
const getLabelsForBucketHelper = async (bucketID: string): Label[] => {
```

ui/src/buckets/actions/thunks.ts
```diff
@@ -51,6 +51,8 @@ import {
   removeBucketLabelFailed,
 } from 'src/shared/copy/notifications'
 import {LIMIT} from 'src/resources/constants'
+import {getBucketsLabels} from 'src/client'
```
Asalem1, line 54:
```suggestion
```

ui/src/buckets/actions/thunks.ts
```diff
@@ -88,6 +99,18 @@ export const getBuckets = () => async (
   }
 }
 
+const getLabelsForBucketHelper = async (bucketID: string) => {
+  try {
+    const r = await getBucketsLabels({bucketID: bucketID})
```
Asalem1, line 104:
```suggestion
    const r = await api.getBucketsLabels({bucketID})
```

ui/src/buckets/actions/thunks.ts
```diff
@@ -101,10 +124,10 @@ export const createBucket = (bucket: OwnBucket) => async (
       throw new Error(resp.data.message)
     }
 
-    const newBucket = normalize<Bucket, BucketEntities, string>(
-      resp.data,
-      bucketSchema
-    )
+    const b: BucketWithLabel = resp.data
+    b.labels = []
```
Asalem1, line 128: I think this should be taken care of by normalization?

ui/src/buckets/actions/thunks.ts
```diff
@@ -116,12 +139,10 @@ export const createBucket = (bucket: OwnBucket) => async (
...
 
 export const updateBucket = (bucket: OwnBucket) => async (
-  dispatch: Dispatch<Action>,
-  getState: GetState
+  dispatch: Dispatch<Action>
 ) => {
   try {
-    const state = getState()
-    const data = denormalizeBucket(state, bucket)
+    const data = denormalizeBucket(bucket)
```
Asalem1, line 145: Since the purpose of `denormalizeBucket` was to add a labels to the bucket and update the type of bucket to be a different type, this no longer seems relevant

ui/src/buckets/actions/thunks.ts
AlirieGray: If I get rid of this part then it crashes when trying to create a new bucket. (Same for deleting, etc.) Is it possible I need to make a change to the normalization or something else?

ui/src/buckets/actions/thunks.ts
AlirieGray: this gets used on line 104 to get the labels for the bucket. is there a better way to make this api call?

ui/src/buckets/actions/thunks.ts
Asalem1: I imagine that this is crashing because the api is expecting the Buckets to resemble the GenBuckets type, when in reality they've been merged to include specific properties like labels and readableRetention. If we want to change the type to GenBuckets we should probably strip those properties in the function, since I imagine those are still being passed in

ui/src/buckets/actions/thunks.ts
AlirieGray: I think the Bucket does need to go from `OwnBucket` type (with a labels property) to `GenBucket` type (no labels). If I remove this part then it crashes.

ui/src/buckets/actions/thunks.ts
AlirieGray: never mind, I just saw your comment below where I can do api.getBucketsLabels. thank you!

ui/src/buckets/actions/thunks.ts
Asalem1: totally:

https://github.com/influxdata/influxdb/pull/17688#discussion_r406321100

If that's the case then we should strip the properties that make it an `OwnBucket` so that the type change accurately reflects the type it's being changed to. More specifically, we can ensure that the `labels` and `readableRetention` are removed before forcing the type change

AlirieGray:
closing per discussion with @ebb-tide @aanthony1243 et al

Issue #6583, opened 2016-05-09
Odd behaviour of time series containing the ":" char in its name


Hypfer:
### Bug report

**System info:** Default config on Debian Jessie. Influx v0.12.2

**Steps to reproduce:**
1. Install Influx v0.12.2
2. Push data to a time series containing ":" in its name e.g. curl -i -XPOST 'http://localhost:8086/write?db=foo' --data-binary 'HALLO\ : value=0.64'
3. Stop Influx
4. Start Influx
5. Try querying this time series. 

**Expected behavior:** Should work fine

**Actual behavior:** Sometimes it looses all data permanently; sometimes it is empty but if you push again everything is back. Just very odd

**Additional info:** Fresh install. No errors in influxdb.log

e-dard:
@Hypfer `:` is not a special character and does not need to be escaped. There are a couple of issues with using `\` as a literal character, which I think  you may be running into.

jasase:
There is a white space between `HALLO` and `:` that need to be escaped.

e-dard:
@jasase are sorry I missed that. I will look into this more today.

jwilder:
I'm not able to reproduce this with 1.1.

PR #17690, opened 2020-04-08
feat(demodata): Demodata UI polish

### LINKED ISSUE #6581: SHOW TAG VALUES WITH KEY WHERE fails if the tag name is "name"
### Bug report

InfluxDB 0.11.1 running on top of Debian Jessie(8.4)

I have a measurement, virtualmachine, which contains data like this(simplified version):
```
host=vcenter-01.adm,cluster=cluster-01,name=vmname-01,datastore=blabla32 cpu_usage_average=1531513,mem_usage_average=135313
```

Using Grafana, i've tried to create a template variable with the datastore(a tag) of a currently selected VM(by the tag name, in another template variable).

Expected behavior:

```
SHOW TAG VALUES WITH KEY ="datastore" WHERE "name" =~ /$name/	
```
I hoped that would give me the datastore on which $name is residing.

However, I get all the datastores, regardless of the value of $name.

I find this behaviour only with the tag "name", if i filter on the tag $host, i get proper data(filtered by host).

I'll probably end up using SELECT instead of SHOW VALUES, but nonetheless i find it weird that the tag "name" doesn't work properly(contrary to other tags).

ebb-tide:
Closes #6581 #6583 # 6584
This is probably a PR to review commit by commit. 
I adjusted UI display of demodata buckets
I added demodata dashboard template
I added logic for creating the dashboard at request for bucket membership
I made it so demo data buckets that have already been instantiated do not show up on add demodata dropdown
Add delete button functionality to demodata buckets
Delete demodata dashboard if it exists


ui/src/buckets/components/BucketList.tsx
```diff
@@ -91,7 +95,13 @@ class BucketList extends PureComponent<Props & WithRouterProps> {
 
     return sortedBuckets.map(bucket => {
       if (bucket.type === 'demodata') {
-        return <DemoDataBucketCard key={bucket.id} bucket={bucket} />
+        return (
+          <DemoDataBucketCard
+            key={bucket.id}
+            bucket={bucket}
+            onRemoveBucket={deleteDemoDataBucketMembership}
```
drdelambre, line 102: this should go into the demodatabucketcard component and not be added through a prop. also, don't you need a `connect` for actions to work?

ebb-tide: it's prop drilled further than you think!!

ui/src/cloud/actions/demodata.ts
```diff
@@ -46,39 +57,77 @@ export const getDemoDataBuckets = () => async (
...
+  const {id: orgID} = getOrg(state)
 
   try {
-    await deleteDemoDataBucketMembershipAJAX(bucketID, userID)
+    await deleteDemoDataBucketMembershipAJAX(bucket.id, userID)
     dispatch(getBuckets())
 
-    // TODO: check for success and error appropriately
-    // TODO: delete associated dashboard
+    const dashboardsResp = await api.getDashboards({query: {orgID}})
```
drdelambre, line 103: why do we need to refresh the dashboards here? and why isn't it using one of the dashboard actions that updates the normalized state?

ui/src/cloud/constants/index.ts
```diff
@@ -7,3 +9,1104 @@ export const ASSET_LIMIT_ERROR_STATUS = 403
 
 export const ASSET_LIMIT_ERROR_TEXT =
   'Oops. It looks like you have exceeded the asset limits allowed as part of your plan. If you would like to increase your limits, reach out to support@influxdata.com.'
+
+export const WebsiteMonitoringDashboardTemplate = (name: string) => {
```
drdelambre, line 13: this should be loaded async to absorb the impact on the bundle

ui/src/buckets/components/BucketList.tsx
ebb-tide: but I'm fixing it.

ui/src/cloud/selectors/demodata.ts
```diff
@@ -0,0 +1,11 @@
+import {get, differenceBy} from 'lodash'
+import {AppState, Bucket, DemoBucket} from 'src/types'
+
+export const getNewDemoBuckets = (state: AppState, ownBuckets: Bucket[]) => {
+  const demoDataBuckets = get(
+    state,
+    'cloud.demoData.buckets',
+    []
+  ) as DemoBucket[]
+  return differenceBy(demoDataBuckets, ownBuckets, b => b.id)
```
drdelambre, line 10: is there an explicit default sort that should be applied here?

ebb-tide: will alphabetize

ui/src/cloud/actions/demodata.ts
```diff
@@ -4,10 +4,29 @@ import {
   getDemoDataBucketMembership as getDemoDataBucketMembershipAJAX,
   deleteDemoDataBucketMembership as deleteDemoDataBucketMembershipAJAX,
 } from 'src/cloud/apis/demodata'
+import {createDashboardFromTemplate} from 'src/templates/api'
+import * as api from 'src/client'
```
drdelambre, line 8: are you using all of the generated client imports?

ui/src/cloud/actions/demodata.ts
```diff
@@ -46,26 +65,63 @@ export const getDemoDataBuckets = () => async (
...
     me: {id: userID},
-  } = getState()
+  } = state
+  const {id: orgID} = getOrg(state)
 
   try {
-    await getDemoDataBucketMembershipAJAX(bucketID, userID)
-
+    await getDemoDataBucketMembershipAJAX(bucket.id, userID)
     dispatch(getBuckets())
```
drdelambre, line 80: why is it required to refresh all buckets here instead of maintaining the state changes of a single bucket? shouldn't this be an await also?

ui/src/cloud/actions/demodata.ts
```diff
@@ -74,11 +130,19 @@ export const deleteDemoDataBucketMembership = (bucketID: string) => async (
   } = getState()
 
   try {
-    await deleteDemoDataBucketMembershipAJAX(bucketID, userID)
+    await deleteDemoDataBucketMembershipAJAX(bucket.id, userID)
```
drdelambre, line 133: this seems like a duplicate pattern from `getDemoDataBucketMembership`

ui/src/cloud/actions/demodata.ts
```diff
@@ -46,26 +65,63 @@ export const getDemoDataBuckets = () => async (
...
 
   try {
-    await getDemoDataBucketMembershipAJAX(bucketID, userID)
-
+    await getDemoDataBucketMembershipAJAX(bucket.id, userID)
     dispatch(getBuckets())
-    // TODO: check for success and error appropriately
-    // TODO: instantiate dashboard template
+
+    const template = DemoDataTemplates[bucket.name]
```
drdelambre, line 82: this is an async function now right?

ui/src/cloud/actions/demodata.ts
ebb-tide: added await. made a request for a single bucket

ui/src/cloud/actions/demodata.ts
ebb-tide: made single bucket request here as well.

ui/src/cloud/actions/demodata.ts
ebb-tide: it's an object

PR #17697, opened 2020-04-09
fix(flux): return internal influxdb error for cursor conflict

jlapacik:

### DIFF
--- a/storage/flux/reader.go
+++ b/storage/flux/reader.go
@@ -17,6 +17,32 @@ import (
 	"github.com/influxdata/influxdb/v2/tsdb/cursors"
 )
 
+// GroupCursorError is returned when two different cursor types
+// are read for the same table.
+type GroupCursorError struct {
+	typ    string
+	cursor cursors.Cursor
+}
+
+func (err *GroupCursorError) Error() string {
+	var got string
+	switch err.cursor.(type) {
+	case cursors.FloatArrayCursor:
+		got = "float"
+	case cursors.IntegerArrayCursor:
+		got = "integer"
+	case cursors.UnsignedArrayCursor:
+		got = "unsigned"
+	case cursors.StringArrayCursor:
+		got = "string"
+	case cursors.BooleanArrayCursor:
+		got = "boolean"
+	default:
+		got = "invalid"
+	}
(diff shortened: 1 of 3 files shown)


storage/flux/table.gen.go
```diff
@@ -182,7 +182,13 @@ func (t *floatGroupTable) advanceCursor() bool {
 		if typedCur, ok := cur.(cursors.FloatArrayCursor); !ok {
 			// TODO(sgc): error or skip?
 			cur.Close()
-			t.err = errors.Errorf("expected float cursor type, got %T", cur)
+			t.err = &influxdb.Error{
+				Code: influxdb.EInternal,
```
jsternberg, line 186: This will report itself as "An internal server error occurred" just like the previous error. Is this intentional?

jlapacik: No that is not intentional. Is that because `Msg` is nil?

jsternberg: It's because of the code. The `EInternal` code indicates it is an internal error that shouldn't be shown to users.

jlapacik: Should it be `Invalid` instead?

jsternberg: Yes, but probably try to make it a more informed error message like "attempted to combine rows of different types" or something. Then use the internal error as the nested error on the `influxdb.Error` type.

storage/flux/reader.go
```diff
@@ -17,6 +17,32 @@ import (
...
+	case cursors.UnsignedArrayCursor:
+		got = "unsigned"
+	case cursors.StringArrayCursor:
+		got = "string"
+	case cursors.BooleanArrayCursor:
+		got = "boolean"
+	default:
+		got = "invalid"
+	}
+	return fmt.Sprintf("schema collision: cannot group %s and %s types together", err.typ, got)
```
jlapacik, line 43: @jsternberg does this error message need more clarity?

jsternberg: Oh I didn't notice this. This error message is great. Yes, just use `EInvalid` and you'll be fine.

jlapacik: Done.

Issue #17708, opened 2020-04-10
Table cells not exported/imported correctly in Packager


russorat:
This is the view of the dashboard before I exported it. Notice all the Table cells have two columns:
![image](https://user-images.githubusercontent.com/4805997/79021115-92839e00-7b2f-11ea-9af5-f06e50552a3e.png)

This is the file what was created when i called `influx pkg export --dashboard id`:
https://raw.githubusercontent.com/russorat/community-templates/master/ufo-sightings/ufo_data.yml

When i imported the dashboard, all the table cells only showed a single column.
![image](https://user-images.githubusercontent.com/4805997/79021083-7aac1a00-7b2f-11ea-8d19-b3d2f5df8ce0.png)

jsteenb2:
@russorat can you provide the raw data for this dashboard?

russorat:
[2020-04-11-13-42_chronograf_data.csv.zip](https://github.com/influxdata/influxdb/files/4465381/2020-04-11-13-42_chronograf_data.csv.zip)

jsteenb2:
@russorat I was asking for the json payload of the dashboard the UI retrieved in that first picture 😬

PR #17720, opened 2020-04-13
fix(pkger): fixup field misassignment in table view export

jsteenb2:
closes: #17708

`influx pkg export --dashboards $ID | influx pkg` results in:
image

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/pkger/models.go
+++ b/pkger/models.go
@@ -2656,9 +2656,6 @@ func (c chart) properties() influxdb.ViewProperties {
 				Visible:      fieldOpt.Visible,
 			})
 		}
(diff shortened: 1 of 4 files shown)


pkger/models.go
```diff
@@ -2656,9 +2656,6 @@ func (c chart) properties() influxdb.ViewProperties {
 				Visible:      fieldOpt.Visible,
 			})
 		}
-		sort.Slice(fieldOptions, func(i, j int) bool {
```
jsteenb2, line 2659: the order it comes in needs to respected, the UI is consuming it directly it turns out

Issue #12137, opened 2019-02-25
Panic on TSI Index Query


wruyi:
**version:v1.7.3**
**Engine: ts1**
**index-version = "tsi1"**

during a large number of writting, exec a query  "SELECT * FROM mydb.rp_name.weather WHERE altitude = '7'" ,then panic

-------------------------------------------------------------------------------------------------------------------
[httpd] 127.0.0.1 - null [25/Feb/2019:15:51:38 +0800] "POST /write?db=mydb&rp=rp_name&precision=n&consistency=one HTTP/1.1" 204 0 "-" "okhttp/3.13.1" 2e6db61a-38d2-11e9-b2a3-00ffad9b1641 40039
[httpd] 127.0.0.1 - null [25/Feb/2019:15:51:38 +0800] "POST /write?db=mydb&rp=rp_name&precision=n&consistency=one HTTP/1.1" 204 0 "-" "okhttp/3.13.1" 2e74b3bb-38d2-11e9-b2a4-00ffad9b1641 43945
ts=2019-02-25T07:51:38.577884Z lvl=info msg="Executing query" log_id=0Dm6XihG000 service=query query="SELECT * FROM mydb.rp_name.weather WHERE altitude = '7'"
unexpected fault address 0x1d72a005
fatal error: fault
[signal 0xc0000005 code=0x0 addr=0x1d72a005 pc=0xc0dd65]

goroutine 2228683 [running]:
runtime.throw(0x1731a8f, 0x5)
	C:/Go/src/runtime/panic.go:608 +0x79 fp=0xc0415aa708 sp=0xc0415aa6d8 pc=0x4339e9
runtime.sigpanic()
	C:/Go/src/runtime/signal_windows.go:207 +0x139 fp=0xc0415aa738 sp=0xc0415aa708 pc=0x448679
github.com/influxdata/roaring.(*shortIterator).next(0xc03834bbe0, 0x0)
	D:/Go-WorkSpace/src/github.com/influxdata/roaring/shortiterator.go:18 +0x35 fp=0xc0415aa750 sp=0xc0415aa738 pc=0xc0dd65
github.com/influxdata/roaring.(*intIterator).Next(0xc02435e450, 0x0)
	D:/Go-WorkSpace/src/github.com/influxdata/roaring/roaring.go:239 +0x4b fp=0xc0415aa778 sp=0xc0415aa750 pc=0xbd7e3b
[...]

Issue #14455, opened 2019-07-25
Segfaults in 1.7.7 with TSI enabled


samhld:
(Log attached at bottom)

__Environment info:__

* InfluxDB OSS 1.7.7
* TSI enabled

__Config:__
```
bind-address = ":8088"
reporting-disabled = false

[admin]
bind-address = ":8083"
enabled = true
https-certificate = "/etc/ssl/influxdb.pem"
https-enabled = false

[[collectd]]
enabled = false

[continuous_queries]
enabled = true
log-enabled = true
run-interval = "1s"

[coordinator]
log-queries-after = "5s"
max-concurrent-queries = 0
max-select-buckets = 0
max-select-point = 0
max-select-series = 0
query-timeout = "0s"
write-timeout = "10s"

[data]
cache-max-memory-size = "8g"
cache-snapshot-memory-size = 26214400
cache-snapshot-write-cold-duration = "10m"
compact-full-write-cold-duration = "4h"
dir = "/var/lib/influxdb/data"
engine = "tsm1"
index-version = "tsi1"
max-concurrent-compactions = 0
max-points-per-block = 0
max-series-per-database = 0
query-log-enabled = true
trace-logging-enabled = false
wal-dir = "/var/lib/influxdb/wal"
wal-logging-enabled = true

[[graphite]]
enabled = false

[http]
auth-enabled = false
bind-address = ":8086"
enabled = true
https-certificate = "/etc/ssl/influxdb.pem"
https-enabled = false
https-private-key = ""
log-enabled = false
max-connection-limit = 0
max-row-limt = 10000
pprof-enabled = false
realm = "InfluDB"
shared-secret = ""
write-tracing = false

[meta]
dir = "/var/lib/influxdb/meta"
logging-enabled = true
retention-autocreate = true

[monitor]
store-database = "_internal"
store-enabled = true
store-interval = "10s"

[[opentsdb]]
enabled = false

[retention]
check-interval = "30m0s"
enabled = true

[shard-precreation]
advance-period = "30m0s"
check-interval = "10m0s"
enabled = true

[subscriber]
enabled = true
http-timeout = "30s"

[[udp]]
enabled = false
```

__Logs:__
[influxdb_segfault_out.log](https://github.com/influxdata/influxdb/files/3432631/influxdb_segfault_out.log)

samhld:
[...]

PR #17721, opened 2020-04-13
fix(tsdb/tsi1): Remove tag value cache.

benbjohnson:
## Overview

The tag value cache stores series ID sets which, because of direct mmapping of Roaring bitmap containers, can potentially leak unmapped data which could cause a SIGSEGV.

See also: https://github.com/influxdata/influxdb/issues/14455, https://github.com/influxdata/EAR/issues/1038, https://github.com/influxdata/EAR/issues/1485

### DIFF
--- a/tsdb/config.go
+++ b/tsdb/config.go
@@ -67,9 +67,6 @@ const (
 	// write-ahead log file will compact into an index file.
 	DefaultMaxIndexLogFileSize = 1 * 1024 * 1024 // 1MB
 
-	// DefaultSeriesIDSetCacheSize is the default number of series ID sets to cache in the TSI index.
-	DefaultSeriesIDSetCacheSize = 100
-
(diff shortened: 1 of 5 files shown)


foobar:
looks like this also fixed #12137?

benbjohnson:
Closed in favor of disabling cache via https://github.com/influxdata/influxdb/pull/18158

benbjohnson:
@foobar Yes, it should fix that issue as well.

foobar:
seems the crash still happens with 1.8.3. Please re-open this issue

coguin:
I have this problem too. I don't understand the explanation that "The tag value cache stores series ID sets which, because of direct mmapping of Roaring bitmap containers, can potentially leak unmapped data which could cause a SIGSEGV."   TagValueSeriesIDCache in cache.go is s map, allocated in heap,  seems irrelevant to the mmap.   Could you explain it in more detail?  
And Is it possible that it;s because a bug in RoaringBitmap ?

PR #17724, opened 2020-04-13
fix(ui): add component to route hierarchy to handle org renaming

TCL735:
Closes idpe 6730

Adds the component for renaming orgs to the correct place in the routing hierarchy

### DIFF
--- a/ui/src/index.tsx
+++ b/ui/src/index.tsx
@@ -404,6 +404,12 @@ class Root extends PureComponent {
                               />
                             </Route>
                             <Route path="labels" component={LabelsIndex} />
+                            <Route path="about" component={OrgProfilePage}>
+                              <Route
+                                path="rename"
+                                component={RenameOrgOverlay}
+                              />
+                            </Route>
                           </Route>
                           <Route path="alerting" component={AlertingIndex}>
                             <Route
@@ -443,9 +449,7 @@ class Root extends PureComponent {
                             path="checks/:checkID"
                             component={CheckHistory}
                           />
-                          <Route path="about" component={OrgProfilePage}>
-                            <Route path="rename" component={RenameOrgOverlay} />
-                          </Route>
+                          <Route path="about" component={OrgProfilePage} />
                           {!CLOUD && (
                             <Route path="members" component={MembersIndex} />
                           )}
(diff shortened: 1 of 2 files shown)


drdelambre:
wasn't there something about replacing links to fix this one?

TCL735:
There was, but i wasn't present when that was discussed. Not sure what is meant by using anchor tags.

ui/src/index.tsx
```diff
@@ -443,9 +449,7 @@ class Root extends PureComponent {
                             path="checks/:checkID"
                             component={CheckHistory}
                           />
-                          <Route path="about" component={OrgProfilePage}>
-                            <Route path="rename" component={RenameOrgOverlay} />
-                          </Route>
+                          <Route path="about" component={OrgProfilePage} />
```
Asalem1, line 452: I'm confused about why we're setting the path here if it already exists here:

https://github.com/influxdata/influxdb/pull/17724/files#diff-58cea1fdba4d8d7058e35cb66db3a171R407

drdelambre:
@hoorayimhelping want to weigh in here? he did some fixes around the same area. iirc, we need to audit our links and migrate the external directed ones (that aren't just being passed around inside of react-router as a SPA) to being link tags.

ui/src/index.tsx
TCL735: The hierarchy was changed when the tree navigation was put in place.

**/orgs/:orgId/about** is the page that handles showing the Organization Profile, including the button that will rename the org.

**/orgs/:orgId/settings/about/rename** is where the overlay appears when renaming the org.

**/orgs/:orgId/settings/about** is the landing page after renaming an org.

All three are needed.

hoorayimhelping:
@TCL735 @drdelambre I think we're okay here with respect to turning things to anchors. That is more relevant to `Link` elements rather than `Route`s.

[Here's a commit](https://github.com/influxdata/influxdb/commit/e55c08514179cd416f8ded136163fe902304a1a6) I did fixing a problem for `/checkout` links. What happens is ReactRouter takes `Link`s and tries to route them internally. It doesn't see a route for `/checkout`, since it's an external link, and throws a 404.

PR #17732, opened 2020-04-14
fix(http/write_handler): propagate context to http request

sranka:
_http.NewRequestWithContext_ (available since golang 1.13) ensures that the supplied context also controls the entire lifetime of a request and its response. Before this PR, the supplied context was not used at all.

- [x] Rebased/mergeable
- [x] Tests pass


stuartcarnie:
The effect of this change is the lifetime of the upstream write request is based on the provided `Context`. In other words, the write request will fail if the context terminates for any reason (e.g. canceled or times-out).

PR #17733, opened 2020-04-14
improvement(query): performance improvement for sorted merge iterator

### LINKED ISSUE #8304: Find a way to futher optimize GROUP BY * queries

foobar:
Sorted merge iterator has cpu-intensive operations to sort the points
from multiple inputs. Typical queries like `SELECT * FROM m GROUP BY *`
do not behave well due to the comparison of points though in many cases
it doesn't necessarily have to use the slow path.

This patch adds a shortcut.  If each input has a single and unique
series we can just return the points input by input.
The detection of the shortcut introduces slight overhead but the gains
are significant in many slow queries.

Closes #8304

### DIFF
--- a/query/iterator.gen.go
+++ b/query/iterator.gen.go
@@ -11,6 +11,7 @@ import (
 	"context"
 	"io"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
@@ -374,6 +375,7 @@ func (itr *floatSortedMergeIterator) pop() (*FloatPoint, error) {
 			}
 			itr.heap.items = append(itr.heap.items, item)
 		}
+		itr.heap.detectFast()
 		heap.Init(itr.heap)
 		itr.init = true
 	}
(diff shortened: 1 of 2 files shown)


jacobmarble:
@dgnorton this looks like a backport of #17596 , which you merged earlier today. Worth adding to 1.8?

benbjohnson:
@jacobmarble I think it's good to keep versions consistent. We usually merge into `master-1.x` and backport from there as far back as we need. Anything that goes into `1.7` first should be forward ported to `1.8` and `master-1.x` so we don't hit performance regressions when users upgrade.

benbjohnson:
@jacobmarble Sorry, I misread the versions. I thought this was another instance of forward porting.

jacobmarble:
This is not a strict backport of #17596. Formatting is off here, and tests are missing. If needed, we can backport the change later.

PR #17754, opened 2020-04-15
fix(ui): add error messaging for cells in dashdboard view

### LINKED ISSUE #17594: Memory allocation errors no longer show on dashboards
this is in OSS. When you have a dashboard cell that returns a "Memory allocation limit error...", the dashboard cell just shows a black screen. When i click edit on the cell, I can see the memory error.

We need to figure out some error message to show to the user when this happens on their dashboard. 

image

This is what i see when I open the cell:
image

TCL735:
Closes #17594 

Add error messaging for cells in the Dashboard View.

Example: when one cell has error
Screen Shot 2020-04-15 at 11 25 11 AM

Example: when all cells have error
Screen Shot 2020-04-15 at 11 25 39 AM

### DIFF
--- a/ui/src/shared/utils/checkQueryResult.ts
+++ b/ui/src/shared/utils/checkQueryResult.ts
@@ -13,7 +13,7 @@
 
   See https://github.com/influxdata/flux/blob/master/docs/SPEC.md#errors.
 */
-export const checkQueryResult = (file: string): void => {
+export const checkQueryResult = (file: string = ''): void => {
   // Don't check the whole file, since it could be huge and the error table
   // will be within the first few lines (if it exists)
   const fileHead = file.slice(0, findNthIndex(file, '\n', 6))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 1. [17694](https://github.com/influxdata/influxdb/pull/17694): Fixed typos in the Flux functions list
 1. [17701](https://github.com/influxdata/influxdb/pull/17701): Allow mouse cursor inside Script Editor for Safari
 1. [17609](https://github.com/influxdata/influxdb/pull/17609): Fixed an issue where Variables could not use other Variables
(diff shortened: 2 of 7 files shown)


Asalem1:
I'm wondering if this might be related to:

https://github.com/influxdata/idpe/issues/6788

alexpaxton:
image

Overall looks good, there are these massive gaps between the error box and the cell header. Not sure where that gap is coming from, but I am happy to pair on removing it if that helps

ui/src/shared/utils/checkQueryResult.ts
```diff
@@ -13,7 +13,7 @@
 
   See https://github.com/influxdata/flux/blob/master/docs/SPEC.md#errors.
 */
-export const checkQueryResult = (file: string): void => {
+export const checkQueryResult = (file: string = ''): void => {
```
drdelambre, line 16: what does this do?

TCL735: **checkQueryResult** is called here https://github.com/influxdata/influxdb/blob/b653c2f975ae6be20e99b580e66058fb7af5a788/ui/src/shared/components/TimeSeries.tsx#L231

At runtime, `result.csv` may be undefined (not a string) when there is an error. The code immediately above the call to **checkQueryResult** throws a runtime error which will hide the fact that **checkQueryResult** will throw its own error when `result.csv` is not a string. By adding a default value to the parameter we prevent checkQueryResult from throwing its own error (which would be visible in the console should the throws above it are ever removed).

TCL735: Specifically, the error messaging will become checkQueryResult's thrown error rather than the actual error. And we want the error message to be useful to the user (rather than the dev). The user won't care or know what to do with "Cannot read property 'slice' of undefined" means.

Screen Shot 2020-04-15 at 11 43 58 AM

TCL735:
@alexpaxton Gaps are removed. I have updated the screenshots.

PR #17756, opened 2020-04-15
chore(pkger): refactor parser notification endpoint state out into stateful type

jsteenb2:
references: #17434

- [x] Rebased/mergeable
- [ ] Tests pass


kit/transport/http/api.go
```diff
@@ -165,20 +165,35 @@ func (a *API) Respond(w http.ResponseWriter, status int, v interface{}) {
 	}
 
 	w.Header().Set("Content-Type", "application/json; charset=utf-8")
-	enc := json.NewEncoder(writer)
+
+	// this marshal block is to catch failures before they hit the http writer.
```
jsteenb2, line 169: found this issue while testing. I had set an 0 value influxdb.ID type and it exploded in my face during encoding. However, my tests never caught it b/c a 200 was written during encoding. This fixes that up and also makes it viable for anyone else that may have hit this issue.

pkger/models.go
```diff
@@ -877,309 +856,6 @@ type SummaryVariable struct {
...
-			Field: fieldMin,
-			Msg:   "min must be < max",
-		})
-	}
-	return vErrs
-}
-
-func toInfluxThresholds(thresholds ...threshold) []icheck.ThresholdConfig {
-	var iThresh []icheck.ThresholdConfig
-	for _, th := range thresholds {
```
jsteenb2, line 929: vast majority of LOCs coming from moving these bits around to their new homes

kit/transport/http/api.go
bthesorceror: did you mean 'body' instead of 'bod' here?

jsteenb2: good catch 👍, I did mean body

PR #17725, opened 2020-04-13
fix: legacy fromFlux translation

drdelambre:
Closes #16474
Closes #17262
Closes influxdata/idpe#6772

replaces the flux parser that lives in giraffe with one that lives in influxdb. it closes the issue because it more closely follows the spec (throws an error during parsing). This PR is a weird side effect of a migration strategy and is wasteful of resources, but i wanted to merge it in while taking care of the todos, as i have no idea what the ramifications of changing that stuff is and this has a nice little bow on it.

PR #17757, opened 2020-04-15
fix: reapply new flux result parser

drdelambre:
this is a redeploy of https://github.com/influxdata/influxdb/pull/17725 for all the same reasons with one minor change:

```
#group,false,false,false
#datatype,string,long,long
#default,_result,,
,result,table,_value
,,0,28

```
vs
```
#group,false,false,false
#datatype,string,long,long
#default,_result,,
,result,table,_value
,,0,28
```

notice the difference?!

WELL NOW THE TESTS DO!

This was previously reverted because of a newline at the end of a single part query adding some null columns to the end of the table. While it worked fine in displaying the results in timemachine and didn't show up on a majority of tools cluster dashboards, our cells in a dashboard have a different rendering path, and when used in a single-stat or gauge view type, triggered a strange message to hide the results from the user.

This PR now takes care of all the whitespaces, even whitespace injections outside of the spec.

all of the cool stuff (reason i believe this can be redeployed) exists in the commit: https://github.com/influxdata/influxdb/pull/17757/commits/dc9f17e001b44f762f85ca857b87a2504fddf06e


ui/src/shared/utils/fromFlux.test.ts
```diff
@@ -209,4 +210,24 @@ there",5
     expect(() => fromFlux('\n\n\n\nyolo\ntrash\n\n\n\n')).not.toThrow()
     expect(fromFlux('\n\n\n\nyolo\ntrash\n\n\n\n').table.length).toEqual(0)
   })
+
+  test('ignores all the wrong whitespace', () => {
```
drdelambre, line 214: this is the test to address the issue

ui/src/shared/utils/fromFlux.test.ts
```diff
@@ -209,4 +210,24 @@ there",5
     expect(() => fromFlux('\n\n\n\nyolo\ntrash\n\n\n\n')).not.toThrow()
     expect(fromFlux('\n\n\n\nyolo\ntrash\n\n\n\n').table.length).toEqual(0)
```
drdelambre, line 211: you'd think that this test would check for leading and trailing whitespace interactions, but it doesn't

PR #17758, opened 2020-04-15
fix: org filtering for both kv and tenant need to match behaviors.

lyondhill:


organization.go
```diff
@@ -71,8 +71,9 @@ var ErrInvalidOrgFilter = &Error{
 
 // OrganizationFilter represents a set of filter that restrict the returned results.
 type OrganizationFilter struct {
-	Name *string
-	ID   *ID
+	Name   *string
+	ID     *ID
+	UserID *ID
```
AlirieGray, line 76: Should this be added to documentation somewhere if it's the API user who sets it? (Or, if it's only used internally, should it be set internally somewhere?)

lyondhill: I will add that. good find!

PR #17763, opened 2020-04-16
fix(ui): unset active plugin to allow user to leave info page

### LINKED ISSUE #16329: UI: telegraf creating screen gets stuck on plugins page when accessed
[Screen Recording 2019-12-26 at 10.10.37 AM.zip](https://github.com/influxdata/influxdb/files/4003190/Screen.Recording.2019-12-26.at.10.10.37.AM.zip)

__Steps to reproduce:__

1. access the create telegraf UI page
2. select system
3. select any plugin on the left
4. try navigating anywhere with buttons, should be stuck there

__Expected behavior:__
Expect to be able to navigate back to the previous view. 

__Actual behavior:__
Stuck on the plugins page.

__Environment info:__

firefox: 71.0

TCL735:
Closes #16329 

The Plugin Config form must always unset the active plugin to allow the user to escape from the form via the `Done` button.

Consume the new icons from clockface.


TCL735:
There is more to this.

The status icons are also missing.

TCL735:
I have updated the icons so that they will visually appear for the user and not just in the DOM.

ui/src/dataLoaders/components/side_bar/SideBarTab.tsx
```diff
@@ -20,8 +20,9 @@ class SideBarTab extends Component<Props> {
 
     return (
       <div className={this.className} onClick={this.handleClick}>
-        {this.icon}
-        {label}
+        <pre>
```
hoorayimhelping, line 23: what problem is this `pre` solving?

TCL735: Without the `pre` there is no space between the icon and the text. We could add the space with styling. But i think it's simpler just to use a `pre` and have the space _be_ text.

PR #17764, opened 2020-04-16
 feat(pkg/csv2lp): add csv to line protocol conversion library

sranka:
#17004 requests CSV support in `influx write` command, #17599 requires a separate library for CSV to line protocol conversion that shall be then used in `influx write`. This PR introduces the library.

[README.md](https://github.com/influxdata/influxdb/blob/17004/csv2lp/pkg/csv2lp/README.md) describes this library and contains examples.

- [x] Rebased/mergeable
- [x] Tests pass


sranka:
@jsteenb2 can you please help with the stakeholders to review this PR

jsteenb2:
@sranka I'd like to see storage folks or someone who is more LP saavy give this a read. I pulled someone in from the auto assigner for the storage team.

sebito91:
Hey folks, I'm looking into this today!

pkg/csv2lp/csvTable.go
```diff
@@ -0,0 +1,464 @@
...
+		return buffer, errors.New("no measurement column found")
+	}
+	measurement := t.cachedMeasurement.Value(row)
+	if measurement == "" {
+		return buffer, CsvColumnError{
+			t.cachedMeasurement.Label,
+			errors.New("no measurement supplied"),
+		}
+	}
+	buffer = append(buffer, escapeMeasurement(measurement)...)
```
sebito91, line 340: These `append` functions can end up being costly. In this case, I would recommend you leave the function to receive a `buffer []byte` but then copy that to a local `bytes.Buffer` and use the `Write` function to add the characters you need.

https://golang.org/pkg/bytes/#Buffer

pkg/csv2lp/csvTable.go
```diff
@@ -0,0 +1,464 @@
...
+			buffer, err = appendConverted(buffer, timeVal, t.cachedTime)
+			if err != nil {
+				return buffer, CsvColumnError{
+					t.cachedTime.Label,
+					err,
+				}
+			}
+		}
+	}
+	return buffer, nil
```
sebito91, line 412: If you leverage the `bytes.Buffer` you can then return your local copy `Bytes()`

pkg/csv2lp/csvTable.go
```diff
@@ -0,0 +1,464 @@
...
+}
+
+func (e CsvColumnError) Error() string {
+	return fmt.Sprintf("column '%s': %v", e.Column, e.Err)
+}
+
+// CsvTable contains metadata about columns and a state of the CSV processing
+type CsvTable struct {
+	// table columns that extract value from data row
+	columns []CsvTableColumn
```
sebito91, line 88: This should become a slice of pointers to the `CsvTableColumn` as you return pointers all over the place.

pkg/csv2lp/csvToProtocolLines.go
```diff
@@ -0,0 +1,132 @@
...
+	return state
+}
+
+// SkipRowOnError controls whether to fail on every CSV conversion error (false) or to log the error and continue (true)
+func (state *CsvToLineReader) SkipRowOnError(val bool) *CsvToLineReader {
+	state.skipRowOnError = val
+	return state
+}
+
+func (state *CsvToLineReader) Read(p []byte) (n int, err error) {
```
sebito91, line 57: Comment here please...

pkg/csv2lp/dataConversion.go
```diff
@@ -0,0 +1,300 @@
...
+const (
+	stringDatatype                = "string"
+	doubleDatatype                = "double"
+	boolDatatype                  = "boolean"
+	longDatatype                  = "long"
+	uLongDatatype                 = "unsignedLong"
+	durationDatatype              = "duration"
+	base64BinaryDataType          = "base64Binary"
+	dateTimeDatatype              = "dateTime"
+	dateTimeDataFormatRFC3339     = "RFC3339"
```
sebito91, line 26: Can you use just the `time` const-elements like `time.RFC3339`?

pkg/csv2lp/dataConversion.go
```diff
@@ -0,0 +1,300 @@
...
+	supportedDataTypes = make(map[string]struct{}, 9)
+	supportedDataTypes[stringDatatype] = struct{}{}
+	supportedDataTypes[doubleDatatype] = struct{}{}
+	supportedDataTypes[boolDatatype] = struct{}{}
+	supportedDataTypes[longDatatype] = struct{}{}
+	supportedDataTypes[uLongDatatype] = struct{}{}
+	supportedDataTypes[durationDatatype] = struct{}{}
+	supportedDataTypes[base64BinaryDataType] = struct{}{}
+	supportedDataTypes[dateTimeDatatype] = struct{}{}
+	supportedDataTypes[""] = struct{}{}
```
sebito91, line 43: What does the empty-type represent?

pkg/csv2lp/dataConversion.go
```diff
@@ -0,0 +1,300 @@
...
+		return val, nil
+	case dateTimeDatatype:
+		switch dataFormat {
+		case "": // number or time.RFC3339
+			t, err := strconv.ParseInt(val, 10, 64)
+			if err != nil {
+				return time.Parse(time.RFC3339, val)
+			}
+			return time.Unix(0, t).UTC(), nil
+		case dateTimeDataFormatRFC3339:
```
sebito91, line 139: I would just use the `time` const here instead of writing your own version.

pkg/csv2lp/csvTable.go
sranka: I also use `appendConverted` fn to append field values or various data types, the `appendConverted` internally uses `strConv.append*` functions that accept and return []byte. A change to `bytes.Buffer` would then require (on the other hand) to create temporary objects for every float, int, uint value that I don't need now. I am not sure that the situation gets better after this change. So it is trade-off situation herein. I would rather keep it the way it is.

I wish to have `append` optimized by the compiler or `bytes.Buffer` also accepting primitive types for writing.

pkg/csv2lp/csvTable.go
sranka: yes, or I could simplify the function signature to use `*bytes.Buffer`

sebito91:
Excellent work @sranka, will review this and get back to you today!

pkg/csv2lp/csv_annotations.go
```diff
@@ -0,0 +1,171 @@
...
+type annotationComment struct {
+	// prefix in a CSV row that recognizes this annotation
+	prefix string
+	// flag is 0 to represent an annotation that is used for all data rows
+	// or a unique bit (>0) between supported annotation prefixes
+	flag uint8
+	// setupColumn setups metadata that drives the way of how column data
+	// are parsed, mandatory when flag > 0
+	setupColumn func(column *CsvTableColumn, columnValue string)
+	// setupColumn setups metadata that drives the way of how the table data
```
sebito91, line 21: Should be `// setupTable ...`

pkg/csv2lp/multi_closer.go
```diff
@@ -0,0 +1,33 @@
+package csv2lp
```
sebito91, line 1: Is this file actually used anywhere? If not we should remove...

pkg/csv2lp/README.md
```diff
@@ -0,0 +1,179 @@
+# CSV to Line Protocol
+csv2lp library converts CSV (comma separated values) to InfluxDB Line Protocol.
+
+  1. it can process CSV result of a (simple) flux query that exports data from a bucket
+  2. it allows the processing of existing CSV files
+
+## Usage
+The entry point is the ``CsvToProtocolLines`` function that accepts a (utf8) reader with CSV data and returns a reader with line protocol data.
```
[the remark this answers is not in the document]
sranka: ok

pkg/csv2lp/multi_closer.go
sranka: It is not used internally by the library. It is referenced in README.MD (Support Existing CSV files) since it helps compose a single input reader out of to compose multiple Readers and ReadClosers.

[csv2lp.MultiCloser](./multi_closer.go) helps with closing multiple io.Closers (files) on input, because [it is not available OOTB](https://github.com/golang/go/issues/20136)

sebito91:
I think this build requires a `go mod tidy` before it'll pass integration tests. If you can make those changes I'll ack the PR and we're good to merge!

sranka:
I would appreciate any insight or explanation regarding the failing tests that are (from time to time) causing unintended interferences and testing failures. @sebito91 I rebased this branch to the latest master, it usually helps.

sebito91:
That'll usually do it, but if you're importing new packages you'll need to run `go mod vendor` and `go mod tidy` to ensure the modules files are updated accordingly.

For example after bringing your branch down into my local repo, it looks like go.mod is missing a reference to `testify`:

```
(base)  [sborza@icebox]:~/src/github.com/influxdata/influxdb (master-1.x *%):$ git diff go.mod
diff --git a/go.mod b/go.mod
index cd2e3407f4..c702f309d2 100644
--- a/go.mod
+++ b/go.mod
@@ -40,6 +40,7 @@ require (
        github.com/segmentio/kafka-go v0.2.0 // indirect
        github.com/smartystreets/goconvey v1.6.4 // indirect
        github.com/spf13/cast v1.3.0
+       github.com/stretchr/testify v1.4.0
        github.com/tinylib/msgp v1.0.2
        github.com/willf/bitset v1.1.3 // indirect
        github.com/xlab/treeprint v0.0.0-20180616005107-d6fb6747feb6
```

PR #17769, opened 2020-04-16
fix: Migrated bucket should have correct retention policy.

### LINKED ISSUE #17257: Ensure migrated buckets have correct retention policy
Currently the `influxd migrate` tool assumes all retention policies are forever, which is not the case in reality.

We should ensure the retention policy duration gets migrated too. This can be done by reading the `meta.db` file in the OSS 1.x `.influxdb/meta` directory, and checking what the retention policy values should be. Then, when creating the bucket in OSS 2.x using that value for the retention interval.

One thought is if the target bucket already exists but has a different retention period? Perhaps emit a warning or ask the operator what they want to do...

patriczek:
This commit fixed that, if is created bucket while migration and the old database had set retention policy, than this policy is used in created bucket.

Closes #17257

Signed-off-by: Patrik Helia <patashelia@gmail.com>


tsdb/migrate/migrate.go
```diff
@@ -283,6 +299,40 @@ func (m *Migrator) createBucket(db, rp string) (influxdb.ID, error) {
...
+			return nil, err
+		}
+		return nil, err
+	}
+	defer f.Close()
+
+	data, err := ioutil.ReadAll(f)
+	if err != nil {
+		return nil, err
+	}
```
benbjohnson, line 318: This can all be shortened to:

```go
data, err := ioutil.ReadFile(file)
if err != nil {
	return nil, err
}
```

Also, your check for `os.IsNotExist(err)` on L308 is redundant since both error cases simply return `err`.

benbjohnson:
@patriczek Thanks for the update to your commit! lgtm

patriczek:
@benbjohnson You are welcome

PR #17772, opened 2020-04-16
feat(SigninPage): Updated the Signin checkSession to correspond to API changes

Asalem1:
### Problem

Previously we were using the current window location href as a means to redirect the user after performing a checkSession on their Auth0 session. However, if the user was making the request from an un-whitelisted origin, the request would fail, thereby preventing the single sign-on flow.

### Solution

Reverted the Signin page redirect logic so that the redirects are happening from the API rather than the UI. This simplifies our authentication flow by reducing the complexity of session validation while maintaining our single sign-on flow

### DIFF
--- a/ui/src/Signin.tsx
+++ b/ui/src/Signin.tsx
@@ -2,7 +2,6 @@
 import React, {ReactElement, PureComponent} from 'react'
 import {withRouter, WithRouterProps} from 'react-router'
 import {connect} from 'react-redux'
-import auth0js from 'auth0-js'
 
 import {client} from 'src/utils/api'
 
@@ -19,9 +18,6 @@ import {CLOUD, CLOUD_SIGNIN_PATHNAME} from 'src/shared/constants'
 
 // Types
 import {RemoteDataState} from 'src/types'
-
-// Utils
-import {getAuth0Config} from 'src/authorizations/apis'
 
 interface State {
   loading: RemoteDataState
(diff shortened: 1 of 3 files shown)


ui/src/onboarding/containers/LoginPageContents.tsx
```diff
@@ -431,8 +433,13 @@ class LoginPageContents extends PureComponent<DispatchProps> {
       auth0Err.code === 'access_denied' ||
       auth0Err.code === 'user_exists'
     ) {
-      const emailError = `An account with that email address already exists. Try logging in instead.`
-      this.setState({...errors, emailError})
+      if (activeTab === ActiveTab.Login) {
+        const emailError = `The email and password combination you submitted are don't match. Please try again`
```
TCL735, line 437: Is there an extraneous word here? 😄

Asalem1: good eyes

ui/src/Signin.tsx
```diff
@@ -85,45 +85,30 @@ export class Signin extends PureComponent<Props, State> {
...
+          // https://auth0.github.io/auth0.js/web-auth_index.js.html#line564
+          const auth0 = new auth0js.WebAuth({
+            domain: config.domain,
+            clientID: config.clientID,
+            state: config.state,
+            redirectUri: config.redirectURL,
+            responseType: 'token',
+          })
+
+          auth0.checkSession({}, error => {
```
Asalem1, line 101: @ischolten should we return here? It seems like what'll end up happening is that we'll continue onward after this operation finishes, and we'll simply call `handleSignin` on the API? I'm also not sure how relevant these changes are now that we're going to be redirecting from IDPE rather than the UI

PR #17786, opened 2020-04-17
feat: ui measurement completion

bthesorceror:

### DIFF
--- a/ui/src/external/monaco.flux.server.ts
+++ b/ui/src/external/monaco.flux.server.ts
@@ -14,11 +14,14 @@ import {
   references,
   definition,
   symbols,
+  formatting,
 } from 'src/external/monaco.flux.messages'
 import {registerCompletion} from 'src/external/monaco.flux.lsp'
 import {AppState, LocalStorage} from 'src/types'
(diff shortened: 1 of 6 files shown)


ui/src/external/monaco.flux.server.ts
```diff
@@ -33,38 +36,76 @@ import {
...
+    } catch (e) {
+      return []
+    }
+  }
+
   updateBuckets(buckets: string[]) {
     this.buckets = buckets
   }
 
+  setOrg(orgID: string) {
```
drdelambre, line 105: shouldn't this send a message to the LSP in case we need to update the buckets or measurements lists? or am i imagining that the LSP has more state than it needs, and if so is a follow up task to handle caching of the measurements query on the monitoring team side?

bthesorceror: The LSP has no state about the selected bucket or measurements and instead does smart detection for when it needs a updated list. 

- if you use a bucket parameter in a function call
- if you use a measurement parameter in a function call
- if you have a `==` with `r._measurement` on the left side

no state required on the LSP side (other than the cache of what the script looks like)

PR #17800, opened 2020-04-20
feat: dbrp service

gianarb:
This PR implements the HTTP endpoints required to expose the database retention mapping feature to the outside

### DIFF
--- a/authz.go
+++ b/authz.go
@@ -129,6 +129,8 @@ const (
 	NotificationEndpointResourceType = ResourceType("notificationEndpoints") // 15
 	// ChecksResourceType gives permission to one or more Checks.
 	ChecksResourceType = ResourceType("checks") // 16
+	// DBRPType gives permission to one or more DBRPs.
+	DBRPResourceType = ResourceType("dbrp") // 17
 )
 
 // AllResourceTypes is the list of all known resource types.
@@ -150,6 +152,7 @@ var AllResourceTypes = []ResourceType{
 	NotificationRuleResourceType,     // 14
 	NotificationEndpointResourceType, // 15
 	ChecksResourceType,               // 16
+	DBRPResourceType,                 // 17
 	// NOTE: when modifying this list, please update the swagger for components.schemas.Permission resource enum.
 }
 
@@ -167,6 +170,7 @@ var OrgResourceTypes = []ResourceType{
 	NotificationRuleResourceType,     // 14
 	NotificationEndpointResourceType, // 15
 	ChecksResourceType,               // 16
+	DBRPResourceType,                 // 17
 }
 
 // Valid checks if the resource type is a member of the ResourceType enum.
@@ -194,6 +198,7 @@ func (t ResourceType) Valid() (err error) {
 	case NotificationRuleResourceType: // 14
 	case NotificationEndpointResourceType: // 15
 	case ChecksResourceType: // 16
+	case DBRPResourceType: // 17
 	default:
 		err = ErrInvalidResourceType
 	}
--- a/cmd/influxd/launcher/launcher.go
+++ b/cmd/influxd/launcher/launcher.go
@@ -22,6 +22,7 @@ import (
 	"github.com/influxdata/influxdb/v2/bolt"
 	"github.com/influxdata/influxdb/v2/chronograf/server"
 	"github.com/influxdata/influxdb/v2/cmd/influxd/inspect"
+	"github.com/influxdata/influxdb/v2/dbrp"
 	"github.com/influxdata/influxdb/v2/endpoints"
 	"github.com/influxdata/influxdb/v2/gather"
 	"github.com/influxdata/influxdb/v2/http"
@@ -762,6 +763,13 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 		}
 	}
 
+	dbrpSvc, err := dbrp.NewService(ctx, authorizer.NewBucketService(bucketSvc, userResourceSvc), m.kvStore)
+	if err != nil {
+		return err
+	}
+
+	dbrpSvc = dbrp.NewAuthorizedService(dbrpSvc)
+
 	var checkSvc platform.CheckService
 	{
 		coordinator := coordinator.NewCoordinator(m.log, m.scheduler, m.executor)
@@ -881,6 +889,7 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 		BucketService:                   storage.NewBucketService(bucketSvc, m.engine),
 		SessionService:                  sessionSvc,
 		UserService:                     userSvc,
+		DBRPService:                     dbrpSvc,
 		OrganizationService:             orgSvc,
 		UserResourceMappingService:      userResourceSvc,
 		LabelService:                    labelSvc,
--- a/dbrp/http_client_dbrp.go
+++ b/dbrp/http_client_dbrp.go
@@ -1,0 +1,130 @@
+package dbrp
+
+import (
+	"context"
+	"fmt"
+	"path"
+
+	"github.com/influxdata/influxdb/v2"
+	"github.com/influxdata/influxdb/v2/kit/tracing"
+	"github.com/influxdata/influxdb/v2/pkg/httpc"
+)
+
+var _ influxdb.DBRPMappingServiceV2 = (*Client)(nil)
+
+// Client connects to Influx via HTTP using tokens to manage DBRPs.
+type Client struct {
+	Client *httpc.Client
+	Prefix string
+}
+
+func NewClient(client *httpc.Client) *Client {
+	return &Client{
+		Client: client,
+		Prefix: PrefixDBRP,
+	}
+}
+
+func (c *Client) dbrpURL(id influxdb.ID) string {
+	return path.Join(c.Prefix, id.String())
+}
+
+func (c *Client) FindByID(ctx context.Context, orgID, id influxdb.ID) (*influxdb.DBRPMappingV2, error) {
+	span, _ := tracing.StartSpanFromContext(ctx)
+	defer span.Finish()
+
+	var resp getDBRPResponse
+	if err := c.Client.
+		Get(c.dbrpURL(id)).
+		QueryParams([2]string{"orgID", orgID.String()}).
+		DecodeJSON(&resp).
+		Do(ctx); err != nil {
+		return nil, err
+	}
+	return resp.Content, nil
+}
+
+func (c *Client) FindMany(ctx context.Context, filter influxdb.DBRPMappingFilterV2, opts ...influxdb.FindOptions) ([]*influxdb.DBRPMappingV2, int, error) {
+	span, _ := tracing.StartSpanFromContext(ctx)
+	defer span.Finish()
+
+	params := influxdb.FindOptionParams(opts...)
+	if filter.OrgID != nil {
+		params = append(params, [2]string{"orgID", filter.OrgID.String()})
+	} else {
+		return nil, 0, fmt.Errorf("please filter by orgID")
+	}
+	if filter.ID != nil {
+		params = append(params, [2]string{"id", filter.ID.String()})
+	}
+	if filter.BucketID != nil {
+		params = append(params, [2]string{"bucketID", filter.BucketID.String()})
+	}
+	if filter.Database != nil {
+		params = append(params, [2]string{"db", *filter.Database})
+	}
+	if filter.RetentionPolicy != nil {
+		params = append(params, [2]string{"rp", *filter.RetentionPolicy})
+	}
+
+	var resp getDBRPsResponse
+	if err := c.Client.
+		Get(c.Prefix).
+		QueryParams(params...).
+		DecodeJSON(&resp).
+		Do(ctx); err != nil {
+		return nil, 0, err
+	}
+	return resp.Content, len(resp.Content), nil
+}
+
+func (c *Client) Create(ctx context.Context, dbrp *influxdb.DBRPMappingV2) error {
+	span, _ := tracing.StartSpanFromContext(ctx)
+	defer span.Finish()
+
+	var newDBRP influxdb.DBRPMappingV2
+	if err := c.Client.
+		PostJSON(createDBRPRequest{
+			Database:        dbrp.Database,
+			RetentionPolicy: dbrp.RetentionPolicy,
+			Default:         dbrp.Default,
+			OrganizationID:  dbrp.OrganizationID,
+			BucketID:        dbrp.BucketID,
+		}, c.Prefix).
+		DecodeJSON(&newDBRP).
+		Do(ctx); err != nil {
+		return err
+	}
+	dbrp.ID = newDBRP.ID
+	return nil
+}
+
+func (c *Client) Update(ctx context.Context, dbrp *influxdb.DBRPMappingV2) error {
+	span, _ := tracing.StartSpanFromContext(ctx)
+	defer span.Finish()
+
+	if err := dbrp.Validate(); err != nil {
+		return err
+	}
+
+	var newDBRP influxdb.DBRPMappingV2
+	if err := c.Client.
+		PatchJSON(dbrp, c.dbrpURL(dbrp.ID)).
+		QueryParams([2]string{"orgID", dbrp.OrganizationID.String()}).
+		DecodeJSON(&newDBRP).
+		Do(ctx); err != nil {
+		return err
+	}
+	*dbrp = newDBRP
+	return nil
+}
+
+func (c *Client) Delete(ctx context.Context, orgID, id influxdb.ID) error {
+	span, _ := tracing.StartSpanFromContext(ctx)
+	defer span.Finish()
+
+	return c.Client.
+		Delete(c.dbrpURL(id)).
+		QueryParams([2]string{"orgID", orgID.String()}).
+		Do(ctx)
+}
--- a/dbrp/http_server_dbrp.go
+++ b/dbrp/http_server_dbrp.go
@@ -1,0 +1,309 @@
+package dbrp
+
+import (
+	"encoding/json"
+	"net/http"
+	"strconv"
+
+	"github.com/go-chi/chi"
+	"github.com/go-chi/chi/middleware"
+	"github.com/influxdata/influxdb/v2"
+	kithttp "github.com/influxdata/influxdb/v2/kit/transport/http"
+	"go.uber.org/zap"
+)
+
+const (
+	PrefixDBRP = "/api/v2/dbrps"
+)
+
+type Handler struct {
+	chi.Router
+	api     *kithttp.API
+	log     *zap.Logger
+	dbrpSvc influxdb.DBRPMappingServiceV2
+}
+
+// NewHTTPHandler constructs a new http server.
+func NewHTTPHandler(log *zap.Logger, dbrpSvc influxdb.DBRPMappingServiceV2) *Handler {
+	h := &Handler{
+		api:     kithttp.NewAPI(kithttp.WithLog(log)),
+		log:     log,
+		dbrpSvc: dbrpSvc,
+	}
+
+	r := chi.NewRouter()
+	r.Use(
+		middleware.Recoverer,
+		middleware.RequestID,
+		middleware.RealIP,
+	)
+
+	r.Route("/", func(r chi.Router) {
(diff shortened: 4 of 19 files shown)


http/swagger.yml
```diff
@@ -339,6 +339,148 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
+  "/dbrp":
+    get:
+      operationId: GetDBRPList
+      tags:
+        - DBRPEndpoints
+      summary: List all database retention policy mapping
```
brettbuddin, line 347: Nit: `mappings`

http/swagger.yml
```diff
@@ -11047,6 +11189,55 @@ components:
     NotificationEndpointType:
       type: string
       enum: ['slack', 'pagerduty', 'http']
+    DBRP:
+      properties:
+        orgID:
```
stuartcarnie, line 11194: We will need an `id` property added to the Swagger. This will necessary so the Get, Update and Delete operations can function correctly with an addressable ID:

```yaml
        id:
          readOnly: true
          type: string
```

That will require we update the `DBRPMapping` struct to include an `ID` property:

https://github.com/influxdata/influxdb/blob/e4539ad04d5fb6f7bc3b0e2516ebc5812036ee49/dbrp_mapping.go#L26-L36

We will also need to amend the `DBRPMappingService` to be able to Get, Update and Delete a mapping by `ID`:

```go
// DBRPMappingService provides a mapping of cluster, database and retention policy to an organization ID and bucket ID.
type DBRPMappingService interface {
	FindMappingByID(ctx context.Context, id ID) (*DBRPMapping, error)
	UpdateMapping(ctx context.Context, id ID, update DBRPMappingUpdate) (*DBRPMapping, error)
	DeleteMapping(ctx context.Context, id ID) error
}
```

authz.go
```diff
@@ -129,6 +129,8 @@ const (
 	NotificationEndpointResourceType = ResourceType("notificationEndpoints") // 15
 	// ChecksResourceType gives permission to one or more Checks.
 	ChecksResourceType = ResourceType("checks") // 16
+	// DBRPType gives permission to one or more DBRPs.
+	DBRPType = ResourceType("dbrp") // 16
```
affo, line 133: @gianarb could this be `DBRPResourceType`?

Just for conformity with the others

cmd/influxd/launcher/launcher.go
```diff
@@ -774,6 +775,13 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 		}
 	}
 
+	dbrpSvc, err := dbrp.NewService(ctx, authorizer.NewBucketService(bucketSvc, userResourceSvc), m.kvStore)
```
affo, line 778: that's great, relying on the bucket auth, IMO,  is the way to go 🎉

dbrp/http_server_dbrp.go
```diff
@@ -0,0 +1,238 @@
...
+	r.Route("/", func(r chi.Router) {
+		r.Post("/", h.handlePostDBRP)
+		r.Get("/", h.handleGetDBRPs)
+
+		r.Route("/{id}", func(r chi.Router) {
+			r.Get("/", h.handleGetDBRP)
+			r.Patch("/", h.handlePatchDBRP)
+			r.Delete("/", h.handleDeleteDBRP)
+		})
+	})
```
affo, line 40: routing is so clear and clean ❤️

dbrp/http_dbrp_server_test.go
```diff
@@ -0,0 +1,115 @@
...
+		t.Fatal(err)
+	}
+
+	server := httptest.NewServer(dbrp.NewHTTPDBRPHandler(zaptest.NewLogger(t), svc))
+	return server, func() {
+		server.Close()
+	}
+}
+
+func Test_handlePostDBRP(t *testing.T) {
```
affo, line 37: @gianarb best would be to test every endpoint 👍

cmd/influxd/launcher/launcher.go
```diff
@@ -774,6 +775,13 @@ func (m *Launcher) run(ctx context.Context) (err error) {
...
+	if err != nil {
+		return err
+	}
+
+	dbrpSvc = dbrp.DBRPMappingAuthorzedService{DBRPMappingServiceV2: dbrpSvc}
```
affo, line 783: oh ok, you also have a proper auth layer for dbrp

dbrp/service_auth.go
```diff
@@ -0,0 +1,51 @@
...
+func (svc DBRPMappingAuthorzedService) FindMany(ctx context.Context, filter influxdb.DBRPMappingFilter, opts ...influxdb.FindOptions) ([]*influxdb.DBRPMapping, int, error) {
+	if _, _, err := authorizer.AuthorizeOrgReadResource(ctx, influxdb.DBRPType, *filter.OrgID); err != nil {
+		return nil, 0, ErrUnauthorized(err)
+	}
+
+	return svc.DBRPMappingServiceV2.FindMany(ctx, filter, opts...)
+}
+
+func (svc DBRPMappingAuthorzedService) Create(ctx context.Context, t *influxdb.DBRPMapping) error {
+	if _, _, err := authorizer.AuthorizeOrgWriteResource(ctx, influxdb.DBRPType, t.OrganizationID); err != nil {
```
affo, line 33: it's actually the same, but it would be better to use `AuthorizeCreate`

inmem/dbrp_mapping_test.go
```diff
@@ -18,26 +18,31 @@ func initDBRPMappingService(f platformtesting.DBRPMappingFields, t *testing.T) (
 }
 
 func TestDBRPMappingService_CreateDBRPMapping(t *testing.T) {
+	t.Skip("not needed anymore because we are moving to a new implementation of dbrp")
```
affo, line 21: @gianarb it would be best to have this set of tests with the new dbrp mapping service 👍

affo:
@gianarb I was just thinking about auth.

When you authorize somebody to create a DBRP, he/should be able to read and write the underlying bucket, or just read?

If "just read" then this implementation is correct.  
Otherwise, the auth layer should ensure that 🤔

gianarb:
> If "just read" then this implementation is correct.

This is how I implement it, feel free to change it if you think differently. I didn't get any opinion on it. I thought it was enough

inmem/dbrp_mapping_test.go
gianarb: Tests like this Test_handlePostDBRP in `dbrp` convers the service as well. I do not see a need for that

dbrp/service_auth.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"context"
+
+	"github.com/influxdata/influxdb/v2"
+	"github.com/influxdata/influxdb/v2/authorizer"
+)
+
+var _ influxdb.DBRPMappingServiceV2 = (*DBRPMappingAuthorzedService)(nil)
+
+type DBRPMappingAuthorzedService struct {
```
GeorgeMac, line 12: Missing an `i` should be `DBRPMappingAuthorizedService`

affo: boom George the eagle

affo:
@gianarb Yeah, that was just to raise awareness.

If you are still editing the PR and can add a comment such as

```
NOTE: this allows to create a DBRP even if the user has only the right to read the bucket.
The user must NOT be able to write the bucket.
If we want to change this behaviour in the future, we should check for that permission as well.
```

So that we have a memo for the future 👍

dbrp/service.go
```diff
@@ -0,0 +1,198 @@
...
+	}); err != nil {
+		return nil, err
+	}
+
+	dbrp := &influxdb.DBRPMappingV2{}
+	if err := json.Unmarshal(b, dbrp); err != nil {
+		return nil, ErrInternalServiceError(err)
+	}
+	// If the given orgID is wrong, it is as if we did not found a DBRP scoped to the org.
+	if dbrp.OrganizationID != orgID {
```
affo, line 66: this is what you wanted to see @GeorgeMac

http/api_handler.go
```diff
@@ -212,6 +214,8 @@ func NewAPIHandler(b *APIBackend, opts ...APIHandlerOptFn) *APIHandler {
 	backupBackend.BackupService = authorizer.NewBackupService(backupBackend.BackupService)
 	h.Mount(prefixBackup, NewBackupHandler(backupBackend))
 
+	h.Mount("/api/v2/dbrp", dbrp.NewHTTPHandler(b.Logger, b.DBRPService))
```
affo, line 217: memo for me, should use `prefixDBRP` here

dbrp_mapping.go
```diff
@@ -37,12 +177,6 @@ type DBRPMapping struct {
 
 // Validate reports any validation errors for the mapping.
 func (m DBRPMapping) Validate() error {
-	if !validName(m.Cluster) {
```
affo, line 40: for me: should restore this

http/swagger.yml
```diff
@@ -345,16 +345,231 @@ paths:
...
+      summary: List all database retention policy mappings
       parameters:
-          - $ref: '#/components/parameters/TraceSpan'
-          - in: query
-            name: type
-            description: The type of plugin desired.
+        - $ref: "#/components/parameters/TraceSpan"
+        - in: query
+          name: orgID
+          required: true
```
affo, line 358: added required here

http/swagger.yml
```diff
@@ -345,16 +345,231 @@ paths:
...
+            name: default
+            description: Specifies filtering on default
+            schema:
+              type: boolean
+        - in: query
+            name: db
+            description: Specifies the database to filter on
+            schema:
+              type: string
+        - in: query
```
affo, line 382: added filtering by any field of the dbrp mapping

http/swagger.yml
```diff
@@ -345,16 +345,231 @@ paths:
...
             schema:
               type: string
+      responses:
+        "200":
+          description: A list of all database retention policy mappings
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/DBRPs"
+        "400":
```
affo, line 394: added errors for every endpoint

affo:
@stuartcarnie @GeorgeMac  I think this is now good to go 👍 

Just added some last tests for the server and fixed some nits (also rebased on master).

I also added some bits to the swagger.

Using my IDE I caused a lot of formatting changes, is there a standard way to format yaml? 🤔

affo:
@hoorayimhelping @desa FYI

@aanthony1243 told me you were interested in seeing a service built from scratch with the new pattern.

This is an example. The DBRP mapping service associates 1.x db-rp pairs to 2.x buckets.

dbrp/service.go
```diff
@@ -0,0 +1,198 @@
...
+		bucket, err := tx.Bucket(bucket)
+		if err != nil {
+			return ErrInternalService(err)
+		}
+		cur, err := bucket.Cursor()
+		if err != nil {
+			return ErrInternalService(err)
+		}
+
+		for k, v := cur.First(); k != nil; k, v = cur.Next() {
```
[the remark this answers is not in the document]
affo: Yep, haven't actually touched this implementation 🤔

dbrp/http_client_dbrp.go
```diff
@@ -0,0 +1,130 @@
...
+func (c *Client) FindMany(ctx context.Context, filter influxdb.DBRPMappingFilterV2, opts ...influxdb.FindOptions) ([]*influxdb.DBRPMappingV2, int, error) {
+	span, _ := tracing.StartSpanFromContext(ctx)
+	defer span.Finish()
+
+	params := influxdb.FindOptionParams(opts...)
+	if filter.OrgID != nil {
+		params = append(params, [2]string{"orgID", filter.OrgID.String()})
+	} else {
+		return nil, 0, fmt.Errorf("please filter by orgID")
+	}
```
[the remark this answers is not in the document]
affo: Yeah, I am doing that.

This makes it impossible to get every dbrp in the system.

Do we want to scope finds to a single org?

That was the assumption I was making.

Do we still want to provide the possibility to get the entire DBRP set?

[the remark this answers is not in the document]
stuartcarnie: Yes, that seems reasonable to be able to find them all in the API. You could have two APIs:

```go
func (c *Client) FindManyByOrgID(...
func (c *Client) FindMany(...
```

dbrp/service.go
```diff
@@ -0,0 +1,489 @@
+package dbrp
+
+// The DBRP Mapping `Service` maps database, retention policy pairs to buckets.
```
affo, line 3: @aanthony1243 here are the comments on doing relational stuff with a kv backend 👍

dbrp/service.go
```diff
@@ -0,0 +1,489 @@
...
+	return ms, len(ms), err
+}
+
+// Create creates a new mapping.
+// If another mapping with same organization ID, database, and retention policy exists, an error is returned.
+// If the mapping already contains a valid ID, that one is used for storing the mapping.
+func (s *Service) Create(ctx context.Context, dbrp *influxdb.DBRPMappingV2) error {
+	if !dbrp.ID.Valid() {
+		dbrp.ID = s.IDGen.ID()
+	}
```
stuartcarnie, line 349: It is unclear why ID can be specified or valid for a create operation. I would expect the ID must always be generated by `Service#Create`. I do note that `Create` checks whether the ID exists, and fails the operation, but the behavior is not what I would expect.

affo: This is mostly for testing, to mock IDs.

Is something that we agreed with @lyondhill and was already done for the tenant service.

This does not impact the HTTP side however. The ID never gets deserialized from the request, and is always generated.

dbrp/service.go
```diff
@@ -0,0 +1,489 @@
...
+// updateDefault updates the default value of the given mapping based on the default value stored.
+func (s *Service) updateDefault(ctx context.Context, m *influxdb.DBRPMappingV2) error {
+	if d, err := s.isDefault(ctx, m); err != nil {
+		return err
+	} else if d {
+		m.Default = true
+	} else {
+		m.Default = false
+	}
+	return nil
```
GeorgeMac, line 147: take it or leave it:
```go
func (s *Service) updateDefault(ctx context.Context, m *influxdb.DBRPMappingV2) (err error) {
    m.Default, err = s.isDefault(ctx, m)
    return
}
```

dbrp/service.go
```diff
@@ -0,0 +1,489 @@
...
+			set = true
+		} else {
+			return defValue, ErrInternalService(err)
+		}
+	}
+	if set {
+		if err := b.Put(compKey, id); err != nil {
+			return defValue, ErrInternalService(err)
+		}
+	}
```
GeorgeMac, line 208: I tried to walk through this behaviour and see if I could make it more readable. I came to this which I think is equivalent, but I could be mistaken:
```go
	_, err = b.Get(compKey)
	// given a nil error (dbrp is set as default) and we do not want to set the dbrp as the default
	if err == nil && !defValue  {
		// then just return default value as provided
		return defValue, nil
	// given a nil error (dbrp is set as default) and we want to set the dbrp as the default
	// or the key was not present (therefore the dbrp is not currently the default)
	} else if (err == nil && defValue) || kv.IsNotFound(err) {
		if err = b.Put(compKey, id); err == nil {
			return true, nil
		}
	}
	return defValue, ErrInternalService(err)
```

Which I think is equivalent and when I read it back it makes me wonder about the behaviour of the method.

There are couple cases that make me double take:
1. if `defValue == false (we want to unset) && the DBRP is currently set as default` it will noop and return `false`.
2. given `the DBRP is currently set as default && defValue == true` it will set it again.

Taking this method out of the context where it is called, these behaviours confused me. Perhaps they make sens in the called context. So I will dig further into where they're called and perhaps this will clear up my understanding :D

dbrp_mapping.go
```diff
@@ -7,6 +7,146 @@ import (
...
+	Database        *string
+	RetentionPolicy *string
+	Default         *bool
+}
+
+func (f DBRPMappingFilterV2) String() string {
+	var s strings.Builder
+	s.WriteString("{")
+
+	s.WriteString(" id:")
```
GeorgeMac, line 105:
```go
s.WriteString("{ id:")
```

kv/index.go
```diff
@@ -277,6 +277,36 @@ func (i *Index) Walk(ctx context.Context, tx Tx, foreignKey []byte, visitFn Visi
 	return indexWalk(ctx, cursor, sourceBucket, visitFn)
 }
 
+// First returns the first entry in the index.
+func (i *Index) First(tx Tx, foreignKey []byte) ([]byte, error) {
+	// return not found if the index is being used purely to write.
+	if !i.canRead {
+		return nil, ErrKeyNotFound
```
GeorgeMac, line 284: ❤️

dbrp/service.go
```diff
@@ -0,0 +1,490 @@
...
+			if err := json.Unmarshal(v, &dbrp); err != nil {
+				return nil, err
+			}
+			return indexForeignKey(dbrp), nil
+		}), kv.WithIndexReadPathEnabled),
+	}, nil
+}
+
+// getDefault gets the default mapping ID inside of a transaction.
+func (s *Service) getDefault(tx kv.Tx, compKey []byte) ([]byte, error) {
```
affo, line 96: @GeorgeMac from here on, simpler methods 🎉

testing/dbrp_mapping_v2.go
```diff
@@ -0,0 +1,1670 @@
...
+				t.Fatalf("failed to retrieve dbrps: %v", err)
+			}
+			if diff := cmp.Diff(dbrpMappings, tt.wants.dbrpMappings, DBRPMappingCmpOptionsV2...); diff != "" {
+				t.Errorf("dbrpMappings are different -got/+want\ndiff %s", diff)
+			}
+		})
+	}
+}
+
+func MiscDBRPMappingV2(
```
affo, line 1580: @GeorgeMac the misc test for verifying interconnected create/delete

affo:
@GeorgeMac fixed things we talked about this morning 👍

dbrp/service.go
affo: I simplified the methods in the last commit. Everything should be clearer.

kv/index.go
affo: removed `First` because we do not need that anymore

dbrp/service.go
```diff
@@ -0,0 +1,490 @@
...
+			}
+
+			for k, v := cur.First(); k != nil; k, v = cur.Next() {
+				if err := add(tx)(k, v); err != nil {
+					return err
+				}
+			}
+			return nil
+		})
+	}
```
GeorgeMac, line 314: Take it or leave it:
This could be one `View` with the conditional inside.

dbrp/service.go
```diff
@@ -0,0 +1,490 @@
...
+			// This means default was unset.
+			// Need to find a new default.
+			first, ferr := s.getFirstBut(tx, compKey, encodedID)
+			if ferr != nil {
+				return ferr
+			}
+			if len(first) > 0 {
+				err = s.setAsDefault(tx, compKey, first)
+			}
+			// If no first was found, then this will remain the default.
```
GeorgeMac, line 432: Do we need to error here?
Since the caller is expecting to unset this as the default, they should probably be presented with something like:
`&influxdb.Error{Code: influxdb.Invalid, Msg: "cannot unset DBRP as default: cannot find alternative default"}`

dbrp/service.go
```diff
@@ -0,0 +1,490 @@
...
+			first, err := s.getFirstBut(tx, compKey, encodedID)
+			if err != nil {
+				return err
+			}
+			if len(first) > 0 {
+				derr = s.setAsDefault(tx, compKey, first)
+			} else {
+				// This means no other mapping is in the index.
+				// Unset the default
+				derr = s.unsetDefault(tx, compKey)
```
GeorgeMac, line 474: Same goes here as for Update. Should we fail Delete if there is no suitable alternative `default`?

dbrp/service.go
affo: This is a design decision. Think this could be updated in a further PR if needed.

dbrp/service.go
affo: I think this is ok as it is.

Because otherwise one could never delete mappings for a database

PR #17801, opened 2020-04-20
feat(query): add interface for the window aggregate capability

jsternberg:
Added an interface for an additional storage capability. This interface
will allow for checking if the reader supports the window aggregate call
and another method for invoking the call if it does.

This is implemented using a single interface. If the reader implements
the interface, it indicates that the client is capable of reading the
response. The `HasXXX` method is intended to check if the store supports
the operation. This method also takes a context because it could require
a remote call or to wait for one.


jsternberg:
I ended up going with this interface instead of the others for these reasons.

1. It is typesafe. You can typecast to learn if the client supports reading and then call the method to check for the server.
2. The two interface method was _more_ typesafe, but it also incurred programmer overhead because you had to learn to deal with and implement two interfaces. If you wanted to implement a capability, you had to implement two interfaces instead of just one.
3. The other method meant you could use a different struct to implement the capability, but I figured this would be something we would, in practicality, not use.
4. I preferred this over using `HasCapability(string)` because this concealed how the capabilities were implemented and made it more typesafe (see 1).

nathanielc:
>The two interface method was more typesafe, but it also incurred programmer overhead because you had to learn to deal with and implement two interfaces. If you wanted to implement a capability, you had to implement two interfaces instead of just one.

Agreed adding one interface per RPC method is better than two.

> The other method meant you could use a different struct to implement the capability, but I figured this would be something we would, in practicality, not use.

I agree, we would not use that ability in practice.

>I preferred this over using HasCapability(string) because this concealed how the capabilities were implemented and made it more typesafe (see 1).

It feels a bit odd to me that you have to first type cast the reader to see if it implements `WindowAggregateCapability` and then you have to explicitly check on that interface if it supports the method.

Would an enum of Capablities be a good middle ground?

The storage reader interface would have a `HasCapability(c Capability) bool` method on it, and then we would define the set of Capabilities like:

```go
type Capability int

const (
    WindowAggregateCapability = iota
    //...
)
```

This way the check for the Capability comes before any typecasting.

jsternberg:
The reason I wanted to avoid the `HasCapability(c Capability) bool` method is because I felt it added an extra disconnect between checking the capability and implementing it. It would be possible for me to implement `HasCapability` for a specific capability without also implementing the interface type for reading. In the method in this PR, since both are part of the same interface, it would be impossible for me to make that mistake. I could implement `HasWindowAggregateCapability() bool`, but if I don't remember to also implement the reader, then it doesn't implement the interface.

In all likelihood, I imagine that each of the implementations will do something like:

```go
func (s *Store) HasWindowAggregateCapability() bool {
    return s.hasCapability(windowAggregateCapability)
}
```

Maybe a better name is more appropriate. I'm going to rename `Capability` to `Reader`.

nathanielc:
@jsternberg Ok that works for me. Let's get one more review.

jsternberg:
There may be cases where that happens since the planner rule is global code and the reader is injected as a dependency. So the runtime has the planner rule but the reader injected as a dependency doesn't know how to read the message.

If we added the method to the reader interface, we break the existing interface and have to update every implementation.

jsternberg:
As a note, I do think that, at the moment, the `influxdb.Reader` only has one implementor. If this is the case, we can potentially refactor this later to remove the `influxdb.Reader` portion of the interface in favor of a single struct implementation. Then we wouldn't have a breaking change when adding methods and wouldn't have to chain the interface conversions.

PR #17804, opened 2020-04-20
 feat(pkger): add stateful management for dashboards

jsteenb2:
references: #17434 

note: the first commit here updates the dashboard update to allow for updating cells (as a collection) as part of the dashboard update.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -2282,7 +2282,17 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/Dashboard"
+                type: object
+                properties:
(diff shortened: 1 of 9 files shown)


http/swagger.yml
```diff
@@ -2282,7 +2282,17 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/Dashboard"
```
jsteenb2, line 2285: this old ref is actually inaccurate for this endpoint. The new stuff below represents what the API supports

PR #17806, opened 2020-04-20
feat: allow for new flag source

drdelambre:
some ground work for: https://github.com/influxdata/idpe/issues/6654
related to: #17851

the intended behavior is that a user gets their feature flags assigned to them on the server, whos state is sent to the client on a `/api/v2/flags` request. If a developer wants to override the feature flag on the client, or add a feature flag that is only used on the client, they can do so using an interface much like the current one.

the api work isn't merged yet, so this just moves the feature into redux, and allows the existing functionality to operate until we start fleshing out the contract


ui/src/shared/actions/me.ts
```diff
@@ -4,25 +4,17 @@ import {CLOUD} from 'src/shared/constants'
...
-export interface SetMe {
-  type: ActionTypes.SetMe
-  payload: {
-    me: MeState
-  }
-}
+export const SET_ME = 'SET_ME'
 
-export type Actions = SetMe
+export type Actions = ReturnType<typeof setMe>
```
drdelambre, line 9: migrated to new reducer coding standard while i was in here

ui/src/shared/selectors/flags.ts
```diff
@@ -0,0 +1,39 @@
...
+import {FlagMap} from 'src/shared/reducers/flags'
+import {CLOUD, CLOUD_BILLING_VISIBLE} from 'src/shared/constants'
+
+export const OSS_FLAGS = {
+  deleteWithPredicate: false,
+  downloadCellCSV: false,
+  telegrafEditor: false,
+  customCheckQuery: false,
+  matchingNotificationRules: false,
+  regionBasedLoginPage: false,
```
Asalem1, line 11:
```suggestion

```

ui/src/shared/selectors/flags.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const CLOUD_FLAGS = {
+  deleteWithPredicate: false,
+  multiUser: false,
+  cloudBilling: CLOUD_BILLING_VISIBLE, // should be visible in dev and acceptance, but not in cloud
+  downloadCellCSV: false,
+  telegrafEditor: false,
+  customCheckQuery: false,
+  matchingNotificationRules: false,
+  regionBasedLoginPage: false,
```
Asalem1, line 24:
```suggestion
```

ui/src/shared/actions/flags.ts
```diff
@@ -0,0 +1,31 @@
...
+
+// NOTE: this doesnt have a type as it will be determined
+// by the backend at a later time and keeping the format
+// open for transformations in a bit
+export const setFlags = flags => {
+  return {
+    type: SET_FEATURE_FLAGS,
+    payload: flags,
+  } as const
+}
```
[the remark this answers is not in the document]
Asalem1: Just to keep things consistent within the same file since that's what we're doing in the next function

ui/src/mockState.tsx
```diff
@@ -50,8 +50,10 @@ export const localState: LocalStorage = {
 const history = createMemoryHistory({entries: ['/']})
 
 export function renderWithRedux(ui, initialState = s => s) {
+  clearStore()
   const seedStore = configureStore(localState, history)
   const seedState = seedStore.getState()
+  clearStore()
```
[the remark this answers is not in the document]
drdelambre: to share the store around the app without recreating stores, i had to make it a singleton, which makes sense for a single page app. maybe a slight interface change to allow for multiple stores to be created will be required in the future, but the desired outcome is to provide the developer the ability to say "give me the store" without explicitly plumbing it everywhere. There be some abstraction dragons here, but i like to think we're using the store as a singleton anyways, just allowing `connect` to be that interface that obfuscates the plumbing for us, so it's not THAT bad. the clearStore works like how we use the `/debug/flush` endpoint for e2e tests, so that we can inject fresh data into the singleton on every test. This specific stuff looks weird because we're using a store to load a store, but overwrite specific things, instead of injecting a default store with overrides, which i can imagine is to allow for greater flexibility in _how_ that default store gets defined / created and more easily mirror actual application state than keeping a derived state in sync with an explicit one.

ui/src/shared/selectors/flags.ts
drdelambre: are you flagging a merge conflict here?

Asalem1: Nah, it's no longer being used

PR #17814, opened 2020-04-21
feat(prometheus): update prometheus remote protocol

foobar:
This PR updates prometheus remote api to latest version which supports streamed read. It imports prometheus package directly instead of making a copy.
This change will lay the groundwork for #18528.


foobar:
@dgnorton could you take a look?

ctmuthu:
@foobar is this the reason why currently remote read from influxdb is not working?

foobar:
> @foobar is this the reason why currently remote read from influxdb is not working?

Sorry I may not get you correctly but currently it is working well. This PR just updates the remote read protocol-bufffer message to alight with prometheus code.

foobar:
@timhallinflux could you please get this reviewed?

jacobmarble:
Perhaps indicate which change [in this changelog](https://github.com/prometheus/client_golang/blob/master/CHANGELOG.md) you're interested in?

foobar:
> @foobar when I change only `github.com/prometheus/client_golang v1.5.1`, the diff is much simpler than what I see in this PR (go.mod gets one other dependency updated). Can you share a link to the dependency change that you would like to import?

The change i made is just adding the following line, but go mod needs to get the dependency updated.

```
	github.com/prometheus/prometheus v0.0.0-20200609090129-a6600f564e3c
```

jacobmarble:
This still leaves me confused. When I add `github.com/prometheus/prometheus v0.0.0-20200609090129-a6600f564e3c` to go.mod, then run `go mod tidy`, the result is close to what you propose. However, the dependency isn't used, so shouldn't be added.

```
$ go mod why github.com/prometheus/prometheus
go: finding github.com/glycerine/goconvey/convey latest
go: finding github.com/glycerine/goconvey latest
# github.com/prometheus/prometheus
(main module does not need package github.com/prometheus/prometheus)
```

Have you tested this PR? If it adds the streaming functionality you're looking for, then it is a side effect of what else is added because of prometheus/prometheus. Or I'm wrong, and maybe you can help me understand what's going on here.

Also, can you provide a quick way for me to test whether the streaming feature is working? I would like to pull the branch down and test myself before we get to merging this change.

foobar:
thanks @jacobmarble for looking into this. 

> This still leaves me confused. When I add `github.com/prometheus/prometheus v0.0.0-20200609090129-a6600f564e3c` to go.mod, then run `go mod tidy`, the result is close to what you propose. However, the dependency isn't used, so shouldn't be added.

what being used is a package from github.com/prometheus/prometheus (github.com/prometheus/prometheus/prompb itself is not a module)
 
```
go mod why github.com/prometheus/prometheus/prompb
# github.com/prometheus/prometheus/prompb
github.com/influxdata/influxdb/prometheus
github.com/prometheus/prometheus/prompb
```


> Have you tested this PR? If it adds the streaming functionality you're looking for, then it is a side effect of what else is added because of prometheus/prometheus. Or I'm wrong, and maybe you can help me understand what's going on here.

It does introduce dependencies change because prometheus also depends on them and uses some newer versions of those dependencies.  I think this is required by go module(https://github.com/golang/go/issues/31578)


> Also, can you provide a quick way for me to test whether the streaming feature is working? I would like to pull the branch down and test myself before we get to merging this change.

This PR doesn't make the streaming/chunked read work by itself. It just pulls in the latest remote read protobuffer which supports streaming read:

https://github.com/prometheus/prometheus/blob/f97d2ddb6e6fce1343d2b026479d19592b934a36/prompb/remote.pb.go#L46-L49

With this pr, we can implement chunked response, but with a new PR.

jacobmarble:
@foobar FYI #18757 should separate dependency upgrade anxiety from this PR.

foobar:
> @foobar FYI #18757 should separate dependency upgrade anxiety from this PR.

I created pr https://github.com/influxdata/influxdb/pull/18766 to update the dependencies to align with prometheus

jacobmarble:
@foobar I'm taking a final look at this today, plan to merge tomorrow.

jacobmarble:
I have built this PR locally, and checked the results from the /metrics endpoint. Things look very much the same there. 👍

PR #17822, opened 2020-04-21
fix: filling views and cells when we grab dashboards

### LINKED ISSUE #17787: Adding a new Query to a new dashboard cell fails
__Steps to reproduce:__
1. go to explore
1. reload the page **this is critical**
1. create a new query
1. save the query to a dashboard and create a new cell
1. failure, undefined variables

drdelambre:
Closes #17787

when fetching a list of dashboards, we weren't filling the cells / views that also come back from the result of this api call.

### DIFF
--- a/ui/src/dashboards/actions/thunks.ts
+++ b/ui/src/dashboards/actions/thunks.ts
@@ -234,6 +234,10 @@ export const getDashboards = () => async (
 
     dispatch(setDashboards(RemoteDataState.Done, dashboards))
 
+    if (!dashboards.result.length) {
+      return
+    }
+
     Object.values(dashboards.entities.dashboards)
       .map(d => {
         return {
--- a/ui/src/dashboards/actions/thunks.ts
+++ b/ui/src/dashboards/actions/thunks.ts
@@ -239,14 +239,14 @@ export const getDashboards = () => async (
     }
 
     Object.values(dashboards.entities.dashboards)
-      .map(d => {
+      .map(dashboard => {
         return {
-          id: d.id,
-          cells: d.cells.map(c => dashboards.entities.cells[c]),
+          id: dashboard.id,
+          cells: dashboard.cells.map(cell => dashboards.entities.cells[cell]),
         }
       })
-      .forEach(e => {
-        const viewsData = viewsFromCells(e.cells, e.id)
+      .forEach(entity => {
+        const viewsData = viewsFromCells(entity.cells, entity.id)
 
         const normViews = normalize<View, ViewEntities, string[]>(
           viewsData,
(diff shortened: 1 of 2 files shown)


ui/src/dashboards/utils/cellGetters.ts
```diff
@@ -50,9 +50,9 @@ export const getNewDashboardCell = (
     status: RemoteDataState.Done,
   }
 
-  const cells = dashboard.cells.map(
-    cellID => state.resources.cells.byID[cellID]
-  )
+  const cells = dashboard.cells
+    .map(cellID => state.resources.cells.byID[cellID])
+    .filter(c => !!c)
```
drdelambre, line 55: this is just to guard from the issue above causing the page to error

ui/src/dashboards/actions/thunks.ts
```diff
@@ -229,6 +233,34 @@ export const getDashboards = () => async (
     )
 
     dispatch(setDashboards(RemoteDataState.Done, dashboards))
+
+    if (!dashboards.result.length) {
+      return
+    }
+
+    Object.values(dashboards.entities.dashboards)
```
drdelambre, line 241: this is the source of the issue; not backfilling views and cells

ui/src/dashboards/actions/thunks.ts
```diff
@@ -229,6 +233,34 @@ export const getDashboards = () => async (
     )
 
     dispatch(setDashboards(RemoteDataState.Done, dashboards))
+
+    if (!dashboards.result.length) {
```
hoorayimhelping, line 237: is `result` guaranteed to be exist and to be a array-like? we often have undefined errors because of assumptions that variables exist, just want to confirm this is kosher.

ui/src/dashboards/actions/thunks.ts
```diff
@@ -229,6 +233,34 @@ export const getDashboards = () => async (
...
+        )
+
+        const normCells = normalize<Dashboard, DashboardEntities, string[]>(
+          e.cells,
+          arrayOfCells
+        )
+
+        dispatch(setViews(RemoteDataState.Done, normViews))
+        dispatch(setCells(e.id, RemoteDataState.Done, normCells))
+      })
```
hoorayimhelping, line 263: i really dislike single letter variable names - they make it very hard to find patterns and to track down loose ends in global searches and require a little extra brainpower to process while reading. writing out a variable name costs us very little.

```suggestion
      .map(dashboard => {
        return {
          id: dashboard.id,
          cells: dashboard.cells.map(cell => dashboards.entities.cells[cell]),
        }
      })
      .forEach(entity => {
        const viewsData = viewsFromCells(entity.cells, entity.id)

        const normViews = normalize<View, ViewEntities, string[]>(
          viewsData,
          arrayOfViews
        )

        const normCells = normalize<Dashboard, DashboardEntities, string[]>(
          entity.cells,
          arrayOfCells
        )

        dispatch(setViews(RemoteDataState.Done, normViews))
        dispatch(setCells(entity.id, RemoteDataState.Done, normCells))
      })
```

ui/src/dashboards/utils/cellGetters.ts
hoorayimhelping, line 55:
```suggestion
    .filter(cell => !!cell)
```

ui/src/dashboards/actions/thunks.ts
drdelambre: totally is by the normalize function above

PR #17830, opened 2020-04-22
revert(Logout): simplified Logout component functionality to maintain the original cloud logout

Asalem1:
fix(package.json): Updated the cloud-dev command to have the CLOUD_URL land on the same default port as Quartz UI

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -19,7 +19,7 @@
     "start": "yarn install && yarn generate && cross-env TS_NODE_PROJECT=\"webpack.tsconfig.json\" && yarn run build:vendor && yarn run start:dev",
     "start:cloud": "yarn install && yarn generate && cross-env TS_NODE_PROJECT=\"webpack.tsconfig.json\" && yarn run build:vendor && yarn run start:dev-cloud",
     "start:dev": "webpack-dev-server --config ./webpack.dev.ts --progress false",
-    "start:dev-cloud": "cross-env CLOUD_LOGOUT_URL=http://localhost:8080/api/v2/signout CLOUD_URL=http://localhost:9999 webpack-dev-server --config ./webpack.dev.ts",
+    "start:dev-cloud": "cross-env CLOUD_LOGOUT_URL=http://localhost:8080/api/v2/signout CLOUD_URL=http://localhost:4000 webpack-dev-server --config ./webpack.dev.ts",
     "start:docker": "yarn generate && yarn build:vendor && yarn run start:dev",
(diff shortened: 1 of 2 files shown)


ebb-tide:
I don't get it but the code looks good!

Asalem1:
> I don't get it but the code looks good!

I'll add some comments specifically with the code to clear up any confusion

ui/package.json
```diff
@@ -19,7 +19,7 @@
     "start": "yarn install && yarn generate && cross-env TS_NODE_PROJECT=\"webpack.tsconfig.json\" && yarn run build:vendor && yarn run start:dev",
     "start:cloud": "yarn install && yarn generate && cross-env TS_NODE_PROJECT=\"webpack.tsconfig.json\" && yarn run build:vendor && yarn run start:dev-cloud",
     "start:dev": "webpack-dev-server --config ./webpack.dev.ts --progress false",
-    "start:dev-cloud": "cross-env CLOUD_LOGOUT_URL=http://localhost:8080/api/v2/signout CLOUD_URL=http://localhost:9999 webpack-dev-server --config ./webpack.dev.ts",
+    "start:dev-cloud": "cross-env CLOUD_LOGOUT_URL=http://localhost:8080/api/v2/signout CLOUD_URL=http://localhost:4000 webpack-dev-server --config ./webpack.dev.ts",
```
Asalem1, line 22: made a change to the `CLOUD_URL` so that the port is set to the same default port as Quartz. This change was made so that we could connect the two UIs locally. While this is by no means a long-term stable solution to developing both UIs locally, it does allow for us to currently play around in the local environment

ui/src/Logout.tsx
```diff
@@ -15,16 +13,6 @@ import {ErrorHandling} from 'src/shared/decorators/errors'
...
-    const config = await getAuth0Config()
-    if (CLOUD && config.socialSignUpOn) {
-      const auth0 = new auth0js.WebAuth({
-        domain: config.domain,
-        clientID: config.clientID,
-      })
-      auth0.logout({})
-      window.location.href = `${CLOUD_URL}${CLOUD_LOGOUT_PATH}`
-      return
-    }
```
Asalem1, line 27: TLDR; we refactored cloud authentication to move away from the UI and to the API. Since the existing CLOUD API implementation for logging out already destroys Auth0 sessions & CLOUD API sessions, we can confidently depend upon the existing code and no longer need a special feature-flag enabled logout setting

PR #17831, opened 2020-04-22
feat: add a workflow to enforce non-empty PR or Issue descriptions

GeorgeMac:
This adds a Github action which ensure that both pull requests and issue bodies have descriptions with at least 1 character in them.

The action is a little militant as it just closes the PR. But you can update the description and re-open it. I could go make a pull request or fork it to just fail the check status via some configuration option.


jeffreyssmith2nd:
Closing in favor of the PR template that was added recently.

PR #17836, opened 2020-04-22
refactor(query): return detailed capabilities for the read window aggregate interfaces

jsternberg:
This modifies the read window aggregate interfaces to future-proof it
if and when we add additional capabilities to the method. Previously,
the interface was all or nothing. If we modified the RPC call itself, we
would have to make a new interface to denote the change to the Go code.

This changes the interface so now a `WindowAggregateCapability` exists.
This way, we can modify the struct to include things like:

```go
    WindowPeriodCapability bool
    MeanAggregateCapability bool
}
```

This way we can learn if the RPC call itself supports some specific
option. If the first iteration doesn't support a mean aggregate or the
mean aggregate is only supported by single server implementations, the
window aggregate can tell the caller that it won't be able to compute
the mean aggregate.

Since it fills in a struct with these capabilities, the struct can
safely introduce new values. If a downstream consumer wants to take
advantage of that functionality, then all interfaces in the chain have
to be updated to consume the upstream capabilities.

### DIFF
--- a/query/stdlib/influxdata/influxdb/storage.go
+++ b/query/stdlib/influxdata/influxdb/storage.go
@@ -149,10 +149,15 @@ type ReadWindowAggregateSpec struct {
 	// TODO(issue #17784): add attributes for the window aggregate spec.
 }
 
+// WindowAggregateCapability describes what is supported by WindowAggregateReader.
+type WindowAggregateCapability struct{}
+
 // WindowAggregateReader implements the WindowAggregate capability.
 type WindowAggregateReader interface {
 	// HasWindowAggregateCapability will test if this Reader source supports the ReadWindowAggregate capability.
-	HasWindowAggregateCapability(ctx context.Context) bool
+	// If WindowAggregateCapability is passed to the method, then the struct
+	// is filled with a detailed list of what the RPC call supports.
+	HasWindowAggregateCapability(ctx context.Context, capability ...*WindowAggregateCapability) bool
 
 	// ReadWindowAggregate will read a table using the WindowAggregate method.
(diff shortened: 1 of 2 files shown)


query/stdlib/influxdata/influxdb/storage.go
```diff
@@ -149,10 +149,15 @@ type ReadWindowAggregateSpec struct {
...
+// WindowAggregateCapability describes what is supported by WindowAggregateReader.
+type WindowAggregateCapability struct{}
+
 // WindowAggregateReader implements the WindowAggregate capability.
 type WindowAggregateReader interface {
 	// HasWindowAggregateCapability will test if this Reader source supports the ReadWindowAggregate capability.
-	HasWindowAggregateCapability(ctx context.Context) bool
+	// If WindowAggregateCapability is passed to the method, then the struct
+	// is filled with a detailed list of what the RPC call supports.
+	HasWindowAggregateCapability(ctx context.Context, capability ...*WindowAggregateCapability) bool
```
yzhang1991, line 160: Why `WindowAggregateCapability` is not directly returned along with that boolean. or I am thinking maybe the function can just return a `WindowAggregateCapability` type. a non-nil value means it is supported?

jsternberg: There's no reason why it doesn't. We could potentially just do that. The main reason is I wanted to be able to do something like this:

```
var capability WindowAggregateCapability
if !HasWindowAggregateCapability(ctx, &capability) {
    return
}
```

If you return more than one parameter, you have to assign it to a variable to use the boolean result.

yzhang1991: Ah I see.

PR #17843, opened 2020-04-23
feat(pkger): add stateful management for notification rules

jsteenb2:
references: #17434

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/pkger/service.go
+++ b/pkger/service.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/influxdata/influxdb/v2"
 	ierrors "github.com/influxdata/influxdb/v2/kit/errors"
+	"github.com/influxdata/influxdb/v2/notification/rule"
 	"github.com/influxdata/influxdb/v2/snowflake"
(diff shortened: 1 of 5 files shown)


pkger/service.go
```diff
@@ -1859,44 +1895,111 @@ func (s *Service) applyNotificationRules(rules []*stateRule) applier {
...
+		return influxRule.NotificationRule, nil
 	}
-
-	return influxRule, nil
 }
 
-func (s *Service) rollbackNotificationRules(rules []*stateRule) error {
+func (s *Service) rollbackNotificationRules(ctx context.Context, userID influxdb.ID, rules []*stateRule) error {
+	rollbackFn := func(r *stateRule) error {
+		existingRuleFn := func(endpointID influxdb.ID) influxdb.NotificationRule {
```
jsteenb2, line 1945: this is rather unfortunate. We have a getter for the rule's endpointID but no setter... this works around that limiation

PR #17850, opened 2020-04-23
refactor(storage/reads): refactor the capabilities interface

jsternberg:
The capabilities interface will now return a mapping of capabilities to
a capabilities object. The capabilities object will contain a list of
features supported by the capability.


nathanielc:
What is the advantage of the `map[string][]string` over a simpler `[]string` data structure? I'd prefer capabilities to be super simple so that we can easily reason about them and talk about them without needing much context.

jsternberg:
My thought was that we could categorize the features so we didn't have to prefix features for a specific capability. As an example:

```go
const (
	WindowAggregateCapability = "WindowAggregate"
	MultipleAggregatesFeature = "MultipleAggregatesFeature"
)

var Capabilities = map[string]*datatypes.Capability{
	WindowAggregateCapability: {
		Features: []string{
			MultipleAggregatesFeature,
		},
	},
}
```

I don't feel too strongly about this though. I also figured it was more extensible for the future. If we needed a new feature, then `Capability` being a message means it is extensible. `[]string` isn't.

nathanielc:
I like the idea of a message. And roughly speaking the map key is the RPC method name and the feature list are the feature that method supports? 

How do we represent that we support the `WindowAggregateCapability` on its own without any extra features? Is that just the existence of the key with an empty list?

jsternberg:
Yes. Existence of the key would indicate the base feature set is supported.

nathanielc:
>Yes. Existence of the key would indicate the base feature set is supported.

Sounds good, can you add some comments to this effect? I.e. the key should normally map to an RPC method name and the sub list are features of that particular method.

jsternberg:
Updated the protobuf file with a comment.

PR #17860, opened 2020-04-24
feat(ui): ability to create a bucket from time machine

alexpaxton:
This change makes creating a bucket much more natural by enabling it without leaving the time machine

create-bucket-on-the-fly

- Refactor `CreateBucketOverlay` to get everything it needs from redux and dispatch
- Add `create-bucket` to overlay controller
- Create a component within `BucketSelector` that shows the create bucket form within a popover
  - This is disabled if a cloud user has maxed out buckets
- Refactor bucket selector in time machine to get list directly from redux instead of `getActiveTimeMachine`

- [x] Rebased/mergeable
- [x] Tests pass


drdelambre:
this is 💯 amazing and blows my mind and delivers all of my hopes and dreams, but write a cypress test for it. "i went to data explorer, clicked on a button, filled out a name and clicked on some more buttons, and then it was there forever"

ui/src/timeMachine/components/SelectorListCreateBucket.tsx
```diff
@@ -0,0 +1,203 @@
...
+  isRetentionLimitEnforced,
+  limitStatus,
+  checkBucketLimits,
+}) => {
+  const triggerRef = useRef<HTMLButtonElement>(null)
+  const [bucketName, setBucketName] = useState<string>('')
+  const [bucketRetentionRules, setBucketRetentionRules] = useState<any>(
+    isRetentionLimitEnforced ? DEFAULT_RULES : []
+  )
+  const [bucketRuleType, setBucketRuleType] = useState<'expire' | null>(
```
121watts, line 78: There's enough state in here to justify moving this into a `useReducer` hook.   It could all even stay in this file but would pull a bunch of this state management into a reducer.

alexpaxton:
@drdelambre I added the e2e test

ui/src/timeMachine/components/SelectorListCreateBucket.tsx
alexpaxton: A similar thought I had was to make a component that is just the contents of the popover, then most of the state can live in there. We'd still have a lot of state though. The component I based this off of just had a whole `Bucket` object as a single piece of state so I flattened it out

alexpaxton:
@121watts alright I rebuilt the component with `useReducer` instead of `useState` check it out

PR #17861, opened 2020-04-24
feat(pkger): add support for stateful label mappings

jsteenb2:
references: #17434

this PR has some small fixes for the http pkg. Label http client pointing to wrong routes, and telegraf http client didn't support update. Those are remedied in this, first 2 commits of PR.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/pkger/service_models.go
+++ b/pkger/service_models.go
@@ -19,7 +19,8 @@ type stateCoordinator struct {
 	mTelegrafs  map[string]*stateTelegraf
 	mVariables  map[string]*stateVariable
 
-	labelMappings []stateLabelMapping
+	labelMappings         []stateLabelMapping
+	labelMappingsToRemove []stateLabelMappingForRemoval
 }
 
 func newStateCoordinator(pkg *Pkg) *stateCoordinator {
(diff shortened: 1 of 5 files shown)


http/label_service.go
```diff
@@ -620,9 +620,8 @@ func (s *LabelService) CreateLabelMapping(ctx context.Context, m *influxdb.Label
 		return err
 	}
 
-	urlPath := resourceIDPath(m.ResourceType, m.ResourceID, "labels")
 	return s.Client.
-		PostJSON(m, urlPath).
+		PostJSON(m, string(m.ResourceType), m.ResourceID.String(), "labels").
```
jsteenb2, line 624: these were hitting wrong routes. The tests in pkger failed when trying to validate label mappings with the http client as the label service dependency. These are fixed and tests are now passing. There is a big gap in tests for this inside `http` pkg itself though. I did not go back and add that at this point.

pkger/service_models.go
```diff
@@ -19,7 +19,8 @@ type stateCoordinator struct {
 	mTelegrafs  map[string]*stateTelegraf
 	mVariables  map[string]*stateVariable
 
-	labelMappings []stateLabelMapping
+	labelMappings         []stateLabelMapping
+	labelMappingsToRemove []stateLabelMappingForRemoval
```
jsteenb2, line 23: separating these two, trying to juggle them with the same type felt icky. Side effect of this decision, is we can now simplify the service implementation dramatically.

PR #17865, opened 2020-04-24
fix(kv): context done

docmerlin:
inmem kv shouldn't keep holding a lock on transactions or run them after their context has expired/been_cancled/etc.

This was adapted from Rob Pike's blogpost https://blogtitle.github.io/go-advanced-concurrency-patterns-part-3-channels/

### DIFF
--- a/inmem/kv.go
+++ b/inmem/kv.go
@@ -23,7 +23,7 @@ const cursorBatchSize = 1000
 
 // KVStore is an in memory btree backed kv.Store.
 type KVStore struct {
-	mu      sync.RWMutex
+	mu      RWMutex
 	buckets map[string]*Bucket
 	ro      map[string]*bucket
 }
@@ -31,6 +31,7 @@ type KVStore struct {
 // NewKVStore creates an instance of a KVStore.
 func NewKVStore() *KVStore {
 	return &KVStore{
+		mu:      NewLock(),
 		buckets: map[string]*Bucket{},
 		ro:      map[string]*bucket{},
 	}
@@ -38,7 +39,9 @@ func NewKVStore() *KVStore {
 
 // View opens up a transaction with a read lock.
 func (s *KVStore) View(ctx context.Context, fn func(kv.Tx) error) error {
-	s.mu.RLock()
+	if !s.mu.RTryLock(ctx) {
+		return ctx.Err()
+	}
 	defer s.mu.RUnlock()
 
 	return fn(&Tx{
@@ -50,7 +53,9 @@ func (s *KVStore) View(ctx context.Context, fn func(kv.Tx) error) error {
 
 // Update opens up a transaction with a write lock.
 func (s *KVStore) Update(ctx context.Context, fn func(kv.Tx) error) error {
-	s.mu.Lock()
+	if !s.mu.TryLock(ctx) {
+		return ctx.Err()
+	}
 	defer s.mu.Unlock()
 
 	return fn(&Tx{
--- a/inmem/kv_test.go
+++ b/inmem/kv_test.go
@@ -7,7 +7,9 @@ import (
 	"os"
 	"reflect"
 	"sort"
+	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/influxdata/influxdb/v2/inmem"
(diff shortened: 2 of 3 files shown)


inmem/kv.go
```diff
@@ -50,9 +60,18 @@ func (s *KVStore) View(ctx context.Context, fn func(kv.Tx) error) error {
...
 // Update opens up a transaction with a write lock.
 func (s *KVStore) Update(ctx context.Context, fn func(kv.Tx) error) error {
-	s.mu.Lock()
+	if !s.mu.TryLock(ctx) {
+		return ctx.Err()
+	}
 	defer s.mu.Unlock()
 
+	// we have to do this because RTryLock doesn't check context before attempting to take the lock
+	select {
```
affo, line 69: @docmerlin why this?

It seems that `TryLock` accounts for `ctx.Done()` in the select statement 🤔

inmem/kv_test.go
```diff
@@ -139,6 +141,103 @@ func TestKVStore_Bucket_CursorHintPredicate(t *testing.T) {
...
+		kv := inmem.NewKVStore()
+
+		ctx, cancel := context.WithCancel(context.Background())
+		go kv.Update(ctx, blockTillDone(&countUpdateStarted))
+		time.Sleep(time.Second / 2)
+		for i := 0; i < 20; i++ {
+			go kv.View(ctx, blockTillDone(&countViewStarted))
+		}
+		time.Sleep(time.Second / 2)
+		cancel()
```
affo, line 202: when you trigger `cancel` should this unlock the views and update their count? 🤔

inmem/kvlock.go
```diff
@@ -0,0 +1,127 @@
...
+	l.readers <- rs
+}
+
+func (l RWMutex) RTryLock(ctx context.Context) bool {
+	// Count current readers. Default to 0.
+	var rs int
+	// Select on the channels without default.
+	// One and only one case will be selected and this
+	// will block until one case becomes available.
+	if ctx == nil {
```
affo, line 76: could you simplify logic and:

```
if ctx == nil {
 ctx = context.Background()
}
```
?

So that you write the logci only once?

inmem/kvlock.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func NewLock() RWMutex {
+	return RWMutex{
+		// This is used as a normal Mutex.
+		write: make(chan struct{}, 1),
+		// This is used to protect the readers count.
+		// By receiving the value it is guaranteed that no
+		// other goroutine is changing it at the same time.
+		readers: make(chan int, 1),
```
GeorgeMac, line 22: I think you're going to want to prime this channel with `readers <- 0` before returning.
Otherwise, the first attempt to read will block on the first retrieve.
This does get unblocked by the first write lock call, but I imagine we wouldn't want to rely on at least one Write before a Read can happen.

GeorgeMac: Oh wait, I see you select on both and increment afterwards. Ignore me.

inmem/kv.go
docmerlin: Woops, that was my fault, thanks affo!

inmem/kvlock.go
docmerlin: I thought about it, but it seemed wasteful to wait on a channel that would never return.

inmem/kvlock.go
```diff
@@ -0,0 +1,150 @@
...
+		l.readers <- rs
+		return true
+	}
+	done := ctx.Done()
+	select {
+	case l.write <- struct{}{}: // One sending case for write.
+		// If the write lock is available we have no readers.
+		// We grab the write lock to prevent concurrent
+		// read-writes.
+		println("grabbed write")
```
affo, line 106: think this should be deleted

inmem/kvlock.go
```diff
@@ -0,0 +1,150 @@
...
+		select {
+		case <-done:
+			println("was done")
```
affo, line 109: this one too

inmem/kvlock.go
```diff
@@ -0,0 +1,150 @@
...
+			return false
+
+		}
+	}
+	done := ctx.Done()
+	select {
+	case l.write <- struct{}{}:
+		select {
+		case <-done:
+			<-l.write
```
affo, line 43: This is not clear to me 🤔 

Why would you grab the lock if the context is done?
Should you release instead

```
l.write <- struct{}{}
```
?

inmem/kv_test.go
affo: @docmerlin still have this doubt 😅

inmem/kvlock.go
docmerlin: That isn't a release.  `<-l.write` is a release
`l.write <- struct{}{}` grabs the write lock.

affo: Sure thing, I got confused 😅

inmem/kv.go
```diff
@@ -23,22 +23,25 @@ const cursorBatchSize = 1000
...
+		mu:      NewLock(),
 		buckets: map[string]*Bucket{},
 		ro:      map[string]*bucket{},
 	}
 }
 
 // View opens up a transaction with a read lock.
 func (s *KVStore) View(ctx context.Context, fn func(kv.Tx) error) error {
-	s.mu.RLock()
+	if !s.mu.RTryLock(ctx) {
```
aanthony1243, line 42: This seems like a straightforward improvement.   Should we consider feature flagging this since it's on a service-critical path?

PR #17874, opened 2020-04-28
refactor: create new authorization package

### LINKED ISSUE #17726: Refactor authorization service
This issue is part of the ongoing work in this epic to refactor the API and enable a new database to replace the kv store. The work in this issue includes:

- [ ] Move auth code involving tokens to a new token package
- [ ] Simplify kv/auth.go to match the pattern established in the tenant service 
- [ ] Use chi router to mount routes

AlirieGray:
Closes #17726

This PR creates a new package `Authorization` for the `Authorization` resource used to store tokens. This package follows the new pattern set by the Tenant service to include the API, service, and storage layers in the same package. The functionality remains mostly unchanged. 

One change to the functionality is that the old kv service implementation did not allow an ID to be provided when creating a new Authorization, and would ignore an ID that was given to it and instead generate a new one. This implementation instead checks that a provided ID is valid and does not already correspond to an Authorization in the system, and will use it if it meets those criteria. A test in `testing/auth.go` was updated to reflect this change. This was done so that we would not have to pass an ID Generator when creating a new Service in order to get the tests to pass, and the Service can create its own ID Generator. 

The comparer options for the Authorization in `testing/Auth.go` was also changed to ignore the ID, Token, and CRUD Log properties since these are now generated internally by the system. (May need additional testing to ensure these are generated and saved properly?)

The code in this PR will be placed behind a feature flag in a subsequent PR. Another PR will be made for the metrics middleware.

As an aside, it consolidates the `http/paging.go` and `influxdb/paging.go` files.

- [x] Rebased/mergeable
- [x] Tests pass


lyondhill:
Looks good so far. I will keep an eye and when the tests are passing and we add metrics I will look again.

PR #17885, opened 2020-04-28
feat(query): Reader implementation for WindowAggregate

yzhang1991:


query/stdlib/influxdata/influxdb/storage.go
```diff
@@ -148,18 +148,15 @@ type TableIterator interface {
...
-// WindowAggregateCapability describes what is supported by WindowAggregateReader.
-type WindowAggregateCapability struct{}
-
 // WindowAggregateReader implements the WindowAggregate capability.
 type WindowAggregateReader interface {
 	// HasWindowAggregateCapability will test if this Reader source supports the ReadWindowAggregate capability.
 	// If WindowAggregateCapability is passed to the method, then the struct
 	// is filled with a detailed list of what the RPC call supports.
-	HasWindowAggregateCapability(ctx context.Context, capability ...*WindowAggregateCapability) bool
+	HasWindowAggregateCapability(ctx context.Context, capability ...*reads.WindowAggregateCapability) bool
```
jsternberg, line 159: This shouldn't be here. We need to duplicate this struct so we don't have an explicit dependency on the reads package. Please keep it as it was.

storage/flux/reader.go
```diff
@@ -79,6 +80,24 @@ func (r *storeReader) ReadGroup(ctx context.Context, spec influxdb.ReadGroupSpec
 	}, nil
 }
 
+func (r *storeReader) HasWindowAggregateCapability(ctx context.Context, capability ...*storage.WindowAggregateCapability) bool {
+	if aggStore, ok := r.s.(storage.WindowAggregateStore); ok {
+		return aggStore.HasWindowAggregateCapability(ctx, capability...)
+	} else {
```
jsternberg, line 86: Don't include a hanging false.

https://golang.org/doc/effective_go.html#if

storage/readservice/store.go
```diff
@@ -153,3 +153,12 @@ func (s *store) GetSource(orgID, bucketID uint64) proto.Message {
...
 	}
 }
+
+func (s *store) HasWindowAggregateCapability(ctx context.Context, capability ...*reads.WindowAggregateCapability) bool {
+	return false
+}
+
+// WindowAggregate will invoke a ReadWindowAggregateRequest against the Store.
+func (s *store) WindowAggregate(ctx context.Context, req *datatypes.ReadWindowAggregateRequest) (reads.ResultSet, error) {
+	return nil, nil
```
jsternberg, line 163: This should return an error.

mock/reader.go
```diff
@@ -5,33 +5,34 @@ import (
 
 	"github.com/influxdata/flux/memory"
 	"github.com/influxdata/influxdb/v2/query/stdlib/influxdata/influxdb"
+	"github.com/influxdata/influxdb/v2/storage/reads"
 )
 
-type StoreReader struct {
+type Reader struct {
```
jsternberg, line 11: Does this name need to be changed? The name `Reader` is a very generic name for a package and since this is a mock package the intention was to say what kind of reader this was. It's a mock of reading the storage engine. If you want, we can name it `StorageReader` to match the changed interface name that is on the algorithm w branch.

yzhang1991: OK `StorageReader` sounds good.

query/stdlib/influxdata/influxdb/storage.go
yzhang1991: Can you explain why we cannot have this dependency?

My understanding is that this `WindowAggregateCapability` struct in the future will contain flags like:
```
    SupportA bool
    SupportB bool
    ....
}
```

If we keep two separate `WindowAggregateCapability` struct definitions, conversion code is needed in this function implementation to convert from `reads.WindowAggregateCapability` to `influxdb.WindowAggregateCapability`.  Dependency on the `reads` package is inevitable, but the implementation of this function is in a different package though.

jsternberg: My thought process on this is that we shouldn't have this dependency as part of the interface because it assumes the reader is the specific reader from `reads.Store` and there's no other part of the interface that is dependent on that package. I can make an `influxdb.Reader` and have it live completely within the flux code. The implementation is in a separate package.

In general, I think it's an anti-pattern if an interface has a dependency on structs from other packages.

We can potentially bring this refactor to a different PR. I'd like to discuss it more, but I don't want to block work that is dependent on this PR.

mock/reader.go
```diff
@@ -40,13 +41,13 @@ func (s *StoreReader) Close() {
 	}
 }
 
-type WindowAggregateStoreReader struct {
-	*StoreReader
-	HasWindowAggregateCapabilityFn func(ctx context.Context) bool
+type WindowAggregateReader struct {
```
jsternberg, line 44: Can this get named the same way as above? `WindowAggregateStorageReader`.

storage/readservice/store.go
```diff
@@ -153,3 +153,12 @@ func (s *store) GetSource(orgID, bucketID uint64) proto.Message {
...
 	}
 }
+
+func (s *store) HasWindowAggregateCapability(ctx context.Context, capability ...*reads.WindowAggregateCapability) bool {
+	return false
+}
+
+// WindowAggregate will invoke a ReadWindowAggregateRequest against the Store.
+func (s *store) WindowAggregate(ctx context.Context, req *datatypes.ReadWindowAggregateRequest) (reads.ResultSet, error) {
+	return nil, errors.New("unimplemented WindowAggregate: https://github.com/influxdata/idpe/issues/6805")
```
jsternberg, line 163: Include this as a comment. I think it's better if we don't include links in errors even if they should never be hit.

yzhang1991:
@adrian-thurston Since you are almost done for the day, I will take a look at your PR https://github.com/influxdata/influxdb/pull/17898 and create a branch off yours with some suggested changes.

PR #17888, opened 2020-04-29
chore(errors): interpret more cases as influxdb error type in http.CheckError

GeorgeMac:
This is just a little proposal PR to restructure some of the `http.CheckError` behaviour to get a little more context in some cases.

We recently experience what appeared to be a response between tasks and gateway with a non 2XX or 3XX with no content type and an empty body. This is turning up in a task run log as an `Unexpected end of JSON input: ""` error.
While I am unsure if we can get much more context, the status code in the situation is getting dropped on the floor. This change ensures that we at least attempt to convert it into one of our influxdb error codes.
It also leverages both `message` and `error` fields in the event of a failure to unmarshal an error as json. The `error` contains the first line of the response body and the `message` contains an explanation and the error received after the attempt to unmarshal.

I also promoted the error code to status code and back as a function in `errors.go`. I am not sure if this is a bad thing. But seems like something we can standardize and share somewhere.

Lastly, it changes the default status code to `500` from `400`, because if we can't figure out what an error is we shouldn't be reporting it as a user error.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/errors.go
+++ b/http/errors.go
@@ -12,7 +12,7 @@ import (
 	"strings"
 
 	platform "github.com/influxdata/influxdb/v2"
-	"github.com/pkg/errors"
+	khttp "github.com/influxdata/influxdb/v2/kit/transport/http"
 )
 
 // AuthzError is returned for authorization errors. When this error type is returned,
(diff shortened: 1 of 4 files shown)


errors.go
```diff
@@ -261,3 +261,49 @@ func decodeInternalError(target interface{}) error {
 type HTTPErrorHandler interface {
 	HandleHTTPError(ctx context.Context, err error, w http.ResponseWriter)
 }
+
+// HTTPStatusCodeToErrorCode maps a http status code integer to an
+// influxdb error code string.
+func HTTPStatusCodeToErrorCode(statusCode int) string {
```
jsteenb2, line 267: did you consider moving this to `kit/http`? imo, this doesn't belong here in the domain. its a http/grpc specific concern. 

the API responder type already does this here: https://github.com/influxdata/influxdb/blob/svc_def/kit/transport/http/api.go#L90

derived from: https://github.com/influxdata/influxdb/blob/svc_def/kit/transport/http/api.go#L232

GeorgeMac: Awesome, cheers @jsteenb2 I will do that. I wasn't sure if it was a little too HTTP specific for the top-level. I will move it.

gianarb: This is related @GeorgeMac https://github.com/influxdata/influxdb/issues/17889

PR #17898, opened 2020-04-29
feat: added PushDownWindowAggregate planner rewrite rule

adrian-thurston:
Added a (disabled and feature-flagged) planner rule that matches:

ReadRange -> window -> { min, max, mean, count, sum }

The rule requires:
 * the pushDownWindowAggregate{Count,Rest} feature flags enabled
 * having WindowAggregateCapability
   (which StorageReader does not currently have)
 * use of "_value" columns only
 * window.period == window.every
 * window.every.months == 0
 * window.every is positive
 * window.offset == 0
 * standard time columns
 * createEmpty is false

### DIFF
--- a/query/stdlib/influxdata/influxdb/rules.go
+++ b/query/stdlib/influxdata/influxdb/rules.go
@@ -11,6 +11,8 @@ import (
 	"github.com/influxdata/flux/semantic"
 	"github.com/influxdata/flux/stdlib/influxdata/influxdb"
 	"github.com/influxdata/flux/stdlib/universe"
+	"github.com/influxdata/influxdb/v2/kit/feature"
+	"github.com/influxdata/influxdb/v2/query"
 )
 
 func init() {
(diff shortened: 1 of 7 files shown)


query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -638,3 +639,81 @@ func (SortedPivotRule) Rewrite(ctx context.Context, pn plan.Node) (plan.Node, bo
...
+	// windowAggregateReade := reader.(reads.WindowAggregateReader)
+	// if !storeReader.HasWindowAggregateCapability(ctx) {
+	//  	return pn, false, nil
+	// }
+
+	// Check the aggregate function spec. Require operation on _value.
+	fnNode := pn
+	switch fnNode.Kind() {
+	case universe.MinKind:
+		minSpec := fnNode.ProcedureSpec().(*universe.MinProcedureSpec)
```
wolffcm, line 675: It occurs to me that the specific functions that are supported in pushdowns will be specific to the storage nodes we're talking to.  Does the capabilities API expose which aggregate functions are supported?

In our initial update to storage, only one aggregate function will be supported.  It would be nice if, once we add support for other aggregate functions, those nodes begin reporting their new capabilities, and we don't have to make any changes to the planner rule to reflect that, it would just start working.

adrian-thurston: Yeah if we assign the capabilities now we can definitely do that. @wolffcm

PR #17900, opened 2020-04-29
chore: promote algorithm W to master

jlapacik:


jlapacik:
@rockstar are you able to address @jsternberg's question?

rockstar:
> I know we already did the work to ensure goreleaser would work, but did we integrate that work into the release process?

I'm not sure, but we definitely can't make macOS binaries right now with the work we did. I guess I'm not sure what the delta is in the release process, as my assumption was that the work we were doing was behind a shared protocol with influxdb.

PR #17901, opened 2020-04-29
refactor: unify WindowAggregateCapability

yzhang1991:
This PR is a follow-up for #17885

We currently have two `WindowAggregateCapability` struct definitions. One is in [query/stdlib/influxdata/influxdb/storage.go](https://github.com/influxdata/influxdb/blob/master/query/stdlib/influxdata/influxdb/storage.go#L154) used by the `WindowAggregateReader` implementations, the other one is in [influxdb/storage/reads/store.go](https://github.com/influxdata/influxdb/blob/master/storage/reads/store.go#L87) used by the `WindowAggregateStore` implementations.

The plan is to add more supported feature flags to those `WindowAggregateCapability` structs in the future as we move forward to add more storage API capabilities. For example:

```go
    SupportMutipleAggregates bool
    SupportSpreadFunction bool
    ...
}
```
Then in OSS we can do something like:
```go
func pushDownSpreadAggregate() bool {
	if aggStore, ok := store.(storage.WindowAggregateStore); ok {
		var capability WindowAggregateCapability
		aggStore.HasWindowAggregateCapability(ctx, &capability)
		return capability.SupportSpreadFunction
	}
	return false
}
```

I want to eliminate the struct definition in [query/stdlib/influxdata/influxdb/storage.go](https://github.com/influxdata/influxdb/blob/master/query/stdlib/influxdata/influxdb/storage.go#L154), only use the one in [influxdb/storage/reads/store.go](https://github.com/influxdata/influxdb/blob/master/storage/reads/store.go#L87) for the following reasons:

1. We don't need to keep both `WindowAggregateCapability` definitions in sync.
2. The `Reader`'s implementation is simpler, otherwise, we need to introduce code to convert from a `Store`'s `WindowAggregateCapability` to a `Reader`'s `WindowAggregateCapability`.

@jsternberg 's opinion is to keep the duplicated structs in their respective package because he does not want the `influxdb` package to have a dependency on the `reads` package. Quote his comment:

>My thought process on this is that we shouldn't have this dependency as part of the interface because it assumes the reader is the specific reader from reads.Store and there's no other part of the interface that is dependent on that package. I can make an influxdb.Reader and have it live completely within the flux code. The implementation is in a separate package.

>In general, I think it's an anti-pattern if an interface has a dependency on structs from other packages.

### DIFF
--- a/storage/reads/store.go
+++ b/storage/reads/store.go
@@ -84,14 +84,13 @@ type Store interface {
 }
 
 // WindowAggregateCapability describes what is supported by WindowAggregateStore.
-type WindowAggregateCapability struct{}
+type WindowAggregateCapability interface{}
 
 // WindowAggregateStore implements the WindowAggregate capability.
 type WindowAggregateStore interface {
-	// HasWindowAggregateCapability checks if this Store supports the capability.
-	// If WindowAggregateCapability is passed to the method, then the struct
-	// is filled with a detailed list of what the RPC call supports.
-	HasWindowAggregateCapability(ctx context.Context, capability ...*WindowAggregateCapability) bool
+	// GetWindowAggregateCapability will get a detailed list of what the RPC call supports
+	// for window aggregate.
+	GetWindowAggregateCapability(ctx context.Context) WindowAggregateCapability
 
 	// WindowAggregate will invoke a ReadWindowAggregateRequest against the Store.
 	WindowAggregate(ctx context.Context, req *datatypes.ReadWindowAggregateRequest) (ResultSet, error)
--- a/mock/reader.go
+++ b/mock/reader.go
@@ -42,19 +42,16 @@ func (s *StorageReader) Close() {
 
 type WindowAggregateStoreReader struct {
 	*StorageReader
-	HasWindowAggregateCapabilityFn func(ctx context.Context) bool
+	GetWindowAggregateCapabilityFn func(ctx context.Context) query.WindowAggregateCapability
 	ReadWindowAggregateFn          func(ctx context.Context, spec query.ReadWindowAggregateSpec, alloc *memory.Allocator) (query.TableIterator, error)
 }
 
-func (s *WindowAggregateStoreReader) HasWindowAggregateCapability(ctx context.Context, capability ...*query.WindowAggregateCapability) bool {
+func (s *WindowAggregateStoreReader) GetWindowAggregateCapability(ctx context.Context) query.WindowAggregateCapability {
 	// Use the function if it exists.
-	if s.HasWindowAggregateCapabilityFn != nil {
-		return s.HasWindowAggregateCapabilityFn(ctx)
+	if s.GetWindowAggregateCapabilityFn != nil {
+		return s.GetWindowAggregateCapabilityFn(ctx)
 	}
-
-	// Provide a default implementation if one wasn't set.
-	// This will return true if the other function was set.
-	return s.ReadWindowAggregateFn != nil
+	return nil
 }
 
 func (s *WindowAggregateStoreReader) ReadWindowAggregate(ctx context.Context, spec query.ReadWindowAggregateSpec, alloc *memory.Allocator) (query.TableIterator, error) {
--- a/query/storage.go
+++ b/query/storage.go
@@ -9,6 +9,7 @@ import (
 	"github.com/influxdata/flux/memory"
 	"github.com/influxdata/flux/plan"
 	"github.com/influxdata/influxdb/v2"
+	"github.com/influxdata/influxdb/v2/storage/reads"
 	"github.com/influxdata/influxdb/v2/storage/reads/datatypes"
 	"github.com/influxdata/influxdb/v2/tsdb/cursors"
 )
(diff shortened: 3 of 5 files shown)


wolffcm:
I think both Ethan and Jonathan have valid concerns here.  My understanding of Jonathan's concern is that it should be possible to implement `WindowAggregateReader` without needing to import any storage packages directly.  Do I have that right?

What if we made `WindowAggregateCapability` an interface in both storage code and in `query/stdlib`, with the latter being defined in terms of the storage interface?

```go
// storage code
type WindowAggregateCapability interface {
    // ...
}

type windowAggregateCapability struct {}

type WindowAggregateStore interface {
	HasWindowAggregateCapability(ctx context.Context, capability ...WindowAggregateCapability) bool
}
```

```go
// query/stdlib/influxdata/influxdb/storage.go
type WindowAggregateCapability interface {
    reads.WindowAggregateCapability
}

	HasWindowAggregateCapability(ctx context.Context, capability ...WindowAggregateCapability) bool
}
```
This way it should be possible to implement `WindowAggregateReader` but the implementor will not need to import any storage packages directly.  Also, storage's `WindowAggregateCapability` will implement our stdlib's corresponding capability interface by design, so extra translation code is not needed, and it's very unlikely the two representations get out of sync.

adrian-thurston:
I agree having two similarly named interfaces is confusing. I already was bitten by it and the only reason I could catch it quickly was that I was already thinking about it. But if I didn't know I'd probably be lost.

query/stdlib/influxdata/influxdb/storage.go
```diff
@@ -151,7 +152,10 @@ type ReadWindowAggregateSpec struct {
...
 
 // WindowAggregateCapability describes what is supported by WindowAggregateReader.
-type WindowAggregateCapability struct{}
+type WindowAggregateCapability interface {
+	reads.WindowAggregateCapability
+}
+type windowAggregateCapability struct {}
 
 // WindowAggregateReader implements the WindowAggregate capability.
 type WindowAggregateReader interface {
```
wolffcm, line 161: You'll want to remove the pointer `*` in the prototype for `HasWindowAggregateCapability` below, since interfaces are usually passed by value in Go.

storage/reads/store.go
```diff
@@ -84,7 +84,9 @@ type Store interface {
 }
 
 // WindowAggregateCapability describes what is supported by WindowAggregateStore.
-type WindowAggregateCapability struct{}
+type WindowAggregateCapability interface{}
```
wolffcm, line 87: Don't you want to define the required methods for the interface here?

yzhang1991: Currently there is no method defined, this is something to be added later.

wolffcm: It seems like to unblock Adrian's work, the methods on the interface will need to be defined, won't they?

yzhang1991: At this stage, as long as we see the `Reader` implements the `WindowAggregateReader` interface, it means we can push down one window aggregate function. Stabilizing the `WindowAggregateReader` interface will be enough to unblock @adrian-thurston. 

The methods in this interface will be for more detailed flags. For example, in the future we want to know if the storage engine can support pushing down *MULTIPLE* aggregates instead of one, then in addition to checking whether the `Reader` implements the `WindowAggregateReader` interface, we need to call `GetWindowAggregateCapability()` and check its `SupportMultipleAggregate()` method.

wolffcm: It seems like the planner rule will want to know which aggregate functions are supported for pushdown, right?  Since only `count` will be supported initially.  It seems like we need some detail around that for the planner rule to work properly. Am I missing something?

yzhang1991: My understanding is that we are only going to start to use the methods in this interface after our first version goes into production (i.e., when we develop V2). I imagine the checks in the planner rule will be like

```go
func aggSupported(aggKind ProcedureKind) bool {
	if reader, ok := r.(influxdb.WindowAggregateReader); !ok {
		return false
	}
	capability := reader.GetWindowAggregateCapability()
	switch aggKind {
	case universe.CountKind:
	case **some other supported kinds in V1**:
		return true
	case universe.MeanKind:
		return capability.SupportMean()
	case **others**:
		...
	}
}
```

wolffcm: Ok, I don't want to block work from proceeding.  I seem to remember some of this discussion, but I'm unclear on the details.  It seems like it would be nice if the capabilities were in place and the rule "just works" and does the right thing as new capabilities come online.

adrian-thurston: @wolffcm @yzhang1991 The underlying communication with the server needs to have a fixed design and must be extensible. For example, a dictionary mapping names to boolean values. You can add items to the dictionary to represent additional capabilities. Newer versions of the queryd service can talk to older versions of storage. The protocol is fixed, just the data is changing.

The Go interface that the planner uses can grow and evolve as we change our needs. It just needs to use the fixed interface for communicating with storage. It can hide those details from the planner.

So for now I can interpret the fact that the reader casts to a WindowAggregateReader as indication that Count will work. Then we can update it as we go.

adrian-thurston: On second thought ... I'll add the `HaveMin`/`HaveMax` etc functions as it will allow me to fill in the rule and test it. Then it should be ready, those functions just need implementations that flip the bit when ready.

adrian-thurston: I'm already adding the feature flags so why not the capabilities.

jsternberg: As a general rule, I try to avoid interfaces for APIs that can be or plan to be changed. The reason for this is that interfaces have some notable drawbacks in Go.

1. There is no default implementation.
2. Adding to an interface breaks code that uses it in contrast to everything else in Go.
3. When you use an interfaces sentinel null value to denote that something doesn't exist, you invite panics when a programmer doesn't realize the function can return null.

For things like capabilities, I think structs work a lot better. The reason is because you can add fields to the struct and they will automatically default to the default value for downstream code. So if we need to add a `HaveFirst` capability in the future, the downstream code that uses this struct, both producing and consuming, automatically gets the default value that it does not support the first aggregate which is correct.

The reason why I had the function return a boolean instead of the struct was because I wanted to make it possible to check the capability without assigning the value. For example, you could do this:

```
GetWindowAggregate(context.Context) (WindowAggregateCapability, bool)
```

But then you have to assign the value when you may not care what the value is as the capability doesn't have any attributes. This is why I preferred to have the capability be set as an optional argument because we didn't need it at that exact moment, but would likely need to expand it in the future.

We can potentially just switch to the above. The `master` branch didn't have the structs in a neutral location so, as long as we have a neutral location such as the `query` package, it might be fine to just keep the implementation there.

PR #17908, opened 2020-04-29
fix: monaco polish

### LINKED ISSUE #17820: There's no indication that a query is running when using ctl+enter
When using the flux editor, there is no indication that a query is actually running when using ctl-enter. When you click the "submit" button, there is a spinner on the button.

We need a clear way to indicate to the user that their query is currently being executed.

### LINKED ISSUE #17821: Standardize keyboard shortcuts
right now, you use ctl+enter to submit and cmd+/ to comment out the lines.

We need to standardize them and update the documentation in the ? to reflect the correct shortcuts.

This is a regression because before monaco, these keyboard shortcuts were standardized.

drdelambre:
Kapture 2020-04-29 at 13 09 07

Closes #17820
Closes #17821

there was some weird code in there about "only show the loading icon if it was clicked by a person", and i threw that out. also added some verbage around "use ctrl OR cmd" to capture the different environments. This turns off ctrl on macs in favor of cmd. @russorat weigh in if we still want to force the ctrl key on macs in addition to cmd.

### DIFF
--- a/ui/src/external/monaco.flux.hotkeys.ts
+++ b/ui/src/external/monaco.flux.hotkeys.ts
@@ -62,7 +62,7 @@ export function comments(editor: EditorType) {
 
 export function submit(editor: EditorType, submitFn: () => any) {
   editor.addCommand(
-    [window.monaco.KeyMod.CtrlCmd | window.monaco.KeyCode.Enter],
+    window.monaco.KeyMod.CtrlCmd | window.monaco.KeyCode.Enter,
     () => {
       submitFn()
     }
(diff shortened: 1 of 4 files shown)


russorat:
> weigh in if we still want to force the ctrl key on macs in addition to cmd.
i'm ok with just having cmd on macs.

Is it possible to update the text in the ? dialog to match the OSS/Browser behavior but if that's a lot of work, i think this is good.

drdelambre:
the `or` bugs me too. I can't find a way to surface how monaco determines which one to use, and hesitate to use anything but their internal definition as it'll just cause perpetual drift between those two components.

PR #17909, opened 2020-04-29
feat(github): removing the Github social sign-up option until further notice

Asalem1:
### Problem 

This feature is set to release very soon. In the event of any failures with the feature, we want to reduce the scope of the problem in order to get a more focused understanding of what is going wrong, and where we should focus our efforts. In order to narrow the scope of this feature, we are commenting out the Github option for social sign-ups.

### Solution

By deleting the Github-related UI, we are limiting social sign-up to only allow for Google users. Once this feature has proven successful, we will reintroduce the code to allow Github integration

Screen Shot 2020-04-29 at 14 13 43

### DIFF
--- a/ui/src/onboarding/containers/LoginPageContents.tsx
+++ b/ui/src/onboarding/containers/LoginPageContents.tsx
@@ -21,7 +21,7 @@ import auth0js, {WebAuth} from 'auth0-js'
 // Components
 import {LoginForm} from 'src/onboarding/components/LoginForm'
 import {SocialButton} from 'src/shared/components/SocialButton'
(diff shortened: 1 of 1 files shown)


Asalem1:
@drdelambre couldn't agree more. Based on discussions had with @lukevmorris it was determined that the best course of action in this case would be to simply comment the code out rather than integrating a feature flag since it should only be behind a flag for a few days

drdelambre:
you can safely delete it and revert this commit later, keeping master clean and achieving the same result. there's been a lot of "but it's only for just now" turning into tech debt recently.

PR #17913, opened 2020-04-29
chore: upgrade ts to 3.8.3

121watts:
### What && Why

TypeScript 3.8 introduces a bunch of [cool features](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html)

Specifically, I'm super excited about [Optional Chaning](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) which would allow us to replace / stop using `lodash/get` and the foo && foo.bar && foo.baz patterns!

All we had to do to upgrade was fix a couple of conflicting import / exports

### DIFF
--- a/ui/src/types/alerting.ts
+++ b/ui/src/types/alerting.ts
@@ -9,12 +9,12 @@ import {
   SMTPNotificationRuleBase,
   PagerDutyNotificationRuleBase,
   HTTPNotificationRuleBase,
-  Check as GenCheck,
-  ThresholdCheck as GenThresholdCheck,
-  DeadmanCheck as GenDeadmanCheck,
+  Check as GCheck,
+  ThresholdCheck as GThresholdCheck,
+  DeadmanCheck as GDeadmanCheck,
   CustomCheck as GenCustomCheck,
-  NotificationRule as GenRule,
-  NotificationEndpoint as GenEndpoint,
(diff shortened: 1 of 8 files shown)


ui/src/types/labels.ts
```diff
@@ -1,9 +1,9 @@
-import {Label as GenLabel} from 'src/client'
+import {Label as GLabel} from 'src/client'
 import {RemoteDataState} from 'src/types'
 
 // GenLabel is the shape of a Label returned from the server -- before it has
 // been modified with UI specific fields
-export type GenLabel = GenLabel
+export type GenLabel = GLabel
```
drdelambre, line 6: well will you look at that

ui/src/types/alerting.ts
```diff
@@ -50,7 +50,7 @@ export type NotificationEndpointBase = Omit<GenEndpointBase, 'labels'> &
 type RuleOverrides = {status: RemoteDataState; activeStatus: TaskStatusType}
 
 // GenRule is the shape of a NotificationRule from the server -- before any UI specific fields are added or modified
-export type GenRule = GenRule
+export type GenRule = GRule
```
Asalem1, line 53: The only rules I live by

PR #17917, opened 2020-04-30
feat(pkger): add list operation to stacks

jsteenb2:
- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -4375,6 +4375,84 @@ paths:
               schema:
                 $ref: "#/components/schemas/Error"
   /packages/stacks:
+    get:
+      operationId: ListStacks
+      tags:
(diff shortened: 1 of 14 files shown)


http/swagger.yml
```diff
@@ -4375,6 +4375,86 @@ paths:
...
+            properties:
+              orgID:
+                description: The organization id of the stacks
+                type: string
+              name[]:
+                description: A collection of names to filter the list by. The list is comma delimited.
+                type: array
+                items:
+                  type: string
+              stackID[]:
```
jsteenb2, line 4396: not sure why, but open-api specs dont' follow the canonical array format for `application/x-www-form-urlencoded` forms... that is why you see the `[]` here

PR #17926, opened 2020-04-30
feat: frontend flags

drdelambre:
Last piece for influxdata/idpe#6654 until we start testing IRL. Works fine with the OSS backend. Mimics the `GetMe` and `GetOrganizations` container patterns in order to ensure the user has loaded flags for themselves before beginning.

sorry about the whitespace only changes in `src/index.jsx`, but i gotta get them stats up somehow

### DIFF
--- a/ui/src/Logout.tsx
+++ b/ui/src/Logout.tsx
@@ -1,5 +1,6 @@
 // Libraries
 import {FC, useEffect} from 'react'
+import {connect} from 'react-redux'
 import {withRouter, WithRouterProps} from 'react-router'
 
 // APIs
@@ -10,8 +11,14 @@ import {CLOUD, CLOUD_URL, CLOUD_LOGOUT_PATH} from 'src/shared/constants'
 
 // Components
 import {ErrorHandling} from 'src/shared/decorators/errors'
+import {reset} from 'src/shared/actions/flags'
 
-const Logout: FC<WithRouterProps> = ({router}) => {
+interface DispatchProps {
+  reset: typeof reset
+}
+
+type Props = DispatchProps & WithRouterProps
+const Logout: FC<Props> = ({router, reset}) => {
   const handleSignOut = async () => {
     if (CLOUD) {
       window.location.href = `${CLOUD_URL}${CLOUD_LOGOUT_PATH}`
@@ -28,9 +35,19 @@ const Logout: FC<WithRouterProps> = ({router}) => {
   }
 
   useEffect(() => {
+    reset()
     handleSignOut()
   }, [])
   return null
 }
 
-export default ErrorHandling(withRouter<WithRouterProps>(Logout))
+const mdtp = {
+  reset: reset,
+}
+
+export default ErrorHandling(
+  connect<{}, DispatchProps>(
+    null,
+    mdtp
+  )(withRouter<WithRouterProps>(Logout))
+)
--- a/ui/src/Logout.tsx
+++ b/ui/src/Logout.tsx
@@ -18,7 +18,7 @@ interface DispatchProps {
 }
 
 type Props = DispatchProps & WithRouterProps
-const Logout: FC<Props> = ({router, reset}) => {
+const Logout: FC<Props> = ({router, resetFeatureFlags}) => {
   const handleSignOut = async () => {
     if (CLOUD) {
       window.location.href = `${CLOUD_URL}${CLOUD_LOGOUT_PATH}`
--- a/ui/src/Logout.tsx
+++ b/ui/src/Logout.tsx
@@ -14,7 +14,7 @@ import {ErrorHandling} from 'src/shared/decorators/errors'
 import {reset} from 'src/shared/actions/flags'
 
 interface DispatchProps {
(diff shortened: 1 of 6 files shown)


ui/src/Logout.tsx
```diff
@@ -10,8 +11,14 @@ import {CLOUD, CLOUD_URL, CLOUD_LOGOUT_PATH} from 'src/shared/constants'
 
 // Components
 import {ErrorHandling} from 'src/shared/decorators/errors'
+import {reset} from 'src/shared/actions/flags'
 
-const Logout: FC<WithRouterProps> = ({router}) => {
+interface DispatchProps {
+  reset: typeof reset
```
Asalem1, line 17:
```suggestion
  resetFeatureFlags: typeof reset
```

ui/src/Logout.tsx
```diff
@@ -28,9 +35,19 @@ const Logout: FC<WithRouterProps> = ({router}) => {
   }
 
   useEffect(() => {
+    reset()
```
Asalem1, line 38:
```suggestion
    resetFeatureFlags()
```

ui/src/Logout.tsx
```diff
@@ -28,9 +35,19 @@ const Logout: FC<WithRouterProps> = ({router}) => {
...
     handleSignOut()
   }, [])
   return null
 }
 
-export default ErrorHandling(withRouter<WithRouterProps>(Logout))
+const mdtp = {
+  reset: reset,
```
Asalem1, line 45:
```suggestion
  resetFeatureFlags: reset,
```

ui/src/Logout.tsx
```diff
@@ -28,9 +35,19 @@ const Logout: FC<WithRouterProps> = ({router}) => {
...
+}
+
+export default ErrorHandling(
```
Asalem1, line 48: The main reason for requesting the name change here is twofold:

1. to be more explicit on what we're resetting without having to investigate
2. To reduce confusion about whether `reset` is being called after being imported or as a prop.

ui/src/shared/actions/flags.ts
```diff
@@ -1,19 +1,33 @@
...
-export const setFlags = flags =>
+export const setFlags = (status: RemoteDataState, flags?: FlagMap) =>
   ({
     type: SET_FEATURE_FLAGS,
-    payload: flags,
+    payload: {
+      status,
+      flags,
+    },
+  } as const)
```
Asalem1, line 26: out of curiosity, what's the benefit of keeping this as a `const` rather than a custom `type`. Genuinely curious on your take

ui/src/shared/actions/flags.ts
```diff
@@ -28,3 +42,23 @@ export const setOverride = (flag: string, value: string | boolean) =>
...
+    if (resp.status !== 200) {
+      throw new Error(resp.data.message)
+    }
+
+    dispatch(setFlags(RemoteDataState.Done, resp.data))
+
+    return resp.data
+  } catch (error) {
+    console.error(error)
+    dispatch(setFlags(RemoteDataState.Error, null))
```
Asalem1, line 62: Assuming that feature flags are internal-facing, should we also include a notification here to let the user know something went wrong to reduce confusion / investigation if something does happen?

ui/src/shared/containers/GetFlags.tsx
```diff
@@ -0,0 +1,53 @@
...
+import {SpinnerContainer, TechnoSpinner} from '@influxdata/clockface'
+
+// Types
+import {RemoteDataState, AppState} from 'src/types'
+
+// Actions
+import {getFlags as getFlagsAction} from 'src/shared/actions/flags'
+
+interface PassedInProps {
+  children: React.ReactElement<any>
```
Asalem1, line 15: love being explicit here, but do we need this?

ui/src/shared/actions/flags.ts
drdelambre: keeps the action naming pattern cleaner. we experimented with a few when watts was doing the resource normalization work (like where you define an action and an actiontype), this had the least amount of boilerplate.  100% ergonomics and maintainability over "correctness" in testing. We found that the interface mapping / typescript benefits come more with the "typeof" during consumption than the tight coupling with the reducer (which exists as convention).

ui/src/shared/actions/flags.ts
drdelambre: i dont understand

ui/src/shared/containers/GetFlags.tsx
drdelambre: :copy: :pasta:

ui/src/shared/actions/flags.ts
Asalem1: Basically, if there’s an error in setting the feature flag, shouldn’t the user know it by receiving a notification rather than by having to investigate in some other channel if their expectation is that it should be on?

drdelambre: offlined and both agreed that this is cabbage's job

PR #17933, opened 2020-05-01
fix(zero_value_checks): checks can now have a value set to 0

### LINKED ISSUE #17729: UI: Check threshold == 0 impossible
### Problem
Create a check.
Set the threshold to "is above 0"
Click the green checkbox.
Wait "schedule every".
Notice that the check has warning "Last run status: Completed(failed)".
Open check.
Notice that the threshold is empy.

### Workaround
Set the threshold to 0.00001.
Save the check.
Wait "schedule every".
Warning clears, threshold not empty.

### Also
It would be nice to have more info than "Completed(failed)". I don't know why it failed.

Asalem1:
Closes #17729

### Problem

1. Checks initialized with the value `0` would not be triggered. When a user would edit their check, the value would return blank
2. On smaller screens, the Configure A Check options would be off the screen:

Screen Shot 2020-05-01 at 13 30 13
 
### Solution

Updated the threshold value to remove `omitEmpty` on thresholds since that was removing values set to zero.

As per [this helpful resource](https://www.sohamkamani.com/golang/2018-07-19-golang-omitempty/), go will remove nil values if the flag `omitEmpty` is present. 

It doesn't seem like there was any obvious reason for this to be present since a check should never be created without a value set for the threshold. The UI doesn't allow for this condition, and the API shouldn't allow for this condition. 

I added an e2e test to ensure that this bug doesn't pop its head again, and I added an additional check to ensure that we can name & rename checks. I also updated the AlertBuilder.scss to fix weird styling issue

- [x] Rebased/mergeable

check

### DIFF
--- a/ui/src/alerting/components/builder/AlertBuilder.scss
+++ b/ui/src/alerting/components/builder/AlertBuilder.scss
@@ -25,13 +25,24 @@
   }
 
   &.alert-builder--meta-card {
-    flex: 1 0 280px !important;
+    flex: 1 0 140px !important;
   }
   &.alert-builder--message-card {
-    flex: 3 0 320px !important;
+    flex: 3 0 140px !important;
   }
   &.alert-builder--conditions-card {
-    flex: 2 0 320px !important;
+    flex: 2 0 200px !important;
+  }
+  @media screen and (min-width: $cf-grid--breakpoint-md) {
+    &.alert-builder--meta-card {
+      flex: 1 0 280px !important;
+    }
+    &.alert-builder--message-card {
+      flex: 3 0 320px !important;
+    }
+    &.alert-builder--conditions-card {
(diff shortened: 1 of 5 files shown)


ui/src/alerting/components/builder/AlertBuilder.scss
```diff
@@ -25,13 +25,24 @@
...
+  @media screen and (min-width: $cf-grid--breakpoint-md) {
+    &.alert-builder--meta-card {
+      flex: 1 0 280px !important;
+    }
+    &.alert-builder--message-card {
+      flex: 3 0 320px !important;
+    }
+    &.alert-builder--conditions-card {
+      flex: 2 0 320px !important;
+    }
```
Asalem1, line 45: @alexpaxton would love your eyes on this

Issue #17554, opened 2020-04-01
Add ability to remove stack and all its resources


jsteenb2:
tbd: how we want to manage this

PR #17934, opened 2020-05-01
feat(pkger): add the ability to remove a stack and all its associated resources

jsteenb2:
closes: #17554

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/pkger/http_server.go
+++ b/pkger/http_server.go
@@ -48,11 +48,14 @@ func NewHTTPServer(log *zap.Logger, svc SVC) *HTTPServer {
 	{
 		r.With(middleware.AllowContentType("text/yml", "application/x-yaml", "application/json")).
 			Post("/", svr.createPkg)
+
 		r.With(middleware.SetHeader("Content-Type", "application/json; charset=utf-8")).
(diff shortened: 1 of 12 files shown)


pkger/http_server.go
```diff
@@ -203,6 +206,63 @@ func (s *HTTPServer) createStack(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	auth, err := pctx.GetAuthorizer(r.Context())
+	if err != nil {
+		s.api.Err(w, err)
+		return
+	}
+	userID := auth.GetUserID()
+
+	err = s.svc.DeleteStack(r.Context(), struct{ OrgID, UserID, StackID influxdb.ID }{
```
hoorayimhelping, line 233: i'm not good enough with go to trace this all the way through, but is this an idempotent operation? will

`http DELETE /stacks/1337`
`http DELETE /stacks/1337`

return `204`s both times?

jsteenb2: ahh shoot, good call. I didn't write a test for that 🤦

jsteenb2: to answer yoru question, yes it should

Issue #17940, opened 2020-05-02
pkger enforce resource metadata.name field is dns compliant


jsteenb2:
Pkger resources metadata.name field should be DNS compliant

- [x] resource names must be between 3 and 63 characters long.
- [x] resource names can consist only of lowercase letters, numbers, dots (.),  and hyphens (-).
- [x] resource names must not be formatted as an IP address (for example, 192.168.5.4).

useful link from std lib:
`isDomainName`: https://golang.org/src/net/dnsclient.go

PR #17941, opened 2020-05-03
feat(pkger): enforce metadata.name dns name compliance

jsteenb2:
closes: #17940

- [x] Rebased/mergeable
- [x] Tests pass


cmd/influxd/launcher/pkger_test.go
```diff
@@ -636,7 +636,7 @@ func TestLauncher_Pkger(t *testing.T) {
 				initialRulePkgName     = "oh_doyle_rules"
 				initialTaskPkgName     = "tap"
 				initialTelegrafPkgName = "teletype"
-				initialVariablePkgName = "laces out dan"
+				initialVariablePkgName = "laces_out_dan"
```
hoorayimhelping, line 639: 90s movies references, 😂

pkger/parser.go
```diff
@@ -1152,14 +1152,14 @@ func (p *Pkg) eachResource(resourceKind Kind, minNameLen int, fn func(o Object)
...
-		if len(k.Name()) < minNameLen {
+		if !isDomainName(k.Name()) {
 			pErr.append(resourceErr{
 				Kind: k.Kind.String(),
 				Idx:  intPtr(i),
 				ValidationErrs: []validationErr{
 					objectValidationErr(fieldMetadata, validationErr{
 						Field: fieldName,
-						Msg:   fmt.Sprintf("must be a string of at least %d chars in length", minNameLen),
+						Msg:   "must be a between 3-64 characters long and may only consist of alphanumeric, dots, underderscores, and hyphen characters.",
```
hoorayimhelping, line 1162:
```suggestion
						Msg:   "must be between 3-64 characters long and may only consist of alphanumeric, dots, underderscores, and hyphen characters.",
```

might be nice to link to official docs on dns name compliance. might not be though if the context is wrong 🤷

rawkode:
The Kubernetes naming conventions are a little more strict than this, based on RFC1123:

```
DNS Label Names
Some resource types require their names to follow the DNS label standard as defined in RFC 1123. This means the name must:

contain at most 63 characters
contain only lowercase alphanumeric characters or ‘-’
start with an alphanumeric character
end with an alphanumeric character
```

pkger/parser.go
```diff
@@ -1434,6 +1434,54 @@ func parseChart(r Resource) (chart, []validationErr) {
...
+// isDomainName is composed of 98% of the std lib implementation, only
+// difference in ours is that we only allow lower case letters.
+func isDomainName(s string) bool {
+	l := len(s)
+	if l == 0 || l > 254 || l == 254 && s[l-1] != '.' {
+		return false
+	}
+
+	last := byte('.')
+	nonNumeric := false // true once we've seen a letter or hyphen
```
hoorayimhelping, line 1446: nit: this is a bit of a confusing variable name - what effect does it being nonNumeric have on this algorithm? 

>resource names can consist only of lowercase letters, numbers, dots (.), underscores (_), and hyphens (-).

given that description, i'm a little confused on what it means to be non numeric. it sounds like it's a proxy for "isDomainName." does it have to do with it not being an ip address? as in, we need a character for it to be a valid dns name, otherwise it's an ip address? 

is there some subtlety i'm missing? i feel like this could be more clearly accomplished with `isValidDomainName`.

the reason i bring it up is because this bit of code is quite intense and the clearer it its, the fewer edge cases we'll miss

jsteenb2:
@rawkode sweet, thanks for the heads up

pkger/parser.go
jsteenb2: I copy pastaed this from the go std lib :-)

pkger/parser.go
jsteenb2: I'm hesitant to link docs from k8s here. I'll link them when this gets added to our internal docs though.

pkger/parser.go
jsteenb2: its going away

PR #17944, opened 2020-05-04
feat: add feature flag for refactored authorization package

AlirieGray:
This PR adds a feature flag for the new authorization package introduced in: https://github.com/influxdata/influxdb/pull/17874

Also included is a new interface `FeatureHandler` which takes in an Enabler (in this case a Flag) and two handlers and uses the Enabler to decide which handler to use. This allows us to switch on feature flagged code at the http level instead of the service level.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/api_handler.go
+++ b/http/api_handler.go
@@ -125,10 +125,6 @@ func NewAPIHandler(b *APIBackend, opts ...APIHandlerOptFn) *APIHandler {
 
 	h.Mount("/api/v2", serveLinksHandler(b.HTTPErrorHandler))
 
-	authorizationBackend := NewAuthorizationBackend(b.Logger.With(zap.String("handler", "authorization")), b)
-	authorizationBackend.AuthorizationService = authorizer.NewAuthorizationService(b.AuthorizationService)
-	h.Mount(prefixAuthorization, NewAuthorizationHandler(b.Logger, authorizationBackend))
-
 	bucketBackend := NewBucketBackend(b.Logger.With(zap.String("handler", "bucket")), b)
 	bucketBackend.BucketService = authorizer.NewBucketService(b.BucketService, noAuthUserResourceMappingService)
 	h.Mount(prefixBuckets, NewBucketHandler(b.Logger, bucketBackend))
--- a/kit/transport/http/feature_controller.go
+++ b/kit/transport/http/feature_controller.go
@@ -1,0 +1,39 @@
+package http
+
+import (
+	"context"
+	"net/http"
+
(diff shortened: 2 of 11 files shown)


mock/tenant_service.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"context"
+
+	"github.com/influxdata/influxdb/v2"
+	platform "github.com/influxdata/influxdb/v2"
+)
+
+// TenantService is a mock implementation of an authorization.TenantService
+type TenantService struct {
```
lyondhill, line 11: the name here is confusing. I know it relates to your http specific tenant service implementation but there is also a influxdb.TenantService. I think this mock could be scoped to your http service.

http/api_handler.go
```diff
@@ -125,10 +125,6 @@ func NewAPIHandler(b *APIBackend, opts ...APIHandlerOptFn) *APIHandler {
 
 	h.Mount("/api/v2", serveLinksHandler(b.HTTPErrorHandler))
 
-	authorizationBackend := NewAuthorizationBackend(b.Logger.With(zap.String("handler", "authorization")), b)
```
brettbuddin, line 128: How does this ladder up to `idpe`? Removing this here will cause the handlers to be missing if someone updates the `influxdb` dependency. I think we'll need to make this tolerant of that scenario.

kit/transport/http/feature_controller.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// Enabler allows the switching between two HTTP Handlers
+type Enabler interface {
+	Enabled(ctx context.Context, flagger ...feature.Flagger) bool
+}
+
+// FeatureHandler is used to switch requests between an exisiting and a feature flagged
+// HTTP Handler on a per-request basis
+type FeatureHandler struct {
```
brettbuddin, line 17: Minor nit: This file needs to be renamed to `feature_handler.go`

http/api_handler.go
AlirieGray: this is how I added it in idpe: https://github.com/influxdata/idpe/pull/7054/files#diff-f047f7de48d9a985ade16057d9d48739R109-R124

brettbuddin: Okay, so the intention is that you will merge this then *immediately* merge the other? I would be careful with this assumption in some cases, because there could be some delay in your ability to merge that's out of your control. Would it be possible to merge the `idpe` changes first?

PR #17950, opened 2020-05-04
feat(session): Build out a new session service

lyondhill:
This new session service has the ability to work independant of other systems
it relies on having its own store type which should allow us to be more flexible
then using the built in kv system.

I have included an in-mem session store.

### DIFF
--- a/inmem/session_store.go
+++ b/inmem/session_store.go
@@ -1,0 +1,86 @@
+package inmem
+
+import (
+	"errors"
+	"sync"
+	"time"
+)
+
+type SessionStore struct {
+	data map[string]string
+
+	timers map[string]*time.Timer
+
+	mu sync.RWMutex
+}
+
+func NewSessionStore() *SessionStore {
+	return &SessionStore{
+		data:   map[string]string{},
+		timers: map[string]*time.Timer{},
+	}
+}
+
+func (s *SessionStore) Set(key, val string, expireAt time.Time) error {
+	if !expireAt.IsZero() && expireAt.Before(time.Now()) {
+		// key is already expired. no problem
+		return nil
+	}
+
+	s.mu.Lock()
+	s.data[key] = val
+	s.mu.Unlock()
+
+	if !expireAt.IsZero() {
+		return s.ExpireAt(key, expireAt)
+	}
+	return nil
+}
+
+func (s *SessionStore) Get(key string) (string, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	return s.data[key], nil
+}
+
+func (s *SessionStore) Delete(key string) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	timer := s.timers[key]
+	if timer != nil {
+		timer.Stop()
+	}
+
+	delete(s.data, key)
+	delete(s.timers, key)
+	return nil
+}
+
+func (s *SessionStore) ExpireAt(key string, expireAt time.Time) error {
+	s.mu.Lock()
+
+	existingTimer, ok := s.timers[key]
+	if ok {
+		if !existingTimer.Stop() {
+			return errors.New("session has expired")
+		}
+
+	}
+
+	duration := time.Until(expireAt)
+	if duration <= 0 {
+		s.mu.Unlock()
+		s.Delete(key)
+		return nil
+	}
+	s.timers[key] = time.AfterFunc(time.Until(expireAt), s.timerExpireFunc(key))
+	s.mu.Unlock()
+	return nil
+}
+
+func (s *SessionStore) timerExpireFunc(key string) func() {
+	return func() {
+		s.Delete(key)
+	}
+}
--- a/session/http_server.go
+++ b/session/http_server.go
@@ -1,0 +1,194 @@
+package session
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/go-chi/chi"
+	"github.com/go-chi/chi/middleware"
+	"github.com/influxdata/influxdb/v2"
+	kithttp "github.com/influxdata/influxdb/v2/kit/transport/http"
+	"go.uber.org/zap"
+)
+
+const (
+	prefixSignIn  = "/api/v2/signin"
+	prefixSignOut = "/api/v2/signout"
+)
+
+// SessionHandler represents an HTTP API handler for authorizations.
+type SessionHandler struct {
+	chi.Router
+	api *kithttp.API
+	log *zap.Logger
+
+	sessionSvc influxdb.SessionService
+	passSvc    influxdb.PasswordsService
+	userSvc    influxdb.UserService
+}
+
+// NewSessionHandler returns a new instance of SessionHandler.
+func NewSessionHandler(log *zap.Logger, sessionSvc influxdb.SessionService, userSvc influxdb.UserService, passwordsSvc influxdb.PasswordsService) *SessionHandler {
+	svr := &SessionHandler{
+		api: kithttp.NewAPI(kithttp.WithLog(log)),
+		log: log,
+
+		passSvc:    passwordsSvc,
+		sessionSvc: sessionSvc,
+		userSvc:    userSvc,
+	}
+
+	return svr
+}
+
+type resourceHandler struct {
+	prefix string
+	*SessionHandler
+}
+
+// Prefix is necessary to mount the router as a resource handler
+func (r resourceHandler) Prefix() string { return r.prefix }
+
+// SignInResourceHandler allows us to return 2 different rousource handler
+// for the appropriate mounting location
+func (h SessionHandler) SignInResourceHandler() *resourceHandler {
+	h.Router = chi.NewRouter()
+	h.Router.Use(
+		middleware.Recoverer,
+		middleware.RequestID,
+		middleware.RealIP,
+	)
+	h.Router.Post("/", h.handleSignin)
+	return &resourceHandler{prefix: prefixSignIn, SessionHandler: &h}
+}
+
+// SignOutResourceHandler allows us to return 2 different rousource handler
+// for the appropriate mounting location
+func (h SessionHandler) SignOutResourceHandler() *resourceHandler {
+	h.Router = chi.NewRouter()
+	h.Router.Use(
+		middleware.Recoverer,
+		middleware.RequestID,
+		middleware.RealIP,
+	)
+	h.Router.Post("/", h.handleSignout)
+	return &resourceHandler{prefix: prefixSignOut, SessionHandler: &h}
+}
+
+// handleSignin is the HTTP handler for the POST /signin route.
+func (h *SessionHandler) handleSignin(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+
+	req, decErr := decodeSigninRequest(ctx, r)
+	if decErr != nil {
+		h.api.Err(w, ErrUnauthorized)
+		return
+	}
+
+	u, err := h.userSvc.FindUser(ctx, influxdb.UserFilter{
+		Name: &req.Username,
+	})
+	if err != nil {
+		h.api.Err(w, ErrUnauthorized)
+		return
+	}
+
+	if err := h.passSvc.ComparePassword(ctx, u.ID, req.Password); err != nil {
+		h.api.Err(w, ErrUnauthorized)
+		return
+	}
+
+	s, e := h.sessionSvc.CreateSession(ctx, req.Username)
+	if e != nil {
+		h.api.Err(w, ErrUnauthorized)
+		return
+	}
+
+	encodeCookieSession(w, s)
+	w.WriteHeader(http.StatusNoContent)
+}
+
+type signinRequest struct {
+	Username string
+	Password string
+}
+
+func decodeSigninRequest(ctx context.Context, r *http.Request) (*signinRequest, *influxdb.Error) {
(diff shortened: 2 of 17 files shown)


session/http_server.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	r := chi.NewRouter()
+	r.Use(
+		middleware.Recoverer,
+		middleware.RequestID,
+		middleware.RealIP,
+	)
+
+	// RESTy routes for "articles" resource
```
GeorgeMac, line 49: looks like copy-paste comment left behind

session/http_server.go
```diff
@@ -0,0 +1,205 @@
...
+
+// handleSessionRequest is used because this is a mounted router and we need to check the prefix
+// to ensure we route to the correct place
+func (h *SessionHandler) handleSessionRequest(w http.ResponseWriter, r *http.Request) {
+	switch {
+	case strings.Contains(r.URL.Path, prefixSignIn):
+		h.handleSignin(w, r)
+	case strings.Contains(r.URL.Path, prefixSignIn):
+		h.handleSignout(w, r)
+	}
```
GeorgeMac, line 86: Why is this using strings contains and not the routers prefix trie?

session/http_server_test.go
```diff
@@ -0,0 +1,102 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			userSVC := mock.NewUserService()
+			userSVC.FindUserFn = func(_ context.Context, f influxdb.UserFilter) (*influxdb.User, error) {
+				return &influxdb.User{ID: 1}, nil
+			}
+			h := NewSessionHandler(zaptest.NewLogger(t), tt.fields.SessionService, userSVC, tt.fields.PasswordsService)
+
+			router := chi.NewRouter()
+			router.Mount("/api/v2/signin", h)
```
GeorgeMac, line 76: I feel like this should be testing the `SessionHandler` route setup and not having to remount things at test time.
Can this be simplified to:
```
h := NewSessionHandler(zaptest.NewLogger(t), tt.fields.SessionService, userSVC, tt.fields.PasswordsService)
server := httptest.NewServer(h)
```

session/service.go
```diff
@@ -0,0 +1,121 @@
...
+	store         *Storage
+	userService   influxdb.UserService
+	urmService    influxdb.UserResourceMappingService
+	sessionLength time.Duration
+
+	idGen    influxdb.IDGenerator
+	tokenGen influxdb.TokenGenerator
+}
+
+func NewService(store *Storage, userService influxdb.UserService, urmService influxdb.UserResourceMappingService, sessionLength time.Duration) *Service {
```
GeorgeMac, line 22: Lots of these things don't have any function signature docs. Maybe I just am a pedant after years of golint screaming at me to add them, but what is our policy for docs? (I genuinely don't know) I probably would at the minimum give the core service types and functions a little explanation.

session/service.go
```diff
@@ -0,0 +1,121 @@
...
+	token, err := s.tokenGen.Token()
+	if err != nil {
+		return nil, err
+	}
+
+	session := &influxdb.Session{
+		ID:          s.idGen.ID(),
+		Key:         token,
+		CreatedAt:   time.Now(),
+		ExpiresAt:   time.Now().Add(s.sessionLength),
```
GeorgeMac, line 90: This is a take it or leave it nit, but if we use the same value for now here then it will always be precisely sessionLength between created and expires.

session/service.go
```diff
@@ -0,0 +1,121 @@
...
+			mappings, _, err = s.urmService.FindUserResourceMappings(ctx, influxdb.UserResourceMappingFilter{UserID: u.ID}, influxdb.FindOptions{Offset: i, Limit: 100})
+			if err != nil {
+				return nil, err
+			}
+			pms, err := permissionSet(mappings, u.ID)
+			if err != nil {
+				return nil, err
+			}
+			permissions = append(permissions, pms...)
+			i += len(mappings)
```
GeorgeMac, line 77: This could be in the `for` loops post-expression:
```go
for i := len(mappings); len(mappings) > 0; i += len(mappings) {
```

session/service.go
```diff
@@ -0,0 +1,121 @@
...
+		if err != nil {
+			return nil, &influxdb.Error{
+				Err: err,
+			}
+		}
+
+		ps = append(ps, p...)
+	}
+
+	return append(ps, influxdb.MePermissions(userID)...), nil
```
GeorgeMac, line 120: This is going to result in `MePermissions * no. of pages`

session/storage.go
```diff
@@ -0,0 +1,135 @@
...
+
+	id, err := influxdb.IDFromString(val)
+	if err != nil {
+		return nil, err
+	}
+	return s.FindSessionByID(ctx, *id)
+}
+
+func (s *Storage) FindSessionByID(ctx context.Context, id influxdb.ID) (*influxdb.Session, error) {
+	val, err := s.store.Get(storePrefix + id.String())
```
GeorgeMac, line 50: Perhaps use `sessionKey(id)` here for consistency.

session/storage.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	if err := s.store.Delete(sessionIndexKey(session.Key)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func sessionKey(id influxdb.ID) string {
```
GeorgeMac, line 129: I think this should be named `sessionID` since session has a `Key`. 
It is confusing that the ID is the key and the key is the index key.

session/storage.go
```diff
@@ -0,0 +1,135 @@
...
+	return storePrefix + id.String()
+}
+
+func sessionIndexKey(key string) string {
+	return storeIndex + key
```
GeorgeMac, line 134: This is a fairly significant change to the existing structure of sessions. So the change is like this right:
```
before:
 sessionsv1 / { session_key }
|--bucket--|

after:
sessionsv1 / { session_id }
sessionsindexv1 / { session_key }
```
Is there a plan for how this is going to be rolled out?

testing/session.go
```diff
@@ -273,14 +273,10 @@ func ExpireSession(
 			err := s.ExpireSession(ctx, tt.args.key)
 			diffPlatformErrors(tt.name, err, tt.wants.err, opPrefix, t)
 
-			session, err := s.FindSession(ctx, tt.args.key)
+			_, err = s.FindSession(ctx, tt.args.key)
 			if err == nil {
 				t.Errorf("expected session to be expired got %v", err)
 			}
-
```
GeorgeMac, line 280: Why are we dropping this comparison? I fixed it up recently because we missed a bug but not implementing it correctly. Which is why I ask.

session/storage.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// use a minute time just so the session will expire if we fail to set the expiration later
+	sessionKey := sessionKey(session.ID)
+	if err := s.store.Set(sessionKey, string(sessionBytes), time.Minute); err != nil {
+		return err
+	}
+
+	// create index
+	indexKey := sessionIndexKey(session.Key)
+	if err := s.store.Set(indexKey, session.ID.String(), time.Minute); err != nil {
```
GeorgeMac, line 80: It seems strange the Set method mandates an expiration which you have to set on separate keys. Ultimately creating a situation where these keys would expire at slightly different times.
Why is the eventual intent for expiration time at the key here? Since we also have a timestamp for expiration on the body of the session itself.

session/storage.go
```diff
@@ -0,0 +1,135 @@
...
+		return err
+	}
+
+	// set expiration
+	if err := s.store.ExpireAt(sessionKey, session.ExpiresAt); err != nil {
+		return err
+	}
+	if err := s.store.ExpireAt(indexKey, session.ExpiresAt); err != nil {
```
GeorgeMac, line 88: I am a little confused why you would want to pass an expiration via Set, which calls ExpireAt, with enough time to call ExpireAt again. If Set has the expiration time in the contract, why do we need to call it again?

session/storage.go
```diff
@@ -0,0 +1,135 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"time"
+
+	"github.com/influxdata/influxdb/v2"
+)
+
+type Store interface {
```
GeorgeMac, line 11: After reflecting on some of my comments, I think this interface is the problem here. We have a Storage type and a Store interface. This Store interface has come session behaviour semantics mixed with key value store requirements.

I personally would just do away with this interface. Instead, I would have a Storage interface defined next to the Service which the concrete type here satisfies and have a few implementations of that interface. e.g. an `inmem.SessionStorage` and a `bolt.SessionStorage`.

GeorgeMac:
Another thing that we haven't answered yet is permissions from authorizations (tokens). OSS still uses these because of hearsay around some features the root user received as a token permission that wasn't granted via a URM. Just need to understand that story before we commit to removing this all together.

session/storage.go
lyondhill: sessions are ephemeral in nature so when the roll out happens users with an active session will be asked to log back in. The other thing is this isn't using kv so there will be no key overlap.

testing/session.go
lyondhill: The test was requiring an error to be returned as well as a session object. I came across a todo about this [here](https://github.com/influxdata/influxdb/pull/17950/files#diff-0e9e292791dba4c98b67f966ab4b9081L82) and decided to bring the behavior of this function call in line with the way we usually write go (either object or error)

session/storage.go
lyondhill: you're right this can be improved.

session/storage.go
lyondhill: As per our conversation this has been resolved by keeping the store interface and removing the storage interface inside session. I believe we would like to draw the line between the data storage layer and the session service. By placing the interface on the store we can translate session to keys in a specific way and allow the back end storage handle expiration.

lyondhill:
> Another thing that we haven't answered yet is permissions from authorizations (tokens). OSS still uses these because of hearsay around some features the root user received as a token permission that wasn't granted via a URM. Just need to understand that story before we commit to removing this all together.

Yep I had missed that. It is in now.

testing/session.go
GeorgeMac: Ah OK cool, I see. If this test expects an error and a nil session, then perhaps we should assert the session is nil and check the error is the specific expired error.

GeorgeMac: I see as well I was wrong. It was the test above that I fixed up. My mistake.

session/middleware_metrics.go
```diff
@@ -0,0 +1,56 @@
...
+
+// NewSessionMetrics creates a new session metrics middleware
+func NewSessionMetrics(reg prometheus.Registerer, s influxdb.SessionService) *SessionMetrics {
+	return &SessionMetrics{
+		rec:        metric.New(reg, "session"),
+		sessionSvc: s,
+	}
+}
+
+// FindSession calls the underlieing session service and tracks RED metrics for the call
```
GeorgeMac, line 30: nit pick: `s/underlieing/underlying/`

session/middleware_metrics.go
```diff
@@ -0,0 +1,56 @@
...
+func (m *SessionMetrics) FindSession(ctx context.Context, key string) (session *influxdb.Session, err error) {
+	rec := m.rec.Record("Find_session")
```
GeorgeMac, line 32: Curiosity comment: Is this metric name in line with other services. Just interested in the name, it is kind of mix of camel and snake case.

testing/session.go
lyondhill: ill add a check here to make sure session is nil :)

session/middleware_metrics.go
lyondhill: your right here snake case is the way we have been doing it. Making a fix now.

inmem/session_store.go
```diff
@@ -0,0 +1,82 @@
...
+		// key is already expired. no problem
+		return nil
+	}
+
+	s.mu.Lock()
+	s.data[key] = val
+	s.mu.Unlock()
+
+	if !expireAt.IsZero() {
+		return s.ExpireAt(key, expireAt)
```
GeorgeMac, line 34: Take it or leave it nit: You could do this in one lock; if you had an un-exported `expireAt` method which both this and `ExpireAt` exported both used.

inmem/session_store.go
```diff
@@ -0,0 +1,82 @@
...
+	delete(s.timers, key)
+	return nil
+}
+
+func (s *SessionStore) ExpireAt(key string, expireAt time.Time) error {
+	s.mu.Lock()
+
+	existingTimer, ok := s.timers[key]
+	if ok {
+		existingTimer.Stop()
```
GeorgeMac, line 64: Food for though: https://golang.org/pkg/time/#Timer.Stop
I know this is mostly for internal / testing use (I assume anyway). The after func goroutine may have already entered here and be waiting on the Lock currently held. In which case `Delete(key)` will be called once the lock here is released, after this function returns.
In terms of practicality it probably doesn't matter too much, this would only happen given a new expiration was added around the same time it was due to expire and Delete doesn't explode if the key does not exist.
Just a little racey, but I can't imagine it being a big problem.
Could always just bottom out here and return nil or return an key expired error or something here when `!existingTime.Stop()`.

lyondhill: I will add some data safety when the timer was already stopped.

PR #17952, opened 2020-05-04
feat(storage): convert ResultSet to table stream for aggregate window

jlapacik:
`handleRead` consumes an aggregate window `ResultSet` and translates it to a flux table stream.

Note this relies on a `WindowAggregateStore` producing a `ResultSet` type upon reading. @rockstar let me know if this is not a valid assumption.

### DIFF
--- a/storage/flux/reader.go
+++ b/storage/flux/reader.go
@@ -401,6 +401,59 @@ const (
 	valueColIdx = 3
 )
 
+func determineTableColsForWindowAggregate(tags models.Tags, typ flux.ColType, aggregate bool) ([]flux.ColMeta, [][]byte) {
+	var size int
+	var cols []flux.ColMeta
+	var defs [][]byte
+
+	if aggregate {
+		// aggregates remove the _time column
+		size = 3
+		cols = make([]flux.ColMeta, size+len(tags))
+		defs = make([][]byte, size+len(tags))
+		cols[0] = flux.ColMeta{
+			Label: execute.DefaultStartColLabel,
+			Type:  flux.TTime,
+		}
+		cols[1] = flux.ColMeta{
+			Label: execute.DefaultStopColLabel,
+			Type:  flux.TTime,
+		}
+		cols[2] = flux.ColMeta{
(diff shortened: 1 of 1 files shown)


storage/flux/reader.go
```diff
@@ -401,6 +401,59 @@ const (
 	valueColIdx = 3
 )
 
+func determineTableColsForWindowAgg(tags models.Tags, typ flux.ColType, aggregate bool) ([]flux.ColMeta, [][]byte) {
```
rockstar, line 404: This is a seemingly small thing, but is there any benefit in truncating "Aggregate" to "Agg"? Could we fully expand it?

storage/flux/reader.go
```diff
@@ -560,11 +616,81 @@ func (wai *windowAggregateIterator) Do(f func(flux.Table) error) error {
...
+			table.Close()
+		}
+		if cur != nil {
+			cur.Close()
+		}
+		rs.Close()
+		wai.cache.Release()
+	}()
+
+READ:
```
rockstar, line 640: Is this the idiomatic way to walk an iterator in go? I'll admit to having some twitchiness over GOTO, but if it's normally how this is done in go, it's fine.

storage/flux/reader.go
jlapacik: To be honest, it's already such a long name that a few more characters won't hurt. There are similar functions with very similar names in this file already, so I didn't want to break from convention. But I do think renaming this function to something shorter would be much more readable.

storage/flux/reader.go
jlapacik: Actually there isn't a GOTO for this label. This label is used because there's a select statement inside of a for loop. By placing a label before the for loop, calling `break LABEL` inside of the select will also break out of the for loop.

This removes the need for multiple break statements and the additional logic that goes with them. It's a pretty common pattern in Go, from what I've observed.

rockstar: Ah yes, okay, this makes sense as a "this is how we get around multiple meanings of break". Thanks for humoring me.

PR #17965, opened 2020-05-06
feat(ui): hide the "Upgrade Now" button for Cloud users when appropriate

TCL735:
Closes idpe 5891

For Cloud Users only.

To provide the best experience for paid users, by default, we hide the "Upgrade Now" button. The button immediately reveals itself if:
- getting org settings from IDPE results in a failure
- getting org settings from IDPE is successful but a particular setting is not found

The particular setting that we are looking for is updated by Quartz and that functionality will be active soon. For paid users, a successful check on this setting means that the button continues to be hidden.

**Screenshots**

getting org settings fails: **Upgrade Now** button reveals itself
upgrade_button_on_failure

getting org settings successful but particular setting not found: **Upgrade Now** button reveals itself
upgrade_button_on_success

getting org settings successful and particular setting found: No upgrade button (no need to upgrade)
no_upgrade_button

### DIFF
--- a/ui/src/cloud/actions/orgsettings.ts
+++ b/ui/src/cloud/actions/orgsettings.ts
@@ -1,5 +1,5 @@
 // API
-import {getOrgSettings as getOrgSettingsAJAX} from 'src/cloud/apis/orgsettings'
+import {fetchOrgSettings} from 'src/cloud/apis/orgsettings'
 
 // Constants
 import {FREE_ORG_HIDE_UPGRADE_SETTING} from 'src/cloud/constants'
@@ -10,36 +10,34 @@ import {GetState, OrgSetting} from 'src/types'
 // Selectors
 import {getOrg} from 'src/organizations/selectors'
 
-export enum ActionTypes {
-  SetOrgSettings = 'SET_ORG_SETTINGS',
-}
+export const SET_ORG_SETTINGS = 'SET_ORG_SETTINGS'
 
-export type Actions = SetOrgSettings
+export type Action = ReturnType<typeof setOrgSettings>
 
-export interface SetOrgSettings {
-  type: ActionTypes.SetOrgSettings
-  payload: {orgSettings: OrgSetting[]}
-}
-
-export const setOrgSettings = (settings: OrgSetting[] = []): SetOrgSettings => {
-  return {
-    type: ActionTypes.SetOrgSettings,
+export const setOrgSettings = (settings: OrgSetting[] = []) =>
+  ({
+    type: SET_ORG_SETTINGS,
     payload: {orgSettings: settings},
-  }
-}
+  } as const)
 
-export const setFreeOrgSettings = (): SetOrgSettings => {
-  return {
-    type: ActionTypes.SetOrgSettings,
+export const setFreeOrgSettings = () =>
+  ({
+    type: SET_ORG_SETTINGS,
     payload: {orgSettings: [FREE_ORG_HIDE_UPGRADE_SETTING]},
-  }
-}
+  } as const)
 
 export const getOrgSettings = () => async (dispatch, getState: GetState) => {
   try {
     const org = getOrg(getState())
 
-    const result = await getOrgSettingsAJAX(org.id)
+    const response = await fetchOrgSettings(org.id)
+
+    if (response.status !== 200) {
+      throw new Error(
+        `Unable to get organization settings: ${response.statusText}`
+      )
+    }
+    const result = await response.json()
     dispatch(setOrgSettings(result.settings))
   } catch (error) {
     dispatch(setFreeOrgSettings())
--- a/ui/src/cloud/actions/orgsettings.ts
+++ b/ui/src/cloud/actions/orgsettings.ts
@@ -1,0 +1,48 @@
+// API
+import {getOrgSettings as getOrgSettingsAJAX} from 'src/cloud/apis/orgsettings'
+
+// Constants
+import {FREE_ORG_HIDE_UPGRADE_SETTING} from 'src/cloud/constants'
+
+// Types
+import {GetState, OrgSetting} from 'src/types'
+
+// Selectors
+import {getOrg} from 'src/organizations/selectors'
+
+export enum ActionTypes {
+  SetOrgSettings = 'SET_ORG_SETTINGS',
+}
+
+export type Actions = SetOrgSettings
+
+export interface SetOrgSettings {
+  type: ActionTypes.SetOrgSettings
+  payload: {orgSettings: OrgSetting[]}
+}
+
+export const setOrgSettings = (settings: OrgSetting[]): SetOrgSettings => {
+  return {
+    type: ActionTypes.SetOrgSettings,
+    payload: {orgSettings: settings},
+  }
+}
+
+export const setFreeOrgSettings = (): SetOrgSettings => {
+  return {
+    type: ActionTypes.SetOrgSettings,
+    payload: {orgSettings: [FREE_ORG_HIDE_UPGRADE_SETTING]},
+  }
+}
+
+export const getOrgSettings = () => async (dispatch, getState: GetState) => {
+  try {
+    const org = getOrg(getState())
+
+    const result = await getOrgSettingsAJAX(org.id)
+    dispatch(setOrgSettings(result.settings))
+  } catch (error) {
+    dispatch(setFreeOrgSettings())
+    console.error(error)
+  }
+}
--- a/ui/src/cloud/actions/orgsettings.ts
+++ b/ui/src/cloud/actions/orgsettings.ts
@@ -21,7 +21,7 @@ export interface SetOrgSettings {
   payload: {orgSettings: OrgSetting[]}
 }
 
-export const setOrgSettings = (settings: OrgSetting[]): SetOrgSettings => {
+export const setOrgSettings = (settings: OrgSetting[] = []): SetOrgSettings => {
   return {
     type: ActionTypes.SetOrgSettings,
     payload: {orgSettings: settings},
--- a/ui/src/cloud/apis/orgsettings.ts
+++ b/ui/src/cloud/apis/orgsettings.ts
@@ -1,13 +1,12 @@
-import AJAX from 'src/utils/ajax'
 import {OrgSettings} from 'src/types'
+import {getAPIBasepath} from 'src/utils/basepath'
 
 export const getOrgSettings = async (orgID: string): Promise<OrgSettings> => {
   try {
-    const {data} = await AJAX({
-      method: 'GET',
-      url: `/api/v2private/orgs/${orgID}/settings`,
-    })
-
+    const response = await fetch(
+      `${getAPIBasepath()}/api/v2private/orgs/${orgID}/settings`
+    )
+    const data = await response.json()
     return data
   } catch (error) {
     console.error(error)
--- a/ui/src/cloud/apis/orgsettings.ts
+++ b/ui/src/cloud/apis/orgsettings.ts
@@ -1,15 +1,7 @@
-import {OrgSettings} from 'src/types'
 import {getAPIBasepath} from 'src/utils/basepath'
+import {OrgSettingsResponse} from 'src/types'
 
-export const getOrgSettings = async (orgID: string): Promise<OrgSettings> => {
-  try {
-    const response = await fetch(
-      `${getAPIBasepath()}/api/v2private/orgs/${orgID}/settings`
-    )
-    const data = await response.json()
-    return data
-  } catch (error) {
-    console.error(error)
-    throw error
-  }
-}
+export const fetchOrgSettings = async (
+  orgID: string
+): Promise<OrgSettingsResponse> =>
+  await fetch(`${getAPIBasepath()}/api/v2private/orgs/${orgID}/settings`)
--- a/ui/src/cloud/apis/orgsettings.ts
+++ b/ui/src/cloud/apis/orgsettings.ts
@@ -1,0 +1,16 @@
+import AJAX from 'src/utils/ajax'
+import {OrgSettings} from 'src/types'
+
+export const getOrgSettings = async (orgID: string): Promise<OrgSettings> => {
+  try {
+    const {data} = await AJAX({
+      method: 'GET',
+      url: `/api/v2private/orgs/${orgID}/settings`,
+    })
+
+    return data
+  } catch (error) {
+    console.error(error)
+    throw error
+  }
+}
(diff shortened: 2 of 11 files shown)


ui/src/cloud/apis/orgsettings.ts
```diff
@@ -0,0 +1,16 @@
+import AJAX from 'src/utils/ajax'
+import {OrgSettings} from 'src/types'
+
+export const getOrgSettings = async (orgID: string): Promise<OrgSettings> => {
+  try {
+    const {data} = await AJAX({
```
drdelambre, line 6: please use `fetch` instead of AJAX

TCL735: Here AJAX is just an alias for axios. Is that okay to use or are we moving away from axios as well?

TCL735: 🎉

ui/src/pageLayout/containers/TreeNav.tsx
```diff
@@ -67,132 +74,134 @@ class TreeSidebar extends PureComponent<Props> {
...
-              {Boolean(item.menu) && (
-                <TreeNav.SubMenu>
-                  {item.menu.map(menuItem => {
-                    const linkElement = (className: string): JSX.Element => {
-                      if (menuItem.link.type === 'href') {
                         return (
-                          <a
-                            href={menuItem.link.location}
+                          <Link
+                            to={menuItem.link.location}
```
drdelambre, line 122: are any of these links to external resources? we've been having issues with the Link component giving us 404s lately

TCL735: I actually didn't change anything here is except wrap the whole thing inside another React component. The spacing was updated and the diff makes it look like I might have changed an anchor into a Link. But rest assured, there was no change here.

ui/src/cloud/components/OrgSettings.tsx
```diff
@@ -0,0 +1,34 @@
...
+// Actions
+import {getOrgSettings as getOrgSettingsAction} from 'src/cloud/actions/orgsettings'
+
+interface DispatchProps {
+  getOrgSettings: typeof getOrgSettingsAction
+}
+
+class OrgSettings extends PureComponent<DispatchProps> {
+  public componentDidMount() {
+    if (CLOUD) {
```
hoorayimhelping, line 17: good defensive coding here. is this check necessary? not sure how components under the `/cloud` directory work.

edit: ah, i see lower down that this component is used in OSS

ui/src/cloud/reducers/orgsettings.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export const orgSettingsReducer = (
+  state = defaultState,
+  action: Actions
+): OrgSettingsState =>
+  produce(state, draftState => {
+    if (action.type === ActionTypes.SetOrgSettings) {
+      const {orgSettings} = action.payload
+      draftState.settings = orgSettings
```
hoorayimhelping, line 23: stylistic nit: why not compact this to a single line? grepping for `orgSettings` will still find this line, so no information or context is lost. 

```suggestion
      draftState.settings = action.payload.orgSettings
```

ui/src/shared/components/CloudUpgradeButton.tsx
```diff
@@ -1,25 +1,56 @@
...
+  HIDE_UPGRADE_CTA_KEY,
+  PAID_ORG_HIDE_UPGRADE_SETTING,
+} from 'src/cloud/constants'
 
-const CloudUpgradeButton: FC = () => {
+// Types
+import {AppState, OrgSetting} from 'src/types'
+
+interface StateProps {
+  show: boolean
```
hoorayimhelping, line 20: this is past style nit and bordering on a blocking suggestion. `show` reads like a verb: `show()`. it is not immediately clear what it represents and leaves some ambiguity about its intent.
```suggestion
  shouldShow: boolean
```

ui/src/shared/components/CloudUpgradeButton.tsx
```diff
@@ -1,25 +1,56 @@
...
   )
 }
 
-export default CloudUpgradeButton
+const mstp = ({
+  cloud: {
+    orgSettings: {settings},
+  },
+}: AppState) => {
+  const hideUpgradeCTA = settings.find(
```
hoorayimhelping, line 44: `hideUpgradeCTA` is this an object? like an object of settings or a boolean? the name kind of reads like a function. i think thinking about the name of this might make the intent a little clearer.

ui/src/cloud/actions/orgsettings.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export type Actions = SetOrgSettings
+
+export interface SetOrgSettings {
+  type: ActionTypes.SetOrgSettings
+  payload: {orgSettings: OrgSetting[]}
+}
+
+export const setOrgSettings = (settings: OrgSetting[] = []): SetOrgSettings => {
```
drdelambre, line 24: in an attempt to have actions be a little less verbose, we're trying to remove the `ActionTypes + Actions + interface + action` paradigm and replace it with the `ReturnType<typeof> + as const` pattern you can find in the resource normalization work:
https://github.com/influxdata/influxdb/blob/master/ui/src/dashboards/actions/creators.ts#L29

ui/src/cloud/actions/orgsettings.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export const setFreeOrgSettings = (): SetOrgSettings => {
+  return {
+    type: ActionTypes.SetOrgSettings,
+    payload: {orgSettings: [FREE_ORG_HIDE_UPGRADE_SETTING]},
+  }
+}
+
+export const getOrgSettings = () => async (dispatch, getState: GetState) => {
```
drdelambre, line 38: there's also a `creators` vs `thunks` concept to separate out direct reducer interaction and more complex "action of actions" patterns, but i think it's a bit overkill for this action set, just wanted to let you know about it
https://github.com/influxdata/influxdb/blob/master/ui/src/dashboards/actions/thunks.ts

ui/src/cloud/apis/orgsettings.ts
```diff
@@ -0,0 +1,15 @@
+import {OrgSettings} from 'src/types'
+import {getAPIBasepath} from 'src/utils/basepath'
+
+export const getOrgSettings = async (orgID: string): Promise<OrgSettings> => {
+  try {
+    const response = await fetch(
+      `${getAPIBasepath()}/api/v2private/orgs/${orgID}/settings`
+    )
+    const data = await response.json()
```
drdelambre, line 9: shouldn't we be looking for a status code here as well, just in case an error message comes back as valid json?

ui/src/pageLayout/containers/TreeNav.tsx
```diff
@@ -67,132 +74,134 @@ class TreeSidebar extends PureComponent<Props> {
...
-          const linkElement = (className: string): JSX.Element => {
-            if (item.link.type === 'href') {
-              return <a href={item.link.location} className={className} />
+      <OrgSettings>
+        <TreeNav
+          expanded={isExpanded}
+          headerElement={<NavHeader link={orgPrefix} />}
+          userElement={<UserWidget />}
+          onToggleClick={handleToggleNavExpansion}
+          bannerElement={showUpgradeButton ? <CloudUpgradeNavBanner /> : null}
```
drdelambre, line 83: why add the concept of "showUpgradeButton" here, and not in the `CloudUpgradeNavBanner` component? from a quick read, it seems like it would unnecessarily couple the two components when `CloudUpgradeNavButton` could just return null

ui/src/pageLayout/containers/TreeNav.tsx
drdelambre: oh, this is a weird diff log. DISREGARD

ui/src/shared/components/CloudUpgradeButton.tsx
```diff
@@ -1,25 +1,56 @@
...
   return (
     <CloudOnly>
-      <Link
-        className="cf-button cf-button-sm cf-button-success upgrade-payg--button"
-        to={`${CLOUD_URL}${CLOUD_CHECKOUT_PATH}`}
-        target="_self"
-      >
-        Upgrade Now
-      </Link>
+      {show ? (
```
drdelambre, line 26: seems like it's _already_ taking care of it's own display state. if you move the OrgSettings logic in here with a useEffect to launch the action on mounting, you can remove all of the code changes in the TreeNav component

ui/src/cloud/apis/orgsettings.ts
```diff
@@ -0,0 +1,15 @@
...
+    return data
+  } catch (error) {
+    console.error(error)
```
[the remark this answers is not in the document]
drdelambre: i like seeing errors when i'm developing for cloud or debugging under load within the tools cluster. Occasionally we get the privilege of people allowing us to debug as their user within prod cloud environments (during an EAR for example) which is where these messages really shine.

I think there'd be a low ROI for the effort of wrapping console to hide logs in production builds if that's where this conversation is going, but in a lot of cases we _aren't_ getting alerted with errors, and `console.error` is the only way to surface "degraded states" (there is an error, but it shouldn't stop execution, but you don't see the changes you wanted). Our thunks being a place that pops up in my mind where we have a lot of `catch error -> notify with an unhelpful message -> drop it on the floor` going on.

Is there any evidence of these messages having a negative impact on our users? I'm running with the assumption that these errors have a lot of benefits for us (even if used sparingly), and don't have any effect on user conversion or cloud unit consumption outside of the error itself, but it'd be good to base the conversation on that data if we could.

ui/src/cloud/apis/orgsettings.ts
hoorayimhelping: >please use fetch instead of AJAX
why are we moving away from axios to fetch?  do people prefer the fetch api? i tend to find it tedious and verbose to send requests and handling responses is a bit confusing and prone to be buggy, especially if some responses send text and some send json. was there ever a discussion about this?

TCL735: My guess is that axios is a dependency while fetch is built into all the major browsers. And also, that we want to put the nail into the coffin of IE users.

ui/src/shared/components/CloudUpgradeButton.tsx
TCL735: The OrgSettings component is only responsible for the action that makes the call for the org settings from IDPE. Ideally, this should be done just once.

There are two separate components, CloudUpgradeButton and CloudUpgradeNavBanner, that prompt the user to "Upgrade Now". Each of these components should **not** be making their own identical calls to IDPE for the same resource. They should just receive props from state for that resource.

That's why the logic is separated. <OrgSettings> is really just a wrapper to fetch the resource. Fetching the resource just once, and having a component (TreeNav) that is reasonably expected to appear only once on the page seemed like an ideal marriage.

And then all of the child components on the page are supported with the resource.

ui/src/pageLayout/containers/TreeNav.tsx
TCL735: Originally, I had the show logic inside the ```CloudUpgradeNavBanner``` but I had a tough time getting it to work because of the React Fragment in there. I will make another attempt to move it back in there.

ui/src/cloud/apis/orgsettings.ts
TCL735: Data is data. As long as what we are looking for is there, why not? 😄 

But you make a a good point. I will improve the handling.

ui/src/cloud/reducers/orgsettings.ts
TCL735: i put that in there to make sure you were paying attention 😁

ui/src/cloud/actions/orgsettings.ts
TCL735: Good catch, thanks. But we are just doing the `ReturnType<typeof>` part only. The normalization is a bit out of scope, because _all_ cloud resources should be normalized together and that is a bigger and separate effort.

ui/src/cloud/actions/orgsettings.ts
TCL735: Noted.

ui/src/cloud/apis/orgsettings.ts
TCL735: We'll go with status 200 as the only valid status. Any other status, including redirects would result in an "error" and we will shout at you to **_OMG UPGRADE NOW_** with shiny buttons.

ui/src/pageLayout/containers/TreeNav.tsx
TCL735: Ok I got it working this time.

`TreeSideBar` is still responsible for fetching the settings. Since we can reasonably expect that there will never be more than one `TreeSideBar`, we can also reasonably expect that we will never make unnecessary fetches for the same resource.

`CloudUpgradeButton` and `CloudUpgradeNavBanner` can hide themselves based on the state props passed in.

ui/src/shared/components/CloudUpgradeButton.tsx
TCL735: `shouldShow` sounds too much like sh!tShow 🤣 

How about `inView`

hoorayimhelping: sounds good to me! you crack me up Timmy!

ui/src/cloud/apis/orgsettings.ts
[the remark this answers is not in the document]
hoorayimhelping: i don't think anyone cares or really notices these errors - maybe software engineers using cloud do, sometimes. but if they're helpful to you then let's do it! i'm just trying to guard against cargo culting in things

TCL735:
> My main question is: what's the intent behind making this a component that wraps other components, rather than just calling the function that the `OrgSettings` component wraps? What problems are you trying to solve there that using a function can't solve?

@hoorayimhelping It makes it easy to tell which components are affected by organization settings. I would rather ready JSX than ComponentDidMount or useEffect. You can just spot it in the render method.

If we had just used the underlying function in `TreeSideNav`, it would not be clear why `TreeSideNav` needs to run some side effect at certain points in the lifecycle that updates a state that `TreeSideNav` does not even care about.

Instead, as a separate component the code reads like this: _when `TreeSideNav` renders, its rendered elements are under the control of `OrgSettings`_.

Simple. So much more visible and easy to reason about.

TCL735:
> awesome! stop digging you struck oil! concise, to the point, makes perfect sense. sorry if the question came off as demanding or put you on the defensive!

It was a very important question. I was ready to put the underlying function into `TreeSideNav` if I didn't have an answer for myself.

PR #17966, opened 2020-05-06
feat: notebooks wip

drdelambre:
current status:
Kapture 2020-05-06 at 16 07 03


ui/src/pageLayout/constants/navigationHierarchy.ts
```diff
@@ -121,6 +121,19 @@ export const generateNavItems = (orgID: string): NavItem[] => {
       },
       activeKeywords: ['data-explorer'],
     },
+    {
+      id: 'notebooks',
+      testID: 'nav-item-notebooks',
+      icon: IconFont.Erlenmeyer,
```
alexpaxton, line 127: Dat flask tho

ui/src/notebooks/components/Notebook.tsx
```diff
@@ -0,0 +1,44 @@
...
+import {NotebookProvider, NotebookContext} from 'src/notebooks/notebook.context'
+
+const NotebookHeader: FC = () => {
+  //const { id } = useContext(NotebookContext)
+  //const { context, updateContext } = useContext(TimeContext)
+
+  return (
+    <>
+      <h1>NOTEBOOKS</h1>
+    </>
```
alexpaxton, line 15: I think you can just do `<Page.Title title="Notebooks" />`

ui/src/pageLayout/constants/navigationHierarchy.ts
drdelambre: ...so good...

drdelambre:
wanted to see how we can reuse resources within the redux state
Kapture 2020-05-07 at 15 32 46

drdelambre:
rough query on query action. feels like we have to tell the user what cell to reference and introduce them to the concept of linking somehow

Kapture 2020-05-07 at 16 55 53

drdelambre:
making an interface that's a little lighter weight, now you can extend queries across visualizations with a global `__PREVIOUS_RESULT__` flux variable. this gets reduced into a collection of static queries that can be run in parallel and then have their results attached back to their original visualization / stage.


Kapture 2020-05-12 at 14 21 53

PR #17968, opened 2020-05-06
docs(clientLibs): improve client docs

bednar:
* `Initialize the Client` always contains full range of required imports
* `Initialize the Client` contains `org`, `bucket` and `token` constant for reuse on all places
* Timestamps are always current time => user view written data in default settings of DataExplorer

### C#
c#

### GO
Screenshot 2020-05-06 at 12 29 27

### Java
Screenshot 2020-05-06 at 12 30 10

### JavaScript/Node.js
Screenshot 2020-05-06 at 12 30 54

### PHP
Screenshot 2020-05-06 at 12 31 27

### Python
Screenshot 2020-05-06 at 12 31 55

### Ruby
Screenshot 2020-05-06 at 12 32 46

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/clientLibraries/components/ClientJavaOverlay.tsx
+++ b/ui/src/clientLibraries/components/ClientJavaOverlay.tsx
@@ -58,11 +58,14 @@ const ClientJavaOverlay: FunctionComponent<Props> = props => {
         template={initializeClientCodeSnippet}
         label="Java Code"
         defaults={{
-          server: 'serverUrl',
+          server: 'basepath',
           token: 'token',
+          org: 'orgID',
+          bucket: 'bucketID',
         }}
         values={{
           server,
+          org,
         }}
       />
       <h5>Write Data</h5>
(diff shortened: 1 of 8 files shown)


ui/src/clientLibraries/components/ClientJavaOverlay.tsx
```diff
@@ -58,62 +58,37 @@ const ClientJavaOverlay: FunctionComponent<Props> = props => {
         template={initializeClientCodeSnippet}
         label="Java Code"
         defaults={{
-          server: 'serverUrl',
+          server: 'basepath',
```
russorat, line 61: is there a way to put the actual URL in here? similar to how we do here? https://github.com/influxdata/influxdb/blob/master/ui/src/clientLibraries/components/ClientLibraries.tsx#L35

russorat: nvm i think it does that already...

PR #17970, opened 2020-05-06
feat(http): Allow specific API handlers to be proxied for the Algo-W rollout

brettbuddin:
This introduces the ability to specify a feature flag driven proxy for specific HTTP handlers known to perform Flux `options` parsing. This change should be seen as a temporary measure to ensure the safe rollout of the Algo-W changes to the API. 

In `influxd`, a no-op structure (`http.NoopProxyHandler`) is used which does not do proxying. The Algo-W branch in our rollout will use this same no-op structure to force the use of in-process logic. The `feature.HTTPProxy` will be used in other contexts to direct the traffic to another endpoint.

The handlers that are targeted in this change are:

- Checks
  - `POST /api/v2/checks`
  - `PUT /api/v2/checks/:id`
  - `PATCH /api/v2/checks/:id`
- Tasks
  - `POST /api/v2/tasks`
  - `PATCH /api/v2/tasks/:id`
- Notification Rules
  - `POST /api/v2/notificationRules`
  - `PUT /api/v2/notificationRules/:id`
  - `PATCH /api/v2/notificationRules/:id`

Open questions:

- Are there any more API endpoints that need to be included in the proxying behavior?

### DIFF
--- a/kit/feature/http_proxy.go
+++ b/kit/feature/http_proxy.go
@@ -1,0 +1,65 @@
+package feature
+
+import (
+	"context"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+
+	"go.uber.org/zap"
+)
+
+// HTTPProxy is an HTTP proxy that's guided by a feature flag. If the feature flag
+// presented to it is enabled, it will perform the proxying behavior. Otherwise
+// it will be a no-op.
+type HTTPProxy struct {
+	proxy   *httputil.ReverseProxy
+	logger  *zap.Logger
+	enabler ProxyEnabler
+}
+
+// NewHTTPProxy returns a new Proxy.
+func NewHTTPProxy(dest *url.URL, logger *zap.Logger, enabler ProxyEnabler) *HTTPProxy {
+	return &HTTPProxy{
+		proxy:   newReverseProxy(dest, enabler.Key()),
+		logger:  logger,
+		enabler: enabler,
+	}
+}
+
+// Do performs the proxying. It returns whether or not the request was proxied.
+func (p *HTTPProxy) Do(w http.ResponseWriter, r *http.Request) bool {
+	if p.enabler.Enabled(r.Context()) {
+		p.proxy.ServeHTTP(w, r)
+		return true
+	}
+	return false
+}
+
+const (
+	// headerProxyFlag is the HTTP header for enriching the request and response
+	// with the feature flag key that precipitated the proxying behavior.
+	headerProxyFlag = "X-Proxy-Flag"
+)
+
+// newReverseProxy creates a new single-host reverse proxy.
+func newReverseProxy(dest *url.URL, enablerKey string) *httputil.ReverseProxy {
+	proxy := httputil.NewSingleHostReverseProxy(dest)
+
+	defaultDirector := proxy.Director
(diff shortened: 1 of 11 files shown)


kit/feature/http_proxy.go
```diff
@@ -0,0 +1,65 @@
...
+func NewHTTPProxy(dest *url.URL, logger *zap.Logger, enabler ProxyEnabler) *HTTPProxy {
+	return &HTTPProxy{
+		proxy:   newReverseProxy(dest, enabler.Key()),
+		logger:  logger,
+		enabler: enabler,
+	}
+}
+
+// Do performs the proxying. It returns whether or not the request was proxied.
+func (p *HTTPProxy) Do(w http.ResponseWriter, r *http.Request) bool {
```
gavincabbage, line 31: Did you specifically name this Do to avoid implementing http.Handler by naming it ServeHTTP? Just curious because that's what I would have done but there's really no reason to and I'm wondering if I'm missing any other tradeoffs.

kit/feature/http_proxy.go
```diff
@@ -0,0 +1,65 @@
...
+		p.proxy.ServeHTTP(w, r)
+		return true
+	}
+	return false
+}
+
+const (
+	// headerProxyFlag is the HTTP header for enriching the request and response
+	// with the feature flag key that precipitated the proxying behavior.
+	headerProxyFlag = "X-Proxy-Flag"
```
gavincabbage, line 42: I recently learned "X-" prefixed headers are discouraged as of 2012: https://stackoverflow.com/questions/3561381/custom-http-headers-naming-conventions

That said, the reasoning is that it makes them hard to promote to public standard later, but that doesn't apply in our case so I think it's fine as is.

I really like that you added this header to the response as well.

kit/feature/http_proxy.go
brettbuddin: Yeah, I wanted to be able to express in the contract that the operation might be a no-op. It might not perform the proxying behavior based on the conditions its testing for. To do that I needed a return value. Once I had that, I figured there was no reason to stick with the `ServeHTTP` nomenclature here. This is a loosely held opinion though.

gavincabbage: Oh duh yeah the return 👍

kit/feature/http_proxy.go
stuartcarnie: There is some precedent for using `X-<vendor>-<header-name>`.

We have used `X-INFLUXDB-` and `X-INFLUX` in the past, though those are now deprecated in favor of `X-Platform-`:

https://github.com/influxdata/influxdb/blob/67307d804ede329c5f7a72933ec10680635a7deb/kit/transport/http/api.go#L16

I'd recommend adding that to be consistent.

stuartcarnie: There is some precedent for using `X-<vendor>-<header-name>`.

We have used `X-INFLUXDB-` and `X-INFLUX` in the past, though those are now deprecated in favor of `X-Platform-`:

https://github.com/influxdata/influxdb/blob/67307d804ede329c5f7a72933ec10680635a7deb/kit/transport/http/api.go#L16

I'd recommend adding that to be consistent.

PR #17977, opened 2020-05-06
fix(storage/flux): fix a race condition in the tags cache

jsternberg:
The tags cache was not thread safe when called from multiple goroutines
at the same time. It was intended that it would be, but the locking was
done incorrectly and in too complicated a way. There was an assumption
that the LRU would only be updated from a single thread which wasn't
true at all.

The tags cache has now been updated to include some test cases that test
for race conditions and data validity. The tags cache itself has been
changed to follow a simpler algorithm.

1. Obtain a read lock.
2. Check if the cached array can be used.
3. Release the read lock.
4. If the above was unusable or did not exist, create an array for the
   tag.
5. Obtain a write lock.
6. Check if the cached array should be replaced and replace if needed.
7. Move the entry to the front of the LRU.
8. Release the write lock.

This simpler algorithm should ensure that this code is correct and that
creating the array is still done outside of the lock since creating the
array is the most expensive operation of the ones above.

### DIFF
--- a/storage/flux/tags_cache.go
+++ b/storage/flux/tags_cache.go
@@ -62,48 +62,80 @@ func (c *tagsCache) GetBounds(b execute.Bounds, l int, mem memory.Allocator) (st
 		return start, stop
 	}
 
-	if c.startColumn != nil {
-		start = c.getOrReplaceBounds(&c.startColumn, b.Start, l, mem)
-	} else {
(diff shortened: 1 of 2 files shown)


jlapacik:
Can you explain how this fixes the panic that was happening when a `yield` was included in a task?

PR #17992, opened 2020-05-07
feat(custom-timerange-tz): timeZone dropdown is now integrated in setting custom time ranges for queries

### LINKED ISSUE #17877: The UTC / Local dropdown does not work
__Steps to reproduce:__

```
```

2.  search for it in flux
```
from(bucket: "test")
```

__Actual behavior:__
I find it

__expected behavior:__
I find it as before

__Actual behavior:__
Not getting any result

Image 27-04-2020 at 15 22

__Actual behavior:__
Image 27-04-2020 at 15 25

__Environment info:__

Asalem1:
Closes #17877

### Purpose

The primary purpose of this PR is to ensure that whenever a user sets UTC as their timeZone and is querying a custom timeRange, their query will query data that was set at that set time in UTC (not the adjusted locale -> UTC time). 

### Problem

The UI for the timeZone (local or UTC) was often confusing and seemed to indicate functionality beyond displaying graphs in UTC.

### Solution

Created a selector to update the timeRange whenever a query is executed with a custom timeRange and the timeZone set to UTC. 

- [x] Rebased/mergeable
- [x] Tests pass


ui/src/dashboards/selectors/index.ts
```diff
@@ -15,6 +15,47 @@ export const getTimeRange = (state: AppState): TimeRange => {
...
+  }
+  if (offset < 0) {
+    // add tz minute difference
+    offsetDate = moment
+      .utc(date)
+      .add(offset, 'minutes')
+      .format()
+  }
+  return offsetDate
+}
```
hoorayimhelping, line 53:
```suggestion
export const setTimeToUTC = (date: string): string => {
  const offset = new Date(date).getTimezoneOffset()
  if (offset > 0) {
    // subtract tz minute difference
    return moment
      .utc(date)
      .subtract(offset, 'minutes')
      .format()
  }
  if (offset < 0) {
    // add tz minute difference
   return moment
      .utc(date)
      .add(offset, 'minutes')
      .format()
  }
  // offset === 0
  return date
}
```

this is your code, but with a little bit further editing - the key was realizing that

```ts
if (offset === 0) {
  return date
}
let offsetDate = date

// if (offset > 0) ...
// if (offset < 0) ...

return offsetDate
```
could be turned into `return date` if it was moved below the `if` statements. the reason this change is worth making is that it removes one branch in an `if` statement, and removes the creation of a variable, and removes the overwriting of that variable and replaces them with a single return statement of the argument passed in. i've found that returning values directly, rather than returning them into a variable and returning makes the logic of function much easier to follow.

ui/src/dashboards/selectors/index.ts
```diff
@@ -15,6 +15,47 @@ export const getTimeRange = (state: AppState): TimeRange => {
...
+  const newTimeRange = Object.assign({}, timeRange)
+  if (timeRange.type === 'custom' && timeZone === 'UTC') {
+    // check to see if the timeRange has an offset
+    newTimeRange.lower = setTimeToUTC(newTimeRange.lower)
+    newTimeRange.upper = setTimeToUTC(newTimeRange.upper)
+  }
+  return newTimeRange
+}
+
+export const setTimeToUTC = (date: string): string => {
```
hoorayimhelping, line 31: can you add a comment here and a link to this issue? just to give a little more context into what this function is used for?

ui/src/dashboards/selectors/index.ts
```diff
@@ -15,6 +15,47 @@ export const getTimeRange = (state: AppState): TimeRange => {
   return state.ranges[contextID] || DEFAULT_TIME_RANGE
 }
 
+export const getTimeRangeWithTimezone = (state: AppState): TimeRange => {
+  const timeRange = getTimeRange(state)
+  const timeZone = getTimeZone(state)
+  // create a copy of the timeRange so as not to mutate the original timeRange
+  const newTimeRange = Object.assign({}, timeRange)
```
hoorayimhelping, line 22:
```suggestion
  const newTimeRange = {...timeRange}
```
i *believe* we prefer es6 notation to `Object.assign`

you can leave the comment if you like, but i think seeing `Object.assign({}, foo)` or `{...foo}` generally implies immutable copying which makes the comment a little redundant.

ui/src/dashboards/selectors/index.ts
```diff
@@ -15,6 +15,47 @@ export const getTimeRange = (state: AppState): TimeRange => {
...
+  if (timeRange.type === 'custom' && timeZone === 'UTC') {
+    // check to see if the timeRange has an offset
```
hoorayimhelping, line 24: ✂️ 

doesn't seem to be what this line is doing. looks like it's conforming the date to utc with the proper offset, not checking if it has an offset.

ui/src/dashboards/selectors/index.test.ts
```diff
@@ -18,34 +22,85 @@ const untypedGetTimeRangeByDashboardID = getTimeRange as (a: {
...
+      lower: '2020-05-05T10:00:00Z',
+      upper: '2020-05-05T11:00:00Z',
+      type: 'custom',
+    }
+
+    expect(app.persisted.timeZone).toEqual('UTC')
+    expect(
+      untypedGetTimeRangeWithTimeZone({ranges, currentDashboard, app})
+    ).toEqual(expected)
+  })
```
hoorayimhelping, line 105: these tests look fine, and are wise in that they don't reference like `today` or `now` but instead solid dates in the past.

but having said that, a career of having tests around checking date/times that fail randomly has made me extremely wary and gunshy of them. just keep an eye on tests that involve dates and time. they can fail and blow up in weird and unexpected and painful ways.

ui/src/timeMachine/components/Vis.tsx
```diff
@@ -164,7 +164,7 @@ const mstp = (state: AppState): StateProps => {
   const fillColumns = getFillColumnsSelection(state)
   const symbolColumns = getSymbolColumnsSelection(state)
 
-  const timeZone = state.app.persisted.timeZone
+  const timeZone = getTimeZone(state)
```
hoorayimhelping, line 167: so the line this is replacing wouldn't set the timeZone to `local` if it `state.app.persisted.timeZone` didn't exist. this function will. is that okay for this component? i imagine it is, but just want to be explicit

ui/src/dashboards/selectors/index.ts
Asalem1: Great suggestion! I'll make sure to push it up with the rest of the changes and ping you when it's up

ui/src/dashboards/selectors/index.ts
Asalem1: definitely

ui/src/dashboards/selectors/index.ts
Asalem1: Great point!

ui/src/timeMachine/components/Vis.tsx
Asalem1: I also noticed that. It's weird because we set `local` as the default in only one place, but I think it's good to have a default for this value, especially if it's a required property. I'm very wary of assigning defaultProps, but in this case I think it's safe to assign a default even if none were specified

ui/src/dashboards/selectors/index.ts
```diff
@@ -15,6 +15,49 @@ export const getTimeRange = (state: AppState): TimeRange => {
...
+  // The purpose of this function is to set a user's custom time range selection
+  // from the local time to the same time in UTC if UTC is selected from the
+  // timezone dropdown. This is feature was original requested here:
+  // https://github.com/influxdata/influxdb/issues/17877
+  // Example: user selected 10-11:00am and sets the dropdown to UTC
+  // Query should run against 10-11:00am UTC rather than querying
+  // 10-11:00am local time (offset depending on timezone)
+  const offset = new Date(date).getTimezoneOffset()
+  if (offset > 0) {
+    // subtract tz minute difference
```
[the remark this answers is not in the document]
Asalem1: ;)

PR #18005, opened 2020-05-07
fix(query): Kill to detach query

ayang64:
Closes #17992 

Currently, killing a query leaves it in a "killed" state such that it is visible in the tasks list after being killed.

Sometimes tasks aren't properly removed from the list.  It seems like a better approach would be to kill and remove a task from the list when the `KILL QUERY` command is issued.

This patch wraps the `TaskManager.DetachDetach()` method with `query.Kill()` -- this way, when we call `TaskManager.KillQuery()`, it properly kills and detaches the task.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/query/task_manager.go
+++ b/query/task_manager.go
@@ -220,18 +220,15 @@ func (t *TaskManager) AttachQuery(q *influxql.Query, opt ExecutionOptions, inter
 	return ctx, func() { t.DetachQuery(qid) }, nil
 }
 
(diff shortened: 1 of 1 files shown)


ayang64:
i still need to:

* update the changelog
* request an update of the documentation if necessary

Asalem1:
Looks like this is accidentally marking a PR as closed rather than an issue. Would you mind updating that?

PR #18006, opened 2020-05-07
feat(demodata): Make dd error messages more explanatory

ebb-tide:


ebb-tide:
Yeah, I definitely smell something there as well. 
But I don't want to continue to add the dashboard if adding bucket has errored so I don't think two try/catch'es will work

ebb-tide:
that works! thanks.. :)

PR #18011, opened 2020-05-07
feat(custom-tz): added the ability to set UTC as the timezone when making custom time range queries

### LINKED ISSUE #17877: The UTC / Local dropdown does not work

Asalem1:
Closes #17877

### Purpose
The primary purpose of this PR is to ensure that whenever a user sets UTC as their timeZone and is querying a custom timeRange, their query will query data that was set at that set time in UTC (not the adjusted locale -> UTC time).

### Problem
The UI for the timeZone (local or UTC) was often confusing and seemed to indicate functionality beyond displaying graphs in UTC.

### Solution
Created a selector to update the timeRange whenever a query is executed with a custom timeRange and the timeZone set to UTC.

- [x] Rebased/mergeable

### DIFF
--- a/ui/src/dashboards/selectors/index.test.ts
+++ b/ui/src/dashboards/selectors/index.test.ts
@@ -3,6 +3,7 @@ import {
   getTimeRange,
   getTimeRangeWithTimezone,
 } from 'src/dashboards/selectors/index'
+import moment from 'moment'
 
 // Types
 import {RangeState} from 'src/dashboards/reducers/ranges'
@@ -34,9 +35,15 @@ describe('Dashboards.Selector', () => {
     '04c6f3976f4b8000',
     '04c6f3976f4b8002',
   ]
+  const lower = '2020-05-05T10:00:00-07:00'
+  const upper = '2020-05-05T11:00:00-07:00'
   const customTimeRange = {
-    lower: '2020-05-05T10:00:00-07:00',
-    upper: '2020-05-05T11:00:00-07:00',
+    lower: moment(lower)
+      .utcOffset(lower)
+      .format(),
+    upper: moment(upper)
+      .utcOffset(upper)
+      .format(),
     type: 'custom',
   } as CustomTimeRange
   const ranges: RangeState = {
--- a/ui/src/dashboards/selectors/index.test.ts
+++ b/ui/src/dashboards/selectors/index.test.ts
@@ -2,6 +2,7 @@
 import {
   getTimeRange,
   getTimeRangeWithTimezone,
+  setTimeToUTC,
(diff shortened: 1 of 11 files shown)


ui/src/dashboards/selectors/index.test.ts
```diff
@@ -18,34 +24,99 @@ const untypedGetTimeRangeByDashboardID = getTimeRange as (a: {
...
+    '04c6f3976f4b8001',
+    '04c6f3976f4b8000',
+    '04c6f3976f4b8002',
+  ]
+  const lower = '2020-05-05T10:00:00-07:00'
+  const upper = '2020-05-05T11:00:00-07:00'
+  const customTimeRange = {
+    lower: moment(lower).toISOString(),
+    upper: moment(upper).toISOString(),
+    type: 'custom',
```
Asalem1, line 44: by setting these values to `ISOString` more accurately reflects how the values are represented when passing them to the function

ui/src/dashboards/selectors/index.test.ts
```diff
@@ -18,34 +24,99 @@ const untypedGetTimeRangeByDashboardID = getTimeRange as (a: {
...
+        navBarState: 'expanded',
+        timeZone: 'UTC' as TimeZone,
+        theme: 'dark',
+      },
+    }
+
+    const expected = {
+      lower: setTimeToUTC(lower),
+      upper: setTimeToUTC(upper),
+      type: 'custom',
```
Asalem1, line 115: Rather than hardcoding the values as I was doing, I set the time more deterministically using the setTimeToUTC function.

hoorayimhelping: i'm not sure this test is actually testing much at this point. when you strip everything away it basically says

```ts
assertEquals(setTimeToUTC(lower), setTimeToUTC(lower))
```

since [this line](https://github.com/influxdata/influxdb/pull/18011/files#diff-be9be89f3c9a10c0102cf057f74e8104R25) is the code under test.

don't think we should remove it, but i do think you should spend some time thinking about ways you can get more confidence from this test.

Asalem1: I totally agree. The only other implementation of this that I could think of would be to calculate the UTCOffset locally and set a value dynamically through a switch statement with all potential values

PR #18027, opened 2020-05-11
feat(demodata): Place dd dashboard errors behind feature flag

ebb-tide:
In order to feel more comfortable releasing the demodata, we'd like to be able to switch the feature off in case of errore. We also want to give users a seamless experience even if demodata is switched off after they have added dd buckets, or dashboards. 
Before this PR- toggling the demodata feature flag off would 
- remove the dd dropdown, and 
- stop displaying dd buckets a user may have subscribed to.
But dd dashboards would still remain and generate a `failed to initialize execute state: could not find bucket` error from flux. 

after the changes in this PR, toggling the demodata feature flag will also:
- show a notification to the user that the "demodata feature is temporarily unavailable" if they already had dashboards set up to query the demodata bucket, and receive a `failed to initialize execute state: could not find bucket` error

### DIFF
--- a/ui/src/cloud/utils/demoDataErrors.ts
+++ b/ui/src/cloud/utils/demoDataErrors.ts
@@ -1,31 +1,17 @@
 import {WebsiteMonitoringBucket} from 'src/cloud/constants'
-import {reportError} from 'src/shared/utils/errors'
 import {CLOUD} from 'src/shared/constants'
 import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
-export const isDemoDataAvailabilityError = (errorMessage: string): boolean => {
+export const isDemoDataAvailabilityError = (
+  errorCode: string,
+  errorMessage: string
+): boolean => {
   if (!CLOUD || isFlagEnabled('demodata')) {
     return false
   }
-  const findBucketErrorMessage =
-    'failed to initialize execute state: could not find bucket /"'
 
-  if (errorMessage.startsWith(findBucketErrorMessage)) {
-    const missingBucket = errorMessage
-      .replace(findBucketErrorMessage, '')
-      .replace('/"', '')
-
-    if (missingBucket === WebsiteMonitoringBucket) {
-      reportError(
-        {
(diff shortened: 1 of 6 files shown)


ui/src/cloud/utils/demoDataErrors.ts
```diff
@@ -0,0 +1,33 @@
...
+  const findBucketErrorMessage =
+    'failed to initialize execute state: could not find bucket /"'
+
+  if (errorMessage.startsWith(findBucketErrorMessage)) {
+    const missingBucket = errorMessage
+      .replace(findBucketErrorMessage, '')
+      .replace('/"', '')
+
+    if (missingBucket === WebsiteMonitoringBucket) {
+      reportError(
```
ebb-tide, line 19: I'm on the fence about whether this is useful

ui/src/shared/copy/notifications.ts
```diff
@@ -476,6 +476,12 @@ export const demoDataSucceeded = (
   link,
 })
 
+export const demoDataSwitchedOff = (): Notification => ({
+  ...defaultErrorNotification,
+  message: `Demo data buckets are temporarily unavailable. Please try again later.`,
```
hoorayimhelping, line 481: is it appropriate to link to https://status.influxdata.com/ even though i know demo data isn't a service that's listed on there?

ui/src/shared/apis/query.ts
```diff
@@ -138,8 +139,9 @@ const processErrorResponse = async (
     const body = await response.text()
     const json = JSON.parse(body)
     const message = json.message || json.error
+    const code = json.code || ''
```
hoorayimhelping, line 142: should this fall back to `null` or `undefined` rather than `''`? trying to think of the meanings of all these things:

`undefined`: doesn't exist. the same as doing `const code = json.code;`
`null`: has been assigned a value of: `no value`
`''`: has been assigned a value of: `no value`, represented as a string

the api, if it returns anything, will return a string. if it doesn't return anything, the value of `code` will be `undefined`. an empty string _might_ imply that we got a response back from the API

PR #18041, opened 2020-05-11
fix: password length

### LINKED ISSUE #17642: [UI - Onboarding] Password length or well formed check no longer applies
__Steps to reproduce:__

1. Open the onboarding page
2. Add a user name
3. Input a short initial password, e.g. 'aa' and confirm it

__Expected behavior:__
Expected to be warned about using a short or easy to guess password.  In the past the minimum length was 8 chars. 

__Actual behavior:__
The continue button is highlighted and no warning is given. 

__Screenshot__
NoPasswordLengthCheck01

__Environment info:__

```
Linux 4.15.0-96-generic x86_64
```
Detected in nightly builds
```
version=nightly commit=bd91a81123 build_date=2020-04-07T07:57:22Z
```
Browser: __Chrome 80__

__Config:__
Started with end to end testing enabled. 
```
--e2e-testing=true
```

121watts:
Closes #17642 

### Frontend Validation
Screen Shot 2020-05-11 at 2 18 49 PM

### Notes
I'm not really a fan of how we handle form validation in the frontend at the moment.  Specifically for the password state.  This hacks together a working solution.  Mayhaps we should do something a bit nicer?

### More Notes
Apparently, we need to be able to onboard users with no password via Quartz...

I've removed the backend validation.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -250,8 +250,11 @@ github.com/influxdata/goreleaser v0.97.0-influx h1:jT5OrcW7WfS0e2QxfwmTBjhLvpIC9
 github.com/influxdata/goreleaser v0.97.0-influx/go.mod h1:MnjA0e0Uq6ISqjG1WxxMAl+3VS1QYjILSWVnMYDxasE=
 github.com/influxdata/httprouter v1.3.1-0.20191122104820-ee83e2772f69 h1:WQsmW0fXO4ZE/lFGIE84G6rIV5SJN3P3sjIXAP1a8eU=
 github.com/influxdata/httprouter v1.3.1-0.20191122104820-ee83e2772f69/go.mod h1:pwymjR6SrP3gD3pRj9RJwdl1j5s3doEEV8gS4X9qSzA=
+github.com/influxdata/influxdb v1.8.0 h1:/X+G+i3udzHVxpBMuXdPZcUbkIE0ouT+6U+CzQTsOys=
 github.com/influxdata/influxql v0.0.0-20180925231337-1cbfca8e56b6 h1:CFx+pP90q/qg3spoiZjf8donE4WpAdjeJfPOcoNqkWo=
 github.com/influxdata/influxql v0.0.0-20180925231337-1cbfca8e56b6/go.mod h1:KpVI7okXjK6PRi3Z5B+mtKZli+R1DnZgb3N+tzevNgo=
+github.com/influxdata/kapacitor v1.5.5 h1:0v34uNwVpeyfc9XZGWvFc3pzA2AajcSrdakUV1GCU2E=
+github.com/influxdata/kapacitor v1.5.5/go.mod h1:vv15yTwFBi1kUhCUrM/PGdfsbh5Y5F8BaCrdOg2S+d8=
 github.com/influxdata/line-protocol v0.0.0-20180522152040-32c6aa80de5e h1:/o3vQtpWJhvnIbXley4/jwzzqNeigJK9z+LZcJZ9zfM=
 github.com/influxdata/line-protocol v0.0.0-20180522152040-32c6aa80de5e/go.mod h1:4kt73NQhadE3daL3WhR5EJ/J2ocX0PZzwxQ0gXJ7oFE=
 github.com/influxdata/promql/v2 v2.12.0 h1:kXn3p0D7zPw16rOtfDR+wo6aaiH8tSMfhPwONTxrlEc=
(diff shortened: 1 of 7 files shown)


go.sum
```diff
@@ -250,8 +250,11 @@ github.com/influxdata/goreleaser v0.97.0-influx h1:jT5OrcW7WfS0e2QxfwmTBjhLvpIC9
 github.com/influxdata/goreleaser v0.97.0-influx/go.mod h1:MnjA0e0Uq6ISqjG1WxxMAl+3VS1QYjILSWVnMYDxasE=
 github.com/influxdata/httprouter v1.3.1-0.20191122104820-ee83e2772f69 h1:WQsmW0fXO4ZE/lFGIE84G6rIV5SJN3P3sjIXAP1a8eU=
 github.com/influxdata/httprouter v1.3.1-0.20191122104820-ee83e2772f69/go.mod h1:pwymjR6SrP3gD3pRj9RJwdl1j5s3doEEV8gS4X9qSzA=
+github.com/influxdata/influxdb v1.8.0 h1:/X+G+i3udzHVxpBMuXdPZcUbkIE0ouT+6U+CzQTsOys=
```
lyondhill, line 253: prolly a `go mod tidy` will clean this up

tenant/error.go
```diff
@@ -43,6 +43,11 @@ var (
 		Code: influxdb.ENotFound,
 		Msg:  "not found",
 	}
+
+	ErrPasswordTooShort = &influxdb.Error{
```
lyondhill, line 47: the user service in tenant/service_user.go uses an error defined in error_user.go called `EShortPassword`. You could either use that error or have the user service use this one, I don't care which.

tenant/service_onboarding.go
```diff
@@ -50,7 +50,7 @@ func (s *OnboardService) OnboardInitialUser(ctx context.Context, req *influxdb.O
 	return s.onboardUser(ctx, req, func(influxdb.ID) []influxdb.Permission { return influxdb.OperPermissions() })
 }
 
-// OnboardUser allows us to onboard a new user if is onboarding is allowd
+// OnboardUser allows us to onboard a new user if is onboarding is allowed
```
Asalem1, line 53: 👏

tenant/service_onboarding.go
```diff
@@ -61,6 +61,10 @@ func (s *OnboardService) onboardUser(ctx context.Context, req *influxdb.Onboardi
 		return nil, ErrOnboardInvalid
 	}
 
+	if len(req.Password) < 8 {
```
lyondhill, line 64: Could this be moved inside the `if req.Password != "" {` There is a circumstance we would want to onboard the a user and not provide a password for them.

121watts:
@Asalem1 it does come up as the user is typing.  So does the password mismatch error.  Like I said in the PR, I'm not a fan of it.  I'd like to change how we do password validation altogether but that's beyond the scope of this PR.

ui/src/onboarding/components/AdminStep.tsx
```diff
@@ -204,6 +200,14 @@ class AdminStep extends PureComponent<Props, State> {
     )
   }
 
+  private get passwordError(): string {
+    return this.state.isPassTooShort && 'Password must be at least 8 characters'
```
hoorayimhelping, line 204: i'm  confused by the intent of this code and what it actually is meant to return. this returns `false` or an error message, but it says it returns a string. I tested this in the console and I got this:

```js
const errorHandler = (arbitraryBool) => { return arbitraryBool && "error message" }

errorHandler(false) // false
errorHandler(true) //  error message
```

am I missing something? Is there a strong preference for doing this on one line instead of doing something like?

```ts
if (this.state.isPassTooShort) {
  return 'Password must be at least 8 characters'
}

return ''
```

ui/src/onboarding/components/AdminStep.tsx
```diff
@@ -221,7 +225,8 @@ class AdminStep extends PureComponent<Props, State> {
     const {confirmPassword} = this.state
     const password = e.target.value
     const isPassMismatched = confirmPassword && password !== confirmPassword
-    this.setState({password, isPassMismatched})
+    const isPassTooShort = password && password.length < 8
```
hoorayimhelping, line 228: what's the intent of checking for `password` here? is it to guard against empty strings triggering validation?

i noticed that there aren't any test cases for "" in the password e2e tests - are these things related or did i miss those?

121watts: that's correct.  i don't want to it to show up when the password is empty.

PR #18043, opened 2020-05-11
feat(query): implement aggregate pushdown

### LINKED ISSUE #6805: influxd binary needs better command-line parameter help
### Feature Request

__Proposal:__ [Description of the feature]

Calling `influxd` with an unrecognized switch should point users to the `influxd help` subcommand. 

Also, the current list of switches/flags is partially incorrect.

__Current behavior:__ [What currently happens]

```
# influxd -help
usage: run [flags]

run starts the InfluxDB server. If this is the first time running the command
then a new cluster will be initialized unless the -join argument is used.

        -config <path>
                          Set the path to the configuration file.

        -hostname <name>
                          Override the hostname, the 'hostname' configuration
                          option will be overridden.

        -pidfile <path>
                          Write process ID to a file.

        -cpuprofile <path>
                          Write CPU profiling information to a file.

        -memprofile <path>
                          Write memory usage information to a file.

run: flag: help requested
```

To get the subcommands you must run `influxd help`, no flag:

```
# influxd help
Configure and start an InfluxDB server.

Usage:

	influxd [[command] [arguments]]

The commands are:

    backup               downloads a snapshot of a data node and saves it to disk
    config               display the default configuration
    restore              uses a snapshot of a data node to rebuild a cluster
    run                  run node with existing configuration
    version              displays the InfluxDB version

"run" is the default command.

Use "influxd help [command]" for more information about a command.

```

__Desired behavior:__ [What you would like to happen]

The above switch output mentions `-join`, which in the post 0.12 world is likely deprecated or removed. It also doesn't mention the `help` command, which is the only way to find out about subcommands like `version` or `backup`.

__Use case:__ [Why is this important (helps with prioritizing requests)]

It's impossible to remember every switch and flag to every binary, and naive POSIX users like me rely on being able to get a reminder output from the binary with something like `influxd -?` or `influxd --help`. I get output from those that is incomplete, because instead I must run `influxd help`, which isn't mentioned in the seemingly complete output from `influxd -?`

yzhang1991:
This patch implements the window pushdown for aggregate counts.

Fixes #6805


query/stdlib/influxdata/influxdb/rules_test.go
```diff
@@ -1207,8 +1207,8 @@ func TestPushDownWindowAggregateRule(t *testing.T) {
 
 	tests := make([]plantest.RuleTestCase, 0)
 
-	// construct a simple plan with a specific window and aggregate function
```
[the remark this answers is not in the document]
rockstar: I'm not sure, really. We did that together, and I couldn't remember, so I left it. If you can't remember, I'm reverting it.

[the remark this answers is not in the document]
yzhang1991: I do not remember that at all.. Let's revert it?

storage/flux/table.gen.go
```diff
@@ -97,6 +97,65 @@ func (t *floatTable) advance() bool {
...
+	}
+
+	// Retrieve the buffer for the data to avoid allocating
+	// additional slices. If the buffer is still being used
+	// because the references were retained, then we will
+	// allocate a new buffer.
+	cr := t.allocateBuffer(l)
+	// regain the window start time from the window end time
+	rangeStart := int64(t.bounds.Start)
+	startTimes := make([]int64, len(a.Timestamps))
```
jsternberg, line 145: This isn't the optimal way to do this and it will result in another untracked allocation. We should create the start column using the arrow builders directly.

```go
startB := arrow.NewIntBuilder(t.alloc)
startB.Resize(len(a.Timestamps))
for i := 0; i < a.Len(); i++ {
    ts := a.Timestamps[i] - t.windowEvery
    if ts < int64(t.bounds.Start) {
        ts = int64(t.bounds.Start)
    }
    startB.Append(ts)
}
cr.cols[startColIdx] = startB.NewArray()
```

The reason we don't do the same for the stop or value columns is mostly because we get those arrays from protobuf so we don't control their allocations. Because the start column is something we create that doesn't exist on the wire, it will minimize copies and track allocations better if we use the arrow builder.

query/stdlib/influxdata/influxdb/rules_test.go
```diff
@@ -1207,8 +1207,8 @@ func TestPushDownWindowAggregateRule(t *testing.T) {
...
-	simplePlanWithWindowAgg := func( window universe.WindowProcedureSpec, agg plan.NodeID, spec plan.ProcedureSpec ) *plantest.PlanSpec {
+	// construct a simple plan with a specific windo construct a simple plan with a specific window and aggregate function
```
jsternberg, line 1210: Nitpick: `windo` -> `window`.

storage/flux/table.gen.go
```diff
@@ -288,6 +347,65 @@ func (t *integerTable) advance() bool {
...
+	for i := 0; i < a.Len(); i++ {
+		startTimes[i] = a.Timestamps[i] - t.windowEvery
+	}
+	if startTimes[0] < rangeStart {
+		startTimes[0] = rangeStart
+	}
+	cr.cols[startColIdx] = arrow.NewInt(startTimes, t.alloc)
```
jsternberg, line 402: I'm a bit confused by the code here. It seems to me like the rewrite rule is focused on only pushing down the pattern `window() |> aggregate()`, but the window table implementation seems to be more focused on `aggregateWindow()`. I think this is good because this loop is likely the best way to implement this. We probably don't want to group the table into arrays of size one for performance reasons.

I was expecting to see a piece of code that wrapped this section and use arrow slices to divide this table into multiple tables so, when `window() |> aggregate()` was seen, we would get back tables with the appropriate group keys. Does that idea make sense or am I mistaken?

storage/flux/table.gen.go
yzhang1991: That is a good point!

rockstar: > The current method will result in an untracked allocation.

This seems to a be a go-internals sort of knowledge point. Could you teach me to fish here? How did you identify that untracked allocation? What am I missing?

jsternberg: It's the `make()` call which is standard for Go, but our memory tracking system won't recognize it so the allocator won't prevent the allocation. When you use the arrow builders, it uses `make()` under the hood, but it track the amount of memory it has used to construct arrays. It's a very flux-specific thing that isn't part of Go itself.

storage/flux/table.gen.go
rockstar: > I was expecting to see a piece of code that wrapped this section and use arrow slices to divide this table into multiple tables so, when window() |> aggregate() was seen, we would get back tables with the appropriate group keys. Does that idea make sense or am I mistaken?

To some extent, yes. We were going to hoist that splitting of single table into multiple tables up to the flux level, and in my mind, that was a separate patch, because this one is already pretty busy. If you don't agree, we can come back and evaluate that.

jsternberg: That makes sense to me. I'm concerned about the `PushDownWindowAggregateRule` being included in this patch. Can we comment that back out? It makes sense to me for that to be part of a different PR, but I'm concerned that we'll accidentally set a feature flag to true when the code isn't feature complete.

yzhang1991: Eh... Actually I am thinking we should do that in the storage... To me it was a mistake I made in the WindowAggregateResultSet

rockstar: My understanding is that if we do it in storage, we send more bytes over the wire.

jsternberg: We should discuss that. I much prefer this implementation for storage because it enables us to assemble the result more efficiently on the client side for the various different arrays of `aggregateWindow`. It also dovetails better into a potential future design of aggregate that will be the same as the full `aggregateWindow`. The current design of window which allocates a lot of small tables performs pretty poorly so we want to optimize for the case where someone uses `aggregateWindow` rather than `window() |> aggregate()` specifically. This is also true because `readWindowAggregate() |> duplicate() |> window(every: inf)` will never be as fast as this version and that's going to be the most common query.

jsternberg: It's also true we'll send more bytes over the wire because we will send the series header after every row rather than sending it once for an entire series.

storage/reads/array_cursor.go
```diff
@@ -63,6 +76,43 @@ func newCountArrayCursor(cur cursors.Cursor) cursors.Cursor {
...
+			tr:                req.Range,
+		}
+	case cursors.BooleanArrayCursor:
+		return &integerBooleanWindowCountArrayCursor{
+			BooleanArrayCursor: cur,
+			every:              req.WindowEvery,
+			tr:                 req.Range,
+		}
+	default:
+		panic(fmt.Sprintf("unreachable: %T", cur))
```
fchikwekwe, line 112: Is this default really unreachable? Is a default case required for a switch?

storage/reads/array_cursor.go
```diff
@@ -22,10 +22,23 @@ func newAggregateArrayCursor(ctx context.Context, agg *datatypes.Aggregate, curs
...
+
+	switch req.Aggregate[0].Type {
 	case datatypes.AggregateTypeSum:
 		return newSumArrayCursor(cursor)
 	case datatypes.AggregateTypeCount:
-		return newCountArrayCursor(cursor)
+		return newWindowCountArrayCursor(cursor, req)
 	default:
 		// TODO(sgc): should be validated higher up
 		panic("invalid aggregate")
```
fchikwekwe, line 44: Is is worth creating an issue to validate this sooner? Could this introduce a bug?

rockstar: I don't _believe_ so. I think we saw that it _is_ validated higher up. This could probably just as well be a "unreachable" sort of panic.

storage/reads/array_cursor.go
rockstar: It's more like an assertion that our _assumption_ that it's unreachable holds. If we get that far, we certainly have a bug, and now we have a stacktrace to show us how we got there.

storage/reads/array_cursor.gen.go.tmpl
```diff
@@ -235,6 +236,61 @@ func (c *integer{{.Name}}CountArrayCursor) Next() *cursors.IntegerArray {
...
 
+type integer{{.Name}}WindowCountArrayCursor struct {
+	cursors.{{.Name}}ArrayCursor
+	every int64
+	tr    datatypes.TimestampRange
+}
+
+func (c *integer{{.Name}}WindowCountArrayCursor) Stats() cursors.CursorStats {
+	return c.{{.Name}}ArrayCursor.Stats()
+}
```
fchikwekwe, line 247: I'm just curious because its not super clear to me, what are these `CursorStats` tracking here?

storage/flux/reader.go
```diff
@@ -395,54 +395,32 @@ func convertGroupMode(m query.GroupMode) datatypes.ReadGroupRequest_Group {
...
 const (
-	startColIdx = 0
-	stopColIdx  = 1
-	timeColIdx  = 2
-	valueColIdx = 3
+	startColIdx         = 0
+	stopColIdx          = 1
+	timeColIdx          = 2
+	windowedValueColIdx = 2
+	valueColIdx         = 3
```
fchikwekwe, line 402: I'm not seeing where the `valueColIdx` get used now. Is there a purpose for it or is it just being left for future changes?

storage/reads/array_cursor.gen.go.tmpl
rockstar: `CursorStats` tracks the number of values and the number of bytes read. I'm not sure how it's used at a higher level.

storage/flux/table.gen.go
```diff
@@ -97,6 +97,76 @@ func (t *floatTable) advance() bool {
...
+			return false
+		}
+		t.idxInArr = 0
+	}
+
+	// Retrieve the buffer for the data to avoid allocating
+	// additional slices. If the buffer is still being used
+	// because the references were retained, then we will
+	// allocate a new buffer.
+	columnReader := t.allocateBuffer(1)
```
jsternberg, line 147: We talked offline about this loop and ways we can optimize it further in the future for `aggregateWindow()`. It looks good for its current purpose so I don't see anything else I would change.

storage/flux/reader.go
yzhang1991: It is actually used a lot. See `determineTableColsForGroup()` and `determineTableColsForSeries()`

PR #18046, opened 2020-05-11
fix: update flux parser to fix issues

bthesorceror:

### DIFF
--- a/ui/src/timeMachine/actions/queries.ts
+++ b/ui/src/timeMachine/actions/queries.ts
@@ -1,5 +1,5 @@
 // Libraries
-import {parse} from '@influxdata/flux-parser'
+import {parse} from 'src/external/parser'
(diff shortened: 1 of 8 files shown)


ui/src/timeMachine/actions/queries.ts
```diff
@@ -76,7 +76,14 @@ export const getOrgIDFromBuckets = (
   text: string,
   allBuckets: Bucket[]
 ): string | null => {
-  const ast = parse(text)
+  let ast: any
```
hoorayimhelping, line 79: did it throw a compiler warning not having `: any` here? typing something as `any` is kind of implied and so i don't think we do it very often.

bthesorceror: actually I meant to remove this, so good catch :)

PR #18047, opened 2020-05-12
feat(demodata):  Add GA events for demodata

ebb-tide:
Closes 7080

### DIFF
--- a/ui/src/dashboards/components/DashboardHeader.tsx
+++ b/ui/src/dashboards/components/DashboardHeader.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {FC} from 'react'
+import React, {FC, useEffect} from 'react'
 import {connect} from 'react-redux'
 import {withRouter, WithRouterProps} from 'react-router'
 
@@ -24,6 +24,9 @@ import {
   setDashboardTimeRange as setDashboardTimeRangeAction,
   updateQueryParams as updateQueryParamsAction,
 } from 'src/dashboards/actions/ranges'
+
+// Utils
+import {fireDashboardViewedEvent} from 'src/shared/utils/analytics'
 
 // Selectors
 import {getTimeRange} from 'src/dashboards/selectors'
@@ -85,6 +88,10 @@ const DashboardHeader: FC<Props> = ({
   router,
   org,
 }) => {
+  useEffect(() => {
+    fireDashboardViewedEvent(dashboard.name)
+  }, [dashboard.id])
+
   const handleAddNote = () => {
     router.push(`/orgs/${org.id}/dashboards/${dashboard.id}/notes/new`)
   }
--- a/ui/src/dashboards/components/DashboardHeader.tsx
+++ b/ui/src/dashboards/components/DashboardHeader.tsx
@@ -1,5 +1,5 @@
 // Libraries
-import React, {Component} from 'react'
+import React, {FC} from 'react'
 import {connect} from 'react-redux'
 import {withRouter, WithRouterProps} from 'react-router'
 
@@ -14,15 +14,15 @@ import TimeZoneDropdown from 'src/shared/components/TimeZoneDropdown'
 import {Button, IconFont, ComponentColor, Page} from '@influxdata/clockface'
 
 // Actions
-import {toggleShowVariablesControls} from 'src/userSettings/actions'
-import {updateDashboard} from 'src/dashboards/actions/thunks'
+import {toggleShowVariablesControls as toggleShowVariablesControlsAction} from 'src/userSettings/actions'
+import {updateDashboard as updateDashboardAction} from 'src/dashboards/actions/thunks'
 import {
-  setAutoRefreshInterval,
-  setAutoRefreshStatus,
+  setAutoRefreshInterval as setAutoRefreshIntervalAction,
+  setAutoRefreshStatus as setAutoRefreshStatusAction,
 } from 'src/shared/actions/autoRefresh'
 import {
-  setDashboardTimeRange,
-  updateQueryParams,
+  setDashboardTimeRange as setDashboardTimeRangeAction,
+  updateQueryParams as updateQueryParamsAction,
 } from 'src/dashboards/actions/ranges'
 
 // Selectors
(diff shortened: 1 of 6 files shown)


ui/src/index.tsx
```diff
@@ -157,6 +157,7 @@ declare global {
 
 // Older method used for pre-IE 11 compatibility
 window.basepath = basepath
+window.dataLayer = window.dataLayer || []
```
drdelambre, line 160: idk how this is supposed to work, dataLayer isn't an array in quartz

121watts: @drdelambre [i think it is](https://github.com/influxdata/quartz/blob/1efa065cf012dcb5f90c0c0bc3bf573d873f51c3/assets/js/app.jsx#L45)

drdelambre: has it changed from this?
https://github.com/influxdata/quartz/blob/d691434c6bcfd564cb54d089654cf57480261423/assets/js/client/GoogleAnalyticsClient.test.ts#L6

drdelambre: just bring it up because we've had a lot of false starts with this one: https://github.com/influxdata/idpe/issues/6298 and it'd be cool to add new context on this one

121watts: that's def newer than the array.  for sure.  im fine with either.  as long as they are the same.

drdelambre: i think the `window.dataLayer = []` is a fix to prevent errors from crashing the page in the event of someone calling `GA.push(event)` before the actual GA client has been loaded

ebb-tide: I don't understand quartz and the UI interact here at all and I don't understand how GA works. if anyone does I would love some help.

hoorayimhelping: So we're all clear: `window.dataLayer` is a data structure used by Google Tag Manager. Appending an item to this array is how we "fire" events for GTM. We can view those events in google analytics or google tag manager. We're using this system absent another one because it's already in place and it's already paid for (i.e. the costs of using it are very low relative to adopting a new one).

Quartz and Cloud both use the same google tag manager system, and quartz has its own set of events that are fired. We are able to correlate user behavior between cloud and quartz, but it's not our (ui engineers) concern on how to do that right now. Basically you can view this as: *as long as you don't overwrite events, everything should interact fine without much conflict*

In cloud, when we deploy, we overwrite our OSS index.html file [with an idpe one](https://github.com/influxdata/idpe/blob/master/apps/chronografd/index.html#L8) which sets up GTM correctly. I love the defensiveness of the line, and having it in will  not hurt anything, but it's not necessary, since this same action is done in the index.html file.


https://developers.google.com/tag-manager/devguide

ui/src/shared/utils/analytics.ts
```diff
@@ -23,3 +25,25 @@ export const fireUserDataReady = function fireUserDataReady(
...
+  } else {
+    fireGAEvent('demoData_queried')
+  }
+}
+
+export const fireDashboardViewedEvent = (dashboardName: string) => {
+  const demo_dataset = DemoDataDashboardNames[dashboardName]
+  if (demo_dataset) {
+    fireGAEvent('demoData_dashboardViewed', {demo_dataset})
+  }
```
hoorayimhelping, line 48: is this event only valuable with the demo dataset? or could something like this also be of value:

```ts
else {
  fireGAEvent('demoData_dashboardViewed')
}
```

ui/src/shared/utils/analytics.ts
```diff
@@ -23,3 +25,25 @@ export const fireUserDataReady = function fireUserDataReady(
     },
   })
 }
+
+export const fireGAEvent = (event: string, payload: object = {}) => {
```
hoorayimhelping, line 29:
```suggestion
export const fireEvent = (event: string, payload: object = {}) => {
```

small thing, but I think dropping mention of GA is probably ideal - we abstract the underlying mechanism and only worry about passing in a string and a payload. we turn this function into a separate thing that handles events, regardless of the underlying mechanism.

ui/src/dashboards/components/DashboardHeader.tsx
```diff
@@ -197,6 +136,60 @@ class DashboardHeader extends Component<Props> {
...
+            selected={autoRefresh}
+          />
+          <TimeRangeDropdown
+            onSetTimeRange={handleChooseTimeRange}
+            timeRange={timeRange}
+          />
+        </Page.ControlBarRight>
+      </Page.ControlBar>
+    </>
+  )
```
hoorayimhelping, line 192: it's hard to tell in this diff, did any code change here, or is this just a reshuffling of code?

ui/src/shared/utils/analytics.ts
hoorayimhelping: also, i just noticed this is JS code, camelCase that variable name!

ui/src/index.tsx
drdelambre: for future reference and in the spirit of sharing knowledge, we haven't used that file since  november-ish? please refer all changes to GTM in https://github.com/influxdata/monitor-ci/blob/master/deploy/Jenkinsfile.deploy#L386

ui/src/dashboards/components/DashboardHeader.tsx
ebb-tide: just added 
```  
useEffect(() => {
    fireDashboardViewedEvent(dashboard.name)
  }, [dashboard.id])
```

ui/src/shared/utils/analytics.ts
hoorayimhelping: had a thought: what do you think of moving this file under `ui/cloud` since events are only really fired in a cloud context?

ui/src/shared/utils/analytics.ts
ebb-tide: I'm trying to keep this PR to only the things that were requested for demo data, when we actually instrument GA, we can take a wholistic view of all the events we should track and how.

ui/src/shared/utils/analytics.ts
ebb-tide: I am assuming all of this will be refactored once the actual implementation of GA eventing goes out.

PR #18059, opened 2020-05-12
fix: Adjusting the width of dropdown

### LINKED ISSUE #18021: Dashboards: Variable values are truncated
the issue here is that the important part of this value is the very end, that is truncated.

- When the drop down is opened, we should expand the list horizontally to show the full value of the the variable.

image

zoesteinkamp:
This will make it so the width of the dropdown will adjust to the longest variable name

Closes #18021 

Ive done a simple math problem that determines the longest length in the values and makes that the width px
Screen Shot 2020-05-12 at 12 04 44 PM

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/src/variables/components/VariableDropdown.tsx
+++ b/ui/src/variables/components/VariableDropdown.tsx
@@ -43,12 +43,12 @@ class VariableDropdown extends PureComponent<Props> {
     const dropdownStatus =
       values.length === 0 ? ComponentStatus.Disabled : ComponentStatus.Default
 
-    const widthLength = Math.max(
-      140,
-      values.reduce(function(a, b) {
+    const longestItemWidth =
+      Math.floor(values.reduce(function(a, b) {
         return a.length > b.length ? a : b
-      }, '').length * 8
-    )
+      }, '').length * 8.5)
+
+    const widthLength = Math.max(140, longestItemWidth)
 
     return (
       <div className="variable-dropdown">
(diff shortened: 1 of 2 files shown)


ui/src/variables/components/VariableDropdown.tsx
```diff
@@ -43,6 +43,13 @@ class VariableDropdown extends PureComponent<Props> {
     const dropdownStatus =
       values.length === 0 ? ComponentStatus.Disabled : ComponentStatus.Default
 
+    const widthLength = Math.max(
+      140,
+      values.reduce(function(a, b) {
+        return a.length > b.length ? a : b
+      }, '').length * 8
```
drdelambre, line 50: how did you choose the number eight?

zoesteinkamp: Im not going to lie, it looked the best in the UI. I tried a few numbers, and 8 looked ideal for the px

drdelambre: is the intent to find out what the max width of the longest variable _could_ be?

drdelambre:
@alexpaxton do you have any fancy tricks for measuring text up your sleeve? i have a fear that we're tightly coupling a font choice to system behavior here

ui/src/variables/components/VariableDropdown.tsx
zoesteinkamp: So the intent is find the longest variable, then take its length and determine how much width in px would be needed to display it

alexpaxton:
If we style the dropdown items to be in the monospace font then all characters will be the same width, and more predictable

Issue #18026, opened 2020-05-10
Misaligned cursor/selection position in Flux query editor


DSpeichert:
__Steps to reproduce:__

1. Create a dashboard in UI
2. Edit a chart writing a query manually
3. Click somewhere in the query and start typing or press END key and start typing

__Expected behavior:__
Cursor should show cursor position.
Selection should show selection.

__Actual behavior:__
Visible cursor position is shifted to the left.
Typing shows that characters are entered in the intended position, not where cursor appears to be.
When selecting text, wrong portion is shown as selected but correct text is actually selected.

![image](https://user-images.githubusercontent.com/1254971/81490767-15ddff80-9254-11ea-9833-92d529564421.png)
1: Click here
2: Cursor shows up here
3. New text will show up here

__Environment info:__

* System info: `Linux 5.3.18-3-pve x86_64`
* InfluxDB version: `InfluxDB 2.0.0-beta.10 (git: d00706dea8) build_date: 2020-05-10T04:07:34Z`
* Web browser: Chrome v81 on Windows

desa:
@Asalem1 do we believe this is currently fixed?

Asalem1:
For anyone that plans on working on this, this might be a helpful comment:

https://stackoverflow.com/a/58349818/7963795

hoorayimhelping:
@DSpeichert I can't reproduce this on Firefox or Chrome on Windows 10.`end` and `home` both worked as expected as did highlighting text

Is this still happening for you?

DSpeichert:
Yes, still happening but i'm running the v2.0.0-beta image that hasn't been updated for a while (d00706d).

hoorayimhelping:
@DSpeichert 

That makes sense. I haven't gone through and figured out when this was fixed, but I'm about 90% certain was fixed sometime between the Beta-10 image you're running and the current version. We are currently in the process of deploying Beta-12, which has a bunch of new features and bug fixes if you're interested in upgrading.

I'm going to close this since I can't reproduce it and it seems fixed in our most recent release. If you update to the latest version and still find this happening, please re-open this and tag me and I'll look into it.

desa:
@DSpeichert let us know if you're still experiencing the issue after upgrade

DSpeichert:
@desa are there any plans to release beta-11 or beta-12 image on quay?
If `v2.0.0-beta` is beta-10 and `nightly` is InfluxDB1, there is no docker image to update to.

desa:
[...]

PR #18063, opened 2020-05-12
feat: move cursor to end of text

drdelambre:
if you turn on the flag `cursorAtEOF`, the cursor will default to the end of a prepopulated flux query when you open monaco. i found the "focus at top of all editors at initialization" behavior annoying when viewing multiple editors with prepolpulated flux comments for the notebooks project (#17966).

prop drilling through the TimeMachineFluxEditor is an artifact of the need for ongoing state decoupling in this area of the code and needed to verify this work.

since the concept of `cursor position` is nested deep in monaco, the tests for this looked a lot like "let a = [], b = a, expect(b).toEqual(a)", which is why we dont see any here.

### DIFF
--- a/ui/src/shared/components/FluxMonacoEditor.tsx
+++ b/ui/src/shared/components/FluxMonacoEditor.tsx
@@ -27,6 +27,7 @@ interface Props {
   onChangeScript: OnChangeScript
   onSubmitScript?: () => void
   setEditorInstance?: (editor: EditorType) => void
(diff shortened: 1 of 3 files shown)


Asalem1:
@drdelambre does this have anything to do with:

https://github.com/influxdata/influxdb/issues/18026

drdelambre:
@Asalem1 unfortunately not

PR #18071, opened 2020-05-12
fix(variable_hydration): limiting variable hydration to changing variables

Asalem1:
Closes https://github.com/influxdata/idpe/issues/7030

### Problem

This PR aims to address the issue of overly hydrating variables when a variable selection is made. 

### Solution

The solution to this is twofold:

- Compare current selection to previous selection to ensure that selecting the same variable value will not trigger an action to hydrate the variables
- Only pass in the variable that has been selected to hydrate that specific variable

Here's how these changes impact the network requests:

dehydrate-variables

### UI Tweak

This PR also includes a minor UI tweak that may have been the result of the clockface pendulum update.

Previous:
Screen Shot 2020-05-12 at 14 26 10

Now:
Screen Shot 2020-05-12 at 15 05 58

- [x] Rebased/mergeable

### DIFF
--- a/ui/src/variables/utils/hydrateVars.ts
+++ b/ui/src/variables/utils/hydrateVars.ts
@@ -129,6 +129,8 @@ const findSubgraph = (
 
   // use an ID array to reduce the chance of reference errors
   const varIDs = variables.map(v => v.id)
+  // create an array of IDs to reference later
+  const graphIDs = []
   for (const node of graph) {
     const shouldKeep =
       varIDs.includes(node.variable.id) ||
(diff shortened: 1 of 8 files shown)


ui/src/variables/utils/hydrateVars.ts
```diff
@@ -126,9 +126,10 @@ const findSubgraph = (
   variables: Variable[]
 ): VariableNode[] => {
   const subgraph: Set<VariableNode> = new Set()
-
   // use an ID array to reduce the chance of reference errors
   const varIDs = variables.map(v => v.id)
+  // create an array of IDs to reference later
+  const graphIDs = []
```
drdelambre, line 132: note for self: this was because `Set.prototype.has` compares by reference, which we dont guarantee in an immutable kind of world

drdelambre, line 132: nit for Ari: this should be an object for a faster lookup on line 148, but i can't imagine the search space being very large, so it's not a blocker

PR #18078, opened 2020-05-13
refactor: add new label package

### LINKED ISSUE #18105: Refactor Label Service
We want to refactor the Label code into a new package using the pattern established in the Tenant Service. This work includes:

- [x] Creating an HTTP Server/Client using chi router
- [x] Separating service and storage code 
    - [x] Write Label CRUD storage functions 
    - [x] Write Label Mapping Create and Delete storage functions
    - [x] Write Label CRUD service functions 
    - [x] Write Label Mapping Create and Delete service functions
- [x] Auth middleware layer
- [x] Logging middleware layer 
- [x] Metrics middleware layer
- [x] Test middleware layers
- [x] Write Service tests
- [x] Write Storage tests
- [x] Write HTTP tests

(make URM code public so I don't have to copy all the functions over to the label package?)

Next steps include Feature Flagging the package and removing the URMs from the Label service. After the new code has been sufficiently tested behind the feature flag, the old service code can be removed.

Also related would be to address https://github.com/influxdata/influxdb/issues/11278

AlirieGray:
Closes #18105

This PR adds the new label service to conform to the pattern established in the tenant and authorization services, where the http, middleware, service, and storage code is contained in the same package.

It also makes a couple of incidental changes, such as moving Metrics helper functions from the tenant and authorization packaged to their own file in kit (this could be moved somewhere else if there's a better place). Also changing `CreateUserResourceMappingForOrg` in kv from private to public (this can be changed back once URMs are removed from the Label Service).

The KV Service passed to the new Label Service can also be removed once URMs are removed from the Label Service.

Next step will be to put this code behind a feature flag and incrementally test.

- [x] Rebased/mergeable
- [x] Tests pass


label/middleware_auth.go
```diff
@@ -0,0 +1,145 @@
...
+		return nil, err
+	}
+	// first fetch all labels for this resource
+	ls, err := s.s.FindResourceLabels(ctx, filter)
+	if err != nil {
+		return nil, err
+	}
+
+	// check the permissions for the resource by the org on the context
+	orgID := orgIDFromContext(ctx)
```
AlirieGray, line 70: @GeorgeMac the original authorizer code passed in an Org Service and got the OrgID by looking up the resource on the filter, I changed it to use the OrgID on the context to make things a little cleaner and more separated. Will this result in unwanted behavior? (It does pass the existing tests if I put the same OrgID on the context but there could be something I'm missing.)

GeorgeMac: Oooh interesting. My only concern is my lack of understanding around what puts this orgID on the context. But if it is consistent perhaps it doesn't matter. Will think about it through this review :D This could be all we need.

kit/metric/metrics_options.go
```diff
@@ -1,4 +1,4 @@
-package tenant
+package metric
```
GeorgeMac, line 1: ❤️  Awesome to see this becoming a shared component.

label/storage_label.go
```diff
@@ -0,0 +1,439 @@
...
+
+	for k, v := cur.Next(); k != nil; k, v = cur.Next() {
+		l := &influxdb.Label{}
+		if err := json.Unmarshal(v, l); err != nil {
+			return err
+		}
+		if !fn(l) {
+			break
+		}
+	}
```
GeorgeMac, line 410: Same thing goes here as for above. Need to check the cursor error and close it.

label/storage_label.go
```diff
@@ -0,0 +1,439 @@
...
+		l, err := s.GetLabel(ctx, tx, id)
+		if l == nil && err != nil {
+			// TODO(jm): return error instead of continuing once orphaned mappings are fixed
+			// (see https://github.com/influxdata/influxdb/issues/11278)
+			continue
+		}
+
+		*ls = append(*ls, l)
+	}
+	return nil
```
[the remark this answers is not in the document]
AlirieGray: good catch, thank you!

label/service.go
```diff
@@ -0,0 +1,283 @@
...
+	err := s.store.Update(ctx, func(tx kv.Tx) error {
+		if err := uniqueLabelName(ctx, tx, l); err != nil {
+			return err
+		}
+
+		if err := s.store.CreateLabel(ctx, tx, l); err != nil {
+			return err
+		}
+
+		if err := s.kvSvc.CreateUserResourceMappingForOrg(ctx, tx, l.OrgID, l.ID, influxdb.LabelsResourceType); err != nil {
```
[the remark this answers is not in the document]
AlirieGray: I just tried to implement this and ran into the problem I had when I initially tried to copy this function over, which is that it calls functions on the kv service in order to do the update (and those each in turn call other functions on the service). I could be missing something but I don't know if it's possible to access this code without including a kv service in there somewhere, even if it were on this other interface

label/storage_label.go
```diff
@@ -0,0 +1,439 @@
...
+func (s *Store) CreateLabel(ctx context.Context, tx kv.Tx, l *influxdb.Label) error {
+	// if the provided ID is invalid, or already maps to an existing Auth, then generate a new one
+	if !l.ID.Valid() {
+		id, err := s.generateSafeID(ctx, tx, labelBucket)
+		if err != nil {
+			return nil
+		}
+		l.ID = id
+	} else if err := uniqueID(ctx, tx, l.ID); err != nil {
+		id, err := s.generateSafeID(ctx, tx, labelBucket)
```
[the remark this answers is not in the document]
AlirieGray: I think in general we want to just ignore the ID and create our own, listening to the ID being passed here at all is incidental

label/middleware_auth.go
AlirieGray: wound up making `OrgIDFromContext` and `ValidResource` public and moving them from tenant into kit/transport/http. Labels is always nested after all so the org ID being added to the context by `ValidResource` when mounting the routes in the tenant service should be sufficient

label/service.go
```diff
@@ -0,0 +1,282 @@
...
+
+	err := s.store.Update(ctx, func(tx kv.Tx) error {
+		if err := uniqueLabelName(ctx, tx, l); err != nil {
+			return err
+		}
+
+		if err := s.store.CreateLabel(ctx, tx, l); err != nil {
+			return err
+		}
+		if err := s.kvSvc.CreateUserResourceMappingForOrg(ctx, tx, l.OrgID, l.ID, influxdb.LabelsResourceType); err != nil {
```
lyondhill, line 42:
```suggestion
	urmCreator UserResourceMappingCreator
}

type UserResourceMappingCreator interface {
    CreateUserResourceMappingForOrg(ctx context.Context, orgID, id, *influxdb.ID, influxdb.ResourceType) error
}

func NewService(st *Store, urmCreator UserResourceMappingCreator) influxdb.LabelService {
	return &Service{
		store: st,
		urmCreator: urmCreator,
	}
}

// CreateLabel creates a new label.
func (s *Service) CreateLabel(ctx context.Context, l *influxdb.Label) error {
	if err := l.Validate(); err != nil {
		return &influxdb.Error{
			Code: influxdb.EInvalid,
			Err:  err,
		}
	}

	l.Name = strings.TrimSpace(l.Name)

	err := s.store.Update(ctx, func(tx kv.Tx) error {
		if err := uniqueLabelName(ctx, tx, l); err != nil {
			return err
		}

		if err := s.store.CreateLabel(ctx, tx, l); err != nil {
			return err
		}
		if err := s.urmCreator.CreateUserResourceMappingForOrg(ctx, tx, l.OrgID, l.ID, influxdb.LabelsResourceType); err != nil {
```

lyondhill: @GeorgeMac is this what your suggestiong?

label/service.go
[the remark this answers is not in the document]
AlirieGray: (just kidding, Lyon explained to me what you meant! I was trying to avoid passing the kv service at all which wasn't working but now I'm just using this to wrap it)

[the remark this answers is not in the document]
AlirieGray: I did, on Lyon's suggestion, add a wrapping interface that just includes the `CreateUserResourceMappingForOrg` function to remove the direct dependence on kv. However, with this I'll still need to export the function in kv and pass a kv service where the label service gets initialized.

label/service.go
GeorgeMac: Almost. I am suggesting `CreateUserResourceMappingForOrg` does not take a `kv.Tx` and instead it is defined back in `kv` package and starts an `Update` and call the unexported one with a `kv.Tx`.
Then here we could call `CreateUserResourceMappingForOrg(ctx, l.OrgID, l.ID, influxdb.LabelsResourceType)` without the tx and we would do so outside of this current transaction.

GeorgeMac: Moving the last line outside this current `Update` transaction, so we can go through the URM service front door. Ultimately, I we will remove this anyway in the next couple weeks as well.

kv/label.go
```diff
@@ -311,7 +311,8 @@ func (s *Service) PutLabel(ctx context.Context, l *influxdb.Label) error {
 	})
 }
 
-func (s *Service) createUserResourceMappingForOrg(ctx context.Context, tx Tx, orgID influxdb.ID, resID influxdb.ID, resType influxdb.ResourceType) error {
+// CreateUserResourceMappingForOrg can be made private once URMs are removed from the Label Service
+func (s *Service) CreateUserResourceMappingForOrg(ctx context.Context, tx Tx, orgID influxdb.ID, resID influxdb.ID, resType influxdb.ResourceType) error {
```
GeorgeMac, line 315:
```suggestion
func (s *Service) CreateUserResourceMappingForOrg(ctx context.Context, orgID influxdb.ID, resID influxdb.ID, resType influxdb.ResourceType) error {
	return s.kv.Update(ctx, func(tx Tx) error {
		return s.createUserResourceMappingForOrg(ctx, tx, orgID, resID, resType)
	}
}

func (s *Service) createUserResourceMappingForOrg(ctx context.Context, tx Tx, orgID influxdb.ID, resID influxdb.ID, resType influxdb.ResourceType) error {
```

This was you can use the lower-case create in the kv package. But externally you can consume the exported version which implicitly creates an update transaction.

label/service.go
GeorgeMac: I added a suggestion below for how to change `kv` to support this.

AlirieGray: Okay so I just implemented this as well as the wrapper that Lyon suggested!

label/service.go
```diff
@@ -0,0 +1,294 @@
...
+		}
+
+		return nil
+	})
+
+	if err != nil {
+		return err
+	}
+	err = s.store.Update(ctx, func(tx kv.Tx) error {
+		if err := s.urmCreator.CreateUserResourceMappingForOrg(ctx, tx, l.OrgID, l.ID, influxdb.LabelsResourceType); err != nil {
```
GeorgeMac, line 54: Feel free to disregard this. But now that you have expose the transaction as an argument, you can actually do this all one transaction.
The only reason I suggested moving it out of the transaction, was because the new mapping creator service interface wasn't going to expose the transaction. The theory being that it removed any concept of kv from that operation. So we could if we wanted to, swap out the a kv one for a remote service for example. But if we're going to leave in the transactionality, then we can keep it all on one Tx.

GeorgeMac: p.s. please don't let this slow you down, go ahead with it :) Just wanted to clarify the reasoning behind not having the Tx in the method.

label/service.go
```diff
@@ -0,0 +1,294 @@
...
+	for k, v := range upd.Properties {
+		if v == "" {
+			delete(label.Properties, k)
+		} else {
+			label.Properties[k] = v
+		}
+	}
+
+	if upd.Name != "" {
+		err := s.store.Update(ctx, func(tx kv.Tx) error {
```
GeorgeMac, line 145: I feel like this logic should live in Store and the `*label.Store.UpdateLabel` should take the LabelUpdate.
Meaning all Service does is open a transaction and pass the right things down to storage.
My concern here is that after updating the name, we could fail validation of the label or fail to actually put the label, which is all `store.UpdateLabel` is doing at the moment.
This would also reduce this function down to 1 transaction instead of opening 3.
So you could get a non-successfull call to this service to update the label, but the name did change successfully.

PR #18082, opened 2020-05-13
fix(useVisYDomainSettings): added check for undefined values

Asalem1:
### Problem

In some situations, storedDomain was being passed as undefined and was triggering a honeybadger error when trying to use the `every` method on undefined

### Solution

Added a check for `undefined` in the `useVisYDomainSettings`

### DIFF
--- a/ui/src/shared/utils/useVisDomainSettings.ts
+++ b/ui/src/shared/utils/useVisDomainSettings.ts
@@ -80,8 +80,7 @@ export const useVisYDomainSettings = (
 ) => {
   const initialDomain = useMemo(() => {
     if (
-      storedDomain === undefined ||
-      storedDomain === null ||
+      !Array.isArray(storedDomain) ||
       storedDomain.every(val => val === null)
     ) {
       return getValidRange(data, timeRange)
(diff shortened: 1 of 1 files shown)


ui/src/shared/utils/useVisDomainSettings.ts
```diff
@@ -79,7 +79,11 @@ export const useVisYDomainSettings = (
   timeRange: TimeRange | null = null
 ) => {
   const initialDomain = useMemo(() => {
-    if (storedDomain === null || storedDomain.every(val => val === null)) {
+    if (
+      storedDomain === undefined ||
+      storedDomain === null ||
+      storedDomain.every(val => val === null)
```
hoorayimhelping, line 85: suggestion: let the language do this work for you. we thought it could only be `null` but it ended up being `undefined` - i get the sense that it's not important what the value is - as long as it's not an array, or an array with no values, it should fall into this loop: 

```suggestion
      !Array.isArray(storedDomain) ||
      storedDomain.every(val => val === null)
```

Asalem1: I hear what you're saying, but if that is in fact the case that we don't care about what the value is, shouldn't we just handle every falsey value?

`!storedDomain || storedDomain.every ... `

?

PR #18084, opened 2020-05-13
chore(shared/constants): increase flux response limit to 27MB

desa:
This was detemined by looking at the 99th percentile of query responses.

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/src/shared/constants/index.ts
+++ b/ui/src/shared/constants/index.ts
@@ -65,7 +65,7 @@ export const CLOUD_LOGOUT_PATH = '/logout'
 export const CLOUD_USERS_PATH = '/users'
 
 export const FLUX_RESPONSE_BYTES_LIMIT = CLOUD
-  ? 10 * 1024 * 1024 // 10 MiB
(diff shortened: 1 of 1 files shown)


ui/src/shared/constants/index.ts
```diff
@@ -65,7 +65,7 @@ export const CLOUD_LOGOUT_PATH = '/logout'
 export const CLOUD_USERS_PATH = '/users'
 
 export const FLUX_RESPONSE_BYTES_LIMIT = CLOUD
-  ? 10 * 1024 * 1024 // 10 MiB
+  ? 27 * 1024 * 1024 // 27 MiB  (desa: this was determined by looking at queries responses in the cloud app)
```
hoorayimhelping, line 68: more medium term, is this something we could inject from some outside configuration? the goal being not having to redeploy to change this value. even just restarting the app after a config change would be preferable to redeploying 🤔

desa: We don't currently have that capability, but we could definitely add it.

PR #18089, opened 2020-05-13
feat: adding notebook context

drdelambre:
creating a context provider for the notebook interaction (#17966) based on the suggested structure from this talk (thanks @121watts):
https://www.youtube.com/watch?v=CDGBTjMBJzg&t=523s

I've anchored the context provider to the page for now. we'll probably want to figure out a better solution than `useState` in the near future so that components can define their own state requirements more towards the leaf nodes, and have that state shared between them, but this technique should hold up through the prototyping phase.

i've included the test mock data i've been using to ensure that interactions between notebook stages are working properly, figuring that they might be useful until we have more of the crud level interactions fleshed out.

### DIFF
--- a/ui/src/notebooks/context/notebook.mock.json
+++ b/ui/src/notebooks/context/notebook.mock.json
@@ -1,0 +1,78 @@
+{
+  "id": "testing",
+  "pipes": [
+    {
+      "type": "query",
+      "activeQuery": 0,
+      "queries": [
+        {
+          "text": "from(bucket: \"project\")\n |> range(start: v.timeRangeStart, stop: v.timeRangeStop)\n |> filter(fn: (r) => r[\"_measurement\"] == \"docker_container_cpu\")",
+          "editMode": "advanced",
+          "builderConfig": {
+            "buckets": [],
+            "tags": [],
+            "functions": []
+          }
+        }
+      ]
+    },
+    {
+      "type": "query",
+      "activeQuery": 0,
+      "queries": [
+        {
+          "text": "// Tip: Use the __PREVIOUS_RESULT__ variable to link your queries\n\n__PREVIOUS_RESULT__\n |> filter(fn: (r) => r[\"_field\"] == \"usage_percent\")",
+          "editMode": "advanced",
+          "builderConfig": {
+            "buckets": [],
+            "tags": [],
+            "functions": []
+          }
+        }
+      ]
+    },
+    {
+        "type": "visualization"
+    },
+    {
+        "type": "visualization"
+    },
+    {
+      "type": "query",
+      "activeQuery": 0,
+      "queries": [
+        {
+          "text": "// Tip: Use the __PREVIOUS_RESULT__ variable to link your queries\n\n__PREVIOUS_RESULT__\n |> sum()",
+          "editMode": "advanced",
+          "builderConfig": {
+            "buckets": [],
+            "tags": [],
+            "functions": []
+          }
+        }
+      ]
+    },
+    {
+        "type": "visualization"
+    },
+    {
+      "type": "query",
+      "activeQuery": 0,
+      "queries": [
+        {
+          "text": "// Tip: Use the __PREVIOUS_RESULT__ variable to link your queries\n\nfrom(bucket: \"project\")\n |> range(start: v.timeRangeStart, stop: v.timeRangeStop)\n |> filter(fn: (r) => r[\"_measurement\"] == \"docker_container_cpu\")",
+          "editMode": "advanced",
+          "builderConfig": {
+            "buckets": [],
+            "tags": [],
+            "functions": []
+          }
+        }
+      ]
+    },
+    {
+        "type": "visualization"
+    }
+  ]
+}
+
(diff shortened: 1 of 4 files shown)


ui/src/notebooks/context/notebook.tsx
```diff
@@ -0,0 +1,78 @@
...
+  pipes: [],
+  addPipe: () => {},
+  updatePipe: () => {},
+  removePipe: () => {},
+}
+
+// NOTE: this just loads some test data for exploring the
+// interactions between types. Make sure it's never true
+// during the pull review process
+const TEST_MODE = false
```
zoesteinkamp, line 26: lets consider hiding this behind a feature flag

ui/src/notebooks/context/notebook.tsx
```diff
@@ -0,0 +1,78 @@
...
+)
+
+export const NotebookProvider: FC = ({children}) => {
+  const [id] = useState(DEFAULT_CONTEXT.id)
+  const [pipes, setPipes] = useState(DEFAULT_CONTEXT.pipes)
+
+  function addPipe(pipe: Pipe) {
+    setPipes(pipes => {
+      pipes.push(pipe)
+      return pipes.slice()
```
zoesteinkamp, line 44: whats our preferred method of doing this? we've also seen this form
```ts
[...pipes] 
```

ui/src/notebooks/context/notebook.mock.json
```diff
@@ -0,0 +1,78 @@
...
+            "functions": []
+          }
+        }
+      ]
+    },
+    {
+        "type": "visualization"
+    },
+    {
+        "type": "visualization"
```
zoesteinkamp, line 38: Whats this for?

hoorayimhelping:
Zoe and I paired up on this review, so the comments are from us together.

One thing I noticed that we might need to be careful how we strike the balance between making this work for the future uses vs being too generic to reasonably follow. I get that the problem we're solving that current data explorer is too coupled will necessarily color this implementation, I just want to make sure we're striking the right balance between generic and understandable.

The concrete freedback from all that is that we spent about 10 minutes talking through how the pipes data structure should work and what it actually represents - it's a generic collection that is meant to be passed between React context rights? But currently, query datums are the only things being passed? It would have been a lot easier to follow if it had just been a data structure that shares queries, I think. I get the need to make it generic, but it felt a little early for that while we're still working out the architectural details, at least in this PR. Not a criticism or anything actionable, just maybe something to keep in mind.

ui/src/notebooks/context/notebook.mock.json
drdelambre: different types

ui/src/notebooks/context/notebook.tsx
drdelambre: not sure why we would, but maybe i'm misunderstanding the intent here.

ui/src/notebooks/context/notebook.tsx
drdelambre: i have no preference, but considering that it's not standardized across the code base, that if we're gonna try to make a choice about this, it not be made here, but opened up to the group and captured in a style guide that we can maintain / distribute / refer to in moments like this.

drdelambre:
@hoorayimhelping yeah, this PR isn't that at all. I think you're reading too much into there being mock data? this is just a react context to introduce and get feedback on the pattern of state manipulation that future PRs are built on (having a functional component returning a provider).

If i'm reading this comment correctly, you're looking for https://github.com/influxdata/influxdb/pull/17966/files#diff-0fce7d4eba40cb20aa35815480502420R6, which is gonna happen in a different PR. And then there'll be a follow up one about how to implement individual types, and on showing how to construct interdependent queries, and another one about registering existing redux state, and one about using local storage as a mock api, and one for adding a task type, and one for adding a markdown type, and one for adding an alert type, and another one about dashboard importing and exporting, etc...

The idea is that I've played with these constructs and how they interact in the linked spike PR in an attempt to derisk some of the assumptions we've made while developing this idea, so that's always there as a reference in case you want some insight into that interaction or how this piece fits into a bigger context, but that we should introduce and discuss the merits of aspects of the subsystems individually instead of working towards a consensus over a giant "CHANGED EVERYTHING, DEAL WITH IT" PR.

hoorayimhelping:
@drdelambre I don't think I communicated what I was thinking very clearly at all, sorry about that. 

>but that we should introduce and discuss the merits of aspects of the subsystems individually instead of working towards a consensus over a giant "CHANGED EVERYTHING, DEAL WITH IT" PR.

Agreed, this is way better, thanks for doing it in this way. It makes all our jobs reviewing code and designs easier, and I think it produces much more effective designs and cleaner code. Teamwork makes the dream work!

What I was trying to say is that the construct of pipes feels a bit generic and abstract for what they're providing in this PR. I imagine there's additional context you have that I don't on how they'll evolve, but from this slice of PR, it looks like they're solving future problems that don't exist right now.

Just meant it as a observation without judgement or suggestion because I know you spiked on this previously.

PR #18093, opened 2020-05-14
feat(cmd/influx/write): enhance the ability to import CSV files using csv2lp library

### LINKED ISSUE #17004: Add the ability to import a simple csv file via the influx cli

sranka:
Closes #17004 #17008

This PR enhances the way of how CSV data are processed by `influx write` so that even external CSV data can be written to InfluxDB. 
- It uses a newly introduced `csv2lp package` to convert CSV to the line protocol. The conversion relies solely upon the content on CSV data on input. It is well documented in https://github.com/influxdata/influxdb/tree/master/pkg/csv2lp/README.md, examples are therein as well.
- there are a few new flags in `influx write` that helps the user to import the existing CSV file without changing it 
   - the CSV file encoding can be specified
   - existing header rows can be skipped and new header/annotation rows can be provided instead in order to describe the conversion to line protocol
  - multiple CSV files can be provided so that additional header/annotation rows can be placed in a file   
  - the user can better debug/tune the CSV to line protocol conversion
```
      --debug                Log CSV columns to stderr before reading data rows
      --encoding string      Character encoding of input files or stdin (default "UTF-8")
  -f, --file stringArray     The path to the file to import
      --header stringArray   One or more header lines to prepend to input data
      --skipHeader int[=1]   Skip the first <n> rows from input data
      --skipRowOnError       Log CSV data errors to stderr and continue with CSV processing
```

- [x] Rebased/mergeable
- [x] Tests pass


sranka:
This PR is a continuation of #17599 that had to be split into two parts. The first part is a csv2lp library that is already in master via #17764. This PR is the second part.

sranka:
Related documentation issue: influxdata/docs-v2#1050

russorat:
I've tested this locally and it works great! looking forward to seeing this merged!

cmd/influx/internal/is_character_device.go
```diff
@@ -0,0 +1,19 @@
+package internal
+
+import (
+	"io"
+	"os"
+)
+
+// IsCharacterDevice returns true if the supplied reader is a character device (a terminal)
+func IsCharacterDevice(reader io.Reader) bool {
```
jsteenb2, line 9: can you move this to the cmd/influx main pkg itself. Not seeing the reuse here

write/csvTable.go
```diff
@@ -1,420 +0,0 @@
-package write
```
jsteenb2, line 1: Can you help me understand why these `write/*` deletions are here?

cmd/influx/internal/is_character_device.go
sranka: moved to write

write/csvTable.go
sranka: The conversion of CSV to line protocol was refactored and improved in https://github.com/influxdata/influxdb/tree/master/pkg/csv2lp  as requested by https://github.com/influxdata/influxdb/pull/17599#issuecomment-614174079

PR #18096, opened 2020-05-14
feat(http): Client disconnections should return a HTTP 499 error code.

benbjohnson:
This commit checks `http.Request.Context().Err()` to see if the context has been canceled or deadlined before writing an error code. It uses the non-standard Nginx `499` error code for client disconnection and `408` for client timeout.

https://httpstatuses.com/499

Helps influxdata/idpe#7124


kit/transport/http/error_handler.go
```diff
@@ -51,12 +51,21 @@ func StatusCodeToErrorCode(statusCode int) string {
 
 // ErrorCodeToStatusCode maps an influxdb error code string to a
 // http status code integer.
-func ErrorCodeToStatusCode(code string) int {
+func ErrorCodeToStatusCode(ctx context.Context, code string) int {
```
jacobmarble, line 54: note: juicy bits here

PR #18108, opened 2020-05-14
feat: adding basic structure for notebooks

drdelambre:
This installment adds the structure around notebooks to contain the individual interaction pieces. An example pipe is given to illustrate how a pipe registers into the system and how it connects to the structure. A point of contention currently is if the panel ownership exists on the Pipe component, or on the Pipe's component (_really_ need some more names in this thing). This PR illustrates the latter as it keeps the structure of TypeRegistration smaller and disconnected from the view logic (if it should allow for the first row to be deleted, dynamic titles, etc). From here the PRs should start to diverge between adding individual display types and iterating on the structural components.

part of the initiative to spin #17966 out into real changes

Kapture 2020-05-14 at 16 16 36

### DIFF
--- a/ui/src/notebooks/components/Header.tsx
+++ b/ui/src/notebooks/components/Header.tsx
@@ -2,7 +2,7 @@ import React, {FC} from 'react'
 
 import {Page} from '@influxdata/clockface'
 
-const Header: FC<{}> = () => {
+const Header: FC = () => {
   return (
     <>
       <Page.Title title="Notebooks" />
--- a/ui/src/notebooks/components/Header.tsx
+++ b/ui/src/notebooks/components/Header.tsx
@@ -1,0 +1,13 @@
+import React, {FC} from 'react'
+
+import {Page} from '@influxdata/clockface'
+
+const Header: FC<{}> = () => {
+  return (
+    <>
+      <Page.Title title="Notebooks" />
+    </>
+  )
+}
+
+export default Header
--- a/ui/src/notebooks/components/Notebook.tsx
+++ b/ui/src/notebooks/components/Notebook.tsx
@@ -2,13 +2,26 @@ import React, {FC} from 'react'
 
 import {Page} from '@influxdata/clockface'
 import {NotebookProvider} from 'src/notebooks/context/notebook'
+import Header from 'src/notebooks/components/Header'
+import PipeList from 'src/notebooks/components/PipeList'
+import AddButtons from 'src/notebooks/components/AddButtons'
+
+// NOTE: uncommon, but using this to scope the project
+// within the page and not bleed it's dependancies outside
+// of the feature flag
+import 'src/notebooks/style.scss'
 
 const NotebookPage: FC = () => {
   return (
     <NotebookProvider>
       <Page titleTag="Notebook">
-        <Page.Header fullWidth={false} />
-        <Page.Contents fullWidth={false} scrollable={true} />
+        <Page.Header fullWidth={false}>
+          <Header />
+        </Page.Header>
+        <Page.Contents fullWidth={false} scrollable={true}>
+          <PipeList />
+          <AddButtons />
+        </Page.Contents>
       </Page>
     </NotebookProvider>
   )
--- a/ui/src/notebooks/components/Panel.tsx
+++ b/ui/src/notebooks/components/Panel.tsx
@@ -1,6 +1,5 @@
 // Libraries
-import React, {FC, ReactChildren, useState} from 'react'
-import classnames from 'classnames'
+import React, {FC, ReactChildren} from 'react'
 
 // Components
 import {
@@ -8,9 +7,10 @@ import {
   ComponentSize,
   AlignItems,
   JustifyContent,
-  SquareButton,
-  IconFont,
 } from '@influxdata/clockface'
+
+import usePanelState from 'src/notebooks/hooks/usePanelState'
+import RemoveButton from 'src/notebooks/components/RemoveButton'
 
 interface Props {
   children: ReactChildren | JSX.Element | JSX.Element[]
@@ -20,17 +20,6 @@ interface Props {
   onRemove?: () => void
 }
 
-type PanelState = 'hidden' | 'small' | 'large'
-interface ProgressMap {
-  [key: PanelState]: PanelState
-}
-
-const PANEL_PROGRESS: ProgressMap = {
-  hidden: 'small',
-  small: 'large',
-  large: 'hidden',
-}
-
 const Panel: FC<Props> = ({
   children,
   title,
@@ -38,28 +27,10 @@ const Panel: FC<Props> = ({
   controlsRight,
   onRemove,
 }) => {
-  const [panelState, setPanelState] = useState<PanelState>('small')
-
-  const panelClassName = classnames('notebook-panel', {
-    [`notebook-panel__${panelState}`]: panelState,
-  })
-  const childrenShouldBeVisible =
-    panelState === 'small' || panelState === 'large'
-
-  const handleToggle = (): void => {
-    setPanelState(PANEL_PROGRESS[panelState])
-  }
-
-  let removePanelButton
-
-  if (onRemove) {
-    removePanelButton = (
-      <SquareButton icon={IconFont.Remove} onClick={onRemove} />
-    )
-  }
+  const [className, showChildren, toggle] = usePanelState()
 
   return (
-    <div className={panelClassName}>
+    <div className={className}>
       <div className="notebook-panel--header">
         <FlexBox
           className="notebook-panel--header-left"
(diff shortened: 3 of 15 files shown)


ui/src/notebooks/components/Header.tsx
```diff
@@ -0,0 +1,13 @@
+import React, {FC} from 'react'
+
+import {Page} from '@influxdata/clockface'
+
+const Header: FC<{}> = () => {
```
121watts, line 5:
```suggestion
const Header: FC = () => {
```

ui/src/notebooks/components/Header.tsx
```diff
@@ -0,0 +1,13 @@
...
+  return (
+    <>
+      <Page.Title title="Notebooks" />
+    </>
+  )
```
121watts, line 10:
```suggestion
  return  <Page.Title title="Notebooks" />
```

ui/src/notebooks/components/Notebook.tsx
```diff
@@ -2,13 +2,26 @@ import React, {FC} from 'react'
 
 import {Page} from '@influxdata/clockface'
 import {NotebookProvider} from 'src/notebooks/context/notebook'
+import Header from 'src/notebooks/components/Header'
```
121watts, line 5: Nit here and I know it _feels_ redundant.  But searching for `NotebookHeader` vs `Header` is much easier

ui/src/notebooks/components/Panel.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+interface Props {
+  children: ReactChildren | JSX.Element | JSX.Element[]
+  title: string
+  controlsLeft?: JSX.Element | JSX.Element[]
+  controlsRight?: JSX.Element | JSX.Element[]
+  onRemove?: () => void
+}
+
+type PanelState = 'hidden' | 'small' | 'large'
```
121watts, line 23: I love these union types.  Much preferred to `enum` in this context.

ui/src/notebooks/components/Panel.tsx
```diff
@@ -0,0 +1,93 @@
...
+  controlsRight,
+  onRemove,
+}) => {
+  const [panelState, setPanelState] = useState<PanelState>('small')
+
+  const panelClassName = classnames('notebook-panel', {
+    [`notebook-panel__${panelState}`]: panelState,
+  })
+  const childrenShouldBeVisible =
+    panelState === 'small' || panelState === 'large'
```
121watts, line 47: All this can be pulled into a `usePanelState` hook.

ui/src/notebooks/components/Panel.tsx
```diff
@@ -0,0 +1,93 @@
...
+    setPanelState(PANEL_PROGRESS[panelState])
+  }
+
+  let removePanelButton
+
+  if (onRemove) {
+    removePanelButton = (
+      <SquareButton icon={IconFont.Remove} onClick={onRemove} />
+    )
+  }
```
121watts, line 59: Make this sucker a component.

ui/src/notebooks/components/Panel.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <FlexBox
+          className="notebook-panel--header-left"
+          alignItems={AlignItems.Center}
+          margin={ComponentSize.Small}
+          justifyContent={JustifyContent.FlexStart}
+        >
+          <div className="notebook-panel--toggle" onClick={handleToggle} />
+          <div className="notebook-panel--title">{title}</div>
+          {childrenShouldBeVisible && controlsLeft}
+        </FlexBox>
```
121watts, line 73: Make this a component possibly named `LeftControls` or w/e

ui/src/notebooks/components/Panel.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <FlexBox
+          className="notebook-panel--header-right"
+          alignItems={AlignItems.Center}
+          margin={ComponentSize.Small}
+          justifyContent={JustifyContent.FlexEnd}
+        >
+          {childrenShouldBeVisible && controlsRight}
+          {removePanelButton}
+        </FlexBox>
```
121watts, line 82: And this one

ui/src/notebooks/components/Pipe.tsx
```diff
@@ -0,0 +1,17 @@
...
+import {NotebookContext} from 'src/notebooks/context/notebook'
+
+import {PIPE_DEFINITIONS, PipeProp} from 'src/notebooks'
+
+const Pipe: FC<PipeProp> = ({idx}) => {
+  const {pipes} = useContext(NotebookContext)
+
+  if (!PIPE_DEFINITIONS.hasOwnProperty(pipes[idx].type)) {
+    throw new Error(`Pipe type [${pipes[idx].type}] not registered`)
+    return null
```
121watts, line 11: Looks like this return is not going to be reached.

ui/src/notebooks/components/PipeList.tsx
```diff
@@ -0,0 +1,14 @@
...
+import Pipe from 'src/notebooks/components/Pipe'
+import {NotebookContext} from 'src/notebooks/context/notebook'
+
+const PipeList: FC = () => {
+  const {id, pipes} = useContext(NotebookContext)
+  const _pipes = pipes.map((_, idx) => (
+    <Pipe idx={idx} key={`pipe-${id}-${idx}`} />
+  ))
+
+  return <>{_pipes}</>
```
121watts, line 11:
```suggestion
  return _pipes
```
you can just straight up return arrays now 😸

ui/src/notebooks/index.ts
```diff
@@ -0,0 +1,41 @@
...
+export interface PipeProp {
+  idx: number
+}
+
+// NOTE: keep this interface as small as possible and
+// don't take extending it lightly. this should only
+// define what ALL pipe types require to be included
+// on the page.
+export interface TypeRegistration {
+  type: string // a unique string that identifies a pipe
```
121watts, line 12:
```suggestion
export interface TypeRegistration<T> {
  type: T // a unique string that identifies a pipe
```

121watts: Would it be useful for the component registering this type to be able to define it as a `const` then the generic `string`?

ui/src/notebooks/pipes/Example/view.tsx
```diff
@@ -0,0 +1,31 @@
+import React, {FC, useContext} from 'react'
+import {NotebookContext} from 'src/notebooks/context/notebook'
+
+import Panel from 'src/notebooks/components/Panel'
+
+interface PipeProps {
+  idx: number
```
121watts, line 7: nit: any opposition to the word `index`? 😆

ui/src/notebooks/components/Notebook.tsx
drdelambre: lol i just switched all of these components from `NotebookComponent` to `Component` as there was a `src/notebooks/components/` + `NOTEBOOK EVERYTHING` folder that made it hard to find things locally. i can switch it back

ui/src/notebooks/components/Pipe.tsx
drdelambre: it will if the data schema provides an invalid type, which shows up as a hydration problem. currently you are correct as we are only adding things that are already registered, but i'm expecting that to change as we migrate to new types and storing this on the server, and tracking this error is how we'll signal that we need to have a more graceful migration strategy that doesn't just leave types on the floor.

ui/src/notebooks/components/PipeList.tsx
drdelambre: ITS THE FUTURE

ui/src/notebooks/index.ts
drdelambre: not here, but definitely in the serialization part where i currently have `type Pipe = any` with a todo to learn typescript. this part should have them all with the same shape to keep a consistent registration process, unless youre talking about some `as const` magic i'm unaware of

ui/src/notebooks/pipes/Example/view.tsx
drdelambre: think of my RSI bro

ui/src/notebooks/components/Panel.tsx
drdelambre: to capture the guard yeah?

ui/src/notebooks/components/Header.tsx
drdelambre: imma ignore this one as the next PR adds it right back

ui/src/notebooks/components/Panel.tsx
121watts: yup exactly

ui/src/notebooks/components/Panel.tsx
drdelambre: i tried this and it just turned into a bunch of prop drilling and code duplication for no tangible benefit. I'd even entertain doing it anyways if this component was meant to grow, but it's really just to inject arbitrary code into the header which should consolidate expansion of functionality into the child component that's instantiating this.

ui/src/notebooks/components/Panel.tsx
drdelambre: same as above

ui/src/notebooks/components/PipeList.tsx
drdelambre: nope.. not the future yet. tsc is giving me this:
```
Type '() => JSX.Element[]' is not assignable to type 'FC<{}>'
```

ui/src/notebooks/components/Notebook.tsx
drdelambre: k. this is gonna show up in a following PR after i get all this first wave out. just mapped out the sequencing

PR #18109, opened 2020-05-14
fix(ui): wait for org to be available then fetch org settings just once

### LINKED ISSUE #18102: OrgSettings should wait until there is an org
Although the code for fetching org settings catches errors, ideally, we want to wait until an org with `id` is available before attempting to fetch org settings.

TCL735:
Closes #18102 

For Cloud only.

Wait for an org to be loaded before attempting to fetch org settings.

### DIFF
--- a/ui/src/cloud/components/OrgSettings.tsx
+++ b/ui/src/cloud/components/OrgSettings.tsx
@@ -1,34 +1,56 @@
 // Libraries
-import {PureComponent} from 'react'
+import {FunctionComponent, useEffect, useState} from 'react'
 import {connect} from 'react-redux'
+import {get} from 'lodash'
 
 // Constants
 import {CLOUD} from 'src/shared/constants'
 
+// Types
+import {AppState, Organization} from 'src/types'
+
 // Actions
 import {getOrgSettings as getOrgSettingsAction} from 'src/cloud/actions/orgsettings'
 
+interface PassedInProps {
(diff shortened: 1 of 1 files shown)


ui/src/cloud/components/OrgSettings.tsx
```diff
@@ -1,34 +1,56 @@
...
+const OrgSettings: FunctionComponent<Props> = ({
+  org,
+  getOrgSettings,
+  children,
+}) => {
+  const [hasFetchedOrgSettings, setHasFetchedOrgSettings] = useState<boolean>(
+    false
+  )
+  useEffect(() => {
+    if (CLOUD && org && !hasFetchedOrgSettings) {
```
drdelambre, line 36: just flying through this, but shouldn't we be reading this off of the redux RemoteDataState?

TCL735: I tried that. RemoteDataState is `Done` before there is an org object. So that doesn't work for me.

TCL735: Specifically, `resources.orgs.status` is the RemoteDataState for all orgs. A specific org doesn't have a RemoteDataState. And because I need the `id` off of a specific org, I can't rely on `resources.orgs.status` as the signal.

drdelambre: lame

PR #18112, opened 2020-05-15
fix: need to rebuild the query request before the second e2e test run

adrian-thurston:
The e2e test driver in influxdb run the tests twice to get past the fact that there
is no way to force order between the write to storage and the read back. When
the json.Marshal call became mandatory it was added to the first run, but not
the second.


adrian-thurston:
> Yes, looks good. I'm sorry that this caused you trouble. Good catch.
> 
> For what it's worth, we have a new function called `wait` or something that is supposed to guarantee sequential execution of streams. I don't think we are using it in this framework yet, but we should be, it could simplify this code a lot.

@wolffcm I actually learned a lot about the system fixing this one ... concurrency model, results processing, e2e tests. I should be thanking you!

PR #18117, opened 2020-05-15
feat: adding basic notebook header

drdelambre:
Another installment of the saga to make #17966 more real. Here I flesh out the header component. Nothing too spectacular as most of this is extending off of ideas put forth in #18089 and the reusing of components from other parts of the codebase, but of note is that one of the contexts is now connected to redux!

Screen Shot 2020-05-15 at 11 32 03 AM

Kapture 2020-05-07 at 15 32 46


drdelambre:
sorry for the auto alerts everyone.. i dont rebase often, but when i do...

ui/src/notebooks/components/Header.tsx
```diff
@@ -1,13 +1,84 @@
...
+        },
+      } as TimeBlock)
+    } else {
+      updateTimeContext(id, {
+        refresh: {
+          status: AutoRefreshStatus.Active,
+          interval,
+        },
+      } as TimeBlock)
+    }
```
121watts, line 43:
```suggestion
     const status = interval === 0 ? Paused : Active
      updateTimeContext(id, {
        refresh: {
          status,
          interval,
        },
      } as TimeBlock)
 
```

ui/src/notebooks/context/app.tsx
```diff
@@ -0,0 +1,62 @@
...
+      }}
+    >
+      {children}
+    </AppSettingContext.Provider>
+  )
+}
+
+const mstp = (state: AppState): StateProps => {
+  return {
+    timeZone: state.app.persisted.timeZone || ('Local' as TimeZone),
```
121watts, line 51: lets hide this in a selector.

ui/src/notebooks/context/time.tsx
```diff
@@ -0,0 +1,88 @@
...
+        timeContext,
+        addTimeContext,
+        updateTimeContext,
+        removeTimeContext,
+      }}
+    >
+      {children}
+    </TimeContext.Provider>
+  )
+}
```
121watts, line 88: this whole file could use some unit tests.

ui/src/timeMachine/components/SubmitQueryButton.tsx
```diff
@@ -26,7 +26,7 @@ interface StateProps {
 }
 
 interface DispatchProps {
-  onSubmit: typeof saveAndExecuteQueries
+  onSubmit: typeof saveAndExecuteQueries | (() => void)
```
121watts, line 29: this temporary until `submit` is implemented?  if so, add a TODO?

drdelambre: no, this is because saveAndExecuteQueries _is_ `() => void`, but carries the context of the redux store along for the ride, specifically that redux is going to have a collection of queries ready in the store to be executed when this button is pressed. We're reusing this component, but without redux, and it a different context, so that's why this is being extended. same interface, just without the implicitness

PR #18141, opened 2020-05-18
feat(ui): polish notebook panels

alexpaxton:
Closes https://github.com/influxdata/idpe/issues/7142

- Copy over styles from notebook design prototyping branch
- Organize panel components into `/notebooks/components/panel`
- Break out `NotebookPanel` elements into individual components
- Move add buttons into `PageControlBar`

Screen Shot 2020-05-18 at 11 34 16 AM

- [x] Rebased/mergeable
- [x] Tests pass


ui/src/notebooks/pipes/Example/view.tsx
```diff
@@ -2,7 +2,7 @@ import React, {FC, useContext} from 'react'
 import {PipeProp} from 'src/notebooks'
 import {NotebookContext} from 'src/notebooks/context/notebook'
 
-import Panel from 'src/notebooks/components/Panel'
+import NotebookPanel from 'src/notebooks/components/panel/NotebookPanel'
```
drdelambre, line 5: why so much nesting?

alexpaxton: Just feels more organized to me, once there are more components in here (minimap, etc.) it will be easier to navigate

drdelambre: i get that, even for it, but i'm trying to think of a happy balance between that intent and ease of use. imagine each one of these pipes as being a self contained node module, and now each one needs to have a concept of the internal structure of the notebook component. Maybe the folder structure is fine as it allows the notebook structure to grow, just gotta figure out how to decouple the notebook structure from the pipe so that we can nest these types of things without changing all of the pipes. off the cuff i'm feeling out a structure like:

```
// in src/notebooks/index.ts
export {
  register,
  NotebookPanel
}
export default Notebook
```

half thoughts and maybe outside of this PR, but what about that as a dependency contract between notebook and pipes?

ui/src/notebooks/components/panel/NotebookPanel.tsx
```diff
@@ -0,0 +1,92 @@
...
+          {controlsLeft}
+        </FlexBox>
+        <FlexBox
+          className="notebook-panel--header-right"
+          alignItems={AlignItems.Center}
+          margin={ComponentSize.Small}
+          justifyContent={JustifyContent.FlexEnd}
+        >
+          {controlsRight}
+          <MovePanelUpButton onMoveUp={onMoveUp} id={id} />
```
drdelambre, line 76: lets change this to `<MovePanelUpButton onClick={ onMoveUp } />` and move the id / position binding to the PipeList component that passes that function into this so that we can keep the application logic and it's requirements centralized by it's manipulation and have the presentation just concerned about firing the event.

drdelambre: i guess that changes the function signature to `onMoveUp: () => void` since it's just signaling, same goes for onMoveDown and remove too

ui/src/notebooks/pipes/Example/view.tsx
alexpaxton: I don't have much of an opinion on how importing / exporting happens

alexpaxton:
- Consolidated `MovePanelUpButton` and `MovePanelDownButton` into a single component because the code is 99% the same
- Remove `id` prop from `RemovePanelButton` and `MovePanelButton`
- Remove `usePanelState` hook
  - Only used by 1 component and don't anticipate that state pattern being re-used elsewhere
- Import and export `NotebookPanel` from `Notebook`
- Change `View` to import `NotebookPanel` from `Notebook`

Issue #17931, opened 2020-05-01
Dashboard list takes a long time to load with a large number of dashboard


desa:
![dashboard-take-long-time.gif](https://images.zenhubusercontent.com/5d07f2d52ee3440ccd748a4e/cd0efd55-a0e6-416d-8d94-38845236a840)

drdelambre:
seems like pagination got removed with normalization: https://github.com/influxdata/influxdb/pull/16535

drdelambre:
scratch that. this never worked on an API level, just on a rendering level, and it seems like that's because the API level wasn't the bottleneck. There are some issues with over rendering and some lifecycle methods being messed up, but fixing those had negligible impact on rendering performance. Biggest improvement to performance has come by removing the `InlineLabels` component from the `DashboardCard` component. Still don't know what that means...

drdelambre:
this change effectively disables the pagination: https://github.com/influxdata/influxdb/pull/17399

drdelambre:
@alexpaxton why do dashboards get 3x faster when i remove this component?
https://github.com/influxdata/influxdb/blob/master/ui/src/dashboards/components/dashboard_index/DashboardCard.tsx#L84

drdelambre:
fixed with #18142

PR #18142, opened 2020-05-18
fix: faster dashboards

drdelambre:
hopefully addresses #17931 and influxdata/idpe#7172

super hard to detangle, but it seems to have been a combination of rendering pagination being turned off for an EAR and a thunk blocking the ui thread when trying to hydrate extra data.

Moved the follow up thunk actions further down the event loop with `setImmediate` to unblock the ui thread, effectively delaying view / cell hydration until after the dashboard list is populated (should be ok as these aren't accessed until the dashboard is opened), as well as turned pagination back on to reduce rendering of nested resources, but this time with a fixed window size instead of using react to measure the number of dashboards that are visible on the page. The EAR was an issue with a small number of dashboards not registering a full window size in firefox during this measuring cycle, so i just stopped measuring.

### DIFF
--- a/ui/src/dashboards/actions/thunks.ts
+++ b/ui/src/dashboards/actions/thunks.ts
@@ -246,20 +246,25 @@ export const getDashboards = () => async (
         }
       })
       .forEach(entity => {
-        const viewsData = viewsFromCells(entity.cells, entity.id)
+        setImmediate(() => {
+          const viewsData = viewsFromCells(entity.cells, entity.id)
 
-        const normViews = normalize<View, ViewEntities, string[]>(
-          viewsData,
-          arrayOfViews
-        )
+          const normViews = normalize<View, ViewEntities, string[]>(
+            viewsData,
+            arrayOfViews
+          )
 
-        const normCells = normalize<Dashboard, DashboardEntities, string[]>(
-          entity.cells,
-          arrayOfCells
-        )
+          dispatch(setViews(RemoteDataState.Done, normViews))
+        })
 
-        dispatch(setViews(RemoteDataState.Done, normViews))
-        dispatch(setCells(entity.id, RemoteDataState.Done, normCells))
+        setImmediate(() => {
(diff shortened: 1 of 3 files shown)


hoorayimhelping:
https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate

>This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

>This method is not expected to become standard, and is only implemented by recent builds of Internet Explorer and Node.js 0.10+. It meets resistance both from Gecko (Firefox) and Webkit (Google/Apple).

this big red warning box on this page saying don't use this in production is kind of persuasive. Is this the best idea?

drdelambre:
we have it polyfilled to an equivalent `setTimeout(() => { }, 0)`, but using the default one when available makes the whole thing much faster as setTimeout of 0 is actually 50ms or something

ui/src/dashboards/actions/thunks.ts
```diff
@@ -246,20 +246,25 @@ export const getDashboards = () => async (
...
       .forEach(entity => {
-        const viewsData = viewsFromCells(entity.cells, entity.id)
+        setImmediate(() => {
+          const viewsData = viewsFromCells(entity.cells, entity.id)
 
-        const normViews = normalize<View, ViewEntities, string[]>(
-          viewsData,
-          arrayOfViews
-        )
+          const normViews = normalize<View, ViewEntities, string[]>(
```
glinton, line 252: nit: convert to tabs (or spaces) :wink: :rofl:

drdelambre:
that's good feedback for issues like these. idk how to do a head to head gif battle, but a before and after picture of the audit comparing the flame graph peaks i was using to narrow this down might be nice as it verifies testing criteria and the effect of the change. e2e tests already blew away my test data for this issue, which is building ~60 dashboards with unique properties, so i'm gonna respectfully decline on this specific one, but noting it down as a best practice for future explorations.

PR #18144, opened 2020-05-18
feat(query): add a planner rule to push down bare aggregates

wolffcm:
This adds a rule to push down aggregates that are directly over a `ReadRange` source.

I followed the existing example of `PushDownWindowAggregateRule` and did some refactoring to avoid duplicating logic.

I'll follow up with a pipeline test and a Flux end-to-end test.

### DIFF
--- a/flags.yml
+++ b/flags.yml
@@ -28,13 +28,13 @@
   expose: true
 
 - name: Push Down Window Aggregate Count
-  description: Enable Count variant of PushDownWindowAggregateRule
+  description: Enable Count variant of PushDownWindowAggregateRule and PushDownBareAggregateRule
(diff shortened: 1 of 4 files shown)


wolffcm:
@adrian-thurston 
> So with this change the feature flags will cover both window aggregate and bare aggregate rules.

Hmm, I hadn't really thought that through. I was thinking of the feature flag as guarding the storage capability, but I guess that's not the case. I guess the question comes down to: is it possible that we introduce a regression with the bare-agg rule but not the window-agg rule? It seems within the realm of possibility.

I would think that we'd want to avoid an explosion of feature flags though. I'm curious about @yzhang1991 thoughts on this.

PR #18145, opened 2020-05-18
feat: moving around

drdelambre:
hopefully this turns moving thing around on

### DIFF
--- a/ui/src/notebooks/components/AddButtons.tsx
+++ b/ui/src/notebooks/components/AddButtons.tsx
@@ -17,8 +17,12 @@ const AddButtons: FC = () => {
         key={def.type}
         text={def.button}
         onClick={() => {
+          let data = def.empty
+          if (typeof data === 'function') {
+            data = data()
+          }
           addPipe({
-            ...def.empty,
+            ...data,
             type,
           })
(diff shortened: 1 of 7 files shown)


ui/src/notebooks/components/AddButtons.tsx
```diff
@@ -17,8 +17,12 @@ const AddButtons: FC = () => {
         key={def.type}
         text={def.button}
         onClick={() => {
+          let data = def.empty
+          if (typeof data === 'function') {
```
drdelambre, line 21: this lets us treat the empty state as a generator

ui/src/notebooks/context/notebook.tsx
```diff
@@ -55,6 +55,22 @@ export const NotebookProvider: FC = ({children}) => {
     })
   }
 
+  function movePipe(currentIdx: number, newIdx: number) {
+    setPipes(pipes => {
+      const idx = ((newIdx % pipes.length) + pipes.length) % pipes.length
```
drdelambre, line 60: all of the mods allow us to be lazy with math later on by ensuring the new index is within the bounds of the pipes array

PR #18150, opened 2020-05-18
Revert "refactor: remove fancy scroll bars"

desa:
Reverts influxdata/influxdb#18016

fixes a broken interaction in tools: 
ScrollbarsNotMoving


desa:
fixed by https://github.com/influxdata/influxdb/pull/18151

PR #18154, opened 2020-05-19
fix: Update CI configuration files

ayang64:
This patch updates the Circle CI and Jenkins related configruation files
to allow automated building and testing with Go 1.13.

This also removes any remaning mentions of Go 1.12.

- [x] Rebased/mergeable
- [x] Tests pass


ayang64:
@gunnaraasen, @jacobmarble -- i tried my best to update the CI related configuration but it seems like it takes a VERY long time for jenkins runs to complete.

any help you can offer would be greatly appreciated.

test.sh
```diff
@@ -123,14 +123,9 @@ case $ENVIRONMENT_INDEX in
...
         ;;
-    4)
-        # go1.12
-        run_test_docker Dockerfile_build_ubuntu64_go1.12 test_64bit --test --junit-report
-        rc=$?
-        ;;
-    "count")
-        echo $ENV_COUNT
-        ;;
+		"count")
```
jacobmarble, line 126: Consistent indentation would be nice here. Looks like tabs-vs-spaces.

ayang64: thanks!

gunnaraasen, line 126: Looks like this is still set to tabs while the rest of this file uses spaces.

ayang64: actually fixed this time.

PR #18158, opened 2020-05-19
fix(tsdb): Disable series id set cache size by default.

benbjohnson:
This commit changes `DefaultSeriesIDSetCacheSize` to zero so that the tag value cache is disabled by default. There is a rare known bug where the cache can cause a segfault which crashes the process. The cache is being disabled instead of removed as some users may still need the cache for performance reasons.

Closes: #14455, #12137


jacobmarble:
@dgnorton FYI

PR #18163, opened 2020-05-19
feat: keep notebook meta out of the pipes

drdelambre:
Kapture 2020-05-19 at 10 24 06

this keeps the notebook metadata (of known type) away from the Pipe data (of unknown type) so that we can build better visualizations around it. code kind of smells when we're keeping two arrays in sync and not keeping order separate from a lookup table, but this is just a demo. Also removed the concept of a NotebookPanel from a Pipe type and a bunch of "future interfaces", not because i have any feelings about then, but because they need to be redone with the context

### DIFF
--- a/ui/src/notebooks/index.ts
+++ b/ui/src/notebooks/index.ts
@@ -2,9 +2,6 @@ import {FunctionComponent, ComponentClass} from 'react'
 
 export interface PipeProp {
   index: number
-  remove: () => void
-  moveUp: () => void
-  moveDown: () => void
 }
 
 // NOTE: keep this interface as small as possible and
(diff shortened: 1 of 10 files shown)


ui/src/notebooks/index.ts
```diff
@@ -1,10 +1,17 @@
...
+}
+
+export type PipeData = any
 
 export interface PipeProp {
-  index: number
-  remove: () => void
-  moveUp: () => void
-  moveDown: () => void
+  data: PipeData
```
drdelambre, line 10: i disconnected the relationship between pipes and notebook context here, so beware

PR #18167, opened 2020-05-19
feat(unify): user and invite contract

121watts:
### What
Proposed contract for the unification API

### Why
So we can start pulling frontend assets over to OSS from Quartz with less headaches.

### Notes
Not a huge fan of the name `CloudUser` but `User` was already taken.  Open to suggestions.  Used [these client calls](https://github.com/influxdata/quartz/blob/master/assets/js/client/index.ts) as inspiration.

Apologies for the huge diff but the inconsistencies in quotations and spacing was bothering me.

The relevant chunks are:
https://github.com/influxdata/influxdb/pull/18167/files#diff-33aa613ab26f77a960b47c92852150afR4467

https://github.com/influxdata/influxdb/pull/18167/files#diff-33aa613ab26f77a960b47c92852150afR8602

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -12,17 +12,17 @@ paths:
       security:
         - BasicAuth: []
       parameters:
-        - $ref: '#/components/parameters/TraceSpan'
+        - $ref: "#/components/parameters/TraceSpan"
       responses:
-        '204':
+        "204":
           description: Successfully authenticated
-        '401':
+        "401":
           description: Unauthorized access
           content:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
-        '403':
+        "403":
           description: user account is disabled
           content:
             application/json:
@@ -39,11 +39,11 @@ paths:
       operationId: PostSignout
       summary: Expire the current session
       parameters:
-        - $ref: '#/components/parameters/TraceSpan'
+        - $ref: "#/components/parameters/TraceSpan"
       responses:
-        '204':
+        "204":
           description: Session successfully expired
-        '401':
+        "401":
           description: Unauthorized access
           content:
             application/json:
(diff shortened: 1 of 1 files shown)


http/swagger.yml
```diff
@@ -4465,6 +4465,161 @@ paths:
...
+      tags:
+        - Invites
+        - Organizations
+      summary: Remove an invite to an organization
+      parameters:
+        - $ref: '#/components/parameters/TraceSpan'
+        - in: path
+          name: inviteID
+          schema:
+            type: string
```
randycoulman, line 4515: Invite IDs are technically numbers, I think, but since it's part of the URL, it probably doesn't matter here, does it?

http/swagger.yml
```diff
@@ -8445,6 +8600,94 @@ components:
...
+          type: string
+          enum:
+            - member
+            - owner
+        withdrawnAt:
+          format: date-time
+          type: string
+        acceptedAt:
+          format: date-time
+          type: string
```
randycoulman, line 8621: We don't want to expose `withdrawnAt` and `acceptedAt` here; they're internal implementation details of the DB schema and shouldn't be part of the API.

http/swagger.yml
```diff
@@ -8445,6 +8600,94 @@ components:
...
+    Invites:
+      type: object
+      properties:
+        links:
+          type: object
+          properties:
+            self:
+              type: string
+              format: uri
+        users:
```
randycoulman, line 8644: Should this key name be `invites`?

http/swagger.yml
```diff
@@ -8445,6 +8600,94 @@ components:
...
+          type: array
+          items:
+            $ref: "#/components/schemas/Invites"
```
randycoulman, line 8647: Should this be `Invite` instead of `Invites`?

http/swagger.yml
```diff
@@ -8445,6 +8600,94 @@ components:
...
+    CloudUser:
+      properties:
+        id:
+          description: the idpe id of the user
+          readOnly: true
+          type: string
+        cloudID:
+          description: the cloud id of the user
+          type: string
```
randycoulman, line 8656: Do we need to expose `cloudID` at all?

121watts:
thanks for catching those!

PR #18168, opened 2020-05-19
feat: add query type for notebooks

drdelambre:
Kapture 2020-05-20 at 13 49 11

adds a timemachine style flux editor to the notebook.

some immediate observations is that we can't just reuse the current flux editor (the fixed height sucks up all the scroll events, and to remove the fixed height means we have to address the "functions/variables" interface), and that we need to jump to the end on every add

[ CURRENTLY LOOKING AT PERF ISSUES ]


ui/src/notebooks/context/notebook.tsx
```diff
@@ -125,7 +145,7 @@ export const NotebookProvider: FC = ({children}) => {
         removePipe,
       }}
     >
-      {children}
+      {chidlin}
```
alexpaxton, line 148: Hmmm

PR #18170, opened 2020-05-19
feat(query): add conditional option to LoggingProxyQueryService

wolffcm:
This updates `LoggingProxyQueryService` to only conditionally log queries, based on a condition function passed in when the service is created.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/dgryski/go-bitstream v0.0.0-20180413035011-3522498ce2c8
-	github.com/docker/docker v1.13.1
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -20,7 +20,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/dgryski/go-bitstream v0.0.0-20180413035011-3522498ce2c8
-	github.com/docker/docker v1.13.1 // indirect
+	github.com/docker/docker v1.13.1
```
nathanielc, line 23: any ideas why this changed?

wolffcm: I have no idea. It was failing in CI, so I ran `go mod tidy` and it added that comment, and now CI passes.

wolffcm: Oh, it's because my IDE chose the `ioutils` package from Docker. I wanted the Go stdlib one. I fixed it.

PR #18173, opened 2020-05-20
fix(api/health): report version

### LINKED ISSUE #13907: InfluxDB 2.0: There is now way how to get version of InfluxDB from API [area/api]
The InfluxDB 1.7.x has defined [`/ping`](https://docs.influxdata.com/influxdb/v1.7/tools/api/#ping-http-endpoint) endpoint with `X-Influxdb-Version` header but in the API 2.0 there is no such thing.

The `/health` endpoint will be useful for that.

sranka:
Closes #13907 

InfluxDB 1.8 "/health" endpoint already returns version, 
```js
{
"checks": [],
"message": "ready for queries and writes",
"name": "influxdb",
"status": "pass",
"version": "1.8.0"
}
```
This PR also adds version to 2.0 "/health" endpoint

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/health.go
+++ b/http/health.go
@@ -3,11 +3,13 @@ package http
 import (
 	"fmt"
 	"net/http"
+
+	platform "github.com/influxdata/influxdb/v2"
 )
 
 // HealthHandler returns the status of the process.
 func HealthHandler(w http.ResponseWriter, r *http.Request) {
-	msg := `{"name":"influxdb", "message":"ready for queries and writes", "status":"pass", "checks":[]}`
+	msg := fmt.Sprintf(`{"name":"influxdb", "message":"ready for queries and writes", "status":"pass", "checks":[], "version": %q}`, platform.GetBuildInfo().Version)
 	w.Header().Set("Content-Type", "application/json; charset=utf-8")
 	w.WriteHeader(http.StatusOK)
 	fmt.Fprintln(w, msg)
--- a/http/health.go
+++ b/http/health.go
@@ -9,7 +9,7 @@ import (
 
 // HealthHandler returns the status of the process.
 func HealthHandler(w http.ResponseWriter, r *http.Request) {
-	msg := fmt.Sprintf(`{"name":"influxdb", "message":"ready for queries and writes", "status":"pass", "checks":[], "version": %q}`, platform.GetBuildInfo().Version)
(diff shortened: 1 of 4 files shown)


http/health_test.go
```diff
@@ -26,28 +27,41 @@ func TestHealthHandler(t *testing.T) {
...
+				t.Errorf("%q. HealthHandler() = %v, want %v", tt.name, contentType, tt.wants.contentType)
 			}
-			if tt.wants.body != "" {
-				if eq, diff, err := jsonEqual(string(body), tt.wants.body); err != nil {
-					t.Errorf("%q, HealthHandler(). error unmarshaling json %v", tt.name, err)
-				} else if !eq {
-					t.Errorf("%q. HealthHandler() = ***%s***", tt.name, diff)
+			var content map[string]interface{}
+			if err := json.Unmarshal(body, &content); err != nil {
+				t.Errorf("%q, HealthHandler(). error unmarshaling json %v", tt.name, err)
```
glinton, line 49: Maybe add a `continue` after this so you can remove the following `else`. More of a nit than request.

glinton:
Personally, I wouldn't mind seeing the commit in there too.

http/health_test.go
glinton: The upside of `continue` over `return` or `t.Fatalf` is that the other tests would continue. I guess it doesn't matter here since there is only one test in the slice.

sranka: **return** from t.Run handler is the way to avoid else branching herein ... repaired.

http/health.go
```diff
@@ -3,11 +3,13 @@ package http
...
 	"fmt"
 	"net/http"
+
+	platform "github.com/influxdata/influxdb/v2"
 )
 
 // HealthHandler returns the status of the process.
 func HealthHandler(w http.ResponseWriter, r *http.Request) {
-	msg := `{"name":"influxdb", "message":"ready for queries and writes", "status":"pass", "checks":[]}`
+	msg := fmt.Sprintf(`{"name":"influxdb", "message":"ready for queries and writes", "status":"pass", "checks":[], "version": %q}`, platform.GetBuildInfo().Version)
```
glinton, line 12: The more i think about it, the more valuable i feel like `commit` would be in here. (`platform.GetBuildInfo().Commit`) :man_shrugging:

sranka: A new "commit" field, I can add it add well

PR #18205, opened 2020-05-22
feat: allow monaco to autogrow

drdelambre:
Kapture 2020-05-22 at 17 26 18

This allows monaco to grow to whatever hight it wants to fill the content, allowing interfaces that grow vertically. Removes minimap and the scrollbar as well, as they dont make sense when you can't scroll

### DIFF
--- a/ui/src/external/monaco.flux.autogrow.ts
+++ b/ui/src/external/monaco.flux.autogrow.ts
@@ -10,7 +10,9 @@ export const registerAutogrow = editor => {
       return
     }
 
-    const lineHeight = editor.getOption(window.monaco.editor.EditorOption.lineHeight)
+    const lineHeight = editor.getOption(
+      window.monaco.editor.EditorOption.lineHeight
+    )
     const lineCount = (editor.getModel() || {}).getLineCount() || 1
     const height = editor.getTopForLineNumber(lineCount + 1) + lineHeight
 
--- a/ui/src/external/monaco.flux.autogrow.ts
+++ b/ui/src/external/monaco.flux.autogrow.ts
@@ -10,7 +10,7 @@ export const registerAutogrow = editor => {
       return
     }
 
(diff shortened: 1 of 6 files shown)


ui/src/external/monaco.flux.autogrow.ts
```diff
@@ -0,0 +1,39 @@
+const MIN_HEIGHT = 100
```
alexpaxton, line 1: Why make this global?

drdelambre: not global, it's scoped to the module instance. usually if you chose an arbitrary number in your code, you can expect it to change for arbitrary reasons. pulling it out as a constant at the top of a module signals "this is a compile time configuration" and makes it easier to change when the time comes without searching through all the app logic thinking "which one of these `100`s corresponded to that change?" and consolidating them in to logical groups if you have multiple instances

ui/src/external/monaco.flux.autogrow.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    const lineHeight = editor.getOption(
+      window.monaco.editor.EditorOption.lineHeight
+    )
+    const lineCount = (editor.getModel() || {}).getLineCount() || 1
+    const height = editor.getTopForLineNumber(lineCount + 1) + lineHeight
+
+    if (prevHeight !== Math.max(MIN_HEIGHT, height)) {
+      prevHeight = Math.max(MIN_HEIGHT, height)
+      editorElement.style.height = `${height}px`
```
Asalem1, line 21: should this be setting the height to the reassigned `prevHeight`? It doesn't seem like we're using `prevHeight` anywhere else

drdelambre: you're probably right. lemme double check

PR #18223, opened 2020-05-26
chore: cleaning up code styling

drdelambre:
in a mad dash to get rendering of pipe types to work, some corners were cut. hopefully this rights some wrongs


ui/src/notebooks/components/NotebookPipe.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+import {PipeContextProps, PipeData} from 'src/notebooks'
+import Pipe from 'src/notebooks/components/Pipe'
+import NotebookPanel from 'src/notebooks/components/panel/NotebookPanel'
+
+export interface NotebookPipeProps {
+  index: number
+  data: PipeData
+  onUpdate: (index: number, pipe: PipeData) => void
+}
```
alexpaxton, line 11: More of a pondering than any meaningful feedback but makes me wonder if we should have a `notebooks/types` directory to centralize these.

drdelambre: i've been tossing that around too. i like how the external facing types are on `src/notebooks/index.ts` but the internal types (anything inside of components and context) are kind just jumbled

PR #18252, opened 2020-05-27
chore(http): remove swagger def of invites and cloud users

desa:


russorat:
@desa let's merge this

wbaker85:
I'm closing this since the swagger file is no longer contained in the `influxdb` repo, and is now in the `influxdata/openapi` repo. It looks like these endpoints already aren't present in the `openapi` contract, so this doesn't appear to be needed there!

PR #18253, opened 2020-05-27
docs: update readme for new flux building instructions

jsternberg:


README.md
```diff
@@ -84,6 +84,33 @@ Logs to stdout by default:
...
+```bash
+$ ./env go build ./cmd/influxd
+```
+
+Another method is to configure the `pkg-config` utility. Follow the instructions [here](https://github.com/influxdata/flux#getting-started) to install and configure `pkg-config` and then the normal go commands will work.
+
+```bash
+# On Debian/Ubuntu
+$ sudo apt-get install -y clang pkg-config
+# On Mac OS X with Homebrew
```
gunnaraasen, line 102: I think breaking this block up into several code blocks will make the steps easier to follow.

PR #18254, opened 2020-05-27
fix(query): Disable compiler's Fused Multiply Add to fix test failure in TestHoltWinters_AusTourists

vivkong:
2 tests are failing (`TestHoltWinters_AusTourists` and `TestSelect/HoltWinters_GroupBy_Agg`) on s390x architecture.  This is due to the compiler's Fused Multiply Add (FMA) instruction selection which can result in different rounding errors in floating point computations.  Proposing to use float64() expression explicitly to disable the compiler's FMA instruction selection.  This will not affect architectures that does not have FMA (e.g. amd64).

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/query/functions.go
+++ b/query/functions.go
@@ -1432,11 +1432,13 @@ func (r *FloatHoltWintersReducer) Emit() []FloatPoint {
 }
 
 // Using the recursive relations compute the next values
+// Use float64() expression explicitly to disable the compiler's Fused Multiply Add (FMA) instruction selection
+// which can result in different rounding errors in floating point computations.
 func (r *FloatHoltWintersReducer) next(alpha, beta, gamma, phi, phiH, yT, lTp, bTp, sTm, sTmh float64) (yTh, lT, bT, sT float64) {
-	lT = alpha*(yT/sTm) + (1-alpha)*(lTp+phi*bTp)
-	bT = beta*(lT-lTp) + (1-beta)*phi*bTp
-	sT = gamma*(yT/(lTp+phi*bTp)) + (1-gamma)*sTm
-	yTh = (lT + phiH*bT) * sTmh
+	lT = float64(alpha*(yT/sTm)) + float64((1-alpha)*(lTp+phi*bTp))
+	bT = float64(beta*(lT-lTp)) + float64((1-beta)*phi*bTp)
+	sT = float64(gamma*(yT/(lTp+phi*bTp))) + float64((1-gamma)*sTm)
+	yTh = float64((lT + phiH*bT) * sTmh)
 	return
 }
 


vivkong:
Hello, I'm looking for feedback for this PR.  I've fixed the commit title but I'm not sure how to trigger the Semantic Pull Request check again to fix the failed check.  Thanks.

dgnorton:
@vivkong Thanks for the PR. The Semantic pull request check is failing for the commit and PR titles. It requires a space after the colon (e.g., `fix(query): Disable` instead of `fix(query):Disable`). I fixed the PR title. Try changing the commit title in the same way and doing a force push to the branch.

vivkong:
@dgnorton Thanks for the suggestion.  I've fixed the commit title and I can see the semantic check is now passing.  Thanks!

vivkong:
Hi, I'd like to request a review of this PR.  Thanks.

psteinbachs:
@dgnorton Please re-review when you get a chance.

vivkong:
Hello, can I please get a review for this PR?  Thanks.

psteinbachs:
@lesam Please review when you get a chance.

lesam:
Hi - thanks for the contribution!

From some quick checking, it looks like newer processors generally do support FMA3 at least.

Could you clarify what you mean by 'wrong' rounding? I would have expected FMA to actually provide a more numerically accurate result - are the results you are seeing with FMA wors?

Finally, since the goal here is to force the compiler to pick a different hardware instruction, would you be able to provide benchmarking on the hardware where you are seeing a difference?

dgnorton:
@lesam another option to consider might be build tags for platform-specific code. E.g., https://github.com/influxdata/influxdb/blob/v1.9.3/pkg/mmap/mmap_unix.go#L1. Note there are several versions of `mmap_*.go` where `*` is a specific platform. Blog post about it:  https://dave.cheney.net/2013/10/12/how-to-use-conditional-compilation-with-the-go-build-tool

vivkong:
Hello, thanks for reviewing.

There's actually nothing wrong - I probably didn't phrase it correctly.  The test failures on s390x was due to the differences in floating point rounding on different architectures.  The results on s390x are different from the expected result since the expected values are for amd64.  My attempt to allow tests to pass was to disable FMA as described in the golang reference (https://golang.org/ref/spec#Floating_point_operators).  

In case my proposed change might break other architectures, I can condition the change for s390x only as suggested by @dgnorton ?

lesam:
If the results are more accurate on the s390x, it might make more sense to condition a test change for s390x - or does that touch too many tests?

I feel we should prioritise numerical accuracy (and I believe FMA should be more accurate) over exactly identical results on all architectures.

vivkong:
> If the results are more accurate on the s390x, it might make more sense to condition a test change for s390x - or does that touch too many tests?

OK Sounds good.  I can update the PR for this.  It shouldn't touch too many tests I don't think.

lesam:
closing this PR in favour of PR from @danxmoran above

danxmoran:
Opened #22419 for the 1.x side of things

PR #18257, opened 2020-05-28
feat: allowing notebooks to submit queries

drdelambre:
Adding in results to the notebook context so that pipes can pull from them!
also some monaco work to make things render properly


ui/src/shared/components/MarkdownMonacoEditor.tsx
```diff
@@ -6,7 +6,7 @@ import MonacoEditor from 'react-monaco-editor'
 
 // Utils
 import LANGID from 'src/external/monaco.markdown.syntax'
-import THEME_NAME from 'src/external/monaco.markdown.theme'
+import THEME_NAME from 'src/external/monaco.flux.theme'
```
drdelambre, line 9: fun fact: monaco doesn't support multiple themes!
https://github.com/microsoft/monaco-editor/issues/338

it's in this PR because i couldn't tell if my `result` context changes were causing the editor to lose reference with monaco because of some rerendering bug or if it's the theme's problem itself. i imported the middleware that handles the tokenizing into our codebase so that we can play with prepending TokenizerState tags with language ids as we start to flesh out theming of markdown vs flux and run into collisions.

A PR should be floating through in the next couple days to harden that theme <> onigasm interface a bit more and include the tomls.

ui/src/notebooks/components/header/Submit.tsx
```diff
@@ -0,0 +1,86 @@
...
+import QueryProvider, {QueryContext} from 'src/notebooks/context/query'
+import {NotebookContext, PipeMeta} from 'src/notebooks/context/notebook'
+import {TimeContext} from 'src/notebooks/context/time'
+import {IconFont} from '@influxdata/clockface'
+
+// Types
+import {RemoteDataState} from 'src/types'
+
+const PREVIOUS_REGEXP = /__PREVIOUS_RESULT__/g
+const COMMENT_REMOVER = /(\/\*([\s\S]*?)\*\/)|(\/\/(.*)$)/gm
```
alexpaxton, line 13: NICE

ui/src/notebooks/context/notebook.tsx
```diff
@@ -1,19 +1,24 @@
 import React, {FC, useState, useCallback} from 'react'
 import {PipeData} from 'src/notebooks'
+import {FromFluxResult} from '@influxdata/giraffe'
 
 export interface PipeMeta {
   title: string
   visible: boolean
+  loading: boolean
```
alexpaxton, line 8: nice nice, probs fine for now but makes me wonder if this should be `RemoteDataState` or something like it?

drdelambre: should most def do that in here. oversight as i was running between things

PR #18268, opened 2020-05-28
chore(http): Remove Algorithm W proxying behavior since that's merged

brettbuddin:
This removes the temporary proxying behavior installed a few weeks ago. Algorithm W has been merged into `master` so this is no longer necessary.


brettbuddin:
It's likely we might need this behavior in some form in the coming weeks with some further Algorithm-W behavior that's not easily testable post-build. I'm going to keep these around for a bit longer until we get over that next hump. I will likely rename these structures away from `AlgoW` nomenclature and into something like `QueryBeta` to indicate "future" query behavior.

PR #18270, opened 2020-05-28
feat(notebooks): introduce minimap

alexpaxton:
Closes https://github.com/influxdata/idpe/issues/7144

- Introduce minimap component
  - can be toggled on/off from page header
  - on/off state stored in `AppSettingsContext`
- Clicking on a minimap item scrolls to that panel in the main list
- Minimap mirrors state of panels
  - Focus & visibility reflected in minimap
- Added a couple items in `NotebookContext` to facilitate the minimap and pipe list talking to each other

jump-to-pipe

- [x] Rebased/mergeable
- [x] Tests pass


alexpaxton:
@drdelambre hope I'm not mucking up notebook context, open to suggestions here

Also, because the toggle state of minimap is stored in `AppSettingsContext` (which isn't currently wrapping anything) the toggling behavior is currently not working

ui/src/notebooks/components/panel/NotebookPanel.tsx
```diff
@@ -15,6 +15,9 @@ import PanelVisibilityToggle from 'src/notebooks/components/panel/PanelVisibilit
 import MovePanelButton from 'src/notebooks/components/panel/MovePanelButton'
 import NotebookPanelTitle from 'src/notebooks/components/panel/NotebookPanelTitle'
 
+// Constants
+import {NOTEBOOK_PANEL_ID} from 'src/notebooks/constants'
+
```
alexpaxton, line 20: I made this a constant cuz I am using it in 2 places and I didn't want them to drift apart

ui/src/notebooks/context/notebook.tsx
```diff
@@ -11,22 +14,26 @@ export interface NotebookContextType {
   id: string
   pipes: PipeData[]
   meta: PipeMeta[] // data only used for the view layer for Notebooks
+  listScrollPosition: number
   addPipe: (pipe: PipeData) => void
   updatePipe: (idx: number, pipe: PipeData) => void
   updateMeta: (idx: number, pipe: PipeMeta) => void
   movePipe: (currentIdx: number, newIdx: number) => void
   removePipe: (idx: number) => void
+  scrollToPipe: (idx: number) => void
```
alexpaxton, line 23: I could see grouping these into a single field for readability, but more a matter of opinion

```tsx
scrolling: {
  listScrollPosition: number
  scrollToPipe: (idx: number) => void
},
```

ui/src/notebooks/context/notebook.tsx
```diff
@@ -131,6 +141,19 @@ export const NotebookProvider: FC = ({children}) => {
     [id]
   )
 
+  const scrollToPipe = useCallback(
+    (idx: number) => {
+      const targetPipe = document.getElementById(`${NOTEBOOK_PANEL_ID}${idx}`)
```
alexpaxton, line 146: Yeehaw vanilla JS! `ref` is probably the more reacty way to do this but also would require me changing more files

ui/src/notebooks/context/app.tsx
```diff
@@ -32,11 +36,20 @@ export const AppSettingContext = React.createContext<AppSettingContextType>(
 
 export const AppSettingProvider: FC<Props> = React.memo(
   ({timeZone, onSetTimeZone, children}) => {
+    const [miniMapVisibility, setMiniMapVisibility] = useState<boolean>(false)
+
+    const onToggleMiniMap = (): void => {
+      console.log('boop')
```
drdelambre, line 42: GASP

ui/src/notebooks/context/app.tsx
```diff
@@ -19,11 +19,15 @@ type Modifier = typeof setTimeZone
...
   onSetTimeZone: Modifier
+  miniMapVisibility: boolean
+  onToggleMiniMap: () => void
 }
 
 export const DEFAULT_CONTEXT: AppSettingContextType = {
   timeZone: 'Local' as TimeZone,
   onSetTimeZone: ((() => {}) as any) as Modifier,
+  miniMapVisibility: true,
+  onToggleMiniMap: () => {},
```
drdelambre, line 30: can we change this to a plain set? i can imagine that we're gonna want to change this programmatically and calling `onSetMiniMap(!miniMapVisibility)` for this case is easier than `miniMapVisibility===indendedValue?noop():onToggleMiniMap()` for the other cases

ui/src/notebooks/context/notebook.tsx
drdelambre: lol we should totally change more files. this has an implicit coupling with a presentation layer that is defined somewhere else. Change one, and you have to somehow remember to change the other. Being explicit at least draws a line representing that connection in the code

ui/src/notebooks/context/notebook.tsx
```diff
@@ -131,6 +141,19 @@ export const NotebookProvider: FC = ({children}) => {
...
+
+      if (targetPipe) {
+        const {offsetTop} = targetPipe
+        console.log('scrollToPipe', offsetTop)
```
drdelambre, line 150: guuuuuuurrrrrl

ui/src/notebooks/context/notebook.tsx
drdelambre: instead of a single field, it might make more sense to put them in their own context? it feels like a very different _kind_ of data on the notebook. I got the feeling that pipes and meta should be split into their own contexts when i started adding `results` to it using the same pattern, but they're coupled on array index for now and have to be moved together until we can break them apart

ui/src/notebooks/context/app.tsx
```diff
@@ -32,11 +36,20 @@ export const AppSettingContext = React.createContext<AppSettingContextType>(
 
 export const AppSettingProvider: FC<Props> = React.memo(
   ({timeZone, onSetTimeZone, children}) => {
+    const [miniMapVisibility, setMiniMapVisibility] = useState<boolean>(false)
```
drdelambre, line 39: feel free to add to the app setting reducer here:
https://github.com/influxdata/influxdb/blob/master/ui/src/shared/reducers/app.ts

seems fitting

ui/src/shared/reducers/app.ts
```diff
@@ -28,6 +29,7 @@ const initialState: AppState = {
     showTemplateControlBar: false,
     timeZone: 'Local',
     navBarState: 'collapsed',
+    notebookMiniMapState: 'expanded',
```
drdelambre, line 32: dope

alexpaxton: I learned from @121watts to use more descriptive values instead of booleans

PR #18280, opened 2020-05-29
fix(dshell): update Dockerfile for Flux dependency on pkg-config

rawkode:
Flux currently needs a wrapper for pkg-config to be available to build,
which directly affects running or building influxd.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,15 +18,21 @@ RUN apt install --yes \
         llvm-dev \
         make \
         nodejs \
+        pkg-config \
         protobuf-compiler \
         ragel \
         rustc \
         yarn
 
+ENV GOPATH=/go
+ENV PATH /go/bin:$PATH
+RUN go get github.com/influxdata/pkg-config
+
 FROM dbuild AS dshell
 
 ARG USERID=1000
 RUN adduser --quiet --home /code --uid ${USERID} --disabled-password --gecos "" influx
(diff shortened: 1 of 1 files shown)


rawkode:
Normally I'd scream "GOAWAY" when I see `RUN chmod` in a `Dockerfile`, but this is a rather special case for an environment that never leaves the developers machine.

So I've made an exception :smile:

Dockerfile
```diff
@@ -18,15 +18,21 @@ RUN apt install --yes \
...
         nodejs \
+        pkg-config \
         protobuf-compiler \
         ragel \
         rustc \
         yarn
 
+ENV GOPATH=/go
+ENV PATH /go/bin:$PATH
+RUN go get github.com/influxdata/pkg-config
```
jsternberg, line 29: I don't think this is needed. The `Makefile` should take care of using an ephemeral `pkg-config` wrapper. The `pkg-config` package that's installed above is definitely needed. What was the error message when these lines were excluded?

rawkode: Ah, this PR was created before I fixed the hard coded path inside this:

https://github.com/influxdata/influxdb/commit/917fcfc2d7b7a7feea67647ec6f9687a7065f1b7

I can probably remove these lines now, I believe you are correct.

PR #18281, opened 2020-05-29
fix(swagger): unique operation id

sranka:
Repaired operationId of /orgs/{orgID}/invites/{inviteID}/resend , swagger operation names must be unique. influxdb client API generators fail when they are not.

- [x] Rebased/mergeable
- [x] Tests pass


sranka:
@bednar FYI

sranka:
Both #18250 or #18252 provide alternative solutions, so there is no need to have a 3rd one.

PR #18283, opened 2020-05-29
test: add test for windowed count aggregate

### LINKED ISSUE #18183: Create unit test for WindowAggregateResultSet
This is where the WindowAggregateResultSet is used in code:
https://github.com/influxdata/influxdb/blob/feat/implement-aggregate-pushdown/storage/readservice/store.go#L188

We need a unit test for it.

Also, please double-check to make sure that this flux end-to-end test runs and passes in OSS:
https://github.com/influxdata/flux/blob/master/stdlib/planner/window_agg_push_test.flux

rockstar:
This patch adds tests for `ReadWindowAggregate` and the one supported
window aggregate, the count aggregate.

Fixes #18183

### DIFF
--- a/storage/reads/aggregate_resultset.go
+++ b/storage/reads/aggregate_resultset.go
@@ -60,5 +60,8 @@ func (r *windowAggregateResultSet) Stats() cursors.CursorStats {
 }
 
 func (r *windowAggregateResultSet) Tags() models.Tags {
+	if r.seriesRow == nil {
+		return models.Tags{}
+	}
 	return r.seriesRow.Tags
 }
--- a/storage/reads/aggregate_resultset_test.go
+++ b/storage/reads/aggregate_resultset_test.go
@@ -1,0 +1,169 @@
+package reads_test
+
+import (
+	"context"
+	"reflect"
+	"testing"
+
+	"github.com/influxdata/influxdb/v2/models"
+	"github.com/influxdata/influxdb/v2/storage/reads"
+	"github.com/influxdata/influxdb/v2/storage/reads/datatypes"
+	"github.com/influxdata/influxdb/v2/tsdb/cursors"
+)
+
+func TestNewWindowAggregateResultSet_Tags(t *testing.T) {
+
+	newCursor := sliceSeriesCursor{
+		rows: newSeriesRows(
+			"clicks click=1 1",
+		)}
+
+	request := datatypes.ReadWindowAggregateRequest{}
+	resultSet, err := reads.NewWindowAggregateResultSet(context.Background(), &request, &newCursor)
+
+	if err != nil {
+		t.Fatalf("error creating WindowAggregateResultSet: %s", err)
+	}
+
+	// If .Next() was never called, seriesRow is nil and tags are empty.
+	expectedTags := "[]"
+	if resultSet.Tags().String() != expectedTags {
+		t.Errorf("expected tags: %s got: %s", expectedTags, resultSet.Tags().String())
+	}
+
+	resultSet.Next()
+	expectedTags = "[{_m clicks}]"
+	if resultSet.Tags().String() != expectedTags {
+		t.Errorf("expected tags: %s got: %s", expectedTags, resultSet.Tags().String())
+	}
+}
+
+type mockIntegerArrayCursor struct {
+	callCount int
+}
+
+func (i *mockIntegerArrayCursor) Close()                     {}
+func (i *mockIntegerArrayCursor) Err() error                 { return nil }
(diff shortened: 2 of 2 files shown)


storage/reads/aggregate_resultset_test.go
```diff
@@ -0,0 +1,168 @@
...
+	newCursor := sliceSeriesCursor{
+		rows: newSeriesRows(
+			"clicks click=1 1",
+		)}
+
+	request := datatypes.ReadWindowAggregateRequest{}
+	resultSet, err := reads.NewWindowAggregateResultSet(context.Background(), &request, &newCursor)
+
+	if err != nil {
+		t.Errorf("Error creating WindowAggregateResultSet: %s", err)
```
jsternberg, line 25: Error messages generally have lowercase strings for the beginning so this would probably be better as `t.Errorf("error ...")`.

Also, if there is an error, then the rest of this test will likely not succeed since you use the `resultSet`. In that case, you should use `t.Fatalf`.

storage/reads/aggregate_resultset_test.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	// If .Next() was never called, seriesRow is nil and tags are empty.
+	expectedTags := "[]"
+	if resultSet.Tags().String() != expectedTags {
+		t.Errorf("Expected tags: %s got: %s", expectedTags, resultSet.Tags().String())
```
jsternberg, line 31: Similar thing here for the message. A good preference is to try to create an error message that looks like this:

```
unexpected tags -want/+got:
    - wanted tags
    + got tags
```

A good template for this is:

```
if want, got := expectedTags, resultSet.Tags().String(); want != got {
    t.Errorf("unexpected tags -want/+got:\n\t- %s\n\t+ %s", want, got)
}
```

If you notice, I also assign the values we are comparing to `want` and `got` and then compare them. This helps in two ways. First, it makes the comparison and then subsequent printing easier. Second, it uses a standard name so it is easier to read and know what you are expecting so you can more easily keep track of the got and wanted.

For whatever reason, we use the word `want` instead of `expected` or `exp`.

`Errorf` is probably fine here since the remainder of the function can still run if this fails. The general rule of thumb is if the error is fatal and the unit test can't continue to use fatal, but try to use error if possible since you will have more problems printed out so fewer runs of the test.

storage/reads/aggregate_resultset_test.go
```diff
@@ -0,0 +1,168 @@
...
+		t.Errorf("Unexpected: resultSet could not advance")
+	}
+	cursor := resultSet.Cursor()
+	if cursor == nil {
+		t.Errorf("Unexpected: cursor was nil")
+	}
+	integerArrayCursor := cursor.(cursors.IntegerArrayCursor)
+	integerArray := integerArrayCursor.Next()
+
+	if !reflect.DeepEqual(integerArray.Values, []int64{2, 5, 1}) {
```
yzhang1991, line 165: Can we also verify the timestamps?

rockstar: Yeah, I _just_ remembered I wanted to do that this morning (it was very late/early when I finally put this up).

storage/reads/aggregate_resultset_test.go
```diff
@@ -0,0 +1,168 @@
...
+			1000000020,
+		},
+		Values: []int64{100, 55, 256, 83, 99, 124, 1979, 4, 67, 49929},
+	}
+}
+
+type mockCursorIterator struct{}
+
+func (i *mockCursorIterator) Next(ctx context.Context, req *cursors.CursorRequest) (cursors.Cursor, error) {
+	// TODO: this is the last little piece to mock out the entirety of the I/O
```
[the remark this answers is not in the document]
rockstar: This should be removed. I leave notes for myself when I leave the computer and need to have context when I come back. This one survived the return. :)

storage/reads/aggregate_resultset.go
```diff
@@ -60,5 +60,8 @@ func (r *windowAggregateResultSet) Stats() cursors.CursorStats {
 }
 
 func (r *windowAggregateResultSet) Tags() models.Tags {
+	if r.seriesRow == nil {
```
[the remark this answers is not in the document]
rockstar: The test found this. :)

storage/reads/aggregate_resultset_test.go
yzhang1991: Yup, I think I want to have coverage for the window calculation.

PR #18288, opened 2020-05-29
feat(storage/flux): implement create empty for the window table reader

jsternberg:
This implements create empty for the window table reader and allows this
table read function to be used when it is specified. It will pass down
the create empty flag from the original window call into the storage
read function.

This also fixes the window table reader so it properly creates
individual tables for each window. Previously, it was constructing one
table for an entire series instead of one table per window.

Tests have been added to verify three edge case behaviors. The first is
the normal read operation where all values are present. The second is
when create empty is specified so null values may be created. The third
is with truncated boundaries to ensure that storage is read from and the
start and stop timestamps get correctly truncated.

### DIFF
--- a/storage/flux/table.gen.go
+++ b/storage/flux/table.gen.go
@@ -9,6 +9,7 @@ package storageflux
 import (
 	"sync"
 
+	"github.com/apache/arrow/go/arrow/array"
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/arrow"
 	"github.com/influxdata/flux/execute"
@@ -102,7 +103,9 @@ type floatWindowTable struct {
 	floatTable
 	windowEvery int64
 	arr         *cursors.FloatArray
+	nextTS      int64
 	idxInArr    int
+	createEmpty bool
 }
 
 func newFloatWindowTable(
@@ -110,6 +113,7 @@ func newFloatWindowTable(
 	cur cursors.FloatArrayCursor,
 	bounds execute.Bounds,
 	every int64,
+	createEmpty bool,
 	key flux.GroupKey,
 	cols []flux.ColMeta,
 	tags models.Tags,
@@ -123,6 +127,11 @@ func newFloatWindowTable(
 			cur:   cur,
 		},
 		windowEvery: every,
+		createEmpty: createEmpty,
+	}
+	if t.createEmpty {
+		start := int64(bounds.Start)
+		t.nextTS = start + (every - start%every)
 	}
 	t.readTags(tags)
 	t.advance()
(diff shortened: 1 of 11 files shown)


storage/flux/table.gen.go.tmpl
```diff
@@ -117,47 +121,128 @@ func new{{.Name}}WindowTable(
...
 }
 
-func (t *{{.name}}WindowTable) advance() bool {
+func (t *{{.name}}WindowTable) Do(f func(flux.ColReader) error) error {
+	return t.do(f, t.advance)
+}
+
+// createNextWindow will read the timestamps from the array
+// cursor and construct the values for the next window.
+func (t *{{.name}}WindowTable) createNextWindow() (start, stop *array.Int64, ok bool) {
```
rockstar, line 142: Normally I shy away from functions that are only ever called once, but this makes this process much more readable.

storage/flux/window.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/influxdata/flux/arrow"
+	"github.com/influxdata/flux/execute"
+	"github.com/influxdata/flux/values"
+	"github.com/influxdata/influxdb/v2"
+)
+
+// splitWindows will split a windowTable by creating a new table from each
+// row and modifying the group key to use the start and stop values from
+// that row.
+func splitWindows(ctx context.Context, in flux.Table, f func(t flux.Table) error) error {
```
rockstar, line 19: I had to trace this all the way back up the stack, and I realize that it's a common pattern, but I'd like to suggest a course correction going forward: this function injection pattern is used in a lot of places, and while the result may end up being less duplicated code, it comes at the cost of readable and maintainable code. This function just adds another layer to the stack. We have to open 6 different files between call point and function declaration.

I'm not asking for anything to be done in this patch. I am asking that we "walk, don't run" from this pattern in the future.

jsternberg: This pattern has a lot to be desired and I agree I don't want it. My first implementation attempt was going to be to use the planner and inject a new transformation. Something like: `readAggregateWindow |> _splitWindows` if it were represented in flux code. I shied away from this implementation mostly for fear of either deadlocking the system or oversaturating it with tables. Because this has the potential for creating so many tables, I feared that it would overwhelm the dispatcher or potentially cause a deadlock for the same reason filter caused a deadlock. The split windows would be occupying a dispatcher slot and would not yield it to another transformation until it had split the entire table. I didn't view this as tenable and I'd like to revisit the execution engine to account for this because we're supposed to be able to account for that.

So this wasn't my preferred method and I'm happy to explore other avenues where we don't need to continuously wrap the table processing function.

rockstar: Right. I trust your judgment that it's the best solution for what we have at the time. I also appreciate your insight into other options you'd considered and (rightfully) chose to avoid. Mostly, I'd just like to start applying some friction to this pattern, so that we can start coming up with something that better lends itself to context-switching this logic regularly without too much cost.

storage/flux/window.go
```diff
@@ -0,0 +1,183 @@
...
+			table := &windowTableRow{
+				buffer: buffer,
+				done:   done,
+			}
+			if err := f(table); err != nil {
+				return err
+			}
+
+			select {
+			case <-done:
```
yzhang1991, line 86: Can you explain why a `done` channel is needed (I've seen this in most of the table structures)? I thought the call of `f(table)` will be blocking and synchronous?

storage/flux/window.go
```diff
@@ -0,0 +1,183 @@
...
+	return wts.Do(f)
+}
+
+type windowTableSplitter struct {
+	ctx context.Context
+	in  flux.Table
+}
+
+func (w *windowTableSplitter) Do(f func(flux.Table) error) error {
+	defer w.in.Done()
```
yzhang1991, line 33: Why is this needed? You are calling `w.in.Do()` anyway

storage/flux/table.gen.go
```diff
@@ -123,47 +127,128 @@ func newFloatWindowTable(
...
+	return v, ok
+}
+
+// isInWindow will check if the given time at stop can be used within
+// the window stop time for ts. The ts may be a truncated stop time
+// because of a restricted boundary while stop will be the true
+// stop time returned by storage.
+func (t *floatWindowTable) isInWindow(ts int64, stop int64) bool {
+	start := stop - t.windowEvery
+	return start < ts && ts <= stop
```
yzhang1991, line 199: I think this should be `start <= ts && ts < stop`?
Given a timestamp `t` and window parameter `every`, the start of the window that `t` belongs to is:
```
start = t - t % every
```
(See: https://github.com/influxdata/flux/blob/master/execute/window.go#L110 and https://github.com/influxdata/influxdb/blob/master/storage/reads/array_cursor.gen.go#L254)

Then we can know:
```
start % every == 0, start - start % every == start
(start + every) - (start + every) % every == start + every
```

Therefore, the window range should include the `start` but exclude the `stop`.

storage/flux/window.go
jsternberg: It's possible for this to fail when reading the schema. Since the call to `Done()` doesn't harm anything and should always happen, this will ensure the table is always marked as processed since that's a requirement for the `Do()` function.

storage/flux/table.gen.go
jsternberg: This was really confusing for me but I think the reason is because we are looking at the stop value. It might be worth a comment. When a window has a stop value of 30 seconds, that means it is covering the window of `[20, 30)` which doesn't actually include 30. So it appears like this should be `start <= ts && ts < stop`, but we're matching stop values for the interval. For the window `[20, 30)`, the stop value can be anywhere from `(20, 30]`. It's impossible for the stop value to be `20` exactly because then it would be the stop value for `[10, 20)`.

storage/flux/window.go
jsternberg: This is to prevent oversaturation of tables. The `Process` call isn't blocking and synchronous. It gets redirected through the consecutive transport. This prevents the source from continuously producing tables and consuming additional memory and keeps the process synchronous.

I think this detail is confusing because it isn't evident as part of this section of code, but as part of the executor and dispatcher.

yzhang1991: Oh right, I forgot about the dispatcher.

PR #18318, opened 2020-06-01
test(e2e): dashboard notes

### LINKED ISSUE #7332: Empty prefix alias for wildcards aggregate functions
Moving and downsampling data between databases or retention policies should allow us to maintain the field names.

I tried

```
SELECT last(*) AS ""
INTO "mfcf_vmware"."autogen".:measurement
FROM /^vmware_.*/
WHERE time >= '2016-01-01T00:00:00Z'
GROUP BY time(30m), "host"
```

but sadly because the prefix key is empty it fallbacks to "last_" in this case.

https://github.com/influxdata/influxdb/pull/7009
@jsternberg would adding this be simple?
@gunnaraasen should something different be done so it uses backreferences?

121watts:
Closes #7332

### DIFF
--- a/ui/cypress/index.d.ts
+++ b/ui/cypress/index.d.ts
@@ -15,8 +15,9 @@ import {
   getByInputValue,
   getByTitle,
   createTask,
-  createVariable,
   createMapVariable,
+  createCSVVariable,
+  createQueryVariable,
   createAndAddLabel,
   createLabel,
   createBucket,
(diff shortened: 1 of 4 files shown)


ui/cypress/index.d.ts
```diff
@@ -39,8 +40,9 @@ declare global {
       signin: typeof signin
       setupUser: typeof setupUser
       createSource: typeof createSource
+      createCSVVariable: typeof createCSVVariable
```
drdelambre, line 43: good to know

PR #18319, opened 2020-06-01
feat(ui): expose bucketid

### LINKED ISSUE #18229: Make it easier to determine bucket ID
__Proposal:__
As a developer using InfluxDB OSS or Cloud
I want to easily figure out the ID from a bucket
So that I can more easily write data to the database

__Current behavior:__
Today, there's no easy way to determine the ID for a bucket. If we look at InfluxDB's Buckets page, we list bucket names, but not the bucket ID:

image

It's only if I click on the **Settings** button, and then inspect the URL, that I can find the bucket ID. For example, in the example below, the bucket ID is _05a0fd81bddcd000_. But I only know that since another InfluxDB employee told me. This isn't documented anywhere as far as I know.

image

__Desired behavior:__
In the list of buckets, display _Bucket ID: <bucket ID>_ in de-emphasized text. Perhaps the same style of text as the data retention length. But that's ultimately a decision to be made by our design team.

__Use case:__
We are getting users of our free product who can't figure out how to write data because they cannot determine their bucket ID.

alexpaxton:
Closes #18229 

- Display `bucketid` on user & demo data buckets
- Clicking bucket id copies it to clipbboard (see attached gif)
- Extracted bucket card meta and actions into separate components (makes `BucketCard` easier to read)

copy-bucket-id

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 1. [18279](https://github.com/influxdata/influxdb/pull/18279): Make all pkg applications stateful via stacks
 
+### UI Improvements
+
+1. [18319](https://github.com/influxdata/influxdb/pull/18319): Display bucket ID in bucket list and enable 1 click copying
+
 ## v2.0.0-beta.11 [2020-05-26]
 
 ### Features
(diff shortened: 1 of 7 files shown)


alexpaxton:
@ebb-tide I did not think of that at all, glad I got you as a reviewer. Just pushed changes that add the "Copy ID" functionality to demo data bucket cards

PR #18320, opened 2020-06-01
feat(flag-redirectTo): using old UI feature flag to flag redirectTo

Asalem1:
This is a temporary solution to a problem that is being addressed elsewhere. This PR aims to add a UI feature flag that can be toggled from the console, rather than using the existing configCat setup since that currently doesn't support unauthenticated requests.


drdelambre:
in case anyone stumbles on this PR; you don't need to add a default value of false for a feature flag, just deploy the dark conditional path. the local override system available (`influx` in the console) will turn it on and off.

Issue #18271, opened 2020-05-28
Export by stack


jsteenb2:
As a user,
I want to to export my existing stack in the form of a template,
so that I can update/patch my package in a controlled manner

AC

- [ ] exporting a pkg with a stack id exports the pkg using the existing 
- [ ] deleted resources, are no exported (there is nothing to export)
- [ ] extend CLI with `influx pkg export stack STACK_ID`

PR #18322, opened 2020-06-01
feat(pkger): add ability to export by stack

jsteenb2:
this ability exports all resources associated with a stack by the same
metadata.name fields as the original application had done it. This can
be used as a means to snapshot the current state of the stack. This can
be used for source control or other means.

closes: #18271

---
**Acceptance Criteria**
- [x] exporting a pkg with a stack id exports the pkg using the existing
- [x] deleted resources, are no exported (there is nothing to export)
- [x] extend CLI with `influx pkg export stack $STACK_ID`

---
- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/pkger/service.go
+++ b/pkger/service.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/url"
 	"path"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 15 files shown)


pkger/service.go
```diff
@@ -333,6 +335,184 @@ func (s *Service) DeleteStack(ctx context.Context, identifiers struct{ OrgID, Us
...
+		)
+
+		switch res.Kind {
+		case KindBucket:
+			bkt, err := s.bucketSVC.FindBucketByID(ctx, res.ID)
+			if influxdb.ErrorCode(err) == influxdb.ENotFound {
+				continue
+			}
+			if err != nil {
+				return nil, ierrors.Wrap(err, fmt.Sprintf("failed to find bucket[%s]", res.ID.String()))
```
hoorayimhelping, line 363: so this failure message will say:

>failed to find bucket[12345876]

or is [%s] some special notation?

jsteenb2: you nailed it. The error message isn't great. I want to go through this service and update the error messages as part of our templates work. For some reason that use case feels like a good time to fixup error messaging... they platform errors we return could dramatically improve across teh platform, we can start with pkger here to try and touch that up 👍

PR #18330, opened 2020-06-02
feat: added a testing default value to feature flags

adrian-thurston:
This change enables the setting of a testing default value for each feature
flag. The TestingFlagger() can then be used in test harnesses instead of the
DefaultFlagger() and the testing values are supplied to the Enabled() functions.

If a testing default value is not specified for a flag, then the default value
is used.

With this change we can default feature flags to true, or supply some other
default value for non-bool feature flags without needing to explicitly set them
in individual tests. Intention is to use this in unit tests, end-to-end tests
in influxdb, and in pipeline tests.

Closes https://github.com/influxdata/idpe/issues/7430


adrian-thurston:
Note I moved the Lifetime struct into it's own package because it is needed by both the code generator and the generated code. If it is in the main feature package then the generated code is included by the generator and the generator itself will not compile if any changes are made.

kit/feature/feature.go
```diff
@@ -128,6 +91,27 @@ func (*defaultFlagger) Flags(_ context.Context, flags ...Flag) (map[string]inter
...
+// TestingFlagger returns a flagger that always returns testing values.
+func TestingFlagger() Flagger {
+	return &testingFlagger{}
+}
+
+// Flags returns a map of testing values. It never returns an error.
+func (*testingFlagger) Flags(_ context.Context, flags ...Flag) (map[string]interface{}, error) {
+	if len(flags) == 0 {
+		flags = Flags()
+	}
```
gavincabbage, line 105: This isn't going to work as expected in IDPE (I've actually got an open issue because the same bug exists in the `defaultFlagger` impl above. Both of these flaggers need to accept a `[]feature.Flag` on construction, and in OSS be passed `influxdb/kit/feature.Flags()` and in IDPE be passed `idpe/kit/feature.Flags()`. This is because OSS `Flags()` in OSS returns only OSS flags, but the corresponding IDPE `Flags()` function returns the union of OSS and IDPE flags.

Basically, it just needs to be something like this 

```
type testingFlagger struct{
    flags []Flag
}

// TestingFlagger returns a flagger that always returns testing values.
func TestingFlagger(flags []Flag) Flagger {
	return &testingFlagger{
            flags: flags,
        }
}

// Flags returns a map of testing values. It never returns an error.
func (t *testingFlagger) Flags(_ context.Context, flags ...Flag) (map[string]interface{}, error) {
	if len(flags) == 0 {
		flags = t.flags
	}
 ```

If you can make the same change to the `defaultFlagger` I would appreciate it, but its less of a big deal there.

gavincabbage:
@adrian-thurston If I understand this correctly, this would mean the test harness would always use the `testingDefault`. Doesn't this mean the _actual_ defaults are no longer tested?

e.g. If a new feature is defaulted to false, but test defaulted to true, and we run the e2e tests in OSS, it will use the new testing flagger and return true to test this case, but it will no longer test the default code path.

This is a big issue since we're relying on that default code path as the "control group" that we can always revert to by disabling a flag if the new feature introduces a bug. We need to maintain testing and confidence in that.

Am I misunderstanding the PR or is this a valid concern? We need both the real default and the testing default to be tested if we're going to use this approach.

adrian-thurston:
@gavincabbage That's right, in testing we would always be running with feature flags enabled (when explicitly specified in the yaml file -- they default to the default for each feature flag).

I hadn't considered that enabling a feature flag could reduce test coverage. On the query team at least we are trying to write tests that explicitly exercise the old behavior as well as the new behavior. So for us, turning on the feature flag shouldn't reduce test coverage. But maybe other teams are not doing it this way? I understand we need something that works for everyone.

@nathanielc would be great to have your input on this.

adrian-thurston:
I guess what we are not testing when we do this is that the feature flag successfully disables the new code path.

gavincabbage:
> I guess what we are not testing when we do this is that the feature flag successfully disables the new code path.

That, and we aren't continuing to test our our "control" state. 

For example, our team recently replaced some implementations of various service interfaces. We used a flag to determine if request handlers should use the old or new impls for each request, based on the identify of the user/org sending the request.

We tested the new experimental service itself on its own, but the end to end tests continued to exercise the "control" code path. This gave us confidence that this code path maintained correctness as we developed the new code path, and we could always revert back to it.

If we had used the approach in this PR as I understand it, and set the testing default to enable the new code path, it would be tested by end to end tests, but the old code path would no longer be tested. This would leave an opening for us (or another team) to introduce a bug in the old code path without it being caught by a test, making it unsuitable as a control state since we can't turn off the flag and return to a "good" state if the new code path turns out to have a bug.

> On the query team at least we are trying to write tests that explicitly exercise the old behavior as well as the new behavior.

If every team did this, it would alleviate my concern above, but I'm not sure how that would work in an end-to-end test. If we had a flag to enable a new service implementation as described above, and the testing default was true, the end to end tests would no longer cover the old path.

Could you include some example tests using that pattern to help me understand? Sorry if I'm being dense 😅

gavincabbage:
Unrelated to the discussion above, could I also ask you to open an IDPE PR vendoring this before (or very quickly after) we merge it? Since there is so much change to the code generation, I want to verify the IDPE `list.go` is still generated correctly.

adrian-thurston:
@gavincabbage 

https://github.com/influxdata/flux/tree/master/stdlib/planner

We use two cases. In one we explicitly disable planner rules using flux options. That covers the flag is false case for us.

gavincabbage:
That works for Flux, but what about other cases like the example I presented?

This won't be dangerous tomorrow, but I'm trying to think of a month from now, when two engineers that aren't you and I are working separately in our shared codebase: One is developing a new service impl and enables this testing default. The other is making a change in the old service impl and expecting their path to be exercised by e2e tests. 

Then the e2e tests run in CI, they're green, dev #2 thinks his change works, but it actually introduces a bug. They deploy their code and it breaks prod, because dev #1's flag isn't enabled in prod, just tests. We've opened up a hole in our test coverage and shipped a bug as result.

cc @aanthony1243 in case he has thoughts - he was actually the one to drill this fear into me when we first started down the feature flag path.

adrian-thurston:
FYI @aanthony1243 going to nuke this PR based on Gavin's concerns and take a more focused approach at flipping FF on for our tests.

PR #18331, opened 2020-06-02
fix(dbrpv2): lookup by org name

gavincabbage:
This change adds support for the `org` name parameter in addition to `orgID` for organization lookup when interacting with the DBRP HTTP APIs for reading, updating and deleting DBRP mappings. It also enhances the DBRP mapping creation operation to consider `organization` in the request body, if `organization_id` is missing 

The logic is enhanced to fallback to look up the org ID from the org name if the ID is not present. If both are provided, the name will be ignored in favor of the ID.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 1. [18279](https://github.com/influxdata/influxdb/pull/18279): Make all pkg applications stateful via stacks
 
+### Bug Fixes
+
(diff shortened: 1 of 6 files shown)


hoorayimhelping:
what does DBRP stand for?

gavincabbage:
@hoorayimhelping "DBRP Mappings" map 1.x database/retention policy pairs to 2.x buckets. This concept was added to support InfluxQL on Cloud 2.

gavincabbage:
alchemy please merge

PR #18332, opened 2020-06-02
fix: update app-metrics path to point to the right place

hoorayimhelping:
`/app-metrics` => `/api/v2/app-metrics`

- [x] Rebased/mergeable
- [x] Tests pass


hoorayimhelping:
>suspicious

PR #18340, opened 2020-06-02
feat(secret): transition the secret service to isolated pattern

lyondhill:
We needed a secret service handler that would be pluggable in the org service. Building in the new vertical stacked pattern made the most sense to solve our problem and move ahead with the strategy for our api.

### DIFF
--- a/secret/http_server.go
+++ b/secret/http_server.go
@@ -1,0 +1,135 @@
+package secret
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/go-chi/chi"
+	"github.com/influxdata/influxdb/v2"
+	kithttp "github.com/influxdata/influxdb/v2/kit/transport/http"
+	"go.uber.org/zap"
+)
+
+type handler struct {
+	log *zap.Logger
+	svc influxdb.SecretService
+	api *kithttp.API
+
+	idLookupKey string
+}
+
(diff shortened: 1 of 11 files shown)


secret/storage.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+
+		if id != orgID {
+			// We've reached the end of the keyspace for the provided orgID
+			break
+		}
+
+		keys = append(keys, key)
+	}
+
```
GeorgeMac, line 107: You'll need to err check and close your cursor here.

Everyone does this, makes me think about the UX of cursors.

On that actually there is a `kv.Walk` that consumes a cursor and lets you provide a visit func. That does the cursor cleanup for you.

secret/http_server.go
```diff
@@ -0,0 +1,135 @@
...
+
+type handler struct {
+	log *zap.Logger
+	svc influxdb.SecretService
+	api *kithttp.API
+
+	idLookupKey string
+}
+
+// NewHandler generates a mountable handler for URMs. It needs to know how it will be looking up your resource id
```
GeorgeMac, line 21: I think this comment is incorrect. I assume it is taken from URMs.

secret/storage.go
lyondhill: oo thanks! I will fix this.

PR #18341, opened 2020-06-02
feat(notebooks): display raw data within query pipes

alexpaxton:
Closes https://github.com/influxdata/idpe/issues/7138

- Create new type `BothResults` that contains both raw data and parsed results
  - Using this temporarily until `RawDataTable` is included in Giraffe and accepts parsed results
- Create `Results` components as part of the `Query` pipe type
- Allow results to be rendered in 3 sizes
  - Small `200px`
  - Medium `400px`
  - Large `600px`

notebook-raw-data-sizes

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/notebooks/pipes/Query/Results.tsx
+++ b/ui/src/notebooks/pipes/Query/Results.tsx
@@ -1,0 +1,63 @@
+// Libraries
+import React, {FC} from 'react'
+import {BothResults} from 'src/notebooks/context/query'
+import {AutoSizer} from 'react-virtualized'
+import classnames from 'classnames'
+
+// Components
+import RawFluxDataTable from 'src/timeMachine/components/RawFluxDataTable'
+import ResultsHeader from 'src/notebooks/pipes/Query/ResultsHeader'
+
+// Types
+import {RawDataSize} from 'src/notebooks/pipes/Query'
+
+interface Props {
+  results: BothResults
+  size: RawDataSize
+  onUpdateSize: (size: RawDataSize) => void
+}
+
+const Results: FC<Props> = ({results, size, onUpdateSize}) => {
+  const resultsExist = !!results.raw
+  const className = classnames('notebook-raw-data', {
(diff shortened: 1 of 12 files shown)


ui/src/notebooks/pipes/Query/Results.tsx
```diff
@@ -0,0 +1,63 @@
...
+    [`notebook-raw-data__${size}`]: resultsExist && size,
+  })
+
+  let resultsBody = (
+    <div className="notebook-raw-data--empty">
+      Run the Notebook to see results
+    </div>
+  )
+
+  if (resultsExist) {
```
drdelambre, line 32: Should we include the loading state here too so that we can have a CTA and empty state?

ui/src/notebooks/pipes/Query/view.tsx
```diff
@@ -1,10 +1,22 @@
+// Libraries
 import React, {FC, useMemo} from 'react'
+import {get} from 'lodash'
```
drdelambre, line 3: It was so clean until this moment

ui/src/notebooks/pipes/Query/view.tsx
```diff
@@ -16,6 +28,11 @@ const Query: FC<PipeProp> = ({data, onUpdate, Context}) => {
     onUpdate({queries: _queries})
   }
 
+  const onUpdateSize = (rawDataSize: RawDataSize): void => {
+    console.log('onUpdateSize Query')
```
drdelambre, line 32: Forgot one

alexpaxton: Derp

ui/src/notebooks/pipes/Query/Results.tsx
alexpaxton: I like that idea. Ideally we could get `RemoteDataState` passed in, then we could also show errors

PR #18342, opened 2020-06-02
feat: perform final aggregation of grouped aggregates on flux side

jlapacik:
When pushing down a group followed by an aggregate, storage will aggregate per series (not per group). Therefore a final aggregation is needed on the query side. This PR modifies the planner's rewrite rules for grouped aggregates accordingly.

Note, this PR adds rewrite rules for `group() |> first()` and `group() |> last()`. Because storage does not yet support these operations, I've updated it to expose its current capabilities.

### DIFF
--- a/query/stdlib/influxdata/influxdb/rules.go
+++ b/query/stdlib/influxdata/influxdb/rules.go
@@ -873,7 +873,9 @@ func (PushDownGroupAggregateRule) Rewrite(ctx context.Context, pn plan.Node) (pl
 	case universe.FirstKind:
 		// ReadGroup() -> first => ReadGroup(first) -> min
 		root := plan.CreatePhysicalNode("min", &universe.MinProcedureSpec{
-			SelectorConfig: execute.DefaultSelectorConfig,
+			SelectorConfig: execute.SelectorConfig{
+				Column: execute.DefaultTimeColLabel,
+			},
 		})
 
 		leaf := plan.CreatePhysicalNode("ReadGroupAggregate", &ReadGroupPhysSpec{
@@ -889,7 +891,9 @@ func (PushDownGroupAggregateRule) Rewrite(ctx context.Context, pn plan.Node) (pl
 	case universe.LastKind:
 		// ReadGroup() -> last => ReadGroup(last) -> max
 		root := plan.CreatePhysicalNode("max", &universe.MaxProcedureSpec{
-			SelectorConfig: execute.DefaultSelectorConfig,
+			SelectorConfig: execute.SelectorConfig{
+				Column: execute.DefaultTimeColLabel,
+			},
 		})
 
 		leaf := plan.CreatePhysicalNode("ReadGroupAggregate", &ReadGroupPhysSpec{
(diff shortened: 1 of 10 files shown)


query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -822,37 +819,126 @@ func (rule PushDownGroupAggregateRule) Pattern() plan.Pattern {
...
+			AggregateMethod:   universe.SumKind,
+		})
+
+		root.AddPredecessors(leaf)
+		leaf.AddSuccessors(root)
+		return root, true, nil
+	case universe.FirstKind:
+		// ReadGroup() -> first => ReadGroup(first) -> min
+		root := plan.CreatePhysicalNode("min", &universe.MinProcedureSpec{
+			SelectorConfig: execute.DefaultSelectorConfig,
```
wolffcm, line 876: This should be `min(_time)` so I don't think you can use the default selector config here. In Flux `first` just means the first row of a table, but users expect that to be the oldest point, and `ReadGroup` won't necessarily return data in time order.

query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -822,37 +819,126 @@ func (rule PushDownGroupAggregateRule) Pattern() plan.Pattern {
...
+			AggregateMethod:   universe.FirstKind,
+		})
+
+		root.AddPredecessors(leaf)
+		leaf.AddSuccessors(root)
+		return root, true, nil
+	case universe.LastKind:
+		// ReadGroup() -> last => ReadGroup(last) -> max
+		root := plan.CreatePhysicalNode("max", &universe.MaxProcedureSpec{
+			SelectorConfig: execute.DefaultSelectorConfig,
```
wolffcm, line 892: Same here, this should be `max(_time)`

query/storage.go
```diff
@@ -24,6 +24,17 @@ type StorageReader interface {
...
 }
 
+type GroupCapability interface {
+	HaveCount() bool
+	HaveSum() bool
+	HaveFirst() bool
+	HaveLast() bool
+}
+
+type GroupAggregator interface {
```
wolffcm, line 34: It seems like the corresponding interface for windowed aggregation is `WindowAggregateReader`. Should this be called something like `AggregatingGroupReader` or something like that?

query/stdlib/influxdata/influxdb/rules.go
jlapacik: Thanks for catching this. I must have been on autopilot.

query/storage.go
jlapacik: The reason I broke from convention here is that this interface doesn't have a "Read*" method. `ReadGroup` is part of the `StorageReader` interface, whereas `ReadAggregateWindow` is part of the `WindowAggregateReader` interface. I thought that might lead to some confusion, but I'm personally fine with either.

query/stdlib/influxdata/influxdb/rules.go
wolffcm: Good point. I'm fine with you leaving it as-is.

wolffcm: Oops, I meant to reply to a different comment.

Issue #18336, opened 2020-06-02
Add user-agent to influx CLI HTTP Calls


jsteenb2:
Acceptance Criteria

- [x] all http calls from influx CLI have a `User-Agent` header
- [x] `User-Agent` header should look like `User-Agent: influx/$VERSION (<system-information>) Sha/$SHA Date/$BUILD_DATE`

note: should the `influx` product be identified as `influx2` to differentiate it from v1?

Issue #14821, opened 2019-08-27
cannot get version of influx CLI


gitirabassi:
Probably is hidden in the code but neither `influx --version`, `influx version`, `influx -v` and when executing `influx --help` there is no mention on how to output a version

kelwang:
@gitirabassi you are talking about the influxdb version, not the cli version, right?

gitirabassi:
I was asking for the CLI, but both would be awesome. When I was running it, i wasn't sure if running the v1 or v2 version of the client.

My primary concern is with people migrating from v1

Also, would be cool to have, other then the version of the server, some metadata of it: if cloud, enterprise or OSS. 

These down here are docker and kubernetes examples. I would be fine with a 2 liner to start with
```sh
bash-3.2$ kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-30T12:22:52Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.10", GitCommit:"37d169313237cb4ceb2cc4bef300f2ae3053c1a2", GitTreeState:"clean", BuildDate:"2019-08-19T10:44:49Z", GoVersion:"go1.11.13", Compiler:"gc", Platform:"linux/amd64"}
bash-3.2$ docker version
Client: Docker Engine - Community
 Version:           19.03.2
 API version:       1.40
 Go version:        go1.12.8
 Git commit:        6a30dfc
 Built:             Thu Aug 29 05:26:49 2019
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.2
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.8
  Git commit:       6a30dfc
  Built:            Thu Aug 29 05:32:21 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.6
  GitCommit:        894b81a4b802e4eb2a91d1ce216b8817763c29fb
 runc:
  Version:          1.0.0-rc8
  GitCommit:        425e105d5a03fabd737a126ad93d62a9eeede87f
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

[...]

PR #18344, opened 2020-06-02
feat(influx): extend CLI with versoin and user-agent

jsteenb2:
Closes: #18336 
Closes: #14821

---
Acceptance Criteria

- [x] all http calls from influx CLI have a `User-Agent` header
- [x] `User-Agent` header should look like `User-Agent: influx/$VERSION (<system-information>) Sha/$SHA Date/$BUILD_DATE`

---
- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influx/main.go
+++ b/cmd/influx/main.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/influxdata/influxdb/v2"
 	"github.com/influxdata/influxdb/v2/bolt"
@@ -26,6 +27,12 @@ import (
 )
 
 const maxTCPConnections = 10
+
+var (
+	version = "dev"
+	commit  = "none"
+	date    = time.Now().UTC().Format(time.RFC3339)
+)
 
 func main() {
 	influxCmd := influxCmd()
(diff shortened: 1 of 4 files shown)


cmd/influx/main.go
```diff
@@ -27,6 +29,12 @@ import (
 
 const maxTCPConnections = 10
 
+var (
+	version = "dev"
+	commit  = "none"
+	date    = time.Now().UTC().Format(time.RFC3339)
```
mark-rushakoff, line 35: Fly-by comment: this isn't the build date (as noted in the PR description); it's the time that the `influx` process was started.

jsteenb2: you're right. I borrowed this from the influxd binary here:

https://github.com/influxdata/influxdb/blob/master/cmd/influxd/main.go#L27

just noticed I dont' have the `fmt.Sprintf`, but that shouldn't matter yah?

It looks like the date isn't provided as an LD flag either, so that definitely can change 🤔. What do you think is the best way forward here?

mark-rushakoff: Agreed the `fmt.Sprint` isn't necessary; `time.Format` already returns a string.

I think setting the date at link time is the correct approach.

mark-rushakoff: Looks like we do set the date as a linker flag in goreleaser after all:
https://github.com/influxdata/influxdb/blob/d21fce5f6d026904a6048704c6761384ce2736d9/.goreleaser.yml#L19
https://github.com/influxdata/influxdb/blob/d21fce5f6d026904a6048704c6761384ce2736d9/.goreleaser-nightly.yml#L19

Any tooling we have to inspect the resulting date would know that `version == "dev"` indicates the date would be meaningless.

Nothing to change here after all.

PR #18346, opened 2020-06-02
fix(variable-hydration): fixed variable overhydration issue 

### LINKED ISSUE #18192: Variable Hydration Filtering / Variable loading isn't fully functioning
This issue is being written as a fast-follow for a PR to resolve some variable hydration issues we were experiencing due to a recent refactor. The refactor was meant to limit the number of queries that are made when a variable selection is updated. This works too well and has exposed bugs in our current system that is causing variables to act unpredictably. In order to resolve some of the underlying issues, I've created a PR to first revert the functionality with the intention of working through the issues that arose.

Asalem1:
Closes #18192 

### Problem
This PR aims to address the issue of overly hydrating variables when a variable selection is made.

#### Solution
The solution is as follows:

- Only pass in the variable that has been selected to hydrate that specific variable
- Create a subgraph of the variables that are being used by performing an in order tree traversal of the parent's child nodes in order to first load children, then their parents

This solution will help reduce the amount of unnecessary query requests made from the UI post order to hydrate variables. However, since it involves a lot of moving parts that currently work, I've set it behind a feature flag in order to make sure that variables are still going to work correctly

- [x] Rebased/mergeable

### DIFF
--- a/ui/src/variables/actions/thunks.ts
+++ b/ui/src/variables/actions/thunks.ts
@@ -33,6 +33,7 @@ import {findDependentVariables} from 'src/variables/utils/exportVariables'
 import {getOrg} from 'src/organizations/selectors'
 import {getLabels, getStatus} from 'src/resources/selectors'
 import {currentContext} from 'src/shared/selectors/currentContext'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 // Constants
 import * as copy from 'src/shared/copy/notifications'
@@ -433,6 +434,10 @@ export const selectValue = (variableID: string, selected: string) => async (
 
   await dispatch(selectValueInState(contextID, variableID, selected))
   // only hydrate the changedVariable
-  dispatch(hydrateChangedVariable(variableID))
+  if (isFlagEnabled('hydratevars')) {
+    dispatch(hydrateChangedVariable(variableID))
+  } else {
+    dispatch(hydrateVariables(true))
+  }
   dispatch(updateQueryVars({[variable.name]: selected}))
 }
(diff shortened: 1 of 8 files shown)


ui/src/variables/actions/thunks.ts
```diff
@@ -433,7 +434,10 @@ export const selectValue = (variableID: string, selected: string) => async (
 
   await dispatch(selectValueInState(contextID, variableID, selected))
   // only hydrate the changedVariable
-  dispatch(hydrateVariables(true))
-  // dispatch(hydrateChangedVariable(variableID))
+  if (isFlagEnabled('hydratevars')) {
```
drdelambre, line 437: this is the most badass thing i've ever seen

Asalem1: pushing the limits of feature flagging:

https://media.giphy.com/media/WsvLlmmjx9tnmeTPNc/giphy.gif

ui/src/variables/utils/hydrateVars.test.ts
```diff
@@ -349,13 +440,241 @@ xdescribe('findSubgraph', () => {
...
+      timeRangeStartVariable,
+    ])
     expect(actual.length).toEqual(1)
     const [subgraph] = actual
     // expect the subgraph to return the passed in variable
     expect(subgraph.variable).toEqual(timeRangeStartVariable)
     // expect the parent to be returned with the returning variable
     expect(subgraph.parents[0].variable).toEqual(associatedVariable)
   })
+  test('should filter out inputs that have already been loaded based on a previous associated variable', async () => {
```
drdelambre, line 472: omg this documentation make my week

russorat:
We tested this locally a few hours ago and this is what we discovered:
 - Setting feature flags in OSS doesn't work
 - The number of calls made for a single cell went from 6 to 3 when a variable value was changed: 1 call to the cell query, then a call to the variable query, then another call to the cell query. IMO, we should be able to get this down to a single call.
 - This does not improve additional calls on the explorer page

PR #18348, opened 2020-06-03
feat: looking at some graphs

drdelambre:
make a graph show up, finish hooking up the loading / error states, general bug fixing and lifecycle definitions (like what happens if you add a visualization to a query cell that has already been called)?

Kapture 2020-06-02 at 20 35 23

### DIFF
--- a/ui/src/notebooks/components/NotebookPipe.tsx
+++ b/ui/src/notebooks/components/NotebookPipe.tsx
@@ -15,6 +15,7 @@ const NotebookPipe: FC<NotebookPipeProps> = ({
   data,
   onUpdate,
   results,
+  loading,
 }) => {
   const panel: FC<PipeContextProps> = useMemo(
     () => props => {
@@ -33,7 +34,13 @@ const NotebookPipe: FC<NotebookPipeProps> = ({
   }
 
   return (
-    <Pipe data={data} onUpdate={_onUpdate} results={results} Context={panel} />
+    <Pipe
+      data={data}
+      onUpdate={_onUpdate}
+      loading={loading}
+      results={results}
+      Context={panel}
+    />
   )
 }
 
--- a/ui/src/notebooks/components/panel/NotebookPanel.tsx
+++ b/ui/src/notebooks/components/panel/NotebookPanel.tsx
@@ -92,7 +92,7 @@ const NotebookPanel: FC<Props> = ({index, children, controls}) => {
 
   useEffect(() => {
     updateMeta(index, {panelRef} as PipeMeta)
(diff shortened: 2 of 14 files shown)


ui/src/notebooks/components/NotebookPipe.tsx
```diff
@@ -15,6 +15,7 @@ const NotebookPipe: FC<NotebookPipeProps> = ({
   data,
   onUpdate,
   results,
+  loading,
```
alexpaxton, line 18: OohooOoooo

ui/src/notebooks/components/panel/NotebookPanel.tsx
```diff
@@ -92,7 +92,7 @@ const NotebookPanel: FC<Props> = ({index, children, controls}) => {
 
   useEffect(() => {
     updateMeta(index, {panelRef} as PipeMeta)
-  })
+  }, [])
```
alexpaxton, line 95: I believe having this run only once on mount breaks the jump to feature. I fiddled with not having to hit `updateMeta` so much but couldn't get the feature to work otherwise

ui/src/notebooks/pipes/Visualization/view.tsx
```diff
@@ -0,0 +1,38 @@
...
+import {PipeProp} from 'src/notebooks'
+import EmptyQueryView, {ErrorFormat} from 'src/shared/components/EmptyQueryView'
+import ViewSwitcher from 'src/shared/components/ViewSwitcher'
+import {checkResultsLength} from 'src/shared/utils/vis'
+
+// NOTE we dont want any pipe component to be directly dependent
+// to any notebook concepts as this'll limit future reusability
+// but timezone seems like an app setting, and its existance within
+// the notebook folder is purely a convenience
+import {AppSettingContext} from 'src/notebooks/context/app'
```
alexpaxton, line 11: Agreed -- I don't think this is a blocker

ui/src/notebooks/components/panel/NotebookPanel.tsx
drdelambre: aight. i'll jump back on that then. i was debugging when something updates the meta, and saw an explosion! maybe the idea is to limit rendering in the first place.... i'll figure it out

drdelambre: offlined a test condition: `when a pipe is added and moved in order, the jump to feature should scroll to the notebook entry exposed in the minimap`
tested that ^^ and it seems to work with the change

PR #18357, opened 2020-06-03
feat: add communityTemplate feature flag and put copypasted container in it

### LINKED ISSUE #18238: Feature Flag Templates
## Background
We're an agile organization that iterates on features. We're going to develop this project in a flag-first way. We'll lead with a feature flag and follow up with small, easy-to-digest commits.

## Acceptance Criteria
* New templates are shown behind a feature flag
* Old templates are hidden behind the same feature flag

hoorayimhelping:
Closes #18238

Adds a feature flag and a simple copy-pasta component that logs that it's a flagged component and puts that nearly identical component behind the flag.

This is the foundational work for the whole of the Community Templates MVP work. If it looks like a copy/paste job, that's because that's the intent of it! We want this to carry as little risk as possible while we build confidence in the feature flag we've set up. If I managed to set up the flags incorrectly, the worst that will happen is a user will have a statement logged to their console. Once this gets out to prod and we confirm the flagging works the way we expect, we'll start in on the radical changes.

### Testing Instructions
* Run the binary with the following command:
    ```sh
    go run ./cmd/influxd/ --assets-path=ui/build --feature-flags=communityTemplates=true
    ```
* Start the frontend normally
* Go to Settings -> Templates
* you should see `CommunityTemplatesIndex - the flag is working` in the console.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/kit/feature/list.go
+++ b/kit/feature/list.go
@@ -30,6 +30,20 @@ func BackendExample() BoolFlag {
 	return backendExample
 }
 
+var communityTemplates = MakeBoolFlag(
+	"Community Templates",
+	"communityTemplates",
+	"Bucky, Johnny Steenbergen (Berg)",
+	false,
+	Permanent,
+	true,
+)
+
+// CommunityTemplates - Replace current template uploading functionality with community driven templates
+func CommunityTemplates() BoolFlag {
+	return communityTemplates
+}
+
 var frontendExample = MakeIntFlag(
 	"Frontend Example",
 	"frontendExample",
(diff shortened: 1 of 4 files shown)


kit/feature/list.go
```diff
@@ -30,6 +30,20 @@ func BackendExample() BoolFlag {
 	return backendExample
 }
 
+var communityTemplates = MakeBoolFlag(
+	"Community Templates",
+	"communityTemplates",
+	"Bucky, Johnny Steenbergen (Berg)",
```
jsteenb2, line 36: 😳

PR #18358, opened 2020-06-03
feat: support all of the view types

drdelambre:
Kapture 2020-06-03 at 16 34 50

### DIFF
--- a/ui/src/notebooks/pipes/Visualization/view.tsx
+++ b/ui/src/notebooks/pipes/Visualization/view.tsx
@@ -4,6 +4,7 @@ import EmptyQueryView, {ErrorFormat} from 'src/shared/components/EmptyQueryView'
 import ViewSwitcher from 'src/shared/components/ViewSwitcher'
 import {ViewTypeDropdown} from 'src/timeMachine/components/view_options/ViewTypeDropdown'
 import {checkResultsLength} from 'src/shared/utils/vis'
+import {ViewType} from 'src/types'
 import {createView} from 'src/views/helpers'
 
 // NOTE we dont want any pipe component to be directly dependent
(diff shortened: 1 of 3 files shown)


ui/src/notebooks/pipes/Visualization/view.tsx
```diff
@@ -2,19 +2,93 @@ import React, {FC, useContext} from 'react'
...
+const Visualization: FC<PipeProp> = ({
+  data,
+  results,
+  onUpdate,
+  Context,
+  loading,
+}) => {
   const {timeZone} = useContext(AppSettingContext)
 
+  const updateType = (type: ViewType) => {
```
drdelambre, line 25: I will 100% concede that this entire function is garbage. converting between types like this is a recipe for bugs by creating a "m:n" relationship that you have to track all the permutations of. but any changes would have to be reflected in TimeMachine as well, increasing the surface area of this feature by... at least a spike cycle

alexpaxton: All good! It's a prototype after all.

Wondering if it makes sense to extract this list of `if` statements into a separate file for readability?

ui/src/timeMachine/components/view_options/ViewTypeDropdown.tsx
```diff
@@ -75,22 +75,22 @@ class ViewTypeDropdown extends PureComponent<Props> {
   }
 
   private get dropdownStatus(): ComponentStatus {
-    const {view} = this.props
+    const {viewType} = this.props
 
-    if (view.properties.type === 'check') {
+    if (viewType === 'check') {
```
alexpaxton, line 80: nice and flat

PR #18369, opened 2020-06-04
feat: remove auto refresh from cloud instances

drdelambre:
Closes influxdata/idpe#7409

this... removes the autorefresh interaction, and... stops sending data for client instances that already have auto refresh set up on their local state


drdelambre:
IT LOOKS WEIRD, but mostly because i'm used to it being there

insekt:
Hi,
What was the reason for this change?
It seems odd not having it. At least you could have toggle "Auto-refresh on/off"?

PR #18371, opened 2020-06-04
feat: start using the new org handler from the tenant service.

lyondhill:
The rest of the tenant system is in place except the org http api handler and the
user api handler.

### DIFF
--- a/cmd/influxd/launcher/launcher.go
+++ b/cmd/influxd/launcher/launcher.go
@@ -991,18 +991,17 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 	}
 
 	// feature flagging for new labels service
-	var labelsHTTPServer *kithttp.FeatureHandler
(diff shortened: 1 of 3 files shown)


bednar:
The new org handler from the tenant service doesn't satisfy swagger. The service doesn't implements:

1. `links` property in an `Organization` schema: https://github.com/influxdata/influxdb/blob/0caee29f249a4404f774cad72646e8461e62a098/http/swagger.yml#L7732
1. `labels` api: https://github.com/influxdata/influxdb/blob/0caee29f249a4404f774cad72646e8461e62a098/http/swagger.yml#L4146
1. `logs` api: https://github.com/influxdata/influxdb/blob/0caee29f249a4404f774cad72646e8461e62a098/http/swagger.yml#L4681

The service should be complete before merge it into master. 

related to: #18389

lyondhill:
@bednar Thank you very much! I will get those in before merging.

PR #18380, opened 2020-06-05
fix: fix points writing to rum endpoint

hoorayimhelping:
the previous writer was in the wrong format, this conforms correctly

- [x] Rebased/mergeable
- [x] Tests pass


hoorayimhelping:
>Just adding the points key/v yah?

yessir, making this conform to the proper shape. i've tested it in a hacky way on staging and feel confident it works all good and stuff.

Issue #18296, opened 2020-05-29
Data Explorer: Clicking submit a query is taking too long


russorat:
This delay is happening in the data explorer whenever i click the submit button. Clicking the button should register instantly for the user.

![image (54)](https://user-images.githubusercontent.com/4805997/83290924-528a8000-a19c-11ea-92bf-fe825e1aaf61.png)

desa:
@Asalem1 does the PR you have solve this issue or just add a cancel button?

Asalem1:
Both. It address the issue that clicking submit should immediately set the button value to a loading state rather than waiting for the query to compile and execute. @russorat I just merged the changes onto master if you wanna see if the issue is resolved for you. If it is, feel free to mark it as closed, otherwise let me know and I'll see about speeding the event up some other way

russorat:
This is what i see in the data explorer in us-west this morning. There is still a noticeable delay when interacting with the button.

![Kapture 2020-06-16 at 8 29 46](https://user-images.githubusercontent.com/4805997/84794850-9c9c9f80-afab-11ea-98b2-dda47f7c2318.gif)

Asalem1:
Hmmm, would you mind sending me a link to a dashboard that's experiencing this? We're setting the state on click so I'm not sure where the delay is coming from. We've got a feature enhancement request coming up to half a second until the `cancel` text displays (so that it doesn't flash so abruptly). I'm wondering if that might help in the perception that the event didn't register?

PR #18387, opened 2020-06-05
feat(cancel-query-request): added abort query request functionality after submitting a query

Asalem1:
Closes:

#18296 

### Problem

This PR aims to address two issues:

1. The perception of slow click events for submitting a query
2. Users don't have the option to cancel their query requests once they've been initiated

### Solution

1. Created a custom thunk that will set the loading state when the click event is triggered
2. By passing in a reference to the AbortController with each request within the component, we are easily able to abort the HTTP request. This will abort the query where it lies, and will prevent any further costs from being incurred based on a user's query duration.

cancellation-in-progress

- [x] Rebased/mergeable


ui/src/timeMachine/actions/queries.ts
```diff
@@ -148,10 +155,14 @@ export const executeQueries = () => async (dispatch, getState: GetState) => {
 
       const extern = buildVarsOption(variableAssignments)
 
-      return runQuery(orgID, text, extern)
+      return runQuery(orgID, text, extern, abortController)
     })
+    const results = await Promise.all(pendingResults.map(r => r.promise)).catch(
+      error => {
```
ebb-tide, line 161: I _think_ if we are catching and then throwing error, it's the same as not catching at all?

Asalem1: That's a great question, you could be right. Let me go ahead and give that another shot. I think I originally added it to help the error bubble out of the promise, but it could be that I had an issue in the catch at the end of all this

Asalem1: Yup, you're totally right. I will ✂️  it out

ui/src/notebooks/context/query.tsx
```diff
@@ -21,6 +21,10 @@ export interface QueryContextType {
   query: (text: string) => Promise<BothResults>
 }
 
+export interface QueryContextType {
```
ebb-tide, line 24: what is the difference between this and the one on line 20?

Asalem1: Literally nothing. Good eyes, looks like I added something that shouldn't have been added

ui/src/timeMachine/actions/queries.ts
```diff
@@ -106,7 +106,14 @@ const isFromBucket = (node: Node) => {
   )
 }
 
-export const executeQueries = () => async (dispatch, getState: GetState) => {
+export const setQueryToLoading = () => dispatch => {
```
ebb-tide, line 109: I don't understand why this is a thunk and not just an action?

ui/src/timeMachine/actions/queries.ts
```diff
@@ -194,13 +201,15 @@ export const executeQueries = () => async (dispatch, getState: GetState) => {
     dispatch(
       setQueryResults(RemoteDataState.Done, files, duration, null, statuses)
     )
-  } catch (e) {
-    if (e.name === 'CancellationError') {
+    return results
+  } catch (error) {
+    if (error.name === 'CancellationError' || error.name === 'AbortError') {
```
ebb-tide, line 206: 👍

ui/src/timeMachine/components/SubmitQueryButton.tsx
```diff
@@ -74,8 +93,24 @@ class SubmitQueryButton extends PureComponent<Props> {
...
+    // In order to allow for requests after cancellations:
+    // https://stackoverflow.com/a/56548348/7963795
+    this.abortController = new AbortController()
+    this.props.onSubmit(this.abortController)
+  }
+
+  private handleCancelClick = (): void => {
+    this.props.onNotify(queryCancelRequest())
+    if (this.abortController) {
+      this.abortController.abort()
```
ebb-tide, line 112: I wonder what happens to abortController after you abort it? I'm wondering if you should be setting abortController to null here as well. not sure.

ui/src/notebooks/components/header/Submit.tsx
```diff
@@ -17,6 +18,7 @@ const COMMENT_REMOVER = /(\/\*([\s\S]*?)\*\/)|(\/\/(.*)$)/gm
 
 export const Submit: FC = () => {
   const {query} = useContext(QueryContext)
+  const {onNotify, setQueryToLoading} = useContext(SubmitQueryButtonContext)
```
drdelambre, line 21: this context doesn't have a provider. what are you trying to do in here?

ui/src/timeMachine/components/SubmitQueryButton.tsx
```diff
@@ -74,8 +93,24 @@ class SubmitQueryButton extends PureComponent<Props> {
     return ComponentStatus.Default
   }
 
+  private abortController: AbortController
+
   private handleClick = (): void => {
-    this.props.onSubmit()
+    // Optimistic UI, set the state to loading when the event
+    // happens rather than when the event trickles down to execution
+    this.props.setQueryToLoading()
```
drdelambre, line 101: doesn't the submit action do this immediately? why add the extra iteration?

ui/src/timeMachine/actions/queries.ts
```diff
@@ -212,9 +221,11 @@ const saveDraftQueries = (): SaveDraftQueriesAction => ({
   type: 'SAVE_DRAFT_QUERIES',
 })
 
-export const saveAndExecuteQueries = () => dispatch => {
+export const saveAndExecuteQueries = (
+  abortController?: AbortController
+) => dispatch => {
   dispatch(saveDraftQueries())
-  dispatch(executeQueries())
```
ebb-tide, line 217: could you dispatch(setQueryToLoading) here instead of as part of the `  private handleClick = (): void => {` method in submit button?

ui/src/timeMachine/actions/queries.ts
Asalem1: Totally can be. Good call

ui/src/timeMachine/components/SubmitQueryButton.tsx
Asalem1: That's a great question. Let me find out

ui/src/notebooks/components/header/Submit.tsx
Asalem1: Basically trying to mock out the functionality that's introduced in wrapping the component to redux. Was a bit confused by what I was actually doing since I just planned to use it as a way of satisfying the linter / following what was happening here

ui/src/timeMachine/components/SubmitQueryButton.tsx
hoorayimhelping: oooh great question! 🤔

ui/src/timeMachine/components/SubmitQueryButton.tsx
Asalem1: It does, but like the comment says, we are doing this as an optimistic UI optimization. The loading state is set after the query is drafted and the query starts to execute. This is just a means of setting the expected state of the button once the button is clicked so that the perception of the action is immediate.

ui/src/timeMachine/actions/queries.ts
Asalem1: That sounds great!

ui/src/timeMachine/components/SubmitQueryButton.tsx
Asalem1: I could be wrong, but according to the docs it looks like once an abort controller is aborted, a rejected promise is returned? Although I'm not entirely certain what the means for the instantiated abortController:

https://dom.spec.whatwg.org/#abortcontroller-api-integration

I don't see why setting the value to `null` would hurt

ui/src/timeMachine/actions/queries.ts
hoorayimhelping: This showed up cause of testing! There are situations where the request can be canceled before the other the `catch` in `runQuery` has a chance to catch it, so it would wind up here without properly being converted into a `CancellationError`

yarn.lock
```diff
@@ -1,4 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
-# yarn lockfile v1
```
ebb-tide, line 2: hrrm wait why is this being generated?

Asalem1: @ebb-tide i removed since I had apparently introduced it by running yarn install at the root directory 7 months ago. Just came across it when I was working in fixing the issue that stemmed from the lock file and figured it should be removed to reduce confusion

PR #18392, opened 2020-06-08
feat(influx): move nested commands to root for stacks, export, and rename pkg to apply

jsteenb2:
this is the beginning of consolidating packages under the moniker templates

the new commands introduced:

```
influx stacks [init | remove]
  - is equivalent to the list command as well as the parent command for stack cmds like init and remove

influx apply
   - is equivalent to the `influx pkg` command
   - all deprecated cli commands still work, but will not show up in help anymore (hidden)
      - i.e. `influx pkg export all` and all its friends will still work

influx export [all | stack]
   - moves the export commands up to root
```

- [x] Rebased/mergeable
- [x] Tests pass


russorat:
@jsteenb2 sounds like a good idea!

PR #18400, opened 2020-06-08
feat: Dashboards maintain sort order after navigating away

### LINKED ISSUE #15292: Dashboard page should remember sort order
as a user
when i sort my dashboard page and then navigate away
then the next time i visit that page, i expect the order to be the same as before.

let's store in user's local data.

reparker837:
Closes #15292

Dashboard page remembers sort order when you go to a different page–does not save across page reloads.

- [X] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/dashboards/actions/creators.ts
+++ b/ui/src/dashboards/actions/creators.ts
@@ -1,5 +1,9 @@
 // Types
-import {RemoteDataState, DashboardEntities} from 'src/types'
+import {
+  RemoteDataState,
+  DashboardEntities,
+  DashboardSortParams,
+} from 'src/types'
 import {NormalizedSchema} from 'normalizr'
 import {setLabelOnResource} from 'src/labels/actions/creators'
 
@@ -10,6 +14,7 @@ export const REMOVE_DASHBOARD = 'REMOVE_DASHBOARD'
 export const REMOVE_DASHBOARD_LABEL = 'REMOVE_DASHBOARD_LABEL'
 export const SET_DASHBOARD = 'SET_DASHBOARD'
 export const SET_DASHBOARDS = 'SET_DASHBOARDS'
+export const SET_DASHBOARD_SORT = 'SET_DASHBOARD_SORT'
 
 export type Action =
   | ReturnType<typeof editDashboard>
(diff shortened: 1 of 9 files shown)


ui/src/dashboards/actions/creators.ts
```diff
@@ -10,6 +14,8 @@ export const REMOVE_DASHBOARD = 'REMOVE_DASHBOARD'
 export const REMOVE_DASHBOARD_LABEL = 'REMOVE_DASHBOARD_LABEL'
 export const SET_DASHBOARD = 'SET_DASHBOARD'
 export const SET_DASHBOARDS = 'SET_DASHBOARDS'
+//add action for setting the dashboard sort info
```
hoorayimhelping, line 17: ✂️

ui/src/dashboards/components/dashboard_index/DashboardsIndex.tsx
```diff
@@ -170,14 +167,17 @@ const mstp = (state: AppState): StateProps => {
   const {
     cloud: {limits},
   } = state
+  const sortOptions = state.resources.dashboards['sortOptions']
```
Asalem1, line 170: Nit: it could be beneficial to use the `getByID` here:

https://github.com/influxdata/influxdb/blob/master/ui/src/resources/selectors/index.ts#L26

ui/src/dashboards/constants/index.ts
```diff
@@ -6,6 +6,10 @@ import {Cell, Dashboard, RemoteDataState} from 'src/types'
 import {DecimalPlaces} from 'src/types'
 import {DEFAULT_TIME_FORMAT} from 'src/shared/constants'
 
+import {SortTypes} from 'src/shared/utils/sort'
+import {Sort} from '@influxdata/clockface'
+import {DashboardSortKey} from 'src/shared/components/resource_sort_dropdown/generateSortItems'
```
Asalem1, line 11: 🤔 it's strange that this type is being set in a decentralized location. It might be beyond the scope of this PR to move this to `src/types` but it is worth considering doing that as a fast follow-up

hoorayimhelping: good suggestion, but I think it's beyond this PR's scope to do that.

PR #18412, opened 2020-06-09
tests(launcher): add a launcher test for read window aggregate push down

### LINKED ISSUE #18184: Create a test to verify in OSS that window aggregate query was indeed pushed down to storage
Similar to https://github.com/influxdata/idpe/issues/6999, we want to have a verification test in OSS. 
We can still read the `query_influxdb_source_read_request_duration_seconds` metric, it's set in influxdb: query/stdlib/influxdata/influxdb/metrics.go

jsternberg:
This adds a launcher test for the read window aggregate push down to
verify that it is done when a query is sent with the appropriate
pattern, the output is correct, and that the metric is incremented that
signals the push down happened.

Fixes #18184.

### DIFF
--- a/cmd/influxd/launcher/query_test.go
+++ b/cmd/influxd/launcher/query_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"html/template"
 	"io"
+	"io/ioutil"
 	"math/rand"
 	nethttp "net/http"
 	"strings"
@@ -15,6 +16,7 @@ import (
 	"time"
 
 	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/csv"
 	"github.com/influxdata/flux/execute"
 	"github.com/influxdata/flux/execute/executetest"
 	"github.com/influxdata/flux/lang"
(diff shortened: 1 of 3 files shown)


cmd/influxd/launcher/query_test.go
```diff
@@ -745,3 +748,79 @@ from(bucket: "%s")
...
+		t.Fatal(err)
+	}
+
+	h, ok := func() (*dto.Histogram, bool) {
+		const metricName = "query_influxdb_source_read_request_duration_seconds"
+		mf := l.Metrics(t)[metricName]
+		for _, m := range mf.Metric {
+			for _, label := range m.Label {
+				if label.GetName() == "op" && label.GetValue() == "readWindowAggregate" {
+					return m.GetHistogram(), true
```
wolffcm, line 813: Did you write this as a function just to avoid using a label?

jsternberg: I think I did it just to avoid the `var found bool` pattern. I don't think I realized that this was a nested for loop at the time. I just like to scope my local variables as much as possible when I can.

cmd/influxd/launcher/query_test.go
```diff
@@ -745,3 +748,79 @@ from(bucket: "%s")
...
+	dec := csv.NewMultiResultDecoder(csv.ResultDecoderConfig{})
+	data := `
+#datatype,string,long,long,string,string,string,dateTime:RFC3339
+#group,false,false,false,true,true,true,false
+#default,_result,,,,,,
+,result,table,_value,_field,_measurement,k,_time
+,,0,5,f,m0,k0,1970-01-01T00:00:05Z
+,,0,5,f,m0,k0,1970-01-01T00:00:10Z
+,,0,5,f,m0,k0,1970-01-01T00:00:15Z
+`
```
fchikwekwe, line 796: Do we maybe wanna consider writing this test so that it can run multiple test cases? You could even just merge as is and I can do it as part of registering the bare aggregate.

jsternberg: I refactored this, but I refactored it mostly to test around all aggregate window calls. I think we can further refactor this to share it amongst the code for the other tests though.

I didn't notice you mentioned bare aggregates until I was already done. Do you want me to refactor it further?

fchikwekwe: No, I can work on the further refactor as part of my PR.

wolffcm:
@fchikwekwe Just as an FYI, I'm also looking to add test cases for `sum` for both the `window` and `group` cases (and maybe the bare case too, depending on whether your work lands before or after mine)

PR #18414, opened 2020-06-09
refactor(storage/reads): refactor/generalize aggregate array cursor code and add tests

wolffcm:
This is a refactoring of the aggregate array cursor code that generalizes aggregation.

I added an `Aggs` field to `array_cursor.gen.go.tmpldata` to specify the initializer and accumulation step for each aggregate function, for each data type supported in InfluxDB fields.

I also added code-generated tests to ensure that cursors are created correctly using the helper methods in `array_cursor.go`.


storage/reads/aggregate_resultset.go
```diff
@@ -44,8 +45,16 @@ func (r *windowAggregateResultSet) Next() bool {
 }
 
 func (r *windowAggregateResultSet) Cursor() cursors.Cursor {
+	agg := r.req.Aggregate[0]
```
jacobmarble, line 48: Is it possible to assemble `ReadWindowAggregateRequest` with zero elements in the `Aggregate` field, and is it possible that such an object arrive here? I'm sure you've thought of this, I'm double-checking.

Perhaps `NewWindowAggregateResultSet()` could check for this and return an error to prevent `Cursor()` from being called in that case.

storage/reads/array_cursor.go
```diff
@@ -18,84 +17,52 @@ func (v *singleValue) Value(key string) (interface{}, bool) {
...
 	case cursors.UnsignedArrayCursor:
-		return newUnsignedWindowCountArrayCursor(cur, req.WindowEvery)
-	case cursors.StringArrayCursor:
-		return newStringWindowCountArrayCursor(cur, req.WindowEvery)
-	case cursors.BooleanArrayCursor:
-		return newBooleanWindowCountArrayCursor(cur, req.WindowEvery)
+		return newUnsignedWindowSumArrayCursor(cur, every)
 	default:
-		panic(fmt.Sprintf("unreachable: %T", cur))
+		panic(fmt.Sprintf("unsupported for aggregate sum: %T", cur))
```
jacobmarble, line 65: Is this strictly unreachable? As this effort matures, it might make sense to return error with the cursor here and above. Just a nit today.

storage/reads/aggregate_resultset.go
yzhang1991: Currently, our pushdown planner rule will ensure that this won't happen. This kind of assumes that the Flux runtime is the only caller of this API. I don't know how true it is.

wolffcm: I think that it's true now. I think adding a check is a good idea and it's easy enough to do, so I will.

storage/reads/array_cursor.go
wolffcm: Yeah, I think we need better error handling all around here. It's tricky to return an error here because the `Cursor()` method is part of an interface, so the interface and all of its implementors would need to be updated. I'd like to defer that work to a later time.

storage/reads/array_cursor.go
```diff
@@ -18,84 +17,52 @@ func (v *singleValue) Value(key string) (interface{}, bool) {
 }
 
 func newAggregateArrayCursor(ctx context.Context, agg *datatypes.Aggregate, cursor cursors.Cursor) cursors.Cursor {
```
yzhang1991, line 19: Since you have the `Agg` section in your template file now, I feel this part of the code should also be generated.

storage/reads/array_cursor_test.go
```diff
@@ -110,7 +109,7 @@ func (a *aggArrayCursorTest) run(t *testing.T) {
 	})
 }
 
-func TestIntegerIntegerCountArrayCursor(t *testing.T) {
+func TestIntegerCountArrayCursor(t *testing.T) {
```
yzhang1991, line 112: Those tests are great!

storage/reads/aggregate_resultset.go
yzhang1991: Yup

PR #18420, opened 2020-06-09
fix: guard against negative stats when non-fired events are 0

hoorayimhelping:
Some of the RUM metrics were reporting negative numbers.

The navigation timing api reports the times that certain events in a sequence happened. To calculate load times, later values are subtracted from earlier values to determine how long an event took. If the later event hasn't finished firing, its value will be 0, resulting in a subtraction looking like `0 - someEarlierValue` leading to a negative value for the metric.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/cloud/utils/rum.ts
+++ b/ui/src/cloud/utils/rum.ts
@@ -34,10 +34,22 @@ export const buildFieldsFromTiming = function buildFieldsFromTiming(
   const totalServer = timing.responseEnd - timing.requestStart
 
   //  UI measurements
-  const domProcessing = timing.domComplete - timing.responseEnd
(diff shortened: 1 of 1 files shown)


hoorayimhelping:
>but should we instead be waiting for everything to complete before sending the metrics? is there a way to do that?

i think that's a good idea, but it's something i'd like to do later. there are two ways to handle this, firing when [window.load](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event) is fired, and try using [a beacon.](https://developer.mozilla.org/en-US/docs/Web/API/Beacon_API)

i think `window.load` is probably the best bet, but i don't know what the effects of it will be and i haven't had time to test it, and i need to work on other things before i can do that, so this is a quick fix

PR #18429, opened 2020-06-10
fix(httpd): add user authentication for prometheus remote read api

### LINKED ISSUE #18328: Prometheus remote read API missing user authentication
For 1.x, prometheus remote read API can accept username/password but influxdb server only checks the password but does not check the privilege against the target database.This means a user can access any database even though it doesn't have read privilege on.

foobar:
Closes #18328

### DIFF
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -1169,6 +1169,8 @@ func (h *Handler) servePromWrite(w http.ResponseWriter, r *http.Request, user me
 // servePromRead will convert a Prometheus remote read request into a storage
 // query and returns data in Prometheus remote read protobuf format.
 func (h *Handler) servePromRead(w http.ResponseWriter, r *http.Request, user meta.User) {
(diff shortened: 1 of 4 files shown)


foobar:
@russorat can you evaluate if this needs to be ported back to 1.7?

services/httpd/handler.go
```diff
@@ -1191,6 +1193,17 @@ func (h *Handler) servePromRead(w http.ResponseWriter, r *http.Request, user met
 	db := r.FormValue("db")
 	rp := r.FormValue("rp")
 
+	if h.Config.AuthEnabled {
```
jacobmarble, line 1196: I agree that the endpoint should be protected with auth. However, this PR could break existing usage where the endpoint is used without auth.

Can you add an option similar to what pprof and ping were given [in this change](https://github.com/influxdata/influxdb/commit/0c4703aaa3cd2cbb245982effd53858cb7443bd9)? I would certainly like to merge this with such an adjustment.

PR #18437, opened 2020-06-10
chore: permanently remove authorizations from max permissions

GeorgeMac:
This is a clean-up PR.

I have built influxd locally and I don't think we need this now.
It was left in for OSS, under the speculation that something depended on it. It was suspected for bootstrapping a new cluster. However, I have managed to bootstrap a cluster with no problems and gone on to use the app successfully.

UPDATE:

There is a test using the initial users session to do an API call to perform a privileged action (create a user). Instead of using the onboarded initial users magical access all areas token.

So the question is... does this user really need those permissions in its session (not token)?
Since I don't believe any of the UI based operations allow this kind of management anyway.

- [x] Rebased/mergeable
- [x] Tests pass


danxmoran:
Closing as stale

PR #18447, opened 2020-06-10
feat(dashboards): improve view spinners

### LINKED ISSUE #18323: Provide an indication to the user that the dashboard cells are still loading
when a user changes the time range on a dashboard, there is no indication which cells are still loading, and which ones are finished.

alexpaxton:
Closes #18323 

Upon investigating the issue we found that there is in fact a loading indicator but it had a few problems. One being that it was extremely subtle and the other being it only displayed on some view types and not on others.

### Our solution:

graph-with-spinner
Redesign the loading state to be way more obvious

graph-with-spinner-timer
If the view is loading for more than `5s` a timer appears to additionally display load time

- Render loading indicator in `RefreshingGraph` instead of individually in each view component
  - Removed a lot of prop drilling
- Redesign & rename loading indicator component
- Start a timer each time the view changes to `Loading` and if it stays in `Loading` state longer than 5s reveal the duration

- [x] Rebased/mergeable
- [x] Tests pass


TCL735:
🔥

Issue #18234, opened 2020-05-26
Landing Page


hoorayimhelping:
# Background
As a user, I want to be able to browse and create new resources from templates, as well as see what templates I've used in the past

# Acceptance Criteria
* [x] I can click a link to view the list of community templates in influxdb github repo
  * Open the link in a new tab/window
* [x] I can choose a template to upload a resource
  * [x] Can use URL field

rickspencer3:
Is there not a significant amount of effort and design implied by:

> I can see my list of my currently installed templates

Does this not imply that there is a source of truth for what templates are implied? This opens a host of questions about how to maintain consistency between that source of truth and changes that the user may make. For example, what happens if a user deletes all of the resources associated with the template?

I think templates are not objects that can managed in this manner, but are rather one way injectors which some ability to not overwrite previously injected objects.

jsteenb2:
@rickspencer3 we have a source of truth for installed templates already. This endpint exists as well. Teh templates are associated with stacks by default. Teh stack can be multiple templates/packages, but it is a record of those applications.

PR #18452, opened 2020-06-10
feat(flagged): wireup simple url and name overlay

hoorayimhelping:
Connects #18234

Infers the name of the template from the github url, launches the first part of the new template overlay. It's still not functional, but it provides a starting point to work from.

It's behind feature flag `communityTemplates`.

To access this flag locally, start your app with:
```sh
```

Screen Shot 2020-06-10 at 4 20 04 PM

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/templates/components/CommunityTemplateImportOverlay.tsx
+++ b/ui/src/templates/components/CommunityTemplateImportOverlay.tsx
@@ -1,0 +1,92 @@
+import React, {PureComponent} from 'react'
+import {withRouter, WithRouterProps} from 'react-router'
+import {connect} from 'react-redux'
+
+// Components
+import {TemplateInstallerOverlay} from 'src/templates/components/TemplateInstallerOverlay'
+
+// Actions
+import {createTemplate as createTemplateAction} from 'src/templates/actions/thunks'
+import {notify as notifyAction} from 'src/shared/actions/notifications'
+
+// Types
+import {AppState, Organization, ResourceType} from 'src/types'
+import {ComponentStatus} from '@influxdata/clockface'
+
+// Utils
+import {getByID} from 'src/resources/selectors'
+
+interface State {
+  status: ComponentStatus
+}
+
+interface DispatchProps {
+  createTemplate: typeof createTemplateAction
+  notify: typeof notifyAction
+}
+
+interface StateProps {
+  org: Organization
+  templateName: string
+}
+
+interface OwnProps extends WithRouterProps {
+  params: {orgID: string; templateName: string}
+}
+
+type Props = DispatchProps & OwnProps & StateProps
+
+class UnconnectedTemplateImportOverlay extends PureComponent<Props> {
+  public state: State = {
+    status: ComponentStatus.Default,
+  }
+
+  public render() {
+    return (
+      <TemplateInstallerOverlay
+        onDismissOverlay={this.onDismiss}
+        onSubmit={this.handleInstallTemplate}
+        status={this.state.status}
+        templateName={this.props.templateName}
+        updateStatus={this.updateOverlayStatus}
+      />
+    )
+  }
+
+  private onDismiss = () => {
+    const {router} = this.props
+
+    router.goBack()
+  }
+
+  private updateOverlayStatus = (status: ComponentStatus) =>
+    this.setState(() => ({status}))
+
+  private handleInstallTemplate = (importString: string) => {
+    importString
+  }
+}
+
+const mstp = (state: AppState, props: Props): StateProps => {
+  const org = getByID<Organization>(
+    state,
+    ResourceType.Orgs,
+    props.params.orgID
+  )
+
+  return {org, templateName: props.params.templateName}
+}
+
+const mdtp: DispatchProps = {
+  createTemplate: createTemplateAction,
+  notify: notifyAction,
+}
+
+export const CommunityTemplateImportOverlay = connect<
+  StateProps,
+  DispatchProps,
+  Props
+>(
+  mstp,
+  mdtp
+)(withRouter(UnconnectedTemplateImportOverlay))
--- a/ui/src/templates/containers/CommunityTemplatesIndex.tsx
+++ b/ui/src/templates/containers/CommunityTemplatesIndex.tsx
@@ -23,6 +23,10 @@ import SettingsHeader from 'src/settings/components/SettingsHeader'
 // Utils
 import {pageTitleSuffixer} from 'src/shared/utils/pageTitles'
 import {getOrg} from 'src/organizations/selectors'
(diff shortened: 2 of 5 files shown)


ui/src/templates/components/CommunityTemplateImportOverlay.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const {router} = this.props
+
+    router.goBack()
+  }
+
+  private updateOverlayStatus = (status: ComponentStatus) =>
+    this.setState(() => ({status}))
+
+  private handleInstallTemplate = (importString: string) => {
+    importString
```
hoorayimhelping, line 66: not functional right now, it would complain if it was an empty function body or a `console.log` statement 🙄

ui/src/templates/containers/CommunityTemplatesIndex.tsx
```diff
@@ -34,14 +38,28 @@ interface StateProps {
...
-type Props = WithRouterProps & StateProps
+interface OwnProps extends WithRouterProps {
+  params: {templateName: string}
+}
+
+type Props = OwnProps & StateProps
 
 @ErrorHandling
-class CTI extends Component<Props> {
+class UnconnectedTemplatesIndex extends Component<Props> {
```
jsteenb2, line 48: what does `Unconnected` signal?

ui/src/templates/containers/CommunityTemplatesIndex.tsx
```diff
@@ -34,14 +38,28 @@ interface StateProps {
...
   state = {
     currentTemplate: '',
   }
 
+  public componentDidMount() {
+    if (this.props.params.templateName) {
+      this.setState({
+        currentTemplate: getGithubUrlFromTemplateName(
```
jsteenb2, line 56: is the input limited to github? I didn't' think it was

ui/src/templates/containers/CommunityTemplatesIndex.tsx
```diff
@@ -105,9 +123,19 @@ class CTI extends Component<Props> {
     )
   }
 
-  private handleImport = () => {
+  private startTemplateInstall = () => {
+    if (!this.state.currentTemplate) {
+      console.error('undefined')
```
jsteenb2, line 128: is this console.error a stub for now?

ui/src/templates/utils/index.ts
```diff
@@ -73,3 +73,20 @@ export const getLabelRelationships = (resource: {
 
 export const getIncludedLabels = (included: {type: TemplateType}[]) =>
   included.filter((i): i is LabelIncluded => i.type === TemplateType.Label)
+
+// See https://github.com/influxdata/community-templates/
+// an example of a url that works with this function:
+// https://github.com/influxdata/community-templates/tree/master/csgo
+export const getTemplateNameFromGithubUrl = (url: string): string => {
+  if (!url.includes('https://github.com/influxdata/community-templates/')) {
```
jsteenb2, line 81: why is this? templates can come from anywhere

ui/src/templates/containers/CommunityTemplatesIndex.tsx
hoorayimhelping: https://react-redux.js.org/api/connect

we return `CommunityTemplatesIndex` as part of the public API of this. `UnconnectedTemplatesIndex` is just the plain React component that hasn't had redux injected into it.

ui/src/templates/containers/CommunityTemplatesIndex.tsx
hoorayimhelping: yep

ui/src/templates/containers/CommunityTemplatesIndex.tsx
hoorayimhelping: i don't think it is, but right now for the mvp version 1, we're using https://github.com/influxdata/community-templates/, so that's what i'm targeting. yagni and what have you.

ui/src/templates/utils/index.ts
hoorayimhelping: for the first version, i only want to target this repo: https://github.com/influxdata/community-templates/ and i'm paranoid of accepting something that we can't parse

jsteenb2: so the pkgs the UI will never parse, at least not in the next long while. The chellange with poeple going to influx, is they need to know to provide the `https://raw.githubusercontent.com/influxdata/community-templates/master/influxdb2_oss_metrics/influxdb2_oss_metrics.yml` instead of the plain `https://github.com` one. I raised this question in our dev inception, but I didn't hear any firm direction for if we will be translating github links to retrieve the raw contents or not, and if so, who does it 🤷.

ui/src/templates/containers/CommunityTemplatesIndex.tsx
jsteenb2: see comment below

ui/src/templates/utils/index.ts
hoorayimhelping: @jsteenb2 thanks for the explanation, that makes sense.

I think it makes most sense for the UI to handle this complexity. Ideally, we'd accept a whole bunch of different link formats and file formats, and as long as we can infer the name from what the user provides us, the UI should be fine translating that into the interface the backend expects. 

seems like the a good way to split responsibilities - part of the job of the UI is to make it easy for people to get data into the backend engine. if people want to do a lot of work repeatedly or quickly, they can use the more constrained cli api and write scripts to do that. if they just want to drag some random resource they got from influx into our UI to experiment or practice or just because they're less technical, that's also supported and doable.

jsteenb2: can you describe the specifics of what you are describing? I'm not following what you are proposing

hoorayimhelping: @jsteenb2 if a user gives us this url `https://github.com/influxdata/community-templates/tree/master/csgo` or this url https://github.com/influxdata/community-templates/blob/master/csgo/csgo.yml, or basically any link inside this directory: https://github.com/influxdata/community-templates/tree/master/csgo, we should be able to figure out they're trying to give us the yaml file.

jsteenb2: @hoorayimhelping, this is going to require significant investment of engineering effort to accomplish. The example you gave, may or may not hold weight. Not all templates are created this way. Take a gander at the community templates. They aren't all the same. Some are json, some are yaml, some have different namign conventions. Its more open that what you eluded too. 

I'd say punt on this complexity altogether. It'd be easier to just inform the user of what specific url we are looking for.

PR #18460, opened 2020-06-11
feat(flows): draggable results panels

alexpaxton:
- Enables results panels to be show/hidden
- Enables results panels to be resized to any height (instead of the arbitrary 3 sizes)

Had to add internal state to the dragging components to make the interaction smoother and to avoid updating context too much.

There's a couple things I wanted to do but I think the work is better suited to a different PR:

- Make visualizations resizable and visibility toggle with similar UX to results panel
- Extract out resize components into notebooks shared

- [x] Rebased/mergeable
- [x] Tests pass


ui/src/notebooks/pipes/Query/ResultsResizer.tsx
```diff
@@ -0,0 +1,133 @@
+// Libraries
+import React, {FC, useRef, useEffect, ReactNode, useState} from 'react'
+import {round} from 'lodash'
```
drdelambre, line 3: i'm drawing a line right here....
use host functions, they're faster, free on resources, and you're not going to need to use hundredths of a pixel, get lodash out of here
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round

alexpaxton: Why is there no "grimace" emoji in github??

ui/src/notebooks/pipes/Query/view.tsx
```diff
@@ -40,10 +47,16 @@ const Query: FC<PipeProp> = ({data, onUpdate, Context, results}) => {
           onSubmitScript={() => {}}
           autogrow
         />
-        <Results results={results} size={size} onUpdateSize={onUpdateSize} />
+        <Results
+          results={results}
+          height={resultsHeight}
+          onUpdateHeight={handleUpdateResultsHeight}
```
drdelambre, line 53: why isn't height, onUpdateHeight, visibility, and onUpdateVisibility self contained within the results component? it's strange to route the data in and out like that instead of just routing the meta data into the results component

ui/src/notebooks/pipes/Query/Results.tsx
```diff
@@ -2,58 +2,67 @@
...
-        size={size}
-        onUpdateSize={onUpdateSize}
-      />
-      {resultsBody}
+    <div className="notebook-raw-data">
+      <ResultsResizer
+        visibility={visibility}
+        height={height}
+        onUpdateHeight={onUpdateHeight}
+        onUpdateVisibility={onUpdateVisibility}
```
drdelambre, line 61: and why are these being routed through the component as well?

ui/src/notebooks/pipes/Query/view.tsx
alexpaxton: Should results just accept `data` and `onUpdate` as props instead of breaking them down first?

ui/src/notebooks/pipes/Query/view.tsx
```diff
@@ -15,7 +14,8 @@ import 'src/notebooks/pipes/Query/style.scss'
 const Query: FC<PipeProp> = ({data, onUpdate, Context, results}) => {
   const {queries, activeQuery} = data
   const query = queries[activeQuery]
-  const size = data.rawDataSize || 'small'
+  const resultsVisibility = data.resultsVisibility || 'visible'
```
drdelambre, line 17: can't we remove these too now?

ui/src/notebooks/pipes/Query/Results.tsx
```diff
@@ -2,58 +2,70 @@
...
-    <div className={className}>
-      <ResultsHeader
-        resultsExist={resultsExist}
-        size={size}
-        onUpdateSize={onUpdateSize}
-      />
-      {resultsBody}
+    <div className="notebook-raw-data">
+      <ResultsResizer
+        visibility={visibility}
```
drdelambre: can change this one to use the "data / onUpdate" as well?

ui/src/notebooks/pipes/Query/ResizerHeader.tsx
```diff
@@ -0,0 +1,66 @@
...
+interface Props {
+  visibility: ResultsVisibility
+  onUpdateVisibility: (visibility: ResultsVisibility) => void
+  onStartDrag: () => void
+  resizingEnabled: boolean
+  dragHandleRef: RefObject<HTMLDivElement>
+}
+
+const ResultsHeader: FC<Props> = ({
+  visibility,
```
drdelambre, line 20: i'm conflicted about leaving this because it breaks the pattern, but lets just leave it until we add another property from data

PR #18461, opened 2020-06-11
chore(influx): remove client timeout on write cmd's retriever

jsteenb2:
also adds skip for flakey test identified in issue: #18463
- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influx/write.go
+++ b/cmd/influx/write.go
@@ -9,7 +9,6 @@ import (
 	"net/url"
 	"os"
 	"strings"
-	"time"
 
 	platform "github.com/influxdata/influxdb/v2"
 	ihttp "github.com/influxdata/influxdb/v2/http"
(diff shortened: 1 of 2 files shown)


cmd/influx/write.go
```diff
@@ -151,13 +150,17 @@ func (writeFlags *writeFlagsType) createLineReader(cmd *cobra.Command, args []st
...
 	if len(writeFlags.URLs) > 0 {
-		client := http.Client{Timeout: 5 * time.Minute}
+		client := http.DefaultClient
 		for _, addr := range writeFlags.URLs {
 			u, err := url.Parse(addr)
 			if err != nil {
 				return nil, csv2lp.MultiCloser(closers...), fmt.Errorf("failed to open %q: %v", addr, err)
 			}
-			resp, err := client.Get(addr)
+			req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, addr, nil)
```
sranka, line 159: The default infinite timeout should be here from beginning, blame on me :) IMHO it would be probably better to require a context as a new parameter of `createLineReader` function and reuse the context with signals that is created in https://github.com/influxdata/influxdb/blob/9c779443d0c2947d6473f5599e1a7899c393cc00/cmd/influx/write.go#L295

jsteenb2: @sranka good points. I'm not super concerned about it as the OS will clean it all up, but is still something worth adding 👍

PR #18467, opened 2020-06-11
feat: cursors for first selection

jlapacik:
This PR adds windowed and non-windowed cursors for the `first` selector. Note that `last` is implemented using `first` cursors.

The table iterators still need to be updated in order to push down windowed first and last, but these changes are enough to implement grouped first and last pushdowns.


storage/reads/array_cursor.gen.go.tmpl
```diff
@@ -188,6 +210,103 @@ func (c *{{.name}}ArrayCursor) nextArrayCursor() bool {
...
+	cursors.{{.Name}}ArrayCursor
+	every int64
+	final int64
+	res {{$arrayType}}
+	tmp {{$arrayType}}
+}
+
+func new{{.Name}}WindowFirstArrayCursor(cur cursors.{{.Name}}ArrayCursor, every int64) *{{.name}}WindowFirstArrayCursor {
+	if every == 0 {
+		every = math.MaxInt64
```
wolffcm, line 252: I opted for using the zero value to represent "one big window" instead of `math.MaxInt64` for the other windowed aggregates. You might consider following that example here, for consistency's sake. @yzhang1991 and I had a long discussion on it:
https://github.com/influxdata/influxdb/pull/18354#discussion_r435667209
https://github.com/influxdata/influxdb/pull/18414#discussion_r437653943

storage/reads/aggregate_resultset.go
```diff
@@ -32,11 +32,16 @@ func NewWindowAggregateResultSet(ctx context.Context, req *datatypes.ReadWindowA
 		return nil, errors.Errorf(errors.InternalError, "attempt to create a windowAggregateResultSet with %v aggregate functions", nAggs)
 	}
 
+	ascending := true
+	if len(req.Aggregate) > 0 && req.Aggregate[0].Type == datatypes.AggregateTypeLast {
+		ascending = false
```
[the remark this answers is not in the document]
jlapacik: It would as the logic for this is located in the `tsdb` package. Storage will create two types of TSM cursors, ascending or descending, depending on the value that is set here. Launcher and pipeline tests however do get us test coverage over this part.

storage/reads/array_cursor.gen.go.tmpl
jlapacik: Thanks, will update.

storage/reads/aggregate_resultset.go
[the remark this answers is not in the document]
wolffcm: Ok, that makes sense.

jlapacik:
The feature flags already exist. They were added as part of the planner rule. But you're right, flux end to end tests do not exist. I will add them.

storage/reads/array_cursor.gen.go.tmpl
```diff
@@ -15,6 +15,28 @@ const (
 	MaxPointsPerBlock = 1000
 )
 
+func newFirstArrayCursor(cur cursors.Cursor) cursors.Cursor {
```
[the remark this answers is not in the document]
jlapacik: Limit works for me.

jlapacik:
@wolffcm I updated the PR. There's a known issue with the series order returned by `ReadGroup` and so I haven't added any flux end to end tests, but I'll create an issue to add them in later.

storage/reads/array_cursor.gen.go
```diff
@@ -286,9 +286,6 @@ type floatWindowLimitArrayCursor struct {
 }
 
 func newFloatWindowLimitArrayCursor(cur cursors.FloatArrayCursor, every int64) *floatWindowLimitArrayCursor {
-	if every == 0 {
```
wolffcm: I would think that instead of moving this logic into the `Next()` method itself, you could just create a non-windowed cursor here?

cmd/influxd/launcher/query_test.go
```diff
@@ -771,87 +881,92 @@ m0,k=k0 f=5i 12000000000
...
-
-			dec := csv.NewMultiResultDecoder(csv.ResultDecoderConfig{})
-			want, err := dec.Decode(ioutil.NopCloser(strings.NewReader(tt.res)))
-			if err != nil {
-				t.Fatal(err)
-			}
-			defer want.Release()
+	}
 
-			if err := executetest.EqualResultIterators(want, got); err != nil {
```
wolffcm, line 848: This test used to compare the iterators instead of just comparing the raw CSV, which is more fragile. My understanding was that in OSS we didn't have problems with non-determinism comparing with result iterators. Is it possible to change this back?

storage/reads/array_cursor.gen.go
jlapacik: I do create a non-windowed cursor if `every == 0`, however I still check that `every != 0` inside `Next()` as that code will panic if the zero value is used.

PR #18468, opened 2020-06-11
feat: save dashboards from notebooks

drdelambre:
still needs:
interaction design
call on "does query follow editor, or visualization?"


ui/src/notebooks/components/header/Submit.tsx
```diff
@@ -19,73 +19,87 @@ export const Submit: FC = () => {
...
 
   useEffect(() => {
     submit()
   }, [!!time && time.range])
 
   const submit = () => {
-    pipes
-      .reduce((stages, pipe, index) => {
-        updateMeta(index, {loading: RemoteDataState.Loading} as PipeMeta)
+    setLoading(RemoteDataState.Loading)
```
drdelambre, line 30: this sets a loading state on the submit button

PR #18480, opened 2020-06-12
feat(cmdClick-task): Added command click functionality to tasks names to open in separate tab

Asalem1:
### Problem

Working with tasks requires a lot of back and forth

### Solution

Integrate command+click functionality on task names so that users can open and browse multiple tasks in different tabs

command-click-task

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## v2.0.0-beta.13 [unreleased]
+
+### Features
+
+1. [18480](https://github.com/influxdata/influxdb/pull/18480): Allows tasks to open in new tabs
+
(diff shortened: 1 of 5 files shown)


Asalem1:
@drdelambre as in why not just remove the action and push to that location in general?

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## v2.0.0-beta.13 [unreleased]
+
+### Features
+
+1. [18480](https://github.com/influxdata/influxdb/pull/18480): Allows tasks to open in new tabs
```
hoorayimhelping, line 5: i don't think this is a feature, it's like, table stakes of a browser.

if it has to be in the changelog, i think it should go under ` ### UI Enhancements`

PR #18482, opened 2020-06-12
fix: clean up old runs when we fail to enqueue it

### LINKED ISSUE #18481: Ensure run is finished when executor fails to create promise
When the executor fails to "createPromise" it doesn't currently finish the run.
This leaves the run in an incomplete state and doesn't clear it from the KV store and archive it into storage.

https://github.com/influxdata/influxdb/blob/76cd40da385f0ceddf71da6657d9edb2e177ddf6/task/backend/executor/executor.go#L212-L219

We need to observe any non-nil error from `createPromise(...)` and call `TaskControlService.FinishRun()` on the run with an appropriate message / status (failed due to inability to enqueue work).

We observed a failure to `FindTaskByID` within `createPromise` caused runs to be left behind.

lyondhill:
We need to make sure the runs are removed if we fail to enqueue the run to be worked.
for this we will now add an error message stating why we failed and move the run out of our kv store.

Closes #18481

### DIFF
--- a/task/backend/executor/executor_test.go
+++ b/task/backend/executor/executor_test.go
@@ -571,7 +571,7 @@ func TestPromiseFailure(t *testing.T) {
 	}
 
 	if len(runs) != 1 {
-		t.Fatalf("expected 0 runs on failed promise: got: %d, %#v", len(runs), runs[0])
+		t.Fatalf("expected 1 runs on failed promise: got: %d, %#v", len(runs), runs[0])
 	}
 
 	if runs[0].Status != "failed" {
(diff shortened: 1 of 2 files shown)


task/backend/executor/executor_test.go
```diff
@@ -535,6 +535,51 @@ func testErrorHandling(t *testing.T) {
...
+		t.Fatalf("expected no promise but recieved one: %+v", promise)
+	}
+
+	runs, _, err := tes.i.FindRuns(context.Background(), influxdb.RunFilter{Task: task.ID})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(runs) != 1 {
+		t.Fatalf("expected 0 runs on failed promise: got: %d, %#v", len(runs), runs[0])
```
GeorgeMac, line 574: nit: `expected 1 run ...`

PR #18484, opened 2020-06-12
feat(query): add test and feature flag for pushing down sum

wolffcm:
This adds the storage capability in OSS for pushing down `sum` in `ReadWindowAggregate`, and associated tests and feature flag.

### DIFF
--- a/query/stdlib/influxdata/influxdb/rules.go
+++ b/query/stdlib/influxdata/influxdb/rules.go
@@ -684,7 +684,7 @@ func canPushWindowedAggregate(ctx context.Context, fnNode plan.Node) bool {
 	}
 
 	// Check the aggregate function spec. Require operation on _value. There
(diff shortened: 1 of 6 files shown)


query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -684,7 +684,7 @@ func canPushWindowedAggregate(ctx context.Context, fnNode plan.Node) bool {
 	}
 
 	// Check the aggregate function spec. Require operation on _value. There
-	// are two feature flags covering all cases. One specifically for Count,
+	// are two feature flags covering all cases. One specifically for Count, one for Sum,
```
jsternberg, line 687: Can you change this comment to something more generic like:

```
// Check the aggregate function spec. Require the operation on _value
// and check the feature flag associated with the aggregate function.
```

This way we don't have to update the comment each time we make a change to the feature flags.

PR #18486, opened 2020-06-12
feat(flows): draggable visualizations

alexpaxton:
- Make `Resizer` and `ResizerHandle` into generic components and move to `notebooks/shared/`
- Implement resizing on Visualization panels

Screen Shot 2020-06-12 at 10 10 31 AM
Screen Shot 2020-06-12 at 10 10 38 AM
Screen Shot 2020-06-12 at 10 10 47 AM

Visualization without data:
Screen Shot 2020-06-12 at 10 10 59 AM

- [x] Rebased/mergeable
- [x] Tests pass


ui/src/notebooks/pipes/Query/index.ts
```diff
@@ -2,15 +2,13 @@ import {register} from 'src/notebooks'
...
-export type Visibility = 'visible' | 'hidden'
-
 register({
   type: 'query',
   component: View,
   button: 'Custom Script',
   initial: {
-    resultsVisibility: 'visible',
-    resultsPanelHeight: 200,
+    panelVisibility: 'visible',
```
drdelambre, line 10: this resizing is a weird concept in here. like, is it notebook specific? but it's not shared by all panels... and a panel in the query context is a child of the notebook panel, but in a visualization, it's the panel itself? most of my hangups come from not having a set of clear definitions around this

ui/src/notebooks/pipes/Visualization/view.tsx
```diff
@@ -89,25 +98,33 @@ const Visualization: FC<PipeProp> = ({
...
-              theme="dark"
-            />
-          </EmptyQueryView>
+      <Resizer
+        data={data}
+        onUpdate={onUpdate}
+        resizingEnabled={!!results.raw}
+        emptyText="No data to visualize"
+        emptyIcon={IconFont.BarChart}
+        toggleVisibilityEnabled={false}
```
drdelambre, line 107: dope

ui/src/notebooks/pipes/Query/index.ts
alexpaxton: Same same. I briefly considered moving it into meta or even its own context but did not because only some types would use it

ui/src/notebooks/shared/Resizer.tsx
```diff
@@ -0,0 +1,175 @@
...
+  /** Icon to display in header when resizing is disabled */
+  emptyIcon?: IconFont
+  /** Text to display when resizing is disabled */
+  emptyText: string
+  /** Text to display when the resizer is collapsed */
+  hiddenText?: string
+  /** When resizing is enabled the panel cannot be resized below this amount */
+  minimumHeight?: number
+  /** Renders this element beneath the visibility toggle in the header */
+  additionalControls?: JSX.Element | JSX.Element[]
```
drdelambre, line 34: the idea being we're gonna add a gear icon for view options? it be better to have a more rigid interface description here if that's the case. like an array of "icon, enabled, onclick, hover text" if it's just all going to be the same buttons, with small differences?

ui/src/notebooks/shared/Resizer.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+  if (resizingEnabled && visibility === 'hidden') {
+    body = <div className="panel-resizer--empty">{hiddenText}</div>
+  }
+
+  return (
+    <div className="panel-resizer">
+      <ResizerHeader
+        emptyIcon={emptyIcon}
+        visibility={visibility}
```
drdelambre, line 160: this shouldn't be passed down

ui/src/notebooks/shared/Resizer.tsx
```diff
@@ -0,0 +1,175 @@
...
+        onStartDrag={handleMouseDown}
```
drdelambre, line 161: why handle the drag logic here, and not in the child component?

ui/src/notebooks/shared/Resizer.tsx
```diff
@@ -0,0 +1,175 @@
...
+        dragHandleRef={dragHandleRef}
+        resizingEnabled={resizingEnabled}
+        additionalControls={additionalControls}
+        onUpdateVisibility={handleUpdateVisibility}
```
drdelambre, line 165: this shouldn't be passed down

drdelambre:
there's a lot of prop drilling going on, can we tighten that up? would it also help if we created a "CurrentPipe" context and provider for all of the NotebookPipes? then we can decouple that drilling even more be removing the "data/onUpdate" props

ui/src/notebooks/shared/Resizer.tsx
drdelambre: note: this is going to be addressed in a clean up pass when adding the "CurrentPipe" context

ui/src/notebooks/shared/Resizer.tsx
drdelambre: note: this is going to be addressed in a clean up pass when adding the "CurrentPipe" context

Issue #15632, opened 2019-10-29
DateRangePicker should not allow for invalid submissions


Asalem1:
There are currently three issues that are not accounted for when submitting invalid date ranges in the DateRangePicker:

1. Users can submit date ranges where the start date is *after* the stop date without encountering any issues. 
2. If a user types an invalid start/stop date in the input field, the DateRangePicker will default the date for that field to the date selected in `ReactDatePicker`.
  a. A solution would be to either disable the `Apply Time Range` button until a valid input is submitted
3. If a user types in an invalid start/stop date in the input field, then selects a date on the `ReactDatePicker`, the input field is not updated to reflect the selection, which still displays the incorrect values and error message

Asalem1:
![daterangeerror](https://user-images.githubusercontent.com/19984220/67794457-e8f59800-fa39-11e9-9be4-142df9fc4daf.gif)

Asalem1:
Here are some incomplete tests that were being written to test out the functionality:

https://github.com/influxdata/influxdb/blob/42d84e410384dcb70cfd5c41dec6fdfa3fe8e1c5/ui/cypress/e2e/explorer.test.ts#L65

https://github.com/influxdata/influxdb/blob/42d84e410384dcb70cfd5c41dec6fdfa3fe8e1c5/ui/cypress/e2e/explorer.test.ts#L91

PR #18491, opened 2020-06-12
chore(min-max/customtimerange): set min & max dates for start & vend time respectively

Asalem1:
Sets min & max dates for max & min date picker (respectively). This means that users cannot select a start date that is after the end date, and users cannot select an end date that is after the start date.
Caveat to all of this is that these limitations are not respected when it comes to setting the time. This is an issue with the dependency that I filed here:

https://github.com/Hacker0x01/react-datepicker/issues/2184

This PR address some of the concerns listed here:

#15632

### DIFF
--- a/ui/cypress/e2e/explorer.test.ts
+++ b/ui/cypress/e2e/explorer.test.ts
@@ -242,7 +242,7 @@ describe('DataExplorer', () => {
   })
 
   describe('select time range to query', () => {
-    it('can select different time ranges', () => {
+    it('can set a custom time range and restricts start & stop selections relative to start & stop dates', () => {
       // find initial value
       cy.get('.cf-dropdown--selected')
         .contains('Past 1')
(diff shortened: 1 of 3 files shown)


ui/cypress/e2e/explorer.test.ts
```diff
@@ -270,6 +270,39 @@ describe('DataExplorer', () => {
...
+
+        // Select the 30th in the Start timerange
+        cy.get('.react-datepicker__day--030')
+          .first()
+          .should('be', 'disabled')
+
+        // Select the 28th in the Stop timerange
+        cy.get('.react-datepicker__day--028')
+          .last()
+          .should('be', 'disabled')
```
121watts, line 303: This could probably be added to another test that selects the time range.  Also, a link to the bug would be 👍

Asalem1: It's not a bug, more a weird UI quirk where we allow start times to be before the start time. It's associated with the issue in the PR comments, but that's a pretty wide-ranging issue.

#15632

PR #18494, opened 2020-06-12
fix: update formatStatValue to correctly handle trailing zeroes in all cases

### LINKED ISSUE #18493: Gauge incorrectly formats the decimal places for the tick labels
In a Gauge, if user selects decimal places, for example 4, none of the tick labels display any decimal places _except_ the first label. All other labels have no decimal places. This is incorrect. A user expects that all numeric values in the labels follow the chosen decimal places.

Screen Shot 2020-06-12 at 2 49 31 PM

TCL735:
Closes #18493 

The util responsible for handling the formatting of single stat values has some logic holes in it. Trailing zeroes were being truncated. This addresses the issue with unit tests.

Screen Shot 2020-06-12 at 2 48 45 PM

### DIFF
--- a/ui/src/shared/utils/formatStatValue.test.ts
+++ b/ui/src/shared/utils/formatStatValue.test.ts
@@ -1,0 +1,231 @@
+import {formatStatValue} from './formatStatValue'
+
+describe('formatStatValue', () => {
+  let prefix: string
+  let value: number | string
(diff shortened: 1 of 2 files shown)


TCL735:
This is an important short-term fix. In the future, giraffe will handle this internally.

PR #18495, opened 2020-06-12
fix: loads function writeNavigationTimingMetrics after window.load runs

### LINKED ISSUE #18421: fire navigation timing metrics after window load
Deniz [brought up](https://github.com/influxdata/influxdb/pull/18420#issuecomment-641540686) a good point that it's probably smarter to wait until the page is finished loading before firing metrics that depend on page load numbers existing.

#### Acceptance Criteria
  * `writeNavigationTimingMetrics` is fired after the `window.load` event is fired.

zoesteinkamp:
Closes #18421

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/src/index.tsx
+++ b/ui/src/index.tsx
@@ -546,6 +546,6 @@ if (rootNode) {
   render(<Root />, rootNode)
 }
 
-window.addEventListener('load', _event => {
+window.addEventListener('load', () => {
   writeNavigationTimingMetrics()
 })
--- a/ui/src/index.tsx
+++ b/ui/src/index.tsx
@@ -546,4 +546,6 @@ if (rootNode) {
   render(<Root />, rootNode)
 }
 
-writeNavigationTimingMetrics()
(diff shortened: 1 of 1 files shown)


ui/src/index.tsx
```diff
@@ -546,4 +546,6 @@ if (rootNode) {
   render(<Root />, rootNode)
 }
 
-writeNavigationTimingMetrics()
+window.addEventListener('load', _event => {
```
hoorayimhelping, line 549:
```suggestion
window.addEventListener('load', () => {
```
generally you use the `_argument` syntax when you have a callback function and you need to use an argument that is later in the list _and_ you don't want to use the earlier argument:

```ts
doSomethingThenCallCallback((_event, metric) => {
  writeMetric(metric)
  // don't care about the first argument `event`, but need to list it
  // to access `metric` in the argument list
})
```

zoesteinkamp: I changed this!

PR #18506, opened 2020-06-15
feat: annotate context with feature flags when handling flux queries

adrian-thurston:
Annotate the context with feature flags when handling flux queries in influxdb.
Taking advantage of this in flux end-to-end tests. Using a custom flagger that
can set overrides based on the test case that is about to be run, allowing us
to enable features in the end-to-end tests.

### DIFF
--- a/cmd/influxd/launcher/launcher.go
+++ b/cmd/influxd/launcher/launcher.go
@@ -850,16 +850,18 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 		Addr: m.httpBindAddress,
 	}
 
-	m.flagger = feature.DefaultFlagger()
-	if len(m.featureFlags) > 0 {
-		f, err := overrideflagger.Make(m.featureFlags, feature.ByKey)
-		if err != nil {
-			m.log.Error("Failed to configure feature flag overrides",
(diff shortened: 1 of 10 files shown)


gavincabbage:
This seems sane @adrian-thurston and avoids the problem of changing flags for _all_ tests, as was my concern with the last approach. Nice one!

I think you could simplify things by avoiding the use of a new flagger implementation. The override flagger should return the default in case an override is not present for a given flag, so simply using an override flagger with an empty override map should be equivalent to using a default flagger. I think you can get rid of `TestFlagger` and just always use an override flagger, in other words.

Let me know if that idea pans out. Besides that idea this looks good to me!

adrian-thurston:
Thank you for having a look @gavincabbage!

I need the flags to change state from one test to the next. Each test is allowed to specify it's own feature flags state. Can I use override flagger for that? I couldn't see how I could alter the flags after passing them to the request handler, via the launcher.

I noticed all the flaggers are passed around by value, in IDPE and influxdb. So I didn't want to go around changing everything to a pointer so I could access it during run time. I thought using a custom flagger with a pointer back to the state would be less disruptive. Am I understanding things right?

PR #18507, opened 2020-06-15
refactor(http): Extract write handler logic for re-use elsewhere

brettbuddin:
This is the first PR in a 3-part story to deliver an API for 1.x writes.

This shuffles the structure of the existing `/api/v2/write` endpoint to allow the `/write` (1.x) endpoint to share the point parse and writing code. Code that was originally placed directly in the main HTTP handler has been moved out into other supporting structures for re-use.

📎 influxdata/idpe#6948


http/api_handler.go
```diff
@@ -208,9 +209,11 @@ func NewAPIHandler(b *APIBackend, opts ...APIHandlerOptFn) *APIHandler {
...
 	h.Mount(prefixWrite, NewWriteHandler(b.Logger, writeBackend,
 		WithMaxBatchSizeBytes(b.MaxBatchSizeBytes),
-		WithParserMaxBytes(b.WriteParserMaxBytes),
-		WithParserMaxLines(b.WriteParserMaxLines),
-		WithParserMaxValues(b.WriteParserMaxValues),
+		WithParserOptions(
+			models.WithParserMaxBytes(b.WriteParserMaxBytes),
+			models.WithParserMaxLines(b.WriteParserMaxLines),
+			models.WithParserMaxValues(b.WriteParserMaxValues),
+		),
```
stuartcarnie, line 216: 👏

kit/io/limited_read_closer.go
```diff
@@ -0,0 +1,48 @@
...
+			err = cerr
+		}
+
+		// only call close once
+		l.close = func() error { return nil }
+	}()
+
+	if l.N < 1 {
+		l.err = ErrReadLimitExceeded
+	}
```
stuartcarnie, line 45: What happens if the source data was exactly `N` bytes? Would this report an error even though it had successfully read all data?

brettbuddin: Oooo great eye. I think we should make that less than 0 instead.

brettbuddin: Nevermind. It looks like they did this intentionally. In the constructor they initialize the `io.LimitReader` with `n + 1`. That's consistent with the `l.N < 1` here. I'm going to leave it alone, but write a few tests to clamp in the behavior.

PR #18508, opened 2020-06-15
fix(dbrp): Change DBRP mapping service to use Buckets resource permissions

brettbuddin:
In order to improve the UX of the 1.x endpoint, we are auto-creating DBRP mappings and Buckets for users if they don't exist. The current DBRP mapping permissions checking has proved problematic, because Bucket write/read tokens generated in the UI only create permissions for Bucket resources. Currently, these tokens don't have permissions to deal with DBRP mappings.

In order to allow for normal Bucket read/write tokens (all buckets and specific buckets) to access this API, I've had to make some minor adjustments to the DBRP mapping service's authorization layer. These changes force the DBRP authorization layer to focus on Buckets resource permissions instead of those of the DBRP resource type.

A few examples to illustration what this means:

- Creating a DBRP mapping will require create permissions for Buckets
- Reading a DBRP mapping will require read access to the Bucket it points to.
- Deleting a DBRP mapping will require write permissions to the Bucket it points
  to.

This raises the question: Should DBRP mappings be publicly visible resources that have permissions of their own? I would argue they shouldn't be. The are an implementation detail of the Bucket resource to provide a mapping for 1.x DB/RP concepts. As such, they should rely on Bucket permissions.

📎 https://github.com/influxdata/influxdb/pull/18507
📎 influxdata/idpe#6948

### DIFF
--- a/authorizer/authorize_find.go
+++ b/authorizer/authorize_find.go
@@ -12,7 +12,7 @@ func AuthorizeFindDBRPs(ctx context.Context, rs []*influxdb.DBRPMappingV2) ([]*i
 	// https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating
 	rrs := rs[:0]
 	for _, r := range rs {
-		_, _, err := AuthorizeRead(ctx, influxdb.DBRPResourceType, r.ID, r.OrganizationID)
+		_, _, err := AuthorizeRead(ctx, influxdb.BucketsResourceType, r.BucketID, r.OrganizationID)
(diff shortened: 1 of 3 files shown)


brettbuddin:
@stuartcarnie Just saw those myself. I know the problem, and will fix them up.

brettbuddin:
@stuartcarnie Got that behavior addressed in the Pipeline tests. Opened an IDPE PR that will need to me merged immediately after this to get those test changes.

PR #18513, opened 2020-06-15
feat: migrate task records to always have an owner id

lyondhill:
Co-authored-by: Alirie Gray <alirie.gray@gmail.com>

### DIFF
--- a/kv/task.go
+++ b/kv/task.go
@@ -1778,7 +1778,7 @@ func taskRunKey(taskID, runID influxdb.ID) ([]byte, error) {
 	return []byte(string(encodedID) + "/" + string(encodedRunID)), nil
 }
 
-func (s *Service) taskOwnerIDUpMigration(ctx context.Context, store Store) error {
+func (s *Service) TaskOwnerIDUpMigration(ctx context.Context, store Store) error {
 	var ownerlessTasks []*influxdb.Task
 	// loop through the tasks and collect a set of tasks that are missing the owner id.
 	err := store.View(ctx, func(tx Tx) error {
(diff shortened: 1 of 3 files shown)


kv/task.go
```diff
@@ -1694,3 +1694,134 @@ func taskRunKey(taskID, runID influxdb.ID) ([]byte, error) {
...
+				return influxdb.ErrInternalTaskServiceError(err)
+			}
+
+			t := kvToInfluxTask(kvTask)
+
+			if !t.OwnerID.Valid() {
+				ownerlessTasks = append(ownerlessTasks, t)
+			}
+		}
+		return c.Err()
```
GeorgeMac, line 1727:
```suggestion
if err := c.Err(); err != nil {
    return err
}

return c.Close()
```

GeorgeMac: Oh I see the defer close up the top. I guess you could do that too.

kv/task.go
```diff
@@ -1694,3 +1694,134 @@ func taskRunKey(taskID, runID influxdb.ID) ([]byte, error) {
...
+
+				for k, v := cur.First(); k != nil; k, v = cur.Next() {
+					m := &influxdb.UserResourceMapping{}
+					if err := json.Unmarshal(v, m); err != nil {
+						return err
+					}
+					if m.ResourceID == t.OrganizationID && m.ResourceType == influxdb.OrgsResourceType && m.UserType == influxdb.Owner {
+						t.OwnerID = m.UserID
+						break
+					}
```
GeorgeMac, line 1798: This is going to be a big scan. Since URMs are prefixed by resource ID. I suggest using forward cursor to do a prefixed search based on org ID.
e.g.
```go
id, err := t.OrganizationID.Encode()
// ...
c, err := b.FowardCursor(id, kv.WithCursorPrefix(id))
// ...
```

kv/task.go
lyondhill: I keep forgetting our close has an error. I will commit your suggestion and remove the defer.

Issue #18243, opened 2020-05-26
Backend: eventing / metrics


hoorayimhelping:
# Background
We need the ability to collect backend for the community template installations and removal.

# Acceptance Criteria
- [ ] Telemetry metrics of the backend of this feature are shipped to tools
- [ ] Dashboards displaying metrics exist on tools

PR #18515, opened 2020-06-15
feat(pkger): extend pkgs with source identifiers

jsteenb2:
a good chunk of this PR is jiggling some bits to remove the high lvl Pkg from the svc interface.

references: #18243 

<details><summary>TODOs</summary>

- [x] add tests around different source types in Parser
- [x] add assertions to pkger_test.go
- [x] extend stack with sources
- [x] ensure API is backwards compatible (when OSS beta13 is released, we can let the axe split)

</details>

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/pkger/http_remote_service.go
+++ b/pkger/http_remote_service.go
@@ -68,7 +68,7 @@ func (s *HTTPRemoteService) ExportStack(ctx context.Context, orgID, stackID infl
 		Get(RoutePrefix, "stacks", stackID.String(), "export").
 		QueryParams([2]string{"orgID", orgID.String()}).
 		Decode(func(resp *http.Response) error {
-			decodedPkg, err := Parse(EncodingJSON, FromReader(resp.Body))
+			decodedPkg, err := Parse(EncodingJSON, FromReader(resp.Body, ""))
 			if err != nil {
 				return err
 			}
@@ -139,7 +139,7 @@ func (s *HTTPRemoteService) CreatePkg(ctx context.Context, setters ...CreatePkgS
 	err := s.Client.
 		PostJSON(reqBody, RoutePrefix).
 		Decode(func(resp *http.Response) error {
-			pkg, err := Parse(EncodingJSON, FromReader(resp.Body))
+			pkg, err := Parse(EncodingJSON, FromReader(resp.Body, "export"))
 			newPkg = pkg
 			return err
 		}).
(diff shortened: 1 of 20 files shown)


pkger/http_remote_service.go
```diff
@@ -157,35 +157,37 @@ func (s *HTTPRemoteService) CreatePkg(ctx context.Context, setters ...CreatePkgS
 // DryRun provides a dry run of the pkg application. The pkg will be marked verified
 // for later calls to Apply. This func will be run on an Apply if it has not been run
 // already.
-func (s *HTTPRemoteService) DryRun(ctx context.Context, orgID, userID influxdb.ID, pkg *Pkg, opts ...ApplyOptFn) (PkgImpactSummary, error) {
-	return s.apply(ctx, orgID, pkg, true, opts...)
+func (s *HTTPRemoteService) DryRun(ctx context.Context, orgID, userID influxdb.ID, opts ...ApplyOptFn) (PkgImpactSummary, error) {
+	return s.apply(ctx, orgID, true, opts...)
```
hoorayimhelping, line 161: not a blocker for this, but a design consideration to have a think about:

boolean arguments trigger me hard because they provide very little context to what's happening. i assume `true` and `false` passed in to `apply` have to do with dry runs, but that is a wild ass guess that required looking around at the context of these calls to even have an idea. point is, boolean arguments are a kind of hard stop on comprehension and require digging into definitions to understand.

i think it's usually worth considering how the intent of this `true` vs `false` value might be captured in a way that gives context to the reader. making it an enum is one way i can think of that doesn't resort to stringly typed stuff.

pkger/http_server.go
```diff
@@ -584,6 +625,21 @@ func pkgEncoding(contentType string) Encoding {
 	}
 }
 
+func convertEncoding(ct, rawURL string) Encoding {
```
hoorayimhelping, line 628: sorry, it's hard for me to tell but is this unit tested? it seems like a prime candidate for unit tests hitting all the branches in the switch statement

pkger/http_remote_service.go
jsteenb2: I hear you, if it were exported, I'd have addressed it. However, it's internal, and don't see it necessary here.

pkger/http_server.go
jsteenb2: not explicitly, its an internal implementation detail. Is tested in other tests that rely on this. It is purposeful that there is very limited white box testing in pkger.

PR #18519, opened 2020-06-15
fix: remove urm create from task create request.

lyondhill:
This will greatly reduce the amount of urms created in the system. To make this change to the system we need to also remove our direct reliance on the urm's in tasks.
Remove the create and delete portion of the task actions that are creating and deleting urms
Remove reliance on urm's when we are doing FindTaskByUser and instead rely on the user filter matching the task.OwnerID

One test had to be changed because the test was explicitly hacking the task to remove the owner ID and then trying to successfully look up the task by ownerID

### DIFF
--- a/kv/task.go
+++ b/kv/task.go
@@ -389,10 +389,10 @@ func (s *Service) findTasksByUserUrmFree(ctx context.Context, tx Tx, filter infl
 		return nil, 0, influxdb.ErrUnexpectedTaskBucketErr(err)
 	}
 
-	// free cursor resources
-	defer c.Close()
-
-	ps, _ := s.maxPermissions(ctx, tx, *filter.User)
+	ps, err := s.maxPermissions(ctx, tx, *filter.User)
+	if err != nil {
+		return nil, 0, err
+	}
 
 	matchFn := newTaskMatchFn(filter, nil)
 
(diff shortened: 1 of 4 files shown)


kv/task.go
```diff
@@ -278,78 +278,68 @@ func (s *Service) findTasks(ctx context.Context, tx Tx, filter influxdb.TaskFilt
...
 
 			if len(ts) >= filter.Limit {
 				break
 			}
 		}
+	}
 
+	if err := c.Err(); err != nil {
+		return nil, 0, err
 	}
```
GeorgeMac, line 340: You'll want to check the error on `cursor.Close()` here

kv/task.go
```diff
@@ -278,78 +278,68 @@ func (s *Service) findTasks(ctx context.Context, tx Tx, filter influxdb.TaskFilt
...
-			if filter.After == nil || afterSeen {
-				return true
-			}
+	// free cursor resources
+	defer c.Close()
 
-			if task.ID == *filter.After {
-				afterSeen = true
-			}
+	ps, _ := s.maxPermissions(ctx, tx, *filter.User)
```
GeorgeMac, line 310: Do we not care about the error here? 🤔

lyondhill:
> Makes me wonder if we do want to flag this change 🤔 What do you think?

Yah the change ended up being larger than I thought. A feature flag would get the create actions creating urm's. I think if we run into problems with this change it will be with the find action or as part of task execute, both of which the feature flag wouldn't be able to help.

kv/task.go
lyondhill: I added error checking on the return.

PR #18521, opened 2020-06-15
fix: don't break on auto detected delimiters

drdelambre:
this is the fix for influxdata/idpe#7553, influxdata/idpe#5420 for the flux parser that sits behind the `fluxParser` feature flag within influxdb.
this had something to do with papaparse choosing its own delimiters for tearing apart a csv. Hard coding them made the whole thing fall away.

Also set up a pattern where we can capture offending flux results from network requests as a CSV and load them straight into the testing framework. It was super cool to not understand what was wrong with the code and still write a test to capture it before trying to fix it. TDD FTW.


ui/src/shared/utils/fromFlux.test.ts
```diff
@@ -230,4 +233,32 @@ there",5
     expect(flux.fluxGroupKeyUnion).toEqual([])
     expect((fluxLegacy.table as any).columns['_value'].data).toEqual([28])
   })
+
+  test('it crushes whatever this bug is', () => {
+    const file = path.join(__dirname, './', 'fromFlux.newline.csv')
```
drdelambre, line 238: we have to use `fs` here, or else typescript starts fumbling over itself to validate the values within a string

ui/src/shared/utils/fromFlux.ts
```diff
@@ -196,7 +195,11 @@ export default function fromFlux(csv: string): ParsedFlux {
       csv.substring(
         chunks[currentChunkIndex].start,
         chunks[currentChunkIndex].stop
-      )
+      ),
+      {
```
drdelambre, line 199: this is the fix

ui/src/shared/utils/fromFlux.ts
```diff
@@ -295,8 +297,8 @@ export default function fromFlux(csv: string): ParsedFlux {
           runningTotal + currentLineIndex - headerLocation - 1
         ] = parseValue(
           parsed[currentLineIndex][currentColumnIndex] ||
-            output[columnKey].default,
-          output[columnKey].type
+            annotations['#default'][columnName],
```
drdelambre, line 300: this change removes an error where results were all grouped by giraffe because they used the same name for the yield

drdelambre:
i honestly dont know what the bug is, or how it was caused, just that adding those things made it work. Those seem like the same defaults we've been relying on for all of the other queries, so why did being specific make them work?

all this to say, i dont know what to change it to?

PR #18526, opened 2020-06-16
chore(flagged): add active template to redux

### LINKED ISSUE #18525: Landing Page: Add Community Templates to redux

hoorayimhelping:
Closes #18525 

Adds community template structure to redux. This is all hidden behind the flag `communityTemplates`

Pasting 
```
https://github.com/influxdata/community-templates/tree/master/csgo
```

into the file field results in the template being added to redux. Next part of the iteration will take the data this puts into redux and populate the modal with it.

Screen Shot 2020-06-15 at 6 18 01 PM

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/templates/containers/CommunityTemplatesIndex.tsx
+++ b/ui/src/templates/containers/CommunityTemplatesIndex.tsx
@@ -20,13 +20,17 @@ import {
 import SettingsTabbedPage from 'src/settings/components/SettingsTabbedPage'
 import SettingsHeader from 'src/settings/components/SettingsHeader'
 
+import {setActiveCommunityTemplate} from 'src/templates/actions/creators'
+import {getOrg} from 'src/organizations/selectors'
+
 // Utils
 import {pageTitleSuffixer} from 'src/shared/utils/pageTitles'
-import {getOrg} from 'src/organizations/selectors'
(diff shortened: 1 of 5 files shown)


ui/src/templates/containers/CommunityTemplatesIndex.tsx
```diff
@@ -123,6 +140,30 @@ class UnconnectedTemplatesIndex extends Component<Props> {
...
+      },
+    }
+    try {
+      const resp = await postPackagesApply(params)
+      if (resp.status < 300) {
+        const template = resp.data.summary
+        this.props.setActiveCommunityTemplate(template)
+        return template
+      }
+      console.error('Failed to apply packages: ', resp)
```
drdelambre, line 161: i've seen the pattern elsewheres in our codebase to flip the conditional logic and throw an error so that we can address how we handle an error (maybe changing from console errors to honey badger alerts) in the same place (the catch)

ie:
```
if (resp.status >= 300) {
  throw new Error(..same mesage..)
}
// you got it
```

hoorayimhelping: sure, works for me

PR #18534, opened 2020-06-16
feat(query): planner rule to push down window and bare first() and la…

yzhang1991:


flags.yml
```diff
@@ -55,9 +55,33 @@
   default: false
   contact: Query Team
 
-- name: Push Down Window Aggregate Rest
-  description: Enable non-Count, non-Sum variants of PushDownWindowAggregateRule and PushDownWindowAggregateRule (stage 2)
-  key: pushDownWindowAggregateRest
+- name: Push Down Window Aggregate Min
```
jlapacik, line 58: Can you add these flags to ConfigCat before merging. That way we won't spam the logs when this code gets deployed.

query/stdlib/influxdata/influxdb/rules_test.go
```diff
@@ -1995,25 +1957,9 @@ func TestPushDownBareAggregateRule(t *testing.T) {
...
-	sumProcedureSpec := func() *universe.SumProcedureSpec {
-		return &universe.SumProcedureSpec{
-			AggregateConfig: execute.AggregateConfig{Columns: []string{"_value"}},
-		}
-	}
-
 	testcases := []plantest.RuleTestCase{
 		{
-			// successful push down
+			// successful push down count()
```
jlapacik, line 1962: Can you replace these types of comment with ones of the form `ReadRange -> first => ReadWindowAggregate`?

PR #18535, opened 2020-06-16
feat: allow restricting TLS version and ciphers

### LINKED ISSUE #15437: Add ability to specify minimum and maximum acceptable TLS version
In https://github.com/influxdata/influxdb/issues/15436 we will add TLS support to InfluxDB 2.0.

There are some features we have in InfluxDB 1.7.x that would be nice to carry over to the 2.x line. Specifically:

```
[tls]
  # Determines the available set of cipher suites. See https://golang.org/pkg/crypto/tls/#pkg-constants
  # for a list of available ciphers, which depends on the version of Go (use the query
  # SHOW DIAGNOSTICS to see the version of Go used to build InfluxDB). If not specified, uses
  # the default settings from Go's crypto/tls package.
  # ciphers = [
  #   "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305",
  #   "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
  # ]

  # Minimum version of the tls protocol that will be negotiated. If not specified, uses the
  # default settings from Go's crypto/tls package.
  # min-version = "tls1.2"

  # Maximum version of the tls protocol that will be negotiated. If not specified, uses the
  # default settings from Go's crypto/tls package.
  # max-version = "tls1.2"
```

That is: 

 - the ability to specify which TLS ciphers that you are willing to support;
 - a minimum version of the TLS protocol that a client and the server may negotiate; and
 - a maximum version of the TLS protocol that a client and the server may negotiate.

rawkode:
Closes #15437 

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influxd/launcher/launcher.go
+++ b/cmd/influxd/launcher/launcher.go
@@ -270,6 +270,18 @@ func buildLauncherCommand(l *Launcher, cmd *cobra.Command) {
 			Desc:    "TLS key for HTTPs",
 		},
 		{
+			DestP:   &l.httpTLSMinVersion,
+			Flag:    "tls-min-version",
+			Default: "1.2",
+			Desc:    "Minimum accepted TLS version",
+		},
+		{
+			DestP:   &l.httpTLSStrictCiphers,
+			Flag:    "tls-strict-ciphers",
+			Default: false,
+			Desc:    "Restrict accept ciphers to: ECDHE_RSA_WITH_AES_256_GCM_SHA384, ECDHE_RSA_WITH_AES_256_CBC_SHA, RSA_WITH_AES_256_GCM_SHA384, RSA_WITH_AES_256_CBC_SHA",
+		},
+		{
 			DestP:   &l.noTasks,
 			Flag:    "no-tasks",
 			Default: false,
@@ -354,10 +366,12 @@ type Launcher struct {
 
 	queryController *control.Controller
 
-	httpPort    int
-	httpServer  *nethttp.Server
-	httpTLSCert string
-	httpTLSKey  string
+	httpPort             int
+	httpServer           *nethttp.Server
+	httpTLSCert          string
+	httpTLSKey           string
+	httpTLSMinVersion    string
+	httpTLSStrictCiphers bool
 
 	natsServer *nats.Server
 	natsPort   int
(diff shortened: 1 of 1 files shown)


rawkode:
I know there are no tests, but I feel this would be really painful to test.

I have ran it locally, with strict ciphers and min set to 1.2:

```
2020/06/16 16:12:30 http: TLS handshake error from [::1]:35270: tls: client offered only unsupported versions: [302 301]
2020/06/16 16:12:38 http: TLS handshake error from [::1]:35278: tls: client offered only unsupported versions: [303 302 301]
```

and this was my TLS cert rejected for using a weak cipher:

```
2020-06-16T15:13:49.071653Z	error	Failed https service	{"log_id": "0NRQwjdl000", "error": "http2: TLSConfig.CipherSuites is missing an HTTP/2-required AES_128_GCM_SHA256 cipher (need at least one of TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 or TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256)."}
```

cmd/influxd/launcher/launcher.go
```diff
@@ -269,6 +269,18 @@ func buildLauncherCommand(l *Launcher, cmd *cobra.Command) {
...
+			DestP:   &l.httpTLSMinVersion,
+			Flag:    "tls-min-version",
+			Default: "1.2",
+			Desc:    "Minimum accepted TLS version",
+		},
+		{
+			DestP:   &l.httpTLSStrictCiphers,
+			Flag:    "tls-strict-ciphers",
+			Default: false,
+			Desc:    "Restrict accept ciphers to: ECDHE_RSA_WITH_AES_256_GCM_SHA384, ECDHE_RSA_WITH_AES_256_CBC_SHA, RSA_WITH_AES_256_GCM_SHA384, RSA_WITH_AES_256_CBC_SHA",
```
dnsmichi, line 282: One thing we did to avoid the boolean behaviour being changed over time: Allow the user to define the cipher list as a string, and set a strong default ourselves. AFAIK this is the same in Apache/Nginx.

This may need additional cipher verification but also allows more fine grained control for enterprise environments.

rawkode: I did consider that, but was keen for InfluxDB to provide opinionated and strong defaults; that hopefully change with the times.

cmd/influxd/launcher/launcher.go
```diff
@@ -1078,7 +1092,43 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 		}
 		transport = "https"
 
-		m.httpServer.TLSConfig = &tls.Config{}
+		// Sensible default
+		var tlsMinVersion uint16 = tls.VersionTLS12
```
e-dard, line 1096: What was this the exact previous default?

rawkode: 1.1 - we can use that as a default if preferred, but 1.2 has been around since 2008

e-dard: I'm happy moving default up to 1.2; perhaps this will need documenting somewhere

cmd/influxd/launcher/launcher.go
dnsmichi: Makes sense. And you possibly don't have to deal with strange openssl C api bugs on ciphers (not) loaded on rhel7, as I had fun with in icinga2. That bug caused all our default secure ciphers to be useless and no handshakes performed then.

cmd/influxd/launcher/launcher.go
dnsmichi: I'd also opt for 1.2 as the safe minimum. If you can control it, and don't need to support old legacy clients, forbid to enable 1.1 or below entirely :)

PR #18540, opened 2020-06-16
refactor: windowed read operation should include aggregate name

jlapacik:


yzhang1991:
I am really happy about this change, it adds a lot more clarity to the system.

PR #18557, opened 2020-06-17
fix(ui): add removal of hidden newlines to the fromFlux parser

TCL735:
Closes #18521


ui/src/shared/utils/fromFlux.ts
```diff
@@ -29,13 +29,16 @@ const TO_COLUMN_TYPE: TypeTranslationMap = {
 type ParsedReturnTypes = string | undefined | null | number | boolean
 
 function parseValue(
-  value: string | undefined,
+  value: string,
   columnType: InternalTypes
 ): ParsedReturnTypes {
-  if (value === undefined) {
-    return undefined
+  if (!value) {
```
ebb-tide, line 35: wait. What if value is zero?

PR #18558, opened 2020-06-17
fix(swagger): update swagger spec with correct list output for stacks

jsteenb2:
- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -4720,54 +4720,57 @@ paths:
           content:
             application/json:
               schema:
-                type: array
-                items:
-                  type: object
-                  properties:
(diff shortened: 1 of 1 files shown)


http/swagger.yml
```diff
@@ -4720,54 +4720,57 @@ paths:
...
-                      format: date-time
-                      readOnly: true
-                    updatedAt:
-                      type: string
-                      format: date-time
-                      readOnly: true
-                    resources:
+                type: object
+                properties:
+                  stacks:
```
OfTheDelmer, line 4725: This looks great!


check

PR #18564, opened 2020-06-17
fix: include Now when making a flux query request from a proxy request

adrian-thurston:
Fixes https://github.com/influxdata/idpe/issues/7543


adrian-thurston:
@wolffcm In my local kind setup this fixes the issue with Now not getting set right for tasks when the flux compiler is used. I'm not sure why testing didn't catch that. I haven't had a chance to unpack the testing code. I'll do that tomorrow and update the PR.

wolffcm:
Nice find. It looks like the time was getting dropped in gateway before forwarding the request on to queryd. Yes, we definitely need to cover that line with a regression test.

adrian-thurston:
Hey @wolffcm, added a test case covering the compiler aspect of the conversion from proxy request to query request. The other aspects I don't know much about, so didn't include them.

http/query_test.go
```diff
@@ -610,3 +610,59 @@ func Test_decodeProxyQueryRequest(t *testing.T) {
...
+func TestProxyRequestToQueryRequest_Compilers(t *testing.T) {
+	tests := []struct {
+		name   string
+		pr     query.ProxyRequest
+		want   QueryRequest
+	}{
+		{
+			name: "flux compiler copied",
+			pr: query.ProxyRequest{
+				Dialect: &query.NoContentDialect{},
```
wolffcm, line 623: I did not know we had a `NoContentDialect`... I guess it will execute the query and then return an empty body in the response?

http/query_test.go
```diff
@@ -610,3 +610,59 @@ func Test_decodeProxyQueryRequest(t *testing.T) {
...
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+
+			got, err := QueryRequestFromProxyRequest( &tt.pr )
+			if err != nil {
+				t.Error(err)
+			} else if !reflect.DeepEqual(*got, tt.want) {
```
wolffcm, line 663: I like this pattern for unit tests
```
if diff := cmp.Diff(got, want); diff != "" {
    t.Errorf("...%v...", diff)
}
```
since it will give pretty formatted output with `+` and `-` prefixes for fields that changed. I'm not requesting a change, just wanted you to be aware of it.

adrian-thurston: Ah right, I just copy-pasted and didn't think about it. I'll keep new tests in `cmp.Diff` style (but please do remind me again if it comes up again in a few months and I've forgotten).

http/query_test.go
adrian-thurston: I guess too! I was just looking for the the most straight-forward dialect option that would pass, as I wasn't really sure what we should be doing there for testing.

PR #18566, opened 2020-06-17
fix(tsdb): Add refs for file-sourced tag keys

benbjohnson:
This commit adds ref counting for files that we pull tag keys from. Previously, files were only ref counted during the time we extracted tag keys but this commit adds additional ref counting for the life of the `Engine.tagKeysNoPredicate()` function.

### DIFF
--- a/tsdb/tsm1/engine_schema.go
+++ b/tsdb/tsm1/engine_schema.go
@@ -358,6 +358,9 @@ func (e *Engine) tagKeysNoPredicate(ctx context.Context, orgID, bucketID influxd
 	var stats cursors.CursorStats
 	var canceled bool
 
+	var files unrefs
+	defer func() { files.Unref() }()
+
 	e.FileStore.ForEachFile(func(f TSMFile) bool {
 		// Check the context before touching each tsm file
 		select {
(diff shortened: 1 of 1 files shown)


tsdb/tsm1/engine_schema.go
```diff
@@ -358,6 +358,12 @@ func (e *Engine) tagKeysNoPredicate(ctx context.Context, orgID, bucketID influxd
 	var stats cursors.CursorStats
 	var canceled bool
 
+	// NOTE(bbj): Adding padding in case additional files are added between files
+	// instantiation & the ForEachFile(). The 'files' reference needs to be stable
+	// so defer can clean it up at the end.
+	files := make(unrefs, 0, e.FileStore.Count()+10)
+	defer files.Unref()
```
stuartcarnie, line 365: You can capture `files` in a closure, it won't matter if it grows:

https://play.golang.org/p/xYwL9_OT07S

benbjohnson: Ah, I like that better. 👍

benbjohnson:
alchemy please merge

benbjohnson:
I checked all the other `FileStore.ForEachFile()` uses and they all appear to be safely done so no other updates needed.

PR #18569, opened 2020-06-17
chore: remove draft: true in order to publish the container image

rawkode:
- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -45,7 +45,7 @@ archives:
     format_overrides:
       - goos: windows
         format: zip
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -82,5 +82,5 @@ dockers:
 release:
   name_template: "v{{.Version}}"
   prerelease: auto # when we go to GA remove this line
-  draft: true # when we go to GA remove this line
+  draft: false # This needs to be false, or the container image will not be published
```
BondAnthony, line 85: This should result in a pre-release candidate since `prerelease: auto` is set and the tag has beta...?

rawkode: Yes, is that a problem?

BondAnthony: No problem, I think your fixing the leftover configurations when v2 wasn't on master.

PR #18570, opened 2020-06-17
refactor: migrator and introduce Store.(Create|Delete)Bucket

GeorgeMac:
Apologies for this big one 🙏 

TL;DR

1. `kx.Tx.Bucket()` can no longer be used to create bucket (if not exists).
Instead see `kv.SchemaStore.CreateBucket`.
2. `*kv.Service.Initialize()` no longer exists.
`kv.SchemaStore` should be ran with `kv/migration/all.Up(ctx, logger, store)` before it is consumed by a service type.
3. All bucket creation and deletion should be done via a migration.
4. We have a new cli tool `go run internal/cmd/kvmigrate/main.go create <add some bucket foo>` to aid in correctly configuring migrations.

## Description

This is a refactor which changes primarily the migrator. However, it also changes the pattern for initializing the `kv` service / store.

## The new `kv.SchemaStore`

`kv.Service` no longer owns the migration of a store. And therefore, no longer needs to initialize itself.

Service initialization used to be responsible for creating buckets on the store.
This PR removes that ability from store and it further removes the behaviour of:

```go
store.Update(ctx, func(tx kv.Tx) error {
    bkt, err := tx.Bucket(someBucket)
    // ...
})
```

This used to be a "get or create bucket" action. Now it is purely a lookup and will fail immediately if the bucket is not created. Which is where the following two new methods have been introduced:

- `CreateBucket(ctx, name)`
- `DeleteBucket(ctx, name)`

They are available on store implementations like `inmem` and `bolt`. However, the `kv.Store` interface remains unchanged. Most services which depend on a store, shouldn't care about bucket creation / deletion. This is now purely a `migration` responsibility.
This is why this PR introduces a superset of store called `kv.SchemaStore`.
This is the type handed to an `Up(...)` or a `Down(...)` migration.

This is why this PR is so large. It changes all the tests to ensure that instead of doing:
```go
svc := kv.NewService(...)
if err := svc.Initialize(ctx); err != nil {
    // ...
}
```

We now do:
```go
import "github.com/influxdata/influxdb/kv/migration/all"

// ...

store := (inmem|bolt).NewKVStore()
// apply migrations to store
if err := all.Up(ctx, logger, store); err != nil {
    // ...
}
// pass store to kv service
svc := kv.NewService(store, ...)
```

## `kvmigrate` internal development cli tool

This repo now contains a buildable cli tool (I just run with `go run` at the moment).

It can be found at `go run internal/cmd/kvmigrate/main.go`

This support one sub-command right now `create`.

For example:

```
$ go run internal/cmd/kvmigrate/main.go create add bucket sessionsv2
Creating new migration: ./kv/migration/all/0005_add_bucket_sessionsv2.go
Inserting migration into ./kv/migration/all/all.go
```

This creates a new migration which is automatically added to the list of `all` migrations for the metadata store. You can edit the migration which initially looks like this:

```go
package all

var Migration0005_AddBucketSessionsv2 = &Migration{}
```

The Migration struct type implements `kv/migration.Spec` (It has a MigrationName, Up and Down method). It has some unexported fields `name`, `up` and `down` which all for handy, quick creation of migrations.

Though there are also some handy utility migrations like:
```go
import  "github.com/influxdata/influxdb/kv/migration"

migration.CreateBuckets("add buckets foo bar and baz", []byte("foo"), []byte("bar"), []byte("baz"))
```

As mentioned the cli tool will also make the appropriate entry in to the `all.Migrations` as well:
```diff
diff --git a/kv/migration/all/all.go b/kv/migration/all/all.go
index 0c1f0ac93e..54782d1f1c 100644
--- a/kv/migration/all/all.go
+++ b/kv/migration/all/all.go
@@ -15,5 +15,7 @@ var Migrations = [...]migration.Spec{
        Migration0003_TaskOwnerIDUpMigration,
        // add dbrp buckets
        Migration0004_AddDbrpBuckets,
+       // add bucket sessionsv2
+       Migration0005_AddBucketSessionsv2,
        // {{ do_not_edit . }}
 }
```


GeorgeMac:
Sorry @jademcgough I must've touched something you own 😓

kv/label.go
```diff
@@ -16,20 +16,16 @@ var (
 	labelIndex         = []byte("labelindexv1")
 )
 
-func (s *Service) initializeLabels(ctx context.Context, tx Tx) error {
-	if _, err := tx.Bucket(labelBucket); err != nil {
+func (s *Service) createLabelBuckets(ctx context.Context, creator BucketCreator) error {
```
GeorgeMac, line 19: note to self: remove

kv/onboarding.go
```diff
@@ -16,9 +16,8 @@ var (
 
 var _ influxdb.OnboardingService = (*Service)(nil)
 
-func (s *Service) initializeOnboarding(ctx context.Context, tx Tx) error {
-	_, err := tx.Bucket(onboardingBucket)
-	return err
+func (s *Service) createOnboardingBuckets(ctx context.Context, creator BucketCreator) error {
```
GeorgeMac, line 19: note to self: remove

kv/service.go
```diff
@@ -49,15 +48,13 @@ type Service struct {
...
-	Migrator *Migrator
-
 	urmByUserIndex *Index
 
 	disableAuthorizationsForMaxPermissions func(context.Context) bool
 }
 
 // NewService returns an instance of a Service.
-func NewService(log *zap.Logger, kv Store, configs ...ServiceConfig) *Service {
+func NewService(log *zap.Logger, kv SchemaStore, configs ...ServiceConfig) *Service {
```
GeorgeMac, line 57: note to self: demote to `kv.Store` (schema store only useful for migrations now)

kv/migration/all/0003_task_owners_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+	"github.com/benbjohnson/clock"
+	"github.com/influxdata/influxdb/v2"
+	"github.com/influxdata/influxdb/v2/inmem"
+	"github.com/influxdata/influxdb/v2/kv"
+	"github.com/influxdata/influxdb/v2/kv/migration"
+	"go.uber.org/zap/zaptest"
+)
+
+func Test_(t *testing.T) {
```
GeorgeMac, line 16: note to self: somehow this got chopped off

GeorgeMac:
Sorry @jacobmarble . Changing the shape of the kv store a bit.
I am treading on everyone's toes.

GeorgeMac:
OK most of the tests are passing. I will whittle away on the last bit of the cypress ones.
So I shall open this up for review.

GeorgeMac:
This is an effort towards https://github.com/influxdata/idpe/issues/7646

tenant/duplicate_reads.go
```diff
@@ -48,7 +48,7 @@ type tenantService struct {
 }
 
 // NewReadOnlyStore returns a Store that cannot update the underlying kv.Store.
-func NewReadOnlyStore(store kv.Store) (*Store, error) {
+func NewReadOnlyStore(store kv.Store) *Store {
```
lyondhill, line 51: I think the duplicate read system is no longer in use. You could delete this whole file if you want

GeorgeMac:
Good shout @jsteenb2 
I will do that.

PR #18573, opened 2020-06-17
feat(influx): extend influx stacks cmd with update

### LINKED ISSUE #18548: Add the ability to update stack metadata
__Proposal:__
Add the ability to update stack metadata such as name and description. Something like:

```sh
influx stacks update --stack-id=xxxxx -n stack-name -d 'stack description'
```

__Current behavior:__
Currently you can only add a name and description when manually initializing a stack. There's no way to update them after a stack is created.

__Desired behavior:__
The ability to update stack metadata.

__Use case:__
- Providing a name and description for an auto-initialized stack
- Correcting mistakes in a stack name or description

jsteenb2:
Closes #18548 

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influx/pkg.go
+++ b/cmd/influx/pkg.go
@@ -557,9 +557,9 @@ func (b *cmdPkgBuilder) cmdPkgValidate() *cobra.Command {
 
 func (b *cmdPkgBuilder) cmdStacks() *cobra.Command {
 	cmd := b.newCmdStackList("stacks")
-	cmd.Short = "List stack(s) and associated templates. Sub commands are useful for managing stacks."
+	cmd.Short = "List stack(s) and associated templates. Subcommands manage stacks."
 	cmd.Long = `
-	List stack(s) and associated templates. Sub commands are useful for managing stacks.
+	List stack(s) and associated templates. Subcommands manage stacks.
 
 	Examples:
 		# list all known stacks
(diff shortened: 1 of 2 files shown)


cmd/influx/pkg.go
```diff
@@ -555,16 +555,42 @@ func (b *cmdPkgBuilder) cmdPkgValidate() *cobra.Command {
 	return cmd
 }
 
-func (b *cmdPkgBuilder) cmdStack() *cobra.Command {
+func (b *cmdPkgBuilder) cmdStacks() *cobra.Command {
 	cmd := b.newCmdStackList("stacks")
 	cmd.Short = "List stack(s) and associated templates. Sub commands are useful for managing stacks."
+	cmd.Long = `
+	List stack(s) and associated templates. Sub commands are useful for managing stacks.
```
sanderson, line 562: ~Sub commands are useful for managing stacks.~ -> Subcommands manage stacks.

cmd/influx/pkg.go
```diff
@@ -819,6 +826,83 @@ func (b *cmdPkgBuilder) stackRemoveRunEFn(cmd *cobra.Command, args []string) err
...
+	cmd.Short = "Update a stack"
+	cmd.Long = `
+	The stack update command updates a stack.
+
+	Examples:
+		# Initialize a stack with a name and description
+		influx stack update -n $STACK_NAME -d $STACK_DESCRIPTION
+
+		# Initialize a stack with a name and urls to associate with stack.
+		influx stack update -n $STACK_NAME -u $PATH_TO_TEMPLATE
```
sanderson, line 840: Do these initialize stacks or just update existing stacks? These should also include a stack ID since it's a required parameter.

jsteenb2: ahhh good point, totally forgot that 🤦

PR #18576, opened 2020-06-17
feat: instrument notebooks

drdelambre:
adds a bunch of product instrumentation in there so that we can look at it later. this also brings up an acute need to stream events to the console so that we can talk more naturally to product about event locations without requiring them to code it themselves


drdelambre:
decided to just stream the events to the console and let people tell me how they feel after the fact. hopefully this will be ok in the mean time for letting people identify existing events within the product and gaps where they might want to add events.

`influx.set('streamEvents', true)` gets the ball rolling

ui/src/notebooks/shared/event.ts
```diff
@@ -0,0 +1,30 @@
...
+interface KeyValue {
+  [key: string]: string
+}
+
+export const event = (title: string, meta: KeyValue = {}): void => {
+  const time = toNano(Date.now())
+
+  if (isFlagEnabled('streamEvents')) {
+    /* eslint-disable no-console */
+    console.log(`Event:  [ ${title} ]`)
```
ebb-tide, line 13: why you console loggin?

ui/src/notebooks/shared/event.ts
```diff
@@ -0,0 +1,30 @@
...
+    }
+    /* eslint-enable no-console */
+  }
+
+  reportEvent({
+    timestamp: time,
+    measurement: title,
+    fields: {},
+    tags: {...meta},
+  })
```
ebb-tide, line 29: // src/shared/utils/analytics.ts
fireEvent(title, meta) would also send this to Google Analytics.

drdelambre: but.. in src/cloud/utils right?

PR #18577, opened 2020-06-17
feat(reporting): events to durations

ebb-tide:
I'm going to try reporting durations for some of the easily duration-able events, like how long a function takes to execute.

### DIFF
--- a/ui/src/timeMachine/actions/queries.ts
+++ b/ui/src/timeMachine/actions/queries.ts
@@ -33,8 +33,9 @@ import {
   demoDataError,
 } from 'src/cloud/utils/demoDataErrors'
 import {
-  reportSimpleQueryPerformanceEvent,
+  reportSimpleQueryPerformanceDuration,
   reportQueryPerformanceEvent,
+  toNano,
 } from 'src/cloud/utils/reporting'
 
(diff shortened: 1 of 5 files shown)


ui/src/timeMachine/actions/queries.ts
```diff
@@ -147,15 +148,15 @@ export const executeQueries = (abortController?: AbortController) => async (
...
 
     pendingResults.forEach(({cancel}) => cancel())
-    reportSimpleQueryPerformanceEvent('executeQueries queries start')
 
     pendingResults = queries.map(({text}) => {
       reportQueryPerformanceEvent({
-        timestamp: Date.now() * 1000000,
-        fields: {},
-        tags: {event: 'executeQueries queries', query: text},
+        timestamp: toNano(Date.now()),
```
hoorayimhelping, line 157: like we mentioned IRL, move this inside of `runQuery` so we have all `runQuery` calls being tracked, and then call to `executeQueries querying` will handle this part :+1:

ebb-tide: this turns out way harder than I had expected, so skipping for now

PR #18581, opened 2020-06-17
feat(cache-cell-results): added cell result caching to reduce cell configuration querying on loads

### LINKED ISSUE #18401: Don't reload query when opening a cell
We should not refresh the query when opening a cell. We already have the results so we should just display them and not re-run the query or any dependent queries.

Asalem1:
Closes #18401 

### Problem

Dashboard cells are queried once when they are initially loaded, then again when the cell is opened for configuration

### Solution

Implemented a reducer-based cache to cache dashboard cell queries and their results based on a hashed query index. Now, if a cell is opened for configuration, the action will check to see if a result exists for that hashed query, otherwise the query will execute. Once the dashboard unmounts, the query is cached. 

- [x] Rebased/mergeable


ui/src/data/actions/index.ts
```diff
@@ -0,0 +1,18 @@
+export type Action =
+  | ReturnType<typeof resetCachedQueryResults>
+  | ReturnType<typeof setQueryResultsByQueryID>
+
+export const hashCode = s =>
```
hoorayimhelping, line 5: it's usually a good idea to paste the link you copy/pasted internet code form so people can get some context on it, and in the case of bugs, go right to the source.

ui/src/data/reducers/index.ts
```diff
@@ -0,0 +1,35 @@
...
+      return produce(state, draftState => {
+        const {queryID, files} = action
+        if (queryID && files.length) {
+          draftState.queryResultsByQueryID[queryID] = files
+        }
+      })
+    }
+
+    case 'RESET_CACHED_QUERY_RESULTS': {
+      return initialState
```
ebb-tide, line 30: let's return a brand new object here just in case.

ui/src/store/configureStore.ts
```diff
@@ -73,6 +74,7 @@ export const rootReducer = combineReducers<ReducerState>({
   }),
   currentPage: currentPageReducer,
   currentDashboard: currentDashboardReducer,
+  data: dataReducer,
```
ebb-tide, line 77: maybe call this `queryCache` and `queryCacheReducer`?. data can mean sooo many things... :/

ui/src/data/reducers/index.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+  }
+
+  return state
+}
```
hoorayimhelping, line 35: is this file necessary? why can't the query results be stored in the time machine reducer?

ui/src/views/actions/thunks.ts
```diff
@@ -112,8 +115,50 @@ export const getViewForTimeMachine = (
...
+  dashboardID: string,
+  cellID: string,
+  timeMachineID: TimeMachineID
+) => (dispatch, getState: GetState): Promise<void> => {
+  try {
+    dispatch(getViewForTimeMachine(dashboardID, cellID, timeMachineID))
+    const state = getState()
+    const {view} = getActiveTimeMachine(state)
+    const queries = view.properties.queries.filter(({text}) => !!text.trim())
+    const queryID = get(queries, '[0].text', '')
```
ebb-tide, line 154: The function name `setQueryResultsByQueryID` suggests that you will be hashing first and then passing queryID rather than queryText to the function. I think you should be hashing on this line.

ui/src/shared/components/TimeSeries.tsx
```diff
@@ -344,6 +349,7 @@ const mstp = (state: AppState, props: OwnProps): StateProps => {
 
 const mdtp: DispatchProps = {
   notify: notifyAction,
+  setQueryResults: setQueryResultsByQueryID,
```
hoorayimhelping, line 352: what's the idea behind aliasing this? it's really confusing - there's already a method being exported from another file called `setQueryResults`.  this really confounds the ability to see where `setQueryResults` is used - this represents a false positive of that.

ui/src/views/actions/thunks.ts
```diff
@@ -112,8 +115,50 @@ export const getViewForTimeMachine = (
...
+    const state = getState()
+    const files = state.data.queryResultsByQueryID[hashCode(queryID)]
+    if (files) {
+      dispatch(setQueryResults(RemoteDataState.Done, files, null, null))
+      return
+    }
+    dispatch(executeQueries())
+  } catch (error) {
+    // if the files don't exist in the cache, we want to execute the query
+    console.error(error)
```
hoorayimhelping, line 139: if i'm understanding this correctly, this logs an error on cache misses? seems noisy - if that's the case i think we should just swallow this error and not log it.

ui/src/dashboards/components/DashboardPage.tsx
```diff
@@ -31,10 +36,14 @@ interface OwnProps {
   autoRefresh: AutoRefresh
 }
 
-type Props = OwnProps & StateProps & ManualRefreshProps
+type Props = OwnProps & StateProps & ManualRefreshProps & DispatchProps
 
 @ErrorHandling
 class DashboardPage extends Component<Props> {
+  public componentWillUnmount() {
+    this.props.resetCachedQueryResults()
```
hoorayimhelping, line 44: :+1: thinking about cache invalidation

ui/src/types/stores.ts
```diff
@@ -40,6 +43,7 @@ export interface AppState {
   }
   currentPage: CurrentPage
   currentDashboard: CurrentDashboardState
+  data: DataState
```
hoorayimhelping, line 46: i'm going to be blunt here: `data` is a really poor name for a reducer - all reducers deal in data. i'm not sure if this reducer needs to exist, but if it does, what's wrong with calling it the `cache` reducer or the `queryCache` reducer?

ui/src/views/actions/thunks.ts
```diff
@@ -112,8 +115,50 @@ export const getViewForTimeMachine = (
...
+  cellID: string,
+  timeMachineID: TimeMachineID
+) => (dispatch, getState: GetState): Promise<void> => {
+  try {
+    dispatch(getViewForTimeMachine(dashboardID, cellID, timeMachineID))
+    const state = getState()
+    const {view} = getActiveTimeMachine(state)
+    const queries = view.properties.queries.filter(({text}) => !!text.trim())
+    const queryID = get(queries, '[0].text', '')
+    if (queryID) {
```
ebb-tide, line 155: This should be where you lookup to see whether the queryID exists in your cache.

ui/src/data/actions/index.ts
```diff
@@ -0,0 +1,18 @@
...
+  s.split('').reduce((a, b) => ((a << 5) - a + b.charCodeAt(0)) | 0, 0)
+
+export const setQueryResultsByQueryID = (queryID: string, files: string[]) =>
```
ebb-tide, line 8: it's really confusing when the variable is called `queryID` but you are actually passing in the queryText, and you will convert it into the ID inside the function. Could you change the name of this function to setQueryResultsForQuery(queryText: string, files: string[]))

ui/src/data/reducers/index.ts
Asalem1: I had originally set the data in the time machine, but it became a little tricky accessing the data since timeMachine is partitioned into segmented subsets. The thinking with this feature is that we might be also benefit from this caching across the platform in the future, so if the same query is run in data-explorer and a dashboard cell, we'd be able to cache those results and reduce querying

ui/src/shared/components/TimeSeries.tsx
Asalem1: Fair point. I think aliasing helps differentiate between the imported action vs the one that's accessible in the props. Without aliasing, I find that my workflow is:

1. Import the action
2. Try and call the action directly within the component
3. Get confused for a bit, do some debugging, then realize the err of my ways

ui/src/views/actions/thunks.ts
Asalem1: Good point.

ui/src/dashboards/components/DashboardPage.tsx
Asalem1: Can't take any credit for it, it was all @ebb-tide 's idea

ui/src/types/stores.ts
Asalem1: Those are really great names, I'll switch them up

ui/src/shared/components/TimeSeries.tsx
ebb-tide: for `differentiate between the imported action vs the one that's accessible in the props` we sometimes use `setX` vs `onSetX` I had personally set a bad example of this in this file.

ui/src/store/configureStore.ts
Asalem1: totally. Just seeing this. how about dashboardQueryCache? or are we thinking something a little less specific?

ui/src/views/actions/thunks.ts
Asalem1: Definitely. I was thinking that as I was walking through this time around. I'll go ahead and import it in and hash it here

ui/src/data/actions/index.ts
Asalem1: Definitely. Great point

ui/src/store/configureStore.ts
ebb-tide: I like queryCache, I can't tell if we will only use this in dashboards or not.

ui/src/data/reducers/index.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const dataReducer = (
+  state: DataState = initialState,
+  action: Action
+): DataState => {
+  switch (action.type) {
+    case 'SET_QUERY_RESULTS_BY_QUERY': {
+      return produce(state, draftState => {
+        const {queryID, files} = action
+        if (queryID && files.length) {
```
ebb-tide, line 23: I think we should be checking these conditions before firing `SET_QUERY_RESULTS_BY_QUERY` action, and then we can rely on them being here.

Asalem1: Sounds good.

ui/src/timeMachine/actions/queries.ts
```diff
@@ -171,6 +168,10 @@ export const executeQueries = (abortController?: AbortController) => async (
     const duration = window.performance.now() - startTime
 
     let statuses = [[]] as StatusRow[][]
+    const {
+      alertBuilder: {id: checkID},
+    } = state
+
```
Asalem1, line 174: Moved this down so it would be place a little closer to the source of where `checkID` is being used

ui/src/views/actions/thunks.ts
```diff
@@ -112,8 +115,49 @@ export const getViewForTimeMachine = (
...
+  }
+}
+
+export const setQueryResultsByQueryID = (queryID: string) => (
+  dispatch,
+  getState: GetState
+): Promise<void> => {
+  try {
+    const state = getState()
+    const files = state.queryCache.queryResultsByQueryID[hashCode(queryID)]
```
ebb-tide, line 131: You are hashing queryID before sending to this function. so you shouldn't be hashing here.

Asalem1: Hmmm, I'm not sure if i am, but I definitely should be since it's a little misleading at this point

ui/src/shared/components/TimeSeries.tsx
```diff
@@ -271,6 +280,12 @@ class TimeSeries extends Component<Props & WithRouterProps, State> {
 
       this.pendingReload = false
 
+      const queryText = queries[0].text
```
ebb-tide, line 283: What happens if there is more than one query in queries ?

Asalem1: I was wondering about that. I've never seen a situation where more than one query is being sent, but is that a possibility? If so, should we cache each one separately?

ui/src/views/actions/thunks.ts
```diff
@@ -112,8 +115,50 @@ export const getViewForTimeMachine = (
...
+  }
+}
+
+export const setQueryResultsForCell = (
+  dashboardID: string,
+  cellID: string,
+  timeMachineID: TimeMachineID
+) => (dispatch, getState: GetState): Promise<void> => {
+  try {
+    dispatch(getViewForTimeMachine(dashboardID, cellID, timeMachineID))
```
ebb-tide, line 149: I'm worried about race conditions here. We can not await the results of a thunk, we just dispatch it, and so if we are depending on its results as we are... on line 151, we're hoping that getViewForTimeMachine has done its thing already and we can get the view. The solutions is to integrate the `getViewForTimeMachine` and `setQueryResultsForCell` functions in to one.

Asalem1: I hear what you're saying, it's definitely a valid concern, but it seems like it might be overstepping the intent of the function. I'm curious as to how moving the dispatches within the scope of `setQueryResultsForCell` would prevent a race condition?

ui/src/views/actions/thunks.ts
```diff
@@ -112,8 +115,50 @@ export const getViewForTimeMachine = (
...
+    const state = getState()
+    const files = state.queryCache.queryResultsByQueryID[queryID]
+    if (files) {
+      dispatch(setQueryResults(RemoteDataState.Done, files, null, null))
+      return
+    }
+    dispatch(executeQueries())
+  } catch (error) {
+    // if the files don't exist in the cache, we want to execute the query
+    dispatch(executeQueries())
```
ebb-tide, line 139: I don't think there is an error here that is relevant. We are already checking if files don't exist on line 132, and executing queries if they don't. on line 136

Asalem1: that's a really good point that I didn't consider. Should we just remove this from the try/catch?

ebb-tide: I think so, yeah.

ui/src/views/actions/thunks.ts
```diff
@@ -112,8 +115,50 @@ export const getViewForTimeMachine = (
...
+  dashboardID: string,
+  cellID: string,
+  timeMachineID: TimeMachineID
+) => (dispatch, getState: GetState): Promise<void> => {
+  try {
+    dispatch(getViewForTimeMachine(dashboardID, cellID, timeMachineID))
+    const state = getState()
+    const {view} = getActiveTimeMachine(state)
+    const queries = view.properties.queries.filter(({text}) => !!text.trim())
+    const queryText = queries.map(({text}) => text).join('')
```
ebb-tide, line 153: queries is an array of strings after line 152, not an object that contains a text property. you only need to join

ui/src/views/actions/thunks.ts
ebb-tide: in TimeSeries- we setQueryResults, here we should read it back from queryResults. 
so if `state.queryResults.Query.queryResultsByQueryID[queryID]` exists we should use it in the visualization, and not executeQueries. (you might need to look into how to use in the visualization)

Asalem1: I think there might be a misunderstanding stemming from naming two different functions the same thing, since this dispatch should be dispatching the action above:

https://github.com/influxdata/influxdb/pull/18581/files#diff-d32f5a03b5c68c499b895f2222a64da9R125

and not the action to set the data

ui/src/views/actions/thunks.ts
```diff
@@ -112,8 +115,50 @@ export const getViewForTimeMachine = (
...
+}
+
+export const setQueryResultsForCell = (
+  dashboardID: string,
+  cellID: string,
+  timeMachineID: TimeMachineID
+) => (dispatch, getState: GetState): Promise<void> => {
+  try {
+    dispatch(getViewForTimeMachine(dashboardID, cellID, timeMachineID))
+    const state = getState()
```
ebb-tide, line 150:
```suggestion
) => async (dispatch, getState: GetState): Promise<void> => {
  try {
    const state = getState()

    let view = getByID<View>(state, ResourceType.Views, cellID) as QueryView

    if (!view) {
      dispatch(setView(cellID, RemoteDataState.Loading))
      view = (await getViewAJAX(dashboardID, cellID)) as QueryView
    }

    dispatch(
      setActiveTimeMachine(timeMachineID, {
        contextID: dashboardID,
        view,
      })
    )
```

ebb-tide: And remove the `getViewForTimeMachine` function.

ui/src/views/actions/thunks.ts
ebb-tide: yes, that is pretty confusing :) I don't think we need a separate `setQueryResultsByQueryID` function here, just roll what it does into this action.

ui/src/dashboards/components/EditVEO.tsx
```diff
@@ -23,30 +23,30 @@ import {AppState, RemoteDataState, QueryView, TimeMachineID} from 'src/types'
...
   router,
+  setQueryResultsForCell,
   view,
 }) => {
   useEffect(() => {
     // TODO split this up into "loadView" "setActiveTimeMachine"
     // and something to tell the component to pull from the context
     // of the dashboardID
-    getViewForTimeMachine(dashboardID, cellID, 'veo')
+    setQueryResultsForCell(dashboardID, cellID, 'veo')
```
ebb-tide, line 49: ok this is a nit, but this function name is a lot less descriptive to me. can we switch it to something that has a reference to view? maybe `getViewAndResultsForVEO`?

PR #18591, opened 2020-06-18
feat: getting rid of extra buttons

drdelambre:
quick feedback from user testing.. no one uses the buttons


drdelambre:
implemented in another PR

PR #18597, opened 2020-06-18
chore: clean up rum - delete unused files, use official api

hoorayimhelping:
Uses the `cloud` points writer instead of a raw fetch call.

Removes old lineWriter that isn't going to be used.

- [x] Rebased/mergeable
- [x] Tests pass


ui/src/cloud/apis/reporting.ts
```diff
@@ -21,7 +21,7 @@ export interface Points {
 export const reportPoints = (points: Points) => {
   try {
     const url = '/api/v2/app-metrics'
-    fetch(url, {
+    return fetch(url, {
```
hoorayimhelping, line 24: added because it's generally best practice to return promises from functions that return them

PR #18598, opened 2020-06-18
feat: add error metric writing method. call it during honeybadger errors

hoorayimhelping:
Connects IDPE 7432. Doesn't completely finish the task.

* Adds a method that reports a single error metric to tools
  * Calls this method on HoneyBadger errors
* Goal: every time we have an error, we report it, then graph each error over time to establish a baseline error rate. If the rate spikes suddenly, we have good indication that something is up.

* Todo: Start sensibly replacing calls to `console.error`. I say sensibly because in the short time I looked around, it seemed clear that there were areas where this isn't appropriate, but we're logging an error with `console.error`

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/shared/utils/errors.ts
+++ b/ui/src/shared/utils/errors.ts
@@ -3,7 +3,7 @@ import HoneyBadger from 'honeybadger-js'
 import {CLOUD, GIT_SHA} from 'src/shared/constants'
 
 import {getUserFlags} from 'src/shared/utils/featureFlag'
-import {reportPoints, PointTags, PointFields} from 'src/cloud/apis/reporting'
+import {reportEvent, PointTags, PointFields} from 'src/cloud/utils/reporting'
 
 if (CLOUD) {
   HoneyBadger.configure({
@@ -19,10 +19,8 @@ export const reportSingleErrorMetric = (
   measurement = 'ui_error'
 ) => {
   if (CLOUD) {
-    const points = {
-      points: [{measurement, tags, fields: {errorCount: 1, ...fields}}],
-    }
-    reportPoints(points)
+    const point = {measurement, tags, fields: {errorCount: 1, ...fields}}
+    reportEvent(point)
   }
 }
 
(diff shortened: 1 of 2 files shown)


drdelambre:
it's been a bit since those tickets were written, and in classic @desa style, it has no description, but iirc the intent was to be able to bisect the interface in the occasion of an uninstrumented error, not just report on it's existence. this is most of the way to that, it's just missing the ability to add a unique identifier to the error boundary (if it happened on a dashboard cell, or on time machine, or on the variables page, etc).

more context here: https://github.com/influxdata/idpe/issues/5949

hoorayimhelping:
@drdelambre word. i think the phrase "this is most of the way" is spot on - i wanted to hook the wires up and make sure it works before trying to make a system out of it. also this is meant to force the conversation a bit - if we have a basic error reporting that can only do very basic stuff, hopefully that will make what we need more clear?

as for adding context, i was thinking of doing it with tags, maybe something like this:
```ts
reportSingleErrorMetric({
  errorType: 'login_error'
})
// ....
reportSingleErrorMetric({
  errorType: 'timemachine_error'
})
```

what do you think?

drdelambre:
tags are totally awesome for this.

adding just that little bit of extra context allows us to switch the conversation from a passive "our app is broken 15.7% of the time" to a more active "our app is broken 15.7% of the time, and we can fix that by looking [**here**]"

ui/src/shared/utils/errors.ts
```diff
@@ -12,6 +13,19 @@ if (CLOUD) {
...
+export const reportSingleErrorMetric = (
+  tags: PointTags = {},
+  fields: PointFields = {},
+  measurement = 'ui_error'
+) => {
+  if (CLOUD) {
+    const points = {
+      points: [{measurement, tags, fields: {errorCount: 1, ...fields}}],
+    }
+    reportPoints(points)
```
drdelambre, line 25: why didn't you use reportEvent here? it comes with free point pooling

hoorayimhelping: @drdelambre i didn't want to worry about passing in timestamps

drdelambre: @hoorayimhelping i did this so that i didn't have to worry about timestamps
https://github.com/influxdata/influxdb/pull/18576/files#diff-a8c1840ed70957a404f34926551d6ad8R8

PR #18611, opened 2020-06-18
chore: move flagging location to lowest possible point - we safe now

hoorayimhelping:
`reportPoints` is the lowest level function in the reporting system we're setting up - all other calls must go through this function.

This PR moves the `appMetrics` flag / implicit cloud check to this function. Any function calling the reporting framework can now do so without needing to check if its in a cloud context or if `appMetrics` is flagged on.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/cloud/apis/reporting.ts
+++ b/ui/src/cloud/apis/reporting.ts
@@ -1,3 +1,5 @@
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+
 export interface Point {
   measurement: string
   fields: PointFields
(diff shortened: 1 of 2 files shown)


hoorayimhelping:
>idk how this effects GA integration, but it captures a hole if people call the api directly

@drdelambre maybe we can have a GA module called `reportGA` similar to `reportPoints` that has the flag check on it. then we can call `reportGA` in `reportEvent`, similar to how we call `reportPoints`. that'll give us coverage over the low level plumbing api, and we can recommend people use the higher level porcelain api (to use git terms) that abstracts all this lower level stuff away, but still has that coverage.

that'll give us the flexibility to use GA or tools-cluster reporting if we need, at the expense of making some more footguns because there are a couple of separate lower level modules that exist and are usable.

PR #18613, opened 2020-06-18
feat(query/stdlib): add a rule to reorder group and window

### LINKED ISSUE #17895: Rewrite rule to push down window through group (group/window/agg optimization)
From #17893 
> Alternatively we could make a ReadAggregateWindow call for all the windows and regroup and reapply the aggregate on the Flux side. This would require understand how to safely re-aggregate if possible. This seems more complex and not as likely to faster. Therefore this issue is solely focused on the planner rule to push down multiple ReadGroup calls.

This issues is about exploring the possibility of also adding a planner rule that would chose this method of pushdown + grouping. Its not clear if this method would ever be faster than the multiple ReadGroup calls method.

We should explore the space and see if we can find the conditions where it makes sense to push down ReadAggregateWindow and re-group on Flux side. 

DOD:

* We understand when it makes sense to use this alternative approach
* We understand what data we need to make that decision

Once we know the conditions and the data we need to make the planner decision  we can create a cost based plan rule.

jsternberg:
This rule reorders group and window so it will switch from using
`ReadGroup` to using `ReadWindowAggregate` when the intent is to
aggregate a grouped window. It will then add a group node that groups by
the given columns and the start and stop columns and then reperform the
aggregate. This is more performant than performing the group first.

Fixes #17895.

### DIFF
--- a/query/stdlib/influxdata/influxdb/rules.go
+++ b/query/stdlib/influxdata/influxdb/rules.go
@@ -29,6 +29,7 @@ func init() {
 		PushDownWindowAggregateRule{},
 		PushDownWindowAggregateByTimeRule{},
 		PushDownBareAggregateRule{},
+		GroupWindowAggregateTransposeRule{},
 		PushDownGroupAggregateRule{},
 		SwitchFillImplRule{},
 	)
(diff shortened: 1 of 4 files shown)


query/stdlib/influxdata/influxdb/rules_test.go
```diff
@@ -1917,6 +1917,518 @@ func TestPushDownWindowAggregateRule(t *testing.T) {
...
+
+	// ReadRange -> group -> window -> count => ReadWindowAggregate -> group -> sum
+	tests = append(tests, plantest.RuleTestCase{
+		Context: haveCaps,
+		Name:    "SimplePassCount",
+		Rules:   rules,
+		Before:  simplePlan(window1m, "count", countProcedureSpec()),
+		After: simpleResult("count", dur1m, false,
+			plan.CreatePhysicalNode("group", groupResult()),
+			plan.CreatePhysicalNode("count", sumProcedureSpec()),
```
yzhang1991, line 2117: I feel this can look confusing. The node is named `count` but has a `sum` spec. I recommend we also rename the node.

jsternberg: I'll try to. I couldn't figure out how to do it without using `plan.ReplaceNode` and I wasn't having a lot of success with that specific function.

jsternberg: I think I did this successfully.

query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -873,6 +878,107 @@ func (p PushDownBareAggregateRule) Rewrite(ctx context.Context, pn plan.Node) (p
...
+// ReadGroupPhys |> window |> { min, max, count, sum }
+//
+// This pattern will use the PushDownWindowAggregateRule to determine
+// if the ReadWindowAggregatePhys operation is available before it will
+// rewrite the above. This rewrites the above to:
+//
+// ReadWindowAggregatePhys |> group(columns: ["_start", "_stop", ...]) |> { min, max, sum }
+//
+// The count aggregate uses sum to merge the results.
+type PushDownWindowGroupAggregateRule struct{}
```
yzhang1991, line 891: It just occurs to me that this is just changing from pushing down `group` to pushing down `window`.
I think more precisely I would like to call it `TransposeGroupWindowAggregateRule`, because it transposes ReadGroup-window to be ReadWindow-group.

yzhang1991:
Just a naming suggestion, all rest is good.

PR #18618, opened 2020-06-18
fix(dbrp): Wait until we know this bucket is default before writing value

brettbuddin:
DBRPs are currently being persisted with a `default` value of `false` even though they are the first DBRP mapping created for a DB. This fix defers writing the DBRP mapping record until after we've determined that it is indeed the default mapping for an Organization + DB. The index that refers to his default bucket is operating as expected.

### DIFF
--- a/dbrp/service.go
+++ b/dbrp/service.go
@@ -370,17 +370,10 @@ func (s *Service) Create(ctx context.Context, dbrp *influxdb.DBRPMappingV2) erro
 	if err != nil {
 		return ErrInvalidDBRPID
 	}
-	b, err := json.Marshal(dbrp)
-	if err != nil {
-		return ErrInternalService(err)
-	}
 
 	return s.store.Update(ctx, func(tx kv.Tx) error {
 		bucket, err := tx.Bucket(bucket)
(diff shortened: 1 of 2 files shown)


dbrp/service.go
```diff
@@ -394,6 +387,15 @@ func (s *Service) Create(ctx context.Context, dbrp *influxdb.DBRPMappingV2) erro
 		if !defSet {
 			dbrp.Default = true
 		}
+
+		b, err := json.Marshal(dbrp)
+		if err != nil {
+			return ErrInternalService(err)
+		}
```
stuartcarnie, line 394: There is a slight benefit to marshaling outside the `Update`, as we reduce the time we hold a lock. It probably isn't a significant deal, but something to consider. Presumably it is unlikely to fail, but if it does, we save taking a lock in that instance too.

brettbuddin: Good call. I'll revisit.

brettbuddin: Going to go ahead and merge this, because I think the risk of having problems is low here. Especially since we aren't hitting cache with this operation.

PR #18620, opened 2020-06-18
feat: push it all to GA

drdelambre:
sends all of our user events we're instrumenting for new features off to GA

### DIFF
--- a/ui/src/cloud/utils/reporting.ts
+++ b/ui/src/cloud/utils/reporting.ts
@@ -32,7 +32,7 @@ export const reportEvent = ({timestamp, measurement, fields, tags}: Point) => {
     fields = {source: 'ui'}
   }
 
-  fireEvent(measurement,{...reportingTags, ...tags})
+  fireEvent(measurement, {...reportingTags, ...tags})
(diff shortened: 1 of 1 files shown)


alexpaxton:
Realized after my previous comment that @ebb-tide is already a reviewer

drdelambre:
totes gotta make a next pass on this. after @hoorayimhelping mentioned moving the analytics.ts file, i dug some digging and realized that we've implemented the same thing three times. need to pull it all together

hoorayimhelping:
@drdelambre i just had a thought - i think we'll want to change [this error call](https://github.com/influxdata/influxdb/blob/master/ui/src/shared/utils/errors.ts#L16-L25) to call `reportPoints` we don't want to send google analytics errors events, right?

drdelambre:
@hoorayimhelping i can see a user flow ending in an error as being a potentially useful funnel to watch. could even help identify which honey badger errors have a measurable user effect against our current baselines in GA, and trigger a message saying "whoa, our bad, here's $5 off" if we feel so inclined

PR #18641, opened 2020-06-22
feat: push down window first to storage

jlapacik:
- Adds launcher tests for windowed first, bare first, and `aggregateWindow(fn: first)`
- Turns on the storage capability for first selection
- Adds a table iterator for windowed selectors

### DIFF
--- a/storage/flux/reader.go
+++ b/storage/flux/reader.go
@@ -9,6 +9,8 @@ import (
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/execute"
 	"github.com/influxdata/flux/memory"
+	"github.com/influxdata/flux/plan"
+	"github.com/influxdata/flux/stdlib/universe"
 	"github.com/influxdata/flux/values"
 	"github.com/influxdata/influxdb/v2/kit/errors"
 	"github.com/influxdata/influxdb/v2/models"
@@ -616,14 +618,22 @@ func (wai *windowAggregateIterator) Do(f func(flux.Table) error) error {
 	return wai.handleRead(f, rs)
 }
 
+// isSelector returns true if given a procedure kind that represents a selector operator.
+func isSelector(kind plan.ProcedureKind) bool {
+	return kind == universe.FirstKind || kind == universe.LastKind || kind == universe.MinKind || kind == universe.MaxKind
+}
+
 func (wai *windowAggregateIterator) handleRead(f func(flux.Table) error, rs storage.ResultSet) error {
 	windowEvery := wai.spec.WindowEvery
 	createEmpty := wai.spec.CreateEmpty
+
+	selector := len(wai.spec.Aggregates) > 0 && isSelector(wai.spec.Aggregates[0])
+
 	timeColumn := wai.spec.TimeColumn
 	if timeColumn == "" {
 		tableFn := f
 		f = func(table flux.Table) error {
-			return splitWindows(wai.ctx, table, tableFn)
+			return splitWindows(wai.ctx, wai.alloc, table, selector, tableFn)
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


storage/flux/table.gen.go.tmpl
```diff
@@ -299,6 +300,370 @@ func (t *{{.name}}WindowTable) advance() bool {
...
 }
 
+// window selector table
+type {{.name}}WindowSelectorTable struct {
+	{{.name}}Table
+	windowEvery int64
+	timeColumn  string
+}
+
+func new{{.Name}}WindowSelectorTable(
```
jsternberg, line 310: I think it might be possible for us to integrate this code with the existing window code although we might need to pass in a selector flag. I'd like to pair on it, but here's the process for how the current window works.

1. createNextBufferTimes which creates the start and stop columns for each window. In the aggregate window version, one of these is taken to become the time column.
2. merge the values from the buffer to match with the windows that were just calculated.

I am wondering if it's possible for us to merge the selector into this. A lot of the logic is still the same. We still need to compute the start and stop times and we still need to merge the values from the the array cursor. I think we should give it a try.

jlapacik: I'm not opposed to it, but I'd prefer if they were separated. While we still need to compute start and stop times, the time stamps are not interpreted in the same way as they are for aggregates. Combining the logic I think is going to lead to confusion.

storage/flux/table.gen.go
```diff
@@ -305,6 +306,370 @@ func (t *floatWindowTable) advance() bool {
...
+}
+
+func (t *floatWindowSelectorTable) startTimes(arr *cursors.FloatArray) *array.Int64 {
+	start := arrow.NewIntBuilder(t.alloc)
+	start.Reserve(arr.Len())
+
+	for _, v := range arr.Timestamps {
+		// modulo should produce non-negative remainder
+		remainder := v % t.windowEvery
+		if remainder < 0 {
```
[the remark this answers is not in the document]
jlapacik: Exactly, before epoch.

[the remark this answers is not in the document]
yzhang1991: I opened this issue a while ago and maybe your change here is the fix: https://github.com/influxdata/flux/issues/2856

storage/flux/table.gen.go.tmpl
yzhang1991: The array that other aggregates return has the window stop time in their timestamps, but selectors return the selected rows real timestamp, right?

storage/flux/table.gen.go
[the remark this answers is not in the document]
jlapacik: Yes I didn't realize there was an issue for it. Yes I believe this change is the fix for that issue. I've added tests that will fail otherwise. I will check the aggregate code.

storage/flux/table.gen.go.tmpl
jlapacik: That's correct.

storage/flux/table.gen.go
```diff
@@ -305,6 +306,370 @@ func (t *floatWindowTable) advance() bool {
...
+			remainder += t.windowEvery
+		}
+
+		windowStart := v - remainder
+		rangeStart := int64(t.bounds.Start)
```
yzhang1991, line 384: This can be moved outside the loop.

storage/flux/reader.go
```diff
@@ -658,20 +667,60 @@ READ:
...
+				table = newIntegerEmptyWindowSelectorTable(done, typedCur, bnds, windowEvery, timeColumn, key, cols, rs.Tags(), defs, wai.cache, wai.alloc)
+			} else {
+				cols, defs := determineTableColsForSeries(rs.Tags(), flux.TInt)
+				table = newIntegerWindowSelectorTable(done, typedCur, bnds, windowEvery, timeColumn, key, cols, rs.Tags(), defs, wai.cache, wai.alloc)
+			}
 		case cursors.FloatArrayCursor:
-			cols, defs := determineTableColsForWindowAggregate(rs.Tags(), flux.TFloat, hasTimeCol)
-			table = newFloatWindowTable(done, typedCur, bnds, windowEvery, createEmpty, timeColumn, key, cols, rs.Tags(), defs, wai.cache, wai.alloc)
+			if !selector {
+				cols, defs := determineTableColsForWindowAggregate(rs.Tags(), flux.TInt, hasTimeCol)
```
yzhang1991, line 682:
```suggestion
				cols, defs := determineTableColsForWindowAggregate(rs.Tags(), flux.TFloat, hasTimeCol)
```

Please update all other places to have correct types.
Maybe some additional test cases are needed since this was not caught by your tests

jlapacik: Great catch. I only added tests for integer tables. A fix is incoming.

storage/flux/table.go
```diff
@@ -14,6 +14,14 @@ import (
 	"github.com/influxdata/influxdb/v2/models"
 )
 
+func offset(t, window int64) int64 {
```
yzhang1991, line 17:
```suggestion
func offset(t, windowEvery int64) int64 {
```

I feel that's more consistent and clear to other people?

storage/flux/table.gen.go.tmpl
yzhang1991: I kinda believe that it is OK for the selector tables to stand out but if @jsternberg would love to give it a try to merge them I am also interested to see his changes.

jlapacik: We've decided to hold off on merging the implementations at this point. While there is some overlapping logic between the two, because there are a decent amount of edge cases to account for during windowing, a separate selector implementation keeps branching to a minimum which makes it much easier to follow the algorithm.

storage/flux/reader.go
```diff
@@ -658,20 +667,60 @@ READ:
...
+			}
 		case cursors.StringArrayCursor:
-			cols, defs := determineTableColsForWindowAggregate(rs.Tags(), flux.TString, hasTimeCol)
-			table = newStringWindowTable(done, typedCur, bnds, windowEvery, createEmpty, timeColumn, key, cols, rs.Tags(), defs, wai.cache, wai.alloc)
+			if !selector {
+				cols, defs := determineTableColsForWindowAggregate(rs.Tags(), flux.TInt, hasTimeCol)
+				table = newStringWindowTable(done, typedCur, bnds, windowEvery, createEmpty, timeColumn, key, cols, rs.Tags(), defs, wai.cache, wai.alloc)
+			} else if createEmpty {
+				cols, defs := determineTableColsForSeries(rs.Tags(), flux.TInt)
+				table = newStringEmptyWindowSelectorTable(done, typedCur, bnds, windowEvery, timeColumn, key, cols, rs.Tags(), defs, wai.cache, wai.alloc)
```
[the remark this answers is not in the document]
jlapacik: I don't share your same sentiment, but I'm also not dead set on the name.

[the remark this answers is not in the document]
yzhang1991: I mean, they are not technically "empty window", they are not empty, but they just come with artificially created empty windows as an addition.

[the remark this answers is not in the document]
jlapacik: I've added a comment to the definition that will clear up any confusion if it should arise.

storage/flux/window.go
```diff
@@ -55,6 +57,8 @@ func (w *windowTableSplitter) Do(f func(flux.Table) error) error {
 			arrs[j] = getColumnValues(cr, j)
 		}
 
+		values := getColumnValues(cr, valueColIdx)
```
jsternberg, line 60: Can you just retrieve this from the array above?

```go
values := arrs[valueColIdx]
```

storage/flux/window.go
```diff
@@ -77,6 +82,7 @@ func (w *windowTableSplitter) Do(f func(flux.Table) error) error {
 			table := &windowTableRow{
 				buffer: buffer,
 				done:   done,
+				empty:  w.createEmpty && values.IsNull(i),
```
jsternberg, line 85: Instead of doing this, can you move this check to line 76 above so we drop the null values when we're using a selector?

```
if w.selector && values.IsValid(i) {
    for j, arr := range arrs {
        buffer.Values[j] = arrow.Slice(arr, int64(i))
    }
} else {
    for j, arr := range arrs {
        // might want to add a utility method for creating an empty array into the arrow package
        buffer.Values[j] = arrow.NewBuilder(cr.Cols()[j].Type, w.alloc).NewArray()
    }
}
```

storage/flux/window.go
```diff
@@ -16,17 +16,19 @@ import (
 // splitWindows will split a windowTable by creating a new table from each
 // row and modifying the group key to use the start and stop values from
 // that row.
-func splitWindows(ctx context.Context, in flux.Table, f func(t flux.Table) error) error {
+func splitWindows(ctx context.Context, in flux.Table, createEmpty bool, f func(t flux.Table) error) error {
```
jsternberg, line 19: Can you change `createEmpty` to be `selector`? It seems like the conditional is whether the function is a selector or not and not whether create empty was used. If create empty is false, I think aggregates and selectors are the same. The main difference seems to be that aggregates will have empty tables with one row and selectors don't.

storage/flux/table.go
```diff
@@ -14,6 +14,14 @@ import (
 	"github.com/influxdata/influxdb/v2/models"
 )
 
+func offset(t, windowEvery int64) int64 {
```
jsternberg, line 17: godoc comment.

storage/flux/reader.go
```diff
@@ -619,11 +620,20 @@ func (wai *windowAggregateIterator) Do(f func(flux.Table) error) error {
 func (wai *windowAggregateIterator) handleRead(f func(flux.Table) error, rs storage.ResultSet) error {
 	windowEvery := wai.spec.WindowEvery
 	createEmpty := wai.spec.CreateEmpty
+
+	selector := false
+	if len(wai.spec.Aggregates) > 0 {
```
jsternberg, line 625: Can you change this entire conditional to a function?

```go
func isSelector(kind plan.ProcedureKind) bool {
}
```

You can either continue to do the conditional in the same way, make a gigantic conditional, or a switch statement. It will help to declutter this method and clearly communicate that we are looking for a selector.

storage/flux/reader.go
```diff
@@ -658,20 +667,60 @@ READ:
 		hasTimeCol := timeColumn != ""
 		switch typedCur := cur.(type) {
 		case cursors.IntegerArrayCursor:
-			cols, defs := determineTableColsForWindowAggregate(rs.Tags(), flux.TInt, hasTimeCol)
-			table = newIntegerWindowTable(done, typedCur, bnds, windowEvery, createEmpty, timeColumn, key, cols, rs.Tags(), defs, wai.cache, wai.alloc)
+			if !selector {
+				cols, defs := determineTableColsForWindowAggregate(rs.Tags(), flux.TInt, hasTimeCol)
+				table = newIntegerWindowTable(done, typedCur, bnds, windowEvery, createEmpty, timeColumn, key, cols, rs.Tags(), defs, wai.cache, wai.alloc)
+			} else if createEmpty && !hasTimeCol {
```
jsternberg, line 673: Can you document the reason for this? My understanding is that the reason is because selectors drop empty tables when aggregate window is used so create empty only works when aggregate window is not used. It would be good to capture this here in case we change this behavior in the future.

storage/flux/reader.go
```diff
@@ -616,14 +618,21 @@ func (wai *windowAggregateIterator) Do(f func(flux.Table) error) error {
 	return wai.handleRead(f, rs)
 }
 
+func isSelector(kind plan.ProcedureKind) bool {
```
jsternberg, line 621: godoc comment

storage/flux/table_test.go
```diff
@@ -1193,6 +1193,361 @@ func TestStorageReader_ReadWindowAggregate_TruncatedBoundsCreateEmpty(t *testing
...
+						TimeSequenceSpec: gen.TimeSequenceSpec{
+							Count: math.MaxInt32,
+							Delta: 5 * time.Second,
+						},
+						DataType: models.Integer,
+						Values: func(spec gen.TimeSequenceSpec) gen.TimeValuesSequence {
+							return gen.NewTimeIntegerValuesSequence(
+								spec.Count,
+								gen.NewTimestampSequenceFromSpec(spec),
+								gen.NewIntegerArrayValuesSequence([]int64{1, 2}),
```
jsternberg, line 1226: Can you put 4 values in here so that the test case will get more than just `1` as the answer? I'd like it to be `[1, 2, 3, 4]` and then when you do make window table you'll alternate `1` and `3` for the value. I think this will make a more robust test.

storage/flux/window.go
```diff
@@ -68,15 +75,26 @@ func (w *windowTableSplitter) Do(f func(flux.Table) error) error {
 				Columns:  cr.Cols(),
 				Values:   make([]array.Interface, len(cr.Cols())),
 			}
-			for j, arr := range arrs {
-				buffer.Values[j] = arrow.Slice(arr, int64(i), int64(i+1))
+
+			empty := false
```
jsternberg, line 79: Can you change this to:

```
empty := w.selector && values.IsNull(i)
if empty {
    // ...
} else {
    // ....
}
```

Issue #18498, opened 2020-06-12
Remove references to tenant services from kv package


AlirieGray:
This is a list of the work necessary to remove the org, urm, user, password, and bucket code from the kv package now that these services have been re-implemented in the tenant service. I have included files and code links where kv code references these services, as well as any relevant details for taking this code out. This code can be broken out into separate issues and worked on simultaneously.

- [x] dashboards
- [x] auth
     - this code has already been moved to its own package, so this one will be solved once the feature flag is removed for that package
- [x] checks - in progress
    - `FindOrganizationByName` is used in Find and Create functions
     - URMs are used in `FindChecks`
- [x] lookup service 
   - This service references buckets, orgs, users, dashboards, and telegrafs. 
- [x] tasks 
     - requires a migration to put owner IDs on tasks that do not have them: #18503
- [ ] document & document options 
   - Replaced by pkger, will be removed after https://github.com/influxdata/influxdb/issues/18245
- [ ] remove org, user, urm, password, bucket functions from kv service definition

Many of the references to the org service in kv are via a function `FindOrganizationByName` which is in kv/org.go but is not found in the definition of an `influxdb.OrganizationService` nor the tenant service implementation of it. This function may need to be added to the tenant service.

PR #18643, opened 2020-06-22
refactor: move name lookup to auth http

AlirieGray:
Addresses #18498

This PR moves the name lookup function that is used by the Authorization service from the kv lookup service to be a function of the Authorization service itself. This will allow us to remove the lookup service from the kv package when the new Authorization service feature flag is fully turned on, which is one of the steps to fully migrate to the tenant service.

This change does remove the functionality of looking up the name of a Telegraf, Dashboard, and Source. However, it doesn't appear that this code path was actually being used. The `Name` property is optional on the `Resource` definition in the swagger, and so I have left the swagger unchanged. 

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/authorization/http_server.go
+++ b/authorization/http_server.go
@@ -21,6 +21,7 @@ type TenantService interface {
 	FindOrganization(ctx context.Context, filter influxdb.OrganizationFilter) (*influxdb.Organization, error)
 	FindUserByID(ctx context.Context, id influxdb.ID) (*influxdb.User, error)
 	FindUser(ctx context.Context, filter influxdb.UserFilter) (*influxdb.User, error)
+	FindBucketByID(ctx context.Context, id influxdb.ID) (*influxdb.Bucket, error)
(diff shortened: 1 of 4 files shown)


authorization/http_server.go
```diff
@@ -331,6 +330,39 @@ func newPermissionsResponse(ctx context.Context, ps []influxdb.Permission, svc i
...
+		return r.Name, nil
+	case influxdb.UsersResourceType:
+		r, err := h.tenantService.FindUserByID(ctx, id)
+		if err != nil {
+			return "", err
+		}
+		return r.Name, nil
+	}
+
+	return "", nil
```
brettbuddin, line 363: Should we return an error here if something is attempting to map a resource type we don't support inside this structure?

AlirieGray: I think it should be okay since it is an optional field

PR #18649, opened 2020-06-22
chore(query Builder): fix query builder state

### LINKED ISSUE #18120: navigating to DE from bucket selection doesn't overwrite script editor state of DE.
if DE state was previously in "query builder" mode then clicking on bucket name from buckets tab takes user to DE with bucket selected in query builder'

if DE state was previously in "script editor" mode then clicking on bucket name from buckets tab takes user to DE ... but retains the last script written by user. should probably have the same behavior as above instead. 

Kapture 2020-05-15 at 14 22 35

rbose22:
Closes #18120 

Originally, when a user begins writing in the script editor in the explore page and then navigates to the data page and clicks on one of the buckets, the user is taken to the explore page, but the script editor is still visible instead of the query builder. This solution updates the redux state of the explore page, so that when navigating from the data page to the explore page, the query builder is always displayed. 

This change does not affect navigation from any other pages. For example, if the user was using the script editor on the explore page and then navigated to the tasks page, when they return to the explore page, the script editor will still be visible instead of the query builder. 

- [ ] Rebased/mergeable
- [ ] Tests pass


ui/cypress/e2e/explorer.test.ts
```diff
@@ -43,6 +43,36 @@ describe('DataExplorer', () => {
...
+      cy.getByTestID('switch-to-script-editor').click()
+      cy.get('.flux-editor').within(() => {
+        cy.get('.view-lines').type(fluxCode)
+      })
+      cy.contains('Submit').click()
+      cy.getByTestID('nav-item-tasks').click()
+      cy.getByTestID('nav-item-data-explorer').click()
+
+      cy.get('.flux-editor').within(() => {
+        cy.get('.view-lines').contains(fluxCode).should('exist')
```
hoorayimhelping, line 59: you can safely drop the `should('exist')` if it follows a `contains` function.  https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#Default-Assertions


would 
```ts
  cy.contains(fluxCode)
```
work? or does this need to be scoped to `.view-lines`?

if you can just find `fluxCode` anywhere, this can be simplified to

```suggestion
        cy..contains(fluxCode)
```

ui/src/timeMachine/actions/queryBuilder.ts
```diff
@@ -155,7 +159,7 @@ export const loadBuckets = () => async (
   const orgID = getOrg(getState()).id
 
   dispatch(setBuilderBucketsStatus(RemoteDataState.Loading))
-
+
```
hoorayimhelping, line 162: ✂️ 

https://stackoverflow.com/questions/30884131/remove-trailing-spaces-automatically-or-with-a-shortcut

PR #18651, opened 2020-06-22
fix: don't re-fetch variables if their loading status is done

hoorayimhelping:
Connects #18296

Checks if variables have been loaded and skips the hydration process for any variables that have already been hydrated.

**previously** _note the delay between double clicking the Submit/Cancel button and the cancel notification firing_
old

**now** _cancel button fires immediately_*
new

\* I know we're planning on changing it so it doesn't flash, but the flash is actually the indicator that the ui is responsive.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/variables/actions/thunks.ts
+++ b/ui/src/variables/actions/thunks.ts
@@ -126,12 +126,20 @@ export const hydrateVariables = (skipCache?: boolean) => async (
 ) => {
   const state = getState()
   const org = getOrg(state)
-  const vars = getVariablesFromState(state)
+  const vars = getVariablesFromState(state).filter(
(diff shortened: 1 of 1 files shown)


ui/src/variables/actions/thunks.ts
```diff
@@ -126,12 +126,20 @@ export const hydrateVariables = (skipCache?: boolean) => async (
 ) => {
   const state = getState()
   const org = getOrg(state)
-  const vars = getVariablesFromState(state)
+  const vars = getVariablesFromState(state).filter(
```
Asalem1, line 129: This is simply great.

PR #18658, opened 2020-06-23
fix(influx): support days as duration type

### LINKED ISSUE #18639: Influx CLI - allow bucket creation RP in 'd' units (days)
__Proposal:__
Allow for buckets to be created with a retention policy in days "d" units. Note that in 1.x, duration units include days **d** and weeks **w** as well:
https://docs.influxdata.com/influxdb/v1.8/query_language/spec/#durations

__Current behavior:__
```
./influx bucket create -n testbucket -r 10d
Error: invalid argument "10d" for "-r, --retention" flag: time: unknown unit d in duration 10d
```

__Desired behavior:__
I would like for the "d" unit to be recognized when creating a bucket since it's typical to create RP in days, like 7d, 30d etc.

__Alternatives considered:__
Convert days to hours and use 'h' unit instead.

__Use case:__
Since most people create RPs in terms of days, this would eliminate having to convert to hours and be less confusing.

jsteenb2:
Closes #18639 

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 1. [18602](https://github.com/influxdata/influxdb/pull/18602): Fix uint overflow during setup on 32bit systems
 1. [18623](https://github.com/influxdata/influxdb/pull/18623): Drop support for --local flag within influx CLI
 1. [18632](https://github.com/influxdata/influxdb/pull/18632): Prevents undefined queries in cells from erroring out in dashboards
+1. [18658](https://github.com/influxdata/influxdb/pull/18658): Add support for 'd' day time identifier in the CLI for bucket and setup commands
 
 ## v2.0.0-beta.12 [2020-06-12]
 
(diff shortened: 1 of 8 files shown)


sanderson:
@kelsiev FYI 👆

jsteenb2:
@sanderson / @kelseiv I decided to just get days in for now. I am not sure how to integrate the months/years into this with all the special cases involved (short months (feb), leap year, yada yada). WIll need to follow up, if product wants it, to support the month/year duration.

kelseiv:
Thank you, @sanderson @jsteenb2. That was quick! Makes sense. Day seems most important per the reasons in @abalone23 's request. It sounds like week is next in importance given we supported that in 1.x. 
Although years and months are more complex to implement, it seems to make sense to eventually support retention policies using any Flux duration literal, as you suggested Johnny.

jsteenb2:
I forgot about the `week` duration. I put up a PR to add that, is a trivial extension to the existing.

https://github.com/influxdata/influxdb/pull/18659

jsteenb2:
we're good now, week is merged :-)

PR #18664, opened 2020-06-23
feat(storage): Add option to disable WritePoints() validation.

benbjohnson:
This commit adds the `WithWritePointsValidationEnabled()` option to disable validation in `Engine.WritePoints()` as the same validation is performed earlier in the call stack by cloud.

### DIFF
--- a/models/points_test.go
+++ b/models/points_test.go
@@ -3230,3 +3230,11 @@ func BenchmarkParsePointsWithOptions(b *testing.B) {
 		})
 	}
 }
+
+func BenchmarkValidToken(b *testing.B) {
+	token := []byte("Hello世界")
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		models.ValidToken(token)
+	}
+}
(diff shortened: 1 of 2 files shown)


models/points_test.go
```diff
@@ -3230,3 +3230,11 @@ func BenchmarkParsePointsWithOptions(b *testing.B) {
...
 	}
 }
+
+func BenchmarkValidToken(b *testing.B) {
+	token := []byte("Hello世界")
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		models.ValidToken(token)
+	}
+}
```
benbjohnson, line 3240: This benchmark was added to test an alternate implementation of `ValidToken()` which turned out to be slower than the existing implementation (which calls `range` over `string(a)`):

```go
func ValidToken(a []byte) bool {
	if !utf8.Valid(a) {
		return false
	}

	for len(a) > 0 {
		r, sz := utf8.DecodeRune(a)
		if !unicode.IsPrint(r) || r == unicode.ReplacementChar {
			return false
		}
		a = a[sz:]
	}
	return true
}
```

The Go compiler is smart enough not to allocate additional memory for the `string(a)` in the current implementation and it doesn't have the overhead of function calls to `utf8.DecodeRune()`.

PR #18667, opened 2020-06-23
feat(service/httpd): Add user configurable HTTP headers

ayang64:
This PR will allow users to add custom HTTP response headers to influx.  The intention is to allow users to supply headers to comply with internal security policies.

This is achieved by first adding the `http-headers` configuration options.  `http-headers` accepts a list of key value pairs containing headers and their values.

This configuration entry is used by a wrapper function which applies those headers to each response.

- [x] Rebased/mergeable
- [x] Tests pass


services/httpd/handler.go
```diff
@@ -1873,6 +1875,39 @@ func requestID(inner http.Handler) http.Handler {
...
+func (h *Handler) SetHeadersWrapper(f func(http.ResponseWriter, *http.Request)) func(http.ResponseWriter, *http.Request) {
+	if len(h.Config.HTTPHeaders) == 0 {
+		return f
+	}
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		log.Printf("attempting to set headers for request")
+		for _, kvp := range h.Config.HTTPHeaders {
+			if len(kvp) != 2 {
+				// NOTE(ayan): maybe this should be an error.  maybe we should ensure
```
ayang64, line 1892: Will fix this comment -- will use standard case and edit the comment to be more concise.

services/httpd/config.go
```diff
@@ -44,6 +44,7 @@ type Config struct {
 	PprofAuthEnabled        bool           `toml:"pprof-auth-enabled"`
 	DebugPprofEnabled       bool           `toml:"debug-pprof-enabled"`
 	PingAuthEnabled         bool           `toml:"ping-auth-enabled"`
+	HTTPHeaders             [][2]string    `toml:"http-headers"`
```
gunnaraasen, line 47: `[][2]string` feels a bit cumbersome. Maybe a map type would be a better fit here.

ayang64: agreed!

services/httpd/config_test.go
```diff
@@ -1,12 +1,65 @@
...
 	"github.com/influxdata/influxdb/services/httpd"
 )
 
+// TestConfig_ParseHeaders ensures we can properly extract the user supplied
+// headers from the configuration file.
+//
+// This test is a success if we get the data out as we put into the TOML
+// values.
+//
+func TestConfig_ParseHeaders(t *testing.T) {
```
gunnaraasen, line 19: This covers config handling, but I'd also like to see a small [handler test](https://github.com/influxdata/influxdb/blob/master-1.x/services/httpd/handler_test.go) to show the headers are being added to a request (maybe something like this [test](https://github.com/influxdata/influxdb/blob/master-1.x/services/httpd/handler_test.go#L1729)).

dgnorton:
@gunnaraasen this PR look okay to you?

PR #18671, opened 2020-06-23
refactor: move checks service to new package

AlirieGray:
Part of #18498

This change copies @GeorgeMac's checks service. This change should not affect any functionality. This will allow us to remove the checks file from the kv package which is the last step to removing all references to orgs and urms from kv so that we can switch entirely to the tenant service remove the dead code from kv.

Blocking https://github.com/influxdata/influxdb/pull/18551

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/checks/service.go
+++ b/checks/service.go
@@ -1,0 +1,566 @@
+package checks
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/influxdata/influxdb/v2"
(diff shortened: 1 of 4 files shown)


GeorgeMac:
Heads up you will need to wire this into `influxd` probably in `launcher.go` ?

GeorgeMac:
My mistake :D you done it haha

PR #18677, opened 2020-06-23
feat(VEO-timerange-toggle): prevents timerange selections in the VEO from rerunning queries until the VEO is closed

### LINKED ISSUE #18583: Configuring a cell's timerange updates the dashboard timerange & issues new queries
When a user configures a cell and updates the timerange for that specific cell, the timerange is updated on the dashboard and all the cell queries dependent upon a timerange are then re-queried for the new timerange. This functionality is intended, but what we need to do is trigger the update for the queries when the VEO is updated, rather than when the timerange is updated:

cell-configure-timerange-bug

Asalem1:
Closes #18583 

### Problem

The problem that this PR aims to resolve is that toggling the timerange dropdown of the VEO would rightfully update the timerange in the dashboard, and cause the hidden-behind-the-overlay cells to query based on the new time range.

### Solution

Created two ephemeral states to determine whether:

1. The VEO has been opened
2. The timerange in the VEO has been toggled

By setting whether the VEO is open or not, we can then conditionally trigger an action that should disable rerunning queries when the timerange is updated in the VEO. 

VEO-timerange-fix

- [x] Rebased/mergeable

### DIFF
--- a/ui/src/shared/components/TimeSeries.tsx
+++ b/ui/src/shared/components/TimeSeries.tsx
@@ -45,7 +45,7 @@ import {TIME_RANGE_START, TIME_RANGE_STOP} from 'src/variables/constants'
 // Actions & Selectors
 import {notify as notifyAction} from 'src/shared/actions/notifications'
 import {setQueryResultsByQueryID} from 'src/queryCache/actions'
-import {isInVEOMode} from 'src/shared/selectors/app'
+import {hasUpdatedTimeRangeInVEO} from 'src/shared/selectors/app'
 
 // Types
 import {
@@ -304,11 +304,14 @@ class TimeSeries extends Component<Props, State> {
   }
 
   private shouldReload(prevProps: Props) {
-    if (this.props.isInVEOMode) {
+    if (this.props.hasUpdatedTimeRangeInVEO) {
       return false
     }
 
-    if (prevProps.isInVEOMode && !this.props.isInVEOMode) {
+    if (
+      prevProps.hasUpdatedTimeRangeInVEO &&
+      !this.props.hasUpdatedTimeRangeInVEO
+    ) {
       return true
     }
 
@@ -352,7 +355,7 @@ const mstp = (state: AppState, props: OwnProps) => {
   ]
 
   return {
-    isInVEOMode: isInVEOMode(state),
+    hasUpdatedTimeRangeInVEO: hasUpdatedTimeRangeInVEO(state),
     queryLink: state.links.query.self,
     buckets: getAll<Bucket>(state, ResourceType.Buckets),
     variables,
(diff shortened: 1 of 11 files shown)


ui/src/timeMachine/components/Queries.tsx
```diff
@@ -134,11 +147,13 @@ const mstp = (state: AppState) => {
     activeQuery,
     autoRefresh,
     isInCheckOverlay: getIsInCheckOverlay(state),
+    inVEOMode: isInVEOMode(state),
```
ebb-tide, line 150: So... I think this state is the same as checking with the router to see if you are in the edit cell path. before querying you could check if you are in VEO path, and not update if so.

ui/src/shared/components/TimeSeries.tsx
```diff
@@ -314,6 +316,17 @@ class TimeSeries extends Component<Props & WithRouterProps, State> {
   }
 
   private shouldReload(prevProps: Props) {
+    if (this.props.hasUpdatedTimeRangeInVEO) {
+      return false
+    }
+
+    if (
+      prevProps.hasUpdatedTimeRangeInVEO &&
+      !this.props.hasUpdatedTimeRangeInVEO
```
ebb-tide, line 325: Do you think our query caching will actually do this for us? without having to indicate whether user has updated time range in VEO? Let's think it through

Asalem1: SO this condition is used specifically to trigger an update after a timeRange selection has been made, AND after the VEO has been closed. Without this, toggling the timeRange (even with the initial cache) will cause us to requery the dashboard, so this is to make sure that that doesn't happen until the configuration process has been completed

ui/src/timeMachine/components/Queries.tsx
Asalem1: That's a really good point that I hadn't considered

ui/src/shared/components/TimeSeries.tsx
ebb-tide: but we'd prevent the re querying of the dashboard because we can tell from the router that the VEO is displayed. Then when we close the VEO, we can reissue queries, and if time range has not changed the results will be returned from the cache. if the time range has changed we won't find a match in the cache, and will issue a query.

Asalem1: I think the difference is that we still want to reissue the queries if the timerange selected is the same timerange selected. So if a user has `-5m` selected, they scroll through the dropdown, select other options, then settle back on `-5m`, the query technically will not have changed, but the event should be registered and should update the other cells

ebb-tide: you're right.

ebb-tide: I think we might want to update all cells whenever any change has been applied to a cell. or else the time frames will be out of sync.

PR #18700, opened 2020-06-24
feat(loadBuckets): loadBuckets sets buckets after req and removed Buckets fetching from DataExplorerPage

### LINKED ISSUE #18510: Data explorer making multiple calls for /buckets
Clicking the explore tab makes 2 calls to the /buckets api, but it should only make 1.

image

Asalem1:
Closes #18510

### Problem

Navigating to the DataExplorer was making two query requests for buckets

### Solution

Removed Buckets from GetResources wrapper around DataExplorerPage and updated the loadBuckets function to set the loaded buckets once they are returned from the API and normalized.

- [ ] Rebased/mergeable

### DIFF
--- a/ui/src/dataExplorer/components/DataExplorerPage.tsx
+++ b/ui/src/dataExplorer/components/DataExplorerPage.tsx
@@ -25,7 +25,7 @@ const DataExplorerPage: FC = ({children}) => {
   return (
     <Page titleTag={pageTitleSuffixer(['Data Explorer'])}>
       {children}
-      <GetResources resources={[ResourceType.Variables, ResourceType.Buckets]}>
+      <GetResources resources={[ResourceType.Variables]}>
         <Page.Header fullWidth={true}>
           <Page.Title title="Data Explorer" />
           <CloudUpgradeButton />
--- a/ui/src/timeMachine/actions/queryBuilder.ts
+++ b/ui/src/timeMachine/actions/queryBuilder.ts
@@ -1,4 +1,5 @@
 // APIs
+import {normalize} from 'normalizr'
 import {queryBuilderFetcher} from 'src/timeMachine/apis/QueryBuilderFetcher'
 import * as api from 'src/client'
 import {get} from 'lodash'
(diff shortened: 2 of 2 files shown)


ui/src/dataExplorer/components/DataExplorerPage.tsx
```diff
@@ -25,7 +25,7 @@ const DataExplorerPage: FC = ({children}) => {
   return (
     <Page titleTag={pageTitleSuffixer(['Data Explorer'])}>
       {children}
-      <GetResources resources={[ResourceType.Variables, ResourceType.Buckets]}>
+      <GetResources resources={[ResourceType.Variables]}>
```
ebb-tide, line 28: I'm worried that `setBuilderBucketIfExists` function runs in `DataExplorer.tsx` before `loadBuckets` is called, so will not set buckets from query parameters, if buckets are not available yet. There are a couple ways we can fix this.. let's talk about it.

ui/src/timeMachine/actions/queryBuilder.ts
```diff
@@ -148,9 +155,14 @@ export const selectAggregateWindow = (period: string) => (
 }
 
 export const loadBuckets = () => async (
-  dispatch: Dispatch<Action | ReturnType<typeof selectBucket>>,
+  dispatch: Dispatch<Action | ReturnType<typeof selectBucket> | BucketAction>,
   getState: GetState
 ) => {
+  if (
+    getStatus(getState(), ResourceType.Buckets) === RemoteDataState.NotStarted
```
hoorayimhelping, line 162: so this will work, and it's used in places in our app, so it's not a criticism of you using this pattern, just a heads up: this is a bit of an inference - we're inferring that the state is complete based on the remote data state. that's fine when the state is `Not Started`, but all that `Loaded` tells us is that he data has been loaded. It doesn't tell us what data is loaded, and whether it's been reloaded, whether the state is good, etc. i have a feeling this is going to become relevant soon as we start thinking about when to run and re-run queries and requests. no action item here but to think about how we know if we need to re-fetch data, in general (not just buckets)

ui/src/dataExplorer/components/DataExplorerPage.tsx
ebb-tide: taking this back.

PR #18701, opened 2020-06-24
chore: cleanup ui eventing

drdelambre:
we were using 3 different implementations of the same eventing architecture throughout our app, worked that down to one


ui/src/cloud/utils/reporting.ts
```diff
@@ -31,21 +31,14 @@ export const updateReportingContext = (properties: KeyValue) => {
   reportingTags = {...reportingTags, ...properties}
 }
 
-export const reportEvent = ({
-  timestamp = toNano(Date.now()),
-  measurement,
-  fields,
-  tags,
-}: Point) => {
+const pooledEvent = ({timestamp, measurement, fields, tags}: Point) => {
```
ebb-tide, line 34: calling this var measurement but then changing it to event is super confusing. Can we just call it eventName or whatever throughout ?

ui/src/cloud/utils/reporting.ts
```diff
@@ -99,48 +92,58 @@ export const reportEvent = ({
...
+
+  if (isNaN(time)) {
+    time = Date.now()
+  }
+
+  delete meta.time
+
+  if (isFlagEnabled('streamEvents')) {
+    /* eslint-disable no-console */
+    console.log(`Event:  [ ${title} ]`)
```
ebb-tide, line 118: I don't see the benefit of console.logging events.

ui/src/cloud/utils/reporting.ts
hoorayimhelping: dig the rename `pooledEvent` 👍

+1 on the suggestion to rename `measurement`

ui/src/cloud/utils/reporting.ts
```diff
@@ -31,21 +31,14 @@ export const updateReportingContext = (properties: KeyValue) => {
...
   if (isEmpty(fields)) {
     fields = {source: 'ui'}
   }
 
-  fireEvent(measurement, {...reportingTags, ...tags})
-
   reportingPoints.push({
-    measurement,
-    tags: {...reportingTags, ...tags},
+    measurement: 'UI Event', // artifact of sharing the tools cluster
```
hoorayimhelping, line 40: this means that every time we add a new call to `event`, we add a new tag to the measurement `UI Event`. This gives us no option to change the measurement and we just have to add tags any time we want to measure something? 

Why does everything need to be an event keyed off a single measurement? Were we asked not to create too many measurements?  What if we want to write metrics of the form:

```
{
  measurement: 'the thing i am measuring',
  tags: {...relevantTags},
  fields: {count: 1}
}
```
what is wrong with this kind of measuring?

ui/src/cloud/utils/reporting.ts
```diff
@@ -31,21 +31,14 @@ export const updateReportingContext = (properties: KeyValue) => {
...
+    tags: {...reportingTags, ...tags, event: measurement},
     fields,
     timestamp,
```
hoorayimhelping, line 43: does this need to be converted to nanoseconds?

ui/src/cloud/utils/reporting.ts
ebb-tide: Nothing is wrong. But since all metrics are going to same `apps` bucket in tools, I asked that we have a uniform measurement to visually remove the things we are not interested in. @drdelambre I wonder if a better way for us to do this, could be to use the  `source: ui` tag we are placing on all points?

ui/src/cloud/utils/reporting.ts
ebb-tide: nevermind - I didn't understand that it was behind 'streamEvents' feature flag?

ui/src/cloud/utils/reporting.ts
drdelambre: @ebb-tide could we solve both concerns by changing the bucket that the data is written to?

ebb-tide: That would be great!

ui/src/cloud/utils/reporting.ts
ebb-tide: I will do this next !

PR #18708, opened 2020-06-24
chore: fix configcat flagging in community templates

hoorayimhelping:
Watch that first step, it's a doozy!

[this commit](https://github.com/influxdata/influxdb/commit/a51b1c23ee89d993bbc441d4c1c239206f0073a0) fixed a bug which actually allowed feature flagging directly from `index.tsx` to work. That bugfix caused the call to `isFlagEnabled('communityTemplates')` in the Route structure to no longer work, which caused templates not to be rendered.

This ticket fixes that by changing the structure of those calls. It's less clean with a single entry point into Templates flagged branched on the value of the flag. Now there are multiple entry points and multiple checks to flags. Them's the breaks.

to toggle the flag, start your local influxd process with:

```sh
```

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/shared/reducers/flags.ts
+++ b/ui/src/shared/reducers/flags.ts
@@ -28,10 +28,16 @@ export default (state = defaultState, action: Actions): FlagState => {
     case SET_FEATURE_FLAGS:
       // just setting the loading state
       if (!action.payload.flags) {
-        return {
+        const newState = {
           ...state,
           status: action.payload.status,
         }
+
+        if (!state.hasOwnProperty('original')) {
+          newState.original = defaultState.original
+        }
+
+        return newState
       }
       return {
         ...state,
--- a/ui/src/templates/components/CommunityTemplateImportOverlay.tsx
+++ b/ui/src/templates/components/CommunityTemplateImportOverlay.tsx
@@ -9,6 +9,8 @@ import {CommunityTemplateInstallerOverlay} from 'src/templates/components/Commun
 import {createTemplate as createTemplateAction} from 'src/templates/actions/thunks'
 import {notify as notifyAction} from 'src/shared/actions/notifications'
 
+import {FlagMap} from 'src/shared/reducers/flags'
+
 // Types
 import {AppState, Organization, ResourceType} from 'src/types'
 import {ComponentStatus} from '@influxdata/clockface'
@@ -26,6 +28,7 @@ interface DispatchProps {
 }
 
 interface StateProps {
+  flags: FlagMap
   org: Organization
   templateName: string
 }
@@ -42,6 +45,10 @@ class UnconnectedTemplateImportOverlay extends PureComponent<Props> {
   }
 
   public render() {
+    if (!this.props.flags.communityTemplates) {
+      return null
+    }
+
     return (
       <CommunityTemplateInstallerOverlay
         onDismissOverlay={this.onDismiss}
@@ -74,7 +81,11 @@ const mstp = (state: AppState, props: Props): StateProps => {
     props.params.orgID
   )
 
-  return {org, templateName: props.params.templateName}
+  return {
+    org,
+    templateName: props.params.templateName,
+    flags: state.flags.original,
+  }
 }
 
 const mdtp: DispatchProps = {
--- a/ui/src/templates/containers/TemplatesIndex.tsx
+++ b/ui/src/templates/containers/TemplatesIndex.tsx
@@ -10,14 +10,18 @@ import SettingsHeader from 'src/settings/components/SettingsHeader'
 import TemplatesPage from 'src/templates/components/TemplatesPage'
 import GetResources from 'src/resources/components/GetResources'
 
+import {CommunityTemplatesIndex} from 'src/templates/containers/CommunityTemplatesIndex'
+
 // Utils
 import {pageTitleSuffixer} from 'src/shared/utils/pageTitles'
 import {getOrg} from 'src/organizations/selectors'
 
 // Types
 import {AppState, Organization, ResourceType} from 'src/types'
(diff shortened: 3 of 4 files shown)


ui/src/shared/reducers/flags.ts
```diff
@@ -28,10 +28,16 @@ export default (state = defaultState, action: Actions): FlagState => {
     case SET_FEATURE_FLAGS:
       // just setting the loading state
       if (!action.payload.flags) {
-        return {
+        const newState = {
           ...state,
           status: action.payload.status,
         }
+
+        if (!state.hasOwnProperty('original')) {
```
drdelambre, line 36: this clears out old flags if none are returned?

ui/src/templates/components/CommunityTemplateContents.tsx
```diff
@@ -22,6 +22,13 @@ interface StateProps {
 }
 
 class CommunityTemplateContentsUnconnected extends PureComponent<StateProps> {
+  private safelyAccessLength = collection => {
```
drdelambre, line 25: this should be normalized to an empty array at the data layer (either right off the api call, before putting into redux, or when retrieving from redux, see @121watts's normalization work around resources) and not the view layer.

ui/src/templates/components/CommunityTemplateImportOverlay.tsx
```diff
@@ -26,6 +28,7 @@ interface DispatchProps {
 }
 
 interface StateProps {
+  flags: FlagMap
```
drdelambre, line 31: i'm confused by your statements this morning. using connect to trigger a rerender on your component on update was what was broken with feature flags?

ui/src/shared/reducers/flags.ts
hoorayimhelping: @drdelambre it makes sure the [flag state's structure](https://github.com/influxdata/influxdb/blob/master/ui/src/shared/reducers/flags.ts#L20) has the right shape and that `original` isn't `undefined`

it used to look like:

```
{status: 'loading', override: {}}
```
now it looks like:
```
{status: 'loading', override: {}, original: {}}
```

[`getFlags` dispatches `setFlags`](https://github.com/influxdata/influxdb/blob/master/ui/src/shared/actions/flags.ts#L50) with an empty second argument when it starts loading the flags. i noticed this while trying to track down this issue and changed it just to see if it fixed it. it didn't fix it, but i kept it in because otherwise it returns a state object that is missing a key.

ui/src/templates/components/CommunityTemplateImportOverlay.tsx
hoorayimhelping: >using connect to trigger a rerender on your component on update was what was broken with feature flags?

@drdelambre i don't think so. there's either a race condition between the `GetFlags` component and the code running behind the `isFlagEnabled` call, or there's something deeper going on with rendering subchildren inside of `Route` components. 

Essentially for flags that require api access, calling `isFlagEnabled` at the route layer in `index.tsx` hasn't worked for me for any flag. you can verify this on latest master, toggle community flags on by running influxd with the flag enabled:

```sh
go run ./cmd/influxd --assets-path=ui/build --feature-flags=communityTemplates=true```

make sure your local storage is clear just in case. see if you can get the flagged version working. if it is working, flip the flag to `false` when running influxd and see if you can get it toggling off.

File: ui/src/templates/components/CommunityTemplateContents.tsx (hunk 2 above)
Comment (hoorayimhelping) [reply ui/src/templates/components/CommunityTemplateContents.tsx:25]: @drdelambre i agree that what this code is doing isn't idiomatic or within the confines of our system, but that was a deliberate decision - we're still pretty working out what template data we're going to use or what shape it needs to be in or what parts of it is even useful to users right now. i don't think it makes sense to normalize until way later in this process. i think all that will do right now is slow down iteration and testing of this. this is still an mvp that we might still throw away.

File: ui/src/templates/containers/TemplatesIndex.tsx [hunk 4]
```diff
@@ -26,7 +30,10 @@ type Props = WithRouterProps & StateProps
 @ErrorHandling
 class TemplatesIndex extends Component<Props> {
   public render() {
-    const {org, children} = this.props
+    const {org, children, flags} = this.props
+    if (flags.communityTemplates) {
+      return <CommunityTemplatesIndex>{children}</CommunityTemplatesIndex>
+    }
```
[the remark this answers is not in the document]
hoorayimhelping: `FeatureFlag` calls out to `isFlagEnabled` which references flags in local storage, flags overridden in the console, and flags from configCat, which are stored in redux. I'm not sure what's going on with the `isFlagEnabled`, but I was using it before, and it wasn't working properly - remember how we had to hack up some code to get the feature to show? I spent a day and a half messing with it, trying to get it working and eventually gave up and tried to get something working so we can show people your changes (which we can't do in prod or acceptance until we fix this). I'll figure out how to do this properly later, after it gets unblocked.

ui/src/templates/components/CommunityTemplateContents.tsx
121watts: it would be super easy to just add this to the data layer.  also, easy to change if the shapes change / we need to rip it out.  also, if it _does_ stay it will be easier to extend and we wont have to back in and normalize after the fact.  if it's not normalized now, it probably wont get normalized.

hoorayimhelping: >if it's not normalized now, it probably wont get normalized.

based on what? if i needs to be normalized, i can normalize it before we officially launch this feature.

this probably doesn't even belong in redux state. this is transitional data and state - it's something that exists as a confirmation before we install the resources the template describes, which are what the users actually care about.

PR #18731, opened 2020-06-25
revert(filter_changes): adding the filter function was causing nested…

Asalem1:
Reverts changes previously made to the way variables were being hydrated. Needs a test


drdelambre:
more context around what was reverted: https://github.com/influxdata/influxdb/pull/18651

Asalem1:
Here's the follow-up:

#18732

PR #18735, opened 2020-06-25
fix(storage/flux): the count aggregate will fill in zero values instead of null

### LINKED ISSUE #18702: Aggregate window push down for count doesn't produce zero for empty windows
__Steps to reproduce:__

Enable the aggregate window count push down for count using the feature flag and then run a query like this:

```
from(bucket: ...)
    |> range(...)
    |> filter(...)
    |> aggregateWindow(every: 1m, fn: count)
```

Ensure that there is at least one empty window in the data.

__Expected behavior:__
The empty window has a zero value.

__Actual behavior:__
The empty window has a null value.

jsternberg:
Fixes #18702.

### DIFF
--- a/cmd/influxd/launcher/query_test.go
+++ b/cmd/influxd/launcher/query_test.go
@@ -978,6 +978,38 @@ from(bucket: v.bucket)
 `,
 		},
 		{
+			name: "count with nulls",
+			data: []string{
+				"m0,k=k0 f=0i 0",
+				"m0,k=k0 f=1i 1000000000",
+				"m0,k=k0 f=2i 2000000000",
+				"m0,k=k0 f=3i 3000000000",
+				"m0,k=k0 f=4i 4000000000",
+				"m0,k=k0 f=6i 10000000000",
+				"m0,k=k0 f=7i 11000000000",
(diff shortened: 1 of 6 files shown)


cmd/influxd/launcher/query_test.go
```diff
@@ -1047,6 +1068,27 @@ from(bucket: v.bucket)
...
+			want: `
+#datatype,string,long,dateTime:RFC3339,long,string,string,string
+#group,false,false,false,false,true,true,true
+#default,_result,,,,,,
+,result,table,_time,_value,_field,_measurement,k
+,,0,1970-01-01T00:00:05Z,10,f,m0,k0
+,,0,1970-01-01T00:00:10Z,22,f,m0,k0
+,,0,1970-01-01T00:00:15Z,35,f,m0,k0
+,,0,1970-01-01T00:00:20Z,5,f,m0,k0
+,,0,1970-01-01T00:00:25Z,,f,m0,k0
```
jlapacik, line 1091: Is count the only aggregate that implicitly fills null values?

jsternberg: I've only looked at sum and count but sum will use a null value while count will use zero.

PR #18737, opened 2020-06-25
chore: update staticcheck and fix newly identified lint checks

jsternberg:


prometheus/prometheus_test.go
```diff
@@ -1,8 +1,8 @@
 package prometheus_test
 
 import (
-	proto "github.com/golang/protobuf/proto"
 	dto "github.com/prometheus/client_model/go"
+	proto "google.golang.org/protobuf/proto"
```
jacobmarble, line 5: Can the alias be removed here?

prometheus/filter_test.go
```diff
@@ -5,10 +5,10 @@ import (
 	"reflect"
 	"testing"
 
-	proto "github.com/golang/protobuf/proto"
 	pr "github.com/influxdata/influxdb/v2/prometheus"
 	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
+	proto "google.golang.org/protobuf/proto"
```
jacobmarble, line 11: Can the alias be removed here?

task.go
```diff
@@ -28,15 +28,21 @@ var (
 	TaskSystemType = "system"
 )
 
+type contextKey string
```
jacobmarble, line 31: What does this type alias benefit? I don't see `contextKey` mentioned anywhere in the diff.

telemetry/push_test.go
```diff
@@ -10,12 +10,12 @@ import (
 	"testing"
 	"time"
 
-	proto "github.com/golang/protobuf/proto"
 	"github.com/google/go-cmp/cmp"
 	"github.com/matttproud/golang_protobuf_extensions/pbutil"
 	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
+	proto "google.golang.org/protobuf/proto"
```
jacobmarble, line 18: Can the alias be removed?

task.go
jsternberg: staticcheck will complain if you don't use a special interface type. If you use a string directly, it complains about the possibility of conflicting with another package.

jsternberg:
@jacobmarble I'll remove the aliases. I got them on some of the packages but missed them in others. Since staticcheck is being updated it will prevent these from coming back since these were all found by staticcheck.

jsternberg:
Yea I'm not really sure about the protobuf change. It seems like staticcheck was saying that the path we were using was deprecated and to switch to the google path instead, but I don't know if the versions are comparable.

PR #18740, opened 2020-06-25
fix(storage): Push-down a predicate to match tags for SHOW MEASUREMENT calls

brettbuddin:
This introduces a predicate to the Storage layer's `MeasurementNames` schema call. This allows InfluxQL to push-down predicates that match measurement names and tags expressed by InfluxQL `SHOW MEASUREMENTS`. Queries using negative comparisons (`!=` and `!~`) are unsupported and will return a `not implemented` error.

📎 influxdata/idpe#7587


tsdb/tsm1/engine_measurement_schema.go
```diff
@@ -52,18 +68,30 @@ func (e *Engine) MeasurementNames(ctx context.Context, orgID, bucketID influxdb.
...
-				name, err := models.ParseMeasurement(key)
+				nameBytes, err := models.ParseMeasurement(key)
 				if err != nil {
 					e.logger.Error("Invalid series key in TSM index", zap.Error(err), zap.Binary("key", key))
 					continue
 				}
 
-				if _, ok := tsmValues[string(name)]; ok {
-					continue
+				name := string(nameBytes)
```
stuartcarnie, line 77: I would recommend you always use this in performance sensitive code (such as this loop):

```go
if _, ok := names[string(nameBytes)] ...
```

As Go won't allocate the string here. Only when you add it to the map, will it allocate a string:

```go
names[string(name)] = struct{}{}
```

stuartcarnie:
@jacobmarble – this PR is ready for your review as a code owner. Note that this fix required for the InfluxQL / Grafana launch next week.

PR #18746, opened 2020-06-26
feat(influx): extend influx export with ability to associate new resources with a stack on export

### LINKED ISSUE #18646: Add ability to associate net new resources with stack-id on export
__Proposal:__
Add the ability to associate newly created resources while exporting them.  This could look like:
`influx export --dashboards <dash id> --labels <label id> <label id> --stack-id <stack id>`

__Current behavior:__
Imagine a scenario where a user has already applied a template and associated the contained resources to a stack.  Now they want to add a new resource (not update an existing one).  They want to then export these new resources and add them to their already-created template manifest file as well as the associated stack.

Currently, even with a stack already initialized, the user must run `influx apply <manifest> --stack-id` to associate new resources with a particular stack.  

This seems like strange behavior for two reasons:
1) The user must first export the new resources (created in the instance) and *then* apply them to the stack.
2) These resources are already present in the UI because they were created there.  It seems redundant to `influx apply` when there is nothing new to add or update in the target instance.

__Desired behavior:__
To be able to associate the output of an export to a stack in the same command.  This will go hand-in-hand with being able to simply update a manifest file as opposed to having to append to it manually.

jsteenb2:
Closes #18646 

- [x] Rebased/mergeable
- [x] Tests pass


cmd/influx/template.go
```diff
@@ -344,6 +345,10 @@ func (b *cmdPkgBuilder) cmdExport() *cobra.Command {
 		# export a stack with resources not associated with the stack
 		influx export --stack-id $STACK_ID --buckets $BUCKET_ID
 
+		# export a stack with resources not associated with the stack
+		# and update the stack resources with all exported resources 
+		influx export --update-stack --stack-id $STACK_ID --buckets $BUCKET_ID
```
jsteenb2, line 350: @sanderson this command has been extended with the new `--update-stack` functionality. I'm still struggling with what to call this flag. I'm not sure about `--append`, I'd like it to be more explicit that the append is to the stack's internally managed resoruces... I'm not sure how to phrase this 😢

sanderson: What about `--include`? "Include resources in the exported template"?

PR #18748, opened 2020-06-26
chore(pkger): drop remove diff field no longer in use

jsteenb2:
this was missed in a previous PR wehre we added the status in place of the
remove boolean flag

- [x] Rebased/mergeable
- [x] Tests pass


jsteenb2:
another failing flux test, going to ignore it

PR #18749, opened 2020-06-26
fix(pkger): update handling of diff dashboard charts

jsteenb2:
backfilling a blind sopt in pkger. The diff dashboard charts were not working correctly across RPC boundaries. The marshal/unmarshal stuffs was off.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/pkger/clone_resource.go
+++ b/pkger/clone_resource.go
@@ -823,7 +823,7 @@ func DashboardToObject(name string, dash influxdb.Dashboard) Object {
 
 	charts := make([]Resource, 0, len(dash.Cells))
 	for _, cell := range dash.Cells {
-		if cell.ID == influxdb.ID(0) {
+		if cell.View == nil {
 			continue
 		}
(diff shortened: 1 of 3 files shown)


pkger/clone_resource.go
```diff
@@ -823,7 +823,7 @@ func DashboardToObject(name string, dash influxdb.Dashboard) Object {
 
 	charts := make([]Resource, 0, len(dash.Cells))
 	for _, cell := range dash.Cells {
-		if cell.ID == influxdb.ID(0) {
+		if cell.View == nil {
```
Asalem1, line 826: curious about this diff. DM me about it

jsteenb2: ahhh that diff, was a hack I added earlier to prevent pkger from exporting empty cells with nothing in them. Now that I have coupled the cell to the view, I can use its availability to signal if it is there or not

PR #18750, opened 2020-06-26
feat(flows): ability to insert cells inline

### LINKED ISSUE #18722: Ability to insert a cell between other cells in a flowbook

alexpaxton:
Closes #18722 

insert-cells-inline

- Render a divider component after each panel which offers an insert menu
- Extend `addCell` function to optionally take an `index` to specify insertion point
- Extend `AddButtons` to optionally take an index and a callback, as well as a required event name for analytics
- Show previous result helper text in all script cells regardless of position
- Add margin at the bottom of the flows list to make resizing the last cell easier
- Add default text to markdown cells
  - If the user adds one and doesn't immediately add text then the cell is blank and strange looking

- [x] Rebased/mergeable
- [x] Tests pass


alexpaxton:
@drdelambre since I used `AddButtons` as a starting point for `InsertCellMenu` there's a lot of repeated code there. I am wondering if it should be extracted into a utility that both components can consume so keeping them in sync isn't an issue down the road

ui/src/notebooks/pipes/Query/index.ts
```diff
@@ -13,7 +13,8 @@ register({
     activeQuery: 0,
     queries: [
       {
-        text: '// Write Flux script here',
+        text:
+          '// Write Flux script here\n// use __PREVIOUS_RESULT__ to continue building from the previous cell',
```
drdelambre, line 17: why this change?

ui/src/notebooks/pipes/markdown/index.ts
```diff
@@ -170,7 +170,7 @@ register({
   component: MarkdownPanel,
   button: 'Markdown',
   initial: () => ({
-    text: TEST_MODE ? TESTCASE : '',
+    text: TEST_MODE ? TESTCASE : 'Type markdown here',
```
drdelambre, line 173: do we wanna make this a markdown comment? `[//]: # (Write Markdown below)`

ui/src/notebooks/components/panel/InsertCellMenu.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+// Constants
+import {NotebookContext} from 'src/notebooks/context/notebook'
+import {PIPE_DEFINITIONS} from 'src/notebooks'
+
+import {event} from 'src/notebooks/shared/event'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
+
+interface Props {
+  index: number
```
drdelambre, line 22: could we just make this an optional prop, and then it's the same thing as the addbuttons?

ui/src/notebooks/components/panel/InsertCellMenu.tsx
```diff
@@ -0,0 +1,87 @@
...
+          text={def.button}
+          onClick={() => {
+            let data = def.initial
+            if (typeof data === 'function') {
+              data = data()
+            }
+
+            onInsert()
+
+            event('Notebook Add Button Clicked', {
```
drdelambre, line 57: should probably change this event name too

drdelambre:
does this also prompt us to drop in an animation library?

drdelambre:
ALSO. if we are unsure of discoverability, we should instrument the hover state (ie: `insert button found` and `insert menu opened`)

ui/src/notebooks/pipes/Query/index.ts
alexpaxton: I added it in the query registration so it appears on all cells regardless of position. I figure showing it all the time doesn't impact usability

ui/src/notebooks/pipes/markdown/index.ts
alexpaxton: I tried the syntax here and it appears as a link, not sure if there is a comment in markdown or nah

PR #18766, opened 2020-06-27
chore: bump dependencies

foobar:
Update some dependencies to align with module prometheus/promethues which
will be need by #17814.


foobar:
Hi @jacobmarble could you take a look ?

jacobmarble:
@foobar this change has only minor dependency updates (that's subjective; dependency updates that I consider minor), so let's close this and rebase #17814 on master-1.x.

Issue #18580, opened 2020-06-17
consolidate all pkger under templates


jsteenb2:
tech debt

Acceptance Criteria
- [x] rename `/packages` and `/packages/apply` to `/templates` and `/templates/apply`
    * this may not be possible yet, need to review collisions with existing templates
- [ ] rename `Pkg` to `Template`
- [x] update CLI as well

PR #18774, opened 2020-06-29
chore(pkger): convert packages endpoints into new naming convention

jsteenb2:
note: going to make this evolution in two steps so that we have a simple
rollback to get back to a working state. We'll be maintaining both packages
and the new templates and stacks endpoints for a while as users start to
move onto a newer CLI version. Sunsetting by end of July.

references: #18580

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/pkger/http_server_packages_deprecated.go
+++ b/pkger/http_server_packages_deprecated.go
@@ -21,19 +21,20 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-const RoutePrefix = "/api/v2/packages"
+const RoutePrefixPackages = "/api/v2/packages"
 
-// HTTPServer is a server that manages the packages HTTP transport.
-type HTTPServer struct {
+// HTTPServerPackages is a server that manages the packages HTTP transport. These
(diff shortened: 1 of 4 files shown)


pkger/http_server_packages_deprecated.go
```diff
@@ -859,7 +860,7 @@ func newJSONEnc(w io.Writer) encoder {
 	return enc
 }
 
-func (s *HTTPServer) encResp(w http.ResponseWriter, r *http.Request, enc encoder, code int, res interface{}) {
+func (s *HTTPServerPackages) encResp(w http.ResponseWriter, r *http.Request, enc encoder, code int, res interface{}) {
 	w.WriteHeader(code)
 	if err := enc.Encode(res); err != nil {
 		s.api.Err(w, r, &influxdb.Error{
```
hoorayimhelping, line 866: is there an explanation of the deprecation? a comment at the top of this file giving a little bit of the context on deprecation would be super helpful to anyone who stumbles upon a file that references being deprecated

jsteenb2: there will be once I get the replacements in place

PR #18776, opened 2020-06-29
fix: update templates contents to handle empty collections

hoorayimhelping:
Template resource collections are made into empty arrays.

Guards against empty or `null` collections coming back from the API.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/templates/components/CommunityTemplateListGroup.tsx
+++ b/ui/src/templates/components/CommunityTemplateListGroup.tsx
@@ -16,7 +16,7 @@ import {
 
 interface Props {
   title: string
-  count?: string
+  count: number
   children?: ReactNode
 }
 
@@ -44,13 +44,11 @@ const CommunityTemplateListGroup: FC<Props> = ({title, count, children}) => {
           <Icon glyph={IconFont.CaretRight} />
         </div>
         <Heading element={HeadingElement.H5}>{title}</Heading>
-        {count && (
-          <Heading
-            element={HeadingElement.Div}
-            appearance={HeadingElement.H6}
-            className="community-templates--list-counter"
-          >{`(${count})`}</Heading>
-        )}
+        <Heading
+          element={HeadingElement.Div}
+          appearance={HeadingElement.H6}
+          className="community-templates--list-counter"
+        >{`(${count})`}</Heading>
       </div>
       {mode === 'expanded' && (
         <FlexBox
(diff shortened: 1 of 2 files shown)


ui/src/templates/reducers/index.ts
```diff
@@ -65,7 +65,17 @@ export const templatesReducer = (
       case SET_ACTIVE_COMMUNITY_TEMPLATE: {
         const {template} = action
 
-        draftState.activeCommunityTemplate = template
+        const activeCommunityTemplate: CommunityTemplate = {}
+        activeCommunityTemplate.dashboards = template.dashboards || []
```
glinton, line 69: any chance template will be null?

ui/src/templates/components/CommunityTemplateListGroup.tsx
```diff
@@ -16,7 +16,7 @@ import {
 
 interface Props {
   title: string
-  count?: string
+  count: number
```
glinton, line 19: i assume the default value of numbers in ts is 0?

ui/src/templates/reducers/index.ts
```diff
@@ -65,7 +65,17 @@ export const templatesReducer = (
...
+        activeCommunityTemplate.telegrafConfigs = template.telegrafConfigs || []
+        activeCommunityTemplate.buckets = template.buckets || []
+        activeCommunityTemplate.checks = template.checks || []
+        activeCommunityTemplate.variables = template.variables || []
+        activeCommunityTemplate.notificationRules =
+          template.notificationRules || []
+        activeCommunityTemplate.labels = template.labels || []
+
+        draftState.activeCommunityTemplate = activeCommunityTemplate
```
alexpaxton, line 78: I wonder if there's a way to handle this using a "default template" and a spread operator. More of a code style thing

ui/src/templates/components/CommunityTemplateListGroup.tsx
hoorayimhelping: @glinton good question - `count` actually comes from https://github.com/influxdata/influxdb/blob/master/ui/src/templates/components/CommunityTemplateContents.tsx#L44 - each collection has `.length` called on it. before, it would crash because it was calling `.length` on `null`. the tricksy change here is to make all the collections default to empty arrays so `.length` always returns 0.

ui/src/templates/reducers/index.ts
hoorayimhelping: @alexpaxton haha that's funny, love where your head's at. I actually had it as

```ts
activeCommunityTemplate.dashboards = [...template.dashboards]
```
but these are set as `null` not `undefined` so spreading doesn't work.

ui/src/templates/reducers/index.ts
glinton: because if so `template.thing` will blow up

PR #18779, opened 2020-06-29
feat(cmd/influx/write): add new processing options and enhancements

### LINKED ISSUE #18741: import csv: add the ability to construct a timestamp from multiple columns
Here is an example CSV:

```
Date,Time,Tag,Value
2020-05-22,00:00:00,test,0
2020-05-22,00:05:00,test,1
2020-05-22,00:10:00,test,2
```

I would like a way to combine multiple time columns into a single timestamp to be used on the import. In this case, the resulting line protocol timestamps would combine the `Date` and the `Time` column to produce:

```
1590105600000000000
1590105900000000000
1590106400000000000
```

other example CSVs that should be able to be parsed with this solution:
```
Year,Month,Day,Hour,Minute,Second,Tag,Value
2020,05,22,00,00,00,test,0
2020,05,22,00,05,00,test,1
2020,05,22,00,10,00,test,2
```

the current behavior prints a warning:
`2020-06-25T21:47:08.731185Z	info	WARNING: at most one dateTime column is expected, 'date' column is ignored`

### LINKED ISSUE #18742: CSV import: add the ability to write rejected records to a specific file
As a user, I would like to use this capability to ingest a large number of CSV files, and some of them might have data quality issues.

I would like a way to provide a file location where the rejected records could be written so that they could be monitored and re-processed later.

something like `influx write -f my.csv --errors-file errors.csv`

This would constantly append any records that it could not process to that file.

### LINKED ISSUE #18744: import csv: add the option to fail the import if datatype is a long and a float is in the data
```
t,f
my_tag,1
my_tag,1.9
```

right now, both of these work:
```
➜  influxdb git:(master) ./bin/darwin/influx write dryrun -f ~/Downloads/float.csv --header "t|tag,f|long" --header "#constant measurement,m" --skipHeader
m,t=my_tag f=1i
m,t=my_tag f=1i
➜  influxdb git:(master) ./bin/darwin/influx write dryrun -f ~/Downloads/float.csv --header "t|tag,f|double" --header "#constant measurement,m" --skipHeader
m,t=my_tag f=1
m,t=my_tag f=1.9
```

as a user, I would like a way to fail the import if i have a column marked as a long and a float value is truncated. possibbly via a `--strict` flag? Also, in all cases, I would expect a warning printed similar to the case of multiple date fields indicating a loss of precision.

sranka:
Closes #18744 : **indicate a loss of precision when parsing long/unsignedLong data types**

When parsing `long` or `unsignedLong` value from a string value with fraction digits, the whole CSV row fails when in a strict mode, which is turned on using a column dataFormat that starts with `strict`, such as `long:strict`. A warning is printed when not in a strict mode, saying `line x: column y: '1.2' truncated to '1' to fit into long data type`.  [Package documentation](https://github.com/influxdata/influxdb/compare/18744/csv2lp?expand=1#diff-845cb71dbfa49b093ae411b57e24e057R161) explains strict parsing.

Closes #18742 : **add the ability to write rejected records to a specific file**

There is a new --error-file option in `influx write`
```
   --errors-file string   The path to the file to write rejected rows to
```
The file then contains rows than cannot be imported because of data quality issues, CSV comments are also written to let the user know of what was wrong, for example:
```
# error : line 3: column 'a': '1.1' cannot fit into long data type
cpu,1.1
```

Closes #18741: **add the ability to construct a timestamp from multiple columns**

A new `#concat` annotation adds a new column that is concatenated from existing columns according to bash-like string interpolation literal with variables referencing existing column labels. For example: `#concat,string,fullName,${firstName} ${lastName}`

- [ ] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/pkg/csv2lp/csv2lp_test.go
+++ b/pkg/csv2lp/csv2lp_test.go
@@ -218,6 +218,10 @@ func Test_CsvLineError(t *testing.T) {
 			CsvLineError{Line: 2, Err: CsvColumnError{"a", errors.New("cause")}},
 			"line 2: column 'a': cause",
 		},
+		{
+			CsvLineError{Line: -1, Err: CsvColumnError{"a", errors.New("cause")}},
+			"column 'a': cause",
+		},
 	}
 	for _, test := range tests {
 		require.Equal(t, test.value, test.err.Error())
--- a/pkg/csv2lp/csv2lp_test.go
+++ b/pkg/csv2lp/csv2lp_test.go
@@ -204,6 +204,40 @@ func Test_CsvToLineProtocol_SkipRowOnError(t *testing.T) {
 	require.Equal(t, messages, 2)
 }
 
+// Test_CsvToLineProtocol_RowSkipped tests that error rows are reported to configured RowSkippedListener
+func Test_CsvToLineProtocol_RowSkipped(t *testing.T) {
+	var buf bytes.Buffer
+	log.SetOutput(&buf)
+	oldFlags := log.Flags()
+	log.SetFlags(0)
+	oldPrefix := log.Prefix()
+	prefix := "::PREFIX::"
+	log.SetPrefix(prefix)
+	defer func() {
+		log.SetOutput(os.Stderr)
+		log.SetFlags(oldFlags)
+		log.SetPrefix(oldPrefix)
+	}()
+
+	csv := "sep=;\n_measurement;a|long:strict\n;1\ncpu;2.1\ncpu;3a\n"
+
+	reader := CsvToLineProtocol(strings.NewReader(csv)).SkipRowOnError(true)
+	reader.RowSkipped = func(src *CsvToLineReader, err error, _row []string) {
+		log.Println(err, string(src.Comma()))
+	}
+	// read all the data
+	ioutil.ReadAll(reader)
+
+	out := buf.String()
+	// fmt.Println(out, string(';'))
+	// ::PREFIX::line 3: column '_measurement': no measurement supplied
+	// ::PREFIX::line 4: column 'a': '2.1' cannot fit into long data type
+	// ::PREFIX::line 5: column 'a': strconv.ParseInt: parsing "3a": invalid syntax
+	messages := strings.Count(out, prefix)
+	require.Equal(t, 3, messages)
+	require.Equal(t, 3, strings.Count(out, ";"))
+}
+
 // Test_CsvLineError tests CsvLineError error format
 func Test_CsvLineError(t *testing.T) {
 	var tests = []struct {
--- a/pkg/csv2lp/csv2lp_test.go
+++ b/pkg/csv2lp/csv2lp_test.go
@@ -204,38 +204,66 @@ func Test_CsvToLineProtocol_SkipRowOnError(t *testing.T) {
 	require.Equal(t, messages, 2)
 }
 
-// Test_CsvToLineProtocol_RowSkipped tests that error rows are reported to configured RowSkippedListener
+// Test_CsvToLineProtocol_RowSkipped tests that error rows are reported to configured RowSkipped listener
 func Test_CsvToLineProtocol_RowSkipped(t *testing.T) {
 	var buf bytes.Buffer
 	log.SetOutput(&buf)
 	oldFlags := log.Flags()
 	log.SetFlags(0)
-	oldPrefix := log.Prefix()
-	prefix := "::PREFIX::"
-	log.SetPrefix(prefix)
 	defer func() {
 		log.SetOutput(os.Stderr)
 		log.SetFlags(oldFlags)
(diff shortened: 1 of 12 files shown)


pkg/csv2lp/csv2lp_test.go
```diff
@@ -204,6 +204,40 @@ func Test_CsvToLineProtocol_SkipRowOnError(t *testing.T) {
...
+		log.SetOutput(os.Stderr)
+		log.SetFlags(oldFlags)
+		log.SetPrefix(oldPrefix)
+	}()
+
+	csv := "sep=;\n_measurement;a|long:strict\n;1\ncpu;2.1\ncpu;3a\n"
+
+	reader := CsvToLineProtocol(strings.NewReader(csv)).SkipRowOnError(true)
+	reader.RowSkipped = func(src *CsvToLineReader, err error, _row []string) {
+		log.Println(err, string(src.Comma()))
```
GeorgeMac, line 226: The other tests in this library intercept the calls to `log` in order to validate the `reader` calls `log` appropriately.
Here `log.Println(...)` is being called explicitly in your test.

Could we change this to gather up all the arguments passed to each call to `RowSkipped` and explicitly check they are as expected. Rather than intercepting `log` calls and ensuring that it is just called three times.

i.e. specifically assert that each of the following contents is within the error:
```
line 3: column '_measurement': no measurement supplied
line 4: column 'a': '2.1' cannot fit into long data type
line 5: column 'a': strconv.ParseInt: parsing "3a": invalid syntax
```

sranka: The test also checks that no warning messages are logged when RowSkipped is set (using a logger). Yes, explicit validattion of RowSkipped parameters will improve the readability of the test.

GeorgeMac:
Thanks for addressing my concern. It looks much better.
I would defer to the other reviewers who may be more familiar with this part. As I don't have much context on this part of the codebase.
But my static pass of the code looks good from my end.

pkg/csv2lp/csv_annotations.go
```diff
@@ -79,10 +79,60 @@ func constantSetupTable(table *CsvTable, row []string) error {
...
+	}
+	// add a virtual column to the table
+	table.extraColumns = append(table.extraColumns, &col)
+	// add validator to report error when no placeholder column is available
+	table.validators = append(table.validators, func(table *CsvTable) error {
+		placeholders := computedReplacer.FindAllString(template, len(template))
+		for _, placeholder := range placeholders {
+			columnLabel := placeholder[2 : len(placeholder)-1] // ${columnLabel}
+			if columnLabel == "$" {
+				return nil // ${$} is a way to print $
```
glinton, line 121: stray comments?

pkg/csv2lp/data_conversion.go
```diff
@@ -82,16 +83,17 @@ func escapeString(val string) string {
...
 // are simply removed, they are typically used to visually separate groups in large numbers.
-// The removeFaction parameter controls whether the returned value can contain also the fraction part.
+// The removeFraction parameter controls whether the returned value can contain also the fraction part.
+// An empty format means ". \n\t\r_"
 //
 // For example, to get a strconv-parseable float from a Spanish value '3.494.826.157,123', use format ",." .
-func normalizeNumberString(value string, format string, removeFraction bool) string {
-	if len(format) == 0 {
+func normalizeNumberString(value string, format string, removeFraction bool) (normalized string, truncated bool) {
+	if format == "" {
```
glinton, line 96:
```suggestion
	if len(format) == 0 {
```
the former method is technically faster

pkg/csv2lp/data_conversion.go
```diff
@@ -110,20 +112,20 @@ func normalizeNumberString(value string, format string, removeFraction bool) str
 			}
 			if c == fractionRune {
 				if removeFraction {
-					break ForAllCharacters
+					return retVal.String(), true
```
glinton, line 115: i assume this new behavior is expected (not looping through values again)?

glinton (approved):

pkg/csv2lp/csv_annotations.go
sranka: :+1:  there is no need to do so

pkg/csv2lp/data_conversion.go
sranka: it is intended, the function was changed to return also a truncated indicator, which is `true` only in this case

pierwill:
The following pages need updating:

- https://docs.influxdata.com/influxdb/v2.0/reference/syntax/annotated-csv/extended/
- https://docs.influxdata.com/influxdb/v2.0/reference/cli/influx/write/
- https://docs.influxdata.com/influxdb/v2.0/write-data/developer-tools/csv/

PR #18780, opened 2020-06-29
feat: Add initial OSS-Fuzz testing integration

pierwill:
For a basic overview of our fuzzing initiative and [OSS-Fuzz](https://google.github.io/oss-fuzz/) integration, see ~the README~ FUZZ.md included in this PR.

There are many ways to integrate fuzz tests into a project. Instead of the structure proposed here, we might prefer to do things [the way Golang does](https://github.com/golang/go/tree/master/src/html), with fuzz tests included in the packages themselves. This can evolve over time as we find something that works for us.

### DIFF
--- a/FUZZ.md
+++ b/FUZZ.md
@@ -1,0 +1,25 @@
+# Fuzzing InfluxDB
+
+## Continuous fuzzing
+
+The InfluxDB repository contains testing targets for InfluxDB's integration with
+[OSS-Fuzz](https://google.github.io/oss-fuzz/).
+
+The build scripts for this integration are contained [in the OSS-Fuzz repo](https://github.com/google/oss-fuzz/tree/master/projects/influxdb).
+
+## Local fuzzing
+
+For local fuzzing, install [go-fuzz](https://github.com/dvyukov/go-fuzz):
+
+```
+$ go get -u github.com/dvyukov/go-fuzz/go-fuzz github.com/dvyukov/go-fuzz/go-fuzz-build
+```
+
+Below is an example of building and running a fuzz test.
+In this case, the test is located at `./jsonweb/fuzz.go`.
+
+```
+$ cd go/src/github.com/influxdata/influxdb/jsonweb
+$ go-fuzz-build github.com/influxdata/influxdb/v2/jsonweb
+$ go-fuzz -bin jsonweb-fuzz.zip
+```
--- a/jsonweb/fuzz.go
+++ b/jsonweb/fuzz.go
@@ -1,0 +1,24 @@
+// +build gofuzz
+
+package jsonweb
+
+// FuzzJsonWeb is the entry point for fuzzing when built with go-fuzz-build.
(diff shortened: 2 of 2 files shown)


fuzz/fuzzJsonWeb.go
```diff
@@ -0,0 +1,24 @@
+// +build gofuzz
+
+package fuzz
+
+import (
+	"github.com/influxdata/influxdb/v2/jsonweb"
+)
+
+func FuzzJsonWeb(data []byte) int {
```
rbetts, line 9: What does this function to?

pierwill:
> Can you clarify what FuzzJsonWeb does? Not sure what's enabled exactly (with OSS fuzz) by this PR.

For go-fuzz to build a binary for fuzzing, we write a function with the following signature: `func Fuzz(data []byte) int`. From go-fuzz:

> Data is a random input generated by go-fuzz, note that in most cases it is invalid. The function must return 1 if the fuzzer should increase priority of the given input during subsequent fuzzing (for example, the input is lexically correct and was parsed successfully); -1 if the input must not be added to corpus even if gives new coverage; and 0 otherwise; other values are reserved for future use.

In this example, `FuzzJsonWeb` sets up the required structs in order to exercise [this function](https://github.com/influxdata/influxdb/blob/0c268e205b5d876f980d78ebe2ba1573bf752208/jsonweb/token.go#L54).

I should note that there is something of an art to identifying good fuzz targets and writing these tests. This particular test (`FuzzJsonWeb`) is included here mostly as a way to test the OSS-Fuzz integration once https://github.com/google/oss-fuzz/pull/4038 merges.

fuzz/fuzzJsonWeb.go
```diff
@@ -0,0 +1,24 @@
+// +build gofuzz
```
jsternberg, line 1: Is it possible to move this file into `internal/fuzz/jsonweb/jsonweb.go` and change the fuzzing function name to the standard `Fuzz`? I think it will make it easier to use.

fuzz/fuzzJsonWeb.go
```diff
@@ -0,0 +1,24 @@
...
+	var keyStore = jsonweb.KeyStoreFunc(func(kid string) ([]byte, error) {
+		if kid != "some-key" {
+			return nil, jsonweb.ErrKeyNotFound
+		}
+
+		return []byte("correct-key"), nil
+	})
+
+	parser := jsonweb.NewTokenParser(keyStore)
+	_, err := parser.Parse(string(data))
```
jsternberg, line 19: Nitpick:

```go
if _, err := parser.Parse(string(data)); err != nil {
    return 0
}
return 1

fuzz/fuzzJsonWeb.go
pierwill: That should work; I'll test it out. Do you think in that case it might be better to promote the targets to live in package directories themselves (as opposed to shadowing them in `internal`)? 🤔

PR #18781, opened 2020-06-29
fix: tweak flagging structure of Community Templates

hoorayimhelping:
Tweaks community templates' structure around feature flagging

Previously the flagging occurred in `index.tsx`, but now the flagging is handled by `TemplatesIndex.tsx`, which either renders itself normally or conditionally renders the flagged template path.

#### Testing Instructions

to toggle the flag, start your local influxd process with:

```sh
```

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/templates/containers/TemplatesIndex.tsx
+++ b/ui/src/templates/containers/TemplatesIndex.tsx
@@ -10,14 +10,18 @@ import SettingsHeader from 'src/settings/components/SettingsHeader'
 import TemplatesPage from 'src/templates/components/TemplatesPage'
 import GetResources from 'src/resources/components/GetResources'
 
+import {CommunityTemplatesIndex} from 'src/templates/containers/CommunityTemplatesIndex'
+
 // Utils
 import {pageTitleSuffixer} from 'src/shared/utils/pageTitles'
 import {getOrg} from 'src/organizations/selectors'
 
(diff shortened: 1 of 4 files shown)


ui/src/templates/containers/TemplatesIndex.tsx
```diff
@@ -26,7 +30,10 @@ type Props = WithRouterProps & StateProps
 @ErrorHandling
 class TemplatesIndex extends Component<Props> {
   public render() {
-    const {org, children} = this.props
+    const {org, children, flags} = this.props
+    if (flags.communityTemplates) {
```
drdelambre, line 34: why pull all the flags in here and not just use the util function of `isFlagEnabled`?

drdelambre: NOTE: to trigger re-render from the `connect` HOC

drdelambre:
context: https://github.com/influxdata/influxdb/pull/18708

ui/src/templates/containers/TemplatesIndex.tsx
hoorayimhelping: let's think of a less ad-hoc way to do this so folks in the future don't have to rely on this, if possible

Issue #7792, opened 2017-01-05
ORDER BY time DESC breaks aggregation without explicit where condition


joshliberty:
### Bug report

__System info:__ 
InfluxDB v1.1.0
OS: Debian 64bit

__Steps to reproduce:__

1. Create a measurement, for example "mes"
2. Insert some data:
```
insert mes,org=3 total=240 1483203600000000000
insert mes,org=3 total=100,errors=100 1483207200000000000 
insert mes,org=4 total=3 1483203600000000000
insert mes,org=4 errors=100,total=100 1483207200000000000
```

3. Run the following query:
`select sum(total), sum(errors) as "errors" from mes where time >= '2016-12-31 17:00:00' AND time < '2016-12-31 19:00:00' group by org, time(1h) order by time DESC`

__Expected behavior:__
I expect there to be 4 total rows in the response, corresponding with the inserted data.

__Actual behavior:__
What I actually get is:

```
> select sum(total), sum(errors) as "errors" from mes where time >= '2016-12-31 17:00:00' AND time < '2016-12-31 19:00:00' group by org, time(1h) order by time DESC   
name: mes                                                                                                                                                              
tags: org=3                                                                                                                                                            
time                    sum     errors                                                                                                                                 
----                    ---     ------                                                                                                                                 
1483207200000000000             100                                                                                                                                    
1483203600000000000                                                                                                                                                    
                                                                                                                                                                       
name: mes                                                                                                                                                              
tags: org=4                                                                                                                                                            
time                    sum     errors                                                                                                                                 
----                    ---     ------                                                                                                                                 
1483207200000000000     100     100                                                                                                                                    
1483203600000000000     3                                                                                                                                              
                                                                                                                                                                       
name: mes                                                                                                                                                              
tags: org=3                                                                                                                                                            
time                    sum     errors                                                                                                                                 
----                    ---     ------                                                                                                                                 
1483207200000000000     100                                                                                                                                            
1483203600000000000     240                                                                                                                                            
```

When I run the same thing without ordering by time descending (either with no ordering, or ordering ascending), then everything is fine. **Alternatively, I can set "and org='3'" and the result will also be fine.**

rbetts:
Let's try to reproduce this against 1.5.x

PR #18785, opened 2020-06-29
fix(tsm1): delimit tsmKeyPrefix with appended comma

### LINKED ISSUE #7589: Import does not work with unsafeSsl flag
### Bug report

__System info:__ [InfluxDB version:1.0.2 (running offical docker image), ubuntu 16.04]

__Steps to reproduce:__

1. Setup Influxdb-Server with selfsigned Certificates
2. Import using the -unsafeSsl flag:
`influx -import -path=myexport -pps 10000 --unsafeSsl --ssl`

__Expected behavior:__

Import should work

__Actual behavior:__

On the console I get the following error:
`ERROR: failed to connect to https://localhost:8086`

The logfile contains the following entries:

```
[httpd] ::1 - admin [05/Nov/2016:10:57:16 +0000] "GET /ping HTTP/1.1" 204 0 "-" "InfluxDBShell/1.0.2" 9d77c543-a346-11e6-8004-000000000000 23
[run] 2016/11/05 10:57:16 http: TLS handshake error from [::1]:33652: remote error: bad certificate
```
__Additional info:__

When connecting to the same database without the `-import` flag, then the connection works:

```
influx --unsafeSsl --ssl
```
```
Connected to https://localhost:8086 version 1.0.2
InfluxDB shell version: 1.0.2
>
```
It is also possible to send data to the db with https, so setup of the certificates seems to be ok.

When disabling https and then importing the same file it also works. It seems the import ignores the unsafeSsl flag.

gavincabbage:
Fixes #7589.

Append a comma to the TSM key prefix when matching a full measurement name to avoid erroneously matching other measurement names that include the prefix in their own name. For example, this prevents matching a measurement "cpu1" when targeting "cpu" by updating the prefix to "cpu,". This relies on the fact that tag key-value pairs are separated by commas.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/tsdb/tsm1/engine_measurement_schema.go
+++ b/tsdb/tsm1/engine_measurement_schema.go
@@ -348,6 +348,7 @@ func (e *Engine) fieldsPredicate(ctx context.Context, orgID influxdb.ID, bucketI
 
 	mt := models.Tags{models.NewTag(models.MeasurementTagKeyBytes, measurement)}
 	tsmKeyPrefix := mt.AppendHashKey(orgBucketEsc)
+	tsmKeyPrefix = append(tsmKeyPrefix, ',')
 
 	var canceled bool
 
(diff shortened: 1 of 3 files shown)


stuartcarnie:
Not sure what was going on with the failures in Jenkins:

```
    --- FAIL: TestFluxEndToEnd/date (0.08s)
        --- FAIL: TestFluxEndToEnd/date/hour_duration (0.00s)
            end_to_end_test.go:225: json: cannot unmarshal string into Go struct field bucketResponse.labels of type []influxdb.Label
        ...
```

 and litmus, but I've restarted them

stuartcarnie:
@gavincabbage I think you will need to wait for #7792 to merge and then rebase again

PR #18795, opened 2020-06-29
chore(pkger): drop mentions of pkgs from within service model

jsteenb2:
references: #18580

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influx/template.go
+++ b/cmd/influx/template.go
@@ -27,29 +27,29 @@ import (
 	input "github.com/tcnksm/go-input"
 )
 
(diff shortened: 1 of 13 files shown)


drdelambre:
WOOOO the contract tests finally kicked in! seems like all of the TS defs that were extended / implemented need to be updated too

jsteenb2:
nice, this is simple to track down the contract's that break from the swagger 👏 . Nice work @influxdata/monitoring-team!

PR #18796, opened 2020-06-29
feat: add api to extract flux option references

yzhang1991:


query/reference.go
```diff
@@ -0,0 +1,214 @@
...
+	return o
+}
+
+func (o *optionPropertyReferenceCollector) Done(node semantic.Node) {}
+
+func (o *optionPropertyReferenceCollector) GetResult() map[string]*OptionPropertyReference {
+	return o.refs
+}
+
+func FindFluxOptionReferences(ctx context.Context, source string, extern json.RawMessage) (map[string]*OptionPropertyReference, error) {
```
jsteenb2, line 171: nice! yessss

PR #18805, opened 2020-06-30
fix(release): create static builds for linux amd64

### LINKED ISSUE #17675: Build Linux Static binaries for InfluxDB 2.0
similar to the ones we have for 1.x, we will need a `influxdb-2.0.0-static_linux_amd64.tar.gz` binary

jsternberg:
Fixes #17675.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -8,13 +8,16 @@ builds:
       - amd64
       - arm64
     main: ./cmd/influx/
+    flags:
+      - -tags={{if and (eq .Os "linux") (eq .Arch "amd64")}}osusergo,netgo,static_build{{end}}
     env:
       - GO111MODULE=on
       - CGO_ENABLED=1
       - CC=xcc
       - PKG_CONFIG=$GOPATH/bin/pkg-config
       - MACOSX_DEPLOYMENT_TARGET=10.11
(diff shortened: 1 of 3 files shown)


.goreleaser.yml
```diff
@@ -8,13 +8,16 @@ builds:
...
+      - -tags={{if and (eq .Os "linux") (eq .Arch "amd64")}}osusergo,netgo,static_build{{end}}
     env:
       - GO111MODULE=on
       - CGO_ENABLED=1
       - CC=xcc
       - PKG_CONFIG=$GOPATH/bin/pkg-config
       - MACOSX_DEPLOYMENT_TARGET=10.11
-    ldflags: -s -w -X main.version={{.Version}} -X main.commit={{.ShortCommit}} -X main.date={{.Date}}
+    ldflags:
+      - -s -w -X main.version={{.Version}} -X main.commit={{.ShortCommit}} -X main.date={{.Date}} {{if and (eq .Os "linux") (eq .Arch "amd64")}}-extldflags "-fno-PIC -static -Wl,-z,stack-size=8388608"{{end}}
```
wolffcm, line 20: I'm curious, why do you need to set `stack-size` here, is that related to the issue Adrian fixed where we were blowing out the stacks in Rust code?

jsternberg: Yes. That's how to reset the stack size for musl back to the 8 MB.

PR #18823, opened 2020-07-01
chore: remove kv Session service and sessionsv1 bucket

GeorgeMac:
This removes the bucket `sessionv1` from the underlying metadata store.
It does so using a migration.

It marks the `sessionsv1` bucket in `kv` as deprecated.

It also removes the now defunct session code.
The top-level `sessions` package handles this now (along with the `sessionv2` bucket).


GeorgeMac:
I did indeed. In cloud the delete is a no op 👍 
But this does stop it from getting warmed in the cache.

PR #17254, opened 2020-03-13
fix(tsi1): wait deleting epoch before dropping shard

DengzhiLiu:
Closes #16834 

Describe your proposed changes here.

- [ ] [CHANGELOG.md](https://github.com/influxdata/influxdb/blob/master/CHANGELOG.md) updated with a link to the PR (not the Issue)
- [x] [Well-formatted commit messages](https://www.conventionalcommits.org/en/v1.0.0-beta.3/)
- [ ] Rebased/mergeable
- [x] Tests pass
- [ ] http/swagger.yml updated (if modified Go structs or API)
- [ ] Documentation updated or issue created (provide link to issue/pr)
- [ ] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)

PR #18827, opened 2020-07-01
fix(tsi1): wait deleting epoch before dropping shard

dgnorton:
Forward port https://github.com/influxdata/influxdb/pull/17254 to 1.8


dgnorton:
My mistake...this has already been forward ported to 1.8.

PR #18836, opened 2020-07-01
fix: window tables are empty if cursor is empty

jlapacik:


storage/flux/table.gen.go.tmpl
```diff
@@ -271,6 +271,9 @@ func (t *{{.name}}WindowTable) appendValues(intervals []int64, appendValue func(
 }
 
 func (t *{{.name}}WindowTable) advance() bool {
+	if !t.nextBuffer() {
```
jacobmarble, line 274: That was easy to review.

PR #18852, opened 2020-07-02
chore(flagged): rename variables

hoorayimhelping:
Community Templates background work. Changes our internal representation of templates to match pkgr's model.

When we send a yaml, json, or jsonnet template file to the pkgr API, it returns an object with a `summary` of the template, and a `diff` of the changes, among other things. Previously, our redux store used `summary` and discarded everything else. Now it's storing the entire object returned from the API. The frontend still only uses `summary` at the moment, but that will change in the future.

These commits were split off of work to install templates which is still underway.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/templates/actions/creators.ts
+++ b/ui/src/templates/actions/creators.ts
@@ -11,7 +11,8 @@ export const ADD_TEMPLATE_SUMMARY = 'ADD_TEMPLATE_SUMMARY'
 export const GET_TEMPLATE_SUMMARIES_FOR_ORG = 'GET_TEMPLATE_SUMMARIES_FOR_ORG'
 export const POPULATE_TEMPLATE_SUMMARIES = 'POPULATE_TEMPLATE_SUMMARIES'
 export const REMOVE_TEMPLATE_SUMMARY = 'REMOVE_TEMPLATE_SUMMARY'
-export const SET_ACTIVE_COMMUNITY_TEMPLATE = 'SET_ACTIVE_COMMUNITY_TEMPLATE'
+export const SET_COMMUNITY_TEMPLATE_TO_INSTALL =
+  'SET_COMMUNITY_TEMPLATE_TO_INSTALL'
 export const SET_EXPORT_TEMPLATE = 'SET_EXPORT_TEMPLATE'
 export const SET_TEMPLATE_SUMMARY = 'SET_TEMPLATE_SUMMARY'
 export const SET_TEMPLATES_STATUS = 'SET_TEMPLATES_STATUS'
@@ -25,7 +26,7 @@ export type Action =
   | ReturnType<typeof setExportTemplate>
   | ReturnType<typeof setTemplatesStatus>
   | ReturnType<typeof setTemplateSummary>
-  | ReturnType<typeof setActiveCommunityTemplate>
+  | ReturnType<typeof setCommunityTemplateToInstall>
   | ReturnType<typeof toggleTemplateResourceInstall>
 
 type TemplateSummarySchema<R extends string | string[]> = NormalizedSchema<
@@ -83,9 +84,9 @@ export const setTemplateSummary = (
     schema,
   } as const)
 
-export const setActiveCommunityTemplate = (template: CommunityTemplate) =>
+export const setCommunityTemplateToInstall = (template: CommunityTemplate) =>
   ({
-    type: SET_ACTIVE_COMMUNITY_TEMPLATE,
+    type: SET_COMMUNITY_TEMPLATE_TO_INSTALL,
     template,
   } as const)
 
--- a/ui/src/templates/components/CommunityTemplateContents.tsx
+++ b/ui/src/templates/components/CommunityTemplateContents.tsx
@@ -21,7 +21,7 @@ import {toggleTemplateResourceInstall} from 'src/templates/actions/creators'
 import {getResourceInstallCount} from 'src/templates/selectors'
 
 interface StateProps {
-  activeCommunityTemplate: CommunityTemplate
+  summary: CommunityTemplate
 }
 
 interface DispatchProps {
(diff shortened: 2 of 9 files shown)


ui/src/templates/components/CommunityTemplateContents.tsx
```diff
@@ -21,7 +21,7 @@ import {toggleTemplateResourceInstall} from 'src/templates/actions/creators'
 import {getResourceInstallCount} from 'src/templates/selectors'
 
 interface StateProps {
-  activeCommunityTemplate: CommunityTemplate
+  summary: CommunityTemplate
```
alexpaxton, line 24: ahhhh so nice

ui/src/templates/actions/creators.ts
```diff
@@ -11,7 +11,8 @@ export const ADD_TEMPLATE_SUMMARY = 'ADD_TEMPLATE_SUMMARY'
 export const GET_TEMPLATE_SUMMARIES_FOR_ORG = 'GET_TEMPLATE_SUMMARIES_FOR_ORG'
 export const POPULATE_TEMPLATE_SUMMARIES = 'POPULATE_TEMPLATE_SUMMARIES'
 export const REMOVE_TEMPLATE_SUMMARY = 'REMOVE_TEMPLATE_SUMMARY'
-export const SET_ACTIVE_COMMUNITY_TEMPLATE = 'SET_ACTIVE_COMMUNITY_TEMPLATE'
+export const SET_COMMUNITY_TEMPLATE_TO_INSTALL =
+  'SET_COMMUNITY_TEMPLATE_TO_INSTALL'
```
alexpaxton, line 15: "Community Template" feels super long but also necessary for the time being. Will there be a point at which we can go back and rename this stuff "Template"?

ui/src/templates/components/CommunityTemplateImportOverlay.tsx
```diff
@@ -89,7 +89,9 @@ const mstp = (state: AppState, props: Props): StateProps => {
     org,
     templateName: props.params.templateName,
     flags: state.flags.original,
-    resourceCount: getTotalResourceCount(state),
+    resourceCount: getTotalResourceCount(
+      state.resources.templates.communityTemplateToInstall.summary
```
alexpaxton, line 93: This seems like a really deep get, are there any guards and/or a reasonable default here?

ui/src/templates/containers/CommunityTemplatesIndex.tsx
```diff
@@ -148,25 +143,14 @@ class UnconnectedTemplatesIndex extends Component<Props> {
...
-        throw new Error((resp.data as PkgError).message)
-      }
+      const summary = await reviewTemplate(orgID, yamlLocation)
 
-      const summary = (resp.data as TemplateSummary).summary
-      this.props.setActiveCommunityTemplate(summary)
+      this.props.setCommunityTemplateToInstall(summary)
       return summary
     } catch (err) {
       console.error(err)
```
alexpaxton, line 156: Is this the error catch for submitting a valid URL or file location but the API can't return a summary? I am wondering if this needs to be visualized in the UI

alexpaxton:
Also, the term "stage" comes to mind. So perhaps `STAGE_TEMPLATE` instead of `SET_TEMPLATE_FOR_INSTALL` 

Just an idea!

ui/src/templates/components/CommunityTemplateContents.tsx
```diff
@@ -196,10 +192,10 @@ class CommunityTemplateContentsUnconnected extends PureComponent<Props> {
         </CommunityTemplateListGroup>
         <CommunityTemplateListGroup
           title="Labels"
-          count={getResourceInstallCount(activeCommunityTemplate.labels)}
+          count={getResourceInstallCount(summary.labels)}
         >
-          {Array.isArray(activeCommunityTemplate.labels) &&
-            activeCommunityTemplate.labels.map(label => {
+          {Array.isArray(summary.labels) &&
+            summary.labels.map(label => {
```
ebb-tide, line 198: I wonder if you can apply defaults before `summary` enters the redux store so you don't have to check `Array.isArray` every time you access these parameters?

ui/src/templates/actions/creators.ts
ebb-tide: not until we resolve the two template concepts in the UI !

hoorayimhelping: good question. i've been calling these community templates to distinguish them from regular templates while they're both live in the wild. if we end up going with the community template implementation, i think a final step in turning off this feature flag and making them part of the system would be renaming them to just Templates. once the old templates go away, we should start calling these just `templates`, yeah.

ui/src/templates/components/CommunityTemplateImportOverlay.tsx
hoorayimhelping: it definitely is, good call. `state.resources.templates` is taken care of at the resources level, `communityTemplateToInstall` is taken care of having a default shape here: https://github.com/influxdata/influxdb/pull/18852/files#diff-03c3cd4b5faeff2eaea0b8c3c19df44bR28

ui/src/templates/components/CommunityTemplateContents.tsx
hoorayimhelping: good callout, i think at this point, these checks aren't necessary. i'll look into ditching them

ui/src/templates/containers/CommunityTemplatesIndex.tsx
hoorayimhelping: @alexpaxton i think we'll need to add some ability to display errors to the users. right now the ui is just kind of swallowing them. i'm not quite sure what errors we'll need to show though

PR #18871, opened 2020-07-06
chore(flagged): update templates to use react router 5 hotness

hoorayimhelping:
Updates community templates to be compatible with react router 5 changes.

Mostly shuffling code around, [this](https://github.com/influxdata/influxdb/pull/18871/files#diff-b8bbbcc08236ad73ba323c538d2d3c51R53) is the only relevant change. 

See: https://codedaily.io/tutorials/48/Use-matchPath-to-Match-Nested-Route-Paths-in-Parent-Routes-with-React-Router

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/templates/components/CommunityTemplateImportOverlay.tsx
+++ b/ui/src/templates/components/CommunityTemplateImportOverlay.tsx
@@ -6,6 +6,7 @@ import {connect} from 'react-redux'
 import {CommunityTemplateInstallerOverlay} from 'src/templates/components/CommunityTemplateInstallerOverlay'
 
 // Actions
+import {setCommunityTemplateToInstall} from 'src/templates/actions/creators'
 import {createTemplate as createTemplateAction} from 'src/templates/actions/thunks'
 import {notify as notifyAction} from 'src/shared/actions/notifications'
 
@@ -19,6 +20,11 @@ import {ComponentStatus} from '@influxdata/clockface'
 
 // Utils
 import {getByID} from 'src/resources/selectors'
+import {
+  getGithubUrlFromTemplateName,
+  getRawUrlFromGithub,
+  reviewTemplate,
+} from 'src/templates/utils'
 
 interface State {
   status: ComponentStatus
@@ -27,6 +33,7 @@ interface State {
 interface DispatchProps {
   createTemplate: typeof createTemplateAction
   notify: typeof notifyAction
+  setCommunityTemplateToInstall: typeof setCommunityTemplateToInstall
 }
 
 interface StateProps {
@@ -45,6 +52,12 @@ class UnconnectedTemplateImportOverlay extends PureComponent<Props> {
     status: ComponentStatus.Default,
   }
 
+  public componentDidMount() {
+    const {org, templateName} = this.props
+
+    this.reviewTemplateResources(org.id, templateName)
+  }
+
   public render() {
     if (!this.props.flags.communityTemplates) {
       return null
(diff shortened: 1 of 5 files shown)


ui/src/templates/containers/CommunityTemplatesIndex.tsx
```diff
@@ -41,38 +41,26 @@ interface StateProps {
...
-        )
+    const match = matchPath(this.props.location.pathname, {
+      path: communityTemplatesImportPath,
+    }) as {
+      params: {
+        templateName: string
+      }
+    }
+
+    if (match && match.params && match.params.templateName) {
```
121watts, line 61:
```suggestion
    if (match?.params?.templateName) {
```
We can now use the new chaining operator hotness (if you'd like)

ui/src/templates/containers/CommunityTemplatesIndex.tsx
```diff
@@ -41,38 +41,26 @@ interface StateProps {
...
-          this.props.org.id,
-          getTemplateNameFromGithubUrl(currentTemplate)
-        )
+    const match = matchPath(this.props.location.pathname, {
+      path: communityTemplatesImportPath,
+    }) as {
+      params: {
+        templateName: string
+      }
+    }
```
121watts, line 59:
```suggestion
    type Params = {params: {templateName: string}}
    const match = matchPath<Params>(this.props.location.pathname, {
      path: communityTemplatesImportPath,
    })
```
`matchPath` takes a generic so you dont need to coerce the type

ui/src/templates/containers/CommunityTemplatesIndex.tsx
```diff
@@ -182,11 +155,7 @@ const mstp = (state: AppState): StateProps => {
   }
 }
 
-const mdtp = {
-  setCommunityTemplateToInstall,
-}
-
-export const CommunityTemplatesIndex = connect<StateProps, DispatchProps, {}>(
+export const CommunityTemplatesIndex = connect<StateProps, {}, {}>(
```
121watts, line 158:
```suggestion
export const CommunityTemplatesIndex = connect<StateProps>(
```
can just remove those empty types entirely!

ui/src/templates/containers/CommunityTemplatesIndex.tsx
```diff
@@ -182,11 +155,7 @@ const mstp = (state: AppState): StateProps => {
...
   mstp,
-  mdtp
+  null
```
121watts, line 160:
```suggestion
```
don't need to explicitly pass null either.  woot!

ui/src/templates/components/CommunityTemplateImportOverlay.tsx
```diff
@@ -62,6 +75,21 @@ class UnconnectedTemplateImportOverlay extends PureComponent<Props> {
...
+  private reviewTemplateResources = async (orgID, templateName) => {
+    const yamlLocation = `${getRawUrlFromGithub(
+      getGithubUrlFromTemplateName(templateName)
+    )}/${templateName}.yml`
+
+    try {
+      const summary = await reviewTemplate(orgID, yamlLocation)
+
+      this.props.setCommunityTemplateToInstall(summary)
+      return summary
```
121watts, line 87: do we need this return if its only used in `componentDidMount`?

ui/src/templates/containers/CommunityTemplatesIndex.tsx
hoorayimhelping: thank you for doing typescript right for me 😆

ui/src/templates/containers/CommunityTemplatesIndex.tsx
hoorayimhelping: 🤦

ui/src/templates/containers/CommunityTemplatesIndex.tsx
hoorayimhelping: sweeeeeeeet! make the machines do the work for me.

ui/src/templates/components/CommunityTemplateImportOverlay.tsx
hoorayimhelping: @121watts  it's not currently used, but i always make a habit of returning promises from async functions to facilitate chaining. i thought it was spec for any function that resolves a promise to return it so it can be chained, but i can't find any references to that so i might be full of it...

PR #18854, opened 2020-07-02
chore: upgrade react router

121watts:
### What was done

Upgrades to...

`react-router` v3 => `react-router-dom` 5.x
`react-redux` v4 => v7.x
`react-router-redux` v4 => `connected-react-router` v6.x

There's massive room for improvement for how our routing is architected.  Now was not the time to do that.  For each deprecated api in `react-router` 3 i.e. `<IndexRoute>` there is an analogous (though different) api in `react-router-dom`.  I tried as much as possible to match those analogues and not change implementation.  

### Notes

I've changed routing to the `DashboardIndex` and it's corresponding child routes from `/dashboards` to `/dashboards-list`.  Routes for the `DashboardIndex` overlays `/dashboards/import` etc were conflicting with the `Dashoard` pages i.e. `/dashboard/:dashboardID` etc.  To prevent issues with bookmarked pages or routing from Quartz, I introduced a component that will forward anyone heading to `/dashboards` to `/dashboard-list`.  This will also make updating the quartz nav easier as we wont have to worry about coordinating the releases. 

### Why
I was, with increasing frequency, battling the outdated router.  I saw the writing on the wall and decided to bite the bullet now, before it became an emergency. 

### What we get
- MUCH better typing.  Have you ever wanted to tell your routed component what query param props its receiving?  Well, now you can with `RouteComponentProps<{orgID: string; otherParam: string}>`
- No more need to use generics in `withRouter`.  The updated types just know what to do.
- `useDispatch` & `useSelector` AKA no more need to `connect`, `mstp`, and `mdtp`!
- a bunch of perf enhancements

PR #18873, opened 2020-07-07
chore: upgrade @types/react-redux

121watts:
### Why
In order for use to use all the fancy new hooks we get with the upgrades made in #18854, we also need to upgrade `react-redux` types.  There are some pretty great advantages with the new typings.  Specifically, the `connect` function is MUCH smarter and can infer types from `mstp` and `mdtp`.  For more info on the new patterns on typing connected components see:

https://react-redux.js.org/using-react-redux/static-typing

### Notes
The react redux docs recommend using hooks for "connecting" components to the store from here on out.

### DIFF
--- a/ui/src/notebooks/components/header/Submit.tsx
+++ b/ui/src/notebooks/components/header/Submit.tsx
@@ -118,6 +118,7 @@ export const Submit: FC = () => {
       submitButtonDisabled={!hasQueries}
       queryStatus={isLoading}
       onSubmit={submit}
+      onNotify={noop}
     />
   )
 }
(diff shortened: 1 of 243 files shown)


ui/src/notebooks/components/header/Submit.tsx
```diff
@@ -16,6 +17,8 @@ import {RemoteDataState} from 'src/types'
 const PREVIOUS_REGEXP = /__PREVIOUS_RESULT__/g
 const COMMENT_REMOVER = /(\/\*([\s\S]*?)\*\/)|(\/\/(.*)$)/gm
 
+const fakeNotify = notify
```
drdelambre, line 20: what does this do?

TCL735: It fakes a notify up the middle, qb drops back, waits, slings it downfield. TOUCHDOWN!

PR #18891, opened 2020-07-08
fix: allow custom set y-axis min to be 0

### LINKED ISSUE #18853: manual minimum y-axis value not rendering in dashboard
__Steps to reproduce:__

1. Create chronograf dashboard that includes any value that would cause the y-axis not to start at 0 (e.g. memory consumption, etc)
2. Edit the dashboard and set the y-axis manually to start at 0 (don't specify a max)
3. The y-axis doesn't change

__Expected behavior:__
The steps above would be nice so we can force the dashboard to start the y-axis at zero, but NOT have to specify a maximum for it. With metrics like network rx/tx, memory consumption, etc it's nice to let the tool visualize the changes at the top-end on our behalf while we set the minimum possible value.

__Actual behavior:__
Graph doesn't change even when setting manual y-axis min.

zero_axis

TCL735:
Closes #18853 

0 is a falsy value and was considered an unallowable custom set minimum value. We will allow 0 and guard against NaN and Infinity instead.

### DIFF
--- a/ui/src/shared/utils/useVisDomainSettings.test.ts
+++ b/ui/src/shared/utils/useVisDomainSettings.test.ts
@@ -14,9 +14,11 @@ describe('getValidRange', () => {
   const unixStart: number = 1573094811000
   const endTime: string = '2019-11-28T14:46:51Z'
   const unixEnd: number = 1574952411000
+
   it('should return null when no parameters are input', () => {
     expect(getValidRange(undefined, undefined)).toEqual(null)
   })
+
   it('should return null when no data is passed', () => {
     const timeRange: CustomTimeRange = {
       type: 'custom',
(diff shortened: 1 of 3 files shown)


ui/src/shared/utils/useVisDomainSettings.test.ts
```diff
@@ -72,24 +71,158 @@ describe('getRemainingRange', () => {
     }
     expect(getRemainingRange([], timeRange, [null, null])).toEqual(null)
   })
-  it("should return the min y-axis if it's set", () => {
+  describe('should return a valid number from the stored domain as the min y-axis', () => {
```
121watts, line 74: nit here but can we please have a newline between each `describe` and `it` block?

PR #18898, opened 2020-07-09
test: use experimental.chain in flux e2e tests to allow a single pass

adrian-thurston:
Force the writing of data and running of the test to happen sequentially. As
the results come out, collect them and report an error only if the diff results
are not empty.

### DIFF
--- a/query/stdlib/testing/end_to_end_test.go
+++ b/query/stdlib/testing/end_to_end_test.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/ast"
 	"github.com/influxdata/flux/execute"
 	"github.com/influxdata/flux/lang"
@@ -155,12 +154,15 @@ func makeTestPackage(file *ast.File) *ast.Package {
 var optionsSource = `
 import "testing"
 import c "csv"
+import "experimental"
 
 // Options bucket and org are defined dynamically per test
 
 option testing.loadStorage = (csv) => {
-	c.from(csv: csv) |> to(bucket: bucket, org: org)
-	return from(bucket: bucket)
+	return experimental.chain(
+		first:  c.from(csv: csv) |> to(bucket: bucket, org: org),
+		second: from(bucket:bucket)
+	)
 }
 `
 var optionsAST *ast.File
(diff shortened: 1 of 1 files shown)


query/stdlib/testing/end_to_end_test.go
```diff
@@ -222,91 +222,42 @@ func testFlux(t testing.TB, l *launcher.TestLauncher, file *ast.File) {
...
+				t.Log( "table ", name, " not present in results" )
 			}
 		}
-		if err := r.Err(); err != nil {
-			t.Error(err)
+		if _, ok := results["diff"]; ok {
+			t.Error("diff table was not empty")
+			logFormatted("diff", results)
+			logFormatted("want", results)
+			logFormatted("got", results)
```
fchikwekwe, line 260: will this improve the way that test diffs are displayed? At present, they are a little confusing to read.

query/stdlib/testing/end_to_end_test.go
```diff
@@ -222,91 +222,42 @@ func testFlux(t testing.TB, l *launcher.TestLauncher, file *ast.File) {
...
 				}
-			}
-		}()
-		for r.More() {
-			v := r.Next()
-			err := execute.FormatResult(&out, v)
-			if err != nil {
-				t.Error(err)
+			} else {
+				t.Log( "table ", name, " not present in results" )
```
fchikwekwe, line 253: Hmm....what does it really mean if the `want`, `got` or `diff` is not present in the results? I'm trying to think of the clearest way to express that for troubleshooting purposes.

query/stdlib/testing/end_to_end_test.go
adrian-thurston: The way it ends up reading is like:

```
diff table was not empty
result "diff"
 <table-data>
result "want"
 <table-data>
result "got"
 <table-data>
```

It won't print any other tables it finds (previously you could see "to1" or other tables made by the test). Also, it will always print in that order. Let me know if there are any other improvements you had in mind!

query/stdlib/testing/end_to_end_test.go
adrian-thurston: I added this after, when I discovered that "got" could be empty and it would never end up in the formatted results. Then decided to just check them all. If we still have flakiness WRG to the data not getting committed to the DB before the test runs, maybe this happen more often.

fchikwekwe: Yeah, that's a good point. Another way to check for causes of flakiness.

PR #18905, opened 2020-07-09
feat(tsdb): SHOW TAG KEYS (no time) query using only TSI data.

brettbuddin:
Continuation of the work @stuartcarnie performed in https://github.com/influxdata/influxdb/pull/18894.

Fixes influxdata/idpe#7843

### DIFF
--- a/tsdb/tsm1/engine_measurement_notime_schema.go
+++ b/tsdb/tsm1/engine_measurement_notime_schema.go
@@ -252,7 +252,9 @@ func (e *Engine) fieldsNoTime(ctx context.Context, orgID, bucketID influxdb.ID, 
 // processing, a non-nil error will be returned along with statistics for the
 // already scanned data.
 func (e *Engine) MeasurementTagKeysNoTime(ctx context.Context, orgID, bucketID influxdb.ID, measurement string, predicate influxql.Expr) (cursors.StringIterator, error) {
-	predicate = AddMeasurementToExpr(measurement, predicate)
+	if measurement != "" {
+		predicate = AddMeasurementToExpr(measurement, predicate)
+	}
 	return e.tagKeysNoTime(ctx, orgID, bucketID, predicate)
 }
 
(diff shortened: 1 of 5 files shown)


tsdb/tsm1/engine_measurement_notime_schema.go
```diff
@@ -240,3 +241,91 @@ func (e *Engine) fieldsNoTime(ctx context.Context, orgID, bucketID influxdb.ID,
...
+
+			skey := sf.SeriesKey(elem.SeriesID)
+			if len(skey) == 0 {
+				continue
+			}
+
+			_, tags := seriesfile.ParseSeriesKeyInto(skey, tagsReuse[:0])
+			for _, t := range tags {
+				keyset[string(t.Key)] = struct{}{}
+			}
```
stuartcarnie, line 318: For the reads package, I created `KeyMerger` which you could use:

https://github.com/influxdata/influxdb/blob/0ae8bebd755cf4aa579b9c38e6ea4792dfdbe53a/storage/reads/keymerger.go#L11

Also, we can optimized`seriesfile.ParseSeriesKetInto` with a specialized implementation that just returns the keys:

```go
func parseSeriesKeys(data []byte, dst [][]byte) [][]byte {
	var name []byte
	_, data = seriesfile.ReadSeriesKeyLen(data)
	_, data = seriesfile.ReadSeriesKeyMeasurement(data)
	tagN, data := seriesfile.ReadSeriesKeyTagN(data)

	if cap(dst) < tagN {
		dst = make([][]byte, tagN)
	} else {
		dst = dst[:tagN]
	}

	for i := 0; i < tagN; i++ {
		dst[i], _, data = seriesfile.ReadSeriesKeyTag(data)
	}

	return dst
}
```

and then we can merge the result to km:

```go
keys = parseSeriesKeys(skey, keys)
km.MergeKeys(keys)
```

tsdb/tsm1/engine_measurement_notime_schema.go
```diff
@@ -240,3 +241,112 @@ func (e *Engine) fieldsNoTime(ctx context.Context, orgID, bucketID influxdb.ID,
...
+		}
+		return nil
+	}(); err != nil {
+		return cursors.NewStringSliceIterator(nil), err
+	}
+
+	for _, v := range km.Get() {
+		vals = append(vals, string(v))
+	}
+	sort.Strings(vals)
```
stuartcarnie, line 329: This should no longer be necessary, as `km.Get()` should return the keys already sorted

PR #18906, opened 2020-07-09
chore: connected community template stacks to redux

### LINKED ISSUE #18552: Landing Page: show list of installed templates
#### Background
We want to show the user what templates they've added.

#### Acceptance Criteria
- [x] Empty state: An encouraging message saying there are no templates
- [x] Non empty state: a list of installed templates ordered by date

hoorayimhelping:
Closes #18552

Cleans up the template install process:
* adds stacks (which represent installed templates) to templates store in redux

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/templates/api/index.ts
+++ b/ui/src/templates/api/index.ts
@@ -34,8 +34,9 @@ import {
   postDashboardsLabel as apiPostDashboardsLabel,
   postDashboardsCell as apiPostDashboardsCell,
   patchDashboardsCellsView as apiPatchDashboardsCellsView,
-  Error as PkgError,
   postTemplatesApply,
+  getStacks,
+  Error as PkgError,
   TemplateSummary,
 } from 'src/client'
 import {addDashboardDefaults} from 'src/schemas/dashboards'
@@ -46,6 +47,7 @@ import {
   DashboardTemplate,
   Dashboard,
   TemplateType,
+  InstalledStack,
   Cell,
   CellIncluded,
   LabelIncluded,
@@ -491,4 +493,14 @@ export const installTemplate = async (orgID: string, templateUrl: string) => {
   }
 
   return applyTemplates(params)
+}
+
+export const fetchStacks = async (orgID: string) => {
+  const resp = await getStacks({query: {orgID}})
+
+  if (resp.status >= 300) {
+    throw new Error((resp.data as PkgError).message)
+  }
+
+  return (resp.data as {stacks: InstalledStack[]}).stacks
 }
--- a/ui/src/templates/components/CommunityTemplatesActivityLog.tsx
+++ b/ui/src/templates/components/CommunityTemplatesActivityLog.tsx
@@ -1,16 +1,23 @@
 import React, {PureComponent} from 'react'
+import {connect, ConnectedProps} from 'react-redux'
 
+// Components
 import {Table} from '@influxdata/clockface'
 
-import {getStacks, Stack, TemplateKind} from 'src/client'
+// Redux
+import {fetchAndSetStacks} from 'src/templates/actions/thunks'
 
-interface Props {
+// Types
+import {AppState} from 'src/types'
+import {TemplateKind} from 'src/client'
+
+interface OwnProps {
   orgID: string
 }
 
-interface State {
-  stacks: Stack[]
-}
+type ReduxProps = ConnectedProps<typeof connector>
+
+type Props = OwnProps & ReduxProps
 
 interface Resource {
   apiVersion?: string
@@ -23,18 +30,10 @@ interface Resource {
   }[]
 }
 
-export class CommunityTemplatesActivityLog extends PureComponent<Props, State> {
-  state = {
-    stacks: [],
-  }
-
-  public async componentDidMount() {
+class CommunityTemplatesActivityLogUnconnected extends PureComponent<Props> {
+  public componentDidMount() {
     try {
-      // todo: move the stacks reponse into redux
-      const resp = await getStacks({query: {orgID: this.props.orgID}})
-
-      const stacks = (resp.data as any).stacks
-      this.setState({stacks})
+      this.props.fetchAndSetStacks(this.props.orgID)
     } catch (err) {
       console.error('error getting stacks', err)
     }
@@ -66,7 +65,7 @@ export class CommunityTemplatesActivityLog extends PureComponent<Props, State> {
   }
 
   render() {
-    if (!this.state.stacks.length) {
+    if (!this.props.stacks.length) {
       return <h4>You haven't installed any templates yet</h4>
     }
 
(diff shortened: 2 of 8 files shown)


ui/src/templates/api/index.ts
```diff
@@ -492,3 +493,13 @@ export const installTemplate = async (orgID: string, templateUrl: string) => {
...
 }
+
+export const fetchStacks = async (orgID: string) => {
+  const resp = await getStacks({query: {orgID}})
+
+  if (resp.status >= 300) {
+    throw new Error((resp.data as PkgError).message)
+  }
+
+  return (resp.data as any).stacks
```
ebb-tide, line 504: does this not get types from the client?

ui/src/templates/components/CommunityTemplatesActivityLog.tsx
```diff
@@ -1,16 +1,23 @@
 import React, {PureComponent} from 'react'
+import {connect, ConnectedProps} from 'react-redux'
 
+// Components
 import {Table} from '@influxdata/clockface'
 
-import {getStacks, Stack, TemplateKind} from 'src/client'
+// Redux Stoffs
```
ebb-tide, line 7: Actions ;)

ui/src/templates/components/CommunityTemplatesActivityLog.tsx
```diff
@@ -1,16 +1,23 @@
...
+import {fetchAndSetStacks} from 'src/templates/actions/thunks'
 
-interface Props {
+// TypeScript thingamabobs
```
ebb-tide, line 10: Types ;)

ui/src/templates/components/CommunityTemplatesActivityLog.tsx
```diff
@@ -105,3 +104,21 @@ export class CommunityTemplatesActivityLog extends PureComponent<Props, State> {
     )
   }
 }
+
+const mstp = (state: AppState) => {
+  return {
+    stacks: state.resources.templates.stacks.filter(
```
ebb-tide, line 110: this could be a getter that protects against things not being ready to go ! if you don't want to make a getter, let's protect agains templates stacks not being there?

ui/src/templates/api/index.ts
hoorayimhelping: @ebb-tide because of the way the generated code is structured, `stacks.data` isn't typed.

```ts
interface GetStacksOKResult {
  status: 200
  headers: Headers
  data: {
    stacks?: Stack[]
  }
}
```

I did change this though to

```ts
(resp.data as {stack: InstalledStack[]}).stacks

Comment (ebb-tide) [reply ui/src/templates/api/index.ts]: I wonder if that is that because it is an optional property in swagger? so the generated code thinks it could be there or not be there. If it is in fact guaranteed by the BE we can maybe fix the swagger to make it required?

File: ui/src/templates/components/CommunityTemplatesActivityLog.tsx (hunk 4 above)
Comment (hoorayimhelping) [reply ui/src/templates/components/CommunityTemplatesActivityLog.tsx:110]: @ebb-tide i think you might be misunderstanding this. this is just hiding stacks that have been deleted by the user. when you delete a stack, it currently deletes the resources associated with the stack, but leaves the stack itself. `stacks` will always be an array, it's initialized as an empty array by the reducer, and it comes back from the api as an array.

File: ui/src/templates/components/CommunityTemplatesActivityLog.tsx [hunk 5]
```diff
@@ -23,18 +30,10 @@ interface Resource {
...
-  public async componentDidMount() {
+class CommunityTemplatesActivityLogUnconnected extends PureComponent<Props> {
+  public componentDidMount() {
     try {
-      // todo: move the stacks reponse into redux
-      const resp = await getStacks({query: {orgID: this.props.orgID}})
-
-      const stacks = (resp.data as any).stacks
-      this.setState({stacks})
+      this.props.fetchAndSetStacks(this.props.orgID)
```
ebb-tide, line 36: is there a reason we aren't using GetResources to fetch these?

ui/src/templates/components/CommunityTemplatesActivityLog.tsx
ebb-tide: we've used getters to filter before.. but getter was just a suggestion :) didn't realize stacks was initialized in reducer, thought it was only defined after fetching. nevermind!

ui/src/templates/api/index.ts
```diff
@@ -492,3 +494,13 @@ export const installTemplate = async (orgID: string, templateUrl: string) => {
...
 }
+
+export const fetchStacks = async (orgID: string) => {
+  const resp = await getStacks({query: {orgID}})
+
+  if (resp.status >= 300) {
+    throw new Error((resp.data as PkgError).message)
+  }
+
+  return (resp.data as {stacks: InstalledStack[]}).stacks
```
ebb-tide, line 505: but they're not `InstalledStack` right?, the BE returns `Stack` and then you have to apply properties we want on FE?

ui/src/templates/components/CommunityTemplatesActivityLog.tsx
hoorayimhelping: @ebb-tide yep, this is an MVP project that is dealing with a rapidly evolving API, while also trying to solve a chicken-and-egg product problem: this feature hasn't seen much use traditionally so we're not sure if we want to invest resources into it. But we're not sure if it hasn't seen much usage because we don't have good UX or because the feature itself isn't useful. So the focus of this is to learn what work we really need to do. It's not clear the ideal shape of the data or the how the UX of this should work, so while that's all in flight, I want to focus on iterating on the things are more visibly important to stakeholders (this is also a very high-visibility project, Evan has mentioned it in stand up this week).

My kind of long term vision if we decide to go with this is: making cleaning this up and making it conform to resources part of the work to replace old templates with community templates. Adding tests, tightening up the patterns, renaming confusing stuff, addressing all this tech debt I'm knowingly taking on, etc. But I don't want to do that work (which is super important and honestly my favorite kind of work) until we're sure it's worth the effort long term.

ui/src/templates/api/index.ts
hoorayimhelping: @ebb-tide correct! But I'm working with Johnny on this to get the API cleaned up around stuff like this.

ui/src/templates/components/CommunityTemplatesActivityLog.tsx
ebb-tide: 👍👍

ui/src/templates/api/index.ts
ebb-tide: And here, I just wanted to keep API honest- if it works it's cool with me

PR #18908, opened 2020-07-09
feat(VariableControlBar): load nested query variables onto the dashboard variable control bar

Asalem1:
This PR is part of a broader initiative to reduce the number of issues stemming from variable hydration. The change here is pretty straightforward. If a query variable depends on another variable that is not directly being called within a query, we should still consider that variable as "used" and it should therefore be displayed in the variable control bar.

load-all-vars

### DIFF
--- a/ui/src/dashboards/components/variablesControlBar/VariablesControlBar.tsx
+++ b/ui/src/dashboards/components/variablesControlBar/VariablesControlBar.tsx
@@ -17,11 +17,7 @@ import {
   getVariables,
   getDashboardVariablesStatus,
 } from 'src/variables/selectors'
-import {
-  filterUnusedVars,
-  getAllUsedVars,
-  createdUsedVarsCache,
-} from 'src/shared/utils/filterUnusedVars'
+import {filterUnusedVars} from 'src/shared/utils/filterUnusedVars'
 
 // Actions
 import {moveVariable} from 'src/variables/actions/thunks'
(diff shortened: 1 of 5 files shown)


ui/src/variables/actions/thunks.ts
```diff
@@ -431,6 +435,8 @@ export const selectValue = (variableID: string, selected: string) => async (
   // Validate that we can make this selection
   const vals = normalizeValues(variable)
   if (!vals.includes(selected)) {
+    // TODO: there is an issue that's causing non-state set values to
```
Asalem1, line 434: This is more of a note-to-self so that we don't lose context on one of the main issues with variable hydration. More specifically, when querying variables from the state that haven't already been set, their values are null. This has an adverse affect on variables who's values are being set in query parameters before they are actually loaded by the app

ui/src/dashboards/components/variablesControlBar/VariablesControlBar.tsx
```diff
@@ -161,13 +165,19 @@ const mstp = (state: AppState): StateProps => {
     },
   } = state
 
+  const usedVariables = filterUnusedVars(
```
drdelambre, line 168: this doesn't feel like the right place to put this logic (in the view). why didn't you instead make `filterUnusedVars` better?

Asalem1: I mostly didn't do that because `filterUnusedVars` is used in another place and I didn't want to create an bugs in implementing a slight modification to the function

PR #18909, opened 2020-07-09
chore: add hook linter

### LINKED ISSUE #18760: Add linting rules for hooks

121watts:
Closes #18760 

### Why
After catching a couple of gnarly bugs that turned out to be from the misuses of our hooks.  @desa asked "can we not use a linter for this?".  Indeed we can.

### Notes
This will most likely require a paradigm shift in how we use our hooks, particularly `useEffect`.  I highly recommend reading:

- https://reacttraining.com/blog/useEffect-is-not-the-new-componentDidMount/
- this issue on the react repository https://github.com/facebook/react/issues/14920
- and this request to disable the linting rule in the create-react-app repository (TLDR they didn't)

After my initial frustration in converting these over.  The linter actually forced me into writing smaller, cleaner code, and caught a few potentials for bugs.  I expect a learning period (much like the conversion to TypeScript) where we get used to this way of thinking and then start enjoying it.

While reading through this PR, you might think to yourself "hmmmm does `dispatch` or `history` change between renders?"  Good news!  They do not.  

Also, you might also see some `// eslint-disable-line react-hooks/exhaustive-deps`.  There were some cases where the juice of refactoring was not worth the squeeze.  I'm less concerned about correcting current code than preventing _future_ code from being miswritten and producing _new_ bugs.

### DIFF
--- a/ui/src/Logout.tsx
+++ b/ui/src/Logout.tsx
@@ -1,7 +1,7 @@
 // Libraries
 import {FC, useEffect} from 'react'
-import {connect, ConnectedProps} from 'react-redux'
-import {withRouter, RouteComponentProps} from 'react-router-dom'
+import {useDispatch} from 'react-redux'
+import {RouteComponentProps} from 'react-router-dom'
 
 // APIs
 import {postSignout} from 'src/client'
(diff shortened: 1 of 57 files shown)


ui/src/Logout.tsx
```diff
@@ -12,36 +12,31 @@ import {CLOUD, CLOUD_URL, CLOUD_LOGOUT_PATH} from 'src/shared/constants'
...
-  return null
-}
+  }, [dispatch, history])
 
-const mdtp = {
-  resetFeatureFlags: reset,
+  return null
 }
 
-const connector = connect(null, mdtp)
```
drdelambre, line 45: hott

ui/src/buckets/components/UpdateBucketOverlay.tsx
```diff
@@ -35,31 +36,34 @@ import {OwnBucket} from 'src/types'
...
 
   const [retentionSelection, setRetentionSelection] = useState(DEFAULT_SECONDS)
 
+  const handleClose = useCallback(() => {
+    history.push(`/orgs/${orgID}/load-data/buckets`)
+  }, [orgID, history])
+
   useEffect(() => {
     const fetchBucket = async () => {
       const resp = await api.getBucket({bucketID})
```
drdelambre, line 63: all this stuff is obviously in the wrong place {view layer instead of a data layer}, and while it's completely unrelated to the PR, it does beg the question; should we still use thunks if we have access to `useDispatch`?

drdelambre: maybe i answered my own question.. move it to the data layer, and then we dont have useDispatch

PR #18921, opened 2020-07-10
Variable protected names

### LINKED ISSUE #17457: Prevent creating variables that are flux keywords
Due to a  bug in the flux parser this will cause issues.

zoesteinkamp:
Closes #17457 
Screen Shot 2020-07-10 at 10 45 42 AM

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/variable.go
+++ b/variable.go
@@ -133,7 +133,7 @@ func (m *Variable) Valid() error {
 		fmt.Printf(inValidNames[x])
 
 		if m.Name == inValidNames[x] {
-			return fmt.Errorf("Protected variable name")
+			return fmt.Errorf("protected variable name")
 		}
 	}
 
--- a/variable.go
+++ b/variable.go
@@ -130,8 +130,10 @@ func (m *Variable) Valid() error {
 	inValidNames := [11]string{"and", "import", "not", "return", "option", "test", "empty", "in", "or", "package", "builtin"}
 
 	for x := 0; x < len(inValidNames); x++ {
+		fmt.Printf(inValidNames[x])
+
 		if m.Name == inValidNames[x] {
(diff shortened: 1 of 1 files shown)


variable.go
```diff
@@ -127,6 +127,16 @@ func (m *Variable) Valid() error {
 		return fmt.Errorf("invalid arguments type")
 	}
 
+	inValidNames := [11]string{"and", "import", "not", "return", "option", "test", "empty", "in", "or", "package", "builtin"}
+
+	for x := 0; x < len(inValidNames); x++ {
```
glinton, line 132: While this is valid, it's not preferred.
```suggestion
	for x := range inValidNames {
```

variable.go
```diff
@@ -127,6 +127,16 @@ func (m *Variable) Valid() error {
...
+		fmt.Printf(inValidNames[x])
```
glinton, line 133: remove this print prior to merge

variable.go
```diff
@@ -127,6 +127,16 @@ func (m *Variable) Valid() error {
...
+
+		if m.Name == inValidNames[x] {
+			return fmt.Errorf("protected variable name")
```
glinton, line 136: return the name of the variable for clarity perhaps?
```suggestion
			return fmt.Errorf("%q is a protected variable name", inValidNames[x])
```

zoesteinkamp: Added it!

variable.go
zoesteinkamp: Good catch, removed

variable.go
zoesteinkamp: got it, i changed it

PR #18928, opened 2020-07-13
feat(httpd): configurable http read timeout

### LINKED ISSUE #15410: http server does not set ReadTimeout
Version 1.7.x

https://github.com/influxdata/influxdb/blob/186289da56221340c57b24d1bec1f375e874ce9a/services/httpd/handler.go#L843-L855

On line 843, ReadFrom() may read data from underlying tcp connection but it does not set a deadline. This may be a problem if the connection is half-open(say, the client is powered down or the network is not reachable).  The read() would block for a long time, maybe a couple of days. 
If a concurrent-write-limit is set, such a connection would occupy a slot. Eventually it may fail to handle any write if half-open connections accumulate. 
A proposal is to set ReadTimeout for the http server so it can recover and close the connection. The ReadTimeout should be longer than enqueued-write-timeout.

foobar:
HTTP read timeout is useful to avoid half-open connections due to
bad network.

Closes #15410

### DIFF
--- a/services/httpd/config.go
+++ b/services/httpd/config.go
@@ -64,6 +64,7 @@ type Config struct {
 	MaxEnqueuedWriteLimit   int               `toml:"max-enqueued-write-limit"`
 	EnqueuedWriteTimeout    time.Duration     `toml:"enqueued-write-timeout"`
 	TLS                     *tls.Config       `toml:"-"`
+	ReadTimeout             time.Duration     `toml:"read-timeout"`
 }
 
 // NewConfig returns a new Config with default settings.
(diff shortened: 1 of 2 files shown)


lesam (changes requested):
I think this approach is fine but needs to be updated to fix conflicts.

Stepping back from the PR, which product are you hoping this is released in? We could potentially backport to 1.8, but it would seem more valuable to investigate a fix for 2.x .

Additionally this should really have some sort of test to prove it is working correctly and does not regress (ideally a unit test).

PR #18934, opened 2020-07-13
build(flux): update Flux to v0.73.0

rockstar:


query/stdlib/testing/testing.go
```diff
@@ -46,12 +46,15 @@ var FluxEndToEndSkipList = map[string]map[string]string{
...
-		"rowfn_with_import":   "unbounded test",
+		"group_nulls":           "unbounded test",
+		"integral":              "unbounded test",
+		"integral_columns":      "unbounded test",
+		"map":                   "unbounded test",
+		"join_missing_on_col":   "unbounded test",
+		"join_use_previous":     "unbounded test (https://github.com/influxdata/flux/issues/2996)",
+		"rowfn_with_import":     "unbounded test",
+		"merge_filter_flag_on":  "unbounded test (https://github.com/influxdata/flux/issues/2995)",
+		"merge_filter_flag_off": "unbounded test (https://github.com/influxdata/flux/issues/2994)",
```
fchikwekwe, line 57: Are these skipped tests?

rockstar: Yeah, we had three unbounded tests (though one being unbounded is an actual bug). The indent of the values makes it look bigger as a diff than it should be. Only three tests added.

fchikwekwe: Do we have an issue for the one that's a bug? Just curious.

rockstar: Yeah, for all three skipped tests, I filed a new issue (per skipped test protocol). This is the one that's a bug: https://github.com/influxdata/flux/issues/2996

fchikwekwe: thanks

PR #18936, opened 2020-07-13
feat(influxdb/flags): updated flags.yml with mergleFilterRule

yuktarakesh:

### DIFF
--- a/flags.yml
+++ b/flags.yml
@@ -134,3 +134,4 @@
   key: mergeFilterRule
   default: false
   contact: Query Team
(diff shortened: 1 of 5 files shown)


gavincabbage:
@yuktarakesh you'll want to run `make flags` as well, to generate the flag code from your new entry in `flags.yml`.

We're supposed to have a CI job to catch this, but this revealed it wasn't configured correctly so I've opened a PR to fix it. Glad I happened to see this, that was helpful!

PR #18937, opened 2020-07-13
fix(query/stdlib): exclude logic and errors in Flux function to()

wolffcm:
This makes a small update to logic for excluding tag columns in `to()`, especially when there is a `fieldFn` specified.

Also, I categorized the errors occurring `to.go` to be either invalid or internal.

Related to influxdata/flux#2837 (but that issue is officially closed in a separate PR)


jlapacik:
Is there an issue that this PR closes? Why is the exclude logic changing?

wolffcm:
@jlapacik Sorry, I could have been more helpful about what this is doing. A user noticed they were getting the error `cannot use reserved tag key "_field"` because the traversal of the `fieldFn` would overwrite the hard-coded column names to be excluded.

We need to exclude `_field` and `_measurement` because if the points writer sees them, it will produce an internal error. Note that the user can specify a different column to be the measurement, and if a column called `_measurement` still exists, we will try to use it as a tag key, which is not what the user would want.

jlapacik:
Ok I get it now. If a `fieldFn` was specified, then the excluded column map would get overwritten so as not to include `_field` anymore.

PR #18940, opened 2020-07-14
fix(ci): add lint-feature-flags to jobs lists

gavincabbage:
I noticed the `lint-feature-flags` check wasn't running on PRs. This aims to fix that by adding it to various jobs lists in the circle config. I added it wherever I saw other lint jobs.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -341,7 +341,6 @@ jobs:
           keys:
             - influxdb-gomod-{{ checksum "go.sum" }} # Matches based on go.sum checksum.
       - run: ./scripts/ci/lint/flags.bash
-      - skip_if_not_master
       - save_cache:
           name: Saving GOCACHE
           key: influxdb-gocache-{{ .Branch }}-{{ .Revision }}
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -341,7 +341,6 @@ jobs:
           keys:
             - influxdb-gomod-{{ checksum "go.sum" }} # Matches based on go.sum checksum.
       - run: ./scripts/ci/lint/flags.bash
-      - skip_if_not_master
       - save_cache:
```
mark-rushakoff, line 344: I don't think we want to save the module or build cache in this job, or in the `golint` job for that matter. It should only be enabled on the `gotest` job.

The build cache should include the test cache, and if the lint jobs save the cache first (which is likely since linting should happen much faster than running tests), then the cache we save will not necessarily contain a warm cache for tests, and subsequent runs will take more time to run the tests from a cold cache.

gavincabbage: Thanks for the explanation! I'v removed the caching for both lint jobs.

PR #18941, opened 2020-07-14
feat(storage): Add additional tracing to tsm1 schema APIs

stuartcarnie:
* Additional tracing was added to the tsm1 schema APIs in an effort to identify significant variance in latency.

* A new pkg/stopwatch was created to facilitate timing and tracing specific operations.

* Added a `meta` property to the Measurement schema APIs to permit sending arbitrary data to influence how the server handles the request. In this instance, we will use it to choose which Measurement schema strategy to use (NoTime)


e-dard:
@stuartcarnie looks good to me. Have you got any thoughts about how this will be sufficiently tested given it looks like it's hard to FF?

stuartcarnie:
@e-dard agree – it isn't really practical to feature flag changes that are specifically for increasing observability (tracing / logging), as they are almost always invasive. 

In reality we'd have to copy it all and create a new function called "MeasurementNamesWithMoreTracing".

The only concern in this code would be if the call to `StartSpanFromContext` returned a nil span:

```go
span, ctx := tracing.StartSpanFromContext(ctx)
defer span.Finish()
```

This API guarantees it does not, and will panic if `ctx` is nil. If `ctx` is nil, existing code would already panic.

tsdb/tsm1/engine_measurement_notime_schema.go
```diff
@@ -259,38 +296,56 @@ func (e *Engine) MeasurementTagKeysNoTime(ctx context.Context, orgID, bucketID i
...
+
 	if err := ValidateTagPredicate(predicate); err != nil {
 		return nil, err
 	}
 
 	orgBucket := tsdb.EncodeName(orgID, bucketID)
 
-	vals := make([]string, 0, 32)
+	var (
+		vals              = make([]string, 0, 32)
```
e-dard, line 309: Is this determined empirically or is there some compile-time known reason for it to be 32? Perhaps a comment or moving to a constant would clarify.

stuartcarnie: It was chosen over the default of 0 for an empty slice, which grows to 1, 2, 4, 8, 16 and eventually 32. The expectation is there will be typically more than 8 values, avoiding at least a number of allocations and copies.

https://play.golang.org/p/y-8AkpU-w-x

stuartcarnie: I will add a comment to clarify 👍

stuartcarnie:
This is no longer relevant for OSS

PR #18943, opened 2020-07-14
fix: add tenant URM respect for offset find options

lyondhill:
When requesting a set of urm's with an offset we need to skip offset number of urms


lyondhill:
@GeorgeMac wrote the same fix but added an additional test #18944 18944

PR #18949, opened 2020-07-14
fix(ui): give user all access token permission to read user

desa:
Users used to be given the permission `orgs/<orgID>/users`, but this
does not help when the user is trying to access the users api (as it is
not nested under orgs in the api). We have changed the
`orgs/<orgID>/users` to be `users/<userID>` so that the token has access
to the user that created it. This will allow users to access their
tokens via the api.

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/src/me/selectors/index.ts
+++ b/ui/src/me/selectors/index.ts
@@ -1,0 +1,20 @@
+// Libraries
+import {get} from 'lodash'
+
+// Types
+import {AppState} from 'src/types'
+import {MeState} from 'src/shared/reducers/me'
+
+// TODO(desa): is this reasonable or can I trust app state?
+const emptyMeState: MeState = {
+  id: '',
+  name: '',
+  links: {
+    self: '',
+    log: '',
+  },
+}
+
+export const getMe = (state: AppState): MeState => {
+  return get(state, 'me', emptyMeState)
+}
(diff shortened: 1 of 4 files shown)


desa:
Also how should I write an end to end test for this? I'd like to create an all access token and then use it that token.

ui/src/me/selectors/index.ts
```diff
@@ -0,0 +1,20 @@
+// Libraries
+import {get} from 'lodash'
+
+// Types
+import {AppState} from 'src/types'
+import {MeState} from 'src/shared/reducers/me'
+
+// TODO(desa): is this reasonable or can I trust app state?
+const emptyMeState: MeState = {
```
121watts, line 9: `AppState` is initialized in the [root node of the application] by the redux `Provider` (https://github.com/influxdata/influxdb/blob/master/ui/src/index.tsx#L74). The reducer which provides the `me` initial state is [located here](https://github.com/influxdata/influxdb/blob/master/ui/src/shared/reducers/me.ts#L14
).  We are guaranteed that we will (at a minimum) have this empty me state.  

So this `emptyMeState` is redundant and can be removed.

ui/src/me/selectors/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  id: '',
+  name: '',
+  links: {
+    self: '',
+    log: '',
+  },
+}
+
+export const getMe = (state: AppState): MeState => {
+  return get(state, 'me', emptyMeState)
```
121watts, line 19:
```suggestion
  return state.me
```

PR #18954, opened 2020-07-14
feat(query/stdlib): update flags.yml with mergleFilterRule and add mergeFilterRule to rules.go with corresponding unit tests.

yuktarakesh:
Updated flags.yml to include a rule for mergeFilter and ran make flags to generate code for the flag. Added unit tests to test flag off and on. Made sure that end_to_end_test.go ran and passed the changed made in flux. 

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/flags.yml
+++ b/flags.yml
@@ -136,9 +136,9 @@
   contact: Lyon Hill
   lifetime: temporary
 
-- name: Push Down Merged Filters
+- name: Merged Filters Rule
   description: Create one filter combining multiple single return statements
-  key: mergeFilterRule
+  key: mergeFiltersRule
   default: false
   contact: Query Team
   lifetime: temporary
--- a/flags.yml
+++ b/flags.yml
@@ -135,3 +135,10 @@
   default: false
   contact: Lyon Hill
   lifetime: temporary
+
+- name: Push Down Merged Filters
+  description: Create one filter combining multiple single return statements
+  key: mergeFilterRule
+  default: false
+  contact: Query Team
+  lifetime: temporary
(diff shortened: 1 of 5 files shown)


flags.yml
```diff
@@ -135,3 +135,10 @@
   default: false
   contact: Lyon Hill
   lifetime: temporary
+
+- name: Push Down Merged Filters
```
wolffcm, line 139: Most of the other flags in this file have the `name` field very similar to the `key` field, can you follow that convention here?

Also, this rule is not really a "push down" since it's not actually moving any query execution into the storage layer, it's just enabling the merging of filters that are executed in Flux. Perhaps a better name would be `Merge Filters Rule` or something like that.

go.mod
```diff
@@ -47,7 +47,7 @@ require (
 	github.com/hashicorp/vault/api v1.0.2
 	github.com/imdario/mergo v0.3.9 // indirect
 	github.com/influxdata/cron v0.0.0-20191203200038-ded12750aac6
-	github.com/influxdata/flux v0.73.0
+	github.com/influxdata/flux v0.73.1-0.20200714212633-e1dbd2426587
```
wolffcm, line 50: Just so you know, we can't merge this to master as it is now, since `influxdb` must point at a Flux release. So we'll need to wait until flux is released again, and update this `go.mod` file to use the new release.

query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -1167,3 +1170,20 @@ func asSchemaMutationProcedureSpec(spec plan.ProcedureSpec) *universe.SchemaMuta
 	}
 	return spec.(*universe.SchemaMutationProcedureSpec)
 }
+
+type MergeFilterRule struct{}
+
+func (MergeFilterRule) Name() string {
+	return universe.MergeFiltersRule{}.Name()
```
wolffcm, line 1177: The `flux` version of the rule is `MergeFiltersRule`, where "filters" is plural, can you rename this rule for consistency's sake?

PR #18956, opened 2020-07-14
feat: push down window offset to storage

jlapacik:
This PR

* Adds offset parameter to window read cursors
* Updates the planner rule to push down window offset calls
* Has storage expose window offset capability
* Adds unit tests
* Adds launcher tests


storage/reads/datatypes/storage_common.proto
```diff
@@ -252,5 +252,6 @@ message ReadWindowAggregateRequest {
   TimestampRange range = 2 [(gogoproto.nullable) = false];
   Predicate predicate = 3;
   int64 WindowEvery = 4;
+  int64 Offset = 6;
   repeated Aggregate aggregate = 5;
```
fchikwekwe, line 256: is there a reason that these are out of order of their enum(?) values? Just curious.

storage/flux/reader.go
```diff
@@ -619,6 +620,7 @@ func (wai *windowAggregateIterator) Do(f func(flux.Table) error) error {
 
 const (
 	CountKind = "count"
+	SumKind   = "sum"
```
fchikwekwe, line 623: oh, was sum not available before? Or are you just using it for the first time in this file?

query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -754,7 +747,8 @@ func isPushableWindow(windowSpec *universe.WindowProcedureSpec) bool {
 	// every and period must be equal
 	// every.months must be zero
 	// every.isNegative must be false
-	// offset: must be zero
+	// offset.months must be zero
+	// offset.isNegative must be false
 	// timeColumn: must be "_time"
 	// startColumn: must be "_start"
 	// stopColumn: must be "_stop"
```
fchikwekwe, line 754: These pushdown rule validators are well organized. Easy to follow.

storage/reads/datatypes/storage_common.proto
jlapacik: Not really. I was just going off the field ordering of the procedure spec.

storage/flux/reader.go
jlapacik: I used it for the first time in a test file, and storage can't import the universe package, so I needed to add it here.

storage/reads/modulo.go
```diff
@@ -0,0 +1,29 @@
...
+		r += modulus
+	}
+	return r
+}
+
+func WindowStart(t, every, offset int64) int64 {
+	mod := Modulo(t, every)
+	off := Modulo(offset, every)
+	beg := t - mod + off
+	if mod < off {
```
yzhang1991, line 15: I stopped here for a moment to think about what is this case about, I guess a comment would be helpful.

storage/reads/array_cursor.gen.go.tmpl
```diff
@@ -350,11 +349,6 @@ func (c *{{.name}}WindowFirstArrayCursor) Next() *cursors.{{.Name}}Array {
 	c.res.Timestamps = c.res.Timestamps[:0]
 	c.res.Values = c.res.Values[:0]
 
-	every := c.every
```
yzhang1991, line 353: why is this removed?

jlapacik: This was dead code so I decided to remove it and be explicit about what window first array cursors expect. They should not be passed `every = 0` or `every = math.MaxInt64`. In both cases a regular first array cursor should be (and is currently) used.

storage/reads/modulo.go
jlapacik: Added doc comments

Issue #18635, opened 2020-06-22
Implement Min and Max pushdown optimization


wolffcm:
Do everything except for the storage bits, which is a separate issue.

NOTE: Currently each planner rule uses its own feature flag for min/max. It would be simpler to consolidate those flags to be a single flag for min and a single flag for max. Meaning that each planner rule would consult the same flag for min and max respectively. 

DOD:
- End to end _test.flux tests that demonstrate correctness
- launcher test for OSS
- Pipeline for idpe
- Add test cases to verify that rule works right
- Turn on feature flag in tools, and then production clusters

wolffcm:
Deployment steps are:
- Turn on the feature flag for your user in tools. Test manually, show that it's correct, and if relevant, faster
- Once your satisfied, enable for the whole tools cluster. Let bake for an hour at least (look at queryd health dashboard)
- Deploy to production clusters, start with us-west-2, 25% of traffic. Bake for an hour
- Turn it on for all of us-west-2, let bake 1 hour
- Turn on everywhere

jsternberg:
All of the test cases for this are complete so we are going through the process of toggling the feature flags.

jsternberg:
This is on in tools and we're going to let it bake for a week before turning it on in production.

jsternberg:
This was intended to include `group |> min/max` so I'll write the tests for that and then close this.

jsternberg:
There's more extensive work to do with group so we will cover it as part of a separate issue.

PR #18957, opened 2020-07-14
feat(storage/reads): add min/max aggregate array cursors

### LINKED ISSUE #18636: Implement min and max aggregate in storage
- Add cursor implementation and related tests
- Add capability and related tests

wolffcm:
This PR adds new `min` and `max` aggregate cursors for the three numeric data types supported by storage.

I had to modify the aggregate cursor template a little bit to make this work right, since `min` and `max` are selectors and so must preserve the timestamps of their input rows (rather than just use the timestamp of the current window's upper bound).

Closes #18636

### DIFF
--- a/storage/reads/array_cursor.gen.go.tmpldata
+++ b/storage/reads/array_cursor.gen.go.tmpldata
@@ -5,19 +5,37 @@
 		"Type":"float64",
 		"Aggs": [
 			{
-		 		"Name":"Count",
-				"AccType":"int64",
-				"AccTypeName":"Integer",
-		 		"AccInit":"0",
-				"Accumulate":"acc++"
-		 	},
+				"Name":"Count",
+				"OutputTypeName":"Integer",
+				"AccDecls":"var acc int64 = 0",
+				"Accumulate":"acc++",
+				"AccEmit": "c.res.Timestamps[pos] = windowEnd; c.res.Values[pos] = acc",
+				"AccReset":"acc = 0"
+			},
 			{
-		 		"Name":"Sum",
-				"AccType":"float64",
-				"AccTypeName":"Float",
-		 		"AccInit":"0",
-				"Accumulate":"acc += a.Values[rowIdx]"
-		 	}
+				"Name":"Sum",
+				"OutputTypeName":"Float",
+				"AccDecls":"var acc float64 = 0",
+				"Accumulate":"acc += a.Values[rowIdx]",
+				"AccEmit":"c.res.Timestamps[pos] = windowEnd; c.res.Values[pos] = acc",
+				"AccReset":"acc = 0"
+			},
+			{
+				"Name":"Min",
+				"OutputTypeName":"Float",
+				"AccDecls":"var acc float64 = math.MaxFloat64; var tsAcc int64",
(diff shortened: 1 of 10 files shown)


storage/reads/array_cursor.gen.go.tmpldata
```diff
@@ -5,19 +5,37 @@
...
-		 	}
+				"Name":"Sum",
+				"OutputTypeName":"Float",
+				"AccDecls":"var acc float64 = 0",
+				"Accumulate":"acc += a.Values[rowIdx]",
+				"AccEmit":"c.res.Timestamps[pos] = windowEnd; c.res.Values[pos] = acc",
+				"AccReset":"acc = 0"
+			},
+			{
+				"Name":"Min",
```
jsternberg, line 24: Is there any overlap between first and min? This template seems very complex to me because it contains larger amounts of code.

A big concern of mine is that these code snippets appear to rely on local variables and be a way to insert code rather than as a data-driven way to instantiate templates.

storage/reads/array_cursor.gen.go.tmpldata
```diff
@@ -5,19 +5,37 @@
...
+				"OutputTypeName":"Float",
+				"AccDecls":"var acc float64 = math.MaxFloat64; var tsAcc int64",
+				"Accumulate":"if !windowHasPoints || a.Values[rowIdx] < acc { acc = a.Values[rowIdx]; tsAcc = a.Timestamps[rowIdx] }",
+				"AccEmit":"c.res.Timestamps[pos] = tsAcc; c.res.Values[pos] = acc",
+				"AccReset":"acc = math.MaxFloat64"
```
jsternberg, line 29: Does `windowHasPoints` need to be reset too? Also, does it need to be initialized?

storage/reads/array_cursor.gen.go.tmpldata
wolffcm: The execution pattern of first/last and min/max is pretty different, because for first/last, we do not need to read all of the blocks in a series. So Josh wrote custom cursors for them.

> A big concern of mine is that these code snippets appear to rely on local variables and be a way to insert code rather than as a data-driven way to instantiate templates.

I agree that it's not ideal. Maybe when we have generics in Go it will be easier. In the meantime, can you explain what you mean by "a data-driven way to instantiate templates"?

storage/reads/array_cursor.gen.go.tmpldata
wolffcm: No, `windowHasPoints` is defined in the template for all the aggregates. I understand that it's confusing that some variables are defined by the code snippets for a particular aggregate, and some are in the template.

storage/reads/array_cursor.gen.go.tmpldata
jsternberg: I mostly mean only using attributes that would be valid if generics in Go existed. Which would limit what could be in the template to values and types rather than code.

But it's not clear to me if that would make things more clear in this example so it's a very vague idea.

wolffcm:
> Is the plan to add end-to-end/launcher tests as part of the pushdown rule?

@jlapacik Yes, all that is part of the other issue: https://github.com/influxdata/influxdb/issues/18635

PR #18961, opened 2020-07-15
fix(tsi): clean up FileSet fields

foobar:
cleans up unused members from FileSet.

PR #18962, opened 2020-07-15
fix(tsi): retaining fileset should not copy pointer

foobar:
Current implementation of `retainFileSet` is not safe because it returns `p.fileSet` directly. It is supposed to be protected by mutex:

```
func (p *Partition) retainFileSet() *FileSet {
	fs := p.fileSet
	fs.Retain()
	return fs
}
```

The bug may lead to deadlock if the fileset is modified while there is compaction.

### DIFF
--- a/tsdb/index/tsi1/file_set.go
+++ b/tsdb/index/tsi1/file_set.go
@@ -30,6 +30,14 @@ func NewFileSet(levels []CompactionLevel, sfile *tsdb.SeriesFile, files []File) 
 	}, nil
 }
 
+func (fs *FileSet) Clone() *FileSet {
+	fs2 := &FileSet{
+		files: make([]File, len(fs.files)),
+	}
+	copy(fs2.files, fs.files)
+	return fs2
+}
+
 // bytes estimates the memory footprint of this FileSet, in bytes.
 func (fs *FileSet) bytes() int {
 	var b int
--- a/tsdb/index/tsi1/partition.go
+++ b/tsdb/index/tsi1/partition.go
@@ -452,7 +452,7 @@ func (p *Partition) RetainFileSet() (*FileSet, error) {
 }
 
 func (p *Partition) retainFileSet() *FileSet {
-	fs := p.fileSet
+	fs := p.fileSet.Clone()
 	fs.Retain()
 	return fs
 }


foobar:
The CI will success when https://github.com/influxdata/influxdb/pull/18961 is merged

tsdb/index/tsi1/partition.go
```diff
@@ -452,7 +452,9 @@ func (p *Partition) RetainFileSet() (*FileSet, error) {
 }
 
 func (p *Partition) retainFileSet() *FileSet {
```
ayang64, line 454: Do you think it might be better to make a `Clone()` method associated with `(*FileSet)` that does the copy?  That way the details of copying a `FileSet` remain near the implementation of `FileSet`.

foobar:
@dgnorton I suppose this fix needs to be ported back to 1.7/1.8

tsdb/index/tsi1/file_set.go
```diff
@@ -30,6 +30,14 @@ func NewFileSet(levels []CompactionLevel, sfile *tsdb.SeriesFile, files []File)
 	}, nil
 }
 
+func (fs *FileSet) Clone() *FileSet {
+	fs2 := &FileSet{
+		files: make([]File, len(fs.files)),
+	}
+	copy(fs2.files, fs.files)
```
ayang64, line 37: I'm not sure this is safe.  `File` is an interface and the value of the interface can contain pointers.

It looks like `LogFile` and `IndexFile` implement `File` and both of those types contain mutexes, pointers, etc.

so those pointers and the contents of the mutexes will be copied.

are we sure this is thread safe?  should be be copying mutex state?  is it okay to copy the pointers?

here's a small program that demonstrates the problems:

```package main

import (
  "log"
  "math/rand"
)

type File interface {
  Open() (int, error)
}

type F struct {
  m map[string]int
  n *int
}

func (f F) Open() (int, error) {
  log.Printf("%p:%p", &f, f.n)
  return 0, nil
}

func main() {
  var files []File
  for i := 0; i < 3; i++ {
    v := rand.Intn(100)
    files = append(files, F{n: &v, m: make(map[string]int)})
  }

  for _, v := range files {
    v.Open()
  }
  newFiles := make([]File, len(files))

  copy(newFiles, files)

  for _, v := range newFiles {
    v.Open()
  }
}
```

if you agree that this is a problem, how would you fix it?

what if each file also had a clone method that returns a safely copied version of itself -- without copying mutexes and maybe deeply copying pointer values?

i'm not sure but this seems like a potential source of bugs.

foobar:
rereview  the code, `FileSet.files` is actually immutable so the clone is not needed.

Issue #17993, opened 2020-05-07
Create a test case for groupResultSet


yzhang1991:
Create a test case exercising the groupResultSet. The test case should cover 4 scenarios: group by none/some columns, w/ or w/o aggregates.

We have similar unit tests for other result set implementations, we need to add one for the groupResultSet type for the 4 scenarios listed above.

DOD
- Tests exist that contain _real_ data, i.e. point data not just group key data.
- Tests exist for all 4 scenarios

PR #18968, opened 2020-07-15
fix: read group resultset

yzhang1991:


storage/flux/table.gen.go.tmpl
```diff
@@ -702,24 +703,87 @@ func (t *{{.name}}GroupTable) Do(f func(flux.ColReader) error) error {
...
-
+	if t.cur == nil {
+		// For group aggregates, we will try to get all the series and all table buffers within those series
+		// all at once and merge them into one row when this advance() function is first called.
+		// At the end of this process, t.advanceCursor() already returns false and t.cur becomes nil.
+		// But we still need to return true to indicate that there is data to be returned.
+		// The second time when we call this advance(), t.cur is already nil, so we directly return false.
 		return false
 	}
+	var a *cursors.{{.Name}}Array
```
rockstar, line 714: I realize this is a common pattern in our codebase, and I'm asking for a diversion from the pattern, but `a` and `l` and the like could have better names. `a` is an array, but an array of what?

storage/flux/table.gen.go.tmpl
```diff
@@ -702,24 +703,87 @@ func (t *{{.name}}GroupTable) Do(f func(flux.ColReader) error) error {
...
-	cr := t.allocateBuffer(l)
-	cr.cols[timeColIdx] = arrow.NewInt(a.Timestamps, t.alloc)
-	cr.cols[valueColIdx] = t.toArrowBuffer(a.Values)
+	// handle the group without aggregate case
+	if t.gc.Aggregate() == nil {
+		// Retrieve the buffer for the data to avoid allocating
+		// additional slices. If the buffer is still being used
+		// because the references were retained, then we will
+		// allocate a new buffer.
+		cr := t.allocateBuffer(l)
```
rockstar, line 733: Even though `cr` doesn't live that long, it could still have a better name. I believe this is short for "column row", so maybe `columnRow`

storage/flux/table.gen.go.tmpl
```diff
@@ -702,24 +703,87 @@ func (t *{{.name}}GroupTable) Do(f func(flux.ColReader) error) error {
...
+	// handle the group with aggregate case
+	var value {{.Type}}
+	// For group count, sum, min, and max, the timestamp here is always math.MaxInt64.
+	// their final result does not contain _time, so this timestamp value can be anything
+	// and it won't matter.
+	// For group first, we need to assign the initial value to math.MaxInt64 so
+	// we can find the row with the smallest timestamp.
+	// Do not worry about data with math.MaxInt64 as its real timestamp.
+	// In OSS we require a |> range() call in the query and a math.MaxInt64 timestamp
+	// cannot make it through.
```
rockstar, line 750: Yeah, this is the right kind of comment: we all have this common knowledge, but this code may need to be modified by someone without this common knowledge.

storage/reads/types.tmpldata
```diff
@@ -2,26 +2,97 @@
 	{
 		"Name":"Float",
 		"name":"float",
-		"Type":"float64"
+		"Type":"float64",
```
rockstar, line 5: This is interesting only in the fact that this had to be added now. What changed that meant we needed to implement this now instead of long ago?

yzhang1991: There are a bunch of `.tmpldata` files scattered in the OSS repo and a lot of them have duplicated content.
My intention is to eliminate those duplicates, but since this PR has been open for a long time and there are several conflicts that I do not want to resolve here. So this part of the change is reverted.

storage/flux/table.gen.go.tmpl
yzhang1991: Renamed to `colReader`

storage/flux/table.gen.go.tmpl
yzhang1991: I renamed it to `arr` to make it more obvious that this is an array. I did not include its type in the naming because It has its type right next to it in the definition, and its lifetime is fairly short (limited to this block).

yzhang1991:
> Can you update the rewrites for grouped aggregates as part of this fix?

I did it, and in doing that the `groupTable` needs to emit table with the correct schema (not including the tags that are not part of the group key and the `_time` column for the group aggregate case), so I made more changes to the code to make that happen.

storage/flux/table.go
```diff
@@ -196,6 +196,10 @@ func (t *table) readTags(tags models.Tags) {
 
 	for _, tag := range tags {
 		j := execute.ColIdx(string(tag.Key), t.cols)
+		// In the case of group aggregates, there might be tags that are excluded so you cannot find it.
+		if j < 0 {
+			continue
+		}
```
jlapacik, line 202: Can you explain what this is doing a bit more?

storage/flux/reader.go
```diff
@@ -506,9 +506,35 @@ func defaultGroupKeyForSeries(tags models.Tags, bnds execute.Bounds) flux.GroupK
...
+func determineTableColsForGroup(tagKeys [][]byte, typ flux.ColType, agg *datatypes.Aggregate, groupKey flux.GroupKey) ([]flux.ColMeta, [][]byte) {
+	var colSize int
+	if agg == nil || IsSelector(agg) {
+		// The group without aggregate or with selector (min, max, first, last) case:
+		// _start, _stop, _time, _value + tags
+		colSize += 4 + len(tagKeys)
+	} else {
+		// The group aggregate case:
+		// Only the group keys + _value are needed.
+		// Note that `groupKey` will contain _start, _stop, plus any group columns specified.
```
jlapacik, line 526: Why are _start and _stop always in the group key?

storage/flux/table.gen.go.tmpl
```diff
@@ -706,26 +707,93 @@ func (t *{{.name}}GroupTable) Do(f func(flux.ColReader) error) error {
...
+	var timestamp int64 = math.MaxInt64
+	if t.gc.Aggregate().Type == datatypes.AggregateTypeLast {
+		timestamp = math.MinInt64
+	}
+	for {
+		// note that for the group aggregate case, len here should always be 1
+		for i := 0; i < len; i++ {
+			switch t.gc.Aggregate().Type {
+			case datatypes.AggregateTypeCount:
+				{{if eq .Name "Integer"}}fallthrough{{else}}panic("unsupported for aggregate count: {{.Name}}"){{end}}
```
jlapacik, line 764: I don't think I follow the logic here. Every datatype should support count right?

yzhang1991: This is for re-aggregating `count`. The counts at this stage can only be integers.

storage/flux/reader.go
yzhang1991: I cannot speak to the design decision behind this, but this is how we currently implement it:
https://github.com/influxdata/influxdb/blob/master/storage/flux/reader.go#L539

jlapacik: Don't they have to be removed from the group key at some point?

yzhang1991: I guess not? Did you see where we do that in the code?

storage/flux/table.gen.go.tmpl
jlapacik: Got it. Thanks for clarifying.

storage/flux/reader.go
jlapacik: I'm looking for it, but I don't see it.

storage/flux/table.go
yzhang1991: In the case of group aggregate, tags that are not referenced in `group()` are not included in the result, but `readTags ()` still get a complete tag list. Here is just to skip the tags that should not present in the result.

storage/flux/table.gen.go.tmpl
```diff
@@ -706,26 +707,93 @@ func (t *{{.name}}GroupTable) Do(f func(flux.ColReader) error) error {
...
+	// Do not worry about data with math.MaxInt64 as its real timestamp.
+	// In OSS we require a |> range() call in the query and a math.MaxInt64 timestamp
+	// cannot make it through.
+	var timestamp int64 = math.MaxInt64
+	if t.gc.Aggregate().Type == datatypes.AggregateTypeLast {
+		timestamp = math.MinInt64
+	}
+	for {
+		// note that for the group aggregate case, len here should always be 1
+		for i := 0; i < len; i++ {
```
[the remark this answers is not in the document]
yzhang1991: Not really, just to be cautious.

yzhang1991:
Resolves #17993

PR #18972, opened 2020-07-16
chore: download tsi1 testdata with go generate

gavincabbage:
This PR removes the large binary tsi1 testdata files and adds a go generate directive to download the files from [a separate repository](https://github.com/influxdata/testdata). The go generate call is added to the Makefile and the directory is added to the gitignore to prevent it being recommitted. This PR is a pre-requisite to copying OSS code into the Cloud 2 codebase.

- [x] Rebased/mergeable
- [x] Tests pass


tsdb/tsi1/gen_test.go
```diff
@@ -0,0 +1,2 @@
+//go:generate sh -c "curl -L https://github.com/influxdata/testdata/raw/master/tsi1testdata.tar.gz | tar xz"
+package tsi1_test
```
mark-rushakoff, line 2: Do you think this could have something that has a very clear failure and remediation steps if `go generate` has not been run? Maybe something like:

```go
func init() {
  if _, err := os.Stat("./testdata/<something>"); err != nil {
    panic("You must run go generate to populate test data.")
  }
}
```

}
```

Comment (jacobmarble) [reply tsdb/tsi1/gen_test.go:2]: I second this comment. People will run `go test ./...` and ask "WTF?" which is a valid question without a clear step to remediate.

Comment (gavincabbage) [reply tsdb/tsi1/gen_test.go:2]: Agreed- great thinking @mark-rushakoff! Pushed that change just now

Comment (gavincabbage) [reply tsdb/tsi1/gen_test.go:2]:
```
ok  	github.com/influxdata/influxdb/v2/tsdb/migrate	0.674s
?   	github.com/influxdata/influxdb/v2/tsdb/migrate/internal	[no test files]
ok  	github.com/influxdata/influxdb/v2/tsdb/seriesfile	53.164s
panic: Run go generate to download testdata directory.

goroutine 1 [running]:
github.com/influxdata/influxdb/v2/tsdb/tsi1_test.init.0()
	/Users/gcabbage/Development/src/github.com/influxdata/influxdb/tsdb/tsi1/gen_test.go:8 +0x64
FAIL	github.com/influxdata/influxdb/v2/tsdb/tsi1	0.511s
ok  	github.com/influxdata/influxdb/v2/tsdb/tsm1	146.228s
?   	github.com/influxdata/influxdb/v2/tsdb/value	[no test files]
?   	github.com/influxdata/influxdb/v2/uuid	[no test files]
?   	github.com/influxdata/influxdb/v2/vault	[no test files]
ok  	github.com/influxdata/influxdb/v2/write	1.614s
?   	github.com/influxdata/influxdb/v2/zap	[no test files]
FAIL
gmake: *** [Makefile:140: test-go] Error 1
```

Comment (mark-rushakoff) [reply tsdb/tsi1/gen_test.go:2]: Approving with the panic, although now that I see the output, it would probably be easier to read as a `fmt.Println` and `os.Exit(1)` so that there isn't a stack trace cluttering the output.

File: tsdb/tsi1/gen_test.go [hunk 2]
```diff
@@ -0,0 +1,10 @@
+//go:generate sh -c "curl -L https://github.com/influxdata/testdata/raw/master/tsi1testdata.tar.gz | tar xz"
```
mark-rushakoff, line 1: It would probably be better to tag the testdata repo so that we don't have to coordinate changes to master over there with changes here. I would suggest a date-based tag, something like `2020.07.16.0` since there isn't necessarily any special meaning to any particular version.

PR #18974, opened 2020-07-16
fix(reduce-variable-hydration): filtering unused variables before hydrating them

Asalem1:
### Problem

Over-hydrating variables whenever any queries are run

### Solution

Filter out the unused variables before passing in the array of variables to update in order to reduce the graph dependencies that are ultimately returned and reduce the amount of queries that are performed in turn.

hydrate-less-vars


ui/src/variables/actions/thunks.ts
```diff
@@ -120,6 +124,20 @@ export const getVariables = () => async (
...
+    )
+  }
+  if (get(state, ['timeMachines', 'activeTimeMachineID']) === 'de') {
+    const de = getActiveTimeMachine(state)
+    return [de.view]
+  }
+  return []
+}
+
 // TODO: make this context aware
```
drdelambre, line 141: you todid this!

PR #18982, opened 2020-07-17
feat(mincore): Add page fault limiter

benbjohnson:
This commit adds `mincore.Limiter` which throttles page faults causedby mmap() data. It works by periodically calling `mincore()` to determine which pages are not resident in memory and using `rate.Limiter` to throttle accessing using a token bucket algorithm.

### DIFF
--- a/cmd/influxd/launcher/launcher.go
+++ b/cmd/influxd/launcher/launcher.go
@@ -73,6 +73,7 @@ import (
 	jaegerconfig "github.com/uber/jaeger-client-go/config"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
+	"golang.org/x/time/rate"
 )
 
 const (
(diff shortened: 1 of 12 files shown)


cmd/influxd/launcher/launcher.go
```diff
@@ -692,13 +701,24 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 		return err
 	}
 
+	// Enable storage layer page fault limiting if rate set above zero.
+	var pageFaultLimiter *rate.Limiter
+	if m.pageFaultRate > 0 {
+		pageFaultLimiter = rate.NewLimiter(rate.Limit(m.pageFaultRate), 1)
```
sebito91, line 707: How does the `1` map to the `mmap` instance here? Should this be a `const` so it's more descriptive? Could/should this change ever?

benbjohnson: It's part of the `golang.org/x/time/rate` package. It's the `burst` value for the limiter, although I couldn't get it to work correctly with any other value than `1`. https://godoc.org/golang.org/x/time/rate#NewLimiter

It seems like it lets in a higher number of values than what's available in the token bucket—i.e. if you have a `burst` of `10` and you pass in `10` immediately (before the token bucket has filled) then it allows it and doesn't wait at all. Instead we just call the underlying `Wait()` several times. It's something we can look at in the future if it seems to be a performance issue but I'm not especially worried about it right now.

benbjohnson:
Thanks. The `mincore` stuff is pretty specific. I'm not sure it's especially useful to be generally in `os`.

The funny thing about this PR is that I tried googling for how to call `mincore()` in Go and I had a small repo I had written 5 years ago (that I forgot about) that came up as the #2 result. 🤦‍♂️

Screen Shot 2020-07-17 at 9 38 24 AM

benbjohnson:
Merging this now. I'll add the other items as follow-on PRs.

Issue #18834, opened 2020-07-01
 Provide API's for accessing and modifying options in a Flux script in TypeScript


nathanielc:
We want to provide an API for mutating options in a Flux script.

It has been decided that the best way to do this is mutate the AST directly. As such we will build a TypeScript utility to easily modify and construct AST nodes and to add/remove options from an AST.

See https://github.com/influxdata/influxdb/issues/17386 for details and motivations.

nathanielc:
See https://github.com/influxdata/influxdb/issues/17386#issuecomment-652655507 for details on what the API should look like. 

I also got this helpful pointers from @desa 

Directory where things should go - https://github.com/influxdata/influxdb/tree/master/ui/src/shared/utils
Example - https://github.com/influxdata/influxdb/blob/master/ui/src/shared/utils/fromFlux.ts
Example Test - https://github.com/influxdata/influxdb/blob/master/ui/src/shared/utils/fromFlux.test.ts

to run your tests while developing run `yarn test:watch` in the ui directory of influxdb.

PR #18988, opened 2020-07-17
feat: typescript api for editing flux options

jlapacik:
Closes https://github.com/influxdata/influxdb/issues/18834.

### DIFF
--- a/ui/src/shared/utils/options.ts
+++ b/ui/src/shared/utils/options.ts
@@ -1,0 +1,140 @@
+import {Package, File, Expression, ObjectExpression} from 'src/types/ast'
+
+/*
+  Helper functions for editing option statements and object expressions
+  from the AST of a flux program.
+
+    option v = {a: 5, b: 6} // declaring an option in flux
+
+  After parsing a flux program via wasm and getting an AST representation,
+  it is sometimes useful/necessary to edit the options that are defined.
+  The functions in this file provide an interface for getting, setting,
+  and deleting options from the AST. The same interface is also provided
+  for the properties of an object expression.
+*/
+
+export function getProperty(r: ObjectExpression, label: string): Expression {
+  for (const prop of r.properties) {
+    if (
+      prop.key.type === 'Identifier' &&
+      'name' in prop.key &&
+      prop.key.name === label
+    ) {
+      return prop.value
+    }
+    if (
+      prop.key.type === 'StringLiteral' &&
+      'value' in prop.key &&
+      prop.key.value === label
+    ) {
+      return prop.value
+    }
+  }
+}
+
+export function setProperty(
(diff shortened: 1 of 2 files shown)


ui/src/shared/utils/options.ts
```diff
@@ -0,0 +1,140 @@
...
+      prop.key.type === 'Identifier' &&
+      'name' in prop.key &&
+      prop.key.name === label
+    ) {
+      return prop.value
+    }
+    if (
+      prop.key.type === 'StringLiteral' &&
+      'value' in prop.key &&
+      prop.key.value === label
```
fchikwekwe, line 28: Is this a TypeScript thing where you have to check if this is a string? In my version of the api, I think this check is implicit my comparing it to the label.

fchikwekwe: Or maybe I did this part wrong.

ui/src/shared/utils/options.ts
```diff
@@ -0,0 +1,140 @@
...
+    type: 'OptionStatement' as const,
+    assignment: {
+      type: 'VariableAssignment' as const,
+      id: {type: 'Identifier' as const, name: name},
+      init: expr,
+    },
+  })
+}
+
+export function deleteOption(f: File, name: string): void {
```
fchikwekwe, line 129: Probably just language nuance, but I return an error here if option cannot be deleted. I think it might be useful informationally to the caller.

fchikwekwe: yeah, looking over all the functions, returning an error seems to be a syntactic difference. Lemme know what you think about that.

ui/src/shared/utils/options.ts
jlapacik: Record labels can be both identifiers as well as string literals. The latter case is very rare but possible according to the AST.

ui/src/shared/utils/options.ts
jlapacik: I could have returned a boolean. It seems like an acceptable difference between the two APIs.

fchikwekwe: @nathanielc  was saying that it might not make a difference to the user whether the option existed or not.  Just that it was deleted if it did exist. I did think that it might be good to inform the user somehow, but I think it might make sense that they don't really need to know that the option wasn't there if they wanted to delete it.

ui/src/shared/utils/options.ts
fchikwekwe: Yeah, that's a good catch. I'm not handling this case atm.

fchikwekwe: I found that I only needed this check in setProperty. I'm not seeing the utility in getProperty.

PR #18991, opened 2020-07-20
fix(tsm): check write-ahead-log size

foobar:
current code checks segment file size for rolling agains the default const, instead of the configured size.


lesam:
@foobar can you provide a test case for this one that demonstrates the problem being solved? This looks reasonable but it would be better to have some proof that we are solving a currently broken case.

foobar:
> @foobar can you provide a test case for this one that demonstrates the problem being solved? This looks reasonable but it would be better to have some proof that we are solving a currently broken case.

@lesam done

lesam:
This looks fine - there's no user facing change as far as I can see though, since there doesn't appear to be any config option to set the WAL segment size. @foobar what's the motivation for the change / why do you want it?

lesam (approved):

PR #19002, opened 2020-07-21
fix(tsi): compact inactive log files on startup

foobar:
If there are multiple log files, inactive ones would not be compacted
and some memory space would be occupied.
This may happen when option max-index-log-file-size is adjusted, or the
program crashes during a tsi compaction.


foobar:
@timhallinflux @dgnorton  can you get this PR processed?

lesam:
I've opened a more complete overhaul of this code at https://github.com/influxdata/influxdb/pull/21943/files

PR #19005, opened 2020-07-21
 fix(eventing): Report duration as field value instead of tag value

ebb-tide:


ui/src/notebooks/components/header/TimeRangeDropdown.tsx
```diff
@@ -11,11 +11,13 @@ const TimeRangeDropdown: FC<TimeContextProps> = ({context, update}) => {
...
-        upper: range.upper as string,
-        lower: range.lower,
-      })
+      event(
+        'Time Range Updated',
+        {
+          type: range.type,
+        },
+        {upper: range.upper as string, lower: range.lower}
+      )
```
ebb-tide, line 20: @121watts is this ok with you?

121watts: 👍

PR #19006, opened 2020-07-21
fix: Variables are no longer stored in localStorage

### LINKED ISSUE #18483: UI: variables in UI show up after being deleted
image

__Steps to reproduce:__

1. create a variable via a template and grab the stack ID 
   ```INFLUX_OUTPUT_JSON=true influx apply -f pkger/testdata/variables.yml --org-id $ORG_ID``` 
   note: this outputs only the stackID
2. load up UI and go to variables list page under settings
3. delete the stack by the ID provided above `influx stacks rm --stack-id $STACK_ID --org-id $ORG_ID`
4. refresh page of UI, variables show up in view, but not in response from server

__Expected behavior:__
To show the response from server

__Actual behavior:__
Looks to provide some cached data

__Environment info:__

System: Darwin 18.7.0 x86_64
Browser: FF

zoesteinkamp:
Closes #18483

The variables were stored in localStorage so when we deleted them not via the UI, like for an example the CLI. It was not updating properly, now it will always call to the backend on page refresh

- [x] Rebased/mergeable
- [x] Tests pass


hoorayimhelping:
@Asalem1 it's good that you're considering unintended consequences, but Zoe and I have already examined alternatives for this. We see the effect of turning off localStorage caching any time we ask someone to check the page in an incognito window - they load the page with an empty localStorage cache just fine.

drdelambre:
this change will push this sev-3 into a sev-2
https://github.com/influxdata/idpe/issues/7700

there's a slew of bugs around variable, variable selection, and queries being run that contain variables, that are currently being mitigated / down graded to sev-3s and 4s by having the results of query variables cached in localStorage. It has something to do with the autoselect functionality wreaking havoc on a bunch of product flows.

we should probably hold off on this until that issue is resolved, if this is the only solution to the problem

Asalem1:
> @Asalem1 it's good that you're considering unintended consequences, but Zoe and I have already examined alternatives for this. We see the effect of turning off localStorage caching any time we ask someone to check the page in an incognito window - they load the page with an empty localStorage cache just fine.

I hear where you're coming from with how the localStorage is playing against us in this particular situation, and it's unfortunate that we don't have enough tests to effectively depend upon when implementing these types of changes, but since dashboards currently rely upon cached variables for the URL parameters, this PR will further break something that is currently a minor issue (as Boatie mentioned)

hoorayimhelping:
>by having the results of query variables cached in localStorage.

@drdelambre i'm confused how these can even become issues. where are we relying on localStorage for functionality? if it's caching, shouldn't an empty local storage be expressed as constant cache misses?

russorat:
I tested this and found that the variables are no longer stored in local storage, but for some reason, they still show up on the variables page after they have been deleted. you need to refresh the page and then they disappear. probably just a slight tweak to fix.

PR #19008, opened 2020-07-21
fix(storage/flux): do not wait for empty tables in the windower

jsternberg:
When splitting windows into their own individual tables, do not wait for
empty tables to be read. Empty tables may never be read so waiting for
them can result in a deadlock.

This was found while writing a test case for in progress work and so a
test case for this condition will follow in a future commit.

### DIFF
--- a/storage/flux/window.go
+++ b/storage/flux/window.go
@@ -67,6 +67,15 @@ func (w *windowTableSplitter) Do(f func(flux.Table) error) error {
 
 			// Rewrite the group key using the new time.
 			key := groupKeyForWindow(cr.Key(), startT, stopT)
+			if w.selector && values.IsNull(i) {
+				// Produce an empty table if the value is null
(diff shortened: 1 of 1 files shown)


jsternberg:
@jlapacik it's not really a mistake since the documentation for `Done()` and the tests for table tests check to make sure that `Done` doesn't need to be called. It might be good if we run the table tests for this though.

https://github.com/influxdata/flux/blob/529ab71b66ce695837db21a94be4adb8b9f69926/result.go#L43

jsternberg:
@jlapacik I agree. I had to make a change to this because apparently the previous code would introduce a memory leak and I can't add the table tests because of this bug: https://github.com/influxdata/flux/pull/3016.

I'd like to revisit this interface. It is too hard to use it correctly.

jlapacik:
I can still reproduce the deadlock.

PR #19010, opened 2020-07-21
feat(QueryCache): adds a query cache to Dashboards

### LINKED ISSUE #18791: Cache queries made in dashboard and VEO
writing to the cache
- [x] On dashboard page, when query is issued for cell, query results should be cached in redux store. cache is keyed by externs and query text. 
- [x] On dashboard page, when VEO is visible, when query is issued in VEO, query results should be cached in redux store.  queries cached in VEO should 'expire' after a few minutes. 
---
invalidating cache:
- [x] When user navigates away from a dashboard, the query cache should be invalidated/cleared so that when user navigates to the same dashboard again, cells are re-queried.
- [x] When a dashboard loads for the first time, all queries must query the database directly and not the cache
- [x] On dashboard Page, when viewing cells, the `reload` button should invalidate the cache and re-query for all cells. 
- [ ] In Autoreload(OSS)
- [ ] When user clicks submit. 

--- 
reading from cache:
- [x] On dashboard page, when user clicks on `configure` to edit cell query, the VEO should load and display query results from the cache. (this is already done) 
- [x] On dashboard page, when user clicks 'cancel' on VEO, background dashboard cells should attempt to query, but results should resolve from the cache, user should see dashboard as it was before they launched VEO even if they had selected a different time range or variable value in the VEO. 
- [x] On dashboard page, when user clicks 'save' on VEO without having ever submitting the query in the VEO (by clicking submit or changing variable value/time range), background dashboard cells should attempt to query, but results should resolve from the cache, user should see dashboard as it was before they launched VEO. 
- [x] On dashboard page, when VEO is visible, background dashboard cells should not attempt to query, even when time windows/variable values change.  (in PR)
- [x] On dashboard page, when user clicks 'save' on VEO, if user had resubmitted query in VEO, either by clicking submit, or changing query/variables/timeRange- all cells on dashboard should re-query. The edited cell should receive results from cache if it has not expired.

Asalem1:
Closes #18791 

### What

This PR introduces the concept of a QueryCache to dashboards in order to reduce the number of API requests generated by the UI when toggling between the Cell Configuration Overlay (known as VEO), and the cells on a dashboard. 

### How

Before each cells runs its respective query, it checks for a matching hashed version of that query in the local cache instance. If one is found, then the data from the cache is returned as promise and `runQuery` is never run. If one is not found (or if there is a mismatch in variables, or the data is stale), then the operation will continue to `runQuery`. 

Similarly, when the VEO is toggled open, the same operation will occur. 

When the cache is instantiated, a watch dog is triggered to clean up any stale data that is currently in the cache.

caching


ui/src/dashboards/components/DashboardPage.tsx
```diff
@@ -46,8 +46,12 @@ const dashRoute = `/${ORGS}/${ORG_ID}/${DASHBOARDS}/${DASHBOARD_ID}`
 
 @ErrorHandling
 class DashboardPage extends Component<Props> {
+  public componentDidmount() {
+    resetQueryCache()
```
ebb-tide, line 50: <3

ui/src/dashboards/components/DashboardPage.tsx
```diff
@@ -46,8 +46,12 @@ const dashRoute = `/${ORGS}/${ORG_ID}/${DASHBOARDS}/${DASHBOARD_ID}`
...
+  }
+
   public componentWillUnmount() {
-    this.props.resetCachedQueryResults()
+    resetQueryCache()
```
ebb-tide, line 54: <3

ui/src/dashboards/selectors/index.ts
```diff
@@ -41,6 +41,9 @@ export const sortDashboardByName = (dashboards: Dashboard[]): Dashboard[] =>
     a.name.toLowerCase() > b.name.toLowerCase() ? 1 : -1
   )
 
+export const isDashboardActive = (state: AppState): boolean =>
+  state.currentPage === 'dashboard'
```
drdelambre, line 45: can you drop a comment in here, or next to the usage, about why we have to track this?

ui/src/mockAppState.tsx
```diff
@@ -50,21 +49,11 @@ export const mockAppState = {
...
-          links: {
-            self: '/api/v2/variables/054b7476389f1000',
-            labels: '/api/v2/variables/054b7476389f1000/labels',
-            org: '/api/v2/orgs/674b23253171ee69',
-          },
-          status: RemoteDataState.NotStarted,
         },
         '05782ef09ddb8000': {
           id: '05782ef09ddb8000',
           orgID: '674b23253171ee69',
```
drdelambre, line 55: nit: we can probably get rid of the orgID too

ui/src/shared/apis/query.test.ts
```diff
@@ -0,0 +1,381 @@
+import {mocked} from 'ts-jest/utils'
+import {runQuery} from 'src/shared/apis/query'
+
+import {RunQuerySuccessResult} from 'src/shared/apis/query'
+import {AppState} from 'src/types'
+
+jest.mock('src/shared/apis/query')
+
+const mockState = ({
+  app: {
```
drdelambre, line 10: do we need this?

ui/src/shared/apis/query.test.ts
```diff
@@ -0,0 +1,381 @@
...
+        '05ba3253105a5000',
+        '05e6e4df2287b000',
+        '05e6e4fb0887b000',
+      ],
+    },
+  },
+} as unknown) as AppState
+
+describe('query', () => {
+  const filePath = 'src/shared/apis/query'
```
drdelambre, line 118: not using this anymore right?

ui/src/shared/apis/query.test.ts
Asalem1: We use this at some point in the process of creating of a query. The tests were complaining when this wasn't included

ui/src/shared/apis/query.test.ts
Asalem1: ✂️  you're right

ui/src/dashboards/selectors/index.ts
Asalem1: Definitely

ui/src/shared/apis/queryCache.ts
```diff
@@ -0,0 +1,164 @@
...
+}
+
+class QueryCache {
+  cache = {}
+
+  _cleanExpiredQueries = (): void => {
+    const now = Date.now()
+    for (const id in this.cache) {
+      if (this.cache[id].isCustomTime) {
+        continue
```
hoorayimhelping, line 48: i'm confused about this logic - why do we not expire the cache if it's a custom time?

ui/src/shared/apis/queryCache.ts
```diff
@@ -0,0 +1,164 @@
...
+    queryID: string,
+    variableID: string,
+    values: RunQueryResult,
+    isCustomTime: boolean = false
+  ): void => {
+    this.cache[queryID] = {
+      dateSet: Date.now(),
+      isCustomTime,
+      values,
+      variables: variableID,
```
hoorayimhelping, line 95: why is this called `variables` when it's a `variableID`?

ui/src/shared/apis/queryCache.ts
```diff
@@ -0,0 +1,164 @@
...
+    if (!this.cache[id]) {
+      return null
+    }
+    // query match with no existing variable match
+    if (this.cache[id].variables !== variableID) {
+      this.resetCacheByID(id)
+      return null
+    }
+    // query & variable match with an expired result
+    if (Date.now() - this.cache[id].dateSet > TIME_INVALIDATION) {
```
hoorayimhelping, line 67: do you think this expression
`Date.now() - this.cache[id].dateSet > TIME_INVLIADATION`

might be worth abstracting into a named function? `isCacheExpired` or something like that?

ui/src/shared/apis/queryCache.ts
```diff
@@ -0,0 +1,164 @@
...
+      this._cleanExpiredQueries()
+    }, TIME_INVALIDATION / 2)
+
+    this._cleanExpiredQueries()
+  }
+}
+
+const queryCache = new QueryCache()
+// Set an interval to check for expired data to invalidate
+queryCache.startWatchDog()
```
hoorayimhelping, line 110: so just by importing this file, we start a timer that runs every five seconds. in this PR i counted 5 imports of this file, not counting tests - does that mean five cache timers are going to be running every 5 seconds?

ui/src/shared/apis/queryCache.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+  }
+  const variableAssignments = variables
+    .map(v => asAssignment(v))
+    .filter(v => !!v)
+  // otherwise query & set results
+  const extern = buildVarsOption(variableAssignments)
+  const results = runQuery(orgID, query, extern, abortController)
+  results.promise = results.promise.then(res => {
+    // if the timeRange is non-relative (i.e. a custom timeRange || the query text has a set time range)
```
hoorayimhelping, line 155:
```suggestion
    // if the timeRange is non-relative (i.e. a custom timeRange or the query text has a set time range)
```
don't mix logical operators in comments. it makes all the `&` you have in your comments look like a bitwise operator.

ui/src/shared/apis/queryCache.ts
```diff
@@ -0,0 +1,164 @@
...
+      cancel: () => controller.abort(),
+    }
+  }
+  const variableAssignments = variables
+    .map(v => asAssignment(v))
+    .filter(v => !!v)
+  // otherwise query & set results
+  const extern = buildVarsOption(variableAssignments)
+  const results = runQuery(orgID, query, extern, abortController)
+  results.promise = results.promise.then(res => {
```
hoorayimhelping, line 154: does this need to conform to the same return type as other query methods? it doesn't look like `results` is ever used, just the promise from it. why not just return the promise?

ui/src/shared/apis/query.test.ts
```diff
@@ -0,0 +1,381 @@
...
+  describe('runQuery', () => {
+    beforeEach(() => {
+      const {resetQueryCache} = require('src/shared/apis/queryCache')
+      // NOTE: as long as you mock children like below, before importing your
+      // component by using a require().default pattern, this will reset your
+      // mocks between tests (alex)
```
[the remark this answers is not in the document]
Asalem1: Great question and point. The comment is irrelevant copy pasta that I forgot to remove. I think we might be able to import some of the files at the top, but the main one (getQueryCachedResults) encounters an issue when importing it at the top of the file, namely that it is dependent upon runQuery and so we need to mock out the module then import it

ui/src/shared/apis/query.test.ts
```diff
@@ -0,0 +1,381 @@
...
+        getRunQueryResults,
+        resetQueryCacheByQuery,
+      } = require('src/shared/apis/queryCache')
+      const orgID = 'orgID'
+      const queryText = '|> get some data fool'
+      const result = getRunQueryResults(orgID, queryText, mockState)
+      expect(runQuery).toHaveBeenCalledTimes(1)
+      result.promise.then(() => {
+        try {
+          // expect the cache to be cleared
```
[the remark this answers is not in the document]
Asalem1: More for clarification about what is happening at this point and less of a todo

ui/src/shared/apis/query.test.ts
```diff
@@ -0,0 +1,381 @@
...
+      expect(runQuery).toHaveBeenCalledTimes(1)
+      setTimeout(() => {
+        try {
+          getRunQueryResults(orgID, queryText, mockState)
+          expect(runQuery).toHaveBeenCalledTimes(2)
+          done()
+        } catch (error) {
+          done(error)
+        }
+        // wait 5100 seconds
```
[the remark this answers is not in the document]
Asalem1: Sounds good

ui/src/shared/apis/queryCache.ts
```diff
@@ -0,0 +1,164 @@
...
+  return {
+    [vari.name]: vari.selected || [],
+  }
+}
+
+// Hashing function found here:
+// https://jsperf.com/hashcodelordvlad
+// Through this thread:
+// https://stackoverflow.com/questions/7616461/generate-a-hash-from-string-in-javascript
+const hashCode = (queryText: string): string => {
```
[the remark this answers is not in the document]
Asalem1: Good point

ui/src/shared/apis/queryCache.ts
Asalem1: We do but we will want to treat that data differently since the data will never change. So instead of clearing out the data after 5s we may want to invalidate it after 10s

ui/src/shared/apis/queryCache.ts
Asalem1: The variableID is the hashed version of the variables used

ui/src/shared/apis/queryCache.ts
Asalem1: Since we’re dealing with relative times, I’m not sure how abstracting a method that handles time might affect the accuracy of this condition

ui/src/shared/apis/queryCache.ts
Asalem1: Fixed this by only importing it once at the top of the file

ui/src/shared/apis/queryCache.ts
Asalem1: I think I may be misunderstanding you, but we are returning the `results` here since it conforms specifically to the type that is expected to be returned in the function (a `CancelBox<RunQueryResult>`) which is an object with a `promise` and a `cancel` property.

ui/src/shared/apis/queryCache.ts
hoorayimhelping: that's really interesting! is there any way to indicate that `variables` are the hash of variables? documenting that in variable name would be very helpful to a reader!

ui/src/shared/apis/queryCache.ts
hoorayimhelping: the original issue still remains i think? any place `apis/queryCache` is imported, a new `queryCache` will be instantiated and `startWatchDog` will be called, which starts the timers.

here's the issue with this: a person importing this file may not know that there are timers running when they import it. that may not be a problem, but in general, timers can be a big source of UI jank and frame drops, so we want to make sure we're not firing a bunch of them off needlessly.

i don't have a great handle on the architecture of this and it may be necessary for this to function correctly, it's just something to keep in the back of your mind in this time of UI performance

ui/src/shared/apis/queryCache.ts
hoorayimhelping: yeah, i figured that's why it's returned like this, so there isn't much to do.

what i was saying that i couldn't find a single place where code acted like this:

```js
const results = getRunQueryResults(orgID, myQuery, appState)

// uses results directly
if (results.foo) {
  foo()
}
```
it was always like this:

```ts
const results = getRunQueryResults(orgID, myQuery, appState)

// only use results.promise
results.promise.then(response => doSomething(response))
// or
// still only using results.promise
Promise.all(results, r => r.promise)
```

ui/src/shared/apis/queryCache.ts
```diff
@@ -0,0 +1,164 @@
...
+    hash = (hash << 5) - hash + rawText.charCodeAt(i)
+    hash |= 0 // Convert to 32bit integer
+  }
+  return `${hash}`
+}
+
+class QueryCache {
+  cache = {}
+
+  private _cleanExpiredQueries = (): void => {
```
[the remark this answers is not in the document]
Asalem1: Totally hear that. Since this is being called in a different method, I thought it might give the reader an idea that the method is private without actually having to navigate to it.

ui/src/shared/apis/queryCache.ts
Asalem1: Great to know. Truthfully, if this condition isn't met at nanosecond precision doesn't necessarily bother me (although @drdelambre might have some thoughts on that) since a false positive will be a definite negative when the watchDog is next run (every 2.5s). You do bring up a good point, and it is something to bear in mind as we roll this out - especially if we find that a 5s caching threshold is too high.

ui/src/shared/apis/queryCache.ts
Asalem1: Definitely. How would you feel about `hashedVariables` as a property & parameter name rather than `variables` and `variableID` (respectively)?

ui/src/shared/apis/queryCache.ts
Asalem1: Definitely a great point - and something I dreamt up some half-baked solutions to last night. As you mentioned, it is worth considering this point if/when we decide to expand the cache into other parts of the app. Something I was thinking about was to try and have the watchDog be context aware, but I'm not sure if that's even a thing? @drdelambre any thoughts?

ui/src/shared/components/TimeSeries.tsx
```diff
@@ -219,6 +226,10 @@ class TimeSeries extends Component<Props, State> {
         const extern = buildVarsOption([...vars, ...windowVars])
 
         event('runQuery', {context: 'TimeSeries'})
+        if (isCurrentPageDashboard) {
+          return getRunQueryResults(orgID, text, appState)
```
ebb-tide, line 230: Total suggestion: naming this function `checkCacheAndRunQuery` ?

ui/src/shared/components/TimeSeries.tsx
```diff
@@ -355,7 +361,9 @@ const mstp = (state: AppState, props: OwnProps) => {
   ]
 
   return {
+    appState: state,
```
ebb-tide, line 364: hrrmmm. does this mean this page is rerendering each time any part of appstate changes?

ui/src/shared/components/TimeSeries.tsx
Asalem1: I'm gonna sound pedantic and say that might imply that we always `runQuery` which we don't in the case where the cache has a corresponding value. What about `getCachedResultsOrRunQuery`? A bit of a mouth full but definitely more explicit

ui/src/shared/components/TimeSeries.tsx
Asalem1: That's not something I had considered. I'm afraid I don't know enough about the mechanics of React to know the answer to this.

ebb-tide: since `getRunQueryResults` is only interested in using state to `getAllVariables(state)`, maybe we can just consume what is returned from `getAllVariables(state)`... ? or..... perhaps `getRunQueryResults` should be a thunk?

ui/src/shared/utils/filterUnusedVars.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    })
+    it('returns a variable when it exists in the query and does not depend on another query', () => {
+      const actual = filterUnusedVarsBasedOnQuery(variables, ['v.bucket'])
+      expect(actual).toEqual([bucketVariable])
+    })
+    it('returns a variable and its dependent variables when dependencies exist', () => {
+      const actual = filterUnusedVarsBasedOnQuery(variables, ['v.deployment'])
+      expect(actual).toEqual([deploymentVariable, bucketVariable])
+    })
+    it('returns a variable and its deeply nested dependent variables', () => {
```
ebb-tide, line 114: ❤️ I love reading test text, it's very soothing for the brain

ui/src/shared/components/TimeSeries.tsx
Asalem1: 🤔 that may work. I think I had shied away from making it a thunk originally since it was causing problems with API request + pending results from that API request being stored in redux, but now that we're not storing it in a reducer that might be a possibility? Let me run it by @drdelambre to see what he thinks

ui/src/shared/apis/queryCache.ts
```diff
@@ -0,0 +1,167 @@
...
+import {buildVarsOption} from 'src/variables/utils/buildVarsOption'
+import {filterUnusedVarsBasedOnQuery} from 'src/shared/utils/filterUnusedVars'
+import {runQuery} from 'src/shared/apis/query'
+
+// Types
+import {RunQueryResult} from 'src/shared/apis/query'
+import {CancelBox} from 'src/types/promises'
+import {AppState, Variable} from 'src/types'
+
+export const TIME_INVALIDATION = 5000
```
ebb-tide, line 15: maybe add a comment that this is ms?

ui/src/shared/apis/queryCache.ts
```diff
@@ -0,0 +1,167 @@
...
+      hashedVariables,
+      isCustomTime,
+      values,
+    }
+  }
+
+  startWatchDog = () => {
+    setInterval(() => {
+      this.cleanExpiredQueries()
+    }, TIME_INVALIDATION / 2)
```
ebb-tide, line 104: Since we refrain from using expired queries in the `getFromCache` function, I'm leaning towards this being unnecessary?

ui/src/shared/components/TimeSeries.tsx
ebb-tide: I love `getCachedResultsOrRunQuery`

ui/src/shared/components/TimeSeries.tsx
ebb-tide: 👍

ui/src/shared/utils/filterUnusedVars.ts
```diff
@@ -48,6 +48,19 @@ export const createdUsedVarsCache = (variables: Variable[]) => {
   }, {})
 }
 
+export const filterUnusedVarsBasedOnQuery = (
+  variables: Variable[],
+  queryTexts: string[]
+) => {
+  const varsInUse = variables.filter(variable =>
+    queryTexts.some(text => isInQuery(text, variable))
```
ebb-tide, line 56: ugh does this take into account nested variables? nested variables are the worst things!

ui/src/timeMachine/actions/queries.ts
```diff
@@ -227,6 +232,11 @@ export const executeQueries = (abortController?: AbortController) => async (
       const extern = buildVarsOption(variableAssignments)
 
       event('runQuery', {context: 'timeMachine'})
+      if (isCurrentPageDashboard(state)) {
+        // reset any existing matching query in the cache
+        resetQueryCacheByQuery(text)
```
ebb-tide, line 237: I don't get this- why are we resetting cache for text and then checking to see if the text is in cache?

Asalem1: That's a great question. I think the use case here is:

Cell A has a query `|> foo` with no variables
Cell B has a query `|> bar` with no variables

Cell B is configured and the query is updated to `|> foo`. We want to make sure that we don't end up pulling up a cached value. In order to do that, we first need to clear out whatever value current exists in the cache (for the new query, since it might not be "new")

ui/src/shared/utils/filterUnusedVars.ts
Asalem1: `isInQuery` does not, but that's why we added this :)

https://github.com/influxdata/influxdb/pull/19010/files/53904bedecb318e191d36c25903d703881ea491a#diff-8e45b03aeca264900ab0daa6f18aacd0R61

ui/src/shared/apis/queryCache.ts
Asalem1: Will do

ui/src/shared/apis/queryCache.ts
Asalem1: This is used to ensure that the browser doesn't crash due to a lack of memory. The only time we explicitly expire (and delete) from the cache (besides this) is when we submit a query for an existing cached query, or for a query that is matching but has different variables

ui/src/shared/components/TimeSeries.tsx
Asalem1: This worked! Just pushed up the changes. I tested to see if the cache broke in unexpected ways (it didn't), and I also did a few screen grabs for performance differences (there weren't). Great call!

PR #19020, opened 2020-07-22
feat: consolidate application state fetching

### LINKED ISSUE #18986: Refactor initial app state to parallel API requests
The  UI is architected in a pattern that does not issue API calls until previous API calls have been resolved. 
```
<Component#1>
    <Component#2 />
</Component#1>
```
and component#1 does not render component#2 until after all the API calls it makes are resolved. 
The biggest culprit here is the use of SpinnerContainer. which blocks the render of its children while it is loading. What happens is that we show one container's Spinner and then its child's spinner, and then its child's spinner.

drdelambre:
Closes #18986

instead of cascading blocking operations, lets grab them all at once!


ui/src/me/components/account/Settings.tsx
```diff
@@ -64,7 +64,7 @@ export class Settings extends PureComponent<StateProps, State> {
 }
 
 const mstp = ({me}: AppState) => ({
-  me,
+  me: me.resource.name,
```
ebb-tide, line 67: I think this is supposed to be returning username?

ui/src/me/components/account/Settings.tsx
```diff
@@ -64,7 +64,7 @@ export class Settings extends PureComponent<StateProps, State> {
 }
 
 const mstp = ({me}: AppState) => ({
```
ebb-tide, line 66:
```suggestion
const mstp = ({me}: AppState): StateProps => ({
```

ui/src/shared/containers/GetAppState.tsx
```diff
@@ -0,0 +1,145 @@
...
+      return {
+        flags: activeFlags(state),
+        flagStatus: state.flags.status || RemoteDataState.NotStarted,
+        orgStatus: state.resources.orgs.status || RemoteDataState.NotStarted,
+        meStatus: ignoreMe ? RemoteDataState.Done : state.me.status,
+      }
+    }
+  )
+
+  const loading = [orgStatus, meStatus, flagStatus].filter(
```
ebb-tide, line 50: I wonder if `some` is more appropriate here?

ui/src/shared/containers/GetAppState.tsx
```diff
@@ -0,0 +1,145 @@
...
+        dispatch(notify(sessionTimedOut()))
+      }
+
+      history.replace(`/signin${returnTo}`)
+    }
+  }, [dispatch, meStatus])
+
+  useEffect(() => {
+    const interval = setInterval(() => {
+      // NOTE: after the initial load, this dispatch will update the loading state, which will cause a rerender
```
ebb-tide, line 105: I didn't quite get this Will it rerender at each interval?

drdelambre: it will without this variable

ebb-tide: got it

drdelambre:
i'll look into that

PR #19021, opened 2020-07-22
fix(tenant): use rand org bucket id generator over snowflake in tenant service

### LINKED ISSUE #19017: Tenant service can generate harmful org and bucket IDs
While investigating flaky tests @adrian-thurston bisected a fault down to the following commit:
https://github.com/influxdata/influxdb/commit/d720661e77564105bfe371a392deff165ce49562

It introduced the new tenant service, which in turn introduced a new implementation for the bucket and organizations services.
He observed that the pattern in IDs generated for buckets had changed. Now the IDs being generated were very similar to one another.

This led us to note that the tenant service was using a different `influxdb.IDGenerator` implementation:
- Using `snowflake` now: https://github.com/influxdata/influxdb/blob/master/tenant/storage.go#L24
- Used to use `rand` implementation : https://github.com/influxdata/influxdb/blob/master/kv/service.go#L62

The `rand` implementation sanitized away some potentially harmful characters:

- `\`
- ` `
- `,`

Further investigation showed that this is problematic when the following sequences occur in either a bucket or org ID:

- `\ ` (`5c20`)
- `\"` (`5c22`)
- `\=` (`5c3b`)
- `\,` (`5c2c`)

Since organization and bucket IDs are concatenated to generate the measurement used in the storage engine. This causes an issue within TSI. Which presents itself as empty responses when querying data that was written to the bucket in question.
Data writes successfully, however, it cannot be retrieved again. As it has been stored at an unreachable location.

We need to go back to the original `rand.OrgBucketID` generator:
https://github.com/influxdata/influxdb/blob/master/rand/id.go#L34

GeorgeMac:
Closes #19017

This move tenant back to using the `rand` packages id generator when generating IDs for orgs and buckets.

Prior to this change, it was possible for harmful IDs to be generated for buckets and organizations. Which could leave buckets or even entire orgs unable to read their data (empty responses).

### DIFF
--- a/tenant/http_server_bucket_test.go
+++ b/tenant/http_server_bucket_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/go-chi/chi"
 	"github.com/influxdata/influxdb/v2"
 	ihttp "github.com/influxdata/influxdb/v2/http"
+	"github.com/influxdata/influxdb/v2/kv"
 	"github.com/influxdata/influxdb/v2/tenant"
 	itesting "github.com/influxdata/influxdb/v2/testing"
 	"go.uber.org/zap/zaptest"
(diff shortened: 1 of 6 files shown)


adrian-thurston:
Hey @alespour, just circling back on this with the intention to merge it and I noticed that the additional test doesn't pass in the context of influxdb. If you're unaware, we run all the tests against an influxdb launcher. The driver is in `influxdb/query/stdlib/testing/`. To make it run I think you just need to eliminate the now option and add a range liike
```
    |> range(start: 2020-01-01T00:00:00Z )
```

PR #19028, opened 2020-07-22
feat(storage/flux): add tests for min/max pushdowns

jsternberg:
#18635


jsternberg:
The flux e2e tests this references are part of this PR: influxdata/flux#3017.

PR #19029, opened 2020-07-22
perf: cancel dashboard page requests

### LINKED ISSUE #18640: Navigating away from a dashboard should cancel all pending queries for that dashboard
as a user, when i navigate away from a dashboard, all pending queries from that dashboard should be cancelled.

121watts:
Closes #18640

### The Problem
A bunch of slow and unresolved requests to `/query` would continue to fetch despite the user navigating away from the page.  Thus making the rest of their experience slow and 💩 

### The Solution
When the the Dashboard Page un-mounts cancel any requests to `/query` that are in flight.  

Kapture 2020-07-22 at 16 39 05

### DIFF
--- a/ui/src/dashboards/actions/thunks.ts
+++ b/ui/src/dashboards/actions/thunks.ts
@@ -324,18 +324,22 @@ export const deleteDashboard = (dashboardID: string, name: string) => async (
   }
 }
 
-export const getDashboard = (dashboardID: string) => async (
-  dispatch,
-  getState: GetState
-): Promise<void> => {
+export const getDashboard = (
+  dashboardID: string,
+  signal?: AbortSignal
+) => async (dispatch, getState: GetState): Promise<void> => {
   try {
     dispatch(creators.setDashboard(dashboardID, RemoteDataState.Loading))
 
     // Fetch the dashboard, views, and all variables a user has access to
     const [resp] = await Promise.all([
-      api.getDashboard({dashboardID, query: {include: 'properties'}}),
-      dispatch(getVariables()),
+      api.getDashboard({dashboardID, query: {include: 'properties'}}, {signal}),
+      dispatch(getVariables(signal)),
     ])
+
+    if (!resp) {
+      return
+    }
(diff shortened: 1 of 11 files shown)


TCL735:
:fire:

ui/src/shared/apis/singleQuery.ts
```diff
@@ -0,0 +1,91 @@
...
+        }),
+        cancel: ret.cancel,
+      }
+    },
+    resolve: (...args) => {
+      let curr
+
+      while (cached.length) {
+        curr = cached.pop()
+        curr.resolve.apply(curr, args) // eslint-disable-line prefer-spread
```
hoorayimhelping, line 62: curious about the `apply` here rather than just calling `curr.resolve(...args)` since it's using `curr` for context

ui/src/dashboards/actions/thunks.ts
```diff
@@ -363,6 +370,10 @@ export const getDashboard = (dashboardID: string) => async (
     dispatch(creators.setDashboard(dashboardID, RemoteDataState.Done, normDash))
     dispatch(updateTimeRangeFromQueryParams(dashboardID))
   } catch (error) {
+    if (error.name === 'AbortError') {
```
hoorayimhelping, line 373: we allso check for `'CancellationError'` in some handlers. not sure if we need that here, but wanted to bring it up

ui/src/shared/apis/singleQuery.ts
```diff
@@ -0,0 +1,91 @@
+import {File, CancelBox} from 'src/types'
+import {runQuery} from 'src/shared/apis/query'
+
+/*\
+
+    Declare this at the module level to create a CancelBox mutex
+    for runQuery. this will allow only one query to be run at a time, and
+    pass its results to any calls made to it while waiting for the api.
+    usage:
+      const mutex = RunQueryPromiseMutex()
```
Asalem1, line 10: Where have I seen this before?

PR #19044, opened 2020-07-23
fix: X axis not accepting value & Y axis only accepting value

### LINKED ISSUE #18884: cannot change the x-axis for line graph
As a user, if there is an option to select a different column for the x-axis, then i would like to be able to select that column.

Kapture 2020-07-08 at 9 20 59

zoesteinkamp:
Closes #18884

These changes will fix the X axis so they will accept _value. It will fix the Y axis so it will accept values other than _values. There are end to end test testing that both the X and the Y can select all values including time, start, stop, and value.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/cypress/e2e/explorer.test.ts
+++ b/ui/cypress/e2e/explorer.test.ts
@@ -739,6 +739,56 @@ describe('DataExplorer', () => {
         cy.getByTestID('form--element-error').should('not.exist')
       })
 
+      it('can set x-axis and y-axis values', () => {
+        // build the query to return data from beforeEach
+        cy.getByTestID(`selector-list m`).click()
+        cy.getByTestID('selector-list v').click()
+        cy.getByTestID(`selector-list tv1`).click()
+
+        cy.getByTestID('time-machine-submit-button').click()
+        cy.getByTestID('cog-cell--button').click()
+
+        // Check stop
+        cy.getByTestID('dropdown-x').click()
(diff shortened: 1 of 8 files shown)


zoesteinkamp:
> this looks solid. i would add unit tests for the reducer also.
We did try to add the unit test for the reducer, but we need to get the mocking done first, so that going to be addressed at a later date

ui/cypress/e2e/explorer.test.ts
```diff
@@ -739,6 +739,56 @@ describe('DataExplorer', () => {
...
+        //check start
+        cy.getByTestID('dropdown-y').click()
+        cy.getByTitle('_start').click()
+        cy.getByTestID('dropdown-y').contains('_start')
+
+        //check time
+        cy.getByTestID('dropdown-y').click()
+        cy.getByTitle('_time').click()
+        cy.getByTestID('dropdown-y').contains('_time')
+      })
```
hoorayimhelping, line 790: nice! is it possible to assert that clicking it does something? maybe something we can add when we pair up on unit tests

hoorayimhelping:
good call @121watts. we actually discussed it and looked into it and realized we'd have to do some work that might take a little while and hold this up, so we wrote it up in a ticket https://github.com/influxdata/influxdb/issues/19046 and are going to talk about it tomorrow when we pair up :+1:

PR #19072, opened 2020-07-24
feat(feature-flagged-query-cache): Added a feature flagged query cache

### LINKED ISSUE #18791: Cache queries made in dashboard and VEO

Asalem1:
Closes #18791

**NOTE**

This is a preliminary version of the query cache that has a feature flag wrapped around every place where it's being invoked so that we can work off of it. It resolves the windowPeriod issue that created the incident on Friday. This issue was resolved by specifically getting the windowPeriod variable from the query text, a pattern that we use here:

https://github.com/influxdata/influxdb/pull/19072/files#diff-4399ae26b73db6259e86e2dea60efd3dR224-R225

This was added here before running the query:

https://github.com/influxdata/influxdb/pull/19072/files?file-filters%5B%5D=.go&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-0be90b2f69a877941cc79979e2747aadR177-R180

### What
This PR introduces the concept of a QueryCache to dashboards in order to reduce the number of API requests generated by the UI when toggling between the Cell Configuration Overlay (known as VEO), and the cells on a dashboard.

### How
Before each cells runs its respective query, it checks for a matching hashed version of that query in the local cache instance. If one is found, then the data from the cache is returned as promise and runQuery is never run. If one is not found (or if there is a mismatch in variables, or the data is stale), then the operation will continue to runQuery.

Similarly, when the VEO is toggled open, the same operation will occur.

When the cache is instantiated, a watch dog is triggered to clean up any stale data that is currently in the cache.


ui/src/dashboards/components/DashboardPage.tsx
```diff
@@ -46,8 +47,16 @@ const dashRoute = `/${ORGS}/${ORG_ID}/${DASHBOARDS}/${DASHBOARD_ID}`
...
 @ErrorHandling
 class DashboardPage extends Component<Props> {
+  public componentDidmount() {
+    if (isFlagEnabled('queryCacheForDashboards')) {
+      resetQueryCache()
+    }
+  }
+
   public componentWillUnmount() {
-    this.props.resetCachedQueryResults()
```
drdelambre, line 50: doesn't this need to called if the flag is disabled?

ui/src/dashboards/components/DashboardPage.tsx
```diff
@@ -101,10 +110,6 @@ const mstp = (state: AppState) => {
   }
 }
 
-const mdtp = {
-  resetCachedQueryResults: resetCachedQueryResults,
```
drdelambre, line 105: and this added back

ui/src/dashboards/components/DashboardPage.tsx
Asalem1: Not since the file has been removed and we're no longer caching things. Even if the feature flag is implemented the previous caching system has been ripped out in this PR

ui/src/dashboards/components/DashboardPage.tsx
Asalem1: Not unless we want to integrate the old system back in before we end up removing it later?

ui/src/shared/apis/queryCache.ts
```diff
@@ -0,0 +1,194 @@
...
+  resetCache = (): void => {
+    this.cache = {}
+  }
+
+  setCacheByID = (
+    queryID: string,
+    hashedVariables: string,
+    values: RunQueryResult
+  ): void => {
+    event('query_cache_miss', {context: 'queryCache'})
```
drdelambre, line 115: what's the context for?

Asalem1: 🤷 I just went based off of what was going on in the other usages. Is there a doc somewhere on event reporting that we should stick to?

ui/src/shared/apis/queryCache.ts
```diff
@@ -0,0 +1,194 @@
...
+export const resetQueryCacheByQuery = (query: string): void => {
+  const queryID = `${hashCode(query)}`
+  queryCache.resetCacheByID(queryID)
+}
+
+export const getCachedResultsOrRunQuery = (
+  orgID: string,
+  query: string,
+  state: AppState
+): CancelBox<RunQueryResult> => {
```
drdelambre, line 149: throw an event up here too so that we can split questions like "how many queries did the cache save?" and it'd be `queries requested` vs `query cache set`

ui/src/shared/apis/queryCache.ts
drdelambre: we might want to add the `queryID` into that object so that we can separate parallelized queries

ui/src/shared/apis/queryCache.ts
Asalem1: i think this should be taken care of already in the sense that we already report events for `runQuery` which this inevitably calls

drdelambre: i dont understand this. feels like an gap between intent vs implementation. explicitly defining the event at the intent (as opposed to using a proxy event rooted in implementation to estimate the intent) helps keep a consistent history to the event, even through implementation changes. if we rewrite all of executeQueries, and forget to put in the eventing, or change the definition of the submodule's eventing structure, our dashboards for cache hits vs missus shouldn't change right? isn't the event you're planning on proxying already invalid because we don't use executeQueries for dashboards?

ui/src/dashboards/components/DashboardPage.tsx
drdelambre: that's what i'm trying to call attention to. doesn't that mean that we have interface changes being deployed that aren't captured / contained by the feature flag?

ui/src/shared/apis/queryCache.ts
Asalem1: You're right in that the TimeSeries component does not depend on the executeQueries thunk to run a query, it does, however, a duplicate event. I do see what you're saying though. I'll add it in

ui/src/shared/apis/queryCache.ts
```diff
@@ -0,0 +1,200 @@
...
+    }
+  }
+  const variableAssignments = variables
+    .map(v => asAssignment(v))
+    .filter(v => !!v)
+
+  const windowVars = getWindowVars(query, allVars)
+
+  // otherwise query & set results
+  const extern = buildVarsOption([...variableAssignments, ...windowVars])
```
drdelambre, line 180: flagging this for the thread's sake as the fix for the outage... generating a `windowPeriod` variable from the query and adding it to the AST partial for the query

ebb-tide: 👍

PR #19076, opened 2020-07-24
feat(cmd/influx): modify the query cli to use the http api

### LINKED ISSUE #19038: Update the influx query command to use the http endpoint
The `influx query` command should be updated to use the HTTP endpoint to submit a query as a flux string.

### Background

The `influx` command has both a `query` and a `repl` command. The `repl` command uses the flux repl which is very heavy weight and is the same as the one that's part of the flux binary. It also runs a lot of the query locally instead of on influxdb itself because of how the repl is implemented to query influxdb. This means that when we add push down optimizations, the remote calls need to be updated to push down those calls in the repl too.

In addition, there are other parts of the repl that make it perform differently than a flux query submitted to the HTTP endpoint.

At the moment, the `query` command uses the same backend as the repl. It will start a repl and then submit the entire file as input to the repl instead of running through the http endpoint. This causes most of the query to run locally within a flux engine native to the client rather than in the flux engine on the server.

This is undesirable for the CLI. When a user uses the CLI, they expect the query to be submitted to the server and don't understand why using `curl` is different from using the native `influx query` command and why the native CLI command sometimes doesn't work correctly or is slower than curl.

But fixing the REPL to submit all queries is difficult because of local variables that are common in flux.

For this reason, since the `flux repl` has all of the functionality of `influx repl`, we [removed the repl](https://github.com/influxdata/influxdb/pull/19040) from influx and someone can go download the flux repl instead and use that to experiment with flux and communicate with influxdb. We are then changing `influx query` to use the public HTTP API and to submit the file in its entirety.

jsternberg:
This modifies the query cli to remove the flux runtime dependency and
uses the public http api to execute the query. It then uses flux's csv
decoder to read the raw response.

This removes any dependency on the flux runtime and substantially
reduces the binary size and the startup execution time.

Fixes #19038.

### DIFF
--- a/cmd/influx/query.go
+++ b/cmd/influx/query.go
@@ -1,22 +1,22 @@
 package main
 
 import (
-	"context"
-	"crypto/tls"
+	"bytes"
+	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"os"
+	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/influxdata/flux"
-	"github.com/influxdata/flux/dependencies/filesystem"
-	"github.com/influxdata/flux/plan"
-	"github.com/influxdata/flux/repl"
-	"github.com/influxdata/flux/runtime"
-	_ "github.com/influxdata/flux/stdlib"
-	"github.com/influxdata/flux/stdlib/influxdata/influxdb"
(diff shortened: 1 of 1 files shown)


jsternberg:
I created https://github.com/influxdata/influxdb/issues/19160 to track creating new tests. I want to merge this as it substantially reduces the size of the binary, but I think we should have tests. But I think they are going to take more time than I would like to allocate for this during this sprint.

PR #19082, opened 2020-07-26
fix(endpoints): do not reset organization's secrets when an endpoint is created or updated

### LINKED ISSUE #19081: Update or creation of a notification endpoint deletes the secrets of all existing endpoints
The lastly updated/creating notification endpoint is working fine, other notification endpoints (with secrets) are then failing.

__Steps to reproduce:__
1. Create 1st notification endpoint (alert tab in UI) with a secret, for example HTTP notification endpoint with Bearer auth method
2. go run ./cmd/influx secret list => OK 1 secret 
060e753a36ab5000-token  28031e0d9c10d773
3. Create 2nd notification endpoint (alert tab in UI) with a secret, for example HTTP notification endpoint with Bearer auth method
4. go run ./cmd/influx secret list => **FAIL** there just 1 secret, with a new id (for the second endpoint)
060e761352ab5000-token  28031e0d9c10d773
5. Update 1st notification endpoint in UI with Bearer credentials
6. go run ./cmd/influx secret list => **FAIL**, there is just 1 secret for the teh 1st notification endpoint

__Expected behavior:__
There will be 2 secrets in Step 4 and 6 

__Actual behavior:__
Existing secrets are deleted/lost when creating/updating a notification endpoint.

__Environment info:__
master branch of influxdb as of 2020-07-26

sranka:
Closes #19081

This PR fixes the endpoint service so that it does not reset all organization secrets when a notification endpoint is created or updated. `SecretService.PutSecrets` call, which that "deletes" all existing secrets, is replaced by `PatchSecrets` herein. A new test was added.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/endpoints/service.go
+++ b/endpoints/service.go
@@ -56,7 +56,7 @@ func (s *Service) CreateNotificationEndpoint(ctx context.Context, edp influxdb.N
 		return nil
 	}
 
-	return s.secretSVC.PutSecrets(ctx, edp.GetOrgID(), secrets)
+	return s.secretSVC.PatchSecrets(ctx, edp.GetOrgID(), secrets)
 }
 
 // UpdateNotificationEndpoint updates a single notification endpoint.
@@ -79,7 +79,7 @@ func (s *Service) UpdateNotificationEndpoint(ctx context.Context, id influxdb.ID
 		return updatedEndpoint, nil
 	}
 
-	if err := s.secretSVC.PutSecrets(ctx, updatedEndpoint.GetOrgID(), secrets); err != nil {
+	if err := s.secretSVC.PatchSecrets(ctx, updatedEndpoint.GetOrgID(), secrets); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


sranka:
There is a good [test](https://github.com/influxdata/influxdb/blob/a2c042c8723192356308181b27bab8982ca7230a/testing/secret.go#L210) that explains the semantics of `PutSecrets`. It would IMHO even better to rename the SecretService's `PutSecrets` method to `ResetSecrets` to clearly indicate that all existing secrets are deleted and the new are set.  I can offer this refactoring upon reviewer's request, it might be opinionated.

I guess that every code that now uses PutSecrets is suspicious to do the same mistake. I found also https://github.com/influxdata/influxdb/blob/88cdf43db13326007fed0d775888687d6f4e5c24/pkger/service.go#L2424-L2426 
It should probably use `PatchSecrets`, @jsteenb2 ?

sranka:
@glinton .. a test was added herein to demonstrate the correct behavior

PR #19091, opened 2020-07-27
fix(tsdb): Fix mincore wait() out of bounds calls

benbjohnson:
This commit fixes a panic caused by `mincore()` checks.

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/tsdb/tsi1/measurement_block.go
+++ b/tsdb/tsi1/measurement_block.go
@@ -91,7 +91,7 @@ func (blk *MeasurementBlock) Elem(name []byte, limiter *mincore.Limiter) (e Meas
 		if offset > 0 {
 			// Parse into element.
 			var e MeasurementBlockElem
-			_ = wait(limiter, blk.hashData[offset:offset+1])
+			_ = wait(limiter, blk.data[offset:offset+1])
 			e.UnmarshalBinary(blk.data[offset:])
 
 			// Return if name match.
--- a/tsdb/tsi1/tag_block.go
+++ b/tsdb/tsi1/tag_block.go
@@ -248,8 +248,8 @@ func (itr *tagBlockValueIterator) Next() TagValueElem {
 
 	// Unmarshal next element & move data forward.
 	itr.e.unmarshal(itr.data)
-	itr.data = itr.data[itr.e.size:]
 	_ = wait(itr.limiter, itr.data[:itr.e.size])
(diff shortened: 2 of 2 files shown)


tsdb/tsi1/measurement_block.go
```diff
@@ -91,7 +91,7 @@ func (blk *MeasurementBlock) Elem(name []byte, limiter *mincore.Limiter) (e Meas
 		if offset > 0 {
 			// Parse into element.
 			var e MeasurementBlockElem
-			_ = wait(limiter, blk.hashData[offset:offset+1])
+			_ = wait(limiter, blk.data[offset:offset+1])
 			e.UnmarshalBinary(blk.data[offset:])
```
benbjohnson, line 95: This was the original panic. It's slicing the wrong byte slice. 🤦

tsdb/tsi1/tag_block.go
```diff
@@ -248,8 +248,8 @@ func (itr *tagBlockValueIterator) Next() TagValueElem {
 
 	// Unmarshal next element & move data forward.
 	itr.e.unmarshal(itr.data)
-	itr.data = itr.data[itr.e.size:]
 	_ = wait(itr.limiter, itr.data[:itr.e.size])
+	itr.data = itr.data[itr.e.size:]
```
benbjohnson, line 252: I found this while doing an audit of the other checks. It's performing the slice after moving the byte slice forward when it should happen before.

philjb:
Is there a test that can be added for this?

benbjohnson:
Not really. It generally should be caught with the regular test suite but these slipped through.

tsdb/tsi1/tag_block.go
philjb: I don't follow the change here: this reslice location shouldn't matter for `wait`. The reslice (`itr.data = itr.data[itr.e.size:]`) isn't performing an access to the slice's underlying array. `unmarshal` does access `itr.data`'s array. 

I'd put it `_ = wait(itr.limiter, itr.data[:itr.e.size])` before `unmarshal`. It'll make the array "hot" for the code using the Next() method.

benbjohnson: I believe there's still a bounds check on the slice even if it isn't accessing it. That's what happened with the other code that panicked.

PR #19103, opened 2020-07-27
feat(ui): Enhance resource creation experience when limits are reached

mavarius:
Change rate-limited create asset experience.

### OLD EXPERIENCE
Create asset button disabled when asset limit is reached for **Dashboards**, **Tasks**, & **Buckets**
image

Trying to create a **Check** fails with an error after going though the process of configuring it
image

Trying to create a **Rule** fails silently with only a console error
image

### NEW EXPERIENCE
Create asset button is always enabled
image

When user clicks to create asset they are presented with an upgrade overlay
image

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/alerting/components/AlertsColumn.tsx
+++ b/ui/src/alerting/components/AlertsColumn.tsx
@@ -3,7 +3,7 @@ import React, {FC, ReactChild, useState} from 'react'
 import {connect} from 'react-redux'
 
 // Types
-import {AppState, ResourceType} from 'src/types'
+import {AppState, ResourceType, ColumnTypes} from 'src/types'
 import {LimitStatus, MonitoringLimits} from 'src/cloud/actions/limits'
 
 // Components
@@ -32,11 +32,6 @@ import {
 // Constants
 import {CLOUD} from 'src/shared/constants'
 
-type ColumnTypes =
-  | ResourceType.NotificationRules
-  | ResourceType.NotificationEndpoints
-  | ResourceType.Checks
-
 interface OwnProps {
   type: ColumnTypes
   title: string
@@ -63,23 +58,18 @@ const AlertsColumnHeader: FC<OwnProps & StateProps> = ({
   const panelClassName = `alerting-index--column alerting-index--${formattedTitle}`
   const resourceName = title.substr(0, title.length - 1)
 
-  const assetCreateButton = (): JSX.Element => {
-    if (
-      CLOUD &&
-      limitStatus[type] === LimitStatus.EXCEEDED &&
-      type !== ResourceType.Checks
-    ) {
-      return (
-        <AssetLimitButton
-          color={ComponentColor.Secondary}
-          buttonText="Create"
-          resourceName={resourceName}
-        />
-      )
-    }
+  const isLimitExceeded =
+    CLOUD &&
+    limitStatus[type] === LimitStatus.EXCEEDED &&
+    type !== ResourceType.Checks
 
-    return createButton
-  }
+  const assetLimitButton = (
+    <AssetLimitButton
+      color={ComponentColor.Secondary}
+      buttonText="Create"
+      resourceName={resourceName}
+    />
+  )
 
   return (
     <Panel
@@ -97,7 +87,7 @@ const AlertsColumnHeader: FC<OwnProps & StateProps> = ({
             tooltipContents={questionMarkTooltipContents}
           />
         </FlexBox>
-        {assetCreateButton()}
+        {isLimitExceeded ? assetLimitButton : createButton}
       </Panel.Header>
       <div className="alerting-index--search">
         <Input
--- a/ui/src/alerting/components/AlertsColumn.tsx
+++ b/ui/src/alerting/components/AlertsColumn.tsx
@@ -4,7 +4,7 @@ import {connect} from 'react-redux'
 
 // Types
 import {AppState, ResourceType} from 'src/types'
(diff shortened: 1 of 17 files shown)


Asalem1:
Just wanna say how great an enhancement this is

ui/src/alerting/components/AlertsColumn.tsx
```diff
@@ -53,6 +61,25 @@ const AlertsColumnHeader: FC<OwnProps & StateProps> = ({
 
   const formattedTitle = title.toLowerCase().replace(' ', '-')
   const panelClassName = `alerting-index--column alerting-index--${formattedTitle}`
+  const resourceName = title.substr(0, title.length - 1)
+
+  const assetCreateButton = (): JSX.Element => {
+    if (
+      CLOUD &&
+      limitStatus[type] === LimitStatus.EXCEEDED &&
+      type !== ResourceType.Checks
```
ebb-tide, line 70: why not ResourceType.Checks?

ui/src/alerting/components/AlertsColumn.tsx
```diff
@@ -70,7 +97,7 @@ const AlertsColumnHeader: FC<OwnProps & StateProps> = ({
             tooltipContents={questionMarkTooltipContents}
           />
         </FlexBox>
-        {createButton}
+        {assetCreateButton()}
```
ebb-tide, line 100: There's actually no need to make this a function see above! :)

ebb-tide, line 100: could this just be a const instead of a function?

ui/src/cloud/actions/limits.ts
```diff
@@ -36,6 +36,10 @@ export enum LimitStatus {
   EXCEEDED = 'exceeded',
 }
 
+export type MonitoringLimits = {
+  [type: string]: LimitStatus
```
ebb-tide, line 40: is type one of `check` / `notificationrule` etc? might be nice to type it? nit.

ui/src/cloud/components/AssetLimitButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+// Components
+import {Button, ComponentColor, IconFont} from '@influxdata/clockface'
+
+// Actions
+import {showOverlay, dismissOverlay} from 'src/overlays/actions/overlays'
+
+interface OwnProps {
+  color?: ComponentColor
+  resourceName?: string
```
ebb-tide, line 13: it looks like resourceName might be a required prop?

ui/src/cloud/utils/limits.ts
```diff
@@ -34,40 +34,33 @@ export const extractTaskMax = (limits: LimitsState): number => {
...
+}
 export const extractRulesMax = (limits: LimitsState): number => {
   return get(limits, 'rules.maxAllowed') || Infinity
 }
 export const extractBlockedRules = (limits: LimitsState): string[] => {
   return get(limits, 'rules.blocked') || []
 }
 
+export const extractEndpointsLimits = (limits: LimitsState): LimitStatus => {
+  return get(limits, 'endpoints.limitStatus')
```
ebb-tide, line 55: should these return some defaults?

ui/src/shared/components/AddResourceDropdown.tsx
```diff
@@ -11,11 +12,21 @@ import {
...
 } from '@influxdata/clockface'
 
+// Actions
+import {showOverlay, dismissOverlay} from 'src/overlays/actions/overlays'
+
+// Types
+import {LimitStatus} from 'src/cloud/actions/limits'
+
+// Constants
+import {CLOUD} from '../constants'
```
ebb-tide, line 22: could this be non-relative address please?

ui/src/overlays/reducers/overlays.ts
```diff
@@ -13,6 +13,7 @@ export type OverlayID =
   | 'telegraf-output'
   | 'switch-organizations'
   | 'create-bucket'
+  | 'asset-limit'
 
 export interface OverlayParams {
   [key: string]: string
```
ebb-tide, line 19: I really don't like that these params are actually required for the overlay to load properly, but they are not properly typed. I know you didn't write this, and I don't think you need to fix this in this PR, but it makes me worry :(

ui/src/shared/components/AddResourceDropdown.tsx
```diff
@@ -11,11 +12,21 @@ import {
...
+
 interface OwnProps {
   onSelectNew: () => void
   onSelectImport: () => void
   onSelectTemplate?: () => void
   resourceName: string
+  limitStatus?: LimitStatus
```
ebb-tide, line 29: Is this prop really optional? couldn't see defaults anywhere.

ui/src/cloud/utils/limits.ts
mavarius: I don't think so. The defaults are set to `LimitStatus.OK` so there's always a value when these gets are called.

ui/src/overlays/reducers/overlays.ts
mavarius: Maybe we should open another tech debt issue for this.

ui/src/shared/components/AddResourceDropdown.tsx
mavarius: Ya, this needs to be optional because there are instances of this that are used without this prop. I added a default at the location it's used.

PR #19102, opened 2020-07-27
refactor: remove outdated routes from swagger

AlirieGray:
Remove unused routes from swagger API document

- [ ] [Well-formatted commit messages](https://www.conventionalcommits.org/en/v1.0.0-beta.3/)
- [ ] Rebased/mergeable
- [ ] Tests pass
- [ ] http/swagger.yml updated (if modified Go structs or API)

PR #19104, opened 2020-07-27
refactor: remove nested orgs labels route

AlirieGray:
This PR removes an outdated swagger definition for the orgs/orgID/labels route which was not being used, and removes the label handler from the org handler to reflect this change.

Continuation of https://github.com/influxdata/influxdb/pull/19102

- [x] Rebased/mergeable
- [x] Tests pass


http/swagger.yml
```diff
@@ -4110,74 +4110,6 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
-    post:
```
glinton, line 4113: what about the `GET` (line 4086-4112)?

AlirieGray: thank you!

stephanie-engel:
@AlirieGray @russorat 
I noticed that the nested org labels route was removed from OSS, but it still exists (and works) in Cloud 2. Is that to be expected, or should there be another PR to remove that code from Cloud 2?

PR #19105, opened 2020-07-28
chore: remove unreferenced code

foobar:
`Tags` is never referenced


lesam:
Thanks for the contribution! I have rebased it as https://github.com/influxdata/influxdb/pull/22930 , closing this one.

PR #19108, opened 2020-07-28
chore(query): use dbrp v2 service in flux influx.v1 dependencies

GeorgeMac:
Updating the DBRP Mapping Service to the v2 version within the flux `v1` packages dependencies.
The interface is almost identical. Save a few changes to Find and the obvious V2 suffix.

Moving to v2 will allow me to use the new DBRP mapping service client. Meaning the DBRP metadata can be retrieved from a remote service instead of directly from the kv store.

Update as this now adds it to more places:

- Query Dependencies
- Transpiler

### DIFF
--- a/cmd/chronograf-migrator/dashboard.go
+++ b/cmd/chronograf-migrator/dashboard.go
@@ -376,22 +376,29 @@ func convertQueries(qs []chronograf.DashboardQuery) []influxdb.DashboardQuery {
 	return ds
 }
 
-type dbrpMapper struct {
+type dbrpMapper struct{}
+
+// FindBy returns the dbrp mapping for the specified ID.
+func (d dbrpMapper) FindByID(ctx context.Context, orgID influxdb.ID, id influxdb.ID) (*influxdb.DBRPMappingV2, error) {
+	return nil, errors.New("mapping not found")
 }
 
-func (m dbrpMapper) FindBy(ctx context.Context, cluster string, db string, rp string) (*influxdb.DBRPMapping, error) {
-	return nil, errors.New("mapping not found")
-}
-func (m dbrpMapper) Find(ctx context.Context, filter influxdb.DBRPMappingFilter) (*influxdb.DBRPMapping, error) {
-	return nil, errors.New("mapping not found")
-}
-func (m dbrpMapper) FindMany(ctx context.Context, filter influxdb.DBRPMappingFilter, opt ...influxdb.FindOptions) ([]*influxdb.DBRPMapping, int, error) {
+// FindMany returns a list of dbrp mappings that match filter and the total count of matching dbrp mappings.
+func (d dbrpMapper) FindMany(ctx context.Context, dbrp influxdb.DBRPMappingFilterV2, opts ...influxdb.FindOptions) ([]*influxdb.DBRPMappingV2, int, error) {
 	return nil, 0, errors.New("mapping not found")
-
 }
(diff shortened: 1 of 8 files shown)


GeorgeMac:
Hey @fchikwekwe ! I just tagged flux team as I am touching some transpiler code. Would you mind throwing your eyes over this for me 🙏 

We've moved from dbrp mapping service v1 to v2 and some future work I am doing will benefit from this :)

GeorgeMac:
Hey @jsternberg I hope it is ok I tagged you on this. Github recommended it for the part of the code and Faith is out this week.

cmd/chronograf-migrator/dashboard.go
```diff
@@ -376,22 +376,29 @@ func convertQueries(qs []chronograf.DashboardQuery) []influxdb.DashboardQuery {
...
+func (d dbrpMapper) FindByID(ctx context.Context, orgID influxdb.ID, id influxdb.ID) (*influxdb.DBRPMappingV2, error) {
 	return nil, errors.New("mapping not found")
 }
-func (m dbrpMapper) FindMany(ctx context.Context, filter influxdb.DBRPMappingFilter, opt ...influxdb.FindOptions) ([]*influxdb.DBRPMapping, int, error) {
-	return nil, 0, errors.New("mapping not found")
 
+// FindMany returns a list of dbrp mappings that match filter and the total count of matching dbrp mappings.
+func (d dbrpMapper) FindMany(ctx context.Context, dbrp influxdb.DBRPMappingFilterV2, opts ...influxdb.FindOptions) ([]*influxdb.DBRPMappingV2, int, error) {
+	return nil, 0, errors.New("mapping not found")
 }
```
fchikwekwe, line 389: Are these just unimplemented for now?

query/influxql/transpiler.go
```diff
@@ -735,7 +734,7 @@ func (t *transpilerState) from(m *influxql.Measurement) (ast.Expression, error)
 								Name: "bucketID",
 							},
 							Value: &ast.StringLiteral{
-								Value: mapping.BucketID.String(),
+								Value: mappings[0].BucketID.String(),
```
fchikwekwe, line 737: I'm not familar with the dbrp mapping concept. Is it safe to assume in this context that the first object is the desired one?

cmd/chronograf-migrator/dashboard.go
GeorgeMac: I am not sure about these. The original implementation was unimplemented, so I just kept it consistent.
I assume that the chronograf migrator does not need a DBRP mapping service, but wants something to go in its place other than nil?

query/influxql/transpiler.go
GeorgeMac: I believe the original implementation `DBRPMappingService.Find` did the same thing. It just returned the first item it found for the search filter:
https://github.com/influxdata/influxdb/blob/b3dd6fead48904bd9329860720e382caf9d5690d/inmem/dbrp_mapping_service.go#L94

The new V2 versions now just supports doing a find many and picking _an answer_. I believe because of the composite key of DB and RP.

cmd/chronograf-migrator/dashboard.go
fchikwekwe: Yeah, makes sense to follow the original pattern.

query/influxql/transpiler.go
fchikwekwe: Ah, okay. I think I get that.

PR #19118, opened 2020-07-28
fix: removing default sort on tables

drdelambre:
Peek 2020-07-28 13-05

this addresses influxdata/ear#1644

the current default behavior is to implement a sort on the first column returned from the backend. this over-writes any sort that was explicitly defined in the query and not in the interface. i made it return data just like it came back from the server unless the user specifies a sort in the view options.

weary about the product level consequences, so i'm wrapping it in a flag to let russ click around before removing the flag


drdelambre:
i would need to do that if rolling out to users required config cat (for reverting live code or whatever), but right now i just need sign off from product that this is what they want

PR #19129, opened 2020-07-29
feat(autoAggregate): BE work for autoAggregate

ebb-tide:
This PR has the backend changes for the autoaggregation work to be incorporated in to the UI:
- Add the new `fillValues` property to the autoAggregate property of the builderConfig in dashboards and in the swagger definition
- Change check flux generation logic to check if a createEmpty exists on a call to aggregateWindow, before adding it
- Fix tests

### DIFF
--- a/http/check_test.go
+++ b/http/check_test.go
@@ -693,7 +693,7 @@ func TestService_handlePostCheck(t *testing.T) {
   "query": {
   	"builderConfig": {
     "aggregateWindow": {
-			"fillValues": false,
+      "fillValues": false,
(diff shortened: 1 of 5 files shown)


http/check_test.go
```diff
@@ -510,7 +512,8 @@ func TestService_handleGetCheck(t *testing.T) {
 		  "query": {
             "builderConfig": {
               "aggregateWindow": {
-                "period": ""
+								"fillValues": false,
```
glinton, line 515: nit: unify indentation

PR #19135, opened 2020-07-29
feat(notification/telegram): add telegram notification

sranka:
This PR takes back the telegram notification (#18218) that was reverted by #19088 + adds a new UI feature flag `notification-endpoint-telegram` that turns on/off telegram notifications (in 04b188a), off by default

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/notifications/endpoints/components/EndpointTypeDropdown.tsx
+++ b/ui/src/notifications/endpoints/components/EndpointTypeDropdown.tsx
@@ -32,6 +32,7 @@ const types: EndpointType[] = [
   {name: 'HTTP', type: 'http', id: 'http'},
   {name: 'Slack', type: 'slack', id: 'slack'},
   {name: 'Pagerduty', type: 'pagerduty', id: 'pagerduty'},
+  {name: 'Telegram', type: 'telegram', id: 'telegram'},
 ]
 
 const EndpointTypeDropdown: FC<Props> = ({
--- a/ui/src/notifications/endpoints/components/EndpointTypeDropdown.tsx
+++ b/ui/src/notifications/endpoints/components/EndpointTypeDropdown.tsx
@@ -10,11 +10,13 @@ import {extractBlockedEndpoints} from 'src/cloud/utils/limits'
 
 // Types
 import {NotificationEndpointType, AppState} from 'src/types'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 interface EndpointType {
   id: NotificationEndpointType
   type: NotificationEndpointType
   name: string
+  flag?: string // feature flag that enable/disable endpoint type
 }
 
 interface StateProps {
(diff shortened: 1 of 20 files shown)


glinton:
@sranka what is the status of this? have you looked into why the initial revert was made in order to avoid another? My assumption is that you will need to split this work into separate prs for the frontend and backend work, but i'm not certain

sranka:
@glinton it was reverted because of deployment problems, a new feature flag that turns off telegram UI is the solution, the name of the flag was suggested by @desa on July, 27th :
> "I’d name the feature-flag something like notification-endpoint-telegram
> [9:19 PM] next steps would be to re-open the telegram pr with the feature-flag"

ui/src/notifications/endpoints/components/EndpointTypeDropdown.tsx
```diff
@@ -40,7 +48,10 @@ const EndpointTypeDropdown: FC<Props> = ({
   blockedEndpoints,
 }) => {
   const items = types
-    .filter(({type}) => !blockedEndpoints.includes(type))
+    .filter(
+      ({type, flag}) =>
+        !blockedEndpoints.includes(type) && (!flag || isFlagEnabled(flag))
```
ebb-tide, line 53: let's create another function that calls `isFlagEnabled`, that gets passed the type of the endpoint. 
```suggestion
        !blockedEndpoints.includes(type) && (isFlaggedOn(type))
```

ui/src/notifications/endpoints/components/EndpointTypeDropdown.tsx
```diff
@@ -40,7 +48,10 @@ const EndpointTypeDropdown: FC<Props> = ({
   blockedEndpoints,
 }) => {
```
ebb-tide, line 49: const isFlaggedOn = (type:string) => {
    if (type === 'telegram'){
      return !flag || isFlagEnabled('notification-endpoint-telegram')
}
return true
}

ui/src/notifications/endpoints/components/EndpointTypeDropdown.tsx
```diff
@@ -10,11 +10,13 @@ import {extractBlockedEndpoints} from 'src/cloud/utils/limits'
 
 // Types
 import {NotificationEndpointType, AppState} from 'src/types'
+import {isFlagEnabled} from 'src/shared/utils/featureFlag'
 
 interface EndpointType {
   id: NotificationEndpointType
   type: NotificationEndpointType
   name: string
+  flag?: string // feature flag that enable/disable endpoint type
```
ebb-tide, line 19: it feels odd to me to add this to the type. I've made suggestions below for something that feels a little better. let me know if it makes sense!

sranka: @ebb-tide there are more new notifications types in the game (teams, sensu, discord, opsgenie) , it seems better to me to store the name of the flag in the endpoint descriptor type to avoid branching for every new type. Your proposed code will extend the `if type==='typename` branching in the code with every new type. It does not seem better to me, but I can get myself easily convinced that it is, since I don't know/own the whole concept herein. Can you pls comment on this?

sranka: OK, I did it your way herein.

PR #19137, opened 2020-07-29
fix: add sec headers

### LINKED ISSUE #7901: Could not add uint64 value with Influxdb client for go
Hello, Influxdb team!

I just tried to add uint64 value with official InfluxDB's Go client and got very painful bug. My values was added as "string" (as consequence all math operation on this data become impossible).

My friend @mrqwer88 found reason of this behaviour and I could quote it here from InfluxDB's code https://github.com/influxdata/influxdb/blob/91ee34b111f2034ec9456a6ce391b470d69e3216/models/points.go#L2013:
```
// TODO: 'uint' should be considered just as "dangerous" as a uint64,
 // perhaps the value should be checked and capped at MaxInt64? We could
 // then include uint64 as an accepted value
```

Do you have any plans to implement uint64 values support? it's widely used type in different systems (linux, snmp, routers) and it's enough complicated to avoid usage of it.

Thank you!

121watts:
Closes idpe#7901

### DIFF
--- a/chronograf/dist/dist.go
+++ b/chronograf/dist/dist.go
@@ -6,7 +6,7 @@ import (
 	"fmt"
 	"net/http"
 
-	"github.com/elazarl/go-bindata-assetfs"
+	assetfs "github.com/elazarl/go-bindata-assetfs"
 )
 
 // DebugAssets serves assets via a specified directory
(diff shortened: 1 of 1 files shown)


hoorayimhelping:
@russorat i clicked the link in that ticket and read over the recommendation. where is the rec coming from? it seems a little arbitrary and possibly unrelated to our server. i just want to make sure we're configuring things with intent rather than copy/pasting or cargo culting.

hoorayimhelping (approved):

wbaker85:
This is pretty old and the code it is changing has moved around quite a bit, but having better security headers generally seems like a good idea.

@jdstrand - can you weigh in on this change? Basically we'd be adding the following headers to requests for static content from OSS servers:

```
w.Header().Add("X-Frame-Options", "SAMEORIGIN")
w.Header().Add("X-XSS-Protection", "1; mode=block")
w.Header().Add("X-Content-Type-Options", "nosniff")
w.Header().Add("Content-Security-Policy", "script-src 'self'; object-src 'self'")
```

PR #19140, opened 2020-07-29
feat: Add Error Reporting to community templates

zoesteinkamp:
Adding the error reporting to the community template pages. So when errors happen we have a paper trail of reporting, so we can patterns, bad apples, etc. 

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/src/templates/components/CommunityTemplateImportOverlay.tsx
+++ b/ui/src/templates/components/CommunityTemplateImportOverlay.tsx
@@ -19,6 +19,7 @@ import {ComponentStatus} from '@influxdata/clockface'
 // Utils
 import {getByID} from 'src/resources/selectors'
 import {getGithubUrlFromTemplateDetails} from 'src/templates/utils'
+import {reportError} from 'src/shared/utils/errors'
 
 import {
   installTemplate,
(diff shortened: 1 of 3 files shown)


ui/src/templates/components/CommunityTemplateImportOverlay.tsx
```diff
@@ -96,6 +97,7 @@ class UnconnectedTemplateImportOverlay extends PureComponent<Props> {
       return summary
     } catch (err) {
       this.props.notify(communityTemplateInstallFailed(err.message))
+      reportError(err, {name: 'The community template install failed'})
```
hoorayimhelping, line 100: this is happening in the review templates section, let's change the name to make it a specific:
```suggestion
      reportError(err, {name: 'The community template fetch for preview failed'})
```

ui/src/templates/containers/CommunityTemplatesIndex.tsx
```diff
@@ -169,6 +170,7 @@ class UnconnectedTemplatesIndex extends Component<Props> {
       )
     } catch (err) {
       this.props.notify(communityTemplateUnsupportedFormatError())
+      reportError(err, {name: 'The community template install failed'})
```
hoorayimhelping, line 173: let's tweak this name to be a little more specific - we already have one that says the install failed.

```suggestion
      reportError(err, {name: 'The community template getTemplateDetails failed'})
```

PR #19146, opened 2020-07-29
fix(dashboard-timezone-selection): toggling timezone should update queries timerange to respect timezone selection

### LINKED ISSUE #17877: The UTC / Local dropdown does not work

Asalem1:
Closes #17877

### Problem

The problem with this was twofold: 

1. Toggling between timezones in the explorer (VEO / DE) was incorrectly adjusting times when timezones were East of GMT.
2. Toggling between timezones on the dashboard page was not updating the timerange parameters to respect the timezone in dashboards
3. Timerange selections with a timezone were not being respected in data visualization

### Solution

The solutions to these issues were:

1. Update the offset to be an absolute value rather than a negative, since that was incorrectly adjusting the times (and was leading to confusing assumptions)
2. Updated the timerange being used for the cells (in the TimeSeries component) to account for timezone selections
3. Updated the timeRange that was being passed to the Views (data visualizations) to respect the timezone selections

### DIFF
--- a/ui/src/dashboards/selectors/index.test.ts
+++ b/ui/src/dashboards/selectors/index.test.ts
@@ -5,7 +5,6 @@ import {
   getTimeRangeWithTimezone,
 } from 'src/dashboards/selectors/index'
 import {getTimezoneOffset} from 'src/dashboards/utils/getTimezoneOffset'
-import moment from 'moment'
 
 jest.mock('src/dashboards/utils/getTimezoneOffset')
 
@@ -44,11 +43,9 @@ describe('Dashboards.Selector', () => {
     '04c6f3976f4b8003',
     '04c6f3976f4b8004',
   ]
-  const lower = `2020-05-05T10:00:00-07:00`
-  const upper = `2020-05-05T11:00:00-07:00`
   const customTimeRangePST = {
-    lower,
-    upper,
+    lower: '2020-05-05T10:00:00-07:00',
+    upper: '2020-05-05T11:00:00-07:00',
     type: 'custom',
   } as CustomTimeRange
   const customTimeRangeCET = {
--- a/ui/src/dashboards/selectors/index.test.ts
+++ b/ui/src/dashboards/selectors/index.test.ts
@@ -1,9 +1,13 @@
 // Funcs
+import {mocked} from 'ts-jest/utils'
 import {
   getTimeRange,
   getTimeRangeWithTimezone,
 } from 'src/dashboards/selectors/index'
+import {getTimezoneOffset} from 'src/dashboards/utils/getTimezoneOffset'
 import moment from 'moment'
+
+jest.mock('src/dashboards/utils/getTimezoneOffset')
 
 // Types
 import {RangeState} from 'src/dashboards/reducers/ranges'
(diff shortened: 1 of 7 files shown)


ui/src/dashboards/selectors/index.ts
```diff
@@ -59,7 +59,7 @@ export const setTimeToUTC = (date: string): string => {
   if (offset < 0) {
     return moment
       .utc(date)
-      .add(offset, 'minutes')
+      .add(Math.abs(offset), 'minutes')
```
hoorayimhelping, line 62:
```suggestion
      .subtract(offset, 'minutes')
```
seems odd to me to explicitly check for a negative number, then take the absolute value of it :shrug:

Asalem1: @hoorayimhelping I hear what you're saying and it's something I considered but wasn't bold enough to try. Let me give it a shot and see if it has the same outcome

Asalem1:
> how about adding some test cases for this bug and its fix: https://github.com/influxdata/influxdb/blob/master/ui/src/dashboards/selectors/index.test.ts

Definitely hear where you're coming from but am a little a loss on how to add some tests for stability here, since it would require mocking out date locales in order to ensure that this feature works across different timezones (which is something that we weren't able to do originally). 

If you'll recall from the previous PR that's related to this issue, this wasn't actually possible due to configuration variables that cannot be manipulated within out current testing framework

ui/src/dashboards/selectors/index.ts
hoorayimhelping: @Asalem1 here's a method i've used in the past to gain confidence:

1. write a test that triggers the bug
    1. swap out your fix and code from master while writing the test if it helps writing the test
    1. confirm the test fails against master
1. apply your fix and get passing tests reliably
1. add your scary negative subtraction
1. if tests pass, confidence. if tests fail, put it back the way it works 👍 

this works because you go through the steps of triggering the bug against master, then fixing it. that is what gives you the confidence in your solutions *and* your tests.

Asalem1: Totally a great idea, unfortunately testing relative time based on a timerange here is not as straightforward as it seems due to the constraints of our current testing framework

ui/src/dashboards/selectors/index.ts
```diff
@@ -50,19 +50,13 @@ export const sortDashboardByName = (dashboards: Dashboard[]): Dashboard[] =>
 // 10-11:00am local time (offset depending on timezone)
 export const setTimeToUTC = (date: string): string => {
   const offset = new Date(date).getTimezoneOffset()
-  if (offset > 0) {
-    return moment
-      .utc(date)
-      .subtract(offset, 'minutes')
-      .format()
+  if (offset === 0) {
+    return moment.utc(date).format()
```
ebb-tide, line 54: Is this needed? if offset = 0, would `.subtract(offset, 'minutes')` not just return zero? I don't know the answer just thought I'd check

Asalem1: That's a great question / point. I think I had originally implemented this so that it wouldn't overwrite formatted data unnecessarily. Let me just check real quick to see if it makes a difference

Asalem1: Great point, totally didn't consider this. You're 100% correct

Asalem1:
The most recent PR addresses some of the limitations that the previous testing suite could not. In order to do so, I had to mock out the getTimezoneOffset functionality for testing purposes. This decision was made after making attempts to mock out the Date object's timezone reference within the test suite itself. However, this proved to be an unlikely path. The most compelling argument I found can be seen here:

https://stackoverflow.com/a/15171030/7963795

ui/src/dashboards/selectors/index.test.ts
```diff
@@ -118,6 +136,70 @@ describe('Dashboards.Selector', () => {
       type: 'custom',
     }
 
+    mocked(getTimezoneOffset).mockImplementation(() => 420)
```
hoorayimhelping, line 139: ~can you add a comment here saying this is the value returned for california in daylight savings time?~

actually ignore the above, it's subtly wrong. this is the value returned for -7 UTC timezone, which is the timezone california is in during daylight savings time. in the winter time, this will be 360.

ui/src/dashboards/selectors/index.test.ts
```diff
@@ -118,6 +136,70 @@ describe('Dashboards.Selector', () => {
...
+      },
+    }
+
+    const expected = {
+      lower: `2020-05-05T10:00:00Z`,
+      upper: `2020-05-05T11:00:00Z`,
+      type: 'custom',
+    }
+
+    mocked(getTimezoneOffset).mockImplementation(() => -120)
```
hoorayimhelping, line 170: same thing here about adding a comment for whichever commie timezone is +2 utc

ui/src/dashboards/utils/getTimezoneOffset.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * PLEASE READ
```
hoorayimhelping, line 2: suggestion:  ✂️ the 'please read' part

a big comment implies it should be read. 'please read' almost comes off like the comment is begging for attention, like `pls respond`

Asalem1: You got it

PR #19158, opened 2020-07-30
feat(query/stdlib): enable read group min max

fchikwekwe:
This PR will enable readGroup min/max in OSS behind a feature flag. 

To do: 
- [x] Create feature flags specific to min/max for ReadGroup
- [x] Update storage capabilities in OSS and/or idpe to include min/max for ReadGroup
- [x] Update planner rule PushDownGroupRule and add tests
- [x] Add tests to storage/flux/table_test.go (tests added, and passing, but there are issues with GroupTable as seen below)
- [x] Add flux end-to-end tests (influxdata/flux#3051)
- [x] Add launcher test to OSS
- [x] Add pipeline test to idpe

### DIFF
--- a/query/stdlib/influxdata/influxdb/rules.go
+++ b/query/stdlib/influxdata/influxdb/rules.go
@@ -1023,6 +1023,8 @@ func (rule PushDownGroupAggregateRule) Pattern() plan.Pattern {
 			universe.SumKind,
 			universe.FirstKind,
 			universe.LastKind,
+			universe.MinKind,
+			universe.MaxKind,
 		},
 		plan.Pat(ReadGroupPhysKind))
 }
@@ -1075,6 +1077,28 @@ func (PushDownGroupAggregateRule) Rewrite(ctx context.Context, pn plan.Node) (pl
 			AggregateMethod:   universe.LastKind,
 		})
 		return node, true, nil
+	case universe.MinKind:
+		// ReadGroup() -> min => ReadGroup(min)
+		if feature.PushDownGroupAggregateMinMax().Enabled(ctx) {
+			node := plan.CreatePhysicalNode("ReadGroupAggregate", &ReadGroupPhysSpec{
+				ReadRangePhysSpec: group.ReadRangePhysSpec,
+				GroupMode:         group.GroupMode,
+				GroupKeys:         group.GroupKeys,
+				AggregateMethod:   universe.MinKind,
+			})
+			return node, true, nil
+		}
+	case universe.MaxKind:
+		// ReadGroup() -> max => ReadGroup(max)
+		if feature.PushDownGroupAggregateMinMax().Enabled(ctx) {
+			node := plan.CreatePhysicalNode("ReadGroupAggregate", &ReadGroupPhysSpec{
+				ReadRangePhysSpec: group.ReadRangePhysSpec,
+				GroupMode:         group.GroupMode,
+				GroupKeys:         group.GroupKeys,
+				AggregateMethod:   universe.MaxKind,
+			})
+			return node, true, nil
+		}
 	}
 	return pn, false, nil
 }
@@ -1102,6 +1126,12 @@ func canPushGroupedAggregate(ctx context.Context, pn plan.Node) bool {
 	case universe.LastKind:
 		agg := pn.ProcedureSpec().(*universe.LastProcedureSpec)
 		return caps.HaveLast() && agg.Column == execute.DefaultValueColLabel
+	case universe.MaxKind:
+		agg := pn.ProcedureSpec().(*universe.MaxProcedureSpec)
+		return caps.HaveMax() && agg.Column == execute.DefaultValueColLabel
+	case universe.MinKind:
+		agg := pn.ProcedureSpec().(*universe.MinProcedureSpec)
+		return caps.HaveMin() && agg.Column == execute.DefaultValueColLabel
 	}
 	return false
 }
--- a/storage/flux/table.gen.go
+++ b/storage/flux/table.gen.go
@@ -7,6 +7,7 @@
 package storageflux
 
 import (
+	"fmt"
 	"math"
 	"sync"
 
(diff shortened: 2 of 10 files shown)


storage/flux/table.gen.go.tmpl
```diff
@@ -742,49 +743,21 @@ func (t *{{.name}}GroupTable) advance() bool {
...
-	// we can find the row with the smallest timestamp.
-	// Do not worry about data with math.MaxInt64 as its real timestamp.
-	// In OSS we require a |> range() call in the query and a math.MaxInt64 timestamp
-	// cannot make it through.
-	var timestamp int64 = math.MaxInt64
-	if t.gc.Aggregate().Type == datatypes.AggregateTypeLast {
-		timestamp = math.MinInt64
+	groupBy, err := determine{{.Name}}AggregateMethod(t.gc.Aggregate().Type)
+	if err != nil {
+		panic(err)
```
fchikwekwe, line 748: I don't think that I should panic here, but I'm not sure how to handle this case.

storage/flux/table.gen.go.tmpl
```diff
@@ -742,40 +743,38 @@ func (t *{{.name}}GroupTable) advance() bool {
...
+		{{if and (ne .Name "Boolean") (ne .Name "String")}}
+		timestamp, next = groupBy(timestamps, values)
+		//todo(faith): trying to aggregate values between array cursors if the subsequent
+		// value is still part of the same group. Unsure of the logic here.
+		if len != 0 {
+			value += next
+		} else {
+			value = next
+		}
+		{{end}}
```
fchikwekwe, line 776: I'm missing something in my logic here to preserve values between array cursors. I feel like the overall logic is pretty close, but there's something off here.

storage/flux/table.gen.go.tmpl
```diff
@@ -797,6 +797,133 @@ func (t *{{.name}}GroupTable) advance() bool {
...
+	//todo(faith): I need a way to keep track of the values between array cursors
+	return math.MaxInt64, values[0]
+}
+{{end}}
+
+{{if and (ne .Name "Boolean") (ne .Name "String")}}
+func groupBySum{{.Name}}(timestamps []int64, values []{{.Type}})(int64, {{.Type}}) {
+	//todo(faith): I need a way to keep track of the values between array cursors
+	return math.MaxInt64, values[0]
+}
```
fchikwekwe, line 896: These are the affected cases that are causing the remaining tests to fail.

storage/flux/table.gen.go
```diff
@@ -801,6 +781,113 @@ func (t *floatGroupTable) advance() bool {
...
+	}
+
+	return timestamp, value
+}
+
+// For group count and sum, the timestamp here is always math.MaxInt64.
+// their final result does not contain _time, so this timestamp value can be anything
+// and it won't matter.
+
+func groupBySumFloat(timestamps []int64, values []float64) (int64, float64) {
```
e-dard, line 855: Typically this is something that would benefit from going as fast as possible, and it shouldn't be too complicated to improve performance for a nice tight loop like this. 

One thing that might be worth trying here is to unroll the loop. Typically I have found with right loops like this you can usually improve performance by 30-40% depending on the situation. 

I haven't tested this but I would try something like:

```go
        // we need to know how many values might not fit into blocks of four.
       // we can sum them up afterwards.
        remainder := len(values) % 4

        // we will store the intermediate results in this array.
	var sums [4]float64

        // sum up any extra values...
	for _, v := range a[len(a)-remainder:] {
		sums[0] += v
	}
        
        // iterate over four values at a time. This can improve overall perf
        // because of a bunch of reasons (bounds checking, pipelining, better cache, fewers instructions). 
	for i := 0; i < len(a)-remainder; i += 4 {
		sums[0] += a[i+3] //  as a habit I put the biggest index first incase it helps reduce bounds checking...
		sums[1] += a[i+2]
		sums[2] += a[i+1]
		sums[3] += a[i]
	}

        // just need to sum up the intermediate results
        return math.MaxInt64, sums[0] + sums[1] + sums[2] + sums[3]
```

Of course, I'm speculating and benchmarks are the source of truth 😄. But I would expect something like the above to be quite a bit faster for larger groups of values.

storage/flux/table.gen.go.tmpl
```diff
@@ -797,6 +777,141 @@ func (t *{{.name}}GroupTable) advance() bool {
...
+// their final result does not contain _time, so this timestamp value can be anything
+// and it won't matter.
+{{if eq .Name "Integer"}}
+func groupByCount{{.Name}}(timestamps []int64, values []{{.Type}})(int64, {{.Type}}) {
+	return groupBySum{{.Name}}(timestamps, values)
+}
+{{end}}
+
+{{if and (ne .Name "Boolean") (ne .Name "String")}}
+func groupBySum{{.Name}}(timestamps []int64, values []{{.Type}})(int64, {{.Type}}) {
```
e-dard, line 878: Since `timestamps` is not used it can be omitted for clarity.

```suggestion
func groupBySum{{.Name}}(_ []int64, values []{{.Type}})(int64, {{.Type}}) {
```

storage/flux/table.gen.go.tmpl
e-dard: I agree. It seems to me that if it's not possible to guarantee that only supported aggregate types will be passed down here then ` advance() bool` needs to be refactored to ` advance() (bool, error)`.

It looks above that `t.do`, which takes the `advance` function returns an error, so perhaps this would be a relatively small refactor?

storage/flux/table_test.go
```diff
@@ -2564,6 +2564,115 @@ func TestStorageReader_EmptyTableNoEmptyWindows(t *testing.T) {
 	}
 }
 
+func TestStorageReader_ReadGroup(t *testing.T) {
```
e-dard, line 2567: 👍 nice tests here

storage/flux/table.gen.go
```diff
@@ -801,6 +781,113 @@ func (t *floatGroupTable) advance() bool {
...
+
+	default:
+		return nil, &influxdb.Error{
+			Code: influxdb.EInvalid,
+			Msg:  fmt.Sprintf("unknown/unimplemented aggregate type: %v", agg),
+		}
+	}
+}
+
+func groupByMinFloat(timestamps []int64, values []float64) (int64, float64) {
```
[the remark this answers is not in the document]
jsternberg: In this specific circumstance I think the timestamps would generally not be sorted. The reason is these methods are all aggregation methods of series that have already been aggregated so we're dealing with aggregating the series together here. So for `min` we can have the first series select a point at 40 seconds and then in the second series at 10 seconds.

storage/flux/table.gen.go
jsternberg: This is a cool idea and I'd be interested in benchmarking this as part of a future PR.

jsternberg:
@e-dard I think I addressed your feedback and I think it passes tests now. Can you re-review?

e-dard:
@jsternberg howdy. Is `advance` still able to panic? I don't know if it's github caching issue but still looks like it's panicking.

jsternberg:
I think it's a caching issue. It doesn't panic anymore. Can you refresh to try and confirm that?

storage/flux/table.gen.go.tmpl
```diff
@@ -797,6 +778,141 @@ func (t *{{.name}}GroupTable) advance() bool {
...
+// aggregate the aggregates.
+func determine{{.Name}}AggregateMethod(agg datatypes.Aggregate_AggregateType) ({{.name}}AggregateMethod, error){
+ 	switch agg {
+	case datatypes.AggregateTypeFirst:
+		return groupByFirst{{.Name}}, nil
+	case datatypes.AggregateTypeLast:
+		return groupByLast{{.Name}}, nil
+	case datatypes.AggregateTypeCount:
+		{{if eq .Name "Integer"}}
+		return groupByCount{{.Name}}, nil
```
wolffcm, line 795: This is a little confusing, since `groupByCount` actually does a `sum`, as it should, but I had to double-check to make sure. Maybe just put a comment here explaining this?

storage/flux/table.gen.go.tmpl
```diff
@@ -797,6 +778,141 @@ func (t *{{.name}}GroupTable) advance() bool {
...
+
+	return timestamp, value
+}
+{{end}}
+
+// For group count and sum, the timestamp here is always math.MaxInt64.
+// their final result does not contain _time, so this timestamp value can be anything
+// and it won't matter.
+{{if eq .Name "Integer"}}
+func groupByCount{{.Name}}(timestamps []int64, values []{{.Type}}) (int64, {{.Type}}) {
```
wolffcm, line 873: Do we need this? Can we just call `groupBySum` directly everywhere this would be called? It might make this a little easier to understand.

jsternberg: I don't think it's needed, but I didn't want it to be confusing. Do you think it would be better named as `aggregateCountGroups`? That might represent more of what this is. It is an aggregation of existing groups rather than the grouping themselves.

query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -1075,6 +1077,46 @@ func (PushDownGroupAggregateRule) Rewrite(ctx context.Context, pn plan.Node) (pl
...
 		})
 		return node, true, nil
+	case universe.MinKind:
+		// ReadGroup() -> min => ReadGroup(min) -> min
+		if feature.PushDownGroupAggregateMinMax().Enabled(ctx) {
+			root := plan.CreatePhysicalNode("min", &universe.MinProcedureSpec{
+				SelectorConfig: execute.SelectorConfig{
+					Column: execute.DefaultValueColLabel,
+				},
+			})
```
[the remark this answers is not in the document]
jsternberg: I'll try this out.

storage/flux/table.gen.go.tmpl
wolffcm: Yes, I like that name just fine.

PR #19164, opened 2020-07-30
feat(tenant): Allow tenant to be composable and still maintain access across services

### LINKED ISSUE #19155: TenantService CreateOrganization should use BucketService
## What

The tenant service implementation of `CreateOrganization` does not use the `BucketService` when creating system buckets:

https://github.com/influxdata/influxdb/blob/c02e83038f7cab64f221982b56b53fd24fea37c4/tenant/service_org.go#L124

If there is middleware associated with the `BucketService`, it will not be called.

- [ ] Change `CreateOrganization` to use `BucketService`

lyondhill:
Tenant services often need to make function calls to other services inside the tenant system. This can cause a problem when a external system needs to inject middleware into the call stack for the secondary system call. For instance when a bucket is created we may need to alert the storage system of the change. To accomplish this we need a way for the org svc to call a middleware influxdb.BucketService that is a tenant type service wrapped with a middleware piece. This is accomplished by allowing the individual services access to the tenant system that was generated. If a middleware is wrapping one of the systems the embedded system can still make internal calls

Closes #19155

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/tenant/service_org.go
+++ b/tenant/service_org.go
@@ -8,8 +8,20 @@ import (
 	"github.com/influxdata/influxdb/v2/kv"
 )
 
+type OrgSvc struct {
+	store *Store
+	svc   *Service
+}
+
+func NewOrganizationSvc(st *Store, svc *Service) *OrgSvc {
+	return &OrgSvc{
+		store: st,
+		svc:   svc,
+	}
+}
+
 // Returns a single organization by ID.
-func (s *Service) FindOrganizationByID(ctx context.Context, id influxdb.ID) (*influxdb.Organization, error) {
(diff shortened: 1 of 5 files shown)


tenant/service_org.go
```diff
@@ -176,32 +187,46 @@ func (s *Service) UpdateOrganization(ctx context.Context, id influxdb.ID, upd in
...
-	return err
+	if err != nil {
+		return err
+	}
+
+	return s.removeResourceRelations(ctx, id)
+}
+
+// removeResourceRelations allows us to clean up any resource relationship that would have normally been left over after a delete action of a resource.
+func (s *OrgSvc) removeResourceRelations(ctx context.Context, resourceID influxdb.ID) error {
```
AlirieGray, line 218: nice

PR #19165, opened 2020-07-30
fix: apply feedback from default sort order

drdelambre:
Peek 2020-07-30 14-27

this closes the loop on this issue #19118

essentially, not applying a default sort is super cool, but we also needed to be able to clear existing sorts


ui/src/dashboards/utils/tableGraph.ts
```diff
@@ -1,6 +1,5 @@
 import _ from 'lodash'
 import {fastMap, fastReduce, fastFilter} from 'src/utils/fast'
-import {isFlagEnabled} from 'src/shared/utils/featureFlag'
```
zoesteinkamp, line 3: BE free! release the code!

PR #19166, opened 2020-07-30
fix(disappearing-cells): wrapping set views and setting cells in a timeout to ensure they get populated

Asalem1:
Related #18598 

### Problem

Clicking on a dashboard immediately after it was set created a weird race where the cells and views were still loading from the initial getDashboards call and were setting over the cells and views that were being gotten from the getDashboard function.

### Solution

Store a reference to the timeOut that by reference of the dashboard ID and clear the timeout if the dashboard is being loaded since the cells and views for the dashboard are already being loaded when the dashboard mounts

### DIFF
--- a/ui/src/dashboards/actions/thunks.ts
+++ b/ui/src/dashboards/actions/thunks.ts
@@ -359,12 +359,20 @@ export const getDashboard = (
     const cellViews: CellsWithViewProperties = resp.data.cells || []
     const viewsData = viewsFromCells(cellViews, dashboardID)
 
-    const normViews = normalize<View, ViewEntities, string[]>(
-      viewsData,
-      arrayOfViews
-    )
+    setTimeout(() => {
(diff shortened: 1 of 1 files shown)


ui/src/dashboards/actions/thunks.ts
```diff
@@ -359,12 +359,20 @@ export const getDashboard = (
     const cellViews: CellsWithViewProperties = resp.data.cells || []
     const viewsData = viewsFromCells(cellViews, dashboardID)
 
-    const normViews = normalize<View, ViewEntities, string[]>(
-      viewsData,
-      arrayOfViews
-    )
+    setTimeout(() => {
+      const normCells = normalize<Dashboard, DashboardEntities, string[]>(
+        resp.data.cells,
```
drdelambre, line 364: this should be `cellViews`

Issue #18887, opened 2020-07-08
"failed to create tokens" when creating a token 


anwar6953:
If I use `Select All` to specify the read and write scopes then it fails.
If I select them individually, then it works fine.

![image](https://user-images.githubusercontent.com/2440977/86957277-58667000-c10f-11ea-9f2e-6d636aa6dbee.png)

![image](https://user-images.githubusercontent.com/2440977/86957386-7fbd3d00-c10f-11ea-8a40-ba99ef1d99fb.png)

desa:
@anwar6953 we believe that this issue should be fixed. Can you confirm that you are no longer seeing this behavior?

anwar6953:
I tested it again just now; it still failed for me.

anwar6953:
Note that it is not a blocker, since I could easily select the buckets individually, but I just wanted surface the broken functionality.

desa:
@anwar6953 is this a local installation or in our cloud product?

anwar6953:
It's the cloud product.

If you click "All Buckets", it works. But if you click "Select All", then it fails.
https://imgur.com/a/LZbtqr3

desa:
@anwar6953 awesome! Just reproduced this myself. Thank you for reporting!

TCL735:
Unable to reproduce, per @ShmuelLotman 

Closing this.

PR #19180, opened 2020-08-03
test: generate Read/Write token - select all buckets

Cubone21:
Covers the issue #18887 - "failed to create tokens" when creating a token


wbaker85:
I'm closing this PR since the UI code it is modifying no longer exists in the `influxdb` repo and is now contained in the `influxdata/ui` repo - feel free to re-open it in `influxdata/ui` if it would still be useful to have there!

PR #19163, opened 2020-07-30
perf: dashboard rendering events

121watts:
Closes #8020

### The Problem
We would like to be able to answer the question:  "How long does it take for a users dashboard to visualize all _visible_ cells?" .

### The Solution
Fire an event when dashboard mounts with
 - dashboardID
 - orgID 
 - mountID // to easily group components in the same render cycle

Fire an event when a cells visualization has mounted but _not_ if the cell has mounted as a result of scrolling into view. Emit the event with the following tags:
 - dashboardID
 - orgID 
 - mountID
 - cellID
 - type // the type of visualization rendered

This should allow us to calculate the delta between dashboard render and last cell visualization render. 

### Notes 
I will greatly simplify the scroll detection logic once https://github.com/influxdata/clockface/issues/523 gets merged in.  

### Bonus
Tired of seeing this?

Screen Shot 2020-07-30 at 2 01 05 PM

Great, so was I.  It's gone now.

PR #19186, opened 2020-08-03
perf: calculate and emit event for cell mount => cell visualized

121watts:
Closes #19163 

### The Problem
Just because a dashboard mounts does not necessarily mean a cell visualization renders.  Users can navigate away, refresh the page, close the browser etc.

### The Solution
Do a some work on the frontend and record the delta between cell mounting and a visualization render.  Emit said event on vis render.

Screen Shot 2020-08-03 at 12 01 33 PM

### DIFF
--- a/ui/src/cloud/utils/reporting.ts
+++ b/ui/src/cloud/utils/reporting.ts
@@ -141,8 +141,15 @@ export const event = (
     /* eslint-disable no-console */
     console.log(`Event:  [ ${measurement} ]`)
     if (Object.keys(meta).length) {
+      console.log('tags')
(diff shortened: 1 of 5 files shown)


ui/src/cloud/utils/reporting.ts
```diff
@@ -141,11 +141,18 @@ export const event = (
     /* eslint-disable no-console */
     console.log(`Event:  [ ${measurement} ]`)
     if (Object.keys(meta).length) {
+      console.log('tags')
```
drdelambre, line 144: dope

PR #19197, opened 2020-08-04
perf: ts mount

121watts:
Updates `renderID` when dashboard is manually refreshed.  Also, provides correct render cycle when user scrolls.

### DIFF
--- a/ui/src/shared/components/TimeSeries.tsx
+++ b/ui/src/shared/components/TimeSeries.tsx
@@ -46,6 +46,7 @@ import {TIME_RANGE_START, TIME_RANGE_STOP} from 'src/variables/constants'
 // Actions & Selectors
 import {notify as notifyAction} from 'src/shared/actions/notifications'
 import {hasUpdatedTimeRangeInVEO} from 'src/shared/selectors/app'
+import {setCellMount as setCellMountAction} from 'src/perf/actions'
 
 // Types
 import {
@@ -71,6 +72,7 @@ interface QueriesState {
 }
 
 interface OwnProps {
+  cellID?: string
   className?: string
   style?: CSSProperties
   queries: DashboardQuery[]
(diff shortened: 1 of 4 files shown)


ui/src/shared/components/TimeSeries.tsx
```diff
@@ -126,13 +128,21 @@ class TimeSeries extends Component<Props, State> {
   private pendingCheckStatuses: CancelBox<StatusRow[][]> = null
 
   public componentDidMount() {
+    const {cellID, setCellMount} = this.props
     this.observer = new IntersectionObserver(entries => {
       entries.forEach(entry => {
         const {isIntersecting} = entry
-        if (!this.isIntersecting && isIntersecting && this.pendingReload) {
+        const reload =
```
hoorayimhelping, line 135:
```suggestion
        const shouldReload =
```

for clarity here, since there's a function called `reload`, and it's easy to mix up a verb representing a function and a noun representing a variable

ui/src/shared/components/TimeSeries.tsx
```diff
@@ -141,9 +151,15 @@ class TimeSeries extends Component<Props, State> {
   }
 
   public componentDidUpdate(prevProps: Props) {
-    if (this.shouldReload(prevProps) && this.isIntersecting) {
+    const {setCellMount, cellID} = this.props
+    const reload = this.shouldReload(prevProps) && this.isIntersecting
```
hoorayimhelping, line 155:
```suggestion
    const shouldReload = this.shouldReload(prevProps) && this.isIntersecting
```

ui/src/shared/components/TimeSeries.tsx
121watts: I used to feel the same exact way.  until typescript.  now what used to be "what is this thing" become very clear.

121watts:
@hoorayimhelping re the visibility change.  the clock starts when a cell becomes visible.  previously, the clock would start when a cell mounted and it could be WAYYYY down under the viewport.

ui/src/shared/components/TimeSeries.tsx
hoorayimhelping: @121watts i don't follow - there's no type hinting in this code, there's no context saying what this is, beyond the code itself. how does typescript help a reader understand the intent of the code here, or communicate what these things are?

hoorayimhelping:
@121watts i was thinking more through what happens when a cell loads, then changes browser tabs or OS windows, then the cell either finishes loading or reloads when they come back to the tab

121watts:
that's def something to look out for.

PR #19200, opened 2020-08-04
feat: allow a metadata key to be required for a query to be logged

adrian-thurston:
Can specify that a key must be present in the query response metadata before
LoggingProxyQueryService logs the query. Will use this in gateway to only log
the query when the connection to queryd fails.

https://github.com/influxdata/idpe/issues/8013


adrian-thurston:
> Looks great, thanks for the test.

No problem, there was already some nice testing infrastructure present. Made it easy.

PR #19201, opened 2020-08-04
feat(tsdb): Only load latest shard during starting up

### LINKED ISSUE #19131: InfluxDB startup too slowly
InfluxDB costs minutes to startup if the number of shards is great.  It will bring  bad effect to users.

Solution:
During startup, we can only load the latest shards and others on demand.

DengzhiLiu:
Closes #19131 

- [ ] Rebased/mergeable
- [ ] Tests pass


DengzhiLiu:
@dgnorton Please help me to take a look, thanks.

lv254165754:
I think You forget to be compatible with queries? 
though you not changed store.shards ,it contains shards that are not open.
so you can not query data from these shard(**without TSMReader because It's not opened**).
@DengzhiLiu

krzysdabro:
I have experimented with this PR for a few days and with minor changes, I got significant improvements on a large database (32 databases, 30 days retention each):
|              | v1.7.6 | #19201 |
|--------------|--------|--------|
| Startup time | ~10min | ~1min  |
| Open files   | ~55K   | ~1K    |
| Memory usage | ~16GB  | ~2G    |

@lv254165754 I have also verified queries and Influx opens the shards when needed.

The shards were opened anyway soon after startup as `Shard.IsIdle()` was called on each one in the `Store.monitorShards()` method. I have made some changes to prevent that:

```diff
diff --git a/tsdb/shard.go b/tsdb/shard.go
index b066140966..f912a93c78 100644
--- a/tsdb/shard.go
+++ b/tsdb/shard.go
@@ -303,13 +303,18 @@ func (s *Shard) Statistics(tags map[string]string) []models.Statistic {
 // Path returns the path set on the shard when it was created.
 func (s *Shard) Path() string { return s.path }
 
+// IsOpen returns true when the shard is open.
+func (s *Shard) IsOpen() bool {
+	return s.status == ShardRunning && s._engine != nil
+}
+
 // Open initializes and opens the shard's store.
 func (s *Shard) Open() error {
 	if err := func() (err error) {
 		s.mu.Lock()
 		defer s.mu.Unlock()
 		// Return if the shard is already open
-		if s.status == ShardRunning || s._engine != nil {
+		if s.IsOpen() {
 			return nil
 		}
 
diff --git a/tsdb/store.go b/tsdb/store.go
index 7cb2e591ad..1ac1ca6c8a 100644
--- a/tsdb/store.go
+++ b/tsdb/store.go
@@ -1946,6 +1946,10 @@ func (s *Store) monitorShards() {
 		case <-t.C:
 			s.mu.RLock()
 			for _, sh := range s.shards {
+				if !sh.IsOpen() {
+					continue
+				}
+
 				if sh.IsIdle() {
 					if err := sh.Free(); err != nil {
 						s.Logger.Warn("Error while freeing cold shard resources",
```

krzysdabro:
@russorat @dgnorton Can you review this PR?

tsdb/store.go
```diff
@@ -351,19 +352,19 @@ func (s *Store) loadShards() error {
...
+			for index, sh := range shardDirs {
 				// Series file should not be in a retention policy but skip just in case.
 				if sh.Name() == SeriesFileDirectory {
 					log.Warn("Skipping series file in retention policy dir", zap.String("path", filepath.Join(s.path, db.Name(), rp.Name())))
 					continue
 				}
 
 				n++
-				go func(db, rp, sh string) {
+				go func(db, rp, sh string, id int) {
```
ayang64, line 363: are we relying on this function closing over any variables?  if not, can we make this a separate function?

ayang64 (changes requested):
Thank you!  These are really promising changes.

krzysdabro:
Seems like @DengzhiLiu is not active. Can we make this PR move forward? I am willing to make requested changes by myself if that's okay.

dgnorton:
If I understand the change correctly, it is converting the tsdb store so that most shards are opened lazily, upon request when a query or write for the shard comes in. I think this is a good idea in general but have concerns about the potential change in behavior in use cases that have many shards and run queries that span multiple shards. In that case, it may cause queries to timeout. Also, a query could suddenly trigger many shards to load, temporarily starving the system of resources and possibly causing other operations, like writes, to fail. Prior to this change, all the resource-intensive work of opening shards was done before the server started accepting writes or queries.

We will need to review this change carefully and test it on some real world use cases.

jcdauchy-moodys:
Hello all, I have tried to build branch 1.8.5 + this PR and just do minimal testing (measure startup time).

For really large database, it is quite an improvement. My 500g+ db is now starting in less than 10 minutes where before it was taking almost an hour to startup.

As I don't know much about Influxdb internals, I would like to understand what is the risk to run this "custom" version :
- in which case it can be counter-productive (burst of CPU and failed writes) ?
- does it alter the db file format or just the startup step ?

Thanks for any feedback.
JC

gregarndt:
We are regularly hitting this with influx 1.8.x.  It's about a 500gb data set with 2 week shards.  Restarting influx takes up to a hour to complete, which makes even changing configuration values a huge pain and incurs downtime. 

Any update on this PR and if it'll be pushed forward?

jcdauchy-moodys:
@krzysdabro Krzysztof Dąbrowski Did you finally used those changes in "production" ?

Issue #18958, opened 2020-07-15
Dashboard has no cells


hoorayimhelping:
#### Background
Clicking a dashboard from the dashboard index page while the page is still loading results in the Dashboard showing a message that says, "Please add Cells"

![Screen Shot 2020-07-14 at 4.59.56 PM.png](https://images.zenhubusercontent.com/5c33b08d24dda4712ef4ae10/173135b3-a2f7-431f-b567-b3592351be18)

#### To Reproduce
1. Open a dashboard index page with data on it (the tools one is a great example https://influxdb.aws.influxdata.io/orgs/844910ece80be8bc/dashboards-list)
1. Before the dashboard index page is finished loading, click a link to a dashboard
1. Observe the page lying directly through its teeth by saying there aren't any dashboard cells (reload to prove the lying page wrong)

desa:
As I started to repro this, I noticed additional weird behavior. Sometimes I see the cells flash for a second and then disappear and other times they appear after multiple seconds of waiting.

Asalem1:
This should be resolved with:

https://github.com/influxdata/influxdb/pull/19166

PR #19212, opened 2020-08-04
fix(flashing-cells): prevent flashing cells by setting all the cell data when the dashboard is loaded

Asalem1:
Related #18958

### Problem
Clicking on a dashboard immediately after it was set created a weird race where the cells and views were still loading from the initial getDashboards call and were setting over the cells and views that were being gotten from the getDashboard function.

### Solution
Set the cells when loading the dashboard and also wrap the process in a setTimeout so that it gets added to the callstack and should (hopefully) resolve once the initial getDashboards timeouts resolve

### Notes
Added an e2e test to ensure dashboard cloning doesn't break (#19207).

crud-cells

### DIFF
--- a/ui/src/dashboards/actions/thunks.ts
+++ b/ui/src/dashboards/actions/thunks.ts
@@ -358,17 +358,25 @@ export const getDashboard = (
 
     const cellViews: CellsWithViewProperties = resp.data.cells || []
     const viewsData = viewsFromCells(cellViews, dashboardID)
+    setTimeout(() => {
+      const normCells = normalize<Dashboard, DashboardEntities, string[]>(
+        cellViews,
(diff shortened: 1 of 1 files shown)


ui/src/dashboards/actions/thunks.ts
```diff
@@ -358,17 +358,25 @@ export const getDashboard = (
 
     const cellViews: CellsWithViewProperties = resp.data.cells || []
     const viewsData = viewsFromCells(cellViews, dashboardID)
+    setTimeout(() => {
+      const normCells = normalize<Dashboard, DashboardEntities, string[]>(
+        cellViews,
```
drdelambre, line 363: dope

PR #19213, opened 2020-08-04
refactor: remove options from storage bucket finder

### LINKED ISSUE #19204: Change shape of storage.BucketFinder interface
This is an issue to track some clean-up work in lieu of the introduction of the `tenant` service.

The `storage` package retention enforcer depends upon a `BucketFinder` to fetch _all_ buckets.

https://github.com/influxdata/influxdb/blob/master/storage/retention.go#L32-L35

This interface is a reflection of the original `influxdb.BucketService` interface. When called by the enforcer it provides an empty `influxdb.FindOptions{}` to signify that it wants to find all buckets.

However, as we move towards stricter service boundaries the `tenant` implementation enforces that an empty `influxdb.FindOptions{}` means that the default limit is applied. The caller is expected to paginate their call to this service.

The original `kv` implementation is still usage by storage. However, in order for `tenant` to be consumed I propose this interface definition be changed to:

```go
type AllBucketsFinder interface {
    FindAllBuckets(context.Context) ([]*influxdb.Bucket, error)
}
```

Then a type defined which implements this interface and does the pagination across the original `influxdb.BucketService` interface.

## Implementation

1. Define `storage.AllBucketsFinder` interface in `storage/retention.go`.
2. Implement a `PagingAllBucketsFinder` which depends on the original `storage.BucketsFinder` interface and does the paginated walk.
3. Replace existing buckets finder with this new one.

## Success Criteria

- No observed change in behaviour.
- Storage depends on `tenant` implementation of bucket service via paginated one.

AlirieGray:
Closes #19204

This PR updates the storage `BucketFinder` type to remove find options, as these are not actually used and the BucketFinder should return all Buckets without pagination. 

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/storage/bucket_finder.go
+++ b/storage/bucket_finder.go
@@ -1,0 +1,62 @@
+package storage
+
+import (
+	"context"
+
+	"github.com/influxdata/influxdb/v2"
+)
+
+var _ BucketFinder = (*AllBucketsFinder)(nil)
+
+// AllBucketsFinder is responsible for iterating over the paginated reponses from FindBuckets to return every bucket
+type AllBucketsFinder struct {
+	bucketSvc influxdb.BucketService
+}
+
+func NewAllBucketsFinder(bucketSvc influxdb.BucketService) *AllBucketsFinder {
+	return &AllBucketsFinder{
+		bucketSvc: bucketSvc,
+	}
+}
+
+func (p *AllBucketsFinder) FindBuckets(ctx context.Context, filter influxdb.BucketFilter) ([]*influxdb.Bucket, int, error) {
+	buckets := []*influxdb.Bucket{}
+	o := influxdb.FindOptions{Limit: influxdb.MaxPageSize}
+
+	res, n, err := p.bucketSvc.FindBuckets(ctx, filter, o)
+	if err != nil {
+		return buckets, 0, err
+	}
+	buckets = append(buckets, res...)
+
+	// used to filter out duplicate system buckets
+	foundTask := false
+	foundMonitoring := false
+
+	if n == o.Limit {
+		for {
+			o.Offset = o.Offset + o.Limit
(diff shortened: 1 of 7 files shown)


storage/bucket_finder_test.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	expected := 25
+	for i := 0; i < expected; i++ {
+		tenant.CreateBucket(ctx, &influxdb.Bucket{Name: fmt.Sprintf("bucket%d", i), OrgID: id})
+	}
+
+	finder := storage.NewAllBucketsFinder(tenant)
+	_, n, err := finder.FindBuckets(ctx, influxdb.BucketFilter{})
+	if err != nil {
```
jacobmarble, line 38: FYI you might find it easier to use `require.NoError(t, err)` and later `assert.Equal(t, expected+2, n)`

storage/bucket_finder.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+var _ BucketFinder = (*AllBucketsFinder)(nil)
+
+// AllBucketsFinder is responsible for iterating over the paginated reponses from FindBuckets to return every bucket
+type AllBucketsFinder struct {
+	bucketSvc influxdb.BucketService
+}
+
+func NewAllBucketsFinder(bucketSvc influxdb.BucketService) *AllBucketsFinder {
```
jacobmarble, line 16: In this PR, I see this new constructor, and one use of it in a test. Is there follow-on work, or do you mean for the contents of this file to live in `bucket_finder_test.go`?

AlirieGray: There is a follow up PR and I also added the constructor to cmd/launcher.go

AlirieGray:
Closing in favor of https://github.com/influxdata/influxdb/issues/19218

PR #19230, opened 2020-08-05
chore: remove rum poc

### LINKED ISSUE #18653: /flags called twice when loading the app
looks like we call the /flags api twice when we load the app. we should only need to call it once.

hoorayimhelping:
Closes #18653
Closes idpe 8062

Now that we have richer metrics coming in from app-metrics, we can remove the RUM POC, which mainly existed as a goal to get the whole system connected up.

This will also involve removing dashboards in tools that render graphs based on this data.

- [x] Rebased/mergeable
- [x] Tests pass


hoorayimhelping:
@Asalem1 there's one dashboard, one check, one notification endpoint, and one notification rule

Asalem1:
@hoorayimhelping sweet. Shoot for the moon

PR #19231, opened 2020-08-05
feat: add tab index to alerts page

hoorayimhelping:
I got annoyed trying to use my keyboard to filter things on the alerts page, so I rage fixed it. You can use `tab` to navigate into the filter text boxes now.

tabaf

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/ui/cypress/support/index.js
+++ b/ui/cypress/support/index.js
@@ -17,6 +17,7 @@
 import './commands'
 
 import 'cypress-pipe'
+import 'cypress-plugin-tab'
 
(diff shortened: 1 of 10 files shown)


ui/cypress/support/index.js
```diff
@@ -17,6 +17,7 @@
 import './commands'
 
 import 'cypress-pipe'
+import 'cypress-plugin-tab'
```
glinton, line 20: js newb so ignore if not relevant, but is this import used anywhere (i assume the linter would complain if not, and that it could be used in another file)?

PR #19248, opened 2020-08-07
fix(dedup-window-vars): added condition and tests for ensuring windowVars don't get passed an extern more than once

Asalem1:
### Problem

The previous implementation of the Query cache appended the windowPeriod variable regardless of whether it was added as a variableAssignment. These duplicated extern values were causing an issue when trying to run the query

### Solution

Added a helper function to check whether the variableAssignment has a reference to the windowPeriod before appending it to the list of variables. Added tests to ensure that windowPeriod (and variableAssignments) are deduplicated before passing them in to runQuery

### DIFF
--- a/ui/src/shared/apis/query.test.ts
+++ b/ui/src/shared/apis/query.test.ts
@@ -207,6 +207,45 @@ describe('query', () => {
         }
       })
     })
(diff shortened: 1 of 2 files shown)


Asalem1:
Will do. I'll message him when it gets deployed onto tools

PR #8081, opened 2017-03-01
Refactor the subquery code and fix outer condition queries

jsternberg:
This change refactors the subquery code into a separate builder class to
help allow for more reuse and make the functions smaller and easier to
read.

The previous function that handled most of the code was too big and
impossible to reason through.

This also goes and replaces the complicated logic of aggregates that had
a subquery source with the simpler IteratorMapper. I think the overhead
from the IteratorMapper will be more, but I also believe that the actual
code is simpler and more robust to produce more accurate answers. It
might be a future project to optimize that section of code, but I don't
have any actual numbers for the efficiency of one method and I believe
accuracy and code clarity may be more important at the moment since I am
otherwise incapable of reading my own code.

Fixes #8045.

- [x] Rebased/mergable
- [x] Tests pass
- [x] CHANGELOG.md updated

PR #19258, opened 2020-08-07
fix: file upload

121watts:
Closes #8081

### The Problem
Somewhere along the line we removed the upload limit.  This caused a bunch of odd state changes that made it unclear as to how a user should proceed.  

### Solution...for now
Re-institute the upload limit. 

### Real Solution
Have the `api/v2/write` accept files so we can take advantage of `xhrs` file progress API or `fetch` streams.  This way, we can upload larger files and properly display the progress to the user.  

### Notes
This was much more tedious than expected.  The state machine that is `LineProtocolWizard` had not been touched in years.  It was overly generalized so it could be re-used elsewhere.  It's not used anywhere else.  This led to a bunch of unnecessary props and state which I attempted to untangle the best I could.  Since no other part of the app cares about the state of this overlay, I pulled its state out of redux and into its own reducer with the `useReducer` hook.

A part of me wishes I had just started from scratch.


ui/cypress/e2e/buckets.test.ts
```diff
@@ -293,31 +293,97 @@ describe('Buckets', () => {
...
+      })
+    })
+
+    it('upload a file and write data', () => {
+      cy.getByTestID('add-data--button').click()
+      cy.getByTestID('bucket-add-line-protocol').click()
+      cy.getByTestID('Upload File').click()
+
+      // When a file is larger than 10MB
+      const bigFile = 'data-big.txt'
```
drdelambre, line 327: instead of sending in a big file that we gotta tote around, why not make that max size configurable for tests? or use a function to generate it?

ui/src/buckets/components/lineProtocol/LineProtocolFooterButtons.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const LineProtocolButtons: FC<Props> = ({onSubmit}) => {
+  const history = useHistory()
+  const [{tab}, dispatch] = useContext(Context)
+  const handleCancel = () => {
+    dispatch(reset())
+  }
+
+  const handleClose = () => {
+    history.goBack()
```
drdelambre, line 25: do we want to be more explicit here on the url in case of a deeplink?

ui/src/buckets/components/lineProtocol/configure/DragAndDropButtons.tsx
```diff
@@ -0,0 +1,65 @@
...
+} from '@influxdata/clockface'
+
+interface Props {
+  fileSize: number
+  uploadContent: string
+  onCancel: () => void
+  onSubmit: () => void
+}
+
+const MAX_FILE_SIZE = 1e7 // 10MB
```
drdelambre, line 16: we can use env vars to set this too

drdelambre: or have a `if NODE_ENV === 'test'` kinda thing

ui/src/buckets/components/lineProtocol/configure/DragInfo.tsx
```diff
@@ -0,0 +1,15 @@
+// Libraries
+import React, {FC} from 'react'
+import cn from 'classnames'
```
drdelambre, line 3: look at you, you rebel

drdelambre:
is there a way we could define `WritePrecision` outside of the old client library since it's not supported and this has been the most i've seen it touched in 4mo?

PR #19264, opened 2020-08-10
build(flux): update Flux to v0.78.0

rockstar:


go.mod
```diff
@@ -32,8 +31,8 @@ require (
 	github.com/go-stack/stack v1.8.0
 	github.com/gogo/protobuf v1.3.1
 	github.com/golang/gddo v0.0.0-20181116215533-9bd4a3295021
-	github.com/golang/mock v1.3.1
-	github.com/golang/protobuf v1.3.3
+	github.com/golang/mock v1.4.3
+	github.com/golang/protobuf v1.4.2
```
nathanielc, line 35: Do you know why this dep was updated? Does Flux need the newer version?

rockstar: Yeah, I'm not sure, and it's certainly causing problems. It's just the result of `go get ...flux@v0.78.0; go mod tidy`

nathanielc: Hmm, how is is causing problems? I checked an Flux doesn't have a dependency on github.com/golang/protobuf  or github.com/golang/mock so its not directly because of Flux. 

We should probably understand why it wants to update before moving forward.

PR #19271, opened 2020-08-10
feat: Community template e2e

### LINKED ISSUE #19176: UI: e2e Tests
#### Background
To go through the whole flow, our e2e tests will depend on github being up and reachable. We may want to install some templates via the CLI as part of a setup process

#### Acceptance Criteria
- [x] Check that template page loads as expected
- [x]  Check the browse community template button launches gituhb
- [x] Confirm the lookup template errors on invalid template
  - [x] empty
  - [x] bad url
  - [x] bad file type (.html for example)
  - [x] passing in a github folder should break
- [x] Confirm look up template accepts valid url and launches overlay
   - [x] raw github url (e.g.: https://raw.githubusercontent.com/influxdata/community-templates/master/downsampling/dashboard.yml)
  - [x] pretty github url (e.g.: https://github.com/influxdata/community-templates/blob/master/downsampling/dashboard.yml)
- [x] Confirm there is a resources dropdown present, and they match the amount of resources listed at top
  - [x] Handles pluralization correctly (1 resource, 0 resources, 2 resources, etc.)
  - [x] 0 resources selected should hide the install button
  - [x] Cannot uncheck buckets
  - [x] Cannot uncheck variables
- [x] Confirm the install template button works and confirm template has been added to install list
   - [ ] Confirm that templates installed from the CLI show up on the index page on page reload
   - [x] Confirm installing the same template twice, creates a new, separate, instance
- [x] Check that you can delete installed templates
- [x] Confirm that clicking on label, task, bucket, telegraf, dashboard, and variable reroute you to the appropriate place
- [x] Confirm that clicking on the source takes you to GitHub with a github source
   - [ ] Shows a file link when installed from the CLI

zoesteinkamp:
Closes #19176 

This issues add a few e2e tests for our community template logic, so we can feel more confident in our code going forward

- [x] Rebased/mergeable
- [x] Tests pass


ui/cypress/e2e/communityTemplate.test.ts
```diff
@@ -0,0 +1,231 @@
+describe('Community Templates', () => {
```
Asalem1, line 1: why not just wrap the entire thing in an `if(isFlagEnabled)` rather than setting the flag in the `config.yml`?

hoorayimhelping: @Asalem1 the tests would never run if the flag isn't turned on

Asalem1: Makes sense. Is there a follow-up issue to update the `config` file for when the feature flag is removed?

hoorayimhelping: @Asalem1 yessir, it's part of the upcoming milestone.

PR #19288, opened 2020-08-11
feat(task_service): respect after and before flags

### LINKED ISSUE #19247: influxdb_client_2.0.0-beta.15_darwin_amd64 task run list not working as expected
Hi! I’m having some issues with the influx cli on mac, seems like the filters for influx task run list do not function as expected. 

- a) it does not seem the `--after` or `--before` args are respected as the returned data is outside of those bounds
- b) the`--limit` argument is required even though the documentation does not specify that. Omitting it errors with ` run limit is out of bounds, must be between 1 and 500.` whereas i would expect a default

Running on mac with 
influxdb_client_2.0.0-beta.15_darwin_amd64

RohanSreerama5:
Closes #19247 

This PR adds an implementation to respect --after and --before time flags. 

Notes: 
- Implementation is currently verbose for debugging purposes. A refactor can be performed in places to shorten the implementation as it is seen fit. 
- Tasks are filtered using the ScheduledFor time attribute

Where I left off (As of 08/21/2020): 
- Previous implementation has been narrowed down into a simpler binary function `between(..)`
- Parsing the user-provided timestamp inside task/backend/analytical_storage.go has now been left up to the Flux script. The Flux runScript uses the time() function to convert timestamp to RFC3339. 
- Latest work: Moved implementation from http layer to kv/task.go and modified the flux runScript to use different start/stop in range() based on the after and before flags provided. 
- before flag works. after flag is unreliable because it always populates one or more task runs that are out of range. 
- Work to be done: Track down where the after flag is being lost in the code 
- Write tests and feature flag this change (note: left a lot of commented code to outline approaches I might've attempted in the past feel free to remove it)


http/task_service.go
```diff
@@ -408,7 +408,7 @@ type runsResponse struct {
 	Runs  []*runResponse    `json:"runs"`
 }
 
-func newRunsResponse(rs []*influxdb.Run, taskID influxdb.ID) runsResponse {
+func newRunsResponse(rs []*influxdb.Run, taskID influxdb.ID) runsResponse { //important
```
gavincabbage, line 411: Remove this comment `//important`

http/task_service.go
```diff
@@ -1564,7 +1566,16 @@ func (t TaskService) FindRuns(ctx context.Context, filter influxdb.RunFilter) ([
...
+	if filter.AfterTime != "" {
+		params = append(params, [2]string{"afterTime", filter.AfterTime})
+	}
+
+	if filter.BeforeTime != "" {
+		params = append(params, [2]string{"beforeTime", filter.BeforeTime})
+	}
+
+	if filter.Limit < 0 || filter.Limit > influxdb.TaskMaxPageSize { //possible area of error
+		fmt.Printf("ERRORING HERE: %v, %v", influxdb.TaskMaxPageSize, filter.Limit)
```
gavincabbage, line 1578: Remove this debug `Printf`

http/task_service.go
```diff
@@ -1580,6 +1591,59 @@ func (t TaskService) FindRuns(ctx context.Context, filter influxdb.RunFilter) ([
 		return nil, 0, err
 	}
 
+	if filter.BeforeTime != "" || filter.AfterTime != "" {
```
gavincabbage, line 1594: Can you provide some more explanation of the goal of this new block of code?

http/task_service.go
```diff
@@ -1580,6 +1591,59 @@ func (t TaskService) FindRuns(ctx context.Context, filter influxdb.RunFilter) ([
...
+
+		var timeStampAfter time.Time
+		if filter.AfterTime != "" {
+			timeStampAfter, err = time.Parse(time.RFC3339, filter.AfterTime)
+			if err != nil {
+				return nil, 0, err
+			}
+		}
+
+		var myRuns []*influxdb.Run //potentially dangerous bc dynamic allocation as slice grows in run-time (consider refactor here)
```
gavincabbage, line 1612: Remove any `my` prefixing of variable names, e.g. `myRuns` can just be `runs`.

http/task_service.go
RohanSreerama5: This is the implementation that handles when a user enters a particular time using the --after or --before flags. It handles the case where only the -after flag is used, only the -before flag is used, and where both flags are used.

http/task_service.go
```diff
@@ -1580,6 +1590,59 @@ func (t TaskService) FindRuns(ctx context.Context, filter influxdb.RunFilter) ([
...
+			for j := range rs.Runs {
+				run = rs.Runs[j].httpRun
+				temp = *run.ScheduledFor
+				if timeStampAfter.Before(temp) {
+					flagRuns = append(flagRuns, convertRun(run))
+				}
+			}
+		} else {
+			return nil, 0, errors.New("error finding task runs")
+		}
```
GeorgeMac, line 1641: I think this could do with some de-duplication. You could do:
```go
var before time.Time
if filter.BeforeTime != "" {
	var err error
	before, err = time.Parse(time.RFC3339, filter.BeforeTime)
	if err != nil {
		return nil, 0, err
	}
}

var after time.Time
if filter.AfterTime != "" {
	var err error
	after, err = time.Parse(time.RFC3339, filter.AfterTime)
	if err != nil {
		return nil, 0, err
	}
}

between := func(scheduled time.Time) bool {
	return (after.IsZero() || after.Before(scheduled)) && (before.IsZero() || scheduled.Before(before))
}

for j := range rs.Runs {
	if run := rs.Runs[j].httpRun; between(*run.ScheduledFor) {
		flagRuns = append(flagRuns, convertRun(run))
	}
}
```

Also see final comment on location of this fix.

http/task_service.go
```diff
@@ -985,13 +986,14 @@ func decodeGetRunsRequest(ctx context.Context, r *http.Request) (*getRunsRequest
...
 		afterID, err := influxdb.IDFromString(id)
+
 		if err != nil {
 			return nil, err
 		}
 		req.filter.After = afterID
 	}
 
-	if limit := qp.Get("limit"); limit != "" {
+	if limit := qp.Get("limit"); limit != "0" {
```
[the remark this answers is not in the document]
RohanSreerama5: Just added a small fix to check for "" and "0" in the conditional. Hope that's alright!

http/task_service.go
RohanSreerama5: That boolean function sums up all of my code in 1 return statement. I love it!

RohanSreerama5: @GeorgeMac  Regarding moving the code to kv/task.go, it seems to me that we're doing filtering in the HTTP TaskService and the kv TaskService. The difference is that we're finding runs in the HTTP area and finding tasks in the kv area. Because my solution is intended to find task runs, I'm not totally sure it would make sense to transfer that solution to the kv TaskService. Of course, I may be mistaken in my understanding of your comment so please do correct me. -Thanks!

http/task_service.go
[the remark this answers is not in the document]
GeorgeMac: Equally as good!

http/task_service.go
GeorgeMac: Ah I see. You're absolutely correct that I was referring to Tasks and not Runs. My mistake, thanks for clarity!.

I would still however push this logic down into the service and filter in the `TaskService.findRuns` method:
https://github.com/influxdata/influxdb/blob/648ce11df35152aedcfb7bbcfc164cae38a1b9d0/kv/task.go#L983-L1018

The primary reason is to keep the `http` transport layer thin and concerned with decoding/encoding requests and responses and then delegating to services. Rather than having it handle any kind of filtering. To keep responsibilities clear.
Then a test case can be added to the `testing` package as well. That should hold true across each implementation for this behavior.

RohanSreerama5: Hey @GeorgeMac, trying to make the solution work in kv here. A few small things: Firstly, my approach is to have the solution in kv and then call into that from the FindRuns function in HTTP. However, I'm running into issues because I have to call the kv FindRuns with a Service method receiver, which I don't exactly have. Can you give me some tips please? -thx

PR #19290, opened 2020-08-11
feat(mosaic graph): Add Mosaic Graph Type to the ui behind a feature flag

### LINKED ISSUE #19122: Mosaic Plot: UI can save mosaic plot types
#### User Story
As a user, I want to my dashboard to save my Mosaic plot types.

#### Background
See [The Mosaic Plot Type Epic](https://app.zenhub.com/workspaces/applicationmonitoring-5e3b05772922e316b3a210a4/issues/influxdata/giraffe/199) for more information.

#### Sequencing
This is part two of a two-part series. The first part involves setting up the backend to accept this change. See [#19121](https://app.zenhub.com/workspaces/applicationmonitoring-5e3b05772922e316b3a210a4/issues/influxdata/influxdb/19121)

#### Acceptance Criteria
- This work is feature flagged
- Dashboards can save Mosaic plot types

reparker837:
Closes #19122

https://github.com/influxdata/influxdb/pull/19328

Added mosaic graph behind a feature flag

Co-authored-by: Rashi Bose <rashibose@gmail.com>

- [x] Rebased/mergeable
- [x] Tests pass


ui/package.json
```diff
@@ -135,7 +135,7 @@
     "@influxdata/clockface": "2.3.1",
     "@influxdata/flux": "^0.5.1",
     "@influxdata/flux-lsp-browser": "^0.5.11",
-    "@influxdata/giraffe": "0.23.0",
+    "@influxdata/giraffe": "^0.24.0",
```
TCL735, line 138: No caret. Let's keep it to a specific version so that it doesn't get updated on accident by other PRs that yarn install.

hoorayimhelping, line 138: let's pull this part out into a separate PR. i know it's a little tedious, and that logically it's all about adding mosaic plot types, but i think from a risk perspective, and from a documenting-our-work-for-posterity perspective, it's better to have this be a separate thing. if something somehow breaks in this PR, it'll be hard to nail it down. if we add in 0.24.0 separately and it doesn't break anything, and then something breaks from this PR, it'll be clear what to fix. it'll also provide clear history in git - upgrading to giraffe 0.24.0 will look like a separate step followed by adding mosaic plot types behind a flag.

hoorayimhelping: shoot, i think me saying 'do a `yarn add @influxdata/giraffe`, rather than modifying package.json' the other day is what caused this. sorry!

ui/src/shared/utils/vis.ts
```diff
@@ -210,8 +210,22 @@ export const getNumberColumns = (table: Table): string[] => {
...
+
+    return columnType === 'string'
+  })
+
+  return stringColumnKeys
+}
+
 export const getGroupableColumns = (table: Table): string[] => {
-  const invalidGroupColumns = new Set(['_value', '_time', 'table'])
+  const invalidGroupColumns = new Set(['_value', '_time', 'table']) //'_value',
```
hoorayimhelping, line 228: ✂️ the comment

ui/src/shared/utils/vis.ts
```diff
@@ -282,6 +297,34 @@ export const defaultYColumn = (
...
+export const mosaicYcolumn = (
+  table: Table,
+  preferredColumnKey?: string
+): string | null => {
+  const validColumnKeys = getStringColumns(table)
+  if (validColumnKeys.includes(preferredColumnKey)) {
+    return preferredColumnKey
+  }
+
+  const invalidMosaicYColumns = new Set([
```
hoorayimhelping, line 309: why a `Set` and not a regular array?

ui/src/shared/utils/vis.ts
```diff
@@ -282,6 +297,34 @@ export const defaultYColumn = (
...
+  )
+  if (preferredValidColumnKeys.length) {
+    return preferredValidColumnKeys[0]
+  }
+
+  if (validColumnKeys.length) {
+    return validColumnKeys[0]
+  }
+  return null
+}
```
hoorayimhelping, line 326: i like it, clean function :+1:

ui/src/timeMachine/components/Vis.tsx
```diff
@@ -133,8 +135,15 @@ const mstp = (state: AppState) => {
 
   const giraffeResult = getVisTable(state)
   const xColumn = getXColumnSelection(state)
-  const yColumn = getYColumnSelection(state)
-  const fillColumns = getFillColumnsSelection(state)
+
+  let yColumn, fillColumns
```
hoorayimhelping, line 139: style tip: just set the values here and avoid re-setting them in an `else` block. it'll save the following characters from having to be written:

`,`, `}`, `else`, `{`, `yColumn`, `fillColumns`, `}`

so it'll look like:
```ts
let yColumn = getYColumnSelection(state)
let fillColumns = getFillColumnsSelection(state)

if (activeTimeMachine.view.properties.type === 'mosaic') {
  yColumn = getMosaicYColumnSelection(state)
  fillColumns = getMosaicFillColumnsSelection(state)
}
```

again, just a tip. no way is right or wrong, but it's generally thought that the fewer bits of code you have to read and write, the fewer mistakes there'll be

ui/src/timeMachine/selectors/index.ts
```diff
@@ -172,6 +191,32 @@ export const getFillColumnsSelection = (state: AppState): string[] => {
...
+    'view.properties.fillColumns'
+  )
+
+  if (preference === null) {
+    for (const key of validFillColumns) {
+      if (key.startsWith('_value')) {
+        return [key]
+      }
+    }
+    if (table.columnKeys.includes('_value')) {
```
hoorayimhelping, line 208: it's _generally_ better style to put single flow control `if` statements like this ahead of looping flow control statements like `for of` above. that way, if this check fails, we exit without looping over something and potentially failing. cheaper to check one thing once than to check one thing once for every item in a loop.

ui/package.json
hoorayimhelping: i think this is my fault - i told rose to install via `yarn add @influxdata/giraffe`

ui/src/shared/utils/vis.ts
rbose22: I don't think there was a particular reason for it. We just happened to use a set here.

ui/src/timeMachine/selectors/index.ts
rbose22: Could you explain a bit more what you mean? I'm not sure I understand.

reparker837: In this code, we're checking if there's a valid column key for _value, and if not, we're checking if _value is a numeric column. The `if` case would be true for both string and numeric columns, so we don't want to check it until we're sure "_value" is not a string[], so I'm not sure if we'll be able to switch the `if` and `for of` in this case. If you think it's worth looking into, we can try to check the type of the value column instead?

ui/src/shared/utils/vis.ts
reparker837: Is it better to use a regular array than a Set?

reparker837: Groupable columns used set, so I was basing it off of that.

hoorayimhelping: i don't think it's better or worse, i think using a [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) implies that you care about uniqueness.

Here's the description from MDN:
>Set objects are collections of values. You can iterate through the elements of a set in insertion order. A value in the Set may only occur once; it is unique in the Set's collection.

ui/src/timeMachine/components/Vis.tsx
```diff
@@ -133,8 +135,15 @@ const mstp = (state: AppState) => {
 
   const giraffeResult = getVisTable(state)
   const xColumn = getXColumnSelection(state)
-  const yColumn = getYColumnSelection(state)
-  const fillColumns = getFillColumnsSelection(state)
+
+  let yColumn = getYColumnSelection(state)
+  let fillColumns = getFillColumnsSelection(state)
+
+  if (activeTimeMachine.view.properties.type === 'mosaic') {
```
ebb-tide, line 142: it would be really nice to do this in the `getYColumnSelection` and `getFillColumnsSelection` functions themselves?

yarn.lock
```diff
@@ -1,4 +1,2 @@
 # THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
ebb-tide, line 1: we should remove this file!

ui/src/timeMachine/components/Vis.tsx
ebb-tide: let me know if it makes sense?

reparker837: done!

yarn.lock
reparker837: I'm having trouble undoing this change because prettier keeps changing it back, how can I remove the file from this pr?

hoorayimhelping:
See: https://github.com/influxdata/influxdb/pull/19328 which was made to fix the git hell this PR was in. it's a cherry-picked version of the commits here.

PR #19322, opened 2020-08-13
fix: increase notification rule limit

### LINKED ISSUE #18684: only displaying 20 notification rules
__Steps to reproduce:__

1. Create 21 notification rules
2. ...
3. ...

__Expected behavior:__
be able to view / edit / delete all the notification rules

__Actual behavior:__
can view / edit only 20 notification rules, also filter only works o
NOTIFICATIONRULES2
n the displayed 20

121watts:
Closes #18684 

### The Problem
User was only seeing 20 of their notification rules.

### The Temporary Solution
Increase the limit to max (100). This should unblock the user until...

### The Real Solution
Dedicate resources to pagination throughout the UI.  

Kapture 2020-08-13 at 10 22 05

### DIFF
--- a/ui/src/notifications/rules/actions/thunks.ts
+++ b/ui/src/notifications/rules/actions/thunks.ts
@@ -48,7 +48,7 @@ import {
 } from 'src/types'
 
 // Constants
-import {LIMIT} from 'src/resources/constants'
+import {RULE_LIMIT} from 'src/resources/constants'
 
 export const getNotificationRules = () => async (
   dispatch: Dispatch<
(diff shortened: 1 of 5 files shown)


ui/src/notifications/rules/actions/thunks.ts
```diff
@@ -45,7 +46,9 @@ import {
   RuleEntities,
   ResourceType,
 } from 'src/types'
-import {incrementCloneName} from 'src/utils/naming'
+
+// Constants
+import {LIMIT} from 'src/resources/constants'
```
drdelambre, line 51: i know this is how we normally do this, but after watching russ bring down production, i dont think this limit is shared across all  api endpoints. some are 100, some are 200, which means we have no guarantee. Might be better to add a new const like RULE_LIMIT

russorat: yes i agree.

121watts: Took a look at resource maxes in the golang.  You're right.  Some resources are `500` some are `100`

PR #19327, opened 2020-08-13
feat(query): enable the schema mutation optimizations

jsternberg:
This removes the feature flag that disabled schema mutation
optimizations so they are now permanently on.


danxmoran:
@jsternberg can this be closed? It looks like the changes were covered in #20634

PR #19348, opened 2020-08-17
fix(system-vars-mapping): updated variable mapping for system variabes to register changes in system variables

### LINKED ISSUE #19346: Changing System Variables queries cached results
The crux of this issue is that the query cache is incorrectly mapping selected system values when they're being toggled. The issue here then becomes that toggling between relative tumeRanges does not issue a new query even though it should.

Asalem1:
Closes #19346

### Problem

Changing the timeRange in the dashboard page does not issue a new query for the newly selected timerange

### Solution

The issue here is that the system variables stored selected values differently than other variables so we are writing a condition to map system variables differently. This PR also clears the cache whenever the timerange selection is updated 

### DIFF
--- a/ui/src/dashboards/components/DashboardHeader.tsx
+++ b/ui/src/dashboards/components/DashboardHeader.tsx
@@ -106,6 +106,9 @@ const DashboardHeader: FC<Props> = ({
   }
 
   const handleChooseTimeRange = (timeRange: TimeRange) => {
+    if (isFlagEnabled('queryCacheForDashboards')) {
+      resetQueryCache()
+    }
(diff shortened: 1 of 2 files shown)


ui/src/dashboards/components/DashboardHeader.tsx
```diff
@@ -106,6 +106,9 @@ const DashboardHeader: FC<Props> = ({
   }
 
   const handleChooseTimeRange = (timeRange: TimeRange) => {
+    if (isFlagEnabled('queryCacheForDashboards')) {
+      resetQueryCache()
```
Asalem1, line 110: @drdelambre do you think this is necessary?

PR #19353, opened 2020-08-17
feat(storage/reads): replace panic on unsupported type with an error

jsternberg:
This replaces the panic that happens when attempting to create a cursor
for an unsupported aggregate type for the mean aggregate. It instead
refactors the result set so the cursor is constructed as part of the
`Next()` call and false is returned when there is an error creating the
cursor. The cursor is then accessed with `Cursor()`. When `Next()`
returns false, an error can be accessed by using `Err()`.


jsternberg:
I updated this to have the error be set by the call to `Next()` and I refactored the cursors so that the cursor was constructed in the call to `Next()` and accessed in `Cursor()`.

PR #19388, opened 2020-08-19
feat: persist data explorer state

### LINKED ISSUE #18536: Flux queries should survive a browser refresh
when editing a flux query in the UI, when i refresh the page, i expect the flux query to still be there. right now, it is lost.

121watts:
Closes #18536 

Kapture 2020-08-18 at 14 25 59


russorat:
We have decided not to implement this. closing.

PR #19392, opened 2020-08-19
fix(notification/rule): Include the edge of the boundary we are observing.

brettbuddin:
We've been witnessing some failed observations of Check triggers in Notification Tasks. In `from(start: -10m, stop: now())`, the `stop` is **exclusive** to the time range—this is so consecutive windows like `A -> B`, `B -> C`, don't have overlap. If a Notification Task's schedule boundaries align with Check trigger observations, we will miss observations if they occur on that boundary.

This change includes the left-hand boundary in what the Notification Task sees of Check triggers.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Bug Fixes
 
 1. [19331](https://github.com/influxdata/influxdb/pull/19331): Add description to auth influx command outputs.
+1. [19392](https://github.com/influxdata/influxdb/pull/19392) Include the edge of the boundary we are observing.
 
 ## v2.0.0-beta.16 [2020-08-07]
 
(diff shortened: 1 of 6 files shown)


brettbuddin:
@lyondhill All @aanthony1243 and @wolffcm's work 😄 

We will need to think about regenerating existing Notification Tasks once this gets up into IDPE.

lyondhill:
> We will need to think about regenerating existing Notification Tasks once this gets up into IDPE.

This might not be too difficult. Could we just cause an update to the check that regenerated the task?

brettbuddin:
@lyondhill That's what I'm thinking.

GeorgeMac (approved):

PR #19420, opened 2020-08-24
fix: lock map before writes

ayang64:
```
    fix(tsi1): Acquire a lock when modifying measurement map
    
    This patch protects an internal map for concurrent use.
    
    (*LogFile).Writes() method calls
    (*LogFile).createMeasurementIfNotExists() which writes to a shared map.
    
    (*LogFile).Writes() acquires a read-lock which leaves
    createMeasurementIfNotExists() open to concurrent writes to its shared
    map.
    
    This commit adds the ExecEntries method to the *LogFile type so that we
    can properly lock calls to (*LogFile).appendEntry() using defer.
    
    (*LogFile).ExecEntries() is used to mostly replace the body of
    (*LogFile).Writes() and incurs another function call since ExecEntries()
    can't be inlined.  Below is the output of build with "-m -m -m" gcflags:
    
      ./log_file.go:1076:6: cannot inline (*LogFile).ExecEntries: unhandled op DEFER
    
    The performace impact of the additional function call should be
    negligable and is outwieghed by the safety and simplicity of using
    defer.
```

- [x] Rebased/mergeable

### DIFF
--- a/tsdb/index/tsi1/log_file.go
+++ b/tsdb/index/tsi1/log_file.go
@@ -1073,10 +1073,9 @@ func (f *LogFile) seriesSketches() (sketch, tSketch estimator.Sketch, err error)
 	return sketch, tSketch, nil
 }
 
-func (f *LogFile) Writes(entries []LogEntry) error {
-	f.mu.RLock()
-	defer f.mu.RUnlock()
-
+func (f *LogFile) ExecEntries(entries []LogEntry) error {
+	f.mu.Lock()
+	defer f.mu.Unlock()
 	for i := range entries {
 		entry := &entries[i]
 		if err := f.appendEntry(entry); err != nil {
@@ -1084,6 +1083,14 @@ func (f *LogFile) Writes(entries []LogEntry) error {
 		}
 		f.execEntry(entry)
 	}
+	return nil
+}
+
+func (f *LogFile) Writes(entries []LogEntry) error {
+	if err := f.ExecEntries(entries); err != nil {
+		return err
+	}
+
 	// Flush buffer and sync to disk.
 	return f.FlushAndSync()
 }


ayang64:
> Just a question: The PR is marked for 1.8.3 milestone, but is being merged into master-1.x. Would it be then be back-ported into 1.8 branch?

Yes -- thats the plan.

e-dard:
@ayang64 is there an issue that this is resolving? I think it needs some careful testing because this could have significant implications for both read and writer performance.

ayang64:
> @ayang64 is there an issue that this is resolving? I think it needs some careful testing because this could have significant implications for both read and writer performance.

@e-dard -- Yes, there is a related EAR for this.

e-dard:
@ayang64 can you link to this issue from that EAR so it's all connected up? :-)

tsdb/index/tsi1/log_file.go
```diff
@@ -1074,8 +1074,8 @@ func (f *LogFile) seriesSketches() (sketch, tSketch estimator.Sketch, err error)
 }
 
 func (f *LogFile) Writes(entries []LogEntry) error {
-	f.mu.RLock()
-	defer f.mu.RUnlock()
+	f.mu.Lock()
+	defer f.mu.Unlock()
```
dgnorton, line 1078: Can you narrow the scope of the lock to just the loop?

ayang64: absolutely.

tsdb/index/tsi1/log_file.go
```diff
@@ -1074,16 +1074,16 @@ func (f *LogFile) seriesSketches() (sketch, tSketch estimator.Sketch, err error)
...
 
 func (f *LogFile) Writes(entries []LogEntry) error {
-	f.mu.RLock()
-	defer f.mu.RUnlock()
-
+	f.mu.Lock()
 	for i := range entries {
 		entry := &entries[i]
 		if err := f.appendEntry(entry); err != nil {
 			return err
```
dgnorton, line 1081: need `f.mu.Unlock()` before `return err`

dgnorton (approved):

PR #19422, opened 2020-08-24
chore: remove feature flag for task executor

lyondhill:
We now default task execution to pull from the user service a permission
set.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/task/backend/executor/executor.go
+++ b/task/backend/executor/executor.go
@@ -292,16 +292,9 @@ func (e *Executor) createPromise(ctx context.Context, run *influxdb.Run) (*promi
 		return nil, err
 	}
 
-	var perm influxdb.PermissionSet
-	if e.flagger != nil && feature.UseUserPermission().Enabled(ctx, e.flagger) {
-		perm, err = e.ps.FindPermissionForUser(ctx, t.OwnerID)
-		if err != nil {
-			return nil, err
-		}
-	}
-
-	if perm == nil {
-		perm = t.Authorization.Permissions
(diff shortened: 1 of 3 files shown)


task/backend/executor/executor.go
```diff
@@ -292,16 +292,9 @@ func (e *Executor) createPromise(ctx context.Context, run *influxdb.Run) (*promi
...
-	var perm influxdb.PermissionSet
-	if e.flagger != nil && feature.UseUserPermission().Enabled(ctx, e.flagger) {
-		perm, err = e.ps.FindPermissionForUser(ctx, t.OwnerID)
-		if err != nil {
-			return nil, err
-		}
-	}
-
-	if perm == nil {
-		perm = t.Authorization.Permissions
```
GeorgeMac, line 304: I can't remember the context on this. But do we still need this bit? Do we have some like legacy tasks with a Token on them somewhere?

GeorgeMac: I guess not since we added the owner migration. So all Tasks should have an owner ID right?

GeorgeMac (approved):
Looks good. Given we're happy removing that persisted token permission set.

task/backend/executor/executor.go
lyondhill: yep exactly. This is the last effort to remove our dependence on the `t.Authorization` which is populated using `maxPermissions`  Once this goes out we should have severed our dependence on URM's in the task system.

PR #19423, opened 2020-08-24
fix: refactor legacy buckets api

### LINKED ISSUE #19226: Proposal: Fix for legacy buckets in bucket list APIs
Related: #19189

The current implementations for listing / reading legacy system buckets leave something to be desired. It also frequently breaks pagination.

-  `kv`: https://github.com/influxdata/influxdb/blob/fe5e57934ede3019936b4ec98d455f772b9ccb07/kv/bucket.go#L353-L381
- `tenant`: https://github.com/influxdata/influxdb/blob/fe5e57934ede3019936b4ec98d455f772b9ccb07/tenant/service_bucket.go#L139-L169

This is a proposal to address it (mostly) once and for all. Or at least until we actually create a migration strategy for legacy data (from old buckets to new).

The proposal is to add a quick lookup in the name index for these buckets before listing.

## Implementation

1. On an org scoped call to find buckets:
2. Do a lookup for `bucketsindexv1/<orgID>/_tasks` if not present set `needsLegacyTasks := true`.
3. Do a lookup for `bucketsindexv1/<orgID>/_monitoring` if not present set `needsLegacyMonitoring := true`.
4. (a) If request is for first page, then add respective legacy system buckets to beginning of response and adjust `offset` and `limit` accordingly.
4. (b) if request is for non-first page then decrement offset based on number of legacy buckets.
5. Request page based on adjusted offset.

## Success Criteria

- a request for up to 20 buckets returns no more than 20 buckets. Sometimes it returns 22 for legacy accounts.
- given an account where their non-legacy buckets do not exist in the first page, the legacy buckets are not wrongfully added to the response.
- everything continues to work, but not broken like it is today.

AlirieGray:
Closes #19226

This PR changes the way we handle legacy system buckets. If the system buckets are not found on lookup on the first page only, we add them and adjust the limit and offset accordingly. This prevents a bug we had previously where legacy system buckets would sometimes appear multiple times when getting multiple pages of Buckets.

I have opted to put this new change only in the new tenant code, not in the old kv code which is soon to be deprecated. Accordingly, I have tests using the old code ignore the new tests.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/tenant/service_bucket.go
+++ b/tenant/service_bucket.go
@@ -14,6 +14,8 @@ type BucketSvc struct {
 	svc   *Service
 }
 
+const NumLegacyBuckets = 2
+
 func NewBucketSvc(st *Store, svc *Service) *BucketSvc {
 	return &BucketSvc{
 		store: st,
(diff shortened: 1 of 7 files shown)


tenant/service_bucket.go
```diff
@@ -157,8 +151,19 @@ func (s *BucketSvc) FindBuckets(ctx context.Context, filter influxdb.BucketFilte
 			Description:     "System bucket for task logs",
 		}
 
-		buckets = append(buckets, tb)
+		// if request is for first page, then add legacy system bucket to beginning of response and
+		// adjust offset and limit
+		// because this is the first bucket in the list, only return it if there is no After field
+		// if request is for non-first page then decrement offset based on number of legacy buckets
+		if o.Offset < 2 && o.After == nil {
```
hoorayimhelping, line 158: can you create a variable named in a way that describes what `2` means and use that, rather than the [magic number 2](https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants) here?

tenant/service_bucket.go
```diff
@@ -167,7 +172,41 @@ func (s *BucketSvc) FindBuckets(ctx context.Context, filter influxdb.BucketFilte
 			Description:     "System bucket for monitoring logs",
 		}
 
-		buckets = append(buckets, mb)
+		// because this is the second bucket in the list, if there is an After field it will only be returned
+		// if it's After the Task bucket
+		if o.Offset < 3 {
```
hoorayimhelping, line 177: can you create a variable that names this magic number please?

tenant/service_bucket.go
AlirieGray: Great point, thank you!

testing/bucket_service.go
```diff
@@ -913,6 +961,244 @@ func FindBuckets(
...
+						Description:     "System bucket for task logs",
+					},
+					{
+						ID:    idTwo,
+						OrgID: idOne,
+						Name:  "abc",
+					},
+				},
+			},
+		},
```
GeorgeMac, line 1165: This is looking great Alirie.

Just wanted to sync up on this a little later. We've had the split now. So we need to port this work.
Also wanted to talk through adding a case or two to this block. To cover things like pages which shouldn't include the system buckets, but for an org which has system buckets.
Also wanted to talk through some of how things IDs now in tests. Since we use the id generators now, instead of the IDs on the fixtures in the test fields.
This is looking great though.

PR #19434, opened 2020-08-25
feat: community template adding the new github raw readme fetch

### LINKED ISSUE #8199: Tags keys and associated values are incorrect when calling opentsdb API
### Bug report

__System info:__ `master` and `1.2` branch

__Steps to reproduce:__

1. Show there are no series prior to executing commands

~~~bash
$ ./influx -execute 'show series' -database=opentsdb -precision=rfc3339
key
---

~~~

2. Post the following payload via **opentsdb** API. Note the values for the tags `APPLE` and `A Foo`

~~~json
{"metric":"test.cpu", "timestamp": 1490324507, "value":18, "tags": {"APPLE":"orange", "A Foo":"bar"} }
~~~

~~~bash
$ curl -X POST -H "Content-Type: application/json" :4242/api/put -d "{\"metric\":\"test.cpu\", \"timestamp\": $(date +'%s'), \"value\":18, \"tags\": {\"APPLE\":\"orange\", \"A Foo\":\"bar\"} }"
~~~

3.
~~~bash
$ ./influx -execute 'show series' -database=opentsdb -precision=rfc3339
key
---

test.cpu,APPLE=bar,A\ Foo=orange

~~~

4.
~~~bash
$ ./influx -execute 'select * from "test.cpu"' -database=opentsdb -precision=rfc3339
name: test.cpu
time                 A Foo  APPLE value
----                 -----  ----- -----
2017-03-24T02:50:56Z orange bar   18
~~~

__Expected behavior:__ 

* **orange** should be associated with **APPLE** tag
* **bar** should be associated with **A Foo** tag

__Actual behavior:__

* **orange** was associated with **A Foo** tag
* **bar** should be associated with **APPLE** tag

## Issue

The issue lies in the [tags.HashKey][] method, which is currently [sorting][] the keys after they have been escaped, however as it [reconstructs the key][] it pulls the [escaped][] values here in the order of the original tags array.

**NOTE:** An invariant of `Tags`, assuming it is constructed from the [NewTags][] function, is the tags are always sorted by the keys. As such, if there are no escaped keys, this would not be an issue.

Also, this issue does not surface when using the line protocol, as it requires the tags already be escaped and during parsing, [sorts the tags][] if they are not already.

### Tests

The following unit test can be added to [points_test.go][] to demonstrate the specific issue

~~~go
func TestTags_HashKey(t *testing.T) {
	tags = models.NewTags(map[string]string{"A FOO": "bar", "APPLE": "orange", "host": "serverA", "region": "uswest"})
	got := tags.HashKey()
	if exp := ",A\\ FOO=bar,APPLE=orange,host=serverA,region=uswest"; string(got) != exp {
		t.Log("got: ", string(got))
		t.Log("exp: ", exp)
		t.Error("invalid match")
	}
}
~~~

[points_test.go]: https://github.com/influxdata/influxdb/blob/master/models/points_test.go
[sorting]: https://github.com/influxdata/influxdb/blob/master/models/points.go#L1794
[tags.HashKey]: https://github.com/influxdata/influxdb/blob/master/models/points.go#L1770
[reconstructs the key]: https://github.com/influxdata/influxdb/blob/master/models/points.go#L1800-L1810
[escaped]: https://github.com/influxdata/influxdb/blob/master/models/points.go#L1807
[NewTags]: https://github.com/influxdata/influxdb/blob/master/models/points.go#L1662
[sorts the tags]: https://github.com/influxdata/influxdb/blob/master/models/points.go#L449-L482

zoesteinkamp:
Closes #8199

With these the read me tab will now feed in the read me from github. If there is no github read me we currently have a bioler plate response that will eventually need to be changed

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/ui/cypress/e2e/communityTemplates.test.ts
+++ b/ui/cypress/e2e/communityTemplates.test.ts
@@ -70,6 +70,10 @@ describe('Community Templates', () => {
 
     //and check that 0 resources pluralization is correct
     cy.getByTestID('template-install-title').should('contain', 'resources')
+
+    //check that valid read me appears
+    cy.getByTestID('readme-tab').click()
+    cy.get('.markdown-format').should('contain', 'Downsampling Template')
   })
 
   describe('Opening the install overlay', () => {
--- a/ui/cypress/e2e/communityTemplates.test.ts
+++ b/ui/cypress/e2e/communityTemplates.test.ts
@@ -72,7 +72,7 @@ describe('Community Templates', () => {
     cy.getByTestID('template-install-title').should('contain', 'resources')
 
     //check that valid read me appears
-    cy.getByTestID('readme-tab').click()
+    cy.getByTestID('community-templates-readme-tab').click()
     cy.get('.markdown-format').should('contain', 'Downsampling Template')
   })
 
--- a/ui/src/templates/api/index.ts
+++ b/ui/src/templates/api/index.ts
@@ -440,3 +440,15 @@ export const updateStackName = async (stackID, name) => {
 
   return resp
 }
+
+export const fetchReadMe = async (directory: string) => {
+  const resp = await fetch(
+    `https://raw.githubusercontent.com/influxdata/community-templates/master/${directory}/README.md`
+  )
+
+  if (!resp.ok) {
+    throw new Error('Network response was not ok')
+  }
+
+  return resp.text()
+}
--- a/ui/src/templates/api/index.ts
+++ b/ui/src/templates/api/index.ts
@@ -446,8 +446,8 @@ export const fetchReadMe = async (directory: string) => {
     `https://raw.githubusercontent.com/influxdata/community-templates/master/${directory}/README.md`
   )
 
-  if (!resp.ok) {
-    throw new Error('Network response was not ok')
+  if (resp.status >= 300) {
+    throw new Error('Network response was not ok' + resp.statusText)
   }
 
   return resp.text()
--- a/ui/src/templates/api/index.ts
+++ b/ui/src/templates/api/index.ts
@@ -447,7 +447,7 @@ export const fetchReadMe = async (directory: string) => {
   )
 
   if (resp.status >= 300) {
-    throw new Error('Network response was not ok' + resp.statusText)
+    throw new Error(`Network response was not ok:' ${resp.statusText}`)
   }
 
   return resp.text()
--- a/ui/src/templates/components/CommunityTemplateImportOverlay.tsx
+++ b/ui/src/templates/components/CommunityTemplateImportOverlay.tsx
@@ -63,16 +63,18 @@ class UnconnectedTemplateImportOverlay extends PureComponent<Props> {
   }
 
   public render() {
+    const templateDetails = getTemplateNameFromUrl(this.props.stagedTemplateUrl)
+    const templateName = templateDetails.name
+    const templateDirectory = templateDetails.directory
+
     return (
       <CommunityTemplateOverlay
         onDismissOverlay={this.onDismiss}
         onInstall={this.handleInstallTemplate}
         resourceCount={this.props.resourceCount}
         status={this.state.status}
-        templateName={getTemplateNameFromUrl(this.props.stagedTemplateUrl).name}
-        templateDirectory={
-          getTemplateNameFromUrl(this.props.stagedTemplateUrl).directory
-        }
+        templateName={templateName}
+        templateDirectory={templateDirectory}
         updateStatus={this.updateOverlayStatus}
       />
     )
--- a/ui/src/templates/components/CommunityTemplateImportOverlay.tsx
+++ b/ui/src/templates/components/CommunityTemplateImportOverlay.tsx
@@ -70,6 +70,7 @@ class UnconnectedTemplateImportOverlay extends PureComponent<Props> {
         resourceCount={this.props.resourceCount}
         status={this.state.status}
         templateName={getTemplateNameFromUrl(this.props.stagedTemplateUrl).name}
+        templateDirectory={this.props.directory}
         updateStatus={this.updateOverlayStatus}
       />
     )
--- a/ui/src/templates/components/CommunityTemplateImportOverlay.tsx
+++ b/ui/src/templates/components/CommunityTemplateImportOverlay.tsx
@@ -70,7 +70,9 @@ class UnconnectedTemplateImportOverlay extends PureComponent<Props> {
         resourceCount={this.props.resourceCount}
         status={this.state.status}
         templateName={getTemplateNameFromUrl(this.props.stagedTemplateUrl).name}
-        templateDirectory={this.props.directory}
+        templateDirectory={
+          getTemplateNameFromUrl(this.props.stagedTemplateUrl).directory
+        }
         updateStatus={this.updateOverlayStatus}
       />
     )
--- a/ui/src/templates/components/CommunityTemplateOverlay.tsx
+++ b/ui/src/templates/components/CommunityTemplateOverlay.tsx
@@ -18,6 +18,7 @@ interface OwnProps {
   resourceCount: number
   status?: ComponentStatus
   templateName: string
+  templateDirectory: string
   updateStatus?: (status: ComponentStatus) => void
 }
 
@@ -44,7 +45,13 @@ class CommunityTemplateOverlayUnconnected extends PureComponent<Props, State> {
   }
 
   public render() {
-    const {isVisible, onInstall, resourceCount, templateName} = this.props
+    const {
+      isVisible,
+      onInstall,
+      resourceCount,
+      templateName,
+      templateDirectory,
+    } = this.props
 
     return (
       <Overlay visible={isVisible}>
@@ -77,7 +84,7 @@ class CommunityTemplateOverlayUnconnected extends PureComponent<Props, State> {
               {this.state.activeTab === Tab.IncludedResources ? (
                 <CommunityTemplateOverlayContents />
               ) : (
-                <CommunityTemplateReadme />
+                <CommunityTemplateReadme directory={templateDirectory} />
               )}
             </Tabs.Container>
           </Overlay.Body>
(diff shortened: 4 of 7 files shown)


ui/src/templates/components/CommunityTemplateOverlay.tsx
```diff
@@ -71,13 +78,14 @@ class CommunityTemplateOverlayUnconnected extends PureComponent<Props, State> {
                   active={this.state.activeTab === Tab.Readme}
                   id="readme"
                   text="Readme"
+                  testID="readme-tab"
```
hoorayimhelping, line 81:
```suggestion
                  testID="community-templates-readme-tab"
```

ui/src/templates/utils/index.ts
```diff
@@ -100,10 +100,12 @@ const getTemplateDetailsFromFileSource = (_source: string): TemplateDetails => {
...
   url: string
-): {name: string; extension: string} => {
-  const fullName = url.split('/').pop()
+): {name: string; extension: string; directory: string} => {
+  const urlSplit = url.split('/')
+  const fullName = urlSplit.pop()
+  const directory = urlSplit.pop()
   const [name, extension] = fullName.split('.')
-  return {name, extension}
+  return {name, extension, directory}
```
hoorayimhelping, line 108: can you add assertions to the tests here: https://github.com/influxdata/influxdb/blob/master/ui/src/templates/utils/index.test.ts#L51 to make sure directory is what you expect? let me know if you'd like help on this one :)

ui/src/templates/api/index.ts
```diff
@@ -440,3 +440,15 @@ export const updateStackName = async (stackID, name) => {
 
   return resp
 }
+
+export const fetchReadMe = async (directory: string) => {
+  const resp = await fetch(
+    `https://raw.githubusercontent.com/influxdata/community-templates/master/${directory}/README.md`
+  )
+
+  if (!resp.ok) {
```
hoorayimhelping, line 449:
```suggestion
  if (resp.status >= 300) {
```

ui/src/templates/api/index.ts
```diff
@@ -440,3 +440,15 @@ export const updateStackName = async (stackID, name) => {
...
+    throw new Error('Network response was not ok')
```
hoorayimhelping, line 450: do you know what `resp` looks like when this condition gets hit? is there any kind of error message we could throw here?

ui/cypress/e2e/communityTemplates.test.ts
```diff
@@ -70,6 +70,10 @@ describe('Community Templates', () => {
 
     //and check that 0 resources pluralization is correct
     cy.getByTestID('template-install-title').should('contain', 'resources')
+
+    //check that valid read me appears
+    cy.getByTestID('readme-tab').click()
+    cy.get('.markdown-format').should('contain', 'Downsampling Template')
```
hoorayimhelping, line 76: :+1:

ui/src/templates/api/index.ts
zoesteinkamp: got it

ui/src/templates/api/index.ts
zoesteinkamp: ive got resp.statusText now added, it says it will be "A string (default value "OK"), which corresponds to the HTTP status code message."

ui/src/templates/components/CommunityTemplateOverlay.tsx
zoesteinkamp: got it

ui/src/templates/components/CommunityTemplateImportOverlay.tsx
```diff
@@ -70,6 +70,9 @@ class UnconnectedTemplateImportOverlay extends PureComponent<Props> {
         resourceCount={this.props.resourceCount}
         status={this.state.status}
         templateName={getTemplateNameFromUrl(this.props.stagedTemplateUrl).name}
+        templateDirectory={
+          getTemplateNameFromUrl(this.props.stagedTemplateUrl).directory
```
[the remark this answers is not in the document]
zoesteinkamp: added!

ui/src/templates/components/CommunityTemplateReadme.tsx
```diff
@@ -1,6 +1,40 @@
...
+  componentDidMount = async () => {
+    try {
+      const response = await fetchReadMe(this.props.directory)
+      this.setState({readMeData: response})
+    } catch (error) {
+      console.error(
+        'There has been a problem with your fetch operation:',
+        error
+      )
+      this.setState({readMeData: '## No read me to display!'})
```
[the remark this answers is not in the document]
zoesteinkamp: added!

ui/src/templates/utils/index.ts
zoesteinkamp: i have added the assertions to the tests!

ui/src/templates/api/index.ts
```diff
@@ -440,3 +440,15 @@ export const updateStackName = async (stackID, name) => {
...
   return resp
 }
+
+export const fetchReadMe = async (directory: string) => {
+  const resp = await fetch(
+    `https://raw.githubusercontent.com/influxdata/community-templates/master/${directory}/README.md`
+  )
+
+  if (resp.status >= 300) {
+    throw new Error('Network response was not ok' + resp.statusText)
```
hoorayimhelping, line 450:
```suggestion
    throw new Error(`Network response was not ok: ${resp.statusText}`)
```

ui/src/templates/components/CommunityTemplateReadme.tsx
```diff
@@ -1,6 +1,40 @@
...
+  "We don't support images in markdown for security purposes"
+
+export class CommunityTemplateReadme extends PureComponent<Props> {
+  state = {readMeData: ''}
+  componentDidMount = async () => {
+    try {
+      const response = await fetchReadMe(this.props.directory)
+      this.setState({readMeData: response})
+    } catch (error) {
+      console.error(
```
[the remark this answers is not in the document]
russorat: just displaying there's no readme in the readme tab is good. It's not an "error" that there's no readme.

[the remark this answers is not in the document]
hoorayimhelping: @zoesteinkamp can you change this `console.error` into a `reportError` call similar to how you did it before in templates?

ui/src/templates/utils/index.test.ts
```diff
@@ -49,43 +49,49 @@ describe('Templates utils', () => {
...
+
     expect(name).toBe('dashboard_params')
     expect(extension).toBe('yml')
+    expect(directory).toBe('testdata')
   })
 
   it('returns the template name and extension from the official community templates github repo', () => {
-    const {name, extension} = getTemplateNameFromUrl(
+    const {name, extension, directory} = getTemplateNameFromUrl(
       'https://github.com/influxdata/community-templates/blob/master/csgo/csgo.yml'
```
hoorayimhelping, line 64:
```suggestion
      'https://github.com/influxdata/community-templates/blob/master/hooray/csgo.yml'
```

ui/src/templates/utils/index.test.ts
```diff
@@ -49,43 +49,49 @@ describe('Templates utils', () => {
...
     )
     expect(name).toBe('csgo')
     expect(extension).toBe('yml')
+    expect(directory).toBe('csgo')
```
hoorayimhelping, line 68:
```suggestion
    expect(directory).toBe('hooray')
```

just because it's good to change things up to make sure tests still work

ui/src/templates/utils/index.test.ts
```diff
@@ -49,43 +49,49 @@ describe('Templates utils', () => {
...
     )
     expect(name).toBe('csgo')
     expect(extension).toBe('json')
+    expect(directory).toBe('csgo')
   })
 
   it('handles non secure arbitrary urls', () => {
-    const {name, extension} = getTemplateNameFromUrl(
+    const {name, extension, directory} = getTemplateNameFromUrl(
       'http://www.example.com/blog/cats/catstuff/memes/csgo/downsampling.yml'
```
hoorayimhelping, line 91:
```suggestion
      'http://www.example.com/blog/cats/catstuff/memes/-------/downsampling.yml'
```

ui/src/templates/utils/index.test.ts
```diff
@@ -49,43 +49,49 @@ describe('Templates utils', () => {
...
     )
     expect(name).toBe('downsampling')
     expect(extension).toBe('yml')
+    expect(directory).toBe('csgo')
```
hoorayimhelping, line 95:
```suggestion
    expect(directory).toBe('-------')
```

why not go nuts and see if this works

ui/src/templates/components/CommunityTemplateReadme.tsx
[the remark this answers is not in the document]
zoesteinkamp: yep!

ui/src/templates/api/index.ts
zoesteinkamp: ah thats my mistake actually, thanks for the fix!

ui/src/templates/components/CommunityTemplateReadme.tsx
```diff
@@ -1,6 +1,43 @@
...
+    try {
+      const response = await fetchReadMe(this.props.directory)
+      this.setState({readMeData: response})
+    } catch (error) {
+      reportError(error, {
+        name: 'The community template fetch github readme failed',
+      })
+
+      this.setState({
+        readMeData: '## We cant find the readme associated with this template',
```
hoorayimhelping, line 26:
```suggestion
        readMeData: '## We can't find the readme associated with this template',
```

hoorayimhelping:
Great work! we can ignore the semantic pull request as long as you make sure you squash your commit into a nice clean single commit.

PR #19437, opened 2020-08-25
feat(query): add a feature flag for query tracing

### LINKED ISSUE #16220: Convert the QUERY_TRACING env var switch into a feature flag
influxdata/flux#1992 and #15459 added query tracing via opentracing to flux and influxdb.
The addition will allow OSS to capture much more detailed, per source/transformation level statistics.

At the time of those work being done, we did not have the feature flagging framework we are using now, so we put the work behind an environment variable `QUERY_TRACING`. The detailed tracing is on if  `QUERY_TRACING` is set to `1`.

We need to remove this environment variable and convert that to be a feature flag,

wolffcm:
Closes #16220.

This adds a feature flag for experimental query tracing, which creates a span for each source and transformation in query requests.

### DIFF
--- a/flags.yml
+++ b/flags.yml
@@ -83,9 +83,9 @@
   contact: Query Team
   lifetime: temporary
 
-- name: Experimental Query Tracing
-  description: Turn on experimental query tracing for queries that are sampled
-  key: experimentalQueryTracing
+- name: Query Tracing
+  description: Turn on query tracing for queries that are sampled
+  key: queryTracing
   default: false
   contact: Query Team
   lifetime: permanent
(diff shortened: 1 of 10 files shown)


flags.yml
```diff
@@ -83,6 +83,13 @@
   contact: Query Team
   lifetime: temporary
 
+- name: Experimental Query Tracing
```
yzhang1991, line 86: With this being a permanent flag, I think we can just call this query tracing, leaving out "experimental"? What do you think?

wolffcm: Yes, I agree, I'll change it.

PR #19439, opened 2020-08-26
fix(storage): ArrayFilterCursor truncation for multi-block data

yzhang1991:
Backport https://github.com/influxdata/influxdb/commit/abfe5a54a0b419d0231762574229caddf971bbfd


storage/reads/gen.go
```diff
@@ -1 +1,3 @@
 package reads
+
+//go:generate env GO111MODULE=on go run github.com/benbjohnson/tmpl -data=@array_cursor.gen.go.tmpldata array_cursor.gen.go.tmpl
```
jsternberg, line 3: Remove this. I removed generate statements for code I copied over to avoid interfering with the build or modifying the files.

PR #19448, opened 2020-08-26
feat: flows index

### LINKED ISSUE #7151: Support High Cardinality Tags and Series
### Feature Request

The database should be able to support higher levels of cardinality for tags and series.  Currently, the full tag set is loaded into an in-memory index for fast query planning.  When tags with a large number of values are written, the in-memory index can consume more memory than is available on the host.  

**Proposal:** 

The database should not require loading the full tag set into an in-memory index.  Higher cardinality series and tags should be able to be stored and queried and not be limited by the amount of RAM on the host. 

**Current behavior:**

Currently, high cardinality data causes the process memory usage to grow quickly increasing the chances of an OOM.  It also slows startup times as the the index needs to scan all the stored data to re-create the in-memory index.

Users also frequently write high cardinality tag data by mistake causing the server to crash.  When in this state, removing the problem data is very difficult as well.

**Desired behavior:**

Storing high-cardinality data should not cause the process to OOM or adversely affect startup times.  Query performance should not be adversely affected by higher cardinality data as well.

**Use case:**

It is more natural and convenient to be able to store higher cardinality data at times.  For example, some tag data is ephemeral in nature (docker containers IDs), but can contribute to high cardinality data issues over time.
### Documentation
- [Requirements Doc](https://github.com/influxdata/influxdb/pull/7175)
- [Design Option #1](https://github.com/influxdata/influxdb/pull/7174)
- [Design Option #2](https://github.com/influxdata/influxdb/pull/7186)
- [Design Option #3](https://github.com/influxdata/influxdb/pull/7209)
- [Cardinality Estimation](https://github.com/influxdata/influxdb/pull/7264)

121watts:
Closes #7151

### Notes
This index is purposefully very, very ugly.  It is not done.  Not even close.  But, you can now create and view multiple flows and differentiate them based on a meaningful name.

Kapture 2020-08-26 at 11 23 05

### DIFF
--- a/ui/src/notebooks/components/FlowsIndexEmpty.tsx
+++ b/ui/src/notebooks/components/FlowsIndexEmpty.tsx
@@ -1,0 +1,32 @@
+import React from 'react'
+import {
+  ComponentColor,
+  ComponentSize,
+  EmptyState,
+  Button,
+  IconFont,
+} from '@influxdata/clockface'
+
+const FlowsIndexEmpty = () => {
+  const handleClickCreate = () => {
+    console.log('TODO: create flow')
+  }
+
+  return (
+    <EmptyState size={ComponentSize.Large}>
+      <EmptyState.Text>
+        Looks like there aren't any <b>Flows</b>, why not create one?
+      </EmptyState.Text>
+      <Button
+        icon={IconFont.Plus}
+        color={ComponentColor.Primary}
+        text="Create Flow"
+        titleText="Click to create a Flow"
+        onClick={handleClickCreate}
+        testID="Create Flow"
+      />
+    </EmptyState>
+  )
+}
+
+export default FlowsIndexEmpty
--- a/ui/src/notebooks/components/FlowsIndexEmpty.tsx
+++ b/ui/src/notebooks/components/FlowsIndexEmpty.tsx
@@ -1,30 +1,14 @@
 import React from 'react'
-import {
-  ComponentColor,
-  ComponentSize,
-  EmptyState,
-  Button,
-  IconFont,
-} from '@influxdata/clockface'
+import {ComponentSize, EmptyState} from '@influxdata/clockface'
+import FlowCreateButton from './FlowCreateButton'
 
 const FlowsIndexEmpty = () => {
-  const handleClickCreate = () => {
-    console.log('TODO: create flow')
-  }
-
   return (
     <EmptyState size={ComponentSize.Large}>
       <EmptyState.Text>
         Looks like there aren't any <b>Flows</b>, why not create one?
       </EmptyState.Text>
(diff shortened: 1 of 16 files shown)


ui/src/pageLayout/constants/navigationHierarchy.ts
```diff
@@ -108,17 +108,17 @@ export const generateNavItems = (orgID: string): NavItem[] => {
       activeKeywords: ['data-explorer'],
     },
     {
-      id: 'notebooks',
-      testID: 'nav-item-notebooks',
+      id: 'flows',
```
drdelambre, line 111: so it begins

ui/src/shared/containers/SetOrg.tsx
```diff
@@ -134,9 +135,13 @@ const SetOrg: FC<Props> = ({
           component={RouteToDashboardList}
         />
 
-        {/* Flows */}
+        {/* Flows  */}
         {isFlagEnabled('notebooks') && (
-          <Route path={`${orgPath}/notebooks`} component={NotebookPage} />
+          <Route path={`${orgPath}/notebooks/:id`} component={NotebookPage} />
```
drdelambre, line 140: can rename this if like!

ui/src/notebooks/components/FlowsIndexEmpty.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import {ComponentSize, EmptyState} from '@influxdata/clockface'
+import FlowCreateButton from './FlowCreateButton'
+
+const FlowsIndexEmpty = () => {
+  return (
+    <EmptyState size={ComponentSize.Large}>
+      <EmptyState.Text>
+        Looks like there aren't any <b>Flows</b>, why not create one?
```
drdelambre, line 9: should we just make one for them instead?

ui/src/notebooks/components/header/index.tsx
```diff
@@ -2,7 +2,7 @@
 import React, {FC, useContext, useCallback} from 'react'
 
 // Contexts
-import {NotebookContext} from 'src/notebooks/context/notebook'
+import {NotebookContext} from 'src/notebooks/context/notebook.current'
```
drdelambre, line 5: well this is embarassing

ui/src/notebooks/context/notebook.tsx
```diff
@@ -14,23 +14,27 @@ export interface PipeMeta {
 // TODO: this is screaming for normalization. figure out frontend uuids for cells
```
drdelambre, line 14: I think we can delete this whole file? maybe later, but i got confused as to how any of this worked

PR #19450, opened 2020-08-26
feat(csv-parser): created a csv parser to format the csv into an expected schema

Asalem1:
This PR aims to add a custom CSV parser to ensure that raw CSV data predictably returns Schema data in order to make bucket metadata onto the Flows query builder.
live-data


ui/src/shared/actions/schemaThunks.ts
```diff
@@ -3,14 +3,15 @@ import {Dispatch} from 'react'
 // import {fromFlux as parse} from '@influxdata/giraffe'
 
 // API
-// import {runQuery} from 'src/shared/apis/query'
+import {runQuery} from 'src/shared/apis/query'
```
drdelambre, line 6: this isn't being used

ui/src/shared/parsing/flux/schemaParser.test.ts
```diff
@@ -0,0 +1,110 @@
...
+      },
+    })
+  })
+
+  it('throws when no metadata is present', () => {
+    expect(() => {
+      parseResponse(RESPONSE_NO_METADATA)
+    }).toThrow()
+  })
+})
```
drdelambre, line 110: mmmmm the feeeeeeeels

ui/src/shared/actions/schemaThunks.ts
Asalem1: as in we shouldn't be using this to run the query? It's used here:

https://github.com/influxdata/influxdb/pull/19450/files/0e33e60063b1fab75c9186be801d5a26cf358532#diff-7003025bc3963a7870d490547eb9c4c0R47

drdelambre: nm then i got confused

PR #19475, opened 2020-09-01
test: token detail view

Cubone21:
Token detail view test. Popup content check.


wbaker85:
I'm closing this PR out since the UI code it is referring to no longer exists in the `influxdb` repository. Feel free to re-open an equivalent PR in https://github.com/influxdata/ui if it is still needed!

PR #19477, opened 2020-09-01
chore(tenant): point tests to new tenant service

### LINKED ISSUE #19157: Update unit tests to use TenantService and other new services
## What

Unit tests such as `TestAnalyticalStore` do not use the newer meta services for creating organizations, buckets, etc:

https://github.com/influxdata/influxdb/blob/bce46912804451ece70d8d0136a6d26a3cc046d0/task/backend/analytical_storage_test.go#L42-L44

This prevents these tests from passing when these services are required to leverage features such as middleware.

- [ ] Update unit tests to use the newer services such as the tenant service

lyondhill:
Closes #19157

Now that the tenant service is complete we should be using it for all
test instead of the `kv.Service`.

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/kv/scrapers_test.go
+++ b/kv/scrapers_test.go
@@ -38,12 +38,6 @@ func initScraperTargetStoreService(s kv.SchemaStore, f influxdbtesting.TargetFie
 		}
 	}
 
-	for _, m := range f.UserResourceMappings {
-		if err := svc.CreateUserResourceMapping(ctx, m); err != nil {
-			t.Fatalf("failed to populate user resource mapping")
-		}
-	}
-
 	for _, o := range f.Organizations {
 		if err := svc.PutOrganization(ctx, o); err != nil {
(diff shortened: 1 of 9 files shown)


kv/scrapers_test.go
```diff
@@ -38,11 +38,11 @@ func initScraperTargetStoreService(s kv.SchemaStore, f influxdbtesting.TargetFie
 		}
 	}
 
-	for _, m := range f.UserResourceMappings {
-		if err := svc.CreateUserResourceMapping(ctx, m); err != nil {
-			t.Fatalf("failed to populate user resource mapping")
-		}
-	}
+	// for _, m := range f.UserResourceMappings {
```
AlirieGray, line 41: did you want to remove this or leave it commented?

lyondhill: oo ill remove that thanks

PR #19478, opened 2020-09-01
fix(onboard): onboarding no longer add admin user

lyondhill:
The admin user should no longer be added to the new user created by
setup.

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/tenant/service_onboarding_test.go
+++ b/tenant/service_onboarding_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	influxdb "github.com/influxdata/influxdb/v2"
+	icontext "github.com/influxdata/influxdb/v2/context"
 	"github.com/influxdata/influxdb/v2/kv"
 	"github.com/influxdata/influxdb/v2/tenant"
 	influxdbtesting "github.com/influxdata/influxdb/v2/testing"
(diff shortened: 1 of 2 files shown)


tenant/service_onboarding_test.go
```diff
@@ -47,3 +48,38 @@ func initOnboardingService(s kv.Store, f influxdbtesting.OnboardingFields, t *te
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(urms) > 1 {
+		t.Fatal("additional URMs created")
+	}
+	if urms[0].UserID != onboard.User.ID {
+		t.Fatal("org assigned to the wrong user")
+	}
```
GeorgeMac, line 84: Take it or leave it. All of this could be reduced to:
```go
urms, _, err := ten.FindUserResourceMappings(ctx, influxdb.UserResourceMappingFilter{ResourceID: onboard.Org.ID})
require.NoError(t, err)
assert.Equal(t, []*influxdb.UserResourceMappings{
    {ResourceID: onboard.Org.ID, UserID: onboard.User.ID},
}, urms)
```

GeorgeMac (approved):
I get what this is doing and see that it delivers on the goal. So looks good tome.
However, it gets me thinking about the implicit URM created by CreateOrganization based on the presence of a user ID on the Authorizer.
I don't like that this behaviour is a mutating side effect of a property on a context. Also, that it leverages another domain model (authorization token) just to masquerade the authorizer.
Food for thought, for ongoing improvements to the org service in tenant. Perhaps we promote an optional parameter called something like `WithOwner(user *influxdb.User)`. Which when provided creates a URM between the user and the Org. Then this would be a clear explicit contract provided by the internal service definition.

lyondhill:
@GeorgeMac I dont love it either. I was thinking we could follow the pattern where on create we pass a "createOrg" struct similar to how we do in some other places `CreateTask(ctx context.Context, t TaskCreate) (*Task, error)` then we could add the "ownerID" to that object and not interfere with the org. I will write up an issue for this.

GeorgeMac:
I like that @lyondhill ! That is a good clear contract.

PR #19516, opened 2020-09-08
test: filtering and sorting scrapers

Cubone21:
2 new tests for filtering and sorting scrapers


wbaker85:
I am closing this PR since the UI code for influxdb is now entirely contained in the `influxdata/ui` repository. If these tests would still be valuable, feel free to open a PR in `influxdata/ui` if they don't already exist there!

PR #19521, opened 2020-09-08
feat: Port 1.x retention policy enforcement service

### LINKED ISSUE #19309: Port 1.x Retention Policy service
## What

The [`services/retention`](https://github.com/influxdata/influxdb/tree/master-1.x/services/retention) package is required to enforce retention policies for the tsdb storage engine. This service periodically evaluates the retention policy for each bucket and will delete data that is outside the retention policy time window. 

## How

Port the package to 2.x, which should be compatible given the tsdb and meta client packages are still available. This service should be started when `influxd` is launched.

stuartcarnie:
Closes #19309

This PR introduces the retention policy enforcement service from InfluxDB 1.x. This service is responsible for evaluating retention policy rules and deleted shards which fall outside the time bounds.

In addition, configuration of the check interval is available via

```
--storage-retention-check-interval=<duration>
```

Finally, due to the introduction of the retention policy service, a number of types and interfaces were removed to keep the static checker happy, as they were no longer used.

- [x] Rebased/mergeable
- [x] Tests pass


v1/services/retention/config.go
```diff
@@ -0,0 +1,49 @@
...
+	return Config{Enabled: true, CheckInterval: toml.Duration(30 * time.Minute)}
+}
+
+// Validate returns an error if the Config is invalid.
+func (c Config) Validate() error {
+	if !c.Enabled {
+		return nil
+	}
+
+	// TODO: Should we enforce a minimum interval?
```
benbjohnson, line 28: I'm not sure it's worth adding minimums. Hopefully people won't be adjusting this config option at all.

v1/services/retention/service.go
```diff
@@ -0,0 +1,163 @@
...
+		PruneShardGroups() error
+	}
+	TSDBStore interface {
+		ShardIDs() []uint64
+		DeleteShard(shardID uint64) error
+	}
+
+	config Config
+	wg     sync.WaitGroup
+	done   chan struct{}
```
benbjohnson, line 28: Minor nit. This is probably just ported over from the previous implementation but I like using a cancelable background context instead of a `done` channel. It handles idempotency during closing and a lot of things already take a `context.Context`.

e.g.

```go
type Service struct {
    ctx context.Context
    cancel func()
}

func NewService() *Service {
    s := &Service{}
    s.ctx, s.cancel = context.WithCancel(context.Background())
    return s
}

benbjohnson (approved):

v1/services/retention/config.go
stuartcarnie: Will remove that comment

v1/services/retention/service.go
stuartcarnie: Good shout – I'll alter the `Open` method to take a `context.Context`

benbjohnson (approved):

PR #19525, opened 2020-09-09
feat: Port shard pre-creation service from InfluxDB 1.x

### LINKED ISSUE #19520: Port 1.x shard pre-creation service
## What

The [`services/precreator`](https://github.com/influxdata/influxdb/tree/master-1.x/services/precreator) package provides a strategy to ensure new shards are created prior to writes being received.

### From the README

During normal operation when InfluxDB receives time-series data, it writes the data to files known as _shards_. Each shard only contains data for a specific range of time. Therefore, before data can be accepted by the system, the shards must exist and InfluxDB always checks that the required shards exist for every incoming data point. If the required shards do not exist, InfluxDB will create those shards. Because this requires a cluster to reach consensus, the process is not instantaneous and can temporarily impact write-throughput.

Since almost all time-series data is written sequentially in time, the system has an excellent idea of the timestamps of future data. Shard precreation takes advantage of this fact by creating required shards ahead of time, thereby ensuring the required shards exist by the time new time-series data actually arrives. Write-throughput is therefore not affected when data is first received for a range of time that would normally trigger shard creation.

Note that the shard-existence check must remain in place in the code, even with shard precreation. This is because while most data is written sequentially in time, this is not always the case. Data may be written with timestamps in the past, or farther in the future than shard precreation handles.

## How

Port the package to 2.x, which should be compatible given the tsdb and meta client packages are still available. This service should be started when `influxd` is launched.

stuartcarnie:
Closes #19520

This PR copies the shard precreator package from InfluxDB 1.x to enable shard pre-creation support. It makes minor modifications to take a `context.Context` for the `Open` function.

In order to configure the service, the following parameters have been exposed:

```
--storage-shard-precreator-advance-period Duration              The default period ahead of the endtime of a shard group that its successor group is created. (default 30m0s)
--storage-shard-precreator-check-interval Duration              The interval of time when the check to pre-create new shards runs. (default 10m0s)
```
- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influxd/launcher/launcher.go
+++ b/cmd/influxd/launcher/launcher.go
@@ -440,6 +440,16 @@ func launcherOpts(l *Launcher) []cli.Opt {
 			Flag:  "storage-retention-check-interval",
 			Desc:  "The interval of time when retention policy enforcement checks run.",
 		},
+		{
+			DestP: &l.StorageConfig.PrecreatorConfig.CheckInterval,
+			Flag:  "storage-shard-precreator-check-interval",
+			Desc:  "The interval of time when the check to pre-create new shards runs.",
+		},
+		{
+			DestP: &l.StorageConfig.PrecreatorConfig.AdvancePeriod,
+			Flag:  "storage-shard-precreator-advance-period",
+			Desc:  "The default period ahead of the endtime of a shard group that its successor group is created.",
+		},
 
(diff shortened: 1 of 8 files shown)


benbjohnson (approved):
🕺

PR #19527, opened 2020-09-09
fix(variables): use Update function for Put Variable

AlirieGray:
This line was supposed to go in https://github.com/influxdata/influxdb/pull/19368

The ReplaceVariable function needs to use `PutUpdate` instead of `PutNew`, otherwise it will return a "conflict" error. ReplaceVariable is also used to create new Variables for testing, so I added a condition that will allow it to create a Variable if it does not exist or replace it if it does exist.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/kv/variable.go
+++ b/kv/variable.go
@@ -215,9 +215,12 @@ func (s *Service) CreateVariable(ctx context.Context, v *influxdb.Variable) erro
 	})
 }
 
-// ReplaceVariable puts a variable in the store
+// ReplaceVariable replaces a variable that exists in the store or creates it if it does not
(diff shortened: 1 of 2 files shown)


stuartcarnie (changes requested):
Looks good, @AlirieGray. I would recommend adding a unit test that verifies this behavior. 

You could disable the change temporarily to verify the test fails (returning a conflict) and then re-enable the fix to verify the test passes.

stuartcarnie (approved):
Great work, @AlirieGray and 🥇 for the unit tests

PR #19542, opened 2020-09-10
test: new buckets tests

Cubone21:
- new tests for creating scraper and telegraf agent through the Add Data button on a specific bucket
- new test for creating a bucket with a retention
- a separated test for filtering buckets


wbaker85:
I am closing this PR since the UI code for influxdb is now entirely contained in the `influxdata/ui` repository. If these tests would still be valuable, feel free to open a PR in `influxdata/ui` if they don't already exist there!

PR #19546, opened 2020-09-10
refactor: Replace ctx.Done() with ctx.Err()

ayang64:
```
Prior to this commit we checked for context cancellation with a select
block and context.Context.Done() without multiplexing over any other
channel like:

  select {
    case <-ctx.Done():
      // handle cancellation
    default:
      // fallthrough
  }

This commit replaces those type of blocks with a simple check of
ctx.Err().  This has the following benefits:

* Calling ctx.Err() is much faster than entering a select block.

* ctx.Done() allocates a channel when called for the first time.

* Testing the result of ctx.Err() is a reliable way of determininging if
  a context.Context value has been canceled.

There are a few places we optimized the select block by only entering it
every cancelCheckInterval times.  This commit checks ctx.Err() every
time -- this increases the cancellation granularity and is only
negligibly slower (sub-nanosecond).
```

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/influxql/query/executor.go
+++ b/influxql/query/executor.go
@@ -311,15 +311,7 @@ LOOP:
 		e.Metrics.Requests.WithLabelValues(statusLabel).Inc()
 
 		// Check if the query was interrupted during an uninterruptible statement.
-		interrupted := false
-		select {
-		case <-ctx.Done():
-			interrupted = true
-		default:
-			// Query has not been interrupted.
-		}
-
-		if interrupted {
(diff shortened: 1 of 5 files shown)


benbjohnson (approved):

ayang64:
@stuartcarnie -- adding you for an extra review because I also addressed a race that came up in testing.  See second commit.

stuartcarnie (approved):
Good find regarding the race condition 👍

Issue #19547, opened 2020-09-10
Remove remnants of the task and monitoring system buckets using fixed IDs


stuartcarnie:
## What

Remove the remnants of the fixed IDs for the task and monitoring buckets.

After the merge of the TSDB storage engine from InfluxDB 1.x, users will be required to start a fresh OSS instance, including all their metadata. Given that, there is no possibility the task and monitoring buckets will have the fixed IDs `10` and `11` respectively. 

## Solution

Remove these constants:

https://github.com/influxdata/influxdb/blob/b91d7785791a8a5e88c783b60375edee19f62a47/bucket.go#L11-L14

and references such as these:

https://github.com/influxdata/influxdb/blob/45a3f2e87c90c2e2d3fbeaae9a800aaa556c46ac/kv/bucket.go#L354-L382

and:

https://github.com/influxdata/influxdb/blob/45a3f2e87c90c2e2d3fbeaae9a800aaa556c46ac/tenant/storage_bucket.go#L121-L138

Remaining references to those constants in unit tests, such as:

https://github.com/influxdata/influxdb/blob/ab31e7763bd20dd5cbb66e6d2a8f748d771d1f94/task/backend/analytical_storage_test.go#L105

should use local constants.

PR #19548, opened 2020-09-10
chore: remove hardcoded constants for ids

yquansah:
Closes #

Description of issue and what this PR seeks to achieve are detailed within this issue: https://github.com/influxdata/influxdb/issues/19547

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/kv/bucket.go
+++ b/kv/bucket.go
@@ -182,7 +182,6 @@ func (s *Service) findBucketByName(ctx context.Context, tx Tx, orgID influxdb.ID
 		switch n {
 		case influxdb.TasksSystemBucketName:
 			return &influxdb.Bucket{
-				ID:              influxdb.TasksSystemBucketID,
 				Type:            influxdb.BucketTypeSystem,
 				Name:            influxdb.TasksSystemBucketName,
 				RetentionPeriod: influxdb.TasksSystemBucketRetention,
@@ -191,7 +190,6 @@ func (s *Service) findBucketByName(ctx context.Context, tx Tx, orgID influxdb.ID
 			}, nil
 		case influxdb.MonitoringSystemBucketName:
 			return &influxdb.Bucket{
-				ID:              influxdb.MonitoringSystemBucketID,
 				Type:            influxdb.BucketTypeSystem,
 				Name:            influxdb.MonitoringSystemBucketName,
 				RetentionPeriod: influxdb.MonitoringSystemBucketRetention,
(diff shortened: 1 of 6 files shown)


kv/bucket.go
```diff
@@ -182,7 +182,6 @@ func (s *Service) findBucketByName(ctx context.Context, tx Tx, orgID influxdb.ID
 		switch n {
 		case influxdb.TasksSystemBucketName:
 			return &influxdb.Bucket{
-				ID:              influxdb.TasksSystemBucketID,
 				Type:            influxdb.BucketTypeSystem,
 				Name:            influxdb.TasksSystemBucketName,
 				RetentionPeriod: influxdb.TasksSystemBucketRetention,
```
stuartcarnie, line 187: This entire switch statement can be deleted and the `if IsNotFound(err)` statement only needs to return the `default` case:

```go
return nil, &influxdb.Error{
    Code: influxdb.ENotFound,
    Msg:  fmt.Sprintf("bucket %q not found", n),
}
```

kv/bucket.go
```diff
@@ -361,7 +359,6 @@ func (s *Service) FindBuckets(ctx context.Context, filter influxdb.BucketFilter,
 
 	if needsSystemBuckets {
```
stuartcarnie, line 360: Here is also legacy, so everything from:

```go
needsSystemBuckets := true
for _, b := range bs {
```

through to line 382 can be deleted.

tenant/storage_bucket.go
```diff
@@ -120,7 +120,6 @@ func (s *Store) GetBucketByName(ctx context.Context, tx kv.Tx, orgID influxdb.ID
 		switch n {
```
stuartcarnie, line 120: Same for this one, the `if kv.IsNotFound(err) {` need only be `return nil, ErrBucketNotFoundByName(n)` as the fixed system buckets are gone.

stuartcarnie (changes requested):
Apologies if my issue was lacking some clarity, @yquansah! 

Previously, these system buckets did not exist in the key/value store, as we had hard coded them with IDs `10` and `11`. All the routines to read buckets had to manually materialize them. We have since made them real data in the key/value store, so these routines will fetch them as expected by reading the data from the key/value store.

yquansah:
@stuartcarnie Excellent. Where does the logic live that stores these system buckets in key/value store? (Hopefully I am not misunderstanding)

stuartcarnie:
Fair question, @yquansah. These are created automatically when an organization is created, and exist as real metadata in the key/value store:

https://github.com/influxdata/influxdb/blob/45a3f2e87c90c2e2d3fbeaae9a800aaa556c46ac/kv/bucket.go#L136-L156

We then look these buckets up by `name` to find them.

stuartcarnie (approved):
Great cleanup, @yquansah 🧹

Issue #9976, opened 2018-06-15
Negation operators don't work as expected for meta queries


e-dard:
The negation operators (`!=` and `!~`) don't work as expected for some meta queries. The following meta queries are not affected:

```
SHOW SERIES
SHOW TAG VALUES
SHOW FIELD KEYS
```

Here are examples of affected meta queries:

#### Example 1 - SHOW MEASUREMENTS

This example fails on `tip` and also on `1.4.2` and `1.3.6`.
```
> create database db
> use db
Using database db
>
> insert cpu,region=west value=1
> insert cpu,region=east value=1
>
> show measurements
name: measurements
name
----
cpu
>
> show measurements where region = 'west'
name: measurements
name
----
cpu
>
> show measurements where region = 'east'
name: measurements
name
----
cpu
>
> show measurements where region != 'east'
>
> show measurements where region != 'west'
>
```

In both of these cases we should expect the `cpu` measurement to be returned. The issue here is that because there exist a series in the measurement that doesn't satisfy `!= 'east'`, the entire measurement is excluded from the result set. To clarify this a little further:

```
> create database db
> use db
Using database db
>
> insert cpu,region=east value=1
> insert cpu,region=west value=1
> insert mem,region=east value=1
>
> show measurements where region = 'east'
name: measurements
name
----
cpu
mem
>
> show measurements where region != 'west'
name: measurements
name
----
mem
>
>
```
The `!= 'west'` returns `mem` because `mem` does not have a series where `region = west`. 

#### Example 2 - SHOW TAG KEYS

```
> drop database db
> create database db
>
> insert cpu,region=west value=1
> insert cpu,region=east value=1
>
> show tag keys
name: cpu
tagKey
------
region
>
> show tag keys where region != 'east'
> show tag keys where region !~ 'west'
ERR: error parsing query: found west, expected regex at line 1, char 30
> show tag keys where region !~ /west/
```

The same problem as `SHOW MEASUREMENTS`. The negation options will exclude any tag key if there is an example series where the predicate is not true.

The issue is to do with the way the [tags are matched](https://github.com/influxdata/influxdb/blob/master/tsdb/index.go#L1310-L1434) when looking for measurements to return.

benbjohnson:
@pauldix This would be a breaking change for the query language. Can you weigh in on this? Should this be implemented or should we defer to Flux?

[...]

PR #19566, opened 2020-09-15
fix(services/storage): multi measurement queries return all applicable series

jsternberg:
This fixes multi measurement queries that go through the storage service
to correctly pick up all series that apply with the filter. Previously,
negative queries such as `!=`, `!~`, and predicates attempting to match
empty tags did not work correctly with the storage service when multiple
measurements or `OR` conditions were included.

This was because these predicates would be categorized as "multiple
measurements" and then it would attempt to use the field keys iterator
to find the fields for each measurement. The meta queries for these did
not correctly account for negative equality operators or empty tags when
finding appropriate measurements and those could not be changed because
it would cause a breaking change to influxql too.

This modifies the storage service to use new methods that correctly
account for the above situations rather than the field keys iterator.

Some queries that appeared to be single measurement queries also get
considered as multiple measurement queries. Any query with an `OR`
condition will be considered a multiple measurement query.

This bug did not apply to single measurement queries where one
measurement was selected and all of the logical operators were `AND`
values. This is because it used a different code path that correctly
handled these situations.

Fixes influxdata/flux#3005.

Related to #9976.

### DIFF
--- a/tsdb/index.go
+++ b/tsdb/index.go
@@ -1326,7 +1326,7 @@ func (is IndexSet) MeasurementNamesByExpr(auth query.Authorizer, expr influxql.E
 
 		// Determine if there exists at least one authorised series for the
 		// measurement name.
-		if is.measurementAuthorizedSeries(auth, e) {
+		if is.measurementAuthorizedSeries(auth, e, nil) {
 			names = append(names, e)
 		}
 	}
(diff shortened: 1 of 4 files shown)


scbrickley (approved):
LGTM

nathanielc:
@jsternberg Can you expand on the concept of a `multiple measurement query`? Is that a concept of the InfluxQL engine? Or is that a concept of tsdb cursors? If its a concept of tsdb cursors can you explain the intent/reasoning of classifying different read paths as multi vs single measurement?

jsternberg:
@nathanielc it's a concept that seems to be just related to flux. They don't exist in influxql. When the storage service gets a query, it will receive a request with filters such as `r._measurement == "m0" and r._field == "f0"`. Since the storage service can retrieve arbitrary requests with filters, it could potentially read multiple measurements in a single cursor. Each measurement has its own index behind it. The storage service attempts to read the filter to determine if the filter only refers to one measurement so it can use a faster method of accessing the index. If it cannot statically determine that there is only one measurement referenced in the filter, such as when `OR` is used, then it will perform a slower method of determining what field keys exist that allows for merging results from multiple measurements. This is what I call the multi measurement queries since it is used for queries with multiple measurements. This same code path is used when `OR` is used because if I had a query like `(r._measurement == "cpu" and r._field == "usage_user") or (r._measurement == "mem")` then I would read from multiple measurements. But the check for this is very sensitive and it also considers `r._measurement == "cpu" and (r.t0 == "a" or r.t0 == "b")` to be a query that could potentially reference multiple measurements.

When I tried to make that code be smarter and consider that one as a single measurement query, I ran into some edge cases that made it hard for me to statically determine if it should be considered one. For example, `r._measurement == "cpu" and (r.t0 == "a" or r._measurement == "mem")` would result in the inner `r._measurement == "mem"` as impossible and should just not be referenced. In a smarter system, it would get optimized to `r._measurement == "cpu" and r.t0 == "a"`. But if I did `and` and then an or condition with two impossible measurements, it was hard for me to report that as an invalid combination so I decided to leave that code untouched.

I hope that helps explain what multi measurement queries are and why some single measurement queries are considered multi measurement ones.

benbjohnson (approved):

nathanielc:
@jsternberg very helpful thanks! Follow up question, this concept of multi vs single measurement queries is only a tsm1 specific correct?  This is because in tsm2 we used the tsm concept of measurement to be the bucket+orgID not the _measurement column. In contrast the tsm1 maps the concept of a measurement to the Flux concept of _measurement and so therefore there are different indexes for each _measurement.

Does that sound right?

jsternberg:
I'm actually not sure of the answer to that question. I think that's probably true based on what you are saying but I don't know enough about the storage engine interface in v2 to confirm that's the case.

PR #19583, opened 2020-09-18
fix: ShardGroupDuration is updated for retention period updates

### LINKED ISSUE #19518: Failed to update bucket: "retention policy duration must be greater than the shard duration"
There is an error in bucket service when updating bucket retention policy:

__Steps to reproduce:__

In chronograf create bucket without retention policy, than edit bucket settings and set Older than 1d :

image

image

Here is the also link to failing test in java client library:
https://app.circleci.com/pipelines/github/influxdata/influxdb-client-java/533/workflows/0fecf1aa-29c6-430f-ad2d-8406e496d4a5/jobs/1806

__Environment info:__

* InfluxDB version: Version 2.0.0 (8757b5b) nightly build

__Logs:__
ts=2020-09-08T11:40:45.608908Z lvl=error msg="api error encountered" log_id=0P7NnYjG000 handler=bucket error="retention policy duration must be greater than the shard duration"

stuartcarnie:
Closes #19518

This PR ensures the `ShardGroupDuration` is updated according to rules defined in the `meta` package, when the retention period is updated. It is necessary that the `ShardGroupDuration` ≤ `RetentionPeriod`. This PR also includes a unit test which sets the retention period to the lowest possible value, to exercise the original bug.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influxd/launcher/storage_test.go
+++ b/cmd/influxd/launcher/storage_test.go
@@ -5,11 +5,14 @@ import (
 	"io/ioutil"
 	nethttp "net/http"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/influxdata/influxdb/v2"
 	"github.com/influxdata/influxdb/v2/cmd/influxd/launcher"
 	"github.com/influxdata/influxdb/v2/http"
+	"github.com/influxdata/influxdb/v2/pkg/testing/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestStorage_WriteAndQuery(t *testing.T) {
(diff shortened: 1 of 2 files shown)


storage/engine.go
```diff
@@ -266,13 +265,18 @@ func (e *Engine) CreateBucket(ctx context.Context, b *influxdb.Bucket) (err erro
 	return nil
 }
 
-func (e *Engine) UpdateBucketRetentionPeriod(ctx context.Context, bucketID influxdb.ID, d time.Duration) (err error) {
+func (e *Engine) UpdateBucketRetentionPeriod(ctx context.Context, bucketID influxdb.ID, d time.Duration) error {
 	span, _ := tracing.StartSpanFromContext(ctx)
 	defer span.Finish()
 
+	// A value of zero ensures the ShardGroupDuration is adjusted to an appropriate value based on the specified
```
brettbuddin, line 272: Nice that you don't have to do that logic here.

cmd/influxd/launcher/storage_test.go
```diff
@@ -152,3 +155,18 @@ func TestLauncher_BucketDelete(t *testing.T) {
...
+
+	bucket, err := l.BucketService(t).FindBucket(ctx, influxdb.BucketFilter{ID: &l.Bucket.ID})
+	require.NoError(t, err)
+	require.NotNil(t, bucket)
+
+	newRetentionPeriod := 1 * time.Hour
+	bucket, err = l.BucketService(t).UpdateBucket(ctx, bucket.ID, influxdb.BucketUpdate{RetentionPeriod: &newRetentionPeriod})
+	require.NoError(t, err)
+	assert.Equal(t, bucket.RetentionPeriod, newRetentionPeriod)
+}
```
brettbuddin, line 172: I see this verifying the `RetentionPeriod` and there doesn't seem to be a way to verify that the `ShardGroupDuration` behaved as you expected. It might be worth setting another expected value here (other than 1 hour), because I see this codepath: 

https://github.com/influxdata/influxdb/blob/025319c3875a73ede0345687e256ebc3249318ab/v1/services/meta/data.go#L1279-L1281

Which returns `1 * time.Hour` as a default here:

https://github.com/influxdata/influxdb/blob/025319c3875a73ede0345687e256ebc3249318ab/v1/services/meta/data.go#L1266-L1274

stuartcarnie: I thought about that, but we're not really testing the `meta.Client` as we're a consumer of that API in this code. I verified that `UpdateRetentionPolicy` API is tested here:

https://github.com/influxdata/influxdb/blob/025319c3875a73ede0345687e256ebc3249318ab/v1/services/meta/client_test.go#L463-L522

The intent of our test is to confirm we're able to set the lowest value and that the `UpdateRetentionPolicy` doesn't complain with an error. I do think will want to adjust the rules for `ShardGroupDuration` and / or allow the users to modify it themselves, just as they can in 1.x. Probably via CLI tooling and 2.0 OSS specific API

brettbuddin (approved):

stuartcarnie:
@brettbuddin thank you for the thoughtful and thorough review 🙇

PR #19584, opened 2020-09-18
feat: add legendOrientationThreshold

### LINKED ISSUE #17227: Long hover legends push values offscreen with no way to see them
It's impossible to see the value of this legend since it is pushed offscreen. We need a way to limit the width of the hover legend. the styling issue is in #17226 

image

TCL735:
Closes #17227 

Adds a property for the number of columns in the legend (tooltip) that will adjust the direction (horizontal | vertical) when it reaches this threshold.


TCL735:
Ready for review.

pkger/service_test.go
```diff
@@ -2232,21 +2240,22 @@ func TestService(t *testing.T) {
...
+									XColumn:                    "x",
+									YColumn:                    "y",
+									UpperColumn:                "upper",
+									MainColumn:                 "main",
+									LowerColumn:                "lower",
+									TimeFormat:                 "",
+									LegendOrientationThreshold: 5,
 								},
 							},
 						},
```
hoorayimhelping, line 2261: would it be worthwhile to test values other than 5, just for the sake of showing that it's not implemented to work for 5?

edit: or is 5 meaningful? is it possible to make it more meaningful by assigning it to a variable and naming it?

hoorayimhelping (approved):
looks good to me, in general. but if it's possible, it would be good to assign a meaningful value to 5.

TCL735:
Based on a demo of a locally running version, we will also be adding `legendOpacity` to make the tooltip's transparency adjustable, so that the user can see the graph underneath.

Putting this work back into WIP temporarily.

pkger/service_test.go
TCL735: The 5 is meaningless. It's in an assertion that the value should be 5 because the testdata has 5 in the yml and json files. It's similar to what was done with another property, like bin size was arbitrarily chosen as 10 and we assert that we must get 10 for bin size. Same here.

TCL735:
@hoorayimhelping @glinton - Can i please get a re-review on this? I added `legendOpacity`. Thanks.

http/dashboard_test.go
```diff
@@ -448,24 +448,26 @@ func TestService_handleGetDashboard(t *testing.T) {
...
+			"note": "",
+			"queries": null,
+			"shadeBelow": false,
+			"hoverDimension": "",
+			"showNoteWhenEmpty": false,
+			"timeFormat": "",
+			"type": "xy",
+			"xColumn": "",
+			"yColumn": "",
+			"legendOpacity": 0,
```
hoorayimhelping, line 469: does opacity default to 1 or 0?

hoorayimhelping (approved):

http/dashboard_test.go
TCL735: Opacity will default to 1. It is 0 here as a zero value. This is not a test for the "default" value. It is a test of the zero value.

Issue #19585, opened 2020-09-18
influx setup allows creating passwords shorter than 8 chars, but cannot login. 


danielr1996:
__Steps to reproduce:__
List the minimal actions needed to reproduce the behavior.

1. create a fresh install of influxdb and start the service
2. run `influx setup --bucket default --org admin --password adminadmin --username admin --force`
3. go to http://localhost:9999 and login with user admin and password admin

__Expected behavior:__
`influx setup` refuses password not meeting the password policy

__Actual behavior:__
I get the `Could not sign in` popup. and cannot login 

I tried to reset the admin password with `influx user password --id 0653c03959f02000` and I got `Password too short - minimum length is 8 characters!`. After choosing a password longer than 8 chars I could login in the webinterface. 

__Environment info:__

* System info: Linux 5.4.0-1018-raspi aarch64
* InfluxDB version: InfluxDB 2.0.0-beta.16 (git: 50964d732c) build_date: 2020-08-07T20:18:22Z
* Influx CLI version: Influx CLI 2.0.0-beta.16 (git: 50964d732c) build_date: 2020-08-07T20:09:20Z

glinton:
closed by https://github.com/influxdata/influxdb/pull/19593

PR #19593, opened 2020-09-18
fix(cli): don't accept short passwords in setup

glinton:
Addresses #19585

Error if `influx setup` is called and a password shorter than 8 characters is specified.


Asalem1 (commented):
Did you happen to speak to @121watts about this? I think he made a similar change a while back and it lead to some unforeseen issues

cmd/influx/setup.go
```diff
@@ -222,6 +222,10 @@ func onboardingRequest() (*influxdb.OnboardingRequest, error) {
 }
 
 func nonInteractive() (*influxdb.OnboardingRequest, error) {
+	if len(setupFlags.password) < 8 {
```
russorat, line 225: should 8 be defined as a constant that could be changed or configured in the future?

Asalem1 (approved):
LGTM. 🤞🏽

PR #19597, opened 2020-09-21
fix: chronograf/organizations: avoid nil context with WithValue

rogpeppe:
Under later Go versions, the tests panic because it's not OK to pass a nil Context to `context.WithValue`.

- [x] Rebased/mergeable
- [x] Tests pass


stuartcarnie (approved):
🍻 one of the benefits of running newer Go versions is we'll be more prepared for upgrading

PR #19598, opened 2020-09-21
fix: storage: close PointsWriter when Engine is closed

rogpeppe:
The PointsWriter has a Close method which seems like it should
be called when the Engine is shut down.

- [x] Rebased/mergeable
- [x] Tests pass


storage/engine.go
```diff
@@ -222,6 +223,10 @@ func (e *Engine) Close() error {
 		retErr = multierror.Append(retErr, fmt.Errorf("error closing TSDB store: %w", err))
 	}
 
+	if err := e.pointsWriter.Close(); err != nil {
+		retErr = multierror.Append(retErr, fmt.Errorf("error closing points writer: %w", err))
```
jacobmarble, line 227: Looks like this is the only package in this repository which imports github.com/hashicorp/go-multierror, but two other packages import go.uber.org/multierr. Do you mind switching this to go.uber.org/multierr?

jacobmarble (approved):
LGTM

storage/engine.go
rogpeppe: done

jacobmarble (approved):

Issue #19367, opened 2020-08-18
Provide -lponly option for influx_inspect export & generate only line protocol


timhallinflux:
This would allow for for line protocol exports to more easily move between 1.x and 2.x.  

Currently, the user needs to crack the export and remove the DML lines.

dgnorton:
@timhallinflux should this be `-noddl`? DDL creates the database. DML is the data/line-proto.

timhallinflux:
yes... remove all lines that are not LP.  2.0 doesn't accept the various non-LP stuff at the top.

ayang64:
oh -- so it is the other way around?  `-noddl`?

dgnorton:
@ayang64 it means getting rid of any line that isn't line protocol. In other words, remove all the lines starting with `#` and also the `CREATE DATABASE` lines.

@timhallinflux since this command line option is removing all of the DDL and part of the DML, leaving only line protocol, should the flag be `-dataonly` or `-rawdata` maybe?

timhallinflux:
`-lponly` ?

dgnorton:
Implemented in:
master-1.x - PR #19609
1.8.3 - PR #19611

PR #19609, opened 2020-09-22
feat: Add -lponly flag to export sub-command

ayang64:
When applied, this patch will add the -lponly flag to the export command
which instructs influx_inspect to only output line protocol without
comments and other out-of-band data.

Addresses: https://github.com/influxdata/influxdb/issues/19367

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influx_inspect/export/export.go
+++ b/cmd/influx_inspect/export/export.go
@@ -7,6 +7,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"math"
 	"os"
 	"path/filepath"
@@ -36,6 +37,7 @@ type Command struct {
 	startTime       int64
 	endTime         int64
 	compress        bool
+	lponly          bool
 
 	manifest map[string]struct{}
 	tsmFiles map[string][]string
@@ -65,6 +67,7 @@ func (cmd *Command) Run(args ...string) error {
 	fs.StringVar(&cmd.retentionPolicy, "retention", "", "Optional: the retention policy to export (requires -database)")
 	fs.StringVar(&start, "start", "", "Optional: the start time to export (RFC3339 format)")
 	fs.StringVar(&end, "end", "", "Optional: the end time to export (RFC3339 format)")
+	fs.BoolVar(&cmd.lponly, "lponly", false, "Only export line protocol")
 	fs.BoolVar(&cmd.compress, "compress", false, "Compress the output")
 
 	fs.SetOutput(cmd.Stdout)
@@ -123,6 +126,7 @@ func (cmd *Command) export() error {
 	if err := cmd.walkWALFiles(); err != nil {
 		return err
 	}
+
 	return cmd.write()
 }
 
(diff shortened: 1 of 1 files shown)


cmd/influx_inspect/export/export.go
```diff
@@ -209,42 +275,27 @@ func (cmd *Command) write() error {
...
-			fmt.Fprintln(cmd.Stdout, "complete.")
-		}
-		if _, ok := cmd.walFiles[key]; ok {
-			fmt.Fprintf(cmd.Stdout, "writing out wal file data for %s...", key)
-			if err := cmd.writeWALFiles(w, cmd.walFiles[key], key); err != nil {
-				return err
-			}
-			fmt.Fprintln(cmd.Stdout, "complete.")
+	// mw is our "meta writer" -- the io.Writer to which meta/out-of-band data
+	// like comments will be sent.  If the lponly flag is set, we mw will be
```
dgnorton, line 279: nit: `we mw` should be `the mw`

cmd/influx_inspect/export/export.go
```diff
@@ -209,42 +275,27 @@ func (cmd *Command) write() error {
...
+	// data..
+	//
+	mw := func() io.Writer {
+		if cmd.lponly {
+			return ioutil.Discard
 		}
-	}
-	return nil
+		return w
+	}()
```
dgnorton: Saves a couple lines of code and simpler...
```
mw := w
if cmd.lponly {
   mw = ioutil.Discard
}
```

cmd/influx_inspect/export/export.go
```diff
@@ -187,64 +191,111 @@ func (cmd *Command) walkWALFiles() error {
...
+	// Otherwise, mw is set to the same writer as the actual DDL and line
+	// protocol DML which will cause the comments to be intermixed with the
+	// data..
+	//
+	mw := func() io.Writer {
+		if cmd.lponly {
+			return ioutil.Discard
+		}
+		return w
+	}()
```
dgnorton, line 292: Saves a couple lines of code and simpler...
```
mw := w
if cmd.lponly {
   mw = ioutil.Discard
}
```

dgnorton (changes requested):
Mostly looks good. Couple small things.

Have you tested this to confirm it generates just line protocol?

ayang64:
> Mostly looks good. Couple small things.
> 
> Have you tested this to confirm it generates just line protocol?

Yup.  Works for me.

cmd/influx_inspect/export/export.go
ayang64: Done.  Just so you know, I did think about this.  My intention here is that I feel like with:

```
mw := w
if cmd.lponly {
  mw = ioutil.Discard
}
```
as you read it, you have to track the value of `mw` through the condition and the variable name is intermixed with the logic for assigning itself.   Personally, I like assigning `mw` only once and keeping all of the logic that dictates is value separate; it feels easier to read to me.  So I think about it -- if maybe came to a different conclusion.

cmd/influx_inspect/export/export.go
ayang64: d'oh -- fixed!

dgnorton (approved):

PR #19613, opened 2020-09-22
fix(cli): user may be created without password

alespour:
`influx` should allow to create users within organization without password. This is requested eg. for users migrated from 1.x. As of now, when `--password` is not specified, `influx` creates the user but returns error code as it tries to set password. This PR fixes that.

- [x] Tests pass

### DIFF
--- a/cmd/influx/user.go
+++ b/cmd/influx/user.go
@@ -209,8 +209,10 @@ func (b *cmdUserBuilder) cmdCreateRunEFn(*cobra.Command, []string) error {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


stuartcarnie:
@alespour this code looks fine, however, I need to verify this is valid behavior.

@timhallinflux @russorat can you confirm it is valid in 2.0 that a user may be created without a password?

russorat:
a user should always have a password in 2.0, even though i think the apis allow for one to be created without.

alespour:
This is a result of my upgrade process misunderstanding.

PR #19622, opened 2020-09-23
fix(tsdb): exclude the stop time from the array cursor

### LINKED ISSUE #19564: Pushed down range stop bound is not longer exclusive
The test [range_stop_test.flux](https://github.com/influxdata/flux/blob/1ca64e34e0a18ef7dbdb5f1f113535059d27e1fe/stdlib/universe/range_stop_test.flux) is failing against post-split against Flux v0.83.1. It looks like storage is treating the upper bound of the requested time range as inclusive. This test was developed explicitly to catch this case.

This test will be skipped here:
https://github.com/influxdata/influxdb/blob/31f4b1a403b0dc5865e8534a8c6774b9c7731303/query/stdlib/testing/testing.go#L3

We should fix storage so that it has the correct behavior, or update the Flux source to filter points that are on the bound. And then we should un-skip this test.

jsternberg:
This is a backport of #14262 to the 1.x storage engine. The 1.x storage
engine is now the primary engine for open source so when we switched we
regressed to the old behavior.

This also fixes `go generate` for the tsm1 package by running `tmpl`
with `go run` instead of assuming the correct one is installed in the
path.

Fixes #19564.

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -39,9 +39,14 @@ import (
 	"go.uber.org/zap"
 )
 
+//go:generate -command tmpl go run github.com/benbjohnson/tmpl
 //go:generate tmpl -data=@iterator.gen.go.tmpldata iterator.gen.go.tmpl engine.gen.go.tmpl array_cursor.gen.go.tmpl array_cursor_iterator.gen.go.tmpl
-//go:generate go run ../../../_tools/tmpl/main.go -i -data=file_store.gen.go.tmpldata file_store.gen.go.tmpl=file_store.gen.go
-//go:generate go run ../../../_tools/tmpl/main.go -i -d isArray=y -data=file_store.gen.go.tmpldata file_store.gen.go.tmpl=file_store_array.gen.go
+// The file store generate uses a custom modified tmpl
+// to support adding templated data from the command line.
+// This can probably be worked into the upstream tmpl
+// but isn't at the moment.
+//go:generate go run ../../../tools/tmpl -i -data=file_store.gen.go.tmpldata file_store.gen.go.tmpl=file_store.gen.go
+//go:generate go run ../../../tools/tmpl -i -d isArray=y -data=file_store.gen.go.tmpldata file_store.gen.go.tmpl=file_store_array.gen.go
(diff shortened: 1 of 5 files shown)


tsdb/engine/tsm1/engine.go
```diff
@@ -39,9 +39,10 @@ import (
 	"go.uber.org/zap"
 )
 
+//go:generate -command tmpl go run github.com/benbjohnson/tmpl
 //go:generate tmpl -data=@iterator.gen.go.tmpldata iterator.gen.go.tmpl engine.gen.go.tmpl array_cursor.gen.go.tmpl array_cursor_iterator.gen.go.tmpl
-//go:generate go run ../../../_tools/tmpl/main.go -i -data=file_store.gen.go.tmpldata file_store.gen.go.tmpl=file_store.gen.go
-//go:generate go run ../../../_tools/tmpl/main.go -i -d isArray=y -data=file_store.gen.go.tmpldata file_store.gen.go.tmpl=file_store_array.gen.go
+//go:generate go run ../../../internal/cmd/tmpl -i -data=file_store.gen.go.tmpldata file_store.gen.go.tmpl=file_store.gen.go
+//go:generate go run ../../../internal/cmd/tmpl -i -d isArray=y -data=file_store.gen.go.tmpldata file_store.gen.go.tmpl=file_store_array.gen.go
```
wolffcm, line 45: So some of this code generation needs to happen with the version of `tmpl` in `github.com/benbjohnson/tmpl` and some needs the modified version specific to this repo, I guess.

What are the differences between the two versions, do you know? Can you document this in a comment somewhere?

jsternberg: Yes I can. The difference seems to be that the modified tmpl seems to support adding extra data (the `isArray=y`). I think it would be possible to bring this into the mainline `tmpl`, but I didn't want to do that as part of this PR.

jsternberg: It looks like this version of `tmpl` was already in the repo at `tools` and I didn't see it so I'll switch to using that one so we don't have multiple copies.

wolffcm (approved):
Looks good.

PR #19640, opened 2020-09-24
feat: turn on community templates

hoorayimhelping:
Flips template flag to be on to match cloud behavior.

- [x] Rebased/mergeable
- [x] Tests pass


drdelambre:
why aren't we just removing the flag?

Asalem1 (approved):
Congratulations

hoorayimhelping:
>why aren't we just removing the flag?

@drdelambre it's only been flagged on a day, we want to reserve the right to revert it if we discover a catastrophe

PR #19644, opened 2020-09-25
feat: add high level parameters to dashboard that can be passed to charts/queries

glinton:
**TL;DR** Shelving this implementation for a bit.

---------------------------------

The following template defines individual parameters per chart query. My though was to add the ability to define dashboard-wide parameters intended for use inside chart queries to limit the number of unique definitions required for the parameters.

**Original parameterized template:**
```yml

apiVersion: influxdata.com/v2alpha1
kind: Dashboard
metadata:
    name: linux-system
spec:
    name: Linux System
    charts:
        queries:
          - query: |-
                option params = {
                  bucket: "v.bucket"
                }

                from(bucket: params.bucket)
                  |> filter(fn: (r) => r._measurement == "system")
                  |> filter(fn: (r) => r._field == "uptime")
                  |> last()
                  |> map(fn: (r) => ({ _value: float(v: r._value) / 86400.00 }))
        suffix: ' days'
```

**Diff of new functionality:**
```diff
7a8,11
>     params:
>       - key: bucket
>         default: "v.bucket"
>         type: string
11,14d14
<                 option params = {
<                   bucket: "v.bucket"
<                 }
< 
```

This idea falls apart when implementing, flux complains that params is not defined, suggesting that the fix is to export the template with the parent params injected into each child query... which would then bring us full circle to needing to define each parameter in each query.

---------------------------

Another implementation idea is to parameterize a variable, then reference that variable in the dashboard chart queries where it is intended to be re-used in a set-once, use-many setting.

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/pkger/parser.go
+++ b/pkger/parser.go
@@ -860,7 +860,6 @@ func (p *Template) graphLabels() *parseErr {
 
 func (p *Template) graphChecks() *parseErr {
 	p.mChecks = make(map[string]*check)
-	// todo: what is the business goal wrt having unique names? (currently duplicates are allowed)
 	tracker := p.trackNames(false)
 
 	checkKinds := []struct {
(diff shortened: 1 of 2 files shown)


lesam:
@glinton can we close this?

PR #19649, opened 2020-09-27
fix(engine) Jump out of the loop when err is nil

### LINKED ISSUE #19841: The createsnapshot() of backup() will be executed repeatedly when err is nil
__Steps to reproduce:__

1. Analysis code Engine.Backup()

__Expected behavior:__
Jump out of the loop when err is nil

__Actual behavior:__
no break when err is nil

__Environment info:__

__Config:__

__Logs:__

__Performance:__

```sh
iostat -xd 1 30 > iostat.txt
```

zddr:
close #19841 
Add judgment logic

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -915,18 +915,17 @@ func (e *Engine) Free() error {
 func (e *Engine) Backup(w io.Writer, basePath string, since time.Time) error {
 	var err error
 	var path string
+LOOP:
 	for i := 0; i < 3; i++ {
 		path, err = e.CreateSnapshot()
-		if err == nil {
-			break
-		} else if err != nil {
-			switch err {
-			case ErrSnapshotInProgress:
-				backoff := time.Duration(math.Pow(32, float64(i))) * time.Millisecond
-				time.Sleep(backoff)
-			default:
-				return err
-			}
+		switch {
+		case err == nil:
+			break LOOP
+		case errors.Is(err, ErrSnapshotInProgress):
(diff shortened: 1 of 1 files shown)


zddr:
@stuartcarnie

tsdb/engine/tsm1/engine.go
```diff
@@ -917,7 +917,9 @@ func (e *Engine) Backup(w io.Writer, basePath string, since time.Time) error {
 	var path string
 	for i := 0; i < 3; i++ {
 		path, err = e.CreateSnapshot()
-		if err != nil {
+		if err == nil {
+			break
+		} else if err != nil {
```
stuartcarnie, line 922: Thanks for the improvement!

The `else if err != nil` is unnecessary with your change and possibly clearer as:

```go
switch {
case err == nil:
	break LOOP
case errors.Is(err, ErrSnapshotInProgress):
	backoff := time.Duration(math.Pow(32, float64(i))) * time.Millisecond
	time.Sleep(backoff)
default:
	return err
}
```

stuartcarnie (changes requested):
Great catch, however, @benbjohnson might want to look, as he may be actively working in this area.

tsdb/engine/tsm1/engine.go
zddr: Good idea. I've changed it

lesam:
Fixed a while ago in https://github.com/influxdata/influxdb/pull/20527, closing this one.

PR #19641, opened 2020-09-25
feat: extending upgrade command skeleton

vlastahajek:
This PR is 1st of 4 upgrade feature PRs and prepares common code for others.  

Mainly introduces cli interface and options checks.

Minimal command line includes parameters for creating admin: `influxd upgrade -b my-bucket -o my-org -u my-user -p my-password`

Complete help:
```
    Upgrades a 1.x version of InfluxDB by performing following actions:
      1. Reads 1.x config file and creates 2.x config file with matching options. Unsupported 1.x options are reported.
      2. Copies and upgrades 1.x database files.
      3. Creates a script for creating 1.x users and their permissions. This scripts needs to be revised and run manually after starting 2.x.

    If config file is not available, 1.x db folder (--v1-dir options) is taken as an input.
    Target 2.x database dir is specified by the --engine-path option. If changed, bolt path should to be changed as well.

Usage:
  influxd upgrade [flags]
  influxd upgrade [command]

Available Commands:
  v1-dump-meta Dump InfluxDB 1.x meta.db
  v2-dump-meta Dump InfluxDB 2.x influxd.bolt

Flags:
  -m, --bolt-path string         path for boltdb database (default "/home/ubuntu/.influxdbv2/influxd.bolt")
  -b, --bucket string            primary bucket name
      --config-file string       optional: Custom InfluxDB 1.x config file path, else the default config file
  -e, --engine-path string       path for persistent engine files (default "/home/ubuntu/.influxdbv2/engine")
  -h, --help                     help for upgrade
      --log-path string          optional: custom log file path (default "/home/ubuntu/upgrade.log")
  -o, --org string               primary organization name
  -p, --password string          password for username
  -r, --retention string         optional: duration bucket will retain data. 0 is infinite. Default is 0.
      --security-script string   optional: generated security upgrade script path (default "/home/ubuntu/influxd-upgrade-security.sh")
  -t, --token string             optional: token for username, else auto-generated
  -u, --username string          primary username
      --v1-dir string            path to source 1.x db directory containing meta,data and wal sub-folders (default "/home/ubuntu/.influxdb")
  -v, --verbose                  verbose output (default true)
```

V1 and v2 dump commands are also extended about more info.

PR #19651, opened 2020-09-28
feat(upgrade): upgrade security

alespour:
This is 4th PR of 4  for the upgrade feature.
Generates script which upgrades users to 2.x. The script uses `influx` tool to create tokens with read/write permission to buckets that were created for 1.x databases. A token represents 1.x user, 1.x user name becomes the description of the token.

_Note: It is branched from PR #19641 which should be merged first to avoid conflicts._ 

- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/cmd/influxd/upgrade/security.go
+++ b/cmd/influxd/upgrade/security.go
@@ -1,11 +1,315 @@
 package upgrade
 
+// Security upgrade implementation.
+// Generates script for upgrading 1.x users into tokens in 2.x.
+
 import (
 	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"runtime"
+	"sort"
+	"strings"
+	"text/template"
+
+	"github.com/influxdata/influxdb/v2/v1/services/meta"
+	"github.com/influxdata/influxql"
 	"go.uber.org/zap"
 )
 
-// generateSecurityScript generates security upgrade script.
-func generateSecurityScript(v1 *influxDBv1, dbBuckets map[string][]string, log *zap.Logger) error {
-	return errors.New("not implemented")
+// generateSecurityScript creates security upgrade script with 1.x users mapped to 2.x tokens.
+func generateSecurityScript(v1 *influxDBv1, targetOptions optionsV2, dbBuckets map[string][]string, log *zap.Logger) error {
+	// check if there any 1.x users at all
+	v1meta := v1.meta
+	if len(v1meta.Users()) == 0 {
+		log.Info("There are no users in 1.x, no script will be created.")
+		return nil
+	}
+
+	// get helper instance
+	helper := newSecurityScriptHelper(log)
+
+	// check if target buckets exists in 2.x
+	proceed := helper.checkDbBuckets(v1meta, dbBuckets)
+	if !proceed {
+		return errors.New("upgrade: there were errors/warnings, please fix them and run the command again")
+	}
+
+	// create output
+	var output *os.File
+	var isFileOutput bool
+	if targetOptions.securityScriptPath == "" {
+		output = os.Stdout
+	} else {
+		file, err := os.OpenFile(targetOptions.securityScriptPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0777)
+		if err != nil {
+			return err
+		}
+		defer file.Close()
+		output = file
+		isFileOutput = true
+	}
+
+	// get `influx` path
+	influxExe, err := helper.getInflux2ExePath()
+	if err != nil {
+		return err
+	}
+
+	// template data type
+	type dataObject struct {
+		AdminUsers   []string
+		IgnoreUsers  []string
+		UpgradeUsers []string
+		UserArgs     map[string][]string
+		Exe          string
+		TargetAdmin  string
+		TargetOrg    string
+		TargetToken  string
+		IsFileOutput bool
+		InfluxExe    string
+	}
+
+	// data
+	data := dataObject{
+		UserArgs:     make(map[string][]string),
+		TargetAdmin:  targetOptions.userName,
+		TargetOrg:    targetOptions.orgName,
+		TargetToken:  targetOptions.token,
+		IsFileOutput: isFileOutput,
+		InfluxExe:    influxExe,
+	}
+
+	// fill data with users and their permissions
+	for _, row := range helper.sortUserInfo(v1meta.Users()) {
+		username := row.Name
+		if row.Admin {
+			data.AdminUsers = append(data.AdminUsers, username)
+		} else if len(row.Privileges) == 0 {
+			data.IgnoreUsers = append(data.IgnoreUsers, username)
+		} else {
+			data.UpgradeUsers = append(data.UpgradeUsers, username)
+			dbList := make([]string, 0)
+			for database := range row.Privileges {
+				dbList = append(dbList, database)
+			}
+			sort.Strings(dbList)
+			accessArgs := make([]string, 0)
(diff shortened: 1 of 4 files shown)


cmd/influxd/upgrade/security_test.go
```diff
@@ -0,0 +1,248 @@
...
+			f := reflect.ValueOf(v1.meta).Elem().Field(4)
+			f = reflect.NewAt(f.Type(), unsafe.Pointer(f.UnsafeAddr())).Elem()
+			f.Set(reflect.ValueOf(data))
+
+			// script output file
+			tmpfile, err := ioutil.TempFile("", "upgrade-security-*" + shSuffix)
+			if err != nil {
+				t.Fatal(err)
+			}
+			tmpfile.Close()
```
stuartcarnie, line 132: Go 1.15 now has support for temporary directories that are automatically cleaned up on completion of the test:

https://golang.org/doc/go1.15#testing

```go
tmpFile, err := ioutil.TempFile(t.TempDir(), "upgrade-security-*" + shSuffix)
require.NoError(err)
require.NoError(tmpfile.Close())
```

cmd/influxd/upgrade/security_test.go
```diff
@@ -0,0 +1,248 @@
...
+			if err != nil {
+				t.Fatal(err)
+			}
```
stuartcarnie, line 135: These lines are redundant

cmd/influxd/upgrade/security_test.go
```diff
@@ -0,0 +1,248 @@
...
+			defer os.Remove(tmpfile.Name())
```
stuartcarnie, line 136: Will not be necessary if `t.TempDir()` is used

cmd/influxd/upgrade/security_test.go
```diff
@@ -0,0 +1,248 @@
...
+
+			// options passed on cmdline
+			targetOptions := optionsV2{
+				userName: "admin",
+				orgName: "demo",
+				token: "ABC007==",
+				securityScriptPath: tmpfile.Name(),
+			}
```
stuartcarnie, line 144: `goimports -w`

cmd/influxd/upgrade/security_test.go
```diff
@@ -0,0 +1,248 @@
...
+
+			// command execution
+			err = generateSecurityScript(v1, targetOptions, tc.db2ids, zap.NewNop())
```
stuartcarnie, line 147:
```go
import "go.uber.org/zap/zaptest"
```

and use `zaptest.NewLogger(t)`, which will print the output if the test fails:

```go
err = generateSecurityScript(v1, targetOptions, tc.db2ids, zaptest.NewLogger(t))
```

cmd/influxd/upgrade/security.go
```diff
@@ -1,11 +1,283 @@
...
+		UserArgs:     make(map[string][]string),
+		TargetAdmin:  targetOptions.userName,
+		TargetOrg:    targetOptions.orgName,
+		TargetToken:  targetOptions.token,
+		IsFileOutput: isFileOutput,
+		InfluxExe:    influxExe,
+	}
+
+	// fill data with users and their permissions
+	for _, row := range helper.sortUserInfo(v1meta.Users()) {
```
stuartcarnie, line 92: This is good that the user data is sorted, to ensure consistent ordering in the script output

cmd/influxd/upgrade/security_test.go
```diff
@@ -0,0 +1,248 @@
...
+
+			// validate result by comparing arrays of non-empty lines of wanted vs actual output
+			parse := func(content string) []string {
+				var lines []string
+				scanner := bufio.NewScanner(strings.NewReader(content))
+				for scanner.Scan() {
+					line := strings.TrimSpace(scanner.Text())
+					if line != "" {
+						// In `influx create auth` command line, read/write bucket args may be in different order,
+						// because user privileges is a map. So reorder them alphabetically for easier comparison.
```
stuartcarnie, line 166: This is not true, as the user privileges is keyed in a map, but stored in a slice, which has a consistent ordering.

Can this code be simplified based on that information?

stuartcarnie (changes requested):
Looking really good! I have mostly nits to correct and a couple of questions.

cmd/influxd/upgrade/security_test.go
alespour: Thank you! I did not know this, I've also missed the `go` directive change to 1.15 in `go.mod`.

cmd/influxd/upgrade/security_test.go
alespour: The comment is not good, I've modified it. The order of bucket IDs is random on the generated command line because order of `database` names in iteration over `meta.UserInfo.Privileges` map in random (https://github.com/influxdata/influxdb/blob/af20f784b2765604fc10eb1012b93405b7087138/cmd/influxd/upgrade/security.go#L96). So buckets permissions has to be sorted to get the same line every test run.

cmd/influxd/upgrade/security.go
```diff
@@ -1,11 +1,283 @@
...
+		output = file
+		isFileOutput = true
+	}
+
+	// get `influx` absolute path
+	exePath, err := os.Executable()
+	if err != nil {
+		return err
+	}
+	influxExe := filepath.Join(filepath.Dir(exePath), "influx")
```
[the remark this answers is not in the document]
alespour: The doc for `exec.Command.Run()` says that `"The returned error is nil if the command runs, has no problems copying stdin, stdout, and stderr, and exits with a zero exit status."` so I guess it is good enough just to check that return error is not `nil`, right?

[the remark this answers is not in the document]
stuartcarnie: Yes, that should be fine.

cmd/influxd/upgrade/security_test.go
stuartcarnie: The overarching goal is the script output should be deterministic.

Ultimately that will require sorting any source data to ensure that property is maintained.

alespour: Now the list authorizations is based on sorted database names list (lines 95-101 in https://github.com/influxdata/influxdb/blob/f6cf893ba6dcd6e940ec6e6c8865b82a480b39df/cmd/influxd/upgrade/security.go#L95).

stuartcarnie (changes requested):
Everything looks great – just one more change and we can get this merged!

cmd/influxd/upgrade/upgrade.go
```diff
@@ -196,6 +206,12 @@ func NewCommand() *cobra.Command {
 			Default: influxConfigPathV1(),
 			Desc:    "optional: Custom InfluxDB 1.x config file path, else the default config file",
 		},
+		{
+			DestP:   &options.target.influx2CommandPath,
+			Flag:    "influx-command-path",
+			Default: influxExePathV2(),
+			Desc:    "optional: influx command path",
```
stuartcarnie, line 213:
```suggestion
			Desc:    "Path to influx command",
```

stuartcarnie (changes requested):
One nit for the help of the `influx-command-path` and then we can merge it

cmd/influxd/upgrade/upgrade.go
alespour: Done.

stuartcarnie (approved):
Looks great – thank you!

stuartcarnie:
@alespour can you rebase your PR on master, as I see other commits to master are passing.

PR #19654, opened 2020-09-28
chore: kit/io: improve LimitedReadCloser

rogpeppe:
A fairly minor change, but this saves two allocations every time
points are written to the API (one allocation for the embedded io.LimitReader,
and one allocation to create the `close` closure).

Also fix the code so that it actually limits to the exact requested number of bytes
rather than one more. We don't really need to layer on top of io.LimitReader,
as that code is fairly minimal.

This change also means that the `LimitedReadCloser` can be used without a constructor
function, the same way that `io.LimitedReader` can.

- [x] Rebased/mergeable
- [ ] Tests pass


stuartcarnie (approved):
Always love these types of improvements 💯

PR #19655, opened 2020-09-28
feat: generate modern profiles

ayang64:
Prior to this commit, influxd was writing legacy profiling data which
often (always?) required an accompanying executable to use.

This commit instructs influxd to write profiles in the new format which
can be examined without a binary.

Closes: #19681

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/services/httpd/pprof.go
+++ b/services/httpd/pprof.go
@@ -121,25 +121,28 @@ func (h *Handler) archiveProfilesAndQueries(w http.ResponseWriter, r *http.Reque
 	//
 	if vals, exists := r.Form["trace"]; exists {
 		// parse the duration encoded in the last "trace" value supplied.
-		duration, err := time.ParseDuration(last(vals))
+		val := last(vals)
+		duration, err := time.ParseDuration(val)
 
 		// If we can't parse the duration or if the user supplies a negative
-		// number, then set it to a reasonable default of 10s.
+		// number, return an appropriate error status and message.
 		//
-		// This allows for stuff like ?trace=true -- which isn't recommened but
-		// makes the code do something reasonable.
-		//
-		// If someone wants to supply a trace duration of 0s I guess that's their
-		// perogative but lets ensure that it is positive.
-		if defaultDuration := time.Second * 10; err != nil || duration < 0 {
-			duration = defaultDuration
+		// In this case it is a StatusBadRequest (400) since the problem is in the
+		// supplied form data.
+		if duration < 0 {
+			http.Error(w, fmt.Sprintf("negative trace durations not allowed"), http.StatusBadRequest)
+			return
+		}
+
+		if err != nil {
+			http.Error(w, fmt.Sprintf("could not parse supplied duration for trace %q", val), http.StatusBadRequest)
+			return
 		}
 
 		// Trace files can get big.  Lets clamp the maximum trace duration to 45s.
 		if maxDuration := time.Second * 45; duration > maxDuration {
 			duration = maxDuration
 		}
-
 		profiles = append(profiles, prof{"trace", duration})
 	}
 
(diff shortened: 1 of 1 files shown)


docmerlin (dismissed):

services/httpd/pprof.go
```diff
@@ -113,15 +116,15 @@ func (h *Handler) archiveProfilesAndQueries(w http.ResponseWriter, r *http.Reque
...
+			if err := prof.WriteTo(&buf, 0); err != nil {
 				http.Error(w, err.Error(), http.StatusInternalServerError)
 				return
 			}
 		}
 
 		// Write the profile file's header.
 		err := tw.WriteHeader(&tar.Header{
-			Name: profile.Name + ".txt",
+			Name: path.Join("profiles", profile.Name+".pb.gz"),
```
dgnorton, line 127: Since we're now instructing the profiler to generate the modern format (non-legacy), which is gzipped, does this `tw` still need to wrap a `gzip.Writer` on line 99 above?

ayang64: ya -- we could probably not gzip the entire tarball.

ayang64: done!

services/httpd/pprof.go
```diff
@@ -68,60 +62,173 @@ type prof struct {
...
+		{Name: "threadcreate"},
 	}
 
-	// Capture a CPU profile?
-	if r.FormValue("cpu") != "" {
-		profile := &prof{Name: "cpu"}
+	// We parse the form here so that we can use the http.Request.Form map.
+	//
+	// Otherwise we'd have to use r.FormValue() which makes it impossible to
+	// distinuish between a form value that exists and has no value and a one
```
dgnorton, line 84: nit: "and a one"

services/httpd/pprof.go
```diff
@@ -68,60 +62,173 @@ type prof struct {
...
+	//
+	// Requests for a trace should look like:
+	//
+	//  ?trace=10s
+	//
+	if vals, exists := r.Form["trace"]; exists {
+		// parse the duration encoded in the last "trace" value supplied.
+		duration, err := time.ParseDuration(last(vals))
+
+		// If we can't parse the duration or if the users supplies a negative
```
dgnorton, line 126: nit: "if the users supplies"

services/httpd/pprof.go
```diff
@@ -68,60 +62,173 @@ type prof struct {
...
+		// number, then set it to a reasonable default of 10s.
+		//
+		// This allows for stuff like ?trace=true -- which isn't recommened but
+		// makes the code do something reasonable.
+		//
+		// If someone wants to supply a trace duration of 0s I guess that's their
+		// perogative but lets ensure that it is positive.
+		if defaultDuration := time.Second * 10; err != nil || duration < 0 {
```
dgnorton, line 134: If the user supplies a negative or zero duration, it should return an error. They might have a bug in a script that requests profiles or traces and not be collecting what they expected to collect. It should error on that so they see the bug instead of masking it.

services/httpd/pprof.go
```diff
@@ -68,60 +62,173 @@ type prof struct {
...
+			duration, _ = time.ParseDuration(val)
+		}
+
+		// ensure that the duration is positive
+		if defaultDuration := time.Second * 30; duration < 0 {
+			duration = defaultDuration
+		}
+
+		// ensure that profile duration is less than 1 minute
+		if maxDuration := time.Minute * 1; duration > maxDuration {
```
dgnorton, line 184: Was there a max duration for profiles before? I'm not seeing it at a glance. If not, 1 minute might be too restrictive. Five minutes maybe? Although, I'm not sure we should have a max.

dgnorton (changes requested):
A few small things. Mostly looks good.

services/httpd/pprof.go
ayang64: makes sense.

services/httpd/pprof.go
ayang64: there wasn't a max.  the standard was 30 seconds.  i figured some kind of restriction is better than none.  but maybe if you're poking this endpoint you know what you're doing.

ayang64: okay -- removed.  i think it *is* reasonable to limit trace duration to 45 seconds.  i figure 30 seconds is probably more than anyone would want and could digest since the traces are so dense.  then i added 50% to allow for a little extra -- so 45 seconds of trace.

services/httpd/pprof.go
ayang64: fixed.

services/httpd/pprof.go
ayang64: fixed.

services/httpd/pprof.go
ayang64: fixed.

dgnorton (approved):

PR #19658, opened 2020-09-29
fix: add ability to define bucket to write to with a BucketFilter

### LINKED ISSUE #19643: cli write command should not require read permissions on a token
i was trying to write data to a bucket using the CLI and a token that had `write` permission only to that bucket. the cli complained that the bucket didn't exist.

https://github.com/influxdata/influxdb/blob/master/cmd/influx/write.go#L288

i think the cli is trying to read the bucket, before writing to it, but it shouldn't need to. we should remove that constraint.

glinton:
Closes #19643

`influx write` was doing a bucket lookup to get the org and bucket id. This pr removes that lookup by implementing a more flexible write method (`WriteTo`) that accepts a `BucketFilter` and sets the query parameters accordingly.


bucket.go
```diff
@@ -117,7 +117,7 @@ func (f BucketFilter) QueryParams() map[string][]string {
 	}
 
 	if f.Name != nil {
-		qp["name"] = []string{*f.Name}
+		qp["bucket"] = []string{*f.Name}
```
glinton, line 120: Seems `name` was a typo before and it appears nothing is relying on it.

zoesteinkamp (approved):
Looks good!

PR #19662, opened 2020-09-29
fix: bufio.Scanner: token too long

### LINKED ISSUE #19586: cli write: Error: Failed to write data: bufio.Scanner: token too long
I'm seeing the error:
`Error: Failed to write data: bufio.Scanner: token too long.`

when i try to write a csv data that is ~164MB to influxdb. Based on https://stackoverflow.com/questions/21124327/how-to-read-a-text-file-line-by-line-in-go-when-some-lines-are-long-enough-to-ca , it looks like we might need to evaluate if we should switch to bufio.Reader? https://github.com/influxdata/influxdb/blob/f09ee881fb33ca06474ca499deac0c5cd1bd9f91/write/batcher.go#L67

I can send the file offline. 

```
➜  influxdb git:(master) ✗ influx query -c tools 'from(bucket: "apps") |> range(start: -1m)' -r | influx write -c default --format csv -b apps
Error: Failed to write data: bufio.Scanner: token too long.
```

stuartcarnie:
Closes #19586

This PR fixes two issues. 

* Fixes a panic that was introduced with a new `LineReader` type.
* Adds an option to override the maximum line length in `write.Batcher` to support line lengths greater than the default 64KiB for a new `bufio.Scanner`
* Adds a CLI argument, `max-line-length`, to configure the maximum line length. The default is 16MB, which should cover the majority of use cases.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/write/batcher.go
+++ b/write/batcher.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"time"
@@ -18,6 +19,11 @@ const (
 	DefaultInterval = 10 * time.Second
 )
 
+var (
+	// ErrLineTooLong is the error returned when reading a line that exceeds MaxLineLength.
+	ErrLineTooLong = errors.New("batcher: line too long")
+)
+
 // batcher is a write service that batches for another write service.
 var _ platform.WriteService = (*Batcher)(nil)
 
@@ -25,6 +31,7 @@ var _ platform.WriteService = (*Batcher)(nil)
 type Batcher struct {
 	MaxFlushBytes    int                   // MaxFlushBytes is the maximum number of bytes to buffer before flushing
 	MaxFlushInterval time.Duration         // MaxFlushInterval is the maximum amount of time to wait before flushing
+	MaxLineLength    int                   // MaxLineLength specifies the maximum length of a single line
 	Service          platform.WriteService // Service receives batches flushed from Batcher.
 }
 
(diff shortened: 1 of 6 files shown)


write/batcher.go
```diff
@@ -59,13 +66,29 @@ func (b *Batcher) Write(ctx context.Context, org, bucket platform.ID, r io.Reade
...
 	scanner := bufio.NewScanner(r)
 	scanner.Split(ScanLines)
+
+	maxLineLength := bufio.MaxScanTokenSize
+	if b.MaxLineLength > 0 {
+		maxLineLength = b.MaxLineLength
+	}
+
+	// limit the initial size of the buffer to a maximum of bufio.MaxScanTokenSize bytes
+	scanner.Buffer(make([]byte, min(bufio.MaxScanTokenSize, maxLineLength)), maxLineLength)
```
rogpeppe, line 90: Or just `scanner.Buffer(nil, maxLineLength)` and let the buffer be grown as needed when lines are read. That way when all lines are short (presumably a common case) we'll use a lot less memory.

rogpeppe (commented):
I'm afraid I didn't have time to finish looking through the code before eod, but here's one small comment in the meantime... (looks good otherwise as far as I saw so far!)

write/batcher.go
stuartcarnie: Good feedback – consider it done!

brettbuddin (approved):
Nice find!

e-dard (approved):
LGTM 👍

PR #19691, opened 2020-10-06
feat: allow disable compaction per shard

dgnorton:
This feature allows compaction to be disabled on a per-shard basis by
creating a file named `do_not_compact` in a shard's directory. When
disabled, a message is logged every 15 minutes with the reason for
compaction being disabled (existance of the file). This makes it easy to
know if compaction has been disabled for any shards by searching the log
for `compaction disabled` or running `find path/to/data -type f -name
do_not_compact`.

Example log message:
```
2020-10-06T13:12:02.200598Z     info    TSM compaction disabled {"log_id": "0PgWKQ1W000", "engine": "tsm1", "db_shard_id": 9, "reason": "/home/dgnorton/.influxdb/data/stress/autogen/9/do_not_compact"}
```

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -87,6 +87,9 @@ const (
 
 	// deleteFlushThreshold is the size in bytes of a batch of series keys to delete.
 	deleteFlushThreshold = 50 * 1024 * 1024
+
+	// DoNotCompactFile is the name of the file that disables compactions.
+	DoNotCompactFile = "do_not_compact"
 )
 
 // Statistics gathered by the engine.
@@ -2055,6 +2058,8 @@ func (e *Engine) compact(wg *sync.WaitGroup) {
 	t := time.NewTicker(time.Second)
 	defer t.Stop()
 
+	var nextDisabledMsg time.Time
+
 	for {
 		e.mu.RLock()
 		quit := e.done
@@ -2065,6 +2070,17 @@ func (e *Engine) compact(wg *sync.WaitGroup) {
 			return
 
 		case <-t.C:
+			// See if compactions are disabled.
+			doNotCompactFile := filepath.Join(e.Path(), DoNotCompactFile)
+			_, err := os.Stat(doNotCompactFile)
+			if err == nil {
+				now := time.Now()
+				if now.After(nextDisabledMsg) {
+					e.logger.Info("TSM compaction disabled", logger.Shard(e.id), zap.String("reason", doNotCompactFile))
+					nextDisabledMsg = now.Add(time.Minute * 15)
+				}
+				continue
+			}
 
 			// Find our compaction plans
 			level1Groups := e.CompactionPlan.PlanLevel(1)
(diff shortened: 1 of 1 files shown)


tsdb/engine/tsm1/engine.go
```diff
@@ -2065,6 +2070,17 @@ func (e *Engine) compact(wg *sync.WaitGroup) {
...
+			doNotCompactFile := filepath.Join(e.Path(), DoNotCompactFile)
+			_, err := os.Stat(doNotCompactFile)
+			if err == nil {
+				now := time.Now()
+				if now.After(nextDisabledMsg) {
+					e.logger.Info("TSM compaction disabled", logger.Shard(e.id), zap.String("reason", doNotCompactFile))
+					nextDisabledMsg = now.Add(time.Minute * 15)
+				}
+				continue
+			}
```
jacobmarble, line 2083: It would help to add a comment like `} // err != nil -> file doesn't exist -> do compact`

jacobmarble (approved):

tsdb/engine/tsm1/engine.go
```diff
@@ -2065,6 +2070,17 @@ func (e *Engine) compact(wg *sync.WaitGroup) {
 			return
 
 		case <-t.C:
+			// See if compactions are disabled.
```
ayang64, line 2073: i'd think about this in terms of `lastDisableMsg` instead of `nextDisableMsg`.

this reads a bit more clearly.  instead of asking, "is now after the nextDisabledMessage" we can ask, "has it been 15 mins since `lastDisabledMsg`?"

```
  6       // See if compactions are disabled.
     5       doNotCompactFile := filepath.Join(e.Path(), DoNotCompactFile)
     4       _, err := os.Stat(doNotCompactFile)
~    3       if err == nil && time.Since(lastDisabledMsg) > time.Minute*15 {
~    2         e.logger.Info("TSM compaction disabled", logger.Shard(e.id), zap.String("reason", doNotCompactFile))
~_   1         lastDisabledMsg = time.Now()
  2079         continue
     1       }

```

or something like that.

this seems clearer, saves an indent lev

ayang64: also, i'd probably change the name of `doNotCompatFile` and `DoNotCompactFile` to a noun like, `compactionDisableFile` and `CompactionDisableFile`.

ayang64 (approved):

tsdb/engine/tsm1/engine.go
jacobmarble: I prefer to keep names private until they to be made public. It's easy to capitalize as part of some other PR that needs access.

ayang64: i agree.  there's already `DoNotCompactFile` though -- was trying to make minimal changes.

ayang64: this feels like bike shedding though. :)

philjb:
Seems ok to me! But it also seems like an undocumented debug tool that will be infrequently used while the code is looking for the signal file once a second for all time for all shards. I do appreciate the log message to help with disabled compactions being "left on" by accident.

PR #19699, opened 2020-10-07
feat(values): add window mean array cursor

fchikwekwe:
Adds the ArrayCursor implementation for window mean. 

Also bumps the flux version v0.83.2 => v0.83.3 to import a duration conversion helper.


wolffcm (commented):
This looks really nice. Can you also move this test over as well?
https://github.com/influxdata/idpe/blob/master/storage/reads/array_cursor_test.go#L1921-L2018

storage/reads/array_cursor.go
```diff
@@ -16,18 +17,24 @@ func (v *singleValue) Value(key string) (interface{}, bool) {
...
-		return newSumArrayCursor(cursor)
+		return newSumArrayCursor(cursor), nil
 	case datatypes.AggregateTypeCount:
-		return newCountArrayCursor(cursor)
+		return newCountArrayCursor(cursor), nil
+	case datatypes.AggregateTypeMean:
+		return newWindowMeanArrayCursor(cursor, window)
 	default:
 		// TODO(sgc): should be validated higher up
+		// TODO(faith): add other aggs to this switch as we port them over
```
rockstar, line 37: Isn't this kind of implied by the `panic`? As we add more aggregates, we'll hit this panic, which would show us we need to add it?

fchikwekwe:
@wolffcm Yeah! Just added it. I was looking for blindspots. Appreciate it.

rockstar (approved):
Thanks for shepherding this.

storage/reads/array_cursor.go
fchikwekwe: Yeah, that's true. I guess it was a mental note for me more than anything. I can remove it.

wolffcm (approved):

PR #19700, opened 2020-10-07
fix(templates): preserve cell colors on export/import

### LINKED ISSUE #19441: UI cell configuration incomplete in template with CLI export/import
When a dashboard with single stat cell is exported/imported using CLI `export/apply` commands as template, cell config fields `id` and `value` are missing in `colors` element and the cell coloring does not work as intended when imported later. The same issue seems to be with other view types such as tables (thresholds configuration is broken after import) etc.

InfluxDB 2.0 beta-16

__Steps to reproduce:__

1. import the attached dashboard in Chronograf/Dashboards  - the single stat cell shows number in purple (`"star"`) color
2. export the dashboard with `influx export --dashboards=$ID --file=xyz.json`
3. import it back using `influx apply --file=xyz.json`

Exported template is missing `id` and `value`  fields in single stat `colors` config section. Even if added manually, it appears to be ignored during import.

__Expected behavior:__
Dashboard should show a number in purple (`"star"`) color

__Actual behavior:__
Dashboard shows the number in default (`"laser"`) color and cannot be changed.

[test.json.zip](https://github.com/influxdata/influxdb/files/5129805/test.json.zip)

glinton:
Closes #19441

Something relies on the ID of the cell's color to be set (seems like `"base"` is a common id). This PR preserves that color ID accross import/export so one can now export and re-import a dashboard and any custom colors in cells should go unchanged.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/pkger/parser_models.go
+++ b/pkger/parser_models.go
@@ -977,6 +977,7 @@ const (
 )
 
 type color struct {
+	ID   string `json:"id,omitempty" yaml:"id,omitempty"`
 	Name string `json:"name,omitempty" yaml:"name,omitempty"`
 	Type string `json:"type,omitempty" yaml:"type,omitempty"`
 	Hex  string `json:"hex,omitempty" yaml:"hex,omitempty"`
@@ -1002,6 +1003,7 @@ func (c colors) influxViewColors() []influxdb.ViewColor {
 	var iColors []influxdb.ViewColor
 	for _, cc := range c {
 		iColors = append(iColors, influxdb.ViewColor{
+			ID:    cc.ID,
 			Type:  cc.Type,
 			Hex:   cc.Hex,
 			Name:  cc.Name,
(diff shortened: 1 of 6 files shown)


pkger/parser_models.go
```diff
@@ -977,6 +977,7 @@ const (
 )
 
 type color struct {
+	ID   string `json:"id,omitempty" yaml:"id,omitempty"`
```
Asalem1, line 980: just checking that there shouldn't be a situation where the ID would empty / 0, right?

Asalem1 (approved):
Quick question. Also, should the swagger file be updated with this?

pkger/parser_models.go
glinton: swagger shouldn't need to be updated as this is only preserving the already present color id. it was previously getting dropped and it appears the ui relies on it (no console errors though) to be set to show that color (it's been `base` in every dashboard i looked at).

glinton: if it is empty, it would be when `apply`ing an old `export` where the id was not set with the color information. the color would still appear broken because i didn't default empty id's to `"base"`, but if the user fixes the color and re-exports, the color would be preserved by this change.

PR #19721, opened 2020-10-10
fix(kv): Reorganize bcrypt password hashing outside transactions

### LINKED ISSUE #19720: Password hashing occurs inside the BoltDB transaction
## What

The processing of the password hash should occur outside the key/value store transaction.

### Background

When performing various password operations on the key/value store, hashing of the password occurs _within_ the BoltDB transaction:

https://github.com/influxdata/influxdb/blob/96d84b9126dd2d7bac78a0504e4805e28e4503ee/kv/passwords.go#L75-L99

This increases the duration for which a read or write lock is held against the key / value store, which may lead to unnecessary contention.

stuartcarnie:
Closes #19720

This ensures that transaction lifetimes are shorter to reduce lock contention.

Additionally, this PR removes the ability to set an empty password when running the initial on-boarding process. According to the original developer, this behavior was required to permit Cloud processes to complete. As this code is no longer shared, this security issue is corrected.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/kv/onboarding.go
+++ b/kv/onboarding.go
@@ -124,15 +124,19 @@ func (s *Service) OnboardInitialUser(ctx context.Context, req *influxdb.Onboardi
 		Token:       req.Token,
 	}
 
+	var passwordHash []byte
+	passwordHash, err = s.generatePasswordHash(req.Password)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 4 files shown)


lyondhill (approved):
Lgtm. It's nice to drop that old dependency

PR #18410, opened 2020-06-09
fix(handler): allow CORS in v2 compatibility endpoints

sranka:
Closes #18391

Describe your proposed changes here.
* add CORS handling so that the web browser clients can access v2 compatibility endpoints as they already can against InfluxDB v2 (#17640)
* cherry-pick and merge /health v2 compatibility endpoint to the master-1.x branch from 1.8 branch (#17252)

See influxdata/influxdb-client-js#199 to know more.

- [x] [CHANGELOG.md](https://github.com/influxdata/influxdb/blob/master/CHANGELOG.md) updated with a link to the PR (not the Issue)
- [x] [Well-formatted commit messages](https://www.conventionalcommits.org/en/v1.0.0-beta.3/)
- [x] Rebased/mergeable
- [x] Tests pass
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)

PR #19727, opened 2020-10-12
fix(http): allow CORS in v2 compatibility endpoints

### LINKED ISSUE #19715: CORS fail with query v2 compatibility API
__Description:__
I am noticing a strange behaviour when querying InfluxDB 1.8.3 with Flux, from an Angular web app that uses `@influxdata/influxdb-client`. In my particular case, my web app is accessible at `http://localhost:4200` while InfluxDB is hosted on another machine at `http://192.168.225.5:8086`.
The query request fails because of CORS policy, since this is an actual cross-origin request. But should it?
According to issue influxdata/influxdb#18391 and the consequent pull request, CORS should be correctly handled by InfluxDB and I supposedly should not be receiving this error:
image
In the network tab, this is the situation:
image
HTTP pre-flight OPTIONS requests look like this:
image

I also tried reverse-proxying InfluxDB through nginx to bypass CORS using [this configuration](https://enable-cors.org/server_nginx.html): at first it seemed to work (the OPTIONS pre-flight request was successful) but any subsequent POST included duplicate Access-Control-* headers (both those returned by InfluxDB and those added by nginx).
Network tab:
image
Bypassed pre-flight OPTIONS request:
image
POST response, including duplicate Access-Control-* headers:
image

This tells me that CORS is indeed handled correctly by InfluxDB, if not for pre-flight requests. That is, the OPTIONS requests on `/api/v2/query` should not return 405 Method Not Allowed but a 2xx success response including the relevant Access-Control-* headers.

__Steps to reproduce:__
You could try creating an Angular web app using `@influxdata/influxdb-client`, but a simple web page including a script that launches an XMLHttpRequest to InfluxDB `/api/v2/query` should be sufficient to trigger the browser's CORS checks.

__Expected behavior:__
Calls to `/api/v2/query` should respond correctly to OPTIONS requests and not fail CORS policy checks.

__Actual behavior:__
OPTIONS pre-flight request to `/api/v2/query` responds with 405 Method Not Allowed and fails CORS policy checks.

__Environment info:__

* System info: Linux 4.19.15-1.ph3-esx x86_64
* InfluxDB version: 1.8.3, Docker image influxdb:1.8 sha256:35de8cc24dfca8ad80fc93bba1bc7125240b3e27b457a1ef529aac5fb6217eef
* Other relevant environment details: Docker 18.06.1 on VMware PhotonOS 3.0

__Config:__
Only customization is `INFLUXDB_HTTP_FLUX_ENABLED=true` environment variable.

sranka:
Closes #19715 ... contains cherry-picks from master1.x branch ( #18410)

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -219,8 +219,8 @@ func NewHandler(c Config) *Handler {
 			"status-head",
 			"HEAD", "/status", false, true, authWrapper(h.serveStatus),
 		},
-		Route{ // Ping
-			"ping",
+		Route{ // Health
+			"health",
 			"GET", "/health", false, true, authWrapper(h.serveHealth),
 		},
 		Route{
@@ -1865,6 +1865,7 @@ func cors(inner http.Handler) http.Handler {
 				`Authorization`,
 				`Content-Length`,
 				`Content-Type`,
+				`User-Agent`,
 				`X-CSRF-Token`,
 				`X-HTTP-Method-Override`,
 			}, ", "))
--- a/services/httpd/handler.go
+++ b/services/httpd/handler.go
@@ -195,6 +195,10 @@ func NewHandler(c Config) *Handler {
 			"write", // Data-ingest route.
 			"POST", "/api/v2/write", true, writeLogEnabled, h.serveWriteV2,
 		},
+		Route{ // Enable CORS
+			"write-options",
+			"OPTIONS", "/api/v2/write", false, true, h.serveOptions,
+		},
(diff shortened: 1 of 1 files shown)


ajaxjiang96:
Is there anyone still facing the CORS issue?

sranka:
The fix is already part of **v1.9.3**  https://github.com/influxdata/influxdb/blob/v1.9.3/services/httpd/handler.go proves it.

PR #19758, opened 2020-10-16
fix(swagger): add token to OnboardingRequest

bednar:
Users wants do onboarding the InfluxDB by our clients (https://github.com/influxdata/influxdb-client-java/issues/165), but the token is missing in the `OnboardingRequest` definition in swagger:

```yml
OnboardingRequest:
  type: object
  properties:
    username:
      type: string
    password:
      type: string
    org:
      type: string
    bucket:
      type: string
    retentionPeriodHrs:
      type: integer
  required:
    - username
    - org
    - bucket
```

The corresponding structure in `go`:
https://github.com/influxdata/influxdb/blob/8c0ff75c6df0148e1699c67f3b873b78dd160aef/onboarding.go#L27

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -10843,6 +10843,8 @@ components:
           type: string
         retentionPeriodHrs:
           type: integer
+        token:
+          type: string
       required:
         - username
         - org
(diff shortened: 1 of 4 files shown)


hoorayimhelping (approved):
lgtm, i'd wait until @lyondhill looks it over since he's more familiar with go conventions

lyondhill:
I believe it is a mistake to allow user defined tokens to be passed in through the api. I would suggest removing "token" from the structure rather than adding it in swagger. @stuartcarnie should probably make a decision (I no longer work at influx so I shouldn't decide)

bednar:
@stuartcarnie What do you think about this? Could we keep the possibility to set up token via API?

docmerlin:
I think tokens are things that should be given by the API not created by the user (as opposed to passwords).  That is my take on it.

PR #19773, opened 2020-10-19
fix(upgrade): fallback to user's home when saving upgraded config

### LINKED ISSUE #19771: 1.x upgrade: make config upgrade more robust
`/etc/influxdb/influxdb.conf` is owned by root and when a user runs `upgrade` command as any other user - which is usual situation - it fails with `Error: open /etc/influxdb/influxdb.toml: permission denied` when it attempts to save the new config to the same folder.

alespour:
Closes #19771

Also removes pointless code that backups already existing 2.x config file: when user would want to perform upgrade again, it is expected that such config is useless. anyway.

- [X] Rebased/mergeable
- [X] Tests pass
- [X] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)


stuartcarnie (approved):
LGTM – thanks for the improvements!

PR #19780, opened 2020-10-20
fix(upgrade): proper default 2.x config filename

alespour:
Current config upgrade saves upgraded configuration to `influxdb.toml`,  but 2.x looks for `config.json|toml|yaml` by default. This tiny PR fixes it. 
Also the related test has been slightly improved and simplified.

- [X] Rebased/mergeable
- [X] Tests pass


stuartcarnie (approved):
Good find!

PR #19783, opened 2020-10-20
chore(kv): delete deprecated kv service code

### LINKED ISSUE #18498: Remove references to tenant services from kv package

GeorgeMac:
Closes #18498

This includes removal of a lot of kv.Service responsibilities. However,
it does not finish the re-wiring.
It removes:
- documents
- telegrafs
- notification rules + endpoints
- checks
- orgs
- users
- buckets
- passwords,
- URMs
- labels
- authorizations.

There are some outstanding pieces that are needed to get kv service compiling (dashboard service urm dependency). Then all the call sites for kv service need updating and the new implementations of telegraf and notification rules + endpoints needed installing (along with any necessary migrations).

- [ ] Rebased/mergeable
- [ ] Tests pass


kv/task.go
```diff
@@ -133,41 +134,23 @@ func (s *Service) findTaskByID(ctx context.Context, tx Tx, id influxdb.ID) (*inf
 		t.LatestCompleted = t.CreatedAt
 	}
 
-	// Attempt to fill in the owner ID based on the auth.
```
GeorgeMac, line 136: note for reader:
https://github.com/influxdata/influxdb/blob/master/kv/migration/all/0003_task_owners.go#L15
This should not be nil anymore as a migration was delivered to backfill missing owners a while ago.

kv/task.go
```diff
@@ -863,12 +815,6 @@ func (s *Service) deleteTask(ctx context.Context, tx Tx, id influxdb.ID) error {
 		return influxdb.ErrUnexpectedTaskBucketErr(err)
 	}
 
-	if err := s.deleteUserResourceMapping(ctx, tx, influxdb.UserResourceMappingFilter{
```
GeorgeMac, line 866: We don't create URMs anymore for tasks (like this anyway). So we don't need this.
This _may_ leave some dangling URMs, but they are harmless cruft.

GeorgeMac:
Second pass: https://github.com/influxdata/influxdb/pull/19878

PR #19790, opened 2020-10-20
fix(authorization): Adds authorizationType property to authorizations

### LINKED ISSUE #19763: Extend authorizations service to support new authorization types
## What

To enable seamless upgrades from InfluxDB OSS 1.x to 2.x using `username:password` tokens, create a new token type which allows a user-specified token (key). 

### Add `AuthorizationType`

`AuthorizationType` is an enumerated string, that will be used in a number of places, similar to the `influxdb.Status` field:

```go
type AuthorizationType string

const (
	AuthorizationTypePlain  AuthorizationType = "plain"
	AuthorizationTypeV1User AuthorizationType = "v1_user"
)
```

### Extend the `Authorization` struct

Extend the `/api/v2/authorizations` API to include an `authorizationType` property that is of type `AuthorizationType`. The default value for existing tokens should be `plain` given they are stored in plain text. When creating authorizations, this is also the default `authorization_type`, to avoid breaking existing clients. The `authorizationType` for the v1 APIs should be `v1_user_token`, to be explicit about their use and expected format.

A client will be able to create a new authorization by specifying the `authorization_type` property as `v1_user` and a `token` property with a value formatted as `username:password`. It MUST be in a format that is parseable by BASIC auth for validation purposes and should be validated before being persisted. When persisted to the key/value store, the `token` should be hashed using a cryptographically secure, one-way hash. SHA256 is recommended unless there is a strong case for another.

This token type will only be used by 1.x APIs (`/write` and `/query`) either via BASIC auth:

https://github.com/influxdata/influxdb/blob/8c0ff75c6df0148e1699c67f3b873b78dd160aef/http/legacy/influx1x_authentication_handler.go#L149-L155

or the `u` (username) and `p` (password) query parameters:

https://github.com/influxdata/influxdb/blob/8c0ff75c6df0148e1699c67f3b873b78dd160aef/http/legacy/influx1x_authentication_handler.go#L123-L129

### Update the `AuthorizationService` `CreateAuthorization` API

The `CreateAuthorization` API should check the `Type` enum and if it is `v1_user`, the `Token` field should be SHA256 hashed to use as the index key. The `Token` field should be cleared before the JSON is marshaled.

It should also be validated for minimum length. The minimum password length is defined by a constant:

https://github.com/influxdata/influxdb/blob/d28e55d3b69013c7d42aae33c35332d02fc24837/kv/passwords.go#L13

### Update the Find APIs

The Find APIs should be updated to search for specific token types and to find tokens by their specific type.

Update `FindAuthorizationByToken` signature to include the `AuthorizationType`:

https://github.com/influxdata/influxdb/blob/f646653b1b12f1b9adcbae83612b8a5624b190a5/auth.go#L97-L98

```go
FindAuthorizationByToken(ctx context.Context, t string, authType AuthorizationType) (*Authorization, error)
```

Add a `Type *AuthorizationType` field to the `AuthorizationFilter` struct:

https://github.com/influxdata/influxdb/blob/f646653b1b12f1b9adcbae83612b8a5624b190a5/auth.go#L114-L124

and update the implementation of the Find API to limit the results if the `Type` field is specified

### Update swagger.yaml

Add `authorization_type` property to `Authorization` schema:

```yaml
            authorizationType:
              readOnly: true
              type: string
              default: plain
              enum:
                - plain
                - user_key
```

Add new `PostAuthorizationRequest` to provide explicit schema for `POST /api/v2/authorizations`:

```yaml
    PostAuthorizationRequest:
      properties:
        status:
          type: string
          default: active
          enum:
            - active
            - inactive
        orgID:
          type: string
        userID:
          type: string
          nullable: true
        description:
          type: string
          nullable: true
        permissions:
          type: array
          minItems: 1
          description: List of permissions for an auth.  An auth must have at least one Permission.
          items:
            $ref: "#/components/schemas/Permission"
        authorizationType:
          type: string
          default: plain
          enum:
            - plain
            - user_key
        token:
          type: string
          nullable: true
```

### Update HTTP handlers

The HTTP handlers should be updated to return the `authorizationType` property in all responses. The `POST` handler should support deserializing and validating requests that include the `authorizationType` property.

stuartcarnie:
Closes #19763

This PR combines some cleanup with the primary goal of adding a new field, `authorizationType`, to the `authorizations` API. In doing so, it is possible to distinguish the default, plain-text tokens with a new type for authorizing InfluxDB API 1.x compatible requests.

### Summary

The new authorization type, `v1_user` is nothing special within the storage layer or service layer of `authorization` package. By the time the `influxdb.Authorization` token is being persisted, it is just another field. 

However, when _creating_ authorization tokens, some behavior has changed:

* The default behavior is to create `plain` tokens. That means if the `authorizationType` field is absent, the default value and semantics are to create a `plain` token. Plain tokens will automatically generate a `token` value using CSPRNG as before
* If `authorizationType` is `v1_user`
    * the `token` field must be set. The API will error if it is not.
    * the set value will be hashed using SHA256 before being stored in the K/V store, as it is not intended to be retrievable after initial creation
* unit tests validate this HTTP API behavior

When fetching authorization tokens,

* those of type `v1_user` will return no value for the `token` property
* may be filtered by type, specifying `authorizationType=(plain,v1_user)` query parameter

### Changes

* Add field, `Type`  to `Authorization`, that persists to the key/value store. It is a string enumeration with two possible values (`plain` and `v1_user`) and defaults to `plain`. `plain` indicates the authorization stores its `token` as plain text.
* `AuthorizationFilter` has a new field to filter by type
* Fix the authorization filter predicate function to properly handle conjunctions
* HTTP requests correctly marshal, unmarshal and validate the `authorizationType` field
* Update swagger definition to include the `authorizationType` property as a string enum, with the set of possible values
* Update existing unit tests to pass with new field
* Add new unit tests to verify the behavior of the HTTP API to create authorizations

- [x] Rebased/mergeable
- [x] Tests pass


stuartcarnie:
Hi @zoesteinkamp, the Swagger has been updated

@GeorgeMac I tagged you on this, as I believe you'll have the most context to review it. I've separated into multiple commits.

PR #19802, opened 2020-10-22
chore(docs): update PostDBRP docs to reflect mutual exclusive requirement of org vs orgID

GeorgeMac:
I am unsure if I am doing this correctly. But this updates the PostDBRP docs to reflect the mutually exclusive requirement of `org` vs `orgID` when creating a DBRP mapping. Since only one or the other is actually required to create the mapping. But the docs currently state both are required.

My knowledge of OpenAPI Spec is a little lacking. But this is what I assertain is correct based on good old SO:

https://stackoverflow.com/questions/21134029/how-to-define-mutually-exclusive-query-parameters-in-swagger-openapi/49199240#49199240

- [x] ~[CHANGELOG.md](https://github.com/influxdata/influxdb/blob/master/CHANGELOG.md) updated with a link to the PR (not the Issue)~
- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -11704,12 +11704,7 @@ components:
       type: string
       enum: ["slack", "pagerduty", "http", "telegram"]
     DBRP:
-      required:
-        - orgID
-        - org
-        - bucketID
-        - database
-        - retention_policy
+      type: object
       properties:
         id:
           type: string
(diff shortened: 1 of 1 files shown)


rbetts (approved):
That was less intuitive than I imagined! Thanks for linking the SO answer :joy:.

GeorgeMac:
@rbetts I know right? I stumbled across some old issues which said it wasn't possible in OAPI 2.0 and was concerned I would have to just settle on a description. But 🤞 this will do it. Thanks for the review!

TCL735 (approved):

sranka:
@GeorgeMac FYI : This PR did the job that is compliant with the spec. The influxdb UI's oats (typescript code generator) generates useless `export type DBRP = any | any`, ... I will fix it for [influxdata/influxdb-client-js](https://github.com/influxdata/influxdb-client-js) . Other API generators will likely create a type/class with all properties being optional.

PR #19804, opened 2020-10-22
refactor(notification): move rule service into own package

### LINKED ISSUE #19798: Move Notification Rule service into own package

We need to isolate notification rule service definitions into own package (outside `kv`) and remove `URM` interactions.

## Success Criteria

- We can rebase the referenced PR and remove need for `*kv.Service` to implement notification rule service.

GeorgeMac:
Closes #19798

This creates the separate (non-URM interaction) notification rule service implementation. It is mostly a cut/paste of the kv service as it is today. Without URM interaction. Which we have shown to not be necessary in order to operate properly.

This doesn't remove the original definition in `*kv.Service` yet.
That will be delivered in a later PR (likely in the larger https://github.com/influxdata/influxdb/pull/19783).

Update:

A few extra things had to be delivered to put a bow on this:

TL;DRs
1. Launcher helpers updated to use notification rule client>
2. Fix added to notification rule client for bug when finding rules by ID.
3. Task http client changed to match service definition interface.
4. Remove the notification rule functionality from kv service.
5. Added new but deprecated `resource.OrgIDResolver` to sever the responsibility from `*kv.Service`.

Exhaustive explanations:

1. Introducing the changed notification rule service to the launcher broke pkger tests. Why? Because it turned out pkger tests were using the launcher helpers and the launcher helpers were often referring directly to the embedded kv service instead of going through the client to the daemon instance itself. I updated the launcher helper to return the client.
2. When returning the client this caused a panic in pkger itself. Further investigation showed that the notification rule client `FindNotificationRuleByID` did not work. Because of some poor unmarshalling of the convoluted json format this endpoint produces, it accidentally failed to unmarshall the embedded rule, resulting in a nil rule and a nil error. Referring to this rule down the line caused a panic.
3. The launcher helper used the task http client type directly (not the service abstraction), because the task client was never actually implemented to match the task service (inconsistent with every other service). So I fixed it up to be consistent and return the domain types.
4. TL;DR Remove the notification rule functionality from kv service. The new rule service no longer exposes an embedded urm or org service, which breaks the old functional style tests in `testing`. So the best option was to just remove those definitions now we have ported them next to the service itself (in `notification/rule/service` package). Doing so left the kv service notification rule service without tests. Since this is deprecating that service to ultimately be removed in the coming week, I decided to just remove it as a part of this PR.
5. The *kv.Service depended on implementing all the different service interfaces in order to support org ID resolution for some other services (see FindOrgIDForResource). This has now been moved into own type.

- [x] Rebased/mergeable
- [x] Tests pass


gavincabbage (approved):
My only nitpick is the stuttering of `ruleservice.NewRuleService` but that's not worth worrying about, just personal preference 😄

stuartcarnie (commented):
This looks good to me. I am tempted to be devil's advocate are ponder if we should keep this code in the same package as Cloud, however, I then convince myself it is a fallacy to think this code will / should remain the same between the two systems.

Once the tests pass, I give it the 🆗

GeorgeMac:
@stuartcarnie I agree. I should have noted this actually. Putting it in the same place leads to the following packages:
```
notification
    endpoint
notifications
    endpoints
```
Which I couldn't bring myself to do. So I moved them flat into the existing:
```
notification
    endpoint
```
But then this created a cycle in dependencies because the services I ported depend on `kv` and `kv` depends (in OSS) on these packages for the domain types.
Long story short :joy: I want to flatten this all later once `kv` definitions are removed. Since `kv` will no longer depend on these notification specific packages.

http/notification_rule.go
```diff
@@ -163,21 +163,41 @@ type notificationRuleResponse struct {
...
+	Labels          []influxdb.Label      `json:"labels"`
+	Links           notificationRuleLinks `json:"links"`
+	Status          string                `json:"status"`
+	LatestCompleted time.Time             `json:"latestCompleted,omitempty"`
+	LatestScheduled time.Time             `json:"latestScheduled,omitempty"`
+	LastRunStatus   string                `json:"lastRunStatus,omitempty"`
+	LastRunError    string                `json:"lastRunError,omitempty"`
+}
+
+func (resp *notificationRuleResponse) UnmarshalJSON(v []byte) (err error) {
```
GeorgeMac, line 176: Note for reader:
The unmarshalling here never worked. Leading this client to result in a `nil` NotificationRule and a `nil` error when positively finding a rule.

GeorgeMac: This was causing a panic when this client was used for pkger tests.

GeorgeMac:
@stuartcarnie and @gavincabbage I updated this PR in a number of ways. See the update in the description. As I kept pulling threads I found a bunch of fun traps. Which the PR should now resolve.

authorizer/urm.go
```diff
@@ -6,19 +6,19 @@ import (
 	"github.com/influxdata/influxdb/v2"
 )
 
-type OrganizationService interface {
+type OrgIDResolver interface {
```
stuartcarnie, line 9: I'm not objecting to the name, but throwing out `OrgIDFinder` as an alternative, in case you like it and the verb is **Find**ResourceOrganizationID.

cmd/influxd/launcher/launcher.go
```diff
@@ -1169,7 +1197,7 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 		SecretService:                   secretSvc,
 		LookupService:                   lookupSvc,
 		DocumentService:                 m.kvService,
-		OrgLookupService:                m.kvService,
+		OrgLookupService:                orgIDResolver,
```
stuartcarnie, line 1200: Seeing ~m.kvService~ makes me so happy :)

stuartcarnie (approved):
Great updates George – I can appreciate this is an arduous process, but your work to eliminate the `kv.Service` is very valued and appreciated 💯

authorizer/urm.go
GeorgeMac: Good shout. I actually rename this again in two PRs time to just `Resolver` (in the Telegraf PR). But I will rename it Finder in that PR as it finds both org IDs and name 👍

PR #19807, opened 2020-10-22
feat(storage): enable window agg mean pushdown

fchikwekwe:
Final issue to port window agg mean and bare agg mean pushdown to OSS. 

As discussed below, this PR will also remove capabilities and consolidate the `Store` and `StoreReader` interfaces as well as any mock interfaces for testing.

### DIFF
--- a/cmd/influxd/launcher/query_test.go
+++ b/cmd/influxd/launcher/query_test.go
@@ -953,13 +953,13 @@ error2","query plan",109,110
 }
 
 func TestQueryPushDowns(t *testing.T) {
-	t.Skip("Not supported yet")
 	testcases := []struct {
 		name  string
 		data  []string
 		query string
 		op    string
 		want  string
+		skip  string
 	}{
 		{
 			name: "range last single point start time",
@@ -979,6 +979,7 @@ from(bucket: v.bucket)
 ,result,table,_start,_stop,_time,_value,_field,_measurement,tag
 ,,0,1970-01-01T00:00:00.000000001Z,1970-01-01T01:00:00Z,1970-01-01T00:00:00.000000001Z,1,f,m,a
 `,
+			skip: "https://github.com/influxdata/idpe/issues/8828",
 		},
 		{
 			name: "window last",
@@ -1018,6 +1019,7 @@ from(bucket: v.bucket)
 ,,3,1970-01-01T00:00:12Z,1970-01-01T00:00:15Z,1970-01-01T00:00:14Z,9,f,m0,k0
 ,,4,1970-01-01T00:00:15Z,1970-01-01T00:00:18Z,1970-01-01T00:00:15Z,5,f,m0,k0
 `,
+			skip: "https://github.com/influxdata/idpe/issues/8828",
 		},
 		{
 			name: "window offset last",
@@ -1056,6 +1058,7 @@ from(bucket: v.bucket)
 ,,2,1970-01-01T00:00:11Z,1970-01-01T00:00:14Z,1970-01-01T00:00:13Z,8,f,m0,k0
 ,,3,1970-01-01T00:00:14Z,1970-01-01T00:00:17Z,1970-01-01T00:00:15Z,5,f,m0,k0
 `,
+			skip: "https://github.com/influxdata/idpe/issues/8828",
 		},
 		{
 			name: "bare last",
@@ -1090,6 +1093,7 @@ from(bucket: v.bucket)
 ,result,table,_start,_stop,_time,_value,_field,_measurement,k
 ,,0,1970-01-01T00:00:05Z,1970-01-01T00:00:20Z,1970-01-01T00:00:15Z,5,f,m0,k0
 `,
+			skip: "https://github.com/influxdata/idpe/issues/8828",
 		},
 		{
 			name: "window empty last",
@@ -1135,6 +1139,7 @@ from(bucket: v.bucket)
 #default,_result,2,1970-01-01T01:00:00Z,1970-01-01T02:00:00Z,,,f,m0,k0
 ,result,table,_start,_stop,_time,_value,_field,_measurement,k
 `,
+			skip: "https://github.com/influxdata/idpe/issues/8828",
 		},
 		{
 			name: "window empty offset last",
@@ -1180,6 +1185,7 @@ from(bucket: v.bucket)
 #default,_result,2,1970-01-01T01:00:00Z,1970-01-01T02:00:00Z,,,f,m0,k0
 ,result,table,_start,_stop,_time,_value,_field,_measurement,k
 `,
+			skip: "https://github.com/influxdata/idpe/issues/8828",
 		},
 		{
 			name: "window aggregate last",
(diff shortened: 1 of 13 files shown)


storage/flux/reader.go
```diff
@@ -54,6 +57,31 @@ type storeReader struct {
 	s storage.Store
 }
 
+type qWAC struct{}
```
fchikwekwe, line 60: This was just added to return appropriate capabilities.

storage/flux/reader.go
```diff
@@ -218,6 +246,276 @@ READ:
...
+			},
+			Offset: &datatypes.Duration{
+				Nsecs:    wai.spec.Window.Offset.Nanoseconds(),
+				Months:   wai.spec.Window.Offset.Months(),
+				Negative: wai.spec.Window.Offset.IsNegative(),
+			},
+		}
+	}
+
+	req.Aggregate = make([]*datatypes.Aggregate, len(wai.spec.Aggregates))
```
fchikwekwe, line 302: `datatypes` was substituted for anywhere that `storageproto` was previously used.

v1/services/storage/store.go
```diff
@@ -54,6 +54,56 @@ type Store struct {
 	Logger     *zap.Logger
 }
 
+type readsWAC struct{}
```
fchikwekwe, line 57: added to return enabled capabilities

v1/services/storage/store.go
```diff
@@ -54,6 +54,56 @@ type Store struct {
...
+	var cur reads.SeriesCursor
+	if ic, err := newIndexSeriesCursor(ctx, req.Predicate, s.TSDBStore.Shards(shardIDs)); err != nil {
+		return nil, err
+	} else if ic == nil { // TODO(jeff): this was a typed nil
+		return nil, nil
+	} else {
+		cur = ic
+	}
+
+	return reads.NewWindowAggregateResultSet(ctx, req, cur)
```
fchikwekwe, line 104: changed to follow OSS convention for similar pushdown types

cmd/influxd/launcher/query_test.go
```diff
@@ -949,13 +949,13 @@ error2","query plan",109,110
...
 
 func TestQueryPushDowns(t *testing.T) {
-	t.Skip("Not supported yet")
 	testcases := []struct {
 		name  string
 		data  []string
 		query string
 		op    string
 		want  string
+		skip  string
```
fchikwekwe, line 958: Unskipped mean tests and left all other skipped for now. There are not tests for monthly durations yet as that improvement has not yet made it to OSS.

jlapacik (commented):
This looks good. Were you planning on adding the pushdown rule too or is that coming in a different PR?

fchikwekwe:
@jlapacik they already existed in oss [here](https://github.com/influxdata/influxdb/blob/feat/enable-mean-pushdown/query/stdlib/influxdata/influxdb/rules.go#L782), plus the launcher tests indicate that the pushdown is occurring. Unless you're saying that i'm missing something else?

jlapacik (approved):
Awesome!

query/storage.go
```diff
@@ -96,6 +96,7 @@ type ReadWindowAggregateSpec struct {
 	Aggregates  []plan.ProcedureKind
 	CreateEmpty bool
 	TimeColumn  string
+	Window      execute.Window
```
yzhang1991, line 99: Why do we need this? Does it duplicate `WindowEvery` and `Offset` in this struct?  I don't see the value of this gets assigned anywhere in the code?

storage/flux/reader.go
yzhang1991: Why do we need this duplicated capability struct? It is exactly the same as `readsWAC`.

To take one step back, I think we do not need to have the capability system in OSS at all since it does not have separate queryd and storaged services and operate everything in single node.

storage/reads/store.go
```diff
@@ -84,3 +85,18 @@ type Store interface {
...
 	GetSource(orgID, bucketID uint64) proto.Message
 }
+
+// WindowAggregateCapability describes what is supported by WindowAggregateStore.
+type WindowAggregateCapability interface {
+	query.WindowAggregateCapability
+}
+
+// WindowAggregateStore implements the WindowAggregate capability.
+type WindowAggregateStore interface {
```
yzhang1991, line 95: I think we do not need to make a separate WindowAggregateStore for OSS. All those designs were for rolling upgrade in idpe.

query/storage.go
fchikwekwe: Oh, yeah that's a remnant from the duration support update. We will need it in the future for when we add monthly durations, but with some revision I can remove it for now and we can add it back when we make that change.

fchikwekwe:
@yzhang1991 yeah, the team discussed we decided to backtrack on leaving the capabilities in for now. We also decided that it makes sense to consolidate the Store interface with the WindowAggregateStore interface. I'll go ahead and make that change.

query/storage.go
fchikwekwe: And for the record, it was used in the added code. it was used to create durations like here: https://github.com/influxdata/influxdb/blob/f34cee7199384e52eaf8dd98664ab072883395a4/storage/flux/reader.go#L278-L300

But I'll be removing that since the duration support update was not an intentional part of this PR.

jlapacik: I'm fine if we keep this since it's just going to be added anyway when we enable the rest of the pushdowns.

fchikwekwe: We don't actually have a current issue for adding duration support right now I don't think.

fchikwekwe: Or are you saying that as we port over all the existing code, we'll get this revision for free as long as we don't remove these bits?

jlapacik: Yes that's correct.

fchikwekwe: I'm fine with leaving it then. @yzhang1991 ?

fchikwekwe: @jlapacik in that case, I think that leaving duration support in place and enabling those tests should be added to the DOD for finalizing the other pushdowns.

query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -680,69 +680,19 @@ func (rule PushDownWindowAggregateRule) Pattern() plan.Pattern {
 }
 
 func canPushWindowedAggregate(ctx context.Context, fnNode plan.Node) bool {
```
fchikwekwe, line 682: We're only pushing down mean as a part of this PR. Cases can be added back for each agg as it is ported.

query/stdlib/influxdata/influxdb/rules_test.go
```diff
@@ -56,26 +52,41 @@ func (c mockGroupCapability) HaveLast() bool  { return c.last }
...
-func (m mockWAC) HaveLast() bool   { return m.Have }
-func (m mockWAC) HaveOffset() bool { return m.Have }
-
 func fluxTime(t int64) flux.Time {
 	return flux.Time{
 		Absolute: time.Unix(0, t).UTC(),
 	}
 }
 
+var skipTests = map[string]string{
```
fchikwekwe, line 61: @jlapacik I had to skip these tests once I changed the planner rules. You can remove them from the map as they are enabled.

jlapacik (approved):

PR #19813, opened 2020-10-23
feat: aggregate array cursors

jlapacik:

### DIFF
--- a/storage/reads/array_cursor.go
+++ b/storage/reads/array_cursor.go
@@ -18,6 +18,10 @@ func (v *singleValue) Value(key string) (interface{}, bool) {
 }
 
 func newAggregateArrayCursor(ctx context.Context, agg *datatypes.Aggregate, cursor cursors.Cursor) (cursors.Cursor, error) {
+	switch agg.Type {
+	case datatypes.AggregateTypeFirst, datatypes.AggregateTypeLast:
+		return newLimitArrayCursor(cursor), nil
+	}
 	return newWindowAggregateArrayCursor(ctx, agg, execute.Window{}, cursor)
 }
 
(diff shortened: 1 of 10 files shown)


storage/reads/array_cursor.go
```diff
@@ -38,131 +51,16 @@ func newWindowAggregateArrayCursor(ctx context.Context, agg *datatypes.Aggregate
...
-			return &m.cursors.b
-		default:
-			panic(fmt.Sprintf("unreachable: %T", cur))
-		}
-	}
-	// if cursor is nil
-	return nil
-}
-
-func newSumArrayCursor(cur cursors.Cursor) cursors.Cursor {
```
jlapacik, line 122: This is subsumed by WindowAggregateArrayCursor now.

storage/reads/array_cursor.go
```diff
@@ -38,131 +51,16 @@ func newWindowAggregateArrayCursor(ctx context.Context, agg *datatypes.Aggregate
...
-		return newIntegerArraySumCursor(cur)
-	case cursors.UnsignedArrayCursor:
-		return newUnsignedArraySumCursor(cur)
-	default:
-		// TODO(sgc): propagate an error instead?
-		return nil
-	}
-}
-
-func newCountArrayCursor(cur cursors.Cursor) cursors.Cursor {
```
jlapacik, line 136: This is also subsumed by WindowAggregateArrayCursor now.

storage/reads/array_cursor.go
```diff
@@ -38,131 +51,16 @@ func newWindowAggregateArrayCursor(ctx context.Context, agg *datatypes.Aggregate
 	}
 }
 
-type arrayCursors struct {
```
jlapacik, line 41: This isn't used at all. OSS uses `multiShardArrayCursors` which is almost identical in structure except for the underlying series cursor.

storage/reads/array_cursor.go
```diff
@@ -38,131 +51,16 @@ func newWindowAggregateArrayCursor(ctx context.Context, agg *datatypes.Aggregate
...
+	itrs cursors.CursorIterators
+	err  error
 }
 
 type multiShardArrayCursors struct {
-	ctx   context.Context
-	limit int64
-	req   cursors.CursorRequest
+	ctx context.Context
+	req cursors.CursorRequest
```
jlapacik, line 63: Aggregation is now handled by WindowAggregateArrayCursor, so there's no need to calculate summary stats like count and limit here.

jsternberg (approved):

PR #19815, opened 2020-10-24
feat(authorization): Create a v1 authorization service

### LINKED ISSUE #19812: Create V1 Authorizations service and HTTP API
## What

To enable seamless upgrades from InfluxDB OSS 1.x to 2.x using `username:password` tokens, create a new V1 authorization service which allows a user-specified token (key).

## How

* Create a new `v1/authorization` package
* Create a service to provide CRUD operations for storing `influxdb.Authorization` tokens
* This service should store these tokens in a separate bucket
* Create a HTTP API that is mounted to a private route (i.e. outside the swagger)
    * This API should provide standard create, read and delete operations

stuartcarnie:
Closes #19812

This PR introduces a new package, `v1/authorization`, that provides an implementation of `influxdb.AuthorizationService` for managing v1 authorization tokens. It is a private API, exposed via `/private/legacy/authorizations`, with the same operations as `/api/v2/authorizations` for creating, reading, updating and deleting data.

As this is a private API, it is not noted in the `CHANGELOG`.

Note that this commit does not hook up the service to the v1 `/query` and `/write`, which will occur in a subsequent PR .

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influxd/launcher/launcher.go
+++ b/cmd/influxd/launcher/launcher.go
@@ -67,6 +67,7 @@ import (
 	"github.com/influxdata/influxdb/v2/tenant"
 	_ "github.com/influxdata/influxdb/v2/tsdb/engine/tsm1" // needed for tsm1
 	_ "github.com/influxdata/influxdb/v2/tsdb/index/tsi1"  // needed for tsi1
+	authv1 "github.com/influxdata/influxdb/v2/v1/authorization"
 	iqlcoordinator "github.com/influxdata/influxdb/v2/v1/coordinator"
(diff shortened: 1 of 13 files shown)


GeorgeMac (approved):
Looks good Stuart. I appreciate mostly a lift and drop from `authorizations` right?
All the mounting looks correct to me :)

stuartcarnie:
> I appreciate mostly a lift and drop from `authorizations` right?

Correct – I figured that code did mostly what we needed. I considered if there were ways to abstract some of the implementation, but then considered it was likely to cause more problems, if the two diverged.

PR #19816, opened 2020-10-25
fix: Type-convert fs.Bavail for portability

ayang64:
Prior to this patch, DiskUsage() would calculate bytes available
by multiplying blocks available by block size in bytes:

  disk.Avail = fs.Bavail * uint64(fs.Bsize)

Under some versions of Unix, fs.Bavail is of type uint64 and on
others (like FreeBSD) it is of type int64.

This causes a compile time error:
```
  $ go build
  # github.com/influxdata/influxdb/v2/pkg/fs
  ./fs_unix.go:81:25: invalid operation: fs.Bavail * uint64(fs.Bsize) (mismatched types int64 and uint64)
```
This patch type-converts fs.Bavail to unit64 to ensure that all
types in the expression align.

This prevents compile time errors under FreeBSD and other platforms
where fs.Bavail isn't uint64.

Closes #

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/pkg/fs/fs_unix.go
+++ b/pkg/fs/fs_unix.go
@@ -71,14 +71,14 @@ func CreateFile(newpath string) (*os.File, error) {
 
 // DiskUsage returns disk usage of disk of path
 func DiskUsage(path string) (*DiskStatus, error) {
-	var disk DiskStatus
 	fs := unix.Statfs_t{}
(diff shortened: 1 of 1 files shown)


vlastahajek (approved):
Thanks for the fix!

PR #19818, opened 2020-10-26
refactor(notifications): isolate endpoint service

### LINKED ISSUE #19799: Move Notification Endpoint service into own package

We need to isolate notification endpoint service definitions into own package (outside `kv`) and remove `URM` interactions.

## Success Criteria

- We can rebase the referenced PR and remove need for `*kv.Service` to implement notification endpoint service.

GeorgeMac:
Closes #19799 

This PR delivers a similar objective to #19804 in that it migrates a service behavior (notification endpoints) off of the `*kv.Service` and isolates it into its own package `notification/endpoint/service`.

A lot of this work has been delivered and validated in influx cloud. It is mostly a lift and drop of the implementation there, which in itself is a lift a drop of the existing `*kv.Servive` implementation minus interactions with User Resource Mappings.

This severs the dependency on user resource mapping service. This is further supporting the endeavor to stop using `*kv.Service` as a tenant service: https://github.com/influxdata/influxdb/pull/19783

Update:

I realize the new check service was never wired in either. As I have started to integrate parts of tenant in place of KV, checks service tests in `kv` started failing. We have had a checks service defined which was never installed in the launcher. So instead of fixing a deprecated service. This PR now also installs the `checks` package implementation instead. Which is already tested and works without URMs.

- [x] Rebased/mergeable
- [x] Tests pass


stuartcarnie (commented):
Looks good so far, @GeorgeMac – as you mentioned in the PR description, this is largely a lift of source from cloud, so I didn't thoroughly review those additions. A do see a few check / notification tests are still failing, so I'll leave those with you. Once those are passing, it has the  green light from me 🚦

cmd/influxd/launcher/launcher_helpers.go
```diff
@@ -367,10 +367,6 @@ func (tl *TestLauncher) BucketService(tb testing.TB) *http.BucketService {
 	return &http.BucketService{Client: tl.HTTPClient(tb)}
 }
 
-func (tl *TestLauncher) CheckService() influxdb.CheckService {
```
GeorgeMac, line 370: Note for the reader:

You'll see this have moved up onto the Launcher for now. I have been moving more of these helpers to use the clients as opposed to in-process service implementations. I went to move this to the `http.CheckService` implementation. But half the methods don't actually match the `influxdb.CheckService` interface. They return their own local check API domain model and accept slightly different arguments.
I started down the path of fixing this client, but It started to get unwieldy. I still want to deliver that though as its own PR later. So I have made a node to create and issue and do that.
Until then this just updates the launcher test suite to interface with the same check service installed within the launcher itself. So that consistency is at least achieved, even if it doesn't exercise the front door as we would like.

stuartcarnie (approved):
Really exciting progress, @GeorgeMac 💯 

Particularly love this :)

image

PR #19819, opened 2020-10-26
refactor(telegraf): isolate telegraf config service and remove URM interactions

### LINKED ISSUE #19800: Move Telegraf service into own package

We need to isolate notification telegraf service definitions into own package (outside `kv`) and remove `URM` interactions.

Will require a migration to support listing telegrafs by org (without URM).

## Success Criteria

- We can rebase the referenced PR and remove need for `*kv.Service` to implement telegraf service.

GeorgeMac:
Closes #19800 

The purpose of this change is to sever the dependency on the URM service from the `TelegrafConfigService`.
It does need it to function (much like endpoints and rules it misused this service).
This change introduces a new isolated Telegraf package implementation of the service and it removes the behaviors from the `*kv.Service` implementation.
Further, supporting the work to remove behaviors from the `*kv.Service`: https://github.com/influxdata/influxdb/pull/19783

UPDATE:

I noticed that I had left a todo to add support for pagination parameters. So I added limit and offset support to find telegrafs.
While doing so, I found it was favorable to update the `kv.VisitFunc` abstraction to return both an error _and_ now a keep walking `bool`. So now you return `false` when you want to stop cursing. This touched a few things, but I believe I got them all. Some places were returning a sentinel error to stay stop walking. I fixed those place to just produce a `false` (stop walking) response instead.

I deleted both `kv` and `http` telegraf config service implementations. Moving us solely to `telegraf` package imple mentation.

- [x] Rebased/mergeable
- [x] Tests pass


GeorgeMac:
I closed this by accident.

brettbuddin (approved):

brettbuddin (approved):

PR #19828, opened 2020-10-26
fix: add 1.x compatible endpoints to swagger

### LINKED ISSUE #19753: Add 1.x compatibility api endpoints to swagger
we should add the additional v1 compatibility APIs to the swagger file.

`/write` in addition to `/api/v2/write`
`/query` in addition to `/api/v2/query`

glinton:
Closes #19753 

### DIFF
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -1750,7 +1750,6 @@ paths:
   /api/v2/write:
     servers:
       - url: /
-        description: V1 compatible api endpoints.
     post:
       operationId: PostWrite
       tags:
@@ -3371,7 +3370,6 @@ paths:
   /api/v2/query:
     servers:
       - url: /
-        description: V1 compatible api endpoints.
     post:
       operationId: PostQuery
       tags:
--- a/http/swagger.yml
+++ b/http/swagger.yml
@@ -1747,9 +1747,7 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
-  /api/v2/write:
-    servers:
-      - url: /
+  /write:
     post:
       operationId: PostWrite
       tags:
@@ -3367,9 +3365,7 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
-  /api/v2/query:
-    servers:
-      - url: /
+  /query:
     post:
       operationId: PostQuery
       tags:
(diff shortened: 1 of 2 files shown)


121watts (approved):

http/swagger.yml
```diff
@@ -3365,7 +3368,10 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
-  /query:
+  /api/v2/query:
+    servers:
+      - url: /
+        description: V1 compatible api endpoints.
```
stuartcarnie, line 3374: This description isn't correct and probably unnecessary?

http/swagger.yml
```diff
@@ -1747,7 +1747,10 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Error"
-  /write:
+  /api/v2/write:
+    servers:
+      - url: /
+        description: V1 compatible api endpoints.
```
stuartcarnie, line 1753: This description isn't correct and probably unnecessary?

stuartcarnie (changes requested):
Looks good; just a couple of nits.



To confirm my understanding, changing `/write` to `/api/v2/write` is required as it would conflict with `/write` based at `/`. You've explicitly overridden the `servers` object to change the base to `/` in both cases. I have confirmed this behavior in this spec:  http://spec.openapis.org/oas/v3.0.3#path-item-object

http/swagger.yml
glinton: regarding the servers object and rename of the original paths, yes, it's to prevent conflicts while still preserving compatibility.. though judging by the tests, i'm not sure compatibility has been preserved as the client is missing functions.

stuartcarnie:
I would recommend you confirm why the tests are failing

http/swagger.yml
stuartcarnie: Heh – right, I was just wondering the same thing

stuartcarnie (approved):
Thanks for working through that – makes sense to separate the v1 compatibility APIs for now, given the CI issues you ran into 👍

http/swagger.yml
glinton: tests failed due to a limited amount of supported content-types https://github.com/influxdata/oats/issues/16

PR #19833, opened 2020-10-27
refactor: allow newIndexSeriesCursor() to accept an influxql.Expr

wolffcm:
In order to let TagKeys and TagValues get the right answer,
sometimes they will need to examine their predicate and
see if using the index is possible, or if a block scan is needed.
For those cases we want to examine the predicate before creating
the index series cursor. This change allows us to create an
index series cursor with an already-deserialized influxql.Expr.

This just a refactoring and should have no effect on behavior.

### DIFF
--- a/v1/services/storage/series_cursor.go
+++ b/v1/services/storage/series_cursor.go
@@ -38,6 +38,18 @@ type indexSeriesCursor struct {
 }
 
 func newIndexSeriesCursor(ctx context.Context, predicate *datatypes.Predicate, shards []*tsdb.Shard) (*indexSeriesCursor, error) {
+	var expr influxql.Expr
+	if root := predicate.GetRoot(); root != nil {
+		var err error
+		if expr, err = reads.NodeToExpr(root, measurementRemap); err != nil {
+			return nil, err
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


v1/services/storage/series_cursor.go
```diff
@@ -38,6 +38,18 @@ type indexSeriesCursor struct {
...
 func newIndexSeriesCursor(ctx context.Context, predicate *datatypes.Predicate, shards []*tsdb.Shard) (*indexSeriesCursor, error) {
+	var expr influxql.Expr
+	if root := predicate.GetRoot(); root != nil {
+		var err error
+		if expr, err = reads.NodeToExpr(root, measurementRemap); err != nil {
+			return nil, err
+		}
+	}
+
+	return newIndexSeriesCursorInfluxQL(ctx, expr, shards)
```
yzhang1991, line 49: The name may not be ideal (cursor for influxql?) but the rest is good.

yzhang1991 (approved):

v1/services/storage/series_cursor.go
wolffcm: How about `newIndexSeriesCursorInfluxQLPred`?

yzhang1991: sounds good

PR #19840, opened 2020-10-28
feat(authorization): Add bcrypt password support to v1 authorizations

### LINKED ISSUE #19837: v1/authorization service should use bcrypt for verification
## What

To further simplify the upgrade process, the v1.x authorization service should store passwords using bcrypt.

## How

- [x] A v1 `Authorization` will be created using the `username` as `Token`
- [x] The `password` should be stored and verified separately in an independent bucket
- [x] `v1/authorization.Service` should implement `influxdb.PasswordsService`
- [x] Passwords are assigned to a token for verification
- [x] Implement a caching `influxdb.PasswordService`

stuartcarnie:
Closes #19837 

This commit extends the `v1/authorization` package to support passwords associated with a token using bcrypt hashing. This feature is added to support upgrading InfluxDB 1.x users and using their existing passwords. A subsequent PR will utilize these APIs to authorize requests to the `/query` and `/write` HTTP endpoints by verifying (comparing) a provided password for a given authorization. These passwords will either be supplied via an `Authorization` header using the `BASIC` authentication scheme or via the `u` (user) and `p` (password) query parameters.

The `influxd upgrade` command will be extended to copy the users from `meta.db` and store the bcrypted version of the user's password so existing clients will not require changes.

The summary of changes include:

* `authorization.Service` implements `influxdb.PasswordsService`
* As an implementation of `influxdb.PasswordsService`:
    * passwords may now be assigned to authorizations
    * verifying (comparing) passwords for a given authorization, to be used by v1 handlers in the future (`/query`, `/write`)
* A separate `influxdb.PasswordsService` that implements a cache when comparing passwords. These are salted and stored in-memory using a SHA256 hash. 
  **NOTE:** This implementation is essentially copied from InfluxDB 1.x.
* Extended HTTP service to set a password using `/private/legacy/authorizations/{id}/password`

As this is a private API, the CHANGELOG is not updated.

- [x] Rebased/mergeable
- [x] Tests pass


v1/authorization/caching_password_service.go
```diff
@@ -0,0 +1,110 @@
...
+// An implementation of influxdb.PasswordsService that will perform
+// ComparePassword requests at a reduced cost under certain
+// conditions. See ComparePassword for further information.
+//
+// The cache is only valid for the duration of the process.
+type CachingPasswordsService struct {
+	inner influxdb.PasswordsService
+
+	mu        sync.RWMutex // protects concurrent access to authCache
+	authCache map[influxdb.ID]authUser
```
yquansah, line 23: What are the bounds on this in memory cache.. or do we not need to worry about it growing unbounded?

yquansah (approved):

v1/authorization/caching_password_service.go
stuartcarnie: Good question – this is only for unique user name tokens, so we'd expect this to be probably dozens at most

PR #19846, opened 2020-10-28
chore: prune some unreferenced packages

mark-rushakoff:
This PR removes a handful of packages that were not referenced anywhere in influxdb.


stuartcarnie (approved):
🧹 💨

PR #19848, opened 2020-10-29
fix: tag keys push down

yzhang1991:
Closes #

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/v1/services/storage/store.go
+++ b/v1/services/storage/store.go
@@ -199,32 +199,65 @@ func (s *Store) ReadGroup(ctx context.Context, req *datatypes.ReadGroupRequest) 
 }
 
 type  metaqueryAttributes struct {
-	orgID influxdb.ID
-	db, rp string
+	orgID      influxdb.ID
+	db, rp     string
 	start, end int64
-	pred influxql.Expr
+	pred       influxql.Expr
+}
+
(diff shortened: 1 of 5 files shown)


v1/services/storage/store.go
```diff
@@ -239,6 +272,17 @@ func (s *Store) TagKeys(ctx context.Context, req *datatypes.TagKeysRequest) (cur
...
+				rp:    rp,
+				start: start,
+				end:   end,
+				pred:  expr,
+			}
+			return s.tagKeysWithFieldPredicate(ctx, mqAttrs, shardIDs)
+		}
 		// this will remove any _field references, which are not indexed
 		//   see https://github.com/influxdata/influxdb/issues/19488
 		expr = influxql.Reduce(RewriteExprRemoveFieldKeyAndValue(influxql.CloneExpr(expr)), nil)
```
wolffcm, line 288: This logic here is no longer needed, since we won't have any predicate on `$` since we would have returned an error above, nor `_field`, since we would have taken the slow path.

wolffcm: It's still worth calling `Reduce(influxql.CloneExpr(expr))` though since something like `(r) => true` might have gotten pushed down.

wolffcm (approved):
Looks good, just had a small comment about some logic that's no longer needed.

PR #19839, opened 2020-10-27
feat: add tick generation properties and legendColorizeRows

TCL735:
Follow on work to https://github.com/influxdata/giraffe/issues/79 and https://github.com/influxdata/giraffe/pull/317

Mirrors idpe pull 9027

Adds properties for storing the user's selections for tick generation, including a `generateAxisTicks` property that will turn on/off the entire feature regardless of the numeric values of the other properties.

Also adds the `legendColorizeRows` property to allow the user to toggle color on/off in the legend.

- [ ] [CHANGELOG.md](https://github.com/influxdata/influxdb/blob/master/CHANGELOG.md) updated with a link to the PR (not the Issue)
- [x] [Well-formatted commit messages](https://www.conventionalcommits.org/en/v1.0.0-beta.3/)
- [x] Rebased/mergeable
- [x] Tests pass
- [x] http/swagger.yml updated (if modified Go structs or API)
- [x] Feature flagged (if modified API)
- [x] Documentation updated or issue created (provide link to issue/pr)
- [x] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)

PR #19850, opened 2020-10-29
feat: update generate ticks into an array of properties for each axis

TCL735:
Follow-up to #19839  - we need to separate the control of the tick generation for each axis, as there will be cases where a user wants to control tick generation on one axis but not both. Also, the property should be an array of strings due to Giraffe's extreme flexibility in its configuration. (Giraffe uses 3 properties to control the tick generation, and any combination of the 3 properties including being missing or null works in the config.) To capture this flexibility for the user, we save the properties selected by the user in an array of strings.

Also, removed the tick generation properties from `CheckView` everywhere to match the swagger.yml

- [x] Rebased/mergeable
- [x] Tests pass


TCL735:
Ready for review.

glinton (approved):

PR #19855, opened 2020-10-29
feat: enable window pushdowns

jlapacik:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -47,7 +47,7 @@ require (
 	github.com/hashicorp/vault/api v1.0.2
 	github.com/imdario/mergo v0.3.9 // indirect
 	github.com/influxdata/cron v0.0.0-20191203200038-ded12750aac6
-	github.com/influxdata/flux v0.91.0
+	github.com/influxdata/flux v0.92.0
 	github.com/influxdata/httprouter v1.3.1-0.20191122104820-ee83e2772f69
 	github.com/influxdata/influxql v0.0.0-20180925231337-1cbfca8e56b6
 	github.com/influxdata/pkg-config v0.2.5
--- a/query/stdlib/influxdata/influxdb/rules.go
+++ b/query/stdlib/influxdata/influxdb/rules.go
@@ -3,6 +3,7 @@ package influxdb
 import (
 	"context"
 	"math"
+	"time"
 
 	"github.com/influxdata/flux"
 	"github.com/influxdata/flux/ast"
@@ -14,7 +15,6 @@ import (
 	"github.com/influxdata/flux/stdlib/universe"
 	"github.com/influxdata/flux/values"
 	"github.com/influxdata/influxdb/v2/kit/feature"
-	"github.com/influxdata/influxdb/v2/query"
 )
 
 func init() {
@@ -683,25 +683,37 @@ func canPushWindowedAggregate(ctx context.Context, fnNode plan.Node) bool {
 	// Check the aggregate function spec. Require the operation on _value
 	// and check the feature flag associated with the aggregate function.
 	switch fnNode.Kind() {
+	case universe.MinKind:
+		minSpec := fnNode.ProcedureSpec().(*universe.MinProcedureSpec)
+		return minSpec.Column == execute.DefaultValueColLabel
+	case universe.MaxKind:
+		maxSpec := fnNode.ProcedureSpec().(*universe.MaxProcedureSpec)
+		return maxSpec.Column == execute.DefaultValueColLabel
 	case universe.MeanKind:
-		if !feature.PushDownWindowAggregateMean().Enabled(ctx) {
-			return false
-		}
 		meanSpec := fnNode.ProcedureSpec().(*universe.MeanProcedureSpec)
-		if len(meanSpec.Columns) != 1 || meanSpec.Columns[0] != execute.DefaultValueColLabel {
-			return false
-		}
-	default:
-		return false
+		return len(meanSpec.Columns) == 1 &&
+			meanSpec.Columns[0] == execute.DefaultValueColLabel
+	case universe.CountKind:
+		countSpec := fnNode.ProcedureSpec().(*universe.CountProcedureSpec)
+		return len(countSpec.Columns) == 1 &&
+			countSpec.Columns[0] == execute.DefaultValueColLabel
+	case universe.SumKind:
+		sumSpec := fnNode.ProcedureSpec().(*universe.SumProcedureSpec)
+		return len(sumSpec.Columns) == 1 &&
+			sumSpec.Columns[0] == execute.DefaultValueColLabel
+	case universe.FirstKind:
+		firstSpec := fnNode.ProcedureSpec().(*universe.FirstProcedureSpec)
+		return firstSpec.Column == execute.DefaultValueColLabel
+	case universe.LastKind:
+		lastSpec := fnNode.ProcedureSpec().(*universe.LastProcedureSpec)
+		return lastSpec.Column == execute.DefaultValueColLabel
 	}
 	return true
 }
 
 func isPushableWindow(windowSpec *universe.WindowProcedureSpec) bool {
 	// every and period must be equal
-	// every.months must be zero
 	// every.isNegative must be false
-	// offset.months must be zero
 	// offset.isNegative must be false
 	// timeColumn: must be "_time"
 	// startColumn: must be "_start"
@@ -709,10 +721,7 @@ func isPushableWindow(windowSpec *universe.WindowProcedureSpec) bool {
 	// createEmpty: must be false
 	window := windowSpec.Window
 	return window.Every.Equal(window.Period) &&
-		window.Every.Months() == 0 &&
 		!window.Every.IsNegative() &&
-		!window.Every.IsZero() &&
-		window.Offset.Months() == 0 &&
 		!window.Offset.IsNegative() &&
 		windowSpec.TimeColumn == "_time" &&
 		windowSpec.StartColumn == "_start" &&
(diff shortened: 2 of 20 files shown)


query/stdlib/influxdata/influxdb/rules.go
```diff
@@ -31,7 +31,7 @@ func init() {
 		PushDownWindowAggregateByTimeRule{},
 		PushDownBareAggregateRule{},
 		GroupWindowAggregateTransposeRule{},
-		PushDownGroupAggregateRule{},
+		// PushDownGroupAggregateRule{},
```
jlapacik, line 34: This PR does not include grouped aggregates. This rule still needs to be enabled and the corresponding launcher tests unskipped.

tsdb/engine/tsm1/array_cursor.gen.go.tmpl
```diff
@@ -192,6 +192,8 @@ func (c *{{$type}}) reset(seek, end int64, cacheValues Values, tsmKeyCursor *Key
 		})
 		if c.cache.pos == len(c.cache.values) {
 			c.cache.pos--
+        } else if c.cache.values[c.cache.pos].UnixNano() == seek {
+            c.cache.pos--
```
jlapacik, line 196: DescendingArrayCursors need to exclude points that fall on the right side of the range boundary.

tsdb/engine/tsm1/array_cursor.gen.go.tmpl
```diff
@@ -288,9 +292,9 @@ func (c *{{$type}}) Next() {{$arrayType}} {
 		}
 	}
 
-	if pos > 0 && c.res.Timestamps[pos-1] <= c.end {
+	if pos > 0 && c.res.Timestamps[pos-1] < c.end {
 		pos -= 2
-		for pos >= 0 && c.res.Timestamps[pos] <= c.end {
+		for pos >= 0 && c.res.Timestamps[pos] < c.end {
```
jlapacik, line 297: DescendingArrayCursors need to include points that fall on the left side of the range boundary.

go.mod
```diff
@@ -47,7 +47,7 @@ require (
 	github.com/hashicorp/vault/api v1.0.2
 	github.com/imdario/mergo v0.3.9 // indirect
 	github.com/influxdata/cron v0.0.0-20191203200038-ded12750aac6
-	github.com/influxdata/flux v0.91.0
+	github.com/influxdata/flux v0.91.1-0.20201029180714-1e9bfd49f21c
```
wolffcm, line 50: I thought that we are only upgrading Flux in OSS to actual releases? @rockstar Can you confirm this?

query/stdlib/influxdata/influxdb/rules.go
wolffcm: Ok. By the way `first` and `last` were not supposed to be a part of this issue either. There were separate issues for them. I'm not complaining though, it's nice to get them out of the way.

storage/flux/reader.go
```diff
@@ -231,276 +230,6 @@ READ:
...
-// isSelector returns true if given a procedure kind that represents a selector operator.
-func isSelector(kind plan.ProcedureKind) bool {
-	return kind == FirstKind || kind == LastKind || kind == MinKind || kind == MaxKind
-}
-
-func isAggregateCount(kind plan.ProcedureKind) bool {
-	return kind == CountKind
-}
-
-func (wai *windowAggregateIterator) handleRead(f func(flux.Table) error, rs storage.ResultSet) error {
```
wolffcm, line 331: Why is all this code being deleted? I would not have expected that. Maybe it will become clear as I continue my review.

storage/flux/reader.go
```diff
@@ -593,19 +328,19 @@ READ:
 		done := make(chan struct{})
 		switch typedCur := cur.(type) {
 		case cursors.IntegerArrayCursor:
-			cols, defs := determineTableColsForGroup(gc.Keys(), flux.TInt)
+			cols, defs := determineTableColsForGroup(gc.Keys(), flux.TInt, gc.Aggregate(), key)
```
wolffcm, line 331: `gc.Aggregate()` should always be `nil` here, correct?

storage/flux/reader.go
```diff
@@ -819,6 +601,234 @@ func groupKeyForGroup(kv [][]byte, spec *query.ReadGroupSpec, bnds execute.Bound
 	return execute.NewGroupKey(cols, vs)
 }
 
+type windowAggregateIterator struct {
```
wolffcm, line 604: Okay, so you moved all of this code? That makes it difficult to see what changed. Did you have to change a lot?

storage/flux/table.gen.go.tmpl
```diff
@@ -360,29 +703,211 @@ func (t *{{.name}}GroupTable) Do(f func(flux.ColReader) error) error {
...
-	t.appendTags(cr)
-	t.appendBounds(cr)
+	ts, v := aggregate(arr.Timestamps, arr.Values)
+	timestamps, values := []int64{ts}, []{{.Type}}{v}
+	for {
+		arr = t.cur.Next()
+		if arr.Len() > 0 {
+			ts, v := aggregate(arr.Timestamps, arr.Values)
+			timestamps = append(timestamps, ts)
+			values = append(values, v)
```
wolffcm, line 754: Perhaps not relevant for this PR, but there is a bug here for the `ReadGroup(min/max)` case because the tag keys for the series (pre-grouping) are not preserved for selectors like min and max. @adrian-thurston is currently working on a fix for the 2.0 Cloud, and we'll need to ensure that fix also lands here once group agg pushdowns are enabled.

tsdb/engine/tsm1/array_cursor.gen.go.tmpl
wolffcm: This is a port of a bug fix that you already fixed elsewhere, correct? Do we have a regression test here on this branch?

wolffcm: Ah, I see lots of tests below.

tsdb/engine/tsm1/array_cursor.gen.go.tmpl
wolffcm: I have the same question here as above, this is a port of a bug fix from elsewhere, right?

wolffcm (commented):
Nothing stands out to me as incorrect here, but this is a very large set of changes, so I have a few questions before you merge this:

How much of this was copy/pasted and how much needed to be edited by hand to get it to work right?

How much of these changes are covered by tests that are on this branch?

Is there any part of this PR that you'd like me to look more closely at?

storage/flux/table.gen.go.tmpl
wolffcm: I guess I'd like to be sure that we fix the `ReadGroup(min/max)` bug *before* we turn on pushdowns that expose the bug. How can we use our process to ensure this, any thoughts?

tsdb/engine/tsm1/array_cursor.gen.go.tmpl
jlapacik: Yes we had a launcher test that failed when it was unskipped. That test now passes and I brought over the specific unit tests for it as well.

tsdb/engine/tsm1/array_cursor.gen.go.tmpl
jlapacik: Correct.

go.mod
jlapacik: There was a bug in one of the flux window functions that the aggregate array cursors call. A few of the launcher test caught it. These changes depend on that fix.

storage/flux/reader.go
jlapacik: Because we're not pushing down grouped aggregates at the moment, yes.

storage/flux/reader.go
jlapacik: My bad, I did not realize I changed the structure of this file. This file is a complete copy and paste. I did not make any manual edits.

jlapacik:
@wolffcm this was mostly just copy paste. There were a few tests that failed due to a bug in flux and one test that failed due to a regression that was reintroduced after the repo split. I was able to find the original fix for the regression and port it (although a few more changes were required to get it working).

I fixed the bug in flux around calculating window boundaries so that the failing launcher tests succeed after updating the flux dependency. I also added an flux end-to-end test to catch any regressions in both repos.

Aside from those fixes, the only other area of manual intervention was from me removing capabilities and unused feature flags and updating the planner tests that used them.

tsdb/engine/tsm1/array_cursor.gen.go.tmpl
jlapacik: Correction, this part of the file was not part of the original fix. These two lines do not appear to be in cloud, and so it's not clear to me whether this regression exists there as well. The launcher test that caught it now only lives in OSS.

wolffcm:
@jlapacik Ok, sounds good. The only remaining concern I have is regarding how we upgrade Flux in OSS now.

tsdb/engine/tsm1/array_cursor.gen.go
```diff
@@ -190,6 +190,8 @@ func (c *floatArrayDescendingCursor) reset(seek, end int64, cacheValues Values,
 		})
 		if c.cache.pos == len(c.cache.values) {
 			c.cache.pos--
+		} else if c.cache.values[c.cache.pos].UnixNano() == seek {
+			c.cache.pos--
```
scbrickley, line 194: Another area of code I'm unfamiliar with. What exactly is the purpose of the array cursor type? And why is this `else if` now necessary?

scbrickley (commented):
Left a few comments and questions, but really only out of curiosity, rather than critique. Happy to take the conversation offline if it's easier to give me an ELI5 that way.

storage/flux/table.go
```diff
@@ -234,15 +240,27 @@ func (t *floatTable) toArrowBuffer(vs []float64) *array.Float64 {
 func (t *floatGroupTable) toArrowBuffer(vs []float64) *array.Float64 {
 	return arrow.NewFloat(vs, t.alloc)
 }
+func (t *floatWindowSelectorTable) toArrowBuffer(vs []float64) *array.Float64 {
```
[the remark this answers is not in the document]
jlapacik: These are used by selector tables to convert a slice of values into an arrow array. These methods are only used by selector table implementations which is why they're just being added right now. Github hides the diffs of generated files by default which is why you might not have noticed these functions being called.

tsdb/engine/tsm1/array_cursor.gen.go
jlapacik: `last()` pushdowns use descending array cursors. If points fall on the boundary of the passed in range, we want to keep the points on the left hand boundary and exclude the points on the right hand boundary. This line concerns the right hand boundary. `seek` represents the `_stop` time. So if we see a point with timestamp equal to `_stop` we want to exclude it.

wolffcm (approved):
Looks great, nice work.

PR #19856, opened 2020-10-29
fix: make tagKeys and tagValues work for edge cases involving _field

### LINKED ISSUE #19794: Using TagValues to get values of _field with tag predicates returns wrong answer
In Flux we use a special storage method called `TagValues` to allow the user to discover the values of a tag key. Most of the time this works fine, but for a case like this test it does not:

https://github.com/influxdata/flux/blob/d115c75ebd33120b58bce1c697ea1f3df223ce4d/stdlib/influxdata/influxdb/schema/measurement_field_keys_test.flux#L71

We are seeking the fields that fulfill the predicate `r._measurement == "sys" and r.host == "host.local"`. Storage's implementation `tagValues` will filter the measurements on the predicate (in this case only `sys` would match) and then return *all* of the fields in the measurement. So when using storage code to do this we get the four fields
```
load3
load7
load8
load9
```
When only `load3` and `load8` fulfill the `r.host == "host.local"` part of the predicate.

It doesn't seem to be possible to filter out fields based on predicates on tags using the index. This is at odds with Flux which wants to treat `_field` as a tag key like any other.

I will be updating the planner rule that Flux uses to push down `tagValues` to storage, in order to avoid giving a wrong answer for this case.

Some More Details
---

Looking at the tag set as it's stored in the index I see this for the Telegraf `cpu` measurement:
https://github.com/influxdata/influxdb/blob/ca2055c16c4d69f7155413aebb6d694f65bfaca6/tsdb/index/tsi1/log_file.go#L402
```
mm.tagSet: map[
  cpu:{
    name:[99 112 117]
    deleted:false
    tagValues:map[
      cpu0:{name:[99 112 117 48] deleted:false series:map[481:{}] seriesSet:<nil>}
      cpu2:{name:[99 112 117 50] deleted:false series:map[673:{}] seriesSet:<nil>}
      cpu6:{name:[99 112 117 54] deleted:false series:map[623:{}] seriesSet:<nil>}]}
  host:{
    name:[104 111 115 116]
    deleted:false
    tagValues:map[
      euterpe.local:{name:[101 117 116 101 114 112 101 46 108 111 99 97 108] deleted:false series:map[623:{} 673:{}] seriesSet:<nil>}
      ip-192-168-1-16.ec2.internal:{name:[105 112 45 49 57 50 45 49 54 56 45 49 45 49 54 46 101 99 50 46 105 110 116 101 114 110 97 108] deleted:false series:map[481:{}] seriesSet:<nil>}]}
  ]
```
Since `_field` or `\0xff` is not listed in the tag set, the index doesn't seem to actually store the field names in the index at all.

One possible workaround would be to implement `tagValues` for cases like this in a way similar to `readFilter`, where we could make a query as if we were actually returning data to the client, but limit each returned series to one point. For each series that has a point, collect the field name and then sort and dedupe the list of fields. Then return this field list. That logic could replace what currently happens here:
https://github.com/influxdata/influxdb/blob/7f94b0d5db279521afbcf45fe3af1e8831f0a0c0/v1/services/storage/store.go#L431

### LINKED ISSUE #19806: Using TagKeys returns wrong answers with respect to "_field"
The storage API `tagKeys` is also affected by the TSI not indexing `_field` as described in #19794.

Specifically, it will get wrong answers if there is a predicate on `_field`.

Here is an example of a test that will fail if `PushDownReadTagkeysRule` is re-enabled and we call `tagKeys`:
https://github.com/influxdata/flux/blob/8b7b2cb5ea6973ad051f1d962d48a528161a9985/stdlib/influxdata/influxdb/schema/show_tag_keys_with_pred_test.flux

### User Impact
The user impact for this issue is similar to #19794. The current status on OSS 2.0 master is that we just don't push down tagValues or tagKeys operations to storage, so metaqueries will be slow for non-trivial amounts of data. In particular this has a big impact on the UI experience, especially the Query Builder.

We should update the implementation of `tagKeys` to work around this limitation of the TSI. Specifically, if there is a predicate on `_field`, we can proceed with a block scan similar to a call to `readFilter` to be able to evaluate the predicate. This solution is similar to the one described in #19794.

DOD:
- `PushDownReadTagkeysRule` is re-enabled in the planner
- Flux end-to-end tests are passing
- Work described above in `tagKeys` implementation is complete
- Unit tests for this work? Currently there does not seem to be unit tests for tagKeys(), will end-to-end tests be enough?

wolffcm:
Fixes #19806
Fixes #19794

This updates the `tagKeys` and `tagValues` methods of the `Store` interface to return the answers that Flux is used to, by properly filtering on `_field` when such a predicate is present. These methods are frequently used by the UI query builder to describe the shape of the users data.

- [x] Rebased/mergeable
- [x] Tests pass


wolffcm:
I had to re-skip the Flux test `show_tag_keys` because the current version of Flux has a bug in the test (type of a field changes for different tag values, which tsm1 apparently does not like).

The tip of Flux master has more tests (all of which pass against this branch), and also fixes this issue with the test. I made a note for myself to unskip this test after Flux has upgraded on Monday.

v1/services/storage/store.go
```diff
@@ -529,3 +557,81 @@ func (s *Store) measurementFields(ctx context.Context, req *datatypes.TagValuesR
...
+	}
+	return l != 0
+}
+
+// tagValuesSlow will determine the tag values for the given tagKey.
+// It's generally faster to use tagValues, measurementFields or
+// MeasurementNames, but those methods will only use the index and metadata
+// stored in the shard. Because fields are not themselves indexed, we have no way
+// of correlating fields to tag values, so we sometimes need to consult tsm to
+// provide an accurate answer.
```
stuartcarnie, line 590: Appreciate the detailed comment 👍

v1/services/storage/store.go
```diff
@@ -232,26 +233,66 @@ func (s *Store) ReadGroup(ctx context.Context, req *datatypes.ReadGroupRequest)
...
+			c := rs.Cursor()
+			if c == nil {
+				// no data for series key + field combination
+				return
+			}
+			defer c.Close()
+			if cursorHasData(c) {
+				for _, tag := range rs.Tags() {
+					m[string(tag.Key)] = struct{}{}
+				}
```
stuartcarnie, line 265: Go doesn't optimize these struct loops, which results in copying every `Tag` struct from the slice. I would recommend:

```go
tags := rs.Tags()
for i := range tags {
	m[string(tags[i].Key)] = struct{}{}
}
```

stuartcarnie (approved):
Looks great to me, @wolffcm – I had one suggestion to improve a struct iteration, as I've seen these loops show up in profiles in the past. Happy to skip the change in this PR and we can address it in a subsequent update if it shows up in profiles.

nathanielc (approved):
:tada: This is great!

PR #19865, opened 2020-10-30
feat(http): Implementation of v1 authorization

### LINKED ISSUE #19768: Update v1 HTTP handlers to use v1 authorization
## What

Authorizing a v1 compatible request via `/write` or `/query` should support authorizing requests using tokens as follows:

IF the request uses BASIC auth scheme or provides the credentials via the `u` and `p` query parameters:

1. Try a v1 authorization token and compare password
2. Try a v2 authorization token

IF the request uses the Token auth scheme

2. Try a v2 authorization token

## How

Update the `Influx1xAuthenticationHandler` by factoring the following code:

https://github.com/influxdata/influxdb/blob/8c0ff75c6df0148e1699c67f3b873b78dd160aef/http/legacy/influx1x_authentication_handler.go#L41-L79

The changes should ultimately result in the following steps:

1. `v1AuthService.FindAuthorizationByToken(ctx, <username>)`
2. If found, `v1AuthService.ComparePassword(ctx, auth.id, <password>)`
    * If valid, set this authorizer on the `context.Context` and proceed 
2. `FindAuthorizationByToken(ctx, creds.Token)` 
    * If found, set this authorizer on the `context.Context` and proceed
3. return `401 Unauthorized`

stuartcarnie:
Closes #19768

The PR introduces a new behavior, `AuthorizerV1` which is used to authorize requests to a 1.x API. The `AuthorizerV1` provides a single function, `Authorize`, which takes a set of credentials as input and returns an `influxdb.Authorizer` or `error` as output. This implementation and an extensive set of tests are found in the `v1/authorization` package.

This `AuthorizerV1` behavior is then used by the `http/legacy` package to authorize `/write` and `/query` requests.

- [x] Rebased/mergeable
- [x] Tests pass


v1/authorization/authorizer.go
```diff
@@ -0,0 +1,124 @@
...
+
+		// check the user is still active
+		if user, userErr := v.User.FindUserByID(ctx, auth.UserID); err != nil {
+			auth, err = nil, v.normalizeError(userErr)
+			return
+		} else if user == nil || user.Status != influxdb.Active {
+			auth, err = nil, influxdb.ErrCredentialsUnauthorized
+			return
+		}
+	}()
```
brettbuddin, line 66: 🎨 Since you're doing a lot of work here in this `defer`, would it be better to graduate it to be logic in this `Authorize` function and push down the work (below) to a subroutine? That way you don't need to juggle return values in this way.

brettbuddin (approved):
Small nit, but otherwise looks great.

v1/authorization/authorizer.go
stuartcarnie: Done if you want to take another look; it actually identified an error:

```go
if user, userErr := v.User.FindUserByID(ctx, auth.UserID); err != nil {
```

should have been:

```go
if user, userErr := v.User.FindUserByID(ctx, auth.UserID); userErr != nil {
```

brettbuddin (approved):
🥳

PR #19869, opened 2020-10-30
fix(tsm1): "snapshot in progress" error during backup

davidby-influx:
When an InfluxDB database is very busy writing new points the backup the process can fail because it can not write a new snapshot. The error is: operation timed out with error: create snapshot: snapshot in progress.This happens because InfluxDB takes almost "continuously" a snapshot from the cache caused by the high number of points ingested.

The fix for this was https://github.com/influxdata/influxdb/pull/16627 but it was for OSS only, and was not in the code path for backups in clusters.

This fix adds a `skipCacheOk` flag to `tsdb.Engine.CreateSnapshot()`.
A value of `true` allows the backup to proceed even if a cache snapshot
cannot be taken.

This flag is set to `true` in `tsm1.Engine.Backup()`, the OSS backup code path
and in `tsdb.Shard.CreateSnapshot()`, the cluster backup code path.

This flag is set to `false` in `tsm1.Engine.Export()`

- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -912,26 +912,16 @@ func (e *Engine) Free() error {
 // of the files in the archive. It will force a snapshot of the WAL first
 // then perform the backup with a read lock against the file store. This means
 // that new TSM files will not be able to be created in this shard while the
-// backup is running. For shards that are still acively getting writes, this
-// could cause the WAL to backup, increasing memory usage and evenutally rejecting writes.
+// backup is running. For shards that are still actively getting writes, this
+// could cause the WAL to backup, increasing memory usage and eventually rejecting writes.
 func (e *Engine) Backup(w io.Writer, basePath string, since time.Time) error {
 	var err error
 	var path string
-	for i := 0; i < 3; i++ {
-		path, err = e.CreateSnapshot()
-		if err != nil {
-			switch err {
-			case ErrSnapshotInProgress:
-				backoff := time.Duration(math.Pow(32, float64(i))) * time.Millisecond
-				time.Sleep(backoff)
-			default:
-				return err
-			}
-		}
+	path, err = e.CreateSnapshot(true)
+	if err != nil {
+		return err
 	}
-	if err == ErrSnapshotInProgress {
-		e.logger.Warn("Snapshotter busy: Backup proceeding without snapshot contents.")
-	}
+
 	// Remove the temporary snapshot dir
 	defer func() {
 		if err := os.RemoveAll(path); err != nil {
@@ -998,7 +988,7 @@ func (e *Engine) timeStampFilterTarFile(start, end time.Time) func(f os.FileInfo
 }
 
 func (e *Engine) Export(w io.Writer, basePath string, start time.Time, end time.Time) error {
-	path, err := e.CreateSnapshot()
+	path, err := e.CreateSnapshot(false)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


tsdb/engine/tsm1/engine.go
```diff
@@ -1947,9 +1937,13 @@ func (e *Engine) WriteSnapshot() (err error) {
 }
```
ayang64, line 1938: I'd avoid using a `bool` value here to control this behavior.  Obviously passing `bool` is fine if you want to convey that value to something like:

```
func SetUserAdmin(b bool)
```

where we're passing it because we actually need the `bool` as a value.  This is a case, I think, it only exists to change behavior.  That usually signals to me that it is time for a new function.

Or, in this case, do you think it'd be possible to just call `e.WriteSnapshot()` instead of `CreateSnapshot()`?

ayang64 (changes requested):
Great first PR.  I'd just reconsider passing a bool to `CreateSnapshot()`.

tsdb/engine/tsm1/engine.go
dgnorton: I believe this review thread should be on the `CreateSnapshot()` method that was modified below and not on the `WriteSnapshot()` method.

I think avoiding `bool` parameters is good practice in general because some functions tend to grow. However, `CreateSnapshot()` has remained simple and mostly unchanged over time. It takes only two steps: 1) snapshot in-mem write cache to a file on disk then 2) take a snapshot of all the files on disk. I think it would be overkill to refactor this into separate functions to avoid the `bool` parameter.

tsdb/engine/tsm1/engine.go
```diff
@@ -1947,9 +1937,13 @@ func (e *Engine) WriteSnapshot() (err error) {
...
 // CreateSnapshot will create a temp directory that holds
-// temporary hardlinks to the underylyng shard files.
-func (e *Engine) CreateSnapshot() (string, error) {
-	if err := e.WriteSnapshot(); err != nil {
+// temporary hardlinks to the underlying shard files.
+// skipCacheOk controls whether it is permissible to fail writing out
+// in-memory cache data when a previous snapshot is in progress
+func (e *Engine) CreateSnapshot(skipCacheOk bool) (string, error) {
+	if err := e.WriteSnapshot(); (err == ErrSnapshotInProgress) && skipCacheOk {
+		e.logger.Warn("Snapshotter busy: Backup or export proceeding without cache snapshot contents.")
```
dgnorton, line 1945: Probably should change this log message to something like `cache snapshotter busy: proceeding without cache contents` since this function can potentially be called for other reasons. I.e., this function doesn't know why it's being called.

tsdb/shard.go
```diff
@@ -1158,7 +1158,7 @@ func (s *Shard) CreateSnapshot() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return engine.CreateSnapshot()
+	return engine.CreateSnapshot(true)
```
dgnorton, line 1161: Is it okay to hard code this to `true`? Wondering if `Shard.CreateSnapshot()` also needs to take a flag.

dgnorton (changes requested):
On the right track. One nit and one question to consider.

Also need to add a unit test for this new behavior.

tsdb/shard.go
davidby-influx: Shard.CreateSnapshot appears to only be called by tests in store_test.go and shard_test.go, so hardcoding to true maintains current behavior for those tests

tsdb/engine/tsm1/engine.go
davidby-influx: Changed, will push.

tsdb/engine/tsm1/engine.go
davidby-influx: Per @dgnorton , will probably leave this as a boolean, but agree on best practices with you.

tsdb/shard.go
dgnorton: I think it's also called from outside of `influxdb`.

tsdb/engine/tsm1/engine.go
davidby-influx: Unit test added in https://github.com/influxdata/influxdb/pull/19869/commits/0dcff81f562809ab471958bb68b4557e4f6aced2

davidby-influx: Comment changed in https://github.com/influxdata/influxdb/pull/19869/commits/6ec446f4222e60f07c698a973cd13a45553c0834

tsdb/shard.go
davidby-influx: Fixed in https://github.com/influxdata/influxdb/pull/19869/commits/6ec446f4222e60f07c698a973cd13a45553c0834

ayang64 (approved):
LGTM!

dgnorton (approved):
lgtm

PR #19870, opened 2020-10-30
fix(models): correctly parse float as 64-bits

mark-rushakoff:
Go tip now enforces that strconv.ParseFloat receives a bitSize parameter
of 32 or 64. Previously, any value that was not 32, was treated as 64.

The previous value of 10 was likely a copy-paste error from integer
parsing.


stuartcarnie (approved):
Thanks for fixing this here too 👍

PR #19906, opened 2020-11-05
feat: vendoring the ui

drdelambre:
to reduce the number of places within our organization that determine the definition of what constitutes our UI offering, we are depreciating the influxdb ui source files in favor of vendoring in compiled assets that are generated from influxdata/ui.

there's some hold up on automating this process, so the first one comes in hand rolled to make the GA

no more waiting for `yarn install` to run! amirite?!


mark-rushakoff:
👎  on vendoring in .wasm files and minified javascript files. That will greatly bloat the size of stored git objects in this repo.

Can you instead publish the assets as a github release in the ui repo, and add a make target to download and untar those assets into a .gitignore-d directory?

wbaker85:
Closing this, as a similar scope was completed by #21090.

Issue #19737, opened 2020-10-13
Port beta.8 Prometheus metrics for storage engine


STRRL:
Hi, I have tried influxdb 2.0 rc on  staging env.  
I found that there are some differences between rc0 and beta8 on /metrics, some metrics are missed, for example:

- query_control_queueing_active
- storage_tsm_files_disk_bytes
- storage_compactions_active

I will provide a response on /metrics with each verisons.

2.0 beta8: https://pastebin.ubuntu.com/p/c3Rpsfq87q/
2.0 rc0: https://pastebin.ubuntu.com/p/V5Q8H9ZxpJ/

These metrics are very useful and powerful for tunning and profiling influxdb. 

__Environment info:__

```shell
#influxd version
InfluxDB 2.0.0-rc.0 (git: df47ec7bb2) build_date: 2020-09-29T22:08:57Z
```

danxmoran:
Hi @STRRL, thanks for the report. I did some digging and can confirm all storage-related metrics got dropped between beta8 and rc0, when the 1.x storage engine was ported over to the 2.x line. We're working on restoring them.

<details>
<summary>Click for full list of missing metrics from the provided pastebins</summary>

```
storage_cache_age_seconds
storage_cache_disk_bytes
storage_cache_inuse_bytes
storage_cache_snapshot_bytes
storage_cache_snapshots_active
storage_cache_writes_total
storage_cache_written_bytes
storage_compactions_active
storage_compactions_duration_seconds_bucket
storage_compactions_duration_seconds_count
storage_compactions_duration_seconds_sum
storage_compactions_queued
storage_compactions_total
storage_reads_cursors
storage_reads_seeks
storage_retention_check_duration_seconds_bucket
storage_retention_check_duration_seconds_count
storage_retention_check_duration_seconds_sum
storage_retention_checks_total
storage_series_file_compactions_total
storage_series_file_disk_bytes
storage_series_file_index_compactions_active
storage_series_file_index_compactions_duration_seconds_bucket
storage_series_file_index_compactions_duration_seconds_count
storage_series_file_index_compactions_duration_seconds_sum
storage_series_file_index_get_duration_last_ns
storage_series_file_index_get_duration_ns_bucket
storage_series_file_index_get_duration_ns_count
storage_series_file_index_get_duration_ns_sum
storage_series_file_index_get_total
storage_series_file_segments_total
storage_series_file_series_created
storage_series_file_series_total
storage_tsi_cache_deletes_total
[...]

PR #19912, opened 2020-11-05
feat: instrument TSI index with prometheus metrics

danxmoran:
Part of #19737 

Still very much draft, but opening for early review of the approach.

- [ ] Rebased/mergeable
- [ ] Tests pass


storage/metrics/metrics.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// namespace is the leading part of all published metrics for the Storage service.
+const namespace = "storage"
+
+// StorageMetrics wrap all collectors of storage-related metrics.
+type StorageMetrics struct {
+	DefaultLabels *prometheus.Labels
+	Cache         *CacheMetrics
+	Partition     *PartitionMetrics
```
danxmoran, line 14: Idea is that this will grow as we instrument more of the storage subsystems

tsdb/index/inmem/inmem.go
```diff
@@ -1116,6 +1119,8 @@ type ShardIndex struct {
 	opt tsdb.EngineOptions
 }
 
+func (idx *ShardIndex) SetDefaultMetricLabels(labels prometheus.Labels) {}
```
danxmoran, line 1122: Leftover, need to delete

tsdb/index/tsi1/cache.go
```diff
@@ -48,11 +52,13 @@ func (c *TagValueSeriesIDCache) get(name, key, value []byte) *tsdb.SeriesIDSet {
 	if mmap, ok := c.cache[string(name)]; ok {
 		if tkmap, ok := mmap[string(key)]; ok {
 			if ele, ok := tkmap[string(value)]; ok {
+				c.tracker.IncGetHit()
```
danxmoran, line 55: These are the calls that actually modify the tracked metrics. I was (mostly) able to replicate the logic 1:1 from the beta.8 code (see [here](https://github.com/influxdata/influxdb/blob/v2.0.0-beta.8/tsdb/tsi1/cache.go))

tsdb/index/tsi1/cache.go
```diff
@@ -198,3 +218,82 @@ type seriesIDCacheElement struct {
 	value       string
 	SeriesIDSet *tsdb.SeriesIDSet
 }
+
+type cacheTracker struct {
```
danxmoran, line 222: From here down is pretty much copy-paste. My main addition was adding `if t == nil` checks to many of the methods, to support running without metrics-tracking (an option in the beta, not sure if we want to keep it).

tsdb/index/tsi1/partition.go
```diff
@@ -237,6 +245,10 @@ func (p *Partition) Open() error {
 	if err := p.buildSeriesSet(); err != nil {
 		return err
 	}
+	p.tracker.SetSeries(p.seriesIDSet.Cardinality())
```
danxmoran, line 248: Similar to cache.go, I was able to make a 1:1 mapping from the beta code for most of this. See [here](https://github.com/influxdata/influxdb/blob/v2.0.0-beta.8/tsdb/tsi1/partition.go).

storage/engine.go
```diff
@@ -155,7 +157,7 @@ func (e *Engine) WithLogger(log *zap.Logger) {
 // PrometheusCollectors returns all the prometheus collectors associated with
 // the engine and its components.
 func (e *Engine) PrometheusCollectors() []prometheus.Collector {
```
danxmoran, line 159: This method is already being called [here](https://github.com/influxdata/influxdb/blob/master/cmd/influxd/launcher/launcher.go#L854) in the `influxd` launcher, so all the collectors we return should get registered in the prometheus HTTP registry

storage/metrics/metrics.go
```diff
@@ -0,0 +1,35 @@
+package metrics
```
dgnorton, line 1: Would it be cleaner to have the types in this package live in the `storage` package?
```
storage.Metrics
storage.TSICacheMetrics
storage.WALMetrics
etc.
```

storage/metrics/tsicache.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sort"
+)
+
+const cacheSubsystem = "tsi_cache"     // sub-system associated with TSI index cache.
+
+// CacheMetrics wrap all collectors of metrics from the TSI index cache.
+type CacheMetrics struct {
```
dgnorton, line 11: If this is for TSI, we should put `TSI` in the name. There is also an inmem write cache that will need to track metrics will probably be called `CacheMetrics`.

storage/metrics/metrics.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// namespace is the leading part of all published metrics for the Storage service.
+const namespace = "storage"
+
+// StorageMetrics wrap all collectors of storage-related metrics.
+type StorageMetrics struct {
+	DefaultLabels *prometheus.Labels
```
danxmoran, line 12: Not a huge fan of `DefaultLabels` because it implies these could/should get overwritten. I believe the intent from the beta was more to support configurable _global_ labels (i.e. to separately label different nodes in a cluster)

tsdb/index/tsi1/cache.go
danxmoran: Need to add a comment explaining why this struct exists

tsdb/index/tsi1/cache_test.go
```diff
@@ -15,7 +15,7 @@ func TestTagValueSeriesIDCache(t *testing.T) {
 	m0k1v2 := tsdb.NewSeriesIDSet()
 	m1k3v0 := tsdb.NewSeriesIDSet(900, 0, 929)
 
-	cache := TestCache{NewTagValueSeriesIDCache(10)}
+	cache := TestCache{NewTagValueSeriesIDCache(10, nil)}
```
danxmoran, line 18: I'd be interested in trying to replace these `nil` values with actual trackers, and extending the tests to assert on the collected values.

tsdb/index/tsi1/partition.go
```diff
@@ -1200,6 +1276,181 @@ func (p *Partition) compactLogFile(logFile *LogFile) {
 	}
 }
 
+type partitionTracker struct {
```
danxmoran, line 1279: Needs a comment explaining why this exists

tsdb/shard.go
```diff
@@ -200,6 +202,11 @@ func (s *Shard) WithLogger(log *zap.Logger) {
 	s.logger = s.baseLogger.With(zap.String("service", "shard"))
 }
 
+// SetDefaultMetricLabels sets prometheus labels for the shard. It must be called before Open.
```
danxmoran, line 205: Comment needs updated

tsdb/index/inmem/inmem.go
```diff
@@ -14,6 +14,8 @@ package inmem
 import (
 	"errors"
 	"fmt"
+	"github.com/influxdata/influxdb/v2/storage/metrics"
```
dgnorton, line 17: We don't need to instrument `inmem` index since support for that is being dropped.

storage/metrics/tsicache.go
```diff
@@ -0,0 +1,73 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sort"
+)
+
+const cacheSubsystem = "tsi_cache"     // sub-system associated with TSI index cache.
```
danxmoran, line 8: Latest mystery: when I test locally, I don't see any metrics from this subsystem. I'm not sure if that's because of bugs in the code or because I'm not exercising the code-paths in the cache.

tsdb/index/tsi1/cache.go
```diff
@@ -2,6 +2,8 @@ package tsi1
 
 import (
 	"container/list"
+	metrics2 "github.com/influxdata/influxdb/v2/storage/metrics"
```
danxmoran, line 5: GoLand trying to be helpful with this `metrics2`, needs changed

storage/metrics/metrics.go
danxmoran: I started with them there, but got cyclic import errors.

I'll see if they work out in the `tsdb` package (the 1.x storage tree). That feels like it could be a cleaner abstraction anyways, since new storage engines would probably(?) have different metrics.

tsdb/index/inmem/inmem.go
danxmoran: Yeah, the stub method here is just to satisfy the `Index` interface

tsdb/store.go
```diff
@@ -651,6 +661,7 @@ func (s *Store) CreateShard(database, retentionPolicy string, shardID uint64, en
 	path := filepath.Join(s.path, database, retentionPolicy, strconv.FormatUint(shardID, 10))
 	shard := NewShard(shardID, path, walPath, sfile, opt)
 	shard.WithLogger(s.baseLogger)
+	shard.WithMetrics(s.metrics)
```
danxmoran, line 664: I'm uneasy about this touch-point. In the beta.8 code, the storage architecture had been refactored so there was a single top-level TSI `Index`. In this implementation there's an `Index` per-shard, with potentially many shards. We don't do anything here to distinguish different shards in the metrics labels, so they'll all end up fusing together. Should we inject a label here [like we do for index partitions](https://github.com/influxdata/influxdb/pull/19912/files#diff-18f80f7e0688cd2a4a53981f14766c61b98bfb8a76bc0170cba27f96f6d49a8bR285-R288)?

chronograf/dist/dist.go
```diff
@@ -33,9 +33,9 @@ func (b *BindataAssets) Handler() http.Handler {
 	return b
 }
 
-// addCacheHeaders requests an hour of Cache-Control and sets an ETag based on file size and modtime
+// addCacheHeaders requests an hour of TSICache-Control and sets an ETag based on file size and modtime
 func (b *BindataAssets) addCacheHeaders(filename string, w http.ResponseWriter) error {
-	w.Header().Add("Cache-Control", "public, max-age=3600")
+	w.Header().Add("TSICache-Control", "public, max-age=3600")
```
danxmoran, line 38: Oops, find and replace gone wrong

tsdb/index/tsi1/cache.go
```diff
@@ -136,11 +142,12 @@ func (c *TagValueSeriesIDCache) Put(name, key, value []byte, ss *tsdb.SeriesIDSe
 
 	// No map for the measurement - first tag key for the measurment.
 	c.cache[string(name)] = map[string]map[string]*list.Element{
-		string(key): map[string]*list.Element{string(value): listElement},
+		string(key): {string(value): listElement},
```
danxmoran, line 145: `gofmt` at work here

tsdb/metrics.go
```diff
@@ -0,0 +1,227 @@
...
+const namespace = "storage"
+
+// Metrics wrap all collectors of storage-related metrics.
+type Metrics struct {
+	GlobalLabels *prometheus.Labels
+	TSICache     *TSICacheMetrics
+	TSIPartition *TSIPartitionMetrics
+}
+
+func NewMetrics(globalLabels *prometheus.Labels) *Metrics {
```
danxmoran, line 18: Looking through the prometheus docs, I'm thinking this `globalLabels` approach might be an anti-pattern... continuing to investigate, but wanted to flag it.

If we ditch `globalLabels` here then we _could_ have all these structs as module-level singletons instantiated by `init()`. That could simplify all the code used to thread references through layers of the storage engine.

tsdb/index/tsi1/cache.go
```diff
@@ -198,3 +217,87 @@ type seriesIDCacheElement struct {
...
+	return l
+}
+
+func (t *cacheTracker) SetSize(sz uint64) {
+	if t == nil {
+		return
+	}
+
+	labels := t.Labels()
+	t.metrics.Size.With(labels).Set(float64(sz))
```
danxmoran, line 254: For other prometheus noobs like me, I just found a great example/walkthrough of different ways a program could set up / register metrics for use across multiple "worker" components: https://github.com/prometheus/client_golang/blob/master/prometheus/examples_test.go#L160

One idea that jumps out at me is: instead of calling `With` every time the code wants to interact with a metric collector, we could just call it once and have the tracker structs keep a reference to the result.

tsdb/index/tsi1/cache.go
```diff
@@ -158,16 +165,21 @@ func (c *TagValueSeriesIDCache) delete(name, key, value []byte, x uint64) {
...
 	}
+	c.tracker.IncDeletesMiss()
 }
 
 // checkEviction checks if the cache is too big, and evicts the least recently used
 // item if it is.
 func (c *TagValueSeriesIDCache) checkEviction() {
-	if c.evictor.Len() <= c.capacity {
+	l := c.evictor.Len()
+	c.tracker.SetSize(uint64(l))
```
danxmoran, line 181: This feels like buggy logic to me...

`checkEviction()` is only called at the end of `Put`, so we're not decrementing the size on `Delete`. We're also not readjusting the size after we evict items.

tsdb/index/tsi1/partition.go
```diff
@@ -480,6 +491,11 @@ func (p *Partition) prependActiveLogFile() error {
 		return err
 	}
 	p.manifestSize = manifestSize
+
+	// Set the file metrics again.
```
danxmoran, line 495: Maybe a bug... this block is almost an exact match to the one above, but it's missing:
```go
p.tracker.SetSeries(p.seriesIDSet.Cardinality())
```

PR #19920, opened 2020-11-06
fix: default http.client may cause many time_wait sockets

Chaiyq:
limit MaxIdleConns

### DIFF
--- a/client/v2/client.go
+++ b/client/v2/client.go
@@ -119,6 +119,8 @@ func NewHTTPClient(conf HTTPConfig) (Client, error) {
 		Proxy:       conf.Proxy,
 		DialContext: conf.DialContext,
 	}
+	tr.MaxIdleConns = 100
+	tr.MaxIdleConnsPerHost = 100
 	if conf.TLSConfig != nil {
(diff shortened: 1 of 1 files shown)


lesam:
Are you seeing user-facing effects of this being unlimited? Otherwise I am hesitant to change something so fundamental.

Chaiyq:
@lesam ,Thank you for your reply, we use influxdb at the iot edge. In some online environments, we found that many sockets are in timewait state, causing new requests to fail to bind the sockets. Then modify the default connection pool，it is ok.

lesam:
I think this would have to be a configuration parameter to safely merge it.

Also, I notice this is against 1.8 - we are focussing most of our development effort on OSS 2.x, do you plan to migrate? If so, do you have the same problem there? If not, what blockers to migration do you have?

Chaiyq:
@lesam ，Thank you for your reply, and agree with you very much. We will use 2.x for test, if the problem persists, it would be suitable to merge.

Chaiyq:
@lesam  Hi, Excuse me,what are the new or optimized features of 2.x vs 1.8

PR #19885, opened 2020-11-02
fix(onboard): remove extra multiplication of RP to avoid overflow

danxmoran:
Closes #19884

(As debugged by @stuartcarnie) Retention policies in onboarding requests are already proper `time.Duration`s, so multiplying by an extra `time.Hour` factor causes the values to be incorrect and potentially overflow.

- [x] [CHANGELOG.md](https://github.com/influxdata/influxdb/blob/master/CHANGELOG.md) updated with a link to the PR (not the Issue)
- [x] [Well-formatted commit messages](https://www.conventionalcommits.org/en/v1.0.0-beta.3/)
- [x] Rebased/mergeable
- [x] Tests pass
- [ ] http/swagger.yml updated (if modified Go structs or API)
- [ ] Feature flagged (if modified API)
- [ ] Documentation updated or issue created (provide link to issue/pr)
- [ ] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)

PR #19928, opened 2020-11-06
fix: don't divide RPs provided over the CLI by a factor of an hour

### LINKED ISSUE #19926: Specifying RP as a CLI option in `influx setup` or `influxd upgrade` causes spurious validation error
Ripple effects from the fix in #19884

__Steps to reproduce:__

For `influx setup`:
1. Run a fresh `influxd`
2. Run:
```
influx setup -f -u test -p testtest -o test -b test -r 72h
```
3. See the error:
```
Error: failed to setup instance: retention policy duration must be at least 1h0m0s
```

I see identical code in the setup logic for `influxd upgrade`.

__Expected behavior:__
No validation error.

__Actual behavior:__
These lines are all dividing the RP passed over the CLI by a factor of an hour:
1. [This line](https://github.com/influxdata/influxdb/blob/master/cmd/influx/setup.go#L286)
1. [This line](https://github.com/influxdata/influxdb/blob/master/cmd/influx/setup.go#L243)
1. [This line](https://github.com/influxdata/influxdb/blob/master/cmd/influxd/upgrade/setup.go#L100)

The onboarding service within `influxd` used to reverse this change by re-multiplying RPs by an hour, but this caused overflows in some cases so we removed it. The tests caught that this change broke non-interactive upgrade, but missed these other spots.

danxmoran:
Closes #19926 

Prior to #19885, `influxd` would multiply RPs from onboarding requests by a factor of an hour. This caused overflows in many cases, so I removed the multiplication. The tests didn't catch all the places where the CLI was anticipating that multiplication...

The 3 lines changed here mirror the fix I made to [non-interactive upgrade](https://github.com/influxdata/influxdb/pull/19885/files#diff-ec11f92a3df5beb5cc1657892bfc169a49bed091ce94ebd1fae3d989a8c6eceeL57-R57) as part of the original PR.

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Bug Fixes
 
 1. [19924](https://github.com/influxdata/influxdb/pull/19924): Remove unused 'security-script' option from upgrade command
+1. [19928](https://github.com/influxdata/influxdb/pull/19928): Fix parsing of retention policy CLI args in `influx setup` and `influxd upgrade`
 
 ## v2.0.0-rc.4 [2020-11-05]
 
--- a/cmd/influx/setup.go
+++ b/cmd/influx/setup.go
@@ -240,7 +240,7 @@ func nonInteractive() (*influxdb.OnboardingRequest, error) {
 		return nil, err
 	}
 	if dur > 0 {
-		req.RetentionPeriod = dur / time.Hour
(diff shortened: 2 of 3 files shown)


psteinbachs (approved):
Looks good.

PR #19935, opened 2020-11-09
fix: Improve the UI for the influx v1 auth commands

### LINKED ISSUE #19933: Improve UI for v1 auth create CLI command
## What

The `influx v1 auth create` command requires the following improvements:

* rename `--auth-token` argument to `--username` to be clear its purpose from a v1 perspective
* prompt for token password by default and ensure the password is set
* add `--no-password` to skip password prompt

## Why

The existing implementation of the `influx v1 auth create` command allows a user to create an auth token, but does not set a password until the user executes the `influx v1 auth set-password` command. Without a password, the token is unusable, returning an unauthorised error. It is unclear from the API that this is because the password is not set.

stuartcarnie:
Closes #19933

This PR improves the `v1 auth` commands of the `influx` CLI tool, and includes some minor improvements to the `kit/cli` package. The enhancements and fixes to the user experience are as follows:

* Rename `--auth-token` to `--username` for the `v1 auth create` command to clarify its purpose in the context of v1 requests

* `v1 auth create` now requires a password when creating a new authorisation. It can be optionally disabled with the `--no-password` flag, however, the user much set a password using the `v1 auth set-password` command prior to using the token.

* Rename `v1 auth active` and `v1 auth inactive` commands to the actions `set-active` and `set-inactive` respectively to clarify their behaviour.
  
  For example, to set a token to active:
  
  ```
  influx v1 auth set-active (--id|--username)
  ```
  
  To set a token to inactive, preventing its use in subsequent requests:
  
  ```
  influx v1 auth set-inactive (--id|--username)
  ```
  
* Universally allow the `--id` or `--username` flags anywhere a single authentication token is
  required

* Use separate global variables to store the flag state of each v1 auth command

* factor out the behavior to find a single authentication token into
  a shared function, `v1FindOneAuthorization`

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influx/v1_authorization.go
+++ b/cmd/influx/v1_authorization.go
@@ -520,7 +520,7 @@ func v1AuthorizationSetInactiveF(cmd *cobra.Command, _ []string) error {
 }
 
 var (
-	errMultipleMarchingAuthorizations = errors.New("multiple authorizations found")
+	errMultipleMatchingAuthorizations = errors.New("multiple authorizations found")
 )
 
 func v1FindOneAuthorization(s *authorization.Client, filter influxdb.AuthorizationFilter) (*influxdb.Authorization, error) {
@@ -534,7 +534,7 @@ func v1FindOneAuthorization(s *authorization.Client, filter influxdb.Authorizati
 	}
 
 	if len(authorizations) > 1 {
-		return nil, errMultipleMarchingAuthorizations
+		return nil, errMultipleMatchingAuthorizations
 	}
 
 	return authorizations[0], nil
@@ -567,7 +567,7 @@ func (f *v1AuthLookupFlags) register(cmd *cobra.Command, persistent bool) {
 func (f *v1AuthLookupFlags) validate() error {
 	switch {
 	case f.id.Valid() && f.username != "":
-		return errors.New("specify id or username")
+		return errors.New("specify id or username, not both")
 	case f.required && (!f.id.Valid() && f.username == ""):
 		return errors.New("id or username required")
 	default:
--- a/cmd/influx/v1_authorization.go
+++ b/cmd/influx/v1_authorization.go
@@ -31,48 +31,44 @@ func cmdV1Auth(f *globalFlags, opt genericCLIOpts) *cobra.Command {
 	cmd.Run = seeHelp
 
 	cmd.AddCommand(
-		v1AuthActiveCmd(f),
-		v1AuthCreateCmd(f),
+		v1AuthCreateCmd(f, opt),
 		v1AuthDeleteCmd(f),
 		v1AuthFindCmd(f),
-		v1AuthInactiveCmd(f),
+		v1AuthSetActiveCmd(f),
+		v1AuthSetInactiveCmd(f),
 		v1AuthSetPasswordCmd(f, opt),
 	)
 
 	return cmd
 }
 
-var v1AuthCRUDFlags struct {
-	id          string
-	json        bool
-	hideHeaders bool
-}
-
 var v1AuthCreateFlags struct {
-	token       string
-	user        string
+	username    string
 	description string
+	noPassword  bool
+	hideHeaders bool
+	json        bool
 	org         organization
 
 	writeBucketPermissions []string
 	readBucketPermissions  []string
 }
 
-func v1AuthCreateCmd(f *globalFlags) *cobra.Command {
+func v1AuthCreateCmd(f *globalFlags, opt genericCLIOpts) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "create",
 		Short: "Create authorization",
-		RunE:  checkSetupRunEMiddleware(&flags)(v1AuthorizationCreateF),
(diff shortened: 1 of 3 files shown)


cmd/influx/v1_authorization.go
```diff
@@ -31,155 +31,175 @@ func cmdV1Auth(f *globalFlags, opt genericCLIOpts) *cobra.Command {
...
+		}
 
-	var permissions []influxdb.Permission
-	for _, bp := range bucketPerms {
-		for _, p := range bp.perms {
-			var id influxdb.ID
-			if err := id.DecodeFromString(p); err != nil {
+		// verify an existing token with the same username doesn't already exist
+		filter := influxdb.AuthorizationFilter{
+			Token: &v1AuthCreateFlags.username,
```
danxmoran, line 107: I find it confusing that the `--username` flag gets mapped to the token name, instead of `user.Name`. Ex: in the demo video the command `v1 auth create --username bar` creates a token with: `Name / Token: bar` and `User Name: my-user`. It's not clear where that `my-user` came from, or how I could set it if I wanted to.

cmd/influx/v1_authorization.go
```diff
@@ -479,9 +519,78 @@ func v1AuthorizationInactiveF(cmd *cobra.Command, args []string) error {
 	})
 }
 
+var (
+	errMultipleMarchingAuthorizations = errors.New("multiple authorizations found")
```
danxmoran, line 523:
```suggestion
	errMultipleMatchingAuthorizations = errors.New("multiple authorizations found")
```

cmd/influx/v1_authorization.go
```diff
@@ -479,9 +519,78 @@ func v1AuthorizationInactiveF(cmd *cobra.Command, args []string) error {
...
+			Persistent: persistent,
+		},
+	}
+	opts.mustRegister(cmd)
+}
+
+func (f *v1AuthLookupFlags) validate() error {
+	switch {
+	case f.id.Valid() && f.username != "":
+		return errors.New("specify id or username")
```
danxmoran, line 570: Could we reword the error to make it more clear that it's an XOR? At first glance I wasn't sure why this was an error.

cmd/influx/v1_authorization.go
stuartcarnie: @danxmoran I agree. However, this is an artefact of existing product deficiencies, product management requirements (upgrade 1.x without touching existing clients) and maintaining existing API compatibility with Cloud by not changing existing Users or Tokens.

1. Product management was forced to use tokens as the User object does not provide the appropriate features to restrict access to a subset of buckets, whereas Tokens do.

2. Tokens are always owned / associated with a real v2 `User` object. This is the `User Name` column and the `my-user` user you identified. By default, when you access an InfluxDB instance via the `influx` CLI tool, you are using a token owned by a v2 User (set via config or an environment variable), in this case `my-user`. So, when the operator create new tokens, they will be owned by their V2 User.

cmd/influx/v1_authorization.go
stuartcarnie: Thanks! I am not used to typing on this laptop keyboard – one of the joys of being in quarantine in Australia right now 🤣

cmd/influx/v1_authorization.go
stuartcarnie: How about 

> specify either id or username, not both

cmd/influx/v1_authorization.go
danxmoran: Hmmm in that case, could we use something like `--token-name` as the option instead of `--username`? Or does `username` have a specific meaning when it comes to V1 that we shouldn't change?

cmd/influx/v1_authorization.go
danxmoran: LGTM

cmd/influx/v1_authorization.go
stuartcarnie: No, `--token-name` is unclear. `--username` is familiar to a v1 user, as this will match their users upgraded from a v1 database.

danxmoran: Fair enough 😄

stuartcarnie: To be clear, users upgraded from v1 are _not_ created as real V2 users, due to the deficiencies I described earlier. They will be created as these special tokens.

stuartcarnie: Our hands a tied somewhat, as we can't leverage the existing V2 Tokens. We chose to create these private tokens, with restricted use only to the 1.x compatible APIs (`/write` and `/query`).

1. We can't change them and break compatibility with Cloud APIs (for example to support bcrypt passwords, for a seamless upgrade from 1.x to 2.x)
2. It is time-consuming to propose a new token type, requiring a thorough review and verification they will be viable and useful in Cloud. Time constraints for OSS 2.0 GA prevented us from taking this approach.

danxmoran (approved):
:shipit: assuming CI passes

PR #19936, opened 2020-11-09
fix(tsi): close series id iterator after merging

foobar:
This use-after-free bug may lead to segfault. The iterators that have
reference to the underlying index files were closed too early while
the bitmaps were still used afterwards. If a compaction occurs
concurrently and removes the index files, it would result in accessing
unmap'd memory address.

### DIFF
--- a/tsdb/index.go
+++ b/tsdb/index.go
@@ -557,9 +557,11 @@ func IntersectSeriesIDIterators(itr0, itr1 SeriesIDIterator) SeriesIDIterator {
 
 	// Create series id set, if available.
 	if a := NewSeriesIDSetIterators([]SeriesIDIterator{itr0, itr1}); a != nil {
+		ss := a[0].SeriesIDSet().And(a[1].SeriesIDSet())
+		// `a` holds references to itr0/itr1 so itr0/itr1 should not be closed when `a` is still in use
 		itr0.Close()
 		itr1.Close()
-		return NewSeriesIDSetIterator(a[0].SeriesIDSet().And(a[1].SeriesIDSet()))
+		return NewSeriesIDSetIterator(ss)
 	}
 
 	return &seriesIDIntersectIterator{itrs: [2]SeriesIDIterator{itr0, itr1}}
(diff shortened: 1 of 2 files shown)


foobar:
@ayang64 PTAL

foobar:
@danxmoran can you review this PR ?

tsdb/index.go
```diff
@@ -557,9 +557,10 @@ func IntersectSeriesIDIterators(itr0, itr1 SeriesIDIterator) SeriesIDIterator {
 
 	// Create series id set, if available.
 	if a := NewSeriesIDSetIterators([]SeriesIDIterator{itr0, itr1}); a != nil {
+		ss := a[0].SeriesIDSet().And(a[1].SeriesIDSet())
```
danxmoran, line 560: Does `a[0] == itr0` and `a[1] == itr1` here? If so, could we add a comment saying so? I think it'll make it more obvious why the `Close()` calls have to happen after this

tsdb/index.go
```diff
@@ -646,10 +647,10 @@ func UnionSeriesIDIterators(itr0, itr1 SeriesIDIterator) SeriesIDIterator {
 
 	// Create series id set, if available.
 	if a := NewSeriesIDSetIterators([]SeriesIDIterator{itr0, itr1}); a != nil {
-		itr0.Close()
-		itr1.Close()
 		ss := NewSeriesIDSet()
 		ss.Merge(a[0].SeriesIDSet(), a[1].SeriesIDSet())
```
danxmoran, line 651: Same comment / request as above

tsdb/index.go
```diff
@@ -733,9 +734,10 @@ func DifferenceSeriesIDIterators(itr0, itr1 SeriesIDIterator) SeriesIDIterator {
 
 	// Create series id set, if available.
 	if a := NewSeriesIDSetIterators([]SeriesIDIterator{itr0, itr1}); a != nil {
+		ss := a[0].SeriesIDSet().AndNot(a[1].SeriesIDSet())
```
danxmoran, line 737: Same comment / request as above

tsdb/index.go
foobar: i added comments. please check if it matches your suggestion.

danxmoran (approved):
LGTM, thanks again! I'll handle wrangling the CI / changelog from here

PR #19937, opened 2020-11-09
fix(upgrade): create CLI configs

### LINKED ISSUE #19925: influxd upgrade does not create a configs file for working with the CLI
as part of the `influx setup` process, we create a `configs` file which stores a CLI configuration for accessing the instance from the command line tools. 

we need to add that to the upgrade command. I think it happens here: https://github.com/influxdata/influxdb/blob/master/cmd/influx/setup.go#L169

it is created in the `~/influxdbv2/configs` file by default

alespour:
Closes #19925

This PR contains code that creates CLI configs during upgrade from v1.

- [X] Rebased/mergeable
- [X] Tests pass
- [X] CHANGELOG.md updated with a link to the PR (not the Issue)


cmd/influxd/upgrade/upgrade.go
```diff
@@ -149,6 +175,13 @@ func NewCommand() *cobra.Command {
 			Desc:    "path for boltdb database",
 			Short:   'm',
 		},
+		{
+			DestP:   &options.target.configsPath,
+			Flag:    "configs-path",
```
danxmoran, line 180: Could we rename the flag to be even more specific, i.e. `influx-configs-path`? I'm worried about confusion with the existing `config-file` argument.

cmd/influxd/upgrade/upgrade.go
```diff
@@ -149,6 +175,13 @@ func NewCommand() *cobra.Command {
 			Desc:    "path for boltdb database",
```
danxmoran, line 175: It's too far away from the diff for me to comment on directly, but could you add a point `3` to the list at [this line](https://github.com/influxdata/influxdb/pull/19937/files#diff-e5c52e2be1a2771a344e5d0402468658d34a70fe9ba50453edb3b3b3e97a4eaeR149) describing the new behavior?

alespour: I hope I got this correctly (wasn't sure where the link point to).

danxmoran: Yup, that was it

danxmoran (approved):

cmd/influxd/upgrade/upgrade.go
```diff
@@ -40,12 +41,36 @@ type configV1 struct {
 		Dir    string `toml:"dir"`
 		WALDir string `toml:"wal-dir"`
 	} `toml:"data"`
+	Http struct {
```
docmerlin, line 44: please don't build the url using string concatination.  Please use the `net.URL` object.

cmd/influxd/upgrade/upgrade_test.go
```diff
@@ -6,6 +6,9 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/BurntSushi/toml"
+	"github.com/google/go-cmp/cmp"
+
```
docmerlin, line 11: remove this line please

docmerlin (changes requested):

dgnorton (approved):

danxmoran (approved):
Re-LGTM

PR #19971, opened 2020-11-10
chore: remove Invites from swagger

121watts:
reference: https://github.com/influxdata/ui/pull/293

removes `Invite` and `Invites` components from `http/swagger.yml`


randycoulman (approved):
👍

Palakp41 (approved):
👍

PR #19972, opened 2020-11-10
fix(upgrade): remove obsolete unused option (influx-command-path)

### LINKED ISSUE #19973: remove unused option from influxd upgrade command
`influx-command-path` is no longer used since security upgrade is done via v1 auth API instead of a script calling `influx` exe.

alespour:
Closes #19973

Removes obsolete unused option `influx-command-path`. which was previously used in security upgrade, but since it is now done at runtime using v1 auth API, this option is obsolete and not used at all.

- [X] Rebased/mergeable
- [X] Tests pass


danxmoran (approved):
Nice!

PR #19979, opened 2020-11-10
feat(tasks): added functionality to filter runs by time

### LINKED ISSUE #19247: influxdb_client_2.0.0-beta.15_darwin_amd64 task run list not working as expected

Nav-aggarwal09:
Closes #19247

This PR will respect the before/after time filters for runs. 
Note that this mirrors the already (new) added functionality from idpe. See influxdata/idpe#8973

- [ ] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/kv/task.go
+++ b/kv/task.go
@@ -951,6 +951,21 @@ func (s *Service) findRuns(ctx context.Context, tx Tx, filter influxdb.RunFilter
 	if filter.Limit < 0 || filter.Limit > influxdb.TaskMaxPageSize {
 		return nil, 0, influxdb.ErrOutOfBoundsLimit
 	}
+	parsedFilterAfterTime := time.Time{}
+	parsedFilterBeforeTime := time.Now().UTC()
+	var err error
+	if filter.AfterTime != "" {
+		parsedFilterAfterTime, err = time.Parse(time.RFC3339, filter.AfterTime)
+		if err != nil {
+			return nil, 0, err
+		}
+	}
+	if filter.BeforeTime != "" {
+		parsedFilterBeforeTime, err = time.Parse(time.RFC3339, filter.BeforeTime)
+		if err != nil {
+			return nil, 0, err
+		}
+	}
 
 	var runs []*influxdb.Run
 	// manual runs
@@ -959,7 +974,9 @@ func (s *Service) findRuns(ctx context.Context, tx Tx, filter influxdb.RunFilter
 		return nil, 0, err
 	}
 	for _, run := range manualRuns {
-		runs = append(runs, run)
+		if run.ScheduledFor.After(parsedFilterAfterTime) && run.ScheduledFor.Before(parsedFilterBeforeTime) {
+			runs = append(runs, run)
+		}
 		if len(runs) >= filter.Limit {
 			return runs, len(runs), nil
 		}
@@ -971,13 +988,16 @@ func (s *Service) findRuns(ctx context.Context, tx Tx, filter influxdb.RunFilter
 		return nil, 0, err
 	}
 	for _, run := range currentlyRunning {
-		runs = append(runs, run)
+		if run.ScheduledFor.After(parsedFilterAfterTime) && run.ScheduledFor.Before(parsedFilterBeforeTime) {
+			runs = append(runs, run)
+		}
 		if len(runs) >= filter.Limit {
 			return runs, len(runs), nil
 		}
 	}
 
 	return runs, len(runs), nil
+
 }
 
 // FindRunByID returns a single run.
--- a/kv/task.go
+++ b/kv/task.go
@@ -954,7 +954,7 @@ func (s *Service) findRuns(ctx context.Context, tx Tx, filter influxdb.RunFilter
 	parsedFilterAfterTime := time.Time{}
 	parsedFilterBeforeTime := time.Now().UTC()
 	var err error
-	if filter.AfterTime != "" {
+	if len(filter.AfterTime) > 0 {
 		parsedFilterAfterTime, err = time.Parse(time.RFC3339, filter.AfterTime)
 		if err != nil {
 			return nil, 0, err
@@ -997,7 +997,6 @@ func (s *Service) findRuns(ctx context.Context, tx Tx, filter influxdb.RunFilter
 	}
 
 	return runs, len(runs), nil
-
 }
 
 // FindRunByID returns a single run.
--- a/task/backend/analytical_storage.go
+++ b/task/backend/analytical_storage.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/influxdata/influxdb/v2/kit/errors"
 	"time"
 
 	"github.com/influxdata/flux"
@@ -154,23 +155,26 @@ func (as *AnalyticalStorage) FindRuns(ctx context.Context, filter influxdb.RunFi
 
 	// creates flux script to filter based on time, if given
 	constructedTimeFilter := ""
-	parsedAfterTime := time.Time{}
-	parsedBeforeTime := time.Now()
-	if len(filter.AfterTime) > 0 {
-		parsedAfterTime, err = time.Parse(time.RFC3339, filter.AfterTime)
-		if err != nil {
-			return nil, 0, fmt.Errorf("failed parsing after time: %s", err.Error())
+	if len(filter.AfterTime) > 0 || len(filter.BeforeTime) > 0 {
+		parsedAfterTime := time.Time{}
+		parsedBeforeTime := time.Now()
+		if len(filter.AfterTime) > 0 {
+			parsedAfterTime, err = time.Parse(time.RFC3339, filter.AfterTime)
+			if err != nil {
+				return nil, 0, fmt.Errorf("failed parsing after time: %s", err.Error())
+			}
 		}
-	}
-	if len(filter.BeforeTime) > 0 {
-		parsedBeforeTime, err = time.Parse(time.RFC3339, filter.BeforeTime)
-		if err != nil {
-			return nil, 0, fmt.Errorf("failed parsing before time: %s", err.Error())
+		if len(filter.BeforeTime) > 0 {
+			parsedBeforeTime, err = time.Parse(time.RFC3339, filter.BeforeTime)
+			if err != nil {
+				return nil, 0, fmt.Errorf("failed parsing before time: %s", err.Error())
+			}
+
+		}
+		if !parsedBeforeTime.After(parsedAfterTime) {
+			return nil, 0, errors.New("Given after time must be prior to before time")
 		}
 
-	}
-
-	if parsedAfterTime.IsZero() || len(filter.BeforeTime) > 0 {
 		constructedTimeFilter = fmt.Sprintf(
 			`|> filter(fn: (r) =>time(v: r["scheduledFor"]) > %s and time(v: r["scheduledFor"]) < %s)`,
 			parsedAfterTime.Format(time.RFC3339),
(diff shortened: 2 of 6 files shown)


kv/task.go
```diff
@@ -951,6 +951,21 @@ func (s *Service) findRuns(ctx context.Context, tx Tx, filter influxdb.RunFilter
 	if filter.Limit < 0 || filter.Limit > influxdb.TaskMaxPageSize {
 		return nil, 0, influxdb.ErrOutOfBoundsLimit
 	}
+	parsedFilterAfterTime := time.Time{}
+	parsedFilterBeforeTime := time.Now().UTC()
+	var err error
+	if filter.AfterTime != "" {
```
glinton, line 957: nit: the following is a marginally faster check for the same thing:
```suggestion
	if len(filter.AfterTime) > 0 {
```

kv/task.go
```diff
@@ -971,13 +988,16 @@ func (s *Service) findRuns(ctx context.Context, tx Tx, filter influxdb.RunFilter
...
+		if run.ScheduledFor.After(parsedFilterAfterTime) && run.ScheduledFor.Before(parsedFilterBeforeTime) {
+			runs = append(runs, run)
+		}
 		if len(runs) >= filter.Limit {
 			return runs, len(runs), nil
 		}
 	}
 
 	return runs, len(runs), nil
+
```
glinton, line 1000: nit: :scissors: ime, it's more beneficial to add spacing inside the function between logical blocks, than to add a space as the first or last line in a function. Julia Ferraioli's "Writing Accessible Go" is a great resource to look at for more accessibility tips.

task/backend/analytical_storage.go
```diff
@@ -152,18 +153,45 @@ func (as *AnalyticalStorage) FindRuns(ctx context.Context, filter influxdb.RunFi
 		filterPart = fmt.Sprintf(`|> filter(fn: (r) => r.runID > %q)`, filter.After.String())
 	}
 
+	parsedAfterTime := time.Time{}
+	parsedBeforeTime := time.Now()
```
glinton, line 157: nit: move these inside the `if` block, since they are only referenced within that.

task/backend/analytical_storage.go
```diff
@@ -152,18 +153,45 @@ func (as *AnalyticalStorage) FindRuns(ctx context.Context, filter influxdb.RunFi
...
+	constructedTimeFilter := ""
+	if feature.TimeFilterFlags().Enabled(ctx) {
+		if filter.AfterTime != "" {
+			tmpParsedAfter, err := time.Parse(time.RFC3339, filter.AfterTime)
```
glinton, line 161: a common trick to avoid a temporary variable where the error is returned is something like the following
```suggestion
			var err error // define the error beforehand, if needed
			// make the assignments
			parsedAfterTime, err = time.Parse(time.RFC3339, filter.AfterTime)
```

task/backend/analytical_storage.go
```diff
@@ -152,18 +153,45 @@ func (as *AnalyticalStorage) FindRuns(ctx context.Context, filter influxdb.RunFi
...
+			if err != nil {
+				return nil, 0, err
```
glinton, line 163: it may be beneficial to add some context to this error to help the user differentiate between this and the before time parsing
```suggestion
				return nil, 0, fmt.Errorf("failed parsing after time: %s", err.Error())
```

task/servicetest/servicetest.go
```diff
@@ -902,7 +909,7 @@ func testTaskRuns(t *testing.T, sys *System) {
 			t.Fatalf("failed to error with out of bounds run limit: %d", influxdb.TaskMaxPageSize+1)
 		}
 
-		requestedAt := time.Now().Add(5 * time.Minute).UTC() // This should guarantee we can make two runs.
+		requestedAt := time.Now().Add(time.Hour * -1).UTC() // This should guarantee we can make two runs.
```
glinton, line 912: was turning these positive time values into a negative one intentional?

glinton (commented):
nice work. :heavy_check_mark:

task/servicetest/servicetest.go
Nav-aggarwal09: Yes since we our default time constraints are beginning of time and time.now(), any attempt to find runs in the future would result in an empty list returned.

Nav-aggarwal09:
@glinton good reviews. Should I reopen the mirrored branch on idpe and make the same changes and set you as a reviewer?

task/backend/analytical_storage.go
```diff
@@ -153,25 +153,24 @@ func (as *AnalyticalStorage) FindRuns(ctx context.Context, filter influxdb.RunFi
 		filterPart = fmt.Sprintf(`|> filter(fn: (r) => r.runID > %q)`, filter.After.String())
 	}
 
-	parsedAfterTime := time.Time{}
-	parsedBeforeTime := time.Now()
 	constructedTimeFilter := ""
 	if feature.TimeFilterFlags().Enabled(ctx) {
```
glinton, line 157: > Should I reopen the mirrored branch on idpe and make the same changes and set you as a reviewer?

Completely up to you as many of these were small nits rather than functionality changes. that reminds me though, feature flags are a cloud concept and can be removed from oss (is my understanding, and should be verified)

task/backend/analytical_storage.go
```diff
@@ -153,25 +152,25 @@ func (as *AnalyticalStorage) FindRuns(ctx context.Context, filter influxdb.RunFi
...
-		var err error
-		if filter.AfterTime != "" {
-			parsedAfterTime, err = time.Parse(time.RFC3339, filter.AfterTime)
-			if err != nil {
-				return nil, 0, fmt.Errorf("failed parsing after time: %s", err.Error())
-			}
-
+	parsedAfterTime := time.Time{}
+	parsedBeforeTime := time.Now()
+	if len(filter.AfterTime) > 0 {
```
glinton, line 159: excellent!

task/backend/analytical_storage.go
```diff
@@ -153,25 +152,25 @@ func (as *AnalyticalStorage) FindRuns(ctx context.Context, filter influxdb.RunFi
...
+	}
+	if len(filter.BeforeTime) > 0 {
+		parsedBeforeTime, err = time.Parse(time.RFC3339, filter.BeforeTime)
+		if err != nil {
+			return nil, 0, fmt.Errorf("failed parsing before time: %s", err.Error())
 		}
+
+	}
+
+	if parsedAfterTime.IsZero() || len(filter.BeforeTime) > 0 {
```
glinton, line 173: This check is slightly confusing, did you mean to ensure that the `parsedBeforeTime` comes after the `parsedAfterTime`:
```go
if parsedBeforeTime.After(parsedAfterTime) {
```

kv/task.go
```diff
@@ -951,6 +951,21 @@ func (s *Service) findRuns(ctx context.Context, tx Tx, filter influxdb.RunFilter
...
+		if err != nil {
+			return nil, 0, err
+		}
+	}
+	if filter.BeforeTime != "" {
+		parsedFilterBeforeTime, err = time.Parse(time.RFC3339, filter.BeforeTime)
+		if err != nil {
+			return nil, 0, err
+		}
+	}
```
glinton, line 968: nit: this logic could someday be moved into a function to dry up the code (it's reused in `task/backend/analytical_storage.go:157`)

glinton (changes requested):
The only change i'm actually requesting is the check before setting the `constructedTimeFilter`. Previously, there was no check, which was ok and the query would error later if the values were mixed up, but checking here could save that execution. "return early" and whatnot.

This is really looking great though, nice work!

task/backend/analytical_storage.go
Nav-aggarwal09: Its intended to check whether there are any time filters specified. If either are specified (aka not nil/0), then construct the flux script. If none, then proceed with empty string

kv/task.go
Nav-aggarwal09: Unsure how I would go about doing that as they are in different packages/folders. Would I just make two of the same helper functions in each file?

task/backend/analytical_storage.go
```diff
@@ -152,18 +152,44 @@ func (as *AnalyticalStorage) FindRuns(ctx context.Context, filter influxdb.RunFi
...
+		}
+
+	}
+
+	if parsedAfterTime.IsZero() || len(filter.BeforeTime) > 0 {
+		constructedTimeFilter = fmt.Sprintf(
+			`|> filter(fn: (r) =>time(v: r["scheduledFor"]) > %s and time(v: r["scheduledFor"]) < %s)`,
+			parsedAfterTime.Format(time.RFC3339),
+			parsedBeforeTime.Format(time.RFC3339))
+	}
```
glinton, line 178: The check as written is essentially `if there is no AfterTime set or BeforeTime is set`. With your default zero value for `parsedAfterTime`, this check would always pass (negative time would return an error above. With your defaults, the constructed time filter (and resulting query) would be valid and could return valid data. However, if the user accidentally flips the times (something like `after = now()` and `before = now()-5m`), the query would never return any data, potentially leaving a confused user.

In order to provide meaningful feedback pre-emptively, you could modify the check and do something like:
```suggestion
	if !parsedBeforeTime.After(parsedAfterTime) {
		return nil, 0, errors.New("after time must be prior to before time") // or some other useful error
	}
	
	constructedTimeFilter = fmt.Sprintf(
		`|> filter(fn: (r) =>time(v: r["scheduledFor"]) > %s and time(v: r["scheduledFor"]) < %s)`,
		parsedAfterTime.Format(time.RFC3339),
		parsedBeforeTime.Format(time.RFC3339))
```

Or remove the check entirely

kv/task.go
glinton: i'm not sure the proper location (maybe in this same file in the `kit` package that you would then import in `task/backend/analytical_storage`), but defining the logic once to a central location and importing/calling it in others helps when that logic needs updated. no need to worry about it this go

task/backend/analytical_storage.go
```diff
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/influxdata/influxdb/v2/kit/errors"
```
glinton, line 8: TIL :) i was thinking the one from stdlib, but this works great.

task/backend/analytical_storage.go
```diff
@@ -154,23 +155,26 @@ func (as *AnalyticalStorage) FindRuns(ctx context.Context, filter influxdb.RunFi
...
+		if len(filter.BeforeTime) > 0 {
+			parsedBeforeTime, err = time.Parse(time.RFC3339, filter.BeforeTime)
+			if err != nil {
+				return nil, 0, fmt.Errorf("failed parsing before time: %s", err.Error())
+			}
 
-	}
+		}
+		if !parsedBeforeTime.After(parsedAfterTime) {
+			return nil, 0, errors.New("Given after time must be prior to before time")
```
glinton, line 175: errors should begin with a lowercase letter. learn more at https://github.com/golang/go/wiki/CodeReviewComments#error-strings

task/backend/analytical_storage.go
Nav-aggarwal09: I actually had no idea I imported that.. I thought I was using the stdlib 😄

task/backend/analytical_storage.go
Nav-aggarwal09: I did a little combo of both. I realized my logic was wrong, but also saw how a user could give faulty data and mess up the constructed string

glinton (approved):
great work on this

Nav-aggarwal09:
Thank you for your quick and great feedback! @glinton

PR #19980, opened 2020-11-10
fix: check write permission in legacy write path

### LINKED ISSUE #19974: Upgrade: User with a read only permission should not be able to write into a bucket
InfluxDB 1.x can have users that are only able to write into a database or only read from a database. If we have a user that has only a read permission, then after upgrade to 2.0 this same user will be able to write into the bucket, even though it does not have permission to do that.
Setup:
- Install InfluxDB 1.8.3 (latest OSS)
- Update configuration to use authentication under [http] section if the influxdb.conf file:
   ```
   # Determines whether user authentication is enabled over HTTP/HTTPS.
   auth-enabled = true
   ```
- Start influxdb v1 service
- Create an admin user and restart the service
- Create a user `A` with a `READ` permission
- Create a user `B` with a `WRITE` permission
- Download v2.0.0 InfluxDB
- Run upgrade
- Write data into migrated database as a user with READ only permissions:
   ```
   curl -vv -u read_test_db_1:read_test_db_1 -X POST http://localhost:8086/write?db=test_db_1 --data-binary "m,r=1 f=1"
   
   *   Trying 127.0.0.1...
   * TCP_NODELAY set
  * Connected to localhost (127.0.0.1) port 8086 (#0). 
   * Server auth using Basic with user 'read_test_db_1'
   POST /write?db=test_db_1 HTTP/1.1Host: localhost:8086
   Authorization: Basic cmVhZF90ZXN0X2RiXzE6cmVhZF90ZXN0X2RiXzE=User-Agent: curl/7.58.0
   Accept: */*
  Content-Length: 9
  Content-Type: application/x-www-form-urlencoded
  * upload completely sent off: 9 out of 9 bytes
    HTTP/1.1 204 No Content
    Date: Tue, 10 Nov 2020 18:30:57 GMT

   * Connection #0 to host localhost left intact
   ```

When running `influx v1 auth list` command it mapps READ and WRITE users to admin user:
```
root@ip-172-31-14-196:/tmp# /usr/bin/influx v1 auth list
ID			Description			Name / Token	User Name	User ID			Permissions
0698cb720526e000	read_test_db_1's Legacy Token	read_test_db_1	admin_user	0698cb71dde6e000	[read:orgs/31ee98c80563ba36/buckets/b9a96022e9ce3e1e read:orgs/31ee98c80563ba36/buckets/5593dffd6d852fef]
0698cb7205e6e000	write_test_db_1's Legacy Token	write_test_db_1	admin_user	0698cb71dde6e000	[write:orgs/31ee98c80563ba36/buckets/b9a96022e9ce3e1e write:orgs/31ee98c80563ba36/buckets/5593dffd6d852fef]
```

danxmoran:
Closes #19974 

This is a big of a hack, it doesn't fit how we structure our auth checks elsewhere. Suggestions appreciated!

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influxd/launcher/launcher.go
+++ b/cmd/influxd/launcher/launcher.go
@@ -954,8 +954,7 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 		}
 	}
 
-	// N.B. the BucketService used by the DBRP service doesn't perform authorization.
-	dbrpSvc := dbrp.NewAuthorizedService(dbrp.NewService(ctx, ts.BucketService, m.kvStore))
+	dbrpSvc := dbrp.NewAuthorizedService(dbrp.NewService(ctx, authorizer.NewBucketService(ts.BucketService), m.kvStore))
 
 	cm := iqlcontrol.NewControllerMetrics([]string{})
 	m.reg.MustRegister(cm.PrometheusCollectors()...)
--- a/http/legacy/write_handler.go
+++ b/http/legacy/write_handler.go
@@ -2,7 +2,7 @@ package legacy
 
 import (
 	"context"
-	"github.com/influxdata/influxdb/v2/authorizer"
+	"fmt"
 	"io"
 	"net/http"
 
@@ -129,6 +129,11 @@ func (h *WriteHandler) handleWrite(w http.ResponseWriter, r *http.Request) {
 	}
 	span.LogKV("bucket_id", bucket.ID)
 
+	if err := checkBucketWritePermissions(auth, bucket.OrgID, bucket.ID); err != nil {
+		h.HandleHTTPError(ctx, err, sw)
+		return
+	}
+
 	parsed, err := points.NewParser(req.Precision).Parse(ctx, auth.OrgID, bucket.ID, req.Body)
 	if err != nil {
 		h.HandleHTTPError(ctx, err, sw)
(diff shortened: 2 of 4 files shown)


cmd/influxd/launcher/launcher.go
```diff
@@ -954,8 +954,7 @@ func (m *Launcher) run(ctx context.Context) (err error) {
 		}
 	}
 
-	// N.B. the BucketService used by the DBRP service doesn't perform authorization.
-	dbrpSvc := dbrp.NewAuthorizedService(dbrp.NewService(ctx, ts.BucketService, m.kvStore))
+	dbrpSvc := dbrp.NewAuthorizedService(dbrp.NewService(ctx, authorizer.NewBucketService(ts.BucketService), m.kvStore))
```
danxmoran, line 957: Adding this back in didn't fix the problem, but I was nervous about what other unknown effects it might be having.

http/legacy/write_handler.go
```diff
@@ -155,7 +156,15 @@ func (h *WriteHandler) findBucket(ctx context.Context, orgID influxdb.ID, db, rp
 		return nil, err
 	}
 
-	return h.BucketService.FindBucketByID(ctx, mapping.BucketID)
+	bucket, err := h.BucketService.FindBucketByID(ctx, mapping.BucketID)
+	if err != nil {
+		return nil, err
+	}
+	_, _, err = authorizer.AuthorizeWrite(ctx, influxdb.BucketsResourceType, bucket.ID, bucket.OrgID)
```
danxmoran, line 163: This is the only hard-coded auth check at this layer, all others are abstracted away behind various middlewares. Not sure if/how it could be refactored to better fit existing patterns.

http/legacy/write_handler_test.go
```diff
@@ -94,7 +93,7 @@ func TestWriteHandler_BucketAndMappingExists(t *testing.T) {
 	handler := NewWriterHandler(&PointsWriterBackend{
 		HTTPErrorHandler:   DefaultErrorHandler,
 		Logger:             zaptest.NewLogger(t),
-		BucketService:      authorizer.NewBucketService(bucketService),
+		BucketService:      bucketService,
```
danxmoran, line 96: I believe this accurately reflects the setup constructed by the `influxd` launcher. It'd be nice to have one setup function that both the tests and main code could call, to ensure they stay in sync.

http/legacy/write_handler_test.go
```diff
@@ -105,6 +104,84 @@ func TestWriteHandler_BucketAndMappingExists(t *testing.T) {
 	assert.Equal(t, "", w.Body.String())
 }
 
+func TestWriteHandler_BucketAndMappingExistsNoPermissions(t *testing.T) {
```
docmerlin, line 107: Yes, good, this should help a lot.

http/legacy/write_handler.go
stuartcarnie: This is great and does follow the same pattern as `/api/v2/write`:

https://github.com/influxdata/influxdb/blob/fcb70c26d54b78d8da1bab9300273ae311f6c600/http/write_handler.go#L174-L177

Can you make the changes so the code looks the same – i.e. move this check outside the `findBucket` function into the `handleWrite` function after the `findBucket` call?

stuartcarnie (changes requested):
This is great – I'd suggested a minor change to follow the same pattern as the v2 API implementation of the write handler.

stuartcarnie (approved):
Solid work – thanks for sorting this out and 💯 for adding additional tests

Issue #19429, opened 2020-08-25
Port influx generate command


stuartcarnie:
## What

Port the `influxd generate` command, which is compatible with InfluxDB 1.x TSM files.

The `pkg/data/gen` package was ported from InfluxDB 1.x. Relevant components must be forward ported from 1.x to write to multiple shards and the `influxd generate` re-enabled.

singlaive:
Any update on this?
Do we have rough timeline when the generate command is back?

PR #19995, opened 2020-11-11
fix(cmd/influxd): suggest running with -h on error instead of printing usage

### LINKED ISSUE #19922: Do not automatically print the help docs for influxd command
when there is an error calling the influxd command, we should not automatically print the help text. we should follow the same conventions as the `influx` command.

danxmoran:
Closes #19922 

I'm not incredibly happy with this whack-a-mole approach, but it seemed like the most minimal approach. I'm going to open a separate issue for refactoring the CLI code here.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influxd/inspect/export_index.go
+++ b/cmd/influxd/inspect/export_index.go
@@ -16,6 +16,7 @@ func NewExportIndexCommand() *cobra.Command {
 		Long: `
 This command will export all series in a TSI index to
 SQL format for easier inspection and debugging.`,
+		Args: cobra.NoArgs,
 	}
 
 	var seriesFilePath, dataPath string
--- a/cmd/influxd/launcher/launcher.go
+++ b/cmd/influxd/launcher/launcher.go
@@ -22,7 +22,6 @@ import (
 	"github.com/influxdata/influxdb/v2/bolt"
 	"github.com/influxdata/influxdb/v2/checks"
 	"github.com/influxdata/influxdb/v2/chronograf/server"
-	"github.com/influxdata/influxdb/v2/cmd/influxd/inspect"
 	"github.com/influxdata/influxdb/v2/dashboards"
 	dashboardTransport "github.com/influxdata/influxdb/v2/dashboards/transport"
 	"github.com/influxdata/influxdb/v2/dbrp"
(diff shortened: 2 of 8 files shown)


cmd/influxd/main.go
```diff
@@ -28,23 +29,25 @@ func main() {
 
 	influxdb.SetBuildInfo(version, commit, date)
 
-	rootCmd := launcher.NewInfluxdCommand(context.Background(),
-		// FIXME
-		//generate.Command,
```
danxmoran, line 33: Issue: #19429

cmd/influxd/main.go
```diff
@@ -28,23 +29,25 @@ func main() {
...
-		//restore.Command,
```
danxmoran, line 34: (As far as I can tell) this was replaced by the backup/restore commands in the `influx` CLI, so it's fine to remove the FIXME without a tracking issue

cmd/influxd/inspect/export_index.go
```diff
@@ -16,6 +16,7 @@ func NewExportIndexCommand() *cobra.Command {
 		Long: `
 This command will export all series in a TSI index to
 SQL format for easier inspection and debugging.`,
+		Args: cobra.NoArgs,
```
danxmoran, line 19: This is the magic needed to trigger errors on unknown / typoed subcommands

cmd/influxd/launcher/launcher.go
```diff
@@ -178,7 +178,6 @@ var vaultConfig vault.Config
 
 func setLauncherCMDOpts(l *Launcher, cmd *cobra.Command) {
 	cli.BindOptions(cmd, launcherOpts(l))
-	cmd.AddCommand(inspect.NewCommand())
```
danxmoran, line 181: Before this, the code was registering both `influxd inspect` and `influxd run inspect` commands. That felt like a mistake, but this technically a breaking change.  I felt decently confident making the change because the public docs site only list `influxd inspect`, but I can revert if needed.

cmd/influxd/launcher/launcher.go
```diff
@@ -101,7 +100,7 @@ const (
 	MaxInt = 1<<uint(strconv.IntSize-1) - 1
 )
 
-func NewInfluxdCommand(ctx context.Context, subCommands ...*cobra.Command) *cobra.Command {
```
danxmoran, line 104: The humble beginnings of my quest to simplify `launcher.go`...

cmd/influxd/launcher/launcher.go
stuartcarnie: This is good with me; `run` shouldn't have any subcommands.

cmd/influxd/launcher/launcher.go
stuartcarnie: And a noble quest it is!

stuartcarnie (approved):
These a nice UX improvements 👏

PR #20002, opened 2020-11-12
chore: Do not use global viper APIs, which breaks testing

### LINKED ISSUE #19998: Port test pipeline and InfluxQL test suite
## What

Port the Cloud implementation of the InfluxQL test suite to OSS

- [ ] `tests` package to provide the `Pipeline` framework
- [ ] `influxql/_v1Tests` package, which provides extensive coverage of the InfluxQL query engine
- [ ] `influxql/_v1Validation` package, which provides a configuration framework for additional test coverage

stuartcarnie:
This PR removes the use of global viper state, by passing instances of `viper.Viper` throughout the CLI tools. This change is required to enable the `TestLauncher` to be used reliably and without data races when running test suites in parallel.

This is the first step in 3 PRs to bring the full pipeline test suite into OSS in order to close #19998.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/cmd/influxd/inspect/inspect.go
+++ b/cmd/influxd/inspect/inspect.go
@@ -2,10 +2,11 @@ package inspect
 
 import (
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 // NewCommand creates the new command.
-func NewCommand() *cobra.Command {
+func NewCommand(_ *viper.Viper) *cobra.Command {
 	base := &cobra.Command{
 		Use:   "inspect",
 		Short: "Commands for inspecting on-disk database data",
--- a/cmd/influxd/inspect/inspect.go
+++ b/cmd/influxd/inspect/inspect.go
@@ -2,11 +2,10 @@ package inspect
 
 import (
 	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
 )
 
 // NewCommand creates the new command.
-func NewCommand(_ *viper.Viper) *cobra.Command {
+func NewCommand() *cobra.Command {
 	base := &cobra.Command{
 		Use:   "inspect",
 		Short: "Commands for inspecting on-disk database data",
--- a/cmd/influxd/main.go
+++ b/cmd/influxd/main.go
@@ -3,17 +3,18 @@ package main
 import (
 	"context"
 	"fmt"
-	"github.com/influxdata/influxdb/v2/cmd/influxd/inspect"
 	_ "net/http/pprof"
 	"os"
 	"time"
 
 	"github.com/influxdata/influxdb/v2"
+	"github.com/influxdata/influxdb/v2/cmd/influxd/inspect"
 	"github.com/influxdata/influxdb/v2/cmd/influxd/launcher"
 	"github.com/influxdata/influxdb/v2/cmd/influxd/upgrade"
 	_ "github.com/influxdata/influxdb/v2/tsdb/engine/tsm1"
 	_ "github.com/influxdata/influxdb/v2/tsdb/index/tsi1"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 var (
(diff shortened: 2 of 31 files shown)


cmd/influxd/inspect/inspect.go
```diff
@@ -2,10 +2,11 @@ package inspect
 
 import (
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 // NewCommand creates the new command.
-func NewCommand() *cobra.Command {
+func NewCommand(_ *viper.Viper) *cobra.Command {
```
danxmoran, line 9: Feels inconsistent to me to require the viper arg here (but not use it), and not do the same for `NewExportIndexCommand()`. Is there a reason why it needs to be passed here?

cmd/influxd/main.go
```diff
@@ -29,10 +30,11 @@ func main() {
...
 
-	rootCmd := launcher.NewInfluxdCommand(context.Background())
+	v := viper.New()
+	rootCmd := launcher.NewInfluxdCommand(context.Background(), v)
 	// upgrade binds options to env variables, so it must be added after rootCmd is initialized
-	rootCmd.AddCommand(upgrade.NewCommand())
-	rootCmd.AddCommand(inspect.NewCommand())
+	rootCmd.AddCommand(upgrade.NewCommand(v))
+	rootCmd.AddCommand(inspect.NewCommand(v))
 	rootCmd.AddCommand(versionCmd())
```
danxmoran, line 38: Does the version command need the viper for anything?

cmd/influxd/upgrade/upgrade.go
```diff
@@ -258,7 +259,7 @@ func NewCommand() *cobra.Command {
 		},
 	}
 
-	cli.BindOptions(cmd, opts)
+	cli.BindOptions(v, cmd, opts)
 	// add sub commands
 	cmd.AddCommand(v1DumpMetaCommand)
 	cmd.AddCommand(v2DumpMetaCommand)
```
danxmoran, line 265: Do these 2 subcommands need the viper for anything?

kit/cli/viper.go
```diff
@@ -63,40 +63,40 @@ func NewCommand(p *Program) *cobra.Command {
 		},
 	}
 
-	viper.SetEnvPrefix(strings.ToUpper(p.Name))
-	viper.AutomaticEnv()
+	v.SetEnvPrefix(strings.ToUpper(p.Name))
+	v.AutomaticEnv()
```
danxmoran, line 67: Side note: I noticed `cmd/influx` duplicates some of this logic, should I open a tech-debt ticket to port it over to `kit/cli`?

danxmoran (commented):
Looks great! Left some comments to make sure we aren't missing spots

cmd/influxd/upgrade/upgrade.go
stuartcarnie: No, they should probably be removed actually.

kit/cli/viper.go
stuartcarnie: Yes, I think that is a good idea

cmd/influxd/main.go
stuartcarnie: It doesn't, as it just returns a struct

cmd/influxd/inspect/inspect.go
stuartcarnie: I'll remove it for now – originally I wanted the consistency at the call site for all the `NewCommand` calls, and given there are a large number of subcommands to be added here, it was possible one might require it. We can leave that for another day.

stuartcarnie: Done – removed the unused argument

kit/cli/viper.go
danxmoran: #20013

danxmoran (approved):

PR #20015, opened 2020-11-12
fix(http): Add same site strict flag to session cookie

pierwill:
See https://golang.org/pkg/net/http/#SameSite.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/http/session_handler.go
+++ b/http/session_handler.go
@@ -176,9 +176,10 @@ func decodeCookieSession(ctx context.Context, r *http.Request) (string, error) {
 // SetCookieSession adds a cookie for the session to an http request
 func SetCookieSession(key string, r *http.Request) {
 	c := &http.Cookie{
-		Name:   cookieSessionName,
-		Value:  key,
-		Secure: true,
+		Name:     cookieSessionName,
+		Value:    key,
+		Secure:   true,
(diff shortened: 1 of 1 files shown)


http/session_handler.go
```diff
@@ -176,9 +176,10 @@ func decodeCookieSession(ctx context.Context, r *http.Request) (string, error) {
 // SetCookieSession adds a cookie for the session to an http request
 func SetCookieSession(key string, r *http.Request) {
 	c := &http.Cookie{
-		Name:   cookieSessionName,
-		Value:  key,
-		Secure: true,
+		Name:     cookieSessionName,
+		Value:    key,
+		Secure:   true,
+		SameSite: 3, // SameSiteStrictMode
```
glinton, line 182: Rather than using the number, use the self-documenting variable.
```suggestion
		SameSite: http.SameSiteStrictMode,
```

glinton (changes requested):
Somewhere in this pr, it would be great to document why strict mode was chosen over lax.

glinton (approved):
I'd still say document why you chose "strict" over "lax" for posterity's sake (ideally it would be configurable to suit individual user's needs :thinking:)

pierwill:
Using "strict" here because it is more restrictive.

PR #20012, opened 2020-11-12
fix(cmd/influxd): validate all input paths to `upgrade` up-front

danxmoran:
Part of #19985

Validate existence / nonexistence of source and target paths provided to the `upgrade` command before doing any work, to avoid annoyance.

- [ ] [CHANGELOG.md](https://github.com/influxdata/influxdb/blob/master/CHANGELOG.md) updated with a link to the PR (not the Issue)
- [ ] [Well-formatted commit messages](https://www.conventionalcommits.org/en/v1.0.0-beta.3/)
- [ ] Rebased/mergeable
- [ ] Tests pass
- [ ] http/swagger.yml updated (if modified Go structs or API)
- [ ] Feature flagged (if modified API)
- [ ] Documentation updated or issue created (provide link to issue/pr)
- [ ] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)

PR #20052, opened 2020-11-16
fix(cmd/influxd): don't validate unused paths in `upgrade`

### LINKED ISSUE #20037: Incorrect input validation in `influxd upgrade`
__Steps to reproduce:__

1. Run `influxd upgrade` pointing to a `--v1-dir` that doesn't exist, and a `--config-file` containing V1 paths that _do_ exist

__Expected behavior:__
As specified by the long help text, `--v1-dir` should only be used if `--config-file` isn't given. Using the two together should ideally be an error. If not that, we shouldn't raise a validation error on missing `--v1-dir` paths when we have valid config.

__Actual behavior:__
The command exists with an error because the `--v1-dir`, or its expected sub-directories don't exist.

This regression was caused by the refactoring in #20012

danxmoran:
Closes #20037 

The help text for `influxd upgrade` says `--v1-dir` is only used if `--config-file` isn't passed, but the validation logic I added in #20012 always validated `--v1-dir`. This fixes the behavior.

I had to refactor config-loading logic to support loading V1 config separately from translating it to V2, so we could still run all validation checks before writing any new files to disk.

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influxd/upgrade/config_test.go
+++ b/cmd/influxd/upgrade/config_test.go
@@ -1,6 +1,7 @@
 package upgrade
 
 import (
+	"github.com/influxdata/influxdb/v2/pkg/testing/assert"
 	"io/ioutil"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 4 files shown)


cmd/influxd/upgrade/config_test.go
```diff
@@ -1,59 +1,44 @@
 package upgrade
 
 import (
+	"github.com/influxdata/influxdb/v2/pkg/testing/assert"
```
stuartcarnie, line 4: nit: `goimports -w`

stuartcarnie (approved):
These improvements look good to me

PR #20064, opened 2020-11-17
fix: Ensure Flux reads across all shards

### LINKED ISSUE #20035: Flux stops reading results on first shard with no data
__Steps to reproduce:__
1. Download [tests.zip](https://github.com/influxdata/influxdb/files/5538677/tests.zip), it contains two annotated csv files, test1.csv and test4.csv
2. Add test1.csv to influxdb using `influx write` Note that there are 80 records in this file spanning a year. Confirm that there are now 80 records in this measurement using the following query
```
from(bucket: "test")
  |> range(start: -1y)
  |> filter(fn: (r) => r["_measurement"] == "Measurement1")
  |> drop(columns: ["team", "scope"])
  |> count()
```
3. Now add test4.csv to influxdb. Run the query above again. Note that there are now only 18 records, even though test4 only adds records to Measurement4.

__Expected behavior:__
When two records have the same timestamp and tags, I expect one to overwrite the other without affecting other measurements.

__Actual behavior:__
One overwrites the other, but then messes up data from other measurements.

__Environment info:__

* System info: Linux 4.19.76-linuxkit x86_64
* InfluxDB version: InfluxDB 2.0.1 (git: 37cc047133) build_date: 2020-11-11T03:53:31Z
* Other relevant environment details: Running using the docker image on Windows 10 using HyperV containers

stuartcarnie:
Closes #20035

This PR ensures all elements of the `cursors.CursorIterator` slice (`itrs`) are read when reading a multi-shard cursor. It is accompanied with a unit test to replicate the incorrect behaviour. 

The root cause is that it is possible a single shard has no data for a given query and this resulted in the read stopping short, as it returned a nil cursor.

Unlike the `createCursor` function, which finds the next non-nil result from the `itrs` slice:

https://github.com/influxdata/influxdb/blob/b7ac9f07be2b668be419bb77293570216be0988e/storage/reads/array_cursor.go#L110-L113

the existing implementation of `nextArrayCursor` was only checking one element of the `itrs` slice:

https://github.com/influxdata/influxdb/blob/632e8f11ed20b28eac688d99a17cdfa510e0ac3c/storage/reads/array_cursor.gen.go#L316-L317

If the call to `Next` returned a nil `Cursor`, that signalled the end of the read.

This fix ensures `itrs` is read in a loop for the next non-nil cursor.

- [x] Rebased/mergeable
- [x] Tests pass


stuartcarnie:
@nathanielc please feel free to add someone from your team for review.

@danxmoran I added you, so you could see what the fix was 👍

jsternberg (approved):

PR #20067, opened 2020-11-17
feat(cli): dbrp cli

### LINKED ISSUE #19950: Add dbrp management to the CLI
We have an area for v1 tooling. let's put CRUD ops for dbrps api there as well: https://docs.influxdata.com/influxdb/v2.0/api/#tag/DBRPs

```
➜  darwin git:(v1-aut-desc) ✗ ./influx v1 -h
InfluxDB v1 management commands

Usage:
  influx v1 [flags]
  influx v1 [command]

Available Commands:
  auth        Authorization management commands for v1 APIs

Flags:
  -h, --help   Help for the v1 command

Use "influx v1 [command] --help" for more information about a command.
```

see also #19333

docmerlin:
Closes #19950 

This adds DBRP management to the CLI.

### DIFF
--- a/cmd/influx/v1_dbrp.go
+++ b/cmd/influx/v1_dbrp.go
@@ -118,7 +118,6 @@ func v1DBRPFindF(cmd *cobra.Command, args []string) error {
 	}
 
 	if v1DBRPFindFlags.Default != nil {
-		fmt.Println("here")
 		filter.Default = v1DBRPFindFlags.Default
 	}
 	dbrps, _, err := s.FindMany(context.Background(), filter, influxdb.FindOptions{})
--- a/cmd/influx/v1_dbrp.go
+++ b/cmd/influx/v1_dbrp.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"fmt"
 	"io"
 
 	"github.com/influxdata/influxdb/v2"
--- a/cmd/influx/v1_dbrp.go
+++ b/cmd/influx/v1_dbrp.go
@@ -1,0 +1,386 @@
+package main
+
+import (
+	"context"
+	"io"
+
+	"github.com/influxdata/influxdb/v2"
+	"github.com/influxdata/influxdb/v2/cmd/influx/internal"
+	"github.com/influxdata/influxdb/v2/dbrp"
+	"github.com/spf13/cobra"
+)
+
+func cmdV1DBRP(f *globalFlags, opt genericCLIOpts) *cobra.Command {
+	cmd := opt.newCmd("dbrp", nil, false)
+	cmd.Aliases = []string{"dbrp"}
+	cmd.Short = "Database retention policy mappings for v1 APIs"
+	cmd.Run = seeHelp
+
+	cmd.AddCommand(
+		v1DBRPCreateCmd(f, opt),
+		v1DBRPFindCmd(f, opt),
+		v1DBRPDeleteCmd(f, opt),
+		v1DBRPUpdateCmd(f, opt),
+	)
+
+	return cmd
+}
+
+var v1DBRPCRUDFlags struct {
+	id          string
+	json        bool
+	hideHeaders bool
+}
+
+var v1DBRPFindFlags struct {
+	BucketID string       // Specifies the bucket ID to filter on
+	DB       string       // Specifies the database to filter on
+	Default  bool         // Specifies filtering on default
+	ID       string       // Specifies the mapping ID to filter on
+	Org      organization // required  // Specifies the organization ID to filter on
+	RP       string       // Specifies the retention policy to filter on
+}
+
+func v1DBRPFindCmd(f *globalFlags, opt genericCLIOpts) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List DBRP",
+		RunE:  checkSetupRunEMiddleware(&flags)(v1DBRPFindF),
+	}
+
+	f.registerFlags(opt.viper, cmd)
+	registerPrintOptions(opt.viper, cmd, &v1DBRPCRUDFlags.hideHeaders, &v1DBRPCRUDFlags.json)
+	v1DBRPFindFlags.Org.register(opt.viper, cmd, false)
+	cmd.Flags().StringVarP(&v1DBRPCreateFlags.BucketID, "bucket-id", "b", "", "the bucket ID to filter on")
+	cmd.Flags().StringVarP(&v1DBRPCreateFlags.DB, "db", "d", "", "the v1 database to map from")
+	cmd.Flags().BoolVar(&v1DBRPCreateFlags.Default, "default", false, "Specify if this mapping represents the default retention policy for the database specificed")
+	cmd.Flags().StringVarP(&v1DBRPCreateFlags.RP, "rp", "r", "", "InfluxDB v1 retention policy")
+
+	return cmd
+}
+
+func v1DBRPFindF(cmd *cobra.Command, args []string) error {
+	if err := v1DBRPFindFlags.Org.validOrgFlags(&flags); err != nil {
+		return err
+	}
+	orgSvc, err := newOrganizationService()
+	if err != nil {
+		return err
+	}
+
+	orgID, err := v1DBRPCreateFlags.Org.getID(orgSvc)
+	if err != nil {
+		return err
+	}
+
+	s, err := newV1DBRPService()
+	if err != nil {
+		return err
+	}
+	filter := influxdb.DBRPMappingFilterV2{OrgID: &orgID}
+
+	if err := filter.OrgID.DecodeFromString(orgID.String()); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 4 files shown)


danxmoran:
Will take a deeper look later, but low-hanging fruit to start: can you add a line to the CHANGELOD.md?

danxmoran:
Failed gotest seems(?) unrelated, I opened #20069 to track it

cmd/influx/v1_dbrp.go
```diff
@@ -0,0 +1,386 @@
...
+		Short: "List DBRP",
+		RunE:  checkSetupRunEMiddleware(&flags)(v1DBRPFindF),
+	}
+
+	f.registerFlags(opt.viper, cmd)
+	registerPrintOptions(opt.viper, cmd, &v1DBRPCRUDFlags.hideHeaders, &v1DBRPCRUDFlags.json)
+	v1DBRPFindFlags.Org.register(opt.viper, cmd, false)
+	cmd.Flags().StringVarP(&v1DBRPCreateFlags.BucketID, "bucket-id", "b", "", "the bucket ID to filter on")
+	cmd.Flags().StringVarP(&v1DBRPCreateFlags.DB, "db", "d", "", "the v1 database to map from")
+	cmd.Flags().BoolVar(&v1DBRPCreateFlags.Default, "default", false, "Specify if this mapping represents the default retention policy for the database specificed")
```
danxmoran, line 56: Does `--default` == the magic "autogen" RP? If so, should we validate up-front that `--rp` isn't passed when `--default` is used?

In the V1 write-path, if no RP is specified in a request then "autogen" is used. Could we mimic that behavior here (instead of having a `--default` flag) for symmetry?

cmd/influx/v1_dbrp.go
```diff
@@ -0,0 +1,386 @@
...
+	cmd.Flags().StringVarP(&v1DBRPCreateFlags.RP, "rp", "r", "", "InfluxDB v1 retention policy")
+
+	return cmd
+}
+
+func v1DBRPFindF(cmd *cobra.Command, args []string) error {
```
danxmoran, line 62: Nitpick, more me figuring out naming conventions: should this func name (and the others) end in `E` instead of `F`? To match the `RunE` it's passed to?

cmd/influx/v1_dbrp.go
```diff
@@ -0,0 +1,386 @@
...
+		return err
+	}
+
+	s, err := newV1DBRPService()
+	if err != nil {
+		return err
+	}
+	filter := influxdb.DBRPMappingFilterV2{OrgID: &orgID}
+
+	if err := filter.OrgID.DecodeFromString(orgID.String()); err != nil {
```
danxmoran, line 82: Could you add a comment explaining why this is needed? Without knowing the code, I'm surprised that the ID returned by `v1DBRPCreateFlags.Org.getId` isn't usable as-is.

cmd/influx/v1_dbrp.go
```diff
@@ -0,0 +1,386 @@
...
+	hideHeaders bool
+}
+
+var v1DBRPFindFlags struct {
+	BucketID string       // Specifies the bucket ID to filter on
+	DB       string       // Specifies the database to filter on
+	Default  bool         // Specifies filtering on default
+	ID       string       // Specifies the mapping ID to filter on
+	Org      organization // required  // Specifies the organization ID to filter on
+	RP       string       // Specifies the retention policy to filter on
```
danxmoran, line 41: AFAICS the fields DB, Default, and RP aren't actually used in the implementation.

cmd/influx/v1_dbrp.go
```diff
@@ -0,0 +1,386 @@
...
+	f.registerFlags(opt.viper, cmd)
+	registerPrintOptions(opt.viper, cmd, &v1DBRPCRUDFlags.hideHeaders, &v1DBRPCRUDFlags.json)
+	v1DBRPCreateFlags.Org.register(opt.viper, cmd, false)
+	cmd.Flags().StringVarP(&v1DBRPCreateFlags.BucketID, "bucket-id", "b", "", "the bucket ID to filter on")
+	_ = cmd.MarkFlagRequired("bucket-id")
+	cmd.Flags().StringVarP(&v1DBRPCreateFlags.DB, "db", "d", "", "the v1 database to map from")
+	_ = cmd.MarkFlagRequired("db")
+	cmd.Flags().BoolVar(&v1DBRPCreateFlags.Default, "default", false, "Specify if this mapping represents the default retention policy for the database specificed")
+	cmd.Flags().StringVarP(&v1DBRPCreateFlags.RP, "rp", "r", "", "InfluxDB v1 retention policy")
+	_ = cmd.MarkFlagRequired("rp")
```
danxmoran, line 138: Similar questions about how `--rp` and `--default` interact

cmd/influx/v1_dbrp.go
```diff
@@ -0,0 +1,386 @@
...
+
+	if v1DBRPFindFlags.BucketID != "" {
+		fID, err := influxdb.IDFromString(v1DBRPFindFlags.ID)
+		if err != nil {
+			return err
+		}
+		filter.BucketID = fID
+	}
+
+	dbrps, _, err := s.FindMany(context.Background(), filter, influxdb.FindOptions{})
```
danxmoran, line 102: Sanity-check: this is hitting an authed endpoint, right?

cmd/influx/v1_dbrp.go
danxmoran: Alt: I see in the delete command that the `Org.getId` result is used as-is, could we delete this statement entirely?

cmd/influx/v1_dbrp.go
```diff
@@ -0,0 +1,386 @@
...
+		hideHeaders: v1DBRPCRUDFlags.hideHeaders,
+		mapping:     dbrpToDelete,
+	})
+}
+
+var v1DBRPUpdateFlags struct {
+	ID      string       // Specifies the mapping ID to filter on
+	Org     organization // required  // Specifies the organization ID to filter on
+	DB      string       //InfluxDB v1 database
+	Default *bool
```
danxmoran, line 294: Intentional pointer?

cmd/influx/v1_dbrp.go
```diff
@@ -0,0 +1,386 @@
...
+	RP      string // InfluxDB v1 retention policy
+}
```
danxmoran, line 296: Should we also have BucketID on this one?

cmd/influx/v1_dbrp.go
```diff
@@ -0,0 +1,386 @@
...
+	return cmd
+}
+
+func v1DBRPUpdateF(cmd *cobra.Command, args []string) error {
+	if err := v1DBRPCreateFlags.Org.validOrgFlags(&flags); err != nil {
+		return err
+	}
+	if defaultFlg := cmd.Flags().Lookup("default"); defaultFlg.Changed {
+		defaultBool, _ := cmd.Flags().GetBool("default")
+		defaultBool = false
```
danxmoran, line 320: Should this be here? Won't `v1DBRPUpdateFlags.Default` always end up `&false`?

danxmoran (commented):
Higher-level questions:
* Would it make sense to unit-test any piece of this?
* Are the errors returned by the server sane enough to return directly from the CLI, or should we wrap them with friendlier messages?

cmd/influx/v1_dbrp.go
docmerlin: As far as I can tell Default is a property of the mapping not the RP.

cmd/influx/v1_dbrp.go
docmerlin: I copied the pattern from the other v1 commands, I don't really care either way though.

cmd/influx/v1_dbrp.go
docmerlin: Yes, so it only updates default if you explicitly set it.

cmd/influx/v1_dbrp.go
docmerlin: The Server ignores everything except RP and Default.  Thanks, though I had forgotten to remove DB from the object (but I removed it from the flags

cmd/influx/v1_dbrp.go
docmerlin: Good eye, I was using this to debug and forgot to remove it.  THANKS.

cmd/influx/v1_dbrp.go
```diff
@@ -0,0 +1,392 @@
...
+var v1DBRPCRUDFlags struct {
+	id          string
+	json        bool
+	hideHeaders bool
+}
+
+var v1DBRPFindFlags struct {
+	BucketID string       // Specifies the bucket ID to filter on
+	DB       string       // Specifies the database to filter on
+	Default  bool         // Specifies filtering on default
```
danxmoran, line 38: I think we're still not using `Default` below

danxmoran (commented):
One more thing

danxmoran (approved):
Will merge when CI passes

Issue #20007, opened 2020-11-12
New release numbers are not updated in UI


mnsmithuk:
After installing 2.0.1, the login screen still says 2.0.0

![image](https://user-images.githubusercontent.com/30718438/98940498-fa91cc80-24e2-11eb-8151-cfd5bd650ac2.png)

PR #20077, opened 2020-11-17
chore(ui): bump version in package.json so it displays correctly

danxmoran:
Relates to #20007 

The UI displays the version that was set in `package.json` at build time. We'll need to fix this "for real" as part of vendoring out the UI, but this will at least make the value correct for 2.0.2.

- [ ] Rebased/mergeable
- [ ] Tests pass


psteinbachs (approved):
lgtm

PR #20083, opened 2020-11-17
fix(api): raise API pagesizes

docmerlin:
Currently, in the API, pagination is not working; to mitigate this, we
can drastically increase the page size.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,10 @@
 
 ### Features
 
-1. [20036](https://github.com/influxdata/influxdb/pull/20036): Warn if V1 users are upgraded, but V1 auth wasn't enabled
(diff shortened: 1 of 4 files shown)


danxmoran (commented):
I don't feel qualified to :+1: or :-1: this, so I'll just go for the low-hanging fruit: can you update the CHANGELOG if we decide to go through with it? 😄

danxmoran (approved):
Looks like Phillip is 👍 so LGTM

psteinbachs (approved):
As discussed, consensus is this is a reasonable stopgap for now.

stuartcarnie (approved):
Thanks for fixing the tests – they'll be much more reliable using the constants too 💯

PR #20091, opened 2020-11-18
fix(api): make the dbrp api match the swagger spec

### LINKED ISSUE #20087: DBRP server implementation does not match swagger specification
See: https://docs.influxdata.com/influxdb/v2.0/api/#operation/PostDBRP

https://github.com/influxdata/influxdb/blob/d93560fdc85f21d8a4720a7882698f38c5be28b2/dbrp/http_server_dbrp.go#L58-L65

The API docs states:
- `org` 
- `orgID`
- `bucketID`

But we shorten organization to org, and we use camel-case in the docs / API spec.
Since the docs are the shared source of truth, we should update the implementation.

docmerlin:
This is to make the DBRP api match the swagger spec.
Closes #20087


danxmoran (approved):

danxmoran:
Oh wait, need to check client code...

danxmoran:
Hit a panic during manual testing, investigating...

danxmoran (approved):
Panic seems unrelated, LGTM

PR #20097, opened 2020-11-19
fix: Ensure Index.Walk fetches matching foreign keys only

### LINKED ISSUE #20096: kv Index.Walk function matches the foreign key by prefix only
## What

When the kv store searches the index, it utilises a forward cursor matching by prefix:

https://github.com/influxdata/influxdb/blob/15b95312736f1ea651d2e14ba8343e483d5cd92a/kv/index.go#L192-L193

The cursor will return all keys that have a matching prefix, in the following loop, including those keys which are longer:

https://github.com/influxdata/influxdb/blob/15b95312736f1ea651d2e14ba8343e483d5cd92a/kv/index.go#L206-L208

For example, if the `foreignKey` is `hello` and the index contains the keys `hello`, `hello-world` and `hello-there`, the loop will gather them all.

In most cases, the index is typically the result of combining 16-byte IDs, so this is not a problem. However, for `dbrp`, it creates an additional `orgID` + `database` foreign key:

https://github.com/influxdata/influxdb/blob/b274e15eacb2cf6c8915a9f75caf63d69fb2aac1/dbrp/service.go#L70-L76

### Creation of DBRPs fail:

1. Create `bucket1`
2. Create a dbrp mapping for `bucket1` with db `jenkins-aws`, rp `never`:

    ```sh
    curl -H 'Content-type: application/json' --header "Authorization: Token $INFLUX_TOKEN" -XPOST "$INFLUX_HOST/api/v2/dbrps" --data-binary '{"bucket_id": "<BUCKET1_ID>","database": "jenkins-aws","default": true, "organization_id": "<org-id>", "retention_policy": "never"}'
    ```

3. Create `bucket2`
4. Attempt to create a dbrp mapping for `bucket2` with db `jenkins`, rp `never`:

    ```sh
    curl -H 'Content-type: application/json' --header "Authorization: Token $INFLUX_TOKEN" -XPOST "$INFLUX_HOST/api/v2/dbrps" --data-binary '{"bucket_id": "<BUCKET2_ID>","database": "jenkins","default": true, "organization_id": "<org-id>", "retention_policy": "never"}'
    ```

    Fails with the following error:

    ```
    another DBRP mapping with same orgID, db, and rp exists
    ```

stuartcarnie:
Closes #20096

This PR ensures `Influx.Walk` validates that every matched key returned by the index prefix cursor is an exact match with the `foreignKey` argument. This change broke an invalid assumption made by the DBRP Find operations when searching for mappings by organisation ID only. This was fixed in a second commit.

The first commit modifies the behaviour of the internal `indexWalk` function to ensure it parses the key parts and matches the foreign key exactly. Unit tests ensure this behaviour.

The second commit adds a new index and migration to the DBRP service for finding all database and retention policy mappings for a single organisation. This change was required to resolve an invalid assumption of the DBRP service, which relied on a prefix match of the `byOrgAndDatabase` index when performing search operations using the organisation ID only.

- [x] Rebased/mergeable
- [x] Tests pass


stuartcarnie:
@GeorgeMac, given your experience with implementing the index, can you review this draft PR and comment about the expected behaviour of `Influx.Walk` and whether we should introduce `Index.WalkExact` for the `dbrp` `orgIDDatabase` index:

https://github.com/influxdata/influxdb/blob/b274e15eacb2cf6c8915a9f75caf63d69fb2aac1/dbrp/service.go#L70-L76

kv/index.go
```diff
@@ -195,16 +195,18 @@ func (i *Index) Walk(ctx context.Context, tx Tx, foreignKey []byte, visitFn Visi
...
 
 // indexWalk consumes the indexKey and primaryKey pairs in the index bucket and looks up their
 // associated primaryKey's value in the provided source bucket.
 // When an item is located in the source, the provided visit function is called with primary key and associated value.
-func indexWalk(ctx context.Context, indexCursor ForwardCursor, sourceBucket Bucket, visit VisitFunc) (err error) {
+func indexWalk(ctx context.Context, foreignKey []byte, indexCursor ForwardCursor, sourceBucket Bucket, visit VisitFunc) (err error) {
 	var keys [][]byte
 	for ik, pk := indexCursor.Next(); ik != nil; ik, pk = indexCursor.Next() {
-		keys = append(keys, pk)
+		if string(foreignKey) == string(ik) {
```
GeorgeMac, line 207:
```suggestion
		if bytes.Equal(foreignKey, bytes.SplitN(ik, []byte{'/'}, 2)[0]) {
```
I get it now :D This might be what you're looking for. The foreign key is all the bits before the `/` in the index key.

GeorgeMac: or even:
```go
if ok := bytes.HasPrefix(ik, foreignKey); ok && len(ik) > len(foreignKey) && ik[len(foreignKey)] == '/' {

}
```

stuartcarnie: Thanks for the info – I see now that the `Index` composes the key as `foreignKey` + `/` + `primaryKey`

stuartcarnie: I have resolved the issue with the latest changes by using the `indexKeyParts` function to parse the foreign key value. Thanks again for leading me in the right direction 👏

kv/index_test.go
```diff
@@ -33,6 +38,102 @@ func Test_Bolt_Index(t *testing.T) {
 	influxdbtesting.TestIndex(t, s)
 }
 
+func TestIndex_Walk(t *testing.T) {
+	t.Run("only selects exact keys", func(t *testing.T) {
+		ctrl := gomock.NewController(t)
+		t.Cleanup(ctrl.Finish)
```
GeorgeMac, line 44: take it or leave it: With 1.14+ this happens by default:
https://godoc.org/github.com/golang/mock/gomock#NewController
```suggestion
```

kv/index.go
```diff
@@ -195,16 +197,20 @@ func (i *Index) Walk(ctx context.Context, tx Tx, foreignKey []byte, visitFn Visi
...
 
 // indexWalk consumes the indexKey and primaryKey pairs in the index bucket and looks up their
 // associated primaryKey's value in the provided source bucket.
 // When an item is located in the source, the provided visit function is called with primary key and associated value.
-func indexWalk(ctx context.Context, indexCursor ForwardCursor, sourceBucket Bucket, visit VisitFunc) (err error) {
+func indexWalk(ctx context.Context, foreignKey []byte, indexCursor ForwardCursor, sourceBucket Bucket, visit VisitFunc) (err error) {
 	var keys [][]byte
 	for ik, pk := indexCursor.Next(); ik != nil; ik, pk = indexCursor.Next() {
-		keys = append(keys, pk)
+		if fk, _, err := indexKeyParts(ik); err != nil {
```
GeorgeMac, line 209: 🙌 awesome

GeorgeMac (approved):
Looks great @stuartcarnie !

kv/index.go
GeorgeMac: Actually @stuartcarnie you know what I've just thought of something better.
If we curse `foreignKey + "/"` then we push down the actual prefix to the db.
This does make me think the foreign key needs to be escaped for `"/"` as well though.

kv/index_test.go
stuartcarnie: I have been waiting for this, however, the docs are misleading as there is no released version of `gomock` with this support yet:

mock@master has the `Cleanup` support:

https://github.com/golang/mock/blob/99aa9272d551d0d16f4bdcc43dc06a3077e69b3a/gomock/controller.go#L136-L141

latest tagged version is mock@1.4.4, which does not have `Cleanup` support:

https://github.com/golang/mock/blob/f7b1909c82a8958747e5c87c6a5c3b2eaed8a33d/gomock/controller.go#L118-L128

kv/index.go
stuartcarnie: It is unclear what I should change here – should we add some code to validate that foreign keys do not have `/` for now?

stuartcarnie: Spoke with @GeorgeMac and I am going to reject foreign keys with a `/`. Clients of the `Index` will be expected to partition their foreign keys with characters other than a `/`

Issue #20098, opened 2020-11-19
Successful writes increment write error statistics incorrectly, fail to increment write success statistics.


davidby-influx:
In v1.8.3 and earlier, the write path through (*PointsWriter) writeToShardWithContext() always increments the WriteErr count in the debug variables, and does not increment the WriteOK count.

https://github.com/influxdata/influxdb/blob/v1.8.3/coordinator/points_writer.go line 450 should be an `else if  err != nil {` instead of an `else`

This has been reported in a customer cloud instance, and verified under a debugger.

Fix should go in master-1.x and 1.8 branches.

davidby-influx:
PR approved and merged.

PR #20100, opened 2020-11-19
fix(write): Successful writes increment write error statistics incorrectly

davidby-influx:
In v1.8.3 and earlier, the write path through (*PointsWriter) writeToShardWithContext() always increments the WriteErr count in the debug variables, and does not increment the WriteOK count.
https://github.com/influxdata/influxdb/blob/v1.8.3/coordinator/points_writer.go line 450 should be an `else if err != nil {` instead of an `else`
This has been reported in a customer cloud instance, and verified under a debugger.

Closes: https://github.com/influxdata/influxdb/issues/20098

- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,6 @@ v1.8.0 [unreleased]
 
 ### Bugfixes
 
--   [#20100](https://github.com/influxdata/influxdb/pull/20100): fix(write): Successful writes increment write error statistics incorrectly
 -	[#10503](https://github.com/influxdata/influxdb/pull/10503): Delete rebuilds series index when series to be deleted are only found in cache.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ v1.8.0 [unreleased]
 
 ### Bugfixes
 
+-   [#20100](https://github.com/influxdata/influxdb/pull/20100): fix(write): Successful writes increment write error statistics incorrectly
```
dgnorton, line 26: The changelog on `master-1.x` isn't currently being maintained. We usually update it on the release branch.

davidby-influx: So, should I remove that entry?  I updated both changelogs.

danxmoran (approved):

Issue #20099, opened 2020-11-19
Successful writes increment write error statistics incorrectly, fail to increment write success statistics. [BACKPORT 1.8]


davidby-influx:
Backport 

https://github.com/influxdata/influxdb/issues/20098

to 1.8

davidby-influx:
PR approved and merged

PR #20101, opened 2020-11-19
fix(write): Successful writes increment write error statistics incorrectly

davidby-influx:
In v1.8.3 and earlier, the write path through (*PointsWriter) writeToShardWithContext() always increments the WriteErr count in the debug variables, and does not increment the WriteOK count.
https://github.com/influxdata/influxdb/blob/v1.8.3/coordinator/points_writer.go line 450 should be an `else if err != nil { `instead of an `else`
This has been reported in a customer cloud instance, and verified under a debugger.

https://github.com/influxdata/influxdb/issues/20098
(cherry picked from commit 3d9f8b5020daf5e3759e6bf1570324d77a4cbe44)

Closes: https://github.com/influxdata/influxdb/issues/20099

- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ v1.8.4 [unreleased]
 
 ### Bugfixes
 
+-   [#20101](https://github.com/influxdata/influxdb/pull/20101): fix(write): Successful writes increment write error statistics incorrectly
 -	[#19696](https://github.com/influxdata/influxdb/pull/19697): fix(flux): add durations to Flux logging
 
 v1.8.3 [2020-09-30]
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ v1.8.3 [2020-09-30]
 
 ### Bugfixes
 
+-   [#20101](https://github.com/influxdata/influxdb/pull/20101): fix(write): Successful writes increment write error statistics incorrectly
```
dgnorton, line 18: This should move up to the `v1.8.4 [unreleased]` section

davidby-influx: Will do.

davidby-influx: Fixed

danxmoran (approved):
Best PR ever

PR #20110, opened 2020-11-19
feat(cmd/influxd): use a better default for upgraded V2 config, and allow users to override it

### LINKED ISSUE #20104: `influxd upgrade` writes `config.toml` to a confusing location
When `influxd upgrade` converts a V1 config file to V2, it writes the output `config.toml` into the same directory as the V1 file. This results in confusing behavior like `config.toml` being written into `~/.influxdb` instead of `~/.influxdbv2`. We don't expose an option to override this behavior.

We should fix the command to:
1. Expose a new CLI option to set the output path of the V2 config
2. Update the default path to be `~/.influxdbv2/config.toml`

danxmoran:
Closes #20104 

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influxd/upgrade/config.go
+++ b/cmd/influxd/upgrade/config.go
@@ -4,7 +4,6 @@ package upgrade
 // The strategy is to transform only those entries for which rule exists.
 
 import (
-	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -109,7 +108,7 @@ func load(path string) ([]byte, error) {
 }
 
 // upgradeConfig upgrades existing 1.x configuration file to 2.x influxdb.toml file.
-func upgradeConfig(v1Config map[string]interface{}, targetOptions optionsV2, log *zap.Logger) (string, error) {
+func upgradeConfig(v1Config map[string]interface{}, targetOptions optionsV2, log *zap.Logger) error {
 	// create and initialize helper
 	cu := &configUpgrader{
 		rules:           configMapRules,
@@ -117,17 +116,14 @@ func upgradeConfig(v1Config map[string]interface{}, targetOptions optionsV2, log
 		log:             log,
 	}
 
+	// rewrite config options from V1 to V2 paths
 	cTransformed := cu.transform(v1Config)
 
 	// update new config with upgrade command options
 	cu.updateV2Config(cTransformed, targetOptions)
 
-	configFileV2, err := cu.save(cTransformed, targetOptions.configPath)
-	if err != nil {
-		return "", err
-	}
-
-	return configFileV2, nil
+	// write the ugpraded config to disk
+	return cu.save(cTransformed, targetOptions.configPath)
 }
 
 // configUpgrader is a helper used by `upgrade-config` command.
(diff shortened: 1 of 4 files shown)


cmd/influxd/upgrade/upgrade.go
```diff
@@ -238,6 +238,12 @@ func NewCommand(v *viper.Viper) *cobra.Command {
 			Flag:  "config-file",
 			Desc:  "optional: Custom InfluxDB 1.x config file path, else the default config file",
 		},
+		{
+			DestP:   &options.target.configPath,
+			Flag:    "v2-config-path",
```
danxmoran, line 243: This breaks the convention for existing V2 output paths (i.e. `--bolt-path` and `--engine-path`), but we already have `--config-file` for pointing at V1 config so I didn't want to use `--config-path` to mark the V2 output.

cmd/influxd/upgrade/config.go
```diff
@@ -152,7 +152,13 @@ func (cu *configUpgrader) save(config map[string]interface{}, path string) (stri
 		return "", err
 	}
 
-	err := ioutil.WriteFile(path, buf.Bytes(), 0666)
+	// Try writing config to the requested location.
```
ayang64, line 155: i wouldn't test if `err == nil`.

i'd test for `err != nil` and handle it here.

cmd/influxd/upgrade/config.go
```diff
@@ -152,7 +152,13 @@ func (cu *configUpgrader) save(config map[string]interface{}, path string) (stri
...
+	var err error
+	err = os.MkdirAll(filepath.Dir(path), 0755)
```
ayang64, line 157: possibly simply `err` value handling by limiting its scope.

instead of assigning it before the condition, i might declare it in the `if` block like:


```
if err := os.MkdirAll(filepath.Dir(path), 0755); err != nil {
  // handle err here.
}
```

ayang64 (changes requested):

cmd/influxd/upgrade/config.go
docmerlin, line 157: Minor nit, 
```go
var err error
err = os.MkdirAll(filepath.Dir(path), 0755)
```
 to `err := os.MkdirAll(filepath.Dir(path), 0755)`

docmerlin (approved):
Approved, with minor nit, which you can ignore.  Can you put periods at the end of the sentences in your CHANGELOG changes?

cmd/influxd/upgrade/config.go
danxmoran: Agreed this is non-standard. The `err != nil` case for both the `MkdirAll` and `WriteFile` cases is to fall through instead of returning `err`. I'll check if we can remove that behavior, it's confused me more times than it's helped me.

danxmoran:
@ayang64 Tim gave a 👍 to removing the fallback case, can you take another look?

cmd/influxd/upgrade/config.go
```diff
@@ -146,20 +142,19 @@ func (cu *configUpgrader) updateV2Config(config map[string]interface{}, targetOp
...
+	if err := os.MkdirAll(filepath.Dir(path), 0755); err != nil {
+		return err
 	}
-
-	err := ioutil.WriteFile(path, buf.Bytes(), 0666)
-	if err != nil { // permission issue possible - try to save the file to home or current dir
-		cu.log.Warn(fmt.Sprintf("Could not save upgraded config to %s, trying to save it to user's home.", path), zap.Error(err))
-		path = filepath.Join(homeOrAnyDir(), filepath.Base(path))
-		err = ioutil.WriteFile(path, buf.Bytes(), 0666)
+	outFile, err := os.OpenFile(path, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
ayang64, line 150: is it okay to append to the file?

danxmoran: The danger of copy-pasting code... the `upgrade` program asserts that there's no file at this target path before the logic reaches this point, but better to be consistent across the board. I'll update the flags to remove it.

ayang64 (approved):
LGTM!

PR #20112, opened 2020-11-20
chore: remove nightly and hourly e2e jobs

gshif:
This PR removes nightly and e2e hourly jobs from circleci,


davidby-influx (approved):
I'm about as far from a Docker expert as exists, but this does look like a straightforward removal of some tasks...

PR #20115, opened 2020-11-20
fix: restore portable backup error

rickif:
This PR is to solve restore problem in influxdb 1.x

`
influxd restore -portable -host [host] -db [db] -newdb [newdb] [backup_path]
`

Restoring portable backup with this command may get error as below
```
2020/11/20 13:55:02 error updating meta: updating metadata on influxd service failed: err=read tcp x.x.x.x:52286->x.x.x.x:8088: read: connection reset by peer, n=16
restore: updating metadata on fctsdb service failed: err=read tcp x.x.x.x:52286->x.x.x.x:8088: read: connection reset by peer, n=16
```
Accordingly, the influxdb error log is as below
```
ts=2020-11-20T13:55:02.555262Z lvl=info msg="failed to decode meta: proto: meta.ShardInfo: illegal tag 0 (wire type 0)" log_id=0QQzP3qG000 service=snapshot
```
This problem happens at random, but more frequently in network with high latency. The `net.func (*TCPConn) Read` dose not promise that exactly len(buf) bytes is read when the call is returned, which leading this problem. Replacing the `net.func (*TCPConn) Read` with `io.ReadFull` could fix this.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ v1.8.5 [unreleased]
 -	[#20909](https://github.com/influxdata/influxdb/pull/20909): fix(tsm1): data race when accessing tombstone stats
 -	[#20912](https://github.com/influxdata/influxdb/pull/20912): fix(tsdb): minimize lock contention when adding new fields or measure
 -	[#20914](https://github.com/influxdata/influxdb/pull/20914): fix: infinite recursion bug (#20862)
+-	[#20115](https://github.com/influxdata/influxdb/pull/20115): fix: restore portable backup bug
 
 v1.8.4 [2021-01-27]
 -------------------
@@ -29,7 +30,6 @@ v1.8.4 [2021-01-27]
 ### Bugfixes
 
 -	[#19696](https://github.com/influxdata/influxdb/pull/19697): fix(flux): add durations to Flux logging
--	[#20115](https://github.com/influxdata/influxdb/pull/20115): fix: restore portable backup error
 
 v1.8.3 [2020-09-30]
 -------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ v1.8.4 [2021-01-27]
 ### Bugfixes
 
 -	[#19696](https://github.com/influxdata/influxdb/pull/19697): fix(flux): add durations to Flux logging
+-	[#20115](https://github.com/influxdata/influxdb/pull/20115): fix: restore portable backup error
 
 v1.8.3 [2020-09-30]
 -------------------
--- a/services/snapshotter/service_test.go
+++ b/services/snapshotter/service_test.go
@@ -165,7 +165,7 @@ func TestSnapshotter_RequestMetastoreBackup(t *testing.T) {
 	}
 	defer l.Close()
 
-	s.MetaClient = &MetaClient{data: data}
+	s.MetaClient = &MetaClient{data: data.Clone()}
 	if err := s.Open(); err != nil {
 		t.Fatalf("unexpected open error: %s", err)
 	}
@@ -214,7 +214,7 @@ func TestSnapshotter_RequestDatabaseInfo(t *testing.T) {
 		return "", fmt.Errorf("no such shard id: %d", id)
 	}
 
-	s.MetaClient = &MetaClient{data: data}
+	s.MetaClient = &MetaClient{data: data.Clone()}
 	s.TSDBStore = &tsdbStore
 	if err := s.Open(); err != nil {
 		t.Fatalf("unexpected open error: %s", err)
@@ -269,7 +269,7 @@ func TestSnapshotter_RequestDatabaseInfo_ErrDatabaseNotFound(t *testing.T) {
 	}
 	defer l.Close()
 
-	s.MetaClient = &MetaClient{data: data}
+	s.MetaClient = &MetaClient{data: data.Clone()}
 	if err := s.Open(); err != nil {
 		t.Fatalf("unexpected open error: %s", err)
 	}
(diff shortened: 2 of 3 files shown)


danxmoran:
Hi @rickif thanks for the contribution! While we review, could you push an update to the `CHANGELOG.md` file with a reference to this PR?

rickif:
@danxmoran OK, the `CHANGELOG.md` is updated now.

danxmoran (commented):
@rickif this change makes sense to me in theory from reading the docs on `Read` and `ReadFull`, but I'm not sure what to expect in practice from the use-case you mention (restoring on a high-latency network). In situations where `Read` used to return "early", will `ReadFull` now sit and wait for more data? Or will it still result in an error, but the error will be easier to understand?

Do you feel up to try adding tests to [`services/snapshotter/service.go`](https://github.com/influxdata/influxdb/blob/1.8/services/snapshotter/service_test.go)? The 2 cases I'd ideally want covered are:
1. `conn` closes before `ReadFull` reads enough bytes to fill `typ[:]`
2. `conn` doesn't "contain" enough bytes to fill `typ[:]` when `ReadFull` is called, but more data arrives afterwards

rickif:
@danxmoran  I agree with u and have added some test codes .But I have doubt about the cases you mentioned.
1. Is it necessary to cover the first case? According to the godoc, `If an EOF happens after reading some but not all the bytes, ReadFull returns ErrUnexpectedEOF`. The result is clear and well-defined by golang.
2. On Linux platform,  the call to `read()` may read less than the requested number of bytes, which may occur if there are fewer bytes available in the socket than were requested in the `read()` call. The possible cause of this is happened in the Linux Kernel Mode, which is not easy to add test to simulate this circumstance. I tried to do this by dividing the payload and sending the pieces one by one, meanwhile `Sleep()` is added after every sending. But it does not work at all. I wonder whether you have any good idea about this?

danxmoran:
> Is it necessary to cover the first case? According to the godoc, If an EOF happens after reading some but not all the bytes, ReadFull returns ErrUnexpectedEOF. The result is clear and well-defined by golang.

A test case covering only a call to `ReadFull` would indeed be redundant. The pieces I'm interested in covering are the functions wrapping the `ReadFull` calls. My impression is that before this change, those functions would either complete successfully (but return partial/corrupted data) or fail with a confusing error. The functions still fail with this change, but with (hopefully) a more clear explanation in the error. If it's easy to capture the new expectation in a test, it'd help guard against future regressions.

That said, I don't want to gatekeep your contribution for too long. If setting up test cases is going to take a significant amount of mocking & other test machinery, I'm ok merging without additional tests. In that case, could you add a comment above each of the `ReadFull` calls explaining why `ReadFull` over `Read`?

danxmoran:
> But it does not work at all. I wonder whether you have any good idea about this?

It sounds like this will be more trouble than it's worth, I wouldn't spend more time on it. Thanks for investigating!

rickif:
> The functions still fail with this change, but with (hopefully) a more clear explanation in the error

This change is not aiming at providing a more clear explanation, but is mainly to solve the problem that the `conn.Read()` may return too early without enough data, which leds a data corruption error even if the data is valid. As you mentioned before, the `io.ReadFull` would block until expected data is read, in which the data could be parsed correctly.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ v1.8.4 [unreleased]
 ### Bugfixes
 
 -	[#19696](https://github.com/influxdata/influxdb/pull/19697): fix(flux): add durations to Flux logging
+-	[#20115](https://github.com/influxdata/influxdb/pull/20115): fix: restore portable backup error
```
danxmoran, line 7:
```suggestion
-	[#20115](https://github.com/influxdata/influxdb/pull/20115): fix: prevent invalid payload errors while restoring backups. Thanks @rickif!
```

services/snapshotter/service_test.go
```diff
@@ -379,6 +381,42 @@ func TestSnapshotter_RequestRetentionPolicyInfo(t *testing.T) {
...
+		Type:                   snapshotter.RequestMetaStoreUpdate,
+		BackupDatabase:         "db0",
+		RestoreDatabase:        "db1",
+		BackupRetentionPolicy:  "autogen",
+		RestoreRetentionPolicy: "autogen",
+		UploadSize:             int64(len(metaBytes)),
+	}
+
+	c := snapshotter.NewClient(l.Addr().String())
+	if _, err := c.UpdateMeta(req, bytes.NewReader(metaBytes)); err != nil {
```
danxmoran, line 415: Is there anything useful we could assert about the success case here?

services/snapshotter/service_test.go
```diff
@@ -163,7 +165,7 @@ func TestSnapshotter_RequestMetastoreBackup(t *testing.T) {
 	}
 	defer l.Close()
 
-	s.MetaClient = &MetaClient{Data: data}
+	s.MetaClient = &MetaClient{data: data.Clone()}
```
danxmoran, line 168: Are these new `Clone()` calls needed for the tests to pass?

danxmoran (commented):
Thanks @rickif, one suggestion to curate the CHANGELOG, and some questions in the new test code.

services/snapshotter/service_test.go
rickif: Yeah! The `Clone()` calls are unnecessary, which is removed now.

services/snapshotter/service_test.go
rickif: Of course, I've add some codes to cover the success case

rickif:
> Thanks @rickif, one suggestion to curate the CHANGELOG, and some questions in the new test code.

Thanks for suggestions. The PR has been updated!

danxmoran (dismissed):
LGTM!

rickif:
@danxmoran Hi, the PR has been pending for a long time. Is there any problem?

rickif:
@dgnorton Hi, would you please do review for this PR?

danxmoran:
@rickif very sorry for the delay, I'm not sure how I lost track of this one. I'll do a final review pass now.

rickif:
> @rickif very sorry for the delay, I'm not sure how I lost track of this one. I'll do a final review pass now.

that would very nice.

lesam:
Re-opened this PR as https://github.com/influxdata/influxdb/pull/21991 so that I can update the changelog, otherwise it is good.

Issue #20121, opened 2020-11-20
[2.0.2] a panic has occurred - on 1.x query/write endpoints


nsteinmetz:
__Steps to reproduce:__
List the minimal actions needed to reproduce the behavior.

1. Upgrade data from 1.8.3 to 2.0.1 with success
2. Upgrade deb package to 2.0.2
3. See huge amounts of log with panic message

__Expected behavior:__

InfluxDB 1.x endpoints should continue to work as expected

__Actual behavior:__

```
[...]

PR #20122, opened 2020-11-20
fix: panic when there is a bad DBRP created with no default

docmerlin:
Related to #20121

### DIFF
--- a/dbrp/service.go
+++ b/dbrp/service.go
@@ -265,7 +265,9 @@ func (s *Service) FindMany(ctx context.Context, filter influxdb.DBRPMappingFilte
 			if err != nil {
 				return false, ErrInternalService(err)
 			}
-
+			if defID == nil {
+				return true, nil
+			}
 			m.Default = m.ID == *defID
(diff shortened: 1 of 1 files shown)


danxmoran:
@docmerlin is this still something we should review / merge?

dbrp/service.go
```diff
@@ -265,7 +265,9 @@ func (s *Service) FindMany(ctx context.Context, filter influxdb.DBRPMappingFilte
 			if err != nil {
 				return false, ErrInternalService(err)
 			}
-
+			if defID == nil {
+				return true, nil
```
danxmoran, line 269: I played around with this locally, and while it prevents the panic I think it results in too many results being returned because `filterFunc` is skipped. The biggest potential issue I see is skipping the filter on `OrgID`.

danxmoran (changes requested):
Not sure if this is still meant to be merged, but leaving a note to remind us next time we consider the PR.

danxmoran:
Closing as superseded by #20168

PR #20149, opened 2020-11-23
fix: enforce bounds on query-related CLI/config parameters to avoid startup panics

### LINKED ISSUE #20144: Panic when setting --query-concurrency to large value
__Steps to reproduce:__

1. Run `influxd --query-concurrency 9223372036854775807`

__Expected behavior:__
Either the program runs successfully, or the CLI errors out telling me the number is too big.

__Actual behavior:__
```
panic: sync: negative WaitGroup counter

goroutine 1 [running]:
sync.(*WaitGroup).Add(0xc0018362d0, 0x7fffffffffffffff)
        /Users/dan/.gvm/gos/go1.15/src/sync/waitgroup.go:74 +0x147
github.com/influxdata/influxdb/v2/query/control.New(0x7fffffffffffffff, 0x0, 0x7fffffffffffffff, 0x0, 0xa, 0xc00169a480, 0x0, 0x0, 0x0, 0xc001562f90, ...)
        /Users/dan/influxdb/query/control/controller.go:189 +0x7f5
github.com/influxdata/influxdb/v2/cmd/influxd/launcher.(*Launcher).run(0xc0005b0e00, 0x760eb80, 0xc00055d480, 0x0, 0x0)
        /Users/dan/influxdb/cmd/influxd/launcher/launcher.go:901 +0x3a38
github.com/influxdata/influxdb/v2/cmd/influxd/launcher.cmdRunE.func1(0x0, 0x0)
        /Users/dan/influxdb/cmd/influxd/launcher/launcher.go:149 +0x14c
github.com/influxdata/influxdb/v2/kit/cli.NewCommand.func1(0xc000130b00, 0xc000ce90e0, 0x0, 0x2, 0x0, 0x0)
        /Users/dan/influxdb/kit/cli/viper.go:62 +0x29
github.com/spf13/cobra.(*Command).execute(0xc000130b00, 0xc000184190, 0x2, 0x2, 0xc000130b00, 0xc000184190)
        /Users/dan/.gvm/pkgsets/go1.15/global/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:842 +0x47c
github.com/spf13/cobra.(*Command).ExecuteC(0xc000130b00, 0xc000d9ff48, 0x1, 0x1)
        /Users/dan/.gvm/pkgsets/go1.15/global/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
        /Users/dan/.gvm/pkgsets/go1.15/global/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887
main.main()
        /Users/dan/influxdb/cmd/influxd/main.go:41 +0x31c
```
This is apparently a known problem in Go's `WaitGroup` impl: https://github.com/golang/go/issues/20687

__Environment info:__
InfluxDB OSS 2.0.2

danxmoran:
Closes #20144 

`math.MaxInt32` is the biggest value that the `WaitGroup` wrapped in the query controller will accept without panicking. When I try running `influxd` with that parameter it takes a loooooong time to boot up, but I assume it'd be faster on a beefed-up VM.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ want to use the default.
 1. [20110](https://github.com/influxdata/influxdb/pull/20110): Use V2 directory for default V2 config path in `influxd upgrade`.
 1. [20137](https://github.com/influxdata/influxdb/pull/20137): Fix panic when writing a point with 100 tags. Thanks @foobar!
 1. [20151](https://github.com/influxdata/influxdb/pull/20151): Don't log bodies of V1 write requests.
+1. [20097](https://github.com/influxdata/influxdb/pull/20097): Ensure Index.Walk fetches matching foreign keys only.
+1. [20149](https://github.com/influxdata/influxdb/pull/20149): Enforce max value of 2147483647 on query concurrency to avoid startup panic.
 
 ## v2.0.2 [2020-11-19]
 
@@ -34,7 +36,6 @@ want to use the default.
 1. [20053](https://github.com/influxdata/influxdb/pull/20053): Upgrade Flux to v0.95.0.
 1. [20058](https://github.com/influxdata/influxdb/pull/20058): UI: Upgrade flux-lsp-browser to v0.5.23.
 1. [20067](https://github.com/influxdata/influxdb/pull/20067): Add DBRP cli commands as `influxd v1 dbrp`.
-1. [20097](https://github.com/influxdata/influxdb/pull/20097): Ensure Index.Walk fetches matching foreign keys only.
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ want to use the default.
 1. [20151](https://github.com/influxdata/influxdb/pull/20151): Don't log bodies of V1 write requests.
 1. [20097](https://github.com/influxdata/influxdb/pull/20097): Ensure Index.Walk fetches matching foreign keys only.
 1. [20149](https://github.com/influxdata/influxdb/pull/20149): Enforce max value of 2147483647 on query concurrency to avoid startup panic.
+1. [20149](https://github.com/influxdata/influxdb/pull/20149): Enforce max value of 2147483647 on query queue size to avoid startup panic.
 
 ## v2.0.2 [2020-11-19]
 
--- a/query/control/controller.go
+++ b/query/control/controller.go
@@ -95,10 +95,20 @@ type Config struct {
 	// This number may be less than the ConcurrencyQuota * MemoryBytesQuotaPerQuery.
 	MaxMemoryBytes int64
 
-	// QueueSize is the number of queries that are allowed to be awaiting execution before new queries are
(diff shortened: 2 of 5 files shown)


query/control/controller.go
```diff
@@ -121,6 +128,9 @@ func (c *Config) validate(isComplete bool) error {
 	if c.ConcurrencyQuota <= 0 {
 		return errors.New("ConcurrencyQuota must be positive")
 	}
+	if c.ConcurrencyQuota > MaxQueryConcurrency {
```
davidby-influx, line 131: Should we define ConcurrencyQuota as an int64 in the struct? I realize that `int` is 64 bits on our target architectures, but the additional clarity couldn't hurt.

query/control/controller_test.go
```diff
@@ -122,6 +122,15 @@ func validateUnusedMemory(t testing.TB, reg *prometheus.Registry, c control.Conf
 	}
 }
 
+func TestConfig_ValidateMaxConcurrencyQuota(t *testing.T) {
+	config.ConcurrencyQuota = control.MaxQueryConcurrency + 1
```
davidby-influx, line 127: My Golang ignorance - does this correctly cast MazQueryConcurrency to an int64 before the addition, and void a wrap-around?

davidby-influx (approved):
You can proceed without my issues if you think they are not useful.  Won't block the commit for these minor nits.

query/control/controller.go
danxmoran: `int32` is what we'd want/need to use, but I'm a fan of the general idea... it might get us some extra validation "for free" too, if the CLI parser is smart about detecting overflows. I'll give it a try in the morning.

query/control/controller_test.go
danxmoran: Once I changed the type of `ConcurrencyQuota` to `int32`:
1. This still compiled fine
2. Running the test resulted in immediate failure, with an overflow error

query/control/controller.go
danxmoran: Using `int32` does give us some validation at the CLI layer. I'll keep heading down this path.

CHANGELOG.md
```diff
@@ -34,7 +36,6 @@ want to use the default.
 1. [20053](https://github.com/influxdata/influxdb/pull/20053): Upgrade Flux to v0.95.0.
 1. [20058](https://github.com/influxdata/influxdb/pull/20058): UI: Upgrade flux-lsp-browser to v0.5.23.
 1. [20067](https://github.com/influxdata/influxdb/pull/20067): Add DBRP cli commands as `influxd v1 dbrp`.
-1. [20097](https://github.com/influxdata/influxdb/pull/20097): Ensure Index.Walk fetches matching foreign keys only.
```
danxmoran, line 37: Accidentally filed under the wrong release

davidby-influx (approved):
Looks much safer. Comments are outstanding!

PR #20155, opened 2020-11-23
fix: various improvements to setup command

adrian-thurston:
When checking for existing configs, check the length of the array after
parsing. There could be an empty config file present.

If --name was specified then use that as the configuration name, regardless of
whether or not there are existing configs.

Allow a 0 (infinite) retention to be specified with --retention when in
interactive mode.

### DIFF
--- a/cmd/influx/setup.go
+++ b/cmd/influx/setup.go
@@ -141,27 +141,13 @@ func setupF(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("instance at %q has already been setup", activeConfig.Host)
 	}
 
-	if _, err := os.Stat(dPath); err == nil {
-		existingConfigs, _ := localConfigSVC.ListConfigs()
-		if len(existingConfigs) > 0 {
-			// If there are existing configs then require that a name be
-			// spcified in order to distinguish this new config from what's
-			// there already.
-			if setupFlags.name == "" {
-				return errors.New("flag name is required if you already have existing configs")
-			}
-			if _, ok := existingConfigs[setupFlags.name]; ok {
-				return &influxdb.Error{
-					Code: influxdb.EConflict,
-					Msg:  fmt.Sprintf("config name %q already existed", setupFlags.name),
-				}
-			}
-		}
+	if err := validateNoNameCollision(localConfigSVC, setupFlags.name); err != nil {
+		return err
 	}
 
 	req, err := onboardingRequest()
 	if err != nil {
-		return fmt.Errorf("failed to retrieve data to setup instance: %v", err)
+		return fmt.Errorf("failed to setup instance: %v", err)
 	}
 
 	result, err := s.OnboardInitialUser(context.Background(), req)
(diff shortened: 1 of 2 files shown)


cmd/influx/setup.go
```diff
@@ -141,17 +141,20 @@ func setupF(cmd *cobra.Command, args []string) error {
...
-			return errors.New("flag name is required if you already have existing configs")
-		}
-		if _, ok := existingConfigs[setupFlags.name]; ok {
-			return &influxdb.Error{
-				Code: influxdb.EConflict,
-				Msg:  fmt.Sprintf("config name %q already existed", setupFlags.name),
+		existingConfigs, _ := localConfigSVC.ListConfigs()
+		if len(existingConfigs) > 0 {
+			// If there are existing configs then require that a name be
+			// spcified in order to distinguish this new config from what's
```
danxmoran, line 148:
```suggestion
			// specified in order to distinguish this new config from what's
```

cmd/influx/setup.go
```diff
@@ -141,17 +141,20 @@ func setupF(cmd *cobra.Command, args []string) error {
...
+			// there already.
+			if setupFlags.name == "" {
+				return errors.New("flag name is required if you already have existing configs")
+			}
+			if _, ok := existingConfigs[setupFlags.name]; ok {
+				return &influxdb.Error{
+					Code: influxdb.EConflict,
+					Msg:  fmt.Sprintf("config name %q already existed", setupFlags.name),
```
danxmoran, line 156:
```suggestion
					Msg:  fmt.Sprintf("config name %q already exists", setupFlags.name),
```

danxmoran (commented):
Thanks for the fixes, they look great! Could you add a line to CHANGELOG.md?

cmd/influx/setup.go
```diff
@@ -141,17 +141,20 @@ func setupF(cmd *cobra.Command, args []string) error {
...
-		// ignore the error if found nothing
-		if setupFlags.name == "" {
-			return errors.New("flag name is required if you already have existing configs")
-		}
-		if _, ok := existingConfigs[setupFlags.name]; ok {
-			return &influxdb.Error{
-				Code: influxdb.EConflict,
-				Msg:  fmt.Sprintf("config name %q already existed", setupFlags.name),
+		existingConfigs, _ := localConfigSVC.ListConfigs()
+		if len(existingConfigs) > 0 {
```
ayang64, line 146: can we invert this logic to save an indent level?

for example, could this be:

```
if len(existingConfigs) == 0 {
 break
}
```

danxmoran:
I'll get this one over the finish line.

cmd/influx/setup.go
danxmoran: `break` doesn't seem to work, since it's not in a loop. I'll poke at it some more.

danxmoran: Solution: move this logic to a separate function.

ayang64 (approved):
Looks great!

adrian-thurston:
Thanks @danxmoran and @ayang64. Came back today to work on it and saw you took care of it. I've been wrapping up a move into a new apartment and haven't been moving on things as quickly as I'd like!

danxmoran:
No worries 😄 hope the move went well!

PR #20168, opened 2020-11-24
fix(kv): add KV migration to repair DBRP mappings broken by schema change

### LINKED ISSUE #20167: Migrate DBRPs from old KV schema to new on startup
EDIT: Updated to reflect new knowledge

In 2.0.2 we changed the implementation of our DBRP API to match the spec. As part of the change, we modified the JSON tags for a DBRP struct. This unknowingly resulted in a breaking change to `influxd`'s KV metadata store, which persists DBRP mappings as JSON. Any DBRP objects created pre-2.0.2 will fail to fully deserialize from JSON after an upgrade to 2.0.2, causing their bucket and organization IDs to be zeroed out. This currently causes panics on the V1 write path, and even with a fix for the panic we see the DBRPs are inaccessible when cut off from their org / bucket.

A [work-around](https://github.com/influxdata/influxdb/issues/20121#issuecomment-731403423) exists to repair the mappings from the outside using the API, but it's fairly complex. We should use a KV migration to automatically rewrite any DBRP mappings using the old schema on startup.

danxmoran:
Closes #20167

I tested the new command locally, happy to refactor and set up automated tests if the reviewers would like.

### DIFF
--- a/kv/migration/all/0013_repair-DBRP-owner-and-bucket-IDs.go
+++ b/kv/migration/all/0013_repair-DBRP-owner-and-bucket-IDs.go
@@ -3,6 +3,7 @@ package all
 import (
 	"context"
 	"encoding/json"
+
 	"github.com/influxdata/influxdb/v2"
 	"github.com/influxdata/influxdb/v2/kv"
 )
@@ -15,8 +16,10 @@ var Migration0013_RepairDBRPOwnerAndBucketIDs = UpOnlyMigration(
 			Database        string      `json:"database"`
 			RetentionPolicy string      `json:"retention_policy"`
 			Default         bool        `json:"default"`
-			OrganizationID  influxdb.ID `json:"organization_id"`
-			BucketID        influxdb.ID `json:"bucket_id"`
+
+			// These 2 fields were renamed.
+			OrganizationID influxdb.ID `json:"organization_id"`
+			BucketID       influxdb.ID `json:"bucket_id"`
 		}
 
 		// Collect DBRPs that are using the old schema.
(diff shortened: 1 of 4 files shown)


danxmoran:
Marking as draft while I convert the logic to a KV migration (thanks for the idea @stuartcarnie!)

kv/migration/all/all.go
```diff
@@ -31,5 +31,9 @@ var Migrations = [...]migration.Spec{
 	Migration0011_PopulateDashboardsOwnerId,
 	// Populate the DBRP service ByOrg index
 	Migration0012_DBRPByOrgIndex,
+	// repair DBRP owner and bucket IDs
+	Migration0013_RepairDBRPOwnerAndBucketIDs,
+	// reindex DBRPs
+	Migration0014_ReindexDBRPs,
```
danxmoran, line 37: The reindex here is required for the `Find` methods in the DBRP service to locate repaired DBRPs. Reordering the migrations so my new `Migration0013` runs before the existing `Migration0012` also works, but that felt like a potentially-dangerous idea.

kv/migration/all/0013_repair-DBRP-owner-and-bucket-IDs.go
```diff
@@ -0,0 +1,119 @@
+package all
+
+import (
+	"context"
+	"encoding/json"
+	"github.com/influxdata/influxdb/v2"
+	"github.com/influxdata/influxdb/v2/kv"
+)
+
+var Migration0013_RepairDBRPOwnerAndBucketIDs = UpOnlyMigration(
```
danxmoran, line 10: The logic & structure of this file is largely copied from the [0012 migration](https://github.com/influxdata/influxdb/blob/master/kv/migration/all/0011_populate-dashboards-owner-id.go).

danxmoran:
To test this change, I:
1. Upgraded a local 1.8 DB using a v2.0.1 `influxd upgrade`
2. Ran a v2.0.2 `influxd` pointing at the upgraded data
3. Used a v2.0.2 `influx v1 dbrp list` to trigger the panic that occurs because of the data corruption
4. Used `influx v1 dbrp create` to create a new DBRP mapping using the v2.0.2 schema
4. Stopped the v2.0.2 `influxd`, and started a new one built from my branch
5. Used the v2.0.2 `influx v1 dbrp list` to confirm the panic was fixed, and all DBRPs were returned

GeorgeMac (approved):
Awesome. Cheers @danxmoran !

PR #20171, opened 2020-11-24
fix: Shutdown http servers before closing httpd service

ayang64:
Closes #

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/services/httpd/service.go
+++ b/services/httpd/service.go
@@ -2,6 +2,7 @@
 package httpd // import "github.com/influxdata/influxdb/services/httpd"
 
 import (
+	"context"
 	"crypto/tls"
 	"fmt"
 	"net"
@@ -10,6 +11,7 @@ import (
 	"path"
 	"runtime"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -56,6 +58,9 @@ type Service struct {
 	limit     int
 	tlsConfig *tls.Config
 	err       chan error
+
+	httpServerMutex sync.Mutex
+	httpServer      []*http.Server
 
 	unixSocket         bool
 	unixSocketPerm     uint32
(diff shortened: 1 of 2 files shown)


services/httpd/service.go
```diff
@@ -192,6 +197,16 @@ func (s *Service) Open() error {
 func (s *Service) Close() error {
 	s.Handler.Close()
 
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	s.httpServerMutex.Lock()
+	// shut down all running http servers
+	for _, server := range s.httpServer {
+		server.Shutdown(ctx)
```
davidby-influx, line 206: This may be overkill for the scenario, but I've seen timed cleanups/shutdowns in separate threads per instance, so one blocking doesn't time out all the others being closed.  That does require a counting semaphore or something similar, rather than a simple mutex.   Or, you may want to scale your timeout on the number of Shutdown() calls being made.

davidby-influx (approved):
Looks good; my comment is a suggestion only, and I approve the PR without any changes.

services/httpd/service.go
ayang64: this is good feedback.  thank you!  will make those changes.

services/httpd/service.go
```diff
@@ -192,6 +197,23 @@ func (s *Service) Open() error {
 func (s *Service) Close() error {
 	s.Handler.Close()
 
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	s.httpServerMutex.Lock()
+	errCh := make(chan error, len(s.httpServer))
+	// shut down all running http servers
+	for _, server := range s.httpServer {
```
davidby-influx, line 206: Should the two `for` loops  be a go func, with the `unlock` in a `defer` so that if the context timeout occurs, you don't hold the mutex?  Maybe it's cleared automatically; I'm not certain how Go does it.  So, lines 206 - 214 become a synchronously called go func (which spawns the asynchronous shutdowns as currently), but has line 215 as a defer on that little func?

davidby-influx (approved):
Comment in the code is a suggestion, not sure if necessary.

PR #20190, opened 2020-11-27
fix: properly shut down httpd service

ayang64:
InfluxDB does not `Shutdown()` http servers when the service is `Closed()`.  This patch properly shuts them down.

Closes #

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/services/httpd/service.go
+++ b/services/httpd/service.go
@@ -2,6 +2,7 @@
 package httpd // import "github.com/influxdata/influxdb/services/httpd"
 
 import (
+	"context"
 	"crypto/tls"
 	"fmt"
 	"net"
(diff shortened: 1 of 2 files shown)


services/httpd/service.go
```diff
@@ -245,10 +267,16 @@ func (s *Service) serveUnixSocket() {
 
 // serve serves the handler from the listener.
 func (s *Service) serve(listener net.Listener) {
+	svr := &http.Server{
```
rogpeppe, line 270: I _think_ that if you just use a single `http.Server` instance for all the listeners, you can just use a single Shutdown call on that.

rogpeppe (commented):
Looks reasonable, but I think this could be a bit simpler. If it can't then perhaps add a comment indicating to future readers why that's the case?

davidby-influx (approved):
This looks good, and I know it will be superseded by the next PR which looks even better.

Issue #19976, opened 2020-11-10
Tech debt: refactor influxd's "run" method for better readability / testability


danxmoran:
influxd's [`run` method](https://github.com/influxdata/influxdb/blob/master/cmd/influxd/launcher/launcher.go#L658) is ~800 lines long. This makes it very hard to tell how different pieces of the DB interact with and depend on one another during setup, especially when using middleware that doesn't affect types (i.e. applying an authorization wrapper).

It should be possible to refactor `run` into smaller pieces and make it easier to understand.

PR #20196, opened 2020-11-30
refactor(cmd/influxd): parse log-level CLI opts directly to correct type

danxmoran:
Part of #19976 

This took more boilerplate than I expected, I could be convinced it's not worth the trouble.

### DIFF
--- a/kit/cli/loglevelflag.go
+++ b/kit/cli/loglevelflag.go
@@ -1,0 +1,42 @@
+package cli
+
+import (
+	"fmt"
+	"github.com/spf13/pflag"
+
+	"go.uber.org/zap/zapcore"
+)
+
+type levelValue zapcore.Level
+
+func newLevelValue(val zapcore.Level, p *zapcore.Level) *levelValue {
+	*p = val
+	return (*levelValue)(p)
+}
+
+func (l *levelValue) String() string {
+	return zapcore.Level(*l).String()
+}
+func (l *levelValue) Set(s string) error {
+	var level zapcore.Level
+	if err := level.Set(s); err != nil {
+		return fmt.Errorf("unknown log level; supported levels are debug, info, warn, error")
+	}
+	*l = levelValue(level)
+	return nil
+}
+
+func (l *levelValue) Type() string {
+	return "Log-Level"
+}
+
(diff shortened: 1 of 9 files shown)


kit/cli/loglevelflag.go
```diff
@@ -0,0 +1,42 @@
+package cli
+
+import (
+	"fmt"
+
+	"github.com/spf13/pflag"
+	"go.uber.org/zap/zapcore"
+)
+
+type levelValue zapcore.Level
```
danxmoran, line 10: This boilerplate file is pretty much identical to `kit/cli/idflag.go`, but with `zapcore.Level` instead of `influxdb.ID`.

kit/cli/viper.go
```diff
@@ -274,6 +275,19 @@ func BindOptions(v *viper.Viper, cmd *cobra.Command, opts []Opt) {
...
+				l = o.Default.(zapcore.Level)
+			}
+			if hasShort {
+				LevelVarP(flagset, destP, o.Flag, string(o.Short), l, o.Desc)
+			} else {
+				LevelVar(flagset, destP, o.Flag, l, o.Desc)
+			}
+			if s := v.GetString(envVar); s != "" {
+				_ = (*destP).Set(v.GetString(envVar))
+			}
```
danxmoran, line 290: This block is mostly boilerplate, following the same pattern as the preceding `case`s

kit/cli/loglevelflag.go
```diff
@@ -0,0 +1,42 @@
+package cli
+
+import (
+	"fmt"
+	"github.com/spf13/pflag"
+
+	"go.uber.org/zap/zapcore"
+)
+
+type levelValue zapcore.Level
```
davidby-influx, line 10: Is this definition just a way to declare methods on the zapcore.Level type?  Nice trick!

davidby-influx (approved):

kit/cli/loglevelflag.go
danxmoran: (I think) yes, specifically so we can redefine methods like `String` and `Set` that already exist on the zapcore.Level type.

PR #20203, opened 2020-11-30
fix(dbrp): Don't use record ID removing orgID index

brettbuddin:
Noticed this while I was going through porting some of this to IDPE. The new Org-based index for DBRP mappings won't get cleaned up properly without this.


stuartcarnie (approved):
Thanks for fixing this!

GeorgeMac (approved):

PR #20219, opened 2020-12-01
chore: update CHANGELOG

danxmoran:
I preemptively carved out a section for 2.0.4 when I thought we were releasing 2.0.3. We've decided to wait a bit longer on that release, so reorganizing to match.


psteinbachs (approved):
lgtm

PR #20220, opened 2020-12-01
fix(dbrp): Don't use record ID removing orgID index

danxmoran:
Backports #20203 

This is a fix for other unreleased logic sitting in the `2.0` branch.


psteinbachs (approved):
Looks fine.

PR #20227, opened 2020-12-01
refactor(cmd/influxd): unify logger-construction in launcher

danxmoran:
Part 3/??? of #19976 

Pushing logger construction out of the `Launcher` doesn't have that big of an effect on the `run` method itself, but I think it's a cleaner division of responsibilities, especially in test code.


cmd/influxd/launcher/cmd.go
```diff
@@ -74,17 +76,28 @@ func cmdRunE(ctx context.Context, o *InfluxdOpts) func() error {
...
+			return err
+		}
+
+		if err := l.run(ctx, o, log); err != nil {
 			return err
 		}
 		<-ctx.Done()
 
 		// Attempt clean shutdown.
 		ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
```
ayang64, line 96: i'm not sure if we can re-use a context this way.  you might need a new context value that isn't cancelled -- i don't think you can un-cancel a context chain.

danxmoran: Will investigate, it'd be nice to squash a bonus bug as part of this

ayang64 (changes requested):

cmd/influxd/launcher/cmd.go
danxmoran: This seems to match the pattern from `context.WithTimeout`'s godoc:
```go
// WithTimeout returns WithDeadline(parent, time.Now().Add(timeout)).
//
// Canceling this context releases resources associated with it, so code should
// call cancel as soon as the operations running in this Context complete:
//
// 	func slowOperationWithTimeout(ctx context.Context) (Result, error) {
// 		ctx, cancel := context.WithTimeout(ctx, 100*time.Millisecond)
// 		defer cancel()  // releases resources if slowOperation completes before timeout elapses
// 		return slowOperation(ctx)
// 	}
```
The following call to `l.Shutdown(ctx, log)` being the slow operation.

cmd/influxd/launcher/cmd.go
```diff
@@ -66,26 +69,45 @@ func setCmdDescriptions(cmd *cobra.Command) {
 `
 }
 
-func cmdRunE(ctx context.Context, o *InfluxdOpts) func() error {
+func cmdRunE(o *InfluxdOpts) func() error {
```
ayang64, line 72: this was better with a passed in context.   prior to this change, the caller could cancel the entire call chain.

cmd/influxd/launcher/cmd.go
```diff
@@ -66,26 +69,45 @@ func setCmdDescriptions(cmd *cobra.Command) {
...
+			Level:  o.LogLevel,
+		}
+		logger, err := logconf.New(os.Stdout)
+		if err != nil {
+			return err
+		}
 
-		l := NewLauncher()
-		if err := l.run(ctx, o); err != nil {
+		if err := l.runUntilCanceled(o, logger); err != nil {
```
ayang64, line 87: probably should just accept a `context.Context` value.

cmd/influxd/launcher/cmd.go
```diff
@@ -66,26 +69,45 @@ func setCmdDescriptions(cmd *cobra.Command) {
...
-		defer cancel()
-		l.Shutdown(ctx)
+		return logger.Sync()
+	}
+}
 
-		return nil
+// runUntilCanceled runs the given launcher and waits, only exiting when a SIGINT or SIGTERM is received.
+func (l *Launcher) runUntilCanceled(o *InfluxdOpts, log *zap.Logger) error {
+	ctx := signals.WithStandardSignals(context.Background())
```
ayang64, line 98: should probably create this context value before calling `cmdRunE()` and pass that value up the call chain.

cmd/influxd/launcher/launcher_helpers.go
```diff
@@ -49,26 +49,35 @@ type TestLauncher struct {
...
 
 	// Flag to act as standard server: disk store, no-e2e testing flag
 	realServer bool
+
+	log *zap.Logger
+}
+
+// RunAndSetupNewLauncherOrFail shorcuts the most common pattern used in testing,
+// building a new TestLauncher, running it, and setting it up with an initial user.
+func RunAndSetupNewLauncherOrFail(tb testing.TB, ctx context.Context, setters ...OptSetter) *TestLauncher {
```
ayang64, line 63: typically `ctx` is the first parameter.

ayang64 (changes requested):

danxmoran:
Closing this for now, going to rework it

PR #20230, opened 2020-12-02
fix(query): Group By queries with offset that crosses a DST boundary can fail

davidby-influx:
Customer reported that a GROUP BY query with an offset that caused an interval
to cross a daylight savings change inserted an extra output row off by one hour.

This fix ensured that the start time for the interval of a GROUP BY operator is
correctly set before calculating the time zone offset for that date and time. 

- [X] Rebased/mergeable
- [X] Tests pass

Fixes https://github.com/influxdata/influxdb/issues/20238

### DIFF
--- a/query/iterator.go
+++ b/query/iterator.go
@@ -815,6 +815,8 @@ func (opt IteratorOptions) Window(t int64) (start, end int64) {
 		start = t - dt
 	}
 
+	start += int64(opt.Interval.Offset)
+
 	// Look for the start offset again because the first time may have been
 	// after the offset switch. Now that we are at midnight in UTC, we can
 	// lookup the zone offset again to get the real starting offset.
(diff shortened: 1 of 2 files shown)


gshif (approved):

query/iterator.go
```diff
@@ -815,6 +815,8 @@ func (opt IteratorOptions) Window(t int64) (start, end int64) {
 		start = t - dt
 	}
 
+	start += int64(opt.Interval.Offset)
```
jsternberg, line 818: Can you add some tests for this behavior?

jsternberg (changes requested):

query/iterator.go
davidby-influx: Can do.  @gshif  developed a truly minimal repro case I can encapsulate in a test.

davidby-influx:
Added a regression test, TestGroupByIterator_DST() in query/iterator_test.go.  @jsternberg, take a look at your convenience.

jsternberg (approved):

PR #20234, opened 2020-12-02
refactor: save field set after restoring

rickif:
This PR is to remove the extra rebuilding of engine's field set in the process of `Restore()/Import()`. 
The engine would load the `.tsm` file to rebuild the field set when `Restore()/Import()` is called, while the field set data is not persisted after the operation is done. After `Restore()/Import()`, the engine would be reopen,  in which the engine would load data and rebuild the field set again.
The second rebuilding is unnecessary and could be avoided via saving the engine's field set to `fields.idx`.

- [x] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -1182,6 +1182,12 @@ func (e *Engine) overlay(r io.Reader, basePath string, asNew bool) error {
 			return err
 		}
 	}
+
+	// Save the field set index so we don't have to rebuild it next time
+	if err := e.fieldset.Save(); err != nil {
+		return err
+	}
+
 	return nil
 }
 


rickif:
@danatinflux Hi, would you please do review for this PR ?

danatinflux:
@rickif Preeeettty sure you have the wrong Dan. :)

lesam:
@rickif I'm concerned that if we merge this, we may have issues with ImportShard calls that import TSM files into an existing shard (as used, for example, by the Enterprise anti-entropy service).

Is there a serious performance concern from doing the rebuild? Also we are trying to encourage users to migrate to 2.x, I would be much more inclined to put in the work to investigate this change if it was against the `master` branch where we are focussing our development effort.

Issue #20249, opened 2020-12-04
unsupported value: +Inf" error not handled gracefully


davidby-influx:
JSON marshalling errors should be returned as properly formatted JSON errors.

__Steps to reproduce:__
```
> create database db
> use db
Using database db
> insert m v=1.7976931348623157E308
> insert m v=1.7976931348623157E308
> 
> select * from m
name: m
time                v
----                -
1604428620720870437 1.7976931348623157e+308
1604428621584887557 1.7976931348623157e+308

> select sum(v) from m
ERR: 
> exit
$ curl "http://localhost:8086/query?db=EAR1834" --data-urlencode "q=select sum(v) from m"  
```

__Expected behavior:__
```
$ curl "http://localhost:8086/query?db=EAR1834" --data-urlencode "q=select sum(v) from m"  
{"error":"json: unsupported value: +Inf"}
```

__Actual behavior:__
```
$ curl "http://localhost:8086/query?db=db" --data-urlencode "q=select sum(v) from m"  
json: error calling MarshalJSON for type httpd.Response: json: error calling MarshalJSON for type *query.Result: json: unsupported value: +Inf

```
__Environment info:__

* v1.8.3

davidby-influx:
Pushed.

ttps://github.com/influxdata/influxdb/pull/20250

PR #20250, opened 2020-12-04
fix(error): unsupported value: +Inf" error not handled gracefully

davidby-influx:
JSON marshaling errors should be returned properly formatted in JSON
like other errors. 

Closes https://github.com/influxdata/influxdb/issues/20249

This fix formats marshaling errors the same way
influxdb formats other query errors.

- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/services/httpd/response_writer.go
+++ b/services/httpd/response_writer.go
@@ -3,6 +3,7 @@ package httpd
 import (
 	"encoding/csv"
 	"encoding/json"
+	"errors"
 	"io"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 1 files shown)


services/httpd/response_writer.go
```diff
@@ -124,19 +129,25 @@ type jsonFormatter struct {
...
 		b, err = json.MarshalIndent(resp, "", "    ")
 	} else {
 		b, err = json.Marshal(resp)
 	}
 
 	if err != nil {
-		_, err = io.WriteString(w, err.Error())
-	} else {
-		_, err = w.Write(b)
+		err = unnestError(err)
```
dgnorton, line 140: Need to format this error as json (e.g., `{"error":"json: unsupported value: +Inf"}`) and write to the client.

dgnorton (changes requested):

services/httpd/response_writer.go
davidby-influx: The error is correctly formatted by the JSON marshaller.  The top level error looks like:

{"error":"json: error calling MarshalJSON for type httpd.Response: json: error calling MarshalJSON for type *query.Result: json: unsupported value: +Inf"}

The deepest error after unnesting is this:

{"error":"json: unsupported value: +Inf"}

davidby-influx: The write to the client happens in the caller, (*responseWriter).WriteResponse(), rather than in (*jsonFormatter).WriteResponse() in the previous version.

dgnorton (approved):

PR #20260, opened 2020-12-06
build: create Dockerfile for CI / release

### LINKED ISSUE #20254: Create a Docker image for building influxdb
The image should contain all the dependencies needed for our CircleCI pipelines, including cross-build support. It should be pushed to `quay.io/influxdb/influxdb-circleci`

danxmoran:
Closes #20254 

Using this image, I was able to build binaries for ARMv6, ARMv7, and ARM64.


Makefile
```diff
@@ -167,24 +167,23 @@ bench:
 
 build: all
 
-goreleaser:
-	curl -sfL -o goreleaser-install https://install.goreleaser.com/github.com/goreleaser/goreleaser.sh
-	sh goreleaser-install v0.142.0
+pkg-config:
+	# FIXME: Use an unofficial release for now, to get the ARM preview out.
+	go get github.com/influxdata/pkg-config@dm-no-modify-cc
```
danxmoran, line 172: See influxdata/pkg-config#28

docker/circleci/Dockerfile
```diff
@@ -0,0 +1,106 @@
...
+        libtinfo5 \
+        libtool \
+        libxml2-dev \
+        llvm-dev \
+        lzma-dev \
+        netcat-openbsd \
+        patch \
+        xutils-dev \
+        zlib1g-dev && \
+    sudo rm -rf rm -rf /var/lib/apt/lists/*
```
gshif, line 27: I think `rm -rf` was meant to be only once

gshif (approved):
It is good to go with addressing one comment about `rm -rf`

docker/circleci/Dockerfile
```diff
@@ -0,0 +1,91 @@
...
+ENV MUSL_VERSION=1.1.24
+RUN curl https://musl.libc.org/releases/musl-${MUSL_VERSION}.tar.gz -O && \
+    tar xzf musl-${MUSL_VERSION}.tar.gz && \
+    cd musl-${MUSL_VERSION} && \
+    ./configure && \
+    make && \
+    sudo make install && \
+    cd .. && \
+    rm -rf musl-${MUSL_VERSION}*
+# TODO: Convert these to curl calls to download from S3.
```
danxmoran, line 40: Before merge: complete this TODO

.goreleaser.yml
```diff
@@ -64,9 +95,9 @@ nfpms:
       rpm:
         replacements:
           amd64: x86_64
-        file_name_template: "influxdb2-{{ .Version }}.{{ .Arch }}{{if .Arm}}{{ if eq .Arm \"5\" }}el{{end}}{{ end }}{{if .Arm}}{{ if eq .Arm \"6\" }}hf{{end}}{{ end }}"
+        file_name_template: "influxdb2-{{ .Version }}.{{ .Arch }}{{ if .Arm }}v{{ .Arm }}hf{{ end }}"
```
danxmoran, line 98: TODO before merge: Make sure these template changes don't break anything.

* The case for 5 was removed because we're not building for `GOARM=5`.
* `GOARM=6` should have the same output.
* `GOARM=7` will now output with an `hf` suffix

danxmoran: Decision: keep v6 as `armhf` suffix, use `armv7hf` for v7

.goreleaser-nightly.yml
```diff
@@ -6,18 +6,31 @@ builds:
...
+      - goos: darwin
+        goarch: arm
+        goarm: 6
+      - goos: darwin
+        goarch: arm
+        goarm: 7
     main: ./cmd/influx/
     flags:
-      - -tags={{if and (eq .Os "linux") (eq .Arch "amd64")}}osusergo,netgo,static_build{{end}}
+      - -tags={{if eq .Os "linux"}}osusergo,netgo,static_build{{end}}
```
danxmoran, line 25: Before merge: I think we might need to set `noasm` in the tags here for non-amd64 `GOARCH`, if this bypasses the `Makefile`

.goreleaser-nightly.yml
```diff
@@ -6,18 +6,31 @@ builds:
       - darwin
     goarch:
       - amd64
-
+      - arm64
+      - arm
+    goarm:
+      - 6
+      - 7
```
danxmoran, line 13: Before merge: @psteinbachs says we've officially dropped support for 32-bit architectures in 2.x, so we can remove all the code to handle GOARCH=arm

docker/circleci/Dockerfile
danxmoran: Update: if all we need is ARM64, we can revert to building the cross-compiler here. It'll still be slow, but not unbearably so.

Makefile
danxmoran: This is now merged, will update go.mod to use the new tag

PR #20266, opened 2020-12-07
chore: Fix build for clustering

lesam:
Closes #

- [ ] Rebased/mergeable
- [ ] Tests pass


services/meta/data.go
```diff
@@ -1363,12 +1363,15 @@ func (sgi ShardGroupInfo) clone() ShardGroupInfo {
...
 }
 
-// ShardFor returns the ShardInfo for a Point.
-func (sgi *ShardGroupInfo) ShardFor(p models.Point) ShardInfo {
+type hashIDer interface {
+	HashID() uint64
+}
+
+// ShardFor returns the ShardInfo for a Point or other hashIDer.
+func (sgi *ShardGroupInfo) ShardFor(p hashIDer) ShardInfo {
```
lesam, line 1371: `hashIDer` is a much easier requirement than `Point` (an interface subset)

danxmoran: Wouldn't have thought of this, I like it!

danxmoran (approved):
LGTM, the ImportShard impl looks identical to the old code

services/meta/data.go
dgnorton: Since `ShardFor` is public, `hashIDer` should probably be public also?

lesam:
> LGTM, the ImportShard impl looks identical to the old code

Yeah I just reverted the PR that removed it with `git revert`

services/meta/data.go
lesam: The interface only needs to be exported if we want to re-use it elsewhere, which we don't (I think), and also this package would not be the right one to export from.

I have validated locally that I can use this to fix the clustering build.

Issue #20256, opened 2020-12-04
tcp.Mux.Serve() closes all net.Listener instances silently on error.


davidby-influx:
__Steps to reproduce:__
In the file influxdb/tcp/mux.go, tcp.Mux.Serve() will silently close all the net.Listener instances in the Mux structure if there is an error on a net.Listener.Accept() call. 

__Expected behavior:__
The error should be logged for diagnosis.  A customer has seen behavior which indicates that this may be happening, but the absence of error-logging hinders investigation. 

__Actual behavior:__
The error is discarded.

__Environment info:__

* InfluxDB version: 1.8.3

davidby-influx:
Pull request:

https://github.com/influxdata/influxdb/pull/20278

PR #20278, opened 2020-12-08
fix: cp.Mux.Serve() closes all net.Listener instances silently on error.

davidby-influx:
A customer has seen a rash of "connection refused" errors to the meta node.
    This fix ensures that when net.Listener instances are closed because of an
    error in Accept(), influxdb logs the error which caused the closures, as well
    as any errors in closing the Listeners.
    
    Fixes https://github.com/influxdata/influxdb/issues/20256

- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/tcp/mux.go
+++ b/tcp/mux.go
@@ -79,7 +79,7 @@ func (mux *Mux) Serve(ln net.Listener) error {
 			continue
 		}
 		if err != nil {
-			mux.Logger.Printf("Listener at %s failed failed to accept a connection, closing all listeners - %s", ln.Addr(), err)
+			mux.Logger.Printf("tcp.Mux: Listener at %s failed failed to accept a connection, closing all listeners - %s", ln.Addr(), err)
 			// Wait for all connections to be demux
 			mux.wg.Wait()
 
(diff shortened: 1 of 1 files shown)


tcp/mux.go
```diff
@@ -79,6 +79,7 @@ func (mux *Mux) Serve(ln net.Listener) error {
 			continue
 		}
 		if err != nil {
+			mux.Logger.Printf("tcp.Mux.Serve: call to Accept() at %s failed, closing all listeners - %s", ln.Addr(), err)
```
danxmoran, line 82: Is it possible to set a level on these log messages? i.e. error or warning?

danxmoran (dismissed):
One Q for a potential improvement, LGTM overall

tcp/mux.go
dgnorton: Most logging in InfluxDB is through the `zap` logger but `Mux` hasn't been updated to use that. We'd either have to update it to use `zap` or simulate the same structured logging format.

@davidby-influx since one intended audience of the log is users that may not read code, we generally avoid logging code. Can you convert these messages to something unique enough to allow engineers to located the source but also more friendly to users / non-coders?

dgnorton (changes requested):

tcp/mux.go
davidby-influx: Rewrote log messages to match formatting of other message from tcp.Mux, made more user-friendly.

dgnorton (approved):

tcp/mux.go
ayang64: In general you may want to consider using the `%v` verb for printing error messages; not `%s`.

PR #20298, opened 2020-12-09
docs: OSX always needs pkg-config

lesam:
Tiny PR to fix up the docs - OSX build appears to need pkg-config with normal 'make' build, so add it to the docs.

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,10 +119,10 @@ Currently the project only depends on `git` and `bzr`.
 
 You need a recent stable version of Rust. We recommend using [rustup](https://rustup.rs/) to install Rust.
 
-You also need `protobuf`, `yarn`, and `make` installed.
+You also need `clang`, `make`, `pkg-config`, `protobuf`, and `yarn` installed.
 
-- OSX: `brew install protobuf yarn make`
-- Linux (Arch): `pacman -S protobuf yarn make`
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -119,9 +119,9 @@ Currently the project only depends on `git` and `bzr`.
 
 You need a recent stable version of Rust. We recommend using [rustup](https://rustup.rs/) to install Rust.
 
-You also need `protobuf`, `yarn`, and `make` installed.
+You also need `protobuf`, `yarn`, and `make` installed. Depending on how you build you may need pkg-config.
 
-- OSX: `brew install protobuf yarn make`
+- OSX: `brew install protobuf yarn make pkg-config`
```
danxmoran, line 124: Can you also add `clang` here, and add both to the sections for linux? To match the setup info in the `Building with the go command` section below.

Tangential, I would advocate for deleting the `Building with the go command` section entirely. I've been poking around the `Makefile` and there's a lot more setup logic there than I knew about, I don't think we should present it as something that can be worked around.

lesam: Updated, without the tangent

lesam:
Rebased

danxmoran (approved):

PR #20305, opened 2020-12-10
fix: Improve v2 packaging

psteinbachs:
Packaging changes:
* Package name is now influxdb2 and conflicts with any previous influxdb package (including initial 2.0.0, 2.0.1, and 2.0.2 packages)
  - Requires explicit removal of previous package before installing new package
* Creates an /etc/default/influxdb2 file if non existent, which sets environment defaults
  - INFLUXD_CONFIG_PATH=/etc/influxdb/config.toml
  - INFLUXD_BOLT_PATH=/var/lib/influxd/influxd.bolt
  - INFLUXD_ENGINE_PATH=/var/lib/influxdb/engine
* Systemd service file now reads /etc/default/influxdb2
* Detects if v2 directory structure/files exist and outputs upgrade message if not
* Generates an upgrade script under /usr/share/influxdb/influxdb2-upgrade.sh
  - Runs 'influxd upgrade' with appropriate arguments
  - If exit code successful, outputs a message about further steps, v1 cleanup via helper script
  - If exit code unsuccessful, outputs a message about further steps, v2 cleanup via helper script

- [x] Rebased/mergeable


scripts/influxdb2-upgrade.sh
```diff
@@ -0,0 +1,38 @@
...
+  -m $INFLUXD_BOLT_PATH -e $INFLUXD_ENGINE_PATH
+
+if [[ $? -eq 0 ]]; then
+cat << EOF
+
+The upgrade completed successfully.  Execute the following to start InfluxDB:
+
+sudo systemctl start influxdb
+EOF
+fi
```
danxmoran, line 34: Should we have a message + early exit here for the failure case? Upgrade isn't idempotent, so it could be nice to print an `rm` command for cleanup in the case of failure.

scripts/influxdb2-upgrade.sh
```diff
@@ -0,0 +1,38 @@
...
+    echo "already upgraded to v2.  Exiting."
+    exit 1
+  fi
+done
+
+# Perform upgrade
+sudo /usr/bin/influxd upgrade \
+  --config-file=/etc/influxdb/influxdb.conf \
+  --v2-config-path=${INFLUXD_CONFIG_PATH} \
+  -m $INFLUXD_BOLT_PATH -e $INFLUXD_ENGINE_PATH
```
danxmoran, line 25: There are a few other outputs for `upgrade` that default to the home directory:
* Log file (`--log-path`)
* CQ export (`--continuous-query-export-path`)

Should we set those here too?

scripts/influxdb2-upgrade.sh
```diff
@@ -0,0 +1,38 @@
...
+if [[ $? -eq 0 ]]; then
+cat << EOF
+
+The upgrade completed successfully.  Execute the following to start InfluxDB:
+
+sudo systemctl start influxdb
+EOF
+fi
+
+sudo cp /root/.influxdbv2/configs /var/lib/influxdb
```
danxmoran, line 36: Would `/etc/influxdb` be a better choice, since it's config?

danxmoran:
Should we include this in CHANGELOG.md? For the ARM build I made a [top-level callout](https://github.com/influxdata/influxdb/pull/20260/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR3-R6), we could maybe broaden it to something like "packaging changes"

scripts/influxdb2-upgrade.sh
psteinbachs: Most recent commit expands the upgrade script for both pass/fail cases, and outputs a respective v1 or v2 data removal script.  They need to be double-checked to make sure it's removing only what's necessary.

scripts/influxdb2-upgrade.sh
psteinbachs: Added these.

scripts/influxdb2-upgrade.sh
psteinbachs: Per our conversation on this one, if we were to write it under /etc/influxdb then the user would have to pass the path as a cli argument or set an environment variable.  Currently we write it to the home directories of root and influxdb users which in theory will just work without additional effort.

CHANGELOG.md
```diff
@@ -33,6 +33,10 @@ want to use the default.
 1. [20155](https://github.com/influxdata/influxdb/pull/20155): Respect the `--name` option in `influx setup` whether configs already exist or not.
 1. [20155](https://github.com/influxdata/influxdb/pull/20155): Allow for 0 (infinite) values for `--retention` in `influx setup`.
 
+### Packaging Improvements
+
+1. [20305](https://github.com/influxdata/influxdb/pull/20305): Set sane defaults and provide upgrade helper scripts
```
danxmoran, line 38: When this is rebased pre-merge, I'd move this up to live in the same section as the ARM64 support (and rename the section to something generic like "Release Improvements").

danxmoran:
```suggestion
1. [20305](https://github.com/influxdata/influxdb/pull/20305): Set sane defaults and provide upgrade helper scripts in release packages
```

scripts/influxdb2-upgrade.sh
```diff
@@ -0,0 +1,82 @@
...
+
+NOTE: This script will erase all previous v2 data and config files!
+
+EOF
+
+cat << EOF > /var/tmp/influxdbv2-remove.sh
+#!/bin/bash
+sudo rm /etc/influxdb/config.toml /var/lib/influxdb/influxd.bolt
+sudo rm /var/lib/influxdb/configs /root/.influxdbv2/configs
+sudo rm -rf /var/lib/influxdb/engine
```
danxmoran, line 78: Should rm `/var/lib/influxdb/continuous_queries.txt` as well, and maybe `/var/log/influxdb/upgrade.log` (unless that'd be good to leave for debugging?)

scripts/influxdb2-upgrade.sh
```diff
@@ -0,0 +1,82 @@
...
+sudo /var/tmp/influxdbv2-remove.sh
+
+NOTE: This script will erase all previous v2 data and config files!
+
+EOF
+
+cat << EOF > /var/tmp/influxdbv2-remove.sh
+#!/bin/bash
+sudo rm /etc/influxdb/config.toml /var/lib/influxdb/influxd.bolt
+sudo rm /var/lib/influxdb/configs /root/.influxdbv2/configs
```
danxmoran, line 77: Should we include `-f` on these 2 lines so it succeeds even if the upgrade process didn't reach the point of writing some files?

scripts/influxdb2-upgrade.sh
psteinbachs: Added continuous_queries.txt removal, but leaving upgrade.log since the message tells the user to review it.

scripts/influxdb2-upgrade.sh
psteinbachs: Added -f for files.

danxmoran (approved):
:shipit:

scripts/influxdb2-upgrade.sh
```diff
@@ -0,0 +1,83 @@
...
+
+sudo /var/tmp/influxdbv1-remove.sh
+
+NOTE: This script will erase all previous v1 data and config files!
+
+EOF
+
+cat << EOF > /var/tmp/influxdbv1-remove.sh
+#!/bin/bash
+sudo rm /etc/influxdb/influxdb.conf
```
danxmoran, line 49: Ah, if you need to rebase anyway, this one could also use a `-f`

CHANGELOG.md
psteinbachs: Updated this with callout blurb.

scripts/influxdb2-upgrade.sh
psteinbachs: Yeah I caught this one too, added.

PR #20313, opened 2020-12-11
fix(tsdb): never use an inmem index

### LINKED ISSUE #20257: InfluxDB 2.0 sometimes uses inmem index instead of TSI
__Steps to reproduce:__
_we ran into this issue while trying to do something else, I have not attempted to create a minimum set of steps_

1. Upgrade an influx 1.8 database with > 1M series cardinality
2. Start inserting data using the new flux line protocol
3. Wait for a fair amount of time

__Expected behavior:__
Data to be ingested as normal

__Actual behavior:__
We are inserting with many different threads, and occasionally some threads get stuck with the following behaviour.

- We occasionally get an (HTTP 500) error with a body like `{"code":"internal error","message":"unexpected error writing points to database: partial write: max-series-per-database limit exceeded: (1000000) dropped=5"}` when writing to InfluxDB

__Environment info:__

`Linux 5.4.0-1029-aws x86_64`

`InfluxDB 2.0.2 (git: 84496e507a) build_date: 2020-11-19T03:59:35Z`

This max-series-per-database setting is not documented for InfluxDB v2 and only exists for the inmem index type which we don’t use (use tsi1 on InfluxDB v1.8, there is no option to change the index type on InfluxDB v2.0 afaik)

Seems like we get a message like this when InfluxDB opens a new shard or is restarted:
(here are those counts over time.. the tsi1_count does seem to be increasing, while the inmem_count decreased after a restart)

```
Dec 01 12:34:14 influxdb-experiment01 influxd[55446]: ts=2020-12-01T12:34:14.272747Z lvl=warn msg="Mixed shard index types" log_id=0QfXUnBl000 service=storage-engine inmem_count=15 tsi1_count=1 db_instance=e06881594d095d90
Dec 01 12:34:31 influxdb-experiment01 influxd[55446]: ts=2020-12-01T12:34:31.212353Z lvl=warn msg="Mixed shard index types" log_id=0QfXUnBl000 service=storage-engine inmem_count=15 tsi1_count=2 db_instance=e06881594d095d90
Dec 01 12:35:35 influxdb-experiment01 influxd[55446]: ts=2020-12-01T12:35:35.135699Z lvl=warn msg="Mixed shard index types" log_id=0QfXUnBl000 service=storage-engine tsi1_count=3 inmem_count=15 db_instance=e06881594d095d90
Dec 01 13:07:36 influxdb-experiment01 influxd[55446]: ts=2020-12-01T13:07:36.543219Z lvl=warn msg="Mixed shard index types" log_id=0QfXUnBl000 service=storage-engine inmem_count=15 tsi1_count=4 db_instance=e06881594d095d90
Dec 01 23:02:15 influxdb-experiment01 influxd[55446]: ts=2020-12-01T23:02:15.668045Z lvl=warn msg="Mixed shard index types" log_id=0QfXUnBl000 service=storage-engine inmem_count=15 tsi1_count=5 db_instance=e06881594d095d90
Dec 02 09:31:01 influxdb-experiment01 influxd[2116776]: ts=2020-12-02T09:31:01.280938Z lvl=warn msg="Mixed shard index types" log_id=0QphRlrW000 service=storage-engine tsi1_count=5 inmem_count=11 db_instance=e06881594d095d90
```

some notes from our engineer investigating this:

> I think possibly what’s happening is that the upgrade process created inmem shard indexes instead of tsi1

> ok staring at more influxdb source code will make me crazy.. I don’t know why we’re getting inmem indexes--those are the only kind that have the max-series-per-database error and the only place we use them is the legacy influx05

> In InfluxDB v1.8 there was a tool for transferring between inmem and tsi1 and inspecting the current indexes (influx_inspect). I can’t find that in InfluxDB v2.0 (the binary is not provided in any of their downloads and while it looks like the command still exists in the source, most of the functionality has been stripped out)

__Config:__
- Tried setting the (undocumented) max-series-per-database option in the config file, had no effect

__Logs:__
see above

__Performance:__
n/a

danxmoran:
Closes #20257 

2.x was never meant to use an inmem index, but there was a fallback code path to handle cases when `influxd upgrade` didn't convert an inmem index into TSI. All code for the inmem index is now deleted, and the fallback is replaced by a one-time reindex operation run on shards w/o a TSI directory.

The reindex logic is inspired by the [`influx inspect` implementation](https://github.com/influxdata/influxdb/blob/master/cmd/influx_inspect/buildtsi/buildtsi.go#L341), but uses methods on `Engine` to read TSM / WAL data instead of listing and reading files directly from the file system.


danxmoran:
@stuartcarnie could you take a look at this one? I'm pretty confident the existing gotests cover the logic (many were failing when I still had bugs in `Reindex()`), but I'm not familiar enough with the code to be 100% sure.

dgnorton (changes requested):

danxmoran:
Other note from Zoom review: try to add an integration test to check that index files are (re)generated if missing, and that query results are the same.

tsdb/engine/tsm1/engine.go
```diff
@@ -3140,3 +3098,75 @@ func varRefSliceRemove(a []influxql.VarRef, v string) []influxql.VarRef {
...
+	n := 0
+
+	reindexBatch := func() error {
+		if n == 0 {
+			return nil
+		}
+
+		for i, key := range keys[:n] {
+			seriesKeys[i], _ = SeriesAndFieldFromCompositeKey(key)
+			names[i], tags[i] = models.ParseKeyBytesWithTags(seriesKeys[i], tags[i])
```
danxmoran, line 3119: Noticed while removing the log line below... the `tags[i]` arg to `ParseKeyBytesWithTags` is fishy...

tsdb/engine/tsm1/engine.go
```diff
@@ -3140,3 +3098,75 @@ func varRefSliceRemove(a []influxql.VarRef, v string) []influxql.VarRef {
...
+			e.logger.Debug(
+				"Read series during reindexing",
+				zap.String("name", string(names[i])),
+				zap.String("tags", tags[i].String()),
+			)
```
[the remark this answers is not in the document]
danxmoran: Didn't see a Trace() option, so I'll move this out of the loop and have it instead debug-log the size of the batch being indexed.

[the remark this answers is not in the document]
danxmoran: Hmmm actually, looks like there is a trace logger in the engine struct. Will look into using both logs.

tsdb/shard_test.go
```diff
@@ -139,132 +129,93 @@ func TestShard_Open_CorruptFieldsIndex(t *testing.T) {
...
-			models.Tags{{Key: []byte("host"), Value: []byte(fmt.Sprintf("server%d", i))}},
-			map[string]interface{}{"value": 1.0},
-			time.Unix(1, 2),
-		)
-		points = append(points, pt)
-	}
+	validateIndex()
 
-	err := sh.WritePoints(points)
+	// and ensure that we can still write data
```
danxmoran, line 163: Before merge: also need to check that the index dir has been re-written.

storage/engine.go
```diff
@@ -138,13 +137,11 @@ func NewEngine(path string, c Config, options ...Option) *Engine {
...
-	fields := []zap.Field{}
-	fields = append(fields, zap.String("service", "storage-engine"))
-	e.logger = log.With(fields...)
+	e.logger = log.With(zap.String("service", "storage-engine"))
 
-	e.tsdbStore.Logger = e.logger
+	e.tsdbStore.WithLogger(e.logger)
 	if pw, ok := e.pointsWriter.(*coordinator.PointsWriter); ok {
-		pw.Logger = e.logger
+		pw.WithLogger(e.logger)
```
danxmoran, line 144: Drive-by fix for engine logging setup

stuartcarnie (approved):
Great work, @danxmoran – removing this will simplify future maintenance of the TSDB engine. 🥇 👏 🧹

PR #20317, opened 2020-12-11
fix(onboarding): don't ignore failures to set initial password

### LINKED ISSUE #20088: Onboarding API ignores errors when trying to set user password
Hello,

I'm using influxdb 2.0 but i coud not sign in. i get this error in logs :
error   api error encountered   {"log_id": "0QYiUwu0000", "handler": "session", "error": "unauthorized access"}

Could you please help.

Best Regards.

danxmoran:
Fixes #20088 

This was the most obvious problem with the onboarding path's handling of passwords. There's also a lot of duplication in the CLI with slight differences between the onboarding paths for interactive vs. noninteractive runs and the `influx setup` vs. `influxd upgrade` runs, but I'm leaving that refactor for a follow-up PR.

### DIFF
--- a/tenant/service_onboarding.go
+++ b/tenant/service_onboarding.go
@@ -6,12 +6,14 @@ import (
 	"github.com/influxdata/influxdb/v2"
 	icontext "github.com/influxdata/influxdb/v2/context"
 	"github.com/influxdata/influxdb/v2/kv"
+	"go.uber.org/zap"
 )
 
 type OnboardService struct {
 	service     *Service
 	authSvc     influxdb.AuthorizationService
 	alwaysAllow bool
+	log         *zap.Logger
 }
 
 type OnboardServiceOptionFn func(*OnboardService)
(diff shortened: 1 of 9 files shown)


tenant/service_onboarding.go
```diff
@@ -6,12 +6,14 @@ import (
...
 	icontext "github.com/influxdata/influxdb/v2/context"
 	"github.com/influxdata/influxdb/v2/kv"
+	"go.uber.org/zap"
 )
 
 type OnboardService struct {
 	service     *Service
 	authSvc     influxdb.AuthorizationService
 	alwaysAllow bool
+	log         *zap.Logger
```
danxmoran, line 16: It's a little odd to have a `log` parameter here when there's a whole middleware layer for logging. IMO the middlewares are more confusing than they're worth and could be replaced by option-setters like `WithOnboardingLogger` below. Holding off on the refactor for now 🙂

tenant/service_onboarding.go
```diff
@@ -98,7 +107,18 @@ func (s *OnboardService) onboardUser(ctx context.Context, req *influxdb.Onboardi
 
 	// create users password
 	if req.Password != "" {
-		s.service.SetPassword(ctx, user.ID, req.Password)
+		if err := s.service.SetPassword(ctx, user.ID, req.Password); err != nil {
```
davidby-influx, line 110: This looks like the best you can do in cascading errors.  Nice that you return the original error here, not the cleanup error. +1

davidby-influx (approved):
I really like to see this level of error handling. Nice work!

docmerlin (approved):

PR #20348, opened 2020-12-15
fix: adjust post-install logic to only display upgrade notice when needed

### LINKED ISSUE #20340: post-install script displays upgrade notice even when no V1 data is on disk
__Steps to reproduce:__

1. Run a fresh ubuntu VM
2. Download & install the 2.0.3 deb archive using instructions from the downloads page

__Expected behavior:__
No upgrade notice is shown after install.

__Actual behavior:__
An upgrade notice is shown after install. Running the upgrade helper-script results in an error, because there's no V1 config / data on disk.

danxmoran:
Closes #20340

In addition to checking for V2 data, check for a V1 config file. Display the upgrade notice when V1 is present and V2 isn't.

To test this, I:
1. Built the .deb from my branch
2. Mounted the .deb into an ubuntu docker container
3. `dpkg -i`'d the .deb, and confirmed:
    1. The upgrade notice didn't show
    2. `config.toml` was populated with default paths for bolt and engine
4. `dpkg -r influxdb2 && rm /etc/influxdb/config.toml`
5. Set up the influxdata repo, then `apt-get install -y influxdb && apt-get remove -y influxdb`
6. `dpkg -i`'d the .deb, and confirmed the upgrade notice showed

### DIFF
--- a/scripts/post-install.sh
+++ b/scripts/post-install.sh
@@ -25,6 +25,23 @@ function install_chkconfig {
     chkconfig --add influxdb
 }
 
+function should_upgrade {
+    if [[ ! -s /etc/influxdb/influxdb.conf ]]; then
+        # No V1 config present, no upgrade needed.
+        return 1
+    fi
+
+    bolt_dir="/root/.influxdbv2 /var/lib/influxdb/.influxdbv2 /var/lib/influxdb"
+    for bolt in $bolt_dir; do
+        if [[ -s ${bolt}/influxd.bolt ]]; then
+            # Found a bolt file, assume previous v2 upgrade.
+            return 1
+        fi
+    done
+
+    return 0
+}
+
 function upgrade_notice {
 cat << EOF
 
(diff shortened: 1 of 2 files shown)


scripts/post-install.sh
```diff
@@ -51,12 +68,18 @@ https://docs.influxdata.com/influxdb/v2.0/get-started/
...
+EOF
+}
+
 # Add defaults file, if it doesn't exist
 if [[ ! -s /etc/default/influxdb2 ]]; then
 cat << EOF > /etc/default/influxdb2
-INFLUXD_CONFIG_PATH=/etc/influxdb/config.toml
-INFLUXD_BOLT_PATH=/var/lib/influxdb/influxd.bolt
-INFLUXD_ENGINE_PATH=/var/lib/influxdb/engine
+INFLUXD_CONFIG_PATH=${INFLUXD_CONFIG_PATH}
```
danxmoran, line 82: `influxd upgrade` will write `bolt-path` and `engine-path` into the `config.toml` output. This change + the one above are meant to make the fresh-install end state look as similar as possible to the upgrade case.

psteinbachs (approved):
Nice improvement.

PR #20349, opened 2020-12-15
fix: ensure init.d-style install sees our default environment variables

### LINKED ISSUE #20132: [2.0.x] Debian package does not read /etc/influxdb/config.toml file
__Steps to reproduce:__

1. Install InfluxDB 2.x OSS
2. Provide /etc/influxdb/config.toml file
3. Starts it with service

__Expected behavior:__

`/etc/influxdb/config.toml` file should be read as default config file.

__Actual behavior:__

file ignored and influx will start in `/var/lib/influxdb/.influxdbv2`

__Fix suggestion__

As systemd service file contains:

```
EnvironmentFile=-/etc/default/influxdb
``` 

We could have for `/etc/default/influxdb` instead of empty file:

``` 
INFLUXD_CONFIG_PATH=/etc/influxdb/config.toml
```

danxmoran:
Addresses #20132 

I think the `set` trick will work since the shebang on the script points at `/bin/bash`, let me know if it's not portable.

Tested by building the package & installing into a fresh ubuntu VM. Without these changes, I see this in the logs:
```
ts=2020-12-15T22:52:12.224551Z lvl=info msg="Resources opened" log_id=0R69W2FG000 service=bolt path=/var/lib/influxdb/.influxdbv2/influxd.bolt
```
After this change, I see:
```
ts=2020-12-15T22:56:05.258672Z lvl=info msg="Resources opened" log_id=0R69jGY0000 service=bolt path=/var/lib/influxdb/influxd.bolt
```

With these env / config file contents:
```
$ cat /etc/default/influxdb2
INFLUXD_CONFIG_PATH=/etc/influxdb/config.toml

$ cat /etc/influxdb/config.toml
bolt-path = "/var/lib/influxdb/influxd.bolt"
engine-path = "/var/lib/influxdb/engine"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 1. [20339](https://github.com/influxdata/influxdb/pull/20339): Include upgrade helper script in goreleaser manifest.
 1. [20348](https://github.com/influxdata/influxdb/pull/20348): Don't show the upgrade notice on fresh `influxdb2` installs.
 1. [20348](https://github.com/influxdata/influxdb/pull/20348): Ensure `config.toml` is initialized on fresh `influxdb2` installs.
+1. [20349](https://github.com/influxdata/influxdb/pull/20349): Ensure `influxdb` service sees default env variables when running under `init.d`.
 
 ## v2.0.3 [2020-12-14]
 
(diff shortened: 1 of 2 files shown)


psteinbachs (approved):
I think this will be fine and shouldn't pose any portability issues.

PR #20355, opened 2020-12-16
fix: don't use a finished context for graceful shutdown

danxmoran:
Part of #19976 

Ayan noticed in a separate PR that we were misusing `ctx` values here, resulting in an immediate timeout of our graceful-shutdown code. Splitting out the fix to make it easy to review.

### DIFF
--- a/cmd/influxd/launcher/cmd.go
+++ b/cmd/influxd/launcher/cmd.go
@@ -71,32 +71,23 @@ func cmdRunE(ctx context.Context, o *InfluxdOpts) func() error {
 		fluxinit.FluxInit()
 
 		l := NewLauncher()
-		if err := l.runUntilDone(ctx, o); err != nil {
+
+		// Start the launcher and wait for it to exit on SIGINT or SIGTERM.
+		runCtx := signals.WithStandardSignals(ctx)
+		if err := l.run(runCtx, o); err != nil {
 			return err
 		}
-		l.gracefulShutdown(ctx)
+		<-runCtx.Done()
+
+		// Tear down the launcher, allowing it a few seconds to finish any
+		// in-progress requests.
+		shutdownCtx, cancel := context.WithTimeout(ctx, 2*time.Second)
+		defer cancel()
+		l.Shutdown(shutdownCtx)
 
 		return nil
 	}
-}
-
-// runUntilDone starts the launcher and waits for it to exit on SIGINT or SIGTERM.
-func (l *Launcher) runUntilDone(ctx context.Context, o *InfluxdOpts) error {
-	ctx = signals.WithStandardSignals(ctx)
-	if err := l.run(ctx, o); err != nil {
-		return err
-	}
-	<-ctx.Done()
-	return nil
-}
-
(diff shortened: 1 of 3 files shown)


cmd/influxd/launcher/cmd.go
```diff
@@ -70,24 +70,34 @@ func cmdRunE(ctx context.Context, o *InfluxdOpts) func() error {
...
-		l.Shutdown(ctx)
+		l.gracefulShutdown(ctx)
 
 		return nil
 	}
 }
 
+// runUntilDone starts the launcher and waits for it to exit on SIGINT or SIGTERM.
+func (l *Launcher) runUntilDone(ctx context.Context, o *InfluxdOpts) error {
+	ctx = signals.WithStandardSignals(ctx)
```
danxmoran, line 85: Split these into separate methods to make it harder for the run & shutdown contexts to collide with one another

cmd/influxd/launcher/cmd.go
```diff
@@ -70,24 +70,34 @@ func cmdRunE(ctx context.Context, o *InfluxdOpts) func() error {
...
-
 		l := NewLauncher()
-		if err := l.run(ctx, o); err != nil {
+		if err := l.runUntilDone(ctx, o); err != nil {
 			return err
 		}
-		<-ctx.Done()
-
-		// Attempt clean shutdown.
-		ctx, cancel := context.WithTimeout(ctx, 2*time.Second)
```
danxmoran, line 83: This is the bug:
* At this point, `ctx` is completed (because we got past the `<-ctx.Done()` call above)
* `context.WithTImeout` returns an already-completed context if the 1st arg is completed

davidby-influx: I think you need a new context here, because the parent forces the new child context closed, as well. WithTimeout() calls WithDeadline() which has the following comment:

/ WithDeadline returns a copy of the parent context with the deadline adjusted
// to be no later than d. If the parent's deadline is already earlier than d,
// WithDeadline(parent, d) is semantically equivalent to parent. The returned
// context's Done channel is closed when the deadline expires, when the returned
// cancel function is called, or when the parent context's Done channel is
// closed, whichever happens first.

davidby-influx (changes requested):

davidby-influx (approved):

cmd/influxd/launcher/cmd.go
```diff
@@ -70,19 +70,20 @@ func cmdRunE(ctx context.Context, o *InfluxdOpts) func() error {
...
 		fluxinit.FluxInit()
 
-		// exit with SIGINT and SIGTERM
-		ctx = signals.WithStandardSignals(ctx)
-
 		l := NewLauncher()
-		if err := l.run(ctx, o); err != nil {
+
+		// Start the launcher and wait for it to exit on SIGINT or SIGTERM.
+		runCtx := signals.WithStandardSignals(ctx)
```
ayang64, line 76: why assign this to `runCtx` and not simply:

```
ctx := signals.WithStandardSignals(ctx)
```


also, if you're using go 1.15, i think you can use `os/signal.NotifyContext()`.
?

PR #20362, opened 2020-12-16
fix(cli): don't overwrite stack name/desc on stack update

### LINKED ISSUE #20344: Updating an existing stack discards the name and description on every update
When updating an existing stack from the command line, the name and description are lost unless you explicitly re-supply them each time.

__Steps to reproduce:__

1. Run the command `influx stacks init -n "example" -d "An example stack"`, note the ID of the created stack.
2. Observe that the stack has a name and description. 
3. Run the command `influx task create 'option task = {name: "Example", every: 10m} from(bucket: "example")'`, note the ID of the created task.
4. Run the command `influx stacks update -i $STACK_ID --addResource=Task=$TASK_ID -f example.yaml`
5. Observe that the stack doesn't have a name or a description anymore.

__Expected behavior:__
Updating a stack without specifying the `-n` or `-d` flags should leave the existing name and description in place. Eliminating the existing name or description of a stack should require passing the `-n` or `-d` flags with the empty string as an argument, e.g. `-n ""`.

__Actual behavior:__
Failing to specify the `-n` and/or `-d` flags when updating an existing stack removes the existing name and description for that stack. To retain the existing name and description you must explicitly provide them them during every invocation of `influx stacks update`.

__Environment info:__
InfluxDB Cloud 2.0
CLI tools version:  `Influx CLI 2.0.3 (git: fe04d346df) build_date: 2020-12-15T01:00:16Z`

glinton:
Closes #20344

This preserves stack name and description if they are not altered on a stack update.

- [x] Rebased/mergeable
- [x] Tests pass


cmd/influx/template_test.go
```diff
@@ -636,6 +636,124 @@ func Test_Template_Commands(t *testing.T) {
...
 			}
 		})
+
+		t.Run("update", func(t *testing.T) {
+			tests := []struct {
+				name          string
+				args          []string
+				envVars       map[string]string
+				expectedStack pkger.Stack
+				shouldErr     bool
```
danxmoran, line 646: Doesn't look like `shouldErr` is ever set to `true`, should we remove it?

danxmoran (approved):
One question about cleanup, but LGTM overall. Can you add a line to the `CHANGELOG.md` before merging?

PR #20374, opened 2020-12-17
refactor: simplify how we set the top-level influxd logger

danxmoran:
Part of #19976 

This started as an attempt to make a tiny dent in the size of `launcher.go`, but it ended up mostly benefiting our test code. With these changes, all logging is routed into a per-test buffer, and displayed on failure. With the logger-setup path hard-coded in this way, I was able to strip some complexity out of our logic for setting up test launchers.


cmd/influxd/launcher/launcher_test.go
```diff
@@ -18,11 +18,9 @@ import (
...
 
 func TestLauncher_Setup(t *testing.T) {
-	l := launcher.NewTestLauncher(nil)
-	if err := l.Run(ctx); err != nil {
-		t.Fatal(err)
-	}
-	defer l.Shutdown(ctx)
+	l := launcher.NewTestLauncher()
+	l.RunOrFail(t, ctx)
+	defer l.ShutdownOrFail(t, ctx)
```
davidby-influx, line 23: It's not code you touched, but should Launcher.Shutdown() return an error, so that ShutdownOrFail() checks for the actual Launcher shutdown, not just the file deletion afterwards, and fails the test if the Launcher didn't shutdown?

davidby-influx (approved):
Added a suggestion for further refactoring, but approved without it.

PR #20380, opened 2020-12-18
fix: don't duplicate text in task errors

### LINKED ISSUE #20055: InfluxDB 2.0.1 OSS; duplicated error message
__Steps to reproduce:__

1. Create a threshold alert
2. Change bucket name
3. Go back to alerts page
4. Hover over red triangle with ! on it

__Expected behavior:__
I hover over the red triangle and the error message is rendered once, not twice.

__Actual behavior:__
```
Last Run Status: Error exhausting result iterator; Err: failed to initialize execute state: could not find bucket "test": failed to initialize execute state: could not find bucket “test"
```

__Environment info:__

* System info: Darwin 18.7.0 x86_64
* InfluxDB version: InfluxDB 2.0.1 (git: 37cc047133) build_date: 2020-11-11T03:53:31Z

danxmoran:
Closes #20055


docmerlin (approved):
Ok, cool, looks like the `Error()` method takes care of including the `Err` field, so we should be good.

PR #20396, opened 2020-12-21
fix(cli): don't ignore errors when marking flags as required

### LINKED ISSUE #20394: CLI helpers silently fail to mark options as required
We set the required marker before we register the corresponding flag, and don't check for an error.

danxmoran:
Closes #20394 

Replaced some `panic`s with `error` return values while I was in the neighborhood.

EDIT: Tests caught a problem in my initial approach (yay!), and I noticed some fishy behavior in the existing implementation. See inline comment.


cmd/influxd/main.go
```diff
@@ -2,7 +2,7 @@ package main
 
 import (
 	"context"
-	"fmt"
+	fmt "fmt"
```
danxmoran, line 5: Weird, TODO revert this

kit/cli/viper.go
```diff
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"fmt"
+	"github.com/spf13/cast"
```
danxmoran, line 5: TODO move to the 2nd import block

kit/cli/viper.go
```diff
@@ -137,8 +121,15 @@ func BindOptions(v *viper.Viper, cmd *cobra.Command, opts []Opt) {
 			} else {
 				flagset.StringVar(destP, o.Flag, d, o.Desc)
 			}
-			mustBindPFlag(v, o.Flag, flagset)
-			*destP = v.GetString(envVar)
+			if err := v.BindPFlag(o.Flag, flagset.Lookup(o.Flag)); err != nil {
+				return err
+			}
+			if envVal != nil {
+				if s, err := cast.ToStringE(envVal); err == nil {
```
danxmoran, line 128: This implementation (and the repetitious blocks below) ignores any errors returned when trying to convert a value pulled from config / the environment into the type expected by the CLI flags. It's a faithful translation of the previous behavior (see the impl for `v.GetString` etc), but I'm surprised it hasn't caused any user frustration. IMO it'd be better to fail fast here when an invalid value is provided in config / the environment, or at least to print a warning. Thoughts?

docmerlin: Yes, failing fast on invalid config is the right behavior.

lesam:
Should we audit some of the higher-level packages with https://github.com/kisielk/errcheck ?

kit/cli/viper.go
lesam: Silently ignoring config flags definitely sounds bad to me.

lesam:
Mostly ok with this - still questions about if we should audit with errcheck and also what the impact would be of fixing the config ignore-config-on-error behaviour

danxmoran:
Created #20438 to track running errcheck

danxmoran:
Update: changing the code to not ignore errors has opened up pandora's box of cobra/viper problems we were apparently hitting but never noticing... need to explore options for fixing the problems w/o rewriting the universe

lesam:
`BindOptions` used to panic but now it returns an error, which is not always checked.

danxmoran:
Going to revert the `panic` changes for now, so we can merge the fix for setting `Required` and not worry about unhandled `err`s elsewhere

danxmoran:
Closing this for now, will reopen a more focused fix for just the required markers, with follow-ups for the other changes.

PR #20397, opened 2020-12-21
build: clean up our CircleCI configs

### LINKED ISSUE #20366: Clean up our caching & gotest logic in CI
Some things to investigate:
1. Our `GOCACHE` is ~3.5 GiB, and sometimes takes longer to download/restore than the rest of the job combined. Can we shrink it?
2. After #20358 we'll be restoring competing GOCACHEs from multiple test suites. Do we need to? If we're considered about caching test results, how can we prevent collisions between test suites?
3. Our `gotest` job runs tests twice, once "with cache" and once in parallel without. Does the 2nd option really not use the cache? If not, can we just run tests once?

danxmoran:
Closes #20366 

* Delete unused configs
* Delete unused test steps
* Reorder our main config file to start with top-level pipelines, and to group conceptually-related jobs
* Introduce new jobs to pull dependencies, to reduce duplicated work across test/lint/build steps
* Rename our caches to bust our current huge sizes
* Use unique names for caches per-step, to avoid conflicts
* Fix result collection for JS tests


.circleci/config.yml
```diff
@@ -22,7 +22,6 @@ jobs:
           destination: raw-daily-output
       - store_test_results:
           path: ~/project
-          destination: daily-junit
```
danxmoran, line 25: Invalid key for `store_test_results` according to IntelliJ's validator

.circleci/config.yml
```diff
@@ -66,13 +64,12 @@ jobs:
...
       - store_test_results:
           path: ~/project
-          destination: nightly-junit
+
   e2e:
     docker:
       - image: quay.io/influxdb/influxdb-circleci:latest
     environment:
       GOCACHE: /tmp/go-cache
-      GOFLAGS: "-mod=readonly -p=4" # Go on Circle thinks 32 CPUs are available, but there aren't.
```
danxmoran, line 75: Bad copy-pasta, Go only gets confused about available CPUs in jobs that use `parallelism` > 1

.circleci/config.yml
```diff
@@ -93,52 +90,13 @@ jobs:
...
           path: ui/junit-results
-          destination: junit-results
       - store_artifacts:
           path: ui/cypress/videos
           destination: videos
       - store_artifacts:
           path: ui/cypress/screenshots
           destination: screenshots
 
-  selenium_accept:
```
danxmoran, line 104: Unused for who knows how long

.circleci/config.yml
```diff
@@ -246,32 +204,31 @@ jobs:
       - restore_cache:
           name: Restoring GOCACHE
           keys:
-            - influxdb-gocache-{{ .Branch }}-{{ .Revision }} # Matches when retrying a single run.
-            - influxdb-gocache-{{ .Branch }}- # Matches a new commit on an existing branch.
-            - influxdb-gocache- # Matches a new branch.
+            - influxdb-gotest-{{ .Branch }}-{{ .Revision }} # Matches when retrying a single run.
+            - influxdb-gotest-{{ .Branch }}- # Matches a new commit on an existing branch.
+            - influxdb-gotest- # Matches a new branch.
```
danxmoran, line 209: Trying to:
1. Bust the caches that are currently in Circle, to see if they rebuild smaller & quicker to download
2. Separate the test caches for different jobs to avoid collisions

.circleci/config.yml
```diff
@@ -246,32 +204,31 @@ jobs:
...
       - run: mkdir -p $TEST_RESULTS
-      - run: make test-go # This uses the test cache so it may succeed or fail quickly.
       - run:
-          name: parallel go race tests
+          name: run parallel race tests
           command: |
-            set +e
+            GO_TEST_CMD="gotestsum --format standard-quiet --junitfile /tmp/test-results/gotestsum.xml -- -p=4"
             TESTFILES=$(go list ./... | circleci tests split --split-by=timings)
-            echo $TESTFILES
```
danxmoran, line 264: The `make` command below should echo the commands it executes, including the value of `TESTFILES`

docmerlin (approved):

PR #20414, opened 2020-12-23
build(ui): fix report generation & collection for JS tests

danxmoran:
Noticed while working on flaky tests. Will comment explanations inline.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -244,7 +244,7 @@ jobs:
             # This should usually be a no-op (fully contained in the Yarn cache above), but we
             # include it to be safe since `yarn test` won't auto-install missing modules.
             yarn install --immutable
-            yarn test:ci
+            yarn test:circleci
       - store_test_results:
           path: ui/coverage
       - store_artifacts:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -229,7 +229,6 @@ jobs:
     docker:
       - image: quay.io/influxdb/influxdb-circleci:latest
     working_directory: /go/src/github.com/influxdata/influxdb
-    parallelism: 8
     steps:
       - checkout
       - restore_cache:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -229,7 +229,6 @@ jobs:
     docker:
       - image: quay.io/influxdb/influxdb-circleci:latest
     working_directory: /go/src/github.com/influxdata/influxdb
-    parallelism: 8
```
danxmoran, line 232: Our JS tests are so quick, there's not much point to parallelizing across many VMs

.circleci/config.yml
```diff
@@ -245,8 +244,7 @@ jobs:
             # This should usually be a no-op (fully contained in the Yarn cache above), but we
             # include it to be safe since `yarn test` won't auto-install missing modules.
             yarn install --immutable
-            TESTFILES=$(circleci tests glob "src/**/*.test.ts*" | circleci tests split --split-by=timings)
-            yarn test:circleci $TESTFILES
+            yarn test:ci
```
danxmoran, line 247: `yarn test:circleci` called into `yarn test:ci` with a limit on `maxWorkers`. Now that we have no parallelism, there's no need for the limit.

ui/jest.config.js
```diff
@@ -33,5 +33,5 @@ module.exports = {
   ],
   coverageDirectory: './coverage',
   coverageReporters: ['html', 'cobertura'],
-  reporters: ['default', 'jest-junit'],
+  reporters: ['default', ['jest-junit', { classNameTemplate: '{filepath}' }]],
```
danxmoran, line 36: Before this change, the JUnit reports generated by `jest` used the test name (free text full of spaces) as the class name. This isn't a perfect replacement, but it's enough to make Circle happy.

davidby-influx: We are assuming we're careful not to have spaces in our filepath, which seems reasonable. Only savages do that.

davidby-influx (approved):
Looks good.

davidby-influx (approved):

PR #20440, opened 2021-01-04
fix(cmd/influxd): improve `influxd upgrade` logging, and require manual confirmation of data copy

### LINKED ISSUE #20383: Make size-check in influxd upgrade more obvious
Before copying any data, `influxd upgrade` checks that the target v2 directory has enough space to copy all v1 files, and exists with an error if not. The check is otherwise silent unless the command is running in verbose mode. We can make the check more obvious by:
1. Removing the verbosity check around the log statement (it's only 1 line)
2. Unless `-f` is given, add a y/n interactive question before proceeding with the copy

We should also call out the copying behavior in our packaging helper scripts.

danxmoran:
Closes #20383

Before this patch, we were:
1. Calculating expected disk size needed to make a full copy of V1 data
2. Logging the info, but only in `verbose` mode
3. Auto-proceeding with the copy if the disk had enough space

After this patch:
1. `verbose` is deprecated/hidden in favor of debug-level logging (reduce the number of levers users need to pull to get useful diagnostics)
2. Most verbose-level logs are now debug, but the size-log is info
3. Add an explicit y/n input point before copying data (skippable by using `--force`)

### DIFF
--- a/cmd/influxd/upgrade/database.go
+++ b/cmd/influxd/upgrade/database.go
@@ -2,20 +2,25 @@ package upgrade
 
 import (
 	"context"
+	"errors"
 	"fmt"
-	"github.com/dustin/go-humanize"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/dustin/go-humanize"
 	"github.com/influxdata/influxdb/v2"
+	"github.com/influxdata/influxdb/v2/cmd/internal"
 	"github.com/influxdata/influxdb/v2/pkg/fs"
 	"github.com/influxdata/influxdb/v2/v1/services/meta"
+	"github.com/tcnksm/go-input"
 	"go.uber.org/zap"
 )
 
 // upgradeDatabases creates databases, buckets, retention policies and shard info according to 1.x meta and copies data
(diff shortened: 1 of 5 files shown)


cmd/influxd/upgrade/database.go
```diff
@@ -222,3 +193,45 @@ func upgradeDatabases(ctx context.Context, v1 *influxDBv1, v2 *influxDBv2, v1opt
...
+
+	if size > diskInfo.Free {
+		return fmt.Errorf("not enough space on target disk of %s: need %d, available %d", v2dir, size, diskInfo.Free)
+	}
+	if !opts.force {
+		if confirmed := internal.GetConfirm(ui, func() string {
+			return fmt.Sprintf(`Proceeding will copy all V1 data to %q
+  Space available: %s
+  Space required:  %s
+`, v2dir, freeBytes, requiredBytes)
```
danxmoran, line 231: @psteinbachs @timhallinflux can you give a :+1: or :-1: to this behavior change? It adds one more y/N interaction point to `influxd upgrade`, to force users to consider how much disk they're about to use. The check is skipped if the `--force` flag is given.

psteinbachs: Yes, given community feedback around this I think it is a welcome change.

davidby-influx (approved):
Looks good.

PR #20442, opened 2021-01-04
fix(http): don't return 500 errors for partial writes

### LINKED ISSUE #20359: 500 error on writing data out of policy range
Writing data outside of policy range should result in a 4xx error that the client can understand not to retry sending the data. It's expected behavior that 5xx errors received by clients should result in the client re-transmitting the same data to avoid data loss.

This issue was raised by a customer here https://github.com/influxdata/telegraf/issues/8571

__Steps to reproduce:__

I've configured a bucket with a retention policy of 2 weeks. When I write some measurements that are older than those two weeks, influxdb response with an 500 error. 

1. setup influxdb v2 
2. create bucket with a retention policy of N weeks
3. send some data to this bucket that is older than N weeks. E.g. via curl.

__Expected behavior:__

server responds with an error indicating client error, eg 4xx. such as. `406 Not Acceptable`  or `422 Unprocessable Entity`

__Actual behavior:__

Server responds with 500 Internal Server Error

__Environment info:__

* InfluxDB version: 2.x

__Logs:__
see logs from https://github.com/influxdata/telegraf/issues/8571

danxmoran:
Closes #20359

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Replacement `tsi1` indexes will be automatically generated on startup for shards
 1. [20380](https://github.com/influxdata/influxdb/pull/20380): Remove duplication from task error messages.
 1. [20313](https://github.com/influxdata/influxdb/pull/20313): Automatically build `tsi1` indexes for shards that need it instead of falling back to `inmem`.
 1. [20313](https://github.com/influxdata/influxdb/pull/20313): Fix logging initialization for storage engine.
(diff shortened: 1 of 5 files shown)


davidby-influx (approved):
Looks good to me.  Does this require a backport to 1.8?

danxmoran:
Looks like it's [already covered](https://github.com/influxdata/influxdb/blob/9e33be2619cd0d7ad5a4323455bf5c3308908f23/services/httpd/handler.go#L1010-L1014) in 1.x, though it's a different flavor of 4xx.

PR #20451, opened 2021-01-05
build: fix cache key used by godeps job

danxmoran:
_facepalms_

This fixes the key of the `godeps` job to match the key expected by all downstream jobs. Important because `godeps` is the step that generates the cache for everything else!


davidby-influx (approved):
The file now has go.sum exclusively as the source for the key.

PR #20473, opened 2021-01-07
feat(cmd/influxd): add --overwrite-existing-v2 flag to influxd upgrade

### LINKED ISSUE #20441: "force" mode in influxd upgrade should remove any conflicting files
By default, `influxd upgrade` will validate that there are no files present at its output paths (failing fast if proceeding would overwrite an existing file). This can be annoying for automated upgrade runs.

Upgrade already has a `--force` flag. It feels natural to think that it should imply any existing output files are overwritten.

danxmoran:
Closes #20441

### DIFF
--- a/cmd/influxd/upgrade/upgrade.go
+++ b/cmd/influxd/upgrade/upgrade.go
@@ -373,8 +373,14 @@ func runUpgradeE(cmd *cobra.Command, options *options, verbose bool) error {
 		options.source.dbURL = (&configV1{}).dbURL()
 	}
 
-	err = validatePaths(&options.source, &options.target)
-	if err != nil {
+	if err := options.source.validatePaths(); err != nil {
+		return err
+	}
+	checkV2paths := options.target.validatePaths
+	if options.force {
+		checkV2paths = options.target.clearPaths
+	}
+	if err := checkV2paths(); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/influxd/upgrade/upgrade.go
```diff
@@ -478,54 +484,85 @@ func runUpgradeE(cmd *cobra.Command, options *options, verbose bool) error {
...
+
+	if _, err := os.Stat(o.cqPath); err == nil {
+		return fmt.Errorf("file present at target path for exported continuous queries '%s'", o.cqPath)
+	}
+
+	return nil
+}
+
+// clearPaths deletes any files already present at the specified V2 output paths.
+func (o *optionsV2) clearPaths() error {
```
lesam, line 545: Seems like bad UX to have `--force` with default arguments remove all the user's data. Maybe a new arg for `--remove-existing-v2` and a helpful error message if the user runs into this situation (`Existing v2 data detected at <blah>, use --remove-existing-v2 to force migration`)

cmd/influxd/upgrade/upgrade.go
```diff
@@ -478,54 +484,85 @@ func runUpgradeE(cmd *cobra.Command, options *options, verbose bool) error {
...
 	}
 
-	if _, err = os.Stat(targetOpts.boltPath); err == nil {
-		return fmt.Errorf("file present at target path for upgraded 2.x bolt DB: '%s'", targetOpts.boltPath)
+	if _, err := os.Stat(o.boltPath); err == nil {
+		return fmt.Errorf("file present at target path for upgraded 2.x bolt DB: '%s'", o.boltPath)
 	}
 
-	if fi, err := os.Stat(targetOpts.enginePath); err == nil {
+	if fi, err := os.Stat(o.enginePath); err == nil {
```
davidby-influx, line 520: Do we want an `else` clause here for if the` os.stat()` can't be reached.  I see two cases, an err if the directory doesn't exist, which would be ok, and an err if there is a permissions problem or missing mount or filesystem error, which isn't ok.  Maybe these error cases are handled later in the directory creation process.

davidby-influx (changes requested):

cmd/influxd/upgrade/upgrade.go
lesam: I don't think it's worth it - for one thing, it's going to be very hard to tell which type of error we hit:

https://golang.org/pkg/os/#Stat

> Stat returns a FileInfo describing the named file. If there is an error, it will be of type *PathError.

Different OS's are going to handle mount failures etc. different. If you're trying to upgrade at an existing path while the drive is mounting you probably deserve problems.

davidby-influx: I'm thinking we catch the Stat() error, and then just call os.IsNotExist(err error) bool. If it's true, we are happy, if it is false, we report the error.

danxmoran: There is an `os.IsExist(err error)` function, I could use it instead of `err == nil` in all of these conditional checks.

danxmoran: Scratch that, looks like `os.IsExist` is checking for a specific error and not a lazy shorthand for `!os.IsNotExist` like I was assuming.

danxmoran: @davidby-influx sanity-checking: your comment isn't really specific to the engine directory, right? We could say the same for all the other `Stat` calls in this function.

davidby-influx: And then return or handle the error if os.IsExist() == false.  That would let you catch the edge cases where you get an error you don't expect and want to surface it.

davidby-influx: Yes, my point (and perhaps we are spending too much time on this small point) is that we want to distinguish the case the code handles, the presence of a file where there should not be one), from the case where we get some file system error trying to check of that existence.

davidby-influx (approved):
Perfect!

PR #20490, opened 2021-01-11
fix(kit/cli): don't ignore failures to mark CLI options as required

### LINKED ISSUE #20394: CLI helpers silently fail to mark options as required

danxmoran:
Closes #20394 

We can only mark a flag as required after it's been registered in the `cobra.Command`.


kit/cli/viper.go
```diff
@@ -272,9 +299,13 @@ func BindOptions(v *viper.Viper, cmd *cobra.Command, opts []Opt) {
 			} else {
 				IDVar(flagset, destP, o.Flag, d, o.Desc)
 			}
-			if s := v.GetString(envVar); s != "" {
-				_ = (*destP).DecodeFromString(v.GetString(envVar))
+			mustBindPFlag(v, o.Flag, flagset)
```
lesam, line 302: Why are we adding mustBindPFlag here (what does this change?)

danxmoran: This makes flags of these types get bound to viper, same as all the others. Right now flags of type `influxdb.ID` and `zapcore.Level` are being left out for no good reason.

I can revert this addition though, because one of the follow-up PRs is going to change the logic across the board for when flags are bound to viper.

lesam (approved):
lgtm

PR #20484, opened 2021-01-09
build: rewrite our CI/CD pipeline to make ARM64 docker build work

danxmoran:
Closes #20479 

Oh boy.

The chain of reason for these changes is:
1. CircleCI recommends against running Docker cross-builds from within a `docker` executor, because of emulation headaches
2. Using a `machine` executor instead of a `docker` executor means we have to install all the system dependencies at CI time instead of as a pre-step in the executor image
3. The original reason for creating a custom CI image was to package all of our cross-building tools in one place. If we're not doing that any more, we can revert all jobs to use the standard Circle images and speed up environment setup
4. Remove cross-building & release commands from the `Makefile` since they require significant setup and we don't intend for them to run outside of CI

Semi-related: The `master` branch had both `.goreleaser.yml` and `.goreleaser-nightly.yml`, but only ever used the nightly variant. I ported the relevant differences from `-nightly` to the main goreleaser config, and deleted nightly. When I backport these changes to `2.0` I'll just delete the `-nightly` file.

[Here](https://app.circleci.com/pipelines/github/influxdata/influxdb/17750/workflows/b820be82-40c8-4d05-803f-9ea79ea14415/jobs/166504) is a successful cross-build run with these changes.

PR #20499, opened 2021-01-12
build: more fixes for CircleCI cross-build setup

### LINKED ISSUE #20479: ARM64 docker build failing in CircleCI because of missing emulators
See an example failure [here](https://app.circleci.com/pipelines/github/influxdata/influxdb/17711/workflows/71eb4a41-6200-452e-9359-311eaa1295aa/jobs/166276). docker/buildx#464 explains the issue, and [this tutorial](https://circleci.com/blog/building-docker-images-for-multiple-os-architectures/) from Circle shows an example of the fix. The tutorial strongly recommends using a `machine` executor for cross-builds, though... we might need to reconsider our CI setup.

danxmoran:
Closes #20479 

Even with #20484 I was seeing intermittent failures on `master`. I haven't seen a failure (yet) with these changes, so fingers crossed...

* Upgrade the Docker installation on the executor instead of bolting the `buildx` plugin onto whatever happens to be there
* Install all binfmt platforms instead of just `arm64`
* Don't hard-code specific platforms when creating the `buildx` builder
* Don't share as many caches between jobs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -163,6 +163,7 @@ commands:
             sudo apt-get install -y --no-install-recommends \
               cmake \
               gnupg \
+              jq \
               libssl-dev \
               libxml2-dev \
               llvm-dev \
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -28,6 +28,10 @@ workflows:
           requires:
             - godeps
             - jsdeps
+      - cross_build:
+          requires:
+            - godeps
+            - jsdeps
```
danxmoran, line 34: I am torn between running this on every PR (catches problems early, but slow) and only running it on `master`. Thoughts?

danxmoran:
Ugh, spoke too soon. Failed job popped up [here](https://app.circleci.com/pipelines/github/influxdata/influxdb/17884/workflows/f0c6b11c-7b61-4d45-94b7-85554aef1eed/jobs/167408).

.circleci/config.yml
psteinbachs: Better I think to weed out problems early.

danxmoran:
We might be intermittently running into docker/buildx#495

danxmoran:
Closing this for now to reduce spam

Issue #20500, opened 2021-01-12
fields.idx frequent writes cause lock contention


davidby-influx:
The fields.idx file is completely rewritten every time a new measurement or field is added.  The locking around this rewrite is causing slow writes at a customer who is creating 5M new fields.

__Steps to reproduce:__
1. Add a new field or a new measurement to a database
2. Note that fields.idx.tmp is written, and renamed to fields.idx each time 
3. During the creation of fields.idx.tmp, a lock is help preventing any new fields or measurements from being added.

__Expected behavior:__
The locking for updating fields.idx should be the minimum necessary. 

__Actual behavior:__
Writes that create new fields and measurements are blocked while fields.idx is replaced.
__Environment info:__

* InfluxDB version: v1.8.3 and earlier

__Performance:__
[block.txt](https://github.com/influxdata/influxdb/files/5803906/block.txt)
[mutex.txt](https://github.com/influxdata/influxdb/files/5803907/mutex.txt)

PR #20504, opened 2021-01-12
fix(tsdb): minimize lock contention when adding new fields or measurements

davidby-influx:
fields.idx frequent writes cause lock contention and fields.idx is recreated
when a field or measurement is added in a WritePointsWithContext()

Closes https://github.com/influxdata/influxdb/issues/20500

This eliminates locking during the actual file rewrite, and limits it to
the times when the MeasurementFieldSet is actually being read or written

- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/tsdb/shard.go
+++ b/tsdb/shard.go
@@ -12,6 +12,7 @@ import (
 	"regexp"
 	"runtime"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -760,11 +761,7 @@ func (s *Shard) createFieldsAndMeasurements(fieldsToCreate []*FieldCreate) error
 		s.index.SetFieldName(f.Measurement, f.Field.Name)
 	}
 
-	if len(fieldsToCreate) > 0 {
-		return engine.MeasurementFieldSet().Save()
-	}
-
-	return nil
+	return engine.MeasurementFieldSet().Save()
 }
 
 // DeleteSeriesRange deletes all values from for seriesKeys between min and max (inclusive)
@@ -1647,16 +1644,20 @@ func (m *MeasurementFields) ForEachField(fn func(name string, typ influxql.DataT
 type MeasurementFieldSet struct {
 	mu     sync.RWMutex
 	fields map[string]*MeasurementFields
-
 	// path is the location to persist field sets
 	path string
+	// ephemeral counters for updating the file on disk
+	memoryVersion  uint64
+	writtenVersion uint64
 }
 
 // NewMeasurementFieldSet returns a new instance of MeasurementFieldSet.
 func NewMeasurementFieldSet(path string) (*MeasurementFieldSet, error) {
 	fs := &MeasurementFieldSet{
-		fields: make(map[string]*MeasurementFields),
-		path:   path,
+		fields:         make(map[string]*MeasurementFields),
+		path:           path,
+		memoryVersion:  0,
+		writtenVersion: 0,
 	}
 
 	// If there is a load error, return the error and an empty set so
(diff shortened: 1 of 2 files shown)


tsdb/shard.go
```diff
@@ -760,11 +761,7 @@ func (s *Shard) createFieldsAndMeasurements(fieldsToCreate []*FieldCreate) error
 		s.index.SetFieldName(f.Measurement, f.Field.Name)
 	}
 
-	if len(fieldsToCreate) > 0 {
```
lesam, line 763: This seems like a good optimisation - ditching it seems bad.

davidby-influx: It is checked as the first statement in the procedure. This second check looks redundant to me.

tsdb/shard.go
```diff
@@ -1801,11 +1778,67 @@ func (fs *MeasurementFieldSet) saveNoLock() error {
 		return err
 	}
 
+	fs.mu.Lock()
+	defer fs.mu.Unlock()
+
+	// Check if a later modification and save of fields has superseded ours
+	// If so, we are successfully done! We were beaten by a later call
+	// to this function
+	if fs.writtenVersion > v {
```
lesam, line 1787: We should delete the temp file here too right?

lesam: nvm the remove is defer'd

tsdb/shard_test.go
```diff
@@ -1696,6 +1697,45 @@ func TestMeasurementFieldSet_InvalidFormat(t *testing.T) {
...
+		t.Fatalf("NewMeasurementFieldSet error: %v", err)
+	}
+	var wg sync.WaitGroup
+
+	wg.Add(4)
+	go testFieldMaker(t, &wg, mf, "cpu", "cvalue", iterations)
+	go testFieldMaker(t, &wg, mf, "cpu", "cvalue", iterations)
+	go testFieldMaker(t, &wg, mf, "cpu", "cvalue", iterations)
+	go testFieldMaker(t, &wg, mf, "cpu", "cvalue", iterations)
+	wg.Wait()
```
lesam, line 1723: I'd like to see some more assertions here that the final file ended up with the correct data.

tsdb/shard_test.go
```diff
@@ -1696,6 +1697,45 @@ func TestMeasurementFieldSet_InvalidFormat(t *testing.T) {
...
+	fields := mf.CreateFieldsIfNotExists([]byte(measurement))
+	for ; count > 0; count -= 1 {
+		if err := fields.CreateFieldIfNotExists([]byte("value"+strconv.FormatInt(count, 10)), influxql.Float); err != nil {
+			t.Fatalf("create field error: %v", err)
+		}
+		if err := mf.Save(); err != nil {
+			t.Fatalf("save error: %v", err)
+		}
+	}
+}
```
lesam, line 1737: Does this change improve the benchmark for this? I assume this one:

`func BenchmarkWritePoints_NewSeries_1000000_Measurement_1_TagKey_1_TagValue(b *testing.B) {`

lesam: Should also check the rest of the benchmarks in the same file to see if there are regressions I think.

davidby-influx: Will check!

davidby-influx: The benchmarks don't really cover this case.  The only one close is BenchmarkWritePoints_ExistingSeries_100K_1_1. We would need a benchmark that just created millions of measurements and fields, and wrote them without batching. The new code and the existing code in master-1.x do not complete the benchmarks, and both timeout at the 11 minute mark. There does not appear to be any regression in the benchmarks, but since they already have errors, it's hard to be sure.

lesam: I'd like some proof that this will actually improve performance - is there anything we can do to show that?

dgnorton: How difficult would it be to fix or add benchmarks to measure the change in performance for this change? If possible, it would be best to get some measurements since this is in the storage layer.

tsdb/shard_test.go
```diff
@@ -1696,6 +1696,71 @@ func TestMeasurementFieldSet_InvalidFormat(t *testing.T) {
...
+	mfs2, err := tsdb.NewMeasurementFieldSet(path)
+	if err != nil {
+		t.Fatalf("NewMeasurementFieldSet error: %v", err)
+	}
+	for i, fs := range ft {
+		mf := mfs.Fields([]byte(mt[i]))
+		mf2 := mfs2.Fields([]byte(mt[i]))
+		for _, f := range fs {
+			if mf2.Field(f) == nil {
+				t.Fatalf("Created field not found on reload:ed MeasurementFieldSet %s", f)
```
lesam, line 1741: `reload:ed` typo?

davidby-influx:
Performance on the BenchmarkWritePoints_NewSeries_1000000_Measurement_1_TagKey_1_TagValue was roughly equivalent to the previous commit in the branch. This benchmark does not have multi-threaded writes, which is the where this fix will improve performance.  
branch | mean | stdev|
--------------|-----------|------|
|master-1.x| 103.0s| 4.95|
|branch| 104.6s| 5.64|

lesam (approved):

PR #20508, opened 2021-01-12
feat(internal/cmd): add internal command for influxdb flux test harness

### LINKED ISSUE #19882: Implement flux test runner
The functionality introduced in influxdata/flux#3298 should also be introduce in influxdb.

DOD
  - Running a single command, outside of go, runs all `*_test.flux` files in flux.
  - Remove the `TestEndToEnd` test, as the functionality will all live inside the new command
  - CI is updated to add the new command, rather than `go test ...`

jsternberg:
The internal command can be used to execute flux tests using the same
internals as the `flux test` command, but it will invoke these tests
against a test influxdb instance that is launched in-process.

Closes #19882.

### DIFF
--- a/internal/cmd/fluxtest-harness-influxdb/test.go
+++ b/internal/cmd/fluxtest-harness-influxdb/test.go
@@ -1,0 +1,233 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/influxdata/flux"
+	"github.com/influxdata/flux/ast"
+	"github.com/influxdata/flux/cmd/flux/cmd"
+	"github.com/influxdata/flux/execute/table"
+	"github.com/influxdata/flux/lang"
+	"github.com/influxdata/flux/parser"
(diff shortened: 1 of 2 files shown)


internal/cmd/fluxtest-harness-influxdb/test.go
```diff
@@ -0,0 +1,233 @@
...
+	return pkg.Files[0]
+}
+
+func (t *testExecutor) Logf(s string, i ...interface{}) {
+	_, _ = fmt.Fprintf(&t.errOutput, s, i...)
+	_, _ = fmt.Fprintln(&t.errOutput)
+}
+
+func (t *testExecutor) Errorf(s string, i ...interface{}) {
+	t.Logf(s, i...)
```
rockstar, line 192: This interface, that _looks_ a lot like `testing.T` interface, is quite neat. I like it a lot, and will likely take out my hack in idpe and put this in its place. Thanks for the idea.

rockstar (approved):
Apologies for the delay; trying to get the idpe side up and going at the same time gets my wire crossed badly. Thanks for doing this.

PR #20522, opened 2021-01-15
fix(kit/cli): fix loading of .yml config files

### LINKED ISSUE #20521: INFLUXD_CONFIG_PATH ignored if pointing to a file with extension ".yml"
__Steps to reproduce:__

1. echo 'log-level: warn' >> ${HOME}/config.yml
2. INFLUXD_CONFIG_PATH=${HOME}/config.yml influxd

__Expected behavior:__
Custom config is used, and log level is set to warn.

__Actual behavior:__
Custom config is ignored, log level stays at default of info.

danxmoran:
Closes #20521

The actual fix is 1 character, but I took the chance to beef up our tests in this area.

### DIFF
--- a/kit/cli/viper.go
+++ b/kit/cli/viper.go
@@ -62,7 +62,7 @@ func NewCommand(v *viper.Viper, p *Program) *cobra.Command {
 
 	if configPath := v.GetString("CONFIG_PATH"); configPath != "" {
 		switch path.Ext(configPath) {
-		case ".json", ".toml", ".yaml", "yml":
+		case ".json", ".toml", ".yaml", ".yml":
 			v.SetConfigFile(configPath)
 		case "":
 			v.AddConfigPath(configPath)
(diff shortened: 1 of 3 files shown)


kit/cli/viper.go
```diff
@@ -62,7 +62,7 @@ func NewCommand(v *viper.Viper, p *Program) *cobra.Command {
 
 	if configPath := v.GetString("CONFIG_PATH"); configPath != "" {
 		switch path.Ext(configPath) {
-		case ".json", ".toml", ".yaml", "yml":
+		case ".json", ".toml", ".yaml", ".yml":
```
davidby-influx, line 65: Is there any chance of running on Windows where we would need case-insensitive comparison?

davidby-influx (approved):

kit/cli/viper.go
danxmoran: Case-insensitivity seems like a good idea either way, will update.

PR #20524, opened 2021-01-15
feat(cmd/influxd): add `print-config` subcommand to support automation

### LINKED ISSUE #20517: Add influxd subcommand to print config resolved from the environment
__Proposal:__
Restore a variant of 1.x's `influxd config` to make it easier for automation to find parameters like the HTTP port being used by `influxd`.

__Current behavior:__
`influxd` can pull parameters from a variety of sources:
1. A config file on disk, in JSON, YAML, or TOML
2. Per-option environment variables
3. CLI options

This flexibility is great for users, but it makes it challenging to write robust automation that boots up an `influxd` instance and calls the API.

__Desired behavior:__
A new `influxd` subcommand that:
1. Performs all the config resolution of the main `influxd` launcher, then
2. Prints the final config in a consistent format

__Alternatives considered:__
* Reimplement the config-resolution within wrapper automation scripts
* Don't support this style of automation + custom user config

__Use case:__
We want the entrypoint for our Docker Hub images to support running arbitrary setup logic.

danxmoran:
Closes #20517 

This command is meant to support our Docker entry-point in two ways.

First, in our documentation for how to customize the `influxd` config we can advise running `influxd print-config` to generate a starting template file. This mirrors the [existing docs](https://github.com/influxdata/docker-docs/blob/master/influxdb/README.md#configuration) for our 1.x image.

Second, we want the container entry-point to support running setup scripts mounted by the user, but _only_ if the DB hasn't already been set up. To do this in a robust way while also supporting custom config / env vars / CLI args, we need to be able to determine the final values of configs like the bolt path & HTTP bind address. We can use `influxd print-config` with the `--key-name` option to do this.

If the user runs something like:
```bash
docker run -e INFLUXD_OPT2=val2 influxdb:2.0 --opt1 val1
```
Then within our entrypoint script we can run:
```bash
opt1=$(influxd print-config --key-name opt1 ${@}) # Expected output: val1
opt2=$(influxd print-config --key-name opt2 ${@}) # Expected output: val2
```

### DIFF
--- a/cmd/influxd/launcher/print_config.go
+++ b/cmd/influxd/launcher/print_config.go
@@ -1,0 +1,88 @@
+package launcher
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/influxdata/influxdb/v2/kit/cli"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"gopkg.in/yaml.v3"
+)
+
+func NewInfluxdPrintConfigCommand(v *viper.Viper, influxdOpts []cli.Opt) *cobra.Command {
+
+	var keyToPrint string
+	printOpts := make([]cli.Opt, len(influxdOpts)+1)
+
+	printOpts[0] = cli.Opt{
+		DestP: &keyToPrint,
+		Flag:  "key-name",
+		Desc:  "config key name; if set, only the resolved value of that key will be printed",
+	}
+	for i, opt := range influxdOpts {
+		printOpts[i+1] = cli.Opt{
+			DestP:  opt.DestP,
+			Flag:   opt.Flag,
+			Hidden: true,
+		}
+	}
+
+	cmd := &cobra.Command{
+		Use:   "print-config",
+		Short: "Print the full influxd config resolved from the current environment",
+		Long: `
+Print config (in YAML) that the influxd server would use if run with the current flags/env vars/config file.
+
+The order of precedence for config options are as follows (1 highest, 3 lowest):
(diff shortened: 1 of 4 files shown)


cmd/influxd/launcher/print_config.go
```diff
@@ -0,0 +1,88 @@
...
+			Flag:   opt.Flag,
+			Hidden: true,
+		}
+	}
+
+	cmd := &cobra.Command{
+		Use:   "print-config",
+		Short: "Print the full influxd config resolved from the current environment",
+		Long: `
+Print config (in YAML) that the influxd server would use if run with the current flags/env vars/config file.
```
lesam, line 35: Where does the config file read happen in this code path (I'm sure it does, looking to be educated)

danxmoran: `cli.NewCommand` (called [here](https://github.com/influxdata/influxdb/pull/20524/files#diff-7002d794d2d06b59d9096a9b955d58ccb645a3c1a9c8d66fb4eb9dd554a25da8R40)) is where config & env get loaded into the `Viper` instance. `cli.BindOptions` (called [here](https://github.com/influxdata/influxdb/pull/20524/files#diff-b37fd099a57923e8dbb2276b8dbcf3ced555f3ef55e95ca3ba5de35846b783f0R64)) is where the CLI flags get bound to local vars, with any config/env values injected as defaults.

lesam (approved):
lgtm

PR #20527, opened 2021-01-16
fix(tsdb): allow backups during snapshotting, and don't leak tmp files

danxmoran:
Closes #20526 

Backports the end combination of #19869 and #20063, which I believe were the relevant PRs in 1.x.

PR #20531, opened 2021-01-18
fix: stop backup temp file leakage

alxric:
Since this loop does not exist on success, snapshot operation is
executed three times regardless, and once succesful it only cleans up
after the last run, leaving temporary files from the previous 2 runs
behind, on top of creating unnecessary load on the system.

- [ x] [Well-formatted commit messages](https://www.conventionalcommits.org/en/v1.0.0-beta.3/)
- [x ] Signed [CLA](https://influxdata.com/community/cla/) (if not already signed)

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -920,16 +920,17 @@ func (e *Engine) Free() error {
 func (e *Engine) Backup(w io.Writer, basePath string, since time.Time) error {
 	var err error
 	var path string
+L:
 	for i := 0; i < 3; i++ {
 		path, err = e.CreateSnapshot()
-		if err != nil {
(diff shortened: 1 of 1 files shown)


danxmoran:
Hi @alxric, thanks for the submission! The git conflict here is caused by the merging of #20527, which we believe fixes the temp file leakage. Can you take a look at the patch there and let me know if you think there's any difference in end behavior between the two approaches?

alxric:
> Hi @alxric, thanks for the submission! The git conflict here is caused by the merging of #20527, which we believe fixes the temp file leakage. Can you take a look at the patch there and let me know if you think there's any difference in end behavior between the two approaches?

Ah I missed that PR. End result looks like it would be the same, so closing this. Thanks!

PR #20544, opened 2021-01-19
feat(tsi): optimize series iteration

### LINKED ISSUE #20543: Improve series iteration for TSI index
__Proposal:__
Queries like `select count(_seriesKey) from /.*/` should be satisfied only from the TSI index, without reference to the tsm engine. We should iterate through the _seriesKey values without storing them all in memory.

__Current behavior:__
Currently a query like the one above would require reading all the series keys into RAM.

__Desired behavior:__
The query should iterate over the series without swapping/OOM-ing if there are too many series to fit in memory

__Alternatives considered:__
'show series cardinality' shows the total series cardinality blazingly fast, but does not show the series cardinality at the measurement level. 'show series cardinality from my measurement' is pretty slow for large numbers of series. This is a first step toward improving that type of query.

__Use case:__
Finding out cardinality by measurement to investigate series cardinality at a more granular level.

lesam:
When using queries like 'select count(_seriesKey) from bigmeasurement`, we
should iterate over the tsi structures to serve the query instead of loading
all the series into memory up front.

Closes #20543

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/tsdb/engine/tsm1/engine.go
+++ b/tsdb/engine/tsm1/engine.go
@@ -2498,6 +2498,40 @@ type indexTagSets interface {
 	TagSets(name []byte, options query.IteratorOptions) ([]*query.TagSet, error)
 }
 
+// createSeriesIterator creates an optimized series iterator if possible.
+// We exclude less-common cases for now as not worth implementing.
+func (e *Engine) createSeriesIterator(measurement string, ref *influxql.VarRef, is tsdb.IndexSet, opt query.IteratorOptions) (query.Iterator, error) {
+	// Main check to see if we are trying to create a seriesKey iterator
+	if ref == nil || ref.Val != "_seriesKey" || len(opt.Aux) != 0 {
+		return nil, nil
+	}
+	// Check some other cases that we could maybe handle, but don't
+	if len(opt.Dimensions) > 0 {
+		return nil, nil
+	}
+	if opt.SLimit != 0 || opt.SOffset != 0 {
+		return nil, nil
+	}
+	if opt.StripName {
+		return nil, nil
+	}
+	if opt.Ordered {
+		return nil, nil
+	}
+	// Actual creation of the iterator
+	seriesCursor, err := is.MeasurementSeriesKeyByExprIterator([]byte(measurement), opt.Condition, opt.Authorizer)
+	if err != nil {
+		seriesCursor.Close()
+		return nil, err
+	}
+	var seriesIterator query.Iterator
+	seriesIterator = newSeriesIterator(measurement, seriesCursor)
+	if opt.InterruptCh != nil {
+		seriesIterator = query.NewInterruptIterator(seriesIterator, opt.InterruptCh)
+	}
+	return seriesIterator, nil
+}
+
 func (e *Engine) createCallIterator(ctx context.Context, measurement string, call *influxql.Call, opt query.IteratorOptions) ([]query.Iterator, error) {
 	ref, _ := call.Args[0].(*influxql.VarRef)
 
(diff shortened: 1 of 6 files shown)


tsdb/engine/tsm1/engine.go
```diff
@@ -2507,6 +2546,28 @@ func (e *Engine) createCallIterator(ctx context.Context, measurement string, cal
 		return nil, nil
 	}
 
+	// check for optimized series iteration for tsi index
+	if e.index.Type() == tsdb.TSI1IndexName {
+		indexSet := tsdb.IndexSet{Indexes: []tsdb.Index{e.index}, SeriesFile: e.sfile}
+		seriesOpt := opt
+		if len(opt.Dimensions) == 0 {
+			// no point ordering the series if we are just aggregating them
+			seriesOpt.Ordered = false
```
lesam, line 2555: Not sure on this particular check - maybe needs to be more granular depending on the type of call iterator?

tsdb/engine/tsm1/engine.go
```diff
@@ -2588,6 +2649,18 @@ func (e *Engine) createVarRefIterator(ctx context.Context, measurement string, o
 		return nil, nil
 	}
 
+	// check for optimized series iteration for tsi index
+	if e.index.Type() == tsdb.TSI1IndexName {
```
lesam, line 2653: This has a unit test but I'm not sure there's actually a way to hit it from an influxql query without `Ordered=true`, so might need to remove this section.

lesam:
Note we also have the `_series` System Iterator for `show series` queries - this also requires everything to be slurped into memory per-measurement.

lesam:
Now with gofmt run

lesam:
We're going to do this in flux instead of influxql, so this can be closed.

lesam:
@benbjohnson We're back to planning on doing this in influxql, so this PR is back on the menu.

tsdb/engine/tsm1/engine.go
benbjohnson: Yeah, I think there might be some call iterators that need sorting.

benbjohnson (dismissed):
Overall I think the code makes sense. I agree that disabling sorting when there is a group by may not always be safe. Otherwise 👍

tsdb/engine/tsm1/engine.go
```diff
@@ -2498,6 +2498,40 @@ type indexTagSets interface {
...
+	if opt.StripName {
+		return nil, nil
+	}
+	if opt.Ordered {
+		return nil, nil
+	}
+	// Actual creation of the iterator
+	seriesCursor, err := is.MeasurementSeriesKeyByExprIterator([]byte(measurement), opt.Condition, opt.Authorizer)
+	if err != nil {
+		seriesCursor.Close()
```
davidby-influx, line 2524: Can the` Close()` cause problems? Are there errors where the `seriesCursor` is nil, or already closed or in any other state where `Close()` would fail or panic? We have the problem elsewhere.

tsdb/engine/tsm1/engine.go
```diff
@@ -2507,6 +2541,28 @@ func (e *Engine) createCallIterator(ctx context.Context, measurement string, cal
...
+	if e.index.Type() == tsdb.TSI1IndexName {
+		indexSet := tsdb.IndexSet{Indexes: []tsdb.Index{e.index}, SeriesFile: e.sfile}
+		seriesOpt := opt
+		if len(opt.Dimensions) == 0 && call.Name == "count" {
+			// no point ordering the series if we are just counting all of them
+			seriesOpt.Ordered = false
+		}
+		seriesIterator, err := e.createSeriesIterator(measurement, ref, indexSet, seriesOpt)
+		if err != nil {
+			return nil, err
```
davidby-influx, line 2554: The reverse of my other comment above: should we call `Close()` on the `seriesIterator` here? My assumption is not to, because I expect when an error is returned from a creation function, the created object is nil or closed. But since we call `Close()` above in a series cursor, just want to ensure that this is different.

tsdb/engine/tsm1/engine.go
```diff
@@ -2507,6 +2541,28 @@ func (e *Engine) createCallIterator(ctx context.Context, measurement string, cal
...
+		}
+		if seriesIterator != nil {
+			callIterator, err := query.NewCallIterator(seriesIterator, opt)
+			if err != nil {
+				seriesIterator.Close()
```
davidby-influx, line 2559: Again, same comment: should `Close()` be called on the `callIterator`?  Assuming not,.

davidby-influx (changes requested):
May not require changes, but just a verification of the questions I ask below: should a cursor that failed to open be closed, and should iterators that failed to open not be closed.

tsdb/engine/tsm1/engine.go
lesam: Looking at the implementation for MeasurementSeriesKeyByExprIterator, the contract is it either returns `nil, error` or `cursor, nil`, never both.

tsdb/engine/tsm1/engine.go
lesam: `seriesIterator` is always nil on errors

tsdb/engine/tsm1/engine.go
lesam: Yep, `callIterator` is always nil on errors

davidby-influx (approved):
Thanks for assuaging my doubts.

PR #20561, opened 2021-01-20
feat(cmd/influxd): add `nats-port` config option to influxd server

### LINKED ISSUE #16442: Allow NATS port to be configurable
I'm using a separate NATS instance for development which a bunch of services talk to. Influx will automatically start NATS up on a different port if it's already in use, but this creates an orchestration issue where if Influx starts before my own NATS service I end up with a port collision.

Please allow specifying the NATS port to start with via a CLI flag.

danxmoran:
Closes #16442

Relevant git archeology:
* `nats.NewDefaultServerOptions()` returns a struct with `Port` set to `4222`
* Unit tests that used the `influxd` launcher sometimes failed with port conflicts because of NATS
* 047afcfc10fec049e09b4abd396b9cd104203886 added a hacky loop that would try to find an unused port closest to `4222`
* fab99c9512f86bd013fbd0c933865bbf17e9bfe7 replaced the loop with a built-in marker for random port selection

Random isn't very good for running in Docker, so we add a config parameter to support setting a specific value. I left the default as random just in case.


davidby-influx (approved):
Per our discussion, looks good.

PR #20565, opened 2021-01-20
fix(influxql): set correct Content-Type on v1 query responses

### LINKED ISSUE #20514: Wrong Content-Type in v1 query API response
__Steps to reproduce:__

1. Setup default DBRP for bucket
2. Run "SHOW MEASUREMENTS" query or any SELECT statement

For example:
```sh
curl -i --request POST 'http://localhost:8086/query?db=mybucket' --header 'Authorization: Token <token>' --header 'Accept: application/json' --data-urlencode 'q=SHOW MEASUREMENTS'
```
gives the following response from InfluxDB 2.0.3:
```
HTTP/1.1 200 OK
Date: Thu, 14 Jan 2021 09:51:36 GMT
Content-Type: text/plain; charset=utf-8
Content-Length: 120
Connection: keep-alive

{"results":[{"statement_id":0,"series":[{"name":"measurements","columns":["name"],"values":[["mymeasurement"]]}]}]}
```

__Expected behavior:__
The Content-Type header should indicate **application/json** encoding.

The same query towards InfluxDB 1.x correctly sets **application/json** content type.

__Actual behavior:__
The Content-Type header is **text/plain** while body is JSON.

__Environment info:__

* InfluxDB 2.0.3 (git: fe04d346df) build_date: 2020-12-15T01:00:16Z

danxmoran:
Closes #20514 

See reasoning for this approach in the issue.

### DIFF
--- a/http/legacy/influxqld_handler.go
+++ b/http/legacy/influxqld_handler.go
@@ -142,7 +142,13 @@ func (h *InfluxqlHandler) handleInfluxqldQuery(w http.ResponseWriter, r *http.Re
 		}
 	}
 
-	encodingFormat := influxql.EncodingFormatFromMimeType(r.Header.Get("Accept"))
+	formatString := r.Header.Get("Accept")
+	encodingFormat := influxql.EncodingFormatFromMimeType(formatString)
+	if encodingFormat == influxql.EncodingFormatUnknown {
+		h.Logger.Warn("request included unknown format in Accept header, using application/json", zap.String("format", formatString))
+		encodingFormat = influxql.EncodingFormatJSON
+	}
+	w.Header().Set("Content-Type", encodingFormat.ContentType())
 
 	req := &influxql.QueryRequest{
 		DB:             r.FormValue("db"),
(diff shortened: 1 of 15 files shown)


http/legacy/influxqld_handler.go
```diff
@@ -142,11 +142,19 @@ func (h *InfluxqlHandler) handleInfluxqldQuery(w http.ResponseWriter, r *http.Re
 		}
 	}
 
+	formatString := r.Header.Get("Accept")
+	encodingFormat := influxql.EncodingFormatFromMimeType(formatString)
+	if encodingFormat == influxql.EncodingFormatUnknown {
+		h.Logger.Warn("request included unknown format in Accept header, using application/json", zap.String("format", formatString))
```
stuartcarnie, line 148: Not sure there is value in logging invalid `Accept` headers. It is a client issue and could result in spamming the logs.

http/swaggerV1Compat.yml
```diff
@@ -114,6 +114,17 @@ paths:
...
+          name: Accept
+          schema:
+            type: string
+            description: Specifies how query results should be encoded in the response.
+            default: application/json
+            enum:
+              - application/json
+              - application/csv
+              - text/csv
+              - application/x-msgpack
```
stuartcarnie, line 127: 👏

stuartcarnie (approved):
Looks good, @danxmoran – I'd caution against logging the invalid accept headers to the logs, as these represent a client issue rather than a server issue. You could consider logging them as Debug level, which should be excluded from regular logs, but they are not a warning that an administrator needs to worry about.

http/legacy/influxqld_handler.go
danxmoran: :+1: I'll remove it

PR #20572, opened 2021-01-22
fix(kv): log migration events at debug level to avoid spam

### LINKED ISSUE #20569: Review & clean up logging for metadata migrations
On startup, the `influxd` server runs migrations against its metadata DB. It currently logs an `info` line per migration. At this point, we have enough migrations for the logging scheme to be spammy & potentially confusing to users who expect a stream of output to mean something's gone wrong.

After a Slack discussion, we decided to switch these logs to `debug` level, and add some higher-level `info` logs.

danxmoran:
Closes #20569

Before:
Screen Shot 2021-01-22 at 7 34 23 AM

After:
Screen Shot 2021-01-22 at 7 35 21 AM

### DIFF
--- a/kv/migration/migration.go
+++ b/kv/migration/migration.go
@@ -7,7 +7,7 @@ import (
 	"fmt"
 	"time"
 
-	influxdb "github.com/influxdata/influxdb/v2"
+	"github.com/influxdata/influxdb/v2"
 	"github.com/influxdata/influxdb/v2/kv"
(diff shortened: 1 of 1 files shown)


danxmoran:
@rickspencer3 does the "after" logging look better to you?

davidby-influx (approved):
This will certainly clean up the log nicely.

Issue #20459, opened 2021-01-06
math.mMin causes serious performance degradations


hardincm2:
__Steps to reproduce:__
List the minimal actions needed to reproduce the behavior.

1. Create a simple custom min function definition `math_min = (x, y) => if x < y then x else y`
2. Swap any instances of `math.mMin()` with custom function
3. Observe noticeable performance improvements (up to 50% shorter execution time)

__Expected behavior:__
Should be no significant performance impact when calculating the min

__Actual behavior:__
Significant performance impact when calculating min

__Environment info:__

* System info: Linux 4.4.0-53-generic x86_64
* InfluxDB version: 2.0.3

PR #20576, opened 2021-01-22
math.mMin causes serious performance degradations

mohammadali-seifkashani:


mohammadali-seifkashani:
In issue 20459 it was said that math.mMin had bad performance. So I removed math.mMin and math.mMax from fluxFunctions.ts file and added simple min_two_num and max_two_num functions instead. Those removed functions were not used in the project anywhere and no test was implemented for them.

mohammadali-seifkashani:
I don't know where to implement my two functions!!?

timhallinflux:
addresses #20459

wbaker85:
Since the UI code this PR refers to is no longer in the `influxdb` repository and the underlying issue has been transferred to Flux repo (see: https://github.com/influxdata/flux/issues/3508), I'm going to close this PR out since we won't be able to merge it here.

Issue #17938, opened 2020-05-02
InfluxDB v2 - E-Mail - Notification Endpoint


StefanSa:
__Proposal:__
Add E-Mail as Endpoint Notification.

__Current behavior:__
No exist this option currently

__Desired behavior:__
I want to enable notifications of alerts and send them to a E-Mail user / group.

__Alternatives considered:__
No alternatives are considered so far.

__Use case:__
This feature exists in Cronograf and I use it. That be great have in InfluxDB v2.

emexelem:
I just upgraded to InfluxDB v2 and started to use it for monitoring and sending alert, but it seems there's no support for email, except for mail services having HTTP API, as described in https://docs.influxdata.com/influxdb/v2.0/monitor-alert/send-email/
Sending email through SMTP is a must have in my case.
Do you have any idea when this will be supported?

mohammadali-seifkashani:
Whick files should be edited and what functions should be implemented?
I'm a beginner.

StefanSa:
Hi there, @hoorayimhelping 
Any ETA for release of this feature?

timhallinflux:
This is the recommended approach for now: https://docs.influxdata.com/influxdb/v2.0/monitor-alert/send-email/

mirkocomparetti-synesis:
Thank you @timhallinflux 
The actual approach does not work for us and we really miss the standard smtp functionality. Hopefully it will make it to the release soon!

yorik:
Would it be possible to call custom shell script or similar for now to send alerts? Using commercial mailing services wouldn't make sense if there is existing email server.

haavee:
Actually, we were considering to start to use InfluxDB but then I checked whether sending email via SMTP on alert was possible (a strict requirement for our environment) and I couldn't find it in v2 - much to my surprise. This alone removes Influx from our list of candidates. If a shell-script-on-alert would be an endpoint, that would work fine as well.

timhallinflux:
@haavee as mentioned here.... there are many ways to integrate with email providers to send email-based alerts.  See examples listed here: https://docs.influxdata.com/influxdb/v2.0/monitor-alert/send-email/  

You can also create your own call out using the HTTP post mechanism: https://docs.influxdata.com/influxdb/v2.0/reference/flux/stdlib/http/

jdstrand:
[...]

PR #20582, opened 2021-01-23
17938

mohammadali-seifkashani:
I don't know where and in what structure should we implement the function!

### DIFF
--- a/notification/endpoint/endpoint.go
+++ b/notification/endpoint/endpoint.go
@@ -12,6 +12,7 @@ const (
 	SlackType     = "slack"
 	PagerDutyType = "pagerduty"
 	HTTPType      = "http"
+	SMTPType	  = "smtp"
 	TelegramType  = "telegram"
 )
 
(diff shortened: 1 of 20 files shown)


notification/endpoint/endpoint.go
```diff
@@ -12,13 +12,15 @@ const (
 	SlackType     = "slack"
 	PagerDutyType = "pagerduty"
 	HTTPType      = "http"
+	SMTPType	  = "smtp"
```
nurlansu, line 15: Alignment.

```suggestion
	SMTPType      = "smtp"
```

timhallinflux:
addresses: https://github.com/influxdata/influxdb/issues/17938

mirkocomparetti-synesis:
Hopefully this will make it... it is really important for us!

notification/endpoint/smtp.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// Type returns the type.
+func (s SMTP) Type() string {
+	return SMTPType
+}
+
+func send(senderMail, senderPassword string, message []byte, to []string) {
+	// smtp server configuration.
+	smtpHost := "smtp.gmail.com"
```
yorik, line 101: It would be nice to make SMTP host and port configurable or rename this module to gmail.

PR #20586, opened 2021-01-25
fix(tsi): retaining fileset should not copy pointer

foobar:


danxmoran:
Thanks for the submission @foobar! The `gotest` failures look real (not a spurious failure) to me, I'll wait on the full review until I see they're working. My first impression is that I'm not sure what problem this is meant to solve. Could you describe the goal so I can better understand?

Issue #20579, opened 2021-01-22
Logging formats for `influxd` and `influxd upgrade` don't match


danxmoran:
`influxd` writes logs with format:
```
<timestamp> <level> <message> {"<field>": "<value">, ...}
```
`influxd upgrade` writes logs with format:
```
{"level": "<level>", "ts": "<timestamp>", "msg": "<message>", "<field>": "<value>", ...}
```

Our new DockerHub wrapper runs `upgrade` followed by plain `influxd` in the same container, and the logs get mixed together. The mismatch will be a pain for any downstream log-processing tools.

PR #20589, opened 2021-01-25
test: replace influxlogger with zaptest logger

danxmoran:
Related to #20579 

The ultimate goal is to adjust the implementation / APIs in the `logger` package for more consistency across our CLI commands. When I started making changes I noticed many tests stopped building.

I think this is a nice stand-alone win. Zaptest's logger automatically sets debug-level logging, and only shows the logs for a test if the test fail or `-v` is passed to `go test`.

### DIFF
--- a/http/router.go
+++ b/http/router.go
@@ -129,14 +129,17 @@ func panicMW(api *kithttp.API) func(http.Handler) http.Handler {
 	}
 }
 
-var panicLogger *zap.Logger
+var panicLogger = zap.NewNop()
 var panicLoggerOnce sync.Once
 
 // getPanicLogger returns a logger for panicHandler.
 func getPanicLogger() *zap.Logger {
 	panicLoggerOnce.Do(func() {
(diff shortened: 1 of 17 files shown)


http/router.go
```diff
@@ -129,14 +129,17 @@ func panicMW(api *kithttp.API) func(http.Handler) http.Handler {
 	}
 }
 
-var panicLogger *zap.Logger
+var panicLogger = zap.NewNop()
 var panicLoggerOnce sync.Once
 
 // getPanicLogger returns a logger for panicHandler.
 func getPanicLogger() *zap.Logger {
 	panicLoggerOnce.Do(func() {
```
lesam, line 137: Should this function be extracted somewhere?

danxmoran: Looked into it. These files are nearly identical and I think we could consolidate into `kit/transport/http`, but would prefer to leave it for a separate issue. Made  #20712 to track it

lesam (approved):
looks good

PR #20590, opened 2021-01-25
build: fix test-result collection for grace_nighty job

danxmoran:
The nightly grace tests have been failing, but I can't tell why because report collection is broken.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,6 +42,14 @@ workflows:
             branches:
               # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
               ignore: /pull\/[0-9]+/
+      # TODO: Remove after build fixed
+      - grace_nightly:
+          requires:
+            - build
+          filters:
+            branches:
+              # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
+              ignore: /pull\/[0-9]+/
       - litmus_daily:
           requires:
             - build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,14 +42,6 @@ workflows:
             branches:
               # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
               ignore: /pull\/[0-9]+/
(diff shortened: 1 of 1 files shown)


danxmoran:
The actual test failure from last night seems to have been a flaky thing (see the successful run [here](https://app.circleci.com/pipelines/github/influxdata/influxdb/18382/workflows/1cd5567c-1ec3-425c-ac13-dffa004b82ea/jobs/170884) from the 2nd to last commit on this branch, with no changes). I modified the `grace` repo to output reports in both the browser-friendly "awesome" format and the Circle-friendly JUnit format, so we should have an easier time debugging failures in the future.

.circleci/config.yml
```diff
@@ -777,11 +777,12 @@ jobs:
       - attach_workspace:
           at: ~/project
       - quay_login
-      - run: docker run --net host -v /var/run/docker.sock:/var/run/docker.sock -e TEST_RESULTS=~/project quay.io/influxdb/grace:latest
+      - run: mkdir -p ~/project/results
+      - run: docker run --net host -v /var/run/docker.sock:/var/run/docker.sock -v ~/project/results:/grace/test-results/grace-results -e TEST_RESULTS=~/project/results quay.io/influxdb/grace:latest
```
lesam, line 781: What does `TEST_RESULTS` do? I would have thought it would need to be the path mounted inside the docker container

danxmoran: That was my first thought as well, but it looks like it's instead used to set the source path for volumes created by `docker compose` within the image: https://github.com/influxdata/grace/search?q=TEST_RESULTS

I'm honestly not 100% sure why it works. Maybe because the `docker compose` CLI is talking to the daemon on the host, so when it passes `source` paths the paths have to be relative to the host and not the intermediary container?

lesam: ok - well, it works, and it's only build harness, so I'll approve. A mystery to solve another day.

lesam (approved):

PR #20598, opened 2021-01-25
build(flux): update flux to v0.102.0

rockstar:


timhallinflux:
should we retire this as Flux has moved on to 0.105.x

PR #20603, opened 2021-01-25
feat(query): Hyper log log operators in influxql

### LINKED ISSUE #717: Add hyperloglog functions
For people that want to calculate distinct counts across large numbers of items, we should support HyperLogLog functions. I was thinking of something like this: http://antirez.com/news/75

Basically have an aggregate that will output a string like the redis implementation. And have a function for summing them and another function for outputting the count. Like this:

``` sql
select hllHash(user_id) from events group by time(1h) into 1h.hll.events

select hllSum(hllHash) from 1h.hll.events

select hllCount(hllSum(hllHash) from 1h.hll.events where time > now() 30d

select hllCount(hllHash(user_id)) from events group by time(1d) where time > now() - 7d
```

Notice the examples where we're chaining them together.

lesam:
Closes #717

Adds HyperLogLog operator to influxql to allow computing count-distinct approximately.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/pkg/estimator/hll/hll.go
+++ b/pkg/estimator/hll/hll.go
@@ -151,9 +151,10 @@ func (h *Plus) Add(v []byte) {
 
 		if uint32(len(h.tmpSet))*100 > h.m {
 			h.mergeSparse()
-			if uint32(h.sparseList.Len()) > h.m {
-				h.toNormal()
-			}
+		}
+		if uint32(h.sparseList.Len()) > h.m {
+			h.mergeSparse()
+			h.toNormal()
 		}
 	} else {
 		i := bextr(x, 64-h.p, h.p) // {x63,...,x64-p}
@@ -241,6 +242,10 @@ func (h *Plus) MarshalBinary() (data []byte, err error) {
 		return nil, nil
 	}
 
+	if h.sparse {
+		h.mergeSparse()
+	}
+
 	// Marshal a version marker.
 	data = append(data, version)
 
@@ -251,7 +256,7 @@ func (h *Plus) MarshalBinary() (data []byte, err error) {
 		// It's using the sparse representation.
 		data = append(data, byte(1))
 
-		// Add the tmp_set
+		// Add the tmp_set (should be empty)
 		tsdata, err := h.tmpSet.MarshalBinary()
 		if err != nil {
 			return nil, err
--- a/query/compile.go
+++ b/query/compile.go
@@ -295,6 +295,8 @@ func (c *compiledField) compileExpr(expr influxql.Expr) error {
 			return c.compileElapsed(expr.Args)
 		case "integral":
 			return c.compileIntegral(expr.Args)
+		case "count_hll":
+			return c.compileCountHll(expr.Args)
 		case "holt_winters", "holt_winters_with_fit":
 			withFit := expr.Name == "holt_winters_with_fit"
 			return c.compileHoltWinters(expr.Args, withFit)
@@ -380,7 +382,7 @@ func (c *compiledField) compileFunction(expr *influxql.Call) error {
 	switch expr.Name {
 	case "max", "min", "first", "last":
 		// top/bottom are not included here since they are not typical functions.
-	case "count", "sum", "mean", "median", "mode", "stddev", "spread":
+	case "count", "sum", "mean", "median", "mode", "stddev", "spread", "sum_hll":
 		// These functions are not considered selectors.
 		c.global.OnlySelectors = false
 	default:
@@ -769,6 +771,19 @@ func (c *compiledField) compileIntegral(args []influxql.Expr) error {
 
 	// Must be a variable reference, wildcard, or regexp.
 	return c.compileSymbol("integral", args[0])
+}
+
+func (c *compiledField) compileCountHll(args []influxql.Expr) error {
+	if exp, got := 1, len(args); exp != got {
+		return fmt.Errorf("invalid number of arguments for integral, expected %d, got %d", exp, got)
+	}
+	c.global.OnlySelectors = false
+	switch arg0 := args[0].(type) {
+	case *influxql.Call:
+		return c.compileExpr(arg0)
+	default:
+		return c.compileSymbol("count_hll", arg0)
+	}
 }
 
 func (c *compiledField) compileHoltWinters(args []influxql.Expr, withFit bool) error {
--- a/query/compile.go
+++ b/query/compile.go
@@ -775,7 +775,7 @@ func (c *compiledField) compileIntegral(args []influxql.Expr) error {
 
 func (c *compiledField) compileCountHll(args []influxql.Expr) error {
 	if exp, got := 1, len(args); exp != got {
-		return fmt.Errorf("invalid number of arguments for integral, expected %d, got %d", exp, got)
+		return fmt.Errorf("invalid number of arguments for count_hll, expected %d, got %d", exp, got)
 	}
 	c.global.OnlySelectors = false
 	switch arg0 := args[0].(type) {
--- a/query/functions.gen.go.tmpl
+++ b/query/functions.gen.go.tmpl
@@ -1,9 +1,13 @@
 package query
 
 import (
+"encoding/binary"
+"bytes"
 "sort"
 "time"
 "math/rand"
+
+"github.com/influxdata/influxdb/pkg/estimator/hll"
 )
 
 {{with $types := .}}{{range $k := $types}}
@@ -16,20 +20,6 @@ type {{$k.Name}}PointAggregator interface {
 // {{$k.Name}}BulkPointAggregator aggregates multiple points at a time.
 type {{$k.Name}}BulkPointAggregator interface {
 	Aggregate{{$k.Name}}Bulk(points []{{$k.Name}}Point)
-}
-
-// Aggregate{{$k.Name}}Points feeds a slice of {{$k.Name}}Point into an
-// aggregator. If the aggregator is a {{$k.Name}}BulkPointAggregator, it will
-// use the AggregateBulk method.
-func Aggregate{{$k.Name}}Points(a {{$k.Name}}PointAggregator, points []{{$k.Name}}Point) {
-	switch a := a.(type) {
-	case {{$k.Name}}BulkPointAggregator:
-		a.Aggregate{{$k.Name}}Bulk(points)
-	default:
-		for _, p := range points {
-			a.Aggregate{{$k.Name}}(&p)
-		}
-	}
 }
 
 // {{$k.Name}}PointEmitter produces a single point from an aggregate.
@@ -109,6 +99,35 @@ func (r *{{$k.Name}}SliceFunc{{if ne $k.Name $v.Name}}{{$v.Name}}{{end}}Reducer)
 	return r.fn(r.points)
 }
 {{end}}
+
+// {{$k.Name}}SumHllReducer returns the HLL sketch for a series, in string form
+type {{$k.Name}}SumHllReducer struct {
+	plus *hll.Plus
+}
+
+// func New{{$k.Name}}SumHllReducer creates a new {{$k.Name}}SumHllReducer
+func New{{$k.Name}}SumHllReducer() *{{$k.Name}}SumHllReducer {
+	return &{{$k.Name}}SumHllReducer{plus:hll.NewDefaultPlus()}
+}
+
+// Aggregate{{$k.Name}} aggregates a point into the reducer.
+func (r *{{$k.Name}}SumHllReducer) Aggregate{{$k.Name}}(p *{{$k.Name}}Point) {
+	{{if eq $k.Type "string"}}
+	b := []byte(p.Value)
+	{{else}}
+	buf := new(bytes.Buffer)
+	binary.Write(buf, binary.BigEndian, p.Value)
+	b := buf.Bytes()
+	{{end}}
+	r.plus.Add(b)
+}
+
+// Emit emits the distinct points that have been aggregated into the reducer.
+func (r *{{$k.Name}}SumHllReducer) Emit() []StringPoint {
+	return []StringPoint{
+		marshalPlus(r.plus, nil),
+	}
+}
 
 // {{$k.Name}}DistinctReducer returns the distinct points in a series.
 type {{$k.Name}}DistinctReducer struct {
--- a/query/functions.go
+++ b/query/functions.go
@@ -2,14 +2,20 @@ package query
 
 import (
 	"container/heap"
+	"encoding/base64"
+	"fmt"
 	"math"
 	"sort"
 	"time"
 
+	"github.com/influxdata/influxdb/pkg/estimator/hll"
 	"github.com/influxdata/influxdb/query/internal/gota"
 	"github.com/influxdata/influxdb/query/neldermead"
 	"github.com/influxdata/influxql"
 )
+
+var hllPrefix = []byte("HLL_")
+var hllErrorPrefix = []byte("HLLERROR ")
 
 // FieldMapper is a FieldMapper that wraps another FieldMapper and exposes
 // the functions implemented by the query engine.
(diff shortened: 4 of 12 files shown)


pkg/estimator/hll/hll.go
```diff
@@ -241,6 +242,10 @@ func (h *Plus) MarshalBinary() (data []byte, err error) {
 		return nil, nil
 	}
 
+	if h.sparse {
+		h.mergeSparse()
+	}
```
lesam, line 247: This change is so that the serialised format of the HLL sketch is repeatable. I could have also made the `set` serialisation repeatable by sorting but this was easier.

query/functions_test.go
```diff
@@ -497,3 +501,50 @@ func TestSample_SampleSizeGreaterThanNumPoints(t *testing.T) {
...
...AAAAAAAAAAAAgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAgAAAAAAAAAAAAAAAAAAAAAAAAEAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAgAAAAAAAAAIAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAEAAAAAAAADAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABAAAAAAAAgABAAAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAwAAAAAAAAAAAAAAAAMAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAIAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAQAAAAAAAAAAAAAAAAIIAAEAAQAAAAEAAAAAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAAAAAACAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAACAAAAAAAAAAAAAAAAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAMAAAAAAAYCAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAQAAAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="
```
lesam, line 540: hah. Github refuses to render this line. Perhaps it's a sign that I should just check a hash + prefix.

query/functions.go
```diff
@@ -2150,3 +2159,90 @@ func (r *UnsignedBottomReducer) Emit() []UnsignedPoint {
...
+	_, _ = base64.StdEncoding.Decode(b, data)
+	h := new(hll.Plus)
+	if err := h.UnmarshalBinary(b); err != nil {
+		return nil, err
+	}
+	return h, nil
+}
+
+func (r *StringMergeHllReducer) AggregateString(p *StringPoint) {
+	// TODO: how to handle errors?
```
lesam, line 2190: Still need to handle this TODO

query/functions.gen.go.tmpl
```diff
@@ -1,37 +1,27 @@
...
 
 // {{$k.Name}}BulkPointAggregator aggregates multiple points at a time.
 type {{$k.Name}}BulkPointAggregator interface {
 	Aggregate{{$k.Name}}Bulk(points []{{$k.Name}}Point)
 }
 
-// Aggregate{{$k.Name}}Points feeds a slice of {{$k.Name}}Point into an
-// aggregator. If the aggregator is a {{$k.Name}}BulkPointAggregator, it will
-// use the AggregateBulk method.
-func Aggregate{{$k.Name}}Points(a {{$k.Name}}PointAggregator, points []{{$k.Name}}Point) {
```
lesam, line 24: This was apparently unused

query/functions.gen.go.tmpl
```diff
@@ -1,37 +1,27 @@
...
+
+"github.com/influxdata/influxdb/pkg/estimator/hll"
 )
 
 {{with $types := .}}{{range $k := $types}}
 
 // {{$k.Name}}PointAggregator aggregates points to produce a single point.
 type {{$k.Name}}PointAggregator interface {
-	Aggregate{{$k.Name}}(p *{{$k.Name}}Point)
+	Aggregate{{$k.Name}}(p *{{$k.Name}}Point) error
```
lesam, line 17: This interface change is in a separate commit - we need to decide if it is a good idea.

query/functions.go
```diff
@@ -2186,17 +2239,16 @@ func unmarshalPlus(s string) (*hll.Plus, error) {
 	return h, nil
 }
 
-func (r *StringMergeHllReducer) AggregateString(p *StringPoint) {
-	// TODO: how to handle errors?
+func (r *StringMergeHllReducer) AggregateString(p *StringPoint) error {
```
lesam, line 2242: the reason for the error changes

lesam:
Note this will need a docs change - not sure how that will work.

pkg/estimator/hll/hll.go
```diff
@@ -151,9 +151,10 @@ func (h *Plus) Add(v []byte) {
 
 		if uint32(len(h.tmpSet))*100 > h.m {
 			h.mergeSparse()
-			if uint32(h.sparseList.Len()) > h.m {
-				h.toNormal()
-			}
+		}
```
davidby-influx, line 154: It feels like this may result in h.mergeSparse() being called repeatedly.  Is it idempotent?

lesam: Yes. we keep a small `tmpSet` and a large `sparseList`. `mergeSparse` merges the values in `tmpSet` into `sparseList` and empties `tmpSet`.

lesam: `mergeSparse` even already has the optimisation to make it a no-op if `tmpSet` is empty

query/functions.gen.go.tmpl
lesam: @dgnorton and I talked and we agree that this seems like a good idea.

lesam:
Another loose thread to tie up before merge: do we need something either in influxql or influxdb to make sure the type inferences are correct? I had opened https://github.com/influxdata/influxql/issues/41 but this seems to be working without any changes on the influxql repo.

query/functions.go
```diff
@@ -11,6 +14,12 @@ import (
 	"github.com/influxdata/influxql"
 )
 
+var hllPrefix []byte
+
+func init() {
+	hllPrefix = []byte("HLL_")
+}
```
stuartcarnie, line 21: This is simpler:

```suggestion
var hllPrefix = []byte("_HLL")
```

query/functions.gen.go.tmpl
stuartcarnie: Does this change have a noticeable impact on performance? It is unclear to me how frequently this API is called, however, if large enough, pushing an extra value (the `error` return) on to the stack may have an impact.

query/iterator.gen.go.tmpl
```diff
@@ -1318,7 +1320,9 @@ func (itr *{{$k.name}}Stream{{$v.Name}}Iterator) reduce() ([]{{$v.Name}}Point, e
 			}
 			itr.m[id] = rp
 		}
-		rp.Aggregator.Aggregate{{$k.Name}}(curr)
+		if err := rp.Aggregator.Aggregate{{$k.Name}}(curr); err != nil {
+            return nil, err
+        }
```
stuartcarnie, line 1325: nit: `gofmt`

stuartcarnie (dismissed):
Overall this looks great and no doubt a welcomed feature to InfluxQL! Only a couple of nits and a question about performance.

I notice this is for 1.x, so are there plans or issues created to add this feature to 2.x OSS and Cloud 2.0?

query/functions.go
lesam: Done

query/iterator.gen.go.tmpl
lesam: Done

query/functions.gen.go.tmpl
lesam: I added BenchmarkIterator_Aggregator and ran several times (see https://gist.github.com/lesam/9337ce60306d638f03a6bfcf17c42a32 for methodology)

```
benchstat without-error-aggregate.txt with-error-aggregate.txt
name                    old time/op  new time/op  delta
Iterator_Aggregator-12  26.4ns ± 7%  26.7ns ± 1%  +1.23%  (p=0.006 n=19+17)
```

Interpretation: This is about 1% worse performance in the absolute worst case (counting a const value). I would expected real-world `count` to have to do more work (e.g. actually writing to the point) so in the real world I would expect less than 1% impact. I think it's worth it for the code cleanliness, but I can fake the error return with a different string if anyone disagrees.

lesam: Thinking about it more, 1% regression is probably enough to change this around, since only count_hll needs it - let me rework this a bit.

dgnorton: > ...in the real world I would expect less than 1% impact

I don't know if < 1% real world impact is worth it. Passing an error via a string will make for confusing code.

query/functions.go
```diff
@@ -2150,3 +2156,110 @@ func (r *UnsignedBottomReducer) Emit() []UnsignedPoint {
...
+func NewStringMergeHllReducer() *StringMergeHllReducer {
+	return &StringMergeHllReducer{plus: nil}
+}
+
+func unmarshalPlus(s string) (*hll.Plus, error) {
+	if string(hllPrefix) != s[:len(hllPrefix)] {
+		if string(hllErrorPrefix) == s[:len(hllErrorPrefix)] {
+			// parse a special error out of the string.
+			return nil, fmt.Errorf("%v", s[len(hllErrorPrefix):])
+		}
```
lesam, line 2174: error case when unmarshalling

query/functions.go
```diff
@@ -2150,3 +2156,110 @@ func (r *UnsignedBottomReducer) Emit() []UnsignedPoint {
...
+	}
+	err = r.plus.Merge(h)
+	if err != nil {
+		r.err = err
+		return
+	}
+}
+
+func marshalPlus(p *hll.Plus, err error) StringPoint {
+	if err != nil {
```
lesam, line 2214: Error case when marshalling

lesam:
We shouldn't really see errors in this except from actual code bugs, and if we do then the `sum_hll` string should show the error reason.

lesam:
Now with good formatting. Should be good for a re-review @stuartcarnie

query/iterator_test.go
```diff
@@ -1568,6 +1568,66 @@ func TestIterator_EncodeDecode(t *testing.T) {
...
+type IntegerConstIterator struct {
+	numPoints int
+	time      int64
+	value     int64
+	Closed    bool
+	stats     query.IteratorStats
+	point     query.IntegerPoint
+}
+
+func BenchmarkIterator_Aggregator(b *testing.B) {
```
lesam, line 1581: Here's the benchmark I used

query/functions.gen.go.tmpl
lesam: The re-write turned out to be pretty understandable and actually a lot less change overall, I think it should be ok.

query/compile.go
```diff
@@ -771,6 +773,19 @@ func (c *compiledField) compileIntegral(args []influxql.Expr) error {
 	return c.compileSymbol("integral", args[0])
 }
 
+func (c *compiledField) compileCountHll(args []influxql.Expr) error {
+	if exp, got := 1, len(args); exp != got {
+		return fmt.Errorf("invalid number of arguments for integral, expected %d, got %d", exp, got)
```
dgnorton, line 778: Should the error be "...for count_hll"?

query/functions.go
dgnorton: How often do `marshalPlus` and `unmarshalPlus` get called compared to the method where the `error` return was removed? I.e., do we know that this performs better than returning an error?

query/functions.go
```diff
@@ -2150,3 +2156,110 @@ func (r *UnsignedBottomReducer) Emit() []UnsignedPoint {
...
+			Value: string(hllErrorPrefix) + err.Error(),
+		}
+	}
+	if p == nil {
+		return StringPoint{
+			Time:  ZeroTime,
+			Value: string(hllPrefix),
+		}
+	}
+	b, _ := p.MarshalBinary() // no possible errors
```
dgnorton, line 2226: If a function has an error return value, it's best to check it even if the current implementation of the function doesn't currently return an error. If someone adds an error return in the future, they're unlikely check every caller and shouldn't have to since the contract says it's okay to return an error.

query/functions_test.go
```diff
@@ -497,3 +503,76 @@ func TestSample_SampleSizeGreaterThanNumPoints(t *testing.T) {
...
+	for _, p := range input[:2000] {
+		s1.AggregateString(p)
+	}
+	point1 := s1.Emit()
+	s2 := query.NewStringSumHllReducer()
+	for _, p := range input[1000:] {
+		s2.AggregateString(p)
+	}
+	point2 := s2.Emit()
+	require.Equal("17190211103962133664", input[2999].Value)
```
dgnorton, line 529: I'm probably missing the obvious but why is this check of the last input string needed?

dgnorton (changes requested):

query/functions.go
dgnorton: I guess this this way impacts `count_hll` but avoids impacting all other aggregates?

query/compile.go
lesam: yes

query/functions.go
lesam: Yes exactly. I'm not too worried about slowing `count_hll`, especially since it actually can have errors, but I don't want to slow down `sum`, `count`, `mean`, etc.

query/functions.go
lesam: Good point. This is left over from when I was hoping not to return any error at all - fixed.

query/functions_test.go
lesam: This is just to show people reading the test the format of the values (they are stringified numbers) and they really are repeatable. I'll add a comment.

dgnorton (approved):
lgtm

lesam:
> Overall this looks great and no doubt a welcomed feature to InfluxQL! Only a couple of nits and a question about performance.
> 
> I notice this is for 1.x, so are there plans or issues created to add this feature to 2.x OSS and Cloud 2.0?

It will be a product decision whether to pull this in to 2.x OSS / cloud. There is an open request to add this to flux: https://github.com/influxdata/flux/issues/354 , and also an aggregated request to add new functions to 2.x influxql: https://github.com/influxdata/influxdb/issues/5930

PR #20640, opened 2021-01-28
fix(cmd/influx): improve logging consistency in `influxd upgrade`

### LINKED ISSUE #20637: Inconsistent (spammy) logging in `influxd upgrade`
Info-level logging in `influxd upgrade` is (rightly) pretty terse, but we still log an info line per migrated user. Those per-user logs should be debug-level, and an info-level log should report the migration summary.

danxmoran:
Closes #20637

The original motivation was to make per-user logs `debug` level, but I found more things while I was in the code.


davidby-influx (approved):
LGTM.

PR #20643, opened 2021-01-28
docs: mention that flux needs to be in PATH in DEVELOPMENT.md

jdstrand:
Following https://github.com/influxdata/influxdb/blob/master/DEVELOPMENT.md#full-length-guide results in:
```
$ go run ./cmd/influxd --assets-path=ui/build
go: downloading github.com/spf13/cobra v1.0.0
...
# pkg-config --cflags  -- flux
Package flux was not found in the pkg-config search path.
Perhaps you should add the directory containing `flux.pc'
to the PKG_CONFIG_PATH environment variable
No package 'flux' found
pkg-config: exit status 1
$
```

After installing `flux` in GOBIN (also see https://github.com/influxdata/flux/pull/3458), things work fine:
```
$ PATH=$GOPATH/bin:$PATH go run ./cmd/influxd --assets-path=ui/build
2021-01-28T20:15:37.399475Z	info	Welcome to InfluxDB	{"log_id": "0Rzf1sil000", "version": "dev", "commit": "none", "build_date": "2021-01-28T20:15:35Z"}
...
```

I signed the CLA just now.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -54,6 +54,7 @@ To get started with Influx, you'll need to install these tools if you don't alre
 1. [Install go](https://golang.org/doc/install)
 1. [Install nodejs](https://nodejs.org/en/download/package-manager/)
 1. [Install yarn](https://yarnpkg.com/lang/en/docs/install/)
+1. [Install flux](https://github.com/influxdata/flux) and make sure it is in your PATH
 
 Yarn is a package manager for nodejs and an alternative to npm.
 


jdstrand:
I should mention that I tried to build master. If I check out tags/v2.0.3, see: `go: downloading github.com/influxdata/flux v0.99.0`

danxmoran:
Ergh, I didn't know we had a `DEVELOPMENT.md` lying around in addition to `CONTRIBUTING.md`... We should consolidate the two, but in the meantime `CONTRIBUTING` has received the most recent face-lift (see #20298 for the latest revisions).

The error you hit is what happens when `go` uses the "standard" `pkg-config` instead of our wrapper that knows how to handle Flux. The installation instructions for Flux happen to include a line to install the custom pkg-config into `GOBIN`, which (I think) is why you see success after installing Flux and adding `GOBIN` to your `PATH`.

Our `Makefile` is wired to set up all the dependencies and `PKG_CONFIG` env var, along with build tags. We don't have a strong story for bypassing `make` in favor of using `go build` or `go run`.

jdstrand:
> The error you hit is what happens when `go` uses the "standard" `pkg-config` instead of our wrapper that knows how to handle Flux. The installation instructions for Flux happen to include a line to install the custom pkg-config into `GOBIN`, which (I think) is why you see success after installing Flux and adding `GOBIN` to your `PATH`.

For context, I did indeed try to build influxdb first without the pkg-config wrapper (since it isn't part of these instructions) and encountered the error. When I went to compile flux (not knowing about the wrapper), I added the pkg-config wrapper to GOBIN and added GOBIN to my PATH but found that I couldn't compile openssl when using cargo with it there (ie, following the flux instructions), so I removed it from PATH and chose to use PKG_CONFIG=path/to/wrapper (as mentioned in the other PR).

danxmoran:
Closing this since `CONTRIBUTING.md` is the up-to-date file. I deleted `DEVELOPMENT.md` in #20747 to avoid future confusiuon

PR #20657, opened 2021-02-01
fix(cmd/influx): allow for creating users without initial passwords in `influx user create`

### LINKED ISSUE #20645: 'influx user create' cmd without -p flag misleading
When running the **`influx user create command -n user_name -o org`** the CLI shows this error message: **`Error: Passwords must be at least 8 characters long.`**

Which is misleading or it's a bug. Im not sure what was the intention here. 

The reason for reporting this is because user is still created even after not supplying the -p flag. 
It's also not mentioned in help or documentantion that -p flag is required, so I assume this error msg must have been overlooked and should not be displayed on `influx user create`

https://docs.influxdata.com/influxdb/v2.0/reference/cli/influx/user/create/ 
This documentation also mentions that -p flag and -o/--org flag are mutually exclusive, which is not the case. User can still be create with both flags.

Screenshot from 2021-01-29 09-51-01

danxmoran:
Closes #20645 

I manually tested the case of setting a password without specifying an org, it's tough to unit-test because the org-flag helpers are coded to read a global config file as a fallback when no flags are passed in.

### DIFF
--- a/cmd/influx/user.go
+++ b/cmd/influx/user.go
@@ -2,15 +2,17 @@ package main
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	"github.com/influxdata/influxdb/v2"
 	"github.com/influxdata/influxdb/v2/cmd/internal"
 	"github.com/influxdata/influxdb/v2/http"
+	ilogger "github.com/influxdata/influxdb/v2/logger"
 	"github.com/influxdata/influxdb/v2/tenant"
 	"github.com/spf13/cobra"
 	"github.com/tcnksm/go-input"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
 type userSVCsFn func() (cmdUserDeps, error)
@@ -34,12 +36,10 @@ type cmdUserBuilder struct {
 
 	svcFn userSVCsFn
 
-	hideHeaders bool
-	id          string
-	json        bool
-	name        string
-	password    string
-	org         organization
+	id       string
+	name     string
+	password string
+	org      organization
 }
 
 func newCmdUserBuilder(svcsFn userSVCsFn, f *globalFlags, opt genericCLIOpts) *cmdUserBuilder {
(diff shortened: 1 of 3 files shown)


cmd/influx/user.go
```diff
@@ -187,35 +204,46 @@ func (b *cmdUserBuilder) cmdCreateRunEFn(*cobra.Command, []string) error {
...
+	if b.password != "" {
+		setPassErr = dep.passSVC.SetPassword(ctx, user.ID, b.password)
+	} else {
+		log.Warn("No initial password set on user, use `influx user password`", zap.String("user", b.name))
 	}
 
-	pass := b.password
-	if orgID == 0 && pass == "" {
-		return b.printUser(userPrintOpts{user: user})
+	orgID, setOrgErr := b.org.getID(dep.orgSvc)
```
davidby-influx, line 214: Do we want to bail earlier (i.e., here) if setPassErr != nil, and not try to set the Org?

cmd/influx/user.go
```diff
@@ -187,35 +204,46 @@ func (b *cmdUserBuilder) cmdCreateRunEFn(*cobra.Command, []string) error {
...
+			UserID:       user.ID,
+			UserType:     influxdb.Member,
+			ResourceType: influxdb.OrgsResourceType,
+			ResourceID:   orgID,
+		})
 	}
 
-	if pass != "" && orgID == 0 {
-		return errors.New("an org id is required when providing a user password")
+	if setPassErr != nil {
```
davidby-influx, line 224: Consider moving this above the attempt to set the organization.

davidby-influx (approved):
Wasn't sure if we should try to set the Org if we can't set the password, but maybe that's correct (saving the user another redo step when they set the password).

Minor point anyway.

cmd/influx/user.go
danxmoran: I'll restore the previous bail-out order (org first if ID is specified, then password). Better to not rock the boat too much on a patch release.

davidby-influx (approved):
Looks clean and clear.

PR #20666, opened 2021-02-01
refactor(cmd/influx): clean up existing tests for `influx write`

danxmoran:
Before we start adding new functionality to `influx write`, I wanted to give the existing tests a facelift.

### DIFF
--- a/cmd/influx/write_test.go
+++ b/cmd/influx/write_test.go
@@ -3,7 +3,6 @@ package main
 import (
 	"bufio"
 	"bytes"
-	"compress/gzip"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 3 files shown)


cmd/influx/write_test.go
```diff
@@ -371,128 +371,50 @@ func Test_fluxWriteDryrunF(t *testing.T) {
...
 
 	t.Run("validates decoding of org-id", func(t *testing.T) {
-		useTestServer()
 		command := cmdWrite(&globalFlags{}, genericCLIOpts{w: ioutil.Discard, viper: viper.New()})
 		command.SetArgs([]string{"--format", "csv", "--org-id", "my-org", "--bucket", "my-bucket"})
 		err := command.Execute()
 		require.Contains(t, fmt.Sprintf("%s", err), "org-id")
 	})
 
-	t.Run("validates error when failed to retrieve buckets", func(t *testing.T) {
```
danxmoran, line 458: These 3 are no longer relevant (at least for unit tests). The write API accepts both IDs and names in write requests (I assume because resolving an ID from a name requires read-access on the target resource, which writers might not have). It's up to the server to do the right thing with all the parameters.

I'll add a replacement case for a generic failure-on-write.

davidby-influx (approved):
LGTM

PR #20669, opened 2021-02-02
fix(write): drop empty lines from write batches

### LINKED ISSUE #20667: Server-side failure when `influx write` sends blank lines
Users & Support have reported failures in Cloud when `influx write` is used on an empty file (no failure is seen on OSS). We can work around the issue for now by having the CLI command filter out blank lines before sending data to the server. Empty files manifest as blank lines within our writer because of our multi-reader setup.

danxmoran:
Closes #20667

The 2nd commit is unrelated, but useful for having Support test with binaries built from branches.

### DIFF
--- a/write/batcher.go
+++ b/write/batcher.go
@@ -96,7 +96,9 @@ func (b *Batcher) read(ctx context.Context, r io.Reader, lines chan<- []byte, er
 	for scanner.Scan() {
 		// exit early if the context is done
 		select {
-		case lines <- scanner.Bytes():
+		// NOTE: We purposefully don't use scanner.Bytes() here because it returns a slice
+		// pointing to an array which is reused / overwritten on every call to Scan().
+		case lines <- []byte(scanner.Text()):
 		case <-ctx.Done():
 			errC <- ctx.Err()
 			return
--- a/write/batcher.go
+++ b/write/batcher.go
@@ -96,7 +96,7 @@ func (b *Batcher) read(ctx context.Context, r io.Reader, lines chan<- []byte, er
 	for scanner.Scan() {
 		// exit early if the context is done
 		select {
-		case lines <- []byte(scanner.Text()):
+		case lines <- scanner.Bytes():
 		case <-ctx.Done():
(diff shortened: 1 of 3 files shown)


write/batcher.go
```diff
@@ -96,7 +96,7 @@ func (b *Batcher) read(ctx context.Context, r io.Reader, lines chan<- []byte, er
 	for scanner.Scan() {
 		// exit early if the context is done
 		select {
-		case lines <- []byte(scanner.Text()):
+		case lines <- scanner.Bytes():
```
lesam, line 99: >  The underlying array may point to data that will be overwritten by a subsequent call to Scan

That sounds scary - are we sure this data is not being overwritten? - i.e. we call `Bytes`, put it on the channel again, and call `Scan` before picking the bytes off the channel.

danxmoran: Hmmmm maybe that's why we were using `Text()`... I'll restore it, and add a comment for the next person to see

danxmoran: Fixed!

write/batcher_test.go
```diff
@@ -304,26 +324,26 @@ func TestBatcher_write(t *testing.T) {
...
 			err := <-tt.args.errC
 			if (err != nil) != tt.wantErr {
 				t.Errorf("ScanLines.read() error = %v, wantErr %v", err, tt.wantErr)
 				return
 			}
 
-			if !cmp.Equal(got, tt.want) {
-				t.Errorf("%q. Batcher.write() = -got/+want %s", tt.name, cmp.Diff(got, tt.want))
-			}
+			require.Equal(t, tt.wantNoCall, !writeCalled)
```
lesam, line 345: This logic would be a bit prettier if this had been called `wantCall`

lesam (approved):
lgtm

write/batcher_test.go
danxmoran: True! I like `wantNoCall` because the zero value is the most common case, so we can leave it unspecified in most of the test-case structs & reduce noise.

PR #20690, opened 2021-02-03
refactor: update window logic to use interval package

jlapacik:


rockstar (approved):
Thanks for doing this

PR #20702, opened 2021-02-05
fix(kit/cli): fix loading config from directories with '.' in the name

### LINKED ISSUE #20696: INFLUXD_CONFIG_PATH does not work if there is a `.` in the directory name
According to the docs: https://docs.influxdata.com/influxdb/v2.0/reference/config-options/#influxdb-configuration-file
> On startup, influxd will check for a config.* in the INFLUXD_CONFIG_PATH directory.

__Steps to reproduce:__

1. Create `config.toml` in a directory, for example `/vol/mounted-drive/config.toml`
2. Set `INFLUXD_CONFIG_PATH` to the parent directory of the config file, e.g. `INFLUXD_CONFIG_PATH=/vol/mounted-drive`
3. Run `influxd`

__Expected behavior:__
`influxd` should find and use `/vol/mounted-drive/config.toml`

__Actual behavior:__
The custom configuration is not used

__Environment info:__

* System info: `Linux 4.4.0-1085-aws x86_64`
* InfluxDB version: `InfluxDB 2.0.3 (git: fe04d346df) build_date: 2020-12-15T01:00:16Z`

---

This may not be a bug, but rather an inaccuracy in the documentation. The workaround is to simply point `INFLUXD_CONFIG_PATH` at `config.toml` itself, rather than the parent directory.

danxmoran:
Closes #20696

Instead of using an extension-check to guess if `INFLUXD_CONFIG_PATH` is pointing to a file vs. a directory, use `os.Stat` + `IsDir()`.

This change introduces a new failure mode when `INFLUXD_CONFIG_PATH` is set in the env, but points to a nonexistent file. I think that's a win, but could be convinced to revert to the previous behavior.

### DIFF
--- a/kit/cli/viper.go
+++ b/kit/cli/viper.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"fmt"
 	"os"
+	"path"
 	"strings"
 	"time"
 
@@ -79,17 +80,11 @@ func initializeConfig(v *viper.Viper) error {
 		configPath = "."
 	}
 
-	fi, err := os.Stat(configPath)
-	if err != nil {
-		return err
-	}
-	if fi.IsDir() {
-		// NOTE: if given a directory, viper will look in the dir for a file
-		// with base name `config` and an extension from `viper.SupportedExts`.
-		// Of those extensions, we only document support for .json|.toml|.yaml|.yml
-		v.AddConfigPath(configPath)
-	} else {
+	switch strings.ToLower(path.Ext(configPath)) {
+	case ".json",".toml",".yaml",".yml":
 		v.SetConfigFile(configPath)
+	default:
+		v.AddConfigPath(configPath)
 	}
 
 	if err := v.ReadInConfig(); err != nil && !os.IsNotExist(err) {
--- a/kit/cli/viper.go
+++ b/kit/cli/viper.go
@@ -3,7 +3,6 @@ package cli
 import (
 	"fmt"
 	"os"
-	"path"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


kit/cli/viper.go
```diff
@@ -60,20 +59,6 @@ func NewCommand(v *viper.Viper, p *Program) *cobra.Command {
 	// This normalizes "-" to an underscore in env names.
 	v.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 
-	if configPath := v.GetString("CONFIG_PATH"); configPath != "" {
-		switch strings.ToLower(path.Ext(configPath)) {
-		case ".json", ".toml", ".yaml", ".yml":
-			v.SetConfigFile(configPath)
-		case "":
-			v.AddConfigPath(configPath)
-		}
```
lesam, line 69: I think this should have just been changed to:

```
		switch strings.ToLower(path.Ext(configPath)) {
 		case ".json", ".toml", ".yaml", ".yml":
 			v.SetConfigFile(configPath) // maybe also check if configPath is actually a directory, and if so, error message.
 		default:
 			v.AddConfigPath(configPath)
 		}
```
if someone has '*.yaml' as their config directory they deserve problems.

danxmoran: I did consider that change originally. The question to answer is: what do we want the behavior to be if a user sets `INFLUXD_CONFIG_PATH` to point at a file that doesn't end in one of our supported extensions?

As it is now, the PR would detect that the env is pointing directly at a file, pass that file along to `viper`, and then `viper` would raise an error about the unsupported extension.

With the suggested change, the code would detect the extension mismatch and add the path to `viper`'s search path. `viper` would then stat `${INFLUXD_CONFIG_PATH}/config.${ext}` for each of its supported extensions (all failing). Another error-handling block below the changes here would ensure the config-loading failure didn't crash the program, so `influxd` would boot up with default config values.

lesam: If the user specifies a config file that doesn't have the right extension, I want `influxd` to refuse to boot.

lesam: If they specify a directory (or a path that doesn't point to a file or directory) I'm fine with them getting default config.

danxmoran: Implemented below

lesam (approved):
lgtm

PR #20705, opened 2021-02-05
fix(http/swagger): repair DBRPs type to match implementation

sranka:
This PR repairs swagger DBRPs type to match actual implementation in https://github.com/influxdata/influxdb/blob/master/dbrp/http_server_dbrp.go#L128

- `notificationEndpoints` property a was copy-paste mistake, it is called `content`
- `links` property removed, it is not returned by the API

This change is also required for InfluxDB client libraries that rely upon code generated from the swagger file.

- [x] Rebased/mergeable
- [x] Tests pass


http/swagger.yml
```diff
@@ -12036,12 +12036,10 @@ components:
           - retention_policy
     DBRPs:
       properties:
-        notificationEndpoints:
+        content:
```
danxmoran, line 12039: To match the convention of other schemas, I think this should be `dbrps`

http/swagger.yml
```diff
@@ -12036,12 +12036,10 @@ components:
...
           type: array
           items:
             $ref: "#/components/schemas/DBRP"
-        links:
-          $ref: "#/components/schemas/Links"
```
danxmoran, line 12044: This seems standard in many other objects returned by the API, I think we should instead try to fix the implementation to fill it in.

danxmoran (changes requested):
We normally consider the API spec to be the source of truth, and fix the implementation to match when the two don't align. Agreed on the need to rename the key in the `DBRPs` schema, but we'll need to coordinate with the Cloud team to make the same change there.

http/swagger.yml
sranka: This is IMHO just a copy-paste error in the swagger file. The existing `influx v1 auth` command already uses the `content` property and it already works well in OSS and in the cloud, the client also uses https://github.com/influxdata/influxdb/blob/master/dbrp/http_server_dbrp.go#L128. IMHO it would be better to fix it to match the implementation state first and then possibly create a defect for a better name and links later as a second step if you want to, WDYT?

danxmoran: Yeah, with fresh eyes it sounds like a good idea to fix for now and then improve later 👍

danxmoran (dismissed):

danxmoran (commented):
Can you add a line to `CHANGELOG.md`?

danxmoran (approved):
Thanks!

Issue #20718, opened 2021-02-06
e2e tests are incredibly flaky


danxmoran:
Our e2e job fails more often than not in CI. See if the changes applied to the `ui` repo in #20482 could also work for our copy of the UI code.

danxmoran:
A sample list of failures from the past few days:

- [ ] The Query Builder from the Data Explorer creates a query, edits it to add another field, then views its results with pride and satisfaction - creates a query, edits it to add another field, then views its results with pride and satisfaction (this is the most common failure)
```
AssertionError: Timed out retrying: Expected to find content: 'Configure' but never did.
    at Context.eval (http://localhost:8086/__cypress/tests?p=cypress/e2e/queryBuilder.test.ts:53:16)
```

- [ ] DataExplorer raw script editing shows flux signatures - shows flux signatures (2nd-most common failure)
```
AssertionError: Timed out retrying: Expected to find element: `.signature`, but never found it.
    at Context.eval (http://localhost:8086/__cypress/tests?p=cypress/e2e/explorer.test.ts:871:42)
```

- [ ] Buckets from the buckets index page can create a bucket - can create a bucket
```
AssertionError: Timed out retrying: Expected to find element: `[data-testid="create-label-form--submit"]`, but never found it.
    at Context.exports.getByTestID (http://localhost:8086/__cypress/tests?p=cypress/support/index.js:870:15)
```

- [ ] DataExplorer raw script editing can filter aggregation functions by name from script editor mode - can filter aggregation functions by name from script editor mode
```
AssertionError: Timed out retrying: expected '<input.cf-input-field>' to have value 'covariance', but the value was 'c'
    at Context.eval (http://localhost:8086/__cypress/tests?p=cypress/e2e/explorer.test.ts:938:18)
```

- [ ] Dashboard can create, clone and destroy cells & toggle in and out of presentation mode - can create, clone and destroy cells & toggle in and out of presentation mode
```
CypressError: `cy.click()` can only be called on a single element. Your subject contained 2 elements. Pass `{ multiple: true }` if you want to serially click each element.

https://on.cypress.io/click
    at mouseAction (http://localhost:8086/__cypress/runner/cypress_runner.js:151230:17)
    at Context.click (http://localhost:8086/__cypress/runner/cypress_runner.js:151374:14)
    at Context.<anonymous> (http://localhost:8086/__cypress/runner/cypress_runner.js:168170:21)
From Your Spec Code:
[...]

PR #18090, opened 2020-05-13
chore: import selenium e2e tests

glinton:
Import selenium e2e tests for influxdb from external repo

PR #20728, opened 2021-02-08
build: clean up e2e code

danxmoran:
The 1st commit deletes the `hourly-e2e` job that has been spamming us with failures. #20718 lists the failures I've seen in the past few days, for tracking.

The 2nd commit deletes an entirely different selenium-based `e2e` suite that hasn't been used for months. It was imported from a Bonitoo repo in #18090, but I'm not sure if it was ever added to CI. I confirmed with Zoe and Kristina from the UI team that the cypress-based `e2e` suites in the `ui/` directory and `influxdata/ui` repo match the coverage of this old suite, so there should be no harm in deleting it.


davidby-influx (approved):
Approving on principle. Burn bad tests.

Issue #20736, opened 2021-02-09
Make meta queries respect QueryTimeout values


davidby-influx:
__Proposal:__
Meta queries (SHOW TAG VALUES, SHOW TAG KEYS, SHOW SERIES CARDINALITY, etc.) do not respect the QueryTimeout config  parameter

__Current behavior:__
Meta queries can only be cancelled in the first few moments of operation; during the actual data gathering and return phases, they do not check the query context object for cancellation or timeout.

__Desired behavior:__
Meta queries should check the query context when possible to allow cancellation and timeout.  This will not be as frequent as regular queries, which use iterators, because meta queries return data in batches.

__Use case:__
Speaking with the support and SA team, this is a request  is a control that would benefit many, particularly those who are trying to enforce the users running dashboards. In one case, they believe the meta queries are coming from end users in Grafana with each key stroke. However bad practices do happen, hence we have a query timeout. This should either be enforced for meta queries as well.

davidby-influx:
PR for master-1.x branch: https://github.com/influxdata/influxdb/pull/20739

PR #20739, opened 2021-02-09
feat: Make meta queries respect QueryTimeout values

davidby-influx:
Meta queries (SHOW TAG VALUES, SHOW TAG KEYS, SHOW SERIES CARDINALITY, etc.) do not respect
the QueryTimeout config parameter. Meta queries should check the query context when possible
to allow cancellation and timeout. This will not be as frequent as regular queries, which
use iterators, because meta queries return data in batches.

Add a context.Context to
(*Store).MeasurementNames()
(*Store).MeasurementsCardinality()
(*Store).SeriesCardinality()
(*Store).TagValues()
(*Store).TagKeys()
(*Store).SeriesSketches()
(*Store).MeasurementsSketches()
which is tested for timeout or cancellation
to allow limitation of time spent in meta queries

Closes https://github.com/influxdata/influxdb/issues/20736

- [X] Rebased/mergeable
- [X] Tests pass

### DIFF
--- a/coordinator/statement_executor.go
+++ b/coordinator/statement_executor.go
@@ -180,11 +180,11 @@ func (e *StatementExecutor) ExecuteStatement(stmt influxql.Statement, ctx *query
 	case *influxql.ShowMeasurementsStatement:
 		return e.executeShowMeasurementsStatement(stmt, ctx)
 	case *influxql.ShowMeasurementCardinalityStatement:
-		rows, err = e.executeShowMeasurementCardinalityStatement(stmt)
+		rows, err = e.executeShowMeasurementCardinalityStatement(stmt, ctx)
 	case *influxql.ShowRetentionPoliciesStatement:
 		rows, err = e.executeShowRetentionPoliciesStatement(stmt)
 	case *influxql.ShowSeriesCardinalityStatement:
-		rows, err = e.executeShowSeriesCardinalityStatement(stmt)
+		rows, err = e.executeShowSeriesCardinalityStatement(stmt, ctx)
 	case *influxql.ShowShardsStatement:
 		rows, err = e.executeShowShardsStatement(stmt)
 	case *influxql.ShowShardGroupsStatement:
@@ -719,7 +719,7 @@ func (e *StatementExecutor) executeShowMeasurementsStatement(q *influxql.ShowMea
 		return ErrDatabaseNameRequired
 	}
 
-	names, err := e.TSDBStore.MeasurementNames(ctx.Authorizer, q.Database, q.Condition)
+	names, err := e.TSDBStore.MeasurementNames(ctx.Context, ctx.Authorizer, q.Database, q.Condition)
 	if err != nil || len(names) == 0 {
 		return ctx.Send(&query.Result{
 			Err: err,
@@ -758,12 +758,12 @@ func (e *StatementExecutor) executeShowMeasurementsStatement(q *influxql.ShowMea
 	})
 }
 
-func (e *StatementExecutor) executeShowMeasurementCardinalityStatement(stmt *influxql.ShowMeasurementCardinalityStatement) (models.Rows, error) {
+func (e *StatementExecutor) executeShowMeasurementCardinalityStatement(stmt *influxql.ShowMeasurementCardinalityStatement, ctx *query.ExecutionContext) (models.Rows, error) {
 	if stmt.Database == "" {
 		return nil, ErrDatabaseNameRequired
 	}
 
-	n, err := e.TSDBStore.MeasurementsCardinality(stmt.Database)
+	n, err := e.TSDBStore.MeasurementsCardinality(ctx.Context, stmt.Database)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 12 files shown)


coordinator/statement_executor.go
```diff
@@ -180,11 +180,11 @@ func (e *StatementExecutor) ExecuteStatement(stmt influxql.Statement, ctx *query
 	case *influxql.ShowMeasurementsStatement:
 		return e.executeShowMeasurementsStatement(stmt, ctx)
 	case *influxql.ShowMeasurementCardinalityStatement:
-		rows, err = e.executeShowMeasurementCardinalityStatement(stmt)
+		rows, err = e.executeShowMeasurementCardinalityStatement(stmt, ctx)
```
lesam, line 183: ctx is conventionally the first parameter of any method it is in.

tsdb/store.go
```diff
@@ -1180,40 +1180,57 @@ func (s *Store) sketchesForDatabase(dbName string, getSketches func(*Shard) (est
...
+		}
 	})
-
+	if err != nil {
+		return 0, err
+	}
 	ss := NewSeriesIDSet()
 	ss.Merge(others...)
-	return int64(ss.Cardinality()), nil
+	select {
```
lesam, line 1214: I don't know if this bail out is necessary - I'd assume by this point there's a minimal amount of processing remaining so we might as well just return the cardinality.

lesam (commented):
Do we know exactly which section is taking forever for the customer that has the issue? (block or perf profiles should tell us). We should make sure we have a ctx.Done check inside whatever loop is taking a while.

coordinator/statement_executor.go
davidby-influx: @lesam - agreed.  However, it was passed as the last parameter in some of the other functions here, so I was trying to follow their pattern.  Elsewhere in the code I pass it as normally done.

tsdb/store.go
davidby-influx: My timings showed that the Cardinality computation could take a modicum of time, so I dropped in this check.  Agreed, it's a later bail-out that the earlier checks.

davidby-influx:
> Do we know exactly which section is taking forever for the customer that has the issue? (block or perf profiles should tell us). We should make sure we have a ctx.Done check inside whatever loop is taking a while.

I did manual timings to get a general sense, and added the checks anywhere I saw a loop that was easy to amend.  I have not asked for customer profiles on this.

coordinator/statement_executor.go
davidby-influx: Do you want me to switch the context to the first argument?  If so, would you like me to also change the function signature of TaskManager.ExecuteStatement(), executeExplainStatement(), executeExplainAnalyzeStatement(), executeSelectStatement(), and the caller, ExecuteStatement(), all of which have the context as the last argument?
My preference was to leave the functions called from ExecuteStatement() as is, with the context as the last argument, match that parameter order for the new signatures, and then use context as the first argument as is conventional everywhere else.

dgnorton: May as well make it consistent, unless that would have a large impact on other repos.

davidby-influx: Will refactor if that's the consensus.

tsdb/store.go
```diff
@@ -1180,40 +1180,57 @@ func (s *Store) sketchesForDatabase(dbName string, getSketches func(*Shard) (est
...
-	s.walkShards(shards, func(sh *Shard) error {
-		index, err := sh.Index()
-		if err != nil {
-			return err
-		}
+	err := s.walkShards(shards, func(sh *Shard) error {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
```
lesam, line 1195: nit: you have one pattern like:

```
    select {
    case <-ctx.Done():
      return ctx.Err()
    default:
    }
    // more code
```

And another like:

```
    select {
    case <-ctx.Done():
      return ctx.Err()
    default:
      //more code
    }
```

I prefer the first one. Not blocking.

davidby-influx: @lesam - Thanks. I like the second pattern, but I used the first to avoid indenting too deeply.  I'll standardize in the future; I think my aesthetics are not yet pure Go.

tsdb/store.go
```diff
@@ -1226,13 +1243,8 @@ func (s *Store) SeriesSketches(database string) (estimator.Sketch, estimator.Ske
...
 // method cannot be combined with any other results.
-func (s *Store) MeasurementsCardinality(database string) (int64, error) {
-	ss, ts, err := s.sketchesForDatabase(database, func(sh *Shard) (estimator.Sketch, estimator.Sketch, error) {
-		if sh == nil {
-			return nil, nil, errors.New("shard nil, can't get cardinality")
-		}
-		return sh.MeasurementsSketches()
-	})
+func (s *Store) MeasurementsCardinality(ctx context.Context, database string) (int64, error) {
+	ss, ts, err := s.MeasurementsSketches(ctx, database)
```
lesam, line 1247: 👍 for removing duplicated code.

lesam (approved):
lgtm - see the one comment

PR #20745, opened 2021-02-10
feat: use count_hll for 'show series cardinality' queries

### LINKED ISSUE #20614: 'SHOW SERIES CARDINALITY' queries should use count_hll instead of count
__Proposal:__
The non-exact versions of `SHOW SERIES CARDINALITY from ... ` should always use hyper-log-log enabled counters.

__Current behavior:__
Currently a plain `SHOW SERIES CARDINALITY` is a highly-optimised hyper-log-log query, but any more complicated queries are rewritten using `select count(distinct(_seriesKey))`.

__Desired behavior:__
 After https://github.com/influxdata/influxdb/pull/20603 merges, we will be able to re-write the non-exact versions using `select count_hll(sum_hll(_seriesKey))`, which is provides a much more efficient estimate.

__Alternatives considered:__
See also: https://github.com/influxdata/influxdb/issues/20613 . Both features would be useful. This one would be more for ad-hoc queries where that one is more for ongoing monitoring.

__Use case:__
Cardinality counting is especially useful for large-cardinality databases using the TSI index to determine which measurements etc. are using up the available cardinality. But currently anything other than a bare `SHOW SERIES CARDINALITY` requires a large amount of RAM to calculate. Optimizing `SHOW SERIES CARDINALITY from MEASUREMENT where ...` type queries would improve usability of these metaqueries for the large installations where they are most valuable.

lesam:
Closes #20614

In addition to the query rewrites, fix a recent bug in seriesKey iteration exposed by the cardinality tests,
and add an extra tests for the recent ingress metric work.

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/query/statement_rewriter.go
+++ b/query/statement_rewriter.go
@@ -202,6 +202,30 @@ func rewriteShowSeriesCardinalityStatement(stmt *influxql.ShowSeriesCardinalityS
 		}
 	}
 
+	// do hll estimation for per-measurement queries
+	if !stmt.Exact && stmt.Dimensions == nil && stmt.Limit == 0 && stmt.Offset == 0 {
+		return &influxql.SelectStatement{
+			Fields: []*influxql.Field{
+				{
+					Expr: &influxql.Call{
+						Name: "count_hll",
+						Args: []influxql.Expr{&influxql.Call{
+							Name: "sum_hll",
+							Args: []influxql.Expr{&influxql.VarRef{Val: "_seriesKey"}},
+						}},
+					},
+					Alias: "cardinality estimation",
+				},
+			},
+			Sources:    rewriteSources2(stmt.Sources, stmt.Database),
+			Condition:  stmt.Condition,
+			Dimensions: stmt.Dimensions,
+			Offset:     stmt.Offset,
+			Limit:      stmt.Limit,
+			OmitTime:   true,
+		}, nil
+	}
+
 	return &influxql.SelectStatement{
 		Fields: []*influxql.Field{
 			{
--- a/tsdb/index.go
+++ b/tsdb/index.go
@@ -1933,7 +1933,7 @@ type measurementSeriesKeyByExprIterator struct {
 }
 
 func (itr *measurementSeriesKeyByExprIterator) Next() ([]byte, error) {
-	if itr == nil {
+	if itr.ids == nil {
 		return nil, nil
 	}
 	for {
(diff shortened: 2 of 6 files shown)


tsdb/index.go
```diff
@@ -2001,7 +2008,7 @@ func (is IndexSet) MeasurementSeriesKeyByExprIterator(name []byte, expr influxql
 	}
 	if ids == nil {
 		release()
-		return nil, nil
+		release = nil
```
lesam, line 2011: For us to successfully use a nil `measurementSeriesKeyByExprIterator`, we would have to return a boxed nil, which is ugly.

Instead just set the `ids` to nil and the methods can check that.

query/statement_rewriter.go
```diff
@@ -202,6 +202,30 @@ func rewriteShowSeriesCardinalityStatement(stmt *influxql.ShowSeriesCardinalityS
 		}
 	}
 
+	// do hll estimation for per-measurement queries
+	if !stmt.Exact && stmt.Dimensions == nil && stmt.Limit == 0 && stmt.Offset == 0 {
```
lesam, line 206: I could take these conditions off (except exact) but I think it would muddy the waters.

`count_hll` works for any kind of query, but the optimized series iteration only works for no dimension/limit/offset.

tsdb/index.go
```diff
@@ -1982,7 +1982,8 @@ func (itr *measurementSeriesKeyByExprIterator) Next() ([]byte, error) {
 }
 
 func (itr *measurementSeriesKeyByExprIterator) Close() error {
-	if itr == nil {
```
davidby-influx, line 1985: Do you want to get rid of the `itr == nil` check? Would that be impossible? Having both seems more defensive.

tsdb/index.go
```diff
@@ -1933,7 +1933,7 @@ type measurementSeriesKeyByExprIterator struct {
 }
 
 func (itr *measurementSeriesKeyByExprIterator) Next() ([]byte, error) {
-	if itr == nil {
+	if itr.ids == nil {
```
davidby-influx, line 1936: As below on Close() - perhaps check for `itr == nil` as well.

davidby-influx (approved):
Minor questions about defensive programming.

tsdb/index.go
lesam: I think I'd prefer the panic - nobody should be using a nil, and we usually don't check it I think.

lesam: Dave Cheney is on my side here, which is good enough for me:

https://dave.cheney.net/practical-go/presentations/gophercon-singapore-2019.html#_never_use_nil_to_indicate_failure

> Tip: Don’t check for a nil receiver, employ high test coverage and vet/lint tools to spot unhandled error conditions resulting in nil receivers.

davidby-influx: Unfortunately, we have that panic happening, and I have to figure it out. It looks like a failed cleanup from the creation of multiple iterators, which are all then closed, and some are nil while others were created.....

PR #20754, opened 2021-02-11
fix: update doc links to point at new site

### LINKED ISSUE #20673: Update documentation links in the product to point to docs.influxdata.com
https://github.com/influxdata/influxdb/search?q=%22v2.docs.influxdata.com%22

These should all be updated to point to the new docs links.

danxmoran:
Closes #20673


russorat (approved):
LGTM, but i added @sanderson to double check the links

ui/src/dataLoaders/components/collectorsWizard/select/SelectCollectorsStep.tsx
```diff
@@ -61,14 +61,14 @@ export class SelectCollectorsStep extends PureComponent<Props> {
           <h5 className="wizard-step--sub-title">
             Looking for other things to monitor? Check out our 200+ other &nbsp;
             <a
-              href="https://v2.docs.influxdata.com/v2.0/reference/telegraf-plugins/#input-plugins"
+              href="https://docs.influxdata.com/telegraf/v1.17/plugins/#input-plugins"
```
sanderson, line 64: Not a show-stopper, but you may update this to `https://docs.influxdata.com/telegraf/latest/plugins/#input-plugins` just so it doesn't need to be updated for each Telegraf release.

```suggestion
              href="https://docs.influxdata.com/telegraf/latest/plugins/#input-plugins"
```

sanderson (approved):
One minor suggestion. Feel free to ignore. LGTM.

ui/src/dataLoaders/components/collectorsWizard/select/SelectCollectorsStep.tsx
danxmoran: Would it make sense to do the same for all the links that start with `docs.influxdata/com/influxdb/v2.0`?

sanderson: This is where it gets tricky. The InfluxDB docs are versioned per minor version of InfluxDB, so it makes sense to leave these as-is and then update them to v2.1 whenever 2.1 rolls out. That way the links in the UI point to the specific version of InfluxDB the user is running. The tricky part is that this same code is used (afaik) in the Cloud UI, so links in cloud are actually pointing to OSS docs. So in short... I don't know 🤷🏻‍♂️ 

In an ideal world, we'd be able to have these links be version/context specific.

danxmoran: What I'll do is:
* Update the links on `master` to use `latest`
* When I backport this to the `2.0` branch, keep the links pointing at `2.0`

PR #20763, opened 2021-02-16
feat(cmd/influx): support reading GZIP data in `influx write`

### LINKED ISSUE #20659: Suppport for ingesting compressed files via `influx write`
__Proposal:__
Allow `influx write` to read compressed LP / CSV files from disk.

__Current behavior:__
`influx write` only supports reading uncompressed files.

__Desired behavior:__
Passing a gzip file to `influx write` Just Works. Some details:
* gzip is auto-detected for files with extension `.gz`
* Auto-detection runs before the CSV vs. LP detection, so `.csv.gz` should work
* A new flag can be used to force gzip or not

__Use case:__
Make it easier for customers to onboard large amounts of data at a time.

danxmoran:
Closes #20659

This is meant to mirror the implementation of the existing CSV support.
* A new CLI option allows for explicitly specifying compression of all inputs
* If the flag isn't set, the line-reader checks the names of input files & URLs to detect GZIP
   * Files or URLs ending in `.gz` enable decompression
   * URLs that respond with `Content-Encoding: gzip` enable decompression

Unlike the existing CSV impl, gzip-or-not is handled on an input-by-input basis when the CLI option isn't set.

### DIFF
--- a/cmd/influx/write.go
+++ b/cmd/influx/write.go
@@ -56,7 +56,6 @@ type writeFlagsBuilder struct {
 	ErrorsFile                 string
 	RateLimit                  string
 	Compression                string
-	ErrorThreshold             uint64
 }
 
 func newWriteFlagsBuilder(svcFn buildWriteSvcFn, f *globalFlags, opt genericCLIOpts) *writeFlagsBuilder {
--- a/cmd/influx/write.go
+++ b/cmd/influx/write.go
@@ -130,7 +130,7 @@ func (b *writeFlagsBuilder) cmd() *cobra.Command {
 	cmd.PersistentFlags().StringVar(&b.Encoding, "encoding", "UTF-8", "Character encoding of input files or stdin")
 	cmd.PersistentFlags().StringVar(&b.ErrorsFile, "errors-file", "", "The path to the file to write rejected rows to")
 	cmd.PersistentFlags().StringVar(&b.RateLimit, "rate-limit", "", "Throttles write, examples: \"5 MB / 5 min\" , \"17kBs\". \"\" (default) disables throttling.")
-	cmd.PersistentFlags().BoolVar(&b.Compressed, "compressed", false, "Indicates that the file is compressed")
+	cmd.PersistentFlags().BoolVar(&b.Compressed, "compressed", false, "Indicates that the input is GZIP-compressed. Defaults to false unless a '.gz' extension")
 
 	cmdDryRun := b.newCmd("dryrun", b.writeDryrunE, false)
 	cmdDryRun.Args = cobra.MaximumNArgs(1)
--- a/cmd/influx/write.go
+++ b/cmd/influx/write.go
@@ -2,13 +2,11 @@ package main
 
 import (
 	"compress/gzip"
-	_ "compress/gzip"
 	"context"
 	"encoding/csv"
 	"fmt"
 	"io"
 	"log"
-	"math"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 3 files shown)


cmd/influx/write.go
```diff
@@ -52,6 +53,8 @@ type writeFlagsBuilder struct {
 	Encoding                   string
 	ErrorsFile                 string
 	RateLimit                  string
+	Compressed                 bool
```
danxmoran, line 56: I think this needs to become `*bool` so we can distinguish not-set from a true `false`

danxmoran:
Closing this for now to do a bit more cleanup / refactoring

cmd/influx/write.go
danxmoran: Made it a `string` instead to match the existing input-format option

cmd/influx/write.go
```diff
@@ -159,13 +164,33 @@ func (b *writeFlagsBuilder) createLineReader(ctx context.Context, cmd *cobra.Com
...
 	// validate and setup decoding of files/stdin if encoding is supplied
 	decode, err := csv2lp.CreateDecoder(b.Encoding)
 	if err != nil {
 		return nil, csv2lp.MultiCloser(closers...), err
 	}
 
+	// utility to manage common steps used to decode / decompress input sources,
+	// while tracking resources that must be cleaned-up after reading.
+	addReader := func(r io.Reader, name string, compressed bool) error {
+		r = decode(r)
```
lesam, line 181: Shouldn't we decode the character set after decompressing??

cmd/influx/write_test.go
```diff
@@ -120,24 +156,157 @@ func Test_writeFlags_createLineReader(t *testing.T) {
...
+			stdIn: stdInLpGzipContents,
+			lines: []string{
+				stdInLpContents,
+			},
+		},
+		{
+			name:      "read LP data from stdin using '-' argument",
+			flags:     writeFlagsBuilder{},
+			stdIn:     strings.NewReader(stdInLpContents),
+			arguments: []string{"-"},
```
lesam, line 224: Do we allow something like `influx write --compression gzip -` to read from stdin? Should we test that case?

cmd/influx/write.go
danxmoran: I guess we should add a test that actually uses the encoding arg 🤦

danxmoran: Fixed, and test added

cmd/influx/write_test.go
danxmoran: Test added

cmd/influx/write_test.go
```diff
@@ -120,24 +156,191 @@ func Test_writeFlags_createLineReader(t *testing.T) {
...
+			lines: []string{
+				stdInLpContents,
+			},
+		},
+		{
+			name: "read compressed LP data from stdin using '-' argument",
```
danxmoran, line 253: Ugh, this one passes if run on its own, but fails when all tests run... investigating

danxmoran: Fixed (locally at least)

cmd/influx/write.go
```diff
@@ -184,10 +208,18 @@ func (b *writeFlagsBuilder) createLineReader(ctx context.Context, cmd *cobra.Com
 				return nil, csv2lp.MultiCloser(closers...), fmt.Errorf("failed to open %q: %v", file, err)
 			}
 			closers = append(closers, f)
-			readers = append(readers, decode(f), strings.NewReader("\n"))
+
+			compressed := b.Compression == "gzip" || (len(b.Compression) == 0 && strings.HasSuffix(file, ".gz"))
+			if compressed {
+				file = strings.TrimSuffix(file, ".gz")
```
lesam, line 214: As a user I would prefer to see the `.gz` suffix in the error message. Not blocking.

lesam (approved):
lgtm

PR #20767, opened 2021-02-16
chore: Upgrade 1.x to the latest flux

### LINKED ISSUE #20766: Upgrade to latest flux for InfluxDB 1.9
__Proposal:__
We should upgrade flux to the latest available version

__Current behavior:__
Flux in InfluxDB master-1.x is on version 0.65.0

__Desired behavior:__
Flux should be upgraded to the latest version (0.105.0)

__Alternatives considered:__
N/A

__Use case:__
Users should be able to use documentation for the latest flux release.

lesam:
Closes #20766

Upgrade to the latest flux

- [ ] Rebased/mergeable
- [ ] Tests pass


lesam:
This is very much in draft state - it compiles, but doesn't work yet.

flux/client/http.go
```diff
@@ -90,57 +55,4 @@ func newClient(scheme string, insecure bool) *http.Client {
...
-		qr.Dialect.CommentPrefix = "#"
-		qr.Dialect.DateTimeFormat = "RFC3339"
-		qr.Dialect.Annotations = d.ResultEncoderConfig.Annotations
-	default:
-		return nil, fmt.Errorf("unsupported dialect %T", d)
-	}
-
-	return qr, nil
-}
+}
```
docmerlin, line 58: add a newline here

flux/control/controller.go
```diff
@@ -40,7 +41,7 @@ func (c *Controller) Query(ctx context.Context, compiler flux.Compiler) (flux.Qu
 		ctx = dep.Inject(ctx)
 	}
 
-	p, err := compiler.Compile(ctx)
+	p, err := compiler.Compile(ctx, runtime.Default)
```
docmerlin, line 44: this doesn't seem right to me, I believe we have to inject some packages here, but I am not certain.

flux/stdlib/influxdata/influxdb/rules.go
```diff
@@ -1,6 +1,10 @@
 package influxdb
 
 import (
+	"context"
+	"github.com/influxdata/influxdb/models"
+	"github.com/influxdata/influxdb/storage/reads/datatypes"
+
 	"github.com/influxdata/flux"
```
docmerlin, line 8: remove space on line 7 and add space between 4 and 5

flux/stdlib/influxdata/influxdb/storage.go
```diff
@@ -3,11 +3,11 @@ package influxdb
 import (
 	"context"
 	"fmt"
+	"github.com/influxdata/influxdb/storage/reads/datatypes"
 
 	"github.com/influxdata/flux"
```
docmerlin, line 8: spacing.

flux/stdlib/influxdata/influxdb/v1/databases.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/influxdata/flux/runtime"
 
 	"github.com/influxdata/flux"
```
docmerlin, line 9: spacing again.

go.mod
```diff
@@ -3,20 +3,30 @@ module github.com/influxdata/influxdb
...
+	github.com/denisenkom/go-mssqldb v0.0.0-20200428022330-06a60b6afbbc // indirect
 	github.com/dgrijalva/jwt-go/v4 v4.0.0-preview1
 	github.com/dgryski/go-bitstream v0.0.0-20180413035011-3522498ce2c8
+	github.com/foxcpp/go-mockdns v0.0.0-20201212160233-ede2f9158d15 // indirect
+	github.com/gofrs/uuid v3.3.0+incompatible // indirect
 	github.com/gogo/protobuf v1.3.1
 	github.com/golang/snappy v0.0.1
 	github.com/google/go-cmp v0.5.0
-	github.com/influxdata/flux v0.65.0
+	github.com/influxdata/flux v0.105.1
```
docmerlin, line 27: woot

docmerlin (dismissed):
I didn't see anything glaring, just some typographical stuff.   I'm also not sure about the package injection.

docmerlin:
Good first pass.  Some of the typoes are just results of not using `goimports -w ./`

lesam:
@stuartcarnie I think you had said that inmem indexing would not work with the latest flux, but I'm not sure why this would be the case.

flux/client/http.go
```diff
@@ -46,40 +34,17 @@ func NewHTTP(u url.URL) (*HTTP, error) {
...
-	}
+func (s *HTTP) Do(hreq *http.Request) (*http.Response, error) {
 	if s.Username != "" {
 		hreq.SetBasicAuth(s.Username, s.Password)
 	}
 
 	hreq.Header.Set("Content-Type", "application/json")
 	hreq.Header.Set("Accept", "text/csv")
-	hreq = hreq.WithContext(ctx)
+	// ?? hreq = hreq.WithContext(ctx)
```
lesam, line 44: TODO: need to understand whether this client can be canceled / if it needs a cancelable context

flux/control/controller.go
lesam: I think that since we are using the global injection method we should be using the global runtime.

pkg-config.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
lesam, line 1: Stolen from the flux repo

wolffcm (approved):
This all looks good to me. In OSS 2.0 and cloud we have a suite of Flux scripts that we run against real InfluxDB storage. Unfortunately they have not been ported here--that would be a bit of work. I would definitely do some manual testing to make sure that queries work right, with an emphasis on `from` and `to`. I'm happy to chat about this.

I'm guessing that a lot of these changes were very similar to the upgrade of Flux that happened in September (I think) to bring it back up to the latest Flux?

stuartcarnie:
> @stuartcarnie I think you had said that inmem indexing would not work with the latest flux, but I'm not sure why this would be the case.

Do you have context for where I mentioned that? The two indices publish the same shared `Index` interface API, so I would expect it works.

flux/client/http.go
lesam: Figured it out I think - this is fine because we set the context on the request in flux: https://github.com/influxdata/flux/blob/master/stdlib/http/post.go#L73

lesam:
> > @stuartcarnie I think you had said that inmem indexing would not work with the latest flux, but I'm not sure why this would be the case.
> 
> Do you have context for where I mentioned that? The two indices publish the same shared `Index` interface API, so I would expect it works.

No I just have rumours from within the team, definitely makes the job easier that we can keep inmem.

flux/client/http.go
lesam: Though actually this appears to be for making a new http call, not the initial one to flux. So maybe it needs a re-look. Will make sure I can hit it in tests and understand what's up with it.

cmd/influx/cli/cli.go
```diff
@@ -211,12 +211,8 @@ func (c *CommandLine) Run() error {
...
 
 	if c.Type == QueryLanguageFlux {
-		repl, err := getFluxREPL(c.URL, c.ClientConfig.Username, c.ClientConfig.Password)
-		if err != nil {
-			return err
-		}
-		repl.Run()
-		os.Exit(0)
+		// TODO(lesam): steal 2.x flux client
+		return fmt.Errorf("ERROR: flux repl missing due to flux upgrade")
```
lesam, line 215: Note these errors - this will be a pretty major change to the `influx` client for 1.x flux queries, as we will be removing the REPL. We should obviously fix this TODO in the next week or two, before 1.9 releases.

flux/stdlib/influxdata/influxdb/to.go
```diff
@@ -1,4 +1,353 @@
...
+	"github.com/influxdata/influxql"
+	"time"
+)
+
+const (
+	ToKind = "influx1x/toKind"
+	DefaultBufferSize = 5000
+)
+
+type ToOpSpec struct {
```
lesam, line 26: right now `to` turns all string columns into keys - this is bad and should be fixed as a follow-up

flux/stdlib/influxdata/influxdb/to.go
```diff
@@ -1,4 +1,353 @@
...
+	}, nil
+}
+
+func (t *ToTransformation) RetractTable(id execute.DatasetID, key flux.GroupKey) error {
+	return t.d.RetractTable(key)
+}
+
+// Process does the actual work for the ToTransformation.
+func (t *ToTransformation) Process(id execute.DatasetID, tbl flux.Table) error {
+	//TODO(lesam): this is where 2.x overrides with explicit tag columns
```
lesam, line 167: Note the todo

flux/stdlib/influxdata/influxdb/to.go
```diff
@@ -1,4 +1,353 @@
...
+			if measurementName == "" {
+				return &flux.Error{
+					Code: codes.Invalid,
+					Msg: fmt.Sprintf("no column with label %s exists", measurementColumn),
+				}
+			}
+
+			var fieldValues values.Object
+			var err error
+			// TODO(lesam): this is where we would support the `fn` argument to `to`
```
lesam, line 262: Need this too for 1.9

docmerlin (approved):
@lesam and I features that need to be added and a couple changes to be made before we release 1.9, but this PR is being approved.

luismedicina23:
@lesam 

I'm new and Brazilian 

So it's hard for me
I'm sorry

luismedicina23 (commented):
@lesam

cmd/influx/cli/cli.go
luismedicina23:
```

Dockerfile_build_ubuntu32

```

luismedicina23: @lesam 

Dockerfile_build_ubuntu32

luismedicina23 (commented):
Dockerfile_build_ubuntu64_go1.13

luismedicina23 (commented):
Dockerfile_build_ubuntu64_go1.13

luismedicina23 (commented):
Dockerfile_build_ubuntu32

luismedicina23 (commented):
@Maintenance

PR #20770, opened 2021-02-18
chore: clean up unused fields in FileSet

foobar:
`levels` and  `manifestSize` are not used

### DIFF
--- a/tsdb/index/tsi1/file_set.go
+++ b/tsdb/index/tsi1/file_set.go
@@ -15,27 +15,21 @@ import (
 
 // FileSet represents a collection of files.
 type FileSet struct {
-	levels       []CompactionLevel
-	sfile        *tsdb.SeriesFile
-	files        []File
-	manifestSize int64 // Size of the manifest file in bytes.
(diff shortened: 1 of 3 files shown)


foobar:
@danxmoran can you review this pr?

danxmoran:
@foobar yes, I'll take a look

tsdb/index/tsi1/file_set.go
```diff
@@ -88,7 +81,7 @@ func (fs *FileSet) Size() int64 {
 	for _, f := range fs.files {
 		total += f.Size()
 	}
-	return total + int64(fs.manifestSize)
```
danxmoran, line 91: I'm not sure about this one (the fact that `manifestSize` is always `0` could be a bug). Will check internally with teammates who know this code better

danxmoran (commented):
Overall I like this PR, but I need to double-check that these fields are really useless

danxmoran (approved):
Thanks for the cleanup @foobar

Issue #20452, opened 2021-01-05
retentionPeriodHrs key in onboarding requests is misleading


jeremy-prater:
__Steps to reproduce:__
List the minimal actions needed to reproduce the behavior.

1. Create a new influxdb
2. Verify the onboarding status is true (ready for onboarding)
3. Using the C# client library v1.13.0 send a Onboarding request with a non-zero retention policy

__Expected behavior:__
Successful onboarding response

__Actual behavior:__
Onboarding fails with `Message: retention policy duration must be at least 1h0m0s`

__Environment info:__

* System info: Darwin 19.6.0 x86_64
* InfluxDB version: InfluxDB 2.0.3 (git: fe04d346df) build_date: 2020-12-15T01:00:16Z
* Other relevant environment details: None

__Config:__
Config is default

__Code:__

```
OnboardingRequest onboardingRequest = new OnboardingRequest(
    adminUsername,
    adminPassword,
    defaultOrg,
    defaultBucket,
    24 * 30 // 30 days
);
var onboardResponse = await influxDBClient.OnboardingAsync(onboardingRequest);
```

__Logs:__
```
ts=2021-01-05T22:59:11.445554Z lvl=debug msg="Onboarding eligibility check finished" log_id=0RXCFZaG000 result=true
ts=2021-01-05T22:59:11.445675Z lvl=debug msg=Request log_id=0RXCFZaG000 service=http method=GET host=localhost:8086 path=/api/v2/setup query= proto=HTTP/1.1 status_code=200 response_size=20 content_length=0 referrer= remote=[::1]:65335 user_agent=influxdb-client-csharp took=0.443ms body=
ts=2021-01-05T22:59:11.527917Z lvl=debug msg="user create" log_id=0RXCFZaG000 store=new took=15.825ms
ts=2021-01-05T22:59:11.626550Z lvl=debug msg="set password" log_id=0RXCFZaG000 store=new took=98.599ms
ts=2021-01-05T22:59:11.646952Z lvl=debug msg="org find by ID" log_id=0RXCFZaG000 store=new took=0.073ms
ts=2021-01-05T22:59:11.670368Z lvl=debug msg="bucket create" log_id=0RXCFZaG000 store=new took=23.487ms
ts=2021-01-05T22:59:11.690045Z lvl=debug msg="org find by ID" log_id=0RXCFZaG000 store=new took=0.064ms
[...]

PR #20794, opened 2021-02-22
refactor: reduce duplication in setup logic

### LINKED ISSUE #20326: Reduce duplication of onboarding path in CLI commands
We have nearly-identical code for initial DB setup in the implementations of `influx setup` and `influxd upgrade`. Inside both of those, we have another layer of duplication between "interactive" and "noninteractive" paths. We should DRY up the logic (within reason) to avoid having to touch 4 different code-paths when we want to fix or change logic in those CLI commands.

danxmoran:
Closes #20326 

Merges the "interactive" and "noninteractive" branches in the `influx setup` and `influxd upgrade` commands to reduce duplication _within_ each command. The goal is to do this before #20452 so we only need to change how we handle retention policies in 2 places (instead of 4).

There is much more room to de-duplicate logic _across_ the two commands, but I don't think it makes sense to take those steps when we're planning to split the `influx` CLI into its own repo. Better for now to keep the diff small and the commands as independent as possible.


davidby-influx (approved):
Looks good to me. I like the use of intermediate variables to clarify code.

PR #20172, opened 2020-11-25
fix(tsm1): fix ring and related test

StoneYunZhao:
I accidentally found that ring.go is not rigorous。I opened two identical PRs before, one of the PR's circleci test failed but another succeed.

1. [failed circleci](https://app.circleci.com/pipelines/github/influxdata/influxdb/15858/workflows/0d1ce4c9-f0c6-400a-8c65-a26a28df0fd7/jobs/158630) triggered by #20117 .
2. [succeed circleci](https://app.circleci.com/pipelines/github/influxdata/influxdb/16024/workflows/b3f7375b-4c3c-46ef-9825-92b8266932a0/jobs/159176) triggered by #20118 .

I'm pretty sure that it's not my commit who caused the failure after read ring.go. I did some fix work for it.

1. [e9db11a](https://github.com/influxdata/influxdb/commit/e9db11a) changed the max partition to 16, but not change benchmark test, so the benchmark test will failed if the partition larger than 16.
2. `newring()` doesn't check the partition if is the power of two, and `TestRing_newRing` can't test that case.
3. `add()` should not add a nil entry in `benchmarkRingkeys`, because it will cause panic when call `keys()`.
4. reset keysHint must use atomic in `reset()`. If not, it will cause the below failure at a very small chance.
5. `remove()` is safe for use by multiple goroutines, so we must ensure `keysHint` not be negative in `remove()`. If negative , may cause panic because it used in the way `make([][]byte, 0, atomic.LoadInt64(&r.keysHint))`

Below is the error log in circleci: 

```=== RUN   TestShard_WritePoints_FieldConflictConcurrent
==================
WARNING: DATA RACE
Write at 0x00c00039c300 by goroutine 90:
  github.com/influxdata/influxdb/tsdb/engine/tsm1.(*ring).reset()
      /root/influxdb/tsdb/engine/tsm1/ring.go:81 +0xac
  github.com/influxdata/influxdb/tsdb/engine/tsm1.(*Cache).ClearSnapshot()
      /root/influxdb/tsdb/engine/tsm1/cache.go:449 +0x3b1
  github.com/influxdata/influxdb/tsdb/engine/tsm1.(*Engine).WriteSnapshot()
      /root/influxdb/tsdb/engine/tsm1/engine.go:1934 +0x5e2
  github.com/influxdata/influxdb/tsdb/engine/tsm1.(*Engine).CreateSnapshot()
      /root/influxdb/tsdb/engine/tsm1/engine.go:1959 +0xbd
[...]

PR #20802, opened 2021-02-24
fix(tsm1): fix data race and validation in cache ring

danxmoran:
Backport of #20797 (which was a forward-port++ of #20172)


lesam (approved):
lgtm

PR #20803, opened 2021-02-24
refactor: Remove unused function add and unused variable keysHint

lesam:
Very small refactor so no tracking issue

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/tsdb/engine/tsm1/ring.go
+++ b/tsdb/engine/tsm1/ring.go
@@ -3,7 +3,6 @@ package tsm1
 import (
 	"fmt"
 	"sync"
-	"sync/atomic"
 
 	"github.com/cespare/xxhash"
 	"github.com/influxdata/influxdb/v2/pkg/bytesutil"
@@ -32,12 +31,6 @@ const partitions = 16
 // key is hashed and the first 8 bits are used as an index to the ring.
 //
 type ring struct {
-	// Number of keys within the ring. This is used to provide a hint for
-	// allocating the return values in keys(). It will not be perfectly accurate
-	// since it doesn't consider adding duplicate keys, or trying to remove non-
-	// existent keys.
-	keysHint int64
-
 	// The unique set of partitions in the ring.
 	// len(partitions) <= len(continuum)
 	partitions []*partition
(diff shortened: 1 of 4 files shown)


tsdb/engine/tsm1/ring.go
```diff
@@ -103,28 +95,16 @@ func (r *ring) write(key []byte, values Values) (bool, error) {
 	return r.getPartition(key).write(key, values)
 }
 
-// add adds an entry to the ring.
-func (r *ring) add(key []byte, entry *entry) {
-	r.getPartition(key).add(key, entry)
-	atomic.AddInt64(&r.keysHint, 1)
```
lesam, line 109: Note that since add is removed (it was unused in production code), keysHint is always 0

e-dard:
This seems fine to me correctness wise but I'm surprised that the `ring.add` call is never used. I would be even more surprised if it _has_ never been used, rather than being removed through some refactor.

The original reason I added these sizing hints was to improve `tsm1.Cache` snapshotting and write performance by reducing allocations. Typically in a running system the same type of data gets continuously written and so if you pre-allocate bits of the cache in the same way after a snapshot you can minimise the number of allocations.

If `add` was just removed as a refactor somewhere I wonder about putting it back so that `keysHint` can become useful again. Having said that, a lot changes in four years and the cache as it runs today may not benefit from these pre-allocation hints anyway 🤷

e-dard (commented):
This seems fine to me correctness wise but I'm surprised that the ring.add call is never used. I would be even more surprised if it has never been used, rather than being removed through some refactor.

The original reason I added these sizing hints was to improve tsm1.Cache snapshotting and write performance by reducing allocations. Typically in a running system the same type of data gets continuously written and so if you pre-allocate bits of the cache in the same way after a snapshot you can minimise the number of allocations.

If add was just removed as a refactor somewhere I wonder about putting it back so that keysHint can become useful again. Having said that, a lot changes in four years and the cache as it runs today may not benefit from these pre-allocation hints anyway 🤷

dgnorton (approved):

luismedicina23:
@Main
- Remove

PR #20811, opened 2021-02-28
fix(tsm1): fix wal's totalOldDiskSize statistics

StoneYunZhao:
The value of the totalOldDiskSize in WAL is wrong. There are two reason caused the inaccuracy.

Reason 1 caused great inaccuracies: TotalOldDiskSize should add rather than set the current segment bytes when rolling a new segment.

Reason 2 caused minor inaccuracies: Must skip the current segment size when statistic the totalOldDiskSize, because total disk size is computed by following method:
```
func (l *WAL) DiskSizeBytes() int64 {
	return atomic.LoadInt64(&l.stats.OldBytes) + atomic.LoadInt64(&l.stats.CurrentBytes)
}
```

- [x] Rebased/mergeable
- [x] Tests pass


tsdb/engine/tsm1/wal.go
```diff
@@ -242,6 +242,11 @@ func (l *WAL) Open() error {
 			}
 		}
 	}
+
+	if l.currentSegmentWriter != nil {
+		totalOldDiskSize -= int64(l.currentSegmentWriter.size)
+	}
+
 	atomic.StoreInt64(&l.stats.OldBytes, totalOldDiskSize)
```
danxmoran, line 250: Could you add a regression test that asserts on the value of `stats.OldBytes`?

tsdb/engine/tsm1/wal.go
```diff
@@ -379,6 +384,11 @@ func (l *WAL) Remove(files []string) error {
 
 		totalOldDiskSize += stat.Size()
 	}
+
+	if l.currentSegmentWriter != nil {
+		totalOldDiskSize -= int64(l.currentSegmentWriter.size)
+	}
+
 	atomic.StoreInt64(&l.stats.OldBytes, totalOldDiskSize)
```
danxmoran, line 392: Could you add a test?

tsdb/engine/tsm1/wal.go
```diff
@@ -565,7 +575,8 @@ func (l *WAL) newSegmentFile() error {
 		if err := l.currentSegmentWriter.close(); err != nil {
 			return err
 		}
-		atomic.StoreInt64(&l.stats.OldBytes, int64(l.currentSegmentWriter.size))
+
+		atomic.AddInt64(&l.stats.OldBytes, int64(l.currentSegmentWriter.size))
```
danxmoran, line 579: Could you add a test?

danxmoran (commented):
I'm not super familiar with this code, so feel free to point out if testing any of the changes would be incredibly hard

StoneYunZhao:
> ree to point out if

Tests added.

I compared monitor data from measurement `influxdb_tsm1_wal` scraped by telegraf and disk usage of shard directory under wal directory from Linux OS, and confirmed that this error exists.

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
 	}
 }
 
+func TestWAL_DistSize(t *testing.T) {
```
danxmoran, line 706:
```suggestion
func TestWAL_DiskSize(t *testing.T) {
```

tsdb/engine/tsm1/wal_test.go
```diff
@@ -6,11 +6,13 @@ import (
 	"io/ioutil"
 	"os"
 	"reflect"
+	"sort"
 	"testing"
 
 	"github.com/golang/snappy"
 	"github.com/influxdata/influxdb/v2/pkg/slices"
 	"github.com/influxdata/influxdb/v2/tsdb/engine/tsm1"
+	"github.com/stretchr/testify/assert"
```
danxmoran, line 15:
```suggestion
	"github.com/stretchr/testify/require"
```
IMO makes the testing simpler, since it bails at the first mismatched expectation

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
 	}
 }
 
+func TestWAL_DistSize(t *testing.T) {
+	test := func(w *tsm1.WAL, oldZero, curZero bool) () {
+		// get disk size by reading file
+		files, err := ioutil.ReadDir(w.Path())
+		if err != nil {
```
danxmoran, line 710: This `if` could be replaced by `require.NoError`

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
...
+		if !oldZero && old == 0 {
+			t.Fatalf("expect old disk size larger than 0, got 0")
+		}
+
+		if curZero && cur > 0 {
+			t.Fatalf("expect current disk size: 0, got: %d", cur)
+		}
+		if !curZero && cur == 0 {
+			t.Fatalf("expect current disk size larger than 0, got 0")
+		}
```
danxmoran, line 740: These `if`s could all be replaced by `require` functions

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
...
+
+		// test method DiskSizeBytes
+		if got, exp := w.DiskSizeBytes(), old+cur; got != exp {
+			t.Fatalf("expect total disk size: %d, got: %d", exp, got)
+		}
```
danxmoran, line 745: Same here (`require.Equal` I think)

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
...
+		// test Statistics
+		ss := w.Statistics(nil)
+		assert.Equal(t, 1, len(ss))
+
+		m := ss[0].Values
+		assert.NotNil(t, m)
+
+		if got, exp := m["oldSegmentsDiskBytes"].(int64), old; got != exp {
+			t.Fatalf("expect old disk size: %d, got: %d", exp, got)
+		}
```
danxmoran, line 756: `require.Equal`

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
...
+
+		// test method DiskSizeBytes
+		if got, exp := w.DiskSizeBytes(), old+cur; got != exp {
+			t.Fatalf("expect total disk size: %d, got: %d", exp, got)
+		}
+
+		// test Statistics
+		ss := w.Statistics(nil)
+		assert.Equal(t, 1, len(ss))
```
danxmoran, line 749:
```suggestion
		require.Equal(t, 1, len(ss))
```

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
...
+
+		m := ss[0].Values
+		assert.NotNil(t, m)
```
danxmoran, line 752:
```suggestion
		require.NotNil(t, m)
```

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
...
+
+		if got, exp := m["oldSegmentsDiskBytes"].(int64), old; got != exp {
+			t.Fatalf("expect old disk size: %d, got: %d", exp, got)
+		}
+
+		if got, exp := m["currentSegmentDiskBytes"].(int64), cur; got != exp {
+			t.Fatalf("expect current disk size: %d, got: %d", exp, got)
+		}
```
danxmoran, line 760: `require.Equal`

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
...
+
+	w := tsm1.NewWAL(dir)
+
+	const segSize = 1024
+	w.SegmentSize = segSize
+
+	// open
+	if err := w.Open(); err != nil {
+		t.Fatalf("error opening WAL: %v", err)
+	}
```
danxmoran, line 774: `require.NoError`

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
...
+			fatal(t, "write points", err)
+		}
+	}
+
+	test(w, false, false)
+
+	// reopen
+	if err := w.Close(); err != nil {
+		t.Fatalf("error cloing WAL: %v", err)
+	}
```
danxmoran, line 803: `require.NoError`

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
...
+
+	if err := w.Open(); err != nil {
+		t.Fatalf("error open WAL: %v", err)
+	}
```
danxmoran, line 807: `require.NoError`

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
...
+
+	test(w, false, false)
+
+	// remove
+	closedSegments, err := w.ClosedSegments()
+	if err != nil {
+		t.Fatalf("get closed segment error: %v", err)
+	}
```
danxmoran, line 815: `require.NoError`

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
...
+
+	if err := w.Remove(closedSegments); err != nil {
+		t.Fatalf("close segment error: %v", err)
+	}
```
danxmoran, line 819: `require.NoError`

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("read dir err: %v", err)
+		}
+
+		sort.Slice(files, func(i, j int) bool {
+			return files[i].Name() < files[j].Name()
+		})
+
+		var old, cur int64
+		for i, f := range files {
```
danxmoran, line 719: This loop confuses me. Is the end goal to have:
```
cur = // size of last file
old = // sum of sizes of all but last file
```
?

danxmoran: If so, would we assign `cur` directly outside of the loop, and have the `range` only loop over the files we want to sum?

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
...
+			t.Fatalf("expect current disk size larger than 0, got 0")
+		}
+
+		// test method DiskSizeBytes
+		if got, exp := w.DiskSizeBytes(), old+cur; got != exp {
+			t.Fatalf("expect total disk size: %d, got: %d", exp, got)
+		}
+
+		// test Statistics
+		ss := w.Statistics(nil)
```
danxmoran, line 748: What is the `nil` here?

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
...
+		"cpu,host=C#!~#value": {tsm1.NewValue(1, 1.0)},
+	}
+
+	if _, err := w.WriteMulti(values); err != nil {
+		fatal(t, "write points", err)
+	}
+
+	test(w, true, false)
+
+	// write with rollSegment
```
danxmoran, line 791: Not super familiar with this code, so I'm confused by this comment when I don't see `rollSegment` below. Can you clarify? Something like:
```suggestion
	// write enough points to force the segment to roll
```

tsdb/engine/tsm1/wal_test.go
```diff
@@ -701,6 +703,124 @@ func TestWALRollSegment(t *testing.T) {
...
+	w.SegmentSize = segSize
+
+	// open
+	if err := w.Open(); err != nil {
+		t.Fatalf("error opening WAL: %v", err)
+	}
+
+	test(w, true, true)
+
+	// write without rollSegment
```
danxmoran, line 778:
```suggestion
	// write some points (not enough to force the segment to roll)
```

danxmoran (commented):
The test makes sense to me, and I'm :+1: to introducing `testify`. Some style comments about cleaning up the test

tsdb/engine/tsm1/wal_test.go
StoneYunZhao: `nil` means no need to add additional tags

StoneYunZhao:
> The test makes sense to me, and I'm 👍 to introducing `testify`. Some style comments about cleaning up the test

Thank you very much for your suggestions, the code is indeed a lot concise.

danxmoran (approved):

PR #20814, opened 2021-03-01
fix(tsm1): limit concurrent WAL encodings to reduce memory pressure under heavy write load

StoneYunZhao:
The comment in `writeToLog` method in wal.go says:
> limit how many concurrent encodings can be in flight.  Since we can only write one at a time to disk, a slow disk can cause the allocations below to increase quickly.  If we're backed up, wait until others have completed.

The code will not work as described, because the `bytesPool` of type `LimitedBytes` would never block when call `Get` method under any conditions. The word 'limit' in `LimitedBytes` means limiting the max size of byte slice instead of the max number of byte slice. So the previous code can cause the memory allocations increase quickly when use a slow disk.

- [ ] Rebased/mergeable
- [ ] Tests pass


tsdb/series_partition.go
```diff
@@ -88,7 +88,7 @@ func (p *SeriesPartition) Open() error {
 		p.index = NewSeriesIndex(p.IndexPath())
 		if err := p.index.Open(); err != nil {
 			return err
-		} else if p.index.Recover(p.segments); err != nil {
+		} else if err := p.index.Recover(p.segments); err != nil {
```
StoneYunZhao, line 91: Here is nothing to do with the PR. As this is such a small typo, I added it incidentally.

danxmoran: Thank you!

tsdb/engine/tsm1/wal.go
```diff
@@ -112,7 +112,9 @@ type WAL struct {
 	SegmentSize int
 
 	// statistics for the WAL
-	stats   *WALStatistics
+	stats *WALStatistics
+
+	// limiter limits the max concurrency of waiting WAL writes.
 	limiter limiter.Fixed
```
danxmoran, line 118: For anyone else looking, this was apparently unused 🤦

tsdb/engine/tsm1/wal.go
```diff
@@ -399,6 +401,9 @@ func (l *WAL) writeToLog(entry WALEntry) (int, error) {
 	// limit how many concurrent encodings can be in flight.  Since we can only
 	// write one at a time to disk, a slow disk can cause the allocations below
 	// to increase quickly.  If we're backed up, wait until others have completed.
+	l.limiter.Take()
+	defer l.limiter.Release()
```
danxmoran, line 405: Hmmm it looks like these statements were originally in the code, then commented out in d7c5dd0a3e2bdcaa669d2d7ece8bd76ef07b075a and eventually removed. Putting them back makes sense to me given the comment, but I'll need to call in another reviewer with more context.

@StoneYunZhao have you seen memory pressure in this code that's been fixed by your changes here? Or did you notice the comment and impl didn't align while reading through the code?

StoneYunZhao: My situation belongs to the latter.

StoneYunZhao: When there is a peak write traffic, I see that the memory will increase a lot. But I am not sure if it is caused by this code, It may be `Cache` in `tsdb/engine/tsm1/cache.go`, or it may be both.

StoneYunZhao: > Hmmm it looks like these statements were originally in the code, then commented out in [d7c5dd0](https://github.com/influxdata/influxdb/commit/d7c5dd0a3e2bdcaa669d2d7ece8bd76ef07b075a) and eventually removed. Putting them back makes sense to me given the comment, but I'll need to call in another reviewer with more context.
> 
> @StoneYunZhao have you seen memory pressure in this code that's been fixed by your changes here? Or did you notice the comment and impl didn't align while reading through the code?

I think the purpose of https://github.com/influxdata/influxdb/commit/d7c5dd0a3e2bdcaa669d2d7ece8bd76ef07b075a is to reduce the frequency of goroutine creation and destruction, but I don't understand why the limiter related code is deleted.

danxmoran (commented):
Apologies for the long delay in reviewing this, something urgent came up on our end so the 2nd reviewer hasn't been able to take a look yet.

In the meantime, I looked through `wal_test.go` and noticed that most of the existing tests & benchmarks skip `WriteMulti` (and therefore `writeToLog`) entirely, in favor of using low-level segment writers/readers. Could you refactor the existing tests with `WriteMulti` in the name to actually use that method, and add a benchmark so we can compare perf with and without the limiter?

StoneYunZhao:
> Apologies for the long delay in reviewing this, something urgent came up on our end so the 2nd reviewer hasn't been able to take a look yet.
> 
> In the meantime, I looked through `wal_test.go` and noticed that most of the existing tests & benchmarks skip `WriteMulti` (and therefore `writeToLog`) entirely, in favor of using low-level segment writers/readers. Could you refactor the existing tests with `WriteMulti` in the name to actually use that method, and add a benchmark so we can compare perf with and without the limiter?

Test refactored and benchmark added.

Below are the benchmark results. High memory consumption without limiter.
> My machine’s hard disk is an SSD. If the disk is slower, I think the memory consumption will be higher.

| concurrency | with limiter ns/op | without limiter ns/op | with limiter B/op | without limiter B/op | with limiter allocs/op | without limiter allocs/op |
| ----------- | ------------------ | --------------------- | --------------------- | -------------------- | ---------------------- | ------------------------- |
| 1           | 26238464           | 24643564              | 667               | 574                  | 4                      | 4                         |
| 12          | 4053306            | 4320493               | 534               | 901                  | 3                      | 3                         |
| 24          | 2333593            | 2087698               | 728               | 215                  | 3                      | 3                         |
| 50          | 2245439            | 1003814               | 196               | 669                  | 3                      | 3                         |
| 100         | 2208063            | 482330                | 204               | 170                  | 3                      | 3                         |
| 200         | 2272190            | 280450                | 396               | 165                  | 3                      | 3                         |
| 300         | 1992442            | 215670                | 224               | 3703                 | 3                      | 3                         |
| 400         | 2191696            | 196756                | 192               | 17616                | 3                      | 3                         |
| 500         | 2325023            | 172394                | 208               | 10315                | 4                      | 3                         |


# With Limiter
```
go test -bench="BenchmarkWAL_WriteMulti" -run=^a -v -benchmem -memprofile mem_with_limiter.prof
go tool pprof mem_with_limiter.prof
```
image
image



# Without Limiter
```
go test -bench="BenchmarkWAL_WriteMulti" -run=^a -v -benchmem -memprofile mem_without_limiter.prof
go tool pprof -http :9091 mem_without_limiter.prof
```
image
image


# Attachment
[profile.zip](https://github.com/influxdata/influxdb/files/6220700/profile.zip)

danxmoran:
Thanks for your patience @StoneYunZhao. I've gotten a :+1: offline from another Influxer, so I'm inclined to merge this as-is. I pinged a few others while waiting on reviews so I'll wait to see if any of them speak up. If not, I'll merge this at my EOD.

tsdb/engine/tsm1/wal.go
```diff
@@ -409,6 +411,9 @@ func (l *WAL) writeToLog(entry WALEntry) (int, error) {
 	// limit how many concurrent encodings can be in flight.  Since we can only
 	// write one at a time to disk, a slow disk can cause the allocations below
 	// to increase quickly.  If we're backed up, wait until others have completed.
+	l.limiter.Take()
```
danxmoran, line 414: Note from @lesam: We'd feel more confident about this if `Take` supported a timeout parameter. We're ok adding that in a follow-up PR since the improvements of the current code seem solid.

danxmoran (approved):

PR #20818, opened 2021-03-01
refactor: delete unused v1 monitoring service

danxmoran:
In 1.x, this service would periodically poll system stats and write them to the `_internal` database. @russorat confirmed this behavior was purposefully dropped in 2.x.

Outside of the `v1/monitor` module, the only usage of the deleted code was references to the `Diagnostics` type. The methods referencing that type were only called by functions in the monitoring service, so they could be cleanly deleted as well.

NOTE: This system also populated the global `expvar` contents while processing stats. 2.x's `/debug/vars` output has been pretty minimal because of these missing values. We intend to disable `/debug/vars` entirely, and add any missing vars to the output of `/metrics`.


docmerlin (approved):
BURN IT DOWN.

PR #20827, opened 2021-03-02
feat(http): allow for disabling pprof

### LINKED ISSUE #20764: i would like to have the ability to disable the pprof endpoint
in 1.x, we had the following config options for controlling when the pprof endpoint was enabled:

```
  pprof-enabled = true
  pprof-auth-enabled = false
  debug-pprof-enabled = false
```

I'm not sure what the difference is between `pprof-enabled` and `debug-pprof-enabled` since there is no `/pprof` endpoint and only `pprof-enabled` seemed to have any effect.

As an oss user, i would like a config option to disable `/debug/pprof` in the latest influxdb. we can use the same name as `pprof-enabled: true` as the default. When set to false, `/debug/pprof` path should return a 404 and the profiles should not be generated.

danxmoran:
Closes #20764

The big chunks of new logic are the `/all` profile, ported from 1.x. I didn't port the pieces of that code that gathered InfluxQL statistics, since we have other endpoints & tooling for scraping that info.

As a side-effect of switching away from using the default Mux, the `/debug/vars` endpoint is no longer exposed. Those vars were mostly empty in 2.x since we turned off the internal diagnostics service. The `/metrics` endpoint is the intended replacement.

### DIFF
--- a/chronograf/server/mux.go
+++ b/chronograf/server/mux.go
@@ -8,8 +8,6 @@ import (
 	"strconv"
 	"strings"
 
-	_ "net/http/pprof"
-
 	"github.com/NYTimes/gziphandler"
 	"github.com/bouk/httprouter"
 	jhttprouter "github.com/influxdata/httprouter"
@@ -33,7 +31,6 @@ type MuxOpts struct {
 	ProviderFuncs []func(func(oauth2.Provider, oauth2.Mux))
 	StatusFeedURL string            // JSON Feed URL for the client Status page News Feed
 	CustomLinks   map[string]string // Any custom external links for client's User menu
-	PprofEnabled  bool              // Mount pprof routes for profiling
 }
 
 // NewMux attaches all the route handlers; handler returned servers chronograf.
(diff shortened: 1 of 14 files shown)


chronograf/server/mux.go
```diff
@@ -131,11 +128,6 @@ func NewMux(opts MuxOpts, service Service) http.Handler {
 		)
 	}
 
-	if opts.PprofEnabled {
-		// add profiling routes
-		router.GET("/debug/pprof/:thing", http.DefaultServeMux.ServeHTTP)
-	}
```
danxmoran, line 137: AFAIK this mux isn't actually connected to anything, so this is just cleanup/consolidation

davidby-influx (approved):
LGTM

cmd/influxd/upgrade/config_test.go
```diff
@@ -438,6 +438,7 @@ storage-validate-keys = false
 storage-wal-fsync-delay = "0s"
 tls-cert = "/etc/ssl/influxdb.pem"
 tls-key = ""
+pprof-enabled = true
```
danxmoran, line 441: Agh, this needs updated to match the new flag name... or I should revert back to `pprof-enabled` for the 2.x flag

PR #20585, opened 2021-01-25
fix(tsm): should not check write-ahead-log size against default size

foobar:
it should check against the local saved SegmentSize instead of the
default const DefaultSegmentSize.

PR #20839, opened 2021-03-02
fix(tsm): use configured value instead of default when checking WAL segment size

### LINKED ISSUE #20742: WAL segment size-check compares against default instead of configured value [backport 2.0.x]
Backport #20585 to the `2.0` branch

danxmoran:
Closes #20742

Backports #20585


psteinbachs (approved):
Good to go.

PR #20843, opened 2021-03-02
fix(tsm1): fix data race and validation in cache ring

### LINKED ISSUE #20801: Fix data race and validation in cache ring [backport 2.0.x]
Backport #20172 and #20803 to the `2.0` branch

danxmoran:
Closes #20801 

Backports #20797, #20803, and #20890


StoneYunZhao:
I'm curious if need to delete `(*Cache) Write([]byte, []Value)` method in `cache.go` too.

danxmoran:
> I'm curious if need to delete (*Cache) Write([]byte, []Value) method in cache.go too.

I'll investigate, and open a follow-up PR if so. In general I'd prefer to keep these backports as close to the original merges as possible, otherwise we need to track forward-ports in addition to backports and it all becomes a mess.

psteinbachs (approved):
Good to go.

Issue #19490, opened 2020-09-02
influx bucket find does not output more than 20 buckets


jsternberg:
__Steps to reproduce:__
List the minimal actions needed to reproduce the behavior.

1. Create 30 buckets
2. Run `influx bucket find` on the organization
3. The number of lines will be 21 (one header line and 20 results)

__Expected behavior:__
I expected all buckets to be output.

__Actual behavior:__
Only 20 buckets were returned.

wbaker85:
This should be fixed by https://github.com/influxdata/influx-cli/pull/307, which implemented bucket pagination in the CLI.

PR #20847, opened 2021-03-02
feat: add limit option to influx bucket list

russorat:
Closes https://github.com/influxdata/influxdb/issues/19490

- [ ] Rebased/mergeable
- [ ] Tests pass

### DIFF
--- a/cmd/influx/bucket.go
+++ b/cmd/influx/bucket.go
@@ -10,6 +10,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const DefaultLimit = 20
+
 type bucketSVCsFn func() (influxdb.BucketService, influxdb.OrganizationService, error)
 
 func cmdBucket(f *globalFlags, opt genericCLIOpts) *cobra.Command {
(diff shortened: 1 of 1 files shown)


cmd/influx/bucket.go
```diff
@@ -179,6 +183,13 @@ func (b *cmdBucketBuilder) cmdList() *cobra.Command {
 			EnvVar: "BUCKET_NAME",
 			Desc:   "The bucket name",
 		},
+		{
+			DestP:  &b.limit,
+			Flag:   "limit",
+			Short:  'l',
+			EnvVar: "LIMIT",
+			Desc:   "The max number of results to return (must be < 100)",
```
danxmoran, line 191: We should probably add this to all the find / list commands

danxmoran: What happens if a user passes > 100? Does the server just return the first 100 things, or does something go boom?

russorat: they get a message: `Error: Failed to retrieve buckets: limit must be between 1 and 100.`

russorat:
I agree we should add this to all list commands. I'm not skilled enough to do that very well though.

russorat:
i'm going to add this in a few more places, and add an offset option as well

wbaker85:
Closing this since https://github.com/influxdata/influx-cli/pull/307 implemented bucket pagination in the CLI.

PR #20862, opened 2021-03-04
fix: infinite recursion bug

### LINKED ISSUE #20249: unsupported value: +Inf" error not handled gracefully

lesam:
Closes #20249

We should avoid infinite recursion!

- [x] Rebased/mergeable
- [x] Tests pass

### DIFF
--- a/services/httpd/response_writer.go
+++ b/services/httpd/response_writer.go
@@ -3,6 +3,7 @@ package httpd
 import (
 	"encoding/csv"
 	"encoding/json"
+	"errors"
 	"io"
 	"net/http"
 	"strconv"
@@ -122,8 +123,9 @@ type jsonFormatter struct {
 	Pretty bool
 }
 
-func (f *jsonFormatter) WriteResponse(w io.Writer, resp Response) (err error) {
+func (f *jsonFormatter) WriteResponse(w io.Writer, resp Response) error {
 	var b []byte
+	var err error
 	if f.Pretty {
 		b, err = json.MarshalIndent(resp, "", "    ")
 	} else {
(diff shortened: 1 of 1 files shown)


services/httpd/response_writer.go
```diff
@@ -98,13 +98,9 @@ func (w *bytesCountWriter) Write(data []byte) (int, error) {
 
 // WriteResponse writes the response using the formatter.
 func (w *responseWriter) WriteResponse(resp Response) (int, error) {
-	n := 0
 	writer := bytesCountWriter{w: w.ResponseWriter}
 	err := w.formatter.WriteResponse(&writer, resp)
-	if err != nil {
-		n, _ = WriteError(w, err)
```
lesam, line 105: WriteError calls WriteResponse, and if the ResponseWriter is bad and returning errors, that calls back to WriteError. This was introduced recently, never released.

davidby-influx (dismissed):
LGTM.  Thanks!

services/httpd/response_writer.go
```diff
@@ -131,15 +132,31 @@ func (f *jsonFormatter) WriteResponse(w io.Writer, resp Response) (err error) {
...
+			b, _ = json.MarshalIndent(resp, "", "    ")
+		} else {
+			b, _ = json.Marshal(resp)
+		}
+		w.Write(b)
+		w.Write([]byte("\n"))
+		return err
 	}
 
+	_, err = w.Write(b)
```
davidby-influx, line 148: Now wondering if we should check for an error on the w.Write ([]byte("\n") call.  Or maybe append the \n and only have one write (in both places it is done).

davidby-influx (dismissed):
One comment to ponder, but not blocking.

services/httpd/response_writer.go
lesam: Fixed the error check

davidby-influx (approved):
All the right errors returned, all the wrong errors ignored!

Issue #20491, opened 2021-01-11
Convert panics in kit/cli to returned errors


danxmoran:
Panics make users think things have _really_ gone wrong, but many failure cases in the CLI code are possible to hit from bad environment variables / config / user inputs. Converting to use returned errors should provide a cleaner UX in the failure case.

Things TODO:
- [ ] Convert the panics in `kit/cli/viper.go` to return errors instead of panic-ing
- [ ] Convert callsites of functions from `kit/cli/viper.go` to return errors (or sys.exit) instead of panic-ing

PR #20863, opened 2021-03-04
refactor: remove panics from CLI kit code and influxd CLI setup

danxmoran:
1st "half" of #20491

Return errors from `kit/cli` code instead of using `panic`. The impact was pretty minimal outside of `cmd/influx`, so I've included the fixed up code here. The impact on the `influx` CLI is very extensive, so I've shimmed a `panic` back in at that layer for now. Will open a follow-up to finish the job after this one merges, in case we decide it's not worth the diff.


davidby-influx (approved):
LGTM

PR #20873, opened 2021-03-05
feat: add arguments to flux to function

### LINKED ISSUE #20853: 'To' flux function should support more arguments
__Proposal:__
In https://github.com/influxdata/influxdb/pull/20767 , we added support for the `to` function in 1.x, with `bucket` as the only supported argument. We should support at least `timeColumn`, `tagColumns`, `fieldFn` as documented at https://docs.influxdata.com/influxdb/v2.0/reference/flux/stdlib/built-in/outputs/to/ . It should also support `measurementColumn` in the same way as 2.x .

__Desired behavior:__
`to` should support `timeColumn`, `tagColumns`, `fieldFn` as documented at https://docs.influxdata.com/influxdb/v2.0/reference/flux/stdlib/built-in/outputs/to/ , and also `measurementColumn` to set the measurement column.

__Alternatives considered:__
n/a

__Use case:__
Often data will not have exactly _measurement, _field, _value, and _time columns. We should support this use case.

lesam:
Also remove some dead code. 

Adds support for timeColumn, tagColumns, fieldFn, measurementColumn. See to_test.go  or the issue description for details.

Closes #20853

- [x] Rebased/mergeable
- [x] Tests pass


flux/client/http.go
```diff
@@ -1,57 +0,0 @@
-package client
```
lesam, line 1: This code was dead

Dockerfile_build_ubuntu64_go1.13
```diff
@@ -24,7 +24,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y \
 
 RUN gem install fpm
 
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain stable -y
```
lesam, line 27: We should pull a stable rust version. Ideally we would either save the builder Dockerfile, or hard-code a rust version, but that's for the future.

flux/stdlib/influxdata/influxdb/to.go
```diff
@@ -147,10 +186,57 @@ func NewToTransformation(ctx context.Context, d execute.Dataset, cache execute.T
...
+		// is referenced in the function expression from being a tag column.
+		if toSpec.Spec.FieldFn.Fn != nil {
+			recordParam := toSpec.Spec.FieldFn.Fn.Parameters.List[0].Key.Name
+			exprNode := toSpec.Spec.FieldFn.Fn
+			colVisitor := newFieldFunctionVisitor(recordParam)
+
+			// Walk the field function expression and record which columns
+			// are referenced. None of these columns will be used as tag columns.
+			semantic.Walk(colVisitor, exprNode)
+			for k, v := range colVisitor.captured {
```
lesam, line 216: Note that 2.x delays this processing to `Process` so it can evaluate the captured columns against the actual table columns. That seems unnecessary to me, much nicer to do it up front.

flux/stdlib/influxdata/influxdb/to.go
```diff
@@ -217,12 +300,12 @@ func (t *ToTransformation) Process(id execute.DatasetID, tbl flux.Table) error {
 			measurementName := ""
 			fields := make(models.Fields)
 			var pointTime time.Time
-			kv = kv[0:]
+			kv = kv[:0]
```
lesam, line 303: This was obviously bad but happened to pretty often work properly, probably due to stable-ish sorting somewhere. The desired behaviour is truncation, not a no-op!!

flux/stdlib/influxdata/influxdb/to_test.go
```diff
@@ -119,6 +127,622 @@ c _value=4 41`),
 				}},
 			},
 		},
+		{
+			name: "default with heterogeneous tag columns",
```
lesam, line 131: New tests taken from 2.x

flux/stdlib/influxdata/influxdb/to.go
docmerlin: nice catch

docmerlin (approved):

PR #20875, opened 2021-03-05
fix: make AST-based option parsing the default, remove feature flag

### LINKED ISSUE #17666: error in task editor when using tableFind()
I've created a task with the below query and getting error when using tableFind().Attaching the screenshot of the error. Shouldn't be tableFind() used in task?

data = from(bucket: "flowEvents_1hr")
	|> range(start: -5h, stop: -4h)
	|> tableFind(fn: (key) => key._field == "c")

image

danxmoran:
Closes #17666

The fix for this issue was already checked in (#19326), but it was hidden behind a feature flag. This removes the feature flag, and makes the gated behavior the default.


ui/cypress/e2e/tasks.test.ts
```diff
@@ -24,25 +24,6 @@ describe('Tasks', () => {
     })
   })
 
-  it('cannot create a task with an invalid to() function', () => {
```
danxmoran, line 27: This test was also deleted from Cloud (I assume for the same reasons), see influxdata/ui@ca989db68d56a4ab36f871a7745c1608c48f0b6b

brettbuddin (approved):

PR #20882, opened 2021-03-06
fix: query regex condition rewrite not working in subquery

### LINKED ISSUE #20881: influxQL query regex condition rewrite not working in subquery
__Proposal:__
Also rewrite subquery regex conditions into static OR query for better series look up performance.
Not just rewrite regex for root query condition

__Current behavior:__
Currently, for example this query can be convert to multi string literal condition after compile

```sql
SELECT mean(usage) FROM cpu WHERE id =~ /^(foo|bar)$/ GROUP BY host
=> SELECT mean(usage) FROM cpu WHERE id = 'foo' OR id = 'bar' GROUP BY host
```

This is helpful in faster series key looking up in index.
However, if this query is subquery, the compiler won't convert the regex into multi string literal OR condition, for example

```sql
SELECT TOP(mean, 10), host FROM (SELECT mean(usage) FROM cpu WHERE id =~ /^(foo|bar)$/ GROUP BY host)
```

Because it is not root query condition so that the rewrite logic is not apply to subquery. So the subquery above will not benefit from the regex rewrite. WHAT a pity...

The code below in MARK 1 only rewrite its own condition. In subquery case, the subquery is identify as sources. So the rewrite is not working for subquery!

```go
func Compile(stmt *influxql.SelectStatement, opt CompileOptions) (Statement, error) {
	c := newCompiler(opt)
	c.stmt = stmt.Clone()
	if err := c.preprocess(c.stmt); err != nil {
		return nil, err
	}
	if err := c.compile(c.stmt); err != nil {
		return nil, err
	}
	c.stmt.TimeAlias = c.TimeFieldName
	c.stmt.Condition = c.Condition

	// Convert DISTINCT into a call.
	c.stmt.RewriteDistinct()

	// Remove "time" from fields list.
	c.stmt.RewriteTimeFields()

	// Rewrite any regex conditions that could make use of the index.
	c.stmt.RewriteRegexConditions()  // ====== MARK 1 ======

	return c, nil
}
```

__Desired behavior:__
We expect that the subquery can also be benefit from the regex rewrite. DEAD simple.

__Use case:__
Why is this important (helps with prioritizing requests)?
Subquery regex can also be formatted into a better AST tree for better query performance. Especially in grafana multi values query.

yujiahaol68:
Closes #20881

- [x] Rebased/mergeable
- [x] Tests pass


CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ the endpoint has been removed. Use the `/metrics` endpoint to collect system sta
 1. [20809](https://github.com/influxdata/influxdb/pull/20809): Fix InfluxDB port in Flux function UI examples. Thanks @sunjincheng121!
 1. [20827](https://github.com/influxdata/influxdb/pull/20827): Remove unauthenticated, unsupported `/debug/vars` HTTP endpoint.
 1. [20856](https://github.com/influxdata/influxdb/pull/20856): Respect 24 hour clock formats in the UI and allow more choices
+1. [20882](https://github.com/influxdata/influxdb/pull/20882): Fix query regex condition rewrite not working in subquery. Thanks @yujiahaol68!
```
danxmoran, line 36: I'd suggest moving this up to be a feature, and rewording:
```suggestion
1. [20882](https://github.com/influxdata/influxdb/pull/20882): Rewrite regex conditions in InfluxQL subqueries for performance. Thanks @yujiahaol68!
```

influxql/query/compile.go
```diff
@@ -189,6 +189,7 @@ func (c *compiledStatement) compile(stmt *influxql.SelectStatement) error {
 			if err := c.subquery(source.Statement); err != nil {
 				return err
 			}
+			source.Statement.RewriteRegexConditions()
```
danxmoran, line 192: Can you add a test that shows this works as expected? We should at least add a case to `TestCompile_Success` in `influxql/query/compile_test.go` to ensure this isn't breaking any compilation. Ideally we'd add a new, separate test case showing the output of compilation actually contains the expected rewrite.

yujiahaol68: Yeah, I think it should be a seperate new test case in `compile_test.go`. I will add them.

yujiahaol68: I add a subquery statement test case with regex condition. And thought a lot about adding a seperate test case but figure out that `query_test` is only test the final statement and compiler is working fine. It is hard to diff the ast tree outside the statement interface to make sure it works inside AST. And `RewriteRegexConditions()` is already tested by influxQL package. So the rewrite is guarantee by the package that it apply to all select statement. Any better suggestion or solution?

danxmoran: I don't think I can push the change directly to your branch since it's in a forked repo; can you add the file above?

yujiahaol68: This test file has been added. Great job !!

danxmoran (approved):

